Issue #7172, opened 2021-05-05
[Tracking issue] New lint metadata collection (a metadata collection monster)


xFrednet:
This issue tracks the effort to replace the current lint documentation collection with a new Rust implementation. The implementation will be based on a lint to access rustc's type information (See #4310 for further information). The first incomplete version of *the metadata collection monster* (as I like to call it) got merged in #6887.

### To-Do list
* [x] Collect lint metadata
    * [x] The lint ID
    * [x] The lint declaration file and location (for #1303)
    * [x] The lint group
    * [x] The documentation
    * [x] The applicability (if resolvable)
    * [x] Is the suggesting a multi-part-suggestion? (if resolvable) 
* [x] Collect deprecated lints 
    * [draft](https://github.com/xFrednet/rust-clippy/tree/4310-depreciated-lints-collection)
    * PR #7197
    * backup implementation to collect from `store.register_removed` [commit](https://github.com/xFrednet/rust-clippy/commit/90edfd514624b979f99d3f133f968dd19d908d09)
* [x] Collect lint configuration values 
    * PR #7214
* [x] Create a simple way to run the lint collector, maybe via a cargo alias
    * PR #7217
* [x] Rerun the collection after touching the `clippy_lints/src/lib.rs` if nothing was collected due to caching (See comments in this issue)
  * PR #7256
* [x] Remove `clippy::` prefix from lint group and add the lint level 
  * PR #7246
* [x] Switch the CI over to use the new collection
  * Discussion about this in #7229
  * PR #7298
* [x] Adapt [Clippy's lint list](https://rust-lang.github.io/rust-clippy/master/index.html) to support the new metadata format (Example output file: [`metadata_collection.json`](https://gist.github.com/xFrednet/6b9e2c3f725f476ba88db9563f67e119))
  * PR #7279
* [x] Remove old implementation
* [x] Collect renamed lints (https://github.com/rust-lang/rust-clippy/pull/8843 Thanks to @Serial-ATA)
* [x] Update documentation
    * [x] Lint configuration doc comment schema
    * [x] Registering deprecated lints
    * [x] Update configuration section template to use `### Configuration` instead of `**Configuration**`
* [x] Collect implementation version (See #6492)

### Needs investigation / discussion
[...]

PR #7197, opened 2021-05-08
Metadata collection monster eating deprecated lints

xFrednet:
This adds the collection of deprecated lints to the metadata collection monster. The JSON output has the same structure with the *new* lint group "DEPRECATED". Here is one of fourteen examples it was able to dig up in Clippy's code:

```JSON
  {
    "id": "assign_op_pattern",
    "id_span": {
      "path": "src/assign_ops.rs",
      "line": 34
    },
    "group": "clippy::style",
    "docs": " **What it does:** Checks for `a = a op b` or `a = b commutative_op a` patterns.\n\n **Why is this bad?** These can be written as the shorter `a op= b`.\n\n **Known problems:** While forbidden by the spec, `OpAssign` traits may have\n implementations that differ from the regular `Op` impl.\n\n **Example:**\n ```rust\n let mut a = 5;\n let b = 0;\n // ...\n // Bad\n a = a + b;\n\n // Good\n a += b;\n ```\n",
    "applicability": {
      "is_multi_part_suggestion": false,
      "applicability": "MachineApplicable"
    }
  }
```

And you! Yes you! Sir or Madam can get all of this **for free** in Clippy if this PR gets merged. (Sorry for the silliness ^^)

---

See: #7172 for the full metadata collection to-do list or to suggest a new feature in connection to it :upside_down_face: 

---

changelog: none

r? @flip1995

### DIFF
--- a/clippy_lints/src/deprecated_lints.rs
+++ b/clippy_lints/src/deprecated_lints.rs
@@ -1,6 +1,13 @@
+/// This struct fakes the `Lint` declaration that is usually created by `declare_lint!`. This
+/// enables the simple extraction of the metadata without changing the current depreciation
+/// declaration.
+pub struct ClippyDeprecatedLint;
+
 macro_rules! declare_deprecated_lint {
-    (pub $name: ident, $_reason: expr) => {
-        declare_lint!(pub $name, Allow, "deprecated lint")
+    { $(#[$attr:meta])* pub $name: ident, $_reason: expr} => {
+        $(#[$attr])*
+        #[allow(dead_code)]
+        pub static $name: ClippyDeprecatedLint = ClippyDeprecatedLint {};
     }
 }
 
--- a/clippy_lints/src/deprecated_lints.rs
+++ b/clippy_lints/src/deprecated_lints.rs
@@ -1,5 +1,5 @@
 /// This struct fakes the `Lint` declaration that is usually created by `declare_lint!`. This
-/// enables the simple extraction of the metadata without changing the current depreciation
+/// enables the simple extraction of the metadata without changing the current deprecation
 /// declaration.
 pub struct ClippyDeprecatedLint;
 
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -162,6 +162,8 @@ macro_rules! extract_msrv_attr {
 mod consts;
 #[macro_use]
 mod utils;
+#[cfg(feature = "metadata-collector-lint")]
+mod deprecated_lints;
 
 // begin lints modules, do not remove this comment, it’s used in `update_lints`
 mod absurd_extreme_comparisons;
--- a/clippy_lints/src/utils/internal_lints/metadata_collector.rs
+++ b/clippy_lints/src/utils/internal_lints/metadata_collector.rs
@@ -28,7 +28,7 @@ use std::path::Path;
 
 use crate::utils::internal_lints::is_lint_ref_type;
 use clippy_utils::{
-    diagnostics::span_lint, last_path_segment, match_function_call, match_path, paths, ty::match_type,
+    diagnostics::span_lint, last_path_segment, match_def_path, match_function_call, match_path, paths, ty::match_type,
     ty::walk_ptrs_ty_depth,
 };
 
@@ -41,6 +41,8 @@ const BLACK_LISTED_LINTS: [&str; 3] = ["lint_author", "deep_code_inspection", "i
 const IGNORED_LINT_GROUPS: [&str; 1] = ["clippy::all"];
 /// Lints within this group will be excluded from the collection
 const EXCLUDED_LINT_GROUPS: [&str; 1] = ["clippy::internal"];
+/// Collected deprecated lint will be assigned to this group in the JSON output
+const DEPRECATED_LINT_GROUP_STR: &str = "DEPRECATED";
 
 const LINT_EMISSION_FUNCTIONS: [&[&str]; 7] = [
     &["clippy_utils", "diagnostics", "span_lint"],
@@ -66,6 +68,7 @@ const SUGGESTION_FUNCTIONS: [&[&str]; 2] = [
     &["clippy_utils", "diagnostics", "multispan_sugg"],
     &["clippy_utils", "diagnostics", "multispan_sugg_with_applicability"],
 ];
+const DEPRECATED_LINT_TYPE: [&str; 3] = ["clippy_lints", "deprecated_lints", "ClippyDeprecatedLint"];
 
 /// The index of the applicability name of `paths::APPLICABILITY_VALUES`
 const APPLICABILITY_NAME_INDEX: usize = 2;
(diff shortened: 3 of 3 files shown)


xFrednet:
Ups, I'll update the PR later today. Thank you 😅

xFrednet:
I've fixed the wrong vocab usage in the main commit itself. I don't think it was worth it to add a new one :upside_down_face:

clippy_lints/src/deprecated_lints.rs
```diff
@@ -1,6 +1,13 @@
+/// This struct fakes the `Lint` declaration that is usually created by `declare_lint!`. This
+/// enables the simple extraction of the metadata without changing the current depreciation
```
flip1995, line 2:
```suggestion
/// enables the simple extraction of the metadata without changing the current deprecation
```

clippy_lints/src/lib.rs
```diff
@@ -162,6 +162,8 @@ macro_rules! extract_msrv_attr {
 mod consts;
 #[macro_use]
 mod utils;
+#[cfg(feature = "metadata-collector-lint")]
+mod deprecated_lints;
```
flip1995, line 166: Wait, this module was dead the whole time? :open_mouth: In that case: couldn't you collect the necessary metadata from the `register_removed` calls, so we could remove this module once we completely switch over to this metadata collector?

xFrednet: Yes it was only used by the collection script ^^.

Well, I decided to do it this way because [`deprecated_lints.rs`](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/deprecated_lints.rs) contains a more in-depth explanation why these lints have been depreciated in comparison to our usage of [store.register_removed](https://github.com/rust-lang/rust-clippy/blob/65951c969ff843bc3a45c470d5f366498172ebe1/clippy_lints/src/lib.rs#L444). But that could easily be updated.

However, I think there is currently no way to retrieve the deprecation reason from the `LintStore`. Removed lints are only stored in the [`by_name`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/struct.LintStore.html#structfield.by_name) field which can not be accessed directly. The function [`find_lints`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_lint/struct.LintStore.html#method.find_lints) can be used to access the `by_name` field but will only return `Err(FindLintError::Removed)` without the reason itself (See [rustc_lint/context.rs](https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_lint/context.rs.html#314-333)). I could create an issue or PR in [rust-lang/rust](https://github.com/rust-lang/rust) to maybe add a new function to the `LintStore` for this. That might also be cleaner when I think about it, but it would probably take some time until that would be implemented and synced back. I could add it to #7172 as a new to-do item. What is your opinion on the matter?

clippy_lints/src/deprecated_lints.rs
xFrednet: Typo number two. One day I'll create a PR without one :joy:

clippy_lints/src/lib.rs
xFrednet: Yes, the implementation uses the `LintStore` to collect the lint group, as it was the easiest way to retrieve it. ([here](https://github.com/rust-lang/rust-clippy/blob/72be4764c57266dc1e71d32a06619d302b3253a4/clippy_lints/src/utils/internal_lints/metadata_collector.rs#L337))

Checking for `register_removed` would also work but checking each expression seemed like an unnecessary overhead since they all get registered in the store already anyways :thinking:

flip1995: > Yes, the implementation uses the `LintStore` to collect the lint group, as it was the easiest way to retrieve it.

Ah, ok. In that case my suggestion to only register this lint, when collecting metadata won't work.

---

But IIUC you're not using the `LintStore` for the deprecated lints anyway? So I don't think it would make a difference if you do this in `check_expr` or in `check_item` and match on the method call instead of the static item.

xFrednet: > Ah, ok. In that case my suggestion to only register this lint, when collecting metadata won't work.

There are other ways to collect the lint group so it's not impossible. And your suggestion was focused on performance, we can simply skip the registration of the actual lint pass (`register_early_pass` and `register_late_pass`). Correct?

---

> So I don't think it would make a difference if you do this in `check_expr` or in `check_item` and match on the method call instead of the static item. 

Correct, I take it that you would prefer the expression via `check_expr`?

flip1995: > we can simply skip the registration of the actual lint pass (`register_early_pass` and `register_late_pass`).

Yeah, right. I always forget that registering lints and registering passes is done separately. 



> Correct, I take it that you would prefer the expression via `check_expr`?

Since we can get rid of a whole dead module that way (not immediately, but in the near future), yes definitely.

clippy_lints/src/lib.rs
```diff
@@ -447,59 +447,80 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     // begin deprecated lints, do not remove this comment, it’s used in `update_lints`
     store.register_removed(
         "clippy::should_assert_eq",
-        "`assert!()` will be more flexible with RFC 2011",
+        "This used to check for `assert!(a == b)` and recommend replacement with \
```
flip1995, line 450: These messages are displayed when the lint is used in the code. So we have to apply the same rules as to every other suggestion: start lowercase and don't put punctuation  at the end.

Those are also tested in our UI tests, so you will have to bless them.

flip1995 (commented):
Hm, now that I see the effect of this, I think having a more detailed documentation in addition to the reason displayed in the error message is a better approach. Otherwise we would dump a multi line error message if someone uses a deprecated lint. 

Before you start working on `clippy_dev`, just revert this to the version you had before my silly idea and lets revisit this later. Sorry about that!

xFrednet:
Yes, I had the same gut feeling during the implementation. Documenting deprecated lints in their own module makes them more approachable to newcomers as well due to the reduced scope, IMO. Anyways, This implementation will be kept in a backup branch in case we need it at a later point in time. I'll remove the commits from this branch. 

Was there anything else I should update in the original implementation?

flip1995:
> Was there anything else I should update in the original implementation?

Nope, the original implementation looked good.

xFrednet:
Okay, I've removed the commits to collect the deprecation reasoning from `store.register_removed`. A backup is in a second branch with a link to it in the tracking issue. So, that should be it IMO.

flip1995:
@bors r+

Thanks! (And sorry for the back and forth)

xFrednet:
No worries :upside_down_face:, and thanks for the quick review cycle. I'll create the configuration collection PR later today and that should be it when it comes to the new collection itself. The next step it then to adapt the CI and website. I'm also looking forward to work into a new topic after that xD

PR #7201, opened 2021-05-09
Remove powi, "square can be computed more efficiently"

### LINKED ISSUE #7058: "square can be computed more efficiently" is wrong
Lint name:
`suboptimal_flops`

I tried this code:

```rust
#![warn(clippy::suboptimal_flops)]

pub fn sample_sqr(x: f64) -> f64 {
    x.powi(2)
}
```

I expected to see this happen: it should pass.

Instead, this happened: it says "square can be computed more efficiently" and suggests to use `x * x` instead.

This is incorrect, because, unlike `powf`, `powi` produces exactly same native code as `x * x`, and is more natural to read (IMO). If anything, due to usage of intrinsic, it might be optimised even more than bare expression, although they're unlikely to be different on practice.

### Meta

- `cargo clippy -V`: clippy 0.1.53 (2e495d2 2021-04-08)
- `rustc -Vv`:
  ```
	  rustc 1.53.0-nightly (2e495d2e8 2021-04-08)
	  binary: rustc
	  commit-hash: 2e495d2e845cf27740e3665f718acfd3aa17253e
	  commit-date: 2021-04-08
	  release: 1.53.0-nightly
	  LLVM version: 12.0.0
  ```

mucinoab:
powi(2) produces exactly the same native code as x * x
powi was part of the [`suboptimal_flops`] lint

fixes #7058
changelog: Remove powi [`suboptimal_flops`], "square can be computed more efficiently"

### DIFF
--- a/clippy_lints/src/floating_point_arithmetic.rs
+++ b/clippy_lints/src/floating_point_arithmetic.rs
@@ -323,7 +323,7 @@ fn check_powi(cx: &LateContext<'_>, expr: &Expr<'_>, args: &[Expr<'_>]) {
                         cx,
                         SUBOPTIMAL_FLOPS,
                         parent.span,
-                        "square can be computed more efficiently",
+                        "multiply and add expressions can be calculated more efficiently and accurately",
                         "consider using",
                         format!(
                             "{}.mul_add({}, {})",
@@ -337,16 +337,6 @@ fn check_powi(cx: &LateContext<'_>, expr: &Expr<'_>, args: &[Expr<'_>]) {
                     return;
                 }
             }
-
-            span_lint_and_sugg(
-                cx,
-                SUBOPTIMAL_FLOPS,
-                expr.span,
-                "square can be computed more efficiently",
-                "consider using",
-                format!("{} * {}", Sugg::hir(cx, &args[0], ".."), Sugg::hir(cx, &args[0], "..")),
-                Applicability::MachineApplicable,
-            );
         }
     }
 }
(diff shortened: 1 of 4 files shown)


clippy_lints/src/floating_point_arithmetic.rs
```diff
@@ -297,60 +297,6 @@ fn check_powf(cx: &LateContext<'_>, expr: &Expr<'_>, args: &[Expr<'_>]) {
...
-                        "square can be computed more efficiently",
-                        "consider using",
-                        format!(
-                            "{}.mul_add({}, {})",
-                            Sugg::hir(cx, &args[0], ".."),
-                            Sugg::hir(cx, &args[0], ".."),
-                            Sugg::hir(cx, other_addend, ".."),
-                        ),
-                        Applicability::MachineApplicable,
-                    );
```
giraffate, line 335: Why does this case also removed?

mucinoab: It is just a special case for https://github.com/rust-lang/rust-clippy/blob/c8d52e35c10ca7cfa841309a7db4052344b710bf/clippy_lints/src/floating_point_arithmetic.rs#L403
Should we keep it?

giraffate: In the related issue, I found that the case of `x * x` was discussed, but I couldn't find the discussion of `mul_add`. So I don't know why this case needs to be removed as same.

If there is no reason, I think it would be better to keep it.

mucinoab:
@giraffate I did some changes. Should we keep the same names for everything?

giraffate (approved):
@mucinoab It looks good! Can you squash the commits and mark this PR as ready?

mucinoab:
@giraffate All done.

giraffate:
@bors r+

Thanks!

PR #7214, opened 2021-05-12
Metadata collection monster searching for Clippy's configuration options

xFrednet:
This PR teaches our lovely metadata collection monster which configurations are available inside Clippy. It then adds a new *Configuration* section to the lint documentation.

---

The implementation uses the `define_Conf!` macro to create a vector of metadata during compilation. This enables easy collection and parsing without the need of searching for the struct during a lint-pass (and it's quite elegant IMO). The information is then parsed into an intermediate struct called `ClippyConfiguration` which will be saved inside the `MetadataCollector` struct itself. It is currently only used to generate the *Configuration* section in the lint documentation, but I'm thinking about adding an overview of available configurations to the website. Saving them in this intermediate state without formatting them right away enables this in the future. 

The new parsing will also allow us to have a documentation that spans over multiple lines in the future. For example, this will be valid when the old script has been removed:
```rust
/// Lint: BLACKLISTED_NAME.
/// The list of blacklisted names to lint about. NB: `bar` is not here since it has legitimate uses
(blacklisted_names: Vec<String> = ["foo", "baz", "quux"].iter().map(ToString::to_string).collect())
``` 

The deprecation reason is also currently being collected but not used any further and that's basically it.

---

See: #7172 for the full metadata collection to-do list or to suggest a new feature in connection to it :upside_down_face: 

---

changelog: none

r? @flip1995
cc @camsteffen It would be great if you could also review this PR as you have recently worked on Clippy's `define_Conf!` macro.

### DIFF
--- a/clippy_lints/src/utils/conf.rs
+++ b/clippy_lints/src/utils/conf.rs
@@ -26,13 +26,13 @@ impl TryConf {
 
 macro_rules! define_Conf {
     ($(
-        #[$doc:meta]
+        #[doc = $doc:literal]
         $(#[conf_deprecated($dep:literal)])?
         ($name:ident: $ty:ty = $default:expr),
     )*) => {
         /// Clippy lint configuration
         pub struct Conf {
-            $(#[$doc] pub $name: $ty,)*
+            $(#[doc = $doc] pub $name: $ty,)*
         }
 
         mod defaults {
@@ -89,6 +89,24 @@ macro_rules! define_Conf {
                 Ok(TryConf { conf, errors })
             }
         }
+
+        #[cfg(feature = "metadata-collector-lint")]
+        pub mod metadata {
+            use crate::utils::internal_lints::metadata_collector::ClippyConfigurationBasicInfo;
+
+            pub(crate) fn get_configuration_metadata() -> Vec<ClippyConfigurationBasicInfo> {
+                vec![
+                    $(
+                        ClippyConfigurationBasicInfo {
+                            name: stringify!($name),
+                            config_type: stringify!($ty),
+                            default: stringify!($default),
+                            doc_comment: $doc,
+                        },
+                    )+
+                ]
+            }
+        }
     };
 }
 
@@ -100,7 +118,7 @@ define_Conf! {
     (blacklisted_names: Vec<String> = ["foo", "baz", "quux"].iter().map(ToString::to_string).collect()),
     /// Lint: COGNITIVE_COMPLEXITY. The maximum cognitive complexity a function can have
     (cognitive_complexity_threshold: u64 = 25),
-    /// DEPRECATED LINT: CYCLOMATIC_COMPLEXITY. Use the Cognitive Complexity lint instead.
+    /// Lint: CYCLOMATIC_COMPLEXITY. Use the Cognitive Complexity lint instead.
     #[conf_deprecated("Please use `cognitive-complexity-threshold` instead")]
     (cyclomatic_complexity_threshold: Option<u64> = None),
     /// Lint: DOC_MARKDOWN. The list of words this lint should not consider as identifiers needing ticks
--- a/clippy_lints/src/utils/conf.rs
+++ b/clippy_lints/src/utils/conf.rs
@@ -92,25 +92,26 @@ macro_rules! define_Conf {
 
         #[cfg(feature = "metadata-collector-lint")]
         pub mod metadata {
-            use crate::utils::internal_lints::metadata_collector::ClippyConfigurationBasicInfo;
+            use crate::utils::internal_lints::metadata_collector::ClippyConfiguration;
 
-            pub(crate) fn get_configuration_metadata() -> Vec<ClippyConfigurationBasicInfo> {
+            macro_rules! wrap_option {
+                () => (None);
+                ($x:literal) => (Some($x));
+            }
+
+            pub(crate) fn get_configuration_metadata() -> Vec<ClippyConfiguration> {
                 vec![
                     $(
                         {
-                            #[allow(unused_mut, unused_assignments)]
-                            let mut deprecation_reason = None;
+                            let deprecation_reason = wrap_option!($($dep)?);
 
-                            // only set if a deprecation reason was set
-                            $(deprecation_reason = Some(stringify!($dep));)?
-
-                            ClippyConfigurationBasicInfo {
-                                name: stringify!($name),
-                                config_type: stringify!($ty),
-                                default: stringify!($default),
-                                doc_comment: $doc,
+                            ClippyConfiguration::new(
+                                stringify!($name),
+                                stringify!($ty),
+                                format!("{:?}", super::defaults::$name()),
+                                $doc,
                                 deprecation_reason,
-                            }
+                            )
                         },
                     )+
                 ]
--- a/clippy_lints/src/utils/conf.rs
+++ b/clippy_lints/src/utils/conf.rs
@@ -97,11 +97,20 @@ macro_rules! define_Conf {
             pub(crate) fn get_configuration_metadata() -> Vec<ClippyConfigurationBasicInfo> {
                 vec![
                     $(
-                        ClippyConfigurationBasicInfo {
-                            name: stringify!($name),
-                            config_type: stringify!($ty),
-                            default: stringify!($default),
-                            doc_comment: $doc,
+                        {
+                            #[allow(unused_mut, unused_assignments)]
+                            let mut deprecation_reason = None;
+
+                            // only set if a deprecation reason was set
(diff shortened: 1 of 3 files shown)


clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -214,6 +258,107 @@ impl Serialize for ApplicabilityInfo {
...
+            let lints: Vec<String> = doc_comment.split_off(DOC_START.len()).split(", ").map(str::to_string).collect();
+
+            Some((lints, documentation))
+        } else {
+            None
+        }
+    }
+}
+
+/// Transforms a given `snake_case_string` to a tasty `kebab-case-string`
```
camsteffen, line 341: 😋

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -214,6 +258,107 @@ impl Serialize for ApplicabilityInfo {
...
+                deprecation_reason: x.deprecation_reason,
+            }
+        })
+        .collect()
+}
+
+fn clarify_default(default: &'static str) -> String {
+    if let Some((_start, init)) = default.split_once('[') {
+        if let Some((init, _end)) = init.split_once(']') {
+            return format!("[{}]", init);
```
camsteffen, line 305: Needs a comment explaining. This seems risky?

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -102,13 +127,32 @@ declare_clippy_lint! {
...
+            lints: BinaryHeap::<LintMetadata>::default(),
+            applicability_info: FxHashMap::<String, ApplicabilityInfo>::default(),
+            config: collect_configs(),
+        }
+    }
+
+    fn get_lint_configs(&self, lint_name: &str) -> Option<String> {
+        self.config
+            .iter()
+            .filter_map(|x| x.lints.iter().any(|x| x == lint_name).then(|| format!("{}", x)))
```
camsteffen, line 152: nit
```suggestion
            .filter(|config| config.lints.iter().any(|lint| lint == lint_name))
            .map(|config| config.to_string())
```

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -214,6 +258,107 @@ impl Serialize for ApplicabilityInfo {
...
+
+#[derive(Debug, Clone, Default)]
+struct ClippyConfiguration {
+    name: String,
+    lints: Vec<String>,
+    doc: String,
+    config_type: &'static str,
+    default: String,
+    deprecation_reason: Option<&'static str>,
+}
```
camsteffen, line 281: Would it be simpler to skip `ClippyConfigurationBasicInfo` and just create `ClippyConfiguration` in the macro?

camsteffen (commented):
Just some ignorable comments.

clippy_lints/src/utils/conf.rs
```diff
@@ -89,6 +89,33 @@ macro_rules! define_Conf {
...
+
+            pub(crate) fn get_configuration_metadata() -> Vec<ClippyConfigurationBasicInfo> {
+                vec![
+                    $(
+                        {
+                            #[allow(unused_mut, unused_assignments)]
+                            let mut deprecation_reason = None;
+
+                            // only set if a deprecation reason was set
+                            $(deprecation_reason = Some(stringify!($dep));)?
```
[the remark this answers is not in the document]
xFrednet: I was searching for a good way to do this, but I haven't found this version. I still find Rusts `macro_rules` a bit confusing TBH. I'll try your suggestion :)

clippy_lints/src/utils/internal_lints/metadata_collector.rs
xFrednet: This simply strips iterator and struct creation code from the default value. The current python script does the same via a regex. The result can be seen in this example: https://rust-lang.github.io/rust-clippy/master/#blacklisted_name

Documenting that in the code might be better though, fair point^^

What do you exactly mean with "rusky" btw?

clippy_lints/src/utils/internal_lints/metadata_collector.rs
xFrednet: The main reason why I have done it this way is to keep the monster contained to one module. It surely has its fangs in some other parts, but doing it this way keeps it mostly locked up.

But thinking about it, I could simply create a function in the module for it. *Sadly* a good suggestion as it requires some changes ^^

clippy_lints/src/utils/internal_lints/metadata_collector.rs
camsteffen: Risky since there could be multiple [ ] chars, possibly inside a string literal.

camsteffen: Could use rsplit_once for the right side.

xFrednet: I found that `format("{:?}", default)` will produce a well formatted and readable string for all of our configurations without any adjustments. So, I'll switch over to using that.

Sadly I'm on a different setup for this weekend that for some reason can't compile or even `cargo check` the project. (Some local miss configuration with windows and nightly). I'm currently checking the code in GitHub actions, but it might take some time, or I'll even wait until I'm back at my usual setup.

xFrednet:
I've addressed @camsteffen's suggestions which were excellent!  I'm happy with this version, but I was sadly unable to fully test it with my current setup. I'll be able to do so on Thursday. However, I strongly believe that this works and can therefor at least be reviewed if not merged :)

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -102,13 +127,33 @@ declare_clippy_lint! {
...
     /// All collected lints
     ///
     /// We use a Heap here to have the lints added in alphabetic order in the export
     lints: BinaryHeap<LintMetadata>,
     applicability_info: FxHashMap<String, ApplicabilityInfo>,
+    config: Vec<ClippyConfiguration>,
+}
+
+impl MetadataCollector {
+    pub fn new() -> Self {
```
flip1995, line 141: Doesn't this trigger our [``new_without_default``](https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default) lint?

flip1995 (approved):
> The new parsing will also allow us to have a documentation that spans over multiple lines in the future. For example, this will be valid when the old script has been removed:

That would be really neat, but I think the macro definition would have to be changed to allow multi-lint doc comments:

```
error: no rules expected the token `doc`
   --> clippy_lints/src/utils/conf.rs:126:5
    |
27  | macro_rules! define_Conf {
    | ------------------------ when calling this macro
...
126 |     /// FILTER_MAP_NEXT, CHECKED_CONVERSIONS, MANUAL_RANGE_CONTAINS, USE_SELF,
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no rules expected this token in macro call
```

LGTM. Works great for me locally. Just a NIT.

flip1995:
> However, I strongly believe that this works and can therefor at least be reviewed if not merged

I'm good with merging this. Feel free to approve this with `r=flip1995`, if you're also happy with this PR @xFrednet 

@bors delegate+

clippy_lints/src/utils/internal_lints/metadata_collector.rs
xFrednet: It's the first time that I've heard of that lint... so, I guess not 🤔.  Unless it got merged after this branch was crated from master. I would also expect it not to trigger as `new()` does more than just assign default values but that might be just my opinion 🙃 .

It didn't trigger during testing. We'll see if bors complains about it 🙃

xFrednet:
> > The new parsing will also allow us to have a documentation that spans over multiple lines in the future. For example, this will be valid when the old script has been removed:
>
> That would be really neat, but I think the macro definition would have to be changed to allow multi-lint doc comments:

That's true and done on purpose for now. The change is pretty simple but it would not be supported by our current python script. I therefore decided to leave the macro definition to be restrictive until we fully switched over. The parsing itself is already setup for it :)

I'm happy with merging this, and it also brings us one step closer to switching over. Thank you for the fast reviews on these PRs! @flip1995 

@bors r=flip1995,camsteffen

The bors syntax should be correct this time to add you as a reviewer, I've also added @camsteffen as he reviewed it as well 🙃

Issue #7006, opened 2021-03-30
clippy::option-if-let-else suggests a worse code.


fxdave:
Lint name: clippy::option-if-let-else

I tried this code:

```rust
if let Some(draggable) = utils::dom::find(target, "draggable") {
    // some code
} else if let Some(relation) = utils::dom::find(target, "relation") {
    // some code
} else {
    // some code
}
```

I expected to see no errors.

Instead, this happened:

```c
error: use Option::map_or instead of an if let/else
  --> src/components/viewport/mod.rs:78:12
   |
78 |       } else if let Some(relation) = utils::dom::find(target, "relation") {
   |  ____________^
...  |
90 | |     }
   | |_____^
   |
   = note: `-D clippy::option-if-let-else` implied by `-D clippy::pedantic`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
help: try
   |
78 |     } else { utils::dom::find(target, "relation").map_or(
              // some code,
               |relation| {
82 |         // some code
```

### Meta

- `cargo clippy -V`: clippy 0.1.53 (4a20eb6 2021-03-28)
- `rustc -Vv`:
  ```
  rustc 1.53.0-nightly (4a20eb6a9 2021-03-28)
  binary: rustc
  commit-hash: 4a20eb6a9da36c88ee929826c4f1eb8d7ea393b2
  commit-date: 2021-03-28
  host: x86_64-unknown-linux-gnu
  release: 1.53.0-nightly
  LLVM version: 12.0.0
  ```

magurotuna:
So do you mean that the suggested code works fine but feels worse than your original code, right?

fxdave:
Yes

camsteffen:
Yeah...I'd be inclined to just never lint `else if let`.

PR #7216, opened 2021-05-13
Stop linting `else if let` pattern in [`option_if_let_else`] lint

ThibsG:
For readability concerns, it is counterproductive to lint `else if let` pattern.
Unfortunately the suggested code is much less readable.

Fixes: #7006 

changelog: stop linting `else if let` pattern in [`option_if_let_else`] lint

### DIFF
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::usage::contains_return_break_continue_macro;
-use clippy_utils::{eager_or_lazy, get_enclosing_block, in_macro, is_lang_ctor};
+use clippy_utils::{eager_or_lazy, get_enclosing_block, in_macro, is_else_clause, is_lang_ctor};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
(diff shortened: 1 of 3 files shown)


clippy_lints/src/option_if_let_else.rs
```diff
@@ -161,13 +161,15 @@ fn detect_option_if_let_else<'tcx>(
...
         if let ExprKind::Match(cond_expr, arms, MatchSource::IfLetDesugar{contains_else_clause: true}) = &expr.kind;
+        if !is_else_clause(cx.tcx, expr);
         if arms.len() == 2;
         if !is_result_ok(cx, cond_expr); // Don't lint on Result::ok because a different lint does it already
         if let PatKind::TupleStruct(struct_qpath, &[inner_pat], _) = &arms[0].pat.kind;
         if is_lang_ctor(cx, struct_qpath, OptionSome);
         if let PatKind::Binding(bind_annotation, _, id, _) = &inner_pat.kind;
         if !contains_return_break_continue_macro(arms[0].body);
         if !contains_return_break_continue_macro(arms[1].body);
+
```
llogiq, line 172: What's that empty line for?

llogiq:
Regardless of the empty line, this is a good improvement. And I don't think an empty line should matter that much. So @bors r+ from my side.

clippy_lints/src/option_if_let_else.rs
ThibsG: Oh! that's en habit for readability, I should have discard it before pushing, sorry

Issue #6479, opened 2020-12-20
False positive of lint `similar-names` for identifiers with trailing underscore


fmease:
I tried this code:

```rust
#![allow(unused_variables)]
#![warn(clippy::similar_names)]

fn main() {
    let types = [1, 2, 3];
    let type_ = 3;
}
```

I expected Clippy to not emit any warnings since `types` and `type_` are visually distinct (in my eyes).
Instead, Clippy reported `similar_names` suggesting to change `type_` to `type__` which does not make much sense to me.

Edit: Together with `types`, I use the identifier `type_` quite a lot in my code base, which is a compiler. The unusual name obviously stems from the fact that `type` is a keyword in Rust and thus I cannot use it (but I prefer it to `typ` or `r#type`). All that to say, this is not an artificial corner case.

### Meta

* `clippy 0.0.212 (1c389ff 2020-11-24)`
* `rustc 1.50.0-nightly (1c389ffef 2020-11-24)`

<details>
<summary>Full Standard Error</summary>
<p>

```
warning: binding's name is too similar to existing binding
 --> src/main.rs:6:9
  |
6 |     let type_ = 3;
  |         ^^^^^
  |
note: the lint level is defined here
 --> src/main.rs:2:9
  |
2 | #![warn(clippy::similar_names)]
  |         ^^^^^^^^^^^^^^^^^^^^^
note: existing binding defined here
 --> src/main.rs:5:9
  |
5 |     let types = [1, 2, 3];
  |         ^^^^^
help: separate the discriminating character by an underscore like: `type__`
 --> src/main.rs:6:9
  |
6 |     let type_ = 3;
  |         ^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#similar_names

warning: 1 warning emitted
```

</p>
</details>

th1000s:
Fixed in PR #7221 by comparing it to a (hardcoded) list of known keywords.

arieluy:
Should this be closed?

fmease:
I guess it should.
Closed via #7221.

PR #7221, opened 2021-05-13
similar_names: No longer suggest inserting or appending an underscore

th1000s:
changelog: [`similar_names`] lint no longer suggests to insert or add an underscore to "fix" too similar names

### DIFF
--- a/clippy_lints/src/non_expressive_names.rs
+++ b/clippy_lints/src/non_expressive_names.rs
@@ -217,7 +217,6 @@ impl<'a, 'tcx, 'b> SimilarNamesNameVisitor<'a, 'tcx, 'b> {
             if allowed_to_be_similar(&interned_name, existing_name.exemptions) {
                 continue;
             }
-            let mut split_at = None;
             match existing_name.len.cmp(&count) {
                 Ordering::Greater => {
                     if existing_name.len - count != 1
@@ -268,7 +267,6 @@ impl<'a, 'tcx, 'b> SimilarNamesNameVisitor<'a, 'tcx, 'b> {
                                 // or too many chars differ (foo_x, boo_y) or (foox, booy)
                                 continue;
                             }
-                            split_at = interned_name.char_indices().rev().next().map(|(i, _)| i);
                         }
                     } else {
                         let second_i = interned_chars.next().expect("we know we have at least two chars");
@@ -281,7 +279,6 @@ impl<'a, 'tcx, 'b> SimilarNamesNameVisitor<'a, 'tcx, 'b> {
                             // or too many chars differ (x_foo, y_boo) or (xfoo, yboo)
                             continue;
                         }
-                        split_at = interned_name.chars().next().map(char::len_utf8);
                     }
                 },
             }
@@ -292,17 +289,6 @@ impl<'a, 'tcx, 'b> SimilarNamesNameVisitor<'a, 'tcx, 'b> {
                 "binding's name is too similar to existing binding",
                 |diag| {
                     diag.span_note(existing_name.span, "existing binding defined here");
-                    if let Some(split) = split_at {
-                        diag.span_help(
-                            ident.span,
-                            &format!(
-                                "separate the discriminating character by an \
-                                 underscore like: `{}_{}`",
-                                &interned_name[..split],
-                                &interned_name[split..]
-                            ),
-                        );
-                    }
                 },
             );
             return;
--- a/tests/ui/similar_names.stderr
+++ b/tests/ui/similar_names.stderr
@@ -10,11 +10,6 @@ note: existing binding defined here
    |
 LL |     let apple: i32;
    |         ^^^^^
-help: separate the discriminating character by an underscore like: `b_pple`
-  --> $DIR/similar_names.rs:15:9
-   |
-LL |     let bpple: i32;
-   |         ^^^^^
 
 error: binding's name is too similar to existing binding
   --> $DIR/similar_names.rs:17:9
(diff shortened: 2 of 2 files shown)


giraffate (commented):
Why do only specific keywords fixed here? It looks like that it's not limited to specific keywords. For example, `book__` is suggested in case below.
```
let books = ["a", "b", "c"];
let book_ = "a";
````

File: clippy_lints/src/non_expressive_names.rs [hunk 1]
```diff
@@ -168,6 +168,16 @@ fn allowed_to_be_similar(interned_name: &str, list: &[&str]) -> bool {
...
 }
 
+fn avoiding_keyword_clash(maybe_keyword: &str) -> bool {
+    let rust_keywords = [
+        "as", "break", "const", "continue", "crate", "else", "enum", "extern", "false", "fn", "for", "if", "impl",
+        "in", "let", "loop", "match", "mod", "move", "mut", "pub", "ref", "return", "self", "Self", "static", "struct",
+        "super", "trait", "true", "type", "unsafe", "use", "where", "while", "async", "await", "dyn", "abstract",
+        "become", "box", "do", "final", "macro", "override", "priv", "typeof", "unsized", "virtual", "yield", "union",
+    ];
+    rust_keywords.iter().any(|kw| *kw == maybe_keyword)
```
flip1995, line 178: This could be 

```suggestion
    Ident::with_dummy_span(Symbol::intern(maybe_keyword)).is_used_keyword()
```

flip1995 (changes requested):
In addition to @giraffate question, my question is why this is a fix for this lint? The lint documentation states:

> It's hard to distinguish between names that differ only by a single character.

`types` and `type_` only differ by a single character. As do `types` and `type`. So the lint triggering here is correct. Since this lint is `pedantic` and triggering here is correct by the lint definition, I wouldn't change this. I don't see a linked issue about this, was there discussions of this change anywhere?

If you want a keyword as a variable name, you should add `r#` in front of the keyword instead of adding a `_`. So `types` and `r#type` should be fine. Does this lint trigger on those?

th1000s:
In the case of `books` /  `book_` the name `book` is available: One trailing character is not counted, maybe to allow for plurals such as tihis one.

The mentioned issue #6479 notes that an underscore is nicer to read that `r#type` which may look a bit out of place. I also agree that an underscore is visually distinct enough and would even suggest to not count trailing underscores at all, keyword or not.

clippy_lints/src/non_expressive_names.rs
th1000s: Thanks, much better than hardcoding the list. To also catch `yield` etc. I'd even use `is_reserved()`.

flip1995:
`r#<keyword>` exists in Rust, so that you can use keywords as idents. So 

> an underscore is nicer to read that `r#type`

But `r#type` is the Rust way to do it, so this isn't a valid argument for a change in Clippy. Also I disagree personally.

>  I also agree that an underscore is visually distinct enough and would even suggest to not count trailing underscores at all, keyword or not.

I'm not convinced that we should encourage adding an `_` as the only distinguishing thing between variable names.

---

That being said, the suggestion with a trailing underscore is quite bad. I would just remove the help message if the distinguishing character is `_`.

clippy_lints/src/non_expressive_names.rs
```diff
@@ -205,6 +215,13 @@ impl<'a, 'tcx, 'b> SimilarNamesNameVisitor<'a, 'tcx, 'b> {
             // these bindings are typically unused or represent an ignored portion of a destructuring pattern
             return;
         }
+        if interned_name
+            .strip_suffix('_')
```
riking, line 219: Consider also stripping prefix underscores

th1000s:
> suggest[ing] a trailing underscore is quite bad.

Indeed, that just technically fixes the lint. Will fix that in any case.

What do @phansch and @camsteffen think? Should an exception be made when evading known keywords as the issue suggests? Maybe even go further and never count an underscore-character match, as `abc` and `a_c` are distinct enough?

camsteffen:
The keyword thing doesn't make sense to me. The fact that `type` is a keyword doesn't make `type_` and `types` any less similar.

> I would just remove the help message if the distinguishing character is `_`.

> ...never count an underscore-character match, as `abc` and `a_c` are distinct enough?

These statements are saying the same thing. I think this makes sense. Then I would add another caveat that `foo_` and `foo__` should still lint.

flip1995:
> > I would just remove the help message if the distinguishing character is `_`.
> 
> > ...never count an underscore-character match, as `abc` and `a_c` are distinct enough?
> 
> These statements are saying the same thing. I think this makes sense.

Not really, I think? I'm saying, that we should remove (or improve) the help message telling the user to change the name to a double underscore. I'm not suggesting to stop linting in those cases.

camsteffen:
Oh I misunderstood. That makes sense too.

I think we could stop linting those cases since one underscore vs no underscores is a more visually distinct than different letters IMO.

flip1995:
It's still pretty similar though. And this lint is a `pedantic` lint. So I wouldn't change it and let the lint be pedantic.

I would like to see a test if `r#type` vs `types` gets linted by this lint.

giraffate:
ping from triage @th1000s. Does it still need to be discussed?

th1000s:
Ok, then I'll just reword the current suggestion to "use a different name". Just adding a trailing `_` "fixes" the lint but not the intent behind it.

giraffate:
ping from triage @th1000s. Can you have any update on this?

th1000s:
The help text now always reads:

```help: use a more distinct name for either `bpple`, `apple`, or both```

camsteffen:
I don't find that very helpful. The two variables are already pointed out. How bout we just remove the help message?

th1000s:
Very well, done.

giraffate (approved):
I think it looks good.

giraffate:
@bors r+

Thanks!

PR #7225, opened 2021-05-15
New lint: `unused_async`

### LINKED ISSUE #7176: New lint: unused-async
### What it does

Checks for functions that are declared `async` but have no `.await`s inside of them.

### Categories (optional)

- Kind: pedantic

Reduces code smell and overhead by encouraging async functions to be refactored to synchronous functions.

### Drawbacks
There are valid cases where a programmer may want to wrap synchronous code in an asynchronous future: for example, spawning a synchronous task on an asynchronous scheduler, or implementing an async function for a trait where the implementation is synchronous. These cases are the reason this lint is pedantic. However, there are a few additional steps I want to take with this lint to avoid causing hassle or confusion for programmers:
- This lint should avoid trait functions, as there are valid reasons why an async trait function might have a non-async implementation.
- This lint should recommend to the programmer the proper way to make an synchronous task into a future: using the `async {}` block.

### Example

```rust
async fn get_random_number() -> i64 {
    4 // Chosen by fair dice roll. Guaranteed to be random.
}
```

Could be written as:

```rust
fn get_random_number() -> i64 {
    4 // Chosen by fair dice roll. Guaranteed to be random.
}
```

Something like this, however, should **not** be caught by clippy:
```rust
#[async_trait]
trait AsyncTrait {
    async fn foo();
}

struct Bar;

#[async_trait]
impl AsyncTrait for Bar {
    async fn foo() {
        println!("bar");
    }
}
```

InquisitivePenguin:
changelog: Adds a lint, `unused_async`, which checks for async functions with no await statements

`unused_async` is a lint that reduces code smell and overhead by encouraging async functions to be refactored into synchronous functions.

Fixes #7176 

### Examples

```rust
async fn get_random_number() -> i64 {
    4 // Chosen by fair dice roll. Guaranteed to be random.
}
```

Could be written as:

```rust
fn get_random_number() -> i64 {
    4 // Chosen by fair dice roll. Guaranteed to be random.
}
```

Something like this, however, should **not** be caught by clippy:
```rust
#[async_trait]
trait AsyncTrait {
    async fn foo();
}

struct Bar;

#[async_trait]
impl AsyncTrait for Bar {
    async fn foo() {
        println!("bar");
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2538,6 +2538,7 @@ Released 2018-09-13
 [`unsound_collection_transmute`]: https://rust-lang.github.io/rust-clippy/master/index.html#unsound_collection_transmute
 [`unstable_as_mut_slice`]: https://rust-lang.github.io/rust-clippy/master/index.html#unstable_as_mut_slice
 [`unstable_as_slice`]: https://rust-lang.github.io/rust-clippy/master/index.html#unstable_as_slice
+[`unused_async`]: https://rust-lang.github.io/rust-clippy/master/index.html#unused_async
 [`unused_collect`]: https://rust-lang.github.io/rust-clippy/master/index.html#unused_collect
 [`unused_io_amount`]: https://rust-lang.github.io/rust-clippy/master/index.html#unused_io_amount
 [`unused_self`]: https://rust-lang.github.io/rust-clippy/master/index.html#unused_self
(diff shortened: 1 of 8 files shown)


llogiq:
Thank you for this lint. I think there might be other forms of false positives besides being a trait functions. One could have a pointer to an async function and the type of that pointer might change if the function is no longer async. But as this is a pedantic lint, that's no need for concern.

@bors r+

PR #7234, opened 2021-05-17
Remove dead code after #7216

camsteffen:
changelog: none


flip1995:
@bors r+

Thanks!

flip1995:
That was the 2nd time that this happened today. This is a rustup 1.24.1 issue, which is already fixed in 1.24.2, but the windows runner still has 1.24.1. I wouldn't take any action here, since this should fix itself in the near future.

@bors retry

Issue #7231, opened 2021-05-17
Panic in `clippy_lints/src/implicit_return.rs:150:66`


phil-opp:
Link to CI build where the issue occured: https://github.com/phil-opp/blog_os/runs/2600559391?check_suite_focus=true#step:4:226

Panic output with backtrace:

```
thread 'rustc' panicked at 'called `Option::unwrap()` on a `None` value', src/tools/clippy/clippy_lints/src/implicit_return.rs:150:66
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe72845f7bb6a77b9e671e6a4f32fe714962cec4/library/std/src/panicking.rs:515:5
   1: core::panicking::panic_fmt
             at /rustc/fe72845f7bb6a77b9e671e6a4f32fe714962cec4/library/core/src/panicking.rs:92:14
   2: core::panicking::panic
             at /rustc/fe72845f7bb6a77b9e671e6a4f32fe714962cec4/library/core/src/panicking.rs:50:5
   3: <clippy_utils::visitors::visit_break_exprs::V<F> as rustc_hir::intravisit::Visitor>::visit_expr
   4: rustc_hir::intravisit::walk_expr
   5: clippy_lints::implicit_return::lint_implicit_returns
   6: clippy_lints::implicit_return::lint_implicit_returns
   7: <rustc_lint::late::LateLintPassObjects as rustc_lint::passes::LateLintPass>::check_fn
   8: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_fn
   9: rustc_hir::intravisit::walk_item
  10: rustc_hir::intravisit::Visitor::visit_nested_item
  11: rustc_hir::intravisit::walk_mod
  12: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_mod
  13: rustc_hir::intravisit::Visitor::visit_nested_item
  14: rustc_hir::intravisit::walk_mod
  15: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_mod
  16: rustc_hir::intravisit::Visitor::visit_nested_item
  17: rustc_hir::intravisit::walk_mod
  18: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_mod
  19: rustc_hir::intravisit::walk_crate
  20: rustc_lint::late::late_lint_pass_crate
[...]

PR #7242, opened 2021-05-18
Fix ICE in `implicit_return`

Jarcho:
fixes: #7231

changelog: Fix ICE in `implicit_return` for async functions


flip1995:
@bors r+

Thanks! (Luckily this hasn't hit beta yet)

PR #7243, opened 2021-05-18
Add new lint: `strlen_on_c_strings`

### LINKED ISSUE #7145: lint on `libc::strlen(some_cstr.as_ptr())`
### What it does

This lint would flag the use of `libc::strlen` on a `CString` or `CStr` value, and suggest calling `to_bytes().len()` instead.

### Categories (optional)

- Kind: `clippy::complexity`

This avoids calling an unsafe `libc` function. Currently, it also avoids calculating the length; in the future, this may become equivalent to calling `strlen` at runtime, but it'll still avoid the need for an unsafe `libc` call.

### Drawbacks

None.

### Example

```rust
libc::strlen(name.as_ptr())
```

Could be written as:

```rust
name.to_bytes().len()
```

mgacek8:
~~This is WIP, linting in case of `CString` has been added, but for `CStr`, its diagnostic item needs to be available for clippy. 
[PR that adds diagnostic item for CStr on rust repo](https://github.com/rust-lang/rust/pull/85439).~~
Ready for the review. Please take a look. 
fixes #7145
changelog: Add new lint: `strlen_on_c_strings`, that lints on `libc::strlen(some_cstring.as_ptr())`

### DIFF
--- a/clippy_lints/src/strlen_on_c_strings.rs
+++ b/clippy_lints/src/strlen_on_c_strings.rs
@@ -1,0 +1,82 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::in_macro;
+use clippy_utils::paths;
+use clippy_utils::source::snippet_with_macro_callsite;
+use clippy_utils::ty::{is_type_diagnostic_item, is_type_ref_to_diagnostic_item};
+use if_chain::if_chain;
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::{sym, Symbol};
+
+declare_clippy_lint! {
+    /// **What it does:** Checks for usage of `libc::strlen` on a `CString` or `CStr` value,
+    /// and suggest calling `as_bytes().len()` or `to_bytes().len()` respectively instead.
+    ///
+    /// **Why is this bad?** This avoids calling an unsafe `libc` function.
+    /// Currently, it also avoids calculating the length.
+    ///
+    /// **Known problems:** None.
+    ///
+    /// **Example:**
+    ///
+    /// ```rust, ignore
+    /// use std::ffi::CString;
+    /// let cstring = CString::new("foo").expect("CString::new failed");
+    /// let len = unsafe { libc::strlen(cstring.as_ptr()) };
+    /// ```
+    /// Use instead:
+    /// ```rust, no_run
+    /// use std::ffi::CString;
+    /// let cstring = CString::new("foo").expect("CString::new failed");
+    /// let len = cstring.as_bytes().len();
+    /// ```
+    pub STRLEN_ON_C_STRINGS,
+    complexity,
+    "using `libc::strlen` on a `CString` or `CStr` value, while `as_bytes().len()` or `to_bytes().len()` respectively can be used instead"
+}
+
+declare_lint_pass!(StrlenOnCStrings => [STRLEN_ON_C_STRINGS]);
+
+impl LateLintPass<'tcx> for StrlenOnCStrings {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>) {
+        if in_macro(expr.span) {
+            return;
+        }
+
+        if_chain! {
+            if let hir::ExprKind::Call(func, [recv]) = expr.kind;
+            if let hir::ExprKind::Path(hir::QPath::Resolved(_, path)) = func.kind;
+
+            if (&paths::LIBC_STRLEN).iter().map(|x| Symbol::intern(x)).eq(
+                path.segments.iter().map(|seg| seg.ident.name));
+            if let hir::ExprKind::MethodCall(path, _, args, _) = recv.kind;
+            if args.len() == 1;
+            if !args.iter().any(|e| e.span.from_expansion());
+            if path.ident.name == sym::as_ptr;
+            then {
+                let cstring = &args[0];
+                let ty = cx.typeck_results().expr_ty(cstring);
+                let val_name = snippet_with_macro_callsite(cx, cstring.span, "..");
+                let sugg = if is_type_diagnostic_item(cx, ty, sym::cstring_type){
+                    format!("{}.as_bytes().len()", val_name)
+                } else if is_type_ref_to_diagnostic_item(cx, ty, sym::CStr){
+                    format!("{}.to_bytes().len()", val_name)
+                } else {
+                    return;
+                };
+
+                span_lint_and_sugg(
+                    cx,
+                    STRLEN_ON_C_STRINGS,
+                    expr.span,
+                    "using `libc::strlen` on a `CString` or `CStr` value",
+                    "try this (you might also need to get rid of `unsafe` block in some cases):",
+                    sugg,
+                    Applicability::Unspecified // Sometimes unnecessary `unsafe` block
+                );
+            }
+        }
+    }
+}
--- a/tests/ui/strlen_on_c_strings.stderr
+++ b/tests/ui/strlen_on_c_strings.stderr
@@ -1,0 +1,25 @@
+error: using `libc::strlen` on a `CString` or `CStr` value
+  --> $DIR/strlen_on_c_strings.rs:11:24
+   |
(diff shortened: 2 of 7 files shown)


clippy_lints/src/strlen_on_c_strings.rs
```diff
@@ -0,0 +1,85 @@
...
+declare_lint_pass!(StrlenOnCStrings => [STRLEN_ON_C_STRINGS]);
+
+impl LateLintPass<'tcx> for StrlenOnCStrings {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>) {
+        if in_macro(expr.span) {
+            return;
+        }
+
+        if_chain! {
+            if let hir::ExprKind::Call(func, args) = expr.kind;
```
camsteffen, line 47: drive-by nit
```suggestion
            if let hir::ExprKind::Call(func, [recv]) = expr.kind;
```

giraffate:
ping from triage @mgacek8. https://github.com/rust-lang/rust/pull/85439 has been merged, so is this ready for review?

mgacek8:
> 
> 
> ping from triage @mgacek8. [rust-lang/rust#85439](https://github.com/rust-lang/rust/pull/85439) has been merged, so is this ready for review?

Not yet, I resolved conflicts, but I need to investigate why `is_type_diagnostic_item(cx, ty, sym::CStr)` doesn't return true when type is `CStr` and fix it.

mgacek8:
It is ready for the review, please take a look. Can we change the label ?

clippy_lints/src/strlen_on_c_strings.rs
```diff
@@ -0,0 +1,82 @@
...
+                let val_name = snippet_with_macro_callsite(cx, cstring.span, "..");
+                let sugg = if is_type_diagnostic_item(cx, ty, sym::cstring_type){
+                    format!("{}.as_bytes().len()", val_name)
+                }
+                else if is_type_ref_to_diagnostic_item(cx, ty, sym::CStr){
+                    format!("{}.to_bytes().len()", val_name)
+                }
+                else {
+                    return;
+                };
```
giraffate, line 68: Doesn't rustfmt work properly?
```suggestion
                let sugg = if is_type_diagnostic_item(cx, ty, sym::cstring_type){
                    format!("{}.as_bytes().len()", val_name)
                } else if is_type_ref_to_diagnostic_item(cx, ty, sym::CStr){
                    format!("{}.to_bytes().len()", val_name)
                } else {
                    return;
                };
```

clippy_lints/src/strlen_on_c_strings.rs
```diff
@@ -0,0 +1,82 @@
...
+        if in_macro(expr.span) {
+            return;
+        }
+
+        if_chain! {
+            if let hir::ExprKind::Call(func, [recv]) = expr.kind;
+            if let hir::ExprKind::Path(hir::QPath::Resolved(_, path)) = func.kind;
+
+            if (&paths::LIBC_STRLEN).iter().map(|x| Symbol::intern(x)).eq(
+                path.segments.iter().map(|seg| seg.ident.name));
```
giraffate, line 51: Can `match_def_path` util method be used here?

clippy_lints/src/strlen_on_c_strings.rs
```diff
@@ -0,0 +1,82 @@
...
+use clippy_utils::ty::{is_type_diagnostic_item, is_type_ref_to_diagnostic_item};
+use if_chain::if_chain;
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::{sym, Symbol};
+
+declare_clippy_lint! {
+    /// **What it does:** Checks for usage of `libc::strlen` on a `CString` or `CStr` value, and suggest calling `as_bytes().len()` or `to_bytes().len()` respectively instead.
```
giraffate, line 14: nits
```suggestion
    /// **What it does:** Checks for usage of `libc::strlen` on a `CString` or `CStr` value,
    /// and suggest calling `as_bytes().len()` or `to_bytes().len()` respectively instead.
```

clippy_lints/src/strlen_on_c_strings.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    /// **Why is this bad?** This avoids calling an unsafe `libc` function. Currently, it also avoids calculating the length.
```
giraffate, line 16: nits
```suggestion
    /// **Why is this bad?** This avoids calling an unsafe `libc` function.
    /// Currently, it also avoids calculating the length.
```

giraffate (commented):

clippy_lints/src/strlen_on_c_strings.rs
mgacek8: There was a problem that `let path = cx.get_def_path(did);` returns the following path (on Windows):
```
 [clippy_utils\src\lib.rs:1265] &path = [
     "libc",
     "windows",
     "",
     "strlen",
 ]
```

clippy_lints/src/strlen_on_c_strings.rs
mgacek8: Seems like no inside of a macro

clippy_lints/src/strlen_on_c_strings.rs
giraffate: Oh, you're right.

tests/ui/strlen_on_c_strings.stderr
```diff
@@ -0,0 +1,16 @@
+error: using `libc::strlen` on a `CString` or `CStr` value
+  --> $DIR/strlen_on_c_strings.rs:11:24
+   |
+LL |     let len = unsafe { libc::strlen(cstring.as_ptr()) };
+   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `cstring.as_bytes().len()`
```
giraffate, line 5: The suggested code don't need unsafe block. Can unsafe block be checked? Or this comments needs to be improved to include removing unsafe block.

mgacek8: Good catch, yes, unsafe block is redundant here. It can be removed when it's the only one unsafe operation inside of it.
What about for now to lint like this: "help: try this: `cstring.as_bytes().len()`, you might also need to get rid of `unsafe` block in some cases"?

giraffate: Yes, it looks good as a first step! That unsafe block might be removed is one of purposes of this lint.

If unsafe block has only `libc::strlen(cstr.as_ptr())`, it would be better to remove unsafe block in suggestion. But I think it can be fixed with another follow-up PR.

tests/ui/strlen_on_c_strings.stderr
```diff
@@ -0,0 +1,16 @@
+error: using `libc::strlen` on a `CString` or `CStr` value
+  --> $DIR/strlen_on_c_strings.rs:11:24
+   |
+LL |     let len = unsafe { libc::strlen(cstring.as_ptr()) };
+   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `cstring.as_bytes().len(), you might also need to get rid of `unsafe` block in some cases`
```
giraffate, line 5: Can this comment be fixed like this?
```suggestion
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `cstring.as_bytes().len()`, you might also need to get rid of `unsafe` block in some cases
```

mgacek8: Seems like `span_lint_and_sugg` doesn't allow that. 
What about: "help: try this (you might also need to get rid of `unsafe` block in some cases): `cstring.as_bytes().len()`" ?

giraffate: Ok, it looks good!

giraffate (approved):

giraffate:
@bors r+

Thanks!

tests/ui/strlen_on_c_strings.stderr
mgacek8: Created the issue for that: https://github.com/rust-lang/rust-clippy/issues/7436

PR #7253, opened 2021-05-20
Early return from LintPass registration when collecting metadata

flip1995:
This speeds up the metadata collection by 2-2.5x on my machine. During
metadata collection other lint passes don't have to be registered, only
the lints themselves.

cc #7172 

r? @xFrednet 

changelog: none

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -995,291 +995,6 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     ]);
     // end register lints, do not remove this comment, it’s used in `update_lints`
 
-    // all the internal lints
-    #[cfg(feature = "internal-lints")]
-    {
-        store.register_early_pass(|| box utils::internal_lints::ClippyLintsInternal);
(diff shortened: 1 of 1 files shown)


clippy_lints/src/lib.rs
```diff
@@ -2048,32 +1763,320 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
...
+        LintId::of(strings::STRING_LIT_AS_BYTES),
+        LintId::of(transmute::USELESS_TRANSMUTE),
+        LintId::of(use_self::USE_SELF),
+    ]);
+
+    #[cfg(feature = "metadata-collector-lint")]
+    {
+        if std::env::var("ENABLE_METADATA_COLLECTION").eq(&Ok("1".to_string())) {
+            store.register_late_pass(|| box utils::internal_lints::metadata_collector::MetadataCollector::new());
+            return;
```
flip1995, line 1797: This line is the only actual change in this PR.

Everything else is just moving the `register_*_pass` calls below the group registration.

xFrednet:
LGTM! This change sped up the collection on my machine about 2x as well.

@bors r+

PR #7255, opened 2021-05-20
Allow wparam and lparam in similar_names

whatisaphone:
`wparam` and `lparam` are often used as generic parameter names in win32 (for example [WindowProc](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms633573(v=vs.85))). This PR adds them to the similar_names exception list.

changelog: [`similar_names`] don't treat wparam and lparam as similar

### DIFF
--- a/clippy_lints/src/non_expressive_names.rs
+++ b/clippy_lints/src/non_expressive_names.rs
@@ -125,6 +125,7 @@ const ALLOWED_TO_BE_SIMILAR: &[&[&str]] = &[
     &["args", "arms"],
     &["qpath", "path"],
     &["lit", "lint"],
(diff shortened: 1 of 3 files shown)


giraffate:
Thanks for opening PR!

The code looks good to me. But I'm not familiar to Windows, so I don't know if it's reasonable. I would like to hear the opinions of others, here <https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Allow.20wparam.20and.20lparam.20in.20similar_names>.

giraffate:
@bors r+

It looks good, thanks!

PR #7256, opened 2021-05-20
Adding the ability to invalidate caches to force metadata collection

xFrednet:
This adds the discussed hack to touch `clippy_lints/src/lib.rs` to invalidate cargos cache and force metadata collection. I've decided to use the [`filetime`](https://github.com/alexcrichton/filetime) crate instead of the touch command to make is cross-platform and just cleaner in general. Looking at the maintainers I would say that it's a save crate to use xD.

---

cc: #7172 I know this ID without looking it up now... This is not good

changelog: none

r? @flip1995

### DIFF
--- a/tests/dogfood.rs
+++ b/tests/dogfood.rs
@@ -179,7 +179,38 @@ fn dogfood_subprojects() {
 #[ignore]
 #[cfg(feature = "metadata-collector-lint")]
 fn run_metadata_collection_lint() {
+    use std::fs::File;
+    use std::time::SystemTime;
+
+    // Setup for validation
+    let metadata_output_path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("util/gh-pages/metadata_collection.json");
+    let start_time = SystemTime::now();
+
+    // Run collection as is
     std::env::set_var("ENABLE_METADATA_COLLECTION", "1");
+    run_clippy_for_project("clippy_lints");
+
+    // Check if cargo caching got in the way
+    if let Ok(file) = File::open(metadata_output_path) {
+        if let Ok(metadata) = file.metadata() {
+            if let Ok(last_modification) = metadata.modified() {
+                if last_modification > start_time {
+                    // The output file has been modified. Most likely by a hungry
+                    // metadata collection monster. So We'll return.
+                    return;
+                }
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


tests/dogfood.rs
```diff
@@ -179,8 +179,39 @@ fn dogfood_subprojects() {
...
+
+    // Run collection as is
     std::env::set_var("ENABLE_METADATA_COLLECTION", "1");
     run_clippy_for_project("clippy_lints");
+
+    // Check if cargo caching got in the way
+    if let Ok(file) = File::open(metadata_output_path) {
+        if let Ok(metadata) = file.metadata() {
+            if let Ok(last_modification) = metadata.modified() {
+                if last_modification > start_time {
```
flip1995, line 197: Is this even possible? This only happens if someone modifies the file between 6 lines before and this line.

flip1995 (changes requested):
I'm good with adding ``filetime``, since it's just another crate by Alex.

tests/dogfood.rs
xFrednet: This can happen if you don't remove the `metadata_collection.json` file before running this test. It would also be an option to delete it beforehand and simply check if a new one was created. This just removes the deletion step.

flip1995: I meant the other way around: If you don't delete the file, it exists (step 1 in the timeline). Then you run the metadata collection (step 2 in the timeline). So `start_time > last_modified` in this scenario.  The only way that start time is smaller than the last modification is, if you modify it during the execution of those 6 LoC. I don't think a human is capable of such a precise timing.

xFrednet: Ahh, the six previous lines also include run_clippy_for_project("clippy_lints"); which runs the metadata collector itself. This makes it humanly possible (if no cache is used) but most importantly makes it possible for Clippy to rewrite the file. And this time check is targeted towards manipulation from our monster.

The test basically only touches clippy_lints/src/lib.rs if cargo actually used caching.

flip1995: > Ahh, the six previous lines also include `run_clippy_for_project("clippy_lints");` 

Oh, I missed that. This should be fine then.

flip1995 (approved):

flip1995:
@bors r+

Thanks!

Issue #7249, opened 2021-05-19
`while_let_on_iterator` suggests broken fix when iterator is captured in FnMut


zgotsch:
I have noticed a similar problem to #5844 and #6231, where Clippy suggests a for loop instead of a `while let`, causing an illegal move of the target iterator. In my case, it was because the iterator was borrowed by reference in an `FnMut` closure.

Minimal example:
```rust
fn close_over(mut f: impl FnMut()) {
    f();
    f();
}

fn main() {
    let v = vec![1, 2, 3];
    let mut it = v.into_iter();
    close_over(|| {
        while let Some(x) = it.next() {
        // for x in it {
            println!("{}", x);
        }
    });
}
```
[Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8c2e758fd71a6a3b62bec27977ffa56c)

Clippy suggests changing the `while let Some(x) = it.next()` to `for x in it`, but we cannot move out of the mutable reference to `it`. The correct suggestion is probably `for x in &mut it`, similar to fixes in #6966, but it seems more difficult to detect a `FnMut` without type information, which I'm not sure Clippy has access to. I believe the suggestion `for x in it` is better when the iterator has been moved into the closure.

### Meta

- `cargo clippy -V`: `clippy 0.1.52 (88f19c6d 2021-05-03)`
- `rustc -Vv`:
```
rustc 1.52.0 (88f19c6da 2021-05-03)
binary: rustc
commit-hash: 88f19c6dab716c6281af7602e30f413e809c5974
commit-date: 2021-05-03
host: x86_64-apple-darwin
release: 1.52.0
LLVM version: 12.0.0
```

PR #7262, opened 2021-05-21
fix `while_let_on_iterator` suggestion in a closure

Jarcho:
fixes: #7249

A future improvement would be to check if the closure is being used as `FnOnce`, in which case the original suggestion would be correct.

changelog: Suggest `&mut iter` inside a closure for `while_let_on_iterator`

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -861,14 +861,16 @@ pub fn get_enclosing_block<'tcx>(cx: &LateContext<'tcx>, hir_id: HirId) -> Optio
     })
 }
 
-/// Gets the loop enclosing the given expression, if any.
-pub fn get_enclosing_loop(tcx: TyCtxt<'tcx>, expr: &Expr<'_>) -> Option<&'tcx Expr<'tcx>> {
+/// Gets the loop or closure enclosing the given expression, if any.
+pub fn get_enclosing_loop_or_closure(tcx: TyCtxt<'tcx>, expr: &Expr<'_>) -> Option<&'tcx Expr<'tcx>> {
     let map = tcx.hir();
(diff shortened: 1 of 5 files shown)


clippy_utils/src/lib.rs
```diff
@@ -861,14 +861,16 @@ pub fn get_enclosing_block<'tcx>(cx: &LateContext<'tcx>, hir_id: HirId) -> Optio
     })
 }
 
-/// Gets the loop enclosing the given expression, if any.
-pub fn get_enclosing_loop(tcx: TyCtxt<'tcx>, expr: &Expr<'_>) -> Option<&'tcx Expr<'tcx>> {
+/// Gets the loop or closure enclosing the given expression, if any.
+pub fn get_enclosing_loop_or_closure(tcx: TyCtxt<'tcx>, expr: &Expr<'_>) -> Option<&'tcx Expr<'tcx>> {
```
xFrednet, line 865: It might be better to split this up into two functions to make it more versatile. Something like: `get_enclosing_loop` and `get_enclosing_closure`

The rest of the code and logic LGTM :upside_down_face:

Jarcho: Only the innermost of the two is needed for the lint, so it can't really be two separate functions.

xFrednet: Okay, that makes sense. In that case you can resolve this comment :)

xFrednet:
Hey @flip1995, I've reviewed this PR before and now have gone over it a second time. I think it can be merged as it is. Could you maybe double-check it? :upside_down_face: 

r? @flip1995

xFrednet (approved):

flip1995:
@bors r+

Thanks! And thanks for the review @xFrednet

Issue #7224, opened 2021-05-14
Redundant closure incorrectly triggers on `.map(|()| vec![])`


rbtying:
Lint name: redundant_closure

I tried this code:

```rust
pub fn foo() -> Vec<Vec<()>> {
    let x: Vec<()> = vec![(), (), ()];
    x.into_iter().map(|()| vec![]).collect()
}
```

I expected to see this happen: no lint error

Instead, this happened: `cargo clippy` returns

```
warning: redundant closure
 --> src/lib.rs:3:23
  |
3 |     x.into_iter().map(|()| vec![]).collect()
  |                       ^^^^^^^^^^^ help: replace the closure with `Vec::new`: `std::vec::Vec::new`
  |
  = note: `#[warn(clippy::redundant_closure)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
```

This is invalid, because a closure accepting an empty tuple and a closure accepting no arguments is not equivalent.

If we follow the suggestion and replace the closure with `Vec::new`, as follows:
```
pub fn foo() -> Vec<Vec<()>> {
    let x: Vec<()> = vec![(), (), ()];
    x.into_iter().map(Vec::new).collect()
}
```

then `rustc` will fail with

```
 --> src/lib.rs:3:23
  |
3 |     x.into_iter().map(Vec::new).collect()
  |                       ^^^^^^^^ expected function that takes a single 0-tuple as argument
```

### Meta

- `cargo clippy -V`: clippy 0.1.52 (9bc8c42 2021-05-09)
- `rustc -Vv`:
 ```
rustc 1.52.1 (9bc8c42bb 2021-05-09)
binary: rustc
commit-hash: 9bc8c42bb2f19e745a63f3445f1ac248fb015e53
commit-date: 2021-05-09
host: x86_64-unknown-linux-gnu
release: 1.52.1
LLVM version: 12.0.0
  ```

miDeb:
This does not only happen for empty tuples, but for any number and any type of closure arguments.
Code example:
```rust
fn foo() -> Vec<u8> {
    bar(|x, y| vec![])
}

fn bar<U>(f: impl FnOnce(usize, usize) -> U) -> U {
    f(0, 0)
}
```

gives
```plaintext
  |
2 |     bar(|x, y| vec![])
  |         ^^^^^^^^^^^^^ help: replace the closure with `Vec::new`: `std::vec::Vec::new`
  |
[...]

PR #7263, opened 2021-05-21
Fix `redundant_closure` for `vec![]` macro in a closure with arguments

Jarcho:
fixes: #7224
changelog: Fix `redundant_closure` for `vec![]` macro in a closure with arguments


llogiq:
Thank you! @bors r+

Issue #7260, opened 2021-05-21
Suggest to change `std::iter::repeat(s).take(n).collect::<String>()` to `s.repeat(n)`


r00ster91:
### What it does

It suggests the user to change `std::iter::repeat(string).take(n).collect::<String>()` to `string.repeat(n)`. It should work without the `String` type argument to `collect` too and clippy has to find out based on context whether it does collect to a `String`.
Additionally, when the argument to `std::iter::repeat` is a constant char like `'a'` it should suggest to change that to `"a".repeat(n)`. I said only constant chars because I'm not sure how well it will work if it does that for variable chars as well.

### Categories (optional)

- Kind: `clippy::perf`

```rust
use criterion::{criterion_group, criterion_main, Criterion};

fn repeat1(count: usize) -> String {
    std::iter::repeat("a").take(count).collect()
}

fn repeat2(count: usize) -> String {
    "a".repeat(count)
}


fn criterion_benchmark(c: &mut Criterion) {
    c.bench_function("std::iter::repeat", |b| b.iter(|| repeat1(100)));
    c.bench_function("str::repeat", |b| b.iter(|| repeat2(100)));
}

criterion_group!(benches, criterion_benchmark);
criterion_main!(benches);
```
```
std::iter::repeat       time:   [433.62 ns 434.36 ns 435.09 ns]                              
Found 10 outliers among 100 measurements (10.00%)
  4 (4.00%) low mild
  5 (5.00%) high mild
  1 (1.00%) high severe

str::repeat             time:   [52.540 ns 52.658 ns 52.857 ns]                        
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
```

It's faster and generates a lot less instructions, in the `char` case too.
It's also easier to read.

### Drawbacks

None.

### Example

```rust
std::iter::repeat("hello").take(10).collect::<String>();
std::iter::repeat('x').take(10).collect::<String>();
```

Could be written as:

```rust
"hello".repeat(10);
"x".repeat(10);
```

See also: https://github.com/rust-lang/rust/pull/85538.

mikerite:
https://doc.rust-lang.org/std/primitive.str.html#method.repeat
MSRV 1.16 
There will be code that uses `iter::repeat` because it was written before `str::repeat` was added.

r00ster91:
Then I suppose Clippy should check if the version is >=1.16.0 if that's a thing.

[...]

PR #7265, opened 2021-05-22
Add lint `manual_str_repeat`

Jarcho:
fixes: #7260

There's a similar function for slices. Should this be renamed to include it, or should that be a separate lint? If we are going to have them as one lint a better name will be needed. `manual_repeat` isn't exactly clear as it's replacing a call to `iter::repeat`.

changelog: Add new lint `manual_str_repeat`


mikerite:
I tested this lint against a lot of crates and it worked perfectly.

Some of the type checks could be tightened up through because it can get confused  by custom implementations of `FromIterator`. An example is below. I couldn't find an actual example of this in real code.


```
#![warn(clippy::manual_str_repeat)]

use core::iter;
use core::iter::FromIterator;

#[derive(Clone)]
struct S {}

impl FromIterator<Box<S>> for String {
    fn from_iter<I: IntoIterator<Item = Box<S>>>(_: I) -> String {
        "abc".to_string()
    }
}

fn main() {
    let _: String = iter::repeat(Box::new(S {})).take(3).collect();
    //Bad suggestion:
    //let _ = (&Box::new(S {})).repeat(3);
}
```

Jarcho:
I forgot you could do that with `Box`. Fixed.

clippy_lints/src/methods/manual_str_repeat.rs
```diff
@@ -0,0 +1,95 @@
...
+            (ty.is_str() || is_type_diagnostic_item(cx, ty, sym::string_type)).then(|| RepeatKind::Str)
+        }
+    }
+}
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, method_name: Symbol, args: &[Expr<'_>]) {
+    let ctxt = expr.span.ctxt();
+    if_chain! {
+        if let [take_expr] = args;
+        if method_name.as_str() == "collect";
```
giraffate, line 55:
```suggestion
        if method_name == sym!(collect);
```

clippy_lints/src/methods/manual_str_repeat.rs
```diff
@@ -0,0 +1,95 @@
...
+        if let ExprKind::MethodCall(take_name, _, [repeat_expr, take_arg], _) = take_expr.kind;
+        if take_name.ident.as_str() == "take";
```
giraffate, line 57:
```suggestion
        if take_name.ident.name == sym!(take);
```

clippy_lints/src/methods/manual_str_repeat.rs
```diff
@@ -0,0 +1,95 @@
...
+        if let Some(repeat_kind) = parse_repeat_arg(cx, repeat_arg);
+        if ctxt == take_expr.span.ctxt();
+        if ctxt == repeat_expr.span.ctxt();
+        then {
+            let mut app = Applicability::MachineApplicable;
+            let (val_snip, val_is_mac) = snippet_with_context(cx, repeat_arg.span, ctxt, "..", &mut app);
+            let count_snip = snippet_with_context(cx, take_arg.span, ctxt, "..", &mut app).0;
+
+            let val_str = match repeat_kind {
+                RepeatKind::Str if !val_is_mac && repeat_arg.precedence().order() < PREC_POSTFIX => {
```
giraffate, line 75: This is just a question. What does this do? `repeat_arg.precedence().order() < PREC_POSTFIX`

giraffate (changes requested):
Can you add this lint to the lint documentation in `clippy_lints/src/utils/conf.rs`,  and tests for MSRV?

clippy_lints/src/methods/manual_str_repeat.rs
Jarcho: It's to deal with operator precedence. `repeat(*x).take(y).collect()` would change to `*x.repeat(y)` otherwise, which is dereferencing the newly created string.

clippy_lints/src/methods/manual_str_repeat.rs
Jarcho: This is done both ways in clippy right now, near a 50/50 split. The string comparison is the more efficient version as `sym!` is just a wrapper around `Symbol::intern`.

giraffate: Why is it more efficient? This is just a question. I thought `Symbol::as_str` is slow because docs says so.

https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/symbol/struct.Symbol.html#method.as_str
> This is a slowish operation because it requires locking the symbol interner.

clippy_lints/src/methods/manual_str_repeat.rs
flip1995: AFAICT this is necessary for building the suggestion? Why can't you use `clippy_utils::Sugg` here? It should already implement something like this.

clippy_lints/src/methods/manual_str_repeat.rs
Jarcho: Both `as_str` and `intern` lock the interner. `as_str` is then just indexing into a `Vec`. `intern` involves a lookup in a hash table, which both hashes the string, and compares at least two strings for equality (assuming the string is found, which it most certainly will be in this case).

camsteffen: I slightly prefer `as_str()` since it won't ever intern a string unnecessarily. I considered enforcing this in the internal `UNNECESSARY_SYMBOL_STR` lint, but decided it doesn't matter enough to lint.

giraffate: I got it , thanks!

Jarcho:
What change do you want with the documentation?

giraffate:
> What change do you want with the documentation?

Here: https://github.com/rust-lang/rust-clippy/blob/6eea598645be5489f518f91e4b80a0f04a315fda/clippy_lints/src/utils/conf.rs#L127

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

Issue #6722, opened 2021-02-11
suspicious-operation-groupings false positive


emilk:
Lint name: [suspicious-operation-groupings](https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_operation_groupings)

I tried this code:

```rust
pub fn truncated_cone_volume([r0, r1]: [f32; 2], height: f32) -> f32 {
    use std::f32::consts::PI;
    PI * height * (r0 * r0 + r0 * r1 + r1 * r1) / 3.0
}
```

Latest clippy gives me this:
```
warning: This sequence of operators looks suspiciously like a bug.
 --> src/lib.rs:3:40
  |
3 |     PI * height * (r0 * r0 + r0 * r1 + r1 * r1) / 3.0
  |                                        ^^^^^^^ help: I think you meant: `PI * r1`
  |
  = note: `#[warn(clippy::suspicious_operation_groupings)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_operation_groupings
```

(playground link: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=4dbce29a2cf5bed0efc48214538bea53)

---

Minimal reproduce:

``` rust
pub fn false_positive(r0: f32, r1: f32) -> f32 {
    r0 * r0 + r0 * r1 + r1 * r1
}
```

```
2 |     r0 * r0 + r0 * r1 + r1 * r1
  |                         ^^^^^^^ help: I think you meant: `r0 * r1`
```

---

I don't know how this lint is implemented, but maybe it should only trigger if the number of variables involved is the same as the number of binary expressions?

For instance, this is a _true positive_:

``` rust
pub fn true_positive(r0: f32, r1: f32, r2: f32) -> f32 {
    r0 * r0 + r1 * r1 + r2 * r1
}
```

…although the suggestion is probably not the best:
```
2 |     r0 * r0 + r1 * r1 + r2 * r1
  |               ^^^^^^^ help: I think you meant: `r0 * r1`
```

soniasingla:
I am working on this issue 🎤

warycat:
I have similar issues.
```
error: This sequence of operators looks suspiciously like a bug.
  --> rustgym/src/leetcode/d16/_1620_coordinate_with_maxium_network_quality.rs:15:68
   |
15 |                     if (x - xi) * (x - xi) + (y - yi) * (y - yi) > radius * radius {
[...]

PR #7266, opened 2021-05-24
Downgrade suspicious_op..._groupings to Nursery

mikerite:
This addresses #6722.

changelog: Downgrade [`suspicious_lint_operations`] to Nursery


giraffate (approved):
It looks reasonable to downgrade to nursery.

Issue #7209, opened 2021-05-11
Prefer `vec.append(vec)` over `vec.extend(vec.drain(..))`


seritools:
### What it does

This lint would suggest using the specialized `Vec::append` instead of `vec.extend(other_vec.drain(..))`.

### Categories (optional)

- Kind: `clippy::style`

*What is the advantage of the recommended code over the original code*

- Slightly more succinct code
- Potentially better perf?
- Helps raise awareness of `append` even existing :)

### Drawbacks

None that I could come up with

### Example

```rust
let mut a = vec![0u8; 1024];
let mut b = Vec::new();

b.extend(a.drain(..));
```

Could be written as:

```rust
let mut a = vec![0u8; 1024];
let mut b = Vec::new();

b.append(&mut a);
```

Valentine-Mario:
> Hey, I've noticed that this issue is still open after the implementation was merged in #7270. @Valentine-Mario here is a neat-trick for the next time: you can close issues when a PR is merged by including a `closes #<id>` in the PR description. ([GitHub documentation for more keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)) 
> 
> cc: @flip1995 could you close this issue?

Oh. Was looking for the close issue button. Lol. Thanks

PR #7270, opened 2021-05-25
Vec extend to append

Valentine-Mario:
This PR adds a check to suggest changes of vector from 

```
vec.extend(other_vec.drain(..))
```

could be written as 

```
vec![].append(&mut vec![]);
```

changelog: Add vec_extend_to_append lint
issue: #7209


Valentine-Mario:
> r? @xFrednet This seems to be a good PR to review.

The test keeps failing. Not sure what I'm doing wrong. I'm pretty new to the repo 😪

clippy_lints/src/vec_extend_to_append.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+}
+
+fn is_vec_an_append<'tcx>(_cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>)->bool{
+    if_chain!{
+        if let ExprKind::MethodCall(method_name, _, exprs, _) = expr.kind;
+        if method_name.ident.as_str() == "extend";
+        if let ExprKind::MethodCall(method_name2, _, _exprs2, _) = &exprs[1].kind;
+        if method_name2.ident.as_str() == "drain";
```
xFrednet, line 47: Your idea of this implementation is good, but there are some missing aspects.

1. You don't check for the specific types here. This means that any `a.extend(b.drain(..))` would be linted even if a and b are not of type `std::vec::Vec`
    * You can check the types by first getting the type of `exprs[0]` (the self object) and then using [`is_type_diagnostic_item`](https://github.com/rust-lang/rust-clippy/blob/a248648862cabc05a5fefe320bd9fcaa9a0a8c7b/clippy_utils/src/ty.rs#L234-L237)
    ```rs
    // Draft implementation
    let ty = cx.typeck_results().expr_ty(exprs[0]).peel_refs();
    is_type_diagnostic_item(cx, ty, sym::vec_type)
    ```
2. The implementation also doesn't check the given range inside the drain call. For instance `a.extend(b.drain(1..3))` should be valid as it's not equivalent to `a.append(&mut b)`
3. Now the next thing is probably something you couldn't know. Clippy already as a bunch of lints focusing on methods. They have therefor all been combined in a single module and lint pass. New lints focussing on methods should be injected into the existing context. The [`STRING_EXTEND_CHARS`]() is probably a good example for you as it does something very similar. Here are some links:
    * Injecting into current method lint-pass [-->](https://github.com/rust-lang/rust-clippy/blob/cd241b33cb8379311286f53c4afc6e671ddfded1/clippy_lints/src/methods/mod.rs#L1985)
    * Lint implementation: [-->](https://github.com/rust-lang/rust-clippy/blob/21083875d211c29fcfa4a21fcd66d4601d2b618b/clippy_lints/src/methods/string_extend_chars.rs#L13)
    * You can also notice that that implementation provides a suggestion with applicability. You can ignore that part for now and focus on the other notes.

flip1995:
You can update the tests with `cargo dev bless`. Also see CI:

```
Not all lints defined properly. Please run `cargo dev update_lints` to make sure all lints are defined properly.
```

To get started with working on Clippy, make sure to read our documentation in `doc/`. Especially `adding_lints.md` and `basics.md` should help you out.

xFrednet (changes requested):
Hey @Valentine-Mario, you've come to the right place to ask :). I've added some comments to your code with explanations, how it can be improved. 

> The test keeps failing. Not sure what I'm doing wrong. I'm pretty new to the repo :sleepy:

You have two tests which are failing:
1. **UI tests**: these tests run Clippy on the `.rs` files inside `tests/ui/` directory and then compare the output with the `.stderr` file with the same name. You've correctly written the `tests/ui/vec_extend_to_append.rs` file but the corresponding `tests/ui/vec_extend_to_append.stderr` is missing. Our dev tool can copy it to the correct location with `cargo dev bless`.
2. **Test update_lints**: This test verifies that `cargo dev update_lints` produces no changes. Your merge from master most likely altered something that results in a new output. You can simply rerun `cargo dev update_lints`.
3. **Test fmt**: This test didn't run, but it's likely that it would fail. You can simply fix this by running `cargo dev fmt` (Note: The `dev` is required because we're using a slightly different version than `cargo fmt`)

It's usually good to run a few commands before every commit. This is roughly my procedure:

```sh
# Formatting the code base and tests
cargo dev fmt
# Running Clippy's tests
cargo test
# Updating UI test outputs if anything failed
cargo dev bless
# I usually only run this if `Test update_lints` fails but you can also run it every time as it's quite fast
cargo dev update_lints

# git add -A
# git commit -m "Your awesome commit message"
# git push
```

---

All of this might be a lot to take in, but your effort is appreciated, everyone started out like this. Feel free to ask if you're stuck or something is unclear :upside_down_face:

clippy_lints/src/vec_extend_to_append.rs
Valentine-Mario: > 2\. a.extend(b.drain(1..3))

Thanks for the suggestion. Updated the lint and it seems to be working for the most part. Just one slight thing left. In the case of `a.extend(b.drain(1..3))` how do I get the `drain` func args from the exprs2 to verify that is its `1..3`? Thanks

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,61 @@
...
+    ///
+    /// **Example:**
+    ///
+    /// ```rust
+    /// // example code where clippy issues a warning
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// // example code which does not raise clippy warning
+    /// ```
```
xFrednet, line 25: You've now committed the default lint documentation and description in your latest commit. You can leave this for now, this is just a reminder, to make sure we don't forget :)

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,61 @@
...
+
+fn is_vec_an_append<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) -> bool {
+    if_chain! {
+        if let ExprKind::MethodCall(method_name, _, exprs, _) = expr.kind;
+        if method_name.ident.as_str() == "extend";
+        let ty = cx.typeck_results().expr_ty(&exprs[0]).peel_refs();
+        if is_type_diagnostic_item(cx, ty, sym::vec_type);
+        if let ExprKind::MethodCall(method_name2, _, _exprs2, _) = &exprs[1].kind;
+        if method_name2.ident.as_str() == "drain";
+        then {
```
xFrednet, line 56: Okay, you've now added the type check for the first object. We now need the exact same check for the second struct as well to make sure that it's also a `std::vec::Vec<>`.

This as well as the range check will add new if checks. It is often useful to group these checks and annotate them with a comment, like this:
```rust
// Validate target object 
if let ExprKind::MethodCall(method_name, _, exprs, _) = expr.kind;
if method_name.ident.as_str() == "extend";
let ty = cx.typeck_results().expr_ty(&exprs[0]).peel_refs();
if is_type_diagnostic_item(cx, ty, sym::vec_type);
// Check source object
if let ExprKind::MethodCall(src_method, _, src_args, _) = &exprs[1].kind;
if src_method.ident.as_str() == "drain";
let src_ty = cx.typeck_results().expr_ty(&src_args[0]).peel_refs();
if is_type_diagnostic_item(cx, src_ty, sym::vec_type);
// Check drain range
...
then {
```

---

Now about the range check:

Clippy has a utility function that tries to parse a given expression into range. You can find the function here: [`use clippy_utils::higher::range;`](https://github.com/rust-lang/rust-clippy/blob/35e8be7407198565c434b69c5b9f85c71f156539/clippy_utils/src/higher.rs#L51). You can find a usage example from a different lint [here](https://github.com/rust-lang/rust-clippy/blob/12fce557669a0de230399cf8e6eee4f5307bf87b/clippy_lints/src/methods/iter_next_slice.rs#L31-L32)

xFrednet (commented):
Hey, I've added a comment with guidance for the next steps.

In regard to commits. You've now overridden your previous commits with a force push. This can be useful to clean up at the end but right now complicates the review and destroys links from previous comments. Could you please add the next changes in a new commit and also describe a bit better what you're doing in the message itself :upside_down_face:

clippy_lints/src/vec_extend_to_append.rs
```diff
@@ -0,0 +1,53 @@
...
+    ///
+    /// **Why is this bad?** Slightly more succinct code.
+    ///
+    /// **Known problems:** None.
+    ///
+    /// **Example:**
+    /// ```rust
+    /// vec![].extend(vec![1, 2, 3].drain(..));
+    /// ```
+    pub VEC_EXTEND_TO_APPEND,
```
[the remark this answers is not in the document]
xFrednet: You can resolve this :+1:

clippy_lints/src/append_instead_of_extend.rs
Valentine-Mario: Thanks for the updates. Added the check and test are passing. I used a slightly different method though to check the range `is_type_lang_item(cx, ty, LangItem::RangeFull)` .

xFrednet (changes requested):
Hey @Valentine-Mario, I've reviewed your recent changes and added some suggestions. Don't let them overwhelm you, the implementation is getting better each time.

clippy_lints/src/append_instead_of_extend.rs
xFrednet: The use of `LangItem::RangeFull` is a nice solution and I've learned something new. Feel free to resolve this comment @Valentine-Mario :)

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,69 @@
...
+                    && vector_1
+                    && vector_2
+                {
+                    span_lint_and_help(
+                        cx,
+                        APPEND_INSTEAD_OF_EXTEND,
+                        expr.span,
+                        "use of `vec.extend(other_vec.drain(..))`",
+                        None,
+                        "consider using `vec![].append(&mut vec![]);` instead",
```
xFrednet, line 63: Here you can go one step further by not just using `vec![]` but the actual name of the used objects via rusts `snippet` functions. [Here is an example in practice -->](https://github.com/rust-lang/rust-clippy/blob/21083875d211c29fcfa4a21fcd66d4601d2b618b/clippy_lints/src/methods/string_extend_chars.rs#L36-L41) The applicability is a nice bonus but definitely not required, so you can ignore it if you like :)

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,69 @@
...
+declare_lint_pass!(AppendInsteadOfExtend => [APPEND_INSTEAD_OF_EXTEND]);
+
+impl<'tcx> LateLintPass<'tcx> for AppendInsteadOfExtend {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        //validate target object
+        if let ExprKind::MethodCall(method_name, _, exprs, _) = expr.kind {
+            //check source object
+            if let ExprKind::MethodCall(src_method, _, src_args, _) = &exprs[1].kind {
+                //check drain range
+                let ty = cx.typeck_results().expr_ty(&src_args[1]);
```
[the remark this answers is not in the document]
flip1995: I always prefer to use pattern matching instead of length check + indexing. So in the `if let` above, you could write 

```rust
if let ExprKind::MethodCall(src_method, _, [_, src_arg], _) = &exprs[1].kind
```

and then use `src_arg` directly. Same for the `if let` above that with `exprs` of course.

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,69 @@
...
+}
+
+declare_lint_pass!(AppendInsteadOfExtend => [APPEND_INSTEAD_OF_EXTEND]);
+
+impl<'tcx> LateLintPass<'tcx> for AppendInsteadOfExtend {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        //validate target object
+        if let ExprKind::MethodCall(method_name, _, exprs, _) = expr.kind {
+            //check source object
+            if let ExprKind::MethodCall(src_method, _, src_args, _) = &exprs[1].kind {
```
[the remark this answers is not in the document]
Valentine-Mario: I reverted from using the if_chain because I could not print out items to see their structure. I can always every back. And my test all pass though some other test failed. Wasn't sure why. Thanks for pointing that out.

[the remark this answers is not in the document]
Valentine-Mario: Oh, thanks for this 👍🏾 I'll definitely use it.

clippy_lints/src/append_instead_of_extend.rs
Valentine-Mario: Added this. Thanks for the suggestion

tests/ui/append_instead_of_extend.rs
```diff
@@ -0,0 +1,28 @@
...
+    let mut vec3 = vec![0u8; 1024];
+    let mut vec4: std::vec::Vec<u8> = Vec::new();
+
+    vec4.extend(vec3.drain(..));
+
+    let mut vec5 = vec![0u8; 1024];
+    let mut vec6: std::vec::Vec<u8> = Vec::new();
+
+    vec5.extend(vec6.drain(..4));
+}
```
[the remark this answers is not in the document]
Valentine-Mario: Added this too. Thanks

Valentine-Mario:
> Hey @Valentine-Mario, I've reviewed your recent changes and added some suggestions. Don't let them overwhelm you, the implementation is getting better each time.

Added the updates. Sorry for the delay. It was weekend

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+declare_lint_pass!(AppendInsteadOfExtend => [APPEND_INSTEAD_OF_EXTEND]);
+
+impl<'tcx> LateLintPass<'tcx> for AppendInsteadOfExtend {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if does_function_use_append(cx, expr) {
+            let mut applicability = Applicability::MachineApplicable;
+            if let ExprKind::MethodCall(_, _, exprs, _) = expr.kind {
+                if let ExprKind::MethodCall(_, _, src_args, _) = &exprs[1].kind {
```
xFrednet, line 42: You use these two `if let` expressions only to get the arguments for the `snipped` function. You could also move the lint emission into the if chain where you already extracted and validated these arguments. So to line 78. That would make it more lenient and removed a bit of complexity from the implementation.

Another possibility would be to return an `Option<>` from the check function which returns the two expressions if the lint should be triggered.

clippy_lints/src/append_instead_of_extend.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// a.append(&mut b);
+    /// ```
+    pub APPEND_INSTEAD_OF_EXTEND,
+    perf,
+    "default lint description"
+}
+
+declare_lint_pass!(AppendInsteadOfExtend => [APPEND_INSTEAD_OF_EXTEND]);
+
+impl<'tcx> LateLintPass<'tcx> for AppendInsteadOfExtend {
```
xFrednet, line 37: I've previously that we've combined all lint passes regarding methods into a single lint pass for performance reasons ([here](https://github.com/rust-lang/rust-clippy/pull/7270#discussion_r638725620)). Could you move this implementation into the `methods` module with the single lint pass as well? Here was the previous rough guide:

> 3. Now the next thing is probably something you couldn't know. Clippy already as a bunch of lints focusing on methods. They have therefore all been combined in a single module and lint pass. New lints focussing on methods should be injected into the existing context. The [`STRING_EXTEND_CHARS`]() is probably a good example for you as it does something very similar. Here are some links:
>    * Injecting into current method lint-pass [-->](https://github.com/rust-lang/rust-clippy/blob/cd241b33cb8379311286f53c4afc6e671ddfded1/clippy_lints/src/methods/mod.rs#L1985)
>    * Lint implementation: [-->](https://github.com/rust-lang/rust-clippy)

tests/ui/append_instead_of_extend.rs
```diff
@@ -0,0 +1,32 @@
+#![warn(clippy::append_instead_of_extend)]
```
xFrednet, line 1: You can add the following as the first line in the file to also test the automated application of the suggestion:
```rs
// run-rustfix
```

You only need to run the test afterwards once and bless the results. The rest should be taken care of for you :)

xFrednet (commented):
Hey, thank you for the changes. You've also added the automatic applicability of the suggestion correctly. :+1: 

Here are some last annotations. You can fix the red pipeline by adjusting the lint description. I'll do a final review when you've addressed these comments.

> Added the updates. Sorry for the delay. It was weekend

That's totally okay, you can take your time with these changes :upside_down_face:

clippy_lints/src/append_instead_of_extend.rs
Valentine-Mario: Done moved lint to method

clippy_lints/src/append_instead_of_extend.rs
Valentine-Mario: Used the first approach

giraffate:
Can you do a rebase, not merge? We follow a no-merge commit policy: <https://github.com/rust-lang/rust-clippy/blob/master/doc/basics.md#pr>

clippy_lints/src/methods/mod.rs
```diff
@@ -1031,6 +1032,30 @@ declare_clippy_lint! {
...
+    ///
+    /// // Bad
+    /// a.extend(b.drain(..));
+    ///
+    /// // Good
+    /// a.append(&mut b);
+    /// ```
+    pub APPEND_INSTEAD_OF_EXTEND,
+    perf,
+    "default lint description"
```
giraffate, line 1056: Can you update this?

flip1995 (changes requested):
The rebase gone wrong: You have to force push after a rebase, instead of merging the rebased branch. You should be able to fix this by running:

```bash
$ git fetch upstream # assuming upstream is the remote of this repo
$ git rebase upstream/master
$ git diff origin/vec_extend_to_append  # < This should be empty
$ git push --force-with-lease
```

tests/ui/append_instead_of_extend.rs
```diff
@@ -0,0 +1,33 @@
+// run-rustfix
+#![warn(clippy::append_instead_of_extend)]
+use std::collections::BinaryHeap;
+fn main() {
+    let mut vec1 = vec![0u8; 1024];
+    let mut vec2: std::vec::Vec<u8> = Vec::new();
+
+    vec2.extend(vec1.drain(..));
```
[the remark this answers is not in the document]
Valentine-Mario: Added more test update

tests/ui/append_instead_of_extend.fixed
```diff
@@ -0,0 +1,33 @@
...
+    vec4.append(&mut vec3);
+
+    let mut vec5 = vec![0u8; 1024];
+    let mut vec6: std::vec::Vec<u8> = Vec::new();
+
+    vec5.extend(vec6.drain(..4));
+
+    let mut heap = BinaryHeap::from(vec![1, 3]);
+    let mut heap2 = BinaryHeap::from(vec![]);
+    heap2.extend(heap.drain())
```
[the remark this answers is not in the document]
Valentine-Mario: Added more test update

clippy_lints/src/methods/append_instead_of_extend.rs
```diff
@@ -0,0 +1,43 @@
...
+        if method_name.ident.as_str() == "extend";
+        let ty = cx.typeck_results().expr_ty(&exprs[0]).peel_refs();
+        if is_type_diagnostic_item(cx, ty, sym::vec_type);
+        //check source object
+        if let ExprKind::MethodCall(src_method, _, src_args, _) = &exprs[1].kind;
+        if src_method.ident.as_str() == "drain";
+        let src_ty = cx.typeck_results().expr_ty(&src_args[0]).peel_refs();
+        if is_type_diagnostic_item(cx, src_ty, sym::vec_type);
+        //check drain range
+        let src_ty_range = cx.typeck_results().expr_ty(&src_args[1]).peel_refs();
```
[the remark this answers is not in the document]
Valentine-Mario: done

tests/ui/append_instead_of_extend.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let mut vec3 = vec![0u8; 1024];
+    let mut vec4: std::vec::Vec<u8> = Vec::new();
+
+    vec4.extend(vec3.drain(..));
+
+    let mut vec10 = return_vector();
+    let mut vec11: std::vec::Vec<u8> = Vec::new();
+
+    vec11.extend(vec10.drain(..));
```
flip1995, line 19: Last thing: I meant to write a test like

```suggestion
    vec11.extend(return_vector().drain(..));
```

without binding it beforehand.

flip1995 (approved):
Thanks! Please squash your commit and this should be ready to go.

flip1995:
There are some dogfood errors and you have to run `cargo dev update_lints`

Valentine-Mario:
> There are some dogfood errors and you have to run `cargo dev update_lints`

done

flip1995:
There are still 2 dogfood errors.

clippy_lints/src/methods/append_instead_of_extend.rs
```diff
@@ -0,0 +1,41 @@
...
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind, LangItem};
+use rustc_lint::LateContext;
+use rustc_span::symbol::sym;
+
+use super::APPEND_INSTEAD_OF_EXTEND;
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, recv: &Expr<'_>, arg: &Expr<'_>) {
+    if_chain! {
+        if let ty = cx.typeck_results().expr_ty(recv).peel_refs();
```
flip1995, line 14: That works, but you should put this in front of the `if_chain! {`. So:
```suggestion
    let ty = cx.typeck_results().expr_ty(recv).peel_refs();
    if_chain! {
```

Same for the `if let mut applicability` below, which should be moved into the `then { }` block.

---

Unrelated: I'm surprised Clippy doesn't have a lint for irrefutable bindings in `if let` expressions.

flip1995 (changes requested):

Valentine-Mario:
> There are still 2 dogfood errors.

Done

flip1995 (approved):

flip1995:
@bors r+

Thanks!

0969397110:
https://github.com/rust-lang/rust-clippy/pull/7270#issue-652113459

0969397110:
> > There are still 2 dogfood errors.
> 
> 
> 
> Done

Valentine-Mario:
closes #7209

xFrednet:
It has to be in the PR description before the merge. It's setup that way to allow the maintainers to double-check it beforehand. You can add it to future PRs :)

Issue #4310, opened 2019-07-31
Website: Show which lints are auto-correctable with cargo fix


phansch:
Now that `cargo fix` support is on the way for Clippy, it would be nice to update the lint docs to show which lints support auto-correct.

Since a lint can be emitted in different ways with different applicabilities, we have to decide which lints to include:

- Lints where _all_ emitted diagnostics are `MachineApplicable`
- Lints that emit _at least_ one `MachineApplicable` diagnostic

Apart from that we also have to somehow obtain this data. This is probably not going to be easy since our existing python tools for the website have pretty much no way to find out the applicability of lints.

flip1995:
That would be great. 

Maybe we could write an internal Clippy lint, that does this analysis? For performance reasons I would hide the registration of this lint behind a feature flag though.

phansch:
> Maybe we could write an internal Clippy lint, that does this analysis?

Seems like the best option. And I really want to have this feature.. So I'm going to try and write an internal 'lint'/data collector for this.

xFrednet:
@phansch are you still working on this issue? It sounds interesting, and I would like to work on it, if that's okay for you? 🙃

And if so, have you already crated a plan how to implement it? I have a rough idea how it can be done, but I would be happy about pointers if you have already put some thoughts into it.

flip1995:
> This is probably not going to be easy since our existing python tools for the website have pretty much no way to find out the applicability of lints.

This suggest that it is time to RIIR our documentation generation scripts. 

So I see 2 ways how to do this:
1. Implement an internal lint, that collects the data and writes it to a JSON or similar file, that can then be processed by the python scripts. 
    a. If it is implemented in clippy_lints: We would have to figure out a system on how to suppress/enable the generation of this JSON file, since we don't want to always run this, only if we generate the documentation (`deploy` CI workflow)
    b. If it is implemented in clippy_dev: We would need to use `rustc_private` and add a minimal rustc wrapper to `clippy_dev`
2. RIIR our python documentation scripts and then include the fix data collector there, to avoid having to go through an output file (I think this could only be done with implementing variant 1.b. though)

1.a. would be the quick and dirty solution, 1.b. the not so quick, but still dirty, and 2. the slow and clean way.

How would you go about implementing this? @xFrednet

phansch:
I'm not working on this currently, feel free to pick it up again =)

[...]

Issue #1303, opened 2016-10-27
link from the lint in the docs to the code


oli-obk:
The idea came up in the lint workshop at rustbeltrust

We could simply generate a github search link to the uppercase name of the lint

mcarton:
What?

oli-obk:
In the lint list ( https://github.com/Manishearth/rust-clippy#lints ) we link to the docs. The suggestion is to also link to the source code of the lint.

mcarton:
Most users should not need this, and having two links in that already quite big table would be confusing :confused: 
Also, how do you define what the source code of a lint is?

oli-obk:
I know, but it could be like the src link in rustdocs, just a small thing.

Well the source is the line where the lint name is mentioned ;)

mcarton:
> where the lint name is mentioned

Which one?
If you mean [the one where the lint doc is defined](https://github.com/Manishearth/rust-clippy/blob/502416fa78fdb47e98e9983da5a101de49454b0a/clippy_lints/src/len_zero.rs#L31-L52), I think that's pretty useless, as the doc is on the wiki/the website. Otherwise, it's [not](https://github.com/Manishearth/rust-clippy/blob/502416fa78fdb47e98e9983da5a101de49454b0a/clippy_lints/src/len_zero.rs#L105) [unique](https://github.com/Manishearth/rust-clippy/blob/502416fa78fdb47e98e9983da5a101de49454b0a/clippy_lints/src/len_zero.rs#L139).

oli-obk:
We can link to a github search url on the clippy repo, that'll list all the use sites

Manishearth:
Y'all have a lint workshop? What's it all about? 😀

oli-obk:
I taught ppl how to write clippy lints. But the wireless was down all the time, so it was hard. Also we have a lot of work ahead of us to improve the lint writing experience for new ppl. I've already got a few ideas that I'll explore.

PR #7279, opened 2021-05-26
Adapting the lint list to Clippy's new metadata format

xFrednet:
This is close to the end of a long living project to rewrite the lint metadata collection for Clippy. Progress on this has been tracked in #7172. This PR adds one of the last missing puzzle pieces, the actual display of all the changes that have been done in the background. A preview can be seen here: [Clippy's lint list](https://xfrednet.github.io/rust-clippy/master/index.html)

The styling has been discussed on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Styling.20feedback.20for.20Clippy's.20lint.20list/near/239601067) but is still open to suggestion.

Side note: It has been fun working on the website where we don't have unit tests and everything is just tested by playing around. However, it's good that this chaos is contained into this one part of Clippy. :bug: 

---

Closes: #1303
Closes: #4310

This actually closes fewer things than I thought it would...

changelog: Reworked Clippy's [website](https://rust-lang.github.io/rust-clippy/master/index.html):
changelog: * Added applicability information about lints
changelog: * Added a link to jump to the specific lint implementation
changelog: * Adapted some styling and improved loading time

r? @flip1995

### DIFF
--- a/clippy_lints/src/utils/internal_lints/metadata_collector.rs
+++ b/clippy_lints/src/utils/internal_lints/metadata_collector.rs
@@ -114,6 +114,8 @@ const DEPRECATED_LINT_TYPE: [&str; 3] = ["clippy_lints", "deprecated_lints", "Cl
 
 /// The index of the applicability name of `paths::APPLICABILITY_VALUES`
 const APPLICABILITY_NAME_INDEX: usize = 2;
+/// This applicability will be set for unresolved applicability values.
+const APPLICABILITY_UNRESOLVED_STR: &str = "Unresolved";
 
 declare_clippy_lint! {
     /// **What it does:** Collects metadata about clippy lints for the website.
@@ -192,7 +194,7 @@ impl Drop for MetadataCollector {
         let mut lints = std::mem::take(&mut self.lints).into_sorted_vec();
         lints
             .iter_mut()
-            .for_each(|x| x.applicability = applicability_info.remove(&x.id));
+            .for_each(|x| x.applicability = Some(applicability_info.remove(&x.id).unwrap_or_default()));
 
         // Outputting
         if Path::new(OUTPUT_FILE).exists() {
@@ -208,7 +210,7 @@ struct LintMetadata {
     id: String,
     id_span: SerializableSpan,
     group: String,
-    level: &'static str,
+    level: String,
     docs: String,
     /// This field is only used in the output and will only be
     /// mapped shortly before the actual output.
@@ -221,7 +223,7 @@ impl LintMetadata {
             id,
             id_span,
             group,
-            level,
+            level: level.to_string(),
             docs,
             applicability: None,
         }
@@ -269,14 +271,16 @@ impl Serialize for ApplicabilityInfo {
     where
         S: Serializer,
     {
-        let index = self.applicability.unwrap_or_default();
-
         let mut s = serializer.serialize_struct("ApplicabilityInfo", 2)?;
         s.serialize_field("is_multi_part_suggestion", &self.is_multi_part_suggestion)?;
-        s.serialize_field(
-            "applicability",
-            &paths::APPLICABILITY_VALUES[index][APPLICABILITY_NAME_INDEX],
-        )?;
+        if let Some(index) = self.applicability {
+            s.serialize_field(
+                "applicability",
+                &paths::APPLICABILITY_VALUES[index][APPLICABILITY_NAME_INDEX],
+            )?;
+        } else {
+            s.serialize_field("applicability", APPLICABILITY_UNRESOLVED_STR)?;
+        }
         s.end()
     }
 }
@@ -486,6 +490,13 @@ fn extract_attr_docs_or_lint(cx: &LateContext<'_>, item: &Item<'_>) -> Option<St
 /// ```
 ///
 /// Would result in `Hello world!\n=^.^=\n`
+///
+/// ---
+///
+/// This function may modify the doc comment to ensure that the string can be displayed using a
+/// markdown viewer in Clippy's lint list. The following modifications could be applied:
+/// * Removal of leading space after a new line. (Important to display tables)
+/// * Ensures that code blocks only contain language information
 fn extract_attr_docs(cx: &LateContext<'_>, item: &Item<'_>) -> Option<String> {
     let attrs = cx.tcx.hir().attrs(item.hir_id());
     let mut lines = attrs.iter().filter_map(ast::Attribute::doc_str);
@@ -510,7 +521,12 @@ fn extract_attr_docs(cx: &LateContext<'_>, item: &Item<'_>) -> Option<String> {
                 continue;
             }
         }
-        docs.push_str(line);
+        // This removes the leading space that the macro translation introduces
+        if let Some(stripped_doc) = line.strip_prefix(' ') {
+            docs.push_str(stripped_doc);
+        } else if !line.is_empty() {
+            docs.push_str(line);
+        }
     }
     Some(docs)
 }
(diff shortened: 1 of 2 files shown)


util/gh-pages/index.html
```diff
@@ -395,6 +495,18 @@ <h4 class="list-group-item-heading">
...
+            // works for all doc versions (master, beta, rust-1.52.0) with the exception
+            // of "stable" asa this is neither a tag or branch in our repo.
+            //
+            // The dirty hack is to link against beta in that case as that is the closes
+            // known branch to the current stable version. (The version would need to be
+            // queried) 
+            $scope.docVersion = window.location.pathname.split('/')[2];
+            if ($scope.docVersion === "stable") {
+                $scope.docVersion = "beta";
+            }
```
flip1995, line 509: The current stable release is saved in the ``stable`` symlink to the current stable. Don't know, if you can use that here.

util/gh-pages/index.html
```diff
@@ -1,10 +1,17 @@
...
+
+Otherwise, have a great day =^.^=
+-->
 <html lang="en">
 <head>
     <meta charset="UTF-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
 
-    <title>ALL the Clippy Lints</title>
+    <title>Clippy's lint list</title>
```
flip1995, line 14: Why not just 

```suggestion
    <title>Clippy Lints</title>
```

? Anyway, use title case here. Same below.

xFrednet: I would like a different title than "ALL the Clippy Lints" as it always seems a bit long and weird when used in a sentence :sweat_smile:. "Clippy's Lint List" has a nice ring to it, but I'm not sure if it totally fits in the context and tone of Rust.  So, it's your call "Clippy Lints" is also fine with me :)

util/gh-pages/index.html
xFrednet: I wanted to discuss where I might get the current version from. Getting it from the symlink seems a bit complicated. The easiest solution would be to have a stable branch or maybe set a value during deployment that can easily be accessed in the script. 

The value is later used in this link template which works for tags and branches:
```
https://github.com/rust-lang/rust-clippy/tree/{$scope.docVersion}/clippy_lints
```
Examples:

| Lint list URL                                        | Corresponding code URL                                                 |
| ---------------------------------------------------- | ---------------------------------------------------------------------- |
| https://rust-lang.github.io/rust-clippy/master/      | https://github.com/rust-lang/rust-clippy/tree/master/clippy_lints      |
| https://rust-lang.github.io/rust-clippy/beta/        | https://github.com/rust-lang/rust-clippy/tree/beta/clippy_lints        |
| https://rust-lang.github.io/rust-clippy/rust-1.52.0/ | https://github.com/rust-lang/rust-clippy/tree/rust-1.52.0/clippy_lints |

Do you have a simple idea for this?

flip1995: We could add a `stable` branch. This branch would be just a dummy branch holding the commit of the latest release. Since we almost never do stable backports maintaining this shouldn't be much of an overhead.

util/gh-pages/index.html
flip1995: I would prefer just `Clippy Lints`, since the fact that it is a list is pretty obvious. IMO shorter headings are always better, if it has the same meaning.

util/gh-pages/index.html
flip1995: Leaving this as "unresolved", as a reminder for me to create a stable branch (I have to check that it doesn't break deployment and whatnot).

flip1995: Stable branch created: https://github.com/rust-lang/rust-clippy/tree/stable marking as resolved.

xFrednet:
New to-dos:
* [ ] Add the new lint group once #7350 is merged
* [ ] Update the syntax highlighting code once #7352 is merged

flip1995:
I got another idea for the website (definitely not for this PR though): Add a link to playground for ```` ```rust```` code blocks from the documentation. No idea how hard that would be. @xFrednet do you think this would somehow be possible?

xFrednet:
Hey, I think that should be possible to implement this in some form, rustdoc already has similar functionality. I believe the biggest hurdle would be to adapt all lint documentation to have valid code. Some examples are currently using just a code snipped that would be invalid on its own. (Examples: [(1)](https://rust-lang.github.io/rust-clippy/master/index.html#almost_swapped) [(2)](https://rust-lang.github.io/rust-clippy/master/index.html#bool_comparison) [(3)](https://rust-lang.github.io/rust-clippy/master/index.html#builtin_type_shadow)). The code would also need to include a lint level for the lint like `#[deny(lint_name)]` to make sure that it gets triggered.

What is the main objective of this feature? @flip1995 

I would only use it to view the output of such a lint. However, I think that that use case might be better covered by an implementation that includes the lint output in the documentation itself. Rustc does this by running the example code and replacing the string `{{produces}}` with the actual strerr output (from my understanding). Having the lint output in the documentation can make it a bit overloaded, but the output could easily be hidden inside a foldable section to create something like this:

```rust
// Bad
let _ = "Hello".bytes().nth(3);

// Good
let _ = "Hello".as_bytes().get(3);
```

<details>
<summary>Lint output</summary>

```
warning: called `.byte().nth()` on a `str`
 --> src/main.rs:4:9
  |
4 | let _ = "Hello".bytes().nth(3);
  |         ^^^^^^^^^^^^^^^^^^^^^^ help: try: `"Hello".as_bytes().get(3)`
  |
  = note: `#[warn(clippy::bytes_nth)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#bytes_nth
```

</details>

(Sorry if the last part totally misses the point of your question. This was a thing that I thought about while working on a lint in rustc and my mind made the connection to your question)

flip1995:
All code in the documentation tagged with ```` ```rust ```` can be run, since we already check it with rustdoc tests. We include `# ...` lines for this:

https://github.com/rust-lang/rust-clippy/blob/6ae0835df01f51c75b516b02ab2983576cf6b0b3/clippy_lints/src/swap.rs#L50-L55

Yeah my intention is pretty much to run Clippy on the code and see the output. I don't think this feature will be used very often, so it shouldn't be a priority. 

This is more something that would be nice to have, if you or someone else has an idea on how to add this _non-intrusively_. If it is much work to add it, I really don't think it is worth adding at all.

xFrednet:
Okay, I'll add that as an idea to the tracking issue :upside_down_face:

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -509,7 +520,12 @@ fn extract_attr_docs(cx: &LateContext<'_>, item: &Item<'_>) -> Option<String> {
                 continue;
             }
         }
-        docs.push_str(line);
+        // This removes the leading space that the macro translation introduces
+        if let Some(stripped_doc) = line.strip_prefix(' ') {
+            docs.push_str(stripped_doc);
+        } else if !line.is_empty() {
+            docs.push_str(&line);
+        }
```
xFrednet, line 528: The rebase might make it difficult to see where and how the conflicts were resolved. I added this block to strip the leading space on every filled line and added a doc comment to the method itself, the rest stayed the same. I also removed the JS implementation of the language extraction in a new commit.

flip1995:
I don't understand CSS or JS, so I can't really review those changes, but a quick skim over the changes and looking at the resulting website LGTM.

Approving to do a rollup.

@bors r+ rollup

PR #7281, opened 2021-05-26
Fix missing_docs_in_private_items false negative

### LINKED ISSUE #7247: Use rustc_lint::builtin::has_doc
See https://github.com/rust-lang/rust/pull/85457#discussion_r634861596

camsteffen:
changelog: Fix [`missing_docs_in_private_items`] false negative when the item has any `#[name = "value"]` attribute

Closes #7247 (decided not to use the rustc method since it calls `Session::check_name`, which is for rustc only)

### DIFF
--- a/clippy_lints/src/missing_doc.rs
+++ b/clippy_lints/src/missing_doc.rs
@@ -93,9 +93,9 @@ impl MissingDoc {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


flip1995:
@bors r+

Thanks!

PR #7282, opened 2021-05-27
Fix allow on some statement lints

### LINKED ISSUE #7171: Can't disable `no_effect`
I'm using a `struct` with a custom `DerefMut` implementation that internally updates the object every time it's mutably de-referenced, as a way to check whether it has changed. To force an update, I'm using the following code:

```rust
#[allow(clippy::no_effect)]
&mut *custom_deref;
```

Despite appearances, the second line is not actually a no-op. I can't blame Clippy for not being able to figure that out. However, I'm not able to disable the lint warning, even with the `#[allow(clippy::no_effect)]` line. 

The following code replicates this bug. Note how `&mut *custom_deref;` is not a no-op, and note how the warning still shows despite disabling it.

```rust
struct CustomDeref(i32);

impl std::ops::Deref for CustomDeref {
    type Target = i32;

        println!("Derefenced!");
        &self.0
    }
}

impl std::ops::DerefMut for CustomDeref {
    fn deref_mut(&mut self) -> &mut Self::Target {
        println!("Mutably dereferenced!");
        &mut self.0
    }
}

fn main() {
    let mut custom_deref = CustomDeref(0);

    #[allow(clippy::no_effect)]
    &mut *custom_deref;
}
```

Output after `cargo clippy`:

```text
warning: statement with no effect
  --> src\main.rs:23:5
   |
23 |     &mut *custom_deref;
   |     ^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::no_effect)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#no_effect

warning: 1 warning emitted
```

### LINKED ISSUE #7202: can't allow a single needless_collect
I tried this code:

```rust
fn main() {
    let v = vec![1, 2, 3, 2, 4];
    #[allow(clippy::needless_collect)]
    let v: Vec<i32> = v.into_iter().map(|x| x * 2).collect();
}
```
This code should not warn but it does.
````
warning: unused variable: `v`
 --> src/main.rs:4:9
  |
4 |     let v: Vec<i32> = v.into_iter().map(|x| x * 2).collect();
  |         ^ help: if this is intentional, prefix it with an underscore: `_v`
  |

warning: avoid using `collect()` when not needed
 --> src/main.rs:4:52
  |
4 |     let v: Vec<i32> = v.into_iter().map(|x| x * 2).collect();
  |                       -------------                ^^^^^^^
  |                       |
  |                       the iterator could be used here instead
  |
  = note: `#[warn(clippy::needless_collect)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect
help: use the original Iterator instead of collecting it and then producing a new one
  |
4 |     v.into_iter().map(|x| x * 2).map(|x| x * 2).collect();
  |    --                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: 2 warnings emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.68s
````

A global `#![allow(clippy::needless_collect)]` works though.. :thinking: 

[Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=67d84a3a1892e64fc1d2abc6ffc43101)

### Meta
`clippy 0.1.54 (881c1ac 2021-05-08)`

camsteffen:
changelog: Fix `#[allow(..)]` over statements for [`needless_collect`], [`short_circuit_statement`] and [`unnecessary_operation`]

Fixes #7171
Fixes #7202

### DIFF
--- a/clippy_lints/src/loops/needless_collect.rs
+++ b/clippy_lints/src/loops/needless_collect.rs
@@ -1,5 +1,5 @@
 use super::NEEDLESS_COLLECT;
-use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
+use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_hir_and_then};
 use clippy_utils::source::{snippet, snippet_with_applicability};
 use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_diagnostic_item;
@@ -116,9 +116,10 @@ fn check_needless_collect_indirect_usage<'tcx>(expr: &'tcx Expr<'_>, cx: &LateCo
                     // Suggest replacing iter_call with iter_replacement, and removing stmt
                     let mut span = MultiSpan::from_span(collect_span);
                     span.push_span_label(iter_call.span, "the iterator could be used here instead".into());
-                    span_lint_and_then(
+                    span_lint_hir_and_then(
                         cx,
                         super::NEEDLESS_COLLECT,
+                        init_expr.hir_id,
                         span,
                         NEEDLESS_COLLECT_MSG,
                         |diag| {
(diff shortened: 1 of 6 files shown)


flip1995:
@bors r+

Thanks!

PR #7285, opened 2021-05-27
Move mini-macro to tests/ui/auxilary

camsteffen:
changelog: none

Merges `/mini-macro` into `/tests/ui/auxilary/proc_macro_derive.rs`.

The mini-macro crate is an artifact of the distant past. A lot has changed (#2284) and it doesn't make sense as a top-level crate anymore. Especially since we can use the auxilary folder to accompolish the same thing.


tests/ui/crashes/procedural_macro.rs
```diff
@@ -1,11 +0,0 @@
...
-extern crate clippy_mini_macro_test;
-
-#[deny(warnings)]
-fn main() {
-    let x = Foo;
-    println!("{:?}", x);
-}
-
-#[derive(ClippyMiniMacroTest, Debug)]
-struct Foo;
```
camsteffen, line 11: Deleted this for convenience since /crashes has a separate auxilary folder. The test is redundant and history shows no evidence that it ever produced an ICE (see #3743).

flip1995: I also can't find the reason why this exists. Probably to check that the derive doesn't produce an ICE. But since we use this in other tests too, deleting this test should be fine.

flip1995 (approved):

flip1995:
@bors r+

Thanks! This is long overdue.

Issue #7286, opened 2021-05-27
ICE Mismatched brackets in string in closure `let _ = || "}{";`


colelawrence:
### Code

```rust
// lib.rs of cargo new project
pub fn hmm() {
    let _ = || "}{";
}
```

### Meta

- `cargo clippy -V`: clippy 0.1.54 (ff2c947c 2021-05-25)
- `rustc -Vv`:
```
rustc 1.54.0-nightly (ff2c947c0 2021-05-25)
binary: rustc
commit-hash: ff2c947c00f867b9f012e28ba88cecfbe556f904
commit-date: 2021-05-25
host: x86_64-apple-darwin
release: 1.54.0-nightly
LLVM version: 12.0.1
```

### Error output

`cargo clippy`
```
thread 'rustc' panicked at 'begin <= end (3 <= 1) when slicing `"}{"`', src/tools/clippy/clippy_lints/src/functions/too_many_lines.rs:24:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

<details><summary>Backtrace</summary>
  <p>
  
  ```
RUST_BACKTRACE=1 cargo clippy       
    Checking clippy-ice-breckets v0.1.0 (/Users/cole/story/clippy-ice-breckets)
thread 'rustc' panicked at 'begin <= end (3 <= 1) when slicing `"}{"`', src/tools/clippy/clippy_lints/src/functions/too_many_lines.rs:24:26
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::str::slice_error_fail
   3: clippy_lints::functions::too_many_lines::check_fn
   4: <clippy_lints::functions::Functions as rustc_lint::passes::LateLintPass>::check_fn
   5: <rustc_lint::late::LateLintPassObjects as rustc_lint::passes::LateLintPass>::check_fn
   6: rustc_hir::intravisit::walk_expr
   7: rustc_hir::intravisit::walk_local
   8: rustc_hir::intravisit::walk_stmt
   9: rustc_hir::intravisit::walk_block
  10: rustc_hir::intravisit::walk_expr
  11: rustc_hir::intravisit::walk_body
  12: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_nested_body
  13: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_fn
[...]

Issue #7272, opened 2021-05-25
ICE when using declarative macro being defined by another declarative macro


mwu-tow:
### Code
Reduced code is at https://github.com/mwu-tow/clippy-ice-repro

Run the nightly clippy on this repo to reproduce the issue,

It consists of `lib.rs` and `main.rs`.  Lib defines a macro with a macro (using a trick where first argument expands to `$` sign in the nested macro), `main.rs` uses it.

<details><summary>lib.rs</summary>
  <p>

```rust
pub fn warn<T>(_:T) {}

macro_rules! define_macro {
    ($d:tt $lower:ident $upper:ident) => {
        #[macro_export] macro_rules! $upper  {
            ($d arg:tt) => { $crate:: $lower($d arg) }
        }
    }
}

define_macro!{$ warn  WARNING}
```
  </p>
</details>

<details><summary>main.rs</summary>
  <p>

```rust
use repro::*;

pub fn main() { || WARNING!("Style changed!"); }
```
  </p>
</details>

I haven't been able to reduce this to use only a single cargo target.

This is a regression from stable 1.52.1. 
The `nightly-2021-03-25-x86_64-pc-windows-gnu` is confirmed to work correctly.
The `nightly-2021-04-13-x86_64-pc-windows-gnu` is confirmed to ICE on this repo.

### Meta

- `cargo clippy -V`: clippy 0.1.54 (79e50bf7 2021-05-10)
- `rustc -Vv`:
```
rustc 1.54.0-nightly (79e50bf77 2021-05-10)
binary: rustc
commit-hash: 79e50bf77928f374921a6bcafee3fcff1915f062
commit-date: 2021-05-10
host: x86_64-pc-windows-gnu
release: 1.54.0-nightly
LLVM version: 12.0.1
```

### Error output

```
❯ cargo clippy
    Checking repro v0.1.0 (H:\NBO\clippy-ice-repro)
warning: statement with no effect
 --> src\main.rs:3:17
  |
3 | pub fn main() { || WARNING!("Style changed!"); }
  |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(clippy::no_effect)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#no_effect

thread 'rustc' panicked at 'begin <= end (59 <= 42) when slicing `$crate:: $lower($d arg) }
        }
    }
}

[...]

PR #7287, opened 2021-05-27
Fix ICE in `too_many_lines`

Jarcho:
fixes: #7272
fixes: #7286

#7272 looks like it's caused by a bug in rust-c. The span it's giving for the closure is:

```rust
$crate:: $lower($d arg) }
        }
    }
}
```

The correct span would be `$crate:: $lower($d arg)` without all the closing braces.

#7286 is definitely a clippy bug

changelog: none

### DIFF
--- a/clippy_lints/src/functions/too_many_lines.rs
+++ b/clippy_lints/src/functions/too_many_lines.rs
@@ -4,7 +4,7 @@ use rustc_middle::lint::in_external_macro;
 use rustc_span::Span;
 
 use clippy_utils::diagnostics::span_lint;
-use clippy_utils::source::snippet;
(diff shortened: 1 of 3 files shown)


flip1995 (approved):

flip1995:
r=me with ``cargo dev fmt`` @bors delegate+

Jarcho:
Pushed a different implementation that fixes the issue better.

rustfmt is also broken here, it keeps trying to indent those two lines no matter how many times it's run.

Jarcho:
Should be fixed now. @flip1995 want to do a quick check just to make sure everything looks good?

flip1995:
LGTM, Thanks!

@bors r+

Issue #7245, opened 2021-05-19
Warn on SplitN with a count of 1


mibes404:
### What it does

It warns the user when SplitN is used where n == 1, specifically when used in combination with .next(). When used in this way, splitn is not doing anything and can be omitted altogether. It is more likely that the user meant to only get the first part after splitting, which means n has to be at least 2. 

### Categories (optional)

- Kind: clippy::correctness

*What is the advantage of the recommended code over the original code*

In the most likely case it will point the user to undesired behaviour, in the other case - where the written code is actually the user's intent - it will make the code more concise. 

### Drawbacks

None.

### Example

```rust
fn main() {
   let slice = [10, 40, 30, 20, 60, 50];
   let iter = slice.splitn(1, |num| *num % 3 == 0).next();
   println!("{:?}", iter);
}
```
Output:
```
Some([10, 40, 30, 20, 60, 50])
```

Could be written as:

```rust
fn main() {
   let slice = [10, 40, 30, 20, 60, 50];
   let iter = Some(slice);
   println!("{:?}", iter);
}
```
Output:
```
Some([10, 40, 30, 20, 60, 50])
```

Should (most likely) be written as:

```rust
fn main() {
   let slice = [10, 40, 30, 20, 60, 50];
   let iter = slice.splitn(2, |num| *num % 3 == 0).next();
   println!("{:?}", iter);
}
```
Output:
```
Some([10, 40])
```

camsteffen:
Should lint `splitn(0, ..` too.

hi-rustin:
I would like to try it.

@rustbot claim

PR #7292, opened 2021-05-30
Add lint `suspicious_splitn`

Jarcho:
fixes: #7245
changelog: Add lint `suspicious_splitn`


clippy_lints/src/methods/suspicious_splitn.rs
```diff
@@ -0,0 +1,52 @@
...
+    self_arg: &Expr<'_>,
+    count_arg: &Expr<'_>,
+) {
+    if_chain! {
+        // Ignore empty slice literal
+        if !matches!(self_arg.kind, ExprKind::Array([]));
+        // Ignore empty string literal
+        if !matches!(self_arg.kind, ExprKind::Lit(Spanned { node: LitKind::Str(s, _), .. }) if s.is_empty());
+        if let ExprKind::Lit(count_lit) = &count_arg.kind;
+        if let LitKind::Int(count, _) = count_lit.node;
```
flip1995, line 23: Would it make sense to use ``clippy_utils::consts`` here, so that also constant expressions evaluating to ``0`` or ``1`` get linted?

clippy_lints/src/methods/mod.rs
```diff
@@ -1633,6 +1634,35 @@ declare_clippy_lint! {
     "replace `.iter().count()` with `.len()`"
 }
 
+declare_clippy_lint! {
+    /// **What it does:** Checks for calls to `splitn` and related functions with
```
flip1995, line 1638: Please insert a link to the ``splitn`` method here.

```suggestion
    /// **What it does:** Checks for calls to [`splitn`](https://doc.rust-lang.org/std/primitive.str.html#method.splitn) and related functions with
```

flip1995 (commented):
LGTM overall

clippy_lints/src/methods/suspicious_splitn.rs
Jarcho: I imagine it would be unlikely to occur, but I see no reason not to. It could be a false positive with something cfg dependent, but that could just as likely be a bug.

flip1995: Ok, if you also can't think of a reason against it, let's lint on all constants. This is also the usual behavior of Clippy, so it should be fine here too.

Jarcho: Done now.

flip1995:
@bors r+

Thanks!

Issue #6492, opened 2020-12-21
Add the clippy/rust verison of lint introduced to the website


xFrednet:
It might be useful to include the Rust and Clippy version that a lint was introduces on the [lint list](https://rust-lang.github.io/rust-clippy/master/). This feature is useful for two reasons:
1. It can happen that users pinned the compiler version for a project. (I believe that this is more common in other languages as Rust has a high compatibility grantee). It might than be interesting to know if a lint is already available for the compiler version they are using.
2. The current README links to the lint list of master this already includes lints that are not included in the current stable rust version. Displaying the introduction version could reduce confusion

The versions should in my opinion be a part of the lint data ([lints.json](https://github.com/rust-lang/rust-clippy/blob/gh-pages/master/lints.json)) for the website as loading the version dynamically would take up unnecessary performance. 

@rustbot label +L-website, +C-needs-discussion

PR #7298, opened 2021-05-31
Switch CI to new metadata collection

flip1995:
r? @xFrednet 

Things we have to keep in mind:

- This removes the template files and the scripts used for deployment from the checkout. This was added in #5517. I don't think we ever needed those there. Not sure though.
- ~~As a result, we can't remove the python scripts yet. We have to wait until this hits a stable Clippy release.~~ I'll just break the next stable deploy and do it by hand once.
- This should be merged together with #7279. Me and @xFrednet will coordinate the switch
- ...?

I still have to try out some things:

- [x] Is it worth caching? Yes
- [x] ~~Is it worth to do a release build?~~ Nope
- [x] Does it actually work? With a few changes, yes
- [ ] ...?

changelog: Clippy now uses a lint to generate its documentation :tada:


.github/workflows/deploy.yml
```diff
@@ -39,10 +39,19 @@ jobs:
...
 
-    - name: Use scripts and templates from master branch
+    - name: Use deploy script from master branch
       run: |
         git fetch --no-tags --prune --depth=1 origin master
-        git checkout origin/master -- .github/deploy.sh util/gh-pages/ util/*.py
+        git checkout origin/master -- .github/deploy.sh
+
+    - name: cargo update
+      run: cargo update
```
xFrednet, line 48: I'm not sure if this is actually needed. The `Cargo.lock` file is not part of our repository and also not cached based on the documentation. This will not hurt but might be unnecessary.

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -31,7 +31,7 @@ use clippy_utils::{
 };
 
 /// This is the output file of the lint collector.
-const OUTPUT_FILE: &str = "../util/gh-pages/metadata_collection.json";
+const OUTPUT_FILE: &str = "../util/gh-pages/lints.json";
```
xFrednet, line 34: The monster has grown up so fast :cry:

xFrednet (commented):
The deployment script  has some special handling for tags and the beta deployment (if I understand it correctly). In #7279 we've discussed that a `stable` branch should be added to support the *View Code* link works for all website paths. I believe that should be added to the deployment script as well.

The rest looks straight forward and should hopefully work. You mentioned that you've wanted to document the CI a bit better. Do you want to do that as part of this PR or after everything works?

Thank you for taking over the CI work :)

---

Another question that I had recently that is slightly related: The Clippy CI currently starts on every push even in forks or it at least does for me [xFrednet/rust-clippy/actions](https://github.com/xFrednet/rust-clippy/actions). Is that intended behavior? I'm fine with it eating up my CI time, but other users might not be in favor of it.

.github/workflows/deploy.yml
```diff
@@ -39,10 +39,19 @@ jobs:
       if: github.ref == 'refs/heads/beta'
       run: echo "BETA=true" >> $GITHUB_ENV
 
-    - name: Use scripts and templates from master branch
+    - name: Use deploy script from master branch
       run: |
         git fetch --no-tags --prune --depth=1 origin master
-        git checkout origin/master -- .github/deploy.sh util/gh-pages/ util/*.py
+        git checkout origin/master -- .github/deploy.sh
```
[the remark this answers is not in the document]
xFrednet: This is probably what you meant with:

> This removes the template files and the scripts used for deployment from the checkout. This was added in #5517. I don't think we ever needed those there. Not sure though.

We still need the content of `util/gh-pages/` as it contains the index file and the `versions.py` to export the version overview. 

The other python files are no longer needed but still referenced in the deploy script itself. You probably need to update that script in addition to your changes :)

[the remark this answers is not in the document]
flip1995: The contents of `util/gh-pages` is also on the `beta` branch and on every tagged commit. So checking out the master `index.html` file would even break the deployment for beta and stable, since it assumes that there are information in the `lints.json` file that aren't there yet on beta and stable.

Good point on the `versions.py` script. I don't _think_ we have to check it out, but I'll double check.

The `deploy.sh` script is something I have to check on. It should be written in a way, that it doesn't assume anything of the Clippy repo structure, except the necessary files for deployment. That way, if a bug occurs in the deploy script, it can be fixed on master and this fix will be immediately available on beta and stable. So that the python files are called from the deploy script is kind of a problem. I'll see how to best handle that.

.github/workflows/deploy.yml
```diff
@@ -39,10 +39,19 @@ jobs:
...
+
+    - name: cargo update
+      run: cargo update
+
+    - name: Cache
+      uses: Swatinem/rust-cache@v1
```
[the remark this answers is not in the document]
flip1995: Yes, will do. This is **really** WIP and I haven't tested anything yet. This PR was more for writing down my notes rather than the finished CI changes.

.github/workflows/deploy.yml
flip1995: I think it is necessary. The cache action will use a hash of the `Cargo.lock` file in the cache key. If we wouldn't have the `Cargo.lock` file the deploy script would build Clippy with other deps than the other workflows, that build Clippy from scratch, which might break the deploy build. I rather have more cache misses, because the `lock` file changes slightly than having to deal with breakages because of cached deps.

flip1995:
> In #7279 we've discussed that a `stable` branch should be added to support the _View Code_ link works for all website paths. I believe that should be added to the deployment script as well.

Quite the opposite. The `stable` branch should not be deployed, since it would just duplicate the deployed tag. The stable branch will just be a dummy branch pushed at every release together with the tag. 


>  Do you want to do that as part of this PR or after everything works?

Thanks for reminding / forcing me to do this. I want to do this separately. In this PR I only want the really necessary things, so if something goes wrong, we can just revert one commit and everything is fine again.

> Another question that I had recently that is slightly related: The Clippy CI currently starts on every push even in forks or it at least does for me [xFrednet/rust-clippy/actions](https://github.com/xFrednet/rust-clippy/actions). Is that intended behavior?

Yes, this only applies for the `clippy.yml` workflow. I think `@matthiaskrgr` complained when I removed the automatic builds on forks. It is just one job that runs, so it shouldn't eat that much of CI time. People probably won't even notice.

xFrednet:
> Quite the opposite. The stable branch should not be deployed, since it would just duplicate the deployed tag. The stable branch will just be a dummy branch pushed at every release together with the tag.

Yes, that's what I meant.

> Thanks for reminding / forcing me to do this. I want to do this separately. In this PR I only want the really necessary things, so if something goes wrong, we can just revert one commit and everything is fine again.

Sure, makes sense. I remember reading the documentation one or two times before my first contribution to not ask any stupid questions. And it stayed in my focus since then to help other newcomers as well :upside_down_face: 

> Yes, this only applies for the `clippy.yml` workflow. I think `@matthiaskrgr` complained when I removed the automatic builds on forks. It is just one job that runs, so it shouldn't eat that much of CI time. People probably won't even notice.

Okay, thanks for confirming :+1:

.github/workflows/deploy.yml
xFrednet: The selected caching implementation can also just use the `Cargo.toml` at least according to the documentation. 

> a hash of the Cargo.lock / Cargo.toml files (if present). 
> ~(Source: https://github.com/Swatinem/rust-cache#cache-details)

It might be good to add the `Cargo.locked`to the project, but not necessarily required. If we do, we should also think about how the locked file will be updated in the future. This `cargo update` call doesn't commit the file from my understanding.

.github/workflows/deploy.yml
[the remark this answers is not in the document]
xFrednet: > Good point on the `versions.py` script. I don't think we have to check it out, but I'll double check.

Please check that. I also thought about rewriting the version collection in our dev tool and implementing #6492 along with it. But that's pretty low on my to-do list right now. :upside_down_face:

.github/workflows/deploy.yml
flip1995: Yes, it also uses the `Cargo.toml`. But if you have for example `dep_xyz = "0.6"` in your `Cargo.toml`, everything from `0.6.0` up to `0.6.∞` is a valid version to use. If `0.6.2` is currently cached, `cargo` will just use that instead of downloading a new version. But since the main Clippy workflows  build Clippy from scratch they always use the latest versions of the deps. So it could happen, that Clippy in `clippy_bors.yml` is build with `dep_xyz = "0.6.3"`, but the deploy workflow still uses `0.6.2`, since it is cached. This can lead to build failures. This is avoided by running `cargo update`, since it produces the `Cargo.lock` file. I just noticed, that `cargo generate-lockfile` does this also and will switch to that.

Clippy can't have a `Cargo.lock` file since it is part of the `rust-lang/rust` repo and would cause all kinds of problems if it had one.

.github/workflows/deploy.yml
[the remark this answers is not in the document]
flip1995: I don't think RIIRing the `versions.py` script makes sense :smile:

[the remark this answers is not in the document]
xFrednet: I mainly thought about doing it for adding the lint implementation version to the website (somehow...). But a RIIR is not necessary. I agree with you :upside_down_face:

.github/workflows/deploy.yml
xFrednet: Ahh, I had a misconception about how this part would work. Thank you for clarifying :upside_down_face:

flip1995:
This deploy workflow run won't actually work. For that I have to reopen this PR from a branch directly in this repo.

flip1995:
@xFrednet I think I addressed everything we discussed in your previous review. Before I push it to a branch in this repo for testing, can you take another look, if you still notice things that might need another fix?

.github/workflows/deploy.yml
```diff
@@ -39,10 +40,20 @@ jobs:
       if: github.ref == 'refs/heads/beta'
       run: echo "BETA=true" >> $GITHUB_ENV
 
-    - name: Use scripts and templates from master branch
+    - name: Use deploy script from master branch
       run: |
         git fetch --no-tags --prune --depth=1 origin master
-        git checkout origin/master -- .github/deploy.sh util/gh-pages/ util/*.py
+        git checkout origin/master -- .github/deploy.sh util/versions.py util/gh-pages/versions.html
```
xFrednet, line 46: This checkout is missing the `util/gh-pages/index.html` for the lint list itself. The `index.html` is referenced in `deply.sh:10`

xFrednet (commented):
The rest looks ready for a test IMO :+1:

.github/workflows/deploy.yml
flip1995: This is intentional. If we'd check that out, we'd get weird incompatibilities between the output of the metadata collector output and the index.html file. This file still exists in the beta branch and on tags.

flip1995:
> Is it intended, that the CI is read in this PR?

I guess you mean "red"? Or do you mean that it is run? In the first case not intended, but expected, in the second, yes.

.github/workflows/deploy.yml
xFrednet: Okay, so this is basically intended for the next version until the new metadata collector is also used for beta and tags, right?

xFrednet:
> I guess you mean "red"?

Yes :sweat_smile:

Everything should be covered in that case :)

.github/workflows/deploy.yml
flip1995: Nope, we have to keep it like this, since we can't check out a newer version of the metadata collector anymore. Before we checked out the documentation generation script ``export.py`` **and** the ``index.html`` so it was always in a consistent state.

xFrednet: Is the deploy script working in a clone of the repo?

flip1995: Yes

xFrednet: That makes sense now. Thank you for clarifying. 

I guess @flip1995 r+ as you wanted to merge this into a repo branch. ^^

flip1995:
Alright, I tested everything in #7401. So we should be ready for the move, after I squashed the commits.

I won't get to do this this weekend, but plan to do it next Saturday 2021-07-03

xFrednet:
Awesome!!! I'll review this PR again and also look at the test. Next weekend also works for me, I'll rebase #7279 and create a new PR (shortly before the switch) to update the documentation to use hashtags instead of bold text. Is there anything else you need help with? :upside_down_face:

xFrednet (approved):

flip1995:
> Is there anything else you need help with?

Nope, I think I'm good 👍

flip1995:
>     * As a result, we can't remove the python scripts yet. We have to wait until this hits a stable Clippy release.

This isn't the case anymore, since I decided to break the stable deploy for one cycle and do it by hand instead.

PR #7299, opened 2021-05-31
Add macro_braces lint to check for irregular brace use in certain macros

### LINKED ISSUE #7278: New lint: Check `quote!` macro is always used with curyl braces
### What it does

*What does this lint do?*
### What it does

Check `quote!` is always used with curly braces, this could maybe apply to other macros or other bracing combos?

### Categories

- Nursery

This is mostly a consistency lint although using `()` or `[]` doesn't give you a semicolon in item position, which can be unexpected.

### Drawbacks

None :shrug:

### Example

```rust
quote! ( some tokens )
```

Could be written as:

```rust
quote! { some tokens }
```

DevinR528:
The name is a bit long but this sounds good as `#[allow(unconventional_macro_braces)]` and it seems more clear that we are talking about the macro call not macro definitions, any feedback let me know. Thanks!
fixes #7278

changelog: Add ``[`unconventional_macro_braces`]`` lint that checks for uncommon brace usage with macros.


clippy_lints/src/unconventional_macro_braces.rs
```diff
@@ -0,0 +1,146 @@
...
+    ///
+    /// ```rust
+    /// vec!{1, 2, 3};
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// vec![1, 2, 3];
+    /// ```
+    pub UNCONVENTIONAL_MACRO_BRACES,
+    nursery,
```
jplatte, line 27: To my understanding `nursery` is only for lints with known issues. I think this should be `style` or `pedantic`.

clippy_lints/src/unconventional_macro_braces.rs
```diff
@@ -0,0 +1,146 @@
...
+        braces: ("(", ")"),
+    },
+    MacroMatcher {
+        name: "vec",
+        braces: ("[", "]"),
+    },
+    MacroMatcher {
+        name: "quote",
+        braces: ("{", "}"),
+    },
```
jplatte, line 75: I think all of the above should be clear-cut but `quote!` is also used as `quote!()` a lot. Maybe this lint could be warn-by-default with `quote` removed (which users could then re-add as a config item)?

tests/ui/unconventional_macro_braces.stderr
```diff
@@ -0,0 +1,29 @@
+error: use of irregular braces for `vec!` macro
+  --> $DIR/unconventional_macro_braces.rs:5:13
+   |
+LL |     let x = vec! {1, 2, 3};
+   |             ^^^^^^^^^^^^^^
+   |
+   = note: `-D clippy::unconventional-macro-braces` implied by `-D warnings`
+help: consider writing `vec! [1, 2, 3]`
```
jplatte, line 8: The space between `!` and the opening brace is only standard style for curly braces, would be nice if the suggestion could take that into account.

llogiq:
Agree with @jplatte 's comments. Also I'd consider "nonstandard" instead of "unconventional", it's slightly shorter and consistent with "nonstandard_style".

DevinR528:
Ahh thanks for the "nonstandard" I spent a half-hour trying to figure out a good word :laughing:!

clippy_lints/src/unconventional_macro_braces.rs
DevinR528: I went with `style`.

clippy_lints/src/nonstandard_macro_braces.rs
```diff
@@ -0,0 +1,246 @@
...
+        MacroMatcher {
+            name: $name.to_owned(),
+            braces: ($open.to_owned(), $close.to_owned()),
+        }
+    };
+}
+pub(crate) use macro_matcher;
+
+fn macro_braces(conf: FxHashSet<MacroMatcher>) -> Vec<MacroMatcher> {
+    let mut braces = vec![
```
llogiq, line 149: I think `braces` should be a Set instead of a `Vec`. The way it is now does not allow `conf` to determinisitically override settings.

clippy_lints/src/utils/conf.rs
```diff
@@ -196,6 +196,8 @@ define_Conf! {
     (upper_case_acronyms_aggressive: bool = false),
     /// Lint: _CARGO_COMMON_METADATA. For internal testing only, ignores the current `publish` settings in the Cargo manifest.
     (cargo_ignore_publish: bool = false),
+    /// Lint: NONSTANDARD_MACRO_BRACES. Enforce specific macros always use certain braces.
```
llogiq, line 199: I think this may warrant some additional docs?

tests/ui-toml/nonstandard_macro_braces/conf_nonstandard_macro_braces.rs
```diff
@@ -0,0 +1,12 @@
...
+
+extern crate quote;
+
+use quote::quote;
+
+#[rustfmt::skip]
+fn main() {
+    let x = vec! {1, 2, 3};
+    let y = format!["ugh {} stop being such a good compiler", "hello"];
+    let z = quote!(let x = 1;);
```
llogiq, line 11: I'd like to see some cases that don't trigger the lint to guard against false positives.

llogiq (changes requested):
This looks mostly good, I only have a few comments, otherwise it's ready to merge.

clippy_lints/src/utils/conf.rs
DevinR528: Are you thinking of an example or just more info?
This is what I came up with and I added multi-line to the `Conf` macro decl.

```rust
/// Lint: NONSTANDARD_MACRO_BRACES. Enforce the named macros always use the braces specified.
///
/// A `MacroMatcher` looks like this in toml `{ name = "macro_name", brace = "(" }`. If the macro is always going to be used
/// with a full path it should named as such `crate_name::macro_name`.
```

Is there a link to what gets generated by the `Conf` struct?

llogiq: Not sure about the link, but I like your proposed change.

tests/ui-toml/nonstandard_macro_braces/conf_nonstandard_macro_braces.rs
DevinR528: I added a few that won't trigger the lint to the first commit.

llogiq:
The only thing left is the question whether configuration can change brace style (perhaps someone has their own style guide that differs from the standard).

DevinR528:
Do you mean since `print` is already defined as `(` should we be able to override that {which is the way it currently works)? If so then I think yes users should be able to override the default `MacroMatcher`s.

llogiq:
Good! I like that this is also tested. I think once this is rebased, it's ready to merge.

DevinR528:
Should I squash? Oh also I thought maybe the lint should also `check_ty` in the `EarlyLintPass` macros in type position is a thing right?

jplatte:
Yes, macros in type position are a thing, e.g. [syn::Token](https://docs.rs/syn/1.0.73/syn/macro.Token.html).

llogiq:
I'm not sure if bors does it automatically, but it probably won't hurt to squash.

flip1995:
bors merges as-is. So no auto-squashing.

DevinR528:
Squashed, thanks for all the help everyone!

llogiq:
Thank you! @bors r+

PR #7308, opened 2021-06-02
add Arc to `redundant_allocation`

### LINKED ISSUE #7303: extend redundant_allocation to Arc
Currently redundant_allocation check Rc and Box. And it can be extended to Arc, which is quite similar to Rc.
redundant_allocation should check:
```
Arc<Arc<T>>
Arc<Box<T>>
Arc<&T>
```

If it's ok, I can give a pr.

lengyijun:
fixes #7303
changelog:  add Arc to `redundant_allocation`


clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -60,23 +60,73 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
...
-                    applicability,
-                );
-                true
-            })
+        });
+    }
+
+    if cx.tcx.is_diagnostic_item(sym::Arc, def_id) {
+        if let Some(ty) = is_ty_param_diagnostic_item(cx, qpath, sym::Arc) {
+            let mut applicability = Applicability::MachineApplicable;
```
xFrednet, line 82: This is from what I can tell a copy of the `Rc<T>` implementation above with slight adjustments when it comes to the checked type and lint messages. Could you move the code into its own check function that can be reused for both types?

You can use the `format!()` macro to adjust the message for the specific type. Example for formatting the lint message before emission:

```rust
// assuming that `type_sym` is the `Symbol` of the checked type
span_lint_and_sugg(
    cx,
    REDUNDANT_ALLOCATION,
    hir_ty.span,
    &format!("usage of `{}<Box<T>>`", type_sym), // formatted message
    "try",
    format!(
        "{}<{}>",
        type_sym,
        snippet_with_applicability(cx, inner_span, "..", &mut applicability)
    ),
    applicability,
);
```

xFrednet (changes requested):
Hey lengyijun, welcome and thank you for contributing to Clippy.

The new `Arc<>` type should also be added to the lint documentation in [`clippy_lints/src/types/mod.rs`](https://github.com/rust-lang/rust-clippy/blob/527fb42a324296343f94548a116c51646de16065/clippy_lints/src/types/mod.rs#L177-L199). Maybe you can also make the explanation more general and simply add a bullet point list with all checked types. But that is fully up to you.

Could you also move the change log entry in your PR description to be on the same line as "changelog:" ? Take makes writing it a bit easier in the end.

Edit: For the next time, it would also be good if you pick a better commit message. The "But a function is too long" part seems a bit out of place. It also doesn't link to your GitHub account. I'm not sure if there is any policy for it, but it's always better for tracking changes if the account is linked.

The rest of your changes look good to me :+1: 

---

r? @flip1995 I would like to take over this PR review if that works for you :upside_down_face:.

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -25,19 +25,40 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
...
-                applicability,
-            );
-            true
+    let sym_outer = if cx.tcx.is_diagnostic_item(sym::Rc, def_id) {
+        Some(sym::Rc)
+    } else if cx.tcx.is_diagnostic_item(sym::Arc, def_id) {
+        Some(sym::Arc)
+    } else {
+        None
+    };
```
lengyijun, line 34:
```
    let sym_outer = if cx.tcx.is_diagnostic_item(sym::Rc, def_id) {
        Some("Rc")
    } else if cx.tcx.is_diagnostic_item(sym::Arc, def_id) {
        Some("Arc")
    } else {
        None
    };
```
Maybe better?

xFrednet: That's definitely better IMO. You can also use an early return to avoid the whole `Option<>` handling. Something similar like this:

```
let sym_outer = if cx.tcx.is_diagnostic_item(sym::Rc, def_id) {
    sym::Rc
} else if cx.tcx.is_diagnostic_item(sym::Arc, def_id) {
    sym::Arc
} else {
    return false;
};
```

That reduces nesting and saves one if statement in line 36

```
if let Some(sym_outer) = sym_outer {
    ...
```

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -25,19 +25,40 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
...
+                let inner_span = match get_qpath_generic_tys(qpath_inner).next() {
+                    Some(ty) => ty.span,
+                    None => return false,
+                };
+                let mut applicability = Applicability::MachineApplicable;
+                span_lint_and_sugg(
+                    cx,
+                    REDUNDANT_ALLOCATION,
+                    hir_ty.span,
+                    &format!("usage of `{}<Rc<T>>`", sym_outer),
```
xFrednet, line 51: The inner type of this lint emission can be `Rc` or `Arc` due to line 38 - 39. Could you adjust the message to reflect the correct type?

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -25,19 +25,40 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
...
-            true
+    let sym_outer = if cx.tcx.is_diagnostic_item(sym::Rc, def_id) {
+        Some(sym::Rc)
+    } else if cx.tcx.is_diagnostic_item(sym::Arc, def_id) {
+        Some(sym::Arc)
+    } else {
+        None
+    };
+
+    if let Some(sym_outer) = sym_outer {
```
xFrednet, line 36: I think it would still be better to move the content of this entire if block into a new function for readability. The current function is still quite long, and it's not clear right away that this checks reference counting types.

But that's not totally necessary :)

xFrednet (commented):
Thank you for applying the changes so quickly and relinking your GH account in the first commit :+1:

clippy_lints/src/types/redundant_allocation.rs
xFrednet: You can resolve this comment :)

clippy_lints/src/types/redundant_allocation.rs
lengyijun: I have made this function cleaner in latest commit.

clippy_lints/src/types/redundant_allocation.rs
xFrednet: Your refactoring looks awesome and is very clean. Nicely done, you can resolve this comment :)

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,61 @@ use rustc_span::symbol::sym;
...
+    } else if cx.tcx.is_diagnostic_item(sym::Rc, def_id) {
+        "Rc"
+    } else if cx.tcx.is_diagnostic_item(sym::Arc, def_id) {
+        "Arc"
+    } else {
+        return false;
+    };
+
+    if let Some(span) = utils::match_borrows_parameter(cx, qpath) {
+        let mut applicability = Applicability::MachineApplicable;
```
xFrednet, line 23: Hey @flip1995, I've tested the current implementation of this lint on this *very useful* function [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8e56aa3623759c811660250cab26b243):
```rs
#[allow(dead_code)]
pub fn test_fn(useful_value: &u128) -> Box<&u128> {
    Box::new(useful_value)
}
```
The suggested refactoring looks like this:
```rs
pub fn test_fn(useful_value: &u128) -> &u128 {
    Box::new(useful_value)
}
```
Which is invalid due to a type mismatch. I believe that this is a problem with every lint that suggest changes to types, is there any special rule that we allow applicability on types even if that could cause invalid code? Otherwise, we should change it to `Unspecified` IMO :thinking:. (The previous implementation also used `Applicability::MachineApplicable`)

Sorry, for pinging you so much lately :sweat_smile: 

[`Applicability` documentation](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/enum.Applicability.html)

xFrednet (commented):
The refactoring is fantastic. Thank you, for applying them so quickly. This is just a small change I overlooked last time and a question on something I'm unsure about. :upside_down_face:

clippy_lints/src/types/redundant_allocation.rs
xFrednet: Yes, the changes are really cool, you can resolve this comment. :+1:

clippy_lints/src/types/redundant_allocation.rs
xFrednet: This one is also implemented :+1:

clippy_lints/src/types/redundant_allocation.rs
flip1995: We usually use `MaybeIncorrect` in those cases, even though it isn't meant for that use case. But there isn't anything better that says "This suggestion is probably wrong, since it requires further changes".

> Sorry, for pinging you so much lately

No problem, keep doing it :)

---

Sidenote: Since this lint changes function signatures, it should probably not apply to public items by default. We have a config flag for this.

lengyijun: May I change `MachineApplicable` to `MaybeIncorrect` in `redundant_allocation`?

xFrednet: > Sidenote: Since this lint changes function signatures, it should probably not apply to public items by default. We have a config flag for this.

You're probably referring to [`avoid_breaking_exported_api`](https://github.com/rust-lang/rust-clippy/blob/2464ee92840f8d081bb2f546020581e73ab789fa/clippy_lints/src/utils/conf.rs#L125-L126). Supporting that configuration on all type lints would be a good step, but it would require changes to the type module in general. I think it would be better to do that with other type lints together in a different PR. What do you think? @flip1995 I could create an issue for it. :upside_down_face:

xFrednet: Hey @lengyijun, could you change the applicability to `MaybeIncorrect` as discussed earlier.

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,61 @@ use rustc_span::symbol::sym;
...
+
+    let inner_qpath = match &ty.kind {
+        TyKind::Path(inner_qpath) => inner_qpath,
+        _ => return false,
+    };
+    let inner_span = match get_qpath_generic_tys(inner_qpath).next() {
+        Some(ty) => ty.span,
+        None => return false,
+    };
+    let mut applicability = Applicability::MachineApplicable;
```
xFrednet, line 54: Also, this applicability.

xFrednet (changes requested):
These are the last changes at a quick glance, I'll do a final review when the applicability has been updated. Thank you for applying all the requested changes :upside_down_face:

0969397110 (approved):

clippy_lints/src/types/redundant_allocation.rs
flip1995: It isn't necessary for all type lints. Just for type lints that may suggest changing the signature of public API, for example function signatures. So, I wouldn't add it to all type lints, just think that it might make sense for this lint, just from the example, you gave above.

xFrednet (approved):
This version looks excellent to me :upside_down_face: . @flip1995 could you double-check my review?

I think it would be good if the review commits and implementation commits would be squashed in this case, as there are several commits addressing review changes. However, I would leave that decision up to @flip1995, as I'm not totally sure on this yet :sweat_smile:.

BTW. I also checked with `cargo lintcheck` and it doesn't report any new lint triggers or ICEs, so that is good as well IMO

flip1995 (changes requested):
>  It also doesn't link to your GitHub account

That doesn't really matter

> I think it would be good if the review commits and implementation commits would be squashed

Yes, squashing some commits together would definitely be good.

---

I have 2 more things:

1. This lint still suggests to change public API. So it should only trigger iff the `avoid_breaking_public_api` is set to `false`, like discussed here: https://github.com/rust-lang/rust-clippy/pull/7308#discussion_r648610311
2. We can't really tell what type the programmer actually wanted. For example `Rc<Arc<T>>`: We don't really know if the atomic ref counting provided by `Arc` is important. For `Arc/Rc/Box<&T>` this doesn't matter and suggesting `&T` is fine. But if more complex types like `Arc`, `Rc` and `Box` are involved and put together this matters. Right now, always the outer type is suggested as the replacement. I don't think this is a good suggestion. What I would do is to add 2 suggestions to the lint: One that suggests the inner type and one that suggests the outer type as the replacement. But keep in mind that we can't use `run-rustfix` tests with two suggestions and the test file has to be split up into the simple case `Arc/Rc/Box<&T> -> &T` and the other cases.

(I'm only 90% sure that my concern in 2. is valid. So tell me if I'm wrong here)

xFrednet:
> 1. This lint still suggests to change public API. So it should only trigger iff the avoid_breaking_public_api is set to false, like discussed here: #7308 (comment)

I think that this will require some changes to the type module in general and that other type lints (which also affect the public API, like `vec_box`) are missing the flag as well. I therefore suggested in that [comment](https://github.com/rust-lang/rust-clippy/pull/7308#discussion_r648610311) to do that all in another PR. I would be happy to do so once this is merged and that's the reason why I didn't press further on the matter. Would that be okay for you two?

flip1995:
> I would be happy to do so once this is merged and that's the reason why I didn't press further on the matter. Would that be okay for you two?

Yes, that should also be fine :+1:

lengyijun:
>     2\. We can't really tell what type the programmer actually wanted. For example `Rc<Arc<T>>`: We don't really know if the atomic ref counting provided by `Arc` is important. For `Arc/Rc/Box<&T>` this doesn't matter and suggesting `&T` is fine. But if more complex types like `Arc`, `Rc` and `Box` are involved and put together this matters. Right now, always the outer type is suggested as the replacement. I don't think this is a good suggestion. What I would do is to add 2 suggestions to the lint: One that suggests the inner type and one that suggests the outer type as the replacement. But keep in mind that we can't use `run-rustfix` tests with two suggestions and the test file has to be split up into the simple case `Arc/Rc/Box<&T> -> &T` and the other cases.

Should we auto fix `Box<Box<T>>`, `Rc<Rc<T>>`, `Arc<Arc<T>>` ?

flip1995:
> Should we auto fix `Box<Box<T>>`, `Rc<Rc<T>>`, `Arc<Arc<T>>` ?

Yes, those should be fine.

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,78 @@ use rustc_span::symbol::sym;
...
+            &format!(
+                "you seem to be trying to use `{}<{}<T>>`. Consider using just `{}<T>` of `{}<T>`",
+                outer_sym, inner_sym, outer_sym, inner_sym,
+            ),
+            None,
+            &format!(
+                "`{}<T>` is already on the heap, `{}<{}<T>>` makes an extra allocation",
+                inner_sym, outer_sym, inner_sym,
+            ),
+        );
```
xFrednet, line 83: The lint message already contains a help message. It would be better to add that as a help message. You can use `span_lint_and_then(..)` to add two help messages. I would also change the current "help" message to be a "note" as it's more an explanation.

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,78 @@ use rustc_span::symbol::sym;
...
+            applicability,
+        );
     } else {
-        false
+        span_lint_and_help(
+            cx,
+            REDUNDANT_ALLOCATION,
+            hir_ty.span,
+            &format!(
+                "you seem to be trying to use `{}<{}<T>>`. Consider using just `{}<T>` of `{}<T>`",
```
xFrednet, line 75:
```suggestion
                "you seem to be trying to use `{}<{}<T>>`. Consider using just `{}<T>` or `{}<T>`",
```

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,78 @@ use rustc_span::symbol::sym;
...
+
+    let inner_qpath = match &ty.kind {
+        TyKind::Path(inner_qpath) => inner_qpath,
+        _ => return false,
+    };
+    let inner_span = match get_qpath_generic_tys(inner_qpath).next() {
+        Some(ty) => ty.span,
+        None => return false,
+    };
+    let mut applicability = Applicability::MaybeIncorrect;
```
xFrednet, line 54: NIT: I believe you can move this value inside the if block

tests/ui/redundant_allocation_fixed.fixed
```diff
@@ -0,0 +1,75 @@
+// run-rustfix
```
xFrednet, line 1: I think `..._fixable.rs` would be a better name then just fixed.

clippy_lints/src/types/redundant_allocation.rs
lengyijun: Sorry for making so many mistakes.

xFrednet: It's totally fine, don't worry about it. They are all pretty minor and some, were actually just NITs from the previous implementation. I for my part had fun reviewing this PR and to see it take shape, and I'm also happy that you're still motivated :upside_down_face:.

lengyijun: I' m not sure about note information. If this `span_lint_and_note` is ok?

xFrednet: You could in theory use `span_lint_and_note` but that would only give you the ability to add a single note to a diagnostic. `span_lint_and_then` enables you to add several messages.

In this case, it can look something like this (Haven't tested it locally, but it should work with some slight adjustments).
```rs
span_lint_and_then(
    cx,
    REDUNDANT_ALLOCATION,
    hir_ty.span,
    &format!("use of `{}<{}<T>>`", outer_sym, inner_sym),
    |diag| {
        diag.note(&format!(
            "`{}<T>` is already on the heap, `{}<{}<T>>` makes an extra allocation",
            inner_sym, outer_sym, inner_sym,
        ));
        diag.help(&format!("consider using just `{}<T>` or `{}<T>`", outer_sym, inner_sym));
    },
);
```

That should give you a message like:

```
error: use of `Box<Arc<T>>`.
  --> $DIR/redundant_allocation.rs:33:30
   |
LL |     pub fn box_test9<T>(foo: Box<Arc<T>>) -> Box<Arc<SubT<T>>> {
   |                              ^^^^^^^^^^^
   |
   = note: `Arc<T>` is already on the heap, `Box<Arc<T>>` makes an extra allocation
   = help: consider using just `Box<T>` or `Arc<T>`
```

Other links:
* [Example from inside Clippy](https://github.com/rust-lang/rust-clippy/blob/12fce557669a0de230399cf8e6eee4f5307bf87b/clippy_lints/src/fallible_impl_from.rs#L130-L140)
* [DiagnosticBuilder docs](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/struct.DiagnosticBuilder.html#method.note)

lengyijun: I found the note information is different among `Box<Box<T>>` and `Box<Rc<T>>`.
So, in order to add a note to `span_lint_and_sugg`, I change it into `span_lint_and_then`.

xFrednet (approved):
LGTM

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,95 @@ use rustc_span::symbol::sym;
...
+            &format!("usage of `{}<&T>`", outer_sym),
+            |diag| {
+                diag.span_suggestion(
+                    hir_ty.span,
+                    "try",
+                    format!("{}", snippet_with_applicability(cx, span, "..", &mut applicability)),
+                    applicability,
+                );
+                diag.note(&format!(
+                    "`&T` is already a pointer, `{}<&T>` allocate a pointer on heap",
```
flip1995, line 37:
```suggestion
                    "`&T` is already a pointer, `{}<&T>` allocates a pointer on the heap",
```

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,95 @@ use rustc_span::symbol::sym;
...
+            cx,
+            REDUNDANT_ALLOCATION,
+            hir_ty.span,
+            &format!("usage of `{}<{}<T>>`", outer_sym, inner_sym),
+            |diag| {
+                diag.note(&format!(
+                    "`{}<T>` is already on the heap, `{}<{}<T>>` makes an extra allocation",
+                    inner_sym, outer_sym, inner_sym,
+                ));
+                diag.help(&format!("consider using just `{}<T>` or `{}<T>`", outer_sym, inner_sym,));
```
flip1995, line 98:
```suggestion
                diag.help(&format!("consider using just `{}<T>` or `{}<T>`", outer_sym, inner_sym));
```

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,95 @@ use rustc_span::symbol::sym;
...
+                        snippet_with_applicability(cx, inner_span, "..", &mut applicability)
+                    ),
                     applicability,
                 );
-                true
-            })
-        }
+                diag.note(&format!(
+                    "`{}<T>` is already on the heap, `{}<{}<T>>` makes an extra allocation",
+                    inner_sym, outer_sym, inner_sym,
```
flip1995, line 83: This could be improved by also using the result of `let generic_snippet = snippet_with_applicability(cx, inner_span, "..", &mut applicability)` here instead of just a generic `T`. You can do this nicely with format args:

```suggestion
                    "`{inner}<{generic}>` is already on the heap, `{outer}<{inner}<{generic}>>` makes an extra allocation",
                    inner=inner_sym, outer=outer_sym, generic=generic_snippet
```

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -9,74 +9,95 @@ use rustc_span::symbol::sym;
...
     } else {
-        false
+        span_lint_and_then(
+            cx,
+            REDUNDANT_ALLOCATION,
+            hir_ty.span,
+            &format!("usage of `{}<{}<T>>`", outer_sym, inner_sym),
+            |diag| {
+                diag.note(&format!(
+                    "`{}<T>` is already on the heap, `{}<{}<T>>` makes an extra allocation",
```
flip1995, line 95: Same here: use the snippet of the generic arg instead of a generic type param.

flip1995 (commented):
Impl looks really good now. The error message still could be slightly improved. Remember that you will have to run `cargo dev bless` after applying my suggested changes.

flip1995 (approved):

flip1995:
Thanks, LGTM now! Sorry for the long time without a response, GitHub sometimes doesn't give notifications on force pushes.

Can you squash your commits and ping me afterwards, so we can merge this?

lengyijun:
Do I need to squash all commits to one commit?

flip1995:
Not necessarily. The remaining commits should make sense to keep as separate commits though. E.g. one commit for adding tests, one commit for adding the visitor and one for the actual fix.

lengyijun:
Thx.
I prefer to squash them into single commit to make the git tree clean.

flip1995:
@bors r=xFrednet,flip1995

Thanks!

PR #7315, opened 2021-06-03
Add disallowed_type lint, this adds a field to the conf struct

### LINKED ISSUE #7277: New lint: disallowed_types based on config
### What it does

Check certain types are not used according to a config list (similar to [disallowed_method](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_method)).

### Categories

- Nursery

Allow a project to specify types that are not allowed.

### Drawbacks

None :shrug:

### Example

```rust
```

Could be written as:
possibly with a short explanation that is included in the config somehow?
```rust
// Since non of the maps will be large the cost of hashing is greater than the cost of lookup.
use std::collections::BTreeMap;
```

DevinR528:
Fixes #7277

changelog: Add ``[`disallowed_type`]`` a lint that can enforce banning types specified in the config.


DevinR528:
Looks like we got someone brought ICE to the party! CI failed because of 2 different ICE panics maybe just one?

clippy_lints/src/disallowed_type.rs
```diff
@@ -0,0 +1,257 @@
...
+                if let GenericArg::Type(t) = arg {
+                    walk_to_path(cx, t, disallowed);
+                }
+            }
+        },
+        TyKind::TraitObject(polyref, _, _) => walk_trait_obj(cx, &polyref.iter().collect::<Vec<_>>(), disallowed),
+        TyKind::Typeof(anonconst) => {
+            if_chain! {
+                if let Some(tyck) = cx.maybe_typeck_results();
+                if let ty::Adt(ty::AdtDef { did, .. }, _) = tyck.node_type(anonconst.hir_id).kind();
```
giraffate, line 150: To fix ICE in CI, it would be better to use `node_type_opt`:
```suggestion
                let var_ty = tyck.node_type_opt(anonconst.hir_id);
                if let Some(var_ty) = var_ty;
                if let ty::Adt(ty::AdtDef { did, .. }, _) = var_ty.kind();
```

giraffate (changes requested):

clippy_lints/src/disallowed_type.rs
DevinR528: Oddly enough this lint isn't affected by this, should I work on changing uses throughout the lints. It seems `node_type` is responsible for most/all of the failures?

giraffate (approved):
It looks good, thanks! Can you squash commits?

DevinR528:
Squashed.

giraffate:
@bors r+

Thanks!

PR #7316, opened 2021-06-04
Add new lint: `rc_mutex`

lengyijun:
changelog: Add new lint `rc_mutex`. 

It lints on `Rc<Mutex<T>>`.  

`Rc<Mutex<T>>` should be corrected to `Rc<RefCell<T>>`


llogiq (changes requested):
This should work well enough as a restriction lint. However, I think the suggestion is incomplete and the tests should be fleshed out more.

tests/ui/rc_mutex.rs
```diff
@@ -0,0 +1,28 @@
...
+pub struct SubT<T> {
+    foo: T,
+}
+
+pub enum MyEnum {
+    One,
+    Two,
+}
+
+pub fn test1<T>(foo: Rc<Mutex<T>>) {}
```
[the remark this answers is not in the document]
lengyijun: Could you please check if I added correct test in my latest commit?
I'm not sure I did it right.

tests/ui/rc_mutex.rs
```diff
@@ -0,0 +1,27 @@
...
+
+pub fn test2(foo: Rc<Mutex<MyEnum>>) {}
+
+pub fn test3(foo: Rc<Mutex<SubT<usize>>>) {}
+
+fn main() {}
```
llogiq, line 27: How about calling `test1`, `test2`, etc. from here?

This is something the lint might want to consider, btw., and also the reason not to apply it automatically.

flip1995:
Is there an issue open for this lint? Or was it discussed in any other place beforehand? If so, please link to it in the PR description.

lengyijun:
There is no related issue and discussion.

clippy_lints/src/types/mod.rs
```diff
@@ -250,12 +251,39 @@ declare_clippy_lint! {
     "usage of very complex types that might be better factored into `type` definitions"
 }
 
+declare_clippy_lint! {
+    /// **What it does:** Checks for `Rc<Mutex<T>>`.
+    ///
+    /// **Why is this bad?** `Rc` is used in single thread and `Mutex` is used in multi thread.
+    /// Consider using `Rc<RefCell<T>>` in single thread or `Arc<Mutex<T>>` in multi thread.
+    ///
+    /// **Known problems:** None.
```
llogiq, line 260: I think there may be instances where combining generics and traits could require this actual type combination, and following the suggestion would make it impossible to implement the traits.

We should consider those cases false positives and make a note of a possible problem. I don't think this problem is worrisome for the lint (it's unlikely to happen in practise), but we should note it nonetheless.

llogiq:
Sorry for letting you wait for so long. I had some $life keeping me busy.

I'd like a more fleshed out explanation of the possible false positives. How about: "Sometimes combining generic types can lead to the requirement that a type use `Rc` in conjunction with `Mutex`. We must consider those cases false positives, but alas they are quite hard to rule out. Luckily they are also rare."

llogiq:
Thank you! @bors r+

PR #7319, opened 2021-06-04
Don't warn about `cfg!(..)` as a constant in assertions

m-ou-se:
This makes clippy understand that `cfg!(..)` is not just a hardcoded `true` or `false` (even though it expands to one of those).

cc @khyperia

changelog: Don't treat `cfg!(..)` as a constant in [`assertions-on-constants`]


camsteffen:
Thanks!

@bors r+

PR #7321, opened 2021-06-04
Release doc: add explanation for updating the stable branch

flip1995:
r? @xFrednet 

[Rendered](https://github.com/flip1995/rust-clippy/blob/doc-release-stable-branch/doc/release.md)

changelog: none

### DIFF
--- a/doc/release.md
+++ b/doc/release.md
@@ -101,6 +101,21 @@ After this, the release should be available on the Clippy [release page].
 
 [release page]: https://github.com/rust-lang/rust-clippy/releases
 
+## Update the `stable` branch
+
+At this step you should have already checked out the commit of the `rust-1.XX.0`
+tag. Updating the stable branch from here is as easy as:
+
+```bash
+# Assuming the current directory corresponds to the Clippy repository and the
+# commit of the just created rust-1.XX.0 tag is checked out.
+$ git push upstream rust-1.XX.0:stable  # `upstream` is the `rust-lang/rust-clippy` remote
+```
+
+_NOTE: Usually there are no stable backports for Clippy, so this update should
+be possible without force pushing or anything like this. If there should have
+happened a stable backport, make sure to re-merge those changes just as with the
+`beta` branch._
 
 ## Update `CHANGELOG.md`
 


xFrednet (commented):
Using a `stable` tag instead of a branch is a really nice solution!

doc/release.md
```diff
@@ -101,6 +101,21 @@ After this, the release should be available on the Clippy [release page].
...
+```bash
+# Assuming the current directory corresponds to the Clippy repository and the
+# commit of the just created rust-1.XX.0 tag is checked out.
+$ git push upstream rust-1.XX.0:stable  # `upstream` is the `rust-lang/rust-clippy` remote
+```
+
+_NOTE: Usually there are no stable backports for Clippy, so this update should
+be possible without force pushing or anything like this. If there should have
+happened a stable backport, make sure to re-merge those changes just as with the
+`beta` branch._
```
xFrednet, line 118: I've tested this procedure in a test repo and I had to force update the stable tag on the second run because it already existed. Could you check if this is also the case for ypi?

My test script:
```sh
# Create a file
$ git add -A
$ git commit -m "Test commit 1"
$ git tag test-tag-1
$ git push origin test-tag-1
$ git push origin test-tag-1:stable # This works the first time

# Delete that file
$ git add -A
$ git commit -m "Test commit 2"
$ git tag test-tag-2
$ git push origin test-tag-2
$ git push origin test-tag-2:stable
   # To github.com:xFrednet/some-hidden-test-repo.git
   #  ! [rejected]        test-2 -> stable (already exists)
   # error: failed to push some refs to 'git@github.com:xFrednet/some-hidden-test-repo.git'
   # hint: Updates were rejected because the tag already exists in the remote.
$ git push origin test-2:stable --force
   # Total 0 (delta 0), reused 0 (delta 0)
   # To github.com:xFrednet/some-hidden-test-repo.git
   #  + 28b0f5a...2e2272f test-2 -> stable (forced update)
```

flip1995:
> Using a `stable` tag instead of a branch is a really nice solution!

It's not a tag, it's a branch. 

> I've tested this procedure in a test repo and I had to force update the stable tag on the second run because it already existed.

The reason, why your test failed is (and I'm assuming here without testing it), that you didn't create the ``stable`` branch first and therefore created/pushed the ``test-tag-1`` twice. Once under the name ``test-tag-1`` and once under the name ``stable``. In other words, your command expanded to ``git push origin refs/tags/test-tag-1:refs/tags/stable``. But if a branch named ``stable`` already exists, than git will expand the command to ``git push origin refs/tags/test-tag-1:refs/heads/stable`` and pushes to the branch, since ``git`` assumes that you meant the branch and not an non-existent tag.

So if you insert a ``git checkout -b stable && git push [-u] origin stable && git checkout master`` after the first ``git commit`` it should work.

xFrednet:
Yeah, you're correct, everything works now. Nice work!

Should we merge this now or with the rest of the new collection? There is no real reason to wait with this one, right?

flip1995:
Can be merged now. I think we won't remove the stable branch.

xFrednet:
@bors r+

In that case we're one step closer, thank you :)

PR #7330, opened 2021-06-08
Refactoring identity function lints

xFrednet:
I've noticed that we have several lints that all check for identity functions and each used their own check implementation. I moved the `is_expr_identity_function` function to `clippy_utils` and adapted all lints to reuse that one function. This should make the addition of new lints like this also easier in the future.

I've also moved the `map_identity` lint into the `methods` module. It's probably the best to review this PR by checking each commit individually. And that's it, have a great day :upside_down_face: 

changelog: none


flip1995:
@bors r+

Thanks!

PR #7333, opened 2021-06-09
Factor out match_var and get_pat_name utils

camsteffen:
...because checking vars by name is bad, because of shadowing.

changelog: none


llogiq:
Thank you, that's a solid improvement.

@bors r+

PR #7338, opened 2021-06-09
Re-write shadow lints

### LINKED ISSUE #318: shadow lint should use visitor or ExprUseVisitor
We're duplicating a lot of the work Visitor does.

ExprUseVisitor might work, but there's no way of handling scopes.

cc @birkenfeld thoughts?

### LINKED ISSUE #2890: Lint request: warn when variable declared in match pattern shadows other variable
"The Book" (TRPL) [warns](https://doc.rust-lang.org/book/second-edition/ch18-03-pattern-syntax.html#matching-named-variables) about variables declared in match patterns shadowing other variables. I have once been hit by this myself, so this seems to be a good idea for a lint.

### LINKED ISSUE #6563: shadow_unrelated false positive with question mark
Lint name: shadow_unrelated

I tried this code:

```rust
#[warn(clippy::shadow_unrelated)]
fn foo() -> Option<i32> {
    let a = 1_i32;
    let a = Some(1).map(|_| a)?;
    Some(a)
}
```

I expected to see this happen: no lint

Instead, this happened: shadow_unrelated lint

I needed a closure and a question mark to reproduce this. 🤷

### LINKED ISSUE #7588: shadow_unrelated false positives for variable named `val`
When I name a variable `val`, next fn call with `?` produces `shadow_unrelated` warning.
If I rename the variable, the warning disappears. Both on stable and nightly.
Reproducible with `cargo clippy` on this repo: https://github.com/romand/shadowed_val

I tried this code:

```rust
const fn bar() -> Result<(), Error> {
    Ok(())
}

pub fn foo() -> Result<(), Error> {
    let val = 1;
    bar()?;
    println!("{}", val);
    Ok(())
}
```

I expected to see this happen: no warnings

Instead, this happened: 
```
    Checking shadowed_val v0.1.0 (/home/romand/shadowed_val)
error: `bar()?` is being shadowed
  --> src/lib.rs:14:5
   |
14 |     bar()?;
   |     ^^^^^^
   |
   = note: `-D clippy::shadow-unrelated` implied by `-D warnings`
note: initialization happens here
  --> src/lib.rs:14:5
   |
14 |     bar()?;
   |     ^^^^^^
note: previous binding is here
  --> src/lib.rs:13:9
   |
13 |     let val = 1;
   |         ^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

error: could not compile `shadowed_val`

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: `bar()?` is being shadowed
  --> src/lib.rs:14:5
   |
14 |     bar()?;
   |     ^^^^^^
   |
   = note: `-D clippy::shadow-unrelated` implied by `-D warnings`
note: initialization happens here
  --> src/lib.rs:14:5
   |
14 |     bar()?;
   |     ^^^^^^
note: previous binding is here
  --> src/lib.rs:13:9
   |
13 |     let val = 1;
   |         ^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

error: build failed

cargo-clippy exited abnormally with code 101 at Thu Aug 19 13:58:18
```

### Meta

- `cargo clippy -V`: clippy 0.1.53 (53cb7b0 2021-06-17)
- `rustc -Vv`:
```
binary: rustc
commit-date: 2021-06-17
release: 1.53.0
LLVM version: 12.0.1
```

### LINKED ISSUE #7620: shadow_unrelated does not register a variable being shadowed
I accidentally shadowed a variable without knowing and the linter did not warn me about that. Not a huge deal but just wanted to report the issue.

Lint name:
shadow_unrelated

I tried this code:

```rust
let s = "&str";
println!("{}", s);

let s = String::from("String");
println!("{}", s);

let x = 1;
println!("{}", x);

let x = "2".parse::<i32>().unwrap();
println!("{}", x);
```

I expected to see this happen: Warning shown about variables `s` and `x` being shadowed when running `cargo clippy -- -W clippy::shadow_reuse -W clippy::shadow_same -W clippy::shadow_unrelated`.

Instead, this happened: No warning for variable `s` was shown.

### Meta

```
binary: rustc
commit-date: 2021-07-26
release: 1.54.0
LLVM version: 12.0.1
```

camsteffen:
changelog: Move shadow_unrelated to restriction
changelog: The shadow lints find a lot more shadows and are not limited to certain patterns

Drastically simplifies the implementation. Catches a lot more cases.

I removed the "initialization happens here" note. It is not helpful IMO.

Closes #318
Fixes #2890
Fixes #6563
Fixes #7588
Fixes #7620

### DIFF
--- a/clippy_lints/src/lib.register_pedantic.rs
+++ b/clippy_lints/src/lib.register_pedantic.rs
@@ -81,7 +81,6 @@ store.register_group(true, "clippy::pedantic", Some("clippy_pedantic"), vec![
     LintId::of(redundant_else::REDUNDANT_ELSE),
     LintId::of(ref_option_ref::REF_OPTION_REF),
     LintId::of(semicolon_if_nothing_returned::SEMICOLON_IF_NOTHING_RETURNED),
-    LintId::of(shadow::SHADOW_UNRELATED),
     LintId::of(strings::STRING_ADD_ASSIGN),
     LintId::of(trait_bounds::TRAIT_DUPLICATION_IN_BOUNDS),
     LintId::of(trait_bounds::TYPE_REPETITION_IN_BOUNDS),
--- a/clippy_lints/src/lib.register_restriction.rs
+++ b/clippy_lints/src/lib.register_restriction.rs
@@ -50,6 +50,7 @@ store.register_group(true, "clippy::restriction", Some("clippy_restriction"), ve
     LintId::of(same_name_method::SAME_NAME_METHOD),
     LintId::of(shadow::SHADOW_REUSE),
     LintId::of(shadow::SHADOW_SAME),
+    LintId::of(shadow::SHADOW_UNRELATED),
     LintId::of(strings::STRING_ADD),
     LintId::of(strings::STRING_TO_STRING),
     LintId::of(strings::STR_TO_STRING),
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -324,7 +324,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|| Box::new(same_name_method::SameNameMethod));
     store.register_late_pass(|| Box::new(map_clone::MapClone));
     store.register_late_pass(|| Box::new(map_err_ignore::MapErrIgnore));
-    store.register_late_pass(|| Box::new(shadow::Shadow));
+    store.register_late_pass(|| Box::new(shadow::Shadow::default()));
     store.register_late_pass(|| Box::new(unit_types::UnitTypes));
     store.register_late_pass(|| Box::new(loops::Loops));
     store.register_late_pass(|| Box::new(main_recursion::MainRecursion::default()));
--- a/clippy_lints/src/shadow.rs
+++ b/clippy_lints/src/shadow.rs
@@ -1,17 +1,14 @@
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::diagnostics::span_lint_and_note;
 use clippy_utils::source::snippet;
-use clippy_utils::{contains_name, higher, iter_input_pats};
-use rustc_hir::intravisit::FnKind;
-use rustc_hir::{
-    Block, Body, Expr, ExprKind, FnDecl, Guard, HirId, Local, MutTy, Pat, PatKind, Path, QPath, StmtKind, Ty, TyKind,
-    UnOp,
-};
-use rustc_lint::{LateContext, LateLintPass, LintContext};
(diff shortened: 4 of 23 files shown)


mikerite:
About re-categorizing:

I think we should have some kind of shadow lint on by default as it is a source of bugs.

I thought a while ago that we could have a more permissive shadow lint that only lints when the two bindings are unrelated _and_ have the same type. It's more likely you'll get a compiler error than a bug if you confuse two bindings with different types. Maybe you can try this variant?

llogiq:
I like both this simplification and @mikerite's idea of only linting on same-type shadows.

It still should be either a pedantic or restriction lint. And perhaps we could reinstate the `shadow` lint group that was lost when we created the other grouos.

llogiq:
The problem with this config usage is that it's one-size-fits all and there is no way to allow(shadow) partially.

camsteffen:
Can you give an example? The config seems to give enough control to me. But maybe I need to explain it...

The `shadow-allow-related` config creates a sliding scale of permissiveness. You are forced to pick a point on that scale and I think that is good. "*How much* does the shadow have to be related to be allowed?" Each option listed is one step more permissive than the previous one **and includes the previous options**. So `shadow-allow-related: reuse` means `allow(shadow_same, shadow_reuse)` and `warn(shadow_unrelated, shadow_unrelated_same_type)` Without the config, you are able to have a setup that doesn't make sense. For example, it doesn't make sense to just have `#![warn(shadow_reuse)]` because an instance of either `shadow_unrelated` or `shadow_unrelated_same_type` is a "worse" shadow and it won't be linted.

After this PR, you should be able to `#[allow]` on a single statement that creates a shadow (but I'll need to test that).

llogiq:
How do you configure the items where the lints apply or not?

camsteffen:
You can do `#[allow(clippy::shadow)]` on a function or a statement. Is that not enough?

llogiq:
But what if I want to allow only some variants in one part of the code and other variants elsewhere?

camsteffen:
I guess that level of tweaking just doesn't seem valuable to me. But I understand it would be a slight loss of functionality and that could frustrate users.

llogiq:
@camsteffen any news on this PR? What's your plan going forward?

camsteffen:
@llogiq thanks for the ping. This has been in the back of my mind. If it's okay with you, I would like to make the shadow lints restriction (just move shadow_unrelated) and leave a possible allow-by-default lint to a future enhancement.

camsteffen:
@llogiq @rustbot ready

llogiq:
r=me after rebase

PR #7346, opened 2021-06-13
fix 5707

### LINKED ISSUE #5707: redundant_clone: False positive with borrowed data
```rust
use std::path::PathBuf;

fn main() {
    let p = PathBuf::from("/");

    let mut h: HashMap<&str, &str> = HashMap::new();
    h.insert("orig-p", p.to_str().unwrap());

    let mut q = p.clone();
    q.push("foo");

    println!("{:?} {}", h, q.display());
}
```

This warns about a redundant clone:

```
warning: redundant clone
  --> src/main.rs:10:18
   |
10 |     let mut q = p.clone();
   |                  ^^^^^^^^ help: remove this
   |
   = note: `#[warn(clippy::redundant_clone)]` on by default
note: this value is dropped without further use
  --> src/main.rs:10:17
   |
10 |     let mut q = p.clone();
   |                 ^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
```

However, removing the clone as suggested is illegal, as we can't move `p` because of the reference stored in `h`:

```
error[E0505]: cannot move out of `p` because it is borrowed
  --> src/main.rs:10:17
   |
8  |     h.insert("orig-p", p.to_str().unwrap());
   |                        - borrow of `p` occurs here
9  |
10 |     let mut q = p;
   |                 ^ move out of `p` occurs here
...
13 |     println!("{:?} {}", h, q.display());
   |                         - borrow later used here

```

cargo clippy -V: clippy 0.0.212 (d4092ace 2020-05-11)

lengyijun:
changelog: ``[`redundant_clone`]``, fix #5707 

# Root problem of #5707 :
```
&2:&mut HashMap = &mut _4;
&3:&str = & _5;
_1 = HashMap::insert(move _2,move _3, _);
```

generate PossibleBorrower(_2,_1) and PossibleBorrower(_3,_1).  

However, it misses PossibleBorrower(_3,_2).

# My solution to #5707 :

When meet a function call, we should:
1. build PossibleBorrower between borrow parameters and return value (currently)
2. build PossibleBorrower between immutable borrow parameters and mutable borrow parameters (*add*)
3. build PossibleBorrower inside mutable borrow parameters (*add*)

For example: 
``` 
_2: &mut _22;
_3: &mut _;
_4: & _;
_5: & _;
_1 = call(move _2, move _3, move _4, move _5);
```
we need to build
1. return value with parameter(current implementataion)
 PossibleBorrower(_2,_1)
 PossibleBorrower(_3,_1)
 PossibleBorrower(_4,_1)
 PossibleBorrower(_5,_1)

2. between mutable borrow and immutable borrow
PossibleBorrower(_4,_2)
PossibleBorrower(_5,_2)
PossibleBorrower(_4,_3)
PossibleBorrower(_5,_3)

3. between mutable borrow and mutable borrow
PossibleBorrower(_3,_2)
PossibleBorrower(_2,_3)

  But that's not enough. 
 Modification to _2 actually apply to _22.
  So I write a `PossibleBorrowed` visitor, which tracks (borrower => possible borrowed) relation.
  For example (_2 => _22).
  However, a lot of problems exist here.

## Known Problems:
  1. not sure all `&mut`'s origin are collected.
  I'm not sure how to deal with `&mut` when meet a function call, so I didn't do it currently.
  Also, my implement is not flow sensitive, so it's not accurate.

```
foo(_2:&mut _, _3: &_)
```
This pr doesn't count _3 as origin of _2.

 2. introduce false negative
`foo(_2, _3)` will  emit PossibleBorrower(_3,_2) in this pr, but _3 and _2 may not have relation.
Clippy may feel that _3 is still in use because of _2, but actually, _3 is on longer needed and can be moved.

## Insight
  The key problem is determine where every `&mut` come from accurately.
  I think Polonius is an elegant solution to it. Polonius is flow sensitive and accurate.
  But I'm uncertain about whether we can import Polonius in rust-clippy currently.
  This pr actually is part of Polonius' functionality, I think.

# TODO
1. `cargo test` can't pass yet due to similar variable name

### DIFF
--- a/clippy_lints/src/redundant_clone.rs
+++ b/clippy_lints/src/redundant_clone.rs
@@ -88,8 +88,8 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
 
         let mir = cx.tcx.optimized_mir(def_id.to_def_id());
 
-        let possible_borrowed = {
-            let mut vis = PossibleBorrowedVisitor::new(mir);
+        let possible_origin = {
+            let mut vis = PossibleOriginVisitor::new(mir);
             vis.visit_body(mir);
             vis.into_map(cx)
         };
@@ -99,7 +99,7 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
             .iterate_to_fixpoint()
             .into_results_cursor(mir);
         let mut possible_borrower = {
-            let mut vis = PossibleBorrowerVisitor::new(cx, mir, possible_borrowed);
+            let mut vis = PossibleBorrowerVisitor::new(cx, mir, possible_origin);
             vis.visit_body(mir);
             vis.into_map(cx, maybe_storage_live_result)
         };
@@ -515,20 +515,20 @@ struct PossibleBorrowerVisitor<'a, 'tcx> {
     possible_borrower: TransitiveRelation<mir::Local>,
     body: &'a mir::Body<'tcx>,
     cx: &'a LateContext<'tcx>,
-    possible_borrowed: FxHashMap<mir::Local, HybridBitSet<mir::Local>>,
+    possible_origin: FxHashMap<mir::Local, HybridBitSet<mir::Local>>,
 }
 
 impl<'a, 'tcx> PossibleBorrowerVisitor<'a, 'tcx> {
     fn new(
         cx: &'a LateContext<'tcx>,
         body: &'a mir::Body<'tcx>,
-        possible_borrowed: FxHashMap<mir::Local, HybridBitSet<mir::Local>>,
+        possible_origin: FxHashMap<mir::Local, HybridBitSet<mir::Local>>,
     ) -> Self {
         Self {
             possible_borrower: TransitiveRelation::default(),
             cx,
             body,
-            possible_borrowed,
+            possible_origin,
         }
     }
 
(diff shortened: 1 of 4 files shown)


0969397110:
H

flip1995:
There are dogfood errors and you have to run `cargo dev fmt`.

Since this change heavily involves knowledge of MIR, I'd like to give this review to r? @oli-obk. Also cc @sinkuu since you've done the most work on this lint until now.

oli-obk:
I won't get to this for the next three weeks

oli-obk:
> Will introduce false negative

can you add a test showing this?

> not sure all &mut's origin are collected.

you already mentioned feeding &mut into function calls and getting something out. A similar situation occurs with deref projections. Dereferencing a raw pointer can subvert any checks we could possibly make.

The implementation of this PR lgtm as far as in "is an improvement to the status quo", as you're rejecting *more* code and now allowing anything that wasn't allowed before. I will merge it, but I think we (wg-mir-opt) should figure out how to make writing and analyzing mir optimizations better.

One of my main worries with this lint (and other mir analyses) is that it is hard to reason about their behaviour at the source level. We generate and track a lot of information, but we cannot "visualize" (textually or otherwise) this information and thus analyze it in detail. While there are such visualizations for specific analyses, we should figure out something generalizeable.

oli-obk:
Clippy dogfood is failing, and likely it would be good to rebase this PR first. After that r=me

lengyijun:
Thanks for reviewing such a longthy pr!

> can you add a test showing this?

Newly added test: `false_negative_5707`

>  I think we (wg-mir-opt) should figure out how to make writing and analyzing mir optimizations better.

I'd like to help it. You can appoint specific job for me. I can make a try.

oli-obk:
You still have the dogfood test failure (running clippy on clippy) to address before we can merge this.

> I'd like to help it. You can appoint specific job for me. I can make a try.

Let's talk about this in https://rust-lang.zulipchat.com/#narrow/stream/189540-t-compiler.2Fwg-mir-opt/topic/debugging.20tools.20for.20dataflow.20and.20mir.20opts

oli-obk:
@bors r+

Thanks!

PR #7352, opened 2021-06-14
Fix metadata code block syntax highlighting

camsteffen:
changelog: none

Fixes code fence block in the metadata collector to change ` ``` ` or ` ```ignore` or ` ```rust,ignore` to ` ```rust`. This makes syntax highlighting work more consistently on the website.

### DIFF
--- a/clippy_lints/src/utils/internal_lints/metadata_collector.rs
+++ b/clippy_lints/src/utils/internal_lints/metadata_collector.rs
@@ -9,6 +9,7 @@
 //! a simple mistake)
 
 use if_chain::if_chain;
+use rustc_ast as ast;
 use rustc_data_structures::fx::FxHashMap;
 use rustc_hir::{
     self as hir, def::DefKind, intravisit, intravisit::Visitor, ExprKind, Item, ItemKind, Mutability, QPath,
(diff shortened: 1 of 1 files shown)


clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -485,16 +486,44 @@ fn extract_attr_docs_or_lint(cx: &LateContext<'_>, item: &Item<'_>) -> Option<St
...
+                docs.push_str("```");
+                match parts.next() {
+                    // if no language is present, fill in "rust"
+                    None => docs.push_str("rust"),
+                    Some(first) => {
+                        docs.push_str(first);
+                        for part in parts {
+                            docs.push(',');
+                            docs.push_str(part);
+                        }
```
xFrednet, line 518: Are there actually any additional parts we want to include here? The website uses highlight.js which can only take one language and no additional parameters to my understanding. The best solution would probably to only keep the main language and drop the rest, unless someone knows of a case where we need this.

I actually added some extra handling in the website code that does the same [here](https://github.com/rust-lang/rust-clippy/pull/7279/files#diff-c618f97d2a35c1d819a1df7a09772499726a909c94959c8f6a4e5c5d7230ed0cR374-R381). Having it in the collector itself is something I didn't think of. Nicely done.

camsteffen: Yeah dropping the rest would be fine. I waffled on this. YAGNI.

Oh, the JS solution is nice too! I don't have a strong preference for one or the other. But we should just keep one. I'll let you decide which.

xFrednet: Good, we can merge this once the additional parts are gone.

Having it here makes more sense IMO. The other solution runs for each block on each page load, that's a bit excessive in comparison. And you added the nice part that no language means `rust`. I've also touched this method to adjust some markdown for the website in that PR, so it's the correct place I would say :upside_down_face:.

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -485,16 +486,32 @@ fn extract_attr_docs_or_lint(cx: &LateContext<'_>, item: &Item<'_>) -> Option<St
...
+        if let Some(info) = line.trim_start().strip_prefix("```") {
+            in_code_block = !in_code_block;
+            if in_code_block {
+                // remove rustdoc directives
+                let mut parts = info
+                    .trim()
+                    .split(',')
+                    .filter(|&s| !matches!(s, "" | "ignore" | "no_run" | "should_panic"));
+                // if no language is present, fill in "rust"
+                let lang = parts.next().unwrap_or("rust");
```
xFrednet, line 506: You can actually use `find()` here as we only want to take the first part now. (I've missed that on the first review, sorry)

xFrednet:
Thank you for the quick changes, it was fun racing you with the review :wink: 

@bors r+

PR #7355, opened 2021-06-15
Fix link in docs

Pyther99:
The [docs](https://github.com/rust-lang/rust-clippy/blob/master/doc/common_tools_writing_lints.md) had an incorrect link for "Checking if a type defines a specific method".

changelog: none


camsteffen:
Thanks!

@bors r+

PR #7357, opened 2021-06-15
check for unbalanced tick pairs in doc-markdown lint

### LINKED ISSUE #6753: Unbalanced tick marks lead to confusing/inverted error messages from clippy::doc-markdown
The title of the issue spoils the fun a bit, but I'll lead with the confusing error message a teammate ran into yesterday:

```
warning: you should put `encompassed_by` between ticks in the documentation
 --> src/main.rs:2:16
  |
2 | /// should be `encompassed_by` tick marks because they `contain_underscores`.
  |                ^^^^^^^^^^^^^^
  |
  = note: `-W clippy::doc-markdown` implied by `-W clippy::pedantic`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown

warning: you should put `contain_underscores` between ticks in the documentation
 --> src/main.rs:2:57
  |
2 | /// should be `encompassed_by` tick marks because they `contain_underscores`.
  |                                                         ^^^^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown

warning: you should put `unbalanced_tick` between ticks in the documentation
 --> src/main.rs:3:29
  |
3 | /// Because of the initial `unbalanced_tick` pair, the error message is
  |                             ^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown

warning: you should put `confusing_and_misleading` between ticks in the documentation
 --> src/main.rs:4:11
  |
4 | /// very `confusing_and_misleading`.
  |           ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown

warning: 4 warnings emitted
```

The doc comment that produces these odd warnings:

```rust
/// This is a doc comment with `unbalanced_tick marks and several words that
/// should be `encompassed_by` tick marks because they `contain_underscores`.
/// Because of the initial `unbalanced_tick` pair, the error message is
/// very `confusing_and_misleading`.
pub fn testing() {}
```

`unbalanced_tick` on the first line is missing its closing tick mark, which causes the "tick-mark-ness" of the remainder of the comment to be inverted. So technically all of the warnings are correct: `encompassed_by` is not inside tick marks (because the tick mark that appears to be at the beginning of `encompassed_by` is actually closing the tick mark from the first line, and the tick mark that appears to be at the ending of `encompassed_by` is actually opening a new tick mark section). However, they're very confusing! :)

ebobrow:
fixes #6753 

changelog: check for unbalanced tick pairs in doc-markdown lint


xFrednet:
Hey, have you decided where you want to span the lint after your changes? I think that that would affect the implementation a bit. The example output would suggest to me that your change points to the first back-tick. Is that correct?

I was thinking that the implementation could be simplified by spanning the lint over the complete paragraph instead of a single backtick. The bug ticket is actually more about the fact that the wrong spans have been marked. I believe the issue would also count as fixed when the whole thing would be marked with a message like: *the backticks are unbalanced, one might be missing a pair*. The implementation might then be pretty straight forward, as you only have to check if text contains backticks.

<details>
<summary>I've done some testing by printing the parsed tokens and these are the results:</summary>

```
/// Normal valid paragraph (Ok)
///
/// One missing `backtick (Lint)
///
/// Double unbalanced backtick from ``here to here` the end (Lint)
///
/// Double balanced back ticks ``start end`` looks good (Ok)
///
/// A longer `paragraph-where` the mistake is `here the `rest-is` valid again (Lint)
```

Results:
```
/// Normal valid paragraph (Ok)
    Start(Paragraph)
    Text(Borrowed("Normal valid paragraph (Ok)"))
    End(Paragraph)

/// One missing `backtick (Lint)
    Start(Paragraph)
    Text(Boxed("One missing `backtick (Lint)"))
    End(Paragraph)

/// Double unbalanced backtick from ``here to here` the end (Lint)
    Start(Paragraph)
    Text(Boxed("Double unbalanced backtick from ``here to here` the end (Lint)"))
    End(Paragraph)

/// Double balanced back ticks ``start end`` looks good (Ok)
    Start(Paragraph)
    Text(Borrowed("Double balanced back ticks "))
    Code(Borrowed("start end"))
    Text(Borrowed(" looks good (Ok)"))
    End(Paragraph)

/// A longer `paragraph-where` the mistake is `here the `rest-is` valid again (Lint)
    Start(Paragraph)
    Text(Borrowed("A longer "))
    Code(Borrowed("paragraph-where"))
    Text(Borrowed(" the mistake is "))
    Code(Borrowed("here the "))
    Text(Boxed("rest-is` valid again (Lint)"))
    End(Paragraph)
```

</details>

---

ebobrow:
Yes, currently I have the span only covering the first backtick. The other issue I was having is that if there are other errors after the unbalanced backtick, the spans on their lints are not correct because `pulldown_cmark` is treating them as code, but clippy is treating them as text. For example, the failing test:

```
error: you should put `unbalanced_tick` between ticks in the documentation
  --> $DIR/doc.rs:208:23
   |
LL | /// Because of the initial unbalanced_tick pair, the error message is
   |                       ^^^^^^^^^^^^^^^
```

I'm not sure if this is possible to fix. Should I just stop clippy from linting on a paragraph once it discovers unbalanced ticks?

> I was thinking that the implementation could be simplified by spanning the lint over the complete paragraph instead of a single backtick.

That's a good idea, thanks!

xFrednet:
> I'm not sure if this is possible to fix. Should I just stop clippy from linting on a paragraph once it discovers unbalanced ticks?

This is not necessary but would be a very nice feature. It could otherwise happen that a single change like adding one tick produces several other lint messages. I think that having this lint shadow all/most other doc lint's for one pharagraph would be good! :)

ebobrow:
Oops, not so sure what happened there. Sorry about that. Is there a way to remove those extra commits or does it not matter?

xFrednet:
Yes there is, you can try this: https://github.com/rust-lang/rust-clippy/pull/7270#pullrequestreview-675961183

I know that it worked for someone else :)

ebobrow:
Thank you! That worked. Seems like I just went too far back in the commit history.

I think this PR is ready for review now. Thanks for your help!

xFrednet:
Your welcome 🙃

I would also take over the review of this PR if that works with you? @flip1995 🙃

r? @flip1995

tests/ui/doc.rs
```diff
@@ -203,6 +203,12 @@ fn issue_2343() {}
 /// __|_ _|__||_|
 fn pulldown_cmark_crash() {}
 
+/// This is a doc comment with `unbalanced_tick marks and several words that
+/// should be `encompassed_by` tick marks because they `contain_underscores`.
+/// Because of the initial `unbalanced_tick` pair, the error message is
+/// very `confusing_and_misleading`.
+fn issue_6753() {}
```
xFrednet, line 210: Could you also add a test that has multiple Markdown paragraphs and headlines etc.

And one that has a backtick inside a code block like this:
```
// Moo - I'm a cow
` <-- notice the single backtick
// some more text
```

clippy_lints/src/doc.rs
```diff
@@ -485,6 +485,8 @@ fn check_doc<'a, Events: Iterator<Item = (pulldown_cmark::Event<'a>, Range<usize
     let mut in_heading = false;
     let mut is_rust = false;
     let mut edition = None;
+    let mut ticks_unbalanced = false;
```
xFrednet, line 488: You currently check for unbalanced backticks in the complete doc comment. I think it would make sense to only lint single paragraphs at once. You can use the `Start(_)` and `End(_)` events for that.

Example, this should only span the lint over B and check A as usual:
```rs
/// A
/// A
///
/// B` <- backtick
/// B
```

The bottom block that calls `check_text(...)` would also need to be moved.

clippy_lints/src/doc.rs
```diff
@@ -516,34 +518,54 @@ fn check_doc<'a, Events: Iterator<Item = (pulldown_cmark::Event<'a>, Range<usize
...
-                    Err(e) => e - 1,
-                };
-                let (begin, span) = spans[index];
-                if in_code {
-                    if is_rust {
-                        let edition = edition.unwrap_or_else(|| cx.tcx.sess.edition());
-                        check_code(cx, &text, edition, span);
-                    }
+                if text.contains('`') && !in_code {
+                    let (_, start) = spans[0];
```
xFrednet, line 528: Just a NIT: You can use [`clippy_util::line_span(..)`](https://github.com/rust-lang/rust-clippy/blob/d18441383362c1a55f72949677f972a2c42972c2/clippy_utils/src/lib.rs#L786-L801) to expand the message to the entire line. That will make the message a bit cleaner in this case.

```txt
// It should change this:
LL |   /// This is a doc comment with `unbalanced_tick marks and several words that
   |  ____^
LL | | /// should be `encompassed_by` tick marks because they `contain_underscores`.

// To only show the lines without a start indication:
LL |   /// This is a doc comment with `unbalanced_tick marks and several words that
LL |   /// should be `encompassed_by` tick marks because they `contain_underscores`.
```

clippy_lints/src/doc.rs
```diff
@@ -516,34 +518,54 @@ fn check_doc<'a, Events: Iterator<Item = (pulldown_cmark::Event<'a>, Range<usize
...
+                        span,
+                        "backticks are unbalanced; one may be missing a pair",
+                    );
+                    ticks_unbalanced = true;
                 } else {
-                    // Adjust for the beginning of the current `Event`
-                    let span = span.with_lo(span.lo() + BytePos::from_usize(range.start - begin));
-
-                    check_text(cx, valid_idents, &text, span);
+                    headers.safety |= in_heading && text.trim() == "Safety";
```
xFrednet, line 542: The following is a copy from the previous implementation. I think it would be better to add a `continue` in line 540 and unindent the code again :upside_down_face:

xFrednet (changes requested):
The implementation before and after your changes ignores HTML, could you add that as a note to the lint documentation in the *Known Problems* section.

The rest of the implementation looks good :+1: :upside_down_face:

ebobrow:
Thanks for the review! I've applied your changes. I also had to change some tests to pass CI because there is a 200 line limit on `.stderr` files. I removed some parts that seemed redundant, but if it's better I can split the tests into two files to keep them all.

clippy_lints/src/doc.rs
```diff
@@ -512,24 +516,45 @@ fn check_doc<'a, Events: Iterator<Item = (pulldown_cmark::Event<'a>, Range<usize
...
+                let (_, span) = get_current_span(spans, range.start);
+                paragraph_span = first_line_of_span(cx, span);
+            },
+            End(Paragraph) => {
+                if ticks_unbalanced {
+                    span_lint(
+                        cx,
+                        DOC_MARKDOWN,
+                        paragraph_span,
+                        "backticks are unbalanced; one may be missing a pair",
```
xFrednet, line 530: This message seems a bit off to me in comparison to other messages, but I can't quite place why. @flip1995 could you give this a second look?

xFrednet (changes requested):
> Thanks for the review! I've applied your changes. I also had to change some tests to pass CI because there is a 200 line limit on .stderr files. I removed some parts that seemed redundant, but if it's better I can split the tests into two files to keep them all.

It's usually better to add a second test file in such a case, that will also make it easier in the future to expand the tests if needed. For instance, to add lint support in headlines or edge cases.

Thank you for applying the requested changes so quickly. The implementation looks good to me besides the marked places. :+1:

clippy_lints/src/doc.rs
flip1995: If there is punctuation in between the message, you usually want to split it up in error message + help message. Which you can do here with `span_lint_and_help`. So put the first part in the error message and the second part in the help message.

Bonus points when you also provide the span to the unmatched backtick as the span for the help message.

xFrednet: Splitting the message up into `span_lint_and_help` would make sense to me as well :+1: 

> Bonus points when you also provide the span to the unmatched backtick as the span for the help message.

It is sadly a bit difficult to automatically determine which backtick is unmatched in longer paragraphs with multiple snippets. It could maybe show the backtick if there is only one?

Example:
```rust
/// Hello World! `some-code-here` more
/// text in the next line with an unmatched `backtick
/// and finally `the-end`
```

How should the lint know that it's the middle backtick that it should highlight? Marking the whole paragraph seems elegant enough to me.

flip1995: Yeah, that's why I said bonus points. If @ebobrow has a good idea how to do this, that would be nice, but if not, just pointing to the paragraph is definitely good enough.

ebobrow: Agreed. I don't think it's possible.

For the error message, does something like this look good?

```
error: backticks are unbalanced
  --> $DIR/unbalanced_ticks.rs:7:1
   |
LL | / /// This is a doc comment with `unbalanced_tick marks and several words that
LL | | /// should be `encompassed_by` tick marks because they `contain_underscores`.
LL | | /// Because of the initial `unbalanced_tick` pair, the error message is
LL | | /// very `confusing_and_misleading`.
   | |____________________________________^
   |
   = note: `-D clippy::doc-markdown` implied by `-D warnings`
   = help: a backtick may be missing a pair
```

xFrednet: I think that looks good 👍

xFrednet (changes requested):
I've notices one last false negative by running `cargo lintcheck`. The current implementation ignores items inside lists, as those use the `Start(Item)` and `End(Item)` events. Here is a simple reproduction:
```rs
#![deny(clippy::doc_markdown)]

/// - machine: x86_64,
pub fn last_fp() {}
// The x86_64 should trigger the lint validation for identifiers as it not inside backticks
```
I would approve of this PR once this last FP is fixed and test are added for it. The rest looks wonderful to me :upside_down_face: 

The tool otherwise noted one correct unbalanced backtick [here](https://docs.rs/rand/0.7.3/src/rand/distributions/weighted/mod.rs.html#49).

ebobrow:
Thank you! While I'm at it, are there any other tags I should include? The full list is fonud [here](https://docs.rs/pulldown-cmark/0.8.0/pulldown_cmark/enum.Tag.html). It probably wouldn't hurt to add Emphasis, Strong, and Strikethrough, right?

ebobrow:
Nevermind, none of the other tags can have backticks in them. I've added support for list items.

xFrednet (approved):
LGTM

cc: @flip1995 could you double check my review?

clippy_lints/src/doc.rs
```diff
@@ -485,6 +487,9 @@ fn check_doc<'a, Events: Iterator<Item = (pulldown_cmark::Event<'a>, Range<usize
     let mut in_heading = false;
     let mut is_rust = false;
     let mut edition = None;
+    let mut ticks_unbalanced = false;
+    let mut text_to_check: Vec<(CowStr<'_>, Span)> = Vec::new();
+    let mut paragraph_span = spans[0].1; // First line
```
flip1995, line 492: Can you be sure that `spans` isn't empty?

flip1995 (commented):
Just one question

clippy_lints/src/doc.rs
ebobrow: I don't think that should ever panic. I just did some tests where a function has no doc comment, or has an empty doc comment, and they were all fine. The only time that spans would be empty is if there isn't a doc comment, and if that's the case the `check_doc` function is never called:

https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/doc.rs#L435-L441

flip1995: That's good! Can you then add a comment or, even better, replace this with `spans.get(0).expect("<explanation>")`, please?

flip1995:
Thanks!

@bors r=xFrednet,flip1995

Issue #5394, opened 2020-03-30
[Roadmap] Turn `cargo dev` into Clippys `./x.py`


flip1995:
There's already `cargo dev` which makes Clippy development easier and more
pleasant. This can still be expanded, so that it covers more areas of the
development process.

Steps to completion:
- [x] `bless` execute UI-tests and update reference files
- [ ] ~~`init` include `setup-toolchain.sh` in `clippy_dev`~~ (outdated)
- [ ] `pr` (?) Run everything (fmt, update_lints, ...?), so that the changes probably pass CI
- [x] `deprecate` to deprecate a Clippy lint. This is especially useful if people want to uplift Clippy lints.
- [ ] Better integrate `cargo dev` in `rust-lang/rust` so that commands are also usable there
- [x] Add git commit hooks running `cargo dev` commands before committing (#7361)
- [x] `cargo dev rename_lint`(See issue #7799)
- [x] `cargo dev dogfood --fix`
- [ ] ... ideas?

These are some ideas, that I had, maybe we can even further expand the `clippy_dev` tool.

phansch:
I'm going to work on adding the `bless` subcommand to `clippy_dev` this week.

phansch:
@xFrednet Assuming, you want to work on the `cargo dev` integration into
`x.py`, here's a short summary of the current state of `bless`ing in rustc and
Clippy:

Rustc's `bless` is implemented in the `compiletest` library living [inside the rustc tree](https://github.com/rust-lang/rust/tree/master/src/tools/compiletest).
Clippy is using a custom fork of that library: [compiletest-rs](https://github.com/laumann/compiletest-rs).

The custom fork that Clippy is using, currently doesn't have a `bless`
feature. Instead, Clippy's `bless` is a custom implementation in `clippy_dev`.

Two months later, I think that was a bad choice on my end. It would have been
much better to add bless support to `compiletest-rs`. This way we would have
more similarities between rustc tools and the tools Clippy is using. It would
have also allowed `miri` to get bless support for their testsuite, I think.
Finally, I believe it would also be easier to integrate it with `x.py` this way.

Considering all that, I think we would be better off throwing away our custom bless
implementation and adding bless support to `compiletest-rs` instead.

My proposal would be:

1. I'm opening a PR to `compiletest-rs` with the addition of the `bless`
   feature. That is mostly copy/pasting of code.
2. The interesting part then,
   would be to change `cargo dev bless` to use the blessing mechanism of
[...]

PR #7361, opened 2021-06-16
Added `cargo dev setup git-hook` and updated `cargo dev setup intellij` including a `remove` command

xFrednet:
This PR enables our dev tool to install a git hook that formats the code before each commit and also runs `update_lints` to make sure that everything is registered correctly. The script is located at `util/etc/pre-commit.sh`. I found it reasonable to locate it in the `util` folder and decided to add a `etc` in correlation to the main rust repo and to bring a bit of structure into it.

* The hook can be installed via: `cargo dev setup git-hook`
* And removed via: `cargo dev remove git-hook`

cc: #5394 

The refactoring of `src/ide_setup.rs` to `src/setup/intellij.rs` is an extra commit to simplify the review.

---

Changes:
* Added `cargo dev setup git-hook` for formatting before every commit
* Added `cargo dev remove git-hook` to remove the hook again
* Added `cargo dev remove intellij` to remove rustc source path dependencies
* Changed `cargo dev ide_setup` to `cargo dev setup intellij`

changelog: none

This is only an internal change and therefore not worth an entry in the general change log.

---

Tested on:
* [x] Linux (by @xFrednet)
* [ ] Windows (All used commands run inside the git bash, so it's very likely to work as well @xFrednet)
* [ ] macOS


flip1995:
How much work would it be to add a ``cargo dev remove intellij``?

xFrednet:
Well, not much I believe, but I haven't looked into it, as I don't use IntelliJ. I also plan to add `cargo dev setup vscode-tasks` and the removal of it. I can take a look at the IntelliJ removal as well if you like. :upside_down_face:

flip1995:
That would be great! If you find that it would be too much work for this PR, let me know and I'll review it in the current state.

xFrednet:
Converted to draft until I've reviewed the work required to undo the IntelliJ setup command. The current version should still be fully working.

xFrednet:
I've looked into adding the `cargo dev remove intellij` command. It looks pretty simple, however I noticed that the current implementation also seems to be a bit broken:
* It doesn't add the new source paths to the `clippy_util/Cargo.toml`
* It also adds `clippy_util` as a dependency in `clippy_lints/Cargo.toml`
  ```
  clippy_utils = { path = "/home/xfrednet/workspace/rust/rust/compiler/clippy_utils" }
  ```

I want to fix them as well, I guess that this will be about 150 additional changes.

xFrednet:
I'm not done yet, this commit only cleans up some message and makes the setup more expandable. The rest of the work should still be reasonably straight forward. :upside_down_face:

xFrednet:
Okay, I've updated the dev tool to also include a remove command for intellij and give some nicer error messages when calling the setup command.
```sh
# Inject dependencies
cargo dev setup intellij
# Remove dependencies again
cargo dev remove intellij
```

I was debating if it should be renamed to something like `cargo dev setup rustc-sources --repo-path ...` and `cargo dev remove rustc-sources` but that also sounded a bit weird. I'm open to suggestions, or we leave it as it is :upside_down_face:. I'll try to test the git-hook on Windows soon, but I don't have access to a mac to test the hook on there as well.

clippy_dev/src/setup/git_hook.rs
```diff
@@ -0,0 +1,76 @@
...
+    let git_path = Path::new(REPO_GIT_DIR);
+    if !git_path.exists() || !git_path.is_dir() {
+        println!("error: clippy_dev was unable to find the `.git` directory");
+        return Err(());
+    }
+
+    // Make sure that we don't override an existing hook by accident
+    let path = Path::new(HOOK_TARGET_FILE);
+    if path.exists() {
+        if force_override || super::ask_yes_no_question("Do you want to override the existing pre-commit hook it?") {
```
flip1995, line 49: I wouldn't ask, but just print an error suggesting to add the flag.

clippy_dev/src/setup/git_hook.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    Ok(())
+}
+
+pub fn remove_hook() {
+    let path = Path::new(HOOK_TARGET_FILE);
+    if path.exists() {
+        if delete_git_hook_file(path).is_ok() {
+            println!("Git hook successfully removed :)");
```
flip1995, line 62:
```suggestion
            println!("git hook successfully removed");
```

Not in user facing code please.

clippy_dev/src/setup/git_hook.rs
```diff
@@ -0,0 +1,76 @@
...
+        }
+    } else {
+        println!("No pre-commit hook was found. You're good to go :)");
```
flip1995, line 65:
```suggestion
        println!("No pre-commit hook was found");
```

clippy_dev/src/setup/git_hook.rs
```diff
@@ -0,0 +1,76 @@
...
+    }
+}
+
+fn delete_git_hook_file(path: &Path) -> Result<(), ()> {
+    if fs::remove_file(path).is_err() {
+        println!("error: unable to delete existing pre-commit git hook");
```
flip1995, line 71:
```suggestion
        eprintln!("error: unable to delete existing pre-commit git hook");
```

clippy_dev/src/setup/git_hook.rs
```diff
@@ -0,0 +1,76 @@
...
+    if path.exists() {
+        if delete_git_hook_file(path).is_ok() {
+            println!("Git hook successfully removed :)");
+        }
+    } else {
+        println!("No pre-commit hook was found. You're good to go :)");
+    }
+}
+
+fn delete_git_hook_file(path: &Path) -> Result<(), ()> {
```
flip1995, line 69: Since you return `()` in both `Ok` and `Err` case, you could just return `bool`? (Could be a lint maybe?)

clippy_dev/src/setup/git_hook.rs
```diff
@@ -0,0 +1,76 @@
...
+            println!("Git hook successfully installed :)");
+        },
+        Err(err) => println!(
+            "error: unable to copy `{}` to `{}` ({})",
+            HOOK_SOURCE_FILE, HOOK_TARGET_FILE, err
+        ),
+    }
+}
+
+fn check_precondition(force_override: bool) -> Result<(), ()> {
```
flip1995, line 38: Same here: Why not just return `bool`?

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
+use std::fs;
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::{Path, PathBuf};
+
+// This module takes an absolute path to a rustc repo and alters the dependencies to point towards
+// the respective rustc subcrates instead of using extern crate xyz.
+// This allows rust analyzer to analyze rustc internals and show proper information inside clippy
```
flip1995, line 8:
```suggestion
// This allows intellij to analyze rustc internals and show proper information inside clippy
```

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
...
+// code. See https://github.com/rust-analyzer/rust-analyzer/issues/3517 and https://github.com/rust-lang/rust-clippy/issues/5514 for details
```
flip1995, line 9:
```suggestion
// code. See https://github.com/rust-lang/rust-clippy/issues/5514 for details
```

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
...
+    let path = path.join("compiler");
+    println!("info: looking for compiler sources at: {}", path.display());
+
+    if !path.exists() {
+        eprintln!("error: the given path does not exist");
+        return Err(());
+    }
+
+    if !path.is_dir() {
+        eprintln!("error: the given path is a file and not a directory");
```
flip1995, line 78:
```suggestion
        eprintln!("error: the given path is not a directory");
```

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
...
+        return Err(());
+    }
+
+    Ok(path)
+}
+
+fn inject_deps_into_project(rustc_source_dir: &Path, project: &ClippyProjectInfo) -> Result<(), ()> {
```
flip1995, line 85: `Result<(), ()>` -> `bool`

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
...
+            "error: unable to find the `{}` file for the project {}",
+            file_name, project
+        );
+        return Err(());
+    }
+
+    match fs::read_to_string(path) {
+        Ok(content) => Ok(content),
+        Err(err) => {
+            println!(
```
flip1995, line 116:
```suggestion
            eprintln!(
```

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
...
+        Err(())
+    } else {
+        Ok(())
+    }
+}
+
+/// `clippy_dev` expects to be executed in the root directory of Clippy. This function
+/// loads the given file or returns an error. Having it in this extra function ensures
+/// that the error message looks nice.
+fn read_project_file(file_path: &str, file_name: &str, project: &str) -> Result<String, ()> {
```
flip1995, line 103: I don't really see the need for passing the `file_name` here. Couldn't you just use the `file_path` for the error message? I even think this would make the message a bit clearer.

clippy_dev/src/setup/intellij.rs
```diff
@@ -0,0 +1,228 @@
...
+        return Ok(());
+    }
+
+    let extern_crates = lib_rs
+        .lines()
+        // only take dependencies starting with `rustc_`
+        .filter(|line| line.starts_with("extern crate rustc_"))
+        // we have something like "extern crate foo;", we only care about the "foo"
+        //              ↓          ↓
+        // extern crate rustc_middle;
```
flip1995, line 146:
```suggestion
        // extern crate rustc_middle;
        //              ^^^^^^^^^^^^
```

Avoid utf-8 when possible if there are still editors out there that aren't able to display them.

clippy_dev/src/setup/git_hook.rs
xFrednet: Yeah, I was a bit unsure if the smile was a bit to childish ^^ I'll know it for the next time :upside_down_face:

clippy_dev/src/setup/git_hook.rs
xFrednet: Simply using a `bool` can sometimes be a bit ambiguous, IMO. But you're right that it might be better to use a `bool` here.

clippy_dev/src/setup/intellij.rs
xFrednet: I used a `Result<(), ()>` in this case to use the `?` operator. I think it looks cleaner than two if statements in this case, or is there a nicer way to use this with bools?

flip1995: The only usage of `inject_deps_into_project` is in line 45 as `if inject_deps_into_project.is_err() { return; }`. So this is probably not the case anymore?

xFrednet: The command was referring to lines 86 and 87. The following code uses the result value from `Ok()` which would make it look something like this:

```diff
- let cargo_content = read_project_file(project.cargo_file)?;
- let lib_content = read_project_file(project.lib_rs_file)?;

+ let cargo_content = if let Ok(cargo_content) = read_project_file(project.cargo_file) {
+     cargo_content
+ } else {
+     return false;
+ }
+ let lib_content = if let Ok(lib_content) = read_project_file(project.lib_rs_file) {
+     lib_content
+ } else {
+     return false;
+ }
```

flip1995: Ah ok, in that case you can keep it as-is.

flip1995:
Code LGTM now. I'll give those commands a try locally and then this should be good to merge.

flip1995:
Rebased and good to merge. Thanks, great work!

@bors r+

xFrednet:
You're welcome! I plan to also add a setup to add vs-code tasks to compile and test Clippy. Could you check the git-hook suggestion in the linked ticket as that has been implemented in this PR #5394.

Was there a reason for the rebase? :sweat_smile: (Not complaining just interested :))

flip1995:
I had to rebase locally anyway (since this branch still used the old nightly), so I figured, why not push it. I didn't think this was necessary though.

I checked it off in #5394

xFrednet:
Ahh makes sense. Thank you :)

Issue #7328, opened 2021-06-06
New lint: suggest the use of Result::unwrap_or_else over Result::map_or_else


jhscheer:
### What it does
Suggest the use of `Result::unwrap_or_else` over `Result::map_or_else` **if** `map_or_else` is just used to unpack a successful result while handling an error.

### Categories (optional)

*What is the advantage of the recommended code over the original code*

Code will be shorter (and cleaner).

### Drawbacks

None.

### Example

```rust
.map_or_else(
    |e| handle_the_error(e),
    |n| n,
)
```

Could be written as:

```rust
.unwrap_or_else(|e| handle_the_error(e))
```

camsteffen:
I think an enhancement to `map_identity` would do.

KisaragiEffective:
I've found another case.

```
let res: Result<usize, usize> = foo();
let v: usize = foo.map_or_else(|x| x, |x| x);
```

can be following code:

```
let v: usize = foo.unwrap_or_else(|x| x)
```

GuillaumeGomez:
Adding a lint for it.

PR #7363, opened 2021-06-16
 use of Result::unwrap_or_else over Result::map_or_else

Valentine-Mario:
This PR is the lint suggested in #7328

### DIFF
--- a/clippy_lints/src/unwrap_or_else_over_map_or_else.rs
+++ b/clippy_lints/src/unwrap_or_else_over_map_or_else.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_help;
 use if_chain::if_chain;
-use rustc_hir::*;
+use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
@@ -35,7 +35,7 @@ declare_clippy_lint! {
     /// ```
     pub UNWRAP_OR_ELSE_OVER_MAP_OR_ELSE,
     pedantic,
-    "default lint description"
+    "use 'Result::unwrap_or_else' over 'Result::map_or_else'"
 }
 
 declare_lint_pass!(UnwrapOrElseOverMapOrElse => [UNWRAP_OR_ELSE_OVER_MAP_OR_ELSE]);
@@ -44,7 +44,7 @@ impl LateLintPass<'_> for UnwrapOrElseOverMapOrElse {
     fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
         if_chain! {
             //check if this is a method call eg func.map_or_else()
-            if let ExprKind::MethodCall(method, _t_span, args, _) = expr.kind;
+            if let ExprKind::MethodCall(method, t_span, args, _) = expr.kind;
             //check if the function name is map_or_else
             if method.ident.as_str() == "map_or_else";
             //check if the first arg is a closure
@@ -54,7 +54,7 @@ impl LateLintPass<'_> for UnwrapOrElseOverMapOrElse {
                 span_lint_and_help(
                     cx,
                     UNWRAP_OR_ELSE_OVER_MAP_OR_ELSE,
-                    _t_span,
+                    t_span,
                     "Result::unwrap_or_else is shorter and more succinet",
                     None,
                     "consider unwrap_or_else(|e| handle_the_error(e)) to unpack result",
--- a/clippy_lints/src/unwrap_or_else_over_map_or_else.rs
+++ b/clippy_lints/src/unwrap_or_else_over_map_or_else.rs
@@ -1,4 +1,5 @@
 use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::is_adjusted;
 use if_chain::if_chain;
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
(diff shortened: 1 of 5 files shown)


llogiq:
Hey there! I see you have put the basic structure together, now you may want to start fleshing out the test, then implement the actual lint.

Let me know if you need any help.

Valentine-Mario:
> Hey there! I see you have put the basic structure together, now you may want to start fleshing out the test, then implement the actual lint.
> 
> Let me know if you need any help.

Thank you. I definitely would reach out when I need help

Valentine-Mario:
> Hey there! I see you have put the basic structure together, now you may want to start fleshing out the test, then implement the actual lint.
> 
> Let me know if you need any help.

Hey, @llogiq I'm currently facing a challenge. So my test is passing even when they are not supposed to, thus I can't run `cargo dev bless`. But I've written a simple linter that would check for map_or_else method ident and verify if it is a closure. Your help would be greatly appreciated. Thanks

Valentine-Mario:
@llogiq  can you please help me with review

clippy_lints/src/unwrap_or_else_over_map_or_else.rs
```diff
@@ -0,0 +1,68 @@
...
+    ///    Ok("An even number :)")
+    /// }
+    ///
+    /// func_result(2).map_or_else(|e| println!("{:?}", e), |n| println!("{}", n))
+    /// Use instead:
+    /// ```rust
+    /// // example code which does not raise clippy warning
+    ///  /// let c= func_result(3).unwrap_or_else(|e| {
+    ///     println!("Error: {:?}", e);
+    ///    process::exit(1);
```
llogiq, line 32: I somehow doubt that exiting the process makes for a good suggestion.

llogiq (commented):
I feel the example suggestion is worse than the code it replaces. Ultimately I don't understand the reasoning behind this lint.

Both `map_or_else` and `unwrap_or_else` are about getting the return value, yet your examples only ever return `()`. Worse, `process::exit` for control flow is not generally applicable and in most cases misguided.

llogiq:
Having read the underlying issue, I think the lint was meant to catch an `id`-implementing closure in the `map` part of the `map_or_else`.

Valentine-Mario:
> I feel the example suggestion is worse than the code it replaces. Ultimately I don't understand the reasoning behind this lint.
> 
> Both `map_or_else` and `unwrap_or_else` are about getting the return value, yet your examples only ever return `()`. Worse, `process::exit` for control flow is not generally applicable and in most cases misguided.

I've updated the doc and test to give a clearer picture to the lint usage

llogiq:
I'm still not happy about the example.

Things I would consider lint-worthy:

`a.map_or_else(|_| 0, |n| n)` (but only in the absence of type adjustments) or `b.map_or_else(|x| something_else(x), std::convert::identity)`.

Note that both are expressions that return some value (of the `Ok`-part of the type of `a` and `b`, respectively).

As soon as the `map` part does some operation, I would consider linting the expression a false positive.

Valentine-Mario:
> I'm still not happy about the example.
> 
> Things I would consider lint-worthy:
> 
> `a.map_or_else(|_| 0, |n| n)` (but only in the absence of type adjustments) or `b.map_or_else(|x| something_else(x), std::convert::identity)`.
> 
> Note that both are expressions that return some value (of the `Ok`-part of the type of `a` and `b`, respectively).
> 
> As soon as the `map` part does some operation, I would consider linting the expression a false positive.

Thanks for that. I'm kind of new to contributing, so how do I check if the map part performs any form of operation?

llogiq:
For that you need to match on the second argument expression of the `map_or_else` method call (`args[2]`, because `[0]` is the receiver). Either it is an `ExprPath`, then you check if that matches `std::convert::identity`. Or it is an `ExprClosure` then you need to get the `Body` and see if it is an `ExprPath` with just one segment and the same ident as the closure's sole argument. In this case, check no type adjustments took place (see `eta_reduction.rs` as a reference).

flip1995:
> how do I check if the map part performs any form of operation?

With #7330 you can do this with the function `clippy_utils::is_expr_identity_function`

camsteffen:
Any thoughts on enhancing `map_identity` instead of creating a new lint? We could also include `Result::map_or`, `Option::map_or`, `Option::map_or_else`.

Valentine-Mario:
> For that you need to match on the second argument expression of the `map_or_else` method call (`args[2]`, because `[0]` is the receiver). Either it is an `ExprPath`, then you check if that matches `std::convert::identity`. Or it is an `ExprClosure` then you need to get the `Body` and see if it is an `ExprPath` with just one segment and the same ident as the closure's sole argument. In this case, check no type adjustments took place (see `eta_reduction.rs` as a reference).

Thanks for the explanation. I used a slightly different approach though. After getting the closure expr body, I checked if it was an expr block. Then if  block.stmts.is_empty(), then we could say that there is no type adjustments

Valentine-Mario:
> For that you need to match on the second argument expression of the `map_or_else` method call (`args[2]`, because `[0]` is the receiver). Either it is an `ExprPath`, then you check if that matches `std::convert::identity`. Or it is an `ExprClosure` then you need to get the `Body` and see if it is an `ExprPath` with just one segment and the same ident as the closure's sole argument. In this case, check no type adjustments took place (see `eta_reduction.rs` as a reference).

Thanks for this. In my case, I checked if it was an ```ExprKind::Lit```, then I checked the adjstments

llogiq:
As @flip1995 said, we already have a utils function to find identities. You can simply use that one.

And @camsteffen is right, we should extend existing lints that fit before creating a new one.

Valentine-Mario:
> As @flip1995 said, we already have a utils function to find identities. You can simply use that one.
> 
> And @camsteffen is right, we should extend existing lints that fit before creating a new one.

Not sure how the extending should work. Since map_identity matches for just `map` method name

camsteffen:
> Not sure how the extending should work. Since map_identity matches for just `map` method name

You can just edit the implementation of `map_identity` to match `map_or_else`. Ultimately you just need to do `span_lint_and_help(cx, MAP_IDENTITY, ..)`.

giraffate:
ping from triage @Valentine-Mario. Can you have any update on this?

giraffate:
ping from triage @Valentine-Mario. According to the triage procedure, I'm closing this because 2 weeks have passed with no activity. If you have more time to work on this, feel free to reopen this.

PR #7377, opened 2021-06-19
Do not trigger match-same-arms lint on never-type arms

### LINKED ISSUE #7331: match_same_arms should ignore todo!() branches
Lint name:
match_same_arms

I tried [this code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a07b322a2f140d8d42e008bc8282c5d7):

```rust
#![warn(clippy::pedantic)]
#![allow(dead_code, clippy::blacklisted_name)]

#[derive(Copy, Clone)]
enum Foo {A, B}

fn test(foo: Foo) {
    match foo {
        Foo::A => todo!(),
        Foo::B => todo!()
    }
}
```

No warnings. In this case, it's pretty clear that, eventually, we intend to do different things for A and B; combining them would only create more work when we inevitably split them apart later.

It's also worth noting that this structure (several match arms containing `todo!()`) is generated by rust-analyzer's "fill match arms" feature.

Instead, this happened:

Clippy warns me:

```
warning: this `match` has identical arm bodies
  --> src/lib.rs:10:19
   |
10 |         Foo::B => todo!()
   |                   ^^^^^^^
   |
  --> src/lib.rs:1:9
   |
1  | #![warn(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::match_same_arms)]` implied by `#[warn(clippy::pedantic)]`
note: same as this
  --> src/lib.rs:9:19
   |
9  |         Foo::A => todo!(),
   |                   ^^^^^^^
help: consider refactoring into `Foo::A | Foo::B`
  --> src/lib.rs:9:9
   |
9  |         Foo::A => todo!(),
   |         ^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_same_arms
   = note: this warning originates in the macro `$crate::panic::panic_2015` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Meta

Note: the code in this issue was tested in the playground, which I can't figure out how to get version information for; but I had the same issue (with real code) locally, with the below versions.

- `cargo clippy -V`: clippy 0.1.52 (07e0e2ec 2021-03-24)
- `rustc -Vv`:
  ```
  rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)
  binary: rustc
  commit-hash: 07e0e2ec268c140e607e1ac7f49f145612d0f597
  commit-date: 2021-03-24
  host: x86_64-apple-darwin
  release: 1.53.0-nightly
  LLVM version: 12.0.0
  ```

popzxc:
fixes #7331 

With this PR, never-type arms (such as `todo!()` and `unimplemeted!()`) will not result in the lint being triggered.

This has a side effect of `A => panic!("foo"), B => panic!("foo")` not being triggered, but I think it's hard to decide whether it's a placeholder for the future or not. I lean towards not considering it a false negative, at least unless someone directly reports it. 

changelog: `match-same-arms` lint is no longer triggered for arms that always panic (e.g. `todo`).

### DIFF
--- a/clippy_lints/src/matches.rs
+++ b/clippy_lints/src/matches.rs
@@ -2229,6 +2229,9 @@ fn lint_match_arms<'tcx>(cx: &LateContext<'tcx>, expr: &Expr<'_>) {
                 && SpanlessEq::new(cx)
                     .expr_fallback(eq_fallback)
                     .eq_expr(lhs.body, rhs.body)
+                // we don't need to spawn the lint for always-panicking branches,
+                // e.g. `A => todo!(), B => todo!()`
+                && !cx.typeck_results().expr_ty(lhs.body).is_never()
                 // these checks could be removed to allow unused bindings
                 && bindings_eq(lhs.pat, local_map.keys().copied().collect())
                 && bindings_eq(rhs.pat, local_map.values().copied().collect())
--- a/tests/ui/match_same_arms.rs
+++ b/tests/ui/match_same_arms.rs
@@ -53,4 +53,23 @@ mod issue4244 {
     }
 }
 
+mod issue7331 {
+    #![warn(clippy::pedantic)]
+    #![allow(dead_code, clippy::blacklisted_name)]
+
+    #[derive(Copy, Clone)]
+    enum Foo {
+        A,
+        B,
+    }
+
+    // Lint should not be triggered for panicking branches.
+    fn test(foo: Foo) {
+        match foo {
+            Foo::A => todo!(),
+            Foo::B => todo!(),
+        }
+    }
+}
+
 fn main() {}


flip1995:
See my comment on the issue: https://github.com/rust-lang/rust-clippy/issues/7331#issuecomment-864874396

I don't think this needs a fix. But as I wrote there:

> We may want to add a general message, that another possible fix would be to change the match arm body, instead of combining the match arms.

So if you want to improve this lint, I think this would be the better thing to do.

popzxc:
Makes sense. I guess I'll close this PR, and will open a new one (if I'll have some spare time for that).

PR #7379, opened 2021-06-19
Do not spawn blacklisted_name lint in test context

### LINKED ISSUE #7305: `blacklisted_name` lint should not run inside the test code
Lint name: `blacklisted_name`

I tried this code:

```rust
#[test]
fn test() {
   let foo_encoded = encode_object("foo"); // Actual functions are not important.
   let foo = parse_object(foo_encoded);
   assert_eq!(foo, "foo");
}
```

I expected to see this happen: `blacklisted_name` lint is not emitted. Foo is an OK name for tests, displaying something that is not really important as an object itself and telling programmers to focus on other things.

Instead, this happened: `blacklisted_name` lint is emitted.

### Meta

- `cargo clippy -V`: clippy 0.1.52 (9bc8c42b 2021-05-09)
- `rustc -Vv`:

  ```
  rustc 1.52.1 (9bc8c42bb 2021-05-09)
  binary: rustc
  commit-hash: 9bc8c42bb2f19e745a63f3445f1ac248fb015e53
  commit-date: 2021-05-09
  host: aarch64-apple-darwin
  release: 1.52.1
  LLVM version: 12.0.0
  ```

popzxc:
---

fixed #7305 

changelog: `blacklisted_name` lint is not spawned in the test context anymore.

### DIFF
--- a/clippy_lints/src/blacklisted_name.rs
+++ b/clippy_lints/src/blacklisted_name.rs
@@ -1,6 +1,6 @@
-use clippy_utils::diagnostics::span_lint;
+use clippy_utils::{diagnostics::span_lint, is_test_module_or_function};
 use rustc_data_structures::fx::FxHashSet;
-use rustc_hir::{Pat, PatKind};
+use rustc_hir::{Item, Pat, PatKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 
@@ -25,18 +25,37 @@ declare_clippy_lint! {
 #[derive(Clone, Debug)]
 pub struct BlacklistedName {
     blacklist: FxHashSet<String>,
+    test_modules_deep: u32,
 }
 
 impl BlacklistedName {
     pub fn new(blacklist: FxHashSet<String>) -> Self {
-        Self { blacklist }
+        Self {
+            blacklist,
+            test_modules_deep: 0,
+        }
+    }
+
+    fn in_test_module(&self) -> bool {
+        self.test_modules_deep != 0
     }
 }
 
 impl_lint_pass!(BlacklistedName => [BLACKLISTED_NAME]);
 
 impl<'tcx> LateLintPass<'tcx> for BlacklistedName {
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if is_test_module_or_function(cx.tcx, item) {
+            self.test_modules_deep = self.test_modules_deep.saturating_add(1);
+        }
+    }
+
     fn check_pat(&mut self, cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>) {
+        // Check whether we are under the `test` attribute.
+        if self.in_test_module() {
+            return;
+        }
+
         if let PatKind::Binding(.., ident, _) = pat.kind {
             if self.blacklist.contains(&ident.name.to_string()) {
                 span_lint(
@@ -46,6 +65,12 @@ impl<'tcx> LateLintPass<'tcx> for BlacklistedName {
                     &format!("use of a blacklisted/placeholder name `{}`", ident.name),
                 );
             }
+        }
+    }
+
+    fn check_item_post(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if is_test_module_or_function(cx.tcx, item) {
+            self.test_modules_deep = self.test_modules_deep.saturating_sub(1);
         }
     }
 }
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -1727,3 +1727,15 @@ pub fn is_hir_ty_cfg_dependant(cx: &LateContext<'_>, ty: &hir::Ty<'_>) -> bool {
         }
     }
 }
+
+/// Checks whether item either has `test` attribute applied, or
+/// is a module with `test` in its name.
+pub fn is_test_module_or_function(tcx: TyCtxt<'_>, item: &Item<'_>) -> bool {
+    if let Some(def_id) = tcx.hir().opt_local_def_id(item.hir_id()) {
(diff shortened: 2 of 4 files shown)


clippy_lints/src/blacklisted_name.rs
```diff
@@ -37,6 +38,14 @@ impl_lint_pass!(BlacklistedName => [BLACKLISTED_NAME]);
 
 impl<'tcx> LateLintPass<'tcx> for BlacklistedName {
     fn check_pat(&mut self, cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>) {
+        // Check whether we are under the `test` attribute.
+        if let Some(def_id) = cx.tcx.hir().opt_local_def_id(pat.hir_id) {
```
camsteffen, line 42: I think this is happening: This is always `None` because a `Pat` is not an item. The test does not lint because `#[test]` is not compiled in ui tests.

popzxc: Should be fixed now. I'm not really happy with the hack required to test the code (not only check for `test`, but also for `ignored`), but I didn't found a more elegant way to do it.

If there is a more correct way, I'd be grateful for any hints.

clippy_lints/src/blacklisted_name.rs
```diff
@@ -37,6 +38,24 @@ impl_lint_pass!(BlacklistedName => [BLACKLISTED_NAME]);
...
+            if cx.tcx.has_attr(def_id, sym::test) {
+                // We don't want to check for blacklisted names in the test code.
+                return;
+            }
+
+            // Hack: `compiletest-rs` do not compile code with `#[test]` attribute.
+            // In order to test that this part of code actually works, we use `#[ignore]`
+            // in tests (as an attribute that does not cause code to be not compiled,
+            // but at the same time highly tied with tests).
+            if cx.tcx.has_attr(def_id, sym::ignore) {
```
flip1995, line 54: We can't change the lint behavior just so it works with our test suite. Our test suite is not the production environment for Clippy.

clippy_lints/src/blacklisted_name.rs
```diff
@@ -37,6 +38,24 @@ impl_lint_pass!(BlacklistedName => [BLACKLISTED_NAME]);
 
 impl<'tcx> LateLintPass<'tcx> for BlacklistedName {
     fn check_pat(&mut self, cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>) {
+        // Check whether we are under the `test` attribute.
+        if let Some(enclosing) = cx.enclosing_body {
```
flip1995, line 42: I don't think this works. E.g.:

```rust
#[test]
fn test_outer() {
    fn no_test_inner() {
        let foo = 42;
    }
    no_test_inner();
}
```

This will probably lint, since the enclosing body of `let foo ...` is `no_test_inner` which doesn't have the `#[test]` attribute.

flip1995 (changes requested):
We already have some lints, that shouldn't trigger in test code. This is how we usually handle those situations:

When going into a test context, save this information in the lint struct:

https://github.com/rust-lang/rust-clippy/blob/48fa1dc8bb843aaa7b912beed1666ca9c90f5363/clippy_lints/src/wildcard_imports.rs#L108-L111

When leaving the test context:

https://github.com/rust-lang/rust-clippy/blob/48fa1dc8bb843aaa7b912beed1666ca9c90f5363/clippy_lints/src/wildcard_imports.rs#L186-L190

clippy_lints/src/blacklisted_name.rs
```diff
@@ -48,4 +67,14 @@ impl<'tcx> LateLintPass<'tcx> for BlacklistedName {
...
+    fn check_item_post(&mut self, _: &LateContext<'_>, item: &Item<'_>) {
+        if is_test_module_or_function(item) {
+            self.test_modules_deep = self.test_modules_deep.saturating_sub(1);
+        }
+    }
+}
+
+fn is_test_module_or_function(item: &Item<'_>) -> bool {
+    matches!(item.kind, ItemKind::Mod(..)) && item.ident.name.as_str().contains("test")
 }
```
flip1995, line 80: Can you move this function to `clippy_utils` please and add a check, if the item has the `#[test]` attribute, in addition to the check for the module name, please?

clippy_utils/src/lib.rs
```diff
@@ -51,6 +51,7 @@ pub mod ptr;
 pub mod qualify_min_const_fn;
 pub mod source;
 pub mod sugg;
+pub mod test_module;
```
flip1995, line 54: No need for this to be in its own module. Just put it in `lib.rs`

tests/ui/blacklisted_name.rs
```diff
@@ -43,3 +43,19 @@ fn issue_1647_ref_mut() {
...
 }
+
+mod tests {
+    fn issue_7305() {
+        // `blackisted_name` lint should not be triggered inside of the test code.
+        let foo = 0;
+
+        // However, we an additional problematic code here that triggers another lint
+        // just to ensure that this code was compiled & processed.
+        let nOn_SnakeCase_ident = 10u8;
```
flip1995, line 54: I don't think this is necessary anymore and would like to avoid triggering unrelated lints in this test file.

flip1995 (changes requested):

tests/compile-test.rs
```diff
@@ -77,7 +77,7 @@ fn default_config() -> compiletest::Config {
     }
 
     config.target_rustcflags = Some(format!(
-        "--emit=metadata -L {0} -L {1} -Dwarnings -Zui-testing {2}",
+        " --emit=metadata -L {0} -L {1} -Dwarnings -Zui-testing {2}",
```
flip1995, line 80: Unnecessary change?

flip1995 (approved):
Thanks! LGTM overall. One small thing left. Please squash the commits after fixing that.

flip1995:
@bors r+

Thanks!

Issue #7343, opened 2021-06-10
Tests fail with conflicting external dependencies in the target directory


camsteffen:
This is about the "multiple matching crates found" error or "multiple rlibs found" error that occurs rather frequently when running Clippy UI tests.

There are already multiple issues opened, but I want to explain the problem, tell what has been done to try and fix it, centralize discussion and possibly find a solution.

In short, the current state of things is okay/working, but fragile. For instance, every time I run `cargo check` during local development, or build with a different set of feature flags, I bump into this error and have to `rm target/debug/deps/lib*` and try again. The error also occurs now and then in CI and it's a real show stopper.

One thing to keep in mind is that this error can be seen as a good thing as it may prevent rustc bootstrapping from needlessly recompiling external crates, adding more time to a process that we want to quicken in any way we can. So a fix along the lines of "just rebuild in another target directory" may not be desirable.

### Detailed Explanation

Clippy UI tests use external crate dependencies from crates.io. It is the only project in the rustc repository with this feature. For instance, we have some lints that are specific to serde. We use (a [fork](https://github.com/Manishearth/compiletest-rs) of) compiletest like rustc. And compiletest does not have first-class support of this feature. In order to use external dependencies with compiletest, you need to have them compiled beforehand (by cargo) and provide `-L dir` or `--extern file` rust flags to compiletest. This is what Clippy [does](https://github.com/rust-lang/rust-clippy/blob/f7d09b45e9dbb32c5524d5f3b0838401c848bbf2/tests/compile-test.rs#L80). We provide the Clippy target directory (e.g. target/debug/deps) so that any crates listed in the dependencies of the main `Cargo.toml` will be compiled and available for use in UI tests with `extern crate`.

This fails when there are multiple versions of the same library in the `-L` target dir(s) because rustc won't know which one to use and will fail with a "multiple matching crates" error. As long as only one version of every library is compiled, with the same features, everything is fine. That is basically the current solution. But a variety of scenarios can break this requirement. It's hard to maintain.

We have one partial fix where we give preference to libs in the clippy (or stage-tools) target dir by [scanning](https://github.com/rust-lang/rust-clippy/blob/f7d09b45e9dbb32c5524d5f3b0838401c848bbf2/tests/compile-test.rs#L35) the directory and using `--extern` with libs found there. This is where we have our dreaded "multiple rlibs found" panic case.

[...]

PR #7380, opened 2021-06-20
Improve panic message on "Found multiple rlibs" error in compile-test

popzxc:
Related to #7343 

When I first met this error I was pretty much confused, so I thought it may be a good idea to:

- Give a hint on what to do to users that don't want to dig into specifics and just want to quickly resolve the issue.
- Give a link for those who are interested in details.

## Old appearance:

Screenshot 2021-06-20 at 08 30 34

## New appearance:

Screenshot 2021-06-20 at 08 32 18

changelog: none


jyn514:
`cargo clean` will do nothing in rust-lang/rust. Can you suggest "Remove the stageN-tools directory" instead when RUSTC_BOOTSTRAP is set in the environment? (or any other way of detecting x.py; BOOTSTRAP_PYTHON should work too)

flip1995:
@bors r+

Thanks!

PR #7382, opened 2021-06-20
Fix wrong config option being suggested for deprecated wrong_pub_self_convention lint

matthiaskrgr:
Problem:
for code like
````rust
#![warn(clippy::wrong_pub_self_convention)]
fn main() {
    println!("Hello, world!");
}
````
clippy will issue a warning to use a clippy.toml option instead:

````
warning: lint `clippy::wrong_pub_self_convention` has been removed: set the `avoid_breaking_exported_api` config option to `false` to enable the `wrong_self_convention` lint for public items
 --> src/main.rs:2:9
  |
2 | #![warn(clippy::wrong_pub_self_convention)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(renamed_and_removed_lints)]` on by default
````

But using the lint name as seen in the warning message
`echo "avoid_breaking_exported_api = true\n" > clippy.toml`

Will cause an error:
````
error: error reading Clippy's configuration file `/tmp/clippytest/clippy.toml`: unknown field `avoid_breaking_exported_api`, expected one of `avoid-breaking-exported-api`, ...
````

Replace the underscores with dashes in the deprecation message.

changelog: avoid_breaking_exported_api: suggest correct clippy config toml option in the deprecation message

### DIFF
--- a/clippy_lints/src/deprecated_lints.rs
+++ b/clippy_lints/src/deprecated_lints.rs
@@ -149,7 +149,7 @@ declare_deprecated_lint! {
     /// enables the `enum_variant_names` lint for public items.
     /// ```
     pub PUB_ENUM_VARIANT_NAMES,
-    "set the `avoid_breaking_exported_api` config option to `false` to enable the `enum_variant_names` lint for public items"
+    "set the `avoid-breaking-exported-api` config option to `false` to enable the `enum_variant_names` lint for public items"
 }
 
 declare_deprecated_lint! {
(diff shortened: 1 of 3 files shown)


matthiaskrgr:
Hmm, apparently `cargo dev update_lints` will revert my fix :laughing:

matthiaskrgr:
Seems that
````rust
    file_change |= replace_region_in_file(
        Path::new("clippy_lints/src/lib.rs"),
        "begin deprecated lints",
        "end deprecated lints",
        false,
        update_mode == UpdateMode::Change,
        || gen_deprecated(deprecated_lints.iter()),
    )
    .changed;
````
is causing this (`update_lints.rs`)

matthiaskrgr:
I think i fixed it

flip1995:
@bors r+

Thanks!

PR #7394, opened 2021-06-22
Update opener.

ehuss:
This updates the opener dependency, to try to reduce the duplicate dependencies in the rust-lang/rust repo.  The changelog is [here](https://github.com/Seeker14491/opener/blob/master/CHANGELOG.md#050---2021-06-11), and I don't expect it to have much of a change to anyone.

changelog: none


flip1995:
Tested it locally in the only  place I think this is used `cargo dev serve` and this still works for me.

Before I'm merging this here, do you want to do this directly in the `rust-lang/rust` repo, so you don't have to wait for it to sync back in a week?

Otherwise r=me. @bors delegate+

ehuss:
I'm in no hurry, this can ride with the rest of the updates. Thanks!

@bors r+

Issue #7398, opened 2021-06-24
clippy-driver usage returns the "cargo clippy" which will make user confused


sungguoshuai:
clippy-driver usage returns the "cargo clippy" which will make user confused

#clippy-driver --help
.......usage: cargi clippy.....

I found the latest code is the same. Could I modify the user prompt?

xFrednet:
I believe that the `clippy-driver` executable gets indirectly called via `cargo clippy` and therefore uses the name `cargo clippy` inside the help information. (I might be wrong on that)

Is there a reason why  the user might call `clippy-driver` directly, outside of Clippy and rustc development?

sungguoshuai:
Thank you for your reply,sir.
I just install the clippy in fedora,and find it supply the /usr/bin/clippy-driver executable, just confused about this phenomenon.
If no scenes user minght call it directly ,close the issue please.
In addition, I found the clippy-driver's help information has one more option than cargo clippy,that's "--rustc",so I feel it should be different from cargo clippy.

xFrednet:
That makes sense, I would wait on some more feedback from a more knowledgeable contributor before closing this and the PR as I'm not totally sure on this point.

You're welcome, sir :upside_down_face:.

PR #7399, opened 2021-06-24
clippy-driver usage should tell user the usage about clippy-driver but not cargo clippy

sungguoshuai:

fixes:  #7398 

---

changelog: modify the usage prompt for clippy-driver.

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -126,7 +126,7 @@ fn display_help() {
 Checks a package to catch common mistakes and improve your Rust code.
 
 Usage:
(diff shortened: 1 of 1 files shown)


camsteffen:
We can't just change it to say `clippy-driver` because the help message should not reference `cargo-clippy` when a user runs `cargo clippy --help`. I suppose we could make the output dynamic, depending on whether cargo was used. But I don't think `cargo-clippy` is used directly very often.

camsteffen:
Closing due to inactivity. Feel free to continue discussion.

PR #7401, opened 2021-06-25
WIP: PR to test #7298

flip1995:
r? ghost

nothing to see here, just testing #7298

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -2,7 +2,7 @@
 uitest = "test --test compile-test"
 dev = "run --target-dir clippy_dev/target --package clippy_dev --bin clippy_dev --manifest-path clippy_dev/Cargo.toml --"
 lintcheck = "run --target-dir lintcheck/target --package lintcheck --bin lintcheck --manifest-path lintcheck/Cargo.toml  -- "
-collect-metadata = "test --test dogfood --features metadata-collector-lint -- run_metadata_collection_lint --ignored"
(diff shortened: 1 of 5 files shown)


flip1995:
Whoops forgot the @ before ghost...

flip1995:
Hmmm, I'm inclined to break stable deploy for one cycle. This checking out of the master deploy script really helped in the past, but now it is a bit of a PITA. I would like to keep doing this, since it helps usually more than it hurts. 

With the next release, the deployment on stable will break, since it still has to run the python script to create the documentation.  This will require a fix by hand. But I rather do this, than copying the deploy script for a whole release cycle, and then spend another release cycle to remove the copy again.

Beta shouldn't be affected by this, since it will receive the new doc generation before the deploy job is run.

flip1995:
Hmmm, I'm inclined to break stable deploy for one cycle. This checking out of the master deploy script really helped in the past, but now it is a bit of a PITA. I would like to keep doing this, since it helps usually more than it hurts. 

With the next release, the deployment on stable will break, since it still has to run the python script to create the documentation.  This will require a fix by hand. But I rather do this, than copying the deploy script for a whole release cycle, and then spend another release cycle to remove the copy again.

Beta shouldn't be affected by this, since it will receive the new doc generation before the deploy job is run.

flip1995:
Debug build: 2 min clean build; 50 seconds cached build

Release build: 5 min clean build; Always does a clean build...

PR #7406, opened 2021-06-26
Don't use exact definition of std's ErrorKind in test.

m-ou-se:
Every time we add something to `std::io::ErrorKind`, this clippy test breaks. This is especially annoying because it usually only happens when bors is testing it (e.g. in a rollup), as clippy tests are normally not run for PRs with just library changes.

For example: https://github.com/rust-lang/rust/pull/85746#issuecomment-868886893

This changes the test to not rely on the definition from std.

---

changelog: none

### DIFF
--- a/tests/ui/auxiliary/non-exhaustive-enum.rs
+++ b/tests/ui/auxiliary/non-exhaustive-enum.rs
@@ -1,0 +1,23 @@
+#[non_exhaustive]
+pub enum ErrorKind {
+    NotFound,
+    PermissionDenied,
+    ConnectionRefused,
+    ConnectionReset,
+    ConnectionAborted,
+    NotConnected,
+    AddrInUse,
+    AddrNotAvailable,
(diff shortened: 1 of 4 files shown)


tests/ui/auxiliary/non-exhaustive-enum.rs
```diff
@@ -0,0 +1,23 @@
...
+    InvalidInput,
+    InvalidData,
+    TimedOut,
+    WriteZero,
+    Interrupted,
+    Other,
+    UnexpectedEof,
+    Unsupported,
+    OutOfMemory,
+}
```
camsteffen, line 23: Maybe trim this down and/or rename it so it doesn't even seem like we're trying to match std.

flip1995 (approved):
LGTM, except Cam's comment. I also don't think we need that many variants and/or try to match std. 2 or 3 should be enough.

PR #7407, opened 2021-06-26
Don't suggest doc(hidden) or unstable variants in wildcard lint

m-ou-se:
Clippy's wildcard lint would suggest doc(hidden) and unstable variants for non_exhaustive enums, even though those aren't part of the public interface (yet) and should only be matched on using a `_`, just like potential future additions to the enum. There was already some logic to exclude a *single* doc(hidden) variant. This extends that to all hidden variants, and also hides `#[unstable]` variants.

See https://github.com/rust-lang/rust/pull/85746#issuecomment-868886893

This PR includes https://github.com/rust-lang/rust-clippy/pull/7406 as the first commit.

Here's the diff that this PR adds on top of that PR: https://github.com/m-ou-se/rust-clippy/compare/std-errorkind...m-ou-se:doc-hidden-variants

---

changelog: No longer suggest unstable and doc(hidden) variants in wildcard lint. wildcard_enum_match_arm, match_wildcard_for_single_variants


clippy_lints/src/matches.rs
```diff
@@ -1033,7 +1033,7 @@ fn check_wild_enum_match(cx: &LateContext<'_>, ex: &Expr<'_>, arms: &[Arm<'_>])
 
     // Accumulate the variants which should be put in place of the wildcard because they're not
     // already covered.
-    let mut missing_variants: Vec<_> = adt_def.variants.iter().collect();
+    let mut missing_variants: Vec<_> = adt_def.variants.iter().filter(|x| !is_hidden(cx, x)).collect();
```
xFrednet, line 1036: I think here it would be good to differentiate between the enums origin. It would make sense to me to also cover hidden/unstable variations if they originate from the same crate. (I couldn't find a check for that in the current implementation)

I'm also debating if that should be configurable, but I believe there is no real reason for that.

clippy_lints/src/matches.rs
```diff
@@ -1118,7 +1118,7 @@ fn check_wild_enum_match(cx: &LateContext<'_>, ex: &Expr<'_>, arms: &[Arm<'_>])
 
     match missing_variants.as_slice() {
         [] => (),
-        [x] if !adt_def.is_variant_list_non_exhaustive() && !is_doc_hidden(cx, x) => span_lint_and_sugg(
+        [x] if !adt_def.is_variant_list_non_exhaustive() => span_lint_and_sugg(
```
xFrednet, line 1121: Your implementation effected the emission of the `match_wildcard_for_single_variants` lint. I'm not sure if the lint could now maybe be triggered by something like this:

```rs
enum AlphabetEnum {
    A,
    B,
    C,
    #[doc(hidden)]
    __Private,
}
match Enum::A {
    AlphabetEnum::A => (),
    AlphabetEnum::B => (),
    _ => (),
}
```

Could you add a test for that case :)

xFrednet (changes requested):
Could you update the changelog message to mention the specific lints that are effected by this change? That makes writing the change log easier in the end. In this case, it should be the lint `wildcard_enum_match_arm`

The rest looks good to me.

---

Hey @flip1995, I would take over the review of this PR if that works for you :)

flip1995:
(Un-assigning from phansch, since he's taking a break, thanks for taking over the review @xFrednet!)

clippy_lints/src/matches.rs
flip1995: I don't think this has to be done in this PR. I like this idea though. Can you open a good-first-issue issue for this, please?

clippy_lints/src/matches.rs
flip1995: Great catch!

flip1995:
@xFrednet can you review my changes, please? If any changes should be required I'll do them in a follow-up PR, I want to get this merged, so I can sync Clippy to Rust. 

@bors r+

flip1995:
The github issue preventing bors to do anything should be fixed 

@bors retry

clippy_lints/src/matches.rs
xFrednet: cc: #7419

tests/ui/wildcard_enum_match_arm.stderr
```diff
@@ -1,38 +1,44 @@
...
 LL |         _ => {},
-   |         ^ help: try this: `ErrorKind::PermissionDenied | ErrorKind::ConnectionRefused | ErrorKind::ConnectionReset | ErrorKind::ConnectionAborted | ErrorKind::NotConnected | ErrorKind::AddrInUse | ErrorKind::AddrNotAvailable | ErrorKind::BrokenPipe | ErrorKind::AlreadyExists | ErrorKind::WouldBlock | ErrorKind::InvalidInput | ErrorKind::InvalidData | ErrorKind::TimedOut | ErrorKind::WriteZero | ErrorKind::Interrupted | ErrorKind::Other | ErrorKind::UnexpectedEof | ErrorKind::Unsupported | ErrorKind::OutOfMemory | _`
+   |         ^ help: try this: `ErrorKind::PermissionDenied | _`
 
-error: aborting due to 5 previous errors
+error: wildcard matches known variants and will also match future added variants
+  --> $DIR/wildcard_enum_match_arm.rs:101:13
+   |
+LL |             _ => (),
+   |             ^ help: try this: `Enum::B | _`
```
xFrednet, line 41: I find it confusing that the lint suggests here to use the enum value in the same arm as the wildcard. That would mean that future additions would also run into this branch.

It would be good to either add a note to explain why a wildcard has to be present here or to make the suggestion to have two arms like this:
```rs
Enum::B => {},
_ => { /* Hidden values */ }
```
Adding a note that the enum includes hidden, or unstable values would still be good.

xFrednet (commented):
The rest looks good to me :+1:

tests/ui/wildcard_enum_match_arm.stderr
flip1995: A `X | _` pattern is already caught by another lint, so this shouldn't be an issue.

PR #7411, opened 2021-06-28
Simplify use_self impl

camsteffen:
changelog: none

Mainly to remove an extra visit and simplify the "in body?" logic.

### DIFF
--- a/clippy_lints/src/use_self.rs
+++ b/clippy_lints/src/use_self.rs
@@ -87,59 +87,42 @@ const SEGMENTS_MSG: &str = "segments should be composed of at least 1 element";
 
 impl<'tcx> LateLintPass<'tcx> for UseSelf {
     fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if !is_item_interesting(item) {
+            // This does two things:
+            //  1) Reduce needless churn on `self.stack`
+            //  2) Don't push `StackItem::NoCheck` when entering `ItemKind::OpaqueTy`,
+            //     in order to lint `foo() -> impl <..>`
+            return;
+        }
         // We push the self types of `impl`s on a stack here. Only the top type on the stack is
         // relevant for linting, since this is the self type of the `impl` we're currently in. To
         // avoid linting on nested items, we push `StackItem::NoCheck` on the stack to signal, that
         // we're in an `impl` or nested item, that we don't want to lint
-        //
-        // NB: If you push something on the stack in this method, remember to also pop it in the
-        // `check_item_post` method.
-        match &item.kind {
-            ItemKind::Impl(Impl {
-                self_ty: hir_self_ty,
-                of_trait,
-                ..
-            }) => {
-                let should_check = if let TyKind::Path(QPath::Resolved(_, item_path)) = hir_self_ty.kind {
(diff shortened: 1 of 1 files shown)


clippy_lints/src/use_self.rs
```diff
@@ -220,15 +217,15 @@ impl<'tcx> LateLintPass<'tcx> for UseSelf {
             if !in_macro(hir_ty.span);
             if meets_msrv(self.msrv.as_ref(), &msrvs::TYPE_ALIAS_ENUM_VARIANTS);
             if let Some(StackItem::Check {
-                hir_id,
+                impl_id,
                 types_to_lint,
                 types_to_skip,
                 ..
```
flip1995, line 208: The `..` shouldn't be necessary anymore.

clippy_lints/src/use_self.rs
```diff
@@ -182,52 +182,40 @@ impl<'tcx> LateLintPass<'tcx> for UseSelf {
...
-        if let Some(StackItem::Check {
-            impl_id,
-            types_to_lint,
-            types_to_skip,
-            ..
-        }) = self.stack.last_mut()
-        {
-            let self_ty = cx.tcx.type_of(*impl_id);
+        if let Some(&mut StackItem::Check { ref mut in_body, .. }) = self.stack.last_mut() {
+            *in_body += 1;
```
flip1995, line 190: To be save, I always prefer using `saturating_{sub,add}` in those things. This should not overflow, since `2^32-1` nested bodies is unrealistic, and it also shouldn't underflow. But I would prefer the `saturating_{sub,add}` method anyway, just to be save and so that we'll never have to deal with weird overflow errors in the future.

clippy_lints/src/use_self.rs
```diff
@@ -359,36 +299,11 @@ fn lint_path_to_variant(cx: &LateContext<'_>, path: &Path<'_>) {
...
-        if let TyKind::Path(QPath::Resolved(_, path)) = hir_ty.kind;
-        then {
-            !matches!(path.res, Res::SelfTy(..) | Res::Def(DefKind::TyParam, _))
-        } else {
-            false
-        }
-    }
+fn is_item_interesting(item: &Item<'_>) -> bool {
+    #[allow(clippy::enum_glob_use)]
+    use rustc_hir::ItemKind::*;
```
flip1995, line 304: I would keep importing those by name.

```suggestion
    use rustc_hir::ItemKind::{Const, Enum, Fn, Impl, Static, Struct, Trait, Union};
```

flip1995 (commented):
Is `type_of` a new query or did I just totally miss that when I implemented this? :thinking: 

Changes LGTM overall, so small things I noticed.

flip1995:
r? @flip1995 

@bors try (Just to make sure that this doesn't ICE. I'm really wary about the `in_body` handling, since this made the most problems for me when implementing this)

I'll also run the lintcheck tool if the try build succeeds.

camsteffen:
Comments are addressed!

> Is `type_of` a new query or did I just totally miss that when I implemented this? 🤔

I think it's not new? I randomly discovered one time that it can be used on impl blocks - certainly not very obvious.

flip1995:
@bors r+

Thanks! (Also passed my lintcheck run locally)

PR #7437, opened 2021-07-05
suggest `&mut` for redundant FnMut closures

### LINKED ISSUE #6903: Redundant_closure false positive with moved FnMut
Lint name: redundant_closure

I tried this code (reduced to a minimal example):

```rust
fn fp(x: Vec<i32>, y: Vec<i32>) -> Vec<i32> {
    let mut res = Vec::new();
    let mut add_to_res = |n| res.push(n);
    x.into_iter().for_each(|x| add_to_res(x));
    y.into_iter().for_each(add_to_res);
    res
}
```

If I apply the suggestion (redundant_closure on the fourth line), resulting in the following code:
```rust
fn fp(x: Vec<i32>, y: Vec<i32>) -> Vec<i32> {
    let mut res = Vec::new();
    let mut add_to_res = |n| res.push(n);
    x.into_iter().for_each(add_to_res);
    y.into_iter().for_each(add_to_res);
    res
}
```

I get the following error:

```
error[E0382]: use of moved value: `add_to_res`
 --> src/lib.rs:7:28
  |
6 |     x.into_iter().for_each(add_to_res);
  |                            ---------- value moved here
7 |     y.into_iter().for_each(add_to_res);
  |                            ^^^^^^^^^^ value used here after move
  |
note: closure cannot be moved more than once as it is not `Copy` due to moving the variable `res` out of its environment
 --> src/lib.rs:5:30
  |
5 |     let mut add_to_res = |n| res.push(n);
  |                              ^^^
```

### Meta

- `cargo clippy -V`: e.g. clippy 0.0.212 (98d6634 2020-11-14)
- `rustc -Vv`:
  ```
  rustc 1.50.0-nightly (98d66340d 2020-11-14)
  binary: rustc
  commmit-hash: 98d66340d6e63eda115afc8b0da1d87965881936
  commit-date: 2020-11-14
  release: 1.50.0-nightly
  ```

ebobrow:
fixes #6903

changelog: suggest `&mut` for redundant FnMut closures

### DIFF
--- a/clippy_lints/src/eta_reduction.rs
+++ b/clippy_lints/src/eta_reduction.rs
@@ -1,15 +1,16 @@
 use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
-use clippy_utils::higher;
 use clippy_utils::higher::VecArgs;
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::{implements_trait, type_is_unsafe_function};
+use clippy_utils::usage::UsedAfterExprVisitor;
+use clippy_utils::{get_enclosing_loop_or_closure, higher};
 use clippy_utils::{is_adjusted, iter_input_pats};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::{def_id, Expr, ExprKind, Param, PatKind, QPath};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
-use rustc_middle::ty::{self, Ty};
+use rustc_middle::ty::{self, ClosureKind, Ty};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
 declare_clippy_lint! {
@@ -86,7 +87,7 @@ impl<'tcx> LateLintPass<'tcx> for EtaReduction {
     }
 }
 
-fn check_closure(cx: &LateContext<'_>, expr: &Expr<'_>) {
+fn check_closure<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
     if let ExprKind::Closure(_, decl, eid, _, _) = expr.kind {
         let body = cx.tcx.hir().body(eid);
         let ex = &body.value;
@@ -131,7 +132,18 @@ fn check_closure(cx: &LateContext<'_>, expr: &Expr<'_>) {
 
             then {
                 span_lint_and_then(cx, REDUNDANT_CLOSURE, expr.span, "redundant closure", |diag| {
-                    if let Some(snippet) = snippet_opt(cx, caller.span) {
+                    if let Some(mut snippet) = snippet_opt(cx, caller.span) {
+                        if_chain! {
+                            if let ty::Closure(_, substs) = fn_ty.kind();
+                            if let ClosureKind::FnMut = substs.as_closure().kind();
+                            if UsedAfterExprVisitor::is_found(cx, caller)
+                                || get_enclosing_loop_or_closure(cx.tcx, expr).is_some();
+
+                            then {
+                                // Mutable closure is used after current expr; we cannot consume it.
+                                snippet = format!("&mut {}", snippet);
+                            }
+                        }
                         diag.span_suggestion(
                             expr.span,
                             "replace the closure with the function itself",
--- a/clippy_utils/src/usage.rs
+++ b/clippy_utils/src/usage.rs
@@ -199,3 +199,50 @@ pub fn contains_return_break_continue_macro(expression: &Expr<'_>) -> bool {
     recursive_visitor.visit_expr(expression);
     recursive_visitor.seen_return_break_continue
 }
+
+pub struct UsedAfterExprVisitor<'a, 'tcx> {
+    cx: &'a LateContext<'tcx>,
+    expr: &'tcx Expr<'tcx>,
+    definition: HirId,
+    past_expr: bool,
+    used_after_expr: bool,
+}
+impl<'a, 'tcx> UsedAfterExprVisitor<'a, 'tcx> {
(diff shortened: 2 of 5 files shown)


clippy_lints/src/eta_reduction.rs
```diff
@@ -131,7 +132,22 @@ fn check_closure(cx: &LateContext<'_>, expr: &Expr<'_>) {
 
             then {
                 span_lint_and_then(cx, REDUNDANT_CLOSURE, expr.span, "redundant closure", |diag| {
-                    if let Some(snippet) = snippet_opt(cx, caller.span) {
+                    if let Some(mut snippet) = snippet_opt(cx, caller.span) {
+                        if_chain! {
+                            if let ty::Closure(_, substs) = fn_ty.kind();
+                            if let ClosureKind::FnMut = substs.as_closure().kind();
+                            if let Some(closure_def) = path_to_local(caller);
+                            if let Some(scope) = cx.tcx.hir().get_enclosing_scope(closure_def);
```
flip1995, line 140: We have the utility function `get_enclosing_block` for this.

clippy_lints/src/eta_reduction.rs
```diff
@@ -174,6 +190,42 @@ fn check_closure(cx: &LateContext<'_>, expr: &Expr<'_>) {
     }
 }
 
+fn used_after_expr<'tcx>(mut visitor: LocalUsedVisitor<'tcx>, block: &Block<'tcx>, expr: &Expr<'_>) -> bool {
```
flip1995, line 193: This impl looks wrong/unnecessary to me. Staring at this for a while, I think I figured out why:

1. You use spans to check where an expression is. This will probably fail in some cases, for example, when macros are involved. 
2. The loop detection seems too complicated to me. This is the HIR, so you shouldn't have to differentiate between loop kinds.
3. We have a function  `clippy_utils::get_enclosing_loop_or_closure` you can just use to check if the closure is called in a context where `&mut` is necessary.
4. A "used after `expr`" sounds like something we're already doing in other places. Looking for it, I only found lint specific implementations though. But I think that this is a good opportunity to implement a visitor for this in `clippy_utils::usage`. With a visitor you can check if you already passed an expression like this: https://github.com/rust-lang/rust-clippy/blob/64d74df19c582a10ad41ddce496528e597b06765/clippy_lints/src/loops/utils.rs#L175-L176 This should be implemented as a visitor.

flip1995 (changes requested):
I don't think this is sufficient to fix all cases. I made some suggestion on how to improve this check.

clippy_lints/src/eta_reduction.rs
ebobrow: Thanks for your help! Can you point me towards the other lints that implement a "used after `expr`" so that I can switch to the visitor there, or should that be a separate PR?

flip1995: If at all this should be a separate PR. This is mostly done in other visitors that are pretty specific to the lints they're used in, so I'm not sure if it's worth splitting that out.

clippy_utils/src/usage.rs
```diff
@@ -199,3 +199,56 @@ pub fn contains_return_break_continue_macro(expression: &Expr<'_>) -> bool {
...
+            return;
+        }
+
+        if expr.hir_id == self.expr.hir_id {
+            self.past_self = true;
+
+            if let Some(Expr {
+                kind: ExprKind::Loop(..),
+                ..
+            }) = utils::get_enclosing_loop_or_closure(self.cx.tcx, expr)
```
flip1995, line 244: I don't think this belongs in this visitor. If loops are a concern, then this function should be called explicitly, instead of hiding it away in this visitor.

Also you don't have to differentiate between loops and closures.

flip1995 (commented):
One more note.

ebobrow:
Thanks for the review! I think this should be better now. Sorry, I'm not sure how to change the label back to `waiting-on-review`.

clippy_utils/src/usage.rs
```diff
@@ -199,3 +199,50 @@ pub fn contains_return_break_continue_macro(expression: &Expr<'_>) -> bool {
     recursive_visitor.visit_expr(expression);
     recursive_visitor.seen_return_break_continue
 }
+
+pub struct UsedAfterExprVisitor<'a, 'tcx> {
+    cx: &'a LateContext<'tcx>,
+    expr: &'tcx Expr<'tcx>,
+    definition: HirId,
+    past_self: bool,
+    used_after_self: bool,
```
flip1995, line 208: NIT: I would rename those to 
```suggestion
    past_expr: bool,
    used_after_expr: bool,
```

`self` is confusing to me.

tests/ui/eta.rs
```diff
@@ -220,3 +220,18 @@ impl std::ops::Deref for Bar {
 fn test_deref_with_trait_method() {
     let _ = [Bar].iter().map(|s| s.to_string()).collect::<Vec<_>>();
 }
+
+fn mutable_closure_used_twice(x: Vec<i32>, y: Vec<i32>) {
+    let mut res = Vec::new();
+    let mut add_to_res = |n| res.push(n);
+    x.into_iter().for_each(|x| add_to_res(x));
+    y.into_iter().for_each(|x| add_to_res(x));
```
flip1995, line 228: Can you add this line a third time? I expect that your changes will just work here and suggest to use `&mut` for the first 2 lines, but having this as a regression test will help in the future.

flip1995:
No problem, it's 

@rustbot label -S-waiting-on-author +S-waiting-on-review

flip1995 (commented):
2 small things left. Those should really be the last before we can merge this.

flip1995 (approved):

flip1995:
@bors r+

Thanks!

Issue #7433, opened 2021-07-04
`panic` lints for `assert` with formatting


iliakonnov:
**Lint name**: `panic`

**I tried this code:**

```rust
#![warn(clippy::panic)]
fn main() {
    assert!(true, "True is not true!");
}
```

**I expected to see this happen**: No warning

**Instead, this happened**: `panic` should not be present in production code

For example, following code does not produce this warning:
```rust
#![warn(clippy::panic)]
fn main() {
    assert!(true);
}
```

I believe that behaviour should be the same, regardless of optional message existence. Both the warn on `assert!` and no warning will be reasonable, but the latter looks better for me.

This issue looks very similar to https://github.com/rust-lang/rust-clippy/issues/7062.

### Meta

- `cargo clippy -V`: clippy 0.1.55 (71b8742 2021-07-03)
- `rustc -Vv`:
  ```
  rustc 1.55.0-nightly (71b8742bb 2021-07-03)
  binary: rustc
  commit-hash: 71b8742bbcbed2cd908dbc031d6552d8b528c037
  commit-date: 2021-07-03
  host: x86_64-unknown-linux-gnu
  release: 1.55.0-nightly
  LLVM version: 12.0.1
  ```

PR #7439, opened 2021-07-06
Fix false-positive `assert` in `panic`

iobtl:
This PR fixes a false-positive in `clippy::panic` when using the `assert` macro with its optional message parameter.

Fixes: #7433 

changelog: `panic_unimplemented.rs`: added condition to exclude `assert` macro, similar to `debug_assert`
changelog: `panicking_macros.rs`: relevant tests to check for `assert` usage.


flip1995:
@bors r+

Thanks!

PR #7442, opened 2021-07-07
Refactor `format_args!` expansion parsing

camsteffen:
Motivated by rust-lang/rust#83302, so I only have to change Clippy in one place. Fixed an FP along the way.

I also allowed `needless_bool` in macros because I often want to do `if_chain! { .. then { true } else { false } }`.

changelog: Fix false positive in `useless_format` when some text is appended or prepended to a single string with some useless formatting params
changelog: Allow `needless_bool` in macros


xFrednet:
Sure, I'm happy to take them over! Could you maybe also set me as the assignee? 🙃

xFrednet:
^^ thank you! I believe that @flip1995  would also like to stay assigned to double-check my review later. 🙃

clippy_lints/src/format.rs
```diff
@@ -44,130 +43,79 @@ declare_lint_pass!(UselessFormat => [USELESS_FORMAT]);
...
     }
 }
 
-fn span_useless_format<T: LintContext>(cx: &T, span: Span, help: &str, mut sugg: String) {
-    let to_replace = span.source_callsite();
-
+fn span_useless_format(cx: &LateContext<'_>, span: Span, mut sugg: String, applicability: Applicability) {
     // The callsite span contains the statement semicolon for some reason.
-    let snippet = snippet(cx, to_replace, "..");
+    let snippet = snippet(cx, span, "..");
```
xFrednet, line 95: Could you use `snippet_with_applicability` here

xFrednet (changes requested):
Hey, awesome work, it took some time digging though it all but the result is worth it. Thank you for the nice refactoring!

`clippy_lints/src/explicit_write.rs` still contains some lint messages with punctuation and a suggestion in the main message (lines 111ff). Could you update the lint emission to display the help message using `span_lint_and_help`?

camsteffen:
Thanks for the thorough review @xFrednet. I addressed your comments.

xFrednet (approved):
Awesome, this looks good to me :+1: 

Hey @flip1995, I'm done with my review and would hand this over to you now. If you weren't taking over, I would ask `@camsteffen` to squash his fixup commits before merging it. :upside_down_face:

flip1995 (approved):
@xFrednet Thanks for doing the review!

@camsteffen please squash the fixup commits and then r=xFrednet :+1:

PR #7446, opened 2021-07-08
`default_numeric_fallback`: Fix FP with floating literal

### LINKED ISSUE #7445: FP: `default_numeric_fallback` with floating literal
Lint name: `default_numeric_fallback`

I tried this code:

```rust
#![deny(clippy::default_numeric_fallback)]

fn foo(_f: f64) {}

fn main() {
    foo(12.0);
}
```

I expected to see this happen: no lint triggered.

Instead, this happened: 
```
error: default numeric fallback might occur
 --> src/main.rs:6:9
  |
6 |     foo(12.0);
  |         ^^^^ help: consider adding suffix: `12.0_f64`
  |
```

### Meta

- `cargo clippy -V`: clippy 0.1.55 (d2b04f0 2021-07-07)
- `rustc -Vv`: 
  ```
  rustc 1.55.0-nightly (d2b04f075 2021-07-07)
  binary: rustc
  commit-hash: d2b04f075c0ce010758c4c8674152ff89d1d73f3
  commit-date: 2021-07-07
  release: 1.55.0-nightly
  LLVM version: 12.0.1
  ```

Y-Nak:
Fix #7445 

changelog: `default_numeric_fallback`: Fix FP with floating literal

### DIFF
--- a/clippy_lints/src/default_numeric_fallback.rs
+++ b/clippy_lints/src/default_numeric_fallback.rs
@@ -1,5 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::source::snippet;
+use clippy_utils::numeric_literal;
+use clippy_utils::source::snippet_opt;
 use if_chain::if_chain;
 use rustc_ast::ast::{LitFloatType, LitIntType, LitKind};
 use rustc_errors::Applicability;
@@ -80,14 +81,23 @@ impl<'a, 'tcx> NumericFallbackVisitor<'a, 'tcx> {
                             LitKind::Int(_, LitIntType::Unsuffixed) | LitKind::Float(_, LitFloatType::Unsuffixed));
                 if !ty_bound.is_numeric();
                 then {
-                    let suffix = match lit_ty.kind() {
-                        ty::Int(IntTy::I32) => "i32",
-                        ty::Float(FloatTy::F64) => "f64",
+                    let (suffix, is_float) = match lit_ty.kind() {
+                        ty::Int(IntTy::I32) => ("i32", false),
+                        ty::Float(FloatTy::F64) => ("f64", true),
                         // Default numeric fallback never results in other types.
                         _ => return,
                     };
 
-                    let sugg = format!("{}_{}", snippet(self.cx, lit.span, ""), suffix);
+                    let src = if let Some(src) = snippet_opt(self.cx, lit.span) {
+                        src
+                    } else {
+                        match lit.node {
+                            LitKind::Int(src, _) => format!("{}", src),
+                            LitKind::Float(src, _) => format!("{}", src),
+                            _ => return,
+                        }
+                    };
+                    let sugg = numeric_literal::format(&src, Some(suffix), is_float);
                     span_lint_and_sugg(
                         self.cx,
                         DEFAULT_NUMERIC_FALLBACK,
--- a/clippy_lints/src/default_numeric_fallback.rs
+++ b/clippy_lints/src/default_numeric_fallback.rs
@@ -78,7 +78,7 @@ impl<'a, 'tcx> NumericFallbackVisitor<'a, 'tcx> {
                 if let Some(ty_bound) = self.ty_bounds.last();
                 if matches!(lit.node,
                             LitKind::Int(_, LitIntType::Unsuffixed) | LitKind::Float(_, LitFloatType::Unsuffixed));
-                if !ty_bound.is_integral();
+                if !ty_bound.is_numeric();
                 then {
                     let suffix = match lit_ty.kind() {
                         ty::Int(IntTy::I32) => "i32",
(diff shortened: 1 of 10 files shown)


xFrednet (commented):
The changes LGTM.

The FP was a result of tests which only checked for the int type. I've looked through the test file, and it seems to me that the first test cases still miss a float counterpart. Could you also add them as a precaution?

```rs
// in the `basic_expr` module
        let x = [1.0, 2.0, 3.0];
        let x = if true { (1.0, 2.0) } else { (3.0, 4.0) };
        let x = match 1.0 {
            1.0 => 1.0,
            _ => 2.0,
        };
 // in the `nested_local` module
         let x: _ = {
            // Should lint this because this literal is not bound to any types.
            let y = 1.0;

            // Should NOT lint this because this literal is bound to `_` of outer `Local`.
            1.0
        };
```

Y-Nak:
Changes:
1. Add more test cases for floating literal
2. Split the i32 and f64 test cases into their own files, otherwise `cargo dev limit_stderr_length` fails

tests/ui/default_numeric_fallback_f64.rs
```diff
@@ -0,0 +1,172 @@
...
+
+mod basic_expr {
+    fn test() {
+        // Should lint unsuffixed literals typed `f64`.
+        let x = 0.12;
+        let x = [1., 2., 3.];
+        let x = if true { (1., 2.) } else { (3., 4.) };
+        let x = match 1. {
+            _ => 1.,
+        };
```
xFrednet, line 21: NIT: This triggers the `clippy::match_single_binding` lint. Could you allow this lint to keep the file error files to only contain `default_numeric_fallback` lint triggers?

```txt
error: this match could be replaced by its body itself
  --> $DIR/default_numeric_fallback_f64.rs:19:17
   |
LL |           let x = match 1. {
   |  _________________^
LL | |             _ => 1.,
LL | |         };
   | |_________^ help: consider using the match body instead: `1.`
   |
   = note: `-D clippy::match-single-binding` implied by `-D warnings`
```

Y-Nak: Good catch! I fixed that.

xFrednet (approved):
LGTM, thank you for the quick fixes and nice work!

@flip1995 I would hand it over to you now :)

tests/ui/default_numeric_fallback_f64.rs
```diff
@@ -0,0 +1,174 @@
+// aux-build:macro_rules.rs
+
+#![warn(clippy::default_numeric_fallback)]
+#![allow(unused)]
+#![allow(clippy::never_loop)]
+#![allow(clippy::no_effect)]
+#![allow(clippy::unnecessary_operation)]
+#![allow(clippy::branches_sharing_code)]
+#![allow(clippy::branches_sharing_code)]
```
flip1995, line 9: dup

tests/ui/default_numeric_fallback_f64.rs
```diff
@@ -0,0 +1,174 @@
+// aux-build:macro_rules.rs
```
flip1995, line 1: Is there a reason, why neither the floating point, nor the integer test are tested with `// run-rustfix`? Looking at the `stderr` files this should be possible.

Could you please test if adding `// run-rustfix` Just Works(tm)?

Y-Nak: I just forgot to add them.

clippy_lints/src/default_numeric_fallback.rs
```diff
@@ -80,14 +81,23 @@ impl<'a, 'tcx> NumericFallbackVisitor<'a, 'tcx> {
...
 
-                    let sugg = format!("{}_{}", snippet(self.cx, lit.span, ""), suffix);
+                    let src = if let Some(src) = snippet_opt(self.cx, lit.span){
+                        src
+                    } else {
+                        match lit.node  {
+                            LitKind::Int(src, _) =>format!("{}", src),
+                            LitKind::Float(src, _) => format!("{}", src),
+                            _ => { return; }
+                        }
```
flip1995, line 98: Formattings:

```suggestion
                    let src = if let Some(src) = snippet_opt(self.cx, lit.span) {
                        src
                    } else {
                        match lit.node {
                            LitKind::Int(src, _) => format!("{}", src),
                            LitKind::Float(src, _) => format!("{}", src),
                            _ => return,
                        }
```

Y-Nak:
Changes:
* Fix `NumericLiteral::format`, there was a possibility of producing invalid literals like `1._f32`
*  Remove duplication of `#![allow(clippy::branches_sharing_code)]`
* Add `// run-rustfix`

flip1995 (commented):
One more NIT

clippy_lints/src/default_numeric_fallback.rs
Y-Nak: Just a question:
Could you tell me how do you find formatting problems inside `if_chain`? Is there a good way to find them inside macros?

flip1995: I don't know of anything that would help here. Maybe implementing formatting rules in `rustfmt` for the if_chain macro. Not sure if that would be accepted there though, since this isn't a std macro.

flip1995:
@bors r=xFrednet,flip1995

Thanks!

clippy_lints/src/default_numeric_fallback.rs
Y-Nak: I understand. Thanks for the answer!

PR #7448, opened 2021-07-08
Rename run_lints -> lints_enabled

flip1995:
Just a quick rename of a utilities function. `run_lints` kinda suggested that the lints were run by this function. But the only thing this function does is to check if the lints are enabled in the context of the `hir_id`

changelog: none

### DIFF
--- a/clippy_lints/src/cargo_common_metadata.rs
+++ b/clippy_lints/src/cargo_common_metadata.rs
@@ -3,7 +3,7 @@
 use std::path::PathBuf;
 
 use clippy_utils::diagnostics::span_lint;
-use clippy_utils::run_lints;
+use clippy_utils::lints_enabled;
 use rustc_hir::{hir_id::CRATE_HIR_ID, Crate};
 use rustc_lint::{LateContext, LateLintPass};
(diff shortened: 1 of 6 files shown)


flip1995:
rust-highfive bot was down because of a invalid json file https://github.com/rust-lang/highfive/pull/343

camsteffen:
Can we just remove it and use `is_allowed` instead? We're always passing a single `LintId` anyways. Maybe rename to `is_lint_allowed`. I've always though `is_allowed` is too vague. But I think I'd prefer "allowed" to the opposite "enabled" because it refers to a specific `Level`. "enabled" doesn't really correspond to a rustc concept. Also `run_lints` is missing `Level::ForceWarn`.

camsteffen:
One more thought. We could have a variant that uses `LateContext::last_node_with_lint_attrs`, similar to `span_lint`. So we could have `is_lint_allowed` and `is_lint_allowed_at`.

flip1995:
> Can we just remove it and use `is_allowed` instead?

Oh I missed `is_allowed` when I added the `run_lints` function. I agree with your analysis and will rename `is_allowed` to `is_lint_allowed` and remove `lints_enabled`.

PR #7449, opened 2021-07-09
Remove lints_enabled

flip1995:
r? @camsteffen 

cc https://github.com/rust-lang/rust-clippy/pull/7448#issuecomment-876497862

I haven't added a variant with `last_node_with_lint_attrs` yet, since I didn't see a usecase for this. Also this field is not documented, so I'm wondering how it behaves with command line lints and so on.

changelog: none

### DIFF
--- a/clippy_lints/src/cargo_common_metadata.rs
+++ b/clippy_lints/src/cargo_common_metadata.rs
@@ -2,7 +2,7 @@
 
 use std::path::PathBuf;
 
-use clippy_utils::{diagnostics::span_lint, is_allowed};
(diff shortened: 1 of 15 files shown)


camsteffen:
Whenever we `span_lint_*` without a `HirId` (`span_lint_hir`), `last_node_with_lint_attrs` is [used](https://github.com/rust-lang/rust/blob/ac8c3bfffb2829d0af075a5d4020fb7e9e86c3a2/compiler/rustc_lint/src/context.rs#L802) to set the appropriate lint level (and skip if the lint is allowed).

PR #7450, opened 2021-07-09
Fix syntax highlighting for no_run blocks

frogtd:
I was reading the lint documentation and I noticed that some of the lints weren't getting syntax highlighting.

changelog: Fix syntax highlighting for no_run blocks.

### DIFF
--- a/util/export.py
+++ b/util/export.py
@@ -23,7 +23,7 @@ def parse_code_block(match):
 
     for line in match.group(0).split('\n'):
         # fix syntax highlighting for headers like ```rust,ignore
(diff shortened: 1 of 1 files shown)


camsteffen:
We're in the process of migrating to a new "metadata collector" (#7298) where this issue is fixed ([code](https://github.com/rust-lang/rust-clippy/blob/e405c68b3c1265daa9a091ed9b4b5c5a38c0c0ba/clippy_lints/src/utils/internal_lints/metadata_collector.rs#L505)) and the python script will be obsolete. Thanks for the fix but I think it's better to just wait for #7298 at this point.

PR #7458, opened 2021-07-13
Add fixed example to unused_unit documentation

flip1995:
changelog: none

(don't think this is worth a changelog mention)


giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #7369, opened 2021-06-17
branches_sharing_code incorrectly detects branches using the same code


rockstar:
Lint name: branches_sharing_code

I upgrade to rust 1.53 and tried this code:

```rust
    fn format_node_with_parens(&mut self, node: &Node) {
        if has_parens(node) {
            // If the AST already has parens here do not double add them
            self.format_node(node);
        } else {
            self.write_rune('(');
            self.format_node(node);
            self.write_rune(')')
        }
    }
```

I expected clippy to be fine with this code.

Instead, this this code was linted incorrectly with the following:

```
error: all if blocks contain the same code at the end
   --> flux-core/src/formatter/mod.rs:799:9
    |
799 | /             self.format_node(node);
800 | |             self.write_rune(')')
801 | |         }
    | |_________^
    |
note: the lint level is defined here
   --> flux-core/src/lib.rs:1:38
    |
1   | #![cfg_attr(feature = "strict", deny(warnings, missing_docs))]
    |                                      ^^^^^^^^
    = note: `#[deny(clippy::branches_sharing_code)]` implied by `#[deny(warnings)]`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#branches_sharing_code
help: consider moving the end statements out like this
    |
799 |         }
800 |         self.format_node(node);
801 |         self.write_rune(')')
    |

error: all if blocks contain the same code at the end
   --> flux-core/src/formatter/mod.rs:799:9
    |
799 | /             self.format_node(node);
800 | |             self.write_rune(')')
801 | |         }
    | |_________^
    |
note: the lint level is defined here
   --> flux-core/src/lib.rs:1:38
    |
1   | #![cfg_attr(feature = "strict", deny(warnings, missing_docs))]
    |                                      ^^^^^^^^
    = note: `#[deny(clippy::branches_sharing_code)]` implied by `#[deny(warnings)]`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#branches_sharing_code
help: consider moving the end statements out like this
    |
799 |         }
800 |         self.format_node(node);
801 |         self.write_rune(')')
    |
```

The suggested code changes what the code actually does, rather than just finding a better way to express them. In this case, if we made the change, we'd change the functionality and (thankfully) break tests asserting specific behavior.

### Meta

- `cargo clippy -V`: `clippy 0.1.53 (53cb7b09 2021-06-17)`
- `rustc -Vv`:
```
[...]

Issue #7452, opened 2021-07-10
branches_sharing_code reporting a false positive


MGlolenstine:
I tried this code:

```rust
fn test() {
    let v = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 0];
    let mut counter = 0;
    for i in v {
        if counter == 0 {
            counter += 1;
            println!("first");
        } else if counter == 1 {
            counter += 1;
            println!("second");
        } else {
            counter += 1;
            println!("other: {}", i);
        }
    }
}

```

I expected to see this happen: I expected the lint to pass without a hitch, as the counter is used in the `if` statement.

Instead, this happened: 
I got a lint report, that I could change the if statements, as they all start with the same line of code, but doing that makes the code invalid.
```
warning: all if blocks contain the same code at the start
 --> src/lib.rs:5:9
  |
5 | /         if counter == 0 {
6 | |             counter += 1;
  | |_________________________^
  |
  = note: `#[warn(clippy::branches_sharing_code)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#branches_sharing_code
help: consider moving the start statements out like this
  |
5 |         counter += 1;
6 |         if counter == 0 {
  |
```

### Meta

- `cargo clippy -V`: clippy 0.1.53 (53cb7b0 2021-06-17)
- `rustc -Vv`:
```
rustc 1.53.0 (53cb7b09b 2021-06-17)
binary: rustc
commit-hash: 53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b
commit-date: 2021-06-17
host: x86_64-unknown-linux-gnu
release: 1.53.0
LLVM version: 12.0.1
```

MGlolenstine:
Suggesting label change to `I-suggestion-causes-error`, as it's correctly detected the case, but the fix causes a problem.

xFrednet:
This kind of fits both labels, I'm going over a possible solution in my head and might claim this issue :)

MGlolenstine:
@xFrednet Can this issue be closed now that the #7462 has been merged?
Thanks for the great work!

xFrednet:
#7462 only documented this as a known issue. It's possible to actually fix this issue to not trigger the lint at all. I would keep this issue open until that is done, or it's decided that it's not worth it. 

You're welcome, thank you for the report. :upside_down_face:

youknowone:
[...]

PR #7462, opened 2021-07-13
FP fix and documentation for `branches_sharing_code` lint

xFrednet:
Closes rust-lang/rust-clippy#7369

Related rust-lang/rust-clippy#7452 I'm still thinking about the best way to fix this. I could simply add another visitor to ensure that the moved expressions don't modify values being used in the condition, but I'm not totally happy with this due to the complexity. I therefore only documented it for now 

changelog: [`branches_sharing_code`] fixed false positive where block expressions would sometimes be ignored.

### DIFF
--- a/clippy_lints/src/copies.rs
+++ b/clippy_lints/src/copies.rs
@@ -120,7 +120,10 @@ declare_clippy_lint! {
     ///
     /// **Why is this bad?** Duplicate code is less maintainable.
     ///
-    /// **Known problems:** Hopefully none.
+    /// **Known problems:**
+    /// * The lint doesn't check if the moved expressions modify values that are beeing used in
+    ///   the if condition. The suggestion can in that case modify the behavior of the program.
+    ///   See [rust-clippy#7452](https://github.com/rust-lang/rust-clippy/issues/7452)
     ///
     /// **Example:**
     /// ```ignore
@@ -358,8 +361,7 @@ fn scan_block_for_eq(cx: &LateContext<'tcx>, blocks: &[&Block<'tcx>]) -> Option<
         expr_eq &= block_expr_eq;
     }
 
-    let has_expr = blocks[0].expr.is_some();
-    if has_expr && !expr_eq {
+    if !expr_eq {
         end_eq = 0;
     }
 
--- a/tests/ui/branches_sharing_code/false_positives.rs
+++ b/tests/ui/branches_sharing_code/false_positives.rs
@@ -1,0 +1,28 @@
+#![allow(dead_code)]
+#![deny(clippy::if_same_then_else, clippy::branches_sharing_code)]
+
+// ##################################
+// # Issue clippy#7369
+// ##################################
+#[derive(Debug)]
+pub struct FooBar {
+    foo: Vec<u32>,
+}
+
+impl FooBar {
+    pub fn bar(&mut self) {
+        if true {
+            self.foo.pop();
+        } else {
+            self.baz();
+
+            self.foo.pop();
+
+            self.baz()
+        }
+    }
+
+    fn baz(&mut self) {}
+}
+
+fn main() {}


tests/ui/branches_sharing_code/false_positives.rs
```diff
@@ -0,0 +1,48 @@
...
+            println!("first");
+        } else if counter == 1 {
+            counter += 1;
+            println!("second");
+        } else {
+            counter += 1;
+            println!("other: {}", i);
+        }
+    }
+}
```
xFrednet, line 46: This is the issue that is not fixed yet. I found it reasonable to add a test for it (even if it fails) as it's also documented in the lint description

clippy_lints/src/copies.rs
```diff
@@ -225,7 +228,7 @@ fn lint_same_then_else<'tcx>(
             blocks,
             expr,
         );
-    } else if end_eq != 0 || (has_expr && expr_eq) {
+    } else if (end_eq != 0 && !has_expr) || (has_expr && expr_eq) {
```
camsteffen, line 231: I think the real problem is that `scan_block_for_eq` can return `end_eq > 0 && expr_eq == false`? End statements should not be evaluated if the expressions are not equal.

tests/ui/branches_sharing_code/false_positives.rs
camsteffen: I'm not sure we gain anything by adding this. Tests are valuable when they verify correct behavior.

camsteffen:
For #7452, I would maybe first collect all the variables referenced in the condition. Then, when determining if statements are equal, if there is a mutable reference to the variable, the statements are not equal. So I think you would add `SpanlessEq::disallowed_mut_locals: FxHashSet<HirId>`.

tests/ui/branches_sharing_code/false_positives.rs
xFrednet: Yeah, I wasn't sure with this one that's, the reason why I pointed it out. I'll remove it from this PR and then squash the two commits :upside_down_face:

xFrednet:
> For #7452, I would maybe first collect all the variables referenced in the condition. Then, when determining if statements are equal, if there is a mutable reference to the variable, the statements are not equal. So I think you would add `SpanlessEq::disallowed_mut_locals: FxHashSet<HirId>`.

Interesting idea, I thought about expending the `UsedValueFinderVisitor` to also collect mutations. Putting it into `SpanlessEq` feels like the wrong place, as it's not really part of the comparison in my eyes, and I feel like `SpanlessEq` is already quite complex :thinking:

clippy_lints/src/copies.rs
xFrednet: You're right. `scan_block_for_eq` even contained an if block for this purpose, but the condition was not quite correct, it's fixed in the latest version. :upside_down_face:

xFrednet:
The FP is now fixed in `scan_block_for_eq` where it originated from (very nice catch!!). I've squashed the commits right away, having a fix up commit that just reverted the previous implementation and the FP test seemed a bit unnecessary. I hope that is okay for you :upside_down_face:

camsteffen:
> Interesting idea, I thought about expending the `UsedValueFinderVisitor` to also collect mutations. Putting it into `SpanlessEq` feels like the wrong place, as it's not really part of the comparison in my eyes, and I feel like `SpanlessEq` is already quite complex 🤔

Yeah I know what you mean. That would work. One more idea for `SpanlessEq` - `expr_fallback` could have a counterpart `expr_precondition`.

camsteffen:
Looks good, thanks!

@bors r+

xFrednet:
> Yeah I know what you mean. That would work. One more idea for `SpanlessEq` - `expr_fallback` could have a counterpart `expr_precondition`.

`SpanlessEq` - `expr_fallback` doesn't ring a bell for me right now, but it also has been some time since I’ve dug through the code :thinking: Thank you for the suggestions :)

Issue #5393, opened 2020-03-30
Move our paths over to diagnostic items


Manishearth:
Currently we have [a ton of hardcoded paths](https://github.com/rust-lang/rust-clippy/blob/master/clippy_utils/src/paths.rs) for stdlib types.

Rustc has a way of tagging types for diagnostics via a "diagnostic item" API (https://github.com/rust-lang/rust/pull/60966/). We should go through rustc and tag most of these diagnostic items, and replace the hardcoded paths with hardcoded diagnostic item names instead. We can still keep the path framework around for rapid prototyping, but have a comment recommending people switch to diagnostic items whenever possible.

flip1995:
Marking it as tracking issue and pinning it, since this is probably done in multiple steps.

lzutao:
Could we use lang_items if possible ?

flip1995:
If there is a `LangItem` we should use it. But this issue is rather about adding `#[rustc_diagnostic_item(..)]` in rustc and then use it in Clippy. We can do this easier now that Clippy is a subtree in rustc.

xFrednet:
Heyho, #7466 updates Clippy to use diagnostic items where currently possible. There are still a bunch of paths which don't have a corresponding diagnostic item yet. (See: [`paths.rs`](https://github.com/rust-lang/rust-clippy/blob/ecf85f4bdc319f9d9d853d1fff68a8a25e64c7a8/clippy_utils/src/paths.rs) after the update).

I would be happy to go through rustc and add the missing diagnostic items. Is this still wanted as stated in the lint description? And is there any documentation regarding the diagnostic items? The compiler currently has a mixture of naming conventions, and it sometimes uses `#[cfg_attr(not(test), rustc_diagnostic_item = "...")]` instead of `rustc_diagnostic_item` which got me a little confused. Any guidance would be greatly appreciated :upside_down_face:

Manishearth:
@xFrednet yes, please add them. I don't remember why the compiler does that though, perhaps ask on one of the compiler zulip channels

clarfonthey:
Just commenting here that the issue description links the old path to `paths.rs`, which does not exist on the latest commit.

Comparing:

* Listed: https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/utils/paths.rs
* Actual: https://github.com/rust-lang/rust-clippy/blob/master/clippy_utils/src/paths.rs

xFrednet:
[...]

PR #7466, opened 2021-07-14
Use diagnostic items where possible

xFrednet:
Clippy still uses a bunch of paths in places that could easily use already defined diagnostic items. This PR updates all references to such paths and also removes a bunch of them that are no longer needed after this cleanup.

Some paths are also used to construct new paths and can therefore not be removed that easily. I've added a doc comment to those instances that recommends the use of the diagnostic item where possible.

And that's it, cleaning crew signing off :broom: :wastebasket: 

---

changelog: none

(only internal improvements)

cc: #5393


flip1995:
@bors r+

Thanks! (For some reason I totally missed that this PR was opened)

PR #7469, opened 2021-07-15
Add new lint `macros_hiding_unsafe_code`

### LINKED ISSUE #7323: Macros wrapping expressions in `unsafe` blocks when using `unsafe_op_in_unsafe_fn`
### What it does

For projects using `unsafe_op_in_unsafe_fn` (and specially those requiring `// SAFETY` comments), it would be nice to detect cases where a macro is wrapping an expression passed as a parameter in a "hidden" `unsafe` block, and thus bypassing the intention of `unsafe_op_in_unsafe_fn`.

The logic could be something like:
  - If inside an `unsafe fn`...
  - And there is a macro call outside an `unsafe` block...
  - And `unsafe_op_in_unsafe_fn` is enabled in this context...
  - And if the node/AST/tree of a parameter appears as-is...
  - Then check that node/AST/tree is not inside an introduced `unsafe` block.

Callers of the macro would need to wrap the entire macro call in an `unsafe` block to avoid hitting the lint (plus perhaps playing with `#[allow(unused_unsafe)]` inside the macro to avoid the unneeded nested `unsafe` warning from `rustc`).

### Categories (optional)

- Kind: `clippy::pedantic`?

### Drawbacks

None.

### Example

Assume `#![deny(unsafe_op_in_unsafe_fn)]`, and that `container_of!` wraps the first parameter in an `unsafe` block. Then:

```rust
let reg = container_of!((*file).private_data, Self, mdev);
```

Could be written as:

```rust
let reg = unsafe { container_of!((*file).private_data, Self, mdev) };
```

flip1995:
This lint only triggers if the `unsafe_op_in_unsafe_fn` lint is enabled.

This lint is placed in the restriction group, since it suggests code,
that requires to allow the `unused_unsafe` lint locally. Instead of
allowing this lint, the unsafe block in the macro could be removed.

Closes #7323 

changelog: Add new lint [`macros_hiding_unsafe_code`]

### DIFF
--- a/clippy_lints/src/macros_hiding_unsafe_code.rs
+++ b/clippy_lints/src/macros_hiding_unsafe_code.rs
@@ -1,0 +1,140 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, in_macro, is_lint_allowed, source::snippet_with_applicability};
+use rustc_errors::Applicability;
+use rustc_hir::{intravisit::FnKind, Block, BlockCheckMode, Body, FnDecl, FnSig, HirId, UnsafeSource, Unsafety};
+use rustc_lint::{
+    builtin::{UNSAFE_OP_IN_UNSAFE_FN, UNUSED_UNSAFE},
+    LateContext, LateLintPass, LintContext,
+};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// **What it does:** Checks for macro calls inside an `unsafe` function which expansion
+    /// contains an `unsafe` block, while the macro call is not wrapped in an `unsafe` block
+    /// itself. This lint only triggers in functions where the `unsafe_op_in_unsafe_fn` lint is
+    /// enabled.
+    ///
+    /// **Why is this bad?** This hides an unsafe operation inside a macro call. This is against
+    /// the intention of the `unsafe_op_in_unsafe_fn` lint, which is meant to make unsafe code more
+    /// visible by requiring `unsafe` blocks also in `unsafe` functions.
+    ///
+    /// **Known problems:**
+    /// - In some cases the intention of the macro is to actually hide the unsafety, because the
+    /// macro itself ensures the safety of the `unsafe` operation.
+    /// - For local macros, either an `#[allow(unused_unsafe)]` has to be added to the new unsafe
+    /// block or the unsafe block inside the macro has to be removed.
+    ///
+    /// **Example:**
+    ///
+    /// ```rust
+    /// macro_rules! unsafe_macro {
+    ///     ($e:expr) => {
+    ///         unsafe { $e };
+    ///     };
+    /// }
+    ///
+    /// #[warn(unsafe_op_in_unsafe_fn)]
+    /// unsafe fn foo(x: *const usize) {
+    ///     unsafe_macro!(std::ptr::read(x));
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # macro_rules! unsafe_macro {
+    /// #     ($e:expr) => {
+    /// #         unsafe { $e };
+    /// #     };
+    /// # }
+    /// #[warn(unsafe_op_in_unsafe_fn)]
+    /// unsafe fn foo(x: *const usize) {
+    ///     #[allow(unused_unsafe)] unsafe { unsafe_macro!(std::ptr::read(x)) };
+    /// }
+    /// ```
+    pub MACROS_HIDING_UNSAFE_CODE,
+    restriction,
+    "macros hiding unsafe code, while `unsafe_op_in_unsafe_fn` is enabled"
+}
+
+#[derive(Clone, Copy, Default)]
+pub struct MacrosHidingUnsafeCode {
+    in_unsafe_fn: bool,
+    in_unsafe_block: usize,
+}
+
+impl_lint_pass!(MacrosHidingUnsafeCode => [MACROS_HIDING_UNSAFE_CODE]);
+
+impl LateLintPass<'_> for MacrosHidingUnsafeCode {
+    fn check_block(&mut self, cx: &LateContext<'_>, block: &Block<'_>) {
+        if let BlockCheckMode::UnsafeBlock(UnsafeSource::UserProvided) = block.rules {
+            if in_macro(block.span) {
+                if self.in_unsafe_fn
+                    && self.in_unsafe_block == 0
+                    && !is_lint_allowed(cx, UNSAFE_OP_IN_UNSAFE_FN, block.hir_id)
+                {
+                    let macro_call_span = block.span.source_callsite();
+                    let unused_unsafe_sugg = if !in_external_macro(cx.sess(), block.span)
+                        && !is_lint_allowed(cx, UNUSED_UNSAFE, block.hir_id)
+                    {
+                        "#[allow(unused_unsafe)] "
+                    } else {
+                        ""
+                    };
+                    let mut applicability = Applicability::MaybeIncorrect;
+                    span_lint_and_sugg(
+                        cx,
+                        MACROS_HIDING_UNSAFE_CODE,
+                        macro_call_span,
+                        "this macro call hides unsafe code",
+                        "wrap it in an `unsafe` block",
+                        format!(
+                            "{}unsafe {{ {} }}",
+                            unused_unsafe_sugg,
+                            snippet_with_applicability(cx, macro_call_span, "...", &mut applicability),
+                        ),
+                        applicability,
+                    );
+                }
+            } else {
+                self.in_unsafe_block = self.in_unsafe_block.saturating_add(1);
+            }
+        }
+    }
+
+    fn check_block_post(&mut self, _: &LateContext<'_>, block: &Block<'_>) {
+        if let BlockCheckMode::UnsafeBlock(UnsafeSource::UserProvided) = block.rules {
+            if !in_macro(block.span) {
+                self.in_unsafe_block = self.in_unsafe_block.saturating_sub(1);
+            }
+        }
+    }
+
+    fn check_fn(&mut self, _: &LateContext<'_>, fn_kind: FnKind<'_>, _: &FnDecl<'_>, _: &Body<'_>, _: Span, _: HirId) {
+        if is_unsafe_fn(fn_kind) {
+            self.in_unsafe_fn = true;
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


clippy_lints/src/macros_hiding_unsafe_code.rs
```diff
@@ -0,0 +1,140 @@
...
+        _: &LateContext<'_>,
+        fn_kind: FnKind<'_>,
+        _: &FnDecl<'_>,
+        _: &Body<'_>,
+        _: Span,
+        _: HirId,
+    ) {
+        if is_unsafe_fn(fn_kind) {
+            self.in_unsafe_fn = false;
+        }
```
camsteffen, line 129: I think this needs to be adjusted for nested functions.

```rust
unsafe fn f() {
    fn g() {}
    unsafe_macro!();
}
```

clippy_lints/src/macros_hiding_unsafe_code.rs
```diff
@@ -0,0 +1,140 @@
...
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// **What it does:** Checks for macro calls inside an `unsafe` function which expansion
+    /// contains an `unsafe` block, while the macro call is not wrapped in an `unsafe` block
+    /// itself. This lint only triggers in functions where the `unsafe_op_in_unsafe_fn` lint is
+    /// enabled.
+    ///
+    /// **Why is this bad?** This hides an unsafe operation inside a macro call. This is against
```
popzxc, line 18: I don't think that being against the intention of lint is a strong motivation since it's the nature of the lint to catch what it's supposed to catch. Like, chicken and egg problem.

I'd emphasize that the macro containing an unsafe can hide the fact that it uses unsafe, and thus makes it easier to use in the context where required invariants are not held.

clippy_lints/src/macros_hiding_unsafe_code.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// the intention of the `unsafe_op_in_unsafe_fn` lint, which is meant to make unsafe code more
+    /// visible by requiring `unsafe` blocks also in `unsafe` functions.
+    ///
+    /// **Known problems:**
+    /// - In some cases the intention of the macro is to actually hide the unsafety, because the
+    /// macro itself ensures the safety of the `unsafe` operation.
+    /// - For local macros, either an `#[allow(unused_unsafe)]` has to be added to the new unsafe
```
popzxc, line 25: IMHO, suggesting introducing an unused unsafe block is not a good thing to do; and suggesting `allow(unused_unsafe)` is worse.

Imagine the following chain of actions:

1. Someone introduces a macro with an unsafe code.
2. In the use places it gets wrapped with `#[allow(unused_unsafe)] unsafe { ... }`.
3. Unsafe code is removed from macro.

There is a good  chance that this `unused_unsafe` will remain there unnoticed.

popzxc (commented):
While I totally like the idea of this lint, I think its design should be thought through a bit more carefully.
This lint should not suggest anything a user normally won't do.

Additionally, if some library has such a macro already, dependent crates will have no choice but to either get rid of dependency or add a bunch of `allow(...)`, since they can't directly fix the issue.

Logically, it's the macro fault that it hides `unsafe` code, so the lint should be placed on the macro. Not sure whether it can be implemented though.

clippy_lints/src/macros_hiding_unsafe_code.rs
flip1995: Good point, I'll word this differently.

clippy_lints/src/macros_hiding_unsafe_code.rs
flip1995: Ah right. Your example shouldn't break this code, but 

```rs
unsafe fn f() {
    unsafe fn g() {}
    unsafe_macro!();
}
```

will.

clippy_lints/src/macros_hiding_unsafe_code.rs
flip1995: Yes, I also think this isn't nice. But I don't really see how else to handle this, without rendering the lint completely useless.

flip1995:
> I think its design should be thought through a bit more carefully.

I'm also not really satisfied with how the lint currently behaves. 

> if some library has such a macro already, dependent crates will have no choice but to either get rid of dependency or add a bunch of allow(...), since they can't directly fix the issue.

I don't think this is a problem. I placed this lint in `restriction`, so that you'll have to enable it explicitly if you want it. If a dependent crate uses this lint and uses an external macro, that hides unsafe operations, the dependent crate probably wants to wrap this macro in an `unsafe` block, rather than adding `allow`s.

We could just not lint on external macros, but as I just explained, this lint is supposed to restrict what you can and cannot do with macros that hide unsafe code, no matter the origin.

> Logically, it's the macro fault that it hides unsafe code, so the lint should be placed on the macro.

With that, I think you would run into the problem with dependent crates. If you would lint a macro definition and tell the user to remove an `unsafe` block from the macro, all dependent crates using that macro would now have to add an `unsafe` block around the macro.

We could just not lint exported macros, but I'm not even sure if we're able to lint macro definitions at all, since Clippy is run after macro expansion.

flip1995:
I'll address all comments and further lint design decision once we have consensus  on how the lint should behave.

clippy_lints/src/macros_hiding_unsafe_code.rs
flip1995: When implementing the lint and noticing, that this is necessary, I also thought about whether we should allow lints that contradict `rustc` lints at all. We have lints contradicting other Clippy lints in the `restriction` group. But we don't have lints the contradict `rustc` lints (yet).

popzxc:
> With that, I think you would run into the problem with dependent crates. If you would lint a macro definition and tell the user to remove an `unsafe` block from the macro, all dependent crates using that macro would now have to add an `unsafe` block around the macro.

Actually, I think that's the way it should work. If the crate author thinks that `unsafe` is justified and always valid, they'll just put an `allow` statement on it and leave it as-is. No problem here.
Otherwise, they'll prepare a semver major release explaining that they discovered a potentially dangerous place, so since now users have to wrap the macro usage in `unsafe`. It's also not a problem, and even a good thing, since now the API of the dependency crate is more transparent.

camsteffen:
I see this lint as an *extra* precaution for macro *users*. Library authors with macros should not worry about triggering the lint, and they should not `allow` it. Users of this lint should expect false positives. A macro should have `unsafe {}` if and only if it can guarantee safety. If the guarantee is good, the lint is a false positive. If the guarantee is bad, the macro is wrong. But we can only lint expanded code. I think we should still lint for external macros.

For the lint suggestion, I'm not sure it makes sense to add an unsafe block to appease this lint when it will be deemed unused by rustc. I think it would be better to just allow the lint? This also can be a lint that you run manually to see what you see and just ignore the false positives. I see this lint as being in conflict with rustc's unsafe rules, in its false positive cases, so there is no "fix" to be made.

camsteffen:
Another reason to not suggest adding `unsafe {}` is that it could widen the scope of `unsafe {}`.

flip1995:
Thanks for all the input!

What do you think about suggesting to add an `unsafe` block only for 
1. external macros 
2. optionally: if the `unused_unsafe` lint is allowed

But for local macros the correct fix would be to either remove the `unsafe` block from the macro and put it around every macro call (or the macro argument that makes the unsafe block necessary), OR allow this lint on the macro calls as a marker, that the macro is unsafe. But without producing an auto-applicable suggestion.

popzxc:
Sounds reasonable to me.

camsteffen:
For local macros, yes I agree.

For external macros, I guess it is more reasonable to add `unsafe {}`, but I think allow-this-lint is still the preferred way to suppress because of the conflict with `unused_unsafe`. Maybe write something like this:

```text
help: suppress this lint by writing `#[allow(macros_hiding_unsafe_code)] { macro_name!(..) }`
help: or write `unsafe { macro_name!(..) }`, but this will trigger the `unused_unsafe` lint
```

camsteffen:
I think `macros_hiding_unsafe` would be a better name. Technically it is looking for an unsafe block, where `_unsafe_code` kinda makes it sound like macros can just call unsafe functions anywhere.

flip1995:
> For external macros, I guess it is more reasonable to add `unsafe {}`, but I think allow-this-lint is still the preferred way to suppress because of the conflict with `unused_unsafe`. Maybe output something like this:

The `unused_unsafe` lint doesn't trigger for external macros. I've added a test for this [here](https://github.com/rust-lang/rust-clippy/pull/7469/files#diff-2a6c4afbeffbeb25c79c60aeae5d0611427c63fd3851b9791269e49395dea3daR44). Do you still think suggesting to just allow this lint in this case makes more sense, than suggesting unsafe code. I'm good with suggesting both alternatives.

> I think `macros_hiding_unsafe` would be a better name. Technically it is looking for an unsafe block, where `_unsafe_code` kinda makes it sound like macros can just call unsafe functions anywhere.

Ah yes, thanks for the better name! Now I know what irritated me with the name that I chose :smile:

camsteffen:
> The `unused_unsafe` lint doesn't trigger for external macros.

Huh. Isn't that a bug?

bjorn3:
No, macros may contain unsafe blocks as they need to be callable from a safe context, but still need to do unsafe things internally. If you were to use such a macro inside an `unsafe fn`, then if `unused_unsafe` were to trigger for external macros that would cause an unactionable warning.

camsteffen:
This is about an unsafe block surrounding the external macro call.

bjorn3:
The `unused_unsafe` lint doesn't distinguish `unsafe fn` and `unsafe {}` afaik, but even when looking at `unsafe {}` `unused_unsafe` not triggering for external macros is not a bug as you may have something like `unsafe { unsafe_call(unsafe_containing_macro!()) }` where moving `unsafe_containing_macro!()` out of the unsafe block may be awkward or even impossible. Or you may have more than one statement in the `unsafe` block for example because you temporarily broke a type invariant and want to put all code that may need to handle this broken invariant in the same `unsafe` block.

camsteffen:
> you may have something like `unsafe { unsafe_call(unsafe_containing_macro!()) }`

You seem to be talking about cases where the unsafe block contains unsafe code in addition to the macro (`unsafe_call`). Of course `unused_unsafe` does not apply then.

flip1995:
This has been open way to long with no update by me. This is pretty low on my priority list, so I'm closing it for now, maybe I'll get back to it some time in the future.

Issue #7447, opened 2021-07-08
ICE checking `symbolic-minidump` with 2021-07-07 clippy


Swatinem:
### Code

I haven’t started minimizing this yet, but here is the complete code: https://github.com/getsentry/symbolic/tree/master/symbolic-minidump

```rust
```

### Meta

- `cargo clippy -V`: clippy 0.1.55 (d2b04f07 2021-07-07)
- `rustc -Vv`:
```
rustc 1.55.0-nightly (d2b04f075 2021-07-07)
binary: rustc
commit-hash: d2b04f075c0ce010758c4c8674152ff89d1d73f3
commit-date: 2021-07-07
host: x86_64-pc-windows-msvc
release: 1.55.0-nightly
LLVM version: 12.0.1
```

`cargo check` and `cargo build` work just fine for the crate.

### Error output

```
thread 'rustc' panicked at 'assertion failed: !value.has_escaping_bound_vars()', compiler\rustc_middle\src\ty\sty.rs:968:9

error: internal compiler error: unexpected panic

[…]

query stack during panic:
#0 [is_sized_raw] computing whether `alloc::sync::ArcInner<symbolic_common::byteview::ByteViewBacking<'_>>` is `Sized`
#1 [layout_raw] computing layout of `*const alloc::sync::ArcInner<symbolic_common::byteview::ByteViewBacking<'_>>`
#2 [layout_raw] computing layout of `std::ptr::NonNull<alloc::sync::ArcInner<symbolic_common::byteview::ByteViewBacking<'_>>>`
#3 [layout_raw] computing layout of `std::sync::Arc<symbolic_common::byteview::ByteViewBacking<'_>>`
#4 [layout_raw] computing layout of `symbolic_common::ByteView<'_>`
#5 [layout_raw] computing layout of `cfi::CfiCacheV1<'_>`
#6 [layout_raw] computing layout of `cfi::CfiCacheInner<'_>`
#7 [layout_raw] computing layout of `cfi::CfiCache<'_>`
#8 [analysis] running analysis passes on this crate
end of query stack
error: could not compile `symbolic-minidump`
```

<details><summary>Backtrace</summary>
  <p>
  
```
   0:     0x7ff9995573cf - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h3163d06c760e8856
   1:     0x7ff99957ff7a - core::fmt::write::h7e004fee938dc3bd
[...]

PR #7470, opened 2021-07-15
Add check if ty has_escaping_bound_vars in zero_sized_map_values lint

DevinR528:
Fixes: #7447

changelog: fix ICE in [`zero_sized_map_values`]

### DIFF
--- a/clippy_lints/src/zero_sized_map_values.rs
+++ b/clippy_lints/src/zero_sized_map_values.rs
@@ -4,7 +4,7 @@ use clippy_utils::ty::{is_normalizable, is_type_diagnostic_item, match_type};
 use if_chain::if_chain;
 use rustc_hir::{self as hir, HirId, ItemKind, Node};
 use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::ty::{Adt, Ty};
+use rustc_middle::ty::{Adt, Ty, TypeFoldable};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::sym;
 use rustc_target::abi::LayoutOf as _;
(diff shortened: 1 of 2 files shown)


DevinR528:
Not exactly sure if this is the best way, it seems that checking for `Cow` here would also work but the way it is now, seems more robust since that is the method call causing the ICE?

flip1995:
I think this is a more general problem in Clippy. The same error with `layout_of` happened to me yesterday, when I did the rustup. The fix was the same. Clippy has about 20 calls to `layout_of`. So maybe we should implement a utils function for this, which calls `has_escaping_bound_vars` before calling `layout_of`?

flip1995:
Let's ask about it: https://github.com/rust-lang/rust/pull/86867#issuecomment-881255693

DevinR528:
Cool in a day or so I can just preemptively add this to all the `layout_of` calls, but I will follow the conversation in the rust repo.

flip1995:
> can just preemptively add this to all the `layout_of` calls,

Yeah. Or better: implement a ``clippy_utils::layout_of``, that looks like this:

```rust
fn layout_of(cx: LateContext, ty: Ty) -> Option<_> {
    if ty.has_escaping_bound_vars() {
        return None;
    }
    cx.tcx.layout_of(ty)
}
```

If you don't have time for it, let me know and I'll do this tomorrow. We have to sync back those fixes before the release in 1 and a half weeks.

flip1995:
Since we didn't get any more reports on this, I'm holding off from replacing all of the `layout_of` calls for now. I have a patch ready here: https://github.com/flip1995/rust-clippy/tree/layout_of-util

@bors r+

DevinR528:
Cool thanks! Sorry I didn't get around to it I was doing a bit more traveling than I thought I'll keep an eye on any ICE that could be related.

PR #7477, opened 2021-07-20
Enhance needless continue to detect loop {continue;}

### LINKED ISSUE #7417: Lint on `loop { continue }`
### What it does

`loop { continue }` is heavily used in the embedded ecosystem to sidestep an LLVM miscompilation. However, nowadays it results in the same MIR and thus LLVM IR as `loop {}`, and the miscompilation of `loop {}` has been fixed, so this is unnecessary.

### Categories (optional)

- Kind: `clippy::style` probably, since it's just a confusing way to write `loop {}`

Looks clearer

### Drawbacks

None.

Linting on macros that expand to `continue;` should probably be avoided though.

### Example

```rust
loop { continue; }
```

Could be written as:

```rust
loop {}
```

F3real:
Fixes #7417

changelog: Report [`needless_continue`] in `loop { continue; }` case

### DIFF
--- a/clippy_lints/src/needless_continue.rs
+++ b/clippy_lints/src/needless_continue.rs
@@ -370,15 +370,14 @@ fn suggestion_snippet_for_continue_inside_else<'a>(cx: &EarlyContext<'_>, data: 
 fn check_and_warn<'a>(cx: &EarlyContext<'_>, expr: &'a ast::Expr) {
     if_chain! {
         if let ast::ExprKind::Loop(loop_block, ..) = &expr.kind;
-        if !loop_block.stmts.is_empty();
-        if let ast::StmtKind::Expr(ref statement)
-        | ast::StmtKind::Semi(ref statement) = loop_block.stmts.last().unwrap().kind;
-        if let ast::ExprKind::Continue(_) = statement.kind;
+        if let Some(last_stmt) = loop_block.stmts.last();
+        if let ast::StmtKind::Expr(inner_expr) | ast::StmtKind::Semi(inner_expr) = &last_stmt.kind;
+        if let ast::ExprKind::Continue(_) = inner_expr.kind;
         then {
             span_lint_and_help(
                 cx,
                 NEEDLESS_CONTINUE,
-                loop_block.stmts.last().unwrap().span,
+                last_stmt.span,
                 MSG_REDUNDANT_CONTINUE_EXPRESSION,
                 None,
                 DROP_CONTINUE_EXPRESSION_MSG,
--- a/clippy_lints/src/needless_continue.rs
+++ b/clippy_lints/src/needless_continue.rs
@@ -370,14 +370,14 @@ fn suggestion_snippet_for_continue_inside_else<'a>(cx: &EarlyContext<'_>, data: 
 fn check_and_warn<'a>(cx: &EarlyContext<'_>, expr: &'a ast::Expr) {
     if_chain! {
         if let ast::ExprKind::Loop(loop_block, ..) = &expr.kind;
-        if loop_block.stmts.len() == 1;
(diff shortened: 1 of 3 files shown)


tests/ui/needless_continue.rs
```diff
@@ -49,6 +49,9 @@ fn main() {
 
         println!("bleh");
     }
+    loop {
+        continue;
+    }
```
flip1995, line 54: What about the more general case

````rust
loop {
    some_other_stmt();
    continue;
}
```

So, why not lint on every loop where ``continue`` is the last stmt/expr?

F3real: I covered the issue mentioned in linked issue first, but that's a good point. I expanded it now to cover that case as well.

tests/ui/needless_continue.rs
```diff
@@ -51,6 +51,19 @@ fn main() {
...
+        continue; // should lint here
+    }
+}
+
+fn simple_loop2() {
+    loop {
+        println!("bleh");
+        continue; // should lint here
+    }
+}
```
flip1995, line 65: Thanks for extending it! With that I just thought of another case: No semicolon after the `continue`. I would add a test for this for both cases, the simple one with only `continue` in the loop and the more general case with something before `continue` in the loop.

F3real: Covered this case. I am not sure how much value it adds as this code is automatically changed to `continue;` once user runs fmt (but it is more consistent).

clippy_lints/src/needless_continue.rs
```diff
@@ -364,6 +368,23 @@ fn suggestion_snippet_for_continue_inside_else<'a>(cx: &EarlyContext<'_>, data:
...
+        if let ast::ExprKind::Loop(loop_block, ..) = &expr.kind;
+        if !loop_block.stmts.is_empty();
+        if let ast::StmtKind::Expr(ref statement)
+        | ast::StmtKind::Semi(ref statement) = loop_block.stmts.last().unwrap().kind;
+        if let ast::ExprKind::Continue(_) = statement.kind;
+        then {
+            span_lint_and_help(
+                cx,
+                NEEDLESS_CONTINUE,
+                loop_block.stmts.last().unwrap().span,
```
flip1995, line 381: Formatting looks off here and I don't like the unwrapping. If `loop_block.stmts.last()` already is `Some(_)` you don't have to check if it is empty first. So this could be written as:

```suggestion
        if let Some(last_stmt) = loop_block.stmts.last();
        if let StmtKind::Expr(expr) | StmtKind::Semi(expr) = &last_stmt.kind;
        if let ExprKind::Continue(_) = expr.kind;
        then {
            span_lint_and_help(
                cx,
                NEEDLESS_CONTINUE,
                last_stmt.span,
```

Also import `StmtKind` and `ExprKind` at the top of the file instead of full qualifying it here.

F3real: I agree it is definitely better without `unwrap()`, I wasn't sure how to avoid it. 

The rust fmt just suggested to break line with pattern matching into two lines, but I also wasn't keen on the end result. Is there some other auto formatter that you use as default?

flip1995: Nope, we use `rustfmt`. But `rustfmt` (currently) can't format code in macros. Since this code is in an `if_chain!` macro, `rustfmt` doesn't know how to format it. What `rustfmt` does though is to check if a line is too long regardless if it is in a macro or not. You then get a error message something like "this line is too long, but I don't know how to format it" from `rustfmt`.

flip1995:
@bors r+

Thanks for the contribution and implementing my additional suggestions!

PR #7478, opened 2021-07-20
Fix nonstandard_macro_braces FP and docs of disallowed_types

DevinR528:
changelog: Fix FP in [`nonstandard_macro_braces`] lint

### DIFF
--- a/clippy_lints/src/nonstandard_macro_braces.rs
+++ b/clippy_lints/src/nonstandard_macro_braces.rs
@@ -94,7 +94,8 @@ impl EarlyLintPass for MacroBraces {
 
 fn is_offending_macro<'a>(cx: &EarlyContext<'_>, span: Span, mac_braces: &'a MacroBraces) -> Option<MacroInfo<'a>> {
     if_chain! {
-        if in_macro(span);
+        // Make sure we are only one level deep otherwise there are to many FP's
+        if in_macro(span) && !in_macro(span.ctxt().outer_expn_data().call_site);
         if let Some((name, braces)) = find_matching_macro(span, &mac_braces.macro_braces);
         if let Some(snip) = snippet_opt(cx, span.ctxt().outer_expn_data().call_site);
         // we must check only invocation sites
--- a/clippy_lints/src/nonstandard_macro_braces.rs
+++ b/clippy_lints/src/nonstandard_macro_braces.rs
@@ -99,7 +99,7 @@ fn is_offending_macro<'a>(cx: &EarlyContext<'_>, span: Span, mac_braces: &'a Mac
         if let Some(snip) = snippet_opt(cx, span.ctxt().outer_expn_data().call_site);
         // we must check only invocation sites
         // https://github.com/rust-lang/rust-clippy/issues/7422
-        if snip.starts_with(name);
+        if snip.starts_with(&format!("{}!", name));
         // make formatting consistent
         let c = snip.replace(" ", "");
         if !c.starts_with(&format!("{}!{}", name, braces.0));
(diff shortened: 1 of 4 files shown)


DevinR528:
I could make this 2 PRs I just wasn't sure if it would be worth the change?

giraffate (commented):
Can you add tests?

tests/ui-toml/nonstandard_macro_braces/clippy.toml
```diff
@@ -3,4 +3,5 @@ standard-macro-braces = [
     { name = "quote::quote", brace = "{" },
     { name = "eprint", brace = "[" },
     { name = "type_pos", brace = "[" },
+    { name = "printlnfoo", brace = "[" },
```
jplatte, line 6: The bug in Ruma was that the name in the config was prefix-matched, no? Shouldn't there be a test that using `println` in the config doesn't trigger for `printlnfoo`, rather than the other way around?

DevinR528: Yes, good catch!

tests/ui-toml/nonstandard_macro_braces/conf_nonstandard_macro_braces.rs
```diff
@@ -44,7 +44,7 @@ fn main() {
     let _ = format!["ugh {} stop being such a good compiler", "hello"];
     let _ = quote!(let x = 1;);
     let _ = quote::quote!(match match match);
-    let _ = test!();
+    let _ = test!(); // don't trigger for macro calls inside macros
```
jplatte, line 47: You know if the macro is defined within the same crate..... I would argue it makes sense to still emit the lint, but maybe it's not worth complicating the logic further.

DevinR528: I think I got something that works, that was a bit of a rabbit hole, `Span` is a complicated thing...

DevinR528:
With https://github.com/rust-lang/rust-clippy/pull/7488 I'm going to remove the commit to fix the docs of `disallowed_type` and move it to the other PR.

clippy_lints/src/nonstandard_macro_braces.rs
```diff
@@ -93,13 +93,21 @@ impl EarlyLintPass for MacroBraces {
 }
 
 fn is_offending_macro<'a>(cx: &EarlyContext<'_>, span: Span, mac_braces: &'a MacroBraces) -> Option<MacroInfo<'a>> {
+    let unnested_or_local = || {
+        let nested = in_macro(span.ctxt().outer_expn_data().call_site);
+        let in_local_macro = nested
```
llogiq, line 98: Having `nested &&` here and returning `!nested || ..` might befuddle at cursory reading. I'd suggest using a guard `if` instead.

llogiq (commented):
This looks good in general, I only have one small suggestion for readability.

clippy_lints/src/nonstandard_macro_braces.rs
DevinR528: Is this what you mean?
```rust
let unnested_or_local = || {
    let nested = in_macro(span.ctxt().outer_expn_data().call_site);
    let in_local_macro = matches!(
        span.macro_backtrace().last().and_then(|e| e.macro_def_id),
        Some(defid) if nested && defid.is_local() // adding `nested` to the if guard?
    );
    
    !nested || in_local_macro
};
```

llogiq: I was more thinking of

```
let unnested_or_local = || {
    let nested = in_macro(span.ctxt().outer_expn_data().call_site);
    !nested || span.macro_backtrace().last().map_or(false, |e| 
        e.macro_def_id.map_or(false, |defid| defid.is_local()))
    )
};
```

DevinR528: Oh right duh :facepalm: I was asking the same question two times in a row (`!nested` then `nested && ..`)

DevinR528:
Should I squash, or is there anything else, thanks again!

llogiq:
No, that should be fine.

@bors r+

Issue #7306, opened 2021-06-01
`repeat_once` triggers when the parameter is a constant


f-hanzlik:
Lint name: `repeat_once`

I tried this code:

```rust
fn main() {
    const REPEAT_COUNT: usize = 1;
    
    "x".repeat(REPEAT_COUNT);
}
```

I expected to see this happen: the lint is not triggered because if I used a constant as a parameter I probably intend to allow changing the constant value to something else in the future.

Instead, this happened: the lint triggers

### Meta

- `cargo clippy -V`: clippy 0.1.54 (1c6868a 2021-05-27
- `rustc -Vv`:
  ```
  rustc 1.54.0-nightly (1c6868aa2 2021-05-27)
  binary: rustc
  commit-hash: 1c6868aa21981b37cbd3fc95828ee3b0ac22d494
  commit-date: 2021-05-27
  host: x86_64-unknown-linux-gnu
  release: 1.54.0-nightly
  LLVM version: 12.0.1
  ```

xFrednet:
Hey @giraffate and @flip1995 the lint implementation contains a deliberate test that resolves the value of constants, like in this case. You two discussed the check [here](https://github.com/rust-lang/rust-clippy/pull/5773#discussion_r453650766). Would you also consider this to be a false positive, or working as intended? :upside_down_face:

giraffate:
Yes, that was the intention. I thought that warning is valid in many cases at that time.

> because if I used a constant as a parameter I probably intend to allow changing the constant value to something else in the future.

However, if this is a common case, it's a good idea to change that behavior.

f-hanzlik:
I'd like to note that I don't feel too strongly about this - I can always just disable the lint. I just wanted to report it since it felt a little surprising.

flip1995:
> because if I used a constant as a parameter I probably intend to allow changing the constant value to something else in the future.

I mean, if you change this constant the lint will no longer trigger. But as long as the constant is `1` the lint will tell you that it is `1`. So I think the lint is right here. Even if in this case it will rather remind you to change the constant than that the `repeat(CONST)` call is wrong. I see value in that too.

So I would keep the current behavior. Especially because code like

```rust
"x".repeat(NUMBER + const_fn() * ANOTHER_CONST);
```

may result in an unintentional `"x".repeat(1)`. In this case, the lint is really valuable.

xFrednet:
[...]

PR #7482, opened 2021-07-23
Documented constant expression evaluation for `repeat_once`

xFrednet:
Documents the fact that the `repeat_once` lint evaluates constant expressions

---

closes: #7306 

changelog: none
(I don't think it's worth a change log entry, as nothing has really changed)

r? @giraffate as you've implemented the lint and were part of the discussion in the issue :upside_down_face:


giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #7484, opened 2021-07-23
Some `clippy::author` improvements

camsteffen:
changelog: none

* Use `Debug` instead of re-implementing it for some things
* Fix block trailing expression handing
* Don't double print on stmt/expr with `#[clippy::author]` attribute

### DIFF
--- a/clippy_lints/src/utils/author.rs
+++ b/clippy_lints/src/utils/author.rs
@@ -554,6 +554,8 @@ impl<'tcx> Visitor<'tcx> for PrintVisitor {
             self.current = self.next("trailing_expr");
             println!("    if let Some({}) = &{}.expr;", self.current, block_name);
             self.visit_expr(expr);
+        } else {
+            println!("    if {}.expr.is_none();", block_name);
         }
     }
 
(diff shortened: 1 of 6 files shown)


clippy_lints/src/utils/author.rs
```diff
@@ -536,14 +544,17 @@ impl<'tcx> Visitor<'tcx> for PrintVisitor {
...
         for (i, stmt) in block.stmts.iter().enumerate() {
-            self.current = format!("{}.stmts[{}]", current, i);
+            self.current = format!("{}.stmts[{}]", block_name, i);
             self.visit_stmt(stmt);
         }
+        if let Some(expr) = block.expr {
+            self.current = self.next("trailing_expr");
+            println!("    if let Some({}) = &{}.expr;", self.current, block_name);
+            self.visit_expr(expr);
+        }
```
xFrednet, line 557: I think it would make sense to add an else case here that checks that the expression is none
```suggestion
        } else {
            println!("    if {}.expr.is_none();", block_name)
        }
```

flip1995:
@bors r+

Thanks!

PR #7488, opened 2021-07-24
Add primitive type support to disallowed_type lint

### LINKED ISSUE #6845: Lint request: `char_is_the_wrong_abstraction`
### What it does

We want a restriction lint that prevents using `char` and `std::str::Chars`. We want to discourage ppl contributing to our code from using `char` because we believe it's just the wrong abstraction, and would rather it just errored so we can explain to ppl that things like grapheme clusters and whatnot exist and they should be preferred. [The std suggests you may want to look into grapheme clusters instead of using `str.chars`](https://doc.rust-lang.org/std/primitive.str.html#method.chars), which is a great suggestion but unfortunately it's just not made prominent enough, and this would fix that. Further, even the std documentation often passes the wrong idea about `char`-based APIs, leading to bugs in user code (this isn't meant to say those APIs should be deprecated, or those docs shouldn't be fixed. it's just mean to show that even std makes mistakes, so `char` is just a bad idea in general).

### Categories (optional)

- Kind: restriction
- Internationalization

### Drawbacks

Being able to use an `fn(char)->bool` as a `Pattern` is actually somewhat useful, but it's easy enough to slap an `#[allow]` on those so it's not a major drawback.

### Example

```rust
char
```

Could be written as:

```rust
SmallString<[u8; 8]>
```

### Motivation

See also https://github.com/kdeldycke/awesome-falsehood#internationalization and in particular https://eev.ee/blog/2015/09/12/dark-corners-of-unicode/ and https://manishearth.github.io/blog/2017/01/14/stop-ascribing-meaning-to-unicode-code-points/

DevinR528:
Closes #6845

changelog: Enable the banning of primitive types in [`disallowed_type`]

### DIFF
--- a/clippy_lints/src/disallowed_type.rs
+++ b/clippy_lints/src/disallowed_type.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::span_lint;
 
 use rustc_data_structures::fx::FxHashSet;
 use rustc_hir::{
-    def::Res, def_id::DefId, Crate, Item, ItemKind, PolyTraitRef, TraitBoundModifier, Ty, TyKind, UseKind,
+    def::Res, def_id::DefId, Crate, Item, ItemKind, PolyTraitRef, PrimTy, TraitBoundModifier, Ty, TyKind, UseKind,
 };
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
@@ -15,14 +15,12 @@ declare_clippy_lint! {
     ///
     /// **Known problems:** None.
     ///
-    /// N.B. There is no way to ban primitive types.
-    ///
     /// **Example:**
     ///
     /// An example clippy.toml configuration:
     /// ```toml
     /// # clippy.toml
-    /// disallowed-methods = ["std::collections::BTreeMap"]
+    /// disallowed-types = ["std::collections::BTreeMap"]
     /// ```
     ///
     /// ```rust,ignore
@@ -42,7 +40,8 @@ declare_clippy_lint! {
 #[derive(Clone, Debug)]
 pub struct DisallowedType {
     disallowed: FxHashSet<Vec<Symbol>>,
-    def_ids: FxHashSet<(DefId, Vec<Symbol>)>,
+    def_ids: FxHashSet<DefId>,
+    prim_tys: FxHashSet<PrimTy>,
 }
 
 impl DisallowedType {
@@ -53,6 +52,23 @@ impl DisallowedType {
                 .map(|s| s.split("::").map(|seg| Symbol::intern(seg)).collect::<Vec<_>>())
                 .collect(),
             def_ids: FxHashSet::default(),
+            prim_tys: FxHashSet::default(),
+        }
+    }
+
+    fn check_res_emit(&self, cx: &LateContext<'_>, res: &Res, span: Span) {
+        match res {
+            Res::Def(_, did) => {
+                if self.def_ids.contains(did) {
+                    emit(cx, &cx.tcx.def_path_str(*did), span);
+                }
+            },
+            Res::PrimTy(prim) => {
+                if self.prim_tys.contains(prim) {
+                    emit(cx, prim.name_str(), span);
+                }
+            },
+            _ => {},
         }
     }
 }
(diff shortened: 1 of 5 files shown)


tests/ui-toml/toml_disallowed_type/clippy.toml
```diff
@@ -6,4 +6,5 @@ disallowed-types = [
     "std::thread::Thread",
     "std::time::Instant",
     "std::io::Read",
+    "std::primitive::usize"
```
jplatte, line 9: Is this required with the currently implementation, or could one just write `"usize"` here?

camsteffen: Agreed, that would be better. The docs for std::primitive says it is re-exports of primitives. The simple names are the canonical names.

DevinR528: The problem is that [path_to_res](https://github.com/rust-lang/rust-clippy/blob/master/clippy_utils/src/lib.rs#L466) doesn't allow non `crate::name` paths, and looking at the code I'm not sure it could be altered to work either. Is there a way to go from `Symbol` to `DefId` or search all items visible to the current crate?

camsteffen: You can use `PrimTy::from_name`.

camsteffen:
Closes #6845

camsteffen:
Closes #6845

clippy_lints/src/disallowed_type.rs
```diff
@@ -65,6 +65,12 @@ impl_lint_pass!(DisallowedType => [DISALLOWED_TYPE]);
 impl<'tcx> LateLintPass<'tcx> for DisallowedType {
     fn check_crate(&mut self, cx: &LateContext<'_>, _: &Crate<'_>) {
         for path in &self.disallowed {
+            if path.len() == 1 {
+                if let Some(ty) = PrimTy::from_name(path[0]) {
+                    self.prim_ty.insert(ty);
+                    continue;
+                }
+            }
```
DevinR528, line 73: This seemed the most efficient and direct way to do this, if you were thinking something else let me know!

camsteffen (changes requested):
Some minor requests. This is a nice enhancement!

clippy_lints/src/disallowed_type.rs
DevinR528: Unfortunately, I can't use `path_to_res` for primitive types unless they are of the form `crate::[mod]::name` that's why I was asking about `Symbol` to `DefId` (although looking further at how rustc handles primitives they seem to be a separate thing having their `DefId` defined in a macro via `LangItems` or something like that).

clippy_lints/src/disallowed_type.rs
```diff
@@ -15,14 +15,15 @@ declare_clippy_lint! {
     ///
     /// **Known problems:** None.
     ///
-    /// N.B. There is no way to ban primitive types.
+    /// N.B. Primitive types can be banned using their full import path
+    /// i.e. `std::primitive::usize` or simply `usize`.
```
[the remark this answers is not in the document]
DevinR528: I'll get rid of the comment but leave the ability to resolve `std::primitive::usize` right? as in all the `match` with `Res::PrimTy(prim)`.

clippy_lints/src/disallowed_type.rs
```diff
@@ -74,31 +88,44 @@ impl<'tcx> LateLintPass<'tcx> for DisallowedType {
...
-            if let TyKind::Path(path) = &ty.kind;
-            if let Some(did) = cx.qpath_res(path, ty.hir_id).opt_def_id();
-            if let Some((_, name)) = self.def_ids.iter().find(|(id, _)| *id == did);
-            then {
-                emit(cx, name, path.span());
+        if let TyKind::Path(path) = &ty.kind {
+            match cx.qpath_res(path, ty.hir_id) {
+                Res::Def(_, did) => {
+                    if let Some(name) = self.def_ids.get(&did) {
+                        emit(cx, name, path.span());
```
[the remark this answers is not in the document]
DevinR528: So the  last commit has what I think you mean by factor `check_qpath` but I'm unsure?

As for using `cx.txc.def_path_str` I'm fine with it I just wanted you to be aware of my reasoning for not using it. The user seeing the exact path that they banned in the `Conf` struct may be less surprising than banning a re-export and getting a different name in the warning. I'm fine either way just thought to mention it :shrug:

clippy_lints/src/disallowed_type.rs
DevinR528: Ahh do you mean add the above code to the `path_to_res` function?

clippy_lints/src/disallowed_type.rs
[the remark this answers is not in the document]
camsteffen: Yes, path_to_res should work for re-exports, PrimTy or otherwise.

clippy_lints/src/disallowed_type.rs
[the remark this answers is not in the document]
camsteffen: I think that's okay if it doesn't match the config exactly.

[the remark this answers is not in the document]
DevinR528: Oh yeah, it seems to actually follow the re-exports (notice syn::Ident) nice thanks for putting up with all my questions!

camsteffen (approved):
Looks great! Please squash commits.

clippy_lints/src/disallowed_type.rs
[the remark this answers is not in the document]
camsteffen: No problem! `cx.get_def_path` gets the absolute path so it's confusing.

DevinR528:
As always thanks for all the help!

PR #7420, opened 2021-07-02
Update lint documentation to use markdown headlines

xFrednet:
This PR updates all lint documentation to use markdown headlines. It additionally removed the *Known problems* section for lints without any problems. I've double-checked all automatic replacements, but a second pair of eyes is definitely appreciated!

I wasn't sure when you wanted to switch to the new metadata collection tomorrow, I therefore prepared this PR today. And that's it this is a standalone PR to keep the other related PRs reviewable. 

changelog:  none

r? @flip1995 

cc: #7172 

Note: This should be merged with the other metadata collection related PRs.

PR #7350, opened 2021-06-14
Add suspicious group

camsteffen:
changelog: Introduce `clippy::suspicious` 🤔 group and move several lints into the group

Closes #6366. CC #6626.

A number of lints are moved from each of `correctness`, `style` and `complexity` groups. Notably I didn't move `suspicious_splitn` since I think that is a `correctness` lint despite the name.

Lints moved to `clippy::suspicious`:
* `blanket_clippy_restriction_lints` (was `clippy::style`)
* `empty_loop` (was `clippy::style`)
* `eval_order_dependence` (was `clippy::complexity`)
* `float_equality_without_abs` (was `clippy::correctness`)
* `for_loops_over_fallibles` (was `clippy::correctness`)
* `misrefactored_assign_op` (was `clippy::complexity`)
* `mut_range_bound` (was `clippy::complexity`)
* `mutable_key_type` (was `clippy::correctness`)
* `suspicious_arithmetic_impl` (was `clippy::correctness`)
* `suspicious_assignment_formatting` (was `clippy::style`)
* `suspicious_else_formatting` (was `clippy::style`)
* `suspicious_map` (was `clippy::complexity`)
* `suspicious_op_assign_impl` (was `clippy::correctness`)
* `suspicious_unary_op_formatting` (was `clippy::style`)

PR #7498, opened 2021-07-28
Updated changelog for 1.55

xFrednet:
This has again been a bit of work, but I'm happy to notice that my English is still improving, and I'm getting faster at these things. That's a very nice side effect of contributing and getting feedback on reviews :blush:

Moving on, there were a few things that I was unsure about:
* The PR rust-lang/rust#86717 changes an old entry in the change log, is this worth mentioning? I've left it out for now.
* The stabilization of `cargo clippy --fix` is quite awesome and important IMO. It sadly gets a bit lost in the *Other* entry, as it's the last one. Do we maybe want to move it somewhere else or change the headline order for this release?
* I've listed the introduction of the new `suspicious` group under the *Moves and Deprecations* section. Is this alright, or should it be moved to the *Other* section as well?
* Last but definitely not least, some fun! I've used the :tada: emoji in the `cargo clippy --fix` entry, is this okay?

Sorry for the bombardment of questions xD

---

The PR already includes the entries for the new metadata collection and website updates. These are not merged yet, but should probably be to make this correct. This might also require the commit hashes to be updated (Not sure on this, though). It would actually be super fitting to get this into this release as we also stabilize `--fix`. TODOs:
* [x] Merge metadata collection PRs:
  1. #7279
  2. #7298
  3. #7420 (Hope to not get any merge conflicts)

---

[Rendered :newspaper:](https://github.com/xFrednet/rust-clippy/blob/changelog-1-55/CHANGELOG.md)

r? @flip1995

changelog: none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,141 @@ document.
 
 ## Unreleased / In Rust Nightly
 
-[3ae8faf...master](https://github.com/rust-lang/rust-clippy/compare/3ae8faf...master)
+[74d1561...master](https://github.com/rust-lang/rust-clippy/compare/74d1561...master)
+
+## Rust 1.55
+
+Current beta, release 2021-09-09
+
+[3ae8faf...74d1561](https://github.com/rust-lang/rust-clippy/compare/3ae8faf...74d1561)
+
+### Important Changes
+
+* Stabilized `cargo clippy --fix` :tada:
+  [#7405](https://github.com/rust-lang/rust-clippy/pull/7405)
+
+### New Lints
+
+* [`rc_mutex`]
+  [#7316](https://github.com/rust-lang/rust-clippy/pull/7316)
+* [`nonstandard_macro_braces`]
+  [#7299](https://github.com/rust-lang/rust-clippy/pull/7299)
+* [`strlen_on_c_strings`]
+  [#7243](https://github.com/rust-lang/rust-clippy/pull/7243)
+* [`self_named_constructors`]
+  [#7403](https://github.com/rust-lang/rust-clippy/pull/7403)
+* [`disallowed_script_idents`]
+  [#7400](https://github.com/rust-lang/rust-clippy/pull/7400)
+* [`disallowed_type`]
+  [#7315](https://github.com/rust-lang/rust-clippy/pull/7315)
+* [`missing_enforced_import_renames`]
+  [#7300](https://github.com/rust-lang/rust-clippy/pull/7300)
+* [`extend_with_drain`]
+  [#7270](https://github.com/rust-lang/rust-clippy/pull/7270)
+
+### Moves and Deprecations
+
+* Moved [`from_iter_instead_of_collect`] to `pedantic`
+  [#7375](https://github.com/rust-lang/rust-clippy/pull/7375)
+* Added `suspicious` as a new lint group for *code that is most likely wrong or useless*
+  [#7350](https://github.com/rust-lang/rust-clippy/pull/7350)
+  * Moved [`blanket_clippy_restriction_lints`] to `suspicious`
+  * Moved [`empty_loop`] to `suspicious`
+  * Moved [`eval_order_dependence`] to `suspicious`
+  * Moved [`float_equality_without_abs`] to `suspicious`
+  * Moved [`for_loops_over_fallibles`] to `suspicious`
+  * Moved [`misrefactored_assign_op`] to `suspicious`
+  * Moved [`mut_range_bound`] to `suspicious`
+  * Moved [`mutable_key_type`] to `suspicious`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,11 +6,141 @@ document.
...
+
+### Others
+
+* Stabilized `cargo clippy --fix` :tada:
+  [#7405](https://github.com/rust-lang/rust-clippy/pull/7405)
 
 ## Rust 1.54
 
-Current beta, release 2021-07-29
+Current stable, release 2021-07-29
```
flip1995, line 143:
```suggestion
Current stable, released 2021-07-29
```

CHANGELOG.md
```diff
@@ -6,11 +6,141 @@ document.
...
+* [`append_instead_of_extend`]
+  [#7270](https://github.com/rust-lang/rust-clippy/pull/7270)
+
+### Moves and Deprecations
+
+* Moved [`from_iter_instead_of_collect`] to `pedantic`
+  [#7375](https://github.com/rust-lang/rust-clippy/pull/7375)
+* Added `suspicious` as a new lint group for *code that is most likely wrong or useless*
+  [#7350](https://github.com/rust-lang/rust-clippy/pull/7350)
+  * Moved [`blanket_clippy_restriction_lints`] to `suspicious`
```
flip1995, line 42: I would write those as 

> Moved [`blanket_clippy_restriction_lints`] from `<lint_group>`

You can get the old lint groups from the PR body of #7350 

No need to mention here, that they were moved TO `suspicious` since this is already stated one line above.

flip1995 (commented):
Just 2 things everything else LGTM. Thanks for doing this!

- No need to mention this
- We could introduce a new section on top of the list for important changes. Not sure how to name this though.
- I had done the same thing
- I don't mind

CHANGELOG.md
```diff
@@ -6,11 +6,141 @@ document.
...
+* [`wildcard_enum_match_arm`], [`match_wildcard_for_single_variants`]: Now allows wildcards for enums with unstable
+  or hidden variants
+  [#7407](https://github.com/rust-lang/rust-clippy/pull/7407)
+* [`redundant_allocation`]: Now additionally supports the `Arc<>` type
+  [#7308](https://github.com/rust-lang/rust-clippy/pull/7308)
+* [`blacklisted_name`]: Now allows blacklisted names in test code
+  [#7379](https://github.com/rust-lang/rust-clippy/pull/7379)
+* [`redundant_closure`]: Suggests `&mut` for `FnMut`
+  [#7437](https://github.com/rust-lang/rust-clippy/pull/7437)
+* [`disallowed_method`], [`disallowed_type`]: The configuration values `disallowed-method` and `disallowed_type`
```
giraffate, line 75: nits
```suggestion
* [`disallowed_method`], [`disallowed_type`]: The configuration values `disallowed-method` and `disallowed-type`
```

CHANGELOG.md
xFrednet: I wrote them this way to always have the same format for moved lints regardless of the context and I don't think that the original linting group is that interesting to the end user. If someone is really interested they can checkout the linked PR IMO. The main interest to me would actually be if the lint level has changed but most users don't directly know what group has which lint level. :thinking:

xFrednet:
Thank you for the feedback! I'll fix the NITs later today, that should definitely be in time for the sync tomorrow :upside_down_face: 

> We could introduce a new section on top of the list for important changes. Not sure how to name this though.

How about something like *Highlight* or *Major changes* ?

CHANGELOG.md
flip1995: Yeah makes sense. Then never mind me :)

flip1995:
How about "Important Changes"? Stabilizing something is not necessarily a "Major" change.

camsteffen:
>     * The stabilization of `cargo clippy --fix` is quite awesome and important IMO. It sadly gets a bit lost in the _Other_ entry, as it's the last one. Do we maybe want to move it somewhere else or change the headline order for this release?

Just put it at the top without a header?

camsteffen:
>  I've used the 🎉 emoji in the `cargo clippy --fix` entry, is this okay?

Absolutely not. This is a professional work environment.

flip1995:
> Just put it at the top without a header?

Also a good idea. +1 for that.

flip1995:
Hijacking this PR to reopen the tree

@bors treeclosed-

xFrednet:
Having it without a headline is a possibility but than it seems to be a bit like an outsider IMO

image

vs:

image

So I would add the *Important Changes* if that works for everyone :upside_down_face:

flip1995 (approved):
Will merge once release is done.

flip1995 (approved):

PR #7504, opened 2021-07-29
Rename two lints to comply with our lint naming convention

flip1995:
self_named_constructor -> self_named_constructors
append_instead_of_extend -> extend_with_drain

We don't need to `register_renamed` those lints, since I'll backport them to beta, so the old names won't hit stable.

changelog: none
(I'll adapt the changelog before merging #7498)


flip1995:
Those new names were discussed and agreed upon on Zulip: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Misnamed.20lint.20.60append_instead_of_extend.60/near/247352538

So merging myself

@bors r+

PR #7506, opened 2021-07-29
Add xor case to `manual swap` lint

### LINKED ISSUE #6598: Add XOR swap algorithm detection to (existing) clippy::manual_swap
### What it does

*This would warn the user that there is a better alternative to using a XOR swapping algorithm to swap the values of two variables*

### Categories (optional)

- Kind: *clippy::complexity*

*The std::mem::swap function exposes the intent better (this is somewhat the same reason for the warning for swapping using a temporary variable)*

For example:
- People reading the recommended code will see that it likely swaps two variables, as opposed to if they see a XOR swap algorithm and are unfamiliar with it.

### Drawbacks

None.

### Example

```rust
fn main() {
    let mut a = 5;
    let mut b = 10;
    
    // Swap them.
    a ^= b;
    b ^= a;
    a ^= b;
    
    println!("a: {}, b: {}", a, b);
}
```

Could be written as:

```rust
fn main() {
    let mut a = 5;
    let mut b = 10;
    
    // Swap them.
    std::mem::swap(&mut a, &mut b);
    
    println!("a: {}, b: {}", a, b);
}
```

HKalbasi:
Continue of #7153
closes #6598

changelog: Add "xor swap" case to [`manual_swap`]


HKalbasi:
I think @llogiq is busy.
r? @camsteffen

tests/ui/swap.rs
```diff
@@ -66,13 +66,61 @@ fn vec() {
...
+    b ^= c;
+    a ^= c;
+    c ^= a;
+}
+
+fn xor_unswappable_slice() {
+    let foo = &mut [vec![1, 2], vec![3, 4]];
+    foo[0][1] ^= foo[1][0];
+    foo[1][0] ^= foo[0][0];
+    foo[0][1] ^= foo[1][0];
```
camsteffen, line 109: Can you add a comment like this?

> this could use `split_at_mut` and `mem::swap`, but that is not much simpler.

clippy_lints/src/swap.rs
```diff
@@ -262,3 +264,82 @@ fn check_suspicious_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
+                match check_for_slice(cx, lhs0, rhs0) {
+                    Slice::Swappable(slice, idx0, idx1) => {
+                        if let Some(slice) = Sugg::hir_opt(cx, slice) {
+                            span_lint_and_sugg(
+                                cx,
+                                MANUAL_SWAP,
+                                span,
+                                &format!(
+                                    "this xor-based swap of the elements of `{}` can be \
+                                    more clearly expressed using `.swap`",
```
camsteffen, line 291:
```suggestion
                                    more clearly expressed using `slice::swap`",
```

clippy_lints/src/swap.rs
```diff
@@ -262,3 +264,82 @@ fn check_suspicious_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
+                let mut applicability = Applicability::MachineApplicable;
+                match check_for_slice(cx, lhs0, rhs0) {
+                    Slice::Swappable(slice, idx0, idx1) => {
+                        if let Some(slice) = Sugg::hir_opt(cx, slice) {
+                            span_lint_and_sugg(
+                                cx,
+                                MANUAL_SWAP,
+                                span,
+                                &format!(
+                                    "this xor-based swap of the elements of `{}` can be \
```
camsteffen, line 290: I think we can safely simply the message like this, and do similarly with the other message. But you may disagree.

```suggestion
                                    "this xor-based swap can be \
```

camsteffen (changes requested):
Thanks for picking this up!

clippy_lints/src/swap.rs
```diff
@@ -262,3 +264,82 @@ fn check_suspicious_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
+            if let Some((lhs1, rhs1)) = extract_sides_of_xor_assign(&window[1]);
+            if let Some((lhs2, rhs2)) = extract_sides_of_xor_assign(&window[2]);
+            if eq_expr_value(cx, lhs0, rhs1);
+            if eq_expr_value(cx, rhs1, lhs2);
+            if eq_expr_value(cx, rhs0, lhs1);
+            if eq_expr_value(cx, lhs1, rhs2);
+            then {
+                let span = window[0].span.to(window[2].span);
+                let mut applicability = Applicability::MachineApplicable;
+                match check_for_slice(cx, lhs0, rhs0) {
```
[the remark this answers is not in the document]
HKalbasi: `check_for_slice` was there before this PR. So I don't think it is useless. And your changes seems doesn't detect vector and array:
```
let foo = &mut [1, 2];
-    foo.swap(0, 1)
+    std::mem::swap(&mut foo[0], &mut foo[1])
```

clippy_lints/src/swap.rs
HKalbasi: I unified messages. So now it is: "This looks like you are swapping elements of `foo` manually."

HKalbasi:
One question. In previous manual swap there is this test:
```
let mut bar = Bar { a: 1, b: 2 };
let temp = bar.a;
bar.a = bar.b;
bar.b = temp;
```
And it is ignored by lint, but we can also swap it with `std::mem::swap`, what is reason behind this?

camsteffen:
🤷 Maybe whoever wrote the lint assumed that the borrow checker wouldn't allow it. But this was always allowed AFAICT.

clippy_lints/src/swap.rs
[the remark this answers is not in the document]
camsteffen: > `check_for_slice` was there before this PR. So I don't think it is useless.

Well I didn't say it was useless. But that's okay to leave it.

> And your changes seems doesn't detect vector and array:

Ah my bad. I thought the indexing operation always derefs to slice but that is not the case.

clippy_lints/src/swap.rs
camsteffen: Hmm I actually think that mentioning the method `slice::swap` is the most important part. The error is: "this method is available and you are not using it".

HKalbasi: It is being mentioned. The full message is:
```
error: this looks like you are swapping elements of `foo` manually
  --> $DIR/swap.rs:91:5
   |
LL | /     foo[0] ^= foo[1];
LL | |     foo[1] ^= foo[0];
LL | |     foo[0] ^= foo[1];
   | |_____________________^ help: try: `foo.swap(0, 1)`
```

In-line with:
```
error: this looks like you are swapping elements of `foo` manually
  --> $DIR/swap.rs:36:5
   |
LL | /     let temp = foo[0];
LL | |     foo[0] = foo[1];
LL | |     foo[1] = temp;
   | |_________________^ help: try: `foo.swap(0, 1)`
```
I'm not in favor of changing old code too much. But if message isn't clear we can change it.

clippy_lints/src/swap.rs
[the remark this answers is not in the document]
HKalbasi: I found a false negative in old code:
```rust
fn distinct_slice() {
    let foo = &mut [vec![1, 2], vec![3, 4]];
    let bar = &mut [vec![1, 2], vec![3, 4]];
    let temp = foo[0][1];
    foo[0][1] = bar[1][0];
    bar[1][0] = temp;
}
```
So now I'm with you on removing `check_for_slice`. Is there a function that checks if we can mutably borrow two expression at the same time? It will also solve fields problem.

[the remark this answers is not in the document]
camsteffen: Oof no I'm afraid not. I think you would have to manually traverse the expression.

If you are swapping `foo.a...` and `foo.b...` that is okay because `a` and `b` differ. But if you are swapping `foo.a[b]...` and `foo.a[c]...` that is not okay because indices differ. So its tricky because you traverse into the expression "right to left" but you'll have to compare them "left to right".

You don't have to fix the false negative in this PR but its up to you.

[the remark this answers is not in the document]
HKalbasi: I done it, but there may be a better way. Also I removed `check_for_slice`.

clippy_lints/src/swap.rs
```diff
@@ -70,9 +71,133 @@ impl<'tcx> LateLintPass<'tcx> for Swap {
...
+                (ep1, ep2)
+            },
+            (ExprKind::Index(ep1, _), ExprKind::Field(ep2, _)) | (ExprKind::Field(ep1, _), ExprKind::Index(ep2, _)) => {
+                (ep1, ep2)
+            },
+            _ => return false,
+        };
+        e1 = ep1;
+        e2 = ep2;
+    }
```
camsteffen, line 134: There are a lot of redundant HIR traversals here: `root_of_borrow`, each `eq_expr_value` call, each`go_up` call, and finally this loop. You can traverse the HIR just once for `e1` and `e2` and push `Expr`s onto a stack.

clippy_lints/src/swap.rs
```diff
@@ -70,9 +71,107 @@ impl<'tcx> LateLintPass<'tcx> for Swap {
...
+            if let ExprKind::Index(lhs2, idx2) = e2.kind {
+                if eq_expr_value(cx, lhs1, lhs2) {
+                    let ty = cx.typeck_results().expr_ty(lhs1).peel_refs();
+
+                    if matches!(ty.kind(), ty::Slice(_))
+                        || matches!(ty.kind(), ty::Array(_, _))
+                        || is_type_diagnostic_item(cx, ty, sym::vec_type)
+                        || is_type_diagnostic_item(cx, ty, sym::vecdeque_type)
+                    {
+                        if let Some(slice) = Sugg::hir_opt(cx, lhs1) {
```
camsteffen, line 125: This is pre-existing but `Sugg::hir_opt` would be better as `Sugg::hir_with_applicability` here and below.

clippy_lints/src/swap.rs
```diff
@@ -70,9 +71,107 @@ impl<'tcx> LateLintPass<'tcx> for Swap {
...
+
+    let (replace, what, sugg) = if let (Some(first), Some(second)) = (Sugg::hir_opt(cx, e1), Sugg::hir_opt(cx, e2)) {
+        (
+            true,
+            format!(" `{}` and `{}`", first, second),
+            format!("std::mem::swap({}, {})", first.mut_addr(), second.mut_addr()),
+        )
+    } else {
+        (true, String::new(), String::new())
+    };
```
camsteffen, line 156: `replace` is now always true.

clippy_lints/src/swap.rs
```diff
@@ -70,9 +71,107 @@ impl<'tcx> LateLintPass<'tcx> for Swap {
...
+    false
+}
+
+fn generate_swap_warning(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>, span: Span, is_xor_based: bool) {
+    let mut applicability = Applicability::MachineApplicable;
+
+    if !is_swappable(cx, e1, e2) {
+        if let ExprKind::Index(lhs1, idx1) = e1.kind {
+            if let ExprKind::Index(lhs2, idx2) = e2.kind {
+                if eq_expr_value(cx, lhs1, lhs2) {
```
camsteffen, line 117: Another `eq_expr_value` should not be necessary at this point. You may need to change `is_swappable` to having three possible outcomes (again). Suggestion: `Option<Swappable>`, `enum Swappable { MemSwap, SliceSwap }`. I know that is effectively the same as `Slice` but I think much clearer.

clippy_lints/src/swap.rs
```diff
@@ -70,9 +71,107 @@ impl<'tcx> LateLintPass<'tcx> for Swap {
...
+        MANUAL_SWAP,
+        span,
+        &format!("this looks like you are swapping{} manually", what),
+        |diag| {
+            if !sugg.is_empty() {
+                diag.span_suggestion(span, "try", sugg, applicability);
+
+                if replace && !is_xor_based {
+                    diag.note("or maybe you should use `std::mem::replace`?");
+                }
```
camsteffen, line 169: Why is this dependent on xor?

camsteffen (changes requested):

clippy_lints/src/swap.rs
HKalbasi: Why we suggest replace? I think `mem::replace` it is equal to two first statements of `temp` version and it isn't related to `xor` version. By this logic I turn it off for `xor` version.

clippy_lints/src/swap.rs
camsteffen: Okay, that's fair. You could rename `is_swappable` to `can_mut_borrow_both` and put it in clippy_utils.

clippy_lints/src/swap.rs
camsteffen: That makes sense.

clippy_lints/src/swap.rs
HKalbasi: Is it now good?

clippy_lints/src/swap.rs
```diff
@@ -70,9 +71,107 @@ impl<'tcx> LateLintPass<'tcx> for Swap {
...
+            break;
+        }
+    }
+    result.reverse();
+    result
+}
+
+fn is_swappable(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>) -> bool {
+    let s1 = subpart_stack(e1);
+    let s2 = subpart_stack(e2);
```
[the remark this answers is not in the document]
HKalbasi: I renamed it to `projection_stack`. Do we need to do something for `subslice` and `deref` cases as well? I think `subslice` is already handled, but don't know about `deref`.

clippy_utils/src/lib.rs
```diff
@@ -558,6 +558,54 @@ pub fn trait_ref_of_method<'tcx>(cx: &LateContext<'tcx>, hir_id: HirId) -> Optio
...
+/// this method will return a tuple, composed of a `Vec`
+/// containing the `Expr`s for `v[0], v[0].a, v[0].a.b, v[0].a.b[x]`
+/// and a `Expr` for root of them, `v`
+fn projection_stack<'a, 'hir>(mut e: &'a Expr<'hir>) -> (Vec<&'a Expr<'hir>>, &'a Expr<'hir>) {
+    let mut result = vec![];
+    let root = loop {
+        e = match e.kind {
+            ExprKind::Index(ep, _) | ExprKind::Field(ep, _) => {
+                result.push(e);
+                ep
```
camsteffen, line 574:
```suggestion
        match e.kind {
            ExprKind::Index(ep, _) | ExprKind::Field(ep, _) => {
                result.push(e);
                e = ep;
```

camsteffen (changes requested):
This is looking good. Thanks for making changes. One more nit and please squash commits.

clippy_lints/src/swap.rs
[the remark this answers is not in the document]
camsteffen: Maybe but that can be left for a future enhancement.

clippy_lints/src/swap.rs
camsteffen: yep!

clippy_utils/src/lib.rs
HKalbasi: Finally. Thank you for your patience :)

camsteffen:
Can you squash the "add xor swap lint" commit as well, since that lint is removed in the next commit?

HKalbasi:
I kept it for credit of user @hrishisd. Is squash necessary?

HKalbasi:
I kept it for credit of user @hrishisd. Is squash necessary?

camsteffen:
Gotcha. Thanks!

@bors r+

PR #7509, opened 2021-07-29
Updated config value name in change log for 1.54

xFrednet:
Hey, I've noticed that the change log for 1.54 references the avoid_breaking_exported_api configuration with its Rust name, not the name it actually has in the config file. This should fix it for everyone, as the blog post references our change log on the master branch. Just a very small PR xD

changelog: none

(This is the change log xD)

r? @flip1995


flip1995:
@bors r+

Thanks!

PR #7521, opened 2021-07-31
lintcheck always copies in a fresh crate when provided with a crate path

rukai:
changelog: none

When lintcheck is run on local crates it copies the crate to `target/lintcheck/sources/crate_name` on the first run only.
Then in subsequent runs of lintcheck it reuses this same copy.
This caching behaviour makes sense when dealing with immutable crates.io releases and git commits.
However it is quite surprising that the changes to my local crate are not used when I run lintcheck.

To fix this I removed the copy, instead clippy runs directly off the provided crate folder.
I have tested this and have not observed any negative effects from doing this.
But maybe i'm missing something as im not familiar with clippy!

Alternatively we could make it copy the entire crate every run, but that seems problematic to me as multi-gigabyte target folders will take a long time to copy and wear down SSDs for developers who frequently run lintcheck.

### DIFF
--- a/lintcheck/src/main.rs
+++ b/lintcheck/src/main.rs
@@ -21,6 +21,7 @@ use clap::{App, Arg, ArgMatches};
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use walkdir::{DirEntry, WalkDir};
 
 #[cfg(not(windows))]
 const CLIPPY_DRIVER_PATH: &str = "target/debug/clippy-driver";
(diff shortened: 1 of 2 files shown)


camsteffen (approved):
LGTM. @flip1995 can you also take a look? I don't have much experience with lintcheck.

matthiaskrgr:
Hmm, I think the reason for copying the crate every time is so that lintcheck can run clippy with --fix mode without having to worry about the crates state, but I am not 100% certain.

Does that still work?

camsteffen:
Ah I didn't realize there is a `--fix` feature. Indeed that may be surprising if lintcheck actually applies fixes. I also just realized that we risk modifying `Cargo.lock`. I have a list of ideas, just brainstorming.

* Only copy the directory in fix mode
* Use [`cargo_metadata`](https://crates.io/crates/cargo_metadata) to identify target directories and avoid copying them.
* Add a `--update-paths` flag to re-copy paths.
* Tell the user if a cached path is used.
* Use `cargo --locked` to avoid updating `Cargo.lock`
* Add a `--path-direct-mode` (straw man) to run directly in paths and don't worry about modifying stuff

I do feel that copying target directories is pretty bad.

flip1995:
@matthiaskrgr is the one to ask about the lintcheck tool. So I'm passing this decision to him. :+1:

camsteffen:
@rukai I don't think this can be merged as is because we can't modify paths. Do you want to pursue any of the ideas mentioned?

rukai:
I'm happy to implement some of these ideas. But I wanted to know which direction we were taking before I started work on it.

camsteffen:
Instead of cargo metadata, I think a better option is to use gitignore. So I'd say do that and/or `--update-paths`.

camsteffen:
Ooh here's an even easier idea - look for CACHEDIR.TAG files.

rukai:
I went for the CACHEDIR.TAG approach.
Happy to try other approaches if we decide thats not enough.

The copying logic completely ignores any symlinks.

lintcheck/src/main.rs
```diff
@@ -193,32 +194,35 @@ impl CrateSource {
...
-                        crate_root.display()
-                    );
-                } else {
-                    println!("Copying {} to {}", path.display(), copy_dest.display());
+                // copy path into the dest_crate_root but skip directories that contain a CACHEDIR.TAG file.
+                // The target/ directory contains a CACHEDIR.TAG file so it is the most commonly skipped directory
+                // as a result of this filter.
+                let dest_crate_root = PathBuf::from(LINTCHECK_SOURCES).join(name);
+                println!("Copying {:?} to {:?}", path, dest_crate_root);
+                std::fs::remove_dir_all(&dest_crate_root).ok();
```
camsteffen, line 202: Unwrap if the error is not `NotFound`. Also add a `println!` for if the directory is found and removed.

camsteffen (changes requested):

camsteffen (approved):
LGTM. Can you please squash commits?

rukai:
squashed

camsteffen:
Thanks! @bors r+

PR #7522, opened 2021-08-01
Cover `Result` on `map_flatten` lint

### LINKED ISSUE #7496: map_flatten does not cover Result.map
Lint name: `map_flatten`

### I tried this code:
```rust
let a: Result<&str, ParseIntError> = Ok("1");
let c: i32 = a.map(|b| b.parse()).flatten().unwrap();
```
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=3c1eced0e47dac7f563945134ddc3bc8
### I expected to see this happen:
This should fall under lint rule `map_flatten`.
The lint should advise the user to use `andThen`, instead of the `.map(..).flatten()`

### Instead, this happened:
Result is not currently covered by the lint rule.
However, it is nearly the same as Option's map_flatten, and therefore makes sense to be included.

### Meta

- `cargo clippy -V`: e.g. clippy 0.1.53 (53cb7b09 2021-06-17)
-- Although I did run `rustup update`, I'm not sure if this is the latest dev version, but the lint rule doesn't seem to have changed in any recent versions.
- `rustc -Vv`:
```
binary: rustc
commit-date: 2021-06-17
host: x86_64-pc-windows-msvc
release: 1.53.0
LLVM version: 12.0.1
```

dswij:
Closes #7496 

changelog: `[map_flatten]` handles `Result` type

### DIFF
--- a/clippy_lints/src/methods/map_flatten.rs
+++ b/clippy_lints/src/methods/map_flatten.rs
@@ -52,18 +52,32 @@ pub(super) fn check<'tcx>(
         );
     }
 
-    // lint if caller of `.map().flatten()` is an Option
-    if is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(recv), sym::option_type) {
-        let func_snippet = snippet(cx, map_arg.span, "..");
-        let hint = format!(".and_then({})", func_snippet);
-        span_lint_and_sugg(
-            cx,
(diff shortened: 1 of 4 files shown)


llogiq:
Thank you for the PR! Bonus points for extending the test. @bors r+

PR #7525, opened 2021-08-02
Updated `define_Conf!` to support multi-line doc comments

xFrednet:
Updated the `define_Conf!` macro to support multi-line doc comments for readability. This also enables configuration documentation to have multiple paragraphs.

I've also added the `metadata-collector-lint` feature to the CI build and testing tasks. (I would think that we want this, now that we officially switched over)  

---

Now a small informal explanation what this PR changes (just for fun): 

* *Once upon a time there was a monster. It was handsome, supportive and happy to dig through the best source code it has ever seen. Spanning over hundreds of lines and reading over complete crates, it was purely marvels!*

    *However, there was one region in its territory that wasn't clean and well formatted like the rest. That was the mighty `define_Conf` macro. The monster would have cleaned it up a long time ago but, the previous ruler of this kingdom a powerful Python script was prohibiting it. But now that the old king was slain in the great battle of rust-1.54.0, everything was possible again. Our lovely monster was now able to grab its cleaning equipment and get some nice formatting into the now unprotected area.*

    *Said and done! Let this day go down in history!*

(I'll be the first to admit, that I'm a bit lost today. And I'm also procrastinating on some other discussions where I have some catching up to do... Oh, well, this was fun)

---

changelog: none

r? @flip1995

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -49,13 +49,13 @@ jobs:
         echo "LD_LIBRARY_PATH=${SYSROOT}/lib${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}" >> $GITHUB_ENV
 
     - name: Build
-      run: cargo build --features deny-warnings,internal-lints
+      run: cargo build --features deny-warnings,internal-lints,metadata-collector-lint
 
     - name: Test
-      run: cargo test --features deny-warnings,internal-lints
+      run: cargo test --features deny-warnings,internal-lints,metadata-collector-lint
 
     - name: Test clippy_lints
-      run: cargo test --features deny-warnings,internal-lints
+      run: cargo test --features deny-warnings,internal-lints,metadata-collector-lint
       working-directory: clippy_lints
 
     - name: Test rustc_tools_util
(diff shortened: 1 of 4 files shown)


.github/workflows/clippy.yml
```diff
@@ -49,13 +49,13 @@ jobs:
         echo "LD_LIBRARY_PATH=${SYSROOT}/lib${LD_LIBRARY_PATH+:${LD_LIBRARY_PATH}}" >> $GITHUB_ENV
 
     - name: Build
-      run: cargo build --features deny-warnings,internal-lints
+      run: cargo build --features deny-warnings,internal-lints,metadata-collector-lint
```
flip1995, line 52: Just to verify, this does not _run_ the metadata collection?

flip1995 (approved):
> I'll be the first to _emit_

That's a nice typo :smile:

.github/workflows/clippy.yml
xFrednet: Correct. It only includes the source code in the build. Running it still requires the `ENABLE_METADATA_COLLECTION` environment value to be set to `1`.

The following line changes `run: cargo test --features deny-warnings,internal-lints,metadata-collector-lint` will also not run the collection, as the test `dogfood::run_metadata_collection_lint()` is marked as *ignored* :upside_down_face:.

This will therefore basically just ensure that it compiles, which is already a strong insurance in Rust. ^^

flip1995:
@bors r+

Thanks!

Issue #7413, opened 2021-06-30
In some cases, clippy's `manual_map` lint suggests changes that don't compile


dcormier:
I tried this code:

```rust
use std::collections::HashMap;

fn main() {
    #[derive(Debug)]
    struct Thing {
        name: Option<&'static str>,
    }

    let map: HashMap<_, _> = vec![
        Thing { name: Some("a") },
        Thing { name: None },
        Thing { name: Some("b") },
    ]
    .into_iter()
    // Clippy suggests changing this manual map
    // https://rust-lang.github.io/rust-clippy/master/index.html#manual_map
    .filter_map(|thing| match &thing.name {
        Some(name) => Some((name.to_string(), thing)),
        None => None,
    })
    // Clippy's recommendation doesn't compile:
    // .filter_map(|thing| thing.name.as_ref().map(|name| (name.clone(), thing)))
    .collect();

    println!("{:?}", map);
}
```

[Playground.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=1b3ecd4156b5d4432193bcb16fb64f13)

I expected to see this happen: Clippy would have no complaints.

Instead, this happened: Clippy complained about the [manual map](https://rust-lang.github.io/rust-clippy/master/index.html#manual_map):

```
warning: manual implementation of `Option::map`
  --> src/main.rs:17:25
   |
17 |       .filter_map(|thing| match &thing.name {
   |  _________________________^
18 | |         Some(name) => Some((name.to_string(), thing)),
19 | |         None => None,
20 | |     })
   | |_____^ help: try this: `thing.name.as_ref().map(|name| (name.to_string(), thing))`
   |
   = note: `#[warn(clippy::manual_map)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_map
```

But the suggested change doesn't compile.

```
error[E0505]: cannot move out of `thing` because it is borrowed
  --> src/main.rs:22:49
   |
22 |     .filter_map(|thing| thing.name.as_ref().map(|name| (name.clone(), thing)))
   |                         ----------          --- ^^^^^^                ----- move occurs due to use in closure
   |                         |                   |   |
   |                         |                   |   move out of `thing` occurs here
   |                         |                   borrow later used by call
   |                         borrow of `thing.name` occurs here
```

### Meta

- `cargo clippy -V`: clippy 0.1.53 (53cb7b09 2021-06-17)
- `rustc -Vv`:
```
[...]

PR #7531, opened 2021-08-04
Manual map 7413

Jarcho:
fixes: #7413 

This only fixes the specific problem from #7413, not the general case. The full fix requires interacting with the borrow checker to determine the lifetime of all the borrows made in the function. I'll open an issue about it later.

changelog: Don't suggest using `map` when the option is borrowed in the match, and also consumed in the arm.
changelog: Locals declared within the would-be closure will not prevent the closure from being suggested in `manual_map` and `map_entry`

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -67,19 +67,20 @@ use rustc_data_structures::unhash::UnhashMap;
 use rustc_hir as hir;
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::def_id::DefId;
-use rustc_hir::hir_id::HirIdSet;
+use rustc_hir::hir_id::{HirIdMap, HirIdSet};
 use rustc_hir::intravisit::{self, walk_expr, ErasedMap, FnKind, NestedVisitorMap, Visitor};
 use rustc_hir::LangItem::{ResultErr, ResultOk};
 use rustc_hir::{
     def, Arm, BindingAnnotation, Block, Body, Constness, Destination, Expr, ExprKind, FnDecl, GenericArgs, HirId, Impl,
-    ImplItem, ImplItemKind, IsAsync, Item, ItemKind, LangItem, Local, MatchSource, Node, Param, Pat, PatKind, Path,
-    PathSegment, PrimTy, QPath, Stmt, StmtKind, TraitItem, TraitItemKind, TraitRef, TyKind, UnOp,
+    ImplItem, ImplItemKind, IsAsync, Item, ItemKind, LangItem, Local, MatchSource, Mutability, Node, Param, Pat,
+    PatKind, Path, PathSegment, PrimTy, QPath, Stmt, StmtKind, TraitItem, TraitItemKind, TraitRef, TyKind, UnOp,
 };
 use rustc_lint::{LateContext, Level, Lint, LintContext};
 use rustc_middle::hir::exports::Export;
 use rustc_middle::hir::map::Map;
 use rustc_middle::ty as rustc_ty;
-use rustc_middle::ty::{layout::IntegerExt, DefIdTree, Ty, TyCtxt, TypeFoldable};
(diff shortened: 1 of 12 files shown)


clippy_utils/src/lib.rs
```diff
@@ -559,7 +561,12 @@ pub fn trait_ref_of_method<'tcx>(cx: &LateContext<'tcx>, hir_id: HirId) -> Optio
 }
 
 /// Checks if the top level expression can be moved into a closure as is.
-pub fn can_move_expr_to_closure_no_visit(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, jump_targets: &[HirId]) -> bool {
```
camsteffen, line 562: Can you document what this function actually checks for? What are the conditions that must be met to return true?

clippy_utils/src/lib.rs
```diff
@@ -575,25 +582,115 @@ pub fn can_move_expr_to_closure_no_visit(cx: &LateContext<'tcx>, expr: &'tcx Exp
...
 }
 
-/// Checks if the expression can be moved into a closure as is.
-pub fn can_move_expr_to_closure(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) -> bool {
+/// How a local is captured by a closure
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CaptureKind {
+    Value,
+    Ref(Mutability),
+}
```
camsteffen, line 612: I'm concerned that this is basically re-implementing `ExprUseVisitor`. Have you tried using that instead? There maybe nuances there that we need to catch like `fake_read`.

Jarcho: It implements a subset of that. `fake_read` doesn't need to be considered, that's needed to adjust things for the borrow checker.

I could make use of it. It would mean running two visitors, but I doubt the speed difference would be noticeable even with the extra work `ExprUseVisitor` does.

camsteffen: I'll leave it up to you.

clippy_utils/src/lib.rs
Jarcho: Should be enough there now. The comment is now as long as the function.

clippy_utils/src/lib.rs
Jarcho: After getting it to work properly with patterns and nested closures, should have just gone with `ExprUseVisitor`. If `capture_local_usage` is something that would be used somewhere else it's worth it, but I don't see it being used.

camsteffen: Are you saying that you might switch to `ExprUseVisitor`?

Jarcho: I would say keep it as is. It's already written. It shouldn't need to touched much until captures switch to working on fields which would require changes anyways.

camsteffen (approved):
One minor code suggestion. Also please update the PR description to include a changelog entry for each lint changed. Then good to merge.

clippy_utils/src/lib.rs
```diff
@@ -642,25 +681,156 @@ pub fn can_move_expr_to_closure_no_visit(cx: &LateContext<'tcx>, expr: &'tcx Exp
...
+
+/// Given an expression referencing a local, determines how it would be captured in a closure.
+/// Note as this will walk up to parent expressions until the capture can be determined it should
+/// only be used while making a closure somewhere a value is consumed. e.g. a block, match arm, or
+/// function argument (other than a receiver).
+pub fn capture_local_usage(cx: &LateContext<'tcx>, e: &Expr<'_>) -> CaptureKind {
+    fn pat_capture_kind(cx: &LateContext<'_>, pat: &Pat<'_>) -> CaptureKind {
+        let mut capture = CaptureKind::Ref(Mutability::Not);
+        pat.each_binding(|_, id, span, _| {
```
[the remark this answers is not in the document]
Jarcho: That it can.

camsteffen:
Thanks! @bors r+

PR #7541, opened 2021-08-06
`never_loop`: suggest using an `if let` instead of a `for` loop

### LINKED ISSUE #7537: never_loop in for loop
People use a never_loop for loop for picking first element (if exists) of an iterator. It is possible without such a loop with `if let Some(elem) = x.iter().next() {` instead of `for elem in x {` but is it considered a better way? If it isn't, we should disable never_loop for `for` loops. If it is, we can provide a auto-fix suggestion so people who are not aware of this, don't disable lint.

LeSeulArtichaut:
changelog: suggest using an `if let` statement instead of a `for` loop that [`never_loop`]s

Fixes #7537, r? @camsteffen.

### DIFF
--- a/clippy_lints/src/loops/never_loop.rs
+++ b/clippy_lints/src/loops/never_loop.rs
@@ -1,13 +1,36 @@
+use super::utils::make_iterator_snippet;
 use super::NEVER_LOOP;
-use clippy_utils::diagnostics::span_lint;
-use rustc_hir::{Block, Expr, ExprKind, HirId, InlineAsmOperand, Stmt, StmtKind};
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::higher;
+use clippy_utils::source::snippet;
+use rustc_errors::Applicability;
+use rustc_hir::{Block, Expr, ExprKind, HirId, InlineAsmOperand, LoopSource, Node, Pat, Stmt, StmtKind};
 use rustc_lint::LateContext;
 use std::iter::{once, Iterator};
 
 pub(super) fn check(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
-    if let ExprKind::Loop(block, _, _, _) = expr.kind {
+    if let ExprKind::Loop(block, _, source, _) = expr.kind {
         match never_loop_block(block, expr.hir_id) {
-            NeverLoopResult::AlwaysBreak => span_lint(cx, NEVER_LOOP, expr.span, "this loop never actually loops"),
+            NeverLoopResult::AlwaysBreak => {
+                span_lint_and_then(cx, NEVER_LOOP, expr.span, "this loop never actually loops", |diag| {
+                    if_chain! {
+                        if let LoopSource::ForLoop = source;
+                        if let Some((_, Node::Expr(parent_match))) = cx.tcx.hir().parent_iter(expr.hir_id).nth(1);
+                        if let Some((pat, iterator, _, for_span)) = higher::for_loop(parent_match);
+                        then {
+                            // Suggests using an `if let` instead. This is `Unspecified` because the
+                            // loop may (probably) contain `break` statements which would be invalid
+                            // in an `if let`.
+                            diag.span_suggestion_verbose(
+                                for_span.with_hi(iterator.span.hi()),
+                                "if you need the first element of the iterator, try writing",
+                                for_to_if_let_sugg(cx, iterator, pat),
+                                Applicability::Unspecified,
+                            );
+                        }
+                    };
+                });
+            },
             NeverLoopResult::MayContinueMainLoop | NeverLoopResult::Otherwise => (),
         }
     }
@@ -169,4 +192,15 @@ fn never_loop_expr_all<'a, T: Iterator<Item = &'a Expr<'a>>>(es: &mut T, main_lo
 fn never_loop_expr_branch<'a, T: Iterator<Item = &'a Expr<'a>>>(e: &mut T, main_loop_id: HirId) -> NeverLoopResult {
     e.map(|e| never_loop_expr(e, main_loop_id))
         .fold(NeverLoopResult::AlwaysBreak, combine_branches)
+}
+
+fn for_to_if_let_sugg(cx: &LateContext<'_>, iterator: &Expr<'_>, pat: &Pat<'_>) -> String {
+    let pat_snippet = snippet(cx, pat.span, "_");
+    let iter_snippet = make_iterator_snippet(cx, iterator, &mut Applicability::Unspecified);
+
+    format!(
+        "if let Some({pat}) = {iter}.next()",
+        pat = pat_snippet,
+        iter = iter_snippet
+    )
 }
--- a/tests/ui/never_loop.stderr
+++ b/tests/ui/never_loop.stderr
@@ -75,6 +75,11 @@ LL | |             _ => return,
 LL | |         }
 LL | |     }
    | |_____^
+   |
+help: if you need the first element of the iterator, try writing
+   |
+LL |     if let Some(x) = (0..10).next() {
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 error: this loop never actually loops
   --> $DIR/never_loop.rs:157:5


tests/ui/never_loop.stderr
```diff
@@ -75,6 +75,13 @@ LL | |             _ => return,
 LL | |         }
 LL | |     }
    | |_____^
+   |
+help: if you need the first element of the iterator, try writing
+   |
+LL |     for (0..10).into_iter().next() {
```
HKalbasi, line 81: It should be `if let Some(x) = (0..10).into_iter().next() {`

HKalbasi (changes requested):

tests/ui/never_loop.stderr
LeSeulArtichaut: Oh whoops

clippy_lints/src/loops/never_loop.rs
```diff
@@ -170,3 +198,27 @@ fn never_loop_expr_branch<'a, T: Iterator<Item = &'a Expr<'a>>>(e: &mut T, main_
...
+    };
+    let iterator_snippet = snippet_opt(cx, iterator.span)?;
+    let pat_snippet = snippet(cx, pat.span, "_");
+
+    #[rustfmt::skip]
+    let sugg = format!(
+"if let Some({pat}) = {open_paren}{iterator}{close_paren}.into_iter().next() {{
+    /* ... */
+}}", pat = pat_snippet, open_paren = open_paren, iterator = iterator_snippet, close_paren = close_paren);
+    Some(reindent_multiline(sugg.into(), true, indent_of(cx, loop_expr.span)).into())
```
camsteffen, line 223: Instead of having `{ /* ... */ }` in the suggestion, can you try just replacing the "head" of the loop block? Then I think you can do something like `span_label(entire_loop_span, "in this loop")` to show more context in the output.

camsteffen, line 223: Use `has_iter_method` to suggest `iter()` instead of `into_iter()` when applicable.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -1,13 +1,41 @@
...
 
 pub(super) fn check(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
-    if let ExprKind::Loop(block, _, _, _) = expr.kind {
+    if let ExprKind::Loop(block, _, source, _) = expr.kind {
         match never_loop_block(block, expr.hir_id) {
-            NeverLoopResult::AlwaysBreak => span_lint(cx, NEVER_LOOP, expr.span, "this loop never actually loops"),
+            NeverLoopResult::AlwaysBreak => {
+                span_lint_and_then(cx, NEVER_LOOP, expr.span, "this loop never actually loops", |diag| {
+                    if_chain! {
+                        if let LoopSource::ForLoop = source;
```
camsteffen, line 18: Use `higher::for_loop` to parse the `for` loop.

camsteffen (changes requested):

clippy_lints/src/loops/never_loop.rs
LeSeulArtichaut: :o

clippy_lints/src/loops/never_loop.rs
LeSeulArtichaut: Then the suggestion won't compile if there are any `break`s or `continue` in the original loop. Is that ok?

HKalbasi: Reachable continues can't be there (because it's a never loop). I think a fallback to your original `{ /* ... */ }` can be good in precedence of `break`.

camsteffen: It's okay to make a suggestion that would require additional changes, just use `Applicability::Unspecified`. You could additionally add a note for the user pointing out `break`s but I don't think it's necessary.

LeSeulArtichaut:
@camsteffen Resolved your review comments.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -1,13 +1,37 @@
...
+                        let parent_arm = cx.tcx.hir().get_parent_node(expr.hir_id);
+                        let parent_match = cx.tcx.hir().expect_expr(cx.tcx.hir().get_parent_node(parent_arm));
+                        if let Some((pat, iterator, _, _)) = higher::for_loop(parent_match);
+                        if let Some(sugg) = for_to_if_let_sugg(cx, iterator, pat);
+                        then {
+                            diag.span_suggestion_verbose(
+                                cx.tcx.sess.source_map().guess_head_span(expr.span),
```
camsteffen, line 25: A "guess" shouldn't be necessary. Try `span.with_hi(iterator.span)` where `span` is the last element returned from `higher::for_loop`.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -1,13 +1,37 @@
...
+                                "if you need the first element of the iterator, try writing",
+                                sugg,
+                                Applicability::Unspecified,
+                            );
+                            // diag.span_label(expr.span, "in this loop");
```
camsteffen, line 30: Delete this if you decided against it. (I think that's fine)

tests/ui/never_loop.stderr
```diff
@@ -75,6 +75,11 @@ LL | |             _ => return,
 LL | |         }
 LL | |     }
    | |_____^
+   |
+help: if you need the first element of the iterator, try writing
+   |
+LL |     if let Some(x) = (0..10).into_iter().next() {
```
camsteffen, line 81: Hmm this should be `(0..10).next()`. Sorry I didn't realize this till now - we have a `make_iterator_snippet` util that should handle all three cases.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -1,13 +1,37 @@
...
+                        let parent_arm = cx.tcx.hir().get_parent_node(expr.hir_id);
+                        let parent_match = cx.tcx.hir().expect_expr(cx.tcx.hir().get_parent_node(parent_arm));
+                        if let Some((pat, iterator, _, _)) = higher::for_loop(parent_match);
+                        if let Some(sugg) = for_to_if_let_sugg(cx, iterator, pat);
+                        then {
+                            diag.span_suggestion_verbose(
+                                cx.tcx.sess.source_map().guess_head_span(expr.span),
+                                "if you need the first element of the iterator, try writing",
+                                sugg,
+                                Applicability::Unspecified,
```
camsteffen, line 28: Add a comment why this is `Unspecified`.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -1,13 +1,37 @@
...
-    if let ExprKind::Loop(block, _, _, _) = expr.kind {
+    if let ExprKind::Loop(block, _, source, _) = expr.kind {
         match never_loop_block(block, expr.hir_id) {
-            NeverLoopResult::AlwaysBreak => span_lint(cx, NEVER_LOOP, expr.span, "this loop never actually loops"),
+            NeverLoopResult::AlwaysBreak => {
+                span_lint_and_then(cx, NEVER_LOOP, expr.span, "this loop never actually loops", |diag| {
+                    if_chain! {
+                        if let LoopSource::ForLoop = source;
+                        let parent_arm = cx.tcx.hir().get_parent_node(expr.hir_id);
+                        let parent_match = cx.tcx.hir().expect_expr(cx.tcx.hir().get_parent_node(parent_arm));
```
camsteffen, line 20: This could be `cx.tcx.hir().parent_iter(expr.hir_id).nth(1)` if you like

camsteffen (changes requested):

clippy_lints/src/loops/never_loop.rs
LeSeulArtichaut: Oops, sorry. For the record: the loop is already pointed at in the primary span of the diagnostic, this just adds a label in that same span.

camsteffen:
Thanks! @bors r+

PR #7542, opened 2021-08-07
Cleanup usage of `span_to_snippet` and `LintContext::sess`

LeSeulArtichaut:
- avoid using `SourceMap::span_to_snippet` directly and use `clippy_utils::source::snippet_opt` instead
- don't use `LintContext::sess()` on `EarlyContext`s which have a `sess` field directly available, saving the import of `LintContext`

changelog: none


flip1995:
@bors r+

Thanks!

PR #7544, opened 2021-08-07
Clean up examples in new lint suggestion template

r00ster91:
I'm pretty sure they meant to write `bounds checking` when they wrote `bounce checking` but I could be wrong. After that I thought I could improve it further and ended up with this.

changelog: none


flip1995:
@bors r+

Thanks!

PR #7556, opened 2021-08-11
No effect inclusive range

F3real:
I noticed during last PR that range expression is `ExprKind::Struct` while inclusive range is `ExprKind::Call` which was why it was not handled. This PR adds check for this case.

changelog: [`no_effect]` Report inclusive range in no_effect lint


flip1995:
@bors r+

Thanks!

PR #7557, opened 2021-08-11
Add the `undocumented_unsafe_blocks` lint

### LINKED ISSUE #7464: `missing_unsafe_doc`, explain why an `unsafe` block is safe/sound
### What it does

Kinda like `missing_safety_doc`, but this lint would warn about a missing comment on each `unsafe {}` block, for which a comment must exist explaining *why* the `unsafe` block is safe/sound.

### Categories (optional)

- Kind: `style`

### What is the advantage of the recommended code over the original code

There'd be no recommended code, but the lint would suggest adding
```rs
// Safety: [...]
```
To the new code as a start. The lint would also raise a warning if that comment is copied verbatim.

### Drawbacks

This would be rather bothersome to those who know what they're doing, or only make small `unsafe` blocks all around.

### Example

```rust
pub(crate) fn coerce<T>(ffi_pointer: *const usize) -> T {
  unsafe {
      some_unchecked_function(pointer)
  }
}
```

Could be written as:

```rust
pub(crate) fn coerce<T>(ffi_pointer: *const usize) -> T {
  // Safety: function visibility is crate-internal, ffi_pointer could only be constructed in this crate.
  unsafe {
      some_unchecked_function(pointer)
  }
}
```

The following code would still raise a warning (not because of what `transmute` is doing here, but because the safety comment still has "[...]", as suggested. The developer would need to update it with an actual sentence.);
```rust
// Safety: [...]
let totally_a_usize = unsafe {
    std::mem::transmute::<String, usize>(string)
};
```

LeSeulArtichaut:
changelog: add the [`undocumented_unsafe_blocks`] lint.

Fixes #7464.

The lint was proposed as a `style` lint, I implemented it as a `restriction` lint, which is preferred?

### DIFF
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -1,0 +1,128 @@
+use clippy_utils::diagnostics::span_lint;
+use if_chain::if_chain;
+use rustc_ast::ast::{BlockCheckMode, ExprKind, UnsafeSource};
+use rustc_data_structures::fx::FxHashMap;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::{source_map::SourceMap, BytePos, FileName, Span};
+use std::convert::TryInto;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `unsafe` blocks without a `// SAFETY` comment
+    /// explaining why the unsafe operations performed inside
+    /// the block are safe.
+    ///
+    /// ### Why is this bad?
+    /// Undocumented unsafe blocks are hard to check and maintain.
+    /// On the other hand, writing safety comments helps uncovering
+    /// unsoundness holes and bugs.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # use std::ptr::NonNull;
+    /// # let a = &mut 42;
+    ///
+    /// let ptr = unsafe { NonNull::new_unchecked(a) };
+    /// ```
+    /// You should explain why calling `NonNull::new_unchecked` is safe:
+    /// ```rust
+    /// # use std::ptr::NonNull;
+    /// # let a = &mut 42;
+    ///
+    /// // SAFETY: references are guaranteed to be non-null.
+    /// let ptr = unsafe { NonNull::new_unchecked(a) };
+    /// ```
+    pub UNDOCUMENTED_UNSAFE_BLOCKS,
+    restriction,
+    "unsafe blocks without safety comments"
+}
+
+struct SafetyComment {
+    hi: BytePos,
+    hi_line: usize,
+}
+
+#[derive(Default)]
+pub struct UndocumentedUnsafeBlocks {
+    safety_comments: FxHashMap<FileName, Option<Vec<SafetyComment>>>,
+}
+
+impl UndocumentedUnsafeBlocks {
+    fn safety_comments(&mut self, sm: &SourceMap, span: Span) -> Option<&mut Vec<SafetyComment>> {
+        let file = sm.span_to_filename(span);
+        self.safety_comments
+            .entry(file.clone())
+            .or_insert_with(|| Self::gather_safety_comments(sm, &file))
+            .as_mut()
+    }
+
+    // Inspired by `rustc_ast::utils::comments::gather_comments`.
+    fn gather_safety_comments(sm: &SourceMap, file: &FileName) -> Option<Vec<SafetyComment>> {
+        let source_file = sm.get_source_file(file)?;
+        let src = source_file.src.as_deref()?;
+        let start_bpos = source_file.start_pos;
+        let mut comments = Vec::new();
+
+        let mut inside_comment = false;
+        let mut pos = rustc_lexer::strip_shebang(src).unwrap_or(0);
+        for token in rustc_lexer::tokenize(&src[pos..]) {
+            match token.kind {
+                rustc_lexer::TokenKind::LineComment { doc_style: None }
+                | rustc_lexer::TokenKind::BlockComment {
+                    doc_style: None,
+                    terminated: true,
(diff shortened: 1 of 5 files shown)


clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,128 @@
...
+            pos += token.len;
+        }
+        Some(comments)
+    }
+}
+
+impl_lint_pass!(UndocumentedUnsafeBlocks => [UNDOCUMENTED_UNSAFE_BLOCKS]);
+
+impl EarlyLintPass for UndocumentedUnsafeBlocks {
+    fn check_expr(&mut self, cx: &EarlyContext<'_>, expr: &rustc_ast::Expr) {
```
flip1995, line 101: There's also a [`check_block`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/passes/trait.EarlyLintPass.html#method.check_block) method.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,128 @@
...
+    // Inspired by `rustc_ast::utils::comments::gather_comments`.
+    fn gather_safety_comments(sm: &SourceMap, file: &FileName) -> Option<Vec<SafetyComment>> {
+        let source_file = sm.get_source_file(file)?;
+        let src = source_file.src.as_deref()?;
+        let start_bpos = source_file.start_pos;
+        let mut comments = Vec::new();
+
+        let mut inside_comment = false;
+        let mut pos = rustc_lexer::strip_shebang(src).unwrap_or(0);
+        for token in rustc_lexer::tokenize(&src[pos..]) {
```
flip1995, line 69: Since we don't really care about correct and complete syntax in the file, but only finding comments immediately above the `unsafe` block, you should be able to optimize this. I think you can use the [`get_enclosing_scope`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/struct.Map.html#method.get_enclosing_scope) in combination with the [`opt_span`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/struct.Map.html#method.opt_span) methods and then slice the to-tokenize `src` with the start of the enclosing scope up until the start of the block.

To do this, you would have to turn this lint into a `LatePassLint`, but that is a tradeoff I'm willing to take in order to not re-lex whole files just to find a comment, of which we already know where it _should_ be.

To make sure that just slicing the `src` works as expected, please add a test case with unicode chars in the comment and the code above the `unsafe` block. 

---

With this change, you also might be able to just directly return if the checked block has a `SAFETY` comment right above it, without having to do `BytePos` and `span.hi()/.lo()` gymnastics. Basically you have to track if after lexing the suggested part of the file, the last thing you saw was a comment and if it was a `SAFETY` comment.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+        Some(comments)
+    }
+}
+
+impl_lint_pass!(UndocumentedUnsafeBlocks => [UNDOCUMENTED_UNSAFE_BLOCKS]);
+
+impl EarlyLintPass for UndocumentedUnsafeBlocks {
+    fn check_expr(&mut self, cx: &EarlyContext<'_>, expr: &rustc_ast::Expr) {
+        let (block, comments) = if_chain! {
```
flip1995, line 102: I think you can add a `in_external_macro` check here, since this lint shouldn't be triggered for those.

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,74 @@
+#![warn(clippy::undocumented_unsafe_blocks)]
```
flip1995, line 1: I'd like to see some tests of how this lint behaves in macros.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,128 @@
...
+            // we're searching for is likely to be at the beginning of the vector,
+            // so this is better than a binary search.
+            if let Some((i, comment)) = comments.iter().enumerate().take_while(|(_, c)| c.hi <= block.span.lo()).last();
+            let block_line = cx.sess.source_map().lookup_char_pos(block.span.lo()).line;
+            if block_line == comment.hi_line + 1 || block_line == comment.hi_line;
+            then {
+                comments.remove(i);
+            }
+            else {
+                span_lint(cx, UNDOCUMENTED_UNSAFE_BLOCKS, expr.span, "this `unsafe` block is missing a SAFETY comment");
```
flip1995, line 124: I think we should produce a help message here, suggesting to add a `// SAFETY: ...`  above the `unsafe` block.

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,74 @@
...
+
+#[rustfmt::skip]
+fn good_inline_block() {
+    /* SAFETY */ unsafe {}
+}
+
+fn good_statement() {
+    let _ =
+        // SAFETY
+        unsafe {};
```
flip1995, line 39: There's a test missing for the example in the documentation:

```rs
// SAFETY: bla
let _ = unsafe {};
```

flip1995 (changes requested):
I have some concerns when it comes to performance with this lint. With the current implementation, this lint will potentially re-lex every file in the crate, if there are unsafe blocks.

I agree that this lint fits best in `restriction`, but I would also add a 
```rs
if is_lint_allowed(...) {
    return;
}
```
to the top of the `check_expr`/`check_block` method. That and my suggestions below should deal with the performance impact.

giraffate:
ping from triage @LeSeulArtichaut. Can you have any update on this?

giraffate:
ping from triage @LeSeulArtichaut. According to the triage procedure, I'm closing this because 2 weeks have passed with no activity. If you have more time to work on this, feel free to reopen this.

PR #7562, opened 2021-08-13
Fix false positive on `filter_next`

### LINKED ISSUE #7561: filter_next false positive
Lint name:
filter_next

I tried this code:

```rust
struct Foo;
impl Foo {
    fn filter(&self, _spec: i32) -> std::vec::IntoIter<i32> {
        unimplemented!()
    }
}

fn main() {
    let empty = Foo.filter(42).next().is_none();
    println!("{}", empty);
}
```

I expected to see this happen: Clippy doesn't suggest anything

Instead, this happened:
Clippy says this:
```
warning: called `filter(..).next()` on an `Iterator`. This is more succinctly expressed by calling `.find(..)` instead
 --> src/main.rs:9:17
  |
9 |     let empty = Foo.filter(42).next().is_none();
  |                 ^^^^^^^^^^^^^^^^^^^^^ help: try this: `Foo.find(42)`
  |
```

There is no find method on `Foo`.

### Meta

clippy 0.1.56 (ccffcaf 2021-08-11)

  ```
rustc 1.56.0-nightly (ccffcafd5 2021-08-11)
binary: rustc
commit-hash: ccffcafd55e58f769d4b0efc0064bf65e76998e4
commit-date: 2021-08-11
release: 1.56.0-nightly
LLVM version: 12.0.1
  ```

dswij:
fixes #7561

changelog: Fix false positive on [`filter_next`] when a method does not implement `Iterator`

### DIFF
--- a/clippy_lints/src/methods/filter_next.rs
+++ b/clippy_lints/src/methods/filter_next.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg};
-use clippy_utils::is_trait_method;
 use clippy_utils::source::snippet;
+use clippy_utils::ty::implements_trait;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_lint::LateContext;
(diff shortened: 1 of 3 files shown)


dswij:
Seems like [the docs is still suggesting](https://github.com/rust-lang/rust-clippy/blob/b4d76b42fd12d963eddafb7e1d9b8559951f73be/doc/common_tools_writing_lints.md#checking-if-a-type-implements-a-specific-trait) to use `match_trait_method`, which is both deprecated and is prone to false positive such as the one above.

I think it would be good if the docs explicitly explains why `is_trait_method` is not the preferred way.

xFrednet:
Oh, you're right. We should update that documentation, would you like to do that? :upside_down_face: 

Diagnostic items in general are sadly not as good documented as they should. There is an issue for that in rust-lang/rustc-dev-guide#1188 to cover the entire topic. (I plan to work on it in the future when I have a bit more time again)

xFrednet (approved):
Looks good to me. Thank you for the fix! :upside_down_face:

xFrednet:
If you would like to do it, it would be good to do it in another PR to also get general feedback on the documentation update. Having it in this PR could let it fly under the radar.

@bors r+

PR #7565, opened 2021-08-14
New lint `manual_split_once`

Jarcho:
This is a WIP because it still needs to recognize more patterns. Currently handles:

```rust
s.splitn(2, ' ').next();
s.splitn(2, ' ').nth(0)
s.splitn(2, ' ').nth(1);
s.splitn(2, ' ').skip(0).next();
s.splitn(2, ' ').skip(1).next();
s.splitn(2, ' ').next_tuple(); // from itertools

// as well as `unwrap()` and `?` forms
```

Still to do:

```rust
let mut iter = s.splitn(2, ' ');
(iter.next().unwrap(), iter.next()?)

let mut iter = s.splitn(2, ' ');
let key = iter.next().unwrap();
let value = iter.next()?;
```

Suggestions on other patterns to check for would be useful. I've done a search on github for uses of `splitn`. Still have yet to actually look through the results.

There's also the question of whether the lint shouold trigger on all uses of `splitn` with two values, or only on recognized usages. The former could have false positives where it couldn't be replaced, but I'm not sure how common that would be.

changelog: Add lint `manual_split_once`


tests/ui/manual_split_once.stderr
```diff
@@ -0,0 +1,34 @@
+error: manual implementation of `split_once`
+  --> $DIR/manual_split_once.rs:13:13
+   |
+LL |     let _ = "key=value".splitn(2, '=').nth(0).unwrap();
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `"key=value".split_once('=').unwrap().0`
```
mikerite, line 5: This suggestion isn't the same in general.
`"no_equals".splitn(2, '=').nth(0).unwrap()` returns `"no_equals"`.
`"no_equals".split_once('=').unwrap().0` panics.

Jarcho: `map_or` is suggested now.

llogiq:
This already looks quite comprehensive! Great job! I think it'd be ready to merge and we can extend it with a later PR if you're OK with that.

My suggestion is informed by

* The already high code quality :+1:
* The rather low risk of false positives
* The helpful suggestions as showcased by the tests

Jarcho:
Just updated the docs as they mentioned something that currently isn't checked for. I see nothing wrong with merging it as is. It will probably be a week before I get to finishing the rest of the checks.

llogiq:
Thank you! @bors r+

xFrednet:
It seems like #7566 which was merged shortly before this, again uses the `LocalUsedVisitor` visitor

llogiq:
Let's try that again – @bors r+

PR #7566, opened 2021-08-14
Check expr usage for  `manual_flatten`

### LINKED ISSUE #6784: manual_flatten suggestin make code not compilable
Lint name: `manual_flatten`

I tried this code:
Code is simplified so `workers` are never set, but to trigger the incorrect sugestion it is enough.
The full code starts same with all workers doing nothing, and later in the loop assign work to them, and here early in the loop let them finish the work if it is the right time.
```rust
fn main() {
    let mut time = 0;
    let mut workers : [Option<(u8, i32)>; 5] = [None, None, None, None, None];
    loop {
        for w in &mut workers {
            if let Some((_to, when)) = w {
                if time == *when {
                    // ...
                    *w = None;
                }
            }
        }
        // ...
        time += 1;
    }
}
```

I expected to see this happen: No suggestion probably.

Instead, this happened: 
```
warning: unnecessary `if let` since only the `Some` variant of the iterator element is used
  --> src\main.rs:6:9
   |
6  |           for w in &mut workers {
   |           ^        ------------ help: try: `workers.iter_mut().flatten()`
   |  _________|
   | |
7  | |             if let Some((_to, when)) = w {
8  | |                 if time == *when {
9  | |                     *w = None;
10 | |                 }
11 | |             }
12 | |         }
   | |_________^
   |
   = note: `#[warn(clippy::manual_flatten)]` on by default
help: ...and remove the `if let` statement in the for loop
  --> src\main.rs:7:13
   |
7  | /             if let Some((_to, when)) = w {
8  | |                 if time == *when {
9  | |                     *w = None;
10 | |                 }
11 | |             }
   | |_____________^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_flatten

warning: 1 warning emitted
```
In this simplified example line 9 `*w = None;` is visible.
Am I missing a way how to handle this line? I asked on Rust community server and no one find a way how to make this work, so I think it must be false positive.

### Meta

- `cargo clippy -V`: clippy 0.1.52 (fe1bf8e0 2021-02-23)
- `rustc -Vv`:
```
rustc 1.52.0-nightly (fe1bf8e05 2021-02-23)
binary: rustc
commit-hash: fe1bf8e05c39bdcc73fc09e246b7209444e389bc
commit-date: 2021-02-23
host: x86_64-pc-windows-msvc
release: 1.52.0-nightly
LLVM version: 11.0.1
```

### LINKED ISSUE #7538: `manual_flatten` suggests code that does not compile
Lint name:

`manual_flatten`

I tried this code:

```rust
for server in self.servers.iter_mut() {
    if let Some(server_inner) = server {
        if server_inner.pid == target_pid {
            *server = None;
        }
    }
}
```

I expected to see this happen: The lint should not have been emitted. Alternately, the suggestion should have resulted in code that compiled.

Instead, this happened: The `Quick fix` resulted in code that did not compile.

### Meta

- `cargo clippy -V`: clippy 0.1.54 (a178d032 2021-07-26)
- `rustc -Vv`:
  ```
  binary: rustc
  commit-date: 2021-07-26
  host: x86_64-pc-windows-msvc
  release: 1.54.0
  LLVM version: 12.0.1
  ```

dswij:
Fixes #6784
Fixes #7538 

`manual_flatten` should not trigger when `if let` match expression will be used.

changelog: [`manual_flatten`] checks for expr usage after `if let`

### DIFF
--- a/clippy_lints/src/loops/manual_flatten.rs
+++ b/clippy_lints/src/loops/manual_flatten.rs
@@ -1,6 +1,7 @@
 use super::utils::make_iterator_snippet;
 use super::MANUAL_FLATTEN;
 use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::visitors::LocalUsedVisitor;
 use clippy_utils::{is_lang_ctor, path_to_local_id};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
@@ -37,16 +38,18 @@ pub(super) fn check<'tcx>(
         if_chain! {
             if let Some(inner_expr) = inner_expr;
             if let ExprKind::Match(
-                match_expr, match_arms, MatchSource::IfLetDesugar{ contains_else_clause: false }
+                match_expr, [true_arm, _else_arm], MatchSource::IfLetDesugar{ contains_else_clause: false }
             ) = inner_expr.kind;
             // Ensure match_expr in `if let` statement is the same as the pat from the for-loop
             if let PatKind::Binding(_, pat_hir_id, _, _) = pat.kind;
             if path_to_local_id(match_expr, pat_hir_id);
             // Ensure the `if let` statement is for the `Some` variant of `Option` or the `Ok` variant of `Result`
-            if let PatKind::TupleStruct(ref qpath, _, _) = match_arms[0].pat.kind;
+            if let PatKind::TupleStruct(ref qpath, _, _) = true_arm.pat.kind;
             let some_ctor = is_lang_ctor(cx, qpath, OptionSome);
             let ok_ctor = is_lang_ctor(cx, qpath, ResultOk);
             if some_ctor || ok_ctor;
+            // Ensure epxr in `if let` is not used afterwards
+            if !LocalUsedVisitor::new(cx, pat_hir_id).check_arm(true_arm);
             then {
                 let if_let_type = if some_ctor { "Some" } else { "Ok" };
                 // Prepare the error message
--- a/tests/ui/manual_flatten.rs
+++ b/tests/ui/manual_flatten.rs
@@ -91,6 +91,19 @@ fn main() {
         }
     }
 
+    struct Test {
+        a: usize,
+    }
+
+    let mut vec_of_struct = [Some(Test { a: 1 }), None];
+
+    // Usage of `if let` expression should not trigger lint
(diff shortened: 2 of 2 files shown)


clippy_lints/src/loops/manual_flatten.rs
```diff
@@ -47,6 +48,9 @@ pub(super) fn check<'tcx>(
             let some_ctor = is_lang_ctor(cx, qpath, OptionSome);
             let ok_ctor = is_lang_ctor(cx, qpath, ResultOk);
             if some_ctor || ok_ctor;
+            // Ensure epxr in `if let` is not used afterwards
+            let mut used_visitor = LocalUsedVisitor::new(cx, pat_hir_id);
+            if !match_arms.iter().any(|arm| used_visitor.check_arm(arm));
```
xFrednet, line 53: Small NIT: The usage of an iterator can be avoided here by deconstructing the slice in line 40ff like this:
```rs
if let ExprKind::Match(
    match_expr, [true_arm, _else_arm], MatchSource::IfLetDesugar{ contains_else_clause: false }
) = inner_expr.kind;
```

Then it should be possible to directly call `if !LocalUsedVisitor::new(cx, pat_hir_id).check_arm(true_arm);` and the direct index call in line 47 can also be avoided.

---

The rest looks good to me. I'm happy to merge this afterwards :upside_down_face: 

r? @xFrednet

dswij: Ah that was a nice one. I should re-emphasize Rust's powerful pattern matching in my bible :)

xFrednet (approved):
Awesome work and bonus points for having a Rust bible :P

clippy_lints/src/loops/manual_flatten.rs
xFrednet: Looking at this, it's actually surprising that we don't have a lint for this. I'll create am issue, even if this might ignore macros as most lints do :upside_down_face:

xFrednet: Created #7569 and also liked this PR as an example how this improves code quality. I also suggested to lint inside `if_chain!` macros to catch these cases inside Clippy. Thank you again for the contributing and inspiration :upside_down_face:

Issue #7567, opened 2021-08-14
False positive in option_if_let_else inside const fn


dtolnay:
```rust
#![deny(clippy::option_if_let_else)]

pub const fn f() -> Option<i8> {
    None
}

pub const fn g() -> i8 {
    if let Some(x) = f() {
        1 + x
    } else {
        0
    }
}
```

```console
$ cargo clippy

error: use Option::map_or instead of an if let/else
  --> src/main.rs:8:5
   |
8  | /     if let Some(x) = f() {
9  | |         1 + x
10 | |     } else {
11 | |         0
12 | |     }
   | |_____^ help: try: `f().map_or(0, |x| 1 + x)`
```

However, Clippy's suggested replacement code does not compile.

```console
error[E0015]: calls in constant functions are limited to constant functions, tuple structs and tuple variants
 --> src/main.rs:8:5
  |
8 |     f().map_or(0, |x| 1 + x)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^
```

Originally reported by @lopopolo in https://github.com/rust-lang/rust-clippy/issues/6137#issuecomment-705605688.

### Meta

- `cargo clippy -V`: clippy 0.1.56 (0fa3190 2021-08-12)
- `rustc -Vv`:
  ```
  rustc 1.56.0-nightly (0fa319039 2021-08-12)
  binary: rustc
  commit-hash: 0fa3190394475a84360b34e074e719d519bc40f1
  commit-date: 2021-08-12
  host: x86_64-unknown-linux-gnu
  release: 1.56.0-nightly
  LLVM version: 12.0.1
  ```

Issue #6737, opened 2021-02-13
Invalid suggestion of `option_if_let_else` causes a compilation error `error[E0382]: Borrow of moved value`


xFrednet:
I tried this code:

```rust
#[allow(dead_code, clippy::unnecessary_wraps)]
#[deny(clippy::option_if_let_else)]
fn test_fn(acc: Option<String>) -> Option<String> {
    // Computation of mv_value
    let mv_value = "Hello clippy".to_string();
    
    if let Some(mut x) = acc {
        x.push_str(&mv_value);
        Some(x)
    } else {
        Some(mv_value)
    }
}
```

and got the following lint trigger:

```
error: use Option::map_or instead of an if let/else
  --> src/main.rs:8:5
   |
8  | /     if let Some(mut x) = acc {
9  | |         x.push_str(&mv_value);
10 | |         Some(x)
11 | |     } else {
12 | |         Some(mv_value)
13 | |     }
   | |_____^
   |
note: the lint level is defined here
  --> src/main.rs:3:8
   |
3  | #[deny(clippy::option_if_let_else)]
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
help: try
   |
8  |     acc.map_or(Some(mv_value), |mut x| {
9  |         x.push_str(&mv_value);
10 |         Some(x)
11 |     })
   |
```

Using the suggested code leads to:

```
error[E0382]: borrow of moved value: `mv_value`
  --> src/main.rs:16:32
   |
6  |     let mv_value = "Hello clippy".to_string();
   |         -------- move occurs because `mv_value` has type `String`, which does not implement the `Copy` trait
...
16 |     acc.map_or(Some(mv_value), |mut x| {
   |                     --------   ^^^^^^^ value borrowed here after move
   |                     |
   |                     value moved here
17 |         x.push_str(&mv_value);
   |                     -------- borrow occurs due to use in closure
```

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4bdec68d896548d9e398613a10bf0e30)

The `if let` is used in this case to use `mv_value` in both branches. This is a limitation of the `map_or` suggestion.

### Meta

- `cargo clippy -V`: clippy 0.1.51 (3682750 2021-02-02)
- `rustc -Vv`:
    ```
    rustc 1.51.0-nightly (368275062 2021-02-02)
[...]

Issue #6137, opened 2020-10-08
option_if_let_else suggestion does not compile


lopopolo:
I tried this code:

```rust
#![warn(clippy::all)]
#![warn(clippy::pedantic)]

use core::convert::TryFrom;

#[derive(Clone, Copy)]
struct X;

struct Foo;

impl From<X> for Option<Foo> {
    fn from(_: X) -> Self {
        Some(Foo)
    }
}

#[derive(Clone)]
struct Bar;

impl From<X> for Option<Bar> {
    fn from(_: X) -> Self {
        Some(Bar)
    }
}

pub struct NoBlockGiven(X);

impl From<X> for NoBlockGiven {
    fn from(value: X) -> Self {
        Self(value)
    }
}

impl TryFrom<X> for Bar {
    type Error = NoBlockGiven;

    fn try_from(value: X) -> Result<Self, Self::Error> {
        if let Some(block) = value.into() {
            Ok(block)
        } else {
            Err(NoBlockGiven::from(value))
        }
    }
}
```

Clippy made this suggestion:

```console
$ cargo clippy
    Checking clippy-test v0.1.0 (/Users/lopopolo/Downloads/clippy-test)
warning: use Option::map_or instead of an if let/else
  --> src/lib.rs:38:9
   |
38 | /         if let Some(block) = value.into() {
39 | |             Ok(block)
40 | |         } else {
41 | |             Err(NoBlockGiven::from(value))
42 | |         }
   | |_________^ help: try: `value.into().map_or(Err(NoBlockGiven::from(value)), |block| Ok(block))`
   |
note: the lint level is defined here
  --> src/lib.rs:2:9
   |
2  | #![warn(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::option_if_let_else)]` implied by `#[warn(clippy::pedantic)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else

warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
```

Applying the fix results in code that does not compile:

```rust
#![warn(clippy::all)]
#![warn(clippy::pedantic)]

use core::convert::TryFrom;

#[derive(Clone, Copy)]
struct X;

struct Foo;

impl From<X> for Option<Foo> {
    fn from(_: X) -> Self {
        Some(Foo)
    }
}

#[derive(Clone)]
struct Bar;

[...]

Issue #5822, opened 2020-07-19
option_if_let_else suggestion inapplicable when mutable borrow or owned value is used in both branches


Arnavion:
```rust
#[deny(clippy::option_if_let_else)]
pub fn foo(o: Option<i32>, mut f: impl FnMut(i32)) {
    if let Some(i) = o {
        f(i)
    }
    else {
        f(5)
    }
}
```

Apart from having the same problem as #5821, in this particular case even the suggestion of `o.map_or_else(|| f(5), |i| f(i))` would be wrong because `f` is being used as a `&mut`, so it cannot be used with two closures at the same time. This also happens for `f: impl FnOnce(i32)`.

I don't think there's any better way to write this code using any of the `Option` combinators, so it seems to be the lint should not fire at all.

### Meta

- `cargo clippy -V`: `clippy 0.0.212 (39d5a61 2020-07-17)`
- `rustc -Vv`:
  ```
  rustc 1.47.0-nightly (39d5a61f2 2020-07-17)
  binary: rustc
  commit-hash: 39d5a61f2e4e237123837f5162cc275c2fd7e625
  commit-date: 2020-07-17
  host: x86_64-unknown-linux-gnu
  release: 1.47.0-nightly
  LLVM version: 10.0
  ```

rail-rain:
I've found a real world example inside Clippy where presumably this bug led to an incorrect use of `crate::util::snippet_with_applicability`, which slipped into the code base [along with the lint itself](https://github.com/rust-lang/rust-clippy/pull/5301/commits/5150277a4f765bb113e163adc7eb495dcbb57129). I'm not sure whether we should immediately revert the change with `#[allow(clippy::option_if_let_else)]` for the time being or can leave it until this issue gets fixed though.

https://github.com/rust-lang/rust-clippy/blob/23db5426e6c90a18700e6bf30af537e5bf6b14c7/clippy_lints/src/write.rs#L299-L306

ebroto:
Nice catch! I would say let's allow the lint for now and fix that.

BTW wouldn't this be a nice candidate for a lint? I can't think now of a case where you actually want to pass a `&mut` to a temporary to a function.

rail-rain:
OK, I made a PR for it (and it's just quickly merged while I'm writing this). Thanks for the reply!

Regarding an new lint, it makes sense for me; considering a function taking a mutable reference probably means the reference is some kind of an "out" parameter.

brightly-salty:
[...]

PR #7568, opened 2021-08-14
Downgrade option_if_let_else to nursery

dtolnay:
I believe that this lint's loose understanding of ownership (#5822, #6737) makes it unsuitable to be enabled by default in its current state, even as a pedantic lint.

Additionally the lint has known problems with type inference (#6137), though I may be willing to consider this a non-blocker in isolation if it weren't for the ownership false positives.

A fourth false positive involving const fn: #7567.

But on top of these, for me the biggest issue is I basically fully agree with https://github.com/rust-lang/rust-clippy/issues/6137#issuecomment-705605688. In my experience this lint universally makes code worse even when the resulting code does compile.

---

changelog: remove [`option_if_let_else`] from default set of enabled lints


llogiq:
Thank you! @bors r+

PR #7570, opened 2021-08-16
Add the `derivable_impls` lint

### LINKED ISSUE #7550: Detect redundant `Default` implementations
### What it does

Detects custom `std::default::Default` implementations that are identical to a derived implementation. This could probably be done with other derive attributes, but I think `Default` is a fairly common one that people may implement themselves.

### Categories (optional)

- Kind: `clippy::style`

It is more concise.

### Drawbacks

I have no clue about the internals of Clippy, so I'm not sure how hard it would be to implement.

### Example

```rust
struct Foo {
    bar: bool
}

impl std::default::Default for Foo {
    fn default() -> Self {
        Self {
            bar: false
        }
    }
}
```

Could be written as:

```rust
#[derive(Default)]
struct Foo {
    bar: bool
}
```

HKalbasi:
Fix #7550 

changelog: Add new derivable_impls lint. mem_replace_with_default now covers non constructor cases.

### DIFF
--- a/clippy_lints/src/derivable_impls.rs
+++ b/clippy_lints/src/derivable_impls.rs
@@ -1,0 +1,108 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::{in_macro, is_automatically_derived, is_default_equivalent, remove_blocks};
+use rustc_hir::{
+    def::{DefKind, Res},
+    Body, Expr, ExprKind, Impl, ImplItemKind, Item, ItemKind, Node, QPath,
+};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::TypeFoldable;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects manual `std::default::Default` implementations that are identical to a derived implementation.
+    ///
+    /// ### Why is this bad?
+    /// It is less concise.
+    ///
+    /// ### Example
+    /// ```rust
+    /// struct Foo {
+    ///     bar: bool
+    /// }
+    ///
+    /// impl std::default::Default for Foo {
+    ///     fn default() -> Self {
+    ///         Self {
+    ///             bar: false
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    ///
+    /// Could be written as:
+    ///
+    /// ```rust
+    /// #[derive(Default)]
+    /// struct Foo {
+    ///     bar: bool
+    /// }
+    /// ```
+    ///
+    /// ### Known problems
+    /// Derive macros [sometimes use incorrect bounds](https://github.com/rust-lang/rust/issues/26925)
+    /// in generic types and the user defined `impl` maybe is more generalized or
+    /// specialized than what derive will produce. This lint can't detect the manual `impl`
+    /// has exactly equal bounds, and therefore this lint is disabled for types with
+    /// generic parameters.
+    ///
+    pub DERIVABLE_IMPLS,
+    complexity,
+    "manual implementation of the `Default` trait which is equal to a derive"
+}
+
+declare_lint_pass!(DerivableImpls => [DERIVABLE_IMPLS]);
+
+fn is_path_self(e: &Expr<'_>) -> bool {
+    if let ExprKind::Path(QPath::Resolved(_, p)) = e.kind {
+        matches!(p.res, Res::SelfCtor(..) | Res::Def(DefKind::Ctor(..), _))
+    } else {
+        false
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for DerivableImpls {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'_>) {
+        if_chain! {
+            if let ItemKind::Impl(Impl {
+                of_trait: Some(ref trait_ref),
+                items: [child],
+                ..
+            }) = item.kind;
+            if let attrs = cx.tcx.hir().attrs(item.hir_id());
+            if !is_automatically_derived(attrs);
+            if !in_macro(item.span);
+            if let Some(def_id) = trait_ref.trait_def_id();
+            if cx.tcx.is_diagnostic_item(sym::Default, def_id);
+            if let impl_item_hir = child.id.hir_id();
+            if let Some(Node::ImplItem(impl_item)) = cx.tcx.hir().find(impl_item_hir);
+            if let ImplItemKind::Fn(_, b) = &impl_item.kind;
+            if let Body { value: func_expr, .. } = cx.tcx.hir().body(*b);
+            if let Some(adt_def) = cx.tcx.type_of(item.def_id).ty_adt_def();
+            then {
+                if cx.tcx.type_of(item.def_id).definitely_has_param_types_or_consts(cx.tcx) {
+                    return;
+                }
+                let should_emit = match remove_blocks(func_expr).kind {
+                    ExprKind::Tup(fields) => fields.iter().all(|e| is_default_equivalent(cx, e)),
+                    ExprKind::Call(callee, args)
+                        if is_path_self(callee) => args.iter().all(|e| is_default_equivalent(cx, e)),
+                    ExprKind::Struct(_, fields, _) => fields.iter().all(|ef| is_default_equivalent(cx, ef.expr)),
+                    _ => false,
+                };
+                if should_emit {
+                    let path_string = cx.tcx.def_path_str(adt_def.did);
+                    span_lint_and_help(
+                        cx,
+                        DERIVABLE_IMPLS,
+                        item.span,
+                        "this `impl` can be derived",
+                        None,
+                        &format!("try annotating `{}` with `#[derive(Default)]`", path_string),
+                    );
+                }
+            }
+        }
+    }
+}
--- a/clippy_lints/src/mem_replace.rs
+++ b/clippy_lints/src/mem_replace.rs
@@ -1,9 +1,9 @@
 use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::source::{snippet, snippet_with_applicability};
-use clippy_utils::{in_macro, is_diag_trait_item, is_lang_ctor, match_def_path, meets_msrv, msrvs, paths};
+use clippy_utils::ty::is_non_aggregate_primitive_type;
+use clippy_utils::{in_macro, is_default_equivalent, is_lang_ctor, match_def_path, meets_msrv, msrvs, paths};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::def_id::DefId;
 use rustc_hir::LangItem::OptionNone;
 use rustc_hir::{BorrowKind, Expr, ExprKind, Mutability, QPath};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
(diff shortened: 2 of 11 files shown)


camsteffen:
You can use `is_default_equivalent_ctor`. Otherwise I think you got it covered. Careful not to lint macros where a dynamically generated impl just happens to be Default-equivalent.

HKalbasi:
I added `is_default_equivalent_ctor` and some more cases to `is_default`.

> Careful not to lint macros where a dynamically generated impl just happens to be Default-equivalent.

I think it should be ok, do you have special test in mind that I can add it?

r? @camsteffen

HKalbasi:
Seems my changes in `mem-replace-with-default` are catching this in dogfood: `std::mem::replace(&mut self.is_break, false);` in `redundant_else.rs:123`. I think it should be `self.is_break = false;` and `replace` (or `take`) are pointless. Am I missing something?

camsteffen:
The lint is technically correct because the result of `mem::replace` is the return value of the function. Although I don't really like the suggestion in that case. I think we should allow `mem::replace` for primitive types.

HKalbasi:
Or maybe just for boolean? For integers take isn't better?

camsteffen:
`mem::take` makes the *most* sense when you have something with heap allocation like a `Vec` where you are literally taking the heap-allocated data from the `Vec`. `mem::take` with integers is okay but (at least personally) I don't think it is better than `mem::replace(n, 0)`. Allowing primitives is the simplest way to draw the line here.

camsteffen:
> > Careful not to lint macros where a dynamically generated impl just happens to be Default-equivalent.
> 
> I think it should be ok, do you have special test in mind that I can add it?

Here's a couple tests - [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b329bc6a895c3b77608704b564618aa0)

HKalbasi:
I disabled that lint for primitives. But it is still enable for stack allocated things like `(0, false)`.

@rustbot label +S-waiting-on-review

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+fn is_path_self(e: &Expr<'_>) -> bool {
+    if let ExprKind::Path(QPath::Resolved(_, p)) = e.kind {
+        if p.segments.len() != 1 {
+            return false;
+        }
+        if p.segments[0].ident.name == rustc_span::symbol::kw::SelfUpper {
+            return true;
+        }
```
camsteffen, line 51: You can do `matches!(path.res, Res::SelfTy(..))`

tests/ui/derivable_impls.rs
```diff
@@ -0,0 +1,121 @@
...
+    c: u64,
+    d: Vec<i32>,
+    e: FooND1,
+    f: FooND2,
+    g: HashMap<i32, i32>,
+    h: (i32, Vec<i32>),
+    i: [Vec<i32>; 3],
+    j: [i32; 5],
+    k: Option<i32>,
+}
```
camsteffen, line 15: How about a slice?

clippy_utils/src/lib.rs
```diff
@@ -606,6 +606,62 @@ pub fn can_mut_borrow_both(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>) -
...
+                        .iter()
+                        .any(|&symbol| cx.tcx.is_diagnostic_item(symbol, adt.did));
+                }
+            }
+        }
+    }
+    false
+}
+
+pub fn is_default(cx: &LateContext<'_>, e: &Expr<'_>) -> bool {
```
camsteffen, line 638: Please add docs. `is_default_equivalent` would be better since this function does not only check for `Default::default()`.

clippy_lints/src/mem_replace.rs
```diff
@@ -194,64 +193,36 @@ fn check_replace_with_uninit(cx: &LateContext<'_>, src: &Expr<'_>, dest: &Expr<'
...
+    }
+    // disable lint for Option since it is covered in another lint
+    if let ExprKind::Path(q) = &src.kind {
+        if is_lang_ctor(cx, q, OptionNone) {
+            return;
         }
     }
-    false
-}
+    if is_default(cx, src) && !in_external_macro(cx.tcx.sess, expr_span) {
```
camsteffen, line 207: Why are the changes to this lint needed?

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,114 @@
+use clippy_utils::diagnostics::span_lint;
+use clippy_utils::{in_macro, is_automatically_derived, is_default, match_def_path, paths};
+use rustc_hir::{Block, Expr, ExprField, ExprKind, Impl, ImplItemKind, Item, ItemKind, Node, QPath};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for double comparisons that could be simplified to a single expression.
```
camsteffen, line 9: Looks like all the docs are copied from another lint.

camsteffen (changes requested):

clippy_utils/src/lib.rs
```diff
@@ -606,6 +606,61 @@ pub fn can_mut_borrow_both(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>) -
...
+    match &e.kind {
+        ExprKind::Lit(lit) => match lit.node {
+            LitKind::Bool(x) => !x,
+            LitKind::Int(x, _) => x == 0,
+            LitKind::Str(s, _) => s.is_empty(),
+            _ => false,
+        },
```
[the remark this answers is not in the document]
HKalbasi: S isn't `&str`, but `Symbol`. I merged two first ones:
```rust
ExprKind::Lit(lit) => match lit.node {
            LitKind::Bool(false) | LitKind::Int(0, _) => true,
            LitKind::Str(s, _) => s.is_empty(),
            _ => false,
        },
```

clippy_lints/src/mem_replace.rs
HKalbasi: It now covers more cases like `mem::replace(&mut x, (vec![], HashMap::new());`.

clippy_lints/src/derivable_impls.rs
HKalbasi: Ah, sorry, I forgot to write docs.

tests/ui/derivable_impls.rs
HKalbasi: My current `is_default_equivalent` doesn't detect slice. Is it generally true that `&(T::default())` is equal to `(&T)::default()`?

camsteffen: I don't think so. But you can see the Default impl [in the docs] (https://doc.rust-lang.org/std/primitive.slice.html#impl-Default).

clippy_lints/src/derivable_impls.rs
HKalbasi: It doesn't pass tests

camsteffen:
Whoops I somehow assumed this was already assigned to me.

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,88 @@
...
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects custom `std::default::Default` implementations that are identical to a derived implementation.
+    ///
+    ///
+    /// ### Why is this bad?
+    /// It is more concise.
```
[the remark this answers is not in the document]
HKalbasi: hmm, I blindly copied it from original issue :)

clippy_lints/src/mem_replace.rs
```diff
@@ -194,64 +193,36 @@ fn check_replace_with_uninit(cx: &LateContext<'_>, src: &Expr<'_>, dest: &Expr<'
...
-                    return std_types_symbols
-                        .iter()
-                        .any(|&symbol| cx.tcx.is_diagnostic_item(symbol, adt.did));
-                }
-            }
+fn check_replace_with_default(cx: &LateContext<'_>, src: &Expr<'_>, dest: &Expr<'_>, expr_span: Span) {
+    // disable lint for primitives
+    if let ExprKind::Lit(_) = src.kind {
+        return;
+    }
```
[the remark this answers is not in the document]
HKalbasi: I done it but not sure if it is the best line we can draw. Specially I don't like disabling lint for slices and `&str`.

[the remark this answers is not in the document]
camsteffen: Hmm good point. `&str`, slices and tuples are considered "aggregate data types", so we could lint those? I think that also removes "recursively" from the picture so just `is_non_aggregate_primitive`.

@rust-lang/clippy any thoughts? We currently only lint things like `mem::replace(x, Vec::new())` where there is some constructor call or `Default::default()`. I think suggesting `mem::take` instead of `mem::replace(x, false)` or `0` is weird so I suggested in my previous comment above to exclude all primitives. But now I'm thinking to exclude "non-aggregate primitives".

camsteffen:
We should watch out for cases where a manual impl is needed because a derive adds different type bounds (https://github.com/rust-lang/rust/issues/26925). For example, a struct with `Option<T>` does not require `T: Default`, but a derive adds that type bound anyways. A simple approach would be to just ignore impls with any type variables. Or, unless I'm missing something, I think it would be correct to lint if all type variables have the `Default` constraint. In any case, we should probably mention this issue in the lint docs.

clippy_lints/src/mem_replace.rs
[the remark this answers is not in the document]
Manishearth: i think that could work!

HKalbasi:
I now disabled lint for non-lifetime generics, because in addition to that issue, specialized impls can create false positives, and detection of them is not always easy:
```rust
struct SpecializedImpl<A, B> {
    a: A,
    b: B,
}

impl<T: Default> Default for SpecializedImpl<T, T> {
    fn default() -> Self {
        Self {
            a: T::default(),
            b: T::default(),
        }
    }
}
```
Not sure if there is real use case for specialized impls of default trait, but `#[derive(Default)]` behave differently in this case.

camsteffen:
Wouldn't that example be the same as derive?

HKalbasi:
No? I think derive will generate this:
```rust
impl<A: Default, B: Default> Default for SpecializedImpl<A, B> {
    fn default() -> Self {
        Self {
            a: A::default(),
            b: B::default(),
        }
    }
}
```

clippy_lints/src/mem_replace.rs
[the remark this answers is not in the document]
camsteffen: @HKalbasi does that sound good to you?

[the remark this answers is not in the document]
HKalbasi: Yes it is good, and `if let ExprKind::Lit(_) = src.kind` was doing the same but it depend on the implementation of `is_default_equivalent`. I didn't find `is_non_aggregate_primitive`. If it doesn't exist with another name, I should add this if I understand correctly in `ty.rs`:
```rs
pub fn is_non_aggregate_primitive_type(ty: Ty<'_>) -> bool {
    matches!(ty.kind(), ty::Bool | ty::Char | ty::Int(_) | ty::Uint(_) | ty::Float(_))
}
```
Is it covers all non aggregate primitives? And another option is putting `matches!` directly in if condition.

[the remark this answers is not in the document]
camsteffen: > Yes it is good, and `if let ExprKind::Lit(_) = src.kind` was doing the same but it depend on the implementation of `is_default_equivalent`.

I think `ExprKind::Lit` would exclude string literals.

> If it doesn't exist with another name, I should add this if I understand correctly in `ty.rs`:

Looks good. I think it should be a util.

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,113 @@
...
+                    for arg in generic.args {
+                        if !matches!(arg, GenericArg::Lifetime(_)) {
+                            return;
+                        }
+                    }
+                }
+                let should_emit = match remove_blocks(func_expr).kind {
+                    ExprKind::Tup(fields) => fields.iter().all(|e| is_default_equivalent(cx, e)),
+                    ExprKind::Call(callee, args)
+                        if is_path_self(callee) => args.iter().all(|e| is_default_equivalent(cx, e)),
```
camsteffen, line 88: It looks like the lint will not work if `Self` is replaced with the type name (should be tested). Would it work to simply remove the `is_path_self` check?

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,113 @@
...
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects manual `std::default::Default` implementations that are identical to a derived implementation.
+    ///
+    /// ### Why is this bad?
+    /// It is less concise.
+    ///
+    /// ### Example
```
camsteffen, line 17: Please add a `Known problems` section to indicate that impls with type parameters are not linted and why.

camsteffen (changes requested):
A couple more things and this is almost ready! Can you be a little more specific for the mem_replace_with_default changelog?

clippy_lints/src/derivable_impls.rs
HKalbasi: > Would it work to simply remove the is_path_self check?

No because any function call can `callee` be. Now I check it against `self_ty` and added test.

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,121 @@
...
+                ..
+            } = self_ty;
+            then {
+                if let Some(generic) = self_ty_path.segments.last().and_then(|s| s.args) {
+                    for arg in generic.args {
+                        if !matches!(arg, GenericArg::Lifetime(_)) {
+                            return;
+                        }
+                    }
+                }
```
camsteffen, line 92: This check actually should be `cx.tcx.type_of(item.def_id).has_param_types_or_consts()`. And this is changed to `definitely_has_param_types_or_consts` with the next Rustup. It looks like the Rustup PR will be merged soon so we should rebase after that.

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub DERIVABLE_IMPLS,
+    complexity,
+    "manual implementation of the `Default` trait which is equal to a derive"
+}
+
+declare_lint_pass!(DerivableImpls => [DERIVABLE_IMPLS]);
+
+fn is_path_self(cx: &LateContext<'_>, e: &Expr<'_>, self_path: &Path<'_>) -> bool {
+    if let ExprKind::Path(QPath::Resolved(_, p)) = e.kind {
+        matches!(p.res, Res::SelfCtor(..)) || SpanlessEq::new(cx).eq_path(p, self_path)
```
camsteffen, line 57: `SpanlessEq` shouldn't be necessary and might be incorrect if the path is qualified differently. Can you just check for `Res::Def(DefKind::Ctor(..), _)`? You could also do a check on the DefId there but I don't think that is needed since a constructor call must be the constructor of the type being returned.

clippy_lints/src/derivable_impls.rs
```diff
@@ -0,0 +1,121 @@
...
+                };
+                if should_emit {
+                    let path_string = self_ty_path
+                        .segments
+                        .iter()
+                        .map(|segment| {
+                            segment.ident.to_string()
+                        })
+                        .collect::<Vec<_>>()
+                        .join("::");
```
camsteffen, line 108: Can you use `cx.tcx.def_path_str` with the `DefId` from `cx.tcx.type_of(item.def_id).ty_adt_def()` here?

camsteffen (changes requested):
Sorry, found a couple more things.

clippy_lints/src/derivable_impls.rs
HKalbasi: I can't find these functions. I think rustup PR is merged and I rebased but that is renamed to something else.

camsteffen: Check TypeFoldable

camsteffen:
Thanks!

@bors r+

HKalbasi:
changelog: is there, what is wrong?

camsteffen:
Oh it has to be on the same line like `changelog: <stuff>`

HKalbasi:
Like this? Please edit it in the desired way if it is wrong.

PR #7573, opened 2021-08-16
Fix `option_if_let_else`

Jarcho:
fixes: #5822
fixes: #6737
fixes: #7567

The inference from #6137 still exists so I'm not sure if this should be moved from the nursery. Before doing that though I'd almost want to see this split into two lints. One suggesting `map_or` and the other suggesting `map_or_else`.

`map_or_else` tends to have longer expressions for both branches so it doesn't end up much shorter than a match expression in practice. It also seems most people find it harder to read. `map_or` at least has the terseness benefit of being on one line most of the time, especially when the `None` branch is just a literal or path expression.

changelog: `break` and `continue` statments local to the would-be closure are allowed in `option_if_let_else`
changelog: don't lint in const contexts  in `option_if_let_else`
changelog: don't lint when yield expressions are used  in `option_if_let_else`
changelog: don't lint when the captures made by the would-be closure conflict with the other branch  in `option_if_let_else`
changelog: don't lint when a field of a local is used when the type could be pontentially moved from  in `option_if_let_else`
changelog: in some cases, don't lint when scrutinee expression conflicts with the captures of the would-be closure  in `option_if_let_else`

### DIFF
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -174,9 +174,9 @@ fn detect_option_if_let_else<'tcx>(
                     ExprKind::Field(e, _) | ExprKind::AddrOf(_, _, e) => Some(e),
                     _ => None,
                 });
-                if let ExprKind::Path(QPath::Resolved(None, Path { res: Res::Local(l), .. })) = e.kind {
-                    match some_captures.get(l)
-                        .or_else(|| (method_sugg == "map_or_else").then(|| ()).and_then(|_| none_captures.get(l)))
+                if let ExprKind::Path(QPath::Resolved(None, Path { res: Res::Local(local_id), .. })) = e.kind {
+                    match some_captures.get(local_id)
+                        .or_else(|| (method_sugg == "map_or_else").then(|| ()).and_then(|_| none_captures.get(local_id)))
                     {
                         Some(CaptureKind::Value | CaptureKind::Ref(Mutability::Mut)) => return None,
                         Some(CaptureKind::Ref(Mutability::Not)) if as_mut => return None,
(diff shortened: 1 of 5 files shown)


clippy_lints/src/option_if_let_else.rs
```diff
@@ -153,6 +166,24 @@ fn detect_option_if_let_else<'tcx>(
...
             };
+            // Check if captures the closure will need conflict with borrows made in the scrutinee.
+            // TODO: check all the references made in the scrutinee expression. This will require interacting
+            // with the borrow checker. Currently only `<local>[.<field>]*` is checked for.
+            if as_ref || as_mut {
+                let e = peel_hir_expr_while(cond_expr, |e| match e.kind {
+                    ExprKind::Field(e, _) | ExprKind::AddrOf(_, _, e) => Some(e),
+                    _ => None,
+                });
+                if let ExprKind::Path(QPath::Resolved(None, Path { res: Res::Local(l), .. })) = e.kind {
```
giraffate, line 177: nits
```suggestion
                if let ExprKind::Path(QPath::Resolved(None, Path { res: Res::Local(hir_id), .. })) = e.kind {
```

clippy_lints/src/option_if_let_else.rs
```diff
@@ -153,6 +166,24 @@ fn detect_option_if_let_else<'tcx>(
...
+                    match some_captures.get(l)
+                        .or_else(|| (method_sugg == "map_or_else").then(|| ()).and_then(|_| none_captures.get(l)))
+                    {
+                        Some(CaptureKind::Value | CaptureKind::Ref(Mutability::Mut)) => return None,
+                        Some(CaptureKind::Ref(Mutability::Not)) if as_mut => return None,
```
giraffate, line 182: Is there a test for this case?

giraffate (commented):
Sorry for the late review.

> changelog: don't lint when yield expressions are used in `option_if_let_else`

What kind of case is this? Is there a test for this?

giraffate:
Additionally, it would be helpful if you split the PR or commits for each issue next time. I'm okay with that this time.

giraffate:
@bors r+

Thanks!

PR #7583, opened 2021-08-18
Tweak common tool docs on type-implement-trait check

dswij:
See https://github.com/rust-lang/rust-clippy/pull/7562#issuecomment-898240963.

changelog: none


doc/common_tools_writing_lints.md
```diff
@@ -102,8 +103,7 @@ impl LateLintPass<'_> for MyStructLint {
 ```
 
 > Prefer using lang items, if the target trait is available there.
-
-A list of defined paths for Clippy can be found in [paths.rs][paths]
+> `is_trait_method` checks for method call and can introduce unwanted behaviours.
```
xFrednet, line 106: What do you exactly mean with the second part that it "can introduce unwanted behaviours"? :sweat_smile:

xFrednet (commented):
Thank you for the documentation update :+1:

doc/common_tools_writing_lints.md
dswij: Good point. I wanted to refer to the problem in #7562, where the original caller does not implement a trait, but calls a method that returns a type that implements the trait, it could sometimes lead to false positive, and hence using lang items is preferred.

But I have some trouble expressing this 😅, figured I would open a PR to ask for help

I have thought of moving it to a new section for `checking if a method call belongs to a trait`, if that's better?

xFrednet: Ahh, yes! I've mixed up two methods, this makes sense! What do you think about this for line 105 and 106?

```md
Clippy can also use paths to check for types and trait methods. Please prefer diagnostic items
where possible, as paths can sometimes be unreliable. The diagnostic item of a trait or struct can be found
using the following steps:
1. Open the type documentation (Example: [`Iterator`])
2. Open the source code for the type in the documentation.
3. Find the `rustc_diagnostic_item` attribute (Example: `#[rustc_diagnostic_item = "Iterator"]`)

[`Iterator`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html
```

Rendered:

---

Clippy can also use paths to check for types and trait methods. Please prefer diagnostic items
where possible, as paths can sometimes be unreliable. The diagnostic item of a trait or struct can be found
using the following steps:
1. Open the type documentation (Example: [`Iterator`])
2. Open the source code for the type in the documentation.
3. Find the `rustc_diagnostic_item` attribute (Example: `#[rustc_diagnostic_item = "Iterator"]`)

[`Iterator`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html

----

I think that would cover that statement and give some additional information on how to find diagnostic items.

xFrednet: > I have thought of moving it to a new section for `checking if a method call belongs to a trait`, if that's better?

Sorry, I skipped over this at first. I think it's valid to keep it here for now. In the future, we should move this into the general documentation of diagnostic items, but until then I would leave it here :upside_down_face:

dswij: > Clippy can also use paths to check for types and trait methods.

While this is how `match_trait_method` works, I don't think this applies to `is_trait_method` since `is_trait_method` is using diagnostic item, if I'm not misunderstanding it ?

https://github.com/rust-lang/rust-clippy/blob/1fc1aee7be6e7e71f12eacd60b9b1040e7640fe4/clippy_utils/src/lib.rs#L327-L332

xFrednet: Oh yes, you're correct. Sorry, I mixed some stuff up :sweat_smile: .

I would say that the function is working as intended. The bug you fixed in #7562 originated from the fact that `is_trait_method` was only called on the second expression of two consecutive calls. Saying that it can introduce unwanted behavior seems a bit too much. I would either leave it out or state that: when checking for sequential, the diagnostic item has to checked for each expression and not just the last one. (Worded a bit differently)

doc/common_tools_writing_lints.md
```diff
@@ -78,14 +78,15 @@ impl LateLintPass<'_> for MyStructLint {
 There are two ways to do this, depending if the target trait is part of lang items.
 
 ```rust
-use clippy_utils::{implements_trait, match_trait_method, paths};
+use clippy_utils::{implements_trait, is_trait_method};
+use rustc_span::symbol::sym;
 
 impl LateLintPass<'_> for MyStructLint {
     fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
         // 1. Using expression and Clippy's convenient method
```
xFrednet, line 86: These labels should also be updated to reflect what they are actually using, IMO. What do you think about these:
```rs
// 1. Using diagnostic items with the expression
// 2. Using lang items with the expression type
// 3. Using the type path with the expression
```

doc/common_tools_writing_lints.md
```diff
@@ -102,8 +103,7 @@ impl LateLintPass<'_> for MyStructLint {
 ```
 
 > Prefer using lang items, if the target trait is available there.
```
xFrednet, line 105:
```suggestion
> Prefer using diagnostic and lang items, if the target trait has one.
```

doc/common_tools_writing_lints.md
```diff
@@ -102,8 +103,7 @@ impl LateLintPass<'_> for MyStructLint {
...
-
-A list of defined paths for Clippy can be found in [paths.rs][paths]
+> Note that using `is_trait_method` requires checking each expression on sequentials.
```
xFrednet, line 106: Correct me if this is wrong, but this is not specific to `is_trait_method` but just a result of sequential calls, right? In that case, maybe something like this:
```suggestion
> Note that linting a sequence of trait calls requires checking each expression for the trait.
```

xFrednet (commented):
Thank you so much for the update. This time I read the whole section and found some NITs where it would be great if you could look at them too :upside_down_face:

doc/common_tools_writing_lints.md
dswij: I'd say that we do not need to check each expression, but rather it depends on the context?

After re-reading this, this seems to give more confusion rather than clarity. I suppose we can omit this one, what do you think @xFrednet

xFrednet: I agree! So, sounds good to me :+1:

PR #7584, opened 2021-08-18
Extend unnecessary_unwrap to look for expect in addition to unwrap

### LINKED ISSUE #7581: Extend unnecessary_unwrap to look for `expect` in addition to `unwrap`
### What it does

Applies the same logic as `unwrap` to `expect`.

```rust
fn main() {
    let t = Some(0);
    if t.is_some() {
        // let _v = t.unwrap(); // reported by clippy::unnecessary_unwrap
        let _v = t.expect("x"); // not reported
    }
}
```

### Categories

- Kind: complexity

### Drawbacks

None.

### Example

```rust
fn main() {
    let t = Some(0);
    if t.is_some() {
        let _v = t.expect("x");
    }
}
```

Could be written as:

```rust
fn main() {
    let t = Some(0);
    if let Some(_v) = t {}
}
```

shepmaster:
changelog: Extend ``[`unnecessary_unwrap`]`` to also check for `Option::expect` and `Result::expect`. Also give code suggestions in some cases.

Fixes #7581

### DIFF
--- a/clippy_lints/src/unwrap.rs
+++ b/clippy_lints/src/unwrap.rs
@@ -1,10 +1,11 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::higher;
 use clippy_utils::ty::is_type_diagnostic_item;
-use clippy_utils::{differing_macro_contexts, usage::is_potentially_mutated};
+use clippy_utils::{differing_macro_contexts, path_to_local, usage::is_potentially_mutated};
 use if_chain::if_chain;
+use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, walk_fn, FnKind, NestedVisitorMap, Visitor};
-use rustc_hir::{BinOpKind, Body, Expr, ExprKind, FnDecl, HirId, Path, QPath, UnOp};
+use rustc_hir::{BinOpKind, Body, Expr, ExprKind, FnDecl, HirId, PathSegment, UnOp};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::map::Map;
 use rustc_middle::lint::in_external_macro;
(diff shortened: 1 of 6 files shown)


tests/ui/unnecessary_unwrap.stderr
```diff
@@ -0,0 +1,36 @@
+error: you checked before that `unwrap()` cannot fail, instead of checking and unwrapping, it's better to use `if let` or `match`
```
camsteffen, line 1: Tests should be added to existing tests at `simple_conditionals.rs` or `complex_conditionals.rs`.

camsteffen, line 1: Do you mind adjusting the output for this lint while you're here? I find it hard to read currently. Something like this would be nice:

```text
error: called `unwrap` on `opt` after checking its variant with `is_some`
  --> $DIR/unnecessary_unwrap.rs:11:17
   |
LL |     if opt.is_some() {
   |        ------------- the variant is checked here
LL |         let _ = opt.expect("an error message");
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `expect` is called here
help: try using `if let` or `match` instead
```
An actual code suggestion would be even better, but not necessary.

camsteffen (changes requested):

tests/ui/unnecessary_unwrap.stderr
shepmaster: >  to existing tests at simple_conditionals.rs or complex_conditionals.rs

As a first-time contributor, it was not easy to find these tests (and you can see that I _didn't_).

camsteffen: I understand. Feel free to add a comment pointing to the tests. Personally I look for tests by searching all files for "you checked before that", which is definitely weird.

camsteffen:
You bumped into the stderr length. I think you can remove some of your tests. It isn't necessary to add an `expect` test for every `unwrap` test.

tests/ui/unnecessary_unwrap.stderr
shepmaster: I think that means I need something like 

```rust
&format!(
    "called `{}` on `{}` after checking its variant with `{}`",
    method_name.ident.name,
    unwrappable.ident,
    unwrappable.check,
),
```

However, `Path` and `Expr` do not implement `Display`, and I don't see anything obvious to get what I want. Any clues on where to look for examples?

shepmaster: Alright, a revised version is up.

> An actual code suggestion would be even better

Any tips on where to look for how to do that?

camsteffen: You would need to add `diag.span_suggestion(span, suggestion)` to the closure in `span_lint_and_then`. The `span` would be from the `Expr` of `opt.is_some()` and the suggestion would be like `"let Some(value) = opt"`.

shepmaster:
> > An actual code suggestion would be even better
> 
> Any tips on where to look for how to do that?

I haven't found any examples of how to suggest that a line of code be deleted.

```rust
if __option__.is_some() {
    let __var__ = __option__.unwrap();
}
```

Should be suggested to

```rust
if let Some(__var__) = __option__ {
}
```

So I'd need to:

1. track the target variable name (if there is one)
1. track if it's an `Option` or `Result`
1. write the if into an if-let, preserving the body
1. remove the redundant `unwrap`/`expect` expression / statement

shepmaster:
> You bumped into the stderr length limit. I think you can remove some of your tests.

Looks like the `help` caused this to happen again — any preferred tests I should remove this round?

camsteffen:
For the code suggestion, I think it's okay to keep it simple and just make a suggestion to replace the `if` condition like `if let Some(..) = __option__`. So that is leaving it to the user to fill in a variable name and make other code changes. The suggestion would be `Applicability::Unspecified`. Let me know if you need more help with that.

> Looks like the `help` caused this to happen again — any preferred tests I should remove this round?

I'm seeing if we can increase that that limit ([Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Increase.20test.20stderr.20length.20limit.3F)). If not, please split the test into another file.

camsteffen:
The stderr limit is removed! #7593

shepmaster:
Ok, I added a commit with a potential suggestion implementation (I was thinking you wanted auto-fixable, which is why I was a bit 😰).

If you like it, I can squash the second and third commits.

tests/ui/checked_unwrap/complex_conditionals.stderr
```diff
@@ -179,11 +179,11 @@ LL |     if x.is_ok() || !(y.is_ok() && z.is_err()) {
 LL |         z.unwrap(); // will panic
    |         ^^^^^^^^^^
 
-error: you checked before that `unwrap_err()` cannot fail, instead of checking and unwrapping, it's better to use `if let` or `match`
+error: called `unwrap_err` on `z` after checking its variant with `is_err`
   --> $DIR/complex_conditionals.rs:50:9
    |
 LL |     if x.is_ok() || !(y.is_ok() && z.is_err()) {
-   |                                    ---------- the check is happening here
+   |                                    ---------- help: try using `if let` or `match`: `if let Ok(..) = z`
```
camsteffen, line 186: This should suggest `if let Err(..)`

camsteffen (changes requested):

clippy_lints/src/unwrap.rs
```diff
@@ -170,24 +200,41 @@ impl<'a, 'tcx> Visitor<'tcx> for UnwrappableVariablesVisitor<'a, 'tcx> {
...
+                            &format!(
+                                "called `{}` on `{}` after checking its variant with `{}`",
+                                method_name.ident.name,
+                                unwrappable_variable_name,
+                                unwrappable.check_name.ident.as_str(),
+                            ),
+                            |diag| {
+                                diag.span_suggestion(
+                                    unwrappable.check.span,
+                                    "try using `if let` or `match`",
```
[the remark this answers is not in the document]
shepmaster: > it's a bit noisy/redundant

Agree, but that loses the "or `match`" part — not that important?

shepmaster:
All right, pushed up a new version. Let me know if you like it. If so, I'll squash the last two commits.

shepmaster:
Addressed!

camsteffen:
Thanks! @bors r+

PR #7585, opened 2021-08-18
README: Replace regular title with a logo

### LINKED ISSUE #3158: Logo proposal
Hello there, i am a graphic designer and open-source is my habit lol. I would love to contribute a logo to clippy that's if you don't already have one.

Turbo87:
Resolves #3158 by adding a logo to the README file. It is a combination of the official Rust logo and https://icon-icons.com/icon/clippy/136771 (licensed under Apache 2.0), plus "rust-clippy" in FiraSans font.

clippy

see https://github.com/Turbo87/rust-clippy/tree/logo#readme

changelog: none


Turbo87:
favicon

@xFrednet this is how it would look like at a size of 48px

Turbo87:
and at 32px and 16px:

favicon32

favicon16

favicons do support SVG files these days though, so on high resolution screens they should look better than this 😅

xFrednet:
Thank you! That should work :+1:

camsteffen:
Do we need to check in with someone from Rust because this logo is derivative of the Rust logo? See https://www.rust-lang.org/policies/media-guide

Manishearth:
Yeah, someone  on the clippy team should email trademark@rust-lang.org, cc core@rust-lang.org and clippy@rust-lang.org. We are using the trademark _within_ the rust project (and be careful to specify that) so I suspect the trademark itself is not a huge deal, but picking "a logo", even if just for the repo, might be.

xFrednet:
Summary / Collection of data:
* From #3158:
  > https://office-watch.com/2021/why-has-microsoft-finally-trademarked-clippy/
  > the image is trademarked (the name itself, also I checked, is not)
* The trademark states:
  > "The mark consists of a drawing of a paperclip with eyeballs and eyebrows intended to look like a face."
  >
  > Source [justia](https://trademarks.justia.com/907/82/n-90782739.html)
* Summary off related [Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202021-08-24/near/250499160):
  > "We like the logo, but have to figure out legal stuff."

---

@rustbot label -S-waiting-on-review +S-needs-discussion

Turbo87:
let's reopen this if we get a positive response at some point :)

flip1995:
I still haven't heard from the trademark team.

flip1995:
I got a response from the foundation about the logo (thanks Rebecca Rumbul!). It isn't the response we hoped for, but the response we expected:

> We cannot provide official legal advice, but think that the proposed use of a Clippy image based/adapted from the Microsoft Clippy logo is likely to fall outside of 'reasonable use'. The Foundation is concerned that if legal action was pursued against the use of this image/character in future, then a full rebrand would be required, placing additional work on those invested in this project. 

They also suggested that it may be an option to contact Microsoft directly about this. I would expect though, that this would lead to a lot of bureaucracy and legal back and forth, so I'm not really convinced that we should try this.

---

Thanks for your work @Turbo87! Even though I think that the logo is really great, I also think that because of the project name we won't be able to have such a nice logo in the future :neutral_face:

xFrednet:
Hmm, that's a bit disappointing, even if as you said expected. Maybe we could use a logo that just has a paperclip without the eyes and smile :thinking:. That also sounds like a stretch, even if that should not be covered by the trademark. Sadly, that would look less friendly in comparison to this design.

flip1995:
> Maybe we could use a logo that just has a paperclip without the eyes and smile

With that we would still kind of imply that this Clippy has something to do with Microsofts Clippy, which is also not allowed by the "don'ts" section in the link you posted.

xFrednet:
I sadly agree with that -.-

Turbo87:
thanks for following up on this, @flip1995. sad to see this not happening, but I guess it was indeed expected. :)

PR #7596, opened 2021-08-24
New lint: option_needless_deref

### LINKED ISSUE #7571: Add lint checking for no-op uses of `Option::{as_deref,as_deref_mut}`
### What it does

Checks for no-op uses of `Option::{as_deref,as_deref_mut}`.

### Categories (optional)

- Kind: `complexity`

Removes useless code, removing clutter and improving readability.

### Drawbacks

None.

### Example

```rust
let a = Some(&1);
let b = a.as_deref(); // goes from Option<&i32> to Option<&i32>
```

Could be written as:

```rust
let a = Some(&1);
let b = a;
```

lengyijun:
changelog: [`option_needless_deref`]
fix #7571

### DIFF
--- a/clippy_lints/src/needless_option_as_deref.rs
+++ b/clippy_lints/src/needless_option_as_deref.rs
@@ -38,10 +38,7 @@ declare_lint_pass!(OptionNeedlessDeref=> [
 
 impl<'tcx> LateLintPass<'tcx> for OptionNeedlessDeref {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
-        if expr.span.from_expansion() {
-            return;
-        }
-        if in_macro(expr.span) {
+        if expr.span.from_expansion() || in_macro(expr.span) {
             return;
         }
         let typeck = cx.typeck_results();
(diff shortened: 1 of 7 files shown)


camsteffen:
I think the lint name should include `as_deref`. `needless_as_deref` would be good IMO.

lengyijun:
`option_needless_as_deref` or `needless_as_deref` ?

camsteffen:
`needless_option_as_deref` if anything. But I'd pick `needless_as_deref`.

lengyijun:
I did some fix.

`Result` also has a `as_deref` method, but this lint doesn't work on it. 
So I temporally rename to `needless_option_as_deref`.

clippy_lints/src/needless_option_as_deref.rs
```diff
@@ -0,0 +1,69 @@
...
+]);
+
+impl<'tcx> LateLintPass<'tcx> for OptionNeedlessDeref {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if expr.span.from_expansion() {
+            return;
+        }
+        if in_macro(expr.span) {
+            return;
+        }
```
llogiq, line 46:
```suggestion
        if expr.span.from_expansion() || in_macro(expr.span) {
            return;
        }
```

llogiq (approved):
I only have a small nit, otherwise this looks good.

llogiq:
Thank you! @bors r+

PR #7605, opened 2021-08-26
Issue 7548 fix

### LINKED ISSUE #7548: bool_assert_comparison false positive when one side is non-bool with PartialEq<bool> impl
Lint name: bool_assert_comparison

When testing `serde_json::Value` values, or more generally more-complex-than-bool types that implement `PartialEq<bool>`, `assert_eq!` testing should not be linted against.

### Example

```rust
// serde_json = "1"

let val: serde_json::Value = serde_json::from_str(r#"{"bar": true, "baz": 123}"#).unwrap();

assert_eq!(val["bar"], true);
assert_eq!(val["baz"], false);
```

These trigger _bool_assert_comparison_ lint
```
 --> src/main.rs:5:5
  |
5 |     assert_eq!(val["baz"], false);
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `assert!(..)`
```

However, these aren't trivially fixable with `assert!(..)`. Moreover, converting to a bool for testing is worse code as it provides less info in a test failure scenario. For example, running the example code tells us not only that the 2nd assert fails, but exactly what the actual value was.

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `Number(123)`,
 right: `false`', src/main.rs:5:5
```

A workaround like `assert_eq!(val["baz"].as_bool(), Some(false))` would make this test code worse, and does not seem to be the intention of this lint.

### Meta

- `cargo clippy -V`: `clippy 0.1.56 (ad981d5 2021-08-08)`
- `rustc -Vv`:
  ```
  rustc 1.56.0-nightly (ad981d58e 2021-08-08)
  binary: rustc
  commit-hash: ad981d58e1ca16bcf4072577934630deb11c5e14
  commit-date: 2021-08-08
  release: 1.56.0-nightly
  LLVM version: 12.0.1
  ```

xordi:
Close #7548 

changelog: [`bool_assert_comparison`] fixes should be emitted only in case they are comparing a value of a type that implements the `Not` trait with an output of type `bool` against a boolean literal.

### DIFF
--- a/clippy_lints/src/bool_assert_comparison.rs
+++ b/clippy_lints/src/bool_assert_comparison.rs
@@ -40,23 +40,23 @@ fn is_bool_lit(e: &Expr<'_>) -> bool {
     ) && !e.span.from_expansion()
 }
 
-fn impl_not_trait_with_bool_out(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> bool {
+fn is_impl_not_trait_with_bool_out(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> bool {
     let ty = cx.typeck_results().expr_ty(e);
 
     cx.tcx
         .lang_items()
         .not_trait()
-        .filter(|id| implements_trait(cx, ty, *id, &[]))
-        .and_then(|id| {
-            cx.tcx.associated_items(id).find_by_name_and_kind(
+        .filter(|trait_id| implements_trait(cx, ty, *trait_id, &[]))
+        .and_then(|trait_id| {
+            cx.tcx.associated_items(trait_id).find_by_name_and_kind(
                 cx.tcx,
                 Ident::from_str("Output"),
                 ty::AssocKind::Type,
-                id,
+                trait_id,
             )
         })
-        .map_or(false, |item| {
-            let proj = cx.tcx.mk_projection(item.def_id, cx.tcx.mk_substs_trait(ty, &[]));
+        .map_or(false, |assoc_item| {
+            let proj = cx.tcx.mk_projection(assoc_item.def_id, cx.tcx.mk_substs_trait(ty, &[]));
             let nty = cx.tcx.normalize_erasing_regions(cx.param_env, proj);
 
             nty.is_bool()
@@ -82,7 +82,7 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
                             return;
                         }
 
-                        if !impl_not_trait_with_bool_out(cx, a) || !impl_not_trait_with_bool_out(cx, b) {
+                        if !is_impl_not_trait_with_bool_out(cx, a) || !is_impl_not_trait_with_bool_out(cx, b) {
                             // At this point the expression which is not a boolean
                             // literal does not implement Not trait with a bool output,
                             // so we cannot suggest to rewrite our code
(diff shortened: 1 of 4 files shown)


clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -28,45 +30,77 @@ declare_clippy_lint! {
...
-            kind: LitKind::Bool(_),
+            node: LitKind::Bool(_),
             ..
         })
     ) && !e.span.from_expansion()
 }
 
-impl EarlyLintPass for BoolAssertComparison {
-    fn check_expr(&mut self, cx: &EarlyContext<'_>, e: &Expr) {
+fn impl_not_trait_with_bool_out(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> bool {
```
giraffate, line 43: nits
```suggestion
fn is_impl_not_trait_with_bool_out(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> bool {
```

tests/ui/bool_assert_comparison.rs
```diff
@@ -11,14 +13,67 @@ macro_rules! b {
     };
 }
 
+// Implements the Not trait but with an output type
+// that's not bool. Should not suggest a rewrite
+#[derive(Debug)]
+enum A {
```
giraffate, line 19: nits
```suggestion
enum ImplNotTraitWithoutBool {
```

tests/ui/bool_assert_comparison.rs
```diff
@@ -11,14 +13,67 @@ macro_rules! b {
...
+            A::VariantY(0) => A::VariantY(1),
+            A::VariantY(_) => A::VariantY(0),
+        }
+    }
+}
+
+// This type implements the Not trait with an Output of
+// type bool. Using assert!(..) must be suggested
+#[derive(Debug)]
+struct B;
```
giraffate, line 48:
```suggestion
struct ImplNotTraitWithBool;
```

clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -28,45 +30,77 @@ declare_clippy_lint! {
...
+        .filter(|id| implements_trait(cx, ty, *id, &[]))
+        .and_then(|id| {
+            cx.tcx.associated_items(id).find_by_name_and_kind(
+                cx.tcx,
+                Ident::from_str("Output"),
+                ty::AssocKind::Type,
+                id,
+            )
+        })
+        .map_or(false, |item| {
```
giraffate, line 58: nits
```suggestion
        .map_or(false, |assoc_item| {
```

clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -28,45 +30,77 @@ declare_clippy_lint! {
...
-impl EarlyLintPass for BoolAssertComparison {
-    fn check_expr(&mut self, cx: &EarlyContext<'_>, e: &Expr) {
+fn impl_not_trait_with_bool_out(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> bool {
+    let ty = cx.typeck_results().expr_ty(e);
+
+    cx.tcx
+        .lang_items()
+        .not_trait()
+        .filter(|id| implements_trait(cx, ty, *id, &[]))
+        .and_then(|id| {
```
giraffate, line 50: nits
```suggestion
        .filter(|trait_id| implements_trait(cx, ty, *id, &[]))
        .and_then(|trait_id| {
```

giraffate (commented):
Sorry for the late review.

Overall looks good! I made some comments for variable names.

xordi:
Thanks for the review @giraffate! I've made the changes you suggested in the naming.

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

Issue #6072, opened 2020-09-21
FP mut_range_bound: range bound modified inside enclosing loop


matthiaskrgr:
In this code we first create the mut range bound, and mutate it in another loop which means the loop that actually uses the range is executed several times with different bounds (even if it is true that inside the usage of the range, changing the index has no effect on the inner loop.
```rust
fn main() {
    let mut i = 3;
    while i < 1000 { // this invalidates the lint warning
        for x in i..9 { // this will be unaffected by the i+=4 but we modify i AND LOOP AGAIN!
            i += 4; // lint warns about this
        }

        if i > 12 {
            return;
        }
    }
}
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=30343889f1b8be663adb15c2d8184c2a

The lint should probably only fire if range bound variable and the range bounds usage are in the same scope?

### Meta
`clippy 0.0.212 (1fd5b9d 2020-09-20)`

flip1995:
I guess the only valid fix would be to check all parent expressions of the for loop if the range bound is used afterwards. This is a hard fix IMO?

PR #7607, opened 2021-08-29
`mut_range_bound` check for immediate break after mutation

### LINKED ISSUE #7532: `mut_range_bound` on immediately broken loop
I tried this code:

```rust
fn bytes_needed_for_level(_level: u32) -> usize {
    unimplemented!()
}

pub fn load(data: &[u8], mut level_count: u32) {
    let mut bytes_needed = 0;
    for level in 0..level_count {
        bytes_needed += bytes_needed_for_level(level);
        if bytes_needed > data.len() {
            eprintln!(
                "not enough bytes for the specified number of levels;
                loading only the first {} levels",
                level_count
            );
            level_count = level;
            break;
        }
    }

    // use level_count while loading the data...
    println!("level_count is {}", level_count);
}
```

I expected this to happen: no diagnostic. `level_count` is mutated for when it is used *after* the loop, and the loop is immediately broken, so the lack of effect changing the bound has on the number of iterations of the loop is moot.

Instead, this happened: the following diagnostic is emitted:

```
warning: attempt to mutate range bound within loop; note that the range of the loop is unchanged
  --> src/lib.rs:15:13
   |
15 |             level_count = level;
   |             ^^^^^^^^^^^
   |
   = note: `#[warn(clippy::mut_range_bound)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#mut_range_bound

warning: `playground` (lib) generated 1 warning
```

### Meta

clippy 0.1.56 (2021-08-03 a6ece56) on [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a7ad2f6b9be791def91122246fd21f6a)

dswij:
closes #7532 

`mut_range_bound` ignores mutation on range bounds that is placed immediately before break. Still warns if the break is not always reachable.

changelog: [`mut_range_bound`] ignore range bound mutations before immediate break

### DIFF
--- a/clippy_lints/src/loops/mod.rs
+++ b/clippy_lints/src/loops/mod.rs
@@ -397,6 +397,21 @@ declare_clippy_lint! {
     /// ### Why is this bad?
     /// One might think that modifying the mutable variable changes the loop bounds
     ///
+    /// ### Known problems
+    /// False positive when mutation is followed by a `break`, but the `break` is not immediately
+    /// after the mutation:
+    ///
+    /// ```rust
+    /// let mut x = 5;
+    /// for _ in 0..x {
+    ///     x += 1; // x is a range bound that is mutated
+    ///     ..; // some other expression
+    ///     break; // leaves the loop, so mutation is not an issue
+    /// }
+    /// ```
+    ///
+    /// False positive on nested loops ([#6072](https://github.com/rust-lang/rust-clippy/issues/6072))
+    ///
     /// ### Example
     /// ```rust
     /// let mut foo = 42;
--- a/clippy_lints/src/loops/mut_range_bound.rs
+++ b/clippy_lints/src/loops/mut_range_bound.rs
@@ -1,24 +1,27 @@
 use super::MUT_RANGE_BOUND;
-use clippy_utils::diagnostics::span_lint;
-use clippy_utils::{higher, path_to_local};
+use clippy_utils::diagnostics::span_lint_and_note;
+use clippy_utils::{get_enclosing_block, higher, path_to_local};
 use if_chain::if_chain;
-use rustc_hir::{BindingAnnotation, Expr, HirId, Node, PatKind};
+use rustc_hir::intravisit::{self, NestedVisitorMap, Visitor};
+use rustc_hir::{BindingAnnotation, Expr, ExprKind, HirId, Node, PatKind};
 use rustc_infer::infer::TyCtxtInferExt;
 use rustc_lint::LateContext;
+use rustc_middle::hir::map::Map;
 use rustc_middle::{mir::FakeReadCause, ty};
 use rustc_span::source_map::Span;
 use rustc_typeck::expr_use_visitor::{Delegate, ExprUseVisitor, PlaceBase, PlaceWithHirId};
 
 pub(super) fn check(cx: &LateContext<'_>, arg: &Expr<'_>, body: &Expr<'_>) {
-    if let Some(higher::Range {
-        start: Some(start),
-        end: Some(end),
-        ..
-    }) = higher::Range::hir(arg)
-    {
-        let mut_ids = vec![check_for_mutability(cx, start), check_for_mutability(cx, end)];
-        if mut_ids[0].is_some() || mut_ids[1].is_some() {
-            let (span_low, span_high) = check_for_mutation(cx, body, &mut_ids);
+    if_chain! {
+        if let Some(higher::Range {
+            start: Some(start),
+            end: Some(end),
+            ..
+        }) = higher::Range::hir(arg);
+        let (mut_id_start, mut_id_end) = (check_for_mutability(cx, start), check_for_mutability(cx, end));
+        if mut_id_start.is_some() || mut_id_end.is_some();
+        then {
+            let (span_low, span_high) = check_for_mutation(cx, body, mut_id_start, mut_id_end);
             mut_warn_with_span(cx, span_low);
             mut_warn_with_span(cx, span_high);
         }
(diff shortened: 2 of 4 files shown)


tests/ui/mut_range_bound.rs
```diff
@@ -8,6 +8,9 @@ fn main() {
     immut_range_bound();
     mut_borrow_range_bound();
     immut_borrow_range_bound();
+    mut_range_bound_break();
+    mut_range_bound_nested_break();
+    multiple_mut_range_bound_break();
```
dswij, line 13: Not sure why this needs to be here

flip1995: Back in the day those were necessary to suppress `unused` warnings. Now those warnings are disabled by default.

Long story short: Yes you can remove them.

clippy_lints/src/loops/mut_range_bound.rs
```diff
@@ -116,3 +120,49 @@ impl MutatePairDelegate<'_, '_> {
...
+        NestedVisitorMap::None
+    }
+
+    fn visit_expr(&mut self, expr: &'tcx Expr<'tcx>) {
+        if self.break_after_expr {
+            return;
+        }
+        if matches!(&expr.kind, ExprKind::If(..)) || matches!(&expr.kind, ExprKind::Match(..)) {
+            return;
+        }
```
flip1995, line 158: I think this is there to make sure that control flow doesn't get in the way? There are way more things that could break this analysis, like other loops for example.

I think you want to achieve, that `break;` doesn't have to follow the assignment immediately. 

I wouldn't care about this that much, since you should be able to move the assignment immediately in front of the `break`. E.g.:

```rs
m = some_expr;
x = 2 + m;
break;
```

could also be written as

```rs
res = some_expr
x = 2 + res;
m = res;
break;
```

---

On the other hand, it would be nice, if this would also work on the above example, but this is difficult to implement correctly.

dswij: I'm not sure that I completely follow the example here

flip1995: I'm not sure if I understand the code in the visitor correctly. Why do you check for `if` and `match`?

dswij: This is to cover for

https://github.com/dswij/rust-clippy/commit/de0f3de605e6d4c51c812322692df545a6cad343#diff-26bc8b24b8058cb2b9f8fdc7ed87d6cb272e71386db63e328905d827ec242adaR89-R99

The `break` inside the control flow is not always reachable, so I think it should still be warned. The check is there to ignore `break`s inside control flow. 

That being said, I do agree that there are more things that might break this, such as #6072 😕

flip1995: Yes, I thought this would be the reason. 

So I think this would be easier to implement, to just allow `break;`s  right after the expression that mutates the range bound variable. This would mean that

```rs
m = 5;
some_other_expr;
break;
```

would still lint, but 

```rs
some_other_expr;
m = 5;
break;
```

would no longer lint.

dswij: Ah OK, I understand the first comment now. 

Yes, I think that this actually makes more sense in term of what we want to achieve rather than complicate things.

tests/ui/mut_range_bound.rs
```diff
@@ -61,3 +53,32 @@ fn immut_range_bound() {
...
+        m = 2; // warning because it is not immediately followed by break
+        if m == 4 {
+            break;
+        }
+    }
+
+    let mut n = 3;
+    for i in n..10 {
+        if n == 4 {
+            n = 1; // warning because is is not immediately followed by break
```
flip1995, line 79:
```suggestion
            n = 1; // FIXME: warning because is is not immediately followed by break
```

And can you mention this in the lint documentation under "### Known Problems" please?

clippy_lints/src/loops/mut_range_bound.rs
```diff
@@ -1,23 +1,26 @@
...
-    {
+    if_chain! {
+        if let Some(higher::Range {
+            start: Some(start),
+            end: Some(end),
+            ..
+        }) = higher::Range::hir(arg);
         let mut_ids = vec![check_for_mutability(cx, start), check_for_mutability(cx, end)];
-        if mut_ids[0].is_some() || mut_ids[1].is_some() {
+        if mut_ids[0].is_some() || mut_ids[1].is_some();
```
flip1995, line 22: While you're at it, ca you change this to

```rust
let (start_mut, end_mut) = (check_for_mutability(cx, start), check_for_mutability(cx, end));
if start_mut.is_some() || end_mut.is_some();
```

Allocating a vector here, seems weird to me. The `&mut_ids` below can be replaced with `&[start_mut, end_mut]`.

tests/ui/mut_range_bound.stderr
```diff
@@ -1,34 +1,46 @@
...
    |                      ^
 
-error: aborting due to 5 previous errors
+error: attempt to mutate range bound within loop; note that the range of the loop is unchanged
+  --> $DIR/mut_range_bound.rs:70:9
+   |
+LL |         m = 2; // warning because it is not immediately followed by break
+   |         ^
+
+error: attempt to mutate range bound within loop; note that the range of the loop is unchanged
```
flip1995, line 39: Fixing/improving old lints comes with also fixing some technical debt.

Can you swap the lint emitting with `span_lint(...)` with `span_lint_and_note(...)` and put the `note` part as a note with just the message "the range of the loop is unchanged"?

flip1995 (commented):
LGTM now. All comments are about technical debt with this lint, which you uncovered by working on this lint :stuck_out_tongue:

clippy_lints/src/loops/mod.rs
```diff
@@ -397,6 +397,10 @@ declare_clippy_lint! {
     /// ### Why is this bad?
     /// One might think that modifying the mutable variable changes the loop bounds
     ///
+    /// ### Known problems
+    /// False positive when mutation is not immediately followed by a break statement
+    /// False positive on nested expressions ([#6072](https://github.com/rust-lang/rust-clippy/issues/6072))
```
flip1995, line 402:
```suggestion
    /// False positive when mutation is followed by a `break`, but the `break` is not immediately after the mutation:
    ///
    /// ```
    /// x += 1; // x is range bound
    /// some_other_expr();
    /// break; // leaves the loop so mutating is not an issue
    /// ```
    ///
    /// False positive on nested loops ([#6072](https://github.com/rust-lang/rust-clippy/issues/6072))
```

flip1995 (commented):
Last comment. Thank for addressing all of my complaints :smile:

flip1995:
@bors r+

Thanks!

PR #7612, opened 2021-09-01
[`bool_assert_comparaison`] improve suggestion

### LINKED ISSUE #7598: `bool_assert_comparison` improve the suggestion to include the expression
The lint message for the  [`bool_assert_comparison`](https://rust-lang.github.io/rust-clippy/master/index.html#bool_assert_comparison) lint currently look like this:

```rs
 --> src/main.rs:5:5
  |
5 |     assert_eq!(some_bool_expression, false);
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `assert!(..)`
```

I think the suggestion can be improved to include the actual expression that should be put in the `assert!` instead of just having the placeholder `..`. 

The text for the suggestion can be retrieved with functions from [`clippy_utils::source`](https://github.com/rust-lang/rust-clippy/blob/master/clippy_utils/src/source.rs).

ABouttefeux:
close #7598

now code like
```rust
assert_eq!( a, true);
```
Will suggest to use `assert(a)` wherease previously it was `assert(..)`.

Also what I would like to add is with
```rust
assert_eq!( a, true, "message {}", b);
```
is to suggest `assert!( a, "message {}", b)`.
It is not the case at the moment.

changelog: [`bool_assert_comparaison`] improve suggestion

### DIFF
--- a/clippy_lints/src/bool_assert_comparison.rs
+++ b/clippy_lints/src/bool_assert_comparison.rs
@@ -1,5 +1,5 @@
 use clippy_utils::{
-    diagnostics::span_lint_and_sugg, higher::AssertExpn, is_direct_expn_of, source, sugg::Sugg, ty::implements_trait,
+    diagnostics::span_lint_and_sugg, higher::AssertExpn, is_direct_expn_of, source, ty::implements_trait,
 };
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
--- a/clippy_lints/src/bool_assert_comparison.rs
+++ b/clippy_lints/src/bool_assert_comparison.rs
@@ -1,4 +1,4 @@
-use clippy_utils::{diagnostics::span_lint_and_sugg, higher, is_direct_expn_of, ty::implements_trait};
+use clippy_utils::{diagnostics::span_lint_and_sugg, higher, is_direct_expn_of, source, ty::implements_trait};
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, Lit};
@@ -30,14 +30,19 @@ declare_clippy_lint! {
 
 declare_lint_pass!(BoolAssertComparison => [BOOL_ASSERT_COMPARISON]);
 
-fn is_bool_lit(e: &Expr<'_>) -> bool {
-    matches!(
-        e.kind,
+fn bool_lit(e: &Expr<'_>) -> Option<bool> {
+    match e.kind {
         ExprKind::Lit(Lit {
-            node: LitKind::Bool(_),
-            ..
-        })
-    ) && !e.span.from_expansion()
+            node: LitKind::Bool(b), ..
+        }) => {
+            if e.span.from_expansion() {
+                None
+            } else {
+                Some(b)
+            }
+        },
+        _ => None,
+    }
 }
 
 fn is_impl_not_trait_with_bool_out(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> bool {
@@ -68,19 +73,27 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
         let macros = ["assert_eq", "debug_assert_eq"];
         let inverted_macros = ["assert_ne", "debug_assert_ne"];
 
-        for mac in macros.iter().chain(inverted_macros.iter()) {
+        for (mac, is_eq) in macros
+            .iter()
+            .map(|el| (el, true))
+            .chain(inverted_macros.iter().map(|el| (el, false)))
+        {
             if let Some(span) = is_direct_expn_of(expr.span, mac) {
                 if let Some(args) = higher::extract_assert_macro_args(expr) {
                     if let [a, b, ..] = args[..] {
-                        let nb_bool_args = is_bool_lit(a) as usize + is_bool_lit(b) as usize;
+                        //let nb_bool_args = is_bool_lit(a) as usize + is_bool_lit(b) as usize;
 
-                        if nb_bool_args != 1 {
-                            // If there are two boolean arguments, we definitely don't understand
-                            // what's going on, so better leave things as is...
-                            //
-                            // Or there is simply no boolean and then we can leave things as is!
-                            return;
-                        }
+                        let (lit_value, other_expr) = match (bool_lit(a), bool_lit(b)) {
+                            (Some(lit), None) => (lit, b),
+                            (None, Some(lit)) => (lit, a),
+                            _ => {
+                                // If there are two boolean arguments, we definitely don't understand
+                                // what's going on, so better leave things as is...
+                                //
+                                // Or there is simply no boolean and then we can leave things as is!
+                                return;
+                            },
+                        };
 
                         if !is_impl_not_trait_with_bool_out(cx, a) || !is_impl_not_trait_with_bool_out(cx, b) {
                             // At this point the expression which is not a boolean
@@ -90,13 +103,28 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
                         }
 
                         let non_eq_mac = &mac[..mac.len() - 3];
+                        let expr_string = if lit_value ^ is_eq {
+                            format!("!({})", source::snippet(cx, other_expr.span, ""))
+                        } else {
+                            source::snippet(cx, other_expr.span, "").to_string()
+                        };
+
+                        let suggestion = if args.len() <= 2 {
+                            format!("{}!({})", non_eq_mac, expr_string)
+                        } else {
+                            let mut str = format!("{}!({}", non_eq_mac, expr_string);
+                            for el in &args[2..] {
+                                str = format!("{}, {}", str, source::snippet(cx, el.span, ""));
+                            }
+                            format!("{})", str)
+                        };
                         span_lint_and_sugg(
                             cx,
                             BOOL_ASSERT_COMPARISON,
                             span,
                             &format!("used `{}!` with a literal bool", mac),
                             "replace it with",
-                            format!("{}!(..)", non_eq_mac),
+                            suggestion,
                             Applicability::MaybeIncorrect,
                         );
                         return;
--- a/clippy_lints/src/bool_assert_comparison.rs
+++ b/clippy_lints/src/bool_assert_comparison.rs
@@ -81,8 +81,8 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
             .chain(inverted_macros.iter().map(|el| (el, false)))
         {
             if let Some(span) = is_direct_expn_of(expr.span, mac) {
-                if let Some(args) = AssertExpn::parse(expr).map(|v| v.argument_vector()) {
-                    if let [a, b, ref fmt_args @ ..] = args[..] {
+                if let Some(parse_assert) = AssertExpn::parse(expr) {
+                    if let [a, b] = parse_assert.assert_arguments()[..] {
                         let (lit_value, other_expr) = match (bool_lit(a), bool_lit(b)) {
                             (Some(lit), None) => (lit, b),
                             (None, Some(lit)) => (lit, a),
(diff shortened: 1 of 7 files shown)


clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -68,19 +73,27 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
...
+        for (mac, is_eq) in macros
+            .iter()
+            .map(|el| (el, true))
+            .chain(inverted_macros.iter().map(|el| (el, false)))
+        {
             if let Some(span) = is_direct_expn_of(expr.span, mac) {
                 if let Some(args) = higher::extract_assert_macro_args(expr) {
                     if let [a, b, ..] = args[..] {
-                        let nb_bool_args = is_bool_lit(a) as usize + is_bool_lit(b) as usize;
+                        //let nb_bool_args = is_bool_lit(a) as usize + is_bool_lit(b) as usize;
```
camsteffen, line 84: delete this commented code

clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -90,13 +103,28 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
                         }
 
                         let non_eq_mac = &mac[..mac.len() - 3];
+                        let expr_string = if lit_value ^ is_eq {
+                            format!("!({})", source::snippet(cx, other_expr.span, ""))
```
camsteffen, line 107: `snippet` should be directly imported. Use `Sugg::hir_with_applicability` and `Sugg::maybe_par`. The default value should be `".."` instead of `""`.

clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -68,19 +73,27 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
...
 
-        for mac in macros.iter().chain(inverted_macros.iter()) {
+        for (mac, is_eq) in macros
+            .iter()
+            .map(|el| (el, true))
+            .chain(inverted_macros.iter().map(|el| (el, false)))
+        {
             if let Some(span) = is_direct_expn_of(expr.span, mac) {
                 if let Some(args) = higher::extract_assert_macro_args(expr) {
                     if let [a, b, ..] = args[..] {
```
camsteffen, line 83: Use this instead of `args[2..]`.
```suggestion
                    if let [a, b, fmt_args @ ..] = args[..] {
```

clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -90,13 +103,28 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
...
+                        let expr_string = if lit_value ^ is_eq {
+                            format!("!({})", source::snippet(cx, other_expr.span, ""))
+                        } else {
+                            source::snippet(cx, other_expr.span, "").to_string()
+                        };
+
+                        let suggestion = if args.len() <= 2 {
+                            format!("{}!({})", non_eq_mac, expr_string)
+                        } else {
+                            let mut str = format!("{}!({}", non_eq_mac, expr_string);
```
camsteffen, line 115: Avoid variable names that are primitive types.

tests/ui/bool_assert_comparison.stderr
```diff
@@ -2,135 +2,135 @@ error: used `assert_eq!` with a literal bool
...
 LL |     debug_assert_eq!("a".is_empty(), false, "tadam {}", true);
-   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `debug_assert!(..)`
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `debug_assert!(!("a".is_empty()))`
 
 error: used `debug_assert_eq!` with a literal bool
   --> $DIR/bool_assert_comparison.rs:120:5
    |
 LL |     debug_assert_eq!(false, "a".is_empty(), "tadam {}", true);
-   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `debug_assert!(..)`
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace it with: `debug_assert!(!("a".is_empty()))`
```
camsteffen, line 133: Format args are missing here.

clippy_lints/src/bool_assert_comparison.rs
```diff
@@ -90,13 +103,28 @@ impl<'tcx> LateLintPass<'tcx> for BoolAssertComparison {
...
+                            for el in &args[2..] {
+                                str = format!("{}, {}", str, source::snippet(cx, el.span, ""));
+                            }
+                            format!("{})", str)
```
camsteffen, line 119: Use the `String` API (`push` and `push_str`).

Or better yet, get just one snippet of all the format args with something like

```rust
match fmt_args {
    [] => None,
    [a] => Some(a.span),
    [a, .., b] => Some(a.span.to(b.span)),
}

**Review by camsteffen — CHANGES_REQUESTED** [review]

**Comment by ABouttefeux** [comment]
The problem I see is that `higher::extract_assert_macro_args(expr)` does not give me the formatting arguments. I do not know how to get them.

**Comment by ABouttefeux** [comment]
I still not able to have the formats arg. I do not how to extract them.

**Comment by camsteffen** [comment]
> The problem I see is that `higher::extract_assert_macro_args(expr)` does not give me the formatting arguments. I do not know how to get them.

You can open a Rust Playground with an `assert_eq!` and run "Tools > Expand macros" to see what the expansion looks like. You will see a `match` expression. Currently the macro parsing in `extract_assert_macro_args` [basically stops there](https://github.com/rust-lang/rust-clippy/blob/a8c2c7b712c273f13258de8b87b233918d8726b9/clippy_utils/src/higher.rs#L431). You need to extend the parsing farther to the part where you see `Option::Some(...)`. The contents of the `Some` is `format_args!`, so that expression can be parsed with `FormatArgsExpn::parse` and just add `Option<FormatArgsExpn>` to the output of `extract_assert_macro_args`. Let me know if that makes sense.

**Comment by ABouttefeux** [comment]
Yes it does thank you.

**Review by balarajupetta — APPROVED** [review]

**Comment by camsteffen** [comment]
Hi @ABouttefeux, it looks like tests are not passing. Also please do a rebase instead of a merge since we follow rust's no-merge policy: https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy

**Comment by ABouttefeux** [comment]
I am not able yet to extract the fist formatting argument, I change it to ".." temporarily. I will work on it later. Do you have any idea how to extract it ?

**Comment by camsteffen** [comment]
That would come from `FormatArgsExpn::parse`. See [this](https://github.com/rust-lang/rust/blob/c9db3e0fbc84d8409285698486375f080d361ef3/library/core/src/macros/mod.rs#L57).

File: clippy_utils/src/higher.rs [hunk 7]
```diff
@@ -502,6 +521,8 @@ impl FormatExpn<'tcx> {
 
 /// A parsed `format_args!` expansion
 pub struct FormatArgsExpn<'tcx> {
+    /// The fist argument, the fromat string, as an expr
+    pub format_string: &'tcx Expr<'tcx>,
     /// Span of the first argument, the format string
     pub format_string_span: Span,
```
ABouttefeux, line 527: Should I just remove `format_string_span` as the span can be easily be found from the expr

clippy_utils/src/higher.rs
```diff
@@ -428,12 +428,31 @@ pub fn extract_assert_macro_args<'tcx>(e: &'tcx Expr<'tcx>) -> Option<Vec<&'tcx
...
+                    if !arms.is_empty();
+                    if let ExprKind::Block(Block{expr: Some(if_expr),..},_) = arms[0].body.kind;
+                    if let ExprKind::If(_, if_block, _) = if_expr.kind;
+                    if let ExprKind::Block(Block{stmts: stmts_if_block,..},_) = if_block.kind;
+                    if stmts_if_block.len() >= 2;
+                    if let StmtKind::Expr(call_assert_failed)
+                        | StmtKind::Semi(call_assert_failed) = stmts_if_block[1].kind;
+                    if let ExprKind::Call(_, args_assert_failed) = call_assert_failed.kind;
+                    if args_assert_failed.len() >= 4;
+                    if let ExprKind::Call(_, args) =  args_assert_failed[3].kind;
```
camsteffen, line 447: Prefer destructuring for stmts, args, etc.
```suggestion
                    if let ExprKind::Call(_, [arg]) =  args_assert_failed[3].kind;
```

clippy_utils/src/higher.rs
```diff
@@ -428,12 +428,31 @@ pub fn extract_assert_macro_args<'tcx>(e: &'tcx Expr<'tcx>) -> Option<Vec<&'tcx
...
+                    if !args.is_empty();
+                    if let Some (mut format_arg_expn) = FormatArgsExpn::parse(&args[0]);
+                    then {
+                        vec_arg.push(format_arg_expn.format_string);
+                        vec_arg.append(&mut format_arg_expn.value_args);
```
camsteffen, line 452: I think we should change `extract_assert_macro_args` to be more consistent with other utils here. We could have `AssertExpn::parse` and `AssertExpn` is a struct that contains a `FormatArgsExpn`.

camsteffen (changes requested):
Nice work getting the macro stuff working! Just some feedback with how the utils are structured and we should enable rustfix for tests now that we have valid suggestions.

clippy_utils/src/higher.rs
ABouttefeux: Thanks I have done something similar, but now I would like to test parsing assert(expr, "format {}", value), how can I test that ?

camsteffen: Use Expand macros in the rust playground to see what you are trying to parse. For other tips, use `dbg!()` to set a checkpoint in your code and see if that line is being reached. Use `dbg!(expr)` to see the HIR. Use `let _ = dbg!();` in `if_chain!`s.

clippy_utils/src/higher.rs
```diff
@@ -416,56 +419,169 @@ pub fn binop(op: hir::BinOpKind) -> ast::BinOpKind {
...
+/// macro.
+pub struct AssertExpn<'tcx> {
+    /// the first agrument of the assret e.g. `var` in element `assert!(var, ...)`
+    pub first_assert_argument: &'tcx Expr<'tcx>,
+    /// second argument of the asset for case `assert_eq!`,
+    /// `assert_ne!` etc ... Eg var_2 in `debug_assert_eq!(x, var_2,..)`
+    pub second_assert_argument: Option<&'tcx Expr<'tcx>>,
+    /// The format argument passed at the end of the macro
+    pub format_arg: Option<FormatArgsExpn<'tcx>>,
+}
```
camsteffen, line 434: I think an enum here would be good.

```rust
enum AssertExpnKind {
    Bool(Expr),
    Eq(Expr, Expr),
}
```

And maybe a flag `is_debug: bool`.

clippy_utils/src/higher.rs
```diff
@@ -416,56 +419,169 @@ pub fn binop(op: hir::BinOpKind) -> ast::BinOpKind {
...
+                                        format_arg: Some(format_arg_expn), // FIXME actually parse the aguments
+                                    });
+                                }
+                                else{
+                                    return Some(Self {
+                                        first_assert_argument: condition,
+                                        second_assert_argument: None,
+                                        format_arg: None,
+                                    });
+                                }
```
camsteffen, line 477: Try to avoid `else` in `if_chain!` here since it adds an else clause to every `if` in the expansion. You should only need to "fork" at a single point.

clippy_utils/src/higher.rs
```diff
@@ -416,56 +419,169 @@ pub fn binop(op: hir::BinOpKind) -> ast::BinOpKind {
...
+    /// Currently working with:
+    /// - `assert!`, `assert_eq!` and `assert_ne!`
+    /// - `debug_assert!`, `debug_assert_eq!` and `debug_assert_ne!`
+    /// For example:
+    /// `assert!(expr)` will return `Some(AssertExpn { first_assert_argument: expr,
+    /// second_assert_argument: None, format_arg:None })` `debug_assert_eq!(a, b)` will return
+    /// `Some(AssertExpn { first_assert_argument: a, second_assert_argument: Some(b),
+    /// format_arg:None })`
+    pub fn parse(e: &'tcx Expr<'tcx>) -> Option<Self> {
+        if let ExprKind::Block(block, _) = e.kind {
```
camsteffen, line 447: Consider doing something like this as a first step

```rust
if let ExpnKind::Macro(_, name) = expr.span.ctxt().outer_expn_data().kind;
match *name.as_str() {
    "assert" | "debug_assert" => ..,
    "assert_eq" | "debug_assert_eq" => ..,
}
```

giraffate:
ping from triage @ABouttefeux. It looks like there are a few things left to fix. Can you address these?

camsteffen:
Hey @ABouttefeux, let me know when you are ready for review.

Also I want to point out #7843 - we may need to take a different direction for parsing the assert macros that does not parse the exact structure of the expansion. Although this is somewhat pre-existing so maybe it's okay to continue with the current approach for this PR. (@rust-lang/clippy thoughts?) I think it is possible to take a different approach using a visitor and checking span expansion data, but this is quite tricky.

camsteffen:
@ABouttefeux After #8219 is merged, have another look if you're still interested. There are new utils in place for assert macros.

flip1995:
ping from triage @ABouttefeux. #8219 was merged, and you should be able to use those new utils to implement this lint and get this PR merged.

camsteffen:
Closing due to inactivity. Feel free to reopen if you'd like to pick this up again.

PR #7625, opened 2021-09-02
Set cargo's output directory to `<repo root>/target/`

kneasle:
Makes the tests pass immediately for people (like me) who use a unified `target/` directory by default.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -6,4 +6,5 @@ collect-metadata = "test --test dogfood --features metadata-collector-lint -- ru
 
 [build]
 # -Zbinary-dep-depinfo allows us to track which rlib files to use for compiling UI tests
-rustflags = ["-Zunstable-options", "-Zbinary-dep-depinfo"]
+rustflags = ["-Zunstable-options"]
+target-dir = "target/" # Tests require that the build files are placed in <repo root>/target


camsteffen:
Thanks for the PR @kneasle. Can you provide more detail as to what problem this solves?

kneasle:
> Thanks for the PR @kneasle. Can you provide more detail as to what problem this solves?

The commit message had some more info - when building Clippy, the UI tests runner expects the binary to be in a hard-coded location in the `<clippy repo>/target` directory.  However, Cargo lets you customise this output directory, which is (not uncommonly) used to put all the build artifacts in one folder like `~/.build/rust/`.  In this scenario, Clippy's UI tests would still look for the built binary in the `<clippy repo>/target` directory, making the test runner fail without a terribly obvious error message.  It's not a disaster - I would expect anyone who's customised their target directory to know how to undo it - but it's a pain to have to manually enforce this, and it'd be nice if it just worked out of the box.

So this PR overrides Cargo's output directory option in Clippy's own `.cargo/config` file so that Clippy always puts its build files into `<clippy repo>/target` (and the change will only apply inside clippy's repo).

camsteffen:
I wonder if we could actually fix the problem at the source by making the tests work with a different target. But if not, I guess this is fine. We should just make sure it doesn't break the rust repo tests/CI.

kneasle:
If [this comment](https://github.com/rust-lang/rust-clippy/blob/master/tests/compile-test.rs#L33-L35) is anything to go by, it seems you're not the first to think this (and if git blame is anything to go by, that person was @tesuji).  But it seems like this requires way more knowledge of compiler internals than I have :sweat_smile:.  I figured this PR would at least get things working for new contributors out of the box (getting everything working was otherwise a very smooth experience).

flip1995:
I'll have to test if this breaks something in the Rust repo. But since the sync situation between those two repos is currently _difficult_ this will take me about a week.

kneasle:
> I'll have to test if this breaks something in the Rust repo. But since the sync situation between those two repos is currently _difficult_ this will take me about a week.

That's fine, no stress :smiley:.

camsteffen:
You should be able to use your custom target directory with #7646.

kneasle:
> You should be able to use a custom target directory with #7646.

Oh excellent, indeed it does - cheers!

camsteffen:
Actually I just found out that I had to add in the cargo config like you have here. But it should be safe to override it.

kneasle:
> Actually I just found out that I had to add in the cargo config like you have here. But it should be safe to override it.

Same here, so I've re-opened this PR.  For #7640, I had to set it locally and avoid checking the change into git.

camsteffen:
I mean, the change in this PR is also needed in #7646, so it's already included there.

kneasle:
> I mean, the change in this PR is also needed in #7646, so it's already included there.

Oh I seeeeee! Too many similar PR numbers :sweat_smile:.

Issue #7626, opened 2021-09-03
False positive for `unused-unsafe`


cmichi:
Lint name: `unused-unsafe`

I tried this code:

`src/lib.rs`:
```rust 
pub extern fn test() {
  unsafe { core::arch::wasm32::unreachable(); }
}
```

`Cargo.toml`:
```toml
[package]
name = "test"
version = "0.1.0"
edition = "2018"

[lib]
name = "test"
path = "src/lib.rs"
crate-type =["cdylib"]

[dependencies]
```

I expected to see this happen: No lint error should be reported.

Instead, this happened: 
```bash
$ cargo clippy --verbose --no-default-features --target wasm32-unknown-unknown -- -D warnings
…
error: unnecessary `unsafe` block
 --> src/lib.rs:2:2                                                                                                                                            
  |                                                                                                                                                            
2 |     unsafe { core::arch::wasm32::unreachable(); }                                                                                                          
  |     ^^^^^^ unnecessary `unsafe` block                                                                                                                      
  |                                                                            
  = note: `-D unused-unsafe` implied by `-D warnings`
…
```
The function is marked as `unsafe` though: https://doc.rust-lang.org/stable/core/arch/wasm32/fn.unreachable.html.

There is also no lint error reported if I switch to `rustup override set nightly-2021-08-13`.

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.56.0-nightly (371f3cd3f 2021-09-02)
binary: rustc
commit-hash: 371f3cd3fe523d0b398ed1db1620667c53ba7d02
commit-date: 2021-09-02
host: x86_64-unknown-linux-gnu
release: 1.56.0-nightly
LLVM version: 13.0.0
```

cmichi:
Turns out the function was indeed recently marked as safe: https://doc.rust-lang.org/nightly/core/arch/wasm32/fn.unreachable.html.

>This function is safe to call and immediately aborts the execution.

Issue #7599, opened 2021-08-25
Test issue (Testing the label command from rustbot)


xFrednet:
This should add the label:

@rustbot label +C-an-interesting-project

And this should not:

xFrednet:
Hmm

xFrednet:
* @rustbot label +C-an-interesting-project

xFrednet:
Okay, I got my answers. If someone is reading this: Here, have a cookie :cookie:

xFrednet:
Wait, not quite done yet

xFrednet:
Hmmmm

xFrednet:
What about this:

xFrednet:
And this:

@rustbot label -C-Question

xFrednet:
That's an even better solution... I should really stop with the spam now xD

xFrednet:
\@rustbot label ss

xFrednet:
\@rustbot label +C-Bug

xFrednet:
(at)rustbot label +C-Bug

xFrednet:
One more test (@)rustbot label +Hello

xFrednet:
@ rustbot label +Hello

xFrednet:
What about this:

xFrednet:
Now again without the backticks


@rustbot label +C-Bug

PR #7627, opened 2021-09-03
Updating issue templates again for rustbot

xFrednet:
It turns out that our current issue template can sometimes trigger a rustbot error message, as can be seen in [#7626](https://github.com/rust-lang/rust-clippy/issues/7626). I originally tested this in #7599, but it's apparently a bit inconsistent. This PR adds backticks to the commands, as correctly suggested by @mikerite in the comments. (Thank you!)

`@rustbot label +S-blocked`

---

Now I also pushed a tiny link fix as well. :upside_down_face: 

---

changelog: none

### DIFF
--- a/.github/ISSUE_TEMPLATE/blank_issue.md
+++ b/.github/ISSUE_TEMPLATE/blank_issue.md
@@ -5,9 +5,10 @@ about: Create a blank issue.
 
 
 <!--
-Additional labels can be added to this issue by including the following command:
+Additional labels can be added to this issue by including the following command
+(without the space after the @ symbol):
 
-@rustbot label +<label>
+`@rustbot label +<label>`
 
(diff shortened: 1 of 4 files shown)


mikerite:
Why not surround the whole comment in backticks? I'm sure people will know to remove them.

Or does that have the same issue?

xFrednet:
The issue is that GitHub ignores backticks inside comments. At least rustbot still triggers with them. I'll ping you inside an HTML comment with backticks as an example

xFrednet:
Well now it's somehow not triggering rustbot with the backticks. Maybe I missed that in my tests earlier

xFrednet:
You're right, nice catch!

llogiq:
From my side, this looks Ok. r=me if you think it's ready to merge.

PR #7630, opened 2021-09-04
Fix `multiple rlibs` error in `compile-test.rs`

Jarcho:
fixes: #7343

This solves the problem by looking up the dependencies of the currently running test.  There is a requirement that all the crates are mentioned in `Cargo.toml`, but I don't see that as a big downside (It also happens to guarantee they're built).

The only downside is this will break if/when cargo changes how fingerprint hashes are calculated. Really this should depend on cargo for that, but it's a large dependency and I don't think cargo is in the sysroot.

changelog: none


camsteffen:
Looks like we worked on this at the same time. I found a different solution that is less dependent on cargo internals. I may open a PR soon.

PR #7640, opened 2021-09-05
Improve accuracy of `mut_key`

### LINKED ISSUE #6745: mutable_key_type false positive for raw pointers
Lint name: mutable_key_type

I tried this code:

```rust
use std::collections::HashSet;

fn main() {
   let s: HashSet<*mut ()> = HashSet::new(); 
}
```

I expected to see this happen: lint not triggered because raw pointers are compared and hashed by pointer value rather than by pointed-to value

Instead, this happened: lint triggered

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=601da4e6cf6eb0b2c178c32fcab35390

### Meta

- `cargo clippy -V`: clippy 0.1.52 (d1206f9 2021-02-15)
- `rustc -Vv`:
  ```
  rustc 1.52.0-nightly (d1206f950 2021-02-15)
  binary: rustc
  commit-hash: d1206f950ffb76c76e1b74a19ae33c2b7d949454
  commit-date: 2021-02-15
  release: 1.52.0-nightly
  LLVM version: 11.0.1
  ```

kneasle:
Fixes #6745.

Whilst writing the tests for this, I noticed what I believe is a false negative (the code in @xFrednet's [comment](https://github.com/rust-lang/rust-clippy/issues/6745#issuecomment-909658267) doesn't trigger the lint).  Currently the tests contain a case for this (which is blatantly ignored), but I'm not at all sure how to implement this (since the lint currently behaves completely differently for ADTs).  I'm not sure what should be done - on the one hand the extra test cases are misleading, but on the other hand they don't cause much harm and would save effort for anyone fixing that false negative.

---

changelog: Improve accuracy of `clippy::mutable_key_type`.

### DIFF
--- a/clippy_lints/src/mut_key.rs
+++ b/clippy_lints/src/mut_key.rs
@@ -103,26 +103,39 @@ fn check_sig<'tcx>(cx: &LateContext<'tcx>, item_hir_id: hir::HirId, decl: &hir::
 fn check_ty<'tcx>(cx: &LateContext<'tcx>, span: Span, ty: Ty<'tcx>) {
     let ty = ty.peel_refs();
     if let Adt(def, substs) = ty.kind() {
-        if [sym::hashmap_type, sym::BTreeMap, sym::hashset_type, sym::BTreeMap]
+        let is_map_type = [sym::hashmap_type, sym::BTreeMap, sym::hashset_type, sym::BTreeMap]
             .iter()
-            .any(|diag_item| cx.tcx.is_diagnostic_item(*diag_item, def.did))
-            && is_mutable_type(cx, substs.type_at(0), span)
-        {
+            .any(|diag_item| cx.tcx.is_diagnostic_item(*diag_item, def.did));
+        if is_map_type && is_mutable_type(cx, substs.type_at(0), span, true) {
             span_lint(cx, MUTABLE_KEY_TYPE, span, "mutable key type");
         }
     }
 }
 
-fn is_mutable_type<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>, span: Span) -> bool {
+fn is_mutable_type<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>, span: Span, is_top_level_type: bool) -> bool {
     match *ty.kind() {
-        RawPtr(TypeAndMut { ty: inner_ty, mutbl }) | Ref(_, inner_ty, mutbl) => {
-            mutbl == hir::Mutability::Mut || is_mutable_type(cx, inner_ty, span)
+        RawPtr(TypeAndMut { ty: inner_ty, mutbl }) => {
+            if is_top_level_type {
+                // Raw pointers are hashed by the address they point to, not what is pointed to.
+                // Therefore, using a raw pointer to any type as the top-level key type is OK.
+                // Using raw pointers _in_ the key type is not, because the wrapper type could
+                // provide a custom `impl` for `Hash` (which could deref the raw pointer).
+                //
+                // see:
+                // - clippy issue: https://github.com/rust-lang/rust-clippy/issues/6745
+                // - std code: https://github.com/rust-lang/rust/blob/1.54.0/library/core/src/hash/mod.rs#L717-L736
+                false
+            } else {
+                mutbl == hir::Mutability::Mut || is_mutable_type(cx, inner_ty, span, false)
+            }
         },
-        Slice(inner_ty) => is_mutable_type(cx, inner_ty, span),
+        Ref(_, inner_ty, mutbl) => mutbl == hir::Mutability::Mut || is_mutable_type(cx, inner_ty, span, false),
+        Slice(inner_ty) => is_mutable_type(cx, inner_ty, span, false),
         Array(inner_ty, size) => {
-            size.try_eval_usize(cx.tcx, cx.param_env).map_or(true, |u| u != 0) && is_mutable_type(cx, inner_ty, span)
+            size.try_eval_usize(cx.tcx, cx.param_env).map_or(true, |u| u != 0)
+                && is_mutable_type(cx, inner_ty, span, false)
         },
-        Tuple(..) => ty.tuple_fields().any(|ty| is_mutable_type(cx, ty, span)),
+        Tuple(..) => ty.tuple_fields().any(|ty| is_mutable_type(cx, ty, span, false)),
         Adt(..) => {
             !ty.has_escaping_bound_vars()
                 && cx.tcx.layout_of(cx.param_env.and(ty)).is_ok()
--- a/clippy_lints/src/mut_key.rs
+++ b/clippy_lints/src/mut_key.rs
@@ -3,7 +3,7 @@ use clippy_utils::trait_ref_of_method;
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::TypeFoldable;
-use rustc_middle::ty::{Adt, Array, RawPtr, Ref, Slice, Tuple, Ty, TypeAndMut};
+use rustc_middle::ty::{Adt, Array, Ref, Slice, Tuple, Ty};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::source_map::Span;
 use rustc_span::symbol::sym;
@@ -19,10 +19,29 @@ declare_clippy_lint! {
     /// so having types with interior mutability is a bad idea.
     ///
     /// ### Known problems
-    /// It's correct to use a struct, that contains interior mutability
-    /// as a key, when its `Hash` implementation doesn't access any of the interior mutable types.
-    /// However, this lint is unable to recognize this, so it causes a false positive in theses cases.
-    /// The `bytes` crate is a great example of this.
+    ///
+    /// #### False Positives
+    /// It's correct to use a struct that contains interior mutability as a key, when its
+    /// implementation of `Hash` or `Ord` doesn't access any of the interior mutable types.
+    /// However, this lint is unable to recognize this, so it will often cause false positives in
+    /// theses cases.  The `bytes` crate is a great example of this.
+    ///
+    /// #### False Negatives
+    /// For custom `struct`s/`enum`s, this lint is unable to check for interior mutability behind
+    /// indirection.  For example, `struct BadKey<'a>(&'a Cell<usize>)` will be seen as immutable
+    /// and cause a false negative if its implementation of `Hash`/`Ord` accesses the `Cell`.
+    ///
+    /// This lint does check a few cases for indirection.  Firstly, using some standard library
+    /// types (`Option`, `Result`, `Box`, `Rc`, `Arc`, `Vec`, `VecDeque`, `BTreeMap` and
+    /// `BTreeSet`) directly as keys (e.g. in `HashMap<Box<Cell<usize>>, ()>`) **will** trigger the
+    /// lint, because the impls of `Hash`/`Ord` for these types directly call `Hash`/`Ord` on their
+    /// contained type.
+    ///
+    /// Secondly, the implementations of `Hash` and `Ord` for raw pointers (`*const T` or `*mut T`)
+    /// apply only to the **address** of the contained value.  Therefore, interior mutability
+    /// behind raw pointers (e.g. in `HashSet<*mut Cell<usize>>`) can't impact the value of `Hash`
+    /// or `Ord`, and therefore will not trigger this link.  For more info, see issue
+    /// [#6745](https://github.com/rust-lang/rust-clippy/issues/6745).
     ///
     /// ### Example
     /// ```rust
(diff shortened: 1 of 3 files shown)


clippy_lints/src/mut_key.rs
```diff
@@ -103,26 +103,39 @@ fn check_sig<'tcx>(cx: &LateContext<'tcx>, item_hir_id: hir::HirId, decl: &hir::
...
+fn is_mutable_type<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>, span: Span, is_top_level_type: bool) -> bool {
     match *ty.kind() {
-        RawPtr(TypeAndMut { ty: inner_ty, mutbl }) | Ref(_, inner_ty, mutbl) => {
-            mutbl == hir::Mutability::Mut || is_mutable_type(cx, inner_ty, span)
+        RawPtr(TypeAndMut { ty: inner_ty, mutbl }) => {
+            if is_top_level_type {
+                // Raw pointers are hashed by the address they point to, not what is pointed to.
+                // Therefore, using a raw pointer to any type as the top-level key type is OK.
+                // Using raw pointers _in_ the key type is not, because the wrapper type could
+                // provide a custom `impl` for `Hash` (which could deref the raw pointer).
```
camsteffen, line 122: I'm not sure the "top-level" disambiguation is necessary.

> the wrapper type could provide a custom `impl` for `Hash` (which could deref the raw pointer).

Do you have an example of this? It seems unlikely/wrong to me to have a specialized `impl Hash` that derefs the pointer. I'm pretty sure this would not happen for any primitive (like array) or std types. Also I think it's okay to lean on the side of having false negatives rather than false positives.

tests/ui/mut_key.rs
```diff
@@ -24,18 +25,50 @@ impl Hash for Key {
     }
 }
 
+struct PtrToCell(*const Cell<usize>);
+#[derive(Eq, PartialEq)]
+struct MutPtr(*mut usize);
```
camsteffen, line 30: Note that having a type that contains a field that is a pointer is different from having a type with a generic type parameter that may be a pointer type. `substs.type_at(0)` is the latter. So I'm not sure this test is meaningful.

camsteffen:
> Whilst writing the tests for this, I noticed what I believe is a false negative (the code in @xFrednet's [comment](https://github.com/rust-lang/rust-clippy/issues/6745#issuecomment-909658267) doesn't trigger the lint). Currently the tests contain a case for this (which is blatantly ignored), but I'm not at all sure how to implement this (since the lint currently behaves completely differently for ADTs). I'm not sure what should be done - on the one hand the extra test cases are misleading, but on the other hand they don't cause much harm and would save effort for anyone fixing that false negative.

My personal opinion is that this case is out of scope for the lint. I think if a user defines a type with a pointer and a custom `impl Hash`, they are on their own.

clippy_lints/src/mut_key.rs
kneasle: > Do you have an example of this?

Surely `Vec` is an example - it contains a raw pointer to some data (contained inside a `Unique<T>`), and its `impl Hash` depends on that data.  So, currently, `HashMap::<Cell<usize>>, ()>` triggers the lint but `HashMap::<Vec<Cell<usize>>, ()>` doesn't (despite being equally dangerous).  This makes sense, because the lint's case for `Adt`s doesn't truly check for mutability - it instead checks for the [`Freeze` trait](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/core/marker/trait.Freeze.html) which only checks for `UnsafeCell`s **without** indirection.  This also makes sense because (as I understand it) the `Freeze` trait simply says _"the bytes of this type can't change without a mutable reference"_.

So, in fact, I think the ADT clause should recurse over the (fully qualified) types of its fields (i.e. an ADT is mutable iff any of its subtypes are).  Not sure if that's worthy of another PR (or if it's even possible :sweat_smile:).

tests/ui/mut_key.rs
kneasle: > Note that having a type that contains a field that is a pointer is different from having a type with a generic type parameter that may be a pointer type.

How come?  Surely they're going to generate exactly the same code, so should be treated equivalently by the compiler and this lint?

clippy_lints/src/mut_key.rs
camsteffen: `HashMap<&[*mut Cell<usize>], ()>` or `HashMap<Vec<*mut Cell<usize>>, ()>` should not lint because `impl Hash for Vec<T>` or `&[T]` does not deref the pointer in the generic type.

For the ADT case, what we could do is check if the type is a known std collection (any type where generic parameters are hashed), and if it is, recurse over generic types. This would fix the `Vec<Cell<..>>` case.

tests/ui/mut_key.rs
camsteffen: They are different language constructs and they are represented differently when Clippy inspects them. The lint does not look at field types. And I don't think it should either. It would be expensive to recurse all fields, and we don't necessarily know which fields are hashed.

kneasle: > They are different language constructs

Ah OK that makes sense.  I still feel like it would be more logical to make them the same, but I'm sure there's a reason that the compiler folks have chosen this way of doing things.  In that case, it would make sense to remove these tests.

clippy_lints/src/mut_key.rs
kneasle: Ah yes I see this is more complicated than I first thought :sweat_smile:

> For the ADT case, what we could do is check if the type is a known std collection (any type where generic parameters are hashed), and if it is, recurse over generic types. This would fix the `Vec<Cell<..>>` case.

I like this solution - if someone has an ADT with some funky pointer magic and implements `Hash` by (unsafely) dereferencing that pointer, then I think it's justified for Clippy to not lint.  But equally it's worth making the lint as good as possible within performance/dev time/complexity constraints, so I think this is a good middle line.

I'll probs get on to this tomorrow, since I'm out of time for this evening...

kneasle: Right, this has been done!

tests/ui/mut_key.rs
kneasle: Righto these test cases have been removed.

clippy_lints/src/mut_key.rs
```diff
@@ -103,30 +103,63 @@ fn check_sig<'tcx>(cx: &LateContext<'tcx>, item_hir_id: hir::HirId, decl: &hir::
...
-            !ty.has_escaping_bound_vars()
-                && cx.tcx.layout_of(cx.param_env.and(ty)).is_ok()
-                && !ty.is_freeze(cx.tcx.at(span), cx.param_env)
+        Adt(def, substs) => {
+            // `LinkedList`, `Vec`, `VecDeque`, `BTreeSet`, `BTreeMap` all implement `Hash` by
+            // hashing their contents (`BTreeMap` hashes both its keys and values).  All of these
+            // store their contents behind (at least) one level of indirection, meaning that
+            // checking the `Freeze` trait for the type alone will cause false negatives for
+            // obviously bad cases like `HashSet<Vec<Cell<usize>>>`.  So, we handle these
+            // collections explicitly.
```
camsteffen, line 145: This list should also include Option, Result, HashMap, HashSet. `Option` and `Result` don't have any "indirection" so I'm not sure about that part. I would just say "generic containers where the contained value(s) are hashed and can be modified through interior mutability".

clippy_lints/src/mut_key.rs
```diff
@@ -103,30 +103,63 @@ fn check_sig<'tcx>(cx: &LateContext<'tcx>, item_hir_id: hir::HirId, decl: &hir::
...
+        // Raw pointers are never considered 'mutable' by this lint, since there are two cases:
+        // 1. The raw pointer is hashed/ordered directly (e.g. in `HashSet<*const Cell<usize>>` or
+        //    `HashSet<*mut usize>`.  These impls depend only on the memory **address** of the
+        //    referred value, rather than the value itself.  Therefore, checking the contained type
+        //    causes a false positive.  Links (links refer to Rust version 1.54):
+        //      - clippy issue: https://github.com/rust-lang/rust-clippy/issues/6745
+        //      - hash impl: https://github.com/rust-lang/rust/blob/1.54.0/library/core/src/hash/mod.rs#L717-L736
+        //      - ord impls:
+        //        - https://github.com/rust-lang/rust/blob/master/library/core/src/ptr/const_ptr.rs#L1030-L1043
+        //        - https://github.com/rust-lang/rust/blob/master/library/core/src/ptr/mut_ptr.rs#L1345-L1357
```
camsteffen, line 126: We shouldn't link to std implementation. The line numbers will go stale anyways.

clippy_lints/src/mut_key.rs
```diff
@@ -103,30 +103,63 @@ fn check_sig<'tcx>(cx: &LateContext<'tcx>, item_hir_id: hir::HirId, decl: &hir::
...
+        //
+        // 2. The raw pointer is contained inside an ADT.  Using `#[derive(Hash)]` will call
+        //    `<*const T as Hash>::hash()`, which will have the same behaviour as case (1).  If the
+        //    ADT uses a custom `impl Hash` which (unsafely) dereferences the raw pointer, then
+        //    they are on their own - checking this in the general case would be impossible.
```
camsteffen, line 131: This is not really relevant for this function since we are looking at the *type* and so we can't see ADT fields anyways. You could put this in a `Known problems` section of the lint docs.

clippy_lints/src/mut_key.rs
```diff
@@ -103,30 +103,63 @@ fn check_sig<'tcx>(cx: &LateContext<'tcx>, item_hir_id: hir::HirId, decl: &hir::
...
             span_lint(cx, MUTABLE_KEY_TYPE, span, "mutable key type");
         }
     }
 }
 
 fn is_mutable_type<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>, span: Span) -> bool {
     match *ty.kind() {
-        RawPtr(TypeAndMut { ty: inner_ty, mutbl }) | Ref(_, inner_ty, mutbl) => {
-            mutbl == hir::Mutability::Mut || is_mutable_type(cx, inner_ty, span)
-        },
```
camsteffen, line 120: Can we name and document the function `is_mutable_type` more precisely? Then everything else here is more or less self-explanatory. Comments in the function should ideally be written with respect to what this function does, not the lint.

```rust
/// Returns true for types where the hash of an instance of the type can be modified using
/// interior mutability (without a mutable reference).
/// For example, this includes `Cell` and `Vec<Cell<_>>`.
fn type_has_interior_mutable_hash(...) {
```

tests/ui/mut_key.rs
```diff
@@ -52,4 +61,16 @@ fn main() {
...
     tuples::<()>(&mut HashMap::new());
     tuples_bad::<()>(&mut HashMap::new());
+
+    raw_ptr_is_ok(&mut HashMap::new());
+    raw_mut_ptr_is_ok(&mut HashMap::new());
+
+    // Currently fail as expected:
+    let _map = HashMap::<Cell<usize>, usize>::new();
+    let _map = HashMap::<&mut Cell<usize>, usize>::new();
+    // Collection types from `std` which hash/order their type parameter
```
camsteffen, line 71: not sure what you mean by "order"

tests/ui/mut_key.rs
```diff
@@ -52,4 +61,16 @@ fn main() {
     tuples::<Key>(&mut HashMap::new());
     tuples::<()>(&mut HashMap::new());
     tuples_bad::<()>(&mut HashMap::new());
+
+    raw_ptr_is_ok(&mut HashMap::new());
+    raw_mut_ptr_is_ok(&mut HashMap::new());
+
+    // Currently fail as expected:
```
camsteffen, line 68: This comment won't make sense in the future.

camsteffen (changes requested):

tests/ui/mut_key.rs
kneasle: This lint triggers on both `Hash(Map|Set)` (which requires `T: Hash`) as well as `BTree(Map|Set)` (which requires `T: Ord`).  Interior mutability is bad in both cases, but 'order' was referring to `Ord::cmp` required by `BTree(Map|Set)`

clippy_lints/src/mut_key.rs
kneasle: Righto - links have gone :smiley:.

clippy_lints/src/mut_key.rs
kneasle: We don't need `Hash(Map|Set)`, because they don't implement either of `Hash` or `Ord` and can't be used as keys in the types which trigger this lint.  We do need a recursive case for `Option` and `Result` - cases like `Option<Cell<usize>>` would lint anyway but more complex cases like `Option<Vec<Cell<usize>>>` or `Option<&mut usize>` would not lint unless we recurse.  I've also added tests for these cases.

I've added simpler comment, but one which mentions that we care about both `Hash` (for `Hash(Map|Set)`) _and_ `Ord` (for `BTree(Map|Set)`).

tests/ui/mut_key.rs
kneasle: Right I've updated the comment.

tests/ui/mut_key.rs
kneasle: It's gone :smiley:.

clippy_lints/src/mut_key.rs
kneasle: Done!

clippy_lints/src/mut_key.rs
kneasle: Update: I've also added smart pointers like `Box`, `Arc`, `Rc`, `cell::{Ref, RefMut}`, etc.

clippy_lints/src/mut_key.rs
kneasle: I think `is_interior_mutable_type` would be good - the lint cares about both `Hash` _and_ `Ord`, and I don't feel like bloating the function name with that much detail.  I've added a doc comment which hopefully explains things.  I'll leave this conversation unresolved in case you have more feedback.

clippy_lints/src/mut_key.rs
```diff
@@ -19,10 +19,14 @@ declare_clippy_lint! {
...
-    /// However, this lint is unable to recognize this, so it causes a false positive in theses cases.
-    /// The `bytes` crate is a great example of this.
+    /// It's correct to use a struct that contains interior mutability as a key, when its
+    /// implementation of `Hash` or `Ord` doesn't access any of the interior mutable types.
+    /// However, this lint is unable to recognize this, so it will often cause false positives in
+    /// theses cases.  The `bytes` crate is a great example of this.
+    ///
+    /// Additionally, if a struct/enum contains interior mutability behind indirection then this
+    /// lint is unlikely to catch it.  For example, `struct BadKey<'a>(&'a Cell<usize>)` will not
+    /// be detected by this lint.
```
camsteffen, line 29:
```suggestion
    /// Additionally, if a struct/enum contains a field with interior mutability, then this
    /// lint will not catch it.  For example, `struct BadKey<'a>(&'a Cell<usize>)` will not
    /// be detected by this lint.
```

tests/ui/mut_key.rs
```diff
@@ -52,4 +63,28 @@ fn main() {
...
+    let _map = HashMap::<Result<&mut usize, ()>, usize>::new();
+    // Smart pointers from `std` who's impl of `Hash` or `Ord` delegate their type parameters
+    let _map = HashMap::<Box<Cell<usize>>, usize>::new();
+    let _map = HashMap::<Rc<Cell<usize>>, usize>::new();
+    let _map = HashMap::<Arc<Cell<usize>>, usize>::new();
+    let _map = HashMap::<Ref<Cell<usize>>, usize>::new();
+    let _map = HashMap::<RefMut<Cell<usize>>, usize>::new();
+
+    // should not lint, because we'd need a `&mut RefMut` to get mutable access to its contents
+    let _map = HashMap::<RefMut<usize>, usize>::new();
```
camsteffen, line 89: Hmm RefMut doesn't implement Hash/Ord.

clippy_lints/src/mut_key.rs
```diff
@@ -19,10 +19,14 @@ declare_clippy_lint! {
...
     ///
     /// ### Example
     /// ```rust
```
camsteffen, line 32: Huh isn't this example an instance of the false negative described above? Could you update that while you're at it? A simple `Cell` example would be good. Also feel free to update lint docs and description to be more precise (use "interior mutability").

camsteffen (changes requested):
A few nits and this is nearly ready. Please update the changelog in your post to reflect added changes.

clippy_lints/src/mut_key.rs
camsteffen: That's fine since it's isolated to this lint.

clippy_lints/src/mut_key.rs
kneasle: > Additionally, if a struct/enum contains _a field with interior mutability_, then this lint will not catch it

This isn't strictly true: what matters is the indirection because that's what impacts the `Freeze` trait, which is what the `Adt` case checks.  For example:

This will lint, because the bytes of `BadKey` can change without a mutable reference (i.e. `BadKey` is not `Freeze`):
```rust
struct BadKey {
    uh_oh: Cell<usize>,
}
```
but this won't (despite being just as bad) because the mutability happens to memory _owned_ by `SilentlyBadKey`, not `SilentlyBadKey` itself:
```rust
struct SilentlyBadKey {
    uh_oh: Box<Cell<usize>>,
}
```

The other change is correct, though - I'll fix that now.

tests/ui/mut_key.rs
kneasle: Huh that is indeed correct.  In fact, neither does `Ref`.  Beats me why it's OK to create a `HashMap` where the key doesn't implement `Hash`, but whatever.  I'll remove the lint and tests for `Ref` and `RefMut`.

clippy_lints/src/mut_key.rs
camsteffen: Good point. How about "a field with interior mutability behind a pointer like `Box<Cell<usize>>`"?

clippy_lints/src/mut_key.rs
kneasle: Huh?  The first example is a false positive, so this false negative can't be an instance of it.  I've updated the docs though to hopefully make for easier reading :smile:.

kneasle: I've rebased the changes into the same 3 commit messages as before, so feel free to close this conversation if the changes are good :grin:.

kneasle:
Wait hold on what's going on with the unit tests?  It seems to be that my local build of clippy is producing different test results to the build in the PR checker.  So when I `cargo dev bless` my local copy it causes the PR check to fail.  @camsteffen what d'ya reckon I should do?  I can update my PR branch to be what the test runner wants (which will let this PR be merged), but I feel like I can't be alone in having this discrepancy.

camsteffen:
Your test output is not including changes from recent PRs. Try rebasing, don't run with TESTNAME, and maybe try cargo clean. Or just don't check in unrelated changes.

clippy_lints/src/mut_key.rs
camsteffen: I was referring to the ###Example section.

kneasle:
Aha I've figured out what's going - I appear to not have `target-dir` not set in `<clippy>/.cargo/config`.  So it was compiling my new changes into `~/.build/rust` but the test runner was still reading the old version from `<clippy>/test`.

clippy_lints/src/mut_key.rs
kneasle: Ah I see!  The example looks fine to me - it's definitely on the verbose side, but it does correctly trigger the lint.  We could potentially remove the `impl` blocks and the example would still work.

kneasle:
Yes finally!  The tests are passing at last.

clippy_lints/src/mut_key.rs
camsteffen: Oh it wasn't firing for me because it doesn't fire on `let _: .. = ..` 🤦 . Yeah if you could fix that and trim out the impls that would be good.

kneasle: > Oh it wasn't firing for me because it doesn't fire on `let _: .. = ..` facepalm .

Ah yes I got gotcha-ed by that exact same thing when writing the tests :sweat_smile:

>  Yeah if you could fix that and trim out the impls that would be good

Right I'll do this.  Although, thinking about it, I think the lint becomes a false positive if I remove the `impl` blocks.  We can't `#[derive(Hash, PartialEq, Eq)]` because `AtomicUsize` doesn't impl any of those 3 traits.  So I think the example is as short as it can be without being subtly wrong.  What d'ya reckon @camsteffen?  If you're OK with it, then feel free to resolve this conversation :).

camsteffen: Makes sense

camsteffen:
Please squash some commits and I'll merge.

kneasle:
> Please squash some commits and I'll merge.

Rebased down to two commits :).

camsteffen:
Thanks! @bors r+

kneasle:
:tada::tada::tada::tada:

Issue #7569, opened 2021-08-16
New lint: `slice_indexing_after_deconstruction`


xFrednet:
### What it does

The lint checks for slices that come from value deconstruction and are only used to access individual slice values.

### Categories (optional)

- Kind: *pedantic* (This could also be *style*, but I'm not sure if we want it warn-by-default.)

### Advantages

Deconstruction of slices allows direct value access without using indexes on the slice, which can lead to panics. It also enables the user to give slice values a better name and indirectly check the correct length of the slice. See [\#7566 files](https://github.com/rust-lang/rust-clippy/pull/7566/files) as an example of how this can be an improvement.

Clippy already had a few instances where ICEs would have been prevented by using deconstruction of slices. 

### Example

```rust
fn main() {
    let slice: Option<&[u32]> = Some(&[1, 2, 3]);
    
    if let Some(slice) = slice {
        println!("{}", slice[0]);
    }
}
```

Could be written as:

```rust
fn main() {
    let slice: Option<&[u32]> = Some(&[1, 2, 3]);
    
    if let Some([first, ..]) = slice {
        println!("{}", first);
    }
}
```

### Notes:

* Clippy uses pattern matching extensively, often in  `if_chain` macros with `if let` statements. I think it would be reasonable for this lint to also check if statements inside `if_chain` macros.
* The lint probably shouldn't warn if the slice instance is used for something else besides direct value access.

camsteffen:
We have to pick some max threshold for indexes to lint. Like we don't want to lint `slice[100]`. Maybe `slice[3]` is a reasonable max to lint? In other words, suggest a slice pattern with up to 4 elements.

xFrednet:
Good suggestion, a threshold of 3 would catch all instance in Clippy that I've seen. We could also have this configurable and have the default be 3 :+1:.

PR #7643, opened 2021-09-07
New lint `index_refutable_slice` to avoid slice indexing 

xFrednet:
A new lint to check for slices that could be deconstructed to avoid indexing. This lint should hopefully prevent some panics in other projects and ICEs for us. See #7569 for an example

The implementation specifically checks for immutable bindings in `if let` statements to slices and arrays. Then it checks if these bindings are only used for value access using indices and that these indices are lower than the configured limit. I did my best to keep the implementation small, however the check was sadly quite complex. Now it's around 300 lines for the implementation and the rest are test.

---

Optional future improvements:
* Check for these instances also in `match` statements
* Check for mutable slice bindings that could also be destructed

---

changelog: New lint [`index_refutable_slice`]

I've already fixed a bunch of lint triggers in #7638 to make this PR smaller

Closes: #7569


camsteffen:
I'm afraid of false positives as this lint changes semantics. Maybe this is acceptable for a pedantic lint, but I'm not sure. It's like suspicious-pedantic...

```rust
if let Some(slice) = opt {
    if !slice.is_empty() {
        println!("first: {}", slice[0]);
    }
}
if let Some(slice) = opt {
    if interesting_stuff(slice) {
        // interesting_stuff never returns true for empty slices
        println!("first: {}", slice[0]);
    }
}
if let Some(slice) = foo.bar {
    if foo.is_interesting {
        // foo.is_interesting is only true if the slice is not empty
        println!("first: {}", slice[0]);
    }
}
```

xFrednet:
I was also a bit afraid of FPs. The implementation is therefore quite cautious, it at least covers all cases that I could think of and should only lint if the slice is used for indexing. I remove it from the list if it's used in any other way, but FPs could maybe still happen. The first two cases should definitely be covered, the third one should work as well. I'll add these tests tomorrow and also try to run lintcheck with it :upside_down_face:

camsteffen:
Yeah I'm wary of lints where we can never really be sure that we're giving a false positive. You can add heuristics to be cautious, but ultimately we can't know from static analysis why the coder might have certain assumptions about the length of the slice, and how those assumptions change at different points in the code. The heuristics remove false positives and add false negatives at the same time. I think the pedantic group allows for *picky* lints, but not necessarily more false positives. Maybe this can just go in nursery or restriction. @rust-lang/clippy thoughts?

camsteffen:
> I mainly wrote it with Clippy in mind xD

That's fine, but I also think the FP rate in Clippy will be lower than in most other domains. For example, we often have a slice of arguments and we simply expect the number of arguments to be one static value.

flip1995:
I like the idea of the lint. We could put it in nursery for now and try with the `lintcheck` tool how often it triggers in the wild. 

I haven't read the code, so I don't know what the scope of patterns is this lint triggers on. The tests seem to only address cases where the slice is immediately indexed after the `if let`. So I think the scope of this lint is an open question.

xFrednet:
Okay, then I'll change the group to nursery.

> The tests seem to only address cases where the slice is immediately indexed after the `if let`

The lint uses a `UseExprVisitor` to move through the entire block. The tests are rather simple, as you've noticed. I've already fixed the lint triggers inside Clippy in #7638 which displays some more complex example. I'll put it on the to-do list to add some more complex examples.

Sorry, for not adding them sooner, my semester is currently taking up a lot of time. It should be better in the upcoming week :upside_down_face:

camsteffen:
Alright, maybe I'm being overly pessimistic. :)

I have an idea for naming the lint/config if you like it. `index_refutable_slice` as in "index a slice binding from a refutable pattern". This conveys that a let statement would never be linted. For the config, `max_suggested_slice_pattern_length` as I think this is more "to the point" of what we want to configure. It could be reused by another lint, but if not that's fine.

clippy_lints/src/avoidable_slice_indexing.rs
```diff
@@ -0,0 +1,291 @@
...
+            }
+        }
+    }
+
+    extract_msrv_attr!(LateContext);
+}
+
+fn find_slice_values(cx: &LateContext<'_>, pat: &hir::Pat<'_>) -> Option<FxHashMap<hir::HirId, SliceLintInformation>> {
+    let mut removed_pat: FxHashSet<Symbol> = FxHashSet::default();
+    let mut ident_to_id: FxHashMap<Symbol, hir::HirId> = FxHashMap::default();
```
camsteffen, line 91: I like to track bindings by `HirId` rather that `Symbol`. Then I can use `cx.tcx.hir().name(id)` and `cx.tcx.hir().span(id)` when those are needed. Would that remove the need for `ident_to_id` and `pattern_spans`?

clippy_lints/src/avoidable_slice_indexing.rs
```diff
@@ -0,0 +1,291 @@
...
+                }
+            })
+            .collect::<Vec<_>>();
+        let pat_sugg = format!("[{}, ..]", pat_sugg_idents.join(", "));
+
+        span_lint_and_then(
+            cx,
+            AVOIDABLE_SLICE_INDEXING,
+            slice.ident.span,
+            "this slice can be deconstructured to avoid indexing",
```
camsteffen, line 160:
```suggestion
            "this binding can be a slice pattern to avoid indexing",
```

clippy_lints/src/avoidable_slice_indexing.rs
```diff
@@ -0,0 +1,291 @@
...
+            |diag| {
+                diag.multipart_suggestion(
+                    "try destructing the slice with a pattern here",
```
camsteffen, line 163:
```suggestion
                    "try using a slice pattern here",
```

clippy_lints/src/avoidable_slice_indexing.rs
```diff
@@ -0,0 +1,291 @@
...
+                        .map(|(index, span)| (*span, value_name(*index)))
+                        .collect(),
+                    Applicability::MaybeIncorrect,
+                );
+
+                // I thought about adding a note to the lint message to inform the user that these
+                // refactorings will remove the expressions that determine which access should be
+                // used. However, I decided against this, as `filter_lintable_slices` will only
+                // return slices where all access indices are known at compile time. They will
+                // therefore not have any side effects when they are removed.
```
camsteffen, line 186: Hmm what do you mean be "determine which access should be used"?

clippy_lints/src/avoidable_slice_indexing.rs
```diff
@@ -0,0 +1,291 @@
...
+        self.check_place(place, expr_id);
+    }
+
+    fn borrow(&mut self, place: &PlaceWithHirId<'tcx>, expr_id: hir::HirId, _bk: ty::BorrowKind) {
+        self.check_place(place, expr_id);
+    }
+
+    fn mutate(&mut self, place: &PlaceWithHirId<'tcx>, expr_id: hir::HirId) {
+        self.check_place(place, expr_id);
+    }
```
camsteffen, line 288: It seems like this could be accomplished with a normal visitor and `path_to_local_id`?

tests/ui/avoidable_slice_indexing/if_let_slice_destruction.rs
```diff
@@ -0,0 +1,166 @@
...
+    let mut slice: Option<[String; 1]> = Some([String::from("Penguin")]);
+    if let Some(ref mut slice) = slice {
+        slice[0] = String::from("Mr. Penguin");
+    }
+    println!("Use after modification: {:?}", slice);
+
+    // Mut access on reference
+    let mut slice: Option<[String; 1]> = Some([String::from("Cat")]);
+    if let Some(slice) = &mut slice {
+        slice[0] = String::from("Lord Meow Meow");
```
camsteffen, line 152: hahahahaha

clippy_lints/src/avoidable_slice_indexing.rs
```diff
@@ -0,0 +1,291 @@
...
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::{symbol::Ident, Span, Symbol};
+use rustc_trait_selection::infer::TyCtxtInferExt;
+use rustc_typeck::expr_use_visitor::{Delegate, ExprUseVisitor, Place, PlaceBase, PlaceWithHirId};
+use std::convert::TryInto;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// The lint checks for slices that come from value deconstruction and
+    /// are only used to access individual slice values.
```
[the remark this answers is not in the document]
xFrednet: Good point, I noticed while writing the lint that *deconstruction* is actually the wrong therm :upside_down_face:

clippy_lints/src/avoidable_slice_indexing.rs
xFrednet: This refers to the index expression. The new suggestion removes them completely (`slice[CONST]` → `slice_x`). However, this shouldn't have any side effects as the value is already constant.

I've changed the comment a bit to clearly state this :upside_down_face:

clippy_lints/src/avoidable_slice_indexing.rs
xFrednet: Also true, I at first used the provided information by the `ExprUseVisitor` but then switched over to this. Nice catch

clippy_lints/src/avoidable_slice_indexing.rs
xFrednet: Each binding inside a pattern has its own `HirId` even if they are later used as the same value. I at first tried to only have the `slices` map in the next line, but that missed other bindings with the same name. We therefore need to match all bindings somehow.

The implementation could use `cx.tcx.hir().name(id)` to avoid `ident_to_id` and use `HirId`s in `removed_pat`. However, each newly found binding would then need to first call `cx.tcx.hir().name(id)` on itself and then on every key inside `ident_to_id` to check if there is already an entry for it. I personally find this more readable and okay, as the scope is limited to only this function. Everything else uses the `HirId` to track the slice. :upside_down_face:

camsteffen: > Each binding inside a pattern has its own `HirId` even if they are later used as the same value.

I believe the `Pat` has a unique `HirId` but the `HirId` in `PatKind::Binding` should be the same.

xFrednet: Ahh yes, you're correct. I've at first used [`Pat::each_binding()`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/struct.Pat.html#method.each_binding) which has the described problem. I'll apply this refactoring with the Visitor refactoring :upside_down_face:

xFrednet:
I've updated the code now, this should hopefully address all comments and make the pipeline green ^^. 

@rustbot  label -S-waiting-on-author +S-waiting-on-review

xFrednet:
> I have an idea for naming the lint/config if you like it. index_refutable_slice as in "index a slice binding from a refutable pattern". This conveys that a let statement would never be linted. For the config, max_suggested_slice_pattern_length as I think this is more "to the point" of what we want to configure.

I missed this review comment earlier. Thank you for the suggestion, I like them :+1:. I'll add them in another commit

camsteffen:
Nah go ahead.

xFrednet:
I've also rebased on `upstream/master` now :upside_down_face:

tests/ui-toml/max_suggested_slice_pattern_length/avoidable_slice_indexing_limit.rs
```diff
@@ -0,0 +1,23 @@
+#![deny(clippy::index_refutable_slice)]
```
camsteffen, line 1: This file has the old lint name.

tests/ui-toml/max_suggested_slice_pattern_length/avoidable_slice_indexing_limit.rs
```diff
@@ -0,0 +1,23 @@
...
+        // This would usually not be linted but is included now due to the
+        // index limit in the config gile
+        println!("{}", slice[7]);
+    }
+}
+
+fn above_limit() {
+    let slice: Option<&[u32]> = Some(&[1, 2, 3]);
+    if let Some(slice) = slice {
+        // This will be linted as 8 is above the limit
```
camsteffen, line 15:
```suggestion
        // This will not be linted as 8 is above the limit
```

tests/ui/index_refutable_slice/if_let_slice_binding.stderr
```diff
@@ -0,0 +1,158 @@
...
+
+error: this binding can be a slice pattern to avoid indexing
+  --> $DIR/if_let_slice_binding.rs:19:17
+   |
+LL |     if let Some(slice) = slice {
+   |                 ^^^^^
+   |
+help: try using a slice pattern here
+   |
+LL |     if let Some([ref slice_0, ..]) = slice {
```
camsteffen, line 29: I don't think this should have `ref`?

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,278 @@
...
+    /// ```rust
+    /// let slice: Option<&[u32]> = Some(&[1, 2, 3]);
+    ///
+    /// if let Some(&[first, ..]) = slice {
+    ///     println!("{}", first);
+    /// }
+    /// ```
+    pub INDEX_REFUTABLE_SLICE,
+    nursery,
+    "avoid indexing on slices which could be deconstructed"
```
camsteffen, line 50: I believe this is the more common term
```suggestion
    "avoid indexing on slices which could be destructured"
```

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,278 @@
...
+
+                // Get info to later lint this exporession
+                if let Some(use_info) = self.slice_lint_info.get_mut(&local_id);
+                then {
+                    use_info.index_use.push((index_value, map.span(parent_expr.hir_id)));
+                    return;
+                }
+            }
+
+            // The slice was used for something else than indexing
```
camsteffen, line 273:
```suggestion
            // The slice was used for something other than indexing
```

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,278 @@
...
+
+    let slices: FxHashMap<hir::HirId, SliceLintInformation> = visitor
+        .slice_lint_info
+        .drain()
+        .filter(|(_key, value)| !value.index_use.is_empty())
+        .collect();
+    (!slices.is_empty()).then(|| slices)
+}
+
+struct SliceIndexLintingDelegate<'a, 'tcx> {
```
camsteffen, line 231: This is not a "delegate" anymore.

camsteffen (changes requested):
FWIW it's okay by me to squash in changes when it's a significant refactor.

xFrednet:
Just a small note for triage: I'm currently quite busy, but I still plan to continue this PR. Thank you for the review :upside_down_face:

clippy_lints/src/utils/conf.rs
```diff
@@ -284,6 +284,12 @@ define_Conf! {
     ///
     /// The list of unicode scripts allowed to be used in the scope.
     (allowed_scripts: Vec<String> = vec!["Latin".to_string()]),
+    /// Lint: INDEX_REFUTABLE_SLICE.
+    ///
+    /// The limit after which slice indexing will not be linted as avoidable.
+    ///
+    /// As an example, the default limit of `3` would allow `slice[3]` but lint `slice[2]`.
```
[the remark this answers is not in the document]
xFrednet: I'm not quite sure how to create a generic configuration documentation for this. :sweat_smile:  I've updated the comment now to say:

```rs
    /// The limit after which slice indexing will not be linted.
    ///
    /// As an example, the default limit of `3` would allow `slice[3]` but lint `slice[2]`.
```

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,278 @@
...
+
+                let slice_info = slices
+                    .entry(value_hir_id)
+                    .or_insert_with(|| SliceLintInformation::new(ident, needs_ref));
+                slice_info.pattern_spans.push(pat.span);
+            }
+        }
+    });
+
+    (!slices.is_empty()).then(|| slices)
```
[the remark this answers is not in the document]
xFrednet: Sure, that should be simple. :upside_down_face:

xFrednet:
I'm starting to revive this branch. I'll ping you and switch the labels once I'm done :upside_down_face:

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,278 @@
...
+        index_limit,
+    };
+
+    intravisit::walk_expr(&mut visitor, scope);
+
+    let slices: FxHashMap<hir::HirId, SliceLintInformation> = visitor
+        .slice_lint_info
+        .drain()
+        .filter(|(_key, value)| !value.index_use.is_empty())
+        .collect();
```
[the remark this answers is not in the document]
xFrednet: TIL, good call!

tests/ui-toml/max_suggested_slice_pattern_length/avoidable_slice_indexing_limit.rs
xFrednet: This is the new lint name, otherwise it would also trigger the "unknown lint" lint :thinking:

tests/ui/index_refutable_slice/if_let_slice_binding.stderr
xFrednet: Good catch, the logic to detect the need for refs was a bit flawed ^^. It should be fixed in  the next commit :+1:

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,278 @@
...
+                        .iter()
+                        .map(|(index, span)| (*span, value_name(*index)))
+                        .collect(),
+                    Applicability::MaybeIncorrect,
+                );
+
+                // I thought about adding a note to the lint message to inform the user that these
+                // refactorings will remove the index expression. However, I decided against this,
+                // as `filter_lintable_slices` will only return slices where all access indices are
+                // known at compile time. The removal should therefore not have any side effects.
```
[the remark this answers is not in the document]
xFrednet: Yes, this is referring to the expression inside the index. I was referring to a case like this:
```rs
fn give_index() -> usize {
    println!("I have side effects");
    println!("I print stuff");
    println!("I important function");
    1
}

fn main() {
    // ...
    if let Some(ref slice) = slice {
        println!("{:?}", slice[give_index()]);
    //                         ^^^^^^^^^^^^
    // Removing this also removes the side effects of `give_index`. This is
    // not directly stated by "and replacing the index expressions here".
    // But it's very much implied and wouln't be linted anyway as `give_index`
    // is not a const expression at compiletime
    // ...
}
```

xFrednet:
Thank you for waiting on this review! This should now be ready for review again @camsteffen :upside_down_face: 

> FWIW it's okay by me to squash in changes when it's a significant refactor.

Good to know, I've squashed the changes from your review into the commit as well.

clippy_lints/src/index_refutable_slice.rs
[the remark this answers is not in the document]
camsteffen: I see. So it sounds like "I thought about doing X in order to mitigate problem Y, but then I realized problem Y doesn't exist because Z". Perhaps reduce that to "Problem Y doesn't exist because Z". But I don't mean to linger on this.

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+}
+
+impl_lint_pass!(IndexRefutableSlice => [INDEX_REFUTABLE_SLICE]);
+
+impl LateLintPass<'_> for IndexRefutableSlice {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>) {
+        if_chain! {
+            if !in_macro(expr.span) || is_expn_of(expr.span, "if_chain").is_some();
```
camsteffen, line 74: `in_macro` was just removed

```suggestion
            if !expr.span.from_expansion() || is_expn_of(expr.span, "if_chain").is_some();
```

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,284 @@
...
+            let bound_ty = cx.typeck_results().node_type(pat.hir_id);
+            if let ty::Slice(inner_ty) | ty::Array(inner_ty, _) = bound_ty.peel_refs().kind() {
+                // The values need to use the `ref` keyword if they can't be copied.
+                // This will need to be adjusted if the lint want to support multable access in the future
+                let src_is_ref = bound_ty.is_ref() && binding != hir::BindingAnnotation::Ref;
+                let is_copy = cx
+                    .tcx
+                    .lang_items()
+                    .copy_trait()
+                    .map_or(false, |trait_id| implements_trait(cx, inner_ty, trait_id, &[]));
```
camsteffen, line 124: I think we have a `is_copy` util

clippy_lints/src/index_refutable_slice.rs
```diff
@@ -0,0 +1,284 @@
...
+fn lint_slice(cx: &LateContext<'_>, slice: &SliceLintInformation) {
+    let used_indices = slice
+        .index_use
+        .iter()
+        .map(|(index, _)| *index)
+        .collect::<FxHashSet<_>>();
+
+    let value_name = |index| format!("{}_{}", slice.ident.name, index);
+
+    if let Some(max_index) = used_indices.iter().max() {
```
camsteffen, line 147: You already checked `slice.index_use` is not empty so you can unwrap `max()`. Maybe move the empty check to this function.

clippy_lints/src/utils/conf.rs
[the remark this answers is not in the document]
camsteffen: Something like

> When Clippy suggests using a slice pattern, this is the maximum number of elements allowed in the slice pattern that is suggested. If more elements would be necessary, the lint is suppressed. For example, `[_, _, _, e, ..]` is a slice pattern with 4 elements.

Here I'm implying that `..` doesn't count. That's consistent with the current impl right? But maybe it should count? I don't have a strong opinion on that.

tests/ui-toml/max_suggested_slice_pattern_length/avoidable_slice_indexing_limit.rs
camsteffen: I mean the file name.

tests/ui-toml/max_suggested_slice_pattern_length/avoidable_slice_indexing_limit.rs
```diff
@@ -0,0 +1,23 @@
+#![deny(clippy::index_refutable_slice)]
+
+fn below_limit() {
+    let slice: Option<&[u32]> = Some(&[1, 2, 3]);
+    if let Some(slice) = slice {
+        // This would usually not be linted but is included now due to the
+        // index limit in the config gile
```
camsteffen, line 7:
```suggestion
        // index limit in the confif gile
```

tests/ui-toml/min_rust_version/min_rust_version.rs
```diff
@@ -59,10 +59,20 @@ fn manual_strip_msrv() {
     }
 }
 
+fn check_avoidable_slice_indexing() {
+    // This shouldn't trigger `clippy::avoidable_slice_indexing` as the suggestion
```
camsteffen, line 63: changed lint name

camsteffen (commented):
Just some minor stuff

clippy_lints/src/index_refutable_slice.rs
xFrednet: I found no simple way to move the `is_empty` check to this function. I usually try to avoid calling unwrap in Clippy code, and  here I think it's fine to add one level of indention. Is that alright with you? Otherwise, an `unwrap` call with an explanatory comment is also fine IMO

camsteffen: I was thinking to just remove the `retain` call but maybe I'm missing something. Yeah this is fine as is.

xFrednet: Alright, I missed that simple solution. Updated the code to remove the retain statement :upside_down_face:

camsteffen: Well now `filter_lintable_slices` doesn't filter anything 😆. Maybe rename to `find_slice_usages`.

Totally optional, but you could pass it `&mut found_slices` and use `expr_visitor` instead of `impl Visitor`.

xFrednet: The visitor still filters slices that are used for operations other than index operations. The remove method can be called in line 272.

camsteffen: Ah okay!

camsteffen (approved):
LGTM, thanks! Please squash.

xFrednet:
Done! Thanks for the review!

camsteffen:
Er maybe it's too late to r- 🤷 @bors r+

PR #7646, opened 2021-09-08
Target directory cleanup

camsteffen:
changelog: none

* .cargo/config now has `target-dir` specified so that it is inherited by child projects. The target directory needs to be shared with clippy_dev, but not necessarily at the project root. (cc #7625)
* Uses `std::env::current_exe` (and its parent directories) whenever possible
* `CLIPPY_DRIVER_PATH` and `TARGET_LIBS` are no longer required from rustc bootstrap (but `HOST_LIBS` still is). These can be removed from the rustc side after merging.
* `CLIPPY_DOGFOOD` and the separate target directory are removed. This was originally added to mitigate #7343.

r? @flip1995

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,7 +1,7 @@
 [alias]
 uitest = "test --test compile-test"
-dev = "run --target-dir clippy_dev/target --package clippy_dev --bin clippy_dev --manifest-path clippy_dev/Cargo.toml --"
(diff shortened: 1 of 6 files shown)


flip1995:
Tested this in rustc and it seems that everything still works. Changes LGTM and this simplifies things. So:

@bors r+

Thanks!

Issue #7632, opened 2021-09-04
`method_duplicate_name`


lengyijun:
### What it does

*What does this lint do?*

A struct has two methods with the same name: 
one defined in `impl trait`, another in normal `impl`.

We shouldn't lint two method in two traits.This may break a lot. (I'm not sure)

### Categories

`clippy::all`, `clippy::style`

*What is the advantage of the recommended code over the original code*

- Remove confusing method call. 

### Drawbacks

Maybe meet difficulty in renaming the methods.

### Example

```
trait T1 {
    fn foo(&self) {}
}

struct S;

impl S {
    fn foo(&self) {
        dbg!("impl");
    }
}

impl T1 for S {
    fn foo(&self) {
        dbg!("T1");
    }
}

fn bar<T: T1>(x: T) {
    x.foo(); // which foo?
}

fn main() {
    bar(S);

    let s = S;
    s.foo(); // which foo?
}
```

PR #7653, opened 2021-09-10
New lint: `same_name_method`

lengyijun:
changelog: ``[`same_name_method`]``
fix: https://github.com/rust-lang/rust-clippy/issues/7632

It only compares a method in `impl` with another in `impl trait for`
It doesn't lint two methods in two traits.

I'm not sure my approach is the best way. I meet difficulty in other approaches.

### DIFF
--- a/clippy_lints/src/same_name_method.rs
+++ b/clippy_lints/src/same_name_method.rs
@@ -1,0 +1,160 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use rustc_data_structures::fx::FxHashMap;
+use rustc_hir::def::{DefKind, Res};
+use rustc_hir::{Crate, Impl, ItemKind, Node, Path, QPath, TraitRef, TyKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::AssocKind;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::Symbol;
+use rustc_span::Span;
+use std::collections::{BTreeMap, BTreeSet};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// It lints if a struct has two method with same time:
+    /// one from a trait, another not from trait.
+    ///
+    /// ### Why is this bad?
+    /// Confusing.
+    ///
+    /// ### Example
+    /// ```rust
+    /// trait T {
+    ///     fn foo(&self) {}
+    /// }
+    ///
+    /// struct S;
+    ///
+    /// impl T for S {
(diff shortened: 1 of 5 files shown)


clippy_lints/src/same_name_method.rs
```diff
@@ -0,0 +1,160 @@
...
+    /// impl T for S {
+    ///     fn foo(&self) {}
+    /// }
+    ///
+    /// impl S {
+    ///     fn foo(&self) {}
+    /// }
+    /// ```
+    pub SAME_NAME_METHOD,
+    pedantic,
```
flip1995, line 37: On second thought, this should probably be a `restriction` lint.

llogiq:
Apart from making this a `restriction` lint, I think this looks merge-worthy. I don't care too much about struct names in tests.

lengyijun:
@llogiq Thx. I have removed the todo and looking for your reply.

lengyijun:
Thx! I have removed the todos.
Although not satisfied with  `restriction`, but being accepted is also a good start for this lint.

llogiq:
What's your take on restriction vs. pedantic? Why would you not be satisfied with the former?

lengyijun:
I have no deep insight into the category of lint. Please help we with this hard decision !
I just want this lint can be used more and get more feedback.

llogiq:
The idea behind `pedantic` is to group lints that are deemed beneficial to code style, but have enough false positives not to warn by default. `restriction` lints are for things that are not universally seen as problematic, but may be in some contexts.

As I don't see how your lints could have false positives, yet depending on the context people may choose to ignore it, this fits the `restriction` category somewhat better than `pedantic`.

lengyijun:
I agree on `restriction` now.

llogiq:
r=me with the changed lint category :+1:

lengyijun:
r=me

clippy_lints/src/same_name_method.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub SAME_NAME_METHOD,
+    restriction,
+    "two method with same name"
+}
+
+declare_lint_pass!(SameNameMethod => [SAME_NAME_METHOD]);
+
+struct ExistingName {
+    impl_methods: BTreeMap<SymbolStr, Span>,
+    trait_methods: BTreeMap<SymbolStr, Vec<Span>>,
```
camsteffen, line 45: These should have `Symbol` instead of `SymbolStr` to avoid needless interner access.

lengyijun:
@llogiq

llogiq:
Thank you @lengyijun ! @bors r+

PR #7661, opened 2021-09-10
Fix various redundant_closure bugs

### LINKED ISSUE #3071: Redundant closure lint recommends invalid code
I've found an issue in the [redundant_closure] lint where it'll suggest a closure which won't actually compile. 

I think what's happening is `clippy` sees a closure that just calls a function, and because the input/output types match up it thinks the closure is redundant. However, when you get rid of the closure it fails because of lifetimes/HKT. 

```rust
use std::path::Path;

fn redundant<P: AsRef<Path>>(path: P) {
    let path = path.as_ref();
    println!("{}", path.display());
}

fn use_it<F>(thunk: F)
where
    F: FnOnce(&str),
{
    thunk("/bin/bash");
}

fn main() {
    use_it(|s| redundant(s));
    // use_it(redundant); <- recommended code that doesn't compile
}
```

And `rustc` gives the following compilation error:

```
error[E0631]: type mismatch in function arguments
  --> src/main.rs:17:5
   |
3  | fn redundant<P: AsRef<Path>>(path: P) {
   | ------------------------------------- found signature of `fn(_) -> _`
...
17 |     use_it(redundant); //<- recommended code that doesn't compile
   |     ^^^^^^ expected signature of `for<'r> fn(&'r str) -> _`
   |
note: required by `use_it`

...

error[E0271]: type mismatch resolving `for<'r> <fn(_) {redundant::<_>} as std::ops::FnOnce<(&'r str,)>>::Output == ()`
  --> src/main.rs:17:5
   |
17 |     use_it(redundant); //<- recommended code that doesn't compile
   |     ^^^^^^ expected bound lifetime parameter, found concrete lifetime
   |
note: required by `use_it`
```

([playground](https://play.rust-lang.org/?gist=b633a139243508cb37505a1173a59032&version=stable&mode=debug&edition=2015))

This may also be a simpler example for #1439.

[redundant_closure]: https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#redundant_closure

### LINKED ISSUE #4002: False positive in redundant_closure when closure value's lifetime is constrained
```rust
pub struct Arg;

pub trait Trait {
    fn f(_: Arg);
}

fn take<T: 'static>(_: T) {}

pub fn f<G: Trait>() {
    take(|v| G::f(v));
}
```

Clippy recommends `take(G::f)`:

```console
warning: redundant closure found
  --> src/main.rs:10:10
   |
10 |     take(|v| G::f(v));
   |          ^^^^^^^^^^^ help: remove closure as shown: `G::f`
   |
   = note: #[warn(clippy::redundant_closure)] on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
```

The original code compiles, while the code suggested by Clippy does not:

```console
error[E0310]: the parameter type `G` may not live long enough
  --> src/main.rs:10:5
   |
9  | pub fn f<G: Trait>() {
   |          -- help: consider adding an explicit lifetime bound `G: 'static`...
10 |     take(G::f);
   |     ^^^^
   |
note: ...so that the type `fn(Arg) {<G as Trait>::f}` will meet its required lifetime bounds
  --> src/main.rs:10:5
   |
10 |     take(G::f);
   |     ^^^^
```

I believe the difference is that `|v| G::f(v)` is an ordinary 'static closure `(Arg) -> ()` that captures nothing and `G` does not appear in its type, while `G::f` is according to the compiler a `fn(Arg) {<G as Trait>::f}` in which `G` does appear. Due to `G` appearing in its official type, the closure cannot be considered 'static unless `G` is 'static.

Mentioning @dgreid who encountered this in Chrome OS.

---

<kbd>clippy 0.0.212 (fbb3a47 2019-04-14)</kbd>

camsteffen:
changelog: Fix various false negatives and false positives for [`redundant_closure`]

Closes #3071
Closes #4002

This lint is full of weird nuances and this is basically a re-write to tighten up the logic.

### DIFF
--- a/clippy_lints/src/disallowed_type.rs
+++ b/clippy_lints/src/disallowed_type.rs
@@ -48,7 +48,7 @@ impl DisallowedType {
         Self {
             disallowed: disallowed
                 .iter()
-                .map(|s| s.split("::").map(|seg| Symbol::intern(seg)).collect::<Vec<_>>())
+                .map(|s| s.split("::").map(Symbol::intern).collect::<Vec<_>>())
                 .collect(),
             def_ids: FxHashSet::default(),
             prim_tys: FxHashSet::default(),
--- a/clippy_lints/src/eta_reduction.rs
+++ b/clippy_lints/src/eta_reduction.rs
@@ -1,16 +1,16 @@
 use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::higher::VecArgs;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::ty::{implements_trait, type_is_unsafe_function};
 use clippy_utils::usage::UsedAfterExprVisitor;
-use clippy_utils::{get_enclosing_loop_or_closure, higher};
-use clippy_utils::{is_adjusted, iter_input_pats};
+use clippy_utils::{get_enclosing_loop_or_closure, higher, path_to_local_id};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::{def_id, Expr, ExprKind, Param, PatKind, QPath};
-use rustc_lint::{LateContext, LateLintPass, LintContext};
-use rustc_middle::lint::in_external_macro;
-use rustc_middle::ty::{self, ClosureKind, Ty};
+use rustc_hir::def_id::DefId;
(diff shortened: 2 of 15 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #7656, opened 2021-09-10
rsplit_once suggest error


flisky:
```
error: manual implementation of `rsplit_once`
  --> src/utility/mod.rs:47:5
   |
47 | /     std::any::type_name::<T>()
48 | |         .rsplitn(2, ':')
49 | |         .next()
   | |_______________^ help: try this: `Some(std::any::type_name::<T>().rsplit_once(':').map_or(std::any::type_name::<T>(), |x| x.0))`
   |
   = note: `-D clippy::manual-split-once` implied by `-D clippy::all`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_split_once
```

the `rsplitn` returns last to first, and `rsplit_once` returns `Option<(first, last)>`, so it should go to `|x| x.1`. Right?

PR #7663, opened 2021-09-10
Fix result order for `manual_split_once` when `rsplitn` is used

Jarcho:
fixes: #7656

changelog: Fix result order for `manual_split_once` when `rsplitn` is used


llogiq:
Thank you! @bors r+

PR #7664, opened 2021-09-10
bump clippy crates to edition 2021

matthiaskrgr:
Also helps with dogfooding edition 2021 a bit. :)
Tests passed locally.

---

changelog: bump edition from 2018 to 2021


xFrednet (approved):
This looks good to me. However, I would like to also get some feedback from @flip1995 before merging this :upside_down_face:.

r? @flip1995

flip1995:
Is there already a `rust_2021_idioms` lint group we can enable? This diff looks suspiciously small.

matthiaskrgr:
There is `rust_2021_compatibility` I turned that on now and fix the only warning :upside_down_face: 

````
warning: changes to closure capture in Rust 2021 will affect drop order
   --> clippy_utils/src/lib.rs:995:26
    |
995 |     v.allow_closure.then(|| v.captures)
    |                          ^^^----------
    |                             |
    |                             in Rust 2018, this closure captures all of `v`, but in Rust 2021, it will only capture `v.captures`
996 | }
    | - in Rust 2018, `v` is dropped here, but in Rust 2021, only `v.captures` will be dropped here as part of the closure
    |
   --> clippy_utils/src/lib.rs:16:9
    |
16  | #![warn(rust_2021_compatibility)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^
    = note: `#[warn(rust_2021_incompatible_closure_captures)]` implied by `#[warn(rust_2021_compatibility)]`
    = note: for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/disjoint-capture-in-closures.html>
help: add a dummy let to cause `v` to be fully captured
    |
995 |     v.allow_closure.then(|| { let _ = &v; v.captures })
    |                             +++++++++++++            +

````

File: clippy_utils/src/lib.rs [hunk 1]
```diff
@@ -990,7 +992,10 @@ pub fn can_move_expr_to_closure(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) ->
         captures: HirIdMap::default(),
     };
     v.visit_expr(expr);
-    v.allow_closure.then(|| v.captures)
+    v.allow_closure.then(|| {
+        let _ = &v;
+        v.captures
+    })
```
flip1995, line 998: We don't really care about fully capturing `v`, do we?

clippy_utils/src/lib.rs
```diff
@@ -8,10 +8,12 @@
 #![allow(clippy::missing_errors_doc, clippy::missing_panics_doc, clippy::must_use_candidate)]
 // warn on the same lints as `clippy_lints`
 #![warn(trivial_casts, trivial_numeric_casts)]
+#![warn(rust_2021_compatibility)]
 // warn on lints, that are included in `rust-lang/rust`s bootstrap
 #![warn(rust_2018_idioms, unused_lifetimes)]
 // warn on rustc internal lints
 #![warn(rustc::internal)]
+#![warn(rust_2021_compatibility)]
```
flip1995, line 16: Added twice. Also I'd prefer if this would just be added to the same `warn` attribute as `rust_2018_idioms`. But I'd understand if you don't want to go over every file again, so if you don't want, just ignore this :+1:

flip1995 (commented):
Nice, let's also keep the `rust_2021_compatibility` :+1:

camsteffen:
I think `rust_2021_compatibility` is only meant to be enabled *before* transitioning?

flip1995:
From [the rustc book](https://doc.rust-lang.org/rustc/lints/groups.html):

> Lints used to transition code from the 2018 edition to 2021

A rust-2021-idioms lint does not exist yet. So I guess we don't want to enable the compatibility group. But good to know, that only one of our closures "fails" this.

matthiaskrgr:
removed the `rust_2021_compatibility` warnings again.

flip1995:
@bors r+

Thanks!

PR #7669, opened 2021-09-14
New lint `if_then_panic`

### LINKED ISSUE #7645: New lint: Manual assert
### What it does

Detects `if`-then-`panic!` that can be replaced with `assert!`

### Categories (optional)

- Kind: style

It is simpler

### Drawbacks

None.

### Example

```rust
if !sad_people.is_empty() {
    panic!("there are sad people: {:?}", sad_people);
}
```

Could be written as:

```rust
assert!(sad_people.is_empty(), "there are sad people: {:?}", sad_people);
```

Lint name: `manual_assert` or `if_then_panic`

Labelray:
changelog: add the new lint [`if_then_panic`]
fix #7645

### DIFF
--- a/clippy_lints/src/if_then_panic.rs
+++ b/clippy_lints/src/if_then_panic.rs
@@ -1,0 +1,97 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::higher::PanicExpn;
+use clippy_utils::is_expn_of;
+use clippy_utils::source::snippet_with_applicability;
+use rustc_errors::Applicability;
+use rustc_hir::{Block, Expr, ExprKind, StmtKind, UnOp};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects `if`-then-`panic!` that can be replaced with `assert!`.
+    ///
+    /// ### Why is this bad?
+    /// `assert!` is simpler than `if`-then-`panic!`.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let sad_people: Vec<&str> = vec![];
+    /// if !sad_people.is_empty() {
+    ///     panic!("there are sad people: {:?}", sad_people);
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let sad_people: Vec<&str> = vec![];
+    /// assert!(sad_people.is_empty(), "there are sad people: {:?}", sad_people);
+    /// ```
+    pub IF_THEN_PANIC,
+    style,
+    "`panic!` and only a `panic!` in `if`-then statement"
+}
+
+declare_lint_pass!(IfThenPanic => [IF_THEN_PANIC]);
+
+impl LateLintPass<'_> for IfThenPanic {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if_chain! {
+            if let Expr {
+                kind: ExprKind:: If(cond, Expr {
+                    kind: ExprKind::Block(
+                        Block {
+                            stmts: [stmt],
+                            ..
+                        },
+                        _),
+                    ..
+                }, None),
+                ..
+            } = &expr;
+            if is_expn_of(stmt.span, "panic").is_some();
+            if !matches!(cond.kind, ExprKind::Let(_, _, _));
+            if let StmtKind::Semi(semi) = stmt.kind;
+            if !cx.tcx.sess.source_map().is_multiline(cond.span);
+
+            then {
+                let span = if let Some(panic_expn) = PanicExpn::parse(semi) {
+                    match *panic_expn.format_args.value_args {
+                        [] => panic_expn.format_args.format_string_span,
+                        [.., last] => panic_expn.format_args.format_string_span.to(last.span),
+                    }
+                } else {
+                    if_chain! {
+                        if let ExprKind::Block(block, _) = semi.kind;
+                        if let Some(init) = block.expr;
+                        if let ExprKind::Call(_, [format_args]) = init.kind;
+
+                        then {
+                            format_args.span
+                        } else {
+                            return
+                        }
+                    }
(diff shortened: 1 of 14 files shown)


camsteffen:
Use `cx.sess().source_map().is_multiline(span)` to make sure the condition fits on on line. Add a negative test for this.

I think providing a code suggestion is important for this lint. The fix may not be intuitive and people may not be aware that `assert!` accepts a message and formatting args. We probably need `PanicMacroExpn` similar to `FormatExpn`.

clippy_utils/src/higher.rs
```diff
@@ -608,3 +608,106 @@ pub fn is_from_for_desugar(local: &hir::Local<'_>) -> bool {
 
     false
 }
+
+/// A parsed `panic!` expansion
+pub struct PanicExpn<'tcx> {
+    /// Span of `panic!(..)`
+    pub call_site: Span,
+    /// Inner `format_args!` expansion
+    pub format_args: PanicArgsExpn<'tcx>,
```
camsteffen, line 617: We should reuse `FormatArgsExpn` here.

camsteffen:
I'll leave it to @giraffate do to a full review. But note I still don't see the multi-line check. @Labelray don't hesitate to ask if something doesn't make sense!

Labelray:
> I'll leave it to @giraffate do to a full review. But note I still don't see the multi-line check. @Labelray don't hesitate to ask if something doesn't make sense!

@camsteffen I didn't check the number of lines but I checked the number of statements. I think it is better.

camsteffen:
A long condition can be a single statement.

```rust
if some_function(foo, bar, baz) == another_function(hi, whats, up)
    && some_function(a, b, c) == another_function(a, b, c) {
    panic!();
}
```

clippy_lints/src/if_then_panic.rs
```diff
@@ -0,0 +1,96 @@
...
+                };
+                let mut applicability = Applicability::MachineApplicable;
+                let sugg = snippet_with_applicability(cx, span, "..", &mut applicability);
+
+                let cond_sugg =
+                if let ExprKind::DropTemps(Expr{kind: ExprKind::Unary(UnOp::Not, not_expr), ..}) = cond.kind {
+                    snippet_with_applicability(cx, not_expr.span, "..", &mut applicability).to_string()
+                } else {
+                    format!("!{}", snippet_with_applicability(cx, cond.span, "..", &mut applicability))
+                };
```
giraffate, line 83: nits: Doesn't rustfmt work?
```suggestion
                let cond_sugg = if let ExprKind::DropTemps(Expr{kind: ExprKind::Unary(UnOp::Not, not_expr), ..}) = cond.kind {
                    snippet_with_applicability(cx, not_expr.span, "..", &mut applicability).to_string()
                } else {
                    format!("!{}", snippet_with_applicability(cx, cond.span, "..", &mut applicability))
                };
```

clippy_lints/src/if_then_panic.rs
```diff
@@ -0,0 +1,96 @@
...
+
+                span_lint_and_help(
+                    cx,
+                    IF_THEN_PANIC,
+                    expr.span,
+                    "only a `panic!` in `if`-then statement",
+                    None,
+                    &format!("considering use `assert!({}, {})` instead", cond_sugg, sugg)
+                );
```
giraffate, line 92: Can this lint have a suggestion using `span_lint_and_sugg` method? The `// run-rustfix` line needs to be added to the beginning of the test file too.

clippy_lints/src/if_then_panic.rs
Labelray: I wrote the code as your suggest at first. But rustfmt generated a mistake.
```
error[internal]: line formatted, but exceeded maximum width (maximum: 120 (see `max_width` option), found: 126)
  --> /Users/labelray/Projects/rust-clippy/clippy_lints/src/if_then_panic.rs:78:78:121
   |
78 |                 let cond_sugg = if let ExprKind::DropTemps(Expr{kind: ExprKind::Unary(UnOp::Not, not_expr), ..}) = cond.kind {
   |                                                                                                                         ^^^^^^
   |

warning: rustfmt has failed to format. See previous 1 errors.
```

tests/ui/fallible_impl_from.stderr
```diff
@@ -40,6 +40,16 @@ LL |             panic!();
...
+error: only a `panic!` in `if`-then statement
+  --> $DIR/fallible_impl_from.rs:28:9
+   |
+LL | /         if i != 42 {
+LL | |             panic!();
+LL | |         }
+   | |_________^ help: try: `assert!(!i != 42, "explicit panic");`
+   |
+   = note: `-D clippy::if-then-panic` implied by `-D warnings`
+
```
giraffate, line 52: Can you add `#![allow(clippy::if_then_panic)]` and remove these warnings in this test file? It looks a little confusing.

giraffate (commented):
I made a comment but overall looks good, thanks!

giraffate (approved):
Can the conflict be resolved?

giraffate:
We follow [a no merge-commit policy](https://github.com/rust-lang/rust-clippy/blob/master/doc/basics.md#pr). Can you do a rebase, not a merge?

giraffate:
@bors r+

Thanks!

tests/ui/fallible_impl_from.stderr
pravic: If this suggestion even valid?

`!i != 42` is `(!i) != 42` which is not what is intended to be suggested here. It should be `!(i != 42)`.

camsteffen:
Hi @Jasperav, thanks for the feedback.

> This issue causes regression: `panic!` can be catch with `std::panic::catch_unwind`, but `assert!` not.

I don't think this is true, unless I misunderstood. See [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=a5b53124d61639ccfb86776667f39570).

> I would suggest reverting this lint or don't make it active by default

The lint is being moved to pedantic (off by default) in #7810.

> which... is weird. Why negate an integer?

This should be fixed from #7741.

PR #7671, opened 2021-09-14
Downgrade many_single_char_names to pedantic

dtolnay:
As suggested by @flip1995 in https://github.com/rust-lang/rust-clippy/issues/7666#issuecomment-918993215, by today's standards this lint would be considered `pedantic`.

This is one of the most widely suppressed Clippy lints on crates.io according to https://github.com/dtolnay/noisy-clippy.

In my opinion this lint is just too domain specific for Clippy to have reliable visibility into. Sure there are some cases where the author is just being lazy and could use a kick in the butt, but we're still left with an enormous number of suppressions where single chars are the most appropriate name. For example in the context of colors, a function using `h`, `s`, `l`, `r`, `g`, `b` is 100% sensible and spelling all those out is silly, but it's past the default lint threshold.

---

changelog: Moved [`many_single_char_names`] to `pedantic`


xFrednet (approved):

xFrednet:
The changes LGTM. Thank you for the updated list in #7666 and this PR!

Note: I've updated the change log entry to state the group the lint moved to :upside_down_face: 

@bors r+

PR #7677, opened 2021-09-16
fix bug for large_enum_variants

surechen:
Fix the discussion problem in the issue of https://github.com/rust-lang/rust-clippy/issues/7666#issuecomment-919654291

About the false positive problem of case:
```rust
enum LargeEnum6 {
    A,
    B([u8;255]),
    C([u8;200]),
}
```

changelog: Fix largest_enum_variant wrongly identifying the second largest variant.

### DIFF
--- a/clippy_lints/src/large_enum_variant.rs
+++ b/clippy_lints/src/large_enum_variant.rs
@@ -1,13 +1,14 @@
 //! lint when there is a large size difference between variants on an enum
 
 use clippy_utils::diagnostics::span_lint_and_then;
-use clippy_utils::source::snippet_opt;
+use clippy_utils::source::snippet_with_applicability;
 use rustc_errors::Applicability;
-use rustc_hir::{Item, ItemKind, VariantData};
+use rustc_hir::{Item, ItemKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::lint::in_external_macro;
 use rustc_middle::ty::layout::LayoutOf;
 use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::source_map::Span;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -58,6 +59,17 @@ impl LargeEnumVariant {
     }
 }
 
+struct FieldInfo {
+    ind: usize,
+    size: u64,
+}
+
+struct VariantInfo {
+    ind: usize,
+    size: u64,
+    fields_size: Vec<FieldInfo>,
+}
+
 impl_lint_pass!(LargeEnumVariant => [LARGE_ENUM_VARIANT]);
 
 impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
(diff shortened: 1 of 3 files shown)


tests/ui/large_enum_variant.stderr
```diff
@@ -31,6 +31,23 @@ help: consider boxing the large fields to reduce the total size of the enum
...
+note: and the second-largest variant is 8 bytes:
+  --> $DIR/large_enum_variant.rs:41:5
+   |
+LL |     StructLikeLittle { x: i32, y: i32 },
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+help: consider boxing the large fields to reduce the total size of the enum
+  --> $DIR/large_enum_variant.rs:39:5
+   |
+LL |     ContainingMoreThanOneField(i32, [i32; 8000], [i32; 9500]),
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
camsteffen, line 49: Hmm this suggestion is missing `Box`es?

surechen: > Hmm this suggestion is missing `Box`es?

Hi, thanks.  Yes, this situation should not trigger lint.

camsteffen: I think it actually should trigger. Though it would take a little more work to detect which fields to box. A simpler fix here would be to just change the `span_help` to `help` (the span made me think it was a code suggestion when it's actually not).

surechen: > I think it actually should trigger. Though it would take a little more work to detect which fields to box. A simpler fix here would be to just change the `span_help` to `help` (the span made me think it was a code suggestion when it's actually not).

Hi,  Should the result be like following two way？
1. 
Is ContainingMoreThanOneField(i32, [i32; 8000], [i32; 9500]),  -> 

enum LargeEnum3 {
    ContainingMoreThanOneField(i32, Box<[i32; 8000]>, Box<[i32; 9500]>),
    VoidVariant,
    StructLikeLittle { x: i32, y: i32 },
}

Or

2.

error: large size difference between variants
  --> $DIR/large_enum_variant.rs:39:5
   |
LL |     ContainingMoreThanOneField(i32, [i32; 8000], [i32; 9500]),
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this variant is 70004 bytes
   |
note: and the second-largest variant is 8 bytes:
  --> $DIR/large_enum_variant.rs:41:5
   |
LL |     StructLikeLittle { x: i32, y: i32 },
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: consider boxing the large fields in ContainingMoreThanOneField to reduce the total size of the enum

camsteffen: Either is good. The first is better.

surechen: > Either is good. The first is better.

Done.

clippy_lints/src/large_enum_variant.rs
```diff
@@ -113,23 +114,27 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
+                            let mut sugg: Vec<(Span, String)> = vec![];
+                            let fields = match def.variants[i].data {
+                                VariantData::Struct(fields, ..) | VariantData::Tuple(fields, ..) => fields,
+                                VariantData::Unit(..) => unreachable!(),
+                            };
+                            let length = variant.fields.len() as u64;
+                            for (ind, val) in variant.fields.iter().enumerate() {
+                                let ty = cx.tcx.type_of(val.did);
+                                if let Some(size_info) = cx.layout_of(ty).ok().map(|l| l.size.bytes()) {
+                                    if size_info >= self.maximum_size_difference_allowed / length {
```
camsteffen, line 126: This isn't quite right. We should Box as many fields as needed to bring the size difference below what is allowed.

surechen: > This isn't quite right. We should Box as many fields as needed to bring the size difference below what is allowed.

Done. Thanks for your patient review.

clippy_lints/src/large_enum_variant.rs
```diff
@@ -113,23 +114,35 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
-                                        format!("Box<{}>", snip),
-                                        Applicability::MaybeIncorrect,
-                                    );
-                                    return;
+
+                            let mut sugg: Vec<(Span, String)> = vec![];
+                            let fields = match def.variants[i].data {
+                                VariantData::Struct(fields, ..) | VariantData::Tuple(fields, ..) => fields,
+                                VariantData::Unit(..) => unreachable!(),
+                            };
```
camsteffen, line 122:
```suggestion
                            let fields = match def.variants[i].data.fields();
```

clippy_lints/src/large_enum_variant.rs
```diff
@@ -113,23 +114,35 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
+                            let mut fields_size = Vec::new();
+                            for (ind, val) in variant.fields.iter().enumerate() {
+                                let ty = cx.tcx.type_of(val.did);
+                                if let Some(size_info) = cx.layout_of(ty).ok().map(|l| l.size.bytes()) {
+                                    fields_size.push((ind, size_info));
                                 }
                             }
+                            fields_size.sort_by(|a, b| (a.1.cmp(&b.1)));
```
camsteffen, line 130: This is mostly repeating what has been done earlier in the function. Can we combine some of the logic? For instance, if this sorted Vec is available earlier, then the largest and second largest variants are simply the first two elements. Also there is too much tuple field access `.1` and structs would help.

camsteffen (changes requested):

clippy_lints/src/large_enum_variant.rs
surechen: > This is mostly repeating what has been done earlier in the function. Can we combine some of the logic? For instance, if this sorted Vec is available earlier, then the largest and second largest variants are simply the first two elements. Also there is too much tuple field access `.1` and structs would help.

Hi, thanks for reviewing.
For example:
enum LargeEnum8 {
     VariantOk(i32, u32),
     ContainingMoreThanOneField([i32; 8000], [i32; 2], [i32; 9500], [i32; 30]),
}
The first calculation is for the size of VariantOk and ContainingMoreThanOneField, and the second is calculating size for the fields of ContainingMoreThanOneField, such as [i32; 8000], [i32; 2]...
Because the fields of ContainingMoreThanOneField only need to be calculated under the condition of difference > maximum_size_difference_allowed, I tend to separate them.

camsteffen: The first loop does get the layout and size of every field of every variant, so there isn't any work being saved. You could have `VariantInfo { size: usize, fields: Vec<FieldInfo> }`, create a `Vec<VariantInfo>`, sort it by size, and then only sort `fields` for the largest variant.

surechen: > The first loop does get the layout and size of every field of every variant, so there isn't any work being saved. You could have `VariantInfo { size: usize, fields: Vec<FieldInfo> }`, create a `Vec<VariantInfo>`, sort it by size, and then only sort `fields` for the largest variant.

Yes. You are right, I got it wrong. fixed

clippy_lints/src/large_enum_variant.rs
```diff
@@ -68,69 +80,77 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
+                    size,
+                    fields_size,
+                });
             }
+            variants_size.sort_by(|a, b| (b.size.cmp(&a.size)));
 
-            if let (Some(largest), Some(second)) = (largest_variant, second_variant) {
-                let difference = largest.0 - second.0;
-
+            if variants_size.len() >= 2 {
```
camsteffen, line 110: This check should be sooner.

clippy_lints/src/large_enum_variant.rs
```diff
@@ -68,69 +80,77 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
-                                    return;
+
+                            let fields = def.variants[variants_size[0].ind].data.fields();
+                            variants_size[0].fields_size.sort_by(|a, b| (a.size.cmp(&b.size)));
+
+                            let mut sugg: Vec<(Span, String)> = vec![];
+                            while difference > self.maximum_size_difference_allowed
+                                && !variants_size[0].fields_size.is_empty()
+                            {
+                                if let Some(val) = variants_size[0].fields_size.pop() {
```
camsteffen, line 136: `pop` is always `Some` here. What about this?

```rust
let sugg = variants_size[0].fields_size.into_iter().rev()
    .take_while(|_| difference > self.maximum_size_difference_allowed)
    .map(|val| ...)
    .collect();
```

clippy_lints/src/large_enum_variant.rs
```diff
@@ -68,69 +80,77 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
+                                    difference = if difference > val.size {
+                                        difference - val.size
+                                    } else {
+                                        0
+                                    };
+                                    if let Some(snip) = snippet_opt(cx, fields[val.ind].ty.span) {
```
camsteffen, line 142: Use `snippet_with_applicability`

camsteffen: You could also suggest `Box<` and  `>` separately to avoid snippets.

clippy_lints/src/large_enum_variant.rs
```diff
@@ -68,72 +80,99 @@ impl<'tcx> LateLintPass<'tcx> for LargeEnumVariant {
...
+                            .fields_size
+                            .iter()
+                            .rev()
+                            .take_while(|val| {
+                                let judge = difference > self.maximum_size_difference_allowed;
+                                difference = if difference > val.size {
+                                    difference - val.size
+                                } else {
+                                    0
                                 };
```
camsteffen, line 148: This should be `difference = difference.saturating_sub(val.size)`. And this should be *after* `take_while` since you should only change `difference` for values that are used in the suggestion.

surechen: > This should be `difference = difference.saturating_sub(val.size)`. And this should be _after_ `take_while` since you should only change `difference` for values that are used in the suggestion.

Hi, Thanks for reviewing.  In fact I don’t understand the meaning here. I think it is for values that are used in the suggestion now. 
For example for  ContainingMoreThanOneField([i32; 8000], [i32; 2], [i32; 9500], [i32; 30]),this code can get [i32; 8000] and [i32; 9500] now. 
Do you mean change code like this？
.take_while(|val| {
                                difference = difference.saturating_sub(val.size);
                                difference > self.maximum_size_difference_allowed
                            })
If I  change it like this, it can only get  [i32; 9500];

camsteffen: I mean like this

```rust
    .take_while(|val| difference > self.maximum_size_difference_allowed)
    .map(|val| {
        difference = difference.saturating_sub(val.size);
        (
            ...
```

By performing the subtraction in `map`,  you would only subtract for values where the `take_while` condition is true.

surechen: > I mean like this
> 
> ```rust
>     .take_while(|val| difference > self.maximum_size_difference_allowed)
>     .map(|val| {
>         difference = difference.saturating_sub(val.size);
>         (
>             ...
> ```
> 
> By performing the subtraction in `map`, you would only subtract for values where the `take_while` condition is true.

Hi, I tried code like this before , but the compiler reported a compilation error. For the variable difference, there is a mutable borrowing and an immutable borrowing at the same time.

camsteffen: Ah, the good ol’ borrow checker! Well there's a couple things you could do here to appease the borrow checker and keep the same semantics. I think I would reach for `map_while` (you can add the feature flag for it).

surechen: > Ah, the good ol’ borrow checker! Well there's a couple things you could do here to appease the borrow checker and keep the same semantics. I think I would reach for `map_while` (you can add the feature flag for it).

Hi, I use map_while. In order to pass CI I need to remove #![feature(iter_map_while)], but rustc 1.57 cannot be installed via rustup yet, so I am not sure whether it will cause local errors for others.
https://github.com/rust-lang/rust-clippy/runs/3743685037#:~:text=error%3A%20the%20feature,rs%3A6%3A12

```
error: the feature `iter_map_while` has been stable since 1.57.0 and no longer requires an attribute to enable
  --> clippy_lints/src/lib.rs:6:12
   |
6  | #![feature(iter_map_while)]
   |            ^^^^^^^^^^^^^^
   |
  --> clippy_lints/src/lib.rs:13:45
   |
13 | #![cfg_attr(feature = "deny-warnings", deny(warnings))]
   |                                             ^^^^^^^^
   = note: `#[deny(stable_features)]` implied by `#[deny(warnings)]`

error: could not compile `clippy_lints` due to previous error
```

camsteffen: That's okay, Clippy is used as a binary, not a library.

surechen:
> Looking good! Please squash commits.

Done. Thanks.

camsteffen:
@bors r+

Thanks!

camsteffen:
@bors retry

Added changelog

Issue #7659, opened 2021-09-10
while_let_on_iterator should not trigger on borrowed iterators if the iterator is subsequently used.


peterjoel:
Lint name: while_let_on_iterator

I tried this code:

```rust
fn parse_value<V>(tokens: TokenStream) -> parse::Result<(ValueExpr<V>, TokenStream)>
where
    ValueExpr<V>: Parse,
{
    let mut it = tokens.into_iter();
    let mut value = TokenStream::new();
    while let Some(tt) = it.next() {
        if let TokenTree::Punct(p) = &tt {
            if p.as_char() == ',' {
                break;
            }
        }
        value.append(tt);
    }
    Ok((ValueExpr::parse.parse2(value)?, it.collect()))
}
```

I expected to see this happen:

This should not trigger the lint because the iterator is used again after the loop. The purpose of explicitly calling `next()` is to make it clear that the iterator may not be fully consumed in the loop and the remaining items can be processed elsewhere.

Instead, this happened: The lint triggers and tells me to use a `for` loop.

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.55.0 (c8dfcfe04 2021-09-06)
binary: rustc
commit-hash: c8dfcfe046a7680554bf4eb612bad840e7631c4b
commit-date: 2021-09-06
host: x86_64-apple-darwin
release: 1.55.0
LLVM version: 12.0.1
```

camsteffen:
The rustbot comment seems to be caused by #7627 @xFrednet

xFrednet:
It sadly seems that way. The bot is a bit inconsistent when it comes to mentions inside html comments. I'll create another fix PR

Jarcho:
It's hard to call this one a false positive as the suggestion is correct in that you can use a `for` loop there. You have a subjective point about why not to use one which I personally disagree with. I would assume you were using the iterator within the loop, not necessarily afterwards. It might be a good idea to split the into two so you could disable just part of it.

As an aside I would be writing the code more like this (if you like combinators):

```rust
let mut it = tokens.into_iter();
let value: TokenStream = (&mut it).take_while(|x| match x {
    TokenTree::Punct(p) => p.as_char() != ',',
    _ => true,
}).collect();
Ok((ValueExpr::parse.parse2(value)?, it.collect()))
```

jjl:
[...]

PR #7688, opened 2021-09-18
Add lint `ref_mut_iter_method_chain`

Jarcho:
Mentioned here: https://github.com/rust-lang/rust-clippy/issues/7659#issuecomment-920582174

changelog: New lint: [`ref_mut_iter_method_chain`]
[#7688](https://github.com/rust-lang/rust-clippy/pull/7688)

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -44,6 +44,7 @@ mod option_as_ref_deref;
 mod option_map_or_none;
 mod option_map_unwrap_or;
 mod or_fun_call;
+mod ref_mut_iter_method_chain;
 mod search_is_some;
 mod single_char_add_str;
 mod single_char_insert_string;
@@ -1861,6 +1862,30 @@ declare_clippy_lint! {
     "replace `.splitn(2, pat)` with `.split_once(pat)`"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Check for `&mut iter` followed by a method call.
+    ///
+    /// ### Why is this bad?
+    /// This requires using parenthesis to signify precedence.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let mut iter = ['a', 'b', '.', 'd'].iter();
+    /// let before_dot = (&mut iter).take_while(|&&c| c != '.').collect::<Vec<_>>();
+    /// let after_dot = iter.collect::<Vec<_>>();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let mut iter = ['a', 'b', '.', 'd'].iter();
+    /// let before_dot = iter.by_ref().take_while(|&&c| c != '.').collect::<Vec<_>>();
+    /// let after_dot = iter.collect::<Vec<_>>();
+    /// ```
+    pub REF_MUT_ITER_METHOD_CHAIN,
+    style,
+    "`&mut iter` used in a method chain"
+}
+
 pub struct Methods {
     avoid_breaking_exported_api: bool,
     msrv: Option<RustcVersion>,
@@ -1939,7 +1964,8 @@ impl_lint_pass!(Methods => [
     SUSPICIOUS_SPLITN,
     MANUAL_STR_REPEAT,
     EXTEND_WITH_DRAIN,
-    MANUAL_SPLIT_ONCE
+    MANUAL_SPLIT_ONCE,
+    REF_MUT_ITER_METHOD_CHAIN
 ]);
 
 /// Extracts a method call name, args, and `Span` of the method name.
@@ -1973,7 +1999,7 @@ impl<'tcx> LateLintPass<'tcx> for Methods {
             hir::ExprKind::Call(func, args) => {
                 from_iter_instead_of_collect::check(cx, expr, args, func);
             },
-            hir::ExprKind::MethodCall(method_call, ref method_span, args, _) => {
+            hir::ExprKind::MethodCall(method_call, ref method_span, args @ [self_arg, ..], _) => {
                 or_fun_call::check(cx, expr, *method_span, &method_call.ident.as_str(), args);
                 expect_fun_call::check(cx, expr, *method_span, &method_call.ident.as_str(), args);
                 clone_on_copy::check(cx, expr, method_call.ident.name, args);
@@ -1982,6 +2008,7 @@ impl<'tcx> LateLintPass<'tcx> for Methods {
                 single_char_add_str::check(cx, expr, args);
                 into_iter_on_ref::check(cx, expr, *method_span, method_call.ident.name, args);
                 single_char_pattern::check(cx, expr, method_call.ident.name, args);
+                ref_mut_iter_method_chain::check(cx, self_arg);
             },
             hir::ExprKind::Binary(op, lhs, rhs) if op.node == hir::BinOpKind::Eq || op.node == hir::BinOpKind::Ne => {
                 let mut info = BinaryExprInfo {
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -1867,19 +1867,17 @@ declare_clippy_lint! {
     /// Check for `&mut iter` followed by a method call.
     ///
     /// ### Why is this bad?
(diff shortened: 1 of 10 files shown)


clippy_lints/src/methods/mod.rs
```diff
@@ -1796,6 +1797,30 @@ declare_clippy_lint! {
     "replace `.splitn(2, pat)` with `.split_once(pat)`"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Check for `&mut iter` followed by a method call.
+    ///
+    /// ### Why is this bad?
+    /// This requires using parenthesis to signify precedence.
```
flip1995, line 1805: Let's be explicit here.

```suggestion
    /// Using `(&mut iter)._` requires using parenthesis to signify precedence.
```

clippy_lints/src/methods/mod.rs
```diff
@@ -1796,6 +1797,30 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// let mut iter = ['a', 'b', '.', 'd'].iter();
+    /// let before_dot = (&mut iter).take_while(|&&c| c != '.').collect::<Vec<_>>();
+    /// let after_dot = iter.collect::<Vec<_>>();
```
flip1995, line 1811: What does this `after_dot` example show?

clippy_lints/src/methods/ref_mut_iter_method_chain.rs
```diff
@@ -0,0 +1,40 @@
...
+
+pub(crate) fn check(cx: &LateContext<'_>, self_arg: &Expr<'_>) {
+    if_chain! {
+        if let ExprKind::AddrOf(BorrowKind::Ref, Mutability::Mut, base_expr) = self_arg.kind;
+        if !in_macro(self_arg.span);
+        if let Some(&iter_trait) = cx.tcx.all_diagnostic_items(()).get(&sym::Iterator);
+        if implements_trait(cx, cx.typeck_results().expr_ty(base_expr).peel_refs(), iter_trait, &[]);
+        then {
+            let snip_span = match base_expr.kind {
+                ExprKind::Unary(UnOp::Deref, e) if cx.typeck_results().expr_ty(e).is_ref() && !in_macro(base_expr.span)
```
flip1995, line 21: I assume this is to provide the correct suggestion in the last test case? What happens if something is dereferenced twice?

flip1995 (commented):
LGTM overall. Some small questions.

Not sure if this should be warn-by-default.

clippy_lints/src/methods/mod.rs
```diff
@@ -1796,6 +1797,30 @@ declare_clippy_lint! {
...
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let mut iter = ['a', 'b', '.', 'd'].iter();
+    /// let before_dot = iter.by_ref().take_while(|&&c| c != '.').collect::<Vec<_>>();
+    /// let after_dot = iter.collect::<Vec<_>>();
+    /// ```
+    pub REF_MUT_ITER_METHOD_CHAIN,
```
xFrednet, line 1819: The PR #7813 was just merged, which added a new `clippy::version` attribute to all lints. Just a heads-up that the attribute needs to added for a green CI after the next rebase :upside_down_face: 
```suggestion
    #[clippy::version = "1.58.0"]
    pub REF_MUT_ITER_METHOD_CHAIN,
```

clippy_lints/src/methods/mod.rs
Jarcho: Nothing related to the lint. It's more of a practical use case for using `&mut iter` in the first place.

clippy_lints/src/methods/ref_mut_iter_method_chain.rs
Jarcho: It's to make a nicer suggestion on a re-borrow. e.g.

```rust
let x = &mut iter;
(&mut *iter).next();
```

would become

```rust
let x = &mut iter;
iter.by_ref().next();
```

It also does happen to fix a precedence issue with prefix operators before a function call. Multiple can safely be removed so long as they don't call a custom deref implementation.

clippy_lints/src/methods/mod.rs
flip1995: Can you add a comment to the documentation then? Usually we only have good vs bad examples. Also that it is duplicated in the "use instead" example is a bit confusing IMO.

Jarcho: Just removed the line.

clippy_lints/src/methods/ref_mut_iter_method_chain.rs
Jarcho: The precedence issue is fixed now. It will keep the parenthesis if they're needed. This defeats the stated purpose of the lint, but it's an edge case anyways.

clippy_lints/src/methods/ref_mut_iter_method_chain.rs
```diff
@@ -0,0 +1,40 @@
...
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::REF_MUT_ITER_METHOD_CHAIN;
+
+pub(crate) fn check(cx: &LateContext<'_>, self_arg: &Expr<'_>) {
+    if_chain! {
+        if let ExprKind::AddrOf(BorrowKind::Ref, Mutability::Mut, base_expr) = self_arg.kind;
+        if !self_arg.span.from_expansion();
+        if let Some(&iter_trait) = cx.tcx.all_diagnostic_items(()).name_to_id.get(&sym::Iterator);
```
flip1995, line 16:
```suggestion
        if let Some(iter_trait) = cx.tcx.get_diagnostic_item(sym::Iterator);
```

should work here.

clippy_utils/src/source.rs
```diff
@@ -306,6 +306,147 @@ pub fn snippet_with_context(
...
+) -> String {
+    let (snip, is_mac_call) = snippet_with_context(cx, expr.span, ctxt, "(..)", app);
+
+    match snip {
+        Cow::Borrowed(snip) => snip.to_owned(),
+        Cow::Owned(snip) if is_mac_call => snip,
+        Cow::Owned(mut snip) => {
+            let ctxt = expr.span.ctxt();
+
+            // Attempt to determine if parenthesis are needed base on the target position. The snippet may have
```
flip1995, line 374: Isn't this the same as what `sugg::Sugg` does? Also I think the enum `ExprPosition` is pretty similar to the [`ExprPrecedence`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/util/parser/enum.ExprPrecedence.html) enum of rustc.

Would using the `Sugg` utility also work here? If not, do you think it would be better to expand the `Sugg` utility, rather than implementing this new utils function?

Jarcho: Adding to the left or right of an operator needs to do different things. `1+1` added to the left of a subtraction doesn't need parenthesis, but added to the right it does. `ExprPrecedence` doesn't allow that distinction.

Jarcho: `Sugg` currently doesn't allow adding parenthesis based on where the will be inserted. I'm in the middle of reworking `Sugg` now, but it's going to be a larger change to fit this in nicely.

flip1995: Hm, I prefer reusing `Sugg` here anyway. If there are unnecessary parenthesis, rustc will warn on that afterwards. As long as the suggestion is semantically correct, I don't mind if there are a pair of parenthesis too many.

Especially when you're working on `Sugg` anyway, I don't want to include a half-baked solution now. (Also big thanks for improving all of our utilities and not only the lints :heart:)

Jarcho: Working on #7986 which should be a nicer solution. We can hold off merging this until that's done.

flip1995: > We can hold off merging this until that's done.

Yes, let's do this. :+1:

xFrednet:
Hey @Jarcho, this is a ping from triage. Has there been any update to the blocked status?

xFrednet:

PR #7690, opened 2021-09-19
Change `while_let_on_iterator` suggestion to use `by_ref()`

Jarcho:
It came up in the discussion #7659 that suggesting `iter.by_ref()` is a clearer suggestion than `&mut iter`. I personally think they're equivalent, but if `by_ref()` is clearer to people then that should be the suggestion.

changelog: Change `while_let_on_iterator` suggestion when using `&mut` to use `by_ref()`

### DIFF
--- a/clippy_lints/src/loops/while_let_on_iterator.rs
+++ b/clippy_lints/src/loops/while_let_on_iterator.rs
@@ -8,7 +8,7 @@ use clippy_utils::{
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, ErasedMap, NestedVisitorMap, Visitor};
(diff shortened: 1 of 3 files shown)


xFrednet (approved):
Thank you for the PR.

I can see the reasoning why `by_ref()` might be preferred, even if they are equivalent. Using `by_ref()` looks a bit cleaner IMO, and it's also easier to find documentation for it. The std docs contains an example which illustrates the usage of `by_ref()` quite well. :upside_down_face: 

One related question: It seems like this resolves the discussion from #7659 should we close the issue after this got merged? (I'll r+ this PR now, as it looks good to me. We can close the issue manually afterwards)

Jarcho:
There's still the question of whether the lint should be split between the two cases (whether `by_ref()` is needed or not).

PR #7694, opened 2021-09-20
Retry on some download errors in lintcheck

Jarcho:
I'm currently on spotty wifi right now. It is shocking the number of things that break when you lose connection for a few seconds. Some 500 errors should probably also be retried, but this fixes my issue.

changelog: None


matthiaskrgr (approved):
Looks good to me, thanks! :)

PR #7705, opened 2021-09-23
Restriction lint for casts from function pointer to *any* numeric type

michaelsproul:
The existing lints for function pointer casts cover the cases where a cast is non-portable or would result in truncation, however there's currently no way to forbid numeric function pointer casts entirely.

I've added a new lint `fn_to_numeric_cast_any`, which allows one to ban numeric function pointer casts, including to `usize`. This is useful if you're writing high-level Rust and want to opt-out of potentially surprising behaviour, avoiding silent bugs from forgotten parentheses, e.g.

```rust
fn foo() -> u32 {
    10
}

fn main() {
    let _ = foo as usize; // oops, forgot to call foo and got a random address instead!
}
```

I'm open to suggestions for the naming of the lint, because `fn_to_numeric_cast_any` is a bit clunky. Ideally I'd call this lint `fn_to_numeric_cast`, but that name is already taken for the more specific lint.

I'm also open to changing the suggestion behaviour, as adding parentheses is only one of many possible ways to fix the lint.

changelog: add ``[`fn_to_numeric_cast_any`]`` restriction lint


clippy_lints/src/casts/fn_to_numeric_cast_any.rs
```diff
@@ -0,0 +1,34 @@
...
+        ty::FnDef(..) | ty::FnPtr(_) => {
+            let mut applicability = Applicability::MaybeIncorrect;
+            let from_snippet = snippet_with_applicability(cx, cast_expr.span, "x", &mut applicability);
+
+            span_lint_and_sugg(
+                cx,
+                FN_TO_NUMERIC_CAST_ANY,
+                expr.span,
+                &format!("casting function pointer `{}` to `{}`", from_snippet, cast_to),
+                "try",
```
camsteffen, line 27: Since we don't really know the coder's intention:

```suggestion
                "did you mean to invoke the function?",
```

clippy_lints/src/casts/fn_to_numeric_cast_any.rs
```diff
@@ -0,0 +1,34 @@
...
+    // We only want to check casts to `ty::Uint` or `ty::Int`
+    match cast_to.kind() {
+        ty::Uint(_) | ty::Int(..) => { /* continue on */ },
+        _ => return,
+    }
+
+    match cast_from.kind() {
+        ty::FnDef(..) | ty::FnPtr(_) => {
+            let mut applicability = Applicability::MaybeIncorrect;
+            let from_snippet = snippet_with_applicability(cx, cast_expr.span, "x", &mut applicability);
```
camsteffen, line 20: You can avoid the snippet by using `span_lint_and_then` and suggest using `cast_expr.span.hi()` and `"()"`.

clippy_lints/src/casts/mod.rs
```diff
@@ -251,6 +252,36 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// // Bad: forgot to call `fn1` before casting to `usize`
+    /// fn fn1() -> u16 {
+    ///     1
+    /// };
+    /// let _ = fn1 as usize;
+    ///
+    /// // Good: lint ensures that you remember to call `fn2`
```
camsteffen, line 274: This should be re-worded since the lint cannot assume that the coder intended to call the function.

camsteffen (changes requested):
Could we generalize the lint even more, ban a cast to *any* type, and name it `fn_cast`? That makes the lint name situation a little better.

clippy_lints/src/casts/fn_to_numeric_cast_any.rs
michaelsproul: Updated

clippy_lints/src/casts/fn_to_numeric_cast_any.rs
michaelsproul: Sorry I don't understand what you mean. I based this on the code in `fn_to_numeric_cast`

clippy_lints/src/casts/mod.rs
michaelsproul: Done

michaelsproul:
Thanks for the feedback @camsteffen! I've addressed all but one of your suggestions and I'm liking the new `fn_cast` name a lot!

clippy_lints/src/casts/fn_to_numeric_cast_any.rs
camsteffen: Something like this

```rust
span_lint_and_then(
    cx,
    FN_CAST,
    cast_expr.span,
    &format!("casting function pointer `{}` to `{}`", from_snippet, cast_to),
    |diag| {
        diag.span_suggestion(
            cast_expr.span.hi(),
            "did you mean to invoke the function?",
            "()".to_string(),
        );
    },
);
```

michaelsproul: Oh, I see, thanks!

The `from_snippet` is still used in the "casting function pointer `{}` to `{}`" message though, so I don't think it can be removed without altering that message.

I feel like the snippet adds something to both the error message and the help text, do you think we could merge as-is?

camsteffen: The zero-sized span would show something like

    f() as usize
     ++

which makes the change a little more clear, but not a big deal.

camsteffen:
Sorry, I'm second guessing what to do for the lint name. I didn't realize that this lint still allows `as fn()` ([function item](https://doc.rust-lang.org/reference/types/function-item.html) to [function pointer](https://doc.rust-lang.org/reference/types/function-pointer.html)), so I don't think `fn_cast` really works. Like you said earlier, `fn_to_numeric_cast` would be good but is already taken.

I feel that we need to rename `fn_to_numeric_cast` because its existence alongside this lint would be very confusing. So I suggest renaming it to `nonstandard_fn_cast`. This more clearly communicates the intent - "don't cast to a non-usize integer" - and that it's just a style thing.

For this new lint, I suggest `fn_cast_to_integer`. Importantly, this is a new name so it does not change the meaning of an existing lint. I'm using "integer" instead of "numeric" because Rust doesn't allow `func as f32`.

That leaves `fn_to_numeric_cast_with_truncation`, which would no longer have a neat similarity to any other lint names, but I think it's okay as is.

In sum:
* Rename `fn_to_numeric_cast` to `nonstandard_fn_cast`
* New lint: `fn_cast_to_integer`
* `fn_to_numeric_cast_with_truncation` - leave it

@rust-lang/clippy thoughts?

michaelsproul:
Alternatively we could keep the `fn_cast` name and forbid fn item to fn pointer casts, although it would forbid code like this (from `quickcheck`'s readme):

```rust
fn prop(xs: Vec<isize>) -> TestResult {
    if xs.len() != 1 {
        return TestResult::discard()
    }
    TestResult::from_bool(xs == reverse(&xs))
}
quickcheck(prop as fn(Vec<isize>) -> TestResult);
```

Maybe this pattern is rare enough that hitting it with a restriction lint is considered acceptable. I'm not familiar with the process for removing a lint entirely... is it painful?

flip1995:
>     * Rename `fn_to_numeric_cast` to `nonstandard_fn_cast`

I wouldn't rename it, because I think it makes sense to keep the name close to the `fn_to_numeric_cast_with_truncation` because those two lints are related.

I don't see a problem with keeping the old names and introduce a new `fn_cast_to_integer` lint. Especially since it is a restriction lint and the common Clippy user won't encounter this similarity in names, as long as they don't actively look for it. And at this point they probably read the documentation either way.

camsteffen:
> I wouldn't rename it

Yeah I agree...

Then it makes less sense to use "integer" for the new lint because it sounds more specific when it is actually less specific. So now I've come full circle and think the new lint should be `fn_to_numeric_cast_any`. 😂

camsteffen:
> Maybe this pattern is rare enough that hitting it with a restriction lint is considered acceptable.

We should decide what behavior we want for the lint, and name it accordingly. Not the other way around ☺️. I don't think linting function item to
function pointer would make sense because it is very low risk.

flip1995:
> I've come full circle and think the new lint should be `fn_to_numeric_cast_any`

By now you're my go-to person for lint names :smile: So I'd be fine with that.

Agreed on not linting fn -> fn_ptr.

camsteffen:
One minor thing and then ready to merge!

PR #7709, opened 2021-09-23
Implement `non_send_field_in_send_ty` lint

### LINKED ISSUE #7703: New Lint: `drop_non_send`
### What it does

Warns about a field in a `Send` struct that is neither `Send` nor `Copy`.

Sending the struct to another thread and drops it there will also drop the field in the new thread. This effectively changes the ownership of the field type and breaks the non-`Send` requirement which is unsound.

This is one of the most common unsound bug patterns we observed during [Rudra](https://github.com/sslab-gatech/Rudra) project.

Bug examples:

* [disrustor](https://github.com/sklose/disrustor/blob/966b949dae93d7c56f796d5c5f49be8e5a73a9c6/src/ringbuffer.rs#L4-L8) / RUSTSEC-2020-0150
* [noise_search](https://github.com/pipedown/noise/blob/5a582a7d0af1d7038646bfacaced40c3c19d4421/src/index.rs#L395-L398) / RUSTSEC-2020-0141
* [async-coap](https://github.com/google/rust-async-coap/blob/6a7b592a23de0c9d86ca399bf40ecfbf0bff6e62/async-coap/src/arc_guard.rs#L128-L132) / RUSTSEC-2020-0124
* [rusb](https://github.com/a1ien/rusb/blob/12ee91dd081428c610d3ccb5bba422198beee084/src/device_handle.rs#L107-L112) / RUSTSEC-2020-0098

### Categories (optional)

- Kind: Nursery (for now)

### Known Problems

* Raw pointers in data structures could cause false positives, such as `Vec<*const T>`. They can be partially mitigated by checking if any pointer type is used as a generic type parameter.
* A field that implements `Copy` but not `Send` can actually be a problem, e.g., [RUSTSEC-2020-0136](https://rustsec.org/advisories/RUSTSEC-2020-0136.html). However, they would be better handled in a separate lint because this lint focuses on unsoundness in drop and removing the `Copy` bound would cause too much false positives.

### Example

```rust
// Example from google/async-coap, RUSTSEC-2020-0124
#[derive(Debug, Clone)]
pub struct ArcGuard<RC, T> {
    inner: T,
    head: Arc<RC>, // This drops Arc<RC> (and in turn RC) which might not be `Send`
}

unsafe impl<RC, T: Send> Send for ArcGuard<RC, T> {} // There is no `RC: Send` bound here
```

@rustbot claim

Qwaz:
changelog: Implement [`non_send_fields_in_send_ty`] lint

Fixes #7703


clippy_lints/src/non_send_field_in_send_ty.rs
```diff
@@ -0,0 +1,129 @@
...
+                                    cx,
+                                    NON_SEND_FIELD_IN_SEND_TY,
+                                    field_hir_id,
+                                    field_span,
+                                    "non-`Send` field found in a `Send` struct",
+                                    |diag| {
+                                        diag.span_note(
+                                            item.span,
+                                            &format!(
+                                                "type `{}` doesn't implement `Send` when `{}` is `Send`",
```
xFrednet, line 91: The "when" statement seems a bit weird, as the type restriction is not conditional, from my understanding.

I would structure the error message to first show the item span in the `span_lint` with the error message and then add additional information. Making it look like this is more direct and the user knows where this originated from:

```
error: this implementation is unsound, as some fields in `MvccRwLock<T>` are `!Send`
  --> $DIR/non_send_field_in_send_ty.rs:23:1
   |
LL | unsafe impl<T> Send for MvccRwLock<T> {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the field type of `lock` doesn't implement `Send`
  --> $DIR/non_send_field_in_send_ty.rs:20:5
   |
LL |     lock: Mutex<Box<T>>,
   |     ^^^^^^^^^^^^^^^^^^^
   |
   = help: use a type that implements `Send`
   = help: or consider adding a `T: Send` bound
```
I've also changed the lint messages in the example to be a bit more direct. The second help text should only be added when the type actually uses a generic. You can use `clippy_utils::is_lint_allowed(cx, NON_SEND_FIELD_IN_SEND_TY, field_hir_id)` to check for `#[allow]` attributes on the field itself.

clippy_lints/src/non_send_field_in_send_ty.rs
```diff
@@ -0,0 +1,129 @@
...
+            if let Some(trait_ref) = &hir_impl.of_trait;
+            if let Some(trait_id) = trait_ref.trait_def_id();
+            if send_trait == trait_id;
+            if let ImplPolarity::Positive = hir_impl.polarity;
+            if let Some(ty_trait_ref) = cx.tcx.impl_trait_ref(item.def_id);
+            if let self_ty = ty_trait_ref.self_ty();
+            if let ty::Adt(adt_def, impl_trait_substs) = self_ty.kind();
+            then {
+                for variant in &adt_def.variants {
+                    for field in &variant.fields {
```
xFrednet, line 65: The current lint implementation emits one message per field that doesn't implement `Send`. It would be really nice, if all `!Send` fields would be collected and then emitted as one message per struct. That would require a lint message restructuring as suggested below :upside_down_face:

xFrednet (commented):
The implementation logic is already looking good. I find it awesome to see the security lints coming to Clippy. Thank you for the research you've done in this regard and the work you're putting in the implementation!

Most of my comments are in regard to the documentation and the lint message construction. :)

clippy_lints/src/non_send_field_in_send_ty.rs
Qwaz: Awesome, I also wanted to implement the lint that way but I wasn't aware of `is_lint_allowed()`

Qwaz:
The push was to resolve the merge conflict. I'll start working on the feedbacks shortly.

tests/ui/non_send_field_in_send_ty.stderr
```diff
@@ -0,0 +1,142 @@
...
+LL |     data: Vec<UnsafeCell<T>>,
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: `-D clippy::non-send-field-in-send-ty` implied by `-D warnings`
+note: type `std::vec::Vec<std::cell::UnsafeCell<T>>` doesn't implement `Send` when `RingBuffer<T>` is `Send`
+  --> $DIR/non_send_field_in_send_ty.rs:15:1
+   |
+LL | unsafe impl<T> Send for RingBuffer<T> {}
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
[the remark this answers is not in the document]
Qwaz: I intentionally limited the suggestion logic to direct uses of type parameters because I don't know how to detect the correct bound for type parameters inside a struct. `T: Send` is insufficient in some cases (e.g., `Arc<T>: Send` requires `T: Send + Sync`), and showing "add `T: Send` bound" would be incorrect in such cases. Is there a compiler API to detect such bound? Otherwise, I think the current implementation is the safest option.

Another option considered is to make a suggestion like `UnsafeCell<T>: Send` bound. This would be always correct, but it suggests non-idiomatic bound in many cases (`T: Send` is usually preferred), so I chose to not give any suggestions and rely on programmers in those cases.

clippy_lints/src/non_send_field_in_send_ty.rs
Qwaz: > The "when" statement seems a bit weird, as the type restriction is not conditional, from my understanding.

Trait bound on type parameters may make `Send` impl conditional:
```rust
struct MyStruct {
    field: T,
}

unsafe impl<T: MyTrait> Send for MyStruct<T> {}
```

What I wanted to express was that `T: MyTrait` precondition (= "when `MyStruct<T>` is `Send`") does not guarantee that `T: Send`.

With that being clarified, I really like your suggestion of the lint message. I think specifying the unsafe impl first and saying "this implementation is unsound" provides a clear context for the users.

clippy_lints/src/non_send_field_in_send_ty.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// // There is no `RC: Send` bound here
+    /// unsafe impl<RC, T: Send> Send for ArcGuard<RC, T> {}
+    ///
+    /// #[derive(Debug, Clone)]
+    /// pub struct ArcGuard<RC, T> {
+    ///     inner: T,
+    ///     // Possibly drops `Arc<RC>` (and in turn `RC`) on a wrong thread
+    ///     head: Arc<RC>
+    /// }
+    /// ```
```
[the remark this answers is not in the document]
Qwaz: I would like to keep the generic type parameter in the example because the most common cases of this bug caused by not specifying correct bound on `T` while using implementing `Send` on `ADT<T>`. I'll try to simplify the example a little bit while keeping it. The current example is from an actual bug case, which could be a little complicated as an example.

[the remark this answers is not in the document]
Qwaz: I've updated the example. How does it look now?

Qwaz:
I've addressed the PR feedbacks. Could you take another look at the PR?

I left a few comments above unresolved, so that it is easier for you to leave additional feedbacks. Please resolve them if you are satisfied :)

tests/ui/non_send_field_in_send_ty.stderr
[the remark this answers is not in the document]
xFrednet: Good point, I forgot the fact that implementations can also have several bounds :sweat_smile:. It's unlikely that the compiler provides an API for this, as it's not something the compiler needs. It's a good decision to not include an incomplete help message. :+1:

Then I think it's sufficient to have the solution of *adding bounds* listed in the lint documentation. You can resolve this, as it's already listed as a viable solution :upside_down_face:

xFrednet:
Thank you, I'll have a look at it over the week :upside_down_face:

tests/ui-toml/toml_unknown_key/conf_unknown_key.stderr
```diff
@@ -1,4 +1,4 @@
...
+error: error reading Clippy's configuration file `$DIR/clippy.toml`: unknown field `foobar`, expected one of `avoid-breaking-exported-api`, `msrv`, `blacklisted-names`, `cognitive-complexity-threshold`, `cyclomatic-complexity-threshold`, `doc-valid-idents`, `too-many-arguments-threshold`, `type-complexity-threshold`, `single-char-binding-names-threshold`, `too-large-for-stack`, `enum-variant-name-threshold`, `enum-variant-size-threshold`, `verbose-bit-mask-threshold`, `literal-representation-threshold`, `trivial-copy-size-limit`, `pass-by-value-size-limit`, `too-many-lines-threshold`, `array-size-threshold`, `vec-box-size-threshold`, `max-trait-bounds`, `max-struct-bools`, `max-fn-params-bools`, `warn-on-all-wildcard-imports`, `disallowed-methods`, `disallowed-types`, `unreadable-literal-lint-fractions`, `upper-case-acronyms-aggressive`, `cargo-ignore-publish`, `standard-macro-braces`, `enforced-import-renames`, `allowed-scripts`, `enable-raw-pointer-heuristic`, `third-party` at line 5 column 1
```
flip1995, line 1: Can you add a test where this configuration is not the default. See the [`ui-toml`](https://github.com/rust-lang/rust-clippy/tree/master/tests/ui-toml) tests on how to do that.

clippy_lints/src/utils/conf.rs
```diff
@@ -284,6 +284,10 @@ define_Conf! {
     ///
     /// The list of unicode scripts allowed to be used in the scope.
     (allowed_scripts: Vec<String> = vec!["Latin".to_string()]),
+    /// Lint: NON_SEND_FIELD_IN_SEND_TY.
+    ///
+    /// Whether to apply the raw pointer heuristic in `non_send_field_in_send_ty` lint.
+    (enable_raw_pointer_heuristic: bool = true),
```
xFrednet, line 290: This documentation will currently only be displayed in the lint description. A more generic name and description usually helps with reusability. 
```suggestion
    /// Whether to apply the raw pointer heuristic to determine if a type is `Send`
    (enable_raw_pointer_heuristic_for_send: bool = true),
```
This will rename the value in the `.toml` file. You don't need to change the field inside the lint implementation.

clippy_lints/src/non_send_field_in_send_ty.rs
```diff
@@ -0,0 +1,229 @@
...
+                }
+
+                if !non_send_fields.is_empty() {
+                    span_lint_and_then(
+                        cx,
+                        NON_SEND_FIELD_IN_SEND_TY,
+                        item.span,
+                        &format!(
+                            "this implementation is unsound, as some fields in `{}` are `!Send`",
+                            self_ty
```
xFrednet, line 122: The display function of `Ty<>` sadly prefixed the entire path to the type, this can result in lint message with long paths. To keep the message a bit more concise, it would be better if the type can be copied from the source code:
```rs
clippy_utils::source::snippet(cx, self_ty.span, "..")
```

See #7212 for further information. The same thing should be applied for the help and note messages :upside_down_face:

xFrednet (commented):
This version is looking good, I think the logic is ready. Your lint test cases are awesome, I found a test case for everything I was expecting. There are still some NITs when it comes to the lint messages and some formatting, but all of them should hopefully be simple to fix. Then it should be ready for merge :upside_down_face:.

You can also rebase right away to resolve the merge conflicts.

clippy_lints/src/non_send_field_in_send_ty.rs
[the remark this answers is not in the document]
xFrednet: It looks good, thank you!

clippy_lints/src/non_send_field_in_send_ty.rs
```diff
@@ -0,0 +1,229 @@
...
+        true
+    } else {
+        // The type is known to be `!Send` and `!Copy`
+        match ty.kind() {
+            ty::Tuple(_) => ty
+                .tuple_fields()
+                .all(|ty| ty_allowed_with_raw_pointer_heuristic(cx, ty, send_trait)),
+            ty::Array(ty, _) | ty::Slice(ty) => ty_allowed_with_raw_pointer_heuristic(cx, ty, send_trait),
+            ty::Adt(_, substs) => {
+                if contains_raw_pointer(cx, ty) {
```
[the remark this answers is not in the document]
Qwaz: `walk()` returns all nested types while `substs` only returns a single level generic params. I'll mention "nested" in the comment for `contains_raw_pointer`.
Example: `MyStruct<P, Box<Q, R>>` - `[P, Box<Q, R>, Q, R]` versus `[P, Box<Q, R>]`

clippy_lints/src/non_send_field_in_send_ty.rs
Qwaz: Using the code span will rarely show incomplete type. For instance, consider the output of `Complex` testcase.

```
error: this implementation is unsound, as some fields in `Complex<Q, std::sync::MutexGuard<'static, bool>>` are `!Send`
  --> $DIR/non_send_field_in_send_ty.rs:125:1
   |
LL | unsafe impl<Q: Send> Send for Complex<Q, MutexGuard<'static, bool>> {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: the field `field2` has type `std::sync::MutexGuard<'static, bool>` which is `!Send`
  --> $DIR/non_send_field_in_send_ty.rs:119:5
   |
LL |     field2: B,
   |     ^^^^^^^^^
   = help: use a thread-safe type that implements `Send`
```

If we use span, the output message would become "has type `B`" although `B` is replaced with a concrete type `MutexGuard<'static, bool>`. Would this be okay?

Qwaz: Decided to omit the type in the "note" part, as it becomes redundant with the code snippet if we use span to print it.

Qwaz:
Finished addressing the comments :)

xFrednet (commented):
Thank you for the fast changes! Three more NITs that I've missed so far (The last reviews mainly focussed on the logic). Then I'm happy to merge this :upside_down_face:.

I've also used our [lintcheck tool](https://github.com/rust-lang/rust-clippy/blob/master/lintcheck/README.md) and everything looks fine. The lint had no false positives for our default list of crates.

Regarding the lint group, you set the group to `nursery` as a precaution, is that correct? I think it would be good to keep it, until the next beta has been branched off (sometime in the next 7 days). Then I would move the lint to the `suspicious` group to have it warn-by-default and get some feedback from the nightly users. We currently don't have a process to automatically to do this. Therefore, we should create an issue to change the group, once this is merged. :upside_down_face:

Qwaz:
> Regarding the lint group, you set the group to `nursery` as a precaution, is that correct? I think it would be good to keep it, until the next beta has been branched off (that should be in about 13 - 19 days). Then I would move the lint to the `suspicious` group to have it warn-by-default and get some feedback from the nightly users.

I initially put `nursery` category because I was a little intimidated by "most likely" part of the description "code that is most likely wrong or useless". However, your report about lintcheck result regained me some confidence :) I think it's a good idea to start collecting feedbacks from the nightly users.

The last commit should have addressed the review comments. I'll restart working on uninit_vec lint after this.

xFrednet (approved):
Everything looks good to me. Thank you for all the work you put into this lint! :)

xFrednet:
@bors r+

FYI: I updated the lint description with the new lint name :upside_down_face:. And this comment should merge it :tada:

PR #7713, opened 2021-09-24
Add known problem to redundant_clone suggestion

marcospb19:
changelog:

Add note in known problems list of [`redundant_clone`] explaining that it might fail if the `Clone` has some unexpected behavior.

### DIFF
--- a/clippy_lints/src/redundant_clone.rs
+++ b/clippy_lints/src/redundant_clone.rs
@@ -43,6 +43,9 @@ declare_clippy_lint! {
     /// ### Known problems
     /// False-negatives: analysis performed by this lint is conservative and limited.
     ///
+    /// Sugestion causes errors: If the `Clone` implementation returns a different
+    /// value, removing the `.clone()` call might change the behavior of your code.
+    ///
     /// ### Example
     /// ```rust
     /// # use std::path::Path;


marcospb19:
Here's an example where running `cargo clippy --fix` will trigger `clone_redundant` and applying the change will modify the value, causing the `assert_eq!` to fail:

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=231d825230cec2cad52f3c33c879bc9c).

```rust
#[derive(Debug)]
pub struct Integer(i32);

impl Clone for Integer {
    fn clone(&self) -> Self {
        Integer(i32::MIN)
    }
}

fn main() {
    let integer = Integer(i32::MAX);
    assert_eq!(integer.clone().0, i32::MIN);
}
```

This is because there are no guarantees that `Clone::clone` does not change the value, so this could silently introduce bugs to codebases.

camsteffen:
`Clone` is for "duplicating values" so I would say that is a bug in `impl Clone`...

marcospb19:
It is, I agree.

Still, `Clippy` changing code semantics and introducing bugs silently is something, at least we should document it.

Maybe you mean that "known problems" is not the right section for this? So we should add a note instead?

Or is it a _false positive_? Because in my example `.clone()` is not redundant.

marcospb19:
(force-pushing due to poor writing by me, I previously wrote "might change your code", which was stupidly obvious and redundant)

flip1995:
I agree with @camsteffen here. I would apply the principle of "garbage in, garbage out" here. Or to put it in logic terms of an implication: `false => true | false` (Everything can be derived from a `false` assumption). If your code is just wrong (in a way that can't be statically checked), then you can't rely on your tools to do the right thing.

If we'd had to take into account every possible error that a user could make and think about how a lint could be wrong in this case, we'd had to add a whole bunch of documentation to many lints. With the same reasoning as in this PR, the [`cmp_owned`](https://rust-lang.github.io/rust-clippy/master/index.html#cmp_owned) lint is also wrong. Other examples, that directly come to mind are [`derive_hash_xor_eq`](https://rust-lang.github.io/rust-clippy/master/index.html#derive_hash_xor_eq) and [`derive_ord_xor_partial_ord`](https://rust-lang.github.io/rust-clippy/master/index.html#derive_ord_xor_partial_ord). If you want the impls be different, the lint is wrong, but that's just because you violate the documented precondition.

Even though it's "just documentation" we don't want to get inconsistent here. If we document something like this for this lint, it also implies, that we take care of such things in other lint documentations. But we really don't want to do that.

---

TL;DR: The `Known Problems` section is for known false positives/negatives and bugs in the lint. I don't consider this a FP/FN/bug in the lint though.

marcospb19:
Understood :slightly_smiling_face:, closing.

PR #7722, opened 2021-09-26
Stop suggesting a float truncation that is not shorter

### LINKED ISSUE #7721: False positive in excessive_precision for floats that are already in shortest form
```rust
fn main() {
    let _ = 0.123_456_789_012_345_67;
}
```

```console
warning: float has excessive precision
 --> src/main.rs:2:13
  |
2 |     let _ = 0.123_456_789_012_345_67;
  |             ^^^^^^^^^^^^^^^^^^^^^^^^ help: consider changing the type or truncating it to: `0.123_456_789_012_345_66`
  |
  = note: `#[warn(clippy::excessive_precision)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#excessive_precision
```

Clippy's suggestion is the following diff:

```diff
-     let _ = 0.123_456_789_012_345_67;
+     let _ = 0.123_456_789_012_345_66;
```

It seems wrong to call this excessive precision when both representations have identical precision. Probably there is a reason that the one the human wrote is more meaningful in context and should be left as is.

<br>

### Meta

```
rustc 1.57.0-nightly (addb4da68 2021-09-25)
binary: rustc
commit-hash: addb4da686a97da46159f0123cb6cdc2ce3d7fdb
commit-date: 2021-09-25
release: 1.57.0-nightly
LLVM version: 13.0.0
```

dtolnay:
Fixes #7721.

Previously Clippy would say that a number has excessive precision even if it has the minimum possible precision for the floating point value that it corresponds to.

changelog: Fix [`excessive_precision`] being triggered on floats that are already written in shortest form


giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #7726, opened 2021-09-27
Unseparated literal suffix

### LINKED ISSUE #7658: unseparated_literal_suffix is very opinion oriented `0u8` versus `0_u8`
I just found [unseparated_literal_suffix](https://rust-lang.github.io/rust-clippy/master/index.html#unseparated_literal_suffix) with https://github.com/Geal/nom/pull/1387 and I believe this have never been talk about:

- This has been advice by [unnecessary_cast](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast) and so I have a problem, `0 as u8` is very clear, clippy lint to just typed the `0` make sense to me, but then advice to add an `_` to make it more clear ? when the original code is very clear ?
- `0u8` is used in a LOT in rust literature without any problem I don't think it's more unclear than `0_u8`
- `0_u8` is longer to write
- rustfmt doesn't fmt `0u8` to `0_u8`
- there is no talk about this
- never see `0_u8` anywhere before

Thus I don't understand the lint (and the consequence to other lint) come a little from nowhere in my opinion.

dswij:
Closes #7658

Since `literal_suffix` style is opinionated, we should disable by default and only enforce if it's stated as so.

changelog: [`unseparated_literal_suffix`] is renamed to `literal_suffix`, adds a new configuration `literal-suffix-style` to enforce a certain style writing literal_suffix. Possible values for `literal-suffix-style`: `"separated"`, `"unseparated"`


dswij:
Not sure what's happening there at https://github.com/rust-lang/rust-clippy/runs/3717748423#step:5:12

xFrednet:
The log states:
```
.github/PULL_REQUEST_TEMPLATE.md: no issues found
CHANGELOG.md
  2407:70-2407:100  warning  Found reference to undefined definition  no-undefined-references  remark-lint

⚠ 1 warning
CODE_OF_CONDUCT.md: no issues found
```

The correlating text inside `CHANGELOG.md` looks like this:
```
2405 ## 0.0.81 — 2016-08-14
2406 * Rustup to *rustc 1.12.0-nightly (1deb02ea6 2016-08-12)*
2407 * New lints: [`eval_order_dependence`], [`mixed_case_hex_literals`], [`unseparated_literal_suffix`]
2408 * False positive fix in [`too_many_arguments`]
```

https://github.com/rust-lang/rust-clippy/blame/master/CHANGELOG.md#L2407

However, I'm not sure what causes the warning in this case.

dswij:
Ah, there is a `CHANGELOG.md` changes on 45a89b9. `unseparated_literal_suffix` is renamed to `literal_suffix` in the auto-generated links. But not changing it also causes an error: 

```
Run cargo dev update_lints --check
   Compiling clippy_dev v0.0.1 (/home/runner/work/rust-clippy/rust-clippy/clippy_dev)
    Finished dev [unoptimized + debuginfo] target(s) in 3.67s
     Running `target/debug/clippy_dev update_lints --check`
Not all lints defined properly. Please run `cargo dev update_lints` to make sure all lints are defined properly.
Error: Process completed with exit code 1.
```

https://github.com/rust-lang/rust-clippy/runs/3718070585#step:6:9

Not sure what's the best way to get around it. We can probably remove all old links to `unseparated_literal_suffix` in CHANGELOG.

dswij:
> Actually I think I would prefer having two lints over adding a config option for these "one or the other" scenarios. And this would avoid a lint rename.

This also crossed my mind. The only downside of this approach is that someone might get confused as to why two different and opposing lints co-exists. Maybe a documentation for the two lints can describe this, but I'm open to suggestions.

camsteffen:
Yes the docs should absolutely mention the other lint.

clippy_lints/src/misc_early/literal_suffix.rs
```diff
@@ -0,0 +1,38 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
```
dswij, line 1: Modified version of `clippy_lints/src/misc_early/unseparated_literal_suffix.rs`

dswij:
@flip1995 squashed :slightly_smiling_face:

flip1995:
@bors r+

Thanks!

leonardo-m:
Generally it's better to write 10_u8, it allows to tell apart better the value from the type.

PR #7727, opened 2021-09-27
Update changelog to 1.56

flip1995:
As expected a pretty short changelog, because of the missed sync last release cycle.

[Rendered](https://github.com/flip1995/rust-clippy/blob/changelog/CHANGELOG.md)

changelog: none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,60 +21,58 @@ Current beta, release 2021-10-21
 
 ### Enhancements
 
-* [`needless_continue`] now also lints in `loop { continue; }` case
+* [`needless_continue`]: Now also lints in `loop { continue; }` case
   [#7477](https://github.com/rust-lang/rust-clippy/pull/7477)
-* [`disallowed_type`] now also primitive types can be disallowed
+* [`disallowed_type`]: Now also primitive types can be disallowed
   [#7488](https://github.com/rust-lang/rust-clippy/pull/7488)
-* [`manual_swap`] now also lints on xor swaps
+* [`manual_swap`]: Now also lints on xor swaps
   [#7506](https://github.com/rust-lang/rust-clippy/pull/7506)
-* [`map_flatten`] now also lints on the `Result` type
+* [`map_flatten`]: Now also lints on the `Result` type
   [#7522](https://github.com/rust-lang/rust-clippy/pull/7522)
-* [`no_effect`] now also lints on inclusive ranges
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,11 +6,79 @@ document.
...
+  applicable [#7541](https://github.com/rust-lang/rust-clippy/pull/7541)
+
+### Documentation Improvements
+
+* Reworked Clippy's
+  [website](https://rust-lang.github.io/rust-clippy/master/index.html)
+  [#7279](https://github.com/rust-lang/rust-clippy/pull/7279):
+    * Added applicability information about lints
+    * Added a link to jump to the specific lint implementation
+    * Adapted some styling and improved loading time
```
xFrednet, line 67: This change is still listed under `Rust 1.55` the same goes for the change under *others*. I think we should keep them in this version and remove it below as the commits only landed in 1.56 :upside_down_face:

CHANGELOG.md
```diff
@@ -6,11 +6,79 @@ document.
...
+
+### New Lints
+
+* [`unwrap_or_else_default`]
+  [#7516](https://github.com/rust-lang/rust-clippy/pull/7516)
+
+### Enhancements
+
+* [`needless_continue`] now also lints in `loop { continue; }` case
+  [#7477](https://github.com/rust-lang/rust-clippy/pull/7477)
```
xFrednet, line 25: The previous versions used a colon after the lint name and then continued with a capital letter. I would suggest using the same format for consistency. The same goes for the other messages as well. What are your thoughts about it? :upside_down_face: 
```suggestion
* [`needless_continue`]: now also lints in `loop { continue; }` case
  [#7477](https://github.com/rust-lang/rust-clippy/pull/7477)
```

xFrednet (commented):
Hey, thank you for writing the change log! I just have two small comments that should be easy to fix :upside_down_face:

CHANGELOG.md
flip1995: Oh, how did that happen? :thinking: The commits are already in 1.55, just not in the sync, but the sync date is irrelevant for a change like this. But yeah, let's stay consistent with writing changelog by sync state.

CHANGELOG.md
flip1995: I seem to have forgotten my own formatting here :smile: Will change that.

xFrednet (approved):
Nice work, thank you! And awesome reference with "Lints all the way down" :wink:

PR #7733, opened 2021-09-28
Rustup

flip1995:
This needs a review this time. Especially https://github.com/rust-lang/rust-clippy/commit/521bf8f0fa18c7f130505f0a902ab0e65a76cec2 cc @camsteffen I think this is necessary now, because `itertools` is no longer a dependency of `clippy_utils` and therefore this path can't be found :thinking: 

( I forgot about the sync last week. I should get to document this process better, so other people can do it when I'm not around )

changelog: none


camsteffen:
> Especially [521bf8f](https://github.com/rust-lang/rust-clippy/commit/521bf8f0fa18c7f130505f0a902ab0e65a76cec2) cc @camsteffen I think this is necessary now, because `itertools` is no longer a dependency of `clippy_utils` and therefore this path can't be found 🤔

Indeed. I'd say either `allow` the path with a comment or have `IGNORED_CRATES =  ["itertools"]`.

flip1995:
I've gone for the added `allow`. This seemed cleaner to me.

PR #7741, opened 2021-09-30
Make if_then_panic handle situation of BinOpKind::And || BinOpKind::Or

### LINKED ISSUE #7731: `if-then-panic` suggestion is not equivalent for complex conditions
I tried this code:

```rust
#![warn(clippy::if_then_panic)]

fn main() {
    if false && false { 
        // comment
        panic!("ABC"); 
    };
}
```

I expected to see this happen: The suggestion would be equivalent but use `assert` instead.

Instead, this happened: 
```
 --> src/main.rs:5:5
  |
5 | /     if false && false { 
6 | |         // comment
7 | |         panic!("ABC"); 
8 | |     };
  | |_____^ help: try: `assert!(!false && false, "ABC");`
```
The suggestion doesn't correctly negate the condition and so changes the logic of the code. My original code will never panic. The suggestion will always panic.

### Meta

```
rustc 1.57.0-nightly (fdf65053e 2021-09-07)
binary: rustc
commit-hash: fdf65053e99e8966f9bd83b5a8491326cb33d638
commit-date: 2021-09-07
release: 1.57.0-nightly
LLVM version: 13.0.0
```
@rustbot label +L-suggestion

surechen:
fixes #7731

Make if_then_panic handle situation of cond.kind = ExprKind::DropTemps(ExprKind::Binary(BinOpKind::And || BinOpKind::Or, left, right), ..) =

changelog: [`if_then_panic`] Fix suggestion for more complex conditions

### DIFF
--- a/clippy_lints/src/if_then_panic.rs
+++ b/clippy_lints/src/if_then_panic.rs
@@ -3,7 +3,7 @@ use clippy_utils::higher::PanicExpn;
 use clippy_utils::is_expn_of;
 use clippy_utils::source::snippet_with_applicability;
 use rustc_errors::Applicability;
-use rustc_hir::{Block, Expr, ExprKind, StmtKind, UnOp};
+use rustc_hir::{BinOpKind, Block, Expr, ExprKind, StmtKind, UnOp};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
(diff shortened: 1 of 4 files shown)


surechen:
> fixes #7731
> 
> Make if_then_panic handle situation of cond.kind = ExprKind::DropTemps(ExprKind::Binary(BinOpKind::And || BinOpKind::Or, left, right), ..) =


Hello, @flip1995.  There is no automatic allocation of reviewer, I do not know how to deal with it, can you help me? Thank you

flip1995:
> There is no automatic allocation of reviewer

That's weird. I just self-assign it to me :)

flip1995:
r? @flip1995 

(FYI @surechen  this is how you can assign/request a specific reviewer)

flip1995:
I think your solution is more difficult than it has to be. The easiest fix for this would be to put the `expr` in `if <expr>` into parentheses and be done with it. You can do that more easily by just using `clippy_utils::Sugg`. There should be a `maybe_paren` function somewhere in there that will at parenthesis to the suggestion, if necessary.

teor2345:
> I think your solution is more difficult than it has to be. The easiest fix for this would be to put the `expr` in `if <expr>` into parentheses and be done with it. You can do that more easily by just using `clippy_utils::Sugg`. There should be a `maybe_paren` function somewhere in there that will at parenthesis to the suggestion, if necessary.

This would also handle operators like `==` and `!=`, and other complex expressions.

surechen:
> I think your solution is more difficult than it has to be. The easiest fix for this would be to put the `expr` in `if <expr>` into parentheses and be done with it. You can do that more easily by just using `clippy_utils::Sugg`. There should be a `maybe_paren` function somewhere in there that will at parenthesis to the suggestion, if necessary.

Hi, Done. clippy_utils::Sugg is very cool.  Thank you very much.

surechen:
> > I think your solution is more difficult than it has to be. The easiest fix for this would be to put the `expr` in `if <expr>` into parentheses and be done with it. You can do that more easily by just using `clippy_utils::Sugg`. There should be a `maybe_paren` function somewhere in there that will at parenthesis to the suggestion, if necessary.
> 
> This would also handle operators like `==` and `!=`, and other complex expressions.
> 
> Edit: there doesn't seem to be a simplification for `!(a != b)` -> `a == b`, but that's a totally separate issue.

Agree, the various possibilities here are very complicated. Using sugg is better than trying to cover various possible situations in this lint.

clippy_lints/src/if_then_panic.rs
```diff
@@ -74,12 +74,14 @@ impl LateLintPass<'_> for IfThenPanic {
...
-                    snippet_with_applicability(cx, not_expr.span, "..", &mut applicability).to_string()
+                let cond_sugg = if let ExprKind::DropTemps(e, ..) = cond.kind {
+                    if let Expr{kind: ExprKind::Unary(UnOp::Not, not_expr), ..} = e {
+                         sugg::Sugg::hir_with_applicability(cx, not_expr, "..", &mut applicability).maybe_par().to_string()
+                    } else {
+                       format!("!{}", sugg::Sugg::hir_with_applicability(cx, e, "..", &mut applicability).maybe_par().to_string())
+                    }
                 } else {
-                    format!("!{}", snippet_with_applicability(cx, cond.span, "..", &mut applicability))
+                   format!("!{}", sugg::Sugg::hir_with_applicability(cx, cond, "..", &mut applicability).maybe_par().to_string())
```
flip1995, line 84: `cond` and `e` should have the same `Span` , so the two `format!` calls are pretty much equivalent. I think you only have to change the `else` case and not the `ExprKind::Unary` case.

flip1995:
> Edit: there doesn't seem to be a simplification for `!(a != b)` -> `a == b`, but that's a totally separate issue.

There is the [`nonminimal_bool`](https://rust-lang.github.io/rust-clippy/master/index.html#nonminimal_bool) lint that should take care of this. But this lint is a bit restricted, since simplifying logical expressions is a NP-hard problem (or checking if two expressions are equivalent or something like that...).

clippy_lints/src/if_then_panic.rs
surechen: > `cond` and `e` should have the same `Span` , so the two `format!` calls are pretty much equivalent. I think you only have to change the `else` case and not the `ExprKind::Unary` case.

Hi, I thought the same at the beginning, but if I only modify the else block:
```
let cond_sugg = if let ExprKind::DropTemps(Expr{kind: ExprKind::Unary(UnOp::Not, not_expr), ..}, ..) = cond.kind {
                    sugg::Sugg::hir_with_applicability(cx, not_expr, "..", &mut applicability).maybe_par().to_string()
} else {
                   format!("!{}", sugg::Sugg::hir_with_applicability(cx, cond, "..", &mut applicability).maybe_par().to_string())
};
```
result will be like this：

```
error: only a `panic!` in `if`-then statement
  --> $DIR/if_then_panic.rs:45:5
   |
LL | /     if a.is_empty() && !b.is_empty() {
LL | |         panic!("panic3");
LL | |     }
   | |_____^ help: try: `assert!(!a.is_empty() && !b.is_empty(), "panic3");`
```
So I think "a.is_empty() && !b.is_empty()" is ExprKind::DropTemps{Expr{kind: ExprKind::Binary(BinOp::Or, left, right), ..}} here.

flip1995: Hm, that's weird. I also just tested it locally. In that case, let's merge it as it is.

flip1995:
@bors r+

Thanks!

Issue #7729, opened 2021-09-27
Redundant `to_string`?


smoelius:
I was surprised that there does not seem to be a lint to catch the following. Is it possible that it exists and I am somehow not triggering it?
```rust
use camino::Utf8PathBuf;

fn main() {
    let path = Utf8PathBuf::from("src/main.rs");
    println!("{}", path.to_string()); // to_string is unnecessary since Utf8PathBuf implements Display
    println!("{}", path);
}
```

camsteffen:
Related to #7729. I'd call this `to_string_in_format_args`.

akanalytics:
Is it true that 
a) println!("{:<20}", struct.to_string()); 
gives the same result as 
b) println!("{:<20}", struct); 

?

Doesnt it depend on how the formatter is written. 
Line a) uses the default  "{}" display to render a string then renders the string using the formatting rules :<20
Whereas b) uses the formatting rules of the struct's display, not the string's display trait impl.

camsteffen:
@akanalytics Feel free to file a bug if you found a false positive.

PR #7743, opened 2021-09-30
Add `format_in_format_args` and `to_string_in_format_args` lints

### LINKED ISSUE #7667: Lint suggestion: nested format!
### What it does

Warn on `format!` within the arguments of another macro that does formatting such as `format!` itself, `write!` or `println!`.

Suggests replacing the inner `format!` call with `format_args!` or inlining it entirely (see example).

### Categories (optional)

- Kind: style / perf

The recommended code is both shorter and avoids a temporary allocation.

### Drawbacks

There could be situations when the outer macro is `format_args!` where lifetimes don't work out with the `format!` changed to `format_args!` or inlined entirely. I'm not sure how hard it would be to detect these cases.

### Example

```rust
println!("error: {}", format!("something failed at {}", Location::caller()));
```

Could be written as:

```rust
println!("error: {}", format_args!("something failed at {}", Location::caller()));
```

or inlined entirely (when the argument is used exactly once like here):

```rust
println!("error: something failed at {}", Location::caller());
```

smoelius:
Fixes #7667 and #7729

I put these in `perf` since that was one of @jplatte's suggestions, and `redundant_clone` (which I consider to be similar) lives there as well.

However, I am open to changing the category or anything else.

r? @camsteffen

changelog: Add `format_in_format_args` and `to_string_in_format_args` lints


smoelius:
Sorry for the late addition. I added two more test cases.

clippy_utils/src/format.rs
```diff
@@ -0,0 +1,49 @@
...
+        if let Some(width_field) = fields.iter().find(|f| f.ident.name == sym::width);
+        if let ExprKind::Path(ref width_qpath) = width_field.expr.kind;
+        if last_path_segment(width_qpath).ident.name == sym::Implied;
+        then {
+            return true;
+        }
+    }
+
+    false
+}
```
camsteffen, line 49: Since this is moving to utils, there are some things to fix.

1. Make this a method of `FormatArgsExpn` and give it a "positive" name like `format_args.has_formatting()`
1. Update the docs to say something like "returns true if any formating parameters are used like `{:+2}` instead of just `{}`"
1. Work for any number of args - the `ExprKind::Array` at the top currently only works for one arg.
1. Check for `flags: 0` in addition to checking `precision` and `width`
    * For perf - avoid doing `fields.iter().find()` for every field

clippy_lints/src/format_args.rs
```diff
@@ -0,0 +1,262 @@
...
+{
+    if_chain! {
+        if let Some(format_args) = FormatArgsExpn::parse(expr);
+        let call_site = expr.span.ctxt().outer_expn_data().call_site;
+        if call_site.from_expansion();
+        let expn_data = call_site.ctxt().outer_expn_data();
+        if let ExpnKind::Macro(_, name) = expn_data.kind;
+        if format_args.fmt_expr.map_or(true, check_unformatted);
+        then {
+            assert_eq!(format_args.args.len(), format_args.value_args.len());
```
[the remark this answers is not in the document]
camsteffen: We may need a new abstraction to tie everything together, so we can do

```rust
for arg in format_args.args() {
    if !arg.has_formatting() && arg.is_display() {
```

where `arg` is a new thing like

```rust
struct FormatArgsArg {
    value: &Expr, // this is an element of `value_args` according to `position`
    fmt: &Expr, // this is an element of `args`
}
```

clippy_utils/src/format.rs
smoelius: > 4\. Check for `flags: 0` in addition to checking `precision` and `width`

`useless_format` doesn't seem to like it when we check for `flags: 0`. So, I'm proposing that `has_formatting` be told what formatting types to check for, e.g.:
```rust
bitflags! {
    pub struct Formatting: u32 {
        const FILL = 0b00000001;
        const ALIGN = 0b00000010;
        const FLAGS = 0b00000100;
        const PRECISION = 0b00001000;
        const WIDTH = 0b00010000;
    }
}
```
```rust
                if !format_args.has_formatting(Formatting::PRECISION | Formatting::WIDTH);
```
Is it ok to add `bitflags` as a dependency of `clippy_utils`?

smoelius:
I pushed the changes using `bitflags`.

`bitflags 1.2` is already in `rust`'s lockfile, so hopefully this won't be a problem: https://github.com/rust-lang/rust/blob/a8387aef8c378a771686878062e544af4d5e2245/Cargo.lock#L125

Please let me know what other changes are needed.

clippy_utils/src/format.rs
camsteffen: > `useless_format` doesn't seem to like it when we check for `flags: 0`.

It looks like `useless_format` does not consider `flags` because it happens that none of the "flags" have any effect when the value is a string, and `useless_format` only triggers on strings. What if we just have a `has_string_formatting` which returns true if there is formatting that has an effect for string values? The bitflags seems more complex than necessary, and their general usefulness is limited since they don't let you see actual values like width and alignment.

smoelius: I think I implemented what you described, but please let me know if I misunderstood.

camsteffen (changes requested):

clippy_utils/src/higher.rs
```diff
@@ -569,6 +569,154 @@ impl FormatArgsExpn<'tcx> {
...
+        match self {
+            FormatArgsIter::Formatted {
+                value_args,
+                args,
+                fmt_iter,
+            } => fmt_iter.next().map(|fmt| {
+                if_chain! {
+                    // struct `core::fmt::rt::v1::Argument`
+                    if let ExprKind::Struct(_, [position_field, _], _) = fmt.kind;
+                    let _ = assert_eq!(position_field.ident.name, sym::position);
```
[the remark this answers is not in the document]
smoelius: Why prefer `find()` in this case?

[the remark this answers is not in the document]
camsteffen: So that we don't rely on the field order and avoid a panic case.

camsteffen:
> @camsteffen The paths I added to path.rs appear to be the only ones that refer to macros. I am guessing this is why `INVALID_PATHS` is failing. Assuming this is right, do you have any suggestions on how to fix it?

@smoelius I'd probably have to debug the invalid_paths lint to figure it out. But the paths should be migrated to diagnostic items anyways so it's kinda not worth the effort. So it's okay to use paths for now and allow the lint on those items with a todo. Or, if you feel inclined, you can add the diagnostic items in rustc and wait for them to be synced in the following Rustup PR.

Issue #7238, opened 2021-05-18
New Lint: Unsafe block in safe functions without `SAFETY` comment 


fee1-dead:
### What it does

Detects unsafe blocks used in safe functions missing a comment explaining/justifying the use of `unsafe`.

### Categories (optional)

- Kind: `clippy::pedantic`

*What is the advantage of the recommended code over the original code*

- Explains why the use of `unsafe` in the function is okay
- Makes it easier to maintain

### Drawbacks

None.

### Example

```rust
fn get(self, slice: &[T]) -> Option<T> {
  if self < slice.len() { unsafe { Some(&*self.get_unchecked(slice)) } } else { None }
}
```

Could be written as:

```rust
fn get(self, slice: &[T]) -> Option<T> {
  // SAFETY: `self` is checked to be in bounds.
  if self < slice.len() { unsafe { Some(&*self.get_unchecked(slice)) } } else { None }
}
```

giraffate:
From [API guidelines](https://rust-lang.github.io/api-guidelines/documentation.html), unsafe functions should be documented with a "Safety" section, so why do you want it with unsafe blocks?

FYI, Clippy has the [`missing_safety_doc`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc) lint.

fee1-dead:
There are some requirements for a call to an unsafe function to be safe. A comment with unsafe blocks can explain how the safety contract is respected. As shown in the example, it could be easier to maintain if the developer knows that unsafe block has been checked. This new lint can help developers to make sure that all usage of unsafe is documented.

For checking comments, I don't know if there is a solution. The Rust repository does document their usage of unsafe so I thought "why not a lint?"

fee1-dead:
Since #7748 was merged, I am closing this.

PR #7748, opened 2021-10-02
Add undocumented_unsafe_block_safety lint

### LINKED ISSUE #7464: `missing_unsafe_doc`, explain why an `unsafe` block is safe/sound

Serial-ATA:
changelog: Added a new lint [`undocumented_unsafe_block_safety`]

Fixes #7464, #7238 (?)


xFrednet:
Welcome to the project, it's nice to see that you got a working implementation :upside_down_face:.

@flip1995 would you maybe like to review this, as you've also reviewed #7557? Otherwise, I'll have a look at it during the next week :upside_down_face:

flip1995:
Yes, sure. I have time this afternoon.

r? @flip1995

clippy_lints/src/undocumented_unsafe_block_safety.rs
```diff
@@ -0,0 +1,198 @@
...
+enum CommentType {
+    Safety,
+    Unrelated,
+    None,
+}
+
+impl LateLintPass<'_> for UndocumentedUnsafeBlockSafety {
+    fn check_block(&mut self, cx: &LateContext<'_>, block: &'_ Block<'_>) {
+        if self.local {
+            self.local = false;
```
flip1995, line 63: That will produce bugs:

```rs
let _ = {
    let _ = {
        // ...
    }
    // self.local is false, even though we're still in a local.
}
```

clippy_lints/src/undocumented_unsafe_block_safety.rs
```diff
@@ -0,0 +1,198 @@
...
+            }
+        }
+    }
+
+    fn check_local(&mut self, cx: &LateContext<'_>, local: &'_ Local<'_>) {
+        if_chain! {
+            if !is_lint_allowed(cx, UNDOCUMENTED_UNSAFE_BLOCK_SAFETY, local.hir_id);
+            if !in_external_macro(cx.tcx.sess, local.span);
+            if let Some(init) = local.init;
+            if let ExprKind::Block(block, _) = init.kind;
```
flip1995, line 83: This won't work for 

```rs
let _ = (42, unsafe { *ptr::null_ptr() });
```

Please also add a test for this.

clippy_lints/src/undocumented_unsafe_block_safety.rs
```diff
@@ -0,0 +1,198 @@
...
+    let between_span = enclosing_scope_span.until(block_span);
+
+    let file_name = source_map.span_to_filename(between_span);
+    let source_file = source_map.get_source_file(&file_name)?;
+
+    let lex_start = (between_span.lo().0 + 1) as usize;
+    let mut src_str = source_file.src.as_ref()?[lex_start..between_span.hi().0 as usize].to_string();
+
+    // Remove all whitespace, but retain newlines to verify the block immediately follows the comment
+    src_str.retain(|c| c == '\n' || !c.is_whitespace());
```
flip1995, line 127: That doesn't matter IMO.

```rs
// SAFETY: ...

unsafe { ... }
```

is totally fine. I think the formatting is weird, but if people want to do that, more power to them.

clippy_lints/src/undocumented_unsafe_block_safety.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let mut src_str_split = src_str.lines().rev();
+
+    let mut found_safety_comment = false;
+
+    while let Some(line) =  src_str_split.next() {
```
flip1995, line 133: I think using the Rust lexer here is the way to go, like #7557 did. You already do the span manipulation, before re-lexing the code (by hand), so you already addressed my comment https://github.com/rust-lang/rust-clippy/pull/7557/files#r687488132

flip1995 (changes requested):
If I had to guess, I assume, that you didn't look at #7557 before implementing this?

I think your approach and the approach taken in #7557 combined would be ideal. I left comments about that below. IMO the work done in #7557 is really valuable. The PR was only closed, because it was inactive.

clippy_lints/src/undocumented_unsafe_block_safety.rs
Serial-ATA: I just tested this out:
```rust
    let _ = {
        let _ = {
            let _ = unsafe {};
        };
    };
```
```
error: unsafe block missing a safety comment
  --> $DIR/undocumented_unsafe_blocks.rs:8:13
   |
LL |             let _ = unsafe {};
   |             ^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::undocumented-unsafe-blocks` implied by `-D warnings`
```
It still only ends up checking the local with the unsafe block.

clippy_lints/src/undocumented_unsafe_block_safety.rs
Serial-ATA: Yeah, I just thought it was a little weird. Probably should've asked before deviating from the original PR.

clippy_lints/src/undocumented_unsafe_block_safety.rs
Serial-ATA: Do you mean a UI test, or should it be handled?

clippy_lints/src/undocumented_unsafe_block_safety.rs
flip1995: The problem is not the inner unsafe block, but if you put an unsafe block after leaving one of the inner blocks. We have this situation multiple times. For this we usually count how many layers deep we are with `saturating_add(1)` and `saturating_sub(1)`, instead of setting a boolean.

clippy_lints/src/undocumented_unsafe_block_safety.rs
flip1995: Putting a `SAFETY` comment above this line should also be acceptable.

Serial-ATA:
I corrected all the issues, but I see those Manishearth commits it says I committed. Will that be an issue? I don't know too much about git.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let mut pos = 0;
+    let mut found_safety_comment = false;
+
+    for token in rustc_lexer::tokenize(&src_str) {
+        match token.kind {
+            TokenKind::LineComment { doc_style: None }
+            | TokenKind::BlockComment { doc_style: None, terminated: true } => {
+                if src_str[pos + 2.. pos + token.len].to_ascii_uppercase().contains("SAFETY:") {
+                    found_safety_comment = true;
```
ShadowJonathan, line 131: Please also check if the safety comment is just `...`, i.e. if the person just jammed in the suggestion without actually typing out any safety comment, this'd prevent laziness and false positives from shoddy lint wack-a-mole-ing.

ShadowJonathan (changes requested):
Thanks! Just missing one feature from the original suggestion in #7464;

> The lint would also raise a warning if that comment is copied verbatim.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,221 @@
...
+    let src_str = source_file.src.as_ref()?[lex_start..between_span.hi().0 as usize].to_string();
+
+    let mut pos = 0;
+    let mut comment = false;
+
+    for token in rustc_lexer::tokenize(&src_str) {
+        match token.kind {
+            TokenKind::LineComment { doc_style: None }
+            | TokenKind::BlockComment { doc_style: None, terminated: true }  => {
+                let comment_str = src_str[pos + 2..pos + token.len].to_ascii_uppercase();
```
flip1995, line 194: What happens, if the comment is not ASCII? Please add a test, where the comment contains non-ascii characters.

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,205 @@
...
+fn nested_local() {
+    let _ = {
+        let _ = {
+            // Safety:
+            let _ = unsafe {};
+        };
+    };
+}
+
+fn deep_nest() {
```
flip1995, line 14: That should test this case enough :smile: Thanks!

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,221 @@
...
+            },
+            TokenKind::Whitespace => {},
+            _ => {
+                if comment {
+                    let comment_line_num = source_file.lookup_file_pos_with_col_display(BytePos((lex_start + pos).try_into().unwrap())).0;
+                    let block_line_num = tcx.sess.source_map().lookup_char_pos(block_span.lo()).line;
+
+                    if block_line_num == comment_line_num + 1 || block_line_num == comment_line_num {
+                        return None;
+                    }
```
flip1995, line 210: What does this do? I don't think this is obvious, and I don't want to think about it. Can you comment it and/or introduce some more variables with telling names?

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,205 @@
...
+}
+
+fn local_no_comment() {
+    let _ = unsafe {};
+}
+
+fn local_commented_block() {
+    let _ =
+        // Safety:
+        unsafe {};
```
flip1995, line 184: IMO this shouldn't be linted? Is there a reason why this is linted? If the answer is "it would be too hard to prevent", that's fine, but then it should be documented in `Known Problems`

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,205 @@
...
+
+fn no_comment_unary() {
+    let _ = *unsafe { &42 };
+}
+
+#[allow(clippy::match_single_binding)]
+fn no_comment_match() {
+    let _ = match unsafe {} {
+        _ => {},
+    };
```
flip1995, line 166: Is

```rs
    // Safety: abc
    let _ = match unsafe {} {
        _ => {},
    };
```

getting linted?

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,205 @@
...
+    unsafe {}
+}
+
+fn local_line_comment() {
+    // Safety:
+    let _ = unsafe {};
+}
+
+fn local_block_comment() {
+    /* Safety: */
```
flip1995, line 135: Linting on `Safety: ...` but not on `Safety: ` (with no explanation) is a bit weird. But getting this right might be difficult for example for the lorem ipsum test case. So if you have an idea how to easily handle this, you can make this improvement, otherwise the current behavior is totally fine.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    for token in rustc_lexer::tokenize(&src_str) {
+        match token.kind {
+            TokenKind::LineComment { doc_style: None }
+            | TokenKind::BlockComment { doc_style: None, terminated: true }  => {
+                let comment_str = src_str[pos + 2..pos + token.len].to_ascii_uppercase();
+
+                if comment_str.contains("SAFETY: ...") {
+                    return Some(ErrorKind::Placeholder);
+                } else if comment_str.contains("SAFETY:") {
```
flip1995, line 198: Should we also allow `SAFETY` without a colon `:`? That would also allow other formats, like

```rs
// ## SAFETY
// bla blub
```

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,221 @@
...
+            },
+        ExprKind::Box(Expr {
+                          kind: ExprKind::Block(block, _),
+                          ..
+                      }) |
+        ExprKind::Unary(_, Expr {
+            kind: ExprKind::Block(block, _),
+            ..
+        }) |  ExprKind::Match(Expr {
+                                                            kind: ExprKind::Block(block, _),
```
flip1995, line 121: Is GitHub just displaying this weirdly or is the formatting totally broken?

flip1995 (commented):
Impl LGTM overall now. A few minor improvements could be made.

clippy_lints/src/undocumented_unsafe_blocks.rs
Serial-ATA: Yeah, I've been getting some weird results with `cargo dev fmt`.

tests/ui/undocumented_unsafe_blocks.rs
Serial-ATA: All of the `no_comment_*` tests would pass with a safety comment. I'll also make those tests.

tests/ui/undocumented_unsafe_blocks.rs
Serial-ATA: Yeah, I think I'll just go back to the old behavior. I don't think I could handle all possible cases.

clippy_lints/src/undocumented_unsafe_blocks.rs
Serial-ATA: That would allow something like

```rust
// We don't care about safety
unsafe { ... }
```

to pass.

tests/ui/undocumented_unsafe_blocks.rs
Serial-ATA: This wouldn't be hard to do, but what about in the case of:

```rust
// Safety:
let _ = (42,
         unsafe { … },
         unsafe { … },
         // Safety:
         unsafe { … });
```

Should the overall comment on the local cover the 2 undocumented blocks?

flip1995: Hm good question. I would say so. That leaves the possibility open, to only document the unsafe behavior once, if it is (almost) duplicated in both blocks.

clippy_lints/src/undocumented_unsafe_blocks.rs
flip1995: Good point. I think the current behavior is fine then :+1:

clippy_lints/src/undocumented_unsafe_blocks.rs
flip1995: I wonder if this is a bug in rustfmt :thinking: Let's deal with this right before merging this PR.

Serial-ATA: I just formatted it in playground and `cargo dev fmt --check` still passes. Is it just not doing anything then?

tests/ui/undocumented_unsafe_blocks.rs
Serial-ATA: @ShadowJonathan ^

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,273 @@
...
+        self.local_level = self.local_level.saturating_sub(1);
+
+        if self.local_level == 0 {
+            self.local_checked = false;
+            self.local_span = None;
+        }
+    }
+}
+
+fn find_block_in_expr(expr_kind: &ExprKind<'hir>) -> Option<(&'tcx Block<'hir>, u32)> {
```
flip1995, line 109: This should be way easier to implement with a [visitor](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/intravisit/trait.Visitor.html)

This would look something like this:

```rs
fn check_expr(...) {
    if let ExprKind::Block(block) = expr.kind { 
        self.block = block;
        return;
    }
    walk_expr(expr);
}
```

If you only want to go one level deep and not walk down the whole expr, you can add a counter and early return.

flip1995 (commented):
This PR should be almost done. I have some suggestions on how to simplify the code a bit.

The tests look awesome! The only test that is missing is a macro test. Two important tests here would be

- `unsafe` block in the macro **call**: `m!(unsafe{ ... }`
- `unsafe` block in the macro **definition**

clippy_lints/src/undocumented_unsafe_blocks.rs
flip1995: I'm not sure. I'll test this myself before merging. Not sure what's going on.

tests/ui/undocumented_unsafe_blocks.rs
ShadowJonathan: Hmm, I still don't like that, but I guess this can be re-introduced somewhere down the line

ShadowJonathan:
Just for posterity; the behaviour that makes sure that "..." is also checked and errored on is removed, I think that's fine for now to get this PR through, but I'd still like to see it somewhere in the future

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,273 @@
...
+
+impl LateLintPass<'_> for UndocumentedUnsafeBlocks {
+    fn check_block(&mut self, cx: &LateContext<'_>, block: &'_ Block<'_>) {
+        if_chain! {
+            if !self.local_checked;
+            if !is_lint_allowed(cx, UNDOCUMENTED_UNSAFE_BLOCKS, block.hir_id);
+            if !in_external_macro(cx.tcx.sess, block.span);
+            if let BlockCheckMode::UnsafeBlock(UnsafeSource::UserProvided) = block.rules;
+            if let Some(enclosing_scope_hir_id) = cx.tcx.hir().get_enclosing_scope(block.hir_id);
+            if block_has_safety_comment(cx.tcx, enclosing_scope_hir_id, block.span) == Some(false);
```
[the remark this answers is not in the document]
Serial-ATA: I just changed this around since I saw #7762 get merged.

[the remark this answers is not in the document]
flip1995: Ah, hmmm. I don't like this lint. I think pattern matching on an option is pretty common in Rust and therefore easy to read. But that's not something to address in this issue. So nothing to do here.

tests/ui/undocumented_unsafe_blocks.stderr
```diff
@@ -0,0 +1,171 @@
...
+LL |             unsafe {}
+   |             ^^^^^^^^^
+...
+LL |     t!();
+   |     ----- in this macro invocation
+   |
+   = note: this error originates in the macro `t` (in Nightly builds, run with -Z macro-backtrace for more info)
+help: consider adding a safety comment
+   |
+LL ~             // Safety: ...
```
Serial-ATA, line 126: It supports macro calls and definitions now, but there's an issue with the indentation I don't know how to deal with.

clippy_lints/src/undocumented_unsafe_blocks.rs
Serial-ATA: The check just failed on some missing semicolons, so its doing *something* at least.

tests/ui/undocumented_unsafe_blocks.stderr
flip1995: That is because the `snippet` function uses the code of the expanded macro. I don't think we can suggest to add it inside the macro definition, since we don't have the span information on where the macro is defined. 

This suggestion may get confusing for more complex macro calls. So I would suggest to not produce a suggestion for `in_macro(unsafe_block_span)`, but just a help message saying "consider adding a safety comment in the macro definition".

flip1995:
One more thing that came to mind: What if the unsafe block has multi-line code in it? How does the suggestion look in that case? I expect that the full unsafe block is shown. We may want to avoid this, by using the [`span_until_char`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_span/source_map/struct.SourceMap.html#method.span_until_char) function ( or similar functions)

flip1995:
Looks great now! Could you address the multi-line unsafe block case? After that this should be good to go. :rocket:

flip1995:
> Did I not take care of that? The `no_comment_match` and `internal_comment` tests only show the first line now.

Oh I missed that. In that case all good :+1: 

(`no_comment_match` doesn't test that, because the `unsafe` block is still empty and all on one line in this test case, but the other test case does)

flip1995:
@bors r+

Thanks for sticking around for all the review iterations. Great work!

PR #7755, opened 2021-10-03
exclude enum from derivable impls

### LINKED ISSUE #7753: derivable_impls wants derive(Default) on enum
Lint name: [`derivable_impls`](https://rust-lang.github.io/rust-clippy/master/index.html#derivable_impls)

I tried [this code:](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=4a7402ede26296690167908770a84331)

```rust
pub enum IntOrString {
    Int(i32),
    String(String),
}

impl Default for IntOrString {
    fn default() -> Self {
        IntOrString::Int(0)
    }
}
```

Clippy says:

```
warning: this `impl` can be derived
  --> src/lib.rs:6:1
   |
6  | / impl Default for IntOrString {
7  | |     fn default() -> Self {
8  | |         IntOrString::Int(0)
9  | |     }
10 | | }
   | |_^
   |
   = note: `#[warn(clippy::derivable_impls)]` on by default
   = help: try annotating `IntOrString` with `#[derive(Default)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#derivable_impls
```

This is incorrect because rustc cannot derive `Default` on enums (until [`feature(derive_enum_default)`](https://github.com/rust-lang/rust/issues/87517) is stabilized.)

### Meta

```
rustc 1.57.0-nightly (2b862bed9 2021-09-23)
binary: rustc
commit-hash: 2b862bed9889808b69629fd7246317189b9517a5
commit-date: 2021-09-23
release: 1.57.0-nightly
LLVM version: 13.0.0
```

The playground repro uses `2021-10-02 f03eb6bef8ced8a24385`

---

cc @HKalbasi

HKalbasi:
fix #7753 

changelog: Exclude enum from ``[`derivable_impls`]``

### DIFF
--- a/clippy_lints/src/derivable_impls.rs
+++ b/clippy_lints/src/derivable_impls.rs
@@ -83,6 +83,7 @@ impl<'tcx> LateLintPass<'tcx> for DerivableImpls {
             if !attrs.iter().any(|attr| attr.doc_str().is_some());
             if let child_attrs = cx.tcx.hir().attrs(impl_item_hir);
             if !child_attrs.iter().any(|attr| attr.doc_str().is_some());
+            if !adt_def.is_enum();
             then {
                 if let TyKind::Path(QPath::Resolved(_, p)) = self_ty.kind {
                     if let Some(PathSegment { args: Some(a), .. }) = p.segments.last() {
--- a/clippy_lints/src/derivable_impls.rs
+++ b/clippy_lints/src/derivable_impls.rs
@@ -83,7 +83,7 @@ impl<'tcx> LateLintPass<'tcx> for DerivableImpls {
             if !attrs.iter().any(|attr| attr.doc_str().is_some());
             if let child_attrs = cx.tcx.hir().attrs(impl_item_hir);
             if !child_attrs.iter().any(|attr| attr.doc_str().is_some());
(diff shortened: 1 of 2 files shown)


clippy_lints/src/derivable_impls.rs
```diff
@@ -83,6 +83,7 @@ impl<'tcx> LateLintPass<'tcx> for DerivableImpls {
             if !attrs.iter().any(|attr| attr.doc_str().is_some());
             if let child_attrs = cx.tcx.hir().attrs(impl_item_hir);
             if !child_attrs.iter().any(|attr| attr.doc_str().is_some());
+            if !adt_def.is_enum();
```
xFrednet, line 86: I think this should check for `if adt_def.is_struct();` instead, to make sure that we don't get any FPs on unions, as they can also not derive `Default`. Otherwise, this looks good to me :upside_down_face:

HKalbasi:
If the problem was Enter after changlog, I deleted it.

xFrednet:
Yes that it most likely the problem they have to be on the same line. I've updated the description :upside_down_face: 

@bors retry

PR #7775, opened 2021-10-06
Detect underscored variables with no side effects

### LINKED ISSUE #7545: Detect useless underscored variables with no side effects
### What it does

It detects variables like `let _unused = 1;` and suggests deleting that variable because it serves 0% purpose and has no side effects whatsoever. This is different from `let _unused_result = function();` which likely has a side effect (unless the function is empty/the function contents have no side effect, maybe that should be detected too? Probably not but I'm not sure).

See also: https://github.com/rust-lang/rust-clippy/issues/7527#issuecomment-894302979

### Categories (optional)

- Kind: `clippy::suspicious`.

It removes unused variables that serve no purpose.

One could argue that there are use cases for this like tests (for `rustc` I suppose) or examples merely showing off the ability to declare such variables, but one can of course always turn this lint off and in the most common case the user will actually want to remove this underscored variable with 0% side effects, I believe.

### Drawbacks

- The requirement to mute the lint, in very rare cases where the user wants this.

### Example

```rust
fn main() {
    let _i_serve_no_purpose = 1;
}
```

Could be written as:

```rust
fn main() { }
```

If the variable is used afterwards then of course there should be no lint regardless of whether there is a side effect in the variable declaration itself or not.

F3real:
Fixes #7545

changelog: Lint on underscored variables with no side effects in [`no_effect`]


flip1995:
This lint should be allowed in all test cases, that don't explicitly test for this lint. Otherwise all the stderr files are just messy.

Please also add a test case, where the assigned value implements `Drop`, which shouldn't lint, e.g. `let _x = vec![1];`. This allocated memory and keeps it allocated until the end of the scope, so definitely has an effect.

I'm also not so sure if this should be an extension of the `no_effect` lint, since it seems pretty noisy. E.g. what happens if the `_x` variable is used afterwards? A leading `_` is no guarantee that the variable is unused. A possible real world example:

```rs
let _x = 42; // only required if *one of* f1 or f2 is enabled, 
             // but has to be declared with `_` to avoid warning if none of those features are enabled
let mut y = 0;

// If feature 1 is enabled call foo with `_x - 1`
#[cfg(feature = "f1")]
{ y += foo(_x - 1); }

// If feature 2 is enabled call foo with `_x + 1`
#[cfg(feature = "f2")]
{ y += foo(_x + 1); }

y += foo(0);
```

PR #7776, opened 2021-10-06
Fix typos

tsoutsman:
I'm not sure whether I should add links to `bool`, `char`, and `str`. `slice` could also be linked to.

changelog: none

### DIFF
--- a/clippy_utils/src/ty.rs
+++ b/clippy_utils/src/ty.rs
@@ -224,15 +224,15 @@ fn is_normalizable_helper<'tcx>(
     result
 }
 
-/// Returns true iff the given type is a non aggregate primitive (a bool or char, any integer or
-/// floating-point number type). For checking aggregation of primitive types (e.g. tuples and slices
-/// of primitive type) see `is_recursively_primitive_type`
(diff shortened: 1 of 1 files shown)


clippy_utils/src/ty.rs
```diff
@@ -224,15 +224,15 @@ fn is_normalizable_helper<'tcx>(
     result
 }
 
-/// Returns true iff the given type is a non aggregate primitive (a bool or char, any integer or
-/// floating-point number type). For checking aggregation of primitive types (e.g. tuples and slices
-/// of primitive type) see `is_recursively_primitive_type`
+/// Returns `true` if the given type is a non aggregate primitive (a `bool` or `char`, any
```
flip1995, line 227: `iff` is not a typo, but stands for [`if and only if`](https://en.wikipedia.org/wiki/If_and_only_if). But I guess for those two functions, it doesn't really matter.

```
type is primitive <=> return true
type is primitive  => return true
```

Is pretty much the same in this context.

flip1995:
@bors r+

Thanks!

PR #7780, opened 2021-10-07
Revert `update_lints` module list generating code

ghost:
This commit reverts the module list generation code to what it was
before the change to `include!` it and generates better output.

changelog: none


llogiq:
@bors r+

Thank you!

PR #7791, opened 2021-10-08
Allow giving reasons for `disallowed_types`

JamesHinshelwood:
Similar to #7609 but for the `disallowed_type` lint. The permitted form of configuration is the same as for `disallowed_methods`.

changelog: Allow giving reasons for [`disallowed_type`]


giraffate:
@bors r+

It looks good, thanks!

Issue #6312, opened 2020-11-09
field_reassign_with_default could be false positive


Fullstop000:
I tried this code:

```rust
#[derive(Clone, Default)]
pub struct FileNode {
    name: String,
    delay_data_sync: Arc<AtomicBool>,
}

impl FileNode {
    fn new(name: &str) -> Self {
        let mut f = FileNode::default();
        f.name = name.to_owned();
        f
    }
}
```

I expected to see this happen: 
Clippy reports ok as the `delay_data_sync` does not implement `Copy`

Instead, this happened: 
Clippy complains about `field assignment outside of initializer for an instance created with Default::default()`. If refactor the codes as the suggestion clippy provides, `rustc` just throws the error:
```
cannot move out of type `FileNode`, which implements the `Drop` trait
move occurs because value has type `Arc<AtomicBool>`, which does not implement the `Copy` trait rustc(E0509)
```

### Meta

- `cargo clippy -V`: clippy 0.0.212 (1773f60 2020-11-08)
- `rustc -Vv`:
  ```
rustc 1.49.0-nightly (1773f60ea 2020-11-08)
binary: rustc
commit-hash: 1773f60ea5d42e86b8fdf78d2fc5221ead222bc1
commit-date: 2020-11-08
host: x86_64-unknown-linux-gnu
release: 1.49.0-nightly
  ```

flip1995:
Can you provide an example on playground? I don't get this error after applying the lint: [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=fd91b8b2b9816020ea07f4751cefb99b)

Fullstop000:
@flip1995 Sorry I forgot to post the `Drop` part
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=3f6e5e3e6ca7da0317fb753cf0d2a1bd

ebroto:
Indeed, the `Drop` part is important. We hit [E0509](https://doc.rust-lang.org/nightly/error-index.html#E0509) because [functional record update](https://doc.rust-lang.org/reference/expressions/struct-expr.html#functional-update-syntax) `moves or copies the remaining fields`. Since `Arc<AtomicBool>` does not implement `Copy` we are moving out of the struct that implements `Drop`.

To fix this, we should avoid linting if the type implements `Drop` and has any field which is not `Copy`.

PR #7794, opened 2021-10-09
Fix false positive when `Drop` and `Copy` involved in `field_reassign_with_default` lint

ThibsG:
Fix FP in `field_reassign_with_default` lint when type implements `Drop` but not all fields are `Copy`.

fixes: #6312

changelog: [`field_reassign_with_default`] Fix FP when `Drop` and `Copy` are involved


llogiq:
Looks good! I always cheer when we remove false positives. 🥳 Thank you!

@bors r+

PR #7800, opened 2021-10-10
Refactor `clippy::match_ref_pats` to check for multiple reference patterns

### LINKED ISSUE #7740: clippy::match_ref_pats probably shouldn't fire for if let
Lint name:

I tried this code:

```rust
if let &KeyboardInput {
    keycode: Some(keycode),
    state,
    ..
} = event
{
    // ...
}
```

I expected to see this happen: warning free

Instead, this happened: `clippy::match_ref_pats` matched

The point of `clippy::match_ref_pats` is that it's less notation to put a single deref on the matched value than a ref on every pattern, but with `if let` this doesn't apply, as there's only one pattern to begin with.

Either the description/help message should change when only one pattern is present, or `clippy::match_ref_pats` should only be emitted when more than one pattern are present.

### Meta

```
rustc 1.56.0-nightly (ad02dc46b 2021-08-26)
binary: rustc
commit-hash: ad02dc46badee510bd3a2c093edf80fcaade91b1
commit-date: 2021-08-26
host: x86_64-pc-windows-msvc
release: 1.56.0-nightly
LLVM version: 13.0.0
```

1nF0rmed:
fixes #7740 

When there is only one pattern, to begin with, i.e. a single deref(`&`), then in such cases we suppress `clippy::match_ref_pats`. 
This is done by checking the count of the reference pattern and emitting `clippy::match_ref_pats` when more than one pattern is present.

Removed certain errors in the `stderr` tests as they would not be triggered.  

changelog: Refactor `clippy::match_ref_pats` to check for multiple reference patterns

### DIFF
--- a/tests/ui/match_ref_pats.rs
+++ b/tests/ui/match_ref_pats.rs
@@ -72,4 +72,46 @@ mod ice_3719 {
     }
 }
 
+mod issue_7740 {
+    macro_rules! foobar_variant(
+        ($idx:expr) => (FooBar::get($idx).unwrap())
+    );
+
+    enum FooBar {
+        Foo,
+        Bar,
+        FooBar,
+        BarFoo,
+    }
+
+    impl FooBar {
+        fn get(idx: u8) -> Option<&'static Self> {
+            match idx {
+                0 => Some(&FooBar::Foo),
+                1 => Some(&FooBar::Bar),
+                2 => Some(&FooBar::FooBar),
+                3 => Some(&FooBar::BarFoo),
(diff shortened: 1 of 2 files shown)


llogiq:
Cool stuff! The only downside I see is that some test cases are now effectively disabled. Would you be OK with changing them so they apply again and adding one or two new tests that show single ref patterns are not linted?

llogiq:
I'm only talking about the lint's own tests. To make the error re-occur, you may need an enum with more than two cases.

PR #7806, opened 2021-10-12
Add match_str_case_mismatch lint

### LINKED ISSUE #7440: Flag non-lowercase string literals in match expression on .to_lowercase()
### What it does

Detect match statements of form `str_variable.to_lowercase()` with non-lowercase string literals as match arms. Should also apply to `to_ascii_lowercase`, `to_uppercase`, and `to_ascii_uppercase`.

### Categories (optional)

- Kind: `clippy::correctness`

The flagged match arm in the original code wouldn't ever be triggered, which is likely a bug in the program.

### Drawbacks

None.

### Example

```rust
match &*kind.to_ascii_lowercase() {
    "foo" => {},
    "Bar" => {},
    _ => {},
}
```

Should be written as:

```rust
match &*kind.to_ascii_lowercase() {
    "foo" => {},
    "bar" => {},
    _ => {},
}
```

Serial-ATA:
changelog: Added a new lint [`match_str_case_mismatch`]

Fixes #7440

### DIFF
--- a/clippy_lints/src/match_str_case_mismatch.rs
+++ b/clippy_lints/src/match_str_case_mismatch.rs
@@ -1,6 +1,7 @@
-use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::ty::is_type_diagnostic_item;
 use rustc_ast::ast::LitKind;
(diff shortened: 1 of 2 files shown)


llogiq:
Thank you! @bors r+

PR #7811, opened 2021-10-12
avoid `eq_op` in test code

llogiq:
Add a check to `eq_op` that will avoid linting in functions annotated with `#[test]`

---

changelog: avoid `eq_op` in test functions


llogiq:
Ah, I see! I change the compiletest parameters without cloning the Config, and as a result, everything afterwards gets compiled with `--test`!

Luckily that's easy to fix.

llogiq:
Now all tests pass this should be ready for review.

llogiq:
@xFrednet Done!

xFrednet:
Thank you for the changes! :upside_down_face: 

@bors r+

Issue #6623, opened 2021-01-22
[Roadmap] False Positive Rate


flip1995:
In the worst case, new lints are only available in nightly for 2 weeks, before
hitting beta and ultimately stable. This and the fact that fewer people use
nightly Rust nowadays makes it more probable that a lint with many FPs hits
stable. This leads to annoyed users, that will disable these new lints in the
best case and to more annoyed users, that will stop using Clippy in the worst.
A process should be developed and implemented to prevent this from happening.

#6429

Steps to completion:
- [x] Implement crater aka `cargo dev-lintcheck` for Clippy (https://github.com/rust-lang/rust-clippy/pull/6469)
- [ ] Implement framework so that lints can be enabled in nightly only
    - Previous attempt to implement this in Clippy #8211
- [x] Define clear process for stabilizing lints
    - We now have a FCP process where 2 team members have to approve on a new lint and no other team member vetos it.
- [ ] ~~CFP: Ask people using Clippy to submit their crates, so we can check new lints on them before releasing the lints to stable~~
    - We test ~200 crates on CI, so that wouldn't help much more.
- [x] Make it easy to run the `crater` tool on peoples private code / CI to get even more input
    - Lintcheck is now run on CI

EdJoPaTo:
Something I think that would be helpful to deal with false positives would be #3122. In case false positives are found they will be allowed short term.
In the long term this false positive will be fixed and the allow comment shouldn't be there anymore.

flip1995:
Good news: @xFrednet is currently working on this: https://github.com/rust-lang/rust/pull/87835

xFrednet:
I would like to take a stab at the second point: *"Implement framework so that lints can be enabled in nightly only"*. I have three ideas how this can be implemented, with advantages and drawbacks. I'll create an issue outlining the possibilities and to ask for feedback. :upside_down_face:

matthiaskrgr:
To add an additional datapoint, I have run clippy on all files (standalone) in the rustc repo [ [1]](github.com/matthiaskrgr/icemaker/)
(basically: run cargo clippy --fix -Wall::lints on a file, check if rustfix fails to apply clippys suggestion, bisect the offending lint), and these are the lints and the number of times a applied lint suggestion caused a compilation failure (aka invalid suggestion)

````
      1       "--force-warn clippy::assign-op-pattern",
      1       "--force-warn clippy::blocks-in-if-conditions",
      1       "--force-warn clippy::bool-comparison",
[...]

PR #7813, opened 2021-10-12
Add Clippy version to Clippy's lint list

xFrednet:
Hey, hey, the semester is finally over, and I wanted to get back into hacking on Clippy. It has also been some time since our metadata collection monster has been feed. So, this PR adds a new attribute `clippy::version` to document which version a lint was stabilized. I considered using `git blame` but that would be very hacky and probably not accurate. 

I'm also thinking that this attribute can be used to have a `clippy::nightly` lint group which is allow-by-default that delays setting the actual lint group until the defined version is reached. Just something to consider regarding #6623 :upside_down_face:

This PR only adds the version to 4 lints to keep it reviewable. I'll do a followup PR to add the version to other lints if the implementation is accepted :upside_down_face: 

image

Also, mobile approved xD

image

---

r? @flip1995

cc: #7172 

closes: #6492

changelog: [Clippy's lint list](https://rust-lang.github.io/rust-clippy/master/index.html) now displays the version a lint was added. :tada: 

---

Example lint declaration after this update:

```rs
declare_clippy_lint! {
    /// [...]
    ///
    /// ### Example
    /// ```rust
    /// // Bad
    /// let x = 3.14;
    /// // Good
    /// let x = std::f32::consts::PI;
    /// ```
    pub APPROX_CONSTANT,
    correctness,
    "the approximate of a known float constant (in `std::fXX::consts`)"
}
```

### DIFF
--- a/clippy_dev/Cargo.toml
+++ b/clippy_dev/Cargo.toml
@@ -12,6 +12,7 @@ opener = "0.5"
 regex = "1.5"
 shell-escape = "0.1"
 walkdir = "2.3"
+cargo_metadata = "0.12"
 
 [features]
 deny-warnings = []
--- a/clippy_dev/Cargo.toml
+++ b/clippy_dev/Cargo.toml
@@ -12,7 +12,7 @@ opener = "0.5"
 regex = "1.5"
 shell-escape = "0.1"
 walkdir = "2.3"
-cargo_metadata = "0.12"
+cargo_metadata = "0.14"
 
 [features]
 deny-warnings = []
--- a/clippy_dev/src/new_lint.rs
+++ b/clippy_dev/src/new_lint.rs
@@ -141,7 +141,7 @@ fn get_stabilisation_version() -> String {
         }
     }
 
-    String::from("<TODO set version(see doc/adding_lints.md)")
+    String::from("<TODO set version(see doc/adding_lints.md)>")
 }
 
 fn get_test_file_contents(lint_name: &str, header_commands: Option<&str>) -> String {
--- a/clippy_dev/src/new_lint.rs
+++ b/clippy_dev/src/new_lint.rs
@@ -132,6 +132,18 @@ fn to_camel_case(name: &str) -> String {
         .collect()
 }
 
+fn get_stabilisation_version() -> String {
+    let mut command = cargo_metadata::MetadataCommand::new();
+    command.no_deps();
+    if let Ok(metadata) = command.exec() {
+        if let Some(pkg) = metadata.packages.iter().find(|pkg| pkg.name == "clippy") {
+            return format!("{}.{}.0", pkg.version.minor, pkg.version.patch);
+        }
+    }
+
+    String::from("<TODO set version(see doc/adding_lints.md)")
+}
+
 fn get_test_file_contents(lint_name: &str, header_commands: Option<&str>) -> String {
     let mut contents = format!(
         indoc! {"
@@ -178,6 +190,7 @@ fn get_lint_file_contents(lint: &LintData<'_>, enable_msrv: bool) -> String {
         },
     };
 
+    let version = get_stabilisation_version();
     let lint_name = lint.name;
     let category = lint.category;
     let name_camel = to_camel_case(lint.name);
@@ -212,7 +225,7 @@ fn get_lint_file_contents(lint: &LintData<'_>, enable_msrv: bool) -> String {
     });
 
     result.push_str(&format!(
-        indoc! {"
+        indoc! {r#"
             declare_clippy_lint! {{
                 /// ### What it does
                 ///
@@ -226,11 +239,13 @@ fn get_lint_file_contents(lint: &LintData<'_>, enable_msrv: bool) -> String {
                 /// ```rust
                 /// // example code which does not raise clippy warning
                 /// ```
+                #[clippy::version = "{version}"]
                 pub {name_upper},
                 {category},
-                \"default lint description\"
+                "default lint description"
             }}
-        "},
+        "#},
+        version = version,
         name_upper = name_upper,
         category = category,
     ));
--- a/clippy_dev/src/update_lints.rs
+++ b/clippy_dev/src/update_lints.rs
@@ -18,6 +18,7 @@ static DEC_CLIPPY_LINT_RE: SyncLazy<Regex> = SyncLazy::new(|| {
         r#"(?x)
     declare_clippy_lint!\s*[\{(]
     (?:\s+///.*)*
+    (?:\s*\#\[clippy::version\s*=\s*"[^"]*"\])?
     \s+pub\s+(?P<name>[A-Z_][A-Z_0-9]*)\s*,\s*
     (?P<cat>[a-z_]+)\s*,\s*
     "(?P<desc>(?:[^"\\]+|\\(?s).(?-s))*)"\s*[})]
--- a/clippy_dev/src/update_lints.rs
+++ b/clippy_dev/src/update_lints.rs
@@ -18,7 +18,7 @@ static DEC_CLIPPY_LINT_RE: SyncLazy<Regex> = SyncLazy::new(|| {
         r#"(?x)
     declare_clippy_lint!\s*[\{(]
     (?:\s+///.*)*
-    (?:\s*\#\[clippy::version\s*=\s*"[^"]*"\])
+    (?:\s*\#\[clippy::version\s*=\s*"[^"]*"\])?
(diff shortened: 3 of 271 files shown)


xFrednet:
> > And I decided to write a script using nushell because why not? This was a mistake... I spend way more time on this than I would like to admit. It has definitely been more than 4 hours
> 
> 4 hours is not too bad for this.

I always underestimate the time a small change will take. A coworker once told me that it's usually a good idea to multiply the early estimated time for a task by 2. For me, this multiplier is about 5 times, as I always oversimplify stuff xD.

> I would have written a few simple python scripts for this.

That would have probably been the smart move. Python is a nice scripting language, it has been some time since I used it, though. :sweat_smile: 

> Since my suggestions may require some new scripts, let me know if you want me to write them or want some help writing them. (I can only help with python scripts, because I don't know anything about nushell)

The adjustments should be pretty simple with some wizardry :sparkles:. I can simply rename the [`v0.0.212`](https://github.com/rust-lang/rust-clippy/tree/gh-pages/v0.0.212) folder to `rust-1.28.0` run the script again and then use `sed` to replace each `#[clippy::version = "1.28.0"]` with `#[clippy::version = "pre 1.29.0"]`. The other changes required some manual adjustment anyway, they are luckily limited to a few lints :upside_down_face:. Thank you for the offer, though, I'll ping you if I get stuck.

flip1995:
Thanks for keeping this up-to-date-ish! Before I do a final review and we merge this, I want to ask in today's meeting what other people think of it.

xFrednet:
The text for the version label has been discussed on Zulip in [today's meeting](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202021-11-02/near/260017257). We've agreed that "Added in" might work well. I've tried the other suggestions and also wanted to showcase them.

1. "Added in"

    image

2. "Added in Rust"
    
   image

3. "Added in Rust version"

    image

4. "Added in version"

    image

I've also updated the label to use a monospace font, I think that makes the version more readable. The other labels remain the same.

From image to image
 
cc: @rust-lang/clippy

flip1995:
I like the "Added in" version most. Thanks for providing the comparison!

clippy_dev/src/new_lint.rs
```diff
@@ -132,6 +132,18 @@ fn to_camel_case(name: &str) -> String {
...
+fn get_stabilisation_version() -> String {
+    let mut command = cargo_metadata::MetadataCommand::new();
+    command.no_deps();
+    if let Ok(metadata) = command.exec() {
+        if let Some(pkg) = metadata.packages.iter().find(|pkg| pkg.name == "clippy") {
+            return format!("{}.{}.0", pkg.version.minor, pkg.version.patch);
+        }
+    }
+
+    String::from("<TODO set version(see doc/adding_lints.md)")
```
flip1995, line 144:
```suggestion
    String::from("<TODO set version(see doc/adding_lints.md)>")
```

clippy_dev/Cargo.toml
```diff
@@ -12,6 +12,7 @@ opener = "0.5"
 regex = "1.5"
 shell-escape = "0.1"
 walkdir = "2.3"
+cargo_metadata = "0.12"
```
flip1995, line 15:
```suggestion
cargo_metadata = "0.14"
```

flip1995 (commented):
I also noticed, that no lints were actually added in 1.29.0. I guess this is because 0.1.29 is the same version as 0.0.212. I don't think a action is required here.

Sorry to kill the fun in the comments :stuck_out_tongue:

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -552,6 +569,13 @@ fn extract_attr_docs(cx: &LateContext<'_>, item: &Item<'_>) -> Option<String> {
     Some(docs)
 }
 
+fn get_lint_version(cx: &LateContext<'_>, item: &Item<'_>) -> String {
+    extract_clippy_version_value(cx, item).map_or_else(
+        || VERION_DEFAULT_STR.to_string(),
+        |version| version.as_str().to_string(),
+    )
```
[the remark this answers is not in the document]
xFrednet: No, `version.as_str()` returns a `SymbolStr` which is not the same as `VERION_DEFAULT_STR`. Referencing the symbol string also doesn't work, as that would take a reference of a temporary value.

clippy_lints/src/utils/internal_lints.rs
```diff
@@ -314,6 +316,26 @@ declare_clippy_lint! {
...
+    /// ```txt
+    /// +-------------------------------+
+    /// | Valid values are:             |
+    /// |  * "pre 1.29.0"               |
+    /// |  * any valid semantic version |
+    /// +-------------------------------+
+    ///                  \
+    ///                   \  (^v^)
+    ///                     <(   )>
+    ///                       w w
```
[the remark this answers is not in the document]
xFrednet: I totally forgot this masterpiece :joy:. It's removed, no problem

clippy_dev/src/update_lints.rs
```diff
@@ -495,20 +497,23 @@ fn test_parse_contents() {
     let result: Vec<Lint> = parse_contents(
         r#"
 declare_clippy_lint! {
+    #[clippy::version = "Hello Clippy!"]
```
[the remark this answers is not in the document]
xFrednet: Agreed, I opted for a lint to have a better error message. Having the validation here only mentioned that one lint declaration was malformed. The lint was also simpler to implement.

xFrednet:
> I also noticed, that no lints were actually added in 1.29.0. I guess this is because 0.1.29 is the same version as 0.0.212. I don't think an action is required here.

Ups, good to know ^^. The difference shouldn't really matter :)

> Sorry to kill the fun in the comments

No problem ^^. I'm sometimes a bit unsure what is okay and what not, when it comes to these kinds of things. :upside_down_face: It should be ready now :)

clippy_lints/src/deprecated_lints.rs
```diff
@@ -192,6 +207,7 @@ declare_deprecated_lint! {
     /// ### Deprecation reason
     /// The `avoid_breaking_exported_api` config option was added, which
     /// enables the `wrong_self_conversion` lint for public items.
+    #[clippy::version = "1.54.0"]
```
flip1995, line 210: What is being displayed for deprecated lints? Also "Added in: <version>"? If so, this is probably wrong, since we decided earlier to put the version when the lint was removed here.

flip1995 (commented):
What happens when the version is not defined for a lint? Do we have an automated check, that makes sure that a version is defined for every lint?

clippy_lints/src/deprecated_lints.rs
xFrednet: Good catch! I've updated the page to display "Deprecated in `<version_label>`" for these lints.

xFrednet:
> What happens when the version is not defined for a lint? Do we have an automated check, that makes sure that a version is defined for every lint?

No, we don't. I misunderstood you before and thought it shouldn't be mandatory. I'll add a new lint internal lint to make the attribute mandatory. That ensure nice error messages and enables us to allow it for internal lints :upside_down_face:

flip1995:
> I misunderstood you before and thought it shouldn't be mandatory. I'll add a new lint internal lint to make the attribute mandatory.

Oh I think I said that before :smile: But now, I think it's better to just require it for every lint to avoid getting to a point where we mark _some_ of our new lints when we remember to do so, but get more inconsistent over time.

flip1995 (approved):
LGTM. Let me quickly try to build the website locally, and this should be ready to merge.

flip1995:
@bors r+

Thanks! And sorry for taking so long for reviewing such a bit-rotty PR.

(The next possible improvement for our website would be to add some more filters. E.g. filtering for version added, applicability, ... This will also involve making the existent filters more usable)

xFrednet:
No worries :upside_down_face:, thanks for the review!

> (The next possible improvement for our website would be to add some more filters. E.g. filtering for version added, applicability, ... This will also involve making the existent filters more usable)

My next goal for the website was collecting renamed lints and create some kind of indicator, when somebody searches for the old name. Expanding on the filters is also a good idea, a copy-lint-name function would also be nice. Currently, I'm trying to wrap up my larger project before starting on my bachelor thesis in December. I'll create issues for the proposed improvements, these can also be labeled as good-first-issues :upside_down_face:

xFrednet:
I've created #7958 and #7959. Could you maybe expand your suggestion a bit, what you meant with "making the existent filters more usable" in  #7958.

PR #7819, opened 2021-10-14
avoid linting `possible_truncation` on bit-reducing operations

llogiq:
---

changelog: avoid linting `possible_truncation` on bit-reducing operations


llogiq:
There's still something wrong; I have apparently picked the false test case. Will fix shortly.

llogiq:
@flip1995 this is now ready for review.

llogiq:
Anything else we need to look at?

flip1995:
> Anything else we need to look at?

There are probably other cases, where reductions could be applied, but those can just be added to the `apply_reductions` function once they come up.

@bors r+

Thanks!

PR #7833, opened 2021-10-17
Add reference to another doc with explanation

Dmitry-Borodin:
Add reference to another doc that explains which repository should be passed in this command since this is not covered in the command help itself.
changelog: none


Dmitry-Borodin:
Btw I don't understand yet why I should point to a copy of rustc repository, instead of original one. It wont commit staff there anyway, right?
Not sure if I should ask it here or open issue for that..

Dmitry-Borodin:
@giraffate done.

PR #7877, opened 2021-10-25
[`swap`] lints now check if there is `no_std` or `no_core` attribute

### LINKED ISSUE #7858: clippy suggests std::foo with #[!no_std] being enabled
Lint name: clippy::manual_swap

I tried this code:

```rust
#![no_std]
 ....

        let t = self.0;
        self.0 = self.1;
        self.1 = t;

```

I expected to see this happen: clippy sits quiet

Instead, this happened: got a "warning: this looks like you are swapping `self.0` and `self.1` manually"
Proposed change won't compile.

### Meta

```
commit-hash: 09c42c45858d5f3aedfa670698275303a3d19afa
commit-date: 2021-10-18
release: 1.56.0
LLVM version: 13.0.0
```

dswij:
Closes #7858 

changelog: [`swap`] lints now check if there is `no_std` or `no_core` attribute

### DIFF
--- a/clippy_lints/src/swap.rs
+++ b/clippy_lints/src/swap.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::source::snippet_with_applicability;
 use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_diagnostic_item;
-use clippy_utils::{can_mut_borrow_both, differing_macro_contexts, eq_expr_value};
+use clippy_utils::{can_mut_borrow_both, differing_macro_contexts, eq_expr_value, std_or_core};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::{BinOpKind, Block, Expr, ExprKind, PatKind, QPath, Stmt, StmtKind};
@@ -113,6 +113,8 @@ fn generate_swap_warning(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>, spa
 
     let first = Sugg::hir_with_applicability(cx, e1, "..", &mut applicability);
     let second = Sugg::hir_with_applicability(cx, e2, "..", &mut applicability);
+    let Some(sugg) = std_or_core(cx) else { return };
+
     span_lint_and_then(
         cx,
         MANUAL_SWAP,
@@ -122,11 +124,11 @@ fn generate_swap_warning(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>, spa
             diag.span_suggestion(
                 span,
                 "try",
-                format!("std::mem::swap({}, {})", first.mut_addr(), second.mut_addr()),
+                format!("{}::mem::swap({}, {})", sugg, first.mut_addr(), second.mut_addr()),
                 applicability,
             );
             if !is_xor_based {
-                diag.note("or maybe you should use `std::mem::replace`?");
+                diag.note(&format!("or maybe you should use `{}::mem::replace`?", sugg));
             }
         },
     );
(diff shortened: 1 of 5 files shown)


clippy_lints/src/swap.rs
```diff
@@ -113,6 +113,18 @@ fn generate_swap_warning(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>, spa
 
     let first = Sugg::hir_with_applicability(cx, e1, "..", &mut applicability);
     let second = Sugg::hir_with_applicability(cx, e2, "..", &mut applicability);
+    let sugg = if !is_no_std_crate(cx) {
+        "std"
+    } else if is_no_core_crate(cx) {
+        ""
+    } else {
+        "core"
+    };
```
camsteffen, line 122: Could use one more util `std_or_core() -> Option<&'static str>`. Shouldn't use empty strings as "none"`.

tests/ui/crate_level_checks/no_std_swap.rs
```diff
@@ -0,0 +1,23 @@
...
+}
+
+#[allow(clippy::empty_loop)]
+#[panic_handler]
+fn panic(_info: &PanicInfo) -> ! {
+    loop {}
+}
+
+#[lang = "eh_personality"]
+extern "C" fn eh_personality() {}
```
camsteffen, line 23: What is this for?

clippy_lints/src/swap.rs
dswij: Was going for a direct return there on the last else statement, but I did this as a workaround.

```rust
    let sugg = if !is_no_std_crate(cx) {
        "std"
    } else if !is_no_core_crate(cx) {
        "core"
    } else {
        return;
    };
```

Now that https://github.com/rust-lang/rust-clippy/pull/7895 is merged, I think we can go for this, wdyt?

tests/ui/crate_level_checks/no_std_swap.rs
dswij: Since we don't explicitly declare unconditional abort on panic, we need to declare `eh_personality` for the `panic_handler` above.

clippy_lints/src/swap.rs
camsteffen: I think `std_or_core` would be good to have in `clippy_utils`. You could sneak in a let-else.

```rust
let Some(sugg) = std_or_core(cx) else { return };
```

tests/ui/crate_level_checks/no_std_swap.rs
camsteffen: Oh, ha TIL. Can we just use `#![crate_type = "lib"]`?

dswij: Oh, you're right, we can use `#![crate_type = "lib"]` for this. That might be better

clippy_lints/src/swap.rs
dswij: `let-else` sounds amazing

dswij: Oh whoops, `semicolon-if-nothing-returned` FP on this?

```rust
let Some(sugg) = std_or_core(cx) else { return };
```

produces

```
   --> src/swap.rs:116:5
    |
116 |     let Some(sugg) = std_or_core(cx) else { return };
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: add a `;` here: `let Some(sugg) = std_or_core(cx) else { return };;`
    |
    = note: `-D clippy::semicolon-if-nothing-returned` implied by `-D clippy::pedantic`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#semicolon_if_nothing_returned
```

I will open an issue for this 😅

camsteffen (approved):
LGTM but we need to do something about the false positive in the CI build. Would you be interested in fixing that in another PR?

dswij:
FP in CI build should be fixed by #7955

camsteffen:
Nice. Can you squash some commits?

dswij:
> Nice. Can you squash some commits?

Done

PR #7890, opened 2021-10-27
Ignore references to type aliases in ptr_arg

### LINKED ISSUE #7699: ptr-arg warning on reference to type alias
Lint name: ptr_arg

I have a type alias somewhere:
```rust
type DocumentMap = Vec<Vec<Something>>;
```

and I have a function that takes a refence to one:

```rust
fn foo(map: &DocumentMap) { .. }
```
Clippy, run with `-Dwarnings`, returns this suggestion:
```
error: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices
   --> etc
    |
193 |     map: &DocumentMap,
    |                   ^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

```

I think this suggestion is bad -- I want to keep this type alias. I have to pepper each function that receives such an arg with `#[allow(clippy::ptr_arg)]` and a comment explaining why.

### Meta

```
rustc 1.55.0 (c8dfcfe04 2021-09-06)
binary: rustc
commit-hash: c8dfcfe046a7680554bf4eb612bad840e7631c4b
commit-date: 2021-09-06
release: 1.55.0
LLVM version: 12.0.1
```
clippy version:
```
clippy 0.1.55 (c8dfcfe 2021-09-06)
```

Alexendoo:
Works using the fact that the hir path will point to a TyAlias, rather than being resolved to the underlying type

Fixes #7699

changelog: [`ptr_arg`] No longer lints references to type aliases

### DIFF
--- a/clippy_lints/src/ptr.rs
+++ b/clippy_lints/src/ptr.rs
@@ -3,16 +3,16 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::ptr::get_spans;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::ty::{is_type_diagnostic_item, match_type, walk_ptrs_hir_ty};
+use clippy_utils::ty::walk_ptrs_hir_ty;
 use clippy_utils::{expr_path_res, is_lint_allowed, match_any_diagnostic_items, paths};
 use if_chain::if_chain;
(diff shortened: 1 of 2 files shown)


camsteffen:
Thanks!

@bors r+

There still might be an issue with the span in the original issue, but that is a separate issue.

Issue #7868, opened 2021-10-23
ICE with `clippy::undocumented_unsafe_blocks`


dylni:
### Code

main.rs:

```rust
#![warn(clippy::undocumented_unsafe_blocks)]

mod zero;

fn main() {}
```

zero.rs:

```rust
fn zero() {
    unsafe { 0 };
}
```

### Meta

**Rust version (`rustc -v`):**
```
rustc 1.58.0-nightly (514b38779 2021-10-22)
```

### Error output

```
thread 'rustc' panicked at 'byte index 36 is out of bounds of `fn zero() {
    unsafe { 0 };
}
`', src/tools/clippy/clippy_lints/src/undocumented_unsafe_blocks.rs:149:23
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust-clippy/issues/new

note: Clippy version: clippy 0.1.58 (514b387 2021-10-22)

query stack during panic:
#0 [analysis] running analysis passes on this crate
end of query stack
```

<details><summary>Backtrace</summary>
  <p>
  
  ```
   0: rust_begin_unwind
             at /rustc/514b3877956dc594823106b66c164f8cdbc8b3da/library/std/src/panicking.rs:495:5
   1: core::panicking::panic_fmt
             at /rustc/514b3877956dc594823106b66c164f8cdbc8b3da/library/core/src/panicking.rs:107:14
   2: core::str::slice_error_fail
   3: clippy_lints::undocumented_unsafe_blocks::UndocumentedUnsafeBlocks::block_has_safety_comment
   4: <clippy_lints::undocumented_unsafe_blocks::UndocumentedUnsafeBlocks as rustc_lint::passes::LateLintPass>::check_block
   5: <rustc_lint::late::LateLintPassObjects as rustc_lint::passes::LateLintPass>::check_block
   6: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_block
   7: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_expr
   8: <rustc_lint::late::LateContextAndPass<T> as rustc_hir::intravisit::Visitor>::visit_stmt
   9: rustc_hir::intravisit::walk_block
[...]

PR #7891, opened 2021-10-28
Fix ice in `undocumented_unsafe_blocks`

giraffate:
Fix https://github.com/rust-lang/rust-clippy/issues/7868

changelog: Fix ice in [`undocumented_unsafe_blocks`]


dylni:
Thanks @giraffate!

PR #7894, opened 2021-10-29
Extend author lint

Serial-ATA:
changelog: none

* Print float and int suffixes
* Print labels
* Struct field checks
* Repeat length expression check
* Destructure method calls
* Destructure closures


clippy_lints/src/utils/author.rs
```diff
@@ -190,17 +193,37 @@ impl PrintVisitor {
...
 }
 
-struct PrintVisitor {
+struct PrintVisitor<'a, 'tcx> {
     /// Fields are the current index that needs to be appended to pattern
     /// binding names
     ids: FxHashMap<&'static str, usize>,
     /// the name that needs to be destructured
     current: String,
+    cx: Option<&'a LateContext<'tcx>>,
```
camsteffen, line 223: Why is this optional?

clippy_lints/src/utils/author.rs
```diff
@@ -190,17 +193,37 @@ impl PrintVisitor {
...
+            let label_name_bind = self.next("label_name");
+            let label_name = label.ident.name;
+
+            println!(
+                "    if {}.ident.name.as_str() == {:?};",
+                label_name_bind,
+                label_name.as_str()
+            );
+        } else {
+            println!("    if {}.is_none();", self.current);
```
camsteffen, line 212: I'd skip the None case

clippy_lints/src/utils/author.rs
```diff
@@ -400,9 +475,29 @@ impl<'tcx> Visitor<'tcx> for PrintVisitor {
...
+                    "Closure(ref {}, ref {}, _, _, ref {}, _) = {};",
+                    capture_by_pat, async_pat, fn_expr_pat, current
+                );
+
+                println!("    if let CaptureBy::{:?} = {}", capture_clause, capture_by_pat);
+
+                let hir = self.cx.unwrap().tcx.hir();
+                let body = hir.body(body);
+
+                if let Some(GeneratorKind::Async(AsyncGeneratorKind::Closure)) = body.generator_kind {
```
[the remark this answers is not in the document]
camsteffen: I think no because the generator is something that gets desugared alongside the body that is directly from code.

Use `-Zunpretty=hir-tree` to see what the HIR is for any given code.

tests/ui/author/blocks.stdout
```diff
@@ -1,13 +1,23 @@
...
     if let ExprKind::Block(ref block) = expr.kind;
-    if block.stmts.len() == 2;
+    if block.stmts.len() == 3;
     if let StmtKind::Local(ref local) = block.stmts[0].kind;
     if let Some(ref init) = local.init;
     if let ExprKind::Lit(ref lit) = init.kind;
-    if let LitKind::Int(42, _) = lit.node;
+    if let LitKind::Int(42, ref suffix) = lit.node;
+    if let LitIntType::Signed(ref int_ty) = suffix
+    if let IntTy::I32 = int_ty
```
camsteffen, line 9:
```suggestion
    if let LitKind::Int(42, LitIntType::Signed(IntTy::I32)) = lit.node;
```

tests/ui/author/blocks.stdout
```diff
@@ -38,3 +48,14 @@ if_chain! {
         // report your lint here
     }
 }
+if_chain! {
+    if let ExprKind::Closure(ref capture_by, ref async, _, _, ref fn_expr, _) = expr.kind;
+    if let CaptureBy::Value = capture_by
+    if let Async::No = async
```
camsteffen, line 54:
```suggestion
    if let ExprKind::Closure(CaptureBy::Value, Async::No, _, _, ref fn_expr, _) = expr.kind;
```

tests/ui/author/for_loop.stdout
```diff
@@ -33,31 +40,176 @@ if_chain! {
...
+    if label.is_none();
+    if let PatKind::Binding(BindingAnnotation::Mutable, _, name4, None) = arms[0].pat.kind;
+    if name4.as_str() == "iter";
+    then {
+        // report your lint here
+    }
+}
+if_chain! {
+    if let ExprKind::DropTemps(ref expr) = expr.kind;
+    if let ExprKind::Match(ref expr1, ref arms, MatchSource::ForLoopDesugar) = expr.kind;
```
camsteffen, line 77: While you're at it, can you integrate `higher`?
```suggestion
    if let Some(ForLoop { pat, arg, body, .. }) = higher::ForLoop::parse(expr);
```

camsteffen (changes requested):

clippy_lints/src/utils/author.rs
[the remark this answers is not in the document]
Serial-ATA: Thanks, that flag makes things a lot easier.

clippy_lints/src/utils/author.rs
Serial-ATA: I was only using it in `visit_expr`, but I guess it doesn't really matter.

tests/ui/author/blocks.stdout
Serial-ATA: Should the same be done with float suffixes?

tests/ui/author/for_loop.stdout
Serial-ATA: Is author supposed to print the pattern as it would be in `rustc_ast` or `rust_hir`? I made `ExprKind::Closure` print as it would be in `rustc_ast`, but `ExprKind::Match` is printed as it is in `rustc_hir`.

tests/ui/author/blocks.stdout
camsteffen: Yes I think so

tests/ui/author/for_loop.stdout
camsteffen: Author outputs code for matching the HIR.

tests/ui/author/loop.stdout
```diff
@@ -0,0 +1,145 @@
...
+    if block.stmts.len() == 1;
+    if let StmtKind::Semi(ref e, _) = block.stmts[0].kind
+    if let ExprKind::Break(ref destination, None) = e.kind;
+    if block.expr.is_none();
+    if let ExprKind::Block(ref block1, ref label1) = else_expr.kind;
+    if block1.stmts.len() == 1;
+    if let StmtKind::Expr(ref e1, _) = block1.stmts[0].kind
+    if let ExprKind::Break(ref destination1, None) = e1.kind;
+    if block1.expr.is_none();
+    if let ExprKind::Loop(ref body1, ref label2, LoopSource::While) = label1;
```
camsteffen, line 91: Something's not right here.

clippy_lints/src/utils/author.rs
```diff
@@ -190,43 +195,343 @@ impl PrintVisitor {
...
+        let else_ = if opt_else.is_some() {
+            format!("Some({})", else_pat)
+        } else {
+            String::from("None")
+        };
+
+        match cond.kind {
+            ExprKind::DropTemps(cond) => {
+                println!(
+                    "    if let Some(If {{ {} @ cond, {} @ then, r#else: {}}}) = higher::If::hir({})",
```
camsteffen, line 327: I think this notation is more common for field bindings.

```suggestion
                    "    if let Some(If {{ cond: {}, {} @ then, r#else: {}}}) = higher::If::hir({})",
```

clippy_lints/src/utils/author.rs
```diff
@@ -190,43 +195,343 @@ impl PrintVisitor {
...
+
+                        self.current = body_pat;
+                        self.visit_expr(body_expr);
+                    },
+                    ExprKind::Let(pat, expr, _) => {
+                        let let_pat = self.next("let_pat");
+                        let let_expr_pat = self.next("let_expr");
+                        let if_then_pat = self.next("if_then");
+
+                        println!("    if let Some(WhileLet {{ {} @ let_pat, {} @ let_expr, {} @ if_then }}) = higher::WhileLet::hir({})", let_pat, let_expr_pat, if_then_pat, self.current);
```
camsteffen, line 282: Ideally each of the `higher` outputs correspond to actually using `higher` in the `author` implementation. We shouldn't re-implement `higher`, but instead do `if let Some(...) = higher...` for every thing on every expression.

Although you can make an exception for `if` and `if let` since these are very simple and are not really desugaring. I wouldn't mind not even using `higher` for those.

One thing you could do is `match expr.span.ctxt().outer_expn_data().kind` to see what is desugared before invoking `higher`, but not necessary.

tests/ui/author/if.stdout
```diff
@@ -1,32 +1,48 @@
...
     if block.stmts.len() == 1;
     if let StmtKind::Semi(ref e, _) = block.stmts[0].kind
     if let ExprKind::Binary(ref op, ref left, ref right) = e.kind;
     if BinOpKind::Eq == op.node;
-    if let ExprKind::Lit(ref lit) = left.kind;
-    if let LitKind::Int(2, _) = lit.node;
-    if let ExprKind::Lit(ref lit1) = right.kind;
-    if let LitKind::Int(2, _) = lit1.node;
+    if let ExprKind::Lit(ref lit1) = left.kind;
+    if let LitKind::Int(1, LitIntType::Unsuffixed) = lit1.node;
```
camsteffen, line 13: Hmm how did the number change?

```suggestion
    if let LitKind::Int(2, LitIntType::Unsuffixed) = lit1.node;
```

camsteffen (changes requested):

tests/ui/author/if.stdout
Serial-ATA: Looks like it used to check the else block first. https://github.com/rust-lang/rust-clippy/blob/310ecb0a5c0d9161679c256052822f34c6eb548a/tests/ui/author/if.stdout#L4-L5

tests/ui/author/loop.stdout
Serial-ATA: Oops, missed an early return there.

clippy_lints/src/utils/author.rs
```diff
@@ -190,43 +191,300 @@ impl PrintVisitor {
...
+    cx: &'a LateContext<'tcx>,
 }
 
-impl<'tcx> Visitor<'tcx> for PrintVisitor {
+impl<'a, 'tcx> Visitor<'tcx> for PrintVisitor<'a, '_> {
     type Map = Map<'tcx>;
 
     #[allow(clippy::too_many_lines)]
     fn visit_expr(&mut self, expr: &Expr<'_>) {
-        print!("    if let ExprKind::");
```
camsteffen, line 208: Is there a particular reason to change this?

Serial-ATA: Nope, forgot to add that back after adding `check_higher`. Thanks :)

camsteffen (approved):
LGTM after rebase

camsteffen:
Thanks!

@bors r+

PR #7895, opened 2021-10-29
Disable "if_not_else" lints from firing on else-ifs

### LINKED ISSUE #7892: `if_not_else` lints firing on `else-if`s
Lint name: `if_not_else`

I tried this code:

```rust
fn foo() {}
fn bar() {}
fn nothing() {}

#[warn(clippy::if_not_else)]
fn main() {
    let foo_disabled = true;
    let bar_disabled = true;
    
    if !foo_disabled {
        foo()
    }
    else if !bar_disabled {
        bar()
    } else {
        nothing()
    }
} 
```

I expected to see this happen: nothing

Instead, this happened: clippy suggests swapping the order of `else-if` and `else` block, which makes it less clear imo. There are scenarios where using not in `else-if`s makes it more readable like the one above - you would expect it to be written in order of priority/preference. (try `foo()`, then try `bar()`, else do nothing)

```rust
warning: unnecessary boolean `not` operation
  --> src/main.rs:13:10
   |
13 |       else if !bar_disabled {
   |  __________^
14 | |         bar()
15 | |     } else {
16 | |         nothing()
17 | |     }
   | |_____^
   |
  --> src/main.rs:5:8
   |
5  | #[warn(clippy::if_not_else)]
   |        ^^^^^^^^^^^^^^^^^^^
   = help: remove the `!` and swap the blocks of the `if`/`else`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else

    Finished dev [unoptimized + debuginfo] target(s) in 4.06s
```

### Meta

```
rustc 1.46.0-nightly (f455e46ea 2020-06-20)
binary: rustc
commit-hash: f455e46eae1a227d735091091144601b467e1565
commit-date: 2020-06-20
release: 1.46.0-nightly
LLVM version: 10.0
```

See https://github.com/rust-lang/rust-clippy/pull/7877#discussion_r735263149

ahmedkrmn:
Fixes #7892

1. Convert `['if_not_else']` to `LateLintPass` and use `clippy_utils::is_else_clause` for checking.
2. Update tests.

changelog: [`if_not_else`] now ignores else if statements.

### DIFF
--- a/clippy_lints/src/if_not_else.rs
+++ b/clippy_lints/src/if_not_else.rs
@@ -2,9 +2,9 @@
 //! on the condition
 
 use clippy_utils::diagnostics::span_lint_and_help;
-use rustc_ast::ast::{BinOpKind, Expr, ExprKind, UnOp};
-use rustc_lint::{EarlyContext, EarlyLintPass};
-use rustc_middle::lint::in_external_macro;
+use clippy_utils::is_else_clause;
+use rustc_hir::{BinOpKind, Expr, ExprKind, UnOp};
+use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
 declare_clippy_lint! {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/if_not_else.rs
```diff
@@ -46,13 +47,21 @@ declare_clippy_lint! {
 
 declare_lint_pass!(IfNotElse => [IF_NOT_ELSE]);
 
-impl EarlyLintPass for IfNotElse {
-    fn check_expr(&mut self, cx: &EarlyContext<'_>, item: &Expr) {
-        if in_external_macro(cx.sess, item.span) {
+impl LateLintPass<'_> for IfNotElse {
+    fn check_expr(&mut self, cx: &LateContext<'_>, item: &Expr<'_>) {
+        if in_external_macro(cx.sess(), item.span) {
```
camsteffen, line 52: This will fix it if my suspicion is right - changing to LateLintPass means that desugaring occurs before the lint runs. `from_expansion` will catch desugaring in addition to macros.

```suggestion
        if item.span.from_expansion() {
```

ahmedkrmn: Thanks @camsteffen. This is my first time working with rustc. I didn't know this life saver function existed!
I've been debugging the issue. After dumping the expression trees, I could see that the while loop was indeed represented as an `ExprKind::If` when using `LateLintPass`which caused this problem. I thought about reverting back to `EarlyLintPass` and looking for a workaround, but your comment came just in time!

Issue #7889, opened 2021-10-27
manual_split_once lint causes a change in behaviour


Marwes:
Lint name: manual_split_once

I tried this code:

```rust
path.rsplitn(2, '/').next().unwrap()
```

I expected to see this happen: `cargo clippy` should not suggest I change it to `rsplit_once` as that changes the behaviour.

Instead, this happened: 

```
manual implementation of `rsplit_once`
   --> flux-core/src/semantic/nodes.rs:465:21
    |
465 |             None => path.rsplitn(2, '/').next().unwrap(),
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `path.rsplit_once('/').unwrap().1`
    |
```

`cargo clippy --fix` rewrote this to `path.rsplit_once('/').unwrap().1` which is not equivalent code as `rsplitn` will always return at least one item, whereas `rsplit_once` will return `None` if the separator does not exist in the string causing the unwrap to panic.

`path.rsplit_once('/').map_or(path, |t| t.1)` would be a valid transformation.

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.58.0-nightly (e269e6bf4 2021-10-26)
binary: rustc
commit-hash: e269e6bf47f40c9046cd44ab787881d700099252
commit-date: 2021-10-26
host: x86_64-unknown-linux-gnu
release: 1.58.0-nightly
LLVM version: 13.0.0
```

PR #7896, opened 2021-10-29
Fix for #7889 and add new lint needless_splitn

surechen:
fixes: #7889
1. Fix the problem of manual_split_once changing the original behavior.
2. Add a new lint needless_splitn.

changelog: Fix the problem of manual_split_once changing the original behavior and add a new lint needless_splitn.

### DIFF
--- a/clippy_lints/src/methods/str_splitn.rs
+++ b/clippy_lints/src/methods/str_splitn.rs
@@ -11,7 +11,13 @@ use rustc_span::{symbol::sym, Span, SyntaxContext};
 
 use super::MANUAL_SPLIT_ONCE;
 
-pub(super) fn check(cx: &LateContext<'_>, method_name: &str, expr: &Expr<'_>, self_arg: &Expr<'_>, pat_arg: &Expr<'_>) {
+pub(super) fn check_manual_split_once(
+    cx: &LateContext<'_>,
+    method_name: &str,
+    expr: &Expr<'_>,
+    self_arg: &Expr<'_>,
+    pat_arg: &Expr<'_>,
+) {
     if !cx.typeck_results().expr_ty_adjusted(self_arg).peel_refs().is_str() {
         return;
     }
@@ -36,7 +42,7 @@ pub(super) fn check(cx: &LateContext<'_>, method_name: &str, expr: &Expr<'_>, se
             format!("{}.{}({})", self_snip, method_name, pat_snip)
         },
         IterUsageKind::RNextTuple => format!("{}.{}({}).map(|(x, y)| (y, x))", self_snip, method_name, pat_snip),
-        IterUsageKind::Next => {
+        IterUsageKind::Next | IterUsageKind::Second => {
             let self_deref = {
                 let adjust = cx.typeck_results().expr_adjustments(self_arg);
                 if adjust.is_empty() {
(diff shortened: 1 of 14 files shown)


tests/ui/manual_split_once.stderr
```diff
@@ -22,13 +22,13 @@ error: manual implementation of `split_once`
   --> $DIR/manual_split_once.rs:17:13
    |
 LL |     let _ = "key=value".splitn(2, '=').nth(1).unwrap();
-   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `"key=value".split_once('=').unwrap().1`
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `"key=value".split_once('=').map_or("key=value", |x| x.1)`
```
camsteffen, line 25: This changes behavior since `nth(1)` is not guaranteed to return `Some`. The issue in question is `[r]splitn(..).next().unwrap()` which never panics because `splitn` always has at least one element.

tests/ui/manual_split_once.stderr
```diff
@@ -76,13 +76,13 @@ error: manual implementation of `rsplit_once`
   --> $DIR/manual_split_once.rs:41:13
    |
 LL |     let _ = "key=value".rsplitn(2, '=').next().unwrap();
-   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `"key=value".rsplit_once('=').unwrap().1`
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `"key=value".rsplit_once('=').map_or("key=value", |x| x.1)`
```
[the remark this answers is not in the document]
surechen: > This suggestion doesn't seem like an improvement from the original code to me. I think it would make more sense to do `.rsplit('=').next().unwrap()`. But then it doesn't really fit with the `manual_split_once` lint and could be a new lint like...
> 
> > `needless_splitn`: lints usage of `splitn` or `rsplitn` where `split` or `rsplit` would be equivalent.

Hi, thanks for your patient review. According to your opinion, should I ignore this situation in this pr for this lint, and then develop a new lint and submit a new pr to support this situation.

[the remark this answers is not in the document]
camsteffen: Sorry for the delay. Let's ignore this case for now and you can add a lint in a separate PR.

> In this example, I think rsplit will change the behavior of rsplitn. The result of rsplit depends on the string it processes.

Maybe I'm missing something. Do you have an example where `x.splitn(..).next()` is different from `x.split(..).next()`?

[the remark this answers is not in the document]
camsteffen: Also for the new lint, I'd prefer the implementations to be together in one file named `str_split.rs` since they are closely related.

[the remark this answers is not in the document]
surechen: > Sorry for the delay. Let's ignore this case for now and you can add a lint in a separate PR.
> 
> > In this example, I think rsplit will change the behavior of rsplitn. The result of rsplit depends on the string it processes.
> 
> Maybe I'm missing something. Do you have an example where `x.splitn(..).next()` is different from `x.split(..).next()`?

Hi, I can only think of this example at present:
let a1 = "key=value=add".splitn(1, '=').next().unwrap();
let a2 = "key=value=add".split('=').next().unwrap();
println!("{} {}", a1, a2);
result is : `key=value=add` and `key`
When pat >= 2 for splitn(), I think they are the same.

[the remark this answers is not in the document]
camsteffen: > When pat >= 2 for splitn(), I think they are the same.

Exactly. I think that's where `needless_splitn` can lint.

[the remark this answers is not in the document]
surechen: @rustbot label +S-waiting-on-author

clippy_lints/src/methods/str_splitn.rs
```diff
@@ -209,3 +242,120 @@ fn parse_iter_usage(
...
+                    expr.span,
+                    msg,
+                    "try this",
+                    format!(
+                    "{}.{}({})",
+                    snippet_with_context(cx, self_arg.span, ctxt, "..", &mut app).0,
+                    method_sugg,
+                    snippet_with_context(cx, pat_arg.span, ctxt, "..", &mut app).0
+                    ),
+                    app);
```
camsteffen, line 315: Using `splitn` on a string literal seems really unrealistic to me. Is it even worth it for Clippy to detect this? Suggesting to write `"a,b".split(',')` seems questionable since they probably shouldn't use `split` at all. (could write `["a", "b"]`)

camsteffen (changes requested):

clippy_lints/src/methods/str_splitn.rs
surechen: > Using `splitn` on a string literal seems really unrealistic to me. Is it even worth it for Clippy to detect this? Suggesting to write `"a,b".split(',')` seems questionable since they probably shouldn't use `split` at all. (could write `["a", "b"]`)

Hi，I'm sorry, I have more things to deal with than I thought. I guess I can modify this PR according to your suggestions during the weekend. Thank you very much.

camsteffen: To be clear I'm suggesting to not detect this case. Maybe we can add it in a different PR.

tests/ui/manual_split_once.stderr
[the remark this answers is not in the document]
surechen: > Sorry for the delay. Let's ignore this case for now and you can add a lint in a separate PR.
> 
> > In this example, I think rsplit will change the behavior of rsplitn. The result of rsplit depends on the string it processes.
> 
> Maybe I'm missing something. Do you have an example where `x.splitn(..).next()` is different from `x.split(..).next()`?

Ignore the case of "key=value".rsplitn(2, '=').next().unwrap() for manual_split_once.

clippy_lints/src/methods/str_splitn.rs
```diff
@@ -209,3 +238,82 @@ fn parse_iter_usage(
...
+    pat_arg: &Expr<'_>,
+    count: u128,
+) {
+    if !cx.typeck_results().expr_ty_adjusted(self_arg).peel_refs().is_str() {
+        return;
+    }
+    let ctxt = expr.span.ctxt();
+    let mut app = Applicability::MachineApplicable;
+    if_chain! {
+        if method_name == "splitn";
```
camsteffen, line 258: Why not rplitn?

surechen: > Why not rsplitn?

I think the result of rsplitn depends on the string it processes.
For example:
"key=value=add".rsplitn(2, "=").next().unwrap(); is not same as "key=value".rsplitn(2, "=").next().unwrap();

This is why I wanted to deal with rsplitn only for literal at first.

camsteffen: But isn't `string.rsplitn(2, ',').next()` the same as `string.rsplit(',').next()`?

surechen: > But isn't `string.rsplitn(2, ',').next()` the same as `string.rsplit(',').next()`?

I'm very sorry, I was amused by my mistake.

camsteffen: Lol! No worries, sometimes brains don't work.

camsteffen (approved):
LGTM! Please squash some commits.

surechen:
> LGTM! Please squash some commits.

Done. Thanks.

PR #7897, opened 2021-10-29
Replace `in_macro` usage with `from_expansion`

camsteffen:
changelog: none

Generally replace `in_macro(span)` with `span.from_expansion()`. If we're just trying to avoid expanded code, this seems more appropriate because any kind of expanded code is prone to false positives. One place I did not touch is `macro_use.rs`. I think this lint could use a rewrite so I moved `in_macro` there, the only place it is still used.

### DIFF
--- a/clippy_lints/src/macro_use.rs
+++ b/clippy_lints/src/macro_use.rs
@@ -1,5 +1,4 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::in_macro;
 use clippy_utils::source::snippet;
 use hir::def::{DefKind, Res};
 use if_chain::if_chain;
@@ -9,6 +8,7 @@ use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::hygiene::ExpnKind;
 use rustc_span::{edition::Edition, sym, Span};
 
 declare_clippy_lint! {
@@ -212,4 +212,8 @@ impl<'tcx> LateLintPass<'tcx> for MacroUseImports {
             }
         }
     }
+}
+
+fn in_macro(span: Span) -> bool {
+    span.from_expansion() && !matches!(span.ctxt().outer_expn_data().kind, ExpnKind::Desugaring(..))
 }
--- a/doc/common_tools_writing_lints.md
+++ b/doc/common_tools_writing_lints.md
@@ -8,7 +8,7 @@ You may need following tooltips to catch up with common operations.
   - [Checking for a specific type](#checking-for-a-specific-type)
   - [Checking if a type implements a specific trait](#checking-if-a-type-implements-a-specific-trait)
   - [Checking if a type defines a specific method](#checking-if-a-type-defines-a-specific-method)
-  - [Dealing with macros](#dealing-with-macros)
+  - [Dealing with macros](#dealing-with-macros-and-expansions)
 
 Useful Rustc dev guide links:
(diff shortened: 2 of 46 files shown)


camsteffen:
Eh let's wait for the sync

flip1995:
Ah nice, so the for-loop lowering PR got merged and this will get through without test regressions? Nice! I'll make sure to include this PR in the sync-back to rustc 👍

flip1995 (approved):

camsteffen:
To be clear there are a few changes beyond simply replacing the function, but it's all fairly straightforward.

clippy_lints/src/macro_use.rs
```diff
@@ -213,3 +213,7 @@ impl<'tcx> LateLintPass<'tcx> for MacroUseImports {
         }
     }
 }
+
+fn in_macro(span: Span) -> bool {
+    span.from_expansion() && !matches!(span.ctxt().outer_expn_data().kind, ExpnKind::Desugaring(..))
```
flip1995, line 218: Ah, so this is the only place where we can't get rid of the distinction between macro and desugaring, yet.

flip1995 (approved):
Now also reviewed the other small changes.

xFrednet:
Could you also update the [*Dealing with macros*](https://github.com/rust-lang/rust-clippy/blob/master/doc/common_tools_writing_lints.md#dealing-with-macros) documentation in accordance to this change? It would be good to mention that it's generally better to use `.from_expansion()` :upside_down_face:

flip1995:
> I'll make sure to include this PR in the sync-back to rustc

Since this is a NFC, I decided to rather get the sync through as fast as possible, instead of trying to get this included in the sync.

camsteffen:
Rebased. Removed some redundant `in_external_macro` calls. Edited the docs.

clippy_lints/src/macro_use.rs
camsteffen: Right. Actually this lint probably should look for `ExpnKind::Macro(MacroKind::Bang`

doc/common_tools_writing_lints.md
```diff
@@ -182,64 +182,76 @@ impl<'tcx> LateLintPass<'tcx> for MyTypeImpl {
...
+   extern crate a_crate_with_macros;
+
+   // `foo` is defined in `a_crate_with_macros`
+   foo!("bar");
+
+   // if we lint the `match` of `foo` call and test its span
+   assert_eq!(in_external_macro(cx.sess(), match_span), true);
+   ```
+
+   ```rust
```
flip1995, line 238: That code block seems to be misplaced. Did you mean to remove it? If not, this probably belongs in the `span.ctxt()` section.

camsteffen: Oh...I just realized I conflated this PR with another idea I had to remove `differing_macro_contexts`...so what do you think of that? 😄 

For now I'll just restore the lines deleted here.

camsteffen: > For now I'll just restore the lines deleted here.

This only makes sense if you view without whitespace. I'll revert changes to the `differing_macro_contexts` section.

flip1995:
Thanks! This looks good now.

About removing `differing_macro_contexts`: IIRC, the only thing that it does is returning `span1.ctxt() != span2.ctxt()`, right? So yeah, having a wrapper function for this doesn't really help complexity-wise. The question is if the implicit documentation of having this function helps. I personally don't think that it is that important, since `span1.ctxt() != span2.ctxt()` is already telling enough IMO. So I guess, it could be removed. But let's do this in an additional PR.

@bors r+ p=1 (codebase wide change)

PR #7898, opened 2021-10-29
Don't show no_effect warning on unit structs implementing fn_once

### LINKED ISSUE #7792: Calling a struct with a manual implementation of `FnOnce` trips `clippy::no_effect` despite having side effects
Lint name: `clippy::no_effect`

I tried this code ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=ef0cf18168440a92810107cf8bc078c2)):

```rust
#![feature(fn_traits, unboxed_closures)]

struct GreetStruct;

impl FnOnce<(&str,)> for GreetStruct {
    type Output = ();

    extern "rust-call" fn call_once(self, (who,): (&str,)) -> Self::Output {
        println!("hello {}", who);
    }
}

fn greet_fn(who: &str) {
    println!("hello {}", who);
}

fn main() {
    // no lint
    greet_fn("world");
    // no lint
    GreetStruct.call_once(("world",));
    // warning: statement with no effect
    GreetStruct("world");
}
```

I expected to see this happen: Clippy succeeds with no warnings.

Instead, this happened:
```
warning: statement with no effect
  --> src/main.rs:23:5
   |
23 |     GreetStruct("world");
   |     ^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::no_effect)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#no_effect
```

This statement does have an effect, as removing it changes the program's output from this:
```
hello world
hello world
hello world
```

to this:
```
hello world
hello world
```

### Meta

**Rust version (`rustc +nightly -Vv`):**
```
rustc 1.57.0-nightly (25ec82738 2021-10-05)
binary: rustc
commit-hash: 25ec8273855fde2d72ae877b397e054de5300e10
commit-date: 2021-10-05
release: 1.57.0-nightly
LLVM version: 13.0.0
```

F3real:
Fixes #7792 

changelog: Don't show [`no_effect`] or [`unecessary_operation`] warning for unit struct implementing FnOnce

### DIFF
--- a/clippy_lints/src/no_effect.rs
+++ b/clippy_lints/src/no_effect.rs
@@ -133,9 +133,12 @@ fn has_no_effect(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
         },
         ExprKind::Call(callee, args) => {
             if let ExprKind::Path(ref qpath) = callee.kind {
-                let res = cx.qpath_res(qpath, callee.hir_id);
+                if cx.typeck_results().type_dependent_def(expr.hir_id).is_some() {
+                    // type-dependent function call like `impl FnOnce for X`
+                    return false;
+                }
                 let def_matched = matches!(
-                    res,
+                    cx.qpath_res(qpath, callee.hir_id),
                     Res::Def(DefKind::Struct | DefKind::Variant | DefKind::Ctor(..), ..)
                 );
                 if def_matched || is_range_literal(expr) {
@@ -238,6 +241,10 @@ fn reduce_expression<'a>(cx: &LateContext<'_>, expr: &'a Expr<'a>) -> Option<Vec
         },
         ExprKind::Call(callee, args) => {
             if let ExprKind::Path(ref qpath) = callee.kind {
+                if cx.typeck_results().type_dependent_def(expr.hir_id).is_some() {
+                    // type-dependent function call like `impl FnOnce for X`
+                    return None;
+                }
                 let res = cx.qpath_res(qpath, callee.hir_id);
                 match res {
                     Res::Def(DefKind::Struct | DefKind::Variant | DefKind::Ctor(..), ..)
--- a/tests/ui/no_effect.rs
+++ b/tests/ui/no_effect.rs
@@ -1,4 +1,4 @@
-#![feature(box_syntax)]
+#![feature(box_syntax, fn_traits, unboxed_closures)]
 #![warn(clippy::no_effect_underscore_binding)]
 #![allow(dead_code)]
 #![allow(path_statements)]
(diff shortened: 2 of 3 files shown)


clippy_lints/src/no_effect.rs
```diff
@@ -130,10 +131,12 @@ fn has_no_effect(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
         },
         ExprKind::Call(callee, args) => {
             if let ExprKind::Path(ref qpath) = callee.kind {
-                let res = cx.qpath_res(qpath, callee.hir_id);
                 let def_matched = matches!(
-                    res,
-                    Res::Def(DefKind::Struct | DefKind::Variant | DefKind::Ctor(..), ..)
+                    cx.qpath_res(qpath, callee.hir_id),
+                    Res::Def(
+                        DefKind::Struct | DefKind::Variant | DefKind::Ctor(_, CtorKind::Fn | CtorKind::Fictive),
```
camsteffen, line 137: I don't think this is the right fix. Perhaps try inspecting `tcx.type_dependent_def(callee.hir_id)`.

Add a test with
```rust
struct GreetStruct(u32);

impl FnOnce<(&str,)> for GreetStruct {
    type Output = ();

    extern "rust-call" fn call_once(self, (who,): (&str,)) -> Self::Output {
        println!("hello {}", who);
    }
}

fn main() {
    GreetStruct(1)("world");
}
```

F3real: @camsteffen Thanks for suggestion, I am not that familiar with the rustc API. 
I looked into it briefly but it gives `none` as result, I think that it requires MethodCall and not Call as ExprKind.

camsteffen: Hmm yeah I haven't seen it work with `Call` either, it was just a guess...

camsteffen: Ah did you try `expr.hir_id`?

F3real: @camsteffen I've tried it with expr.hir_id and it works. So one option would be to add something like this:
```rust
fn is_fn_once(cx: &LateContext<'_>, expr: &'a Expr<'a>) -> bool {
    if_chain! {
        if let Some(def_id) = cx.typeck_results().type_dependent_def_id(expr.hir_id);
        if match_def_path(cx, def_id, &paths::FN_ONCE_CALL_ONCE);
        then {
            return true;
        }
    }
    false
}
```
Hardcoded path is used as it seems `FnOnce` trait has no corresponding Symbol. But I've noticed we also have false positive for `FnMut` and probably `Fn` case, so those should also be added.

Other way that we could go is more similar to first solution I had, if we use:
`Res::Def(DefKind::Variant | DefKind::Ctor(_, CtorKind::Fn | CtorKind::Fictive), ..)` in match instead of `Res::Def(DefKind::Struct | DefKind::Variant | DefKind::Ctor(..), ..)` we avoid these cases. I think original intention was to cover construction of touples (as far as I can see from tests) but implementation seems to be too broad.

camsteffen: I think you can just do this

```rust
ExprKind::Call(callee, args) => {
    if cx.typeck_results().type_dependent_def_id(expr.hir_id).is_some() {
        // type-dependent function call like `impl FnOnce for X`
        return false;
    }
    ..
```

camsteffen:
Thanks!

@bors r+

PR #7901, opened 2021-10-30
Prevent clippy::needless_lifetimes false positive in async function definition

### LINKED ISSUE #7893: Lifetime cannot be elided
Lint name: clippy::needless_lifetimes

I tried this code:

```rust
async fn func<'a>(args: &[&'a str]) -> Option<&'a str> {
    args.get(0).cloned()
}
```

I expected to see this happen: no warning as the `'a` lifetime is needed. Moreover the lint does not trigger is the function is not marked as `async`.

Instead, this happened: 

```
warning: explicit lifetimes given in parameter types where they could be elided (or replaced with `'_` if needed by type declaration)
 --> src/main.rs:1:1
  |
1 | async fn func<'a>(args: &[&'a str]) -> Option<&'a str> {
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(clippy::needless_lifetimes)]` on by default
```

### Meta

```
rustc 1.58.0-nightly (4e0d3973f 2021-10-27)
binary: rustc
commit-hash: 4e0d3973fafdfb1c51011bc74e44257b5e3863f1
commit-date: 2021-10-27
release: 1.58.0-nightly
LLVM version: 13.0.0
```

samueltardieu:
Scan `OpaqueDef` bounds for lifetimes as well. Those `OpaqueDef` instances are generated while desugaring an `async` function definition.

This fixes #7893

changelog: Prevent [`clippy::needless_lifetimes`] false positive in `async` function definition

### DIFF
--- a/tests/ui/needless_lifetimes.rs
+++ b/tests/ui/needless_lifetimes.rs
@@ -27,6 +27,11 @@ fn multiple_in_and_out_2<'a, 'b>(x: &'a u8, _y: &'b u8) -> &'a u8 {
     x
 }
 
+// No error; multiple input refs
+async fn func<'a>(args: &[&'a str]) -> Option<&'a str> {
+    args.get(0).cloned()
+}
+
 // No error; static involved.
 fn in_static_and_out<'a>(x: &'a u8, _y: &'static u8) -> &'a u8 {
     x
--- a/clippy_lints/src/lifetimes.rs
+++ b/clippy_lints/src/lifetimes.rs
@@ -378,11 +378,15 @@ impl<'a, 'tcx> Visitor<'tcx> for RefVisitor<'a, 'tcx> {
 
     fn visit_ty(&mut self, ty: &'tcx Ty<'_>) {
         match ty.kind {
-            TyKind::OpaqueDef(item, _) => {
+            TyKind::OpaqueDef(item, bounds) => {
                 let map = self.cx.tcx.hir();
                 let item = map.item(item);
                 walk_item(self, item);
                 walk_ty(self, ty);
+                self.lts.extend(bounds.iter().filter_map(|bound| match bound {
+                    GenericArg::Lifetime(l) => Some(RefLt::Named(l.name.ident().name)),
+                    _ => None,
+                }));
             },
             TyKind::BareFn(&BareFnTy { decl, .. }) => {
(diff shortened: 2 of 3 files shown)


xFrednet (approved):
This looks good to me :+1:. I'm not quite sure how these `OpaqueDef` types are used inside rustc, I'll therefore ask some team members to double-check this PR :upside_down_face:

samueltardieu:
> I'm not quite sure how these `OpaqueDef` types are used inside rustc,

According to [the documentation of `OpaqueDef`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/enum.TyKind.html#variant.OpaqueDef), this is the type used to expand `impl Trait` (and `async fn` gets expanded into a function returning `impl Future<_>`). The documentation also indicates that the second field of `OpaqueDef` contains the lifetimes bound on the `impl Trait`, hence this patch.

tests/ui/needless_lifetimes.rs
```diff
@@ -1,3 +1,4 @@
+// edition:2018
```
flip1995, line 1: This shouldn't be necessary anymore, since we use the 2021 edition by default for our tests now.

flip1995 (approved):
Thanks for the further explanation. This LGTM. There's a small NIT with the test.

tests/ui/needless_lifetimes.rs
samueltardieu: Do we? If I remove it I get an error: "Rust 2015 does not permit the use of `async fn`", and the 3 commits that are not in my "master" branch don't seem to change that.

flip1995: Huh? Let me actually try out my suggestion myself. It _should_ work after rebasing.

samueltardieu: I was out of sync with upstream, I'll rebase and update the test!

flip1995: Thanks! I rebased on the latest upstream master and it worked for me locally.

samueltardieu: You were right, rebasing allowed me to remove this line.

flip1995:
@bors r=xFrednet,flip1995

Thanks!

Issue #7737, opened 2021-09-29
Add functionality to automatically put variable names in markdown docs in ticks


DerDoni:
Wenn calling clippy with pedantic settings, it advises the user to put variables (and others) into ticks, like so:

> warning: you should put `crate::common::LogLevel` between ticks in the documentation
>   --> logger/src/rpc/mod.rs:52:38
>    |
> 52 | ///     * `loglevel` (optional): u8 [crate::common::LogLevel]
>    |                                      ^^^^^^^^^^^^^^^^^^^^^^^
>    |
>    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown

I tried the following command to automatically fix this, but to no avail
`cargo clippy --fix -- --warn clippy::doc_markdown`

Following the advice manually and possibly searching for hundreds or thousands of places and in the code is cumbersome and should be automated.
@rustbot C-enhancement

PR #7904, opened 2021-10-31
Add suggestion to missing backticks error

Serial-ATA:
changelog: Add a machine applicable suggestion for the [`doc_markdown`] missing backticks lint

closes: #7737


tests/ui/doc/doc.stderr
```diff
@@ -1,160 +1,160 @@
-error: you should put `foo_bar` between ticks in the documentation
+error: item in documentation missing backticks
   --> $DIR/doc.rs:8:9
    |
 LL | /// The foo_bar function does _nothing_. See also foo::bar. (note the dot there)
-   |         ^^^^^^^
+   |         ^^^^^^^ help: try: ``foo_bar``
```
xFrednet, line 5: Is there a reason why the suggestion uses two backticks? Markdown usually handles one and two backticks the same:

```md
Hello `how` are ``you`` doing?
```
-> Hello `how` are ``you`` doing?

tests/ui/doc/doc.stderr
```diff
@@ -1,160 +1,160 @@
-error: you should put `foo_bar` between ticks in the documentation
+error: item in documentation missing backticks
```
xFrednet, line 1: Could you add `// run-rustfix` in the first line of this and other `.rs` test file? That'll create a `.fixed` file and validate that the suggestion has been applied correctly :upside_down_face:

clippy_lints/src/doc.rs
```diff
@@ -747,17 +747,20 @@ fn check_word(cx: &LateContext<'_>, word: &str, span: Span) {
...
     if has_underscore(word) || word.contains("::") || is_camel_case(word) {
-        span_lint(
+        span_lint_and_sugg(
             cx,
             DOC_MARKDOWN,
             span,
-            &format!("you should put `{}` between ticks in the documentation", word),
+            "item in documentation missing backticks",
+            "try",
+            format!("`{}`", snippet(cx, span, "..")),
```
xFrednet, line 762: Here it would be better to use `snippet_with_applicability` to ensure that the suggestion is only `MachineApplicable` if the snipped was extracted correctly :upside_down_face: 

[Here is an example usage of `snippet_with_applicability`](https://github.com/rust-lang/rust-clippy/blob/1f955158ddc2035a054fdd4d7cc3bd2297b5a7ba/clippy_lints/src/len_zero.rs#L420-L433)

clippy_lints/src/doc.rs
```diff
@@ -747,17 +747,20 @@ fn check_word(cx: &LateContext<'_>, word: &str, span: Span) {
...
     }
 
     if has_underscore(word) || word.contains("::") || is_camel_case(word) {
-        span_lint(
+        span_lint_and_sugg(
             cx,
             DOC_MARKDOWN,
             span,
-            &format!("you should put `{}` between ticks in the documentation", word),
+            "item in documentation missing backticks",
```
xFrednet, line 760: Here I would suggest adding an "is" that makes it sound better IMO :upside_down_face: 
```suggestion
            "item in documentation is missing backticks",
```

xFrednet (commented):
Thank you for the PR, the changes look good in general, just some minor NITs. :upside_down_face:

tests/ui/doc/doc.stderr
Serial-ATA: All suggestions are put in backticks, but they aren't actually included in the fix. I don't know how to make it a single pair and still include them.

clippy_lints/src/doc.rs
```diff
@@ -747,17 +747,22 @@ fn check_word(cx: &LateContext<'_>, word: &str, span: Span) {
...
+        let mut applicability = Applicability::MachineApplicable;
+
+        span_lint_and_sugg(
             cx,
             DOC_MARKDOWN,
             span,
-            &format!("you should put `{}` between ticks in the documentation", word),
+            "item in documentation is missing backticks",
+            "try",
+            format!("`{}`", snippet_with_applicability(cx, span, "..", &mut applicability)),
```
giraffate, line 764: There are needless backticks. This causes two backticks in suggestions.
```suggestion
            snippet_with_applicability(cx, span, "..", &mut applicability).into_owned(),
```

xFrednet: These backticks are actually intentinal to make the suggestion machine applicable, but the suggestion sadly does look a bit strange :) See [comment](https://github.com/rust-lang/rust-clippy/pull/7904#discussion_r739807872)

tests/ui/doc/missing_ticks.rs
```diff
@@ -0,0 +1,9 @@
+// run-rustfix
```
xFrednet, line 1: The usage of `// run-rustfix` is correct. However, could you add the comment to the `tests/ui/doc/doc.rs` file? It's usually better to have all test cases for a lint in one file, and I would like the suggestion to work for each suggestion and not just one. The rest looks good to me. :upside_down_face:

tests/ui/doc/doc.stderr
xFrednet: Makes sense! Thank you :+1:

clippy_lints/src/doc.rs
giraffate: Oh, I got it, thanks!

tests/ui/doc/doc.stderr
```diff
@@ -1,190 +1,166 @@
...
+   |                                                                                   ^^^^^^^^^^^^^ help: try: ``Foo::some_fun``
 
-error: you should put `a::global:path` between ticks in the documentation
-  --> $DIR/doc.rs:11:15
+error: item in documentation is missing backticks
+  --> $DIR/doc.rs:12:15
    |
 LL | /// Here be ::a::global:path.
-   |               ^^^^^^^^^^^^^^
+   |               ^^^^^^^^^^^^^^ help: try: ``a::global:path``
```
xFrednet, line 25: Side note: This lint case could be improved by including the first double colon for global identifiers. I've created a new issue #7914 to keep this in mind. This is not a result of your changes. Therefore, I would suggest not to fix it in this PR, also to keep the changes manageable :upside_down_face:

tests/ui/doc/missing_ticks.rs
xFrednet: I think this test file can be removed again now that `tests/ui/doc/doc.rs` has the fix comment and has more in depth tests

tests/ui/doc/issue_1832.rs
```diff
@@ -0,0 +1,9 @@
+/// Ok: <http://www.unicode.org/reports/tr9/#Reordering_Resolved_Levels>
```
xFrednet, line 1: Is there a reason why you extracted these two tests?

xFrednet (commented):
Alright, I found two more smaller things where it would be great if you could address them. Then I would merge it. Thank you for the quick changes :upside_down_face:

tests/ui/doc/issue_1832.rs
Serial-ATA: Those two tests fail with rustfix because they have a suggestion that isn't machine applicable, and it's compiled with `-Dwarnings`.

https://github.com/rust-lang/rust-clippy/blob/c68af43b3dffc90b4653063c5012f039cf209784/tests/compile-test.rs#L131-L132

xFrednet: That explains why some lint test files are split up into a fixable and not fixable test file. Today I learned. Would you mind renaming the `doc.rs` file to `doc-fixable.rs`? You can also do that in an additional commit ^^ then everything is really ready :upside_down_face:

xFrednet (approved):

xFrednet:
Fantastic! Thank you very much for addressing my sometimes nit picky comments. I appreciate it! :upside_down_face: 

This looks really good to me now :+1: 

@bors r+

PR #7907, opened 2021-10-31
Move non_ascii_literal to restriction

Alexendoo:
It feels like the more apt category, since cases where you'd want it enabled would be pretty specific

changelog: Move [`non_ascii_literal`] to `restriction`


flip1995:
Seems reasonable to me. 

@bors r+

Thanks!

PR #7909, opened 2021-11-01
Fix false negative in [`match_overlapping_arms`]

mikerite:
changelog: Fix false negative in [`match_overlapping_arms`]


camsteffen:
Thanks!

@bors r+

PR #7916, opened 2021-11-02
Advise to put a :: prefix inside the ticks

### LINKED ISSUE #7914: Enhance missing backtick detection for global identifiers `doc_markdown`
The lint [`clippy::doc_markdown`](https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown) can detect missing backticks around identifiers. Global identifiers like `::a::global::item` are only detected after the first double colon `::`. It would be good if the suggestion also included the initial `::` for a global identifier.

**Test File**

https://github.com/rust-lang/rust-clippy/blob/f37dc2521b8b8cdb27dd36e364b8e838a5aa51b6/tests/ui/doc/doc.rs#L12

**Lint output**

https://github.com/rust-lang/rust-clippy/blob/f37dc2521b8b8cdb27dd36e364b8e838a5aa51b6/tests/ui/doc/doc.stderr#L21-L25

It should instead span over the complete `::a::global:path` and suggest `` `::a::global:path` ``

---

@rustbot label +C-enhancement

samueltardieu:
Fixes #7914.

changelog: Keep an initial `::` when [`doc_markdown`] suggests to use ticks

### DIFF
--- a/clippy_lints/src/doc.rs
+++ b/clippy_lints/src/doc.rs
@@ -689,10 +689,18 @@ fn check_text(cx: &LateContext<'_>, valid_idents: &FxHashSet<String>, text: &str
     for word in text.split(|c: char| c.is_whitespace() || c == '\'') {
         // Trim punctuation as in `some comment (see foo::bar).`
         //                                                   ^^
-        // Or even as in `_foo bar_` which is emphasized.
-        let word = word.trim_matches(|c: char| !c.is_alphanumeric());
+        // Or even as in `_foo bar_` which is emphasized. Also preserve `::` as a prefix/suffix.
+        let mut word = word.trim_matches(|c: char| !c.is_alphanumeric() && c != ':');
 
-        if valid_idents.contains(word) {
+        // Remove leading or trailing single `:` which may be part of a sentence.
+        if word.starts_with(':') && !word.starts_with("::") {
+            word = word.trim_start_matches(':');
+        }
+        if word.ends_with(':') && !word.ends_with("::") {
+            word = word.trim_end_matches(':');
+        }
+
+        if valid_idents.contains(word) || word.chars().all(|c| c == ':') {
             continue;
         }
 
--- a/tests/ui/doc/doc.rs
+++ b/tests/ui/doc/doc.rs
@@ -8,7 +8,9 @@
 /// The foo_bar function does _nothing_. See also foo::bar. (note the dot there)
 /// Markdown is _weird_. I mean _really weird_. This \_ is ok. So is `_`. But not Foo::some_fun
 /// which should be reported only once despite being __doubly bad__.
-/// Here be ::a::global:path.
+/// Here be ::a::global:path, and _::another::global::path_.  :: is not a path though.
+/// Import an item from ::awesome::global::blob:: (Intended postfix)
+/// These are the options for ::Cat: (Intended trailing single colon, shouldn't be linted)
 /// That's not code ~NotInCodeBlock~.
 /// be_sure_we_got_to_the_end_of_it
 fn foo_bar() {
--- a/tests/ui/doc/doc.stderr
+++ b/tests/ui/doc/doc.stderr
@@ -18,173 +18,191 @@ error: you should put `Foo::some_fun` between ticks in the documentation
 LL | /// Markdown is _weird_. I mean _really weird_. This /_ is ok. So is `_`. But not Foo::some_fun
    |                                                                                   ^^^^^^^^^^^^^
 
-error: you should put `a::global:path` between ticks in the documentation
-  --> $DIR/doc.rs:11:15
+error: you should put `::a::global:path` between ticks in the documentation
+  --> $DIR/doc.rs:11:13
    |
-LL | /// Here be ::a::global:path.
-   |               ^^^^^^^^^^^^^^
+LL | /// Here be ::a::global:path, and _::another::global::path_.  :: is not a path though.
+   |             ^^^^^^^^^^^^^^^^
+
+error: you should put `::another::global::path` between ticks in the documentation
+  --> $DIR/doc.rs:11:36
+   |
+LL | /// Here be ::a::global:path, and _::another::global::path_.  :: is not a path though.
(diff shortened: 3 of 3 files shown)


samueltardieu:
Asking for review by @xFrednet as they opened the original issue.

r? @xFrednet

clippy_lints/src/doc.rs
```diff
@@ -690,7 +690,16 @@ fn check_text(cx: &LateContext<'_>, valid_idents: &FxHashSet<String>, text: &str
...
+        let word = {
+            let w = word.trim_matches(|c: char| !c.is_alphanumeric());
+            let offset = w.as_ptr() as usize - word.as_ptr() as usize;
+            // Preserve an initial `::` prefix, even as in `_::a::b_` which is emphasized.
+            if offset >= 2 && &word.as_bytes()[offset - 2..offset] == b"::" {
+                &word[offset - 2..offset + w.len()]
+            } else {
+                w
+            }
+        };
```
xFrednet, line 702: Using pointers casted to `usize` seems very error prune to me, even if I know exactly what this is doing. The usage of since indexing on the string seems a bit complicated and has caused some errors in the past due to byte and string index mismatches. Could this entire block maybe be replaced by changing the `trim_matches` statement like this?

```suggestion
        let word = word.trim_matches(|c: char| !c.is_alphanumeric() && c != ':');
```

That would allow for identifiers like `:::::::way_to_many_colons::` but that would the current as well, if I'm seeing this correctly. (It's also debatable if this would be an FP)

tests/ui/doc/doc.rs
```diff
@@ -8,7 +8,7 @@
 /// The foo_bar function does _nothing_. See also foo::bar. (note the dot there)
 /// Markdown is _weird_. I mean _really weird_. This \_ is ok. So is `_`. But not Foo::some_fun
 /// which should be reported only once despite being __doubly bad__.
-/// Here be ::a::global:path.
+/// Here be ::a::global:path and _::another::global::path_.
```
xFrednet, line 11: Could you also add a test where you just have `::` two single colons like this in the text? I think such cases shouldn't be linted, but that's not a clear line. What do you think? :upside_down_face:

xFrednet (commented):
That was a quick PR :+1:. The output is already looking good, but I'm a bit unsure about the documentation. I left comments on the related code segments :upside_down_face:

clippy_lints/src/doc.rs
samueltardieu: I wanted to avoid those too many colons. But I agree with you that this leads to convoluted code, I don't like this pointer arithmetic much myself. Maybe I should use different trims for the start and the end (this won't take more time) so that at least we don't have trailing `:`. I'll see if removing `:` as long as the resulting string starts with `:::` (so that we keep only two colons) leads to clean code.

tests/ui/doc/doc.rs
samueltardieu: Agreed, I'll add that.

clippy_lints/src/doc.rs
xFrednet: The trailing `:` is a good point. The code could either use `trim_start_matches` and `trim_end_matches` to only allow the prefix or validate that the word at least ends with two colons. I think the later option would make more sense. Then we can still say:

```rs
/// Import an item from ::awesome::global::blob:: (Intended postfix)
/// These are the options for ::Cat: (Indended trailing single colon, shouln't be linted)
```

samueltardieu: I was testing that `and:::std::mem::forget` was correctly parsed as `::std::mem::forget` but it is indeed parsed as `and:::std::mem::forget` since we don't check that we don't have more than two `::` in a row. So right now I won't remove extra `:` at the beginning, but maybe if we add a check that we don't have `:::` sequences to determine if a word is valid it would be worth filtering extra `:` at the beginning of the word.

xFrednet: I also think that it's fine if such cases are linted. If someone writes and:::std::mem::forget it should be pointed out :upside_down_face:

xFrednet: And as you said, this was also previously accepted, and we didn't receive any complaints about that as far as I know :upside_down_face:

samueltardieu:
To summarize, leading and trailing `:` are now included in the lint as long as there are at least two of them in a row. Single leading or trailing `:` are not included. Is that better?

samueltardieu:
(pushed a new version with a clearer comment)

tests/ui/doc/doc.stderr
```diff
@@ -18,173 +18,197 @@ error: you should put `Foo::some_fun` between ticks in the documentation
...
+  --> $DIR/doc.rs:11:36
+   |
+LL | /// Here be ::a::global:path, and _::another::global::path_.  :: is not a path though.
+   |                                    ^^^^^^^^^^^^^^^^^^^^^^^
+
+error: you should put `::` between ticks in the documentation
+  --> $DIR/doc.rs:11:63
+   |
+LL | /// Here be ::a::global:path, and _::another::global::path_.  :: is not a path though.
+   |                                                               ^^
```
xFrednet, line 37: The single `::` is now linted as a path. Do we want to keep it this way? In that case, it would be better to update the doc comment :upside_down_face:

xFrednet (commented):
One last tiny NIT, then it should be good to go, IMO :upside_down_face:

tests/ui/doc/doc.stderr
samueltardieu: Darn, you're perfectly right to nitpick here, I had missed that in the last update! I'll fix it, it makes no sense to lint it IMO.

xFrednet: Good, I sometimes feel like I'm to nitpicky :joy:

samueltardieu: Updated, thanks for the catch!

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update :+1: 

@bors r+

PR #7917, opened 2021-11-02
Add `cargo dev lint` to manually run clippy on a file

Alexendoo:
I found the manual run command really useful, this makes it a bit easier to type

Not sure if this belongs in the changelog or not

changelog: Add `cargo dev lint` to manually run clippy on a file

### DIFF
--- a/clippy_dev/src/main.rs
+++ b/clippy_dev/src/main.rs
@@ -3,7 +3,7 @@
 #![warn(rust_2018_idioms, unused_lifetimes)]
 
 use clap::{App, AppSettings, Arg, ArgMatches, SubCommand};
-use clippy_dev::{bless, fmt, new_lint, serve, setup, update_lints};
+use clippy_dev::{bless, fmt, lint, new_lint, serve, setup, update_lints};
 fn main() {
     let matches = get_clap_config();
 
@@ -54,6 +54,10 @@ fn main() {
             let port = matches.value_of("port").unwrap().parse().unwrap();
             let lint = matches.value_of("lint");
             serve::run(port, lint);
+        },
+        ("lint", Some(matches)) => {
+            let filename = matches.value_of("filename").unwrap();
+            lint::run(filename);
         },
         _ => {},
     }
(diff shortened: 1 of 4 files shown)


clippy_dev/src/main.rs
```diff
@@ -219,5 +223,15 @@ fn get_clap_config<'a>() -> ArgMatches<'a> {
                 )
                 .arg(Arg::with_name("lint").help("Which lint's page to load initially (optional)")),
         )
+        .subcommand(
+            SubCommand::with_name("lint")
+                .about("Manually run clippy on a file")
+                .arg(
+                    Arg::with_name("INPUT")
+                        .multiple(true)
```
giraffate, line 231: I don't know if I'm doing it right, but I tried multiple inputs and got an error like below:
```sh
cargo dev lint tests/ui/absurd-extreme-comparisons.rs tests/ui/bind_instead_of_map.rs    
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
     Running `target/debug/clippy_dev lint tests/ui/absurd-extreme-comparisons.rs tests/ui/bind_instead_of_map.rs`
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/clippy-driver -L ./target/debug -Z no-codegen --edition 2021 tests/ui/absurd-extreme-comparisons.rs tests/ui/bind_instead_of_map.rs`
error: multiple input filenames provided (first two filenames are `tests/ui/absurd-extreme-comparisons.rs` and `tests/ui/bind_instead_of_map.rs`)
```

Is `multiple` needed?

Alexendoo: It does only accept one filename, I put the `multiple` in so that you can pass through extra flags like `cargo dev lint a.rs -- --crate-type lib`. I'm not familiar with clap though so that might not be the right way to do that

I could remove it, the cases where you'd need extra flags seem pretty rare

giraffate: Hm, can you remove it? If there's a case for it, we can add it then.

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #7918, opened 2021-11-02
Use .cargo/config.toml instead of .cargo/config

samueltardieu:
`.cargo/config.toml` is the preferred form for the local configuration file. This is emphasized in _The Cargo Book_ with the following note:

> Note: Cargo also reads config files without the `.toml` extension, such as `.cargo/config`. Support for the `.toml` extension was added in version 1.39 and is the preferred form.

Moreover, this helps with toml-aware text editors as they will recognize the file extension.

changelog: none


xFrednet:
This seems like a strait forward change. I'm not sure if the rust repo depends on this file somehow. Therefore, I would rather pass this along to someone else :)

r? @flip1995

flip1995:
`x.py` in the Rust repo just ignores this file altogether AFAIK, so we should be good.

@bors r+

flip1995:
> so we should be good.

Famous last words. :smile:

camsteffen:
We could do the same thing for `rust-toolchain`.

PR #7930, opened 2021-11-04
new lint: `needless_deref`

lengyijun:
changelog: ``[`needless_deref`]``

Related pr: #6837 #7577
@Jarcho Could you please give a review?

`cargo lintcheck` gives no false negative (but tested crates are out-of-date).

TODO:
1. Not sure the name. `deref_on_immutable_ref` or some others?
2. I'll rebase before merge


clippy_lints/src/needless_deref.rs
```diff
@@ -0,0 +1,84 @@
...
+    ///
+    /// ### Why is this bad?
+    /// When people deref on an immutable reference `&T`, they may expect return `&U`.
+    /// Accutually `&* (&T)` is still `&T`.
+    /// if you want to deref explicitly, `&** (&T)` is what you need.
+    /// If you want to reborrow, `&T` is enough (`&T` is Copy).
+    ///
+    /// ### Known problems
+    /// None.
+    ///
```
flip1995, line 21: Not required, if there aren't any known problems.

clippy_lints/src/needless_deref.rs
```diff
@@ -0,0 +1,84 @@
...
+                }
+
+                span_lint_and_help(
+                    cx,
+                    NEEDLESS_DEREF,
+                    e.span,
+                    "deref on an immutable borrow",
+                    None,
+                    &format!(
+                        "consider using `{}` if you would like to deref\nconsider using `{}` if you would like to reborrow",
```
flip1995, line 76: Just add multiple help messages with `span_lint_and_then` instead of trying to format with `\n`.

tests/ui/needless_deref.rs
```diff
@@ -0,0 +1,26 @@
...
+}
+
+mod should_not_lint {
+    fn foo(x: &mut &u32) {
+        *x = &1;
+    }
+
+    fn main() {
+        let mut x = &0;
+        foo(&mut &*x); // should not lint
```
flip1995, line 20: What's wrong with removing `&*` here?

clippy_lints/src/needless_deref.rs
```diff
@@ -0,0 +1,84 @@
...
+                    if span.from_expansion() {
+                        return;
+                    }
+                    let parent_node = map.find(parent_hir_id);
+                    if let Some(rustc_hir::Node::Expr(parent_expr)) = parent_node {
+                        if  matches!(parent_expr.kind, ExprKind::Unary(UnOp::Deref, ..)) ||
+                            matches!(parent_expr.kind, ExprKind::AddrOf(..) ){
+                                return;
+                            }
+                    }
```
flip1995, line 66: Why is this necessary?

flip1995 (commented):
Some general comments. I wasn't involved in the previous discussions about this lint, so I don't really know how we wanted to move forward with this lint.

camsteffen:
This lint is very similar to [deref_addrof](https://rust-lang.github.io/rust-clippy/master/index.html#deref_addrof) but with the operators swapped. How about the name `addrof_deref_ref` - `&*<ref-typed-expr>`.

camsteffen:
Though we probably should say "borrow" instead of "addrof" since that's what the [docs](https://doc.rust-lang.org/reference/expressions/operator-expr.html#borrow-operators) say - so maybe `borrow_deref_ref`.

tests/ui/needless_deref.rs
lengyijun:
```
let mut x = &0;
foo(&mut &*x); 
// foo(&mut x); // assert will fail
// foo(&mut &**x); // compile fail
assert_eq!(*x, 0);
```
So `foo(&mut &*x);` is the only correct way.
Whenever `&*(&T)` is wrapped with `&` or `&mut`, the lint should ignore it.

clippy_lints/src/needless_deref.rs
lengyijun: `AddrOf` is explained by previous example

lengyijun: ## `Unary` :
`* &* (&T)`  will trigger `deref_addrof`.
I think we should deal with `deref_addrof` first.

tests/ui/needless_deref.rs
camsteffen: Very interesting.

> Whenever `&*(&T)` is wrapped with `&` or `&mut`, the lint should ignore it.

Wouldn't it only be when wrapped with `&mut`? An immutable borrow is no risk since it's immutable I think.

Could you also add that the `&T` must be a `ExprKind::Path` where the base is a mutable binding? For example `&mut &*fun()` should be lintable.

clippy_lints/src/needless_deref.rs
camsteffen: I'd just lint twice. Noisy code = noisy clippy.

tests/ui/needless_deref.rs
lengyijun: ## Why the lint excludes `& &* (&T)` case
```
fn main() {
    let x = &12;
    let addr_x = &x as *const _ as usize;
    let addr_y= &&*x as *const _ as usize; // assert ok
    // let addr_y = &x as *const _ as usize; // assert fail                                                                                       
    assert_ne!(addr_x, addr_y);
}
```

lengyijun: > Could you also add that the &T must be a ExprKind::Path where the base is a mutable binding? For example &mut &*fun() should be lintable.

Sure, I have done it.

camsteffen: I think if we suppressed lints whenever it would change the pointer equality of two expressions, that would be too limiting.

Jarcho: I'll second @camsteffen on pointer equality here. It's very unlikely to be used for that purpose. I might consider allowing `let x = &*y` as an explicit reborrow.

Jarcho: Don't forget to handle `&mut &*x.y`.

clippy_lints/src/needless_deref.rs
Jarcho: The suggestion for both lints would be the same. `deref_addrof` will suggest removing the first `*&`, this lint will suggest removing the `&*` part. Both of those suggestions will result in `*&T`. Since it's easy to detect I would say it's reasonable to silence one of them.

A check if `deref_addrof` is allowed on the outer deref is needed though. Use `is_lint_allowed`.

tests/ui/needless_deref.rs
lengyijun: > I might consider allowing let x = &*y as an explicit reborrow.

If any others think so? Or I'll retract this pr.

camsteffen: > > I might consider allowing let x = &*y as an explicit reborrow.
> 
> If any others think so? Or I'll retract this pr.

This seems like a misunderstanding. @Jarcho is suggesting to allow a `let` statement that that doesn't do anything except "re-borrow". That doesn't apply to most instances of this lint.

lengyijun: @camsteffen @Jarcho No longer lint on reborrow now.
Frankly speaking, my intuitive is to lint on reborrow.

clippy_lints/src/needless_deref.rs
lengyijun: > A check if deref_addrof is allowed on the outer deref is needed though. Use is_lint_allowed.

Done.

tests/ui/needless_deref.rs
Jarcho: I just realized I had a typo earlier. I meant to say consider allowing the lint on `let x = &&*y`. This would be used for when a different pointer value is actually needed. I'm not sure if there's ever a need for it, but it is a semantic change and this gives the person a more explicit way to do that without having to allow the lint.

As for `let x = &*y` I had erroneously mentioned earlier, I think it only matters if `y` is a mutable reference, as then it's changing the mutability of the pointer which shouldn't be linted.

lengyijun: > As for let x = &*y I had erroneously mentioned earlier, I think it only matters if y is a mutable reference, as then it's changing the mutability of the reference which shouldn't be linted.

This pr doesn't deal with `&mut T`.

lengyijun: @Jarcho Thanks. I guess we have the same view now!

@camsteffen What do you think of `let x = &&*y` ? I guess Jarcho and I support not linting on it.
In `cargo linkcheck`, I didn't find such pattern.

camsteffen: The `let` cases seem a bit speculative to me. Personally I would "keep it simple" until a real FP is found.

lengyijun:
## Conclusion and current status:
We lint on:
1. `&&* (&T)`
2. `let x=&* (&T)`

We don't lint on:
1. `&mut &* (&T)`

flip1995:
Here's another interesting case, I stumbled upon working on rustc today: [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=de166f63ae387db7628ff03f466b9c08)

```rust
struct S<'a> {
    a: &'a mut u32,
}

fn f(s: &S<'_>) {
    let x = &*s.a;
    // let x = s.a; // <- errors, since &mut must be moved and cannot be copied
    println!("{:?}", x);
}
```

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+            if let ExprKind::Unary(UnOp::Deref, deref_expr) = addrof_expr.kind;
+            if !matches!(deref_expr.kind, ExprKind::Unary(UnOp::Deref, ..) );
+            let ref_ty = cx.typeck_results().expr_ty(deref_expr);
+            if let ty::Ref(_, inner_ty, Mutability::Not) = ref_ty.kind();
+            then{
+
+                let map = cx.tcx.hir();
+                if let Some(parent_hir_id) = map.find_parent_node(e.hir_id){
+                    let span = map.span(parent_hir_id);
+                    if span.from_expansion() {
```
Jarcho, line 67: You don't need to check the span here. This can only happen when the address-of deref expression is an argument to a macro and that argument is then expanded within an expression created by the macro. Whether or not the expression is a macro argument doesn't change whether you want to lint or not. Only whether a mutable reference is taken will do that.

Removing this check here will require you to add it to the check for `DEREF_ADDROF`

Jarcho (changes requested):

clippy_lints/src/borrow_deref_ref.rs
lengyijun: You are right. But I  don't know how to deal with this example, without checking parent:
```
// no other lints (deref_addrof) triggered.
fn main() {
    let x=&&12;

    println!("{}",*x); // will lint on
    println!("{}",x);  // remove * is ok

    println!("{:p}",*x); // will lint on
    println!("{:p}",x);  // remove *, the result is different !
}
```

Jarcho: Really neither of those cases should be linted as the borrow occurs inside a macro expansion. This should have been fixed by the earlier change to make sure neither the deref nor the borrow expressions were from a macro expansion.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+declare_lint_pass!(BorrowDerefRef => [BORROW_DEREF_REF]);
+
+impl LateLintPass<'_> for BorrowDerefRef {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx rustc_hir::Expr<'_>) {
+        if_chain! {
+            if !e.span.from_expansion();
+            if let ExprKind::AddrOf(_, Mutability::Not, addrof_expr) = e.kind;
+            if let ExprKind::Unary(UnOp::Deref, deref_expr) = addrof_expr.kind;
```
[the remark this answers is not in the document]
lengyijun: I'm sorry I didn't get your meaning. 
I try in 082ae7bca3372d22dc985004331a33b00de950c6 , but it still doesn't work.

clippy_lints/src/borrow_deref_ref.rs
lengyijun: I'm sorry I didn't get your meaning. 
I try in 082ae7bca3372d22dc985004331a33b00de950c6 , but it still doesn't work.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// let a: &String = &* s;
+    /// foo(&*s);
+    ///
+    /// // Good
+    /// let a: &String = s;
+    /// foo(&**s);
+    ///
+    /// fn foo(_: &str){ }
+    /// ```
+    pub BORROW_DEREF_REF,
```
xFrednet, line 48: The PR #7813 was just merged, which added a new `clippy::version` attribute to all lints. Just a heads-up that the attribute needs to added for a green CI after the next rebase :upside_down_face: 
```suggestion
    #[clippy::version = "1.58.0"]
    pub BORROW_DEREF_REF,
```

clippy_lints/src/borrow_deref_ref.rs
Jarcho: Looks like `format_args` is broken here. That macro borrows every argument, but the span it gives to that borrow is the same as the argument's span, rather than a span indicating it's from a macro. This should probably be changed on rustc's side.

camsteffen: lo and behold I'm already trying to do that https://github.com/rust-lang/rust/pull/90131

lengyijun:
I guess it's ready to another review.
I' ll rebase it before merge.

lengyijun:
I have rebased.
I guess it's ready to another review.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,113 @@
...
+            let ref_ty = cx.typeck_results().expr_ty(deref_target);
+            if let ty::Ref(_, inner_ty, Mutability::Not) = ref_ty.kind();
+            then{
+
+                if let Some(parent_expr) = get_parent_expr(cx, e){
+                    if matches!(deref_target.kind, ExprKind::Path(..) | ExprKind::Field(..)) {
+                        if matches!(parent_expr.kind, ExprKind::AddrOf(_, Mutability::Mut, _)) {
+                            return;
+                        }
+                        if matches!(parent_expr.kind, ExprKind::Unary(UnOp::Deref, ..)) &&
```
Jarcho, line 72: Formatting nit

```rust
if matches!(parent_expr.kind, ExprKind::Unary(UnOp::Deref, ..))
    && !is_lint_allowed(cx, DEREF_ADDROF, parent_expr.hir_id)
{
    return;
}
```

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,113 @@
...
+                                "consider using `{}` if you would like to reborrow",
+                                &snippet_opt(cx, deref_target.span).unwrap(),
+                             )
+                        );
+
+                        // has deref trait -> give 2 help
+                        // doesn't have deref trait -> give 1 help
+                        if let Some(deref_trait_id) = cx.tcx.lang_items().deref_trait(){
+                            if !implements_trait(cx, inner_ty, deref_trait_id, &[]) {
+                                return;
```
Jarcho, line 96: The help message can just be put directly here rather than using an early return.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                }
+
+                span_lint_and_then(
+                    cx,
+                    BORROW_DEREF_REF,
+                    e.span,
+                    "deref on an immutable reference",
+                    |diag| {
+                        diag.help(
```
Jarcho, line 85: This should be a suggestion in the case where the type doesn't implement `Deref`.

Jarcho (changes requested):
Sorry for the late review. December's been very busy.

Both help messages should be suggestions instead.

lengyijun:
WIP
I will finish it soon

lengyijun:
>  Both help messages should be suggestions instead.
Suggestion looks much better!

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,118 @@
...
+            let ref_ty = cx.typeck_results().expr_ty(deref_target);
+            if let ty::Ref(_, inner_ty, Mutability::Not) = ref_ty.kind();
+            then{
+
+                if let Some(parent_expr) = get_parent_expr(cx, e){
+                    if matches!(deref_target.kind, ExprKind::Path(..) | ExprKind::Field(..)) {
+                        if matches!(parent_expr.kind, ExprKind::AddrOf(_, Mutability::Mut, _)) {
+                            return;
+                        }
+                        if matches!(parent_expr.kind, ExprKind::Unary(UnOp::Deref, ..))
```
Jarcho, line 73: If I understand that correctly this should be moved into the outer if block. I assume this is to not lint on `*&*x` when `DEREF_ADDROF` also does.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,118 @@
...
+            if let ExprKind::Unary(UnOp::Deref, deref_target) = addrof_target.kind;
+            if !deref_target.span.from_expansion();
+            if !matches!(deref_target.kind, ExprKind::Unary(UnOp::Deref, ..) );
+            let ref_ty = cx.typeck_results().expr_ty(deref_target);
+            if let ty::Ref(_, inner_ty, Mutability::Not) = ref_ty.kind();
+            then{
+
+                if let Some(parent_expr) = get_parent_expr(cx, e){
+                    if matches!(deref_target.kind, ExprKind::Path(..) | ExprKind::Field(..)) {
+                        if matches!(parent_expr.kind, ExprKind::AddrOf(_, Mutability::Mut, _)) {
```
Jarcho, line 70: This needs a check for auto-borrow as well. Something like `cx.typeck_results().expr_adjustments(e).first().map_or(false, |a| matches!(a.kind, Adjust::Borrow(AutoBorrow::Ref(_, AutoRefMutability::Mut {..}))))`

Jarcho:
I think just those last two points and that's it. @Manishearth anything else?

clippy_lints/src/borrow_deref_ref.rs
lengyijun: > I assume this is to not lint on *&*x when DEREF_ADDROF also does.

Right

clippy_lints/src/borrow_deref_ref.rs
lengyijun: I guess `if !e.span.from_expansion();` excludes this case. (I'm not sure)

Jarcho: No. Auto-borrow comes from the receiver in a method call. e.g. `(&*x).foo()` can borrow `&*x`

camsteffen:
oh no I did it again 😆

Manishearth:
I mean given that Jarcho has already reviewed it this is probably a rubber stamp, but I just wanted to be sure your concerns were handled too :smile:

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,127 @@
...
+                span_lint_and_then(
+                    cx,
+                    BORROW_DEREF_REF,
+                    e.span,
+                    "deref on an immutable reference",
+                    |diag| {
+                        diag.span_suggestion(
+                            e.span,
+                            "if you want to reborrow, use",
+                            snippet_opt(cx, deref_target.span).unwrap(),
```
camsteffen, line 98: This can fail and panic. Either omit the suggestion, use `snippet(.., "..")`, or avoid getting a snippet by suggesting to replace `"&*"` with `""`.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,127 @@
...
+                            Applicability::MachineApplicable
+                        );
+
+                        // has deref trait -> give 2 suggestions
+                        // doesn't have deref trait -> give 1 suggestion
+                        match cx
+                            .tcx
+                            .lang_items()
+                            .deref_trait()
+                            .map(|deref_trait_id| implements_trait(cx, inner_ty, deref_trait_id, &[]) )
```
camsteffen, line 108: Should not suggest by default

```suggestion
                            .map_or(false, |deref_trait_id| implements_trait(cx, inner_ty, deref_trait_id, &[]) )
```

clippy_lints/src/borrow_deref_ref.rs
lengyijun: I'm not sure I understand your meaning correctly.
Do you mean an `&mut` is auto-adjusted to `&*_` , and output `&mut &* _` ?

Jarcho: When calling a method taking `&self` or `&mut self` the compiler will insert a borrow to make the type match if needed. For references this can only happen when a trait is implemented on a method (e.g. `io::Read` is implemented on `&[u8]).
A read example would be `(&*bytes).read(count)` which is intended to leave `bytes` unmodified after calling `read`.

camsteffen: Here's a minimal example (can be a test). The `&*` cannot be removed because it would make the auto borrow not possible.

```rust
let src = &[0][..];
(&*src).read_to_end(&mut vec![]);
```

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,140 @@
...
+    ///
+    /// // Bad
+    /// let a: &String = &* s;
+    /// foo(&*s);
+    ///
+    /// // Good
+    /// let a: &String = s;
+    /// foo(&**s);
+    ///
+    /// fn foo(_: &str){ }
```
camsteffen, line 46: nit
```suggestion
    /// fn foo(_: &str) {}
```

tests/ui/needless_deref.rs
camsteffen: Is everything resolved here?

clippy_lints/src/borrow_deref_ref.rs
camsteffen: Looks like this is fixed now.

tests/ui/borrow_deref_ref.rs
```diff
@@ -0,0 +1,76 @@
+fn main() {}
```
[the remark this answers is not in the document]
lengyijun: If gives 2 suggestions, `run-rustfix` fails.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,140 @@
...
+
+                    // `&mut e` should trigger when `e` looks like :
+                    // 1. foo()
+                    //
+                    // `&mut e` should not trigger when `e` looks like :
+                    // 1. variable: `x`
+                    // 2. field: `x.a`
+                    //
+                    if matches!(parent_expr.kind, ExprKind::AddrOf(_, Mutability::Mut, _))
+                        && matches!(deref_target.kind, ExprKind::Path(..) | ExprKind::Field(..)) {
```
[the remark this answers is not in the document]
lengyijun: Maybe if there is another way to visit `x`, `x.a`, `x[0]`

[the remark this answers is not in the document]
lengyijun: `Unary` should also be excluded.
But it's already excluded in `if_chain` yet.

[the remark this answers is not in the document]
lengyijun: > What is the rationale for that?

1. If the reference is fresh. 
Call, MethodCall and Block always return a fresh reference.
`x`, `x.a`, `x[0]` : we can recreate it again.
  
2. If we remove `&*`, the mutability of `x` changed.
|            |                 |
| ---------- | --------------- |
| &mut &*x   | &mut x          |
| x : &usize | x :  &mut usize |

|              |              |
| ------------ | ------------ |
| &mut &*id(x) | &mut id(x)   |
| x : &usize   | x :  & usize |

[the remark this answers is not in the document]
lengyijun: 7930 drawio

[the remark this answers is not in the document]
camsteffen: Sorry I'm not following. What would help is a minimal code example where you can say "without that check, the lint would incorrectly fire here".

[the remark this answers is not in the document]
Jarcho: Just going to jump in here as a change needs to happen. This is checking for a temporary vs named value. Re-borrowing is sometimes used to change a named value to a temporary value (e.g. if x is `&[u8]` then `(&*x).read(..)` avoids changing `x` as `read` takes `&mut &[u8]`).

Both `ExprKind::Index` and `ExprKind::Unary(UnOp::Deref, ..)` can also result in a named value, so those need to be handled as well.

camsteffen:
@lengyijun Please address open conversations to move forward. If you're not sure what to do about some of them let me know. Also feel free to squash history at any time to avoid difficult rebasing.

xFrednet:
Hey @lengyijun, this is a ping from triage, do you plan to continue working on this PR?

tests/ui/borrow_deref_ref.rs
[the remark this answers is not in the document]
Jarcho: You can split the tests into two files. One which can be fixed, and one which can't be. There are a couple of lints like that.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,118 @@
...
+                if let Some(parent_expr) = get_parent_expr(cx, e){
+                    // modification to `&mut &*x` is different from `&mut x`
+                    if matches!(deref_target.kind, ExprKind::Path(..)
+                                                | ExprKind::Field(..)
+                                                | ExprKind::Index(..)
+                                                | ExprKind::Unary(UnOp::Deref, ..)) {
+                        if matches!(parent_expr.kind, ExprKind::AddrOf(_, Mutability::Mut, _)) {
+                            return;
+                        }
+                        if matches!(parent_expr.kind, ExprKind::Unary(UnOp::Deref, ..)) &&
```
Jarcho, line 74: Just a question. Should this not happen no matter what kind of expression `deref_target` is? `deref_addrof` doesn't check that when linting.

xFrednet:
Hey @Jarcho, could you leave a normal comment (not a review) on this PR, so I can assign it to you?

Jarcho:
Can do.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -0,0 +1,118 @@
...
+                        }
+
+                        diag.span_suggestion(
+                            e.span,
+                            "if you would like to deref, try using `&**`",
+                            format!(
+                                "&**{}",
+                                &snippet_opt(cx, deref_target.span).unwrap(),
+                             ),
+                            Applicability::MachineApplicable
```
Jarcho, line 109: This should be `Applicability::MaybeIncorrect`. It's a semantic change the person may not want.

Manishearth:
@bors delegate=Jarcho

but yeah sure!

Jarcho:
Everything looks good. Just have the changes to `rustc_tools_util/LICENSE-APACHE` and `rustc_tools_util/LICENSE-MIT` to revert. Feel free to squash while you're at it.

lengyijun:
Run `cargo test --features deny-warnings,internal` locally emit no error,
but CI give some errors.

xFrednet:
On master, we've added a new test, that ensures that lints with automatically applicable suggestions also have a `.fixed` file. It looks like the CI failed due to this. It might be good to rebase on master again :)

Jarcho:
You're going to need to split the test file into two: `borrow_deref_ref.rs` and `borrow_deref_ref_unfixable.rs`. Move everything that produces two suggestions over to the unfixable one and then add an exemption to `RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS` in `compile-test.rs` for the unfixable test.

Jarcho:
Thanks for keeping up with everything!

@bors r+

PR #7938, opened 2021-11-05
Introduce `expr_visitor` and `expr_visitor_no_bodies`

camsteffen:
changelog: none

A couple utils that satisfy a *lot* of visitor use cases. Factoring in every possible usage would be really big so I just focused on cleaning clippy_utils.

### DIFF
--- a/clippy_utils/src/usage.rs
+++ b/clippy_utils/src/usage.rs
@@ -1,7 +1,7 @@
 use crate as utils;
+use crate::visitors::{expr_visitor, expr_visitor_no_bodies};
 use rustc_hir as hir;
-use rustc_hir::intravisit;
-use rustc_hir::intravisit::{NestedVisitorMap, Visitor};
+use rustc_hir::intravisit::{self, Visitor};
 use rustc_hir::HirIdSet;
 use rustc_hir::{Expr, ExprKind, HirId};
 use rustc_infer::infer::TyCtxtInferExt;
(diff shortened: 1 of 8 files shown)


clippy_utils/src/usage.rs
```diff
@@ -148,96 +148,47 @@ impl<'a, 'tcx> intravisit::Visitor<'tcx> for BindingUsageFinder<'a, 'tcx> {
...
-
-    fn visit_expr(&mut self, ex: &'tcx Expr<'tcx>) {
-        if self.seen_return_break_continue {
-            // No need to look farther if we've already seen one of them
-            return;
+pub fn contains_return_break_continue_macro(expression: &Expr<'_>) -> bool {
+    let mut seen_return_break_continue = false;
+    expr_visitor_no_bodies(|ex| {
+        if seen_return_break_continue {
+            return false;
```
xFrednet, line 155: This is the third time this pattern showed up, do you think it might be worth it to add a `expr_find_visitor_no_bodies` that does exactly this? An alternative could be a function like `contains_expr` similar to how `clippy_utils::visitors::contains_try` now works, just more generic.

This is not necessary but might be another nice wrapper for this pattern :upside_down_face:

xFrednet (approved):
LGTM, thanks for the changes :upside_down_face:.

I have one question regarding the usage. You can `r=me` if you prefer to keep the current version as it is :upside_down_face:.

clippy_utils/src/usage.rs
camsteffen: Yeah maybe we can add that some time. Just don't want to add too many wrappers since they could multiply quickly.

xFrednet: Alright :+1:

Issue #7922, opened 2021-11-03
deprecated_cfg_attr hit in MSRV 1.29.0


Kixunil:
Lint name: deprecated_cfg_attr

The suggested code doesn't work in the configured MSRV.

I tried this code:

```rust
#[cfg_attr(rustfmt, rustfmt_skip)]
macro_rules! foo {
...
}
```

**with this configuration**

```toml
msrv = "1.29"
```

I expected to see this happen: Clippy should see that `#[rustfmt::skip]` is not available in 1.29 and ignore it

Instead, this happened: Clippy suggested `#[rustfmt::skip]`, which fails with old version of Rust.

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.58.0-nightly (18bc4bee9 2021-11-02)
binary: rustc
commit-hash: 18bc4bee9710b181b440a472635150f0d6257713
commit-date: 2021-11-02
host: x86_64-unknown-linux-gnu
release: 1.58.0-nightly
LLVM version: 13.0.0
```

@rustbot label +I-suggestion-causes-error

PR #7944, opened 2021-11-07
Add MSRV to deprecated_cfg_attr

Serial-ATA:
changelog: Add MSRV to [`deprecated_cfg_attr`]

closes: #7922


clippy_lints/src/lib.rs
```diff
@@ -404,10 +404,21 @@ use crate::utils::conf::TryConf;
...
+
+    let msrv = conf.msrv.as_ref().and_then(|s| {
+        parse_msrv(s, None, None).or_else(|| {
+            sess.err(&format!(
+                "error reading Clippy's configuration file. `{}` is not a valid Rust version",
+                s
+            ));
+            None
+        })
+    });
```
Serial-ATA, line 418: I don't like copying these lines. Is there a reason these 3 lints can't just be a part of `register_plugins`?

giraffate (changes requested):
Can you add a test and update `clippy_lints/src/utils/conf.rs`? Here is a reference:
https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv

clippy_utils/src/msrvs.rs
```diff
@@ -27,6 +27,7 @@ msrv_aliases! {
     1,36,0 { ITERATOR_COPIED }
     1,35,0 { OPTION_COPIED, RANGE_CONTAINS }
     1,34,0 { TRY_FROM }
+    1,30,0 { TOOL_ATTRIBUTES }
     1,30,0 { ITERATOR_FIND_MAP }
```
giraffate, line 31:
```suggestion
    1,30,0 { ITERATOR_FIND_MAP, TOOL_ATTRIBUTES }
```

clippy_lints/src/lib.rs
giraffate: There seems to be no specific reason. IMHO, as the comments says, It might be easier to keep them separate as separate functions as they are now, since it is desirable that they are eventually removed.

giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #7934, opened 2021-11-05
ICE with `clippy::undocumented_unsafe_blocks`


calebzulawski:
### Code

I couldn't figure out how to reduce this (simple examples pass), but this occurs when running clippy on Portable SIMD (https://github.com/rust-lang/portable-simd, check out the `feature/safety` branch) after enabling this lint.  I'm pretty sure it has something to do with macros.

### Meta

**Rust version (`rustc -Vv`):**
```
cargo 1.58.0-nightly (94ca096af 2021-10-29)
release: 1.58.0
commit-hash: 94ca096afbf25f670e76e07dca754fcfe27134be
commit-date: 2021-10-29
host: x86_64-unknown-linux-gnu
libgit2: 1.3.0 (sys:0.13.23 vendored)
libcurl: 7.79.1-DEV (sys:0.4.49+curl-7.79.1 vendored ssl:OpenSSL/1.1.1l)
os: Ubuntu 20.04 (focal) [64-bit]
```

Clippy version from master: `clippy 0.1.58 (84a4ab7f3 2021-11-04)`

### Error output

```
thread 'rustc' panicked at 'assertion failed: self.start_pos.to_u32() + total_extra_bytes <= bpos.to_u32()', compiler/rustc_span/src/lib.rs:1682:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust-clippy/issues/new

note: Clippy version: clippy 0.1.58 (84a4ab7f3 2021-11-04)

query stack during panic:
#0 [analysis] running analysis passes on this crate
end of query stack
error: could not compile `core_simd`
```

<details><summary>Backtrace</summary>
  <p>
  
  ```

thread 'rustc' panicked at 'assertion failed: self.start_pos.to_u32() + total_extra_bytes <= bpos.to_u32()', compiler/rustc_span/src/lib.rs:1682:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/baba6687df3e83fdb15cc6ec239b4a1c75a30505/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/baba6687df3e83fdb15cc6ec239b4a1c75a30505/library/core/src/panicking.rs:106:14
   2: core::panicking::panic
[...]

PR #7945, opened 2021-11-07
Fix ICE in undocumented_unsafe_blocks

Serial-ATA:
changelog: Fix ICE in [`undocumented_unsafe_blocks`]

closes: #7934


calebzulawski:
I built this locally and confirmed that it now correctly lints `portable-simd` 👍

flip1995:
Thanks! I have the feeling, that we might see more ICEs with this lint... I hope I'm wrong.

@bors r+

PR #7948, opened 2021-11-08
Lint for bool to integer casts in `cast_lossless`

### LINKED ISSUE #7947: cast_lossless should lint about bool -> integer casts?
Lint name: `cast_lossless`

I tried this code:

```rust
#![forbid(clippy::cast_lossless)]

fn main() {
    let _x = true as u8;
}
```

I expected to see this happen: Clippy to complain about the bool -> u8 cast, since it can be replaced with `u8::from(true)`.

Instead, this happened: No lint was triggered

---

The lint description does say it is a cast between *numeric* types which can be replaced with safe conversion functions, and `bool` isn't strictly a numeric type, but `u8::from(bool)` does exist. So it's unclear to me if this warning should be on `cast_lossless` or a new lint, but I feel it should exist.

### Meta

```
rustc 1.58.0-nightly (18bc4bee9 2021-11-02)
binary: rustc
commit-hash: 18bc4bee9710b181b440a472635150f0d6257713
commit-date: 2021-11-02
release: 1.58.0-nightly
LLVM version: 13.0.0
```

(clippy version 0.1.58 (18bc4be 2021-11-02))

5225225:
The lint description says

> Checks for casts between *numerical* types that may be replaced by safe conversion functions.

Which is strictly speaking being violated here, but it seems within the spirit of the lint. I think it is still a useful lint to have, and having a different lint for just this feels excessive. Thoughts?

Fixes #7947

changelog: Lint for bool to integer casts in [`cast_lossless`]


5225225:
Just realised this lint doesn't seem to check for the MSRV. Looks like all the `From` impls for bools were added in 1.28. (And the ones for integers are 1.5/1.6, old enough to not care about I suppose)

Worth adding a MSRV check for the bools?

llogiq:
The linting part looks really thoughtfully done. Great job! However, I think the lint message doesn't exactly fit the cast. I doubt we ever get to the point where we get a `u0` type that cannot be created from a bool.

So I think we should change the message for bool → int casts.

5225225:
True, but then again, I can't really think of a *better* message. It *can* become silently lossy if you change the type (of the bool). But yeah, you're right, there's no possible change of the right hand side which would make it invalid.

```
error: casting `bool` to `i64` may become silently lossy if you later change the bool's type
error: casting `bool` to `i64` may become silently lossy if you later change the bool
error: casting `bool` to `i64` may become silently lossy if you later change the type of the bool
```

perhaps?

None of those seem obviously better than just leaving the message as is to me.

llogiq:
I'm currently short on time, but I'll think of something.

llogiq:
I don't think it's very likely that someone changes a value from bool to some integer type while forgetting some `as` conversions. How about 

```
error: Casting `bool` to `u64` is more cleanly stated with `i64::from(_)`.
```

5225225:
Okay, that's changed.

I used a literal `_` in the message rather than actually showing the real expression since we show the real expression in the suggestion anyways.

llogiq:
That's totally OK. No need to duplicate the suggestion. @bors r+

Issue #7921, opened 2021-11-02
cargo clippy --fix fails fixing redundant_pattern_matching


Yuri6037:
I tried this code:

```rust
pub trait OptionExtension<T>
{
    fn get_or_insert_with_err<TError, F: FnOnce() -> Result<T, TError>>(
        &mut self,
        f: F
    ) -> Result<&mut T, TError>;
}

impl<T> OptionExtension<T> for Option<T>
{
    fn get_or_insert_with_err<TError, F: FnOnce() -> Result<T, TError>>(
        &mut self,
        f: F
    ) -> Result<&mut T, TError>
    {
        if let None = *self {
            *self = Some(f()?);
        }

        match self {
            Some(v) => Ok(v),
            // SAFETY: a `None` variant for `self` would have been replaced by a `Some`
            // variant in the code above.
            None => unsafe { std::hint::unreachable_unchecked() }
        }
    }
}
```

I expected to see this happen: Clippy fixing this redundant_pattern_matching

Instead, this happened: I got a bug report from clippy:
```
warning: failed to automatically apply fixes suggested by rustc to crate `testafs`

after fixes were automatically applied the compiler reported errors within these files:

  * src/lib.rs

This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see 
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:
error[E0614]: type `bool` cannot be dereferenced
  --> src/lib.rs:16:12
   |
16 |         if *self.is_none() {
   |            ^^^^^^^^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0614`.
Original diagnostics will follow.
```

### Meta

**Rust version (`rustc -Vv`):**

```
rustc 1.56.0 (09c42c458 2021-10-18)
binary: rustc
commit-hash: 09c42c45858d5f3aedfa670698275303a3d19afa
commit-date: 2021-10-18
host: x86_64-unknown-linux-gnu
release: 1.56.0
LLVM version: 13.0.0
```

PR #7949, opened 2021-11-08
Fix suggestion for deref expressions in redundant_pattern_matching

Serial-ATA:
changelog: Fix suggestion for deref expressions in [`redundant_pattern_matching`]

closes: #7921

### DIFF
--- a/clippy_lints/src/matches.rs
+++ b/clippy_lints/src/matches.rs
@@ -1785,7 +1785,8 @@ mod redundant_pattern_match {
     use super::REDUNDANT_PATTERN_MATCHING;
     use clippy_utils::diagnostics::span_lint_and_then;
     use clippy_utils::higher;
-    use clippy_utils::source::{snippet, snippet_with_applicability};
+    use clippy_utils::source::snippet;
+    use clippy_utils::sugg::Sugg;
     use clippy_utils::ty::{implements_trait, is_type_diagnostic_item, is_type_lang_item, match_type};
     use clippy_utils::{is_lang_ctor, is_qpath_def_path, is_trait_method, paths};
     use if_chain::if_chain;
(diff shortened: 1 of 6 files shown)


tests/ui/redundant_pattern_matching_option.rs
```diff
@@ -95,3 +97,8 @@ const fn issue6067() {
         None => true,
     };
 }
+
+#[allow(clippy::deref_addrof)]
+fn issue7921() {
+    if let None = *(&None::<()>) {}
```
flip1995, line 103: Does this also work for 

```suggestion
    if let None = *&None::<()> {}
```

I suspect, that the suggestion for this will be `if &None::<()>.is_none() {`, which will produce a compile error.

If that's the case, we should use our `clippy_utils::Sugg` struct with `.maybe_paren()`.

tests/ui/redundant_pattern_matching_option.rs
```diff
@@ -63,6 +63,8 @@ fn main() {
     } else {
         3
     };
+
+    issue7921();
```
flip1995, line 67: This shouldn't be necessary, since unused code is generally allowed in UI tests. 

EDIT: It may be necessary for the fixed test to pass. If so, just ignore my comment and resolve this thread.

clippy_lints/src/matches.rs
```diff
@@ -2069,12 +2072,15 @@ mod redundant_pattern_match {
...
 
-                let mut app = if needs_drop {
+                let app = if needs_drop {
                     Applicability::MaybeIncorrect
                 } else {
                     Applicability::MachineApplicable
                 };
-                let sugg = snippet_with_applicability(cx, op_span, "_", &mut app);
+
+                let sugg = Sugg::hir_with_macro_callsite(cx, result_expr, "_")
```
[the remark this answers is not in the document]
flip1995: Ah, just saw the test update. Nvm.

flip1995:
@bors r+

Thanks!

PR #7956, opened 2021-11-11
Author improvements

camsteffen:
changelog: none

Various aspects of the author implementation are re-imagined to be much less repetitive. Also fixes some bugs. I hope this makes author more fun to work on for future contributors.

The last commit is pretty heavy but I tried to at least separate some changes so that the test file diffs per commit are simple.


llogiq:
Wow, that's a solid improvement! Also having those commits split up to review was pretty helpful. However, it seems that you're still actively working on it, so please let me know when I should do a final round of review.

camsteffen:
Sorry I did a tiny tweak when I found an extra `, _` in the output for `StmtKind::Expr`. It's ready now.

llogiq:
Great! Thank you! @bors r+

PR #7963, opened 2021-11-12
Use rustfmt version from `rust-toolchain`

Jarcho:
changelog: None


flip1995:
Hm, I wonder why we didn't do this earlier. I can't think of anything that would prevent this change.

@bors r+

Thanks!

PR #7966, opened 2021-11-12
Run rustfmt on batches of multiple files

Alexendoo:
changelog: none

This gives `cargo dev fmt` a nice speed boost, down from 90s (because old) on my laptop and 120s (because windows) on my desktop to ~5s on both

250 at a time was to give windows a good amount of headroom (failed at ~800, https://github.com/rust-lang/rust/issues/40384)

Also adds rustfmt to the toolchain file and has the clippy_dev workflow test using the pinned version as a follow up to #7963

### DIFF
--- a/clippy_dev/src/fmt.rs
+++ b/clippy_dev/src/fmt.rs
@@ -1,6 +1,7 @@
 use crate::clippy_project_root;
+use itertools::Itertools;
 use shell_escape::escape;
-use std::ffi::OsStr;
+use std::ffi::{OsStr, OsString};
 use std::path::Path;
 use std::process::{self, Command};
 use std::{fs, io};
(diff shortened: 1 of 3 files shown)


clippy_dev/src/fmt.rs
```diff
@@ -64,7 +65,11 @@ pub fn run(check: bool, verbose: bool) {
                 continue;
             }
 
-            success &= rustfmt(context, path)?;
+            paths.push(entry.into_path());
+        }
+
+        for chunk in paths.chunks(250) {
+            success &= rustfmt(context, chunk)?;
```
camsteffen, line 72: I think we should avoid collecting *all* the paths into the Vec at once. You could use `Itertools::chunks`.

Alexendoo: 👍 changed

camsteffen:
This was sorely needed, thanks! `cargo dev fmt` goes from 70s to 3.2s on my machine!

@bors r+

flip1995:
Wait it took so long for you? It was only at 10s on every machine for me, so I didn't bother looking into it. 😄 Thanks so much for this!

camsteffen:
I was surprised that it was that long. Has it gotten slower??

Alexendoo:
Now that you mention it the before numbers do seem a bit odd, I measured it after a rebase but it never felt like 2 minutes, more like 20-30s

Alexendoo:
Surprisingly it got slower in 24d561f0081b047d3a54f70dbff2a1b621f62749. Looks like rustup takes ~100ms extra if you don't manually specify the toolchain, which adds up over lots of tests

```
Benchmark 1: rustfmt --version
  Time (mean ± σ):     130.3 ms ±   2.0 ms    [User: 4.7 ms, System: 5.9 ms]
  Range (min … max):   127.1 ms … 133.7 ms    21 runs

Benchmark 2: rustfmt +nightly-2021-11-04 --version
  Time (mean ± σ):      37.5 ms ±   1.5 ms    [User: 2.7 ms, System: 8.4 ms]
  Range (min … max):    34.5 ms …  41.9 ms    58 runs
```

flip1995:
Is this maybe a perf regression of rustfmt since 2021-11-04? Those numbers seem really high and would explain why it took up to 2 minutes.

Alexendoo:
I believe it's not on rustfmt's end, it happens with rustc as well, here on stable (with no override/rust-toolchain file around)

```
Benchmark 1: rustc --version
  Time (mean ± σ):     135.8 ms ±   2.8 ms    [User: 4.3 ms, System: 7.7 ms]
  Range (min … max):   131.7 ms … 143.7 ms    19 runs

Benchmark 2: rustc +stable --version
  Time (mean ± σ):      44.0 ms ±   2.3 ms    [User: 5.3 ms, System: 8.6 ms]
  Range (min … max):    40.4 ms …  51.4 ms    55 runs
```

I plan to take a better look tomorrow

Alexendoo:
I think it's down to https://github.com/rust-lang/rustup/issues/2626

Issue #7820, opened 2021-10-14
False positive in manual_map with unsafe fn


dtolnay:
```rust
pub trait OptionExt<T> {
    #[allow(clippy::missing_safety_doc)]
    unsafe fn unsafe_map<U>(self, op: unsafe fn(T) -> U) -> Option<U>;
}

impl<T> OptionExt<T> for Option<T> {
    unsafe fn unsafe_map<U>(self, op: unsafe fn(T) -> U) -> Option<U> {
        match self {
            Some(t) => Some(op(t)),
            None => None,
        }
    }
}
```

```console
$ cargo clippy -- -Dclippy::all

error: manual implementation of `Option::map`
  --> src/main.rs:8:9
   |
8  | /         match self {
9  | |             Some(t) => Some(op(t)),
10 | |             None => None,
11 | |         }
   | |_________^ help: try this: `self.map(op)`
   |
   = note: `-D clippy::manual-map` implied by `-D clippy::all`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_map
```

Clippy's suggested replacement does not compile.

```console
error[E0277]: expected a `FnOnce<(T,)>` closure, found `unsafe fn(T) -> U`
 --> src/main.rs:8:18
  |
8 |         self.map(op)
  |              --- ^^ expected an `FnOnce<(T,)>` closure, found `unsafe fn(T) -> U`
  |              |
  |              required by a bound introduced by this call
  |
  = help: the trait `FnOnce<(T,)>` is not implemented for `unsafe fn(T) -> U`
```

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.57.0-nightly (dfc5add91 2021-10-13)
binary: rustc
commit-hash: dfc5add915e8bf4accbb7cf4de00351a7c6126a1
commit-date: 2021-10-13
host: x86_64-unknown-linux-gnu
release: 1.57.0-nightly
LLVM version: 13.0.0
```

PR #7968, opened 2021-11-13
Fix `manual_map` with unsafe functions

Jarcho:
fixes: #7820
changelog: Fix `manual_map` suggestion when used with unsafe functions and unsafe blocks

### DIFF
--- a/clippy_lints/src/manual_map.rs
+++ b/clippy_lints/src/manual_map.rs
@@ -179,37 +179,22 @@ impl LateLintPass<'_> for ManualMap {
                     } else {
                         ""
                     };
+                    let expr_snip = snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0;
                     if some_expr.needs_unsafe_block {
-                        format!(
-                            "|{}{}| unsafe {{ {} }}",
-                            annotation,
-                            some_binding,
-                            snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0
-                        )
+                        format!("|{}{}| unsafe {{ {} }}", annotation, some_binding, expr_snip)
                     } else {
-                        format!(
-                            "|{}{}| {}",
-                            annotation,
-                            some_binding,
-                            snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0
-                        )
+                        format!("|{}{}| {}", annotation, some_binding, expr_snip)
                     }
                 }
             }
         } else if !is_wild_none && explicit_ref.is_none() {
             // TODO: handle explicit reference annotations.
+            let pat_snip = snippet_with_context(cx, some_pat.span, expr_ctxt, "..", &mut app).0;
+            let expr_snip = snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0;
             if some_expr.needs_unsafe_block {
-                format!(
-                    "|{}| unsafe {{ {} }}",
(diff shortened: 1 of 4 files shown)


clippy_lints/src/manual_map.rs
```diff
@@ -176,21 +179,38 @@ impl LateLintPass<'_> for ManualMap {
...
+                    snippet_with_context(cx, some_pat.span, expr_ctxt, "..", &mut app).0,
+                    snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0
+                )
+            } else {
+                format!(
+                    "|{}| {}",
+                    snippet_with_context(cx, some_pat.span, expr_ctxt, "..", &mut app).0,
+                    snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0
+                )
+            }
```
xFrednet, line 213: Both `format!` macros use the same arguments, here it would be better to store them in a value beforehand. That avoids duplicated expressions and is better readable IMO :upside_down_face: 

```rs
let closure_expr = snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0;
let closure_args = snippet_with_context(cx, some_pat.span, expr_ctxt, "..", &mut app).0;
if some_expr.needs_unsafe_block {
    format!("|{}| unsafe {{ {} }}", closure_args, closure_expr)
} else {
    format!("|{}| {}", closure_args, closure_expr)
}
```

clippy_lints/src/manual_map.rs
```diff
@@ -176,21 +179,38 @@ impl LateLintPass<'_> for ManualMap {
...
+                        )
+                    } else {
+                        format!(
+                            "|{}{}| {}",
+                            annotation,
+                            some_binding,
+                            snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0
+                        )
+                    }
+                }
```
xFrednet, line 197: These can also be shortened if you add a `let closure_expr = ..` variable for the closure body beforehand.

```rs
let closure_expr = snippet_with_context(cx, some_expr.expr.span, expr_ctxt, "..", &mut app).0;
if some_expr.needs_unsafe_block {
    format!("|{}{}| unsafe {{ {} }}", annotation, some_binding, closure_expr)
} else {
    format!("|{}{}| {}", annotation, some_binding, closure_expr)
}
```

You can add the `closure_expr` variable before the `let body_str = ...` in line 161 as all uses have the same expression.

tests/ui/manual_map_option_2.rs
```diff
@@ -53,4 +53,18 @@ fn main() {
...
+        let _ = match Some(0) {
+            Some(x) => Some(f(x)),
+            None => None,
+        };
+    }
+    let _ = match Some(0) {
+        Some(x) => unsafe { Some(f(x)) },
+        None => None,
+    };
 }
```
xFrednet, line 70: On additional test would be good for the expression itself, like this:
```rs
    let _ = match Some(0) {
        Some(x) => Some(unsafe { f(x) }),
        None => None,
    };
```

xFrednet (commented):
The adjustments look good to me! Just some small suggestions for better formatting, and then it's ready to be merged IMO :upside_down_face:

clippy_lints/src/manual_map.rs
Jarcho: Line 167 (the `then` block of the `if_chain`) doesn't use it.

xFrednet (approved):

xFrednet:
LGTM, thank you for the fixes!

@bors r+

Issue #7960, opened 2021-11-11
The suggested code by `option_map_or_none` is warned again in some case


giraffate:
```rs
fn main() {
    let opt = Some(3);
    opt.map_or(None, |a| Some(a + 1));
}
```
Clippy warns like below on this:
```
  |
3 |     opt.map_or(None, |a| Some(a + 1));
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try using `and_then` instead: `opt.and_then(|a| Some(a + 1))`
  |
```
So I updated it according to the warning:
```rs
fn main() {
    let opt = Some(3);
    opt.and_then(|a| Some(a + 1));
}
```
But, Clippy warns like below again:
```
  |
3 |     opt.and_then(|a| Some(a + 1));
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `opt.map(|a| a + 1)`
  |
```

In this case, I want `option_map_or_none` to suggest the last code suggested from the beginning.

PR #7971, opened 2021-11-13
fix suggestion in option_map_or_none

togami2864:
fix: #7960 
changelog: change suggestion in the lint rule `option_map_or_none`

### DIFF
--- a/tests/ui/option_map_or_none.fixed
+++ b/tests/ui/option_map_or_none.fixed
@@ -4,13 +4,18 @@
 
 fn main() {
     let opt = Some(1);
+    let bar = |_| {
+       Some(1)
+    };
 
     // Check `OPTION_MAP_OR_NONE`.
     // Single line case.
-    let _ = opt.map(|x| Some(x + 1));
+    let _ :Option<i32> = opt.map(|x| x + 1);
     // Multi-line case.
     #[rustfmt::skip]
-    let _ = opt.map(|x| {
-                        Some(x + 1)
+    let _ :Option<i32> = opt.map(|x| {
+                        x + 1
                        });
+    // function returning `Option`
+    let _ :Option<i32> = opt.and_then(bar);
 }
--- a/tests/ui/option_map_or_none.fixed
+++ b/tests/ui/option_map_or_none.fixed
@@ -14,4 +14,9 @@ fn main() {
     let _: Option<i32> = opt.map(|x| x + 1);
     // function returning `Option`
     let _: Option<i32> = opt.and_then(bar);
+    let _: Option<i32> = opt.and_then(|x| {
+        let offset = 0;
+        let height = x;
+        Some(offset + height)
+    });
 }
--- a/tests/ui/option_map_or_none.fixed
+++ b/tests/ui/option_map_or_none.fixed
@@ -4,16 +4,14 @@
 
 fn main() {
     let opt = Some(1);
-    let bar = |_| {
-        Some(1)
-    };
+    let bar = |_| Some(1);
 
     // Check `OPTION_MAP_OR_NONE`.
     // Single line case.
-    let _ :Option<i32> = opt.map(|x| x + 1);
+    let _: Option<i32> = opt.map(|x| x + 1);
     // Multi-line case.
     #[rustfmt::skip]
-    let _ :Option<i32> = opt.map(|x| x + 1);
+    let _: Option<i32> = opt.map(|x| x + 1);
     // function returning `Option`
-    let _ :Option<i32> = opt.and_then(bar);
+    let _: Option<i32> = opt.and_then(bar);
 }
--- a/tests/ui/option_map_or_none.fixed
+++ b/tests/ui/option_map_or_none.fixed
@@ -7,10 +7,10 @@ fn main() {
 
     // Check `OPTION_MAP_OR_NONE`.
     // Single line case.
-    let _ = opt.and_then(|x| Some(x + 1));
+    let _ = opt.map(|x| Some(x + 1));
     // Multi-line case.
     #[rustfmt::skip]
-    let _ = opt.and_then(|x| {
+    let _ = opt.map(|x| {
                         Some(x + 1)
                        });
 }
--- a/tests/ui/option_map_or_none.fixed
+++ b/tests/ui/option_map_or_none.fixed
@@ -5,7 +5,7 @@
 fn main() {
     let opt = Some(1);
     let bar = |_| {
-       Some(1)
+        Some(1)
     };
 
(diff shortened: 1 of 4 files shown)


tests/ui/option_map_or_none.fixed
```diff
@@ -7,10 +7,10 @@ fn main() {
 
     // Check `OPTION_MAP_OR_NONE`.
     // Single line case.
-    let _ = opt.and_then(|x| Some(x + 1));
+    let _ = opt.map(|x| Some(x + 1));
```
llogiq, line 10: I think you should also remove the `Some(  )`, otherwise you'll change the type from `Option<_>` to `Option<Option<_>>`.

llogiq (changes requested):
I think this introduces a change of behavior. Please add a test case with the type of the operation given, e.g. change `let _ = ..` to `let _: Option<i32>` or some such.

tests/ui/option_map_or_none.fixed
togami2864: Is it correct that I should somehow remove `Some ( )` in this line?
https://github.com/rust-lang/rust-clippy/blob/f51fb341dd3ff79f400ee816dceb81eb2d0c5106/clippy_lints/src/methods/option_map_or_none.rs#L54

llogiq: Exactly. The problem here is that while this should usually be `Some(_)` which is an `ExprCall(..)` with a single item in `args`, it could also be any other expression that evaluates to an `Option`, e.g. `{ let foo = Some(bar); foo }` or `function_returning_option()`.

The right way to do this is to check if this is a `ExprKind::Call` to `Some(_)`, then remove the call and use the zeroeth argument with the `map`, or using the original expression with `and_then` otherwise.

togami2864: Now, I don't understand how to identify whether `Some` is in the code or not.

### What I did
#### case1: 
I tried to identify it by match with `expr.kind` and write like this.
```
if_chain! {
                if let hir::ExprKind::MethodCall(path, _, _,_ ) = &expr.kind;
                if let hir::ExprKind::Path(ref qpath) = path[0].kind;
                 .....
            }
```
But I couldn't get the path of `Some( )` in `Closure` which is expected as the oneth arg in `map_or`.

#### case2:
I tried to identify it by match with `map_arg`. 
https://github.com/rust-lang/rust-clippy/blob/f82bf470d029cc4c998bfbe15acec8ec701c2aab/clippy_lints/src/methods/option_map_or_none.rs#L20
However, the type of `map_arg` differs depending on the code you lint. (in this case `ExprKind::Closure` or `ExprKind::Path`.)


In both cases, I couldn't specify `Some()` in code.
Would you have any advice? Due to my limited understanding, it is possible that my approach is fundamentally wrong.

llogiq: First, this is not an `ExprKind::MethodCall`, but an `ExprKind::Call`. The rest has been done before, e.g. in [`clippy_lints/src/methods/chars_cmp.rs#L22-26`](https://github.com/rust-lang/rust-clippy/blob/f82bf470d029cc4c998bfbe15acec8ec701c2aab/clippy_lints/src/methods/chars_cmp.rs#L22-L26).

togami2864: Thanks.
I tried to output `expr.kind`, but it seems to be judged as `ExprKind::MethodCall`🤔
https://github.com/rust-lang/rust-clippy/blob/f82bf470d029cc4c998bfbe15acec8ec701c2aab/clippy_lints/src/methods/option_map_or_none.rs#L17

```Rust
println!({:?}, expr.kind)
```

Result
```
MethodCall(PathSegment { ident: map_or#0, hir_id: Some(HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 26 }), res: Some(Err), args: None, infer_args: true }, tests/ui/option_map_or_none.rs:13:30: 13:36 (#0), [Expr { hir_id: HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 28 }, kind: Path(Resolved(None, Path { span: tests/ui/option_map_or_none.rs:13:26: 13:29 (#0), res: Local(HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 7 }), segments: [PathSegment { ident: opt#0, hir_id: Some(HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 27 }), res: Some(Local(HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 7 })), args: None, infer_args: true }] })), span: tests/ui/option_map_or_none.rs:13:26: 13:29 (#0) }, Expr { hir_id: HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 30 }, kind: Path(Resolved(None, Path { span: tests/ui/option_map_or_none.rs:13:37: 13:41 (#0), res: Def(Ctor(Variant, Const), DefId(2:36962 ~ core[489a]::option::Option::None::{constructor#0})), segments: [PathSegment { ident: None#0, hir_id: Some(HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 29 }), res: Some(Err), args: None, infer_args: true }] })), span: tests/ui/option_map_or_none.rs:13:37: 13:41 (#0) }, Expr { hir_id: HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 41 }, kind: Closure(Ref, FnDecl { inputs: [Ty { hir_id: HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 40 }, kind: Infer, span: tests/ui/option_map_or_none.rs:13:44: 13:45 (#0) }], output: DefaultReturn(tests/ui/option_map_or_none.rs:13:47: 13:47 (#0)), c_variadic: false, implicit_self: None }, BodyId { hir_id: HirId { owner: DefId(0:3 ~ option_map_or_none[aa57]::main), local_id: 39 } }, tests/ui/option_map_or_none.rs:13:43: 13:46 (#0), None), span: tests/ui/option_map_or_none.rs:13:43: 13:58 (#0) }], tests/ui/option_map_or_none.rs:13:30: 13:59 (#0))
```

llogiq: I wasn't talking about the `.map_or(..)` call expr, but the 2nd argument, which is a closure of which you need to get the `Body` first. Sorry I forgot about that.

llogiq:
Great! I think this is ready for merging now. Thank you for your patience in seeing this through.

@bors r+

Issue #7972, opened 2021-11-13
False positive redundant_slicing report when slicing as immutable slice to invoke Read trait methods taking (&mut self)


EdMcBane:
Lint name: redundant_slicing

I tried this code:

```rust
    let data = vec![1,2,3];
    let src: &[u8] = &data;
    let mut dest = vec![];
    (&src[..]).read_to_end(&mut dest).unwrap();
```

I expected no lint warnings, due to the fact that the slicing in the last line is needed to call read_to_end on the immutable slice src.

Instead clippy reports redundant slicing of the whole range
```
warning: redundant slicing of the whole range
 --> src/main.rs:6:5
  |
6 |     (&src[..]).read_to_end(&mut dest).unwrap();
  |     ^^^^^^^^^^ help: use the original value instead: `src`
  |
  = note: `#[warn(clippy::redundant_slicing)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_slicing
```

Consequently, `cargo clippy --fix` fails with:
```
fdegrassi@homerunner:/tmp/asd$ cargo clippy --fix
    Checking asd v0.1.0 (/tmp/asd)
warning: failed to automatically apply fixes suggested by rustc to crate `asd`

after fixes were automatically applied the compiler reported errors within these files:

  * src/main.rs

This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see 
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:
error[E0596]: cannot borrow `src` as mutable, as it is not declared as mutable
 --> src/main.rs:6:5
  |
4 |     let src: &[u8] = &data;
  |         --- help: consider changing this to be mutable: `mut src`
5 |     let mut dest = vec![];
6 |     src.read_to_end(&mut dest).unwrap();
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable

error: aborting due to previous error

For more information about this error, try `rustc --explain E0596`.
Original diagnostics will follow.

warning: redundant slicing of the whole range
 --> src/main.rs:6:5
  |
6 |     (&src[..]).read_to_end(&mut dest).unwrap();
[...]

PR #7976, opened 2021-11-14
`redundant_slicing` precedence fix

Jarcho:
fixes: #7972
changelog: Fix `redundant_slicing` suggestion when the result is used in a postfix expression

### DIFF
--- a/clippy_lints/src/redundant_slicing.rs
+++ b/clippy_lints/src/redundant_slicing.rs
@@ -1,11 +1,12 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::get_parent_expr;
-use clippy_utils::source::snippet_with_context;
+use clippy_utils::source::{position_of_expr, snippet_expr, ExprPosition};
 use clippy_utils::ty::is_type_lang_item;
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::{BorrowKind, Expr, ExprKind, LangItem, Mutability};
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::adjustment::{Adjust, Adjustment, AutoBorrow, AutoBorrowMutability};
 use rustc_middle::ty::TyS;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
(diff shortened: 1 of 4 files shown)


clippy_utils/src/source.rs
```diff
@@ -306,6 +306,199 @@ pub fn snippet_with_context(
     )
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ExprPosition {
```
camsteffen, line 310: Can we use `ExprPrecedence` from rustc?

Jarcho: Doesn't distinguish between the left and right. Given `a - b` replacing `a` with `x + y` doesn't require parenthesis, replacing `b` does.

camsteffen: Here is how [pretty print](https://github.com/rust-lang/rust/blob/b0535508474fb4353a3b83ef79df87817f028a71/compiler/rustc_ast_pretty/src/pprust/state.rs#L1839-L1847) does it. I guess you'd still have to pass `is_rhs: bool` in some form so 🤷.

xFrednet: Seams like there is no ideal solution. I think this way is acceptable, are you also fine with it @camsteffen?

clippy_lints/src/redundant_slicing.rs
```diff
@@ -57,34 +58,47 @@ impl LateLintPass<'_> for RedundantSlicing {
...
-                    format!("{}{}", reborrow_str, snip),
+                    if !reborrow_str.is_empty() && position > ExprPosition::Prefix {
+                        format!("({}{})", reborrow_str, snip)
+                    } else {
+                        format!("{}{}", reborrow_str, snip)
+                    },
                     app,
                 );
             }
+
```
xFrednet, line 101:
```suggestion
```

clippy_utils/src/source.rs
```diff
@@ -306,6 +306,199 @@ pub fn snippet_with_context(
...
+/// snippet of the chile of the borrow expression will get a snippet of what `vec![]` expands in to.
+/// With the target context set to the same as the borrow expression, this will get a snippet of the
+/// call to the macro.
+///
+/// The applicability will be modified in two ways:
+/// * If a snippet can't be extracted it will be changed from `MachineApplicable` or
+///   `MaybeIncorrect` to `HasPlaceholders`.
+/// * If the snippet is taken from a macro expansion then it will be changed from
+///   `MachineApplicable` to `MaybeIncorrect`.
+pub fn snippet_expr(
```
xFrednet, line 359: The tests don't have any instance where the parenthesis are added by this function. Can you please add some test cases to ensure that this works correctly?

clippy_utils/src/source.rs
```diff
@@ -306,6 +306,199 @@ pub fn snippet_with_context(
...
+
+/// Extracts a snippet of the given expression taking into account the `SyntaxContext` the snippet
+/// needs to be taken from. Parenthesis will be added if needed to place the snippet in the target
+/// precedence level. Returns a placeholder (`(..)`) if a snippet can't be extracted (e.g. an
+/// invalid span).
+///
+/// The `SyntaxContext` of the expression will be walked up to the given target context (usually
+/// from the parent expression) before extracting a snippet. This allows getting the call to a macro
+/// rather than the expression from expanding the macro. e.g. In the expression `&vec![]` taking a
+/// snippet of the chile of the borrow expression will get a snippet of what `vec![]` expands in to.
```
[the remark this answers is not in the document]
Jarcho: I swear I know how to type.

clippy_utils/src/source.rs
camsteffen: Yeah this works.

Jarcho:
Just btw I'm working on #7986 which is an alternative to the extra util functions.

clippy_utils/src/source.rs
```diff
@@ -306,6 +306,199 @@ pub fn snippet_with_context(
...
+        Cow::Owned(mut snip) => {
+            let ctxt = expr.span.ctxt();
+
+            // Attempt to determine if parenthesis are needed base on the target position. The snippet may have
+            // parenthesis already, so attempt to find those.
+            // TODO: Remove parenthesis if they aren't needed at the target position.
+            let needs_paren = match expr.peel_drop_temps().kind {
+                ExprKind::Binary(_, lhs, rhs)
+                    if (ctxt == lhs.span.ctxt() && expr.span.lo() != lhs.span.lo())
+                        || (ctxt == rhs.span.ctxt() && expr.span.hi() != rhs.span.hi()) =>
```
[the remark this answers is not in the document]
Jarcho: It has to do with macro expansions and whether parenthesis are from the current expression or the sub expression. If the span on the current expansion doesn't match with the outer edges of the two sub expressions then the current expression has parenthesis already, but this doesn't work if either sub expression is the result of a macro expansion.

xFrednet:
Hey @Jarcho, are you planning to continue working on this PR or are you waiting on #7986 before doing more work here? :upside_down_face:

Jarcho:
I'll give it a fix for the specific issue as that doesn't need anything added to utils.

Jarcho:
Closing in favour of #8218.

Issue #5327, opened 2020-03-17
needless_borrow: suggest removing several references at a time.


disconnect3d:
Hey,

It would be nice if Clippy could detect needless references as in here:
```rust
fn main() {
    let st = String::from("zomg");
    myfunc(&&&&&st);
}

fn myfunc(param: &str) {
    println!("param: {}", param);
}
```

The `myfunc(&&&&&st);` should be `myfunc(&st);`.

Side note: there is a similar, needless lifetime parameter detecting, that would trigger in e.g. `fn myfunc<'a>(param: &'a str)`

matthiaskrgr:
Hi, this is [already implemented](https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow) but you have to opt-in manually (`#[warn(clippy::needless_borrow)]
`)

````rust
#[warn(clippy::needless_borrow)]
fn main() {
    let st = String::from("zomg");
    myfunc(&&&&&st);
}

fn myfunc(param: &str) {
    println!("param: {}", param);
}

````
should make clippy warn about it. :)

**Comment by disconnect3d** [comment]
@matthiaskrgr is there a way to opt in with a CLI argument? Also, how do I list all linters/detectors?

**Comment by disconnect3d** [comment]
@matthiaskrgr Also, the suggestion given is not the right one and applying it and re-running `cargo clippy` will still warn:

```
➜  refs git:(master) ✗ cargo clippy
    Checking refs v0.1.0 (/Users/dc/playground/rust-2/refs)
warning: this expression borrows a reference that is immediately dereferenced by the compiler
 --> src/main.rs:6:12
  |
6 |     myfunc(&&&&&st);
  |            ^^^^^^^ help: change this to: `&&&&st`
  |
note: lint level defined here
 --> src/main.rs:1:8
  |
1 | #[warn(clippy::needless_borrow)]
  |        ^^^^^^^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
```

seems like something to be improved

matthiaskrgr:
`cargo clippy -- -Wclippy::needless-borrow` should work.

Mhh, looks like the link only looks at one reference at a time.. :thinking:

flip1995:
All lints are listed here: https://rust-lang.github.io/rust-clippy/

[...]

Issue #1726, opened 2017-05-05
needless borrow does not trigger on block trailing expression


oli-obk:
```rust
    let x = 5;
    let foo = match 42 {
        44 => &x,
        45 => {
            println!("foo");
            &&x // does not trigger needless_borrow
        },
        46 => &&x, // triggers needless_borrow
        _ => panic!(),
    };
```

oli-obk:
Ok... so this doesn't trigger on any expression where the location of the autoderef is disconnected from the location of the address-taking. In the match example, the issue is probably in the expansion of match arms.

e.g.

```rust
    let intermediate = &&5;
    let b: &i32 = intermediate; // doesn't lint
    let c: &i32 = &&5; // lints
```

This just became much harder to solve.

Any complaints about a lint on locals with `&&T` type which are only ever used by (auto-)derefing? The suggestion would be to deref at the initialization site instead of at each use site.

Issue #1212, opened 2016-09-02
Unnecessary references for casting from immutable reference to mutable reference.


ticki:
A warning should be generated with:

``` rust
let mut a = 2;
let mut a_mut = &mut a;
let a_ref: &u8 = &a_mut;
```

which is more approriately written as:

``` rust
let mut a = 2;
let mut a_mut = &mut a;
let a_ref: &u8 = a_mut;
```

(when possible)

mcarton:
I would have guessed this was already covered by [`needless_borrow`](https://manishearth.github.io/rust-clippy/current/#needless_borrow), but it seems like not.

oli-obk:
I think the issue is that this is a conversion from a `&&mut T` to a `&T`. I'll check how we can improve the `needless_borrow` lint.

PR #7977, opened 2021-11-15
Improve `needless_borrow` lint

Jarcho:
fixes: #5327
fixes: #1726
fixes: #1212

This is merging `needless_borrow` into the `dereference` pass in preparation for `explicit_auto_deref`. `explicit_auto_deref` needs to implement most of what `needless_borrow` implements in order to work.

There is a minor regression here where `let x: &str = &x.deref()` will trigger `needless_borrow` without triggering `explicit_deref_methods`. Removing the redundant borrow will cause `explicit_deref_methods` to trigger. This will be fixed when `explicit_auto_deref` is implemented.

changelog: Lint `needless_borrow` when a borrow is auto-derefed more than once
changelog: Lint `needless_borrow` in the trailing expression of a block for a match arm


xFrednet (approved):

xFrednet:
Looks good to me. The code is nicely good job :+1: 

@bors r+

PR #7980, opened 2021-11-16
Fix FP on `if_then_some_else_none` when there is early return

### LINKED ISSUE #7870: False positive with `clippy::if_then_some_else_none`
Lint name:

`clippy::if_then_some_else_none`

I tried this code:

```rust
#![warn(clippy::if_then_some_else_none)]

fn f(b: bool, v: Option<()>) -> Option<()> {
    if b {
        v?;

        Some(())
    } else {
        None
    }
}
```

I expected to see this happen: Clippy does not warn `clippy::if_then_some_else_none`.

Instead, this happened: Clippy  warns that code can be simplified.

### Meta

```
rustc 1.58.0-nightly (91b931926 2021-10-23)
binary: rustc
commit-hash: 91b931926fd49fc97d1e39f2b8206abf1d77ce7d
commit-date: 2021-10-23
host: x86_64-pc-windows-msvc
release: 1.58.0-nightly
LLVM version: 13.0.0
```

dswij:
closes #7870 

changelog: [`if_then_some_else_none`] now does not fire when there is early return

### DIFF
--- a/clippy_lints/src/if_then_some_else_none.rs
+++ b/clippy_lints/src/if_then_some_else_none.rs
@@ -1,9 +1,9 @@
 use clippy_utils::diagnostics::span_lint_and_help;
 use clippy_utils::source::snippet_with_macro_callsite;
-use clippy_utils::{higher, is_else_clause, is_lang_ctor, meets_msrv, msrvs};
+use clippy_utils::{contains_return, higher, is_else_clause, is_lang_ctor, meets_msrv, msrvs};
 use if_chain::if_chain;
 use rustc_hir::LangItem::{OptionNone, OptionSome};
-use rustc_hir::{Expr, ExprKind};
+use rustc_hir::{Expr, ExprKind, Stmt, StmtKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
 use rustc_semver::RustcVersion;
@@ -82,6 +82,7 @@ impl LateLintPass<'_> for IfThenSomeElseNone {
             if let Some(els_expr) = els_block.expr;
             if let ExprKind::Path(ref qpath) = els_expr.kind;
             if is_lang_ctor(cx, qpath, OptionNone);
+            if !stmts_contains_early_return(then_block.stmts);
             then {
                 let cond_snip = snippet_with_macro_callsite(cx, cond.span, "[condition]");
                 let cond_snip = if matches!(cond.kind, ExprKind::Unary(_, _) | ExprKind::Binary(_, _, _)) {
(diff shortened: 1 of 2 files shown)


xFrednet (commented):
The implementation looks good to me. 

---

For future reference: The example in the issue could in theory still use `bool::then()` and then call flatten on the result like this([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=88c0a3f85ca9a40fac08ec7f935188a1))

```rs
 fn f2(b: bool, v: Option<()>) -> Option<()> {
    b.then(|| {
        v?;
        Some(())
    })
    .flatten()
}
```

However, this only works if the early return is also inside the return expression of function. I like this solution implemented in this PR. Just wanted to document it for future reference :upside_down_face:

clippy_lints/src/if_then_some_else_none.rs
```diff
@@ -73,6 +73,7 @@ impl LateLintPass<'_> for IfThenSomeElseNone {
         if_chain! {
             if let Some(higher::If { cond, then, r#else: Some(els) }) = higher::If::hir(expr);
             if let ExprKind::Block(then_block, _) = then.kind;
+            if !stmts_contains_early_return(then_block.stmts);
```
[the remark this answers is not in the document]
dswij: Done 🙂

clippy_lints/src/if_then_some_else_none.rs
```diff
@@ -99,18 +100,26 @@ impl LateLintPass<'_> for IfThenSomeElseNone {
...
+                    );
                 span_lint_and_help(
                     cx,
                     IF_THEN_SOME_ELSE_NONE,
                     expr.span,
                     "this could be simplified with `bool::then`",
                     None,
                     &help,
-                );
+                    );
```
xFrednet, line 111: rustfmt sadly doesn't work inside macros. Could you fix this indention and the one in line 103? Then it's all good :+1:

xFrednet (approved):
Looks good, only one small formatting NIT :sweat_smile:

clippy_lints/src/if_then_some_else_none.rs
dswij: Hm, I thought that this change was because of rustfmt 🤔

xFrednet: These indentations also often happen in my editor, when I move code around. It's good that rustfmt is really solid otherwise, and this might be fixed when Rust supports `if let` chains in the future :upside_down_face:

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the bug fix! :upside_down_face: 

@bors r+

Issue #7975, opened 2021-11-13
`needless_collect` false positive around closures and mutable borrows.


Blckbrry-Pi:
Lint name:
`needless_collect`

Clippy emits a `needless_collect` warning, but after replacing the lint-failing code with the suggested code, it no longer compiles.

This is the offending code:
```rust
let test_vec: Vec<()> = vec![];

let mut vec_2: Vec<()> = vec![];
let mut_ref = &mut vec_2;

let collected_vec: Vec<_> = test_vec.into_iter().map(|_| mut_ref.push(())).collect();

let double_collected_vec: Vec<_> = collected_vec.into_iter().map(|_| mut_ref.push(())).collect();
```

Clippy's suggestion:
```
warning: avoid using `collect()` when not needed
  --> src/main.rs:6:76
   |
6  | let collected_vec: Vec<_> = test_vec.into_iter().map(|_| mut_ref.push(())).collect();
   |                                                                            ^^^^^^^
7  | 
8  | let double_collected_vec: Vec<_> = collected_vec.into_iter().map(|_| mut_ref.push(())).collect();
   |                                    ------------------------- the iterator could be used here instead
   |
   = note: `#[warn(clippy::needless_collect)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect
help: use the original Iterator instead of collecting it and then producing a new one
   |
6  + 
7  | 
8  ~ let double_collected_vec: Vec<_> = test_vec.into_iter().map(|_| mut_ref.push(())).map(|_| mut_ref.push(())).collect();
   |
```
This suggestion will not compile because both closures attempt to borrow the `&mut Vec<()>` as mutable at the same time.

**Expectation:** The code should not fail the lint because (at least from what I can see) there is no simple change that can be made so that the code does not use collect.

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.57.0-beta.3 (708d57e28 2021-11-01)
binary: rustc
commit-hash: 708d57e288d051a6238ed56039ffeac158e10e84
commit-date: 2021-11-01
host: aarch64-apple-darwin
release: 1.57.0-beta.3
LLVM version: 13.0.0
```
@rustbot label +I-suggestion-causes-error

SKyletoft:
This still affects me

```rust
use std::collections::HashMap;

fn main() {
        // Original
	let mut map = HashMap::<u64, u64>::new();
[...]

PR #7982, opened 2021-11-16
Fix `needless_collect`'s tendency to suggest code requiring multiple mutable borrows of the same value.

Blckbrry-Pi:
Fixes error specified in #7975.

changelog: [`needless_collect`] no longer suggests removal of `collect` when removal would create code requiring mutably borrowing a value multiple times.

### DIFF
--- a/clippy_lints/src/loops/needless_collect.rs
+++ b/clippy_lints/src/loops/needless_collect.rs
@@ -12,7 +12,7 @@ use rustc_hir::{Block, Expr, ExprKind, HirId, HirIdSet, Local, Mutability, Node,
 use rustc_lint::LateContext;
 use rustc_middle::hir::map::Map;
 use rustc_middle::ty::subst::GenericArgKind;
-use rustc_middle::ty::{TyKind, TyS};
+use rustc_middle::ty::{self, TyS};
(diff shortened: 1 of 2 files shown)


clippy_lints/src/loops/needless_collect.rs
```diff
@@ -237,12 +311,61 @@ impl<'a, 'tcx> Visitor<'tcx> for UsedCountVisitor<'a, 'tcx> {
...
     visitor.visit_block(block);
-    if visitor.seen_other { None } else { Some(visitor.uses) }
+    if visitor.seen_other {
+        None
+    } else {
+        Some(visitor.uses.into_iter().flatten().collect())
+    }
+}
+
+#[allow(rustc::usage_of_ty_tykind)]
```
llogiq, line 338: You can replace the `TyKind::` with `ty::` and `use rustc_middle::ty` above, then remove the `#[allow(..)]` here.

llogiq (commented):
This is already looking quite good. I only have some minor readability nits.

Blckbrry-Pi:
I have fixed the suggestions.  Thank you for making them!

llogiq:
Thank you! @bors r+

PR #7984, opened 2021-11-17
Pluralize `disallowed_type` lint

phansch:
This was brought up in [Zulip] and is also mentioned in the lint naming
conventions. Since this is still a nursery lint, I think there shouldn't
be any problem in renaming it.

[Zulip]: rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/disallow_type.20vs.20disallowed-types

changelog: Rename nursery lint [`disallowed_type`] to [`disallowed_types`].


clippy_lints/src/utils/conf.rs
```diff
@@ -15,10 +15,10 @@ pub struct Rename {
     pub rename: String,
 }
 
-/// A single disallowed method, used by the `DISALLOWED_METHOD` lint.
+/// A single disallowed method, used by the `DISALLOWED_METHODS` lint.
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
-pub enum DisallowedMethod {
+pub enum DisallowedMethods {
```
phansch, line 21: I think this type should still be singular as it represents a single method

flip1995: Oh right. I'll revert this. Good catch!

xFrednet (approved):

xFrednet:
LGTM, thanks!

@bors r+

PR #7986, opened 2021-11-17
WIP Suggestion creation macro

Jarcho:
This is a minimal and hacky implementation right now, but it should get the idea across.

A large number of suggestions in clippy are basically extracting a snippet of an expression (or multiple), adding something around it, and then replacing an expression. This is a proc-macro made to simplify that as much as possible. The syntax is a regular rust expression with meta-variables to insert snippets. It should handle keeping track of precedence and inserting parenthesis around snippets as needed, as well as keeping track of the final precedence 

Some examples:

```
expr_sugg!(&*$snip(e).foo);
expr_sugg!(&$mut(m) *foo);
expr_sugg!(&$mut(m) foo[$snip(e)]);
```

changelog: None

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -178,9 +178,11 @@ pub mod _internal {
 
     pub fn needs_parens(expr: ExprPrecedence, position: ExprPosition) -> bool {
         match expr {
-            ExprPrecedence::Closure | ExprPrecedence::Ret | ExprPrecedence::Break | ExprPrecedence::Yield => {
-                position > ExprPosition::Closure
-            },
+            ExprPrecedence::Closure
+            | ExprPrecedence::Ret
+            | ExprPrecedence::Break
+            | ExprPrecedence::Yield
+            | ExprPrecedence::Yeet => position > ExprPosition::Closure,
             ExprPrecedence::Assign | ExprPrecedence::AssignOp => position > ExprPosition::AssignRhs,
             ExprPrecedence::Range => position >= ExprPosition::RangeLhs,
             ExprPrecedence::Binary(BinOpKind::Or) => position > ExprPosition::OrLhs,
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -57,6 +57,265 @@ pub mod ty;
 pub mod usage;
 pub mod visitors;
 
+pub mod _internal {
+    use crate::diagnostics::span_lint_and_then;
+    use crate::get_parent_expr;
+    use crate::source::snippet_with_context;
+    use rustc_ast::{util::parser::ExprPrecedence, BinOpKind};
+    use rustc_errors::Applicability;
+    use rustc_hir::{Expr, ExprKind};
+    use rustc_lint::{LateContext, Lint};
+    use rustc_span::SyntaxContext;
+    use std::borrow::Cow;
+
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+    pub enum ExprPosition {
+        Closure,
+        AssignRhs,
+        AssignLhs,
+        RangeLhs,
+        RangeRhs,
+        OrLhs,
+        OrRhs,
+        AndLhs,
+        AndRhs,
+        Let,
+        EqLhs,
+        EqRhs,
+        BitOrLhs,
+        BitOrRhs,
+        BitXorLhs,
+        BitXorRhs,
+        BitAndLhs,
+        BitAndRhs,
+        ShiftLhs,
+        ShiftRhs,
+        AddLhs,
+        AddRhs,
+        MulLhs,
+        MulRhs,
+        Cast,
+        Prefix,
+        Suffix,
+        Callee,
+    }
+
+    pub trait ExprSugg {
+        fn into_sugg<'tcx>(self, cx: &LateContext<'tcx>, replace: &'tcx Expr<'tcx>, app: &mut Applicability) -> String;
+    }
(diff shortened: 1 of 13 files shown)


clippy_utils/src/lib.rs
```diff
@@ -56,6 +56,81 @@ pub mod ty;
 pub mod usage;
 pub mod visitors;
 
+pub mod _internal {
```
llogiq, line 59: Why the underscore? I would've thought the name makes it clear that this is not for public consumption. Perhaps add `#[doc(hidden)]` instead?

llogiq (commented):
Cool idea! I like how you did this without using `syn`. I do wonder how much this actually simplifies things though. Perhaps add another commit that uses it in a few lints once the tests pass so we can see the wins to be had?

clippy_utils/src/lib.rs
```diff
@@ -56,6 +56,81 @@ pub mod ty;
...
+            Some(x) if has_enclosing_paren(&x) => return x,
+            Some(x) => x,
+            None => return "(..)".into(),
+        };
+        match expr.kind {
+            ExprKind::Unary(..) | ExprKind::AddrOf(..) | ExprKind::Box(..) if position > Prefix => {
+                format!("({})", snip)
+            },
+            ExprKind::Closure(..) | ExprKind::Ret(..) | ExprKind::Break(..) | ExprKind::Yield(..)
+                if position > Closure =>
```
llogiq, line 105: That would be equal to `position != Closure`, right?

Jarcho: Yes.

clippy_utils/src/lib.rs
Jarcho: It's temporary anyways. Everything used will be regular `pub` items once I'm done.

llogiq: Ah, Ok then.

clippy_utils/src/lib.rs
llogiq: Then I'd prefer `!=`, because it's not susceptible to reordering.

Jarcho: `ExprPostion` is already in precedence order and relies on it.

llogiq: Ok then.

llogiq:
Anything needed or should I review this now?

camsteffen:
I think a good goal here would be for the macro to expand into code that could be written without the macro. This would take a lot of complexity out of the macro implementation.

For example, I would like to be able to write:

```rust
let sugg = Suggestion::borrow( // adds '&' and parenthesis according to arg precedence
    Suggestion::from_expr(cx, expr) // gets snippet
)
.position(cx, expr.hir_id) // adds outer parenthesis if needed when replacing `expr`
.to_string();
```

Jarcho:
That would strictly be more work. Currently it's being handled like a format string with the addition of keeping track of what precedence level snippets are being inserted at, and what the lowest precedence level in the expression is. 

> ```rust
> let sugg = Suggestion::borrow( // adds '&' and parenthesis according to arg precedence
>     Suggestion::from_expr(cx, expr) // gets snippet
> )
> .position(cx, expr.hir_id) // adds outer parenthesis if needed when replacing `expr`
> .to_string();
> ```

This kind of output requires building an ast as something like `x * $snip * y` and `x + $snip * y` would have to build the suggestion in different orders (the first being `mul(mul(x, $snip), y)` the second being `add(x, mul($snip, y))`).

That being said keeping `Sugg` around for more complicated things is still useful. (e.g. conditionally negating a suggestion)

Jarcho:
Just playing around with syntax options:

1. `$expr(expr1) + $expr(expr2)`
2. `$expr + $expr, expr1, expr2`
3. `{expr} + {expr}, expr1, expr2`

There's also another option where the type of the expansion has a default value, and only needs to be specified sometimes:

1. `$(expr1) + &$mut(mut1) foo`
2. `$_ + &$mut foo, expr1, mut1`
3. `{} + &{mut} foo, expr1, mut1`

I like the look of the format string versions, but it does mean any block expressions would need double curly braces. (e.g. `|x| { x }` would need to be `|x| {{ x }}`)

The types could also be specified after out of band (e.g. `{} = &{} foo, expr1, mut: mut1`), but that would require a second pass over the input as the trailing arguments would have to be parsed first.

llogiq:
I still wonder whether the macro is going to pull its weight in terms of complexity. While we have a number of rather complex suggestions, I doubt they add up to 1.2k lines of code.

Jarcho:
I have a few plans to shrink the amount of code, so there's still hope there.

Ultimately I'd like an interface along the lines of

```rust
lint_expr(cx, LINT_NAME, msg, expr, Applicability::MachineApplicable, sugg!($expr(foo) + 1));
```

This would cover a good chunk of lints which just replace one expression with another. Cover everything else with `span_lint_and_then` with some helper function to take the result of the `sugg!` macro.

llogiq:
Sorry for letting you wait, I had a very busy week. This should be good to merge after yet another rebase.

llogiq:
Sorry, I had a busy week. But another question arised: How does this mesh with the new [diagnostics translation effort](https://blog.rust-lang.org/inside-rust/2022/08/16/diagnostic-effort.html) (which may include clippy in the future)?

Jarcho:
As far as I can tell this shouldn't be any more work than what we would currently need to do.

xFrednet:
Since there hasn't been any activity since the last comment asking about the status, I'm going to close this PR. @Jarcho you're welcome to reopen it or create a new one, if you're interested in continuing this work :D

Issue #7979, opened 2021-11-16
ICE on undocumented_unsafe_blocks lint


pymongo:
```
73 ~     // Safety: ...
74 +     unsafe {
   |

thread 'rustc' panicked at 'internal error: entered unreachable code: attempted to remap an already remapped filename', compiler/rustc_span/src/source_map.rs:1107:34
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust-clippy/issues/new

note: Clippy version: clippy 0.1.58 (891ca5f 2021-11-15)

query stack during panic:
#0 [analysis] running analysis passes on this crate
```

## step to reproduce
1. git clone https://github.com/pymongo/leetcode-rust
2. git checkout 5432649be7c7fb82e24c5eac83585fe4936e1b77
3. cargo clippy

### Meta

**Rust version (`rustc -Vv`):**

```
rustc 1.58.0-nightly (891ca5f63 2021-11-15)
binary: rustc
commit-hash: 891ca5f63c3b3cfe3939710a728671243e881ed6
commit-date: 2021-11-15
host: x86_64-unknown-linux-gnu
release: 1.58.0-nightly
LLVM version: 13.0.0
```

### If I apply some `// Safety: ...` line to solve undocumented_unsafe_blocks lint, then cause another ICE panic

```
    Checking leetcode-rust v0.1.1 (/home/w/repos/my_repos/leetcode-rust)
thread 'rustc' panicked at 'assertion failed: self.start_pos.to_u32() + total_extra_bytes <= bpos.to_u32()', compiler/rustc_span/src/lib.rs:1677:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

error: internal compiler error: unexpected panic

note: the compiler unexpectedly panicked. this is a bug.

note: we would appreciate a bug report: https://github.com/rust-lang/rust-clippy/issues/new

note: Clippy version: clippy 0.1.58 (891ca5f 2021-11-15)
```

pymongo:
@ rustbot label +I-ICE

pymongo:
[...]

PR #7988, opened 2021-11-17
Fix ICE on `undocumented_unsafe_blocks`

giraffate:
fix https://github.com/rust-lang/rust-clippy/issues/7979

changelog: Fix ICE on `undocumented_unsafe_blocks`

### DIFF
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -137,10 +137,10 @@ impl UndocumentedUnsafeBlocks {
 
         let between_span = if block_span.from_expansion() {
             self.macro_expansion = true;
-            enclosing_scope_span.with_hi(block_span.hi())
+            enclosing_scope_span.with_hi(block_span.hi()).source_callsite()
         } else {
             self.macro_expansion = false;
-            enclosing_scope_span.to(block_span)
+            enclosing_scope_span.to(block_span).source_callsite()
         };
 
         let file_name = source_map.span_to_filename(between_span);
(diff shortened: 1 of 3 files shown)


flip1995:
@bors r+

Thanks! And another one from this lint...

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -137,10 +137,10 @@ impl UndocumentedUnsafeBlocks {
 
         let between_span = if block_span.from_expansion() {
             self.macro_expansion = true;
-            enclosing_scope_span.with_hi(block_span.hi())
+            enclosing_scope_span.with_hi(block_span.hi()).source_callsite()
         } else {
             self.macro_expansion = false;
-            enclosing_scope_span.to(block_span)
+            enclosing_scope_span.to(block_span).source_callsite()
```
camsteffen, line 143: This fixes the case in question but I think there are more cases unaccounted for. In general I think two spans should be ensured to have the same `span.ctxt()` before combining them with `to`, `with_hi`, etc, otherwise the operation is unpredictable. There can be any combination of contexts between the two spans.

A general fix might be...

```rust
if enclosing_scope_span.ctxt() != block_span.ctxt() {
    // get the span of the macro call
    // mac! { unsafe {} }
    enclosing_scope_span = hygiene::walk_chain(enclosing_scope_span, block_span.ctxt());
    if enclosing_scope_span.ctxt() != block_span.ctxt() {
        // if the contexts are still not equal after walk_chain,
        // the unsafe block span context is a descendant of the enclosing scope span context
        // { macro_with_unsafe!() }
        // return since I don't know what span to use
        return;
    }
}
```

But then I don't know if this lint is designed to work with the span of a macro call, where there can be an arbitrary amount of code between the start of the macro call and the unsafe block.

camsteffen: @flip1995 @giraffate

PR #7174, opened 2021-05-05
Fix eval_order_dependence async false positive

camsteffen:
changelog: Fix [`eval_order_dependence`] false positive in async code

Fixes #6925

PR #7997, opened 2021-11-19
Fixes shadow_same's false positive for #7915

### LINKED ISSUE #7915: `shadow_same` is just plain wrong in an async context
https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=bb8b58e91b2a00d51cd67875ebce1da6

**Code**

```rust
#![deny(clippy::shadow_same)]

pub async fn foo(_a: i32) {
}
```

**Output**

```
error: `_a` is shadowed by itself in `_a
```

surechen:
Fix shadow_same's false positive for async function's params(Fixes #7915):

Example Code:
```rust
#![deny(clippy::shadow_same)]

pub async fn foo(_a: i32) {
}
```
Output:
```
error: `_a` is shadowed by itself in `_a
```

Hir:
```rust
pub async fn foo(_a: i32)
 ->
     /*impl Trait*/ #[lang = "from_generator"](move |mut _task_context|
                                                   {
                                                       let _a = _a;
                                                       { let _t = { }; _t }
                                                   })
```

Skip checking async function's params.

changelog: Fix shadow_same's false positive for async function's params

### DIFF
--- a/tests/ui/shadow.rs
+++ b/tests/ui/shadow.rs
@@ -79,4 +79,10 @@ fn question_mark() -> Option<()> {
     None
 }
 
+pub async fn foo1(_a: i32) {}
+
+pub async fn foo2(_a: i32, _b: i64) {
+    let _b = _a;
+}
+
 fn main() {}
--- a/clippy_lints/src/shadow.rs
+++ b/clippy_lints/src/shadow.rs
@@ -105,11 +105,16 @@ impl<'tcx> LateLintPass<'tcx> for Shadow {
             PatKind::Binding(_, hir_id, ident, _) => (hir_id, ident),
             _ => return,
         };
+
+        if pat.span.desugaring_kind().is_some() {
+            return;
+        }
+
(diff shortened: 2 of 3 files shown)


tests/ui/shadow.rs
```diff
@@ -79,4 +79,10 @@ fn question_mark() -> Option<()> {
     None
 }
 
+pub async fn foo1(_a: i32) {}
+
+pub async fn foo2(_a: i32, _b: i64) {
+    let _b = _a;
```
giraffate, line 85: I think this should be warned in `shadow_unrelated`.

giraffate (commented):
It might be simpler to check `DesugaringKind`. This may help us:
https://github.com/rust-lang/rust-clippy/pull/7174.

surechen:
> It might be simpler to check `DesugaringKind`. This may help us: #7174.
Thank you very much.  By checking [DesugaringKind] is more concise and more accurate than my method.

giraffate:
@bors r+

It looks good, thanks!

Issue #8004, opened 2021-11-20
suboptimal_flops in const functions


leonardo-m:
On this code:

```rust
#![feature(const_fn_floating_point_arithmetic)]
#![allow(dead_code)]

#![warn(clippy::nursery)]

const fn mul_add(a: f32, b: f32, c: f32) -> f32 { a * b + c }

fn main() {}
```

Clippy gives me:

```
Checking foo1 v0.1.0 (...\foo1)
warning: multiply and add expressions can be calculated more efficiently and accurately
 --> src\main.rs:6:51
  |
6 | const fn mul_add(a: f32, b: f32, c: f32) -> f32 { a * b + c }
  |                                                   ^^^^^^^^^ help: consider using: `a.mul_add(b, c)`
  |
note: the lint level is defined here
 --> src\main.rs:4:9
  |
4 | #![warn(clippy::nursery)]
  |         ^^^^^^^^^^^^^^^
  = note: `#[warn(clippy::suboptimal_flops)]` implied by `#[warn(clippy::nursery)]`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suboptimal_flops
```

But currently mul_add doesn't work in const functions.

PR #8009, opened 2021-11-20
Allow `suboptimal_flops` in const functions

xFrednet:
This PR allows `clippy::suboptimal_flops` in constant functions. The check also effects the `clippy::imprecise_flops` lint logic. However, this doesn't have any effects as all functions checked for are not const and can therefore not be found in such functions. 

---

changelog: [`suboptimal_flops`]: No longer triggers in constant functions

Closes: rust-lang/rust-clippy#8004


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8011, opened 2021-11-20
Avoid inline hints with double backticks for `doc-markdown`

### LINKED ISSUE #8002: doc_markdown: suggestion looks like double backticks needed
For example:
```
error: item in documentation is missing backticks
   --> src/tcp.rs:135:58
    |
135 |     /// closed immediately.  The route can be added from TwinCAT, or this
    |                                                          ^^^^^^^ help: try: ``TwinCAT``
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown
```

Since all "try" hints are already enclosed in backticks, we end up with double backticks in the suggestion, which looks like it could be actual markup (especially if you also use ReST).

I suggest to omit the backticks in the hint, leading to ```try: `TwinCAT` ``` which together with the error text makes it obvious what is meant.

birkenfeld:
The easiest route here was to ensure that the suggestion is always shown on
its own line, where no additional backticks are added by the diagnostic formatter.

Fixes #8002 

---

changelog: Avoid inline hints with double backticks for `doc-markdown`

### DIFF
--- a/clippy_lints/src/doc.rs
+++ b/clippy_lints/src/doc.rs
@@ -1,5 +1,5 @@
 use clippy_utils::attrs::is_doc_hidden;
-use clippy_utils::diagnostics::{span_lint, span_lint_and_help, span_lint_and_note, span_lint_and_sugg};
+use clippy_utils::diagnostics::{span_lint, span_lint_and_help, span_lint_and_note, span_lint_and_then};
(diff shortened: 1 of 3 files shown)


xFrednet (approved):

xFrednet:
Using `SuggestionStyle::ShowAlways` is also a very nice solution, good catch! Thank you for the quick fix!

@bors r+

PR #8030, opened 2021-11-25
Ignore associated types in traits when considering type complexity

### LINKED ISSUE #1013: Type complexity should not be considered when defining an associated type
Consider this example:

``` rust
use std::vec::IntoIter;
use std::iter::{Map, Filter};

struct S;

impl IntoIterator for S {
    type Item = i32;
    type IntoIter = Filter<Map<IntoIter<i32>, fn(i32) -> i32>, fn(&i32) -> bool>;

        fn m(a: i32) -> i32 { a }
        fn p(_: &i32) -> bool { true }
        vec![1i32, 2, 3].into_iter().map(m as fn(_) -> _).filter(p)
    }
}
```

Clippy warns that `Filter<Map<IntoIter<i32>, fn(i32) -> i32>, fn(&i32) -> bool>` is a very complex type and suggest creating a `type` definition, but it already is type definition.

WaffleLapkin:
changelog: Ignore associated types in traits when checking ``[`type_complexity`]`` lint.

fixes #1013


tests/ui/type_complexity_issue_1013.rs
```diff
@@ -0,0 +1,23 @@
...
+
+    fn into_iter(self) -> Self::IntoIter {
+        fn m(a: i32) -> i32 {
+            a
+        }
+        fn p(_: &i32) -> bool {
+            true
+        }
+        vec![1i32, 2, 3].into_iter().map(m as fn(_) -> _).filter(p)
+    }
```
flip1995, line 20: Is this "complicated" implementation necessary to test the lint? I think you could also implement the test without using `std` traits, but rather write a "dummy" trait yourself, with just one assoc type in it. 

---

One a related note: Please put this test into the test file that already exists for this lint.

WaffleLapkin: I've reused the dummy trait that was already present in the test 👍🏻

tests/ui/type_complexity.stderr
```diff
@@ -73,22 +73,34 @@ LL |     fn def_method(&self, p: Vec<Vec<Box<(u32, u32, u32, u32)>>>) {}
...
+   |
+LL |     const A: Vec<Vec<Box<(u32, u32, u32, u32)>>> = vec![];
+   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: very complex type used. Consider factoring parts into `type` definitions
+  --> $DIR/type_complexity.rs:38:25
+   |
+LL |     fn method(&self, p: Vec<Vec<Box<(u32, u32, u32, u32)>>>) {}
+   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
WaffleLapkin, line 86: Should these warnings also not be produced? These types are set by the trait definition (and warnings are produced for them there), and they can use local generics/assoc types which makes them hard to simplify.

llogiq: So you suggest allowing the lint on trait impls? Yes, that likely makes sense. If the trait is local to the crate, it will already get a warning, so the additional warnings don't have much value.

flip1995: Yes not linting in trait implementations makes sense here, I think. There's a small exception though:

```rust
trait T {
    type A;
    fn f(a: Self::A); // type of `a` defined with the assoc type
}

impl T for _ {
    type A = Really<Complex<Type<Thingy>>>; // should *not* get linted as assoc type def
    fn f(a: Really<Complex<Type<Thingy>>>) { ... } // should get linted, since Self::A could be used like in the trait definition
}
```

So if a complex type is used in the impl, where the trait is using an assoc type (or more general, a type that wouldn't get linted), then this lint should trigger in the trait impl and suggest to use the type that is used in the trait def.

But I know from the `use_self` lint that this is hard to detect and even harder to produce a suggestion for. So nothing that would've to be done in this PR, but maybe worth an issue, if we stop linting on trait impls.

llogiq:
Any updates on this? Should I have a look now or are you still trying to offer suggestions for associated types?

WaffleLapkin:
Ah, sorry, I forget that I had this PR 😅 

I'll remove warnings for trait impls leaving a `FIXME` saying that we should ideally lint them if they can be simplified. I'm going to leave a `FIXME` instead of actually fixing this because it seems hard (https://github.com/rust-lang/rust-clippy/pull/8030#discussion_r756808646) and I don't think I have the capacity for this rn.

WaffleLapkin:
I'm not sure if I did everything right, but it seems to work :D

This is ready for review.

llogiq:
Thank you! @bors r+

PR #8032, opened 2021-11-25
Use new template forms for issue templates

flip1995:
GitHub has a new way to define issue templates. This gives an even more consistent formatting of our issues.

You can find (read-only) rendered versions [here](https://github.com/flip1995/rust-clippy/tree/new_templates/.github/ISSUE_TEMPLATE). If you want to try to fill out this kind of issue, you can do this in the [`cargo` Repo](https://github.com/rust-lang/cargo/issues/new/choose).

changelog: Better issue templates


flip1995:
Of course there is still the question if we want to do this. I'll bring it up in the meeting on Tuesday.

xFrednet:
Very fancy, I like this a lot! :+1:

camsteffen:
I'd like to see error output for false positive and lint name (optional) for new lint. But that's orthogonal to using the fancy new forms, so I think this is good!

.github/ISSUE_TEMPLATE/new_lint.yml
```diff
@@ -0,0 +1,65 @@
...
+      description: What does this lint do?
+    validations:
+      required: true
+  - type: dropdown
+    id: category
+    attributes:
+      label: Category
+      description: >
+        What category should this lint go into? If you're unsure you can select
+        multiple categories.
```
camsteffen, line 21: Maybe you could say to see the category descriptions in the readme.

.github/ISSUE_TEMPLATE/new_lint.yml
```diff
@@ -0,0 +1,65 @@
...
+    attributes:
+      label: Drawbacks
+      description: What might be possible drawbacks of such a lint?
+  - type: textarea
+    id: example
+    attributes:
+      label: Example
+      description: >
+        Include a backtrace in the code block by setting `RUST_BACKTRACE=1` in
+        your environment. E.g. `RUST_BACKTRACE=1 cargo clippy`.
```
camsteffen, line 53: Copy paste error?

flip1995:
> I'd like to see error output for false positive

I modified the default value of the text area. I think adding an additional section for just an error output would be overkill.

camsteffen (approved):
r=me after squash. Thanks!

flip1995:
Rebased+Squashed. Let's get this merged.

@bors r=camsteffen

Issue #8013, opened 2021-11-21
non_ascii_literal doesn't catch `char` literals in single-quotes


zedseven:
Lint name: `non_ascii_literal`

I apologize if this has been reported before or this is the wrong issue category. I checked for pre-existing issues and this category seemed to fit best.

I tried this code:

```rust
#![warn(clippy::non_ascii_literal)]
// ...
fn main() {
  const EMPTY_BLOCK: char = '▱';
  const FULL_BLOCK: char = '▰';
}
```

This could be replaced with:

```rust
#![warn(clippy::non_ascii_literal)]
// ...
fn main() {
  const EMPTY_BLOCK: char = '\u{25b1}';
  const FULL_BLOCK: char = '\u{25b0}';
}
```

I expected to see Clippy warn about the non-ASCII literal within the single-quotes.

Instead, Clippy allowed it. I suspect Clippy only checks for non-ASCII characters within string literals, but that's not really what the name of the lint communicates and I think it would be helpful to check within single-quotes as well.

After changing the constants to single-character strings, Clippy recognized the characters and warned as expected.

If this improvement is made, the *What it does* section in the [lint documentation](https://rust-lang.github.io/rust-clippy/master/index.html#non_ascii_literal) will need to be updated. At the moment it states: "Checks for non-ASCII characters in string literals."

### Meta

**Rust version (`rustc -Vv`):**

```
rustc 1.58.0-nightly (c9c4b5d72 2021-11-17)
binary: rustc
commit-hash: c9c4b5d7276297679387189d96a952f2b760e7ad
commit-date: 2021-11-17
host: x86_64-pc-windows-gnu
release: 1.58.0-nightly
LLVM version: 13.0.0
```

PR #8034, opened 2021-11-26
Non ascii in sigle literal

togami2864:
fix: #8013 
changelog: `non_ascii_literal` warn against non-ASCII within `char`, not just `strings`


clippy_lints/src/unicode.rs
```diff
@@ -28,7 +28,7 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for non-ASCII characters in string literals.
+    /// Checks for non-ASCII characters in string literals or char.
```
llogiq, line 31:
```suggestion
    /// Checks for non-ASCII characters in string and char literals.
```

llogiq (approved):
Just a small wording nit, otherwise this looks good. Perhaps you want to rename `check_str` to `check_str_or_char` or `check_non_ascii` or something, but if not, that's OK, too.

llogiq:
Thank you! @bors r+

PR #8042, opened 2021-11-29
Peel blocks and statements utils

camsteffen:
changelog: none

* Rename `remove_blocks` to `peel_blocks`
* Add `peel_blocks_and_stmts`
* Various refactors to use the above utils
* The utils also now check `block.rules`

### DIFF
--- a/clippy_lints/src/loops/manual_flatten.rs
+++ b/clippy_lints/src/loops/manual_flatten.rs
@@ -3,11 +3,11 @@ use super::MANUAL_FLATTEN;
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::higher;
 use clippy_utils::visitors::is_local_used;
-use clippy_utils::{is_lang_ctor, path_to_local_id};
+use clippy_utils::{is_lang_ctor, path_to_local_id, peel_blocks_with_stmt};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::LangItem::{OptionSome, ResultOk};
-use rustc_hir::{Expr, ExprKind, Pat, PatKind, StmtKind};
+use rustc_hir::{Expr, Pat, PatKind};
 use rustc_lint::LateContext;
 use rustc_middle::ty;
 use rustc_span::source_map::Span;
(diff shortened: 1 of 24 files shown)


clippy_utils/src/lib.rs
```diff
@@ -1224,6 +1224,57 @@ pub fn get_parent_as_impl(tcx: TyCtxt<'_>, id: HirId) -> Option<&Impl<'_>> {
...
+    {
+        expr = inner;
+    }
+    expr
+}
+
+/// Remove blocks around an expression or single statement.
+///
+/// Example: `x`, `{ x }`, `{{ x }}` or `{{ x; }}` all yield `x` but `{ x; y }` yields itself.
+pub fn peel_blocks_and_stmts<'a>(mut expr: &'a Expr<'a>) -> &'a Expr<'a> {
```
flip1995, line 1250: I first reviewed the fallout fixes and after that this function. From reading the function name without looking at the function, I would have expected that this removes all blocks and also all statements, and returning the `expr` if available.

Maybe `peel_blocks_around_stmt`. Definitely the plural of `stmts` is confusing here.

camsteffen: I was also wondering about it so I'm glad you raised this. How about `peel_stmt_blocks`? Or `peel_blocks_and_stmt`, `peel_blocks_with_stmt`.

flip1995: Of those three I like `peel_blocks_with_stmt` the most. I find it really hard to come up with a short and telling name here...

flip1995: Or `peel_blocks_from_stmt`?

camsteffen: Ideally the name would convey "blocks having one expression with or without a statement" instead of "blocks with one statement". It is certainly tricky. I'm now leaning to `peel_blocks_with_stmt`.

clippy_utils/src/lib.rs
```diff
@@ -1224,6 +1224,57 @@ pub fn get_parent_as_impl(tcx: TyCtxt<'_>, id: HirId) -> Option<&Impl<'_>> {
     }
 }
 
+/// Remove blocks around an expression.
+///
+/// Ie. `x`, `{ x }` and `{{{{ x }}}}` all give `x`. `{ x; y }` and `{}` return
+/// themselves.
```
xFrednet, line 1230: The implementation will also halt at blocks which contain a single statement. I think it's worth to mention this and provide an example. Also, to make clear what distinguishes this function from the other one. It would also be good to mention that this only peels normal blocks.

Maybe a table or something might be better than a sentence for the examples, even if this is still not perfect.

```rs
/// | Input           | Output        |
/// |-----------------|---------------|
/// | `x`             | `x`           |
/// | `{ x }`         | `x`           |
/// | `{{{{ x }}}}`   | `x`           |
/// | `{}`            | `{}`          |
/// | `{ y; x }`      | `{ y; x }`    |
/// | `{ y; }`        | `{ y; }`      |
/// | `{ {x}; }`      | `{ {x}; }`    |
/// | `{ unsafe {x} }`| `unsafe {x} ` |
```

clippy_lints/src/loops/manual_flatten.rs
```diff
@@ -21,71 +21,54 @@ pub(super) fn check<'tcx>(
...
+        if let Some(higher::IfLet { let_pat, let_expr, if_then, if_else: None })
+            = higher::IfLet::hir(cx, inner_expr);
+        // Ensure match_expr in `if let` statement is the same as the pat from the for-loop
+        if let PatKind::Binding(_, pat_hir_id, _, _) = pat.kind;
+        if path_to_local_id(let_expr, pat_hir_id);
+        // Ensure the `if let` statement is for the `Some` variant of `Option` or the `Ok` variant of `Result`
+        if let PatKind::TupleStruct(ref qpath, _, _) = let_pat.kind;
+        let some_ctor = is_lang_ctor(cx, qpath, OptionSome);
+        let ok_ctor = is_lang_ctor(cx, qpath, ResultOk);
+        if some_ctor || ok_ctor;// Ensure epxr in `if let` is not used afterwards
```
xFrednet, line 35: NIT in case you have to change something else:
```suggestion
        if some_ctor || ok_ctor; // Ensure epxr in `if let` is not used afterwards
```

clippy_lints/src/needless_bool.rs
```diff
@@ -331,41 +326,20 @@ fn suggest_bool_comparison<'a, 'tcx>(
...
-            if let LitKind::Bool(value) = lit_ptr.node {
-                Expression::Bool(value)
-            } else {
-                Expression::Other
-            }
-        },
-        ExprKind::Ret(Some(expr)) => match fetch_bool_expr(expr) {
-            Expression::Bool(value) => Expression::RetBool(value),
-            _ => Expression::Other,
-        },
```
xFrednet, line 368: This previously handled return statements, which the new implementation doesn't. Is this an intentional behavior change?

clippy_utils/src/lib.rs
xFrednet: I think the name should better reflect the difference between this and the other until function. `peel_blocks_with_stmt` sounds to me like it would ignore all statements and only return the innermost expression while ignoring everything else. What do you think about `extract_expr_or_stmt` with the upper being called `extract_expr`? Or if we want to keep `peel` in the name, maybe `peel_all_blocks`?

I'm not quite sold on these names, but I feel like they capture the behavior a bit better :upside_down_face:

xFrednet:
The other changes look good to me, it's nice to remove some duplicate code! Good work :upside_down_face:

clippy_utils/src/lib.rs
camsteffen: It sounds like you had the same misconception as flip1995, unfortunately.

I think "block" is quite important - removing blocks is the primary function of the...function.

The word "peel" is consistent with some other methods like `Ty::peel_refs` and `Expr::peel_drop_temps`. It's not crucial, but I like how short it is.

Here is my thought process: to me the word "peel" means remove a layer and get *everything* inside of it. It doesn't mean remove a layer and then pick out some unspecified piece of what's inside. So I assume that `peel_blocks` returns everything inside the block, or else it will not "peel". And since it is `Expr -> Expr`, I assume it's looking for blocks containing a single expr. Now I admit that those are pretty hefty assumptions. But we have to choose *some* assumptions if we want a function name that is reasonably short. Documentation fills in the gaps. Honestly I think this is one of those cases where you're just not going know until you go read the docs.

`extract_expr` lends to the misconception you had even more strongly IMO, since "extract" conveys taking a part from the whole.

`peel_blocks` is already plural, so `peel_all_blocks` doesn't add any information.

So I'm currently liking `peel_blocks` and `peel_blocks_with_stmt`, but certainly open to discuss.

clippy_lints/src/needless_bool.rs
camsteffen: It shouldn't be a behavior change if you look at how the function is used. Return statements are still found in `fetch_bool_block`. In the other usage of this function, returns were being ignored.

xFrednet: Ahh, okay now I see, the control flow confused me at first :)

clippy_utils/src/lib.rs
xFrednet: > It sounds like you had the same misconception as flip1995, unfortunately.

I guessed as much from the previous discussion. However, I think that this shows that the original name was somewhat confusing.

Let's take `peel_blocks_with_stmt` for now, the more I think about it the more I like it :upside_down_face:

camsteffen:
Rebased, renamed to `peel_blocks_with_stmt`, and improved docs. Let me know if you think the docs are clear enough.

clippy_utils/src/lib.rs
```diff
@@ -1224,6 +1224,68 @@ pub fn get_parent_as_impl(tcx: TyCtxt<'_>, id: HirId) -> Option<&Impl<'_>> {
...
 
+/// Removes blocks around an expression, only if the block contains just one expression
+/// and no statements.
+///
+/// Examples:
+///  * `{}`       -> `{}`
+///  * `{ x }`    -> `x`
+///  * `{{ x }}`  -> `x`
+///  * `{ x; }`   -> `{ x; }`
+///  * `{ x; y }` -> `{ x; y }`
```
xFrednet, line 1235: I would like one more example that shows that `unsafe` blocks will not be peeled. The same for the other function.


```rs
///  * `{ unsafe { x } }` -> `unsafe { x }`
```

xFrednet (approved):
LGTM, thank you for the changes! I would like one small change related to the documentation, then you can `r=me` :upside_down_face:

camsteffen:
Thanks for the review! @bors r=xFrednet

camsteffen:
@bors r+

(sorry, I accidentally had thrown in unrelated commits)

xFrednet:
That was a rollercoaster of emotions xD

PR #8049, opened 2021-12-01
Add test for pattern_type_mismatch.

### LINKED ISSUE #7946: `pattern_type_mismatch` false positive when `&str` is matched against a `const _: &str`
Lint name: `pattern_type_mismatch`

I tried this code ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3a64e0312d5d6be65ffd24ecd00bb284)):

```rust
#![deny(clippy::pattern_type_mismatch)]

const FOO: &str = "foo";

fn foo(s: &str) -> i32 {
    match s {
        FOO => 1,
        _ => 0,
    }
}
```

I expected to see this happen: the code should pass the lint, the type of the pattern is exactly the same as the type of the discriminant expression.

Instead, this happened: I got the "type of pattern does not match the expression type" error message.

### Meta

**Clippy version:** 0.1.58 (2021-11-07 46b8e74)
**Rust version:** 1.56.1 stable (on the Playground)

surechen:
This issue has been fixed by [commit](https://github.com/rust-lang/rust-clippy/commit/8c1c763c2d5897c66f52bd5e7c16f48ae66c894c)
This PR is used for close #7946(Fixes #7946).

changelog: Add test for pattern_type_mismatch.

### DIFF
--- a/tests/ui/pattern_type_mismatch/mutability.rs
+++ b/tests/ui/pattern_type_mismatch/mutability.rs
@@ -37,4 +37,13 @@ fn should_not_lint() {
         Some(_) => (),
         _ => (),
     }
+
+    const FOO: &str = "foo";
+
+    fn foo(s: &str) -> i32 {
+        match s {
+            FOO => 1,
+            _ => 0,
+        }
+    }
 }


flip1995:
@bors r+

Thanks!

PR #8066, opened 2021-12-02
Parenthesize blocks in `needless_bool` suggestion

### LINKED ISSUE #8052: Clippy emitting wrong lint for unsafe
### Summary

```
unsafe fn no(arg: i32) -> i32 {
    arg
}

---

warning: this if-then-else expression returns a bool literal
  --> src\main.rs:6:5
   |
6  | /     if unsafe { no(4) } & 1 != 0 {
8  | |     } else {
9  | |         false
10 | |     }
   | |_____^ help: you can reduce it to: `unsafe { no(4) } & 1 != 0
```

The produced expression above is not valid, the `& 1 != 0` should be inside the `unsafe` block.

### Reproducer

I tried this code:

```rust
unsafe fn no(arg: i32) -> i32 {
    arg
}

fn state() -> bool {
    if unsafe { no(4) } & 1 != 0 {
        true
    } else {
        false
    }
}

fn main() {
    println!("{}", if state() { "hello" } else { "bye" });
}
```

Upon running `cargo clippy` I'd expect it to reduce state to `unsafe { no(4) & 1 != 0 }` 

Instead, this happened:
It reduced state to `unsafe { no(4) } & 1 != 0 ` which is invalid syntax

### Version

```text
rustc 1.59.0-nightly (686e313a9 2021-11-27)
binary: rustc
commit-hash: 686e313a9aa14107c8631ffe48fa09110a7692db
commit-date: 2021-11-27
host: i686-pc-windows-msvc
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

llogiq:
Because the `if .. {}` statement already puts the condition in expression scope, contained blocks would be parsed as complete
statements, so any `&` binary expression whose left operand ended in a block would lead to a non-compiling suggestion.

We identify such expressions and add parentheses. Note that we don't make a difference between normal and unsafe blocks because the parsing problems are the same for both.

This fixes #8052.

---

changelog: none

### DIFF
--- a/clippy_lints/src/needless_bool.rs
+++ b/clippy_lints/src/needless_bool.rs
@@ -9,8 +9,10 @@ use clippy_utils::sugg::Sugg;
 use clippy_utils::{is_else_clause, is_expn_of};
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
+use rustc_hir::intravisit::{walk_expr, NestedVisitorMap, Visitor};
 use rustc_hir::{BinOpKind, Block, Expr, ExprKind, StmtKind, UnOp};
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::hir::map::Map;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::source_map::Spanned;
 use rustc_span::Span;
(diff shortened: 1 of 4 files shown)


camsteffen (changes requested):
In the test you added, I don't think parentheses are needed since the expression is not moved to statement position. So I think we should check if the parent of the `if` node is a statement.

We can take inspiration from rustc [here](https://github.com/rust-lang/rust/blob/2a9e0831d6603d87220cedd1b1293e2eb82ef55c/compiler/rustc_lint/src/unused.rs#L460-L475) and [here](https://github.com/rust-lang/rust/blob/c2a408840ad18f74280805535f0b7193528ff3df/compiler/rustc_ast/src/util/classify.rs#L14-L25): I would avoid using a visitor since you could easily skip into where parentheses are not needed. There are more types of expressions in addition to blocks where parentheses are needed.

Maybe add utils like `expr_needs_parens_to_be_stmt` and `expr_is_stmt`.

llogiq:
If you remove the parentheses, the parser thinks that the block is one statement and the `&` is a borrow instead of a bitwise and, then you get a compiler error.

As I wrote, the `if` parses the condition as an expression, where a plain statement won't.

camsteffen:
> If you remove the parentheses, the parser thinks that the block is one statement

Only if that block is at the beginning of a statement. See [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=933d8ccaa1ac2c9d03191b5975ebabf9).

llogiq:
Ah, I see. So the correct check is whether the `if` is in the first statement of the parent block and the condition, peeling all binops, starts with a block?

camsteffen:
> whether the `if` is in the first statement of the parent block

Not necessarily the first statement, but the direct child of any statement.

> peeling all binops

See the first linked rustc snippet. Also peel `ExprKind::Call` and other things that have a left and right side.

> starts with a block?

See the second linked rustc snippet. Not only a block, but `match`, `loop` and other things that have curly braces.

llogiq:
@camsteffen like this?

camsteffen:
Still need a check for whether `if` is a direct child of a statement. I still don't think the test you added should have parentheses as written.

llogiq:
You're right, I think. I've added the check now. However, there's something wrong with this – the fixed version fails to compile.

camsteffen:
Looks good! Thanks for the updates.

@bors r+

Issue #8060, opened 2021-12-02
clippy --fix fails to escape backslash when replacing r"\" into '\\' E0762


frobiac:
### Summary

`clippy --fix` will not escape a backslash when replacing it in a single-character string constant.
                  

### Reproducer

Minimal example:

```rust
// original
let s = r#"abc\xyz/"#.find(r"\");

// This should be fixed into
let s = r#"abc\xyz/"#.find('\\');
```

But clippy did leave out the escaping backslash.

```shell
The following errors were reported:
error[E0762]: unterminated character literal
 --> src/main.rs:2:32
  |
2 |     let s = r#"abc\xyz/"#.find('\');
  |                                ^^^^^
```

### Full log

```shell
warning: failed to automatically apply fixes suggested by rustc to crate `clippy_err`

after fixes were automatically applied the compiler reported errors within these files:

  * src/main.rs

This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:
error[E0762]: unterminated character literal
 --> src/main.rs:2:32
  |
2 |     let s = r#"abc\xyz/"#.find('\');
  |                                ^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0762`.
Original diagnostics will follow.

warning: single-character string constant used as pattern
 --> src/main.rs:2:32
  |
2 |     let s = r#"abc\xyz/"#.find(r"\");
  |                                ^^^^ help: try using a `char` instead: `'\'`
  |
  = note: `#[warn(clippy::single_char_pattern)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern

```

### Version

```text
rustc 1.56.1 (59eed8a2a 2021-11-01)
binary: rustc
commit-hash: 59eed8a2aac0230a8b53e89d4e99d55912ba6b35
commit-date: 2021-11-01
host: x86_64-unknown-linux-gnu
release: 1.56.1
LLVM version: 13.0.0


clippy 0.1.56 (59eed8a 2021-11-01)
```

frobiac:
I'll give it a go...

frobiac:
[...]

PR #8067, opened 2021-12-02
Escape backslash in single_char_pattern.rs

frobiac:
Escape backslash in single_char_pattern.

Previously, the proposed clippy fix for a single backslash in raw strings ```r"\"``` or ```r#"\"#``` was also only an unescaped, *single* ```'\'```:
```shell
warning: single-character string constant used as pattern
2 |     let s = r#"abc\xyz/"#.find(r"\");
  |                                ^^^^ help: try using a `char` instead: `'\'`
  |
  = note: `#[warn(clippy::single_char_pattern)]` on by default
```

This PR corrects this to a properly escaped *double* backslash ```'\\'```.
I haven't come up with any other problematic cases, a single quote was already handled.

Closes: #8060 

changelog:  Escape backslash in ``[`single_char_pattern`]``

### DIFF
--- a/clippy_lints/src/methods/utils.rs
+++ b/clippy_lints/src/methods/utils.rs
@@ -66,7 +66,13 @@ pub(super) fn get_hint_if_single_char_arg(
                 // for regular string: "a"
                 &snip[1..(snip.len() - 1)]
             };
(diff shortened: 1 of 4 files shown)


giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #8071, opened 2021-12-03
Add new lint to warn when #[must_use] attribute should be used on a method

GuillaumeGomez:
This lint is somewhat similar to https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate but also different: it emits a warning by default and only targets methods (so not functions nor associated functions).

Someone suggested it to me after this tweet: https://twitter.com/m_ou_se/status/1466439813230477312

I think it would reduce the number of cases of API misuses quite a lot.

What do you think?

---

changelog: Added new [`return_self_not_must_use`] lint

### DIFF
--- a/clippy_lints/src/return_self_not_must_use.rs
+++ b/clippy_lints/src/return_self_not_must_use.rs
@@ -1,0 +1,105 @@
+use clippy_utils::{diagnostics::span_lint, must_use_attr, nth_arg, return_ty};
+use rustc_hir::def_id::LocalDefId;
+use rustc_hir::intravisit::FnKind;
+use rustc_hir::{Body, FnDecl, HirId, TraitItem, TraitItemKind};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint warns when a method returning `Self` doesn't have the `#[must_use]` attribute.
+    ///
+    /// ### Why is this bad?
+    /// It prevents to "forget" to use the newly created value.
+    ///
+    /// ### Limitations
+    /// This lint is only applied on methods taking a `self` argument. It would be mostly noise
+    /// if it was added on constructors for example.
+    ///
+    /// ### Example
+    /// ```rust
+    /// pub struct Bar;
+    ///
+    /// impl Bar {
+    ///     // Bad
+    ///     pub fn bar(&self) -> Self {
+    ///         Self
+    ///     }
+    ///
+    ///     // Good
+    ///     #[must_use]
+    ///     pub fn foo(&self) -> Self {
+    ///         Self
+    ///     }
+    /// }
+    /// ```
+    #[clippy::version = "1.59.0"]
+    pub RETURN_SELF_NOT_MUST_USE,
+    suspicious,
+    "missing `#[must_use]` annotation on a method returning `Self`"
+}
+
+declare_lint_pass!(ReturnSelfNotMustUse => [RETURN_SELF_NOT_MUST_USE]);
+
+fn check_method(cx: &LateContext<'tcx>, decl: &'tcx FnDecl<'tcx>, fn_def: LocalDefId, span: Span, hir_id: HirId) {
+    if_chain! {
+        // If it comes from an external macro, better ignore it.
+        if !in_external_macro(cx.sess(), span);
+        if decl.implicit_self.has_implicit_self();
+        // We only show this warning for public exported methods.
+        if cx.access_levels.is_exported(fn_def);
+        if cx.tcx.visibility(fn_def.to_def_id()).is_public();
+        // No need to warn if the attribute is already present.
+        if must_use_attr(cx.tcx.hir().attrs(hir_id)).is_none();
+        let ret_ty = return_ty(cx, hir_id);
+        let self_arg = nth_arg(cx, hir_id, 0);
+        // If `Self` has the same type as the returned type, then we want to warn.
+        //
+        // For this check, we don't want to remove the reference on the returned type because if
+        // there is one, we shouldn't emit a warning!
+        if self_arg.peel_refs() == ret_ty;
+
+        then {
+            span_lint(
+                cx,
+                RETURN_SELF_NOT_MUST_USE,
+                span,
+                "missing `#[must_use]` attribute on a method returning `Self`",
+            );
+        }
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for ReturnSelfNotMustUse {
+    fn check_fn(
+        &mut self,
+        cx: &LateContext<'tcx>,
+        kind: FnKind<'tcx>,
+        decl: &'tcx FnDecl<'tcx>,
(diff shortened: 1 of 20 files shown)


clippy_lints/src/must_use_self_return.rs
```diff
@@ -0,0 +1,78 @@
...
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint warns when a method returning `Self` doesn't have the `#[must_use]` attribute.
+    ///
+    /// ### Why is this bad?
+    /// It prevents to "forget" to use the newly created type.
```
sdroege, line 13:
```suggestion
    /// It prevents to "forget" to use the newly created value.
```

GuillaumeGomez: Thanks!

clippy_lints/src/must_use_self_return.rs
```diff
@@ -0,0 +1,78 @@
...
+        &mut self,
+        cx: &LateContext<'tcx>,
+        kind: FnKind<'tcx>,
+        decl: &'tcx FnDecl<'tcx>,
+        _: &'tcx Body<'tcx>,
+        span: Span,
+        hir_id: HirId,
+    ) {
+        if_chain! {
+            // We are only interested in methods, not in functions or associated functions.
```
sdroege, line 51: Why not? What about `fn new() -> Self`, `fn new_foo(bar: Baz) -> Self` or `fn from_blurb(blurb: Blurb) -> Self`?

tests/ui/must_use_self_return.rs
```diff
@@ -0,0 +1,42 @@
+#![deny(clippy::must_use_self_return)]
+#![crate_type = "lib"]
+
+pub struct Bar;
+
+pub trait Whatever {
+    // There should be no warning here!
```
sdroege, line 7: Why not? That's basically like the `Clone` trait :)

GuillaumeGomez: Because I only want it on the implementation, not on the "declaration".

sdroege: I think the attribute only has an effect on the declaration in traits, not on implementations. Or am I confused?

clippy_lints/src/must_use_self_return.rs
GuillaumeGomez: They look like different use cases to me. What I'm trying to prevent here is cases like `0f32.rad(2.)`. It's much less likely (in my opinion) if it's not a method.

tests/ui/must_use_self_return.rs
GuillaumeGomez: No, you are perfectly right, I'll change this condition then.

clippy_lints/src/must_use_self_return.rs
sdroege: Ah sure, if you don't want to cover that case here that makes sense. I also think it's quite unlikely that someone calls such a constructor function and forgets to actually do something with the result. Adding `#[must_use]` on all these might add quite a bit of noise to the code.

GuillaumeGomez:
Updated with @sdroege's suggestions.

camsteffen:
I'd call it `return_self_not_must_use` so that it makes sense with `allow`.

For the category I'd do `suspicious` or maybe `pedantic`, depending on the supposed false positive rate. I'm not really sure if there will be false positives, but it's possible there are methods that you can call for a side effect and drop the returned value.

GuillaumeGomez:
@camsteffen This is indeed much better, I renamed the lint.

GuillaumeGomez:
And updated the category to pedantic.

clippy_lints/src/return_self_not_must_use.rs
```diff
@@ -0,0 +1,101 @@
...
+use rustc_hir::intravisit::FnKind;
+use rustc_hir::{Body, FnDecl, HirId, TraitItem, TraitItemKind};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint warns when a method returning `Self` doesn't have the `#[must_use]` attribute.
```
xFrednet, line 12: > They look like different use cases to me. What I'm trying to prevent here is cases like 0f32.rad(2.). It's much less likely (in my opinion) if it's not a method.
> ~ `@GuillaumeGomez`
>
> Ah sure, if you don't want to cover that case here that makes sense. I also think it's quite unlikely that someone calls such a constructor function and forgets to actually do something with the result. Adding #[must_use] on all these might add quite a bit of noise to the code.
> ~`@sdroege`

(I copied parts of [the discussion](https://github.com/rust-lang/rust-clippy/pull/8071#discussion_r762047475), as I was unable to comment on it for some reason :thinking: )

I understand the reasoning behind this decision, and I'm also in favor of this solution. Could you mention this in the lint description as well, to avoid false negative reports? You can add a *Limitations* section before the example or mention here that only methods will be linted which take a `self` argument.

clippy_utils/src/lib.rs
```diff
@@ -1329,6 +1329,13 @@ pub fn return_ty<'tcx>(cx: &LateContext<'tcx>, fn_item: hir::HirId) -> Ty<'tcx>
     cx.tcx.erase_late_bound_regions(ret_ty)
 }
 
+/// Convenience function to get the nth argument of a function.
```
xFrednet, line 1332:
```suggestion
/// Convenience function to get the nth argument type of a function.
```

xFrednet (commented):
The implementation logic looks good to me. There are some minor improvements to be made, and then it should be ready to be merged. :+1: 

Also, a thank you to the ones who have given feedback already. I appreciate it! :upside_down_face:

clippy_lints/src/return_self_not_must_use.rs
GuillaumeGomez: Done!

tests/ui/auxiliary/option_helpers.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(dead_code, unused_variables)]
```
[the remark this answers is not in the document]
GuillaumeGomez: It should not be allowed by default. I guess it's `pedantic` which changed that... I'll find another one.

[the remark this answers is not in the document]
GuillaumeGomez: I moved it into "suspicious" and it's not allowed by default anymore. Thanks a lot, completely missed that. :)

GuillaumeGomez:
Updated with comments from @xFrednet. :)

xFrednet (approved):

xFrednet:
Everything looks good to me, thank you for the work!

I ran the lint on 25 crates with lintcheck and got about 60 reports, the once I've looked at seem to be correct. :upside_down_face: 

@bors r+

xFrednet:
Clippy created the changelog based on the lint description. I added the corresponding message :upside_down_face: 

@bors retry

GuillaumeGomez:
Thanks!

PR #8074, opened 2021-12-04
Consider NonNull as a pointer type

Qwaz:
PR 1/2 for issue #8045. Add `NonNull` as a pointer class to suppress false positives like `UnsafeCell<NonNull<()>>`. However, this change is not sufficient to handle the cases shared in gtk-rs and Rug in the issue.

changelog: none

r? @xFrednet

### DIFF
--- a/clippy_lints/src/non_send_fields_in_send_ty.rs
+++ b/clippy_lints/src/non_send_fields_in_send_ty.rs
@@ -1,11 +1,12 @@
 use clippy_utils::diagnostics::span_lint_and_then;
-use clippy_utils::is_lint_allowed;
 use clippy_utils::source::snippet;
 use clippy_utils::ty::{implements_trait, is_copy};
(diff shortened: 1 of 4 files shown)


Qwaz:
I tried and failed to avoid using hard-coded paths for `NonNull`. I couldn't find `NonNull` in diagnostic items. Is there a better way to handle that comparison?

xFrednet:
> I tried and failed to avoid using hard-coded paths for NonNull. I couldn't find NonNull in diagnostic items. Is there a better way to handle that comparison?

It sadly doesn't have another identifier besides that. We can add them in the rust repo, but that will take at least a week to be available here again. It's okay to have it like this until we go through them again and add them in rust :upside_down_face:

xFrednet (approved):

xFrednet:
The changes are looking good, thank you very much! :upside_down_face: 

@bors r+

PR #8075, opened 2021-12-04
Clarify the purpose of the non_send lint

Qwaz:
PR 2/2 for issue #8045. Tried to tone down the warning message and clarify the intention of the lint. Specifically, I added a description that this lint tries to detect "types that are not safe to be sent to another thread".

changelog: none

r? @xFrednet

### DIFF
--- a/clippy_lints/src/non_send_fields_in_send_ty.rs
+++ b/clippy_lints/src/non_send_fields_in_send_ty.rs
@@ -17,7 +17,7 @@ declare_clippy_lint! {
     /// contains fields that are not safe to be sent across threads.
     /// It tries to detect fields that can cause a soundness issue
     /// when sent to another thread (e.g., `Rc`) while allowing `!Send` fields
-    /// that are expected to exist in a `Send` type such as raw pointers.
+    /// that are expected to exist in a `Send` type, such as raw pointers.
     ///
     /// ### Why is this bad?
     /// Sending the struct to another thread effectively sends all of its fields,
@@ -124,7 +124,7 @@ impl<'tcx> LateLintPass<'tcx> for NonSendFieldInSendTy {
                         NON_SEND_FIELDS_IN_SEND_TY,
                         item.span,
                         &format!(
-                            "there are some fields in `{}` are not safe to be sent to another thread",
+                            "some fields in `{}` are not safe to be sent to another thread",
                             snippet(cx, hir_impl.self_ty.span, "Unknown")
                         ),
                         |diag| {
--- a/clippy_lints/src/non_send_fields_in_send_ty.rs
+++ b/clippy_lints/src/non_send_fields_in_send_ty.rs
@@ -13,24 +13,28 @@ use rustc_span::sym;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Warns about fields in struct implementing `Send` that are neither `Send` nor `Copy`.
+    /// This lint warns about a `Send` implementation for a type that
+    /// contains fields that are not safe to be sent across threads.
+    /// It tries to detect fields that can cause a soundness issue
+    /// when sent to another thread (e.g., `Rc`) while allowing `!Send` fields
+    /// that are expected to exist in a `Send` type such as raw pointers.
     ///
     /// ### Why is this bad?
-    /// Sending the struct to another thread will transfer the ownership to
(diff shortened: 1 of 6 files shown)


clippy_lints/src/non_send_fields_in_send_ty.rs
```diff
@@ -118,14 +122,14 @@ impl<'tcx> LateLintPass<'tcx> for NonSendFieldInSendTy {
                         NON_SEND_FIELDS_IN_SEND_TY,
                         item.span,
                         &format!(
-                            "this implementation is unsound, as some fields in `{}` are `!Send`",
+                            "there are some fields in `{}` are not safe to be sent to another thread",
```
xFrednet, line 125: > There are some fields in XYZ are not

Sounds wrong to me, I would suggest:

```suggestion
                            "some fields in `{}` are not safe to be sent to another thread",
```

or

```suggestion
                            "there are some fields in `{}` which are not safe to be sent to another thread",
```

clippy_lints/src/non_send_fields_in_send_ty.rs
```diff
@@ -118,14 +122,14 @@ impl<'tcx> LateLintPass<'tcx> for NonSendFieldInSendTy {
...
                             snippet(cx, hir_impl.self_ty.span, "Unknown")
                         ),
                         |diag| {
                             for field in non_send_fields {
                                 diag.span_note(
                                     field.def.span,
-                                    &format!("the type of field `{}` is `!Send`", field.def.ident.name),
+                                    &format!("it is not safe to send field `{}` to another thread", field.def.ident.name),
```
xFrednet, line 132: Here I liked the mentioning of the `Send` trait. What about:
```suggestion
                                    &format!("it is not safe to send field `{}` to another thread, as the type does not implement `Send`", field.def.ident.name),
```

xFrednet (commented):
Some minor adjustments, but the overall change looks good to me. :+1:  Your first PR created some merge conflicts. To resolve them, please rebase on the current master. Thank you for coming back to this lint :upside_down_face:

clippy_lints/src/non_send_fields_in_send_ty.rs
```diff
@@ -46,7 +50,7 @@ declare_clippy_lint! {
     #[clippy::version = "1.57.0"]
     pub NON_SEND_FIELDS_IN_SEND_TY,
     suspicious,
```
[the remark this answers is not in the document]
Qwaz: The original example in the issue would be solved by this PR, but not the others shared later in the thread. I'm especially concerned about the intermediate struct example shared from `gtk-rs` which I know to be common. I believe it can be solvable with new rules, but for the time being I think it's better to stay in `nursery` to not disrupt those use cases. If you think it can be done as part of this PR, I'll add another commit to adjust the category soon.

clippy_lints/src/non_send_fields_in_send_ty.rs
Qwaz: Omitting `Send` from the description was in fact intentional. `!Send` has subsets, (1) types that are fundamentally unsafe to send (e.g., `Rc`) and (2) types that are marked as `!Send` ["as a lint"](https://doc.rust-lang.org/nomicon/send-and-sync.html) such as pointers. The goal of this lint is to warn about the first subset and avoid raise warning for the second subset, and it seems that mentioning `Send` trait leads to unintended interpretation of the lint's purpose according to the people's response.

Qwaz:
Ugh let me try once more. I think rust-lang/triagebot#1449 needs more prioritization.
@bors ready

clippy_lints/src/non_send_fields_in_send_ty.rs
xFrednet: Ahh, okay, you're right :+1:

xFrednet (approved):

xFrednet:
> I think [rust-lang/triagebot#1449](https://github.com/rust-lang/triagebot/issues/1449) needs more prioritization.

Yes, I should probably also mention the comment to switch the labels before just assigning them, thanks for pointing that out. The syntax was correct, except that the pinged bot should have been `@rustbot`. I'll take care to indicate how the label can be switched from now on, maybe we could also adjust the label description :thinking: 

The changes look good to me, thank you for coming back to the lint and making the description more detailed!

@bors r+

PR #8077, opened 2021-12-05
Fix some false negatives for [`single_char_pattern`]

nixxquality:

changelog: Fix some false negatives for [`single_char_pattern`]

I noticed that clippy wasn't complaining about my usage of `split_once("x")` in a personal project so I updated the list of functions.

I had to update the test case for an unrelated issue because replace is now included in the list of functions to be linted.

### DIFF
--- a/clippy_lints/src/map_unit_fn.rs
+++ b/clippy_lints/src/map_unit_fn.rs
@@ -190,7 +190,7 @@ fn unit_closure<'tcx>(
 /// Anything else will return `a`.
 fn let_binding_name(cx: &LateContext<'_>, var_arg: &hir::Expr<'_>) -> String {
     match &var_arg.kind {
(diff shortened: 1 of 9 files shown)


nixxquality:
What's up with these failing tests? I did test before I committed and I re-tested now as well and nothing fails locally.

llogiq:
It seems like our code doesn't live up to your improved requirements. The dogfood tests shows where updates are needed.

nixxquality:
How do I run those tests locally?
Following the instructions in basics.md to "only run dogfood tests" using `cargo test --test dogfood` doesn't produce any failing results.

camsteffen:
Not sure why `cargo test` didn't work, but CI is passing now! Changes look good. Please squash commits (or change the commit message to something like "Fix dogfood").

camsteffen:
Thanks! @bors r+

PR #8086, opened 2021-12-07
Fix bad suggestion on `option_if_let_else` when there is complex subpat

### LINKED ISSUE #7991: `option_if_let_else` incorrect suggestion, ignores semantics of existing code
I tried this code:

```rust
enum Enum {
    One(u8),
    Two,
}

fn main() {
    let x = Some(Enum::One(1));

    let y = if let Some(one @ Enum::One(..)) = x {
        1
    } else {
        2
    };
}
```

Expected: option_if_let_else should probably not suggest anything here, it can't be trivially replaced with a map or unwrap function.

Instead, clippy suggested replacing the if let with:

```rust
x.map_or(2, |one| 1)
```

This is an incorrect suggestion that will compile but will change the behaviour of the code.

### Meta

```
binary: rustc
commit-date: 2021-11-01
release: 1.56.1
LLVM version: 13.0.0

```

I usually disable `option_if_let_else` because I find it usually makes the code harder to read, but every now and then I'll run pedantic/nursery to see if there are lints I should be enabling. Unlike most of the broken suggestions for this lint, this one will actually compile but break the semantics of the existing code.

@rustbot label +I-suggestion-causes-error +I-false-positive

dswij:
closes #7991 

Prefer not warning any complex subpat in `option_if_let_else` rather than suggesting obscure suggestions.

changelog: [`option_if_let_else`] does not warn when complex subpat is present

### DIFF
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -112,7 +112,7 @@ fn detect_option_if_let_else<'tcx>(cx: &LateContext<'tcx>, expr: &Expr<'tcx>) ->
         if !is_result_ok(cx, let_expr); // Don't lint on Result::ok because a different lint does it already
         if let PatKind::TupleStruct(struct_qpath, [inner_pat], _) = &let_pat.kind;
         if is_lang_ctor(cx, struct_qpath, OptionSome);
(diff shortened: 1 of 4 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8108, opened 2021-12-09
Split `needless_collect` and downgrade the split one

giraffate:
As discussed in [zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/split.20and.20downgrade.20.60needless_collect.60), `needless_collect` has many false positives, so it's time to downgrade it to `nursery`. But false positives have been reported for the indirect usage, not for the direct usage. So `needless_collect` is split into two and the indirect usage is downgraded to nursery.

The naming of the new lint is bothering me.

changelog: Split `needless_collect` and downgrade the split one as the new lint `needless_indirect_collect`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3033,6 +3033,7 @@ Released 2018-09-13
 [`needless_continue`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_continue
 [`needless_doctest_main`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_doctest_main
 [`needless_for_each`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_for_each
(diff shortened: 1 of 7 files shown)


llogiq:
Regarding the name, I found it a bit hard to understand the distinction between needless direct and indirect collect. And the docs don't make that distinction very clear either. So perhaps by trying to tease out the actual difference, we may be able to come up with a better name, or at least find our peace with the current one?

Otherwise this looks ready to merge.

giraffate:
You're right. And I can't come up with a good name for it either.

I'm thinking that it might be a good idea to downgrade the whole `needless_colect` to nursery for now and file the naming issue away, what do you think?

llogiq:
That would be a workable stopgap measure. I'd be happier with actually understanding the distinction well enough to explain it. I'll probably need some time to re-read and wrap my head around the issue.

giraffate:
I don't have enough time to think about this, so I'm closing this.

PR #8127, opened 2021-12-15
Fix `enum_variants` FP on prefixes that are not camel-case

### LINKED ISSUE #8090: enum_variant_names false positive when first word has common prefix as part of word
### Summary

The lint warns if all variants have the same prefix word. If the first variant has the prefix word at the start of its first word, it wrongly triggers the lint even if the first word is longer than the prefix word. If this variant is not the first variant, the warning is not triggered.

### Lint Name

enum_variant_names

### Reproducer

I tried this code:

```rust
enum _E {
    Normal,
    NoLeft,
    NoRight,
}
```

I saw this happen:

```
warning: all variants have the same prefix: `No`
 --> src/lib.rs:1:1
  |
1 | / enum _E {
2 | |     Normal,
3 | |     NoLeft,
4 | |     NoRight,
5 | | }
  | |_^
  |
  = note: `#[warn(clippy::enum_variant_names)]` on by default
  = help: remove the prefixes and use full paths to the variants instead of glob imports
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names
```

I expected no warning.

### Version

```text
rustc 1.58.0-beta.2 (0e07bcb68 2021-12-04)
binary: rustc
commit-hash: 0e07bcb68b82b54c0c4ec6fe076e9d75b02109cf
commit-date: 2021-12-04
release: 1.58.0-beta.2
LLVM version: 13.0.0

clippy 0.1.58 (0e07bcb 2021-12-04)
```

### Additional Labels

_No response_

dswij:
closes #8090 

Fix FP on `enum_variants` when prefixes are only a substring of a camel-case word. Also adds some util helpers on `str_utils` to help parsing camel-case strings.

This changes how the lint behaves:

1. previously if the Prefix is only a length of 1, it's going to get ignored, i.e. these were previously ignored and now is warned
```rust
enum Foo {
    cFoo,
    cBar,
    cBaz,
}

enum Something {
    CCall,
    CCreate,
    CCryogenize,
}
```
  
2. non-ascii characters that doesn't have casing will not be split, 
```rust
enum NonCaps {
    PrefixXXX,
    PrefixTea,
    PrefixCake,
}
```
will be considered as `PrefixXXX`, `Prefix`, `Prefix`, so this won't lint as opposed to fired previously.

changelog: [`enum_variant_names`] Fix FP when first prefix are only a substring of a camel-case word. 

---

 (Edited by @xFrednet removed some non ascii characters)

### DIFF
--- a/clippy_lints/src/enum_variants.rs
+++ b/clippy_lints/src/enum_variants.rs
@@ -3,7 +3,7 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_help};
 use clippy_utils::source::is_present_in_source;
 use clippy_utils::str_utils::{self, count_match_end, count_match_start};
-use rustc_hir::{EnumDef, Item, ItemKind};
+use rustc_hir::{EnumDef, Item, ItemKind, Variant};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::source_map::Span;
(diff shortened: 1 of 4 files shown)


tests/ui/enum_variants.stderr
```diff
@@ -60,31 +72,31 @@ LL | | }
...
    |
    = help: remove the prefixes and use full paths to the variants instead of glob imports
 
-error: all variants have the same prefix: `Prefix`
-  --> $DIR/enum_variants.rs:87:1
+error: all variants have the same prefix: `WithOut`
+  --> $DIR/enum_variants.rs:81:1
    |
-LL | / enum NonCaps {
-LL | |     Prefix的,
```
dswij, line 79: 的 is not lowercase nor uppercase, so I think it's pointless to check if this is a camel-case. Rather, we just not warn.

dswij: yeah, camel case (or case in this sense) is not properly defined in some non-ascii based characters, so it doesn't make sense to impose some sort of lint based on casing on them. If someone chose to use a non-ascii identifier, then I think it's fair to assume that they won't care about camelcase that much.

dswij: Forgot to add that, we don't really ignore non-ascii characters, but non-ascii characters that does not have casing will just be joined to the previous word

xFrednet: I mixed up the casing you were checking for in this example when I wrote my answer ^^. I would consider these letters like numbers, as they can also not be capitalizes. Your solution sounds good to me :+1:

xFrednet:
I would like to review this, I hope it's okay to *steal* the assignment ^^

r? @xFrednet

(It's on my todo list later today :upside_down_face: )

llogiq:
Sure! It's all yours, @xFrednet !

clippy_lints/src/enum_variants.rs
```diff
@@ -115,77 +115,80 @@ impl EnumVariantNames {
...
-        post = &post[post_end..];
-        let post_camel = str_utils::camel_case_start(post);
-        post = &post[post_camel.byte_index..];
+        pre = pre
+            .iter()
+            .copied()
+            .zip(variant_split.iter().copied())
+            .take_while(|(a, b)| a == b)
+            .map(|e| e.0)
+            .collect();
```
xFrednet, line 176: Here and below, you can drop the two `copied()` calls by dereferencing `e.0`
```rs
        pre = pre
            .iter()
            .zip(variant_split.iter())
            .take_while(|(a, b)| a == b)
            .map(|e| *e.0)
            .collect();
```
or using `into_iter` respectively. I think the suggested version above looks better.

tests/ui/enum_variants.stderr
xFrednet: Could you add this limitation to the lint description? You can use a `Known Problems` or `Limitations` section for that.

xFrednet (commented):
Thank you for the changes, overall, they look good to me and the code is also more readable now. I found some things which can probably be improved, but nothing major. I've also been a bit more nit picky than usual in `str_utils` :sweat_smile:. Afterwards, it should be ready to be merged :upside_down_face:

clippy_utils/src/str_utils.rs
```diff
@@ -78,9 +89,13 @@ pub fn camel_case_start(s: &str) -> StrIndex {
     } else {
         return StrIndex::new(char_count, s.len());
     }
+
     let mut down = true;
     let mut last_index = StrIndex::new(char_count, s.len());
     for (char_index, (byte_index, c)) in iter {
+        if byte_index < start_idx {
+            continue;
```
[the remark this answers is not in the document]
dswij: Ah it's a rev iter, good find!

clippy_utils/src/str_utils.rs
```diff
@@ -98,9 +113,55 @@ pub fn camel_case_start(s: &str) -> StrIndex {
...
+            idxs_iter.collect()
+        } else {
+            Vec::<usize>::from([0]).into_iter().chain(idxs_iter).collect()
+        }
+    } else {
+        return vec![s];
+    };
+    let split_points: Vec<(&usize, &usize)> = idxs[..idxs.len() - 1].iter().zip(&idxs[1..]).collect();
+
+    split_points.iter().map(|(&start, &stop)| &s[start..stop]).collect()
```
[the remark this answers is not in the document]
dswij: Yeah, this seems much better, I was trying to be too fancy there 😅

xFrednet (approved):
LGTM. One last NIT, Rust and Clippy have a [no-merge policy](https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy), meaning that no merge-commits should be added in PRs. Could you please remove the merge commit from this branch and rebase on master. I'm guessing the easiest way is to drop the commit first with `git rebase -i HEAD~9` (This opens the rebase interface) and afterwards using `git rebase upstream/master` or something similar depending on your setup. Let me know if you need help with that.

Then I think it's ready to be merged :upside_down_face:

clippy_utils/src/str_utils.rs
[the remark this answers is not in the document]
xFrednet: This also often happens to me, it's just more fun to go fancy ^^. BTW, there is also a `std::iter::once()` function to create a one time iterator, that's more efficient than `Vec::<usize>::from([0]).into_iter()` as it doesn't allocate any memory. :upside_down_face:

dswij:
@xFrednet BTW, this is rebased and ready to be re-reviewed :slightly_smiling_face:

xFrednet (approved):

xFrednet:
Awesome thank you very much! This looks good to me :+1: 

@bors r+

(Bors seems to have some problems since yesterday. Let's hope that it runs now :upside_down_face: )

dswij:
Lots of failed downloads on GH action, I suppose we can wait for a while :slightly_smiling_face: 

```
 Warning: Failed to download action 'https://api.github.com/repos/actions/download-artifact/tarball/18f0f591fbc635562c815484d73b6e8e3980482e'. Error: The operation was canceled.
Warning: Back off 12.276 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/actions/download-artifact/tarball/18f0f591fbc635562c815484d73b6e8e3980482e'. Error: The operation was canceled.
Warning: Back off 15.908 seconds before retry.
Error: The operation was canceled.
```

xFrednet:
GitHub had some outages, not everything seems to work again. let's try it one more time

@bors retry

PR #8133, opened 2021-12-17
Fix 8128

### LINKED ISSUE #8128: skip(..).next() lint incorrectly suggest using nth() in some cases
### Summary

Clippy incorrectly reports that I should switch to using `nth(..)` instead of the `skip(..).next()` but does not account for the fact that my variable is not mutable, and thus cannot use `nth(..)`. I found this while attempting a `cargo clippy --fix` and it applied this fix and caused compile error and prevented any further clippy fixes from being applied.

### Reproducer

I tried this code:

```rust
fn main() {
    let test_string = "1|1 2";
    let sp = test_string.split('|').map(|s| s.trim());
    let out: Vec<&str> = sp.skip(1).next().unwrap().split(' ').collect();

    println!("{:?}", out);
}

```

Either for clippy to realize I cannot use `nth()` as suggested because the variable is not mutable, or for it to change that variable to mutable as part of the fix (I could see how this could be complicated/bad).

I suppose changing the help to mention that nth() would be available if the variable was mutable and that the the user should consider changing it.

Instead, this happened:

Clippy reports that I should be using `nth()` instead of `skip(1).next()` but without other changes in my code this is not an allowable fix.

### Version

```text
binary: rustc
commit-date: 2021-11-01
host: x86_64-pc-windows-msvc
release: 1.56.1
LLVM version: 13.0.0
```

### Additional Labels

* `l-suggestion-causes-error`

surechen:
Fixes #8128

changelog: Fix  error suggestion of `skip(..).next()` for immutable variable.

### DIFF
--- a/clippy_lints/src/methods/iter_skip_next.rs
+++ b/clippy_lints/src/methods/iter_skip_next.rs
@@ -1,8 +1,10 @@
-use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::is_trait_method;
+use clippy_utils::path_to_local;
 use clippy_utils::source::snippet;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
+use rustc_hir::{BindingAnnotation, Node, PatKind};
 use rustc_lint::LateContext;
 use rustc_span::sym;
 
@@ -11,14 +13,34 @@ use super::ITER_SKIP_NEXT;
 pub(super) fn check(cx: &LateContext<'_>, expr: &hir::Expr<'_>, recv: &hir::Expr<'_>, arg: &hir::Expr<'_>) {
     // lint if caller of skip is an Iterator
     if is_trait_method(cx, expr, sym::Iterator) {
-        span_lint_and_sugg(
+        let mut application = Applicability::MachineApplicable;
+        span_lint_and_then(
             cx,
             ITER_SKIP_NEXT,
             expr.span.trim_start(recv.span).unwrap(),
             "called `skip(..).next()` on an iterator",
-            "use `nth` instead",
-            format!(".nth({})", snippet(cx, arg.span, "..")),
-            Applicability::MachineApplicable,
+            |diag| {
+                if_chain! {
+                    if let Some(id) = path_to_local(recv);
+                    if let Node::Binding(pat) = cx.tcx.hir().get(id);
+                    if let PatKind::Binding(ann, _, _, _)  = pat.kind;
+                    if ann != BindingAnnotation::Mutable;
+                    then {
+                        application = Applicability::Unspecified;
+                        diag.span_help(
+                            pat.span,
+                            &format!("for this change `{}` has to be mutable", snippet(cx, pat.span, "..")),
+                        );
+                    }
+                }
+
+                diag.span_suggestion(
+                    expr.span.trim_start(recv.span).unwrap(),
+                    "use `nth` instead",
+                    format!(".nth({})", snippet(cx, arg.span, "..")),
+                    application,
+                );
+            },
         );
     }
 }
--- a/tests/ui/iter_skip_next_unfixable.rs
+++ b/tests/ui/iter_skip_next_unfixable.rs
@@ -1,0 +1,19 @@
+#![warn(clippy::iter_skip_next)]
+#![allow(dead_code)]
+
+/// Checks implementation of `ITER_SKIP_NEXT` lint
+fn main() {
+    // fix #8128
+    let test_string = "1|1 2";
+    let sp = test_string.split('|').map(|s| s.trim());
+    let _: Vec<&str> = sp.skip(1).next().unwrap().split(' ').collect();
+    if let Some(s) = Some(test_string.split('|').map(|s| s.trim())) {
+        let _: Vec<&str> = s.skip(1).next().unwrap().split(' ').collect();
+    };
+    fn check<T>(s: T)
+    where
+        T: Iterator<Item = String>,
+    {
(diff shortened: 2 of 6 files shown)


surechen:
I want to add a counter example. It runs correctly in playground. [link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=69560f1f2c155241e3af1eec69765d51)
But the CI [failed](https://github.com/rust-lang/rust-clippy/runs/4555862500?check_suite_focus=true#: ~:text=thread%20%27-,compile_ test,-%27%20panicked%20at%20%27Some)

clippy_lints/src/methods/iter_skip_next.rs
```diff
@@ -11,6 +14,23 @@ use super::ITER_SKIP_NEXT;
...
+            if let PatKind::Binding(ann, _, _, _)  = pat.kind;
+            if ann != BindingAnnotation::Mutable;
+            then {
+                span_lint_and_note(
+                    cx,
+                    ITER_SKIP_NEXT,
+                    expr.span.trim_start(recv.span).unwrap(),
+                    "called `skip(..).next()` on an iterator",
+                    Some(pat.span),
+                    &format!("can use `nth(..)` if change {} to mut", snippet_opt(cx, pat.span).unwrap()),
```
flip1995, line 29: You could also add the `arg.span` in the help message, like it is done below. But I would check if `arg.span` is a multiline span first (if it is, I would just print the `..`). You can do this with `x.sess().source_map().is_multiline(arg.span);`

I would also split this up with `span_lint_and_then` and use `span_help` for the note to make the binding mutable and `span_suggestion` for the `.nth` part.

flip1995 (changes requested):
The tests are failing, because `rustfix` can only apply suggestions produces by `span_lint_and_sugg`/`span_suggestion`. So you probably have to create a separate test file, that doesn't have the `// run rustfix` comment at the top.

clippy_lints/src/methods/iter_skip_next.rs
surechen: > You could also add the `arg.span` in the help message, like it is done below. But I would check if `arg.span` is a multiline span first (if it is, I would just print the `..`). You can do this with `x.sess().source_map().is_multiline(arg.span);`
> 
> I would also split this up with `span_lint_and_then` and use `span_help` for the note to make the binding mutable and `span_suggestion` for the `.nth` part.

Done. Thanks.

surechen:
Hi, could you help me see this? Is it consistent with your modification suggestion? Thanks.
r? @xFrednet

tests/ui/iter_skip_next_unfixable.rs
```diff
@@ -0,0 +1,9 @@
+#![warn(clippy::iter_skip_next)]
+
+/// Checks implementation of `ITER_SKIP_NEXT` lint
+fn main() {
+    // fix #8128
+    let test_string = "1|1 2";
+    let sp = test_string.split('|').map(|s| s.trim());
+    let _: Vec<&str> = sp.skip(1).next().unwrap().split(' ').collect();
```
xFrednet, line 8: Could you also add a test where the iterator comes from a function or closure argument to make sure that those patterns also work? A closure might be easier as it doesn't require a type annotations :). A check if `if let` would also be awesome :upside_down_face:

xFrednet (commented):
> Hi, could you help me see this? Is it consistent with your modification suggestion? 

Sure, this is what I had in mind. I again looked into a way to check this differently, but I don't see a viable solution to get the information from the [type](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyS.html) of the `self` argument of `skip`. And I think this is fine, since `@flip1995` didn't mention another solution in his first review.

I have two suggestions how the output could be a bit more readable, but otherwise this looks good to me. Thank you for the work :upside_down_face:

surechen:
> > Hi, could you help me see this? Is it consistent with your modification suggestion?
> 
> Sure, this is what I had in mind. I again looked into a way to check this differently, but I don't see a viable solution to get the information from the [type](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.TyS.html) of the `self` argument of `skip`. And I think this is fine, since `@flip1995` didn't mention another solution in his first review.
> 
> I have two suggestions how the output could be a bit more readable, but otherwise this looks good to me. Thank you for the work 🙃

Done. Thanks.

xFrednet (approved):
Looks good to me! :+1: 

Could you please squash your commits before the merge? If you want to keep the messages, you can squash the `fix #8128` commits into one and all review commits into another one.

A related note: In Rust we usually avoid mentioning the issue in commit messages as that creates a new link in the issue with every rebase. I do see the advantage of mentioning it and use that feature in other projects as well. I think the idea is more to use a descriptive commit message that can be understood without opening an issue. We can also merge commits with such messages, and I'm happy to do so if you want to keep the messages. Just something I wanted to mention, this also depends on your personal style :upside_down_face:

surechen:
> Looks good to me! 👍
> 
> Could you please squash your commits before the merge? If you want to keep the messages, you can squash the `fix #8128` commits into one and all review commits into another one.
> 
> A related note: In Rust we usually avoid mentioning the issue in commit messages as that creates a new link in the issue with every rebase. I do see the advantage of mentioning it and use that feature in other projects as well. I think the idea is more to use a descriptive commit message that can be understood without opening an issue. We can also merge commits with such messages, and I'm happy to do so if you want to keep the messages. Just something I wanted to mention, this also depends on your personal style 🙃

Done. Thank you very much.

xFrednet (approved):

xFrednet:
LGTM, thank you for the update and quick replies! :+1: 

@bors r+

PR #8138, opened 2021-12-17
Fix `SAFETY` comment tag casing in undocumented_unsafe_blocks

r00ster91:
This changes the lint introduced in #7748 to suggest adding a `SAFETY` comment instead of a `Safety` comment.

Searching for `// Safety:` in rust-lang/rust yields 67 results while `// SAFETY:` yields 1072.
I think it's safe to say that this comment tag is written in upper case, just like `TODO`, `FIXME` and so on are. As such I would expect this lint to follow the official convention as well.

Note that I intentionally introduced some casing diversity in `tests/ui/undocumented_unsafe_blocks.rs` to test more cases than just `Safety:`.

changelog: Capitalize `SAFETY` comment in [`undocumented_unsafe_blocks`]

### DIFF
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -15,7 +15,7 @@ use std::borrow::Cow;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for `unsafe` blocks without a `// Safety: ` comment
+    /// Checks for `unsafe` blocks without a `// SAFETY: ` comment
     /// explaining why the unsafe operations performed inside
     /// the block are safe.
     ///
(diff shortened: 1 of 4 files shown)


xFrednet:
I like standardizing the capitalization. :+1:  You might also want to take a look at the (`missing_safety_doc`)[https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc] lint. That one lints missing safety heading in doc comments. :upside_down_face:

r00ster91:
@xFrednet I think in that case `/// # Safety` is fine as-is because it denotes a markdown section like e.g. `/// # Examples` does (which is not uppercased either). I believe that and this comment tag `// SAFETY: ` which is directly followed by the explanation about why it is safe are two slightly different things.

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8144, opened 2021-12-19
Add suggestion for neg_multiply lint

### LINKED ISSUE #8115: neg_multiply should show a suggestion
### Description

```
  --> worker/src/proc.rs:78:19
   |
78 |         let pid = pid as i32 * -1;
   |                   ^^^^^^^^^^^^^^^
   |
   = note: `#[warn(clippy::neg_multiply)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#neg_multiply
```

There's no suggestion how this should be done and the [docs](https://rust-lang.github.io/rust-clippy/master/index.html#neg_multiply) just say:

> It’s more readable to just negate.

but don't show how this is supposed to be done.

I had to google "how to negate an integer" and clippy should just suggest this:

```rust
let pid = -(pid as i32);
```

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Gh0stm4chine:
This fixes #8115 by adding a suggestion for [neg_multiply].

My first issue on Github, any feedback or input is welcome 😃 

changelog: create a suggestion for `neg_multiply`

### DIFF
--- a/clippy_lints/src/neg_multiply.rs
+++ b/clippy_lints/src/neg_multiply.rs
@@ -1,8 +1,9 @@
 use clippy_utils::consts::{self, Constant};
 use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::source::snippet_with_applicability;
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::{BinOpKind, Expr, ExprKind, QPath, UnOp};
+use rustc_hir::{BinOpKind, Expr, ExprKind, UnOp};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::source_map::Span;
@@ -28,7 +29,7 @@ declare_clippy_lint! {
     #[clippy::version = "pre 1.29.0"]
     pub NEG_MULTIPLY,
     style,
-    "multiplying integers with `-1`"
+    "multiplying integers by `-1`"
 }
 
 declare_lint_pass!(NegMultiply => [NEG_MULTIPLY]);
@@ -54,17 +55,15 @@ fn check_mul(cx: &LateContext<'_>, span: Span, lit: &Expr<'_>, exp: &Expr<'_>) {
         if let ExprKind::Lit(ref l) = lit.kind;
         if consts::lit_to_constant(&l.node, cx.typeck_results().expr_ty_opt(lit)) == Constant::Int(1);
         if cx.typeck_results().expr_ty(exp).is_integral();
-        if let ExprKind::Path(QPath::Resolved(_, var_path)) = exp.kind;
 
         then {
-            let var_name = var_path.segments[0].ident.name.as_str();
-            let suggestion = format!("-{var}",var=var_name);
-            let applicability = Applicability::MachineApplicable;
+            let mut applicability = Applicability::MachineApplicable;
+            let suggestion = format!("-{}", snippet_with_applicability(cx, exp.span, "..", &mut applicability));
             span_lint_and_sugg(
                     cx,
                     NEG_MULTIPLY,
                     span,
-                    "this `multiplication with -1` can be written more succinctly",
+                    "this multiplication by -1 can be written more succinctly",
                     "consider using",
                     suggestion,
                     applicability,
--- a/clippy_lints/src/neg_multiply.rs
+++ b/clippy_lints/src/neg_multiply.rs
@@ -1,7 +1,8 @@
 use clippy_utils::consts::{self, Constant};
-use clippy_utils::diagnostics::span_lint;
+use clippy_utils::diagnostics::span_lint_and_sugg;
 use if_chain::if_chain;
-use rustc_hir::{BinOpKind, Expr, ExprKind, UnOp};
+use rustc_errors::Applicability;
+use rustc_hir::{BinOpKind, Expr, ExprKind, QPath, UnOp};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::source_map::Span;
@@ -18,7 +19,11 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```ignore
-    /// x * -1
+    /// // Bad
+    /// let a = x * -1;
+    ///
+    /// // Good
+    /// let b = -x;
     /// ```
     #[clippy::version = "pre 1.29.0"]
     pub NEG_MULTIPLY,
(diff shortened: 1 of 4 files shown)


tests/ui/neg_multiply.stderr
```diff
@@ -1,16 +1,16 @@
...
-   |     ^^^^^^
+   |     ^^^^^^ help: consider using: `-x`
    |
    = note: `-D clippy::neg-multiply` implied by `-D warnings`
 
-error: negation by multiplying with `-1`
+error: this `multiplication with -1` can be written more succinctly
   --> $DIR/neg_multiply.rs:29:5
    |
 LL |     -1 * x;
```
xFrednet, line 12: Could you also add some test cases where the second factor is more than a value. Here are some examples you can take:

```rs
100 + x * -1;
(100 + x) * -1;
-1 * 17;
0xcafe | 0xff00 * -1
```

Then this should be sufficiently tested :upside_down_face:

clippy_lints/src/neg_multiply.rs
```diff
@@ -18,7 +19,11 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```ignore
-    /// x * -1
+    /// // Bad
+    /// let a = x * -1;
+    ///
+    /// // Good
+    /// let b = -x;
```
xFrednet, line 26: Nice addition :+1: 

(You can resolve this comment :upside_down_face: )

xFrednet (commented):
Hey @Gh0stm4chine and welcome to the project,
I've made some suggestions on how the code can be improved, the general direction was already good :+1:.

One more thing related to the PR description: Clippy creates it's changelog based on the description. Could you include a text like this: ``changelog: create a suggestion for [`neg_multiply`]`` or something similar? Otherwise, everything looks good. Let me know if you have any more questions :upside_down_face:

clippy_lints/src/neg_multiply.rs
```diff
@@ -49,8 +54,21 @@ fn check_mul(cx: &LateContext<'_>, span: Span, lit: &Expr<'_>, exp: &Expr<'_>) {
...
         then {
-            span_lint(cx, NEG_MULTIPLY, span, "negation by multiplying with `-1`");
+            let var_name = var_path.segments[0].ident.name.as_str();
+            let suggestion = format!("-{var}",var=var_name);
+            let applicability = Applicability::MachineApplicable;
+            span_lint_and_sugg(
+                    cx,
+                    NEG_MULTIPLY,
+                    span,
+                    "this `multiplication with -1` can be written more succinctly",
```
r00ster91, line 67:
```suggestion
                    "this multiplication with -1 can be written more succinctly",
```
It looks strange to me to put "multiplication with -1" in grave accents. Is there a reason for that? I would understand if it were `* -1` or something similar that's actually code.
The old string (above) used grave accents only for the number `-1` (which can be code):
```diff
-             span_lint(cx, NEG_MULTIPLY, span, "negation by multiplying with `-1`");
```
Maybe you meant to do that instead, if at all?

tests/ui/neg_multiply.stderr
```diff
@@ -1,16 +1,16 @@
-error: negation by multiplying with `-1`
+error: this `multiplication with -1` can be written more succinctly
```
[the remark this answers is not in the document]
Gh0stm4chine: Hi @xFrednet , thank you for your feedback. 
Actually rustfix can't apply the correct suggestion for the test case `0xcafe | 0xff00 * -1`:
`LL |     0xcafe | 0xff00 * -1
   |              ^^^^^^^^^^^ help: consider using: -0xff00` 
I get the following error : `failed to apply suggestions for "tests\\ui\\neg_multiply.rs" with rustfix`

clippy_lints/src/neg_multiply.rs
Gh0stm4chine: Hi @r00ster91 , 
Yes you are right 👍 I will fix it.

tests/ui/neg_multiply.stderr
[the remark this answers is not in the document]
Gh0stm4chine: Nevermind.. I was missing `;` at the end of the line 😑

tests/ui/neg_multiply.stderr
Gh0stm4chine: Seems good 👍

clippy_lints/src/neg_multiply.rs
```diff
@@ -49,8 +54,21 @@ fn check_mul(cx: &LateContext<'_>, span: Span, lit: &Expr<'_>, exp: &Expr<'_>) {
         if let ExprKind::Lit(ref l) = lit.kind;
         if consts::lit_to_constant(&l.node, cx.typeck_results().expr_ty_opt(lit)) == Constant::Int(1);
         if cx.typeck_results().expr_ty(exp).is_integral();
+        if let ExprKind::Path(QPath::Resolved(_, var_path)) = exp.kind;
```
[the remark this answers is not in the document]
Gh0stm4chine: Thank you @xFrednet. 
I took into account your suggestion, that's better indeed.

xFrednet (approved):

xFrednet:
It looks good to me :+1:. The pipeline failed as it was unable to apply one suggestion. This can happen when multiple suggestions overlap, but that shouldn't be the case here. I've restarted the pipeline to see if it works now.

Also thank you to @r00ster91 for the review comment :+1:

tests/ui/neg_multiply.rs
```diff
@@ -1,3 +1,4 @@
+// run-rustfix
 #![warn(clippy::neg_multiply)]
 #![allow(clippy::no_effect, clippy::unnecessary_operation)]
```
xFrednet, line 3:
```suggestion
#![allow(clippy::no_effect, clippy::unnecessary_operation, clippy::precedence)]
```
The `clippy::precedence` lint triggers for the `0xcafe | 0xff00 * -1;` expression. This causes the pipeline to fail. If you allow that lint and rerun the tests and update the output, everything should be fine :upside_down_face:

Gh0stm4chine: Done 😃

Gh0stm4chine: The pipeline fails because it can't compile the fixed code. I tested it locally, it compiles fine 😕

xFrednet: Some tests can't be executed on Windows, if you use that it might explain this behavior. I'll check it out and try to find the problem :upside_down_face:

xFrednet: The tests fail due to the configuration of `compiletest-rs`. In the normal UI tests all lints inside the `unused` group are allowed, this is sadly not true for `.fixed` files. The problem can be solved by adding `#![allow(unused)]` to the file, as that will allow these lints and prevent the warnings that make the test fail.

(This is something I want to fix at some point, but for now, this is the best and easiest solution. Sorry that you ran into this :sweat_smile:)

Gh0stm4chine: > Some tests can't be executed on Windows, if you use that it might explain this behavior. I'll check it out and try to find the problem 🙃

Yes, I am on Windows, but actually running the test locally was failing too. I thought the problem was about compiling the .fixed file.





> The tests fail due to the configuration of `compiletest-rs`. In the normal UI tests all lints inside the `unused` group are allowed, this is sadly not true for `.fixed` files. The problem can be solved by adding `#![allow(unused)]` to the file, as that will allow these lints and prevent the warnings that make the test fail.

I was not sure if the `#![allow(unused)]` should be added in the .fixed only even though it's a generated file ?
I added it in test file for now.

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the quick changes :upside_down_face: 

@bors r+

tests/ui/neg_multiply.rs
xFrednet: > Yes, I am on Windows, but actually running the test locally was failing too. I thought the problem was about compiling the .fixed file.

Alright :+1:, your assumption was correct, it was with the compilation of the `.fixed` files. 

> I was not sure if the `#![allow(unused)]` should be added in the .fixed only even though it's a generated file?
I added it in test file for now.

You did it correctly, that's what I meant :+1:

PR #8146, opened 2021-12-19
Don't emit RETURN_SELF_NOT_MUST_USE lint if `Self` already is marked as `#[must_use]`

GuillaumeGomez:
New bug discovered with this lint. Hopefully, this is the last one.

---

changelog: none

### DIFF
--- a/clippy_lints/src/return_self_not_must_use.rs
+++ b/clippy_lints/src/return_self_not_must_use.rs
@@ -1,4 +1,5 @@
-use clippy_utils::{diagnostics::span_lint, must_use_attr, nth_arg, return_ty};
+use clippy_utils::ty::is_must_use_ty;
+use clippy_utils::{diagnostics::span_lint, nth_arg, return_ty};
 use rustc_hir::def_id::LocalDefId;
 use rustc_hir::intravisit::FnKind;
 use rustc_hir::{Body, FnDecl, HirId, TraitItem, TraitItemKind};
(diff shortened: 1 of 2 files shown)


clippy_lints/src/return_self_not_must_use.rs
```diff
@@ -50,6 +51,7 @@ fn check_method(cx: &LateContext<'tcx>, decl: &'tcx FnDecl<'tcx>, fn_def: LocalD
         if decl.implicit_self.has_implicit_self();
         // We only show this warning for public exported methods.
         if cx.access_levels.is_exported(fn_def);
+        if !cx.tcx.hir().attrs(hir_id).iter().any(|attr| attr.has_name(sym::must_use));
```
xFrednet, line 54: Isn't this case covered by the `must_use_attr` check in line 57?

xFrednet (commented):
One small question, it otherwise looks good to me :upside_down_face:

clippy_lints/src/return_self_not_must_use.rs
GuillaumeGomez: Indeed but apparently it's not working as it should. I'll clean up though. Good catch!

GuillaumeGomez: (Which was fixed in #8143 but I realized it would be better checked here instead.)

GuillaumeGomez:
Updated!

GuillaumeGomez:
And removed the warning as well.

xFrednet (approved):

xFrednet:
LGTM, thank you for following up on these bugs :+1:. Let's hope it was the last one :upside_down_face: 

@bors r+

GuillaumeGomez:
I hope as well! Thanks for the quick reviews!

GuillaumeGomez:
@xFrednet Just to confirm: I still have this bug showing up when working on glib on the latest rust nightly:

```
$ rustc --version
rustc 1.59.0-nightly (e100ec5bc 2021-12-21)
$ cargo clippy --version
clippy 0.1.59 (e100ec5 2021-12-21)
```

Has this patch been "merged" into the nightly or not yet?

xFrednet:
Hey @GuillaumeGomez, the fix has not been merged into nightly. The sync is usually done Thursdays, meaning today it should be synced. However, it might be delayed due to holidays and such. I can ping you on the sync PR or here if you like :upside_down_face:

GuillaumeGomez:
Yes please! Thank you!

flip1995:
The regular sync will be next week Thursday. I can do an out-of-cycle sync today, if you need this fix tomorrow.

GuillaumeGomez:
No hurry, don't worry. Before beta/stable release would be nice though. ;)

flip1995:
Yes, this will get in before beta branch 👍 (famous last words)

Issue #2852, opened 2018-06-18
Lint for useless tuple binding.


c-edw:
```rust
fn main() {
    let a = 1;
    
    match (a, 2) {
        (a, 2) => {},
        _ => {},
    }
}
```

The `match (a, 2)` can be reduced to just `match 2`, and the `(a, 2) => {}` can be reduced to `2 => {}`.

mati865:
> The match (a, 2) can be reduced to just match 2

You meant `match a`?

c-edw:
Some examples.

```rust
fn main() {
    let a = 1;
    
    match (a, 2) {
        (a, 2) => {},
        _ => {},
    }
}
```
should be reduced to
```rust
fn main() {
    let a = 1;
    
    match 2 {
        2 => {},
        _ => {},
    }
}
```rust
fn main() {
    let a = 1;
    
    match (a, 2) {
        (1, 2) => println!("{}", 1),
        (_, 2) => println!("{}", a),
    }
}

camsteffen:
Related to if not the same as #7122

PR #8157, opened 2021-12-23
New lint: `single_field_patterns`

### LINKED ISSUE #7122: New Lint: `match_single_field`
### What it does

Finds `match` or `if let` expressions where only one inner field is actually being matched. This could be a struct field, a tuple struct field or an array element (but not slices).

### Categories

- Kind: style

It is simpler.

### Drawbacks

None.

### Example

```rust
struct Strct { bar: Option<u32> }
struct Tup(Option<u32>, u32);

fn fun(struct: Strct, arr: [Option<u32>; 2], tup: Tup) {
	  if let Strct { bar: Some(_), .. } = strct { todo!() }
 
	  match arr {
	      [.., None] => todo!(),
	      [.., Some(_)] => todo!(),
	  }

	  if let Tup(Some(_), _) = tup { todo!() }
}
```

Could be written as:

```rust
fn fun(struct: Strct, arr: [Option<u32>; 2], tup: Tup) {
	  if let Some(_) = strct.bar { todo!() }

	  match arr[1] {
	      None => todo!(),
	      Some(_) => todo!(),
	  }

	  if let Some(_) = tup.0 { todo!() }
}
```

Note: It is possible for the reduced match to look like `if let 1 = x[0]` which is better expressed as `if x[0] == 1`.

Another lint name idea: `reducible_match`

hahohihu:
changelog: Add ``[`single_field_patterns`]`` lint

fixes #7122 

Comments:
1. I wasn't sure how to handle #2852 in a relevant way since it seems to require context outside of the expression/statement - which could make the code more complex - if I'm missing something, or there's a smart way of approaching it, I'm all ears.
2. I don't lint [.., x] or (.., x) because those communicate "get the last element", where array[800] or tuple.4 would feel a bit magical (less relevant in the tuple case imo, that's a bit iffy)
3. I allow `if let 1 = x[0]` to be suggested because handling it would require effectively re-implementing equatable_if_let, and it deals with the same issues. So I think dealing with that should be equatable_if_let's responsibility, and if people don't like it, they should enable that lint and do a second pass.


mikerite:
This lint generates too many warnings to be warn by default. I tested it on around 1000 popular crates. I got 541 warnings. The 10 worst crates were:

| Number | Crate |
|------------|---------|
|25 | chrono-0.4.19 |
|24 | petgraph-0.6.0 |
|21 | hdrhistogram-7.4.0 |
|16 | mockall_derive-0.11.0 |
|16 | html5ever-0.25.1 |
|15 | num-bigint-0.4.3 |
|13 | rust-crypto-0.2.36 | 
|12 | crossbeam-epoch-0.9.5  |
|11 | rayon-1.5.1 |
|11 | image-0.23.14 |

Personally, I strongly prefer the pattern matching to the tuple indexing. I'd like to hear other people's opinions on this.

---

From the implementation side, the only problem I found is that the suggestions don't handle macros correctly.

```rust
#![allow(unused_variables)]

macro_rules! m {
    () => { (1, 2, 30) }
}

fn main() {
    let (_, x, _) = m!();
}
```
outputs

```
warning: this single-variant pattern only matches one field
 --> src/main.rs:8:5
  |
8 |     let (_, x, _) = m!();
  |     ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(clippy::single_field_patterns)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_field_patterns
help: try this
  |
4 ~     () => { (1, 2, 30).1 }
5 | }
6 | 
7 | fn main() {
8 ~     let x = m!();
  |
```

hahohihu:
> Personally, I strongly prefer the pattern matching to the tuple indexing. I'd like to hear other people's opinions on this.
I think you're in the majority. Breaking down 34 lints from `cargo lintcheck`, I got:
- 28 plain tuples
- 3 tuple structs
- 3 structs with named fields

The frequency of plain tuple matching suggests that people are doing it intentionally.

Also, of the struct lints, 5/6 were newtypes, which may also be something some people do intentionally.

> This lint generates too many warnings to be warn by default.
Would it make more sense if I made it configurable, and ignored plain tuples + newtypes by default?



> From the implementation side, the only problem I found is that the suggestions don't handle macros correctly.

clippy_lints/src/comparison_chain.rs
```diff
@@ -73,7 +73,7 @@ impl<'tcx> LateLintPass<'tcx> for ComparisonChain {
         }
 
         // Check that there exists at least one explicit else condition
-        let (conds, _) = if_sequence(expr);
+        let conds = if_sequence(expr).0;
         if conds.len() < 2 {
             return;
         }
```
matthiaskrgr, line 79: Hmm, my two cents;
I would not see this as an improvement in my own code. :/

If for example we do some code changes and it turns out we need the second return type of `if_sequence()` as well, we'd have to revert back to `(conds, ..)` which is more work than just replacing `_` with a variable binding.

hahohihu:
Considering the feedback, I've opted to only handle structs with named fields. I could add tuples/arrays/etc. back as a pedantic lint if anyone was interested, and assuming that's an appropriate category for them (they seem controversial).

I don't think there's any getting around the loss of flexibility, but I think in some of these cases the decrease in verbosity is a fair trade-off. Here it seem reasonable to me - the type is no longer specified, and it's less verbose overall:
```rust
-if let Spanned {
-    node: LitKind::Bool(lit),
-    ..
-} = *span
-{
+if let LitKind::Bool(lit) = span.node {
```

Does that seem like a reasonable correction?

Also, there are currently cases where it requires repetition, which isn't good - I'm planning on making changes to not lint these, but wanted to make sure I'm not wasting time with a bad design:
```rust
-let Range { start, .. } = self.pixel_indices_unchecked(sx, sy);
+let start = self.pixel_indices_unchecked(sx, sy).start;
```

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+        None => Some(SingleField::Unused),
+    }
+}
+
+fn struct_sf(pat: &PatKind<'_>) -> Option<SingleField> {
+    match *pat {
+        PatKind::Struct(_, pats, _) => get_sf(pats.iter().map(|field| (field.ident, field.pat))),
+        _ => None,
+    }
+}
```
flip1995, line 92: This should then also go into the impl block. Since the get_sf fn is only used in this function, there's probably no reason to separate those.

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+/// e.g. 1 | (2 | 9) | 3..5
+struct FlatPatterns<'hir, I>
+where
+    I: Iterator<Item = &'hir Pat<'hir>>,
+{
+    patterns: I,
+    stack: Vec<&'hir Pat<'hir>>,
+}
+
+impl<I: Iterator<Item = &'hir Pat<'hir>>> FlatPatterns<'hir, I> {
```
flip1995, line 104: Is specifying the bound on `I` here again necessary, if it is already defined in the struct definition?

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+impl LateLintPass<'_> for SingleFieldPatterns {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &Expr<'tcx>) {
+        if expr.span.from_expansion() {
+            return;
+        }
+        match IfLetOrMatch::parse(cx, expr) {
+            Some(IfLetOrMatch::Match(scrutinee, arms, MatchSource::Normal)) => {
+                lint_sf(cx, expr.span, scrutinee, arms.iter().map(|arm| arm.pat));
+            },
+            Some(IfLetOrMatch::IfLet(scrutinee, pat, ..)) => lint_sf(cx, expr.span, scrutinee, once(pat)),
```
flip1995, line 242: Big NIT, but it took me some time to figure out what the `once` function does, until I remembered that it is a function from `iter`.

```suggestion
            Some(IfLetOrMatch::IfLet(scrutinee, pat, ..)) => lint_sf(cx, expr.span, scrutinee, iter::once(pat)),
```

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+        SINGLE_FIELD_PATTERNS,
+        span,
+        "this single-variant pattern only matches one field",
+        |diag| {
+            multispan_sugg_with_applicability(diag, "try this", Applicability::MachineApplicable, sugg);
+        },
+    );
+}
+
+fn remove_deref<'a>(mut scrutinee: &'a Expr<'a>) -> &'a Expr<'a> {
```
flip1995, line 194: Are those lifetimes necessary?

```suggestion
fn remove_deref(mut scrutinee: &Expr<'_>) -> &Expr<'_> {
```

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+    if scrutinee.span.from_expansion() {
+        return;
+    }
+    let scrutinee_name = if let Some(name) = snippet_opt(cx, remove_deref(scrutinee).span) {
+        name
+    } else {
+        return;
+    };
+    match cx.typeck_results().expr_ty(scrutinee).kind() {
+        ty::TyKind::Adt(def, ..) if def.is_struct() => {
```
flip1995, line 217: This should be re-exported.

```suggestion
        ty::Adt(def, ..) if def.is_struct() => {
```

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+            }
+        }
+        unreachable!("Or always has 2 patterns, so one of the prior returns must return");
+    }
+}
+
+fn find_sf_lint<'hir>(
+    cx: &LateContext<'_>,
+    patterns: impl Iterator<Item = &'hir Pat<'hir>>,
+    leaf_sf: &impl Fn(&PatKind<'hir>) -> Option<SingleField>,
```
flip1995, line 137: Why is this passed? There's only one place this function is called. Probably a relic of a previous version?

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+    let fields = FlatPatterns::new(patterns).map(|p| {
+        (
+            p.span,
+            if matches!(p.kind, PatKind::Wild) {
+                Some(SingleField::Unused)
+            } else {
+                leaf_sf(&p.kind)
+            },
+        )
+    });
```
flip1995, line 148:
```suggestion
    let fields = FlatPatterns::new(patterns).filter_map(|p| {
        if matches!(p.kind, PatKind::Wild) {
            Some(SingleField::Unused)
        } else {
            leaf_sf(&p.kind)
        }.map(|sf| (p.span, sf))
    });
```

With that you don't have to check for `Some(sf) = sf` in the for loop.

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+}
+
+fn apply_lint_sf(cx: &LateContext<'_>, span: Span, sugg: impl IntoIterator<Item = (Span, String)>) {
+    span_lint_and_then(
+        cx,
+        SINGLE_FIELD_PATTERNS,
+        span,
+        "this single-variant pattern only matches one field",
+        |diag| {
+            multispan_sugg_with_applicability(diag, "try this", Applicability::MachineApplicable, sugg);
```
flip1995, line 189: We/I try to get away from those generic "try this", because they are really unhelpful if the suggestion can't be fully displayed (i.e. multiline span too big).
```suggestion
            multispan_sugg_with_applicability(diag, "try accessing the field directly", Applicability::MachineApplicable, sugg);
```

and

```suggestion
            multispan_sugg_with_applicability(diag, "try this", Applicability::MaybeIncorrect, sugg);
```

because this can lead to rustc warnings (duplicated `_ => {}` match arms).

clippy_lints/src/single_field_patterns.rs
```diff
@@ -0,0 +1,264 @@
...
+fn struct_sf(pat: &PatKind<'_>) -> Option<SingleField> {
+    match *pat {
+        PatKind::Struct(_, pats, _) => get_sf(pats.iter().map(|field| (field.ident, field.pat))),
+        _ => None,
+    }
+}
+
+/// This handles recursive patterns and flattens them out lazily
+/// e.g. 1 | (2 | 9) | 3..5
+struct FlatPatterns<'hir, I>
```
flip1995, line 96: Is there a test for the issue this tries to resolve?

tests/ui/single_field_patterns.rs
```diff
@@ -0,0 +1,89 @@
...
+        Struct {
+            field1: Some(_) | None, ..
+        } => {},
+    }
+    while let Struct { field1: Some(5), .. } = struct1 {}
+}
+
+fn lint_ref(struct1: &&Struct) {
+    // this should suggest struct1.field1, NOT **struct1.field1
+    let Struct { field1, .. } = **struct1;
```
flip1995, line 36: Please also add a test case where the field is `ref`ed:

```suggestion
    let Struct { ref field1, .. } = **struct1;
```

or `ref mut`ed

tests/ui/single_field_patterns.rs
```diff
@@ -0,0 +1,89 @@
...
+    match struct1 {
+        Struct { field1: Some(1), .. } => {},
+        Struct {
+            field1: _,
+            field2: Some(1),
+        } => {},
+        _ => {},
+    }
+    let s @ Struct { field1, .. } = struct1;
+}
```
flip1995, line 87: Please add tests for tuples, to document that it was an active decision not to lint those.

flip1995 (changes requested):
Sorry for the way too late review, _$day_job_ and christmas vacation got in the way..

This lint and its impl looks really good overall. I agree that tuples should not be linted.

> Also, there are currently cases where it requires repetition, which isn't good - I'm planning on making changes to not lint these, but wanted to make sure I'm not wasting time with a bad design:

What do you mean by "repetition"?

clippy_lints/src/single_field_patterns.rs
hahohihu: It seems to be - or at least I wasn't able to figure out how to not include it

clippy_lints/src/single_field_patterns.rs
hahohihu: I think so - if I try to use the suggested change, I get a build error

clippy_lints/src/single_field_patterns.rs
hahohihu: filter_map has slightly different semantics. leaf_sf (which I've since inlined), returns None on PatKind::Binding - and we shouldn't lint one of those

hahohihu:
> Sorry for the way too late review, _$day_job_ and christmas vacation got in the way..

No worries, I was also preoccupied with work and some other stuff for a while.

> What do you mean by "repetition"?

By repetition, I meant something it would lint something like `let Struct { field, .. } = s` and suggest `let field = s.field` - though I suppose whether or not that's bad depends on whether you would rather repeat the struct name or the field name. Should I avoid that case?



Also, I'm not sure why CI is failing - my local machine is building/passing tests with rust 1.60.0 nightly, and the failing code doesn't seem problematic to me. Any ideas why that might be happening?

flip1995:
> not sure if I should resolve conversations myself or wait for you to.

You can resolve them yourself.

> By repetition, I meant it would lint something like `let Struct { field, .. } = s` and suggest `let field = s.field`

I would lint those too. If the user would care about the name, they would have named it in the `Struct` match already.



> Also, I'm not sure why CI is failing - my local machine is building/passing tests with rust 1.60.0 nightly,

You probably need to rebase on the latest master. With that rebase you'll also get the same nightly that CI uses (the one specified in rust-toolchain.toml on the master branch)

tests/ui/single_field_patterns.stderr
```diff
@@ -115,5 +126,27 @@ help: try this
...
+
+error: this single-variant pattern only matches one field
+  --> $DIR/single_field_patterns.rs:39:5
+   |
+LL |     let Struct { ref mut field1, .. } = **struct1;
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+help: try this
+   |
+LL |     let ref mut field1 = struct1.field1;
```
flip1995, line 148: I think a better suggestion would be for the `ref` and `ref mut` pattern to borrow the field, instead of keeping it in the pattern. This is just more idiomatic Rust IMO.

tests/ui/single_field_patterns.rs
```diff
@@ -0,0 +1,132 @@
+#![warn(clippy::single_field_patterns)]
```
flip1995, line 1: This should be `// run-rustfix`, since this lint is auto applicable in theory. With that comment on top of the test file, the UI tests will make sure that the suggestions are applied cleanly and don't break anything.

flip1995 (commented):
I have one comment left. Not sure how hard it would be to implement this. If you can't find an easy way to do that, feel free to keep the current behavior. This can also be done in a future PR.

How to move forward:

Please rebase on top of the latest master. Also, please squash (some of)  your commits. If you need help with that, let me know!

After that, I'll give it a final review and this should be ready to merge!

flip1995:
Closing because of inactivity. @hahohihu  if you want to continue to work on this, let me know!

hahohihu:
@flip1995 Sorry for abandoning this - life got in the way. Do you think I could pick it back up?

nk9:
Seems like the author wants to continue work on this. Can it be reopened?

@rustbot label S-triage

samueltardieu:
The author can reopen a new PR, it makes little sense to continue here when the last activity was more than 2 years ago.

PR #8163, opened 2021-12-24
Improve `unwrap_or_else_default` when handling `unwrap_or_else(XXX::new)`

pmnoxx:
changelog:  change `unwrap_or_else_default` to work with std constructors like `Vec::new`, `HashSet::new`, `HashMap::new`.

Notes:
- Code to handle detecting those constructors is already there. I moved it out to `is_default_equivalent_call`


Manishearth (approved):

xFrednet:
The changelog entry has to be on the same line like the `changelog` label. Now it should work without a problem :upside_down_face: 

@bors retry

PR #8168, opened 2021-12-25
Update pulldown-cmark version

GuillaumeGomez:
Linked to https://github.com/rust-lang/rust/pull/92252.


xFrednet:
Thanks for the PR the change look good to me. For Clippy we usually update the dependencies in rust-lang/rust and then sync it back into this repo. We do this to ensure that it builds in rust-lang/rust and that we don't use different versions of crates than other tools. Would you mind making this change in rust-lang/rust? You can also add it to rust-lang/rust#92252 and ping me to approve the Clippy changes there :upside_down_face:.

GuillaumeGomez:
I see, will do it like this then!

PR #8170, opened 2021-12-25
new lint: `init-numbered-fields`

### LINKED ISSUE #7985: Lint suggestion: explicit construction of tuple struct by named fields
### What it does

*What does this lint do?*

Detects if someone has decided to create a tuple struct by its named (.0, .1 etc) fields.

See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8006290eca4f50c8e6dd065b60ce0cb7) for proof that this actually works.

### Categories (optional)

- Kind: `clippy::style`

Constructing tuple structs by their named fields is extremely unidiomatic, verbose and unclear.

### Drawbacks

May be spurious if users are using struct update syntax with tuple structs.

### Example

```rust
struct MyTuple(isize, isize);

let my_tuple = MyTuple {
  0: 1,
  1: 2,
};
```

Could be written as:

```rust
struct MyTuple(isize, isize);

let my_tuple = MyTuple(0, 1);
```

llogiq:
This fixes #7985.

r? @xFrednet 

---

changelog: new lint: [`init_numbered_fields`]

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -1,13 +1,15 @@
 // error-pattern:cargo-clippy
 
+#![feature(binary_heap_into_iter_sorted)]
 #![feature(box_patterns)]
+#![feature(control_flow_enum)]
 #![feature(drain_filter)]
 #![feature(in_band_lifetimes)]
+#![feature(iter_intersperse)]
+#![feature(let_else)]
 #![feature(once_cell)]
 #![feature(rustc_private)]
 #![feature(stmt_expr_attributes)]
-#![feature(control_flow_enum)]
-#![feature(let_else)]
 #![recursion_limit = "512"]
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
 #![allow(clippy::missing_docs_in_private_items, clippy::must_use_candidate)]
@@ -242,6 +244,7 @@ mod indexing_slicing;
 mod infinite_iter;
 mod inherent_impl;
 mod inherent_to_string;
+mod init_numbered_fields;
 mod inline_fn_without_body;
 mod int_plus_one;
 mod integer_division;
@@ -854,6 +857,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_early_pass(|| Box::new(octal_escapes::OctalEscapes));
     store.register_late_pass(|| Box::new(needless_late_init::NeedlessLateInit));
     store.register_late_pass(|| Box::new(return_self_not_must_use::ReturnSelfNotMustUse));
+    store.register_late_pass(|| Box::new(init_numbered_fields::NumberedFields));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/tests/ui/numbered_fields.rs
+++ b/tests/ui/numbered_fields.rs
@@ -1,0 +1,41 @@
+//run-rustfix
+#![warn(clippy::init_numbered_fields)]
+
+#[derive(Default)]
+struct TupleStruct(u32, u32, u8);
+
+// This shouldn't lint because it's in a macro
+macro_rules! tuple_struct_init {
+    () => {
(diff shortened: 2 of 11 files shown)


clippy_lints/src/numbered_fields.rs
```diff
@@ -0,0 +1,77 @@
...
+
+declare_lint_pass!(NumberedFields => [NUMBERED_FIELDS]);
+
+impl<'tcx> LateLintPass<'tcx> for NumberedFields {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
+        if let ExprKind::Struct(path, fields, None) = e.kind {
+            if !fields.is_empty()
+                && fields
+                    .iter()
+                    .all(|f| f.ident.as_str().as_bytes().iter().all(u8::is_ascii_digit))
```
llogiq, line 49: I wondered if I should introduce a `fn is_numbered_field(&FieldExpr) -> bool`, but it didn't seem to be worth the hassle.

alercah:
🎉 

Will this be suppressed in macros by default? I think that macros might be a legitimate use of numbered fields to let them treat tuples like other structs.

llogiq:
Good point, I'll add a macro check.

clippy_lints/src/lib.rs
```diff
@@ -1,13 +1,15 @@
 // error-pattern:cargo-clippy
 
+#![feature(binary_heap_into_iter_sorted)]
 #![feature(box_patterns)]
+#![feature(control_flow_enum)]
 #![feature(drain_filter)]
 #![feature(in_band_lifetimes)]
+#![feature(iter_intersperse)]
```
xFrednet, line 8: Have you used this somewhere in your lint logic?

clippy_lints/src/numbered_fields.rs
xFrednet: Agreed. I'm considering if an iterator over chars would be more readable or faster etc but I think this is the best solution :+1:

tests/ui/numbered_fields.rs
```diff
@@ -0,0 +1,34 @@
...
+
+fn main() {
+    let tuple_struct = TupleStruct::default();
+
+    // This should lint
+    let _ = TupleStruct {
+        0: 1u32,
+        1: 42,
+        2: 23u8,
+    };
```
xFrednet, line 22: Could you also add a test where the fields are out of order? Something like this:

```rs
    let _ = TupleStruct {
        2: 23u8,
        0: 1u32,
        1: 42,
    };
```

xFrednet (commented):
TIL that you can construct tuples like structs ^^. The implementation looks good to me, it would be good if you could add one more test to see if the suggestion is correct for out of order fields :upside_down_face:. Then it's ready to be merged :+1:

xFrednet:
Do we maybe want to change the lint name to something more precise? When I first read it, I thought this would be linting struct definitions like `struct A {1: u32, 2: u32}` and not the struct like initialization for tuples. I like the name `struct-like-init-for-tuples`

llogiq:
Numbers are not valid field names in struct definitions, so I don't think we need to explicitly rule out that case. And I'd like to keep it short. If we were to be exact, it would need to be called `struct-init-for-tuple-struct` or something like it.

xFrednet:
If we want to keep it short then I have one more suggestion what about `init_numbered_fields`? :upside_down_face:

llogiq:
Ok, renamed. I kept the test's name though. Is that OK with you?

xFrednet (approved):
> Ok, renamed. I kept the test's name though. Is that OK with you?

Sure, thank you for renaming the lint, I like this name more!

xFrednet:
And thank you for the lint implementation! :+1: 

@bors r+

llogiq:
That is weird. I didn't have any problems when running the tests locally.

xFrednet:
Bors is acting up a bit today, the last PR merged in rust-lang/rust was also yesterday. I'm guessing we can try another retry

@bors retry

PR #8174, opened 2021-12-26
new lint: `missing-spin-loop`

### LINKED ISSUE #7809: New lint: loop with atomics without other side effects
### What it does

The lint would check that a loop has no side effects other than atomic access and warn to park the thread or call `std::hint::spin_loop()`.

### Categories (optional)

- Kind: perf

If the thread is parked it will stop consuming resources and avoid priority inversion.
If `spin_loop()` hint is used the code will consume less power on some architectures or perform better.
Presence of `spin_loop()` could also better communicate to reviewer that author did **not** forget to park the thread.

### Drawbacks

None that I know of but I'm not very knowledgeable about the topic.
In some cases it may be better to use futex instead of `thread::park()`.

### Example

```rust
while some_atomic.load(atomic::Ordering::Acquire) == NOT_READY {}
```

Could be written as:

```rust
while some_atomic.load(atomic::Ordering::Acquire) == NOT_READY {
    std::hint::spin_loop();
}
```

or

```rust
while some_atomic.load(atomic::Ordering::Acquire) == NOT_READY {
    thread::park();
}
```

with `thread::unpark()` at the place where a value is stored into atomic

### Additional context

Loosely related: https://github.com/rust-lang/miri/issues/1388#issuecomment-940464052

llogiq:
This fixes #7809. I went with the shorter name because the function is called `std::hint::spin_loop`. It doesn't yet detect `while let` loops. I left that for a follow-up PR.

---

changelog: new lint: [`missing_spin_loop`]

### DIFF
--- a/clippy_lints/src/loops/missing_spin_loop.rs
+++ b/clippy_lints/src/loops/missing_spin_loop.rs
@@ -1,0 +1,56 @@
+use super::MISSING_SPIN_LOOP;
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::is_no_std_crate;
+use rustc_errors::Applicability;
+use rustc_hir::{Block, Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_middle::ty;
+use rustc_span::sym;
+
+fn unpack_cond<'tcx>(cond: &'tcx Expr<'tcx>) -> &'tcx Expr<'tcx> {
+    match &cond.kind {
+        ExprKind::Block(
+            Block {
+                stmts: [],
+                expr: Some(e),
+                ..
+            },
+            _,
+        )
+        | ExprKind::Unary(_, e) => unpack_cond(e),
+        ExprKind::Binary(_, l, r) => {
+            let l = unpack_cond(l);
+            if let ExprKind::MethodCall(..) = l.kind {
+                l
+            } else {
+                unpack_cond(r)
+            }
+        },
+        _ => cond,
+    }
+}
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, cond: &'tcx Expr<'_>, body: &'tcx Expr<'_>) {
+    if_chain! {
+        if let ExprKind::Block(Block { stmts: [], expr: None, ..}, _) = body.kind;
+        if let ExprKind::MethodCall(method, [callee, ..], _) = unpack_cond(cond).kind;
+        if [sym::load, sym::compare_exchange, sym::compare_exchange_weak].contains(&method.ident.name);
+        if let ty::Adt(def, _substs) = cx.typeck_results().expr_ty(callee).kind();
+        if cx.tcx.is_diagnostic_item(sym::AtomicBool, def.did);
+        then {
+            span_lint_and_sugg(
+                cx,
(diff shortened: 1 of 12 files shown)


clippy_lints/src/loops/missing_spin_loop.rs
```diff
@@ -0,0 +1,33 @@
...
+//use rustc_middle::ty;
+use rustc_span::sym;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, cond: &'tcx Expr<'_>, body: &'tcx Expr<'_>) {
+    if_chain! {
+        if let ExprKind::Block(Block { stmts: [], expr: None, ..}, _) = body.kind;
+        if let ExprKind::MethodCall(method, _, [_callee, _ordering], _) = cond.kind;
+        if method.ident.name == sym::load;
+        //if let ty::Adt(def, _substs) = cx.typeck_results().expr_ty(callee).kind();
+        //if cx.tcx.is_diagnostic_item(sym::AtomicBool, def.did);
```
flip1995, line 16: Without this, this will lint on all calls to `load` on an empty loop. I guess this is fine, because empty loops are probably rare and regardless of the type `load` is called on, this is suspicious and probably warrants a `hint::spin_loop()`.

Please remove those lines and possibly replace them with a comment.

flip1995 (approved):
LGTM overall, just a NIT.

clippy_lints/src/loops/missing_spin_loop.rs
llogiq: I'd actually like to check this, but I think it fails because `AtomicBool` is missing the `rustc_diagnostic_item` attribute. I've opened a PR to add it.

flip1995: Doesn't this lint also apply to `AtomicU8` and all the integer types? Or do you plan to check for all of them?

llogiq:
This is blocked on [rust/#92870](https://github.com/rust-lang/rust/pull/92870).

llogiq:
No longer blocked. No idea why the base test was cancelled, but everything seems to be in order here locally.

r? @flip1995

clippy_lints/src/loops/missing_spin_loop.rs
```diff
@@ -0,0 +1,57 @@
...
+    }
+}
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, cond: &'tcx Expr<'_>, body: &'tcx Expr<'_>) {
+    if_chain! {
+        if let ExprKind::Block(Block { stmts: [], expr: None, ..}, _) = body.kind;
+        if let ExprKind::MethodCall(method, [callee, ..], _) = unpack_cond(cond).kind;
+        if [sym::load, sym::compare_exchange, sym::compare_exchange_weak].contains(&method.ident.name);
+        if let ty::Adt(def, _substs) = cx.typeck_results().expr_ty(callee).kind();
+        //FIXME: The AtomicBool sym is apparently not correctly assigned, once it is, uncomment:
```
flip1995, line 39: I think this fixme can now be removed.

flip1995 (approved):
LGTM. r=me after removing the `FIXME`.

PR #8183, opened 2021-12-28
Limit the ``[`identity_op`]`` lint to integral operands.

alex-ozdemir:
changelog: limit ``[`identity_op`]`` to integral operands

In the ``[`identity_op`]`` lint, if the operands are non-integers, then the lint is likely
wrong.

### DIFF
--- a/clippy_lints/src/identity_op.rs
+++ b/clippy_lints/src/identity_op.rs
@@ -71,8 +71,8 @@ fn is_allowed(cx: &LateContext<'_>, cmp: BinOp, left: &Expr<'_>, right: &Expr<'_
 }
 
 fn check(cx: &LateContext<'_>, e: &Expr<'_>, m: i8, span: Span, arg: Span) {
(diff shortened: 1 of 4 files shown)


camsteffen:
Do you have an example false positive that this fixes?

alex-ozdemir:
Ah yes. [good_lp](https://github.com/rust-or/good_lp) overloads the shift operators to construct bounds in an ILP. See, for example [this overload](https://docs.rs/good_lp/latest/good_lp/struct.Expression.html#impl-Shl%3CRHS%3E).

I encountered the false positive [here](https://github.com/circify/circ/blob/master/src/target/ilp/trans.rs#L466).

camsteffen:
I see. Can you add a test for that?

Also I think you should use `peel_refs().is_integral()` since there can be `+ &0`.

alex-ozdemir:
Good suggestions.

First commit avoids ignoring refs.

Second commit actually catches `x << &0`, by peeling refs around constants.

camsteffen:
Thanks!

@bors r+

PR #8185, opened 2021-12-28
`needless_return` suggest return unit type on void returns

### LINKED ISSUE #8177: `needless_return` suggests empty block instead of void return
### Summary

`needless_return` suggests `{}` instead of `()` for empty returns in match statements

### Reproducer

I tried this code:

```rust
fn foo(x: bool) {
    match x {
        true => (),
        false => return, 
    }
}
```

I expected to see this happen: clippy suggests replacing return with `()`

Instead, this happened: clippy suggests replacing `return` with `{}`

```
 --> src/main.rs:8:18
  |
8 |         false => return,
  |                  ^^^^^^ help: replace `return` with an empty block: `{}`
  |
```

### Version

```text
rustc 1.59.0-nightly (34926f0a1 2021-12-22)
binary: rustc
commit-hash: 34926f0a1681458588a2d4240c0715ef9eff7d35
commit-date: 2021-12-22
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

@rustbot label +L-suggestion

dswij:
closes #8177 

previously, `needless_return` suggests an empty block `{}` to replace void `return` on match arms, this PR improve the suggestion by suggesting a unit instead.

changelog: `needless_return` suggests `()` instead of `{}` on match arms

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -73,6 +73,7 @@ declare_clippy_lint! {
 enum RetReplacement {
     Empty,
     Block,
+    Unit,
 }
 
 declare_lint_pass!(Return => [LET_AND_RETURN, NEEDLESS_RETURN]);
@@ -212,7 +213,7 @@ fn check_final_expr<'tcx>(
         // (except for unit type functions) so we don't match it
         ExprKind::Match(_, arms, MatchSource::Normal) => {
             for arm in arms.iter() {
-                check_final_expr(cx, arm.body, Some(arm.body.span), RetReplacement::Block);
+                check_final_expr(cx, arm.body, Some(arm.body.span), RetReplacement::Unit);
             }
         },
         ExprKind::DropTemps(expr) => check_final_expr(cx, expr, None, RetReplacement::Empty),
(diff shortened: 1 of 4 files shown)


llogiq:
I'm not completely happy with calling an empty tuple a "unit type" – you're returning a value, not a type. Please change to either "unit value", "unit expression" or "empty tuple".

Otherwise this looks good to merge.

llogiq:
Thank you! @bors r+ squash

llogiq:
This was actually merged, but github UI seems to have some problems.

xFrednet:
It would be interesting to see what happens with a retry, but I think the best solution would be to just close it :sweat_smile:

llogiq:
I did this on another PR, it was merged twice. So as it's merged I'm going to close this PR manually. Thanks @dswij and sorry for the inconvenience.

dswij:
> I did this on another PR, it was merged twice. So as it's merged I'm going to close this PR manually. Thanks @dswij and sorry for the inconvenience.

Nothing to be sorry about here, thanks for the review :slightly_smiling_face:

Issue #7830, opened 2021-10-17
Warn when intended Deref is not triggered correctly


jamesmcm:
### What it does

Aim would be to warn when intended Deref is not triggered correctly.

E.g. for https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c78429c73f85949bd3799a9fdcaa4d74

```rust
use std::ops::Deref;

struct TestType;

impl Deref for TestType {
    type Target = str;

    fn deref(&self) -> &Self::Target {
        "test"
    }
}

impl std::fmt::Display for TestType {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(f, "{}", &**self)
    }
}

fn main() {
    let test = TestType {};
    println!("Hello, world!, {}", test);
}
```

If we write:
```rust
        write!(f, "{}", &self)
```

or

```rust
        write!(f, "{}", &*self)
```

Then the Deref to str is not triggered and we got a stack overflow due to recursively calling `Display::fmt`

At the moment there are no clippy or rustc warnings or errors, and we just get a stack overflow.

### Categories (optional)

- Kind: `clippy::suspicious`

*What is the advantage of the recommended code over the original code*

The recommended code correctly triggers the Deref and avoids the stack overflow.

camsteffen:
Duplicate of #2691

Issue #2691, opened 2018-04-21
new lint: Recursive `Display` impl


oherrala:
It's possible to implement Display trait in such a way which recursively calls itself until stack is exhausted. There's probably more variations how to encounter this bug. I provide one example.

Consider the following code:

```rust
use std::fmt::{Display, Formatter, Result};

struct Foo;

impl Display for Foo {
    fn fmt(&self, f: &mut Formatter) -> Result {
        write!(f, "{}", self)
    }
}

fn main() {
    println!("{}", Foo);
}
```

Compiles perfectly fine, but there's stack trace at run time:

```console
$ cargo run -q

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
Abort trap: 6
```

The bug is in line `write!(f, "{}", self)` which recursively calls itself by trying to use Display trait implementation for self.

clarcharr:
Ah right, the lint only checks one layer of function calls, so `fn a() { b() } fn b() { a() }` doesn't lint.

mimoo:
any update here :) ? This would be an awesome lint!

flip1995:
There is the [`to_string_in_display`](https://rust-lang.github.io/rust-clippy/master/index.html#to_string_in_display) lint, which already checks for one case. This lint could be enhanced to also check for things like using the formatting machinery inside the `Display` implementation.

giraffate:
It looks like that the [`recursive_format_impl`](https://rust-lang.github.io/rust-clippy/master/index.html#recursive_format_impl) covers this, so I'm closing this.

PR #8188, opened 2021-12-28
new lint: `recursive_format_impl`

jamesmcm:
The to_string_in_display lint is renamed to recursive_format_impl
A check is added for the use of self formatted with Display or Debug inside any format string in the same impl
The to_string_in_display check is kept as is - like in the format_in_format_args lint

This is my first contribution so please check it for better / more idiomatic checks + error messages. Note the format macro paths are shared with the `format_in_format_args` lint - maybe those could be moved to clippy utils too.

This relates to issues #2691 and #7830 

------

changelog: Renamed `to_string_in_display` lint to [`recursive_format_impl`] with new check for any use of self as Display or Debug inside the same format trait impl.


mikerite:
It seems like lint doesn't handle nested structures correctly.

Here is a false positive:

```
#![allow(dead_code)]
#![warn(clippy::all)]

use std::fmt;

enum Tree {
    Leaf,
    Node(Vec<Tree>),
}

impl fmt::Display for Tree {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        match self {
            Tree::Leaf => write!(f, "*"),
            Tree::Node(children) => {
                write!(f, "(")?;
                for child in children.iter() {
                    write!(f, "{},", child)?;
                }
                write!(f, ")")
            }
        }
    }
}

fn main() {}

```

```
warning: using `self` in `fmt::Display` implementation might lead to infinite recursion
  --> src/main.rs:18:21
   |
18 |                     write!(f, "{},", child)?;
   |                     ^^^^^^^^^^^^^^^^^^^^^^^
   |
  --> src/main.rs:2:9
   |
2  | #![warn(clippy::all)]
   |         ^^^^^^^^^^^
   = note: `#[warn(clippy::recursive_display_impl)]` implied by `#[warn(clippy::all)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#recursive_display_impl
   = note: this warning originates in the macro `$crate::format_args` (in Nightly builds, run with -Z macro-backtrace for more info)
```

jamesmcm:
Thanks, seems it is because enum variants aren't distinct types so there it is just treating Tree as the same type as self?

I'll add it as a test case and try to work around it, thanks!

jamesmcm:
Fixed by dealing with the single deref of `&self` to `self` inside `fmt()` as a special case, instead of just checking the types match (that caused the false positive above).

camsteffen:
I realized with #8253 that this lint could work for any formatting trait (e.g. Debug). The two lints can share a LintPass and a naming convention. For the common part of the name I'm thinking either "fmt" or "format_impl".

jamesmcm:
It's a good idea @camsteffen but note the implementations would differ a fair bit since atm this uses:
https://github.com/rust-lang/rust-clippy/pull/8188/files#diff-37e08397df0148cab790ac270ccda86159451eb034ef01a7b59bc46b00074ca4R168-R174

```rust
                    if !arg.is_display() {
                        continue;
                    }
```
to filter the format args, unfortunately there isn't an equivalent for debug atm. See the source code here - https://doc.rust-lang.org/nightly/nightly-rustc/src/clippy_utils/higher.rs.html#680-688

Also note that it's okay to use Debug in the Display Impl and vice versa.

I'll try to look at adding the above for Debug when I get some time free.

camsteffen:
It would be perfectly fine to leave other traits as a future enhancement. I was just thinking about the lint name since we like to get that right the first time.

jamesmcm:
@camsteffen renamed to `recursive_format_trait_impl` and rebased on upstream (there were some changes to the FormatArg macro parsing).

atm I only check for Debug and Display, but one could add the others (Binary, etc.) pretty easily. It also doesn't check the case where there's a reference cycle i.e. Debug calls Display, and Display calls Debug - we could detect this if we store more information about the other format trait impls as we do the pass, but it doesn't seem like something that is likely to come up in real code anyway.

camsteffen:
I think `recursive_format_impl` would be clear enough and shorter?

giraffate:
@camsteffen You've already reviewed some of this, so can you review it as is?

camsteffen:
Okay.

r? @camsteffen

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+    correctness,
+    "Format trait method called while implementing the same Format trait"
+}
+
+#[derive(Default)]
+pub struct RecursiveFormatImpl {
+    // Whether we are inside Display or Debug trait impl - None for neither
+    format_trait_impl: Option<ImplTrait>,
+    // hir_id of self parameter of method inside Display Impl - i.e. fmt(&self)
+    self_hir_id: Option<HirId>,
```
camsteffen, line 82: I think you can avoid storing this and just check whether the variable name is `kw::SelfLower`.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+                    ImplTrait::Debug => {
+                    if arg.format_trait != sym::Debug {
+                        continue;
+                    }},
+
+                    };
+                    check_format_arg_self(cx, expr, self_hir_id, arg, impl_trait);
+                }
+            }
+    }
```
camsteffen, line 197: nit: extra indentation

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+                "using `to_string` in `fmt::Display` implementation might lead to infinite recursion",
+            );
+        }
+    }
+}
+
+fn check_self_in_format_args(cx: &LateContext<'_>, expr: &Expr<'_>, self_hir_id: HirId, impl_trait: ImplTrait) {
+    // Check each arg in format calls - do we ever use Display on self (directly or via deref)?
+    if_chain! {
+            if let Some(format_args) = FormatArgsExpn::parse(cx, expr);
```
camsteffen, line 167: We have some new macro utils that should be used here. First call `root_macro_call_first_node` and then `FormatArgsExpn::parse` (this will avoid some calls to `FormatArgsExpn::parse` which can be expensive). The output of `root_macro_call_first_node` can be used instead of `outermost_expn_data`.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+            let expr_expn_data = expr.span.ctxt().outer_expn_data();
+            let outermost_expn_data = outermost_expn_data(expr_expn_data);
+            if let Some(macro_def_id) = outermost_expn_data.macro_def_id;
+            if FORMAT_MACRO_PATHS
+                .iter()
+                .any(|path| match_def_path(cx, macro_def_id, path))
+                || FORMAT_MACRO_DIAG_ITEMS
+                    .iter()
+                    .any(|diag_item| cx.tcx.is_diagnostic_item(*diag_item, macro_def_id));
+            if let ExpnKind::Macro(_, _name) = outermost_expn_data.kind;
```
camsteffen, line 177: This check will be unnecessary after addressing the above comment.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+            if let Some(args) = format_args.args();
+            then {
+                for (_i, arg) in args.iter().enumerate() {
```
camsteffen, line 180:
```suggestion
                for arg in args {
```

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+                    // In Display, we only care about Display (it is okay to use Debug)
+                        ImplTrait::Display => {
+                    if arg.format_trait != sym::Display {
+                        continue;
+                    }},
+                    // In Debug, we only care about Debug (it is okay to use Display and ToString)
+                    ImplTrait::Debug => {
+                    if arg.format_trait != sym::Debug {
+                        continue;
+                    }},
```
camsteffen, line 191: You could have `ImplTrait::name() -> Symbol` and then just do `impl_trait.name() == arg.format_trait`.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+                    "using `self` as Display in `fmt::Display` implementation might lead to infinite recursion",
+                );
+            },
+            ImplTrait::Debug => {
+                span_lint(
+                    cx,
+                    RECURSIVE_FORMAT_IMPL,
+                    expr.span,
+                    "using `self` as Debug in `fmt::Debug` implementation might lead to infinite recursion",
+                );
```
camsteffen, line 227: The two `span_lint` calls can be collapsed.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+    cx: &LateContext<'_>,
+    expr: &Expr<'_>,
+    self_hir_id: HirId,
+    arg: &FormatArgsArg<'_>,
+    impl_trait: ImplTrait,
+) {
+    // Handle multiple dereferencing of references e.g. &&self
+    // Handle single dereference of &self -> self that is equivalent (i.e. via *self in fmt() impl)
+    // Since the argument to fmt is itself a reference: &self
+    let reference = single_deref(deref_expr(arg.value));
```
camsteffen, line 210: You should handle *any* number of borrow/derefs (loop and `match e.kind`). Just make sure that the expression behind the deref operator is a reference.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+        then {
+            // Is it for Display trait?
+            if match_def_path(cx, did, &paths::DISPLAY_TRAIT) {
+                Some(ImplTrait::Display)
+            }
+            // Is it for Debug trait?
+            else if match_def_path(cx, did, &paths::DEBUG_TRAIT) {
+                Some(ImplTrait::Debug)
+            } else {
+            None
```
camsteffen, line 263: Note that code inside `if_chain!` is not automatically formatted.

camsteffen, line 263: Prefer diagnostic items over paths whenever possible. Here you should use `cx.tcx.get_diagnostic_name`.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+    // Since the argument to fmt is itself a reference: &self
+    let reference = single_deref(deref_expr(arg.value));
+    if path_to_local_id(reference, self_hir_id) {
+        match impl_trait {
+            ImplTrait::Display => {
+                span_lint(
+                    cx,
+                    RECURSIVE_FORMAT_IMPL,
+                    expr.span,
+                    "using `self` as Display in `fmt::Display` implementation might lead to infinite recursion",
```
camsteffen, line 218:
```suggestion
                    "using `self` as `Display` in `impl Display` will cause infinite recursion",
```

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+        if let Some(expr_def_id) = cx.typeck_results().type_dependent_def_id(expr.hir_id);
+        if is_diag_trait_item(cx, expr_def_id, sym::ToString);
+        // Is the method is called on self
+        if path_to_local_id(self_arg, self_hir_id);
+        then {
+            span_lint(
+                cx,
+                RECURSIVE_FORMAT_IMPL,
+                expr.span,
+                "using `to_string` in `fmt::Display` implementation might lead to infinite recursion",
```
camsteffen, line 158: I think we can be confident about it leading to infinite recursion. Unless you have a counterexample in mind?
```suggestion
                "using `self.to_string` in `fmt::Display` implementation will cause infinite recursion",
```

camsteffen (changes requested):
This is looking good! Nice tests. I have some comments but it's pretty minor stuff.

It would be nice to have a separate commit to rename the lint. Also we need to add a `register_renamed` call and add to the `rename.rs` test.

Not necessary to do now but it would be easier to review with a separate commit that renames the lint.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+}
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for recursive use of `Display` or `Debug` traits inside their implementation.
+    ///
+    /// ### Why is this bad?
+    /// This is unconditional recursion and so will lead to infinite
+    /// recursion and a stack overflow.
```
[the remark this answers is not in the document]
jamesmcm: It was copied from the other lint :joy_cat: I'll just remove the first clause.

clippy_lints/src/recursive_format_impl.rs
jamesmcm: Cool, this works. I was wondering if there's ever a case where this is less powerful? But it seems all the tests pass either way :slightly_smiling_face:

clippy_lints/src/recursive_format_impl.rs
jamesmcm: But you can have a case like:
```rust
struct J4 {}

impl std::ops::Deref for J4 {
    type Target = str;

        "test"
    }
}

impl std::fmt::Display for J4 {
    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
        write!(f, "{}", &&**&&*self)
    }
}
```

I'm not sure how to catch this too since we have two derefs next to each other, so it's not as simple as only handling if it the next part is a reference :confused: 

This is unlikely to come up in practice though, and would trigger the immediate deref lint anyway.

clippy_lints/src/recursive_format_impl.rs
jamesmcm: Which HIR node / expr should be checked here?

It works if I just use:
```
        if let Some(macro_call) = root_macro_call(expr.span);
```

But trying to add the first node check with the `expr` results in it always being None.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,269 @@
...
+    &paths::ASSERT_EQ_MACRO,
+    &paths::ASSERT_MACRO,
+    &paths::ASSERT_NE_MACRO,
+    &paths::EPRINT_MACRO,
+    &paths::EPRINTLN_MACRO,
+    &paths::PRINT_MACRO,
+    &paths::PRINTLN_MACRO,
+    &paths::WRITE_MACRO,
+    &paths::WRITELN_MACRO,
+];
```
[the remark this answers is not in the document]
jamesmcm: Done, but I'll leave making the changes to the FormatArgs lint to another PR so as not to make this huge.

How could we re-use the lint pass, when here we need to check we are inside the impl Display or Debug, etc. ?

clippy_lints/src/recursive_format_impl.rs
camsteffen: For one of those deref operators, the expression behind it is not a reference. `expr_ty(e).is_ref()`.

jamesmcm: Fixed, so the only remaining issue is `root_macro_call` vs `root_macro_call_first_node` - which expr should be passed if the latter is used? (passing the macro expr resulted in no matches).

clippy_lints/src/recursive_format_impl.rs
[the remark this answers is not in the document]
camsteffen: > Done, but I'll leave making the changes to the FormatArgs lint to another PR so as not to make this huge.

I still would like to avoid duplicating the list in this PR. I small refactor can be done in the other lint.

> How could we re-use the lint pass, when here we need to check we are inside the impl Display or Debug, etc. ?

It seems like it should be possible to merge the two lint passes and maintain all cases. But I may be underestimating the difficulty. Again I'll leave it up to you to decide if that ultimately makes sense to do.

clippy_lints/src/recursive_format_impl.rs
camsteffen: Okay, I think the problem is that `format_args!` is not the root macro call. It is nested within a format macro. So you should check the outer macro with `root_macro_call_first_node` and `is_format_macro` and then use `FormatArgsExpn::find_nested` to find the nested `format_args!` call.

clippy_lints/src/recursive_format_impl.rs
```diff
@@ -0,0 +1,204 @@
...
+        // Get the hir_id of the object we are calling the method on
+        if let ExprKind::MethodCall(path, _, [ref self_arg, ..], _) = expr.kind;
+        // Is the method to_string() ?
+        if path.ident.name == sym!(to_string);
+        // Is the method a part of the ToString trait? (i.e. not to_string() implemented
+        // separately)
+        if let Some(expr_def_id) = cx.typeck_results().type_dependent_def_id(expr.hir_id);
+        if is_diag_trait_item(cx, expr_def_id, sym::ToString);
+        // Is the method is called on self
+        if path_to_local(self_arg).map(|x| map.name(x)) == Some(kw::SelfLower);
```
camsteffen, line 121: This avoids a lookup by ID.

```suggestion
        if let ExprKind::Path(QPath::Resolved(_, path)) = self_arg.kind;
        if let [segment] = path.segments;
        if segment.ident.name == kw::SelfLower;
```

clippy_lints/src/recursive_format_impl.rs
camsteffen: Oh I'm realizing this late but we also have an existing util for this - `peel_ref_operators`.

(I responded in the other thread about `root_macro_call`)

clippy_utils/src/format_macros.rs
```diff
@@ -0,0 +1,28 @@
...
+const FORMAT_MACRO_DIAG_ITEMS: &[Symbol] = &[sym::format_macro, sym::std_panic_macro];
+
+pub fn is_format_macro(cx: &LateContext<'_>, macro_def_id: DefId) -> bool {
+    FORMAT_MACRO_PATHS
+        .iter()
+        .any(|path| match_def_path(cx, macro_def_id, path))
+        || FORMAT_MACRO_DIAG_ITEMS
+            .iter()
+            .any(|diag_item| cx.tcx.is_diagnostic_item(*diag_item, macro_def_id))
+}
```
camsteffen, line 28: This can live in `clippy_utils::macros` instead of creating a new module.

clippy_lints/src/recursive_format_impl.rs
jamesmcm: Nice, this also has the benefit of providing the outer macro name for the error message :slightly_smiling_face:

clippy_lints/src/recursive_format_impl.rs
jamesmcm: Is there any documentation on this, and the use of paths vs. diagnostic items for example? It's not obvious when writing it as a newcomer.

clippy_lints/src/recursive_format_impl.rs
jamesmcm: Thanks, it's perfect for this case.

clippy_lints/src/recursive_format_impl.rs
camsteffen: Migrating to diagnostic items is a long-term effort and there's a pinned issue about it. Path nodes to local variables seem weird because Path nodes are general enough to handle complex paths to external items. I think we could use a section on path nodes in the contributing docs.

clippy_lints/src/recursive_format_impl.rs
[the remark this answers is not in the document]
camsteffen: @jamesmcm did you decide against merging the lint passes or still looking into it?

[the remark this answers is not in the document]
jamesmcm: I took a look at it, in theory it should be possible, it's not trivial though since the format_args case also needs to work when the macro isn't nested. It might be best to do in another PR though.

Btw is there any way to share the pass without combining the source files (and error files)?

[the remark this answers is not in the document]
camsteffen: If we combine them, I think it would be better to do it in this PR since it would be less churn and reviewing. The tests can still be separated per lint.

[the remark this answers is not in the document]
jamesmcm: Cool, looking at it in detail I don't think it is worth combining them. Since handling the nested macro difference means the code isn't really shared between the non format trait impl and format trait impl cases, so it's more just moving it around and I don't think that's worth merging them (since at that point you might as well merge #8253 too for example).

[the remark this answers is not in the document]
jamesmcm: @camsteffen considering this, I think this is mergeable as is?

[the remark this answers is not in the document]
camsteffen: Sorry for the delay.

> (since at that point you might as well merge #8253 too for example).

I am planning to have that lint merged with this one at least. But if that doesn't change your decision, yes this is ready to merge!

camsteffen:
Thanks for your patience, and great first contribution!

@bors r+

PR #8190, opened 2021-12-28
Stop using `in_band_lifetimes`

camsteffen:
Per rust-lang/rust#91867

changelog: none


flip1995:
Thanks! Also for the unnecessary-lifetime cleanup that came with this!

@bors r+

PR #8192, opened 2021-12-29
Expand mismatched_target_os lint to check macro position usage

### LINKED ISSUE #7839: `mismatched_target_os` false negative when using the `cfg` macro instead of the attribute.
Lint name: `mismatched_target_os `

I tried this code:

```rust
fn main() {
    // Clippy throws error
    //#[cfg(linux)]
    //println!("Hello, Linux!");

    // Clippy doesn't care
    if cfg!(linux) {
        println!("Hello, Linux!");
    }
}
```

I expected to see the error `operating system used in target family position` here like you would see if you used the attribute `#[cfg(linux)]`

Instead, clippy doesn't give any error.

### Meta

```
rustc 1.52.1 (9bc8c42bb 2021-05-09)
binary: rustc
commit-hash: 9bc8c42bb2f19e745a63f3445f1ac248fb015e53
commit-date: 2021-05-09
host: x86_64-pc-windows-msvc
release: 1.52.1
LLVM version: 12.0.0
```

fanzeyi:
changelog: improving [`mismatched_target_os`] to check `cfg!` macro usage.

fixes #7839


fanzeyi:
Oops, there are some panics with the entire test suite. I am working on fixing those now. Sorry about it.

flip1995:
I'm not sure if doing this is a good idea. We try to get rid of pre-expansion passes, since they are semi-deprecated in rustc. This would add another dependency on pre-expansion passes. I lean towards not doing this.

flip1995:
Even though there wasn't much discussion on the Zulip thread, I think we shouldn't introduce more code into pre-expansion passes.

FYI: This kind of check is planned to get implemented in rustc: https://github.com/rust-lang/rust/issues/82450, with an WIP PR already open here: https://github.com/rust-lang/rust/pull/89346 

I'm therefore closing this PR. Thanks for the work you've done so far! (The rustc PR seems stale, so if you want to continue to work on something in that direction, you may offer you're expertise there)

PR #8198, opened 2021-12-30
Remove method_call! macro

camsteffen:
This is possible now that `SymbolStr` is removed from rustc.

changelog: none


flip1995:
@bors r+

Thanks! Good catch!

Issue #8148, opened 2021-12-20
`unnecessary_to_owned` suggestion causes error when iterator is re-borrowed mutably within a loop


teor2345:
### Summary

When running:
```sh
cargo clippy --fix
```

On https://github.com/ZcashFoundation/zebra/tree/410133435e875336875aab32c338a284470ceab4/zebra-chain

I get an error:
> after fixes were automatically applied the compiler reported errors

### Lint Name

unnecessary_to_owned

### Reproducer

I tried this code:

```rust
for mut input in transaction.inputs().to_vec().into_iter() { 
    // use input and &mut transaction
}
```

https://github.com/ZcashFoundation/zebra/blob/410133435e875336875aab32c338a284470ceab4/zebra-chain/src/block/arbitrary.rs#L554-L578

I saw this happen:

```
after fixes were automatically applied the compiler reported errors within these files:

  * zebra-chain/src/block/arbitrary.rs
...
The following errors were reported:                                                                                                                                                            
error[E0502]: cannot borrow `transaction` as mutable because it is also borrowed as immutable                                                                                                  
   --> zebra-chain/src/block/arbitrary.rs:559:17                                                                                                                                               
    |                                                                                                                                                                                          
554 |     for mut input in transaction.inputs().iter().cloned() {                                                                                                                              
    |                      ------------------------------------                                                                                                                                
    |                      |                                                                                                                                                                   
    |                      immutable borrow occurs here                                                                                                                                        
    |                      immutable borrow later used here                                                                                                                                    
...                                                                                                                                                                                            
559 |                 &mut transaction,                                                                                                                                                        
    |                 ^^^^^^^^^^^^^^^^ mutable borrow occurs here   
```

I expected to see this happen:

Either no lint, or a fix that compiles correctly.

### Version

```text
rustc 1.59.0-nightly (7abab1efb 2021-12-17)
binary: rustc
commit-hash: 7abab1efb21617ba6845fa86328dffa16cfcf1dc
commit-date: 2021-12-17
host: x86_64-unknown-linux-gnu
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

@rustbot label +I-suggestion-causes-error

smoelius:
I'll get started on a fix right away.

teor2345:
Thanks!

I'm not sure if this will help, but here's the workaround we ended up merging:
```rust
    let original_inputs = transaction.inputs().to_vec();
[...]

PR #8201, opened 2021-12-31
Change `unnecessary_to_owned` `into_iter` suggestions to `MaybeIncorrect`

smoelius:
I am having a hard time finding a good solution for #8148, so I am wondering if is enough to just change the suggestion's applicability to `MaybeIncorrect`?

I apologize, as I realize this is a bit of a cop out.

changelog: none

### DIFF
--- a/clippy_lints/src/methods/unnecessary_iter_cloned.rs
+++ b/clippy_lints/src/methods/unnecessary_iter_cloned.rs
@@ -18,7 +18,7 @@ pub fn check(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, method_name: Symbol
         if let Some(callee_def_id) = fn_def_id(cx, parent);
         if is_into_iter(cx, callee_def_id);
         then {
-            check_for_loop_iter(cx, parent, method_name, receiver)
+            check_for_loop_iter(cx, parent, method_name, receiver, false)
         } else {
             false
         }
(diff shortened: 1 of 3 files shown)


clippy_lints/src/methods/unnecessary_iter_cloned.rs
```diff
@@ -34,6 +34,7 @@ pub fn check_for_loop_iter(
     expr: &'tcx Expr<'tcx>,
     method_name: Symbol,
     receiver: &'tcx Expr<'tcx>,
+    applicability: Applicability,
```
camsteffen, line 37: Passing applicability to this function is difficult to make sense of. Does this sound right? If it is cloned (or to_owned) *before* `into_iter` is called, that's where we have issues.

```suggestion
    cloned_before_iter: bool,
```

camsteffen:
> I apologize, as I realize this is a bit of a cop out.

No worries! This is one of those very difficult problems where we basically have to reverse-engineer the borrow checker.

clippy_lints/src/methods/unnecessary_iter_cloned.rs
smoelius: That's right: if there is a call to `clone`/`to_owned` before the call to `into_iter`, the suggestion to remove the `clone`/`to_owned` call could be incorrect, as could follow-on suggestions to remove uses of `.copied()`/`.cloned()` and `&`.

Should the effect of `cloned_before_iter` be the same: to determine the applicability of `check_for_loop_iter`'s suggestions?

camsteffen: Yes, purely non-functional change to make the code easier to follow.

smoelius: Following the change, I felt like the comments should be adjusted. I can change them back, if preferred.

camsteffen:
Do you mind adding a "Known Problems" section to the lint docs? Otherwise looks good.

smoelius:
I hope I used the right level of detail.

camsteffen:
Thanks! @bors r±

camsteffen:
That's weird
@bors r+

PR #8203, opened 2021-12-31
New lint: `iter_overeager_cloned`

### LINKED ISSUE #8202: New lint: `iter_overeager_cloned`
### What it does

Calling `.iter().cloned.last()` involved doing unnecessary work.
It's better to write it as `.iter().last().cloned()`

Same thing can rule can be applied to:
-  `.iter().cloned().next()` 
-  `.iter().cloned().count()` 
-  `.iter().cloned().flatten()`
-  `.iter().cloned().take(...)`
-  `.iter().cloned().skip(...)`
-  `.iter().cloned().nth(...)` 

### Lint Name

cloned_last

### Category

style, perf

### Advantage

- More efficient code

### Drawbacks

- I don't see any

### Example

```rust
    let _: Option<String> = vec!["1".to_string(), "2".to_string(), "3".to_string()]
        .iter().cloned().last()
```

Could be written as:

```rust
    let _: Option<String> = vec!["1".to_string(), "2".to_string(), "3".to_string()]
        .iter().last().cloned()
```

pmnoxx:
Closes #8202 

changelog: New lint: [`iter_overeager_cloned`]

### DIFF
--- a/clippy_lints/src/methods/iter_overeager_cloned.rs
+++ b/clippy_lints/src/methods/iter_overeager_cloned.rs
@@ -1,0 +1,62 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::source::snippet;
+use clippy_utils::ty::{get_iterator_item_ty, is_copy};
+use itertools::Itertools;
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::LateContext;
+use rustc_middle::ty;
+use std::ops::Not;
+
+use super::ITER_OVEREAGER_CLONED;
+use crate::redundant_clone::REDUNDANT_CLONE;
+
+/// lint overeager use of `cloned()` for `Iterator`s
+pub(super) fn check<'tcx>(
+    cx: &LateContext<'tcx>,
+    expr: &'tcx hir::Expr<'_>,
+    recv: &'tcx hir::Expr<'_>,
+    name: &str,
+    map_arg: &[hir::Expr<'_>],
+) {
+    // Check if it's iterator and get type associated with `Item`.
+    let inner_ty = match get_iterator_item_ty(cx, cx.typeck_results().expr_ty_adjusted(recv)) {
+        Some(ty) => ty,
+        _ => return,
+    };
+
+    match inner_ty.kind() {
+        ty::Ref(_, ty, _) if !is_copy(cx, ty) => {},
+        _ => return,
+    };
+
+    let (lint, preserve_cloned) = match name {
+        "count" => (REDUNDANT_CLONE, false),
+        _ => (ITER_OVEREAGER_CLONED, true),
+    };
+    let wildcard_params = map_arg.is_empty().not().then(|| "...").unwrap_or_default();
+    let msg = format!(
+        "called `cloned().{}({})` on an `Iterator`. It may be more efficient to call `{}({}){}` instead",
+        name,
+        wildcard_params,
+        name,
+        wildcard_params,
+        preserve_cloned.then(|| ".cloned()").unwrap_or_default(),
+    );
+
+    span_lint_and_sugg(
+        cx,
+        lint,
+        expr.span,
+        &msg,
+        "try this",
+        format!(
+            "{}.{}({}){}",
+            snippet(cx, recv.span, ".."),
+            name,
+            map_arg.iter().map(|a| snippet(cx, a.span, "..")).join(", "),
+            preserve_cloned.then(|| ".cloned()").unwrap_or_default(),
+        ),
+        Applicability::MachineApplicable,
+    );
+}
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -2319,7 +2319,7 @@ fn check_methods<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, msrv: Optio
             ("is_file", []) => filetype_is_file::check(cx, expr, recv),
             ("is_none", []) => check_is_some_is_none(cx, expr, recv, false),
             ("is_some", []) => check_is_some_is_none(cx, expr, recv, true),
-            ("last" | "skip", args @ []) => {
+            ("last", args @ []) | ("skip", args @ [_]) => {
                 if let Some((name2, [recv2, args2 @ ..], _span2)) = method_call(recv) {
                     if let ("cloned", []) = (name2, args2) {
                         iter_overeager_cloned::check(cx, expr, recv2, name, args);
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -30,6 +30,7 @@ mod iter_count;
 mod iter_next_slice;
 mod iter_nth;
 mod iter_nth_zero;
+mod iter_overeager_cloned;
 mod iter_skip_next;
 mod iterator_step_by_zero;
 mod manual_saturating_arithmetic;
@@ -104,6 +105,41 @@ declare_clippy_lint! {
     pub CLONED_INSTEAD_OF_COPIED,
     pedantic,
     "used `cloned` where `copied` could be used instead"
+}
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `_.cloned().<func>()` where call to `.cloned()` can be postponed.
+    ///
+    /// ### Why is this bad?
+    /// It's often inefficient to clone all elements of an iterator, when eventually, only some
+    /// of them will be consumed.
+    ///
+    /// ### Examples
+    /// ```rust
+    /// # let vec = vec!["string".to_string()];
+    ///
+    /// // Bad
+    /// vec.iter().cloned().take(10);
+    ///
+    /// // Good
+    /// vec.iter().take(10).cloned();
+    ///
+    /// // Bad
+    /// vec.iter().cloned().last();
+    ///
+    /// // Good
+    /// vec.iter().last().cloned();
+    ///
+    /// ```
+    /// ### Known Problems
+    /// This `lint` removes the side of effect of cloning items in the iterator.
+    /// A code that relies on that side-effect could fail.
+    ///
+    #[clippy::version = "1.59.0"]
+    pub ITER_OVEREAGER_CLONED,
+    perf,
+    "using `cloned()` early with `Iterator::iter()` can lead to some performance inefficiencies"
 }
 
 declare_clippy_lint! {
@@ -1950,6 +1986,7 @@ impl_lint_pass!(Methods => [
     CLONE_ON_COPY,
     CLONE_ON_REF_PTR,
     CLONE_DOUBLE_REF,
+    ITER_OVEREAGER_CLONED,
     CLONED_INSTEAD_OF_COPIED,
     FLAT_MAP_OPTION,
     INEFFICIENT_TO_STRING,
@@ -2243,9 +2280,10 @@ fn check_methods<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, msrv: Optio
                 },
                 _ => {},
             },
-            ("count", []) => match method_call(recv) {
-                Some((name @ ("into_iter" | "iter" | "iter_mut"), [recv2], _)) => {
-                    iter_count::check(cx, expr, recv2, name);
+            (name @ "count", args @ []) => match method_call(recv) {
+                Some(("cloned", [recv2], _)) => iter_overeager_cloned::check(cx, expr, recv2, name, args),
+                Some((name2 @ ("into_iter" | "iter" | "iter_mut"), [recv2], _)) => {
+                    iter_count::check(cx, expr, recv2, name2);
                 },
                 Some(("map", [_, arg], _)) => suspicious_map::check(cx, expr, recv, arg),
                 _ => {},
@@ -2266,10 +2304,10 @@ fn check_methods<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, msrv: Optio
                 flat_map_identity::check(cx, expr, arg, span);
                 flat_map_option::check(cx, expr, arg, span);
             },
-            ("flatten", []) => {
-                if let Some(("map", [recv, map_arg], _)) = method_call(recv) {
-                    map_flatten::check(cx, expr, recv, map_arg);
-                }
+            (name @ "flatten", args @ []) => match method_call(recv) {
+                Some(("map", [recv, map_arg], _)) => map_flatten::check(cx, expr, recv, map_arg),
+                Some(("cloned", [recv2], _)) => iter_overeager_cloned::check(cx, expr, recv2, name, args),
+                _ => {},
             },
             ("fold", [init, acc]) => unnecessary_fold::check(cx, expr, init, acc, span),
             ("for_each", [_]) => {
@@ -2281,6 +2319,13 @@ fn check_methods<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, msrv: Optio
             ("is_file", []) => filetype_is_file::check(cx, expr, recv),
             ("is_none", []) => check_is_some_is_none(cx, expr, recv, false),
             ("is_some", []) => check_is_some_is_none(cx, expr, recv, true),
+            ("last", args @ []) => {
+                if let Some((name2, [recv2, args2 @ ..], _span2)) = method_call(recv) {
+                    if let ("cloned", []) = (name2, args2) {
+                        iter_overeager_cloned::check(cx, expr, recv2, name, args);
+                    }
+                }
+            },
             ("map", [m_arg]) => {
                 if let Some((name, [recv2, args @ ..], span2)) = method_call(recv) {
                     match (name, args) {
@@ -2296,20 +2341,22 @@ fn check_methods<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, msrv: Optio
                 map_identity::check(cx, expr, recv, m_arg, span);
             },
             ("map_or", [def, map]) => option_map_or_none::check(cx, expr, recv, def, map),
-            ("next", []) => {
-                if let Some((name, [recv, args @ ..], _)) = method_call(recv) {
-                    match (name, args) {
-                        ("filter", [arg]) => filter_next::check(cx, expr, recv, arg),
-                        ("filter_map", [arg]) => filter_map_next::check(cx, expr, recv, arg, msrv),
-                        ("iter", []) => iter_next_slice::check(cx, expr, recv),
-                        ("skip", [arg]) => iter_skip_next::check(cx, expr, recv, arg),
+            (name @ "next", args @ []) => {
+                if let Some((name2, [recv2, args2 @ ..], _)) = method_call(recv) {
+                    match (name2, args2) {
+                        ("cloned", []) => iter_overeager_cloned::check(cx, expr, recv2, name, args),
+                        ("filter", [arg]) => filter_next::check(cx, expr, recv2, arg),
+                        ("filter_map", [arg]) => filter_map_next::check(cx, expr, recv2, arg, msrv),
+                        ("iter", []) => iter_next_slice::check(cx, expr, recv2),
+                        ("skip", [arg]) => iter_skip_next::check(cx, expr, recv2, arg),
                         ("skip_while", [_]) => skip_while_next::check(cx, expr),
                         _ => {},
                     }
                 }
             },
-            ("nth", [n_arg]) => match method_call(recv) {
+            ("nth", args @ [n_arg]) => match method_call(recv) {
                 Some(("bytes", [recv2], _)) => bytes_nth::check(cx, expr, recv2, n_arg),
+                Some(("cloned", [recv2], _)) => iter_overeager_cloned::check(cx, expr, recv2, name, args),
                 Some(("iter", [recv2], _)) => iter_nth::check(cx, expr, recv2, recv, n_arg, false),
                 Some(("iter_mut", [recv2], _)) => iter_nth::check(cx, expr, recv2, recv, n_arg, true),
                 _ => iter_nth_zero::check(cx, expr, recv, n_arg),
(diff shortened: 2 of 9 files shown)


clippy_lints/src/methods/cloned_last.rs
```diff
@@ -0,0 +1,36 @@
...
+                   `.last.cloned()` instead";
+        let iter_snippet = snippet(cx, recv.span, "..");
+        // add note if not multi-line
+        span_lint_and_sugg(
+            cx,
+            CLONED_LAST,
+            expr.span,
+            msg,
+            "try this",
+            format!("{}.last().cloned()", iter_snippet),
```
llogiq, line 32: If the `.last()` is `unwrapped` or `?`d, we could even suggest to `.clone()` that. Just a suggestion for extending the lint, this won't stop us from merging.

llogiq (approved):
This looks good to me. I have an idea how we could make that lint even better. If you want to do that within this PR, go ahead, otherwise let me know and I'll r+.

clippy_lints/src/methods/cloned_last.rs
```diff
@@ -0,0 +1,32 @@
...
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::CLONED_LAST;
+
+/// lint use of `cloned().next()` for `Iterators`
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>, recv: &'tcx hir::Expr<'_>) {
+    // lint if caller of `.filter().next()` is an Iterator
```
llogiq, line 13: This comment looks out of place here.

clippy_lints/src/methods/cloned_last.rs
pmnoxx: @llogiq Actually, you are right. The result of `last()` is an `Option`, so we should be using `close()` instead.

I updated the PR.

clippy_lints/src/methods/cloned_last.rs
pmnoxx: @llogiq Thanks, I fixed the comment.

clippy_lints/src/methods/cloned_last.rs
pmnoxx: @llogiq That's a great suggestion. 

I see:  for `Option<>` /  `Result<>`.
`recv.cloned().unwrap()`
could be
`recv.unwrap().clone`.

I see that `cloned` should be moved furthermost the chain as long as possible, not just for `last`.
For example here:
```rust
 vec.iter().cloned().filter(...).skip(5).take(10).next().last().unwrap()
```
That could be just written as
```rust
let val = vec.cloned().filter(...).skip(5).take(10).next().last().unwrap().clone()
```

This may even allow us to drop the `clone()` entirely, in some cases, when clone() is not needed.

It may be a good idea to propagate clone from before, to after
- `last`
- `take`
- `skip`
- `next`.
- `unwrap` 
- `filter` // has lambda, so requires some more checks
- `find`.// has lambda,so requires some more checks
- ... probably some others as well

The question is, should each be a separate lint, or should they be combined together?

clippy_lints/src/methods/cloned_last.rs
```diff
@@ -0,0 +1,31 @@
...
+use clippy_utils::source::snippet;
+use clippy_utils::ty::implements_trait;
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::CLONED_LAST;
+
+/// lint use of `cloned().next()` for `Iterators`
```
llogiq, line 11: Shouldn't that be ` cloned().last()`?

On the other hand, you could extend the lint to `needless_cloned_iter` and lint `.last()`, `.next()` and `.count()` (the latter simply removing `cloned()`...

llogiq:
I'd be OK with merging this with the `.cloned().next()` replaced by `.cloned().last()` in the two comments, or if you want you can tackle making the lint more powerful. Ping me if you need any help!

tests/ui/cloned_last.rs
```diff
@@ -7,4 +7,16 @@ fn main() {
         .iter()
         .cloned()
         .last();
+
+    #[rustfmt::skip]
```
llogiq, line 11: Why skip rustfmt?

pmnoxx: I copied it from another example. It's not really needed, so I removed it

pmnoxx:
> I'd be OK with merging this with the `.cloned().next()` replaced by `.cloned().last()` in the two comments, or if you want you can tackle making the lint more powerful. Ping me if you need any help!

I added implementation for
- skip
- take
- count
- last
- next
- nth


TODO
- update documentation

Questions:
- what name should be used for new lint? `needless_cloned` ?
- should it be applied to `clippy::all` or `clippy::nursery`
- maybe this should only be applies to types ,which are bigger than 8 bytes, as it's more efficient to deference types 8 bytes and shorter

I think that support for `.cloned().next().unwrap()` should be in a separate PR. 

 Fixing `option_receiver.cloned().unwrap()` or `result_receiver.cloned().unwrap()`, shouldn't be hard to to either, I'll take a look at it later.

clippy_lints/src/methods/cloned_last.rs
llogiq: I think your example is a bit off, but otherwise full ack.

I'd be OK with multiple lints if there are useful distinctions to make. E.g. if one part of the check can have false positives, it would be a good idea to put that into a different `nursery` lint. Or if one part could reasonably be `style` and another `perf`. Otherwise, putting all the checks into one lint seems natural. One thing I that comes to mind is that we already have the `needless_clone` lint, so we could add cases where the `cloned` can be removed outright to that, making the rest `early_iter_cloned` or some such.

Again, I'm also OK if you restrict the scope of this PR and do a followup later. Have fun!

llogiq:
My name suggestion as per comment above is: Add the cases where the `cloned` call can be removed outright to `needless_clone`, the others could be `needless_iter_cloned` or `early_iter_cloned` or `overeager_iter_cloned`. For `last` and `skip`, the latter seems to be most appropriate.

llogiq:
Oh, and I would like to run some tests before deciding whether to put it in `style`, `perf` or `nursery`. If we find false positives or lots of noise, I'd suggest the latter. Otherwise I'd be OK with either of the former.

tests/ui/cloned_last.rs
```diff
@@ -0,0 +1,36 @@
+// run-rustfix
+#![warn(clippy::all)]
+
+fn main() {
+    let _: Option<String> = vec!["1".to_string(), "2".to_string(), "3".to_string()]
```
llogiq, line 5: How about `["1", "2", "3"].into_iter().cloned().collect::<Vec<_>>()` to avoid the `to_string()`s?

Also perhaps put it in a binding and clone as needed.

tests/ui/cloned_last.rs
```diff
@@ -0,0 +1,36 @@
...
+        .iter()
+        .cloned()
+        .skip(2)
+        .collect();
+
+    let _ = vec!["1".to_string(), "2".to_string(), "3".to_string()]
+        .iter()
+        .cloned()
+        .nth(2);
+}
```
llogiq, line 36: I'd like to see some tests against false positives if possible: Methods after `.cloned()` which are OK, and if you do anything that would require more complex tests (e.g. `.map(_)`), there should be tests for those checks, too.

clippy_lints/src/methods/mod.rs
```diff
@@ -107,6 +108,29 @@ declare_clippy_lint! {
     "used `cloned` where `copied` could be used instead"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `_.cloned().last()`.
```
llogiq, line 113: The documentation should also reflect the complete lint.

clippy_lints/src/methods/mod.rs
```diff
@@ -107,6 +108,29 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Why is this bad?
```
llogiq, line 115: Not only performance, but also style: For `.cloned().next()`, the call will very likely be inlined anyway, but as code changes, keeping a consistent style is also a good thing.

I'd also add a `### Known Problems` section, because the lint is partially behavior-changing, because `.cloned().last()` will clone all elements unlike `.last().cloned()`. This cloning could be observed as a side effect, which would be removed by this lint. Given the contract of `Clone` and the likely perf benefit, I find this acceptable, but we should still document it.

tests/ui/cloned_last.rs
pmnoxx:
```rust
143 |                     let x = ["1", "2", "3"].into_iter().cloned().collect::<Vec<_>>();
    |                                                         ^^^^^^ the trait `Clone` is not implemented for `str`
```

pmnoxx: I could write it like this:
```
let _ = ["1", "2", "3"].into_iter().map(ToString::to_string).collect::<Vec<_>>()
```

clippy_lints/src/methods/mod.rs
```diff
@@ -107,6 +108,29 @@ declare_clippy_lint! {
...
+    /// vec.iter().cloned().last();
+    /// ```
+    /// Could be written as
+    /// ```rust
+    /// # let vec = vec!["string".to_string()];
+    /// vec.iter().last().cloned();
+    /// ```
+    #[clippy::version = "1.59.0"]
+    pub CLONED_LAST,
+    complexity,
```
llogiq, line 130: I think this should be `perf` or `nursery`, depending on the number of additional warnings in lintcheck vs. master. If it's too noisy, I'd prefer the latter, otherwise I'd look into the added warnings and see if the benefit outweighs the added effort to fix the lint.

tests/ui/cloned_last.rs
pmnoxx: Although, that would deny the purpose of having this test.

pmnoxx:
> My name suggestion as per comment above is: Add the cases where the `cloned` call can be removed outright to `needless_clone`, the others could be `needless_iter_cloned` or `early_iter_cloned` or `overeager_iter_cloned`. For `last` and `skip`, the latter seems to be most appropriate.

I agree, as a use of clippy I think like to see it divided maybe into
- `needless_cloned` - `count`, `len`, `is_empty`   - `cloned` is remoed
- `overeager_iter_cloned` - `skip`, `take`, `last`, `nth` - affects performance; `cloned` is used less often` - has side effect
-  `early_iter_cloned` - `next` - clone can be done later, perhaps this will allow further optimizations later on; - moves clone further into the chain, may allow further performance optimizations. 

- `filter` - can be added later to  `overeager_iter_cloned`
- `map`. `flat_map` - not sure, sometimes may be `map` / `flat_map` doesn't need cloned version of the item.

clippy_lints/src/methods/mod.rs
pmnoxx: @llogiq It looks to me that maybe the lint should be split. Sometimes it may not make a sense to do it, if it's just `.cloned().next()`, though it depend on whats after.

Other times it may be worth it, because you may avoid cloning the whole collection.

pmnoxx: I was planning to write a more complicated check, which can cover more cases, but I think this PR is a good start. There are a few things that I don't know how to use yet.

llogiq: I don't think that would warrant a split: Just because the code has `.next()` now doesn't mean that it won't get changed to `.nth(_)` later. Also even when only changing `.cloned().next()`, to `.next().cloned()`, this will reduce compile time if ever so slightly, so that's a perf benefit anyway.

llogiq:
I'm ok with the `needless_clone` / `overeager_iter_cloned` split.

I would avoid adding `needless_cloned`, because a) that's *very* similar to `needless_clone` which we already have and b) also conceptually similar enough we may as well just extend that lint.

I don't think the `early_iter_cloned` warrants it's own lint. As I explained above, code changes, and doing it now has no drawbacks and benefits compile time even if only slightly.

Note that we'll have to extend the `needless_clone` documentation to include the cases if we go that route.

tests/ui/cloned_last.rs
pmnoxx: @llogiq I added some examples, where doing automatic replacement would be bad, because the signature of methods would change.

I added a list of TODOs as well.

clippy_lints/src/methods/mod.rs
pmnoxx: @llogiq I updated the documentation.

clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -0,0 +1,78 @@
...
+        implements_trait(cx, cx.typeck_results().expr_ty(recv), id, &[])
+    });
+    if recv_impls_iterator {
+        let (lint, msg) = match name {
+            "count" => (
+                REDUNDANT_CLONE,
+                format!(
+                    "called `cloned().{}()` on an `Iterator`. It may be more efficient to call\
+                    `.{}()` instead",
+                    name, name
```
pmnoxx, line 31: @llogiq I'm not sure if this is the preferred way to return `REDUNDANT_CLONE` lint.

If not I'll change the code of `redundant_clone.rs`.

llogiq: Why would it not be? Issuing the lint here keeps the code simple. Perhaps add a comment to the `redundant_clone` code that links to the implementation.

pmnoxx: > Why would it not be? Issuing the lint here keeps the code simple. Perhaps add a comment to the `redundant_clone` code that links to the implementation.

@llogiq This depends on `clippy` design; Whenever or or all `check` always run or not. I suspect that's the case, but I didn't read that part of the code to make sure.

pmnoxx: And also on how checks are used. I don't know whenever, there is a rule that says "all lints should be generated from file with the same name". It was the case for some lints, but I didn't check if that's true. Though, I see you don't see a problem, so I guess it's fine.

pmnoxx:
@llogiq I think some more changes are needed.

--) Cloning small values, references of references is actually beneficial. 

Consider following 
```rust
/// cloned() should not be removed
let _ = [0u8, 0u8].iter().cloned()....
/// cloned() should not be removed
let _ = Vec::<&&String>.iter().cloned()....
```
Cloning `&u8` is a good idea, this will improve performance, as there is no need to do dereferencing.
Same thing can be said about `&&String`. 

I think we should apply the same rules as `clippy_lints/src/pass_by_ref_or_value.rs` uses.

If the type after cloning is `<=8 bytes`, then we should keep the `cloned`, otherwise postpone it.
For example, changing `&&String` to `&String`, keep the size the same, but removes one extra deference needed.

llogiq:
That should be `.copied()`, not `.cloned()`, but we already have that lint. Regardless, not linting if the element type is of `size <= size_of::<usize>()` makes sense.

pmnoxx:
> That should be `.copied()`, not `.cloned()`, but we already have that lint. Regardless, not linting if the element type is of `size <= size_of::<usize>()` makes sense.

I added a test, and I see that this would be applied automatically:
```
diff --git i/tests/ui/iter_overeager_cloned.fixed w/tests/ui/iter_overeager_cloned.fixed
index 01b64ff7a..7295ad871 100644
--- i/tests/ui/iter_overeager_cloned.fixed
+++ w/tests/ui/iter_overeager_cloned.fixed
@@ -39,4 +39,7 @@ fn main() {
 
     // Not implemented yet
     let _ = vec.iter().cloned().any(|x| x.len() == 1);
+
+    // Should probably stay as it is.
+    let _ = [0, 1, 2, 3, 4].iter().take(10).cloned();
 }
diff --git i/tests/ui/iter_overeager_cloned.rs w/tests/ui/iter_overeager_cloned.rs
index fede4c733..dd04e33a4 100644
--- i/tests/ui/iter_overeager_cloned.rs
+++ w/tests/ui/iter_overeager_cloned.rs
@@ -41,4 +41,7 @@ fn main() {
 
     // Not implemented yet
     let _ = vec.iter().cloned().any(|x| x.len() == 1);
+
+    // Should probably stay as it is.
+    let _ = [0, 1, 2, 3, 4].iter().cloned().take(10);
 }
diff --git i/tests/ui/iter_overeager_cloned.stderr w/tests/ui/iter_overeager_cloned.stderr
index 4a58687f4..343614e75 100644
--- i/tests/ui/iter_overeager_cloned.stderr
+++ w/tests/ui/iter_overeager_cloned.stderr
@@ -54,5 +54,11 @@ LL ~     let _ = [Some(Some("str".to_string())), Some(Some("str".to_string()))]
 LL ~         .iter().flatten().cloned();
    |
 
-error: aborting due to 7 previous errors
+error: called `cloned().take(...)` on an `Iterator`. It may be more efficient to call`.take(...).cloned()` instead
+  --> $DIR/iter_overeager_cloned.rs:46:13
+   |
+LL |     let _ = [0, 1, 2, 3, 4].iter().cloned().take(10);
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `[0, 1, 2, 3, 4].iter().take(10).cloned()`
+
+error: aborting due to 8 previous errors
```

clippy_lints/src/methods/mod.rs
```diff
@@ -107,6 +108,42 @@ declare_clippy_lint! {
     "used `cloned` where `copied` could be used instead"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `_.cloned().<func>()` where call to `.cloned()` can be postponed,
+    /// or often removed entirely.
```
llogiq, line 114: Don't we issue `REDUNDANT_CLONE` when the `.cloned()` can be removed?

llogiq:
Cool stuff! My remaining nits are with the documentation, otherwise this looks great.

clippy_lints/src/methods/mod.rs
pmnoxx: @llogiq Ok, fixed.

clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -0,0 +1,78 @@
...
+
+/// lint use of `cloned().last()` for `Iterators`
+pub(super) fn check<'tcx>(
+    cx: &LateContext<'tcx>,
+    expr: &'tcx hir::Expr<'_>,
+    recv: &'tcx hir::Expr<'_>,
+    name: &str,
+    map_arg: &[hir::Expr<'_>],
+) {
+    // lint if caller of `.cloned().last()` is an Iterator
```
llogiq, line 20:
```suggestion
    // lint if callee is an Iterator
```

clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -0,0 +1,78 @@
...
+use clippy_utils::ty::implements_trait;
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::ITER_OVEREAGER_CLONED;
+use crate::redundant_clone::REDUNDANT_CLONE;
+
+/// lint use of `cloned().last()` for `Iterators`
```
llogiq, line 12:
```suggestion
/// lint overeager use of `cloned()` for `Iterator`s
```

llogiq (commented):
A few more nits to fix. Do you need help with the size check?

pmnoxx:
> A few more nits to fix. Do you need help with the size check?

I'm still thinking whenever the check is needed. For example, for

- `count` - 

> pass_by_ref_or_value

I'm not entirely convinced whenever the check is needed. For example.

```rust
[0u8, 1u8, 2u8, 3u8].iter().cloned().count() /// cloned() should be removed
 [0, 1, 2, 3, 4].last().cloned().last();          /// it's better to not deference values, and move cloned() last
 [0, 1, 2, 3, 4].last().cloned().nth(40);          /// same here
 [0, 1, 2, 3, 4].last().cloned().skip(40);          /// same here
 [0, 1, 2, 3, 4].last().cloned().next();     // probably it's better not to print - not sure
 [0, 1, 2, 3, 4].last().cloned().take(50);     // probably it's better not to print - not sure
 [0, 1, 2, 3, 4].last().cloned().flatten();     // probably it's better not to print - not sure
```

@llogiq I'm not sure how to get  the type from the iterator. I'll have to spend some time and figure out a better way to navigate through code. Type hinting doesn't seem to work correctly with Rust plugin for Intellij, which makes things harder.

My understanding is that the receiver is some generic iterator, that implements `Iterator<Item = SomeItem>`. I'm not sure yet how to do get it.

llogiq:
I just had a discussion with someone from libs. There's a Rust PR [#90209](https://github.com/rust-lang/rust/pull/90209) that aims to remove more clones through specialization. Apparently the libs team thinks that as there never was any guarantee of current behavior, they're free to change it. Which is fine by me. That somewhat reduces the need for this lint, but doesn't render it moot, because as we all know specialization is a fickle beast and may sometimes fail to apply. So we should still add this lint, but perhaps note the possibility of not seeing a perf win in the docs.

llogiq:
That PR was closed, so we're left with the clippy lint and some docs on `Iterator::cloned` I just wrote. Hopefully this will suffice to have people avoid most needless clones on iterating.

pmnoxx:
> That PR was closed, so we're left with the clippy lint and some docs on `Iterator::cloned` I just wrote. Hopefully this will suffice to have people avoid most needless clones on iterating.

Ok, just let me know what you think is best. Whenever we should ships this PR as it is. Or improve it, or close this PR.

llogiq:
Re-reading this, I think we can look into refining the lint later. For now let's just merge it.

@bors r+

llogiq:
I overlooked that you need to rebase the PR on current master. Shouldn't be too hard.

pmnoxx:
> I overlooked that you need to rebase the PR on current master. Shouldn't be too hard.

I rebased the PR.

PR #8204, opened 2021-12-31
[`erasing_op`] lint ignored when operation `Output` type is different from the type of constant `0`

### LINKED ISSUE #7210: erasing_op and identity_op should only be triggered for internal
Lint name: 
    - erasing_op
    - identity_op

I tried this code:

```rust
struct L(i32);
struct Meter;
impl core::ops::Mul<Meter> for  i32 {
    type Output = L;
    fn mul(self,_:Meter) -> L {
        L(self)
    }
}
fn check() {
  let _ = 0 * Meter;  //=> Fire clippy::erasing_op
  let _ = 1 * Meter;  //=> Fire clippy::identity_op
}
```
Multiply by 0 is not an erasing operation as the type is changed.
Multiply by 1 is not an identity operation as the type is changed.

I suppose that the lint `erasing_op` may only fire for where `0*x` is `0`:
   ```
   let a = AType::new(...);
 
    let res: i32 = 0 * a; //the result is of the type of the 0
   assert!(res;0);  //and the result is 0 <=> this is an erasing_op
  ```
And for `erasing_op` only where `1*x` is `x` (`1*x` and `x` should have the same state): e.g:
 ```
   let a = AType::new(...);
   let res: AType = 1* a; //the result is of the type of a
   assert_same_binary_representation!(res;a);  //and the result is a
   ```

  

### Meta

- `cargo clippy -V`:  clippy 0.1.54 (881c1ac40 2021-05-08)
- `rustc -Vv`:
  ```
  rustc 1.54.0-nightly (881c1ac40 2021-05-08)
  binary: rustc
  commit-hash: 881c1ac408d93bb7adaa3a51dabab9266e82eee8
  commit-date: 2021-05-08
  release: 1.54.0-nightly
  LLVM version: 12.0.0
  ```

wigy-opensource-developer:
fixes #7210

changelog: [`erasing_op`] lint ignored when operation `Output` type is different from the type of constant `0`

### DIFF
--- a/clippy_lints/src/erasing_op.rs
+++ b/clippy_lints/src/erasing_op.rs
@@ -1,9 +1,11 @@
 use clippy_utils::consts::{constant_simple, Constant};
 use clippy_utils::diagnostics::span_lint;
+use clippy_utils::ty::same_type_and_consts;
+
 use rustc_hir::{BinOpKind, Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::TypeckResults;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::source_map::Span;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -35,24 +37,34 @@ impl<'tcx> LateLintPass<'tcx> for ErasingOp {
             return;
         }
         if let ExprKind::Binary(ref cmp, left, right) = e.kind {
+            let tck = cx.typeck_results();
             match cmp.node {
                 BinOpKind::Mul | BinOpKind::BitAnd => {
-                    check(cx, left, e.span);
-                    check(cx, right, e.span);
+                    check(cx, tck, left, right, e);
+                    check(cx, tck, right, left, e);
                 },
-                BinOpKind::Div => check(cx, left, e.span),
+                BinOpKind::Div => check(cx, tck, left, right, e),
                 _ => (),
             }
         }
     }
 }
 
-fn check(cx: &LateContext<'_>, e: &Expr<'_>, span: Span) {
-    if constant_simple(cx, cx.typeck_results(), e) == Some(Constant::Int(0)) {
+fn different_types(tck: &TypeckResults<'tcx>, input: &'tcx Expr<'_>, output: &'tcx Expr<'_>) -> bool {
+    let input_ty = tck.expr_ty(input).peel_refs();
+    let output_ty = tck.expr_ty(output).peel_refs();
+    !same_type_and_consts(input_ty, output_ty)
+}
+
+fn check(cx: &LateContext<'cx>, tck: &TypeckResults<'cx>, op: &Expr<'cx>, other: &Expr<'cx>, parent: &Expr<'cx>) {
+    if constant_simple(cx, tck, op) == Some(Constant::Int(0)) {
+        if different_types(tck, other, parent) {
+            return;
+        }
         span_lint(
             cx,
             ERASING_OP,
-            span,
+            parent.span,
             "this operation will always return zero. This is likely not the intended outcome",
         );
     }
--- a/tests/ui/erasing_op.rs
+++ b/tests/ui/erasing_op.rs
@@ -1,3 +1,34 @@
+struct Length(u8);
+struct Meter;
+
+impl core::ops::Mul<Meter> for u8 {
+    type Output = Length;
+    fn mul(self, _: Meter) -> Length {
+        Length(self)
+    }
+}
+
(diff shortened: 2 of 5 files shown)


clippy_lints/src/erasing_op.rs
```diff
@@ -37,22 +39,32 @@ impl<'tcx> LateLintPass<'tcx> for ErasingOp {
...
+                BinOpKind::Div => check(cx, left, e),
                 _ => (),
             }
         }
     }
 }
 
-fn check(cx: &LateContext<'_>, e: &Expr<'_>, span: Span) {
-    if constant_simple(cx, cx.typeck_results(), e) == Some(Constant::Int(0)) {
+fn different_types(typeck_results: &TypeckResults<'tcx>, left: &'tcx Expr<'_>, right: &'tcx Expr<'_>) -> bool {
```
xFrednet, line 52: NIT: Here I would try to give the two arguments a better name to indicate what they are. At first when reading this I suspected that they are the left and right expression of the multiplication, while they are actually the input and output. Could you maybe update this: 
```suggestion
fn different_types(typeck_results: &TypeckResults<'tcx>, input: &'tcx Expr<'_>, output: &'tcx Expr<'_>) -> bool {
```

tests/ui/erasing_op.rs
```diff
@@ -1,3 +1,13 @@
+struct Length(u8);
+struct Meter;
+
+impl core::ops::Mul<Meter> for u8 {
+    type Output = Length;
+    fn mul(self, _: Meter) -> Length {
+        Length(self)
+    }
+}
```
xFrednet, line 9: The example in the issue was used to create a new type, like the one you added here. Could you add a second test case there the argument stays the same, to ensure that operations returning the same type is still being linted? You can use this:

```rs
#![allow(unused)]

struct Vec2 {
    x: u32,
    y: u32,
}

impl core::ops::Mul<Vec2> for u8 {
    type Output = Vec2;
    fn mul(self, _: Vec2) -> Vec2 {
        Vec2 {
            x: 0,
            y: 0,
        }
    }
}
```

xFrednet (commented):
Welcome to Clippy :wave:. Your changes look good to me, I only have two smaller comments.

I've also looked at the implementation of `identity_op` which restricts the lint emissions to only trigger on number types. I think your solution of comparing the types is the better solution. We might want to think about also making this change to the `identity_op` implementation :upside_down_face:

tests/ui/erasing_op.rs
wigy-opensource-developer: Thanks for the prompt feedback even in these holiday seasons. I have added something similar, but the current logic allows to scale such a vector by a constant 0. If you have a good idea how to make the 2 extra test cases I have added fail the lint without failing `0 * Meter`, feel free to advise or just commit on my branch.

There are 3 types involved in a binary op, and it is easy to create false positives or ignore something relevant. If one of the arguments is a constant integer 0, I would test the other argument's type against the output type, and ignore the lint if those are different. But I am unsure whether that would be a sufficient implementation.

xFrednet: You're welcome! I'm currently using Clippy to take a break from some other work that I should do, it's a little like a holiday. ^^

Regarding your suggestion, that would actually be the route I'd take. You're right that it might have some disadvantages, even if I can't really think of any ATM. The current suggestion isn't concerned with the return type and lints in either case, looking at the open issues it seems like the issue you're fixing was the only relevant report. Therefore, I think it's safe to say that doing this comparison would be a good fix as it increases the precision of the lint. So, if you have the time, it would be perfect if you could implement your suggestion :upside_down_face:

wigy-opensource-developer:
Implemented based on the suggestion, squashed and rebased on master.

xFrednet (approved):
LGTM

xFrednet:
Looks good to me, thank you for the swift response, could you bless the tests one more time and squash it, then it's ready to be merged! :upside_down_face:

wigy-opensource-developer:
Aye, sir, just noticed the breakage. Fixed.

xFrednet:
Alright, looks good to me, I hope you also had fun while working on this :upside_down_face: 

@bors r+

xFrednet:
The changelog label and entry have to be on the same line, it should work now :upside_down_face: 

@bors retry

wigy-opensource-developer:
Oh. Thanks. Sure, I enjoyed it. First time I looked below the hood and it was less scary than I though. I only made a Roslyn check for C# before, that was worse 🤣

xFrednet:
That's good to hear, the manageable code base was actually the reason why I started to contribute to Clippy. Rustc and other compiler projects seemed to intimidating xD

PR #8208, opened 2022-01-01
wrong_self_convention: Match `SelfKind::No` more restrictively

### LINKED ISSUE #8142: wrong_self_convention: FN with `is_*` taking `&mut self`
### Summary

My interpretation of the table at https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention is that a method called `is_*` is allowed to take no self, or `&self`, but not `self` or `&mut self`. In reality, only `self` produces a warning. Is this intentional?

### Lint Name

wrong_self_convention

### Reproducer

I tried this code:

```rust
#![allow(unused)]
struct S { field: String }

impl S {
    fn is_foo(&mut self) -> bool { todo!() }
}

fn main() {}
```

```
warning: methods called `is_*` usually take `self` by reference or no `self`
  --> src/main.rs:16:15
   |
16 |     fn is_foo(&mut self) -> bool {
   |               ^^^^
```

Instead, this happened:

No warning

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

nmathewson:
The `wrong_self_convention` lint uses a `SelfKind` type to decide
whether a method has the right kind of "self" for its name, or whether
the kind of "self" it has makes its name confusable for a method in
a common trait.  One possibility is `SelfKind::No`, which is supposed
to mean "No `self`".

Previously, SelfKind::No matched everything _except_ Self, including
references to Self.  This patch changes it to match Self, &Self, &mut
Self, Box<Self>, and so on.

For example, this kind of method was allowed before:

```
impl S {
    // Should trigger the lint, because
    // "methods called `is_*` usually take `self` by reference or no `self`"
    fn is_foo(&mut self) -> bool { todo!() }
}
```

But since SelfKind::No matched "&mut self", no lint was triggered
(see #8142).

With this patch, the code above now gives a lint as expected.

fixes #8142

changelog: [`wrong_self_convention`] rejects `self` references in more cases

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -2535,11 +2535,17 @@ impl SelfKind {
             implements_trait(cx, ty, trait_def_id, &[parent_ty.into()])
         }
 
+        fn matches_none<'a>(cx: &LateContext<'a>, parent_ty: Ty<'a>, ty: Ty<'a>) -> bool {
+            !matches_value(cx, parent_ty, ty)
+                && !matches_ref(cx, hir::Mutability::Not, parent_ty, ty)
+                && !matches_ref(cx, hir::Mutability::Mut, parent_ty, ty)
+        }
+
         match self {
             Self::Value => matches_value(cx, parent_ty, ty),
             Self::Ref => matches_ref(cx, hir::Mutability::Not, parent_ty, ty) || ty == parent_ty && is_copy(cx, ty),
             Self::RefMut => matches_ref(cx, hir::Mutability::Mut, parent_ty, ty),
-            Self::No => ty != parent_ty,
+            Self::No => matches_none(cx, parent_ty, ty),
         }
     }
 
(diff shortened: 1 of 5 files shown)


nmathewson:
(FWIW, `lintcheck` reports that there are are a couple of new positives here; I haven't evaluated whether they are desirable or not.)

xFrednet (approved):
Looks good to me, thank you very much! I've also looked at the output of lintcheck, we have five new reports which all seem to be correct. This is a very nice fix :+1: 


file | lint | message
-- | -- | --
cargo-0.49.0/src/cargo/core/source/mod.rs:103:18 | clippy::wrong_self_convention | "methods called is_* usually take self by reference or no self"
cargo-0.49.0/src/cargo/ops/tree/graph.rs:166:27 | clippy::wrong_self_convention | "methods called from_* usually take no self"
cargo-0.49.0/src/cargo/sources/registry/index.rs:441:22 | clippy::wrong_self_convention | "methods called is_* usually take self by reference or no self"
xsv-0.13.0/src/config.rs:282:33 | clippy::wrong_self_convention | "methods called from_* usually take no self"
xsv-0.13.0/src/config.rs:300:38 | clippy::wrong_self_convention | "methods called from_* usually take no self"

xFrednet:
btw a second side note, you can assign yourself to an issue including `@rustbot claim` in your comment. That indicates to others that you want to work on that issue :upside_down_face: 

Thank you very much for this PR :+1: 

@bors r+

PR #8217, opened 2022-01-03
Fix `needless_borrow` causing mutable borrows to be moved

### LINKED ISSUE #8191: Bad suggestion for `&mut *(&mut &mut T)` from `clippy::needless_borrow`
### Summary

`clippy::needless_borrow` suggested a change that only works if the compiler was going to reborrow the reference or the reference could be moved/copied. Since neither conditions were true the suggested code produced an error.

### Reproducer

I originally encounter the issue here
https://github.com/bevyengine/bevy/blob/601cc0cbe3dc73c0db1993184f4ec47783e83884/crates/bevy_ecs/src/bundle.rs#L492 

A reduced version:

```rust
struct Inner {
    num: u8,
}

enum Outer<'a> {
    InnerBorrowed(&'a mut Inner),
    _InnerOwned(Inner),
}

impl<'a> Outer<'a> {
    fn get_inner_mut(&mut self) -> &mut Inner {
        let inner = match self {
            Outer::InnerBorrowed(inner) => &mut *inner,
            Outer::_InnerOwned(inner) => inner,
        };
        inner.num += 1;
        inner
    }
}

fn main() {
    let mut inner = Inner { num: 0 };
    let mut outer = Outer::InnerBorrowed(&mut inner);
    let _inner_inside_outer = outer.get_inner_mut();
}
```

`cargo clippy --fix` should change `Outer::InnerBorrowed(inner) => &mut *inner,` into `Outer::InnerBorrowed(inner) => inner,`

Instead, this happened:

`cargo clippy --fix` tries to change `Outer::InnerBorrowed(inner) => &mut *inner,` into `Outer::InnerBorrowed(inner) => *inner,` and then produces the following console output

<details>
    <summary>Console Output</summary>

    ❯ cargo clippy --fix                                 
        Checking test-234234 v0.1.0 (/extra/Projects/test)
    warning: failed to automatically apply fixes suggested by rustc to crate `test_234234`

    * src/main.rs

    error[E0507]: cannot move out of `*inner` which is behind a mutable reference
    --> src/main.rs:13:44
    |
    13 |             Outer::InnerBorrowed(inner) => *inner,
    |                                            ^^^^^^
    |                                            |
    |                                            move occurs because `*inner` has type `&mut Inner`, which does not implement the `Copy` trait
    |                                            help: consider borrowing here: `&*inner`

    For more information about this error, try `rustc --explain E0507`.

    warning: this expression borrows a reference (`&mut Inner`) that is immediately dereferenced by the compiler
    --> src/main.rs:13:44
    |
    13 |             Outer::InnerBorrowed(inner) => &mut *inner,
    |                                            ^^^^^^^^^^^ help: change this to: `*inner`
    |

    warning: `test-234234` (bin "test-234234") generated 1 warning (1 duplicate)
    warning: `test-234234` (bin "test-234234" test) generated 1 warning
        Finished dev [unoptimized + debuginfo] target(s) in 0.44s

</details>

### Version

```text
rustc 1.59.0-nightly (f8abed9ed 2021-12-26)
binary: rustc
commit-hash: f8abed9ed48bace6be0087bcd44ed534e239b8d8
commit-date: 2021-12-26
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

Jarcho:
fixes #8191

changelog: Fix `needless_borrow` causing mutable borrows to be moved
changelog: Rename `ref_in_deref` to `needless_borrow`
changelog: Suggest removing the borrow on method call receivers in `needless_borrow`


clippy_lints/src/dereference.rs
```diff
@@ -226,40 +225,58 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
...
-                                self.state = Some((
-                                    // Subtract one for the current borrow expression, and one to cover the last
-                                    // reference which can't be removed (it's either reborrowed, or needed for
-                                    // auto-deref to happen).
-                                    State::DerefedBorrow {
+                                // If the next adjustment is a mutable borrow, then check to see if the compiler will
+                                // insert a re-borrow here. If not, leave an extra borrow here to avoid attempting to
+                                // move the a mutable reference.
+                                let (i, target_mut) = if let Some(&Adjust::Borrow(AutoBorrow::Ref(_, mutability))) =
+                                    adjust.or_else(|| iter.next()).map(|a| &a.kind)
```
camsteffen, line 241: Wow this gets hairy. Nice work! I think I understand enough of what's happening to approve the approach here. Can we make a couple tweaks to make this more readable?
1. The first `i` is okay since it is an index, but reusing the name gets confusing and I'm not really sure what it means semantically. Do you have a more meaningful name?
2. Add a `let` binding for `adjust.or_else(|| iter.next())`.

Jarcho: I'll try writing it without iterators and see if that helps.

Jarcho: There's been a bit of a cleanup in that section, should be a little easier to follow.

Jarcho:
This now includes all cases covered by `ref_in_deref`, which lints `(&_).field`. The description for both would lead one to believe they lint the same thing. Should I just rename it?

camsteffen:
I think we can remove/deprecate `ref_in_deref` then?

Tangentially, I noticed the other day that we don't lint `(*x).y`. I wonder where that could fit.

Jarcho:
I'm (very slowly) working on an explicit deref lint. It would fit in there whenever that's done.

Jarcho:
That successfully broke the changelog... Not really sure what to change there.

camsteffen:
You need to add `declare_deprecated_lint!` in `clippy_lints/src/deprecated_lints.rs` and also add a deprecation test in `tests/ui/deprecated.rs`. Then run `cargo dev update_lints` and it should be good.

Jarcho:
That would mean `register_renamed` can't be used, right?

camsteffen:
`register_renamed` would make it so any usage of `ref_in_deref` counts as `needless_borrow`. That may or may not be a good thing but I think the lints are different enough that we don't want that? In any case the error message should point to `needless_borrow`.

Jarcho:
The description for `ref_in`deref` is

> Checks for references in expressions that use auto dereference.

The description for `needless_borrow` is

> Checks for address of operations (&) that are going to be dereferenced immediately by the compiler.

Those describe exactly the same thing. o the point where anything covered by one and not he other would be a clear false negative. Given that renaming sound like the appropriate thing to do.

camsteffen:
Okay, good point. So I think you can just add `register_renamed` by hand. To fix the error, change any `ref_in_deref` links in the changelog to be not be links anymore.

Jarcho:
I was trying to avoid editing the changelog, but I don't see any other option.

tests/ui/needless_borrow.stderr
```diff
@@ -1,64 +1,118 @@
...
+  --> $DIR/needless_borrow.rs:56:23
+   |
+LL |     let y: &mut i32 = &mut &mut x;
+   |                       ^^^^^^^^^^^ help: change this to: `x`
+
+error: this expression creates a reference which is immediately dereferenced by the compiler
+  --> $DIR/needless_borrow.rs:67:13
+   |
+LL |     let _ = (&s).len();
+   |             ^^^^ help: change this to: `s`
```
camsteffen, line 91: I think the message should be adjusted in this case to something like "borrowed before a method call". "which is immediately dereferenced by the compiler" is perhaps correct, but only on a technicality (re-borrowing?) that is probably not interesting to users.

camsteffen:
One idea for the lint message but otherwise looks good. Please update "changelog" to include the lint rename and added functionality.

camsteffen:
Squash some commits then will merge!

PR #8219, opened 2022-01-03
New macro utils

### LINKED ISSUE #7843: Clippy breaks when we change anything about the std macros
Clippy recognizes macros like `panic!()` and `format_args!()` by its exact expansion, such as here:

https://github.com/rust-lang/rust-clippy/blob/389a74b31aabca4aac3289763eeb2eccedd1b988/clippy_utils/src/higher.rs#L719-L736

And here:

https://github.com/rust-lang/rust-clippy/blob/389a74b31aabca4aac3289763eeb2eccedd1b988/clippy_utils/src/higher.rs#L505-L675

This is very fragile, as it needs to be changed whenever we improve/change their implementations in std/core. For example: https://github.com/rust-lang/rust-clippy/issues/7723 removes a `&`, after which clippy doesn't Recognize the arguments to panic anymore.

Worse, since Clippy is now not a submodule but a subtree in `rust-lang/rust`, it means that we can no longer make any changes at all to those macros without it getting blocked in CI on clippy. Clippy's dependence on all these implementations details effectively blocks development on `panic`, `assert*`, and `format_args`/`fmt::Arguments` in `rust-lang/rust`.

So as a maintainer of `std`, I'd like to ask again: please do not depend on these implementation details. We change this stuff. I'm expecting some big changes to fmt::Arguments and the assert macros in the future, and the panic macros we've also repeatedly changed tiny details of over the past year, and and they will change again.

See also:

- https://github.com/rust-lang/rust/pull/88860
- https://github.com/rust-lang/rust-clippy/issues/7723
- https://github.com/rust-lang/rust/pull/78343#issuecomment-719981591
- https://github.com/rust-lang/rust/pull/79145
- https://github.com/rust-lang/rust/pull/80851#issuecomment-767436784
- https://github.com/rust-lang/rust/pull/79100
- https://github.com/rust-lang/rust/issues/76495

---

I don't know what the best solution is, but making Clippy depend heavily on every detail of our macros is just going to cause a big headache down the road for someone. We should probably work on some way in std or rustc to make it easy to for clippy to have the information without having to do these hacks to revert a macro expansion.

Maybe we can have an empty and hidden `#[inline(always)] fn macro_hint<T>(hint: &str, _: T) {}` somewhere that we can call in our macros (e.g. `macro_hint("hey clippy, this is an assert_eq", (left, right))`). Then Clippy could recognize that instead.

camsteffen:
changelog: none

Sorry, this is a big one. A lot of interrelated changes and I wanted to put the new utils to use to make sure they are somewhat battle-tested. We may want to divide some of the lint-specific refactoring commits into batches for smaller reviewing tasks. I could also split into more PRs.

Introduces a bunch of new utils at `clippy_utils::macros::...`. Please read through the docs and give any feedback! I'm happy to introduce `MacroCall` and various functions to retrieve an instance. It feels like the missing puzzle piece. I'm also introducing `ExpnId` from rustc as "useful for Clippy too". @rust-lang/clippy 

Fixes #7843 by not parsing every node of macro implementations, at least the major offenders.

I probably want to get rid of `is_expn_of` at some point.


llogiq:
Wow, that's a rather big one. Will take me a while to review, but I'm on it.

llogiq:
Hat's off. This looks great. I found the macro origin messages annoying when I updated the test suite to the 2021 edition. And now they're gone. Thank you!

@bors r+ squash

camsteffen:
> Will take me a while to review

So much for that 😆 thanks!

llogiq:
Well, I had a few log-running tests. :laughing:

camsteffen:
Hmm Github doesn't see that this is merged.

PR #8220, opened 2022-01-04
Consider auto-deref when linting `manual_swap`

### LINKED ISSUE #8154: manual_swap suggests invalid multiple mutable borrow
### Description

Given the following function:

```
fn swap_xy(mut v: IVec3) -> IVec3 {
    let x = v.x;
    v.x = v.y;
    v.y = x;

    v
}
```

Clippy is suggesting the following:

```
warning: this looks like you are swapping `v.x` and `v.y` manually
  --> day19/src/main.rs:15:5
   |
15 | /     let x = v.x;
16 | |     v.x = v.y;
17 | |     v.y = x;
   | |___________^ help: try: `std::mem::swap(&mut v.x, &mut v.y)`
   |
   = note: `#[warn(clippy::manual_swap)]` on by default
   = note: or maybe you should use `std::mem::replace`?
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_swap
```

which produces the following error:

```
error[E0499]: cannot borrow `v` as mutable more than once at a time
  --> day19/src/main.rs:15:35
   |
15 |     std::mem::swap(&mut v.x, &mut v.y);
   |     --------------      -         ^ second mutable borrow occurs here
   |     |                   |
   |     |                   first mutable borrow occurs here
   |     first borrow later used by call
```

It looks like this may have been fixed in https://github.com/rust-lang/rust-clippy/issues/981 and then maybe reintroduced?

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Jarcho:
fixes #8154

changelog: Don't lint `manual_swap` when a field access involves auto-deref

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -621,6 +621,19 @@ fn projection_stack<'a, 'hir>(mut e: &'a Expr<'hir>) -> (Vec<&'a Expr<'hir>>, &'
     (result, root)
 }
 
+/// Gets the mutability of the custom deref adjustment, if any.
+pub fn expr_custom_deref_adjustment(cx: &LateContext<'_>, e: &Expr<'_>) -> Option<Mutability> {
+    cx.typeck_results()
+        .expr_adjustments(e)
+        .iter()
+        .find_map(|a| match a.kind {
+            Adjust::Deref(Some(d)) => Some(Some(d.mutbl)),
+            Adjust::Deref(None) => None,
+            _ => Some(None),
+        })
+        .and_then(|x| x)
+}
+
 /// Checks if two expressions can be mutably borrowed simultaneously
 /// and they aren't dependent on borrowing same thing twice
 pub fn can_mut_borrow_both(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>) -> bool {
@@ -629,7 +642,15 @@ pub fn can_mut_borrow_both(cx: &LateContext<'_>, e1: &Expr<'_>, e2: &Expr<'_>) -
     if !eq_expr_value(cx, r1, r2) {
         return true;
     }
+    if expr_custom_deref_adjustment(cx, r1).is_some() || expr_custom_deref_adjustment(cx, r2).is_some() {
+        return false;
+    }
+
     for (x1, x2) in s1.iter().zip(s2.iter()) {
+        if expr_custom_deref_adjustment(cx, x1).is_some() || expr_custom_deref_adjustment(cx, x2).is_some() {
+            return false;
+        }
+
         match (&x1.kind, &x2.kind) {
             (ExprKind::Field(_, i1), ExprKind::Field(_, i2)) => {
                 if i1 != i2 {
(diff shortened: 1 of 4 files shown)


clippy_utils/src/lib.rs
```diff
@@ -621,6 +621,20 @@ fn projection_stack<'a, 'hir>(mut e: &'a Expr<'hir>) -> (Vec<&'a Expr<'hir>>, &'
...
 
+/// Checks if the expression is borrowed via an adjustment.
+pub fn expr_borrow_adjustment(cx: &LateContext<'_>, e: &Expr<'_>) -> Option<Mutability> {
+    for a in cx.typeck_results().expr_adjustments(e) {
+        match a.kind {
+            Adjust::Deref(Some(d)) => return Some(d.mutbl),
+            Adjust::Deref(..) => (),
+            Adjust::Borrow(AutoBorrow::Ref(_, m)) => return Some(m.into()),
+            Adjust::Borrow(AutoBorrow::RawPtr(m)) => return Some(m),
+            _ => return None,
```
camsteffen, line 632: Are the borrow branches necessary? I would think the overloaded deref is what matters and there's an early return there anyways.

Jarcho: To fix the problem with `manual_swap`, no. The function name would need to be changed.

Jarcho: Done. Also now handles `Box`'s deref.

camsteffen: So the borrow branches are only needed for box? What is meant by "custom deref adjustment"? Why not just have `expr_has_overloaded_deref(cx, expr) -> bool`? Anything more specific than that seems pretty niche for clippy_utils.

Jarcho: The borrow branches were originally to match the function name, which would have handled just a borrow being added as well (e.g. `x.foo()` can add a borrow to `x`). That isn't needed for `manual_swap` as it will never look at just the receiver of a method call.

`Box` dereferencing is handled the same as a regular reference (Adjust::Deref(None)) rather than through a custom deref, but it still requires borrowing the `Box`, so there's special handling to detect that case. I'm pretty sure the plan is to remove all the special handling for `Box` from rustc at some point, so this can hopefully be removed eventually.

Naming is hard here as it only handles implicit deref's inserted by the compiler, so any name that doesn't mention that is going to be ambiguous. Since they're already called adjustments by rustc there isn't really a reason to call it anything else. I guess `overloaded_deref_adjustment` or `overloaded_implicit_deref` would work.

camsteffen: The following compiles, so I think box deref is okay?

```rust
let mut x = Box::new((String::new(), String::new()));
std::mem::swap(&mut x.0, &mut x.1);
```

Jarcho: So it does. I remember running into a case where it didn't match regular deref behaviour. I'll remove it and deal with it later if need be.

camsteffen:
r=me with squash and passing CI

camsteffen:
Thanks! @bors r+

PR #8221, opened 2022-01-04
Better detect when a field can be moved from in `while_let_on_iterator`

### LINKED ISSUE #8113: Incorrect suggestion in `while_let_on_iterator` if iterator cannot be moved
### Summary

Same as #7846, but for `while_let_on_iterator`

### Lint Name

while_let_on_iterator

### Reproducer

I tried this code:

```rust
fn main() {
    let mut arr = [vec![1, 2, 3].into_iter()];
    while let Some(x) = arr[0].next() {}
}
```

I saw this happen:

```
warning: this loop could be written as a `for` loop
 --> src\main.rs:3:5
  |
3 |     while let Some(x) = arr[0].next() {}
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `for x in arr[0]`
  |
  = note: `#[warn(clippy::while_let_on_iterator)]` on by default
```

the suggestion causes the error

```
error[E0508]: cannot move out of type `[std::vec::IntoIter<i32>; 1]`, a non-copy array
 --> src\main.rs:3:14
  |
3 |     for x in arr[0] {}
  |              ^^^^^^
  |              |
  |              cannot move out of here
  |              move occurs because `arr[_]` has type `std::vec::IntoIter<i32>`, which does not implement the `Copy` trait
```

No lint emitted

### Version

```text
rustc 1.59.0-nightly (e6b883c74 2021-12-08)
binary: rustc
commit-hash: e6b883c74f49f32cb5d1cbad3457f2b8805a4a38
commit-date: 2021-12-08
host: x86_64-pc-windows-msvc
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

Jarcho:
fixes #8113 

changelog: Better detect when a field can be moved from in `while_let_on_iterator`

### DIFF
--- a/clippy_lints/src/loops/while_let_on_iterator.rs
+++ b/clippy_lints/src/loops/while_let_on_iterator.rs
@@ -8,12 +8,13 @@ use clippy_utils::{
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, ErasedMap, NestedVisitorMap, Visitor};
-use rustc_hir::{def::Res, Expr, ExprKind, HirId, Local, PatKind, QPath, UnOp};
+use rustc_hir::{def::Res, Expr, ExprKind, HirId, Local, Mutability, PatKind, QPath, UnOp};
 use rustc_lint::LateContext;
-use rustc_span::{symbol::sym, Span, Symbol};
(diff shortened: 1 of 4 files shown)


llogiq:
Nice improvement! Thank you! @bors r+

PR #8223, opened 2022-01-04
Remove in_macro from clippy_utils

camsteffen:
changelog: none

Previously done in #7897 but reverted in #8170. I'd like to keep `in_macro` out of utils because if a span is from expansion in any way (desugaring or macro), we should not proceed without understanding the nature of the expansion IMO.

r? @llogiq

### DIFF
--- a/clippy_lints/src/init_numbered_fields.rs
+++ b/clippy_lints/src/init_numbered_fields.rs
@@ -1,5 +1,4 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::in_macro;
 use clippy_utils::source::snippet_with_applicability;
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind};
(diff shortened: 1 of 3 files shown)


clippy_lints/src/init_numbered_fields.rs
```diff
@@ -46,7 +45,7 @@ impl<'tcx> LateLintPass<'tcx> for NumberedFields {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
         if let ExprKind::Struct(path, fields, None) = e.kind {
             if !fields.is_empty()
-                && !in_macro(e.span)
+                && !e.span.from_expansion()
```
llogiq, line 48: This is not completely behavior-preserving? What if the expansion was a desugaring?

camsteffen: Right. If it is desugaring then it doesn't make sense to continue because the user can't change it.

llogiq: I see. Ok then.

llogiq:
Thank you! @bors r+

PR #8226, opened 2022-01-05
`manual_memcpy` fix

### LINKED ISSUE #8160: Bad suggestion to `clone_from_slice` from `VecDeque`
### Summary

Copying items from `VecDeque` to an array using a for-loop triggers a `manual_memcpy` lint.

The suggested code does not work since `VecDeque` is not trivially sliceable i.e. `&vec_deque[..]` is not valid code.

### Lint Name

manual-memcpy

### Reproducer

I tried this code:

```rust
use std::collections::VecDeque;

fn main() {
    let mut dst = [0; 10];
    let src = (0..dst.len()).collect::<VecDeque<_>>();
    
    for i in 0..dst.len() {
        dst[i] = src[i];
    }
}
```

I saw this happen:

```
warning: it looks like you're manually copying between slices
 --> src/main.rs:7:5
  |
7 | /     for i in 0..dst.len() {
8 | |         dst[i] = src[i];
9 | |     }
  | |_____^ help: try replacing the loop by: `dst.clone_from_slice(&src[..dst.len()]);`
  |
  = note: `#[warn(clippy::manual_memcpy)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_memcpy
```

Probably no lint? Or one needs to do two separate `clone_from_slice` calls for both parts of the `VecDeque.as_slices()`.

And since `T` is copy it should be suggesting `copy_from_slice`.

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

@rustbot label +`I-suggestion-causes-error`

Jarcho:
fixes #8160 

Ideally this would work with `VecDeque`, but the current interface is unsuitable for it. At a minimum something like `range_as_slices` would be needed.

changelog: Don't lint `manual_memcpy` on `VecDeque`
changelog: Suggest `copy_from_slice` for `manual_memcpy` when applicable

### DIFF
--- a/clippy_lints/src/loops/manual_memcpy.rs
+++ b/clippy_lints/src/loops/manual_memcpy.rs
@@ -2,6 +2,7 @@ use super::{IncrementVisitor, InitializeVisitor, MANUAL_MEMCPY};
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet;
 use clippy_utils::sugg::Sugg;
+use clippy_utils::ty::is_copy;
 use clippy_utils::{get_enclosing_block, higher, path_to_local, sugg};
 use if_chain::if_chain;
 use rustc_ast::ast;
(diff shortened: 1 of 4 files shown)


flip1995:
Can you add a test case for the `VecDeque` case, please? Otherwise this LGTM.

flip1995:
Perfect, thanks!

@bors r+

PR #8228, opened 2022-01-05
fix `iter_not_returning_iterator`

### LINKED ISSUE #8225: FP iter_not_returning_iterator with associated type
### Summary

I think associated types are not evaluated correctly when checking whether the return type implements `Iterator`. 

### Lint Name

iter_not_returning_iterator

### Reproducer

I tried this code: [play](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&code=%23!%5Bwarn(clippy%3A%3Aiter_not_returning_iterator)%5D%0A%0Atrait%20Foo%20%7B%0A%20%20%20%20type%20I%3B%0A%20%20%20%20fn%20iter(%26self)%20-%3E%20Self%3A%3AI%3B%0A%7D%0A%0Aimpl%20Foo%20for%20()%20%7B%0A%20%20%20%20type%20I%20%3D%20std%3A%3Avec%3A%3AIntoIter%3Cu8%3E%3B%0A%20%20%20%20fn%20iter(%26self)%20-%3E%20%3CSelf%20as%20Foo%3E%3A%3AI%20%7B%0A%20%20%20%20%20%20%20%20todo!()%0A%20%20%20%20%7D%0A%7D%0A)

```rust
#![warn(clippy::iter_not_returning_iterator)]

trait Foo {
    type I;
    fn iter(&self) -> Self::I;
}

impl Foo for () {
    type I = std::vec::IntoIter<u8>;
    fn iter(&self) -> Self::I {
        todo!()
    }
}
```

I saw this happen:

```
warning: this method is named `iter` but its return type does not implement `Iterator`
```

I expected the associated type to be substituted like a type alias.

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Jarcho:
fixes #8225 

changelog: Handle type projections in `iter_not_returning_iterator`
changelog: Don't lint `iter_not_returning_iterator` in trait implementations
changelog: Lint `iter_not_returning_iterator` in trait definitions

### DIFF
--- a/tests/ui/iter_not_returning_iterator.stderr
+++ b/tests/ui/iter_not_returning_iterator.stderr
@@ -12,5 +12,11 @@ error: this method is named `iter_mut` but its return type does not implement `I
 LL |     fn iter_mut(&self) -> Counter2 {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: aborting due to 2 previous errors
+error: this method is named `iter` but its return type does not implement `Iterator`
+  --> $DIR/iter_not_returning_iterator.rs:49:5
+   |
+LL |     fn iter(&self) -> Self::I;
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: aborting due to 3 previous errors
 
--- a/tests/ui/iter_not_returning_iterator.stderr
+++ b/tests/ui/iter_not_returning_iterator.stderr
@@ -13,7 +13,7 @@ LL |     fn iter_mut(&self) -> Counter2 {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 error: this method is named `iter` but its return type does not implement `Iterator`
-  --> $DIR/iter_not_returning_iterator.rs:49:5
+  --> $DIR/iter_not_returning_iterator.rs:50:5
    |
 LL |     fn iter(&self) -> Self::I;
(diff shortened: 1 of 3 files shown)


tests/ui/iter_not_returning_iterator.stderr
```diff
@@ -12,5 +12,11 @@ error: this method is named `iter_mut` but its return type does not implement `I
 LL |     fn iter_mut(&self) -> Counter2 {
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-error: aborting due to 2 previous errors
+error: this method is named `iter` but its return type does not implement `Iterator`
+  --> $DIR/iter_not_returning_iterator.rs:50:5
+   |
+LL |     fn iter(&self) -> Self::I;
```
giraffate, line 18: From the issue it looks like false positive, is this fixed? I tried the case described in the issue, but it seems that I am still warned.

Jarcho: The error occurs in a different spot. Before it errored on the trait impl, now it errors on the trait def. The trait def in the issue does not restrict the type in the trait so this is correct. The following test checks if the projection is correctly resolved to a type implementing `Iterator`.

Either fix would have fixed the fp in the issue.

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8231, opened 2022-01-05
Fix `implicit_clone` for `&&T`

### LINKED ISSUE #8227: FP implicit-clone: &&Path.to_path_buf()
### Summary

clippy suggestes invalid code when cloning `&&Path`s 

### Lint Name

clippy::implicit-clone

### Reproducer

I tried this code:

```rust
use std::path::PathBuf;

fn main() {
    let a: Vec<&PathBuf> = Vec::new();
    let mut c: Vec<PathBuf> = Vec::new();

    for path /* &&PathBuf */  in a.iter() {
        c.push(path.to_path_buf());
    }
}

```

I saw this happen:
`cargo clippy -- -Wclippy::implicit_clone `
```
    Checking clippy_bug v0.1.0 (/tmp/clippy_bug)
warning: implicitly cloning a `PathBuf` by calling `to_path_buf` on its dereferenced type
 --> src/main.rs:8:21
  |
8 |         c.push(path.to_path_buf());
  |                     ^^^^^^^^^^^ help: consider using: `clone`
  |
  = note: requested on the command line with `-W clippy::implicit-clone`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#implicit_clone
```

I expected the suggested code to compile, but it didn't :sweat_smile: 

````
 --> src/main.rs:8:16
  |
8 |         c.push(path.clone());
  |                ^^^^^^^^^^^^- help: try using a conversion method: `.to_path_buf()`
  |                |
  |                expected struct `std::path::PathBuf`, found `&std::path::PathBuf`

````
Maybe clippy does not understand that we have a double-reference here?

### Version

```text
clippy 0.1.59 (7d6f948 2022-01-04)

rustc 1.59.0-nightly (7d6f94817 2022-01-04)
binary: rustc
commit-hash: 7d6f948173ccb18822bab13d548c65632db5f0aa
commit-date: 2022-01-04
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

Jarcho:
fixes #8227

changelog: Don't lint `implicit_clone` on `&&T`

### DIFF
--- a/clippy_lints/src/methods/implicit_clone.rs
+++ b/clippy_lints/src/methods/implicit_clone.rs
@@ -1,31 +1,40 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::source::snippet_with_context;
+use clippy_utils::ty::peel_mid_ty_refs;
 use clippy_utils::{is_diag_item_method, is_diag_trait_item};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_lint::LateContext;
 use rustc_middle::ty::TyS;
-use rustc_span::{sym, Span};
+use rustc_span::sym;
 
 use super::IMPLICIT_CLONE;
 
-pub fn check(cx: &LateContext<'_>, method_name: &str, expr: &hir::Expr<'_>, recv: &hir::Expr<'_>, span: Span) {
+pub fn check(cx: &LateContext<'_>, method_name: &str, expr: &hir::Expr<'_>, recv: &hir::Expr<'_>) {
     if_chain! {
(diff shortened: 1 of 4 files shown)


camsteffen:
Hmm I feel like treating this as a false positive goes against the intention of the lint. I also get that suggesting `(*x).clone()` is not great, but I think that is the best we can do. In reality my suggestion would be to avoid `&&T` variables, but suggesting a fix for *that* is not really feasible.

Jarcho:
It's at least a pedantic lint so a less-than-great suggestion is fine. I'm sure we'll also get a bug report saying the suggestion sucks though.

PR #8234, opened 2022-01-06
Allow running lintcheck with a renamed rust-clippy dir

Alexendoo:
I have Clippy checked out in `rust/clippy` rather than `rust/rust-clippy`, this allows lintcheck to still run in that case

changelog: none


xFrednet (approved):

xFrednet:
LGTM, thank you for this change! I also found this restriction some time ago (but never fixed it like you did).

@bors r+

PR #8236, opened 2022-01-06
[squash-merged] new lint: `single_char_lifetime_names`

### LINKED ISSUE #8233: Introduce a lint for single letter lifetimes
### What it does

Highlights when a named lifetime is introduced that is a single character long. This feels like a very opinionated lint, and definitely overly pedantic in many cases, but I think there could be contexts where it's useful.

For example, I had to explain what the following impl meant in our Rocket web service:
```rust
impl<'r, 'o: 'r> Responder<'r, 'o> for Error {
    // ...
}
```
After reading the docs, I found a comment saying that `'r` was the lifetime of the request, and `'o` was the lifetime of the response. Admittedly this is somewhat intuitive by looking at the uses of the references in the type signature of the function in the trait, but I can't help but feel if it was instead written as: `impl <'req, 'res: 'req> ...`, I wouldn't even have to look at the docs to understand, which is significantly more valuable to those less familiar with Rust and borrowing/lifetimes.

### Lint Name

single_char_lifetime_name

### Category

pedantic

### Advantage

Easier to understand meaning of a specific named lifetime, especially if new to Rust

### Drawbacks

Likely overly pedantic in some codebases/teams/contexts
 - some contexts have only one lifetime that is sensible, and a single letter lifetime would be fine, e.g.:
```rust
struct Foo<'foo> {
  inner: &'foo str
}
```
is not (in my opinion) more readable/understandable than:
```rust
struct Foo<'a> {
  inner: &'a str
}
```
 - some teams/companies are familiar enough with the language that reading function signatures to work out the meaning of lifetimes doesn't add significant overhead

### Example

```rust
fn dispatch<'r, 'o: 'r>(req: &'r Request) -> &'o Repsonse {
  todo!()
}
```

Could be written as:

```rust
fn dispatch<'req, 'res: 'req>(req: &'req Request) -> &'res Repsonse {
  todo!()
}
```

PatchMixolydic:
This pull request adds a lint against single character lifetime names, as they might not divulge enough information about the purpose of the lifetime. This can make code harder to understand. I placed this in `restriction` rather than `pedantic` (as suggested in #8233) since most of the Rust ecosystem already uses single character lifetime names (to my knowledge, at least) and since single character lifetime names aren't incorrect. I'd be happy to change this upon request, however. Fixes #8233.

- [x] Followed lint naming conventions
- [x] Run `cargo dev fmt`

changelog: new lint: [`single_char_lifetime_names`]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3253,6 +3253,7 @@ Released 2018-09-13
 [`should_implement_trait`]: https://rust-lang.github.io/rust-clippy/master/index.html#should_implement_trait
 [`similar_names`]: https://rust-lang.github.io/rust-clippy/master/index.html#similar_names
 [`single_char_add_str`]: https://rust-lang.github.io/rust-clippy/master/index.html#single_char_add_str
+[`single_char_lifetime_names`]: https://rust-lang.github.io/rust-clippy/master/index.html#single_char_lifetime_names
 [`single_char_pattern`]: https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern
 [`single_component_path_imports`]: https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports
 [`single_element_loop`]: https://rust-lang.github.io/rust-clippy/master/index.html#single_element_loop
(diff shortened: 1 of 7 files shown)


llogiq (commented):
Just one small nit

PatchMixolydic:
I've added a test to confirm that `'_` isn't linted.

llogiq:
Thank you! @bors r+ squash

PatchMixolydic:
This seems to have landed in `master` ~5 hours ago, so I'm going to go ahead and close this.

Issue #8247, opened 2022-01-08
deref_addrof should not fire on ptr::addr_of


Ralith:
### Summary

`deref_addrof` suggests replacing `*ptr::addr_of!((*ptr).field)` with `(*ptr).field`, but this is not semantically equivalent: if `ptr` addresses a struct wherein `field` is initialized but other parts are not, the former is legal but the latter is UB. See e.g. https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-a-struct-field-by-field.

### Lint Name

deref_addrof

### Reproducer

I tried this code:

```rust
*ptr::addr_of!((*ptr).linear_velocity)
```

I saw this happen:

```
warning: immediately dereferencing a reference
   --> openxr/src/space.rs:263:26
    |
263 |                 .then(|| *ptr::addr_of!((*ptr).linear_velocity)),
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `(*ptr).linear_velocity`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#deref_addrof
```

I expected to see this happen: no warning

### Version

```text
rustc 1.56.1 (59eed8a2a 2021-11-01)
binary: rustc
commit-hash: 59eed8a2aac0230a8b53e89d4e99d55912ba6b35
commit-date: 2021-11-01
host: x86_64-unknown-linux-gnu
release: 1.56.1
LLVM version: 13.0.0
```

taiki-e:
Filed #8259 to downgrade this lint to `nursery` group that allowed by default.

PR #8259, opened 2022-01-10
Downgrade deref_addrof to nursery

taiki-e:
This lint currently suggests replacing the sound code with unsound code: #8247
Until that problem is fixed, I don't think it is good for this lint to be warned by default.

changelog: Moved [`deref_addrof`] to `nursery`


taiki-e:
replaced by #8268

PR #8261, opened 2022-01-10
Warn disallowed_methods and disallowed_types by default

### LINKED ISSUE #7841: Enable disallowed-method and disallowed-type by default
Could `disallowed-method` and `disallowed-type` be moved to `restriction` and made warn (or even deny) by default? I can't find any known issues with them and you need to configure them via `.clippy.toml` anyways before they do anything.

@rustbot label +C-enhancement

taiki-e:
Closes #7841

changelog: Moved [`disallowed_methods`] and [`disallowed_types`] to `style`

### DIFF
--- a/clippy_lints/src/disallowed_methods.rs
+++ b/clippy_lints/src/disallowed_methods.rs
@@ -11,6 +11,9 @@ declare_clippy_lint! {
     /// ### What it does
     /// Denies the configured methods and functions in clippy.toml
     ///
+    /// Note: Even though this lint is warn-by-default, it will only trigger if
+    /// methods are defined in the clippy.toml file.
+    ///
     /// ### Why is this bad?
     /// Some methods are undesirable in certain contexts, and it's beneficial to
     /// lint for them as needed.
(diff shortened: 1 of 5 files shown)


clippy_lints/src/disallowed_methods.rs
```diff
@@ -49,7 +49,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.49.0"]
     pub DISALLOWED_METHODS,
-    nursery,
+    style,
```
taiki-e, line 52: I'm not sure if style is the right group (I felt like it is "user-defined code styles"), but I couldn't find another group that looked good. (restriction that #7841 originally proposed is allowed by default.)

flip1995:
Yeah, I think style is fine here.

I would add a note to the `What it does` section, that says somehting like:

> Note: Even though this lint is warn-by-default, it will only trigger if methods/types are defined in the clippy.toml file.

taiki-e:
@flip1995 Thanks. I applied your suggestion.

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #8262, opened 2022-01-10
Improve documentation for `borrowed-box` lint

### LINKED ISSUE #8161: Improve documentation of `borrowed_box`
### Description

This came up in #3971.

The documentation could need some improvement. It should mention, that by removing the Box from ``&Box<T>``, you no longer have to box the parameter when it is passed to the function. For more information, see the discussion in #3971.

### Version

_No response_

### Additional Labels

_No response_

1nF0rmed:
fixes #8161 

Updates documentation to elaborate more on how removing Box from a function parameter can generalize the function.

changelog: none

### DIFF
--- a/clippy_lints/src/types/mod.rs
+++ b/clippy_lints/src/types/mod.rs
@@ -167,8 +167,9 @@ declare_clippy_lint! {
     /// Check the [Box documentation](https://doc.rust-lang.org/std/boxed/index.html) for more information.
     ///
     /// ### Why is this bad?
-    /// Any `&Box<T>` can also be a `&T`, which is more
-    /// general.
+    /// A `&Box<T>` parameter requires the function caller to box `T` first before passing it to a function.
+    /// Using `&T` defines a concrete type for the parameter and generalizes the function, this would also
+    /// auto-deref to `&T` at the function call site if passed a `&Box<T>`.
     ///
     /// ### Example
     /// ```rust,ignore


clippy_lints/src/types/mod.rs
```diff
@@ -167,8 +167,8 @@ declare_clippy_lint! {
     /// Check the [Box documentation](https://doc.rust-lang.org/std/boxed/index.html) for more information.
     ///
     /// ### Why is this bad?
-    /// Any `&Box<T>` can also be a `&T`, which is more
-    /// general.
+    /// A `&Box<T>` parameter requires the function caller to box `T` first before passing it to a function.
+    /// Using `&T` defines a concrete type for the parameter and generalizes the function.
```
camsteffen, line 171: I think it should also include something like "a `&Box<T>` will auto-deref to `&T` at the function call site".

1nF0rmed: Would something like this work:
A `&Box<T>` parameter will auto-deref to `&T` at the function call site and requires the function caller to box `T` first before passing it to a function. 
Using `&T` defines a concrete type for the parameter and generalizes the function.

camsteffen: I mean that if the parameter type is changed to `&T`, then you may still have `&Box<T>` passed to the function and it will automatically deref to `&T` with having to add a deref operator. The coder may not be aware of this, and that is IMO a likely reason that they choose to keep the Box in the first place.

1nF0rmed: My bad, in that case, it would be something like this:
A `&Box<T>` parameter requires the function caller to box `T` first before passing it to a function. 
Using `&T` defines a concrete type for the parameter and generalizes the function, this would also auto-deref to `&T` at the function call site if passed a `&Box<T>`.

camsteffen: Ya!

camsteffen:
Thanks! @bors r+

(I also changed the post with `changelog: none` since this doesn't need a changelog entry)

PR #8266, opened 2022-01-10
Some test code cleanup

camsteffen:
changelog: none

Mainly moves /clippy_workspace_tests into /tests and combines the two dogfood tests which can't run concurrently.


tests/dogfood.rs
```diff
@@ -3,179 +3,26 @@
...
+    if IS_RUSTC_TEST_SUITE {
         return;
     }
 
-    // NOTE: `path_dep` crate is omitted on purpose here
-    for project in &[
-        "clippy_workspace_tests",
-        "clippy_workspace_tests/src",
-        "clippy_workspace_tests/subcrate",
-        "clippy_workspace_tests/subcrate/src",
```
camsteffen, line 167: These aren't dogfooded anymore which is less testing strictly speaking. But we're already running clippy against these in the actual workspace test.

.github/workflows/clippy_bors.yml
```diff
@@ -133,14 +133,6 @@ jobs:
...
       working-directory: clippy_dev
 
-    - name: Test cargo-clippy
-      run: ../target/debug/cargo-clippy
-      working-directory: clippy_workspace_tests
-
-    - name: Test cargo-clippy --fix
-      run: ../target/debug/cargo-clippy clippy --fix
-      working-directory: clippy_workspace_tests
-
```
camsteffen, line 143: These could be replaced with the new directory. I'm just not sure what this is supposed to test, and the test should probably be contained in the new `workspace.rs`.

flip1995: Yeah, no idea why this is tested here.

tests/dogfood.rs
```diff
@@ -3,179 +3,26 @@
 //!
 //! See [Eating your own dog food](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) for context
 
-// Dogfood cannot run on Windows
-#![cfg(not(windows))]
```
flip1995, line 7: The windows and mac runners are our slowest runners in CI and also the current bottleneck. This added another 1-2min to the windows runner. So I wonder if we should only run those tests on `unix`. An  argument against that is, that contributors using windows/mac, won't see dogfood errors until they submit a PR.

flip1995 (approved):
LGTM. One more question about dogfood on windows. If you think it should be enabled feel free to r=me.

tests/dogfood.rs
camsteffen: Thanks for reviewing this.

Maybe I can find a way to get the best of both worlds using a flag or environment variable.

flip1995:
That's a neat solution! Thanks!

PR #8271, opened 2022-01-13
Check usages in `ptr_arg`

### LINKED ISSUE #214: false positive on `ptr_arg` against &Vec<_>
It's not possible to call any of `Vec`'s methods on a slice, whether mutable or not. So if any method of Vec is called, the lint message has to be considered a false positive.

Example: `fn foo(v: &Vec<Name>, n: Name) { v.push(n); }`

### LINKED ISSUE #1981: Typedefs and `ptr_arg` lint
I have the following situation:
```rust
type Value = Vec<u8>;

fn foo(_: &Value) {}
```
Typedef is used for clarity, but Clippy [produces](https://play.rust-lang.org/?gist=7f94e3fa12b35627ec98aedb56babbc8&version=stable) a [warning](https://github.com/rust-lang-nursery/rust-clippy/wiki#ptr_arg):
> warning: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices. Consider changing the type to `&[...]`

Probably this lint can be disabled for typedes?

### LINKED ISSUE #3381: ptr_arg: clarify about original type when newtype is just a Vec<_>
`clippy 0.0.212 (b1d0343 2018-10-19)`

````rust
fn main() {}

type SomeType = Vec<String>;

pub fn do_not_eat(a: &SomeType) -> SomeType {
	for i in a {
		println!("{}", i);
	}
	a.to_vec()
}
````

clippy warns
````
warning: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices.
 --> src/main.rs:5:22
  |
5 | pub fn do_not_eat(a: &SomeType) -> SomeType {
  |                      ^^^^^^^^^
  |
  = note: #[warn(clippy::ptr_arg)] on by default
````
but I think the message looks like a false positive if we don't see (as in this example) that the type `SomeType` is actually just a `Vec<_>` in disguise.

Can we perhaps print some detail on what the actual type is in case of newtypes to make this slightly more clear?

I found similar code in rls where `syntax::ast::GenericBounds` is actually just `type GenericBounds = Vec<GenericBound>;`

### LINKED ISSUE #6406: Suggest changing `&mut Vec<_>` parameters to `&mut [_]` when only slice methods are used.
This mistake comes up fairly frequently on URLO, such as 
https://users.rust-lang.org/t/iterators-vs-index-loops-performance/52131?u=scottmcm
https://users.rust-lang.org/t/we-all-know-iter-is-faster-than-loop-but-why/51486/3?u=scottmcm

### What it does

- For a free function or inherent method that takes `&mut Vec<_>` ~~or `&Vec<_>`~~ [ed: the `&Vec<_>` one already exists]
- But only uses it via its deref to slice (len, iter_mut, etc)
- Suggest changing the parameter to a slice instead

### Categories (optional)

- Kind: `perf` & `style` both seem appropriate

- More general in what it accepts, as one can pass sub-slices of `Vec`s instead of only being able to pass entire ones
- Faster, because LLVM optimizes it better when it doesn't have to do the extra hop)
- Clearer, because it makes it obvious that you're not changing the length of the passed `Vec`

For example:
- Remove bounce checking inserted by ...
- Remove the need to duplicating/storing/typo ...

### Drawbacks

None.

### Example

```rust
fn use_for(vec: &mut Vec<usize>) {
    for i in 0..vec.len() {
        vec[i] = i;
    }
}
```

Could be written as:

```rust
fn use_for(vec: &mut [usize]) {
    for i in 0..vec.len() {
        vec[i] = i;
    }
}
```

### LINKED ISSUE #6964: (Possibly) confusing message in ptr_arg
ptr_arg prints the message:

```warning: writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices```

I was confused for a while about the "one more reference", as switching does not seem to remove any references.

After some thought, is this referring to the internal implementation of Vec requiring another indirection? In that case, perhaps replace "involves one more reference" with "may produce slower code", or something like that?

Jarcho:
fixes #214
fixes #1981
fixes #3381
fixes #6406
fixes #6964

This does not take into account the return type of the function currently, so `(&Vec<_>) -> &Vec<_>` functions may still be false positives.

The name given for the type also has to match the real type name, so `type Foo = Vec<u32>` won't trigger the lint, but `type Vec = Vec<u32>` will. I'm not sure if this is the best way to handle this, or if a note about the actual type should be added instead.

changelog: Check if the argument is used in a way which requires the original type in `ptr_arg`
changelog: Lint mutable references in `ptr_arg`

### DIFF
--- a/clippy_lints/src/ptr.rs
+++ b/clippy_lints/src/ptr.rs
@@ -1,30 +1,36 @@
 //! Checks for usage of  `&Vec[_]` and `&String`.
 
 use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg, span_lint_and_then};
-use clippy_utils::ptr::get_spans;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::ty::walk_ptrs_hir_ty;
-use clippy_utils::{expr_path_res, is_lint_allowed, match_any_diagnostic_items, paths};
+use clippy_utils::ty::expr_sig;
+use clippy_utils::{
+    expr_path_res, get_expr_use_or_unification_node, is_lint_allowed, match_any_diagnostic_items, path_to_local, paths,
+};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::def::Res;
+use rustc_hir::def_id::DefId;
+use rustc_hir::hir_id::HirIdMap;
+use rustc_hir::intravisit::{walk_expr, NestedVisitorMap, Visitor};
 use rustc_hir::{
-    BinOpKind, BodyId, Expr, ExprKind, FnDecl, FnRetTy, GenericArg, Impl, ImplItem, ImplItemKind, Item, ItemKind,
-    Lifetime, MutTy, Mutability, Node, PathSegment, QPath, TraitFn, TraitItem, TraitItemKind, Ty, TyKind,
+    self as hir, AnonConst, BinOpKind, BindingAnnotation, Body, Expr, ExprKind, FnDecl, FnRetTy, GenericArg,
+    ImplItemKind, ItemKind, Lifetime, LifetimeName, Mutability, Node, Param, ParamName, PatKind, QPath, TraitFn,
+    TraitItem, TraitItemKind, TyKind,
 };
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::hir::map::Map;
+use rustc_middle::ty::{self, AssocItems, AssocKind, Ty};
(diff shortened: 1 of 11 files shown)


flip1995 (commented):
This looks good to me. I left some minor comments.

flip1995:
@bors r+

Thanks!

PR #8070, opened 2021-12-03
Improve styles of filtering options for Clippy's lint list

CrazyRoka:
Partially solves #7958 

Updated styles for filtering options. It now uses dropdown menus.

image

changelog: none

PR #8274, opened 2022-01-13
Update markdown-it version

andrewarchi:
Fixes improperly-escaped pipes in Markdown tables for the [documentation of bad_bit_mask](https://rust-lang.github.io/rust-clippy/master/index.html#bad_bit_mask) and [ineffective_bit_mask](https://rust-lang.github.io/rust-clippy/master/index.html#ineffective_bit_mask). A column pipe takes precedence over inline code markers, so some back ticks are displayed literally and the pipes need to be escaped. I found no other occurrences of the same problem when searching rust-clippy by ```\|.*`.*\|```.

changelog: none

### DIFF
--- a/clippy_lints/src/bit_mask.rs
+++ b/clippy_lints/src/bit_mask.rs
@@ -18,14 +18,14 @@ declare_clippy_lint! {
     /// {`!=`, `>=`, `>`, `!=`, `>=`, `>`}) can be determined from the following
     /// table:
     ///
-    /// |Comparison  |Bit Op|Example     |is always|Formula               |
-    /// |------------|------|------------|---------|----------------------|
-    /// |`==` or `!=`| `&`  |`x & 2 == 3`|`false`  |`c & m != c`          |
-    /// |`<`  or `>=`| `&`  |`x & 2 < 3` |`true`   |`m < c`               |
-    /// |`>`  or `<=`| `&`  |`x & 1 > 1` |`false`  |`m <= c`              |
-    /// |`==` or `!=`| `|`  |`x | 1 == 0`|`false`  |`c | m != c`          |
-    /// |`<`  or `>=`| `|`  |`x | 1 < 1` |`false`  |`m >= c`              |
-    /// |`<=` or `>` | `|`  |`x | 1 > 0` |`true`   |`m > c`               |
+    /// |Comparison  |Bit Op|Example      |is always|Formula               |
+    /// |------------|------|-------------|---------|----------------------|
+    /// |`==` or `!=`| `&`  |`x & 2 == 3` |`false`  |`c & m != c`          |
+    /// |`<`  or `>=`| `&`  |`x & 2 < 3`  |`true`   |`m < c`               |
+    /// |`>`  or `<=`| `&`  |`x & 1 > 1`  |`false`  |`m <= c`              |
+    /// |`==` or `!=`| `\|` |`x \| 1 == 0`|`false`  |`c \| m != c`         |
+    /// |`<`  or `>=`| `\|` |`x \| 1 < 1` |`false`  |`m >= c`              |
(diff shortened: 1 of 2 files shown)


camsteffen:
Hmm the docs already look right to me at least on the website. Are you seeing it rendered incorrectly somewhere else?

andrewarchi:
The incorrectly-rendered tables shows on my Firefox instance, but not any other browser, so I did some digging and it turns out the culprit is the [LocalCDN](https://www.localcdn.org/) extension, which substitutes the CDNJS-hosted markdown-it (7.0.0) with a local version (12.3.0). If I disable LocalCDN for the Clippy docs, then it renders as expected.

This reveals the larger issue is that the markdown-it version needs updating.

For comparison, I have tested the tables in question in VS Code (which also uses markdown-it) and on GitHub and both render the same as the current markdown-it version. For consistency, I think we ought to update the version.

The `bad_bit_mask` table (without my escaping) rendered by GitHub:

|Comparison  |Bit Op|Example     |is always|Formula               |
|------------|------|------------|---------|----------------------|
|`==` or `!=`| `&`  |`x & 2 == 3`|`false`  |`c & m != c`          |
|`<`  or `>=`| `&`  |`x & 2 < 3` |`true`   |`m < c`               |
|`>`  or `<=`| `&`  |`x & 1 > 1` |`false`  |`m <= c`              |
|`==` or `!=`| `|`  |`x | 1 == 0`|`false`  |`c | m != c`          |
|`<`  or `>=`| `|`  |`x | 1 < 1` |`false`  |`m >= c`              |
|`<=` or `>` | `|`  |`x | 1 > 0` |`true`   |`m > c`               |

andrewarchi:
The other dependencies could be updated as follows:

[angular.js](https://cdnjs.com/libraries/angular.js): 1.4.12 -> 1.8.2
[highlight.js](https://cdnjs.com/libraries/highlight.js): 9.5.0 -> 11.4.0
[markdown-it](https://cdnjs.com/libraries/markdown-it): 7.0.0 -> 12.3.2
[twitter-bootstrap](https://cdnjs.com/libraries/twitter-bootstrap): 3.3.6 -> 4.6.1

I reviewed the changelogs for highlight.js and markdown-it and it seems that those could be simply updated without any other changes.

camsteffen:
In that case do you want to upgrade those two in this PR?

The other upgrades are probably worthwhile as well but those can be separate PRs and also should wait for #8070.

I also wonder if we should move to Angular (not AngularJS) or perhaps Vue but that's another discussion.

andrewarchi:
I've upgraded markdown-it and highlight.js and I agree that AngularJS and Bootstrap are separate issues.

I have not tested it on my end as I am not sure of your pipeline for generating the GH Pages tree.

camsteffen:
Try `cargo dev serve` to test

andrewarchi:
The markdown-it fixes look to work properly,

Bumping the highlight.js version messes with the highlighting colors though. The page pulls in a custom [highlight.css](https://github.com/rust-lang/rust-clippy/blob/3cd151def9bd0f58ca6984f4c27144732f6865e0/util/gh-pages/index.html#L21) theme from rust-lang/mdBook, which overrides some of the styles from the github.min.css highlight.js theme. It looked fine before, but there must have been changes since then in highlight.js with the GitHub theme or class names. If either github.min.css or highlight.css are disabled, then it looks fine, though highlight.css is required for the color-theme-changing feature.

I think I'll revert the highlight.js version bump because I don't want to fix the bespoke themes in mdBook too. Interestingly, mdBook upgraded highlight.js to v11 last year, but [reverted](https://github.com/rust-lang/mdBook/pull/1621) shortly thereafter.

camsteffen:
Thanks!

@bors r+

PR #8280, opened 2022-01-14
Add `msrv` config for `map_clone`

xFrednet:
Just a small PR to have some fun with Clippy and to clear my head a bit :sweat_smile: 

changelog: [`map_clone`]: The suggestion takes `msrv` into account


clippy_lints/src/map_clone.rs
```diff
@@ -105,6 +116,8 @@ impl<'tcx> LateLintPass<'tcx> for MapClone {
             }
         }
     }
+
+    extract_msrv_attr!(LateContext);
```
flip1995, line 120: Seeing this PR, I have the suspicion, that this is not always added when new MSRVs get introduced. I will have to check that across Clippy. Same thing with the lint names in the conf.rs docs.

flip1995:
(full review follows soon)

clippy_lints/src/map_clone.rs
xFrednet: I would think that the conf.rs docs are quite up-to-date, but the `extract_msrv_attr` might not be. Now the question is if I'm going to look into it and procrastinate a bit more...

xFrednet: Found one without the macro!

camsteffen: Sounds like an internal lint, ya dig?

xFrednet: Could be, we could probably just check if the struct has a `msrv: Option<RustcVersion>,` field to determine if it should use the macro :thinking:

xFrednet: And a second one, the other lint passes all have that macro :upside_down_face:

xFrednet: Now I've checked, all lints are now listed in the `conf.rs`. Only `manual_bits` and `borrow_as_prt` were missing.

flip1995:
Thanks!

@bors r+

Issue #8282, opened 2022-01-15
`single_match` should not lint exhaustive `match`


camsteffen:
### Summary

`single_match` should not lint an exhaustive `match`

### Lint Name

single_match

### Reproducer

I tried this code:

```rust
enum Foo { Bar }

match Some(Foo::Bar) {
    Some(Foo::Bar) => {dbg!()}
    None => {}
}
```

I saw this happen:

```
warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
 --> src/main.rs:5:5
  |
5 | /     match Some(Foo::Bar) {
6 | |         Some(Foo::Bar) => {dbg!()}
7 | |         None => {}
8 | |     }
  | |_____^ help: try this: `if let Some(Foo::Bar) = Some(Foo::Bar) {dbg!()}`
```

I expected to see this happen: nothing

camelid:
Context: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Lint.20group.20for.20.60return_self_not_must_use.60/near/268089511

jubnzv:
The problem is in [this line](https://github.com/rust-lang/rust-clippy/tree/7e3e6c5df6d137a0566427b108e6bf7aef98e8c1/clippy_lints/src/matches.rs#L781), for some reasons `implements_trait(cx, ty, pe_trait_id, &[ty.into()])` doesn't work as expected for `Option<Foo>` type. I'm looking at this.

camsteffen:
I think the important thing here is that there is no wild pattern (`_`) anywhere in an exhaustive match. You can use [`Pat::walk_short`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_hir/hir/struct.Pat.html#method.walk_short) to search each pattern for one with `PatKind::Wild`.

camelid:
There's also `match` arms like `(_, _)` -- basically, when matching on a single-constructor (excluding constructors that are uninhabited) type, applying that constructor to wildcards behaves like a plain wildcard. Not sure how you want to handle those.

camsteffen:
The presence of `_` in `(_, _)` means it's a non-exhaustive match, right?

Also, a wild in a branch with an `if` guard doesn't count.

camelid:
> The presence of `_` in `(_, _)` means it's a non-exhaustive match, right?

[...]

Issue #8215, opened 2022-01-03
v.checked_sub(1).unwrap_or(v) is also an implicit_saturating_sub


gilescope:
### Description

we already detect saturating_subs done one way:

https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/implicit_saturating_sub.rs

I came across someone doing it another way: `v.checked_sub(1).unwrap_or(v)`

Thought it might be one to add into the logic. Maybe an easy one for a new person to hack on?

### Version

```text
n/a
```

giraffate:
As noted above, the [`manual_saturating_arithmetic`](https://rust-lang.github.io/rust-clippy/master/index.html#manual_saturating_arithmetic) covers this, so I'm closing this.

Issue #8212, opened 2022-01-02
New lint: `redundant_sort_by_key`.


pmnoxx:
### What it does

I saw an interesting bug today. Apparently the following code is valid.

```rust
[3,1,2].sort_by_key(|x|{x;}|);
```
The extra semicolon makes this function return `()`, which is valid, but this makes the sorting pointless.

### Lint Name

`useless_sort_by_key`

### Category

correctness

### Advantage

- This is likely a bug, and should be fixed. I had it happen to my code, once.
- Or it can be just called redundant sort_by_key.

### Drawbacks

- I don't see any

### Example

```rust
[3,1,2].sort_by_key(|x|{x;}|);
```

Could be written as:

That codes does nothing can be removed.
Or can be fixed, but it may be impossible to figure out what the user intended
```rust
[3,1,2].sort_by_key(|x|{x}|);
```

I think this should be marked as an potential error.

5225225:
IMO this should apply to most/all `_by_key` methods.

So you'd be looking for all calls of a specific method with a signature like

```rust
fn max_by_key<B, F>(self, f: F) -> Option<Self::Item> where
    B: Ord,
    F: FnMut(&Self::Item) -> B, 
```

where `B` is `()`.

Though I suppose there may be legitimate use cases for a unit `_by_key` for some methods, so it might be a good idea to restrict it to just some types / be a bit smarter and let people explicitly return unit.

PR #8289, opened 2022-01-15
Add `default_union_representation` lint

### LINKED ISSUE #8235: Warn on unions with no `#[repr]` attribute
### What it does

Unions in rust have unspecified layout by default, despite many people thinking that they lay out each type at the start of the union (like C does)

### Lint Name

unspecified_layout_union

### Category

pedantic

### Advantage

If the user intended the union to be `#[repr(C)]` because they are using it to read fields other than the most recently written one, the code becomes defined behaviour.

If the user did not intend any specific representation because they are using it in an externally tagged way, the `#[allow(unspecified_layout_union)]` calls this out specifically. 

### Drawbacks

Some unions don't rely on the representation being C-like, and this may be the majority of the unions that have no explicit representation.

This lint, once added, would fire on all of these unions, as it's currently impossible to specify the rust repr explicitly. This might be a source of noise. I originally thought about `suspicious` as a category for this lint, but downgraded it to `pedantic` because of that.

### Example

```rust
union Foo {
    a: i32,
    b: u32,
}

fn main() {
    let _x: u32 = unsafe {
        Foo { a: 0_i32 }.b // UB, `b` is allowed to be padding
    };
}
```

Could be written as:

```rust
#[repr(C)]
union Foo {
    a: i32,
    b: u32,
}

fn main() {
    let _x: u32 = unsafe {
        Foo { a: 0_i32 }.b // Now defined behaviour, this is just an i32 -> u32 transmute.
    };
}
```

jubnzv:
Closes #8235

changelog: Added a new lint  [`default_union_representation`]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2930,6 +2930,7 @@ Released 2018-09-13
 [`declare_interior_mutable_const`]: https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
 [`default_numeric_fallback`]: https://rust-lang.github.io/rust-clippy/master/index.html#default_numeric_fallback
 [`default_trait_access`]: https://rust-lang.github.io/rust-clippy/master/index.html#default_trait_access
+[`default_union_representation`]: https://rust-lang.github.io/rust-clippy/master/index.html#default_union_representation
 [`deprecated_cfg_attr`]: https://rust-lang.github.io/rust-clippy/master/index.html#deprecated_cfg_attr
 [`deprecated_semver`]: https://rust-lang.github.io/rust-clippy/master/index.html#deprecated_semver
 [`deref_addrof`]: https://rust-lang.github.io/rust-clippy/master/index.html#deref_addrof
--- a/clippy_lints/src/default_union_representation.rs
+++ b/clippy_lints/src/default_union_representation.rs
@@ -1,0 +1,105 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use rustc_hir::{self as hir, HirId, Item, ItemKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::layout::LayoutOf;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+use rustc_typeck::hir_ty_to_ty;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Displays a warning when a union is declared with the default representation (without a `#[repr(C)]` attribute).
+    ///
+    /// ### Why is this bad?
+    /// Unions in Rust have unspecified layout by default, despite many people thinking that they
+    /// lay out each field at the start of the union (like C does). That is, there are no guarantees
+    /// about the offset of the fields for unions with multiple non-ZST fields without an explicitly
+    /// specified layout. These cases may lead to undefined behavior in unsafe blocks.
+    ///
+    /// ### Example
+    /// ```rust
+    /// union Foo {
+    ///     a: i32,
+    ///     b: u32,
+    /// }
+    ///
+    /// fn main() {
+    ///     let _x: u32 = unsafe {
+    ///         Foo { a: 0_i32 }.b // Undefined behaviour: `b` is allowed to be padding
+    ///     };
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// #[repr(C)]
+    /// union Foo {
+    ///     a: i32,
+    ///     b: u32,
+    /// }
+    ///
+    /// fn main() {
+    ///     let _x: u32 = unsafe {
+    ///         Foo { a: 0_i32 }.b // Now defined behaviour, this is just an i32 -> u32 transmute
+    ///     };
+    /// }
+    /// ```
+    #[clippy::version = "1.60.0"]
+    pub DEFAULT_UNION_REPRESENTATION,
+    restriction,
+    "unions without a `#[repr(C)]` attribute"
+}
+declare_lint_pass!(DefaultUnionRepresentation => [DEFAULT_UNION_REPRESENTATION]);
+
+impl<'tcx> LateLintPass<'tcx> for DefaultUnionRepresentation {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
+        if is_union_with_two_non_zst_fields(cx, item) && !has_c_repr_attr(cx, item.hir_id()) {
+            span_lint_and_help(
+                cx,
+                DEFAULT_UNION_REPRESENTATION,
+                item.span,
+                "this union has the default representation",
+                None,
+                &format!(
+                    "consider annotating `{}` with `#[repr(C)]` to explicitly specify memory layout",
+                    cx.tcx.def_path_str(item.def_id.to_def_id())
+                ),
+            );
+        }
+    }
+}
+
+/// Returns true if the given item is a union with at least two non-ZST fields.
+fn is_union_with_two_non_zst_fields(cx: &LateContext<'_>, item: &Item<'_>) -> bool {
+    if let ItemKind::Union(data, _) = &item.kind {
+        data.fields().iter().filter(|f| !is_zst(cx, f.ty)).count() >= 2
+    } else {
+        false
+    }
+}
+
+fn is_zst(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>) -> bool {
+    if hir_ty.span.from_expansion() {
+        return false;
+    }
+    let ty = hir_ty_to_ty(cx.tcx, hir_ty);
+    if let Ok(layout) = cx.layout_of(ty) {
+        layout.is_zst()
+    } else {
+        false
+    }
+}
+
+fn has_c_repr_attr(cx: &LateContext<'_>, hir_id: HirId) -> bool {
+    cx.tcx.hir().attrs(hir_id).iter().any(|attr| {
+        if attr.has_name(sym::repr) {
+            if let Some(items) = attr.meta_item_list() {
+                for item in items {
+                    if item.is_word() && matches!(item.name_or_empty(), sym::C) {
+                        return true;
+                    }
+                }
+            }
+        }
+        false
+    })
+}
--- a/clippy_lints/src/lib.register_lints.rs
+++ b/clippy_lints/src/lib.register_lints.rs
@@ -92,6 +92,7 @@ store.register_lints(&[
     default::DEFAULT_TRAIT_ACCESS,
     default::FIELD_REASSIGN_WITH_DEFAULT,
     default_numeric_fallback::DEFAULT_NUMERIC_FALLBACK,
+    default_union_representation::DEFAULT_UNION_REPRESENTATION,
     dereference::EXPLICIT_DEREF_METHODS,
     dereference::NEEDLESS_BORROW,
     dereference::REF_BINDING_TO_REFERENCE,
--- a/clippy_lints/src/lib.register_restriction.rs
+++ b/clippy_lints/src/lib.register_restriction.rs
@@ -12,6 +12,7 @@ store.register_group(true, "clippy::restriction", Some("clippy_restriction"), ve
     LintId::of(create_dir::CREATE_DIR),
     LintId::of(dbg_macro::DBG_MACRO),
     LintId::of(default_numeric_fallback::DEFAULT_NUMERIC_FALLBACK),
+    LintId::of(default_union_representation::DEFAULT_UNION_REPRESENTATION),
     LintId::of(disallowed_script_idents::DISALLOWED_SCRIPT_IDENTS),
     LintId::of(else_if_without_else::ELSE_IF_WITHOUT_ELSE),
     LintId::of(exhaustive_items::EXHAUSTIVE_ENUMS),
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -189,6 +189,7 @@ mod create_dir;
 mod dbg_macro;
 mod default;
 mod default_numeric_fallback;
+mod default_union_representation;
 mod dereference;
 mod derivable_impls;
 mod derive;
@@ -860,6 +861,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_early_pass(|| Box::new(single_char_lifetime_names::SingleCharLifetimeNames));
     store.register_late_pass(move || Box::new(borrow_as_ptr::BorrowAsPtr::new(msrv)));
     store.register_late_pass(move || Box::new(manual_bits::ManualBits::new(msrv)));
+    store.register_late_pass(|| Box::new(default_union_representation::DefaultUnionRepresentation));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/tests/ui/default_union_representation.rs
+++ b/tests/ui/default_union_representation.rs
@@ -1,0 +1,78 @@
+#![feature(transparent_unions)]
+#![warn(clippy::default_union_representation)]
+
+union NoAttribute {
+    a: i32,
+    b: u32,
+}
+
+#[repr(C)]
+union ReprC {
+    a: i32,
+    b: u32,
+}
+
+#[repr(packed)]
+union ReprPacked {
+    a: i32,
+    b: u32,
+}
+
+#[repr(C, packed)]
+union ReprCPacked {
+    a: i32,
+    b: u32,
+}
+
+#[repr(C, align(32))]
+union ReprCAlign {
+    a: i32,
+    b: u32,
+}
+
+#[repr(align(32))]
+union ReprAlign {
+    a: i32,
+    b: u32,
+}
+
+union SingleZST {
+    f0: (),
+}
+union ZSTsAndField1 {
+    f0: u32,
+    f1: (),
+    f2: (),
+    f3: (),
+}
+union ZSTsAndField2 {
+    f0: (),
+    f1: (),
+    f2: u32,
+    f3: (),
+}
+union ZSTAndTwoFields {
+    f0: u32,
+    f1: u64,
+    f2: (),
+}
+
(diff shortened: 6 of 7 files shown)


camsteffen:
I think if we have a lint that unilaterally enforces `repr(C)` for unions (or other types), it would have to be restriction.

jubnzv:
Well, this lint is implemented as suggested in #8235. It enforces the user to explicitly set any `#[repr]` attribute for the unions, to make sure he understands their memory layout.

I personally don't like in this check that the user will have to set `#[allow(unspecified_layout_union)]` on each union which have the default layout, because it is not possible to explicitly set `#[repr(Rust)` in the code. But how can it be implemented better?

camsteffen:
> I personally don't like in this check that the user will have to set #[allow(unspecified_layout_union)] on each union which have the default layout

Agreed.

> But how can it be implemented better?

Other than making it a restriction lint, I don't know.

Do you have a particular motivation for adding this lint? Maybe that would help us decide how the lint should work. Or, if you're just looking for an issue to get started with contributing to Clippy, maybe I could suggest a different issue to start with?

jubnzv:
> Or, if you're just looking for an issue to get started with contributing to Clippy, maybe I could suggest a different issue to start with?

Yes, I'm just looking for tasks to start contributing to Clippy. I would really appreciate, if you can advice some issues.
Maybe there is something that would be most useful for the project right now?

> Do you have a particular motivation for adding this lint? Maybe that would help us decide how the lint should work.

I think, this lint could be useful in two kinds of projects:
* Embedded systems, in particular, the implementation of various communication protocols. If they use structures to describe packages, that are transmitted in the binary format, we always need to know their exact layout.
* Something with FFI and interaction with C code.

Most likely, in other projects, this lint will almost always be globally suppressed.

I haven't written such things on Rust yet, so this is just my guess. Probably,  @5225225 can provide a better rationale for this.

5225225:
My reasoning behind it is I see people assuming that unions are an okay way to transmute between two types, even without a `#[repr(C)]` annotation.

Which is wrong, you must never read from a variant that hasn't been written to at least once (and if your type has validity invariants, this statement might be stronger, you must never read from a variant that isn't the *most recently* written one, if the two variants overlap).

Though now that I look at it, simply warning on *all* unions that have no `#[repr(C)]` might be wrong (Since that would call MaybeUninit and other unions like it wrong).

What might be a more useful lint is to look for all unions that have at least 2 non-ZST variants. Since that would allow the definition of MaybeUninit, but *not* allow

```rust
union FloatOrInt {
    a: f64,
    b: u64,
}
```

(which is not correct)

One bug that this lint *would* have caught is https://github.com/ImageOptim/libimagequant/blob/e99d92ad9abc08c8c912f8d3dc19bc054afb468e/src/hist.rs#L78-L82 , where the union here *must* be laid out overlapping, and if you don't, it's UB.

camsteffen:
I'm generally wary of lints along the lints of "there's nothing wrong with this code, but did you realize...". I just don't think that's what lints are for. Lints should detect code that is deterministically wrong or sub-optimal.

> What might be a more useful lint is to look for all unions that have at least 2 non-ZST variants.

I think you mean "fields" instead of "variants"? This would eliminate some cases from the lint, but largely the problem still remains.

> Yes, I'm just looking for tasks to start contributing to Clippy. I would really appreciate, if you can advice some issues.
> Maybe there is something that would be more useful for the project right now?

In general the ones labelled good-first-issue are good (but I know you already tried that 😄). Some specific ones that look good to me: #8282, #8212, #8215.

Maybe others will have a different view. @rust-lang/clippy?

5225225:
> I'm generally wary of lints along the lines of "there's nothing wrong with this code, but did you realize...". I just don't think that's what lints are for. Lints should detect code that is deterministically wrong or sub-optimal.

Well, there *could* be something wrong with the code. It depends how the union is used.

In the PR above for libimagequant, there was *definitely* a bug there. I'm not sure clippy is powerful enough to determine stuff like "Oh, you wrote to this field of the union, and then you read from a different one, that's UB" (especially if it was cross function, as it might be).

It could *maybe* detect union fields that are never written to/have a pointer taken, but *are* read from? That might catch some cases, but feels like it might have too many false negatives.

And in any case, there's plenty of existing lints for "there *might* not be something wrong with this code, but there *probably* is something wrong here.", as I understand it, that's what `suspicious` is for. Though this lint might belong in `pedantic`.

Searching down the list of non-repr(C) unions found by doing https://github.com/search?o=desc&p=1&q=%22pub+union%22+language%3Arust&s=indexed&type=Code , it does find a few cases that are definitely unsound, and some that are *probably* fine

Non-repr(C) unions that have multiple non-ZST fields are rare, but I found these 4 cases

1. https://github.com/Pathal/mips_rust_int/blob/e90d86d3979c83a2907bd993954989388678592a/src/register.rs#L267
2. https://github.com/unic0rn9k/slas/blob/006d2184dcd2b05330e8872358f5c221c411c92e/src/lib.rs#L185
3. https://github.com/dgoodlad/crabdac/blob/bd51ea819e8b163cc6e4efff223907598bb93cc6/stm32f4/src/stm32f413/mod.rs#L111
4. https://github.com/mathias234/gray/blob/0db68d62288a5ef087eced535ff504cd9ed53425/crates/gray/src/compiler/compiler.rs#L41

`1` is definitely unsound, simply create a register with the `int` field init, and then call `.get_f32()`
`2` looks sound, it's externally tagged as to what variant is active. So this would be a false positive.
`3`.... looks unsound but not for this reason, they're using a non-`#[repr(C)]` thing in what looks to be FFI. So.... half positive? We're right to lint here, but for the wrong reasons.
`4` looks like a similar case to `2`, externally tagged.

So 1.5/4 on being correct, but also this lint really wouldn't fire that often. 2.5/5 if we count the issue I fixed.

jubnzv:
> What might be a more useful lint is to look for all unions that have at least 2 non-ZST variants.

I agree with this point. Unions with a single non-ZST field [have layout the same as the layout of that field](https://rust-lang.github.io/unsafe-code-guidelines/layout/unions.html#layout-of-unions-with-a-single-non-zero-sized-field). I updated PR to make the lint only checks unions with at least two non-ZST fields.

I also improved the lint to make it warn about any `repr` other than `C`, because we don't actually know the layout of the union, if it uses the default (`#[repr(Rust)]`) layout.

> In general the ones labelled good-first-issue are good (but I know you already tried that smile). Some specific ones that look good to me:

Thank you! I managed to localize #8282 in the Clippy code. It takes some time to fix it, because I'm new to the `rustc` internals.

5225225:
By the way, the unsafe code guidelines is *not* what is promised.

It's what they think *should* be promised.

It says this both at the top of that page, and in the introduction.

It is currently legal for a compiler to add padding into a single field type of *any* kind, including unions.

5225225:
I'd also add some tests for repr(transparent) unions

granted, you can't have a repr transparent union with multiple non-ZST fields, but we shouldn't lint if they do that.

flip1995:
> Maybe others will have a different view.

I didn't read the whole thread, but I see parallels with the recently discussed [`non_send_fields_in_send_ty`](https://rust-lang.github.io/rust-clippy/master/index.html#non_send_fields_in_send_ty) lint. You have to write `unsafe` code in order to do the thing the lint complains about.

In order to get the UB described in the lint description, you have to read from the union. And in order to do that you have to write `unsafe` code, which already is a barrier.

So all this lint would do is pointing to the [documentation of unions](https://doc.rust-lang.org/reference/items/unions.html#reading-and-writing-union-fields) in Rust. I agree with @camsteffen that this is not really the point of lints. If we would add this, it would have to be `restriction`, because having a `union` without a `repr` attr is totally fine by itself, which means that this lint would _restrict_ the language.

camsteffen:
> which means that this lint would _restrict_ the language.

Oh! 🤯

5225225:
I disagree with putting this lint in `restriction`, seeing as it *does* find soundness bugs.

And you probably wouldn't have written the bugs if you *knew* to look out for them, and in order to look out for them using this lint if it was in `restriction`, you'd have to explicitly name this lint. (As in, the only case the lint would be useful is if you're in a `union` heavy crate and you *know* you messed this up in the past, but even then, simply being aware of it makes it pretty easy to spot)

After all, other lints for things that are *often* perfectly sound but *might* indicate a bug exist in `pedantic`, like `cast_ptr_alignment`

camsteffen:
> I disagree with putting this lint in `restriction`, seeing as it _does_ find soundness bugs.

Soundness bugs are indeed serious bugs. But severity is not enough. It has to be a *likely* bug in order to be `pedantic` or `suspicious`. To say that `union` by itself is "likely a bug" is a far reach.

We get a lot of opportunities to compromise on "drawing the line" like this. If we don't take a hard stance, Clippy gets very noisy very quickly.

> And you probably wouldn't have written the bugs if you _knew_ to look out for them

This is what documentations is for, not lints. Lints are for telling you when you've already made a detectable mistake. And like flip1995 said the coder is already given a warning with unsafe blocks.

> After all, other lints for things that are _often_ perfectly sound but _might_ indicate a bug exist in `pedantic`, like `cast_ptr_alignment`

That lint detects a code pattern that is suspicious in its own right. But if the lint isn't correct *at least* for a majority of cases (preferably a very large majority) in practice, then it shouldn't be `pedantic` or `suspicious`.

5225225:
Okay, suggestion:

We land this lint as-is, in `restriction`. Still only warning on more likely to be incorrect usages (not linting for literally "you used a union", but a union with multiple non-ZST fields which aren't forced to overlap by a repr(C))

And then a new lint is made, with correctness level, looking for non-repr C unions that have a field being read from, that *definitely* couldn't have been init at that point in the code. Probably leaving it open ended as to exactly what it lints on, but only linting on stuff that's definitely UB (depending on layout)

(Say, you make a union specifying field `a`, you pass a pointer to field `b` to some FFI function, and then you read field `c`, all in the same fn. Or you have a union that had any field that is not public, is *never* written to/has an offset taken/has a pointer taken, but *is* read from)

---

And in the mean time, I go fix the `union` docs warning people about this, because their *example code* has UB, and is doing the exact thing this lint complains about.

flip1995:
> I disagree with putting this lint in `restriction`, seeing as it _does_ find soundness bugs.

My issue with this lint is, that it doesn't find unsoundness bugs, but rather tells you that there might be or might not be unsoundness in your code. It doesn't even tell you where there might be unsoundness, it just spells out the documentation for you again and sends you off on a journey to find the unsoundness yourself.

> And then a new lint is made, with correctness level, looking for non-repr C unions that have a field being read from, that _definitely_ couldn't have been init at that point in the code. Probably leaving it open ended as to exactly what it lints on, but only linting on stuff that's definitely UB (depending on layout)

That would also be my suggestion. We shouldn't lint on the union type, but on the union read. Only linting on something that is "definitely UB" might be very hard or even impossible (if it were possible, shouldn't the compiler error on this?). But I can see a lint in `pedantic` that lints _all_ reads of non-`repr(C)` unions.

jubnzv:
> I'd also add some tests for repr(transparent) unions (both with / without ZST fields, as well as multiple non-ZSTs) 
> granted, you can't have a repr transparent union with multiple non-ZST fields, but we shouldn't lint if they do that.

Thank you, that's a good idea. I added tests for `#[repr(transparent)]`.

> If we would add this, it would have to be restriction, because having a union without a repr attr is totally fine by itself, which means that this lint would restrict the language.

Well, it sounds reasonable, because we don't actually check if we the use of these unions in unsafe blocks. I updated the level of the lint to `restriction`.

camsteffen:
Okay so I think we're on the same page that this lint would have to be restriction. But now I have to ask is this lint worth it as a restriction lint? If the motivation for the lint were "I want to enforce that all unions in my crate have a non-default repr", then this lint would make perfect sense. But rather the motivation seems to be "let's make an attempt to catch unsoundness bugs". And we know that restriction lints are not so good for catching bugs. The former motivation sounds plausible, but I don't want to blindly assume that use case exists.

> The only thing that confuses me, is why the `#[repr(Rust)]` is present in the documentation, but cannot be explicitly specified by the user. Is this an intentional design decision?

That confused me too so it should probably be clarified. :) But yes I'm pretty sure it's intentional - it's like pseudo code for documentation purposes only.

jubnzv:
> But now I have to ask is this lint worth it as a restriction lint? If the motivation for the lint were "I want to enforce that all unions in my crate have a non-default repr", then this lint would make perfect sense.

I see the motivation for this lint as: "I want to ensure that all unions have a _defined_ layout, so I can't get unexpected behavior in unsafe blocks". We have already seen [examples](https://github.com/rust-lang/rust-clippy/pull/8289#issuecomment-1013963336) of potential errors that could be fixed with this lint.

flip1995:
I have a pretty low bar for restriction lints like this. However, the lint documentation should not claim, that it finds unsoundness bugs, but rather that it enforces a `repr` for unions (which side effect may be to fix unsoundness bugs). I think the current documentation conveys this quite well.

If some people want a lint that restricts the language for them, like this lint does AND the lint helps with finding real world problems, I think it is fine to include it in Clippy.

clippy_lints/src/unspecified_layout_union.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// }
+    ///
+    /// fn main() {
+    ///     let _x: u32 = unsafe {
+    ///         Foo { a: 0_i32 }.b // Now defined behaviour, this is just an i32 -> u32 transmute
+    ///     };
+    /// }
+    /// ```
+    #[clippy::version = "1.60.0"]
+    pub UNSPECIFIED_LAYOUT_UNION,
```
camsteffen, line 47: I think a better name would be `default_union_representation` which is consistent with the [docs](https://doc.rust-lang.org/reference/type-layout.html#representations).

clippy_lints/src/unspecified_layout_union.rs
```diff
@@ -0,0 +1,105 @@
...
+use rustc_hir::{self as hir, HirId, Item, ItemKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::layout::LayoutOf;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+use rustc_typeck::hir_ty_to_ty;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Displays a warning when a union is declared without a `#[repr]` attribute.
```
camsteffen, line 11:
```suggestion
    /// Displays a warning when a union is declared with the default representation (without a `repr` attribute).
```

clippy_lints/src/unspecified_layout_union.rs
```diff
@@ -0,0 +1,105 @@
...
+declare_lint_pass!(UnspecifiedLayoutUnion => [UNSPECIFIED_LAYOUT_UNION]);
+
+impl<'tcx> LateLintPass<'tcx> for UnspecifiedLayoutUnion {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
+        if !has_c_repr_attr(cx, item.hir_id()) && is_union_with_two_non_zst_fields(cx, item) {
+            span_lint_and_help(
+                cx,
+                UNSPECIFIED_LAYOUT_UNION,
+                item.span,
+                "union which is not marked with a `#[repr(C)]` attribute",
```
camsteffen, line 60:
```suggestion
                "this union has the default representation",
```

clippy_lints/src/unspecified_layout_union.rs
```diff
@@ -0,0 +1,105 @@
...
+                    cx.tcx.def_path_str(item.def_id.to_def_id())
+                ),
+            );
+        }
+    }
+}
+
+/// Returns true if the given item is a union with at least two non-ZST fields.
+fn is_union_with_two_non_zst_fields(cx: &LateContext<'_>, item: &Item<'_>) -> bool {
+    if let ItemKind::Union(data, _) = &item.kind {
```
camsteffen, line 73: The `kind` check should come before `has_c_repr_attr`.

clippy_lints/src/unspecified_layout_union.rs
```diff
@@ -0,0 +1,105 @@
...
+    }
+    let ty = hir_ty_to_ty(cx.tcx, hir_ty);
+    if let Ok(layout) = cx.layout_of(ty) {
+        layout.is_zst()
+    } else {
+        false
+    }
+}
+
+fn has_c_repr_attr(cx: &LateContext<'_>, hir_id: HirId) -> bool {
```
camsteffen, line 92: I think the lint should allow any `repr`, not just `repr(C)`.

camsteffen (changes requested):
Let's go ahead with this then. Just some minor feedback.

clippy_lints/src/unspecified_layout_union.rs
jubnzv: Fixed

clippy_lints/src/unspecified_layout_union.rs
jubnzv: Even if the union has some alignment attribute (like `align` or `packed`), it still has no any guarantees about the layout, if the default `Rust` representation is used. I think, those cases more likely will contain bugs, because the user may want to set a specific layout using only these attributes.

camsteffen: Gotcha. Let's just add a comment somewhere that C is the only non-default representation for unions.

jubnzv: We have the following description:

> ### What it does
> Displays a warning when a union is declared with the default representation (without a `#[repr(C)]` attribute)

Should we explain this in more detail?

camsteffen: I think it's okay but you're welcome to expand that.

jubnzv: Well, I'm not sure how best to expand this so as not to annoy the user with unnecessary repetitions. I think, it's enough, because if the user wants to restrict the language, they at least understand which layouts are possible in Rust.

camsteffen (approved):
Looks good! Please squash commits.

camsteffen:
@jubnzv and @workingjubilee thank you for your patience as we worked through this.

@bors r+

PR #8292, opened 2022-01-15
issue #8239: Printed hint for lint or_fun_call is cropped and does no…

### LINKED ISSUE #8239: Printed hint for lint `or_fun_call` is cropped and does not show the solution
### Desctiption
Following code snippet was correctly marked by clippy as `or_fun_call`. However it's output could be better. Problem here is at the last line of this block, but clippy's hint is cropped and this bottom part is not shown. It seams that clippy is omitting different lines in the warning and in the hint.

### Code
```Rust
self.frames
    .iter()
    .map(|f| f.encode())
    .reduce(|mut acc, mut f| {
        acc.append(&mut f);
        acc
    })
    .unwrap_or(Vec::new())

```

### Clippy's output
```
warning: use of `unwrap_or` followed by a call to `new`
  --> src/server/frame.rs:89:9
   |
89 | /         self.frames
90 | |             .iter()
91 | |             .map(|f| f.encode())
92 | |             .reduce(|mut acc, mut f| {
...  |
95 | |             })
96 | |             .unwrap_or(Vec::new())
   | |__________________________________^
   |
   = note: `#[warn(clippy::or_fun_call)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
help: try this
   |
89 ~         self.frames
90 +             .iter()
91 +             .map(|f| f.encode())
92 +             .reduce(|mut acc, mut f| {
93 +                 acc.append(&mut f);
94 +                 acc
 ...
```

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

@rustbot label +C-enhancement

marekdownar:
fixes rust-lang/rust-clippy#8239

changelog: [`or_fun_call`]: if suggestion contains more lines than MAX_SUGGESTION_HIGHLIGHT_LINES it is stripped to one line

### DIFF
--- a/clippy_lints/src/methods/or_fun_call.rs
+++ b/clippy_lints/src/methods/or_fun_call.rs
@@ -24,6 +24,7 @@ pub(super) fn check<'tcx>(
     args: &'tcx [hir::Expr<'_>],
 ) {
     /// Checks for `unwrap_or(T::new())` or `unwrap_or(T::default())`.
+    #[allow(clippy::too_many_arguments)]
     fn check_unwrap_or_default(
         cx: &LateContext<'_>,
         name: &str,
@@ -32,6 +33,7 @@ pub(super) fn check<'tcx>(
         arg: &hir::Expr<'_>,
         or_has_args: bool,
         span: Span,
+        method_span: Span,
     ) -> bool {
         let is_default_default = || is_trait_item(cx, fun, sym::Default);
 
@@ -53,21 +55,24 @@ pub(super) fn check<'tcx>(
 
             then {
                 let mut applicability = Applicability::MachineApplicable;
-                let hint = ".unwrap_or_default()";
+                let hint = "unwrap_or_default()";
+                let mut shrink = span;
+
                 let mut sugg: String = format!(
-                    "{}{}",
+                    "{}.{}",
                     snippet_with_applicability(cx, self_expr.span, "..", &mut applicability),
                     hint
                 );
 
                 if sugg.lines().count() > MAX_SUGGESTION_HIGHLIGHT_LINES {
+                    shrink = method_span.with_hi(span.hi());
                     sugg = hint.to_string();
                 }
 
                 span_lint_and_sugg(
                     cx,
                     OR_FUN_CALL,
-                    span,
+                    shrink,
                     &format!("use of `{}` followed by a call to `{}`", name, path),
                     "try this",
                     sugg,
@@ -173,7 +178,7 @@ pub(super) fn check<'tcx>(
         match inner_arg.kind {
             hir::ExprKind::Call(fun, or_args) => {
                 let or_has_args = !or_args.is_empty();
-                if !check_unwrap_or_default(cx, name, fun, self_arg, arg, or_has_args, expr.span) {
+                if !check_unwrap_or_default(cx, name, fun, self_arg, arg, or_has_args, expr.span, method_span) {
                     let fun_span = if or_has_args { None } else { Some(fun.span) };
                     check_general_case(cx, name, method_span, self_arg, arg, expr.span, fun_span);
                 }
--- a/clippy_lints/src/methods/or_fun_call.rs
+++ b/clippy_lints/src/methods/or_fun_call.rs
@@ -4,6 +4,7 @@ use clippy_utils::source::{snippet, snippet_with_applicability, snippet_with_mac
 use clippy_utils::ty::{implements_trait, match_type};
 use clippy_utils::{contains_return, is_trait_item, last_path_segment, paths};
 use if_chain::if_chain;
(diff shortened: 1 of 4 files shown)


clippy_lints/src/methods/or_fun_call.rs
```diff
@@ -52,16 +53,24 @@ pub(super) fn check<'tcx>(
...
                     cx,
                     OR_FUN_CALL,
                     span,
                     &format!("use of `{}` followed by a call to `{}`", name, path),
                     "try this",
-                    format!(
-                        "{}.unwrap_or_default()",
-                        snippet_with_applicability(cx, self_expr.span, "..", &mut applicability)
-                    ),
+                    sugg,
```
xFrednet, line 73: If you change the suggestion to only display `.unwrap_or_default()` then the span in line 70 also has to change to only replace the previous function call. This would currently replace the entire expression with just the string `.unwrap_or_default()`.

I would recomment writing a test first and then checking the output. :upside_down_face: The test is also needed to ensure that this doesn't happen again :)

llogiq:
As @xFrednet wrote, I would expect at least a test to ensure correct behavior.

marekdownar:
@xFrednet 
I will try to provide fix it in about 2 hours with suitable test case.

xFrednet:
Thank you, feel free to reach out if you need help :upside_down_face:

tests/ui/or_fun_call.fixed
```diff
@@ -176,4 +176,52 @@ mod issue6675 {
...
+        let iter = frames.iter();
+        iter.map(|f: &String| f.to_lowercase())
+            .reduce(|mut acc, f| {
+                let _ = "";
+                acc.push_str(&f);
+                acc
+            }).unwrap_or_default();
+    }
+
+    unsafe fn less_than_max_suggestion_highest_lines() {
```
xFrednet, line 216: Is there a reason why all of these test functions are unsafe? :) There is nothing wrong with it, I'm just curious.

clippy_lints/src/methods/or_fun_call.rs
```diff
@@ -52,16 +55,27 @@ pub(super) fn check<'tcx>(
 
             then {
                 let mut applicability = Applicability::MachineApplicable;
+                let hint = "unwrap_or_default()";
+                let mut shrink = span;
```
xFrednet, line 59: This is really nit-picky, and I'm fine with keeping it like this. But you could give this span a better name, like: `sugg_span`.

xFrednet (approved):
This version looks good to me, one optional NIT and one question.

BTW, all your commits are marked as *Unverified*. This is not a problem, but I wanted to mention it in case this was unintended :upside_down_face:

tests/ui/or_fun_call.fixed
marekdownar: > This version looks good to me, one optional NIT and one question.
> 
> BTW, all your commits are marked as _Unverified_. This is not a problem, but I wanted to mention it in case this was unintended 🙃

Should I upload my public GPG key to GitHub to verify my commits ?

marekdownar: > Is there a reason why all of these test functions are unsafe? :) There is nothing wrong with it, I'm just curious.

I made them unsafe without any reason - will fix it in next commit.

clippy_lints/src/methods/or_fun_call.rs
marekdownar: > This is really nit-picky, and I'm fine with keeping it like this. But you could give this span a better name, like: `sugg_span`.

Ok, next commit will change it.

tests/ui/or_fun_call.fixed
xFrednet: > Should I upload my public GPG key to GitHub to verify my commits ?

Either is fine for us. We can merge this as it is. I just like to point these things out :upside_down_face:

marekdownar: > > Should I upload my public GPG key to GitHub to verify my commits ?
> 
> Either is fine for us. We can merge this as it is. I just like to point these things out 🙃

Ok uploaded public GPG key but still Unverified in here. Probably next PR will be Verified :) Will C :)

xFrednet: Alright :+1:

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the changes! :+1: 

FYI @marekdownar: I've updated the lint description, as the changelog entry has to be on the same line for the pipeline to pass.

I hope @llogiq will be fine with me *stealing* this r+

@bors r+

PR #8295, opened 2022-01-15
Handle implicit named arguments in `useless_format`

### LINKED ISSUE #8290: useless_format does not play well with format arg capture
### Summary

Clippy does not yet understand fully understand `format!("{x}");` yet.
This probably needs to be a `msrv`lint. 

### Reproducer

I tried this code:

```rust
fn main() {
    let x = "hello";
    let _string = format!("{x}");
}
```

Clippy should suggest something like `x.to_string()` but instead suggests `{x}.to_string()`

Instead, this happened:
````
warning: useless use of `format!`
 --> src/main.rs:3:19
  |
3 |     let _string = format!("{x}");
  |                   ^^^^^^^^^^^^^^ help: consider using `.to_string()`: `{x}.to_string()`
  |
  = note: `#[warn(clippy::useless_format)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_format
````

### Version

```text
`clippy 0.1.60 (38c22af 2022-01-15)`
```

### Additional Labels

_No response_

Jarcho:
fixes #8290

Ideally this would fix the macro parsing code to handle this, but this is a smaller change and easier to back port.

changelog: Handle implicit named arguments in `useless_format`

### DIFF
--- a/tests/ui/format.rs
+++ b/tests/ui/format.rs
@@ -75,4 +75,10 @@ fn main() {
     let _s: String = format!("{}", &*v.join("\n"));
 
     format!("prepend {:+}", "s");
+
+    // Issue #8290
+    let x = "foo";
+    let _ = format!("{x}");
+    let _ = format!("{x:?}"); // Don't lint on debug
+    let _ = format!("{y}", y = x);
 }
--- a/clippy_lints/src/format.rs
+++ b/clippy_lints/src/format.rs
@@ -9,7 +9,7 @@ use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::symbol::kw;
-use rustc_span::{sym, Span};
+use rustc_span::{sym, BytePos, Span};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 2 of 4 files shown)


tests/ui/format.rs
```diff
@@ -75,4 +75,8 @@ fn main() {
     let _s: String = format!("{}", &*v.join("\n"));
 
     format!("prepend {:+}", "s");
+
+    // Issue #8290
+    let x = "foo";
+    let _ = format!("{x}");
```
xFrednet, line 81: Could you also add a test were the value is format and where a named argument is used, like this :upside_down_face::

```rs
// Formatted (Ignore)
let y = 10;
format!("{y:+}");

// named argument
format!("{x}", x=y)

giraffate (changes requested):
Can you add the `msrv`? The named arguments was added in 1.58

camsteffen:
I don't think msrv is needed since we aren't making a code suggestion with the new feature.

I wonder if we could use `path_to_local` and `hir().name(id)`. But this seems to work well enough.

giraffate:
> Can you add the `msrv`? The named arguments was added in 1.58

Oh, is there a compile error in the first place?

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8297, opened 2022-01-16
Don't lint `if_same_then_else` with `if let` conditions

### LINKED ISSUE #7579: FP in if_same_then_else for if-let
followup to #7383 and pinging @xFrednet

Lint name: `if_same_then_else`

I tried this code:

```rust
#[allow(unused_variables)]
fn main() {
    let a = true;
    let b = Some(true);
    
    if let Some(b) = b {
        println!("hello");
    }
    else if a {
        println!("hello");
    }
}
```
see https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=c49e249051ac25681b4fe4522994a524

no lint warning, since I cannot express this until *if-let-chains* are supported

Instead, this happened: 
```sh
error: this `if` has identical blocks
  --> src/main.rs:6:24
   |
6  |       if let Some(b) = b {
   |  ________________________^
7  | |         println!("hello");
8  | |     }
   | |_____^
   |
   = note: `#[deny(clippy::if_same_then_else)]` on by default
note: same as this
  --> src/main.rs:9:15
   |
9  |       else if a {
   |  _______________^
10 | |         println!("hello");
11 | |     }
   | |_____^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else

error: could not compile `playground` due to previous error
```

### Meta

- `cargo clippy -V`: clippy 0.1.56 (0035d9dc 2021-08-16)
- `rustc -Vv`:
```
binary: rustc
commit-date: 2021-07-26
host: x86_64-apple-darwin
release: 1.54.0
LLVM version: 12.0.1
```

Jarcho:
fixes #7579 

changelog: Don't lint `if_same_then_else` with `if let` conditions

### DIFF
--- a/clippy_lints/src/copies.rs
+++ b/clippy_lints/src/copies.rs
@@ -183,7 +183,7 @@ impl<'tcx> LateLintPass<'tcx> for CopyAndPaste {
                 lint_same_cond(cx, &conds);
                 lint_same_fns_in_if_cond(cx, &conds);
                 // Block duplication
-                lint_same_then_else(cx, &blocks, conds.len() == blocks.len(), expr);
+                lint_same_then_else(cx, &conds, &blocks, conds.len() == blocks.len(), expr);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Jarcho:
Should be fixed now. Didn't account for the number of conditions and the number of blocks being different at first.

PR #8298, opened 2022-01-16
fix op_ref false positive

### LINKED ISSUE #7572: op_ref: Following lint leads to unconditional recursion
Lint name: op_ref

I tried this code:

```rust
use std::ops::Mul;

#[derive(Clone, Copy)]
struct Vec3D {
    pub x: f64,
    pub y: f64,
    pub z: f64,
}

impl Mul<&Vec3D> for Vec3D {
    type Output = f64;

    fn mul(self, rhs: &Vec3D) -> Self::Output {
        self.x * rhs.x + self.y * rhs.y + self.z * rhs.z
    }
}

impl Mul<Vec3D> for Vec3D {
    type Output = f64;

    fn mul(self, rhs: Vec3D) -> Self::Output {
        self * &rhs
    }
}

fn main() {
    let a = Vec3D{x: 1.0, y: 2.0, z: 3.0};
    let b = &Vec3D{x: 1.0, y: 2.0, z: 3.0};
    println!("{}", a * b);
    println!("{}", a * a);
}
```
The lint suggests removing the ampersand in the `Mul<Vec3D> for Vec3D` implementation. Doing that, however, leads to a warning about unconditional recursion, and, naturally, it causes the program to not function correctly. 

### Meta

- `cargo clippy -V`: clippy 0.1.56 (2d2bc94c 2021-08-15)
- `rustc -Vv`:
```
rustc 1.56.0-nightly (2d2bc94c8 2021-08-15)
binary: rustc
commit-hash: 2d2bc94c8c3aa778e191f80261c726e4777439f1
commit-date: 2021-08-15
host: x86_64-pc-windows-msvc
release: 1.56.0-nightly
LLVM version: 12.0.1
  ```

ebobrow:
fixes #7572 

changelog: `op_ref` don't lint for unnecessary reference in BinOp impl if removing the reference will lead to unconditional recursion

### DIFF
--- a/clippy_lints/src/eq_op.rs
+++ b/clippy_lints/src/eq_op.rs
@@ -1,12 +1,16 @@
 use clippy_utils::diagnostics::{multispan_sugg, span_lint, span_lint_and_then};
+use clippy_utils::get_enclosing_block;
 use clippy_utils::macros::{find_assert_eq_args, first_node_macro_backtrace};
 use clippy_utils::source::snippet;
 use clippy_utils::ty::{implements_trait, is_copy};
 use clippy_utils::{ast_utils::is_useless_with_eq_exprs, eq_expr_value, is_in_test_function};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::{BinOpKind, BorrowKind, Expr, ExprKind};
+use rustc_hir::{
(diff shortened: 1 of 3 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8304, opened 2022-01-17
Remove __CLIPPY_INTERNAL_TESTS env var

camsteffen:
changelog: none

It doesn't have any effect AFAICT.


flip1995:
I thought this was used to force-enable clippy in UI tests. But if that isn't necessary (anymore), I think we can just remove this.

camsteffen:
I guess that was true at one time. I don't know what changed.

flip1995:
I mean that is a relic from the past, that was already there when I joined the Clippy team, so :shrug: 

@bors r+

PR #8305, opened 2022-01-17
Factor out several utils, add `path_def_id`

camsteffen:
changelog: none

This is generally an effort to reduce the total number of utils. `path_def_id` is added which I believe is more "cross-cutting" and also complements `path_to_local`. Best reviewed one commit at a time.

Added:
* `path_def_id`
* `path_res`

Removed:
 * `is_qpath_def_path`
 * `match_any_diagnostic_items`
 * `expr_path_res`
 * `single_segment_path`
 * `differing_macro_contexts`
 * `is_ty_param_lang_item`
 * `is_ty_param_diagnostic_item`
 * `get_qpath_generics`

Renamed:
* `path_to_res` to `def_path_res`
* `get_qpath_generic_tys` to `qpath_generic_tys`

CC @Jarcho since this relates to some of your work and you may have input.

### DIFF
--- a/clippy_lints/src/blocks_in_if_conditions.rs
+++ b/clippy_lints/src/blocks_in_if_conditions.rs
@@ -1,8 +1,8 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg};
+use clippy_utils::get_parent_expr;
 use clippy_utils::higher;
 use clippy_utils::source::snippet_block_with_applicability;
 use clippy_utils::ty::implements_trait;
-use clippy_utils::{differing_macro_contexts, get_parent_expr};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, Visitor};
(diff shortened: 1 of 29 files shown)


Jarcho:
The only thing I would question here is how much utility `path_def_id` really gives over just having `path_res`. It's only slightly shorter and does the obvious thing (it's 10 characters shorter).

Other than that I think this is good. It's an improvement over what currently exists and no matter what changes happen afterwards related to identifying items, paths will need to be resolved. Having all path resolution go through the same place will probably make future changes easier.

camsteffen:
Yeah `path_res(expr).opt_def_id()` isn't bad. I was thinking that `path_def_id` is so widely applicable that it's worth it but now I'm not so sure.

Also is the distinction between `def_path_res` and `path_res` good enough? `path_res` should be much more widely used so the short name is nice. But it could be `path_node_res`...liking that as I type it.

Jarcho:
`def_path_res` should be rare enough that it isn't that big of a deal. Could name it `resolve_def_path` as it's actually resolving the path and not just looking up what the given path resolves to.

`path_node_res` sounds fine.

camsteffen:
Rebased. I searched the code for `opt_def_id()` and we really do that a lot so I think `path_def_id` is worth it (again). I also figured adding `_node` doesn't ultimately add much clarity. 🤷

flip1995 (approved):
This looks like a great simplification of our utils! 

Regarding the naming, I'm fine with how it is now. But I'll leave that decision up to you @Jarcho & @camsteffen.

r=me if you settled on the naming of the functions.

Jarcho:
I think the name is fine as is. One small quip here is it doesn't actually take a `Path` or `QPath` despite what the name might lead one to believe. Not really an issue though.

PR #8315, opened 2022-01-20
`trait_duplication_in_bounds` checks path segments for trait items

### LINKED ISSUE #8306: Misleading error message on traits after #8229
### Summary

I have a simple trait that extends the iterators (just like `itertools` does). This trait has `Iterator` bound, but my function also has `Iterator` in its `Self` argument (just a more precise one).

Before #8229 it was fine, now it hits `trait_duplication_in_bounds` (which is great!) but the error message is misleading and points to the function's `where`-clause, however removing that bound does not solve this issue.

### Reproducer

I tried this code:

```rust
pub trait Iter: Iterator {
    fn into_group_btreemap<K, V>(self) -> BTreeMap<K, Vec<V>>
    where
        Self: Iterator<Item = (K, V)> + Sized,
        K: Ord + Eq,
    {
        into_group_btreemap(self)
    }
}
```

The lint should show that the trait bound is superfluous:
```
3 |     pub trait Iter: Iterator {
  |                     ^^^^^^^^
```

Instead, this happened:
The lint shows the real bound:
```
6 |         Self: Iterator<Item = (K, V)> + Sized,
  |               ^^^^^^^^^^^^^^^^^^^^^^^
```

### Version

```text
rustc 1.60.0-nightly (bd3cb5256 2022-01-16)
binary: rustc
commit-hash: bd3cb52565faab2755ff1bdb54d88bc91f47b4b9
commit-date: 2022-01-16
release: 1.60.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

dswij:
closes #8306 

changelog: [`trait_duplication_in_bounds`] Fix FP when path segments exists for trait items

### DIFF
--- a/clippy_lints/src/trait_bounds.rs
+++ b/clippy_lints/src/trait_bounds.rs
@@ -3,7 +3,7 @@ use clippy_utils::source::{snippet, snippet_with_applicability};
 use clippy_utils::{SpanlessEq, SpanlessHash};
 use core::hash::{Hash, Hasher};
 use if_chain::if_chain;
-use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::unhash::UnhashMap;
 use rustc_errors::Applicability;
 use rustc_hir::def::Res;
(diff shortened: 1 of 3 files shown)


giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #8322, opened 2022-01-20
single_match: Don't lint non-exhaustive matches; support tuples

### LINKED ISSUE #8282: `single_match` should not lint exhaustive `match`

jubnzv:
`single_match` lint:
* Don't lint exhaustive enum patterns without a wild.
  Rationale: The definition of the enum could be changed, so the user can get non-exhaustive match after applying the suggested lint (see https://github.com/rust-lang/rust-clippy/issues/8282#issuecomment-1013566068 for context).
* Lint `match` constructions with tuples (as suggested at https://github.com/rust-lang/rust-clippy/issues/8282#issuecomment-1015621148)

Closes #8282

---

changelog: [`single_match`]: Don't lint exhaustive enum patterns without a wild.
changelog: [`single_match`]: Lint `match` constructions with tuples


jubnzv:
This is not finished yet, because I need some time to restore logic that calls `match_types` in the condition. This is required to handle cases, when the user has aliases named as standard structures (`Err`, `Ok`, etc.), I'd like to write tests for this.

I'm publishing this PR now to discuss overall design of changes, because there was some discussion in #8282. I'm open to ideas if there are better suggestions to implementing this.

clippy_lints/src/matches.rs
```diff
@@ -849,25 +848,135 @@ fn check_single_match_opt_like(
...
+    match pat.kind {
+        PatKind::Wild => true,
+        PatKind::Tuple(inner, _) => {
+            for p in inner {
+                if !collect_pat_paths(acc, p) {
+                    return false;
+                }
             }
-            rustc_hir_pretty::to_string(rustc_hir_pretty::NO_ANN, |s| s.print_qpath(path, false))
+            true
```
llogiq, line 886: How about `inner.iter().all(|p| collect_pat_paths(acc, p))`?

clippy_lints/src/matches.rs
```diff
@@ -849,25 +848,135 @@ fn check_single_match_opt_like(
...
+/// match x {
+///     (Some(_), _) => todo!(),
+///     (None, _) => {}
+/// }
+/// ```
+fn form_exhaustive_tuples(left: &Pat<'_>, right: &Pat<'_>) -> bool {
+    match (&left.kind, &right.kind) {
+        (PatKind::Wild, _) | (_, PatKind::Wild) => true,
+        (PatKind::Tuple(left_in, left_pos), PatKind::Tuple(right_in, right_pos)) => {
+            // We don't actually know the position and presence of the `..` (dotdot) operator in
```
llogiq, line 939: This puzzles me. Why don't we know at which argument #no is the `..`? And why do we need the span for this? It seems like a very roundabout way to implement this.

tests/ui/single_match.rs
```diff
@@ -145,6 +145,56 @@ fn if_suggestion() {
...
+        (_, _) => {},
+    }
+
+    // lint
+    match (Some(42), Some(E::V), Some(42)) {
+        (.., Some(E::V), _) => {},
+        (..) => {},
+    }
+
+    // don't lint
```
llogiq, line 179: I'd like to see a short explanation why this shouldn't lint.

tests/ui/single_match.rs
```diff
@@ -145,6 +145,56 @@ fn if_suggestion() {
...
+    match (Some(E::V), Some(E::V), Some(E::V)) {
+        (.., Some(E::V), _) => {},
+        (.., None, _) => {},
+    }
+
+    // don't lint
```
llogiq, line 185: Same here.

tests/ui/single_match.rs
```diff
@@ -145,6 +145,56 @@ fn if_suggestion() {
...
+    match (Some(E::V), Some(E::V), Some(E::V)) {
+        (Some(E::V), ..) => {},
+        (None, ..) => {},
+    }
+
+    // don't lint
```
llogiq, line 191: And here.

llogiq (commented):
I commented where I found shorter ways or was befuddled with the implementation. In general, this should be mostly merge-worthy already apart from the nitpick comments I added.

clippy_lints/src/matches.rs
jubnzv: The point is that we don't know the actual number of the patterns replaced by `..`. 

For example:

```rust
match (Some(E::V), Some(E::V), Some(E::V), Some(E::V)) {
  (Some(E::V), .., None) => todo!(), // `..` replaces [1, 2]
  (.., None) => todo!(), // `..` replaces [0, 1, 2]
}
```

In this example we want to iterate to the both arms at the same time, to make sure that the elements with the same index makes the exhaustive match. So, the logic, implemented here basically evaluates the maximum possible length of the patterns and traverses both arms, considering entries in `..` as wildcards (`_`).

tests/ui/single_match.rs
jubnzv: I updated the description in 87e00c9. That's what the #8282 was about.

clippy_lints/src/matches.rs
jubnzv: Thank you!

clippy_lints/src/matches.rs
llogiq: In that case shouldn't, `position`- and `rposition`ing the `..` pattern in the subpatterns be enough? Also "span" has a different meaning in most lints, so the naming could be improved (if needed at all).

camsteffen:
Just a note that the term "exhaustive match" doesn't really make sense like I thought it did when I posted the issue. *All* `match`es are exhaustive by necessity. The problem is more specific to enums, like "exhaustive enum patterns without a wild".

clippy_lints/src/matches.rs
jubnzv: I fixed the naming to make it differ from `rustc`'s `Span`s.
I'm not sure if we want to remove these variables at all, because this would complicate the loop in places when we need to evaluate the relative offset from the current pattern to the `..` position.

llogiq:
Thank you! @bors r+

xFrednet:
The changelog entries were missing in the PR description, I added them now :upside_down_face: 

@bors retry

camelid:
Btw, does this work with tuple structs (`S(_, _)`) too?

jubnzv:
> Btw, does this work with tuple structs (S(_, _)) too?

No, this PR doesn't implement any additional logic with structs.

I think, we could process them just like tuples, i.e.:

```rust
// lint: S(_, _) forms exhaustive match, so it could be removed
match s {
  S(42, a) => dummy(a),
  S(_, _) => {},
}

// lint: S(..) forms exhaustive match, so it could be removed
match s {
  S(42, a) => dummy(a),
  S(..) => {},
}
```

Am I right?

jubnzv:
I added support for tuple structs in #8395.

PR #8329, opened 2022-01-21
Don't suggest an empty variant name in `enum_variant_names`

### LINKED ISSUE #8324: `enum_variant_names` warns even if removing the suffix would leave an empty string
### Summary

`enum_variant_names` sometimes issues warnings that don't make sense and cause an error.

### Lint Name

enum_variant_names

### Reproducer

I tried this code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6c5a826b31f5f8c74ca7991ac042002a

```rust
#[allow(dead_code)]
enum Phase {
    PreLookup,
    Lookup,
    PostLookup,
}
```

I saw this happen:

```
warning: all variants have the same postfix: `Lookup`
 --> src/lib.rs:1:1
  |
1 | / enum Phase {
2 | |     PreLookup,
3 | |     Lookup,
4 | |     PostLookup,
5 | | }
  | |_^
  |
  = note: `#[warn(clippy::enum_variant_names)]` on by default
  = help: remove the postfixes and use full paths to the variants instead of glob imports
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names
```

I expected to see this happen: No warning, because removing the `Lookup` postfix would leave one of the enum variants as an empty string.

### Version

```text
rustc 1.60.0-nightly (1bd4fdc94 2022-01-12)
clippy 0.1.59 (1bd4fdc 2022-01-12)
```

### Additional Labels

Alexendoo:
changelog: false positive fix: [`enum_variant_names`]: No longer suggests an empty variant name

Fixes #8324

### DIFF
--- a/clippy_lints/src/enum_variants.rs
+++ b/clippy_lints/src/enum_variants.rs
@@ -172,6 +172,9 @@ fn check_variant(cx: &LateContext<'_>, threshold: u64, def: &EnumDef<'_>, item_n
         let name = var.ident.name.as_str();
 
         let variant_split = camel_case_split(name);
+        if variant_split.len() == 1 {
+            return;
+        }
 
         pre = pre
             .iter()
--- a/tests/ui/enum_variants.rs
+++ b/tests/ui/enum_variants.rs
@@ -151,4 +151,11 @@ enum North {
     NoRight,
 }
 
+// #8324
+enum Phase {
+    PreLookup,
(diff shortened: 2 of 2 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8330, opened 2022-01-21
Update Changelog

flip1995:
[Rendered](https://github.com/flip1995/rust-clippy/blob/changelog/CHANGELOG.md)

changelog: none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,184 @@ document.
 
 ## Unreleased / In Rust Nightly
 
-[e181011...master](https://github.com/rust-lang/rust-clippy/compare/e181011...master)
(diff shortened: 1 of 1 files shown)


llogiq:
Thank you! This is going to be a good release. Should we add notes for the 1.58.1 release, too?

matthiaskrgr:
Just a side note: I wonder if we should also inform about the lints default severity (deny, warn, allow..) so that people can see at a glance "ah, this will lint by default on my crate" or "I'll have to enable this one manually if I'm interested".

flip1995:
I added the changes of `1.58.1` as a subsection of `1.58`. 

I also added comments about the updated lint severity.

llogiq:
Thank you! @bors r+

PR #8337, opened 2022-01-23
Add new lint manual_memmove

### LINKED ISSUE #8335: New lint: manual_memmove
### What it does

Like manual_memcpy, find for-loops equivalent to `slice::copy_within`.

### Lint Name

manual_memmove

### Category

perf

### Advantage

It's shorter and faster.

### Drawbacks

any case where memmove is not actually faster? (should probably file as bug against rust though)

### Example

```rust
let mut arr: [u8; 4] = [1, 2, 3, 4];
                                     
for i in 0..2 {
    arr[i] = arr[i + 2];
}

```

Could be written as:

```rust
arr.copy_within(2..4, 0);
```

untitaker:
Add a very basic lint that checks against unnecessary reimplementations of `copy_within`. The lint reuses some now-public parts of the memcpy lint, but supports almost none of the fanciness that the memcpy lint has. The main reason being that overlapping indices/ranges change behavior in memmove, but do not in memcpy (because src is immutable). The secondary reason being that this is my first lint.

Fixes #8335 

----

changelog: New lint [`manual_memmove`]

### DIFF
--- a/clippy_lints/src/loops/manual_memmove.rs
+++ b/clippy_lints/src/loops/manual_memmove.rs
@@ -1,0 +1,233 @@
+use super::MANUAL_MEMCPY;
+use crate::loops::manual_memcpy::{
+    apply_offset, get_assignment, get_assignments, get_details_from_idx, get_loop_counters, get_slice_like_element_ty,
+    IndexExpr, MinifyingSugg, Start, StartKind,
+};
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::source::snippet;
+use clippy_utils::sugg::Sugg;
+use clippy_utils::ty::is_copy;
+use clippy_utils::{higher, path_to_local, sugg};
+use if_chain::if_chain;
+use rustc_ast::ast;
+use rustc_errors::Applicability;
+use rustc_hir::{BinOpKind, Expr, ExprKind, Pat, PatKind};
+use rustc_lint::LateContext;
+use rustc_middle::ty;
+use rustc_span::symbol::sym;
+use std::iter::Iterator;
+
+/// Checks for for loops that sequentially copy items within a slice
+pub(super) fn check<'tcx>(
+    cx: &LateContext<'tcx>,
+    pat: &'tcx Pat<'_>,
+    arg: &'tcx Expr<'_>,
+    body: &'tcx Expr<'_>,
+    expr: &'tcx Expr<'_>,
+) -> bool {
+    if let Some(higher::Range {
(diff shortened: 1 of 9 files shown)


untitaker:
Still found some bugs, ignore.

tests/ui/manual_memmove.stderr
```diff
@@ -0,0 +1,22 @@
+error: it looks like you're manually copying within a slice
+  --> $DIR/manual_memmove.rs:8:5
+   |
+LL | /     for i in 0..(arr.len() - 4) {
+LL | |         // left shift by 4
+LL | |         arr[i] = arr[i + 4];
+LL | |     }
+   | |_____^ help: try replacing the loop by: `arr.copy_within(4..((arr.len() - 4) + 4), 0);`
```
untitaker, line 8: The range bound could be simplified. My suspicion is that MinifyingSugg doesn't work properly here, but I would be "fine" with this output.

Jarcho:
Is there a reason you didn't just add a small extension to `manual_memcpy` to handle when the source and destination are the same? The detection code should be identical between this ant `manual_memcpy`.

I would consider using the same lint as well, possibly renaming to something a little more general.

clippy_lints/src/loops/manual_memmove.rs
```diff
@@ -0,0 +1,233 @@
...
+                    // we currently do not support more than one assignment, as it's "too hard" to
+                    // prove that the to-be-moved slices (+ their destinations) are nonoverlapping
+                    v.len() == 1
+                })
+                .map(|v| v.join("\n    "));
+
+            if let Some(big_sugg) = big_sugg {
+                span_lint_and_sugg(
+                    cx,
+                    MANUAL_MEMCPY,
```
untitaker, line 106: need to change this

untitaker:
@Jarcho i'll try that next time I work on this, indeed i was reusing a lot of stuff from manual_memcpy, but that comes from cargo-cult, and the potential future challenges with manual_memmove (see testcases) are quite different

flip1995:
Wat?

blyxyas:
@untitaker are you interested in continuing with this lint? Or should it be closed ヘ(^･ω･^=)\~

untitaker:
I think this lint was kind of in an OK state from what I can remember but could use further refactor to deduplicate code.

blyxyas:
Don't worry, no hard feeling (and you left some public work already done for future contributors). Have fun on other projects! (●ↀωↀ●) :heart:

PR #8344, opened 2022-01-24
Fix FP on `option_if_let_else` on conflicting mut borrow

### LINKED ISSUE #8141: option_if_let_else suggests invalid code when using mut references in the else clause
### Summary

There is a conditional branching code that uses `if let else` for the return value of `HashMap::get_mut()`.
Even though the else clause of this code uses a mutable reference of the HashMap, `option_if_let_else` suggests to use `map_or_else()`, but the suggested code causes an error in the borrow checker.

### Lint Name

option_if_let_else

### Reproducer

I tried this code:

```rust
// feature_ids is a HashMap.
let Some(v) = feature_ids.get_mut(&fid) {
    *v += 1.0;
} else {
    feature_ids.insert(fid, 1.0);
}
```

I saw this happen:

```
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
help: try
    |
150 ~                 feature_ids.get_mut(&fid).map_or_else(|| {
151 +                     feature_ids.insert(fid, 1.0);
152 +                 }, |v| {
153 +                     *v += 1.0;
154 +                 })
```

The suggested code is not compilable:
```
    |
150 |                 feature_ids.get_mut(&fid).map_or_else(|| {
    |                 ------------------------- ----------- ^^ second mutable borrow occurs here
    |                 |                         |
    |                 |                         first borrow later used by call
    |                 first mutable borrow occurs here
151 |                     feature_ids.insert(fid, 1.0);
    |                     ----------- second borrow occurs due to use of `feature_ids` in closure
```

### Version

```text
rustc 1.59.0-nightly (7abab1efb 2021-12-17)
binary: rustc
commit-hash: 7abab1efb21617ba6845fa86328dffa16cfcf1dc
commit-date: 2021-12-17
host: x86_64-apple-darwin
release: 1.59.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

dswij:
closes #8141 

changelog: [`option_if_let_else`] fix FP on conflicting mut borrow

### DIFF
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::{
-    can_move_expr_to_closure, eager_or_lazy, higher, in_constant, is_else_clause, is_lang_ctor, peel_blocks,
+    can_move_expr_to_closure, eager_or_lazy, fn_def_id, higher, in_constant, is_else_clause, is_lang_ctor, peel_blocks,
     peel_hir_expr_while, CaptureKind,
 };
 use if_chain::if_chain;
@@ -129,6 +129,18 @@ fn detect_option_if_let_else<'tcx>(cx: &LateContext<'tcx>, expr: &Expr<'tcx>) ->
             let (as_ref, as_mut) = match &let_expr.kind {
                 ExprKind::AddrOf(_, Mutability::Not, _) => (true, false),
                 ExprKind::AddrOf(_, Mutability::Mut, _) => (false, true),
+                ExprKind::MethodCall(..) => {
+                    if let Some(fn_did) = fn_def_id(cx, let_expr) {
+                        let fn_sig = cx.tcx.fn_sig(fn_did).input(0);
+                        if let Some(Mutability::Mut) = cx.tcx.erase_late_bound_regions(fn_sig).ref_mutability() {
+                            (false, true)
+                        } else {
+                            (false, false)
+                        }
+                    } else {
+                        (false, false)
+                    }
+                }
                 _ => (bind_annotation == &BindingAnnotation::Ref, bind_annotation == &BindingAnnotation::RefMut),
             };
             let cond_expr = match let_expr.kind {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/option_if_let_else.rs
```diff
@@ -129,6 +129,18 @@ fn detect_option_if_let_else<'tcx>(cx: &LateContext<'tcx>, expr: &Expr<'tcx>) ->
             let (as_ref, as_mut) = match &let_expr.kind {
                 ExprKind::AddrOf(_, Mutability::Not, _) => (true, false),
                 ExprKind::AddrOf(_, Mutability::Mut, _) => (false, true),
+                ExprKind::MethodCall(..) => {
+                    if let Some(fn_did) = fn_def_id(cx, let_expr) {
+                        let fn_sig = cx.tcx.fn_sig(fn_did).input(0);
+                        if let Some(Mutability::Mut) = cx.tcx.erase_late_bound_regions(fn_sig).ref_mutability() {
```
camsteffen, line 135: This should be the same since you're not doing anything with regions.
```suggestion
                        if let Some(Mutability::Mut) = cx.tcx.fn_sig(fn_did).skip_binder().inputs()[0].ref_mutability() {
```

clippy_lints/src/option_if_let_else.rs
```diff
@@ -141,7 +153,9 @@ fn detect_option_if_let_else<'tcx>(cx: &LateContext<'tcx>, expr: &Expr<'tcx>) ->
             // with the borrow checker. Currently only `<local>[.<field>]*` is checked for.
             if as_ref || as_mut {
                 let e = peel_hir_expr_while(cond_expr, |e| match e.kind {
-                    ExprKind::Field(e, _) | ExprKind::AddrOf(_, _, e) => Some(e),
+                    ExprKind::Field(e, _)
+                    | ExprKind::AddrOf(_, _, e)
+                    | ExprKind::MethodCall(_, _, [e, ..], _) => Some(e),
```
camsteffen, line 158: This seems incomplete. In the change higher up you look at the first `MethodCall` but here you can peel any number of method calls. Also, a function may take a mutable reference as input without necessarily returning a value that borrows the input.

This is a tricky problem and we have a lot of bugs like this. I think what we really need is a util like:

```rust
fn locals_borrowed_by_expr(&Expr) -> HirIdMap<Mutability>
```

and I think it should share some logic used by `ExprUseVisitor` but that logic is currently not public to Clippy.

dswij: The more I look at this, the more I feel that this is actually hacky.

There is a recent issue that got opened for this specifically #8346.

camsteffen: @dswij Are you okay with closing this PR, or do you want to give this another try?

Issue #7040, opened 2021-04-06
Remove unnecessary match


SaadiSave:
### What it does

*What does this lint do?*

Remove unnecessary match for error handling when function signatures are the same.

### Categories (optional)

- Kind: `style`, `correctness` ?

*What is the advantage of the recommended code over the original code*

Not sure if there are any performance improvements. The code is more readable and shorter though. This lint also applies for any function pair that returns the same enum, like `Option<T>` for example.

### Drawbacks

None.

### Example

```rust
// For example, a function has the signature:
fn fname(...) -> Result<T, E> { ... }

// and it is called in another function with the same signature:
fn fname2(...) -> Result<T, E> {
    // This is obviously unnecessary
    match fname(...) {
        Ok(s) => Ok(s),
        Err(e) => Err(e),
    }
}
```

Could be written as:

```rust
// For example, a function has the signature:
fn fname(...) -> Result<T, E> { ... }

// and it is called in another function with the same signature:
fn fname2(...) -> Result<T, E> {
    // Much better
    fname(...)
}
```

#### Real world example
```rust
async fn destructure_reqwest_response(res: Response) -> Result<Generic, InternalError> {
    ...
}

async fn get_count(...) -> Result<Generic, InternalError> {
    ...
    // Changed this later. Even clippy::pedantic did not detect this, so I assume that a lint is missing.
    match res {
        Ok(r) => match destructure_reqwest_response(r).await {
            Ok(r) => Ok(r),
            Err(e) => Err(e),
        },
        Err(_) => Err(INTERNALERROR),
    }
    ...
}
```

giraffate:
It would be better for categories to be `style`.

SaadiSave:
@giraffate 
Correctness is described as code that is completely unnecessary. Don't the examples above have completely unnecessary code? What makes `style` a better category?

giraffate:
> code that is outright wrong or very useless

`correctness` is described above in README. The original code isn't wrong and doesn't seems to lead some bugs. So I think `style (default warn)`, not `correctness (default deny)`, is appropriate for this.

SaadiSave:
Okay. Will change it

SaadiSave:
Can someone estimate how complex implementing this lint is?

camsteffen:
I think this should be `correctness` since it is "very useless". It could point out a bug where the coder meant to actually do something instead of nothing.

Name idea: `nop_match`.

> Can someone estimate how complex implementing this lint is?

[...]

PR #8356, opened 2022-01-27
Simplify if let statements

### LINKED ISSUE #8288: Simplify `if let` statements that immediately return unwrapped value
### What it does

Reduce simple `if let` statements that immediately return the unwrapped value by replacing the `if let` with a `?`.

### Lint Name

collapsible_if_let_return

### Category

style

### Advantage

Encourages concise code and proper use of the `?` operator.

### Drawbacks

None that I can think of.

### Example

```
fn example_fn() -> Result<(), CustomError> {
    if let Err(err) = run_some_process() {
        return Err(err);
    }
    // Other function logic...
    Ok(())
}
```

Could be written as:

```
fn example_fn() -> Result<(), CustomError> {
    run_some_process()?;
    // Other function logic...
    Ok(())
}
```

And, if there were no other function logic, could further be reduced to:

```
fn example_fn() -> Result<(), CustomError> {
    run_some_process()
}
```

J-ZhengLi:
fixes: #8288 

---

changelog: Allowing [`qustion_mark`] lint to check `if let` expressions that immediatly return unwrapped value

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -566,17 +566,20 @@ declare_clippy_lint! {
     ///
     /// ### Why is this bad?
     /// Readability, this can be written more concisely as
-    /// `_.flat_map(_)`
+    /// `_.flat_map(_)` for `Iterator` or `_.and_then(_)` for `Option`
     ///
     /// ### Example
     /// ```rust
     /// let vec = vec![vec![1]];
+    /// let opt = Some(5);
     ///
     /// // Bad
     /// vec.iter().map(|x| x.iter()).flatten();
+    /// opt.map(|x| Some(x * 2)).flatten();
     ///
     /// // Good
     /// vec.iter().flat_map(|x| x.iter());
+    /// opt.and_then(|x| Some(x * 2));
     /// ```
     #[clippy::version = "1.31.0"]
     pub MAP_FLATTEN,


tests/ui/question_mark.rs
```diff
@@ -186,6 +186,43 @@ fn f() -> NotOption {
...
+            return Err(err);
+        }
+        let mut res = 1_u8;
+        if some_flag {
+            res = 0_u8;
+        }
+        Ok(res)
+    }
+
+    /// No other function logic, same result, replace whole function body with `foo()`
```
dswij, line 209: I'd say that the check for 'same result' for the whole function body is 'too much' - the only case where it will happen is iff the function result can be determined easily and there is no other logic and the two result is exactly equal.

I think we can leave out this specific case.

tests/ui/question_mark.rs
```diff
@@ -186,6 +186,43 @@ fn f() -> NotOption {
...
+    }
+
+    /// No other function logic, different result, don't replace whole function body
+    fn qux() -> Result<u8, CustomError> {
+        if let Err(err) = foo() {
+            return Err(err);
+        }
+        Ok(2)
+    }
+}
```
dswij, line 224: nit: Let's also add a test case for when the `if let` does not immediately return `Err` i.e.

```rust
if let Err(err) = foo() {
    do_something();
    return Err(err);
}
```

clippy_lints/src/question_mark.rs
```diff
@@ -122,8 +128,22 @@ impl QuestionMark {
...
 
-    fn result_check_and_early_return(cx: &LateContext<'_>, expr: &Expr<'_>, nested_expr: &Expr<'_>) -> bool {
-        Self::is_result(cx, expr) && Self::expression_returns_unmodified_err(cx, nested_expr, expr)
+    fn check_lang_items(cx: &LateContext<'_>, qpath: &QPath<'_>, items: &[rustc_hir::LangItem]) -> bool {
+        for lang_item in items {
+            if is_lang_ctor(cx, qpath, *lang_item) {
+                return true;
+            }
+        }
+        false
```
dswij, line 137: nit: I think we can use `items.iter().any(..)` here

clippy_lints/src/question_mark.rs
```diff
@@ -95,19 +95,25 @@ impl QuestionMark {
...
+            if let Some(higher::IfLet { let_pat, let_expr, if_then, if_else })
                 = higher::IfLet::hir(cx, expr);
             if let PatKind::TupleStruct(ref path1, fields, None) = let_pat.kind;
-            if (Self::option_check_and_early_return(cx, let_expr, if_else) && is_lang_ctor(cx, path1, OptionSome)) ||
-                (Self::result_check_and_early_return(cx, let_expr, if_else) && is_lang_ctor(cx, path1, ResultOk));
-
-            if let PatKind::Binding(annot, bind_id, _, _) = fields[0].kind;
+            // Only check one of the blocks
+            let nested_expr = if_else.unwrap_or(if_then);
+            if Self::check_lang_items(cx, path1, &[OptionSome, ResultOk, ResultErr]);
```
dswij, line 103: I don't think this is correct to place it here, this will result in a different behavior compared to the original. When the let pattern is `Result` we want the return to also be `Result`, but placing it here will make `option_check_and_early_return` called for `Result`.

For example, I believe this will trigger when `check_lang_items` is here

```rust
fn test_1() -> Option<String> {
    if let Err(err) = function_returning_result() {
        return Some(err);
    }

    None
}
```

J-ZhengLi: Ahhh, I see, thank you for all the feedbacks!

tests/ui/question_mark.rs
J-ZhengLi: And this one shouldn't trigger right?

clippy_lints/src/question_mark.rs
J-ZhengLi: the original code kinda has this problem as well, yes, your example did trigger this lint, but I guess maybe it is because [this block here](https://github.com/rust-lang/rust-clippy/blob/1c87b2df8ae13dfb404805840c041789778310e6/clippy_lints/src/question_mark.rs#L193)? where I only check for `err` symbol inside the parentheses... 

How do I compare `Err()` and `Some()` tho? Their HirId.owner look the same but I can't directly compare them right?

dswij: Before the changes, it's checked here

https://github.com/rust-lang/rust-clippy/blob/ea4db3a6c6fdd30abb0b746f7a68d403a07d24bf/clippy_lints/src/question_mark.rs#L101-L102

This does not trigger before ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4db6d0ba767bc645f714e97e7996db64))

J-ZhengLi: Cuz it didn't check `ExprKind::Call` return and `if without else` at all

dswij: Whoops, that playground example I gave is wrong.

But yeah, the problem is a little different now, I think what I meant to say is that, previously we checked it with those lines and it's sufficient, but this is not the case anymore. 

For the changes, I think you can just check if the return value is an `Err()` because we matched `if let Err() ...` in the let pattern, without checking the owner.

J-ZhengLi: That helps a lot, thanks!

J-ZhengLi:
Upon attempting to fix this issue, I found a new problem here, maybe I'm just a little bit confused.

After removing [this line below](https://github.com/rust-lang/rust-clippy/blob/7ceffdee9bc8dc7b645cb1b91bec5df8a6716087/clippy_lints/src/question_mark.rs#L106)
```rust
if path_to_local_id(peel_blocks(if_then), bind_id);
```

This lint will also triggered by:

```rust
fn check_option(x: Option<i32>) -> Option<i32> {
    do_something();
    if let Some(a) = x {
        Some(a)
    } else {
        None
    }
}
```

which is already triggering ``[`manual_map`]``, and offered a suggestion of replacing the if let with `x.map(|a| a)`

shouldn't a `x?` be more intuitive in this case?

camsteffen:
@J-ZhengLi Sorry for the delay. I'll take a look at this soon. It looks like some internal lints are failing. Try running `cargo test --features internal`.

camsteffen:
> shouldn't a `x?` be more intuitive in this case?

That is a case where the `match` doesn't do anything. There is a suggested lint for that at #7040. For now, I would just let it be linted by `manual_map` only.

J-ZhengLi:
> @J-ZhengLi Sorry for the delay. I'll take a look at this soon. It looks like some internal lints are failing. Try running `cargo test --features internal`.

Don't worry, I'm on a vacation anyway XD... 

> For now, I would just let it be linted by `manual_map` only.

Thanks, and I'll be working on it in two days

J-ZhengLi:
@camsteffen I hope it's good now, let me if anything comes up

tests/ui/question_mark.rs
```diff
@@ -186,6 +186,68 @@ fn f() -> NotOption {
     NotOption::First
 }
 
+#[allow(dead_code)]
+mod issue8288 {
```
camsteffen, line 190: The mod isn't necessary. I'd only reference the issue if it has non-obvious context.

clippy_lints/src/question_mark.rs
```diff
@@ -93,21 +93,52 @@ impl QuestionMark {
...
+            // otherwise it will cause trouble when replacing blocks with question mark.
+            if !Self::expr_is_else_if(cx, expr);
             if let PatKind::TupleStruct(ref path1, fields, None) = let_pat.kind;
-            if (Self::option_check_and_early_return(cx, let_expr, if_else) && is_lang_ctor(cx, path1, OptionSome)) ||
-                (Self::result_check_and_early_return(cx, let_expr, if_else) && is_lang_ctor(cx, path1, ResultOk));
-
-            if let PatKind::Binding(annot, bind_id, _, _) = fields[0].kind;
-            let by_ref = matches!(annot, BindingAnnotation::Ref | BindingAnnotation::RefMut);
-            if path_to_local_id(peel_blocks(if_then), bind_id);
+            let nested_expr = if_else.unwrap_or(if_then);
```
camsteffen, line 125: It doesn't seem right to treat the if-block and the else-block the same here.

clippy_lints/src/question_mark.rs
```diff
@@ -122,8 +153,56 @@ impl QuestionMark {
...
+        let if_then_is_ret_stmt = matches!(if_then_kind, ExprKind::Ret(..));
+
+        if if_else.is_none() {
+            return if_then_is_ret_stmt;
+        }
+        if let ExprKind::Ret(_) = peel_blocks_with_stmt(if_else.unwrap()).kind {
+            return if_then_is_ret_stmt;
+        }
+        false
+    }
```
camsteffen, line 197: I would expect this to be implemented by checking whether the parent node of the expression is a statement.

clippy_lints/src/question_mark.rs
```diff
@@ -122,8 +153,56 @@ impl QuestionMark {
...
+        if_chain! {
+            if let ExprKind::If(..) = expr.kind;
+            if let Some(parent_expr) = clippy_utils::get_parent_expr(cx, expr);
+            if let ExprKind::If(..) = parent_expr.kind;
+            then {
+                return true;
+            }
+        }
+        false
+    }
```
camsteffen, line 167: Can you use `clippy_utils::is_else_clause`?

clippy_lints/src/question_mark.rs
```diff
@@ -156,10 +235,32 @@ impl QuestionMark {
...
+        expr: &Expr<'_>,
+        cond_expr: &Expr<'_>,
+        pat_symbol: Option<Symbol>,
+    ) -> bool {
+        match &peel_blocks_with_stmt(expr).kind {
+            ExprKind::Ret(Some(ret_expr)) => {
+                Self::expression_returns_unmodified_err(cx, ret_expr, cond_expr, pat_symbol)
+            },
             ExprKind::Path(_) => path_to_local(expr).is_some() && path_to_local(expr) == path_to_local(cond_expr),
+            ExprKind::Call(call_expr, args_expr) => {
```
camsteffen, line 249: This is no longer requiring `return ..`?

clippy_lints/src/question_mark.rs
```diff
@@ -122,8 +153,56 @@ impl QuestionMark {
...
+        false
+    }
+
+    fn result_check_and_early_return(
+        cx: &LateContext<'_>,
+        expr: &Expr<'_>,
+        nested_expr: &Expr<'_>,
+        pat_symbol: Option<Symbol>,
+    ) -> bool {
+        Self::is_result(cx, expr) && Self::expression_returns_unmodified_err(cx, nested_expr, expr, pat_symbol)
```
camsteffen, line 205: I think we could use some refactoring to avoid code repetition in a lot of these utils. For example, don't get the expression type in all the `is_*` functions. And just look for `ExprKind::Ret` in one place rather than having a bunch of `*_and_early_return` utils. Maybe you can have a refactor commit before other changes?

camsteffen (changes requested):

tests/ui/question_mark.rs
J-ZhengLi: well, I did that because I wanted to suppress dead_code warning for all those functions, I'll think of a better way then :)

camsteffen: You can add `#![allow(dead_code)]` at the top of the file.

J-ZhengLi: Awesome, I was kinda afraid to do that for some reason

camsteffen: Yeah we try to keep allows minimal but allowing dead/unused is generally okay for tests.

J-ZhengLi: Awesome, I was overthinking other people on this lint might want it, now I guess those function call in `main()` won't be needed

clippy_lints/src/question_mark.rs
J-ZhengLi: Sorry, but I'm not sure I understand this, why "no longer requiring `return`?", can you explain this a bit more? XD

camsteffen: Oh I thought this function is only supposed to return true for `return ..` expressions (`ExprKind::Ret`) but I guess that is not true. The name is just misleading.

J-ZhengLi: Oh I see~

clippy_lints/src/question_mark.rs
J-ZhengLi: Did you suggesting `clippy_utils::get_parent_node`? if yes, I don't understand how am I going to do with it, since the parent node of them will give me the entire if expression. Or perhaps I checked the wrong children...

clippy_lints/src/question_mark.rs
J-ZhengLi: I tried refactoring it, and it still looks like speghatti unfortunatly

camsteffen: Instead of `is_early_return(..) || is_early_return(..)`, try doing the following. I think that will make things a lot simpler.

```rust
if let Some(adt) = caller_ty.ty_adt_ty();
if let Some(name)  = cx.tcx.get_diagnostic_name(adt.did);
if matches!(name, sym::Option | sym::Result);
```

clippy_lints/src/question_mark.rs
camsteffen: I'm not sure what you mean. Wouldn't this work?

```rust
let requires_semi = matches!(get_parent_node(expr), Some(Node::Stmt(_)));

Comment (J-ZhengLi) [reply clippy_lints/src/question_mark.rs:133]: Yep, it works, I was just being dumb thinking you want me to check the parent node of the return value, like the `if_then_is_ret_stmt` in this case, sorry for causing confusion.

File: clippy_lints/src/question_mark.rs (hunk 10 above)
Comment (J-ZhengLi) [reply clippy_lints/src/question_mark.rs]: Ummm, I tried copy-paste this code, but I think it just brings back the issue #7859.

File: clippy_lints/src/question_mark.rs [hunk 11]
```diff
@@ -39,135 +42,185 @@ declare_clippy_lint! {
...
+}
 
-    fn expression_returns_unmodified_err(expr: &Expr<'_>, cond_expr: &Expr<'_>) -> bool {
-        match peel_blocks_with_stmt(expr).kind {
-            ExprKind::Ret(Some(ret_expr)) => Self::expression_returns_unmodified_err(ret_expr, cond_expr),
-            ExprKind::Path(_) => path_to_local(expr).is_some() && path_to_local(expr) == path_to_local(cond_expr),
-            _ => false,
-        }
+fn offer_suggestion(cx: &LateContext<'_>, expr: &Expr<'_>, suggestion: String, applicability: Applicability) {
+    if !suggestion.is_empty() {
```
dswij, line 213: Let's use `Option<String>` rather than check for emptiness here.

dswij: I think we can even remove this check, I don't see a scenario where `suggestion` should be empty.

clippy_lints/src/question_mark.rs
```diff
@@ -39,135 +42,185 @@ declare_clippy_lint! {
...
+            let by_ref = !caller_ty.is_copy_modulo_regions(cx.tcx.at(caller.span), cx.param_env) &&
+                !matches!(caller.kind, ExprKind::Call(..) | ExprKind::MethodCall(..));
+            let mut suggestion = String::new();
+            if let Some(else_inner) = r#else {
+                if eq_expr_value(cx, caller, peel_blocks(else_inner)) {
+                    suggestion = format!("Some({}?)", receiver_str);
                 }
+            } else {
+                suggestion = format!("{}{}?;", receiver_str, if by_ref { ".as_ref()" } else { "" });
             }
```
dswij, line 106: nit: Now that we're using let-chains, we can probably do:

```rust
if let Some(else_inner) = r#else && eq_expr_value(cx, caller, peel_blocks(else_inner) {
	suggestion = format!("Some({}?)", receiver_str);
} else {
	suggestion = format!("{}{}?;", receiver_str, if by_ref { ".as_ref()" } else { "" });
}
```

dswij (commented):
Apologies that this got sidelined for a while. Thanks for the PR and for being patient!

The change itself looks great to me, just some small comments and I after that I think it's ready to be merged

clippy_lints/src/question_mark.rs
J-ZhengLi: finally~ I saw a pr which introducing it at the time, and I think I even posted a commit changing all the if-chains to let-chains, but I eventually revert back after that pr was closed due to concerns of formatting etc.

btw I'll get back to this pr once I got my hard drive fixed 🤦‍♂️

J-ZhengLi: > nit: Now that we're using let-chains, we can probably do:
> 
> ```rust
> if let Some(else_inner) = r#else && eq_expr_value(cx, caller, peel_blocks(else_inner) {
> 	suggestion = format!("Some({}?)", receiver_str);
> } else {
> 	suggestion = format!("{}{}?;", receiver_str, if by_ref { ".as_ref()" } else { "" });
> }
> ```

*hooray! finally got a new hard drive*

well, I did that but only realized it breaks one of the original tests, like when the return in `else` block is different than the caller:

```rust
if Some(1).is_none() {
    return None;
} else {
    return Some(0);
}
```

So I guess... I can't change too mach on this part?

dswij: Oh wait, you're right. They're not exactly the same. :facepalm:

dswij:
Thanks so much for this amazing work and refactors!
@bors r+ squash

PR #8363, opened 2022-01-28
extending `question_mark` lint for issue#8288

### LINKED ISSUE #8288: Simplify `if let` statements that immediately return unwrapped value

J-ZhengLi:
fixes #8288 

I somehow broke my last draft PR by deleting too many commits... 🤪 so here is a new one...

---

changelog:
for ``[`question_mark`]``
* Add new test cases
* Add code to check for unmodified `Err()`

**a million thanks to** @dswij 

Is it safe too assume that clippy will be using `let_chains` in the future?

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -5,6 +5,7 @@
 #![feature(control_flow_enum)]
 #![feature(drain_filter)]
 #![feature(iter_intersperse)]
+#![feature(let_chains)]
 #![feature(let_else)]
 #![feature(once_cell)]
 #![feature(rustc_private)]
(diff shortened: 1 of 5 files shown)


J-ZhengLi:
oops... I breaks uitest for ``[`manual_map_option`]``

camsteffen:
> I somehow broke my last draft PR by deleting too many commits... 🤪 so here is a new one...

You should be able to fix it with something like...

```shell
git checkout broken-branch
git reset fixed-branch --hard
```
Feel free to do that and reopen the first PR.

camsteffen:
Oh unless they are the same branch? Maybe just `git rebase master`? Oh well, no big deal.

PR #8365, opened 2022-01-28
Add `explicit_write` suggestions for `write!`s with format args

### LINKED ISSUE #4542: Revive suggestion on explicit_write
```rust
writeln!(std::io::stderr(), "foo {}", "bar").unwrap();
```

should be

```rust
eprintln!("foo {}","bar");
```

but clippy suggests

```
  |
4 |     writeln!(std::io::stderr(), "foo {}", "bar").unwrap();
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `eprintln!("foo ")`
```

It should be

```
  |
4 |     writeln!(std::io::stderr(), "foo {}", "bar").unwrap();
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `eprintln!("foo {}", "bar")`
```

([Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=50eace297cae2cc57d27af4f7322ad48))

---

## UPDATE

#4544 removed the wrong suggestion. We should find the right way and revive suggestions.

Alexendoo:
changelog: Add [`explicit_write`] suggestions for `write!`s with format args

Fixes #4542

```rust
writeln!(std::io::stderr(), "macro arg {}", one!()).unwrap();
```

Now suggests:

```
error: use of `writeln!(stderr(), ...).unwrap()`
  --> $DIR/explicit_write.rs:36:9
   |
LL |         writeln!(std::io::stderr(), "macro arg {}", one!()).unwrap();
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `eprintln!("macro arg {}", one!())`
```

---------

r? @camsteffen (again, sorry 😛) for the `FormatArgsExpn` change

Before this change `inputs_span` returned a span pointing to just `1` in

```rust
macro_rules! one {
    () => { 1 };
}

`writeln!(std::io::stderr(), "macro arg {}", one!()).unwrap();`
```

And the `source_callsite` of that span didn't include the format string, it was just `one!()`


clippy_utils/src/macros.rs
```diff
@@ -461,11 +462,13 @@ impl<'tcx> FormatArgsExpn<'tcx> {
             .collect()
     }
 
-    /// Span of all inputs
+    /// Source callsite span of all inputs
     pub fn inputs_span(&self) -> Span {
+        let format_string = self.format_string_span.source_callsite();
```
camsteffen, line 467: If this is needed then I think `self.format_string_span` needs to be fixed?

Alexendoo: It wasn't needed there, didn't think to check individually

clippy_utils/src/macros.rs
```diff
@@ -461,11 +462,11 @@ impl<'tcx> FormatArgsExpn<'tcx> {
             .collect()
     }
 
-    /// Span of all inputs
+    /// Source callsite span of all inputs
     pub fn inputs_span(&self) -> Span {
         match *self.value_args {
             [] => self.format_string_span,
-            [.., last] => self.format_string_span.to(last.span),
+            [.., last] => self.format_string_span.to(last.span.source_callsite()),
```
camsteffen, line 469: If the `format_args!(..)` call is inside a macro definition, then we don't want to walk all the way to the source root.
```suggestion
            [.., last] => self.format_string_span.to(hygiene::walk_chain(last.span, self.format_string_span.ctxt())),
```

Alexendoo: 👍

Changed

clippy_lints/src/explicit_write.rs
```diff
@@ -80,13 +81,8 @@ impl<'tcx> LateLintPass<'tcx> for ExplicitWrite {
...
                 let msg = format!("use of `{}.unwrap()`", used);
-                if let [write_output] = *format_args.format_string_parts {
-                    let mut write_output = write_output.to_string();
-                    if write_output.ends_with('\n') {
-                        write_output.pop();
-                    }
-
-                    let sugg = format!("{}{}!(\"{}\")", prefix, sugg_mac, write_output.escape_default());
+                if let Some(inputs_snippet) = snippet_opt(cx, format_args.inputs_span()) {
+                    let sugg = format!("{}{}!({})", prefix, sugg_mac, inputs_snippet);
```
camsteffen, line 85: I think you can use `snippet_with_applicability` here and remove the else `span_lint_and_sugg` case. Otherwise this looks good!

camsteffen:
Thanks!

@bors r+

PR #8370, opened 2022-01-29
Format `if_chain` invocations in clippy_utils

smoelius:
Not realizing it was [already obsolete](https://github.com/rust-lang/rust-clippy/pull/8360), I built a [tool to format inside `if_chain` invocations](https://crates.io/crates/rustfmt_if_chain).

This PR applies the tool to clippy_utils. (If you apply it to clippy_lints, the changes are extensive.)

Anyway, I'm making it known here in case anyone wants to use it while `if-let` chain support is developed for `rustfmt`. (There could be a few Clippy PRs between now and then, and IMHO, the code looks better with the `if_chain` invocations formatted.)

Cheers.

---

changelog: none


flip1995:
Thanks! This is an improvement of the formatting. Probably not necessary soon, once we switch to let_chains, but until we can make the move, I'm happy to merge this.

@bors r+

xFrednet:
The changelog entry was missing. I added `changelog: none` as this PR only changed some formatting :)

@bors retry

PR #8376, opened 2022-01-31
[`chars_next_cmp`] Fix unescaped suggestion

### LINKED ISSUE #8373: clippy fix error: `character constant must be escaped: '\n'`
### Description

For the example below, clippy raises the warning `you should use the 'ends_with' method` for line 3:

```rust
fn main() {
    let mut buffer = String::from("foo");
    if buffer.chars().last().unwrap() == '\n' {
        buffer.pop();
    }
}
```

However, `clippy --fix` is not able to fix this as it aborts with the following error:

```
error: character constant must be escaped: `\n`
 --> src/main.rs:3:26
  |
3 |       if buffer.ends_with('
  |  __________________________^
4 | | ') {
  | |_ help: escape the character: `\n`

```

### Version

```text
binary: rustc
commit-hash: db9d1b20bba1968c1ec1fc49616d4742c1725b4b
commit-date: 2022-01-20
release: 1.58.1
LLVM version: 13.0.0
```

### Additional Labels

_No response_

dswij:
closes #8373 

changelog: [`chars_next_cmp`] Fix unescaped suggestion

### DIFF
--- a/clippy_lints/src/methods/chars_cmp_with_unwrap.rs
+++ b/clippy_lints/src/methods/chars_cmp_with_unwrap.rs
@@ -32,7 +32,7 @@ pub(super) fn check<'tcx>(
                         if info.eq { "" } else { "!" },
                         snippet_with_applicability(cx, args[0][0].span, "..", &mut applicability),
                         suggest,
-                        c),
+                        c.escape_default()),
                 applicability,
             );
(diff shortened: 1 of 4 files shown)


camsteffen:
Thanks!

@bors r+

PR #8385, opened 2022-02-01
add doc_link_with_quotes lint

### LINKED ISSUE #8383: detect ['foo'] (quotes, not backticks) in doc comments
### What it does

Highlights when someone writes:
```rust
/// ['foo::Bar']
```
when they probably meant
```rust
/// [`foo::Bar`]
```
This recently caught me out. `cargo doc` reported no broken links, but I had a couple lurking that were using this format, so were missed by the broken link check.

The characters are visually similar enough to be very easy to confuse

### Lint Name

doc_link_with_quotes

### Category

suspicious

### Advantage

Typing ['foo::Bar'] (with single quotes) is almost certainly a mistake, either because they genuinely think that is the correct syntax, or as a typo. Having clippy flag likely leads to better 

### Drawbacks

There are likely some false positives, since this syntax is meaningful in other languages (e.g. in JS/TS, Python and Dart (probably more than that) they represent list containing a string literal, though in those cases I suspect it would already be enclosed in backticks, so the actual false positive rate is likely quite low

### Example

```rust
/// Does the same thing as ['bar']
fn foo() {}
```

Could be written as:

```rust
/// Does the same thing as [`bar`]
fn foo() {}
```

cameron1024:
I'm not sure about wording, it seems OK to me but happy to change if other people have better ideas

closes #8383 

---
changelog: add [`doc_link_with_quotes`] lint


tests/ui/doc_link_with_quotes.rs
```diff
@@ -0,0 +1,12 @@
...
+fn main() {
+    foo()
+}
+
+/// Calls ['bar']
+pub fn foo() {
+    bar()
+}
+
+pub fn bar() {}
```
llogiq, line 12: I'd like to see a test case against false positives, e.g.

```
/// ['This week in Rust'](https://this-week-in-rust.org)
fn twir() {}
```

llogiq (commented):
I think we should look at more cases to ensure the false positive rate is low. The best way would presumably to check if the link text contains a valid path, but that may be too complex.

cameron1024:
The particular case that motivated my example was looking for broken links, but I came across something with a broken link *and* quotes instead of backticks. Ideally the lint would catch these, but I guess that comes at the cost of false positives. Anecdotally, the use of `['something']` seems to be mostly limited to being already within backticks, but I haven't gathered any data for that 🤷 

Though if by "valid path" you mean a much simpler syntactic check (rather than checking if it points to an actual item), that sounds reasonable. i.e.:
 - `['invalid_identifier,.<>/?']` won't trigger the lint
 - `['valid::but::doesnt::exist']` still would

Does that seem like a good way to go?

llogiq:
That's exactly what I meant.

llogiq:
You will need to rebase on master.

llogiq:
So this has waited long enough, I fixed the conflicts. Thanks for your patience.

@bors r+

PR #8395, opened 2022-02-05
single_match: Lint structs and tuple structs

jubnzv:
These changes allow `single_match` lint to suggest the simplification of `match` constructions for structs and tuple structs with the same name, if they form an exhaustive match.

For example:

```rust
// lint: S(_, _) forms an exhaustive match, so it could be removed
match s {
    S(42, _a) => {},
    S(_, _) => {},
}
```

Context: https://github.com/rust-lang/rust-clippy/pull/8322#issuecomment-1025316474

---

changelog: [`single_match`]: Lint tuple structs with the same name


clippy_lints/src/matches.rs
```diff
@@ -849,63 +868,119 @@ fn check_single_match_opt_like<'a>(
...
-            true
+        (
+            PatKind::TupleStruct(left_qpath, left_in, left_pos),
+            PatKind::TupleStruct(right_qpath, right_in, right_pos),
+        ) => {
+            let are_structs_with_the_same_name =
+                || -> bool { cx.qpath_res(left_qpath, left.hir_id) == cx.qpath_res(right_qpath, right.hir_id) };
+            let are_known_enums =
+                || -> bool { known_enums::is_known_enum(cx, left) && known_enums::is_known_enum(cx, right) };
+            if are_structs_with_the_same_name() || are_known_enums() {
```
camsteffen, line 911: Just check the type on one side. We can assume the types are equal.

clippy_lints/src/matches.rs
```diff
@@ -849,63 +868,119 @@ fn check_single_match_opt_like<'a>(
...
+            }
+            false
         },
-        PatKind::Path(ref path) => {
-            let path = rustc_hir_pretty::to_string(rustc_hir_pretty::NO_ANN, |s| {
-                s.print_qpath(path, false);
-            });
-            acc.push((path, ty));
-            true
+        (PatKind::Binding(BindingAnnotation::Unannotated, .., ident, None), _) if contains_only_wilds(right) => {
```
camsteffen, line 916: The `BindingAnnotation` shouldn't matter here.
```suggestion
        (PatKind::Binding(.., ident, None), _) if contains_only_wilds(right) => {
```

Also, I think you can treat `Binding` the same as `Wild`?

Thirdly, I think we should handle the case with `Some` at the end. You could do this at the top of this function:

```rust
fn peel_subpatterns(pat) {
    while let PatKind::Binding(.., Some(sub)) = pat.kind {
        pat = sub;
    }
    pat
}
let left = peel_subpatterns(left);
let right = peel_subpatterns(right);
```

clippy_lints/src/matches.rs
```diff
@@ -849,63 +868,119 @@ fn check_single_match_opt_like<'a>(
...
-        }),
-        PatKind::TupleStruct(ref path, ..) => {
-            let path = rustc_hir_pretty::to_string(rustc_hir_pretty::NO_ANN, |s| {
-                s.print_qpath(path, false);
-            });
-            acc.push((path, ty));
-            true
+/// Returns true if the given arms of the pattern forms exhaustive match that can be simplified to
+/// `if let` construction.
+fn check_exhaustive<'a>(cx: &LateContext<'a>, left: &Pat<'_>, right: &Pat<'_>, ty: Ty<'a>) -> bool {
```
camsteffen, line 897: I think this should say "returns false if the patterns exhaustively match an enum"

camsteffen (changes requested):
I reviewed about 75% so far.

clippy_lints/src/matches.rs
jubnzv: Thanks, I agree. You meant "returns true", right?

camelid:
I'm not sure what you mean by "with the same name". Any other name would result in a type error?

clippy_lints/src/matches.rs
camsteffen: No. By exhaustive I mean than all enum variants are present between the two patterns (no wild or binding).

jubnzv:
> I'm not sure what you mean by "with the same name". Any other name would result in a type error before clippy ran, right?

Yes, that would be a compile error.

clippy_lints/src/matches.rs
```diff
@@ -849,63 +868,119 @@ fn check_single_match_opt_like<'a>(
...
+    /// simplified by the `matches` lint.
+    pub fn is_known_enum(cx: &LateContext<'_>, pat: &Pat<'_>) -> bool {
+        match pat.kind {
+            PatKind::Path(ref path) => contains(cx, &qpath_to_string(path), cx.typeck_results().pat_ty(pat)),
+            PatKind::TupleStruct(ref path, ..) => {
+                let ty = cx.typeck_results().pat_ty(pat);
+                let path = &qpath_to_string(path);
+                (contains_only_wilds(pat) || contains_single_binding(pat)) && contains(cx, path, ty)
+            },
+            _ => false,
```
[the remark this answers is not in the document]
jubnzv: Done.

clippy_lints/src/matches.rs
jubnzv: Could you please explain in more detail?

We actually check here if we have structures with the same name. This is required to correctly handle the following case (added as unit test at [7fc3545](https://github.com/rust-lang/rust-clippy/pull/8395/commits/7fc3545cd6eae7d94aff3b2e07d3317be8b2cae7)):
```rust
    enum E {
        S1(i32, i32),
        S2(i32, i32),
    }
    let s = E::S1(1, 2);

    // don't lint
    match s {
        E::S1(..) => {},
        E::S2(..) => {},
    }
```

We detect these cases using `are_structs_with_the_same_name`. This check should be performed for `TupleStructs` patterns only, because any other case will cause compile error. For example, we cannot match for `Struct` patterns with the different types.

The second condition `are_known_enums` is required to check if we have one of the standard `TupleStruct` patterns in the both arms. This allows us to handle cases like this:

```rust
match x {
  // `Err` and `Ok` are both `TupleStruct`'s, so we must process them here.
  Err(..) => {},
  Ok(..) => {},
}
```

clippy_lints/src/matches.rs
jubnzv: I agree, I added support for bindings as you suggest.

But I have some problems with test suite after the changes.
[The error](https://github.com/jubnzv/rust-clippy/runs/5104199888?check_suite_focus=true#step:7:826) states: `'[ui] ui/patterns.rs' panicked at 'failed to apply suggestions for "tests/ui/patterns.rs" with rustfix: Cannot replace slice of data that was already replaced`.

As far as I understand, the issue occurs when the test suite tries to apply `rustfix` on `patterns.rs` to check if it matches `patterns.fixed`.  According to stack trace, the problem is in `rustfix` library, and it is not related with `clippy` code. I found a few similar issues [[1]](https://github.com/rust-lang/rustfix/issues/114) [[2]](https://github.com/rust-lang/rust-clippy/pull/8144#discussion_r773426034) in `clippy` and `rustfix` repositories, but it didn't help me much.

So at the moment I'm not sure how to fix it and I need some time to figure it out. Any help would be helpful.

clippy_lints/src/matches.rs
camsteffen: In your first example the type of both patterns is `E`. In the second example the type of both patterns is `Result`. The type is always the same. The name of the enum variant or the enum doesn't matter since you can just check the type.

clippy_lints/src/matches.rs
camsteffen: That error means that rustfix applied your suggestion, reran the lint on the updated code, and then got another suggestion for the same code. Like an infinite loop. Try manually applying the suggestions and then see if you get another suggestion in the same code.

clippy_lints/src/matches.rs
camsteffen: > But we want to make sure that the both variants have the same type as well.

This doesn't make sense. You are checking the type of the pattern and the type of the pattern is the enum, not the variant.

I disagree that it should lint if the variants are the same. That is still an exhuastive match of the enum. The point of #8282 is that the `match` expression will no longer compile if a variant is added to the enum, and that is what the user wants because the compiler helps them update their code.

clippy_lints/src/matches.rs
jubnzv: Thank you. I understood.

The problem was that after my changes clippy suggests to replace these lines:
```rust
    match v {
        Some(x) => (),
        y @ _ => (),
    }
```

To these:
```rust
if let Some(x) = x = { () }
```

But this code itself could be simplified to:
```rust
if let Some(x) = x = {}
```

So we got an infinite loop.

I fixed this file to keep its original logic and not trigger `single_match` lint.

clippy_lints/src/matches.rs
jubnzv: Well, my idea was is that the user may want to extend the enum with more variants later. So the `match` expression will give him warnings about non-exhaustive matching.

But suggesting to replace such `match` to `if let` is reasonable and should not be considered as a false positive. So, I agree, let's lint these patterns as well.

camsteffen: > The case we want to lint is if the enum has a variant with the single struct, and we match this struct in both arms:

> ```rust
>     // lint
>     match s {
>         E::S1(42, ..) => {},
>         E::S1(..) => {},
>     }
> ```

> I disagree that it should lint if the variants are the same.

> So, I agree, let's lint these patterns as well.

Our communication broke down somewhere. I think we should not lint this case.

clippy_lints/src/matches.rs
camsteffen: Hmm but I think the first example *should* trigger `single_match`. Which lint fired after the suggestion was applied?

jubnzv: > Hmm but I think the first example should trigger single_match

Yes, it seems logical. But for some reasons we have [the following check](
https://github.com/rust-lang/rust-clippy/tree/51045172c1313ed912742a521cd02b639c120fc4/clippy_lints/src/matches/single_match.rs#L29) for the `els` branch. And such cases are explicitly rejected.

According to `git blame` this check was added [here](https://github.com/jubnzv/rust-clippy/commit/6f25adbd5a1600653e8edf8de7135c397d73c07a#diff-f94a8b7cc3dc3012f81240077f02b32e16e5c1b43af7792c386468c82a01cc1fR610). There are no tests or additional context here, so I'm not sure, what problem this solves. Probably, it's better to leave this logic as is in this MR and come back to it later.

> Which lint fired after the suggestion was applied?

First, `clippy` suggests to replace `y @ _` to `y` with `redundant_pattern` check. Then we get code that can be simplified with `single_match` check, and we try to execute `rustfix` again to fix it.

clippy_lints/src/matches.rs
jubnzv: Ah, sorry for that. These corner cases are not easy to understand.

You mean, that we should not lint cases with non-standard enums on both arms at all? Even if they have a single member.

e.g.:
```rust
   // don't lint
    match s {
        E::S1(42, ..) => {},
        E::S1(..) => {},
    }

    // lint, because we have a wild pattern
    match s {
        E::S1(42, ..) => {},
        _ => {},
    }
```

And we could simplify the condition for the lint case just to `is_known_enum(left_tuple_pattern)`, right? We could do this, because if the first arm have one of the standard types (`Result`, `Option`, etc.), we can be sure, that the second one have the same type or wildcard/binding pattern.

So the condition will become:
```rust
        (PatKind::TupleStruct(_, left_in, left_pos), PatKind::TupleStruct(_, right_in, right_pos))
            if contains_only_wilds(right) =>
        {
            if !contains_only_known_enums(cx, &left) {
                return false;
            }
            check_exhaustive_tuples(cx, left_in, left_pos, right_in, right_pos)
        },
```

camsteffen: That sounds right 👍

clippy_lints/src/matches/single_match.rs
```diff
@@ -140,74 +140,128 @@ fn check_opt_like<'a>(
...
+}
 
-    let mut paths_and_types = Vec::new();
-    if !collect_pat_paths(&mut paths_and_types, cx, arms[1].pat, ty) {
-        return;
+/// Resturns true if the given type is one of the standard `Enum`s we know will never get any more
+/// members.
+fn is_known_enum(cx: &LateContext<'_>, ty: Ty<'_>) -> bool {
+    const CANDIDATES: &[&[&str; 3]] = &[&paths::COW, &paths::OPTION, &paths::RESULT];
+    CANDIDATES.iter().any(|&ty_path| match_type(cx, ty, ty_path))
```
camsteffen, line 152: One more improvement here is to use diagnostic items, which are preferred over paths.

```rust
let Some(adt) = ty.ty_adt_def() else { return false };
matches!(cx.tcx.get_diagnostic_name(adt.did), Some(sym::Cow | sym::Option | sym::Result))
```

clippy_lints/src/matches/single_match.rs
```diff
@@ -216,54 +270,10 @@ fn contains_only_wilds(pat: &Pat<'_>) -> bool {
...
-                }
-                if !contains_only_wilds(&left_in[i - left_dot_space])
-                    && !contains_only_wilds(&right_in[i - right_dot_space])
-                {
-                    return false;
-                }
-            }
-            true
-        },
+fn contains_single_binding(pat: &Pat<'_>) -> bool {
```
camsteffen, line 273: I don't think this function is needed since wilds and bindings are effectively the same.

clippy_lints/src/matches/single_match.rs
```diff
@@ -140,74 +140,128 @@ fn check_opt_like<'a>(
...
+}
+
+fn contains_only_known_enums(cx: &LateContext<'_>, pat: &Pat<'_>) -> bool {
+    match pat.kind {
+        PatKind::Path(..) => is_known_enum(cx, cx.typeck_results().pat_ty(pat)),
+        PatKind::TupleStruct(..) => {
+            let ty = cx.typeck_results().pat_ty(pat);
+            (contains_only_wilds(pat) || contains_single_binding(pat)) && is_known_enum(cx, ty)
+        },
+        _ => false,
```
camsteffen, line 162: `Struct` is still missing here. There should be a test for it.

clippy_lints/src/matches.rs
camsteffen: > Yes, it seems logical. But for some reasons we have [the following check](https://github.com/rust-lang/rust-clippy/tree/51045172c1313ed912742a521cd02b639c120fc4/clippy_lints/src/matches/single_match.rs#L29) for the `els` branch. And such cases are explicitly rejected.

That check is there because we only want to lint if the second arm is `.. => ()` (or `{}`), so I that shouldn't be a problem for `y @ _ => (),`?

clippy_lints/src/matches/single_match.rs
```diff
@@ -140,74 +140,128 @@ fn check_opt_like<'a>(
...
-                s.print_qpath(path, false);
-            });
-            acc.push((path, ty));
-            true
+/// Returns false if the patterns exhaustively match an enum.
+fn check_exhaustive<'a>(cx: &LateContext<'a>, left: &Pat<'_>, right: &Pat<'_>, ty: Ty<'a>) -> bool {
+    let left = peel_subpatterns(left);
+    let right = peel_subpatterns(right);
+    match (&left.kind, &right.kind) {
+        (PatKind::Wild, _) | (_, PatKind::Wild) => true,
```
camsteffen, line 178: Wouldn't this work?
```suggestion
        (PatKind::Wild | PatKind::Binding(..), _) | (_, PatKind::Wild | PatKind::Binding(..)) => true,
```

camsteffen (changes requested):

clippy_lints/src/matches/single_match.rs
jubnzv: We need to handle wildcards only when processing an `els` branch, because we cannot bind identifier to it.

For example, we don't want to lint the following case:
```rust
match s {
  a => (),
  a => (), // don't lint, we cannot bind `a` to the `else` branch
}
```

clippy_lints/src/matches/single_match.rs
jubnzv: No, because we should lint bindings iff the `else` branch (`right` pattern) contains wildcards only.

clippy_lints/src/matches.rs
jubnzv: > That check is there because we only want to lint if the second arm is .. => () (or {}), so I that shouldn't be a problem for y @ _ => (),?

Yes, and in that case we'll get the recursion error in `rustfix`.

clippy_lints/src/matches/single_match.rs
camsteffen: I think we can lint that case. The binding will be unused in the second branch.

jubnzv: It doesn't work in such cases. We don't actually know, which branch will be `if` and which is `els`, so after this change we'll get a lot of false positives. I think, this out of scope of this MR, because it  adds support for `Struct` and `TupleStruct` patterns.

If we need to improve precision of this lint, this should be a different MR.

jubnzv:
@camsteffen friendly ping.
Could you please take a look at latest comments when you have time? I'm not sure we want to tweak any branch processing logic in the PR related to adding `Struct` and `TupleStruct` types.

jubnzv:
Closed, since there is no response from maintainers.
Feel free to pick it, this PR is ready to merge.

jubnzv:
@flip1995 No problem, I understand. I don't have time for any side projects right now either, so I prefer to close it.

Issue #3999, opened 2019-04-18
[lint request] unspecified_repr_Rust_transmute


gnzlbg:
There are some `repr(Rust)` types (structs, tuples, tuple structs, enums, unions, ...) for which we guarantee the layout (e.g. `Option<NonZeroU32>`), but we don't guarantee the layout of most `repr(Rust)` types, their layout is unspecified and allowed to change.

We should add a lint that warns about these, but I think we should try to at least compile a list of cases for which we do guarantee layout, and not warn on those. 

cc @eddyb

oli-obk:
Maybe we can "document" these in the `ty::Layout` so semverver can use the same logic for figuring out when changing a type is a breaking change (moving from guaranteed layout to not guaranteed layout on a public type is a breaking change)

Issue #546, opened 2016-01-13
[lint idea] dangerous transmutes


durka:
Along the lines of `useless_transmute`, it seem plausible to warn on `transmute::<X, Y>(x)` where `X` and `Y` aren't both primitive types or marked `#[repr(C)]`, since as @ubsan keeps reminding us this is likely to cause UB.

ubsan:
@durka heh :)

The only valid transmutes of a `repr(Rust)` struct that I know of are in `std::repr`, so it should(?) be fairly simple to special case those.

Ryman:
I would love to see something like this, it's very easy to forget about and have seen it come up a number of times.

durka:
@ubsan I expect running clippy on libstd would trigger more lints than that :)

oli-obk:
also casting a `#[repr(C)]` struct with padding to anything where the padding becomes accessible is UB

durka:
Does clippy have access to ask rustc whether a struct has any padding, or
would the logic have to be duplicated?

On Thu, May 12, 2016 at 11:15 AM, Oliver Schneider <notifications@github.com

oli-obk:
clippy can ask rustc, at least miri can do it (since a few weeks ago), so clippy should be able to do it, too.

felix91gr:
Upon inspection, I believe this issue is still relevant:

1. The Rustonomicon states that layouts are only known to be the same if both types are [either both `repr(C)` or both `repr(transparent)`](https://doc.rust-lang.org/nomicon/transmutes.html).
2. The last attempt at a lint for this was [@Jarcho's work](https://rust-lang.github.io/rust-clippy/master/index.html#transmute_undefined_repr), which is now in Nursery due to known issues which have been investigated and documented at https://github.com/rust-lang/rust-clippy/issues/8496.

It might not be a `good-first-issue` though. Maybe `E-hard`?

Jarcho:
There are only two things holding `transumte_undefined_repr` back. One is a trivial one line change I never made, and the other is a question about whether fat pointers should be linted with it.

felix91gr:
> There are only two things holding `transumte_undefined_repr` back. One is a trivial one line change I never made, and the other is a question about whether fat pointers should be linted with it.

Fat pointers, like... `dst`, right? 🤔 lemme check the Rustonomicon.

Okay, here's what I've got:

felix91gr:
[...]

PR #8398, opened 2022-02-06
Add lint `transmute_undefined_repr`

Jarcho:
Partially implements #3999 and #546

This doesn't consider `enum`s at all right now as those are going to be a pain to deal with. This also allows `#[repr(Rust)]` structs with only one non-zero sized fields. I think those are technically undefined when transmuted.

changelog: Add lint `transmute_undefined_repr`


llogiq:
Just one very small nit, otherwise this looks fine. Great work, as always.

llogiq:
Thank you! @bors r+

Issue #6680, opened 2021-02-05
[Tracking issue] Rearrange location of some of the existing lints


magurotuna:
Currently, the lint rules grouped as "methods" reside in `clippy_lints/src/methods` as there are a number of rules belonging to the "methods" group.
However, although the other groups such as "loops", "types", "transmute" and "misc" also have many relavent rules, they reside not in `clippy_lints/src/<group_name>` but in `clippy_lints/src`.
It would be great to have these groups located under `<group_name>` directory just like "methods". So I open this issue to do this refactoring and to keep track of it.

- [ ] loops
- [ ] types
- [ ] transmute
- [ ] misc

How to rearrange is like the following, proposed by @flip1995 on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/lint.20file.20location.20clippy.236643/near/225318775)

> I would keep the declare_clippy_lint! in the mod.rs Best case would be that mod.rs only contains the lint definitions and the LintPass impl. Utility functions of the module should probably go in methods/utils.rs and then reexported for the module in mod.rs.

CC @nahuakang @camsteffen

nahuakang:
@magurotuna @camsteffen Shall we kick this off sometime this month? I could focus on `loops` since I just came from it (and there are only 18 lints). But I think for `methods` we'd need to split carefully since it has 50+ lints :thinking:

flip1995:
Also feel free to split this not only in multiple commits, but also in multiple PRs.

camsteffen:
@nahuakang Sure! Frankly I don't want to promise anything from me at this point. For organization, I would recommend we comment here on what group you're currently working on. But you don't have to do an entire group.

phansch:
> clippy_lints/src/<group_name>/<lint_name>.rs

If `group_name` refers to lint groups like `restriction` or `pedantic`, it's worth noting that these change quite often compared to lint names. So, if someone changes the lint group of a lint they will have to move it to the correct directory as well, adding some extra work. Not sure if that's fine?

If it just refers to unrelated group names, ignore what I said =)

nahuakang:
@phansch The `group_name` would be `methods` or `loops` so that new contributors have an easier time finding lint's or understanding the structure of the repo 😁 You are right that it shouldn't be the lint group categories.

magurotuna:
I'll work on "transmute" group :)

Y-Nak:
I'll work on `types`!

[...]

PR #8400, opened 2022-02-06
Split matches

Jarcho:
Part of #6680

changelog: None


clippy_lints/src/matches/mod.rs
```diff
@@ -0,0 +1,716 @@
...
+        }
+
+        if let ExprKind::Match(ex, arms, MatchSource::Normal) = expr.kind {
+            single_match::check(cx, ex, arms, expr);
+            match_bool::check(cx, ex, arms, expr);
+            overlapping_arms::check(cx, ex, arms);
+            match_wild_err_arm::check(cx, ex, arms);
+            match_wild_enum::check(cx, ex, arms);
+            match_as_ref::check(cx, ex, arms, expr);
+            check_wild_in_or_pats(cx, arms);
```
flip1995, line 626: There are some lints left in `mod.rs`. I guess this is because the linting code is too small to extract them into their own file?

flip1995 (commented):
All the current lint extractions LGTM. Thanks for splitting this up!

clippy_lints/src/matches/mod.rs
Jarcho: I actually missed `check_wild_in_or_pats`. It's small enough it doesn't really matter though. The other two I just left because they were in different check functions. Could all be moved out though.

flip1995: Let's move them all out then. get the `mod.rs` file completely cleaned up, while you're at it :+1:

Jarcho: And done.

PR #8414, opened 2022-02-11
Optimize `redundant_clone`

### LINKED ISSUE #8412: redundant_clone is too slow
### Description

I noticed that a slowness of clippy (around 3 minutes to run) while writing a code that has some [code generated by a macro](https://github.com/MiSawa/xq/blob/ed1d41e2593693c25608de6ae878a0c734723a4b/src/lang2/lexer.rs#L107). It turned out that this is due to `clippy::redundant_clone` lint, especially because of [TransitiveRelation::add](https://github.com/rust-lang/rust/blob/e273fca380c5d28bc32b25ac1a885c61d4c5e75e/compiler/rustc_data_structures/src/transitive_relation.rs) method call.
This method takes `Θ(t)` where `t` is the number of distinct `add` calls so far, so takes `Θ(m^2)` in total where `m` is the total number of relations.

Though use of this `TransitiveRelation` seems to be overkill, because we only use it as "add a lot of relations, then get all transitive relations" (i.e. as a static data structure) where the `TransitiveRelation` can handle interleaved "add" and "get" (i.e. an incremental data structure). To mitigate the problem, we can implement a simpler version of `TransitiveRelation` that takes (expected amortized) `O(1)` for `add` and `O(sum-of-num-of-reachable-edges)` for "get all transitive relations".

Reproduce steps:
```sh
$ git clone https://github.com/MiSawa/xq.git
$ cd xq
$ git switch -d ed1d41e2593693c25608de6ae878a0c734723a4b
$ cargo clippy
```

### Version

```text
rustc 1.60.0-nightly (e646f3d2a 2022-02-10)
binary: rustc
commit-hash: e646f3d2a9541952310778288854943678738ea9
commit-date: 2022-02-10
release: 1.60.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

MiSawa:
Fixes #8412 

changelog: none

### DIFF
--- a/clippy_lints/src/redundant_clone.rs
+++ b/clippy_lints/src/redundant_clone.rs
@@ -762,7 +762,6 @@ impl TransitiveRelation {
 
     fn reachable_from(&self, a: mir::Local, domain_size: usize) -> HybridBitSet<mir::Local> {
         let mut seen = HybridBitSet::new_empty(domain_size);
-        seen.insert(a);
         let mut stack = vec![a];
         while let Some(u) = stack.pop() {
             if let Some(edges) = self.relations.get(&u) {
@@ -773,7 +772,6 @@ impl TransitiveRelation {
                 }
             }
         }
-        seen.remove(a);
         seen
     }
 }
--- a/clippy_lints/src/redundant_clone.rs
+++ b/clippy_lints/src/redundant_clone.rs
@@ -3,7 +3,7 @@ use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::{has_drop, is_copy, is_type_diagnostic_item, walk_ptrs_ty_depth};
 use clippy_utils::{fn_has_unsatisfiable_preds, match_def_path, paths};
 use if_chain::if_chain;
-use rustc_data_structures::{fx::FxHashMap, transitive_relation::TransitiveRelation};
+use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::FnKind;
 use rustc_hir::{def_id, Body, FnDecl, HirId};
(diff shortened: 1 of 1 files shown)


MiSawa:
@giraffate Hi, I'm wondering if I'm on the right track about getting the change reviewed and merged. e.g. is there something that I should be registering so that this will be queued? Don't mean to rush you. I'm new to this repo, so just unsure about the process and wondering if there are action items for me.

giraffate:
@MiSawa Sorry for the late review, and thanks for reminding me. It's just that I missed it. And, I'm not sure if there is anything wrong with this, so I will have another members take a look just to be sure.

@rust-lang/clippy Can anyone review this?

MiSawa:
Don't be sorry, thank you for moving this forward!

clippy_lints/src/redundant_clone.rs
```diff
@@ -766,3 +750,30 @@ impl PossibleBorrowerMap<'_, '_> {
...
+        while let Some(u) = stack.pop() {
+            if let Some(edges) = self.relations.get(&u) {
+                for &v in edges {
+                    if seen.insert(v) {
+                        stack.push(v);
+                    }
+                }
+            }
+        }
+        seen.remove(a);
```
smoelius, line 776: Is it certain that `a` cannot be inserted on line 770?

MiSawa: Thank you for the review! I was imagined that borrow-borrower relation is a DAG (directed acyclic graph), but I'm actually uncertain. I'll remove L:765 and L:776 so it would work even if it wasn't a DAG (with less than 2x traversal which is just a constant factor).

flip1995:
Because this lint operates on the MIR, I kindly ask @oli-obk to give this a look.

oli-obk:
@bors r+

yea, this makes a lot of sense. The original datastructure is tuned to the access patterns of lifetime lookups and overkill here.

MiSawa:
Awesome, thank you all!

Issue #8417, opened 2022-02-12
transmute_undefined_repr is wrong, transmute does not require a defined layout


dtolnay:
### Summary

The description of transmute_undefined_repr says:

> **What it does**
> Checks for transmutes either to or from a type which does not have a defined representation.
>
> **Why is this bad?**
> The results of such a transmute are not defined.

I think this is not correct. Transmute does not require either side to have a defined representation. It only requires that both sides have the same representation, which can be guaranteed without defining what that representation is.

### Lint Name

transmute_undefined_repr

### Reproducer

```rust
#[repr(C)]
struct Wrapper(String);

fn main() {
    let string = String::new();
    let _ = unsafe { std::mem::transmute::<String, Wrapper>(string) };
}
```

```console
$ cargo clippy

error: transmute from `std::string::String` which has an undefined layout
 --> src/main.rs:6:22
  |
6 |     let _ = unsafe { std::mem::transmute::<String, Wrapper>(string) };
  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[deny(clippy::transmute_undefined_repr)]` on by default
  = note: the contained type `std::vec::Vec<u8>` has an undefined layout
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_undefined_repr
```

### Version

```text
rustc 1.60.0-nightly (e789f3a3a 2022-02-11)
binary: rustc
commit-hash: e789f3a3a3d96ebf99b7bbd95011527e5be32a11
commit-date: 2022-02-11
host: x86_64-unknown-linux-gnu
release: 1.60.0-nightly
LLVM version: 13.0.0
```

Jarcho:
Hilariously it works if you remove `#[repr(C)]`. I can fix it later today.

PR #8418, opened 2022-02-12
Downgrade transmute_undefined_repr to nursery

dtolnay:
Reason: #8417. I am skeptical of this lint but maybe there is a narrower subset of types on which it is useful, so keeping it for now but moving to nursery for further development.

---

changelog: Remove [`transmute_undefined_repr`] from default set of enabled lints


llogiq:
Thank you! @bors r+

Issue #6446, opened 2020-12-12
clippy::await_holding_lock reports explicitly dropped mutex guards as if they weren't dropped


jazevedo620:
I tried this code:

```rust
#[deny(clippy::pedantic)]
pub async fn doesnt_work() {
    use std::collections::HashSet;
    use std::sync::Mutex;
    use std::time::Duration;

    let lock: Mutex<HashSet<u8>> = Mutex::new(HashSet::new());
    let mut set = lock.lock().unwrap();
    set.insert(0);
    drop(set);
    tokio::time::sleep(Duration::from_secs(1)).await;
}
```

I expected to see this happen: it should pass Clippy with no linting errors because the `MutexGuard` is explicitly dropped before the await point with a call to `std::mem::drop`.

Instead, this happened: Clippy failed the lint check because of a violation of `clippy::await_holding_lock`:

```
$ cargo clippy
...
error: this MutexGuard is held across an 'await' point. Consider using an async-aware Mutex type or ensuring the MutexGuard is dropped before calling await.
  --> src/lib.rs:8:9
   |
8  |     let mut set = lock.lock().unwrap();
   |         ^^^^^^^
   |
note: the lint level is defined here
  --> src/lib.rs:1:8
   |
1  | #[deny(clippy::pedantic)]
   |        ^^^^^^^^^^^^^^^^
   = note: `#[deny(clippy::await_holding_lock)]` implied by `#[deny(clippy::pedantic)]`
note: these are all the await points this lock is held through
  --> src/lib.rs:8:5
   |
8  | /     let mut set = lock.lock().unwrap();
9  | |     set.insert(0);
10 | |     drop(set);
11 | |     tokio::time::sleep(Duration::from_secs(1)).await;
12 | | }
   | |_^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_lock

error: aborting due to previous error
```

Running with `--verbose` does not provide any additional information.

Additionally, of note, the following code does pass Clippy:

```rust
#[deny(clippy::pedantic)]
pub async fn doesnt_work() {
    use std::collections::HashSet;
    use std::sync::Mutex;
    use std::time::Duration;

    let lock: Mutex<HashSet<u8>> = Mutex::new(HashSet::new());
    {
        let mut set = lock.lock().unwrap();
        set.insert(0);
    }
    tokio::time::sleep(Duration::from_secs(1)).await;
}
```

### Meta

[...]

PR #8419, opened 2022-02-12
Fix `await_holding_lock` not linting `parking_lot` Mutex/RwLock

flip1995:
This adds tests for `RwLock` and `parking_lot::{Mutex, RwLock}`, which were added before in https://github.com/rust-lang/rust-clippy/commit/2dc8c083f54454ca87bb09d691577eada2d23539, but never tested in UI tests. I noticed this while reading [fasterthanli.me](https://fasterthanli.me/articles/a-rust-match-made-in-hell) latest blog post, complaining that Clippy doesn't catch this for `parking_lot`. (Too many people read his blog, he's too powerful)

Some more things:
- Adds a test for #6446
- Improves the lint message

changelog: [`await_holding_lock`]: Now also lints for `parking_lot::{Mutex, RwLock}`


tests/ui/await_holding_lock.rs
```diff
@@ -1,64 +1,194 @@
 #![warn(clippy::await_holding_lock)]
 
-use std::sync::Mutex;
+extern crate parking_lot;
```
llogiq, line 3: Is this needed because we're in a UI test? Elsewhere I'd just `use`.

llogiq (commented):
Not the approver, but this looks good in general. One point: if this still has bugs, we should document them in the lint docs. Otherwise should we re-group this to `suspicious` so it at least warns by default?

tests/ui/await_holding_lock.rs
flip1995: Not sure. I just copied this from the `let_underscore` test, that also tests for the parking_lot crate. I'll try if it also works with `use`, now that we're using the 2018/2021 (?) edition.

flip1995:
> One point: if this still has bugs, we should document them in the lint docs. 

I think the bug for which I added a test is already documented under `Known Problems`, but the documentation could be improved.

> Otherwise should we re-group this to `suspicious` so it at least warns by default?

I'll try to do that and run the `lintcheck` tool.

flip1995:
> I'll try to do that and run the `lintcheck` tool.

I ran the lintcheck tool on the top 279 downloaded crates on crates.io, where grepping for `.await` gave 7645 matches. The `await_holding_lock` lint didn't trigger once. So IMO we could move it to `suspicious`.

Actual changes in this rebase to previous version:
- Moved lints to `suspicious`
- removed `extern crate` from test
- Improved documentation (This is now doc-tested)

llogiq:
Great job! Fingers crossed that it'll work as great in production.

@bors r+

Issue #8390, opened 2022-02-03
New Lint: function argument only used in recursion


buttercrab:
### What it does

Arguments that are used only in recursion to itself.
Since it is used, the warning `unused_variables` does not come out, but it is useless.

### Lint Name

only_used_in_recursion

### Category

perf

### Advantage

- Lowers the complexity of the function.
- Better performance when the useless argument have some operation (like below example)

### Example

```rust
fn f(a: usize, b: usize) -> usize {
    if a == 0 {
        1
    } else {
        f(a - 1, b + 1)
    }
}

fn main() {
    print!("{}", f(1, 1));
}
```

Could be written as:

```rust
fn f(a: usize) -> usize {
    if a == 0 {
        1
    } else {
        f(a - 1)
    }
}

fn main() {
    print!("{}", f(1));
}
```

buttercrab:
Can I claim this if it looks good?

PR #8422, opened 2022-02-12
new lint: `only_used_in_recursion`

buttercrab:
changed:
- added `only_used_in_recursion`.
- fixed code that variables are only used in recursion.
- this would not lint when `unused_variable`

This fixes: #8390 

-----

changelog: add lint [`only_used_in_recursion`]

### DIFF
--- a/clippy_lints/src/only_used_in_recursion.rs
+++ b/clippy_lints/src/only_used_in_recursion.rs
@@ -126,29 +126,36 @@ impl<'tcx> LateLintPass<'tcx> for OnlyUsedInRecursion {
                 }
             }
 
-            let mut pre_order = FxHashMap::default();
-
-            visitor.graph.iter().for_each(|(_, next)| {
-                next.iter().for_each(|i| {
-                    *pre_order.entry(*i).or_insert(0) += 1;
-                });
-            });
-
             for (id, span, ident) in param_span {
                 // if the variable is not used in recursion, it would be marked as unused
-                if !visitor.has_side_effect.contains(&id)
-                    && *pre_order.get(&id).unwrap_or(&0) > 0
-                    && visitor.graph.contains_key(&id)
-                {
-                    span_lint_and_sugg(
-                        cx,
-                        ONLY_USED_IN_RECURSION,
-                        span,
-                        "parameter is only used in recursion with no side-effects",
-                        "if this is intentional, prefix with an underscore",
-                        format!("_{}", ident.name.as_str()),
-                        Applicability::MaybeIncorrect,
-                    );
+                if !visitor.has_side_effect.contains(&id) {
+                    let mut queue = VecDeque::new();
+                    let mut visited = FxHashSet::default();
+
+                    queue.push_back(id);
+
+                    while let Some(id) = queue.pop_front() {
+                        if let Some(next) = visitor.graph.get(&id) {
+                            for i in next {
+                                if !visited.contains(i) {
+                                    visited.insert(id);
+                                    queue.push_back(*i);
+                                }
+                            }
+                        }
+                    }
+
+                    if visited.contains(&id) {
+                        span_lint_and_sugg(
+                            cx,
+                            ONLY_USED_IN_RECURSION,
+                            span,
+                            "parameter is only used in recursion with no side-effects",
+                            "if this is intentional, prefix with an underscore",
+                            format!("_{}", ident.name.as_str()),
+                            Applicability::MaybeIncorrect,
+                        );
+                    }
                 }
             }
         }
@@ -200,7 +207,7 @@ impl<'tcx> Visitor<'tcx> for SideEffectVisit<'tcx> {
             StmtKind::Local(Local {
                 pat, init: Some(init), ..
             }) => {
-                self.visit_pat_expr(pat, init);
+                self.visit_pat_expr(pat, init, false);
             },
             StmtKind::Item(i) => {
                 let item = self.ty_ctx.hir().item(i);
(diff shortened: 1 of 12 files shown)


clippy_lints/src/only_used_in_recursion.rs
```diff
@@ -0,0 +1,586 @@
...
+use rustc_middle::ty;
+use rustc_middle::ty::{Ty, TyCtxt, TypeckResults};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::kw;
+use rustc_span::symbol::Ident;
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for arguments that is only used in recursion with no side-effects.
```
est31, line 24:
```suggestion
    /// Checks for arguments that are only used in recursion with no side-effects.
```

clippy_lints/src/only_used_in_recursion.rs
```diff
@@ -0,0 +1,563 @@
...
+            }
+
+            for (id, span, ident) in param_span {
+                if !visitor.has_side_effect.contains(&id) {
+                    span_lint_and_sugg(
+                        cx,
+                        ONLY_USED_IN_RECURSION,
+                        span,
+                        "parameter is only used in recursion with no side-effects",
+                        "if this is intentional, prefix with an underscore",
```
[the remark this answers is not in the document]
buttercrab: If the function is from trait implementation, the argument needs to be there. Also, I thought since `unused_variables` suggests prefixing with an underscore, this way would be better for consistency.

clippy_lints/src/only_used_in_recursion.rs
```diff
@@ -0,0 +1,629 @@
...
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for arguments that are only used in recursion with no side-effects.
+    /// The arguments can be involved in calculations and assignments but as long as
+    /// the calculations have no side-effects (function calls or mutating dereference)
+    /// and the assigned variables are also only in recursion, it is useless.
+    ///
+    /// ### Why is this bad?
+    /// The could contain a useless calculation and can make function simpler.
```
llogiq, line 30:
```suggestion
    /// The could contain a useless calculation and can make the function simpler.
```

llogiq (commented):
This looks mostly good. Small documentation nits from my side, and I'd like to see a test case with a trait method / trait method impl (the latter should not lint).

Also I'd add the drawback that if the method is publicly visible, following the lint would be API-breaking.

buttercrab:
@llogiq I forgot about traits. I'll add some code to support trait impls.

llogiq:
Please note that we should *not* lint trait *impls* (because the argument is part of the trait definition and the impl cannot change that), but may lint definitions.

tests/ui/only_used_in_recursion.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn method(a: usize, b: usize) -> usize {
+        if a == 0 { 1 } else { A::method(a - 1, b - 1) }
+    }
+
+    fn method2(&self, a: usize, b: usize) -> usize {
+        if a == 0 { 1 } else { self.method2(a - 1, b + 1) }
+    }
+}
+
+trait B {
```
llogiq, line 73: How about adding an implemented method (which should be linted) here?

llogiq (commented):
One small suggestion, otherwise this looks merge-worthy.

tests/ui/only_used_in_recursion.rs
buttercrab: Since trait definition can be in other crate and only some of impls are linted and others are not, I think that current way is better in my opinion.

llogiq: We only lint the current crate. So if it contains a `trait B { fn rec(a: usize) { if a > 0 { rec(a - 1) } else { 42 } } .. }`, we can opt to lint `B::rec`, as the current crate defines and implements it.

buttercrab: I see. I'll work on it.

tests/ui/only_used_in_recursion.rs
```diff
@@ -0,0 +1,107 @@
...
+        if a == 0 { 1 } else { A::hello(a - 1, b + 1) }
+    }
+
+    fn hello2(&self, a: usize, b: usize) -> usize {
+        if a == 0 { 1 } else { self.hello2(a - 1, b + 1) }
+    }
+}
+
+trait C {
+    fn hello(a: usize, b: usize) -> usize {
```
llogiq, line 90: Why is this not caught?

buttercrab: My bad. I'll fix it.

llogiq:
Thank you for this PR and sorry for drawing out this process for so long.

@bors r+

buttercrab:
@llogiq I've changed the PR body due to error. Could you try again?

llogiq:
Wouldn't you need to push something first?

buttercrab:
It failed on [changelog test](https://github.com/rust-lang/rust-clippy/runs/5522331585?check_suite_focus=true).

buttercrab:
Should I need to push a commit even though it failed on changelog test?

llogiq:
I see. @bors r+

buttercrab:
@llogiq I fixed the code. Could you try again?

llogiq:
Sure! @bors r+

PR #8429, opened 2022-02-14
Document `pub` requirement for `new_without_default` lint

### LINKED ISSUE #8415: Docs for `new_without_default` don't mention `pub` requirement
### Description

The [docs](https://rust-lang.github.io/rust-clippy/master/#new_without_default) for the `new_without_default` lint include the following example:

```rust
struct Foo(Bar);

impl Foo {
    fn new() -> Self {
        Foo(Bar::new())
    }
}
```

If we run Clippy against this example (plus a definition for `Bar`), we get no warnings for `new_without_default`, see the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b9418408c963904581f81c1043c3739d).

This seems to be because the warning will only trigger if both the type and the `new` method are `pub`. See [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d403757c51e896afb1e1a731767d39ef) similar example which does get the `new_without_default` warning.

The docs should be clear on this `pub` requirement, and the example should reflect this.

### Version

```text
binary: rustc
commit-hash: db9d1b20bba1968c1ec1fc49616d4742c1725b4b
commit-date: 2022-01-20
release: 1.58.1
LLVM version: 13.0.0
```

### Additional Labels

@rustbot label +A-documentation

nsunderland1:
fixes #8415

Also adds some UI tests that ensure that `pub` is required on both the struct _and_ the field. The only thing I'm not sure about is that the lint actually [checks](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/new_without_default.rs#L102) if `new` is _reachable_, not _public_. To the best of my understanding, both the struct and the method need to be public for the method to be reachable for external crates (I certainly didn't manage to craft a counterexample).

changelog: Document `pub` requirement for ``[`new_without_default`]`` lint.

### DIFF
--- a/clippy_lints/src/new_without_default.rs
+++ b/clippy_lints/src/new_without_default.rs
@@ -13,7 +13,7 @@ use rustc_span::sym;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for types with a `fn new() -> Self` method and no
+    /// Checks for public types with a `pub fn new() -> Self` method and no
     /// implementation of
     /// [`Default`](https://doc.rust-lang.org/std/default/trait.Default.html).
     ///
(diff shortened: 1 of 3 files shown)


llogiq:
Thank you! I wouldn't worry with the distinction between public and exported here. You could in theory have a `struct X; pub use X`, but I don't know anyone coding that way.

@bors r+

Issue #8091, opened 2021-12-07
`redundant_closure` FP when functions returns `!`


austinabell:
### Summary

Unsure if directly related to https://github.com/rust-lang/rust-clippy/issues/8073 so I will open a separate issue. Feel free to close if it is handled in the same cases, just want to be sure this variant is handled as well. Clippy suggests removing the redundant closure but it cannot be removed because the closure return is different.

### Reproducer

I tried this code:

```rust
fn exit() -> ! {
    panic!()
}

fn main() {
    let value = Some(8u8);
    println!("{}", value.unwrap_or_else(|| exit()));
}
```

I expected to see this happen:

No clippy warning since the suggestion doesn't compile

Instead, this happened:

```
warning: redundant closure
 --> src/main.rs:7:41
  |
7 |     println!("{}", value.unwrap_or_else(|| exit()));
  |                                         ^^^^^^^^^ help: replace the closure with the function itself: `exit`
  |
  = note: `#[warn(clippy::redundant_closure)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure

warning: `playground` (bin "playground") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 1.01s
```

### Version

```text
rustc 1.57.0 (f1edd0429 2021-11-29)
binary: rustc
commit-hash: f1edd0429582dd29cccacaf50fd134b05593bd9c
commit-date: 2021-11-29
host: x86_64-unknown-linux-gnu
release: 1.57.0
LLVM version: 13.0.0
```

Issue #7812, opened 2021-10-12
redundant_closure: false positive for never return type


Bromeon:
Lint name: `redundant_closure`

With this code:
```rust
fn explode() -> ! {
    panic!("never returns")
}

fn main() {
    Some(32).unwrap_or_else(|| explode());
}
```
Clippy emits the following warning:
```rs
warning: redundant closure
 --> src\main.rs:6:29
  |
6 |     Some(32).unwrap_or_else(|| explode());
  |                             ^^^^^^^^^^^^ help: replace the closure with the function itself: `explode`
  |
  = note: `#[warn(clippy::redundant_closure)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
  ```

---

However, I cannot inline the closure and use the function pointer directly -- this code:
```rs
fn explode() -> ! {
    panic!("never returns")
}

fn main() {
    Some(32).unwrap_or_else(explode);
}
```
fails to compile with:
```rs
error[E0271]: type mismatch resolving `<fn() -> ! {explode} as std::ops::FnOnce<()>>::Output == {integer}`
 --> src\main.rs:6:14
  |
6 |     Some(32).unwrap_or_else(explode);
  |              ^^^^^^^^^^^^^^ expected `!`, found integer
  |
  = note: expected type `!`
             found type `{integer}`

For more information about this error, try `rustc --explain E0271`
```

---

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.57.0-nightly (5b210643e 2021-10-11)
binary: rustc
commit-hash: 5b210643ebf2485aafdf2494de8cf41941a64e95
commit-date: 2021-10-11
host: x86_64-pc-windows-msvc
release: 1.57.0-nightly
LLVM version: 13.0.0
```

Related, but not identical issues ( `redundant_closure` false positives):
* https://github.com/rust-lang/rust-clippy/issues/1608
* https://github.com/rust-lang/rust-clippy/issues/5594
* https://github.com/rust-lang/rust-clippy/issues/5939
* https://github.com/rust-lang/rust-clippy/issues/7497

PR #817, opened 2016-03-30
Fix FP in `REDUNDANT_CLOSURE` with divergent functions

mcarton:
Fix #815.

PR #8431, opened 2022-02-14
`redundant_closure` fix FP on coerced closure

### LINKED ISSUE #8416: clippy::redundant_closure: false positive when function returns ! but closure doesn't
### Summary

For tests I have a mock for `std::process::exit()` that needs to return, so when `std::process::exit()` is passed in as the callback I wrap it in a closure to change the return type from ! to ().

### Lint Name

clippy::redundant_closure

### Reproducer

I tried this code:

```rust
fn do_thing_with_exit<F: FnMut(i32)>(mut exit_provider: F) {
    //Do thing
    exit_provider(0);
}

fn main() {
    // Example test code:
    let mut ret_code = -1i32;
    do_thing_with_exit(|v| { ret_code = v; });
    assert_eq!(ret_code, 0);

    // Example non-test code:
    do_thing_with_exit(|v| std::process::exit(v));
}
```

I saw this happen:

```
warning: redundant closure
  [--> src/main.rs:13:24
13 |     do_thing_with_exit(|v| std::process::exit(v));
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `std::process::exit`
   |
   = note: `#[warn(clippy::redundant_closure)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
```

I applied the suggestion:

```
fn do_thing_with_exit<F: FnMut(i32)>(mut exit_provider: F) {
    //Do thing
    exit_provider(0);
}

fn main() {
    // Example test code:
    let mut ret_code = -1i32;
    do_thing_with_exit(|v| { ret_code = v; });
    assert_eq!(ret_code, 0);

    // Example non-test code:
    do_thing_with_exit(std::process::exit);
}
```

This fails to compile with the error:

```
error[[E0271]](https://doc.rust-lang.org/stable/error-index.html#E0271): type mismatch resolving `<fn(i32) -> ! {exit} as FnOnce<(i32,)>>::Output == ()`
  [--> src/main.rs:13:5
13 |     do_thing_with_exit(std::process::exit);
   |     ^^^^^^^^^^^^^^^^^^ expected `()`, found `!`
   |
   = note: expected unit type `()`
                   found type `!`
note: required by a bound in `do_thing_with_exit`
  [--> src/main.rs:1:26
1  | fn do_thing_with_exit<F: FnMut(i32)>(mut exit_provider: F) {
   |                          ^^^^^^^^^^ required by this bound in `do_thing_with_exit`

For more information about this error, try `rustc --explain E0271`.
```

### Version

```text
I used 1.58.1 stable from play.rust-lang.org.

The clippy version is 0.1.60 (2022-02-10 e646f3d)
```

### Additional Labels

_No response_

dswij:
Closes https://github.com/rust-lang/rust-clippy/issues/8416,
Closes https://github.com/rust-lang/rust-clippy/issues/7812
Closes https://github.com/rust-lang/rust-clippy/issues/8091

~~Seems like this is fixed in #817 and regressed.~~

Ignore coerced closure false positives, but this will lead to some false negatives on resolvable generics. IMO, this is still an overall improvement

changelog: [`redundant_closure`] ignores coerced closure

### DIFF
--- a/clippy_lints/src/eta_reduction.rs
+++ b/clippy_lints/src/eta_reduction.rs
@@ -3,7 +3,7 @@ use clippy_utils::higher::VecArgs;
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::usage::local_used_after_expr;
-use clippy_utils::{higher, path_to_local, path_to_local_id};
+use clippy_utils::{higher, is_adjusted, path_to_local, path_to_local_id};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::def_id::DefId;
@@ -103,6 +103,7 @@ impl<'tcx> LateLintPass<'tcx> for EtaReduction {
         let closure_ty = cx.typeck_results().expr_ty(expr);
 
         if_chain!(
+            if !is_adjusted(cx, &body.value);
             if let ExprKind::Call(callee, args) = body.value.kind;
             if let ExprKind::Path(_) = callee.kind;
             if check_inputs(cx, body.params, args);
@@ -144,6 +145,7 @@ impl<'tcx> LateLintPass<'tcx> for EtaReduction {
         );
 
         if_chain!(
+            if !is_adjusted(cx, &body.value);
             if let ExprKind::MethodCall(path, args, _) = body.value.kind;
             if check_inputs(cx, body.params, args);
             let method_def_id = cx.typeck_results().type_dependent_def_id(body.value.hir_id).unwrap();
--- a/tests/ui/eta.stderr
+++ b/tests/ui/eta.stderr
@@ -25,12 +25,6 @@ LL |     all(&[1, 2, 3], &&2, |x, y| below(x, y)); //is adjusted
    |                          ^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `below`
 
 error: redundant closure
-  --> $DIR/eta.rs:40:27
-   |
-LL |     let e = Some(1u8).map(|a| divergent(a));
-   |                           ^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `divergent`
-
-error: redundant closure
   --> $DIR/eta.rs:41:27
    |
 LL |     let e = Some(1u8).map(|a| generic(a));
@@ -122,11 +116,5 @@ error: redundant closure
 LL |         Some(1).map(|n| in_loop(n));
    |                     ^^^^^^^^^^^^^^ help: replace the closure with the function itself: `in_loop`
 
-error: redundant closure
-  --> $DIR/eta.rs:236:21
-   |
-LL |     map_str_to_path(|s| s.as_ref());
-   |                     ^^^^^^^^^^^^^^ help: replace the closure with the method itself: `std::convert::AsRef::as_ref`
-
-error: aborting due to 21 previous errors
+error: aborting due to 19 previous errors
 
--- a/tests/ui/eta.fixed
+++ b/tests/ui/eta.fixed
@@ -37,7 +37,7 @@ fn main() {
     }
 
     // See #815
-    let e = Some(1u8).map(divergent);
+    let e = Some(1u8).map(|a| divergent(a));
     let e = Some(1u8).map(generic);
     let e = Some(1u8).map(generic);
     // See #515
@@ -233,7 +233,7 @@ fn late_bound_lifetimes() {
     {
     }
     map_str(|s| take_asref_path(s));
-    map_str_to_path(std::convert::AsRef::as_ref);
+    map_str_to_path(|s| s.as_ref());
 }
 
 mod type_param_bound {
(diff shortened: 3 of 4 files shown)


tests/ui/eta.stderr
```diff
@@ -24,12 +24,6 @@ error: redundant closure
 LL |     all(&[1, 2, 3], &&2, |x, y| below(x, y)); //is adjusted
    |                          ^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `below`
 
-error: redundant closure
-  --> $DIR/eta.rs:40:27
-   |
-LL |     let e = Some(1u8).map(|a| divergent(a));
-   |                           ^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `divergent`
-
```
camsteffen, line 32: This seems like a false negative because `map` allows a function with *any* return type but the bug in the example expects a function that returns nothing (or unit). It seems like it should be possible to detect the difference. Maybe there is an expr type adjustment to coerce `!` to `()` on the closure or the function call within the closure. Or maybe you can detect this scenario by comparing `call_ty` and `closure_ty` (one of them has `-> ()` and the other has `-> !`?).

dswij: Ah, I definitely misunderstood the bug in the example. I will take a further look

dswij: Seems like checking for coerced closure is not enough. In the case above, it seems like the closure is coerced to `()` before being passed on to `map` (not sure why this is the case, though). AFAIU, in the example above, `map` accepts `divergent` nonetheless, probably because of generics. For non-generics, this is not the case and type check wouldn't resolve (hence the problem with issue #8416 and similar, see [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ee3ed9301b745e22ceba0406588b0bc6)).

Ignoring the coerced closure is not hard, but it does produce false negatives such as the case above. Although I'd say that this is still an improvement over all the false positives.

A full solution would be to check if type-check is resolvable for the call/method-call taking  `Fn(..) -> T` without the coerced closure, but I'm not sure that this is trivial to do.

dswij:
Hey @flip1995, I'd love to fix the merge conflicts, but this PR has not seen any activities for some time now. 
Is there anything I can do to help this move forward?

xFrednet:
We currently sadly have a long reviewing queue and only a few reviewers. That's why stale PRs are sadly not always picked up. That being said, I'll take over the review, my queue is still manageable (If I ignore some university and works stuff xD). 

The linked issue contains two references to duplicate issues. Could you add tests and check that they are also fixed? If so, you can add them to the `Closed: ` list in your issue :upside_down_face: 

r? @xFrednet

clippy_lints/src/eta_reduction.rs
```diff
@@ -145,6 +146,7 @@ impl<'tcx> LateLintPass<'tcx> for EtaReduction {
         );
 
         if_chain!(
+            if !is_adjusted(cx, &body.value);
```
xFrednet, line 149: I think that `is_adjusted` is a bit too restrictive. The example cases all are related to the never type. Therefore, I would only suppress the lint if the wrapped function returns the never type, in all other cases it seems like the normal adjustment has worked its magic.

Have you tried the following two options?
1. Check the return type of the body expression of the closure. The expression might still return the never type if it gets adjusted, but I'm not sure about that.
* Otherwise we might have to resolve the referenced function and check what that returns.

dswij: > The example cases all are related to the never type.

The problem is not `never` type, but it's when the wrapped function can only be resolved by typechecks by adjusting it to return `()` type.  

> Check the return type of the body expression of the closure. The expression might still return the never type if it gets adjusted, but I'm not sure about that.

It gets adjusted to `()` type.

> Otherwise we might have to resolve the referenced function and check what that returns.

Yes, the only way that I see for a complete solution is to check if the function typecheck can be resolved to accept the original return type of the closure. I'm not sure this is trivial to do.

dswij:
> We currently sadly have a long reviewing queue and only a few reviewers

Hope the team can get enough bandwidth soon 🙂

xFrednet:
> > We currently sadly have a long reviewing queue and only a few reviewers
> 
> Hope the team can get enough bandwidth soon slightly_smiling_face

Btw, I've pinged you on Zulip. Not sure, if that's your preferred method of communication.

I'll put the next review on my todo list :upside_down_face:

xFrednet:
I've looked through our open issues, and believe that this also fixes:
#7812 and #8091

clippy_lints/src/eta_reduction.rs
xFrednet: > > The example cases all are related to the never type.
> 
> The problem is not `never` type, but it's when the wrapped function can only be resolved by typechecks by adjusting it to return `()` type.

Looks like I also misunderstood the issue at first. The type adjustment to `()` sounds like a quirk of the type checking system.

In that case, I think that your fix is the correct one. It will have some FNs but those are better than FPs IMO. :upside_down_face:

xFrednet (approved):

xFrednet:
FYI, for anyone reading this in the future. This fix introduceses some false negatives, but we believe that the traitof is worth it. Avoiding these false negatives would take several complicated checks which also speaks for this solution.

@dswij Thank you for this change and your pations. It's highly appreciated!

Could you rebase one last time? Than you can use bors with `r=xFrednet`

@bors delegate

giraffate:
@dswij Can you resolve the conflicts?

PR #8433, opened 2022-02-14
Don't lint Default::default if it is the udpate syntax base

hellow554:
changelog: Don't lint `Default::default` it is part of the update syntax

Current clippy warns about this:

```
warning: calling `Foo::default()` is more clear than this expression
  --> src/main.rs:12:11
   |
12 |         ..Default::default()
   |           ^^^^^^^^^^^^^^^^^^ help: try: `Foo::default()`
   |
```

With these changes, it will not lint that particular expression anymore.

### DIFF
--- a/clippy_lints/src/default.rs
+++ b/clippy_lints/src/default.rs
@@ -1,7 +1,7 @@
 use clippy_utils::diagnostics::{span_lint_and_note, span_lint_and_sugg};
 use clippy_utils::source::snippet_with_macro_callsite;
 use clippy_utils::ty::{has_drop, is_copy};
-use clippy_utils::{any_parent_is_automatically_derived, contains_name, match_def_path, paths};
+use clippy_utils::{any_parent_is_automatically_derived, contains_name, get_parent_expr, match_def_path, paths};
 use if_chain::if_chain;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::Applicability;
@@ -88,6 +88,7 @@ impl<'tcx> LateLintPass<'tcx> for Default {
             if let ExprKind::Path(ref qpath) = path.kind;
             if let Some(def_id) = cx.qpath_res(qpath, path.hir_id).opt_def_id();
             if match_def_path(cx, def_id, &paths::DEFAULT_TRAIT_METHOD);
+            if !is_update_syntax_base(cx, expr);
             // Detect and ignore <Foo as Default>::default() because these calls do explicitly name the type.
             if let QPath::Resolved(None, _path) = qpath;
             let expr_ty = cx.typeck_results().expr_ty(expr);
@@ -287,6 +288,19 @@ fn field_reassigned_by_stmt<'tcx>(this: &Stmt<'tcx>, binding_name: Symbol) -> Op
             Some((field_ident, assign_rhs))
         } else {
             None
+        }
+    }
+}
+
+/// Returns whether `expr` is the update syntax base: `Foo { a: 1, .. base }`
+fn is_update_syntax_base<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) -> bool {
+    if_chain! {
+        if let Some(parent) = get_parent_expr(cx, expr);
+        if let ExprKind::Struct(_, _, Some(base)) = parent.kind;
+        then {
+            base.hir_id == expr.hir_id
+        } else {
+            false
         }
     }
 }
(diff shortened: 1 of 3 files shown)


clippy_lints/src/default.rs
```diff
@@ -290,3 +291,16 @@ fn field_reassigned_by_stmt<'tcx>(this: &Stmt<'tcx>, binding_name: Symbol) -> Op
...
     }
 }
+
+/// Returns whether `expr` is the update syntax base: `Foo { a: 1, .. base }`
+fn is_update_syntax_base<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) -> bool {
+    if_chain! {
+        if let Some(parent) = get_parent_expr(cx, expr);
+        if let ExprKind::Struct(_, _, Some(base)) = parent.kind;
+        then {
+            base.hir_id == expr.hir_id
```
hellow554, line 301: This looks weird to me. Is this the correct way if two expressions are the same?

tests/ui/default_trait_access.fixed
```diff
@@ -46,9 +46,14 @@ fn main() {
...
+    let s20 = UpdateSyntax {
+        s: "foo",
+        ..Default::default()
+    };
+
     println!(
-        "[{}] [{}] [{}] [{}] [{}] [{}] [{}] [{}] [{}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}], [{:?}]",
-        s1, s2, s3, s4, s5, s6, s7, s8, s9, s10, s11, s12, s13, s14, s15, s16, s17, s18, s19,
+        "[{}] [{}] [{}] [{}] [{}] [{}] [{}] [{}] [{}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}] [{:?}]",
+        s1, s2, s3, s4, s5, s6, s7, s8, s9, s10, s11, s12, s13, s14, s15, s16, s17, s18, s19, s20,
```
hellow554, line 56: There was a spurious `,`. I removed it :)

clippy_lints/src/default.rs
```diff
@@ -81,6 +81,7 @@ impl<'tcx> LateLintPass<'tcx> for Default {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
         if_chain! {
             if !expr.span.from_expansion();
+            if !is_update_syntax_base(cx, expr);
```
hellow554, line 84: We can discuss about the position of the check inside of the `if_chain`

hellow554:
friendly ping @flip1995

clippy_lints/src/default.rs
flip1995: Depends on what you want to do. Since you want to check that those two expressions are _exactly_ the same (in the same position and the same "content"), yes comparing `HirId` is the way to go.

FYI: If you want to compare 2 expressions if they "look" the same, we have `SpanlessEq` in `clippy_utils` that does that for you.

clippy_lints/src/default.rs
flip1995: I would move it down to the point where we at least know that we're dealing with a `Default::default()` call. I don't think performance matters here, though I think the code is easier to read (because you know what `expr` is) if it is placed later in the `if_chain`

flip1995:
Sorry, I'm currently really busy with my _$day_job_ :/ 

What's the reasoning behind not linting this case? Is there an issue about it?

Impl LGTM, except for the position in the `if_chain`, see comment above.

hellow554:
> Sorry, I'm currently really busy with my $day_job :/

Oh. This doesn't have to hurry, it can rest one week or more ;)

> What's the reasoning behind not linting this case? Is there an issue about it?

Sort of.  
[Fields assign with default](https://rust-lang.github.io/rust-clippy/master/index.html#field_reassign_with_default) recommends using the `.. Default::default()` syntax explicitly and TBH you already write `Foo { .... }`, which means it is obvious which type you are currently try to *default*. It should be the same reasoning as with `<Foo as Default>::default`:

https://github.com/rust-lang/rust-clippy/blob/b34cd79c0cf2f5425aa56a77cbe95459bef1d98f/clippy_lints/src/default.rs#L91

> Impl LGTM, except for the position in the if_chain, see comment above.

I moved it down right after the check that that expression is a default call. Thanks for that.

flip1995:
That's a convincing argument :+1: 

Thanks!

@bors r+

PR #8434, opened 2022-02-14
Add lint for holding a `tracing` Span Guard Across an `.await`

davidbarsky:
Hello! This PR adds a new pedantic lint, ``[`await_holding_span_guard`]`` that prohibits the usage of a tracing span guard across an await boundary. If a `tracing` Span guard is used across an await boundary, tracing will record incorrect traces that _don't appear until concurrent execution of futures occurs_. In practice, this means that this issue won't occur while testing locally. While this is mentioned [in tracing's documentation](https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code), I believe that this is a footgun that is best addressed through a lint or [some language changes](https://internals.rust-lang.org/t/pre-rfc-lifecycle-trait/12311). Until such a feature can be landed, I think that a Clippy lint is the most expedient approach.

As for whether this lint should be included within Clippy, _I_ think it should, but I can see the arguments why it shouldn't. In any case, here are the reasons in favor of inclusion within Clippy:
- Some prominent, third-party crates (regex, parking_lot, and tokio) have lints within Clippy. While `tracing` is probably the smallest of those three crates, it does have some non-trivial usage across the ecosystem (e.g., Tokio, rustc, chalk). Unfortunately, I don't have a principled policy in mind for what crates can and cannot be linted by Clippy and I understand if linting tracing is a step too far.
- The causes as to why tracing's guards can't be held across an await point are similar to that of mutex guards. However, instead of that causing a deadlock, incorrect traces will be reported.

Some additional notes for review:
- I'm not sure what clippy's policy for deprecating lints are, but once [`#[must_not_suspend]`](https://github.com/rust-lang/rust/issues/83310) lands in rustc, I expect that tracing will move to it as soon as its MSRV allows it to. Once that occurs, I would be in favor of removing this lint from clippy.
- In keeping with the naming of `#[must_not_suspend]`, should this lint be renamed to ``#[`suspend_holding_span_guard`]``? Should this lint have `tracing` in the name?

---

changelog: Add a lint for holding a `tracing` Span guard across an `.await` boundary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3051,6 +3051,7 @@ Released 2018-09-13
 [`async_yields_async`]: https://rust-lang.github.io/rust-clippy/master/index.html#async_yields_async
 [`await_holding_lock`]: https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_lock
 [`await_holding_refcell_ref`]: https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_refcell_ref
+[`await_holding_span_guard`]: https://rust-lang.github.io/rust-clippy/master/index.html#await_holding_span_guard
 [`bad_bit_mask`]: https://rust-lang.github.io/rust-clippy/master/index.html#bad_bit_mask
 [`bind_instead_of_map`]: https://rust-lang.github.io/rust-clippy/master/index.html#bind_instead_of_map
 [`blacklisted_name`]: https://rust-lang.github.io/rust-clippy/master/index.html#blacklisted_name
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ syn = { version = "1.0", features = ["full"] }
 futures = "0.3"
 parking_lot = "0.11.2"
 tokio = { version = "1", features = ["io-util"] }
+tracing = "0.1"
 
 [build-dependencies]
 rustc_tools_util = { version = "0.2", path = "rustc_tools_util" }
--- a/clippy_lints/src/await_holding_invalid.rs
+++ b/clippy_lints/src/await_holding_invalid.rs
@@ -95,7 +95,74 @@ declare_clippy_lint! {
     "Inside an async function, holding a RefCell ref while calling await"
 }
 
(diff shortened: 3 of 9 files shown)


flip1995 (commented):
This adds another dependency to an external crate. Clippy is not a tool to implement crate-specific lints. We have exceptions for `regex`, `serde` and `parking_lot` . 

- regex is a repo of the rust-lang org
- serde is IMHO a relic of the past, for which I don't see the justification for the 2 lints we have for it.
- parking_lot was added, because many people use those mutexes over the `std` mutexes in production code and there is the possibility that `parking_lot` will replace the `std` implementation at some point.

So I'm not sure if we should continue sliding down this slope of adding more and more lints for external/non-std crates.

davidbarsky:
Thanks for leaving your thoughts! As for:

> So I'm not sure if we should continue sliding down this slope of adding more and more lints for external/non-std crates.

No worries, I think that this is should be ironed out. I know the feeling of being responsible for crates/code that you'd rather not be and finding a clear policy is tricky, and if this PR is rejected _but a clear policy comes of it_, then I'd consider that this PR to be a net benefit.

> Adding a lint to Clippy with a deadline in mind when it can be removed is questionable. Instead of adding the lint, wouldn't it make more sense to put that work into the feature so it get's stabilized quicker?

Possibly, but the work needed to land `#[must_not_suspend]` seems [tricky and cross-cutting](https://github.com/rust-lang/rust/issues/89562#issuecomment-940392321) (especially the resolving https://github.com/rust-lang/rust/issues/69663), but I can ask and find out.

camsteffen:
Given that this may (probably) be fixed by `#[must_not_suspend]` and that this lint detects a low-severity issue, I don't think it is quite worth the added dependency personally.

davidbarsky:
> Given that this may (probably) be fixed by #[must_not_suspend] and that this lint detects a low-severity issue, I don't think it is quite worth the added dependency personally.

No worries! Then it might be better to see what can be done to improve `#[must_not_suspend]`. I'll close this PR in that case.

PR #8355, opened 2022-01-27
Add lint `explicit_auto_deref` take 2

Jarcho:
fixes: #234
fixes: #8367
fixes: #8380

Still things to do:

* ~~This currently only lints `&*<expr>` when it doesn't trigger `needless_borrow`.~~
* ~~This requires a borrow after a deref to trigger. So `*<expr>` changing `&&T` to `&T` won't be caught.~~
* The `deref` and `deref_mut` trait methods aren't linted.
* Neither ~~field accesses~~, nor method receivers are linted.
* ~~This probably shouldn't lint reborrowing.~~
* Full slicing to deref should probably be handled here as well. e.g. `&vec[..]` when just `&vec` would do

changelog: new lint `explicit_auto_deref`

PR #8441, opened 2022-02-17
Don't lint `needless_borrow` in method receiver positions

### LINKED ISSUE #8367: `needless_borrow` suggestion doesn't compile
### Summary

On the most recent nightly, the `needless_borrow` lint produces a suggestion that doesn't compile.

`cargo +nightly-2022-01-29 clippy` triggers this bug, but `cargo +nightly-2022-01-28 clippy` does not.

cc: @Jarcho (maybe?)

### Reproducer

**I tried this code:**

```rust
fn main() {
    let s0 = String::new();
    let mut chars = s0.chars();
    let s1 = (&mut chars).take(0).collect::<String>();
    println!("{}", s1);
    drop(chars);
}
```

**I expected to see this happen:**

Nothing.

**Instead, this happened:**

```
warning: this expression borrows a value the compiler would automatically borrow
 --> src/main.rs:4:14
  |
4 |     let s1 = (&mut chars).take(0).collect::<String>();
  |              ^^^^^^^^^^^^ help: change this to: `chars`
  |
```
If you make the suggested change, you get:
```
error[E0382]: use of moved value: `chars`
    --> src/main.rs:6:10
     |
3    |     let mut chars = s0.chars();
     |         --------- move occurs because `chars` has type `Chars<'_>`, which does not implement the `Copy` trait
4    |     let s1 = chars.take(0).collect::<String>();
     |                    ------- `chars` moved due to this method call
5    |     println!("{}", s1);
6    |     drop(chars);
     |          ^^^^^ value used here after move
     |
note: this function takes ownership of the receiver `self`, which moves `chars`
    --> /home/smoelius/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:1317:13
     |
1317 |     fn take(self, n: usize) -> Take<Self>
     |             ^^^^
```

### Version

```text
rustc 1.60.0-nightly (e0a55f449 2022-01-28)
binary: rustc
commit-hash: e0a55f4491a729bffc63b402ba903d90858c806b
commit-date: 2022-01-28
release: 1.60.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #8380: needless_borrow false positive, with broken suggestion
### To reproduce

Run `cargo clippy` with this in `src/main.rs`:

```rust
struct A { }

#[derive(Debug)]
struct B { }

impl From<&A> for B {
    fn from(_a: &A) -> B { B { } }
}

fn main() {
    let a = A { };
    
    // warning: this expression borrows a value the compiler would automatically borrow
    // 12 | let b: B = (&a).into();
    //                 ^^^^ help: change this to: `a`
    let b: B = (&a).into();
    
    // error[E0277]: the trait bound `B: std::convert::From<A>` is not satisfied
    // let b: B = a.into();
    
    println!("{:?}", &b)
}
```

### Output with current nightly (bad)

```
    Checking foo v0.1.0 (/home/rustcargo/d/foo)
warning: this expression borrows a value the compiler would automatically borrow
  --> src/main.rs:16:16
   |
16 |     let b: B = (&a).into();
   |                ^^^^ help: change this to: `a`
   |

warning: `foo` (bin "foo") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
```

### Discussion

The disagnostic is not correct.  Changing the program as clippy suggests does not work (see comment in my example).  Both nightly and stable reject the modified program.

Stable clippy IMO-rightly accepts the original program without complaint.

### Meta

```
rustcargo@zealot:~/d/foo$ rustc +stable --version
rustcargo@zealot:~/d/foo$ rustc +nightly --version
rustc 1.60.0-nightly (6abb6385b 2022-01-26)
rustcargo@zealot:~/d/foo$ 
```

### LINKED ISSUE #8391: `needless_borrow` suggestion changes a reference to a move which breaks surrounding code
### Summary

Consider `std::io::Read`, which has methods that take `self`, but the trait is also explicitly implemented for `&mut R` where `R: Read`.

Consider the following code:
```rust
use std::io::{Cursor, Read};
let mut r = Cursor::new("hello world");
let a = (&mut r).take(3);
// do stuff, drop a
let b = r.take(3);
```
`a` lets you still use `r` afterward, but `b` does not, because it moves `r` into the wrapper.

### Lint Name

needless_borrow

### Reproducer

I ran into this problem with code that uses a reader in a loop, using `take` to read it in chunks, similar to this:

```rust
use std::io::{self, Cursor, Read, Write};

fn main() {
    let mut data = Cursor::new("line 1\n\
        line 2\n\
        line 3\n\
        line 4\n");
    let stdout = io::stdout();
    let mut stdout = stdout.lock();
    loop {
        let mut chunk = (&mut data).take(2); // `(&mut data)` can't be replaced by `data`!
        let n = io::copy(&mut chunk, &mut stdout).unwrap();
        if n == 0 {
            break;
        }
        stdout.write_all(b".").unwrap();
    }
}
```

Clippy says:

```
warning: this expression borrows a value the compiler would automatically borrow
  [--> src/main.rs:11:25
11 |         let mut chunk = (&mut data).take(2); // `(&mut data)` can't be replaced by `data`!
   |                         ^^^^^^^^^^^ help: change this to: `data`
   |
```

But that won't compile:
```
error[[E0382]](https://doc.rust-lang.org/stable/error-index.html#E0382): use of moved value: `data`
   [--> src/main.rs:12:25
](https://play.rust-lang.org/#)    |
4   |     let mut data = Cursor::new("line 1\n\
    |         -------- move occurs because `data` has type `std::io::Cursor<&str>`, which does not implement the `Copy` trait
...
12  |         let mut chunk = data.take(2);
    |                         ^^^^ ------- `data` moved due to this method call, in previous iteration of loop
    |
note: this function takes ownership of the receiver `self`, which moves `data`
```

### Version

```text
rustc 1.60.0-nightly (27f5d830e 2022-02-02)          
binary: rustc                                        
commit-hash: 27f5d830eb11cd7bdc834d6f0d78120976f75443
commit-date: 2022-02-02                              
release: 1.60.0-nightly                              
LLVM version: 13.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #8407: Incorrect: `this expression borrows a value the compiler would automatically borrow`
### Summary

Clippy claims that a compiler would automatically borrow specific user borrows. E.g. `(&impl Struct)`. These explicit borrows are done to dispatch to a implementation on the reference. So that owned code can utilize the same implementation.

### Lint Name

needless_borrow

### Reproducer

I tried this code:

```rust
use std::ops::Add;

struct Bar {
    val: i32
}

impl Add for &Bar {
    type Output = Bar;

    fn add(self, rhs: Self) -> Self::Output {
        let new = self.val + rhs.val;
        Bar {
            val: new
        }
    }
}

impl Add for Bar {
    type Output = Bar;

    fn add(self, rhs: Self) -> Self::Output {
        // dispatch to ref impl
        (&self).add(&rhs)
    }
}
```

I saw this happen:

```
warning: this expression borrows a value the compiler would automatically borrow
  --> src/main.rs:58:9
   |
58 |         (&self).add(&rhs)
   |         ^^^^^^^ help: change this to: `self`
   |


```

No warning.

Doing as the lint suggests, leads to recursive call site

```
warning: function cannot return without recursing
  --> src/main.rs:56:5
   |
56 |     fn add(self, rhs: Self) -> Self::Output {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot return without recursing
57 |         // dispatch to ref impl
58 |         self.add(rhs)
   |         ------------- recursive call site
   |
   = note: `#[warn(unconditional_recursion)]` on by default
   = help: a `loop` may express intention better if this is on purpose

```

So that's a catch 22.

### Version

```text
rustc 1.60.0-nightly (0c292c966 2022-02-08)
binary: rustc
commit-hash: 0c292c9667f1b202a9150d58bdd2e89e3e803996
commit-date: 2022-02-08
release: 1.60.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #8408: Incorrect `needless_borrow`
### Summary

Clippy suggests removing a borrow that would lead to ownership being moved.

### Lint Name

needless_borrow

### Reproducer

I tried this code:

```rust
fn main() {
    let mut iter = vec![1, 2, 3].into_iter();

    let mut sum = 0;
    for v in (&mut iter).enumerate().take(1) {
        sum += v.1;
    }
    
    let b = iter.count();
}
```

I saw this happen:

```
warning: this expression borrows a value the compiler would automatically borrow
  --> src/main.rs:65:14
   |
65 |     for v in (&mut iter).enumerate().take(1) {
   |              ^^^^^^^^^^^ help: change this to: `iter`

```

If we follow the lint:

```rust
fn main() {
    let mut iter = vec![1, 2, 3].into_iter();

    let mut sum = 0;
    for v in iter.enumerate().take(1) {
        sum += v.1;
    }

    let b = iter.count();
}
```

Our code does not compile:

```
error[E0382]: use of moved value: `iter`
   --> src/main.rs:69:13
    |
62  |     let mut iter = vec![1, 2, 3].into_iter();
    |         -------- move occurs because `iter` has type `std::vec::IntoIter<i32>`, which does not implement the `Copy` trait
...
65  |     for v in iter.enumerate().take(1) {
    |                   ----------- `iter` moved due to this method call
...
69  |     let b = iter.count();
    |             ^^^^ value used here after move
    |
note: this function takes ownership of the receiver `self`, which moves `iter`
   --> /home/ritchie46/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:928:18
    |
928 |     fn enumerate(self) -> Enumerate<Self>
    |                  ^^^
```

### Version

```text
rustc 1.60.0-nightly (e7aca8959 2022-02-09)
binary: rustc
commit-hash: e7aca895980f25f6d2d3c48e10fd04656764d1e4
commit-date: 2022-02-09
release: 1.60.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Jarcho:
fixes #8408 
fixes #8407 
fixes #8391
fixes #8367
fixes #8380

This is a temporary fix for `needless_borrow`. The proper fix is included in #8355.

This should probably be merged into rustc before beta branches on Friday. This issue has been reported six or seven times in the past couple of weeks.

changelog: Fix various issues with `needless_borrow` n´. Note to changelog writer: those issues might have been introduced in this release cycle, so this might not matter in the changelog.

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -528,7 +528,7 @@ fn is_auto_reborrow_position(parent: Option<Node<'_>>) -> bool {
 fn is_auto_borrow_position(parent: Option<Node<'_>>, child_id: HirId) -> bool {
     if let Some(Node::Expr(parent)) = parent {
         match parent.kind {
-            ExprKind::MethodCall(_, [self_arg, ..], _) => self_arg.hir_id == child_id,
+            // ExprKind::MethodCall(_, [self_arg, ..], _) => self_arg.hir_id == child_id,
             ExprKind::Field(..) => true,
             ExprKind::Call(f, _) => f.hir_id == child_id,
             _ => false,
--- a/tests/ui/needless_borrow.fixed
+++ b/tests/ui/needless_borrow.fixed
@@ -64,9 +64,9 @@ fn main() {
     *x = 5;
 
     let s = String::new();
-    let _ = s.len();
-    let _ = s.capacity();
-    let _ = s.capacity();
+    // let _ = (&s).len();
+    // let _ = (&s).capacity();
+    // let _ = (&&s).capacity();
 
     let x = (1, 2);
     let _ = x.0;
--- a/tests/ui/needless_borrow.rs
+++ b/tests/ui/needless_borrow.rs
@@ -64,9 +64,9 @@ fn main() {
     *x = 5;
 
     let s = String::new();
-    let _ = (&s).len();
-    let _ = (&s).capacity();
-    let _ = (&&s).capacity();
+    // let _ = (&s).len();
+    // let _ = (&s).capacity();
+    // let _ = (&&s).capacity();
 
     let x = (1, 2);
     let _ = (&x).0;
--- a/tests/ui/needless_borrow.stderr
+++ b/tests/ui/needless_borrow.stderr
@@ -85,24 +85,6 @@ LL |     let y: &mut i32 = &mut &mut x;
    |                       ^^^^^^^^^^^ help: change this to: `x`
 
 error: this expression borrows a value the compiler would automatically borrow
-  --> $DIR/needless_borrow.rs:67:13
-   |
-LL |     let _ = (&s).len();
-   |             ^^^^ help: change this to: `s`
-
-error: this expression borrows a value the compiler would automatically borrow
-  --> $DIR/needless_borrow.rs:68:13
-   |
-LL |     let _ = (&s).capacity();
-   |             ^^^^ help: change this to: `s`
-
-error: this expression creates a reference which is immediately dereferenced by the compiler
-  --> $DIR/needless_borrow.rs:69:13
-   |
-LL |     let _ = (&&s).capacity();
-   |             ^^^^^ help: change this to: `s`
-
-error: this expression borrows a value the compiler would automatically borrow
   --> $DIR/needless_borrow.rs:72:13
    |
 LL |     let _ = (&x).0;
@@ -114,5 +96,5 @@ error: this expression borrows a value the compiler would automatically borrow
 LL |     let _ = unsafe { (&*x).0 };
    |                      ^^^^^ help: change this to: `(*x)`
 
-error: aborting due to 19 previous errors
+error: aborting due to 16 previous errors
 


camsteffen:
LGTM. Let's go ahead and close whatever issues are fixed here.

fixes: #8367
fixes: #8380

r? @flip1995 - passing this to you since you might get this into rustc.

flip1995:
> LGTM. Let's go ahead and close whatever issues are fixed here.

I think you have to put those directly in the PR body. I did so and also added a changelog.



> passing this to you since you might get this into rustc.

I submitted this commit directly to rust-lang/rust, so we don't have to do a full sync: https://github.com/rust-lang/rust/pull/94085

But let's also merge it here to not diverge too much from the rust-lang/rust repo

@bors r+

Issue #8291, opened 2022-01-15
cargo clippy --explain <lintname>


matthiaskrgr:
### Description

Rustc has this nice `rustc --explain E060` feature.
I wonder if we could have the same in form of `cargo clippy --explain needless_borrow` for example.
This would print the lints explanation of the doc-comment ( https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow ) into the terminal, the advantage being that we don't need a web browser and internet to access the lint explanation :slightly_smiling_face:

xFrednet:
I've added the nominated tag. Having this sounds like a good idea :upside_down_face:

xFrednet:
FYI: This was discussed in today's meeting. See [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202022-01-25/near/269275900) Everyone agreed that this is a good idea and @llogiq wants to work on this.

@rustbot label -I-nominated

llogiq:
I wonder how to best store the collected data. To start, I'm going to build a macro that will move the doc comment to a string argument.

xFrednet:
Sounds like a good idea, the Metadata collector for our lint list collects the docs from the doc comment. If you change the macro it would be good to reapply the doc comments to the lint item.

Storage will be an interesting problem, I've looked into some ideas while I was working on #8211 but haven't found a nice solution yet. I even thought about building a wrapper type about rustc's lint to store some additional Clippy data, but that didn't feel quite right to only implement nightly lints. I'm interested to see what you come up with :upside_down_face:

PR #8444, opened 2022-02-17
Showcase nightly lints with macro

xFrednet:
This is a showcase of how we can use Clippy's `declare_clippy_lint!` macro to implement nightly lints. The work's macro work is loosely based on https://github.com/rust-lang/rust-clippy/pull/6830 by @camsteffen.

Some background, this uses the `CFG_DISABLE_UNSTABLE_FEATURES` environment value to check at compile time if this is a nightly build of Clippy. The same value is used by [`UnstableFeatures::from_environment()`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_feature/enum.UnstableFeatures.html#method.from_environment). It was pointed out on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/.E2.9C.94.20Target.20release.20channel.20for.20Clippy/near/266637254) that this works but the actual intended and clean way would be to use [`UnstableFeatures::from_environment()`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_feature/enum.UnstableFeatures.html#method.from_environment) which is sadly not const.

In this version, I was unable to support a second environment value like `CLIPPY_NIGHTLY` to give the user control as all logic happens at compile time. The result is quite clean, but I would like to give more control to the user.

I tried to dynamically initialize lints with [`std::lazy::SyncLazy`](https://doc.rust-lang.org/std/lazy/struct.SyncLazy.html) in [another branch](https://github.com/xFrednet/rust-clippy/tree/8211-nightly-lints-dynamic-no) but that required dereferencing the instance for each usage to access the inner lint.

---

I can't really say which option I like the most. If we were starting from scratch, I would suggest using [`std::lazy::SyncLazy`](https://doc.rust-lang.org/std/lazy/struct.SyncLazy.html) and just dereference every lint usage. However, in Clippy that would requite countless changes, even if simple once, and it might have a performance impact. Initializing it at compile time, like this PR does is quite clean in code, but doesn't give any power to the user to opt in or out of Clippy's nightly lints which I would like. The solution from https://github.com/rust-lang/rust-clippy/pull/8435 works quite well, but is also not ideal.

Another option could be to add a new `Nightly` level to rustc. However, I would guess that that level would require a feature flag to enable emission, which we wanted to avoid. I would like to hear what you think about this and how nightly lints should be managed in general. :)

---

changelog: New lints can now be restricted to nightly runs to enable better testing

cc: https://github.com/rust-lang/rust-clippy/issues/8211

cc: @camsteffen @flip1995 

r? @flip1995 (for now, feel free to reassign)

BTW: I will rebase this and the other PR once we have decided which version we want to keep or if we want an entirely different implementation :upside_down_face:


camsteffen:
> I tried to dynamically initialize lints with [`std::lazy::SyncLazy`](https://doc.rust-lang.org/std/lazy/struct.SyncLazy.html) in [another branch](https://github.com/xFrednet/rust-clippy/tree/8211-nightly-lints-dynamic-no) but that required dereferencing the instance for each usage to access the inner lint.

To avoid all the dereferencing, another possibility is to have a new type for `declare_clippy_lint!`

```rust
struct ClippyLint {
    lint: SyncLazy<Lint>,
    category: .. // maybe?
}
```

and then the lint utils would accept `&ClippyLint`. This could have a side benefit of simplifying the metadata collection code.

xFrednet:
True, that's also something I suggested for https://github.com/rust-lang/rust-clippy/issues/8291. All rustc functions would still take `&'static Lint` but those instance would most likely be manageable. That's something I would work. However, I first want to be sure that we want to go that route. :upside_down_face:

flip1995:
Will this version produce `unkown_lints` warnings, if you try to `allow(nightly_lint)` on a stable version? If so, that would be quite bad, because there would be no way to disable those.

I think the first version looks better to me. 

I'm not opposed to a wrapper struct for Clippy lints, where we may be able to store the documentation, category or other Clippy-specific lint informations.

---

Another question would be, if we want to add first-class support of nightly lints to rustc, instead of trying to hammer this into Clippy. I could imagine that rustc might be better suited to handle lint emission suppression or deal with `allow`/`warn`/... attributes with nightly lints. This will probably require a compiler MCP, but we could first just casually ask on Zulip.

xFrednet:
I'll create a topic in `t-compiler/wg-diagnostics` to ask. Otherwise, I think that having a wrapper type for Lints would be the cleanest solution :upside_down_face:

xFrednet:
I'm going to close this PR as it's likely that we can implement nightly lints in rustc directly. Otherwise, this will also be simple to reopen :upside_down_face:

PR #8445, opened 2022-02-18
Llint for casting between raw slice pointers with different element sizes

asquared31415:
This lint disallows using `as` to convert from a raw pointer to a slice (e.g. `*const [i32]`, `*mut [Foo]`) to any other raw pointer to a slice if the element types have different sizes.  When a raw slice pointer is cast, the data pointer and count metadata are preserved.  This means that when the size of the inner slice's element type changes, the total number of bytes pointed to by the count changes.  For example a `*const [i32]` with length 4 (four `i32` elements) is cast `as *const [u8]` the resulting pointer points to four `u8` elements at the same address, losing most of the data.  When the size *increases* the resulting pointer will point to *more* data, and accessing that data will be UB.

On its own, *producing* the pointer isn't actually a problem, but because any use of the pointer as a slice will either produce surprising behavior or cause UB I believe this is a correctness lint.  If the pointer is not intended to be used as a slice, the user should instead use any of a number of methods to produce just a data pointer including an `as` cast to a thin pointer (e.g. `p as *const i32`) or if the pointer is being created from a slice, the `as_ptr` method on slices.  Detecting the intended use of the pointer is outside the scope of this lint, but I believe this lint will also lead users to realize that a slice pointer is only for slices.

There is an exception to this lint when either of the slice element types are zero sized (e.g `*mut [()]`).  The total number of bytes pointed to by the slice with a zero sized element is zero.  In that case preserving the length metadata is likely intended as a workaround to get the length metadata of a slice pointer though a zero sized slice.

The lint does not forbid casting pointers to slices with the *same* element size as the cast was likely intended to reinterpret the data in the slice as some equivalently sized data and the resulting pointer will behave as intended.

---

changelog: Added ``[`cast_slice_different_sizes`]``, a lint that disallows using `as`-casts to convert between raw pointers to slices when the elements have different sizes.


clippy_lints/src/casts/cast_slice_different_sizes.rs
```diff
@@ -0,0 +1,119 @@
...
+                        );
+
+                        diag.span_suggestion(
+                            expr.span,
+                            &format!("replace with `ptr::slice_from_raw_parts{mutbl_fn_str}`"),
+                            sugg,
+                            rustc_errors::Applicability::HasPlaceholders,
+                        );
+
+                        diag.help("the count in elements of raw pointers to slices is preserved across `as`-casts so the slice covers a different number of bytes");
```
llogiq, line 83: I think the help is a bit superfluous and also quite wordy. Either tailor it to the case at hand ("this will only contain as many `u8` as the casted slice had `i32`s") or leave it out entirely.

llogiq (commented):
This looks good, albeit a bit wordy. The help could be either more specific or removed wholesale, I don't think it's pulling it's weight.

clippy_lints/src/casts/mod.rs
```diff
@@ -409,6 +410,50 @@ declare_clippy_lint! {
...
+    /// // The data pointer is cast to a pointer to the target `u8` not `[u8]`
+    /// // The length comes from the known length of 4 i32s times the 4 bytes per i32
+    /// let new_ptr = core::ptr::slice_from_raw_parts(old_ptr as *const u8, 16);
+    /// unsafe {
+    ///     println!("{:?}", &*new_ptr);
+    /// }
+    /// ```
+    #[clippy::version = "1.60.0"]
+    pub CAST_SLICE_DIFFERENT_SIZES,
+    correctness,
```
llogiq, line 453: I wonder if there could be cases where the code is actually correct, even if surprising.

If that is the case, I worry that having a deny by default lint might lead to a downgrade to nursery.

llogiq (commented):
This looks good in general. Just a small style nit on the docs.

clippy_lints/src/casts/mod.rs
asquared31415: The exceptions when one or both target slices have a zero sized inner type should cover what I believe are the only correct uses:
 - To get the length metadata of a slice pointer on stable Rust by going through a `&[()]` (because there's zero bytes operated on).
 - To come from a pointer like `*const [()]` that has the correct inner data and size metadata (would be really weird but probably not wrong)
 
I asked on the Rust community discord if anyone had any other ideas about possible correct uses of a cast like this and nobody had any ideas.

Any time that data is actually accessed behind the pointer it either can be done with a thin pointer or with a slice pointer with a correct length.  While I suppose it's technically the dereference that could cause behavior, I don't think there's ever a way to correctly use a slice pointer with the incorrect length, so I believe that it should be deny by default.

clippy_lints/src/casts/mod.rs
```diff
@@ -409,6 +410,50 @@ declare_clippy_lint! {
     "casts from an enum type to an integral type which will truncate the value"
 }
 
+declare_clippy_lint! {
+    /// Checks for `as` casts between raw pointers to slices with elements with different sizes.
```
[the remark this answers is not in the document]
asquared31415: That reads much better, thanks.

llogiq:
Thank you!

@bors r+

asquared31415:
Edited the PR description, I think it didn't like the newline right after the `changelog:`

xFrednet:
Sounds correct, thanks for the change :upside_down_face: 

@bors retry

xFrednet:
>  Downloading artifact 'target' to: '/home/runner/work/rust-clippy/rust-clippy/target'
Error: The HTTP request timed out after 00:01:40.
Error: Exit code 1 returned from process: file name '/home/runner/runners/2.288.1/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Artifact.DownloadArtifact, Runner.Plugins"'.

The CI seams to be acting up. Here is one more try ^^

@bors retry

llogiq:
@bors retry

more CI flakyness...

Issue #8449, opened 2022-02-19
`undocumented_unsafe_blocks` false positive when safety comment is on the previous line


digama0:
### Description

This is somewhat annoying when the project uses minimal-scope `unsafe` blocks. I would prefer for the lint to only trigger if the previous line from the word `unsafe` is not part of a `// Safety` comment. Currently it requires that the comment *directly* precede the `unsafe` token, which leads to a false positive witth very awkward formatting in the suggestion.

```rust
#![warn(clippy::undocumented_unsafe_blocks)]

fn main() {
    let x = &0 as *const _;
    // Safety: x points to a valid value
    println!("{}", unsafe { *x })
}
```
```
warning: unsafe block missing a safety comment
 --> src/main.rs:6:19
  |
6 |     println!("{}", unsafe { *x })
  |                    ^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/main.rs:1:9
  |
1 | #![warn(clippy::undocumented_unsafe_blocks)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks
help: consider adding a safety comment
  |
6 ~     println!("{}", // SAFETY: ...
7 ~     unsafe { *x })
  |
```

### Version

```text
rustc 1.60.0-nightly (b17226fcc 2022-02-18)
binary: rustc
commit-hash: b17226fcc11587fed612631be372a5b4cb89988a
commit-date: 2022-02-18
host: x86_64-unknown-linux-gnu
release: 1.60.0-nightly
LLVM version: 14.0.0
```

digama0:
If my reading of the code is correct, the reason for this behavior is here:

https://github.com/rust-lang/rust-clippy/blob/29ee5e25d1096f142729ea44c1286e50823762e8/clippy_lints/src/undocumented_unsafe_blocks.rs#L129-L137

it only looks for comments in the enclosing scope. One way to fix this would be to keep going up the syntax tree until you find an enclosing scope that starts on a line different from the unsafe block itself. That would allow things like:
```rust
fn main() {
    // Safety: <- too far away to count
    { // <- this is the enclosing scope within which safety comments are permitted
        // Safety: <- safety comment permitted here 
        let x = { unsafe { foo() } }; // <- surrounding block scope is ignored because it starts on same line
    }
}
```

Issue #8264, opened 2022-01-10
Unfixable `clippy::undocumented_unsafe_blocks` in macro output


CraftSpider:
### Summary

When a proc macro emits an unsafe block with a span somewhere inside the crate using it, it triggers the `undocumented_unsafe_blocks` lint, even if there's documentation on it in the using crate, or if using `quote`, on the original unsafe block

### Lint Name

undocumented_unsafe_blocks

### Reproducer

I tried this code:

foo-proc crate
```rust
use proc_macro::TokenStream;
use quote::quote_spanned;
use quote::spanned::Spanned;

#[proc_macro]
pub fn foo(ts: TokenStream) -> TokenStream {
    quote_spanned!(proc_macro2::TokenStream::from(ts).__span() => {
        // SAFETY: Foobar
        unsafe { ::core::mem::transmute::<i32, i32>(1) }
    }).into()
}
```

foo crate
```
#![warn(clippy::undocumented_unsafe_blocks)]

use foo_proc::foo;

fn main() {
    // SAFETY: Foo
    let _ = foo!(a b c);
}
```

I saw this happen:

```
warning: unsafe block missing a safety comment
 --> src/main.rs:7:5
  |
7 |     let _ = foo!(a b c);
  |     ^^^^^^^^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/main.rs:1:9
  |
1 | #![warn(clippy::undocumented_unsafe_blocks)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks
help: consider adding a safety comment
  |
7 ~     // SAFETY: ...
8 +     let _ = foo!(a b c);
  |
```

I expected to see this happen:

No output. Preferably, there should be no SAFETY comment required in the user's code, as I'd expect it to be the macro's responsibility to handle its safety requirements.

### Version

```text
rustc 1.59.0-nightly (8f3238f89 2022-01-02)
binary: rustc
commit-hash: 8f3238f898163f09726c3d2b2cc9bafb09da26f3
commit-date: 2022-01-02
host: x86_64-unknown-linux-gnu
release: 1.59.0-nightly
LLVM version: 13.0.0
```

PR #8450, opened 2022-02-19
Rework `undocumented_unsafe_blocks`

Jarcho:
fixes: #8264
fixes: #8449

One thing came up while working on this. Currently comments on the same line are supported like so:

```rust
/* SAFETY: reason */ unsafe {}
```

Is this worth supporting at all? Anything other than a couple of words doesn't really fit well.

edit: [zulip topic](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60undocumented_unsafe_blocks.60.20same.20line.20comment)

changelog: Don't lint `undocumented_unsafe_blocks` when the unsafe block comes from a proc-macro.
changelog: Don't lint `undocumented_unsafe_blocks` when the preceding line has a safety comment and the unsafe block is a sub-expression.

### DIFF
--- a/tests/ui/undocumented_unsafe_blocks.rs
+++ b/tests/ui/undocumented_unsafe_blocks.rs
@@ -1,4 +1,8 @@
+// aux-build:proc_macro_unsafe.rs
+
 #![warn(clippy::undocumented_unsafe_blocks)]
+
+extern crate proc_macro_unsafe;
 
 // Valid comments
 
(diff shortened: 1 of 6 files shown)


Jarcho:
This should be good to go now.

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -89,11 +93,6 @@ fn block_comment_newlines() {
     unsafe {}
 }
 
-#[rustfmt::skip]
-fn inline_block_comment() {
-    /* Safety: */unsafe {}
```
giraffate, line 94: It would be better to leave this as it is and add comments not to support this style.

giraffate (commented):
Can the comment about style that are no longer supported be added to `What it does` section on this lint?

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

Issue #8321, opened 2022-01-20
large_enum_variant


karim-agha:
### Summary

Incorrectly receiving this lint message when one of the enum variants is a generic struct and the other is not generic.

### Reproducer

I tried this code:

```rust
pub struct Vote {
  pub validator: Pubkey,
  pub target: Multihash,
  pub justification: Multihash,
  pub signature: Signature,
}

pub struct Produced<D>
{
  pub proposer: Pubkey,
  pub parent: Multihash,
  pub height: u64,
  pub signature: (Pubkey, Signature),
  pub data: D,
  pub votes: Vec<Vote>,
}

pub enum NetworkEvent<D>
{
  BlockReceived(block::Produced<D>),
  VoteReceived(Vote),
}
```

I saw this happen:

```
warning: large size difference between variants
  --> src/network.rs:64:3
   |
64 |   VoteReceived(Vote),
   |   ^^^^^^^^^^^^^^^^^^ this variant is 256 bytes
   |
   = note: `#[warn(clippy::large_enum_variant)]` on by default
note: and the second-largest variant is 0 bytes:
  --> src/network.rs:63:3
   |
63 |   BlockReceived(block::Produced<D>),
   |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
help: consider boxing the large fields to reduce the total size of the enum
   |
64 |   VoteReceived(Box<Vote>),
   |                ~~~~~~~~~


```

I expected to see this happen:

This message is invalid because none of the variants are zero bytes, they are actually both over 256 bytes.

### Version

```text
clippy 0.1.58 (02072b4 2022-01-11)
rustc 1.58.0 (02072b482 2022-01-11)
```

tamaroning:
It seems that the size of struct is not measured properly if its field has a generic type.

PR #8453, opened 2022-02-21
fix false positives of large_enum_variant

tamaroning:
fixes: #8321 
The size of enums containing generic type was calculated to be 0. 
I changed [large_enum_variant] so that such enums are not linted.

changelog: none


camsteffen:
Thanks! @bors r+

Issue #8463, opened 2022-02-23
False positive in `ptr_arg` when function pushes to `&mut Vec`


jorgecarleitao:
### Summary

When a function has the signature `&mut Vec<T>` and pushes to it, it cannot be converted to `&mut [T]`. However, in some cases, clippy recommends doing so.

### Lint Name

ptr_arg

### Reproducer

It is not easy to trigger this, so I will use the example I found it:

```
git clone git@github.com:jorgecarleitao/arrow2.git
cargo clippy --features full
```

I saw this happen:

```
warning: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
   --> src/io/ipc/write/serialize.rs:209:12
    |
209 |     nodes: &mut Vec<ipc::FieldNode>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `&mut [ipc::FieldNode]`
    |
    = note: `#[warn(clippy::ptr_arg)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

warning: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
   --> src/io/ipc/write/serialize.rs:298:12
    |
298 |     nodes: &mut Vec<ipc::FieldNode>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `&mut [ipc::FieldNode]`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

warning: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
   --> src/io/ipc/write/serialize.rs:341:12
    |
341 |     nodes: &mut Vec<ipc::FieldNode>,
    |            ^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `&mut [ipc::FieldNode]`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

warning: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
   --> src/io/ipc/write/serialize.rs:642:14
    |
642 |     buffers: &mut Vec<ipc::Buffer>,
    |              ^^^^^^^^^^^^^^^^^^^^^ help: change this to: `&mut [ipc::Buffer]`
    |
[...]

PR #8464, opened 2022-02-23
Fix `ptr_arg`

Jarcho:
fixes: #8463

changelog: Fix `ptr_arg` when multiple arguments are being checked in one function


camsteffen:
Thanks! @bors r+

PR #8471, opened 2022-02-25
new lint that detects useless match expression

### LINKED ISSUE #7040: Remove unnecessary match

J-ZhengLi:
fixes #7040 

changelog: Add new  lint [`needless_match`] under complexity lint group

### DIFF
--- a/clippy_lints/src/matches/mod.rs
+++ b/clippy_lints/src/matches/mod.rs
@@ -21,6 +21,7 @@ mod redundant_pattern_match;
 mod rest_pat_in_fully_bound_struct;
 mod single_match;
 mod wild_in_or_pats;
+mod nop_match;
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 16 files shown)


J-ZhengLi:
Finally got some time to get back to this PR!

I could use some suggestions for my code, such as any missing cases, improvement for readability etc, thanks~ @rustbot ready

clippy_lints/src/matches/mod.rs
```diff
@@ -566,6 +567,49 @@ declare_clippy_lint! {
...
+    /// fn foo() -> Result<(), i32> {
+    ///     result
+    /// }
+    ///
+    /// fn bar() -> Option<i32> {
+    ///     option
+    /// }
+    /// ```
+    #[clippy::version = "1.61.0"]
+    pub NOP_MATCH,
```
llogiq, line 608: NOP is uncommon vocabulary. I'd suggest naming it `needless_match` or `match_identity` (as we already have the quite similar `map_identity`).

clippy_lints/src/matches/mod.rs
```diff
@@ -566,6 +567,49 @@ declare_clippy_lint! {
...
+    correctness,
```
llogiq, line 609: I don't see this as a correctness lint. At best, it's suspicious, or complexity.

tests/ui/nop_match.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[derive(Clone, Copy)]
+enum Choice {
+    A,
+    B,
+    C,
+    D,
+}
+
+fn useless_match(x: i32) {
```
llogiq, line 14: How about adding a test case with a `x => x` arm?

llogiq (commented):
This looks good! :+1: The naming should be more consistent, and we should reconsider the category, and I have a suggestion for further test cases, but otherwise this is merge-worthy.

tests/ui/nop_match.rs
J-ZhengLi: That's a good point, I didn't check for that in my code. :D 

Also, that reminds me, should I be checking `ref x => *x` as well? Or is that too much?

llogiq:
Thank you!

@bors r+

J-ZhengLi:
> Thank you!
> 
> @bors r+

Thank you so much for the approval! But before I change the PR description which possiblly causing this being merged, may I request one last review for the next commit I'm about to push, I have taken your suggestion of adding a `x => x` check along with an extra. 😃

PR #8473, opened 2022-02-25
Added copy lint name button to lint list

### LINKED ISSUE #7959: Add copy lint name function to Clippy's lint list
I often want to copy the lint name when searching for something on Clippy's lint list. It would be helpful to have a "copy lint name"-button next to the name, similar to the "link"-button that currently exist.

image

The HTML document of the page can be found under [`util/gh-pages/index.html`](https://github.com/rust-lang/rust-clippy/blob/master/util/gh-pages/index.html). `cargo dev serve` can be used to test the changes locally.

Main tracking issue #7172

Mastermindaxe:
Fixes #7959 

changelog: Added 'copy lint name' button to lint list

### DIFF
--- a/util/gh-pages/index.html
+++ b/util/gh-pages/index.html
@@ -322,6 +322,7 @@ Otherwise, have a great day =^.^=
                         <div class="panel-title-name">
                             <span>{{lint.id}}</span>
                             <a href="#{{lint.id}}" class="anchor label label-default" ng-click="open[lint.id] = true; $event.stopPropagation()">&para;</a>
+                            <a href="" class="anchor label label-default" title="Copy to clipboard" ng-click="copyToClipboard(lint); $event.stopPropagation()">&#128203;</a>
                         </div>
 
                         <div class="panel-title-addons">
@@ -497,6 +498,11 @@ Otherwise, have a great day =^.^=
                 }
 
                 return true;
+            }
+
+            $scope.copyToClipboard = function (lint) {
+                navigator.clipboard.writeText(lint.id);
+                alert('Copied `' + lint.id + '` to clipboard!');
             }
 
             // Get data


Mastermindaxe:
I added a tooltip for the button and an alert for when you've clicked it

xFrednet:
I wanted this for quite some time now :heart:. I find the alert a bit distracting, but we should give some kind of visual feedback. Maybe an icon change like rustdoc has for items would be better.

image -> image

rustdoc uses `.svg` images for that, we could also consider doing that if icons don't look so good. I think the current clipboard emoji is a little to small :thinking:

Mastermindaxe:
Sounds like a plan. I'm gonna try to get these changes in over the next few days

@camsteffen What exactly do you mean by prefixing? Just putting `clippy::` in front of the lint name in the clipboard? Wouldn't that defeat the purpose?

xFrednet:
Thank you! To solve this, you can just add the prefix in line 504 like this:

```diff
- navigator.clipboard.writeText(lint.id);
+ navigator.clipboard.writeText("clippy::" + lint.id);
```

Or any other variation of JS string formatting. I believe that most people would expect the prefix to be there. In discussions in this repo we often leave it out as the domain is clear, but having it wouldn't hurt. So, with some consideration, I would suggest prefixing it every time. :upside_down_face: 

A second option would be to have a setting or key combination to control the prefix, but I believe that that is more confusing than helpful for 99% of the users.

If there are complaints / suggestions to remove the prefix it's an easy change :)

xFrednet:
Hey @Mastermindaxe, are you still working on this? If you need any support, we're happy to help. :upside_down_face:

Mastermindaxe:
Sorry for being inactive so long. Have been swamped with work and private stuff. I'm hoping that I can squeeze it in some time in the next two weeks. No promises but I'm trying!

xFrednet:
No problem @Mastermindaxe and thank you for the response, if you're busy, you can also reach out. I'd be happy to add the finishing touches and ask another team member to review the final work :upside_down_face:

PR #8476, opened 2022-02-26
Update edition in rustfmt.toml

xFrednet:
I noticed that our `rustfmt.toml` still has the edition 2018 listed. This updates the configuration to use 2021.

This luckily doesn't introduce any formatting changes :upside_down_face: 

changelog: none

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -2,6 +2,6 @@ max_width = 120
 comment_width = 100
 match_block_trailing_comma = true
 wrap_comments = true
-edition = "2018"
+edition = "2021"
 error_on_line_overflow = true
 version = "Two"


giraffate (approved):

giraffate:
@bors r+

Thanks!

xFrednet:
> FYI: https://github.com/rust-lang/rustfmt#rusts-editions
> 
> > Rustfmt is able to pick up the edition used by reading the Cargo.toml file if executed through the Cargo's formatting tool cargo fmt. Otherwise, the edition needs to be specified in rustfmt.toml, e.g., with edition = "2018".

I also found that when I looked up what the configuration actually does. I'm guessing that the value was or is needed for the rust repo as that doesn't use cargo. That's just a guess, though :upside_down_face:. Thanks for the review

PR #8483, opened 2022-02-28
Use `.into_iter()` rather than `.drain(..)`

### LINKED ISSUE #1908: Suggest using into_iter() instead of drain(..) on Vec
It should be possible to rewrite this:
```rust
let mut a = vec![1, 2, 3];
let b = a.drain(..).filter(|item| 1 == item % 2).collect::<Vec<_>>();
println!("{:?}", b);
```
as this:
```rust
let a = vec![1, 2, 3];
let b = a.into_iter().filter(|item| 1 == item % 2).collect::<Vec<_>>();
println!("{:?}", b);
```

ldm0:
Replacing `.drain(..)` with `.into_iter()` makes my project's binary size smaller.

Fixes #1908

Applicability of this suggestion is `MaybeIncorrect` rather than `MachineApplicable` due to the complexity of "checking otherwise usage" X-|

changelog: Add new lint [`iter_with_drain`]

### DIFF
--- a/clippy_lints/src/methods/iter_with_drain.rs
+++ b/clippy_lints/src/methods/iter_with_drain.rs
@@ -1,0 +1,72 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::is_integer_const;
+use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::{
+    higher::{self, Range},
+    SpanlessEq,
+};
+use rustc_ast::ast::RangeLimits;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind, QPath};
+use rustc_lint::LateContext;
+use rustc_span::symbol::{sym, Symbol};
+use rustc_span::Span;
+
+use super::ITER_WITH_DRAIN;
+
+const DRAIN_TYPES: &[Symbol] = &[sym::Vec, sym::VecDeque];
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, recv: &Expr<'_>, span: Span, arg: &Expr<'_>) {
+    let ty = cx.typeck_results().expr_ty(recv).peel_refs();
+    if let Some(drained_type) = DRAIN_TYPES.iter().find(|&&sym| is_type_diagnostic_item(cx, ty, sym)) {
+        // Refuse to emit `into_iter` suggestion on draining struct fields due
+        // to the strong possibility of processing unmovable field.
+        if let ExprKind::Field(..) = recv.kind {
+            return;
+        }
+
+        if let Some(range) = higher::Range::hir(arg) {
+            let left_full = match range {
+                Range { start: Some(start), .. } if is_integer_const(cx, start, 0) => true,
+                Range { start: None, .. } => true,
+                _ => false,
+            };
+            let full = left_full
+                && match range {
+                    Range {
+                        end: Some(end),
+                        limits: RangeLimits::HalfOpen,
+                        ..
+                    } => {
+                        // `x.drain(..x.len())` call
+                        if_chain! {
+                            if let ExprKind::MethodCall(len_path, len_args, _) = end.kind;
+                            if len_path.ident.name == sym::len && len_args.len() == 1;
+                            if let ExprKind::Path(QPath::Resolved(_, drain_path)) = recv.kind;
+                            if let ExprKind::Path(QPath::Resolved(_, len_path)) = len_args[0].kind;
+                            if SpanlessEq::new(cx).eq_path(drain_path, len_path);
+                            then { true }
+                            else { false }
+                        }
+                    },
+                    Range {
+                        end: None,
+                        limits: RangeLimits::HalfOpen,
+                        ..
+                    } => true,
+                    _ => false,
+                };
+            if full {
+                span_lint_and_sugg(
+                    cx,
+                    ITER_WITH_DRAIN,
+                    span.with_hi(expr.span.hi()),
+                    &format!("`drain(..)` used on a `{}`", drained_type),
+                    "try this",
+                    "into_iter()".to_string(),
+                    Applicability::MaybeIncorrect,
+                );
+            }
+        }
+    }
+}
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -32,6 +32,7 @@ mod iter_nth;
 mod iter_nth_zero;
 mod iter_overeager_cloned;
 mod iter_skip_next;
+mod iter_with_drain;
 mod iterator_step_by_zero;
 mod manual_saturating_arithmetic;
 mod manual_str_repeat;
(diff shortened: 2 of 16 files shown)


clippy_lints/src/methods/mod.rs
```diff
@@ -1118,6 +1119,25 @@ declare_clippy_lint! {
...
+    /// // example code where clippy issues a warning
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// // example code which does not raise clippy warning
+    /// ```
+    #[clippy::version = "1.60.0"]
+    pub ITER_WITH_DRAIN,
+    perf,
+    "default lint description"
```
giraffate, line 1138: Can this be filled in?

giraffate (commented):
It looks like that some tests are failing. Can they be addressed?

clippy_lints/src/methods/mod.rs
ldm0: Sorry, forget to push the last commit.

clippy_lints/src/methods/iter_with_drain.rs
```diff
@@ -0,0 +1,76 @@
...
+                    } => true,
+                    _ => false,
+                };
+            if full {
+                span_lint_and_sugg(
+                    cx,
+                    ITER_WITH_DRAIN,
+                    span.with_hi(expr.span.hi()),
+                    "use `into_iter` instead of `drain` for iterating all elements by value in a container",
+                    "try this",
```
flip1995, line 69: The lint message should state the problem, not the suggestion on how to fix it. So 

```rust
"`drain` used on a `Vec`/`VecDeque`", // consider making the `Vec` part dependent on the actual type that is used.
"try using `into_iter` instead",
```

clippy_lints/src/methods/iter_with_drain.rs
```diff
@@ -0,0 +1,76 @@
...
+                        ..
+                    } => {
+                        // `x.drain(..x.len())` call
+                        if_chain! {
+                            if let ExprKind::MethodCall(_, drain_args, _) = expr.kind;
+                            if let ExprKind::MethodCall(len_path, len_args, _) = end.kind;
+                            if len_path.ident.name == sym::len && len_args.len() == 1;
+                            if let ExprKind::Path(QPath::Resolved(_, drain_path)) = drain_args[0].kind;
+                            if let ExprKind::Path(QPath::Resolved(_, len_path)) = len_args[0].kind;
+                            if SpanlessEq::new(cx).eq_path_segments(drain_path.segments, len_path.segments);
```
flip1995, line 51: Isn't there an `eq_path` as well?

clippy_lints/src/methods/iter_with_drain.rs
```diff
@@ -0,0 +1,76 @@
...
+        if_chain! {
+            if let ExprKind::MethodCall(_, drain_args, _) = expr.kind;
+            if drain_args.len() >= 1;
+            if let ExprKind::Field(..) = drain_args[0].kind;
+            then { return }
+        }
+
+        if let Some(range) = higher::Range::hir(arg) {
+            let left_full = match range {
+                Range { start: Some(start), .. } if is_integer_const(cx, start, 0) => true,
```
flip1995, line 33: This could be extended in the future to suggest `.into_iter().skip(N)` for constants `!= 0`.

Doesn't necessarily have to be done in this PR though.

tests/ui/iter_with_drain.rs
```diff
@@ -0,0 +1,51 @@
+#![warn(clippy::iter_with_drain)]
```
flip1995, line 1: Since we're using `span_lint_and_sugg` for generating the suggestion, this should also be tested with 

```suggestion
// run-rustfix
#![warn(clippy::iter_with_drain)]
```

clippy_lints/src/methods/iter_with_drain.rs
ldm0: Added now :-)

clippy_lints/src/methods/iter_with_drain.rs
ldm0: Done.

clippy_lints/src/methods/mod.rs
```diff
@@ -1118,6 +1119,31 @@ declare_clippy_lint! {
...
+    /// let mut foo = vec![0, 1, 2, 3];
+    /// let bar: HashSet<usize> = foo.drain(..).collect();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # use std::collections::HashSet;
+    /// let foo = vec![0, 1, 2, 3];
+    /// let bar: HashSet<usize> = foo.into_iter().collect();
+    /// ```
+    #[clippy::version = "1.60.0"]
```
flip1995, line 1141:
```suggestion
    #[clippy::version = "1.61.0"]
```

flip1995 (commented):
LGTM overall. Just one small comment.

I leave the final review and approval to @giraffate

clippy_lints/src/methods/mod.rs
ldm0: Fixed

clippy_lints/src/methods/iter_with_drain.rs
```diff
@@ -0,0 +1,76 @@
...
+    let ty = cx.typeck_results().expr_ty(recv).peel_refs();
+    if let Some(drained_type) = DRAIN_TYPES.iter().find(|&&sym| is_type_diagnostic_item(cx, ty, sym)) {
+        // Refuse to emit `into_iter` suggestion on draining struct fields due
+        // to the strong possibility of processing unmovable field.
+        if_chain! {
+            if let ExprKind::MethodCall(_, drain_args, _) = expr.kind;
+            if !drain_args.is_empty();
+            if let ExprKind::Field(..) = drain_args[0].kind;
+            then { return }
+        }
```
giraffate, line 29: Can `drain_args[0]` be replaced with `recv`?
```suggestion
        if let ExprKind::Field(..) = recv.kind; {
            return
        }
```

clippy_lints/src/methods/iter_with_drain.rs
```diff
@@ -0,0 +1,76 @@
...
+                        end: Some(end),
+                        limits: RangeLimits::HalfOpen,
+                        ..
+                    } => {
+                        // `x.drain(..x.len())` call
+                        if_chain! {
+                            if let ExprKind::MethodCall(_, drain_args, _) = expr.kind;
+                            if let ExprKind::MethodCall(len_path, len_args, _) = end.kind;
+                            if len_path.ident.name == sym::len && len_args.len() == 1;
+                            if let ExprKind::Path(QPath::Resolved(_, drain_path)) = drain_args[0].kind;
```
giraffate, line 49: Can `drain_args[0]` be replaced with `recv`?
```suggestion
                            if let ExprKind::MethodCall(len_path, len_args, _) = end.kind;
                            if len_path.ident.name == sym::len && len_args.len() == 1;
                            if let ExprKind::Path(QPath::Resolved(_, drain_path)) = recv.kind;
```

clippy_lints/src/methods/iter_with_drain.rs
ldm0: Done

PR #8489, opened 2022-03-01
Add `unnecessary_find_map` lint

### LINKED ISSUE #8467: `find_map`s that could be simplified to `find`
### What it does

Finds occurrences of `find_map` that could be simplified to `find`.

### Lint Name

simplifiable_find_map(?)

### Category

complexity

### Advantage

Essentially the same benefits of the `filter_next` lint (in fact, this could perhaps be an enhancement of that lint).

### Drawbacks

None(?).

### Example

```rust
pub fn main() {
    let xs = [0, 1, 2];
    let odd = xs.iter().find_map(|x| if x % 2 == 1 { Some(x) } else { None });
    println!("{:?}", odd);
}
```

Could be written as:

```rust
pub fn main() {
    let xs = [0, 1, 2];
    let odd = xs.iter().find(|x| *x % 2 == 1);
    println!("{:?}", odd);
}
```

A tricky part could be that `find`'s function argument expects a reference, but `find`'s does not.

smoelius:
This PR adds an `unnecessary_find_map` lint. It is essentially just a minor enhancement of `unnecessary_filter_map`.

Closes #8467

changelog: New lint `unnecessary_find_map`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3498,6 +3498,7 @@ Released 2018-09-13
 [`unit_return_expecting_ord`]: https://rust-lang.github.io/rust-clippy/master/index.html#unit_return_expecting_ord
 [`unnecessary_cast`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
 [`unnecessary_filter_map`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_filter_map
+[`unnecessary_find_map`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_find_map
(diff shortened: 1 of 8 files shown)


tests/ui/unnecessary_filter_map.rs
```diff
@@ -2,26 +2,43 @@
 
 fn main() {
     let _ = (0..4).filter_map(|x| if x > 1 { Some(x) } else { None });
+    let _ = (0..4).find_map(|x| if x > 1 { Some(x) } else { None });
```
llogiq, line 5: Could you factor those tests out into their own ui test file?

llogiq:
Thank you!

@bors r+

PR #8491, opened 2022-03-01
tests: use std::thread::available_parallelism() instead of num_cpus to get thread count

matthiaskrgr:
removes the dependency added in https://github.com/rust-lang/rust-clippy/pull/8451
---

changelog: tests: use std::thread::available_parallelism() instead of num_cpus to get thread count

### DIFF
--- a/tests/compile-test.rs
+++ b/tests/compile-test.rs
@@ -168,7 +168,11 @@ fn run_ui() {
     let _threads = VarGuard::set(
         "RUST_TEST_THREADS",
         // if RUST_TEST_THREADS is set, adhere to it, otherwise override it
-        env::var("RUST_TEST_THREADS").unwrap_or_else(|_| num_cpus::get().to_string()),
+        env::var("RUST_TEST_THREADS").unwrap_or_else(|_| {
+            std::thread::available_parallelism()
+                .map_or(1, std::num::NonZeroUsize::get)
+                .to_string()
+        }),
     );
     compiletest::run_tests(&config);
 }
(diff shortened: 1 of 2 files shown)


tests/compile-test.rs
```diff
@@ -168,7 +168,13 @@ fn run_ui() {
     let _threads = VarGuard::set(
         "RUST_TEST_THREADS",
         // if RUST_TEST_THREADS is set, adhere to it, otherwise override it
-        env::var("RUST_TEST_THREADS").unwrap_or_else(|_| num_cpus::get().to_string()),
+        env::var("RUST_TEST_THREADS").unwrap_or_else(|_| {
+            match std::thread::available_parallelism() {
+                Ok(threads) => threads.get(),
+                _ => 1,
+            }
```
llogiq, line 175: Wouldn't that be the same as

```suggestion
            std::thread::available_parallelism().map_or(1, |t| t.get())
```

matthiaskrgr: To `map_or` not to `map`, that is the question!

Yeah, should be identical, thanks :D

matthiaskrgr: Also apparently we don't have a lint for that yet :P

matthiaskrgr: https://github.com/rust-lang/rust-clippy/issues/8492

flip1995:
I really like what GitHub did to the formatting of the PR title :smile: 

@bors r+

Thanks!

PR #8509, opened 2022-03-06
Fix `unncessary_to_owned` false positive

### LINKED ISSUE #8507: `unnecessary_to_owned` suggestion causes error due to lifetime and boxed trait objects
### Summary

When applying the suggested fix to the following code, the compiler emits an error.
I observed the original behavior in a project using the `warp` framework.

Clippy is a great project, I use it daily and learned a lot from the suggestions!

The issue seems to be with the `build` function, which takes a type with `AsRef<str>` as an argument. But instead of calling `AsRef::as_ref`, the ownership of the argument is transferred to the returned struct. The suggestion removes the `to_string()` call on the input, which causes the returned struct to be bound by the lifetime of the input, thus, the compiler error.
The lint is no longer applied if the `s.to_string()` call is moved to a separate line.

### Lint Name

unnecessary_to_owned

### Reproducer

I tried this code:

```rust
struct Opaque<P>(P);

pub trait Abstracted {}

impl<P> Abstracted for Opaque<P> {}

fn build<P>(p: P) -> Opaque<P>
where
    P: AsRef<str>,
{
    Opaque(p)
}

pub fn test(s: &str) -> Box<dyn Abstracted> {
    Box::new(build(s.to_string()))
}
```

I saw this happen:

```
warning: unnecessary use of `to_string`
  [--> src/main.rs:19:20
19 |     Box::new(build(s.to_string()))
   |                    ^^^^^^^^^^^^^ help: use: `s`
   |
```

The lint should not have triggered.

### Version

```text
rustc 1.61.0-nightly
2022-03-04 9fcbc32053d5084d1de7
```

### Additional Labels

_No response_

smoelius:
Fix #8507

changelog: none

### DIFF
--- a/clippy_lints/src/methods/unnecessary_to_owned.rs
+++ b/clippy_lints/src/methods/unnecessary_to_owned.rs
@@ -2,7 +2,9 @@ use super::implicit_clone::is_clone_like;
 use super::unnecessary_iter_cloned::{self, is_into_iter};
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::ty::{get_associated_type, get_iterator_item_ty, implements_trait, is_copy, peel_mid_ty_refs};
+use clippy_utils::ty::{
+    contains_ty, get_associated_type, get_iterator_item_ty, implements_trait, is_copy, peel_mid_ty_refs,
+};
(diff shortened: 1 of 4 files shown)


giraffate:
Sorry for the late reviewing. I'm a little busy this week, so I'll review this next week.

smoelius:
No problem. Thanks for the update, @giraffate.

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

Issue #7923, opened 2021-11-03
`single_component_path_imports` false positive for macros used in the same module


SabrinaJewson:
Lint name: `single_component_path_imports`

I tried this code:

```rust
m!();
macro_rules! m { () => {} }
use m;
```

I expected to see this happen: Clippy would not warn

Instead, this happened: Clippy showed this warning:

```
warning: this import is redundant
 --> src/lib.rs:3:1
  |
3 | use m;
  | ^^^^^^ help: remove it entirely
  |
  = note: `#[warn(clippy::single_component_path_imports)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports
```
### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.58.0-nightly (29b124802 2021-10-25)
binary: rustc
commit-hash: 29b1248025b19bd132c8047fc710ea9314b9b76b
commit-date: 2021-10-25
host: x86_64-unknown-linux-gnu
release: 1.58.0-nightly
LLVM version: 13.0.0
```

cc #7106

@rustbot label +I-suggestion-causes-error

BartMassey:
Just ran into this.

```rust
pub mod m {
    pub fn f() {
        super::g!();
    }
}

macro_rules! g {
    () => {
        println!("hello");
    };
}
use g;
```

As noted, following Clippy's suggestion fails compilation. `g` needs to be accessible inside `m` and there's no other way (that I am aware of) to make that possible— except `#[macro_export]`, which will make it visible outside the containing crate.

It appears that `single_component_path_imports` is global-only and cannot be suppressed on a case-by-case basis. Compare with #7290.

PR #8537, opened 2022-03-14
Allow `single_component_path_imports` for all macros

xFrednet:
Closes: https://github.com/rust-lang/rust-clippy/issues/7923

It can be useful to have `use macro_name` regardless of the visibility. This removes the visibility filter.

changelog: [`single_component_path_imports`]: no longer triggers on macros


BartMassey:
Dang you are quick. Thanks much!

llogiq:
Thanks!

@bors r+

PR #8544, opened 2022-03-15
Moved ``[try_err]`` to restriction

### LINKED ISSUE #5757: invert check for `clippy::try_err` and disable it inside of try blocks
Right now clippy lints instances of returning bare `Err`'s via `?`

```rust
Err("foo")?; // clippy hates this

return Err("foo".into()); // clippy's preference
```

This perhaps not the best default going forward. Here are the upcoming concerns that make me prefer using `Err("")?`.

## Try Blocks

Once try blocks are stabilized `return Err` and `Err?` will start to mean two different things. A bare return will return to the function scope, where as `?` returns are scoped to the innermost try block. To resolve this we should just not apply this lint to expressions inside of try blocks.

## Error Return Traces and Try Trait interop

In future rust may have error return traces added to the Try trait where each `?` location passes a location to the inner error type to store so it can build up its own lightweight no_std compatible backtrace type that doesn't include any irrelevant frames. These frames would most likely be passed in by the implementation of the `Try` trait for `Result`. By returning via `return Err("".into());` instead of `?` we avoid this codepath and prevent the error return traces from being populated.

In order to ensure the best compatibility with current and future version of the `Try` trait we should prefer utilizing the core error propagation logic when propagating errors rather than recreating a low budget version of that logic manually.

nellorocca:
Fixes #5757 

changelog: Moved try_err lint to rescrition for #5757

### DIFF
--- a/clippy_lints/src/lib.register_all.rs
+++ b/clippy_lints/src/lib.register_all.rs
@@ -289,7 +289,6 @@ store.register_group(true, "clippy::all", Some("clippy_all"), vec![
     LintId::of(transmute::UNSOUND_COLLECTION_TRANSMUTE),
     LintId::of(transmute::WRONG_TRANSMUTE),
     LintId::of(transmuting_null::TRANSMUTING_NULL),
-    LintId::of(try_err::TRY_ERR),
     LintId::of(types::BORROWED_BOX),
     LintId::of(types::BOX_COLLECTION),
     LintId::of(types::REDUNDANT_ALLOCATION),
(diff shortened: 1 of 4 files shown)


flip1995:
Added a "Fixes" to the PR description, so the issue will be closed when this is merged. I think there is nothing more to do on that issue. If we get a `yeet` operator in the future as Josh pointed out, we can open a new issue for that. But until then, no need to do anything about that lint. I also don't think adding an "inverted" lint will bring any benefits.

Thanks for picking this up so quick, right after it was announced in TWiR! And of course: Welcome to Clippy! :paperclip: :heart: 

@bors r+

Issue #8503, opened 2022-03-03
transmute_undefined_repr: False positive on tuples of size 1


dtolnay:
### Summary

I believe `transmute_undefined_repr` intentionally avoids linting on structs / tuple structs containing 1 field, but it currently *does* lint on **tuples** containing 1 field. These should be treated consistently with 1-field tuple structs: either lint both or neither.

### Lint Name

transmute_undefined_repr

### Reproducer

```rust
fn main() {
    let _ = unsafe { std::mem::transmute::<&i32, &(i32,)>(&0) };
}
```

```console
$ cargo clippy

error: transmute into `&(i32,)` which has an undefined layout
 --> src/main.rs:2:22
  |
2 |     let _ = unsafe { std::mem::transmute::<&i32, &(i32,)>(&0) };
  |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[deny(clippy::transmute_undefined_repr)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_undefined_repr
```

```rust
fn main() {
    struct Tuple1(i32);
    let _ = unsafe { std::mem::transmute::<&i32, &Tuple1>(&0) };
}
```

```console
$ cargo clippy
$ # (no lint)
```

### Version

```text
rustc 1.61.0-nightly (8769f4ef2 2022-03-02)
binary: rustc
commit-hash: 8769f4ef2fe1efddd1f072485f97f568e7328f79
commit-date: 2022-03-02
host: x86_64-unknown-linux-gnu
release: 1.61.0-nightly
LLVM version: 14.0.0
```

Issue #8501, opened 2022-03-03
transmute_undefined_repr fails to adjust argument types.


llogiq:
### Summary

If I have a `mem::transmute<U, _>(v)` where the type of `v` is autoderef-adjusted to `U`, the lint ignores this type adjustment and lints as if the concrete type of `v` would be transmuted.

### Reproducer

I tried this code:

```rust
#![warn(clippy::transmute_undefined_repr)]
struct Foo([u8]);
fn main() {
    let x = vec![0_u8, 1, 2, 3];
    // this makes the lint think it's transmuting `&Vec<u8> to &Foo`
    let _ = unsafe { std::mem::transmute::<&[u8], &Foo>(&x) };
}
```

I saw this happen:

```
warning: transmute from `&std::vec::Vec<u8>` which has an undefined layout
  --> src/test_transmute.rs:5:22
   |
55 |             unsafe { mem::transmute::<&[_], _>(&x) }
   |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: requested on the command line with `-W clippy::transmute-undefined-repr`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_undefined_repr
```

I expected to see this happen:

(nothing, the lint should not apply here).

### Version

```text
rustc 1.61.0-nightly (f0c4da499 2022-03-01)
binary: rustc
commit-hash: f0c4da49983aa699f715caf681e3154b445fb60b
commit-date: 2022-03-01
host: x86_64-unknown-linux-gnu
release: 1.61.0-nightly
LLVM version: 14.0.0
```

Issue #8498, opened 2022-03-02
transmute_undefined_repr: false positive on transmutes involving MaybeUninit


dtolnay:
### Summary

MaybeUninit\<T\> has the same layout as T.

### Lint Name

transmute_undefined_repr

### Reproducer

```rust
use std::mem::{self, MaybeUninit};

struct Struct {
    a: u32,
    b: u32,
}

fn f(s: Struct) -> MaybeUninit<Struct> {
    unsafe { mem::transmute::<Struct, MaybeUninit<Struct>>(s) }
}
```

```console
error: transmute from `Struct` which has an undefined layout
 --> src/main.rs:9:14
  |
9 |     unsafe { mem::transmute::<Struct, MaybeUninit<Struct>>(s) }
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[deny(clippy::transmute_undefined_repr)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_undefined_repr
```

> transmute from \`Struct\` which has an undefined layout

Whether or not `Struct` has a defined layout shouldn't be relevant, since MaybeUninit\<T\> has the same layout as T regardless of the layout of T. This is similar to what happened in https://github.com/rust-lang/rust-clippy/issues/8417.

### Version

```text
rustc 1.61.0-nightly (f0c4da499 2022-03-01)
binary: rustc
commit-hash: f0c4da49983aa699f715caf681e3154b445fb60b
commit-date: 2022-03-01
host: x86_64-unknown-linux-gnu
release: 1.61.0-nightly
LLVM version: 14.0.0
```

dtolnay:
More realistic example of this in https://github.com/dtolnay/cxx/blob/c53e4bc81b876c3d19652e9e35d85379239cdc31/src/rust_string.rs.

PR #8547, opened 2022-03-16
More `transmute_undefined_repr` fixes

Jarcho:
fixes: #8498
fixes: #8501
fixes: #8503

changelog: Allow `transumte_undefined_repr` between fat pointers and `(usize, usize)`
changelog: Allow `transumte_undefined_repr` when one side is a union
changelog: Fix `transumte_undefined_repr` on tuples with one non-zero-sized type.


flip1995:
@bors r+

Thanks!

Issue #8542, opened 2022-03-15
needless_match in nightly (2022-03-14) ignores enum arguments


vakaras:
### Summary

When a matched enum has multiple arguments and one of them is replaced, the `needless_match` lint still reports an error even though the match is not needless.

### Lint Name

needless_match

### Reproducer

I tried this code:

```rust
pub enum E {
    Variant1(u32, bool),
    Variant2(u32, bool),
}

impl E {
    pub fn set_b(self, b: bool) -> Self {
        match self {
            E::Variant1(a, _) => E::Variant1(a, b),
            E::Variant2(a, _) => E::Variant2(a, b),
        }
    }
}
```

I saw this happen:

```
> cargo clippy
    Checking needless_match v0.1.0 (/tmp/bla/needless_match)
warning: this match expression is unnecessary
  --> src/lib.rs:9:9
   |
9  | /         match self {
10 | |             E::Variant1(a, _) => E::Variant1(a, b),
11 | |             E::Variant2(a, _) => E::Variant2(a, b),
12 | |         }
   | |_________^ help: replace it with: `self`
   |
   = note: `#[warn(clippy::needless_match)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_match

warning: `needless_match` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s

```

I expected to see this happen: no warning should be emitted because the new value is constructed with a new `b`.

### Version

```text
rustc 1.61.0-nightly (285fa7ecd 2022-03-14)
binary: rustc
commit-hash: 285fa7ecd05dcbfdaf2faaf20400f5f92b39b3c6
commit-date: 2022-03-14
host: x86_64-unknown-linux-gnu
release: 1.61.0-nightly
LLVM version: 14.0.0
```

teor2345:
The lint also triggers on nightly-2022-03-14 for code like this ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=cbbd0259c8b3396ea27c56c119fc64c9)):

```rust
 enum Resultt<T, E> {
     Okk(T),
     Errr(E),
 }
 
 use Resultt::*;
 
 impl<T, E> Resultt<T, E> {
    pub const fn as_ref(self: &Resultt<T, E>) -> Resultt<&T, &E> {
        match self {
            Okk(x) => Okk(x),
            Errr(x) => Errr(x),
        }
[...]

PR #8549, opened 2022-03-16
quick fix of issue#8542: FP in lint `[needless_match]`

J-ZhengLi:
fixes: #8542 

---

changelog: fix FP in lint `[needless_match]`


J-ZhengLi:
I remove the `ref x => *x` match arm check, it seems problematic, and implementing a suitable solution without affecting other "correct" code would possibly make this lint check overly complicated (I think). If anything, maybe having a dedicated lint to check for `[needless_references]`? I don't know... 😢

J-ZhengLi:
I remove the `ref x => *x` match arm check as it's logic seems problematic, I was already unsure about whether I should or not including it when I add the lint. Right now, removing it does seem to be a quick fix of what @teor2345 mentioned in https://github.com/rust-lang/rust-clippy/issues/8542#issuecomment-1068676368 , I just don't see the need to add it back, but maybe I'll do it when someone actually suggesting it. 🤦

clippy_lints/src/matches/needless_match.rs
```diff
@@ -116,44 +126,68 @@ fn strip_return<'hir>(expr: &'hir Expr<'hir>) -> &'hir Expr<'hir> {
...
+                return !same_type_and_consts(results.node_type(local.hir_id), results.expr_ty(match_expr));
+            },
+            // compare match_expr ty with RetTy in `fn foo() -> RetTy`
+            Node::Item(..) => {
+                if let Some(fn_decl) = p_node.fn_decl() {
+                    if let FnRetTy::Return(ret_ty) = fn_decl.output {
+                        return !same_type_and_consts(
+                            hir_ty_to_ty(cx.tcx, ret_ty),
+                            cx.typeck_results().expr_ty(match_expr),
+                        );
```
J-ZhengLi, line 147: Oh, I settled for `hir_ty_to_ty`, will it cause ICE? @flip1995

flip1995: This should be fine. The converted type is guaranteed to be in a function signature and therefore cannot be in a body.

J-ZhengLi:
Also fixes #8595, #8599 

It's kinda harsh seeing my mistakes being addressed multiple times tbh 😭 , so if anyone is not busy at the moment (sorry to bother if you are), please help me reviewing this, thanks~~~

clippy_lints/src/matches/needless_match.rs
```diff
@@ -116,44 +126,68 @@ fn strip_return<'hir>(expr: &'hir Expr<'hir>) -> &'hir Expr<'hir> {
     }
 }
 
+/// Check if the type of the match operand differs with the assigned local or function return type
+///
+/// Can also be used to check for the `let_expr` in `IfLet` as well.
+fn is_type_differ(cx: &LateContext<'_>, match_expr: &Expr<'_>, expr: &Expr<'_>) -> bool {
+    if let Some(p_node) = get_parent_node(cx.tcx, expr.hir_id) {
```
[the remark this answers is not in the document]
llogiq: We have `clippy_utils::is_adjusted(cx, expr)`, which should do the trick more easily.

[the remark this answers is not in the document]
J-ZhengLi: > We have `clippy_utils::is_adjusted(cx, expr)`, which should do the trick more easily.

Well... I later realized that I might still need to keep it this way, because I do need to make sure that the type of match input and output are the same, not just coercion but also as what #8599 suggested. unless there is another util function that I can use to do the check.

clippy_lints/src/matches/needless_match.rs
```diff
@@ -171,27 +191,30 @@ fn pat_same_as_expr(pat: &Pat<'_>, expr: &Expr<'_>) -> bool {
     false
 }
 
-fn has_identical_segments(left_segs: &[PathSegment<'_>], right_segs: &[PathSegment<'_>]) -> bool {
+fn same_segments(left_segs: &[PathSegment<'_>], right_segs: &[PathSegment<'_>]) -> bool {
```
llogiq, line 194: I think we have an `over` function for such things skmewhere in `clippy_utils`.

llogiq (commented):
This looks good, I just have a very small suggestion. r=me when applied.

llogiq:
Thank you!

@bors r+

Issue #1502, opened 2017-02-01
Let_unit_value false positive when return type is type parameter


dtolnay:
```rust
fn main() {
    let _: () = f();
}

fn f<T: Default>() -> T {
    T::default()
}
```

```
warning: this let-binding has unit value. Consider omitting `let _ =`, #[warn(let_unit_value)] on by default
 --> src/main.rs:2:5
  |
2 |     let _: () = f();
  |     ^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://github.com/Manishearth/rust-clippy/wiki#let_unit_value
```

Maybe this specific case would be clearer with `f::<()>()` but I consider this a false positive. Came up in serde_json unit tests where we test that serde_json::from_str can parse various representations of ().

oli-obk:
we need to check whether the rhs contains any type inference.

oli-obk:
I'm not sure how to solve this. This will need some forensics first to figure out whether rustc can tell us if it used inference.

dtolnay:
I think it would work to recognize just this specific pattern: return type consists entirely of a type parameter that is not used by any of the argument types. In that case permit let_unit_value.

hellow554:
This is an interesting thing I stumpled upon today.

I would say that your code @dtolnay should still be linted, because you assign the unit value to a named variable (although it has an underscore, but that doesn't matter here ;) )

I would let clippy suggest doing this:

```rust
let () = f();
```

This would still let the compiler infer the type but telling the user, that he can't do anything with the unit value.
Or can you think of a specific use case where you need a variable name?

dtolnay:
`let _: ()` does not assign a named variable.

hellow554:
I somehow saw `let _a: ()`, don't ask me why... (:

lilyball:
This fix for this doesn't work for me. Approximate code:

```rust
let (_tx, rx) = oneshot::channel();
let handle = spawn(async {
    let () = rx.await.unwrap();
});
```

This triggers the lint in 1.62.0 (now that it defaults to warn). Using `let _: () = …` also triggers it. The let-binding here is most certainly required for type inference.

lilyball:
It sounds like https://github.com/rust-lang/rust-clippy/issues/8998 might cover my case, though I don't know how that is different from the issue described here.

PR #8563, opened 2022-03-17
Don't lint `let_unit_value` when needed for type inferenece

Jarcho:
fixes: #1502

Pinging @dtolnay. I think this is enough to fix the issue. Do you have a good list crates to test this on?

changelog: Don't lint `let_unit_value` when needed for type inference

### DIFF
--- a/tests/ui/let_unit.rs
+++ b/tests/ui/let_unit.rs
@@ -61,3 +61,30 @@ fn multiline_sugg() {
 
 #[derive(Copy, Clone)]
 pub struct ContainsUnit(()); // should be fine
+
+fn _returns_generic() {
+    fn f<T>() -> T {
+        unimplemented!()
+    }
+    fn f2<T, U>(_: T) -> U {
+        unimplemented!()
+    }
+    fn f3<T>(x: T) -> T {
+        x
+    }
+    fn f4<T>(mut x: Vec<T>) -> T {
+        x.pop().unwrap()
+    }
+
+    let _: () = f(); // Ok
+    let x: () = f(); // Lint.
+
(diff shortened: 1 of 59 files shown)


tests/ui/let_unit.rs
```diff
@@ -22,7 +22,7 @@ fn main() {
 
     multiline_sugg();
 
-    let_and_return!(()) // should be fine
+    let_and_return!(()) // should be fine}
```
llogiq, line 25: Is this intended?

llogiq (commented):
I don't know which crates suffered this problem, but code-wise this looks good to me.

tests/ui/let_unit.rs
Jarcho: Fixed.

llogiq:
Sorry for letting this wait for so long, but I lack insight into actual cases in the wild, so I'm not sure if they are all caught 

However, as even if this misses some cases it is still a good improvement, r=me with a rebase.

Jarcho:
This will now handle block expression and branches when checking if type inference is needed.

The only thing I can think of that would also be a false positive would be a something like:
```rust
trait Foo {
    type T;
    fn get_t(&self) -> Self::T;
    fn use_t(&self, _: T);
}
struct Bar;
impl Foo for Bar {
    type T = ();
    ...
}

fn f(bar: Bar) {
    let x = bar.get_t();
    bar.use_t(x);
}
```
But I'm pretty sure that already doesn't lint.

llogiq:
Ok, I might be offline for a while, but r+ when rebased. Thanks and again sorry for the wait.

@bors delegate+

Issue #6356, opened 2020-11-20
False positive for useless_transmute


digama0:
Source: ([SO question](https://stackoverflow.com/q/64902078/890016), [playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=32065895f3b4f7aabebb48d7a9eea3b6))

```rust
#![allow(clippy::needless_lifetimes)]
#![warn(clippy::useless_transmute)]

pub fn insert<'a, K: Eq, V>(this: &'a mut Vec<(K, V)>, key: K, val: V) -> &'a mut V {
    for (key1, val1) in & /* 'b */ mut *this {
        if key == *key1 {
            return unsafe {
                std::mem::transmute::<&/* 'b */ mut V, &/* 'a */ mut V>(val1)
            }
        }
    }
    let this = & /* 'c */ mut *this;
    this.push((key, val));
    &mut this.last_mut().unwrap().1
}
```

Result:
```
warning: transmute from a type (`&mut V`) to itself
 --> src/lib.rs:8:17
  |
8 |                 std::mem::transmute::<&/* 'b */ mut V, &/* 'a */ mut V>(val1)
  |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/lib.rs:2:9
  |
2 | #![warn(clippy::useless_transmute)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_transmute
```

The transmute is not useless because without it the code does not compile. The comments make it a little clearer what's going on: the transmute is being used to change lifetimes, but since this is not represented in the syntax it looks like it's casting a type to itself. My suggested fix would be to skip the lint if the type being cast to itself contains embedded lifetimes (either a shared or unique reference, or a lifetime parameter in the type).

### Meta

- `cargo clippy -V`: clippy 0.0.212 (1773f60 2020-11-08)
- `rustc -Vv`:
  ```
  rustc 1.49.0-nightly (1773f60ea 2020-11-08)
  binary: rustc
  commit-hash: 1773f60ea5d42e86b8fdf78d2fc5221ead222bc1
  commit-date: 2020-11-08
  host: x86_64-unknown-linux-gnu
  release: 1.49.0-nightly
  ```

[...]

Issue #2906, opened 2018-07-08
False positives when using `transmute()` with raw trait object pointers to change trait object lifetimes


okready:
With clippy 0.0.212, I'm getting what appears to be false positives when using `transmute()` to cast either to or from a raw trait object pointer where the input and output types have invariant lifetimes, or if the output type has a longer lifetime such as `'static` (casting using `as` isn't allowed in such cases). No warnings are triggered if the source and target types are both references.

```rust
use std::mem::transmute;

pub trait DummyTrait {}

pub unsafe fn ref_to_ref<'a, 'b>(obj: &'a (DummyTrait + 'a)) -> &'b (DummyTrait + 'b) {
    // No warnings.
    transmute(obj)
}

pub unsafe fn ptr_to_ptr<'a, 'b>(obj: *const (DummyTrait + 'a)) -> *const (DummyTrait + 'b) {
    // Triggers `transmute_ptr_to_ptr` warning.
    transmute(obj)
}

pub unsafe fn ptr_to_ref<'a, 'b>(obj: *const (DummyTrait + 'a)) -> &'b (DummyTrait + 'b) {
    // Triggers `transmute_ptr_to_ref` warning.
    transmute(obj)
}

pub unsafe fn ref_to_ptr<'a, 'b>(obj: &'a (DummyTrait + 'a)) -> *const (DummyTrait + 'b) {
    // Triggers `useless_transmute` warning.
    transmute(obj)
}
```

rail-rain:
The warning has changed due to #5343. Here is the lastest one for your convinience.

```rust
#![allow(clippy::missing_safety_doc)]
// `useless_transmute` is now an allowed-by-default lint.
#![warn(clippy::useless_transmute)]

use std::mem::transmute;

pub trait DummyTrait {}

pub unsafe fn ref_to_ref<'a, 'b>(obj: &'a (dyn DummyTrait + 'a)) -> &'b (dyn DummyTrait + 'b) {
    // Triggers `useless_transmute` warning.
    // Note: this is the false positive introduced #5343
    transmute(obj)
}

pub unsafe fn ptr_to_ptr<'a, 'b>(obj: *const (dyn DummyTrait + 'a)) -> *const (dyn DummyTrait + 'b) {
    // Triggers `useless_transmute` warning.
    // Note: `useless_transmute` took over because `useless_transmute` has a higher priority than `transmute_ptr_to_ptr`
    // besides the false positive above.
    transmute(obj)
}

[...]

PR #8564, opened 2022-03-18
Don't lint `useless_transmute` on types with erased regions

Jarcho:
fixes: #6356

This should get a proper fix at some point, but this at least gets the lint running on some types.

changelog: Don't lint `useless_transmute` on types with erased regions


Alexendoo:
I'll take a look at this one @xFrednet

Alexendoo (commented):
Could add the relevant tests from #2906 as well

Alexendoo (approved):

PR #8571, opened 2022-03-21
add `empty_drop`

### LINKED ISSUE #8352: New lint: empty drop
### What it does

Warns the user on a `Drop` impl that does nothing

### Lint Name

empty_drop

### Category

correctness

### Advantage

- Empty drops do nothing, as the value recursively drops the remaining values after running the `Drop` impl

### Drawbacks

None

### Example

```rust
struct Foo;

impl Drop for Foo {
    fn drop(&mut self) {}
}
```

Could be written as:

```rust
struct Foo;
```

PyroTechniac:
Closes #8352 

changelog: New lint [`empty_drop`]

### DIFF
--- a/clippy_lints/src/empty_drop.rs
+++ b/clippy_lints/src/empty_drop.rs
@@ -1,0 +1,65 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, peel_blocks};
+use if_chain::if_chain;
+use rustc_errors::Applicability;
+use rustc_hir::{Body, ExprKind, Impl, ImplItemKind, Item, ItemKind, Node};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for empty `Drop` implementations.
+    ///
+    /// ### Why is this bad?
+    /// Empty `Drop` implementations have no effect when dropping an instance of the type. They are
+    /// most likely useless. However, an empty `Drop` implementation prevents a type from being
+    /// destructured, which might be the intention behind adding the implementation as a marker.
+    ///
+    /// ### Example
+    /// ```rust
+    /// struct S;
+    ///
+    /// impl Drop for S {
+    ///     fn drop(&mut self) {}
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// struct S;
+    /// ```
+    #[clippy::version = "1.61.0"]
+    pub EMPTY_DROP,
+    restriction,
+    "empty `Drop` implementations"
+}
+declare_lint_pass!(EmptyDrop => [EMPTY_DROP]);
+
+impl LateLintPass<'_> for EmptyDrop {
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if_chain! {
+            if let ItemKind::Impl(Impl {
+                of_trait: Some(ref trait_ref),
+                items: [child],
+                ..
+            }) = item.kind;
+            if trait_ref.trait_def_id() == cx.tcx.lang_items().drop_trait();
+            if let impl_item_hir = child.id.hir_id();
+            if let Some(Node::ImplItem(impl_item)) = cx.tcx.hir().find(impl_item_hir);
+            if let ImplItemKind::Fn(_, b) = &impl_item.kind;
+            if let Body { value: func_expr, .. } = cx.tcx.hir().body(*b);
+            let func_expr = peel_blocks(func_expr);
+            if let ExprKind::Block(block, _) = func_expr.kind;
+            if block.stmts.is_empty() && block.expr.is_none();
+            then {
+                span_lint_and_sugg(
+                    cx,
+                    EMPTY_DROP,
+                    item.span,
+                    "empty drop implementation",
+                    "try removing this impl",
+                    String::new(),
+                    Applicability::MaybeIncorrect
+                );
+            }
+        }
+    }
+}
--- a/tests/ui/empty_drop.rs
+++ b/tests/ui/empty_drop.rs
@@ -1,0 +1,30 @@
+// run-rustfix
+#![warn(clippy::empty_drop)]
+#![allow(unused)]
+
+// should cause an error
+struct Foo;
+
+impl Drop for Foo {
+    fn drop(&mut self) {}
+}
+
+// shouldn't cause an error
+struct Bar;
+
+impl Drop for Bar {
+    fn drop(&mut self) {
+        println!("dropping bar!");
+    }
+}
+
+// should error
+struct Baz;
+
+impl Drop for Baz {
+    fn drop(&mut self) {
+        {}
+    }
+}
+
+fn main() {}
--- a/tests/ui/empty_drop.stderr
+++ b/tests/ui/empty_drop.stderr
@@ -1,0 +1,22 @@
+error: empty drop implementation
+  --> $DIR/empty_drop.rs:8:1
+   |
+LL | / impl Drop for Foo {
+LL | |     fn drop(&mut self) {}
+LL | | }
+   | |_^ help: try removing this impl
+   |
+   = note: `-D clippy::empty-drop` implied by `-D warnings`
+
+error: empty drop implementation
+  --> $DIR/empty_drop.rs:24:1
+   |
+LL | / impl Drop for Baz {
+LL | |     fn drop(&mut self) {
+LL | |         {}
+LL | |     }
+LL | | }
+   | |_^ help: try removing this impl
+
+error: aborting due to 2 previous errors
+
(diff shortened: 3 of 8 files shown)


PyroTechniac:
After some discussion with fellow devs, they advise I switch this from `correctness` to `style`, but would like input from the devs

flip1995:
As you can see on many Clippy tests, lints behave differently on types that have a `Drop` implementation. The same is true for `rustc`. So even though this `Drop` might not do anything, it might still be used as a marker. (I'm not sure if this has real world use cases, but removing the `Drop` impl changes compiler behavior.)

One example where you might want to use this is [`rustc --explain E0509`](https://godbolt.org/z/9K9MrGrjh). You can implement an empty `Drop` impl to prevent any code from moving fields out of a struct.

Since there isn't any alternative to implementing `Drop` to tell the compiler that a type has drop glue / a destructor, I'm not sure if this lint should be warn-by-default. If you implement an empty `Drop` you probably have a good reason to do so. And if the empty `Drop` impl is by accident it doesn't do harm during runtime.

So for me the question is if this lint should be `pedantic` or `restriction`. I tend towards `restriction`.

PyroTechniac:
I have changed the lint to be a restriction lint

clippy_lints/src/empty_drop.rs
```diff
@@ -0,0 +1,54 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+// use clippy_utils::ty::is_type_lang_item;
```
flip1995, line 2:
```suggestion
```

clippy_lints/src/empty_drop.rs
```diff
@@ -0,0 +1,54 @@
...
+use rustc_hir::{Body, ExprKind, Impl, ImplItemKind, Item, ItemKind, Node};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for empty `Drop` implementations.
+    ///
+    /// ### Why is this bad?
+    /// Empty `Drop` implementations do nothing except hurt code readability. They also prevent the type from being destructured.
```
flip1995, line 13: This is contradicting itself. I would formulate it like this:

```suggestion
    /// Empty `Drop` implementations have no effect when dropping an instance of the type. They are most likely useless. However, an empty `Drop` implementation prevents a type from being destructured, which might be the intention behind adding the implementation as a marker. 
```

This obviously needs formatting.

clippy_lints/src/empty_drop.rs
```diff
@@ -0,0 +1,54 @@
...
+                items: [child],
+                ..
+            }) = item.kind;
+            if trait_ref.trait_def_id() == cx.tcx.lang_items().drop_trait();
+            if let impl_item_hir = child.id.hir_id();
+            if let Some(Node::ImplItem(impl_item)) = cx.tcx.hir().find(impl_item_hir);
+            if let ImplItemKind::Fn(_, b) = &impl_item.kind;
+            if let Body { value: func_expr, .. } = cx.tcx.hir().body(*b);
+            if let ExprKind::Block(block, _) = func_expr.kind;
+            if block.stmts.is_empty() && block.expr.is_none();
```
flip1995, line 48: Funny edge case:

```rust
fn drop(&mut self) {{}}
```

Can you add a test if this is also linted? If not, you might want to use the `clippy_utils::peel_blocks` function on the `func_expr` before checking if the block is empty.

clippy_lints/src/empty_drop.rs
```diff
@@ -0,0 +1,54 @@
...
+            then {
+                span_lint_and_help(cx, EMPTY_DROP, item.span, "empty drop implementation", None, "try removing this impl");
```
flip1995, line 50: This could be a `span_lint_and_sugg` and an empty sugg with `Applicability::MaybeIncorrect`.

tests/ui/crashes/ice-7410.rs
```diff
@@ -6,6 +6,7 @@
 #![no_std]
 #![allow(clippy::if_same_then_else)]
 #![allow(clippy::redundant_pattern_matching)]
+#![allow(clippy::empty_drop)]
```
flip1995, line 9: Since this is a restriction lint now, those allows on unrelated tests shouldn't be necessary anymore.

tests/ui/empty_drop.rs
```diff
@@ -0,0 +1,12 @@
+#![warn(clippy::empty_drop)]
```
flip1995, line 1: After using the `span_lint_and_sugg` method, we should test the suggestion with `run-rustfix`

```suggestion
// run-rustfix
#![warn(clippy::empty_drop)]
```

flip1995 (commented):
Thanks I only noticed an edge case that you might want to address and some comments about the documentation. Great first contribution overall!

clippy_lints/src/empty_drop.rs
```diff
@@ -0,0 +1,56 @@
...
+            if let impl_item_hir = child.id.hir_id();
+            if let Some(Node::ImplItem(impl_item)) = cx.tcx.hir().find(impl_item_hir);
+            if let ImplItemKind::Fn(_, b) = &impl_item.kind;
+            if let Body { value: func_expr, .. } = cx.tcx.hir().body(*b);
+            let func_expr = peel_blocks(&func_expr);
+            if let ExprKind::Block(block, _) = func_expr.kind;
+            if block.stmts.is_empty() && block.expr.is_none();
+            then {
+                // span_lint_and_help(cx, EMPTY_DROP, item.span, "empty drop implementation", None, "try removing this impl");
+                span_lint_and_sugg(cx, EMPTY_DROP, item.span, "empty drop implementation", "try removing this impl", String::new(), Applicability::MaybeIncorrect);
```
flip1995, line 52:
```suggestion
                span_lint_and_sugg(
                    cx,
                    EMPTY_DROP,
                    item.span,
                    "empty drop implementation",
                    "try removing this impl",
                    String::new(),
                    Applicability::MaybeIncorrect,
                );
```

clippy_lints/src/empty_drop.rs
```diff
@@ -0,0 +1,56 @@
...
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_errors::Applicability;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for empty `Drop` implementations.
+    ///
+    /// ### Why is this bad?
+    /// Empty `Drop` implementations have no effect when dropping an instance of the type. They are most likely useless. However, an empty `Drop` implementation prevents a type from being destructured, which might be the intention behind adding the implementation as a marker.
```
flip1995, line 13: This needs formatting.

tests/ui/empty_drop.stderr
```diff
@@ -0,0 +1,20 @@
+error: empty drop implementation
+  --> $DIR/empty_drop.rs:8:1
+   |
+LL | / impl Drop for Foo {
+LL | |     fn drop(&mut self) {}
+LL | | }
+   | |_^ help: try removing thsi impl
```
flip1995, line 7: Run `cargo dev bless` to update those tests.

PyroTechniac:
Do I still need the `main` fn in my tests?

flip1995:
> Do I still need the `main` fn in my tests?

Not sure if you can completely remove it (I don't think so), but you can leave it empty.

flip1995 (approved):
LGTM. With squashed commits and CI passing, this should be ready to get merged.

PyroTechniac:
I'm not quite a git master, is there anything I have to do for squashing commits?

flip1995:
Let me do it for you then. You can do this with `git rebase -i master` and then force pushing the new commit. Before force pushing you should make sure that your diff with the remote (`origin/empty-drop` in this case is empty though)

flip1995:
@bors r+

Thanks!

flip1995:
This will need another rebase.

PyroTechniac:
Sorry! Got caught up this weekend, I'll get to this sometime tonight however.

PyroTechniac:
And by tonight I meant 2 days later, sorry for the delay!

flip1995 (approved):
Thanks, this LGTM now. Great first contribution.

Can you squash your commits please. While you're at it you can also rebase them on the current master to get rid of merge commits. The Rust porject has a [no-merge-policy](https://github.com/rust-lang/rustc-dev-guide/blob/1b56b094728188d1970ef94e727b6c384ac98b7c/src/git.md#no-merge-policy)

You can do this with:
```
git rebase -i master
# An editor will open, where you can "squash" your commits
# You might have to fix some conflicts
git diff -- clippy_lints/src/empty_drop.rs  # Should be empty
```

I recommend that you run the usual commands again:

```
cargo dev update_lints
cargo dev fmt
cargo test
```

If all that passes, you can force push your changes

```
git push --force-with-lease
```

flip1995:
This rebased seemed to have gone wrong. Let me try to fix that.

PyroTechniac:
Thank you! Someone had made a fixed branch for me but I didn't get the chance to fix it yesterday

Issue #8570, opened 2022-03-21
`needless_join` for `Iter<String>`


yoav-lavi:
### What it does

When transforming an `Iter<String>` to `String`, you can either use:

`.collect::<Vec<String>>().join("")`:
```rs
let vector = vec!["hello",  "world"];
let output = vector.iter().map(|item| item.to_uppercase()).collect::<Vec<String>>().join("");
println!("{}", output);
```

or 

`.collect::<String>()`:
```rs
let vector = vec!["hello",  "world"];
let output = vector.iter().map(|item| item.to_uppercase()).collect::<String>();
println!("{}", output);
```

The latter being terser and more performant:

```rs
fn criterion_benchmark(criterion: &mut Criterion) {
    criterion.bench_function("1", |bencher| {
        bencher.iter(|| {
            vec!["hello", "world"]
                .iter()
                .map(|item| item.to_uppercase())
                .collect::<Vec<String>>()
                .join("")
        })
    });
    criterion.bench_function("2", |bencher| {
        bencher.iter(|| {
            vec!["hello", "world"]
                .iter()
                .map(|item| item.to_uppercase())
                .collect::<String>()
        })
    });
}
```

```
1                       time:   [155.79 ns 155.97 ns 156.16 ns]                   
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe
slope  [155.79 ns 156.16 ns] R^2            [0.9971885 0.9971738]
mean   [155.87 ns 156.58 ns] std. dev.      [672.36 ps 2.9345 ns]
median [156.12 ns 156.26 ns] med. abs. dev. [209.02 ps 650.40 ps]

2                       time:   [118.11 ns 118.20 ns 118.28 ns]                   
Found 8 outliers among 100 measurements (8.00%)
  8 (8.00%) high severe
slope  [118.11 ns 118.28 ns] R^2            [0.9987699 0.9987697]
mean   [118.31 ns 118.87 ns] std. dev.      [794.06 ps 2.0461 ns]
median [117.96 ns 118.43 ns] med. abs. dev. [154.11 ps 692.31 ps]
```

I suggest adding a new lint called `needless_join` that detects the case of `.collect<Vec<String>>.join("")` on a `String` iterator.

Thank you for your consideration!

### Lint Name

needless_join

### Category

perf

### Advantage

- It is shorter
- It is more performant
- The intent is clearer

### Drawbacks

None that I'm aware of

[...]

PR #8573, opened 2022-03-22
`unnecessary_join` lint

yoav-lavi:
changelog: Adds a lint called ``[`unnecessary_join`]`` that detects cases of `.collect::<Vec<String>>.join("")` or `.collect::<Vec<_>>.join("")` on an iterator, suggesting `.collect::<String>()` instead

Fixes: https://github.com/rust-lang/rust-clippy/issues/8570


flip1995:
To get the type of the turbofish inside the `collect` call, I would just check the type of the expr `.iter()...collect()`. You may have to use `expr_ty_adjusted()` to get the actual type.

Looking at your implementation: You try to get the `parent` expression inside the `check`  function. You should be able to get the iterator from the `methods/mod.rs` file. The iterator expression preceding the `collect` call should be `_arg` when matching to the `"join"` method name.

flip1995:
> Any input on how I would check that `join` is receiving an empty `&str` / `String`?

For `""` you can just match again a `ExprKind::Lit` and check if it is the empty string. I'm not sure if it is worth handling the other cases, because writing code like this is really questionable in the first place. A separate lint about using an owned empty string where also a `""` would be possible might be better to put time in than trying to cover all cases in this lint.

> Any comment on the validity of the lint itself? I opened the issue recently so I want to make sure that I'm not missing anything that would make this invalid.

This lint sounds reasonable to me. I can't really think of when you would want to use `.join("")` over directly collecting to a string. Maybe there is a case where you can collect into `Vec<String>` but not directly into `String`? This could be covered by checking for a `FromIterator<[iterator-item-type]>` implementation for `String` and only trigger the lint if there is one.

yoav-lavi:
@flip1995 made some updates and added tests, seems to work correctly

clippy_lints/src/methods/mod.rs
```diff
@@ -2049,6 +2050,31 @@ declare_clippy_lint! {
...
+    /// let output = vector.iter().map(|item| item.to_uppercase()).collect::<Vec<String>>().join("");
+    /// println!("{}", output);
+    /// ```
+    /// The correct use would be:
+    /// ```rust
+    /// let vector = vec!["hello",  "world"];
+    /// let output = vector.iter().map(|item| item.to_uppercase()).collect::<String>();
+    /// println!("{}", output);
+    /// ```
+    #[clippy::version = "pre 1.29.0"]
```
yoav-lavi, line 2072: Not completely sure what this should be

flip1995: The current nightly the lint is added in. This will probably be `1.61` or `1.62` depending on if we get this merged before or after Thursday.

clippy_lints/src/methods/unnecessary_join.rs
```diff
@@ -0,0 +1,45 @@
...
+use rustc_middle::ty;
+use rustc_span::sym;
+
+use super::UNNECESSARY_JOIN;
+
+pub(super) fn check<'tcx>(context: &LateContext<'tcx>, join: &'tcx hir::Expr<'tcx>) {
+    let applicability = Applicability::MachineApplicable;
+
+    if_chain! {
+        if let ExprKind::MethodCall(_path, [join_self_arg, join_arg], _span) = &join.kind;
```
flip1995: You have this information in the `mod.rs` file and can just pass it to the `check` function.

clippy_lints/src/methods/unnecessary_join.rs
```diff
@@ -0,0 +1,45 @@
...
+        let collect_output_type = context.typeck_results().expr_ty(join_self_arg);
+        // the current join method is being called on a vector
+        // e.g .join("")
+        if is_type_diagnostic_item(context, collect_output_type, sym::Vec);
```
flip1995: I think this check is redundant. I would just move the check for `::<Vec<String>>` up and remove this check.

clippy_lints/src/methods/unnecessary_join.rs
```diff
@@ -0,0 +1,45 @@
...
+        if let ty::Ref(_, ref_type, _) = collect_output_adjusted_type.kind();
+        if let ty::Slice(slice) = ref_type.kind();
+        if is_type_diagnostic_item(context, *slice, sym::String);
+        then {
+            span_lint_and_sugg(
+                context,
+                UNNECESSARY_JOIN,
+                join.span,
+                &format!(
+                    "called `.collect<Vec<String>>().join(\"\")` on a {}. Using `.collect::<String>()` is clearer and more concise", collect_output_type,
```
flip1995: No need to explain the solution in the lint message and again in the suggestion.

```suggestion
                    "called `.collect<Vec<String>>().join(\"\")` on a {}", collect_output_type,
```

And in the suggestion text:

```rust
"try using `collect::<String>()`"
```

tests/ui/unnecessary_join.rs
```diff
@@ -0,0 +1,35 @@
+#![warn(clippy::unnecessary_join)]
```
flip1995, line 1:
```suggestion
// run-rustfix

#![warn(clippy::unnecessary_join)]
```

To get rustfix to check that the suggestion works.

clippy_lints/src/methods/unnecessary_join.rs
```diff
@@ -0,0 +1,45 @@
...
+        // the turbofish for collect is ::<Vec<String>>
+        let collect_output_adjusted_type = &context.typeck_results().expr_ty_adjusted(join_self_arg);
+        if let ty::Ref(_, ref_type, _) = collect_output_adjusted_type.kind();
+        if let ty::Slice(slice) = ref_type.kind();
+        if is_type_diagnostic_item(context, *slice, sym::String);
+        then {
+            span_lint_and_sugg(
+                context,
+                UNNECESSARY_JOIN,
+                join.span,
```
flip1995, line 35: You will have to limit the span from the start of `.collect::<Vec<_>>()` to the end of `.join("")`. See other method lints on how to do this. It will involve calling `with_hi` on the correct span.

flip1995 (commented):
I will have to think about the case where directly collecting into a `String` might not be possible.

But until then, I left some comments.

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Fair enough, will do

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Done

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Specifically to be able to show the method chain until `collect` without `collect` and `join`? Will look into this, couldn't find an example that behaved that way

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Done

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Just to verify, do you mean for the fix example or for the error message?

yoav-lavi: Done for the span itself, let me know regarding the message

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Done

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Actually I still need to include the . in .collect

yoav-lavi: Seems to behave correctly now

yoav-lavi: Although I'm using `with_lo` and it's pointing to the newline character in the line before, it's fixing correctly but I'm assuming there's a span that includes the . in collect without the newline

yoav-lavi: Looking into this

yoav-lavi:
@flip1995 apparently this can be slower in specific cases: https://github.com/rust-lang/rust/pull/95216

clippy_lints/src/methods/unnecessary_join.rs
yoav-lavi: Done

CHANGELOG.md
```diff
@@ -3508,7 +3508,6 @@ Released 2018-09-13
 [`unnecessary_filter_map`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_filter_map
 [`unnecessary_find_map`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_find_map
 [`unnecessary_fold`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_fold
-[`unnecessary_join`]: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_join
```
yoav-lavi: @flip1995 not sure why these were removed by `cargo dev update_lints`

yoav-lavi:
I suggest also adding `.DS_Store` to the `.gitignore` file, can do this in this PR or in a new one if this is agreed upon

yoav-lavi:
I'm not sure whether I'm supposed to use "r?" to transfer the PR to "waiting for review", will try and let me know if this is wrong

yoav-lavi:
r?

yoav-lavi:
@rust-highfive r?

CHANGELOG.md
flip1995: Me neither. Without registering the lint properly it won't do anything. So try re-running it.

flip1995:
The correct command is 

@rustbot ready

yoav-lavi:
Thanks @flip1995!

CHANGELOG.md
yoav-lavi: I did multiple times, it always removes it. Could this be related to the fact that this lint is in the methods folder?

yoav-lavi: I could add the lines manually but they'd just be erased on the next commit

clippy_lints/src/methods/mod.rs
```diff
@@ -2049,6 +2050,31 @@ declare_clippy_lint! {
...
+    /// ```
+    /// The correct use would be:
+    /// ```rust
+    /// let vector = vec!["hello",  "world"];
+    /// let output = vector.iter().map(|item| item.to_uppercase()).collect::<String>();
+    /// println!("{}", output);
+    /// ```
+    #[clippy::version = "pre 1.61.0"]
+    pub UNNECESSARY_JOIN,
+    perf,
```
flip1995, line 2074: With the perf analysis you did in the Rust PR, I don't think we can put this lint in "perf". Loop unrolling usually gives a pretty big perf benefit. 

I'm happy to include this lint in Clippy, because it might improve code size and is easier to read. 

Maybe the different code generation needs to be fixed in rustc? But until then I would put it in `pedantic`. I don't want to put it in `style` and make it warn-by-default, because it could have a negative perf impact.

Please also add a `Known Problems` section to the lint documentation explaining that using `.collect::<String>` might prevent loop unrolling and thus have a negative perf impact.

flip1995 (commented):
Impl looks great!

clippy_lints/src/methods/mod.rs
yoav-lavi: That sounds reasonable, will add this

yoav-lavi: Done

yoav-lavi: Not sure about the "most cases" part since logically it should be and is in the cases I've measured, but it's hard to know what usually happens. On the other hand, using wording like "can be" would be underselling it a bit IMO

flip1995: Your formulation sounds ideal!

clippy_lints/src/methods/mod.rs
```diff
@@ -2049,6 +2050,35 @@ declare_clippy_lint! {
...
+    /// println!("{}", output);
+    /// ```
+    /// ### Known problems
+    /// While `.collect::<String>()` is more performant in most cases, there are cases where
+    /// using `.collect::<String>()` over `.collect::<Vec<String>>().join("")`
+    /// will prevent loop unrolling and will result in a negative performance impact.
+    #[clippy::version = "1.61.0"]
+    pub UNNECESSARY_JOIN,
+    pedantic,
+    r#"using `.collect::<Vec<String>>().join("")` on an iterator"#
```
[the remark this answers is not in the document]
yoav-lavi: Yep, that was the issue. Fixed

yoav-lavi:
@flip1995 I think that's everything we discussed, there's the issue of cases where directly collecting into a String might not be possible, but I'm not sure that's still an issue if this is in `pedantic`

flip1995:
I think it is pretty much impossible to add a `Join` implementation that would break this lint: [Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=659eb24a6ccdc07374bde8fc5cc0de6d). So this shouldn't be a problem.

flip1995 (approved):
Thanks for addressing the review so quickly! Please squash (some of) your commits and then we can merge this PR.

yoav-lavi:
Sorry closed by mistake, reopening

PR #8576, opened 2022-03-24
Add `crate_in_macro_def` lint

### LINKED ISSUE #4798: Lint request: Warn when importing inside a macro and not using the $ keyword.
I suggest adding a lint for when a programmer makes a macro, imports something into said macro and forgets `$` before `crate` keyword. Exporting macros across crates can be a hassel, atleast for me.

smoelius:
This PR adds a lint to check for `crate` as opposed to `$crate` used in a macro definition.

I think this can close #4798. That issue focused on the case where the macro author "imports something into said macro."

But I think use of `crate` is likely to be a bug whether it appears in a `use` statement or not. There could be some use case I am failing to see, though. (cc: @nilscript @flip1995)

changelog: `crate_in_macro_def`

### DIFF
--- a/clippy_lints/src/crate_in_macro_def.rs
+++ b/clippy_lints/src/crate_in_macro_def.rs
@@ -1,0 +1,100 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use rustc_ast::ast::MacroDef;
+use rustc_ast::node_id::NodeId;
+use rustc_ast::token::{Token, TokenKind};
+use rustc_ast::tokenstream::{TokenStream, TokenTree};
+use rustc_errors::Applicability;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for use of `crate` as opposed to `$crate` in a macro definition.
+    ///
+    /// ### Why is this bad?
+    /// `crate` refers to macro call's crate, whereas `$crate` refers to the macro
+    /// definition's crate. Rarely is the former intended. See:
+    /// https://doc.rust-lang.org/reference/macros-by-example.html#hygiene
+    ///
+    /// ### Example
+    /// ```rust
+    /// macro_rules! print_message {
+    ///     () => {
+    ///         println!("{}", crate::MESSAGE);
+    ///     };
+    /// }
+    /// pub const MESSAGE: &str = "Hello!";
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// macro_rules! print_message {
+    ///     () => {
+    ///         println!("{}", $crate::MESSAGE);
+    ///     };
+    /// }
+    /// pub const MESSAGE: &str = "Hello!";
+    /// ```
+    #[clippy::version = "1.61.0"]
+    pub CRATE_IN_MACRO_DEF,
+    correctness,
+    "using `crate` in a macro definition"
+}
+declare_lint_pass!(CrateInMacroDef => [CRATE_IN_MACRO_DEF]);
+
+impl EarlyLintPass for CrateInMacroDef {
+    fn check_mac_def(&mut self, cx: &EarlyContext<'_>, macro_def: &MacroDef, _: NodeId) {
+        let tts = macro_def.body.inner_tokens();
+        if let Some(span) = contains_unhygienic_crate_reference(&tts) {
+            span_lint_and_sugg(
+                cx,
+                CRATE_IN_MACRO_DEF,
+                span,
+                "reference to the macro call's crate, which is rarely intended",
+                "if reference to the macro definition's crate is intended, use",
+                String::from("$crate"),
+                Applicability::MachineApplicable,
+            );
+        }
+    }
+}
+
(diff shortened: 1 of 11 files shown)


tests/ui/crate_in_macro_def_allow.rs
```diff
@@ -0,0 +1,19 @@
+#![warn(clippy::crate_in_macro_def)]
+
+#[macro_use]
+mod intentional {
+    // For cases where use of `crate` is intentional, applying `allow` to the macro definition
+    // should suppress the lint.
+    #[allow(clippy::crate_in_macro_def)]
```
Alexendoo, line 7: This can go in the same test file as above or be removed, as you aren't having to do something specific to get it working

Could be worth noting in the description that the allow attribute has to go on the `macro_rules` itself, ie

```rust
#[allow(clippy::crate_in_macro_def)]
macro_rules! ok { ... crate::foo ... }

macro_rules! lints {
...
    // Still errors since the allow attribute is part of the macro token stream
     #[allow(clippy::crate_in_macro_def)]
    crate::foo
...
}
```

clippy_lints/src/crate_in_macro_def.rs
```diff
@@ -0,0 +1,100 @@
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// macro_rules! print_message {
+    ///     () => {
+    ///         println!("{}", crate::MESSAGE);
+    ///     };
+    /// }
+    /// pub const MESSAGE: &str = "Hello!";
+    /// ```
```
Alexendoo, line 28: For internal macros `crate` is fine, you could check for a `macro_export` attribute using `check_item` to avoid some false positives

tests/ui/crate_in_macro_def_allow.rs
smoelius: > This can go in the same test file as above or be removed

My concern is that someone uses `crate` intentionally, they apply `allow` to the macro definition, and then a future change to Clippy breaks their uses of `allow`. E.g., it would be annoying to have to use `allow` at every call site. The test is meant to help prevent this possibility.

clippy_lints/src/crate_in_macro_def.rs
smoelius: I am embarrassed that I didn't think of this. :grimacing:

smoelius:
Thank you for the review @Alexendoo. I think [cb307bb](https://github.com/rust-lang/rust-clippy/commit/cb307bbfcd123d3bae8f210f2b51eebe68525c27) addresses all of your comments.

clippy_lints/src/crate_in_macro_def.rs
```diff
@@ -0,0 +1,126 @@
...
+    ///
+    /// Note that if the use of `crate` is intentional, an `allow` attribute can be applied to the
+    /// macro definition, e.g.:
+    /// ```rust,ignore
+    /// #[allow(clippy::crate_in_macro_def)]
+    /// macro_rules! ok { ... crate::foo ... }
+    /// ```
+    #[clippy::version = "1.61.0"]
+    pub CRATE_IN_MACRO_DEF,
+    correctness,
```
llogiq, line 48: I don't think this should be a correctness lint. Correctness implies the lint level is `deny`. While the lint will be right in the general case, there may be exceptions and I would not want to stop the build for that. My suggestion here would be to make it a `suspicious` lint (which is `warn` by default).

clippy_lints/src/crate_in_macro_def.rs
```diff
@@ -0,0 +1,126 @@
...
+            if item.attrs.iter().any(is_macro_export);
+            if let ItemKind::MacroDef(macro_def) = &item.kind;
+            let tts = macro_def.body.inner_tokens();
+            if let Some(span) = contains_unhygienic_crate_reference(&tts);
+            then {
+                span_lint_and_sugg(
+                    cx,
+                    CRATE_IN_MACRO_DEF,
+                    span,
+                    "reference to the macro call's crate, which is rarely intended",
```
llogiq, line 65:
```suggestion
                    "`crate` references the macro call's crate",
```

clippy_lints/src/crate_in_macro_def.rs
```diff
@@ -0,0 +1,126 @@
...
+                    "if reference to the macro definition's crate is intended, use",
```
llogiq, line 66:
```suggestion
                    "to reference the macro definition's crate, use",
```

clippy_lints/src/crate_in_macro_def.rs
```diff
@@ -0,0 +1,126 @@
...
+fn is_macro_export(attr: &Attribute) -> bool {
+    if_chain! {
+        if let AttrKind::Normal(attr_item, _) = &attr.kind;
+        if let [segment] = attr_item.path.segments.as_slice();
+        if segment.ident.name == sym::macro_export;
+        then {
+            true
+        } else {
+            false
+        }
```
llogiq, line 84:
```suggestion
        then {
            segment.ident.name == sym::macro_export
        } else {
            false
        }
```

llogiq (commented):
This looks good. I have but a small number of suggestions. The only thing that is really worrying me is the lint level.

llogiq:
Thank you! @bors r+

smoelius:
@llogiq Sorry, sorry. One more was needed to fix the error message.

PR #8588, opened 2022-03-26
`indexing_slicing` should not fire if a valid array index comes from a constant function that is evaluated at compile-time

### LINKED ISSUE #8348: `indexing_slicing` should not fire if a valid array index comes from a constant function that is evaluated at compile-time
1. From a inline constant block (https://github.com/rust-lang/rust-clippy/issues/8348#issuecomment-1021303011).

```rust
#![feature(inline_const)]
#![forbid(clippy::indexing_slicing)]

const FOO: [i32; 1] = [1];

const fn idx() -> usize {
    0
}

pub fn main() {
    assert_eq!(FOO[const { idx() }], 1);
}
```

2. From a `const` declaration (https://github.com/rust-lang/rust-clippy/issues/8348#issuecomment-1021359538).

```rust
#![forbid(clippy::indexing_slicing)]

const FOO: [i32; 1] = [1];
const REF: &i32 = &FOO[idx()];

const fn idx() -> usize {
    0
}

pub fn main() {
  assert_eq!(REF, &1);
}
```

pitaj:
fix #8348

changelog: [`indexing_slicing`] fewer false positives in `const` contexts and with `const` indices

### DIFF
--- a/clippy_lints/src/indexing_slicing.rs
+++ b/clippy_lints/src/indexing_slicing.rs
@@ -96,6 +96,10 @@ declare_lint_pass!(IndexingSlicing => [INDEXING_SLICING, OUT_OF_BOUNDS_INDEXING]
 
 impl<'tcx> LateLintPass<'tcx> for IndexingSlicing {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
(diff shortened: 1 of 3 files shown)


pitaj:
@flip1995 bump

flip1995:
@bors r+

Thanks!

Sorry for the delay.

PR #8592, opened 2022-03-26
Do not fire `panic` in a constant environment

c410-f3r:
Let rustc handle panics in constant environments.

Since https://github.com/rust-lang/rust-clippy/issues/8348 I initially thought that such modification would require a lot of work but thanks to https://github.com/rust-lang/rust-clippy/pull/8588 I now know that it is not the case.

### DIFF
--- a/tests/ui/panicking_macros.rs
+++ b/tests/ui/panicking_macros.rs
@@ -1,7 +1,22 @@
+#![allow(clippy::assertions_on_constants, clippy::eq_op,)]
+#![feature(inline_const)]
 #![warn(clippy::unimplemented, clippy::unreachable, clippy::todo, clippy::panic)]
-#![allow(clippy::assertions_on_constants, clippy::eq_op)]
 
 extern crate core;
+
+const _: () = {
+    if 1 == 0 {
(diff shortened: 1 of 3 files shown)


tests/ui/panicking_macros.rs
```diff
@@ -1,8 +1,23 @@
...
 extern crate core;
 
+const _: () = {
+    if 1 == 0 {
+        panic!("A balanced diet means a cupcake in each hand");
+    }
+};
+
+fn inline_const() {
+    let _ = const {
```
flip1995, line 14: Why not just make the function `const`? No need to enable a feature in a test.

c410-f3r: Constant functions can be called in non-constant environments, therefore, they can still panic at run-time. On the other hand, inline blocks will always be evaluated at compile-time.

flip1995: Ah right, now I also understand what this PR is about. :sweat_smile:

flip1995:
Added `changelog:` line to PR body.

@bors r+

Thanks!

PR #8593, opened 2022-03-27
Add a lint for static items with large type alignment

Jaic1:
The goal of this PR is to warn the usage of static items which have type alignment larger than page size (typically 4KB).

Due to some known unsound issues, the type alignment may not be fulfilled.
For more information, see rust-lang/rust#70022 and rust-lang/rust#70143.

changelog: add lint [`static_items_large_align`]


clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,92 @@
...
+    "static items with large type alignment, which may trigger unsound problems"
+}
+impl_lint_pass!(StaticItemsLargeAlign => [STATIC_ITEMS_LARGE_ALIGN]);
+
+impl LateLintPass<'_> for StaticItemsLargeAlign {
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if_chain! {
+            if let ItemKind::Static(hir_ty, _, _) = item.kind;
+            let ty = hir_ty_to_ty(cx.tcx, hir_ty);
+            if let Some(adt_ref) = self.check_ty_alignment(cx.tcx, ty);
```
oli-obk, line 49: I did not think about references at all... that's a really good catch. I'm wondering if we'll miss anything with such an approach. I guess we could use the const validator to look at the actual allocations, but not sure we'll have other problems due to that.

Either way, your function looks good and we can add more types to recurse on later.

One thing your PR made me realize is that promoteds suffer from the same issue, but those can even be generic... not sure how to best handle them. Maybe with a mir visitor looking at the types of all promoteds. Doesn't have to be in this PR tho

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+impl StaticItemsLargeAlign {
+    /// It checks a type with the following steps:
+    /// 1. picks out this type if its kind is among adt, array, tuple or ref to them;
+    ///    otherwise return None
+    /// 2. check if its (or its inner fields') alignment are larger than page size
+    /// 3. return one of them;
+    ///    otherwise return None
+    fn check_ty_alignment<'tcx>(&self, tcx: TyCtxt<'tcx>, ty: Ty<'tcx>) -> Option<AdtDef<'tcx>> {
```
oli-obk, line 72: I wonder if we can do a bit better on the spans and also collect the spans of all fields of intermediate structs. Not sure how to get a struct field's span though.

Jaic1: How about just report the intermediate types in mir but without the spans (from hir)? The difficult part in collecting the spans is dealing with the generics in hir.

A simple example:
```rust
#[repr(align(0x100000))]
struct Aligned(u8);

struct AlignedWrapper {
    f: u8,
    g: Aligned,
}

struct AlignedGeneric<T>(T);

static XG: AlignedGeneric<AlignedWrapper> = ...;
```
Then the intermediate types in mir would be:
```rust
AlignedGeneric<AlignedWrapper>
AlignedWrapper
Aligned
```
But in hir we can only get:
```rust
AlignedGeneric<T>(T)
// T needs to be resolved to AlignedWrapper
// and it is not straightforward to do it in hir
...
```

As we get more involved in connecting the mir and hir (the solution I have been trying), it is likely the lint/checking would become unmanageable and be unsound (i.e., having false negative).

Here is a more involved example (as a reference):
```rust
#[repr(align(0x100000))]
struct AlignedA(u8);

#[repr(align(0x100000))]
struct AlignedB(u8);

struct FnPtr<T>(fn() -> Box<T>);

struct AG<T>(T);

type AGAlias<T> = AG<T>;

struct AGWithArgs<A, B> {
    not_aligned: FnPtr<A>,
    aligned: AGAlias<B>,
}

static XG: AGWithArgs<AlignedA, AlignedB> = AGWithArgs {
    not_aligned: FnPtr(box_aligned),
    aligned: AG(AlignedB(0)),
};

fn box_aligned() -> Box<AlignedA> {
    Box::new(AlignedA(0))
}
```
The report in mir would be straightforward (since it is more easy to mono mir types):
```rust
AGWithArgs<AlignedA, AlignedB>
AG(AlignedB(0))
AlignedB
```
But it is difficult to report the relevant spans (from hir) like this:
```rust
struct AGWithArgs<A, B> {
    not_aligned: FnPtr<A>,
    aligned: AGAlias<B>,
    ^^^^^^^^^^^^^^^^^^^
}
type AGAlias<T> = AG<T>;
                  ^^^^^
struct AG<T>(T);
            ^^^
AlignedB
```

Or maybe there is a good way to work with (e.g., some subst mechanisms) generics in hir I am missing? And now I also start to wonder how rustc deal with generics when reporting (type) errors to user.

clippy_lints/src/static_items_large_align.rs
Jaic1: > I guess we could use the const validator to look at the actual allocations

Could you provide more context on const validator and allocations?

> One thing your PR made me realize is that promoteds suffer from the same issue, but those can even be generic...

Your mention of generic reminds me to think more about it when collecting the spans... As for  promoteds, I guess I will take a peek at const eval.

clippy_lints/src/static_items_large_align.rs
oli-obk: Yea, that seems reasonable. You can use `tcx.def_span` to get the span of the structs.

clippy_lints/src/static_items_large_align.rs
oli-obk: The actual memory as we will pass it to llvm can be fetched with https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.eval_static_initializer

This gives you access to an Allocation which has an `align` field. It also gives you a list of relocations, which are ids of other allocations referenced by this static (essentially pointers to other memory). You can then use https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.get_global_alloc to fetch the allocations themselves and recurse. This will not give us any information about the spans or similar, but it will make sure we don't miss anything.

We could as a next step employ https://doc.rust-lang.org/nightly/nightly-rustc/rustc_const_eval/interpret/visitor/index.html but that is a bit more involved

oli-obk:
All my comments are for follow up improvements. IMO we can merge this as is

Jarcho:
Ping @Jaic1. Are you willing to continue working on this? It doesn't seem like there are any major blockers for merging.

Jaic1:
@Jarcho Sure, I can resolve the conflicts at the moment. For the potential improvements mentioned above, maybe I can pick it up some day.

oli-obk:
Please rebase instead of merge. When you rebase, the merge commit gets removed automatically without you needing to do anything

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+    ///     let x = Aligned(0);                 // Good
+    ///     println!("{:#x}", &x as *const _ as usize);
+    ///     println!("{:#x}", &X as *const _ as usize);
+    ///     let b = Box::new(Aligned(0));       // Good
+    ///     println!("{:#x}", Box::into_raw(b) as usize);
+    /// }
+    /// ```
+    #[clippy::version = "1.61.0"]
+    pub STATIC_ITEMS_LARGE_ALIGN,
+    pedantic,
```
flip1995, line 41: I think this would be a good candidate for `correctness` or at least the `suspicious` group. Or do you have concerns about drastic false positives this lint might trigger?

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+    ///
+    /// fn main() {
+    ///     let x = Aligned(0);                 // Good
+    ///     println!("{:#x}", &x as *const _ as usize);
+    ///     println!("{:#x}", &X as *const _ as usize);
+    ///     let b = Box::new(Aligned(0));       // Good
+    ///     println!("{:#x}", Box::into_raw(b) as usize);
+    /// }
+    /// ```
+    #[clippy::version = "1.61.0"]
```
flip1995, line 39:
```suggestion
    #[clippy::version = "1.66.0"]
```

This will most likely be merged in the 1.66 cycle.

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+    pub STATIC_ITEMS_LARGE_ALIGN,
```
flip1995, line 40: The lint name sounds a bit weird to me. Maybe `static_items_with_large_alignment` but then again, that's a bit wordy... Maybe the current lint name is a good compromise.

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+                }
+                span_notes.push(self.report_la_ty(cx, &la_ty));
+
+                span_lint_and_then(
+                    cx,
+                    STATIC_ITEMS_LARGE_ALIGN,
+                    item.span,
+                    "this static item (itself or its subfield) has a type alignment,\n\
+                    which is larger than page size and may not be fulfilled,\n\
+                    for more information, see <https://github.com/rust-lang/rust/issues/70022>.",
```
flip1995, line 76:
```suggestion
                    "static item with large type alignment",
```

The lint message is not meant to explain the whole world :P Clippy will add a link to its documentation to the first lint emission anyway, so adding links is not necessary.

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+        if let ty::Adt(adt_def, substs_ref) = ty.kind() {
+            Some((
+                cx.tcx.def_span(adt_def.did()),
+                if substs_ref.is_empty() {
+                    format!("{} contains an inner type with large alignment", ty.sort_string(cx.tcx))
+                } else {
+                    // TODO - can we use :#?
+                    format!(
+                        "{} with substitutions {:#?},\n\
+                        contains an inner type with large alignment",
```
flip1995, line 155:
```suggestion
                        "{} with substitutions {:#?}, contains an inner type with large alignment",
```

no newline.

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    fn report_im_ty(cx: &LateContext<'_>, im_ty: &IntermediateTy<'_>) -> Option<(Span, String)> {
+        let ty = im_ty.ty;
+        if let ty::Adt(adt_def, substs_ref) = ty.kind() {
+            Some((
+                cx.tcx.def_span(adt_def.did()),
+                if substs_ref.is_empty() {
+                    format!("{} contains an inner type with large alignment", ty.sort_string(cx.tcx))
+                } else {
+                    // TODO - can we use :#?
```
flip1995, line 152: Can we? Do you have a test for this? This TODO comment should be resolved before merging.

clippy_lints/src/static_items_large_align.rs
```diff
@@ -0,0 +1,186 @@
...
+    /// 1. Check if the type is already visited (for a static item),
+    ///     if not, continue;
+    ///     otherwise, return `None` early.
+    /// 2. Push the type in the checked types.
+    /// 3. Pick out this type if its kind is among adt, tuple, array, ref or raw ptr to them;
+    ///     otherwise return `None`.
+    /// 4. Check if its (or its inner fields') alignment are larger than page size.
+    /// 5. Return one of them;
+    ///     otherwise pop the current checked type and return `None`.
+    fn check_ty_alignment<'tcx>(
```
flip1995, line 99: Doesn't rustc have a query that just gives you the type alignment? :thinking: Implementing type alignment calculation in Clippy feels a bit wrong to me.

flip1995 (changes requested):
There are a few changes left to do before we can merge this.

blyxyas:
@Jaic1 I'm closing this for inactivity. It seems like you're not active in the OSS ecosystem anymore, so, have luck in whatever else you're doing! :heart:.

If you'd like to continue with this PR, don't hesitate to comment and I'll reopen it! Thanks for trying ~(=^‥^)/.

PR #8596, opened 2022-03-28
Fix unnecessary_cast suggestion for type aliasses

### LINKED ISSUE #6923: unnecessary_cast suggestion for type aliasses
Lint name: unnecessary_cast

I tried this code:

```rust
#[allow(non_camel_case_types)]
pub type size_t = usize;

pub struct Foo {
    pub len: size_t,
} 

impl Foo {
    #[allow(dead_code)]
    fn new() -> Foo {
        Foo {
            len: 0 as size_t,
        }
    }
}
```

No lint triggered, or, ideally, a suggestion to omit the cast completely and write just `len: 0,`

Instead, this happened: 
Clippy suggested using `0_usize` instead of `0 as size_t`. 
The suggestion is not good, because then changing `size_t` to e.g. `u64` causes the compiler error.

### Meta

- clippy 0.1.52 (2021-03-16 f5d8117)
- rustc the Stable version: 1.50.0

Jaic1:
Fix #6923. The [`unnecessary_cast`] lint now will skip casting to non-primitive type.

changelog: fix lint [`unnecessary_cast `]

### DIFF
--- a/clippy_lints/src/casts/unnecessary_cast.rs
+++ b/clippy_lints/src/casts/unnecessary_cast.rs
@@ -4,7 +4,8 @@ use clippy_utils::source::snippet_opt;
 use if_chain::if_chain;
 use rustc_ast::{LitFloatType, LitIntType, LitKind};
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind, Lit, UnOp};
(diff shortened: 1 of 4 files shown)


flip1995:
@bors r+

Thanks!

Issue #8527, opened 2022-03-12
Clippy crashes in iter_overeager_cloned::check


remoun:
I ran into an unknown issue that crashes Clippy. I can reliably reproduce this with `nightly-2022-02-22` and `nightly-2022-03-10`

Output:
```
error: internal compiler error: compiler/rustc_middle/src/ty/normalize_erasing_regions.rs:179:90: Failed to normalize <std::option::Option<&std::option::Option<std::time::Instant>> as std::iter::Iterator>::Item, maybe try to call `try_normalize_erasing_regions` instead

thread 'rustc' panicked at 'Box<dyn Any>', compiler/rustc_errors/src/lib.rs:1236:9
stack backtrace:
   0: std::panicking::begin_panic::<rustc_errors::ExplicitBug>
   1: std::panic::panic_any::<rustc_errors::ExplicitBug>
   2: <rustc_errors::HandlerInner>::bug
   3: <rustc_errors::Handler>::bug
   4: rustc_middle::ty::context::tls::with_opt::<rustc_middle::util::bug::opt_span_bug_fmt<rustc_span::span_encoding::Span>::{closure#0}, ()>
   5: rustc_middle::util::bug::opt_span_bug_fmt::<rustc_span::span_encoding::Span>
   6: rustc_middle::util::bug::bug_fmt
   7: <rustc_middle::ty::normalize_erasing_regions::NormalizeAfterErasingRegionsFolder as rustc_middle::ty::fold::TypeFolder>::fold_ty
   8: clippy_utils::ty::get_associated_type
   9: clippy_lints::methods::iter_overeager_cloned::check
  10: <clippy_lints::methods::Methods as rustc_lint::passes::LateLintPass>::check_expr
  11: <rustc_lint::late::LateLintPassObjects as rustc_lint::passes::LateLintPass>::check_expr
  12: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  13: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_block
  14: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
[...]

PR #8602, opened 2022-03-29
Fix ICE for `iter_overeager_cloned`

giraffate:
Fix https://github.com/rust-lang/rust-clippy/issues/8527

changelog: Fix ICE for [`iter_overeager_cloned`]


llogiq:
Thank you! @bors r+

PR #8607, opened 2022-03-29
Allow running `cargo dev lint` on a package directory

Alexendoo:
Allows you run the local clippy in a specified directory, e.g. allowing

```sh
# Lint a ui-cargo test
cargo dev lint tests/ui-cargo/wildcard_dependencies/fail

# Lint some other project
cargo dev lint ~/my-project
```

The `target` directory is set to a tempdir which isn't ideal for medium/large projects as it would be compiled from scratch. This is to avoid cached clippy messages where you `cargo dev lint dir`, change something in clippy, and run `cargo dev lint dir` again

changelog: Dev: `cargo dev lint` can now be run on a package directory

### DIFF
--- a/clippy_dev/Cargo.toml
+++ b/clippy_dev/Cargo.toml
@@ -11,6 +11,7 @@ itertools = "0.10.1"
 opener = "0.5"
 regex = "1.5"
 shell-escape = "0.1"
+tempfile = "3.3"
 walkdir = "2.3"
 cargo_metadata = "0.14"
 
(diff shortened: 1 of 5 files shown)


giraffate (commented):
It looks good, but I'm unfamiliar with this code, so can anyone do another review for this?
@rust-lang/clippy

flip1995:
LGTM. The `tempfile` doesn't seem to increase the time this takes to build on CI, so adding this should be fine and don't lead to increased build times. They were just reduced in https://github.com/rust-lang/rust-clippy/pull/8605 and I don't want to increase them again.

@bors r=flip1995,giraffate

PR #8611, opened 2022-03-30
Handle relative paths in module_files lints

### LINKED ISSUE #8123: mod_module_files does nothing
### Summary

TL;DR: clippy::mod_module_files seems to do absolutely nothing, whether as a crate attribute or as a command line arg.

I'm using clippy via rustup

This occurs both on my laptop (WSL, Ubuntu 18.04) as well as a remote server (Ubuntu 20.04), but most of this was done on my laptop, I just checked the command line option on the remote to rule out this being WSL related.

`cargo clippy -V`: clippy 0.1.57 (f1edd04 2021-11-29)

EDIT:
In case this is useful:  I cloned the clippy repo, and ran the tests (which passed) but if I manually `cd` into the test directories and run `cargo clippy` (or nightly) the results did not match up with what the relevant `src/main.stderr` contained.

i.e. 
```
~/rust-clippy$ TESTNAME="module_style" cargo uitest
    Finished test [unoptimized + debuginfo] target(s) in 0.27s
     Running tests/compile-test.rs (target/debug/deps/compile_test-2c69db81b6b806f2)

running 1 test
test [ui] ui-cargo/module_style/fail_no_mod/src/main.rs ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

etc...
```
vs
```
~/rust-clippy/tests/ui-cargo/module_style/fail_mod$ cargo clippy
    Checking fail v0.1.0 (/home/username/rust-clippy/tests/ui-cargo/module_style/fail_mod)
    Finished dev [unoptimized + debuginfo] target(s) in 0.15s
```
^ would expect an error based on the contents of `tests/ui-cargo/module_style/fail_mod/src/main.stderr` (unless I'm misunderstanding how these tests work...)

### Reproducer

I used `cargo new my_project` to start a new project.
I created a file `my_project/src/my_module/mod.rs` with the (presumably unimportant) contents:
```rust
pub fn my_fn() {
    println!("Text");
}
```
`my_project/src/main.rs`:
```rust
// seems to do nothing
#![forbid(clippy::mod_module_files)]
// seems to do nothing
#![forbid(clippy::self_named_module_files)]
// works fine
#![forbid(clippy::expect_used)]

mod my_module;

fn main() {
    my_module::my_fn();
    let option = Some(5);
    let seven = option.expect("see previous line") + 2;
    println!("seven is {}", seven);
}
```
running `cargo clippy ` correctly gives an error for the use of `.expect` but no module errors

I expected to see an error due to a `mod.rs` file (I included self_named as well in case I somehow mixed the two up. - in a larger project that has both mod.rs and self named modules, neither lint seems to do anything.)

I thought this could be related to #6610 so I added a `.cargo/config.toml`:
```
[target.'cfg(feature = "cargo-clippy")']
rustflags = [
  "-Dclippy::mod_module_files",
  "-Dclippy::expect_used",
]
```
( I tried it with just expect, just module_files, both, neither)
and then I tried running
`cargo clippy -- -W clippy::mod_module_files` to no avail

all three of these methods worked with a variety of other lints I tried, and all three give an error if the named lint does not exist (or if the name is misspelled), I tried them each individually and in concert.  I think this rules out e.g. accidentally using the wrong clippy version or a typo on my part (as there would have been an error that the lint does not exist, which did happen when I first tested on another machine, until I ran `rustup update`).

I also initially ran `mod_module_files` (and self named) on a larger project (the above was to redo it on a minimal example to make sure it wasn't some weird configuration issue or something on the larger project).  I cannot get `mod_module_files` to give a warning/error on anything at all.

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Alexendoo:
The problem being that when clippy is run in the project's directory `lp` would be a relative path, this wasn't caught by the tests as there `lp` is an absolute path. Being a relative path it did not start with `trim_src_path` and so was ignored

Also allowed the removal of some `.to_os_string`/`.to_owned`s

changelog: Fixes [`self_named_module_files`] and [`mod_module_files`] not linting

Fixes #8123, cc @DevinR528

### DIFF
--- a/clippy_lints/src/module_style.rs
+++ b/clippy_lints/src/module_style.rs
@@ -1,13 +1,10 @@
-use std::{
-    ffi::OsString,
-    path::{Component, Path},
-};
-
 use rustc_ast::ast;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_lint::{EarlyContext, EarlyLintPass, Level, LintContext};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::{FileName, RealFileName, SourceFile, Span, SyntaxContext};
+use std::ffi::OsStr;
+use std::path::{Component, Path};
 
 declare_clippy_lint! {
     /// ### What it does
@@ -82,11 +79,7 @@ impl EarlyLintPass for ModStyle {
 
         let files = cx.sess().source_map().files();
 
-        let trim_to_src = if let RealFileName::LocalPath(p) = &cx.sess().opts.working_dir {
-            p.to_string_lossy()
-        } else {
-            return;
-        };
+        let RealFileName::LocalPath(trim_to_src) = &cx.sess().opts.working_dir else { return };
 
         // `folder_segments` is all unique folder path segments `path/to/foo.rs` gives
         // `[path, to]` but not foo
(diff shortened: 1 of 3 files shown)


DevinR528 (commented):
Not a Clippy contributor but I implemented the original lint, this looks good to me!

llogiq:
@DevinR528 as you wrote the original lint, you *are* a contributor. 😃

@Alexendoo this is a solid improvement, and the code is also looking good. Thank you for this!

@bors r+

Issue #8379, opened 2022-01-31
`transmute_int_to_char` incorrectly reported in `const` contexts


SabrinaJewson:
### Summary

Neither [`from_u32`](https://doc.rust-lang.org/stable/std/char/fn.from_u32.html) nor [`from_u32_unchecked`](https://doc.rust-lang.org/stable/std/char/fn.from_u32_unchecked.html) are stably `const`, so this in a `const` context Clippy should not warn on calling `std::mem::transmute::<u32, char>`.

### Lint Name

transmute_int_to_char

### Reproducer

I tried this code:

```rust
const fn foo() -> char {
    unsafe { std::mem::transmute(3486) }
}
```

I saw this happen:

```
warning: transmute from a `i32` to a `char`
 --> src/lib.rs:2:14
  |
2 |     unsafe { std::mem::transmute(3486) }
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `std::char::from_u32(3486 as u32).unwrap()`
  |
  = note: `#[warn(clippy::transmute_int_to_char)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_int_to_char
```

I expected to see this happen: No error.

### Version

```text
rustc 1.58.0 (02072b482 2022-01-11)
binary: rustc
commit-hash: 02072b482a8b5357f7fb5e5637444ae30e423c40
commit-date: 2022-01-11
host: x86_64-unknown-linux-gnu
release: 1.58.0
LLVM version: 13.0.0
```

### Additional Labels

@rustbot label +I-suggestion-causes-error

rossmacarthur:
Got a similar error when transmuting bytes to a `&str`

```
const VAR: &str = unsafe { core::mem::transmute::<&[u8], &str>(&[]) };
```

```
warning: transmute from a `&[u8]` to a `&str`
 --> src/main.rs:2:5
  |
2 |     core::mem::transmute::<&[u8], &str>(&[])
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `std::str::from_utf8(&[]).unwrap()`
  |
  = note: `#[warn(clippy::transmute_bytes_to_str)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#transmute_bytes_to_str
 ```
 
 It should suggest the const `std::str::from_utf8_unchecked`

PR #8612, opened 2022-03-30
Suggest from_utf8_unchecked in const contexts

SabrinaJewson:
Unfortunately I couldn't figure out how to check whether a given expression is in an `unsafe` context or not, so I just unconditionally emit the wrapping `unsafe {}` block in the suggestion. If there is an easy way to get it to work better then I would love to hear it.

changelog: Suggest `from_utf8_unchecked` instead of `from_utf8` in const contexts for ``[`transmute_bytes_to_str`]``

refs: #8379

### DIFF
--- a/clippy_lints/src/transmute/transmute_ref_to_ref.rs
+++ b/clippy_lints/src/transmute/transmute_ref_to_ref.rs
@@ -41,7 +41,7 @@ pub(super) fn check<'tcx>(
                     &format!("transmute from a `{}` to a `{}`", from_ty, to_ty),
                     "consider using",
                     if const_context {
-                        format!("unsafe {{ std::str::from_utf8_unchecked{postfix}({snippet}) }}")
+                        format!("std::str::from_utf8_unchecked{postfix}({snippet})")
                     } else {
                         format!("std::str::from_utf8{postfix}({snippet}).unwrap()")
                     },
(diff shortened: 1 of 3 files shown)


Jarcho:
`transmute` already requires an unsafe block so you don't need to suggest adding one.

SabrinaJewson:
Oh haha, I don't know how I didn't realize that 😅 . Thanks for pointing it out.

clippy_lints/src/transmute/transmute_ref_to_ref.rs
```diff
@@ -32,17 +32,19 @@ pub(super) fn check<'tcx>(
...
-                        "std::str::from_utf8{}({}).unwrap()",
-                        postfix,
-                        snippet(cx, arg.span, ".."),
-                    ),
+                    if const_context {
+                        format!("std::str::from_utf8_unchecked{postfix}({snippet})")
+                    } else {
+                        format!("std::str::from_utf8{postfix}({snippet}).unwrap()")
+                    },
                     Applicability::Unspecified,
```
flip1995, line 48: I think this can be changed to 
```suggestion
                    Applicability::MaybeIncorrect,
```

flip1995 (commented):
Only one comment, LGTM otherwise.

flip1995:
@bors r+

Thanks!

PR #8614, opened 2022-04-01
assertions_on_constants: ignore indirect `cfg!`

### LINKED ISSUE #7597: assertions_on_constants incorrectly lints indirect usage of `cfg!`
Lint name: assertions_on_constants

I tried this code:

```rust
const CFG_FLAG: bool = cfg!(config_thing);

fn main() {
    assert!(!CFG_FLAG); // This gives a warning
    assert!(!cfg!(config_thing)); // But direct usage of cfg! is suppressed
}
```

Indirect usage of `cfg!()` flags should not give a warning, just like direct usage.

### Meta

- `cargo clippy -V`: `clippy 0.1.56 (0fa3190 2021-08-12)`
- `rustc -Vv`: 

```
rustc 1.56.0-nightly (0fa319039 2021-08-12)
binary: rustc
commit-hash: 0fa3190394475a84360b34e074e719d519bc40f1
commit-date: 2021-08-12
release: 1.56.0-nightly
LLVM version: 12.0.1
```

This false positive also happens on the latest stable as well...

Thanks for your time!

pitaj:
Fixes #7597

changelog: [`assertions_on_constants`] ignore constants indirectly based on `cfg!`

### DIFF
--- a/clippy_utils/src/consts.rs
+++ b/clippy_utils/src/consts.rs
@@ -5,7 +5,7 @@ use if_chain::if_chain;
 use rustc_ast::ast::{self, LitFloatType, LitKind};
 use rustc_data_structures::sync::Lrc;
 use rustc_hir::def::{DefKind, Res};
-use rustc_hir::{BinOp, BinOpKind, Block, Expr, ExprKind, HirId, QPath, UnOp};
+use rustc_hir::{BinOp, BinOpKind, Block, Expr, ExprKind, HirId, Item, ItemKind, Node, QPath, UnOp};
 use rustc_lint::LateContext;
 use rustc_middle::mir::interpret::Scalar;
 use rustc_middle::ty::subst::{Subst, SubstsRef};
@@ -400,6 +400,22 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
         let res = self.typeck_results.qpath_res(qpath, id);
         match res {
             Res::Def(DefKind::Const | DefKind::AssocConst, def_id) => {
+                // Check if this constant is based on `cfg!(..)`,
+                // which is NOT constant for our purposes.
+                if let Some(node) = self.lcx.tcx.hir().get_if_local(def_id) &&
+                let Node::Item(&Item {
+                    kind: ItemKind::Const(_, body_id),
+                    ..
+                }) = node &&
+                let Node::Expr(&Expr {
+                    kind: ExprKind::Lit(_),
+                    span,
+                    ..
+                }) = self.lcx.tcx.hir().get(body_id.hir_id) &&
+                is_direct_expn_of(span, "cfg").is_some() {
+                    return None;
+                }
+
                 let substs = self.typeck_results.node_substs(id);
                 let substs = if self.substs.is_empty() {
                     substs
--- a/tests/ui/assertions_on_constants.rs
+++ b/tests/ui/assertions_on_constants.rs
@@ -1,4 +1,4 @@
-#![allow(non_fmt_panics)]
+#![allow(non_fmt_panics, clippy::needless_bool)]
 
 macro_rules! assert_const {
     ($len:expr) => {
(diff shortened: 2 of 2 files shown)


clippy_utils/src/consts.rs
```diff
@@ -400,6 +400,27 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
         let res = self.typeck_results.qpath_res(qpath, id);
         match res {
             Res::Def(DefKind::Const | DefKind::AssocConst, def_id) => {
+                // Check if this constant is based on `cfg!(..)`,
+                // which is NOT constant for our purposes.
+                if let Some(mut node) = self.lcx.tcx.hir().get_if_local(def_id) {
+                    loop {
+                        match node {
+                            Node::Item(&Item {
+                                kind: ItemKind::Const(_, body_id),
```
flip1995, line 409: How would code look like that has nested `const` items? I don't see any tests for this. I have a feeling that walking down this chain isn't necessary.

tests/ui/assertions_on_constants.rs
```diff
@@ -28,6 +28,19 @@ fn main() {
...
     assert!(cfg!(feature = "hey") || cfg!(not(feature = "asdf")));
+
+    let flag: bool = cfg!(not(feature = "asdf"));
+    assert!(flag);
+
+    const CFG_FLAG: &bool = &cfg!(feature = "hey");
+    assert!(!CFG_FLAG);
+
+    const INDIRECT_FLAG: bool = !foo::CFG_FLAG;
+    assert!(CFG_FLAG);
```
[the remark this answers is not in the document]
pitaj: Good catch

clippy_utils/src/consts.rs
```diff
@@ -400,6 +400,85 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
         let res = self.typeck_results.qpath_res(qpath, id);
         match res {
             Res::Def(DefKind::Const | DefKind::AssocConst, def_id) => {
+                // Check if this constant is based on `cfg!(..)`,
+                // which is NOT constant for our purposes.
+                if let Some(node) = self.lcx.tcx.hir().get_if_local(def_id) {
+                    fn node_contains_cfg_expn(tcx: TyCtxt<'_>, n: Node<'_>) -> bool {
```
pitaj, line 406: Not sure if this should live somewhere else. It could maybe be useful to others as as more generic `clippy_utils::contains_expn_of(e: &Expr, name: &str) -> bool`. Or it could just live in the `assertions_on_constants` lint.

clippy_utils/src/consts.rs
```diff
@@ -567,6 +575,69 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
     }
 }
 
+fn node_contains_cfg_expn(tcx: TyCtxt<'_>, n: Node<'_>) -> bool {
```
pitaj, line 578: Not sure where this should live. I can see a few options:
- here
- more generic version `clippy_utils::contains_expn_of(e: &Expr, name: &str)`
- in the `assertions_on_constants` lint

I can see it being useful in other lints like `needless_bool`.

flip1995: I agree with @giraffate The current impl would also no longer lint on things like `const _: bool = [cfg!("bla"); 10].len() > 5;`, just because there is a `cfg!` somewhere in the expression. But the `cfg!` in this case has absolutely no influence on the value of the resulting `bool`. 

I would also just check for direct `cfg!` and possibly extend this checks if other issues are opened about more complex cases that would make sense to allow.

pitaj: I figure a false negative is better than a false positive, but I understand not wanting to add so much complexity for little benefit.

giraffate (approved):
It looks good. Can you squash the commits?

pitaj:
@giraffate squashed

giraffate:
@bors r+

Thanks!

PR #8616, opened 2022-04-01
single_element_loop: handle arrays for Edition2021

pitaj:
changelog: [`single_element_loop`] handle arrays in Edition 2021, handle `.iter_mut()` and `.into_iter()`, and wrap in parens if necessary


llogiq:
Thank you!

@bors r+

PR #8617, opened 2022-04-01
`needless_late_init`: ignore `if let`, `let mut` and significant drops

### LINKED ISSUE #8613: Suppress needless_late_init on if-let and long if/match expr
### Summary

According to [the lint docs](https://rust-lang.github.io/rust-clippy/master/index.html#needless_late_init), the sole justification for this lint is to reduce repetitive code. Yes, the examples in the docs look better as suggested, but others don't.

Slightly wordy but clean code:
```rust
        let attr_widget;
        if let Some(i) = index {
            let attr = attrs.remove(i);
            let (_, tokens) = parse_attr_group(attr.tokens)?;
            attr_widget = syn::parse2(tokens)?;
        } else {
            attr_widget = Default::default();
        }
```
Shorter code, but first line contains two `=` tokens (ugly):
```rust
        let attr_widget = if let Some(i) = index {
            let attr = attrs.remove(i);
            let (_, tokens) = parse_attr_group(attr.tokens)?;
            syn::parse2(tokens)?
        } else {
            Default::default()
        };
```
It's easy to imagine worse examples:
```rust
let foo = if let Some(bar) = thing.iter().find(|t| t.is_good()) {
    ...
} else {
    ...
};
```

Suggestion: suppress lint when first line of `if` / `match` is long or includes an `if let` binding.

### Lint Name

needless_late_init

### Reproducer

_No response_

### Version

_No response_

### Additional Labels

_No response_

Alexendoo:
No longer lints `if let`, personal taste on this one is pretty split, so it probably shouldn't be warning by default. Fixes #8613

```rust
let x = if let Some(n) = y {
    n
} else {
    1
}
```

No longer lints `let mut`, things like the following are not uncommon and look fine as they are

https://github.com/shepmaster/twox-hash/blob/b169c16d86eb8ea4a296b0acb9d00ca7e3c3005f/src/sixty_four.rs#L88-L93

Avoids changing the drop order in an observable way, where the type of `x` has a drop with side effects and something between `x` and the first use also does, e.g.

https://github.com/denoland/rusty_v8/blob/48cc6cb791cac57d22fab1a2feaa92da8ddc9a68/tests/test_api.rs#L159-L167

The implementation of `type_needs_ordered_drop_inner` was changed a bit, it now uses `Ty::has_significant_drop` and reordered the ifs to check diagnostic name before checking the implicit drop impl

changelog: [`needless_late_init`]: No longer lints `if let` statements, `let mut` bindings and no longer significantly changes drop order

### DIFF
--- a/clippy_lints/src/needless_late_init.rs
+++ b/clippy_lints/src/needless_late_init.rs
@@ -1,10 +1,14 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::path_to_local;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::visitors::{expr_visitor, is_local_used};
+use clippy_utils::ty::needs_ordered_drop;
+use clippy_utils::visitors::{expr_visitor, expr_visitor_no_bodies, is_local_used};
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::Visitor;
-use rustc_hir::{Block, Expr, ExprKind, HirId, Local, LocalSource, MatchSource, Node, Pat, PatKind, Stmt, StmtKind};
+use rustc_hir::{
+    BindingAnnotation, Block, Expr, ExprKind, HirId, Local, LocalSource, MatchSource, Node, Pat, PatKind, Stmt,
+    StmtKind,
+};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::Span;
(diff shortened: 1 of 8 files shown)


giraffate:
Sorry for the late reviewing. I should be able to start the review by next week.

clippy_lints/src/needless_late_init.rs
```diff
@@ -73,6 +77,31 @@ fn contains_assign_expr<'tcx>(cx: &LateContext<'tcx>, stmt: &'tcx Stmt<'tcx>) ->
     seen
 }
 
+fn contains_let(cond: &Expr<'_>) -> bool {
```
giraffate, line 80: Is it possible to replace this with `higher::IfLet::hir`?

tests/ui/needless_late_init.rs
```diff
@@ -53,7 +56,29 @@ fn main() {
...
+    let y = String::new();
+    let y = vec![3.0];
+    let y = HashMap::<usize, usize>::new();
+    let y = BTreeMap::<usize, usize>::new();
+    let y = HashSet::<usize>::new();
+    let y = BTreeSet::<usize>::new();
+    let y = Box::new(4);
+    x = SignificantDrop;
+
+    println!();
```
giraffate, line 81: nits
```suggestion
```

tests/ui/needless_late_init.rs
```diff
@@ -177,5 +202,34 @@ fn does_not_lint() {
...
+    x = 2;
+    x = 3;
+
+    // should not move the declaration if `x` has a significant drop, and there
+    // is another binding with a significant drop between it and the first usage
+    let x;
+    let y = SignificantDrop;
+    x = SignificantDrop;
+
+    println!();
```
giraffate, line 234: nits
```suggestion
```

tests/ui/needless_late_init_fixable.rs
```diff
@@ -14,12 +14,8 @@ fn main() {
...
-    e = 1;
-    e = 2;
+    let e;
+    e = format!("{}", d);
 
-    let h;
-    h = format!("{}", e);
-
-    println!("{}", a);
+    println!();
```
giraffate, line 20: nits
```suggestion
```

clippy_lints/src/needless_late_init.rs
Alexendoo: It wouldn't be for let chains

giraffate:
@bors r+

Thanks!

PR #8620, opened 2022-04-02
Run fmt test before compile-test/dogfood

Alexendoo:
I seem to always forget to run `cargo dev fmt` before doing a test. This lets it fail fast rather than going through the much longer compile-test/dogfood tests first

changelog: none


flip1995:
I mean, why not. Renaming the test file seems like a hack, but after a quick google search it seems this is the only way how to do this.

@bors r+

PR #8621, opened 2022-04-02
Rename `eval_order_dependence` to `mixed_read_write_expression`, move to nursery

Jarcho:
As per the [reference](https://doc.rust-lang.org/1.51.0/reference/expressions.html#evaluation-order-of-operands) evaluation order is now defined.

I'm pretty sure rust always compiled with this evaluation order anyways so there's no reason the put an msrv limit on the lint.

changelog: Rename `eval_order_dependence` to `mixed_read_write_expression`, move to nursery


xFrednet:
I'm not sure if we should deprecate the lint just because the evaluation order has been defined. Looking at the [example from the lint description](https://rust-lang.github.io/rust-clippy/master/index.html#eval_order_dependence), I would still expect this to be linted or later mark that as confusion in the code review.

```rs
// Example from the lint description
let a = {
    x = 1;
    1
} + x;
```

CC: @rust-lang/clippy

llogiq:
I also think deprecation is too far a step. It may be debatable if `suspicious` is still the right category or if it should be a `restriction` lint, though.

Jarcho:
Two reasons for the deprecation. 
1. It should be renamed and moved into a different group at the very least. The current name is no longer accurate.
2. It's current state is not good given the fact it's not catching any code which is _wrong_, only code of a questionable style. For catching style issues it lints on too many things, specifically using `syn` is broken.

If not deprecating it, I would say move it to the nursery with an issue open to deal with what should happen to it.

xFrednet:
Given the background, I agree, it should at least be moved and renamed. Having a lint for the style from the example sounds good, but we can add that afterwards. :+1:

Jarcho:
Any nominations for a new name? If not I'll just update the PR to move it into the nursery with a note to deal with it later.

flip1995:
`unclear_eval_order` / `suspicious_eval_order`?

I think `restriction` is a good category for this lint, even if not renamed right away. If someone really cares about eval order and enables this lint, some FPs shouldn't be too bad.

Jarcho:
Maybe `mixed_read_write_in_expression`. It's a little more direct as to what's linted. `restriction` sounds fine.

xFrednet:
I think this sounds like a plan. @Jarcho Could you rename the lint (I like `mixed_read_write_in_expression`) and move it to restriction? Then we can move forward with this :upside_down_face:

Jarcho:
Had to add the renamed lints to the link list in the changelog. Do we want to have that link to the page for the new name? Currently I have it using the page for the old name.

It could be either
```markdown
[`old_name`]: https://rust-lang.github.io/rust-clippy/master/index.html#old_name
```
or
```markdown
[`old_name`]: https://rust-lang.github.io/rust-clippy/master/index.html#new_name
```

flip1995:
> Had to add the renamed lints to the link list in the changelog

You can just remove the `[`, `]` around mentions of the old lint name.

Jarcho:
Deprecated lints are already in the link list. This keeps renamed lints consistent with those. The links aren't useful as the lints aren't in the docs. (They should be, but that's a different issue)

xFrednet:
Adding renamed lints as an alias with a note to our lint list is still on my mind, somewhere at the bottom of my todo list :sweat_smile:

flip1995:
Deprecated lints are in the docs if you explicitly filter for them with the lint-group filter: https://rust-lang.github.io/rust-clippy/master/index.html#assign_ops (I consider it a bug that it doesn't show up with this link, because the filter excludes it. I think we have an issue open for this or at least talked about it before? :thinking:)

Renamed lints can't be in the documentation, because the old lint isn't defined anywhere anymore (except for its name).

llogiq:
I was a bit discombobulated by the added links in the lint list, but I now understand. Sorry for taking so long to review.

r=me after rebase

PR #8624, opened 2022-04-02
New lint `is_digit_ascii_radix`

### LINKED ISSUE #6399: Lint: is_ascii_digit instead of is_digit
### What it does

Finds usages of `is_digit` that can be replaced with `is_ascii_digit` or `is_ascii_hexdigit`.

### Categories (optional)

- Kind: style

It is simpler since the radix does not need to be specified.

### Drawbacks

None.

### Example

```rust
fn bad(c: char) {
    c.is_digit(10);
    c.is_digit(16);
}
```

Could be written as:

```rust
fn good(c: char) {
    c.is_ascii_digit();
    c.is_ascii_hexdigit();
}
```

pitaj:
Closes #6399 

changelog: Added [`is_digit_ascii_radix`]: recommend `is_ascii_digit()` or `is_ascii_hexdigit()` in place of `is_digit(10)` and `is_digit(16)`

### DIFF
--- a/clippy_lints/src/methods/is_digit_ascii_radix.rs
+++ b/clippy_lints/src/methods/is_digit_ascii_radix.rs
@@ -1,0 +1,50 @@
+//! Lint for `c.is_digit(10)`
+
+use super::IS_DIGIT_ASCII_RADIX;
+use clippy_utils::{
+    consts::constant_full_int, consts::FullInt, diagnostics::span_lint_and_sugg, meets_msrv, msrvs,
+    source::snippet_with_applicability,
+};
+use rustc_errors::Applicability;
+use rustc_hir::Expr;
+use rustc_lint::LateContext;
+use rustc_semver::RustcVersion;
+
+pub(super) fn check<'tcx>(
+    cx: &LateContext<'tcx>,
+    expr: &'tcx Expr<'_>,
+    self_arg: &'tcx Expr<'_>,
+    radix: &'tcx Expr<'_>,
+    msrv: Option<&RustcVersion>,
+) {
+    if !meets_msrv(msrv, &msrvs::IS_ASCII_DIGIT) {
+        return;
+    }
+
+    if !cx.typeck_results().expr_ty_adjusted(self_arg).peel_refs().is_char() {
+        return;
+    }
+
+    if let Some(radix_val) = constant_full_int(cx, cx.typeck_results(), radix) {
+        let (num, replacement) = match radix_val {
+            FullInt::S(10) | FullInt::U(10) => (10, "is_ascii_digit"),
+            FullInt::S(16) | FullInt::U(16) => (16, "is_ascii_hexdigit"),
+            _ => return,
+        };
+        let mut applicability = Applicability::MachineApplicable;
+
+        span_lint_and_sugg(
+            cx,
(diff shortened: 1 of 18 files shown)


clippy_lints/src/methods/is_digit_ascii_radix.rs
```diff
@@ -0,0 +1,43 @@
...
+    cx: &LateContext<'tcx>,
+    expr: &'tcx Expr<'_>,
+    self_arg: &'tcx Expr<'_>,
+    radix: &'tcx Expr<'_>,
+) {
+    if !cx.typeck_results().expr_ty_adjusted(self_arg).peel_refs().is_char() {
+        return;
+    }
+
+    if let Some(radix) = constant_full_int(cx, cx.typeck_results(), radix) {
```
xFrednet, line 21: [`is_ascii_digit`](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_digit) and [is_ascii_hexdigit](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_hexdigit) has been stabilized in `1.24.0` while `is_digit(...)` has been stable since 1.0. Clippy has a configuration that is used to specify what the minimum rust version that the suggestion has to work for.

You can pass the current `msrv` version in as an argument and then check the version like this:

```rs
if meets_msrv(msrv, &msrvs::IS_ASCII_DIGIT) ...
```

See: [*Specifying the lint's minimum supported Rust version (MSRV)*](https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv)

They also have a minimum rust version for const evaluation, however, since `is_digit` isn't const we don't have to check for that :upside_down_face:

clippy_lints/src/methods/is_digit_ascii_radix.rs
```diff
@@ -0,0 +1,43 @@
...
+            FullInt::S(16) | FullInt::U(16) => "is_ascii_hexdigit",
+            _ => return,
+        };
+        let mut applicability = Applicability::MachineApplicable;
+
+        span_lint_and_sugg(
+            cx,
+            IS_DIGIT_ASCII_RADIX,
+            expr.span,
+            "use of `char::is_digit(..)` with literal radix of 10 or 16",
```
xFrednet, line 33: Since you mention the used radix, you could go one step further and only name the one been using. For that, you would store the found value and then use format, like so:

```rs
&format!("use of `char::is_digit(..)` with literal radix of {}", radiax_int),
```

or maybe:

```rs
&format!("use of `char::is_digit({})`", radiax_int),
```

This is optional, the current lint message already works well, and I'm also happy if you want to keep it this way :upside_down_face:

xFrednet (commented):
This already looks good to me :+1: I've added two comments, and then it should be ready to be merged. :upside_down_face:

pitaj:
Done.

@rustbot -S-waiting-on-author +S-waiting-on-review

pitaj:
Alright, moved the MSRV check and resolved the conflicts.

xFrednet:
Alright, the code looks good. Could you rebase your commit again on master? it looks like you accidentally pulled in a second commit from someone else 🙃

PR #8626, opened 2022-04-04
New lint `format_add_strings`

### LINKED ISSUE #6261: Warn on push_str(&format!())
### What it does

Emits a warning on code like ```rust
some_string.push_str(&format!(...));
// or
some_string += &format!(...);
```
and suggests rewriting it to ```rust
write!(&mut some_string, ...);
```

### Categories
- Kind: `Perf`

It avoids one extra heap allocation by writing directly to the existing String instead of creating a new temporary String.

### Drawbacks
The user has to `use std::io::Write` for the `write!` macro to work (which I don't get tbh, why can't `write!` just import the trait on its own...)

### Example
```rust
let mut string = String::new();
for foo in foo_list {
    string += &format!("{:?}", foo);
}
```

Could be written as:

```rust
use std::io::Write;

let mut string = String::new();
for foo in foo_list {
    write!(&mut string, "{:?}", foo);
}
```

pitaj:
Closes #6261 

changelog: Added [`format_add_string`]: recommend using `write!` instead of appending the result of  `format!`

### DIFF
--- a/clippy_lints/src/format_push_string.rs
+++ b/clippy_lints/src/format_push_string.rs
@@ -1,0 +1,77 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::{match_def_path, paths, peel_hir_expr_refs};
+use rustc_hir::{BinOpKind, Expr, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects cases where the result of a `format!` call is
+    /// appended to an existing `String`.
+    ///
+    /// ### Why is this bad?
+    /// Introduces an extra, avoidable heap allocation.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let mut s = String::new();
+    /// s += &format!("0x{:X}", 1024);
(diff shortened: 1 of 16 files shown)


clippy_lints/src/format_push_string.rs
```diff
@@ -0,0 +1,77 @@
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// let mut s = String::new();
+    /// s += &format!("0x{:X}", 1024);
+    /// s.push_str(&format!("0x{:X}", 1024));
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// use std::fmt::Write;
```
llogiq, line 25: Should we suggest `use std::fmt::Write as _;` here to avoid putting the item name in scope?

llogiq (approved):
This looks good to me. I have but one small suggestion and you will need to rebase the changes. Once that's done, I'll be happy to merge.

pitaj:
Changed to recommend `as _` for import and rebased.

llogiq:
Thanks! It will need another rebase, then it'll be ready to merge. Feel free to r+ it yourself

@bors delegate+

pitaj:
@llogiq why does it need another rebase? There are no conflicts.

xFrednet:
Maybe he missed your push? I'd say we let bors decide if this is ready or not. I don't want to steal your honer to use bors. You can merge this by commenting `@bors r+ llogiq`

llogiq:
In that case @bors r+

PR #8630, opened 2022-04-04
Add lints `drop_non_drop` and `forget_non_drop`

### LINKED ISSUE #1897: mem::forget on !Drop
We should warn if you call `mem::forget` on:

 - a generic type known to be `Copy`
 - a non generic type known to not be `Drop`

since it's a no op, and likely means you `mem::forgot` the wrong thing.

(I hit this issue today)

Jarcho:
fixes #1897

changelog: Add lints `drop_non_drop` and `forget_non_drop`


Manishearth (approved):

clippy_lints/src/drop_forget_ref.rs
```diff
@@ -103,6 +102,75 @@ declare_clippy_lint! {
     "calls to `std::mem::forget` with a value that implements Copy"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to `std::mem::drop` with a value that does not implement `Drop`.
+    ///
+    /// ### Why is this bad?
+    /// Calling `std::mem::drop` is no different than dropping such a type. A different value may
```
cherryblossom000, line 110: Maybe the wording of this could be improved. It looks like it was copied and pasted from the `forget` lint and I think this description is confusing – yes `std::mem::drop` is in fact no different than dropping it, what’s the problem?

PR #8632, opened 2022-04-04
Don't lint `cast_ptr_alignment` when used for unaligned reads and writes

### LINKED ISSUE #2881: cast_ptr_alignment is acceptable in a narrow set of contexts
Clippy may not be able to see when this is acceptable but if so there are several functions that can handle unaligned ptrs. If the unaligned pointer is only used as an input to those then it ideally wouldn't be flagged.

The obvious
std::ptr::read_unaligned
std::ptr::write_unaligned

Anything with memcpy or memmove semantics
std::ptr::copy
std::ptr::copy_nonoverlapping

I'm unsure about std::ptr::swap and std::ptr::swap_nonoverlapping

And in 
std::arch::x86_64
std::arch::x86
*storeu*
*loadu*
*lddqu*

There might be some others.

Jarcho:
fixes #2881

Ideally this would trace the usage of the value rather than only looking at the parent expression, but that would require dataflow analysis. e.g.
```rust
let x = ptr as *const u16;
c.read_unaligned(x);
```

Arch specific intrinsic functions need to be checked for ones which could take an unaligned pointer. This can be another PR.

changelog: Don't lint `cast_ptr_alignment` when used for unaligned reads and writes


llogiq:
Thank you! Great as always, a joy to review.

@bors r+

Issue #8627, opened 2022-04-04
Incorrect link in mod_module_files docs


Kyuuhachi:
### Description

The docs for the [mod_module_files](https://rust-lang.github.io/rust-clippy/master/index.html#mod_module_files) link contains a link to [mod.rs](//mod.rs), which is about wooden furniture.

### Version

```text
Current master (1cec8b30fac2c13504de7e7f950bd36c497858b1).
```

### Additional Labels

@rustbot label +A-documentation

Kyuuhachi:
Same in `self_named_module_files`, might be others too.

PR #8633, opened 2022-04-04
Escape `mod.rs` file mentions to avoid links in our documentation

xFrednet:
As the title says nothing special, still a fun fix :)

Closes: #8627

changelog: none


xFrednet:
Changelog entry was missing

@bors retry

PR #8606, opened 2022-03-29
Add [`err_expect`] lint

InfRandomness:
[`expect_err`] lint

- \[ ] Followed [lint naming conventions][lint_naming]
- \[x] Added passing UI tests (including committed `.stderr` file)
- \[x] `cargo test` passes locally
- \[x] Executed `cargo dev update_lints`
- \[x] Added lint documentation
- \[x] Run `cargo dev fmt`

Fixes https://github.com/rust-lang/rust-clippy/issues/1435

changelog: Added a lint to detect usage of .err().expect()

PR #8635, opened 2022-04-05
Add a lint to detect cast to unsigned for abs() and suggest unsigned_…

pbor:
…abs()

changelog: Add a [`cast_abs_to_unsigned`] that checks for uses of `abs()` that are cast to the corresponding unsigned integer type and suggest to replace them with `unsigned_abs()`.


tests/ui/cast.rs
```diff
@@ -46,12 +46,6 @@ fn main() {
     i64::MAX as u64;
     i128::MAX as u128;
 
-    (-1i8).abs() as u8;
-    (-1i16).abs() as u16;
-    (-1i32).abs() as u32;
-    (-1i64).abs() as u64;
-    (-1isize).abs() as usize;
-
```
giraffate, line 54: It would be better not to remove these tests. If we just suppress warnings, we can put `allow(clippy:: cast_abs_to_unsigned)`.

clippy_lints/src/casts/mod.rs
```diff
@@ -473,6 +474,28 @@ declare_clippy_lint! {
     "casts from an enum tuple constructor to an integer"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for uses of the `abs()` method that cast the result to unsigned.
+    ///
+    /// ### Why is this bad?
+    /// The `unsigned_abs()` method avoids panic when called on the MIN value.
```
giraffate, line 482: It looks like that the use of `wrapping_abs()` avoids panic. Why this lint catch the use of it?

giraffate (changes requested):
The `unsinged_abs` methods was added at 1.51, so we need [msrv](https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv) for this.

clippy_lints/src/casts/mod.rs
pbor: Because now that is available `unsigned_abs()` is nicer than `wrapping_abs() as u32`... It is true that in this case it is just a style matter, but since it was easy to catch I added it. If you prefer to remove it let me know.

tests/ui/cast.rs
pbor: Ok, thanks, I was not sure what was the preferred solution.

Both this and msrv should be ok now.

clippy_lints/src/casts/mod.rs
giraffate: IMO it would be better to remove `wrapping_abs()` because it confuses users.

pbor: Okay! done

giraffate (changes requested):
Can `clippy_lints/src/utils/conf.rs` and `tests/ui/min_rust_version_attr.rs` be updated for msrv? These will help you:
- https://github.com/rust-lang/rust-clippy/pull/8606
- https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv

pbor:
> Can `clippy_lints/src/utils/conf.rs` and `tests/ui/min_rust_version_attr.rs` be updated for msrv? These will help you:
> 
>     * #8606
> 
>     * https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv

Done. I also added a separate commit which actually fixes the `ERR_EXPECT` name in the conf :)

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8636, opened 2022-04-05
Add documentation on how to do a minimal changelog update

flip1995:
This ensures that the link to the Clippy version in the Rust release
blog post works correctly. The additional `(beta)` behind the previous
beta version breaks that link otherwise.

[Rendered](https://github.com/flip1995/rust-clippy/blob/release_changelog_quick_update/doc/release.md)

changelog: none

### DIFF
--- a/doc/release.md
+++ b/doc/release.md
@@ -121,4 +121,25 @@ happened a stable backport, make sure to re-merge those changes just as with the
 
 For this see the document on [how to update the changelog].
 
+If you don't have time to do a complete changelog update right away, just update
+the following parts:
+
+- Remove the `(beta)` from the new stable version:
+
+  ```markdown
+  ## Rust 1.XX (beta) -> ## Rust 1.XX
+  ```
+
+- Update the release date line of the new stable version:
+
+  ```markdown
+  Current beta, release 20YY-MM-DD -> Current stable, released 20YY-MM-DD
+  ```
+
+- Update the release date line of the previous stable version:
+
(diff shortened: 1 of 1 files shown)


xFrednet:
Alternatively, we could manually add an invisible html item with an id that stays constant regardless of the section title. I've seen that on an RFC recently. Not sure if that is better though 🤔

flip1995:
> Alternatively, we could manually add an invisible html item with an id that stays constant regardless of the section title. I've seen that on an RFC recently. Not sure if that is better though thinking

This would also have to be moved with every changelog update. This has to be done anyway, so it can also just be done during the release. The release is also connected with the sync, so this can also be just another commit in the sync PR.

Also it's not only that the Rust blog post link is wrong (That's not that big of a problem, since the changelog of the most recent stable is at the top anyway) but, that it might be confusing for people if we only write the beta changelog 4-5 weeks (like in this release cycle) after the release and the `## Rust 1.XX (beta)` heading stays there for the whole time.

xFrednet (approved):

flip1995:
I will do this quick changelog update on Thursday anyway. I just thought I might document it beforehand :smile:

xFrednet:
Thanks :+1: 

@bors r+

xFrednet:
Sounds good, I plan to write the full changelog for the next release (unless you want to xD)

flip1995:
> unless you want to xD

Definitely not :smile:

PR #8639, opened 2022-04-06
`trivially_copy_pass_by_ref` fixes

### LINKED ISSUE #2961: False positive with trivially_copy_pass_by_ref
The following code falsely assumes that `value` can be converted from reference to value when the value is returned as a reference by the function.

https://github.com/CraneStation/cranelift/blob/01de6842b1a3230db645a00f4ecf0176b5fdc65f/lib/codegen/src/regalloc/virtregs.rs#L100-L106

**ref_slice**

https://github.com/CraneStation/cranelift/blob/fb4acf8b5aaf63a96b9364d9b6f8d373fd8913c8/lib/codegen/src/ref_slice.rs#L12-L14

### LINKED ISSUE #5953: trivially_copy_pass_by_ref does not consider pointers
This is a similar issue to https://github.com/rust-lang/rust-clippy/issues/2946.

Recently it was discovered that a change recommended by the `trivially_copy_pass_by_ref` lint had started causing undefined behavior in Rust-SDL2: https://github.com/Rust-SDL2/rust-sdl2/pull/1020.

[This is the problematic change](https://github.com/Rust-SDL2/rust-sdl2/commit/a9893f0c0d42f012c312280ed3b0a785fcf9abcc#diff-fcecab478ea38ffde5351fe366399462L776):

```diff
 impl Point {
     // ...
-    pub fn raw(&self) -> *const sys::SDL_Point {
+    pub fn raw(self) -> *const sys::SDL_Point {
         self.raw
     }
 }
```

`Point` is a simple wrapper around an owned `sys::SDL_Point`, which itself is just a simple FFI-compatible data structure. Since `Point` implements `Copy`, the `trivially_copy_pass_by_ref` lint was recommended on `Point::raw` - even though making that change causes `Point::raw` to return a pointer to an immediately-dropped copy of one of its own fields.

Here's a minimal reproduction of the issue. The `Point::raw` method correctly produces a consistent pointer to the internal `RawPoint`, whereas the `Point::raw_linted` method produces a different pointer on each call.

```rust
#![forbid(clippy::pedantic)]

#[derive(Clone, Copy)]
struct RawPoint {
    pub x: u8,
}

#[derive(Clone, Copy)]
struct Point {
    pub raw: RawPoint,
}

impl Point {
    pub fn raw(&self) -> *const RawPoint {
        &self.raw
    }

    pub fn raw_linted(self) -> *const RawPoint {
        &self.raw
    }
}

fn main() {
    let p = Point { raw: RawPoint { x: 10 } };

    // This passes
    assert_eq!(p.raw(), p.raw());
    // This fails
    assert_eq!(p.raw_linted(), p.raw_linted());
}
```

This would be a bit more challenging to decide than the solution in https://github.com/rust-lang/rust-clippy/pull/2951. Pointers could be present as arbitrarily nested fields of a returned struct, which would not itself be a pointer. Further, pointers could be manipulated by long-lasting side effects (e.g. sent over a channel) during a function without actually ever being returned. Arguably it is the responsibility of the author of the `unsafe` dereference site to guarantee that the pointer remains valid, but it's still unreasonable to _recommend_ this change in these cases.

### Meta

- `cargo clippy -V`: clippy 0.0.212 (346aec9 2020-07-11)
- `rustc -Vv`:
  ```
  rustc 1.46.0-nightly (346aec9b0 2020-07-11)
  binary: rustc
  commit-hash: 346aec9b02f3c74f3fce97fd6bda24709d220e49
  commit-date: 2020-07-11
  release: 1.46.0-nightly
  LLVM version: 10.0
  ```

Jarcho:
fixes #5953
fixes #2961

The fix for #5953 is overly aggressive, but the suggestion is so bad that it's worth the false negatives. Basically three things together:
* It's not obviously wrong
* It compiles
* It may actually work when tested

changelog: Don't lint `trivially_copy_pass_by_ref` when unsafe pointers are used.
changelog: Better track lifetimes when linting `trivially_copy_pass_by_ref`.

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -7,6 +7,7 @@
 #![feature(let_chains)]
 #![feature(let_else)]
 #![feature(lint_reasons)]
+#![feature(never_type)]
 #![feature(once_cell)]
 #![feature(rustc_private)]
 #![feature(stmt_expr_attributes)]
--- a/clippy_lints/src/pass_by_ref_or_value.rs
+++ b/clippy_lints/src/pass_by_ref_or_value.rs
@@ -3,17 +3,19 @@ use std::iter;
 
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet;
-use clippy_utils::ty::is_copy;
+use clippy_utils::ty::{for_each_top_level_late_bound_region, is_copy};
 use clippy_utils::{is_self, is_self_ty};
+use core::ops::ControlFlow;
 use if_chain::if_chain;
 use rustc_ast::attr;
+use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_hir::intravisit::FnKind;
 use rustc_hir::{BindingAnnotation, Body, FnDecl, HirId, Impl, ItemKind, MutTy, Mutability, Node, PatKind};
 use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::ty;
 use rustc_middle::ty::layout::LayoutOf;
+use rustc_middle::ty::{self, RegionKind};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::def_id::LocalDefId;
 use rustc_span::{sym, Span};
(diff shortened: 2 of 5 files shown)


dswij (approved):
Looks great, thanks for this!

There's some conflicts that needs to be resolved, and after that I think this is good to merge

dswij:
Thanks! @bors r+

PR #8579, opened 2022-03-24
`unnecessary_join` lint

yoav-lavi:
Adds a lint called `unnecessary_join` that detects cases of `.collect::<Vec<String>>.join("")` or `.collect::<Vec<_>>.join("")` on an iterator, suggesting `.collect::<String>()` instead

Fixes: https://github.com/rust-lang/rust-clippy/issues/8570

This is a reopen of https://github.com/rust-lang/rust-clippy/pull/8573

changelog: add lint [`unnecessary_join`]

PR #8640, opened 2022-04-06
`unnecessary_join` documentation update

yoav-lavi:
changelog: none

Updates the description of `unnecessary_join` in accordance with https://github.com/rust-lang/rust-clippy/pull/8579#issuecomment-1089969859. I've also added a line regarding differences in assembly output, please let me know if it should also make it in.


flip1995:
Also answering your question from the other PR: I think having this in `pedantic` is fine. I don't think the performance impact is that big in cases where the collect+join is faster. 

Please squash the commits and we can merge this.

PR #8647, opened 2022-04-06
Only lint `mut_from_ref` when unsafe code is used

### LINKED ISSUE #6326: mut_from_ref should not trigger on code which is fully safe
The reason that is given for the lint is that "This is trivially unsound, as one can create two mutable references from the same (immutable!) source. This error actually lead to an interim Rust release 1.15.1." but this warning only makes sense when some of the code is unsafe.

This can occur without issue when using arenas, at least some of which follow the condition of producing an exclusive reference to new data using the shared reference to the arena.

```
$ cargo clippy -V                                                                                                                                      ~/code/rust/lang
clippy 0.0.212 (5404efc 2020-11-11)
```
```
$ rustc -Vv
rustc 1.49.0-nightly (5404efc28 2020-11-11)
binary: rustc
commit-hash: 5404efc28a0cddee103ef6396c48ea71ff9631c8
commit-date: 2020-11-11
release: 1.49.0-nightly
```

Jarcho:
fixes #6326

changelog: Only lint `mut_from_ref` when unsafe code is used.

### DIFF
--- a/clippy_lints/src/ptr.rs
+++ b/clippy_lints/src/ptr.rs
@@ -3,6 +3,7 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::expr_sig;
+use clippy_utils::visitors::contains_unsafe_block;
 use clippy_utils::{get_expr_use_or_unification_node, is_lint_allowed, path_def_id, path_to_local, paths};
 use if_chain::if_chain;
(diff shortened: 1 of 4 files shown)


giraffate (commented):
Can an explanation of why only unsafe is covered be added to `What it does` section?

Jarcho:
Just rewrote the whole thing. Functions caught by this lint are not trivially unsound.

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8648, opened 2022-04-06
Fix `unsound_collection_transmute`

### LINKED ISSUE #7706: Vec<MaybeUninit<u8>> transmute to Vec<u8>
Lint name: `unsound_collection_transmute`

I tried this code:

```rust
unsafe fn foo(v: Vec<std::mem::MaybeUninit<u8>>) -> Vec<u8> {
    std::mem::transmute(v)
}
```

I expected to see this happen: assuming MaybeUninit has been initialized, this should be safe, since MaybeUninit should have same layout as inner type

Instead, this happened:

```
error: transmute from `std::vec::Vec<std::mem::MaybeUninit<u8>>` to `std::vec::Vec<u8>` with mismatched layout is unsound
 --> src/lib.rs:2:5
  |
2 |     std::mem::transmute(v)
  |     ^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[deny(clippy::unsound_collection_transmute)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unsound_collection_transmute
```

Btw, it seems strange, but if I instead make the function generic, everything is ok:

```rust
unsafe fn foo_generic<T>(v: Vec<std::mem::MaybeUninit<T>>) -> Vec<T> {
    std::mem::transmute(v)
}

unsafe fn foo(v: Vec<std::mem::MaybeUninit<u8>>) -> Vec<u8> {
    foo_generic(v)
}
```

### Meta

Ran on [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=e042fbdde01589282a8fe321f82cc418)

Jarcho:
fixes #7706

changelog: Better check size and alignment requirements in `unsound_collection_transmute`

### DIFF
--- a/clippy_lints/src/transmute/utils.rs
+++ b/clippy_lints/src/transmute/utils.rs
@@ -1,10 +1,9 @@
 use clippy_utils::last_path_segment;
 use clippy_utils::source::snippet;
-use clippy_utils::ty::is_normalizable;
 use if_chain::if_chain;
 use rustc_hir::{Expr, GenericArg, QPath, TyKind};
 use rustc_lint::LateContext;
-use rustc_middle::ty::{self, cast::CastKind, Ty};
+use rustc_middle::ty::{cast::CastKind, Ty};
 use rustc_span::DUMMY_SP;
 use rustc_typeck::check::{cast::CastCheck, FnCtxt, Inherited};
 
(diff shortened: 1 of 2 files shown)


xFrednet (approved):
Looks good to me, thank you for the fix

Kixiron:
This is an incorrect "fix", the layout of `Vec` is not specified so `Vec<T>`  and `Vec<MaybeUninit<T>>` can have different layouts

Jarcho:
True, but the lint isn't to check for that. It's to check for when the contained type has a different size/alignment. The `Vec` to`Vec` transmute is formally unsound, but "works" (In this case, works in every version of rust so far).

The `transmute_undefined_repr` lint catches the rest. It's in the nursery now, but at some point it'll work.

Issue #7143, opened 2021-04-28
Suggest using find rather than implementing it using a for loop


teor2345:
### What it does

Suggests replacing a for loop that searches for an item in an iterable, with the find iterator method.

### Categories (optional)

- Kind: correctness

It's easier to make mistakes in a manual implementation. Using find is simpler and easier to maintain, focusing the attention on the predicate.

### Drawbacks

Suggestions might not work in complex edge cases.

If the lookup does a mapping, it might be harder to create a good suggestion. If it finds an index, we should suggest index instead.

### Example

```rust
const ARRAY: [u32; 5] = [2, 7, 1, 9, 3];

fn lookup(n: u32) -> Option<u32> {
    for v in ARRAY {
        if v == n {
            return Some(n);
        }
    }
    None
}
```

Could be written as:

```rust
const ARRAY: [u32; 5] = [2, 7, 1, 9, 3];

fn lookup(n: u32) -> Option<u32> {
    ARRAY.iter().find(|v| v == n)
}
```

llogiq:
That loop code example returns the element itself, not the index, so the suggestion would change behavior.

teor2345:
> That loop code example returns the element itself, not the index, so the suggestion would change behavior.

Thanks, I've made a playground and fixed a few other issues with the code.

camsteffen:
This could be generalized a bit to suggest `find_map` when applicable (possibly under a different lint).

ebobrow:
I thought I'd give this a try, but I've hit a bit of a wall. Currently, I have an implementation that replaces the for loop with an `if let` no matter what, so for something like the example code that would produce
```rust
const ARRAY: &[u32; 5] = &[2, 7, 1, 9, 3];
                                                        
fn lookup(n: u32) -> Option<u32> {
    if let Some(&v) = ARRAY.iter().find(|&&v| v == n) {
        return Some(v);
    }
    None
}

```
which is not great. Would it be worth it to hard code certain use cases (function return, variable assignments, etc.) to create better suggestions, and if so, what use cases? I'm drawing a blank because I'm guessing it would be more complicated than this in real code.

xFrednet:
I think for this, it's reasonable to hard-code for certain use cases. You probably first want to write the detection part for the `for`-loop, and then identify the use case. As for the cases, I would first focus on one, like the return or value assignment, and then create a PR. More cases can be added afterwards.

[...]

PR #8649, opened 2022-04-06
add [`manual_find`] lint for function return case

ebobrow:
part of the implementation discussed in #7143 

changelog: add [`manual_find`] lint for function return case

### DIFF
--- a/clippy_lints/src/loops/manual_find.rs
+++ b/clippy_lints/src/loops/manual_find.rs
@@ -1,0 +1,158 @@
+use super::utils::make_iterator_snippet;
+use super::MANUAL_FIND;
+use clippy_utils::{
+    diagnostics::span_lint_and_then, higher, is_lang_ctor, path_res, peel_blocks_with_stmt,
+    source::snippet_with_applicability, ty::implements_trait,
+};
+use if_chain::if_chain;
+use rustc_errors::Applicability;
+use rustc_hir::{
+    def::Res, lang_items::LangItem, BindingAnnotation, Block, Expr, ExprKind, HirId, Node, Pat, PatKind, Stmt, StmtKind,
+};
+use rustc_lint::LateContext;
+use rustc_span::source_map::Span;
+
+pub(super) fn check<'tcx>(
+    cx: &LateContext<'tcx>,
+    pat: &'tcx Pat<'_>,
+    arg: &'tcx Expr<'_>,
+    body: &'tcx Expr<'_>,
+    span: Span,
+    expr: &'tcx Expr<'_>,
+) {
+    let inner_expr = peel_blocks_with_stmt(body);
+    // Check for the specific case that the result is returned and optimize suggestion for that (more
+    // cases can be added later)
+    if_chain! {
+        if let Some(higher::If { cond, then, r#else: None, }) = higher::If::hir(inner_expr);
+        if let Some(binding_id) = get_binding(pat);
+        if let ExprKind::Block(block, _) = then.kind;
+        if let [stmt] = block.stmts;
+        if let StmtKind::Semi(semi) = stmt.kind;
+        if let ExprKind::Ret(Some(ret_value)) = semi.kind;
+        if let ExprKind::Call(Expr { kind: ExprKind::Path(ctor), .. }, [inner_ret]) = ret_value.kind;
+        if is_lang_ctor(cx, ctor, LangItem::OptionSome);
+        if path_res(cx, inner_ret) == Res::Local(binding_id);
+        if let Some((last_stmt, last_ret)) = last_stmt_and_ret(cx, expr);
+        then {
+            let mut applicability = Applicability::MachineApplicable;
+            let mut snippet = make_iterator_snippet(cx, arg, &mut applicability);
+            // Checks if `pat` is a single reference to a binding (`&x`)
+            let is_ref_to_binding =
+                matches!(pat.kind, PatKind::Ref(inner, _) if matches!(inner.kind, PatKind::Binding(..)));
+            // If `pat` is not a binding or a reference to a binding (`x` or `&x`)
+            // we need to map it to the binding returned by the function (i.e. `.map(|(x, _)| x)`)
+            if !(matches!(pat.kind, PatKind::Binding(..)) || is_ref_to_binding) {
+                snippet.push_str(
+                    &format!(
+                        ".map(|{}| {})",
+                        snippet_with_applicability(cx, pat.span, "..", &mut applicability),
+                        snippet_with_applicability(cx, inner_ret.span, "..", &mut applicability),
+                    )[..],
+                );
+            }
+            let ty = cx.typeck_results().expr_ty(inner_ret);
+            if cx.tcx.lang_items().copy_trait().map_or(false, |id| implements_trait(cx, ty, id, &[])) {
+                snippet.push_str(
+                    &format!(
+                        ".find(|{}{}| {})",
+                        "&".repeat(1 + usize::from(is_ref_to_binding)),
+                        snippet_with_applicability(cx, inner_ret.span, "..", &mut applicability),
+                        snippet_with_applicability(cx, cond.span, "..", &mut applicability),
+                    )[..],
+                );
+                if is_ref_to_binding {
+                    snippet.push_str(".copied()");
+                }
+            } else {
+                applicability = Applicability::MaybeIncorrect;
+                snippet.push_str(
+                    &format!(
+                        ".find(|{}| {})",
+                        snippet_with_applicability(cx, inner_ret.span, "..", &mut applicability),
+                        snippet_with_applicability(cx, cond.span, "..", &mut applicability),
+                    )[..],
+                );
+            }
+            // Extends to `last_stmt` to include semicolon in case of `return None;`
+            let lint_span = span.to(last_stmt.span).to(last_ret.span);
+            span_lint_and_then(
+                cx,
+                MANUAL_FIND,
+                lint_span,
+                "manual implementation of `Iterator::find`",
+                |diag| {
+                    if applicability == Applicability::MaybeIncorrect {
+                        diag.note("you may need to dereference some variables");
+                    }
+                    diag.span_suggestion(
+                        lint_span,
+                        "replace with an iterator",
+                        snippet,
+                        applicability,
+                    );
+                },
+            );
+        }
+    }
+}
+
+fn get_binding(pat: &Pat<'_>) -> Option<HirId> {
+    let mut hir_id = None;
+    let mut count = 0;
+    pat.each_binding(|annotation, id, _, _| {
+        count += 1;
+        if count > 1 {
+            hir_id = None;
+            return;
+        }
+        if let BindingAnnotation::Unannotated = annotation {
+            hir_id = Some(id);
+        }
+    });
+    hir_id
+}
+
+// Returns the last statement and last return if function fits format for lint
+fn last_stmt_and_ret<'tcx>(
+    cx: &LateContext<'tcx>,
+    expr: &'tcx Expr<'_>,
+) -> Option<(&'tcx Stmt<'tcx>, &'tcx Expr<'tcx>)> {
+    // Returns last non-return statement and the last return
+    fn extract<'tcx>(block: &Block<'tcx>) -> Option<(&'tcx Stmt<'tcx>, &'tcx Expr<'tcx>)> {
+        if let [.., last_stmt] = block.stmts {
+            if let Some(ret) = block.expr {
+                return Some((last_stmt, ret));
+            }
+            if_chain! {
+                if let [.., snd_last, _] = block.stmts;
+                if let StmtKind::Semi(last_expr) = last_stmt.kind;
+                if let ExprKind::Ret(Some(ret)) = last_expr.kind;
+                then {
+                    return Some((snd_last, ret));
+                }
+            }
+        }
+        None
+    }
+    let mut parent_iter = cx.tcx.hir().parent_iter(expr.hir_id);
+    if_chain! {
+        // This should be the loop
+        if let Some((node_hir, Node::Stmt(..))) = parent_iter.next();
+        // This should be the funciton body
+        if let Some((_, Node::Block(block))) = parent_iter.next();
+        if let Some((last_stmt, last_ret)) = extract(block);
+        if last_stmt.hir_id == node_hir;
+        if let ExprKind::Path(path) = &last_ret.kind;
+        if is_lang_ctor(cx, path, LangItem::OptionNone);
+        if let Some((_, Node::Expr(_block))) = parent_iter.next();
+        // This includes the function header
+        if let Some((_, func)) = parent_iter.next();
+        if func.fn_kind().is_some();
+        then {
+            Some((block.stmts.last().unwrap(), last_ret))
+        } else {
+            None
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,198 @@
...
+                    self.visit_pat(field.pat, ref_level);
+                }
+            },
+            PatKind::TupleStruct(_, pats, _) | PatKind::Or(pats) | PatKind::Tuple(pats, _) => {
+                for pat in pats {
+                    self.visit_pat(pat, ref_level);
+                }
+            },
+            PatKind::Box(p) => self.visit_pat(p, ref_level),
+            PatKind::Ref(p, _) => self.visit_pat(p, ref_level + 1),
```
Alexendoo, line 174: This doesn't quite work when references cross the boundary of another Pat, e.g. ref -> tuple -> ref, examples:

```rust
fn tuple_with_ref(v: [(u8, &u8); 3]) -> Option<u8> {
    for (_, &x) in v {
        if x > 10 {
            return Some(x);
        }
    }
    None
}
```
```rust
fn ref_to_tuple_with_ref(v: &[(u8, &u8)]) -> Option<u8> {
    for &(_, &x) in v {
        if x > 10 {
            return Some(x);
        }
    }
    None
}
```

Alexendoo (commented):
Hi, thanks for the PR!

So far I've mostly been looking at the pattern/ident stuff, I think it could do with being more conservative to deal with all the wild patterns out there, say for where `PAT` binds `binding`

```rust
for PAT in input {
    if COND {
        return Some(binding);
    }
    None
}
```

default to the assumption we're going to suggest

```rust
input.[into_iter/iter/etc]().map(|PAT| binding).find(|&binding| COND)
```

Then add more focused variants, like if PAT is already a `PatKind::Binding` omit the `.map()`, or if it's a single ref to a binding use `.copied()` instead

ebobrow:
Thanks for the review @Alexendoo and apologies for the delay! I think this should address your comments.

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,189 @@
...
+            if let Node::Item(parent_fn)
+                = cx.tcx.hir().get_by_def_id(cx.tcx.hir().get_parent_item(body.hir_id));
+            if let ItemKind::Fn(_, _, body_id) = parent_fn.kind;
+            let parent_fn = &cx.tcx.hir().body(body_id).value;
+            let ret_exprs = collect_ret_exprs(cx, parent_fn);
+            if ret_exprs.len() == 2;
+            // Some(binding)
+            if let ExprKind::Call(_, [binding]) = &ret_exprs[0].kind;
+            // None
+            if let ExprKind::Path(..) = &ret_exprs[1].kind;
```
Alexendoo, line 63: Checking the ret values runs into trouble with

```rust
fn for_in_condition(a: &[i32], b: bool) -> Option<i32> {
    if b {
        for &x in a {
            if x == 1 {
                return Some(x);
            }
        }
    }
    None
}
```
```rust
fn intermediate_statements(a: &[i32]) -> Option<i32> {
    for &x in a {
        if x == 1 {
            return Some(x);
        }
    }

    println!("side effect");

    None
}
```

I can't think of a pleasant way to check this other than mechanically using [`parent_iter`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/struct.Map.html#method.parent_iter). Passing `expr` into `manual_kind::check` from `mod.rs`, and calling `parent_iter` on that should give you:

- A `Stmt`
- A `Block`   
  - The above statement should be the last one of the block
  - The [`block.expr`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/struct.Block.html#structfield.expr) should be `Some(Expr { ExprKind::Path(path), .. })` where `is_lang_ctor(cx, path, OptionNone)`
- A fn-like thing, you can check [`Node::fn_sig`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/enum.Node.html#method.fn_sig) for this

there may be some extra `Expr`s mixed in like `ExprKind::Block` you have to skip over. You'd probably want to stash all this in a different function

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,189 @@
...
+
+#[derive(Default)]
+struct PatIdentCollector {
+    idents: Vec<Ident>,
+}
+
+impl PatIdentCollector {
+    fn visit_pat(&mut self, pat: &Pat<'_>) {
+        match pat.kind {
+            PatKind::Binding(BindingAnnotation::Unannotated, _, ident, None) => self.idents.push(ident),
```
Alexendoo, line 154: If there are idents we aren't handling that can cause problems, here `.idents` would have length 1

```rust
fn mixed_binding_modes(arr: Vec<(i32, String)>) -> Option<i32> {
    for (x, mut s) in arr {
        if x == 1 && s.as_mut_str().len() == 2 {
            return Some(x);
        }
    }
    None
}
```

We could have some function that returns `Option<Ident>` implemented using [`Pat::each_binding`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/struct.Pat.html#method.each_binding), if multiple idents appear or there's  one we aren't going to handle (`mut/ref/ref mut`) then it should return `None`

clippy_lints/src/loops/manual_find.rs
ebobrow: This works great, thanks! I've run into some trouble with the `fn_sig` in the case that the find implementation occurs within a closure, but it seems sufficient with the tests that I have to just check the `Stmt` and `Block` like you described. Is there an edge case that I'm not considering that means we have to check for a `fn_sig` as well?

Alexendoo: Ah whoops, I didn't know closures have no `FnSig`. Looks like [`fn_kind`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/enum.Node.html#method.fn_kind) instead would do the trick

Here's one I think would be a FP without such a check:

```rust
fn in_block(a: &[i32]) -> Option<i32> {
    let should_be_none = {
        for &x in a {
            if x == 1 {
                return Some(x);
            }
        }
        None
    };
    
    assert!(should_be_none.is_none());
    
    should_be_none
}
```

Transforming it into a find would remove the early return

ebobrow:
Thanks for the review! This should address your comments.

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,166 @@
...
+                    }
+                },
+                ExprKind::Call(..) => borrows = 1,
+                ExprKind::MethodCall(_, [caller, ..], _) => {
+                    if !ident_in_expr(cx, ident, caller) {
+                        borrows = 1;
+                    }
+                },
+                _ => {},
+            }
```
Alexendoo, line 77: I have found quite the spanner to throw in the works:

```rust
fn vec_string(strings: Vec<String>) -> Option<String> {
    for s in strings {
        if s == String::new() {
            return Some(s);
        }
    }
    None
}
```

both `|s|` and `|&s|` would result in errors:

```
error[E0277]: can't compare `&std::string::String` with `std::string::String`
  --> t.rs:25:36
   |
25 |     strings.into_iter().find(|s| s == String::new())
   |                                    ^^ no implementation for `&std::string::String == std::string::String`
   |
   = help: the trait `std::cmp::PartialEq<std::string::String>` is not implemented for `&std::string::String`
```
```
error[E0507]: cannot move out of a shared reference
  --> t.rs:25:31
   |
25 |     strings.into_iter().find(|&s| s == String::new())
   |                               ^-
   |                               ||
   |                               |data moved here
   |                               |move occurs because `s` has type `std::string::String`, which does not implement the `Copy` trait
   |                               help: consider removing the `&`: `s`
```

I believe the only way to create a correct suggestion here would be

```rust
strings.into_iter().find(|s| *s == String::new())
```

Alexendoo: One thing that you could do is check if `inner_ret` is `Copy`, in that case you can add the extra borrow unconditionally and the body of `find` can be unchanged

For the non `Copy` case it's more tricky, I *think* turning every use of `ident` into `(*ident)` would be correct, though it would be a fair bit of work creating such a snippet. I'd need to think a bit longer on the cases where you can leave it as `ident`

Another option would be to have an `Applicability::HasPlaceholders` suggestion for the non-`Copy` case, e.g. for the strings case suggesting

```rust
strings.into_iter().find(|s| { ... })
```

ebobrow: I went with the `Applicability::HasPlaceholders` implementation for now, but I can change it to dereference each `ident`. The only way I can think to do that would be to walk the tree and add partial snippets with asterisks and parentheses in between.

As for the cases that don't have to be dereferenced, all I can think of so far is a method called on `ident`, but I'm probably missing some.

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,145 @@
...
+// Returns last_return if function fits format for lint
+fn last_ret<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) -> Option<&'tcx Expr<'tcx>> {
+    let mut parent_iter = cx.tcx.hir().parent_iter(expr.hir_id);
+    let mut last_return = None;
+    if_chain! {
+        if let Some((node_hir, Node::Stmt(..))) = parent_iter.next();
+        if let Some((_, Node::Block(block))) = parent_iter.next();
+        if let [.., last_stmt] = block.stmts;
+        if last_stmt.hir_id == node_hir;
+        if let Some(last_ret) = block.expr;
```
flip1995, line 134: This won't cover `return None` as the last block expression/statement. I suggest to move the `strip_return` function in `matches::needless_match` to `clippy_utils` and reuse it here.

Please also add a test for this then.

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,145 @@
...
+                );
+            }
+            let mut needs_copy = false;
+            match cond.kind {
+                ExprKind::Binary(_, l, r) => {
+                    for side in [l, r] {
+                        if !matches!(
+                            side.kind,
+                            ExprKind::MethodCall(_, [caller, ..], _) if is_local_used(cx, caller, binding_id)
+                        ) && is_local_used(cx, side, binding_id)
```
flip1995, line 60: This code seems to deal with a really specific edge cases. Can you add some comments what those edge cases are, please?

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,145 @@
...
+        if let ExprKind::Call(Expr { kind: ExprKind::Path(ctor), .. }, [inner_ret]) = ret_value.kind;
+        if is_lang_ctor(cx, ctor, OptionSome);
+        if path_res(cx, inner_ret) == Res::Local(binding_id);
+        if let Some(last_ret) = last_ret(cx, expr);
+        then {
+            let mut applicability = Applicability::MachineApplicable;
+            let mut snippet = make_iterator_snippet(cx, arg, &mut applicability);
+            let is_ref_to_binding =
+                matches!(pat.kind, PatKind::Ref(inner, _) if matches!(inner.kind, PatKind::Binding(..)));
+            if !(matches!(pat.kind, PatKind::Binding(..)) || is_ref_to_binding) {
```
flip1995, line 44: Same here. It is not at all obvious to me what this code checks for.

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,145 @@
...
+                        ".find(|{}| {{ ... }})",
+                        snippet_with_applicability(cx, inner_ret.span, "..", &mut applicability),
+                    )[..],
+                );
+            }
+            span_lint_and_sugg(
+                cx,
+                MANUAL_FIND,
+                span.to(last_ret.span),
+                "manual implementation of Iterator::find",
```
flip1995, line 100:
```suggestion
                "manual implementation of `Iterator::find`",
```

code snippets should be in backticks.

clippy_lints/src/loops/mod.rs
```diff
@@ -597,6 +598,37 @@ declare_clippy_lint! {
...
+    /// Use instead:
+    /// ```rust
+    /// fn example(arr: Vec<i32>) -> Option<i32> {
+    ///     arr.into_iter().find(|&el| el == 1)
+    /// }
+    /// ```
+    #[clippy::version = "1.61.0"]
+    pub MANUAL_FIND,
+    complexity,
+    "Manual implementation of Iterator::find"
```
flip1995, line 629:
```suggestion
    "manual implementation of `Iterator::find`"
```

should be lower case plus backticks for code snippets.

clippy_lints/src/loops/manual_find.rs
Alexendoo: I believe this `match cond.kind` can be removed now, it only works for `Copy` types but if the type is `Copy` we should add the `&` unconditionally

ebobrow:
Thanks for the review @flip1995! I've applied your suggestions. I didn't end up using `strip_return` because I already had to do most of the pattern matching there to check if the last statement _was_ a return, but let me know if I'm missing something and should use it.

clippy_lints/src/loops/manual_find.rs
```diff
@@ -0,0 +1,158 @@
...
+        if path_res(cx, inner_ret) == Res::Local(binding_id);
+        if let Some((last_stmt, last_ret)) = last_stmt_and_ret(cx, expr);
+        then {
+            let mut applicability = Applicability::MachineApplicable;
+            let mut snippet = make_iterator_snippet(cx, arg, &mut applicability);
+            // Checks if `pat` is a single reference to a binding (`&x`)
+            let is_ref_to_binding =
+                matches!(pat.kind, PatKind::Ref(inner, _) if matches!(inner.kind, PatKind::Binding(..)));
+            // If `pat` is not a binding or a reference to a binding (`x` or `&x`)
+            // we need to map it to the binding returned by the function (i.e. `.map(|(x, _)| x)`)
```
flip1995, line 44: Why can't you do this in the `find(|(x, _)| ...)` closure?

flip1995 (commented):
Thanks for addressing this so quickly! I have one more question left.

clippy_lints/src/loops/manual_find.rs
ebobrow: If we did that, the function would return the tuple instead of just the `x`. In the imperative version, the code would look something like

```
for (x, _) in arr {
    if /* ... */ {
        return Some(x);
    }
}
```

so it only returns the `x` and ignores the rest of the pattern, but in a `find` the elements in the iterator remain as the entire tuple.

ebobrow:
Hey @flip1995, just checking in! Are you going to review this or have you been waiting for me to do something? (Apologies if so!)

flip1995:
I think you answered my question. I'll have another look on the PR soon and will then merge it. I don't think there's anything left for you to do.

Sorry for taking so long. My real life is a bit busy recently, so open source suffers... :/

ebobrow:
No worries! I'm in no rush, just wanted to make sure this hadn't been lost.

flip1995 (approved):

flip1995:
Thanks for the patience again!

@bors r+

PR #8655, opened 2022-04-07
Parse raw strings in `update_lints`

Alexendoo:
Fixes it not picking up `err_expect`, whose description is a raw string

changelog: none


flip1995:
Closing in favor of #8657  Thanks for addressing this so quickly!

PR #8657, opened 2022-04-07
Allow raw lint descriptions

flip1995:
update_lints now understands raw strings in declare_clippy_lint descriptions.

Supersedes  #8655

cc @Alexendoo thanks for addressing this so quickly. I build a little bit simpler version of your patch. I don't think it really matters what `Literal` we're trying to tokenize, since we assume later, that it is some sort of `str`.

changelog: none


Alexendoo:
If a raw string contained any `\`s it would matter, but I think the error would be ignored there so it would still build, the message might just be a bit off

flip1995:
Yeah, I guess we can live with that, honestly. I think the `desc` field is only used when printing the lints to the command line? But this is only for debugging purposes of Clippy devs. So if the message is off, then this is fine.

Issue #8663, opened 2022-04-08
dev experience: `cargo dev remove` + `cargo dev setup ... --remove`


yoav-lavi:
### Description

I think that having (in addition to the existing command) a `remove` flag / alias under setup (e.g. `cargo dev setup intellij ...`, `cargo dev setup intellij --remove`) might be good for visibility.

If it's been a while since you've setup the editor and now need to remove to commit, you're more likely to remember what you did or press up multiple times than remember the feedback info when running the command.

It also makes sense to me as setup and intellij are subcommands which you want to perform an additional operation on when removing the setup.

I've personally ended up rolling back the files manually since I didn't remember the additional command, so this would have been helpful.

Would be willing to open a PR for this if it's agreed upon!

(of course if there's a way to keep the setup permanent locally with a separate file and a .gitignore, that would be even better)

PR #8660, opened 2022-04-07
`unnecessary_owned_empty_strings`

yoav-lavi:
[`unnecessary_owned_empty_strings`]

Fixes https://github.com/rust-lang/rust-clippy/issues/8650

- \[x] Followed [lint naming conventions][lint_naming]
- \[x] Added passing UI tests (including committed `.stderr` file)
- \[x] `cargo test` passes locally
- \[x] Executed `cargo dev update_lints`
- \[x] Added lint documentation
- \[x] Run `cargo dev fmt`

[lint_naming]: https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints

changelog: Adds `unnecessary_owned_empty_strings`, a lint that detects passing owned empty strings to a function expecting `&str`

PR #8664, opened 2022-04-08
Allow passing `--remove` to `cargo dev setup <SUBCOMMAND>`

yoav-lavi:
changelog: none

Allows passing `--remove` to `cargo dev setup <SUBCOMMAND>` as an alternative to `cargo dev remove ...`

Fixes https://github.com/rust-lang/rust-clippy/issues/8663

### DIFF
--- a/clippy_dev/src/main.rs
+++ b/clippy_dev/src/main.rs
@@ -36,13 +36,31 @@ fn main() {
             }
         },
         ("setup", Some(sub_command)) => match sub_command.subcommand() {
-            ("intellij", Some(matches)) => setup::intellij::setup_rustc_src(
-                matches
-                    .value_of("rustc-repo-path")
-                    .expect("this field is mandatory and therefore always valid"),
-            ),
-            ("git-hook", Some(matches)) => setup::git_hook::install_hook(matches.is_present("force-override")),
-            ("vscode-tasks", Some(matches)) => setup::vscode::install_tasks(matches.is_present("force-override")),
+            ("intellij", Some(matches)) => {
+                if matches.is_present("remove") {
+                    setup::intellij::remove_rustc_src();
+                } else {
+                    setup::intellij::setup_rustc_src(
+                        matches
+                            .value_of("rustc-repo-path")
+                            .expect("this field is mandatory and therefore always valid"),
+                    );
+                }
+            },
+            ("git-hook", Some(matches)) => {
+                if matches.is_present("remove") {
+                    setup::git_hook::remove_hook();
+                } else {
+                    setup::git_hook::install_hook(matches.is_present("force-override"));
+                }
+            },
+            ("vscode-tasks", Some(matches)) => {
+                if matches.is_present("remove") {
+                    setup::vscode::remove_tasks();
+                } else {
+                    setup::vscode::install_tasks(matches.is_present("force-override"));
+                }
+            },
             _ => {},
         },
         ("remove", Some(sub_command)) => match sub_command.subcommand() {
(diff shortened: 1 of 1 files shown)


xFrednet (commented):
The changes look good to me. I think that `setup XYZ --remove` reads a bit weird, but I'm no UI designer, and it's worth the addition, if it helps at least one contributor.

Could you squash your commits? Afterwards, this can be merged :upside_down_face:

yoav-lavi:
@xFrednet Fair enough, it might be a bit weird.

I think what's throwing me off here is that it's verb noun rather than noun verb (due to the options being subcommands rather than arbitrary values) so I would possibly expect it to be `cargo dev git-hooks install / remove` (as in `git commit --amend`, `gh repo clone` etc.). 

You see verb noun more in cases with arbitrary values I think, e.g. `cargo install itertools`.

That being said, it's a matter of personal preference and it may need some more thought / input from others

xFrednet:
I chose this order for two reasons:
1. rustc's `x.py` file also has a setup command, which opens an interactive TUI to set up the project.
2. I wanted all setup and remove commands to be grouped together. `git commit --amend` is a good ui. Git also only has a single job. `clippy_dev` is a weird mixture of several tools, which can make it a bit messy IMO.

I don't think that there is a right or wrong here. Alternatively, we could add a command to the `setup` and `setup <object>` that mentions the `remove` command.

Since this is an internal Clippy-specific tool, I think just adding `--remove` as a kind of alias is alright. If you say that this is more intuitive for you, then there might be more users thinking the same :upside_down_face:

yoav-lavi:
I understand, just to clarify I didn't mean to say that the current method is wrong, just to explain why I personally saw the need to open the PR.

This is possibly a silly question but since I've had some issues with this:

I'm used to working with "squash and merge" or just normal commits and haven't needed to squash on my end up until now (and the previous PRs to clippy), what specifically am I expected to do when asked to squash my commits? The guides I've followed didn't really work as expected.

xFrednet:
Hey @yoav-lavi, it's not a stupid question. I only learned about squashing from this project.  :upside_down_face:

Squashing can basically combine multiple commits into one. There are a few ways to do it, and this is my way:

```sh
# On your branch
git rebase -i HEAD~6
```

The `6` is the number of commits you want to effect. This will open your editor with something like this:

```
pick 6c1216f1f Allow passing --remove to `cargo dev setup <SUBCOMMAND>`
pick 62d94c754 add missing args
pick 7d42ae967 unque name not needed
pick c2c3a3e63 more descriptive help
pick bf0212b28 formatting fixes
pick 3260bbf9a missing quote
```

Now you can indicate if and how you would like to change commits. In this case, you can use `s` to keep the other commit messages in the description in the base commit or `f` to just add the changes to the previous commit. I would do the following:

```
pick 6c1216f1f Allow passing --remove to `cargo dev setup <SUBCOMMAND>`
f 62d94c754 add missing args
f 7d42ae967 unque name not needed
f c2c3a3e63 more descriptive help
f bf0212b28 formatting fixes
f 3260bbf9a missing quote
```

and then save and close the file. If you pick `s` for squashing, you'll also get the option to change the commit message from the base commit. Afterwards, you just need to force push your changes:

```sh
git push --force-with-lease
# I have an alias for this and other git utilities. Here is my alias, if you want it
# alias git-push='git push --force-with-lease'
```

For a deeper understanding, I would recommend reading the documentation but this should you though almost everything :upside_down_face: 

> also, it seems like this is said on every PR, perhaps a bot could just post that when approving? To make reviewers' lives easier

The goal of squashing is to create a good history. With larger PRs it makes sense to split the implementation into several well named commits. They can also help with the review, to separate old changed from new ones. It's possible to automatically squash with `@bors` but that sadly looses the commit author. There is an open issue to fix this :)

For me, it depends on the type of commits that are present. Here we have one well named one and several commits that just fix something. That's why I asked you to squash. But I appreciate you watching out for us :upside_down_face:

yoav-lavi:
@xFrednet thanks! That's more or less what I've been doing using the guide I found, the problem mainly arises when pulling from master at some point, but it worked correctly on this attempt (although I tried again before seeing your message so the commit message might be a combined one rather than the first, wasn't aware of `f`).

xFrednet:
You're welcome! Are you referring to conflicts? Those are sadly more difficult to fix with rebasing in comparison to merging :/ .

I've heard that you can just merge and then squash the merge commit. But don't quote me on that ^^

xFrednet:
The changes and commit look good to me, thank you for the update and squashing the commits :upside_down_face: 

@bors r+

yoav-lavi:
I think the issue was mainly having X new commits that also need to be squashed, but I don't remember exactly what happened last time around. Will let you know if I notice it again. I wonder if bors could use commit impersonation to fix the issue, but the commits would not be verified (not sure if that's an issue)

yoav-lavi:
By the way @xFrednet (and feel completely free not to do so, only if you feel like it and have the time), I'm also waiting on https://github.com/rust-lang/rust-clippy/pull/8660, @flip1995 is assigned and I don't want to skip his review but since I don't think he's currently available I would love to know if there's something obvious that needs to be fixed since I may have time today / tomorrow to fix it.

Again, only if you want to!

xFrednet:
Sure, you can ping me the next time when you have a problem :upside_down_face: 

>  I wonder if bors could use commit impersonation to fix the issue, but the commits would not be verified (not sure if that's an issue)

That's the main thing holding this feature back. They've discussed impersonating the user, but that would mark them as unverified. I'm sadly not sure what the best solution would be. This works well enough for now :upside_down_face:

flip1995:
We should add "how to squash / rebase" instructions to our documentation. I always recommend to check the git diff with the remote branch and make sure it is empty before force pushing to the remote branch.

Having an alias for force pushing is really bad style in my opinion, because it makes it easier to do by accident. My alias is `git push --fwl` :smile: :stuck_out_tongue_closed_eyes:

PR #8666, opened 2022-04-08
Don't lint `while_let_loop` when significant drop order would change

### LINKED ISSUE #5715: lint result.unwrap_or_else(|| value)
Most of the `result.unwrap_or_else(|| <some value, or object property>)` code can be replaced with `result.unwrap_or(<some value>)`. This is mostly a readability issue and probably gives extra work to the inliner/optimizer similar to other redundant closures.

### LINKED ISSUE #5717: while_let_loop suggestion does not compile
```rust
use std::cell::RefCell;
use std::rc::Rc;

#[derive(Clone)]
struct X {
    x: Rc<RefCell<Option<X>>>,
}

fn main() {
    let mut x = X {
        x: Rc::new(RefCell::new(None)),
    };
    loop {
        let tmp = match *x.x.borrow() {
            Some(ref y) => y.clone(),
            _ => break,
        };
        x = tmp;
    }  // try: `while let Some(ref y) = *x.x.borrow() { .. }
}
```

But

```
error[E0506]: cannot assign to `x` because it is borrowed
  --> src/main.rs:22:9
   |
20 |     while let Some(ref y) = *x.x.borrow() {
   |                              ------------
   |                              |          |
   |                              |          ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `std::cell::Ref<'_, std::option::Option<X>>`
   |                              borrow of `x` occurs here
   |                              a temporary with access to the borrow is created here ...
21 |         let tmp = y.clone();
22 |         x = tmp;
   |         ^ assignment to borrowed `x` occurs here
```

### LINKED ISSUE #7226: False `clippy::while_let_loop` when doing `loop { if let }` in spawned thread.
Lint name: `while_let_loop`

I tried this code:

[(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=064acd8b79add4ff31946d92be059a83)

```rust
use std::{
    sync::{Arc, RwLock},
    thread::{sleep, spawn, JoinHandle},
    time::Duration,
};

// NOTE: the following function is a simplified version of the true function whose signature is generic.
pub fn spawn_observer(observed: Arc<RwLock<Option<i32>>>, observer: fn(i32), interval: Duration) -> JoinHandle<()> {
    spawn(move || loop {
        if let Some(value) = *observed.read().unwrap() {
            observer(value);
        } else {
            break;
        }
        sleep(interval);
    })
}
```

I expected to see this happen: No warnings.

Instead, this happened: It suggests `while let` loop, which would cause `observer` to lock the `observed`, blocking the other threads that also read/mutate the `observed`.

### Meta

- `cargo clippy -V`: clippy 0.1.51 (2fd73fa 2021-03-23)
- `rustc -Vv`:
  ```
  rustc 1.51.0 (2fd73fabe 2021-03-23)
  binary: rustc
  commit-hash: 2fd73fabe469357a12c2c974c140f67e7cdd76d0
  commit-date: 2021-03-23
  release: 1.51.0
  LLVM version: 11.0.1
  ```

### LINKED ISSUE #7913: clippy::while_let_loop incorectly scope variable
Lint name: clippy::while_let_loop

I tried this code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e0c751f49cce697e1f06bcc1fed429e

I simplify it by changing instances to only 5, and "work" to just short `sleep` to run better on playground, but I have 32 core CPU, so using more than one core at time is important for me.

```rust
loop {
// lock mutex to obtain next thing to process, and unlock it before processing it
    let seed = if let Some(seed) = seeds.lock().unwrap().pop_front() {
        seed
    } else {
        break
    };
    // do something
}
```

I expected to see this happen: no lint, or suggestion that does not break the code

Instead, this happened: suggested code is running one thread at time, instead of many, which is significant performance impact.

### Meta

```
rustc 1.58.0-nightly (29b124802 2021-10-25)
binary: rustc
commit-hash: 29b1248025b19bd132c8047fc710ea9314b9b76b
commit-date: 2021-10-25
host: x86_64-pc-windows-msvc
release: 1.58.0-nightly
LLVM version: 13.0.0
```

Jarcho:
fixes #7226
fixes #7913
fixes #5717

For #5717 it may not stay fully fixed. This is only completely fixed right now due to all the allowed drop impls have `#[may_dangle]` on their drop impls. This may get changed in the future based on how significant drops are determined, but the example listed with `RefCell` shouldn't break.

changelog: Don't lint `while_let_loop` when the order of significant drops would change

### DIFF
--- a/clippy_lints/src/loops/while_let_loop.rs
+++ b/clippy_lints/src/loops/while_let_loop.rs
@@ -2,71 +2,60 @@ use super::WHILE_LET_LOOP;
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::higher;
 use clippy_utils::source::snippet_with_applicability;
+use clippy_utils::ty::needs_ordered_drop;
+use clippy_utils::visitors::any_temporaries_need_ordered_drop;
 use rustc_errors::Applicability;
-use rustc_hir::{Block, Expr, ExprKind, MatchSource, Pat, StmtKind};
-use rustc_lint::{LateContext, LintContext};
-use rustc_middle::lint::in_external_macro;
+use rustc_hir::{Block, Expr, ExprKind, Local, MatchSource, Pat, StmtKind};
+use rustc_lint::LateContext;
 
 pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, loop_block: &'tcx Block<'_>) {
-    // extract the expression from the first statement (if any) in a block
-    let inner_stmt_expr = extract_expr_from_first_stmt(loop_block);
-    // or extract the first expression (if any) from the block
-    if let Some(inner) = inner_stmt_expr.or_else(|| extract_first_expr(loop_block)) {
-        if let Some(higher::IfLet {
-            let_pat,
-            let_expr,
-            if_else: Some(if_else),
-            ..
-        }) = higher::IfLet::hir(cx, inner)
-        {
-            if is_simple_break_expr(if_else) {
-                could_be_while_let(cx, expr, let_pat, let_expr);
+    let (init, has_trailing_exprs) = match (loop_block.stmts, loop_block.expr) {
+        ([stmt, stmts @ ..], expr) => {
+            if let StmtKind::Local(&Local { init: Some(e), .. }) | StmtKind::Semi(e) | StmtKind::Expr(e) = stmt.kind {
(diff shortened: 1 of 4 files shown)


dswij (approved):
This looks great, thanks for the changes!

dswij:
Looks like there's some conflicts, and after that I think this is good to merge

dswij:
Thanks! @bors r+

PR #8669, opened 2022-04-09
fix comments in test for split_once

kyoto7250:
This PR fixed comments in test.

`split_once` was stabilized in 1.52, so I think the comments maybe be wrong.

ref:
https://doc.rust-lang.org/std/string/struct.String.html#method.split_once

thank you in advance.

changelog: none


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update :upside_down_face: 

@bors r+

PR #8672, opened 2022-04-10
Adding test about msrv for map clone

### LINKED ISSUE #8623: map_clone should not suggest copied() for MSRV < 1.36
### Summary

I'm backporting some code to work on 1.32 and noticed that the map_clone lint suggests using copied() even though I've configured Clippy with MSRV = 1.32. Would be nice if this wouldn't happen.

### Lint Name

map_clone

### Reproducer

_No response_

### Version

_No response_

### Additional Labels

_No response_

kyoto7250:
Close #8623

This PR adds tests for `map_clone`, because this rule does not have tests about MSRV.

I have tried, but the contents of the issue could not be verified.

The rule seems to separate `copied` and `cloned` according to the value of MSRV. `cloned` was [established in 1.1.0](https://doc.rust-lang.org/stable/std/iter/struct.Cloned.html), so this logic seems to be correct. (maybe misreading of clone and copied?)

https://github.com/rust-lang/rust-clippy/blob/c7e68638431b2a6639fcbeb44de790619de3b9b1/clippy_lints/src/map_clone.rs#L151-L155

However, [other tests seem to test according to the msrv value](https://github.com/rust-lang/rust-clippy/blob/c7e68638431b2a6639fcbeb44de790619de3b9b1/tests/ui/if_then_some_else_none.rs#L69-L88),
so I created PR to add tests.

Thank you in advance.

changelog: None

### DIFF
--- a/tests/ui/map_clone.fixed
+++ b/tests/ui/map_clone.fixed
@@ -1,4 +1,6 @@
 // run-rustfix
+
+#![feature(custom_inner_attributes)]
 #![warn(clippy::map_clone)]
 #![allow(
     clippy::clone_on_copy,
@@ -60,4 +62,15 @@ fn main() {
 
         let _ = Some(RefCell::new(String::new()).borrow()).map(|s| s.clone());
     }
+}
+
+fn _msrv_1_35() {
+    #![clippy::msrv = "1.35"]
+    // `copied` was stabilized in 1.36, so clippy should use `cloned`.
+    let _: Vec<i8> = vec![5_i8; 6].iter().cloned().collect();
+}
+
+fn _msrv_1_36() {
+    #![clippy::msrv = "1.36"]
+    let _: Vec<i8> = vec![5_i8; 6].iter().copied().collect();
 }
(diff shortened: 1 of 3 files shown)


giraffate:
We've already have the tests for `msrv` of `map_clone` in https://github.com/rust-lang/rust-clippy/blob/5344236715fafa2b8d27320839f44d0f4b5ca955/tests/ui-toml/min_rust_version/min_rust_version.rs#L72-L75

At first, we need the reproduction, so I'm asking it in the issue.

giraffate:
@kyoto7250 We need to cover the `Iterator::copied()` case for this: https://github.com/rust-lang/rust-clippy/issues/8623#issuecomment-1094705576

kyoto7250:
@giraffate 

I do not think it is necessary to merge this Pull Request, so this PR close and another PR open for fixing about `[unnecessary_to_owned.rs](http://unnecessary_to_owned.rs/)`

kyoto7250:
the PR is here.
https://github.com/rust-lang/rust-clippy/pull/8692

giraffate:
@kyoto7250 

> I do not think it is necessary to merge this Pull Request, so this PR close and another PR open for fixing about unnecessary_to_owned.

Yes, I'm closing this. Anyway, thanks for working on this!

> Can I also request a review of a new PR?

Okay!

PR #8646, opened 2022-04-06
Fix `as_deref_mut` false positives in `needless_option_as_deref`

Alexendoo:
Also moves it into `methods/`

Fixes #7846
Fixes #8047

changelog: [`needless_option_as_deref`]: No longer lints for `as_deref_mut` on Options that cannot be moved

supersedes #8064

PR #8676, opened 2022-04-10
Check for loops/closures in `local_used_after_expr`

Alexendoo:
Follow up to #8646, catches when a local is used multiple times because it's in a loop or a closure

changelog: none


xFrednet (approved):

xFrednet:
:wave: 

Looks good to me, thank you for the update! :upside_down_face: 

@bors r+

PR #8213, opened 2022-01-03
Add `manual_bits` lint

paolobarbolini:
Closes #6670 

---

changelog: new lint: [`manual_bits`]

PR #8677, opened 2022-04-10
Add `usize` cast to `clippy::manual_bits` suggestion

xFrednet:
A fix for the suggestion from https://github.com/rust-lang/rust-clippy/pull/8213

changelog: [`manual_bits`]: The suggestion now includes a cast for proper type conversion

### DIFF
--- a/clippy_lints/src/manual_bits.rs
+++ b/clippy_lints/src/manual_bits.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::source::snippet_opt;
-use clippy_utils::{meets_msrv, msrvs};
+use clippy_utils::source::snippet_with_applicability;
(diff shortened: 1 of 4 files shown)


giraffate:
The test fails in `ui/manual_bits.rs`, can this be updated?

xFrednet:
Thank you for the message, I've missed the failed CI. Now it should be ready, the update only added `clippy::useless_conversion` to the allowed lint list in the test file :)

giraffate (commented):
Can the example in doc be updated like below?
```
    /// ### Example
    /// ```rust
    /// std::mem::size_of::<usize>() * 8;
    /// ```
    /// Use instead:
    /// ```rust
    /// usize::BITS as usize;
    /// ```
```

giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #2624, opened 2018-04-03
New Lint: parameter of function only used in recursive call


flip1995:
If a parameter of a function is only used in a recursive call of this functions body, it is probably unintended and useless:
```rust
fn rec_fn(a: i32) {
    //some conditions, no a... 
    rec_fn(a) 
} 
```
See also https://github.com/rust-lang-nursery/rust-clippy/pull/2622#comment-178952187

jamesmcm:
This should be addressed by #8422

xFrednet:
Thank you for pointing that out @jamesmcm

PR #8678, opened 2022-04-10
Add extra check to format_impl lint for self.fmt()

jamesmcm:
As mentioned in this Reddit comment:
https://www.reddit.com/r/rust/comments/tq602q/my_first_clippy_lint_detecting_use_of_self_as/i2fhdy1/

Previously use of self.fmt within the Format trait wasn't detected

Note we don't check fully qualified paths atm, but that shouldn't happen by accident anyway.

changelog: format_impl lint now checks for recursive use of self.fmt() too.

### DIFF
--- a/clippy_lints/src/format_impl.rs
+++ b/clippy_lints/src/format_impl.rs
@@ -131,8 +131,35 @@ impl<'tcx> LateLintPass<'tcx> for FormatImpl {
             check_to_string_in_display(cx, expr);
         }
 
+        check_self_fmt_in_fmt(cx, expr, format_trait_impl);
         check_self_in_format_args(cx, expr, format_trait_impl);
         check_print_in_format_impl(cx, expr, format_trait_impl);
+    }
+}
+
+fn check_self_fmt_in_fmt(cx: &LateContext<'_>, expr: &Expr<'_>, format_trait: FormatTrait) {
+    if_chain! {
+        // Get the hir_id of the object we are calling the method on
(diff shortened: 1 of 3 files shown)


giraffate (commented):
Would this occur regardless of `Display` or `Debug` trait? It appears to be simply a recursive call.

If so, it would be better to implement this as a new lint. It may be similar to https://github.com/rust-lang/rust-clippy/issues/2624.

jamesmcm:
That's a good point, and yeah that would catch it too and is a more general solution.

The only difficult part there is writing the lint so it works for general functions and trait methods.

jamesmcm:
It'd be similar to #8422 but that doesn't catch the case of ever having identical recursion e.g. recursing with exactly the same arguments we take in, at least when they are edited in another conditional branch atm.

i.e. it didn't trigger on:

```rust

enum Foo {
    Foo(usize),
    Bar(usize),
}

impl std::fmt::Display for Foo {
        match self {
            Self::Foo(_) => self.fmt(f),
            Self::Bar(x) => x.fmt(f),
        }
    }
}
```

giraffate (commented):
Thanks!

It might be better to enhance the `only_used_in_recursion` lint to catch this case.

blyxyas:
@jamesmcm I'm going to close this for now (we're doing cleaning of old PRs :heart:) are you interested in continuing development as a fix to `only_used_in_recursion`?

If you're still interested in this PR, don't hesitate in re-open it and continue your work.

PR #8687, opened 2022-04-11
Fix subtraction overflow in `cast_possible_truncation`

Alexendoo:
changelog: Fix false negative due to subtraction overflow in `cast_possible_truncation`

I *think* a false negative is the worst that can happen from this


xFrednet:
Looks like highfive is a bit slow rn. ^^

r? @xFrednet

xFrednet:
Looks like highfive is a bit slow rn. ^^

r? @xFrednet

xFrednet:
Looks good to me, thank you for the update. This even ICEs in debug builds :upside_down_face: 

@bors r+

PR #8691, opened 2022-04-12
Prevent infinite (exponential) recursion in only_used_in_recursion

### LINKED ISSUE #8689: Infinite loop in `clippy::only_used_in_recursion`
### Summary

Since 1.61.0-beta.1, one of our crate gets stuck in an infinite loop. I was able to GDB into clippy and figure out that it gets stuck while running the [`clippy::only_used_in_recursion`](https://github.com/rust-lang/rust-clippy/blob/d0cf3481a84e3aa68c2f185c460e282af36ebc42/clippy_lints/src/only_used_in_recursion.rs) that was recently introduced. Unfortunately, I was not able to understand why it gets stuck.

### Reproducer

Running `cargo +beta clippy` on the [`dnp3-schema`](https://github.com/stepfunc/dnp3/tree/488aeb11038da8670fd7696e8d37ccafd8ecd19f/ffi/dnp3-schema) crate gets stuck in an infinite loop in 1.61.0-beta.1 and up (including nightly)

### Version

```text
rustc 1.61.0-beta.2 (7c13df853 2022-04-09)
binary: rustc
commit-hash: 7c13df853721b60a03e7c0bb084d2eb1e27a9caa
commit-date: 2022-04-09
release: 1.61.0-beta.2
LLVM version: 14.0.0
```

### Additional Labels

_No response_

flip1995:
This simplifies the visitor code a bit and prevents checking expressions
multiple times. I still think this lint should be removed for now,
because its code isn't really tested.

Fixes #8689 

**NOTE:** Before merging this, we should talk about removing and revisiting this lint. See my comment in #8689 

changelog: prevent infinite recursion in [`only_used_in_recursion`]

### DIFF
--- a/clippy_lints/src/only_used_in_recursion.rs
+++ b/clippy_lints/src/only_used_in_recursion.rs
@@ -8,7 +8,7 @@ use rustc_errors::Applicability;
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::def_id::DefId;
 use rustc_hir::definitions::{DefPathData, DisambiguatedDefPathData};
-use rustc_hir::intravisit::{walk_expr, FnKind, Visitor};
+use rustc_hir::intravisit::{walk_expr, walk_stmt, FnKind, Visitor};
 use rustc_hir::{
     Arm, Block, Body, Expr, ExprKind, Guard, HirId, ImplicitSelfKind, Let, Local, Pat, PatKind, Path, PathSegment,
     QPath, Stmt, StmtKind, TyKind, UnOp,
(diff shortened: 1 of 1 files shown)


clippy_lints/src/only_used_in_recursion.rs
```diff
@@ -250,40 +245,30 @@ pub struct SideEffectVisit<'tcx> {
...
         }
+        self.ret_vars.clear();
     }
 
     fn visit_expr(&mut self, ex: &'tcx Expr<'tcx>) {
+        if !self.visited_exprs.insert(ex.hir_id) {
+            return;
+        }
         match ex.kind {
             ExprKind::Array(exprs) | ExprKind::Tup(exprs) => {
```
flip1995, line 273: An example for missing tests: This whole arm could be removed completely and the tests keep passing, even though this changes the semantics of the lint.

llogiq:
Thank you! That's what I would have done there, too.

@bors r+

flip1995:
> **NOTE:** Before merging this, we should talk about removing and revisiting this lint. See my comment in #8689

That still stands. But let's talk about this on Zulip

Issue #8492, opened 2022-03-01
suggest map_or  for simple None substitutions


matthiaskrgr:
### What it does

Find patterns like
````rust
pub fn a(s: Option<String>) -> usize {
    match s {
        Some(string) => string.len(),
        None => 1, 
    }
}
````

and suggest to use `map_or` instead:

````rust
pub fn a(s: Option<String>) -> usize {
    s.map_or(1, |string| string.len())
}
````

We might need to consider restricting such a lint to simple matches that don't have a lot of "stuff going on"

### Lint Name

map_or_match? matching_map_or? mmh

### Category

style

### Advantage

for simple matches, this reduces lines from 4 to 1, and I think it's a bit easier to grasp what is going on if you are familiar with `map_or`.

### Drawbacks

:shrug: 

We might want to bail out on matches where a single arm spans several lines.

### Example

```rust
pub fn a(s: Option<String>) -> usize {
    match s {
        Some(string) => string.len(),
        None => 1, 
    }
}
```

Could be written as:

```rust
pub fn a(s: Option<String>) -> usize {
    s.map_or(1, |string| string.len())
}
```

J-ZhengLi:
Seems like this hasn't been fixed yet... @rustbot claim

J-ZhengLi:
> Just a note for anyone taking this. There is `option_if_let_some_else_none` which covers this only for `if let ... else ...`.

should I extend that lint and change its name or add a new one?

PR #8696, opened 2022-04-14
check for if-some-or-ok-else-none-or-err

J-ZhengLi:
fixes: #8492 

---

changelog: make [`option_if_let_else`] to check for match expression with both Option and Result; **TODO: Change lint name? Add new lint with similar functionality?**

### DIFF
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -7,15 +7,18 @@ use clippy_utils::{
 };
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::LangItem::OptionSome;
-use rustc_hir::{def::Res, BindingAnnotation, Expr, ExprKind, Mutability, PatKind, Path, QPath, UnOp};
+use rustc_hir::LangItem::{OptionNone, OptionSome};
+use rustc_hir::{
+    def::Res, Arm, BindingAnnotation, Expr, ExprKind, MatchSource, Mutability, Pat, PatKind, Path, QPath, UnOp,
+};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::sym;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Lints usage of `if let Some(v) = ... { y } else { x }` which is more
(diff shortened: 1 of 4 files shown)


clippy_lints/src/option_if_let_else.rs
```diff
@@ -99,43 +119,39 @@ fn format_option_in_sugg(cx: &LateContext<'_>, cond_expr: &Expr<'_>, as_ref: boo
...
-        if !expr.span.from_expansion(); // Don't lint macros, because it behaves weirdly
-        if !in_constant(cx, expr.hir_id);
-        if let Some(higher::IfLet { let_pat, let_expr, if_then, if_else: Some(if_else) })
-            = higher::IfLet::hir(cx, expr);
-        if !is_else_clause(cx.tcx, expr);
-        if !is_result_ok(cx, let_expr); // Don't lint on Result::ok because a different lint does it already
-        if let PatKind::TupleStruct(struct_qpath, [inner_pat], _) = &let_pat.kind;
-        if is_lang_ctor(cx, struct_qpath, OptionSome);
-        if let PatKind::Binding(bind_annotation, _, id, None) = &inner_pat.kind;
+        if !is_result_ok(cx, cond_expr); // Don't lint on Result::ok because a different lint does it already
```
J-ZhengLi, line 134: This was with the original code, I wonder what lint it was refering to? cuz I think removing it does nothing new but I was too afraid to remove it. @flip1995

Serial-ATA:
Should the lint be renamed if it's going to cover `match` now as well?

Serial-ATA:
I think this should only check for arms that span a single line.

I don't think going from

```rust
    let _ = match res {
        Err(_) => 1,
        Ok(mut a) => {
            a += 1;
            println!("{a}");
            a -= 1;
            let b = 5;
            a += b;
            a
        }
    };
```

to

```rust
    let _ = res.map_or(1, |mut a| {
            a += 1;
            println!("{a}");
            a -= 1;
            let b = 5;
            a += b;
            a
        });
```

makes much of a difference.

flip1995:
Thanks! Sorry for taking so long. I rebased it and did a minor cleanup. About the lint name: Yes it will need to be changed. But since this lint is currently in `nursery` there's no rush to decide this now. I added a FIXME comment for it.

@bors r+

PR #8699, opened 2022-04-14
Don't lint `vec_init_then_push` when further extended

### LINKED ISSUE #7071: vec_init_then_push should only trigger if there are no further pushes
Lint name: vec_init_then_push

I tried this code:

```rust
    let mut arg_strings: Vec<Box<str>> = Vec::new();
    arg_strings.push(name.to_owned().into_boxed_str());
    for arg in args {
        arg_strings.push(
            arg.as_ref()
                .to_os_string()
                .into_string()
                .unwrap()
                .into_boxed_str(),
        );
    }
```

I expected to see this happen: The lint doesn't trigger, because using `vec![]` and then immediately pushing doesn't improve the code any, it uses two different styles of initializing the vector.

Instead, this happened:

```
warning: calls to `push` immediately after creation
   --> tests/mock/clitools.rs:613:5
    |
613 | /     let mut arg_strings: Vec<Box<str>> = Vec::new();
614 | |     arg_strings.push(name.to_owned().into_boxed_str());
    | |_______________________________________________________^ help: consider using the `vec![]` macro: `let mut arg_strings: Vec<Box<str>> = vec![..];`
```

cc https://github.com/rust-lang/rustup/pull/2718, @kinnison

### Meta

- `cargo clippy -V`: clippy 0.1.53 (dae9d6a 2021-04-09)
- `rustc -v`: rustc 1.53.0-nightly (dae9d6ac3 2021-04-09)

Jarcho:
fixes #7071

This will still lint when a larger number of pushes are done (four currently). The exact number could be debated, but this is more readable then a sequence of pushes so it shouldn't be too large. 

changelog: Don't lint `vec_init_then_push` when further extended.
changelog: Remove `mut` binding from `vec_init_then_push` when possible.

### DIFF
--- a/clippy_lints/src/vec_init_then_push.rs
+++ b/clippy_lints/src/vec_init_then_push.rs
@@ -18,6 +18,13 @@ declare_clippy_lint! {
     /// ### What it does
     /// Checks for calls to `push` immediately after creating a new `Vec`.
     ///
+    /// If the `Vec` is created using `with_capacity` this will only lint if the capacity is a
+    /// constant and the number of pushes is greater than or equal to the initial capacity.
+    ///
+    /// If the `Vec` is extended after the initial sequence of pushes and it was default initialized
+    /// then this will only lint after there were at least four pushes. This number may change in
+    /// the future.
+    ///
     /// ### Why is this bad?
     /// The `vec![]` macro is both more performant and easier to read than
     /// multiple `push` calls.
@@ -56,7 +63,7 @@ struct VecPushSearcher {
 }
 impl VecPushSearcher {
     fn display_err(&self, cx: &LateContext<'_>) {
-        let min_pushes_for_extension = match self.init {
+        let required_pushes_before_extension = match self.init {
             _ if self.found == 0 => return,
             VecInitKind::WithConstCapacity(x) if x > self.found => return,
             VecInitKind::WithConstCapacity(x) => x,
@@ -98,6 +105,8 @@ impl VecPushSearcher {
                         && adjusted_mut == Mutability::Mut
                         && !adjusted_ty.peel_refs().is_slice() =>
                 {
+                    // No need to set `needs_mut` to true. The receiver will be either explicitly borrowed, or it will
+                    // be implicitly borrowed via an adjustment. Both of these cases are already handled by this point.
                     return ControlFlow::Break(true);
                 },
                 ExprKind::Assign(lhs, ..) if e.hir_id == lhs.hir_id => {
@@ -110,7 +119,7 @@ impl VecPushSearcher {
         });
 
         // Avoid allocating small `Vec`s when they'll be extended right after.
-        if res == ControlFlow::Break(true) && self.found <= min_pushes_for_extension {
+        if res == ControlFlow::Break(true) && self.found <= required_pushes_before_extension {
             return;
         }
 
--- a/clippy_lints/src/vec_init_then_push.rs
+++ b/clippy_lints/src/vec_init_then_push.rs
@@ -1,14 +1,18 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::higher::{get_vec_init_kind, VecInitKind};
 use clippy_utils::source::snippet;
-use clippy_utils::{path_to_local, path_to_local_id};
-use if_chain::if_chain;
+use clippy_utils::visitors::for_each_local_use_after_expr;
+use clippy_utils::{get_parent_expr, path_to_local_id};
+use core::ops::ControlFlow;
 use rustc_errors::Applicability;
-use rustc_hir::{BindingAnnotation, Block, Expr, ExprKind, HirId, Local, PatKind, Stmt, StmtKind};
+use rustc_hir::def::Res;
+use rustc_hir::{
+    BindingAnnotation, Block, Expr, ExprKind, HirId, Local, Mutability, PatKind, QPath, Stmt, StmtKind, UnOp,
+};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
 use rustc_session::{declare_tool_lint, impl_lint_pass};
-use rustc_span::Span;
+use rustc_span::{Span, Symbol};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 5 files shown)


xFrednet:
Hey @Jarcho, could you also rebase. I think the CI is blocked on that :upside_down_face:

clippy_lints/src/vec_init_then_push.rs
```diff
@@ -43,26 +54,86 @@ pub struct VecInitThenPush {
...
+                        == Some(Mutability::Mut)
+                        || get_parent_expr(cx, last_place)
+                            .map_or(false, |e| matches!(e.kind, ExprKind::AddrOf(_, Mutability::Mut, _)));
+                },
+                ExprKind::MethodCall(_, [recv, ..], _)
+                    if recv.hir_id == e.hir_id
+                        && adjusted_mut == Mutability::Mut
+                        && !adjusted_ty.peel_refs().is_slice() =>
+                {
+                    return ControlFlow::Break(true);
```
xFrednet, line 108: Shouldn't this also set `needs_mut = true`?

clippy_utils/src/visitors.rs
```diff
@@ -436,3 +436,61 @@ pub fn for_each_value_source<'tcx, B>(
...
+            }
+            if path_to_local_id(e, self.local_id) {
+                self.res = (self.f)(e);
+            } else {
+                walk_expr(self, e);
+            }
+        }
+    }
+
+    if let Some(b) = get_enclosing_block(cx, local_id) {
```
xFrednet, line 482: NIT: Only getting the enclosing body can be a bit problematic, if the expression passed as `expr_id` is inside a nested block like a function. For the current use case and probably most others, this is fine, though. Could you add a comment to the function explaining this?

```rs
let mut v = vec![];
for i in 0..10 {
    v.push(i); // Called for this expression
    println!("{v:?}");
}

v.push(99); // Misses this expression
```

Otherwise, really nice utility! I was thinking of something like this, and might "steal" this implementation. Thank you!

xFrednet, line 482: NIT: Only getting the enclosing body can be a bit problematic, if the expression passed as `expr_id` is inside a nested block like a function. For the current use case and probably most others, this is fine, though. Could you add a comment to the function explaining this?

```rs
let mut v = vec![];
for i in 0..10 {
    v.push(i); // Called for this expression
    println!("{v:?}");
}

v.push(99); // Misses this expression
```

Otherwise, really nice utility! I was thinking of something like this, and might "steal" this implementation. Thank you!

tests/ui/vec_init_then_push.rs
```diff
@@ -44,3 +44,45 @@ pub fn no_lint() -> Vec<i32> {
...
     }
 }
+
+fn _from_iter(items: impl Iterator<Item = u32>) -> Vec<u32> {
+    let mut v = Vec::new();
+    v.push(0);
+    v.push(1);
+    v.extend(items);
+    v
+}
```
xFrednet, line 54: Could you add a second test like this, that reaches the threshold and triggers the lint? From my understanding, it should be as simple as:

```rs
fn _from_iter_trigger(items: impl Iterator<Item = u32>) -> Vec<u32> {
    let mut v = Vec::new();
    v.push(0);
    v.push(1);
    v.push(2);
    v.push(3);
    v.extend(items);
    v
}
```

Also, a test with the capacity would be good, like:

```rs
   let mut cap_err = Vec::with_capacity(4);
    cap_err.push(0);
    cap_err.push(1);
    cap_err.push(2);
```
That shouldn't lint either from my understanding

xFrednet, line 54: Could you add a second test like this, that reaches the threshold and triggers the lint? From my understanding, it should be as simple as:

```rs
fn _from_iter_trigger(items: impl Iterator<Item = u32>) -> Vec<u32> {
    let mut v = Vec::new();
    v.push(0);
    v.push(1);
    v.push(2);
    v.push(3);
    v.extend(items);
    v
}
```

xFrednet (commented):
Looks good to me overall, a few nits and that it should be ready.

Also thank you for your patience, I forgot this was on my todo list :sweat_smile:

clippy_lints/src/vec_init_then_push.rs
Jarcho: That case doesn't lint.

clippy_utils/src/visitors.rs
Jarcho: What's wrong with getting the enclosing block of the local? In you example it would get the block enclosing `let mut v = vec![];`.

Jarcho: On a side note `get_enclosing_block` doesn't work with trait functions. I'll add a fix for that in a bit.

tests/ui/vec_init_then_push.rs
Jarcho: The reason for avoiding the boundary case is to avoid having to change the test if we decide to nudge it a little. It doesn't really matter either way, just that the specific number isn't important.

clippy_lints/src/vec_init_then_push.rs
xFrednet: But it could, couldn't it? The code later checks:

```rs
if res == ControlFlow::Break(true) && self.found <= required_pushes_before_extension {
    return;
}
```

Which could still be false if `self.found` is > 4. Or should that be an or check in the if statement?

clippy_utils/src/visitors.rs
xFrednet: > What's wrong with getting the enclosing block of the local?

Nothing, it's just not clear, that this will always be the block of local, it could also be of some random expression. Using the outer block of the local would be ideal IMO :upside_down_face:

clippy_lints/src/vec_init_then_push.rs
xFrednet: But it could, couldn't it? The code later checks:

```rs
if res == ControlFlow::Break(true) && self.found <= required_pushes_before_extension {
    return;
}
```

Which could still be false if `self.found` is > 4. Or should that be an or check in the if statement?

clippy_utils/src/visitors.rs
xFrednet: Ohhh, I misread the code, sorry, this is fine. Thank you!

tests/ui/vec_init_then_push.rs
xFrednet: > The reason for avoiding the boundary case is to avoid having to change the test if we decide to nudge it a little.

Okay, then could you add one with some more space, it doesn't have to be at the boundary, but I don't see one for this case :upside_down_face:

clippy_lints/src/vec_init_then_push.rs
Jarcho: Ignore my previous comment. I don't know what I was looking.

It will already be set by that point, either by explicitly taking a mutable reference, or from an adjustment taking a mutable reference.

tests/ui/vec_init_then_push.rs
Jarcho: The last test covers it with a much larger number.

xFrednet: The last test covers an if statement. The lint already took if statements into account, that's why I didn't consider this a test to this case. Since you want to add a note about the other comment, could you add a test with `extend` or some other method?

xFrednet:
@Jarcho, this is a ping from triage. Could you address the review comments? (I believe the only real change was a comment that should be added :upside_down_face: )

@rustbot author

PR #8700, opened 2022-04-15
Fix needless_match false positive for if-let when the else block doesn't match to given expr

### LINKED ISSUE #8695: needless_match doesn't consider the `else` of `if let`
### Summary

The `needless_match` lint produces a nonsensical suggestion in some cases. It seems like it fires entirely based on the `if`-clause of an `if let`, ignoring the `else` branch's value entirely. This is a regression, most likely from #8549 (but I haven't confirmed that).

### Lint Name

needless_match

### Reproducer

[I tried this code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=db9096fe9508b1f92d94b985ec04996a):

```rust
pub fn bad() -> Option<i32> {
    if let Some(existing) = None {
        Some(existing)
    } else {
        Some(123)
    }
}
```

I saw this happen:

```
warning: this if-let expression is unnecessary
 [--> src/lib.rs:2:5
](https://play.rust-lang.org/#)  |
2 | /     if let Some(existing) = None {
3 | |         Some(existing)
4 | |     } else {
5 | |         Some(123)
6 | |     }
  | |_____^ help: replace it with: `None`
  |
  = note: `#[warn(clippy::needless_match)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_match
```

I expected to see no lint error.

### Version

```text
rustc 1.62.0-nightly (1f7fb6413 2022-04-10)
binary: rustc
commit-hash: 1f7fb6413d6d6c0c929b223e478e44c3db991b03
commit-date: 2022-04-10
release: 1.62.0-nightly
LLVM version: 14.0.0
```

### Additional Labels

_No response_

youknowone:
fix #8695

changelog: Fixed ``[`needless_match`]`` false positive when else block expression differs.

### DIFF
--- a/clippy_lints/src/matches/needless_match.rs
+++ b/clippy_lints/src/matches/needless_match.rs
@@ -99,7 +99,7 @@ fn check_if_let(cx: &LateContext<'_>, if_let: &higher::IfLet<'_>) -> bool {
                 if let ExprKind::Path(ref qpath) = ret.kind {
                     return is_lang_ctor(cx, qpath, OptionNone) || eq_expr_value(cx, if_let.let_expr, ret);
                 }
-                return true;
+                return false;
             }
             return eq_expr_value(cx, if_let.let_expr, ret);
         }
--- a/tests/ui/needless_match.rs
+++ b/tests/ui/needless_match.rs
@@ -103,6 +103,18 @@ fn if_let_option() {
     } else {
         None
     };
+
+    // Don't trigger
+    let _ = if let Some(a) = Some(1) { Some(a) } else { Some(2) };
+}
+
+fn if_let_option_result() -> Result<(), ()> {
+    fn f(x: i32) -> Result<Option<i32>, ()> {
+        Ok(Some(x))
+    }
+    // Don't trigger
+    let _ = if let Some(v) = f(1)? { Some(v) } else { f(2)? };
+    Ok(())
 }
 
 fn if_let_result() {
--- a/tests/ui/needless_match.fixed
+++ b/tests/ui/needless_match.fixed
@@ -80,6 +80,18 @@ fn if_let_option() {
     } else {
         None
     };
+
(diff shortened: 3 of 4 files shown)


clippy_lints/src/matches/needless_match.rs
```diff
@@ -99,7 +99,7 @@ fn check_if_let(cx: &LateContext<'_>, if_let: &higher::IfLet<'_>) -> bool {
                 if let ExprKind::Path(ref qpath) = ret.kind {
                     return is_lang_ctor(cx, qpath, OptionNone) || eq_expr_value(cx, if_let.let_expr, ret);
                 }
-                return true;
+                return false;
```
J-ZhengLi, line 102: lol, I didn't realize it was such small mistake here 🤣 
thank you~

Alexendoo (approved):
LGTM @xFrednet

tests/ui/needless_match.rs
```diff
@@ -103,6 +103,23 @@ fn if_let_option() {
     } else {
         None
     };
+
+    // Don't trigger
+    let _ = if let Some(a) = Some(1) {
+        do_something();
```
Alexendoo, line 109: Though actually one small nit, the `do_something` should go as this statement doesn't trigger the FP

```suggestion
```

youknowone: @Alexendoo thank you! I fixed it

xFrednet:
This version looks good to me, I'm taking the previous approval from @Alexendoo also as an okay. Thank you very much for the contribution and quick update @youknowone. :+1: 

Also thank you to @Alexendoo for taking over the review! :tada: 

@bors r+

youknowone:
Thank you for everyone. I am excited to make my first contribution to clippy, finally. I want to bring more patches in near future.

Issue #8674, opened 2022-04-10
Remove duplicate trait bound or where clause


aldhsu:
### What it does

This lint warns about repetitions of the same trait in trait bounds or where clauses.

### Lint Name

repetition_where_clause_or_trait_bound

### Category

pedantic

### Advantage

Leaving duplicate trait bounds is less readable.

### Drawbacks

None

### Example

```rust
pub fn foo<T: Copy + Copy>(t: T) {}
```
Could be written as:

```rust
pub fn foo<T: Copy>(t: T) {}
```

OR

```rust
pub fn foo<T>(t: T) where T: Copy + Copy {}
```
Could be written as:

```rust
pub fn foo<T>(t: T) where T: Copy {}
```

aldhsu:
I saw there is an existing lint that compares where clauses and trait bounds for duplicates. This could be an extension of that lint.

aldhsu:
Closed by #8703.

PR #8703, opened 2022-04-15
Add `repeated_where_clause_or_trait_bound` lint

aldhsu:
I thought I would try and scratch my own itch for #8674.

1. Is comparing the `Res` the correct way for ensuring we have the same trait?
2. Is there a way to get the spans for the bounds and clauses for suggestions?
I tried to use `GenericParam::bounds_span_for_suggestions` but it only gave me an empty span at the end of the spans.
I tried `WhereClause::span_for_predicates_or_empty_place` and it included the comma.
3. Is there a simpler way to get the trait names? I have used the spans of the traits because I didn't see a way to get it off the `Res` or `Def`.

changelog: Add ``[`repeated_where_clause_or_trait_bound`]`` lint.

### DIFF
--- a/clippy_lints/src/trait_bounds.rs
+++ b/clippy_lints/src/trait_bounds.rs
@@ -1,5 +1,5 @@
-use clippy_utils::diagnostics::span_lint_and_help;
-use clippy_utils::source::{snippet, snippet_with_applicability};
+use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_sugg};
+use clippy_utils::source::{snippet, snippet_opt, snippet_with_applicability};
 use clippy_utils::{SpanlessEq, SpanlessHash};
 use core::hash::{Hash, Hasher};
 use if_chain::if_chain;
@@ -9,8 +9,8 @@ use rustc_data_structures::unhash::UnhashMap;
 use rustc_errors::Applicability;
 use rustc_hir::def::Res;
 use rustc_hir::{
-    GenericBound, Generics, Item, ItemKind, Node, Path, PathSegment, PredicateOrigin, QPath, TraitBoundModifier,
-    TraitItem, Ty, TyKind, WherePredicate,
+    GenericArg, GenericBound, Generics, Item, ItemKind, Node, Path, PathSegment, PredicateOrigin, QPath,
+    TraitBoundModifier, TraitItem, TraitRef, Ty, TyKind, WherePredicate,
 };
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
@@ -36,7 +36,7 @@ declare_clippy_lint! {
     #[clippy::version = "1.38.0"]
     pub TYPE_REPETITION_IN_BOUNDS,
     nursery,
-    "Types are repeated unnecessary in trait bounds use `+` instead of using `T: _, T: _`"
+    "types are repeated unnecessary in trait bounds use `+` instead of using `T: _, T: _`"
 }
 
 declare_clippy_lint! {
@@ -63,10 +63,26 @@ declare_clippy_lint! {
     ///
     /// fn func<T>(arg: T) where T: Clone + Default {}
     /// ```
+    ///
+    /// ```rust
+    /// fn foo<T: Default + Default>(bar: T) {}
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn foo<T: Default>(bar: T) {}
+    /// ```
+    ///
+    /// ```rust
+    /// fn foo<T>(bar: T) where T: Default + Default {}
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn foo<T>(bar: T) where T: Default {}
+    /// ```
     #[clippy::version = "1.47.0"]
     pub TRAIT_DUPLICATION_IN_BOUNDS,
     nursery,
-    "Check if the same trait bounds are specified twice during a function declaration"
+    "check if the same trait bounds are specified more than once during a generic declaration"
 }
 
 #[derive(Copy, Clone)]
@@ -87,6 +103,19 @@ impl<'tcx> LateLintPass<'tcx> for TraitBounds {
     fn check_generics(&mut self, cx: &LateContext<'tcx>, gen: &'tcx Generics<'_>) {
         self.check_type_repetition(cx, gen);
         check_trait_bound_duplication(cx, gen);
+        check_bounds_or_where_duplication(cx, gen);
+    }
+
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
+        // special handling for self trait bounds as these are not considered generics
+        // ie. trait Foo: Display {}
+        if let Item {
+            kind: ItemKind::Trait(_, _, _, bounds, ..),
+            ..
+        } = item
+        {
+            rollup_traits(cx, bounds, "these bounds contain repeated elements");
+        }
     }
 
     fn check_trait_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx TraitItem<'tcx>) {
@@ -241,6 +270,26 @@ fn check_trait_bound_duplication(cx: &LateContext<'_>, gen: &'_ Generics<'_>) {
     }
 }
 
+#[derive(PartialEq, Eq, Hash, Debug)]
+struct ComparableTraitRef(Res, Vec<Res>);
+
+fn check_bounds_or_where_duplication(cx: &LateContext<'_>, gen: &'_ Generics<'_>) {
+    if gen.span.from_expansion() {
+        return;
+    }
+
+    for predicate in gen.predicates {
+        if let WherePredicate::BoundPredicate(ref bound_predicate) = predicate {
+            let msg = if predicate.in_where_clause() {
+                "these where clauses contain repeated elements"
+            } else {
+                "these bounds contain repeated elements"
+            };
+            rollup_traits(cx, bound_predicate.bounds, msg);
+        }
+    }
+}
+
 fn get_trait_info_from_bound<'a>(bound: &'a GenericBound<'_>) -> Option<(Res, &'a [PathSegment<'a>], Span)> {
     if let GenericBound::Trait(t, tbm) = bound {
         let trait_path = t.trait_ref.path;
@@ -255,5 +304,73 @@ fn get_trait_info_from_bound<'a>(bound: &'a GenericBound<'_>) -> Option<(Res, &'
         Some((trait_path.res, trait_path.segments, trait_span))
     } else {
         None
+    }
+}
+
+// FIXME: ComparableTraitRef does not support nested bounds needed for associated_type_bounds
+fn into_comparable_trait_ref(trait_ref: &TraitRef<'_>) -> ComparableTraitRef {
+    ComparableTraitRef(
+        trait_ref.path.res,
+        trait_ref
+            .path
+            .segments
+            .iter()
+            .filter_map(|segment| {
+                // get trait bound type arguments
+                Some(segment.args?.args.iter().filter_map(|arg| {
+                    if_chain! {
+                        if let GenericArg::Type(ty) = arg;
+                        if let TyKind::Path(QPath::Resolved(_, path)) = ty.kind;
+                        then { return Some(path.res) }
+                    }
+                    None
+                }))
+            })
+            .flatten()
+            .collect(),
+    )
+}
+
+fn rollup_traits(cx: &LateContext<'_>, bounds: &[GenericBound<'_>], msg: &str) {
+    let mut map = FxHashMap::default();
+    let mut repeated_res = false;
+
+    let only_comparable_trait_refs = |bound: &GenericBound<'_>| {
+        if let GenericBound::Trait(t, _) = bound {
+            Some((into_comparable_trait_ref(&t.trait_ref), t.span))
+        } else {
+            None
+        }
+    };
+
+    for bound in bounds.iter().filter_map(only_comparable_trait_refs) {
+        let (comparable_bound, span_direct) = bound;
+        if map.insert(comparable_bound, span_direct).is_some() {
+            repeated_res = true;
+        }
+    }
+
+    if_chain! {
+        if repeated_res;
+        if let [first_trait, .., last_trait] = bounds;
+        then {
+            let all_trait_span = first_trait.span().to(last_trait.span());
+
+            let mut traits = map.values()
+                .filter_map(|span| snippet_opt(cx, *span))
+                .collect::<Vec<_>>();
+            traits.sort_unstable();
+            let traits = traits.join(" + ");
+
+            span_lint_and_sugg(
+                cx,
+                TRAIT_DUPLICATION_IN_BOUNDS,
+                all_trait_span,
+                msg,
+                "try",
+                traits,
+                Applicability::MachineApplicable
+            );
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


aldhsu:
Good catch @giraffate! This also uncovers a bug in the existing TRAIT_DUPLICATION_IN_BOUNDS lint.

I'll have a think about it because it is a recursive problem. Do you have any suggestions for doing an easy comparison?

flip1995:
You should rebase your branch on top of the current master. Some things changed with how generics are lowered in rustc just a few days ago. You might have to revisit some parts of the code.

aldhsu:
Thanks for the heads up @flip1995. I have pushed a fixup that addresses the changes.

I left it unsquashed so how the generic lowering works now could be discussed.

I wasn't able to fix an issue with supporting the unstable associated_type_bounds feature.
Could you help me understand how the new lowering works with nested trait bounds? ie.
```rust
#![feature(associated_type_bounds)]

fn flatten_twice_good<T>(iter: T) -> Flatten<Flatten<T>>
where
    T: Iterator<Item: IntoIterator<Item: IntoIterator>>,
{
    iter.flatten().flatten()
}
```
https://github.com/rust-lang/rust-clippy/pull/8703/commits/c9c31f3be6734978e13bb5f5827c2818ed46bc77 handled this because bounds could contain their nested bounds. I haven't been able to follow how WhereBoundPredicate can nest WhereBoundPredicates yet.

flip1995:
I wouldn't worry too much about supporting this feature with this lint. I'd suggest to just leave a `FIXME` comment that this should be addressed once this feature gets stabilized, which might take quite some time.

aldhsu:
Squashed and added the fixme.

There is one other outstanding issue I haven't been able to figure out.

This case should be detected.
https://github.com/rust-lang/rust-clippy/blob/6040fa64323ebfcd906e6f039b35b776a07a5486/tests/ui/repeated_where_clause_or_trait_bound.rs#L40-L42

Doing some debugging on the Generic, no predicates exist. This only happens for the self trait bound and not the where clause. I haven't dug any further than that.

flip1995:
For your example you should find those bounds in the [`ItemKind::Trait`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/enum.ItemKind.html#variant.Trait). It's the tuple field with the `GenericBounds` type.

aldhsu:
Thanks for the tip @flip1995. I got access to an `ItemKind::Trait` by implementing `check_item` of the LateLintPass for Traitbounds. Is this what you meant?

flip1995:
Thanks! I'd like to delegate the remaining review of this lint to @dswij. @dswij if you already have too many assigned PRs feel free to bounce it back to me though.

dswij (commented):
Many thanks for the PR! Changes overall looks great to me. It's a really nice addition for sure. 

Just left one nit that's not really important to the change.

---

> Is comparing the Res the correct way for ensuring we have the same trait?

Comparing Res within a type bound should be fine.

> Is there a simpler way to get the trait names? I have used the spans of the traits because I didn't see a way to get it off the Res or Def.

Well, we could've used `ident`. But for the suggestions here, using `snippet_opt` with the spans would be preferred

clippy_lints/src/trait_bounds.rs
```diff
@@ -250,10 +293,96 @@ fn check_trait_bound_duplication(cx: &LateContext<'_>, gen: &'_ Generics<'_>) {
...
+        let (comparable_bound, span_direct) = bound;
+        if map.insert(comparable_bound, span_direct).is_some() {
+            repeated_spans = true;
+        }
+    }
+
+    if_chain! {
+        if repeated_spans;
+        if let Some(first_trait) = bounds.get(0);
+        if let Some(last_trait) = bounds.iter().last();
```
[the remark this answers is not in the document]
aldhsu: Very cool.

tests/ui/repeated_where_clause_or_trait_bound.rs
```diff
@@ -0,0 +1,105 @@
+// run-rustfix
+//
+#![allow(unused)]
+#![deny(clippy::repeated_where_clause_or_trait_bound)]
+
+fn bad_foo<T: Clone + Clone + Clone + Copy, U: Clone + Copy>(arg0: T, argo1: U) {
```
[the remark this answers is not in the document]
aldhsu: Added that test case in https://github.com/rust-lang/rust-clippy/pull/8703/commits/e6b34d7b3dfcb15608754ce5b55054bb09b32eab and one for a user specified with a similar name.

clippy_lints/src/trait_bounds.rs
```diff
@@ -250,10 +293,96 @@ fn check_trait_bound_duplication(cx: &LateContext<'_>, gen: &'_ Generics<'_>) {
...
+                }))
+            })
+            .flatten()
+            .collect(),
+    )
+}
+
+fn rollup_traits(cx: &LateContext<'_>, bounds: &[GenericBound<'_>], msg: &str) {
+    let mut map = FxHashMap::default();
+    let mut repeated_spans = false;
```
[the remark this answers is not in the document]
aldhsu: Renamed to repeated_res.

aldhsu:
Thanks for the review @dswij. I have actioned it.

dswij (approved):
Thanks for addressing the comments! It looks good to me.

I'm not sure whether this lint should be in `nursery` group though, I don't see this having lots of false warnings either. Need some opinion on this, what do you guys think? @xFrednet @flip1995

xFrednet:
Thank you for the review @flip1995 offered to do the final review (probably after their vacation :))

> I'm not sure whether this lint should be in nursery group though, I don't see this having lots of false warnings either. Need some opinion on this, what do you guys think?

I think the final group should be `pedantic`. `TRAIT_DUPLICATION_IN_BOUNDS` which is similar, is currently in the nursery group due to some braking changes in rustc. I believe that they're fixed now, but I'm not totally sure. If they are fixed, we should change the lint group. Until then, I would keep this in the same group as `TRAIT_DUPLICATION_IN_BOUNDS`. @flip1995 is more up-to-date on this AFAIK :upside_down_face:

aldhsu:
@dswij I put it in the nursery group because the other trait bound ones were put in there. I didn't have the confidence to say that the new lint was ready when the original lints it was based off were moved to nursery 😅. Original group was `pedantic` as @xFrednet said. 

Offtopic: What is the process of growing out of nursery? I'm going to pick up https://github.com/rust-lang/rust-clippy/issues/8757

xFrednet:
> Offtopic: What is the process of growing out of nursery? I'm going to pick up #8757

It depends on the reason why the lint was moved to the nursery group. The general rule is, that lints can be moved to their respective group once all issues have been resolved.

In this case I remember somebody working on fixing the issues. I'm just not sure if they've finished their work. It could be, that they've been fixed on the rust side and will be enabled again during the weekend. I'll try to keep an eye on it :)

clippy_lints/src/trait_bounds.rs
```diff
@@ -70,6 +70,35 @@ declare_clippy_lint! {
...
+    ///
+    /// ```rust
+    /// fn foo<T>(bar: T) where T: Default + Default {}
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn foo<T>(bar: T) where T: Default {}
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub REPEATED_WHERE_CLAUSE_OR_TRAIT_BOUND,
```
flip1995, line 97: Could this lint be included in the `TRAIT_DUPLICATION_IN_BOUNDS`? When thinking about a good lint name for this lint I thought `TRAIT_DUPLICATION_IN_BOUNDS` would actually the best name for this lint.

If this lint should be kept its own separate lint, the lint name should be

```suggestion
    pub REPEATED_WHERE_CLAUSES_OR_TRAIT_BOUNDS,
```

clippy_lints/src/trait_bounds.rs
```diff
@@ -70,6 +70,35 @@ declare_clippy_lint! {
...
+    nursery,
```
flip1995, line 98: I fixed the issue in rustc and moved the other lints back to pedantic: https://github.com/rust-lang/rust/pull/96770

So regardless whether this lint is merged with the other lint or kept on its own, it should be in pedantic.

clippy_lints/src/trait_bounds.rs
```diff
@@ -70,6 +70,35 @@ declare_clippy_lint! {
...
+    "Traits are repeated within trait bounds or where clause"
```
flip1995, line 99: This should be lower case, just like lint messages. I also noticed that this is also wrong for the other lints in this file. If you want you can also change this for those.
```suggestion
    "traits are repeated within trait bounds or where clause"
```

flip1995 (commented):
Thanks for the review @dswij and thanks for addressing it @aldhsu!

I left some comments. Not sure if this requires a new lint.

clippy_lints/src/trait_bounds.rs
aldhsu: Fixed in https://github.com/rust-lang/rust-clippy/pull/8703/commits/6e2ec70ce3a8ed0f15e33bafdcee400e22bf725c.

clippy_lints/src/trait_bounds.rs
aldhsu: Moved it back to pedantic.

clippy_lints/src/trait_bounds.rs
aldhsu: I think it is possible to merge the two. 

I have a proposal since merging `trait_duplication_in_bounds` and `repeated_where_clause_or_trait_bound` in could be too much work and might mean the PR stalls. 

I propose we:
1. Merge this PR when it is acceptable without adding any more functionality.
2. Fix https://github.com/rust-lang/rust-clippy/issues/8757 and merge PR. I've been working on a fix that is based on some work in this PR to make comparable trait refs. This will also make the two more similar and easier to refactor.
3. I'll work on a refactor PR to merge `trait_duplication_in_bounds` and `repeated_where_clause_or_trait_bound`.

aldhsu:
Thanks for the review @flip1995!

clippy_lints/src/trait_bounds.rs
aldhsu: @flip1995 what do you think of this proposal?

clippy_lints/src/trait_bounds.rs
```diff
@@ -254,10 +296,95 @@ fn check_trait_bound_duplication(cx: &LateContext<'_>, gen: &'_ Generics<'_>) {
...
+
+            let mut traits = map.values()
+                .filter_map(|span| snippet_opt(cx, *span))
+                .collect::<Vec<_>>();
+            traits.sort_unstable();
+            let traits = traits.join(" + ");
+
+            span_lint_and_sugg(
+                cx,
+                REPEATED_WHERE_CLAUSES_OR_TRAIT_BOUNDS,
```
flip1995, line 381: Merging the lints would just mean to change this line and adapt the documentation of the other lint a bit. I don't really want to merge a new lint, that will be renamed later, because if we mistime things, the temporary lint name will get into stable and we have to go through the renaming procedure or backport things.

flip1995 (commented):
So sorry, I completely forgot about this PR... I would do the renaming now, it shouldn't be too much work. If you lost the motivation to work on this after that long time, I'll address this myself in order not to lose this awesome work.

clippy_lints/src/trait_bounds.rs
aldhsu: @flip1995 ah got it. Of course the lint would be just under the same name! I was thinking it had to be done in the same pass.

aldhsu:
@flip1995 I can address it. I had it in my head that "merging" the lint meant doing it in the same pass. I should be able to get it done this weekend.

Thanks for calling it awesome!

clippy_lints/src/trait_bounds.rs
aldhsu: Merging under the existing lint name raises some extra noise with UI test errors since the original lint "catches" some extra errors (#8757). Since this is actually a bug, I have left a comment that the errors generated in `stderr` should be removed once #8757 is fixed. 

With regards to fixing, I've just turned off rustfix since I can't get the fixed file to compile without fixing the original lint.

Once this PR is merged I can work on reducing the noise in the errors so that it's not reporting the same errors twice slightly differently. This sounds ok to me since the lint is back in nursery.

tests/ui/trait_duplication_in_bounds.rs
```diff
@@ -98,4 +99,114 @@ trait FooIter: Iterator<Item = Foo> {
...
+    }
+
+    fn no_error_separate_arg_bounds(program: impl AsRef<()>, dir: impl AsRef<()>, args: &[impl AsRef<()>]) {}
+
+    trait GenericTrait<T> {}
+
+    // This should not warn but currently does see #8757
+    fn good_generic<T: GenericTrait<u64> + GenericTrait<u32>>(arg0: T) {
+        unimplemented!();
+    }
```
flip1995, line 197: If this is the only test case that trips up rustfix, I'd suggest to move it into its own file named something like `tests/ui/trait_duplication_in_bounds_8757.rs`. That way, you can keep rustfix enabled and still have a regression test for the issue.

Or are there other cases that would break rustfix?

aldhsu: Yes, there are other cases that break rustfix.

The existing lint of `trait_duplication_in_bounds` raises a lint for pretty much all the new examples added. I believe the original intention of the lint is supposed to only be for traits appearing in both trait bounds and where clauses as that seems to be the only kinds of tests. I think it accidentally picks up duplicates **within** trait bounds and where clauses. This is evident because the errors produced for these cases are misleading.

Example:
```rust
    fn bad_foo<T: Clone + Clone + Clone + Copy, U: Clone + Copy>(arg0: T, argo1: U) {
        unimplemented!();
    }

```

```
error: this trait bound is already specified in the where clause
  --> $DIR/trait_duplication_in_bounds.rs:103:19
   |
LL |     fn bad_foo<T: Clone + Clone + Clone + Copy, U: Clone + Copy>(arg0: T, argo1: U) {
   |                   ^^^^^
   |
   = help: consider removing this trait bound

```
`this trait bound is already specified in the where clause`
It's error refers to a where clause but there aren't any. 

I can fix this issue along with #8757 .
At that time I can split the fixables and unfixables like other I see done with the other test files.

flip1995: Ah ok, that's interesting. It seems that the old lint already triggers on the cases, the newly added lint code should check for, just with a worse error message + suggestion. I think this is even worse than #8757. Since this lint is in `nursery` I'd be ok merging this. But I would open an issue about it, even when you start working on resolving this right away.

clippy_lints/src/trait_bounds.rs
```diff
@@ -257,3 +306,68 @@ fn get_trait_info_from_bound<'a>(bound: &'a GenericBound<'_>) -> Option<(Res, &'
...
+fn rollup_traits(cx: &LateContext<'_>, bounds: &[GenericBound<'_>], msg: &str) {
+    let mut map = FxHashMap::default();
+    let mut repeated_res = false;
+    for bound in bounds.iter().filter_map(|bound| {
+        if let GenericBound::Trait(t, _) = bound {
+            Some((into_comparable_trait_ref(&t.trait_ref), t.span))
+        } else {
+            None
+        }
+    }) {
```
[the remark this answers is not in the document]
aldhsu: Resolved in https://github.com/rust-lang/rust-clippy/pull/8703/commits/a0c2da924ea381c221bcbdd06436a1c9c96936e1

tests/ui/trait_duplication_in_bounds.rs
aldhsu: Issue here #9151

aldhsu:
@flip1995 squashed. Thanks for the review!

flip1995:
Thanks! Sorry again for forgetting about this PR

@bors r+

PR #8706, opened 2022-04-15
Fix formatting of `cast_abs_to_unsigned` docs

lupd:
The "use instead" section of the example was not being formatted as Rust code, and the "configuration" documentation was being formatted as Rust code.

changelog: `[cast_abs_to_unsigned]` Fix example/configuration formatting


xFrednet:
Very nice catch, thank you for the change :upside_down_face: 

@bors r+

PR #8707, opened 2022-04-15
Add `await_holding_invalid_type` lint

lilymara-onesignal:
changelog: [`await_holding_invalid_type`]

This lint allows users to create a denylist of types which are not allowed to be
held across await points. This is essentially a re-implementation of the
language-level [`must_not_suspend`
lint](https://github.com/rust-lang/rust/issues/83310). That lint has a lot of
work still to be done before it will reach Rust stable, and in the meantime
there are a lot of types which can trip up developers if they are used
improperly.

I originally implemented this specifically for `tracing::span::Entered`, until I discovered #8434 and read the commentary on that PR. Given this implementation is fully user configurable, doesn't tie clippy to any one particular crate, and introduces no additional dependencies, it seems more appropriate.


clippy_lints/src/await_holding_invalid.rs
```diff
@@ -146,46 +213,68 @@ impl LateLintPass<'_> for AwaitHolding {
...
+    let (type_name, reason) = match disallowed {
+        DisallowedType::Simple(path) => (path, &None),
+        DisallowedType::WithReason { path, reason } => (path, reason),
+    };
+
+    span_lint_and_then(
+        cx,
+        AWAIT_HOLDING_INVALID_TYPE,
+        span,
+        &format!("`{type_name}` may not be held across an `await` point according to config",),
```
llogiq, line 269:
```suggestion
        &format!("`{type_name}` may not be held across an `await` point per `clippy.toml`",),
```

clippy_lints/src/await_holding_invalid.rs
```diff
@@ -146,46 +213,68 @@ impl LateLintPass<'_> for AwaitHolding {
...
+        |diag| {
+            if let Some(reason) = reason {
+                diag.note(format!("{reason} (according to clippy.toml)"));
```
llogiq, line 272:
```suggestion
                diag.note(reason.clone());
```

clippy_lints/src/lib.rs
```diff
@@ -515,7 +515,12 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     }
 
     store.register_late_pass(|| Box::new(utils::author::Author));
-    store.register_late_pass(|| Box::new(await_holding_invalid::AwaitHolding));
+    let await_holding_invalid_types = conf.await_holding_invalid_types.clone();
+    store.register_late_pass(move || {
+        Box::new(await_holding_invalid::AwaitHolding::new(
+            await_holding_invalid_types.clone(),
```
llogiq, line 521:
```suggestion
            await_holding_invalid_types,
```
No need for double clone, right?

llogiq (approved):
This looks good to me, I have but a few small suggestions.

clippy_lints/src/lib.rs
lilymara-onesignal: I tried this initially but borrowck doesn't like it

```
error[E0507]: cannot move out of `await_holding_invalid_types`, a captured variable in an `Fn` closure
   --> clippy_lints/src/lib.rs:519:88
    |
518 |     let await_holding_invalid_types = conf.await_holding_invalid_types.clone();
    |         --------------------------- captured outer variable
519 |     store.register_late_pass(move || Box::new(await_holding_invalid::AwaitHolding::new(await_holding_invalid_types)));
    |                              ----------------------------------------------------------^^^^^^^^^^^^^^^^^^^^^^^^^^^--
    |                              |                                                         |
    |                              |                                                         move occurs because `await_holding_invalid_types` has type `std::vec::Vec<DisallowedType>`, which does not implement the `Copy` trait
    |                              captured by this `Fn` closure
```

clippy_lints/src/await_holding_invalid.rs
```diff
@@ -127,17 +130,78 @@ declare_clippy_lint! {
...
+    /// from a memory access perspective but will cause bugs at runtime if they
+    /// are held in such a way.
+    ///
+    /// ### Known problems
+    ///
+    /// ### Example
+    ///
+    /// ```toml
+    /// await-holding-invalid-types = [
+    ///   "CustomLockType",
```
llogiq, line 149: I'd like an additional entry here that shows the "reason" variant of the config. This is the only place users will read to know how to configure this.

llogiq (commented):
One final request re the lint docs, otherwise this is ready to merge.

llogiq:
Thank you! ❤️

@bors r+

PR #8716, opened 2022-04-18
Less authoritative stable_sort_primitive message

### LINKED ISSUE #8241: Make stable_sort_primitive message less authoritative
### Description

The current message for `stable_sort_primitive` says:
```
an unstable sort would perform faster without any observable difference for this data type
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#stable_sort_primitive
```
Although it is a reasonable guideline and typically true for random data, it is not always true. For example, the append benchmark in https://gist.github.com/jongiddy/b29ce7d2251775dcf24c595bc245891b demonstrates that stable sort can be significantly faster for a scenario where a sorted vector is extended with new data and resorted.

```
running 6 tests
test almost_sorted_stable   ... bench:      12,290 ns/iter (+/- 97)
test almost_sorted_unstable ... bench:       3,392 ns/iter (+/- 74)
test append_stable          ... bench:       2,545 ns/iter (+/- 19)
test append_unstable        ... bench:       8,450 ns/iter (+/- 236)
test random_stable          ... bench:      16,803 ns/iter (+/- 396)
test random_unstable        ... bench:       6,149 ns/iter (+/- 156)
```
Because of this, the initial message should be less definite:
```
an unstable sort typically performs faster without any observable difference for this data type
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#stable_sort_primitive
```
and the linked message should also be less definite, and suggest benchmarking specific situations.

### Version

_No response_

### Additional Labels

_No response_

binggh:
fixes #8241 

Hey all - first contribution here so I'm deciding to start with something small.

Updated the linked message to be less authoritative as well as moved the lint grouping from `perf` to `pedantic` as suggested by @camsteffen under the issue.

changelog: [`stable_sort_primitive`]: emit less authoritative message and move to `pedantic`

### DIFF
--- a/clippy_lints/src/lib.register_all.rs
+++ b/clippy_lints/src/lib.register_all.rs
@@ -278,7 +278,6 @@ store.register_group(true, "clippy::all", Some("clippy_all"), vec![
     LintId::of(single_component_path_imports::SINGLE_COMPONENT_PATH_IMPORTS),
     LintId::of(size_of_in_element_count::SIZE_OF_IN_ELEMENT_COUNT),
     LintId::of(slow_vector_initialization::SLOW_VECTOR_INITIALIZATION),
-    LintId::of(stable_sort_primitive::STABLE_SORT_PRIMITIVE),
     LintId::of(strings::STRING_FROM_UTF8_AS_BYTES),
     LintId::of(strlen_on_c_strings::STRLEN_ON_C_STRINGS),
     LintId::of(suspicious_trait_impl::SUSPICIOUS_ARITHMETIC_IMPL),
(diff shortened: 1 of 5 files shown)


giraffate (commented):
Can the `Known problem` section be added to the [document](https://github.com/rust-lang/rust-clippy/blob/7a73e09e35900703d2d58353f30358cfeb1cb4ea/clippy_lints/src/stable_sort_primitive.rs#L20)? And it would be better to add the comments and refer to the issue.

binggh:
Sure, my bad - missed that section.

giraffate:
@bors r+

Thanks!

PR #8717, opened 2022-04-18
`manual_split_once`: lint manual iteration of `SplitN`

Alexendoo:
changelog: `manual_split_once`: lint manual iteration of `SplitN`

Now lints:

```rust
let mut iter = "a.b.c".splitn(2, '.');
let first = iter.next().unwrap();
let second = iter.next().unwrap();

let mut iter = "a.b.c".splitn(2, '.');
let first = iter.next()?;
let second = iter.next()?;

let mut iter = "a.b.c".rsplitn(2, '.');
let first = iter.next().unwrap();
let second = iter.next().unwrap();

let mut iter = "a.b.c".rsplitn(2, '.');
let first = iter.next()?;
let second = iter.next()?;
```

It suggests (minus leftover whitespace):

```rust
let (first, second) = "a.b.c".split_once('.').unwrap();

let (first, second) = "a.b.c".split_once('.')?;

let (second, first) = "a.b.c".rsplit_once('.').unwrap();

let (second, first) = "a.b.c".rsplit_once('.')?;
```

Currently only lints if the statements are next to each other, as detecting the various kinds of shadowing was tricky, so the following won't lint

```rust
let mut iter = "a.b.c".splitn(2, '.');
let something_else = 1;
let first = iter.next()?;
let second = iter.next()?;
```

### DIFF
--- a/clippy_lints/src/methods/str_splitn.rs
+++ b/clippy_lints/src/methods/str_splitn.rs
@@ -1,14 +1,19 @@
 use clippy_utils::consts::{constant, Constant};
-use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::source::snippet_with_context;
-use clippy_utils::{is_diag_item_method, match_def_path, meets_msrv, msrvs, paths};
+use clippy_utils::usage::local_used_after_expr;
+use clippy_utils::visitors::expr_visitor;
+use clippy_utils::{is_diag_item_method, match_def_path, meets_msrv, msrvs, path_to_local_id, paths};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind, HirId, LangItem, Node, QPath};
+use rustc_hir::intravisit::Visitor;
+use rustc_hir::{
+    BindingAnnotation, Expr, ExprKind, HirId, LangItem, Local, MatchSource, Node, Pat, PatKind, QPath, Stmt, StmtKind,
+};
 use rustc_lint::LateContext;
 use rustc_middle::ty;
 use rustc_semver::RustcVersion;
-use rustc_span::{symbol::sym, Span, SyntaxContext};
+use rustc_span::{sym, Span, Symbol, SyntaxContext};
 
 use super::{MANUAL_SPLIT_ONCE, NEEDLESS_SPLITN};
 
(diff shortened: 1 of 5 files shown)


tests/ui/manual_split_once.rs
```diff
@@ -41,13 +41,107 @@ fn main() {
...
+    let other = 1;
+    let l = iter.next()?;
+    let r = iter.next()?;
+
+    let mut iter = "a.b.c".splitn(2, '.');
+    let mut mut_binding = iter.next()?;
+    let r = iter.next()?;
+
+    let mut iter = "a.b.c".splitn(2, '.');
+    let tuple = (iter.next()?, iter.next()?);
```
dswij, line 73: Can we also add the limitations to the lint's documentation?

Something like [this](https://github.com/rust-lang/rust-clippy/blob/cbdf17c884116f17b1efe8c6dd3c41eb45d1342c/clippy_lints/src/enum_variants.rs#L21-L25)

clippy_lints/src/methods/str_splitn.rs
```diff
@@ -107,16 +113,171 @@ fn check_manual_split_once(
...
+        && let mut stmts = enclosing_block
+            .stmts
+            .iter()
+            .skip_while(|stmt| stmt.hir_id != iter_stmt_id)
+            .skip(1)
+
+        && let first = indirect_usage(cx, stmts.next()?, iter_binding_id, ctxt)?
+        && let second = indirect_usage(cx, stmts.next()?, iter_binding_id, ctxt)?
+        && first.unwrap_kind == second.unwrap_kind
+        && first.name != second.name
```
dswij, line 146: I'm assuming this is so that

```rust
    let mut iter = "a.b.c".splitn(2, '.');
    let l = iter.next().unwrap();
    let l = iter.next().unwrap();
```

wouldn't be suggested to `let (l, l) = ...`?

I wonder if there's an easy way to cover this scenario to suggest `let (_, l) = ...` instead if the above happens

dswij (commented):
LGTM! Just a question and a suggestion to add the limitations to docs, otherwise the changes look great.

clippy_lints/src/methods/str_splitn.rs
Alexendoo: It is yeah, I think it's fine to not emit anything for that case, as there would be an unused warning for the first `l`

tests/ui/manual_split_once.rs
xFrednet: We can add that. I usually go by feel if a limitation has to be added to the description or not. Having it in the documentation is never wrong (If it's up-to-date).

Could you also add the limitation, that only `next()` calls directly following the `split()` call get linted?

xFrednet (commented):
This version already looks good to me. I like the suggestion to add a `limitations` section to the docs, and then we can merge this. :+1:

Thank you very much @dswij for the review!

tests/ui/manual_split_once.rs
Alexendoo: Added!

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update! :+1:

@bors r=dswij,xFrednet

Issue #8723, opened 2022-04-20
match_single_binding suggest wrong code


jo-so:
### Summary

The code suggestion of clippy doesn't work.

```text
warning: this match could be written as a `let` statement
  --> src/html_output.rs:59:19
   |
59 |               val = match val.split_at(idx) {
   |  ___________________^
60 | |                 (pre, suf) => {
61 | |                     pre.write_to(out)?;
62 | |                     suf
63 | |                 }
64 | |             };
   | |_____________^
   |
   = note: `#[warn(clippy::match_single_binding)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_single_binding
help: consider using `let` statement
   |
59 ~             val = let (pre, suf) = val.split_at(idx);
60 +             {
61 +                 pre.write_to(out)?;
62 +                 suf
63 ~             };
   |
```

### Reproducer

I tried this code:

```rust
fn main() {
    let (mut val, idx) = ("a b", 1);

    val = match val.split_at(idx) {
        (pre, suf) => {
            println!("{}", pre);
            suf
        }
    };

    let _ = val;
}
```

I expected to see this happen:
```text
59 ~             val = {
60 +                 let (pre, suf) = val.split_at(idx);
61 +                 pre.write_to(out)?;
62 +                 suf
63 ~             };
```

Instead, this happened:
```text
59 ~             val = let (pre, suf) = val.split_at(idx);
60 +             {
61 +                 pre.write_to(out)?;
62 +                 suf
63 ~             };
```

### Version

```text
rustc 1.62.0-nightly (311e2683e 2022-04-18)
binary: rustc
commit-hash: 311e2683e1bad87715b1558f7900e294d24ce491
commit-date: 2022-04-18
host: x86_64-unknown-linux-gnu
release: 1.62.0-nightly
LLVM version: 14.0.1
```

```
clippy 0.1.62 (311e268 2022-04-18)
```

PR #8726, opened 2022-04-20
Fix `match_single_binding` suggestion for assign expressions

Serial-ATA:
changelog: Fix suggestion for assign expressions in [`match_single_binding`]
closes: #8723


dswij (approved):
LGTM, thanks for the changes!

@xFrednet

tests/ui/match_single_binding.stderr
```diff
@@ -176,5 +176,25 @@ LL | |         _ => println!("Not an array index start"),
...
+LL | |     };
+   | |_____^
+   |
+help: consider removing the `match` expression
+   |
+LL ~     let (pre, suf) = val.split_at(idx);
+LL +     val = {
+LL +         println!("{}", pre);
+LL +         suf
+LL ~     };
```
xFrednet, line 196: Shouldn't this suggest the following instead:

```rs
    val = {
        let (pre, suf) = val.split_at(idx);
        println!("{}", pre);
        suf
    };
```

Having the `let <pattern>` before the linted assignment can potentially shadow values. It also extends the lifetime of the pattern values. Now we can access `pre` and `suf` after the assignment. Which we couldn't previously.

xFrednet (commented):
The refactoring already looks good to me. However, I believe the suggestion should be changed a bit, to keep the same semantics :upside_down_face:

tests/ui/match_single_binding.stderr
Serial-ATA: Should it? I know that's what the suggestion was in the issue, but that's how it behaved before I started this.

dswij: Yeah, I was under the impression that that's how the suggestion behaved before. But the suggestion will introduce a change in behavior, and that's certainly not what we want.

I think @xFrednet's suggestion makes sense

xFrednet: Okay, then we should change the suggestion. @Serial-ATA Could you make this change? It should be fairly simple, from what I'm guessing. If not, we can also create a new issue and merge this as it is. Your changes are already a solid improvement :upside_down_face:

Serial-ATA: Alright, I'll try it out 🙂

xFrednet:
Hey @Serial-ATA, what's the status on this PR? The last unresolved comment was an optional improvement. Do we want to get this merged and created an issue for this possible enhancement?

If so, you would just need to rebase, as all changes looked good to us :upside_down_face:

Serial-ATA:
Completely forgot about this :smile:. I tried for some time to make the change, but I couldn't get it to work reliably, unless I'm missing something obvious. I'll make an issue and come back to it soon enough if no one claims it.

xFrednet (approved):

PR #8727, opened 2022-04-21
Add `large_include_file` lint

### LINKED ISSUE #7005: Lint on large `std::{include_str!, include_bytes!};` calls.
### What it does

At Embark we've been doing some optimisations on the size of the final output of our binaries, and one thing that we thought there could be lint to help with this process. The lint will warn you when you have a have included a large enough file, that might significantly contribute to the size of the final binary.

### Categories (optional)

- Kind: pedantic

For example:
- It allows you to easily find locations in code that contribute heavily to your final binary size, allowing you to slim down your size by optimising or removing your included assets.

### Drawbacks

None.

### Unresolved Questions
- How big of a file should be linted against? Every project has different definition of what size is too large. This might be configurable in a world where there's a `clippy.toml` but assuming that isn't an option, are there good defaults we can use instead?
    - My initial thinking was that there could be two lints `clippy::large_include_files` and `clippy::very_large_include_files` which are the same lint but set to two different limits (e.g. >1MB and >10MB), so people can choose what's considered too much for their project.

Serial-ATA:
changelog: Add [`large_include_file`] lint
closes #7005


tests/ui-toml/large_include_file/clippy.toml
```diff
@@ -0,0 +1 @@
+max-include-file-size = 400
```
xFrednet, line 1: Could you add a line break here?
```suggestion
max-include-file-size = 400

```

tests/ui-toml/large_include_file/large_include_file.rs
```diff
@@ -0,0 +1,11 @@
+#![warn(clippy::large_include_file)]
+
+// Good
+const GOOD_INCLUDE_BYTES: &[u8; 354] = include_bytes!("large_include_file.rs");
+const GOOD_INCLUDE_STR: &str = include_str!("large_include_file.rs");
+
+// Bad
+const TOO_BIG_INCLUDE_BYTES: &[u8; 571] = include_bytes!("too_big.txt");
+const TOO_BIG_INCLUDE_STR: &str = include_str!("too_big.txt");
+
```
xFrednet, line 10: Since this is dealing with macros, could you add an explicit test for the allow attribute like this:

```rs
#[allow(clippy::large_include_file)]
const ALLOWED_TOO_BUG_INCLUDE: &str = include_str!("too_big.rs");
```

I'm 99.9% sure that this works without a problem :upside_down_face:

tests/ui-toml/large_include_file/too_big.txt
```diff
@@ -0,0 +1 @@
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ultricies mi quis hendrerit dolor. Sit amet tellus cras adipiscing enim eu turpis egestas. Faucibus nisl tincidunt eget nullam non nisi est. Ipsum faucibus vitae aliquet nec ullamcorper sit. Maecenas pharetra convallis posuere morbi leo urna. Purus in mollis nunc sed id semper. Sit amet tellus cras adipiscing enim eu turpis egestas. Mauris rhoncus aenean vel elit scelerisque mauris pellentesque. Massa vitae tortor condimentum lacinia quis vel.
```
xFrednet, line 1: Unrelated, but I wanted to share my thoughts ^^. You can resolve this :upside_down_face: 

Lorem ipsum, the blind text of all blind texts. I would have probably spent 20+ minutes just searching for a text that can be included. There are so many possibilities to choose from. We could even write a poem about Clippy, the brave paperclip. That's just the inner troll in me *Lorem ipsum* is a good choice IMO.

clippy_lints/src/large_include_file.rs
```diff
@@ -0,0 +1,77 @@
...
+                }
+
+                span_lint_and_help(
+                    cx,
+                    LARGE_INCLUDE_FILE,
+                    expr.span,
+                    "attempted to include a large file",
+                    None,
+                    &format!(
+                        "this lint is configured with a maximum size of {} bytes",
```
xFrednet, line 70: I really like that you mention the configuration here. I would suggest using `span_lint_and_note`, since help is intended for solutions or explanations, while this is more a general information.

Maybe a message like this would be more direct:

```suggestion
                        "the configuration allows a maximum size of {} bytes",
```

The current message is also fine with me :upside_down_face:

clippy_lints/src/large_include_file.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Checks for the inclusion of large files via `include_bytes!()`
+    /// and `include_str!()`
+    ///
+    /// ### Why is this bad?
+    /// Including large files can increase the size of the binary
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// let include = include_str!("very_large_file.txt");
+    /// ```
```
xFrednet, line 22: The description is still missing a solution or better way. Probably also do to the fact, that it's hard to give the correct suggestion. Therefore, I would just add a generic command like:

```suggestion
    /// ```
    /// Instead, you can load the file at runtime, to reduce the binary size.
```

xFrednet (commented):
This version already looks good to me. A few small NITs that should all be simple to fix :upside_down_face:

tests/ui-toml/large_include_file/too_big.txt
Serial-ATA: A Clippy poem would've been great. Wish I was creative enough for something like that :smile:

xFrednet:
LGTM, thank you for this lint!

@bors r+

lopopolo:
Hi folks, is this a lint that can be moved to restriction rather than pediatric? It seems similar in spirit to the restriction lints in that there is nothing inherently wrong with using include str and include bytes with large contents, but for certain cases you may wish to lint.

xFrednet:
Moving it to the `restriction` group sounds good to me. The group was actually one thing I was unsure about, but it slipped through the review :sweat_smile:

xFrednet:
You're right, I didn't look up what group it currently is in. :sweat_smile:

lopopolo:
Oh, it is in restriction. Sorry I didn't look at the code that closely. The linked ticket mentioned pedantic.

Thanks folks!

PR #8730, opened 2022-04-22
[FP] identity_op in front of if

### LINKED ISSUE #8724: [FP] identity_op in front of if
### Summary

The lint triggers, although the suggested code procudes a syntax error.

Problem is most likely, that the lint only views the if expression and not the whole one, e.g. it's a sum of (at least) three operands, where the first one is the identity (`0` in this case). But the expression is invalid without that, because that's how `if` is parsed in rust right now.

### Lint Name

identity_op

### Reproducer

I tried this code:

```rust
pub fn decide(a: bool, b: bool) -> u32 {
    0 + if a { 1 } else { 2 } + if b { 3 } else { 5 }
}
```

I saw this happen:

```
warning: the operation is ineffective. Consider reducing it to `if a { 1 } else { 2 }`
 [--> src/lib.rs:2:5
](https://play.rust-lang.org/#)  |
2 |     0 + if a { 1 } else { 2 } + if b { 3 } else { 5 }
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(clippy::identity_op)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#identity_op
```

### Version

```text
rustc 1.62.0-nightly (311e2683e 2022-04-18)
binary: rustc
commit-hash: 311e2683e1bad87715b1558f7900e294d24ce491
commit-date: 2022-04-18
release: 1.62.0-nightly
LLVM version: 14.0.1
```

### Additional Labels

@rustbot label: I-suggestions-causes-error

tamaroning:
fix #8724 

changelog: FP: [`identity_op`]: is now allowed in front of if statements, blocks and other expressions where the suggestion would be invalid.
 
Resolved simular problems with blocks, mathces, and loops.
identity_op always does NOT suggest reducing `0 + if b { 1 } else { 2 } + 3` into  `if b { 1 } else { 2 } + 3` even in the case that the expression is in `f(expr)` or `let x = expr;` for now.

### DIFF
--- a/clippy_lints/src/identity_op.rs
+++ b/clippy_lints/src/identity_op.rs
@@ -1,3 +1,4 @@
+use clippy_utils::get_parent_expr;
 use clippy_utils::source::snippet;
 use rustc_hir::{BinOp, BinOpKind, Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
@@ -22,6 +23,11 @@ declare_clippy_lint! {
     /// # let x = 1;
     /// x / 1 + 0 * 1 - 0 | 0;
     /// ```
+    ///
+    /// ### Known problems
+    /// False negatives: `f(0 + if b { 1 } else { 2 } + 3);` is reducible to
+    /// `f(if b { 1 } else { 2 } + 3);`. But the lint doesn't trigger for the code.
+    /// See [#8724](https://github.com/rust-lang/rust-clippy/issues/8724)
     #[clippy::version = "pre 1.29.0"]
     pub IDENTITY_OP,
     complexity,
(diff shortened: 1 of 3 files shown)


tests/ui/identity_op.rs
```diff
@@ -77,4 +77,20 @@ fn main() {
     (x + 1) % 3; // no error
     4 % 3; // no error
     4 % -3; // no error
+
+    // See #8724
+    let a = 0;
+    let b = true;
+    0 + if b { 1 } else { 2 };
```
hellow554, line 84: I would like to see a testcase like `let _c = 0 + if b { 1 } else { 2 };`
Maybe also a function like I wrote in #8724:
```rust
pub fn decide(a: bool, b: bool) -> u32 {
    0 + if a { 1 } else { 2 } + if b { 3 } else { 5 }
}
```

tamaroning:
Hmm, uitest somehow fails.
Despite of attribute #[allow(clippy::identity_op)], the warning of identity_op occurs.
```
---- compile_test stdout ----
diff of stderr:

+error: the operation is ineffective. Consider reducing it to `i`
error: test failed, to rerun pass '--test compile-test'
+  --> $DIR/without_loop_counters.rs:104:13
+   |
+LL |         dst[i - 0] = src[i];
+   |             ^^^^^
+   |
+   = note: `-D clippy::identity-op` implied by `-D warnings`
+
```

https://github.com/rust-lang/rust-clippy/blob/ed22428b7292db18f8e0b90e06b91a802c0bf647/tests/ui/manual_memcpy/without_loop_counters.rs#L102-L105

clippy_lints/src/identity_op.rs
```diff
@@ -31,33 +32,99 @@ declare_clippy_lint! {
...
-                },
-                BinOpKind::Div => check(cx, right, 1, e.span, left.span),
-                BinOpKind::BitAnd => {
-                    check(cx, left, -1, e.span, right.span);
-                    check(cx, right, -1, e.span, left.span);
-                },
-                BinOpKind::Rem => check_remainder(cx, left, right, e.span, left.span),
-                _ => (),
+        }
+    }
```
tamaroning, line 69: Checks if `left op ...right` is reducible to `...right` based on the information gathered by traverser `ExprVisitor`.

Manishearth:
r? @Alexendoo 

(I'm travelling rn and might not be able to do full reviews, can rubber-stamp after you've had a look at it)

clippy_lints/src/identity_op.rs
```diff
@@ -31,33 +32,99 @@ declare_clippy_lint! {
...
+        };
+        visitor.visit_expr(&body.value);
+    }
+}
+
+struct ExprVisitor<'a, 'tcx> {
+    cx: &'a LateContext<'tcx>,
+    leftmost: bool,
+    top_level_binop: bool,
+}
```
Alexendoo, line 52: I think a visitor is too confusing here, e.g. `leftmost` is always true by the time it reaches the `self.reducible_to_right` calls

A simpler alternative would be to go back to using `check_expr` in the lint pass and checking if `e`'s [parent expr](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.get_parent_expr.html) is an `ExprKind::Binary` to determine if it's top level or not

Alexendoo (commented):
👋

clippy_lints/src/identity_op.rs
tamaroning: Thanks for your review. I fixed it.

Alexendoo: I think the visitor can be removed entirely, at least for all the test cases `leftmost` is still always true in `self.reducible_to_right` because the order statements are visited in are

```rust
     a + b + c + d;
// 1 LLLLLLLLL   R
// 2 LLLLL   R
// 3 L   R
```

I believe just checking the `ExprKind` of right as you do in `reducible_to_right` + `is_toplevel_binary` is enough

tamaroning: Ohh I misunderstood what you meant. And it seems to work for test cases.

tamaroning: Fixed it.

clippy_lints/src/identity_op.rs
```diff
@@ -31,36 +32,66 @@ declare_clippy_lint! {
...
         }
     }
 }
 
+// Checks if `left op ..right` can be actually reduced to `right`
+// e.g. `0 + if b { 1 } + else { 2 } +  if b { 3 } + else { 4 }`
+// cannot be reduced to `if b { 1 } + else { 2 } +  if b { 3 } + else { 4 }`
+// See #8724
+fn reducible_to_right(cx: &LateContext<'_>, binary: &Expr<'_>, right: &Expr<'_>) -> bool {
+    if let ExprKind::If(..) | ExprKind::Match(..) = right.kind {
```
Alexendoo, line 80: Can also include `ExprKind::Block` and `ExprKind::Loop`

tests/ui/identity_op.rs
```diff
@@ -77,4 +77,23 @@ fn main() {
...
+    0 + match a { 0 => 10, _ => 20 } + match a { 0 => 30, _ => 40 }; // no error
+    0 + if b { 1 } else { 2 } + match a { 0 => 30, _ => 40 }; // no error
+    0 + match a { 0 => 10, _ => 20 } + if b { 3 } else { 4 }; // no error
+    // nested
+    0 + if b { 0 + 1 } else { 2 };
+    0 + match a { 0 =>  0 + 10, _ => 20 };
+    0 + if b { 0 + 1 } else { 2 } + match a { 0 => 0 + 30, _ => 40 };
+    // condition, scrutinee
+    let _ = 0 + if 0 + 1 > 0 { 1 } else { 2 } + if 0 + 1 > 0 { 3 } else { 4 };
+    let _ = 0 + match 0 + 1 { 0 => 10, _ => 20 } + match 0 + 1  { 0 => 30, _ => 40 };
```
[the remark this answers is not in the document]
tamaroning: Thanks.
But I think it seems to be enough to use `get_parent_expr`.
Is there need to check node variants, using `get_parent_node`?

tests/ui/manual_memcpy/without_loop_counters.rs
```diff
@@ -99,7 +100,6 @@ pub fn manual_copy(src: &[i32], dst: &mut [i32], dst2: &mut [i32]) {
         dst[i] = src[i - from];
     }
 
-    #[allow(clippy::identity_op)]
```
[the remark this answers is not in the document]
tamaroning: Fixed it.

clippy_lints/src/identity_op.rs
tamaroning: Fixed it.

tests/ui/identity_op.rs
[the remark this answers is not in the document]
Alexendoo: Those cases don't lint currently but they could do is what I mean, as in

```rust
// doesn't trigger the lint
f(0 + if b { 1 } else { 2 } + 3);
// but it could, as the following is valid
f(if b { 1 } else { 2 } + 3);
```

It'd be also fine to leave it as is and make a note of that under a `Known problems` section in the lint description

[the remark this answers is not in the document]
tamaroning: I didn't aware of this case. I will leave it as it is and add documents.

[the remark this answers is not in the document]
tamaroning: Added documents

Alexendoo:
Thanks @tamaroning!

Changes look good, please could you squash your commits into one with a descriptive message and then it'll be good to go

Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

xFrednet:
Hey, the changelog entry from the PR description is still missing and github is showing some merge conflicts. Could you rebase once more and add a changelog entry? Then all tests should pass :upside_down_face:

tamaroning:
@xFrednet Sorry, could you rerun the test?

xFrednet:
No problem @tamaroning, usually I would edit the changelog entry myself if it's the last thing, but here I wasn't sure anymore how it was exactly fixed. :upside_down_face:  Thank you for the changes :+1: 

@bors retry

PR #8731, opened 2022-04-22
dogfood: allow unknown lints when not running with `internal` feature

Alexendoo:
changelog: none

https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/unknown.20lint.20in.20test.20dogfood_clippy

It's only a warning so this wasn't causing the test to fail, but if you had another error somewhere or used `--nocapture` the extra warnings would be shown


xFrednet (approved):

xFrednet:
LGTM, Thanks!

@bors r+

PR #8737, opened 2022-04-23
Extend `extra_unused_lifetimes` to handle impl lifetimes

### LINKED ISSUE #6437: `extra_unused_lifetimes` doesn't warn for lifetimes declared on `impl` and never used
I tried this code (simplified from the actual code):

```rust
pub struct Scalar;

impl<'a> std::ops::AddAssign<&Scalar> for &mut Scalar {
    fn add_assign(&mut self, _rhs: &Scalar) {
        unimplemented!();
    }
}

impl<'b> Scalar {
    pub fn something<'c>() -> Self {
        Self
    }
}
```

I expected to see this happen: I expected to get an unused lifetime warning for all the lifetimes in this example: `'a` because it's not used anywhere in the `AddAssign` impl block (and `AddAssign`'s definition doesn't have any lifetimes), `'b` because it's not used anywhere in the `impl Scalar` block, and `'c` because it's not used anywhere in `fn something`.

Instead, this happened: I only see a warning for `'c`:

```
warning: this lifetime isn't used in the function definition
  --> src/lib.rs:10:22
   |
10 |     pub fn something<'c>() -> Self {
   |                      ^^
   |
   = note: `#[warn(clippy::extra_unused_lifetimes)]` on by default
```

### Meta

- `cargo clippy -V`: On the playground: 0.0.212 (2020-12-03 5be3f9f)
- `rustc -Vv`: Stable on the playground: 1.48.0

smoelius:
Fixes #6437 (cc: @carols10cents)

changelog: fix #6437

### DIFF
--- a/clippy_lints/src/lifetimes.rs
+++ b/clippy_lints/src/lifetimes.rs
@@ -13,7 +13,7 @@ use rustc_hir::{
     TraitFn, TraitItem, TraitItemKind, Ty, TyKind, WhereClause, WherePredicate,
 };
 use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::hir::nested_filter as mir_nested_filter;
+use rustc_middle::hir::nested_filter as middle_nested_filter;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::source_map::Span;
 use rustc_span::symbol::{kw, Ident, Symbol};
@@ -563,7 +563,7 @@ fn report_extra_impl_lifetimes<'tcx>(cx: &LateContext<'tcx>, impl_: &'tcx Impl<'
             _ => None,
         })
         .collect();
-    let mut checker = LifetimeChecker::<mir_nested_filter::All>::new(cx, hs);
+    let mut checker = LifetimeChecker::<middle_nested_filter::All>::new(cx, hs);
 
     walk_generics(&mut checker, &impl_.generics);
     if let Some(ref trait_ref) = impl_.of_trait {
--- a/clippy_lints/src/lifetimes.rs
+++ b/clippy_lints/src/lifetimes.rs
@@ -1,16 +1,19 @@
 use clippy_utils::diagnostics::span_lint;
 use clippy_utils::trait_ref_of_method;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_hir::intravisit::nested_filter::{self as hir_nested_filter, NestedFilter};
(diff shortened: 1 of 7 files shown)


giraffate (commented):
Would these cases be false positive? I found these here:
- https://github.com/rust-lang/cargo/blob/0.49.0/src/cargo/core/source/mod.rs#L111
- https://github.com/serde-rs/serde/blob/v1.0.118/serde/src/de/mod.rs#L476
```
// first case
use serde::de::Visitor;
pub trait Expected {
    fn fmt(&self, formatter: &mut std::fmt::Formatter);
}

impl<'de, T> Expected for T
where
    T: Visitor<'de>,
{
    fn fmt(&self, formatter: &mut std::fmt::Formatter) {}
}

// second case
pub trait Source {
    fn hey();
}

impl<'a, T: Source + ?Sized + 'a> Source for Box<T> {
    fn hey() {}
}
```

smoelius:
> Would these cases be false positive?

The answer is "yes," of course. :) I think that just pushed commit fixes the problem.

clippy_lints/src/lifetimes.rs
```diff
@@ -1,16 +1,19 @@
...
 };
 use rustc_hir::FnRetTy::Return;
 use rustc_hir::{
-    BareFnTy, BodyId, FnDecl, GenericArg, GenericBound, GenericParam, GenericParamKind, Generics, ImplItem,
+    BareFnTy, BodyId, FnDecl, GenericArg, GenericBound, GenericParam, GenericParamKind, Generics, Impl, ImplItem,
     ImplItemKind, Item, ItemKind, LangItem, Lifetime, LifetimeName, ParamName, PolyTraitRef, TraitBoundModifier,
     TraitFn, TraitItem, TraitItemKind, Ty, TyKind, WhereClause, WherePredicate,
 };
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::hir::nested_filter as mir_nested_filter;
```
giraffate, line 16: I don't know why this is named `mir`. Can this be explained?

giraffate (commented):
Overall looks good, thanks!

I made one comment for naming.

clippy_lints/src/lifetimes.rs
smoelius: I think I misread the crate name. Maybe it's better now (`middle_nested_filter`)?

giraffate:
@bors r+

Thanks!

PR #8742, opened 2022-04-24
mistyped_literal_suffix: improve integer suggestions, avoid wrong float suggestions

### LINKED ISSUE #6129: `mistyped_literal_suffixes` false positive and bad suggestion
```rust
 let _ = 123_32.123;
```

I expected to see this happen: `mistyped_literal_suffixes` not to fire.

Instead, this happened: 
```
  |
8 |     let _ = 123_32.123;
  |             ^^^^^^^^^^ help: did you mean to write: `123.123_f32`
  |
```

goth-turtle:
This PR fixes 2 things:
- The known problem that integer types are always suggested as signed, by suggesting an unsigned suffix for literals that wouldnt fit in the signed type, and ignores any literals too big for the corresponding unsigned type too.
- The lint would only look at the integer part of any floating point literals without an exponent, this causing #6129. This just ignores those literals.

Examples:
```rust
let _ = 2_32; // still 2_i32
let _ = 234_8; // would now suggest 234_u8

// these are now ignored
let _ = 500_8;
let _ = 123_32.123;
```

changelog: suggest correct integer types in [`mistyped_literal_suffix`], ignore float literals without an exponent
fixes #6129

### DIFF
--- a/clippy_lints/src/literal_representation.rs
+++ b/clippy_lints/src/literal_representation.rs
@@ -308,7 +308,7 @@ impl LiteralDigitGrouping {
         let (part, mistyped_suffixes, is_float) = if let Some((_, exponent)) = &mut num_lit.exponent {
             (exponent, &["32", "64"][..], true)
         } else if num_lit.fraction.is_some() {
-            (&mut num_lit.integer, &["32", "64"][..], true)
+            return true;
         } else {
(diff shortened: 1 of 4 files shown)


llogiq:
Thank you!

@bors r+

Issue #8732, opened 2022-04-22
`redundant_pub_crate` FP for macro definition


Kage-Yami:
### Summary

I've written an "internal" macro in my crate, and don't want it to be exported at the root/publicly, so instead of annotating it with `#[macro_export]`, I've added `pub(crate) use bit;` after the definition. Following the recommendation from `clippy::redundant_pub_crate` to remove `(crate)` prevents the crate from compiling because the macro:

> is only public within the crate, and cannot be re-exported outside

### Lint Name

redundant_pub_crate

### Reproducer

I tried this code:

```rust
// in lib.rs
#![warn(clippy::nursery)]
mod util;

// in util.rs
macro_rules! bit {
    ($bit:literal, $value:ident) => {
        $value >> $bit & 1 == 1
    };

    ($bit:ident, $value:ident) => {
        $value >> $bit & 1 == 1
    };
}

pub(crate) use bit;
```

I saw this happen:

```log
warning: pub(crate) import inside private module
  --> src\util.rs:40:1
   |
40 | pub(crate) use bit;
   | ----------^^^^^^^^
   | |
   | help: consider using: `pub`
   |
note: the lint level is defined here
  --> src\lib.rs:1:24
   |
1  | #![warn(clippy::cargo, clippy::nursery, clippy::pedantic)]
   |                        ^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::redundant_pub_crate)]` implied by `#[warn(clippy::nursery)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pub_crate
```

I expected to not see a lint warning.

### Version

```text
rustc 1.60.0 (7737e0b5c 2022-04-04)
binary: rustc
commit-hash: 7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c
commit-date: 2022-04-04
host: x86_64-pc-windows-msvc
release: 1.60.0
LLVM version: 14.0.0
```

### Additional Labels

@rustbot label +I-suggestion-causes-error

Kage-Yami:
(Updated OP to add missing `#![warn(clippy::nursery)]` into `lib.rs`; lint won't trigger otherwise.)

Kage-Yami:
And it turns out I can't even explicitly allow it at the usage site, because then `clippy::useless_attribute` fires (which would also be a FP?), which is deny-by-default.

Though it does work if I then allow `clippy::useless_attribute`. 😂

```rust
[...]

PR #8743, opened 2022-04-24
ignore `redundant_pub_crate` in `useless_attribute`

Alexendoo:
changelog: [`useless_attribute`] no longer lints [`redundant_pub_crate`]

As mentioned in https://github.com/rust-lang/rust-clippy/issues/8732#issuecomment-1106489634

> And it turns out I can't even explicitly allow it at the usage site, because then `clippy::useless_attribute` fires (which would also be a FP?), which is deny-by-default.
> 
> Though it does work if I then allow `clippy::useless_attribute`. 😂
> 
> ```rust
> #[allow(clippy::useless_attribute)]
> #[allow(clippy::redundant_pub_crate)]
> pub(crate) use bit;
> ```
> 
> The originally-reported warning now no longer occurs.

### DIFF
--- a/clippy_lints/src/attrs.rs
+++ b/clippy_lints/src/attrs.rs
@@ -335,9 +335,6 @@ impl<'tcx> LateLintPass<'tcx> for Attributes {
                     }
                     if let Some(lint_list) = &attr.meta_item_list() {
                         if attr.ident().map_or(false, |ident| is_lint_level(ident.name)) {
-                            // permit `unused_imports`, `deprecated`, `unreachable_pub`,
-                            // `clippy::wildcard_imports`, and `clippy::enum_glob_use` for `use` items
-                            // and `unused_imports` for `extern crate` items with `macro_use`
                             for lint in lint_list {
                                 match item.kind {
                                     ItemKind::Use(..) => {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/attrs.rs
```diff
@@ -335,20 +335,19 @@ impl<'tcx> LateLintPass<'tcx> for Attributes {
...
                                             || is_word(lint, sym::deprecated)
                                             || is_word(lint, sym!(unreachable_pub))
                                             || is_word(lint, sym!(unused))
-                                            || extract_clippy_lint(lint)
-                                                .map_or(false, |s| s.as_str() == "wildcard_imports")
-                                            || extract_clippy_lint(lint)
-                                                .map_or(false, |s| s.as_str() == "enum_glob_use")
+                                            || extract_clippy_lint(lint).map_or(false, |s| {
+                                                matches!(
+                                                    s.as_str(),
```
llogiq, line 347: Why did you change the symbol lookup to a str comparison?

llogiq (commented):
This looks good in general, I only have one question.

clippy_lints/src/attrs.rs
Alexendoo: It was already a str comparison, the change I made was to use `matches!` rather than add a third `|| extract...`

llogiq:
Thank you! @bors r+

PR #8744, opened 2022-04-24
Extend `needless_collect`

Jarcho:
Extends when `is_empty` and `contains` are linted. 

`is_empty` will be linted when `<IterTy as Iterator>::Item` is the same as `<CollectTy as IntoIterator>::Item`. This can be a false positive if the `FromIterator` implementation filters out items, but I don't know of any which do that also implement `IntoIterator` with a matching `Item` type.

`contains` will be linted when the argument type is `&<IterTy as Iterator>::Item`. It has the same false positives as `is_empty` with the same note that I know of nothing that actually causes that in practice.

changelog: Lint `needless_collect` when `is_empty` or `contains` is called on some non-std types


Jarcho:
r? @llogiq
Assign this to somebody still doing reviews.

clippy_lints/src/methods/needless_collect.rs
```diff
@@ -2,94 +2,112 @@ use super::NEEDLESS_COLLECT;
...
-                match method_name {
-                    "len" => "count()".to_string(),
-                    "is_empty" => is_empty_sugg,
-                    "contains" => {
-                        let contains_arg = snippet_with_applicability(cx, args[1].span, "??", &mut applicability);
-                        let (arg, pred) = contains_arg
-                            .strip_prefix('&')
-                            .map_or(("&x", &*contains_arg), |s| ("x", s));
-                        format!("any(|{}| x == {})", arg, pred)
+#[allow(clippy::too_many_lines)]
```
llogiq, line 25: Perhaps factor out some functions?

clippy_lints/src/methods/needless_collect.rs
```diff
@@ -2,94 +2,112 @@ use super::NEEDLESS_COLLECT;
...
+) {
+    if let Some(parent) = get_parent_node(cx.tcx, collect_expr.hir_id) {
+        match parent {
+            Node::Expr(parent) => {
+                if let ExprKind::MethodCall(name, args, _) = parent.kind {
+                    let mut app = Applicability::MachineApplicable;
+                    let name = name.ident.as_str();
+                    let collect_ty = cx.typeck_results().expr_ty(collect_expr);
+
+                    if name == "len" {
```
llogiq, line 41: Let chains could also be used here to reduce both lines of code and rightwards drift.

clippy_lints/src/methods/needless_collect.rs
```diff
@@ -2,94 +2,112 @@ use super::NEEDLESS_COLLECT;
...
+                }
+            },
+            Node::Local(l) => {
+                if let PatKind::Binding(BindingAnnotation::Unannotated | BindingAnnotation::Mutable, id, _, None)
+                    = l.pat.kind
+                    && let ty = cx.typeck_results().expr_ty(collect_expr)
+                    && (is_type_diagnostic_item(cx, ty, sym::Vec) ||
+                        is_type_diagnostic_item(cx, ty, sym::VecDeque) ||
+                        is_type_diagnostic_item(cx, ty, sym::BinaryHeap) ||
+                        is_type_diagnostic_item(cx, ty, sym::LinkedList))
```
llogiq, line 105:
```suggestion
                    && [sym::Vec, sym::VecDeque,   sym::BinaryHeap, sym::LinkedList].into_iter().any(|item| is_type_diagnostic_item(cx, ty, item))
```

clippy_utils/src/ty.rs
```diff
@@ -590,3 +591,133 @@ pub fn is_c_void(cx: &LateContext<'_>, ty: Ty<'_>) -> bool {
...
 }
+
+/// Makes the projection type for the named associated type in the given impl or trait impl.
+///
+/// This function is for associated types which are "known" to exist, and as such, will only return
+/// `None` when debug assertions are disabled in order to prevent ICE's. With debug assertions
+/// enabled this will check that the named associated type exists, the correct number of
+/// substitutions are given, and that the correct kinds of substitutions are given (lifetime,
+/// constant or type). This will not check if type normalization would succeed.
+pub fn make_projection<'tcx>(
```
llogiq, line 602: I must admit that I don't quite understand what this is about.

llogiq (commented):
This looks good in general, although some of the stuff goes over my head.

Unfortunately I got some pink eye today, so my first review is rather cursory, as I try to limit my screen time. Will come back to this once I got some antibiotics, but today is a bank holiday here in 🇩🇪.

clippy_utils/src/ty.rs
Jarcho: If some type implements a `T`, then the projected type `T::Ty` will exist. This is the intended use of the function. It's possible while writing lints to mix things up and make an invalid assumption about this, or to mess up the generic substitutions. In that case `evaluate_obligations` will fail or crash when it gets called.

Both `make_projection` and `make_normalized_projection` will crash on debug builds so the error is obvious on debug builds. For released version of clippy this will avoid crashing and instead risk resulting in false positives or negatives. Running to the end with slightly wrong results is generally more useful than crashing part way through.

llogiq:
Ok, I'm back. My eyes have fully recovered. Also, this looks good in general, and if you don't want to refactor it, I'm fine with merging it after a rebase and doing the refactor myself later (if nobody beats me to it).

Jarcho:
I'll get to it this sometime this week.

llogiq:
Ping @Jarcho anything I can do to help?

Jarcho:
This is still on my todo list. Ended up getting sick instead of working on this. Better now though.

llogiq:
Ping! So I guess this fell off the top of your Todo list. No hard feelings here, but it would be great to see this through.

@rustbot author

Jarcho:
Should be good now.

llogiq:
The current commit fails dogfood. Apparently because of the recently introduced format args lint.

llogiq:
Thanks!

@bors r+

PR #8761, opened 2022-04-29
`undocumented_unsafe_blocks` does not trigger on unsafe trait impls

### LINKED ISSUE #8505: `undocumented_unsafe_blocks` does not trigger on unsafe trait impls
### Summary

`missing_safety_doc` and `undocumented_unsafe_blocks` complement each other to provide full safety documentation coverage. While `missing_safety_doc` triggers on unsafe trait definitions and warn that safety docs must be provided, `undocumented_unsafe_blocks` does not trigger on unsafe trait implementations.

### Lint Name

undocumented_unsafe_blocks

### Reproducer

I tried this code:

```rust
#![warn(clippy::undocumented_unsafe_blocks)]

/// # Safety
///
/// The implementor must justify with a safety comment.
unsafe trait Foo {}

unsafe impl Foo for () {}

/// # Safety
///
/// The caller must justify with a safety comment.
unsafe fn foo() {}

fn main() {
    // SAFETY: It's justified.
    unsafe {
        foo();
    }
}
```

```
warning: unsafe impl missing a safety comment
  --> src/main.rs:8:1
```

Instead, this happened:

No lint triggered for the unsafe impl.

### Version

```text
binary: rustc
commit-date: 2022-02-23
host: x86_64-apple-darwin
release: 1.59.0
LLVM version: 13.0.0
```

tamaroning:
Closes #8505 

changelog: This lint checks unsafe impls NOT from macro expansions and checks ones in macro declarations.
~~`unsafe impl`s from macro invocations don't trigger the lint for now.~~
~~This lint checks unsafe impls from/not from macro expansions~~

### DIFF
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -142,20 +142,19 @@ fn block_has_safety_comment(cx: &LateContext<'_>, block: &hir::Block<'_>) -> boo
 /// Checks if the lines immediately preceding the item contain a safety comment.
 #[allow(clippy::collapsible_match)]
 fn item_has_safety_comment(cx: &LateContext<'_>, item: &hir::Item<'_>) -> bool {
-    if span_from_macro_expansion_has_safety_comment(cx, item.span) || span_in_body_has_safety_comment(cx, item.span) {
+    if span_from_macro_expansion_has_safety_comment(cx, item.span) {
         return true;
     }
 
     if item.span.ctxt() == SyntaxContext::root() {
         if let Some(parent_node) = get_parent_node(cx.tcx, item.hir_id()) {
-            let comment_start;
-            match parent_node {
+            let comment_start = match parent_node {
                 Node::Crate(parent_mod) => {
-                    comment_start = comment_start_before_impl_in_mod(cx, parent_mod, parent_mod.spans.inner_span, item);
+                    comment_start_before_impl_in_mod(cx, parent_mod, parent_mod.spans.inner_span, item)
                 },
                 Node::Item(parent_item) => {
                     if let ItemKind::Mod(parent_mod) = &parent_item.kind {
-                        comment_start = comment_start_before_impl_in_mod(cx, parent_mod, parent_item.span, item);
+                        comment_start_before_impl_in_mod(cx, parent_mod, parent_item.span, item)
                     } else {
                         // Doesn't support impls in this position. Pretend a comment was found.
                         return true;
@@ -164,16 +163,14 @@ fn item_has_safety_comment(cx: &LateContext<'_>, item: &hir::Item<'_>) -> bool {
                 Node::Stmt(stmt) => {
                     if let Some(stmt_parent) = get_parent_node(cx.tcx, stmt.hir_id) {
                         match stmt_parent {
-                            Node::Block(block) => {
-                                comment_start = walk_span_to_context(block.span, SyntaxContext::root()).map(Span::lo);
-                            },
+                            Node::Block(block) => walk_span_to_context(block.span, SyntaxContext::root()).map(Span::lo),
                             _ => {
                                 // Doesn't support impls in this position. Pretend a comment was found.
                                 return true;
                             },
                         }
                     } else {
-                        // Doesn't support impls in this position. Pretend a comment was found.
+                        // Problem getting the parent node. Pretend a comment was found.
                         return true;
                     }
                 },
@@ -181,7 +178,7 @@ fn item_has_safety_comment(cx: &LateContext<'_>, item: &hir::Item<'_>) -> bool {
                     // Doesn't support impls in this position. Pretend a comment was found.
                     return true;
                 },
-            }
+            };
 
             let source_map = cx.sess().source_map();
             if let Some(comment_start) = comment_start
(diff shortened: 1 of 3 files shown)


giraffate:
Sorry, I can't take much time this week, but next week I can take time for a review.

@Jarcho Could you do a review on this?

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -86,6 +87,65 @@ impl LateLintPass<'_> for UndocumentedUnsafeBlocks {
...
+        item_and_spans.sort_by_key(|e| e.1.lo());
+
+        for (idx, (item, item_span)) in item_and_spans.iter().enumerate() {
+            if let hir::ItemKind::Impl(imple) = &item.kind
+                && imple.unsafety == hir::Unsafety::Unsafe
+                && !item_span.from_expansion()
+                && !is_lint_allowed(cx, UNDOCUMENTED_UNSAFE_BLOCKS, hir_id)
+            {
+                // Checks if the lines immediately preceding the impl contain a safety comment.
+                let impl_has_safety_comment = {
```
Jarcho, line 110: Minor nit here. This block isn't needed. The first statement can be moved out, then the if assigned directly.

Jarcho (changes requested):

tamaroning:
Thanks for your review @Jarcho 
I fixed code.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -171,13 +136,85 @@ fn block_has_safety_comment(cx: &LateContext<'_>, block: &Block<'_>) -> bool {
     // won't work. This is to avoid dealing with where such a comment should be place relative to
     // attributes and doc comments.
 
+    span_from_macro_expansion_has_safety_comment(cx, block.span) || span_in_body_has_safety_comment(cx, block.span)
+}
+
+/// Checks if the lines immediately preceding the item contain a safety comment.
+fn item_has_safety_comment(cx: &LateContext<'_>, item: &hir::Item<'_>) -> bool {
+    if span_from_macro_expansion_has_safety_comment(cx, item.span) || span_in_body_has_safety_comment(cx, item.span) {
```
Jarcho, line 144: This should be returning false if the item is within a function, but doesn't have a safety comment. e.g.
```rust
unsafe trait Foo {}
fn f() {
    Struct S;
    // no comment
    unsafe impl Foo for S {}
}
```

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -171,13 +136,85 @@ fn block_has_safety_comment(cx: &LateContext<'_>, block: &Block<'_>) -> bool {
...
+        return true;
+    }
+
+    if item.span.ctxt() == SyntaxContext::root() {
+        if let Some(parent_node) = get_parent_node(cx.tcx, item.hir_id()) {
+            let mut span_before_item = None;
+            let mut hi = false;
+            if let Node::Item(parent_item) = parent_node {
+                if let ItemKind::Mod(parent_mod) = &parent_item.kind {
+                    for (idx, item_id) in parent_mod.item_ids.iter().enumerate() {
```
Jarcho, line 154: It would be better to use `Iterator::find_map` here. the extraction of either `Span::lo` or `Span::hi` could be done here as well.

This should get rid of the need to have `span_before_item` and `hi` as mutable.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -171,13 +136,85 @@ fn block_has_safety_comment(cx: &LateContext<'_>, block: &Block<'_>) -> bool {
...
+                                //         ^---------------^ finally checks the text in this range
+                                hi = true;
+                                span_before_item = Some(prev_item.span);
+                            }
+                            break;
+                        }
+                    }
+                }
+            }
+            let span_before_item = span_before_item.unwrap();
```
Jarcho, line 175: This can fail when the impl is inside a function and I think the crate root.

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -171,13 +136,85 @@ fn block_has_safety_comment(cx: &LateContext<'_>, block: &Block<'_>) -> bool {
...
+
+            let source_map = cx.sess().source_map();
+            if let Some(item_span) = walk_span_to_context(item.span, SyntaxContext::root())
```
Jarcho, line 178: `walk_span_to_context` isn't needed here. `item.span` is already at the root context.

Jarcho (changes requested):

tamaroning:
Thanks for your review @Jarcho 
I fixed all.

Jarcho (changes requested):
Just two more small things. I think it's good otherwise, but I'll take one more look at it later. Thanks for keeping up with the review.

tamaroning:
Thanks! I removed  the call to `span_in_body_has_safety_comment`, but I am not sure impls in a body 
 always has a `Node::Stmt(..)` as a parent node and a `Node::Block(..)` as a grandparent node.

Jarcho:
The only way an item can appear in a function is through `StmtKind::Item`, and the only way a statement can appear is through a block.

Jarcho:
Everything looks good now. Good job.

@giraffate

tamaroning:
Thank you so much! @Jarcho

giraffate (approved):

giraffate:
@bors r=Jarcho

@tamaroning @Jarcho It looks good, thanks!

Jarcho:
You'll need to add `changelog: ...` on a single line in the PR body (first post).

PR #8762, opened 2022-04-29
Replace `expr_visitor` with  `for_each_expr`

Jarcho:
This is a minor change which uses `ControlFlow` rather than a boolean. This also runs the visitor rather than returning the visitor, which results in a small readability win as well.

changelog: None


Jarcho:
Great. The compiler broke...

flip1995:
LGTM. Will merge after the next sync.

flip1995:
Huh. Forgot about it. Can you rebase and then r+ yourself, if CI passes?

Jarcho:
Looks like this is still broken. rust-lang/rust#101410 is a possible fix.

Jarcho:
Narrowed the issue down to something with let-chains mir opt levels. Pretty good match with rust-lang/#99852. Just added a workaround until the rustc bug is fixed.

@bors r+

Issue #8721, opened 2022-04-19
manual_range_contains misbehaves when bounds have different signs


xTibor:
### Summary

The `manual_range_contains` lint misbehaves when the upper and lower bounds of the range have different signs. There are four test cases in the reproducer:

- Test 1 - Upper and lower bounds are both positive. Suggestions are okay.
- Test 2 - Upper and lower bounds are both negative. Suggestions are okay.
- Test 3 - Upper bound is positive, lower bound is negative. Clippy misses this case, no warnings, no suggestions.
- Test 4 - Upper bound is negative, lower bound is positive. This range is non-sense, however Clippy makes an invalid suggestion here that later triggers an another warning about reversed ranges. Clippy should warn about that non-sense condition instead.

### Reproducer

```rust
fn main() {
    let value = 0;

    // OK suggestion: `(5..=10).contains(&value)`
    let _test_1_pos_pos = (value >= 5) && (value <= 10);

    // OK suggestion: `(-10..=-5).contains(&value)`
    let _test_2_neg_neg = (value >= -10) && (value <= -5);

    // Missing suggestion. Expected: `(-10..=10).contains(&value)`
    let _test_3_neg_pos = (value >= -10) && (value <= 10);

    // Invalid suggestion: `(10..=-10).contains(&value)`
    // The suggested code triggers `clippy::reversed_empty_ranges`.
    let _test_4_pos_neg = (value >= 10) && (value <= -10);
}
```

### Version

```text
rustc 1.62.0-nightly (311e2683e 2022-04-18)
binary: rustc
commit-hash: 311e2683e1bad87715b1558f7900e294d24ce491
commit-date: 2022-04-18
host: x86_64-unknown-linux-gnu
release: 1.62.0-nightly
LLVM version: 14.0.1
```

PR #8763, opened 2022-04-30
Support negative ints in manual_range_contains

arieluy:
fixes: #8721 
changelog: Fixes issue where ranges containing ints with different signs would be
incorrect due to comparing as unsigned.

### DIFF
--- a/clippy_lints/src/ranges.rs
+++ b/clippy_lints/src/ranges.rs
@@ -219,15 +219,17 @@ fn check_possible_range_contains(cx: &LateContext<'_>, op: BinOpKind, l: &Expr<'
         _ => return,
     };
     // value, name, order (higher/lower), inclusiveness
-    if let (Some((lval, lid, name_span, lval_span, lord, linc)), Some((rval, rid, _, rval_span, rord, rinc))) =
-        (check_range_bounds(cx, l), check_range_bounds(cx, r))
+    if let (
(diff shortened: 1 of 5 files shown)


clippy_lints/src/ranges.rs
```diff
@@ -292,7 +294,10 @@ fn check_possible_range_contains(cx: &LateContext<'_>, op: BinOpKind, l: &Expr<'
     }
 }
 
-fn check_range_bounds(cx: &LateContext<'_>, ex: &Expr<'_>) -> Option<(Constant, HirId, Span, Span, Ordering, bool)> {
+fn check_range_bounds<'a>(
+    cx: &'a LateContext<'_>,
+    ex: &'a Expr<'_>,
+) -> Option<(Constant, &'a Expr<'a>, HirId, Span, Span, Ordering, bool)> {
```
xFrednet, line 300: NIT: This return type is very complex and not descriptive. Would you mind creating a struct with named values instead?

If not, we can also create an issue to this :upside_down_face:

xFrednet (commented):
Hey and welcome to Clippy :wave:. Your changes look good to me.

I have one small change request, if you don't mind. Then everything is ready to be merged :upside_down_face:

clippy_lints/src/ranges.rs
arieluy: Thanks! I added a struct, how does it look?

PR #8764, opened 2022-04-30
Update GitHub Actions actions/checkout@v2 to v3

dtolnay:
The v2 implementation uses Node 12, which is end-of-life on April 30, 2022. See https://nodejs.org/en/about/releases/. Update to v3, which is based on Node 16 whose support lasts until April 30, 2024.

---

changelog: none


llogiq:
Phew, just in time. 😅 Thank you!

@bors r+

Issue #8765, opened 2022-04-30
missing double quote in source example on master folder in `large_include_file` lint


alexzanderr:
### Description

if you go on this url: https://rust-lang.github.io/rust-clippy/master/index.html#large_include_file

you will see that the source example has a missing `"`:
```rust
let included_str = include_str!("very_large_file.txt");
let included_bytes = include_bytes!("very_large_file.txt);
                                                       ^
                                                       |

error[E0765]: unterminated double quote string
 --> test.rs:5:37
  |
5 |   let included_bytes = include_bytes!("very_large_file.txt);
  |  _____________________________________^
6 | | }
  | |__^
```

please fix that, because im making a program that validates that source code and i need it to compile.

i know that its an insignificant issue, im sorry, but my program will fail.

so take your time.

xFrednet:
Hey @alexzanderr, thank you for the report. I've created a PR to fix this. For your tool it would probably be good to ignore example code blocks that are marked with `ignore`.

Out of interest, would you mind providing some more information about the tool you're creating? It sounds interesting :upside_down_face:

alexzanderr:
cargo-clippy-explain

its a tool that does web scraping for a lint and gets the examples and explanations which then will be syntax highlighted and a little beautyfied in the terminal.

since the text in the artcile tag has no classes and no ids,
i need to identify which lines from example are `valid rust source code`, so im compiling the code to see if `let some_vec = vec![0, 1, 2, 3];` its rust or not or if `The correct use would be:` its rust or not.

after some filtering and beautyfication, the output is printed to stdout.

this will get me and others understand how to write clean code, since understanding the clippy lints is key factor to write clean code, this tool its going to be great.

i just began to be rust dev.
i would love to help the community in all the ways.
i love rust. just moved from python.

xFrednet:
In rust doc everything is expected to be valid rust code unless a different language is specified. You could also try to use `cargo collect-metadata` in our repo, that will give you a json file with all lint information that is available over on [Clippy's lint list](https://rust-lang.github.io/rust-clippy/master/index.html)

[...]

PR #8767, opened 2022-04-30
Add missing quite in `large_include_file` example

xFrednet:
Roses are red,
violets are blue,
this fix,
was simple to do

Closes: https://github.com/rust-lang/rust-clippy/issues/8765

changelog: None


xFrednet:
r? @Alexendoo 

@bors delegate=Alexendoo

You can directly r+ if you agree with the changes :upside_down_face:

PR #8769, opened 2022-04-30
introduce rc_clone_in_vec_init lint

### LINKED ISSUE #8719: Warn about using Arc::new() when using the vec![value; size] macro.
### What it does

This lint would warn against using Arc::new in the vec![value; size] macro, because it computes the value once and then clones the value size times instead of generating a new Arc each of the size times.

If the user truly wants a vector of Arcs all pointing to the same arc, then they should make that obvious by constructing the arc before constructing the vector.

### Lint Name

_No response_

### Category

_No response_

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
  #[test]
  fn unexpected_vec_macro_behavior() {
      let v = vec![Arc::new(Mutex::new(0usize)); 100];
      *v[0].lock().unwrap() = 20;
      *v[1].lock().unwrap() = 30;
      assert_eq!(*v[0].lock().unwrap(), 20);
  }
```

Could be written as:

```rust
    #[test]
    fn vec_behavior_is_more_obvious() {
        let cloned_arc = Arc::new(Mutex::new(0usize));
        let v = vec![cloned_arc; 100];
        *v[0].lock().unwrap() = 20;
        *v[1].lock().unwrap() = 30;
        assert_eq!(*v[0].lock().unwrap(), 30);
    }
```

yonip23:
Closes #8719

changelog: Introduce [`rc_clone_in_vec_init`] lint

### DIFF
--- a/clippy_lints/src/rc_clone_in_vec_init.rs
+++ b/clippy_lints/src/rc_clone_in_vec_init.rs
@@ -1,0 +1,90 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::higher::VecArgs;
+use clippy_utils::last_path_segment;
+use clippy_utils::macros::{root_macro_call_first_node, MacroCall};
+use rustc_hir::{Expr, ExprKind, QPath, TyKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::{sym, Symbol};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `Arc::new` or `Rc::new` in `vec![elem; len]`
+    ///
+    /// ### Why is this bad?
+    /// This will create `elem` once and clone it `len` times - doing so with `Arc` or `Rc`
+    /// is a bit misleading, as it will create references to the same pointer, rather
+    /// than different instances.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let v = vec![std::sync::Arc::new("some data".to_string()); 100];
+    /// // or
+    /// let v = vec![std::rc::Rc::new("some data".to_string()); 100];
+    /// ```
+    /// Use instead:
+    /// ```rust
+    ///
+    /// // Initialize each value separately:
+    /// let mut data = Vec::with_capacity(100);
+    /// for _ in 0..100 {
+    ///     data.push(std::rc::Rc::new("some data".to_string()));
+    /// }
+    ///
+    /// // Or if you want clones of the same reference,
+    /// // Create the reference beforehand to clarify that
+    /// // it should be cloned for each value
+    /// let data = std::rc::Rc::new("some data".to_string());
+    /// let v = vec![data; 100];
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub RC_CLONE_IN_VEC_INIT,
+    suspicious,
+    "initializing `Arc` or `Rc` in `vec![elem; len]`"
+}
+declare_lint_pass!(RcCloneInVecInit => [RC_CLONE_IN_VEC_INIT]);
+
+impl LateLintPass<'_> for RcCloneInVecInit {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) else { return; };
+        let Some(VecArgs::Repeat(elem, _)) = VecArgs::hir(cx, expr) else { return; };
+        let Some(symbol) = new_reference_call(cx, elem) else { return; };
+
+        emit_lint(cx, symbol, &macro_call);
+    }
+}
+
+fn emit_lint(cx: &LateContext<'_>, symbol: Symbol, macro_call: &MacroCall) {
+    let symbol_name = symbol.as_str();
+
+    span_lint_and_then(
+        cx,
+        RC_CLONE_IN_VEC_INIT,
+        macro_call.span,
+        &format!("calling `{symbol_name}::new` in `vec![elem; len]`"),
+        |diag| {
+            diag.note(format!("each element will point to the same `{symbol_name}` instance"));
+            diag.help(format!(
+                "if this is intentional, consider extracting the `{symbol_name}` initialization to a variable"
+            ));
+            diag.help("or if not, initialize each element individually");
+        },
+    );
+}
+
+/// Checks whether the given `expr` is a call to `Arc::new` or `Rc::new`
+fn new_reference_call(cx: &LateContext<'_>, expr: &Expr<'_>) -> Option<Symbol> {
+    if_chain! {
+        if let ExprKind::Call(func, _args) = expr.kind;
+        if let ExprKind::Path(ref func_path @ QPath::TypeRelative(ty, _)) = func.kind;
+        if let TyKind::Path(ref ty_path) = ty.kind;
+        if let Some(def_id) = cx.qpath_res(ty_path, ty.hir_id).opt_def_id();
+        if last_path_segment(func_path).ident.name == sym::new;
+
+        then {
+            return cx.tcx.get_diagnostic_name(def_id).filter(|symbol| symbol == &sym::Arc || symbol == &sym::Rc);
+        }
+    }
+
+    None
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3642,6 +3642,7 @@ Released 2018-09-13
 [`range_step_by_zero`]: https://rust-lang.github.io/rust-clippy/master/index.html#range_step_by_zero
 [`range_zip_with_len`]: https://rust-lang.github.io/rust-clippy/master/index.html#range_zip_with_len
 [`rc_buffer`]: https://rust-lang.github.io/rust-clippy/master/index.html#rc_buffer
+[`rc_clone_in_vec_init`]: https://rust-lang.github.io/rust-clippy/master/index.html#rc_clone_in_vec_init
 [`rc_mutex`]: https://rust-lang.github.io/rust-clippy/master/index.html#rc_mutex
 [`recursive_format_impl`]: https://rust-lang.github.io/rust-clippy/master/index.html#recursive_format_impl
 [`redundant_allocation`]: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_allocation
--- a/clippy_lints/src/lib.register_all.rs
+++ b/clippy_lints/src/lib.register_all.rs
@@ -263,6 +263,7 @@ store.register_group(true, "clippy::all", Some("clippy_all"), vec![
     LintId::of(ranges::MANUAL_RANGE_CONTAINS),
     LintId::of(ranges::RANGE_ZIP_WITH_LEN),
     LintId::of(ranges::REVERSED_EMPTY_RANGES),
+    LintId::of(rc_clone_in_vec_init::RC_CLONE_IN_VEC_INIT),
     LintId::of(redundant_clone::REDUNDANT_CLONE),
     LintId::of(redundant_closure_call::REDUNDANT_CLOSURE_CALL),
     LintId::of(redundant_field_names::REDUNDANT_FIELD_NAMES),
--- a/clippy_lints/src/lib.register_lints.rs
+++ b/clippy_lints/src/lib.register_lints.rs
@@ -446,6 +446,7 @@ store.register_lints(&[
     ranges::RANGE_PLUS_ONE,
     ranges::RANGE_ZIP_WITH_LEN,
     ranges::REVERSED_EMPTY_RANGES,
+    rc_clone_in_vec_init::RC_CLONE_IN_VEC_INIT,
     redundant_clone::REDUNDANT_CLONE,
     redundant_closure_call::REDUNDANT_CLOSURE_CALL,
     redundant_else::REDUNDANT_ELSE,
--- a/clippy_lints/src/lib.register_suspicious.rs
+++ b/clippy_lints/src/lib.register_suspicious.rs
@@ -26,6 +26,7 @@ store.register_group(true, "clippy::suspicious", Some("clippy_suspicious"), vec!
     LintId::of(methods::SUSPICIOUS_MAP),
     LintId::of(mut_key::MUTABLE_KEY_TYPE),
     LintId::of(octal_escapes::OCTAL_ESCAPES),
+    LintId::of(rc_clone_in_vec_init::RC_CLONE_IN_VEC_INIT),
     LintId::of(suspicious_trait_impl::SUSPICIOUS_ARITHMETIC_IMPL),
     LintId::of(suspicious_trait_impl::SUSPICIOUS_OP_ASSIGN_IMPL),
 ])
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -345,6 +345,7 @@ mod ptr_offset_with_cast;
 mod pub_use;
 mod question_mark;
 mod ranges;
+mod rc_clone_in_vec_init;
 mod redundant_clone;
 mod redundant_closure_call;
 mod redundant_else;
@@ -890,6 +891,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     let max_include_file_size = conf.max_include_file_size;
     store.register_late_pass(move || Box::new(large_include_file::LargeIncludeFile::new(max_include_file_size)));
     store.register_late_pass(|| Box::new(strings::TrimSplitWhitespace));
+    store.register_late_pass(|| Box::new(rc_clone_in_vec_init::RcCloneInVecInit));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/tests/ui/rc_clone_in_vec_init/arc.rs
+++ b/tests/ui/rc_clone_in_vec_init/arc.rs
@@ -1,0 +1,49 @@
+#![warn(clippy::rc_clone_in_vec_init)]
+use std::sync::{Arc, Mutex};
+
+fn main() {}
+
+fn should_warn_simple_case() {
+    let v = vec![Arc::new("x".to_string()); 2];
+}
+
+fn should_warn_complex_case() {
+    let v = vec![
+        std::sync::Arc::new(Mutex::new({
+            let x = 1;
+            dbg!(x);
+            x
+        }));
+        2
+    ];
+}
+
+fn should_not_warn_custom_arc() {
+    #[derive(Clone)]
+    struct Arc;
+
+    impl Arc {
+        fn new() -> Self {
+            Arc
+        }
+    }
+
+    let v = vec![Arc::new(); 2];
+}
+
+fn should_not_warn_vec_from_elem_but_not_arc() {
+    let v = vec![String::new(); 2];
+    let v1 = vec![1; 2];
+    let v2 = vec![
+        Box::new(std::sync::Arc::new({
+            let y = 3;
+            dbg!(y);
+            y
+        }));
+        2
+    ];
+}
+
+fn should_not_warn_vec_macro_but_not_from_elem() {
+    let v = vec![Arc::new("x".to_string())];
+}
--- a/tests/ui/rc_clone_in_vec_init/arc.stderr
+++ b/tests/ui/rc_clone_in_vec_init/arc.stderr
@@ -1,0 +1,30 @@
+error: calling `Arc::new` in `vec![elem; len]`
+  --> $DIR/arc.rs:7:13
+   |
+LL |     let v = vec![Arc::new("x".to_string()); 2];
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: `-D clippy::rc-clone-in-vec-init` implied by `-D warnings`
+   = note: each element will point to the same `Arc` instance
+   = help: if this is intentional, consider extracting the `Arc` initialization to a variable
+   = help: or if not, initialize each element individually
+
+error: calling `Arc::new` in `vec![elem; len]`
+  --> $DIR/arc.rs:11:13
+   |
+LL |       let v = vec![
+   |  _____________^
+LL | |         std::sync::Arc::new(Mutex::new({
+LL | |             let x = 1;
+LL | |             dbg!(x);
+...  |
+LL | |         2
+LL | |     ];
+   | |_____^
+   |
+   = note: each element will point to the same `Arc` instance
+   = help: if this is intentional, consider extracting the `Arc` initialization to a variable
+   = help: or if not, initialize each element individually
+
+error: aborting due to 2 previous errors
+
--- a/tests/ui/rc_clone_in_vec_init/rc.rs
+++ b/tests/ui/rc_clone_in_vec_init/rc.rs
@@ -1,0 +1,50 @@
+#![warn(clippy::rc_clone_in_vec_init)]
+use std::rc::Rc;
+use std::sync::Mutex;
+
+fn main() {}
+
+fn should_warn_simple_case() {
+    let v = vec![Rc::new("x".to_string()); 2];
+}
+
+fn should_warn_complex_case() {
+    let v = vec![
+        std::rc::Rc::new(Mutex::new({
+            let x = 1;
+            dbg!(x);
+            x
+        }));
+        2
+    ];
+}
+
+fn should_not_warn_custom_arc() {
+    #[derive(Clone)]
+    struct Rc;
+
+    impl Rc {
+        fn new() -> Self {
+            Rc
+        }
+    }
+
+    let v = vec![Rc::new(); 2];
+}
+
+fn should_not_warn_vec_from_elem_but_not_rc() {
+    let v = vec![String::new(); 2];
+    let v1 = vec![1; 2];
+    let v2 = vec![
+        Box::new(std::rc::Rc::new({
+            let y = 3;
+            dbg!(y);
+            y
+        }));
+        2
+    ];
+}
+
+fn should_not_warn_vec_macro_but_not_from_elem() {
+    let v = vec![Rc::new("x".to_string())];
+}
--- a/tests/ui/rc_clone_in_vec_init/rc.stderr
+++ b/tests/ui/rc_clone_in_vec_init/rc.stderr
@@ -1,0 +1,30 @@
+error: calling `Rc::new` in `vec![elem; len]`
+  --> $DIR/rc.rs:8:13
+   |
+LL |     let v = vec![Rc::new("x".to_string()); 2];
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: `-D clippy::rc-clone-in-vec-init` implied by `-D warnings`
+   = note: each element will point to the same `Rc` instance
+   = help: if this is intentional, consider extracting the `Rc` initialization to a variable
+   = help: or if not, initialize each element individually
+
+error: calling `Rc::new` in `vec![elem; len]`
+  --> $DIR/rc.rs:12:13
+   |
+LL |       let v = vec![
+   |  _____________^
+LL | |         std::rc::Rc::new(Mutex::new({
+LL | |             let x = 1;
+LL | |             dbg!(x);
+...  |
+LL | |         2
+LL | |     ];
+   | |_____^
+   |
+   = note: each element will point to the same `Rc` instance
+   = help: if this is intentional, consider extracting the `Rc` initialization to a variable
+   = help: or if not, initialize each element individually
+
+error: aborting due to 2 previous errors
+


yonip23:
Hey there whoever is going to review this 👋🏻 

This is my first contribution to this (amazing) project

I'm not sure my lint's name is the best - if you have a better name please lmk!
Also, I'm not sure that my code is the most coherent way to write this - if it doesn't please feel free to leave some references for examples and I'll do my best!

Thanks ahead 🙁

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,82 @@
...
+        if !macro_is_vec(cx, &macro_call) {
+            return;
+        }
+
+        if_chain! {
+            if let ExprKind::Call(func, args) = expr.peel_drop_temps().kind;
+            if let ExprKind::Path(QPath::Resolved(_ty, path)) = func.kind;
+            if let Some(did) = path.res.opt_def_id();
+            then {
+                if !(match_def_path(cx, did, &paths::VEC_FROM_ELEM) && first_arg_is_arc_new(args)) { return; }
```
Serial-ATA, line 46:
```suggestion
        if let Some(higher::VecArgs::Repeat(elem, _)) = higher::VecArgs::hir(cx, expr) {
```

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// let v = vec![std::sync::Arc::new("some data".to_string()); 100];
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let data = std::sync::Arc::new("some data".to_string());
+    /// let v = vec![data; 100];
+    /// ```
```
Serial-ATA, line 26:
```suggestion
    /// ### What it does
    /// Checks for `Arc::new` in `vec![elem; len]`
    ///
    /// ### Why is this bad?
    /// This will create `elem` once and clone it `len` times - doing so with `Arc`
    /// is a bit misleading, as it will create references to the same pointer, rather
    /// than different `Arc` instances.
    ///
    /// ### Example
    /// ```rust
    /// let v = vec![std::sync::Arc::new("some data".to_string()); 100];
    /// ```
    /// Use instead:
    /// ```rust
    /// let data = std::sync::Arc::new("some data".to_string());
    /// let v = vec![data; 100];
    /// ```
```

I think this is a little better, correct me if I'm wrong :smile:

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+fn macro_is_vec(cx: &LateContext<'_>, macro_call: &MacroCall) -> bool {
+    cx.tcx.is_diagnostic_item(sym::vec_macro, macro_call.def_id)
+}
+
+fn first_arg_is_arc_new(args: &[Expr<'_>]) -> bool {
+    let Some(first_arg) = args.get(0) else { return false; };
+    if_chain! {
+        if let ExprKind::Call(func, _args) = first_arg.kind;
```
Serial-ATA, line 68:
```suggestion
    if_chain! {
    	if let Some(first_arg) = args.get(0);
        if let ExprKind::Call(func, _args) = first_arg.kind;
```

Serial-ATA:
Sorry for the single comments, forgot to start a review :smile:

clippy_lints/src/arc_new_in_vec_from_slice.rs
yonip23: nah, yours sounds much better 😬

clippy_lints/src/arc_new_in_vec_from_slice.rs
yonip23: exactly what i was looking for 👀

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// ```rust
+    /// let v = vec![std::sync::Arc::new("some data".to_string()); 100];
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let data = std::sync::Arc::new("some data".to_string());
+    /// let v = vec![data; 100];
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub ARC_NEW_IN_VEC_FROM_SLICE,
```
Serial-ATA, line 29:
```suggestion
    pub ARC_NEW_IN_VEC,
```

Forgot to suggest this, I'd omit the `from_slice` from the lint name

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,84 @@
...
+    "calling `Arc::new` in `vec![elem; len]`"
+}
+declare_lint_pass!(ArcNewInVecFromSlice => [ARC_NEW_IN_VEC_FROM_SLICE]);
+
+impl LateLintPass<'_> for ArcNewInVecFromSlice {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) else { return; };
+        if !macro_is_vec(cx, &macro_call) {
+            return;
+        }
```
Serial-ATA, line 40:
```suggestion
```
I don't think this check is needed anymore

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,84 @@
...
+fn macro_is_vec(cx: &LateContext<'_>, macro_call: &MacroCall) -> bool {
+    cx.tcx.is_diagnostic_item(sym::vec_macro, macro_call.def_id)
+}
+
+/// Checks whether the given `expr` is a call to `Arc::new`
+fn is_arc_new(expr: &Expr<'_>) -> bool {
+    if_chain! {
+        if let ExprKind::Call(func, _args) = expr.kind;
+        if let ExprKind::Path(ref func_path) = func.kind;
+        if let ExprKind::Path(QPath::TypeRelative(ty, _)) =  func.kind;
```
Serial-ATA, line 72:
```suggestion
        if let ExprKind::Path(ref func_path @ QPath::TypeRelative(ty, _)) = func.kind;
```

Serial-ATA (changes requested):

clippy_lints/src/arc_new_in_vec_from_slice.rs
```diff
@@ -0,0 +1,84 @@
...
+        if let TyKind::Path(ref ty_path) = ty.kind;
+
+        then {
+            let ty_segment = last_path_segment(ty_path);
+            let func_segment = last_path_segment(func_path);
+
+            return ty_segment.ident.name == sym::Arc && func_segment.ident.name == sym::new;
```
Serial-ATA, line 79:
```suggestion
        if let TyKind::Path(ref ty_path) = ty.kind;
		if let Some(def_id) = cx.qpath_res(qpath, ty.hir_id);
		if cx.tcx.is_diagnostic_item(sym::Arc, def_id);
        then {
            let func_segment = last_path_segment(func_path);

            return func_segment.ident.name == sym::new;
```
Missed one, you'll have to check if it's the diagnostic item, otherwise you'll get a FP with this:
```rust
struct Arc;
impl Arc {
    fn new() -> Self { Self }
}

let v = vec![Arc::new(); 2];
```

clippy_lints/src/arc_new_in_vec_from_slice.rs
yonip23: But its ok to have `vec![]` if its not `from_slice` - are you sure omitting it from the name is ok?

Serial-ATA: Maybe instead of `from_slice` it could be `from_elem`, since that makes it clear what it's checking for. The current name sounds like it's checking for `Vec::from(&[Arc])`.

yonip23: Ah, I see - will change to `from_elem`

clippy_lints/src/arc_new_in_vec_from_slice.rs
yonip23: nice catch!

Serial-ATA: You should also add a test for this :slightly_smiling_face:

yonip23: added this case as a test

clippy_lints/src/arc_new_in_vec_from_elem.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// ```rust
+    /// let data = std::sync::Arc::new("some data".to_string());
+    /// let v = vec![data; 100];
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub ARC_NEW_IN_VEC_FROM_ELEM,
+    suspicious,
+    "calling `Arc::new` in `vec![elem; len]`"
+}
+declare_lint_pass!(ArcNewInVecFromSlice => [ARC_NEW_IN_VEC_FROM_ELEM]);
```
Serial-ATA, line 33:
```suggestion
declare_lint_pass!(ArcNewInVecFromElem => [ARC_NEW_IN_VEC_FROM_ELEM]);
```

clippy_lints/src/arc_new_in_vec_from_elem.rs
```diff
@@ -0,0 +1,78 @@
...
+        if let TyKind::Path(ref ty_path) = ty.kind;
+        if let Some(def_id) = cx.qpath_res(ty_path, ty.hir_id).opt_def_id();
+        if cx.tcx.is_diagnostic_item(sym::Arc, def_id);
+
+        then {
+            let ty_segment = last_path_segment(ty_path);
+            let func_segment = last_path_segment(func_path);
+
+            return ty_segment.ident.name == sym::Arc && func_segment.ident.name == sym::new;
+        }
```
Serial-ATA, line 74:
```suggestion
        then {
            let func_segment = last_path_segment(func_path);

            return func_segment.ident.name == sym::new;
        }
```

Serial-ATA (changes requested):
Couple more things and it should be good :slightly_smiling_face:

clippy_lints/src/arc_new_in_vec_from_elem.rs
yonip23: > Couple more things and it should be good slightly_smiling_face


Thanks for the patience 🙂 im just getting used to the clippy concepts and api, sorry for being a bit quirky haha

clippy_lints/src/arc_new_in_vec_from_elem.rs
```diff
@@ -0,0 +1,77 @@
...
+
+fn yield_lint(cx: &LateContext<'_>, macro_call: &MacroCall) {
+    span_lint_and_help(
+        cx,
+        ARC_NEW_IN_VEC_FROM_ELEM,
+        macro_call.span,
+        "calling `Arc::new` in `vec![elem; len]`",
+        None,
+        "consider extracting `Arc` initialization to a variable",
+    );
```
Serial-ATA, line 57:
```suggestion
    span_lint_and_then(
        cx,
        ARC_NEW_IN_VEC_FROM_ELEM,
        macro_call.span,
        "calling `Arc::new` in `vec![elem; len]`",
        |diag| {
            diag.note("each `Arc` will point to the same allocation");
            diag.help("if this is intentional, consider extracting the `Arc` initialization to a variable");
        },
    );
```

Sorry, one more thing actually :smile:. A note here would be nice to explain what's wrong. Afterwards, could you squash the commits? It should be all good after that.

yonip23: sure thing, looks better

yonip23:
just saw that the commit message is `add arc_new_in_vec_from_slice lint` and not `from_elem` - is this ok? or should i re-commit?

Serial-ATA:
Not a big deal, you can `git commit --amend` if you want.

yonip23:
> Not a big deal, you can `git commit --amend` if you want.

Just force-pushed again
I think we're good to go now, right?

Serial-ATA:
I'd say so, just have to see if @xFrednet has any other thoughts :slightly_smiling_face:

Serial-ATA:
😅 Wasn't even thinking, could you also check for `Rc` and change the lint message accordingly? (https://github.com/rust-lang/rust-clippy/issues/8719#issuecomment-1114533408)

yonip23:
> sweat_smile Wasn't even thinking, could you also check for `Rc` and change the lint message accordingly? ([#8719 (comment)](https://github.com/rust-lang/rust-clippy/issues/8719#issuecomment-1114533408))

lol that makes sense, sure thing. Will get to it soon enough

yonip23:
@Serial-ATA regarding the name, i was thinking `implicit_reference_clone` or something similar to make it more generic in order to include `Rc` and `Arc` - wdyt?

yonip23:
well i took the liberty to rename - lmk if you're ok with that 😄

Serial-ATA:
Ah naming, the hardest part :smile:. How about `ref_ptr_in_vec_init`? It's not immediately obvious that `implicit_reference_clone` is only checking `vec![]`.

yonip23:
Uh, but `ref_ptr_in_vec_init` doesn't make it clear that the issue is with cloning the ref 😥
I'll probably get to work on it in the evening of my tz (now its morning here) - if we can't come up with a better name by then I'll change to your suggestion 🙂

yonip23:
Maybe `ref_clone_in_vec_init`?

Serial-ATA:
Yeah, I think that's better.

yonip23:
@Serial-ATA renamed 😃

tests/ui/ref_clone_in_vec_init/rc/custom_arc_strucrt.rs
```diff
@@ -0,0 +1,14 @@
+#![warn(clippy::ref_clone_in_vec_init)]
```
Serial-ATA, line 1: The file name says "arc" and "strucrt"

tests/ui/ref_clone_in_vec_init/arc/custom_arc_strucrt.rs
```diff
@@ -0,0 +1,14 @@
+#![warn(clippy::ref_clone_in_vec_init)]
```
Serial-ATA, line 1: "strucrt" in file name

Serial-ATA (commented):
Just some typos in the file names, but otherwise LGTM. Thanks for addressing my comments so quickly :slightly_smiling_face:.

yonip23:
Np!
Will fix it tomorrow (my tz)

xFrednet:
Thank you, @Serial-ATA for the review support, I highly appreciate it. :upside_down_face: This PR is on my todo list for the weekend :upside_down_face:

yonip23:
I rebased my fork from master and then force pushed - will squash after @xFrednet 's review 🙂

clippy_lints/src/ref_clone_in_vec_init.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) else { return; };
+        let Some(VecArgs::Repeat(elem, _)) = VecArgs::hir(cx, expr) else { return; };
+        let Some(symbol) = new_reference_call(cx, elem) else { return; };
+
+        yield_lint(cx, symbol, &macro_call);
+    }
+}
+
+fn yield_lint(cx: &LateContext<'_>, symbol: Symbol, macro_call: &MacroCall) {
```
xFrednet, line 49: This name is a bit weird to be, maybe `span_lint` or `emit_lint` is a bit better, as this doesn't really yields anything. :upside_down_face:

clippy_lints/src/ref_clone_in_vec_init.rs
```diff
@@ -0,0 +1,81 @@
...
+    suspicious,
+    "initializing `Arc` or `Rc` in `vec![elem; len]`"
+}
+declare_lint_pass!(RefCloneInVecInit => [REF_CLONE_IN_VEC_INIT]);
+
+impl LateLintPass<'_> for RefCloneInVecInit {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) else { return; };
+        let Some(VecArgs::Repeat(elem, _)) = VecArgs::hir(cx, expr) else { return; };
+        let Some(symbol) = new_reference_call(cx, elem) else { return; };
```
xFrednet, line 43: This is an interesting structure, most lint use multiple `if let` statements. Just an observation, this looks good. You can resolve this comment :+1:

clippy_lints/src/ref_clone_in_vec_init.rs
```diff
@@ -0,0 +1,81 @@
...
+    span_lint_and_then(
+        cx,
+        REF_CLONE_IN_VEC_INIT,
+        macro_call.span,
+        &format!("calling `{symbol_name}::new` in `vec![elem; len]`"),
+        |diag| {
+            diag.note(format!("each `{symbol_name}` will point to the same allocation"));
+            diag.help(format!(
+                "if this is intentional, consider extracting the `{symbol_name}` initialization to a variable"
+            ));
```
xFrednet, line 61: Here I would add a second help message saying:

`or if not, initilaize each element indivitually`

If you'd like, you can also go one step further and create actual suggestion snippets for each help message with `diag.span_sugg`. The message could look like the ones in the lint description. This is not necessary, but would be a nice addition. I'm also happy to merge this without :upside_down_face:

clippy_lints/src/ref_clone_in_vec_init.rs
```diff
@@ -0,0 +1,81 @@
...
+fn yield_lint(cx: &LateContext<'_>, symbol: Symbol, macro_call: &MacroCall) {
+    let symbol_name = symbol.as_str();
+
+    span_lint_and_then(
+        cx,
+        REF_CLONE_IN_VEC_INIT,
+        macro_call.span,
+        &format!("calling `{symbol_name}::new` in `vec![elem; len]`"),
+        |diag| {
+            diag.note(format!("each `{symbol_name}` will point to the same allocation"));
```
xFrednet, line 58: I believe this is a bit clearer

```suggestion
            diag.note(format!("each element will point to the same `{symbol_name}` instance"));
```

xFrednet (commented):
The implementation logic looks fantastic. I have a few comments about the documentation and naming, but nothing major. :upside_down_face:

yonip23:
@xFrednet nice catches, thanks for reviewing. I'll get to it in the next day or so 🙂

clippy_lints/src/ref_clone_in_vec_init.rs
yonip23: Ah, I got excited by the `let else` syntax available in nightly - where I work we use only stable 🥲

clippy_lints/src/ref_clone_in_vec_init.rs
yonip23: Since its not clear what exactly the user meant (there are 2 options), I think adding these suggestions will make the lint a bit tedious and hard to read
Wdyt?

xFrednet: Yeah, this is debatable. I, personally, like to make messages as informative as possible. Both suggestions could also be applied using an IDE. But it can make messages overwhelming. I've created a mockup:

```rs
error: calling `Rc::new` in `vec![elem; len]`
  --> $DIR/simple_case.rs:5:13
   |
LL |     let v = vec![Rc::new("x".to_string()); 2];
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D clippy::ref-clone-in-vec-init` implied by `-D warnings`
   = note: each element will point to the same `Rc` instance
help: if this is intentional, consider extracting the `Rc` initialization to a variable
   |
   |    let data = std::rc::Rc::new("some data".to_string());
   |    let v = vec![data; 100];
   |
help: ... or if not, initilaize each element indivitually
   |
   |    let mut v = Vec::with_capacity(100);
   |    for _ in 0..=100 {
   |        v.push(std::rc::Rc::new("some data".to_string()))
   |    }
   |
```

I feel like this is a bit more informative than just the help messages. In rust, I actually focus more on the code snippets and suggestions than the actual message TBH :sweat_smile:. But as said, this is optional :upside_down_face:

yonip23: When I'll get to it I'll try to implement it, lets see how it goes, and how it looks like with a bit more complex snippets 🙂

But since I'm not too familiar with this project, it'd be very kind of you to leave here some references to relevant examples / apis you think i should use to construct these snippets
Thanks ahead 🙏

xFrednet: There are several examples in Clippy, you can search for `diag.span_suggestion`. Here is a link to [one example](https://github.com/rust-lang/rust-clippy/blob/98c6381a387ac05e27c03ccfc64146bf4934f5c8/clippy_lints/src/ranges.rs#L368-L388). The basic idea is, that you take the span of the expressions and turn them into strings and format them with a template. So for this, the code could roughly look like this:

```rs
let count_snippet = snippet(cx, len_expr.span, ".."); // `len_expr` can be retrived from `VecArgs::hir(cx, expr)`
let data_snippet = snippet(cx, elem.span, "..");

// And now we just format this:
let mut loop_sugg = "{\n".to_string()
loop_sugg.push(&format!("    let mut v = Vec::with_capacity({count_snippet})\n"));
loop_sugg.push(&format!("    for _ in 0..={count_snippet} {\n"));
loop_sugg.push(&format!("        v.push({data_snippet})\n"));
loop_sugg.push(&format!("    }\n"));
loop_sugg.push(&format!("    v\n"));
loop_sugg.push(&format!("}\n"));
diag.span_suggestion(
    macro_call.span,
    "... or if not, initilaize each element indivitually",
    loop_sugg,
    Applicability::Unspecified,
)
```

this would now suggest replacing the macro with:
```rs
{
    let mut v = Vec::with_capacity(100);
    for _ in 0..=100 {
        v.push(std::rc::Rc::new("some data".to_string()))
    }
    v
}
```

Which would be valid replacement. We could also check if `data_snippet` spans over multiple lines and emit a placeholder if it's too long:

```rs
let mut data_snippet = snippet(cx, elem.span, "..");
if (data_snippet.contains('\n')) {
    // Use a placeholder for multipline element initializers
    data_snippet = "_".to_string();
}
```

(All of this code is untested, so you might have to do some adjustments)

yonip23: Thanks a lot, its more than enough 🙂

yonip23:
@xFrednet I didn't add the suggestions to the lint for now, since I encountered some difficulties there - wrote you over at Zulip 😅 
except for that, I fixed all your comments, and separated the lint renaming to a different commit so you can review the actual fixes more easily.

B.t.w, I think it can be useful to add a `dev` command to rename a lint - I renamed mine at least 3 times lol
something like `cargo dev rename_lint --old XXX --new YYY`

xFrednet:
Hey, thank you very much for the update! I'll have a look at it over the week. I've also answered you on Zulip :upside_down_face:

> B.t.w, I think it can be useful to add a dev command to rename a lint - I renamed mine at least 3 times lol something like cargo dev rename_lint --old XXX --new YYY

We have a similar tool for that (`cargo dev rename_lint`) it might be nice to add a dev mode to it, that doesn't register it as a renamed lint to rustc. And then also document this.

CC: @Jarcho (In case you're interested :upside_down_face: )

yonip23:
If I'll find some free time in the next few days I'll try to go for the suggestion implementation, but this pr should be ready to merge regardless, right? 🙂 
Worst case, i can open another pr with the additional changes

xFrednet (approved):
This version looks good to me, the suggestion update can be done in a followup PR as you suggested. I've also used lintcheck, and it found no lint trigger -> not false positive, which is nice. It would be awesome if you could squash the commits, as you offered :upside_down_face: 

The last thing, I'm unsure about is the lint group. I agree that this is suspicious, but on the other hand, it's not really wrong to use it. Having it warn-by-default might be a bit too strict. @rust-lang/clippy, @Alexendoo, @dswij what are your thought on the group, should this be suspicious or pedantic? :upside_down_face:

Jarcho:
I think `suspicious` is ok. I can't imagine anyone wanting a vec of the same pointer, but I can see someone making the mistake to use `vec![..]` for this.

Worst case is we change the category afterwards if it's too strict.

yonip23:
@xFrednet squashed, please take one last look to make sure i didn't f**k up anything while squashing, and lets merge it 🙂 
I'll keep working on the lint suggestion separately as we discussed

clippy_lints/src/rc_clone_in_vec_init.rs
```diff
@@ -0,0 +1,88 @@
...
+use rustc_span::{sym, Symbol};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `Arc::new` or `Rc::new` in `vec![elem; len]`
+    ///
+    /// ### Why is this bad?
+    /// This will create `elem` once and clone it `len` times - doing so with `Arc`
+    /// is a bit misleading, as it will create references to the same pointer, rather
+    /// than different `Arc` instances.
```
flip1995, line 17: The `Why is this bad?` still section only mentions `Arc`

clippy_lints/src/rc_clone_in_vec_init.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// let v = vec![std::sync::Arc::new("some data".to_string()); 100];
+    /// // or
+    /// let v = vec![std::rc::Rc::new("some data".to_string()); 100];
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// // Create the reference beforehand to clarify it that
+    /// // this reference should be cloned for each value
+    /// let data = std::rc::Rc::new("some data".to_string());
+    /// let v = vec![data; 100];
```
flip1995, line 30: Is this really a fix that we want to suggest? If so, I would put this example solution second and prefix it with "if you really want to clone the reference"

clippy_lints/src/rc_clone_in_vec_init.rs
```diff
@@ -0,0 +1,88 @@
...
+    ///
+    /// // Or initialize each value separately:
+    /// let mut data = Vec::with_capacity(100);
+    /// for _ in 0..=100 {
```
flip1995, line 34: This will create `101` entries.
```suggestion
    /// for _ in 0..100 {
```

flip1995 (commented):
`sus` as the category SGTM.

I have a few suggestions on how to imrpove the documentation a bit.

Also about the tests: Is there a technical reason why all of those tests have to be in their own file? We usually only separate into different files if the test would otherwise get way too big or if we have auto-fixable and non-auto-fixable suggestions in a lint.

I recommend putting all the tests in one file and put the separate test cases in different functions, that are named like the current file names.

yonip23:
@flip1995 I fixed the documentation, thanks for reviewing!
> Also about the tests: Is there a technical reason why all of those tests have to be in their own file? We usually only separate into different files if the test would otherwise get way too big or if we have auto-fixable and non-auto-fixable suggestions in a lint.

No technical reason, just makes more sense to me to separate each test to a different file. It's more of a personal preference.
Of course if you insist I can rearrange the tests as you asked 🙂

flip1995:
I would prefer to keep the tests in line with our other tests. I think it is easier to get an overview of what is tested and compare lint messages and suggestions when they are all in one file (as long as it doesn't get way to big).

I can see why you would prefer the split up test files and in the PR-view this all looks good. But I think later in the file-view, you would have to click/open too many files to see what tests are there and what tests might be missing.

yonip23:
Got you, np
I'll get to it in the next few days hopefully 🙂

yonip23:
@flip1995 I got to it quicker than I expected 😄 
I took the liberty to separate it into 2 files: `rc.rs` and `arc.rs`, instead of just one file. Hope you're ok with it, it made sense to me.
If not, I'll change it of course

flip1995:
Splitting it into Arc and Rc makes sense. I'm good with tests/documentation now. I didn't review the code, so I leave the final r+ to @xFrednet

xFrednet:
The implementation looks good to me, so everything should be ready since nothing was changed in that regard. Thank you, @flip1995, for the documentation/test review :heart: 

Could you squash your fix commits one last time @yonip23? :upside_down_face:

yonip23:
@xFrednet One last check to see everything is squashed correctly please and I think we're good to go 😬

xFrednet (approved):

xFrednet:
Everything looks good to me, and the squash is also alright :+1:. Thank you for the lint implementation. I hope you also had fun :upside_down_face:

@bors r+

PR #5505, opened 2020-04-22
Avoid running cargo+internal lints when not enabled

flip1995:
r? @matthiaskrgr 

changelog: none

Issue #3142, opened 2018-09-06
"fuzz" clippy with official rustc tests


matthiaskrgr:
It would be great if we had some semi-automatic way of running clippy on all the/most of the rust tests https://github.com/rust-lang/rust/tree/master/src/test/ I believe this will unearth a couple more crashes and/or bugs.

Maybe we can hack the rustc test executor program and clamp clippy-driver in-between somehow...

oli-obk:
OH YES

I want this very much. We want to do the same thing for miri, too. It should be rather simple to get done by setting https://github.com/rust-lang/rust/blob/master/src/tools/compiletest/src/common.rs#L134 to point to the `clippy-driver` binary.

matthiaskrgr:
status:

````bash
#!/bin/bash

for file in `ls -c1 | grep "\.rs$"`; do
	echo ${file}

	RUST_BACKTRACE=Full LD_LIBRARY_PATH=~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/ CLIPPY_TESTS=true  CARGO_MANIFEST_DIR="." ~/vcs/github/rust-clippy/target/debug/clippy-driver  -W clippy::all -W clippy::pedantic -W clippy::nursery ${file}  |& tee clippy.log
	rm -f `echo ${file} |  sed s/\.rs//`

done;
````

matthiaskrgr:
I hacked another small script that runs clippy on crates in the cargo cache
````bash
#!/bin/bash

for archive in `ls -c1 ~/.cargo/registry/cache/github.com-1ecc6299db9ec823/ | sort -n` ; do
echo checking $archive
mkdir /tmp/test_workdir
cd ~/.cargo/registry/cache/github.com-1ecc6299db9ec823/
cp ${archive} /tmp/test_workdir/
cd /tmp/test_workdir
#extract
tar -xvzf ${archive} >& /dev/null
cd ./*/

[...]

PR #8774, opened 2022-05-02
try reading rust-version from Cargo.toml

### LINKED ISSUE #7765: Support the `rust-version` field in `Cargo.toml` to specify the MSRV
The first crates are starting to make use of the `rust-version` field in `Cargo.toml`, which gets stabilized with the 1.56 version of rust, see https://github.com/rust-lang/rust/issues/65262 and https://github.com/rust-lang/cargo/pull/9732.

So far it does not seem like Clippy is able to understand this field. It only supports specifying it in it's own config file and unstably as an inner attribute ([Readme](https://github.com/rust-lang/rust-clippy#specifying-the-minimum-supported-rust-version)).

@rustbot label +C-enhancement

### LINKED ISSUE #8746: Suggestion: use rust-version field
### Description

There is a "new" field in the `Cargo.toml` that can be used to specify the minimum supported rust version: [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)

This feels somewhat redundant to the `msrv` in the `clippy.toml` file: https://github.com/rust-lang/rust-clippy/blob/master/README.md#specifying-the-minimum-supported-rust-version

If I see it correctly, clippy currently ignores the `rust-version` field from `Cargo.toml`.

Would it be a good approach to take that field into account when searching for a msrv?

### Version

```text
rustc 1.62.0-nightly (18f314e70 2022-04-24)
binary: rustc
commit-hash: 18f314e7027fe7084aaab8620c624a0d7bd29e70
commit-date: 2022-04-24
release: 1.62.0-nightly
LLVM version: 14.0.1
```

### Additional Labels

@rustbot modify labels: A-clippy T-cargo

hellow554:
Cargo.toml can contain a field `rust-version`, that acts like a MSRV of
clippy.toml file: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field
This will try to read that field and use it, if the clippy.toml config
has no `msrv` entry

changelog: respect `rust-version` from `Cargo.toml`

closes #8746
closes #7765


flip1995:
This sounds familiar. I think I also worked on that before :thinking: 

....

I did: https://github.com/rust-lang/cargo/pull/9967 and I claimed https://github.com/rust-lang/rust-clippy/issues/7765 back then. I probably forgot about finishing it :smile: 

I think one issue I couldn't resolve with this is that `MetadataCommand` is a quite expensive operation. This isn't really noticeable if compiling a single crate, but for example when running Clippy on rustc test suite, this can increase the time it takes immensely.  See https://github.com/rust-lang/rust-clippy/pull/5505#issuecomment-618425752 

cc @matthiaskrgr Do you still have a way to run Clippy on rustc's test suite to test this branch and if it brings back the issue linked above?

hellow554:
I really tried searching for a duplicate, I'm sorry :D
But thanks for bringing it up my attention.
Should I add a `closes #7765` as well?

flip1995:
> Should I add a `closes #7765` as well?

Yes, please.

clippy_lints/src/lib.rs
```diff
@@ -466,6 +467,23 @@ pub fn read_conf(sess: &Session) -> Conf {
...
+            Ok(metadata) => {
+                conf.msrv = metadata
+                    .packages
+                    .get(0)
+                    .and_then(|x| x.rust_version.as_ref())
+                    .and_then(|r| r.comparators.get(0))
+                    .map(|v| format!("{}.{}.{}", v.major, v.minor.unwrap_or(0), v.patch.unwrap_or(0)));
+            },
+            Err(e) => {
+                sess.struct_err(&format!("could not read cargo metadata: {}", e)).emit();
```
flip1995, line 482: I don't think this is a hard error. For example with `clippy-driver` outside of a `cargo` project, no `Cargo.toml` file can be found. I would just silently keep `conf.msrv = None` here.

flip1995 (commented):
I'm also not sure about the precedence. Should the `rust-version` in the `Cargo.toml` file be valued more or the one in the `clippy.toml` file? Should we deprecate the `msrv` conf option and point to the `rust-version` instead?

hellow554:
> I'm also not sure about the precedence. Should the `rust-version` in the `Cargo.toml` file be valued more or the one in the `clippy.toml` file? Should we deprecate the `msrv` conf option and point to the `rust-version` instead?

Good question. I thought, that the clippy.toml file should have a higher precedence, than the cargo.toml file. Deprecating it would be another idea, yes... But that's above my paygrade 😇

flip1995:
Let's talk about this in tomorrows meeting :+1:

mikerite:
> I think one issue I couldn't resolve with this is that `MetadataCommand` is a quite expensive operation. This isn't really noticeable if compiling a single crate, but for example when running Clippy on rustc test suite, this can increase the time it takes immensely. See [#5505 (comment)](https://github.com/rust-lang/rust-clippy/pull/5505#issuecomment-618425752)

Shouldn't Cargo provide rust-version as an environment variable? They already provide a lot of other manifest data in that way - [cargo docs](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates).

We already use CARGO_PRIMARY_PACKAGE in the driver.
 https://github.com/rust-lang/rust-clippy/blob/95f8b2600240022f0a82e703b35d4bc14a42eb1b/src/driver.rs#L338

flip1995:
We discussed this in today's meeting. We came to the agreement that:

- We shouldn't deprecate Clippy's MSRV config option (yet)
- The clippy.toml MSRV config option should have precedence.
- If both `clippy.toml` MSRV and `Cargo.toml` `rust-version` are defined, but differ, a warning should be emitted.

I think the current implementation covers the first two points. Do you mind adding the warning? If you need help with that let me know.

Also we should answer the perf question and if `cargo` providing this through a env var is and option. Maybe we can even get that from `rustc` itself?

hellow554:
Thanks for the discussion on this topic.

Currently, I only use the expensive MetadataCommand call if clippy.toml hasn't that `msrv` option set. If I want to check them both, I need to make that call every time. I can do that, but that will very much likely make things worse.

I can open an issue/pr @ https://github.com/rust-lang/cargo to insert a `CARGO_RUST_VERSION` field to the environment.  
I'm not sure what you mean by "getting it from `rustc` itself"? `rustc` doesn't read a `Cargo.toml` IIRC, only cargo does.

flip1995:
> If I want to check them both, I need to make that call every time. I can do that, but that will very much likely make things worse.

Not really, because the issue only occurs if you check multiple single files, not a cargo project. But if you don't have a `Cargo.toml` for those single files you also won't have a `clippy.toml`. Maybe it's a none-issue. Let me check that.



> 'm not sure what you mean by "getting it from `rustc` itself"? `rustc` doesn't read a `Cargo.toml` IIRC, only cargo does.

For some reason I had in mind that a `#[rust-version]` attribute exists and that therefore rustc must have some information about that. But that's not the case, so nevermind.

flip1995:
Checking this with flamegraph, the `MetadataCommand` takes up 3-4% of the runtime. 

Looking at `cargo uitest` this has some more impact:

Without this patch:
```console
cargo uitest > /dev/null  75.74s user 26.33s system 886% cpu 11.511 total
cargo uitest > /dev/null  75.96s user 26.91s system 891% cpu 11.545 total
cargo uitest > /dev/null  76.83s user 26.69s system 879% cpu 11.768 total
```

With this patch:
```console
cargo uitest > /dev/null  90.80s user 30.94s system 883% cpu 13.785 total
cargo uitest > /dev/null  91.13s user 30.94s system 880% cpu 13.864 total
cargo uitest > /dev/null  91.20s user 31.08s system 892% cpu 13.697 total
```

It isn't too bad, but definitely a noticable regression.

I think creating an issue in cargo about the env var or (maybe better) asking on Zulip about it would be good. I'm open to merge this with `MetadataCommand` if the cargo team doesn't want to add the env var.

hellow554:
Thanks for the measurement!
I'm not really firm with zulip, would you mind asking there? I would try to prepare a PR for that?

flip1995:
Zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/env.20var.20for.20rust-version.20metadata/near/281135045

flip1995:
If no such env var should get added we could just check if `CARGO_MANIFEST_DIR` is set before running `MetadataCommand`. That way we can be sure, that Clippy is run through cargo and not through `clippy-driver`. But that's only a workaround.

hellow554:
So I did things™.

Instead of using `CargoMetadata` I use what [`@bjorn3` suggested](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/env.20var.20for.20rust-version.20metadata/near/281135215) and reading the `$CARGO_MANIFEST_DIR/Cargo.toml` file directly as toml file and parse the `package.rust-version` field from it.
This should give a little speed advantage over the `cargo metadata` call.

What I wanted next, but couldn't achieve because of the way the `Conf` struct is created, is to have the `RustcVersion` directly, instead of storing a string and parsing it all over again. But for this to work I either need to implement `Deserialize` directly for that struct (looking @flip1995 ;) ) or finding a clever workaround...

hellow554:
I really would like to add some tests, but I'm not really sure how to integrate them into the clippy testsuite. After all, I would like to have at least 7 different testcases and therefore 7 different cargo projects.

| file-attr | clippy | cargo |
|----------|-------|--------|
| None    | None  | None  |
| None    | None  | Some |
| None    | Some  | None |
| ...

clippy_lints/src/lib.rs
```diff
@@ -455,15 +477,38 @@ pub fn read_conf(sess: &Session) -> Conf {
...
+            ));
+            None
+        })
+    });
+    if let Some(cargo_msrv) = cargo_toml_msrv() {
+        if let Some(clippy_msrv) = clippy_msrv {
+            // if both files have an msrv, let's compare them and emit a warning if they differ
+            if clippy_msrv != cargo_msrv {
+                sess.warn(&format!(
+                    "The MSRV in `clippy.toml` and `Cargo.toml` differ. Using `{}`",
```
flip1995, line 504:
```suggestion
                    "the MSRV in `clippy.toml` and `Cargo.toml` differ; using `{}`",
```

clippy_lints/src/lib.rs
```diff
@@ -571,16 +616,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
...
-        parse_msrv(s, None, None).or_else(|| {
-            sess.err(&format!(
-                "error reading Clippy's configuration file. `{}` is not a valid Rust version",
-                s
-            ));
-            None
-        })
-    });
-
+    let msrv = conf.msrv.as_ref().and_then(|s| parse_msrv(s, None, None));
```
flip1995, line 619: Why do you have to re-parse it and move the original parsing code to the `read_conf` file.

Couldn't you do something like:

```rust
let mut msrv = <original-parsing-code>;
if let Some(cargo_msrv) = parse_cargo_msrv() {
    if msrv.map_or(false, |msrv| msrv != cargo_msrv) {
        // emit warning
    } else {
        // they are either identical => doesn't matter which
        // or `msrv == None` => use cargo_msrv
        msrv = Some(cargo_msrv)
    }
}
```

flip1995 (commented):
> But for this to work I either need to implement Deserialize directly for that struct (looking @flip1995 ;) ) or finding a clever workaround

I intentionally wrote the `rustc-version` crate without deps. So adding `serde` for this isn't really an option.

You can add tests for this in the `tests/ui-toml` or `tests/ui-cargo` dirs.

flip1995:
> Instead of using `CargoMetadata` I use what [`@bjorn3` suggested](https://rust-lang.zulipchat.com/#narrow/stream/246057-t-cargo/topic/env.20var.20for.20rust-version.20metadata/near/281135215) and reading the `$CARGO_MANIFEST_DIR/Cargo.toml` file directly as toml file and parse the `package.rust-version` field from it.

Not sure if we should re-parse it by hand or just check for the `CARGO_MANIFEST_DIR` env var and then call `MetadataCommand`.

hellow554:
> Not sure if we should re-parse it by hand or just check for the `CARGO_MANIFEST_DIR` env var and then call `MetadataCommand`.

I mean, in the end, the two options should be equal in their outcome.  
Where they can differ is the speed. As said, I avoided using `MetadataCommand`, because of the speed concerns. Parsing a toml file should be relativly easy and in the end, I'm doing the same as `cargo` does, am I right? :/

clippy_lints/src/lib.rs
flip1995: I think I get your intention. However, I don't really understand why it is necessary to update the `conf.msrv` value. This value is unused after the `let msrv = <conf.msrv parsing code>;` statement AFAIK.

flip1995:
> Where they can differ is the speed. As said, I avoided using `MetadataCommand`, because of the speed concerns. Parsing a toml file should be relativly easy and in the end, I'm doing the same as `cargo` does, am I right? :/

Yeah I guess this is only style preference in the end. I don't think the perf difference matters much when linting a whole crate with `cargo clippy`. The perf concern is only with `clippy-driver single-file.rs`.

matthiaskrgr:
@flip1995  sorry, just randomly saw the mention (I'm getting like 30 github mails a day and don't check them all tbh :sweat: )
I think the iterations were tacked in here: https://github.com/rust-lang/rust-clippy/issues/3142 
(from bash for-loop to a semi-fledged rust program (icemaker))
icemaker has a `--clippy` mode but the setup might be a bit tricky (you'll need RTIM master toolchain + clippy installed with the name "master" and systemd-run at least)

clippy_lints/src/lib.rs
hellow554: I checked that, you're right. It's only used once, so I reworked it to your suggestion.

hellow554:
I did the suggested changes and rebased everything into one commit.

Hope this is ready for merging :)

clippy_lints/src/lib.rs
```diff
@@ -443,6 +447,50 @@ pub fn register_pre_expansion_lints(store: &mut rustc_lint::LintStore, sess: &Se
...
+    if let Some(cargo_msrv) = cargo_msrv {
+        if let Some(clippy_msrv) = clippy_msrv {
+            // if both files have an msrv, let's compare them and emit a warning if they differ
+            if clippy_msrv != cargo_msrv {
+                sess.warn(&format!(
+                    "the MSRV in `clippy.toml` and `Cargo.toml` differ; using `{}`",
+                    clippy_msrv
+                ));
+                return Some(clippy_msrv);
+            }
```
flip1995, line 486:
```suggestion
            }
            return Some(clippy_msrv);
```

I think the code is a bit more readable if the `return` is moved out of the error case.

flip1995 (commented):
LGTM. Do you want to add some tests to `tests/ui-cargo`?

hellow554:
Sadly I'm not ablve to use `CARGO_MANIFEST_DIR`, because during `cargo uitest` it points the clippy root:
```none
[clippy_lints/src/lib.rs:477] std::env::var("CARGO_MANIFEST_DIR") = Ok(
    "/home/marcel/projects/clippy",
)
```

I'm using the `MetadataCommand` now as it seems to work properly. Maybe somebody has a different idea :)

Working on the ui tests now.

hellow554:
I choose the [`use_self`](https://rust-lang.github.io/rust-clippy/master/index.html#use_self) for testing, because it's one of the easiest one to test.

Take your time to review it, I'm hopefully done implementing this :)

It all works on my side :)

clippy_lints/src/lib.rs
hellow554: I noticed, that I can omit the `return` entirely, thanks to your suggestion.

Makes it somewhat more readable, you're right

hellow554:
Solved the merge conflict. Ready for review again ;)

flip1995:
I opened https://github.com/rust-lang/cargo/pull/10713 Let's try to do this with the env var.

hellow554:
Thanks flip for the PR. It finally has been merged :)

I updated my approach to use the env var.

Now we need to wait, that cargo is updated in rust nightly. I will try to observe it and make a rebase once it has been landed.

flip1995:
Thanks! Marking this as blocked in the meantime.

r? @flip1995

hellow554:
Meh, I saw a cargo update in the rust repo, but it didn't include your commit. Sorry, let's keep waiting :)

flip1995:
We would need to bump the nightly clippy version in a sync anyway, before you'd be able to use that in this repo. So you only have to check every other week. (next sync is in a week from now).

hellow554:
Alright, didn't know that. :)

I bumbed it locally, so I got that going for me, but yeah. Let's wait a week, it doesn't matter at this point anymore ;)

hellow554:
Hey @flip1995 I need your help again.

I seem to be stuck to the same problem as last time, when I tried to parse `Cargo.toml` manually, namely that when running tests I don't get the rust-version from the tests `Cargo.toml`, but from the project one.

Running `pass_both_same` yields an empty string, but when I change the `Cargo.toml` file in the clippy root directory to something like `1.62.51` I get exactly that, but not the one from the tests `Cargo.toml` file.

Is there a way around this? Else I would have to use `cargo-metadata` again and the whole point failed :(  
Is that a problem in cargo uitest?

flip1995:
We may want to look at the UI-tests and if we can change something there to test it. We should definitely use the env var and not `cargo_metadata`. I'll have to look more in depth into that though...

mikerite:
Would making a UI test with [compiletest header](https://rustc-dev-guide.rust-lang.org/tests/headers.html#environment-variable-headers) to set the environment variable work?

hellow554:
I think I finally found the solution. It passes on my machine™.

Problem is, that compiletest_rs does not parse the `Cargo.toml` file in the tests folder itself.
I see two options: either use the compiletest-header as @mikerite suggests, or, to stick with the current theme, we just parse the `Cargo.toml` file ourself with the toml crate and set the environment variable ourself.

I know, this isn't a perfect solution, but this is the best I can do :)

tests/ui-cargo/multiple_config_files/warn/src/main.stderr
```diff
@@ -1,2 +1,2 @@
-Using config file `$SRC_DIR/tests/ui-cargo/multiple_config_files/warn/.clippy.toml`
-Warning: `$SRC_DIR/tests/ui-cargo/multiple_config_files/warn/clippy.toml` will be ignored.
+Using config file `$SRC_DIR/.clippy.toml`
+Warning: `$SRC_DIR/clippy.toml` will be ignored.
```
hellow554, line 2: This is a "fallout" by setting `CARGO_MANIFEST_DIR`, but I think this **is** the proper way

Cargo.toml
```diff
@@ -52,6 +52,7 @@ futures = "0.3"
 parking_lot = "0.12"
 tokio = { version = "1", features = ["io-util"] }
 rustc-semver = "1.1"
+toml = "0.5"
```
flip1995, line 55: NIT: This is not a UI test dep in that sense. So I would group it with the other test dependencies in lines 31-35.

README.md
```diff
@@ -214,6 +214,13 @@ specifying the minimum supported Rust version (MSRV) in the clippy configuration
 msrv = "1.30.0"
 ```
 
+Alternatively, the [`rust-version` field](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field)
+in the `Cargo.toml` can be used.
+
+```toml
```
flip1995, line 220:
```suggestion
```toml
# Cargo.toml
```

Let's be very clear in which file this has to be added.

clippy_lints/src/lib.rs
```diff
@@ -450,6 +451,39 @@ pub fn register_pre_expansion_lints(store: &mut rustc_lint::LintStore, sess: &Se
...
+            None
+        })
+    });
+
+    if let Some(cargo_msrv) = cargo_msrv {
+        if let Some(clippy_msrv) = clippy_msrv {
+            // if both files have an msrv, let's compare them and emit a warning if they differ
+            if clippy_msrv != cargo_msrv {
+                sess.warn(&format!(
+                    "the MSRV in `clippy.toml` and `Cargo.toml` differ; using `{}`",
```
flip1995, line 473:
```suggestion
                    "the MSRV in `clippy.toml` and `Cargo.toml` differ; using `{}` from `clippy.toml`",
```

tests/ui-cargo/cargo_rust_version/fail_file_attr/src/main.rs
```diff
@@ -0,0 +1,13 @@
+#![feature(custom_inner_attributes)]
+#![clippy::msrv = "1.58.0"]
```
flip1995, line 2: So, this doesn't produce a warning. That is not optimal, but I wouldn't try to address this in this PR.

Can you add a `FIXME`, that this should produce some kind of warning, that it overrides the value from one of the toml files?

flip1995 (commented):
Impl LGTM and also the solution to the tests. Only some minor things left.

tests/ui-cargo/cargo_rust_version/fail_file_attr/src/main.rs
hellow554: Maybe abandon that attribute in the long run? ;)

hellow554:
Done

hellow554:
CI is green and I think this is (finally) ready to merge

PR #8778, opened 2022-05-02
Fix `cast_lossless` to avoid warning on `usize` to `f64` conversion.

### LINKED ISSUE #3689: On 32-bit, usize as f64 recommends f64::from(usize)
On i686, the cast_lossless lint recommends using `f64::from(len)` instead of `len as f64` where `len` is of type `usize`. This would <del>change the behaviour on x86_64</del> fail to compile.

sunfishcode:
Previously, the `cast_lossless` lint would issue a warning on code that
converted a `usize` value to `f64`, on 32-bit targets.

`usize` to `f64` is a lossless cast on 32-bit targets, however there is
no corresponding `f64::from` that takes a `usize`, so `cast_lossless`'s
suggested replacement does not compile.

This PR disables the lint in the case of casting from `usize` or `isize`.

Fixes #3689.

changelog: [`cast_lossless`] no longer gives wrong suggestion on usize,isize->f64

### DIFF
--- a/clippy_lints/src/casts/cast_lossless.rs
+++ b/clippy_lints/src/casts/cast_lossless.rs
@@ -93,7 +93,7 @@ fn should_lint(
             } else {
                 64
             };
-            from_nbits < to_nbits
+            !is_isize_or_usize(cast_from) && from_nbits < to_nbits
         },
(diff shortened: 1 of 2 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8780, opened 2022-05-03
Ignore type aliases in `init_numbered_fields`

### LINKED ISSUE #8638: init_numbered_fields incorrectly triggers for type alias of tuple struct
### Summary

The clippy lint `init_numbered_fields` introduced in 1.59.0 is emitted with an incorrect suggested fix when the struct initialization is for a type alias to a tuple or unit struct.

This case should be omitted from the lint entirely, as there is no means other than use of numbered fields to initialize a tuple struct for which only the type alias can be reliably known, because the type alias cannot be used as a constructor.

### Lint Name

_No response_

### Reproducer

I tried this code:

```rust
fn main() {
    let _ = TypeAlias { 0: 0 };
}


struct AkwardlyNamedAndPossiblyCfgGatedTupleStruct(i32);

type TypeAlias = AkwardlyNamedAndPossiblyCfgGatedTupleStruct;
```

I ran `cargo clippy`

I saw this happen:

```
warning: used a field initializer for a tuple struct
 --> src/main.rs:2:13
  |
2 |     let _ = TypeAlias { 0: 0 };
  |             ^^^^^^^^^^^^^^^^^^ help: try this instead: `TypeAlias(0)`
  |
  = note: `#[warn(clippy::init_numbered_fields)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#init_numbered_fields

warning: `test_clippy_error` (bin "test_clippy_error") generated 1 warning
```

```
<no clippy warning>
```
because "A type alias to a tuple-struct or unit-struct cannot be used to qualify that type's constructor:", this should not trigger when the tuple struct initialization is being invoked through a type alias.

### Version

```text
binary: rustc
commit-date: 2022-02-23
host: x86_64-apple-darwin
release: 1.59.0
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Alexendoo:
changelog: Ignore type aliases in [`init_numbered_fields`]

Fixes #8638

### DIFF
--- a/clippy_lints/src/init_numbered_fields.rs
+++ b/clippy_lints/src/init_numbered_fields.rs
@@ -1,6 +1,7 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet_with_applicability;
 use rustc_errors::Applicability;
+use rustc_hir::def::{DefKind, Res};
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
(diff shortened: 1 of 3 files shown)


flip1995:
@bors r+

Thanks!

(steeling this back :D)

Issue #8782, opened 2022-05-03
Improve only_used_in_recursion lint/testing


flip1995:
I tried to read the code of the lint for over an hour and couldn't really follow what it is doing. I was also able to remove large junks of code from the visitor and the tests kept passing. This signals to me that the logic the lint code is following isn't really tested. 

The tests that are there could mostly also be linted with simpler linting code. Also there is a test, that with slight modifications produces a FP:

https://github.com/rust-lang/rust-clippy/blob/b3bd03afcda90ed458ecf95cf46798c13a3817d3/tests/ui/only_used_in_recursion.rs#L28-L38

If you add a `if c >= 10 { return 4; }` at the top of the function, you still get the lint. But removing the `b` argument changes the semantics of the function (i.e. you won't be able to get the same behavior). [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f13c4bb8081bd3597a5780b10ce3bc11)

I think we should just bail out if the parameter is used anywhere other than directly as its own function argument.

I also have the suspicion that this lint might be a significant perf regression for Clippy.

I think we should move this lint to nursery and not run it if the lint is allowed for now. I like this lint in general, but I think in the current state it is over engineered and not really maintainable.

_Originally posted by @flip1995 in https://github.com/rust-lang/rust-clippy/issues/8689#issuecomment-1096693028_

PR #8783, opened 2022-05-03
Move only_used_in_recursion to nursery

flip1995:
r? @llogiq 

I still think this is the right thing to do for this lint. See: #8782

I want to get this merged before the sync on Thursday, because I want to backport this and the last fix #8691 to beta.

changelog: Move [`only_used_in_recursion`] to nursery


llogiq:
Ok. @bors r+

PR #8789, opened 2022-05-05
Added missing `### What it does`

### LINKED ISSUE #8781: lint `cast_slice_different_sizes` has missing `### What it does` and my tool crashes again
### Description

if you look at the json for this lint
```json
  {
    "id": "cast_slice_different_sizes",
    "id_span": {
      "path": "src/casts/mod.rs",
      "line": 454
    },
    "group": "correctness",
    "level": "deny",
    "docs": " Checks for `as` casts between raw pointers to slices with differently sized elements.\n\n### Why is this bad?\nThe produced raw pointer to a slice does not update its length metadata. The produced\npointer will point to a different number of bytes than the original pointer because the\nlength metadata of a raw slice pointer is in elements rather than bytes.\nProducing a slice reference from the raw pointer will either create a slice with\nless data (which can be surprising) or create a slice with more data and cause Undefined Behavior.\n\n### Example\n// Missing data\n```rust\nlet a = [1_i32, 2, 3, 4];\nlet p = &a as *const [i32] as *const [u8];\nunsafe {\n    println!(\"{:?}\", &*p);\n}\n```\n// Undefined Behavior (note: also potential alignment issues)\n```rust\nlet a = [1_u8, 2, 3, 4];\nlet p = &a as *const [u8] as *const [u32];\nunsafe {\n    println!(\"{:?}\", &*p);\n}\n```\nInstead use `ptr::slice_from_raw_parts` to construct a slice from a data pointer and the correct length\n```rust\nlet a = [1_i32, 2, 3, 4];\nlet old_ptr = &a as *const [i32];\n// The data pointer is cast to a pointer to the target `u8` not `[u8]`\n// The length comes from the known length of 4 i32s times the 4 bytes per i32\nlet new_ptr = core::ptr::slice_from_raw_parts(old_ptr as *const u8, 16);\nunsafe {\n    println!(\"{:?}\", &*new_ptr);\n}\n```",
    "version": "1.60.0",
    "applicability": {
      "is_multi_part_suggestion": false,
      "applicability": "Unresolved"
    }
  },
```
on the key `docs` the string is missing a ` ### What it does\n` at the beginning
you can also see that on the website: https://rust-lang.github.io/rust-clippy/master/#cast_slice_different_sizes 

every other lint has ` ### What it does\n` in the beggining,
please fix it.

json file is from https://raw.githubusercontent.com/rust-lang/rust-clippy/gh-pages/master/lints.json

### Version

_No response_

### Additional Labels

_No response_

fedemartinezdev:
Adds the missing line to ``[`cast-slice-different-sizes`]`` lint documentation
fixes #8781 

changelog: none

### DIFF
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -413,6 +413,7 @@ declare_clippy_lint! {
 }
 
 declare_clippy_lint! {
+    /// ### What it does
     /// Checks for `as` casts between raw pointers to slices with differently sized elements.
     ///
     /// ### Why is this bad?


flip1995:
@bors r+

Thanks!

PR #8792, opened 2022-05-06
Suggest -Zunpretty=ast-tree instead of -Zast-json

jyn514:
-Zast-json is being removed shortly: https://github.com/rust-lang/rust/pull/85993.
ast-tree does essentially the same thing, and still works today even before that PR lands.

changelog: none


flip1995:
@bors r+

Thanks!

PR #8794, opened 2022-05-06
Address `unnecessary_to_owned` false positive

### LINKED ISSUE #8759: unnecessary_to_owned false positive
### Summary

```Rust
#[derive(Default)]
struct View {
    
}

impl std::borrow::ToOwned for View {
	type Owned = View;
	fn to_owned(&self) -> Self::Owned {
		View {}
	}
}

#[derive(Default)]
struct RenderWindow {
    default_view: View,
}

impl RenderWindow {
    fn default_view(&self) -> &View {
        &self.default_view
    }
    fn set_view(&mut self, _view: &View) {
		
    }
}

fn main() {
	let mut rw = RenderWindow::default();
	rw.set_view(&rw.default_view().to_owned());
}
```

Clippy output:

```
warning: unnecessary use of `to_owned`
  --> src/main.rs:29:14
   |
29 |     rw.set_view(&rw.default_view().to_owned());
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `rw.default_view()`

```
If we apply the suggestion, rustc output:
```
error[E0502]: cannot borrow `rw` as mutable because it is also borrowed as immutable
  --> src/main.rs:29:2
   |
29 |     rw.set_view(rw.default_view());
   |     ^^^--------^-----------------^
   |     |  |        |
   |     |  |        immutable borrow occurs here
   |     |  immutable borrow later used by call
   |     mutable borrow occurs here
   ```

### Version

```text
rustc 1.62.0-nightly (69a5d2481 2022-04-27)
binary: rustc
commit-hash: 69a5d2481e856a5a18885390b8cf6950b9ff8dd3
commit-date: 2022-04-27
release: 1.62.0-nightly
LLVM version: 14.0.1
```

smoelius:
My proposed fix for #8759 is to revise the conditions that delineate `redundant_clone` and `unnecessary_to_owned`:
```rust
        // Only flag cases satisfying at least one of the following three conditions:
        // * the referent and receiver types are distinct
        // * the referent/receiver type is a copyable array
        // * the method is `Cow::into_owned`
        // This restriction is to ensure there is no overlap between `redundant_clone` and this
        // lint. It also avoids the following false positive:
        //  https://github.com/rust-lang/rust-clippy/issues/8759
        //   Arrays are a bit of a corner case. Non-copyable arrays are handled by
        // `redundant_clone`, but copyable arrays are not.
```
This change causes a few cases that were previously flagged by `unnecessary_to_owned` to no longer be flagged. But one could argue those cases would be better handled by `redundant_clone`.

Closes #8759

changelog: none

### DIFF
--- a/clippy_lints/src/methods/unnecessary_to_owned.rs
+++ b/clippy_lints/src/methods/unnecessary_to_owned.rs
@@ -65,13 +65,12 @@ fn check_addr_of_expr(
         if let Some(parent) = get_parent_expr(cx, expr);
         if let ExprKind::AddrOf(BorrowKind::Ref, Mutability::Not, _) = parent.kind;
         let adjustments = cx.typeck_results().expr_adjustments(parent).iter().collect::<Vec<_>>();
-        if let Some(target_ty) = match adjustments[..]
-        {
+        if let
             // For matching uses of `Cow::from`
             [
                 Adjustment {
                     kind: Adjust::Deref(None),
-                    ..
+                    target: referent_ty,
                 },
                 Adjustment {
                     kind: Adjust::Borrow(_),
@@ -82,7 +81,7 @@ fn check_addr_of_expr(
             | [
                 Adjustment {
                     kind: Adjust::Deref(None),
-                    ..
+                    target: referent_ty,
                 },
                 Adjustment {
                     kind: Adjust::Borrow(_),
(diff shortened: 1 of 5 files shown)


llogiq:
That sounds sensible. Thank you!

@bors r+

Issue #1781, opened 2017-05-25
Lints to suggest deriving more traits


clarcharr:
A few suggestions:

1. If a type derives `Eq`, it should also derive `Hash`.
2. If a type is trivial (e.g. a unit struct or an enum of only unit variants) we should suggest to implement `Copy`, `Eq`, `Default`, `Hash`, and their dependencies.
3. If a type derives `PartialEq` but could also derive `Eq`, suggest deriving `Eq`.
4. Similar to 3 but with `Ord`.

oli-obk:
`Copy` is special, because it changes the semantics of the type instead of just making it more usable.

clarcharr:
My thought process is that if you can already essentially "copy" a type by trivially constructing new values, it makes sense to just implement `Copy`.

Also adding to this list: implementing `Ord` for unit structs.

oli-obk:
@clarcharr I misunderstood, I thought you meant `struct Foo(());`, too.

nsunderland1:
| 3. If a type derives `PartialEq` but could also derive `Eq`, suggest deriving `Eq`.

I'm interested in taking this one on (and maybe `Ord` too, as far as I can tell from playing around it should be pretty much the same code), Can open a dedicated issue if you think it's needed.

xFrednet:
Hey @nsunderland1, I don't think it's necessary to create a new issue. You can also assign yourself to this one with `@rustbot claim` and later release the assignment :upside_down_face:

nsunderland1:
Sounds good :)

@rustbot claim

nsunderland1:
@rustbot release

#8796 implemented the `Eq` lint. `Ord` should be very similar, but happy to let someone else take that (or any of the other suggestions from OP) up.

arnolddevos:
I think there is a problem with assuming a type should derive `Eq` just because its fields do.  It depends on the intended semantics of the type.   Imagine a type `Id` with a special value `NULLID` that is by design not equal to itself.  `Id` should derive `PartialEq` but not `Eq`.   However, it might happen that the fields of `Id` derive `Eq`.

```rust
pub struct Id(u32, u32);
const NULLID: Id = Id(0,0);
assert!(NULLID != NULLID);
```

Jarcho:
If `PartialEq` were derived there then that assert would fail. Deriving is the same as structural equality.

arnolddevos:
[...]

PR #8796, opened 2022-05-07
New lint: [`derive_partial_eq_without_eq`]

nsunderland1:
Introduces a new lint, [`derive_partial_eq_without_eq`]. 

See: #1781 (doesn't close it though).

changelog: add lint [`derive_partial_eq_without_eq`]


clippy_lints/src/derive.rs
```diff
@@ -156,11 +156,44 @@ declare_clippy_lint! {
...
+    /// ```rust
+    /// #[derive(PartialEq, Eq)]
+    /// struct Foo {
+    ///     i_am_eq: i32,
+    ///     i_am_eq_too: Vec<String>,
+    /// }
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub DERIVE_PARTIAL_EQ_WITHOUT_EQ,
+    nursery,
```
nsunderland1, line 187: Wasn't sure about the right place to put this.

clippy_lints/src/derive.rs
```diff
@@ -419,3 +453,43 @@ impl<'tcx> Visitor<'tcx> for UnsafeVisitor<'_, 'tcx> {
...
+    if_chain! {
+        if let ty::Adt(adt, substs) = ty.kind();
+        if peq_is_automatically_derived;
+        if let Some(eq_trait_def_id) = cx.tcx.get_diagnostic_item(sym::Eq);
+        if let Some(def_id) = trait_ref.trait_def_id();
+        if cx.tcx.is_diagnostic_item(sym::PartialEq, def_id);
+        then {
+            if !implements_trait(cx, ty, eq_trait_def_id, substs) {
+                // If all of our fields implement `Eq`, we can implement `Eq` too
+                for variant in adt.variants() {
```
nsunderland1, line 474: This section is based off [can_type_implement_copy](https://github.com/rust-lang/rust/blob/37b55c8a0cafdb60b9168da34f904acc70157df8/compiler/rustc_trait_selection/src/traits/misc.rs#L19) from rustc. Only thing I wasn't super sure about is that that function does a [check](https://github.com/rust-lang/rust/blob/37b55c8a0cafdb60b9168da34f904acc70157df8/compiler/rustc_trait_selection/src/traits/misc.rs#L490) for `ty.references_error()`, it wasn't sure clear to me if that was necessary here (or indeed what exactly it does)

clippy_lints/src/derive.rs
```diff
@@ -419,3 +453,43 @@ impl<'tcx> Visitor<'tcx> for UnsafeVisitor<'_, 'tcx> {
...
+                    for field in &variant.fields {
+                        let ty = field.ty(cx.tcx, substs);
+
+                        if !implements_trait(cx, ty, eq_trait_def_id, substs) {
+                            return;
+                        }
+                    }
+                }
+
+                span_lint_and_help(
```
nsunderland1, line 484: I tried making this work with a suggestion like this:

```rust
span_lint_and_sugg(
    cx,
    DERIVE_PARTIAL_EQ_WITHOUT_EQ,
    span,
    "you are deriving `PartialEq` and can implement `Eq`",
    "consider deriving `Eq` as well",
    "PartialEq, Eq".to_string(),
)
```

But when I ran this it just wouldn't print a suggestion at all, not sure what I did wrong.

clippy_lints/src/derive.rs
Jarcho: I'm fairly sure you can ignore that. By the time late lint passes run the code already has no errors.

clippy_lints/src/derive.rs
Jarcho: What did it print? That should have worked.

nsunderland1: So when I switch to `span_lint_and_sugg` on my branch, here's the stderr diff `cargo uitest` gives me: https://pastebin.com/v7Eg5kDp

Note that the only difference is that the ``help: consider deriving `Eq` as well`` is dropped.

Here's the JSON that `cargo uitest` spat out for one of the errors: https://pastebin.com/Gr6w967J

Looks like `suggested_replacement` and `suggestion_applicability` are null?

Jarcho: It's because the span is for the impl block generated by the derive. You need to use `span.ctxt().outer_expn_data().call_site`.

nsunderland1: Ahhh, that makes a lot of sense, thanks! Fixed in the latest revision.

clippy_lints/src/derive.rs
```diff
@@ -176,6 +210,7 @@ impl<'tcx> LateLintPass<'tcx> for Derive {
 
             check_hash_peq(cx, item.span, trait_ref, ty, is_automatically_derived);
             check_ord_partial_ord(cx, item.span, trait_ref, ty, is_automatically_derived);
+            check_partial_eq_without_eq(cx, item.span, trait_ref, ty, is_automatically_derived);
```
giraffate, line 213: Can this be moved to `if is_automatically_derived` block and `is_automatically_derived` be removed from args?

clippy_lints/src/derive.rs
giraffate: @rust-lang/clippy @dswij @Alexendoo @Jarcho Which category do you think is appropriate? I think `suspicious` looks good, but the original code isn't wrong.

clippy_lints/src/derive.rs
nsunderland1: Thanks for catching that, will fix

clippy_lints/src/derive.rs
nsunderland1: So this might be me being silly, but when I see a type that implements `PartialEq` but not `Eq`, I kind of read that as an assertion that "This is _not_ an equivalence relation, it's only a partial equivalence relation". In that sense I see it as suspicious to not implement `Eq` when you do have an equivalence relation, but that's pretty pedantic. I think you're correct in saying that this is primarily a question of adhering to API guidelines.

Jarcho: The only reason not to implement `Eq` when it could be derived, other that laziness, is when there's the potential to extend the type in a way which could not implement `Eq`. This is not most cases, but it is a possibility. This possibility is, to me, enough of a reason to not categorize it as `suspicious`.

If not deriving `Eq` would cause the type to be misused, then `suspicious` would be a perfectly reasonable category. I think all our `suspicious` lints basically point out things which are easy to miswrite, or easy to misread in such a way that it creates a bug, but they are still reasonablish things to type. I haven't actually checked through all the `suspicious` lints, so that might not actually describe all `suspicious` lints.

nsunderland1:
Tentatively put the lint under `style` in the latest revision. As a side effect of this, I had to fix cases where tests were hitting the lint. In some of those I just added the derive, unless the tests cared specifically about whether `Eq` was implemented, in which case I just told them to ignore the lint. Tried to be fairly conservative there.

Also had to fix a few places in clippy where trivial enums were not being marked as `Eq`, and added a rustfix test for the lint.

nsunderland1:
Took me a couple of tries to realize that `cargo test --features deny-warnings,internal` would catch everything for me :) Couple more places where this requires an `Eq` added in clippy internals, but all the cases are pretty straightforward.

clippy_lints/src/derive.rs
flip1995: Completely agree with @Jarcho on this one. `sus` lints are for code that is "most likely wrong or useless". Not implementing `Eq` is not wrong, at least there is no way for Clippy to tell if it is intentional (possibly expanding the struct in the future). With that the API concern is the one left and for those, `style` is the best group.

giraffate: Thanks for the comments!

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8799, opened 2022-05-07
Some lintcheck cleanup

Alexendoo:
A grab bag of smaller changes:

Panics if a crate source isn't valid, and fixes the `ryu` crate entry so it's picked up. Took me a while to realise why I couldn't do `cargo lintcheck --only ryu` 😅

Parses messages with `cargo_metadata`, this catches a few more lints that were ignored because the message had an [expansion field](https://docs.rs/cargo_metadata/latest/cargo_metadata/diagnostic/struct.DiagnosticSpanMacroExpansion.html) that contained a path pointing into `.../.toolchains/...`. It also no longer emits some log lines with `null` as the column

It also merges the `run_clippy_lint` invocations and splits `LintcheckConfig` out into its own file

changelog: none

### DIFF
--- a/lintcheck/src/config.rs
+++ b/lintcheck/src/config.rs
@@ -1,0 +1,140 @@
+use clap::{App, Arg, ArgMatches};
+use std::env;
+use std::path::PathBuf;
+
+fn get_clap_config<'a>() -> ArgMatches<'a> {
+    App::new("lintcheck")
+        .about("run clippy on a set of crates and check output")
+        .arg(
+            Arg::with_name("only")
+                .takes_value(true)
+                .value_name("CRATE")
+                .long("only")
+                .help("Only process a single crate of the list"),
+        )
+        .arg(
+            Arg::with_name("crates-toml")
+                .takes_value(true)
+                .value_name("CRATES-SOURCES-TOML-PATH")
+                .long("crates-toml")
+                .help("Set the path for a crates.toml where lintcheck should read the sources from"),
+        )
+        .arg(
+            Arg::with_name("threads")
+                .takes_value(true)
+                .value_name("N")
+                .short("j")
+                .long("jobs")
+                .help("Number of threads to use, 0 automatic choice"),
+        )
+        .arg(
+            Arg::with_name("fix")
+                .long("--fix")
+                .help("Runs cargo clippy --fix and checks if all suggestions apply"),
+        )
+        .arg(
+            Arg::with_name("filter")
+                .long("--filter")
+                .takes_value(true)
+                .multiple(true)
+                .value_name("clippy_lint_name")
+                .help("Apply a filter to only collect specified lints, this also overrides `allow` attributes"),
+        )
+        .arg(
+            Arg::with_name("markdown")
+                .long("--markdown")
+                .help("Change the reports table to use markdown links"),
+        )
+        .get_matches()
+}
+
+#[derive(Debug)]
+pub(crate) struct LintcheckConfig {
+    /// max number of jobs to spawn (default 1)
+    pub max_jobs: usize,
+    /// we read the sources to check from here
+    pub sources_toml_path: PathBuf,
+    /// we save the clippy lint results here
+    pub lintcheck_results_path: PathBuf,
+    /// Check only a specified package
+    pub only: Option<String>,
+    /// whether to just run --fix and not collect all the warnings
+    pub fix: bool,
+    /// A list of lints that this lintcheck run should focus on
+    pub lint_filter: Vec<String>,
+    /// Indicate if the output should support markdown syntax
+    pub markdown: bool,
+}
+
+impl LintcheckConfig {
+    pub fn new() -> Self {
+        let clap_config = get_clap_config();
+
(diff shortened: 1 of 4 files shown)


lintcheck/src/main.rs
```diff
@@ -371,127 +402,15 @@ impl Crate {
...
-        let output_lines = stdout.lines();
-        let warnings: Vec<ClippyWarning> = output_lines
-            .into_iter()
-            // get all clippy warnings and ICEs
-            .filter(|line| filter_clippy_warnings(&line))
-            .map(|json_msg| parse_json_message(json_msg, &self))
+        // get all clippy warnings and ICEs
+        let warnings: Vec<ClippyWarning> = Message::parse_stream(stdout.as_bytes())
+            .filter_map(|msg| ClippyWarning::new(msg.unwrap(), &self))
             .collect();
```
matthiaskrgr, line 409: oh wow, so we can basically pipe rustc output into cargo-metadata and it will let us do some fancy processing of the stdout/stderr? that's super useful to know :grin:

Alexendoo: Yeah it's handy indeed

If you're parsing rustc output directly the format is slightly different from cargo's, the json is untagged and it'll emit [`Diagnostic`s](https://docs.rs/cargo_metadata/latest/cargo_metadata/diagnostic/struct.Diagnostic.html) directly

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
...
+                .takes_value(true)
+                .value_name("N")
+                .short("j")
+                .long("jobs")
+                .help("number of threads to use, 0 automatic choice"),
+        )
+        .arg(
+            Arg::with_name("fix")
+                .long("--fix")
+                .help("runs cargo clippy --fix and checks if all suggestions apply"),
```
giraffate, line 33: nits
```suggestion
                .help("Runs cargo clippy --fix and checks if all suggestions apply"),
```

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
+use clap::{App, Arg, ArgMatches};
+use std::env;
+use std::path::PathBuf;
+
+fn get_clap_config<'a>() -> ArgMatches<'a> {
+    App::new("lintcheck")
+        .about("run clippy on a set of crates and check output")
```
giraffate, line 7:
```suggestion
        .about("Run clippy on a set of crates and check output")
```

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
...
+        .arg(
+            Arg::with_name("only")
+                .takes_value(true)
+                .value_name("CRATE")
+                .long("only")
+                .help("only process a single crate of the list"),
```
giraffate, line 13: nits
```suggestion
                .help("Only process a single crate of the list"),
```

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
...
+        )
+        .arg(
+            Arg::with_name("crates-toml")
+                .takes_value(true)
+                .value_name("CRATES-SOURCES-TOML-PATH")
+                .long("crates-toml")
+                .help("set the path for a crates.toml where lintcheck should read the sources from"),
```
giraffate, line 20: nits
```suggestion
                .help("Set the path for a crates.toml where lintcheck should read the sources from"),
```

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
...
+        )
+        .arg(
+            Arg::with_name("threads")
+                .takes_value(true)
+                .value_name("N")
+                .short("j")
+                .long("jobs")
+                .help("number of threads to use, 0 automatic choice"),
```
giraffate, line 28:
```suggestion
                .help("Number of threads to use, 0 automatic choice"),
```

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
...
+                .long("--fix")
+                .help("runs cargo clippy --fix and checks if all suggestions apply"),
+        )
+        .arg(
+            Arg::with_name("filter")
+                .long("--filter")
+                .takes_value(true)
+                .multiple(true)
+                .value_name("clippy_lint_name")
+                .help("apply a filter to only collect specified lints, this also overrides `allow` attributes"),
```
giraffate, line 41:
```suggestion
                .help("Apply a filter to only collect specified lints, this also overrides `allow` attributes"),
```

lintcheck/src/config.rs
```diff
@@ -0,0 +1,140 @@
...
+        )
+        .arg(
+            Arg::with_name("markdown")
+                .long("--markdown")
+                .help("change the reports table to use markdown links"),
```
giraffate, line 46: nits
```suggestion
                .help("Change the reports table to use markdown links"),
```

giraffate (commented):
Overall looks good to me, thanks!

I made some small comments.

giraffate:
@bors r+

Thanks!

PR #8802, opened 2022-05-08
Optionally allow `expect` and `unwrap` in tests

### LINKED ISSUE #1015: Disable result_unwrap_used / option_unwrap_used in tests
If clippy is used as a dependency, the lints can pick up test code. Since that code does not need to have full-blown error handling, unwrap is ok.

To solve this, we can use `get_parent` to walk up the parent chain and check all items for `#[test]` annotations.

smoelius:
This addresses #1015, except it makes the new behavior optional.

The reason for the msrv-related changes is as follows.

Rather than expand `check_methods` list of arguments, it seemed easier to make `check_methods` a method of `Methods`, so that `check_methods` could access `Methods`' fields.

`check_methods` had an `msrv` parameter, which I consequently made a field of `Methods`. But, to avoid adding a lifetime parameter to `Methods`, I made the field type `Option<RustcVersion>` instead of the parameter's existing type, `Option<&RustcVersion>`. This seemed sensible since `RustcVersion` implements `Copy`. But this broke a lot of code that expected an `Option<&RustcVersion>` or `&Option<RustcVersion>`. I changed all of those occurrences to `Option<RustcVersion>`. IMHO, the code is better as a result of these changes, though.

The msrv-related changes are in their own commit to (hopefully) ease review.

Closes #1015

changelog: optionally allow `expect` and `unwrap` in tests

r? @llogiq


llogiq:
Yeah, the `&` and `.as_ref()`s are noisy. I doubt it'll make much of a difference perf-wise. Thank you!

@bors r+

PR #8805, opened 2022-05-09
Recommend let chains over if_chain in docs

Alexendoo:
Switches over bit_mask.rs to let chains in order to create a nice example

While the rustfmt thing isn't resolved yet, my rust-analyzer isn't a fan of large `if_chains!`s, it stops giving me hover info and such after some number of if statements

changelog: none


flip1995:
Formatting isn't supported for either and IDE support is the same for me.

@bors r+ p=1

PR #8809, opened 2022-05-09
Ignore unfulfilled_lint_expectations in metadata collection

Alexendoo:
r? @xFrednet 

When metadata collection is enabled the regular early/late pass registrations are skipped, which the `#[expect()]`s were no fan of: https://github.com/rust-lang/rust-clippy/runs/6356675328?check_suite_focus=true

changelog: none


xFrednet:
Ahh, you're right, nice quick catch. Thank you very much!

@bors r+

xFrednet:
Now everything is green again :tada:

PR #8813, opened 2022-05-10
Fix redundant_allocation warning for Rc<Box<str>>

### LINKED ISSUE #8604: redundant_allocation warns about usage of Rc<Box<str>>
### Summary

`Rc<str>` is a fat pointer because `str` is an unsized type, so to turn it into a thin pointer either `Rc<Box<str>>` or `Rc<String>` has to be used. Clippy seems to be perfectly fine with the latter, but the former triggers a false positive.

### Lint Name

redundant_allocation

### Reproducer

The following code:
```rs
use std::rc::Rc;

struct S {
    s: Rc<Box<str>>,
}
```
Produces this warning:
```
warning: usage of `Rc<Box<str>>`
 --> src/lib.rs:4:8
  |
4 |     s: Rc<Box<str>>,
  |        ^^^^^^^^^^^^
  |
  = note: `#[warn(clippy::redundant_allocation)]` on by default
  = note: `Box<str>` is already on the heap, `Rc<Box<str>>` makes an extra allocation
  = help: consider using just `Rc<str>` or `Box<str>`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_allocation
```

### Version

```text
rustc 1.61.0-nightly (5f4e06771 2022-03-10)
binary: rustc
commit-hash: 5f4e0677190b82e61dc507e3e72caf89da8e5e28
commit-date: 2022-03-10
release: 1.61.0-nightly
LLVM version: 14.0.0
```

### Additional Labels

_No response_

evantypanski:
changelog: [`redundant_allocation`] Fixes #8604 

Fixes false positives where a fat pointer with `str` type was made thin by another allocation, but that thinning allocation was marked as redundant

### DIFF
--- a/clippy_lints/src/types/redundant_allocation.rs
+++ b/clippy_lints/src/types/redundant_allocation.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::source::{snippet, snippet_with_applicability};
 use clippy_utils::{path_def_id, qpath_generic_tys};
 use rustc_errors::Applicability;
-use rustc_hir::{self as hir, def_id::DefId, QPath, TyKind};
+use rustc_hir::{self as hir, def, def_id::DefId, PrimTy, QPath, TyKind};
 use rustc_lint::LateContext;
 use rustc_span::symbol::sym;
 
@@ -54,8 +54,7 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
     };
     let inner_span = match qpath_generic_tys(inner_qpath).next() {
         Some(ty) => {
-            // Box<Box<dyn T>> is smaller than Box<dyn T> because of wide pointers
-            if matches!(ty.kind, TyKind::TraitObject(..)) {
+            if alloc_makes_pointer_thin(cx, ty) {
                 return false;
             }
             ty.span
@@ -109,4 +108,21 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
         );
     }
     true
+}
+
+/// Returns `true` if the allocation would make `hir_ty` go from fat to thin.
+fn alloc_makes_pointer_thin(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>) -> bool {
+    match &hir_ty.kind {
+        TyKind::TraitObject(..) => true,
+        TyKind::Path(ty_qpath) => {
+            let ty_res = cx.qpath_res(ty_qpath, hir_ty.hir_id);
+            if let def::Res::PrimTy(prim_ty) = ty_res {
+                if matches!(prim_ty, PrimTy::Str) {
+                    return true;
+                }
+            }
+            false
+        },
+        _ => false,
+    }
 }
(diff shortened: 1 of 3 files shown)


xFrednet:
Welcome to Clippy :wave: 

r? @Alexendoo

Alexendoo:
Welcome!

It would be nice to get it working with other types too like `Box<[T]>`, `Box<Path>`. @xFrednet is `hir_ty_to_ty` something we want to avoid more of (I notice it says it's quasi deprecated) or would it be fine to add here?

xFrednet:
I've heard that `hir_ty_to_ty` can cause ICEs, but I'm unsure under which circumstances. If possible it's better to get the type from the expression. @flip1995 What do you think? :upside_down_face:

flip1995:
`hir_ty_to_ty` can not be used in `bodies`. In bodies you can use `tcx.tycheck_results().node_ty()` or similar functions instead.

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -110,3 +109,20 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
...
+            if let def::Res::PrimTy(prim_ty) = ty_res {
+                if matches!(prim_ty, PrimTy::Str) {
+                    return true;
+                }
+            }
+            false
+        },
+        _ => false,
+    }
+}
```
Alexendoo, line 128: Since `redundant_allocation::check` isn't called on locals, I think that means it'd never be in the context of a body, which would make it safe to use

```rust
!hir_ty_to_ty(cx.tcx, hir_ty).is_sized(cx.tcx.at(hir_ty.span), cx.param_env)
```

evantypanski: Awesome, nice and easy! Moved it inline too and added some tests. I couldn't come up with anything to break `hir_ty_to_ty` in this instance, personally.

clippy_lints/src/types/redundant_allocation.rs
```diff
@@ -54,8 +55,8 @@ pub(super) fn check(cx: &LateContext<'_>, hir_ty: &hir::Ty<'_>, qpath: &QPath<'_
     };
     let inner_span = match qpath_generic_tys(inner_qpath).next() {
         Some(ty) => {
-            // Box<Box<dyn T>> is smaller than Box<dyn T> because of wide pointers
-            if matches!(ty.kind, TyKind::TraitObject(..)) {
+            // Reallocation of a fat pointer causes it to become thin
+            if !hir_ty_to_ty(cx.tcx, ty).is_sized(cx.tcx.at(ty.span), cx.param_env) {
```
flip1995, line 59: Can you add a comment here that `hir_ty_to_ty` is never called inside a body, because the `check_ty` function in `mod.rs` is only called on types outside bodies and that this lint is not run on locals (also guaranteed in `mod.rs`).

flip1995, line 59: Can you add a comment here that `hir_ty_to_ty` is never called inside a body, because the `check_ty` function in `mod.rs` is only called on types outside bodies and that this lint is not run on locals (also guaranteed in `mod.rs`).

evantypanski: done!

flip1995:
Thanks!

@bors delegate=Alexendoo  (I leave the final r+ to Alex)

Alexendoo:
@bors r+

Thanks @evantypanski

PR #8814, opened 2022-05-10
add suggestions to rc_clone_in_vec_init

yonip23:
A followup to https://github.com/rust-lang/rust-clippy/pull/8769
I also switch the order of the 2 suggestions, since the loop initialization one is probably the common case.

@xFrednet I'm not letting you guys rest for a minute 😅 
changelog: add suggestions to [`rc_clone_in_vec_init`]

### DIFF
--- a/clippy_lints/src/rc_clone_in_vec_init.rs
+++ b/clippy_lints/src/rc_clone_in_vec_init.rs
@@ -106,9 +106,9 @@ fn elem_snippet(cx: &LateContext<'_>, elem: &Expr<'_>, symbol_name: &str) -> Str
 fn loop_init_suggestion(elem: &str, len: &str, indent: &str) -> String {
     format!(
         r#"{{
-{indent}{indent}let mut v = Vec::with_capacity({len});
-{indent}{indent}(0..{len}).for_each(|_| v.push({elem}));
-{indent}{indent}v
+{indent}    let mut v = Vec::with_capacity({len});
+{indent}    (0..{len}).for_each(|_| v.push({elem}));
+{indent}    v
 {indent}}}"#
     )
 }
@@ -116,8 +116,8 @@ fn loop_init_suggestion(elem: &str, len: &str, indent: &str) -> String {
 fn extract_suggestion(elem: &str, len: &str, indent: &str) -> String {
     format!(
         "{{
-{indent}{indent}let data = {elem};
-{indent}{indent}vec![data; {len}]
+{indent}    let data = {elem};
+{indent}    vec![data; {len}]
 {indent}}}"
     )
 }
--- a/clippy_lints/src/rc_clone_in_vec_init.rs
+++ b/clippy_lints/src/rc_clone_in_vec_init.rs
@@ -3,7 +3,7 @@ use clippy_utils::higher::VecArgs;
 use clippy_utils::last_path_segment;
 use clippy_utils::macros::root_macro_call_first_node;
 use clippy_utils::source::{indent_of, snippet};
-use rustc_errors::{Applicability, Diagnostic};
+use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, QPath, TyKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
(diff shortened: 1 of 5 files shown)


clippy_lints/src/rc_clone_in_vec_init.rs
```diff
@@ -47,27 +49,107 @@ declare_lint_pass!(RcCloneInVecInit => [RC_CLONE_IN_VEC_INIT]);
...
+        let lint_suggestions =
+            construct_lint_suggestions(lint_span, symbol_name, &elem_snippet, len_snippet.as_ref(), indentation);
 
-        emit_lint(cx, symbol, &macro_call);
+        emit_lint(cx, symbol, lint_span, &lint_suggestions);
     }
 }
 
-fn emit_lint(cx: &LateContext<'_>, symbol: Symbol, macro_call: &MacroCall) {
+struct LintSuggestion {
```
yonip23, line 66: I;m wondering whether this can be extracted to a util somewhere

xFrednet:
> I'm not letting you guys rest for a minute

There ain't no rest for the wicked :notes: :notes: 

I'm happy to see the follow-up PR. I'll put it on my todo list. FYI, you can also request a specific reviewer with:

r? @xFrednet

clippy_lints/src/rc_clone_in_vec_init.rs
xFrednet: I think in most cases it's better to use `span_lint_and_then` and pass the `DiagnosticBuilder` around. That's also a bit more efficient when it comes to performance. The `DiagnosticBuilder` will store the suggestions internally. Is there a reason why you want to use this intermediate struct to store the suggestions? :upside_down_face:

yonip23: Just so I can construct the suggestions separately from where I'm emitting the lint, so that the `span_lint_and_then` code will be clean from logistics.

In terms of performance, since I'm passing a `& mut Diagnostic` the performance overhead is basically only the `Vec<LintSuggestion>` creation - which I thought is a small price to pay for readability.

Lmk if there's some other design you'd prefer 🙂

clippy_lints/src/rc_clone_in_vec_init.rs
```diff
@@ -47,27 +49,107 @@ declare_lint_pass!(RcCloneInVecInit => [RC_CLONE_IN_VEC_INIT]);
...
+            elem_snippet.find(&reference_initialization).unwrap() + reference_initialization.len();
+        elem_snippet.replace_range(reference_initialization_end.., "..");
+    }
+    elem_snippet
+}
+
+fn loop_init_suggestion(elem: &str, len: &str, indent: &str) -> String {
+    format!(
+        r#"{{
+{indent}{indent}let mut v = Vec::with_capacity({len});
```
xFrednet, line 124: Having two indents like this, will probably look weird, if the indent is != 4. Did you mean to write 
```suggestion
{indent}    let mut v = Vec::with_capacity({len});
```

clippy_lints/src/rc_clone_in_vec_init.rs
```diff
@@ -47,27 +49,107 @@ declare_lint_pass!(RcCloneInVecInit => [RC_CLONE_IN_VEC_INIT]);
...
+
+fn elem_snippet(cx: &LateContext<'_>, elem: &Expr<'_>, symbol_name: &str) -> String {
+    let mut elem_snippet = snippet(cx, elem.span, "..").to_string();
+    if elem_snippet.contains('\n') {
+        let reference_initialization = format!("{symbol_name}::new");
+        // This string must be found in `elem_snippet`, otherwise we won't be constructing the snippet in
+        // the first place.
+        let reference_initialization_end =
+            elem_snippet.find(&reference_initialization).unwrap() + reference_initialization.len();
+        elem_snippet.replace_range(reference_initialization_end.., "..");
```
xFrednet, line 116: Nice explanatory comment :upside_down_face: 

Why do you use `replace_range`? Wouldn't it be simpler to reassign `elem_snippet` like this?

```rs
if elem_snippet.contains('\n') {
    elem_snippet = format!("{symbol_name}::new(..)");
}
```

clippy_lints/src/rc_clone_in_vec_init.rs
xFrednet: The idea in rustc is actually to create the suggestion inside the closure. Clippy does it a bit differently to have a simpler interface, but if we use `span_lint_and_then` we usually try to also create the suggestion inside. Extracting the specific string creation to single functions like this is nice, but I wouldn't store them in another struct. Also, since the `Span` and `Applicability` is always the same.

> In terms of performance

This comment was more about the internal implementation in rustc. If lints are allowed, they won't be emitted, rustc then never calls the closure. That's what I was referring to. That is super nit picky though, it only came to mind since I've worked on that recently. I sometimes go a bit overboard with my reviews, this side node is an example for this :sweat_smile:. So please ignore the performance comment :upside_down_face:

xFrednet (commented):
The lint message looks fantastic, I left some comments regarding the implementation :upside_down_face:

clippy_lints/src/rc_clone_in_vec_init.rs
yonip23: I'm willing to go as deep into this rabbit hole as you are 🙂 
np, I'll change the design a bit so that everything is being passed to `emit_lint` and it'll construct the suggestion by calling another function inside the clousure

clippy_lints/src/rc_clone_in_vec_init.rs
yonip23: mmm I think it'll be better to show the same indentation for the suggestion's code block as the rest of the file. Why not aligning my indentation with the user's?

clippy_lints/src/rc_clone_in_vec_init.rs
yonip23: lol I played around with the snippet trimming, tried some variations until i came up with this monstrosity 🤣

yonip23:
I think the only comment I didn't fix was the `{indent}` one - lmk wdyt

clippy_lints/src/rc_clone_in_vec_init.rs
xFrednet: The indention is correct, in this case you want to use the base indention plus the four for the block. If you set `indent` to `            ` (12 Spaces) then the suggestion  would look like this:

```rs
LL ~            let v1 = {
LL +                        let data = Arc::new(..);
LL +                        vec![data; 2]
LL ~            };
```

Instead, we only want the first indention to adapt and the second one to be ident + 4

clippy_lints/src/rc_clone_in_vec_init.rs
xFrednet: Alright, now the question remains, what is the benefit of this struct. I think that adding everything together in the `span_lint_and_then` closure would look a bit cleaner (and avoid a vec allocation). For example, it would look like this:

```rs
    span_lint_and_then(
        cx,
        RC_CLONE_IN_VEC_INIT,
        lint_span,
        &format!("calling `{symbol_name}::new` in `vec![elem; len]`"),
        |diag| {
            let len_snippet = snippet(cx, len.span, "..");
            let elem_snippet = elem_snippet(cx, elem, symbol_name);
            let indentation = indent_of(cx, span).unwrap_or(0);
            let indentation = " ".repeat(indentation);
            let suggestions = construct_lint_suggestions(cx, lint_span, symbol_name, elem, len);

            diag.note(format!("each element will point to the same `{symbol_name}` instance"));
            diag.span_suggestion(
                lint_span,
                &format!("consider initializing each `{symbol_name}` element individually"),
                &loop_init_suggestion(&elem_snippet, len_snippet.as_ref(), &indentation),
                Applicability::Unspecified,
            );
            diag.span_suggestion(
                lint_span,
                &format!("or if this is intentional, consider extracting the `{symbol_name}` initialization to a variable"),
                &extract_suggestion(&elem_snippet, len_snippet.as_ref(), &indentation),
                Applicability::Unspecified,
            );
        },
    );
```

This way everything is together in one place, and it still seems very readable to be :upside_down_face:

> I'm willing to go as deep into this rabbit hole as you are 

Okay, how much time do you have? :joy: JK, I only looked at it in that detail to implement something else in rustc

clippy_lints/src/rc_clone_in_vec_init.rs
yonip23: good point, fixed it now 🙂

clippy_lints/src/rc_clone_in_vec_init.rs
yonip23: Didn't notice your comment there, oops 😄 

I really want the closure to be as clean as possible, this is just my style. I ended up still having a struct and a vec, but now everything is being called from within the closure.
Its somewhere in the middle between your idea and mine haha

If you really want to get rid of the struct and the vec, I can change `construct_lint_suggestions` to receive a `&mut Diagnostic` as well, and it'll call `diag.span_lint` inside.

I think its less readable to encapsulate the `diag` api calls inside `construct_lint_suggestions`, but i don't really mind if it'll sit better with you 🙂

yonip23: Question is, how important are performance here? Is vec allocation going to cost too much here?

xFrednet: Not really, this is more about the style, I don't really see the reason for having the struct just to pass them to the `DiagnosticBuilder` afterwards. What about passing in the `DiagnosticBuilder` into the function as an argument?

yonip23: that's basically what I offered, just don't know what is `DiagnosticBuilder` exactly 🙈 
the closure receives a `&mut Diagnostic`.
I'm offering to pass it into `construct_lint_suggestions` such that this function will span the suggestions instead of returning them as a vec.

> I don't really see the reason for having the struct just to pass them to the DiagnosticBuilder afterwards.

Imo passing **named** data between scopes is arguably justifying a struct.
None the less, I'll change it as you wish of course 🙂

yonip23: I ended up basically taking your suggestion from here: https://github.com/rust-lang/rust-clippy/pull/8814#discussion_r870730625

xFrednet: > Imo passing named data between scopes is arguably justifying a struct.

That's totally correct. In this case, I just felt a bit weird, since I would say that the `DiagnosticBuilder` is exactly for this. Sorry for not explaining what the builder is, it's just the argument received from the `span_lint_and_then` argument. This implementation looks good to me on a quick glance, I'll try to do a full review soon :upside_down_face:

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update and diagnostic improvement. I like the new diagnostic message :)

@bors r+

PR #8831, opened 2022-05-14
Add new lint `mismatching_type_param_order`

### LINKED ISSUE #7147: Inconsistent type-parameter order
### What it does

Warn if type parameters are named identically but ordered inconsistently in type definitions and impl-blocks. This is a cousin to the new [`inconsistent_struct_constructor`](https://rust-lang.github.io/rust-clippy/master/index.html#inconsistent_struct_constructor) lint. With _literal constructors_, the order of variables does not matter (`Foo { a, b }` is the same as `Foo { b, a }`). Type parameters, on the other hand, are **never** referred to by name, but only by position, which is exactly reversed to the behavior of struct constructors. [For example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6accb007583d1edc58644ffd3739f78d):

```rust
struct Foo<A, B> {
    a: A,
    b: B,
}

// Potential warning: `B` does not refer to `B` in the struct-definition, because of position.
impl<B, A> Foo<B, A> {
    fn type_names(&self) {
        let name_a = std::any::type_name::<A>();
        let name_b = std::any::type_name::<B>();
        println!("A: {}, B: {}", name_a, name_b);
    }

    fn values(&self)
    where
        A: std::fmt::Debug,
        B: std::fmt::Debug,
    {
        println!("a: {:?}, b: {:?}", self.a, self.b);
    }
}

fn main() {
    let foo = Foo { a: 0, b: "foo" };

    // Prints 'A: &str, B: i32'
    foo.type_names();

    // Prints 'a: 0, b: "foo"'
    foo.values();

    // Wait, what? A `i32` valued as "foo" ?
}

```

### Categories (optional)

- Kind: Possibly `clippy::correctness`, because getting this wrong may cause you to refer to the wrong type parameter.

- Naming the type parameters consistently avoids confusion between type definition and type implementation.

### Drawbacks

None.

### Example

```rust
// Note::
struct Foo<A, B> {
// ...

impl<B, A> Foo<B, A> {
//   ^--|- This type parameter refers to `Foo::A`, but is named `B`.
//      ^- This type parameter refers to `Foo::B`, but is named `A`
```

Could be written as:

```rust
impl<A, B> Foo<A, B> {
```

arieluy:
changelog: Add new lint [`mismatching_type_param_order`] for checking if type parameters are consistent between type definitions and impl blocks.

fixes #7147

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3560,6 +3560,7 @@ Released 2018-09-13
 [`min_max`]: https://rust-lang.github.io/rust-clippy/master/index.html#min_max
 [`misaligned_transmute`]: https://rust-lang.github.io/rust-clippy/master/index.html#misaligned_transmute
 [`mismatched_target_os`]: https://rust-lang.github.io/rust-clippy/master/index.html#mismatched_target_os
+[`mismatching_type_param_order`]: https://rust-lang.github.io/rust-clippy/master/index.html#mismatching_type_param_order
 [`misrefactored_assign_op`]: https://rust-lang.github.io/rust-clippy/master/index.html#misrefactored_assign_op
 [`missing_const_for_fn`]: https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn
 [`missing_docs_in_private_items`]: https://rust-lang.github.io/rust-clippy/master/index.html#missing_docs_in_private_items
--- a/clippy_lints/src/lib.register_lints.rs
+++ b/clippy_lints/src/lib.register_lints.rs
@@ -380,6 +380,7 @@ store.register_lints(&[
     misc_early::UNNEEDED_WILDCARD_PATTERN,
     misc_early::UNSEPARATED_LITERAL_SUFFIX,
     misc_early::ZERO_PREFIXED_LITERAL,
+    mismatching_type_param_order::MISMATCHING_TYPE_PARAM_ORDER,
     missing_const_for_fn::MISSING_CONST_FOR_FN,
     missing_doc::MISSING_DOCS_IN_PRIVATE_ITEMS,
     missing_enforced_import_rename::MISSING_ENFORCED_IMPORT_RENAMES,
--- a/clippy_lints/src/lib.register_pedantic.rs
+++ b/clippy_lints/src/lib.register_pedantic.rs
@@ -67,6 +67,7 @@ store.register_group(true, "clippy::pedantic", Some("clippy_pedantic"), vec![
     LintId::of(methods::UNNECESSARY_JOIN),
     LintId::of(misc::FLOAT_CMP),
     LintId::of(misc::USED_UNDERSCORE_BINDING),
+    LintId::of(mismatching_type_param_order::MISMATCHING_TYPE_PARAM_ORDER),
     LintId::of(mut_mut::MUT_MUT),
     LintId::of(needless_bitwise_bool::NEEDLESS_BITWISE_BOOL),
     LintId::of(needless_continue::NEEDLESS_CONTINUE),
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -298,6 +298,7 @@ mod methods;
 mod minmax;
 mod misc;
 mod misc_early;
+mod mismatching_type_param_order;
 mod missing_const_for_fn;
 mod missing_doc;
 mod missing_enforced_import_rename;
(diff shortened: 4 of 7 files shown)


arieluy:
Merge conflicts are fixed

dswij:
Sorry, I forgot that this was in my review queue. I've taken a brief look, but got distracted. Will review this soon

tests/ui/type_param_mismatch.stderr
```diff
@@ -0,0 +1,64 @@
+error: this type parameter is named B, but it does not refer to Foo::B
```
dswij, line 1: I find the error message to be confusing. It's not clear what `does not refer to Foo::B` is about.

I think the clarity can be improved. Perhaps something like this?

```suggestion
error: `Foo` has a similarly named generic type parameter `B` in its declaration, but in a different order.
```

clippy_lints/src/type_param_mismatch.rs
```diff
@@ -0,0 +1,113 @@
...
+// Checks if impl_param_name is the same as one of type_param_names,
+// and is in a different position
+fn mismatch_param_name(i: usize, impl_param_name: &str, type_param_names: &[String]) -> bool {
+    for (j, type_param_name) in type_param_names.iter().enumerate() {
+        if impl_param_name == type_param_name && i != j {
+            return true;
+        }
+    }
+    false
+}
```
dswij, line 113: Instead of iterating over `type_param_names` and checking the index, we can store `type_param_names` in a hashmap of `name => index` and just fetch the value from there.

clippy_lints/src/type_param_mismatch.rs
```diff
@@ -0,0 +1,113 @@
...
+                for (i, (impl_param_name, impl_param_span)) in impl_params.enumerate() {
+                    if mismatch_param_name(i, &impl_param_name, &type_param_names) {
+                        let msg = format!("this type parameter is named {}, but it does not refer to {}::{}",
+                                  impl_param_name, type_name, impl_param_name);
+                            span_lint_and_sugg(
+                                cx,
+                                TYPE_PARAM_MISMATCH,
+                                impl_param_span,
+                                &msg,
+                                "try",
```
dswij, line 93: This can be more descriptive. The user can actually do two things: use the same name as the one in the type definition, or rename it to something else.

clippy_lints/src/type_param_mismatch.rs
```diff
@@ -0,0 +1,113 @@
...
+                );
+
+                // find the type that the Impl is for
+                // only lint on struct/enum/union for now
+                let defid = match path.res {
+                    Res::Def(DefKind::Struct | DefKind::Enum | DefKind::Union, defid) => defid,
+                    _ => return,
+                };
+
+                let type_name = segment.ident;
```
dswij, line 72: nit: I personally would put this variable closer to where it's first used. Makes reading the code easier in terms of context/scope.

dswij (changes requested):
Thanks for the PR (and waiting patiently for the review)! 👍 

I'd like to apologize first for forgetting to review this for a while.

I think we should name the lint `type_param_order_mismatch` since we're only checking the order. `type_param_mismatch` is a little bit ambiguous. What do you think?

xFrednet:
> I think we should name the lint `type_param_order_mismatch` since we're only checking the order.

Agreed, good catch! The name should ideally be plural to align with rust's [lint naming guidelines](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints). So, `type_param_order_mismatches`. I would maybe suggest the name `mismatching_type_param_order` as it would work a bit better with lint attributes. :upside_down_face:

tests/ui/type_param_mismatch.stderr
arieluy: Thanks for the suggestion, that sounds better.

clippy_lints/src/type_param_mismatch.rs
arieluy: I considered a hashmap, but it seems a bit unnecessary/more loc. The number of things we are iterating through is probably low. 

Willing to change if you prefer, though.

clippy_lints/src/type_param_mismatch.rs
arieluy: Should using the same name as the one in the type definition be the preferred/suggested solution? That seems the least confusing to me.

arieluy:
Renamed lint, thanks for the suggestion. Also resolved merge conflicts and a few other things.

clippy_lints/src/type_param_mismatch.rs
dswij: I prefer using a hashmap, it makes the code easier to understand, and I don't see any downside from it.

clippy_lints/src/type_param_mismatch.rs
dswij: A user might want accidentally use the same name, but they actually want distinct names from the one in type definitions. 

I think this suggestion is fine, extending it makes it clearer, imo.

I'll leave the decision up to you.

arieluy:
Added hashmap, changed the help message, resolved merge conflicts

dswij (approved):

dswij:
This looks good, thanks for this!

@bors r+

dswij:
Whoops, seems like it doesn't like the changelog. 
@arieluy I edited the changelog in PR description

Issue #8812, opened 2022-05-10
Lint suggestion: Path construction using format!


jplatte:
### What it does

Suggest to create `PathBuf`s using `Path::join` (or maybe `PathBuf::push` in some circumstances) instead of string formatting.

### Lint Name

path_format

### Category

suspicious

### Example

```rust
PathBuf::from(format!("{}/foo/bar", base_path)); // or
PathBuf::from(format!("{}\\foo\\bar", base_ath));
```

could be written as

```rust
Path::new(base_path).join("foo").join("bar")
```

merelymyself:
I would like to give this a shot. Could I confirm that something like [this](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/format_push_string.rs) will work?

@rustbot claim

xFrednet:
Yes, the basic idea is the same. You have to find a format macro where the output is used to create a `Path`. Depending on your implementation it could be good to first search for the path creation and than figure out where the string comes from :).

xFrednet:
Hey @jplatte, I've started in the review for the lint and found several examples of paths like `r"C:\windows\system32.dll"` being used in the documentation of [`PathBuf`](https://doc.rust-lang.org/std/path/struct.PathBuf.html). Based on this I wonder if using format for paths is really so wrong. Could you expand a bit why `.join()` is considered better than `format!`.

The main reason I could think of would be the support for `OSStrings` :thinking:

jplatte:
I find it less readable and for portability it could also be better to use the platform default separator. The `join` solution could also be more efficient (but the opposite could also be true for many successive `join`s).

For absolute paths like the example you showed I agree that there isn't much of a point as those are already almost certainly platform-specific. Maybe the lint shouldn't fire if the path is known to be absolute?

xFrednet:
IDK, it seems like `Path` and `PathBuf` are designed to handle separators the same across all platforms. The PR includes some rewrites of format macros with join, and I actually find them less readable than the original format version. If there is no technical benefit, I'd prefer the formatted version. This is not to say we can't have the lint, but in that case, I'd see it more in the `restriction` category.

Jarcho:
Raw paths on windows `\\?\` require backslashes, so if you don't know how the path is rooted then `format!` can break. Also any use of `PathBuf::from(format!("{}/{}", x, y.display()))` should be warned on anyways.

xFrednet:
[...]

PR #8833, opened 2022-05-15
Add new lint `paths_from_format`

merelymyself:

As per #8812 

changelog: Add new lint ``[`path_from_format`]``

### DIFF
--- a/clippy_lints/src/paths_from_format.rs
+++ b/clippy_lints/src/paths_from_format.rs
@@ -124,6 +124,8 @@ fn push_comps(string: &mut String, path: &str) {
     }
 }
 
+// In essence, this checks if the format!() used is made for concatenation of the filenames / folder names itself.
+// This returns true when it is something like `PathBuf::from(format!("/x/folder{}/textfile.txt", folder_number))
 fn is_valid_use_case(string: &str, string2: &str) -> bool {
     !(string.is_empty() || string.ends_with('/') || string.ends_with('\\'))
         || !(string2.is_empty() || string2.starts_with('/') || string2.starts_with('\\'))
--- a/clippy_lints/src/paths_from_format.rs
+++ b/clippy_lints/src/paths_from_format.rs
@@ -1,0 +1,124 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::macros::{root_macro_call, FormatArgsExpn};
+use clippy_utils::sugg::Sugg;
+use clippy_utils::ty::is_type_diagnostic_item;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+use std::fmt::Write as _;
+use std::path::Path;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `PathBuf::from(format!(..))` calls.
+    ///
+    /// ### Why is this bad?
+    /// It is not OS-agnostic, and can be harder to read.
+    ///
+    /// ### Known Problems
+    /// `.join()` introduces additional allocations that are not present when `PathBuf::push` is
+    /// used instead.
+    ///
+    /// ### Example
+    /// ```rust
+    /// use std::path::PathBuf;
+    /// let base_path = "/base";
+    /// PathBuf::from(format!("{}/foo/bar", base_path));
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// use std::path::Path;
+    /// let base_path = "/base";
+    /// Path::new(&base_path).join("foo").join("bar");
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub PATHS_FROM_FORMAT,
+    pedantic,
+    "builds a `PathBuf` from a format macro"
+}
+
+declare_lint_pass!(PathsFromFormat => [PATHS_FROM_FORMAT]);
+
+impl<'tcx> LateLintPass<'tcx> for PathsFromFormat {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if_chain! {
+            if let ExprKind::Call(_, args) = expr.kind;
+            if let ty = cx.typeck_results().expr_ty(expr);
+            if is_type_diagnostic_item(cx, ty, sym::PathBuf);
+            if !args.is_empty();
+            if let Some(macro_call) = root_macro_call(args[0].span);
+            if cx.tcx.item_name(macro_call.def_id) == sym::format;
+            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &args[0], macro_call.expn);
+            then {
+                let format_string_parts = format_args.format_string.parts;
+                let format_value_args = format_args.args;
+                let string_parts: Vec<&str> = format_string_parts.iter().map(rustc_span::Symbol::as_str).collect();
+                let mut applicability = Applicability::MachineApplicable;
+                let real_vars: Vec<Sugg<'_>> = format_value_args.iter().map(|x| Sugg::hir_with_applicability(cx, x.param.value, "..", &mut applicability)).collect();
+                let mut paths_zip = string_parts.iter().take(real_vars.len()).zip(real_vars.clone());
+                let mut sugg = String::new();
+                if let Some((part, arg)) = paths_zip.next() {
+                    if is_valid_use_case(string_parts.first().unwrap_or(&""), string_parts.get(1).unwrap_or(&"")) {
+                        return;
+                    }
+                    if part.is_empty() {
+                        sugg = format!("Path::new(&{arg})");
+                    }
+                    else {
+                        push_comps(&mut sugg, part);
+                        let _ = write!(sugg, ".join(&{arg})");
+                    }
+                }
+                for n in 1..real_vars.len() {
+                    if let Some((part, arg)) = paths_zip.next() {
+                        if is_valid_use_case(string_parts.get(n).unwrap_or(&""), string_parts.get(n+1).unwrap_or(&"")) {
+                            return;
+                        }
+                        else if n < real_vars.len() {
+                            push_comps(&mut sugg, part);
+                            let _ = write!(sugg, ".join(&{arg})");
+                        }
+                        else {
+                            sugg = format!("{sugg}.join(&{arg})");
+                        }
+                    }
+                }
+                if real_vars.len() < string_parts.len() {
+                    push_comps(&mut sugg, string_parts[real_vars.len()]);
+                }
+                span_lint_and_sugg(
+                    cx,
+                    PATHS_FROM_FORMAT,
+                    expr.span,
+                    "`format!(..)` used to form `PathBuf`",
+                    "consider using `Path::new()` and `.join()` to make it OS-agnostic and improve code readability",
+                    sugg,
+                    Applicability::MaybeIncorrect,
+                );
+            }
+        }
+    }
+}
+
+fn push_comps(string: &mut String, path: &str) {
+    let mut path = path.to_string();
+    if !string.is_empty() {
+        path = path.trim_start_matches(|c| c == '\\' || c == '/').to_string();
+    }
+    for n in Path::new(&path).components() {
+        let mut x = n.as_os_str().to_string_lossy().to_string();
+        if string.is_empty() {
+            let _ = write!(string, "Path::new(\"{x}\")");
+        } else {
+            x = x.trim_end_matches(|c| c == '/' || c == '\\').to_string();
+            let _ = write!(string, ".join(\"{x}\")");
+        }
+    }
+}
+
+fn is_valid_use_case(string: &str, string2: &str) -> bool {
+    !(string.is_empty() || string.ends_with('/') || string.ends_with('\\'))
+        || !(string2.is_empty() || string2.starts_with('/') || string2.starts_with('\\'))
+}
--- a/clippy_lints/src/paths_from_format.rs
+++ b/clippy_lints/src/paths_from_format.rs
@@ -123,8 +123,9 @@ fn push_comps(string: &mut String, path: &str) {
     }
 }
 
-// In essence, this checks if the format!() used is made for concatenation of the filenames / folder names itself.
-// This returns true when it is something like `PathBuf::from(format!("/x/folder{}/textfile.txt", folder_number))
+// In essence, this checks if the format!() used is made for concatenation of the filenames / folder
+// names itself. This returns true when it is something like
+// `PathBuf::from(format!("/x/folder{}/textfile.txt", folder_number))`
 fn is_valid_use_case(string: &str, string2: &str) -> bool {
     !(string.is_empty() || string.ends_with('/') || string.ends_with('\\'))
         || !(string2.is_empty() || string2.starts_with('/') || string2.starts_with('\\'))
--- a/clippy_lints/src/paths_from_format.rs
+++ b/clippy_lints/src/paths_from_format.rs
@@ -49,7 +49,7 @@ impl<'tcx> LateLintPass<'tcx> for PathsFromFormat {
             if is_type_diagnostic_item(cx, ty, sym::PathBuf);
             if let Some(macro_call) = root_macro_call(arg.span);
             if cx.tcx.item_name(macro_call.def_id) == sym::format;
-            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &arg, macro_call.expn);
+            if let Some(format_args) = FormatArgsExpn::find_nested(cx, arg, macro_call.expn);
             then {
                 let format_string_parts = format_args.format_string.parts;
                 let format_value_args = format_args.args;
--- a/clippy_lints/src/paths_from_format.rs
+++ b/clippy_lints/src/paths_from_format.rs
@@ -44,13 +44,12 @@ declare_lint_pass!(PathsFromFormat => [PATHS_FROM_FORMAT]);
 impl<'tcx> LateLintPass<'tcx> for PathsFromFormat {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
         if_chain! {
-            if let ExprKind::Call(_, [args, ..]) = expr.kind;
+            if let ExprKind::Call(_, [arg, ..]) = expr.kind;
             if let ty = cx.typeck_results().expr_ty(expr);
             if is_type_diagnostic_item(cx, ty, sym::PathBuf);
-            if !args.is_empty();
-            if let Some(macro_call) = root_macro_call(args[0].span);
+            if let Some(macro_call) = root_macro_call(arg.span);
             if cx.tcx.item_name(macro_call.def_id) == sym::format;
-            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &args[0], macro_call.expn);
+            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &arg, macro_call.expn);
             then {
                 let format_string_parts = format_args.format_string.parts;
                 let format_value_args = format_args.args;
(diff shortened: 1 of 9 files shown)


xFrednet:
Hey @merelymyself, welcome to the project!

r? @dswij (I'd appreciate it if you could leave a normal comment so that I can also assign you to this PR) :upside_down_face:

giraffate:
Could you do a rebase, not merge commit? We follow no merge-commit policy.
https://github.com/rust-lang/rust-clippy/blob/master/doc/basics.md#pr

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,101 @@
...
+                        push_targets.push(temp_string.clone());
+                        temp_string = String::new();
+                    }
+                    else if c == '}' {
+                        temp_string.push_str(&unformatted[1].replace(' ', ""));
+                    }
+                    else if c != '{' && c != '"' {
+                        temp_string.push(c);
+                    }
+                }
```
dswij, line 66: Instead of doing it manually, there's a clippy util to parse `format!` arguments from `Expr`s, maybe we can use that instead 

https://github.com/rust-lang/rust-clippy/blob/f0bf2003de1e50bc722941dc07b6ca593a136000/clippy_utils/src/macros.rs#L449-L467

dswij (commented):

The basic idea for the lint looks good to me 👍 . 
I left some suggestions and questions.

merelymyself:
Hi, @dswij. Made some changes, could you take another look? Thanks!

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,100 @@
...
+            if let ExprKind::Call(_, args) = expr.kind;
+            if let ty = cx.typeck_results().expr_ty(expr);
+            if is_type_diagnostic_item(cx, ty, sym::PathBuf);
+            if !args.is_empty();
+            if let Some(macro_call) = root_macro_call(args[0].span);
+            if cx.tcx.item_name(macro_call.def_id) == sym::format;
+            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &args[0], macro_call.expn);
+            let mut applicability = Applicability::MachineApplicable;
+            let format_args_snip = snippet_with_applicability(cx, format_args.inputs_span(), "..", &mut applicability);
+            if let Some(end_of_literal) = format_args_snip.find("\",");
```
dswij, line 50: can we maybe use `format_args.format_string_parts` instead of using the snippet?

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,100 @@
...
+                        }
+                        format!("Path::new(\"{str1}\").join({real_vars}).join(\"{str2}\")")
+                    }
+                };
+                span_lint_and_sugg(
+                    cx,
+                    PATH_FROM_FORMAT,
+                    expr.span,
+                    "`format!(..)` used to form `PathBuf`",
+                    "consider using `.join()` to avoid the extra allocation",
```
infmagic2047, line 93: This is incorrent: `PathBuf::from(String)` is allocation-free, while `.join()` allocates a new `PathBuf` on every call, so using `.join()` will only result in more allocations, not fewer.

merelymyself: Thanks! What should be put there instead? Is it just about code readability, or is it more efficient in some other way?

infmagic2047: The main advantage is that, `.join()` uses the platform-specific path separator (`/` on Unix and `\` on Windows), so the path is more portable. Readability could also count. I couldn't find a discussion for Rust, but here's the same discussion for Python: https://stackoverflow.com/questions/13944387/why-use-os-path-join-over-string-concatenation

By the way, it is possible to avoid the extra allocation of `.join()` by using `PathBuf::push` instead. I suggest adding this information to the known problems section in the lint description.

merelymyself:
Thanks for the patience and advice, @dswij. Could you take another look?

If there's a situation where a path like 'foo{}/bar' shows up, instead of offering a suggestion, a note is offered instead now.

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+fn push_comps(string: &mut String, path: &Path) {
+    let comps = path.components();
+    for n in comps {
+        let x = n.as_os_str().to_string_lossy().to_string();
+        if string.is_empty() {
+            let _ = write!(string, "Path::new(\"{x}\")");
+        } else {
+            let _ = write!(string, ".join(\"{x}\")");
```
dswij, line 133: Any reason why we're using `write!` macro, instead of using `format!`?

tests/ui/path_from_format.rs
```diff
@@ -0,0 +1,18 @@
...
+
+fn main() {
+    let mut base_path1 = "";
+    let mut base_path2 = "";
+    PathBuf::from(format!("{}/foo/bar", base_path1));
+    PathBuf::from(format!("/foo/bar/{}", base_path1));
+    PathBuf::from(format!("/foo/{}/bar", base_path1));
+    PathBuf::from(format!("foo/{}/bar", base_path1));
+    PathBuf::from(format!("foo/foooo/{base_path1}/bar/barrr"));
+    PathBuf::from(format!("foo/foooo/{base_path1}/bar/barrr"));
```
dswij, line 13: duplicate

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,136 @@
...
+                    else {
+                        let mut string = String::from(string_parts[n]);
+                        if string.starts_with('/') || string.starts_with('\\') {
+                            string.remove(0);
+                        }
+                        push_comps(&mut sugg, Path::new(&string));
+                        let _ = write!(sugg, ".join({})", real_vars[order_of_real_vars[n]]);
+                    }
+                }
+                if !string_parts[real_vars.len()].is_empty() {
```
dswij, line 105: I'm assuming the empty string push in Line 58 is for this scenario, since `string_parts.len` is either `real_vars.len` or `real_vars.len + 1`. the empty string is to ensure that `string_parts.len > real_vars.len`. 

If that's the case, then we can just check if `real_vars.len < string_parts.len`. Then add the last `string_parts` to the suggestion..

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,136 @@
...
+                    let mut string = String::from(string_parts[real_vars.len()]);
+                    if string.starts_with('/') || string.starts_with('\\') {
+                        string.remove(0);
+                    }
```
dswij, line 109: we can dry this by putting it into `push_comps` instead.

dswij (commented):
Thanks for all the changes!

I think we can just not emit the lint for `foo{}/bar`. I'd imagine someone uses it like this, which is legit

```rust
let prefix = "some-prefixes";
let _ Path::new(format!("./{prefix}.csv"));
```

Note that the suggestion is also wrong right now when the variable in `join()` has a path separator:

```rust
let x = "/bar";
let y = Path::new("/foo").join(x); // y = "bar"
```

clippy_lints/src/path_from_format.rs
merelymyself: iirc, it was because using `format!` was not recommended by another clippy lint

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,136 @@
...
+                    push_comps(&mut sugg, Path::new(&string));
+                }
+                span_lint_and_sugg(
+                    cx,
+                    PATH_FROM_FORMAT,
+                    expr.span,
+                    "`format!(..)` used to form `PathBuf`",
+                    "consider using `Path::new()` and `.join()` to make it OS-agnostic and improve code readability",
```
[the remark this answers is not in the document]
merelymyself: I don't really mind either way, but I feel that having consistency is slightly more important. Plus, I don't really see long paths that often in Rust projects (or is that just confirmation bias?)

merelymyself:
Once again, thanks for the patience and advice, @dswij! Do you mind taking another look? Besides the length thing, I think most of the suggestions have been resolved.

merelymyself:
Hi, @dswij. Again, thanks for all the help. Do you mind taking another look?

clippy_lints/src/lib.register_lints.rs
```diff
@@ -458,6 +458,7 @@ store.register_lints(&[
     pass_by_ref_or_value::LARGE_TYPES_PASSED_BY_VALUE,
     pass_by_ref_or_value::TRIVIALLY_COPY_PASS_BY_REF,
     path_buf_push_overwrite::PATH_BUF_PUSH_OVERWRITE,
+    path_from_format::PATH_FROM_FORMAT,
```
dswij, line 461: Clippy has a [naming convention for lints](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#lints) to prefer plural form. So the name would be `path_from_formats` following the convention

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,130 @@
...
+                let order_of_real_vars: Vec<usize> = format_args.formatters.iter().map(|(x, _)| *x).collect();
+                let mut arguments_in_order = Vec::new();
+                for n in 0..real_vars.len() {
+                    arguments_in_order.push(real_vars[order_of_real_vars[n]].clone());
+                }
+                let mut paths_zip = string_parts.iter().take(arguments_in_order.len()).zip(arguments_in_order);
+                let mut sugg = String::new();
+                for n in 0..real_vars.len() {
+                    if let Some((part, arg)) = paths_zip.next() {
+                        if is_valid_use_case(string_parts[n], string_parts.get(n+1).unwrap_or(&"")) {
```
dswij, line 69: Will this always be true `string_parts.len` >= `real_vars.len`? I'd prefer if we avoid indexing so we're sure that it won't panic

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+fn push_comps(string: &mut String, path: &str, trim_first_slash: bool) {
+    let mut path = path.to_string();
+    if (path.starts_with('/') || path.starts_with('\\')) && trim_first_slash {
+        path.remove(0);
+    }
+    let path = Path::new(&path);
+    let comps = path.components();
+    for n in comps {
```
dswij, line 114:
```suggestion
    for component in Path::new(&path).components() {
```

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,130 @@
...
+        let mut x = n.as_os_str().to_string_lossy().to_string();
+        if x.ends_with('/') || x.ends_with('\\') {
+            x.pop();
+        }
```
dswij, line 118: We can use [`str::trim_right_matches`](https://doc.rust-lang.org/std/primitive.str.html#method.trim_end_matches)

clippy_lints/src/path_from_format.rs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    }
+}
+
+fn push_comps(string: &mut String, path: &str, trim_first_slash: bool) {
+    let mut path = path.to_string();
+    if (path.starts_with('/') || path.starts_with('\\')) && trim_first_slash {
+        path.remove(0);
+    }
```
dswij, line 111: we can use `trim_start_matches` with a `Pattern` instead.

dswij (commented):
Sorry for the super long delay and thank you for the patience! I've had some tight schedules lately.

I left some comments - feel free to ping me here or elsewhere. I might be more responsive in Zulip

clippy_lints/src/lib.register_lints.rs
merelymyself: Should it be `paths_from_format` instead, or am I missing something? @dswij

merelymyself:
Hi, @dswij . Sorry for the delay. Mind taking another look? Also, I renamed it to `paths_from_format` instead of `path_from_formats` because of the guidelines; "allow paths from format" makes more sense.

dswij (commented):
I've given a quick look, and it looks good overall. I'll give a full review soon

merelymyself:
btw, I'm not sure why the CI is red now - when I run `cargo dev update_lints --check` on my side no errors are thrown.

merelymyself:
Hi, @dswij ! Sorry to disturb, but do you mind taking a look soon?

xFrednet:
Hey @dswij, this is a ping from triage, could you review this PR? If you don't have the time, feel free to pass it to me.

@merelymyself Would you mind rebasing the branch on master? IDK, what changes have been made to `src/docs.rs` in the meantime, but I hope that the rebase will go smoothly. You're welcome to reach out, if you need assistance :)

dswij:
@xFrednet Thanks for taking time to look at this. I'll pass this to you if you don't mind :)

r? @xFrednet

clippy_lints/src/paths_from_format.rs
```diff
@@ -0,0 +1,124 @@
...
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `PathBuf::from(format!(..))` calls.
+    ///
+    /// ### Why is this bad?
+    /// It is not OS-agnostic, and can be harder to read.
+    ///
+    /// ### Known Problems
+    /// `.join()` introduces additional allocations that are not present when `PathBuf::push` is
+    /// used instead.
```
xFrednet, line 22: Looking at this, I think it would be better to suggest `PathBuf` directly. Even if that might require additional modifications by the user. `PathBuf` is intended for modifications as this one AFAIK :)

Though, in that case we should have the applicability be less than machine applicable

clippy_lints/src/paths_from_format.rs
```diff
@@ -0,0 +1,124 @@
...
+impl<'tcx> LateLintPass<'tcx> for PathsFromFormat {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if_chain! {
+            if let ExprKind::Call(_, args) = expr.kind;
+            if let ty = cx.typeck_results().expr_ty(expr);
+            if is_type_diagnostic_item(cx, ty, sym::PathBuf);
+            if !args.is_empty();
+            if let Some(macro_call) = root_macro_call(args[0].span);
+            if cx.tcx.item_name(macro_call.def_id) == sym::format;
+            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &args[0], macro_call.expn);
```
xFrednet, line 53: I know there has been quite some work, on format macro diagnostics, since this PR. Hopefully, a rebase is all we need :)

clippy_lints/src/paths_from_format.rs
```diff
@@ -0,0 +1,124 @@
...
+    pedantic,
+    "builds a `PathBuf` from a format macro"
+}
+
+declare_lint_pass!(PathsFromFormat => [PATHS_FROM_FORMAT]);
+
+impl<'tcx> LateLintPass<'tcx> for PathsFromFormat {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if_chain! {
+            if let ExprKind::Call(_, args) = expr.kind;
```
xFrednet, line 47:
```suggestion
            if let ExprKind::Call(_, [arg, ..]) = expr.kind;
```

This checks during the match that `args` is not empty and avoids indexing in later code. Indexing has caused some errors in the past, so it's usually nicer to match directly :)

clippy_lints/src/paths_from_format.rs
```diff
@@ -0,0 +1,124 @@
...
+            if !args.is_empty();
+            if let Some(macro_call) = root_macro_call(args[0].span);
+            if cx.tcx.item_name(macro_call.def_id) == sym::format;
+            if let Some(format_args) = FormatArgsExpn::find_nested(cx, &args[0], macro_call.expn);
+            then {
+                let format_string_parts = format_args.format_string.parts;
+                let format_value_args = format_args.args;
+                let string_parts: Vec<&str> = format_string_parts.iter().map(rustc_span::Symbol::as_str).collect();
+                let mut applicability = Applicability::MachineApplicable;
+                let real_vars: Vec<Sugg<'_>> = format_value_args.iter().map(|x| Sugg::hir_with_applicability(cx, x.param.value, "..", &mut applicability)).collect();
```
xFrednet, line 59: rustfmt sadly doesn't work inside `if_chain` macros. Would you mind either manually formatting this, or better rewrite the if to use let chains?

clippy_lints/src/paths_from_format.rs
```diff
@@ -0,0 +1,124 @@
...
+            x = x.trim_end_matches(|c| c == '/' || c == '\\').to_string();
+            let _ = write!(string, ".join(\"{x}\")");
+        }
+    }
+}
+
+fn is_valid_use_case(string: &str, string2: &str) -> bool {
+    !(string.is_empty() || string.ends_with('/') || string.ends_with('\\'))
+        || !(string2.is_empty() || string2.starts_with('/') || string2.starts_with('\\'))
+}
```
xFrednet, line 124: I'm a bit unsure, why these condition make it a valid use. Could you provide an example and maybe add a doc comment for it?

xFrednet (commented):
The basic structure and idea is already good. Though, while reading though the documentation, I found examples, that suggested a different way. I'll add a comment to the original issue and ask the author :)

@rustbot author

clippy_lints/src/paths_from_format.rs
merelymyself: Sorry, to clarify, you mean you want to do some or all of the following?
1. Replace `.join()` with `.push()`
2. Replace the `Path::new` with `PathBuf::from`
3. Instead of suggesting a replacement, just flag it out to the user

xFrednet: Sorry, if my comment wasn't clear. I mean a combination of the first two. So suggest `PathBuf::from()` and then add segments using `push()`. The thing is, that this probably change the type of the buffer and require further adjustments by the user. Therefore, we should display the suggestion, but not make it machine applicable.

Does this make sense, or should I try to describe it better? :)

merelymyself: I'm looking at the docs for `PathBuf` now, and I just realised that there might be a recommended solution for situations like this :face_with_spiral_eyes:.

`[r"C:\", "windows", "system32.dll"].iter().collect::<PathBuf>()`

Will you be ok with this? Given it should still be machine applicable but I don't think it will have the same problems as before.

Jarcho: You have to be careful when using the array as all the values have to be the same type.

clippy_lints/src/paths_from_format.rs
```diff
@@ -0,0 +1,132 @@
...
+        } else {
+            x = x.trim_end_matches(|c| c == '/' || c == '\\').to_string();
+            let _ = write!(string, ".join(\"{x}\")");
+        }
+    }
+}
+
+// In essence, this checks if the format!() used is made for concatenation of the filenames / folder
+// names itself. This returns true when it is something like
+// `PathBuf::from(format!("/x/folder{}/textfile.txt", folder_number))`
```
xFrednet, line 128: Based on the comment in the related issue, I would actually think that such a paht can still be problematic. I'd suggest either, checking if the path starts with `//?/` which makes it a verbatim path under windows or (IMO the better approach) add a config value, to allow this. I have the feeling the lint can be very noisy without, but ignoring these cases can also be wrong.

xFrednet (commented):
Hey, thank you for your patience, the review sadly took a bit longer. You're welcome to ask if you have any questions :)

xFrednet:
Hey @merelymyself this is a ping from triage. Are you still working on this PR? If you have any questions, you're always welcome to reach out :)

merelymyself:
Hi, so sorry. Thanks for all the help so far, but I don't think I'll have much time in the near future to work on this with any level of consistency - again, apologies for any inconvenience caused. I'll close the PR.

xFrednet:
It was no inconvenience at all, I hope you still had some fun!

I'll mark this PR as `S-inactive-closed` it's still a good start for anyone that wants to work on the issue.

Issue #8758, opened 2022-04-28
 dbg-macro should tolerate use of dbg! in tests (redux, even with --tests)


ijackson:
### Summary

`cargo clippy --tests -- -D clippy::dbg-macro` complains about use of `dbg!` in modules marked `#[cfg(test)]`.

(In #8481 I reported this false positive for test functions themselves and that was fixed in #8500, but the problem remains in non-test-entrypoint functions in test modules.)

### Lint Name

dbg-macro

### Reproducer

I tried this code:

```rust
fn main() {
}

#[cfg(test)]
mod test {
    fn wombat() {
        dbg!("wombat");
    }

    #[test]
    fn t() {
        wombat();
    }
}
```

and ran `cargo clippy --locked --offline --tests -- -D clippy::dbg-macro`.

I saw this happen:

```
    Checking foo v0.1.0 (/volatile/rustcargo/Rustup/Arti/experiments)
error: `dbg!` macro is intended as a debugging tool
 --> src/main.rs:7:9
  |
7 |         dbg!("wombat");
  |         ^^^^^^^^^^^^^^
  |
  = note: requested on the command line with `-D clippy::dbg-macro`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#dbg_macro
help: ensure to avoid having uses of it in version control
  |
7 |         "wombat";
  |         ~~~~~~~~

error: could not compile `foo` due to previous error
```

I expected to see this happen: the code should be accepted without complaint.

IMO `--tests` ought to mean just "check my tests too, but treating them as tests", not "apply to my tests the rules for non-test code".

### Version

```text
rustc 1.61.0-nightly (1bfe40d11 2022-03-18)
binary: rustc
commit-hash: 1bfe40d11c3630254504fb73eeccfca28d50df52
commit-date: 2022-03-18
host: x86_64-unknown-linux-gnu
release: 1.61.0-nightly
LLVM version: 14.0.0
```

PR #8838, opened 2022-05-17
[dbg_macro] tolerates use of `dbg!` in items which have `#[cfg(test)]` attribute

tamaroning:
fix: #8758 
changelog: [dbg_macro] tolerates use of `dbg!` in items with `#[cfg(test)]` attribute

### DIFF
--- a/clippy_lints/src/dbg_macro.rs
+++ b/clippy_lints/src/dbg_macro.rs
@@ -1,7 +1,7 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::is_in_test_function;
 use clippy_utils::macros::root_macro_call_first_node;
 use clippy_utils::source::snippet_with_applicability;
+use clippy_utils::{is_in_cfg_test, is_in_test_function};
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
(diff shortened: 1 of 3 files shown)


Jarcho (changes requested):
Overall looks good. Just two little style related nits.

tamaroning:
@Jarcho 
Thank you for your review. I fixed them.

xFrednet (approved):
This looks good to me. I'll leave the final r+ to @Jarcho 

@bors delegate @Jarcho

@Jarchen when you merge, it would be cool if you could use `r=Jarcho,xFrednet` :upside_down_face:

xFrednet:
@bors delegate @Jarcho

(Bors doesn't work in review comments)

xFrednet:
@bors delegate=Jarcho

(Bors doesn't work in review comments)

Jarcho:
Everything looks good. @bors r=Jarcho,xFrednet

afetisov:
That is now a very confusing behaviour on the part of that lint. When I run it, I expect it to find _all_ uses of dbg, regardless of test or no test. When I use dbg in tests, it's also for debugging purposes, and leaving it in will just cause tests to spam debug messages in the console.

A change like that should have been a separate lint with a clear name, which could be separately allowed.

xFrednet:
I would dislike splitting the lint, but we could introduce a config value, which enables the user to specify the behavior in tests. Do you think that is a better idea? @afetisov

afetisov:
I don't have any objections to a config value, I would just want the lint to be enabled everywhere by default.

afetisov:
I think a separate lint is a bit better for discoverability, since Clippy will print its name when it triggers. Config values are afaik mentioned only in the docs. But from a practical perspective, a config value indeed makes more sense.

xFrednet:
The question would be how the lints, should be split. It wouldn't make sense to have one lint that tests everything and one that is specific to anything but functions, as they would overlap. Therefore, we would need one lint for tests and one for other code. For me, that would be more confusing than a config value. But I'm also used to working with the lint list ^^

tamaroning:
I think it is better to introduce a config value too. Mentioning it in warning messages of [dbg_macro] might be useful.

xFrednet:
Would you mind adding it? We have some [documentation](https://github.com/rust-lang/rust-clippy/blob/master/doc/adding_lints.md#adding-configuration-to-a-lint) how this can be done. If not, we can just create a new issue :upside_down_face:

PR #8841, opened 2022-05-19
Remove code block from `pub_enum_variant_names`

Serial-ATA:
changelog: none

Just noticed this empty code block that no other lint has :smile:.
old

r? @xFrednet


xFrednet:
Bors has been a bit slow today :thinking: 

@bors r+

PR #8843, opened 2022-05-19
Collect renamed lints

Serial-ATA:
changelog: Display past names of renamed lints on Clippy's lint list

cc #7172

r? @xFrednet

### DIFF
--- a/clippy_lints/src/utils/internal_lints/metadata_collector.rs
+++ b/clippy_lints/src/utils/internal_lints/metadata_collector.rs
@@ -7,6 +7,7 @@
 //! during any comparison or mapping. (Please take care of this, it's not fun to spend time on such
 //! a simple mistake)
 
+use crate::renamed_lints::RENAMED_LINTS;
 use crate::utils::internal_lints::{extract_clippy_version_value, is_lint_ref_type};
 
 use clippy_utils::diagnostics::span_lint;
@@ -26,6 +27,7 @@ use rustc_span::{sym, Loc, Span, Symbol};
 use serde::{ser::SerializeStruct, Serialize, Serializer};
 use std::collections::BinaryHeap;
 use std::fmt;
+use std::fmt::Write as _;
 use std::fs::{self, OpenOptions};
 use std::io::prelude::*;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


xFrednet:
Have you thought about how we want to display the renamed lint name on the website?

I was considering something like this:

image

Or maybe

image

(Though I like the first option more)

In either case, we can prepare do some work in out metadata collector to make the website faster. That's why I'm asking :upside_down_face: 

r? @dswij

Serial-ATA:
I had a similar idea to the first one, except it was "Past names" instead of "Notes". Since it'll show up in search regardless, I don't think making a new entry to point to the rename is necessary. 🙂

xFrednet:
> I had a similar idea to the first one, except it was "Past names" instead of "Notes". Since it'll show up in search regardless, I don't think making a new entry to point to the rename is necessary.

Past names is a better name. In that case, I would already add this to the documentation in rust, similar to how it's done for configurations. :upside_down_face:

dswij:
Was just thinking if `Notes` can even include recent changelogs to the lint, but that's probably too much noise

What happens when we rename a lint twice? Would `Past Names` have all past lint names?

Serial-ATA:
I updated it to account for multiple renames :slightly_smiling_face:.

xFrednet:
Nice, this still leaves the question where these will be rendered. I'm still for adding it to the lint description like it's done for the configuration :upside_down_face:

Serial-ATA:
Yeah, I think it should go under the configuration as a bulleted list.

xFrednet:
What I meant with my question of *where this will be rendered* was, where in our *tech-stack*. So if this will append the section in the metadata collector (Like the configuration section) or on the website. Where possible, I usually prefer adding it during the collection in rust, since any logic on Clippy's lint list is multiplied by the number of lints and delays the rendering. That's also why I'm mentioning this here and not in the related frontend PR since, we can implement this without touching the frontend. :upside_down_face:

Serial-ATA:
Oh :smile:. In that case, yeah it should definitely be added like the configuration. I'll get that done soon.

Serial-ATA:
Here's how it looks:
example

xFrednet:
The example looks good IMO :+1:

flip1995:
The `Past Names` option also was my first thought when seeing this PR in my notifications :+1:

xFrednet:
I'm stealing this review back. I hope that's alright with you @dswij 

r? @xFrednet

dswij:
Sure, go ahead!

xFrednet (approved):

xFrednet:
Excellent work! I've also tested it locally and everything works perfectly. Thank you very much for this update. I'm happy to see the metadata collection project to make progress and with your work probably also be completed! :heart: This was my first big project in Clippy, and it's also the one that held me here and introduced me to the team. I hope you also had fun with it.

Now enough rambling:

@bors r+

xFrednet:
Sorry for the r-, this is worth a changelog entry. I don't want this to slip under the radar of who ever will write it for that version. I've added the entry and also ticked of the checkbox in the tracking issue

@bors r+

PR #8844, opened 2022-05-19
Check `.fixed` paths' existence in `run_ui`

smoelius:
This PR adds a test to check that there exists a `.fixed` file for every `.stderr` file in `tests/ui` that mentions a `MachineApplicable` lint. The test leverages `compiletest-rs`'s `rustfix_coverage` option.

I tried to add `.fixed` files where they appeared to be missing. However, 38 exceptional `.rs` files remain. Several of those include comments indicating that they are exceptions, though not all do. Apologies, as I have not tried to associate the 38 files with GH issues. (I think that would be a lot of work, and I worry about linking the wrong issue.)

changelog: none


smoelius:
> If we add something like this, I think we should add some option to disable it for specific test files.

I'm not sure if this is what you have in mind, but that sounds like what I did here: https://github.com/rust-lang/rust-clippy/pull/8844/files#diff-30869434167a8e693c01b4875d4855efeb3a28567f6531765b202e82cc398a99R59

tests/dogfood.rs
```diff
@@ -25,43 +25,206 @@ fn dogfood_clippy() {
...
+    }
+
+    #[derive(Debug, Deserialize)]
+    struct ApplicabilityInfo {
+        applicability: Option<String>,
+    }
+
+    const MACHINE_APPLICABLE: &str = "MachineApplicable";
+
+    const KNOWN_EXCEPTIONS: [&str; 33] = [
```
xFrednet, line 59: Here you can use a reference to not hard-code the array length :upside_down_face: 
```suggestion
    const KNOWN_EXCEPTIONS: &[&str] = &[
```

xFrednet:
It looks like nobody has any concerns regarding this change, and I also like it.

r? @Alexendoo

Alexendoo:
I like the idea, trying it locally it does take quite a while to run though, took up to a minute for me. `compiletest` has [`rustfix_coverage`](https://github.com/Manishearth/compiletest-rs/blob/b8bd2107d820572331efc1c55e63bfc9e55329b6/src/common.rs#L224-L227) for this that we could enable in `run_ui`, it creates a text file with the list of relevant files

Unfortunately the version of `rustfix` in the fork crashes when this is enabled, it's fixed upstream (https://github.com/rust-lang/rustfix/pull/199) but I don't know what the process for updating the `compiletest` fork would be (sync all the upstream changes or just bump the `rustfix` version?)

smoelius:
> I don't know what the process for updating the `compiletest` fork would be (sync all the upstream changes or just bump the `rustfix` version?)

Sorry, but I don't understand what the two alternatives are (possibly because I don't know what "the `compiletest` fork" refers to).

One alternative is to bump the `rustfix` version in `compiletest`, and once `compiletest` is updated, (possibly) bump the `compiletest` version in Ciippy (correct?).

What is the other alternative?

Alexendoo:
Clippy uses https://github.com/Manishearth/compiletest-rs, which is extracted from https://github.com/rust-lang/rust/tree/master/src/tools/compiletest. The one in `rust-lang/rust` has the updated rustfix dep of `0.6.0`, but I wasn't sure what the usual way to get upstream changes into the `compiletest-rs` repo is

The sync all thing would be to round up all of the upstream changes in one PR, but looking at other PRs to the repo I don't *think* that's how it's done. So a PR that bumps the `rustfix` version alone should be good

smoelius:
Submitted. I'll update this PR once a new version of `compiletest-rs` is published.

smoelius:
@Alexendoo I'm sorry if you were telling me this and I didn't get it: https://github.com/rust-lang/rustfix/pull/199 is not in `rustfix` 0.6.0. So, I've requested a new release of `rustfix`.

Alexendoo:
Oh whoops I thought it was, my mistake

smoelius:
I've updated the PR to use `rustfix_coverage`. You're welcome to review whenever you have time.

tests/compile-test.rs
```diff
@@ -337,6 +339,82 @@ fn compile_test() {
...
+fn check_rustfix_coverage() {
+    let missing_coverage_path = Path::new("target/debug/test/ui/rustfix_missing_coverage.txt");
+
+    if let Ok(missing_coverage_contents) = std::fs::read_to_string(missing_coverage_path) {
+        assert!(RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS.is_sorted());
+
+        for rs_path in missing_coverage_contents.lines() {
+            let filename = rs_path.strip_prefix("tests/ui/").unwrap();
+            assert!(
+                RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS.binary_search(&filename).is_ok(),
```
Alexendoo, line 392: So it works on windows also:

```suggestion
            let filename = Path::new(rs_path).strip_prefix("tests/ui/").unwrap();
            assert!(
                RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS.binary_search_by_key(&filename, Path::new).is_ok(),
```

tests/compile-test.rs
```diff
@@ -337,6 +339,82 @@ fn compile_test() {
...
+                "`{}` runs `MachineApplicable` diagnostics but is missing a `run-rustfix` annotation",
```
Alexendoo, line 393: Would be good to have a message here that tells people what to do to resolve this, e.g. to add `// run-rustfix` to the top of the test file or if that doesn't work to add it as an exception

tests/compile-test.rs
```diff
@@ -162,8 +163,8 @@ fn base_config(test_dir: &str) -> compiletest::Config {
 }
 
 fn run_ui() {
-    let config = base_config("ui");
-    // use tests/clippy.toml
```
Alexendoo, line 166: I'd like to keep this comment since it explains what the `CARGO_MANIFEST_DIR` `VarGuard` is for

Alexendoo (commented):
Looks great thanks!

You can also now merge these into their respective main test files, or delete them if the contents are duplicates

- `tests/ui/needless_late_init_fixable.rs`
- `tests/ui/unnecessary_cast_fixable.rs`

tests/compile-test.rs
Alexendoo: May also want to update the `is_sorted` call accordingly

smoelius:
@Alexendoo I think I have addressed all of your comments.

tests/compile-test.rs
smoelius: Apologies. I thought I introduced that comment by mistake.

tests/compile-test.rs
```diff
@@ -337,6 +340,86 @@ fn compile_test() {
...
+
+        for rs_path in missing_coverage_contents.lines() {
+            let filename = Path::new(rs_path).strip_prefix("tests/ui/").unwrap();
+            assert!(
+                RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS
+                    .binary_search_by_key(&filename, Path::new)
+                    .is_ok(),
+                "`{}` runs `MachineApplicable` diagnostics but is missing a `run-rustfix` annotation. \
+                Please either add `// run-rustfix` at the top of file or add the file to \
+                `RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS`.",
```
Alexendoo, line 398: Just one last thing! Could you add `in tests/compile-test.rs` or similar, I think I would have assumed `RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS` to be an environment variable if I didn't already know otherwise

Alexendoo (approved):
Thanks @smoelius!

LGTM @xFrednet

tests/compile-test.rs
```diff
@@ -337,6 +340,86 @@ fn compile_test() {
...
+
+        for rs_path in missing_coverage_contents.lines() {
+            let filename = Path::new(rs_path).strip_prefix("tests/ui/").unwrap();
+            assert!(
+                RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS
+                    .binary_search_by_key(&filename, Path::new)
+                    .is_ok(),
+                "`{}` runs `MachineApplicable` diagnostics but is missing a `run-rustfix` annotation. \
+                Please either add `// run-rustfix` at the top of the file or add the file to \
+                `RUSTFIX_COVERAGE_KNOWN_EXCEPTIONS` in `tests/compile-test.rs`.",
```
xFrednet, line 398: I think this output would look better, if we added some `\n`s

xFrednet (approved):

xFrednet:
Looks good to me, thank you for this addition :)

Also thank you, @Alexendoo, for the nice suggestion with the updated dependency. I wouldn't have known about that!

@bors r=Alexendoo,xFrednet

xFrednet:
Hmm, some file is missing only in one container. This might be some CI weirdness

@bors r+

xFrednet:
r-

Alexendoo:
👀 Oh whoops that's my fault #8896

Maybe it would be worth ignoring the `crashes` dir though

smoelius:
> 👀 Oh whoops that's my fault #8896
> 
> Maybe it would be worth ignoring the `crashes` dir though

In a way, this is a success! :D

What is your preference? I rebase and fix this, or ignore the the crashes dir (any maybe revert the changes thereto)?

smoelius:
I pushed a commit for ice-8850. Please let me know if you would rather ignore the crashes dir.

xFrednet:
I think it would be good to ignore the crashes directory. The focus of those UI tests is not to test the lint, but the lint implementation itself. Either option would be correct, though. If it's simple to do, I would appreciate that change. :upside_down_face:

smoelius:
Oops, I forgot about Windows again.

xFrednet:
This version looks good to me, I'll leave the final r+ to @Alexendoo :upside_down_face: 

@bors delegate=Alexendoo

Alexendoo:
Looks all good to me too, thanks @smoelius 

@bors r+

PR #8848, opened 2022-05-19
Auto-detect preferred colorscheme

Serial-ATA:
changelog: none

This just sets the theme to coal by default if the user prefers dark color schemes.

r? @xFrednet

### DIFF
--- a/util/gh-pages/script.js
+++ b/util/gh-pages/script.js
@@ -343,17 +343,23 @@ function setTheme(theme, store) {
     let enableNight = false;
     let enableAyu = false;
 
-    if (theme == "ayu") {
-        enableAyu = true;
-    } else if (theme == "coal" || theme == "navy") {
-        enableNight = true;
-    } else if (theme == "rust") {
-        enableHighlight = true;
-    } else {
-        enableHighlight = true;
-        // this makes sure that an unknown theme request gets set to a known one
-        theme = "light";
+    switch(theme) {
+        case "ayu":
+            enableAyu = true;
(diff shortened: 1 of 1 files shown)


xFrednet (approved):
I love this change, the only suggestion I have is to have the default dark theme be `navy` since I like that one a bit more. This is very subjective, though. Would that change be okay with you? And also out of interest, which is your preferred theme?

util/gh-pages/script.js
```diff
@@ -368,4 +374,10 @@ function setTheme(theme, store) {
 }
 
 // loading the theme after the initial load
-setTheme(localStorage.getItem('clippy-lint-list-theme'), false);
+const prefersDark = window.matchMedia("(prefers-color-scheme: dark)");
+const theme = localStorage.getItem('clippy-lint-list-theme');
+if (prefersDark.matches && !theme) {
+    setTheme("coal", false);
```
xFrednet, line 380:
```suggestion
    setTheme("navy", false);
```

Serial-ATA: Navy feels a little too blue to me to be a default dark theme. I feel like the night and day of coal and light is more of what you'd expect when you visit a site offering a dark theme. Also to answer your question, I think coal is the best theme (and rust is definitely my least favorite :smile:).

xFrednet:
Yeah, the `rust` theme always felt a bit off to me ^^. Then we'll keep it as `coal`, thank you for this update!!!

@bors r+

PR #8857, opened 2022-05-20
Add test for #8855

### LINKED ISSUE #8855: to_string() in a format string when format specifiers are used (eg :<20) is NOT redundant
### Summary

It is not true that
a) println!("{:<20}", struct.to_string());
gives the same result as
b) println!("{:<20}", struct)

It depend on how the formatter is written.
Line a) uses the default "{}" display to render a string then renders the string using the formatting rules :<20
Whereas b) uses the formatting rules of the struct's display, not the string's display trait impl.

Many dsiplay impls for objects do not have logic to cater for alignment for instance, but using struct.to_string() and then formatting with alignment format specifiiers DOES give alignment.

Clippy "fixing" such lines results in behavioural change.

I think clippy should flag only when the format is "{}"

### Lint Name

to_string_in_format_args

### Reproducer

I tried this code:

```rust
     write!(
                f,
                "{:>6} {:>10} {:>3} {:>2}",
                self.bm.uci(),
                self.score.to_string(),
                self.depth,
                self.nt
            )
```

I saw this happen:

```
6   | #![warn(clippy::perf)]
    |         ^^^^^^^^^^^^
    = note: `#[warn(clippy::to_string_in_format_args)]` implied by `#[warn(clippy::perf)]`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#to_string_in_format_args

warning: `to_string` applied to a type that implements `Display` in `println!` args
   --> src/comms/bench.rs:118:19
    |
118 |                 cp.to_string(),
    |                   ^^^^^^^^^^^^ help: remove this
```

This is related to #7729

### Version

```text
rustc 1.60.0 (7737e0b5c 2022-04-04)
binary: rustc
commit-hash: 7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c
commit-date: 2022-04-04
release: 1.60.0
LLVM version: 14.0.0
```

### Additional Labels

_No response_

smoelius:
Fix #8855

Here is what I think is going on.

First, the expression `format!("{:>6} {:>6}", a, b.to_string())` expands to:
```rust
{
    let res =
        ::alloc::fmt::format(::core::fmt::Arguments::new_v1_formatted(&["",
                            " "],
                &[::core::fmt::ArgumentV1::new_display(&a),
                            ::core::fmt::ArgumentV1::new_display(&b.to_string())],
                &[::core::fmt::rt::v1::Argument {
                                position: 0usize,
                                format: ::core::fmt::rt::v1::FormatSpec {
                                    fill: ' ',
                                    align: ::core::fmt::rt::v1::Alignment::Right,
                                    flags: 0u32,
                                    precision: ::core::fmt::rt::v1::Count::Implied,
                                    width: ::core::fmt::rt::v1::Count::Is(6usize),
                                },
                            },
                            ::core::fmt::rt::v1::Argument {
                                position: 1usize,
                                format: ::core::fmt::rt::v1::FormatSpec {
                                    fill: ' ',
                                    align: ::core::fmt::rt::v1::Alignment::Right,
                                    flags: 0u32,
                                    precision: ::core::fmt::rt::v1::Count::Implied,
                                    width: ::core::fmt::rt::v1::Count::Is(6usize),
                                },
                            }], unsafe { ::core::fmt::UnsafeArg::new() }));
    res
}
```
When I dump the expressions that get past the call to `has_string_formatting` [here](https://github.com/rust-lang/rust-clippy/blob/b312ad7d0cf0f30be2bd4658b71a3520a2e76709/clippy_lints/src/format_args.rs#L83), I see more than I would expect.

In particular, I see this subexpression of the above:
```
                &[::core::fmt::ArgumentV1::new_display(&a),
                            ::core::fmt::ArgumentV1::new_display(&b.to_string())],
```

This suggests to me that more expressions are getting past [this call](https://github.com/rust-lang/rust-clippy/blob/b312ad7d0cf0f30be2bd4658b71a3520a2e76709/clippy_lints/src/format_args.rs#L71) to `FormatArgsExpn::parse` than should.

Those expressions are then visited, but no `::core::fmt::rt::v1::Argument`s are found and pushed [here](https://github.com/rust-lang/rust-clippy/blob/b312ad7d0cf0f30be2bd4658b71a3520a2e76709/clippy_utils/src/macros.rs#L407).

As a result, the expressions appear unformatted, hence, the false positive.

My proposed fix is to restrict `FormatArgsExpn::parse` so that it only matches `Call` expressions.

cc: @akanalytics

changelog: none

### DIFF
--- a/tests/ui/format_args.fixed
+++ b/tests/ui/format_args.fixed
@@ -122,3 +122,27 @@ fn issue8643(vendor_id: usize, product_id: usize, name: &str) {
         name
     );
 }
+
+// https://github.com/rust-lang/rust-clippy/issues/8855
+mod issue_8855 {
+    #![allow(dead_code)]
+
+    struct A {}
+
+    impl std::fmt::Display for A {
+        fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+            write!(f, "test")
+        }
+    }
+
+    fn main() {
+        let a = A {};
+        let b = A {};
+
+        let x = format!("{} {}", a, b);
+        dbg!(x);
+
+        let x = format!("{:>6} {:>6}", a, b.to_string());
+        dbg!(x);
+    }
+}
--- a/tests/ui/format_args.rs
+++ b/tests/ui/format_args.rs
@@ -122,3 +122,27 @@ fn issue8643(vendor_id: usize, product_id: usize, name: &str) {
         name
     );
 }
+
+// https://github.com/rust-lang/rust-clippy/issues/8855
+mod issue_8855 {
+    #![allow(dead_code)]
+
+    struct A {}
+
+    impl std::fmt::Display for A {
+        fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+            write!(f, "test")
+        }
+    }
+
+    fn main() {
+        let a = A {};
+        let b = A {};
+
+        let x = format!("{} {}", a, b.to_string());
+        dbg!(x);
+
+        let x = format!("{:>6} {:>6}", a, b.to_string());
+        dbg!(x);
+    }
+}
--- a/tests/ui/format_args.stderr
+++ b/tests/ui/format_args.stderr
@@ -126,5 +126,11 @@ error: `to_string` applied to a type that implements `Display` in `println!` arg
 LL |     println!("{foo}{bar}", bar = "bar", foo = "foo".to_string());
    |                                                    ^^^^^^^^^^^^ help: remove this
 
-error: aborting due to 21 previous errors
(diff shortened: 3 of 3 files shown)


flip1995:
Thanks for this. As you pointed out in #9349 this is closely related to the other format/write/print PRs. So I'll mark this a `blocked` until those are merged and will then test if this fix is still necessary or if the rewrite already resolved it somehow.

smoelius:
#9349 does seem to have resolved the issue.

I think the only question that remains is whether you would like this PR's test.

flip1995:
Ah great! Thanks for revisiting this. I would have gone back to this once the write.rs rewrite was merged. We definitely want the regression test.

@bors r+

PR #8862, opened 2022-05-21
`get_last_with_len`: lint `VecDeque` and any deref to slice

Alexendoo:
changelog: [`get_last_with_len`]: lint `VecDeque` and any deref to slice

Previously only `Vec`s were linted, this will now catch any usages on slices, arrays, etc. It also suggests `.back()` for `VecDeque`s

Also moves the lint into `methods/`


clippy_lints/src/methods/get_last_with_len.rs
```diff
@@ -0,0 +1,55 @@
...
+use rustc_hir::{BinOpKind, Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_middle::ty;
+use rustc_span::source_map::Spanned;
+use rustc_span::sym;
+
+use super::GET_LAST_WITH_LEN;
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, recv: &Expr<'_>, arg: &Expr<'_>) {
+    if let Some(recv_local) = path_to_local(recv)
```
Jarcho, line 15: This is significantly more restrictive than it used to be. Something like `x.y.get(..)` or `x[0].get(..)` would have been linted before.

I'd think going back to using `SpanlessEq`, but adding a call to `Expr::can_have_side_effects` would be better.

clippy_lints/src/methods/get_last_with_len.rs
```diff
@@ -0,0 +1,55 @@
...
+        && lhs_path.ident.name == sym::len
+
+        // RHS of subtraction is 1
+        && let ExprKind::Lit(rhs_lit) = &rhs.kind
+        && let LitKind::Int(1, ..) = rhs_lit.node
+
+        // check that recv == lhs_recv `recv.get(lhs_recv.len() - 1)`
+        && path_to_local(lhs_recv) == Some(recv_local)
+    {
+        let method = match cx.typeck_results().expr_ty_adjusted(recv).peel_refs().kind() {
```
Jarcho, line 36: Quick question here. This allows any type which derefs to a `VecDeque` or a slice. Is that intentional?

Alexendoo: It is yeah

Jarcho:
Can you add a test for a field access as well just to make sure it doesn't break in the future.

Otherwise looks good. @xFrednet I don't know if you want to add anything before merging.

clippy_lints/src/methods/get_last_with_len.rs
Jarcho: I don't know if you want to handle this horrible case or not:
```rust
struct Foo(Vec<u32>);
impl Deref for Foo { type Target = Vec<u32>; ... }
impl Foo {
    // Only replace `last`, but not `get`
    fn last(&self) -> SomeType { ... }
}
```

I don't think it's necessary to do so. It's kind of a pathological case.

Alexendoo:
👍 added

clippy_lints/src/methods/get_last_with_len.rs
Alexendoo: I'm happy to leave that one out, can't imagine it coming up

Jarcho (approved):

xFrednet (approved):

xFrednet:
I'm a big fan of this change, nice cleanup!!! :muscle: 

Thank you, @Jarcho, for the review.

@bors r=Jarcho,xFrednet

xFrednet:
Weird, seems like some CI weirdness. Maybe bors wanted a break ^^

@bors retry

PR #8866, opened 2022-05-21
Add new lint `[unused_rounding]`

### LINKED ISSUE #39: use of round/floor/ceil on known integer value
E.g. ceil(1f32)

botahamec:
fixes #39 

changelog: added a ``[`unused_rounding`]`` lint to check for the rounding of whole-number literals

### DIFF
--- a/clippy_lints/src/unused_rounding.rs
+++ b/clippy_lints/src/unused_rounding.rs
@@ -1,0 +1,69 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use rustc_ast::ast::{Expr, ExprKind, LitFloatType, LitKind};
+use rustc_errors::Applicability;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Detects cases where a whole-number literal float is being rounded, using
+    /// the `floor`, `ceil`, or `round` methods.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// This is unnecessary and confusing to the reader. Doing this is probably a mistake.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let x = 1f32.ceil();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let x = 1f32;
+    /// ```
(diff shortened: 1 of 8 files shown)


clippy_lints/src/unused_rounding.rs
```diff
@@ -0,0 +1,69 @@
...
+        && !args.is_empty()
+        && let ExprKind::Lit(spanned) = &args[0].kind
+        && let LitKind::Float(symbol, ty) = spanned.kind {
+            let f = symbol.as_str().parse::<f64>().unwrap();
+            let f_str = symbol.to_string() + if let LitFloatType::Suffixed(ty) = ty {
+                ty.name_str()
+            } else {
+                ""
+            };
+            if f.fract() < f64::EPSILON {
```
mikerite, line 45: Is the test that `f` is a whole number? `f64::EPSILON` is 2^(-53) but it's possible for a float to be smaller than that.  Try `let _ = 2E-54f64.floor();` in your tests.

clippy_lints/src/unused_rounding.rs
```diff
@@ -0,0 +1,69 @@
...
+    /// let x = 1f32.ceil();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let x = 1f32;
+    /// ```
+    #[clippy::version = "1.62.0"]
+    pub UNUSED_ROUNDING,
+    nursery,
+    "Rounding a whole number literal, which is useless"
```
llogiq, line 28:
```suggestion
    "Uselessly rounding a whole number floating-point literal"
```

llogiq (commented):
Good job! 👍

I'd also like to see the test @mikerite proposed. And I wonder if the lint should be omitted in test code (we have an `in_test` function in `clippy_utils`, but it requires a `LateContext`).

clippy_lints/src/unused_rounding.rs
botahamec: You're right. Thanks! This should be fixed now

botahamec:
@llogiq I made both changes. I'm not sure about omitting it in tests, so I left that the same for now. What would be the reason for doing that?

botahamec:
Ah shoot, I messed up with the rebase, didn't I?

llogiq:
Thank you!

@bors r+

PR #8871, opened 2022-05-22
Add `cargo dev deprecate`

Serial-ATA:
changelog: none

I wrote this awhile ago when `regex` was still a dependency. Is it alright to add it back?


giraffate (commented):
IMO `deprecate` command is good idea. However, in https://github.com/rust-lang/rust-clippy/pull/8605, we removed `regex` from dependencies because of making command faster, so it would be good not to add it again if possible. cc @Jarcho

Jarcho:
You could just change `gather_all` to store the location of each lint declaration. Saves the need to re-parse the file.

I removed `regex` because it was a third of the compilation time for `clippy-dev`. Since it frequently ends up being recompiled it's a nice QOL change. The runtime speed increase was also a nice bonus, but that doesn't apply to this case.

Serial-ATA:
Makes sense, thanks for the clarification 👍.

clippy_dev/src/update_lints.rs
```diff
@@ -316,6 +317,168 @@ pub fn rename(old_name: &str, new_name: &str, uplift: bool) {
...
+        let stripped = content[impl_lint_pass_start..impl_lint_pass_end]
+            .lines()
+            .filter(|l| !l.contains(lint_name_upper))
+            .fold(String::new(), |s, line| s + line + "\n");
+
+        content.replace_range(impl_lint_pass_start..impl_lint_pass_end, &*stripped);
+    }
+
+    if path.exists() {
+        // We can't delete the entire file, just remove the declaration
```
flip1995, line 416: In this case I would emit a message, that code will have to be removed from the module.

flip1995 (commented):
This is a great addition I wanted for a long time, thanks for implementing this!

I only skimmed through the code, so this isn't an in-depth review.

clippy_dev/src/update_lints.rs
Serial-ATA: Good call!

giraffate:
I'll review this in this week.

giraffate (commented):
A simple case works well. For example, in deprecating `needless_late_init`, it works because there is `clippy_lints/src/needless_late_init.rs` and so.
```
➜  rust-clippy git:(pr_8871) cargo dev deprecate needless_late_init
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/clippy_dev deprecate needless_late_init`
`needless_late_init` has successfully been deprecated
note: the deprecation reason must be updated in `clippy_lints/src/deprecated_lints.rs`
note: you must run `cargo uitest` to update the test results
```

However, for example, I don't know how to deprecate `needless_return`. It's included in `clippy_lints/src/returns.rs`. I tried cases below, but it doesn't work. How can I do this?
```
➜  rust-clippy git:(pr_8871) cargo dev deprecate needless_return           
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/clippy_dev deprecate needless_return`
error: lint not found
➜  rust-clippy git:(pr_8871) cargo dev deprecate needless_return returns
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/clippy_dev deprecate needless_return returns`
error: lint not found
```

clippy_dev/src/update_lints.rs
```diff
@@ -326,6 +327,198 @@ pub fn rename(old_name: &str, new_name: &str, uplift: bool) {
...
+    }
+
+    if path.exists() {
+        if let Some(lint) = lints.iter().find(|l| l.name == name) {
+            if lint.module == name {
+                // The lint name is the same as the file, we can just delete the entire file
+                fs::remove_file(path)?;
+            } else {
+                // We can't delete the entire file, just remove the declaration
+                if lint.module != name {
```
giraffate, line 443: It looks like that this `if` expression doesn't be needed.

giraffate (commented):
Thanks for addressing this!

I tried another case, so I found a case which happen ICE. Can this case be addressed?
```
➜  rust-clippy git:(pr_8871) RUST_BACKTRACE=1 cargo dev deprecate unnecessary_fold
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/clippy_dev deprecate unnecessary_fold`
warn: you will have to manually remove any code related to `unnecessary_fold` from `clippy_lints/src/methods/unnecessary_fold.rs`
thread 'main' panicked at 'byte index 42187 is out of bounds of `use clippy_utils::diagnostics::span_lint_and_sugg;
use clippy_utils::source::snippet_with_applicability;
use clippy_utils::{is_trait_method, path_to_local_id, peel_blocks, strip_pat_refs};
use if_chain::if_chain;
use rustc_ast::ast;
use rustc_errors::Appli`[...]', library/core/src/str/mod.rs:107:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/panicking.rs:142:14
   2: core::str::slice_error_fail_rt
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/str/mod.rs:107:9
   3: core::ops::function::FnOnce::call_once
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/ops/function.rs:248:5
   4: core::intrinsics::const_eval_select
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/intrinsics.rs:2375:5
   5: core::str::slice_error_fail
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/str/mod.rs:86:9
   6: core::str::traits::<impl core::slice::index::SliceIndex<str> for core::ops::range::Range<usize>>::index
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/str/traits.rs:218:21
   7: core::str::traits::<impl core::ops::index::Index<I> for str>::index
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/str/traits.rs:65:9
   8: <alloc::string::String as core::ops::index::Index<core::ops::range::Range<usize>>>::index
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/alloc/src/string.rs:2315:10
   9: clippy_dev::update_lints::remove_lint_declaration
             at ./clippy_dev/src/update_lints.rs:459:25
  10: clippy_dev::update_lints::deprecate
             at ./clippy_dev/src/update_lints.rs:380:8
  11: clippy_dev::main
             at ./clippy_dev/src/main.rs:92:13
  12: core::ops::function::FnOnce::call_once
             at /rustc/b31f9cc22bcd720b37ddf927afe378108a5b9a54/library/core/src/ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

clippy_dev/src/update_lints.rs
Serial-ATA: Copy and paste mistake :smile:. Good catch

clippy_lints/src/utils/internal_lints.rs
```diff
@@ -375,42 +417,67 @@ pub struct LintWithoutLintPass {
...
+                                &format!("the lint `{}` has the default lint description", item.ident.name),
+                            );
+                        }
+
+                        self.declared_lints.insert(item.ident.name, item.span);
+                    } else if sym_str == "default deprecation note" {
                         span_lint(
                             cx,
-                            DEFAULT_LINT,
+                            DEFAULT_DEPRECATION_REASON,
```
giraffate, line 475: Can tests for this internal lint be added? It would be better to add tests to `tests/ui-internal/`.

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #8872, opened 2022-05-23
update dependencies

hellow554:
changelog: none

Updating some dependencies, but the biggest one is clap to v3, but without using the derive struct thingy.


llogiq:
Thank you!

@bors r+

PR #8881, opened 2022-05-24
Fix imports for "Checking if a type defines a specific method"

dmarcoux:
The import of `clippy_utils::is_type_diagnostic_item` would cause this error:

```
error[E0432]: unresolved import `clippy_utils::is_type_diagnostic_item
```

changelog: none


llogiq:
Thank you!

@bors r+

PR #8882, opened 2022-05-24
feat(lint): impl lint about use first() instead of get(0)

### LINKED ISSUE #8851: New lint: use `slice.first()` instead of `slice.get(0)`
### What it does

Detects `slice.get(0)` and suggests replacement with `slice.first()`.

### Lint Name

slice_get_zero

### Category

style

### Advantage

No magic numbers.

### Drawbacks

`first()` is longer than `get(0)` by 1 character.

### Example

```rust
let v = vec![0, 1, 2];
println!("{}", vec.get(0));
```

Could be written as:

```rust
let v = vec![0, 1, 2];
println!("{}", vec.first());
```

kyoto7250:
close #8851 

This PR adds new lint about considering replacing .get(0) with .first().

Thank you in advance.

changelog: adds new lint [`get_first`] to consider replacing .get(0) with .first()

### DIFF
--- a/clippy_lints/src/get_first.rs
+++ b/clippy_lints/src/get_first.rs
@@ -1,0 +1,72 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::source::snippet_with_applicability;
+use clippy_utils::{is_slice_of_primitives, match_def_path, paths};
+use if_chain::if_chain;
+use rustc_ast::LitKind;
+use rustc_errors::Applicability;
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::source_map::Spanned;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for using `x.get(0)` instead of
+    /// `x.first()`.
+    ///
+    /// ### Why is this bad?
+    /// Using `x.first()` is easier to read and has the same
+    /// result.
+    ///
+    /// ### Example
+    /// ```rust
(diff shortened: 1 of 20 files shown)


clippy_lints/src/get_first.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub GET_FIRST,
+    style,
+    "Using `x.get(0)` when `x.first()` is simpler"
+}
+declare_lint_pass!(GetFirst => [GET_FIRST]);
+
+impl<'tcx> LateLintPass<'tcx> for GetFirst {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>) {
+        if_chain! {
+            if let hir::ExprKind::MethodCall(_, expr_args, _) = &expr.kind;
```
llogiq, line 43: Using
```suggestion
            if let hir::ExprKind::MethodCall(_, [struct_calling_on, method_arg], _) = &expr.kind;
```
would make the `expr_args.len() == 2` check and the `expr_args.get(_)` checks obsolete.

llogiq (commented):
This is already looking quite good. I only have one small suggestion.

clippy_lints/src/get_first.rs
kyoto7250: Fixed in [this commit](https://github.com/rust-lang/rust-clippy/pull/8882/commits/d0f93c12a2a6f0d693507ff6325fed62720e616b).

kyoto7250:
@llogiq 

I rebased to the latest master and committed your suggestion.

llogiq:
Thank you!

@bors r+

PR #8896, opened 2022-05-26
Only return `DefId`s to `Fn`-like definitions in `clippy_utils::fn_def_id`

### LINKED ISSUE #8850: clippy bug report
[bug.txt](https://github.com/rust-lang/rust-clippy/files/8738400/bug.txt)

Alexendoo:
Fixes #8850

in `returns.rs` `tcx.fn_sig` is called on the result of `fn_def_id`, which panics if the def is a `const`/`static`/etc rather than a functions definition

https://github.com/rust-lang/rust-clippy/blob/bc4d39e5fea64970dded1e6d2132d41435c0ef24/clippy_lints/src/returns.rs#L294-L303

changelog: Fix ICE due to callable `static`/`const`s

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -1916,7 +1916,17 @@ pub fn fn_def_id(cx: &LateContext<'_>, expr: &Expr<'_>) -> Option<DefId> {
                 ..
             },
             ..,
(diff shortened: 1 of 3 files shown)


llogiq:
Thank you!

@bors r+

PR #8899, opened 2022-05-26
Fix `[use_self]` false negative with on struct and tuple struct patterns

### LINKED ISSUE #8845: `use_self` false negative on tuple or struct enum variants
### Summary

`use_self` doesn't trigger for struct or tuple enum variants, only empty variants.

There are false negatives mentioned in the documention for [use_self](https://rust-lang.github.io/rust-clippy/master/index.html#use_self) but I couldn't find an issue tracking them. I don't know if these are the same known false negatives or not.

### Lint Name

use_self

### Reproducer

I tried this code:

```rust
#![warn(clippy::use_self)]
#![allow(unused)]

pub enum Something {
    Empty,
    Num(u8),
    TupleNums(u8, u8),
    StructNums { one: u8, two: u8 },
}

impl Something {
    const fn get_value(&self) -> u8 {
        match self {
            Something::Empty => 0,
            Something::Num(n) => *n,                      
            Something::TupleNums(n, _m) => *n,            
            Something::StructNums { one, two: _ } => *one,
        }                                                 
    }                                                     
}                                                         
```

I should get a warning for all four uses of `Something` in get_value.

Instead, this happened:

I only get a warning for `Something::Empty`. Once I update it to `Self::Empty` I get no more warnings.

### Version

```text
binary: rustc
commit-date: 2022-05-18
release: 1.61.0
LLVM version: 14.0.0
```

botahamec:
fixes #8845 

changelog: Triggered the warning for ``[`use_self`]`` on `TupleStruct` and `Struct` patterns, whereas currently it's only triggered for `Path` patterns

### DIFF
--- a/tests/ui/use_self.rs
+++ b/tests/ui/use_self.rs
@@ -542,3 +542,69 @@ mod use_self_in_pat {
         }
     }
 }
+
+mod issue8845 {
+    pub enum Something {
+        Num(u8),
+        TupleNums(u8, u8),
+        StructNums { one: u8, two: u8 },
+    }
+
+    struct Foo(u8);
+
+    struct Bar {
+        x: u8,
+        y: usize,
+    }
+
+    impl Something {
+        fn get_value(&self) -> u8 {
+            match self {
+                Something::Num(n) => *n,
+                Something::TupleNums(n, _m) => *n,
+                Something::StructNums { one, two: _ } => *one,
+            }
+        }
+
+        fn use_crate(&self) -> u8 {
+            match self {
+                crate::issue8845::Something::Num(n) => *n,
+                crate::issue8845::Something::TupleNums(n, _m) => *n,
+                crate::issue8845::Something::StructNums { one, two: _ } => *one,
+            }
+        }
+
+        fn imported_values(&self) -> u8 {
+            use Something::*;
+            match self {
+                Num(n) => *n,
+                TupleNums(n, _m) => *n,
+                StructNums { one, two: _ } => *one,
+            }
+        }
+    }
+
+    impl Foo {
+        fn get_value(&self) -> u8 {
+            let Foo(x) = self;
+            *x
+        }
+
+        fn use_crate(&self) -> u8 {
+            let crate::issue8845::Foo(x) = self;
+            *x
+        }
+    }
+
+    impl Bar {
+        fn get_value(&self) -> u8 {
+            let Bar { x, .. } = self;
+            *x
+        }
+
+        fn use_crate(&self) -> u8 {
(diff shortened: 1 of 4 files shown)


Alexendoo:
👋

tests/ui/use_self.rs
```diff
@@ -542,3 +542,42 @@ mod use_self_in_pat {
...
+
+    impl Something {
+        const fn get_value(&self) -> u8 {
+            match self {
+                Something::Num(n) => *n,
+                Something::TupleNums(n, _m) => *n,
+                Something::StructNums { one, two: _ } => *one,
+            }
+        }
+    }
```
Alexendoo, line 568: Using the variants directly trigger the lint, suggesting to replace them all with `Self`:

```rust
fn imported_variants(&self) -> u8 {
    use Something::*;

    match self {
        Num(n) => *n,
        TupleNums(n, _m) => *n,
        StructNums { one, two: _ } => *one,
    }
}
```

Alexendoo: Seems that's uncovering an existing bug, the first path segment is picked unconditionally 

```rust
enum Enum {
    A,
}

impl Enum {
    fn f(self) -> usize {
        match self {
            crate::Enum::A => 1,
        }
    }
}
```

```
warning: unnecessary structure name repetition
  --> t.rs:13:13
   |
13 |             crate::Enum::A => 1,
   |             ^^^^^ help: use the applicable keyword: `Self`
   |
```

You could check the `res` field of the `PathSegment`s to pick the right one

botahamec: This was tricky. Using the `res` field on the `PathSegment`s doesn't seem to work. In the playground, I found that most of the `res` fields had errors, so I ended up using a workaround similar to what's done in the `check_expr` method. Luckily, the `lint_path_to_variant` function already exists. But, it's a bit hacky, and I did find that the name of the enum is correctly shown in the `res` fields. I made a version of the function which uses the `res` field, and is a bit less hacky, but it's also a bit less performant. I doubt anyone would notice the difference, but I left it unchanged for now.

Alexendoo:
Thanks! Could you also squash the commits and then I think it'll be good to go

botahamec:
@Alexendoo Should be good now

xFrednet:
@Alexendoo when you're happy with this PR you can r+ it directly. And we no longer need a delegate for that :partying_face:

Alexendoo (approved):

Alexendoo:
Well then. Guess the teams PR needs to happen for that yet 😄

xFrednet:
*Talking in bors voice*

Thank you for the approval

@bors r=Alexendoo

PR #8901, opened 2022-05-26
Rework `branches_sharing_code`

### LINKED ISSUE #7378: branches_sharing_code seems to ignore names in let bindings
When using an `if` to unpack and process some data with a `let` binding, the `branches_sharing_code` lint triggers, even though I'm unpacking the data differently in each branch.

I included a simplified example below. In reality, I'm not just returning the extracted information in a different type, but actually using it to process some data. Following the suggestion would change the behaviour of the code and introduce a bug.

Lint name: `branches_sharing_code`

Minimal example:

```rust
struct Dimensions {
    width: usize,
    height: usize,
    channels: usize,
}

fn extract_size(data_shape: (usize, usize, usize), planar: bool) -> Dimensions {
    if planar {
        let (channels, width, height) = data_shape;
        Dimensions { width, height, channels }
    } else {
        let (width, height, channels) = data_shape;
        Dimensions { width, height, channels }
    }
}
```

The `branches_sharing_code` lint triggers unexpectedly, and it gives a suggestion that changes the behaviour of the code:

```
warning: all if blocks contain the same code at the start
  --> src/main.rs:9:5
   |
9  | /     if planar {
10 | |         let (channels, width, height) = data_shape;
   | |___________________________________________________^
   |
   = note: `#[warn(clippy::branches_sharing_code)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#branches_sharing_code
help: consider moving the start statements out like this
   |
9  |     let (channels, width, height) = data_shape;
10 |     if planar {
   |
```

### Meta

- `cargo clippy -V`: clippy 0.1.54 (ce1d561 2021-06-18)
- `rustc -Vv`:
```
rustc 1.55.0-nightly (ce1d5611a 2021-06-18)
binary: rustc
commit-hash: ce1d5611a28468663e275078649e7ca6eef735a8
commit-date: 2021-06-18
release: 1.55.0-nightly
LLVM version: 12.0.1
```

Jarcho:
fixes #7378

This changes the lint from checking pairs of blocks, to checking all the blocks at the same time. As such there's almost none of the original code left.

changelog: Don't lint `branches_sharing_code` when using different binding names

### DIFF
--- a/clippy_lints/src/copies.rs
+++ b/clippy_lints/src/copies.rs
@@ -1,18 +1,18 @@
 use clippy_utils::diagnostics::{span_lint_and_note, span_lint_and_then};
 use clippy_utils::source::{first_line_of_span, indent_of, reindent_multiline, snippet, snippet_opt};
 use clippy_utils::{
-    both, count_eq, eq_expr_value, get_enclosing_block, get_parent_expr, if_sequence, is_else_clause, is_lint_allowed,
-    search_same, ContainsName, SpanlessEq, SpanlessHash,
+    eq_expr_value, get_enclosing_block, hash_expr, hash_stmt, if_sequence, is_else_clause, is_lint_allowed,
+    search_same, ContainsName, HirEqInterExpr, SpanlessEq,
 };
-use if_chain::if_chain;
-use rustc_data_structures::fx::FxHashSet;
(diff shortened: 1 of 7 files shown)


dswij (approved):
Changes looks great to me, thanks!

@bors r+

PR #8906, opened 2022-05-27
remove `large_enum_variant` suggestion for `Copy` types

### LINKED ISSUE #8894: large_enum_variant should not lint types that are copy
### Summary

````rust
#[derive(Copy, Clone)]
enum Test {
    A(i32),
    B([i32; 8000]),
}
````
the lint will fire here, but it seems that `Box` does not impl the `Copy` trait...?

````
error[E0204]: the trait `Copy` may not be implemented for this type
 --> src/main.rs:1:10
  |
1 | #[derive(Copy, Clone)]
  |          ^^^^
...
4 |     B(Box<[i32; 8000]>),
  |       ---------------- this field does not implement `Copy`
  |
````

### Lint Name

large_enum_variant

llogiq:
Replaces the (erroneous) suggestion on `large_enum_variant` for `Copy` types by a note. This fixes #8894.

---

changelog: none

### DIFF
--- a/clippy_lints/src/large_enum_variant.rs
+++ b/clippy_lints/src/large_enum_variant.rs
@@ -1,7 +1,7 @@
 //! lint when there is a large size difference between variants on an enum
 
-use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::source::snippet_with_applicability;
+use clippy_utils::{diagnostics::span_lint_and_then, ty::is_copy};
 use rustc_errors::Applicability;
 use rustc_hir::{Item, ItemKind};
 use rustc_lint::{LateContext, LateLintPass};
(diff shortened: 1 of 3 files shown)


llogiq:
We may want to suggest moving data to auxillary data structures instead, but that's very dependent on the use case.

Jarcho:
`is_copy` won't work on generic types without a `ParamEnv` matching the constraints on the impl.

llogiq:
So do we need to walk all `TraitImpl`s if the type is generic? Or is there an easier way that I've overlooked?

Jarcho:
I think so. There's no way to know what constraints are applied to the impl without seeing it first. If `TyCtxt::trait_impls_of` had public fields this would be easy.

llogiq:
So I tried to create a test case for this and failed as the lint ignores all generic enum definitions (because of the `return` in line 99). I now have some doubts that trying to get the trait impls is time well spent.

Jarcho:
I guess we can safely ignore that case then. I don't think the impl can even have generics unless either the struct or the trait have them.

llogiq:
I went ahead and added a check for generic Copy impls, using the local trait impls. Unless we have a *lot* of `Copy` impls in a crate, this shouldn't be a problem. Otherwise I might add a cache for the type def ids to avoid the quadratic behavior.

Jarcho:
I've been going through available queries and [non_blanket_impls_for_ty](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.non_blanket_impls_for_ty) looks like it would be a better way to check.

llogiq:
No idea why the job was canceled, but the last commit only changed formatting to be in line with what rustfmt would do if it had let chains, so meh.

Jarcho (approved):

Jarcho:
I guess bors took a nap.

Everything lgtm.

llogiq:
Seems @camsteffen is busy, so r? @Alexendoo

Alexendoo:
👍

@bors r=Jarcho

PR #8912, opened 2022-05-29
needless_late_init: fix ICE when all branches return the never type

### LINKED ISSUE #8911: ICE `assertion failed: !suggestion.is_empty()`
### Summary

```rust
#[allow(unused)]
#[rustfmt::skip]
fn f() {
    let x: ();
    match todo!() {
        _ if {x = (); false} => panic!(),
        _ => panic!(),
    }

    todo!()
}

```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=48f1a5df9248aa1ee64ce8b8d126730c))

### Version

```text
Clippy version: clippy 0.1.63 (1fede17 2022-05-28)
```

### Error output

```
thread 'rustc' panicked at 'assertion failed: !suggestion.is_empty()', /rustc/1fede1753c33f4ce0660ad9b8edbd618d9733daf/compiler/rustc_errors/src/diagnostic.rs:552:9

note: Clippy version: clippy 0.1.63 (1fede17 2022-05-28)

query stack during panic:
#0 [analysis] running analysis passes on this crate
end of query stack
error: could not compile `playground`
```

Alexendoo:
Fixes #8911

When the assignment is done in a match guard or the if condition and all of the branches return the never type `assignment_suggestions` would return an empty `Vec` which caused the ICE. It now returns `None` in that scenario

Also moves some tests to the top of the file

changelog: ICE Fixes: [`needless_late_init`] #8911

### DIFF
--- a/clippy_lints/src/needless_late_init.rs
+++ b/clippy_lints/src/needless_late_init.rs
@@ -194,14 +194,15 @@ fn assignment_suggestions<'tcx>(
         }))
         .collect::<Option<Vec<(Span, String)>>>()?;
 
-    let applicability = if suggestions.len() > 1 {
+    match suggestions.len() {
+        // All of `exprs` are never types
+        // https://github.com/rust-lang/rust-clippy/issues/8911
+        0 => None,
+        1 => Some((Applicability::MachineApplicable, suggestions)),
         // multiple suggestions don't work with rustfix in multipart_suggest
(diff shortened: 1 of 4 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #8930, opened 2022-06-02
fix(manual_find_map and manual_filter_map): check clone method

### LINKED ISSUE #8920: Lint request: filter(is_some).map(unwrap)
### What it does

Lint about usage of `.filter(/* ... */).map(/* ... */)` calls that could be simplified to `.filter_map(/* ... */)`.

### Lint Name

filter_then_map

### Category

_No response_

### Advantage

- Removes an unnecessary `.unwrap()`
  - … or `.expect()`
  - … or `.unwrap_or*` with a never-used default value / expression
- Makes the code easier to read and shorter

### Drawbacks

_No response_

### Example

```rust
struct Foo {
    field: Option<String>,
}

fn main() {
    let vec = vec![Foo {
        field: Some("value".to_owned()),
    }];
    let iter = vec
        .iter()
        .filter(|f| f.field.is_some())
        .map(|f| f.field.clone().unwrap());

    for field in iter {
        dbg!(field);
    }
}
```

Could be written as:

```rust
struct Foo {
    field: Option<String>,
}

fn main() {
    let vec = vec![Foo {
        field: Some("value".to_owned()),
    }];
    let iter = vec.iter().filter_map(|f| f.field.clone());

    for field in iter {
        dbg!(field);
    }
}
```

kyoto7250:
close #8920 

Added conditional branching when the clone method is used.

Thank you in advance.

---

changelog: check `clone()` and other variant preserving methods in [`manual_find_map`] and [`manual_filter_map`]

### DIFF
--- a/clippy_lints/src/methods/filter_map.rs
+++ b/clippy_lints/src/methods/filter_map.rs
@@ -181,9 +181,11 @@ pub(super) fn check<'tcx>(
 }
 
 fn acceptable_methods(method: &PathSegment<'_>) -> bool {
-    let methods: [Symbol; 6] = [
+    let methods: [Symbol; 8] = [
         sym::clone,
         sym::as_ref,
+        sym!(copied),
+        sym!(cloned),
         sym!(as_deref),
         sym!(as_mut),
         sym!(as_deref_mut),
--- a/clippy_lints/src/methods/filter_map.rs
+++ b/clippy_lints/src/methods/filter_map.rs
@@ -155,7 +155,15 @@ pub(super) fn check<'tcx>(
                 }
                 false
             };
-            if SpanlessEq::new(cx).expr_fallback(eq_fallback).eq_expr(filter_arg, map_arg);
+
+            if match map_arg.kind {
+                ExprKind::MethodCall(clone, [original_arg], _) => {
+                    clone.ident.name == sym::clone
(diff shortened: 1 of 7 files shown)


kyoto7250:
@Alexendoo 

I commited your suggestion.
Could you review again?

clippy_lints/src/methods/filter_map.rs
```diff
@@ -171,3 +179,16 @@ pub(super) fn check<'tcx>(
...
 }
+
+fn acceptable_methods(method: &PathSegment<'_>) -> bool {
+    let methods: [Symbol; 6] = [
+        sym::clone,
+        sym::as_ref,
+        sym!(as_deref),
+        sym!(as_mut),
+        sym!(as_deref_mut),
+        sym!(to_owned),
```
Alexendoo, line 190: `cloned` and `copied` should also be applicable

kyoto7250: Sorry, I missed.

I fixed by [this commit](https://github.com/rust-lang/rust-clippy/pull/8930/commits/42cf98553a72f4b6cdac2c4c6378121d30f3bf5f), thank you!

Alexendoo (approved):
nice, thanks!

PR #8933, opened 2022-06-03
Add new lint [`needless_parens_on_range_literals`]

DennisOSRM:
changelog: Adds a new lint [`needless_parens_on_range_literals`] to warn on needless braces on literals in a range statement

For example, the lint would catch

```log
error: needless parenthesis on range literals can be removed
  --> $DIR/needless_parens_on_range_literals.rs:8:13
   |
LL |     let _ = ('a')..=('z');
   |             ^^^^^ help: try: `'a'`
   |
   = note: `-D clippy::needless-parens-on-range-literals` implied by `-D warnings`
```


mikerite:
Not a review. Just some comments.

...

AFAIK, Rust uses the term parenthesis for `(...)` and braces for `{...}` so the name might be wrong.

Here is an example where the suggestion fails because the `f64` has a trailing `.`.
```
fn main() {
    use rand::{thread_rng, Rng};
    
    let mut rng = thread_rng();
    
    let m: f64 = rng.gen_range((1.)..2.);
    println!("{}", m);
}
```

How does the lint handle macros? E.g. if the parens and literal are in a macro, does it recommend replacing the macro call with a literal? Maybe add some tests for that.

clippy_lints/src/needless_braces_on_range_literal.rs
```diff
@@ -0,0 +1,67 @@
...
+//!
+//! For example, the lint would catch
+//!
+//! ```rust
+//! for i in (0)..10 {
+//!   println!("{i}");
+//! }
+//! ```
+//!
+//! And suggest this:
```
Serial-ATA, line 11:
```suggestion
//! Use instead:
```
For consistency with other lints :)

clippy_lints/src/needless_braces_on_range_literal.rs
```diff
@@ -0,0 +1,67 @@
...
+//!
+//! ```rust
+//! for i in 0..10 {
+//!   println!("{i}");
+//! }
+//! ```
+//!
+//! This lint is **warn** by default.
```
Serial-ATA, line 19:
```suggestion
```

clippy_lints/src/needless_braces_on_range_literal.rs
```diff
@@ -0,0 +1,67 @@
...
+
+declare_clippy_lint! {
+  /// ### What it does
+  /// The lint checks for braces on literals in range statements that are
+  /// superflous.
+  ///
+  /// ### Why is this bad?
+  /// Having superflous braces makes the code less legible as the impose an
+  /// overhead when reading.
+
```
Serial-ATA, line 35: It looks like you took the examples out of here and put them at the top of the file. They need to be within this macro for the docs to appear on the [lint list](https://rust-lang.github.io/rust-clippy/master).

clippy_lints/src/needless_braces_on_range_literal.rs
```diff
@@ -0,0 +1,67 @@
...
+                  diag.span_suggestion(e.span, "try", suggestion, Applicability::MachineApplicable);
+              }
+        });
+    }
+    }
+}
+
+impl EarlyLintPass for NeedlessBracesOnRangeLiteral {
+    fn check_expr(&mut self, cx: &EarlyContext<'_>, e: &Expr) {
+        if let ExprKind::Range(Some(start), Some(end), _) = &e.kind {
```
Serial-ATA, line 62: What about the following cases:
```rust
let _ = ('a')..;
let _ = ..('z');
```

clippy_lints/src/needless_braces_on_range_literal.rs
```diff
@@ -0,0 +1,67 @@
...
+    if_chain! {
+      if let ExprKind::Paren(ref start_statement) = &e.kind;
+    if let ExprKind::Lit(ref literal) = start_statement.kind;
+    then {
+      span_lint_and_then(cx, NEEDLESS_BRACES_ON_RANGE_LITERAL, e.span,
+        "needless braces on range literal can be removed",
+        |diag| {
+                if let Some(suggestion) = snippet_opt(cx, literal.span) {
+                  diag.span_suggestion(e.span, "try", suggestion, Applicability::MachineApplicable);
+              }
```
Serial-ATA, line 54: I think this is better expressed with `snippet_with_applicability`.

Serial-ATA (changes requested):
Welcome :slightly_smiling_face:!

I'm not the reviewer, but I have some comments on top of @mikerite's.

DennisOSRM:
> Not a review. Just some comments.
> 
> ...
> 
> AFAIK, Rust uses the term parenthesis for `(...)` and braces for `{...}` so the name might be wrong.
> 
> Here is an example where the suggestion fails because the `f64` has a trailing `.`.
> 
> ```
> fn main() {
>     use rand::{thread_rng, Rng};
>     
>     let mut rng = thread_rng();
>     
>     let m: f64 = rng.gen_range((1.)..2.);
>     println!("{}", m);
> }
> ```
> 
> How does the lint handle macros? E.g. if the parens and literal are in a macro, does it recommend replacing the macro call with a literal? Maybe add some tests for that.

That's a great suggestion. Thanks, will add it

clippy_lints/src/needless_braces_on_range_literal.rs
DennisOSRM: Implemented in afe37e1

clippy_lints/src/needless_braces_on_range_literal.rs
DennisOSRM: Implemented in 496155d

clippy_lints/src/needless_braces_on_range_literal.rs
DennisOSRM: Implemented in 8185ecd7

DennisOSRM:
Rewrote the check by `impl LateLintPass`. This gives access to type information and makes it easy to ignore parenthesis on a floating point start literal. The detection of parenthesis is done by inspecting the span of the corresponding expression and whether its length lines up with the length of the literal expression.

dswij (commented):

1. The lint itself looks good, just some nits on spellings.
2. Can you do a rebase instead of merge commits? We follow [a no-merge policy ](https://github.com/rust-lang/rust-clippy/blob/master/doc/basics.md#pr)
3. We prefer plural form when [naming lint](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints). So the lint name would be `needless_parens_on_range_literals`. Can you also add the name to the changelog?


---

Also, huge thanks to @mikerite and @Serial-ATA for the suggestions and reviews.

DennisOSRM:
> Thanks for the PR, and welcome!
> 
> 1. The lint itself looks good, just some nits on spellings.
> 2. Can you do a rebase instead of merge commits? We follow [a no-merge policy ](https://github.com/rust-lang/rust-clippy/blob/master/doc/basics.md#pr)
> 3. We prefer plural form when [naming lint](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints). So the lint name would be `needless_parens_on_range_literals`. Can you also add the name to the changelog?
> 
> Also, huge thanks to @mikerite and @Serial-ATA for the suggestions and reviews!

Thanks a lot. Did the rebase. The merge commits are now gone. Also, followed your suggestion to use the plural form. CHANGELOG.md is also updated.

dswij (approved):

dswij:
Thanks for this! The lint looks great to me.

@bors r+

PR #8939, opened 2022-06-04
Suggest `pointer::cast` when possible in `transmute_ptr_to_ref`

### LINKED ISSUE #8924: `transmute_ptr_to_ref` false positive for types containing differing lifetimes
### Summary

There's one last FP in #2906 that wasn't fixed

### Lint Name

transmute_ptr_to_ref

### Reproducer

I tried this code:

```rust
#![allow(clippy::missing_safety_doc)]

use std::mem::transmute;

pub unsafe fn ptr_to_ref<'a, 'b>(ptr: *const &'a ()) -> &&'b () {
    transmute(ptr)
}
```

I saw this happen:

```
warning: transmute from a pointer type (`*const &()`) to a reference type (`&&()`)
 --> t.rs:5:5
  |
5 |     std::mem::transmute(ptr)
  |     ^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `&*ptr`
  |
```

It shouldn't lint, as `&*ptr` would fail to compile

### Version

_No response_

### Additional Labels

_No response_

Jarcho:
fixes #8924

changelog: Suggest casting the pointer for any type containing lifetimes in `transmute_ptr_to_ref`.
changelog: Suggest `pointer::cast` when possible in `transmute_ptr_to_ref`.

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -641,7 +641,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|| Box::new(borrow_deref_ref::BorrowDerefRef));
     store.register_late_pass(|| Box::new(no_effect::NoEffect));
     store.register_late_pass(|| Box::new(temporary_assignment::TemporaryAssignment));
-    store.register_late_pass(|| Box::new(transmute::Transmute));
+    store.register_late_pass(move || Box::new(transmute::Transmute::new(msrv)));
     let cognitive_complexity_threshold = conf.cognitive_complexity_threshold;
     store.register_late_pass(move || {
         Box::new(cognitive_complexity::CognitiveComplexity::new(
--- a/clippy_lints/src/transmute/mod.rs
+++ b/clippy_lints/src/transmute/mod.rs
@@ -16,9 +16,10 @@ mod wrong_transmute;
 
 use clippy_utils::in_constant;
 use if_chain::if_chain;
-use rustc_hir::{Expr, ExprKind};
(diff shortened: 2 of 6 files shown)


Jarcho:
Apparently casting `*const &'a T` to `*const &'b T` directly doesn't work with `as`. Not sure if double cast as a workaround is really better than transmute.

clippy_utils/src/msrvs.rs
```diff
@@ -25,7 +25,7 @@ msrv_aliases! {
     1,40,0 { MEM_TAKE, NON_EXHAUSTIVE, OPTION_AS_DEREF }
     1,38,0 { POINTER_CAST }
     1,37,0 { TYPE_ALIAS_ENUM_VARIANTS }
-    1,36,0 { ITERATOR_COPIED }
+    1,36,0 { ITERATOR_COPIED, PTR_CAST }
```
giraffate, line 28: What method is this? I thought it was [this](https://doc.rust-lang.org/std/primitive.pointer.html#method.cast), but it looks like this was added at 1.38.0.

giraffate (commented):
Sorry for the late reviewing. I made comments.

clippy_utils/src/msrvs.rs
Jarcho: That is the right method. I don't know where I got 1.36 from.

giraffate: Okay. Can `1,38,0 { POINTER_CAST }` be used for msrv?

giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #8941, opened 2022-06-04
improve [`for_loops_over_fallibles`] to detect the usage of iter, iter_mut and into_iterator

### LINKED ISSUE #6762: Opt-in warn when using iter()-like methods on Option or Result
### What it does

Warn when iter, iter_mut or into_iter is used on an Option or Result.

### Categories (optional)

- Kind: clippy::pedantic

Treating Option/Result as normal iterables is sometimes useful, but also sometimes produces really confusing behavior, especially on types like `Option<Vec<_>>`. See below for an example where a hidden bug gets uncovered by removing the obfuscation caused by treating Option/Result as an iterable.

### Drawbacks

Treating Option/Result as iterables is sometimes useful, for example in long clever adapter chains. Personally I still think it's a code smell to do that, but to each their own.

### Example

```rust
pub fn get_item_list() -> Result<Vec<Item>, Error> { /* ... */ }

// Confusing, isn't it?
for item in get_item_list() {
    // ...
}
```

Could be written as:

```rust
pub fn get_item_list() -> Result<Vec<Item>, Error> { /* ... */ }

// Now it's obvious what's happening
if let Ok(item) = get_item_list() {
    // ...
}
```

DevAccentor:
fix #6762

detects code like 
```rust
for _ in option.iter() {
    //..
}
```

changelog: Improve [`for_loops_over_fallibles`] to detect `for _ in option.iter() {}` or using `iter_mut()` or `into_iterator()`.

### DIFF
--- a/tests/ui/for_loops_over_fallibles.rs
+++ b/tests/ui/for_loops_over_fallibles.rs
@@ -21,12 +21,12 @@ fn for_loops_over_fallibles() {
     }
 
     // check over a `Result`
-    for x in result.into_iter() {
+    for x in result.iter_mut() {
         println!("{}", x);
     }
 
     // check over a `Result`
-    for x in result.iter_mut() {
+    for x in result.into_iter() {
         println!("{}", x);
     }
 
--- a/tests/ui/for_loops_over_fallibles.rs
+++ b/tests/ui/for_loops_over_fallibles.rs
@@ -2,7 +2,7 @@
 
 fn for_loops_over_fallibles() {
     let option = Some(1);
-    let result = option.ok_or("x not found");
+    let mut result = option.ok_or("x not found");
     let v = vec![0, 1, 2];
 
     // check over an `Option`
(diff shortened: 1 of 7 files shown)


tests/ui/for_loops_over_fallibles.rs
```diff
@@ -2,19 +2,34 @@
...
+    // check over a `Result`
+    for x in result.into_iter() {
+        println!("{}", x);
+    }
+
+    // check over a `Result`
+    for x in result.iter_mut() {
+        println!("{}", x);
+    }
+
```
llogiq, line 32: I would have switched the order of those. If `result` wasn't `Copy`, this would consume it before borrowing it mutably.

llogiq (commented):
Thank you for making clippy better! I have but a small nit, otherwise it looks merge-worthy.

llogiq:
Thanks! :+1:

@bors r+

PR #8952, opened 2022-06-05
add `--explain` subcommand

### LINKED ISSUE #8291: cargo clippy --explain <lintname>

llogiq:
This closes #8291.

---

changelog: add `cargo clippy -- --explain <lintname>` subcommand


llogiq:
I started out trying to get a proc macro to work, but then I found Alexey Kladow's blog post about self-modifying code and I thought this might be just the right application for the technique.

xFrednet:
This one looks interesting, and my reviewing queue is currently relatively small. I would take over the review. (If you want to keep it @dswij, feel free to say so)

r? @xFrednet 

@llogiq could you maybe link the blog post? Your comment reminded me of a post, I've read before. Also, as a side node, I sometimes find it confusing, when to use the double dashes after `cargo clippy` and when not. Therefore, I think it would be cool, if the `--explain` comment would also be accepted by `cargo-clippy` and then passed to `clippy-driver`. But that can also be done in a followup PR :upside_down_face:

llogiq:
Updated my previus comment with the link.

dswij:
Sure, go ahead!

llogiq:
@xFrednet I'm fine with merging this & doing a follow-up for cargo-clippy later.

llogiq:
I noted that -E is already used for "error".

tests/check-docs.rs
```diff
@@ -0,0 +1,60 @@
+use regex::Regex;
+use std::{collections::BTreeMap, fmt::Write, path::PathBuf};
+use walkdir::WalkDir;
+
+fn mangle_comments(s: &str) -> String {
+    let result = s
+        .split('\n')
+        .map(str::trim)
+        .filter_map(|line| line.strip_prefix("/// "))
```
xFrednet, line 9: This removes spacer lines and makes the output look a bit cramped. What do you think about:
```suggestion
        .filter_map(|line| line.strip_prefix("/// ").or_else(line.strip_prefix("///"))
```

Or something similar

tests/check-docs.rs
```diff
@@ -0,0 +1,60 @@
...
+        .filter(|line| !line.starts_with("# ")) // filter out code
+        .map(|line| if line.starts_with("```") { "" } else { line })
```
xFrednet, line 11: I think it would be cool to add indention to code blocks like this:

```
### Example

    let mut v = Vec::new();
    v.push(0);

Use instead:

    let v = vec![0];
```

xFrednet (commented):
The implementation itself looks good to be. I haven't done a full test, but the initial tries seem good.

While I like the concept of using a test, it feels a bit off in Clippy's codebase since we already do this kind of generation in `clippy_dev`. This introduces another location with little benefit in comparison to adding it to `clippy_dev`. At least that's my impression. `clippy_dev` doesn't have the regex crate, but that should be simple to avoid/rewrite.

cc: @Serial-ATA This might be interesting, since you're working on the metadata collection/lint list. Feel free to ignore this as well :upside_down_face:

Serial-ATA:
Couldn't `docs.rs` be generated during the metadata collection lint alongside `lints.json`? Just so you don't have to reinvent the wheel to get the docs.

src/docs.rs
```diff
@@ -0,0 +1,567 @@
+// autogenerated. Please look at /tests/check-docs.rs
+
+#[allow(clippy::too_many_lines)]
+#[rustfmt::skip]
+fn explain(lint: &str) {
+    println!("{}", match lint {
+        "absurd_extreme_comparisons" => "### What it does\nChecks for comparisons where one side of the relation is\neither the minimum or maximum value for its type and warns if it involves a\ncase that is always true or always false. Only integer and boolean types are\nchecked.\n### Why is this bad?\nAn expression like `min <= x` may misleadingly imply\nthat it is possible for `x` to be less than the minimum. Expressions like\n`max < x` are probably mistakes.\n### Known problems\nFor `usize` the size of the current compile target will\nbe assumed (e.g., 64 bits on 64 bit systems). This means code that uses such\na comparison to detect target pointer width will trigger this lint. One can\nuse `mem::sizeof` and compare its value or conditional compilation\nattributes\nlike `#[cfg(target_pointer_width = \"64\")] ..` instead.\n### Example\n\nlet vec: Vec<isize> = Vec::new();\nif vec.len() <= 0 {}\nif 100 > i32::MAX {}\n",
```
flip1995, line 7: Having this file in the repository will probably result in many merge conflicts in PRs that touch documentation in one way or another.

tests/check-docs.rs
```diff
@@ -0,0 +1,60 @@
...
+
+const SUFFIX: &str = r#"        _ => "unknown lint",
+    });
+}
+"#;
+
+static CODE_FILE: &str = "src/docs.rs";
+
+#[test]
+fn gen_docs() {
```
flip1995, line 32: Should this be done in `cargo dev update_lints` rather than in a test?

flip1995:
> Couldn't `docs.rs` be generated during the metadata collection lint alongside `lints.json`?

collect-metadata is only run during deployment and not even in bors CI. However, collect-metadata could use this file to generate the documentation instead of doing it itself.

KarlWithK:
This would be a really cool thing to have -- especially if we can somehow add this feature to rust-analyzer. Thank you for your work!

llogiq:
So, I see two ways to improve on this:

1. I'll move the docs creation to `cargo-dev` so it's included with `update_lints` and re-use our current infrastructure.
2. I'll change the generation to emit a file per lint and a single file to `include_str!` all of those doc files. This way we won't get too many merge conflicts from docs.

src/docs.rs
llogiq: Ok, I'm going to solve this by creating a lot of `src/docs/*.md`  files and using `include_str!()` on each. That way we only get conflicts if two lints are added/removed (and those are easy to fix) or two people work on the same lint docs (but we get that already).

llogiq:
This version should deal with the merge conflicts arising from doc changes.

I didn't get to changing the code to use `cargo dev update_lints` yet. Using the metadata collector would not be optimal, because we don't run it in CI. On the other hand, extending `update_lints` to also parse the doc comments would duplicate some machinery. Oh well.

llogiq:
Hmmm...I should not name the files ".md" when they don't contain markdown...

tests/check-docs.rs
llogiq: I chose to leave the `\`\`\`` for now. `rustc --explain` does the same, so there is some precedent, and for this first version it should be OK.

tests/check-docs.rs
llogiq: I moved this to `update_lints`.

src/main.rs
```diff
@@ -54,6 +57,16 @@ pub fn main() {
         return;
     }
 
+    if let Some(pos) = env::args().position(|a| a == "--explain") {
+        if let Some(mut lint) = env::args().nth(pos + 1) {
+            lint.make_ascii_lowercase();
+            explain(&lint.replace('-', "_"));
```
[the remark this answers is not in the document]
llogiq: Good point. Added in the most recent commit.

llogiq:
@dswij r?

xFrednet:
Sorry, for the late replay, I moved last week and life is currently still a bit chaotic. Reviewing this is on my todo list for this weekend. If @dswij want's to take over that would also be fine for me :upside_down_face:

llogiq:
@xFrednet @dswij ping?

src/main.rs
```diff
@@ -42,6 +43,8 @@ fn show_version() {
     println!("{}", version_info);
 }
 
+include!("docs.rs");
```
xFrednet, line 46: If possible I would prefer to use a module with `mod docs` as rust-analizer still is a bit iffy with files included differently AFAIK

src/main.rs
```diff
@@ -54,6 +57,16 @@ pub fn main() {
         return;
     }
 
+    if let Some(pos) = env::args().position(|a| a == "--explain") {
+        if let Some(mut lint) = env::args().nth(pos + 1) {
+            lint.make_ascii_lowercase();
+            explain(&lint.strip_prefix("clippy::").unwrap_or(&lint).replace('-', "_"));
+        } else {
+            show_help();
```
xFrednet, line 65: I feel like this should actually say that a lint name was expected, but that is very nit picky and the help message works well enough. :upside_down_face:

xFrednet (approved):
Overall, looks good to me. I also like the location in `cargo-clippy`. I've also played around with it and to works well :). My GitHub is actually struggling a bit with this many changes ^^.

Thank you for your patience, the last weeks have been very full for me :sweat_smile:

xFrednet:
Hey @rust-lang/clippy, this PR adds the `--explain` comment to `cargo-clippy`. For this, it stores the lint description in a text file inside `src/docs/<lint_name>.txt`. The documentation files are automatically checked and updated by `cargo dev update_lints`. This seems to work quite well and looks good to me. Does anyone have any concerns regarding this? :upside_down_face:

llogiq:
I need to rebase to make the tests work.

src/main.rs
llogiq: done.

src/docs.rs
```diff
@@ -0,0 +1,589 @@
+// autogenerated. Please look at /tests/check-docs.rs
```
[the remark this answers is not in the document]
llogiq: Ah, good catch! Fixed.

src/main.rs
llogiq: Like "error: expected lint name. Please use with `cargo clippy -- --explain <lint>`"?

xFrednet: Yeah, maybe, but I don't think this is a blocker. Having the help message is good  enough IMO :)

xFrednet (approved):

llogiq:
Thank you for the review, @xFrednet.

PR #8954, opened 2022-06-05
Improve lint doc consistency

Serial-ATA:
changelog: none

This is a continuation of #8908.

Notable changes:
- Removed empty `Known Problems` sections
- Removed "Good"/"Bad" language (replaced with "Use instead")
- Removed (and added some :smile:) duplication
- Ignored the [`create_dir`] example so it doesn't create `clippy_lints/foo` :smile:


clippy_lints/src/implicit_saturating_sub.rs
```diff
@@ -16,17 +16,15 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
-    /// let end: u32 = 10;
-    /// let start: u32 = 5;
-    ///
-    /// let mut i: u32 = end - start;
-    ///
-    /// // Bad
```
xFrednet, line 24: Based on the suggestion and lint description, I'd say, that the subtraction part is still important. Based on the previous example, we should also check if the unsigned type is important :upside_down_face:

clippy_lints/src/casts/mod.rs
```diff
@@ -277,23 +283,31 @@ declare_clippy_lint! {
...
     /// ```rust
-    /// // Bad: fn1 is cast as `usize`
+    /// // fn1 is cast as `usize`
     /// fn fn1() -> u16 {
     ///     1
     /// };
-    /// let _ = fn1 as usize;
+    /// # let _ =
+    /// fn1 as usize;
+    /// ```
```
xFrednet, line 292: While this works, I believe this will look weird, when the lint output is added to the lint list. Then the output would contain a `let _ =` which is not present in the example.

 ```suggestion
    /// # let _ = fn1 as usize;
    /// ```
```

File: clippy_lints/src/casts/mod.rs [hunk 3]
```diff
@@ -219,13 +219,16 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
-    /// // Bad
     /// fn fun() -> i32 { 1 }
-    /// let a = fun as i64;
+    /// # let _ =
+    /// fun as i64;
```
xFrednet, line 224: See comment below
```suggestion
    /// # let _ = fun as i64;
```

clippy_lints/src/indexing_slicing.rs
```diff
@@ -17,19 +17,20 @@ declare_clippy_lint! {
     /// ### Why is this bad?
     /// This will always panic at runtime.
     ///
-    /// ### Known problems
-    /// Hopefully none.
-    ///
     /// ### Example
-    /// ```no_run
+    /// ```rust,no_run
```
xFrednet, line 21: Just a note: `rust` is implicit in rustdoc and also added by the metadata collector for the lint list. I don't think it's wrong to add it, but any implementations should not rely on `rust` as a language indicator being present, IMO.

This can remain, there is no benefit in removing it :upside_down_face:

clippy_lints/src/indexing_slicing.rs
```diff
@@ -49,42 +50,34 @@ declare_clippy_lint! {
...
-    /// &y[..100];
+    /// ```
+    ///
+    /// Use instead:
+    /// ```rust
+    /// # let x = vec![0; 5];
+    /// # let y = [0, 1, 2, 3];
+    /// # let _ =
+    /// x.get(2);
+    /// # let _ =
```
xFrednet, line 74: Why are these actually needed, will we trigger a lint like `must_use` otherwise?

xFrednet (commented):
I'm done with the first batch ~33 files or so. :upside_down_face: Thank you for taking the time, this looks like you've actually read every description, which is impressive!

clippy_lints/src/indexing_slicing.rs
Serial-ATA: Yeah, I just add it when I see it missing. I like being explicit with these things :smile:.

clippy_lints/src/casts/mod.rs
Serial-ATA: Good call, I probably did that in other places. I'll get that fixed.

clippy_lints/src/implicit_saturating_sub.rs
Serial-ATA: How about this:

```rust
# let end: u32 = 10;
# let start: u32 = 5;
let mut i: u32 = end - start;

if i != 0 {
    i -= 1;
}
```

I just didn't think `end` and `start` were too important to understanding what the lint does.

xFrednet (commented):
Looks good overall, a few small comments. Everything should be simple to fix and then ready to be merged.

I hope I wasn't too nit-picky

clippy_lints/src/literal_representation.rs
```diff
@@ -22,11 +22,16 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
-    /// // Bad
-    /// let x: u64 = 61864918973511;
+    /// # let _: u64 =
+    /// 61864918973511
+    /// # ;
```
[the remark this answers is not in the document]
Serial-ATA: For these ones, I thought it looked better to just show the literals themselves. Since it applies in all contexts I didn't think the `let` statement was all that important for the reader to see.

clippy_lints/src/return_self_not_must_use.rs
```diff
@@ -26,37 +26,39 @@ declare_clippy_lint! {
...
     ///     pub fn bar(&self) -> Self {
     ///         Self
     ///     }
     /// }
-    /// ```
+    /// # }
     ///
-    /// Or on the type definition like this:
-    /// ```rust
+    /// # {
```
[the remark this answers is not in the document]
Serial-ATA: That was just to scope the struct definitions so the examples could be next to each other.

clippy_lints/src/needless_bool.rs
```diff
@@ -30,16 +30,22 @@ declare_clippy_lint! {
     /// shorter code.
     ///
     /// ### Example
-    /// ```rust,ignore
+    /// ```rust
+    /// # let x = true;
     /// if x {
     ///     false
+    /// # ;
```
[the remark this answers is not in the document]
Serial-ATA: The semis are needed:
<pre>
 --> src/lib.rs:7:5
  |
4 | / if x {
5 | |     false
6 | | } else {
7 | |     true
  | |     ^^^^ expected `()`, found `bool`
8 | | }
  | |_- expected this to be `()`
</pre>

The example could also just be ignored again.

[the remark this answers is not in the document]
xFrednet: Then I would add the semi at the ed of the if. This lint is for instances where the boolean could be used directly.

clippy_lints/src/return_self_not_must_use.rs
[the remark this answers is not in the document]
xFrednet: :+1:

clippy_lints/src/literal_representation.rs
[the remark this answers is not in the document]
xFrednet: Alright, I'm a bit torn, since I think it looks weird in the code, but the main focus should be the lint list documentation, where this will show correctly.

clippy_lints/src/needless_bool.rs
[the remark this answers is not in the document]
Serial-ATA: Oops, yeah :facepalm: wasn't thinking.

xFrednet (approved):

xFrednet:
Alright, this version looks good to me. Thank you for the help and quick answers.

@bors r+

Serial-ATA:
@xFrednet thanks for getting the review done so quickly, really appreciate it! :heart:

PR #8958, opened 2022-06-06
Lint simple expressions in `manual_filter_map`, `manual_find_map`

Alexendoo:
changelog: Lint simple expressions in [`manual_filter_map`], [`manual_find_map`]

The current comparison rules out `.find(|a| a.is_some()).map(|b| b.unwrap())` because `a` being a reference can effect more complicated expressions, this adds a simple check for that case and adds the necessary derefs

There's some overlap with `option_filter_map` so `lint_filter_some_map_unwrap` now returns a `bool` to indicate it linted

### DIFF
--- a/clippy_lints/src/methods/filter_map.rs
+++ b/clippy_lints/src/methods/filter_map.rs
@@ -8,6 +8,7 @@ use rustc_hir as hir;
 use rustc_hir::def::Res;
 use rustc_hir::{Expr, ExprKind, PatKind, PathSegment, QPath, UnOp};
 use rustc_lint::LateContext;
+use rustc_middle::ty::adjustment::Adjust;
 use rustc_span::source_map::Span;
 use rustc_span::symbol::{sym, Symbol};
 use std::borrow::Cow;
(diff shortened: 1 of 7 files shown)


clippy_lints/src/methods/filter_map.rs
```diff
@@ -58,15 +59,15 @@ fn lint_filter_some_map_unwrap(
     map_arg: &hir::Expr<'_>,
     target_span: Span,
     methods_span: Span,
-) {
+) -> bool {
```
giraffate, line 62: IMO it would be better to separate function into one that check and return `bool` and one that just emit the lint. I was confused that the function name `lint_filter_some_map_unwrap` return `bool`.

giraffate (commented):
Sorry for the late reviewing.

clippy_lints/src/methods/filter_map.rs
Alexendoo: Thanks, changed it to be an `is_` function

PR #7359, opened 2021-06-16
Clippy book

joshrotenberg:
A work in progress Clippy Book using mdbook. See #6011.

This is currently just a moving around of things:

1. The current README.md split up a bit and put into sections.
1. A rough outline of Clippy lint categories (currently no content, potentially add a basic introduction for each and some example, see questions below.
1. The `docs` content repurposed into a top level `Development` section.
1. The current Roadmap.

Some big questions:

1. is `guide/` the right place? I'm modeling after mdbook itself. 
1. What is the relationship between ALL the Clippy Lints and this guide? It seems like they can coexist. Does that mean the guide should just point to the current side with regard to actual lints, and maybe just include some examples to keep it interesting? Keeping both up to date seems like a maintenance nightmare unless its automated somehow. Or should the current ALL the Clippy lints somehow be incorporated into the book?
1. Related to the above, where should this guide be published since the `gh-pages` branch is already in use?
1. This PR doesn't currently change any existing content. Obviously that would make sense assuming the general structure and relocation is an acceptable approach.

---

Open Tasks for follow up PR:
- Set up CI/CD
- Split up Installation and Usage
- Add more content to Usage (and Installation) chapters
- Enhance CI chapter with more examples for different CIs

---

changelog: The first version of the *Clippy Book*
staring, *The Clippy Team*, *Bors* and *rustc*

PR #8959, opened 2022-06-06
Deploy Clippy book through CI/CD

flip1995:
Now that #7359 is finally merged, we can start looking at the deployment of the book. 

Marking this as draft for now, because I created this 2 months ago and I still have to think about if we want to actually do this if we end up publishing the book through the rust-lang/rust repo.

r? @xFrednet 

changelog: none


flip1995:
Closing in favor of #9005

PR #8964, opened 2022-06-07
Warn about read into zero-length `Vec`

### LINKED ISSUE #8886: Warn about read into zero-length `Vec`
### What it does

This lint catches reads into a zero-length `Vec`. For instance, it should flag code like this:

```rust
let mut data = Vec::with_capacity(len);
r.read_exact(&mut data)?;
```

In general, it should flag any code that constructs a `Vec` (whether using `new` or `vec![]` or especially `with_capacity`) and then reads into it (using `read` or `read_exact`, whether sync or async versions) without changing the size of the `Vec`. The lint should warn that the number of bytes read depends on the size of the buffer read into, so this pattern will read 0 bytes. And in the specific case of a call to `with_capacity`, the lint should warn that read gets the number of bytes from the `Vec`'s length, not its capacity.

In the `with_capacity` case, the lint should suggest adding something like `data.resize(len, 0);`; in other cases the lint can more generally suggest resizing before reading, or calling `read_to_end`.

### Lint Name

read_zero_byte_vec

### Category

correctness

### Advantage

Correctness: reading zero bytes is almost certainly not the intended behavior of this code. This lint would catch a bug.

I've seen multiple instances of this bug, and I just wrote one myself.

### Drawbacks

In theory, a very *unusual* read implementation could assign some semantic meaning to zero-byte reads. But it seems exceptionally unlikely that code *intending* to do a zero-byte read would allocate a `Vec` for it.

### Example

```rust
let mut data = Vec::with_capacity(len);
r.read_exact(&mut data)?;
```

Could be written as:

```rust
let mut data = Vec::with_capacity(len);
data.resize(len, 0);
r.read_exact(&mut data)?;
```

-----

```rust
let mut data = Vec::with_capacity(len);
r.read_exact(&mut data).await?;
```

Could be written as:

```rust
let mut data = Vec::with_capacity(len);
data.resize(len, 0);
r.read_exact(&mut data).await?;
```

tamaroning:
Closes #8886 

- \[x] Run `cargo dev fmt`

changelog: none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3678,6 +3678,7 @@ Released 2018-09-13
 [`rc_buffer`]: https://rust-lang.github.io/rust-clippy/master/index.html#rc_buffer
 [`rc_clone_in_vec_init`]: https://rust-lang.github.io/rust-clippy/master/index.html#rc_clone_in_vec_init
 [`rc_mutex`]: https://rust-lang.github.io/rust-clippy/master/index.html#rc_mutex
+[`read_zero_byte_vec`]: https://rust-lang.github.io/rust-clippy/master/index.html#read_zero_byte_vec
 [`recursive_format_impl`]: https://rust-lang.github.io/rust-clippy/master/index.html#recursive_format_impl
 [`redundant_allocation`]: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_allocation
 [`redundant_clone`]: https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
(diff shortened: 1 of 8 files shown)


dswij (approved):
Thanks for the PR! Changes looks great with the tests

Can you help to squash some commits? After that, I think it's good to merge

tamaroning:
@dswij 
Thank you for your review.
Squashed my commits :)

dswij:
Thanks for this!

@bors r+

unrealhoang:
@tamaroning 
I found a false positive case for this lint:
```rust
    let mut v = Vec::new();
    {
        v.resize(10, 0);
        r.read(&mut v).unwrap();
    }
```
will cause this lint to go off.

dswij:
@unrealhoang Seems like this is because the lint does not transverse down scopes. Can you help to create a new issue for this?

unrealhoang:
@dswij I created a new issue: https://github.com/rust-lang/rust-clippy/issues/9274

dswij:
> @dswij I created a new issue: #9274

@unrealhoang Thanks for this!

PR #8976, opened 2022-06-09
Fix some `#[expect]` lint interaction

xFrednet:
Fixing the first few lints that aren't caught by `#[expect]`. The root cause of these examples was, that the lint was emitted at the wrong location.

---

changelog: none

r? @Jarcho 

cc: rust-lang/rust#97660

### DIFF
--- a/clippy_lints/src/async_yields_async.rs
+++ b/clippy_lints/src/async_yields_async.rs
@@ -1,4 +1,4 @@
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::diagnostics::span_lint_hir_and_then;
 use clippy_utils::source::snippet;
 use clippy_utils::ty::implements_trait;
 use rustc_errors::Applicability;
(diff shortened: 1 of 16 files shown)


xFrednet:
Hey @Jarcho, could you review this. You can r+ this directly if you agree with the changes :upside_down_face: 

@bors delegate=Jarcho

Jarcho:
Taking a second look at `async_yields_async` I don't think anything needs to change. It's basically linting the signature of the body. I originally thought it might have an issue because it did something with the span of the final expression.

Jarcho:
Everything else LGTM.

xFrednet:
Well, while `async_yields_async` lints on the entire body, for the user it would be nice, if expect/allow also works on the statement that's shown in the lint message. Every problem you found with the expectation not working was also affecting other lint attributes, from what I can tell :upside_down_face:

So far, I haven't got any conflicts (Made these changes last weekend or so). Should we merge it as it is, or should I add the other lint changes to this PR as well? I'm fine either way :)

Jarcho:
I question whether that's a good span to lint on in general, but that can be a different discussion.

May as well merge now. Waiting just increases the chances of conflicts coming up. @bors r+

PR #8979, opened 2022-06-09
Correct some `#[clippy::version]`s

Alexendoo:
Some were a bit off

changelog: none

### DIFF
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -414,7 +414,7 @@ declare_clippy_lint! {
     /// enum E { X = 256 };
     /// let _ = E::X as u8;
     /// ```
-    #[clippy::version = "1.60.0"]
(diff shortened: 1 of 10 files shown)


flip1995:
How did you find those incorrect versions? Do you think that there could be a way to automatically check those? 

I trust you on the new version numbers :+1: 

@bors r+

xFrednet:
I think this issue would be solved by having nightly lint's and then set the version on a sync, when we enable/release new lints.

I still have the nightly lint Implementation in my backlog, but I'm held up on some implementation details. Until then it would be cool, to detect this somehow 👍

flip1995:
The way I would go about it is checking out the gh-pages branch and write a script tracking when a new lint impl was added and comparing it with the version in the documentation.

But yeah, once we have nightly lints, and a stabilization procedure for those in place, automating it will be way easier.

PR #8981, opened 2022-06-10
Add details about how significant drop in match scrutinees can cause deadlocks

PrestonFrom:
Adds more details about how a significant drop in a match scrutinee can cause a deadlock and include link to documentation.

changelog: Add more details to significant drop lint to explicitly show how temporaries in match scrutinees can cause deadlocks.

### DIFF
--- a/clippy_lints/src/matches/significant_drop_in_scrutinee.rs
+++ b/clippy_lints/src/matches/significant_drop_in_scrutinee.rs
@@ -4,7 +4,7 @@ use clippy_utils::get_attr;
 use clippy_utils::source::{indent_of, snippet};
 use rustc_errors::{Applicability, Diagnostic};
 use rustc_hir::intravisit::{walk_expr, Visitor};
-use rustc_hir::{Expr, ExprKind, MatchSource};
+use rustc_hir::{Arm, Expr, ExprKind, MatchSource};
 use rustc_lint::{LateContext, LintContext};
 use rustc_middle::ty::subst::GenericArgKind;
 use rustc_middle::ty::{Ty, TypeAndMut};
@@ -16,12 +16,21 @@ pub(super) fn check<'tcx>(
     cx: &LateContext<'tcx>,
     expr: &'tcx Expr<'tcx>,
     scrutinee: &'tcx Expr<'_>,
+    arms: &'tcx [Arm<'_>],
     source: MatchSource,
 ) {
     if let Some((suggestions, message)) = has_significant_drop_in_scrutinee(cx, scrutinee, source) {
         for found in suggestions {
             span_lint_and_then(cx, SIGNIFICANT_DROP_IN_SCRUTINEE, found.found_span, message, |diag| {
                 set_diagnostic(diag, cx, expr, found);
+                let s = Span::new(expr.span.hi(), expr.span.hi(), expr.span.ctxt(), None);
+                diag.span_label(s, "original temporary lives until here");
+                if let Some(spans) = has_significant_drop_in_arms(cx, arms) {
+                    for span in spans {
+                        diag.span_label(span, "significant drop in arm here");
+                    }
+                    diag.note("this might lead to deadlocks or other unexpected behavior");
+                }
             });
         }
     }
@@ -80,22 +89,77 @@ fn has_significant_drop_in_scrutinee<'tcx, 'a>(
     let mut helper = SigDropHelper::new(cx);
     helper.find_sig_drop(scrutinee).map(|drops| {
         let message = if source == MatchSource::Normal {
-            "temporary with significant drop in match scrutinee"
+            "temporary with drop impl with side effects in match scrutinee lives to end of block"
         } else {
-            "temporary with significant drop in for loop"
+            "temporary with drop impl with side effects in for loop condition lives to end of block"
         };
         (drops, message)
     })
 }
 
+struct SigDropChecker<'a, 'tcx> {
+    seen_types: FxHashSet<Ty<'tcx>>,
+    cx: &'a LateContext<'tcx>,
+}
+
+impl<'a, 'tcx> SigDropChecker<'a, 'tcx> {
+    fn new(cx: &'a LateContext<'tcx>) -> SigDropChecker<'a, 'tcx> {
+        SigDropChecker {
+            seen_types: FxHashSet::default(),
+            cx,
+        }
+    }
+
+    fn get_type(&self, ex: &'tcx Expr<'_>) -> Ty<'tcx> {
+        self.cx.typeck_results().expr_ty(ex)
+    }
+
+    fn has_seen_type(&mut self, ty: Ty<'tcx>) -> bool {
+        !self.seen_types.insert(ty)
+    }
+
+    fn has_sig_drop_attr(&mut self, cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
+        if let Some(adt) = ty.ty_adt_def() {
+            if get_attr(cx.sess(), cx.tcx.get_attrs_unchecked(adt.did()), "has_significant_drop").count() > 0 {
+                return true;
+            }
+        }
+
+        match ty.kind() {
+            rustc_middle::ty::Adt(a, b) => {
+                for f in a.all_fields() {
+                    let ty = f.ty(cx.tcx, b);
+                    if !self.has_seen_type(ty) && self.has_sig_drop_attr(cx, ty) {
+                        return true;
+                    }
+                }
+
+                for generic_arg in b.iter() {
+                    if let GenericArgKind::Type(ty) = generic_arg.unpack() {
+                        if self.has_sig_drop_attr(cx, ty) {
+                            return true;
+                        }
+                    }
+                }
+                false
+            },
+            rustc_middle::ty::Array(ty, _)
+            | rustc_middle::ty::RawPtr(TypeAndMut { ty, .. })
+            | rustc_middle::ty::Ref(_, ty, _)
+            | rustc_middle::ty::Slice(ty) => self.has_sig_drop_attr(cx, *ty),
+            _ => false,
+        }
+    }
+}
+
 struct SigDropHelper<'a, 'tcx> {
     cx: &'a LateContext<'tcx>,
     is_chain_end: bool,
-    seen_types: FxHashSet<Ty<'tcx>>,
     has_significant_drop: bool,
     current_sig_drop: Option<FoundSigDrop>,
     sig_drop_spans: Option<Vec<FoundSigDrop>>,
     special_handling_for_binary_op: bool,
+    sig_drop_checker: SigDropChecker<'a, 'tcx>,
 }
 
 #[expect(clippy::enum_variant_names)]
@@ -118,11 +182,11 @@ impl<'a, 'tcx> SigDropHelper<'a, 'tcx> {
         SigDropHelper {
             cx,
             is_chain_end: true,
-            seen_types: FxHashSet::default(),
             has_significant_drop: false,
             current_sig_drop: None,
             sig_drop_spans: None,
             special_handling_for_binary_op: false,
+            sig_drop_checker: SigDropChecker::new(cx),
         }
     }
 
@@ -163,7 +227,7 @@ impl<'a, 'tcx> SigDropHelper<'a, 'tcx> {
         if self.current_sig_drop.is_some() {
             return;
         }
-        let ty = self.get_type(expr);
+        let ty = self.sig_drop_checker.get_type(expr);
         if ty.is_ref() {
             // We checked that the type was ref, so builtin_deref will return Some TypeAndMut,
             // but let's avoid any chance of an ICE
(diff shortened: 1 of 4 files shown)


PrestonFrom:
@oli-obk Apologies for the delay, but I have added more details about the lint, linked to documentation generated as part of clippy, and tried to make the lifetime of the temporary clearer.

I believe you mentioned including wg-diagnostics on the code review. Is that a group I can @ here or do I need to send a link to the PR somewhere?

Thank you!

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
```diff
@@ -22,6 +22,13 @@ pub(super) fn check<'tcx>(
         for found in suggestions {
             span_lint_and_then(cx, SIGNIFICANT_DROP_IN_SCRUTINEE, found.found_span, message, |diag| {
                 set_diagnostic(diag, cx, expr, found);
+                diag.span_label(
+                    expr.span,
+                    "temporaries in a match scrutinee are not dropped until the end of the \
+                    match, so side effects from dropping such temporaries are delayed. \
+                    Deadlocks can occur if a lock guard is created in the scrutinee and a new \
+                    lock guard for the same mutex is created in a match arm. \
+                    For more information, see https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_in_scrutinee");
```
flip1995, line 31: This link is automatically added to every clippy lint, unless explicitly disabled through an undocumented env var. So no need to add it here.

I don't think adding such a big paragraph to a lint message is what was suggested. We don't add full lint explanations to lint messages, that's what the lint documentation is for. I can't think of a rustc or Clippy lint where something like this is done. Can you link the discussion where adding this was suggested?

flip1995:
> I believe you mentioned including wg-diagnostics on the code review. Is that a group I can @ here or do I need to send a link to the PR somewhere?

You only have to include wg-diag once this lint gets on the way to be uplifted. Until then all reviews are done by the Clippy team.

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
PrestonFrom: @flip1995 Thank you for the review! It looks like I might have misunderstood Oli's suggestion here: https://github.com/rust-lang/rust/issues/93883#issuecomment-1122007881

Particularly this part:
```
the diagnostic messaging needs to explain why this is a problem, ideally with some links to documentation and examples

    maybe even add a short span pointing to the end of the match stating that the lock holds until there, so if any arms directly or via function calls lock again, you're in a deadlock
```

I've tried to make the messages smaller, but I did notice some PRs opened due to this lint where it was mentioned that the current output was unhelpful and confusing. If you think this is still too much, I will keep trying to trim it down.

PrestonFrom: > This link is automatically added to every clippy lint, unless explicitly disabled through an undocumented env var. So no need to add it here.

Oh, I completely missed this! Thank you! The URL has been removed as well.

flip1995: Ah yeah, we might want to improve the messaging a bit. But just writing a big paragraph won't help here. I think the ideal thing to do would be something like:

```console
error: temporary with drop impl with side effects in match scrutinee lives to end of block
  --> $DIR/significant_drop_in_scrutinee.rs:59:11
   |
LL |     match mutex.lock().unwrap().foo() {
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
LL |         42 => mutex.lock().unwrap().bar()
    |              ^^^^^^^^^^^^ note: new side effect temporary created here
...
LL |     };
   |      - original temporary lives until here
   |
   = note: this might lead to deadlocks or other unexpected behavior
help: try moving the temporary above the match
   |
LL ~     let value = mutex.lock().unwrap().foo();
LL ~     match value {
   |
```

But this would entail, that we also try to find other significant drop temporaries inside the match block, which might be a bit hard.

tests/ui/significant_drop_in_scrutinee.stderr
```diff
@@ -11,163 +14,205 @@ LL ~     let value = mutex.lock().unwrap().foo();
...
 
-error: temporary with significant drop in match scrutinee
+error: temporary with drop impl with side effects in match scrutinee lives to end of block
   --> $DIR/significant_drop_in_scrutinee.rs:201:11
    |
 LL |     match counter.temp_increment().len() {
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+...
+LL |     };
+   |      - temporary lives until here. If mutex is locked within block, deadlocks can occur.
```
Alexendoo, line 54: The message could be confusing when there's no mutexes involved. It would be good to be able to say

> if `counter` is a mutex that is locked within the block, deadlocks can occur

Or something to those effects, if getting the name is tricky

flip1995: Agreed. See my above suggestion https://github.com/rust-lang/rust-clippy/pull/8981#discussion_r897665627 Don't mention mutexes at all, but say what the result might be. Which is "deadlocks or other unexpected behavior".

PrestonFrom:
@flip1995 @Alexendoo 
Thank you both for the comments! This took me a bit longer than anticipated since I started trying to use MIR to look for the "origin" of temporaries with significant drops, but I wasn't finding a good solution.

The new commit will look for any type in the arms that has a significant drop -- this is not 100% correct, since it's possible that a different "parent" is creating the new temporary. All the types found in the arms are indicated with notes and a final note is added with the message `this might lead to deadlocks or other unexpected behavior`.
If no other types with significant drops are found in the arms, none of the new messages are emitted. I'm not sure if that's correct though, since the "parent" could be passed to a function that creates a new temporary as well. (E.g., if the match locks a mutex and then that mutex is passed to a function, it could be locked again in that function.)
I think I need to find the "parent" of the temporary in the scrutinee and look for it creating new temporaries *or* being passed to other functions? Is that possible with MIR? If so, should that be done in this PR or a new PR?

PrestonFrom:
I also had some issues after trying to rebase and ended up force pushing an entirely new squashed commit. I hope I haven't made re-reviewing too annoying for you.

tests/ui/significant_drop_in_scrutinee.stderr
```diff
@@ -1,173 +1,288 @@
...
+   |
+   = note: this might lead to deadlocks or other unexpected behavior
 help: try moving the temporary above the match
    |
 LL ~         let value = mutex1.lock().unwrap().s.len();
 LL ~         match (value, true, mutex2.lock().unwrap().s.len()) {
    |
 
-error: temporary with significant drop in match scrutinee
+error: temporary with drop impl with side effects in match scrutinee lives to end of block
```
flip1995, line 133:
```suggestion
error: temporary with significant `Drop` in `match` scrutinee will live until the end of the `match` expression
```

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
```diff
@@ -16,12 +16,21 @@ pub(super) fn check<'tcx>(
...
         for found in suggestions {
             span_lint_and_then(cx, SIGNIFICANT_DROP_IN_SCRUTINEE, found.found_span, message, |diag| {
                 set_diagnostic(diag, cx, expr, found);
+                let s = Span::new(expr.span.hi(), expr.span.hi(), expr.span.ctxt(), None);
+                diag.span_label(s, "original temporary lives until here");
+                if let Some(spans) = has_significant_drop_in_arms(cx, arms) {
+                    for span in spans {
+                        diag.span_label(span, "significant drop in arm here");
+                    }
+                    diag.note("this might lead to deadlocks or other unexpected behavior");
```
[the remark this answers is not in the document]
PrestonFrom: Done!

tests/ui/significant_drop_in_scrutinee.stderr
PrestonFrom: Done!

tests/ui/significant_drop_in_scrutinee.stderr
```diff
@@ -1,173 +1,288 @@
...
   --> $DIR/significant_drop_in_scrutinee.rs:243:54
    |
 LL |         match (mutex1.lock().unwrap().s.len(), true, mutex2.lock().unwrap().s.len()) {
    |                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-   |
+...
+LL |                 mutex1.lock().unwrap().s.len();
+   |                 ---------------------- significant drop in arm here
```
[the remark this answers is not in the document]
PrestonFrom: Oh, good call, because this could be in a for-loop as well

PrestonFrom:
@flip1995 Thank you for the comments!

My initial assumption was that checking all the arms wouldn't add a lot of overhead because it would only happen when a significant drop was found in the match scrutinee/for-loop condition, but having evidence one way or the other would be great.

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
```diff
@@ -16,12 +16,21 @@ pub(super) fn check<'tcx>(
...
 ) {
     if let Some((suggestions, message)) = has_significant_drop_in_scrutinee(cx, scrutinee, source) {
         for found in suggestions {
             span_lint_and_then(cx, SIGNIFICANT_DROP_IN_SCRUTINEE, found.found_span, message, |diag| {
                 set_diagnostic(diag, cx, expr, found);
+                let s = Span::new(expr.span.hi(), expr.span.hi(), expr.span.ctxt(), None);
+                diag.span_label(s, "original temporary lives until here");
+                if let Some(spans) = has_significant_drop_in_arms(cx, arms) {
+                    for span in spans {
+                        diag.span_label(span, "another temporary with significant `Drop` created here");
```
Alexendoo, line 30: If I'm reading the code right this highlights any expr, not just temporaries

```suggestion
                        diag.span_label(span, "another value with significant `Drop` created here");
```

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
```diff
@@ -16,12 +16,21 @@ pub(super) fn check<'tcx>(
...
     scrutinee: &'tcx Expr<'_>,
+    arms: &'tcx [Arm<'_>],
     source: MatchSource,
 ) {
     if let Some((suggestions, message)) = has_significant_drop_in_scrutinee(cx, scrutinee, source) {
         for found in suggestions {
             span_lint_and_then(cx, SIGNIFICANT_DROP_IN_SCRUTINEE, found.found_span, message, |diag| {
                 set_diagnostic(diag, cx, expr, found);
+                let s = Span::new(expr.span.hi(), expr.span.hi(), expr.span.ctxt(), None);
+                diag.span_label(s, "original temporary lives until here");
```
Alexendoo, line 27: Then if so, given that we're only referring to temporaries in the scrutinee we could change

```suggestion
                diag.span_label(s, "temporary lives until here");
```

since `original` can be a little confusing when there's no other spans highlighted

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
```diff
@@ -330,3 +357,40 @@ impl<'a, 'tcx> Visitor<'tcx> for SigDropHelper<'a, 'tcx> {
         }
     }
 }
+
+struct ArmSigDropHelper<'a, 'tcx> {
+    sig_drop_checker: SigDropChecker<'a, 'tcx>,
+    found_sig_drop_spans: Option<FxHashSet<Span>>,
```
Alexendoo, line 363: minor, but you could make this field non optional

Alexendoo (commented):
Since it's only run when a candidate is found as you say, the extra visitor should be fine performance wise

When `#[allow(clippy::significant_drop_in_scrutinee)]`' is used it seems the `and_then` callback isn't triggered which would mean no problem there either. It is called for `#[expect]` however (cc @xFrednet in case that's a bug)

If you wanted to err on the side of caution you could throw an [`is_lint_allowed`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_lint_allowed.html) at the top of `check`, that would allow skipping the initial significant drop checking too

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
PrestonFrom: Done!

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
PrestonFrom: Good call!

clippy_lints/src/matches/significant_drop_in_scrutinee.rs
PrestonFrom: Excellent point!

flip1995 (approved):
Thanks! This all LGTM now. I still want to benchmark it though.

Alexendoo (approved):

PrestonFrom:
Thanks everyone for your help!
@flip1995 What needs to be done for benchmarking? Is that something I need to run?

flip1995:
Nvm, I just did that myself. I didn't see any significant changes.

Thanks for continuing improving this already awesome lint. Let's get this merged so people can test it in nightly on Friday

@bors r+

PrestonFrom:
@flip1995 My apologies for the delay in responding! I just saw this. Thank you for running the benchmarks and for helping me get this far!

flip1995:
> My apologies for the delay in responding! I just saw this.

No need to apologies. My two messages were in the span of 15 minutes :D

Issue #2387, opened 2018-01-21
Suggest into_owned instead of to_string for Cow<str>


clarcharr:
Note serde-rs/json#405.

We could also lint `to_vec` on `Cow<[T]>` and probably others.

H2CO3:
Note that this only applies when the `Cow` being inspected is owned by the caller. E.g. this lint would suggest code that doesn't compile when the caller has a `&Cow`. This shouldn't happen too often anyway, because the whole point of a `Cow` is that the user doesn't need to keep in his/her head whether it's a reference or an owned value. Actually, while we're here, may I also suggest a lint against passing `Cow` by reference?

Veetaha:
See additional discussions in the duplicate issue https://github.com/rust-lang/rust-clippy/issues/5987

Kixunil:
Note that the claim that `String: From<Cow<'_, str>>` allocates is [no longer true](https://doc.rust-lang.org/src/alloc/string.rs.html#2335-2339)

H2CO3:
@Kixunil it is not `String: From<Cow<str>>` that is of concern here; instead, it is only [`Cow: ToString`](https://doc.rust-lang.org/src/alloc/string.rs.html#2228) which has no choice but to allocate.

Kixunil:
Ah, I misread that code in the serde issue.

clarfonthey:
Basically, this is a specific case of the more general "did you mean to move instead of clone this" set of errors.

A4-Tacks:
Like #5987, temporary values can be implemented first, which is not controversial

PR #8984, opened 2022-06-10
Implemented `suspicious_to_owned` lint to check if `to_owned` is called on a `Cow`

xanathar:
changelog: Add lint ``[`suspicious_to_owned`]``

-----------------
Hi,
posting this unsolicited PR as I've been burned by this issue :)
Being unsolicited, feel free to reject it or reassign a different lint level etc.

This lint checks whether `to_owned` is called on `Cow<'_, _>`. This is done because `to_owned` is very similarly named to `into_owned`, but the effect of calling those two methods is completely different (one makes the `Cow::Borrowed` into a `Cow::Owned`, the other just clones the `Cow`). If the cow is then passed to code for which the type is not checked (e.g. generics, closures, etc.) it might slip through and if the cow data is coming from an unsafe context there is the potential for accidentally cause undefined behavior.
Even if not falling into this painful case, there's really no reason to call `to_owned` on a `Cow` other than confusing people reading the code: either `into_owned` or `clone` should be called.

Note that this overlaps perfectly with `implicit_clone` as a warning, but `implicit_clone` is classified pedantic (while the consequences for `Cow` might be of a wider blast radius than just pedantry); given the overlap, I set-up the lint so that if `suspicious_to_owned` triggers `implicit_clone` will not trigger. I'm not 100% sure this is done in the correct way (I tried to copy what other lints were doing) so please provide feedback on it if it isn't.

### Checklist

- \[x] Run `cargo dev fmt`

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -78,6 +78,7 @@ mod str_splitn;
 mod string_extend_chars;
 mod suspicious_map;
 mod suspicious_splitn;
+mod suspicious_to_owned;
 mod uninit_assumed_init;
 mod unit_hash;
 mod unnecessary_filter_map;
@@ -2055,6 +2056,55 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
+    /// Checks for the usage of `_.to_owned()`, on a `Cow<'_, _>`.
+    ///
+    /// ### Why is this bad?
+    /// Calling `to_owned()` on a `Cow` creates a clone of the `Cow`
+    /// itself, without taking ownership of the `Cow` contents (i.e.
+    /// it's equivalent to calling `Cow::clone`).
+    /// The similarly named `into_owned` method, on the other hand,
+    /// clones the `Cow` contents, effectively turning any `Cow::Borrowed`
+    /// into a `Cow::Owned`.
+    ///
+    /// Given the potential ambiguity, consider replacing `to_owned`
+    /// with `clone` for better readability or, if getting a `Cow::Owned`
+    /// was the original intent, using `into_owned` instead.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # use std::borrow::Cow;
+    /// let s = "Hello world!";
+    /// let cow = Cow::Borrowed(s);
+    ///
+    /// let data = cow.to_owned();
+    /// assert!(matches!(data, Cow::Borrowed(_)))
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # use std::borrow::Cow;
+    /// let s = "Hello world!";
+    /// let cow = Cow::Borrowed(s);
+    ///
+    /// let data = cow.clone();
+    /// assert!(matches!(data, Cow::Borrowed(_)))
+    /// ```
+    /// or
+    /// ```rust
+    /// # use std::borrow::Cow;
+    /// let s = "Hello world!";
+    /// let cow = Cow::Borrowed(s);
+    ///
+    /// let data = cow.into_owned();
+    /// assert!(matches!(data, String))
+    /// ```
+    #[clippy::version = "1.65.0"]
+    pub SUSPICIOUS_TO_OWNED,
+    suspicious,
+    "calls to `to_owned` on a `Cow<'_, _>` might not do what they are expected"
+}
+
+declare_clippy_lint! {
+    /// ### What it does
     /// Checks for calls to [`splitn`]
     /// (https://doc.rust-lang.org/std/primitive.str.html#method.splitn) and
     /// related functions with either zero or one splits.
@@ -3075,6 +3125,7 @@ impl_lint_pass!(Methods => [
     FROM_ITER_INSTEAD_OF_COLLECT,
     INSPECT_FOR_EACH,
     IMPLICIT_CLONE,
+    SUSPICIOUS_TO_OWNED,
     SUSPICIOUS_SPLITN,
     MANUAL_STR_REPEAT,
     EXTEND_WITH_DRAIN,
@@ -3553,7 +3604,12 @@ impl Methods {
                     }
                     unnecessary_lazy_eval::check(cx, expr, recv, arg, "then_some");
                 },
-                ("to_os_string" | "to_owned" | "to_path_buf" | "to_vec", []) => {
+                ("to_owned", []) => {
+                    if !suspicious_to_owned::check(cx, expr, recv) {
+                        implicit_clone::check(cx, name, expr, recv);
+                    }
+                },
+                ("to_os_string" | "to_path_buf" | "to_vec", []) => {
                     implicit_clone::check(cx, name, expr, recv);
                 },
                 ("unwrap", []) => {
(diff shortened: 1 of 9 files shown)


xanathar:
Converted to draft to fix the test issue (that for some reason doesn't manifest locally :( )

xanathar:
CI is happy now, marking ready for review.

mikerite:
This looks OK but shouldn't [redundant_clone](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone) warn about this? See also #2387.

mikerite:
Here are the warnings from running this on something like the 1000 top crates.

```
---> quick-xml-0.23.0/src/events/mod.rs:126:21                        
warning: this call to `to_owned` does not cause the std::borrow::Cow<[u8]> contents to become owned
   --> src/events/mod.rs:126:21
    |
126 |         Self::owned(self.buf.to_owned(), self.name_len)
    |                     ^^^^^^^^^^^^^^^^^^^ help: consider using: `self.buf.into_owned()`
    |
    = note: requested on the command line with `-W clippy::suspicious-to-owned`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_to_owned
                                                                      
---> sentry-0.26.0/src/transports/reqwest.rs:58:26                    
warning: this call to `to_owned` does not cause the std::borrow::Cow<str> contents to become owned
  --> src/transports/reqwest.rs:58:26
   |
58 |         let user_agent = options.user_agent.to_owned();
   |                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `options.user_agent.into_owned()`
   |
   = note: requested on the command line with `-W clippy::suspicious-to-owned`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_to_owned

---> toml-0.5.9/src/de.rs:457:34                                      
warning: this call to `to_owned` does not cause the std::borrow::Cow<str> contents to become owned
   --> src/de.rs:457:34
    |
457 |                         .map(|k| k.1.to_owned())
    |                                  ^^^^^^^^^^^^^^ help: consider using: `k.1.into_owned()`
    |
    = note: requested on the command line with `-W clippy::suspicious-to-owned`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_to_owned
                                                                      
---> toml-0.5.9/src/de.rs:689:20                                      
warning: this call to `to_owned` does not cause the std::borrow::Cow<str> contents to become owned
   --> src/de.rs:689:20
    |
689 |         let name = table.header[table.header.len() - 1].1.to_owned();
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `table.header[table.header.len() - 1].1.into_owned()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_to_owned
                                                                      
---> vcpkg-0.2.15/src/lib.rs:314:21                                   
warning: this call to `to_owned` does not cause the std::borrow::Cow<str> contents to become owned
   --> src/lib.rs:314:21
    |
314 |                     vcpkg_user_targets_path.to_string_lossy().to_owned()
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider using: `vcpkg_user_targets_path.to_string_lossy().into_owned()`
    |
    = note: requested on the command line with `-W clippy::suspicious-to-owned`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_to_owned
```

It looks like that for most of them the advice of using `to_owned()` won't work and the intention is to clone. I haven't looked at them too closely.

xanathar:
Thanks, running this on the top 1000 crates makes sense.

Quick-xml is certainly conscious of the caveats and in some way perpetrates the same `to_owned` vs `into_owned` argument (see [comment in code](https://github.com/tafia/quick-xml/blob/master/src/events/mod.rs#L124) ).

Vpckg use of `to_owned` is totally redundant (code compiles even after removing it) and notably does not seem to raise neither redundant_clone nor implicit_clone (I should really double check though, I admit I checked in a hurry -- anyway it's not the point).

That said, I can see how these usages (especially the choice made in quick-xml, even if I honestly disagree with it) make a strong argument against this lint.

The reason against relying on `implicit_clone` is because it's only at the pedantic level, plus it really is aesthetic only.

The issue I found was when using `Cow`s coming from unsafe contexts like:

```
#[no_mangle]
pub extern "C" fn delayed_print(cstr: *const c_char) {
    let mystr = unsafe { CStr::from_ptr(cstr) }.to_string_lossy().to_owned();

    tokio::task::spawn(async move {
        println!("{}", mystr);
    });
}
```

The above is UB and very prone to all kinds of crash with `to_owned`, while it's safe and correct with `into_owned`. Now, the core of the issue of the above code could be the programmer not jumping through hoops to get a proper lifetime over that `CStr`, but unless one has been burned before the above code seems sound (it's not!) and can easily slip through a code review.

This the reason why I was suggesting it for a more specific lint with a higher warning level, but I can see the reasons against this argument so feel free to close the PR in that case.

Jarcho:
Should the suggestion not be then to suggest using `clone`? I agree that `to_owned` on `Cow` is very misleading while `clone` would be much clearer in intent. A second suggestion to use `into_owned` could also be made using `Applicability::MaybeIncorrect `

clippy_lints/src/methods/mod.rs
```diff
@@ -1961,6 +1962,56 @@ declare_clippy_lint! {
...
+    /// The `into_owned()` method on a `Cow` can be used to clone the
+    /// underlying data and become a `Cow::Owned`. Calling `to_owned()`,
+    /// on the other hand, clones the `Cow` itself, rather than the
+    /// contained data.
+    /// Given the similar naming, the use of `to_owned()` is suspicious
+    /// and can potentially lead to undefined behavior if it's
+    /// inadvertently used on data coming from unsafe contexts.
+    /// Consider replacing `to_owned` with `into_owned` if the desired
+    /// result is getting a `Cow::Owned`, or explitly calling `clone()`
+    /// if the desired result is cloning the `Cow`.
```
llogiq, line 1980: I think this still places too much emphasis on `into_owned`, given that in our checks the suggestion would lead to errors. Perhaps describing the behavior of `to_owned` and why it may be surprising, and then giving both alternatives (`clone` vs. `into_owned`) would help.

llogiq (approved):
Sorry, I was absent for a while. I think the lint docs could use some minor rewording, otherwise I think we can merge this after a rebase.

xanathar:
I rebased and changed the wording so that it's less opinionated between `clone()` and `into_owned` and now it suggests both replacement "depending on intent".

Let me know if it's fine or other changes to the wording are required.

llogiq:
Thank you! r=me with another rebase.

xanathar:
Rebased on the latest master

llogiq:
Thank you!

@bors r+

xanathar:
Not sure about the merge failure here.
I tried to do a rebase on the latest master and it went without conflicts; cargo tests also not failing after it, `update_lints` did not change any files, and neither did `cargo dev fmt`. 

I'd be happy to help this landed but... I need a clue :)

Jarcho:
The `changelog:` part needs to be on one line.

@bors retry

morrisonlevi:
Hello, I don't understand the premise. Can you help me understand? If I am currently borrowing a `Cow<str>`, for instance, and I need to copy this to pass it to something else, why wouldn't `.to_owned()` be called? A clone is insufficient if lifetimes don't match like when the string comes from an FFI source, but since I don't own the Cow, I can't `.into_owned()` it either, right?

Jarcho:
`to_owned` and `clone` on `&Cow<'a, T>` do the same thing. They both create a new `Cow<'a, T>` with the same state as the old one.

PR #8985, opened 2022-06-10
Lint `[single_match]` on `Option` matches

### LINKED ISSUE #8928: `single_match` should lint exhaustive `match` on standard types
### Summary

After #8282 was closed via #8322 (in Rust 1.60) the example below is no longer linting.

The original rationale was that Clippy should not lint `match`es on user-defined enums which explicitly list all possible options. Otherwise extending the enum later would break the original code, but not the fixed one, which may be unexpected for the author.

However, there is little reason to believe that standard types like `std::option::Option` will be extended in the future. I suggest that they're exempt from this check. Another example is `std::result::Result`. Of course, that does not help with user-defined `Result`-like types like in the `anyhow` crate, but still.

### Lint Name

`single_match`

### Reproducer

I tried this code:

```rust
fn main() {
    let x = Some(10);
    match x {
        Some(_) => println!("hi!"),
        None => {},
    };
}
```

Like in Rust 1.58:

```
warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
 --> a.rs:3:5
  |
3 | /     match x {
4 | |         Some(x) => println!("hi! {}", x),
5 | |         None => {},
6 | |     };
  | |_____^ help: try this: `if let Some(x) = x { println!("hi! {}", x) }`
  |
  = note: `#[warn(clippy::single_match)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_match

warning: 1 warning emitted
```

Instead, this happened:

No output

### Version

```text
Clippy 0.1.63 (2022-05-31 e094492)
```

botahamec:
fixes #8928 

changelog: did some cleanup of the logic for ``[`single_match`]`` and ``[`single_match_else`]`` which fixes the bug where `Option` matches were not linted unless a wildcard was used for one of the arms.

### DIFF
--- a/clippy_lints/src/matches/single_match.rs
+++ b/clippy_lints/src/matches/single_match.rs
@@ -242,7 +242,10 @@ fn form_exhaustive_matches<'a>(cx: &LateContext<'a>, ty: Ty<'a>, left: &Pat<'_>,
             }
             true
         },
-        (PatKind::TupleStruct(..), PatKind::Path(_) | PatKind::TupleStruct(..)) => in_candidate_enum(cx, ty),
+        (PatKind::TupleStruct(..), PatKind::Path(_)) => in_candidate_enum(cx, ty),
+        (PatKind::TupleStruct(..), PatKind::TupleStruct(_, inner, _)) => {
+            in_candidate_enum(cx, ty) && inner.iter().all(contains_only_wilds)
+        },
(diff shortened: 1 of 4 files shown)


llogiq:
Thank you! @bors r+

PR #8988, opened 2022-06-12
Test metadata collection in Bors CI workflow

xFrednet:
This PR adds a new check to bors CI workflows, which ensures that the metadata collection success, when it's run as part of the `deploy` script. I've only added it to bors workflows, as the runtime will be high while it'll also succeed most of the time. This is a preparation for rust-lang/rust-clippy#8947.

---

changelog: none

r? @ghost

### DIFF
--- a/.github/workflows/clippy_bors.yml
+++ b/.github/workflows/clippy_bors.yml
@@ -143,6 +143,25 @@ jobs:
       env:
         OS: ${{ runner.os }}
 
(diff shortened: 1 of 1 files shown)


flip1995 (approved):
This doesn't increase the net CI time and we test a bit more. And we can track the perf of collect-metadata.

I don't see any downsides of adding this :+1:

PR #8992, opened 2022-06-13
feat(fix): Do not lint if the target code is inside a loop

### LINKED ISSUE #8753: [maybe] false positive on clippy::needless-collect
### Summary

I think clippy has a false positive on clippy::needless-collect

See also this thread  https://gitter.im/rust-lang/rust?at=62690723d30b6b44ebbfcd12

### Lint Name

clippy::needless-collect

### Reproducer

I tried this code:

```rust
fn expensive_search() -> Vec<String> {
    vec![]
}

fn expensive_transformation(s: String) -> String {
    s
}

fn main() {
    let things: Vec<String> = expensive_search()
        .into_iter()
        .map(expensive_transformation)
        .collect();

    for i in 0..100 {
        if things.contains(&i.to_string()) {
            println!("foo");
        }
    }
}
```

I saw this happen:

```
error: avoid using `collect()` when not needed
  --> bin/main.rs:13:10
   |
13 |         .collect();
   |          ^^^^^^^
...
16 |         if things.contains(&i.to_string()) {
   |            ------------------------------- the iterator could be used here instead
   |
   = note: `-D clippy::needless-collect` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect
help: check if the original Iterator contains an element instead of collecting then checking
   |
10 ~     
11 | 
12 |     for i in 0..100 {
13 ~         if expensive_search()
14 +         .into_iter()
15 ~         .map(expensive_transformation).any(|x| x == i.to_string()) {
   |


```

nothing, should be ok

### Version

```text
rustc 1.60.0 (7737e0b5c 2022-04-04)
binary: rustc
commit-hash: 7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c
commit-date: 2022-04-04
host: x86_64-apple-darwin
release: 1.60.0
LLVM version: 14.0.0
```

### Additional Labels

_No response_

kyoto7250:
close #8753 

we consider the following code.

```rust
fn main() {
    let vec = vec![1];
    let w: Vec<usize> = vec.iter().map(|i| i * i).collect();  // <- once.

    for i in 0..2 {
        let _ = w.contains(&i);
    }
}
```

and the clippy will issue the following warning.

```rust
warning: avoid using `collect()` when not needed
 --> src/main.rs:3:51
  |
3 |     let w: Vec<usize> = vec.iter().map(|i| i * i).collect();
  |                                                   ^^^^^^^
...
6 |         let _ = w.contains(&i);
  |                 -------------- the iterator could be used here instead
  |
  = note: `#[warn(clippy::needless_collect)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_collect
help: check if the original Iterator contains an element instead of collecting then checking
  |
3 ~     
4 | 
5 |     for i in 0..2 {
6 ~         let _ = vec.iter().map(|i| i * i).any(|x| x == i);
```

Rewrite the code as indicated.

```rust
fn main() {
    let vec = vec![1];

    for i in 0..2 {
        let _ = vec.iter().map(|i| i * i).any(|x| x == i);  // <- execute `map` every loop.
    }
}
```

this code is valid in the compiler, but, it is different from the code before the rewrite.
So, we should not lint, If `collect` is outside of a loop.

Thank you in advance.

---

changelog: Do not lint if the target code is inside a loop in `needless_collect`

### DIFF
--- a/clippy_lints/src/loops/needless_collect.rs
+++ b/clippy_lints/src/loops/needless_collect.rs
@@ -185,19 +185,15 @@ struct IterFunctionVisitor<'a, 'tcx> {
 impl<'tcx> Visitor<'tcx> for IterFunctionVisitor<'_, 'tcx> {
     fn visit_block(&mut self, block: &'tcx Block<'tcx>) {
         for (expr, hir_id) in block.stmts.iter().filter_map(get_expr_and_hir_id_from_stmt) {
-            if is_loop(expr) {
+            if check_loop_kind(expr).is_some() {
                 continue;
             }
             self.visit_block_expr(expr, hir_id);
         }
         if let Some(expr) = block.expr {
-            if is_loop(expr) {
-                if let Some(higher::WhileLet { let_expr, .. }) = higher::WhileLet::hir(expr) {
-                    self.visit_block_expr(let_expr, None);
-                } else if let Some(higher::While { condition, .. }) = higher::While::hir(expr) {
-                    self.visit_block_expr(condition, None);
-                } else if let Some(higher::ForLoop { arg, .. }) = higher::ForLoop::hir(expr) {
-                    self.visit_block_expr(arg, None);
+            if let Some(loop_kind) = check_loop_kind(expr) {
+                if let LoopKind::Conditional(block_expr) = loop_kind {
+                    self.visit_block_expr(block_expr, None);
                 }
             } else {
                 self.visit_block_expr(expr, None);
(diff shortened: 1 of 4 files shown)


flip1995:
Thanks and sorry for taking so long! I did a small code cleanup and rebased the branch.

@bors r+

flip1995:
That error was spurious I think, so let's retry

@bors r+

PR #8996, opened 2022-06-13
Add whiteapple602 to Contributors list

whiteapple602:


mikerite:
What is the purpose of this change?

flip1995:
This list is not for current/new contributors, but for contributors that contributed before the license change to Apache/MIT and gave their consent to the re-licensing. Therefore this list is frozen and can't be changed anymore.

PR #8999, opened 2022-06-14
Remove error-pattern comments

Alexendoo:
The `clippy_lints` one [is unused](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60error-pattern.60), the others in `ui-toml` also appear not to have an effect

changelog: none


xFrednet:
LGTM, thank you for the cleanup

@bors r+

PR #9006, opened 2022-06-16
feat(fix): ignore `todo!` and `unimplemented!` in `if_same_then_else`

kyoto7250:
close: #8836
take over:  #8853 

This PR adds  check `todo!` and `unimplemented!` in if_same_then_else. 
( I thought `unimplemented` should not be checked as well as todo!.)

Thank you in advance.

changelog: ignore todo! and unimplemented! in if_same_then_else

r? @Jarcho

### DIFF
--- a/clippy_lints/src/copies.rs
+++ b/clippy_lints/src/copies.rs
@@ -1,5 +1,4 @@
 use clippy_utils::diagnostics::{span_lint_and_note, span_lint_and_then};
-use clippy_utils::macros::macro_backtrace;
 use clippy_utils::source::{first_line_of_span, indent_of, reindent_multiline, snippet, snippet_opt};
 use clippy_utils::{
     eq_expr_value, get_enclosing_block, hash_expr, hash_stmt, if_sequence, is_else_clause, is_lint_allowed,
@@ -13,7 +12,7 @@ use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::hygiene::walk_chain;
 use rustc_span::source_map::SourceMap;
-use rustc_span::{sym, BytePos, Span, Symbol};
+use rustc_span::{BytePos, Span, Symbol};
 use std::borrow::Cow;
 
 declare_clippy_lint! {
@@ -197,8 +196,6 @@ fn lint_if_same_then_else(cx: &LateContext<'_>, conds: &[&Expr<'_>], blocks: &[&
         .enumerate()
         .fold(true, |all_eq, (i, &[lhs, rhs])| {
             if eq.eq_block(lhs, rhs)
-                && !contains_acceptable_macro(cx, lhs)
-                && !contains_acceptable_macro(cx, rhs)
                 && !contains_let(conds[i])
                 && conds.get(i + 1).map_or(true, |e| !contains_let(e))
             {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/copies.rs
```diff
@@ -365,6 +373,35 @@ fn eq_stmts(
...
+    }
+
+    false
+}
+
+fn acceptable_macro(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    if let ExprKind::Call(call_expr, _)  = expr.kind
+        && let ExprKind::Path(QPath::Resolved(None, path)) = call_expr.kind
+        && macro_backtrace(path.span).any(|macro_call| {
+            ACCEPTABLE_MACRO.contains(&cx.tcx.item_name(macro_call.def_id).as_str())
```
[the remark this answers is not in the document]
kyoto7250: I fixed in [a5b6d25](https://github.com/rust-lang/rust-clippy/pull/9006/commits/a5b6d25ca4f1600a158ffb68301dd816c8eda3cf).

clippy_lints/src/copies.rs
```diff
@@ -365,6 +371,38 @@ fn eq_stmts(
         .all(|b| get_stmt(b).map_or(false, |s| eq.eq_stmt(s, stmt)))
 }
 
+fn contains_acceptable_macro(cx: &LateContext<'_>, block: &Block<'_>) -> bool {
+    for stmt in block.stmts {
```
Jarcho, line 375: Could be `block.stmts.iter().any(|s| matches!(s.kind, StmtKind::Semi(e) if acceptable_macro(cx, e)))`

It also seems unnecessary to check all statements for the macro. `todo` isn't usually followed by other things in the same block.

clippy_lints/src/copies.rs
```diff
@@ -365,6 +371,38 @@ fn eq_stmts(
...
+        && acceptable_macro(cx, block_expr)
+    {
+        return true
+    }
+
+    false
+}
+
+fn acceptable_macro(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    if let ExprKind::Call(call_expr, _)  = expr.kind
```
Jarcho, line 392: It would be better to not depend on the implementation of the macro. Just checking the macro backtrace would be sufficient.

Jarcho (changes requested):
This would be better added to `SpanlessEq`. That would fix the same issue in a couple other lints as well. I'll leave it up to you if you want to do that with this PR though.

kyoto7250:
@Jarcho 

I did some commits.

 Could I request another review?

clippy_utils/src/hir_utils.rs
```diff
@@ -65,7 +66,9 @@ impl<'a, 'tcx> SpanlessEq<'a, 'tcx> {
     }
 
     pub fn eq_block(&mut self, left: &Block<'_>, right: &Block<'_>) -> bool {
-        self.inter_expr().eq_block(left, right)
+        !self.cannot_be_compared_block(left)
```
Jarcho, line 69: This would be better on `eq_block` from `HirEqInterExpr`. That way it gets checked on every block seen.

clippy_utils/src/hir_utils.rs
```diff
@@ -83,6 +86,38 @@ impl<'a, 'tcx> SpanlessEq<'a, 'tcx> {
     pub fn eq_path_segments(&mut self, left: &[PathSegment<'_>], right: &[PathSegment<'_>]) -> bool {
         self.inter_expr().eq_path_segments(left, right)
     }
+
+    fn cannot_be_compared_block(&mut self, block: &Block<'_>) -> bool {
+        if block.stmts.first().map_or(false, |stmt| {
```
Jarcho, line 91: The last statement would work better. This should be caught as well:
```rust
let foo = do_something();
todo!("do bar");
```

Jarcho (changes requested):

clippy_utils/src/hir_utils.rs
kyoto7250: I done in https://github.com/rust-lang/rust-clippy/pull/9006/commits/39ffda014debd9d1d4126a8996d39af934fe8d94

clippy_utils/src/hir_utils.rs
kyoto7250: I did in https://github.com/rust-lang/rust-clippy/pull/9006/commits/46d056e2eb7c5cab767c91383cf7ef1828809fd9

Jarcho:
Looks good.

@bors r+

PR #5903, opened 2020-08-13
Needless return

jrqc:
Fixes #5858
changelog: fix false positive [`needless_return`]

PR #9008, opened 2022-06-16
emit `needless_return` on return stmts using borrow

### LINKED ISSUE #5858: `needless_return` false positive when a temporary borrows a local variable
I tried this code:

```rust
#![allow(unused)]

fn read_line() -> String {
    use std::io::BufRead;
    let stdin = ::std::io::stdin();
    return stdin.lock().lines().next().unwrap().unwrap();
}
```

I expected to see this happen: `needless_return` does not fire because the call to `lock()` borrows `stdin` and the temporary would be destroyed after the borrowed local, resulting in a `does not live long enough` compiler error.

Instead, this happened: it fired

The same fix that was applied for `let_and_return` in #5680 should work for this case. The `known problems` section of the lint description could be updated too. 

### Meta

- `cargo clippy -V`: clippy 0.0.212 (05762e3 2020-08-01)
- `rustc -Vv`:
  ```
  rustc 1.45.1 (c367798cf 2020-07-26)
  binary: rustc
  commit-hash: c367798cfd3817ca6ae908ce675d1d99242af148
  commit-date: 2020-07-26
  release: 1.45.1
  LLVM version: 10.0
  ```

### LINKED ISSUE #8978: `needless_return`  has false negative pattern
### Summary

There is a pattern that does not work even though needless_return is enabled.

[playground is here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9f84d1d415ac46d0e0fdb123b4a46462).

### Lint Name

needless_return

### Reproducer

I tried this code:

```rust
#![warn(clippy::needless_return)]
const WORDS: [&[&str]; 2] = [&["foo"], &["bar"]];

fn main() {
    println!("{:?}", issue());
}

fn issue() -> bool {
    return WORDS.iter().any(|&word| word == ["foo"]);
}
```

I expected to see this happen: There should be a suggestion as above follows.

```rust
fn issue() -> bool {
    WORDS.iter().any(|&word| word == ["foo"])
}
```

Instead, this happened: No happen.

### Version

```text
rustc 1.61 (stable)
clippy 1.63
```

dswij:
Fix #8978

This check is added on #5903 to fix #5858. Though, the test added here compiles as of today, so this can be removed. ([playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=7c0b5d980302c2de89d1aa7f59441934))

changelog: [`needless_return`] now emits on on return stmts using borrow

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -181,15 +181,12 @@ fn check_final_expr<'tcx>(
             // allow `#[cfg(a)] return a; #[cfg(b)] return b;`
             let attrs = cx.tcx.hir().attrs(expr.hir_id);
             if !attrs.iter().any(attr_is_cfg) {
-                let borrows = inner.map_or(false, |inner| last_statement_borrows(cx, inner));
-                if !borrows {
-                    emit_return_lint(
-                        cx,
-                        span.expect("`else return` is not possible"),
-                        inner.as_ref().map(|i| i.span),
-                        replacement,
-                    );
-                }
+                emit_return_lint(
+                    cx,
+                    span.expect("`else return` is not possible"),
+                    inner.as_ref().map(|i| i.span),
+                    replacement,
+                );
             }
         },
         // a whole block? check it!
--- a/tests/ui/needless_return.fixed
+++ b/tests/ui/needless_return.fixed
@@ -89,7 +89,7 @@ fn test_nested_match(x: u32) {
 fn read_line() -> String {
     use std::io::BufRead;
     let stdin = ::std::io::stdin();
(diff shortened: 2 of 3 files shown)


Alexendoo:
I think that's due to rust-lang/rust#93965


```rust
use std::cell::RefCell;

fn foo(x: RefCell<String>) -> String {
    let y = x;
    return y.borrow().clone();
}
```

wouldn't compile if the return were removed (from https://github.com/rust-lang/rust/blob/9a74608543d499bcc7dd505e195e8bfab9447315/src/test/ui/span/issue-23338-locals-die-before-temps-of-body.rs)

dswij:
Oh wait, you're right. 😅 
Closing this one

PR #9016, opened 2022-06-18
Use `RefCell` in `needless_return` tests

Alexendoo:
changelog: none

The stdio locks no longer fail to compile if the `return` is removed due to them now being `'static` (#9008)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #9017, opened 2022-06-18
Example for `WHILE_LET_ON_ITERATOR`

alex-semenyuk:
changelog: none

example for `WHILE_LET_ON_ITERATOR`, using `for` instead of `while let`

### DIFF
--- a/clippy_lints/src/loops/mod.rs
+++ b/clippy_lints/src/loops/mod.rs
@@ -346,7 +346,14 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```ignore
-    /// while let Some(val) = iter() {
+    /// while let Some(val) = iter.next() {
+    ///     ..
+    /// }
+    /// ```
+    ///
+    /// Use instead:
+    /// ```ignore
+    /// for val in &mut iter {
     ///     ..
     /// }
     /// ```


clippy_lints/src/loops/mod.rs
```diff
@@ -346,14 +346,21 @@ declare_clippy_lint! {
...
+    /// ```ignore
+    /// for val in &mut iter {
+    ///     ..
+    /// }
+    /// ``` 
     #[clippy::version = "pre 1.29.0"]
     pub WHILE_LET_ON_ITERATOR,
     style,
-    "using a `while let` loop instead of a for loop on an iterator"
+    "using a `for` loop instead of a `while let` loop on an iterator"
```
giraffate, line 363: This is what Clippy pointed out here, so no need to correct it.

giraffate (commented):
Thanks for improving Clippy!

The test fails, so can `cargo dev fmt` be done? And can you squash commits?

giraffate (approved):
It looks good, thanks!

Can you do a rebase, not merge? We follow a no merge-commit policy: https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/basics.md#pr
And can you squash commits into one?

alex-semenyuk:
> It looks good, thanks!
> 
> Can you do a rebase, not merge? We follow a no merge-commit policy: https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/basics.md#pr And can you squash commits into one?

@[giraffate](https://github.com/giraffate) Done

giraffate:
@bors r+

Thanks!

PR #9030, opened 2022-06-21
Remove `cargo_dev` dependency

Serial-ATA:
changelog: none

Sorry about that :sweat_smile:.

r? @flip1995


flip1995:
Thanks for addressing this so quickly!

@bors r+

PR #9031, opened 2022-06-21
Add [`manual_rem_euclid`] lint

### LINKED ISSUE #8883: New lint: manual reimplementation of `.rem_euclid()`
### What it does

Checks for manual reimplementations of the `.rem_euclid()` function.

### Lint Name

_No response_

### Category

complexity

### Advantage

When people search for how to calculate positive modulos of integers, top Google results and Stack Overflow tend to feature the `((a % n) + n) % n` method as a solution to the problem. Rust has function to calculate this, `.rem_euclid()`. It would be great to suggest people using this function instead (only in cases where `n` can be proven as positive).

### Drawbacks

_No response_

### Example

```rust
let manual_1: i32 = ((value % 4) + 4) % 4;
let manual_2: i32 = (4 + (value % 4)) % 4;
let manual_3: i32 = (value % 4 + 4) % 4;
let manual_4: i32 = (4 + value % 4) % 4;
```

Could be written as:

```rust
let expected: i32 = value.rem_euclid(4);
```

evantypanski:
Closes #8883

Adds a lint for checking manual use of `rem_euclid(n)`

changelog: Add [`manual_rem_euclid`] lint


clippy_lints/src/manual_rem_euclid.rs
```diff
@@ -0,0 +1,117 @@
...
+        } else if let Some(int_const) = constant_full_int(cx, cx.typeck_results(), right) {
+            (int_const, left)
+        } else {
+            return None;
+        }
+    } else {
+        return None;
+    };
+
+    if int_const > FullInt::S(0) {
```
mikerite, line 108: I would rewrite this to not have the `unwrap()`. Something like
```
let val = match int_const {
    FullInt::S(s) => s.try_into(),
    FullInt::U(u) => Some(u)
}
Some((val?, other_op))
```
There might be a better way of doing it.

evantypanski: Agreed, believe this is in the spirit of that https://github.com/rust-lang/rust-clippy/pull/9031/commits/75ed0c9f267c8e65d3bbd45c6e10acd11e702077

clippy_lints/src/manual_rem_euclid.rs
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    extract_msrv_attr!(LateContext);
+}
+
+// Takes a binary expression and separates the operands into the int constant and the other
+// operand. Ensures the int constant is positive. Operators that are not commutative must have the
+// constant appear on the right hand side to return a value.
+fn check_for_positive_int_constant<'a>(
```
Jarcho, line 102: This would be better split into two functions. One that checks if an expression is a positive int constant `&Expr -> Option<u128>`, and another that takes two expressions and returns one as a constant `(&Expr, &Expr) -> Option<(u128, &Expr)>`). This will remove the need for the bool argument and get rid of the extra check for `ExprKind::Binary`

clippy_lints/src/manual_rem_euclid.rs
Jarcho: You can match on the constant and skip the bounds check here. i.e.
```rust
match x {
    S(x) => x.try_into().ok(),
    U(x) => Some(x),
}
```

evantypanski: Ah, realized you can't do `x % 0` because zero divisor is invalid, so the equal checks for the consts makes sure the const is nonzero so some checks were completely unnecessary, thanks

Jarcho:
There should be a call to [in_external_macro](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/lint/fn.in_external_macro.html#), and an msrv test for the const version with the msrv set between the two versions. Otherwise it all looks good.

PR #9032, opened 2022-06-21
enum_variant_names should ignore when all prefixes are _

### LINKED ISSUE #9018: `enum_variant_names` shouldn't trigger when all prefixes are `_`
`_` is a special case contrary to any other user-defined prefix.

```rust
enum Enum {
    _Bar,
    _Foo,
}
```

kyoto7250:
close #9018 

When Enum prefix is only an underscore, we should not issue warnings.

changelog: fix false positive in enum_variant_names

### DIFF
--- a/clippy_lints/src/enum_variants.rs
+++ b/clippy_lints/src/enum_variants.rs
@@ -213,7 +213,7 @@ fn check_variant(cx: &LateContext<'_>, threshold: u64, def: &EnumDef<'_>, item_n
 }
 
 #[must_use]
-fn have_no_extra_prefix(prefixes: &Vec<&str>) -> bool {
+fn have_no_extra_prefix(prefixes: &[&str]) -> bool {
     prefixes.is_empty() || prefixes.join("") == "_"
 }
 
--- a/clippy_lints/src/enum_variants.rs
+++ b/clippy_lints/src/enum_variants.rs
@@ -190,7 +190,7 @@ fn check_variant(cx: &LateContext<'_>, threshold: u64, def: &EnumDef<'_>, item_n
             .map(|e| *e.0)
             .collect();
     }
-    let (what, value) = match (pre.is_empty(), post.is_empty()) {
+    let (what, value) = match (have_no_extra_prefix(&pre), post.is_empty()) {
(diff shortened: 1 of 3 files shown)


clippy_lints/src/enum_variants.rs
```diff
@@ -212,6 +212,11 @@ fn check_variant(cx: &LateContext<'_>, threshold: u64, def: &EnumDef<'_>, item_n
     );
 }
 
+#[must_use]
+fn have_no_extra_prefix(prefixes: &Vec<&str>) -> bool {
```
llogiq, line 216: Shouldn't that be

```suggestion
fn have_no_extra_prefix(prefixes: &[&str]) -> bool {
```

I'm a bit surprised this passes dogfood.

clippy_lints/src/enum_variants.rs
```diff
@@ -212,6 +212,11 @@ fn check_variant(cx: &LateContext<'_>, threshold: u64, def: &EnumDef<'_>, item_n
...
+    prefixes.is_empty() || prefixes.join("") == "_"
```
llogiq, line 217:
```suggestion
    prefixes.all(|p| p == "" || p == "_")
```
would admittedly allow multiple underscores (is that a bad thing?), but avoid the overhead of `join`.

llogiq:
Thank you for the PR. I only have a few suggestions / questions, otherwise this should be good to merge.

clippy_lints/src/enum_variants.rs
kyoto7250: I fixed in https://github.com/rust-lang/rust-clippy/pull/9032/commits/d827b834d91383620d587fb4a1e74d8579e3a280

llogiq:
Thank you!

@bors r+

llogiq:
It seems there is some markdown error.

kyoto7250:
I created  https://github.com/rust-lang/rust-clippy/pull/9055

xFrednet:
Since https://github.com/rust-lang/rust-clippy/pull/9055 was merged, we can see if this fixed the error :upside_down_face: 

@bors retry

PR #9034, opened 2022-06-22
Example for VEC_RESIZE_TO_ZERO

alex-semenyuk:
changelog: none

Example for VEC_RESIZE_TO_ZERO - what we should use instead


giraffate:
@bors r+

Thanks!

PR #9036, opened 2022-06-23
Check for `--force-warn` in Clippy's driver run condition

xFrednet:
Just a thing I've noticed while tinkering on the driver. Currently, the driver only checks for `--cap-lints=allow` to determine if Clippy should run on the code, but ignores possible `--force-warn` arguments

---

changelog: Others: Allowing all lints and only `--force-warn`ing some will now work with Clippy's driver

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -60,6 +60,7 @@ fn test_arg_value() {
     assert_eq!(arg_value(args, "--bar", |p| p == "foo"), None);
     assert_eq!(arg_value(args, "--foobar", |p| p == "foo"), None);
     assert_eq!(arg_value(args, "--foobar", |p| p == "123"), Some("123"));
+    assert_eq!(arg_value(args, "--foobar", |p| p.contains("12")), Some("123"));
     assert_eq!(arg_value(args, "--foo", |_| true), None);
 }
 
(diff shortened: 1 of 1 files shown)


src/driver.rs
```diff
@@ -60,9 +60,55 @@ fn test_arg_value() {
     assert_eq!(arg_value(args, "--bar", |p| p == "foo"), None);
     assert_eq!(arg_value(args, "--foobar", |p| p == "foo"), None);
     assert_eq!(arg_value(args, "--foobar", |p| p == "123"), Some("123"));
+    assert_eq!(arg_value(args, "--foobar", |p| p.contains("12")), Some("123"));
     assert_eq!(arg_value(args, "--foo", |_| true), None);
 }
 
+/// This function checks if the arguments contain a `--force-warn` for any clippy lint. This
+/// is different from [`arg_value`] as `--force-warn` can be present more than once.
```
Alexendoo, line 68: Maybe it shouldn't, but `arg_value` does check multiple args from left to right until one returns `true` so I think you could use it here. Found this out wondering why `--cap-lints=allow --cap-lints=warn` didn't force enable clippy

xFrednet: Ohh, you're right! Thank you for pointing that out. Then this change became even simpler ^^

xFrednet:
Yes, the `--force-warn` flag is defined to warn, even if `--cap-lints allow` is set. `--force-warn clippy::` will be accepted by that implementation, but should issue a warning, since `clippy::` is not a valid lint name. The whole `cap_lints_allow` is just a test to safe performance for dependencies, AFAIK.

dswij (approved):
Makes sense. LGTM, thanks for this!

xFrednet:
Thanks for the review! Is there anything else that needs to be done before the merge or did you forget to r+? :upside_down_face:

dswij:
Oops, you're right 😅 

@bors r+

src/driver.rs
```diff
@@ -334,15 +335,13 @@ pub fn main() {
         // - IF Clippy is run on the main crate, not on deps (`!cap_lints_allow`) THEN
         //    - IF `--no-deps` is not set (`!no_deps`) OR
         //    - IF `--no-deps` is set and Clippy is run on the specified primary package
-        let cap_lints_allow = arg_value(&orig_args, "--cap-lints", |val| val == "allow").is_some();
+        let cap_lints_allow = arg_value(&orig_args, "--cap-lints", |val| val == "allow").is_some()
+            && arg_value(&orig_args, "--force-warn", |val| val.contains("clippy::")).is_none();
```
flip1995, line 339: Couldn't there be more than one `--force-warn` flag, one with a rustc lint and another with a clippy lint? I think in that case, this doesn't quite work.

PR #9039, opened 2022-06-23
Add `cargo dev dogfood`

Serial-ATA:
changelog: Add `cargo dev dogfood`

Part of #5394

### DIFF
--- a/book/src/development/basics.md
+++ b/book/src/development/basics.md
@@ -59,7 +59,7 @@ cargo uitest
 # only run UI tests starting with `test_`
 TESTNAME="test_" cargo uitest
 # only run dogfood tests
-cargo test --test dogfood
+cargo dev dogfood
 ```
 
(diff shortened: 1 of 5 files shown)


xFrednet:
Thanks for working on Clippy's internal tools! Wouldn't it be simpler to add new cargo aliases instead of extending `clippy_dev`?

Serial-ATA:
I'm not sure how to best handle `--fix` in that case.

xFrednet:
Good point, I just have never used `--fix` with dogfood, and in general tbh :sweat_smile:. We could have an environment value, like we have to uitest, but that feels hacky. Then I think this is a good solution! :upside_down_face:

giraffate (commented):
Can the document about the dogfood tests be updated?
https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/basics.md#building-and-testing

giraffate:
@bors r+

Thanks!

PR #8518, opened 2022-03-10
Migrate write.rs to a late pass

Alexendoo:
changelog: Migrates write.rs from a pre expansion pass to a late pass
changelog: [`positional_named_format_parameters`] is renamed in favour of the rustc lint `named_arguments_used_positionally`

- Macros are now identified by diagnostic items, so will no longer lint user defined macros named, e.g. a custom `print!`
- `print_literal`/`write_literal` no longer lint no longer lint literals that come from macro expansions, e.g. `env!("FOO")`
- `print_with_newline`/`write_with_newline` no longer lint strings with any internal `\r` or `\n`s

~~A false negative, `print_literal`/`write_literal` don't lint format strings that produce `FormatSpec`s, e.g. ones containing pretty print/width/align specifiers~~

Suggestion changes:
- ~~`print_literal`/`write_literal` no longer have suggestions, as the spans for the `{}`s were not easily obtainable~~
-  `print_with_newline`/`write_with_newline` has a better suggestion for a sole literal newline, but no longer has suggestions for len > 1 strings that end in a literal newline
- ~~`use_debug` spans are less precise, now point to the whole format string~~

The diff for write.rs is pretty unwieldy, other than for the `declare_clippy_lint!`s I think you'd be better off viewing it as a brand new file rather than looking at the diff, as it's mostly written from scratch

cc #6610, fixes #5721, fixes #7195, fixes #8615

PR #9040, opened 2022-06-23
Add new lint [`unused_named_parameter`]

### LINKED ISSUE #9027: Warn about unused named parameter in format strings
### What it does

Very few beginners know that named parameters can be used in format strings. (https://doc.rust-lang.org/std/fmt/index.html#named-parameters) This can often lead to great confusion as it looks like you are setting the variable to that value. 

Additionally since you can now capture identifiers in format strings, named parameters will become less and less common. As such this new lint warns if a named parameter goes unused.

### Lint Name

unused_named_parameter

### Category

suspicious

### Advantage

- Removes redundant code
- Prevents named parameters from being confused with an assignment such as `print!("{}", x = x+1)`

### Drawbacks

_No response_

### Example

```rust
println!("{}", x = 10);
```

Could be written as:

```rust
println!("{}", 10);
```

miam-miam100:

changelog: Add new lint [`unused_named_parameter`] to warn when named parameters in format strings go unused
fixes #9027


clippy_lints/src/write.rs
```diff
@@ -256,6 +258,25 @@ declare_clippy_lint! {
...
+    /// ### Why is this bad?
+    /// It's verbose and it may be confused with an assignment.
+    /// ### Example
+    /// ```rust
+    /// println!("{}", x = 10);
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// println!("{}", 10);
+    /// ```
```
dswij, line 273: Can you help to add some space between the headings?

tests/ui/unused_named_parameter.rs
```diff
@@ -0,0 +1,44 @@
...
+    println!("hello {:0$}!", width = 2);
+    println!("{0} {1:?}", zero = 0, one = 1);
+
+    writeln!(v, "hello {world} {}", random = 4);
+    writeln!(v, "hello {}", world = world);
+    writeln!(v, "{} {number:.5}", prec = 5, number = 0.01);
+    writeln!(v, "{0:.prec$}", number = 0.01, prec = 5);
+    writeln!(v, "{{ {}", x = world);
+    writeln!(v, "hello {:0$}!", width = 2);
+    writeln!(v, "{0} {1:?}", zero = 0, one = 1);
```
dswij, line 43: Nice extensive tests!

dswij (commented):
Thanks for the patience! The changes in the PR look great to me and is certainly a nice addition! Just some comments

clippy_lints/src/write.rs
```diff
@@ -256,6 +258,25 @@ declare_clippy_lint! {
...
+    #[clippy::version = "1.63.0"]
+    pub UNUSED_NAMED_PARAMETER,
```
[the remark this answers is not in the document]
miam-miam100: I'm happy with `unused_named_parameters` but I feel like `unused_named_parameters_in_format` is a bit too verbose, maybe `unused_named_format_parameters` could work instead?

clippy_lints/src/write.rs
```diff
@@ -423,6 +446,10 @@ impl SimpleFormatArgs {
         self.named.iter().find(|x| *n == x.0).map_or(&[], |x| x.1.as_slice())
     }
 
+    fn get_referenced(&self, n: &Path) -> bool {
```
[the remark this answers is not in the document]
miam-miam100: Used `is_referenced` instead as `has_reference` does not make it entirely clear as to what is meant by a "reference".

estebank:
@miam-miam100 it might still be worth it to have a clippy lint against named params *ever* being referred to positionally, which would be noisier/stricter than what this PR and rustc do. That way rustc avoids false positives but there's still a mechanism to detect "iffy" cases.

miam-miam100:
Hmmm so if I've understood correctly we should have a lint that prevents you from writing this?
```rust
println!("hello {world} {}", world = "world");
```

estebank:
Exactly.

miam-miam100:
I don't know what rustc version clippy is using but it looks like my two errors are due to these two PRs not being merged in yet: https://github.com/rust-lang/rust/pull/99263 and https://github.com/rust-lang/rust/pull/9948. @dswij Do you know what I can do? Do I just have to wait?

dswij:
I believe `clippy` use the toolchain version defined here: https://github.com/rust-lang/rust-clippy/blob/master/rust-toolchain.

The job failed when trying to apply the fix for the lint 😕. 
Did this pass on your local machine?

miam-miam100:
The changes do not currently pass on my local machine as the toolchain version clippy uses does not include the changes I need for the lint to work correctly. However you are correct in saying that the rust fix file was incorrectly written and has now been fixed.

dswij:
That makes sense. We can wait for the weekly sync then. The next scheduled one is pretty soon (by the next 24 hours). We can come back and rebase the PR afterward.

nyurik:
I have been hacking on a var inlining lint available since 1.58 - it is based on another possibly conflicting pr https://github.com/rust-lang/rust-clippy/pull/9233

I could really use the code 1)to check if an argument is an alias to another argument or width or precision, 2) get the spans of all three possible usages so I can inline vars and 3) possibly share the same code between rustc and clippy (keep all buggy parsing code in one place 😀). I have written some test cases you might want to use, eg with some unexpected whitespaces.

Awesome to see so much format-related work!

miam-miam100:
That's awesome, that does not sound all too easy so I wish you the best of luck on that! I added the test cases to my code which look like they should all pass once we update the nightly version.

miam-miam100:
Nice! Successfully rebased, @dswij feel free to review me whenever you are free.

nyurik (commented):
@miam-miam100 I made a tiny suggestion, but the bigger question is how this PR should be merged with the @Alexendoo 's #8518 -- it seems both make a number of significant changes to the `clippy_lints/src/write.rs`, nearly rewriting it. Would it make sense to perhaps base your PR on top of @Alexendoo 's, so that all of them use late context?

miam-miam100:
Hmmm I'm not sure that is a great idea since we don't know when that PR will land and this PR does not really benefit from the late pass in any way.

nyurik:
@miam-miam100 my interest is getting the #9233 in, which I was basing on that. I am not yet certain I need to base my lint on that PR, or if I can do it all with the early context.  Perhaps your PR could be refactored a bit so that we can have shared functions to parse params and get back the spans?  My main requirement:

* detect format!(...) - the string and all arguments
* find spans for each part of the parameter:   `{<arg>:<width>$.<precision>}` -->   `struct FormatStrArg {  arg: Span, widthArg: Optional<Span>, precisionArg: Optional<Span> }`  (note that the `arg` span could be empty, but must be present, whereas width/precision cannot be empty, but could be `None`)
* for each arg and non-missing width/precision in `FormatStrArg`, find corresponding argument that can be analyzed checked if it is it a local var / a fn parameter, or is it something else.

Do you think your functions could produce this result? If so, I could base my work on top of your PR that doesn't try to change the overall structure of the format code.  Thx!

miam-miam100:
Ah well fortunately for you, all those things are directly exposed by rustc. All the arg, width and precision spans are given, as you can see here (https://github.com/miam-miam100/rust-clippy/blob/631525a21fa3b4bd0c495da11c3b1c0eec84c5ff/clippy_lints/src/write.rs#L543). As for checking if the corresponding argument is a local var you can do that here https://github.com/miam-miam100/rust-clippy/blob/631525a21fa3b4bd0c495da11c3b1c0eec84c5ff/clippy_lints/src/write.rs#L667 (Checking if it is an ident). From what you've told me you should just be able to check that there and span the lin

Alexendoo:
To explain the situation a bit, `write.rs` currently uses a pre-expansion pass, a variant of an early pass that runs before macro expansion has happened https://github.com/rust-lang/rust-clippy/blob/53a09d48557119d2b2a875d77a284b470a95e1c8/clippy_lints/src/lib.rs#L419-L428

The plan is to move away from the pre-expansion pass as it has a few issues, as an example for this lint:

```rust
// lib.rs
#![allow(clippy::positional_named_format_parameters)]

mod foo;
```
```rust
fn x() {
    // this would still trigger the lint, ignoring the crate wide allow
    println!("hello {world} {}", world = "world");
}
```

Since it would be pretty unlikely someone will want to `allow` this lint, I'm fine with it being added to `write.rs` in the meanwhile. As you say the rewrite may take a while to be merged as it's a rather large change

For #9233 I could see people wanting to `allow` it crate wide however, either due to personal taste or to avoid churn. For that reason I think it should stick to a late pass. Some time in the next few days I'll see if it's feasible to split off the `clippy_utils::macro` changes into their own PR so you don't have to wait for the whole thing @nyurik, as well as adding the missing details you need

nyurik:
@Alexendoo thank you for the detailed answer!  What are you thoughts about integrating the new rustc format string parsing capabilities like `rustc_parse_format::Parser`, i.e. as used in [here](https://github.com/miam-miam100/rust-clippy/blob/631525a21fa3b4bd0c495da11c3b1c0eec84c5ff/clippy_lints/src/write.rs#L543)?

dswij (commented):
The lint itself looks great, just the lint description that can be improved, IMO.

Can you help squash the commits, and after that, I think it's good to merge.

Alexendoo:
@nyurik Yeah, just waiting for the next sync that includes https://github.com/rust-lang/rust/pull/99987, I'll ping you in the PR once it's ready

tests/ui/positional_named_format_parameters.rs
```diff
@@ -0,0 +1,56 @@
+// run-rustfix
+#![allow(unused_must_use)]
+#![allow(named_arguments_used_positionally)] // Unstable at time of writing.
```
Alexendoo, line 3: Is there any overlap between this lint and `named_arguments_used_positionally`? `clippy::positional_named_format_parameters` will reach stable at the same time or after `named_arguments_used_positionally` does so ideally we wouldn't be double warning

miam-miam100:
Right that's great thanks, I'm currently on holiday so I'll do that when I get back. (a week's time)

tests/ui/positional_named_format_parameters.rs
miam-miam100: Yes for a very specific case (when a named argument is used only once positionally, is never named and does not have a width, modifier or precision) so it should be fine.

Additionally inhibiting such a case may prove more confusing to the end user since they would expect this lint to catch such a case.

clippy_lints/src/write.rs
```diff
@@ -256,6 +257,28 @@ declare_clippy_lint! {
     "writing a literal with a format string"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint warns when a named parameter in a format string is used as a positional one.
+    ///
+    /// ### Why is this bad?
+    /// It's verbose and it may be confused with an assignment.
```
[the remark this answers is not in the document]
miam-miam100: Updated the description :+1:

miam-miam100:
@dswij All squashed!

dswij:
Huge thanks for this @miam-miam100! This is a nice lint to have for sure.

@bors r+

PR #9046, opened 2022-06-25
Fix `#[expect]` for most clippy lints

xFrednet:
This PR fixes most `#[expect]` - lint interactions listed in rust-lang/rust#97660. [My comment in the issue](https://github.com/rust-lang/rust/issues/97660#issuecomment-1147269504) shows the current progress (Once this is merged). I plan to work on `duplicate_mod` and `multiple_inherent_impl` and leave the rest for later. I feel like stabilizing the feature is more important than fixing the last few nits, which currently also don't work with `#[allow]`.

---

changelog: none

r? @Jarcho 

cc: rust-lang/rust#97660


clippy_utils/src/lib.rs
```diff
@@ -1539,9 +1539,13 @@ pub fn is_try<'tcx>(cx: &LateContext<'_>, expr: &'tcx Expr<'tcx>) -> Option<&'tc
...
-/// Returns `true` if the lint is allowed in the current context
+/// Returns `true` if the lint is allowed in the current context. This is useful for
+/// skipping long running code when it's unnecessary
 ///
-/// Useful for skipping long running code when it's unnecessary
+/// This function should check the lint level for the same node, that the lint will
+/// be emitted at. If the information is buffered to be emitted at a later point, please
+/// make sure to use `span_lint_hir` functions to emit the lint. This ensures that
+/// expectations at the checked nodes will be fulfilled.
 pub fn is_lint_allowed(cx: &LateContext<'_>, lint: &'static Lint, id: HirId) -> bool {
```
flip1995, line 1549: Doesn't `LintContext` now have a `get_lint_level` function, that we could use to also perform this check on `EarlyLintPass`es?

tests/ui/boxed_local.rs
```diff
@@ -202,3 +203,8 @@ mod issue4804 {
         fn foo(x: Box<u32>) {}
     }
 }
+
+#[expect(clippy::boxed_local)]
```
Jarcho, line 207: Shouldn't this be on the argument?

xFrednet: I wasn't sure, sometimes I forget that attributes also work on arguments ^^. I've updated the test.

PR #9053, opened 2022-06-25
Fix suggestions for `async` closures in redundant_closure_call

### LINKED ISSUE #9052: redundant_closure_call gives invalid suggestion for an async closure
### Summary

I was creating an async closure and calling it immediately to get an opaque object I could pass to `impl Future`. Clippy suggested I replace this with a non-async block, which gave a compile error.

### Reproducer

Running Clippy on this minimal example:

```rust
#![feature(async_closure)]

use futures::Future;

fn run_in_background(future: impl Future<Output = ()>) {
    // Imagine this did something useful!
}

fn main() {
    run_in_background((async || {
        // Do something...
    })())
}
```

Gives this lint:

```
warning: try not to call a closure in the expression where it is declared
  --> src/main.rs:10:23
   |
10 |       run_in_background((async || {
   |  _______________________^
11 | |         // Do something...
12 | |     })())
   | |________^
   |
   = note: `#[warn(clippy::redundant_closure_call)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_call
help: try doing something like
   |
10 ~     run_in_background({
11 +         // Do something...
12 ~     })
   |
```

But using this suggestion gives a compile error:

```
error[[E0277]](https://doc.rust-lang.org/nightly/error-index.html#E0277): `()` is not a future
  --> src/main.rs:10:23
   |
10 |       run_in_background({
   |  _____-----------------_^
   | |     |
   | |     required by a bound introduced by this call
11 | |         // Do something...
12 | |     })
   | |_____^ `()` is not a future
   |
   = help: the trait `futures::Future` is not implemented for `()`
   = note: () must be a future or must implement `IntoFuture` to be awaited
note: required by a bound in `run_in_background`
  --> src/main.rs:5:35
   |
5  | fn run_in_background(future: impl Future<Output = ()>) {
   |                                   ^^^^^^^^^^^^^^^^^^^ required by this bound in `run_in_background`
```

Whereas this suggestion would have worked fine:

```
run_in_background(async {
    // Do something...
})
```

### Version

```text
rustc 1.61.0-nightly (5f3700105 2022-03-22)
binary: rustc
commit-hash: 5f37001055c29982f4c27ee9edd90449c8e07774
commit-date: 2022-03-22
host: aarch64-apple-darwin
release: 1.61.0-nightly
LLVM version: 14.0.0
```

### Additional Labels

AaronC81:
Fixes #9052

changelog: Fix suggestions given by [`redundant_closure_call`] for async closures

### DIFF
--- a/clippy_lints/src/redundant_closure_call.rs
+++ b/clippy_lints/src/redundant_closure_call.rs
@@ -1,5 +1,5 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_then};
-use clippy_utils::source::snippet_with_applicability;
+use clippy_utils::sugg::Sugg;
 use if_chain::if_chain;
 use rustc_ast::ast;
 use rustc_ast::visit as ast_visit;
@@ -69,7 +69,7 @@ impl EarlyLintPass for RedundantClosureCall {
         if_chain! {
             if let ast::ExprKind::Call(ref paren, _) = expr.kind;
             if let ast::ExprKind::Paren(ref closure) = paren.kind;
-            if let ast::ExprKind::Closure(_, _, _, _, ref decl, ref block, _) = closure.kind;
+            if let ast::ExprKind::Closure(_, _, ref r#async, _, ref decl, ref block, _) = closure.kind;
             then {
                 let mut visitor = ReturnVisitor::new();
                 visitor.visit_expr(block);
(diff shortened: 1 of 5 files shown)


AaronC81:
Just realised the suggestion is incorrect for single-expression blocks: `(async || 42)()` gives the suggestion `async 42` which is a syntax error. I'll see if I can sort this out

flip1995:
@bors r+

Thanks!

flip1995:
What happened here? :thinking: Was bors on vacation?

@bors r+

PR #9056, opened 2022-06-27
Allow `let_unit_value` in more cases

### LINKED ISSUE #8998: `let_unit_value` triggers on generic return value
### Summary

When using [`objc`](https://github.com/SSheldon/rust-objc), a library commonly used when dealing with Objective-C, you often have to use `let _: () = ...` to tell the compiler what the return type, since the return type is generic.

This triggers the lint. Trying to apply the lint's suggestion results in a compilation error. The lint doesn't trigger on Rust `1.61.0`.

### Lint Name

`let_unit_value`

### Reproducer

I tried the following code (see [this playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c7def16278b6fe293de7d23b422f8ce6) for a simplified testcase):
```rust
// objc v0.2.7
use objc::{class, msg_send};
use objc::runtime::Object;

let cls = class!(NSObject);
let obj: *mut Object = unsafe { msg_send![cls, new] };
let _: () = unsafe { msg_send![obj, release] };
```

I expected the lint to not trigger, instead it triggers on the last statement.

### Version

```text
rustc 1.63.0-nightly (ec55c6130 2022-06-10)
binary: rustc
commit-hash: ec55c61305eaf385fc1b93ac9a78284b4d887fe5
commit-date: 2022-06-10
host: x86_64-apple-darwin
release: 1.63.0-nightly
LLVM version: 14.0.5
```

### Additional Labels

@rustbot label +regression-from-stable-to-nightly

Jarcho:
fixes #8998

changelog: Lint `let_unit_value` less aggressively when the type is specified

### DIFF
--- a/clippy_lints/src/unit_types/let_unit_value.rs
+++ b/clippy_lints/src/unit_types/let_unit_value.rs
@@ -18,20 +18,23 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, local: &'tcx Local<'_>) {
         && !in_external_macro(cx.sess(), local.span)
         && cx.typeck_results().pat_ty(local.pat).is_unit()
     {
-        if local.ty.is_some() && expr_needs_inferred_result(cx, init) {
-            if !matches!(local.pat.kind, PatKind::Wild) {
+        if (local.ty.map_or(false, |ty| !matches!(ty.kind, TyKind::Infer))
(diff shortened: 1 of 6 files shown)


Jarcho:
Ping @llogiq. I don't remember if I changed anything since you last checked this.

llogiq:
In any case, this looks good.

@bors r+

PR #9073, opened 2022-06-30
add `[unnecessary_reserve]` lint

### LINKED ISSUE #8982: Unnecessary Vec/VecDeque reserve lint
### What it does

Inspired by:
https://github.com/rust-lang/rust/pull/97922

In a basic case like this Clippy could suggest to remove the call to reserve() because it's unnecessary:

### Lint Name

unnecessary_reserve

### Category

perf

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
fn test1(v: &mut Vec<u32>, b: &[u32]) {
    v.reserve(b.len());
    v.extend(b);
}
```
The same for VecDeque.

kyoto7250:
close #8982 

This PR adds `[unnecessary_reserve]` lint.

~~note: This lint targets next stable versions (1.62), so we cannot merge it now.~~

---

changelog: add `[unnecessary_reserve]` lint
[#9073](https://github.com/rust-lang/rust-clippy/pull/9073)

### DIFF
--- a/clippy_lints/src/unnecessary_reserve.rs
+++ b/clippy_lints/src/unnecessary_reserve.rs
@@ -1,0 +1,141 @@
+use clippy_utils::diagnostics::span_lint;
+use clippy_utils::{match_def_path, meets_msrv, msrvs, paths, visitors::expr_visitor_no_bodies};
+use rustc_hir::{intravisit::Visitor, Block, ExprKind, QPath, StmtKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty;
+use rustc_semver::RustcVersion;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint checks for `reserve` before calling the `extend` method.
+    /// ### Why is this bad?
+    /// vec::reserve method before vec::extend is no longer makes sense in rustc version >= 1.62
+    /// ### Example
+    /// ```rust
+    /// let mut vec: Vec<usize> = vec![];
+    /// let array: &[usize] = &[1, 2];
+    /// vec.reserve(array.len());
+    /// vec.extend(array);
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let mut vec: Vec<usize> = vec![];
+    /// let array: &[usize] = &[1, 2];
+    /// vec.extend(array);
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub UNNECESSARY_RESERVE,
+    pedantic,
+    "`reserve` method before `extend` is no longer makes sense in rustc version >= 1.62"
+}
+
+pub struct UnnecessaryReserve {
+    msrv: Option<RustcVersion>,
+}
+
+impl UnnecessaryReserve {
+    #[must_use]
+    pub fn new(msrv: Option<RustcVersion>) -> Self {
+        Self { msrv }
+    }
+}
+
+impl_lint_pass!(UnnecessaryReserve => [UNNECESSARY_RESERVE]);
+
+impl<'tcx> LateLintPass<'tcx> for UnnecessaryReserve {
+    fn check_block(&mut self, cx: &LateContext<'tcx>, block: &Block<'tcx>) {
+        if !meets_msrv(self.msrv, msrvs::UNNECESSARY_RESERVE) {
+            return;
+        }
+
+        for (idx, stmt) in block.stmts.iter().enumerate() {
+            if let StmtKind::Semi(semi_expr) = stmt.kind
+                && let ExprKind::MethodCall(_, [struct_calling_on, _], _) = semi_expr.kind
+                && let Some(expr_def_id) = cx.typeck_results().type_dependent_def_id(semi_expr.hir_id)
+                && (match_def_path(cx, expr_def_id, &paths::VEC_RESERVE) ||
+                    match_def_path(cx, expr_def_id, &paths::VEC_DEQUE_RESERVE))
+                && acceptable_type(cx, struct_calling_on)
+                && let Some(next_stmt_span) = check_extend_method(cx, block, idx, struct_calling_on)
+                && !next_stmt_span.from_expansion()
+            {
+                span_lint(
+                    cx,
+                    UNNECESSARY_RESERVE,
+                    next_stmt_span,
+                    "this `reserve` no longer makes sense in rustc version >= 1.62",
+                );
+            }
+        }
+    }
+
+    extract_msrv_attr!(LateContext);
+}
+
+#[must_use]
+fn acceptable_type(cx: &LateContext<'_>, struct_calling_on: &rustc_hir::Expr<'_>) -> bool {
+    let acceptable_types = [sym::Vec, sym::VecDeque];
+    acceptable_types.iter().any(|&acceptable_ty| {
+        match cx.typeck_results().expr_ty(struct_calling_on).peel_refs().kind() {
+            ty::Adt(def, _) => cx.tcx.is_diagnostic_item(acceptable_ty, def.did()),
+            _ => false,
+        }
+    })
+}
+
+#[must_use]
+fn check_extend_method(
+    cx: &LateContext<'_>,
+    block: &Block<'_>,
+    idx: usize,
+    struct_expr: &rustc_hir::Expr<'_>,
+) -> Option<rustc_span::Span> {
+    let mut read_found = false;
+    let next_stmt_span;
+
+    let mut visitor = expr_visitor_no_bodies(|expr| {
+        if let ExprKind::MethodCall(_, [struct_calling_on, _], _) = expr.kind
+            && let Some(expr_def_id) = cx.typeck_results().type_dependent_def_id(expr.hir_id)
+            && match_def_path(cx, expr_def_id, &paths::ITER_EXTEND)
+            && acceptable_type(cx, struct_calling_on)
+            && equal_ident(struct_calling_on, struct_expr)
+        {
+            read_found = true;
+        }
+        !read_found
+    });
+
+    if idx == block.stmts.len() - 1 {
+        if let Some(e) = block.expr {
+            visitor.visit_expr(e);
+            next_stmt_span = e.span;
+        } else {
+            return None;
+        }
+    } else {
+        let next_stmt = &block.stmts[idx + 1];
+        visitor.visit_stmt(next_stmt);
+        next_stmt_span = next_stmt.span;
+    }
+    drop(visitor);
+
+    if read_found {
+        return Some(next_stmt_span);
+    }
+
+    None
+}
+
+#[must_use]
+fn equal_ident(left: &rustc_hir::Expr<'_>, right: &rustc_hir::Expr<'_>) -> bool {
+    fn ident_name(expr: &rustc_hir::Expr<'_>) -> Option<rustc_span::Symbol> {
+        if let ExprKind::Path(QPath::Resolved(None, inner_path)) = expr.kind
+            && let [inner_seg] = inner_path.segments {
+            return Some(inner_seg.ident.name);
+        }
+        None
+    }
+
+    ident_name(left) == ident_name(right)
+}
--- a/clippy_lints/src/unnecessary_reserve.rs
+++ b/clippy_lints/src/unnecessary_reserve.rs
@@ -65,7 +65,7 @@ impl<'tcx> LateLintPass<'tcx> for UnnecessaryReserve {
                     cx,
                     UNNECESSARY_RESERVE,
                     next_stmt_span,
-                    "this `reserve` no longer makes sense in rustc version >= 1.62",
+                    "unnecessary call to `reserve`",
                     |diag| {
                         diag.span_suggestion(
                             semi_expr.span,
--- a/clippy_lints/src/unnecessary_reserve.rs
+++ b/clippy_lints/src/unnecessary_reserve.rs
@@ -11,7 +11,7 @@ declare_clippy_lint! {
     /// ### What it does
     /// This lint checks for a call to `reserve` before `extend` on a `Vec` or `VecDeque`.
     /// ### Why is this bad?
-    /// vec::reserve method before vec::extend is no longer makes sense in rustc version >= 1.62
+    /// Since Rust 1.62, `extend` implicitly calls `reserve`
     /// ### Example
     /// ```rust
     /// let mut vec: Vec<usize> = vec![];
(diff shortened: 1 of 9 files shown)


clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,141 @@
...
+use rustc_hir::{intravisit::Visitor, Block, ExprKind, QPath, StmtKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty;
+use rustc_semver::RustcVersion;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint checks for `reserve` before calling the `extend` method.
```
Serial-ATA, line 12:
```suggestion
    /// This lint checks for a call to `reserve` before `extend` on a `Vec` or `VecDeque`.
```

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// Use instead:
+    /// ```rust
+    /// let mut vec: Vec<usize> = vec![];
+    /// let array: &[usize] = &[1, 2];
+    /// vec.extend(array);
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub UNNECESSARY_RESERVE,
+    pedantic,
+    "`reserve` method before `extend` is no longer makes sense in rustc version >= 1.62"
```
Serial-ATA, line 31:
```suggestion
    "`reserve` before `extend` no longer makes sense in rustc version >= 1.62"
```

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,141 @@
...
+use rustc_middle::ty;
+use rustc_semver::RustcVersion;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint checks for `reserve` before calling the `extend` method.
+    /// ### Why is this bad?
+    /// vec::reserve method before vec::extend is no longer makes sense in rustc version >= 1.62
```
Serial-ATA, line 14:
```suggestion
    /// Since Rust 1.62, `extend` implicitly calls `reserve`
```

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,141 @@
...
+                    match_def_path(cx, expr_def_id, &paths::VEC_DEQUE_RESERVE))
+                && acceptable_type(cx, struct_calling_on)
+                && let Some(next_stmt_span) = check_extend_method(cx, block, idx, struct_calling_on)
+                && !next_stmt_span.from_expansion()
+            {
+                span_lint(
+                    cx,
+                    UNNECESSARY_RESERVE,
+                    next_stmt_span,
+                    "this `reserve` no longer makes sense in rustc version >= 1.62",
```
Serial-ATA, line 67: This message is shown on the `extend` call.

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,141 @@
...
+            if let StmtKind::Semi(semi_expr) = stmt.kind
+                && let ExprKind::MethodCall(_, [struct_calling_on, _], _) = semi_expr.kind
+                && let Some(expr_def_id) = cx.typeck_results().type_dependent_def_id(semi_expr.hir_id)
+                && (match_def_path(cx, expr_def_id, &paths::VEC_RESERVE) ||
+                    match_def_path(cx, expr_def_id, &paths::VEC_DEQUE_RESERVE))
+                && acceptable_type(cx, struct_calling_on)
+                && let Some(next_stmt_span) = check_extend_method(cx, block, idx, struct_calling_on)
+                && !next_stmt_span.from_expansion()
+            {
+                span_lint(
```
Serial-ATA, line 63: Couldn't this have a `MaybeIncorrect` suggestion?

Serial-ATA (changes requested):

kyoto7250:
@Serial-ATA 

Thanks for your review 👍🏻
I will check and fix.

clippy_lints/src/unnecessary_reserve.rs
kyoto7250: I had missed it, Thanks!

clippy_lints/src/unnecessary_reserve.rs
kyoto7250: I fixed in https://github.com/rust-lang/rust-clippy/pull/9073/commits/2c4bca26889eae4324a7d241aa6b824b4517d3bc

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,150 @@
...
+    /// Use instead:
+    /// ```rust
+    /// let mut vec: Vec<usize> = vec![];
+    /// let array: &[usize] = &[1, 2];
+    /// vec.extend(array);
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub UNNECESSARY_RESERVE,
+    pedantic,
+    "`reserve` before `extend` no longer makes sense in rustc version >= 1.62"
```
Serial-ATA, line 32:
```suggestion
    "calling `reserve` before `extend` on a `Vec` or `VecDeque`, when it will be called implicitly"
```
This should explain why it no longer makes sense. I think this works.

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,150 @@
...
+                    match_def_path(cx, expr_def_id, &paths::VEC_DEQUE_RESERVE))
+                && acceptable_type(cx, struct_calling_on)
+                && let Some(next_stmt_span) = check_extend_method(cx, block, idx, struct_calling_on)
+                && !next_stmt_span.from_expansion()
+            {
+                span_lint_and_then(
+                    cx,
+                    UNNECESSARY_RESERVE,
+                    next_stmt_span,
+                    "this `reserve` no longer makes sense in rustc version >= 1.62",
```
Serial-ATA, line 68:
```suggestion
                    "unnecessary call to `reserve`",
```

There should be a note explaining why the call is unnecessary as well, similar to the above comment.

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,150 @@
...
+    None
+}
+
+#[must_use]
+fn equal_ident(left: &rustc_hir::Expr<'_>, right: &rustc_hir::Expr<'_>) -> bool {
+    fn ident_name(expr: &rustc_hir::Expr<'_>) -> Option<rustc_span::Symbol> {
+        if let ExprKind::Path(QPath::Resolved(None, inner_path)) = expr.kind
+            && let [inner_seg] = inner_path.segments {
+            return Some(inner_seg.ident.name);
+        }
```
Serial-ATA, line 145:
```suggestion
            && let [inner_seg] = inner_path.segments
        {
            return Some(inner_seg.ident.name);
        }
```

Formatting nit

Serial-ATA (changes requested):

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,151 @@
...
+use rustc_semver::RustcVersion;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint checks for a call to `reserve` before `extend` on a `Vec` or `VecDeque`.
+    /// ### Why is this bad?
+    /// Since Rust 1.62, `extend` implicitly calls `reserve`
+    /// ### Example
```
flip1995, line 16: NIT: readability

```suggestion
    /// This lint checks for a call to `reserve` before `extend` on a `Vec` or `VecDeque`.
    ///
    /// ### Why is this bad?
    /// Since Rust 1.62, `extend` implicitly calls `reserve`
    ///
    /// ### Example
```

clippy_utils/src/msrvs.rs
```diff
@@ -12,7 +12,7 @@ macro_rules! msrv_aliases {
 
 // names may refer to stabilized feature flags or library items
 msrv_aliases! {
-    1,62,0 { BOOL_THEN_SOME }
+    1,62,0 { BOOL_THEN_SOME, UNNECESSARY_RESERVE }
```
flip1995, line 15: This should be named after the feature that was introduced rather than after the lint.

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn check_block(&mut self, cx: &LateContext<'tcx>, block: &Block<'tcx>) {
+        if !meets_msrv(self.msrv, msrvs::UNNECESSARY_RESERVE) {
+            return;
+        }
+
+        for (idx, stmt) in block.stmts.iter().enumerate() {
+            if let StmtKind::Semi(semi_expr) = stmt.kind
+                && let ExprKind::MethodCall(_, [struct_calling_on, _], _) = semi_expr.kind
+                && let Some(expr_def_id) = cx.typeck_results().type_dependent_def_id(semi_expr.hir_id)
+                && (match_def_path(cx, expr_def_id, &paths::VEC_RESERVE) ||
```
flip1995, line 58: These new `paths` might not be necessary, if you compare the method name and then check the type of the receiver (`struct_calling_on`). The type of the receiver can be checked with `is_ty_diagnostic_item(..., sym::Vec)` and similar for `VecDeque`.

EDIT: Just saw, that you're already doing this in `acceptable_type`. So this path comparison is not really necessary. Just compare the method name.

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,151 @@
...
+            next_stmt_span = e.span;
+        } else {
+            return None;
+        }
+    } else {
+        let next_stmt = &block.stmts[idx + 1];
+        visitor.visit_stmt(next_stmt);
+        next_stmt_span = next_stmt.span;
+    }
+    drop(visitor);
```
flip1995, line 130: Why? :thinking:

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,151 @@
...
+
+#[must_use]
+fn check_extend_method(
+    cx: &LateContext<'_>,
+    block: &Block<'_>,
+    idx: usize,
+    struct_expr: &rustc_hir::Expr<'_>,
+) -> Option<rustc_span::Span> {
+    let mut read_found = false;
+    let next_stmt_span;
```
flip1995, line 104: I would not pre-declare this, but rather use a more Rust like style below with

```rust
let next_stmt_span = if idx == block.stmts.len() - 1 {
```

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    if read_found {
+        return Some(next_stmt_span);
+    }
+
+    None
+}
+
+#[must_use]
+fn equal_ident(left: &rustc_hir::Expr<'_>, right: &rustc_hir::Expr<'_>) -> bool {
```
flip1995, line 140: I'm pretty sure, you can use our `SpanlessEq` utility for this.

tests/ui/unnecessary_reserve.rs
```diff
@@ -0,0 +1,95 @@
...
+    msrv_1_62();
+}
+
+fn vec_reserve() {
+    let mut vec: Vec<usize> = vec![];
+    let array: &[usize] = &[1, 2];
+
+    // do lint
+    vec.reserve(1);
+    vec.extend([1]);
```
flip1995, line 20: I'm unsure if this should lint. I think this should only lint for

```rust
vec.reserve(expr.len());
vec.extend(expr);
```

where `expr` is just any arbitrary expression where `len` is called on in the `reserve` call and then used in the `extend` call. To check if the expression is the same, you can use `SpanlessEq`.

Or am I missing something?

tests/ui/unnecessary_reserve.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::collections::HashMap;
+use std::collections::VecDeque;
+
+fn main() {
+    vec_reserve();
+    vec_deque_reserve();
+    hash_map_reserve();
+    msrv_1_62();
+}
+
```
flip1995, line 13: NIT: This should be at the end of the test file.

tests/ui/unnecessary_reserve.stderr
```diff
@@ -0,0 +1,52 @@
...
+LL |     vec.reserve(array.len());
+   |     ------------------------ help: remove this line
+LL |     vec.extend(array);
+   |     ^^^^^^^^^^^^^^^^^^
+
+error: unnecessary call to `reserve`
+  --> $DIR/unnecessary_reserve.rs:29:9
+   |
+LL |         vec.reserve(1);
+   |         -------------- help: remove this line
```
flip1995, line 23: I think the span of the suggestion is wrong? Shouldn't it include the `;` as well?

clippy_lints/src/unnecessary_reserve.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[must_use]
+    pub fn new(msrv: Option<RustcVersion>) -> Self {
+        Self { msrv }
+    }
+}
+
+impl_lint_pass!(UnnecessaryReserve => [UNNECESSARY_RESERVE]);
+
+impl<'tcx> LateLintPass<'tcx> for UnnecessaryReserve {
+    fn check_block(&mut self, cx: &LateContext<'tcx>, block: &Block<'tcx>) {
```
flip1995, line 49: I'm pretty sure you don't have to check the whole block, but rather just the expression and if you find a `reserve` call, you can use [`get_enclosing_block`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.get_enclosing_block.html), find the next expression/statement in the block and then just check that. I even doubt that you need a visitor for this.

flip1995 (changes requested):
I think this lint would be better placed in the `methods` module, since you check for a specific method call. That way you also don't have to implement all of the method matching logic.

clippy_lints/src/unnecessary_reserve.rs
Serial-ATA: `read_found` was used in the closure, and need to be checked.

flip1995: Ah ok. I think the refactor in #8762 will make this unnecessary. Let's see what will get merged first.

xFrednet:
Hey @kyoto7250 this is a ping from triage. Are there any updates on this PR? If you're stuck or need assistance, you're always welcome to reach out.

kyoto7250:
@xFrednet 
Sorry, currently I am too busy, so I cannot maintain this PR.
You may close this PR.

I wan to re-open and respond when I have more time.

xFrednet:
That's totally fine, thank you for the swift response and work, you already put into this!

PR #9082, opened 2022-07-01
Fix direct `#[allow]` attributes in `let_unit_value`

### LINKED ISSUE #9080: let_unit_value when using constants to limit values of `const usize`
### Summary

The pattern `let _ = CONSTANT_VALUE` is used to ensure that various consts conform to expected values. For example, I have a const that I need to be aligned to 4096 bytes.

This check is done by using the `CONSTANT_VALUE` during some const constructor and throwing away the result. However, Clippy reports that this code is unused, and needs to be discarded.

Furthermore, clippy appears to ignore `#[allow(clippy::let_unit_value)]`.

### Lint Name

let_unit_value

### Reproducer

I tried this code:

```rust
#[repr(C, align(4096))]
pub struct Buffer<const N: usize = 4096> {
    data: [u8; N],
}

impl<const N: usize> Buffer<N> {
    // Ensure that `N` is a multiple of 4096
    const CHECK_ALIGNED: () = if N & 4095 != 0 {
        panic!("Buffer size must be a multiple of 4096")
    };

    pub const fn new() -> Self {
        #[allow(clippy::let_unit_value)]
        let _ = Self::CHECK_ALIGNED;
        Buffer { data: [0u8; N], }
    }
}

fn main() {
    let buf1 = Buffer::<4096>::new();
    // let buf2 = Buffer::<4097>::new();
    let buf3 = Buffer::<8192>::new();
    println!("Size of buf1: {}", core::mem::size_of_val(&buf1));
    println!("Size of buf3: {}", core::mem::size_of_val(&buf3));
}
```

I saw this happen:

```
warning: this let-binding has unit value
  --> src/main.rs:14:9
   |
14 |         let _ = Self::CHECK_ALIGNED;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: omit the `let` binding: `Self::CHECK_ALIGNED;`
   |
   = note: `#[warn(clippy::let_unit_value)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value
```

No warning

### Version

```text
binary: rustc
commit-date: 2022-06-27
host: x86_64-pc-windows-msvc
release: 1.62.0
LLVM version: 14.0.5
```

### Additional Labels

_No response_

Alexendoo:
Fixes part of #9080

Not sure why it doesn't work when the lint is emitted at the statement, but switching it to the local works fine

changelog: Fix direct `#[allow]` attributes in [`let_unit_value`]

### DIFF
--- a/clippy_lints/src/unit_types/let_unit_value.rs
+++ b/clippy_lints/src/unit_types/let_unit_value.rs
@@ -4,18 +4,17 @@ use clippy_utils::visitors::for_each_value_source;
 use core::ops::ControlFlow;
 use rustc_errors::Applicability;
 use rustc_hir::def::{DefKind, Res};
-use rustc_hir::{Expr, ExprKind, PatKind, Stmt, StmtKind};
+use rustc_hir::{Expr, ExprKind, Local, PatKind};
 use rustc_lint::{LateContext, LintContext};
 use rustc_middle::lint::in_external_macro;
 use rustc_middle::ty::{self, Ty, TypeFoldable, TypeSuperFoldable, TypeVisitor};
 
 use super::LET_UNIT_VALUE;
 
-pub(super) fn check(cx: &LateContext<'_>, stmt: &Stmt<'_>) {
-    if let StmtKind::Local(local) = stmt.kind
-        && let Some(init) = local.init
+pub(super) fn check(cx: &LateContext<'_>, local: &Local<'_>) {
+    if let Some(init) = local.init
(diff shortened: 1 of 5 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the fix.

Sir bors, I declare this PR as r+ed. Please test it and merge it, if it's green :upside_down_face:

PR #9091, opened 2022-07-02
chore: Set permissions for GitHub actions

nathannaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/clippy_bors.yml
+++ b/.github/workflows/clippy_bors.yml
@@ -261,6 +261,8 @@ jobs:
   # ALL THE PREVIOUS JOBS NEED TO BE ADDED TO THE `needs` SECTION OF THIS JOB!
 
   end-success:
+    permissions:
+      contents: none
     name: bors test finished
     if: github.event.pusher.name == 'bors' && success()
     runs-on: ubuntu-latest
@@ -271,6 +273,8 @@ jobs:
         run: exit 0
 
   end-failure:
+    permissions:
+      contents: none
     name: bors test finished
     if: github.event.pusher.name == 'bors' && (failure() || cancelled())
     runs-on: ubuntu-latest
--- a/.github/workflows/clippy_dev.yml
+++ b/.github/workflows/clippy_dev.yml
@@ -16,6 +16,9 @@ on:
 env:
   RUST_BACKTRACE: 1
 
+permissions:
+  contents: read
+
 jobs:
   clippy_dev:
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


.github/workflows/clippy_bors.yml
```diff
@@ -261,6 +261,8 @@ jobs:
   # ALL THE PREVIOUS JOBS NEED TO BE ADDED TO THE `needs` SECTION OF THIS JOB!
 
   end-success:
+    permissions:
+      contents: none
```
Alexendoo, line 265: Do these permissions apply to jobs in `needs` ?

flip1995:
We only use the `GITHUB_TOKEN` in the changelog step in the `clippy_bors.yaml` workflow. So IIUC, it would be enough to set that job to

```yaml
permissions:
  pull-requests: read
```

And for everything else set the permissions to `none`. WDYT?

Or does this also set permissions for other secrets? I couldn't find that  mentioned in the documentation, so I assume that it doesn't.

flip1995:
> even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Assuming they succeed, couldn't they also just set the permissions to `write-all` then? This might be an ignorant question.

flip1995:
Oh wait, we also use it for the rust-lang action to cancel workflows, when other workflows are canceled. So those probably need 

```yaml
permissions:
  actions: read
```

nathannaveen:
I don't have time at the moment, so could you please cherry-pick and merge the commit?

flip1995:
Waiting for a response for my above questions. @nathannaveen don't worry if it takes you a week or two to respond to them!

r? @flip1995

flip1995:
Closing this PR, because neither the author nor I currently have time to deal with it and I don't see great security risks by not doing this. Thanks for the work @nathannaveen. If you have more time for this and want to pick it up again, let me know.

PR #9099, opened 2022-07-02
Extend unnecessary_lazy_eval to cover `bool::then` -> `bool::then_some`

### LINKED ISSUE #9097: `bool::then` with trivial body should use `bool::then_some`
### What it does

With `bool::then_some` stabilized in Rust 1.62, uses of `bool::then` with a closure consisting of a simple body without side effects should become `bool::then_some`.

This should be part of the existing `unnecessary_lazy_evaluations`.

### Lint Name

unnecessary_lazy_evaluations

### Category

style

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
fn f(cond: bool) -> Option<u64> {
    cond.then(|| 42)
}
```

Could be written as:

```rust
fn f(cond: bool) -> Option<u64> {
    cond.then_some(42)
}
```

joshtriplett:
fixes #9097

changelog: Extend `unnecessary_lazy_eval` to convert `bool::then` to `bool::then_some`


joshtriplett:
This is WIP because I haven't managed to get the change to actually fire on the test case; I would appreciate advice on what I'm missing.

clippy_lints/src/methods/unnecessary_lazy_eval.rs
```diff
@@ -20,8 +20,9 @@ pub(super) fn check<'tcx>(
 ) {
     let is_option = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(recv), sym::Option);
     let is_result = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(recv), sym::Result);
+    let is_bool = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(recv), sym::bool);
```
Serial-ATA, line 23:
```suggestion
    let is_bool = cx.typeck_results().expr_ty(recv).is_bool();
```

There's no `bool` diagnostic item :slightly_smiling_face:

mikerite:
This should probably have a MSRV check for 1.62.

PR #9102, opened 2022-07-02
Added the `[unused_box]` lint

### LINKED ISSUE #5: Returning Box<T>
We explicitly say in the pointer guide that you usually shouldn't do this, return T instead and box it.

botahamec:
fixes #5 

I'm not confident in the name of this lint. Let me know if you can think of something better

changelog: Added the ``[`unused_box`]`` lint to warn when the return type of a function is `Box<T>` where `T` implements `Sized`.


dswij:
Thanks for the PR! Sorry I haven't been able to take a look at this. I will get to it asap this week when I have the time

clippy_lints/src/unused_box.rs
```diff
@@ -0,0 +1,79 @@
...
+        cx: &LateContext<'_>,
+        fn_kind: FnKind<'_>,
+        decl: &FnDecl<'_>,
+        _: &Body<'_>,
+        _: Span,
+        _: HirId,
+    ) {
+        // it's unclear what part of a closure you would span, so for now it's ignored
+        // if this is changed, please also make sure not to call `hir_ty_to_ty` below
+        if let FnKind::Closure = fn_kind {
```
dswij, line 49: we can avoid the Yoda condition by using `matches!` for this instead of using `if let`

```suggestion
        if matches!(fn_kind, FnKind::Closure) {
```

This might be a good improvement for `equatable_if_let` lint :thinking:

dswij (commented):
Thanks for the PR and the patience! The lint logic itself looks great. For the naming, I think `unnecessary_box_returns` is slightly more descriptive. What do you think?

I'm actually curious whether this will fire a lot in other rust codebases. I will try to run this lint for some of the most popular crates.

botahamec:
@dswij I like `unnecessary_box_return`. I think an 's' at the end is unnecessary.

clippy_lints/src/unused_box.rs
```diff
@@ -0,0 +1,79 @@
...
+use rustc_typeck::hir_ty_to_ty;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Checks for a return type containing a `Box<T>` where `T` implements `Sized`
+    ///
+    /// ### Why is this bad?
+    ///
+    /// It's better to just return `T` in these cases
```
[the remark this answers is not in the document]
botahamec: I just added some elaboration

clippy_lints/src/unused_box.rs
botahamec: Just made that change

botahamec:
@xFrednet Ah, ok. I'll make that change then

Jarcho:
The lint itself is fine, but it requires placement syntax before it's generally applicable. As far as I know that's completely stalled as of now. `Box::new(foo())` can't optimize out the temporary from calling `foo()` unless it's inlined (and not necessarily then).

As a `restriction` lint the downside is fine. A size limit would also be fine. Just leaving it in the nursery until placement syntax is stabilized would also work.

dswij:
> The lint itself is fine, but it requires placement syntax before it's generally applicable. As far as I know that's completely stalled as of now. `Box::new(foo())` can't optimize out the temporary from calling `foo()` unless it's inlined (and not necessarily then). 

@botahamec I think we can add this to the test case, and I think it's good to be merged once the merge conflict is addressed.

dswij:
So, I ran this lint on 250 crates:

# Summary

## Build failures

Total: 1

- vsdb-0.45.2

## Warnings

Total: 17

 Crate                   | Count
:------------------------|------:
 bincode-1.3.3           |     4
 clap-3.2.15             |     2
 crossbeam-channel-0.5.6 |     1
 crossbeam-epoch-0.9.10  |     1
 miniz_oxide-0.5.3       |     2
 parking_lot_core-0.9.3  |     1
 pest-2.1.3              |     4
 regex-1.6.0             |     1
 structopt-0.3.26        |     1

---

There are some improvements that needs to be made. e.g. these suggestion aren't the most accurate.

```log
---> pest-2.1.3/src/parser_state.rs:1107:57
warning: function returns `Box<parser_state::ParserState<'i, R>>` when `parser_state::ParserState<'i, R>` implements `Sized`
    --> src/parser_state.rs:1107:57
     |
1107 |     pub(crate) fn checkpoint_ok(mut self: Box<Self>) -> Box<Self> {
     |                                                         ^^^^^^^^^ help: change the return type to: `parser_state::ParserState<'i, R>`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_box_returns
```

```log
---> crossbeam-channel-0.5.6/src/flavors/zero.rs:57:27
warning: function returns `Box<flavors::zero::Packet<T>>` when `flavors::zero::Packet<T>` implements `Sized`
  --> src/flavors/zero.rs:57:27
   |
57 |     fn empty_on_heap() -> Box<Packet<T>> {
   |                           ^^^^^^^^^^^^^^ help: change the return type to: `flavors::zero::Packet<T>`
   |
   = note: requested on the command line with `-W clippy::unnecessary-box-returns`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_box_returns
```

```log
---> structopt-0.3.26/src/lib.rs:1218:53
warning: function returns `Box<T>` when `T` implements `Sized`
    --> src/lib.rs:1218:53
     |
1218 |     fn from_clap(matches: &clap::ArgMatches<'_>) -> Self {
     |                                                     ^^^^ help: change the return type to: `T`
     |
     = note: requested on the command line with `-W clippy::unnecessary-box-returns`
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_box_returns
```

We can address this in this PR or later as an improvement since this lint is going into `nursery`. What do you think? @botahamec

botahamec:
@dswij I think it at least makes sense to take a look at the third one. I'd honestly be ok with merging on the first two examples. Let me know if there's anything else I should look at though. I don't have my desktop with me at the moment, so it might take a while.

dswij:
No rush. Feel free to ping me when it's ready

dswij:
hey @botahamec, sorry to nag, but it seems like there's not a lot of activity on this PR lately. Let me know if you don't have much time, and I might be able to continue the work on this PR.

botahamec:
@dswij Oh yeah, feel free.

dswij:
r? @xFrednet 

I've rebased this and update some outdated stuff. This should be ready to go, any improvements can be in future PRs since this is going to nursery. 

Would you mind to help take a look if you have the time?

xFrednet:
Sure, I'll give you a review soon! :)

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,81 @@
...
+        if implements_trait(cx, boxed_ty, sized_trait, &[]) {
+            span_lint_and_sugg(
+                cx,
+                UNNECESSARY_BOX_RETURNS,
+                return_ty_hir.span,
+                format!("function returns `Box<{boxed_ty}>` when `{boxed_ty}` implements `Sized`").as_str(),
+                "change the return type to",
+                boxed_ty.to_string(),
+                // the return value also needs to be changed, so this can't be MachineApplicable
+                Applicability::HasPlaceholders,
```
xFrednet, line 77: This will also propagate to function callers. Therefore, it should be unspecified IMO
```suggestion
                Applicability::Unspecified,
```

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub UNNECESSARY_BOX_RETURNS,
+    nursery,
+    "Needlessly returning a Box"
+}
+declare_lint_pass!(UnnecessaryBoxReturns => [UNNECESSARY_BOX_RETURNS]);
+
+impl LateLintPass<'_> for UnnecessaryBoxReturns {
+    fn check_fn(
```
xFrednet, line 40: I believe this function is also called for functions inside `impl <Trait> for` could you add a test for that. Trait implementations shouldn't trigger the lint, if possible the trait declaration should. This might require `check_trait_item` to then iterate over the members.

clippy_lints/src/lib.rs
```diff
@@ -908,6 +909,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|_| Box::new(fn_null_check::FnNullCheck));
     store.register_late_pass(|_| Box::new(permissions_set_readonly_false::PermissionsSetReadonlyFalse));
     store.register_late_pass(|_| Box::new(size_of_ref::SizeOfRef));
+    store.register_late_pass(|_| Box::new(unnecessary_box_returns::UnnecessaryBoxReturns));
```
xFrednet, line 912: This lint suggests changing the signature of functions. Therefore, it should take the config value `avoid-breaking-exported-api` into account.

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn foo() -> String {
+    ///     String::from("Hello, world!")
+    /// }
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub UNNECESSARY_BOX_RETURNS,
+    nursery,
```
xFrednet, line 34: I believe we can directly put it into a proper lint group. We might want to ask lintcheck before, to see if it's too noisy for warn-by-default.

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,81 @@
...
+        let Some(sized_trait) = cx.tcx.lang_items().sized_trait() else { return };
+
+        // it's sometimes useful to return Box<T> if T is unsized, so don't lint those
+        if implements_trait(cx, boxed_ty, sized_trait, &[]) {
+            span_lint_and_sugg(
+                cx,
+                UNNECESSARY_BOX_RETURNS,
+                return_ty_hir.span,
+                format!("function returns `Box<{boxed_ty}>` when `{boxed_ty}` implements `Sized`").as_str(),
+                "change the return type to",
```
xFrednet, line 74: What do you think about this? It makes the message shorter and only prints out the type once. This improves readability for long types :)
```suggestion
                &format!("boxed return of the sized type `{boxed_ty}`"),
                "try",
```

xFrednet (commented):
A good start to a new lint. I added some comments, feel free to ask if anything is unclear :)

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,81 @@
...
+    ///     Box::new(String::from("Hello, world!"))
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn foo() -> String {
+    ///     String::from("Hello, world!")
+    /// }
+    /// ```
+    #[clippy::version = "1.64.0"]
```
botahamec, line 32: Just looked through it real quick. I noticed this is still here. I know it won't make too big of a difference for people using it, but for completeness sake, maybe this should be updated?

xFrednet: Updating it would be cool, usually these versions are also checked during the PR writing process :)

xFrednet:
Hey @botahamec, this is a ping from triage. There hasn't been any updates in a long time, are you stuck on an issue or need help with anything?

botahamec:
@xFrednet Oh not at all. I just assumed for a while that @dswij was working on the comments. But the comments seem like things I can resolve if @dswij is ok with that.

xFrednet:
Ahh, sorry, I missed that @dswij took over, thank you! @dswij do you need help with anything? :)

clippy_lints/src/unnecessary_box_returns.rs
botahamec: That makes sense. Done.

clippy_lints/src/lib.rs
botahamec: I believe I fixed this just now. It's a bit weird using `check_item` now to check functions, but I wasn't able to figure out how to do it with `check_fn`. Let me know if I'm missing something.

clippy_lints/src/unnecessary_box_returns.rs
botahamec: I'm guessing this should still be triggered on something like `impl <Struct>`. I wasn't able to figure out how to check for that. Any suggestions?

clippy_lints/src/unnecessary_box_returns.rs
botahamec: I tried setting the category to "perf" and ran `cargo lintcheck` and didn't see anything for this lint, so you might be right. Does "perf" seem like the right category here? Personally, I think before that happens, we should at least look into making this lint dependent on the size of the struct.

clippy_lints/src/unnecessary_box_returns.rs
botahamec: Done

clippy_lints/src/unnecessary_box_returns.rs
xFrednet: `cargo lintcheck` also has a `--filter <lint>` option to ensure all cases are emitted.

For the `perf` group a size check should be required. Even than, I'd be careful to move it directly from `Nursery` to `perf`. I think `pedantic` would be the correct group :)

clippy_lints/src/lib.rs
```diff
@@ -924,6 +925,11 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
         ))
     });
     store.register_late_pass(|_| Box::new(no_mangle_with_rust_abi::NoMangleWithRustAbi));
+    store.register_late_pass(move |_| {
+        Box::new(unnecessary_box_returns::UnnecessaryBoxReturns::new(
+            avoid_breaking_exported_api,
```
xFrednet, line 930: Could you also add the lint to the configuration docs in `clippy_lints::utils::conf`? Running `cargo collect-metadata` should then update the documentation automatically

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,94 @@
...
+impl LateLintPass<'_> for UnnecessaryBoxReturns {
+    fn check_trait_item(&mut self, cx: &LateContext<'_>, item: &TraitItem<'_>) {
+        let TraitItemKind::Fn(signature, _) = &item.kind else { return };
+        self.check_fn_decl(cx, signature.decl, item.owner_id.def_id);
+    }
+
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        let ItemKind::Fn(signature, ..) = &item.kind else { return };
+        self.check_fn_decl(cx, signature.decl, item.owner_id.def_id);
+    }
```
xFrednet, line 93: I see a few ways to check if this is inside an `impl <trait> for <type>`. I guess that the best one would be to get the parent of the `Fn` item and check if it's an `impl` item with a trait. You can take a look at [`cx.tcx.hir().get_parent()`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/hir/map/struct.Map.html#method.get_parent)
 

Although, the uitest already ignores the case `fn baz(&self) -> Box<usize> {` in line 12. I'm not quite sure why, I would expect it to find and trigger on that as well :thinking:

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,94 @@
...
+impl UnnecessaryBoxReturns {
+    pub fn new(avoid_breaking_exported_api: bool) -> Self {
+        Self {
+            avoid_breaking_exported_api,
+        }
+    }
+
+    fn check_fn_decl(&mut self, cx: &LateContext<'_>, decl: &FnDecl<'_>, def_id: LocalDefId) {
+        // we don't want to tell someone to break an exported function if they ask us not to
+        if self.avoid_breaking_exported_api && cx.effective_visibilities.is_exported(def_id) {
```
xFrednet, line 52: Very nice check and I like the comment ^^

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,94 @@
...
+        if !return_ty.is_box() {
+            return;
+        }
+
+        let boxed_ty = return_ty.boxed_ty();
+        let Some(sized_trait) = cx.tcx.lang_items().sized_trait() else { return };
+
+        // it's sometimes useful to return Box<T> if T is unsized, so don't lint those
+        if implements_trait(cx, boxed_ty, sized_trait, &[]) {
+            span_lint_and_sugg(
```
xFrednet, line 70: I think here it would be good to use `span_lint_and_then` to include a help message that the return locations also need to be adjusted.

This is an optional improvement :)

xFrednet (commented):
A few comments, but this version already looks quite good :)

clippy_lints/src/unnecessary_box_returns.rs
botahamec: Are you referring to the return expressions or the calling expressions?

clippy_lints/src/unnecessary_box_returns.rs
botahamec: I think functions in impls are considered `ImplItem`s and not `Item`s. That seems to be mutually exclusive. I got it working using that.

clippy_lints/src/unnecessary_box_returns.rs
botahamec: It's now in pedantic

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        if !return_ty.is_box() {
+            return;
+        }
+
+        let boxed_ty = return_ty.boxed_ty();
+        let Some(sized_trait) = cx.tcx.lang_items().sized_trait() else { return };
+
+        // it's sometimes useful to return Box<T> if T is unsized, so don't lint those
+        if implements_trait(cx, boxed_ty, sized_trait, &[]) {
```
Jarcho, line 69: This should be `boxed_ty.is_sized(cx.tcx, cx.param_env)`.

clippy_lints/src/unnecessary_box_returns.rs
xFrednet: I'm referring to the expression that returns the value from the function. So:

```rs
fn hello(b: bool) -> u32 {
	if b {
		return 0; // This
	}
	6 // This
}
```

If you want, you can include spans of these in the help message, but that could get very verbose. A text message should be sufficient IMO :)

clippy_lints/src/unnecessary_box_returns.rs
botahamec: I tried it and clippy panicked on the `needless_lifetimes` test. I have no idea why.

clippy_lints/src/unnecessary_box_returns.rs
botahamec: I agree with that last statement. How does this look?

clippy_lints/src/unnecessary_box_returns.rs
Jarcho: You can try using `cx.tcx.liberate_late_bound_regions(def_id, cx.tcx.fn_sig(def_id).skip_binder()).output()` to get the return type. `hir_ty_to_ty` is very hackish.

botahamec: That seems to work now

clippy_lints/src/unnecessary_box_returns.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+
+        let FnRetTy::Return(return_ty_hir) = &decl.output else { return };
+
+        let return_ty = cx
+            .tcx
+            .erase_late_bound_regions(cx.tcx.fn_sig(def_id).skip_binder())
+            .output();
+
+        if !return_ty.is_box() {
```
xFrednet, line 62: I've tested the lint with `lintcheck` and it detected a method called `boxed`, the standard library also has a few util methods, with a similar name. I think it would be good, if we check the name of the function and allow the lint, if it has `boxed` in the name. I believe the check could look like this:

```
cx.tcx.hir().expect_item(def_id).ident.name.as_str().contains("boxed")
```

I haven't tested it and the code might not be super beautiful, but it should be a step in the right direction :)

xFrednet (commented):
This version looks good to me, thank you for finishing the last todos. I've checked the lint with our lintcheck tool and found one small improvement, which should be simple.

For the merge, would you mind squashing the commits? Having 1 to 5 commits would be ideal for the size of this PR :)

xFrednet:
Hey @botahamec, since you pushed since my last review, have you seen the comments I left? Besides those two I think this is ready for merge :). Your welcome to reach out if you need help with the rebases or have any other questions :)

botahamec:
@xFrednet Yeah. I was hoping to get it done last week, but I never got around to it. I don't have my computer this week, so hopefully I'll finish it next week.

clippy_lints/src/unnecessary_box_returns.rs
botahamec: While I'm waiting to get back to my apartment, I guess I'll mention this. I don't have any strong opinions on this idea, and neither did anyone I mentioned this to, but someone pointed out that this since this is going into pedantic, it might be ok that this function gets linted. I'm ok either way, but I wanted to know what your thought on that was?

Anyway, when I do get the chance, I think it would make more sense to see if it contains "box" than "boxed".

xFrednet:
Alright, no worries, just wanted to make sure you're not stuck on anything :). Enjoy your week!

clippy_lints/src/unnecessary_box_returns.rs
botahamec: Just did this

botahamec:
Alright. I'm done squashing. That was surprisingly convoluted, but it's down to just four commits now.

xFrednet (approved):

Issue #629, opened 2016-02-06
lint creating a &str (or String) from invalid UTF-8 bytes


llogiq:
This is also undefined behavior (according to the rust book), calling `std::str::from_utf8_unchecked(..)` on something which is not UTF-8.

Manishearth:
I don't see why someone would ever use from_unchecked on a literal?

llogiq:
Perhaps within a macro?

John-Steidley:
Perhaps because of a refactor. Perhaps because they passed the literal into a function.

PR #9105, opened 2022-07-03
Add `invalid_utf8_in_unchecked`

Serial-ATA:
changelog: Add [`invalid_utf8_in_unchecked`]
closes: #629

Don't know how useful of a lint this is, just saw this was a really old issue :smile:.

### DIFF
--- a/clippy_lints/src/invalid_utf8_in_unchecked.rs
+++ b/clippy_lints/src/invalid_utf8_in_unchecked.rs
@@ -1,0 +1,74 @@
+use clippy_utils::diagnostics::span_lint;
+use clippy_utils::{match_function_call, paths};
+use rustc_ast::{BorrowKind, LitKind};
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::source_map::Spanned;
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `std::str::from_utf8_unchecked` with an invalid UTF-8 literal
+    ///
+    /// ### Why is this bad?
+    /// Creating such a `str` would result in undefined behavior
+    ///
+    /// ### Example
+    /// ```rust
+    /// # #[allow(unused)]
+    /// unsafe {
+    ///     std::str::from_utf8_unchecked(b"cl\x82ippy");
+    /// }
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub INVALID_UTF8_IN_UNCHECKED,
+    correctness,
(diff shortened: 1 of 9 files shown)


clippy_lints/src/invalid_utf8_in_unchecked.rs
```diff
@@ -0,0 +1,72 @@
...
+    #[clippy::version = "1.64.0"]
+    pub INVALID_UTF8_IN_UNCHECKED,
+    correctness,
+    "using a non UTF-8 literal in `std::std::from_utf8_unchecked`"
+}
+declare_lint_pass!(InvalidUtf8InUnchecked => [INVALID_UTF8_IN_UNCHECKED]);
+
+impl<'tcx> LateLintPass<'tcx> for InvalidUtf8InUnchecked {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let Some(args) = match_function_call(cx, expr, &paths::STR_FROM_UTF8_UNCHECKED) {
```
Jarcho, line 33: Should be `Some([arg])`

clippy_lints/src/invalid_utf8_in_unchecked.rs
```diff
@@ -0,0 +1,72 @@
...
+            match &args[0].kind {
+                ExprKind::Lit(Spanned { node: lit, .. }) => {
+                    if let LitKind::ByteStr(bytes) = &lit
+                        && std::str::from_utf8(bytes).is_err()
+                    {
+                        lint(cx, expr.span);
+                    }
+                },
+                ExprKind::AddrOf(BorrowKind::Ref, _, Expr { kind: ExprKind::Array(args), .. }) => {
+                    let elements = args.iter().filter_map(|e|{
```
Jarcho, line 43: This can result in a false positive with something like `[0b11000001, x]` assuming `x` is a valid tail byte. I can't imagine anyone writing this though.

Jarcho (changes requested):
I agree this is likely useless. Passing a constant to `from_utf8_unchecked` is rather strange as it could have just been a string literal. On the other hand this is UB, so I guess it's justified.

clippy_lints/src/invalid_utf8_in_unchecked.rs
Serial-ATA: Yeah, I don't think it's worth doing any extra work to check for this, though.

Jarcho: It would only require to collecting into an `Option<Vec<_>>` You can just bail out if there's anything that isn't a literal in the array.

Serial-ATA: Ah, misunderstood what you meant :sweat_smile:. Good catch.

Jarcho:
Thank you.

@bors r+

PR #9106, opened 2022-07-03
corrected README.md book link

usefulmove:
This change corrects the book link on the Clippy Book README.md.

changelog: none


flip1995:
Ah thanks, it appears that I missed that lint when moving those files.

@bors r+

PR #9115, opened 2022-07-04
`new_without_default`: ignore const generics/lifetime params on `fn new`

### LINKED ISSUE #9113: Clippy's Default implementation suggestion does not take const generics into account
### Summary

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a41d1a552a1e2c14f1a1eb5262a4f934)

I have a struct with `new` function that has a const generic parameter. Clippy complains and suggests implementing `Default` by calling `new` without any parameters.

The suggestion does not compile because the const generic parameter - no suriprises - can't be inferred.

I believe this is a false positive, because Clippy does not fire the `new_without_default` lint if the `new` function has a parameter and I think it's fair to consider the const generic parameter a parameter of the new function.

### Lint Name

new_without_default

### Reproducer

I tried this code:

```rust
pub struct S;
impl S {
    pub fn new<const FOO:usize>() -> Self {
        S
    }
}
```

I saw this happen:

```
   |
16 | /     pub fn new<const FOO:usize>() -> Self {
17 | |         S
18 | |     }
   | |_____^
   |
   = note: `#[warn(clippy::new_without_default)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default
help: try adding this
   |
15 + impl Default for S {
16 +     fn default() -> Self {
17 +         Self::new()
18 +     }
19 + }
   |
```

I expected to see this happen: I expected the lint to not be triggered.

### Version

```text
Whatever version is behind playground's 1.62.0
```

### Additional Labels

@rustbot label + I-suggestion-causes-error

Alexendoo:
Fixes #9113

No longer lints if `fn new` has any params

changelog: [`new_without_default`]: no longer lints const generics and lifetime params on `fn new`

### DIFF
--- a/clippy_lints/src/new_without_default.rs
+++ b/clippy_lints/src/new_without_default.rs
@@ -88,15 +88,9 @@ impl<'tcx> LateLintPass<'tcx> for NewWithoutDefault {
                             // shouldn't be implemented when it is hidden in docs
                             return;
                         }
-                        if impl_item
-                            .generics
-                            .params
-                            .iter()
-                            .any(|gen| matches!(gen.kind, hir::GenericParamKind::Type { .. }))
-                        {
-                            // when the result of `new()` depends on a type parameter we should not require
-                            // an
-                            // impl of `Default`
+                        if !impl_item.generics.params.is_empty() {
+                            // when the result of `new()` depends on a parameter we should not require
+                            // an impl of `Default`
                             return;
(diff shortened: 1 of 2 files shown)


llogiq:
Thank you!

@bors r+

PR #9121, opened 2022-07-05
Fix stderr for cast_size_32bit

alex-semenyuk:
changelog: fix stderr for cast_size_32bit


flip1995:
Wait, this shouldn't be necessary. If it were necessary, master would be broken. Why does this pass CI? :thinking:

flip1995:
Counting the errors by hand, there are indeed only 18 errors. Why doesn't this fail compiletest? [Wat?](https://www.destroyallsoftware.com/talks/wat)

Anyway @bors r+

Issue #8903, opened 2022-05-27
Arithmetic lint


c410-f3r:
Are you guys open to include an allow-by-default restricted lint that will warn `+`/`-`/`/`/`*` regardless of the involved type? For example, `num-bigint` (among others) has arithmetic implementations that aren't triggered by `clippy::integer_arithmetic`.

If yes, then I will create a PR. If not, then feel free to close this issue.

llogiq:
This could be a restriction lint, and we may even want to make types to block arithmetic on configurable.

c410-f3r:
Like @llogiq said, types should be configurable. Built-in types that don't panic would then be listed as allowed somehow.

I was thinking about creating an `arithmetic` lint that could suppress `integer_arithmetic` in the long run.

PR #9130, opened 2022-07-07
Add `Arithmetic` lint

c410-f3r:
Fixes https://github.com/rust-lang/rust-clippy/issues/8903

r? @llogiq 

changelog: Add `Arithmetic` lint

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -548,6 +548,8 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
         store.register_late_pass(|| Box::new(utils::internal_lints::MsrvAttrImpl));
     }
 
+    let arithmetic_allowed = conf.arithmetic_allowed.clone();
+    store.register_late_pass(move || Box::new(operators::arithmetic::Arithmetic::new(arithmetic_allowed.clone())));
     store.register_late_pass(|| Box::new(utils::dump_hir::DumpHir));
     store.register_late_pass(|| Box::new(utils::author::Author));
     let await_holding_invalid_types = conf.await_holding_invalid_types.clone();
--- a/tests/ui/arithmetic.fixed
+++ b/tests/ui/arithmetic.fixed
@@ -1,0 +1,27 @@
+// run-rustfix
+
+#![allow(clippy::unnecessary_owned_empty_strings)]
+#![feature(saturating_int_impl)]
+#![warn(clippy::arithmetic)]
+
+use core::num::{Saturating, Wrapping};
+
+pub fn hard_coded_allowed() {
+    let _ = Saturating(0u32) + Saturating(0u32);
+    let _ = String::new() + "";
+    let _ = Wrapping(0u32) + Wrapping(0u32);
+
+    let saturating: Saturating<u32> = Saturating(0u32);
+    let string: String = String::new();
+    let wrapping: Wrapping<u32> = Wrapping(0u32);
+
+    let inferred_saturating = saturating + saturating;
+    let inferred_string = string + "";
+    let inferred_wrapping = wrapping + wrapping;
+
+    let _ = inferred_saturating + inferred_saturating;
+    let _ = inferred_string + "";
+    let _ = inferred_wrapping + inferred_wrapping;
+}
+
+fn main() {}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3437,6 +3437,7 @@ Released 2018-09-13
 [`almost_complete_letter_range`]: https://rust-lang.github.io/rust-clippy/master/index.html#almost_complete_letter_range
 [`almost_swapped`]: https://rust-lang.github.io/rust-clippy/master/index.html#almost_swapped
 [`approx_constant`]: https://rust-lang.github.io/rust-clippy/master/index.html#approx_constant
+[`arithmetic`]: https://rust-lang.github.io/rust-clippy/master/index.html#arithmetic
 [`as_conversions`]: https://rust-lang.github.io/rust-clippy/master/index.html#as_conversions
 [`as_underscore`]: https://rust-lang.github.io/rust-clippy/master/index.html#as_underscore
 [`assertions_on_constants`]: https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_constants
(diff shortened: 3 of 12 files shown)


tests/ui/arithmetic.fixed
```diff
@@ -0,0 +1,23 @@
...
+    let saturating: Saturating<u32> = Saturating(0u32);
+    let wrapping: Wrapping<u32> = Wrapping(0u32);
+    let string: String = String::new();
+
+    let _ = saturating + saturating;
+    let _ = wrapping + wrapping;
+    let _ = string + "";
+}
+
+fn main() {}
```
c410-f3r, line 23: Not sure why this file is necessary

tests/ui/integer_arithmetic.stderr
```diff
@@ -12,158 +12,182 @@ error: this operation will panic at runtime
...
+   |     ^^^^^^
+
+error: arithmetic detected
+  --> $DIR/integer_arithmetic.rs:30:5
+   |
+LL |     i /= 0;
+   |     ^^^^^^
+
+error: arithmetic detected
+  --> $DIR/integer_arithmetic.rs:31:5
```
c410-f3r, line 91: Replacing `#![warn(clippy::integer_arithmetic, clippy::float_arithmetic)]` with `#![warn(clippy::arithmetic)]` actually improved diagnostics.

If you guys want to deprecate `integer_arithmetic` and `float_arithmetic` in favour of `arithmetic`, then I can scan for primitives to recover the lost "integer/float arithmetic detected" messages. Although that would imply more processing and such thing doesn't seem too useful at all.

rust-toolchain
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "nightly-2022-06-30"
-components = ["cargo", "llvm-tools-preview", "rust-src", "rust-std", "rustc", "rustc-dev", "rustfmt"]
+channel = "nightly-2022-06-29"
+components = ["llvm-tools-preview", "rust-src", "rust-std", "rustc", "rustc-dev", "rustfmt"]
```
c410-f3r: Please ignore this change. Local stuff...

clippy_lints/src/lib.rs
```diff
@@ -547,6 +548,8 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
         store.register_late_pass(|| Box::new(utils::internal_lints::MsrvAttrImpl));
     }
 
+    let arithmetic_allowed = conf.arithmetic_allowed.clone();
+    store.register_late_pass(move || Box::new(arithmetic::Arithmetic::new(arithmetic_allowed.clone())));
```
c410-f3r, line 552: Not sure how to avoid double clones

tests/ui/arithmetic.fixed
c410-f3r, line 23: Not sure why this file is necessary

clippy_lints/src/arithmetic.rs
```diff
@@ -0,0 +1,143 @@
...
+                } else if let hir::QPath::TypeRelative(ref ty, _) = qpath
+                    && let hir::TyKind::Path(hir::QPath::Resolved(_, ref path)) = ty.kind
+                {
+                    path
+                }
+                else {
+                    return false;
+                };
+                for segment in path.segments {
+                    if self.allowed.contains(segment.ident.as_str()) {
```
c410-f3r, line 64: `Ident::as_str` actually does some processing. Maybe use `FxHashSet<Ident>` instead?

clippy_lints/src/arithmetic.rs
```diff
@@ -0,0 +1,143 @@
...
+                        path
+                    }
+                } else if let hir::QPath::TypeRelative(ref ty, _) = qpath
+                    && let hir::TyKind::Path(hir::QPath::Resolved(_, ref path)) = ty.kind
+                {
+                    path
+                }
+                else {
+                    return false;
+                };
```
c410-f3r, line 62: Only catches

```rust
// Declared types
let local: Wrapping<u32> = Wrapping::new(0);
local + local

// Constructors
Wrapping::new(0) + Wrapping::new(0)
```

I.e., locals without explicit types aren't evaluated. Does `clippy_utils` has an utility to iterate a path in order to retrieved its inferred type?

```rust
let n = Wrapping::new(0);
let another = n;
another + another
```

tests/ui/integer_arithmetic.stderr
Jarcho: `integer_arithmetic` specifically checks for uses which may panic. `float_arithmetic` could maybe be done, but that can be a later problem.

clippy_lints/src/arithmetic.rs
llogiq: At least pull the `as_str()` out of the loop.

Jarcho:
Just a general comment about the approach here. This would be better done using resolved items rather than paths. `LateContext::qpath_res` will resolve any `QPath` if you need it. Generally you should be able to use `cx.typeck_results().expr_ty(_)` to get the type of an `Expr`. You can also look at the `disallowed_types` lint for how it deals with configuring which types to block.

c410-f3r:
> Just a general comment about the approach here. This would be better done using resolved items rather than paths. `LateContext::qpath_res` will resolve any `QPath` if you need it. Generally you should be able to use `cx.typeck_results().expr_ty(_)` to get the type of an `Expr`. You can also look at the `disallowed_types` lint for how it deals with configuring which types to block.

Exactly what I was looking for! Thanks!

c410-f3r:
I think the code is now in a good shape for review.

tests/ui/integer_arithmetic.stderr
c410-f3r: `i /= 0` and `i %= 0` will panic when the compiler is not able to deduct the RHS at compile-time so why the lint currently omits them?

Regardless, I reverted these changes for now but they can be reviewed in the future.

Jarcho: Both `/= 0` and `%= 0` are caught by the `unconditional_panic` lint.

c410-f3r:
@llogiq Perhaps another reviewer?

llogiq:
No, I just had private stuff keeping me from reviewing. Sorry for the delay. I agree, the code is in good shape.

@bors r+

c410-f3r:
> No, I just had private stuff keeping me from reviewing. Sorry for the delay. I agree, the code is in good shape.
> 
> @bors r+

Thank you @llogiq

PR #9134, opened 2022-07-07
Improve `while_let_on_iterator` suggestion inside an `FnOnce` closure

Jarcho:
changelog: Improve `while_let_on_iterator` suggestion inside an `FnOnce` closure

### DIFF
--- a/tests/ui/while_let_on_iterator.rs
+++ b/tests/ui/while_let_on_iterator.rs
@@ -7,7 +7,8 @@
     unused_mut,
     dead_code,
     clippy::equatable_if_let,
-    clippy::manual_find
+    clippy::manual_find,
+    clippy::redundant_closure_call
 )]
 
 fn base() {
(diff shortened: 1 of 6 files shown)


tests/ui/while_let_on_iterator.rs
```diff
@@ -259,7 +260,7 @@ fn issue1924() {
         fn f(&mut self) -> Option<u32> {
             // Used as a field.
             while let Some(i) = self.0.next() {
-                if i < 3 || i > 7 {
```
dswij, line 262: Why are we changing this one?

dswij (commented):
Thanks for the PR! I think overall it looks great. Just one question on the test

tests/ui/while_let_on_iterator.rs
Jarcho: It was triggering `manual_range_contains`. Since the change doesn't affect what the test is checking for I just applied the suggestion.

PR #9136, opened 2022-07-08
Enhance `needless_borrow` to consider trait implementations

smoelius:
The proposed enhancement causes `needless_borrow` to suggest removing `&` from `&e` when `&e` is an argument position requiring trait implementations, and `e` implements the required traits. Example:
```
error: the borrowed expression implements the required traits
  --> $DIR/needless_borrow.rs:131:51
   |
LL |     let _ = std::process::Command::new("ls").args(&["-a", "-l"]).status().unwrap();
   |                                                   ^^^^^^^^^^^^^ help: change this to: `["-a", "-l"]`
```

r? @Jarcho

changelog: Enhance `needless_borrow` to consider trait implementations

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -1,24 +1,31 @@
 use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_hir_and_then};
 use clippy_utils::source::{snippet_with_applicability, snippet_with_context};
 use clippy_utils::sugg::has_enclosing_paren;
-use clippy_utils::ty::{expr_sig, peel_mid_ty_refs, ty_sig, variant_of_res};
-use clippy_utils::{get_parent_expr, is_lint_allowed, path_to_local, walk_to_expr_usage};
+use clippy_utils::ty::{contains_ty, expr_sig, is_copy, peel_mid_ty_refs, ty_sig, variant_of_res};
+use clippy_utils::{fn_def_id, get_parent_expr, is_lint_allowed, meets_msrv, msrvs, path_to_local, walk_to_expr_usage};
 use rustc_ast::util::parser::{PREC_POSTFIX, PREC_PREFIX};
 use rustc_data_structures::fx::FxIndexMap;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_ty, Visitor};
 use rustc_hir::{
-    self as hir, BindingAnnotation, Body, BodyId, BorrowKind, Closure, Expr, ExprKind, FnRetTy, GenericArg, HirId,
-    ImplItem, ImplItemKind, Item, ItemKind, Local, MatchSource, Mutability, Node, Pat, PatKind, Path, QPath, TraitItem,
-    TraitItemKind, TyKind, UnOp,
+    self as hir, def_id::DefId, BindingAnnotation, Body, BodyId, BorrowKind, Closure, Expr, ExprKind, FnRetTy,
+    GenericArg, HirId, ImplItem, ImplItemKind, Item, ItemKind, Local, MatchSource, Mutability, Node, Pat, PatKind,
+    Path, QPath, TraitItem, TraitItemKind, TyKind, UnOp,
 };
+use rustc_index::bit_set::BitSet;
 use rustc_infer::infer::TyCtxtInferExt;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::adjustment::{Adjust, Adjustment, AutoBorrow, AutoBorrowMutability};
-use rustc_middle::ty::{self, Binder, BoundVariableKind, List, Ty, TyCtxt, TypeVisitable, TypeckResults};
+use rustc_middle::ty::{
+    self, subst::Subst, Binder, BoundVariableKind, EarlyBinder, FnSig, GenericArgKind, List, ParamTy, PredicateKind,
+    ProjectionPredicate, Ty, TyCtxt, TypeVisitable, TypeckResults,
+};
+use rustc_semver::RustcVersion;
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::{symbol::sym, Span, Symbol, DUMMY_SP};
-use rustc_trait_selection::infer::InferCtxtExt;
+use rustc_trait_selection::infer::InferCtxtExt as _;
+use rustc_trait_selection::traits::{query::evaluate_obligation::InferCtxtExt as _, Obligation, ObligationCause};
+use std::collections::VecDeque;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -151,6 +158,7 @@ pub struct Dereferencing {
     /// been finished. Note we can't lint at the end of every body as they can be nested within each
     /// other.
     current_body: Option<BodyId>,
+
     /// The list of locals currently being checked by the lint.
     /// If the value is `None`, then the binding has been seen as a ref pattern, but is not linted.
     /// This is needed for or patterns where one of the branches can be linted, but another can not
@@ -158,6 +166,19 @@ pub struct Dereferencing {
     ///
     /// e.g. `m!(x) | Foo::Bar(ref x)`
     ref_locals: FxIndexMap<HirId, Option<RefPat>>,
+
+    // `IntoIterator` for arrays requires Rust 1.53.
+    msrv: Option<RustcVersion>,
+}
+
+impl Dereferencing {
+    #[must_use]
+    pub fn new(msrv: Option<RustcVersion>) -> Self {
+        Self {
+            msrv,
+            ..Dereferencing::default()
+        }
+    }
 }
 
 struct StateData {
@@ -170,6 +191,7 @@ struct StateData {
 struct DerefedBorrow {
     count: usize,
     msg: &'static str,
+    snip_expr: Option<HirId>,
 }
 
 enum State {
@@ -250,7 +272,7 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
         match (self.state.take(), kind) {
             (None, kind) => {
                 let expr_ty = typeck.expr_ty(expr);
-                let (position, adjustments) = walk_parents(cx, expr);
+                let (position, adjustments) = walk_parents(cx, expr, self.msrv);
 
                 match kind {
                     RefOp::Deref => {
@@ -331,20 +353,23 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
                         let deref_msg =
                             "this expression creates a reference which is immediately dereferenced by the compiler";
                         let borrow_msg = "this expression borrows a value the compiler would automatically borrow";
+                        let impl_msg = "the borrowed expression implements the required traits";
 
-                        let (required_refs, msg) = if position.can_auto_borrow() {
-                            (1, if deref_count == 1 { borrow_msg } else { deref_msg })
+                        let (required_refs, msg, snip_expr) = if position.can_auto_borrow() {
+                            (1, if deref_count == 1 { borrow_msg } else { deref_msg }, None)
+                        } else if let Position::ImplArg(hir_id) = position {
+                            (0, impl_msg, Some(hir_id))
                         } else if let Some(&Adjust::Borrow(AutoBorrow::Ref(_, mutability))) =
                             next_adjust.map(|a| &a.kind)
                         {
                             if matches!(mutability, AutoBorrowMutability::Mut { .. }) && !position.is_reborrow_stable()
                             {
-                                (3, deref_msg)
+                                (3, deref_msg, None)
                             } else {
-                                (2, deref_msg)
+                                (2, deref_msg, None)
                             }
                         } else {
-                            (2, deref_msg)
+                            (2, deref_msg, None)
                         };
 
                         if deref_count >= required_refs {
@@ -354,6 +379,7 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
                                     // can't be removed without breaking the code. See earlier comment.
                                     count: deref_count - required_refs,
                                     msg,
+                                    snip_expr,
                                 }),
                                 StateData { span: expr.span, hir_id: expr.hir_id, position },
                             ));
@@ -510,7 +536,7 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
                             spans: vec![pat.span],
                             app,
                             replacements: vec![(pat.span, snip.into())],
-                            hir_id: pat.hir_id
+                            hir_id: pat.hir_id,
                         }),
                     );
                 }
@@ -542,6 +568,8 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
             self.current_body = None;
         }
     }
+
+    extract_msrv_attr!(LateContext);
 }
 
 fn try_parse_ref_op<'tcx>(
@@ -594,6 +622,7 @@ enum Position {
     /// The method is defined on a reference type. e.g. `impl Foo for &T`
     MethodReceiverRefImpl,
     Callee,
+    ImplArg(HirId),
     FieldAccess(Symbol),
     Postfix,
     Deref,
@@ -630,7 +659,7 @@ impl Position {
             | Self::Callee
             | Self::FieldAccess(_)
             | Self::Postfix => PREC_POSTFIX,
-            Self::Deref => PREC_PREFIX,
+            Self::ImplArg(_) | Self::Deref => PREC_PREFIX,
             Self::DerefStable(p, _) | Self::ReborrowStable(p) | Self::Other(p) => p,
         }
     }
@@ -639,8 +668,12 @@ impl Position {
 /// Walks up the parent expressions attempting to determine both how stable the auto-deref result
 /// is, and which adjustments will be applied to it. Note this will not consider auto-borrow
 /// locations as those follow different rules.
-#[allow(clippy::too_many_lines)]
-fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &'tcx [Adjustment<'tcx>]) {
+#[expect(clippy::too_many_lines)]
+fn walk_parents<'tcx>(
+    cx: &LateContext<'tcx>,
+    e: &'tcx Expr<'_>,
+    msrv: Option<RustcVersion>,
+) -> (Position, &'tcx [Adjustment<'tcx>]) {
     let mut adjustments = [].as_slice();
     let mut precedence = 0i8;
     let ctxt = e.span.ctxt();
@@ -732,13 +765,20 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
                     .iter()
                     .position(|arg| arg.hir_id == child_id)
                     .zip(expr_sig(cx, func))
-                    .and_then(|(i, sig)| sig.input_with_hir(i))
-                    .map(|(hir_ty, ty)| match hir_ty {
-                        // Type inference for closures can depend on how they're called. Only go by the explicit
-                        // types here.
-                        Some(hir_ty) => binding_ty_auto_deref_stability(cx, hir_ty, precedence, ty.bound_vars()),
-                        None => ty_auto_deref_stability(cx, cx.tcx.erase_late_bound_regions(ty), precedence)
-                            .position_for_arg(),
+                    .and_then(|(i, sig)| {
+                        sig.input_with_hir(i).map(|(hir_ty, ty)| match hir_ty {
+                            // Type inference for closures can depend on how they're called. Only go by the explicit
+                            // types here.
+                            Some(hir_ty) => binding_ty_auto_deref_stability(cx, hir_ty, precedence, ty.bound_vars()),
+                            None => {
+                                if let ty::Param(param_ty) = ty.skip_binder().kind() {
+                                    needless_borrow_impl_arg_position(cx, parent, i, *param_ty, e, precedence, msrv)
+                                } else {
+                                    ty_auto_deref_stability(cx, cx.tcx.erase_late_bound_regions(ty), precedence)
+                                        .position_for_arg()
+                                }
+                            },
+                        })
                     }),
                 ExprKind::MethodCall(_, args, _) => {
                     let id = cx.typeck_results().type_dependent_def_id(parent.hir_id).unwrap();
@@ -779,12 +819,17 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
                                 Position::MethodReceiver
                             }
                         } else {
-                            ty_auto_deref_stability(
-                                cx,
-                                cx.tcx.erase_late_bound_regions(cx.tcx.fn_sig(id).input(i)),
-                                precedence,
-                            )
-                            .position_for_arg()
+                            let ty = cx.tcx.fn_sig(id).skip_binder().inputs()[i];
+                            if let ty::Param(param_ty) = ty.kind() {
+                                needless_borrow_impl_arg_position(cx, parent, i, *param_ty, e, precedence, msrv)
+                            } else {
+                                ty_auto_deref_stability(
+                                    cx,
+                                    cx.tcx.erase_late_bound_regions(cx.tcx.fn_sig(id).input(i)),
+                                    precedence,
+                                )
+                                .position_for_arg()
+                            }
                         }
                     })
                 },
@@ -946,6 +991,205 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
     v.0
 }
 
+// Checks whether:
+// * child is an expression of the form `&e` in an argument position requiring an `impl Trait`
+// * `e`'s type implements `Trait` and is copyable
+// If the conditions are met, returns `Some(Position::ImplArg(..))`; otherwise, returns `None`.
+//   The "is copyable" condition is to avoid the case where removing the `&` means `e` would have to
+// be moved, but it cannot be.
+fn needless_borrow_impl_arg_position<'tcx>(
+    cx: &LateContext<'tcx>,
+    parent: &Expr<'tcx>,
+    arg_index: usize,
+    param_ty: ParamTy,
+    mut expr: &Expr<'tcx>,
+    precedence: i8,
+    msrv: Option<RustcVersion>,
+) -> Position {
+    let destruct_trait_def_id = cx.tcx.lang_items().destruct_trait();
+    let sized_trait_def_id = cx.tcx.lang_items().sized_trait();
+
+    let Some(callee_def_id) = fn_def_id(cx, parent) else { return Position::Other(precedence) };
+    let fn_sig = cx.tcx.fn_sig(callee_def_id).skip_binder();
+    let substs_with_expr_ty = cx
+        .typeck_results()
+        .node_substs(if let ExprKind::Call(callee, _) = parent.kind {
+            callee.hir_id
+        } else {
+            parent.hir_id
+        });
+
+    let predicates = cx.tcx.param_env(callee_def_id).caller_bounds();
+    let projection_predicates = predicates
+        .iter()
+        .filter_map(|predicate| {
+            if let PredicateKind::Projection(projection_predicate) = predicate.kind().skip_binder() {
+                Some(projection_predicate)
+            } else {
+                None
+            }
+        })
+        .collect::<Vec<_>>();
+
+    let mut trait_with_ref_mut_self_method = false;
+
+    // If no traits were found, or only the `Destruct`, `Sized`, or `Any` traits were found, return.
+    if predicates
+        .iter()
+        .filter_map(|predicate| {
+            if let PredicateKind::Trait(trait_predicate) = predicate.kind().skip_binder()
+                && trait_predicate.trait_ref.self_ty() == param_ty.to_ty(cx.tcx)
+            {
+                Some(trait_predicate.trait_ref.def_id)
+            } else {
+                None
+            }
+        })
+        .inspect(|trait_def_id| {
+            trait_with_ref_mut_self_method |= has_ref_mut_self_method(cx, *trait_def_id);
+        })
+        .all(|trait_def_id| {
+            Some(trait_def_id) == destruct_trait_def_id
+                || Some(trait_def_id) == sized_trait_def_id
+                || cx.tcx.is_diagnostic_item(sym::Any, trait_def_id)
+        })
+    {
+        return Position::Other(precedence);
+    }
+
+    // `substs_with_referent_ty` can be constructed outside of `check_referent` because the same
+    // elements are modified each time `check_referent` is called.
+    let mut substs_with_referent_ty = substs_with_expr_ty.to_vec();
+
+    let mut check_referent = |referent| {
+        let referent_ty = cx.typeck_results().expr_ty(referent);
+
+        if !is_copy(cx, referent_ty) {
+            return false;
+        }
+
+        // https://github.com/rust-lang/rust-clippy/pull/9136#pullrequestreview-1037379321
+        if trait_with_ref_mut_self_method && !matches!(referent_ty.kind(), ty::Ref(_, _, Mutability::Mut)) {
+            return false;
+        }
+
+        if !replace_types(
+            cx,
+            param_ty,
+            referent_ty,
+            fn_sig,
+            arg_index,
+            &projection_predicates,
+            &mut substs_with_referent_ty,
+        ) {
+            return false;
+        }
+
+        predicates.iter().all(|predicate| {
+            if let PredicateKind::Trait(trait_predicate) = predicate.kind().skip_binder()
+                && cx.tcx.is_diagnostic_item(sym::IntoIterator, trait_predicate.trait_ref.def_id)
+                && let ty::Param(param_ty) = trait_predicate.self_ty().kind()
+                && let GenericArgKind::Type(ty) = substs_with_referent_ty[param_ty.index as usize].unpack()
+                && ty.is_array()
+                && !meets_msrv(msrv, msrvs::ARRAY_INTO_ITERATOR)
+            {
+                return false;
+            }
+
+            let predicate = EarlyBinder(predicate).subst(cx.tcx, &substs_with_referent_ty);
+            let obligation = Obligation::new(ObligationCause::dummy(), cx.param_env, predicate);
+            cx.tcx
+                .infer_ctxt()
+                .enter(|infcx| infcx.predicate_must_hold_modulo_regions(&obligation))
+        })
+    };
+
+    let mut needless_borrow = false;
+    while let ExprKind::AddrOf(_, _, referent) = expr.kind {
+        if !check_referent(referent) {
+            break;
+        }
+        expr = referent;
+        needless_borrow = true;
+    }
+
+    if needless_borrow {
+        Position::ImplArg(expr.hir_id)
+    } else {
+        Position::Other(precedence)
+    }
+}
+
+fn has_ref_mut_self_method(cx: &LateContext<'_>, trait_def_id: DefId) -> bool {
+    cx.tcx
+        .associated_items(trait_def_id)
+        .in_definition_order()
+        .any(|assoc_item| {
+            if assoc_item.fn_has_self_parameter {
+                let self_ty = cx.tcx.fn_sig(assoc_item.def_id).skip_binder().inputs()[0];
+                matches!(self_ty.kind(), ty::Ref(_, _, Mutability::Mut))
+            } else {
+                false
+            }
+        })
+}
+
+// Iteratively replaces `param_ty` with `new_ty` in `substs`, and similarly for each resulting
+// projected type that is a type parameter. Returns `false` if replacing the types would have an
+// effect on the function signature beyond substituting `new_ty` for `param_ty`.
+// See: https://github.com/rust-lang/rust-clippy/pull/9136#discussion_r927212757
+fn replace_types<'tcx>(
+    cx: &LateContext<'tcx>,
+    param_ty: ParamTy,
+    new_ty: Ty<'tcx>,
+    fn_sig: FnSig<'tcx>,
+    arg_index: usize,
+    projection_predicates: &[ProjectionPredicate<'tcx>],
+    substs: &mut [ty::GenericArg<'tcx>],
+) -> bool {
+    let mut replaced = BitSet::new_empty(substs.len());
+
+    let mut deque = VecDeque::with_capacity(substs.len());
+    deque.push_back((param_ty, new_ty));
+
+    while let Some((param_ty, new_ty)) = deque.pop_front() {
+        // If `replaced.is_empty()`, then `param_ty` and `new_ty` are those initially passed in.
+        if !fn_sig
+            .inputs_and_output
+            .iter()
+            .enumerate()
+            .all(|(i, ty)| (replaced.is_empty() && i == arg_index) || !contains_ty(ty, param_ty.to_ty(cx.tcx)))
+        {
+            return false;
+        }
+
+        substs[param_ty.index as usize] = ty::GenericArg::from(new_ty);
+
+        // The `replaced.insert(...)` check provides some protection against infinite loops.
+        if replaced.insert(param_ty.index) {
+            for projection_predicate in projection_predicates {
+                if projection_predicate.projection_ty.self_ty() == param_ty.to_ty(cx.tcx)
+                    && let ty::Term::Ty(term_ty) = projection_predicate.term
+                    && let ty::Param(term_param_ty) = term_ty.kind()
+                {
+                    let item_def_id = projection_predicate.projection_ty.item_def_id;
+                    let assoc_item = cx.tcx.associated_item(item_def_id);
+                    let projection = cx.tcx
+                        .mk_projection(assoc_item.def_id, cx.tcx.mk_substs_trait(new_ty, &[]));
+
+                    if let Ok(projected_ty) = cx.tcx.try_normalize_erasing_regions(cx.param_env, projection)
+                        && substs[term_param_ty.index as usize] != ty::GenericArg::from(projected_ty)
+                    {
+                        deque.push_back((*term_param_ty, projected_ty));
+                    }
+                }
+            }
+        }
+    }
+
+    true
+}
+
 struct TyPosition<'tcx> {
     position: Position,
     ty: Option<Ty<'tcx>>,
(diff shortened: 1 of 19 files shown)


Jarcho:
Probably won't have time to look at this for a few days. Does this handle something like:
```rust
struct S<T>(Vec<T>);
impl<T> S<T> {
    fn push(&mut self, item: T) where T: Eq { .. }
}
```

smoelius:
No problem at all.

Something like `S.push(&0)` would be ignored. There is a condition that the changed argument's type parameter not appear elsewhere in the function signature. In this case, `T` appears in the type of the `&mut self` argument.

smoelius:
However, you caused me to realize I was incorrectly assuming the argument type was a type parameter.

I have corrected the error and added a relevant test.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +931,115 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+            return false;
+        }
+
+        // * If an applicable trait is found and `referent_ty` implements it, `needless_borrow` is set to
+        //   `true`.
+        // * If an applicable trait is found that `referent_ty` does not implement, `false` is returned
+        //   immediately.
+        // * If no applicable traits are found, `needless_borrow` remains `false`.
+        let mut needless_borrow = false;
+        for &(trait_def_id, substs) in &traits {
```
Jarcho, line 1014: This could use `traits.iter().all(..)`. I don't see a reason to treat `Sized` specially. All the predicates must be met no matter what they are and it's already been checked that there's at least one additional predicate.

Jarcho (changes requested):
This should probably ignore cases where the trait has a `&mut self` method, but the unreferenced type is not a mutable reference. e.g.
```rust
#[derive(Clone, Copy)]
struct Iter(..);
impl Iterator for Iter { .. }

fn takes_iter(_: impl Iterator) {}
fn dont_warn(mut x: Iter) -> {
    takes_iter(&mut x);
}
fn warn(mut x: &mut Iter) -> {
    takes_iter(&mut x)
}
```

Iterators shouldn't implement `Copy` for exactly this reason, but that doesn't stop it from happening.

mikerite:
I got a crash with this branch on the crossbeam-epoch-0.9.9 crate.
```
thread 'rustc' panicked at 'assertion failed: !value.has_escaping_bound_vars()'
```

<details>
<summary>Backtrace</summary>

```
thread 'rustc' panicked at 'assertion failed: !value.has_escaping_bound_vars()', /rustc/ddcbba036aee08f0709f98a92a342a278eae5c05/compiler/rustc_middle/src/ty/sty.rs:996:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/ddcbba036aee08f0709f98a92a342a278eae5c05/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/ddcbba036aee08f0709f98a92a342a278eae5c05/library/core/src/panicking.rs:142:14
   2: core::panicking::panic
             at /rustc/ddcbba036aee08f0709f98a92a342a278eae5c05/library/core/src/panicking.rs:48:5
   3: <rustc_middle::ty::sty::Binder<rustc_middle::ty::sty::TraitRef>>::dummy
   4: <rustc_infer::infer::InferCtxt as rustc_trait_selection::infer::InferCtxtExt>::type_implements_trait
   5: rustc_infer::infer::InferCtxtBuilder::enter
   6: clippy_utils::ty::implements_trait_with_env
   7: clippy_utils::ty::implements_trait
   8: clippy_lints::dereference::needless_borrow_impl_arg_position
   9: <clippy_lints::dereference::Dereferencing as rustc_lint::passes::LateLintPass>::check_expr
  10: <rustc_lint::late::LateLintPassObjects as rustc_lint::passes::LateLintPass>::check_expr
  11: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  12: rustc_hir::intravisit::walk_expr::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  13: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  14: rustc_hir::intravisit::walk_expr::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  15: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  16: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_block
  17: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  18: rustc_hir::intravisit::walk_expr::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  19: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  20: rustc_hir::intravisit::walk_block::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  21: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_block
  22: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  23: rustc_hir::intravisit::walk_expr::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  24: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  25: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  26: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_block
  27: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_expr
  28: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_nested_body
  29: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_fn
  30: rustc_hir::intravisit::walk_impl_item::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  31: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_nested_impl_item
  32: rustc_hir::intravisit::walk_impl_item_ref::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  33: rustc_hir::intravisit::walk_item::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  34: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_nested_item
  35: rustc_hir::intravisit::walk_mod::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  36: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_mod
  37: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_nested_item
  38: rustc_hir::intravisit::walk_mod::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>>
  39: <rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects> as rustc_hir::intravisit::Visitor>::visit_mod
  40: rustc_lint::late::late_lint_pass_crate::<rustc_lint::late::LateLintPassObjects>
  41: rustc_lint::late::late_lint_crate::<rustc_lint::BuiltinCombinedLateLintPass>
  42: rustc_data_structures::sync::join::<rustc_lint::late::check_crate<rustc_lint::BuiltinCombinedLateLintPass, rustc_interface::passes::analysis::{closure#5}::{closure#1}::{closure#2}::{closure#0}::{closure#0}>::{closure#0}, rustc_lint::late::check_crate<rustc_lint::BuiltinCombinedLateLintPass, rustc_interface::passes::analysis::{closure#5}::{closure#1}::{closure#2}::{closure#0}::{closure#0}>::{closure#1}, (), ()>
  43: <rustc_session::session::Session>::time::<(), rustc_interface::passes::analysis::{closure#5}::{closure#1}::{closure#2}::{closure#0}>
  44: <core::panic::unwind_safe::AssertUnwindSafe<rustc_interface::passes::analysis::{closure#5}::{closure#1}> as core::ops::function::FnOnce<()>>::call_once
  45: <rustc_session::session::Session>::time::<(), rustc_interface::passes::analysis::{closure#5}>
  46: rustc_interface::passes::analysis
  47: <rustc_query_system::dep_graph::graph::DepGraph<rustc_middle::dep_graph::dep_node::DepKind>>::with_task::<rustc_middle::ty::context::TyCtxt, (), core::result::Result<(), rustc_errors::ErrorGuaranteed>>
  48: rustc_query_system::query::plumbing::try_execute_query::<rustc_query_impl::plumbing::QueryCtxt, rustc_query_system::query::caches::DefaultCache<(), core::result::Result<(), rustc_errors::ErrorGuaranteed>>>
  49: rustc_query_system::query::plumbing::get_query::<rustc_query_impl::queries::analysis, rustc_query_impl::plumbing::QueryCtxt>
  50: <rustc_interface::passes::QueryContext>::enter::<rustc_driver::run_compiler::{closure#1}::{closure#2}::{closure#3}, core::result::Result<(), rustc_errors::ErrorGuaranteed>>
  51: <rustc_interface::interface::Compiler>::enter::<rustc_driver::run_compiler::{closure#1}::{closure#2}, core::result::Result<core::option::Option<rustc_interface::queries::Linker>, rustc_errors::ErrorGuaranteed>>
  52: rustc_span::with_source_map::<core::result::Result<(), rustc_errors::ErrorGuaranteed>, rustc_interface::interface::create_compiler_and_run<core::result::Result<(), rustc_errors::ErrorGuaranteed>, rustc_driver::run_compiler::{closure#1}>::{closure#1}>
  53: <scoped_tls::ScopedKey<rustc_span::SessionGlobals>>::set::<rustc_interface::interface::run_compiler<core::result::Result<(), rustc_errors::ErrorGuaranteed>, rustc_driver::run_compiler::{closure#1}>::{closure#0}, core::result::Result<(), rustc_errors::ErrorGuaranteed>>
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

note: Clippy version: clippy 0.1.64 (2507d2ad8 2022-07-09)
```

</details>

smoelius:
Thanks, @mikerite!

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +931,115 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+// * child is an expression of the form `&e` in an argument position requiring an `impl Trait`
+// * `e`'s type implements `Trait` and is copyable
+// If the conditions are met, returns `Some(Position::ImplArg(..))`; otherwise, returns `None`.
+//   The "is copyable" condition is to avoid the case where removing the `&` means `e` would have to
+// be moved, but it cannot be.
+fn needless_borrow_impl_arg_position<'tcx>(
+    cx: &LateContext<'tcx>,
+    parent: &Expr<'tcx>,
+    child_arg_index: usize,
+    child_arg_ty: Ty<'tcx>,
```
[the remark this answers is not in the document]
smoelius: Is there an easy way to obtain a `ParamTy` here?

The `ParamTy` would replace both `child_arg_index` and `child_arg_ty` arguments, correct? If so, how would you imagine this line changing? https://github.com/rust-lang/rust-clippy/blob/2507d2ad883b56acd28506f34f4ae6918a32173c/clippy_lints/src/dereference.rs#L996

clippy_lints/src/dereference.rs
```diff
@@ -751,12 +758,14 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
...
-                        // Type inference for closures can depend on how they're called. Only go by the explicit
-                        // types here.
-                        Some(ty) => binding_ty_auto_deref_stability(ty, precedence),
-                        None => param_auto_deref_stability(ty.skip_binder(), precedence),
+                    .and_then(|(i, sig)| {
+                        sig.input_with_hir(i).map(|(hir_ty, ty)| match hir_ty {
+                            // Type inference for closures can depend on how they're called. Only go by the explicit
+                            // types here.
+                            Some(ty) => binding_ty_auto_deref_stability(ty, precedence),
+                            None => needless_borrow_impl_arg_position(cx, parent, i, ty.skip_binder(), child_id)
```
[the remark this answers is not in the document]
smoelius: Is the idea that, when `needless_borrow_impl_arg_position` applies, `child_id` should refer to the same expression as the one passed into `walk_parents`? (I'm trying to understand why it seems to work now.)

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +931,115 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+        .enumerate()
+        .any(|(i, ty)| i != child_arg_index && contains_ty(ty, child_arg_ty))
+    {
+        return None;
+    }
+
+    let check_referent = |referent| {
+        let referent_ty = cx.typeck_results().expr_ty(referent);
+
+        if !is_copy(cx, referent_ty) {
```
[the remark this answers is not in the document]
smoelius: What would you recommend here? I looked to [`trivially_copy_pass_by_ref`](https://rust-lang.github.io/rust-clippy/master/#trivially_copy_pass_by_ref) for inspiration, but it seems to have no size limit by default.

clippy_lints/src/dereference.rs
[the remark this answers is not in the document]
Jarcho: It would only replace `child_arg_ty`. Would look like
```rust
if let ty::Param(ty) = ty {
    needless_borrow_impl_arg_position(...)
} else {
    param_auto_deref_stability(...)
}
```
You would have to change `needless_borrow_impl_arg_position` to return `Position::Other` which would be correct for any generic param.

clippy_lints/src/dereference.rs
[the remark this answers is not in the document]
Jarcho: It's limited to 8 bytes by default. A little extreme as a fat pointer should still be passed by value in general. Can you check through all the types that were caught in clippy and see what the sizes are?

clippy_lints/src/dereference.rs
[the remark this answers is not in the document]
smoelius: Are there no cases where `ty` is a `Param` and `needless_borrow_impl_arg_position` would return `None` but `param_auto_deref_stability` returns `Some(..)`? Put another way, are we sure it is okay to elide the current code's `unwrap_or_else`?

Separately, could you please explain why this change necessitates the use of `Position::Other`? I am currently using `Position::ImplArg` to determine the error message: https://github.com/rust-lang/rust-clippy/blob/2507d2ad883b56acd28506f34f4ae6918a32173c/clippy_lints/src/dereference.rs#L340-L341

clippy_lints/src/dereference.rs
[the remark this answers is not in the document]
smoelius: It looks like this was the largest at 160 bytes: https://github.com/rust-lang/rust-clippy/blob/2507d2ad883b56acd28506f34f4ae6918a32173c/tests/lint_message_convention.rs#L20-L31

clippy_lints/src/dereference.rs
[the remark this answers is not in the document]
Jarcho: `param_auto_deref_stability` returns `Position::Other(_)` for generic types.

There's no need to return an `Option` from `needless_borrow_impl_arg_position `. Either you can remove borrows, or nothing interesting can be done with the location the expression is in. The latter case uses `Position::Other(_)`.

smoelius:
I think the most recent push addresses all of the review comments up to this point, and the crash. Please let me know if I misunderstood anything.

Regarding the "is copyable" condition, I imposed a hard limit of 256 bytes. Please let me know if you would prefer a different approach.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +943,148 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+
+    let Some(callee_def_id) = fn_def_id(cx, parent) else { return Position::Other(precedence) };
+
+    let (trait_predicates, projection_predicates) =
+        get_input_traits_and_projections(cx, callee_def_id, param_ty.to_ty(cx.tcx));
+
+    // If no traits were found, or only the `Sized` or `Any` traits were found, return.
+    if trait_predicates.iter().all(|trait_predicate| {
+        let trait_def_id = trait_predicate.def_id();
+        Some(trait_def_id) == sized_trait_def_id || match_def_path(cx, trait_def_id, &paths::ANY_TRAIT)
```
Jarcho, line 977: Should use `cx.tcx.is_diagnostic_item(sym::Any, trait_def_id)`

Jarcho:
A couple of comments now that I've had time to think through the implementation.
* You can ignore my earlier comment about the size of parameters. The calling convention for large types is that they're passed by reference anyways. Sorry for wasting time on that one.
* Currently this only handles `T: SomeTrait` predicates, but `T: SomeTrait<U>` and `U: SomeTrait<T>` predicates also need to be handled. This requires a few steps:
    1. Get the current substitution list for the function call.
    2. Replace the substitution for `T` with the new type.
    3. Apply the new substitution list to each predicate.
    4. Check each new substituted predicate with [`predicate_must_hold_modulo_regions`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/query/evaluate_obligation/trait.InferCtxtExt.html#tymethod.predicate_must_hold_modulo_regions). The `InferCtxt` can be created with `cx.tcx.infer_ctxt().enter(|infcx| ...)`

  You'll still want to keep the rest of the restrictions in place in order to avoid false positives.

smoelius:
> * Currently this only handles `T: SomeTrait` predicates, but `T: SomeTrait<U>` and `U: SomeTrait<T>` predicates also need to be handled. This requires a few steps:
>   
>   1. Get the current substitution list for the function call.
>   2. Replace the substitution for `T` with the new type.
>   3. Apply the new substitution list to each predicate.
>   4. Check each new substituted predicate with [`predicate_must_hold_modulo_regions`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_trait_selection/traits/query/evaluate_obligation/trait.InferCtxtExt.html#tymethod.predicate_must_hold_modulo_regions). The `InferCtxt` can be created with `cx.tcx.infer_ctxt().enter(|infcx| ...)`

You're saying this should be done for all predicates, not just trait predicates, correct?

If so, are there any tools you know of to make iii easier? I.e., something easier than performing a match and handling each `PredicateKind` individually?

Jarcho:
The [`subst`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/subst/trait.Subst.html#tymethod.subst) method will perform the actual substitution. It exists on both `List<Predicate>>` and `Predicate` itself

smoelius:
I think the second round of comments are addressed.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +941,177 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+        // https://github.com/rust-lang/rust-clippy/pull/9136#pullrequestreview-1037379321
+        if !matches!(referent_ty.kind(), ty::Ref(_, _, Mutability::Mut))
+            && trait_predicates
+                .iter()
+                .any(|TraitPredicate { trait_ref, .. }| has_ref_mut_self_method(cx, trait_ref.def_id))
+        {
+            return false;
+        }
+
+        let substs_with_expr_ty = substs_with_expr_ty.split_at(callee_generics.parent_count);
```
Jarcho, line 1035: This whole thing can be replaced with `substs[param_ty.index as usize] = ty::GenericArg::from(referent_ty);` assuming `substs` is a slice containing the original substitutions.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +941,177 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+                                    }
+                                })
+                                .unwrap_or(*arg)
+                        }
+                    }),
+            )
+            .collect::<Vec<_>>();
+
+        let predicates = EarlyBinder(predicates).subst(cx.tcx, &substs_with_referent_ty);
+        if !predicates.iter().all(|predicate| {
```
Jarcho, line 1073: The result of `all` can just be returned directly.

clippy_lints/src/dereference.rs
smoelius: Wow. That is a lot simpler.

smoelius:
Please let me know if I misunderstood any of your suggestions, @Jarcho.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +941,150 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+            && trait_predicates
+                .iter()
+                .any(|TraitPredicate { trait_ref, .. }| has_ref_mut_self_method(cx, trait_ref.def_id))
+        {
+            return false;
+        }
+
+        let mut substs_with_referent_ty = substs_with_expr_ty.to_vec();
+        substs_with_referent_ty[param_ty.index as usize] = ty::GenericArg::from(referent_ty);
+        for projection_predicate in &projection_predicates {
```
Jarcho, line 1035: This loop can just be removed. It takes any predicate `<T as Trait>::Assoc = U` and modifies the substitution list replacing `U` with `<V as Trait>::Assoc` where `V` is the argument type minus the reference, and `U` is some type parameter.

The only item in the substitution list that needs to change is the one for the argument being modified.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +941,150 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+        // https://github.com/rust-lang/rust-clippy/pull/9136#pullrequestreview-1037379321
+        if !matches!(referent_ty.kind(), ty::Ref(_, _, Mutability::Mut))
+            && trait_predicates
+                .iter()
+                .any(|TraitPredicate { trait_ref, .. }| has_ref_mut_self_method(cx, trait_ref.def_id))
+        {
+            return false;
+        }
+
+        let mut substs_with_referent_ty = substs_with_expr_ty.to_vec();
```
Jarcho, line 1033: The `Vec` can be constructed outside the closure. The same item is being modified in every call.

clippy_lints/src/dereference.rs
smoelius: Without the loop on line 1035, this test fails: https://github.com/rust-lang/rust-clippy/blob/cd70a3d3b1c8e019c8938b259bac496b4179dd75/tests/ui/needless_borrow.rs#L131
In particular, `substs_with_referent_ty` looks like this:
```
[[&str; 2], &&str]
```
And the following predicate fails:
```
Binder(ProjectionPredicate(ProjectionTy { substs: [[&str; 2]], item_def_id: DefId(2:8368 ~ core[32c2]::iter::traits::collect::IntoIterator::Item) }, Ty(&&str)), [])
```
With the loop, `substs_with_referent_ty` looks like this:
```
[[&str; 2], &str]
```
Essentially, the types projected from `&T` and `T` differ.

I am open to doing this differently. What I have now feels manual and is certainly not general.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +941,150 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+        } else {
+            parent.hir_id
+        });
+
+    let predicates = cx.tcx.param_env(callee_def_id).caller_bounds();
+    let trait_predicates = predicates
+        .iter()
+        .filter_map(|predicate| {
+            if let PredicateKind::Trait(trait_predicate) = predicate.kind().skip_binder()
+                && is_substs_for_type(trait_predicate.trait_ref.substs, param_ty.to_ty(cx.tcx))
```
Jarcho, line 974: This could be `trait_predicate.trait_ref.self_ty() == param_ty.to_ty(cx.tcx)`

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +941,150 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+    let substs_with_expr_ty = cx
+        .typeck_results()
+        .node_substs(if let ExprKind::Call(callee, _) = parent.kind {
+            callee.hir_id
+        } else {
+            parent.hir_id
+        });
+
+    let predicates = cx.tcx.param_env(callee_def_id).caller_bounds();
+    let trait_predicates = predicates
```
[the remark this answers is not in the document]
smoelius: This vector is also used in the `has_ref_mut_self_method` check, actually.

smoelius:
@Jarcho How's this?

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +944,191 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+                .inputs_and_output
+                .iter()
+                .enumerate()
+                .all(|(i, ty)| (replaced.is_empty() && i == arg_index) || !contains_ty(ty, param_ty.to_ty(cx.tcx))))
+        {
+            return false;
+        }
+
+        // The next line should not panic, but panicking seems better than going into an infinite loop.
+        assert!(replaced.insert(param_ty.index));
```
Jarcho, line 1109: I believe this panics on:
```rust
trait X {
    type X;
}
impl X for () {
    type X = ();
}

fn f<T, U, Z>(_: T)
where T: X<X = U>,
    U: X<X = Z>,
    T: X<X = Z>,
{}

fn main() {
    f(());
}
```

I'm not sure why anyone would type this, but you can safely continue to the next type in this case.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +944,191 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+    let mut replaced = BitSet::new_empty(substs.len());
+
+    let mut deque = VecDeque::with_capacity(substs.len());
+    deque.push_back((param_ty, new_ty));
+
+    while let Some((param_ty, new_ty)) = deque.pop_front() {
+        let new_generic_arg = ty::GenericArg::from(new_ty);
+
+        // If `replaced.is_empty()`, then `param_ty` and `new_ty` are those initially passed in.
+        if !(substs[param_ty.index as usize] == new_generic_arg
```
Jarcho, line 1098: This check could be done right before pushing the types to the queue thereby removing the need to check for the initial argument.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +944,191 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+                && let ty::Term::Ty(term_ty) = projection_predicate.term
+                && let ty::Param(term_param_ty) = term_ty.kind()
+            {
+                let item_def_id = projection_predicate.projection_ty.item_def_id;
+                let assoc_item = cx.tcx.associated_item(item_def_id);
+                let projection = cx.tcx
+                    .mk_projection(assoc_item.def_id, cx.tcx.mk_substs_trait(new_ty, &[]));
+                let projected_ty = cx.tcx.normalize_erasing_regions(cx.param_env, projection);
+
+                deque.push_back((*term_param_ty, projected_ty));
```
Jarcho, line 1124: Combined with the previous change this could modify the substitution for `term_param_ty` immediately rather than storing it for later. The check for `replaced.insert(...)` could also be moved here before pushing the item onto the queue.

Jarcho (changes requested):

clippy_lints/src/dereference.rs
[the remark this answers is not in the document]
Jarcho: Give the current code all the checks could be combined into a single `any` call.

smoelius:
Is this about what you had in mind? If so, I am concerned that the optimizations are starting to affect the code's maintainability.

I personally do not like making the `any`'s closure not `Fn`.

And while unrolling loop eliminates a check, it also causes some code to be duplicated.

I don't have a strong opinion, but I do think the code was more clear with the allocation and the loop rolled.

Apologies if I have misunderstood your suggestions.

Jarcho:
You could use `inspect` to separate the two, or just use a for loop.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +944,201 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+    new_ty: Ty<'tcx>,
+    fn_sig: FnSig<'tcx>,
+    arg_index: usize,
+    projection_predicates: &[ProjectionPredicate<'tcx>],
+    substs: &mut [ty::GenericArg<'tcx>],
+) -> bool {
+    let mut replaced = BitSet::new_empty(substs.len());
+    let mut deque = VecDeque::with_capacity(substs.len());
+
+    if fn_sig
```
Jarcho, line 1088: I had missed that you removed the check from before `replace_types`. I had thought you still had it near the beginning of `needless_borrow_impl_arg_position`. You can revert the loop unrolling.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +944,201 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+        .inputs_and_output
+        .iter()
+        .enumerate()
+        .any(|(i, ty)| i != arg_index && contains_ty(ty, param_ty.to_ty(cx.tcx)))
```
Jarcho, line 1092: I had missed that you removed the check from before `replace_types`. I had thought you still had it near the beginning of `needless_borrow_impl_arg_position`. You can revert the loop unrolling.

clippy_lints/src/dereference.rs
```diff
@@ -920,6 +944,201 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
...
+                let inserted = replaced.insert(term_param_ty.index);
+
+                let new_generic_arg = ty::GenericArg::from(projected_ty);
+
+                if substs[term_param_ty.index as usize] == new_generic_arg {
+                    continue;
+                }
+
+                // The next line provides some protection against infinite loops.
+                assert!(inserted);
```
Jarcho, line 1122: This can still trigger, though the case is much more contrived and requires explicitly passing a generic argument. You can do:
```rust
if insert {
    // All the work in here
}
```
The only way you can fail to insert is if the type has already been placed in the queue. If this happens the inference has already been propagated, or it will be in a future iteration. In both cases nothing needs to happen.

smoelius:
I reverted the loop unrolling and made it so that `has_ref_mut_self_method` is called from a closure called by `inspect`.

smoelius:
> * `Some("").unwrap_or(&"")` gives the new error message when it should give the original one.

What would make sense here? Ignore `core::marker::Destruct`, like we do `Any` and `Sized`? https://doc.rust-lang.org/src/core/option.rs.html#776-778

Jarcho:
For this case `T` appears three times in the signature, so extra bounds shouldn't matter. `Destruct` should probably be ignored as well.

smoelius:
> For this case `T` appears three times in the signature, so extra bounds shouldn't matter.

For the example of `Some("").unwrap_or(&"")`, the substitution for `T` in `unwrap_or` is `&str` like in `Option::Some`. So the left side of this `&&` fails, avoiding the `return false`: https://github.com/rust-lang/rust-clippy/blob/5c5bf96be958a72effd9cce4f6379823bdbe93a2/clippy_lints/src/dereference.rs#L1098-L1103
(Admittedly, I am surprised this example compiles at all.)

However, ignoring `Destruct` produces the original error message, as desired.

Please pay particular attention to how I added the MSRV checks, as this is my first attempt at it.

Jarcho:
There's a type inference hack to make it adjustments work on types inferred earlier. It doesn't compile if you flip the where the reference is.

Is that first condition necessary? As far as I can tell it should only happen in this case which will be picked up by the normal logic, and if the type has already been modified in an earlier iteration. The early escape on the iteration can be handled by putting everything within the `replaced.insert` check.

The msrv check looks good. 👍

smoelius:
> Is that first condition necessary?

The first condition was mean to address https://github.com/rust-lang/rust-clippy/pull/9136#discussion_r927212757:
> * `U` doesn't exist in the argument list or return position.

> ...you can still lint, but only if the projection type normalizes to the same type for both the original type and the replacement type.

But maybe I'm not thinking about this right.

Jarcho:
You would need to use [`try_normalize_erasing_regions`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.try_normalize_erasing_regions) to do that. Two projection types from different bases are never considered equal, but the can normalize to the same type.

I think the conditions you want on the loop would be:
* `replaced.insert(..) -> continue`: The type has already been replaced so there's no need to do anything.
* `original_ty == replacement_ty -> continue`: The type is the same so there's nothing to change.
* `sig_contains_ty_param -> return false`: One of the types can't be replaced.

Jarcho:
I think that should work. The test would be calling a function like `fn f<T, U>(_: T, _: U) where T: Trait<Assoc=U> {}`. Both `&T` and `T` would need to have the same type for `Assoc`.

smoelius:
What you're describing sounds very similar to this test: https://github.com/rust-lang/rust-clippy/blob/7e18f4fd9b1377ad95deda49f60f31ecc3ccdd54/tests/ui/needless_borrow.rs#L135

But I think I am losing the thread. What behavior do you expect when calling a function like that?

Jarcho:
I really should have waited for when I had more time to review this. Totally missed that you were already normalizing the projection when making that last comment.

smoelius:
No problem. But do you think the check was better in its previous position?

Jarcho:
The current position is fine. It also avoids the same reading problem I made.

I think that needs to be `try_normalize_erasing_regions`. You don't have a case where `&T` implements a trait with a projection, but `T` doesn't. I think the `normalize_erasing_regions` ICE's in that case.

smoelius:
Please consider this paused. I have to fix this.

smoelius:
OK. I think this is fixed now. The new test ICEs without `try_normalize_erasing_regions`.

Jarcho:
That should be the last thing. I want to run one more check on more crates, but that will take a couple of days as I have other tests running right now.

Jarcho:
Looks like it mostly catches `AsRef` and `IntoIterator` with `Serialize` being a very distant third. Didn't see any FP's in the top 5000 crates.

You can squash down some of the commits and then it's good to go.

smoelius:
Done.

I left in the change that moved `get_input_traits_and_projections` to `clippy_utils`, even though this change was ultimately not needed. Please let me know if you would like me to undo this change.

Jarcho:
It would be better to remove it. This would get removed as an unused util function if we ever clean up useless/overlapping util functions.

I'm going to put a hold on merging this as it conflicts #9126 and the lint fixed there is about to beta. I can't guarantee I'll have time to deal with the merge conflict in the next week.

smoelius:
> It would be better to remove it. This would get removed as an unused util function if we ever clean up useless/overlapping util functions.

I moved `get_input_traits_and_projections` back into `unnecessary_to_owned.rs`, but kept a few changes that came out of this review.

> I'm going to put a hold on merging this as it conflicts #9126 and the lint fixed there is about to beta. I can't guarantee I'll have time to deal with the merge conflict in the next week.

No problem at all.

Jarcho:
Ping @smoelius just in case github didn't. This is good to merge once the conflicts are resolved. Sorry for dumping that on you.

smoelius:
Sorry, I misunderstood your previous message.

If it helps to verify the rebase, here is the diff between the old patch (ad778ad) and the new (032f112):
```rust
15c15
< index d513a229b..47bdc41a2 100644
---
> index 3b27f061e..357cf6fc4 100644
25c25
<              return Err(CliError::RaSetupActive);
---
>              return Err(CliError::IntellijSetupActive);
50c50
< index c089f4d8c..c831fb831 100644
---
> index 05e79a241..c503142e5 100644
72c72
< index a90f894a7..7d2b526bc 100644
---
> index e0b94f719..e38704701 100644
79c79
< -use clippy_utils::ty::{expr_sig, peel_mid_ty_refs, variant_of_res};
---
> -use clippy_utils::ty::{expr_sig, peel_mid_ty_refs, ty_sig, variant_of_res};
81c81
< +use clippy_utils::ty::{contains_ty, expr_sig, is_copy, peel_mid_ty_refs, variant_of_res};
---
> +use clippy_utils::ty::{contains_ty, expr_sig, is_copy, peel_mid_ty_refs, ty_sig, variant_of_res};
99c99
< -use rustc_middle::ty::{self, Ty, TyCtxt, TypeVisitable, TypeckResults};
---
> -use rustc_middle::ty::{self, Binder, BoundVariableKind, List, Ty, TyCtxt, TypeVisitable, TypeckResults};
101,102c101,102
< +    self, subst::Subst, EarlyBinder, FnSig, GenericArgKind, ParamTy, PredicateKind, ProjectionPredicate, Ty, TyCtxt,
< +    TypeVisitable, TypeckResults,
---
> +    self, subst::Subst, Binder, BoundVariableKind, EarlyBinder, FnSig, GenericArgKind, List, ParamTy, PredicateKind,
> +    ProjectionPredicate, Ty, TyCtxt, TypeVisitable, TypeckResults,
106c106
<  use rustc_span::{symbol::sym, Span, Symbol};
---
>  use rustc_span::{symbol::sym, Span, Symbol, DUMMY_SP};
196c196
< @@ -521,7 +547,7 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
---
> @@ -510,7 +536,7 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
205c205
< @@ -553,6 +579,8 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
---
> @@ -542,6 +568,8 @@ impl<'tcx> LateLintPass<'tcx> for Dereferencing {
214c214
< @@ -605,6 +633,7 @@ enum Position {
---
> @@ -594,6 +622,7 @@ enum Position {
222c222
< @@ -638,7 +667,7 @@ impl Position {
---
> @@ -630,7 +659,7 @@ impl Position {
228c228
<              Self::DerefStable(p) | Self::ReborrowStable(p) | Self::Other(p) => p,
---
>              Self::DerefStable(p, _) | Self::ReborrowStable(p) | Self::Other(p) => p,
231c231
< @@ -647,8 +676,12 @@ impl Position {
---
> @@ -639,8 +668,12 @@ impl Position {
246c246
< @@ -751,12 +784,19 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
---
> @@ -732,13 +765,20 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
254,255c254,256
< -                        Some(ty) => binding_ty_auto_deref_stability(ty, precedence),
< -                        None => param_auto_deref_stability(ty.skip_binder(), precedence),
---
> -                        Some(hir_ty) => binding_ty_auto_deref_stability(cx, hir_ty, precedence, ty.bound_vars()),
> -                        None => ty_auto_deref_stability(cx, cx.tcx.erase_late_bound_regions(ty), precedence)
> -                            .position_for_arg(),
260c261
< +                            Some(ty) => binding_ty_auto_deref_stability(ty, precedence),
---
> +                            Some(hir_ty) => binding_ty_auto_deref_stability(cx, hir_ty, precedence, ty.bound_vars()),
265c266,267
< +                                    param_auto_deref_stability(ty.skip_binder(), precedence)
---
> +                                    ty_auto_deref_stability(cx, cx.tcx.erase_late_bound_regions(ty), precedence)
> +                                        .position_for_arg()
272c274
< @@ -797,7 +837,12 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
---
> @@ -779,12 +819,17 @@ fn walk_parents<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> (Position, &
276c278,283
< -                            param_auto_deref_stability(cx.tcx.fn_sig(id).skip_binder().inputs()[i], precedence)
---
> -                            ty_auto_deref_stability(
> -                                cx,
> -                                cx.tcx.erase_late_bound_regions(cx.tcx.fn_sig(id).input(i)),
> -                                precedence,
> -                            )
> -                            .position_for_arg()
281c288,293
< +                                param_auto_deref_stability(ty, precedence)
---
> +                                ty_auto_deref_stability(
> +                                    cx,
> +                                    cx.tcx.erase_late_bound_regions(cx.tcx.fn_sig(id).input(i)),
> +                                    precedence,
> +                                )
> +                                .position_for_arg()
286c298
< @@ -920,6 +965,205 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
---
> @@ -946,6 +991,205 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
489,492c501,504
<  // Checks whether a type is stable when switching to auto dereferencing,
<  fn param_auto_deref_stability(ty: Ty<'_>, precedence: i8) -> Position {
<      let ty::Ref(_, mut ty, _) = *ty.kind() else {
< @@ -1026,7 +1270,8 @@ fn report<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, state: State, data
---
>  struct TyPosition<'tcx> {
>      position: Position,
>      ty: Option<Ty<'tcx>>,
> @@ -1084,7 +1328,8 @@ fn report<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, state: State, data
503c515
< index 8444532f6..ef8c33a03 100644
---
> index e6a405f81..521739c28 100644
506c518
< @@ -807,7 +807,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
---
> @@ -821,7 +821,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
516c528
< index b4c6bfb31..ad55ee63c 100644
---
> index 99b56da7a..e4624167a 100644
519,521c531,533
< @@ -5,9 +5,8 @@ use clippy_utils::source::snippet_opt;
<  use clippy_utils::ty::{
<      contains_ty, get_associated_type, get_iterator_item_ty, implements_trait, is_copy, peel_mid_ty_refs,
---
> @@ -6,9 +6,8 @@ use clippy_utils::ty::{
>      contains_ty, get_associated_type, get_iterator_item_ty, implements_trait, is_copy, is_type_diagnostic_item,
>      peel_mid_ty_refs,
530c542
< @@ -360,25 +359,15 @@ fn get_input_traits_and_projections<'tcx>(
---
> @@ -373,25 +372,15 @@ fn get_input_traits_and_projections<'tcx>(
```

Jarcho:
Thank you for pushing through and implementing all this.

@bors r+

smoelius:
Thanks for the detailed and thorough review!

PR #9137, opened 2022-07-08
Fix example header

alex-semenyuk:
changelog: fix example header


xFrednet (approved):

xFrednet:
Nice catch, thank you for the update :upside_down_face: 

@bors r+

PR #9144, opened 2022-07-10
Fix span for or_fun_call

### LINKED ISSUE #9033: Improve span of `or_fun_call` lint
### Description

The lint message currently spans through the entire method chain expression:

```console
error: use of `unwrap_or` followed by a call to `new`
  --> $DIR/or_fun_call.rs:182:9
   |
LL | /         frames
LL | |             .iter()
LL | |             .map(|f: &String| f.to_lowercase())
LL | |             .reduce(|mut acc, f| {
...  |
LL | |             })
LL | |             .unwrap_or(String::new());
   | |_____________________________________^
```

It would be enough to just point to the `unwrap_or`.

_Originally posted by @flip1995 in https://github.com/rust-lang/rust/pull/98261#discussion_r903425629_

### Version

_No response_

### Additional Labels

_No response_

alex-semenyuk:
Closes #9033
changelog: [`or_fun_call`]: span points to the `unwrap_or` only instead of through the entire method chain expression

### DIFF
--- a/clippy_lints/src/methods/or_fun_call.rs
+++ b/clippy_lints/src/methods/or_fun_call.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::eager_or_lazy::switch_to_lazy_eval;
-use clippy_utils::source::{snippet, snippet_with_applicability, snippet_with_macro_callsite};
+use clippy_utils::source::{snippet, snippet_with_macro_callsite};
 use clippy_utils::ty::{implements_trait, match_type};
 use clippy_utils::{contains_return, is_trait_item, last_path_segment, paths};
 use if_chain::if_chain;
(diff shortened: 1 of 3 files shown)


clippy_lints/src/methods/or_fun_call.rs
```diff
@@ -52,24 +52,15 @@ pub(super) fn check<'tcx>(
...
-                    snippet_with_applicability(cx, self_expr.span, "..", &mut applicability),
-                    hint
-                );
-
+                let span_replace_word = method_span.with_hi(span.hi());
                 span_lint_and_sugg(
                     cx,
                     OR_FUN_CALL,
-                    sugg_span,
+                    span_replace_word,
```
giraffate, line 59: The `span_replace_word` name looks like that it's not `span` but something about word, so it would be better to change its name to `replace_word_span` or just put it as it is like below.
```suggestion
                    method_span.with_hi(span.hi()),
```

giraffate (commented):

clippy_lints/src/methods/or_fun_call.rs
alex-semenyuk: done

giraffate:
@bors r+

Thanks!

PR #9149, opened 2022-07-11
change applicability type to MaybeIncorrect in `explicit_counter_loop`

### LINKED ISSUE #9013: Incorrect fix for variable being used as a loop counter
### Summary

I think the automatic fix for an index being used as a loop counter is broken. My guess is that it ought to be removing the original variable declaration and the increment statement. I'm getting the same error on both stable and nightly.

### Reproducer

I tried this code:

```rust
fn main() {
    let vector = vec!['a', 'b','c'];
    let mut index:usize = 0;
    
    for v in vector{
        println!("{v} {index}");
        index+=1;
    }
}
```

I ran `cargo clippy --fix` expecting either no change or refactoring to use `.into_iter().enumerate()` 

Instead, this happened:

```
warning: failed to automatically apply fixes suggested by rustc to crate `bugtest`

  * src\main.rs

warning: unused variable: `index`
 --> src\main.rs:3:13
  |
3 |     let mut index:usize = 0;
  |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_index`
  |

warning: value assigned to `index` is never read
 --> src\main.rs:7:9
  |
7 |         index+=1;
  |         ^^^^^
  |
  = note: `#[warn(unused_assignments)]` on by default
  = help: maybe it is overwritten before being read?

warning: variable does not need to be mutable
 --> src\main.rs:3:9
  |
3 |     let mut index:usize = 0;
  |         ----^^^^^
  |         |
  |         help: remove this `mut`
  |
  = note: `#[warn(unused_mut)]` on by default

error[E0384]: cannot assign twice to immutable variable `index`
 --> src\main.rs:7:9
  |
5 |     for (index, v) in vector.into_iter().enumerate(){
  |          -----
  |          |
  |          first assignment to `index`
  |          help: consider making this binding mutable: `mut index`
6 |         println!("{v} {index}");
7 |         index+=1;
  |         ^^^^^^^^ cannot assign twice to immutable variable

error: aborting due to previous error; 3 warnings emitted
```

### Version

```text
binary: rustc
commit-date: 2022-05-18
host: x86_64-pc-windows-msvc
release: 1.61.0
LLVM version: 14.0.0
```

### Additional Labels

kyoto7250:
close #9013 

This PR changes  applicability type to `MaybeIncorrect`,  because the suggestion is not `MachineApplicable`.

changelog: change applicability type to MaybeIncorrect in `explicit_counter_loop`

### DIFF
--- a/clippy_lints/src/loops/explicit_counter_loop.rs
+++ b/clippy_lints/src/loops/explicit_counter_loop.rs
@@ -1,5 +1,5 @@
 use super::{make_iterator_snippet, IncrementVisitor, InitializeVisitor, EXPLICIT_COUNTER_LOOP};
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::source::snippet_with_applicability;
 use clippy_utils::{get_enclosing_block, is_integer_const};
 use if_chain::if_chain;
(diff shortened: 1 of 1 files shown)


clippy_lints/src/loops/explicit_counter_loop.rs
```diff
@@ -34,24 +34,29 @@ pub(super) fn check<'tcx>(
...
-                                format!(
-                                    "for ({}, {}) in {}.enumerate()",
-                                    name,
-                                    snippet_with_applicability(cx, pat.span, "item", &mut applicability),
-                                    make_iterator_snippet(cx, arg, &mut applicability),
-                                ),
-                                applicability,
+                                |diag| {
+                                    diag.span_suggestion(
+                                        expr.span.with_hi(arg.span.hi()),
```
Jarcho, line 49: This should be done once before the `span_lint` call.

kyoto7250: I commited https://github.com/rust-lang/rust-clippy/pull/9149/commits/134d0dee41846d1c0867fa3745b8de25b52ae7af

Jarcho:
Thank you.

@bors r+

PR #9152, opened 2022-07-11
Fix Clippy version in `derive_partial_eq_without_eq` lint

mati865:
It was first added to Rust in https://github.com/rust-lang/rust/pull/97248 which missed 1.62 just by few days.

changelog: none


xFrednet:
LGTM, thank you for the fix. I'm going to check the versions from now on, while writing the changelog :upside_down_face: 

@bors r+

PR #9156, opened 2022-07-12
fix [`manual_flatten`] help texts order

### LINKED ISSUE #8948: manual_flatten: help texts are in the wrong order
### Description

Running clippy on [this file](https://github.com/rust-lang/miri/blob/master/cargo-miri/bin.rs), I get the following suggestion:
```
warning: unnecessary `if let` since only the `Ok` variant of the iterator element is used
   --> bin.rs:503:5
    |
503 | /     for arg in ArgSplitFlagValue::new(
504 | |         env::args().skip(3), // skip the program name, "miri" and "run" / "test"
505 | |         config_flag,
506 | |     ) {
...   |
509 | |         }
510 | |     }
    | |_____^
    |
    = note: `#[warn(clippy::manual_flatten)]` on by default
help: ...and remove the `if let` statement in the for loop
   --> bin.rs:507:9
    |
507 | /         if let Ok(config) = arg {
508 | |             cmd.arg(config_flag).arg(config);
509 | |         }
    | |_________^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_flatten
help: try
    |
503 ~     for arg in ArgSplitFlagValue::new(
504 +         env::args().skip(3), // skip the program name, "miri" and "run" / "test"
505 +         config_flag,
506 ~     ).flatten() {
    |
```
The two `help` seem to be in the wrong order. At least, it seems to start a sentence with `..and`, which is rather confusing. ;) It took me a while to parse this lint.

### Version

```text
rustc 1.63.0-nightly (c7b0452ec 2022-06-04)
binary: rustc
commit-hash: c7b0452ece11bf714f7cf2003747231931504d59
commit-date: 2022-06-04
release: 1.63.0-nightly
LLVM version: 14.0.5
```

### Additional Labels

_No response_

Rqnsom:
fixes  #8948 

Whenever suggestion for this lint does not fit in one line,
legacy solution has some unexpected/unhandled behavior:
lint will then generate two help messages which seem to be shown in the wrong order.
The second help message in that case will contain the suggestion.

The first help message always refers to a suggestion message,
and **it should adapt** depending on the location of the suggestion:
- inline suggestion within the error/warning message
- suggestion separated into a second help text

This is my first contribution here, so I hope I didn't miss anything for creating this PR.

changelog: fix [`manual_flatten`] help texts order

### DIFF
--- a/clippy_lints/src/loops/manual_flatten.rs
+++ b/clippy_lints/src/loops/manual_flatten.rs
@@ -51,22 +51,32 @@ pub(super) fn check<'tcx>(
                 _ => ""
             };
 
+            let sugg = format!("{arg_snippet}{copied}.flatten()");
+
+            // If suggestion is not a one-liner, it won't be shown inline within the error message. In that case,
+            // it will be shown in the extra `help` message at the end, which is why the first `help_msg` needs
+            // to refer to the correct relative position of the suggestion.
+            let help_msg = if sugg.contains('\n') {
+                "remove the `if let` statement in the for loop and then..."
+            } else {
+                "...and remove the `if let` statement in the for loop"
+            };
+
             span_lint_and_then(
                 cx,
                 MANUAL_FLATTEN,
                 span,
                 &msg,
                 |diag| {
-                    let sugg = format!("{}{}.flatten()", arg_snippet, copied);
                     diag.span_suggestion(
                         arg.span,
                         "try",
                         sugg,
-                        Applicability::MaybeIncorrect,
+                        applicability,
                     );
                     diag.span_help(
                         inner_expr.span,
-                        "...and remove the `if let` statement in the for loop",
+                        help_msg,
                     );
                 }
             );
--- a/tests/ui/manual_flatten.rs
+++ b/tests/ui/manual_flatten.rs
@@ -106,4 +106,20 @@ fn main() {
     for n in vec![Some(1), Some(2), Some(3)].iter().flatten() {
         println!("{}", n);
     }
+
+    run_unformatted_tests();
+}
+
+#[rustfmt::skip]
+fn run_unformatted_tests() {
+    // Skip rustfmt here on purpose so the suggestion does not fit in one line
+    for n in vec![
+        Some(1),
+        Some(2),
+        Some(3)
+    ].iter() {
+        if let Some(n) = n {
+            println!("{:?}", n);
+        }
+    }
 }
--- a/tests/lint_message_convention.rs
+++ b/tests/lint_message_convention.rs
@@ -42,6 +42,7 @@ impl Message {
             r".*remove .*the return type...",
             r"note: Clippy version: .*",
             r"the compiler unexpectedly panicked. this is a bug.",
+            r"remove the `if let` statement in the for loop and then...",
         ])
         .unwrap();
 
--- a/tests/ui/manual_flatten.stderr
+++ b/tests/ui/manual_flatten.stderr
@@ -167,5 +167,33 @@ LL | |             println!("{:?}", n);
 LL | |         }
    | |_________^
 
-error: aborting due to 8 previous errors
+error: unnecessary `if let` since only the `Some` variant of the iterator element is used
+  --> $DIR/manual_flatten.rs:116:5
+   |
+LL | /     for n in vec![
+LL | |         Some(1),
+LL | |         Some(2),
+LL | |         Some(3)
+...  |
+LL | |         }
+LL | |     }
+   | |_____^
+   |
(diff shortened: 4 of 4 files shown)


clippy_lints/src/loops/manual_flatten.rs
```diff
@@ -51,13 +51,23 @@ pub(super) fn check<'tcx>(
...
+                "...and remove the `if let` statement in the for loop"
+            };
+
             span_lint_and_then(
                 cx,
                 MANUAL_FLATTEN,
                 span,
                 &msg,
                 |diag| {
-                    let sugg = format!("{}{}.flatten()", arg_snippet, copied);
```
Rqnsom, line 60: I moved these variables outside the closure, to save a few indentation blocks in variable declarations, and also to make the closure a bit shorter.
I hope that's alright.

tests/ui/manual_flatten.rs
```diff
@@ -106,4 +106,24 @@ fn main() {
...
+        fn get_some___________________________________________________________________(
+            p1: Option<i32>,
+        ) -> Vec<Option<i32>> {
+            vec![p1]
+        }
+    }
+
+    // Make sure length of the next line with arguments is above 120, so the
+    // suggestion provided by clippy cannot be shown inline within the error message
+    for n in LongConstructorForSomeArray::get_some___________________________________________________________________(
```
Rqnsom, line 122: I'm really not sure if there is some smarter method to achieve multi-line suggestion here.
My solution is simply `bruteforce` in a way.

Rqnsom: Ah right!
Let's rather use  `#[rustfmt::skip]`, it feels like more elegant solution here.
Thanks for the suggestion!

Alexendoo: Yep that is good thanks, a single commit is usually preferred

clippy_lints/src/loops/manual_flatten.rs
```diff
@@ -66,7 +76,7 @@ pub(super) fn check<'tcx>(
                     );
```
Alexendoo, line 76: Not an issue with your PR, but while reading yours I noticed the line above this should be `applicability` rather than `Applicability::MaybeIncorrect`. If you could fix that while you're working in this area that'd be lovely

Rqnsom: Okay, no problem. I'm always glad I could help here.

clippy_lints/src/loops/manual_flatten.rs
```diff
@@ -51,13 +51,23 @@ pub(super) fn check<'tcx>(
                 _ => ""
             };
 
+            let sugg = format!("{arg_snippet}{copied}.flatten()");
+
+            // If suggestion is not a one-liner, it won't be shown inline within the error message. In that case,
+            // it will be shown in the extra `help` message at the end, which is why the first `help_msg` needs
+            // to refer to the correct relative position of the suggestion.
+            let help_msg = if sugg.contains('\n') {
+                "remove the `if let` statement in the for loop and then"
```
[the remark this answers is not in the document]
Rqnsom: > I liked the trailing ... from your example in #8948 (comment) 

I also liked that version more, but I forgot to mention why I changed it here.
When I've ran locally `cargo test`, I've got a failure from `lint_message_convetion` test:
```
running 1 test
test lint_message_convention ... FAILED

failures:

---- lint_message_convention stdout ----
error: the test 'tests/ui/manual_flatten.stderr' contained the following nonconforming lines :
help: remove the `if let` statement in the for loop and then...

Lint message should not start with a capital letter and should not have punctuation at the end of the message unless multiple sentences are needed.
```

Not sure if we could make some workaround for this.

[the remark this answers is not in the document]
Alexendoo: It seems the workaround is to add the message to https://github.com/rust-lang/rust-clippy/blob/07ab0e87417e8cea018a34c7cd46fde0693d32a5/tests/lint_message_convention.rs#L34-L46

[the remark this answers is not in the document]
Rqnsom: Okay, I could have checked the test to see if there were such exceptions. Sorry about that...

Now everything passes, thank you!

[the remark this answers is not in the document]
Alexendoo: No worries, I checked the test curious to see how it was implemented, didn't actually expect to see such a literal list of exceptions

Alexendoo (approved):

Alexendoo:
That's great, thanks! Welcome to Clippy

@bors r+

PR #9157, opened 2022-07-12
update pull request template

Rqnsom:
Improved suggestion for formatting lint names in the PR template to use this format:
[`lint_name`]

changelog: none

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,11 +3,15 @@ Thank you for making Clippy better!
 We're collecting our changelog from pull request descriptions.
 If your PR only includes internal changes, you can just write
 `changelog: none`. Otherwise, please write a short comment
-explaining your change. Also, it's helpful for us that
-the lint name is put into brackets `[]` and backticks `` ` ` ``,
-e.g. ``[`lint_name`]``.
+explaining your change.
 
-If your PR fixes an issue, you can add "fixes #issue_number" into this
+It's also helpful for us that the lint name is put within backticks (`` ` ` ``),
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,11 +3,13 @@ Thank you for making Clippy better!
...
 `changelog: none`. Otherwise, please write a short comment
-explaining your change. Also, it's helpful for us that
-the lint name is put into brackets `[]` and backticks `` ` ` ``,
-e.g. ``[`lint_name`]``.
+explaining your change.
 
-If your PR fixes an issue, you can add "fixes #issue_number" into this
+It's also helpful for us that the lint name is put within backticks (`` ` ` ``),
+and then encapsulated by square brackets (`[]`),
+e.g. [`lint_name`].
```
Rqnsom, line 10: >   10:6-10:19  warning  Found reference to undefined definition  no-undefined-references  remark-lint

This is the failure.
I think the problem is that `remark` tool fails because it is trying to use square brackets as a part of a hyperlink for referencing.

> You can create an inline link by wrapping link text in brackets [ ], and then wrapping the URL in parentheses ( ).

Rqnsom: > Unfortunately that makes it not a link, so it wouldn't render right in the changelog

Aha, but I didn't expect it to be used with `\` characters except only in the template text... but then people might get confused whether they ought to use `\` inside their actual PR text and title. I see, I see.

> I like the suggestion on putting the example in triple backticks. That would resolve this and would look in the parsed and the raw version.

Agreed!

flip1995:
@bors r+

Thanks!

PR #9161, opened 2022-07-12
Move format_push_string to restriction

### LINKED ISSUE #9077: format_push_string could use some more information
### Description

`clippy::format_push_string` currently outputs the following:

(the following is part of a fork of [`human_panic`](https://github.com/rust-cli/human-panic))

```
error: `format!(..)` appended to existing `String`
   --> my/project/file/mod.rs:119:27
    |
119 |           Some(location) => explanation.push_str(&format!(
    |  ___________________________^
120 | |             "Panic occurred in file '{}' at line {}\n",
121 | |             location.file(),
122 | |             location.line()
123 | |         )),
    | |__________^
    |
    = note: `-D clippy::format-push-string` implied by `-D clippy::all`
    = help: consider using `write!` to avoid the extra allocation
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string

error: could not compile `project_name` due to previous error
```

I think this description could use a bit more information regarding the difference between `format!` and `write!`, as `write!` returns a `Result`, which the [linked example](https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string) instructs you to ignore:

```rust
let _ = write!(s, "0x{:X}", 1024);
```

The issue with this is that if you're not familiar with `write!`, you don't know which errors you're ignoring by making this change, and if you use `-D clippy::all` (as e.g. we do in my project) you have to either make this change or use `#[allow(clippy::format_push_string)]` for your code to compile. This isn't ideal as you may end up making a decision without all the needed information.

If you look into `write!`, you'll find that the `write_xyz` methods have the following description:

>[Errors](https://doc.rust-lang.org/std/fmt/trait.Write.html#errors-1)
>This function will return an instance of [Error](https://doc.rust-lang.org/std/fmt/struct.Error.html) on error.

Which doesn't really give you any more information. (edit: apparently this is meant to reflect any implementation rather than the default implementation, which is why it's a bit vague)

`write_fmt`, used by `write!` (on `$dst` which may mean it's using something other than the linked documentation as I'm not familiar with what `$dst` is in this context), doesn't have an `# Errors` section. 

If you continue going down the rabbit hole, `write_fmt` calls `write`, which also does not have an `# Errors` section. It itself calls `write_str`, and has an `unsafe` block that returns a `Result` as well.

My point isn't to complain or criticize, but rather to point out that if you're not familiar with the intricacies of `write!` the lint info currently doesn't give you a good idea of what potential errors you're ignoring by making this switch if using it as directed. Even if you do add handling code and change the function you used `format!` in to return a `Result`, the result (no pun intended) is that you now have a new failure scenario you don't know enough about.

My assumption is that `format!` calls `write!` somewhere and ignores the error and that's the reason why they were deemed equivalent, but I'd add that to the description if that's the case.

Thanks!

### Version

```text
binary: rustc
commit-date: 2022-06-27
host: aarch64-apple-darwin
release: 1.62.0
LLVM version: 14.0.5
```

### Additional Labels

_No response_

Victor-N-Suadicani:
Fixes #9077 (kinda) by moving the lint to the restriction group. As I noted in that issue, I think the suggested change is too much and as the OP of the issue points out, the ramifications of the change are not necessarily easily understood. As such I don't think the lint should be enabled by default.

changelog: [`format_push_string`]: moved to restriction (see #9077).

### DIFF
--- a/clippy_lints/src/format_push_string.rs
+++ b/clippy_lints/src/format_push_string.rs
@@ -29,7 +29,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.62.0"]
     pub FORMAT_PUSH_STRING,
-    pedantic,
+    restriction,
     "`format!(..)` appended to existing `String`"
 }
 declare_lint_pass!(FormatPushString => [FORMAT_PUSH_STRING]);
--- a/clippy_lints/src/format_push_string.rs
+++ b/clippy_lints/src/format_push_string.rs
@@ -29,7 +29,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.62.0"]
     pub FORMAT_PUSH_STRING,
-    perf,
+    pedantic,
     "`format!(..)` appended to existing `String`"
 }
 declare_lint_pass!(FormatPushString => [FORMAT_PUSH_STRING]);
--- a/clippy_lints/src/format_push_string.rs
+++ b/clippy_lints/src/format_push_string.rs
@@ -14,6 +14,12 @@ declare_clippy_lint! {
     /// ### Why is this bad?
     /// Introduces an extra, avoidable heap allocation.
     ///
+    /// ### Known problems
+    /// `format!` returns a `String` but `write!` returns a `Result`.
+    /// Thus you are forced to ignore the `Err` variant to achieve the same API.
+    ///
+    /// While using `write!` in the suggested way should never fail, this isn't necessarily clear to the programmer.
+    ///
     /// ### Example
     /// ```rust
     /// let mut s = String::new();
--- a/clippy_lints/src/lib.register_all.rs
+++ b/clippy_lints/src/lib.register_all.rs
@@ -71,7 +71,6 @@ store.register_group(true, "clippy::all", Some("clippy_all"), vec![
     LintId::of(format_args::TO_STRING_IN_FORMAT_ARGS),
     LintId::of(format_impl::PRINT_IN_FORMAT_IMPL),
     LintId::of(format_impl::RECURSIVE_FORMAT_IMPL),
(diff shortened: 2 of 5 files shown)


yoav-lavi:
Hey, author of #9077.

Thank you!

I'd like to note that using `clippy::pedantic` is (anecdotally) popular from what I've seen, so perhaps restrictions?

Victor-N-Suadicani:
I totally forgot that restriction even existed and I agree that pedantic wasn't the best fit. I've moved it to restriction now :)

flip1995 (commented):
Thanks! I agree with the reasoning in the linked issue and that it should be moved to `restriction`. Can you please add a summary of this issue to a new `### Known Issues` section in the link documentation after the `why is this bad` section? This should just summarize, that changing the code to `write` will lead to usage of a fallible API.

Jarcho:
Just going to add a couple points here.
* `format!` can fail for exactly the same reason `write!` to a string can fail. It just calls `expect` internally.
* The only way this can return an error is if a `Display` implementation creates the error. This would be a violation of the stated [API](https://doc.rust-lang.org/std/fmt/index.html#formatting-traits)

I see this as a combination of the lint docs not being enough to justify the change, and an unfortunate side-effect of having to make the `unwrap` call explicit.

yoav-lavi:
@Jarcho 
>format! can fail for exactly the same reason write! to a string can fail. It just calls expect internally.
That was my assumption, the lint doesn't make that clear enough though (with the added issue of encouraging ignoring an error)

Victor-N-Suadicani:
@flip1995 Added the known problems section :)

flip1995:
@bors r+

Thanks!

PR #9169, opened 2022-07-13
Use `LazyLock` for `lint_message_convention` regexes

Alexendoo:
They were being recompiled for `Message::new` call, for me this shaves 7s off the time it takes to run the test. Also removes a redundant exception from the list and joins the various `message...` exceptions into one

changelog: none


flip1995:
Nice!

@bors r+

Issue #31, opened 2015-02-18
Box<Vec<T>> warning triggered spuriously


maurer:
The `Box<Vec<T>>` warning appears to be being triggered when I box a closure (necessary in this case to make the struct `Sized`).

```
src/native_types.rs:93:13: 93:65 note: Vec<T> is already on the heap, Box<Vec<T>> makes an extra allocation
src/native_types.rs:93   pub run : Box<Fn(Vec<HValue>) -> Vec<HValue> + 'static + Send>
                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

pythonesque:
This is (or should be) fixed now.

llogiq:
I'll add a test to `tests/compile-fail/box_vec.rs`. If successful, I'll close.

llogiq:
Yep, it's no longer triggered. Closing.

PR #9170, opened 2022-07-13
[`box_collection`]: raise warn for all std collections

Rqnsom:
So far, only [`Vec`, `String`, `HashMap`] were considered.

Extend collection checklist for this lint with:
- `HashSet`
- `VecDeque`
- `LinkedList`
- `BTreeMap`
- `BTreeSet`
- `BinaryHeap`

changelog: [`box_collection`]: raise warn for all std collections

### DIFF
--- a/tests/ui/box_collection.rs
+++ b/tests/ui/box_collection.rs
@@ -6,7 +6,7 @@
     unused
 )]
 
-use std::collections::HashMap;
+use std::collections::{BTreeMap, BTreeSet, BinaryHeap, HashMap, HashSet, LinkedList, VecDeque};
 
 macro_rules! boxit {
     ($init:expr, $x:ty) => {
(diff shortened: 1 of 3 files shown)


tests/ui/box_collection.rs
```diff
@@ -18,7 +18,7 @@ fn test_macro() {
     boxit!(Vec::new(), Vec<u8>);
 }
 
-fn test(foo: Box<Vec<bool>>) {}
+fn test1(foo: Box<Vec<bool>>) {}
 
 fn test2(foo: Box<dyn Fn(Vec<u32>)>) {
     // pass if #31 is fixed
```
Rqnsom, line 24: `    // pass if #31 is fixed`

Not sure what this means here. Is it referring to issue #31?

Jarcho: That looks like the reason.

Jarcho:
Thank you.

@bors r+

tests/ui/box_collection.rs
Rqnsom: Maybe we can remove that comment within some future commit.

PR #9172, opened 2022-07-13
Rename rustcSource in contributing documentation

Guilherme-Vasconcelos:
According to [rust-analyzer docs](https://rust-analyzer.github.io/manual.html#toolchain:~:text=rust%2Danalyzer.rustc.source), rustcSource has been renamed to rustc.source.

changelog: none

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,11 +114,11 @@ As of [#6869][6869], [`rust-analyzer`][ra_homepage] can understand that Clippy u
 using `extern crate` when `package.metadata.rust-analyzer.rustc_private` is set to `true` in Clippy's `Cargo.toml.`
 You will require a `nightly` toolchain with the `rustc-dev` component installed.
 Make sure that in the `rust-analyzer` configuration, you set
-```
-{ "rust-analyzer.rustcSource": "discover" }
+```json
+{ "rust-analyzer.rustc.source": "discover" }
 ```
 and
-```
+```json
 { "rust-analyzer.updates.channel": "nightly" }
 ```
 You should be able to see information on things like `Expr` or `EarlyContext` now if you hover them, also


CONTRIBUTING.md
```diff
@@ -112,13 +112,17 @@ Just make sure to remove the dependencies again before finally making a pull req
 ### Rust Analyzer
 As of [#6869][6869], [`rust-analyzer`][ra_homepage] can understand that Clippy uses compiler-internals
 using `extern crate` when `package.metadata.rust-analyzer.rustc_private` is set to `true` in Clippy's `Cargo.toml.`
+
 You will require a `nightly` toolchain with the `rustc-dev` component installed.
+Notice that [rust-toolchain](./rust-toolchain) pins a specific nightly version, but for the
+configuration below to work you will need `rustc-dev` for the **regular** nightly.
```
dswij, line 118: I might be confusing the wording here. Is this trying to say that: it's ok if `rustc-dev` version used is not the same as the one in `rust-toolchain`?

I guess this being placed after L116 is what making me confused.

flip1995: This is trying to say that the `rustc-dev` version that is automatically installed through the `rust-toolchain` file isn't enough and that you need a `rustc-dev` component in a `nightly` toolchain.

However, I don't think this is the case, because it works for me and I don't have a `rustc-dev` component in my `nightly` toolchain.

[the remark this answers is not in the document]
Guilherme-Vasconcelos: Hey!

I think I must have made a mistake during my first setup. Today I tried uninstalling `rustc-dev` from the nightly toolchain, did a `cargo clean`, recompiled everything and my rust-analyzer is working perfectly fine!

It may have something to do with the fact that for the past two days I experimented a little bit with some different ways of compiling clippy. So maybe at some point rust-analyzer was using the regular nightly instead of the pinned one. Another hypothesis that comes to mind is that I activated my rust-analyzer before creating the configuration mentioned in the documentation, so maybe in the first time I compiled it it wasn't using the correct settings and this caused problems.

But either way, thanks for pointing out that it wasn't needed. I can confirm that, now that I set everything up correctly, it works with just the pinned nightly's rustc-dev.

Guilherme-Vasconcelos: Thanks! I undid the change. I explained a little bit more [in the other thread](https://github.com/rust-lang/rust-clippy/pull/9172#discussion_r921580969), but it was likely due to some mistake I made during the initial setup.

flip1995:
Thanks for updating this documentation! This was in the back of my mind for at least a week now.

@bors r+

xFrednet:
I feel like bors is stuck on this PR for some reason, I'll r- it for now to see if that fixes it

@bors r-

xFrednet:
After the r- bors could test the other PRs as normal. @flip1995 could it be that this doesn't trigger the CI since it's only an `.md` file change? Otherwise, I'm not sure what could confuse bors here. :thinking: I'll leave this r-ed, feel free to r+ it again :upside_down_face:

xFrednet:
It looks like the PR is still stuck. Let's see if try works.

@bors try

xFrednet:
:joy: I see

@bors r-

Guilherme-Vasconcelos:
I think it glitched because I did a commit --amend, you can see at the top I pushed commit https://github.com/rust-lang/rust-clippy/commit/3106bbc35a4966d0ba2e6a5f86afaa75a3a76c38 to https://github.com/rust-lang/rust-clippy/commit/12da7a162a56c6b6f1ff591c837ef77b7745bb40 (just to delete an extra unintentional space), but once flip approved it, it still says 3106bbc was approved

xFrednet:
Could very well be. Could you maybe try to rebase and push again? Then the hash should change:)

xFrednet:
Alright, let's see if that did the trick :upside_down_face: 

@bors r=flip1995

PR #9174, opened 2022-07-14
Rustup

flip1995:
r? @ghost

changelog: none


flip1995:
insufficient disk space on the windows runner. Please be a one-time thing :pray: 

@bors retry

flip1995:
```
  = note: D:\a\rust-clippy\rust-clippy\target\debug\deps\clippy_driver.exe : fatal error LNK1180: insufficient disk space to complete link
```

image

xFrednet:
Hmmm, windows, haven't we had this problem before? I think we had to ping the infra team, if I'm not mistaken

flip1995:
I don't think so. But then again, if there are problems, it is with windows and I usually suppress windows memories.

xFrednet:
I can't remember where I saw that message before. Zulip and GH search isn't too helpful either :thinking:. There is a long thread with a message like this on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/1.2E58.2E1.20release/near/268526756), but it seems to be focussed on a release.

> But then again, if there are problems, it is with windows and I usually suppress windows memories.

:joy: I switched to Linux about 3 - 4 years ago and haven't regretted it one bit.

flip1995:
Sadly I don't have time to debug this today. Maybe I can look into it this evening

mikerite:
It seems related to generating the pdb file. Maybe try testing with `--release`.

Jarcho:
Something is going wrong here. Just built this locally and the target dir is almost 12GB after running `cargo test`. `clippy_driver.exe` itself is 1GB. For reference current master has the target dir at 4.3GB with `clippy_driver.exe` at a reasonable 17MB. `clippy_utils` rlib is also ~800MB on this pr.

Both use about 2GB for incremental stuff.

flip1995:
Found a rust issue for this: https://github.com/rust-lang/rust/issues/99143

flip1995:
Blocking this for now, hoping for a fix on the rustc side soon-ish. I will work on a workaround if no fix is in sight by the end of the week.

mati865:
You might have some luck with (temporarily) replacing windows-msvc with windows-gnu. It's another Tier 1 target which AFAIK is not affected by this issue.

flip1995:
@bors try

This should now work with the fix merged just before the nightly got released.

@Jarcho Can you please double check, that this indeed works on windows and then r+ this PR?

Jarcho:
Thank you.

@bors r+

PR #9182, opened 2022-07-15
Add `--type` flag to `dev new_lint`

Serial-ATA:
changelog: none

This only works with directories, `--type={copies, doc, ...}` won't work. Should they be counted?

### DIFF
--- a/book/src/development/adding_lints.md
+++ b/book/src/development/adding_lints.md
@@ -10,6 +10,10 @@ because that's clearly a non-descriptive name.
 - [Adding a new lint](#adding-a-new-lint)
   - [Setup](#setup)
   - [Getting Started](#getting-started)
+    - [Defining Our Lint](#defining-our-lint)
+      - [Standalone](#standalone)
+      - [Specific Type](#specific-type)
+      - [Tests Location](#tests-location)
   - [Testing](#testing)
     - [Cargo lints](#cargo-lints)
   - [Rustfix tests](#rustfix-tests)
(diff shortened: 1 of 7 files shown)


xFrednet (commented):
Hey, this PR has not been forgotten, my two weeks have just been pretty full, with rl stuff. I'm away over the weekend, but should be able to give you a review next week. I've added one comment about something that stuck out to me :)

clippy_lints/src/operators/mod.rs
```diff
@@ -25,6 +21,10 @@ mod ptr_eq;
 mod self_assignment;
 mod verbose_bit_mask;
 
+use rustc_hir::{Body, Expr, ExprKind, UnOp};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
```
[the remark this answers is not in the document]
Serial-ATA: That was a manual change. I just switched them to have the modules at the top so I can add new entries without searching around.

clippy_dev/src/new_lint.rs
```diff
@@ -312,6 +297,247 @@ fn get_lint_file_contents(lint: &LintData<'_>, enable_msrv: bool) -> String {
...
+                        "EarlyLintPass" => Some("EarlyContext"),
+                        _ => continue,
+                    };
+                }
+            },
+            _ => {},
+        }
+    }
+
+    drop(iter);
```
xFrednet, line 479: I don't quite get, why this drop is needed. Is here any advantage in having this?

clippy_dev/src/new_lint.rs
```diff
@@ -38,25 +39,35 @@ pub fn create(
...
-    add_lint(&lint, msrv).context("Unable to add lint to clippy_lints/src/lib.rs")
+
+    if lint.ty.is_none() {
+        add_lint(&lint, msrv).context("Unable to add lint to clippy_lints/src/lib.rs")?;
+    }
+
+    Ok(())
 }
 
 fn create_lint(lint: &LintData<'_>, enable_msrv: bool) -> io::Result<()> {
```
xFrednet, line 63: Just an optional idea. Maybe it would be nice, if this would print the generated file name to the console to let the user know where they should start implementing the lint. Something simple like: `println!("Generated lint file: {}")`

clippy_dev/src/new_lint.rs
```diff
@@ -312,6 +297,247 @@ fn get_lint_file_contents(lint: &LintData<'_>, enable_msrv: bool) -> String {
...
+                pub(super) fn check(cx: &{context_import}) {{
+                    todo!();
+                }}
+           "#,
+            context_import = context_import,
+            name_upper = name_upper,
+        );
+    }
+
+    write_file(lint_file_path, lint_file_contents)?;
```
xFrednet, line 395: This function creates the lint declaration, the lint file and adds the lint to the lint pass lints. However, it doesn't add the call to `lint_name::check` in the lint pass. This is okay, since there is no way to implement this for the different types we have. However, I believe this can be confusing for new users. I think a message about this would be good, something simple like: ``println!("Please add a call to `lint_name::check`  in `clippy_lints/ty/mod.rs`")``

clippy_dev/src/new_lint.rs
```diff
@@ -312,6 +297,247 @@ fn get_lint_file_contents(lint: &LintData<'_>, enable_msrv: bool) -> String {
...
+            }}
+        "#},
+        version = get_stabilization_version(),
+        name_upper = name_upper,
+        category = category,
+    )
+}
+
+fn generate_from_ty(lint: &LintData<'_>, enable_msrv: bool, ty: &str) -> io::Result<()> {
+    if ty == "cargo" {
```
xFrednet, line 329: Optional NIT: The opposite could also be checked, that other types don't have the cargo category. Currently, `cargo dev new_lint --name test1 --type methods -c cargo` works just fine.

xFrednet (commented):
The implementation looks good overall, I found some small nits, where some might be too nitpicky. :sweat_smile:

Could you maybe also be so kind to update the related docs in `book/src/development/adding_lints.md`? Mentioning the type option is probably enough :upside_down_face:

clippy_dev/src/new_lint.rs
xFrednet: Never mind, I checked it locally and this is needed, due to mutable borrows

clippy_dev/src/new_lint.rs
Serial-ATA: Good call :slightly_smiling_face:.

clippy_dev/src/new_lint.rs
Serial-ATA: I'll also make `--category=cargo` implicitly set `--type=cargo` if there isn't one specified. Right now it treats cargo lints as any other lint, and makes a file in `clippy_lints/src` :smile:.

book/src/development/adding_lints.md
```diff
@@ -36,17 +40,37 @@ See the [Basics](basics.md#get-the-code) documentation.
...
+This command will create a new file: `clippy_lints/src/foo_functions.rs`, as well
+as [register the lint](#lint-registration).
+
+### Specific Type
+Command: `cargo dev new_lint --name=foo_functions --type=functions --category=pedantic`
+
+This command will create a new file: `clippy_lints/src/{type}/foo_functions.rs`.
+
+Notice how this command has a `--type` flag instead of `--pass`. Unlike a standalone
+definition, this lint won't be registered, so a pass doesn't apply.
```
xFrednet, line 65:
```suggestion
Notice how this command has a `--type` flag instead of `--pass`. Unlike a standalone
definition, this lint has to be manually registered in the type lintpass, which can be found in `clippy_lints/src/{type}/mod.rs`.
```

xFrednet (approved):

xFrednet:
I suggest rephrasing one paragraph in the documentation. Otherwise, everything looks good to me. If you agree with the change, you can comment this and merge it with `(at)bors r=xFrednet`

Thank you for this addition, it was something I wanted to see for a longer time!

@bors delegate+

PR #9187, opened 2022-07-16
Add lint recommending using `std::iter::once` and `std::iter::empty` 

### LINKED ISSUE #9186: Lints recommeding to use `std::iter::once` and `std::iter::empty`
### What it does

It seems suitable to separate it in two lints, one for `iter::once` and one for `iter::empty`. They'd look at patterns such as `[foo].into_iter()` and `[].into_iter()` and suggest using `once` and  `empty` instead.

### Lint Name

iter_once and iter_empty

### Category

pedantic

### Advantage

`once` and `empty` take less memory than their the arrays or slice iterators for the corresponding use case.

### Drawbacks

Depending on the use case, the fix can cause type issues, for example if the iterator is then passed to a function that expects an `std::slice::Iter` and not a genreric `impl Iterator`

### Example

```rust
let a = [foo].iter();
let b = [bar].iter_mut();
let c = [baz].into_iter();
let d = Some(foo).iter();
let e = Some(bar).iter_mut();
let f = Some(baz).into_iter();
```
Could be written as:
```rust
use std::iter;
let a = iter::once(&foo);
let b = iter::once(&mut bar);
let c = iter::once(bar);
let d = iter::once(&foo);
let e = iter::once(&mut bar);
let f = iter::once(bar);
```

and 

```rust
let a = [].iter();
let b = [].iter_mut();
let c = [].into_iter();
let d = None.iter();
let e = None.iter_mut();
let f = None.into_iter();
```

could be written as:

```rust
use std::iter;
let a = iter::empty();
let b = iter::empty();
let c = iter::empty();
let d = iter::empty();
let e = iter::empty();
let f = iter::empty();
```

sgued:
```
changelog: [`iter_once`]: add new lint
changelog: [`iter_empty`]: add new lint
```

fixes #9186 

- \[x] Run `cargo dev fmt`

The lint doesn't really follow the naming conventions. I don't have any better idea so I'm open to suggestions.


clippy_lints/src/iter_once_empty.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// - `[].iter()`
+    /// - `[].iter_mut()`
+    /// - `[].into_iter()`
+    /// - `None.iter()`
+    /// - `None.iter_mut()`
+    /// - `None.into_iter()`
+    ///
+    /// ### Why is this bad?
+    ///
+    /// It is simpler to use the empty function from the standard library:
```
Serial-ATA, line 68: This should also mention `iter::once`.

clippy_lints/src/iter_once_empty.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// ### Example
+    ///
+    /// ```rust
+    /// use std::{array, slice, option};
+    /// let a: slice::Iter<i32> = [].iter();
+    /// let b: slice::IterMut<i32>= [].iter_mut();
+    /// let c: array::IntoIter<i32, 0>= [].into_iter();
+    /// let d: option::Iter<i32> = None.iter();
+    /// let e: option::IterMut<i32> = None.iter_mut();
+    /// let f: option::IntoIter<i32> = None.into_iter();
```
Serial-ATA, line 79: You don't need an exhaustive list here, you can just put a few examples.

clippy_lints/src/iter_once_empty.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// ### Example
+    ///
+    /// ```rust
+    /// let a = [123].iter();
+    /// let b = [123].iter_mut();
+    /// let c = [123].into_iter();
+    /// let d = Some(123).iter();
+    /// let e = Some(123).iter_mut();
+    /// let f = Some(123).into_iter();
```
Serial-ATA, line 32: You don't need an exhaustive list here, you can just put a few examples.

Serial-ATA (changes requested):

clippy_lints/src/iter_once_empty.rs
sgued: Why? The two lints are distinct.

clippy_lints/src/iter_once_empty.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// The type of the resulting iterator might become incompatible with its usage
+    #[clippy::version = "1.64.0"]
+    pub ITER_EMPTY,
+    pedantic,
+    "Iterator for empty array"
+}
+
+declare_lint_pass!(IterOnceEmpty => [ITER_ONCE, ITER_EMPTY]);
+
+impl EarlyLintPass for IterOnceEmpty {
```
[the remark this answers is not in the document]
sgued: If the option is not constructed just before calling `iter` this lint shouldn't apply.

clippy_lints/src/iter_once_empty.rs
Serial-ATA: Ignore this. That was me being too hasty :sweat_smile:.

clippy_lints/src/iter_once_empty.rs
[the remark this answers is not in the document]
Serial-ATA: Hypothetically, someone could make their own `Some` and `None`. This should only check for `Option`. [Here's an example](https://github.com/rust-lang/rust-clippy/blob/4562dd0a8e45af4fc2a49951df07ce5d3f1108ec/clippy_lints/src/if_then_some_else_none.rs#L77-L79) of how you can check for this.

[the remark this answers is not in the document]
sgued: Ok, I changed it to both check the path **and** check the type so we're sure that it's not something else that's also called `Some` for some reason

clippy_lints/src/iter_once_empty.rs
```diff
@@ -0,0 +1,180 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
```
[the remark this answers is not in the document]
sgued: Done

clippy_lints/src/iter_once_empty.rs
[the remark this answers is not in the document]
sgued: Ok, I didn't see your message for some reason. So I guess I implemented the right thing

clippy_lints/src/methods/mod.rs
```diff
@@ -2293,6 +2294,83 @@ declare_clippy_lint! {
...
+    /// ### What it does
+    ///
+    /// Checks for usage of:
+    ///
+    /// - `[].iter()`
+    /// - `[].iter_mut()`
+    /// - `[].into_iter()`
+    /// - `None.iter()`
+    /// - `None.iter_mut()`
+    /// - `None.into_iter()`
```
flip1995, line 2345: Same here.

clippy_lints/src/methods/mod.rs
```diff
@@ -2293,6 +2294,83 @@ declare_clippy_lint! {
...
+    /// Use instead:
+    /// ```rust
+    /// use std::iter;
+    /// let a: iter::Empty<i32> = iter::empty();
+    /// let b: iter::Empty<i32> = iter::empty();
+    /// ```
+    ///
+    /// ### Known problems
+    ///
+    /// The type of the resulting iterator might become incompatible with its usage
```
flip1995, line 2367: I'm a little concerned about this. From how many allows are necessary in Clippy already, I think this is a bigger problem, that has to be addressed before we can merge this.

clippy_lints/src/methods/iter_once_empty.rs
```diff
@@ -0,0 +1,70 @@
...
+            _ => return,
+        };
+        span_lint_and_sugg(
+            cx,
+            ITER_EMPTY,
+            expr.span,
+            msg,
+            "try",
+            "std::iter::empty()".to_string(),
+            Applicability::Unspecified,
```
flip1995, line 67: Applicability is `MaybeIncorrect` for this lint.

clippy_lints/src/methods/iter_once_empty.rs
```diff
@@ -0,0 +1,70 @@
...
+            "into_iter" => "this `into_iter` call can be replaced with std::iter::empty",
+            _ => return,
+        };
+        span_lint_and_sugg(
+            cx,
+            ITER_EMPTY,
+            expr.span,
+            msg,
+            "try",
+            "std::iter::empty()".to_string(),
```
flip1995, line 66: The suggestion creation should depend on if it is a `no_std` crate or not. We have other lints that already do that, so you can reuse code from there.

clippy_lints/src/methods/iter_once_empty.rs
```diff
@@ -0,0 +1,70 @@
...
+            "into_iter" => (
+                format!("std::iter::once({})", snippet(cx, i.span, "...")),
+                "this `into_iter` call can be replaced with std::iter::once",
+            ),
+            _ => return,
+        };
+        span_lint_and_sugg(cx, ITER_ONCE, expr.span, msg, "try", sugg, Applicability::Unspecified);
+    } else {
+        let msg = match method_name {
+            "iter" => "this `iter call` can be replaced with std::iter::empty",
```
flip1995, line 55: The message should describe the problem, not the solution. This is what the suggestion is for. So this should be something along the lines of "calling `iter` on an empty collection". With this change, all of this match block should collapse into a single `format!` macro.

flip1995 (changes requested):

clippy_lints/src/methods/mod.rs
sgued: Does clippy have a way to know whether a suggestion type-checks?
Maybe we could limit it to expressions within a `chain`, which would likely lead to less false positives but not 0. It would also create false negatives in cases like:

```rust
let iter = if some_condition {
    [a].into_iter()
} else {
    [b].into_iter()
}
  
iter.chain(other_iter)
```

flip1995: Do you have examples where applying the suggestion does not type check? Looking at all the `allow`s you added to Clippy, I don't really see why you couldn't use `iter::empty` there.

flip1995: Hm, if I'd had to guess, I would say that this fails type checking when using `iter::empty` because it can't infer the type anymore? I don't think this is a problem for `iter_once` though.

flip1995 (commented):
I'll be thinking about a a lint name before merging this.

clippy_lints/src/methods/mod.rs
sgued: Say you have something like this:

```rust
let iter = match smth {
    "a" => [1,2].iter(),
    "b" => [3].iter(),
    _ => [].iter(),
};
```

The first branch forces `iter` to be of type `slice::Iter<i32>` but suggestions would be to change it to:

```rust
let iter = match smth {
    "a" => [1,2].iter(),
    "b" => iter::once(&3),
    _ => iter::empty(),
};
```

so now `iter` must be at the same time `Once<i32>`, `slice::Iter<i32>` and `Empty<i32>`

flip1995: Hm, limiting this to iter chains also doesn't solve this, e.g.:

```rust
let iter = match smth {
    "a" => [[1],[2]].iter().flatten(),
    "b" => [[3]].iter().flatten(),
    _ => [].iter().flatten(),
};
    
let iter = match smth {
    "a" => [[1],[2]].iter().flatten(),
    //"b" => [[3]].iter().flatten(),
    "b" => std::iter::once([3]).flatten(),
    _ => std::iter::empty().flatten(),
};
```

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e6b3994fbecbb6b4ab4d9a369c9328db)

I don't really have a solution to this... I like the idea of the lint, but think that this is too big of a FP to have it as a `pedantic` lint. I'll have to check how often this FP appears in the wild.

flip1995: Running it on the top 50 crates, those lints were only emitted on `serde_derive` and all 3 cases were fixable. Maybe I making a mountain out of a molehill here.

flip1995: I asked on Zulip about this https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60iter_empty.60.2F.60iter_once.60.20new.20lints.20clippy.239187/near/290350623

xFrednet: > Running it on the top 50 crates, those lints were only emitted on serde_derive and all 3 cases were fixable. Maybe I making a mountain out of a molehill here.

@flip1995 have you ensured that the lint was enabled when you ran it on those crates, or could it be that they haven't enabled `pedantic` lints?

flip1995: I used `@mikerite`s tool to check those lints. I'm pretty sure they were enabled. I mean one of those triggered on serde_derive.

sgued: I ran the tool on the top 200 crates to have a better idea.


The results are:
| Crate                        | Count | False Positive (FP)| Partial FP |
|:-----------------------------|:-----:|:-----:             | :--------: |
| chrono-0.4.19                |     1 |               0     |0  |
| petgraph-0.6.2               |     2 |                   1|  0|
| rayon-1.5.3                  |     1 |                   1|  0|
| serde_derive-1.0.140         |     3 |                   0| 0|
| synstructure-0.12.6          |     1 |                   0| 0 |
| tokio-1.20.0                 |     2 |                   0| 2          |
| unicode-normalization-0.1.21 |     6 |                   6| 1          |
| clippy | 7 | 2 | 0 |

*Partial FP* means the lint is right to trigger but the suggestion doesn't typecheck.

#### [petgraph](https://github.com/petgraph/petgraph/blob/master/src/graphmap.rs#L355=) and  [rayon](https://github.com/rayon-rs/rayon/blob/master/src/vec.rs#L279=) 

False positive, a `slice::Iter(Mut)` is needed

#### `serde_derive` ([1](https://github.com/serde-rs/serde/blob/master/serde_derive/src/bound.rs#L344=) [2](https://github.com/serde-rs/serde/blob/master/serde_derive/src/de.rs#L3001=) [3](https://github.com/serde-rs/serde/blob/master/serde_derive/src/de.rs#L3067=)) and [chrono](https://github.com/chronotope/chrono/blob/main/src/datetime/mod.rs#L622=)

Each time the perfect true positive for this lint

#### [tokio](https://github.com/tokio-rs/tokio/blob/master/tokio/src/net/addr.rs#L57=)

Two "false positive". In the sense that the suggestion doesn't TypeCheck but by changing some type annotations it does. This could have required a breaking change if it hadn't been on some private `impls` though.

#### [unicode-normalization](https://github.com/unicode-rs/unicode-normalization/blob/master/src/lib.rs#L173=) 
6 times the same false positive for type checking, because a trait is implemented specifically for `Option::IntoIter`, but it might be interesting for them to implement the trait also for `iter::Once`. I don't know how the crate is used and whether it would be useful though.
 
#### Clippy itself (in this PR)

The lint is fixed 5 times and ignored twice. There was one "partial false positive" which required manually writing the type of `Empty<ty::subst::GenericArg<'_>>`, which was a bit of a pain.

flip1995:
Thanks for doing the analysis (https://github.com/rust-lang/rust-clippy/pull/9187#discussion_r928122711). I nominated this PR for the meeting this evening to figure out if we want to get this lint into `pedantic` or have it in `nursery` for starters.

The meeting is at 3pm UTC: https://everytimezone.com/s/deeec323

flip1995:
So in the Clippy meeting, we decided to place those lints in `nursery` for now and lift them up to `pedantic` once at least the most obvious FPs are addressed. Jarcho left some instructions on how this could be done [here](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60iter_empty.60.2F.60iter_once.60.20new.20lints.20clippy.239187/near/290379301). However I don't want to block this PR on this.

sgued:
I see on Zulip you proposed `iter_on_empty_collection` for the name of `iter_empty`. I like it. I suggest `iter_on_singleton` for `iter_once`

clippy_lints/src/methods/mod.rs
sgued: 3b45a28090ea95a6ad0a589a581a1297070521cc based on [this](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60iter_empty.60.2F.60iter_once.60.20new.20lints.20clippy.239187/near/290379301) fixed the FP in petgraph and clippy itself. The Lint still triggers in Tokio and unicode-normalization though.

flip1995:
I think of something different when I think about singletons. How about `iter_on_single_items`?

flip1995 (approved):
This LGTM now. renaming the lints, please squash some (doesn't have to be all) of the commits and this should be ready to go.

flip1995:
Can you squash the commits a bit further, please? Most of those are trivial/one-line changes, that we don't really need in the Clippy history.

sgued:
Is it squashed enough now or do you want even more?

flip1995:
Commits [f30d7c2](https://github.com/rust-lang/rust-clippy/pull/9187/commits/f30d7c2495b3188bfa0cf8bd6dbbf76494c9845e) and [b247594](https://github.com/rust-lang/rust-clippy/pull/9187/commits/b247594a39cb8e2c3d8d193860eac0eb1c63e1ae) could also be squashed just looking at the commit message. But that's just me being pedantic, so I won't block this PR on that.

Thanks for squashing. That's how I (almost :yum:) like my commit messages.

@bors r+

PR #9215, opened 2022-07-20
Enable test for entrypoint_recursion for windows

alex-semenyuk:
Verified that this test actually works on windows
changelog: none


Jarcho:
Seems to work. Thank you.

@bors r+

PR #9223, opened 2022-07-21
unwrap_used: Don't recommend using `expect` when the `expect_used` lint is not allowed

### LINKED ISSUE #9222: `unwrap_used` recommends using `expect` even when the `expect_used` lint is also enabled
### Description

# Short description

When both the `unwrap_used` and `expect_used` lints are opted in (with `#![warn(clippy::unwrap_used, clippy::expect_used)]`) the unwrap_used lint still recommend using `expect`, even though it would still cause an error.

# Example
With this: 
```rust
#![warn(clippy::unwrap_used, clippy::expect_used)]

fn main() {
    Some(3).unwrap();

    Some(3).expect("Hello wrld!");
}
```

The error message looks like:

```
warning: used `unwrap()` on `an Option` value
 --> src/main.rs:4:5
  |
4 |     Some(3).unwrap();
  |     ^^^^^^^^^^^^^^^^
  |
 --> src/main.rs:1:9
  |
1 | #![warn(clippy::unwrap_used, clippy::expect_used)]
  |         ^^^^^^^^^^^^^^^^^^^
  = help: if you don't want to handle the `None` case gracefully, consider using `expect()` to provide a better panic message
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used
```

The help message recommends using `expect` though it would still raise a warning.

# Fix suggestion

I think in that case the help message should be the same that is used in `expect_used`: "if this value is an `None`, it will panic`

### Version

```text
Tested with nightly in the [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=be2ec742b6bcc4cc9e32a3aa983e669a)

Nightly version: 1.64.0-nightly (2022-07-20 d68e7ebc38cb42b8b237)
```

### Additional Labels

@rustbot label + C-enhancement

sgued:
Fixes #9222 

```
changelog: [`unwrap_used`]: Don't recommend using `expect` when the `expect_used` lint is not allowed
```

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -197,7 +197,7 @@ declare_clippy_lint! {
     /// result.unwrap();
     /// ```
     ///
-    /// If [expect_used](#expect_used) is allowed, instead:
+    /// Use instead:
     /// ```rust
     /// # let option = Some(1);
     /// # let result: Result<usize, ()> = Ok(1);
@@ -205,7 +205,7 @@ declare_clippy_lint! {
     /// result.expect("more helpful message");
     /// ```
     ///
-    /// Otherwise try using
+    /// If [expect_used](#expect_used) is enabled, instead:
     /// ```rust,ignore
     /// # let option = Some(1);
     /// # let result: Result<usize, ()> = Ok(1);
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -197,12 +197,21 @@ declare_clippy_lint! {
     /// result.unwrap();
     /// ```
     ///
-    /// Use instead:
+    /// If [expect_used](#expect_used) is allowed, instead:
     /// ```rust
     /// # let option = Some(1);
     /// # let result: Result<usize, ()> = Ok(1);
     /// option.expect("more helpful message");
     /// result.expect("more helpful message");
+    /// ```
+    ///
+    /// Otherwise try using
+    /// ```rust
+    /// option?;
+    ///
+    /// // or
+    ///
+    /// result?;
     /// ```
     #[clippy::version = "1.45.0"]
     pub UNWRAP_USED,
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -206,7 +206,9 @@ declare_clippy_lint! {
     /// ```
     ///
     /// Otherwise try using
-    /// ```rust
+    /// ```rust,ignore
+    /// # let option = Some(1);
+    /// # let result: Result<usize, ()> = Ok(1);
     /// option?;
     ///
     /// // or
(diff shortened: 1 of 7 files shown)


sgued:
I also noticed that `unwrap_err` and `expect_err` don't trigger those lint and i think they should. Someone who doesn't want any `unwrap` in their codebase probably doesn't want any `unwrap_err` either. I'll make another PR after this one.

clippy_lints/src/methods/unwrap_used.rs
```diff
@@ -24,17 +24,22 @@ pub(super) fn check(cx: &LateContext<'_>, expr: &hir::Expr<'_>, recv: &hir::Expr
     }
 
     if let Some((lint, kind, none_value)) = mess {
+        let help = if is_lint_allowed(cx, EXPECT_USED, expr.hir_id) {
+            format!(
+                "if you don't want to handle the `{none_value}` case gracefully, consider \
+                using `expect()` to provide a better panic message"
+            )
+        } else {
+            format!("if this value is an `{none_value}`, it will panic")
```
giraffate, line 33: On `Option`, it would be better to be "a \`None\`" not "an \`None\`".

giraffate (commented):
Can the documentation for `unwrap_used` be updated?
https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used

clippy_lints/src/methods/unwrap_used.rs
sgued: Nice catch. I also fixed it in `expect_used`

sgued:
How would you change the documentation ? I agree pointing to `expect` there might not be ideal but it still says "at least". Maybe I could mention the `?` operator at the end?

giraffate:
@sgued At least, I want the explanation that the suggestion change if `expect_used` is enabled. It would be good to just add this to the example section. Of course, we can add more details.

sgued:
I'll add at the end:

If panicking with `expect` isn't acceptable either, consider handling the error using `match` or the `?` operator.

sgued:
I don't really know where to put that.

Serial-ATA:
@sgued You could probably change "Use instead" to "If `expect_used` is allowed, use instead".

clippy_lints/src/methods/mod.rs
```diff
@@ -197,13 +197,24 @@ declare_clippy_lint! {
     /// result.unwrap();
     /// ```
     ///
-    /// Use instead:
+    /// If [expect_used](#expect_used) is allowed, instead:
```
giraffate, line 200:
```suggestion
    /// Use instead:
```

clippy_lints/src/methods/mod.rs
```diff
@@ -197,13 +197,24 @@ declare_clippy_lint! {
...
     /// ```rust
     /// # let option = Some(1);
     /// # let result: Result<usize, ()> = Ok(1);
     /// option.expect("more helpful message");
     /// result.expect("more helpful message");
     /// ```
+    ///
+    /// Otherwise try using
```
giraffate, line 208: The `expect_used` lint is allow-by-default, so it would be better to show the case where this is enabled.
```suggestion
    /// If [expect_used](#expect_used) is enabled, instead:
```

giraffate (commented):
Thanks! Overall looks good!

giraffate:
@bors r+

Thanks!

PR #9225, opened 2022-07-22
add `[assertions_on_result_states]` lint

### LINKED ISSUE #9162: Deny `assert!(res.is_ok())` in tests
### What it does

It denies any form of `assert!(r.is_ok())`. Instead, it suggests to use `r.unwrap()` directly.

### Lint Name

_No response_

### Category

style

### Advantage

- In tests, using an `unwrap()` displays the specific offending error and helps diagnosis.

### Drawbacks

_No response_

### Example

```rust
#[test]
fn test_dummy() {
  let r = do_test();
  assert!(r.is_ok());
}
```

Could be written as:

```rust
#[test]
fn test_dummy() {
  do_test().unwrap();
}
```

tabokie:
Close #9162 

changelog: add `[assertions_on_result_states]` lint

Signed-off-by: tabokie <xy.tao@outlook.com>

### DIFF
--- a/clippy_lints/src/assertions_on_result_states.rs
+++ b/clippy_lints/src/assertions_on_result_states.rs
@@ -1,0 +1,98 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::macros::{find_assert_args, root_macro_call_first_node, PanicExpn};
+use clippy_utils::path_res;
+use clippy_utils::source::snippet_with_context;
+use clippy_utils::ty::{implements_trait, is_copy, is_type_diagnostic_item};
+use clippy_utils::usage::local_used_after_expr;
+use rustc_errors::Applicability;
+use rustc_hir::def::Res;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::{self, Ty};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `assert!(r.is_ok())` or `assert!(r.is_err())` calls.
+    ///
+    /// ### Why is this bad?
+    /// An assertion failure cannot output an useful message of the error.
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// # let r = Ok::<_, ()>(());
+    /// assert!(r.is_ok());
+    /// # let r = Err::<_, ()>(());
+    /// assert!(r.is_err());
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub ASSERTIONS_ON_RESULT_STATES,
+    style,
+    "`assert!(r.is_ok())`/`assert!(r.is_err())` gives worse error message than directly calling `r.unwrap()`/`r.unwrap_err()`"
+}
+
+declare_lint_pass!(AssertionsOnResultStates => [ASSERTIONS_ON_RESULT_STATES]);
+
+impl<'tcx> LateLintPass<'tcx> for AssertionsOnResultStates {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
+        if let Some(macro_call) = root_macro_call_first_node(cx, e)
+            && matches!(cx.tcx.get_diagnostic_name(macro_call.def_id), Some(sym::assert_macro))
+            && let Some((condition, panic_expn)) = find_assert_args(cx, e, macro_call.expn)
+            && matches!(panic_expn, PanicExpn::Empty)
+            && let ExprKind::MethodCall(method_segment, [recv], _) = condition.kind
+            && let result_type_with_refs = cx.typeck_results().expr_ty(recv)
+            && let result_type = result_type_with_refs.peel_refs()
+            && is_type_diagnostic_item(cx, result_type, sym::Result)
+            && let ty::Adt(_, substs) = result_type.kind()
+        {
+            if !is_copy(cx, result_type) {
+                if result_type_with_refs != result_type {
+                    return;
+                } else if let Res::Local(binding_id) = path_res(cx, recv)
+                    && local_used_after_expr(cx, binding_id, recv) {
+                    return;
+                }
+            }
+            let mut app = Applicability::MachineApplicable;
+            match method_segment.ident.as_str() {
+                "is_ok" if has_debug_impl(cx, substs.type_at(1)) => {
+                    span_lint_and_sugg(
+                        cx,
+                        ASSERTIONS_ON_RESULT_STATES,
+                        macro_call.span,
+                        "called `assert!` with `Result::is_ok`",
+                        "replace with",
+                        format!(
+                            "{}.unwrap()",
+                            snippet_with_context(cx, recv.span, condition.span.ctxt(), "..", &mut app).0
+                        ),
+                        app,
+                    );
+                }
+                "is_err" if has_debug_impl(cx, substs.type_at(0)) => {
+                    span_lint_and_sugg(
+                        cx,
+                        ASSERTIONS_ON_RESULT_STATES,
+                        macro_call.span,
+                        "called `assert!` with `Result::is_err`",
+                        "replace with",
+                        format!(
+                            "{}.unwrap_err()",
+                            snippet_with_context(cx, recv.span, condition.span.ctxt(), "..", &mut app).0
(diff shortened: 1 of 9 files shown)


clippy_lints/src/assert_ok.rs
```diff
@@ -0,0 +1,81 @@
...
+    pub ASSERT_OK,
+    style,
+    "`assert!(r.is_ok())` gives worse error message than directly calling `r.unwrap()`"
+}
+
+declare_lint_pass!(AssertOk => [ASSERT_OK]);
+
+impl<'tcx> LateLintPass<'tcx> for AssertOk {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
+        if_chain! {
```
Jarcho, line 35: This should use let chains instead of `if_chain!`. i.e.
```rust
if let pat = expr
    && let pat = expr
{}
```

clippy_lints/src/assert_ok.rs
```diff
@@ -0,0 +1,81 @@
...
+            if let Some(error_type) = get_error_type(cx, result_type);
+            if has_debug_impl(error_type, cx);
+            if let Some(method_receiver_snippet) = snippet_opt(cx, method_receiver.span);
+            then {
+                span_lint_and_sugg(
+                    cx,
+                    ASSERT_OK,
+                    macro_call.span,
+                    &format!(
+                        "`assert!({}.is_ok())` gives bad error message",
```
Jarcho, line 53: The lint message shouldn't contain the code that was written as it's already highlighted by the span. You also don't need to state the reasoning behind the lint. Instead, aim for a single short sentence that highlights the code pattern causing the lint.

Jarcho (changes requested):
Two small things, and then one suggested extension which should be easy enough to add.

Can you also add a check for `is_err` with the suggestion to use `unwrap_err` to the same lint. Could be named something like `assert_result_state`.

tabokie:
@Jarcho Thanks for the review.

> Can you also add a check for `is_err` with the suggestion to use `unwrap_err` to the same lint.

I have doubts about this. `unwrap` is more favorable in the error message part, but `unwrap_err` is pretty much equivalent to `is_err`. So I think we shouldn't enforce it, at least not in the same level as `is_ok`.

clippy_lints/src/assert_ok.rs
tabokie: Is there a particular reason that let chain is better? I see a few `let`s with dependency which can't be collapsed, and a few non-`let` conditions as well.

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -0,0 +1,99 @@
...
+
+declare_lint_pass!(AssertionsOnResultStates => [ASSERTIONS_ON_RESULT_STATES]);
+
+impl<'tcx> LateLintPass<'tcx> for AssertionsOnResultStates {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
+        if let Some(macro_call) = root_macro_call_first_node(cx, e)
+          && matches!(cx.tcx.get_diagnostic_name(macro_call.def_id), Some(sym::assert_macro))
+          && let Some((condition, panic_expn)) = find_assert_args(cx, e, macro_call.expn)
+          && matches!(panic_expn, PanicExpn::Empty)
+          && let ExprKind::MethodCall(method_segment, args, _) = condition.kind {
```
Jarcho, line 40: Use a slice pattern here. i.e. `ExprKind::MethodCall(method_segment, [revc], _)`

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -0,0 +1,99 @@
...
+                    ),
+                    Applicability::MaybeIncorrect,
+                );
+            }
+        }
+    }
+}
+
+/// Given a `Result<T, E>` type, return its value type (`T`) or error type (`E`).
+fn get_result_child_type<'a>(cx: &LateContext<'_>, ty: Ty<'a>, err: bool) -> Option<Ty<'a>> {
```
Jarcho, line 81: In general, functions taking boolean parameters to change there behavior are best avoided. For short functions it can be better to split them into two functions unless you need the ability to dispatch at runtime.

For this case it would be better to inline the function. e.g.
```rust
if ...
    && if let ty::Adt(def, substs) = ty.kind()
    && cx.tcx.is_diagnostic_item(sym::Result, def.did())
{
    if name == "is_ok"
        && let Some(ty) = substs.type_at(0)
        ...
    {}
}
```

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -0,0 +1,99 @@
...
+    #[clippy::version = "1.64.0"]
+    pub ASSERTIONS_ON_RESULT_STATES,
+    style,
+    "`assert!(r.is_ok())`/`assert!(r.is_err())` gives worse error message than directly calling `r.unwrap()`/`r.unwrap_err()`"
+}
+
+declare_lint_pass!(AssertionsOnResultStates => [ASSERTIONS_ON_RESULT_STATES]);
+
+impl<'tcx> LateLintPass<'tcx> for AssertionsOnResultStates {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
```
Jarcho, line 35: Note that rustfmt currently doesn't format let chains yet so you have to format everything manually.

Continued lines should be indented four spaces. The opening brace of a multi-line if condition should be on it's own line, not indented.
```rust
if let Some(x) = something()
    && let Some(y) = x.something_else()
    && is_other_thing(y)
{
    so_something_here();
}
```

This will help avoid a large fixup commit when rustfmt finally does format these correctly.

Jarcho (changes requested):

Jarcho:
There's one more issue to handle after the previous smaller things. This change does result in a behaviour change. Namely `unwrap` consumes it's argument whereas `is_ok` does not. If you restrict this to `Result` types which implement `Copy` (use `clippy_utils::ty::is_copy) that would be fine.

If you want to go further you can check if the value is both movable, and not used afterwards. I wouldn't recommend this for the current PR as it's a fair amount of extra work.

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -0,0 +1,92 @@
...
+        if let Some(macro_call) = root_macro_call_first_node(cx, e)
+            && matches!(cx.tcx.get_diagnostic_name(macro_call.def_id), Some(sym::assert_macro))
+            && let Some((condition, panic_expn)) = find_assert_args(cx, e, macro_call.expn)
+            && matches!(panic_expn, PanicExpn::Empty)
+            && let ExprKind::MethodCall(method_segment, [recv], _) = condition.kind
+            && let Res::Local(binding_id) = path_res(cx, recv)
```
Jarcho, line 44: This will make `foo().is_ok()` not lint.

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            let result_type = cx.typeck_results().expr_ty(recv);
```
Jarcho, line 46: This can be included in the previous let chain.
```rust
    ...
    && let Res::Local(binding_id) = path_res(cx, recv)
    && let result_type = cx.typeck_results().expr_ty(recv)
    && is_type_diagnostic_item(cx, result_type, sym::Result)
    ...
```

Jarcho (changes requested):

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -0,0 +1,97 @@
...
+                "is_ok" if has_debug_impl(cx, substs.type_at(1)) => {
+                    span_lint_and_sugg(
+                        cx,
+                        ASSERTIONS_ON_RESULT_STATES,
+                        macro_call.span,
+                        "called `assert!` with `Result::is_ok`",
+                        "replace with",
+                        format!(
+                            "{}.unwrap()",
+                            snippet_opt(cx, recv.span).unwrap()
```
Jarcho, line 67: This should use `snippet_with_context()` with `condition.span.ctxt()`. This will handle `assert!(mac!().is_ok())`

Jarcho (approved):
Everything looks good. You can just squash it down and we're good to merge.

tabokie:
@Jarcho It's done. Thanks.

Jarcho:
Thank you.

@bors r+

Issue #8368, opened 2022-01-29
New lint: suggest using "implicit named arguments"


stanislav-tkach:
### What it does

This lint suggests using the feature added in the 1.58 version of Rust: implicit named arguments for formatting macros.

### Lint Name

implicit_named_arguments

### Category

style

### Advantage

- Slightly less verbose code.
- (Arguably) less error prone code.

### Drawbacks

The only drawback that I can see is that not everyone would probably be happy with this lint, so perhaps it should be opt-in (pedantic).

### Example

```rust
println!("1 = {}, 2 = {}, 3 = {}", first, second, third);
```

Could be written as:

```rust
println!("1 = {first}, 2 = {second}, 3 = {third}");
```

stanislav-tkach:
Perhaps this issue duplicated https://github.com/rust-lang/rust-clippy/issues/633.

nyurik:
I began av implementation of this lint based on the https://github.com/rust-lang/rust-clippy/pull/8518 by @Alexendoo. So far I got it to identify most cases, but I am still struggling with the internal API -- for some reason the suggestions do not modify the source code. I am also not certain how to parse the structs in case the same value is used multiple times, i.e. `("{0} {0}", var)` case.  Any help and feedback is greatly appreciated :)

Alexendoo:
You could do the same as in `check_literal`, using a map to only consider arguments that are used once

https://github.com/rust-lang/rust-clippy/blob/818a91e8c83b5acd040594bb03199966f6394f1a/clippy_lints/src/write.rs#L423-L430

nyurik:
I added some more implementation code, and some of it even works (kinda).  Any feedback is welcome - probably best to do it on the #9233 PR.  Thanks!

nyurik:
Forgot to comment on this one... DONE! :D

PR #9233, opened 2022-07-24
new uninlined_format_args lint to inline explicit arguments

nyurik:
Implement https://github.com/rust-lang/rust-clippy/issues/8368 - a new lint to inline format arguments such as `print!("{}", var)` into `print!("{var}")`.

### Supported cases

code | suggestion | comment
---|---|---
`print!("{}", var)` | `print!("{var}")` |  simple variables
`print!("{0}", var)` | `print!("{var}")` |  positional variables
`print!("{v}", v=var)` | `print!("{var}")` |  named variables
`print!("{0} {0}", var)` | `print!("{var} {var}")` |  aliased variables
`print!("{0:1$}", var, width)` | `print!("{var:width$}")` |  width support
`print!("{0:.1$}", var, prec)` | `print!("{var:.prec$}")` |  precision support
`print!("{:.*}", prec, var)` | `print!("{var:.prec$}")` |  asterisk support

### Known Problems

* There may be a false positive if the format string is wrapped in a macro call:
```rust
# let var = 42;
macro_rules! no_param_str { () => { "{}" }; }
macro_rules! pass_through { ($expr:expr) => { $expr }; }
println!(no_param_str!(), var);
println!(pass_through!("{}"), var);
```

* Format string uses an indexed argument that cannot be inlined.
Supporting this case requires re-indexing of the format string.
Until implemented, `print!("{0}={1}", var, 1+2)` should be changed to `print!("{var}={0}", 1+2)` by hand.

changelog: [`uninlined_format_args`]: A new lint to inline format arguments, i.e. `print!("{}", var)` into `print!("{var}")`


Alexendoo (commented):
👍

I had a look through most of the `TODO`s. Also worth mentioning you can use [let chains](https://github.com/rust-lang/rust/pull/94927) rather than `if_chain`, it plays better with rust_analyzer (though also doesn't get automatically formatted yet)

nyurik:
Thanks @Alexendoo for the review!  I made a number of changes based on that, but still a few todos.  My biggest concern is how to detect if an argument is also used as precision and/or width.  I know some work is being done in the rustc, and there seem to be at least 3 devs involved, so I wonder who/how can coordinate these efforts to be consistent and reuse each other's work.

est31:
If it's not in your TODOs yet, the lint should be integrated into the MSRV system of clippy and be gated on 1.58.0, as [that's when format capturing was added](https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1580-2022-01-13).

nyurik:
@est31 thx, fixed.  I think I am done with this lint -- it now solves the majority of cases, and it avoids expanding width/precision arguments by checking if the argument span contains a `$` symbol.  Not ideal, but much better than nothing.

xFrednet:
@Alexendoo Do you want to take over the full review? I believe you know the format macro better than me. If you don't have the time, feel free to give it back. I should have more time for Clippy again :upside_down_face:

Alexendoo:
Sure! Sounds good to me. I'll take a full look at it once it's unblocked

xFrednet:
Awesome, thank you!

r? @Alexendoo

nyurik:
Thx @Alexendoo !  Is there something i can do to help the #8518 ?  Are there any fundamental blocks on that, or is it juts a matter of patiently waiting? :)

Alexendoo:
Yeah just a matter of waiting for the large amount of reviewer bandwidth needed

clippy_lints/src/inline_format_args.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// ### Example
+    /// ```rust
+    /// // example code where clippy issues a warning
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// // example code which does not raise clippy warning
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub INLINE_FORMAT_ARGS,
```
flip1995, line 24: The lint name is wrong. It suggests that this **lints on** inline format args rather than that it **suggests** using them.

Maybe `needless_explicit_format_args`? :thinking:

flip1995 (commented):
Thanks for this PR! I learned something about the format macro and what is possible with it.

The impl LGTM overall, but some housekeeping, like documentation, has to be done still.

clippy_utils/src/macros.rs
```diff
@@ -562,6 +574,8 @@ pub struct FormatArgsArg<'tcx> {
     pub spec: Option<&'tcx Expr<'tcx>>,
     /// Span of the `{}`
     pub span: Span,
+    /// Span of the "argument" part of the `{arg:...}`. Can be empty.
+    pub argument_span: Span,
```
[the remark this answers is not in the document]
nyurik: @flip1995 argument can be empty, but it still exists:   `format!("{}", foo)` is an empty argument span, but its **position** is there.  This is different from width and precision, which might be omitted (optional).

tests/ui/inline_format_args.rs
```diff
@@ -0,0 +1,108 @@
...
+    println!("{}", fn_arg);
+    println!("{:?}", local_value);
+    println!("{:#?}", local_value);
+    println!("{:4}", local_value);
+    println!("{:04}", local_value);
+    println!("{:<3}", local_value);
+    println!("{:#010x}", local_value);
+    println!("{:.1}", local_value2);
+    println!("Hello {} is {:.*}", "x", local_value, local_value2);
+    println!("{0}", local_value);
```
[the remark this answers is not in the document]
nyurik: @flip1995 I could, but with the recent changes to the underlying format string parsing code, it might be possible for me to handle more cases - which SHOULD be inlined, but I didn't have the needed info to handle them.

clippy_lints/src/inline_format_args.rs
nyurik: ah, naming... :)  Maybe `explicit_format_args` is a bit simpler?  or `uninlined_format_args` / `noninlined_format_args`?

nyurik:
@flip1995 thanks for the review! I addressed some of your points, and made a few clarifications. I am still in a bit of a limbo state - waiting for the upstream #8518 conflicts to be resolved, plus now there is #9349 - so I am not certain at which point I should start resolving conflicts and getting it back to the "ready to merge" state again.

clippy_utils/src/macros.rs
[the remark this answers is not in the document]
flip1995: Ah so it is the span between the `{}`, excluding the `{}`? If so, that wasn't immediately clear to me from the documentation.

clippy_lints/src/inline_format_args.rs
flip1995: I'm not the best at lint naming. Let's decide this right before merging. We can also ask other on Zulip by then.

flip1995:
No worries. Let's get the other PRs merged first. You can work on this as you see fit. But also feel free to just let this PR sit as-is until we resolved the other issues.

nyurik:
I decided to spend a bit of time rewriting inlining -- now it's part of the same `format_args.rs` as the other similar lints, and uses the new parser.

New functionality:  now supports aliases, width, and precision argument inlining, including the starred syntax (eeek!).

While at it, I discovered that my unit tests detected a regression #9468.  In this PR, I simply revert the offending patch, but I hope it will get fixed before merging.

One big TODO: handle the positional references, and renumber them if needed.

nyurik:
This PR is now rebased on main branch, ready for review. The biggest TODO is what should we name this thing?  The current name suggests that this lints on inline format args rather than that it suggests using them (per @flip1995)

Suggestions so far:
* needless_explicit_format_args
* explicit_format_args
* uninlined_format_args
* noninlined_format_args
* ???

est31:
My suggestion: `inlineable_format_args`.

From the ones you suggested @nyurik `explicit_format_args` is my favourite.

Also there is the question about which things to abbreviate. Should it be `args` or `arguments`? Should it be `fmt` or `format`?

nyurik:
Looking at all the lints, there are no `FMT`, only `FORMAT`. There is a mix of `ARG` (6) and `ARGUMENT` (2), so it seems the `ARG` variant is more common.

est31:
Hmm yeah, the macro in std is also called `format_args`, so I guess writing out `format` and shortening `args` is the best way.

nyurik:
Seems it is very common to use `needless` in the lint names, so per Lukas (zulip) suggestion, renaming it to `needless_format_args` for now.

clippy_utils/src/macros.rs
```diff
@@ -723,7 +727,7 @@ pub struct FormatArg<'tcx> {
 pub struct FormatArgsExpn<'tcx> {
     /// The format string literal.
     pub format_string: FormatString,
-    // The format arguments, such as `{:?}`.
+    /// The format arguments, such as `{:?}`.
```
Alexendoo, line 730: oops, thanks 😄

clippy_lints/src/format_args.rs
```diff
@@ -86,11 +149,64 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
...
         }
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+enum ParamType {
+    Argument,
+    Width,
+    Precision,
+}
```
Alexendoo, line 173: This is pretty useful, I think it'd be worth adding to `FormatParam` (maybe with a different name so there aren't both `kind` and `type` fields), that way you could have like `for param in format_args.params() { check_inline(..) }`

Alexendoo (commented):
Thanks for waiting :D

tests/ui/needless_format_args.rs
```diff
@@ -0,0 +1,126 @@
...
+    let local_f64 = 2.0;
+    let local_opt: Option<i32> = Some(3);
+    let width = 4;
+    let prec = 5;
+    let val = 6;
+
+    // make sure this file hasn't been corrupted with tabs converted to spaces
+    // TODO: ideally this should be a const compiler check, similar to
+    // https://docs.rs/static_assertions/1.1.0/static_assertions/macro.const_assert_eq.html
+    assert_eq!("	 	", "\t \t");
```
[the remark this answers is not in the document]
Alexendoo: or `let _: &[u8; 3] = b"[tab] [tab]";` for the current form

clippy_lints/src/format_args.rs
```diff
@@ -86,11 +149,64 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
...
+    if matches!(param.kind, Implicit | Starred | Named(_) | Numbered)
+        && let ExprKind::Path(QPath::Resolved(None, path)) = param.value.kind
+        && let Path { span, segments, .. } = path
+        && let [segment] = segments
+    {
+        let c = inline_spans.get_or_insert_with(Vec::new);
+        // TODO: Note the inconsistency here, that we may want to address separately:
+        // implicit, numbered, and starred `param.span` spans the whole relevant string:
+        // the empty space between `{}`, or the entire value `1$`, `.2$`, or `.*`
+        // but the named argument spans just the name itself, without the surrounding `.` and `$`.
```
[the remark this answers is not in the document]
nyurik: do you want me to apply some change to macros.rs?  I'm not certain which it is from the diff.  Or do you want to submit it separatelly?

clippy_lints/src/format_args.rs
nyurik: thx, yeah, makes things cleaner

clippy_lints/src/format_args.rs
```diff
@@ -76,7 +123,23 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
             if is_format_macro(cx, macro_def_id);
             if let ExpnKind::Macro(_, name) = outermost_expn_data.kind;
             then {
+                // if at least some of the arguments/format/precision are referenced by an index,
+                // e.g.  format!("{} {1}", foo, bar)  or  format!("{:1$}", foo, 2)
+                // we cannot remove an argument from a list until we support renumbering.
+                // We are OK if we inline all numbered arguments.
+                let mut do_inline = true;
+                // if we find one or more suggestions, this becomes a Vec of replacements
+                let mut inline_spans = None;
```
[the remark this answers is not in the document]
nyurik: dope, i thought `Vec::new()` always allocates memory, even if empty.  Should have read the docs.

clippy_lints/src/format_args.rs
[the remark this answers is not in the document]
Alexendoo: You can do it as part of this PR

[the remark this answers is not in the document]
nyurik: yes, but which part?  Apply your diff?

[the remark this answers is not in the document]
nyurik: @Alexendoo I tried applying your change, and it did indeed makes it a bit cleaner, thanks!

clippy_lints/src/format_args.rs
```diff
@@ -86,11 +149,54 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
                     check_format_in_format_args(cx, outermost_expn_data.call_site, name, arg.param.value);
                     check_to_string_in_format_args(cx, name, arg.param.value);
                 }
+
+                // if at least some of the arguments/format/precision are referenced by an index,
+                // e.g.  format!("{} {1}", foo, bar)  or  format!("{:1$}", foo, 2)
+                // we cannot remove an argument from a list until we support renumbering.
+                // We are OK if we inline all numbered arguments.
```
Alexendoo, line 156: nit: this comment should be a lil further down, it doesn't apply to the MSRV stuff

clippy_utils/src/macros.rs
```diff
@@ -549,15 +549,23 @@ fn span_from_inner(base: SpanData, inner: rpf::InnerSpan) -> Span {
...
     Named(Symbol),
     /// An implicit named parameter, such as the `y` in `format!("{y}")`.
     NamedInline(Symbol),
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum FormatParamUsage {
+    Argument,
+    Width,
+    Precision,
```
Alexendoo, line 566: A doc comment on these would be nice, something along the lines of `/// Appears as an argument, e.g. {param}` and similar for width/precision

tests/ui/needless_format_args.fixed
```diff
@@ -0,0 +1,122 @@
...
+    println!("{}", local_opt.unwrap());
+    println!(
+        "val='{local_i32
+    }'"
+    );
+}
+
+fn main() {
+    tester(42);
+}
```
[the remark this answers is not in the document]
nyurik: @Alexendoo I tried to add a test like you suggested, but for some strange reason it is not working (the .fixed file shows both as expanded).  It seems the `FormatArgs` is instantiated with `msrv=None`. Is there some other magical thing I have to do somewhere to ensure that clippy gets the right msrv?

nyurik (commented):
thx @Alexendoo , addressed all your feedback, but ran into an issue with msrv -- see my other comment

clippy_lints/src/format_args.rs
```diff
@@ -64,7 +66,66 @@ declare_clippy_lint! {
...
+    #[clippy::version = "1.64.0"]
+    pub NEEDLESS_FORMAT_ARGS,
+    nursery,
+    "using non-inlined variables in `format!` calls"
+}
+
+impl_lint_pass!(FormatArgs => [FORMAT_IN_FORMAT_ARGS, NEEDLESS_FORMAT_ARGS, TO_STRING_IN_FORMAT_ARGS]);
+
+pub struct FormatArgs {
+    format_args_capture: bool,
```
Alexendoo, line 118:
```suggestion
    msrv: Option<RustcVersion>,
```

+ other changes as needed, moving the `meets_msrv(..)` to the `check_expr`

clippy_lints/src/format_args.rs
```diff
@@ -86,11 +147,55 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
...
+                        outermost_expn_data.call_site,
+                        "variables can be used directly in the `format!` string",
+                        |diag| {
+                            diag.multipart_suggestion("change this to", inline_spans, Applicability::MachineApplicable);
+                        },
+                    );
+                }
             }
         }
     }
```
Alexendoo, line 175:
```suggestion
    }
    
    extract_msrv_attr!(LateContext);
```

Alexendoo (commented):
The MSRV changes:

Also needed is to add the lint name to https://github.com/rust-lang/rust-clippy/blob/e120fb10c60a27670c72e7ec99e93d16c045f493/clippy_lints/src/utils/conf.rs#L216

clippy_lints/src/format_args.rs
nyurik: Aha!  This is the piece i was missing. Writing code with macros... oh well. Thanks!!!

nyurik:
Thanks @Alexendoo !  The extract_msrv_attr! maagic did the trick.

Alexendoo:
I looked through some `lintcheck` output and it looks good, but multiline formats run into https://github.com/rust-lang/rust/issues/92741 pretty hard:

image

The suggestion is still correct as far as the lint is concerned and `cargo fix` will apply it correctly, the output is just a bit rough

nyurik:
@Alexendoo how can this be fixed? Any suggested changes?

nyurik:
One more thing: when i switched to "style", it caused a lot of uitest errors because many test files use the uninlined formats. Should I fix them, or add a clippy allow to them all?

Alexendoo:
This would all be easier if clippy were a workspace but yeah there's also `clippy_dev` to do

nyurik:
@Alexendoo thanks.  I just discovered a new bug with the code when using the [indoc!](https://crates.io/crates/indoc) crate:

```rust
    println!(indoc!{"foo={}"}, local_i32);
    printdoc!("foo={}", local_i32);
```

The first line is actually inlined:

```
    println!(indoc!{"foo={local_i32}"});
```

indoc uses AST token rewriting, and yet somehow the current format args code manages to inline it. How can I detect such cases, and exclude them?

Alexendoo:
That is a pain, since `indoc!` is a proc macro that is producing a string literal with its span set to its input it's not marked as from expansion, and since it's compatible tokenization wise `clippy_utils::is_from_proc_macro` wouldn't catch it either

nyurik:
@Alexendoo but we must detect this case, otherwise the lint produces uncompilable code :(

See also https://github.com/dtolnay/indoc/issues/50

Alexendoo:
For that issue - as the message explains there's nothing the `indoc` crate can do to make `format!(indoc!())` work, a string coming from a macro can't capture variables

On our end one option would be to tokenize the space between the format string and the first value arg, it should find only whitespace/comments/a comma

nyurik:
@Alexendoo thx, is there an example of how i can do this?  I'm still a bit new to clippy internals

nyurik:
P.S. what should i do about the thousands of `tests/ui/*.rs` files which are not part of any project?  I tried running `cargo dev lint tests/ui/blocks_in_if_conditions.rs -- --fix` hoping that i could just run it for each file in that dir, but it refuses the `--fix` if I use lint with a file instead of a dir.  It would be quiet a task to modify all test files by hand

Alexendoo:
Something along the lines of

```rust
let str_span = format_args.format_string.span;

let between = str
    .with_lo(str_span.hi())
    .with_hi(hygiene::walk_chain(first_value.span, str_span.ctxt()));

if let Some(between_snippet) = snippet_opt(cx, between)
    && tokenize(between_snippet).all(|token| matches!(token.kind, [the comment/comma/whitespace kinds] ))
```

(untested)

nyurik:
@Alexendoo I'm afraid I will need a bit more help with that code.  I tried to implement it, but the span does not capture the needed  code. Plus I am having a very hard time attaching a debugger to the uitest - it keeps stopping at panics in the debug mode.

This is where i'm at:
https://github.com/rust-lang/rust-clippy/pull/9233/files#diff-d77429793a4ce2aaa4372bc9cae50f80c6195c0e46ff9c49358077bff0e60a72R185

Basically i'm not sure what the first value is in your example above. Also, while I think we can test for commas/comments, it might be easier to test for `CloseParen | CloseBrace | CloseBracket`, or even better -- try to detect if there is any token between the outermost span and the format string -- e.g. try to see if format str is wrapper by anything  - indoc in this case: `println!(  indoc!(  "format str" ), ...)`

Alexendoo:
Personally I'm inclined to leave it as a known issue for now. It's not limited to just this lint, any other lint working with string literals may have a problem with this style of proc macro, `print_literal` is affected at least

I plan to replace `expand_past_previous_comma` with something more robust, I can probably catch this issue at the same time with that

nyurik:
@Alexendoo ok, thx, I squashed and cleaned up the PR - should be in a good state now.  I also added a few more unit tests that would also fail until your changes.

Cargo.toml
```diff
@@ -46,6 +46,7 @@ clap = { version = "3.1", features = ["derive"] }
 clippy_utils = { path = "clippy_utils" }
 derive-new = "0.5"
 if_chain = "1.0"
+indoc = "1.0"
```
Alexendoo, line 49: We can leave this out, don't need to have a test for it currently as the behaviour isn't fixed yet, when it is we'll use an `aux-build` proc macro or just the passthrough `macro_rules` to avoid an extra dependency

clippy_lints/src/format_args.rs
```diff
@@ -86,9 +153,69 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
...
+                    check_uninlined_args(cx, &format_args, outermost_expn_data.call_site);
+                }
             }
         }
     }
+
+    extract_msrv_attr!(LateContext);
+}
+
+fn check_uninlined_args(cx: &LateContext<'_>, args: &FormatArgsExpn<'_>, call_site: Span) {
```
Alexendoo, line 166: You can add a check for `args.format_string.span.from_expansion()`, that'll catch the case where the string comes from a macro (excluding the weird proc macro case)

clippy_lints/src/format_args.rs
```diff
@@ -86,9 +153,69 @@ impl<'tcx> LateLintPass<'tcx> for FormatArgs {
...
+    };
+    fixes.push((param.span, replacement));
+    let arg_span = expand_past_previous_comma(cx, *span);
+    fixes.push((arg_span, String::new()));
+    true  // successful inlining, continue checking
+} else {
+    // if we can't inline a numbered argument, we can't continue
+    param.kind != Numbered
+    }
 }
```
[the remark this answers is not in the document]
nyurik: good catch, thx

clippy_lints/src/format_args.rs
nyurik: @Alexendoo thanks, it does fix the `println!(no_param_str!(), local_i32);` case, where the macro expands into a format string, but it doesn't handle these cases with the pass_through macro (`from_expansion` returns false):

```rust
    macro_rules! pass_through { ($expr:expr) => { $expr }; }
    ...
    println!(pass_through!("foo={local_i32}"), local_i32 = local_i32);
    println!(pass_through!("foo={}"), local_i32);
```

Alexendoo: Yeah in that example there's no distinguishing the span of `pass_through!("str")` and `"str"`

nyurik: @alexendoo ok, thanks! How would you like to proceed? We could either merge it as a pedantic lint due to false positives, and fix it later, or postpone it and possibly make it as style once fixed?

clippy_lints/src/format_args.rs
```diff
@@ -64,7 +66,72 @@ declare_clippy_lint! {
...
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # let var = 42;
+    /// # let width = 1;
+    /// # let prec = 2;
+    /// format!("{var}");  // implied, positional, and named variables
+    /// format!("{var} {var}");  // aliased variables
+    /// format!("{var:width$}");  // width support
+    /// format!("{var:.prec$}");  // precision and asterisk support
```
Alexendoo, line 100:
```suggestion
    /// format!("{}", var);
    /// format!("{v:?}", v = var);
    /// format!("{0} {0}", var);
    /// format!("{0:1$}", var, width);
    /// format!("{:.*}", prec, var);
    /// ```
    /// Use instead:
    /// ```rust
    /// # let var = 42;
    /// # let width = 1;
    /// # let prec = 2;
    /// format!("{var}");
    /// format!("{var:?}");
    /// format!("{var} {var}");
    /// format!("{var:width$}");
    /// format!("{var:.prec$}");
```

Alexendoo (commented):
Looks good! Just some documentation nits and then it's good for merging

Thank you for the endurance 😄

clippy_lints/src/format_args.rs
Alexendoo: My plan is to merge it, I've kept it waiting long enough 😄

Alexendoo:
Good ol' `update_lints` :D

Could give it one final squash too

nyurik:
Bummer, I can't,  I'm at a playground))))

nyurik:
I'll do it in a bit. Thanks for all the help!!!

nyurik:
@Alexendoo ok, I think it's finally ready 😀

Alexendoo (approved):

Alexendoo:
Awesome, thanks @nyurik!

@bors r+

PR #9235, opened 2022-07-24
Fix src link

alex-semenyuk:
changelog: none


llogiq:
Thank you! Good catch.

@bors r+

PR #9246, opened 2022-07-25
check macro statements in `[non_copy_const]`

### LINKED ISSUE #8493: declare_interior_mutable_const shouldn't fire for const thread-local initializers
### Summary

`thread_local!` has an optimization for `const` initializers. Also, thread-locals often have interior mutability. When using these two features together, the `declare_interior_mutable_const` lint fires, as it detects an interior-mutable `const` item. Which is true, but it is not being used in an incorrect manner; the use happens inside of the `thread_local!` macro.

### Lint Name

declare_interior_mutable_const

### Reproducer

```rust
use std::cell::Cell;
thread_local! {
    static THREAD_LOCAL: Cell<i32> = const { Cell::new(0) };
}
```

I saw this happen:

```
warning: a `const` item should never be interior mutable
 --> src/lib.rs:2:1
  |
2 | / thread_local! {
3 | |     static THREAD_LOCAL: Cell<i32> = const { Cell::new(0) };
4 | | }
  | |_^
  |
  = note: `#[warn(clippy::declare_interior_mutable_const)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
  = note: this warning originates in the macro `$crate::__thread_local_inner` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Version

```text
rustc 1.60.0-nightly (bd3cb5256 2022-01-16)
binary: rustc
commit-hash: bd3cb52565faab2755ff1bdb54d88bc91f47b4b9
commit-date: 2022-01-16
host: x86_64-apple-darwin
release: 1.60.0-nightly
LLVM version: 13.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #9224: declare_interior_mutable_const triggered in tokio::task_local
### Summary

@kyoto7250 this might be of interest to you given you solved a similar issue before.

as of tokio 1.20.0, `tokio::task_local!` triggers this lint.

I thought that https://github.com/rust-lang/rust-clippy/pull/9015 might have resolved the issue but it should be in the latest nightly and the issue is still occurring in the latest nightly.

### Lint Name

declare_interior_mutable_const

### Reproducer

I tried this code:

```rust
tokio::task_local! {
    pub static CONTEXT_CHAIN_NAME: String;
}
```

I saw this happen:

```
warning: a `const` item should never be interior mutable
   --> shotover-proxy/src/transforms/mod.rs:373:1
    |
373 | / tokio::task_local! {
374 | |     pub static CONTEXT_CHAIN_NAME: String;
375 | | }
    | |_^
    |
    = note: `#[warn(clippy::declare_interior_mutable_const)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
    = note: this warning originates in the macro `$crate::__thread_local_inner` which comes from the expansion of the macro `tokio::task_local` (in Nightly builds, run with -Z macro-backtrace for more info)

```

not trigger the lint, I expect tokio to have a sound implementation, if it doesnt then we should raise an issue on the tokio repo instead.

### Version

```text
rustc 1.64.0-nightly (d68e7ebc3 2022-07-20)
binary: rustc
commit-hash: d68e7ebc38cb42b8b237392b28045edeec761503
commit-date: 2022-07-20
release: 1.64.0-nightly
LLVM version: 14.0.6
```

### Additional Labels

_No response_

kyoto7250:
close #8493 
close #9224 

This PR fixes false positives in `[non_copy_const]`.

changelog: fix false positives in`[non_copy_const]`

---

r? @Jarcho

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ serde = { version = "1.0.125", features = ["derive"] }
 syn = { version = "1.0", features = ["full"] }
 futures = "0.3"
 parking_lot = "0.12"
-tokio = { version = "1", features = ["io-util"] }
+tokio = { version = "1", features = ["full"] }
 rustc-semver = "1.1"
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ serde = { version = "1.0.125", features = ["derive"] }
 syn = { version = "1.0", features = ["full"] }
 futures = "0.3"
 parking_lot = "0.12"
-tokio = { version = "1", features = ["full"] }
+tokio = { version = "1", features = ["io-util"] }
 rustc-semver = "1.1"
 
 [build-dependencies]
--- a/clippy_lints/src/non_copy_const.rs
+++ b/clippy_lints/src/non_copy_const.rs
@@ -251,14 +251,7 @@ impl<'tcx> LateLintPass<'tcx> for NonCopyConst {
     fn check_item(&mut self, cx: &LateContext<'tcx>, it: &'tcx Item<'_>) {
         if let ItemKind::Const(hir_ty, body_id) = it.kind {
             let ty = hir_ty_to_ty(cx.tcx, hir_ty);
-            if !macro_backtrace(it.span).last().map_or(false, |macro_call| {
-                matches!(
-                    cx.tcx.get_diagnostic_name(macro_call.def_id),
-                    Some(sym::thread_local_macro)
-                )
-            }) && is_unfrozen(cx, ty)
-                && is_value_unfrozen_poly(cx, body_id, ty)
-            {
+            if !ignored_macro(cx, it) && is_unfrozen(cx, ty) && is_value_unfrozen_poly(cx, body_id, ty) {
                 lint(cx, Source::Item { item: it.span });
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -52,7 +52,7 @@ serde = { version = "1.0.125", features = ["derive"] }
 syn = { version = "1.0", features = ["full"] }
 futures = "0.3"
 parking_lot = "0.12"
-tokio = { version = "1", features = ["io-util"] }
+tokio = { version = "1", features = ["full"] }
```
kyoto7250, line 55: This change is necessary to use `tokio::task_local`.

Jarcho: It would be better not to depend on the implementation of `tokio::task_local` for the test. It looks like it was caused by having a nested macro call so the newly added macro to the test is enough.

kyoto7250: I commied https://github.com/rust-lang/rust-clippy/pull/9246/commits/b6c301055edc9d769bb7ac8e4b8bc083e31ac748

Jarcho:
Thank you.

@bors r+

PR #9249, opened 2022-07-27
Use `action/checkout@v3` in the book

JohnTitor:
As this type of document is often copied/pasted, using a newer version of `actions/checkout` would be better.

changelog: none

Signed-off-by: Yuki Okushi <jtitor@2k36.org>


flip1995:
Great catch, thank you! That was way outdated :open_mouth: 

@bors r+

PR #9252, opened 2022-07-27
Read and use deprecated configuration (as well as emitting a warning)

bossmc:
Original change written by @flip1995 I've simply rebased to master and fixed up the formatting/tests.  This change teaches the configuration parser which config key replaced a deprecated key and attempts to populate the latter from the former.  If both keys are provided this fails with a duplicate key error (rather than attempting to guess which the user intended).

Currently this on affects `cyclomatic-complexity-threshold` -> `cognitive-complexity-threshold` but will also be used in #8974 to handle `blacklisted-names` -> `disallowed-names`.

```
changelog: deprecated configuration keys are still applied as if they were provided as their non-deprecated name.
```

- [x] Run `cargo dev fmt`

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -487,7 +487,7 @@ pub fn read_conf(sess: &Session) -> Conf {
         },
     };
 
-    let TryConf { conf, errors } = utils::conf::read(&file_name);
+    let TryConf { conf, errors, warnings } = utils::conf::read(&file_name);
     // all conf errors are non-fatal, we just use the default conf in case of error
     for error in errors {
         sess.err(&format!(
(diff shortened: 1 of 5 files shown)


flip1995:
I only vaguely remember writing this :sweat_smile: Thanks for seeing it to an end!

Jarcho:
One small issue. This won't warn about duplicate fields when the old name appears after the new name. It's not a big deal as you still get the warning about the deprecated field.

Otherwise LGTM.

bossmc:
I've added a check to handle duplicates in either order and added a test for duplicates.  I also extended the error message to be clear what's happening if one of the duplicates is using the deprecated name (though this only works on one direction).

Jarcho:
Thank you. That should be good enough. The deprecation warning mentions both names in either case.

@bors r+

PR #9254, opened 2022-07-28
Enable test empty_loop_no_std on windows

alex-semenyuk:
Verified that it actully works on windows
changelog: none

### DIFF
--- a/tests/ui/empty_loop_no_std.rs
+++ b/tests/ui/empty_loop_no_std.rs
@@ -1,6 +1,5 @@
 // compile-flags: -Clink-arg=-nostartfiles
 // ignore-macos
-// ignore-windows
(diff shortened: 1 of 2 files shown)


dswij (approved):
Thanks!

I couldn't find the reason we're ignoring this though, curious about this.

xFrednet:
I'm guessing it has something to do with the Compiler flag? 🤔

alex-semenyuk:
Not sure why, but seems it was here https://github.com/rust-lang/rust-clippy/pull/5086 Probably, @Areredify remember why so?

dswij:
Well if the test is passing, I guess it's fine then. Thanks again for this!

@bors r+

Issue #854, opened 2016-04-13
Useless `peekable` iterator


mcarton:
Just noticed that when refactoring. We could warn when making an iterator [peekable](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.peekable), but never actually peeking it(/returning it/passing it to a function).

birkenfeld:
Sounds pretty specific. Are you sure it's worth it?

lukaslueg:
@birkenfeld FWIW I've seen this in the wild, too

PR #9258, opened 2022-07-29
Add [`unused_peekable`] lint

Serial-ATA:
changelog: Add [`unused_peekable`] lint
closes: #854

### DIFF
--- a/clippy_lints/src/unused_peekable.rs
+++ b/clippy_lints/src/unused_peekable.rs
@@ -1,0 +1,208 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::ty::{match_type, peel_mid_ty_refs_is_mutable};
+use clippy_utils::{fn_def_id, path_to_local_id, paths, peel_ref_operators};
+use rustc_ast::Mutability;
+use rustc_hir::intravisit::{walk_expr, Visitor};
+use rustc_hir::lang_items::LangItem;
+use rustc_hir::{Block, Expr, ExprKind, HirId, Local, Node, PatKind, PathSegment, StmtKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for the creation of a `peekable` iterator that is never `.peek()`ed
+    ///
+    /// ### Why is this bad?
+    /// Creating a peekable iterator without using any of its methods is likely a mistake,
+    /// or just a leftover after a refactor.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let collection = vec![1, 2, 3];
+    /// let iter = collection.iter().peekable();
+    ///
+    /// for item in iter {
+    ///     // ...
+    /// }
+    /// ```
+    ///
+    /// Use instead:
+    /// ```rust
+    /// let collection = vec![1, 2, 3];
+    /// let iter = collection.iter();
+    ///
+    /// for item in iter {
+    ///     // ...
+    /// }
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub UNUSED_PEEKABLE,
+    suspicious,
+    "creating a peekable iterator without using any of its methods"
+}
+
+#[derive(Default)]
+pub struct UnusedPeekable {
+    visited: Vec<HirId>,
+}
+
+impl_lint_pass!(UnusedPeekable => [UNUSED_PEEKABLE]);
+
+impl<'tcx> LateLintPass<'tcx> for UnusedPeekable {
+    fn check_block(&mut self, cx: &LateContext<'tcx>, block: &Block<'tcx>) {
+        // Don't lint `Peekable`s returned from a block
+        if let Some(expr) = block.expr
+            && let Some(ty) = cx.typeck_results().expr_ty_opt(peel_ref_operators(cx, expr))
+            && match_type(cx, ty, &paths::PEEKABLE)
+        {
+            return;
+        }
+
+        for (idx, stmt) in block.stmts.iter().enumerate() {
+            if !stmt.span.from_expansion()
+                && let StmtKind::Local(local) = stmt.kind
+                && !self.visited.contains(&local.pat.hir_id)
+                && let PatKind::Binding(_, _, ident, _) = local.pat.kind
+                && let Some(init) = local.init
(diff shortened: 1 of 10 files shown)


clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,151 @@
...
+                    self.found_peek_call = true;
+                    return;
+                }
+            },
+            _ => {},
+        }
+
+        walk_expr(self, ex);
+    }
+}
```
Alexendoo, line 151: This is a tricky one, there's a lot of ways a peekable could end up being used and I'm sure we'll miss some extra ones on top of that

e.g.

```rust
let mut lexer = Lexer { tokens: peekable };
lexer.tokens.peek();

let mut new_binding = peekable;
new_binding.peek();

let by_ref = &mut peekable;
by_ref.peek();
```

A potential way to handle this would be in the visitor check if `ex` is a path to the peekable binding, then use `parent_iter` to check any potential uses it has, considering it used if we're unsure

clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,208 @@
...
+                                    self.found_peek_call = true;
+                                    return;
+                                }
+                            },
+                            // Don't bother if moved into a struct
+                            ExprKind::Struct(..) => {
+                                self.found_peek_call = true;
+                                return;
+                            },
+                            _ => {},
```
Alexendoo, line 177: I think this should be flipped, something like

```suggestion
                            ExprKind::AddrOf(..) | ExprKind::Unary(..) | ExprKind::DropTemps(_) => {},
                            _ => {
                                self.found_peek_call = true;
                                return;
                            },
```

So we're sure we're catching all the usages, e.g. tuples is missed here currently I think. But also weirder stuff like index assignments, returns/breaks, etc.

May need to add `return` to the end of both `Call/MethodCall` arms to terminate them earlier for this to work

clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,208 @@
...
+        Self {
+            cx,
+            expected_hir_id,
+            found_peek_call: false,
+        }
+    }
+}
+
+impl<'tcx> Visitor<'_> for PeekableVisitor<'_, 'tcx> {
+    fn visit_expr(&mut self, ex: &'_ Expr<'_>) {
```
Alexendoo, line 119: Could add an early return here `if self.found_peek_call` so we don't have to keep walking if we already found one

clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,208 @@
...
+                                        && let Some(ty) = self
+                                            .cx
+                                            .typeck_results()
+                                            .expr_ty_opt(arg)
+                                            .map(Ty::peel_refs)
+                                        && match_type(self.cx, ty, &paths::PEEKABLE)
+                                    {
+                                        self.found_peek_call = true;
+                                        return;
+                                    }
```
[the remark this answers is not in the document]
Serial-ATA: Looks like it is :smile:.

clippy_lints/src/unused_peekable.rs
Serial-ATA: Every time `self.found_peek_call` is set to `true` it returns anyway.

clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,208 @@
...
+                                    || method_name == "next_if"
+                                    || method_name == "next_if_eq")
+                                    && let ExprKind::Path(_) = arg.kind
+                                    && let Some(ty) = self.cx.typeck_results().expr_ty_opt(arg).map(Ty::peel_refs)
+                                    && match_type(self.cx, ty, &paths::PEEKABLE)
+                                {
+                                    self.found_peek_call = true;
+                                    return;
+                                }
+                            },
```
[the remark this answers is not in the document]
Serial-ATA: That case seems to pass already.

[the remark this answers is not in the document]
Alexendoo:
```rust
fn main() {
    struct S;
    impl S {
        fn method(&self, _o: Peekable<impl Iterator>) {}
    }

    let peekable = "".chars().peekable();
    let s = S;
    s.method(peekable);
}
```

is what I had in mind, gives me a warning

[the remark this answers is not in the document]
Alexendoo: Ah just saw the edit 👍

Serial-ATA:
One of the tests has a FN now, I won't be able to fix it for a few days.

clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,230 @@
...
+                && let PatKind::Binding(_, _, ident, _) = local.pat.kind
+                && let Some(init) = local.init
+                && !init.span.from_expansion()
+                && let Some(ty) = cx.typeck_results().expr_ty_opt(init)
+                && let (ty, _, Mutability::Mut) = peel_mid_ty_refs_is_mutable(ty)
+                && match_type(cx, ty, &paths::PEEKABLE)
+            {
+                self.visited.push(local.pat.hir_id);
+
+                let mut vis = PeekableVisitor::new(cx, local.pat.hir_id, &mut self.visited);
```
Alexendoo, line 75: Rather than the pattern's `hir_id` this should use `binding` from `&& let PatKind::Binding(_, binding, ident, _) = local.pat.kind`. Since it works I guess they're the same in this case, but maybe that could change in the future

clippy_lints/src/unused_peekable.rs
```diff
@@ -0,0 +1,230 @@
...
+                                    return;
+                                }
+
+                                // foo.some_method() excluding Iterator methods
+                                if remaining_args.iter().any(|arg| arg_is_mut_peekable(self.cx, arg))
+                                    && !is_trait_method(self.cx, expr, sym::Iterator)
+                                {
+                                    self.found_peek_call = true;
+                                    return;
+                                }
```
Alexendoo, line 178: `Iterator::by_ref` should be included as well

```rust
let mut x = "abc".chars().peekable();
let y = x.by_ref().peek();
```

Serial-ATA: Good call!

Alexendoo (approved):

Alexendoo:
Looks good thanks!

@bors r+

Serial-ATA:
Thanks for all the help on this @Alexendoo!

PR #9259, opened 2022-07-29
Fix `to_string_in_format_args` false positive

### LINKED ISSUE #9256: Incorrect suggested removal of to_string on slice
### Summary

`clippy::to_string_in_format_args` gives incorrect output when you call to_string on a str slice.
The code it suggests doesn't compile.

### Reproducer

I tried this code:

```rust
fn print_substring(original: &str) {
    assert!(original.len() > 10);
    println!("{}", original[..10].to_string());
}

fn main() {
    print_substring("Hello, world!");
}
```

Clippy would either produce no error, or offer a valid correction

Instead, this happened:
It offered this alternative, which doesn't compile.
original[..10] doesn't have a known size, so can't be passed to println!
It should be &original[..10]
```
warning: `to_string` applied to a type that implements `Display` in `println!` args
 --> src/main.rs:3:34
  |
3 |     println!("{}", original[..10].to_string());
  |                                  ^^^^^^^^^^^^ help: remove this
  ```

### Version

```text
binary: rustc
commit-date: 2022-06-27
release: 1.62.0
LLVM version: 14.0.5
```

### Additional Labels

smoelius:
Fix #9256

changelog: none

### DIFF
--- a/clippy_lints/src/format_args.rs
+++ b/clippy_lints/src/format_args.rs
@@ -131,34 +131,49 @@ fn check_to_string_in_format_args(cx: &LateContext<'_>, name: Symbol, value: &Ex
         if is_diag_trait_item(cx, method_def_id, sym::ToString);
         let receiver_ty = cx.typeck_results().expr_ty(receiver);
         if let Some(display_trait_id) = cx.tcx.get_diagnostic_item(sym::Display);
+        let (n_needed_derefs, target) =
+            count_needed_derefs(receiver_ty, cx.typeck_results().expr_adjustments(receiver).iter());
+        if implements_trait(cx, target, display_trait_id, &[]);
+        if let Some(sized_trait_id) = cx.tcx.lang_items().sized_trait();
         if let Some(receiver_snippet) = snippet_opt(cx, receiver.span);
         then {
-            let (n_needed_derefs, target) = count_needed_derefs(
-                receiver_ty,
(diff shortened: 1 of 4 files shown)


clippy_lints/src/format_args.rs
```diff
@@ -134,34 +134,49 @@ fn check_to_string_in_format_args(cx: &LateContext<'_>, name: Symbol, value: &Ex
...
-                        format!("{:*>width$}{}", "", receiver_snippet, width = n_needed_derefs),
-                        Applicability::MachineApplicable,
-                    );
-                }
+            let needed_ref = if implements_trait(cx, receiver_ty, sized_trait_id, &[]) {
+                ""
+            } else {
+                "&"
+            };
+            if n_needed_derefs == 0 && needed_ref.is_empty() {
```
llogiq, line 148: How about a `needs_ref` instead of a `needed_ref` so we don't need an `is_empty()` here?

llogiq (approved):
Just one small nit pick, otherwise this looks good.

smoelius:
Hopefully, I understood correctly. Thanks for reviewing, @llogiq.

llogiq:
Yes, that's what I meant. Sorry for being absent for a while. r=me after rebase.

smoelius:
Rebase done. Thanks again, @llogiq.

llogiq:
Thank you!

@bors r+

PR #9269, opened 2022-07-31
Lint `collapsible_str_replace`

### LINKED ISSUE #6651: New lint: `collapsible_str_replace`
### What it does

Checks for consecutive calls to `str::replace` (2 or more) that can be collapsed into a single call.

### Categories (optional)

- Kind: perf

Faster since the string is only scanned once. Also less repetitive code.

### Drawbacks

None.

### Example

```rust
"hesuo worpd"
    .replace('s', "l")
    .replace("u", "l")
    .replace('p', "l")
```

Could be written as:

```rust
"hesuo worpd".replace(|c| matches!(c, 's' | 'u' | 'p'), "l")
```

Using `matches!` is potentially faster than a slice of chars (`replace(&['s', 'u', 'p'], "l")`). But if any of the chars are variables, the lint can fall back to suggesting a slice of chars.

nahuakang:
fixes #6651

```
changelog: [`collapsible_str_replace`]: create new lint `collapsible_str_replace`
```

- \[ ] Executed `cargo dev update_lints`
- \[x] Run `cargo dev fmt`

### DIFF
--- a/clippy_lints/src/methods/collapsible_str_replace.rs
+++ b/clippy_lints/src/methods/collapsible_str_replace.rs
@@ -1,0 +1,6 @@
+use rustc_hir as hir;
+use rustc_hir::*;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &hir::Expr<'_>) {}
--- a/clippy_lints/src/methods/collapsible_str_replace.rs
+++ b/clippy_lints/src/methods/collapsible_str_replace.rs
@@ -1,3 +1,5 @@
+// run-rustfix
+
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::get_parent_expr;
 use clippy_utils::visitors::for_each_expr;
@@ -7,7 +9,7 @@ use rustc_ast::ast::LitKind;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
-use rustc_hir::*;
+use rustc_hir::{ExprKind, Path, QPath};
 use rustc_lint::LateContext;
 use rustc_middle::ty;
 use rustc_span::source_map::Spanned;
@@ -21,44 +23,31 @@ pub(super) fn check<'tcx>(
     expr: &'tcx hir::Expr<'tcx>,
     name: &str,
     recv: &'tcx hir::Expr<'tcx>,
-    args: &'tcx [hir::Expr<'tcx>],
 ) {
-    match (name, args) {
-        ("replace", ..) => {
-            // The receiver of the method call must be `str` type to lint `collapsible_str_replace`
-            let original_recv = find_original_recv(recv);
-            let original_recv_ty_kind = cx.typeck_results().expr_ty(original_recv).peel_refs().kind();
-            let original_recv_is_str_kind = matches!(original_recv_ty_kind, ty::Str);
+    if name == "replace" {
+        // The receiver of the method call must be `str` type to lint `collapsible_str_replace`
+        let original_recv = find_original_recv(recv);
+        let original_recv_ty_kind = cx.typeck_results().expr_ty(original_recv).peel_refs().kind();
+        let original_recv_is_str_kind = matches!(original_recv_ty_kind, ty::Str);
 
-            if_chain! {
-                if original_recv_is_str_kind;
-                if let Some(parent) = get_parent_expr(cx, expr);
-                if let Some((name, ..)) = method_call(parent);
+        if_chain! {
+            if original_recv_is_str_kind;
+            if let Some(parent) = get_parent_expr(cx, expr);
+            if let Some((name, ..)) = method_call(parent);
+            if name == "replace";
 
-                then {
-                    match name {
-                        // If the parent node is a `str::replace` call, we've already handled the lint, don't lint again
-                        "replace" => return,
-                        _ => {
-                            check_consecutive_replace_calls(cx, expr);
-                            return;
-                        },
-                    }
-                }
+            then {
+                // If the parent node is a `str::replace` call, we've already handled the lint, don't lint again
+                return;
             }
+        }
 
-            match method_call(recv) {
-                // Check if there's an earlier `str::replace` call
-                Some(("replace", ..)) => {
-                    if original_recv_is_str_kind {
-                        check_consecutive_replace_calls(cx, expr);
-                        return;
-                    }
-                },
-                _ => {},
+        if let Some(("replace", ..)) = method_call(recv) {
+            // Check if there's an earlier `str::replace` call
+            if original_recv_is_str_kind {
+                check_consecutive_replace_calls(cx, expr);
             }
-        },
-        _ => {},
+        }
     }
 }
 
@@ -116,7 +105,7 @@ fn check_consecutive_replace_calls<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx hir
 
 /// Check if all the `from` arguments of a chain of consecutive calls to `str::replace`
 /// are all of `ExprKind::Lit` types. If any is not, return false.
-fn replace_call_from_args_are_only_lit_chars<'tcx>(from_args: &Vec<&'tcx hir::Expr<'tcx>>) -> bool {
+fn replace_call_from_args_are_only_lit_chars<'tcx>(from_args: &[&'tcx hir::Expr<'tcx>]) -> bool {
     let mut only_lit_chars = true;
 
     for from_arg in from_args.iter() {
@@ -159,9 +148,9 @@ fn get_replace_call_from_args_if_all_char_ty<'tcx>(
 
     if all_from_args_are_chars {
         return Some(from_args);
-    } else {
-        return None;
     }
+
+    None
 }
 
 /// Return a unique String representation of the `to` argument used in a chain of `str::replace`
@@ -186,13 +175,13 @@ fn get_replace_call_unique_to_arg_repr<'tcx>(expr: &'tcx hir::Expr<'tcx>) -> Opt
 
     // let mut to_arg_repr_set = FxHashSet::default();
     let mut to_arg_reprs = Vec::new();
-    for &to_arg in to_args.iter() {
+    for &to_arg in &to_args {
         if let Some(to_arg_repr) = get_replace_call_char_arg_repr(to_arg) {
             to_arg_reprs.push(to_arg_repr);
         }
     }
 
-    let to_arg_repr_set = FxHashSet::from_iter(to_arg_reprs.iter().cloned());
+    let to_arg_repr_set = to_arg_reprs.iter().cloned().collect::<FxHashSet<_>>();
     // Check if the set of `to` argument representations has more than one unique value
     if to_arg_repr_set.len() != 1 {
         return None;
--- a/clippy_lints/src/methods/collapsible_str_replace.rs
+++ b/clippy_lints/src/methods/collapsible_str_replace.rs
@@ -1,17 +1,16 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-// use clippy_utils::source::snippet_with_context;
+use clippy_utils::get_parent_expr;
 use clippy_utils::visitors::for_each_expr;
 use core::ops::ControlFlow;
 use if_chain::if_chain;
 use rustc_ast::ast::LitKind;
+use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_hir::*;
 use rustc_lint::LateContext;
 use rustc_middle::ty;
 use rustc_span::source_map::Spanned;
-use std::unreachable;
-// use rustc_span::Span;
 
 use super::method_call;
 use super::COLLAPSIBLE_STR_REPLACE;
@@ -25,28 +24,46 @@ pub(super) fn check<'tcx>(
 ) {
     match (name, args) {
         ("replace", [from, to]) => {
-            // Check for `str::replace` calls with char slice for linting
+            // The receiver of the method call must be `str` type to lint `collapsible_str_replace`
             let original_recv = find_original_recv(recv);
-            let original_recv_ty = cx.typeck_results().expr_ty(original_recv).peel_refs();
+            let original_recv_ty_kind = cx.typeck_results().expr_ty(original_recv).peel_refs().kind();
+            let original_recv_is_str_kind = matches!(original_recv_ty_kind, ty::Str);
+
             if_chain! {
-                // Check the receiver of the method call is `str` type
-                if matches!(original_recv_ty.kind(), ty::Str);
-                let from_ty = cx.typeck_results().expr_ty(from).peel_refs();
-                if let ty::Array(array_ty, _) = from_ty.kind();
+                // Check for `str::replace` calls with char slice for linting
+                if original_recv_is_str_kind;
+                let from_ty_kind = cx.typeck_results().expr_ty(from).peel_refs().kind();
+                if let ty::Array(array_ty, _) = from_ty_kind;
                 if matches!(array_ty.kind(), ty::Char);
                 then {
                     check_replace_call_with_char_slice(cx, from, to);
+                    return;
+                }
+            }
+
+            if_chain! {
+                if original_recv_is_str_kind;
+                if let Some(parent) = get_parent_expr(cx, expr);
+                if let Some((name, [..], _)) = method_call(parent);
+
+                then {
+                    match name {
+                        "replace" => return,
+                        _ => {
+                            check_consecutive_replace_calls(cx, expr);
+                            return;
+                        },
+                    }
                 }
             }
 
             match method_call(recv) {
                 // Check if there's an earlier `str::replace` call
-                Some(("replace", [prev_recv, prev_from, prev_to], prev_span)) => {
-                    println!("Consecutive replace calls");
-                    // Check that the original receiver is of `ty::Str` type
-                    // Check that all the `from` args are char literals
-                    // Check that all the `to` args are the same variable or has the same &str value
-                    // If so, then lint
+                Some(("replace", [_, _, _], _)) => {
+                    if original_recv_is_str_kind {
+                        check_consecutive_replace_calls(cx, expr);
+                        return;
+                    }
                 },
                 _ => {},
             }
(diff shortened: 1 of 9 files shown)


tests/ui/collapsible_str_replace.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    let replacement = misspelled.replace(s, "l").replace(u, "l").replace('p', "l");
+    println!("{replacement}");
+
+    let replacement = misspelled.replace(s, "l").replace(u, "l").replace(p, "l");
+    println!("{replacement}");
+
+    // If the first argument to a single `str::replace` call is a slice and one or more chars
+    // are variables, do not recommend `collapsible_str_replace`
+    let replacement = misspelled.replace(&['s', u, 'p'], "l");
```
flip1995, line 39: I think the comment, why this shouldn't be linted is wrong. If there's another `replace` behind this `replace` it could be added to the slice. But producing this suggestion might be hard. So this is not necessary for the first version of this lint. 

But yeah, this is not linted because there is only one `replace`, not because the pattern is a slice.

flip1995 (commented):
Only one comment about the test, otherwise the tests LGTM.

This could be included in the `methods` module. Then you already get the method matching for free. You'll also have to check the type of the receiver and check if it is a string-ish (`&str`, `String`, ...) type. I think we have a helper for this(?).

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+                return;
+            }
+        }
+
+        if let Some(("replace", ..)) = method_call(recv) {
+            // Check if there's an earlier `str::replace` call
+            if original_recv_is_str_kind {
+                check_consecutive_replace_calls(cx, expr);
+            }
+        }
```
flip1995, line 50: Why not check for this in `mod.rs`?

clippy_lints/src/methods/mod.rs
```diff
@@ -2752,6 +2780,10 @@ impl Methods {
                         unnecessary_lazy_eval::check(cx, expr, recv, arg, "or");
                     }
                 },
+                ("replace" | "replacen", [arg1, arg2] | [arg1, arg2, _]) => {
+                    no_effect_replace::check(cx, expr, arg1, arg2);
+                    collapsible_str_replace::check(cx, expr, name, recv);
+                },
```
flip1995, line 2786:
```suggestion
                (name @ "replace" | "replacen", [arg1, arg2] | [arg1, arg2, _]) => {
                    no_effect_replace::check(cx, expr, arg1, arg2);
                    if name == "replace" && let Some(("replace", ..)) = method_call(recv) {
                        collapsible_str_replace::check(cx, expr, name, recv);
                    }
                },
```

So something like this.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+            _ => only_lit_chars = false,
+        }
+    }
+
+    only_lit_chars
+}
+
+/// Collect and return all of the `from` arguments of a chain of consecutive `str::replace` calls
+/// if these `from` arguments's expressions are of the `ty::Char` kind. Otherwise return `None`.
+fn get_replace_call_from_args_if_all_char_ty<'tcx>(
```
flip1995, line 123: Function names like this scream that this function does too much and is too complex.

Why not have one function that simply walks up the method call chain (so the `if let Some(("replace", [_, args @ ..], _)) = method_call(e)` part) and collects all the `replace` `MethodCall`s and then one function that takes those exprs and only has to collect the args. 

In the second function, if you find a argument that is **not** a `char`, just early return `None`.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+    // Return the single representation value
+    to_arg_reprs.pop()
+}
+
+/// Get the representation of an argument of a `str::replace` call either of the literal char value
+/// or variable name, i.e. the resolved path segments `ident`.
+/// Return:
+/// - the str literal with double quotes, e.g. "\"l\""
+/// - the char literal with single quotes, e.g. "'l'"
+/// - the variable as a String, e.g. "l"
```
flip1995, line 199:
```suggestion
/// - the str literal with double quotes, e.g. `"l"`
/// - the char literal with single quotes, e.g. `'l'`
/// - the variable as a String, e.g. `l`
```

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+        if let Some(to_arg_repr) = get_replace_call_char_arg_repr(to_arg) {
+            to_arg_reprs.push(to_arg_repr);
+        }
+    }
+
+    let to_arg_repr_set = to_arg_reprs.iter().cloned().collect::<FxHashSet<_>>();
+    // Check if the set of `to` argument representations has more than one unique value
+    if to_arg_repr_set.len() != 1 {
+        return None;
+    }
```
flip1995, line 188: This seems unnecessary inefficient. You already know from the first `replace` what the `to` argument is. So you only have to check for this. And as soon as you find a string/char/whatever that doesn't match that, you should stop your search.

I think this should be done while collecting all of the `replace` calls and as soon as you find something different than the first replace call, you return the already collected list of replace calls.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+        // The receiver of the method call must be `str` type to lint `collapsible_str_replace`
+        let original_recv = find_original_recv(recv);
+        let original_recv_ty_kind = cx.typeck_results().expr_ty(original_recv).peel_refs().kind();
+        let original_recv_is_str_kind = matches!(original_recv_ty_kind, ty::Str);
+
+        if_chain! {
+            if original_recv_is_str_kind;
+            if let Some(parent) = get_parent_expr(cx, expr);
+            if let Some((name, ..)) = method_call(parent);
+            if name == "replace";
```
flip1995, line 37: I noticed that this doesn't really work. You'll also have to compare if the `to` argument is the same as the one of the last `replace` call in the current chain. Otherwise this wouldn't get linted:

```
.replace(_, "l").replace(_, "l").replace(_, "x")
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+    to_arg_reprs.pop()
+}
+
+/// Get the representation of an argument of a `str::replace` call either of the literal char value
+/// or variable name, i.e. the resolved path segments `ident`.
+/// Return:
+/// - the str literal with double quotes, e.g. "\"l\""
+/// - the char literal with single quotes, e.g. "'l'"
+/// - the variable as a String, e.g. "l"
+fn get_replace_call_char_arg_repr<'tcx>(arg: &'tcx hir::Expr<'tcx>) -> Option<String> {
```
flip1995, line 200: I think this is not necessary. You practically doing type checking again. You can just use `utils::SpanlessEq` to compare the `to` expressions. It also will cover more cases.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+                .iter()
+                .map(|path_seg| path_seg.ident.name.as_str())
+                .collect();
+            Some(path_segment_ident_names.join("::"))
+        },
+        _ => None,
+    }
+}
+
+fn get_earliest_replace_call_span<'tcx>(expr: &'tcx hir::Expr<'tcx>) -> Span {
```
flip1995, line 236: This should be a one-liner, if you collect the to-be-linted replace calls in a more efficient way.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,275 @@
...
+        } else {
+            ControlFlow::Continue(())
+        }
+    });
+
+    earliest_replace_call_span
+}
+
+/// Find the original receiver of a chain of `str::replace` method calls.
+fn find_original_recv<'tcx>(recv: &'tcx hir::Expr<'tcx>) -> &'tcx hir::Expr<'tcx> {
```
flip1995, line 257: Same here.

clippy_lints/src/methods/collapsible_str_replace.rs
flip1995: On second thought, maybe you want to do this while collecting the `MethodCall` expressions. 

But the method should have a signature like:

```rust
fn collect_replace_calls(..., to_arg: &Expr) -> ReplaceMethods
```

where 

```rust
struct ReplaceMethods {
    methods: Vec<&Expr>,
    args: Vec<&Expr>,
}
```

and in that function you should collect as long as (pseudo code)

```rust
SpanlessEq(to_arg, current_to_arg) && expr_ty(arg).peel_refs().is_char()
```

flip1995 (changes requested):
This implementation is way too complicated. You walk the method chain way too often with `for_each_expr`. You should walk it once and then return a vector of expressions to all the method calls, that you then reuse for all other checks you want to do.

I left multiple suggestions on how to simplify the code. Just comment if the suggestions are unclear or you have questions about them.

clippy_lints/src/methods/collapsible_str_replace.rs
nahuakang: This might be the only one I'm keeping so far as I don't know how else to get the value out of an expression as `"l"` or `'l'` or `l` for suggestion :) Please let me know if there's something I could use?

flip1995: I don't know why you would want to get the value out at all. You can just use the `SpanlessEq` util and compare the `to` arguments of the `replace` methods with that. You don't really care about the actual value, as long as the expression is the same.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+
+use super::method_call;
+use super::COLLAPSIBLE_STR_REPLACE;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'tcx>, to: &'tcx hir::Expr<'tcx>) {
+    if_chain! {
+        if let Some(replace_methods) = collect_replace_calls(cx, expr, to);
+        if let Some(earliest_replace_call) = replace_methods.methods.last();
+        if let Some((_, [earliest_recv, ..], _)) = method_call(earliest_replace_call);
+        if cx.typeck_results().expr_ty(earliest_recv).peel_refs().is_str();
```
flip1995, line 23: Please don't use `if_chain`

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        then {
+            // If the parent node's `to` argument is the same as the `to` argument
+            // of the last replace call in the current chain, don't lint as it would be duplicated lint
```
flip1995, line 27:
```suggestion
            // of the last replace call in the current chain, don't lint as was already linted
```

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+use rustc_span::source_map::Spanned;
+use rustc_span::Span;
+
+use super::method_call;
+use super::COLLAPSIBLE_STR_REPLACE;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'tcx>, to: &'tcx hir::Expr<'tcx>) {
+    if_chain! {
+        if let Some(replace_methods) = collect_replace_calls(cx, expr, to);
+        if let Some(earliest_replace_call) = replace_methods.methods.last();
```
flip1995, line 21: You should use a `Deque` instead of a `Vec` and use `push_front`, so that the logic is a bit cleaner.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+use rustc_lint::LateContext;
+use rustc_span::source_map::Spanned;
+use rustc_span::Span;
+
+use super::method_call;
+use super::COLLAPSIBLE_STR_REPLACE;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'tcx>, to: &'tcx hir::Expr<'tcx>) {
+    if_chain! {
+        if let Some(replace_methods) = collect_replace_calls(cx, expr, to);
```
flip1995, line 20: You can add an `if` condition here that `replace_methods.methods.len() > 1`

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+    to_arg: &'tcx hir::Expr<'tcx>,
+) -> Option<ReplaceMethods<'tcx>> {
+    let mut methods = Vec::new();
+    let mut from_args = Vec::new();
+    let mut return_none = false;
+
+    let _: Option<()> = for_each_expr(expr, |e| {
+        if let Some((name, [_, args @ ..], _)) = method_call(e) {
+            match (name, args) {
+                ("replace", [from, to]) => {
```
flip1995, line 61:
```suggestion
        if let Some(("replace", [_, from, to], _)) = method_call(e) {
```

should do the same.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+    });
+
+    if return_none {
+        return None;
+    }
+
+    Some(ReplaceMethods {
+        methods,
+        args: from_args,
```
flip1995, line 86: Why not name this field `from_args`. It seems to be the better name :sweat_smile:

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+            if let Some(parent) = get_parent_expr(cx, expr)
+                && let Some((name, [_, _, current_to], _)) = method_call(parent)
+                && name == "replace"
+                && SpanlessEq::new(cx).eq_expr(to, current_to)
+            {
+                return;
+            }
+
+            // Check if there's an earlier `str::replace` call
+            if let Some((_, [..], span_lo)) = method_call(earliest_replace_call) {
```
flip1995, line 37: `replace_methods` already contains this information. No need to check it again here.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+                ("replace", [from, to]) => {
+                    if SpanlessEq::new(cx).eq_expr(to_arg, to)
+                        && cx.typeck_results().expr_ty(from).peel_refs().is_char()
+                    {
+                        methods.push(e);
+                        from_args.push(from);
+                        ControlFlow::Continue(())
+                    } else {
+                        return_none = true;
+                        ControlFlow::BREAK
```
flip1995, line 70: This is not quite correct. E.g.
```rust
.replace(_, "x").replace(_, "t").replace(_, "t")
```

would return `None` even though the last 2 replace calls could be linted.

I'm not sure if or why I suggested to return an `Option` here, but you can just return the ReplaceMethods struct and then infer if you want to lint by checking if the methods Vec has more than one element.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+                segments: path_segments,
+                ..
+            },
+        )) => {
+            // join the path_segments values by "::"
+            let path_segment_ident_names: Vec<&str> = path_segments
+                .iter()
+                .map(|path_seg| path_seg.ident.name.as_str())
+                .collect();
+            Some(path_segment_ident_names.join("::"))
```
flip1995, line 156: This could just be `snippet(expr.span, "_")`

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+/// - the variable as a String, e.g. `l`
+fn get_replace_call_char_arg_repr<'tcx>(arg: &'tcx hir::Expr<'tcx>) -> Option<String> {
+    match arg.kind {
+        ExprKind::Lit(Spanned {
+            node: LitKind::Str(to_arg_val, _),
+            ..
+        }) => {
+            let repr = to_arg_val.as_str();
+            let double_quote = "\"";
+            Some(double_quote.to_owned() + repr + double_quote)
```
flip1995, line 134: Why not just `format!(r#""{to_arg_val}""#)`?

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut methods = Vec::new();
+    let mut from_args = Vec::new();
+    let mut return_none = false;
+
+    let _: Option<()> = for_each_expr(expr, |e| {
+        if let Some((name, [_, args @ ..], _)) = method_call(e) {
+            match (name, args) {
+                ("replace", [from, to]) => {
+                    if SpanlessEq::new(cx).eq_expr(to_arg, to)
+                        && cx.typeck_results().expr_ty(from).peel_refs().is_char()
```
flip1995, line 63: You were right. This should not only check for `char` but also strings, and `paths`. So pretty much the same you do later when formatting the `from_args`. It should probably share the same code. So I would add a function

```rust
fn is_valid_from_arg(...) {
    matches!(expr.kind, Str | Char | Path)
}
```

And then simplify the `get_replace_call_char_arg_repr` function to

```rust
snippet(from_arg, "_").replace_start("'", '"').replace_end("'", '"')
```

You don't even have to check for the `ExprKind` then, because if it would be of a wrong `ExprKind` the `from_arg` would've never been added to the `from_args` list.

flip1995 (changes requested):
Some more suggestiong on how to simplify the code. Thanks for addressing this so quickly!

clippy_lints/src/methods/collapsible_str_replace.rs
flip1995: I addressed this again in my review below.

clippy_lints/src/methods/collapsible_str_replace.rs
flip1995: See suggestion below

clippy_lints/src/methods/collapsible_str_replace.rs
flip1995: See suggestion below

clippy_lints/src/methods/collapsible_str_replace.rs
nahuakang: > You don't even have to check for the ExprKind then, because if it would be of a wrong ExprKind the from_arg would've never been added to the from_args list.
You're right here. The `is_valid_from_arg` is also not necessary since `cx.typeck_results().expr_ty(from).peel_refs().is_char()` checks if the underlying value of a `Path` is a `char` and we don't check for `str` in this iteration. So in the end we only need `snippet`.

clippy_lints/src/methods/collapsible_str_replace.rs
```diff
@@ -0,0 +1,90 @@
...
+    to_arg: &'tcx hir::Expr<'tcx>,
+) {
+    let from_args = &replace_methods.from_args;
+    let from_arg_reprs: Vec<String> = from_args
+        .iter()
+        .map(|from_arg| snippet(cx, from_arg.span, "..").to_string())
+        .collect();
+    let app = Applicability::MachineApplicable;
+    let earliest_replace_call = replace_methods.methods.front().unwrap();
+    if let Some((_, [..], span_lo)) = method_call(earliest_replace_call) {
```
flip1995, line 75: Couldn't you get the span with just `earliest_replace_call.span`? :thinking:

tests/ui/collapsible_str_replace.rs
```diff
@@ -0,0 +1,68 @@
...
+    let _ = misspelled.replace(s, "l").replace(u, "l").replace(p, "l");
+
+    let _ = misspelled.replace('s', "l").replace('u', "l").replace('p', "x");
+
+    let _ = misspelled.replace('s', "x").replace('u', "l").replace('p', "l");
+
+    // NO LINT CASES
+    let _ = misspelled.replace('s', "l");
+
+    let _ = misspelled.replace(s, "l");
```
flip1995, line 45: Every test case with only one `replace` is more or less redundant. This is because the code won't even get to the point that it checks the `replace` arguments. I think the test you wrote are viable, but you should add another `replace` to the single-replace cases.

tests/ui/collapsible_str_replace.rs
```diff
@@ -0,0 +1,68 @@
...
+fn main() {
+    let misspelled = "hesuo worpd";
+
+    let p = 'p';
+    let s = 's';
+    let u = 'u';
+    let l = "l";
+
+    // LINT CASES
+    let _ = misspelled.replace('s', "l").replace('u', "l");
```
flip1995, line 18: IIUC only `char`s in the `from_arg` position get linted now, correct? So not even strings. I think this is an easy extension to also lint strings that can be done in this PR.

I really wouldn't use the `expr_ty` for that, but really the `ExprKind`.

tests/ui/collapsible_str_replace.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    let _ = misspelled.replace(&['s', 'u'], "l").replace('p', "l");
+
+    let _ = misspelled.replace(&['s', u, 'p'], "l");
+
+    let _ = misspelled.replace(&[s, u, 'p'], "l");
+
+    let _ = misspelled.replace(&[s, u, p], "l");
+
+    let _ = misspelled.replace(&[s, u], "l").replace(&[u, p], "l");
```
flip1995, line 67: Please add a comment here, that this could get linted in the future and combined to `[s, u, p]`.

tests/ui/collapsible_str_replace.rs
```diff
@@ -0,0 +1,68 @@
...
+}
```
flip1995, line 68: Please also add another test that uses `iter.next().unwrap()` in the `to` place. It won't get linted in the current implementation, but it's good to have it as an regression test.

flip1995 (commented):
This review round, that comments are more about the tests. The code looks way cleaner and easier to understand now!

clippy_lints/src/methods/collapsible_str_replace.rs
nahuakang: I tried that before but the lint would cover `misspelled.replace.(...)...` so I needed the span inside.

tests/ui/collapsible_str_replace.rs
nahuakang: Actually, if you [play this playground code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82605f473fd8895d1013628957753846), I don't think we can lint strings this way. The [pattern](https://doc.rust-lang.org/std/str/pattern/trait.Pattern.html) must be `&[char]`.

I've improved and added tests to handle the following pattern though:

```rust
let _ = "hesudo worpd".replace(d, "l").replace('p', "l").replace("su", "l");
```

tests/ui/collapsible_str_replace.rs
```diff
@@ -0,0 +1,73 @@
...
+    let _ = "hesudo worpd".replace("su", "l").replace('d', "l").replace('p', "l");
+
+    let _ = "hesudo worpd".replace(d, "l").replace('p', "l").replace("su", "l");
+
+    // NO LINT CASES
+    let _ = "hesuo world".replace('s', "l").replace('u', "p");
+
+    let _ = "hesuo worpd".replace('s', "l").replace('p', l);
+
+    let _ = "hesuo world".replace(&get_filter(), "l").replace('s', "l");
```
flip1995, line 52: This doesn't get linted because `get_filter` returns a `str`, right? What if you write the same test case with a function that returns `char`?

flip1995 (commented):
Linting logic LGTM now. If the 2 test suggestions pass, this is ready to get merged.

flip1995:
@bors r+

Thanks! That implementation turned out nicely!

nahuakang:
@flip1995 @xFrednet Thank you both for your patience and help 🙏 
It's absolutely just wonderful to come back to the project and receive so much support and help ❤️

PR #9270, opened 2022-07-31
Make it easier to find the developer guide

schubart:
changelog: none

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,10 +30,10 @@ All contributors are expected to follow the [Rust Code of Conduct].
 ## The Clippy book
 
 If you're new to Clippy and don't know where to start the [Clippy book] includes
-a developer guide and is a good place to start your journey.
+a [developer guide] and is a good place to start your journey.
 
-<!-- FIXME: Link to the deployed book, once it is deployed through CI -->
-[Clippy book]: book/src
+[Clippy book]: https://doc.rust-lang.org/nightly/clippy/index.html
+[developer guide]: https://doc.rust-lang.org/nightly/clippy/development/index.html
 
 ## High level approach
 


CONTRIBUTING.md
```diff
@@ -30,10 +30,11 @@ All contributors are expected to follow the [Rust Code of Conduct].
 ## The Clippy book
 
 If you're new to Clippy and don't know where to start the [Clippy book] includes
-a developer guide and is a good place to start your journey.
+a [developer guide] and is a good place to start your journey.
 
 <!-- FIXME: Link to the deployed book, once it is deployed through CI -->
 [Clippy book]: book/src
+[developer guide]: book/src/development
```
xFrednet, line 37: Could you maybe update this, to link to the book version instead of the markdown files? :)

xFrednet (commented):
Hey and welcome to Clippy. Thank you for looking out for new contributors. These things sometimes slip under the radar since we know where for find most things. Getting feedback from newcomers is quite helpful to improve on this 🙃

CONTRIBUTING.md
schubart: I'd be happy to, but I can't seem to find the book version. What's the URL, please?

xFrednet: The book is quite new and a bit harder to find then I would like. :sweat_smile: Could you also update the `[Clippy book]` link? Here are the links :upside_down_face: 

```md
[Clippy book]: https://doc.rust-lang.org/nightly/clippy/index.html
[developer guide]: https://doc.rust-lang.org/nightly/clippy/development/index.html
```

schubart: Tanks, done.

I did it by fixing up my commit and force-pushing it. I hope that's OK. It probably does not matter for such a small change, but what's the policy for incorporating PR feedback? Fix up and force push or add additional commits?

xFrednet: That's perfect! There is no real policy, enforcing it would be more work than it's worth since we have so many contributors. GitHub used to not displaying force pushes correctly, but now it's a viable alternative. Generally, I do the following:

* On small PRs like this one, just fixup and force push, that's totally fine and allows us to merge the PR directly
* For larger PRs with more changes I create a normal new commit describing the change
* For larger PRs with small changes or suggestions, I create a commit like "Address review feedback" and then squash sequential changes with that one.

It's also fine to add a new commit for every review. If there are multiple/too many commits, we just ask you to squash them right before the merge.

xFrednet (approved):

schubart:
Thanks @xFrednet for your support. Making my first contribution to Clippy was a really pleasant experience!

PR #9273, opened 2022-08-01
Move [`assertions_on_result_states`] to restriction

### LINKED ISSUE #9263: `assertions_on_result_states` leads to less clear tests
### Summary

IMO using `assert!(foo().is_err())` is clearer within a test to imply that the assertion is the point of the test rather than `foo().unwrap_err()`.

### Lint Name

assertions_on_result_states

### Reproducer

I tried [this code](https://github.com/Nullus157/stylish-rs/blob/e92b0bc2e223f9d6c40affc31e9b6744b327dd92/tests/tests.rs#L58-L71):

```rust
    #[test]
    fn error() {
        struct Foo;

        impl stylish::Display for Foo {
            fn fmt(&self, _: &mut stylish::Formatter<'_>) -> std::fmt::Result {
                Err(std::fmt::Error)
            }
        }

        let mut s = Vec::<u8>::new();
        let mut writer = stylish::io::Plain::new(&mut s);
        assert!(stylish::writeln!(writer, "{:s}", Foo).is_err());
    }
```

I [saw this happen](https://github.com/Nullus157/stylish-rs/runs/7588631386?check_suite_focus=true):

```
error: called `assert!` with `Result::is_err`
Error:   --> tests/tests.rs:70:9
   |
70 |         assert!(stylish::writeln!(writer, "{:s}", Foo).is_err());
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace with: `stylish::writeln!(writer, "{:s}", Foo).unwrap_err()`
   |
   = note: `-D clippy::assertions-on-result-states` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_result_states
```

### Version

```text
info: syncing channel updates for 'nightly-x86_64-unknown-linux-gnu'
info: latest update on 2022-07-30, rust version 1.64.0-nightly (3924dac7b 2022-07-29)
```

### Additional Labels

_No response_

tabokie:
Close #9263

This lint causes regression on readability of code and log output. And printing runtime values is not particularly useful for majority of tests which should be reproducible.

changelog: Move [`assertions_on_result_states`] to restriction and don't lint it for unit type

Signed-off-by: tabokie <xy.tao@outlook.com>

### DIFF
--- a/clippy_lints/src/assertions_on_result_states.rs
+++ b/clippy_lints/src/assertions_on_result_states.rs
@@ -53,13 +53,14 @@ impl<'tcx> LateLintPass<'tcx> for AssertionsOnResultStates {
                 if result_type_with_refs != result_type {
                     return;
                 } else if let Res::Local(binding_id) = path_res(cx, recv)
-                    && local_used_after_expr(cx, binding_id, recv) {
+                    && local_used_after_expr(cx, binding_id, recv)
+                {
                     return;
                 }
             }
             let mut app = Applicability::MachineApplicable;
             match method_segment.ident.as_str() {
-                "is_ok" if has_debug_impl(cx, substs.type_at(1)) => {
+                "is_ok" if type_suitable_to_unwrap(cx, substs.type_at(1)) => {
                     span_lint_and_sugg(
                         cx,
                         ASSERTIONS_ON_RESULT_STATES,
@@ -73,7 +74,7 @@ impl<'tcx> LateLintPass<'tcx> for AssertionsOnResultStates {
                         app,
                     );
                 }
-                "is_err" if has_debug_impl(cx, substs.type_at(0)) => {
+                "is_err" if type_suitable_to_unwrap(cx, substs.type_at(0)) => {
                     span_lint_and_sugg(
                         cx,
                         ASSERTIONS_ON_RESULT_STATES,
@@ -98,4 +99,8 @@ fn has_debug_impl<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
     cx.tcx
         .get_diagnostic_item(sym::Debug)
         .map_or(false, |debug| implements_trait(cx, ty, debug, &[]))
+}
+
+fn type_suitable_to_unwrap<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
+    has_debug_impl(cx, ty) && !ty.is_unit() && !ty.is_never()
 }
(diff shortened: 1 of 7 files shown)


xFrednet:
I agree with these changes @flip1995, the lint was added 10 days ago in #9225. If we merge it now, it won't make it on the beta branch, if I'm correct. Do we maybe want to do the change in rust-lang/rust or backport it later?

(This is another place where nightly lints would be so helpful. It's on my todo list after `lint_reasons` has been stabilized :sweat_smile: )

flip1995:
We would need an out-of-cycle sync to get it into the rust repo before the release. I'm good with that, as long as it doesn't add too many other things (like new lints). 

We can also just do the `restriction` move in the Rust repo and let the other changes in this PR ride the train. That way, we don't need a sync.

clippy_lints/src/assertions_on_result_states.rs
```diff
@@ -96,3 +100,15 @@ fn has_debug_impl<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
         .get_diagnostic_item(sym::Debug)
         .map_or(false, |debug| implements_trait(cx, ty, debug, &[]))
 }
+
+fn is_unit_type(ty: Ty<'_>) -> bool {
```
flip1995, line 104: Can't you call `ty.is_unit() || ty.is_never()` instead of this function?

flip1995 (commented):
Only one comment, otherwise LGTM

clippy_lints/src/assertions_on_result_states.rs
tabokie: Good idea. Was copying it from another older lint before.

flip1995: Oh which one? We should also fix it there... (not in this PR though.

flip1995:
@bors r+

Thanks!

clippy_lints/src/assertions_on_result_states.rs
tabokie: Here: https://github.com/rust-lang/rust-clippy/blob/7c427f0fdadf07c117d23d0b3352a4870c9b17d3/clippy_lints/src/map_unit_fn.rs#L99

flip1995: Are you motivated to fix it also there and open a PR for it? :)

tabokie: Sure, will fix it tomorrow.

PR #9279, opened 2022-08-02
chore: a few small improvements to code quality

tabokie:
Some improvements:
- Simplify implementation of `is_unit_type`
- Use slice matching to destruct `Call` or `MethodCall` whenever possible

changelog: none

r? @flip1995

### DIFF
--- a/clippy_lints/src/map_err_ignore.rs
+++ b/clippy_lints/src/map_err_ignore.rs
@@ -113,10 +113,10 @@ impl<'tcx> LateLintPass<'tcx> for MapErrIgnore {
         }
 
         // check if this is a method call (e.g. x.foo())
-        if let ExprKind::MethodCall(method, args, _) = e.kind {
+        if let ExprKind::MethodCall(method, [_, arg], _) = e.kind {
             // only work if the method name is `map_err` and there are only 2 arguments (e.g. x.map_err(|_|[1]
             // Enum::Variant[2]))
-            if method.ident.as_str() == "map_err" && args.len() == 2 {
+            if method.ident.name == sym!(map_err) {
                 // make sure the first argument is a closure, and grab the CaptureRef, BodyId, and fn_decl_span
                 // fields
                 if let ExprKind::Closure(&Closure {
@@ -124,7 +124,7 @@ impl<'tcx> LateLintPass<'tcx> for MapErrIgnore {
                     body,
                     fn_decl_span,
                     ..
-                }) = args[1].kind
+                }) = arg.kind
                 {
                     // check if this is by Reference (meaning there's no move statement)
                     if capture_clause == CaptureBy::Ref {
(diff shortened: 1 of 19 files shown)


clippy_lints/src/map_err_ignore.rs
```diff
@@ -113,18 +113,18 @@ impl<'tcx> LateLintPass<'tcx> for MapErrIgnore {
         }
 
         // check if this is a method call (e.g. x.foo())
-        if let ExprKind::MethodCall(method, args, _) = e.kind {
+        if let ExprKind::MethodCall(method, [_, arg], _) = e.kind {
             // only work if the method name is `map_err` and there are only 2 arguments (e.g. x.map_err(|_|[1]
             // Enum::Variant[2]))
-            if method.ident.as_str() == "map_err" && args.len() == 2 {
+            if method.ident.as_str() == "map_err" {
```
flip1995, line 119: This should then also be changed:

```suggestion
            if method.ident == sym!(map_err) {
```

flip1995 (changes requested):
Thanks for addressing for going above and beyond with improving the code! The destructuring of arguments is a really nice change.

The symbol comparison changes however are not a code improvement. `as_str` on symbols is not a no-op, but rather an indexing of a vector requiring to acquire a lock first. See [here](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_span/symbol.rs.html#1936-1938). Interning an already interned string is pretty much the same, just that it does a `HashMap` lookup. But then the string comparison is more expensive than the symbol comparison, because the symbol comparison is just an integer comparison. Because those `check_*` functions are ran on every expr/stmt/item/..., it is in the end cheaper to intern the string once and then do a symbol comparison rather than a string comparison. 

(The `as_str` function additionally goes through a [`with_session_globals`](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_span/lib.rs.html#147-149), but I don't know how expensive that is).

TL;DR: Please revert the symbol changes, the rest is great.

tabokie:
@flip1995 Okay. Having looked closer, it's basically `lock+strcmp`(`as_str`) versus `lock+strhash+intcmp`(`intern`). I would argue that `strcmp` might be faster than `strhash` because it can exit early. But I agree whatever the perf difference is, it doesn't matter that much to be fixed this late.

flip1995:
Another non-perf related advantage is, that if a symbol should get pre-inserted by rustc, we have a lint that will tell us to remove the `sym!` macro call and just compare to `sym::`. (I don't think we get this with `as_str()`).

Yeah, your right, it might have equal performance. It probably doesn't make much of a difference. But I'd rather keep the `sym!` approach.

flip1995 (approved):

flip1995:
Perfect, thank you!

@bors r+

PR #9282, opened 2022-08-02
add paren before '?' when suggesting deref for clone_on_copy

### LINKED ISSUE #9277: clippy::clone_on_copy fix on code with `?` leads to code with wrong operator precedence
### Summary

`?` has higher precedence than `*` (dereference), but `cargo clippy --fix` will not parenthesize an expression when it introduces a `*` in an expression ending in `?`, at least for the `clone_on_copy` lint, leading to broken code.

The correct fix requiring explicit parentheses could be seen as the lint being a false positive in that case  - that's subjective of course.

### Reproducer

I tried this code:

```rust
fn foo() -> Option<()> {
    let opt = &None;

    let value = opt.clone()?;  // <--

    Some(value)
}
```

I expected to see this happen: The marked line should be fixed to be `let value = (*opt)?;`

Instead, this happened: The line became `let value = *opt?;`, i.e. equivalent to `let value = *(opt?);`, which does not compile.

### Version

```text
binary: rustc
commit-date: 2022-06-27
release: 1.62.0
LLVM version: 14.0.5
```

### Additional Labels

macovedj:
changelog: none

fixes #9277

### DIFF
--- a/clippy_lints/src/methods/clone_on_copy.rs
+++ b/clippy_lints/src/methods/clone_on_copy.rs
@@ -4,7 +4,7 @@ use clippy_utils::source::snippet_with_context;
 use clippy_utils::sugg;
 use clippy_utils::ty::is_copy;
 use rustc_errors::Applicability;
-use rustc_hir::{BindingAnnotation, Expr, ExprKind, MatchSource, Node, PatKind};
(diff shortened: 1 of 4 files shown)


flip1995:
@bors r+

Thanks!

PR #9283, opened 2022-08-02
Add `if_none_blocks` lint

### LINKED ISSUE #9094: none_else_block
### What it does

transform if else block into then_some() call if else block returns None

### Lint Name

none_else_block

### Category

pedantic

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
let a = true;

if a == true {
      Some("asd")
} else {
      None
}
  |
  v
a.then_some("asd");
```

mkrasnitski:
Closes #9094; this lint detects `None` blocks in both `then` and `else` arms. Some thoughts I have right now are:
 - I think the error messages could be prettier, maybe by using a MultiSpan? I'd like some help with this.
 - I think a special case help message can be introduced if the condition should be negated, and the expression is a single variable. Instead of saying "first consider negating the condition", say "first consider negating `a`".
   - On a related note, should the case when the condition is a single variable be the *only* time this lint is enabled, instead of on all boolean conditions? I'm not sure if `(long_boolean_condition_chain_here).then_some()` is more or less idiomatic than a simple if statement.
 - Should suggestion be enabled for this lint? Maybe some heuristic could be used in cases where the block returning `Some(_)` is quite complex.

This is my first contribution to Rust, so I'm excited but also a bit nervous :) Bikeshedding on the lint's name and error/help messages is appreciated, to get the wording right.

Finally, this currently changes the stderr for one ui test (`filter_map_next`) and also catches one thing when dogfooding, but I'll wait to resolve those until point 2 above has been settled.

PR checklist:
- \[ ] `cargo test` passes locally
- \[x] Run `cargo dev fmt`

---

changelog: [`if_none_blocks`]: Add lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3610,6 +3610,7 @@ Released 2018-09-13
 [`if_let_mutex`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_let_mutex
 [`if_let_redundant_pattern_matching`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_let_redundant_pattern_matching
 [`if_let_some_result`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_let_some_result
+[`if_none_blocks`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_none_blocks
 [`if_not_else`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else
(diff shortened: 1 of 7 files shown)


mkrasnitski:
Turns out [`if_then_some_else_none`](https://rust-lang.github.io/rust-clippy/master/index.html#if_then_some_else_none) already exists, which lints this exact situation. However, it could serve to be expanded to also suggest `bool::then_some`, not just `bool::then`, so I'll try adding that instead.

PR #9287, opened 2022-08-03
`transmute_undefined_repr` fix

Jarcho:
changelog: Don't lint `transmute_undefined_repr` when the the first field of a `repr(C)` type is compatible with the other type

### DIFF
--- a/tests/ui/transmute_undefined_repr.rs
+++ b/tests/ui/transmute_undefined_repr.rs
@@ -4,6 +4,7 @@
 use core::any::TypeId;
 use core::ffi::c_void;
 use core::mem::{size_of, transmute, MaybeUninit};
+use core::ptr::NonNull;
 
 fn value<T>() -> T {
     unimplemented!()
@@ -117,6 +118,9 @@ fn main() {
 
         let _: *const Ty2<u32, u32> = transmute(value::<*const Ty2C<u32, Ty2<u32, u32>>>()); // Err
         let _: *const Ty2C<u32, Ty2<u32, u32>> = transmute(value::<*const Ty2<u32, u32>>()); // Err
+
+        let _: NonNull<u8> = transmute(value::<NonNull<(String, String)>>()); // Ok
+        let _: NonNull<(String, String)> = transmute(value::<NonNull<u8>>()); // Ok
     }
 }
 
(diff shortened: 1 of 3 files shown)


tests/ui/transmute_undefined_repr.rs
```diff
@@ -109,6 +109,14 @@ fn main() {
         let _: Ty2<u32, u32> = transmute(value::<MaybeUninit<Ty2<u32, u32>>>()); // Ok
 
         let _: Ty<&[u32]> = transmute::<&[u32], _>(value::<&Vec<u32>>()); // Ok
+
+        let _: *const Ty2<u32, u32> = transmute(value::<*const Ty2C<Ty2<u32, u32>, u32>>()); // Ok
+        let _: *const Ty2C<Ty2<u32, u32>, u32> = transmute(value::<*const Ty2<u32, u32>>()); // Ok
```
xFrednet, line 114: Why is this okay? This transmutes a `Ty2<u32, u32>` to a larger value. Wouldn't this mean that the last `u32` of `Ty2C` is undefined?

xFrednet (commented):
The changes look good, just one question regarding the output :upside_down_face:

tests/ui/transmute_undefined_repr.rs
Jarcho: This would be the same as down-casting a type. Not necessarily correct, but it's possible.

xFrednet: I feel like that should be linted, but that's not necessarily the scope of this PR. :upside_down_face:

xFrednet:
Could you have a look at the CI? Then we can merge this :)

tests/ui/transmute_undefined_repr.rs
Jarcho: It would be better as a different lint. This is for transmutes where none of the fields have a defined value because they could be in an unexpected order.

Jarcho:
The most recent change allows casts between `*const/mut int_type` and any type as a form of type erasure. `*u8/i8` are used as an untyped buffer. I've also seen `*u16/i16` used due to windows api's counting by them.

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update!

@bors r+

Issue #9275, opened 2022-08-01
Don't use PartialEq to check if an Option is None


lukaslueg:
### What it does

I've seen people use `foo == None` or `foo != None` to check if `Option<T>::is_none()` / `is_some()`. This works fine as long as `T` is `PartialEq`, yet is unnecessary and may lead to surprising type errors later on in the programmer's life when the `T` is not `PartialEq`. The pattern most likely stems from other programming languages, where checking for null is expressed as `foo is None` or something similar. The programmer may not be aware that the comparison piggybacks on `T: PartialEq`, which is only practically useful in the `Some`-case.

Equality comparisons with `Option::None` should not be done in that way, and we can recommend `Option::is_none()` / `is_some()` or structural comparison, which are strictly more powerful and work for any `Option<T>` regardless if `T: PartialEq`

### Lint Name

partialeq_to_none

### Category

style, pedantic

### Advantage

- Intention is expressed more clearly
- The triggering code is using the `PartialEq` trait on `T` in `Option<T>`, which the programmer is probably not aware of
- We can replace the triggering code with a strictly more powerful alternative.

### Example

```rust
pub fn foo(x: Option<u32>) -> &'static str {
    if x != None {
        "yay"
    } else {
        "nay"
    }
}
```

Could be written as:

```rust
pub fn foo(x: Option<u32>) -> &'static str {
    if x.is_some() {
        "yay"
    } else {
        "nay"
    }
}
```

jmj0502:
@xFrednet Hey! Hope you're doing great. I'm a Rust newbie and would like to make a contribution. Can I take this issue (I may need some guidance along the way)?

xFrednet:
@jmj0502 hey I'm doing great, hopefully you as well and welcome to Clippy! Sure you can claim it with `@rustbot claim`.

If you need any help feel free to ping me! To get started you might want to look at our documentation. `CONTRIBUTING.md` should contain links to it 🙃

jmj0502:
@xFrednet I'm glad you're doing cool! I'm doing great as well! c:
Thanks for all the support! Do have any idea of what files should I check to start working on this specific issue (aside from the `CONTRIBUTING.md` of course)?

xFrednet:
[...]

PR #9288, opened 2022-08-03
Add partialeq_to_none lint

lukaslueg:
Initial implementation of #9275, adding lint `partialeq_to_none`. This is my first time working on `clippy`, so please review carefully.

I'm unsure especially about the `Sugg`, as it covers the entire `BinOp`, instead of just covering one of the sides and the operator (see the multi-line example). I was unsure if pinpointing the suggestion wouldn't be brittle...

changelog: [`PARTIALEQ_TO_NONE`]: Initial commit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3823,6 +3823,7 @@ Released 2018-09-13
 [`panic_params`]: https://rust-lang.github.io/rust-clippy/master/index.html#panic_params
 [`panicking_unwrap`]: https://rust-lang.github.io/rust-clippy/master/index.html#panicking_unwrap
 [`partialeq_ne_impl`]: https://rust-lang.github.io/rust-clippy/master/index.html#partialeq_ne_impl
+[`partialeq_to_none`]: https://rust-lang.github.io/rust-clippy/master/index.html#partialeq_to_none
 [`path_buf_push_overwrite`]: https://rust-lang.github.io/rust-clippy/master/index.html#path_buf_push_overwrite
 [`pattern_type_mismatch`]: https://rust-lang.github.io/rust-clippy/master/index.html#pattern_type_mismatch
 [`possible_missing_comma`]: https://rust-lang.github.io/rust-clippy/master/index.html#possible_missing_comma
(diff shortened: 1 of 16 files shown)


Jarcho:
Implementation looks fine. Two things before merging:

* Should this be a `style` lint or a `pedantic` lint. I'm leaning towards `style`.
* I don't particularly like `PartialEq` in the lint name. There's nothing really wrong with it and it's not misleading, but it seems like unnecessary detail. `eq_none` might be better.

lukaslueg:
```rust
          // We are only interested in comparisons between `Option` and a literal `Option::None`
          let scrutinee = match (
              is_none_ctor(left_side) && is_ty_option(right_side),
              is_none_ctor(right_side) && is_ty_option(right_side),
          ) {
              (true, false) => right_side,
              (false, true) => left_side,
              _ => return,
          };
```

Do we even need the `is_ty_option()` (don't mind the `right_side` bug 👀😅) to rule out comparisons like `1f32 == None`? AFAICS, as a `LateLintPass` comes after type-resolution, non-applicable comparisons are already ruled out?!

lukaslueg:
Also: `x != &None` does not trigger because the `Expr` `&None` is not `is_lang_ctor(.., .., LangItem::OptionNone)`. How to peel off the references from a `QPath` akin to `Ty::peel_refs`?

Jarcho:
You do need to check for an option type as `PartialEq` can be implemented across types.

You can use `clippy_utils::peel_hir_expr_refs`.

lukaslueg:
Moving from `nursery` to `style` enables the lint by default, which causes it to get triggered in other ui-tests. For instance:

https://github.com/rust-lang/rust-clippy/blob/5721ca9a13562f5168c6ba9f43dc19bb76c6f7bb/tests/ui/manual_assert.rs#L19-L21

Should I update the triggering code in the ui-tests, bless the results, or `allow(clippy::partialeq_to_none)` in those ui-tests?

Jarcho:
Normally adding `allow` in tests is the better approach. It's possible the change can cause the test to no longer function.

For the lint name, use whichever one you prefer.

lukaslueg:
I've had to make some changes to get the suggestions right (see the `Box`-example). LGTM

Jarcho:
Looks good. You can squash the changes.

Jarcho (approved):

lukaslueg:
Squashed

Jarcho:
Thank you. @bors r+

trevyn:
I'm hitting this lint for code of the form `foo == option_env!("BAR")`, is that intended?

llogiq:
Perhaps we should add a macro check for the `None` expr.

lukaslueg:
@trevyn Thanks for the comment. The lint is certainly wrong as it would hardcode whatever `option_env` expanded to _at that time_.

PR #9289, opened 2022-08-04
Extend `if_then_some_else_none` to also suggest `bool::then_some`

### LINKED ISSUE #9094: none_else_block

mkrasnitski:
Closes #9094.

changelog: Extend `if_then_some_else_none` to also suggest `bool::then_some`

### DIFF
--- a/clippy_lints/src/if_then_some_else_none.rs
+++ b/clippy_lints/src/if_then_some_else_none.rs
@@ -1,7 +1,7 @@
 use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::eager_or_lazy::switch_to_eager_eval;
 use clippy_utils::source::snippet_with_macro_callsite;
 use clippy_utils::{contains_return, higher, is_else_clause, is_lang_ctor, meets_msrv, msrvs, peel_blocks};
-use if_chain::if_chain;
 use rustc_hir::LangItem::{OptionNone, OptionSome};
 use rustc_hir::{Expr, ExprKind, Stmt, StmtKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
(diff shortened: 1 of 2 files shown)


clippy_lints/src/if_then_some_else_none.rs
```diff
@@ -70,43 +70,47 @@ impl<'tcx> LateLintPass<'tcx> for IfThenSomeElseNone {
...
+            } else {
+                cond_snip.into_owned()
+            };
+            let arg_snip = snippet_with_macro_callsite(cx, then_arg.span, "");
+            let mut method_body = if then_block.stmts.is_empty() {
+                arg_snip.into_owned()
+            } else {
+                format!("{{ /* snippet */ {} }}", arg_snip)
+            };
+            let method_name = if switch_to_eager_eval(cx, expr) && meets_msrv(self.msrv, msrvs::BOOL_THEN_SOME) {
```
flip1995, line 95: Hmm, why does this MSRV already exist? What lint uses this?

mkrasnitski: It was added in #9099. This PR more or less brings this lint in line with that one.

flip1995: Ah cool, thanks!

xFrednet (approved):
Looks good to me. As a suggestion, could you mention `bool:then_some` in the lint description? The current example suggests `bool::then` which could/should be `bool::then_some`. As another change, you could just mention `bool::then_some` in the *"Why is this bad?"*  section. Something like:

> For simple calculation and known values you can also use `bool::then_some` which is eagerly avaluated in comparison to `bool:then`.

Otherwise, this is ready to merge. :+1: Thank you for the contribution and patience :upside_down_face:

mkrasnitski:
Thanks for the review! I've pushed some changes to the lint's description.

xFrednet (approved):

xFrednet:
Perfect, thank you for the implementation and quick follow up. I hope you had fun while working on Clippy 🙃

@bors r+

PR #9301, opened 2022-08-07
New `[suspicious_arguments]` lint for possibly swapped arguments

5225225:
changelog: [`suspicious_arguments`]: add lint for probably accidentally swapped arguments

---

### DIFF
--- a/clippy_lints/src/suspicious_arguments.rs
+++ b/clippy_lints/src/suspicious_arguments.rs
@@ -1,10 +1,10 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::path_res;
+use rustc_data_structure::fs::FxHashMap;
 use rustc_hir::{Expr, ExprKind, PatKind, QPath};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::Span;
-use std::collections::HashMap;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -39,7 +39,7 @@ declare_clippy_lint! {
 declare_lint_pass!(SuspiciousArguments => [SUSPICIOUS_ARGUMENTS]);
 
 fn arguments_are_sus(cx: &LateContext<'_>, definition: &[(String, Span)], call: &[Option<(String, Span)>]) {
-    let idxs: HashMap<&String, usize> = definition
+    let idxs: FxHashMap<&String, usize> = definition
         .iter()
         .enumerate()
         .map(|(idx, (item, _))| (item, idx))
--- a/clippy_lints/src/suspicious_arguments.rs
+++ b/clippy_lints/src/suspicious_arguments.rs
@@ -1,0 +1,141 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use rustc_hir::{def, def_id, Expr, ExprKind, PatKind, QPath};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+use std::collections::HashMap;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Checks for calls to a function where the parameters look swapped
+    ///
+    /// ### Why is this bad?
+    ///
+    /// This likely indicates an error, where the arguments were reversed.
+    ///
+    /// ### Example
+    /// ```rust
+    /// fn resize(width: usize, height: usize) {}
+    ///
+    /// let height = 100;
+    /// let width = 200;
+    /// resize(height, width);
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn resize(width: usize, height: usize) {}
+    ///
+    /// let height = 100;
+    /// let width = 200;
+    /// resize(width, height);
+    /// ```
+    #[clippy::version = "1.64.0"]
+    pub SUSPICIOUS_ARGUMENTS,
+    suspicious,
+    "function call with probably swapped arguments"
+}
+declare_lint_pass!(SuspiciousArguments => [SUSPICIOUS_ARGUMENTS]);
+
+
+fn arguments_are_sus(cx: &LateContext<'_>, definition: &[(String, Span)], call: &[Option<(String, Span)>]) {
+    let idxs: HashMap<&String, usize> = definition
+        .iter()
+        .enumerate()
+        .map(|(idx, (item, _))| (item, idx))
+        .collect();
+
+    for (call_idx, arg_and_span) in call.iter().enumerate() {
+        if let Some((arg, call_span)) = arg_and_span {
+            if let Some(&def_idx) = idxs.get(arg) {
+                if call_idx != def_idx {
+                    if let Some((reverse_call, reverse_call_span)) = &call[def_idx] {
+                        let def_for_call = &definition[call_idx];
+                        if reverse_call == &def_for_call.0 {
+                            // This is technically being called twice, but it's being
+                            // deduplicated?
+                            span_lint_and_then(
+                                cx,
+                                SUSPICIOUS_ARGUMENTS,
+                                vec![*call_span, *reverse_call_span],
+                                "these arguments are possibly swapped",
+                                |diag| {
+                                    let second_span = definition[def_idx].1;
+                                    diag.span_note(vec![def_for_call.1, second_span], "the arguments are defined here");
+                                },
+                            );
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for SuspiciousArguments {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if expr.span.from_expansion() {
+            return;
+        }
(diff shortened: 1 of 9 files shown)


clippy_lints/src/suspicious_arguments.rs
```diff
@@ -0,0 +1,141 @@
...
+                            local_def_index: defid.index,
+                        };
+                        let hirid = cx.tcx.hir().local_def_id_to_hir_id(local);
+                        hirid
+                    }
+                    QPath::LangItem(_, _, _) => { return; }
+                };
+
+
+                let Some(bodyid) = cx.tcx.hir().maybe_body_owned_by(hirid) else { return; };
```
Alexendoo, line 117: There's a few utilities here & there that can do the heavy lifting for this part (`path_res` being `clippy_utils::path_res`)

```suggestion
        if let ExprKind::Call(f, args) = expr.kind
            && let Some(def_id) = path_res(cx, f).opt_def_id()
            && let Some(node) = cx.tcx.hir().get_if_local(def_id)
            && let Some(body_id) = node.body_id()
        {
```

5225225: Ah thank you! that's much neater, wasn't aware of that.

5225225:
```rust
let defnames: Map<&str, usize> = ...;
let to_def_idx: Vec<Option<usize>> = vec![None; definition.len()];

for (i, _) in call.iter().enumerate() {
  // set to_def_idx[i] to Some(_) if the name matches a defname
}

let mut cycle = Vec::new();
for callidx in 0..to_def_idx.len() {
  cycle.clear();
  let mut cur = callidx;
  while let Some(defidx) = to_def_idx[cur].take() {
    cycle.push(cur);
    cur = callidx;
  }
  if Some(start) = cycle.iter().position(|e| *e == cur) {
    // cycle detected at `cycle[start..]`, emit lint
  }
}
```

> regarding bigger permutations, I feel it should be possible to do well
> and not that complex
> record the edges 'callsite name' -> 'param name' in a Vec<Option<usize>> 
> and you can just follow the graph along to find perms
> so, to_def_idx defines a (partial) permutation between the args indices
> then, the idea of my algorithm is to follow the edges of the permutation until you find a cycle

A thought of a better algorithm that would detect `tri_rotate` from @moulins

Untested but worth considering

moulins:
Here's a better version of the algorithm I sketched out yesterday on Discord, with bugs fixed and some explanatory comments:
```rs
// Take care of degenerate cases.
if def_args.len() != call_args.len() || call_args.len() < 2 {
  return;
}

// The mapping of argument names to their 'correct' position.
let def_names: Map<&str, usize> = ...;

// For each call argument, the index of the matching name in the definition.
let to_def_idx: Vec<Option<usize>> = def_args.iter().map(|arg| {
  arg_to_ident_str(arg).and_then(|name| def_names.get(name))
});

let mut chain = Vec::new();
for call_idx in 0..to_def_idx.len() {
  // Follow the 'chain' of indices until we get stuck.
  // We remove the links as we traverse them to avoid infinite loops in cycles.
  chain.clear();
  let mut cur = call_idx;
  while let Some(def_idx) = to_def_idx[cur].take() {
    cycle.push(cur);
    cur = def_idx;
  }

  // If the chain contains `cur`, then we have a cycle. Note that 1-cycles shouldn't be linted,
  // as they correspond to well-positioned arguments, and so we ignore the last element.
  let cycle_start = {
    let mut iter = chain.iter();
    iter.next_back();
    iter.position(|e| *e == cur)
  };

  if Some(start) = cycle_start {
    let cycle = &chain[start..];
    // We have found a cyclic permutation of the arguments, emit the lint.
  }
}
```

tests/ui/suspicious_arguments.stderr
```diff
@@ -0,0 +1,102 @@
...
+  --> $DIR/suspicious_arguments.rs:102:20
+   |
+LL |     std::mem::swap(y, x);
+   |                    ^  ^
+   |
+note: the arguments are defined here
+  --> /home/jess/.rustup/toolchains/nightly-2022-07-28-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/mem/mod.rs:711:22
+   |
+LL | pub const fn swap<T>(x: &mut T, y: &mut T) {
+   |                      ^          ^
```
Alexendoo, line 99: You can use [an `// aux-build` comment](https://rustc-dev-guide.rust-lang.org/tests/compiletest.html?highlight=aux-build#building-auxiliary-crates) to test the separate crate version without running into these path issues

clippy_lints/src/suspicious_arguments.rs
```diff
@@ -0,0 +1,140 @@
...
+        .map(|(idx, (item, _))| (item, idx))
+        .collect();
+
+    for (call_idx, arg_and_span) in call.iter().enumerate() {
+        if let Some((arg, call_span)) = arg_and_span {
+            if let Some(&def_idx) = idxs.get(arg) {
+                if call_idx != def_idx {
+                    if let Some((reverse_call, reverse_call_span)) = &call[def_idx] {
+                        let def_for_call = &definition[call_idx];
+                        if reverse_call == &def_for_call.0 {
```
llogiq, line 54: There is quite some rightwards drift here. We can use either the `if_chain` macro or the recently reverted-on-stable-but-available-on-nightly let-chains feature (basically `if let Some((arg, call_span)) = arg_and_span && let Some(&def_idx) = idxs.get(arg) && call_idx != def_idx && ..` etc.).

clippy_lints/src/suspicious_arguments.rs
```diff
@@ -0,0 +1,140 @@
...
+            // We *could* usefully lint on this (and it seems to work for *foreign* items, just not
+            // local ones?)
+            if cx.tcx.fn_sig(def_id).c_variadic() {
+                return;
+            }
+
+            let mut def_args = Vec::new();
+            for ident in cx.tcx.fn_arg_names(def_id) {
+                def_args.push((ident.to_string(), ident.span));
+            }
```
llogiq, line 124:
```suggestion
            let def_args = cx.tcx.fn_arg_names(def_id).map(|ident|{
                (ident.to_string(), ident.span)
            }).collect::<Vec<_>>();
```

clippy_lints/src/suspicious_arguments.rs
```diff
@@ -0,0 +1,140 @@
...
+
+            let mut call_args = Vec::new();
+
+            for call_arg in args {
+                call_args.push(guess_ident(call_arg));
+            }
```
llogiq, line 130:
```suggestion
            let call_args = args.map(guess_ident).collect::<Vec<_>>();
```

llogiq (approved):
Sorry for being absent for so long and thanks to @Alexendoo for taking over the review. I have a few nitpicks and one question, otherwise this looks mergeworthy.

tests/ui/suspicious_arguments.rs
```diff
@@ -0,0 +1,145 @@
+// edition:2015
```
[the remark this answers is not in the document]
5225225: Yeah, I was testing anonymous args, which you can only get with a trait in 2015. I'll move that to a different file.

llogiq:
Your call to `fn_arg_names` probably needs a bit more than just a bare `tcx`, as it seems to cause an ICE in the test. I'm not exactly sure why right now, might look into it later. Also you'll need to `cargo dev fmt` on the code.

afetisov:
It's not uncommon to have dummy parameter names which don't indicate anything, and thus cannot be reasonably used to guess accidental swaps. `mem::swap` in the tests is one such example, but there are many others. I would be very annoyed if the lint fired in those situations.

One example is arithmetic traits. It's common to implement unordered operation for different types based on a swap of some canonical order. E.g.
```rust
impl std::ops::Add<Foo> for Bar { /* snip */ }

impl std::ops::Add<Bar> for Foo {
    type Output = <Bar as Add<Foo>>::Output;
    fn add(&self, other: &Bar) -> Self::Output {
        std::ops::Add::add(other, self)
    }
}
```

I'm not sure if this specific lint will fire on this specific example (`self` probably doesn't have an associated identifier, does it?), but the pattern itself is quite common for essentially symmetric functions and traits.

It's hard to detect dummy identifiers in general, but the specific case of one-letter names is easy and common. It's unreasonable to expect that if I name my indexing variables `i`, `j`, `k`, or my closures `f`, `g`, `h`, then those names carry any meaning and must be matched with whatever functions I pass them to. And what should I do if the functions themselves name the arguments inconsistently? What should I do if the arguments with essentially the same semantics (e.g. closure for default value and for mapping, like in `Option::map_or_else`) are named `g`, `f` in one function, and `f`, `g` in the other?

Also, the lint ignores argument types (even has an explicit test for that with `itertools::all(f, iterable)` on line 93). But that would already be an error in the compiler. What's the point of emitting a warning on the same items and for the same reason as the error? That just causes notification overload.

blyxyas:
@5225225 I'm going to close this for now (365 days without activity), feel free to reopen if you still want to work on it! :heart:

PS: In [this blog post](https://www.5snb.club/posts/2020/distortion/), you can use [SVG Filters](https://yoksel.github.io/svg-filters/#/presets) instead of JS for better performance (letting the GPU do the rendering work instead of modifying the image with the CPU) ฅ՞•ﻌ•՞ฅ

Issue #9072, opened 2022-06-30
False positive `significant_drop_in_scrutinee` if matching on a `MutexGuard` even if it is still used afterwards


sdroege:
### Summary

See reproducer below

### Lint Name

significant_drop_in_scrutinee

### Reproducer

I tried this code:

```rust
fn main() {
    let x = std::sync::Mutex::new(vec![1, 2, 3]);
    let x_guard = x.lock().unwrap();
    match x_guard[0] {
        1 => println!("1!"),
        x => println!("{x}"),
    }
    drop(x_guard); // Some "usage"
}
```

I saw this happen:

```
warning: temporary with significant drop in match scrutinee
 --> src/main.rs:4:11
  |
4 |     match x_guard[0] {
  |           ^^^^^^^^^^
  |
  = note: `#[warn(clippy::significant_drop_in_scrutinee)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_in_scrutinee
help: try moving the temporary above the match
  |
4 ~     let value = x_guard[0];
5 ~     match value {
  |
```

I expected to see this happen: No warning

### Version

```text
rustc 1.63.0-beta.2 (6c1f14289 2022-06-28)
binary: rustc
commit-hash: 6c1f1428998a08864495bb599840a5997ec175f4
commit-date: 2022-06-28
host: x86_64-unknown-linux-gnu
release: 1.63.0-beta.2
LLVM version: 14.0.5
```

### Additional Labels

@rustbot label +I-false-positive

sdroege:
Arguably even without the later usage this shouldn't produce a warning because the `MutexGuard` is not only temporary during the `match` but outlives it.

fxdave:
Here it is another example:
```rust
let mut db = db.lock().unwrap();
match db.insert(WarehouseEntry {
    name: data.0.name,
    parent_id: data.0.parent_id.unwrap_or(db::ROOT_PARENT_ID.to_string()),
    variant: db::WarehouseEntryVariant::Item,
}) {
    Ok(entry) => HttpResponse::InternalServerError().json(CreateSuccessResponse { entry }),
    Err(error) => HttpResponse::InternalServerError().json(GenericError {
        error: error.to_string(),
    }),
}
```

However, the fix was easy in this case:
```rust
let mut db = db.lock().unwrap();
let result = db.insert(WarehouseEntry {
    name: data.0.name,
[...]

Issue #8987, opened 2022-06-12
significant_drop_in_scrutinee is confusing and not very useful for for loops


crumblingstatue:
### Description

The following code
```Rust
use std::sync::Mutex;

fn main() {
    let mutex_vec = Mutex::new(vec![1, 2, 3]);
    for number in mutex_vec.lock().unwrap().iter() {
        dbg!(number);
    }
}
```
produces this clippy output:
```Rust
warning: temporary with significant drop in for loop
 --> src/main.rs:5:19
  |
5 |     for number in mutex_vec.lock().unwrap().iter() {
  |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(clippy::significant_drop_in_scrutinee)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_in_scrutinee
  ```
The first confusing thing is the name. Scrutinee only makes sense for match, it's not used in the context of for loops.
If one visits the lint page, it doesn't talk about for loops at all, and what the ideal code would look like for for loops.

The naive thing to do here would be to bind the guard before the for loop (ignoring that it still triggers the lint (#8963)).

```Rust
use std::sync::Mutex;

fn main() {
    let mutex_vec = Mutex::new(vec![1, 2, 3]);
    let guard = mutex_vec.lock().unwrap();
    for number in guard.iter() {
        dbg!(number);
    }
}
```

However, that actually increases the scope of the guard, and now it will be locked for the entire lifetime of the guard binding.
The proper thing to do would be to drop it after the for loop.

```Rust
use std::sync::Mutex;

fn main() {
    let mutex_vec = Mutex::new(vec![1, 2, 3]);
    let guard = mutex_vec.lock().unwrap();
    for number in guard.iter() {
        dbg!(number);
    }
    drop(guard);
}
```

However, the lint page never talks about this, and this could easily be a trap for beginners/people not paying attention.
And is it really better than just locking in the for loop expression?
I personally think it's pretty clear that the guard lasts for the whole loop.
Abiding this lint just adds a new opportunity for error (dropping the guard late).

### Version

```text
rustc 1.63.0-nightly (ec55c6130 2022-06-10)
binary: rustc
commit-hash: ec55c61305eaf385fc1b93ac9a78284b4d887fe5
commit-date: 2022-06-10
host: x86_64-unknown-linux-gnu
release: 1.63.0-nightly
[...]

PR #9302, opened 2022-08-08
Move `significant_drop_in_scrutinee` into `nursey`

Jarcho:
The current suggestion of extending the lifetime of every sub-expression is not great and doesn't fix the error given in the lint's example, though it does make the potential deadlock easier to see, but it can also cause it's own issues by delaying the drop of the lock guard.

e.g.
```rust
match x.lock().foo {
    ..
}
// some stuff
let y = x.lock();
```
The suggestion would create a deadlock at the second `x.lock()` call.

This also lints even when a significant drop type isn't created as a temporary. (#9072)

I agree @kpreid (https://github.com/rust-lang/rust-clippy/issues/8987#issuecomment-1207464440) that this should be back-ported before the lint hits stable.

changelog: Move `significant_drop_in_scrutinee` into `nursey`

### DIFF
--- a/clippy_lints/src/lib.register_all.rs
+++ b/clippy_lints/src/lib.register_all.rs
@@ -144,7 +144,6 @@ store.register_group(true, "clippy::all", Some("clippy_all"), vec![
     LintId::of(matches::MATCH_STR_CASE_MISMATCH),
     LintId::of(matches::NEEDLESS_MATCH),
     LintId::of(matches::REDUNDANT_PATTERN_MATCHING),
-    LintId::of(matches::SIGNIFICANT_DROP_IN_SCRUTINEE),
     LintId::of(matches::SINGLE_MATCH),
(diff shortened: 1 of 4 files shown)


flip1995:
The `Vec::drain` bug was already fixed AFAIK. The for loop message was improved.

But yeah the suggestion of just moving it in front of the loop/match has other deadlock potential. This should probably be addressed by scoping the match block.

That it warns even if the mutex is not used inside the loop/match was intentional, but maybe should also be revisited.

But moving it back to nursery SGTM. Beta was already branched though. I try to backport it anyway.

@bors r+

mhnap:
I still find this lint very useful, as demonstrated [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e7b93719419c4fa42466b5480a8ca4bd).

PR #9308, opened 2022-08-09
Use `check_proc_macro` for `missing_const_for_fn`

### LINKED ISSUE #8854: `missing_const_for_fn` triggers for derives
### Summary

`missing_const_for_fn` accidentally triggers for code generated by derives which you generally don't have any influence on.

### Lint Name

missing_const_for_fn

### Reproducer

I tried this code:

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=458253aaa8421c3ef13bdc0c18bbef05)
```rust
#![warn(clippy::missing_const_for_fn)]

#[derive(Debug, Copy, Clone, bytemuck::Pod, bytemuck::Zeroable)]
#[repr(C)]
struct Header {
    version: u16,
    count: u16,
    storage_offset: u16,
}
```

I saw this happen:

```
warning: this could be a `const fn`
 --> src\layout\parser\font_resolving\name.rs:7:1
  |
7 | #[repr(C)]
  | ^
  |
 --> src\lib.rs:6:5
  |
6 |     clippy::missing_const_for_fn,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn
```

Weirdly the span is a bit different on the playground.

The lint shouldn't trigger. Also even if it would trigger, the span is somehow on the `repr` and not on the derive.

### Version

```text
binary: rustc
commit-date: 2022-05-18
host: x86_64-pc-windows-msvc
release: 1.61.0
LLVM version: 14.0.0
```

### Additional Labels

_No response_

daxpedda:
This uses @Jarcho's #8694 implementation to fix `missing_const_for_fn` linting in proc-macros.
I'm not 100% sure what I'm doing here, any feedback is appreciated.

Previously: https://github.com/Jarcho/rust-clippy/pull/1.
Fixes #8854.

changelog: [`missing_const_for_fn`]: No longer lints in proc-macros

### DIFF
--- a/clippy_utils/src/check_proc_macro.rs
+++ b/clippy_utils/src/check_proc_macro.rs
@@ -15,8 +15,8 @@
 use rustc_ast::ast::{IntTy, LitIntType, LitKind, StrStyle, UintTy};
 use rustc_hir::{
     intravisit::FnKind, Block, BlockCheckMode, Body, Closure, Destination, Expr, ExprKind, FieldDef, FnHeader, HirId,
-    Impl, ImplItem, ImplItemKind, IsAuto, Item, ItemKind, LoopSource, MatchSource, QPath, TraitItem, TraitItemKind,
-    UnOp, UnsafeSource, Unsafety, Variant, VariantData, YieldSource,
+    Impl, ImplItem, ImplItemKind, IsAuto, Item, ItemKind, LoopSource, MatchSource, Node, QPath, TraitItem,
+    TraitItemKind, UnOp, UnsafeSource, Unsafety, Variant, VariantData, YieldSource,
 };
 use rustc_lint::{LateContext, LintContext};
 use rustc_middle::ty::TyCtxt;
@@ -251,23 +251,20 @@ fn variant_search_pat(v: &Variant<'_>) -> (Pat, Pat) {
 }
 
 fn fn_kind_pat(tcx: TyCtxt<'_>, kind: &FnKind<'_>, body: &Body<'_>, hir_id: HirId) -> (Pat, Pat) {
-    let (start_pat, end_pat, visibility) = match kind {
-        FnKind::ItemFn(.., header) => (
-            fn_header_search_pat(*header),
-            Pat::Str(""),
-            tcx.visibility(tcx.hir().local_def_id(hir_id)),
-        ),
-        FnKind::Method(.., sig) => (
-            fn_header_search_pat(sig.header),
-            Pat::Str(""),
-            tcx.visibility(tcx.hir().local_def_id(hir_id)),
-        ),
+    let (start_pat, end_pat) = match kind {
+        FnKind::ItemFn(.., header) => (fn_header_search_pat(*header), Pat::Str("")),
+        FnKind::Method(.., sig) => (fn_header_search_pat(sig.header), Pat::Str("")),
         FnKind::Closure => return (Pat::Str(""), expr_search_pat(tcx, &body.value).1),
     };
-    if visibility.is_public() {
-        (Pat::Str("pub"), end_pat)
-    } else {
+    let vis_span = match tcx.hir().get(hir_id) {
+        Node::Item(Item { vis_span, .. }) | Node::ImplItem(ImplItem { vis_span, .. }) => Some(vis_span),
+        Node::TraitItem(_) => None,
+        _ => unreachable!(),
+    };
+    if matches!(vis_span, Some(span) if span.is_empty()) {
         (start_pat, end_pat)
+    } else {
+        (Pat::Str("pub"), end_pat)
     }
 }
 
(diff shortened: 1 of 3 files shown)


clippy_utils/src/check_proc_macro.rs
```diff
@@ -250,6 +250,27 @@ fn variant_search_pat(v: &Variant<'_>) -> (Pat, Pat) {
     }
 }
 
+fn fn_kind_pat(tcx: TyCtxt<'_>, kind: &FnKind<'_>, body: &Body<'_>, hir_id: HirId) -> (Pat, Pat) {
+    let (start_pat, end_pat, visibility) = match kind {
+        FnKind::ItemFn(.., header) => (
+            fn_header_search_pat(*header),
+            Pat::Str(""),
+            tcx.visibility(tcx.hir().local_def_id(hir_id)),
```
daxpedda, line 258: Previously visibility was provided in `FnKind`, this isn't the case anymore, so this is kinda workaround to get it. Not sure if this is right or efficient.

clippy_utils/src/check_proc_macro.rs
```diff
@@ -250,6 +250,27 @@ fn variant_search_pat(v: &Variant<'_>) -> (Pat, Pat) {
...
+        ),
+        FnKind::Closure => return (Pat::Str(""), expr_search_pat(tcx, &body.value).1),
+    };
+    if visibility.is_public() {
+        (Pat::Str("pub"), end_pat)
+    } else {
+        (start_pat, end_pat)
+    }
```
daxpedda, line 271: Not entirely sure what is needed to be done here, do I have to cover more visibility kinds?

clippy_utils/src/check_proc_macro.rs
Jarcho: You can get the HIR node for the item (through `cx.tcx.hir()`) and use the `vis_span` field. If the span isn't empty it will start with `pub`. I don't think `Visibility` gives you an easy way to do this.

Jarcho:
r? @Jarcho
Going to take this as I've already reviewed it.

clippy_utils/src/check_proc_macro.rs
daxpedda: I'm not entirely following, do you mean `tcx.hir().expect_item(tcx.hir().local_def_id(hir_id)).vis_span`?

Jarcho: Use `tcx.hir().get(hir_id)`. It should give `Node::Item`/`ImplItem`/`TraitItem`.

daxpedda: Done! I hope this is correct now.

clippy_utils/src/check_proc_macro.rs
```diff
@@ -250,6 +250,24 @@ fn variant_search_pat(v: &Variant<'_>) -> (Pat, Pat) {
...
+fn fn_kind_pat(tcx: TyCtxt<'_>, kind: &FnKind<'_>, body: &Body<'_>, hir_id: HirId) -> (Pat, Pat) {
+    let (start_pat, end_pat) = match kind {
+        FnKind::ItemFn(.., header) => (fn_header_search_pat(*header), Pat::Str("")),
+        FnKind::Method(.., sig) => (fn_header_search_pat(sig.header), Pat::Str("")),
+        FnKind::Closure => return (Pat::Str(""), expr_search_pat(tcx, &body.value).1),
+    };
+    let vis_span = match tcx.hir().get(hir_id) {
+        Node::Item(Item { vis_span, .. }) | Node::ImplItem(ImplItem { vis_span, .. }) => Some(vis_span),
+        Node::TraitItem(_) => None,
+        _ => unreachable!(),
```
Jarcho, line 262: A few things:
* No need for an option here. The only thing being decided on here is the start pattern which can be assigned directly from the match.
* `TraitItem` should result in `start_pat`, not `pub`, as trait items can't have a visibility modifier.
* The default arm would be better as `Pat::Str("")` just in case something changes on the rustc side.

daxpedda: Done!

> * The default arm would be better as `Pat::Str("")` just in case something changes on the rustc side.

Why `Pat::Str("")` instead of `start_pat`.

Jarcho: It's an unexpected node which could have anything at the start of it. Since there's nothing which will definitely match if it's not a proc-macro `Pat::Str("")` is used to match anything.

daxpedda: I see! Thanks!

Jarcho:
Thank you. @bors r+

PR #9318, opened 2022-08-10
Fix if_let_mutex not checking Mutexes behind refs

### LINKED ISSUE #9193: `if_let_mutex` doesn't trigger if operating on a reference to a mutex
### Summary

Lint doesn't trigger when operating on a reference to a mutex, even though such code is functionally the same and incorrect in the same way (causes a deadlock).

### Lint Name

if_let_mutex

### Reproducer

For the following code:

```rust
#![allow(unused)]

use std::sync::Mutex;

pub fn trigger_1() {
    let mutex = Mutex::new(1);

    if let Ok(_) = mutex.lock() {
        do_thing();
    } else {
        mutex.lock();
    };
}

pub fn trigger_2(mutex: &Mutex<i32>) {
    if let Ok(_) = mutex.lock() {
        do_thing();
    } else {
        mutex.lock();
    };
}

pub fn trigger_3(mutex: Mutex<i32>) {
    if let Ok(_) = mutex.lock() {
        do_thing();
    } else {
        mutex.lock();
    };
}

fn do_thing() {}
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=7fcc496458d492422969d3dc8e0c6f27)

After running `cargo clippy -- -A clippy::all -D clippy::if_let_mutex`

I expected to see the lint trigger for all the methods, but it only triggers for methods `trigger_1` and `trigger_3`

The output I got:
```
> cargo clippy -- -A clippy::all -D clippy::if_let_mutex
    Checking clippy-lints v0.1.0 (/home/dzejkp/Projects/Personal/clippy-lints)
error: calling `Mutex::lock` inside the scope of another `Mutex::lock` causes a deadlock
  --> src/if_let_mutex.rs:8:5
   |
8  | /     if let Ok(_) = mutex.lock() {
9  | |         do_thing();
10 | |     } else {
11 | |         mutex.lock();
12 | |     };
   | |_____^
   |
   = note: requested on the command line with `-D clippy::if-let-mutex`
   = help: move the lock call outside of the `if let ...` expression
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_let_mutex

error: calling `Mutex::lock` inside the scope of another `Mutex::lock` causes a deadlock
  --> src/if_let_mutex.rs:24:5
   |
24 | /     if let Ok(_) = mutex.lock() {
25 | |         do_thing();
26 | |     } else {
27 | |         mutex.lock();
28 | |     };
   | |_____^
   |
   = help: move the lock call outside of the `if let ...` expression
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_let_mutex

error: could not compile `clippy-lints` due to 2 previous errors
```

### Version

```text
binary: rustc
commit-date: 2022-06-27
release: 1.62.0
LLVM version: 14.0.5
```

lukaslueg:
Fixes #9193

We can always peel references because we are looking for a method-call, for which autoderef applies.

---

changelog: [`if_let_mutex`]: detect calls to `Mutex::lock()` if mutex is behind a ref
changelog: [`if_let_mutex`]: Add labels to the two instances of the same Mutex that will deadlock

### DIFF
--- a/clippy_lints/src/if_let_mutex.rs
+++ b/clippy_lints/src/if_let_mutex.rs
@@ -129,7 +129,7 @@ fn is_mutex_lock_call<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) -> Opt
     if_chain! {
         if let ExprKind::MethodCall(path, [self_arg, ..], _) = &expr.kind;
         if path.ident.as_str() == "lock";
-        let ty = cx.typeck_results().expr_ty(self_arg);
+        let ty = cx.typeck_results().expr_ty(self_arg).peel_refs();
         if is_type_diagnostic_item(cx, ty, sym::Mutex);
         then {
             Some(self_arg)
--- a/clippy_lints/src/if_let_mutex.rs
+++ b/clippy_lints/src/if_let_mutex.rs
@@ -64,7 +64,7 @@ impl<'tcx> LateLintPass<'tcx> for IfLetMutex {
                     let diag = |diag: &mut Diagnostic| {
                         diag.span_label(
                             op_mutex.span,
-                            "This Mutex will remain locked for the entire `if let`-block...",
+                            "this Mutex will remain locked for the entire `if let`-block...",
                         );
                         diag.span_label(
                             arm_mutex.span,
--- a/clippy_lints/src/if_let_mutex.rs
+++ b/clippy_lints/src/if_let_mutex.rs
@@ -1,8 +1,9 @@
-use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::higher;
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::SpanlessEq;
 use if_chain::if_chain;
+use rustc_errors::Diagnostic;
 use rustc_hir::intravisit::{self as visit, Visitor};
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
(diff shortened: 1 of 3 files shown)


lukaslueg:
While I was at it, I added some labels which point to the offending usages of the same lock, which I think greatly improves the lint-message

clippy_lints/src/if_let_mutex.rs
```diff
@@ -63,18 +56,28 @@ impl<'tcx> LateLintPass<'tcx> for IfLetMutex {
...
                 arm_visit.visit_expr(if_then);
                 arm_visit.visit_expr(if_else);
 
-                if arm_visit.mutex_lock_called && arm_visit.same_mutex(cx, op_visit.found_mutex.unwrap()) {
-                    span_lint_and_help(
+                if let Some(arm_mutex) = arm_visit.found_mutex_if_same_as(op_mutex) {
+                    let diag = |diag: &mut Diagnostic| {
+                        diag.span_label(
+                            op_mutex.span,
+                            "This Mutex will remain locked for the entire `if let`-block...",
```
xFrednet, line 67: Nice improvements, only one tiny nit and then we can merge this. All diagnostic messages should start with lowercase letters:

```suggestion
                            "this Mutex will remain locked for the entire `if let`-block...",
```

xFrednet (approved):

clippy_lints/src/if_let_mutex.rs
lukaslueg: @xFrednet I'm not exactly familiar with the distinction: Are the `this mutex will remain...` messages `span_label` or `span_note`? The `span_label` currently used is visually less prominent, yet I'd rather go with `span_note`

xFrednet: The guideline, that they should start with a lower letter, applies to all of them. In this case, all that needs to be done is change the first letter and bless the tests again.

Now generally, they are distinguished by what kind of message you want to add. This is my understanding of them:
* `span_label` will display the message and span in the main window. This is intended to show the root problem.
* `span_note` is intended for additional context or showcasing possible connected code
* `span_help`/`span_sugg` for messages with direct help and suggestions

For this and the second message, I believe that either `span_label` or `span_note` is good. I'm fine with this version, but you can try around with `span_note` to see if the message is then even clearer :upside_down_face:

lukaslueg:
```
changelog: [`if_let_mutex`]: Add labels to the two instances of the same Mutex that will deadlock
```

lukaslueg:
@xFrednet LGTM

xFrednet:
LGTM to me as well. Thank you for the work and swift responses!

@bors r+

Issue #8263, opened 2022-01-10
Denied `clippy::non_ascii_literal` cannot be re-allowed for a single test function


TonalidadeHidrica:
### Summary

Suppose that `clippy::non_ascii_literal` is denied globally.  Then I cannot allow it in a single `#[test]` function locally.

### Reproducer

I tried this code:

```rust
#![deny(clippy::non_ascii_literal)]

#[test]
#[allow(clippy::non_ascii_literal)]
pub fn it_works() {
    let _ = "悲しいかな、ここに日本語を書くことはできない。";
}
```

No clippy lint should occur, but a `clippy::non_ascii_literal` is detected.

If I removed `#[test]`, it is allowed as expected.  I suspect that this is due to `test` proc macro producing a code *outside* the testing function.

This issue may be related to, or can be fixed at the same time as, #7739.

### Version

```text
rustc 1.57.0 (f1edd0429 2021-11-29)
binary: rustc
commit-hash: f1edd0429582dd29cccacaf50fd134b05593bd9c
commit-date: 2021-11-29
host: x86_64-apple-darwin
release: 1.57.0
LLVM version: 13.0.0
```

Issue #7739, opened 2021-09-30
non_ascii_literal false positive: non-ascii in comment in test


dtolnay:
Surprisingly the following triggers `non_ascii_literal`.

```rust
#[test]
fn non_ascii_comment() {
    // Ryū crate: https://github.com/dtolnay/ryu
}
```

```console
$ cargo clippy --tests -- -Dclippy::all -Dclippy::pedantic 

error: literal non-ASCII character detected
 --> src/main.rs:2:1
  |
1 |   #[test]
  |   ------- in this procedural macro expansion
2 | / fn non_ascii_comment() {
3 | |     // Ryū crate: https://github.com/dtolnay/ryu
4 | | }
  | |_^
  |
  = note: `-D clippy::non-ascii-literal` implied by `-D clippy::pedantic`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#non_ascii_literal
  = note: this error originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is unexpected because there are no literals in this code and non-ascii in comments should be totally fine.

<br>

### Meta

**Rust version (`rustc -Vv`):**
```
rustc 1.57.0-nightly (11491938f 2021-09-29)
binary: rustc
commit-hash: 11491938f80988c7261a1179cf71a25c379c8783
commit-date: 2021-09-29
host: x86_64-unknown-linux-gnu
release: 1.57.0-nightly
LLVM version: 13.0.0
```

PR #9327, opened 2022-08-13
Fix [`non_ascii_literal`] in tests

Serial-ATA:
changelog: Don't lint [`non_ascii_literal`] when using non-ascii comments in tests
changelog: Don't lint [`non_ascii_literal`] when `allow`ed on tests

closes: #7739
closes: #8263

### DIFF
--- a/clippy_lints/src/unicode.rs
+++ b/clippy_lints/src/unicode.rs
@@ -1,5 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::is_lint_allowed;
+use clippy_utils::macros::span_is_local;
 use clippy_utils::source::snippet;
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
@@ -98,6 +99,10 @@ fn escape<T: Iterator<Item = char>>(s: T) -> String {
 }
 
 fn check_str(cx: &LateContext<'_>, span: Span, id: HirId) {
+    if !span_is_local(span) {
+        return;
+    }
+
     let string = snippet(cx, span, "");
     if string.chars().any(|c| ['\u{200B}', '\u{ad}', '\u{2060}'].contains(&c)) {
         span_lint_and_sugg(
@@ -113,6 +118,7 @@ fn check_str(cx: &LateContext<'_>, span: Span, id: HirId) {
             Applicability::MachineApplicable,
         );
     }
+
     if string.chars().any(|c| c as u32 > 0x7F) {
         span_lint_and_sugg(
             cx,
@@ -128,6 +134,7 @@ fn check_str(cx: &LateContext<'_>, span: Span, id: HirId) {
             Applicability::MachineApplicable,
         );
     }
+
     if is_lint_allowed(cx, NON_ASCII_LITERAL, id) && string.chars().zip(string.nfc()).any(|(a, b)| a != b) {
         span_lint_and_sugg(
             cx,
--- a/tests/ui/unicode.fixed
+++ b/tests/ui/unicode.fixed
@@ -1,4 +1,7 @@
 // run-rustfix
+// compile-flags: --test
+#![allow(dead_code)]
+
 #[warn(clippy::invisible_characters)]
 fn zero() {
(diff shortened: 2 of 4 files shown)


Alexendoo:
Tests storing their contents as a literal sounds rather odd, I'm gonna try and find out what the purpose of that is/if it can be removed

clippy_lints/src/unicode.rs
```diff
@@ -114,6 +114,12 @@ fn check_str(cx: &LateContext<'_>, span: Span, id: HirId) {
         );
     }
     if string.chars().any(|c| c as u32 > 0x7F) {
+        // Tests store their source in a literal, causing non-ascii comments to be linted and
+        // `#[allow]`s to no longer be respected.
```
Alexendoo, line 118: Ah so it's not the whole test that's getting stringified, what's happening is tests get expanded to something like

```rust
#[rustc_test_marker]
pub const test_name: test::TestDescAndFn =
    test::TestDescAndFn {
        desc: test::TestDesc {
            name: test::StaticTestName("test_name"),
            ...
        },
        testfn: test::StaticTestFn(|| test::assert_test_result(test_name())),
    };

#[allow(clippy::non_ascii_literal)]
fn test_name() {
    // 悲
}
```

It's firing on the `"test_name"` literal, because it has the span of the test function and the lint uses `snippet` to check the source, so it sees the string

```
fn test_name() {
    // 悲
}
```

Maybe this lint should be an early pass, that would give access to the `Lit::token` which isn't (un?)escaped without reading back any snippets

Serial-ATA: Thanks for looking into it! It's a late pass since it makes use of `is_lint_allowed`, is there a way to check that in an early pass?

Alexendoo: Ooh I didn't realise that required a late context

The first one I'd say could be removed, I find it surprising that it's applying the result of a separate lint in its suggestion

The second allows an early exit, so removing that may have a minor perf hit but I can't imagine it's noticeable. You could replace the `string.chars().zip(...)` with `unicode_normalization::is_nfc` that has a fast path to make up for it

Serial-ATA: I'm not sure if it can be done this way. `snippet` is just reading from the source, but it looks like when using `Lit::token` it's already escaped, so none of the lints actually fire.

Alexendoo: It should have both, an [`rustc_ast::ast::Lit`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/struct.Lit.html) has  the `token` field of [`rustc_ast::token::Lit`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/token/struct.Lit.html), where `symbol` should be as it appears in source code

That there's two `Lit` structs and two `LitKind`s, and that they contain one another is a bit confusing

If it turns out to be more trouble than it's worth, it'd be reasonable to stick with the current solution, maybe using `clippy_utils::macros::span_is_local` to also catch other macros from elsewhere

Serial-ATA: I don't know if I'm doing something wrong, but `Symbol::as_str` is producing the literal characters "\u{200b}" instead of the ZWSP character.

Alexendoo: Oh that is weird, for the literal `"zws(​) \x7e"` (has a zws between the parenthesis) it's

```rust
Lit {
    token: Lit {
        kind: Str,
        symbol: "zws(\u{200b}) \\x7e",
        suffix: None,
    },
    kind: Str(
        "zws(\u{200b}) ~",
        Cooked,
    ),
    span: t.rs:4:13: 4:26 (#0),
}
```

I wonder why the tokens are escaped like that. But yeah in that case best to stick with the late pass

Alexendoo (approved):

Alexendoo:
Great, thanks!

Thanks also for trying the early lint path even though it didn't pan out

@bors r+

TonalidadeHidrica:
Thanks a lot!

PR #9328, opened 2022-08-13
Remove extra dereference from the borrow_deref_ref lint example

stanislav-tkach:
I don't think such minor change should be mentioned in the changelog.

changelog: none

### DIFF
--- a/clippy_lints/src/borrow_deref_ref.rs
+++ b/clippy_lints/src/borrow_deref_ref.rs
@@ -29,20 +29,15 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
-    /// fn foo(_x: &str) {}
-    ///
     /// let s = &String::new();
     ///
     /// let a: &String = &* s;
-    /// foo(&*s);
     /// ```
     ///
     /// Use instead:
     /// ```rust
-    /// # fn foo(_x: &str) {}
     /// # let s = &String::new();
     /// let a: &String = s;
-    /// foo(&**s);
     /// ```
     #[clippy::version = "1.63.0"]
     pub BORROW_DEREF_REF,


lengyijun:
I prefer `&**` than `&*` .

stanislav-tkach:
@lengyijun This is from the lint description:
```rust
...
fn foo(_x: &str) {}

// Before:
let a: &String = &* s;
foo(&*s); // 1

// Use instead:
let a: &String = s;
foo(&**s); // 2
```
In both cases we have the same type, but in the first (problematic) case the `&*s` construction is used and in the second (fixed) one it is different for some reason. It is confusing: I spent several minutes figuring out why it is different and as we can see, I came to a wrong conclusion. I don't mind using `&**`, but I propose to use the same construction in both places, because it is unrelated to the lint.

clippy_lints/src/borrow_deref_ref.rs
```diff
@@ -42,7 +42,7 @@ declare_clippy_lint! {
     /// # fn foo(_x: &str) {}
     /// # let s = &String::new();
     /// let a: &String = s;
-    /// foo(&**s);
+    /// foo(&*s);
```
giraffate, line 45: It would be better to remove the `foo` function itself.
```suggestion
```

giraffate:
Okay. There are two warnings in the example, first is `borrow_deref_ref` and second is `explicit_auto_deref`. So it would be better to remove second warning.

clippy_lints/src/borrow_deref_ref.rs
stanislav-tkach: Makes sense, done.

giraffate:
@bors r+

Thanks!

PR #9338, opened 2022-08-15
unwrap_used and expect_used: trigger on uses of their _err variants

### LINKED ISSUE #9331: unwrap-used and expect-used don't trigger on unwrap_err/expect_err
### Summary

The `unwrap_used` and `expect_used` lints forbid the use of `unwrap` and `expect`. It seems to me that these lints should also lint `unwrap_err` and `expect_err` as they have essentially the same behavior. I believe people who want to not use unwrap to avoid panic probably don't want a panic to be caused by `unwrap_err` either.

### Reproducer

This does not produce any error related to `unwrap_used` and `expect_used`
```rust
#[deny(clippy::unwrap_used, clippy::expect_used)]
fn main() {
    let mut a = Ok(3);
    a = Err(3);
    a.unwrap_err();
    a = Err(3);
    a.expect_err("expect");
}
```
but i think it should produce errors similar to 
```rust
#[deny(clippy::unwrap_used, clippy::expect_used)]
fn main() {
    let mut a = Ok(3);
    a = Err(3);
    a.unwrap();
    a = Err(3);
    a.expect("expect");
}
```

### Version

```text
rustc 1.65.0-nightly (f22819bcc 2022-08-12)
binary: rustc
commit-hash: f22819bcce4abaff7d1246a56eec493418f9f4ee
commit-date: 2022-08-12
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

@rustbot label -C-bug +C-enhancement

sgued:
changelog: [`unwrap_used`]: lint uses of `unwrap_err`
changelog: [`expect_used`]: lint uses of `expect_err`

fixes #9331

### DIFF
--- a/clippy_lints/src/methods/expect_used.rs
+++ b/clippy_lints/src/methods/expect_used.rs
@@ -7,17 +7,25 @@ use rustc_span::sym;
 
 use super::EXPECT_USED;
 
-/// lint use of `expect()` for `Option`s and `Result`s
-pub(super) fn check(cx: &LateContext<'_>, expr: &hir::Expr<'_>, recv: &hir::Expr<'_>, allow_expect_in_tests: bool) {
+/// lint use of `expect()` or `expect_err` for `Result` and `expect()` for `Option`.
+pub(super) fn check(
+    cx: &LateContext<'_>,
+    expr: &hir::Expr<'_>,
+    recv: &hir::Expr<'_>,
+    is_err: bool,
+    allow_expect_in_tests: bool,
+) {
     let obj_ty = cx.typeck_results().expr_ty(recv).peel_refs();
 
-    let mess = if is_type_diagnostic_item(cx, obj_ty, sym::Option) {
+    let mess = if is_type_diagnostic_item(cx, obj_ty, sym::Option) && !is_err {
         Some((EXPECT_USED, "an Option", "None", ""))
     } else if is_type_diagnostic_item(cx, obj_ty, sym::Result) {
-        Some((EXPECT_USED, "a Result", "Err", "an "))
+        Some((EXPECT_USED, "a Result", if is_err { "Ok" } else { "Err" }, "an "))
     } else {
         None
     };
+
+    let method = if is_err { "expect_err" } else { "expect" };
 
     if allow_expect_in_tests && is_in_test_function(cx.tcx, expr.hir_id) {
         return;
@@ -28,7 +36,7 @@ pub(super) fn check(cx: &LateContext<'_>, expr: &hir::Expr<'_>, recv: &hir::Expr
             cx,
             lint,
             expr.span,
-            &format!("used `expect()` on `{kind}` value"),
+            &format!("used `{method}()` on `{kind}` value"),
             None,
             &format!("if this value is {none_prefix}`{none_value}`, it will panic"),
         );
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -174,7 +174,7 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for `.unwrap()` calls on `Option`s and on `Result`s.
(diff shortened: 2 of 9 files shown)


clippy_lints/src/methods/mod.rs
```diff
@@ -174,7 +174,7 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for `.unwrap()` calls on `Option`s and on `Result`s.
+    /// Checks for `.unwrap()` or `.unwrap_err()` calls on `Option`s and on `Result`s.
```
giraffate, line 177:
```suggestion
    /// Checks for `.unwrap()` or `.unwrap_err()` calls on `Result`s and `.unwrap()` call on `Option`s.
```

clippy_lints/src/methods/mod.rs
```diff
@@ -224,7 +224,7 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for `.expect()` calls on `Option`s and `Result`s.
+    /// Checks for `.expect()` or `.expect_err` calls on `Option`s and `Result`s.
```
giraffate, line 227:
```suggestion
    /// Checks for `.expect()` or `.expect_err` calls on `Result`s and `.expect()` call on `Option`s.
```

clippy_lints/src/methods/expect_used.rs
```diff
@@ -8,17 +8,25 @@ use rustc_span::sym;
 use super::EXPECT_USED;
 
 /// lint use of `expect()` for `Option`s and `Result`s
```
giraffate, line 10:
```suggestion
/// lint use of `expect()` or `expect_err` for `Result` and `expect()` for `Option`.
```

clippy_lints/src/methods/unwrap_used.rs
```diff
@@ -8,17 +8,25 @@ use rustc_span::sym;
 use super::{EXPECT_USED, UNWRAP_USED};
 
 /// lint use of `unwrap()` for `Option`s and `Result`s
```
giraffate, line 10:
```suggestion
/// lint use of `unwrap()` or `unwrap_err` for `Result` and `unwrap()` for `Option`.
```

clippy_lints/src/methods/expect_used.rs
```diff
@@ -8,17 +8,25 @@ use rustc_span::sym;
...
+    let mess = if is_type_diagnostic_item(cx, obj_ty, sym::Option) && !is_err {
         Some((EXPECT_USED, "an Option", "None", ""))
     } else if is_type_diagnostic_item(cx, obj_ty, sym::Result) {
-        Some((EXPECT_USED, "a Result", "Err", "an "))
+        Some((EXPECT_USED, "a Result", if is_err { "Ok" } else { "Err" }, "an "))
     } else {
         None
     };
 
+    let method = if is_err { "expect_err" } else { "expect" };
```
giraffate, line 28: The `expect_err`was added at [1.17](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect_err), so we need to check [msrv](https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv). It would be better to refer to the following:
https://github.com/rust-lang/rust-clippy/blob/a427b12803cf98fb72f1b9c4f9f3a718573fe907/clippy_lints/src/methods/err_expect.rs#L24

giraffate: Oh, sorry, msrv isn't needed here. No need to fix.

giraffate (commented):
Can the tests be added to `tests/ui/unwrap.rs,expect.rs`? The `tests/ui/unwrap_expect_used.rs` file is a test when both `unwrap_used` and `expect_used` are enabled.
https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used
> If [expect_used](https://rust-lang.github.io/rust-clippy/master/index.html#expect_used) is enabled, instead:

giraffate:
@bors r+

Thanks!

PR #9340, opened 2022-08-16
Fix example

alex-semenyuk:
The example didn't show the actual problem [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9d0e0727ca5bbd854767f50da693ca0f)
changelog: none


dswij (approved):
Oh nice find! The example kind of implies that the assignment is the problem. The new one is much clearer.

dswij:
Thanks! @bors r+

PR #9342, opened 2022-08-16
Suggest `Entry::or_default` for `Entry::or_insert(Default::default())`

### LINKED ISSUE #3812: Suggest `Entry::or_default` for `Entry::or_insert_with(Default::default)`
For code that calls `.or_insert_with(Default::default)` on a `std::collections::btree_map::Entry` or `std::collections::hash_map::Entry`, clippy should suggest [`std::collections::btree_map::Entry::or_default`](https://doc.rust-lang.org/std/collections/btree_map/enum.Entry.html#method.or_default) and [`std::collections::hash_map::Entry::or_default`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_default)

relrelb:
Unlike past similar work done in #6228, expand the existing `or_fun_call`
lint to detect `or_insert` calls with a `T::new()` or `T::default()`
argument, much like currently done for `unwrap_or` calls. In that case,
suggest the use of `or_default`, which is more idiomatic.

Note that even with this change, `or_insert_with(T::default)` calls
aren't detected as candidates for `or_default()`, in the same manner
that currently `unwrap_or_else(T::default)` calls aren't detected as
candidates for `unwrap_or_default()`.

Also, as a nearby cleanup, change `KNOW_TYPES` from `static` to `const`,
since as far as I understand it's preferred (should Clippy have a lint
for that?).

Addresses #3812.

changelog: [`or_fun_call`]: Suggest `Entry::or_default` for `Entry::or_insert(Default::default())`

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -825,8 +825,9 @@ declare_clippy_lint! {
 declare_clippy_lint! {
     /// ### What it does
     /// Checks for calls to `.or(foo(..))`, `.unwrap_or(foo(..))`,
-    /// etc., and suggests to use `or_else`, `unwrap_or_else`, etc., or
-    /// `unwrap_or_default` instead.
+    /// `.or_insert(foo(..))` etc., and suggests to use `.or_else(|| foo(..))`,
+    /// `.unwrap_or_else(|| foo(..))`, `.unwrap_or_default()` or `.or_default()`
+    /// etc. instead.
     ///
     /// ### Why is this bad?
     /// The function will always be called and potentially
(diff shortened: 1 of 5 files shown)


dswij (approved):

The change looks great, it's certainly a nice improvement. 

AFAIU, this wouldn't resolve #3812 for `insert_with`? Would you be interested in taking that in another PR?

dswij:
Thanks again for this! @bors r+

xFrednet:
The test failure is a result of some changes to `cargo dev update_lints` on master. To fix it, you will need to rebase. The it should be as simple as running that comment and commiting the changes :)

relrelb:
> The test failure is a result of some changes to `cargo dev update_lints` on master. To fix it, you will need to rebase. The it should be as simple as running that comment and commiting the changes :)

Done.

smoelius:
> Note that even with this change, `or_insert_with(T::default)` calls
> aren't detected as candidates for `or_default()`, in the same manner
> that currently `unwrap_or_else(T::default)` calls aren't detected as
> candidates for `unwrap_or_default()`.

Was this simply unimplemented, or is there an impediment that I am not seeing? Would I be wasting my time by trying to do this?

PR #9348, opened 2022-08-18
Don't lint on match pattern-binding in ´question_mark`

### LINKED ISSUE #9347: false positive for clippy::question_mark
### Summary

Clippy wrongly suggests I can use the question mark operator when returning an error which matches a deep pattern.

The idea is that I'm manually matching an error and then returning, but my match is more specific than what `?` does, so it's a false positive

### Lint Name

question_mark

### Reproducer

I tried this code:

```rust
pub struct Error {
    flag: bool,
}

fn foo() -> Result<(), Error> {
    let res = Ok(());
    
    if let Err(err @ Error { flag: true }) = res {
        return Err(err);
    }
    
    res
}
```

I saw this happen:

```
warning: this block may be rewritten with the `?` operator
  --> src/lib.rs:10:5
   |
10 | /     if let Err(err @ Error { flag: true }) = res {
11 | |         return Err(err);
12 | |     }
   | |_____^ help: replace it with: `res?;`
   |
   = note: `#[warn(clippy::question_mark)]` on by default
```

No warning

### Version

```text
rustc 1.65.0-nightly (29e4a9ee0 2022-08-10)
binary: rustc
commit-hash: 29e4a9ee0253cd39e552a77f51f11f9a5f1c41e6
commit-date: 2022-08-10
host: aarch64-apple-darwin
release: 1.65.0-nightly
LLVM version: 14.0.6
```

### Additional Labels

_No response_

lukaslueg:
Fixes #9347

Technically it is possible to have a blank match-pattern that does nothing, and we fail to lint. But it's easier to be safe than sorry here.

changelog: [`question_mark`]: don't lint `if let`s with subpatterns

### DIFF
--- a/clippy_lints/src/question_mark.rs
+++ b/clippy_lints/src/question_mark.rs
@@ -123,7 +123,7 @@ fn check_if_let_some_or_err_and_early_return<'tcx>(cx: &LateContext<'tcx>, expr:
         if let Some(higher::IfLet { let_pat, let_expr, if_then, if_else }) = higher::IfLet::hir(cx, expr);
         if !is_else_clause(cx.tcx, expr);
         if let PatKind::TupleStruct(ref path1, [field], None) = let_pat.kind;
-        if let PatKind::Binding(annot, bind_id, ident, _) = field.kind;
+        if let PatKind::Binding(annot, bind_id, ident, None) = field.kind;
         let caller_ty = cx.typeck_results().expr_ty(let_expr);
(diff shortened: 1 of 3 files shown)


Alexendoo (approved):

Alexendoo:
Great, thanks!

@bors r+

Alexendoo:
And yeah no need to worry about weird things like `_ @ pat`

PR #9349, opened 2022-08-18
Refactor `FormatArgsExpn`

### LINKED ISSUE #8643: format_in_format_args lint suggests code that results in alternative behaviour
### Summary

I have some code that looks like this

```rust
fn tabulate(devices: Vec<Device>) -> String {
    let mut s = String::from("Vendor ID  Product ID  Name\n");
    s.push_str("---------  ----------  ----------------------------------\n");
    for d in devices {
        s.push_str(&format!(
            "{:<9}  {:<10}  {}",
            format!("0x{:x}", d.vendor_id),
            format!("0x{:x}", d.product_id),
            d.name,
        ));
    }
    s
}
```

Clippy complains about the inner `format!` but replacing them with `format_args!` doesn't align to the columns properly.

### Lint Name

format_in_format_args

### Reproducer

I tried this code:

```rust
fn main() {
    println!(
        "{:>9} {:>10}",
        format!("0x{:x}", 42),
        format!("0x{:x}", 1337)
    );
}
```

I saw this happen:

```text
warning: `format!` in `println!` args
 --> src/main.rs:2:5
  |
2 | /     println!(
3 | |         "{:>9} {:>10}",
4 | |         format!("0x{:x}", 42),
5 | |         format!("0x{:x}", 1337)
6 | |     );
  | |_____^
  |
  = note: `#[warn(clippy::format_in_format_args)]` on by default
  = help: combine the `format!(..)` arguments with the outer `println!(..)` call
  = help: or consider changing `format!` to `format_args!`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#format_in_format_args
```

No lint emitted, or a better suggestion

### Version

```text
rustc 1.62.0-nightly (306ba8357 2022-04-05)
binary: rustc
commit-hash: 306ba8357fb36212b7d30efb9eb9e41659ac1445
commit-date: 2022-04-05
host: aarch64-apple-darwin
release: 1.62.0-nightly
LLVM version: 14.0.0
```

### Additional Labels

Alexendoo:
It now for each format argument `{..}` has:
- The `Expr` it points to, and how it does so (named/named inline/numbered/implicit)
- The parsed `FormatSpec` (format trait/fill/align/etc., the precision/width and any value they point to)
- Many spans

The caller no longer needs to pair up arguments to their value, or separately interpret the `specs` `Expr`s when it isn't `None`

The gist is that it combines the result of [`rustc_parse_format::Parser`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_parse_format/struct.Parser.html) with the macro expansion itself

This unfortunately makes the code a bit longer, however we need to use both as neither have all the information we're after. `rustc_parse_format` doesn't have the information to resolve named arguments to their values. The macro expansion doesn't contain whether the positions are implicit/numbered/named, or the spans for format arguments

Wanted by #9233 and #8518 to be able to port the changes from #9040

Also fixes #8643, previously the format args seem to have been paired up with the wrong values somehow

changelog: [`format_in_format_args`]: Fix false positive due to misattributed arguments

r? @flip1995
cc @nyurik

### DIFF
--- a/clippy_utils/src/macros.rs
+++ b/clippy_utils/src/macros.rs
@@ -1,16 +1,21 @@
 #![allow(clippy::similar_names)] // `expr` and `expn`
 
+use crate::is_path_diagnostic_item;
+use crate::source::snippet_opt;
 use crate::visitors::expr_visitor_no_bodies;
 
 use arrayvec::ArrayVec;
-use if_chain::if_chain;
+use itertools::{izip, Either, Itertools};
 use rustc_ast::ast::LitKind;
 use rustc_hir::intravisit::Visitor;
 use rustc_hir::{self as hir, Expr, ExprKind, HirId, Node, QPath};
+use rustc_lexer::unescape::unescape_literal;
+use rustc_lexer::{tokenize, unescape, LiteralKind, TokenKind};
 use rustc_lint::LateContext;
+use rustc_parse_format::{self as rpf, Alignment};
 use rustc_span::def_id::DefId;
 use rustc_span::hygiene::{self, MacroKind, SyntaxContext};
-use rustc_span::{sym, ExpnData, ExpnId, ExpnKind, Span, Symbol};
+use rustc_span::{sym, BytePos, ExpnData, ExpnId, ExpnKind, Pos, Span, SpanData, Symbol};
 use std::ops::ControlFlow;
 
 const FORMAT_MACRO_DIAG_ITEMS: &[Symbol] = &[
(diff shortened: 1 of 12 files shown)


smoelius:
Maybe this makes #8857 unnecessary?

flip1995 (changes requested):
This all is really specific to how the `format_arg!` macro expands. But that also was the previous code, so I don't see this as a blocker. 

However, I noticed that the pattern is to construct those things from `rpf` information. Can you give an estimation how likely it is to add this information to `rpf` and give Clippy access to it? This feels like we're doing work that is/can already (be) done in `rustc`. 

Those are no blockers. Just a few comments how the code could maybe be improved.

clippy_utils/src/source.rs
```diff
@@ -392,6 +394,55 @@ pub fn trim_span(sm: &SourceMap, span: Span) -> Span {
     .span()
 }
 
+#[derive(Debug)]
+pub struct StrLiteral {
+    pub span: Span,
+    pub snippet: String,
+    /// If the string is raw `r"..."`/`r#""#`, how many `#`s does it have
```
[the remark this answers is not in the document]
Alexendoo: Ah, yes!

clippy_utils/src/source.rs
```diff
@@ -392,6 +394,55 @@ pub fn trim_span(sm: &SourceMap, span: Span) -> Span {
...
+    pub style: Option<usize>,
+    pub unescaped: String,
+}
+
+impl StrLiteral {
+    pub fn new(cx: &impl LintContext, span: Span) -> Option<Self> {
```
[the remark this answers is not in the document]
Alexendoo: The actual format string doesn't exist in the HIR. e.g.

```rust
format_args!(r"a {} b \", 1);
```
expands to
```rust
::core::fmt::Arguments::new_v1(&["a ", " b \\"],
    &[::core::fmt::ArgumentV1::new_display(&1)]);
```

`pieces` is the first arg `&["a ", " b \\"]`, the raw-ness is lost to expansion but it does have the span of `r"a {} b \"`. I'll add some comments for this

Yeah I'll move it back to `macros.rs` as well, not sure why I put it in `source.rs`

clippy_utils/src/macros.rs
```diff
@@ -332,121 +337,496 @@ fn is_assert_arg(cx: &LateContext<'_>, expr: &Expr<'_>, assert_expn: ExpnId) ->
...
-                        formatters.push((val_idx, Symbol::intern(fmt_trait)));
-                    }
-                }
-                if let ExprKind::Struct(QPath::Resolved(_, path), ..) = e.kind {
-                    if path.segments.last().unwrap().ident.name == sym::Argument {
-                        specs.push(e);
-                    }
+impl FormatString {
+    pub fn new(cx: &LateContext<'_>, pieces: &Expr<'_>) -> Option<Self> {
+        // The format string doesn't exist in the HIR, so we reassemble it from source code
```
flip1995, line 357: FormatString is a public struct, so this might make more sense as a doc comment on the struct.

clippy_utils/src/macros.rs
```diff
@@ -332,121 +335,419 @@ fn is_assert_arg(cx: &LateContext<'_>, expr: &Expr<'_>, assert_expn: ExpnId) ->
...
+        } else {
+            None
+        }
+    }
+
+    if let ExprKind::AddrOf(.., array) = fmt_arg.kind
+        && let ExprKind::Array(specs) = array.kind
+    {
+        Some(specs.iter().map(|spec| {
+            let mut positions = ParamPositions::default();
```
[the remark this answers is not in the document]
Alexendoo: I'm not that happy about the name either, but I added some docs on the struct

clippy_utils/src/macros.rs
```diff
@@ -332,121 +337,496 @@ fn is_assert_arg(cx: &LateContext<'_>, expr: &Expr<'_>, assert_expn: ExpnId) ->
...
+            format_string_span,
+        }
+    }
+}
+
+/// The positions of a format argument's value, precision and width
+///
+/// A position is an index into the second argument of `Arguments::new_v1[_formatted]`
+#[derive(Debug, Default, Copy, Clone)]
+struct ParamPositions {
```
flip1995, line 478: Reading the doc comment: Shouldn't this then just be 
```suggestion
/// The positions of a format argument's value, precision and width
///
/// A position is an index into the second argument of `Arguments::new_v1[_formatted]`
#[derive(Debug, Default, Copy, Clone)]
struct ParamPosition {
```

So singular? And then also the variable names in the function below.

flip1995 (commented):
Only 2 minor comments, then we can move on with the other PR :rocket:

Alexendoo:
> However, I noticed that the pattern is to construct those things from rpf information. Can you give an estimation how likely it is to add this information to rpf and give Clippy access to it? This feels like we're doing work that is/can already (be) done in rustc.

I don't think `rustc_parse_format` could gain the full capabilities since it's intended to compile on stable. Resolving the names would need access to the bits (`TokenTree`?) that come after the format string

There is talk of [making `format_args` its own node](https://rust-lang.zulipchat.com/#narrow/stream/233931-t-compiler.2Fmajor-changes/topic/Make.20format_args!.28.29.20its.20own.20AST.20node.20.28ast.E2.80.A6.20compiler-team.23541) though, so maybe in the future we can delete all of this 😄

clippy_utils/src/macros.rs
Alexendoo: Also whoops `new` shouldn't be public

clippy_utils/src/macros.rs
```diff
@@ -332,121 +337,495 @@ fn is_assert_arg(cx: &LateContext<'_>, expr: &Expr<'_>, assert_expn: ExpnId) ->
...
+impl FormatString {
+    fn new(cx: &LateContext<'_>, pieces: &Expr<'_>) -> Option<Self> {
+        // format_args!(r"a {} b \", 1);
+        //
+        // expands to
+        //
+        // ::core::fmt::Arguments::new_v1(&["a ", " b \\"],
+        //      &[::core::fmt::ArgumentV1::new_display(&1)]);
+        //
+        // where `pieces` is the expression `&["a ", " b \\"]`. It has the span of `r"a {} b \"`
```
flip1995, line 365: Oh I meant the whole comment as a doc comment, sorry.

flip1995 (approved):
Thanks, just another detail. After that r=me.

clippy_utils/src/macros.rs
Alexendoo: Ah, I'd say the rest of it is implementation details the caller shouldn't need to worry about

flip1995: Ok then :+1:

flip1995:
@bors r+

Thanks!

Issue #9307, opened 2022-08-08
Allowing `clippy::arithmetic` in const contexts


DrSloth:
### Description

The new `clippy::arithmetic` lint always lints against usage of arithmetic operators. Would it be feasible to not lint in const contexts? The reasoning for the lint is the overflow/panic behavior, there are a lot of scenarios which could be statically checked.

For instance:
```
// This will never panic
let x = 60i32 * 100;
// This will never panic and a panic in a const would appear at compile time
const Y = 100u8 + 17;
```

Would it be feasible to disallow arithmetic only in scenarios which may actually overflow/panic?

### Version

```text
binary: rustc
commit-hash: d394408fb38c4de61f765a3ed5189d2731a1da91
commit-date: 2022-08-07
host: x86_64-unknown-linux-gnu
release: 1.65.0-nightly
LLVM version: 14.0.6
```

### Additional Labels

@rustbot label +C-enhancement

c410-f3r:
Yes, it is in my TODO list. I will probably provide a PR in the following days.

PR #9365, opened 2022-08-23
[Arithmetic] Consider literals

c410-f3r:
Fixes https://github.com/rust-lang/rust-clippy/issues/9307 and makes the `arithmetic` lint behave like `integer_arithmetic`.

It is worth noting that literal integers of a binary operation (`1 + 1`, `i32::MAX + 1`), **regardless if they are in a constant environment**, won't trigger the lint. Assign operations also have similar reasoning. 

changelog: Consider literals in the arithmetic lint


clippy_lints/src/operators/arithmetic.rs
```diff
@@ -50,36 +52,82 @@ impl Arithmetic {
...
     }
+
+    /// Checks assign operators (+=, -=, *=, /=) already handled by the CTFE.
+    fn has_valid_assign_op(op: &Spanned<hir::BinOpKind>, rhs: &hir::Expr<'_>) -> bool {
+        match op.node {
+            hir::BinOpKind::Div | hir::BinOpKind::Rem => match rhs.kind {
+                hir::ExprKind::Lit(_) => true,
+                hir::ExprKind::Unary(hir::UnOp::Neg, expr) => {
+                    let hir::ExprKind::Lit(lit) = &expr.kind else {
+                        return true;
```
Alexendoo, line 63: Could be `i /= -(something that ends up as 1);`

```suggestion
                        return false;
```

clippy_lints/src/operators/arithmetic.rs
```diff
@@ -50,36 +52,82 @@ impl Arithmetic {
         span_lint(cx, ARITHMETIC, expr.span, "arithmetic detected");
         self.expr_span = Some(expr.span);
     }
+
+    /// Checks assign operators (+=, -=, *=, /=) already handled by the CTFE.
+    fn has_valid_assign_op(op: &Spanned<hir::BinOpKind>, rhs: &hir::Expr<'_>) -> bool {
+        match op.node {
```
Alexendoo, line 58: This could do with a check that the types involved are integral, as it could cause a panic for custom types. The comment should also be updated to say it's checking for cases that can't cause overflow rather than that are covered by CTFE

clippy_lints/src/operators/arithmetic.rs
```diff
@@ -50,36 +52,82 @@ impl Arithmetic {
...
+        };
+        if self.is_allowed_ty(cx, lhs) || self.is_allowed_ty(cx, rhs) {
+            return;
+        }
+        let lhs_refs = cx.typeck_results().expr_ty(lhs).peel_refs();
+        let rhs_refs = cx.typeck_results().expr_ty(rhs).peel_refs();
+        if lhs_refs.is_floating_point() && rhs_refs.is_floating_point() {
+            self.issue_lint(cx, expr);
+            return;
+        }
```
Alexendoo, line 113: Should be able to remove this check if both `has_valid_assign_op` and `has_valid_op` are checking the type is integeral

Alexendoo (commented):
👍

Just a couple small bits to suggest

c410-f3r:
Addressed all comments with a new strategy and additional tests

clippy_lints/src/operators/arithmetic.rs
```diff
@@ -46,40 +71,65 @@ impl Arithmetic {
...
+            hir::BinOpKind::Add
+            | hir::BinOpKind::Sub
+            | hir::BinOpKind::Mul
+            | hir::BinOpKind::Div
+            | hir::BinOpKind::Rem
+            | hir::BinOpKind::Shl
+            | hir::BinOpKind::Shr
+        ) = op.node else {
+            return;
+        };
```
Alexendoo, line 109: Can use matches here to avoid pulling in an extra `feature` just for this

Alexendoo (commented):
Nice, thanks!

I think it'd be worth updating the lint documentation to mention that the lint doesn't fire for cases where overflow can't happen

clippy_lints/src/operators/arithmetic.rs
c410-f3r: Done

c410-f3r:
> Nice, thanks!
> 
> I think it'd be worth updating the lint documentation to mention that the lint doesn't fire for cases where overflow can't happen

Done

clippy_lints/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![feature(box_patterns)]
 #![feature(control_flow_enum)]
 #![feature(drain_filter)]
+#![feature(if_let_guard)]
```
Alexendoo, line 6:
```suggestion
```

clippy_lints/src/operators/mod.rs
```diff
@@ -61,25 +61,29 @@ declare_clippy_lint! {
...
+    /// Known built-in types like `Wrapping` or `Saturing`, operations in constant
+    /// environments, allowed types and non-constant operations that won't overflow are ignored.
     ///
     /// ### Why is this bad?
-    /// Integer overflow will trigger a panic in debug builds or will wrap in
-    /// release mode. Division by zero will cause a panic in either mode. In some applications one
-    /// wants explicitly checked, wrapping or saturating arithmetic.
+    /// For integers, an overflow will trigger panics in debug builds or wrap the result in
+    /// release mode; division by zero will cause a panic in either mode. As a result, it is
+    /// desirable to explicit call checked, wrapping or saturating arithmetic methods.
```
Alexendoo, line 76:
```suggestion
    /// desirable to explicitly call checked, wrapping or saturating arithmetic methods.
```

clippy_lints/src/operators/arithmetic.rs
```diff
@@ -5,13 +5,21 @@
...
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
 use rustc_session::impl_lint_pass;
-use rustc_span::source_map::Span;
+use rustc_span::source_map::{Span, Spanned};
 
-const HARD_CODED_ALLOWED: &[&str] = &["std::num::Saturating", "std::string::String", "std::num::Wrapping"];
+const HARD_CODED_ALLOWED: &[&str] = &[
+    "f32",
+    "f64",
```
Alexendoo, line 18: This is surprising to me, I would have expected this lint to be a superset of `float_arithmetic` and `integer_arithmetic`. It should be called out by name that floats aren't checked at least in the lint description

c410-f3r: An oversight. Floats don't "overflow" as regular integers, therefore, they aren't fit for this lint.

`float_arithmetic` even states that it is intended for kernels or embedded systems -> https://rust-lang.github.io/rust-clippy/master/index.html#float_arithmetic

clippy_lints/src/lib.rs
Alexendoo: Oh I missed that, thought it was unused

clippy_lints/src/operators/arithmetic.rs
Alexendoo: Yeah it makes sense, just didn't fit my initial expectation from the lint name

c410-f3r: Perhaps something like `side_effects_arithmetic`?

Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

Renaming the lint may be a good idea yeah. Could ask on zulip for some input on a potential name, my vote would probably be the slight variation `arithmetic_side_effects`

PR #9366, opened 2022-08-23
Rename `manual_empty_string_creation` and move to pedantic

Alexendoo:
Renames it to `manual_string_new` and moves it to the pedantic category

Pedantic because it's a fairly minor style change but could be very noisy

changelog: *doesn't need its own entry, but remember to s/manual_empty_string_creation/manual_string_new/ the changelog entry for #9295*

r? @xFrednet to get it in before the upcoming sync as this isn't a `cargo dev rename_lint` style rename


xFrednet (approved):

xFrednet:
LGTM, thank you for the move :)

@bors r+

PR #9368, opened 2022-08-23
Improvement for  `equatable_if_let`

### LINKED ISSUE #9221: lint matchable `if let` for pattern
### What it does

Checks for `if let` for `pattern`s that can be re-written with `matches!`

This can be an improvement to [`equatable_if_let`](https://rust-lang.github.io/rust-clippy/master/#equatable_if_let) lint.

For background see https://github.com/rust-lang/rust-clippy/pull/9102#discussion_r924363140

### Lint Name

_No response_

### Category

_No response_

### Advantage

Similar to `equatable_if_let`:

1. It reads better and has less cognitive load because equality won’t cause binding.
2. It avoids a [Yoda condition](https://en.wikipedia.org/wiki/Yoda_conditions).
3. Equality is a simple bool expression and can be merged with && and || and reuse if blocks

### Drawbacks

_No response_

### Example

```rust
#![allow(dead_code)]
enum Foo {
    Bar,
    Baz,
}

fn test(foo: Foo) {
    if let Foo::Baz = foo {
        println!("Foo::Baz");
    }
}
```

can be written as

```rust
#![allow(dead_code)]
enum Foo {
    Bar,
    Baz,
}

fn test(foo: Foo) {
    if matches!(foo, Foo::Baz) {
        println!("Foo::Baz");
    }
}
```

nahuakang:
fixes #9221 

This PR makes sure that enums or structs not implementing `PartialEq` trait but still using the `if let` patterns can be linted to be rewritten with `matches!`.

- \[ ] Executed `cargo dev update_lints`
- \[ ] Added lint documentation
- \[x] Run `cargo dev fmt`

---

changelog: Improve [`equatable_if_let`] with additional `matches!` suggestions.

### DIFF
--- a/clippy_lints/src/equatable_if_let.rs
+++ b/clippy_lints/src/equatable_if_let.rs
@@ -1,7 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet_with_context;
 use clippy_utils::ty::implements_trait;
-use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, Pat, PatKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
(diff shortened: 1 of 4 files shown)


dswij (approved):
Thanks for the PR! This looks good to me. 

I'll wait for @flip1995's review if there is any concern with merging this. Particularly, I'm not entirely sure if there's any scenario where we don't want to suggest `matches!`.

dswij:
Just one comment, though, @nahuakang can you help to write the `changelog` field in the PR description

nahuakang:
@flip1995 Btw when you have time please take a look here :)

dswij:
Actually @nahuakang, I found a discussion for this very issue https://github.com/rust-lang/rust-clippy/pull/7777 - and it's unresolved yet.

Apologies for just finding this out after you made this PR 😟

nahuakang:
@dswij No problem :) Thanks for finding it out. So what's the status for either PR? Should I close mine in favor of the other or adjust something?

dswij:
> @dswij No problem :) Thanks for finding it out. So what's the status for either PR? Should I close mine in favor of the other or adjust something?

Seems like the ongoing discussion is not for the `matches!` case although the consensus there is somewhat on making `matches!` as a new lint rather than an improvement of `equatable_if_let`.

I will try to bring this up in Zulip.

flip1995:
The discussion in the other PR is not really affected by this change, so I'll merge it like this.

Wow, I'm really behind on reviews (this is already from 2 months ago). Sorry about that.

@bors r+

PR #9370, opened 2022-08-24
Replace `contains_ty(..)` with `Ty::contains(..)`

mikerite:
This removes some code we don't need and the method syntax is
also more readable IMO.

changelog: none


Jarcho:
Thank you. @bors r+

Issue #6560, opened 2021-01-07
Lint for large error types


Nemo157:
### What it does

Warn when an error type (probably by detecting that it has `impl std::error::Error` and is publicly exported) is over some arbitrary threshold, suggesting boxing the inner data to reduce the size.

### Categories (optional)

- Kind: `clippy::perf`

Large error types result in a lot of redundant memory copying on the success path when the `Ok` value is small.

### Drawbacks

None.

### Example

```rust
struct Error {
    data: [u8; 512],
}

impl std::error::Error for Error {}
```

Could be written as:

```rust

struct Error {
    data: Box<[u8; 512]>,
}

impl std::error::Error for Error {}
```

### Other

There are two existing related issues: https://github.com/rust-lang/rust-clippy/issues/4652 and https://github.com/rust-lang/rust-clippy/issues/3884. I think this is different because in those cases it detects when you use the error internally with a small success variant; whereas even if internally all usage is with large success variants, a user of the library may be mapping the success variant to something small, ending up hitting one of those lints in their usage and having to box the whole error; while it would be easier for all users if the size was reduced at the source.

GnomedDev:
Should this be closed now that we have https://github.com/rust-lang/rust-clippy/pull/9373?

Issue #6409, opened 2020-12-01
Lint idea: suggest std Error trait for custom errors


woodruffw:
### What it does

Libraries regularly implement their own error types, frequently as `enum` wrappers around dependency-specific errors. 

Other libraries like [anyhow](https://github.com/dtolnay/anyhow) exist to make error handling in application code simpler, at least for error types that implement `std::error::Error`.

However, when wrapping/writing custom error types for a library, it's easy to forget to implement `std::error::Error`. It would be nice to have a `clippy` lint that detects custom error types and suggests that the user add an `impl std::error::Error for MyError {}` implementation.

### Categories (optional)

Potentially `clippy::style`.

*What is the advantage of the recommended code over the original code*

The recommended code transitively simplifies error handling in other codebases by ensuring that the standard `Error` trait is implemented. This enables more direct use of libraries like `anyhow`.

### Drawbacks

This lint should only be run on crates that are specified as libraries, not as application code (i.e., producing binaries). 

It might also be somewhat difficult to determine which types are custom error types. Some ideas:

* Look for `Result<T, X>`, where `X` is a crate-local type
* Look for types within `crate::*::error` that match a naming pattern (like `*Error*`)
* Look for `enum` types whose variants' associated datas contain one or more implementations of `std::error::Error`

### Example

```rust
pub enum MyError {
  Io(std::io::Error),
  Other(SomeOtherError),
}
```

Could be written as:

```rust
pub enum MyError {
  Io(std::io::Error),
  Other(SomeOtherError),
}

impl std::error::Error for MyError {
  // trait impl...
}
```

giraffate:
A little difference but similar to https://github.com/rust-lang/rust-clippy/issues/1291.

Issue #4652, opened 2019-10-11
Result<T, E> where size_of::<T>() < 128 and size_of::<E> >= 128


llogiq:
Those will cause LLVM to issue `memcpy` calls to copy them around. Boxing the error type will push them below LLVM's size threshold, thus possibly improving performance, unless the Err case is more likely than the Ok case.

andjo403:
feels like the [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/#large_enum_variant) check but with a limit of 128 by default instead of 200.

llogiq:
Sort of, but specialized to `Result`, because in most cases the `Err` case isn't quite as hot as the `Ok` case, so the wins on the latter can outweigh the losses from boxing on the former.

thomcc:
Also large_enum_variant doesn't fire on Result<T, E>, since (I think?) it's looking at the definition of the enum.

Issue #3884, opened 2019-03-14
New lint: large Err(E) variant, as compared with Ok(T), for Result<T, E>


dekellum:
I've found the `large_enum_variant` lint from clippy quite educational and was hoping it would be able to lint on what I'll call _unbalanced_ `Result<T, E>` specified types, in particular when the E (Error) type is really large but the T type is quite small.  From my reading, this is often a performance concern, e.g. wasting lots of stack space for an infrequently used error path.  On the other hand if the T (Ok, happy path) is larger, then the cost of E should be negligible. Make sense?

After not being able to trigger clippy with `Result<(), BigError>` (where `BigError` is 3,000 bytes) I tried using my own `MyResult`, along the lines of:

``` rust
pub enum MyResult<T, E> {
    Ok(T),
    Err(E),
}

/// MyResult concrete type with zero-bytes `Ok` and 3K bytes `Err` variant,
/// for a fixed stack size of 3001 bytes (asserted).
type Unbalanced = MyResult<(), BigError>;

fn f() -> Unbalanced {
    MyResult::Err(BigError::Large(error::Biggie { blob: [0u8; 3000] }))
}
```

But no clippy lint at present on either the fully specified `type Unbalanced` alias or the `f()` function definition.  

Have/would you consider extending this lint to all generic enums, where they are fully specified, or perhaps more practically, adding a new lint specific to `core::result::Result`? The suggestion to use `Box<E>` would be applicable in the later case.

I don't claim to well understand the clippy code base, but this was the only hint I could find (an @oli-obk comment) that generic enum's were considered (and are likely ignored) for the original lint:

https://github.com/rust-lang/rust-clippy/blob/bb41b16423d5a99f02f62a6a1340499dd4e05f70/clippy_lints/src/large_enum_variant.rs#L71

``` txt
% cargo clippy -V
clippy 0.0.212 (016d92d 2019-03-10)
```

oli-obk:
My comment refers generic enums that are still too generic to compute the size. Your case completely monomorphizes the generic enum to something where we can compute the size, and I agree with you that we should be doing this.

I think we should start small and only look at `Result` types to begin with, since most of the time one does not have control over arbitrary enums from other crates, so there might be no way to change it.

[...]

PR #9373, opened 2022-08-24
Initial implementation `result_large_err`

lukaslueg:
This is a shot at #6560, #4652, and #3884. The lint checks for `Result` being returned from functions/methods where the `Err` variant is larger than a configurable threshold (the default of which is 128 bytes). There has been some discussion around this, which I'll try to quickly summarize:

* A large `Err`-variant may force an equally large `Result` if `Err` is actually bigger than `Ok`.
* There is a cost involved in large `Result`, as LLVM may choose to `memcpy` them around above a certain size.
* We usually expect the `Err` variant to be seldomly used, but pay the cost every time.
* `Result` returned from library code has a high chance of bubbling up the call stack, getting stuffed into `MyLibError { IoError(std::io::Error), ParseError(parselib::Error), ...}`, exacerbating the problem.

This PR deliberately does not take into account comparing the `Ok` to the `Err` variant (e.g. a ratio, or one being larger than the other). Rather we choose an absolute threshold for `Err`'s size, above which we warn. The reason for this is that `Err`s probably get `map_err`'ed further up the call stack, and we can't draw conclusions from the ratio at the point where the `Result` is returned. A relative threshold would also be less predictable, while not accounting for the cost of LLVM being forced to generate less efficient code if the `Err`-variant is _large_ in absolute terms.

We lint private functions as well as public functions, as the perf-cost applies to in-crate code as well.

In order to account for type-parameters, I conjured up `fn approx_ty_size`. The function relies on `LateContext::layout_of` to compute the actual size, and in case of failure (e.g. due to generics) tries to come up with an "at least size". In the latter case, the size of obviously wrong, but the inspected size certainly can't be smaller than that. Please give the approach a heavy dose of review, as I'm not actually familiar with the type-system at all (read: I have no idea what I'm doing).

The approach does, however flimsy it is, allow us to successfully lint situations like

```rust
pub union UnionError<T: Copy> {
    _maybe: T,
    _or_perhaps_even: (T, [u8; 512]),
}

// We know `UnionError<T>` will be at least 512 bytes, no matter what `T` is
pub fn param_large_union<T: Copy>() -> Result<(), UnionError<T>> {
    Ok(())
}
```

I've given some refactoring to `functions/result_unit_err.rs` to re-use some bits. This is also the groundwork for #6409

The default threshold is 128 because of https://github.com/rust-lang/rust-clippy/issues/4652#issue-505670554

`lintcheck` does not trigger this lint for a threshold of 128. It does warn for 64, though.

The suggestion currently is the following, which is just a placeholder for discussion to be had. I did have the computed size in a `span_label`. However, that might cause both ui-tests here and lints elsewhere to become flaky wrt to their output (as the size is platform dependent).

```
error: the `Err`-variant returned via this `Result` is very large
  --> $DIR/result_large_err.rs:36:34
   |
LL | pub fn param_large_error<R>() -> Result<(), (u128, R, FullyDefinedLargeError)> {
   |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ The `Err` variant is unusually large, at least 128 bytes
```

changelog: Add [`result_large_err`] lint


lukaslueg:
I'll deal with CI after review. As far as suggestions go, maybe:

`error: the Err-variant returned from this public function is very large`
`^^^^^^^^^^^^^ The Err-variant is at least [actual size] bytes` (is this a problem for automation?)
`help: Try reducing the size of UnionError<T>, e.g. by boxing elements or replacing it with Box<UnionError<T>>`

clippy_lints/src/functions/result.rs
```diff
@@ -0,0 +1,152 @@
...
+use super::{RESULT_LARGE_ERR, RESULT_UNIT_ERR};
+
+/// comes up with an "at least" guesstimate for the type's size, not taking into
+/// account the layout of type parameters.
+/// For example, a `((u32, u32, T), T, [u8; 4])` will be 12, which is only
+/// correct if `T` is a zst *and* doesn't have alignment...
+fn approx_ty_size<'tcx>(cx: &LateContext<'tcx>, ty: ty::Ty<'tcx>) -> u64 {
+    use rustc_middle::ty::layout::LayoutOf;
+    match (cx.layout_of(ty).map(|layout| layout.size.bytes()), ty.kind()) {
+        (Ok(size), _) => size,
```
Alexendoo, line 22: Could include arrays with something like

```suggestion
        (Ok(size), _) => size,
        (Err(_), ty::Array(t, n)) => n.try_eval_usize(cx.tcx, cx.param_env).map_or(0, |len| len * approx_ty_size(cx, t))
```

clippy_lints/src/functions/result.rs
```diff
@@ -0,0 +1,152 @@
...
+fn result_err_ty<'tcx>(
+    cx: &LateContext<'tcx>,
+    decl: &hir::FnDecl<'tcx>,
+    item_span: Span,
+) -> Option<(&'tcx hir::Ty<'tcx>, ty::Ty<'tcx>)> {
+    if !in_external_macro(cx.sess(), item_span)
+    && let hir::FnRetTy::Return(hir_ty) = decl.output
+    && let ty = hir_ty_to_ty(cx.tcx, hir_ty)
+    && is_type_diagnostic_item(cx, ty, sym::Result)
+    && let ty::Adt(_, substs) = ty.kind() {
```
Alexendoo, line 72: Some formatting as rustfmt doesn't do let chains yet:

```suggestion
        && let hir::FnRetTy::Return(hir_ty) = decl.output
        && let ty = hir_ty_to_ty(cx.tcx, hir_ty)
        && is_type_diagnostic_item(cx, ty, sym::Result)
        && let ty::Adt(_, substs) = ty.kind()
    {
```

clippy_lints/src/functions/result.rs
```diff
@@ -0,0 +1,152 @@
...
+        let err_ty = substs.type_at(1);
+        Some((hir_ty, err_ty))
+    } else {
+        None
+    }
+}
+
+pub(super) fn check_item<'tcx>(cx: &LateContext<'tcx>, item: &hir::Item<'tcx>, large_err_threshold: u64) {
+    if let hir::ItemKind::Fn(ref sig, _generics, _) = item.kind
+        && let Some((hir_ty, err_ty)) = result_err_ty(cx, sig.decl, item.span) {
```
Alexendoo, line 82:
```suggestion
        && let Some((hir_ty, err_ty)) = result_err_ty(cx, sig.decl, item.span)
    {
```

clippy_lints/src/functions/result.rs
```diff
@@ -0,0 +1,152 @@
...
+    // Don't lint if method is a trait's implementation, we can't do anything about those
+    if let hir::ImplItemKind::Fn(ref sig, _) = item.kind
+        && let Some((hir_ty, err_ty)) = result_err_ty(cx, sig.decl, item.span)
+          && trait_ref_of_method(cx, item.def_id).is_none() {
+            if cx.access_levels.is_exported(item.def_id) {
+                let fn_header_span = item.span.with_hi(sig.decl.output.span().hi());
+                check_result_unit_err(cx, err_ty, fn_header_span);
+            }
+            check_result_large_err(cx, err_ty, hir_ty.span, large_err_threshold);
+        }
```
Alexendoo, line 101:
```suggestion
        && let Some((hir_ty, err_ty)) = result_err_ty(cx, sig.decl, item.span)
        && trait_ref_of_method(cx, item.def_id).is_none()
    {
        if cx.access_levels.is_exported(item.def_id) {
            let fn_header_span = item.span.with_hi(sig.decl.output.span().hi());
            check_result_unit_err(cx, err_ty, fn_header_span);
        }
        check_result_large_err(cx, err_ty, hir_ty.span, large_err_threshold);
    }
```

Alexendoo (commented):
Looks good so far!

Might be worth grepping for `layout_of` and checking if any other usages of it could benefit from this `approx_ty_size` fn

clippy_lints/src/functions/result.rs
lukaslueg: 👍 I didn't think for `[(T, u128); 10]` et al

lukaslueg:
All addressed. Will squash if so requested.

I've moved `approx_ty_size` to `clippy_utils::ty`, where it might be useful for other things like [`large_enum_variant`](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant). I may grep and address that in a later PR. Added the `array`-case including `substs` and added tests for that. Modified the suggestion as indicated above, happy for comments.

clippy_lints/src/functions/mod.rs
```diff
@@ -217,17 +217,62 @@ declare_clippy_lint! {
...
+    /// ```
+    /// pub enum ParseError {
+    ///     DoesntMakeMuchSenseAnyway,
+    ///     UnparsedBytes(Box<[u8; 512]>),
+    /// }
+    ///
+    /// // The `Result` is slightly large than a pointer
+    /// pub fn parse() -> Result<(), ParseError> {
+    ///     todo!()
+    /// }
```
Alexendoo, line 255: Couple of doc nits:

```suggestion
    ///     UnparsedBytes([u8; 512]),
    ///     UnexpectedEof,
    /// }
    ///
    /// // The `Result` has at least 512 bytes, even in the `Ok`-case
    /// pub fn parse() -> Result<(), ParseError> {
    ///     Ok(())
    /// }
    /// ```
    /// should be
    /// ```
    /// pub enum ParseError {
    ///     UnparsedBytes(Box<[u8; 512]>),
    ///     UnexpectedEof,
    /// }
    ///
    /// // The `Result` is slightly larger than a pointer
    /// pub fn parse() -> Result<(), ParseError> {
    ///     Ok(())
    /// }
```

Alexendoo (commented):
Sorry for the delay! Just a couple nits then if you squash the commits this should be ready to merge

lukaslueg:
Addressed the nit, squashed. One may decide to close #6560, #4652, and/or #3884 once merged.

lukaslueg:
*yells at CI for catching things `cargo test` doesn't*

Alexendoo (approved):

Alexendoo:
Great, thanks for the lint!

I stuck a `changelog: ..` entry in the PR message

@bors r+

lukaslueg:
@Alexendoo Thanks for the review

PR #9381, opened 2022-08-26
Don't lint `needless_return` if `return` has attrs

### LINKED ISSUE #9361: 'unneeded return' fires when return is needed due to attribute
### Description

Consider code like this:
```rust
        #[allow(clippy::integer_arithmetic)] // cannot overflow
        return discriminant << data_size | data;
```
The `return` is needed here because attributes can only be applied to statements, not expressions. But nevertheless clippy says I should remove the `return`...

### Version

```text
rustc 1.65.0-nightly (4b695f7c4 2022-08-21)
binary: rustc
commit-hash: 4b695f7c4e1a02d160fe7e159abd0f87027c0fcf
commit-date: 2022-08-21
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

_No response_

lukaslueg:
Fixes #9361

The lint used to have a mechanic to allow `cfg`-attrs on naked `return`-statements. This was well-intentioned, yet we can have any kind of attribute, e.g. `allow`, `expect` or even custom `derive`. So the mechanic was simply removed. We now never lint on a naked `return`-statement that has attributes on it.

Turns out that the ui-test had a Catch22 in it: In `check_expect()` the `#[expect(clippy::needless_return)]` is an attribute on the `return` statement that can and will be rustfixed away without side effects. But any other attribute would also have been removed, which is what #9361 is about. The test proved the wrong thing. Removed the test, the body is tested elsewhere as well.

changelog: Ignore [`needless_return`] on `return`s with attrs

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -2,7 +2,6 @@ use clippy_utils::diagnostics::span_lint_hir_and_then;
 use clippy_utils::source::{snippet_opt, snippet_with_context};
 use clippy_utils::{fn_def_id, path_to_local_id};
 use if_chain::if_chain;
-use rustc_ast::ast::Attribute;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, FnKind, Visitor};
 use rustc_hir::{Block, Body, Expr, ExprKind, FnDecl, HirId, MatchSource, PatKind, StmtKind};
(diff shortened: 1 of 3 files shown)


dswij (approved):

dswij:
This looks good, thanks for the PR!

@bors r+

dswij:
Test failed because the PR doesn't have changelog. I edited the PR description with the changelog, hope you don't mind with that @lukaslueg

@bors retry

PR #9389, opened 2022-08-28
Don't lint literal `None` from expansion

lukaslueg:
This addresses https://github.com/rust-lang/rust-clippy/pull/9288#issuecomment-1229398524: If the literal `None` is from expansion, we never lint. This is correct because e.g. replacing the call to `option_env!` with whatever that macro expanded to at the time of linting is certainly wrong.

changelog: Don't lint [`partialeq_to_none`] for macro-expansions


llogiq:
That was fast!

@bors r+

llogiq:
(sorry @Alexendoo for stealing the review work :smile:)

PR #9397, opened 2022-08-30
Fix the emission order of `trait_duplication_in_bounds`

Jarcho:
Makes the lint emit in source order rather than whatever order the hash map happens to be in. This is currently blocking the sync into rustc.

changelog: None

### DIFF
--- a/tests/ui/trait_duplication_in_bounds_unfixable.stderr
+++ b/tests/ui/trait_duplication_in_bounds_unfixable.stderr
@@ -1,8 +1,8 @@
 error: this trait bound is already specified in the where clause
-  --> $DIR/trait_duplication_in_bounds_unfixable.rs:6:23
+  --> $DIR/trait_duplication_in_bounds_unfixable.rs:6:15
    |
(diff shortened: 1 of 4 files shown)


dswij (commented):
Change itself looks good to me. How does this issue cause the sync to fail?

tests/ui/trait_duplication_in_bounds_unfixable.stderr
```diff
@@ -1,8 +1,8 @@
 error: this trait bound is already specified in the where clause
-  --> $DIR/trait_duplication_in_bounds_unfixable.rs:6:23
+  --> $DIR/trait_duplication_in_bounds_unfixable.rs:6:15
    |
 LL | fn bad_foo<T: Clone + Default, Z: Copy>(arg0: T, arg1: Z)
-   |                       ^^^^^^^
+   |               ^^^^^
```
Jarcho, line 5: This and the next warning were both emitted with the order depending on a hash map. I guess there was a change recently that caused the order to change.

dswij (approved):

dswij:
Thanks for the fix! @bors r+

PR #9404, opened 2022-08-31
Fix `suboptimal_float` not linting on `{const}.powf({const})`

### LINKED ISSUE #9201: Use cube root function lint
### What it does

The lint suggest to replace a FP::powf() call with a different method call.

### Lint Name

use_cube_root_function

### Category

perf

### Advantage

It's more precise, shorter and simpler.

### Drawbacks

None, I think.

### Example

```rust
#![warn(clippy::all)]
#![warn(clippy::nursery)]
#![warn(clippy::pedantic)]
fn main() {
    println!("{}", 1.5_f64.powf(1.0 / 3.0));
}
```

Could be written as:

```rust
#![warn(clippy::all)]
#![warn(clippy::nursery)]
#![warn(clippy::pedantic)]
fn main() {
    println!("{}", 1.5_f64.cbrt());
}
```
Also note that they output slightly different results. cbrt is more precise.

### LINKED ISSUE #9402: `imprecise_flops` not raised in some situations
### Summary

The lint `imprecise_flops` is not raised when a literal is used in combination with `powf` (and maybe other functions as well). Discovered in https://github.com/rust-lang/rust-clippy/issues/9201.

### Reproducer

I tried this code:

```rust
#![warn(clippy::imprecise_flops)]

fn main() {
    let _ = 1.5_f64.powf(1.0 / 3.0);
}
```

**I expected to see this happen:**

The lint `imprecise_flops` is raised on the line with a message such as "cube-root of a number can be computed more accurately".

**Instead, this happened:**

No lint is raised.

**However**, it does work as expected with the following code:

```rust
#![warn(clippy::imprecise_flops)]

fn main() {
    let x = 1.5_f64;
    let _ = x.powf(1.0 / 3.0);
}

// λ cargo clippy
//     Checking testuff v0.1.0 (/home/np/code/rust/testuff)
// warning: cube-root of a number can be computed more accurately
//  --> src/main.rs:5:13
//   |
// 5 |     let _ = x.powf(1.0 / 3.0);
//   |             ^^^^^^^^^^^^^^^^^ help: consider using: `x.cbrt()`
//   |
// note: the lint level is defined here
//  --> src/main.rs:1:9
//   |
// 1 | #![warn(clippy::imprecise_flops)]
//   |         ^^^^^^^^^^^^^^^^^^^^^^^
//   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#imprecise_flops
```

### Version

```text
rustc 1.65.0-nightly (addacb587 2022-08-24)
binary: rustc
commit-hash: addacb5878b9970ebc1665768a05cb601e7aea15
commit-date: 2022-08-24
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

_No response_

lukaslueg:
There used to be an early return if the receiver was an effective const but the method was not linted, not taking into account later cases where the receiver and the arguments are both effective consts for different methods. Removed the early return.

Fixes #9402
Fixes #9201

changelog: Fix `suboptimal_flops`, `imprecise_flops` not linting on `{const}.powf({const})` et al

### DIFF
--- a/tests/ui/floating_point_powf.rs
+++ b/tests/ui/floating_point_powf.rs
@@ -18,6 +18,11 @@ fn main() {
     let _ = x.powf(-16_777_215.0);
     let _ = (x as f32).powf(-16_777_215.0);
     let _ = (x as f32).powf(3.0);
+    let _ = (1.5_f32 + 1.0).powf(1.0 / 3.0);
+    let _ = 1.5_f64.powf(1.0 / 3.0);
+    let _ = 1.5_f64.powf(1.0 / 2.0);
+    let _ = 1.5_f64.powf(3.0);
+
     // Cases where the lint shouldn't be applied
     let _ = x.powf(2.1);
     let _ = x.powf(-2.1);
--- a/clippy_lints/src/floating_point_arithmetic.rs
+++ b/clippy_lints/src/floating_point_arithmetic.rs
@@ -238,23 +238,23 @@ fn get_integer_from_float_constant(value: &Constant) -> Option<i32> {
 fn check_powf(cx: &LateContext<'_>, expr: &Expr<'_>, args: &[Expr<'_>]) {
     // Check receiver
     if let Some((value, _)) = constant(cx, cx.typeck_results(), &args[0]) {
-        let method = if F32(f32_consts::E) == value || F64(f64_consts::E) == value {
-            "exp"
+        if let Some(method) = if F32(f32_consts::E) == value || F64(f64_consts::E) == value {
+            Some("exp")
         } else if F32(2.0) == value || F64(2.0) == value {
-            "exp2"
+            Some("exp2")
         } else {
-            return;
-        };
-
-        span_lint_and_sugg(
-            cx,
-            SUBOPTIMAL_FLOPS,
-            expr.span,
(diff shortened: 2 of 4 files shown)


tests/ui/floating_point_powf.rs
```diff
@@ -18,6 +18,9 @@ fn main() {
     let _ = x.powf(-16_777_215.0);
     let _ = (x as f32).powf(-16_777_215.0);
     let _ = (x as f32).powf(3.0);
+    let _ = (1.5_f32 + 1.0).powf(1.0 / 3.0);
+    let _ = 1.5_f64.powf(1.0 / 3.0);
```
giraffate, line 22: These tests is for `imprecise_flops`. Can the tests for `suboptimal_flops` be added? 
```suggestion
    let _ = 1.5_f64.powf(1.0 / 3.0);
    let _ = 1.5_f64.powf(1.0 / 2.0);
    let _ = 1.5_f64.powf(3.0);
```

giraffate (commented):

lukaslueg:
Good catch, I didn't see that there were two distinct lints. Both get triggered in the same execution path though.

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #9410, opened 2022-09-01
Use macro callsite when creating `Sugg` helper

### LINKED ISSUE #9375: collapsible_if fix suggestion breaks in the presence of `match!`
### Summary

When the outer `if` condition involves a `match!` (but the `match!` is not the whole condition), the `collapsible_if` fix suggestion will replace it with the macro definition (quite badly), completely breaking the output.

### Reproducer

I tried this code:

```rust
pub enum Kind {
    Cat,
    Dog,
}

pub fn meow(kind: Kind, meowy: bool, awake: bool) {
    if matches!(kind, Kind::Cat) && meowy {
        if awake {
            println!("MEOW");
        }
    }
}
```

suggestion changing the function body to:

```rust
    if matches!(kind, Kind::Cat) && meowy && awake {
        println!("MEOW");
    }
```

Instead, this happened:

```rust
warning: this `if` statement can be collapsed
  --> src/lib.rs:7:5
   |
7  | /     if matches!(kind, Kind::Cat) && meowy {
8  | |         if awake {
9  | |             println!("MEOW");
10 | |         }
11 | |     }
   | |_____^
   |
   = note: `#[warn(clippy::collapsible_if)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_if
help: collapse nested if block
   |
7  ~     if match $expression {
8  +             $( $pattern )|+ $( if $guard )? => true,
9  +             _ => false
10 +         } && meowy && awake {
11 +         println!("MEOW");
12 +     }
   |
```

When used with `--fix`, this results in an error asking for a bug report.

### Version

```text
rustc 1.65.0-nightly (addacb587 2022-08-24)
binary: rustc
commit-hash: addacb5878b9970ebc1665768a05cb601e7aea15
commit-date: 2022-08-24
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

dswij:
Closes #9375 

changelog: Improvement: [`collapsible_if`]: Suggestions now work with macros, by taking the call site into account.

### DIFF
--- a/clippy_utils/src/sugg.rs
+++ b/clippy_utils/src/sugg.rs
@@ -22,7 +22,7 @@ use std::fmt::{Display, Write as _};
 use std::ops::{Add, Neg, Not, Sub};
 
 /// A helper type to build suggestion correctly handling parentheses.
-#[derive(Clone, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub enum Sugg<'a> {
     /// An expression that never needs parentheses such as `1337` or `[0; 42]`.
     NonParen(Cow<'a, str>),
@@ -177,11 +177,11 @@ impl<'a> Sugg<'a> {
     pub fn ast(cx: &EarlyContext<'_>, expr: &ast::Expr, default: &'a str) -> Self {
         use rustc_ast::ast::RangeLimits;
 
-        let get_whole_snippet = || {
-            if expr.span.from_expansion() {
-                snippet_with_macro_callsite(cx, expr.span, default)
+        let snippet_without_expansion = |cx, span: Span, default| {
+            if span.from_expansion() {
+                snippet_with_macro_callsite(cx, span, default)
             } else {
-                snippet(cx, expr.span, default)
+                snippet(cx, span, default)
             }
         };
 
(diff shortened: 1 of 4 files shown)


clippy_utils/src/sugg.rs
```diff
@@ -177,11 +177,11 @@ impl<'a> Sugg<'a> {
     pub fn ast(cx: &EarlyContext<'_>, expr: &ast::Expr, default: &'a str) -> Self {
         use rustc_ast::ast::RangeLimits;
 
-        let get_whole_snippet = || {
-            if expr.span.from_expansion() {
-                snippet_with_macro_callsite(cx, expr.span, default)
+        let snippet_without_expansion = |cx, span: Span, default| {
```
dswij, line 180: I'm not sure if we want this in clippy utils

xFrednet: For now, I would keep it local to this function. Macros are not the simplest thing to deal with, and I prefer contributors to ignore them rather than trying to add suggestions for them. But that is mostly my personal preference, you can extract it if you want :upside_down_face:

xFrednet (approved):
Looks good to me, thank you for the bug fix.

I left an answer to your question. You can resolve it or extract the function, either is fine. And then you can `r=me` :upside_down_face:

clippy_utils/src/sugg.rs
dswij: That's fair. I'll leave it as it is then

dswij:
Thanks for the review! @bors r=@xFrednet

PR #9418, opened 2022-09-02
Fix `mut_mutex_lock` when Mutex is behind immutable deref

### LINKED ISSUE #9415: False positive mut_mutex_lock on &mut Arc<Mutex<_>>
### Summary

Clippy lints a mut_mutex_lock lint on `&mut Arc<Mutex<_>>`, which is obviously incorrect.

Applying the suggestion causes the error:
> cannot borrow data in an `Arc` as mutable

### Lint Name

mut_mutex_lock

### Reproducer

I tried this code:

```rust
use std::sync::{Arc, Mutex};

fn main() {
    let mut arc_mutex = Arc::new(Mutex::new(()));
    let arc_mutex: &mut Arc<Mutex<()>> = &mut arc_mutex;

    let _guard = arc_mutex.lock().unwrap();
}
```

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=bc5728f0494a32c7483c0d2f6fc9dc00

I saw this happen:

```
warning: calling `&mut Mutex::lock` unnecessarily locks an exclusive (mutable) reference
 --> src/main.rs:7:23
  |
7 |     let _ = arc_mutex.lock().unwrap();
  |                       ^^^^ help: change this to: `get_mut`
  |
  = note: `#[warn(clippy::mut_mutex_lock)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#mut_mutex_lock
```

No warnings.

### Version

```text
rustc 1.65.0-nightly (2e35f954a 2022-09-01)
binary: rustc
commit-hash: 2e35f954ada0f0c777844dc4fa66684efe90a035
commit-date: 2022-09-01
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

@rustbot label  +I-suggestion-causes-error

lukaslueg:
I *think* the problem here is the `if let ty::Ref(_, _, Mutability::Mut) = cx.typeck_results().expr_ty(recv).kind()` line tries to check if the `Mutex` can be mutably borrowed (there already is a test for `Arc<Mutex<_>>`), but gets bamboozled by the `&mut Arc` indirection. And I *think* checking the deref-adjustment to filter immutable-adjust (the deref through the `Arc`, starting from `&mut Arc`) is the correct fix.

Fixes #9415

changelog: Fix `mut_mutex_lock` when Mutex is behind immutable deref

### DIFF
--- a/clippy_lints/src/methods/mut_mutex_lock.rs
+++ b/clippy_lints/src/methods/mut_mutex_lock.rs
@@ -1,5 +1,5 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::{expr_custom_deref_adjustment, ty::is_type_diagnostic_item};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, Mutability};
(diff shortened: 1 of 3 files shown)


llogiq:
Thank you!

@bors r+

Issue #9317, opened 2022-08-10
unnecessary use of `to_string` for a type with custom `ToString` trait


xphoniex:
### Summary

I'm programmatically commiting a patch to a git repo and then writing the hash (a [`git2::Oid`](https://docs.rs/git2/latest/git2/struct.Oid.html)) in a file:

```
let commit = util::commit(...);
fs::write(path, commit);
```

if I use above code, I'll get bytes written to the file whereas I wanted the string of hash, so naturally I change it to `fs::write(path, commit.to_string())`, however that's throwing a lint error:

```
warning: unnecessary use of `to_string`
   --> ...
    |
810 |             commit.to_string(),
    |             ^^^^^^^^^^^^^^^^^^ help: use: `commit`
    |
    = note: `#[warn(clippy::unnecessary_to_owned)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_to_owned
```

solution is to use `format!(...)` instead, but that's pretty ugly.

### Lint Name

unnecessary_to_owned

### Reproducer

I tried this code:

```rust
<code>
```

I saw this happen:

```
<output>
```

I expected to see this happen:

### Version

```text
rustc 1.61.0 (fe5b13d68 2022-05-18)
binary: rustc
commit-hash: fe5b13d681f25ee6474be29d748c65adcd91f69e
commit-date: 2022-05-18
host: x86_64-unknown-linux-gnu
release: 1.61.0
LLVM version: 14.0.0
```

Jarcho:
Note to anyone working on this. `to_string` should only be linted when the type is string-like. e.g. implements `Deref<Target=str/String>` or `AsRef<str/String>`.

DesmondWillowbrook:
I wasn't able to replicate your report with rust v1.61.0 locally nor v1.63.0 ([playground for v1.63.0 - Tools - Clippy ](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8dd84082d027d82a1cc197f115b8d469))
Am I perhaps missing any relevant trait implementations of `Oid`?

rcgoodfellow:
Hitting this too. A simple repro is

main.rs
```rust
fn main() {
    let id = uuid::Uuid::new_v4();

    std::fs::write("bad", &id).unwrap();
    std::fs::write("good", id.to_string()).unwrap();
}
```
Cargo.toml
```toml
[package]
name = "poo"
[...]

PR #9424, opened 2022-09-03
Fix `unnecessary_to_owned` false positive

### LINKED ISSUE #9351: `unnecessary_to_owned` doesn't take type changes into account
### Summary

`unnecessary_to_owned` suggests changing the type of an input but doesn't check if this type is used in other places. This can lead to suggestions that won't compile due to mismatched type errors.

### Lint Name

unnecessary_to_owned

### Reproducer

I tried this code:

```rust
#![allow(unused_variables)]
use std::ops::Deref;
use std::path::PathBuf;

fn main() {
    let path = std::path::Path::new("x");
    let x: PathBuf = require_deref_path(path.to_owned());
    // let x: PathBuf = require_deref_path(path);
}

fn require_deref_path<T: Deref<Target = std::path::Path>>(x: T) -> T {
    x
}
```

I saw this happen:

```
warning: unnecessary use of `to_owned`
 --> src/main.rs:7:41
  |
7 |     let x: PathBuf = require_deref_path(path.to_owned());
  |                                         ^^^^^^^^^^^^^^^ help: use: `path`
  |
```

Applying the suggestion results in this error
```
error[[E0308]](https://doc.rust-lang.org/stable/error-index.html#E0308): mismatched types
  --> src/main.rs:8:41
   |
8  |     let x: PathBuf = require_deref_path(path);
   |                      ------------------ ^^^^- help: try using a conversion method: `.to_path_buf()`
   |                      |                  |
   |                      |                  expected struct `PathBuf`, found `&Path`
   |                      arguments to this function are incorrect
   |
note: function defined here
  --> src/main.rs:11:4
   |
11 | fn require_deref_path<T: Deref<Target = std::path::Path>>(x: T) -> T {
```

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d8f0c9b97e862a49063b2a08d6113205)

No warning.

### Version

```text
Clippy 0.1.65 (2022-08-16 86c6ebe)
```

### Additional Labels

@rustbot +label I-suggestion-causes-error

mikerite:
Fixes #9351.

Note that this commit reworks that fix for #9317. The change
is to check that the type implements `AsRef<str>` before regarding
`to_string` as an equivalent of `to_owned`. This was suggested
by Jarcho in the #9317 issue comments.

The benefit of this is that it moves some complexity out of
`check_other_call_arg` and simplifies the module as a whole.

changelog: FP: [`unnecessary_to_owned`]: No longer lints, if type change would cause errors in the caller function

### DIFF
--- a/clippy_lints/src/methods/unnecessary_to_owned.rs
+++ b/clippy_lints/src/methods/unnecessary_to_owned.rs
@@ -1,21 +1,25 @@
 use super::implicit_clone::is_clone_like;
 use super::unnecessary_iter_cloned::{self, is_into_iter};
+use crate::rustc_middle::ty::Subst;
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::ty::{
-    get_associated_type, get_iterator_item_ty, implements_trait, is_copy, is_type_diagnostic_item, peel_mid_ty_refs,
-};
-use clippy_utils::{fn_def_id, get_parent_expr, is_diag_item_method, is_diag_trait_item};
+use clippy_utils::ty::{get_associated_type, get_iterator_item_ty, implements_trait, is_copy, peel_mid_ty_refs};
+use clippy_utils::visitors::find_all_ret_expressions;
+use clippy_utils::{fn_def_id, get_parent_expr, is_diag_item_method, is_diag_trait_item, return_ty};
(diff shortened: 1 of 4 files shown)


xFrednet (approved):
Looks good to me and lintcheck also *agrees* with these changes :upside_down_face:

xFrednet:
Thank you for the update. I've adjusted the changelog entry a bit :)

@bors r+

Issue #9435, opened 2022-09-07
Using `unwrap_or_default()` in place of `unwrap_or(<default>)`


xphoniex:
### Description

Currently this is used:

```
num.unwrap_or(0);
```

This is ok for non-default values, but I propose we suggest `unwrap_or_default` instead for default values of [primitive types](https://doc.rust-lang.org/rust-by-example/primitives.html), which all happen to be 0, and `false` for bools.

So instead we'd write:

```
num.unwrap_or_default();
```

Our syntax is longer, but it'd hard to come up with a case where default value has more characters than keyword "default" itself, which is 7 chars long. If we have an `unwrap_or_default` method, it makes sense to use it.

Comments please.

tbu-:
> If we have an `unwrap_or_default` method, it makes sense to use it.

It only makes sense if it makes the intent clearer. I don't see that for most default values, `.unwrap_or(false)`, `.unwrap_or(0)`, `.unwrap_or(None)` all seem clearer (and shorter!) than `.unwrap_or_default()`.

PR #9436, opened 2022-09-07
Suggest using `unwrap_or_default()` instead of `unwrap_or(<default>)`

xphoniex:
Closes https://github.com/rust-lang/rust-clippy/issues/9435

changelog: [`or_fun_call`]: Suggest using `unwrap_or_default()` instead of `unwrap_or(<default>)`

### DIFF
--- a/clippy_lints/src/exhaustive_items.rs
+++ b/clippy_lints/src/exhaustive_items.rs
@@ -90,7 +90,7 @@ impl LateLintPass<'_> for ExhaustiveItems {
                     (EXHAUSTIVE_ENUMS, "exported enums should not be exhaustive")
                 };
                 let suggestion_span = item.span.shrink_to_lo();
-                let indent = " ".repeat(indent_of(cx, item.span).unwrap_or(0));
+                let indent = " ".repeat(indent_of(cx, item.span).unwrap_or_default());
                 span_lint_and_then(
                     cx,
                     lint,
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -77,7 +77,7 @@ pub(super) fn check<'tcx>(
             let mut block_str = snippet_with_applicability(cx, block.span, "..", &mut applicability).into_owned();
             block_str.remove(0);
             block_str.pop();
-            let indent = " ".repeat(indent_of(cx, block.stmts[0].span).unwrap_or(0));
+            let indent = " ".repeat(indent_of(cx, block.stmts[0].span).unwrap_or_default());
 
             // Reference iterator from `&(mut) []` or `[].iter(_mut)()`.
             if !prefix.is_empty() && (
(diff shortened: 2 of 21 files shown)


matthiaskrgr:
I don't like this because it ads an unnecessary layer of abstraction without benefit.

fhartwig:
I think using `unwrap_or(specific_value)` is generally better than `unwrap_or_default()` because it's clearer and doesn't require the reader of that line to know what the default value of the relevant type is.

xphoniex:
> I don't like this because it ads an unnecessary layer of abstraction to the code without benefit.

By "layer of abstraction", do you mean syntax-wise or is there a performance penalty when using `unwrap_or_default`

xphoniex:
> I think using `unwrap_or(specific_value)` is generally better than `unwrap_or_default()` because it's clearer and doesn't require the reader of that line to know what the default value of the relevant type is.

Makes sense. But then, why do we have a `unwrap_or_default` in the first place?

xphoniex:
Hey @erickt , I went through the [history](https://github.com/rust-lang/rust/blame/04c58c9f81c5d9b6b1847573a41c8b819893a12a/src/libstd/option.rs#L430) and it seems like you're the one who introduced `unwrap_or_default` syntax. (Correct me if I'm wrong)

Would you care to comment on this?

afetisov:
`unwrap_or_default` can cause issues with type inference, because previously the type was determined by the inner expression.

For primitives, this also looks needlessly complex and obfuscated. The `.unwrap_or(0)` is shorter, specifies the type (so no inference errors) and is explicit about the value. While the default `0: uN` is simple, it's one more thing to remember, and doesn't carry its weight for this simple expression.

It also seems that the lint performs constant evaluation (i.e. folds `.unwrap_or(FOO)` to `.unwrap_or(0)` and lints it). This is just wrong. The entire point of constants is to abstract away the specific value, so that it can easily be changed later, and to give that value a clear and meaningful name in the current context. This lint violates both of those goals.

Finally, I expect this to create a lot of churn, so if added, this lint must be `restriction` or `pedantic`.

erickt:
I agree with @afetisov. If I remember correctly, I added `unwrap_or_default` a long time before 1.0, back when we were just feeling out how APIs would work with traits. Nowadays after reading a lot of rust code I’m not sure if it’s really saving reducing that much cognitive load, and I can’t remember ever actually using it.

afetisov:
I don't have an issue with `unwrap_or_default` when it's used for complex types, and the type is clear for the context (foremost to preserve type inference). For example, I don't think the explicit type expression adds anything in this case:
```rust
let map: HashMap<u64, String> = make_map();
let s = map.get(idx).unwrap_or_else(String::new);
```

The expression is more complex, you need to remember to use `unwrap_or_else` instead of `unwrap_or` and to pass a function (otherwise Clippy will somewhat reasonably complain). If you change the type of values from String to `Vec<Stuff>`, then you need to change the default value expressions for no good reason, and if you use `Default::default` as that expression, then it just `unwrap_or_default` but longer and clunkier.

But the for primitive types an explicit initializer is much more clear.

flip1995:
@xphoniex As we talked before about this (I think in DMs?), I agree with all the others here: `unwrap_or_default` doesn't really make sense with primitives. Just writing the primitive in `unwrap_or(0)` is shorter and contains more information. In addition to all the other arguments that were brought up here.

The reason this lint exists for more complex types is the exact reason @afetisov brought up in the above comment: https://github.com/rust-lang/rust-clippy/pull/9436#issuecomment-1270515858

We can add an extra lint that just lints this for primitives. But then again, I don't see who would use a lint like that. 

So I'm strongly leaning towards not merging this PR.

xphoniex:
Thanks everyone for the input, I'm closing this PR.

PR #9441, opened 2022-09-07
Fix hang in `vec_init_then_push`

### LINKED ISSUE #9433: clippy hangs
### Summary

clippy seems to hang with this code:
````rust
// run-pass
// Example taken from RFC 1238 text

// https://github.com/rust-lang/rfcs/blob/master/text/1238-nonparametric-dropck.md
//     #examples-of-code-that-must-continue-to-work

use std::cell::Cell;

struct Concrete<'a>(#[allow(unused_tuple_struct_fields)] u32, Cell<Option<&'a Concrete<'a>>>);

fn main() {
    let mut data = Vec::new();
    data.push(Concrete(0, Cell::new(None)));
    data.push(Concrete(0, Cell::new(None)));

    data[0].1.set(Some(&data[1]));
    data[1].1.set(Some(&data[0]));
}
````

### Version

```text
clippy 0.1.65 (c07a8b4 2022-08-26)
```

### Error output

_No response_

Jarcho:
fixes #9433
changelog: Fix infinite loop in `vec_init_then_push`

### DIFF
--- a/clippy_lints/src/vec_init_then_push.rs
+++ b/clippy_lints/src/vec_init_then_push.rs
@@ -86,7 +86,7 @@ impl VecPushSearcher {
                 },
                 ExprKind::Unary(UnOp::Deref, _) | ExprKind::Index(..) if !needs_mut => {
                     let mut last_place = parent;
-                    while let Some(parent) = get_parent_expr(cx, parent) {
+                    while let Some(parent) = get_parent_expr(cx, last_place) {
                         if matches!(parent.kind, ExprKind::Unary(UnOp::Deref, _) | ExprKind::Field(..))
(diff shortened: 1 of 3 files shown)


Alexendoo (approved):

Alexendoo:
👍

@bors r+

flip1995:
This issue probably already exists for quite some time. The fix will get into `beta` anyway. We might want to backport this hang fix to beta, so it will hit stable next week. Since the fix is pretty small, I'll probably do that.

flip1995:
https://github.com/rust-lang/rust/pull/101808

Jarcho:
It would be since #8699.

PR #9443, opened 2022-09-08
Rename the arithmetic lint

c410-f3r:
changelog: Rename the `arithmetic` lint


c410-f3r:
Had to re-bless some tests

PR #9446, opened 2022-09-08
Fix `range_{plus,minus}_one` bad suggestions

### LINKED ISSUE #9431: `range_plus_one` doesn't handle macros correctly
### Summary

When a macro contains this item, Clippy suggests removing the macro call with the correction.

### Reproducer

I tried this code:

```rust
#![warn(clippy::range_plus_one)]

macro_rules! m {
    () => {
        for i in 0..4 + 1 {
            println!("{}", i);
        }
    }
}

fn main() {
    m!();
}
```

A warning on the code inside the macro.

Instead, this happened:
```
warning: an inclusive range would be more readable
  --> src/main.rs:12:5
   |
12 |     m!();
   |     ^^^^ help: use: `0..=4`
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9a31cfc1c95cfcf916e59baa11da572c)

### Version

```text
Clippy 0.1.65 (2022-09-04 289279d)
```

### Additional Labels

mikerite:
Fixes #9431.

The current `range_plus_one` and `range_minus_one` suggestions are completely incorrect when macros are involved.

This commit resolves this by disabling the lints for any range expression that is expanded from a macro. The reasons for this are that it is very difficult to create a correct suggestion in this case and that false negatives are less important for pedantic lints.

changelog: Fix `range_{plus,minus}_one` bad suggestions

### DIFF
--- a/clippy_lints/src/ranges.rs
+++ b/clippy_lints/src/ranges.rs
@@ -350,6 +350,7 @@ fn check_range_bounds<'a>(cx: &'a LateContext<'_>, ex: &'a Expr<'_>) -> Option<R
 // exclusive range plus one: `x..(y+1)`
 fn check_exclusive_range_plus_one(cx: &LateContext<'_>, expr: &Expr<'_>) {
     if_chain! {
+        if expr.span.can_be_used_for_suggestions();
         if let Some(higher::Range {
             start,
             end: Some(end),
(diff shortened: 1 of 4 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #9447, opened 2022-09-08
Rustup

flip1995:
r? @ghost

changelog: none


clippy_utils/src/macros.rs
```diff
@@ -644,7 +644,7 @@ impl<'tcx> Count<'tcx> {
                 span,
                 values,
             )?),
-            rpf::Count::CountIsParam(_) => {
+            rpf::Count::CountIsParam(_) | rpf::Count::CountIsStar(_) => {
```
nyurik, line 649: @Jarcho I would like to split up these two use cases because the `.*` is more like an implied autonumbered position (similar to `{}`).  I did the change as part of my PR https://github.com/rust-lang/rust-clippy/pull/9233/commits/3ac65936c5751c5d5539d718b643212701d77ce4 -- note that it now detects more cases because of that change.  Please let me know if this is an ok approach. 

cc: @flip1995

Issue #8822, opened 2022-05-12
Suggest use of Option::filter()


tomers:
### What it does

I stumbled upon cases when developers are not familiar enough with all the functions that come with standard libs, such as `Option::filter()`. Instead, they program that functionality each time, which makes the code much more complex, error prone, and less readable.

### Lint Name

filter

### Category

style, complexity

### Advantage

- Reduces complexity substantially.
- Improves code readability.
- Makes use of lib function instead of coding its functionality in each usage.

### Drawbacks

None that I can think of.

### Example

```rust
match s {
    None => None,
    Some(x) => {
        if x.is_empty() {
            None
        } else {
            Some(x)
        }
    }
}
```

Could be written as:

```rust
s.filter(|x| !x.is_empty())
```

Full example:
```rust
fn bad(s: Option<String>) -> Option<String> {
    match s {
        None => None,
        Some(x) => {
            if x.is_empty() {
                None
            } else {
                Some(x)
            }
        }
    }
}

fn good(s: Option<String>) -> Option<String> {
    s.filter(|x| !x.is_empty())
}

fn main() {
    assert_eq!(bad(None), None);
    assert_eq!(bad(Some("".to_owned())), None);
    assert_eq!(bad(Some("foo".to_owned())), Some("foo".to_owned()));
    assert_eq!(good(None), None);
    assert_eq!(good(Some("".to_owned())), None);
    assert_eq!(good(Some("foo".to_owned())), Some("foo".to_owned()));
}
```

kraktus:
Another example while I can think about it

```rust
option match {
      Some(x) if cond(x) => x,
      _ => x
}
```

PR #9451, opened 2022-09-09
Add `manual_filter` lint for `Option`

kraktus:
Share much of its implementation with `manual_map` and should greatly benefit from its previous feedback.
I'm sure it's possible to even more refactor both and would gladly take input on that as well as any clippy idiomatic usage, since this is my first lint addition.

I've added the lint to the complexity section for now, I don't know if every new lint needs to go in nursery first.

The matching could be expanded to more than `Some(<value>)` to lint on arbitrary struct matching inside the `Some` but I've left it like it was for `manual_map` for now. `needless_match::pat_same_as_expr` provides a more generic match example.

close https://github.com/rust-lang/rust-clippy/issues/8822

changelog: Add lint [`manual_filter`] for `Option`


dswij:
> I don't know if every new lint needs to go in nursery first.

There's no requirement to go to nursery first. I'll try to run this on the most popular crates to see if there's lots of FPs. If not, then I think complexity is fine.

dswij:
I ran this on 900 most popular crates. [Here's the result](https://gist.github.com/dswij/2d6be62da2fe4e1b5fd8bfae3ad1ce9e).

At first glance, it seems to work as expected. I'll try taking a look further when time hopefully allows

clippy_lints/src/matches/manual_filter.rs
```diff
@@ -0,0 +1,191 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::{is_lang_ctor, path_to_local_id};
+use rustc_hir::intravisit::{walk_expr, Visitor};
+use rustc_hir::LangItem::OptionSome;
+use rustc_hir::{Arm, Block, BlockCheckMode, Expr, ExprKind, HirId, Pat, PatKind, UnsafeSource};
+use rustc_lint::LateContext;
+use rustc_span::{sym, SyntaxContext};
+
+use super::manual_map::{check_with, SomeExpr};
```
dswij, line 10: Maybe we better have this common functions somewhere better

clippy_lints/src/matches/manual_filter.rs
```diff
@@ -0,0 +1,191 @@
...
+                    ..
+                },
+                _,
+            ) => {
+                self.0 = true;
+            },
+            _ => walk_expr(self, expr),
+        }
+    }
+}
```
dswij, line 31: We have a check for unsafe block in the util, maybe we can re-use that for this? https://github.com/rust-lang/rust-clippy/blob/e43f5a17d4d9263616ae6fc28a7c5353e80d8a12/clippy_utils/src/visitors.rs#L446-L475

clippy_lints/src/matches/manual_filter.rs
```diff
@@ -0,0 +1,191 @@
...
+
+// function called for each <ifelse> expression:
+// Some(x) => if <cond> {
+//    <ifelse>
+// } else {
+//    <ifelse>
+// }
+// Returns true if <ifelse> resolves to `Some(x)`, `false` otherwise
+fn handle_if_or_else_expr<'tcx>(
```
dswij, line 86: `handle_if_or_else_expr` seems a little bit confusing as name, when it "checks" whether the if block contains an `Option`

clippy_lints/src/matches/manual_filter.rs
```diff
@@ -0,0 +1,191 @@
...
+            },
+            [arg],
+        ) = inner_expr.kind
+        {
+            return ctxt == if_or_else_expr.span.ctxt()
+                && is_lang_ctor(cx, qpath, OptionSome)
+                && path_to_local_id(arg, target);
+        }
+    };
+    false
```
dswij, line 107: Can we use [clippy utils](https://github.com/rust-lang/rust-clippy/blob/master/clippy_utils/src/higher.rs#L56-L83)' `If` helper for HIR to simplify this logic?

clippy_lints/src/matches/manual_filter.rs
```diff
@@ -0,0 +1,191 @@
...
+    scrutinee: &'tcx Expr<'_>,
+    arms: &'tcx [Arm<'_>],
+    expr: &'tcx Expr<'_>,
+) {
+    let ty = cx.typeck_results().expr_ty(expr);
+    if_chain! {
+        if is_type_diagnostic_item(cx, ty, sym::Option);
+        if arms.len() == 2;
+        if arms[0].guard.is_none();
+        if arms[1].guard.is_none();
```
dswij, line 133: style nit:
```suggestion
        if let [first_arm, second_arm] == arms;
        // this might be a little bit too fancy
        // if iter::once(first_arm, second_arm).all(|x| x.guard.is_none());
        if first_arm.guard.is_none() && second_arm.guard.is_none();
```

clippy_lints/src/matches/manual_map.rs
```diff
@@ -53,12 +51,59 @@ fn check<'tcx>(
...
+                )
+            },
+            sugg_info.app,
+        );
+    }
+}
+
+#[expect(clippy::too_many_arguments)]
+#[expect(clippy::too_many_lines)]
+pub(super) fn check_with<'tcx, F>(
```
dswij, line 88: See above on moving this to a shared module

clippy_lints/src/matches/manual_filter.rs
```diff
@@ -0,0 +1,191 @@
...
+            },
+            _ => walk_expr(self, expr),
+        }
+    }
+}
+
+// Function called on the `expr` of `[&+]Some((ref | ref mut) x) => <expr>`
+// Need to check if it's of the `if <cond> {<then_expr>} else {<else_expr>}`
+// AND that only one `then/else_expr` resolves to `Some(x)` while the other resolves to `None`
+// return `cond` if
```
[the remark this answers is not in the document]
kraktus: It wasn't 😄 Edited

clippy_lints/src/matches/manual_filter.rs
kraktus: moved them to `matches/manual_utils.rs`

clippy_lints/src/matches/manual_filter.rs
kraktus: perfect, thanks for the suggestion 👍

clippy_lints/src/matches/manual_filter.rs
kraktus: I don't see how here, I think you were confused by the wrong function name, edited it and its comments

kraktus:
Sounds good! Build failures are on crate where the lint does not trigger indeed

dswij:
Thanks for this! @bors r+

PR #9452, opened 2022-09-09
Fix dev book

kraktus:
fix `implements_trait` and `in_external_macro` import path

Remove example using `match_trait_method` since its deprecated.

changelog: none


flip1995:
@bors r+

Thanks!

Issue #9450, opened 2022-09-09
Fix for `assertions_on_result_states` incorrectly changes return type


glittershark:
### Summary

`assert!(res.is_err())` always returns `()`, but `res.unwrap_err()` returns the err type itself - this can cause auto-fixed code to fail to compile if there's a missing semicolon at the end of a test, for example.

### Reproducer

I tried this code:

```rust
#[test]
fn my_test() {
    let res: Result<i32, i32> = Ok(1);
    assert!(res.is_err())
}
```

I expected to see this happen:

```rust
#[test]
fn my_test() {
    let res: Result<i32, i32> = Ok(1);
    res.unwrap_err();
}
```

Instead, this happened:
```rust
#[test]
fn my_test() {
    let res: Result<i32, i32> = Ok(1);
    res.unwrap_err()
}
```

causing a type error

### Version

```text
❯ rustc -Vv
rustc 1.64.0-nightly (fe3342816 2022-08-01)
binary: rustc
commit-hash: fe3342816a282949f014caa05ea2e669ff9d3d3c
commit-date: 2022-08-01
host: x86_64-unknown-linux-gnu
release: 1.64.0-nightly
LLVM version: 14.0.6
```

### Additional Labels

@rustbot label +I-suggestion-causes-err

PR #9453, opened 2022-09-10
[`assertions_on_result_states`]: Fix suggestion when `assert!` is not in a statement.

kraktus:
fix https://github.com/rust-lang/rust-clippy/issues/9450

changelog: [`assertions_on_result_states`]: Fix suggestion when `assert!` is not in a statement.


Jarcho:
Thank you. @bors r+

Issue #9440, opened 2022-09-07
use_self False positive on enums that derive Serialize


laralove143:
### Summary

When an enum has `#[derive(Serialize)]`, it triggers the `use_self` lint

### Lint Name

use_self

### Reproducer

I tried this code:

```rust
#![warn(clippy::use_self)]

use serde::Serialize;

#[derive(Serialize)]
pub enum Foo {
    Bar = 1,
}
```

I saw this happen:

```
warning: unnecessary structure name repetition
 --> src/lib.rs:6:10
  |
6 | pub enum Foo {
  |          ^^^ help: use the applicable keyword: `Self`
  |
note: the lint level is defined here
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::use_self)]
  |         ^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self

warning: `playground` (lib) generated 1 warning
```

I expected to see this happen:
No warnings

### Version

```text
Playground
```

kraktus:
duplicate of https://github.com/rust-lang/rust-clippy/issues/6902

Issue #8910, opened 2022-05-29
-Wclippy::nursery generates use_self when enum derives strum_macros::Display


Guilherme-Vasconcelos:
### Summary

I am using the [strum](https://crates.io/crates/strum) and [strum_macros](https://crates.io/crates/strum_macros) crates, both on version 0.24.0.

When an enum derives `strum_macros::Display` and clippy is invoked with `-Wclippy::nursery` (i.e. `cargo clippy -- -Wclippy::nursery`), a warning is generated telling me to replace the enum's name with Self.

### Lint Name

clippy::use-self

### Reproducer

I tried this code:

```rust
//// on Cargo.toml
[package]
name = "testing"
edition = "2021"
version = "0.0.1"

[dependencies]
strum = "0.24.0"
strum_macros = "0.24.0"

//// on src/main.rs
mod bug;

fn main() {}

//// on src/bug.rs
use strum_macros::Display;

#[allow(dead_code)]
#[derive(Display)]
enum MyTest {
    Testing
}
```

I saw this happen:

```
warning: unnecessary structure name repetition
 --> src/bug.rs:5:6
  |
5 | enum MyTest {
  |      ^^^^^^ help: use the applicable keyword: `Self`
  |
  = note: `-W clippy::use-self` implied by `-W clippy::nursery`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self

warning: `testing` (bin "testing") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
```

`enum Self` here would not be valid.

### Version

```text
rustc 1.61.0 (Arch Linux rust 1:1.61.0-1)
binary: rustc
commit-hash: unknown
commit-date: unknown
host: x86_64-unknown-linux-gnu
release: 1.61.0
LLVM version: 13.0.1
```

giraffate:
It might be a duplicate of https://github.com/rust-lang/rust-clippy/issues/6902.

Issue #6902, opened 2021-03-14
use_self still gives false positives


imp:
Now that #6818 is fixed and the fix (#6833) has made its way to the nightly there are still cases of false positives:

Lint name: `use_self`

I tried this code:

```rust
#![deny(clippy::use_self)]
use serde::Deserialize;

#[derive(Debug, Deserialize)]
#[serde(untagged)]
enum Direction {
    North,
}

fn main() {
    let d = Direction::North;

    println!("{:?}", d);
}
```

I expected to see this happen: clippy clean - no warning/errors

Instead, this happened:

```
    Checking playground v0.0.1 (/playground)
error: unnecessary structure name repetition
 --> src/main.rs:6:6
  |
6 |   enum Direction {
  |  ______^
7 | |     North,
  | |____^ help: use the applicable keyword: `Self`
  |
note: the lint level is defined here
 --> src/main.rs:1:9
  |
1 | #![deny(clippy::use_self)]
  |         ^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self

error: aborting due to previous error

error: could not compile `playground`

To learn more, run the command again with --verbose.
```

Interestingly enough it the code generated by `#[serde(untagged)]` that triggers this lint. If this line is commented out then clippy seems to be happy

### Meta

- `cargo clippy -V`: `clippy 0.1.52 (2021-03-13 acca818)`

- `rustc -Vv`:
 ```
rustc 1.52.0-nightly (acca81892 2021-03-13)
binary: rustc
commit-hash: acca818928654807ed3bc1ce0e97df118f8716c8
commit-date: 2021-03-13
host: x86_64-unknown-linux-gnu
release: 1.52.0-nightly
LLVM version: 12.0.0
 ```

Y-Nak:
This issue would be automatically fixed when https://github.com/rust-lang/rust/issues/84122 is addressed.

barafael:
For reference, here are two cases in which the same false positive occurs:
* [smart-default](https://docs.rs/smart-default/latest/smart_default/) with the `SmartDefault` derive macro
* [num-derive](https://docs.rs/num-derive/latest/num_derive/) with the `FromPrimitive` and `ToPrimitive` derive macros
    - This PR on `num-derive` proposes to stop using the type name and instead use `Self` in the generated code, solving the problem inside `num-derive`: https://github.com/rust-num/num-derive/pull/49

[...]

PR #9454, opened 2022-09-10
Do not lint `use_self` in proc macro expansion

kraktus:
fix https://github.com/rust-lang/rust-clippy/issues/9440
fix https://github.com/rust-lang/rust-clippy/issues/8910
fix https://github.com/rust-lang/rust-clippy/issues/6902

changelog: [`use_self`]: Do not lint in proc macro expansion

### DIFF
--- a/clippy_lints/src/use_self.rs
+++ b/clippy_lints/src/use_self.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::ty::same_type_and_consts;
-use clippy_utils::{meets_msrv, msrvs};
+use clippy_utils::{is_from_proc_macro, meets_msrv, msrvs};
 use if_chain::if_chain;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::Applicability;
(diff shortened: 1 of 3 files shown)


clippy_lints/src/use_self.rs
```diff
@@ -103,6 +103,7 @@ impl<'tcx> LateLintPass<'tcx> for UseSelf {
             if parameters.as_ref().map_or(true, |params| {
                 !params.parenthesized && !params.args.iter().any(|arg| matches!(arg, GenericArg::Lifetime(_)))
             });
+            if !is_from_proc_macro(cx, item); // expensive, should be last check
```
Alexendoo, line 106: Would `item_path.span.from_expansion()` be enough here?

kraktus: Just tried testing it, and `from_expansion` still produces the false positive.

Alexendoo: huh, what about for `item.span`?

Jarcho: The issue is the proc macro using a real span instead of making a new one.

Alexendoo: The path span happened to do that but the whole item doesn't, so checking its span should be fine

flip1995:
@bors r+

Thanks!

clippy_lints/src/use_self.rs
flip1995: It depends what span the proc macro sets. If it doesn't set a span with expansion info, but just uses the span given from the item/tokens it looks to clippy as if it is unexpanded code.

MrNossiom:
In which version will this patch be released ?

PR #9458, opened 2022-09-10
Use `visit_expr_field` for `ParamPosition`

Alexendoo:
A small change to make it a little simpler

changelog: none

### DIFF
--- a/clippy_utils/src/macros.rs
+++ b/clippy_utils/src/macros.rs
@@ -8,7 +8,7 @@ use arrayvec::ArrayVec;
 use itertools::{izip, Either, Itertools};
 use rustc_ast::ast::LitKind;
 use rustc_hir::intravisit::Visitor;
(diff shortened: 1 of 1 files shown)


giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

nyurik:
I think this PR broke format parsing.  To test, add this to [tests/ui/explicit_write.rs](https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/explicit_write.rs) (e.g. around line 39) -- before this PR, it would suggest a replacement with `eprintln!("{:w$}", val, w = width);`, but now it fails to parse and silently ignores.

```rust
        let val = 1;
        let width = 2;
        writeln!(std::io::stderr(), "{:w$}", val, w = width).unwrap();
```

PR #9461, opened 2022-09-10
BREAKING CHANGE fix running path/to/cargo-clippy --fix

matthiaskrgr:
*[View all comments](https://triagebot.infra.rust-lang.org/gh-comments/rust-lang/rust-clippy/pull/9461)*

Previously, we would assume clippy is called as "cargo clippy --arg1 --arg2" so we stripped the first two items and parsed "--arg1" and "--arg2" as cmdline args.
The problem is when we run cargo-clippy binary directly as "target/debug/cargo-clippy --arg1 --arg2", we would still remove the first two args which would be "..cargo-clippy" and "--arg1" in this case which is wrong.

Fix this by checking if we run clippy via "cargo clippy" or "../path/to/cargo-clippy" and for the latter, only skip the first arg instead of two.

Spotted by Kraktus because lintcheck --fix was no longer working

This is potentially a breaking change because the "target/debug/cargo-clippy clippy --fix" will no longer work!
Use "target/debug/cargo-clippy --fix" directly instead.

---

changelog: Important change: `cargo-clippy` will now no longer ignore the first argument. `cargo-clippy clippy` will still work as documented (the second `clippy` is ignored). However, arguments like `--fix` will issue a warning for now and will no longer be ignored in a few releases. `cargo clippy` (without `-`!) is unaffected by this change.
[#9461](https://github.com/rust-lang/rust-clippy/pull/9461)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,4 +1,5 @@
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
+#![feature(let_chains)]
 // warn on lints, that are included in `rust-lang/rust`s bootstrap
 #![warn(rust_2018_idioms, unused_lifetimes)]
 
@@ -43,7 +44,16 @@ pub fn main() {
         return;
     }
 
-    if let Err(code) = process(env::args().skip(2)) {
+    // if we run "cargo clippy" (as cargo subcommand), we have to strip "cargo clippy" (first 2 args)
+    // but if we are run via "..../target/debug/cargo-clippy", only ommit the first arg, the second one
+    // might be a normal cmdline arg already (which we don't want to ommit)
+    let args = if let Some(first_arg) = env::args().next() && first_arg.ends_with("cargo-clippy") {
+        env::args().skip(1)
+    } else {
+        env::args().skip(2)
+    };
+
+    if let Err(code) = process(args) {
         process::exit(code);
     }
 }
@@ -215,5 +225,15 @@ mod tests {
         let args = "cargo clippy".split_whitespace().map(ToString::to_string);
         let cmd = ClippyCmd::new(args);
         assert_eq!("check", cmd.cargo_subcommand);
+    }
+
+    #[test]
+    fn dont_skip_arg() {
+        let args = "target/debug/cargo-clippy --fix"
+            .split_whitespace()
+            .map(ToString::to_string);
+        let cmd = ClippyCmd::new(args);
+        assert_eq!("fix", cmd.cargo_subcommand);
+        assert!(!cmd.args.iter().any(|arg| arg.ends_with("unstable-options")));
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,7 +47,7 @@ pub fn main() {
     // if we run "cargo clippy" (as cargo subcommand), we have to strip "cargo clippy" (first 2 args)
     // but if we are run via "..../target/debug/cargo-clippy", only ommit the first arg, the second one
     // might be a normal cmdline arg already (which we don't want to ommit)
-    let args = if let Some(first_arg) = env::args().next() && first_arg.ends_with("cargo-clippy") {
+    let args = if let Some(first_arg) = env::args().next() && (first_arg.ends_with("cargo-clippy") || first_arg.ends_with("cargo-clippy.exe")) {
         env::args().skip(1)
     } else {
         env::args().skip(2)
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,8 +47,15 @@ pub fn main() {
     // if we run "cargo clippy" (as cargo subcommand), we have to strip "cargo clippy" (first 2 args)
     // but if we are run via "..../target/debug/cargo-clippy", only ommit the first arg, the second one
     // might be a normal cmdline arg already (which we don't want to ommit)
-    let args = if let Some(first_arg) = env::args().next() && (first_arg.ends_with("cargo-clippy") || first_arg.ends_with("cargo-clippy.exe")) {
-        env::args().skip(1)
+    let args = if let Some(first_arg) = env::args().next()
+        && (first_arg.ends_with("cargo-clippy") || first_arg.ends_with("cargo-clippy.exe"))
+    {
+        // turn this of during the migration
+        // env::args().skip(1)
+        eprintln!(
+            "WARNING: potentially breaking change: 'cargo-clippy arg1 arg2...' will no longer ignore 'arg1' after edition=2024"
+        );
+        env::args().skip(2)
     } else {
         env::args().skip(2)
     };
--- a/tests/dogfood.rs
+++ b/tests/dogfood.rs
@@ -98,7 +98,6 @@ fn run_clippy_for_package(project: &str, args: &[&str]) -> bool {
     command
         .current_dir(root_dir.join(project))
         .env("CARGO_INCREMENTAL", "0")
-        .arg("clippy")
         .arg("--all-targets")
         .arg("--all-features");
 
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -67,7 +68,16 @@ pub fn main() {
         return;
     }
 
-    if let Err(code) = process(env::args().skip(2)) {
+    // if we run "cargo clippy" (as cargo subcommand), we have to strip "cargo clippy" (first 2 args)
+    // but if we are run via "..../target/debug/cargo-clippy", only ommit the first arg, the second one
+    // might be a normal cmdline arg already (which we don't want to ommit)
+    let args = if let Some(first_arg) = env::args().next() && (first_arg.ends_with("cargo-clippy") || first_arg.ends_with("cargo-clippy.exe")) {
```
Alexendoo, line 74: We can avoid any breaking change by checking if the 2nd arg is `clippy`, rather than the first being an invocation of `cargo-clippy[.exe]`

matthiaskrgr: ah so that `cargo clippy` but also `cargo-clippy clippy` still works? clever! :D

I'll revisit this tomorrow after a good portion of sleep :)

Jarcho: That doesn't avoid the breaking change. Anything that uses `cargo-clippy --` would break. Or anything else for that matter.

kraktus: That is what is happening for `lintcheck` for example

Alexendoo: Ah right since any arg can currently be used as the separator

Alexendoo: `cargo clippy` does execute with the first arg as `.../cargo-clippy` though, so we'd still need a different approach

```
execve("/home/alex/.cargo/bin/cargo", ["cargo", "clippy"], 0xffffd242db68 /* 49 vars */) = 0
execve("/home/alex/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/bin/cargo", ["/home/alex/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/bin/cargo", "clippy"], 0xaaaae39b8940 /* 54 vars */) = 0
execve("/home/alex/.cargo/bin/cargo-clippy", ["/home/alex/.cargo/bin/cargo-clippy", "clippy"], 0xaaab0d612f40 /* 57 vars */) = 0
execve("/home/alex/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/bin/cargo-clippy", ["/home/alex/.rustup/toolchains/stable-aarch64-unknown-linux-gnu/bin/cargo-clippy", "clippy"], 0xaaaafea6a490 /* 57 vars */) = 0```

Comment (matthiaskrgr) [reply src/main.rs:74]: I just checked clippy `1.50` which was the lastest version when I wrote lintcheck 2 years ago and it looks like 
````
RUSTC_BOOTSTRAP=1 ~/.rustup/toolchains/1.50-x86_64-unknown-linux-gnu/bin/cargo-clippy --fix -Zunstable-options
````
actually never worked as intended :no_mouth:  :sweat_smile: :sweat_smile:

**Comment by matthiaskrgr** [comment]
so to recap:
`path/to/cargo-clippy --fix` // this doesn't work but I would like it to
`cargo clippy --fix` // this works
`cargo-clippy clippy --fix` // this works but its super weird imo, I personally have no objections to dropping support but its probably better to keep backwards compatibility here :upside_down_face: 
`cargo clippy -- --fix` // this also works but it also looks a bit weird to me
`path/to/cargo-clippy -- --fix` // this also works

**Comment by Jarcho** [comment]
I think the least breaking change we could do here is to check if the first argument is one of clippy's argument (currently
`--no-deps`, `--fix`, `-h`, `--help`, `-V`, `--version`, `--explain`). This does leave an inconsistency with `--` being ignored if it's the first argument, so it only half fixes the problem. `cargo-clippy -- -W some-lint` would still not work.

Alternatively, we could just treat anything with a leading dash as an argument. This will break some scripts, but it will make any direct use of `cargo-clippy` just work. I think this will make everything fail with an invalid argument error, rather than silently running with the wrong arguments.

**Comment by flip1995** [comment]
Doing a [GitHub search](https://cs.github.com/?q=%2Fcargo-clippy+(clippy|-)%2F+NOT+language%3AMarkdown&scope=&scopeName=All+repos) I mostly find `cargo-clippy clippy` things. Or `cargo-clippy --version`, which works as expected. Also `--help` and `--explain` work as the first argument. I would say, as long as we keep supporting `cargo-clippy clippy`, we can just break every other usage of `cargo-clippy` where the second argument is used as a separator. I nominate the breakage discussion for the next meeting.

**Comment by flip1995** [comment]
[Clippy meeting](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202022-10-18/near/304711108) result: 

- We don't want to hard break this now.
- We want to keep `cargo-clippy clippy` as it works now
- We want to emit warnings for all flags that are affected by this bug
    - `--fix`
    - `--no-deps` (not sure if this is affected)
    - Find out if other flags are affected
    - Emit a warning if `cargo-clippy` is followed by one of those flags, but don't change behavior yet
- After 1-3 release cycles, change the behavior and remove the warnings 
    - doesn't have to happen in the same release cycle
    - We can remove the warnings a release cycle later

**Comment by xFrednet** [comment]
Oh, and we should probably add a big disclaimer to the changelog :)

**Comment by xFrednet** [comment]
I've now updated the changelog entry according to the meeting results listed above :)

**Comment by xFrednet** [comment]
I'm marking this as waiting on author, until the discussed changes have been implemented

@rustbot author

**Comment by xFrednet** [comment]
Hey @matthiaskrgr, I'd like to announce this change one or two changelogs, before the version that puts this into effect. Can you address the comments and make this PR ready to be merged? Then I could include it in the `1.67` changelog to be added in `1.69` :upside_down_face:

**Comment by flip1995** [comment]
> `cargo-clippy` now expects `clippy` to be the first argument.

This is not correct. I was unclear / ambiguous in my summary.

The warning we want to emit for `cargo-clippy --*` that exists today is that those flags will **have an effect** in the future. Currently they don't, because you need _something_ (usually `clippy`) before the first flag.

We want to keep supporting `cargo-clippy clippy` as a special case to be backwards compatible. Currently `cargo-clippy foo-bar-baz ...` and `cargo-clippy clippy` are exactly the same. So what we want to change here is to only remove the first arg after `cargo-clippy`, if it is actually `clippy`.

**Comment by xFrednet** [comment]
That makes a lot of sense, I had the feeling my understanding was not quite correct. Could you take a second look at the suggested changelog entry?

**Comment by flip1995** [comment]
Updated the changelog.

**Comment by xFrednet** [comment]
Ping @matthiaskrgr, I'll mention this change in the next changelog, which unblocks this PR from being merged. Can you get this up to speed, with a rebase?

**Comment by matthiaskrgr** [comment]
Uuh i may need a bit to wrap my head around this again after 1.5 years I fear 😅

**Comment by xFrednet** [comment]
Very understandable ^^ (Sorry for the delay). The changelog doesn't mention a specific date or release, so there is no pressure from that side.

**Comment by flip1995** [comment]
In the last meeting we decided to not emit a warning, but rather break the behavior on a [edition boundary](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202024-05-14/near/438599616).

**Comment by xFrednet** [comment]
Is there already a date for the edition? And what is the best way to find out when it's scheduled to be released?

**Comment by flip1995** [comment]
I don't think there's a date for the edition. Not sure who's the edition manager though.

File: src/main.rs [hunk 2]
```diff
@@ -43,7 +44,23 @@ pub fn main() {
...
+        // env::args().skip(1)
+        eprintln!(
+            "WARNING: potentially breaking change: 'cargo-clippy arg1 arg2...' will no longer ignore 'arg1' after edition=2024"
+        );
+        env::args().skip(2)
+    } else {
+        env::args().skip(2)
+    };
+
+    if let Err(code) = process(args) {
```
matthiaskrgr, line 63: :thinking:

hkBst:
Yesterday I was looking at the cargo-clippy code and wondering about the `skip(2)` causing problems.

What's the current plan forwards? Aim for the next edition?

PR #9465, opened 2022-09-11
Fix `unused_peekable` closure and `f(&mut peekable)` false positives

Alexendoo:
changelog: Fix [`unused_peekable`] false positive when peeked in a closure or called as `f(&mut peekable)`

The `return`/`break` changes aren't part of the fix, they allow an earlier return in some cases. `break` is replaced with `return` for style purposes as they do the same thing in this case


Alexendoo:
Not sure on the timing of things but I'll add https://github.com/rust-lang/rust-clippy/labels/beta-nominated here too

flip1995:
@bors r+

Thanks! Since this lint was already reported twice I will do a dummy sync with this commit.

Issue #9460, opened 2022-09-10
Clippy wrongly reports clippy::large_stack_arrays on static struct


tokatoka:
### Summary

Hello, I have a problem with clippy::large_stack_arrays lint.

Clippy reports clippy::large_stack_arrays on a `public static mut` struct defined like

```
/// The global `CmpLog` map for the current `LibAFL` run.
#[no_mangle]
pub static mut libafl_cmplog_map: CmpLogMap = CmpLogMap {
    headers: [CmpLogHeader {
        hits: 0,
        shape: 0,
        kind: 0,
    }; CMPLOG_MAP_W],
    vals: CmpLogVals {
        operands: [[CmpLogInstruction(0, 0); CMPLOG_MAP_H]; CMPLOG_MAP_W],
    },
};
```
and here's the short output from clippy
```
error: allocating a local array larger than 512000 bytes
   --> libafl_targets/src/cmplog.rs:165:19
    |
165 |         operands: [[CmpLogInstruction(0, 0); CMPLOG_MAP_H]; CMPLOG_MAP_W],
```
clippy suggests to put this on heap, but it is not possible because it's a static struct.

Thank you!

### Lint Name

clippy::large_stack_arrays

### Reproducer

1. Clone this repository: `git clone https://github.com/AFLplusplus/LibAFL.git`
2. Checkout to: `git checkout b86314282907aa77d727e3bb9a1b03b75ad441c0`
3. Run clippy: `cargo +nightly clippy --all --all-features`

Clippy reports
```
error: allocating a local array larger than 512000 bytes
   --> libafl_targets/src/cmplog.rs:165:19
    |
165 |         operands: [[CmpLogInstruction(0, 0); CMPLOG_MAP_H]; CMPLOG_MAP_W],
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: the lint level is defined here
   --> libafl_targets/src/lib.rs:7:9
    |
7   | #![deny(clippy::pedantic)]
    |         ^^^^^^^^^^^^^^^^
    = note: `#[deny(clippy::large_stack_arrays)]` implied by `#[deny(clippy::pedantic)]`
    = help: consider allocating on the heap with `vec![[CmpLogInstruction(0, 0); CMPLOG_MAP_H]; CMPLOG_MAP_W].into_boxed_slice()`
[...]

PR #9466, opened 2022-09-11
Don't lint `large_stack_array` inside static items

lukaslueg:
We now check if the linted `Expr` is inside an `ItemKind::Static`, which can't take the suggested `Box<[...]`. I _think_ this is the correct fix for #9460

I removed `if_chain` while I was at it.

changelog: Don't lint `large_stack_array` inside static items


dswij (approved):
LGTM, thanks for this

PR #9469, opened 2022-09-12
Fix FormatArgsExpn parsing of FormatSpec positions

### LINKED ISSUE #9468: [regression] Some format expressions no longer parsed by FormatArgsExpn::parse
### Summary

After #9458, some complex format expressions no longer return a value from `FormatArgsExpn::parse(...)`.

### Reproducer

To test, add this to [tests/ui/explicit_write.rs](https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/explicit_write.rs) (e.g. around line 39).

```rust
        let val = 1;
        let width = 2;
        writeln!(std::io::stderr(), "{:w$}", val, w = width).unwrap();
```

I expected it to produce a suggestion `eprintln!("{:w$}", val, w = width);`, but instead it now silently ignores the above.  Reverting 1b245e7e0e83d387ab9d2cc7a157b081ba5d0ef2 and re-testing starts showing the expected suggestion.

### Version

```text
rustc 1.65.0-nightly (c2804e6ec 2022-09-07)
binary: rustc
commit-hash: c2804e6ec2c29a5c7368600ea173b890e2655c3d
commit-date: 2022-09-07
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

_No response_

Alexendoo:
Woops, forgot visitors don't walk themselves

Fixes #9468

r? @giraffate

changelog: none

### DIFF
--- a/clippy_utils/src/macros.rs
+++ b/clippy_utils/src/macros.rs
@@ -7,7 +7,7 @@ use crate::visitors::expr_visitor_no_bodies;
 use arrayvec::ArrayVec;
 use itertools::{izip, Either, Itertools};
 use rustc_ast::ast::LitKind;
-use rustc_hir::intravisit::Visitor;
+use rustc_hir::intravisit::{walk_expr, Visitor};
 use rustc_hir::{self as hir, Expr, ExprField, ExprKind, HirId, Node, QPath};
 use rustc_lexer::unescape::unescape_literal;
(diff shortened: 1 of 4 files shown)


giraffate (approved):

giraffate:
@bors r+

Nice catch, thanks!

PR #9471, opened 2022-09-13
Add matches! checking to nonstandard_macro_braces

jplatte:
changelog: Enhancement: [`nonstandard_macro_braces`]: Now includes `matches!()` in the default lint config
  [#9471](https://github.com/rust-lang/rust-clippy/pull/9471)

### DIFF
--- a/clippy_lints/src/nonstandard_macro_braces.rs
+++ b/clippy_lints/src/nonstandard_macro_braces.rs
@@ -184,6 +184,10 @@ fn macro_braces(conf: FxHashSet<MacroMatcher>) -> FxHashMap<String, (String, Str
             name: "vec",
             braces: ("[", "]"),
         ),
+        macro_matcher!(
+            name: "matches",
+            braces: ("(", ")"),
+        ),
     ]
     .into_iter()
     .collect::<FxHashMap<_, _>>();
--- a/tests/ui-toml/nonstandard_macro_braces/conf_nonstandard_macro_braces.rs
+++ b/tests/ui-toml/nonstandard_macro_braces/conf_nonstandard_macro_braces.rs
@@ -42,6 +42,7 @@ macro_rules! printlnfoo {
 fn main() {
     let _ = vec! {1, 2, 3};
     let _ = format!["ugh {} stop being such a good compiler", "hello"];
+    let _ = matches!{{}, ()};
     let _ = quote!(let x = 1;);
     let _ = quote::quote!(match match match);
     let _ = test!(); // trigger when macro def is inside our own crate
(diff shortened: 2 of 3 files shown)


clippy_lints/src/nonstandard_macro_braces.rs
```diff
@@ -184,6 +184,10 @@ fn macro_braces(conf: FxHashSet<MacroMatcher>) -> FxHashMap<String, (String, Str
             name: "vec",
             braces: ("[", "]"),
         ),
+        macro_matcher!(
+            name: "matches",
+            braces: ("(", ")"),
+        ),
```
kraktus, line 190: I think it would be nice to add a test even though the change is trivial so that someone modifying the code latter won't break it

jplatte: This is just default configuration, if there are no tests at all for this feature I can kind of see your point, otherwise not. In any case I unfortunately don't have much time to figure out how to add completely new testing for this feature, but if there are already some tests and you insist of having them extended for this change, let me know where to find them.

kraktus: [The tests are here](https://github.com/rust-lang/rust-clippy/blob/master/tests/ui-toml/nonstandard_macro_braces/conf_nonstandard_macro_braces.rs). If you don't have time to add them I don't think it's a blocker, I can add them later

jplatte: Added a test and squashed everything. The output in the particular test case I've added is a *little* strange since it re-"format"s `{}` to `()` inside the macro invocation. Should I open an issue about that or is it something known from other lint suggestions that already has an issue?

kraktus: I've opened an issue: https://github.com/rust-lang/rust-clippy/issues/9498 Looks like it's there from the initial implementation. Would you like to fix it at the same time? If not it can be done independently

kraktus: I've opened https://github.com/rust-lang/rust-clippy/pull/9499 and rebase on your work so you won't have any conflit

xFrednet:
Since I'm assigned to the follow-up PR and flip is quite busy AFAIK, I'm taking over this review :upside_down_face: 

r? @xFrednet 

@jplatte have you already seen the PR of @kraktus, and do we want to close this in favor of it? Otherwise, I can just review this first :upside_down_face:

xFrednet (approved):

xFrednet:
Looks good to me, I'm merging this as is, since the follow-up PR includes this commit as well. Thank you for the addition and thank you to @kraktus for the feedback and follow-up PR :upside_down_face: 

FYI, I've updated the changelog entry a bit. The previous one was already pretty good :+1: 

@bors r+

PR #9475, opened 2022-09-13
Make module-style lints resilient to --remap-path-prefix

Nemo157:
changelog: [`self_named_module_files`], [`mod_module_files`]: Make module-style lints resilient to `--remap-path-prefix`

Without this if a user has configured `--remap-path-prefix` to be used for a prefix containing the current source directory the lints would silently fail to generate a warning.


tests/ui-cargo/module_style/fail_mod_remap/src/main.rs
```diff
@@ -0,0 +1,7 @@
+// compile-flags: --remap-path-prefix /=/remapped
```
Nemo157, line 1: I assume this will not work correctly on windows, not really sure what to do about it...

Nemo157: Found a solution in rust's testsuite.

xFrednet (approved):

xFrednet:
That's cool, I never even knew `--remap-path-prefix` existed. Today I learned. Thank you very much for this update :upside_down_face: 

@bors r+

PR #9478, opened 2022-09-14
Update rust-analyzer documentation, mention linkedProjects

Alexendoo:
r-a uses the `rustc-dev` component from the rustup installed toolchain clippy specifies so it doesn't need to be manually installed. Also remove references to nightly r-a as the feature is long stable

I discovered `rust-analyzer.linkedProjects` recently and it has made working on the crates not referenced by the `clippy` crate so much nicer

changelog: none


flip1995:
@bors r+

Thanks for the tip with the linked projects!

Issue #7730, opened 2021-09-28
`if-then-panic` suggestion discards comments


mikerite:
I tried this code:

```rust
#![warn(clippy::if_then_panic)]

fn main() {
    if false { 
        // comment
        panic!("ABC"); 
    };
}
```

I expected to see this happen: The suggestion would preserve the comment somehow.

Instead, this happened: The suggestion discarded the comment.
```
warning: only a `panic!` in `if`-then statement
 --> src/main.rs:5:5
  |
5 | /     if false { 
6 | |         // comment
7 | |         panic!("ABC"); 
8 | |     };
  | |_____^ help: try: `assert!(!false, "ABC");`
  |
```

### Meta

**Rust version (`rustc -Vv`):**

```
rustc 1.57.0-nightly (fdf65053e 2021-09-07)
binary: rustc
commit-hash: fdf65053e99e8966f9bd83b5a8491326cb33d638
commit-date: 2021-09-07
host: x86_64-unknown-linux-gnu
release: 1.57.0-nightly
LLVM version: 13.0.0
```
@rustbot label +L-suggestion

PR #9479, opened 2022-09-14
[`manual_assert`]: Preserve comments in the suggestion

kraktus:
close https://github.com/rust-lang/rust-clippy/issues/7730

changelog: [`manual_assert`]: Preserve comments in the suggestion

### DIFF
--- a/clippy_lints/src/manual_assert.rs
+++ b/clippy_lints/src/manual_assert.rs
@@ -56,9 +56,9 @@ impl<'tcx> LateLintPass<'tcx> for ManualAssert {
                     comments += "\n";
                 }
                 let (cond, not) = match cond.kind {
-                     ExprKind::Unary(UnOp::Not, e) => (e, ""),
-                     _ => (cond, "!"),
-                 };
+                    ExprKind::Unary(UnOp::Not, e) => (e, ""),
+                    _ => (cond, "!"),
+                };
                 let cond_sugg = sugg::Sugg::hir_with_applicability(cx, cond, "..", &mut applicability).maybe_par();
                 let sugg = format!("assert!({not}{cond_sugg}, {format_args_snip});");
                 // we show to the user the suggestion without the comments, but when applicating the fix, include the comments in the block
--- a/clippy_lints/src/manual_assert.rs
+++ b/clippy_lints/src/manual_assert.rs
@@ -1,9 +1,10 @@
-use clippy_utils::diagnostics::span_lint_and_sugg;
+use crate::rustc_lint::LintContext;
+use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::macros::{root_macro_call, FormatArgsExpn};
 use clippy_utils::source::snippet_with_applicability;
-use clippy_utils::{peel_blocks_with_stmt, sugg};
+use clippy_utils::{peel_blocks_with_stmt, span_extract_comment, sugg};
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind, UnOp};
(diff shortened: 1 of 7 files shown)


tests/ui/manual_assert.edition2021.fixed
```diff
@@ -27,8 +28,8 @@ fn main() {
     {
         panic!("qaqaq{:?}", a);
     }
-    assert!(a.is_empty(), "qaqaq{:?}", a);
-    assert!(a.is_empty(), "qwqwq");
+    assert!(!(!a.is_empty()), "qaqaq{:?}", a);
+    assert!(!(!a.is_empty()), "qwqwq");
```
dswij, line 32: Is this change in suggestion expected?

kraktus: It was expected yes. Ideally this should be handled by `Sugg` to avoid it but will eventually be simplified by the `non_minimal` bool. If it's a deal breaker can revert to the manual implementation in the lint

dswij: Just wondering, what's the issue with the previous implementation?

kraktus: I just felt it should be a case common enough not to be handled manually by each lint which needs to negate an expression 😄

dswij: That's fair, and I agree with it. But I think we don't need to rush to remove this from the lint at the moment 😄

We can leave this PR for the comment for now, and work on improving the above later (maybe move this feature to `Sugg`, etc etc.)

kraktus: Ok 👍 Actually the fix in `Sugg` wouldn't be trivial since it only store a string version of the snippet.

kraktus:
Another option would be to downgrade the applicability when there are comments, but at which level should this be done? `span_lint_and_sugg`?

clippy_lints/src/manual_assert.rs
```diff
@@ -50,20 +51,36 @@ impl<'tcx> LateLintPass<'tcx> for ManualAssert {
...
+                if !comments.is_empty() {
+                    comments += "\n";
+                }
                 let (cond, not) = match cond.kind {
-                    ExprKind::Unary(UnOp::Not, e) => (e, ""),
-                    _ => (cond, "!"),
-                };
+                     ExprKind::Unary(UnOp::Not, e) => (e, ""),
+                     _ => (cond, "!"),
+                 };
```
llogiq, line 61: What's up with the indentation?

llogiq (commented):
I think re-adding the comments on fix is good UX. Users can still do what they want with it.

There is one re-formatting glitch that I was wondering about, otherwise this is mergeworthy.

clippy_lints/src/manual_assert.rs
kraktus: I removed then re-introduced this snippet, weird fmt did not reformat the same way. Will try to remove manually the spaces

llogiq:
Weird. For some reason it couldn't compile the fixed code!?

kraktus:
I don't get the error locally using `TESTNAME=manual_assert cargo uitest`, but not super familiar with edition gated tests. iirc it should run them all though

Alexendoo:
Give it a try after a rebase onto master, 628a854ae215f7ce2a68294467c40febfa6650b9 came after the base of this PR which includes https://github.com/Manishearth/compiletest-rs/pull/258

kraktus:
After rebasing could see and fix the tests

llogiq:
Thank you!

@bors r+

Issue #9485, opened 2022-09-16
unnecessary closure used with `bool::then` in field definition


xerenahmed:
### Summary

....

### Lint Name

unnecessary_lazy_evaluations

### Reproducer

I tried this code:

```rust
// FromForm is from rocket (HTTP library)
#[derive(FromForm, Debug)]
pub struct CreateAlarm {
    pub label: String,
}
```

I saw this happen:

```
warning: unnecessary closure used with `bool::then`
  --> src/routes/alarm/create.rs:10:16
   |
10 |     pub label: String,
   |                ^^^^^^ help: use `then_some(..)` instead: `then_some(label)`
   |
   = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
```

I expected to see this happen: nothing

### Version

```text
rustc 1.65.0-nightly (29e4a9ee0 2022-08-10)
binary: rustc
commit-hash: 29e4a9ee0253cd39e552a77f51f11f9a5f1c41e6
commit-date: 2022-08-10
host: aarch64-apple-darwin
release: 1.65.0-nightly
LLVM version: 14.0.6
```

Issue #9439, opened 2022-09-07
`unnecessary_lazy_evaluations` gets triggered by a derive macro


griffi-gh:
### Summary

`unnecessary_lazy_evaluations` gets triggered by a `FromForm` macro (`rocket` `0.5-rc.2`)

### Lint Name

`unnecessary_lazy_evaluations`

### Reproducer

```rust
#[derive(FromForm)]
pub struct RegisterData<'a> {
  email: &'a str,
  username: &'a str,
  password: &'a str,
  repeat_password: Option<&'a str>,
}
```
...Or any other struct with `#[derive(FromForm)]`

![VSCodium_Lxazpouwme](https://user-images.githubusercontent.com/45996170/188950081-4f9e4f26-69ba-4cdd-a06c-9c2fc9465a9d.png)

### Version

```text
rustc 1.65.0-nightly (29e4a9ee0 2022-08-10)
binary: rustc
commit-hash: 29e4a9ee0253cd39e552a77f51f11f9a5f1c41e6
commit-date: 2022-08-10
host: x86_64-pc-windows-msvc
release: 1.65.0-nightly
LLVM version: 14.0.6
```

PR #9486, opened 2022-09-16
[`unnecessary_lazy_eval`] Do not lint in external macros

kraktus:
fix https://github.com/rust-lang/rust-clippy/issues/9485, fix https://github.com/rust-lang/rust-clippy/issues/9439

I didn't find a way to add a test since fps occur due to an external crate.

changelog: [`unnecessary_lazy_eval`] Do not lint in external macros

### DIFF
--- a/clippy_lints/src/methods/unnecessary_lazy_eval.rs
+++ b/clippy_lints/src/methods/unnecessary_lazy_eval.rs
@@ -1,7 +1,7 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::source::snippet;
 use clippy_utils::ty::is_type_diagnostic_item;
-use clippy_utils::{eager_or_lazy, usage};
+use clippy_utils::{eager_or_lazy, is_from_proc_macro, usage};
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_lint::LateContext;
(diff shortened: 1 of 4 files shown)


dswij:
@kraktus FYI, I've recently added a test for Derive macro FP [here](https://github.com/rust-lang/rust-clippy/pull/9043). Maybe we can do something similar?

kraktus:
Thanks, I've identified the source of the false positive in rocket codebase: https://github.com/SergioBenitez/Rocket/blob/707936177a7fea6d8eca103cf0b3601ca45fe657/core/codegen/src/derive/from_form.rs#L287

Following your example I've tried to reproduce the issue with a dummy derive macro
```rust
#[proc_macro_derive(LazyEvalDerive)]
pub fn unnecessary_lazy_eval(_: TokenStream) -> TokenStream {
    quote!(
        pub fn unnecessary_then() {
            Some(1).unwrap_or_else(|| 2); // should not lint because in external macro
        }
    )
}
```
but it's not linting it even before my patch, so unlikely it'll currently fix the issue

Jarcho:
The fix for this needs `is_from_proc_macro`. You can use the `with_span!` macro in tests using `aux-build:proc_macro_with_span.rs`.

dswij (approved):
LGTM, Thanks for this!

@giraffate Hope you don't mind me stealing this review 😁

Issue #9175, opened 2022-07-14
question-mark usage suggested in const fn


urben1680:
### Summary

Clippy suggests to use the `?` operator in a `const fn` where that operator is not allowed.

I hope I did this correctly, it is my first issue I fill in. Thanks for the great work on Clippy!

### Lint Name

question_mark

### Reproducer

I tried this code:

```rust
const fn clippy_false_positive(option: Option<()>) -> Option<()>{
    if option.is_none(){
        return None;
    }
    //stuff
    Some(())
}
```

I saw this happen:

```
warning: this block may be rewritten with the `?` operator
 --> src/main.rs:2:5
  |
2 | /     if option.is_none(){
3 | |         return None;
4 | |     }
  | |_____^ help: replace it with: `option?;`
  |
  = note: `#[warn(clippy::question_mark)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#question_mark

warning: `playground` (bin "playground") generated 1 warning
```

I expected to not see this to happen as the `?` operator is not allowed in a `const fn`. Trying to do so results in:

```
error[[E0658]](https://doc.rust-lang.org/stable/error-index.html#E0658): `?` is not allowed in a `const fn`
 --> src/main.rs:2:5
  |
2 |     option?;
  |     ^^^^^^^
  |
  = note: [see issue #74935 <https://github.com/rust-lang/rust/issues/74935>](https://github.com/rust-lang/rust/issues/74935) for more information

For more information about this error, try `rustc --explain E0658`.
```
https://github.com/rust-lang/rust/issues/74935

### Version

```text
rust: 1.62.0
clippy 0.1.64 (2022-07-13 87588a2)
```

PR #9487, opened 2022-09-16
Silence [`question_mark`] in const context

kraktus:
fix https://github.com/rust-lang/rust-clippy/issues/9175

When `const_try` is stabilised can be turned into a MSRV

changelog: Silence [`question_mark`] in const context


Jarcho:
Thank you. @bors r+

PR #9488, opened 2022-09-16
Add `#[allow(unused)]` to test in `cargo dev new_lint`

Alexendoo:
`rustfix` tests don't automatically apply `-Aunused` which leads to some tests having `_workarounds`, add it to new test files automatically so people don't have to worry about it

changelog: none


llogiq:
Thank you!

@bors r+

Issue #9383, opened 2022-08-26
`needless_borrow` and `explicit_auto_deref` false positive with unions


Arc-blroth:
### Summary

Clippy suggests removing explicit `&mut` and `*` derefs that are required by [RFC 2514][2514] when working with union fields.

[2514]: https://rust-lang.github.io/rfcs/2514-union-initialization-and-drop.html#potential-pitfalls-around-derefmut

### Lint Name

needless_borrow and explicit_auto_deref

### Reproducer

I tried this code:

```rust
use std::mem::ManuallyDrop;

union Coral {
    crab: ManuallyDrop<Vec<i32>>,
}

union Ocean {
    coral: ManuallyDrop<Coral>,
}

fn main() {
    let mut ocean = Ocean {
        coral: ManuallyDrop::new(Coral {
            crab: ManuallyDrop::new(vec![1, 2, 3]),
        }),
    };

    unsafe {
        // this expression borrows a value the compiler would automatically borrow
        ManuallyDrop::drop(&mut (&mut ocean.coral).crab);
        
        // deref which would be done by auto-deref
        (*ocean.coral).crab = ManuallyDrop::new(vec![4, 5, 6]);
        ManuallyDrop::drop(&mut (*ocean.coral).crab);
        
        ManuallyDrop::drop(&mut ocean.coral);
    }
}
```

I saw this happen:

```
warning: this expression borrows a value the compiler would automatically borrow
  --> src/main.rs:20:33
   |
20 |         ManuallyDrop::drop(&mut (&mut ocean.coral).crab);
   |                                 ^^^^^^^^^^^^^^^^^^ help: change this to: `ocean.coral`
   |
   = note: `#[warn(clippy::needless_borrow)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

warning: deref which would be done by auto-deref
  --> src/main.rs:23:9
   |
23 |         (*ocean.coral).crab = ManuallyDrop::new(vec![4, 5, 6]);
   |         ^^^^^^^^^^^^^^ help: try this: `ocean.coral`
   |
   = note: `#[warn(clippy::explicit_auto_deref)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#explicit_auto_deref

warning: deref which would be done by auto-deref
  --> src/main.rs:24:33
   |
24 |         ManuallyDrop::drop(&mut (*ocean.coral).crab);
[...]

PR #9490, opened 2022-09-17
fix [`needless_borrow`], [`explicit_auto_deref`] FPs on unions

kraktus:
fix https://github.com/rust-lang/rust-clippy/issues/9383

changelog: fix [`needless_borrow`] false positive on unions
changelog: fix [`explicit_auto_deref`] false positive on unions

Left a couple debug derived impls on purpose I needed to debug as I don't think it's noise

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -184,6 +184,7 @@ impl Dereferencing {
     }
 }
 
+#[derive(Debug)]
 struct StateData {
     /// Span of the top level expression
     span: Span,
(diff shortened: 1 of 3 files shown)


Jarcho:
It would be slightly better to add `Position::UnionFieldAccess` as a separate variant.

kraktus:
Can fix in a few days

kraktus:
Thank you for the tip, much cleaner implementation

Jarcho:
Thank you. @bors r+

PR #9507, opened 2022-09-21
[arithmetic-side-effects] Consider references

c410-f3r:
Takes into consideration integer references like `&i32::MAX` because currently things like `let _ = &1 + 0` trigger the lint.

changelog: FP: [`arithmetic_side_effects`]: Now ignores references
  [9507](https://github.com/rust-lang/rust-clippy/pull/9507)


c410-f3r:
It seems that @xFrednet is not currently available. Perhaps another reviewer?

r? rust-lang/clippy

c410-f3r:
r? @Alexendoo
Hope you don't mind

xFrednet:
Sorry, for the last replay, haven't had much time for Clippy in the last weeks. @Alexendoo It would be super if you could take this over. You can pass it back, if you don't have time :upside_down_face:

Alexendoo:
Yep no problem! LGTM

@bors r+

PR #9511, opened 2022-09-21
add `box-default` lint

llogiq:
This adds a `box-default` lint to suggest using `Box::default()` instead of `Box::new(Default::default())`, which offers less moving parts and potentially better performance according to [the perf book](https://nnethercote.github.io/perf-book/standard-library-types.html#box).

---

changelog: add [`box_default`] lint

### DIFF
--- a/clippy_lints/src/box_default.rs
+++ b/clippy_lints/src/box_default.rs
@@ -1,0 +1,61 @@
+use clippy_utils::{diagnostics::span_lint_and_help, is_default_equivalent, path_def_id};
+use rustc_hir::{Expr, ExprKind, QPath};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// checks for `Box::new(T::default())`, which is better written as
+    /// `Box::<T>::default()`.
+    ///
+    /// ### Why is this bad?
+    /// First, it's more complex, involving two calls instead of one.
+    /// Second, `Box::default()` can be faster
+    /// [in certain cases](https://nnethercote.github.io/perf-book/standard-library-types.html#box).
+    ///
+    /// ### Known problems
+    /// The lint may miss some cases (e.g. Box::new(String::from(""))).
+    /// On the other hand, it will trigger on cases where the `default`
+    /// code comes from a macro that does something different based on
+    /// e.g. target operating system.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let x: Box<String> = Box::new(Default::default());
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let x: Box<String> = Box::default();
+    /// ```
+    #[clippy::version = "1.65.0"]
+    pub BOX_DEFAULT,
+    perf,
+    "Using Box::new(T::default()) instead of Box::default()"
+}
+
+declare_lint_pass!(BoxDefault => [BOX_DEFAULT]);
+
+impl LateLintPass<'_> for BoxDefault {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_new, [arg]) = expr.kind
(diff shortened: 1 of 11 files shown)


clippy_lints/src/lib.rs
```diff
@@ -776,7 +777,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
             upper_case_acronyms_aggressive,
         ))
     });
-    store.register_late_pass(|| Box::new(default::Default::default()));
+    store.register_late_pass(|| Box::<default::Default>::default());
```
Alexendoo, line 780: At first this looked rather odd, interesting lint module + struct name we have there 😄

llogiq: Why, it's the default box for the Default lint in the default module. Totally clear, innit? :smile:

clippy_lints/src/box_default.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// ### Why is this bad?
+    /// First, it's more complex, involving two calls instead of one.
+    /// Second, `Box::default()` can be faster
+    /// [in certain cases](https://nnethercote.github.io/perf-book/standard-library-types.html#box).
+    ///
+    /// ### Known problems
+    /// The lint may miss some cases (e.g. Box::new(String::from(""))).
+    /// On the other hand, it will trigger on cases where the `default`
+    /// code comes from a macro that does something different based on
+    /// e.g. target operating system.
```
Alexendoo, line 21: Could add a check for `expr.span.ctxt() == arg.span.ctxt()` to avoid linting across macro boundaries while still catching it in local macro definitions

llogiq: Good idea. Done.

Alexendoo: Can remove the macro note from `Known problems` with that unless there's a case I'm missing

Alexendoo: Oh maybe things like `Box::new(cfg_if! ... { String::new() } else { ... })` where the expressions aren't coming from the macro

llogiq: Exactly. Though if you want, I can remove the note until I find a clearer wording.

clippy_lints/src/box_default.rs
```diff
@@ -0,0 +1,59 @@
...
+declare_lint_pass!(BoxDefault => [BOX_DEFAULT]);
+
+impl LateLintPass<'_> for BoxDefault {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_new, [arg]) = expr.kind
+            && let ExprKind::Path(ref box_new_qpath) = box_new.kind
+            && let ExprKind::Call(..) = arg.kind
+            && !in_external_macro(cx.sess(), expr.span)
+            && expr.span.eq_ctxt(arg.span)
+            && match_qpath(box_new_qpath, &paths::BOX_NEW)
```
Alexendoo, line 46: To avoid adding to `paths` you could do something like

```rust
&& let Some(def_id) = path_def_id(cx, box_new)
&& cx.tcx.parent(def_id) == cx.tcx.lang_items().require(LangItem::OwnedBox)
```

clippy_lints/src/box_default.rs
```diff
@@ -0,0 +1,59 @@
...
+            && is_default_equivalent(cx, arg)
+        {
+            span_lint_and_help(
+                cx,
+                BOX_DEFAULT,
+                expr.span,
+                "`Box::new(_)` of default value",
+                None,
+                 "use `Box::default()` instead",
```
Alexendoo, line 55:
```suggestion
                "use `Box::default()` instead",
```

clippy_lints/src/box_default.rs
Alexendoo: I'd say remove it, apparently you can't use `cfg_if` in an expression position, nor `#[cfg]` as attributes on expressions are unstable

clippy_lints/src/box_default.rs
llogiq: I'm afraid that doesn't work. I'll look into it.

llogiq:
Got it! r?

Alexendoo (approved):

Alexendoo:
Ahh right the parent of `Box::new` would probably be the `impl Box {}` not `Box` itself

@bors r+

PR #9523, opened 2022-09-23
Upgrade `compiletest-rs` dependency

smoelius:
From `0.8` to `0.9`.

The new version includes a [fix](https://github.com/Manishearth/compiletest-rs/pull/259) for what I suspect was one cause of the recent rustup failure: https://github.com/rust-lang/rust-clippy/actions/runs/3106438892/jobs/5033324694#step:11:911

changelog: none


Alexendoo (approved):

Alexendoo:
Thanks for looking into it!

@bors r+

PR #9533, opened 2022-09-25
Fix panic when displaying the backtrace of failing integration tests

Jarcho:
changelog: None

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -6,6 +6,11 @@ use std::env;
 use std::ffi::OsStr;
 use std::process::Command;
 
+#[cfg(not(windows))]
+const CARGO_CLIPPY: &str = "cargo-clippy";
+#[cfg(windows)]
+const CARGO_CLIPPY: &str = "cargo-clippy.exe";
+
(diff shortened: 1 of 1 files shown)


tests/integration.rs
```diff
@@ -51,17 +56,14 @@ fn integration_test() {
...
-            .rfind("error: internal compiler error")
+    if let Some(backtrace_start) = stderr.find("error: internal compiler error") {
+        let backtrace_end = stderr[backtrace_start..]
+            .find("end of query stack")
             .expect("end of backtrace not found");
 
         panic!(
             "internal compiler error\nBacktrace:\n\n{}",
-            &stderr[backtrace_start..backtrace_end]
+            &stderr[backtrace_start..backtrace_start + backtrace_end + 18]
```
llogiq, line 66: I think this could have done with a `static` and using its `.len()` here instead of a magic number, but this is a very minor niggle.

Jarcho: I really should have. I actually got the number wrong the first time I wrote it.

llogiq:
Thank you!

@dswij sorry for stealing the review. :grin:

@bors r+

PR #9541, opened 2022-09-26
Generate lint categories and explanations with `declare_clippy_lint`

Alexendoo:
This means contributors will no longer have to run `cargo dev update_lints` after changing a lints documentation or its category, which may also mean fewer merge conflicts in general

It works by swapping `declare_clippy_lint` out for a `proc_macro` of the same name. The proc macro emits a `LintInfo` alongside the generated `Lint` which are gathered into `declared_lint::LINTS`. The categories/explanations are then read from `declared_lint::LINTS` at runtime

The removal of `src/docs` is split into a separate commit to be more easily ignored

It is slightly slower though, adding a bit under a second to build time. Less noticeable in full builds or with a slower linker (benchmark uses mold)

```bash
hyperfine --warmup 2 \
    --parameter-list commit "declare-proc-macro,master" \
    --command-name "{commit}" \
    --setup "git checkout {commit}" \
    --prepare "touch clippy_lints/src/lib.rs" \
    "cargo build"
```
```
Benchmark 1: declare-proc-macro
  Time (mean ± σ):     10.731 s ±  0.154 s    [User: 7.739 s, System: 1.791 s]
  Range (min … max):   10.598 s … 11.125 s    10 runs

Benchmark 2: master
  Time (mean ± σ):      9.422 s ±  0.094 s    [User: 7.183 s, System: 1.732 s]
  Range (min … max):    9.287 s …  9.624 s    10 runs

Summary
  'master' ran
    1.14 ± 0.02 times faster than 'declare-proc-macro'
```

r? @flip1995
cc @llogiq for `--explain`

changelog: none


clippy_lints/src/lib.rs
```diff
@@ -491,31 +390,102 @@ pub fn read_conf(sess: &Session) -> Conf {
...
+    Restriction,
+    Style,
+    Suspicious,
+    #[cfg(feature = "internal")]
+    Internal,
+}
+
+pub(crate) struct LintInfo {
+    /// Double reference to maintain pointer equality
+    lint: &'static &'static Lint,
```
Alexendoo, line 409: Figuring this out was quite an ordeal, I'm not exactly sure what was going on with just `&'static Lint`

`static`s seem to deeply(?) copy other `static`s if they're in a separate module, even for non-`Copy` structs

```rust
// m.rs
pub static ONE: &crate::S = &crate::S {
    str: "number one",
};
```

```rust
// main.rs
mod m;

use m::ONE;

pub struct S {
    str: &'static str,
}

static TWO: &S = ONE;

fn main() {
    println!("{:p}", ONE);
    println!("{:p}", TWO);

    println!();

    // Even the str contents are copied
    println!("{:p}", ONE.str.as_ptr());
    println!("{:p}", TWO.str.as_ptr());
}
```

prints

```
0xaaaad4f50558
0xaaaad4f50510

0xaaaad4f45c3b
0xaaaad4f45c30
```

Alexendoo: I guess the answer is fairly simple actually, `SOME_STATIC` acts the same as a const, only `&SOME_STATIC` has a stable address

Alexendoo:
Added back the code block `#` removal functionality that I missed, I'll also leave this as conflicting when it happens next as it'll be a pain to keep on top of

llogiq:
I think reducing the code churn may be worth that second of build time. Well done, @Alexendoo!

clippy_lints/src/lib.rs
```diff
@@ -491,31 +390,102 @@ pub fn read_conf(sess: &Session) -> Conf {
...
+    let mut suspicious = Vec::new();
+    #[cfg(feature = "internal")]
+    let mut internal = Vec::new();
+
+    for LintInfo { lint, category, .. } in declared_lints::LINTS {
+        if matches!(category, Correctness | Suspicious | Style | Complexity | Perf) {
+            all.push(LintId::of(lint));
+        }
+
+        let group = match category {
```
flip1995, line 444: Can you implement this as methods on the `LintCategory` enum?

declare_clippy_lint/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "declare_clippy_lint"
+version = "0.1.65"
```
flip1995, line 3: Should be tested to match the top level `Cargo.toml` version. We have a test called ``matches_clippy_version`` or something similar, that already does that for other Clippy crates. Just add this crate there.

declare_clippy_lint/src/lib.rs
```diff
@@ -0,0 +1,170 @@
...
+///     /// ### Example
+///     /// ```rust
+///     /// Insert a short example of code that triggers the lint
+///     /// ```
+///     ///
+///     /// Use instead:
+///     /// ```rust
+///     /// Insert a short example of improved code that doesn't trigger the lint
+///     /// ```
+///     pub LINT_NAME,
```
flip1995, line 112:
```suggestion
///     #[clippy::version = "1.65.0"]
///     pub LINT_NAME,
```

Let's also document, that a version attribute should be added.

flip1995 (commented):
This is great! Thanks for doing this. I'm also not that concerned about the compile time of Clippy here. Not having so many files that can conflict between PRs is definitely worth it IMO.

declare_clippy_lint/Cargo.toml
flip1995: This is the test I meant and where it should be added: https://github.com/rust-lang/rust-clippy/blob/fe57ab7c6414bbef66097c972e613492b8d95065/tests/versioncheck.rs#L26-L31

Alexendoo:
Thanks! If any more conflicts crop up I'll keep on top of them

flip1995:
@bors r+ p=100

Thanks! (Going to bed. If this should fail for some reason, feel free to r=me again)

PR #9527, opened 2022-09-24
fallout2: rework clippy_dev & _lints fmt inlining

nyurik:
Continuing #9525 -- a few more inlining, but this time with some code changes to simplify format strings:

* Inline format args where possible
* simplify a few complex macros into format str
* use formatdoc!() instead format!(indoc!(...))

changelog: none

cc: @llogiq

PR #9525, opened 2022-09-23
pre-fallout: Apply uninlined_format-args lint

nyurik:
This change is needed for the uninlined_format-args lint to be merged. See https://github.com/rust-lang/rust-clippy/pull/9233

changelog: none

PR #9547, opened 2022-09-27
fallout: fix tests to allow uninlined_format_args

nyurik:
In order to switch `clippy::uninlined_format_args` from pedantic to style, all existing tests must not raise a warning. I did not want to change the actual tests, so this is a relatively minor change that:

* add `#![allow(clippy::uninlined_format_args)]` where needed
* normalizes all allow/deny/warn attributes
   * all allow attributes are grouped together
   * sorted alphabetically
   * the `clippy::*` attributes are listed separate from the other ones.
   * deny and warn attributes are listed before the allowed ones

See also https://github.com/rust-lang/rust-clippy/pull/9233, https://github.com/rust-lang/rust-clippy/pull/9525, https://github.com/rust-lang/rust-clippy/pull/9527

cc: @llogiq @Alexendoo

changelog: none

### DIFF
--- a/tests/ui-toml/conf_deprecated_key/conf_deprecated_key.rs
+++ b/tests/ui-toml/conf_deprecated_key/conf_deprecated_key.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::uninlined_format_args)]
+
 fn main() {}
 
 #[warn(clippy::cognitive_complexity)]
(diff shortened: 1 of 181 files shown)


flip1995 (changes requested):
This adds an allow attribute for the `uninlined_format_args` lint for **every** test file, even if it doesn't even use the `format!` macro. This is highly unnecessary and not something we want. Please only add this attribute to the tests that would actually fail due to the new lint.

nyurik:
> This adds an allow attribute for the `uninlined_format_args` lint for **every** test file, even if it doesn't even use the `format!` macro. This is highly unnecessary and not something we want. Please only add this attribute to the tests that would actually fail due to the new lint.

@flip1995 that is not correct - there are 801 test .rs files in the `tests/ui` dir, and I only changed 86 of them - only about 10%, all of which are required. I reevaluated all of them again, and did find one -- `tests/ui/empty_line_after_outer_attribute.rs` that I refactored by accident (reverted).  Just to keep things focused, I also undeed the `moveable -> movable` rename.  What's left are all needed.

nyurik:
@flip1995 is there still something that needs to be done to this PR? Thx!

flip1995 (approved):
Sorry, I was away over the weekend (long weekend in Germany).

All good now. Maybe I saw ghosts, but I thought I saw a file where it wasn't necessary. Thanks for addressing this!

Issue #9540, opened 2022-09-26
Incorrect fix for to_string_in_format_args


nyurik:
~I just discovered another bug in the format args parser `FormatArgsExpn::parse` while creating more tests for #9233.~

Add this test case to the [tests/ui/format_args.rs](https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/format_args.rs) `main()` to see it fail:

```rust
print!("error: something failed at {}", (Location::caller().to_string()));
//                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  (does not include parens)
```

~The resulting `format_args.args[0].param.value` has a span does NOT include the surrounding parenthesis, which means that it suggest this:~

Auto-fix is incorrectly suggesting this:

```
error: `to_string` applied to a type that implements `Display` in `print!` args
  --> $DIR/format_args.rs:86:64
   |
LL |     print!("error: something failed at {}", (Location::caller().to_string()));
   |                                                                ^^^^^^^^^^^^^ help: remove this
```

which results in an incompilabre

```rust
print!("error: something failed at {}", (Location::caller());  // <-- missing closing parenthesis
````

cc: @Alexendoo

Alexendoo:
This is unrelated to `format_args` parsing, the expression does include the parenthesis

nyurik:
@Alexendoo why do you think that?   In [these examples](https://github.com/rust-lang/rust-clippy/pull/9548/files#diff-d108878e5a893f2542507a02753d77b8cc6ee203ce0c0a72cc01ce267f7668ccR140-R143) of uninlined-format-args test that also fails, just like the `format-args` lint, the captured arg is clearly not including the surrounding parens - and that's why when it uses `expand_past_previous_comma()`, it is dis-balanced - on the left side it includes all the non-captured parens plus the comma, whereas on the right side it says where the parsing captured it, without the closing paren.

Alexendoo:
Haven't looked into that one but looks like a different issue. For this it's because the parenthesis are part of the expression's span, when it does https://github.com/rust-lang/rust-clippy/blob/5825ae7bd4b5dcb37b0bb7ccfaa68f57e9e8ee1b/clippy_lints/src/format_args.rs#L270 it's including the closing parenthesis in the thing it's suggesting to remove

nyurik:
Thanks @Alexendoo, you are right uninlined_format_args issue was due to a bug that I now have a PR for: #9588 

[...]

PR #9548, opened 2022-09-28
Add failing uninlined-format-args tests

nyurik:
All these tests are currently failing (generating uncompilable code).

* Implement a simplified `indoc!` proc macro
* added extra parens test `println!("{}", (local_i32))`
  * This bug is caused by https://github.com/rust-lang/rust-clippy/issues/9540
* added a `pass_through!(...)` macro test

The .fixed and .stderr were hacked without the `cargo dev bless`, so might be incorrect.

changelog: [`uninlined_format_args`]: fixed several edge case bugs

CC: @Alexendoo


nyurik:
these tests are no longer needed due to #9586

PR #9549, opened 2022-09-28
new `implicit_saturating_add` lint

### LINKED ISSUE #9393: Add lint `IMPLICIT_SATURATING_ADD`
### What it does

Add lint similar to `IMPLICIT_SATURATING_SUB`, but for `ADD`. So it will be `IMPLICIT_SATURATING_ADD`

### Lint Name

IMPLICIT_SATURATING_ADD

### Category

pedantic

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
#![warn(clippy::implicit_saturating_add)]

fn main() {
    let mut i: u32 = 3;

    if i != MAX {
        i += 1;
    }
    println!("{}", i);
}

```

Could be written as:

```rust
#![warn(clippy::implicit_saturating_add)]

fn main() {
    let mut i: u32 = 3;

    if i != MAX {
        i = i.saturating_add(1);
    }
    println!("{}", i);
}

```

royrustdev:
This fixes #9393 

- \[x] Run `cargo dev fmt`

---

changelog: add [`manual_saturating_add`] lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3917,6 +3917,7 @@ Released 2018-09-13
 [`implicit_clone`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_clone
 [`implicit_hasher`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_hasher
 [`implicit_return`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_return
+[`implicit_saturating_add`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_saturating_add
 [`implicit_saturating_sub`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_saturating_sub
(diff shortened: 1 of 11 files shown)


royrustdev:
r?llogiq

tests/ui/manual_saturating_add.rs
```diff
@@ -0,0 +1,136 @@
...
+    let mut b = 8;
+
+    if a < u8::MAX {
+        b += 1;
+    }
+
+    if u8::MAX > a {
+        b += 1;
+    }
+}
```
llogiq, line 136: I'd like to see tests against additional.statements in either then or else branch (noting that this led to false positives in the implicit-saturating-sub lint).

llogiq (commented):
This already looks pretty good. We best make sure to only lint when nothing is in the else branch of the if expression, otherwise this should be mergeable.

llogiq:
I think you'll need to rebase on current master before CI will run.

llogiq:
Thank you!

@bors r+

PR #9551, opened 2022-09-28
[`unnecessary_lazy_evaluations`] Do not suggest switching to early evaluation when type has custom `Drop`

### LINKED ISSUE #9427: `unnecessary_lazy_evaluations`: Shouldn't warn about values with significant `Drop` impl
### Summary

The case in question that was wrongly reported is the following

```
warning: unnecessary closure used with `bool::then`
   --> glib/src/convert.rs:122:9
    |
122 |         (iconv as isize != -1).then(|| Self(iconv))
    |         ^^^^^^^^^^^^^^^^^^^^^^^--------------------
    |                                |
    |                                help: use `then_some(..)` instead: `then_some(Self(iconv))`
    |
    = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
```

`Self` here has a significant `Drop` impl, specifically it will close the `isize` there (it's basically like a `fd` and closing `-1` is UB).

### Lint Name

unnecessary_lazy_evaluations

### Reproducer

I tried this code:

```rust
struct Foo(i32);

impl Drop for Foo {
    fn drop(&mut self) {
        println!("{}", self.0);
    }
}

fn main() {
    (0 == 1).then(|| Foo(0));
}
```

I saw this happen:

```
warning: unnecessary closure used with `bool::then`
  --> src/main.rs:10:5
   |
10 |     (0 == 1).then(|| Foo(0));
   |     ^^^^^^^^^---------------
   |              |
   |              help: use `then_some(..)` instead: `then_some(Foo(0))`
   |
   = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
```

I expected to see this happen: nothing because the suggestion gives different behaviour

### Version

```text
rustc 1.65.0-nightly (289279de1 2022-09-04)
binary: rustc
commit-hash: 289279de116707f28cf9c18e4bbb8c6ec84ad75b
commit-date: 2022-09-04
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

kraktus:
fix #9427

changelog: [`unnecessary_lazy_evaluations`] Do not suggest switching to early evaluation when type has custom `Drop`

### DIFF
--- a/clippy_utils/src/eager_or_lazy.rs
+++ b/clippy_utils/src/eager_or_lazy.rs
@@ -113,7 +113,17 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
                     },
                     args,
                 ) => match self.cx.qpath_res(path, hir_id) {
-                    Res::Def(DefKind::Ctor(..) | DefKind::Variant, _) | Res::SelfCtor(_) => (),
+                    Res::Def(DefKind::Ctor(..) | DefKind::Variant, _) | Res::SelfCtor(_) => {
+                        if self
+                            .cx
+                            .typeck_results()
+                            .expr_ty(e)
+                            .has_significant_drop(self.cx.tcx, self.cx.param_env)
+                        {
+                            self.eagerness = Lazy;
+                            return;
+                        }
+                    },
                     Res::Def(_, id) if self.cx.tcx.is_promotable_const_fn(id) => (),
                     // No need to walk the arguments here, `is_const_evaluatable` already did
                     Res::Def(..) if is_const_evaluatable(self.cx, e) => {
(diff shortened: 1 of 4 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the fix :upside_down_face: 

@bors r+

Jules-Bertholet:
As mentioned in https://github.com/rust-lang/rust-clippy/issues/9427#issuecomment-1295742590, this PR did not fully fix the issue.

PR #9559, opened 2022-09-30
Fix #9544

### LINKED ISSUE #9544: `arithmetic_side_effects` misses `T / integral`, `T * float` and similar
### Summary

`has_valid_op` doesn't check that the other side of the operation is also integral. There's a similar issue for types present in `HARD_CODED_ALLOWED`

cc @c410-f3r

### Lint Name

arithmetic_side_effects

### Reproducer

The following panics at runtime but isn't linted [(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=03993e2273a32c9b0e8f6e0e152ea5ae)

```rust
#![warn(clippy::arithmetic_side_effects)]

fn main() {
    let c = num_rational::Ratio::new(1u8, 128u8);
    let _ = c / 2;
}
```

For the `HARD_CODED_ALLOWED` side [(playground)](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b25ce350ca18191f5c7793d1ac20a4ad)

```rust
#![warn(clippy::arithmetic_side_effects)]

fn main() {
    let c = ordered_float::NotNan::new(0f64).unwrap();
    let _ = c * f64::INFINITY;
}
```

### Version

_No response_

c410-f3r:
Fix #9544

r? @Alexendoo 

changelog: [`arithmetic_side_effects`]: Fix false negative for `CustomType / usize`, `CustomType * float` and similar

### DIFF
--- a/clippy_lints/src/operators/arithmetic_side_effects.rs
+++ b/clippy_lints/src/operators/arithmetic_side_effects.rs
@@ -16,8 +16,9 @@ const HARD_CODED_ALLOWED: &[&str] = &[
     "f32",
     "f64",
     "std::num::Saturating",
-    "std::string::String",
     "std::num::Wrapping",
+    "std::string::String",
+    "&str",
 ];
 
 #[derive(Debug)]
@@ -77,6 +78,11 @@ impl ArithmeticSideEffects {
         )
     }
 
+    // For example, 8i32 or &i64::MAX.
+    fn is_integral<'expr, 'tcx>(cx: &LateContext<'tcx>, expr: &'expr hir::Expr<'tcx>) -> bool {
+        cx.typeck_results().expr_ty(expr).peel_refs().is_integral()
+    }
+
     // Common entry-point to avoid code duplication.
     fn issue_lint(&mut self, cx: &LateContext<'_>, expr: &hir::Expr<'_>) {
         let msg = "arithmetic operation that can potentially result in unexpected side-effects";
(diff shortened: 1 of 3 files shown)


clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -134,14 +129,18 @@ impl ArithmeticSideEffects {
         ) {
             return;
         };
-        if self.is_allowed_ty(cx, lhs) || self.is_allowed_ty(cx, rhs) {
+        if self.is_allowed_ty(cx, lhs) && self.is_allowed_ty(cx, rhs) {
```
Alexendoo, line 132: That does leave us in the position where e.g. `Complex<usize> * usize` requires allowing `Complex` and `usize`, which would also allow all `usize` operations. Not exactly sure what the best thing to do about that is

c410-f3r: Well, it is not possible to conclude anything in regards to custom types or how they interactive with other things, therefore, it seems plausible to `&&` instead of `II`.

The matter of allowing heterogeneous combinations could be addressed in a future PR asking for users to specify pairs instead of single values in `arithmetic-side-effects-allowed`.

```
arithmetic-side-effects-allowed  = [
  "Complex<usize>,usize",
  "Complex<f64>,*",
  "*,Complex<f64>"
]
```

Alexendoo: Specifying pairs seems reasonable, in that case I'd suggest adding a check that the `lhs` and `rhs` types are the same for now

Alexendoo: Yeah just until the pairs thing is implemented

c410-f3r: Done as requested

Alexendoo (approved):

Alexendoo:
Nice, thanks!

@bors r+

Alexendoo:
Stuck a `changelog:` in the body,

@bors retry

c410-f3r:
Can you r? again?

c410-f3r:
Thanks

PR #9566, opened 2022-10-01
Expand internal lint `unnecessary_def_path`

smoelius:
This PR does essentially two things:
* Separates the internal lints into modules by pass. (`internal_lints.rs` was over 1400 lines, which is a little unruly IMHO.)
* ~Adds a new~ Expands the `unnecessary_def_path` internal lint to flag hardcoded paths to diagnostic and language items.

My understanding is that the latter is currently done by reviewers. Automating this process should make things easier for both reviewers and contributors.

I could make the first bullet a separate PR, or remove it entirely, if desired.

changelog: Add internal lint `diagnostic_item_path`


Jarcho:
There is a lint for this, it's just currently broken. There is a pr to fix said lint (#7962)

smoelius:
> There is a lint for this, it's just currently broken. There is a pr to fix said lint (#7962).

Perhaps there is room for both approaches?

If I understand correctly, `unnecessary_def_path` flags the use of a path. The (much less sophisticated) idea I am proposing is to flag the path itself.

I would expect cases like this to be hard for the former, but catchable by the latter: https://github.com/rust-lang/rust-clippy/pull/9566/files#diff-99cb6bebcf4ed3a60e0e046a65b7a39631a9ce18589a91e9e668c0e0fd6cd9abL102-L105

That said, it would certainly be better that the latter not fire when the former does.

So perhaps we could push #7962 through and then look to merge them?

What is needed to push #7962 through?

Jarcho:
Ahh, that is a little different. I think it would be better to be part of `unnecessary_def_path `, but otherwise seems ok.

smoelius:
I think this is ready for review, @dswij (cc: @Jarcho).

Instead of adding a new lint, the changes now expand `unnecessary_def_path`. Also, none of the new warnings are emitted for a path if any of the lint's existing warnings involve that path.

The last commit is just formatting, and could be omitted.

clippy_lints/src/utils/internal_lints/unnecessary_def_path.rs
```diff
@@ -0,0 +1,343 @@
...
+
+            None
+        })
+        .collect()
+}
+
+// def_path_res will match field names before anything else, but for this we want to match
+// inherent functions first.
+fn inherent_def_path_res(cx: &LateContext<'_>, segments: &[&str]) -> Option<DefId> {
+    def_path_res(cx, segments, None).opt_def_id().map(|def_id| {
```
Jarcho, line 299: Unrelated to this PR. I think this whole function can be replaced with with `def_path_res(..., Some(Namespace::ValueNS))`.

smoelius: Just to be clear, is this what you are suggesting? https://github.com/smoelius/rust-clippy/commit/6a1ad0b7e70c1b9fe0d3e9146fc2ae55bbec97f8

The tests don't seem to pass for me locally. (I haven't dug into why, though.)

Jarcho: Looks like you can ignore this. I thought the namespace parameter was a hint rather than a constraint for some reason.

dswij (approved):
LGTM, thanks for this! Can you help to rebase? I think this is good to merge afterward.

smoelius:
The best way I know to verify that a rebase was done correctly is to diff the old patch against the new. So here is a diff of patches 2c8e473:bb72e43 and 50f192f:5dc54c6 (the interesting lines are the `^[<>] [+-]` ones):
```patch
217c217
< index 05d927dbe..19935bd4d 100644
---
> index de1253c85..5609a4dc9 100644
275c275
< index 2dcefd787..da33ad299 100644
---
> index ebb0f14fe..893410dbf 100644
278c278
< @@ -529,17 +529,23 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
---
> @@ -528,17 +528,23 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
503c503
< index cfcf9596c..265ac7bba 100644
---
> index 78c1b33ed..fb92779be 100644
560,563c560,562
< -            let trait_def_id = match get_trait_def_id(cx, trait_path) {
< +            let trait_def_id = match cx.tcx.get_diagnostic_item(trait_sym) {
<                  Some(did) => did,
<                  None => return false,
---
> -            let Some(trait_def_id) = get_trait_def_id(cx, trait_path) else {
> +            let Some(trait_def_id) = cx.tcx.get_diagnostic_item(trait_sym) else {
>                  return false
564a564
>              implements_trait(cx, ty, trait_def_id, &[parent_ty.into()])
802c802
< index 85bcbc7ad..71f6c9909 100644
---
> index 0d908bf2a..71f6c9909 100644
805c805
< @@ -1,1566 +1,12 @@
---
> @@ -1,1560 +1,12 @@
2217,2220c2217
< -        let then_block = match then.kind {
< -            ExprKind::Block(block, _) => block,
< -            _ => return,
< -        };
---
> -        let ExprKind::Block(then_block, _) = then.kind else { return };
2225,2228c2222
< -        let if_chain_span = match if_chain_span {
< -            None => return,
< -            Some(span) => span,
< -        };
---
> -        let Some(if_chain_span) = if_chain_span else { return };
2774c2768
< index 000000000..81cfd1f7e
---
> index 000000000..883a5c08e
2777c2771
< @@ -0,0 +1,170 @@
---
> @@ -0,0 +1,164 @@
2826,2829c2820
< +        let then_block = match then.kind {
< +            ExprKind::Block(block, _) => block,
< +            _ => return,
< +        };
---
> +        let ExprKind::Block(then_block, _) = then.kind else { return };
2834,2837c2825
< +        let if_chain_span = match if_chain_span {
< +            None => return,
< +            Some(span) => span,
< +        };
---
> +        let Some(if_chain_span) = if_chain_span else { return };
4211c4199
< index e6492d762..f9a9c89de 100644
---
> index dbe75b43c..cbc2fde69 100644
4214c4202
< @@ -1763,6 +1763,7 @@ pub fn any_parent_is_automatically_derived(tcx: TyCtxt<'_>, node: HirId) -> bool
---
> @@ -1766,6 +1766,7 @@ pub fn any_parent_is_automatically_derived(tcx: TyCtxt<'_>, node: HirId) -> bool
4222c4210
< @@ -1780,6 +1781,22 @@ pub fn match_function_call<'tcx>(
---
> @@ -1783,6 +1784,22 @@ pub fn match_function_call<'tcx>(
4246c4234
< index 13938645f..f54d7e220 100644
---
> index 78adc4536..bc8514734 100644
4329c4317,4318
< @@ -131,14 +107,8 @@ pub const REGEX_BYTES_SET_NEW: [&str; 5] = ["regex", "re_set", "bytes", "RegexSe
---
> @@ -125,14 +101,8 @@ pub const REGEX_BYTES_NEW: [&str; 4] = ["regex", "re_bytes", "Regex", "new"];
>  pub const REGEX_BYTES_SET_NEW: [&str; 5] = ["regex", "re_set", "bytes", "RegexSet", "new"];
4331d4319
<  #[expect(clippy::invalid_paths)] // internal lints do not know about all external crates
```
The changes are to account for these:
https://github.com/rust-lang/rust-clippy/pull/9629/files#diff-7f73f6fe28c04b654223c09c42fe02937d2351fc58a91d21ab812aaf6f9b185dR3854-R3855
https://github.com/rust-lang/rust-clippy/pull/9629/files#diff-132b8beb349994a7a16a8740b14a5cb5b1402f746d3c2a75295c2079cad6b70aR1405-R1410

Thank you very much for the review.

dswij:
Thank you for the changes and the cleanup! @bors r+

PR #9569, opened 2022-10-01
use rustc_tools_util dependency from crates.io instead of this repo.

### LINKED ISSUE #9553: Status of rustc_tools_util?
vergen is causing some licensing pain in Miri, so I checked what clippy does to determine the git commit for version information. Looks like clippy has an in-tree crate `rustc_tools_util` for that. An old version of that crate exists on crates.io but clippy uses a path dependency of a more up-to-date version of that crate.

Would it make sense for clippy and Miri to share that crate? clippy should then probably fetch the dependency from crates.io rather than using a path dependency, so that the dependency is actually shared in the rustc workspace.

matthiaskrgr:
Fixes #9553

changelog: none

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ path = "src/driver.rs"
 [dependencies]
 clippy_lints = { path = "clippy_lints" }
 semver = "1.0"
-rustc_tools_util = { path = "rustc_tools_util" }
+rustc_tools_util = "0.2.1"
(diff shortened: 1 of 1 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update. :upside_down_face: 

(I've removed this from the chaneglog as it's probably not too interesting to end users) 

@bors r+

PR #9570, opened 2022-10-01
feat: lint unchecked subtraction of a 'Duration' from an 'Instant'

### LINKED ISSUE #9371: Warn against subtracting a constant Duration from Instant::now()
### What it does

I'm not sure how large the scope of this lint should be, so here's the most conservative approach first:

This lint finds patterns of the form `std::time::Instant::now() - std::time::Duration::from_secs(...)` (or from_millis, from_secs_f32, whatever) and warns against potential underflow panics on certain platforms. Maybe there should also be a check that the duration exceeds a certain threshold (e.g. subtracting 50 milliseconds will pretty much always work, but 50 seconds not necessarily)

The warning message would say something like `This subtraction can panic, depending on OS and system bootup time. Rewrite your code to not require potentially negative Instants, or make the intention to panic explicit with '.checked_sub(...).unwrap()'`

The lint could be expanded to lint for all Durations (not just inline-constructed constant Durations), and/or to lint for all Instants (not just Instant::now()). But I'm not sure if the increased false positive rate is worth it

### Lint Name

potential-instant-underflow

### Category

suspicious

### Advantage

There are no hidden potential panics

### Drawbacks

If the system is _known_ to have been booted for the specified Duration, this lint is a false positive. Example `thread::sleep(Duration::from_secs(60)); Instant::now() - Duration::from_secs(60);`

### Example

```rust
let delete_before = Instant::now() - Duration::from_secs(60 * 60);
self.entries.retain(|entry| entry.timestamp > delete_before);
```

Could be written as:

```rust
let max_age = Duration::from_secs(60 * 60);
self.entries.retain(|entry| Instant::now() - entry.timestamp < max_age);
```

nfejzic:
Hello all, I tried to tackle the open issue #9371 and this is what I came up with. 

I have a difficulty currently - some tests are failing: 

```
failures:
    [ui] ui/manual_instant_elapsed.rs
```

The `manual_instant_elapsed` is failing because of `Instant::now() - duration` test, this now gets also picked by `unchecked_duration_subtraction` lint. 
What is the correct way to proceed in this case? Simply update the `.stderr` file for `manual_instant_elapsed` lint? 

changelog: [`unchecked_duration_subtraction`]: Add lint for unchecked subtraction of a `Duration` from an `Instant`.

fixes #9371


llogiq:
I think the best way would be to quell linting if the manual-instant-elapsed lint already picks up the expr. This might mean merging the two lint's passes to avoid duplicating the code.

clippy_lints/src/unchecked_duration_subtraction.rs
```diff
@@ -0,0 +1,105 @@
...
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Finds patterns of unchecked subtraction of [`Duration`] from [`Instant::now()`].
+    ///
+    /// ### Why is this bad?
+    /// Unchecked subtraction could cause underflow on certain platforms, leading to bugs and/or
+    /// unintentional panics.
```
Alexendoo, line 15: It currently always panics if the subtraction overflows - https://doc.rust-lang.org/1.64.0/src/std/time.rs.html#420-426

nfejzic:
> I think the best way would be to quell linting if the manual-instant-elapsed lint already picks up the expr. This might mean merging the two lint's passes to avoid duplicating the code.

As far as I can see, the `manual_instant_elapsed` lint does not catch expression `Instant::now() - duration`.

In the [test file for `manual_instant_elapse`](https://github.com/nfejzic/rust-clippy/blob/9a1813f72dc38833f9100ea3bf9d7074cc3ff1d0/tests/ui/manual_instant_elapsed.rs#L9) there is this snippet: 
```rust
// don't catch
let duration = prev_instant.elapsed();

Instant::now() - duration;
```

But `manual_instant_elapsed` does not produce a lint for the given snippet in [stderr output file](https://github.com/nfejzic/rust-clippy/blob/9a1813f72dc38833f9100ea3bf9d7074cc3ff1d0/tests/ui/manual_instant_elapsed.stderr). However, the new `unchecked_duration_subtraction` does, and that produces a different `stderr` file, hence the test failing.

The `manual_instant_elapsed` lint catches subtraction of two `Instant`s, but not `Instant` and `Duration` subtraction. One way to __fix__ the test, would be to `#![allow(clippy::unchecked_duration_subtraction)]` together with other allowed lints. 

Either way, I probably can merge the two lints. If we decide to do that, should we rename the (resulting) lint?

llogiq:
No, we should keep the two lints, but we may be able to merge the lint passes, because if I understand correctly, one lint is basically a special case of the other or there is at least some overlap.

nfejzic:
> No, we should keep the two lints, but we may be able to merge the lint passes, because if I understand correctly, one lint is basically a special case of the other or there is at least some overlap.

I'm not sure about the overlap. Basically, the `unchecked_duration_subtraction` detects the duration subtraction in the test file:
```rust
Instant::now() - duration;
```

However, that same line is **NOT** detected by the `manual_instant_elapsed` lint.

Do you have an example of merging two lint passes? Would like to see how it works.

llogiq:
Ah, I see, the manual elapsed lint checks for Instant - Instant, this checks for Instant - Duration.

Merging the LintPass is simply putting the relevant code into one pass. One usually starts by copying the lint definition + `check_*` code from one file to the other and then refactoring common expressions.

nfejzic:
> Merging the LintPass is simply putting the relevant code into one pass. One usually starts by copying the lint definition + `check_*` code from one file to the other and then refactoring common expressions.

I assume the file should be also renamed then? 

I will start working on that and see how it goes. 

Also, I think the `manual_instant_elapsed` checks only against `Instant::now()`. 
What is your opinion on this? Should I make it so that `unchecked_duration_subtraction` lints only against `Instant::now() - duration`, or should `manual_elapsed_lint` also check against any `Instant`, not only `Instant::now()` expressions?

nfejzic:
@llogiq I found some time and managed to merge the `manual_instant_elapsed` and `unchecked_duration_subtraction` lint passes into a single pass. Hopefully I've done this right.

llogiq:
We don't merge from upstream, we rebasse onto it. I usually do this on the command line (`git fetch master && git rebase master`, `git mergetool` if necessary), but most git UI tools also have a workflow for this.

nfejzic:
@llogiq How does it look now? I gave my best 👀

llogiq:
Now for some reason CI seems to have excused itself. Don't worry, I'll look into it.

nfejzic:
I just noticed that `clippy_version` for `MANUAL_INSTANT_ELAPSED` was changed some time ago to `1.65.0`... I should probably push a commit for that, right?

llogiq:
Yes, that makes sense.

nfejzic:
@llogiq I think this is ready for review now.

llogiq:
Thank you!

@bors r+

PR #9571, opened 2022-10-01
use `is_integer_literal` more

llogiq:
I noticed that we have the `is_integer_literal` function in our `clippy_utils`, yet almost everywhere people still match int literal expressions manually. So I searched for instances to replace and shorten the code a bit.

---

changelog: none


Jarcho:
Some of those really shrink down quite nicely. Thank you. @bors r+

PR #9572, opened 2022-10-01
Add `as_ptr_cast_mut` lint

Nilstrieb:
This lint detects calls to a `&self`-taking `as_ptr` method, where the result is then immediately cast to a `*mut T`. Code like this is probably invalid, as that pointer will not have write permissions, and `*mut T` is usually used to write through.

Examples of broken code with this pattern:
https://miri.saethlin.dev/ub?crate=lol_alloc&version=0.1.3
https://miri.saethlin.dev/ub?crate=sophon-wasm&version=0.19.0
https://miri.saethlin.dev/ub?crate=polars-core&version=0.24.2
https://miri.saethlin.dev/ub?crate=ach-cell&version=0.1.17

changelog: Add [`as_ptr_cast_mut`]

### DIFF
--- a/clippy_lints/src/casts/as_ptr_cast_mut.rs
+++ b/clippy_lints/src/casts/as_ptr_cast_mut.rs
@@ -1,0 +1,38 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::source::snippet_opt;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_middle::{
+    mir::Mutability,
+    ty::{self, Ty, TypeAndMut},
+};
+
+use super::AS_PTR_CAST_MUT;
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, cast_expr: &Expr<'_>, cast_to: Ty<'_>) {
+    if let ty::RawPtr(ptrty @ TypeAndMut { mutbl: Mutability::Mut, .. }) = cast_to.kind()
+        && let ty::RawPtr(TypeAndMut { mutbl: Mutability::Not, .. }) =
+            cx.typeck_results().node_type(cast_expr.hir_id).kind()
+        && let ExprKind::MethodCall(method_name, receiver, [], _) = cast_expr.peel_blocks().kind
+        && method_name.ident.name == rustc_span::sym::as_ptr
+        && let Some(as_ptr_did) = cx.typeck_results().type_dependent_def_id(cast_expr.peel_blocks().hir_id)
+        && let as_ptr_sig = cx.tcx.fn_sig(as_ptr_did)
+        && let Some(first_param_ty) = as_ptr_sig.skip_binder().inputs().iter().next()
+        && let ty::Ref(_, _, Mutability::Not) = first_param_ty.kind()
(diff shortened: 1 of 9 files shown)


clippy_lints/src/casts/mod.rs
```diff
@@ -596,6 +597,32 @@ declare_clippy_lint! {
...
+    /// Use instead:
+    /// ```rust
+    /// let mut string = String::with_capacity(1);
+    /// let string = string.as_mut_ptr();
+    /// unsafe { ptr.write(4) };
+    /// ```
+    #[clippy::version = "1.66.0"]
+    pub AS_PTR_CAST_MUT,
+    nursery,
+    "casting the result of the `&self`-taking as_ptr to a mutabe point"
```
saethlin, line 623:
```suggestion
    "casting the result of the `&self`-taking as_ptr to a mutable pointer"
```

clippy_lints/src/casts/mod.rs
```diff
@@ -596,6 +597,32 @@ declare_clippy_lint! {
     "casting a slice created from a pointer and length to a slice pointer"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for the result of a `&self`-taking `as_ptr` being cast to a mutable pointer
+    ///
+    /// ### Why is this bad?
+    /// Since `as_ptr` took a `&self`, the pointer won't have write permissions, making it
```
saethlin, line 605: _Won't_ is a bit too strong. This is not a dangerous pattern if the pointer points into an `UnsafeCell`.

I don't know how technically correct or detailed clippy docs are expected to be, but this should not be so absolute.

Nilstrieb: I've added a `unless interior mutability is used` which makes this correct, but it doesn't sound great in my opinion.

saethlin: Perhaps
```
    /// Since `as_ptr` takes a `&self`, the pointer won't have write permissions unless interior
    /// mutability is used. If this cast to `*mut` exists to enable mutation through this pointer,
    /// that mutation is likely to be UB.
```

dswij (approved):
This LGTM. I think the wording is ok now. If the wording is indeed too confusing, we can improve it in the future.

clippy_lints/src/casts/as_ptr_cast_mut.rs
```diff
@@ -0,0 +1,38 @@
...
+        && let ExprKind::MethodCall(method_name, receiver, [], _) = cast_expr.peel_blocks().kind
+        && method_name.ident.name == rustc_span::sym::as_ptr
+        && let Some(as_ptr_did) = cx.typeck_results().type_dependent_def_id(cast_expr.peel_blocks().hir_id)
+        && let as_ptr_sig = cx.tcx.fn_sig(as_ptr_did)
+        && let Some(first_param_ty) = as_ptr_sig.skip_binder().inputs().iter().next()
+        && let ty::Ref(_, _, Mutability::Not) = first_param_ty.kind()
+        && let Some(recv) = snippet_opt(cx, receiver.span)
+    {
+        // `as_mut_ptr` might not exist
+        let applicability = Applicability::MaybeIncorrect;
```
dswij, line 26: I think we can check if it implements `as_mut_ptr` for the suggestion, but this can be a future improvement

dswij:
Thanks for this @Nilstrieb!

@bors r+

PR #9573, opened 2022-10-01
lint nested patterns and slice patterns in `needless_borrowed_reference`

Alexendoo:
Now lints in positions other than top level, e.g. `Some(&ref a)`. Or patterns are excluded as that can cause issues

Slice patterns of `ref`s are now linted, e.g. `&[ref a, ref b]`. An easy one to walk into as you might expect that to match against a slice you should use `&[]`,  then to get around a `cannot move out of type [T]` error you apply a `ref`

changelog: [`needless_borrowed_reference`]: lint nested patterns and slice patterns


dswij (approved):
LGTM, thanks for this!

PR #9574, opened 2022-10-01
Remove unused `.fixed` files, only run asm_syntax doctests on x86

Alexendoo:
Two small changes, removes some unused `.fixed` and makes `clippy_lints` doctests pass on non x86 arches

changelog: none


Jarcho:
Good catch. @bors r+

PR #9576, opened 2022-10-02
let unnecessary_cast work for trivial non_literal expressions

### LINKED ISSUE #9562: clippy::unnecessary_cast does not warn on basic usage
### Summary

When casting the result of a function to the same type, `clippy::unnecessary_cast` doesn't warn

### Lint Name

clippy::unnecessary_cast

### Reproducer

I tried this code:

```rust
fn foo() -> f32 { 0. }

fn main() {
    let num = foo() as f32;
}
```

I expected to see this happen: Warning that the cast is unnecessary

Instead, this happened: No warning

### Version

_No response_

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

---

changelog: [`unnecessary_cast`]: fix for trivial non_literal expressions

Fixes #9562

### DIFF
--- a/clippy_lints/src/casts/unnecessary_cast.rs
+++ b/clippy_lints/src/casts/unnecessary_cast.rs
@@ -34,7 +34,7 @@ pub(super) fn check<'tcx>(
     let cast_str = snippet_opt(cx, cast_expr.span).unwrap_or_default();
 
     if let Some(lit) = get_numeric_literal(cast_expr) {
-        let literal_str = cast_str;
+        let literal_str = &cast_str;
 
         if_chain! {
             if let LitKind::Int(n, _) = lit.node;
@@ -52,10 +52,12 @@ pub(super) fn check<'tcx>(
 
         match lit.node {
             LitKind::Int(_, LitIntType::Unsuffixed) if cast_to.is_integral() => {
-                lint_unnecessary_cast(cx, expr, &literal_str, cast_from, cast_to);
+                lint_unnecessary_cast(cx, expr, literal_str, cast_from, cast_to);
+                return true;
             },
             LitKind::Float(_, LitFloatType::Unsuffixed) if cast_to.is_floating_point() => {
-                lint_unnecessary_cast(cx, expr, &literal_str, cast_from, cast_to);
+                lint_unnecessary_cast(cx, expr, literal_str, cast_from, cast_to);
+                return true;
             },
             LitKind::Int(_, LitIntType::Unsuffixed) | LitKind::Float(_, LitFloatType::Unsuffixed) => {},
             LitKind::Int(_, LitIntType::Signed(_) | LitIntType::Unsigned(_))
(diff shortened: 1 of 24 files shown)


TennyZhuang:
Some other tests were broken, convert to draft temporally :(

TennyZhuang:
I've added some rules on some test files, but I'm not sure why they doesn't work.

 [783ee63](https://github.com/rust-lang/rust-clippy/pull/9576/commits/783ee635559a522a7a46b95419ccc536c6a28f97)

kraktus:
Those tests need to be blessed (because we also check the code output after `clippy --fix`, which is stored as `my_test.fixed`.

You can bless them with `cargo dev bless` and commit the changes (which normally should just be adding the `allow` parameter to the corresponding `.fixed`)

clippy_lints/src/casts/unnecessary_cast.rs
```diff
@@ -30,8 +30,10 @@ pub(super) fn check<'tcx>(
         }
     }
 
+    let cast_str = snippet_opt(cx, cast_expr.span).unwrap_or_default();
```
kraktus, line 33: nit, use [clippy_utils](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/index.html)::[source](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/source/index.html)::[snippet](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/source/fn.snippet.html#) instead (with canonical fallback being `..`, as shown in examples

clippy_lints/src/casts/unnecessary_cast.rs
```diff
@@ -80,6 +82,17 @@ pub(super) fn check<'tcx>(
...
+        span_lint_and_sugg(
+            cx,
+            UNNECESSARY_CAST,
+            expr.span,
+            &format!("casting to the same type is unnecessary (`{cast_from}` -> `{cast_to}`)"),
+            "try",
+            cast_str,
+            Applicability::MachineApplicable,
+        );
+        return true;
```
kraktus, line 95: Looks like it should already be caught by https://github.com/rust-lang/rust-clippy/blob/783ee635559a522a7a46b95419ccc536c6a28f97/clippy_lints/src/casts/unnecessary_cast.rs#L71
no?

TennyZhuang:
@kraktus Hi, in fact, I blessed them using `cargo dev bless` but found the `allow` does not work. You can run them locally or check [the CI log](https://github.com/rust-lang/rust-clippy/actions/runs/3168136112/jobs/5159119802#step:7:976). The warnings are still produced so I don't push them.

I found that if I use the function level `#[allow(clippy::unnecessary_cast)]`, then everything works. I guess there are another bug but I will use it as a workaround.

clippy_lints/src/casts/unnecessary_cast.rs
TennyZhuang: It's under a condition so it will only work for numeric literal.

kraktus: right, you could refactor it by turning your `if else` into a `if` and remove the afore mentioned line then I think

TennyZhuang:
Sorry, I misunderstood and forget to run `cargo uitest` before `cargo dev bless` :(

kraktus:
No worries!

clippy_lints/src/casts/unnecessary_cast.rs
TennyZhuang: I've tried it when I implement it, but the control flow is too complicated to do that. At least, I will try to extract a common function `lint_same_type`.

TennyZhuang: Updated

clippy_lints/src/casts/unnecessary_cast.rs
TennyZhuang: The original code uses `snippet_opt,` and the only thing I do is move the variable; there are also some other `snippet_opt` usages in the file, so I'm not sure if I should update them all in the PR. It seems a little off-topic.

llogiq:
Looks good to me.

Thank you!

@bors r+

PR #9584, opened 2022-10-03
add tests in `implicit_saturating_sub` lint

royrustdev:
This adds more tests to the `implicit_saturating_sub` lint to rule out certain false positives that have appeared in the past.

Now with those false positives out of the equation, we can move the lint to `style`.

---

changelog: promote [`implicit-saturating-sub`] to the `style` category

### DIFF
--- a/tests/ui/implicit_saturating_sub.rs
+++ b/tests/ui/implicit_saturating_sub.rs
@@ -2,6 +2,21 @@
 #![allow(unused_assignments, unused_mut, clippy::assign_op_pattern)]
 #![warn(clippy::implicit_saturating_sub)]
 
+use std::cmp::PartialEq;
+use std::ops::SubAssign;
+// Mock type
+struct Mock;
+
+impl PartialEq<u32> for Mock {
+    fn eq(&self, _: &u32) -> bool {
+        true
+    }
+}
+
+impl SubAssign<u32> for Mock {
+    fn sub_assign(&mut self, _: u32) {}
+}
+
 fn main() {
     // Tests for unsigned integers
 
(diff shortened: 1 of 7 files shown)


tests/ui/implicit_saturating_sub.rs
```diff
@@ -1,7 +1,22 @@
 // run-rustfix
-#![allow(unused_assignments, unused_mut, clippy::assign_op_pattern)]
+#![allow(unused_assignments, unused_variables, unused_mut, clippy::assign_op_pattern)]
```
llogiq, line 2: Is this really needed?

tests/ui/implicit_saturating_sub.rs
```diff
@@ -211,4 +226,39 @@ fn main() {
...
+    let mut m = Mock;
+    if m != 0 {
+        m -= 1;
+    }
+
+    let a = 12;
+    let mut b = 8;
+    let mut u_32 = 300;
+
+    if a < u8::MAX {
```
llogiq, line 240: The `implicit_sub_assign` lint should catch comparisons to zero, not MAX.

royrustdev: Done. Comparing with `0`

tests/ui/implicit_saturating_sub.rs
royrustdev: Facing wierd error without it as `b` is declared but never used, but `b` is certainly in used.

llogiq: In this case, renaming  the variable to start with an underscore (e.g. `_b`) is the canonical way to get rid of the error.

llogiq:
Just that one small nit about the variable name, then we can merge this.

royrustdev:
@llogiq your suggestions are implemented. 

Please take a look. 
Thanks

llogiq (commented):
Thank you for seeing this through.

@bors r+

llogiq:
You may have missed this, @bors r+

PR #9585, opened 2022-10-03
extend `box-default` lint, add suggestion

llogiq:
This extends the recently added `box-default` lint to also cover `Box::new(vec![])`, `Box::new(String::from(""))` and `Box::new(Vec::from([]))`. Also the lint now suggests a suitable replacement. I did not find a simple way to check whether the type is fully determined by the outside, so I at least checked for some variations to remove the turbofish in those cases.

---

changelog: none


llogiq:
Updated and rebased. r?

Alexendoo:
Thanks!

@bors r6

Alexendoo:
Whoops. @bors r+

PR #9586, opened 2022-10-03
FormatArgsExpn: Find comma spans and ignore weird proc macro spans

Alexendoo:
Fixes the following cases:

A missing `, 1` from the `expect_fun_call` suggestion:

```rust
Some(()).expect(&format!("{x} {}", 1));
```
```
warning: use of `expect` followed by a function call
 --> t.rs:7:14
  |
7 |     Some(()).expect(&format!("{x} {}", 1));
  |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `unwrap_or_else(|| panic!("{x} {}"))`
```

The suggestion removing from the comma in the comment rather than the one after the format string:

```rust
println!(
    "{}",
    // a comment, with a comma in it
    x
);
```
```
warning: variables can be used directly in the `format!` string
  --> t.rs:9:5
   |
9  | /     println!(
10 | |         "{}",
11 | |         // a comment, with a comma in it
12 | |         x
13 | |     );
   | |_____^
   |
help: change this to
   |
10 ~         "{x}",
11 ~         // a comment
   |
```

It also no longer accepts expansions where a format string or argument has a "weird" proc macro span, that is one where the literal/expression it outputs has the span of one of its inputs. Kind of like a `format_args` specific `clippy_utils::is_from_proc_macro`, e.g. `format!(indoc! {" ... "})`

changelog: [`expect_fun_call`]: Fix suggestion for `format!` using captured variables
changelog: [`print_literal`], [`write_literal`], [`uninlined_format_args`]: Fix suggestion when following a comment including a comma

### DIFF
--- a/clippy_utils/src/macros.rs
+++ b/clippy_utils/src/macros.rs
@@ -16,6 +16,7 @@ use rustc_parse_format::{self as rpf, Alignment};
 use rustc_span::def_id::DefId;
 use rustc_span::hygiene::{self, MacroKind, SyntaxContext};
 use rustc_span::{sym, BytePos, ExpnData, ExpnId, ExpnKind, Pos, Span, SpanData, Symbol};
+use std::iter::{once, zip};
 use std::ops::ControlFlow;
 
 const FORMAT_MACRO_DIAG_ITEMS: &[Symbol] = &[
@@ -412,7 +413,8 @@ impl FormatString {
 }
 
 struct FormatArgsValues<'tcx> {
-    /// See `FormatArgsExpn::value_args`
+    /// Values passed after the format string and implicit captures. `[1, z + 2, x]` for
+    /// `format!("{x} {} {y}", 1, z + 2)`.
     value_args: Vec<&'tcx Expr<'tcx>>,
     /// Maps an `rt::v1::Argument::position` or an `rt::v1::Count::Param` to its index in
     /// `value_args`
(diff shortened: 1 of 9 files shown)


tests/ui/uninlined_format_args.rs
```diff
@@ -147,11 +145,16 @@ fn tester(fn_arg: i32) {
...
-    // println!(pass_through!("foo={}"), local_i32);
-    // println!(indoc!("foo={}"), local_i32);
-    // printdoc!("foo={}", local_i32);
+    println!(
+        "{}",
+        // comment with a comma , in it
+        val,
+    );
+    println!("{}", /* comment with a comma , in it */ val);
+    println!("{named}", named = val);
```
nyurik, line 154: why do you need this test case?  It's being covered above too.

nyurik:
Overall looks good to me!  One thing you may want to reuse - I trimmed down the `indoc!` macro in the unit tests for https://github.com/rust-lang/rust-clippy/pull/9548 -- the `simple_indoc` -- would it make sense to merge that unit test into your code here?

tests/ui/uninlined_format_args.rs
Alexendoo: Yeah I'll remove that one, was just leftover from when I was testing the implementation in another file

Alexendoo:
`with_span!` covers the case where a proc macro sets the output span to one of its input

clippy_utils/src/macros.rs
```diff
@@ -675,6 +677,68 @@ impl<'tcx> Count<'tcx> {
...
+/// trailing comma
+///
+/// ```ignore
+/// format!("{} {}", a, b)
+/// //             ^  ^
+/// ```
+///
+/// Ensures that the format string and values aren't coming from a proc macro that sets the output
+/// span to that of its input
+fn comma_spans(cx: &LateContext<'_>, explicit_values: &[&Expr<'_>], fmt_span: Span) -> Option<Vec<Span>> {
```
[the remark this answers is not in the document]
Manishearth: or perhaps as a static fn on the format args expansion helper type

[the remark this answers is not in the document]
Alexendoo: 👍

Changed

clippy_utils/src/macros.rs
```diff
@@ -765,12 +767,82 @@ pub struct FormatArgsExpn<'tcx> {
...
+    ///
+    /// ```ignore
+    /// format!("..", 1, 2, 3,)
+    /// //          ^  ^  ^
+    /// ```
+    comma_spans: Vec<Span>,
+    /// Explicit values passed after the format string, ignoring implicit captures. `[1, z + 2]` for
     /// `format!("{x} {} {y}", 1, z + 2)`.
-    value_args: Vec<&'tcx Expr<'tcx>>,
+    explicit_values: Vec<&'tcx Expr<'tcx>>,
```
nyurik, line 780: we may want to clarify this comment -- the above `format!(...)` doesn't look like valid code.

Alexendoo: Ah whoops, the `{y}` should be `{}` I also forgot to remove [the FIXME](https://github.com/rust-lang/rust-clippy/blob/ac0e10aa68325235069a842f47499852b2dee79e/clippy_lints/src/format_args.rs#L176-L186) if you fancy doing a cleanup PR, otherwise I'll get around to it

PR #9587, opened 2022-10-03
[arithmetic-side-effects] Do not ignore literal references

c410-f3r:
To my utter surprise, `rustc` does does not warn stuff like `let n: u8 = &255 + &1` or `let n: u8 = 255 + &1`.

changelog: [arithmetic-side-effects] Do not ignore literal references


Alexendoo (approved):

Alexendoo:
Thanks again!

@bors r+

PR #9588, opened 2022-10-04
Fix incorrect uninlined_format_args edge case

nyurik:
An edge case with format arguments being wrapped in parenthesis. This might have been the real reason behind the #9540.

```
println!("{}", (local_i32));
```

Note that this changelog line might be irrelevant if the uninlined_format_args itself is added in the same release as this patch.

changelog[`uninlined_format_args`]: fix minor bug not handling params in parethethis

cc: @Alexendoo


nyurik:
closing because now all tests pass thx to #9586

PR #9590, opened 2022-10-04
Fix to_string_in_format_args in parens

### LINKED ISSUE #9540: Incorrect fix for to_string_in_format_args

nyurik:
Fix suggestions like

```
print!("error: something failed at {}", (Location::caller().to_string()));
```

where the parenthesis enclose some portion of the value.

Fixes #9540

changelog: [`to_string_in_format_args`]: fix incorrect fix when value is enclosed in parenthesis

### DIFF
--- a/tests/ui/format_args.rs
+++ b/tests/ui/format_args.rs
@@ -3,6 +3,7 @@
 #![allow(unused)]
 #![allow(
     clippy::assertions_on_constants,
+    clippy::double_parens,
     clippy::eq_op,
     clippy::print_literal,
     clippy::uninlined_format_args
@@ -114,6 +115,8 @@ fn main() {
     println!("error: something failed at {}", my_other_macro!());
     // https://github.com/rust-lang/rust-clippy/issues/7903
     println!("{foo}{foo:?}", foo = "foo".to_string());
+    print!("{}", (Location::caller().to_string()));
+    print!("{}", ((Location::caller()).to_string()));
 }
 
 fn issue8643(vendor_id: usize, product_id: usize, name: &str) {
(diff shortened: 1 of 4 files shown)


tests/ui/format_args.rs
```diff
@@ -114,6 +115,16 @@ fn main() {
...
+    print!("{}", (Location::caller().to_string()));
+    print!("{}", ((Location::caller()).to_string()));
+    #[rustfmt::skip]
+    print!("{}", (Location::caller()
+        .
+        to_string()));
+    #[rustfmt::skip]
+    print!("{}", ((Location::caller())
+        .
+        to_string()));
```
Alexendoo, line 127: Can remove these since they're the same as the above two aside from formatting

```suggestion
```

nyurik: @Alexendoo I made it this way on purpose because I wanted to ensure that the surround spacing is also captured.  I didn't want to just grab the dot, i wanted to grab everything after the ending of the prior expression.

nyurik:
CC: @llogiq in case you want to look at this one too - similar to the format args one

tests/ui/format_args.rs
Alexendoo: Sorry for the delay! 😅 

Generally we don't need to worry about whitespace to that degree as rustfmt will take care of it

Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

PR #9592, opened 2022-10-04
[arithmetic-side-effects] Detect overflowing associated constants of integers

c410-f3r:
Triggers the negation of maximum unsigned integers using associated constants. Rustc already handles `-128i8` but doesn't handle `-i8::MAX`.

At the same time, allows stuff like `-1234`.

changelog: FP: [arithmetic-side-effects] Detect overflowing associated constants of integers

### DIFF
--- a/clippy_lints/src/operators/arithmetic_side_effects.rs
+++ b/clippy_lints/src/operators/arithmetic_side_effects.rs
@@ -1,12 +1,19 @@
 use super::ARITHMETIC_SIDE_EFFECTS;
-use clippy_utils::{consts::constant_simple, diagnostics::span_lint};
+use clippy_utils::{
+    consts::{constant, constant_simple},
+    diagnostics::span_lint,
+    peel_hir_expr_refs,
+};
 use rustc_ast as ast;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::Ty;
 use rustc_session::impl_lint_pass;
-use rustc_span::source_map::{Span, Spanned};
+use rustc_span::{
+    source_map::{Span, Spanned},
+    sym,
+};
 
 const HARD_CODED_ALLOWED: &[&str] = &[
     "&str",
@@ -38,24 +45,6 @@ impl ArithmeticSideEffects {
         }
     }
 
-    /// Assuming that `expr` is a literal integer, checks operators (+=, -=, *, /) in a
-    /// non-constant environment that won't overflow.
-    fn has_valid_op(op: &Spanned<hir::BinOpKind>, expr: &hir::Expr<'_>) -> bool {
-        if let hir::ExprKind::Lit(ref lit) = expr.kind &&
-            let ast::LitKind::Int(value, _) = lit.node
-        {
-            match (&op.node, value) {
-                (hir::BinOpKind::Div | hir::BinOpKind::Rem, 0) => false,
-                (hir::BinOpKind::Add | hir::BinOpKind::Sub, 0)
-                    | (hir::BinOpKind::Div | hir::BinOpKind::Rem, _)
-                    | (hir::BinOpKind::Mul, 0 | 1) => true,
-                _ => false,
-            }
-        } else {
-            false
-        }
-    }
-
     /// Checks if the given `expr` has any of the inner `allowed` elements.
     fn is_allowed_ty(&self, ty: Ty<'_>) -> bool {
         self.allowed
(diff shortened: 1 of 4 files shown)


Jarcho:
`-i32::MIN` panics on debug builds. Is that not the kind of thing the lint is for?

Jarcho:
Ping @c410-f3r. Do you plan to continue working on this?

c410-f3r:
> Ping @c410-f3r. Do you plan to continue working on this?

Yes. Just needed a little break and I will probably get back to it today.

c410-f3r:
> `-i32::MIN` panics on debug builds. Is that not the kind of thing the lint is for?

I am sorry, I really tried to reproduce your statement but didn't achieve success. Can you provide a code snippet?

Jarcho:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e24b1ca1e6e6e74c18ae30e6e48e9a60

c410-f3r:
> https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e24b1ca1e6e6e74c18ae30e6e48e9a60

Errrr... I should have paid more attention 🙄 Thanks!

c410-f3r:
@Jarcho Now the lint will only trigger literal signed integers that represent their maximum negative value. On the contrary, things like `-123i8` are allowed.

c410-f3r:
Ah... NVM. Rustc already handles -(-128i8) but associated constant is a special case that I still need to tackle.

c410-f3r:
Well, I think this PR is now finally done.

clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -178,21 +210,31 @@ impl<'tcx> LateLintPass<'tcx> for ArithmeticSideEffects {
...
 
-/// Tells if an expression is a integer declared by value or by reference.
+/// Tells if an expression is declared by value or by reference.
 ///
-/// If `LiteralIntegerTy::Ref`, then the contained value will be `hir::ExprKind::Lit` rather
-/// than `hirExprKind::Addr`.
-enum LiteralIntegerTy<'expr, 'tcx> {
+/// If `OwnershipTy::Ref`, then the contained value will be the expression followed by `&`.
+#[derive(Clone, Copy)]
+enum ExprOwnershipTy<'expr, 'tcx> {
```
Jarcho, line 217: Looks like this can be replaced with `clippy_utils::peel_hir_expr_refs`.

c410-f3r: It it necessary to discriminate expressions with and without references because the compiler does not warn stuff like `let _ = &200u8 + &200u8`.

Although `peel_hir_expr_refs` does remove references, it doesn't tell if the given expression has references at all.

Jarcho: It both removes references and returns the number of removed references.

c410-f3r: Eeerrrrr... You are right, sorry.

Changed as requested.

clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -129,21 +123,53 @@ impl ArithmeticSideEffects {
...
     }
+
+    fn manage_unary_ops<'tcx>(
+        &mut self,
+        cx: &LateContext<'tcx>,
+        expr: &hir::Expr<'tcx>,
+        un_expr: &hir::Expr<'tcx>,
+        un_op: hir::UnOp,
+    ) {
+        if constant_simple(cx, cx.typeck_results(), un_expr).is_some() {
```
Jarcho, line 134: Would it be better to ignore all constants here? At least for the builtin types. I don't think a constant changing between the minimum value and another is particularly common.

c410-f3r: Using `constant` instead of `constant_simple`? I tried such approach without success because `constant` filters out `let _n = -i32::MIN;` in a runtime context. On the constant side, `const _N: i32 = -i32::MIN;` won't even let Clippy kick in.

c410-f3r:
@Jarcho I did some recent experimentation and realized that `-i32::MIN` is already caught by the compiler even in debug builds, which means that unary operations should only be triggered when dealing with variables (`-n`) but not literal integers (`-(-128)`, `-i32::MAX`, etc).

c410-f3r:
I already put a lot of effort trying to detect binary arithmetic with references so it seems that ignoring unary arithmetic is counter-productive.

But since this PR is already taking too much time, I will then move forward with `constant` and try to fix this reference stuff in `Rustc`.

c410-f3r:
Modified as requested

Jarcho:
Looks fine as is. Thank you, and sorry for the long wait on a review.

For the the issue with references, `&-constant` used to lint in rustc, but broke at some point, `-&constant` never linted in rustc (AFAIK) and requires mir inlining in the lint pass for it to work.

@bors r+

Issue #9197, opened 2022-07-17
useless_attribute with unsafe_rename


TheIronBorn:
### Summary

allowing `unsafe_removed_from_name` on imports produces a false positive on `useless_attribute`

### Lint Name

useless_attribute

### Reproducer

I tried this code:

```rust
#![allow(unused_imports)]

#[allow(clippy::unsafe_removed_from_name)]
use std::cell::UnsafeCell as Cell;
```

I saw this happen:

```
error: useless lint attribute
 --> src\lib.rs:3:1
  |
3 | #[allow(clippy::unsafe_removed_from_name)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: if you just forgot a `!`, use: `#![allow(clippy::unsafe_removed_fro
m_name)]`
  |
  = note: `#[deny(clippy::useless_attribute)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_attribute
```

I expected to see this happen:
No output

### Version

```text
rustc 1.64.0-nightly (d5e7f4782 2022-07-16)
binary: rustc
commit-hash: d5e7f4782e4b699728d0a08200ecd1a54d56a85d
commit-date: 2022-07-16
host: x86_64-pc-windows-msvc
release: 1.64.0-nightly
LLVM version: 14.0.6
```

alex-semenyuk:
You can achieve the result:

```
#![allow(unused_imports)]
#![allow(clippy::unsafe_removed_from_name)]
use std::cell::UnsafeCell as Cell;
```

or 

```
#[allow(unused_imports)]
#[allow(clippy::unsafe_removed_from_name)]
use std::cell::UnsafeCell as Cell;
```

PR #9593, opened 2022-10-05
lint::unsafe_removed_from_name: fix false positive result when allowed

Andy-Python-Programmer:
changelog: [`unsafe_removed_from_name`] Fix allowing on imports produces a false positive on `useless_attribute`.

Fixes: #9197

Signed-off-by: Andy-Python-Programmer <andypythonappdeveloper@gmail.com>


llogiq:
Thank you!

@bors r+

PR #9595, opened 2022-10-05
Replace if_chain with let chains in `clippy::author` output

Alexendoo:
Should help nudge new contributors towards let chains

changelog: none


Jarcho:
Thank you for the reminder that this exists.

@bors r+

PR #9599, opened 2022-10-06
Add a temporary workaround for  multiline formart arg inlining

nyurik:
per suggestion in
https://github.com/rust-lang/rust/pull/102729#discussion_r988990080

workaround for an internal crash when handling multi-line format argument inlining.

changelog: none

(no point for changelog because it is still a new lint being introduced)


flip1995:
@bors r+

Thanks!

PR #9600, opened 2022-10-06
Change uninlined_format_args into a style lint

nyurik:
As [previously discussed](https://github.com/rust-lang/rust-clippy/pull/9233#issuecomment-1256361205), the `uninlined_format_args` should probably be a part of the default style because `println!("{}", foo)` is not as concise or easy to understand as `println!("{foo}")`

changelog: [`uninlined_format_args`]: change to be the default `style`

### DIFF
--- a/clippy_lints/src/format_args.rs
+++ b/clippy_lints/src/format_args.rs
@@ -111,7 +111,7 @@ declare_clippy_lint! {
     /// nothing will be suggested, e.g. `println!("{0}={1}", var, 1+2)`.
     #[clippy::version = "1.65.0"]
     pub UNINLINED_FORMAT_ARGS,
-    pedantic,
+    style,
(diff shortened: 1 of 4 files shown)


llogiq:
Taking over review from @Alexendoo to reduce his load.

I'd like to see an `//edition:2018` test. I personally think the lint should probably not apply there, right?

r? @llogiq

nyurik:
> I'd like to see an `//edition:2018` test. I personally think the lint should probably not apply there, right?

@llogiq take a look at #9605 -- I tried to add a 2018 edition, and fix the bug with the panic, but somehow the test does not run as 2018th edition - could you see if i need to do something else there to fix it?  Thanks!

llogiq:
Ok, so I'll try getting #9605 merged first.

nyurik:
This one is ready to go

PR #9605, opened 2022-10-08
fix: uninlined_format_args shouldn't inline panic! before 2021ed

nyurik:
Before 2021 edition, `panic!("...")` was not treated as a format string.
Clippy autofix of `panic!("{}", foo)` into `panic!("{foo}")` is incorrect.

changelog: [`uninlined_format_args`]: Do not inline panic! macros before 2021 edition

### DIFF
--- a/tests/ui/uninlined_format_args_2021.rs
+++ b/tests/ui/uninlined_format_args_2021.rs
@@ -1,0 +1,23 @@
+// run-rustfix
+// edition:2021
+
+#![warn(clippy::uninlined_format_args)]
+
+fn main() {
+    let var = 1;
+
+    println!("val='{}'", var);
+
+    if var > 0 {
+        panic!("p1 {}", var);
+    }
+    if var > 0 {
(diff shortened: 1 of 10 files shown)


llogiq:
The correct way to determine the edition of code is span.edition(). The cx.sess().opts.edition is just the default from the compiler IIRC.

Since I'm already looking into it, r? @llogiq (Sorry/You're welcome Manish for stealing the review :smile: )

llogiq:
I'm not sure, it may be a relatively recent addition.

tests/ui/uninlined_format_args_2021.rs
```diff
@@ -0,0 +1,23 @@
+// run-rustfix
+// edition:2021
```
llogiq, line 2: This is the default edition, so this shouldn't be required here. But that's not keeping me from r+ing it.

llogiq (approved):

Issue #9603, opened 2022-10-07
Casting hexadecimal int literals to f32 is necessary


porky11:
### Summary

Rust sadly doesn't support non-decimal floating point numbers.

### Lint Name

clippy::unnecessary_cast

### Reproducer

I tried this code:

```rust
-0x400 as f32
```

I saw this happen:

```
warning: casting integer literal to `f32` is unnecessary
  --> src/world/layer.rs:26:29
   |
26 |                 Vector::new(-0x400 as f32, 0x40 as f32),
   |                             ^^^^^^^^^^^^^ help: try: `-0x400_f32`
```

I expected to see this happen:

I expected it not to work anymore after following the hint, and that's what happened.

### Version

```text
rustc 1.64.0 (a55dd71d5 2022-09-19)
binary: rustc
commit-hash: a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52
commit-date: 2022-09-19
host: x86_64-unknown-linux-gnu
release: 1.64.0
LLVM version: 14.0.6
```

kraktus:
I'm unable to reproduce, it runs fine on playground ([repro](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c602b73aa3299ef641534302c655c689)). Can you share a minimum reproducible example @porky11? Thanks

porky11:
@kraktus 

```
fn main() {
    println!("{}", -0x10 as f32);
}
```

It turns out the warning only appears when the hexadecimal number is negative.

porky11:
@kraktus What even is your example?
Something seems wrong here. `-0x400_f32` is of an opaque integer type, not of type f32.

Try this:

```rust
fn main() {
    let x: f32 = -0x400_f32;
    println!("{x}");
}
``` 

Is this a rust bug?

porky11:
Now I get it: In hexadecimal, f32 is a digit. `_` is allowed inside of numbers between digits as well.
So `-0x400_f32` is the same as `0x400F32`. Just a hexadecimal integer.

kraktus:
Nice catch!

PR #9609, opened 2022-10-08
[`unnecessary_cast`] Do not lint negative hexadecimal literals when cast as floats

kraktus:
fix https://github.com/rust-lang/rust-clippy/issues/9603

changelog: [`unnecessary_cast`] Do not lint negative hexadecimal literals when cast as floats

### DIFF
--- a/clippy_lints/src/casts/unnecessary_cast.rs
+++ b/clippy_lints/src/casts/unnecessary_cast.rs
@@ -59,9 +59,6 @@ pub(super) fn check<'tcx>(
                 lint_unnecessary_cast(cx, expr, literal_str, cast_from, cast_to);
                 return false;
             },
-            LitKind::Int(_, LitIntType::Unsuffixed) | LitKind::Float(_, LitFloatType::Unsuffixed) => {
-                return false;
-            },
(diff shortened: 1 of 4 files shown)


clippy_lints/src/casts/unnecessary_cast.rs
```diff
@@ -59,9 +59,6 @@ pub(super) fn check<'tcx>(
                 lint_unnecessary_cast(cx, expr, literal_str, cast_from, cast_to);
                 return false;
             },
-            LitKind::Int(_, LitIntType::Unsuffixed) | LitKind::Float(_, LitFloatType::Unsuffixed) => {
-                return false;
-            },
```
giraffate, line 64: Why are these removed? It looks like that the fix for `clippy_utils/src/numeric_literal.rs` is sufficient to me.

kraktus: It was dead code in the first place

kraktus: To be exact not dead but if the branching is not taken, it will still result to the function returning `false` and since the code triggered afterward is cheap it’s not worth the extra complexity of an early return

giraffate:
@bors r+

Thanks!

Issue #7958, opened 2021-11-11
Improve filtering options for Clippy's lint list


xFrednet:
[Clippy's lint list](https://rust-lang.github.io/rust-clippy/master/index.html) currently allows filtering by lint and group. It would be helpful to add more filters for the applicability and version values which are available. The filter design that is currently used is not set in stone, feel free to experiment a bit and make some design suggestions.

The HTML document of the page can be found under [`util/gh-pages/index.html`](https://github.com/rust-lang/rust-clippy/blob/master/util/gh-pages/index.html). `cargo dev serve` can be used to test the changes locally. The current lint data representation can be found in the `gh-pages` branch [here](https://github.com/rust-lang/rust-clippy/blob/gh-pages/master/lints.json)

Main tracking issue #7172

flip1995:
Possible improvements for the current filters:

- Group filter:
    - Make drop-down menu 
    - Add options to select all / none
- Lint levels:
    - Possibly the same as for groups?
- Search
    - Fuzzy search (for lint names)

Possible new lint groups:

- Version
    - greater
    - equal
    - less
- Applicability
    - Filter by applicability group
    - Same interface as the one suggested for groups and levels

xFrednet:
A *search only in lint name* function would also be nice.

repi:
Would be great to be able to esp. filter on the version the lint was introduced, use that quite frequently to test lints in newer Rust versions.

Serial-ATA:
I'll give the version filter a shot

SpencerAWill:
I'll take a look at the applicability filter

PR #9616, opened 2022-10-09
Add Default to Clippy Lints Lint groups

unvalley:
- related to #7958

This PR adds a default (reset) button to Clippy Lints Lint groups. (change for website)
[The page](https://rust-lang.github.io/rust-clippy/master/index.html) sets only `Deprecated` to false by default.
Certainly it is easy to set only `deprecated` to false, but it may be a bit lazy for beginners.

https://user-images.githubusercontent.com/38400669/194831117-3ade7e0d-c4de-4189-9daf-3be8ea3cdd18.mov

changelog: none

### DIFF
--- a/util/gh-pages/index.html
+++ b/util/gh-pages/index.html
@@ -448,6 +448,12 @@ Otherwise, have a great day =^.^=
                                         None
                                     </label>
                                 </li>
+                                <li class="checkbox">
+                                    <label ng-click="resetGroupsToDefault()">
+                                        <input type="checkbox" class="invisible" />
+                                        Default
+                                    </label>
+                                </li>
                                 <li role="separator" class="divider"></li>
                                 <li class="checkbox" ng-repeat="(group, enabled) in groups">
                                     <label class="text-capitalize">
--- a/util/gh-pages/index.html
+++ b/util/gh-pages/index.html
@@ -443,15 +443,15 @@ Otherwise, have a great day =^.^=
                                     </label>
                                 </li>
                                 <li class="checkbox">
-                                    <label ng-click="toggleGroups(false)">
-                                        <input type="checkbox" class="invisible" />
-                                        None
-                                    </label>
-                                </li>
-                                <li class="checkbox">
                                     <label ng-click="resetGroupsToDefault()">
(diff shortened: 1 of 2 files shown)


flip1995:
r? @xFrednet I think you are the best reviewer for this.

Should the `Default` option enable all lints that are warn-by-default, instead of all groups except for `deprecated`? That would've been what I expected.

The `All`  option can also just set `deprecated` to false IMO.

util/gh-pages/script.js
```diff
@@ -164,6 +168,15 @@
                 }
             };
 
+            $scope.resetGroupsToDefault = function () {
+                const groups = $scope.groups;
+                for (const [key, value] of Object.entries(GROUPS_FILTER_DEFAULT)) {
+                    if (groups.hasOwnProperty(key)) {
```
xFrednet, line 174: This should always be true, shouldn't it?

util/gh-pages/index.html
```diff
@@ -448,6 +448,12 @@ <h1>Clippy Lints</h1>
                                         None
                                     </label>
                                 </li>
+                                <li class="checkbox">
```
xFrednet, line 451: I would maybe put this between "All" and "None", since it's the middle ground. Though this version looks a bit better, as the word length increases

xFrednet (commented):
Overall looks good to me, two smaller suggestions

xFrednet:
> Should the `Default` option enable all lints that are warn-by-default, instead of all groups except for `deprecated`? That would've been what I expected.

I think default should reset to the default of the lint list. If we only want it to be warn-by-default lints, then that should also be the default of the list IMO. With context, I think that this is already covered by the lint level selection. We could maybe add a second button, which is `clippy::all` :thinking:

util/gh-pages/index.html
unvalley: I can agree with you. The order "All", "Default", "None" is natural.

unvalley:
@xFrednet Thanks, I added a commit for your suggestion.

xFrednet (approved):

xFrednet:
Looks good to me and runs perfectly. Thank you for the changes!

@bors r+

PR #9617, opened 2022-10-09
add `cast-nan-to-int` lint

### LINKED ISSUE #371: Casting a known NaN value to an integral type
This is actually undefined behavior, and doesn't even get a warning from rustc.

llogiq:
This fixes #371.

r? @Alexendoo

---

changelog: add [`cast-nan-to-int`] lint

### DIFF
--- a/tests/ui/cast_nan_to_int.stderr
+++ b/tests/ui/cast_nan_to_int.stderr
@@ -1,0 +1,51 @@
+error: casting a known NaN to usize
+  --> $DIR/cast_nan_to_int.rs:5:13
+   |
+LL |     let _ = (0.0_f32 / -0.0) as usize;
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: this always evaluates to 0
+   = note: `-D clippy::cast-nan-to-int` implied by `-D warnings`
+
+error: casting a known NaN to usize
+  --> $DIR/cast_nan_to_int.rs:6:13
+   |
+LL |     let _ = (f64::INFINITY * -0.0) as usize;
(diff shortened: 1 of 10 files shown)


tests/ui/cast_nan_to_int.stderr
```diff
@@ -0,0 +1,40 @@
+error: casting a known NaN to usize
+  --> $DIR/cast_nan_to_int.rs:5:13
+   |
+LL |     let _ = (0.0_f32 / -0.0) as usize;
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: `-D clippy::cast-nan-to-int` implied by `-D warnings`
```
Alexendoo, line 7: Would be worth adding a note explaining that the result will be 0

Alexendoo: Maybe we should just lint all operations that result in `NaN`? Whether cast or not

llogiq: That would mean we'd need to check far more expressions. This way we only look into the constant evaluation when we have a cast from float to int, which is sufficiently rare (also in earlier versions of Rust, this cast was actually UB, but that's quite a while ago now). Also the risk of false positives seems higher.

I'm fine with retiring this lint once we have one that checks for all NaNs and works reasonably well.

llogiq:
I've added the note you suggested. Anything else needed?

Alexendoo (approved):
Nothing else, looks good!

you can r=me it with the conflict resolved

llogiq:
Thank you for the review!

@bors r=@Alexendoo

PR #9623, opened 2022-10-10
add [`unnecessary_vec_drain] lint

### LINKED ISSUE #9339: new lint: `vec.drain(..)` instead of `vec.clear()`
### What it does

Checks for usage of `vec.drain(..)` (iterator dropped immediately) which can be replaced with `vec.clear()`

### Lint Name

unnecessary_vec_drain

### Category

style

### Advantage

- Makes intent clearer
- Marginally more efficient, especially in debug builds

### Drawbacks

_No response_

### Example

```rust
vec.drain(..);
```

Could be written as:

```rust
vec.clear();
```

lana99:
fixes #9339 

detects usage of drain with RangeFull and dropped iterator
```rust
vec.drain(..)
vec.drain(0..vec.len())
```
will suggest to change to
```rust
vec.clear()
```

changelog: New lint: [`unnecessary_vec_drain`]
[#9623](https://github.com/rust-lang/rust-clippy/pull/9623)


clippy_lints/src/unnecessary_vec_drain.rs
```diff
@@ -0,0 +1,96 @@
...
+declare_lint_pass!(UnnecessaryVecDrain => [UNNECESSARY_VEC_DRAIN]);
+
+impl LateLintPass<'_> for UnnecessaryVecDrain {
+    fn check_stmt<'tcx>(&mut self, cx: &LateContext<'tcx>, stmt: &'tcx Stmt<'tcx>) {
+        if let hir::StmtKind::Semi(semi_expr) = &stmt.kind {
+            if let hir::ExprKind::MethodCall(path, rcvr, method_args,drain_span) = &semi_expr.kind
+                && path.ident.name == sym!(drain)
+            {
+                let ty = cx.typeck_results().expr_ty(rcvr);
+                if let [expr_element] = &**method_args
```
flip1995, line 45: You can pull this into the check above and already match for the single argument in `methods_args`.

clippy_lints/src/unnecessary_vec_drain.rs
```diff
@@ -0,0 +1,96 @@
...
+                    && is_type_diagnostic_item(cx, ty, sym::Vec)
```
flip1995, line 46: You're combining `if let` chains with the `if_chain` macro. Please use `if let` chains everywhere. 

You can also use irrefutable patterns in `if let` chains, so this can be all in one chain up until here.

clippy_lints/src/unnecessary_vec_drain.rs
```diff
@@ -0,0 +1,96 @@
...
+                            if let hir::ExprKind::MethodCall(path_seg, vec_expr, _, _) = expr_fields[1].expr.kind;
+
+                            if let hir::hir_id::HirId{owner: owner_expr,..} = &vec_expr.hir_id;
+                            if let hir::hir_id::HirId{owner: owner_rcvr,..} = &rcvr.hir_id;
+
+                            then {
+                                if let Int(start,_) = lit.node
+                                    && path_seg.ident.name == sym!(len)
+                                    && owner_rcvr == owner_expr
+                                    && start == 0
```
flip1995, line 76: NIT: The order of these checks are a bit weird. This check should be done right after matching it to `ExprKind::Lit`.

tests/ui/unnecessary_vec_drain.stderr
```diff
@@ -0,0 +1,16 @@
...
+LL |     vec.drain(..);
+   |     ^^^^^^^^^^^^^ help: consider calling `clear()`: `vec.clear()`
+   |
+   = note: `-D clippy::unnecessary-vec-drain` implied by `-D warnings`
+
+error: unnecessary iterator `Drain` is created and dropped immediately
+  --> $DIR/unnecessary_vec_drain.rs:8:9
+   |
+LL |     vec.drain(0..vec.len());
+   |         ^^^^^^^^^^^^^^^^^^^ help: consider calling `clear()`: `vec.clear()`
```
flip1995, line 13: This suggestion won't work. The code after applying this suggestion would be `vec.vec.clear()`. You'll have to fix the span your using when producing this suggestion to also include `vec`.

flip1995 (changes requested):
Some code cleanup has to be done here.

bluthej:
This can be closed since #10528 closed issue #9339

blyxyas:
Closing this, as #9339 was fixed by #10528. @lana99 Thanks for trying! :heart: ฅ^•ﻌ•^ฅ

PR #9624, opened 2022-10-10
Add new lint `string_as_str`

guerinoni:

We're collecting our changelog from pull request descriptions.
If your PR only includes internal changes, you can just write
explaining your change.

It's also helpful for us that the lint name is put within backticks (`` ` ` ``),
and then encapsulated by square brackets (`[]`), for example:
```
changelog: [`string_as_str`]: Add new lint
```

If your PR fixes an issue, you can add `fixes #issue_number` into this
your PR is merged.

- \[x] Run `cargo dev fmt`

Note that you can skip the above if you are just opening a WIP PR in
order to get feedback.

Delete this line and everything above before opening your PR.

---

changelog: Add new lint `string_as_str`


guerinoni:
Thanks to @flip1995 for some hints :)

clippy_lints/src/string_as_str.rs
```diff
@@ -0,0 +1,98 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, ty::is_type_diagnostic_item};
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind, QPath};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
```
giraffate, line 7: Can this be moved to `Method` lint pass? This commit is helpful: https://github.com/rust-lang/rust-clippy/commit/8acc4d2f1e255e0e97e48b7a285a7d6db4a03bd6

guerinoni: Done :)

tests/ui/string_as_str.stderr
```diff
@@ -0,0 +1,22 @@
...
+  --> $DIR/string_as_str.rs:9:17
+   |
+LL |     fn_with_str(my_string.as_str());
+   |                 ^^^^^^^^^^^^^^^^^^ help: replace it with: `&my_string`
+
+error: used `as_str` for reference
+  --> $DIR/string_as_str.rs:30:13
+   |
+LL |     let a = s.as_str().chars().map(|c| c.is_digit(2)).collect::<Vec<_>>();
+   |             ^^^^^^^^^^ help: replace it with: `&s`
```
flip1995, line 19: This suggestion won't work. It should suggest `(&s)`. You can use `clippy_utils::Sugg` with `maybe_paren`  to automatically produce the parens.

clippy_lints/src/methods/string_as_str.rs
```diff
@@ -0,0 +1,60 @@
...
+
+fn is_string(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    is_type_diagnostic_item(
+        cx,
+        cx.typeck_results().expr_ty(expr).peel_refs(),
+        rustc_span::sym::String,
+    )
+}
+
+fn is_within_borrowed_call(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
```
flip1995, line 38: I'm not sure what this function is supposed to do. It looks like it checks for really specific things in the tests you've written.

clippy_lints/src/methods/string_as_str.rs
```diff
@@ -0,0 +1,60 @@
...
+                Applicability::MaybeIncorrect,
+            );
+        }
+    }
+}
+
+fn is_string(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    is_type_diagnostic_item(
+        cx,
+        cx.typeck_results().expr_ty(expr).peel_refs(),
```
flip1995, line 33:
```suggestion
        cx.typeck_results().expr_ty(expr),
```

I think this should just be this and with that you probably won't need the `is_within_borrowed_call` function. The `peel_refs` causes the lint to trigger on `&String` types.

flip1995 (changes requested):
Some suggestions on how to clean up the lint code.

clippy_lints/src/methods/string_as_str.rs
guerinoni: Don't know how to not lint here https://github.com/rust-lang/rust-clippy/blob/50f192f86a12d4b39d8f7f061fbd8e0ba8007932/clippy_lints/src/missing_enforced_import_rename.rs#L77

flip1995: Why can't this be linted?

guerinoni: Because using `&snip` there are different type on those two branch, one is &str and other &String, don't know if we can change the code in other way

flip1995 (commented):
Took a closer look at this lint. The suggestion of changing this to `&s` only applies when the `String` can be coerced to `str`. This is the case in calls where the argument is of type `&str`. In every other place the suggestion has to be `&*`  I can see two paths forward here: 

1. Check if the expression parent is some kind of call and the parameter of that function/method/closure has type `&str`
2. Just lint everything and suggest `&*`. But in this case, this lint should be `restriction` IMO, because I don't think that `&*s` is better than `s.as_str()`. Even for a `pedantic` lint, I think this is a questionable suggestion.

guerinoni:
@flip1995 Do you have any suggestion for checking parameter of `fn`, I prefer to implement the option 1 because the impact can be more useful for a codebase

tests/ui/string_as_str.stderr
kraktus: I don't think we should lint where parenthesis are needed. This feels like the canonical use of `as_str`

clippy_lints/src/methods/string_as_str.rs
guerinoni: Exactly... Don't sure next step to complete this lint :( 🤔

flip1995:
I think we have a ``expr_fn_sig`` utils function, that you can use to get the method/function signature and with that the type of the argument.

tests/ui/string_as_str.stderr
guerinoni: Avoid to lint this

PR #9629, opened 2022-10-10
Replace manual let else patterns with let else

est31:
Clears the codebase from places where the lint added by #8437 is firing, by adopting let else.

changelog: none


Jarcho:
Looks good. Thank you.

@bors r+

PR #9630, opened 2022-10-11
Fix allow_attributes_without_reason applying to external crate macros

CatDevz:
Previously the `clippy::allow_attributes_without_reason` lint would apply to external crate macros. Many macros in the Rust ecosystem include these `allow` attributes without adding a reason, making this lint pretty much unusable in any sizable Rust project.

This commit fixes that by adding a check to the lint if the attribute is from an external crate macro and returning early.

```
changelog: [`allow_attributes_without_reason`]: allow_attributes_without_reason no longer applies to external crate macros
```


flip1995 (approved):
SGTM

flip1995:
@bors r+

Thanks!

PR #9635, opened 2022-10-12
Fix bug introduced by #9386

smoelius:
#9386 introduced a potential out-of-bounds array access. Specifically, a location returned by `local_assignments` could have  [`location.statement_index` equal to `mir.basic_blocks[location.block].statements.len()`](https://github.com/rust-lang/rust-clippy/blob/b8a9a507bf9e3149d287841454842116c72d66c4/clippy_utils/src/mir/mod.rs#L129), in which case the location would refer to the block terminator:
https://github.com/rust-lang/rust-clippy/blob/b8a9a507bf9e3149d287841454842116c72d66c4/clippy_lints/src/dereference.rs#L1204-L1206
I suspect the bug is not triggerable now, because of checks leading up to where it occurs. But a future code change could make it triggerable. Hence, it should be fixed.

r? @Jarcho 

changelog: none

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -1194,16 +1194,16 @@ fn has_ref_mut_self_method(cx: &LateContext<'_>, trait_def_id: DefId) -> bool {
         })
 }
 
-fn referent_used_exactly_once<'a, 'tcx>(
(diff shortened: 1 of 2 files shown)


Jarcho:
Thank you. @bors r+

PR #9640, opened 2022-10-13
Fix edition revision ui tests

Alexendoo:
#9605 had me wondering how the edition revision tests were working for `manual_assert` but not for @nyurik, but it turns out `manual_assert`'s tests weren't working either. I checked how `rust-lang/rust` does it and apparently it comes down to whitespace, `//[rev] edition:X` works 😬

Removes the revisions from `match_wild_err_arm` as I couldn't find any edition dependant behaviour there

r? @llogiq

changelog: none


tests/ui/manual_assert.rs
```diff
@@ -1,6 +1,6 @@
 // revisions: edition2018 edition2021
-// [edition2018] edition:2018
-// [edition2021] edition:2021
+//[edition2018] edition:2018
+//[edition2021] edition:2021
```
llogiq, line 3: I'm curious: Does this compile the code under both revisions?

Alexendoo: It does yeah, a test is compiled once per revision, the `[revision]` syntax makes the magic comment thingies only apply on that revision

nyurik:
Thanks for digging into this! I feel slightly validated that I wasn't going crazy with it :)  That said, ideally spacing shouldn't play this much of a role IMO, and perhaps we should get a PR for the upstream to ignore extra spaces?  Regardless, thx!

PR #9641, opened 2022-10-13
Enable cargo sparse registry in CI

Alexendoo:
https://blog.rust-lang.org/2022/06/22/sparse-registry-testing.html

The initial registry update takes around 1 minute currently, so this gives quite a nice speed boost to CI build times

r? @flip1995

changelog: none


flip1995:
@bors r+

Thanks!

PR #9644, opened 2022-10-14
add missing comma

hkBst:
changelog: none


flip1995 (approved):
To not be flooded with those one character changes during hacktoberfest, we won't count those PRs. Thanks for improving Clippy a little bit though!

PR #9648, opened 2022-10-14
fix `undocumented-unsafe-blocks` false positive

### LINKED ISSUE #9142: False positive in undocumented_unsafe_blocks (stable-to-stable regression)
### Summary

When defining a variable using an unsafe block, if the unsafe block is not on the same line as the variable definition, the undocumented_unsafe_blocks triggers a warning when enabled. This did not trigger the lint in Rust 1.61. It does trigger the lint in Rust 1.62 as well as the latest nightly.

### Lint Name

undocumented_unsafe_blocks

### Reproducer

I tried this code:

```rust
// SAFETY: We write to the array below before reading from it.
let mut col_coeffs_backing: Aligned<[i32; 64]> =
  unsafe { Aligned::uninitialized() };
```

I saw this happen:

```
warning: unsafe block missing a safety comment
   --> src/transform/forward.rs:129:9
    |
129 |         unsafe { Aligned::uninitialized() };
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
   --> src/lib.rs:69:9
    |
69  | #![warn(clippy::undocumented_unsafe_blocks)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = help: consider adding a safety comment on the preceding line
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks
```

I expected to see no warning.

In Rust 1.61, there is no warning, as expected. Additionally, keeping the code on one line, or moving the SAFETY comment down one line, both resolve the warning. e.g. the following two examples both produce no warning:

```rust
// SAFETY: We write to the array below before reading from it.
let mut col_coeffs_backing: Aligned<[i32; 64]> = unsafe { Aligned::uninitialized() };
```

```rust
let mut col_coeffs_backing: Aligned<[i32; 64]> =
  // SAFETY: We write to the array below before reading from it.
  unsafe { Aligned::uninitialized() };
```

### Version

```text
Reproducible on both:

binary: rustc
commit-date: 2022-06-27
release: 1.62.0
LLVM version: 14.0.5


and:

rustc 1.64.0-nightly (06754d885 2022-07-08)
binary: rustc
commit-hash: 06754d8852bea286a3a76d373ccd17e66afb5a8b
commit-date: 2022-07-08
release: 1.64.0-nightly
LLVM version: 14.0.6
```

### Additional Labels

@rustbot +label regression-from-stable-to-nightly regression-from-stable-to-stable

llogiq:
This fixes #9142 by iterating over the parent nodes as long as within a block, expression, statement, local, const or static.

---

changelog: none

### DIFF
--- a/tests/ui/undocumented_unsafe_blocks.rs
+++ b/tests/ui/undocumented_unsafe_blocks.rs
@@ -490,4 +490,23 @@ unsafe impl CrateRoot for () {}
 // SAFETY: ok
 unsafe impl CrateRoot for (i32) {}
 
+fn issue_9142() {
+    // SAFETY: ok
+    let _ =
+        // we need this comment to avoid rustfmt putting
+        // it all on one line
+        unsafe {};
+
+    // SAFETY: this is more than one level away, so it should warn
+    let _ = {
+        if unsafe { true } {
+            todo!();
+        } else {
+            let bar = unsafe {};
+            todo!();
+            bar
+        }
+    };
+}
+
 fn main() {}
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -68,7 +68,8 @@ impl LateLintPass<'_> for UndocumentedUnsafeBlocks {
             && !in_external_macro(cx.tcx.sess, block.span)
             && !is_lint_allowed(cx, UNDOCUMENTED_UNSAFE_BLOCKS, block.hir_id)
             && !is_unsafe_from_proc_macro(cx, block.span)
-            && !block_has_safety_comment(cx, block)
+            && !block_has_safety_comment(cx, block.span)
+            && !block_parents_have_safety_comment(cx, block.hir_id)
         {
             let source_map = cx.tcx.sess.source_map();
             let span = if source_map.is_multiline(block.span) {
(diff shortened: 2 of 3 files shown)


tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -490,4 +490,20 @@ unsafe impl CrateRoot for () {}
...
+fn issue_9142() {
+    // SAFETY: ok
+    let _ = unsafe {};
+
+    // SAFETY: ok
+    let _ = {
+        if unsafe { true } {
+            todo!();
+        } else {
+            let bar = unsafe {};
```
Jarcho, line 502: Should this really be seen as ok? And if so, what about something like
```rust
// SAFETY: ok
match foo {
    Foo::Bar => unsafe { something() },
    // A couple hundred lines here
   Foo::Baz => unsafe { something_else() },
}
```

llogiq: How would you define locality? Should I remove blocks?

llogiq: I have removed blocks from the equation.

llogiq:
I now no longer look at unsafe blocks.

This would of course still allow e.g. `foo(bar(Baz(unsafe { .. })))` including possible newlines in between and accept the safety comment before `foo`.

tests/ui/undocumented_unsafe_blocks.rs
Jarcho: This lint has a problem that it's very much a 'what would the person reading think' sort of thing. It's very formatting dependent.

Even with the blocks ignored a single expression can span a large number of lines.
```rust
some_fn(
    arg1 + with_some_stuff,
    unsafe { arg2.unsafe_fn() },
    some
        .iter()
        .map(|x| do_something(x))
        .collect(),
    //  A few other arguments
    unsafe { arg3.other_unsafe() },
);
```

llogiq:
Ping @giraffate

Or @Jarcho do you want to take over the review?

Jarcho:
I guess I can take over.

I'm still not sure about walking up expressions in general. Method and let chains can result in some pretty crazy expressions. Clippy in particular is pretty bad about having 40-50 line let chains, although I assume that's unusual in other projects.

I think the best plan here would be to fix the current regression by only walking up a single level if the parent is a let statement (and in the same `SyntaxContext`) and checking there. We can then figure out what forms should be allowed by the lint and deal with that in a different PR.

llogiq:
@Jarcho Ok, this now only looks one parent upwards. r?

llogiq:
That comment was meant for Jarcho, not Rustbot. But nice error 😉.

tests/ui/undocumented_unsafe_blocks.rs
```diff
@@ -490,4 +490,20 @@ unsafe impl CrateRoot for () {}
 // SAFETY: ok
 unsafe impl CrateRoot for (i32) {}
 
+fn issue_9142() {
+    // SAFETY: ok
+    let _ = unsafe {};
```
Jarcho, line 495: Can you change this to be on two lines. e.g.
```rust
// SAFETY: ok
let _ =
    unsafe {}
```

llogiq: Ah, rustfmt pulled this together.

Jarcho:
All looks good now. Thank you.

@bors r+

Issue #9638, opened 2022-10-13
From over Into for external crate types which is impossible to implement


Firstyear:
### Summary

Clippy lints stating that from over into is preferred is impossible to satisfy. You can't implement a trait on a struct from an external crate. 

Also common to hit with should_implement_trait.

Clippy should not recommend From over Into or Should implement trait when the target type is an out of crate type which is impossible to compile for. 

### Lint Name

from_over_into

### Reproducer

```
error: an implementation of `From` is preferred since it gives you `Into<_>` for free where the reverse isn't true
   --> webauthn-rs-proto/src/extensions.rs:186:1
    |
186 | impl Into<js_sys::Object> for &RequestAuthenticationExtensions {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider to implement `From<&extensions::RequestAuthenticationExtensions>` instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into

```

### Version

```text
rustc 1.64.0 (a55dd71d5 2022-09-19)
binary: rustc
commit-hash: a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52
commit-date: 2022-09-19
host: aarch64-apple-darwin
release: 1.64.0
LLVM version: 14.0.6
```

kraktus:
You can implement `From<LocalStruct> for ExternalStruct`, as `From<LocalStruct>` is considered an internal trait. This behavior could be édition gated though I don’t remember the details

Firstyear:
@kraktus It doesn't work with FromIterator though I think?

Alexendoo:
Do you have an example of a `should_implement_trait` case that would fail to compile?

For `from_over_into` we could emit a note saying it's fine to do `From<Local> for Foreign` if a foreign type is involved, as this is a very common source of confusion

Firstyear:
@Alexendoo A more complex example, but https://github.com/kanidm/kanidm/blob/master/kanidmd/lib/src/valueset/spn.rs#L28 . tl;dr you can't do fromIterator on a tuple of strings because tuple is always foreign.

[...]

PR #9649, opened 2022-10-14
Add a suggestion and a note about orphan rules for `from_over_into`

### LINKED ISSUE #7444: Do not warn about from_over_into with external crate types
Lint name: `from_over_into`

This lint should not be attempted on types that live in external crates. Right now it triggers e.g.:
```rust
impl Into<OutOfCrateType> for InCrateType {...}
```

*Not to get this case linted as a warning because the `OutOfCrateType` is external and thus `From` cannot or might not be implementable on it since `InCrateType` is from this crate.*

Instead, this happened: 
*It got linted as a warning.*

### Meta

- `cargo clippy -V`: clippy 0.1.51 (2fd73fa 2021-03-23)
- `rustc -Vv`:
```
rustc 1.51.0 (2fd73fabe 2021-03-23)
binary: rustc
commit-hash: 2fd73fabe469357a12c2c974c140f67e7cdd76d0
commit-date: 2021-03-23
release: 1.51.0
LLVM version: 11.0.1
```

Alexendoo:
Adds a machine applicable suggestion to convert the `Into` impl into a `From` one to `from_over_into`

Also adds a note explaining that `impl From<Local> for Foreign` is fine if the `Into` type is foreign

Closes #7444
Addresses half of #9638

changelog: [`from_over_into`] Add a suggestion and a note about orphan rules

### DIFF
--- a/clippy_lints/src/from_over_into.rs
+++ b/clippy_lints/src/from_over_into.rs
@@ -1,11 +1,19 @@
-use clippy_utils::diagnostics::span_lint_and_help;
-use clippy_utils::{meets_msrv, msrvs};
-use if_chain::if_chain;
-use rustc_hir as hir;
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::macros::span_is_local;
+use clippy_utils::source::snippet_opt;
+use clippy_utils::{meets_msrv, msrvs, path_def_id};
+use rustc_errors::Applicability;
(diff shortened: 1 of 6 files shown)


llogiq:
Very cool! The suggestion comments were a bit befuddling at first, because they all start from the same base when in fact they build one upon another, but that only threw me off for one or two minutes. 🙃

@bors r+

PR #9654, opened 2022-10-15
Enable test no_std_main_recursion

alex-semenyuk:
Verified that test actually works on windows
changelog: none


Alexendoo (approved):

Alexendoo:
Thanks, indeed it does

@bors r+

PR #9658, opened 2022-10-16
Add new lint `partial_pub_fields`

### LINKED ISSUE #9604: `partial_pub_fields`: disallow partial fields of a struct be pub
### What it does

https://matklad.github.io//2022/05/29/binary-privacy.html

Most structs are either `ADT` or `Data`, they shouldn't have a non-empty subset of pub fields.

The rule can be default allowed, since it may break some public APIs.

For `pub(crate)` or `pub(mod path)`, it's commonly used as a friend class alternative, so I think we can treat them as private. If there are some requirements, we can add an option `strict_partial_pub_fields` later.

We can also introduce `partial_pub_variants` lint for enums later.

The lint can't be fixed automatically, since we don't know a struct is `ADT` or `Data`.

### Lint Name

partial_pub_fields

### Category

restriction

### Advantage

- Make a clear boundary of ADT or Data

### Drawbacks

- May break some published APIs

### Example

```rust
#[derive(Default)]
pub struct FileSet {
  pub files: HashMap<VfsPath, FileId>,
  paths: HashMap<FileId, VfsPath>,
}
```

Could be written as:

```rust
#[derive(Default)]
pub struct FileSet {
  files: HashMap<VfsPath, FileId>,
  paths: HashMap<FileId, VfsPath>,
}
```

```rust
pub struct Color {
    pub r,
    pub g,
    b,
}
```

Could be written as:

```rust
pub struct Color {
    pub r,
    pub g,
    pub b,
}
```

TennyZhuang:
Signed-off-by: TennyZhuang <zty0826@gmail.com>

changelog: `partial_pub_fields`: new lint to disallow partial fields of a struct be pub

Resolve #9604

### DIFF
--- a/clippy_lints/src/partial_pub_fields.rs
+++ b/clippy_lints/src/partial_pub_fields.rs
@@ -1,5 +1,5 @@
 use clippy_utils::diagnostics::span_lint_and_help;
-use rustc_ast::ast::*;
+use rustc_ast::ast::{Item, ItemKind};
 use rustc_lint::{EarlyContext, EarlyLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
@@ -58,7 +58,7 @@ impl EarlyLintPass for PartialPubFields {
             if all_priv && field.vis.kind.is_pub() {
                 span_lint_and_help(
                     cx,
-                    &PARTIAL_PUB_FIELDS,
+                    PARTIAL_PUB_FIELDS,
                     field.vis.span,
                     msg,
                     None,
@@ -68,7 +68,7 @@ impl EarlyLintPass for PartialPubFields {
             } else if all_pub && !field.vis.kind.is_pub() {
                 span_lint_and_help(
                     cx,
-                    &PARTIAL_PUB_FIELDS,
+                    PARTIAL_PUB_FIELDS,
                     field.vis.span,
                     msg,
                     None,
--- a/clippy_lints/src/partial_pub_fields.rs
+++ b/clippy_lints/src/partial_pub_fields.rs
@@ -18,17 +18,17 @@ declare_clippy_lint! {
     /// ### Example
     /// ```rust
     /// pub struct Color {
-    ///     pub r,
-    ///     pub g,
-    ///     b,
+    ///     pub r: u8,
+    ///     pub g: u8,
+    ///     b: u8,
     /// }
     /// ```
     /// Use instead:
     /// ```rust
     /// pub struct Color {
-    ///     pub r,
-    ///     pub g,
-    ///     pub b,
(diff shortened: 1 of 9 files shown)


tests/ui/partial_pub_fields.rs
```diff
@@ -0,0 +1,20 @@
+#![allow(unused)]
+#![warn(clippy::partial_pub_fields)]
+
+fn main() {
+    // test code goes here
```
llogiq, line 5: This comment is not really helpful. Perhaps remove it?

tests/ui/partial_pub_fields.rs
```diff
@@ -0,0 +1,20 @@
...
+        files: HashMap<String, u32>,
+        pub paths: HashMap<u32, String>,
+    }
+
+    pub struct Color {
+        pub r: u8,
+        pub g: u8,
+        b: u8,
+    }
+}
```
llogiq, line 20: I'd like to see some tuple structs, too.

llogiq (approved):
This looks good in general, but I think there should be more tests, at least another set with tuple structs (e g. `struct Toople(pub bool, usize);`) and perhaps one test with a field named `r#pub` just because we can 😁.

TennyZhuang:
More tests were added and the CI passed

clippy_lints/src/partial_pub_fields.rs
```diff
@@ -0,0 +1,81 @@
...
+        let ItemKind::Struct(ref st, _) = item.kind else {
+            return;
+        };
+
+        let mut fields = st.fields().iter();
+        let Some(first_field) = fields.next() else {
+            // Empty struct.
+            return;
+        };
+        let all_pub = first_field.vis.kind.is_pub();
```
llogiq, line 52: Using the first field to decide whether all should be private or public is but one possible heuristic. We could also count public vs. private fields and use the first field as a tie-breaker.

That way we'd have less lint messages, too.

llogiq (commented):
I just had another look at the lint heuristic. It's not bad per se, but it might be improved further. I'm ok with merging as it is and doing a follow-up later, though. What do you think?

clippy_lints/src/partial_pub_fields.rs
TennyZhuang: In fact we return here, so there will not be too many messages.

Using the first field is more obvious for users I guess. If you think a more heuristic one should be used, I’d prefer assume all fields should be private. ADTs are much more than Data in most projects.

IMHO, the count is too random to users.

llogiq:
Thank you!

@bors r+

TennyZhuang:
I'll add a similar lint `partial_pub_variants` for enums.

PR #9662, opened 2022-10-16
`result_large_err` show largest variants in err msg

### LINKED ISSUE #9538: result_large_err include largest enum variant
### What it does

Right now `clippy::result_large_err` returns something along the following lines

```text
error: the `Err`-variant returned from this function is very large
  --> maps/src/geom.rs:11:54
   |
11 | pub fn func_name(a: f64) -> Result<f64, Error> {
   |                                                    ^^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 152 bytes
   |
```

However, it does not specify which variant of `Error` contributes to the large size. This can make finding the variant to include Box'd fields annoying.

### Advantage

- Improves code refactoring speed

### Drawbacks

- A larger error message

ebobrow:
fixes #9538 

changelog: Sugg: [`result_large_err`]: Now show largest enum variants in error message

### DIFF
--- a/clippy_lints/src/functions/result.rs
+++ b/clippy_lints/src/functions/result.rs
@@ -2,12 +2,12 @@ use rustc_errors::Diagnostic;
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LintContext};
 use rustc_middle::lint::in_external_macro;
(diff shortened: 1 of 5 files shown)


dswij (approved):
LGTM, thanks for this!

PR #9664, opened 2022-10-17
#9231 suggest TryFrom when truncation possible

### LINKED ISSUE #9231: Have a lint against usize-to-u64 casts (or, against *all* integer casts)
### What it does

I would like clippy to lint against *all* integer casts. So I have set:
```rust
#![warn(
    clippy::cast_possible_wrap, // unsigned -> signed
    clippy::cast_sign_loss, // signed -> unsigned
    clippy::cast_lossless,
    clippy::cast_possible_truncation,
)]
```
However, I just by accident noticed that this does *not* lint against usize-to-u64 casts. I guess `cast_possible_truncation` says "this cannot truncate because we don't have more than 64bit pointer size", and "cast_lossless" says "ah this might be lossy on platforms with pointers larger than 64bit", and then neither of them does anything.

I would be happy to either have one of these lints also trigger on usize-to-u64 casts, or to have a new lint against *all* integer casts.

### Lint Name

cast_integer

### Category

pedantic

### Advantage

Integer casts are subtle and should be done via `From`/`TryFrom`, never via `as`, so I want to rule out *all* of them in my codebase.

### Drawbacks

_No response_

### Example

```rust
pub fn foo(x: usize) -> u64 { x as u64 }
```

Could be written as:

```rust
pub fn foo(x: usize) -> u64 { u64::try_from(x).unwrap() }
```

navh:
changelog: [`cast_possible_truncation`]: add suggestion to use TryFrom

fixes #9231


kraktus:
The truncation can be intentional and then suggesting `tryFrom` would change the behaviour of the code, I don't think the suggestion can be machine applicable.

clippy_lints/src/casts/cast_possible_truncation.rs
```diff
@@ -159,5 +161,18 @@ pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, cast_expr: &Expr<'_>,
...
+    let snippet = snippet_with_applicability(cx, expr.span, "x", &mut applicability);
+    let name_of_cast_from = snippet.split(" as").next().unwrap_or("x");
+
+    span_lint_and_sugg(
+        cx,
+        CAST_POSSIBLE_TRUNCATION,
+        expr.span,
+        &msg,
+        "avoid silent truncation by using",
+        format!("{cast_to}::try_from({name_of_cast_from}).unwrap()"),
```
kraktus, line 175: I think using using `expect` here and hinting developers to explain why it shouldn't truncate would be even better
```suggestion
        format!("{cast_to}::try_from({name_of_cast_from}).expect("truncation should be impossible because ...)"),
```

navh: I've made this change and changed MachineApplicable to HasPlaceholders, is this the best way to classify this?

I tried to come up with some sort of `u32::try_from({} & {cast_to}::MAX as {cast_from}).unwrap()` machine applicable suggestion, but they all bury the intention of the lint.

kraktus: `HasPlaceholders` is fine here 👍 Another option now that I think about it would be to make the whole `expect` reason a placeholder so that the user wouldn't be able to copy-paste the offered suggestion without having to think about the issue before getting a compilable version of their program. What do you think @xFrednet? 
```rust
format!("{cast_to}::try_from({name_of_cast_from}).expect(..)
```

xFrednet: Thank you for the review and feedback @kraktus, it's highly appreciated!

Using `try_from()` changes the semantics of the code. In most cases, users would most likely prefer the value to be truncated over the code panicking in production. This also misses cases, where this behavior might be intentional. Though, looking at the lint description, it's not ideal to emit a lint without any suggestion.

The idea of using `try_from` is good and should be included in the lint message. The applicability of `HasPlaceholders` seamed a bit high to me at first, but good at a second thought.

That being said, I would suggest a different lint message. Something like this:

```rs
error: casting `f64` to `isize` may truncate the value
  --> $DIR/cast.rs:29:5
   |
LL |     1f64 as isize;
   |
help: if this is intensional allow the lint with `#[allow(clippy::cast_precision_loss)]` ...
help: ... or use `try_from` and handle the error accordingly
   |
LL | isize::try_from(1f64)
   |
```

Note that this changes the semantics even more, as we don't instruct the user to use `.expect("..")` or `.unwrap()`. For this, I would make the applicability unspecified.

What are your thoughts about this? :upside_down_face: 

---

You can create more complex messages with: `span_lint_and_then`.

<details>
<summary>Here is a mockup how the code could look:</summary>
```rs
span_lint_and_then(
    cx,
    <LINT>,
    <span>,
    <msg>,
    |diag| {
        diag.help(
            "if this is intensional allow the lint with `#[allow(clippy::cast_precision_loss)]` ...",
        );
        diag.span_suggestion_with_style(
            <span>,
            "... or use `try_from` and handle the error accordingly",
            suggestion,
            Applicability::Unspecified,
            // always show the suggestion in a separate line
            SuggestionStyle::ShowAlways,
        );
    },
);
``` 
</details>

xFrednet: This is correct, i mixed up the formatting. You could try add the #[allow] attribute as another code suggestion, but I think this is better. Only thing is the note in the middle which we sadly can influence directly. 

There is sadly no detailed documentation for the formatting.

Could you also update the lint description to mention these two solutions?

navh: Yes, I put 'why is this bad' back to how it was, and appended a bit to the 'what it does' to reflect both the ignore and try_from suggestions.

xFrednet:
Hey, I'll try to take a look at this soon, by mid next week at the latest.

A general FYI, Clippy has a no merge commit policy (with a few exceptions), PR branches should be rebased instead. Here is an info link on how to [resolve merge conflicts instead](https://rustc-dev-guide.rust-lang.org/git.html#rebasing-and-conflicts). Sometimes it can be simpler to squash all commits into one. I'm mentioning this, as your branch tries to add a bunch of external and merge commits. We can also take a look at this after the next review. You're welcome to reach out if you need help with that :)

navh:
>  Sometimes it can be simpler to squash all commits into one. 

I should have taken this advice, instead I've force pushed a ton.... Documentation changes got lost in the crossfire, so I've re-done it, hopefully this works.

clippy_lints/src/casts/mod.rs
```diff
@@ -80,7 +80,8 @@ declare_clippy_lint! {
     /// ### What it does
     /// Checks for casts between numerical types that may
     /// truncate large values. This is expected behavior, so the cast is `Allow` by
-    /// default.
+    /// default. It suggests user either explicitly ignore the lint,
+    /// or use `try_from()` and handle the truncation, default, or panic explicitly.
```
xFrednet, line 84: Could you also add a `Use instead:` block below, like this:

````md
Use instead:
```
fn as_u8(x: u64) -> u8 {
    if let Ok(x) = u8::try_from(x) {
        x
    } else {
        todo!();
    }
}
// Or
fn as_u16(x: u64) -> u16 {
    #[allow(clippy::cast_possible_truncation)]
    x as u16
}
```
````

xFrednet (commented):
This version already looks good to me. Two small adjustments and then we can have this merged :). Thank you for the quick updates.

xFrednet:
Hey @navh, this is a ping from triage. Are you stuck on anything, that I can help with?

@rustbot author

xFrednet:
I'm closing this in favor of https://github.com/rust-lang/rust-clippy/pull/10038 I tried to keep the original commiter from this PR :)

PR #8692, opened 2022-04-12
fix unnecessary_to_owned about msrv

kyoto7250:
This PR fixes ``[`unnecessary_owned`]``.

## What

```rust
# sample code
fn _msrv_1_35() {
    #![clippy::msrv = "1.35"]
    let _ = &["x"][..].to_vec().into_iter();
}

fn _msrv_1_36() {
    #![clippy::msrv = "1.36"]
    let _ = &["x"][..].to_vec().into_iter();
}
```

If we will check this code using clippy, ``[`unnecessary_owned`]`` will modify the code as follows.

```rust
error: unnecessary use of `to_vec`
  --> $DIR/unnecessary_to_owned.rs:219:14
   |
LL |     let _ = &["x"][..].to_vec().into_iter();
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `["x"][..].iter().copied()`

error: unnecessary use of `to_vec`
  --> $DIR/unnecessary_to_owned.rs:224:14
   |
LL |     let _ = &["x"][..].to_vec().into_iter();
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `["x"][..].iter().copied()`
```

This is incorrect. Because `Iterator::copied` was estabilished in 1.36.

## Why

This bug was caused by not separating "copied" and "clone" by reference to msrv.

https://github.com/rust-lang/rust-clippy/blob/89ee6aa6e32df974ef9dbb3c825afbdce441e856/clippy_lints/src/methods/unnecessary_to_owned.rs#L195

So, I added a conditional branch and described the corresponding test.

Thank you in advance.

changelog: fix wrong suggestions about msrv in [`unnecessary_to_owned`]

r! @giraffate

PR #9667, opened 2022-10-17
add new lint `seek_to_start_instead_of_rewind `

### LINKED ISSUE #8600: Prefer `Seek#rewind()` over `Seek#seek(SeekFrom::Start(0))`
### What it does

Suggests using the helper shortcut method, which is shorter.

https://doc.rust-lang.org/stable/std/io/trait.Seek.html#method.rewind

### Lint Name

_No response_

### Category

_No response_

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
fn foo<T: io::Seek>(t: T) {
    t.seek(io::SeekFrom::Start(0));
}
```

Could be written as:

```rust
fn foo<T: io::Seek>(t: T) {
    t.rewind();
}
```

dorublanzeanu:
changelog: `seek_to_start_instead_of_rewind`: new lint to suggest using `rewind` instead of `seek` to start

Resolve #8600


clippy_lints/src/methods/rewind_instead_of_seek_to_start.rs
```diff
@@ -0,0 +1,34 @@
...
+use rustc_lint::LateContext;
+use rustc_span::Span;
+
+use super::REWIND_INSTEAD_OF_SEEK_TO_START;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, arg: &'tcx Expr<'_>, name_span: Span) {
+    if_chain! {
+        if let ExprKind::Call(func, args1) = arg.kind;
+        if let ExprKind::Path(ref qpath1) = func.kind;
+        if match_qpath(qpath1, &["SeekFrom", "Start"]);
```
giraffate, line 15: Can `match_qpath` be replaced with `match_def_path`? The code below will be helpful:
https://github.com/rust-lang/rust-clippy/blob/502e87c3790059e878868125985726c065b72f85/clippy_lints/src/create_dir.rs#L39-L40

clippy_lints/src/methods/mod.rs
```diff
@@ -3066,6 +3067,37 @@ declare_clippy_lint! {
...
+    /// Use instead:
+    /// ```rust
+    /// # use std::io;
+    /// fn foo<T: io::Seek>(t: &mut T) {
+    ///     t.rewind();
+    /// }
+    /// ```
+    #[clippy::version = "1.66.0"]
+    pub REWIND_INSTEAD_OF_SEEK_TO_START,
+    pedantic,
```
giraffate, line 3097: It would be better to change to `complexity`.

clippy_lints/src/methods/rewind_instead_of_seek_to_start.rs
```diff
@@ -0,0 +1,34 @@
...
+        if args1.len() == 1;
+        if let ExprKind::Lit(ref lit) = args1[0].kind;
+        if let LitKind::Int(0, LitIntType::Unsuffixed) = lit.node;
```
giraffate, line 18: Can the check that receiver type implements `std::io::Seek` trait be added? The method below will be helpful:
https://github.com/rust-lang/rust-clippy/blob/502e87c3790059e878868125985726c065b72f85/clippy_utils/src/ty.rs#L150

clippy_lints/src/methods/rewind_instead_of_seek_to_start.rs
```diff
@@ -0,0 +1,34 @@
...
+use rustc_ast::ast::{LitIntType, LitKind};
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::Span;
+
+use super::REWIND_INSTEAD_OF_SEEK_TO_START;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, arg: &'tcx Expr<'_>, name_span: Span) {
+    if_chain! {
```
giraffate, line 12: Can `if_chain!` be replaced with let-chains?
```rs
    if let ExprKind::Call(func, args1) = arg.kind &&
        let ExprKind::Path(ref qpath1) = func.kind &&
        ...
```

clippy_lints/src/methods/rewind_instead_of_seek_to_start.rs
```diff
@@ -0,0 +1,34 @@
...
+            let method_call_span = expr.span.with_lo(name_span.lo());
+            span_lint_and_then(
+                cx,
+                REWIND_INSTEAD_OF_SEEK_TO_START,
+                method_call_span,
+                "used `seek` to go to the start of the stream",
+                |diag| {
+                    let app = Applicability::MachineApplicable;
+
+                    diag.span_suggestion(method_call_span, "...can be replaced by", "rewind()", app);
```
giraffate, line 29: nits
```suggestion
                    diag.span_suggestion(method_call_span, "replaced with", "rewind()", app);
```

clippy_lints/src/methods/rewind_instead_of_seek_to_start.rs
dorublanzeanu: I implemented this suggestion by using `get_trait_def_id(cx, &paths::STD_IO_SEEK)`.
A different approach would have been to define a new symbol(`IoSeek`) similar to [https://github.com/rust-lang/rust/blob/master/compiler/rustc_span/src/symbol.rs#L216](url) and just use `get_diagnostic_item(sym::IoSeek)`, but that implied a change in the `rust` repo also, right?

giraffate: > but that would have implied a change in the rust repo also, right?

yes

giraffate (commented):
Sorry, I forgot `msrv` in the first review. The [`rewind`](https://doc.rust-lang.org/stable/std/io/trait.Seek.html#method.rewind) was implemented at 1.55.0, so this need to implement `msrv`: https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv

The PR below would be helpful to implement `msvr`: https://github.com/rust-lang/rust-clippy/pull/8692

giraffate (approved):

giraffate:
I'll merge this once the conflicts are resolved.

Alexendoo:
Jumping in late here but I think the lint name should be `manual_rewind` or similar, it's currently reversed as it's saying what to do rather than what the issue is

dorublanzeanu:
It depends on how you look at it, but now that you pointed it out, it looks odd.
`manual_rewind` does not seem clearer to me as `allow/deny manual_rewind` seems to be doing something else. What do you say if I reverse it to `seek_to_start_instead_of_rewind` ? It is a bit long, but to me is clearer.

No worries about the conflicts 😄

Alexendoo:
Yeah that's fine by me

giraffate:
@Alexendoo Thanks for your review!

giraffate:
@bors r+

Thanks!

PR #9670, opened 2022-10-17
Add `missing_trait_methods` lint

### LINKED ISSUE #9661: Add a lint for unintentional usage of default trait methods
### What it does

Fires when a trait implementation relies on a default method. This is primarily useful for "delegating" trait implementations that want to do a bit of work in addition to a lower implementation

### Lint Name

unintentional_default_trait_methods

### Category

restriction

### Advantage

If a new default trait is added, it will be more obvious that we need to update our code to account for that method.

### Drawbacks

There may be new default methods that we _do_ want to rely on. I don't actually know if it's possible, but if the lint could be parameterized to ignore specific default methods, that would be nice.

### Example

```rust
trait ExampleTrait {
    fn alpha(&mut self) -> bool {
        false
    }
    
    fn omega(&mut self) -> bool {
        false
    }
}

struct OuterImplementation<T> {
    inner: T,
    count: usize,
}

// This is bad; we don't know what methods `T` implements.
// If a new default function is added to `ExampleTrait`, we
// won't be doing our additional work.
impl<T: ExampleTrait> ExampleTrait for OuterImplementation<T> {
    fn alpha(&mut self) -> bool {
        self.count += 1;
        self.inner.alpha()
    }
}
```

If `#[deny(unintentional_default_trait_methods)]` was added to the `impl ExampleTrait for OuterImplementation` line, then an error like

```
This implementation of `ExampleTrait` uses the default trait method `omega`, but it should be deliberately implemented
```

There could even be a fix-it that adds the method stub.

Alexendoo:
Closes #9661

changelog: new lint: [`missing_trait_methods`]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4049,6 +4049,7 @@ Released 2018-09-13
 [`missing_panics_doc`]: https://rust-lang.github.io/rust-clippy/master/index.html#missing_panics_doc
 [`missing_safety_doc`]: https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc
 [`missing_spin_loop`]: https://rust-lang.github.io/rust-clippy/master/index.html#missing_spin_loop
+[`missing_trait_methods`]: https://rust-lang.github.io/rust-clippy/master/index.html#missing_trait_methods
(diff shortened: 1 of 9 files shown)


Jarcho:
This probably shouldn't lint impls from external macros. Otherwise LGTM.

shepmaster:
Thanks!

PR #9671, opened 2022-10-17
[`check_result_unit_err`]: Disable on `no_std`

FawazTirmizi:
This lint suggests returning a type implementing the trait `std::error::Error` ([doc](https://doc.rust-lang.org/std/error/trait.Error.html)) for the `Result`, but this trait is not available in `no_std` crates, so it's not possible for them to fix this lint.

This is my first time working in Clippy's codebase, so I may have missed something. Please let me know and I'll fix it.

## Example:
Running Clippy on a project with the following `lib.rs`:
```rust
#![no_std]

use core::result::Result;

pub fn ret_err() -> Result<(),()> {
    if true { Ok(()) } else { Err(()) }
}
```
will yield the following warning:
```
warning: this returns a `Result<_, ()>`
 --> src/lib.rs:6:1
  |
6 | pub fn ret_err() -> Result<(),()> {
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: use a custom `Error` type instead
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_unit_err
  = note: `#[warn(clippy::result_unit_err)]` on by default
```

After the fix, Clippy passes.

---

changelog: [`check_result_unit_err`]: No longer warns on `no_std` projects

### DIFF
--- a/clippy_lints/src/functions/result.rs
+++ b/clippy_lints/src/functions/result.rs
@@ -6,8 +6,8 @@ use rustc_middle::ty::{self, Ty};
 use rustc_span::{sym, Span};
 
 use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_then};
-use clippy_utils::trait_ref_of_method;
 use clippy_utils::ty::{approx_ty_size, is_type_diagnostic_item};
+use clippy_utils::{is_no_std_crate, trait_ref_of_method};
 
 use super::{RESULT_LARGE_ERR, RESULT_UNIT_ERR};
 
@@ -71,7 +71,7 @@ pub(super) fn check_trait_item<'tcx>(cx: &LateContext<'tcx>, item: &hir::TraitIt
 }
 
 fn check_result_unit_err(cx: &LateContext<'_>, err_ty: Ty<'_>, fn_header_span: Span) {
-    if err_ty.is_unit() {
+    if err_ty.is_unit() && !is_no_std_crate(cx) {
         span_lint_and_help(
             cx,
             RESULT_UNIT_ERR,


FawazTirmizi:
> In my eyes, the lint as implemented seems to be correct, even on no_std, but the rationale given by the lint documentation is wrong. On one hand, () is almost never useful as a Result's E type. On the other hand, this lint doesn't actually seem to target using types that don't implement Error at all. ... In my opinion, since MyError is a local type, this should result in a different lint proclaiming that MyError should implement Error (missing_error_impls?) ...

This makes sense. Is this something you'd also like to see in this PR?

> ... changing result_unit_err to lint on the use of any E: !Error in Result could produce unresolveable lints when working with error types from other crates. Because of this, result_unit_err's docs/suggestion should be modified to remove references to the Error trait.

So are you saying that `result_unit_err` should instead suggest simply returning a result where `E != ()`?

> As an aside, it's probably worth noting that on nightly, the Error trait is available in [core::error::Error](https://doc.rust-lang.org/nightly/core/error/trait.Error.html) with the error_in_core feature gate.

Thanks, I was unaware of this.

PatchMixolydic:
> This makes sense. Is this something you'd also like to see in this PR?

I should probably emphasize that I'm just a bystander, not a Clippy maintainer :sweat_smile: Nonetheless, I feel that a `missing_error_impls` lint would be big enough to warrant a separate PR.

> So are you saying that `result_unit_err` should instead suggest returning a `Result` where `E != ()`?

Technically, it already does, though perhaps the suggestion could be reworded slightly:
```diff
  warning: this returns a `Result<_, ()>`
   --> src/lib.rs:6:1
    |
  6 | pub fn ret_err() -> Result<(),()> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
-   = help: use a custom `Error` type instead
+   = help: use a custom error type instead
    = help: for further information visit https://rust-lang.github.io/rust- 
 clippy/master/index.html#result_unit_err
    = note: `#[warn(clippy::result_unit_err)]` on by default
```

I feel that the primary issue is that the documentation focuses on the `Error` trait when the lint doesn't really have much to do with it:

https://github.com/rust-lang/rust-clippy/blob/09e4659a862f8a244fedf473a06b9a85448d3a17/clippy_lints/src/functions/mod.rs#L171-L178

Ideally, this would be reworded to discuss how a custom error type, especially an `enum`, could allow one to provide more information about why the error occurred. The fact that a custom error type can implement `Error` is certainly useful, but probably shouldn't be highlighted as the focus of this lint.

FawazTirmizi:
> > This makes sense. Is this something you'd also like to see in this PR?
> 
> I should probably emphasize that I'm just a bystander, not a Clippy maintainer sweat_smile ...

oops :P

>  Nonetheless, I feel that a `missing_error_impls` lint would be big enough to warrant a separate PR.

Agreed.

> 
> > So are you saying that `result_unit_err` should instead suggest returning a `Result` where `E != ()`?
> 
> Technically, it already does, though perhaps the suggestion could be reworded slightly:
> 
> ```diff
>   warning: this returns a `Result<_, ()>`
>    --> src/lib.rs:6:1
>     |
>   6 | pub fn ret_err() -> Result<(),()> {
>     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     |
> -   = help: use a custom `Error` type instead
> +   = help: use a custom error type instead
>     = help: for further information visit https://rust-lang.github.io/rust- 
>  clippy/master/index.html#result_unit_err
>     = note: `#[warn(clippy::result_unit_err)]` on by default
> ```
> 
> I feel that the primary issue is that the documentation focuses on the `Error` trait when the lint doesn't really have much to do with it:
> 
> https://github.com/rust-lang/rust-clippy/blob/09e4659a862f8a244fedf473a06b9a85448d3a17/clippy_lints/src/functions/mod.rs#L171-L178
> 
> Ideally, this would be reworded to discuss how a custom error type, especially an `enum`, could allow one to provide more information about why the error occurred. The fact that a custom error type can implement `Error` is certainly useful, but probably shouldn't be highlighted as the focus of this lint.

That makes a lot more sense, thanks for the clarification. I'll wait until a maintainer chimes in, but what you're suggesting makes more sense than what I had put out. Thanks!

giraffate:
Thanks for the discussion!

> I feel that the primary issue is that the documentation focuses on the Error trait when the lint doesn't really have much to do with it:

That looks good to me. If `no_std`, it might be better to suggest `core::error::Error` with feature gate. I would suggest creating an issue first and then a PR.

Then, can this issue be closed?

FawazTirmizi:
That's fine, we can continue this discussion on the issue.

Issue #9111, opened 2022-07-03
needless_borrow false positive in case when compiler does not automatically borrow


leighmcculloch:
### Summary

The needless_borrow lint has a false positive where the suggested fix results in code that won't compile, because the compiler doesn't automatically borrow, even though the lint says it does.

### Lint Name

needless_borrow

### Reproducer

I tried this code:

```rust
#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
pub struct VecM<T, const MAX: u32 = { u32::MAX }>(Vec<T>);

impl<T: Clone, const N: usize, const MAX: u32> TryFrom<&[T; N]> for VecM<T, MAX> {
    type Error = ();

    fn try_from(v: &[T; N]) -> Result<Self, ()> {
        todo!()
    }
}

#[cfg(test)]
mod test {
    use crate::VecM;

    #[test]
    fn test_try_from() {
        let _vm: VecM<u8, 32> = (&[]).try_into().unwrap();
        //let _vm: VecM<u8, 32> = [].try_into().unwrap();
    }
}
```

```
❯ cargo clippy --all-targets
    Checking clippy-needless-borrow v0.1.0 (/Users/leighmcculloch/Code/clippy-needless-borrow)
warning: this expression borrows a value the compiler would automatically borrow
  --> src/lib.rs:23:33
   |
18 |         let _vm: VecM<u8, 32> = (&[]).try_into().unwrap();
   |                                 ^^^^^ help: change this to: `[]`
   |
   = note: `#[warn(clippy::needless_borrow)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

warning: `clippy-needless-borrow` (lib test) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
```

Making the change that clippy suggests, results in the following and the following compiler error:

```rust
#[derive(Debug, Clone, Hash, PartialEq, Eq, PartialOrd, Ord)]
pub struct VecM<T, const MAX: u32 = { u32::MAX }>(Vec<T>);

impl<T: Clone, const N: usize, const MAX: u32> TryFrom<&[T; N]> for VecM<T, MAX> {
    type Error = ();

    fn try_from(v: &[T; N]) -> Result<Self, ()> {
        todo!()
    }
}

#[cfg(test)]
mod test {
    use crate::VecM;

    #[test]
    fn test_try_from() {
[...]

PR #9674, opened 2022-10-18
Fix `needless_borrow` false positive

smoelius:
The PR fixes the false positive exposed by @BusyJay's example in: https://github.com/rust-lang/rust-clippy/issues/9111#issuecomment-1277114280

The current approach is described in https://github.com/rust-lang/rust-clippy/pull/9674#issuecomment-1289294201 and https://github.com/rust-lang/rust-clippy/pull/9674#issuecomment-1292225232.

The original approach appears below.

---

The proposed fix is to flag only "simple" trait implementations involving references, a concept
that I introduce next.

Intuitively, a trait implementation is "simple" if all it does is dereference and apply the trait
implementation of a type named by a type parameter. `AsRef` provides a good example of a simple
implementation: https://doc.rust-lang.org/std/convert/trait.AsRef.html#impl-AsRef%3CU%3E-for-%26T

We can make this idea more precise as follows. Given a trait implementation, first determine
whether the implementation is "used defined." If so, then examine its nested obligations.
Consider the implementation simple if-and-only-if:
- there is at least one nested obligation for the same trait
- for each type `X` in the nested obligation's substitution, either `X` is the same as that of
  the original obligation's substitution, or the original type is `&X`

For example, the following implementation from @BusyJay's example is "complex" (i.e., not simple)
because it produces no nested obligations:

```rust
impl<'a> Extend<&'a u8> for A { ... }
```

On the other hand, the following slightly modified implementation is simple, because it produces
a nested obligation for `Extend<X>`:

```rust
impl<'a, X> Extend<&'a X> for A where A: Extend<X> { ... }
```

How does flagging only simple implementations help? One way of interpreting the false positive in
@BusyJay's example is that it separates a reference from a concrete type. Doing so turns a
successful type inference into a failing one. By flagging only simple implementations, we
separate references from type variables only, thereby eliminating this class of false positives.

Note that `Deref` is a special case, as the obligations generated for it already involve the
underlying type.

r? @Jarcho (Sorry to keep pinging you with `needless_borrow` stuff. But my impression is no one knows this code better than you.)

changelog: fix `needless_borrow` false positive

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -1049,7 +1049,7 @@ fn ty_contains_infer(ty: &hir::Ty<'_>) -> bool {
 // If the conditions are met, returns `Some(Position::ImplArg(..))`; otherwise, returns `None`.
 //   The "is copyable" condition is to avoid the case where removing the `&` means `e` would have to
 // be moved, but it cannot be.
-#[expect(clippy::too_many_arguments)]
+#[expect(clippy::too_many_arguments, clippy::too_many_lines)]
 fn needless_borrow_impl_arg_position<'tcx>(
     cx: &LateContext<'tcx>,
     possible_borrowers: &mut Vec<(LocalDefId, PossibleBorrowerMap<'tcx, 'tcx>)>,
@@ -1092,7 +1092,7 @@ fn needless_borrow_impl_arg_position<'tcx>(
         .iter()
         .filter_map(|predicate| {
             if let PredicateKind::Trait(trait_predicate) = predicate.kind().skip_binder()
-                && trait_predicate.trait_ref.self_ty() == param_ty.to_ty(cx.tcx)
+                && trait_predicate.self_ty() == param_ty.to_ty(cx.tcx)
             {
                 Some(trait_predicate.trait_ref.def_id)
             } else {
@@ -1107,6 +1107,16 @@ fn needless_borrow_impl_arg_position<'tcx>(
                 || Some(trait_def_id) == sized_trait_def_id
                 || cx.tcx.is_diagnostic_item(sym::Any, trait_def_id)
         })
+    {
+        return Position::Other(precedence);
+    }
+
+    // See:
+    // - https://github.com/rust-lang/rust-clippy/pull/9674#issuecomment-1289294201
+    // - https://github.com/rust-lang/rust-clippy/pull/9674#issuecomment-1292225232
+    if projection_predicates
+        .iter()
+        .any(|projection_predicate| is_mixed_projection_predicate(cx, callee_def_id, projection_predicate))
     {
         return Position::Other(precedence);
     }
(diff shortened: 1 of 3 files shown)


Jarcho:
Requiring nested obligations is overly restrictive here. The problem for the given example comes from switching which impl is being used, essentially causing a different function to be called.

In the example, `extend` has thjree different generic parameters. `Self` and `A` from the `Extend` trait; and `T` from the function itself. I think the problem comes from seeing the predicate `T: IntoIterator<Item = A>`, which causes the `Item` type from the new `T` to be propagated to `A`. This shouldn't happen as `A` was introduced in the surrounding context, not the current function.

The fix would be to check that the target of the type propagation is introduced by the current function. Should be as easy as checking if the parameter index is greater than the number of parameters in the parent generics list.

smoelius:
> The problem for the given example comes from switching which impl is being used, essentially causing a different function to be called.

I agree with this 100%.

> In the example, `extend` has thjree different generic parameters. `Self` and `A` from the `Extend` trait; and `T` from the function itself. I think the problem comes from seeing the predicate `T: IntoIterator<Item = A>`, which causes the `Item` type from the new `T` to be propagated to `A`. This shouldn't happen as `A` was introduced in the surrounding context, not the current function.
> 
> The fix would be to check that the target of the type propagation is introduced by the current function. Should be as easy as checking if the parameter index is greater than the number of parameters in the parent generics list.

I may be misunderstanding your solution, but wouldn't that allow examples like this to get through?
```rust
#[derive(Clone, Copy)]
struct A;

trait Foo {
    fn bar(self);
}

impl Foo for &A {
    fn bar(self) {
        println!("okay");
    }
}

impl Foo for A {
    fn bar(self) {
        panic!();
    }
}

fn call_bar(foo: impl Foo) {
    foo.bar();
}

fn main() {
    call_bar(&A);
}
```

Jarcho:
That is a pretty pathological case. It requires writing a trait impl for some type and it's reference which do different things. This is a thing which is already surprising. Also note you can do the same thing even with a nested obligation.

```rust
trait Tr { fn foo() }
impl<T> Tr for T { fn foo() { panic!() } }
impl<'a, T: 'a + Tr> Tr for &'a T { fn foo() { call_some_other_fn() } }
```

smoelius:
> That is a pretty pathological case.

I agree with you it's pathological. But, to be honest, I would consider any false positive that doesn't involve nested obligations to be pathological. Would you disagree?

For example, if you run `lintcheck` with the configuration from 332b5b3d8b883087ea1632bf7cfb13721db6d9f9, you get the same 22 "the borrowed expression implements the required traits" warnings with and without the nested obligation check. In other words, the nested obligation check does not eliminate any of the warnings seen in those crates.

Jarcho:
The lack of nested obligations will come into play when the impl for the reference isn't generic. e.g.

```rust
trait Tr {}
impl Tr for u8 {}
impl<'a> Tr &'a u8 {}
```

Here the impl on the reference has no nested obligation, but will still most likely be equivalent to the non-reference impl. This is also the only way to implement this for foreign traits on local types.

smoelius:
I'm not convinced, but I'll adjust the PR to implement your fix.

Thank you for your patient explanation.

smoelius:
I think the most recent commit implements your idea in https://github.com/rust-lang/rust-clippy/pull/9674#issuecomment-1289294201. Please let me know if I misunderstood.

Jarcho:
That looks like your blocking any `T:  Trait` predicate where `T` is from the parent generics list. This will block any function in the trait `trait T: Bound {}`.

The predicates that need to be blocked are `U: Trait<Assoc = T>` and `U::Assoc: Trait<Assoc = T>` where `U` is a type introduced on the current function, and `T` is a type introduced by the parent context.

smoelius:
Is this more like what you had in mind?

Jarcho:
That looks like it. Thank you.

@bors r+

Issue #9647, opened 2022-10-14
collapsible_match suggestion does not compile


haata:
### Summary

collapisble_match warning gives an incorrect suggestion for code that cannot compile. Even with some tweaks, I haven't been able to compress the statement.

I suspect the syn crate may be doing something interesting, but I haven't investigated it deeply.

### Lint Name

collapsible_match

### Reproducer

I tried this code:

```rust
use syn::{parse, Field, Type, Expr, Result, ExprLit, Lit, TypePath};

pub fn parse_type(field: &Field, ft: Type) -> Result<(TypePath, usize)> {
    match ft {
        Type::Array(a) => {
            let mut size: usize = 0;

            if let Expr::Lit(ExprLit { lit, .. }) = a.len {
                if let  = lit {
                    if let Ok(num) = lit.base10_parse::<usize>() {
                        size = num;
                    }
                }
            }
            if size == 0 {
                Err(
                    parse::Error::new(field.ident.as_ref().unwrap().span(), "`#[gen_hid_descriptor]` array has invalid length")
                )
            } else {
                Ok((parse_type(field, *a.elem)?.0, size))
            }
        }
        Type::Path(p) => {
            Ok((p, 1))
        }
        _ => {
            Err(
                parse::Error::new(field.ident.as_ref().unwrap().span(),"`#[gen_hid_descriptor]` cannot handle field type")
            )
        }
    }
}
```

I saw this happen:

```
warning: this `if let` can be collapsed into the outer `if let`
  --> src/main.rs:13:17
   |
13 | /                 if let Lit::Int(lit) = lit {
14 | |                     if let Ok(num) = lit.base10_parse::<usize>() {
15 | |                         size = num;
16 | |                     }
17 | |                 }
   | |_________________^
   |
help: the outer pattern can be modified to include the inner pattern
  --> src/main.rs:12:40
   |
12 |             if let Expr::Lit(ExprLit { lit, .. }) = a.len {
   |                                        ^^^ replace this binding
13 |                 if let Lit::Int(lit) = lit {
   |                        ^^^^^^^^^^^^^ with this pattern
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_match
   = note: `#[warn(clippy::collapsible_match)]` on by default

warning: `collapsible_match` (bin "collapsible_match") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
```

I changed the code to the suggestion:

```rust
[...]

PR #9685, opened 2022-10-21
[`collapsible_match`] specify field name when destructuring structs

kraktus:
changelog: [`collapsible_match`] specify field name when destructuring structs

fix https://github.com/rust-lang/rust-clippy/issues/9647

I wasn't the sure about the best way to convey the message in the lint message since it does not use suggestion. Because I liked the former output highlighting both spans, I've left it as before, only modifying the span label.


llogiq:
Thank you!

@bors r+

Issue #9612, opened 2022-10-08
`unwrap_used`/`expect_used` fire in test module function regardless of `allow-unwrap-in-tests`/`allow-expect-in-tests`


Firionus:
### Summary

Functions inside modules with `#[cfg(test)]` that are themselves not `#[test]` functions are linted for `unwrap_used` and `expect_used` even when `clippy.toml` sets `allow-unwrap-in-tests = true` and `allow-expect-in-tests = true`.

### Lint Name

unwrap_used & expect_used

### Reproducer

I tried this code:

main.rs
```rust
#![warn(clippy::unwrap_used, clippy::expect_used)]

fn main() {
    // warns as expected
    let _a: u8 = 0.try_into().unwrap();
    let _a: u8 = 1.try_into().expect("");
}

#[cfg(test)]
mod tests {
    #[test]
    fn test_fn() {
        // doesn't warn, as expected
        let _a: u8 = 2.try_into().unwrap();
        let _a: u8 = 3.try_into().expect("");

        util();
    }

    fn util() {
        // BUG: warns even though it is in test module
        let _a: u8 = 4.try_into().unwrap();
        let _a: u8 = 5.try_into().expect("");
    }
}
```
clippy.toml
```toml
allow-unwrap-in-tests = true
allow-expect-in-tests = true
```

I saw this happen:

```
$ cargo clippy --tests
warning: used `unwrap()` on `a Result` value
 --> src/main.rs:5:18
  |
5 |     let _a: u8 = 0.try_into().unwrap();
  |                  ^^^^^^^^^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/main.rs:1:9
  |
1 | #![warn(clippy::unwrap_used, clippy::expect_used)]
  |         ^^^^^^^^^^^^^^^^^^^
  = help: if you don't want to handle the `Err` case gracefully, consider using `expect()` to provide a better panic message
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_used

warning: used `expect()` on `a Result` value
 --> src/main.rs:6:18
  |
6 |     let _a: u8 = 1.try_into().expect("");
  |                  ^^^^^^^^^^^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/main.rs:1:30
  |
1 | #![warn(clippy::unwrap_used, clippy::expect_used)]
  |                              ^^^^^^^^^^^^^^^^^^^
[...]

PR #9686, opened 2022-10-21
[`unwrap_used`], [`expect_used`] do not lint in `test` cfg

kraktus:
changelog: [`unwrap_used`], [`expect_used`] do not lint in `test` cfg

fix https://github.com/rust-lang/rust-clippy/issues/9612

I've updated the doc and used `cfg` acronym, not sure if `conditional compiler flag` would have been better

### DIFF
--- a/clippy_lints/src/utils/conf.rs
+++ b/clippy_lints/src/utils/conf.rs
@@ -373,7 +373,7 @@ define_Conf! {
     (max_include_file_size: u64 = 1_000_000),
     /// Lint: EXPECT_USED.
     ///
-    /// Whether `expect` should be allowed in test cfg
+    /// Whether `expect` should be allowed within `#[cfg(test)]`
     (allow_expect_in_tests: bool = false),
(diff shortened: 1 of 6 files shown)


clippy_lints/src/utils/conf.rs
```diff
@@ -373,11 +373,11 @@ define_Conf! {
     (max_include_file_size: u64 = 1_000_000),
     /// Lint: EXPECT_USED.
     ///
-    /// Whether `expect` should be allowed in test functions
+    /// Whether `expect` should be allowed in test cfg
```
llogiq, line 376:
```suggestion
    /// Whether `expect` should be allowed within `#[cfg(test)]`
```

llogiq:
I'd make the config docs more clear by just naming cfg(test) directly. Otherwise this looks good.

flip1995 (approved):

flip1995:
@bors r+

Thanks!

PR #9690, opened 2022-10-22
add `from_raw_with_void_ptr` lint

### LINKED ISSUE #9679: Warn about `Box::from_raw::<c_void>`
### What it does

Often in FFI, some APIs will take `*c_void` or `*mut c_void` to indicate arbitrary objects from the caller side. When interacting with such APIs in Rust it is common to allocate something with `{Box,Rc,Arc}::into_raw` and then casting the raw pointer. For deallocating you often get the `c_void` pointer back from the FFI API and you would turn it into the corresponding owned/shared pointer with `{Box,Rc,Arc}::from_raw`, after which you can drop it.

However, it is easy to run into the pitfall of calling `from_raw` with the `c_void` pointer. Note that the definition of, say, `Box::from_raw` is:
```rust
pub unsafe fn from_raw(raw: *mut T) -> Box<T>
```

meaning that if you pass a `*mut c_void` you will get a `Box<c_void>`. Per the safety requirements in the documentation, for this to be safe, `c_void` would need to have the same memory layout as the original type, which is often not the case. And even if it was the case, if the original type or one of its fields (or one of _its_ fields, recursively) had a `Drop` implementation, that implementation would not run when this newly recreated box is dropped. (And similarly for `Rc` and `Arc`.)

While it is possible to purposefully create a `Box<c_void>`, it's hard to see any case where it would be useful. Therefore, uses of `Box::from_raw::<c_void>` can be safely assumed to be a mistake, and clippy should have a lint for it.

### Lint Name

from-raw-cvoid

### Category

_No response_

### Advantage

- Remove an easy pitfall of FFI, particularly with some C or C++ APIs.

### Drawbacks

- Might conflict with genuine usages of `Box<c_void>`, but those are incredibly rare, and not good practice.

### Example

```rust
use std::ffi::c_void;

let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
// pass `ptr` through C FFI or something.
let _ = unsafe { Box::from_raw(ptr) };
```

Could be written as:

```rust
use std::ffi::c_void;

let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
// pass `ptr` through C FFI or something.
let _ = unsafe { Box::from_raw(ptr as *mut usize) };
```

royrustdev:
This PR `fixes #9679` 

- \[x] Run `cargo dev fmt`

---

changelog:  [`from_raw_with_void_ptr`]: added new lint

### DIFF
--- a/clippy_lints/src/from_raw_with_void_ptr.rs
+++ b/clippy_lints/src/from_raw_with_void_ptr.rs
@@ -1,0 +1,57 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::path_def_id;
+use clippy_utils::ty::is_c_void;
+use rustc_hir::{Expr, ExprKind, QPath};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::RawPtr;
+use rustc_middle::ty::TypeAndMut;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks if we're passing a `c_void` raw pointer to `Box::from_raw(_)`
+    ///
+    /// ### Why is this bad?
+    /// However, it is easy to run into the pitfall of calling from_raw with the c_void pointer.
+    /// Note that the definition of, say, Box::from_raw is:
+    ///
+    /// `pub unsafe fn from_raw(raw: *mut T) -> Box<T>`
+    ///
+    /// meaning that if you pass a *mut c_void you will get a Box<c_void>.
+    /// Per the safety requirements in the documentation, for this to be safe,
+    /// c_void would need to have the same memory layout as the original type, which is often not the case.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # use std::ffi::c_void;
+    /// let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
+    /// let _ = unsafe { Box::from_raw(ptr) };
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # use std::ffi::c_void;
+    /// # let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
+    /// let _ = unsafe { Box::from_raw(ptr as *mut usize) };
+    /// ```
+    ///
+    #[clippy::version = "1.66.0"]
+    pub FROM_RAW_WITH_VOID_PTR,
+    suspicious,
+    "creating a `Box` from a raw void pointer"
+}
+declare_lint_pass!(FromRawWithVoidPtr => [FROM_RAW_WITH_VOID_PTR]);
+
+impl LateLintPass<'_> for FromRawWithVoidPtr {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_from_raw, [arg]) = expr.kind
+        && let ExprKind::Path(QPath::TypeRelative(ty, seg)) = box_from_raw.kind
+        && seg.ident.name == sym!(from_raw)
+        // FIXME: This lint is also applicable to other types, like `Rc`, `Arc` and `Weak`.
+        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box())
+        && let arg_kind = cx.typeck_results().expr_ty(arg).kind()
+        && let RawPtr(TypeAndMut { ty, .. }) = arg_kind
+        && is_c_void(cx, *ty) {
+            span_lint_and_help(cx, FROM_RAW_WITH_VOID_PTR, expr.span, "creating a `Box` from a raw void pointer", Some(arg.span), "cast this to a pointer of the actual type");
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3903,6 +3903,7 @@ Released 2018-09-13
 [`format_push_string`]: https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string
 [`from_iter_instead_of_collect`]: https://rust-lang.github.io/rust-clippy/master/index.html#from_iter_instead_of_collect
 [`from_over_into`]: https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into
+[`from_raw_with_void_ptr`]: https://rust-lang.github.io/rust-clippy/master/index.html#from_raw_with_void_ptr
 [`from_str_radix_10`]: https://rust-lang.github.io/rust-clippy/master/index.html#from_str_radix_10
 [`future_not_send`]: https://rust-lang.github.io/rust-clippy/master/index.html#future_not_send
 [`get_first`]: https://rust-lang.github.io/rust-clippy/master/index.html#get_first
--- a/clippy_lints/src/lib.register_all.rs
+++ b/clippy_lints/src/lib.register_all.rs
@@ -81,6 +81,7 @@ store.register_group(true, "clippy::all", Some("clippy_all"), vec![
     LintId::of(formatting::SUSPICIOUS_ELSE_FORMATTING),
     LintId::of(formatting::SUSPICIOUS_UNARY_OP_FORMATTING),
     LintId::of(from_over_into::FROM_OVER_INTO),
+    LintId::of(from_raw_with_void_ptr::FROM_RAW_WITH_VOID_PTR),
     LintId::of(from_str_radix_10::FROM_STR_RADIX_10),
     LintId::of(functions::DOUBLE_MUST_USE),
     LintId::of(functions::MUST_USE_UNIT),
--- a/clippy_lints/src/lib.register_lints.rs
+++ b/clippy_lints/src/lib.register_lints.rs
@@ -173,6 +173,7 @@ store.register_lints(&[
     formatting::SUSPICIOUS_ELSE_FORMATTING,
     formatting::SUSPICIOUS_UNARY_OP_FORMATTING,
     from_over_into::FROM_OVER_INTO,
+    from_raw_with_void_ptr::FROM_RAW_WITH_VOID_PTR,
     from_str_radix_10::FROM_STR_RADIX_10,
     functions::DOUBLE_MUST_USE,
     functions::MUST_USE_CANDIDATE,
--- a/clippy_lints/src/lib.register_suspicious.rs
+++ b/clippy_lints/src/lib.register_suspicious.rs
@@ -21,6 +21,7 @@ store.register_group(true, "clippy::suspicious", Some("clippy_suspicious"), vec!
     LintId::of(formatting::SUSPICIOUS_ASSIGNMENT_FORMATTING),
     LintId::of(formatting::SUSPICIOUS_ELSE_FORMATTING),
     LintId::of(formatting::SUSPICIOUS_UNARY_OP_FORMATTING),
+    LintId::of(from_raw_with_void_ptr::FROM_RAW_WITH_VOID_PTR),
     LintId::of(loops::EMPTY_LOOP),
     LintId::of(loops::MUT_RANGE_BOUND),
     LintId::of(methods::NO_EFFECT_REPLACE),
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -230,6 +230,7 @@ mod format_impl;
 mod format_push_string;
 mod formatting;
 mod from_over_into;
+mod from_raw_with_void_ptr;
 mod from_str_radix_10;
 mod functions;
 mod future_not_send;
@@ -918,6 +919,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|_| Box::new(implicit_saturating_add::ImplicitSaturatingAdd));
     store.register_early_pass(|| Box::new(partial_pub_fields::PartialPubFields));
     store.register_late_pass(|_| Box::new(missing_trait_methods::MissingTraitMethods));
+    store.register_late_pass(|_| Box::new(from_raw_with_void_ptr::FromRawWithVoidPtr));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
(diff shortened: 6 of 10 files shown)


clippy_lints/src/boxed_void.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// ### Example
+    /// ```rust
+    /// use std::ffi::c_void;
+    /// let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
+    /// let _ = unsafe { Box::from_raw(ptr) };
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// use std::ffi::c_void;
+    /// let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
```
flip1995, line 26:
```suggestion
    /// # use std::ffi::c_void;
    /// # let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
```

I would hide those duplicated lines in the documentation.

clippy_lints/src/boxed_void.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl LateLintPass<'_> for BoxedVoid {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_from_raw, [arg]) = expr.kind
+        && let ExprKind::Path(QPath::TypeRelative(ty, seg)) = box_from_raw.kind
+        && seg.ident.name == sym!(from_raw)
+        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box()) {
+            let arg_kind = cx.typeck_results().expr_ty(arg).kind();
+            if let RawPtr(TypeAndMut { ty, .. }) = arg_kind {
+                if is_c_void(cx, *ty) {
```
flip1995, line 44: You can also include irrefutable patterns in the `if let` chain, so this can all be in one chain:

```suggestion
        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box())
        && let arg_kind = cx.typeck_results().expr_ty(arg).kind()
        && if let RawPtr(TypeAndMut { ty, .. }) = arg_kind
        && if is_c_void(cx, *ty) {
```

clippy_lints/src/boxed_void.rs
```diff
@@ -0,0 +1,50 @@
...
+use rustc_middle::ty::RawPtr;
+use rustc_middle::ty::TypeAndMut;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks if we're passing a `c_void` raw pointer to `Box::from_raw(_)`
+    ///
+    /// ### Why is this bad?
+    /// This is a known pitfall of FFI, particularly with some C or C++ APIs.
```
flip1995, line 15: Can you link to some documentation where this pitfall is documented? If you can't find any official documentation, please explain this pitfall in more detail here.

clippy_lints/src/boxed_void.rs
```diff
@@ -0,0 +1,50 @@
...
+impl LateLintPass<'_> for BoxedVoid {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_from_raw, [arg]) = expr.kind
+        && let ExprKind::Path(QPath::TypeRelative(ty, seg)) = box_from_raw.kind
+        && seg.ident.name == sym!(from_raw)
+        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box()) {
+            let arg_kind = cx.typeck_results().expr_ty(arg).kind();
+            if let RawPtr(TypeAndMut { ty, .. }) = arg_kind {
+                if is_c_void(cx, *ty) {
+                span_lint(cx, BOXED_VOID, expr.span, "don't create a Box from a raw void pointer");
```
flip1995, line 45: I would word the lint message a bit different: ``"creating a `Box` from a raw void pointer"`` and then add a suggestion (with `span_lint_and_sugg`) to tell the user to cast the pointer first.

flip1995 (changes requested):

clippy_lints/src/boxed_void.rs
kraktus: The issue's OP made a good description:

> However, it is easy to run into the pitfall of calling `from_raw` with the `c_void` pointer. Note that the definition of, say, `Box::from_raw` is:
>
> ```rust 
> pub unsafe fn from_raw(raw: *mut T) -> Box<T>
> ```
> meaning that if you pass a *mut c_void you will get a `Box<c_void>`. Per the safety requirements in the documentation, for this to be safe, c_void would need to have the same memory layout as the original type, which is often not the case. And even if it was the case, if the original type or one of its fields (or one of its fields, recursively) had a `Drop` implementation, that implementation would not run when this newly recreated box is dropped. (And similarly for `Rc` and `Arc`.)

clippy_lints/src/boxed_void.rs
```diff
@@ -0,0 +1,50 @@
...
+    /// let _ = unsafe { Box::from_raw(ptr) };
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// use std::ffi::c_void;
+    /// let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
+    /// let _ = unsafe { Box::from_raw(ptr as *mut usize) };
+    /// ```
+    #[clippy::version = "1.66.0"]
+    pub BOXED_VOID,
```
kraktus, line 30: I would maybe make the name more specific, maybe `RAW_BOX_FROM_CVOID_PTR` since `void` and `unit` are often used interchangeably, and we're not talking about the `()` type here

clippy_lints/src/boxed_void.rs
llogiq: We cannot know what type to cast to in general. So the suggestion could at best contain a placeholder, which might be confusing. Wouldn't some help text be better suited here?

flip1995: We could suggest `as *mut _` with applicability `HasPlaceholder`. But I agree that this might be confusing. So I'm also fine with just a help message.

clippy_lints/src/boxed_void.rs
flip1995: SGTM, but I would drop the `C` from the same, so just `RAW_BOX_FROM_VOID_PTR`

clippy_lints/src/boxed_void.rs
royrustdev: Do I mention the issue link in `Why is this bad?` ?

flip1995: No, please put this explanation into the documentation. This will save Clippy users a click.

clippy_lints/src/boxed_void.rs
royrustdev: Done commenting these

clippy_lints/src/boxed_void.rs
royrustdev: Refactored into a single if let chain

```rust
    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
        if let ExprKind::Call(box_from_raw, [arg]) = expr.kind
        && let ExprKind::Path(QPath::TypeRelative(ty, seg)) = box_from_raw.kind
        && seg.ident.name == sym!(from_raw)
        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box())
        && let arg_kind = cx.typeck_results().expr_ty(arg).kind()
        && let RawPtr(TypeAndMut { ty, .. }) = arg_kind
        && is_c_void(cx, *ty) {
            span_lint_and_help(cx, RAW_BOX_FROM_VOID_PTR, expr.span, "creating a `Box` from a raw void pointer", Some(arg.span), "cast this to the actual type");
        }
    }
```
I had to modify your suggestion a little
```rust
// changed this line
&& if let RawPtr(TypeAndMut { ty, .. }) = arg_kind
```
```rust
// into this
&& let RawPtr(TypeAndMut { ty, .. }) = arg_kind
```

clippy_lints/src/boxed_void.rs
```diff
@@ -0,0 +1,50 @@
...
+    "don't create a Box from a raw void pointer"
+}
+declare_lint_pass!(BoxedVoid => [BOXED_VOID]);
+
+impl LateLintPass<'_> for BoxedVoid {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_from_raw, [arg]) = expr.kind
+        && let ExprKind::Path(QPath::TypeRelative(ty, seg)) = box_from_raw.kind
+        && seg.ident.name == sym!(from_raw)
+        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box()) {
```
andreubotella, line 41: This should probably cover `Rc` and `Arc`, not only box. Same with the test.

clippy_lints/src/boxed_void.rs
royrustdev: Renamed the lint as suggested.

clippy_lints/src/boxed_void.rs
royrustdev: Is documentation now correct ? Need help

```rust
declare_clippy_lint! {
    /// ### What it does
    /// Checks if we're passing a `c_void` raw pointer to `Box::from_raw(_)`
    ///
    /// ### Why is this bad?
    /// However, it is easy to run into the pitfall of calling from_raw with the c_void pointer.
    /// Note that the definition of, say, Box::from_raw is:
    ///
    /// ```rust
    /// pub unsafe fn from_raw(raw: *mut T) -> Box<T>
    /// ```
    /// meaning that if you pass a *mut c_void you will get a Box<c_void>.
    /// Per the safety requirements in the documentation, for this to be safe,
    /// c_void would need to have the same memory layout as the original type, which is often not the case.
    /// And even if it was the case, if the original type or one of its fields (or one of its fields, recursively) had a Drop implementation, that implementation would not run when this newly recreated box is dropped.
    ///
    /// ### Example
    /// ```rust
    /// #use std::ffi::c_void;
    /// #let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
    /// let _ = unsafe { Box::from_raw(ptr) };
    /// ```
    /// Use instead:
    /// ```rust
    /// #use std::ffi::c_void;
    /// #let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
    /// let _ = unsafe { Box::from_raw(ptr as *mut usize) };
    /// ```
    ///
    #[clippy::version = "1.66.0"]
    pub RAW_BOX_FROM_VOID_PTR,
    suspicious,
    "creating a `Box` from a raw void pointer"
}
```

clippy_lints/src/boxed_void.rs
royrustdev: Updated into this

```rust
span_lint_and_help(cx, RAW_BOX_FROM_VOID_PTR, expr.span, "creating a `Box` from a raw void pointer", Some(arg.span), "cast this to the actual type");
```

clippy_lints/src/boxed_void.rs
flip1995: I would leave out the `And even if it was the case, ...` line. Otherwise LGTM

clippy_lints/src/boxed_void.rs
flip1995: I would write `cast this to a pointer of the actual type`.

flip1995 (commented):
The impl LGTM now. But this lint could potentially also apply to ``Rc`` ``Arc`` and ``Weak``, not only box. This can also be done in a future PR though, so I don't want to block this PR on it. However, I would chose the lint name, so that applying this lint to other types is possible in the future.

Possible names:
- ``from_raw_with_void_ptr``
- ``from_raw_from_void_ptr``
- ...? Better ideas?

royrustdev:
@flip1995 is the PR okay now ? Please let me know.

clippy_lints/src/from_raw_with_void_ptr.rs
```diff
@@ -0,0 +1,56 @@
...
+    "creating a `Box` from a raw void pointer"
+}
+declare_lint_pass!(FromRawWithVoidPtr => [FROM_RAW_WITH_VOID_PTR]);
+
+impl LateLintPass<'_> for FromRawWithVoidPtr {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if let ExprKind::Call(box_from_raw, [arg]) = expr.kind
+        && let ExprKind::Path(QPath::TypeRelative(ty, seg)) = box_from_raw.kind
+        && seg.ident.name == sym!(from_raw)
+        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box())
```
flip1995, line 49:
```suggestion
        // FIXME: This lint is also applicable to other types, like `Rc`, `Arc` and `Weak`.
        && path_def_id(cx, ty).map_or(false, |id| Some(id) == cx.tcx.lang_items().owned_box())
```

Please add this fixme comment. After that, this should be good to go.

royrustdev: Okay.

clippy_lints/src/from_raw_with_void_ptr.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// `pub unsafe fn from_raw(raw: *mut T) -> Box<T>`
+    ///
+    /// meaning that if you pass a *mut c_void you will get a Box<c_void>.
+    /// Per the safety requirements in the documentation, for this to be safe,
+    /// c_void would need to have the same memory layout as the original type, which is often not the case.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # use std::ffi::c_void;
+    /// # let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
```
flip1995, line 27: This line should not be commented out in the first code block
```suggestion
    /// let ptr = Box::into_raw(Box::new(42usize)) as *mut c_void;
```

royrustdev: Understood

flip1995 (approved):

flip1995:
@bors r+

Thanks!

PR #9691, opened 2022-10-22
Add `lintcheck` to packages linted by `dogfood` test

smoelius:
Currently, `lintcheck` is not checked by the `dogfood` test. I assume that is not intentional. If it is intentional, then this PR can be ignored.

changelog: Add `lintcheck` to packages linted by `dogfood` test


llogiq:
I think the fallout changes show conclusively that this is a good idea.

@flip1995 sorry for stealing the review. 😁

@bors r+

PR #9692, opened 2022-10-22
make ignored internally mutable types for `mutable-key` configurable

### LINKED ISSUE #5325: mutable_key_type false positives
Hello!

We're getting false positives with the `mutable_key_type` rule: https://circleci.com/gh/libra/libra/41319?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

It seems to be triggered by this:

```
HashSet<ProtocolId>
```

where [ProtocolId](https://github.com/libra/libra/blob/07be050b658dd8d4402ef877cc0ac47eaff92c57/network/src/protocols/wire/messaging/v1/mod.rs#L31) is defined as 

```
pub type ProtocolId = protocols::wire::messaging::v1::ProtocolId;
```

which is

```
#[repr(u8)]
#[derive(Clone, Copy, Debug, Hash, Eq, PartialEq, Deserialize_repr, Serialize_repr)]
pub enum ProtocolId {
    ConsensusRpc = 0,
    ConsensusDirectSend = 1,
    MempoolDirectSend = 2,
    StateSynchronizerDirectSend = 3,
    DiscoveryDirectSend = 4,
    HealthCheckerRpc = 5,
    IdentityDirectSend = 6,
}
```

I wasn't able to reproduce it with a simple example, so not sure why clippy is getting triggered here.

llogiq:
We had some false positives where people would create their own types that had interior mutability unrelated to hash/eq. This addition lets you configure this as e.g. `arc-like-types=["bytes::Bytes"]`

This fixes #5325 by allowing users to specify the types whose innards like `Arc` should be ignored (the generic types are still checked) for the sake of detecting inner mutability.

r? @Alexendoo

---

changelog: Allow configuring types to ignore internal mutability in `mutable-key`

### DIFF
--- a/clippy_lints/src/mut_key.rs
+++ b/clippy_lints/src/mut_key.rs
@@ -1,10 +1,12 @@
 use clippy_utils::diagnostics::span_lint;
-use clippy_utils::trait_ref_of_method;
+use clippy_utils::{def_path_res, trait_ref_of_method};
+use rustc_data_structures::fx::FxHashSet;
 use rustc_hir as hir;
+use rustc_hir::def::Namespace;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::TypeVisitable;
 use rustc_middle::ty::{Adt, Array, Ref, Slice, Tuple, Ty};
-use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::source_map::Span;
 use rustc_span::symbol::sym;
 use std::iter;
@@ -78,26 +80,44 @@ declare_clippy_lint! {
     "Check for mutable `Map`/`Set` key type"
 }
 
-declare_lint_pass!(MutableKeyType => [ MUTABLE_KEY_TYPE ]);
+#[derive(Clone)]
+pub struct MutableKeyType {
+    ignore_interior_mutability: Vec<String>,
+    ignore_mut_def_ids: FxHashSet<hir::def_id::DefId>,
+}
+
+impl_lint_pass!(MutableKeyType => [ MUTABLE_KEY_TYPE ]);
 
 impl<'tcx> LateLintPass<'tcx> for MutableKeyType {
+    fn check_crate(&mut self, cx: &LateContext<'tcx>) {
+        self.ignore_mut_def_ids.clear();
+        let mut path = Vec::new();
+        for ty in &self.ignore_interior_mutability {
+            path.extend(ty.split("::"));
+            if let Some(id) = def_path_res(cx, &path[..], Some(Namespace::TypeNS)).opt_def_id() {
+                self.ignore_mut_def_ids.insert(id);
+            }
+            path.clear();
+        }
+    }
+
     fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::Item<'tcx>) {
         if let hir::ItemKind::Fn(ref sig, ..) = item.kind {
-            check_sig(cx, item.hir_id(), sig.decl);
+            self.check_sig(cx, item.hir_id(), sig.decl);
         }
     }
 
     fn check_impl_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::ImplItem<'tcx>) {
         if let hir::ImplItemKind::Fn(ref sig, ..) = item.kind {
             if trait_ref_of_method(cx, item.def_id.def_id).is_none() {
-                check_sig(cx, item.hir_id(), sig.decl);
+                self.check_sig(cx, item.hir_id(), sig.decl);
             }
         }
     }
 
     fn check_trait_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::TraitItem<'tcx>) {
         if let hir::TraitItemKind::Fn(ref sig, ..) = item.kind {
-            check_sig(cx, item.hir_id(), sig.decl);
+            self.check_sig(cx, item.hir_id(), sig.decl);
         }
     }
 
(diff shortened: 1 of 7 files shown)


clippy_lints/src/utils/conf.rs
```diff
@@ -387,6 +387,11 @@ define_Conf! {
     ///
     /// The maximum size of the `Err`-variant in a `Result` returned from a function
     (large_error_threshold: u64 = 128),
+    /// Lint::MUTABLE_KEY
+    ///
+    /// A list of paths to types that should be treated like `Arc`, i.e. ignored but
+    /// for the generic parameters for determining interior mutability
+    (arc_like_types: Vec<String> = Vec::new()),
```
Alexendoo, line 394: Seems like a reasonable escape hatch, I think the name might be better as a bit more direct, `ignore_interior_mutability` or something along those lines as there may be other types that aren't `Arc`-like

What do you think about setting the default to `vec!["bytes::Bytes"]` to cover the seemingly most popular false positive?

clippy_lints/src/mut_key.rs
```diff
@@ -78,98 +78,116 @@ declare_clippy_lint! {
...
-            if is_std_collection || is_box {
-                // The type is mutable if any of its type parameters are
-                substs.types().any(|ty| is_interior_mutable_type(cx, ty, span))
-            } else {
-                !ty.has_escaping_bound_vars()
-                    && cx.tcx.layout_of(cx.param_env.and(ty)).is_ok()
-                    && !ty.is_freeze(cx.tcx.at(span), cx.param_env)
+    // We want to lint 1. sets or maps with 2. not immutable key types and 3. no unerased
+    // generics (because the compiler cannot ensure immutability for unknown types).
+    fn check_ty_<'tcx>(&self, cx: &LateContext<'tcx>, span: Span, ty: Ty<'tcx>) {
```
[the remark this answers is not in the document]
llogiq: That clashes with `LateLintPass::check_ty`.

llogiq:
Thanks @Alexendoo for the review. I incorporated your suggestions.

clippy_lints/src/mut_key.rs
```diff
@@ -78,98 +78,119 @@ declare_clippy_lint! {
...
+            },
+            _ => false,
+        }
+    }
+
+    fn is_arc_like(&self, cx: &LateContext<'_>, did: hir::def_id::DefId) -> bool {
+        !self.ignore_interior_mutability.is_empty()
+            && match_any_def_paths_iter(cx, did, self.ignore_interior_mutability.iter().map(|s| s.split("::")))
+                .is_some()
     }
```
Alexendoo, line 195: Could use `def_path_res(..).opt_def_id()` in a `check_crate` to build a `DefIdSet` one time, similar to how `disallowed_types` does it https://github.com/rust-lang/rust-clippy/blob/191c9839f0bad1c2bfec17d55bacb94d2e83f1a1/clippy_lints/src/disallowed_types.rs#L89-L103

Then checking if a type is ignored can be a single lookup

tests/ui-toml/mut_key/clippy.toml
```diff
@@ -0,0 +1 @@
+arc-like-types = ["mut_key::Counted"]
```
Alexendoo, line 1: Needs a key name update

clippy_lints/src/mut_key.rs
llogiq: The problem here is that the queries used by `def_path_res` don't work on local items. It needed a good bunch of code to work around this. I'm not sure if this is absolutely correct (it might need additional code to cater for the namespace_hint), but it works in my case. If this becomes a problem, we can revisit it in a later PR.

clippy_lints/src/mut_key.rs
```diff
@@ -78,98 +80,130 @@ declare_clippy_lint! {
...
 impl<'tcx> LateLintPass<'tcx> for MutableKeyType {
+    fn check_crate(&mut self, cx: &LateContext<'tcx>) {
+        self.ignore_mut_def_ids.clear();
+        let mut path = Vec::new();
+        for ty in &self.ignore_interior_mutability {
+            path.extend(ty.split("::"));
+            if let Some(id) = def_path_res(cx, &path[..], Some(Namespace::TypeNS)).opt_def_id() {
+                self.ignore_mut_def_ids.insert(id);
+                //} else {
+                //    cx.sess().note_without_error(&format!("not found: {ty}"));
```
Alexendoo, line 100:
```suggestion
```

clippy_utils/src/lib.rs
```diff
@@ -642,7 +647,60 @@ pub fn def_path_res(cx: &LateContext<'_>, path: &[&str], namespace_hint: Option<
...
+                        ItemKind::TyAlias(_, _) => DefKind::TyAlias,
+                        ItemKind::OpaqueTy(_) => DefKind::OpaqueTy,
+                        ItemKind::Union(_, _) => DefKind::Union,
+                        ItemKind::Trait(_, _, _, _, _) => DefKind::Trait,
+                        ItemKind::TraitAlias(_, _) => DefKind::TraitAlias,
+                        ItemKind::Impl(_) => DefKind::Impl,
+                    };
+                    if path.is_empty() {
+                        return Res::Def(def_kind, item_def_id.to_def_id());
+                    }
```
Alexendoo, line 693: I *think* the `cx.tcx.def_kind(item_def_id)` query should work still here

clippy_utils/src/lib.rs
```diff
@@ -1811,6 +1869,22 @@ pub fn match_any_def_paths(cx: &LateContext<'_>, did: DefId, paths: &[&[&str]])
...
+) -> Option<usize>
+where
+    P: IntoIterator<Item = &'a str>,
+{
+    let search_path = cx.get_def_path(did);
+    paths
+        .into_iter()
+        .position(|p| p.into_iter().map(Symbol::intern).eq(search_path.iter().copied()))
+}
+
```
Alexendoo, line 1887: Now unused

```suggestion
```

clippy_utils/src/lib.rs
```diff
@@ -586,14 +586,19 @@ pub fn def_path_res(cx: &LateContext<'_>, path: &[&str], namespace_hint: Option<
...
-        tcx.crates(())
+    fn find_crate(cx: &LateContext<'_>, name: &str) -> Option<DefId> {
+        cx.tcx
+            .crates(())
             .iter()
             .copied()
-            .find(|&num| tcx.crate_name(num).as_str() == name)
+            .find(|&num| cx.tcx.crate_name(num).as_str() == name)
             .map(CrateNum::as_def_id)
     }
```
Alexendoo, line 596: This change looks like it became no longer necessary

clippy_utils/src/lib.rs
```diff
@@ -586,14 +586,19 @@ pub fn def_path_res(cx: &LateContext<'_>, path: &[&str], namespace_hint: Option<
...
 
+    if path.first() == Some(&"crate") || path.first().copied() == cx.sess().opts.crate_name.as_deref() {
+        return def_path_res_local(cx, &path[1..]);
+    }
```
Alexendoo, line 600: We may actually not want to handle `crate`, it would resolve to different paths for each package in the workspace, and I think also each bin/lib target within a single package

Could also move this down one statement to use `base` in place of `path.first()` to guard against any zero length path weirdness

llogiq: Ok, I have removed the `crate`, and used `base`.

clippy_utils/src/lib.rs
llogiq: Reverted.

Alexendoo:
Nice, thanks!

@bors r+

Issue #9314, opened 2022-08-09
`let_underscore_drop` and ignoring error 


uselessgoddess:
I ignore some errors with `try` block. I annotate type via `let _: Result<...> = try { ... }`. But clippy:
```
warning: non-binding `let` on a type that implements `Drop`
  --> some/path:
   |
   | /         let _: Result<_> = try {
   | |            . . .
   | |         };
   | |__________^
   |
   = help: consider using an underscore-prefixed named binding or dropping explicitly with `std::mem::drop`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_drop
```
How can I fix this

PR #9696, opened 2022-10-23
Move `let_underscore_drop` to restriction

Alexendoo:
Because of things like #9314. `let _` is considered the idiomatic way to ignore `must_use`: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute

changelog: [`let_underscore_drop`]: move to restriction

r? @flip1995

### DIFF
--- a/clippy_lints/src/let_underscore.rs
+++ b/clippy_lints/src/let_underscore.rs
@@ -62,8 +62,7 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for `let _ = <expr>`
-    /// where expr has a type that implements `Drop`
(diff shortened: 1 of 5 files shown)


flip1995 (approved):
pedantic is a weird category for this lint, so moving it to restriction SGTM.

flip1995:
@bors r+

Thanks!

Alexendoo:
@bors r-

I just remembered that this actually got uplifted, https://github.com/rust-lang/rust/pull/97739, so I'll prepare a PR that does a rename instead

flip1995:
Ah and it wasn't deprecated in Clippy. Yeah, please remove the lint from Clippy then. Now this is no longer our problem :D

PR #9697, opened 2022-10-23
Mark `let_underscore_lock` and `let_underscore_drop` as uplifted

### LINKED ISSUE #8003: let_underscore_drop complains about idiomatic `let _ = ignorable_result`;
Lint name: let_underscore_drop

I tried this code:

```rust
#![allow(unused)]
#![deny(clippy::let_underscore_drop)]

struct Test;

impl Test {
    fn causes_error(&self) -> Result<(), std::io::Error> {
        todo!()
    }
    
    fn bug(&self) {
        let _ = self.causes_error();
    }
}
```
I expected to see this happen: no clippy errors

Instead, this happened: clippy wants to keep the Result around until end of scope, while `let _ =` is the normally suggested way to deal with results where there is no useful Ok and the Err needs to be ignored.

[edit: this originally suggested that the Drop bound is wrongly inferred.]

### Meta

**Rust version:** Nightly on Playground (2021-11-19 a77da2d)

### LINKED ISSUE #9314: `let_underscore_drop` and ignoring error

Alexendoo:
Here I've renamed both the uplifted lints, however rustc's `let_underscore_lock` is slightly less capable than the clippy lint as it doesn't catch `parking_lot` types or `Result<Guard, ..>`, should we still remove it? The `Result` change looks like it was unintentional to me so that could probably be fixed upstream

changelog: Uplift [`let_underscore_drop`] to rustc https://github.com/rust-lang/rust/pull/97739
changelog: Remove overlap between rustc's `let_underscore_lock` and Clippy's [`let_underscore_lock`]

Closes #8003
Closes #9314

r? @flip1995


flip1995:
Should we limit the ``let_underscore_lock`` lint to just lint on `parking_lot` locks? I wouldn't completely deprecate it.

flip1995 (approved):
Thanks!

Issue #9687, opened 2022-10-21
`#[allow(clippy::fn_params_excessive_bools)]` doesn't work on methods


xFrednet:
### Description

Lint: [`clippy::fn_params_excessive_bools`](https://rust-lang.github.io/rust-clippy/master/index.html#fn_params_excessive_bools)

`#[allow(clippy::fn_params_excessive_bools)]` works on functions and impl blocks, but not on individual methods inside an impl blocks. The problem seams to be, that methods are checked from the impl block, which doesn't process the lint attributes on special functions. See: 

https://github.com/rust-lang/rust-clippy/blob/967f172e256802f9bf0ad5718d9c761da18132a9/clippy_lints/src/excessive_bools.rs#L162-L171

### Example:

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=237ab00916c4568ea8d3b6de8c50c75b)

```rs
#![deny(clippy::fn_params_excessive_bools)]
#![allow(dead_code)]

pub struct Test;

impl Test {
    #[allow(clippy::fn_params_excessive_bools)]
    fn new(
        is_const: bool,
        is_async: bool,
        is_unsafe: bool,
        is_extern: bool,
        is_awesome: bool,
        is_special: bool,
    ) -> Self {
        let _b = is_const || is_async || is_unsafe || is_extern || is_awesome || is_special;
        todo!()
    }
}
```

### Version

```text
rustc 1.66.0-nightly (9062b780b 2022-09-21)
binary: rustc
commit-hash: 9062b780b32d2eab060b4432863e085d9504ca5c
commit-date: 2022-09-21
host: x86_64-unknown-linux-gnu
release: 1.66.0-nightly
LLVM version: 15.0.0
```

PR #9698, opened 2022-10-23
 [`fn_params_excessive_bools`] Make it possible to allow the lint at the method level

kraktus:
changelog: FP: [`fn_params_excessive_bools`]: `#[allow]` now works on methods

fix https://github.com/rust-lang/rust-clippy/issues/9687

Tested without committing but `#[allow]`ing now works. Also rewrote the lint to be a late lint while at it :)
r? @xFrednet

### DIFF
--- a/tests/ui/fn_params_excessive_bools.rs
+++ b/tests/ui/fn_params_excessive_bools.rs
@@ -2,6 +2,7 @@
 #![allow(clippy::too_many_arguments)]
 
 extern "C" {
+    // Should not lint, most of the time users have no control over extern function signatures
     fn f(_: bool, _: bool, _: bool, _: bool);
 }
 
--- a/tests/ui/fn_params_excessive_bools.rs
+++ b/tests/ui/fn_params_excessive_bools.rs
@@ -35,6 +35,8 @@ impl S {
 }
 
 impl Trait for S {
+    // Should not lint because the trait might not be changeable by the user
+    // We only lint in the trait definition
     fn f(_: bool, _: bool, _: bool, _: bool) {}
     fn g(_: bool, _: bool, _: bool, _: Vec<u32>) {}
 }
--- a/tests/ui/fn_params_excessive_bools.rs
+++ b/tests/ui/fn_params_excessive_bools.rs
@@ -23,10 +23,12 @@ fn t(_: S, _: S, _: Box<S>, _: Vec<u32>, _: bool, _: bool, _: bool, _: bool) {}
 
 struct S;
 trait Trait {
+    // should warn for trait functions with and without body
     fn f(_: bool, _: bool, _: bool, _: bool);
     fn g(_: bool, _: bool, _: bool, _: Vec<u32>);
     #[allow(clippy::fn_params_excessive_bools)]
     fn h(_: bool, _: bool, _: bool, _: bool, _: bool, _: bool);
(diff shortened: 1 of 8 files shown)


xFrednet:
Sorry, I missed this over the week, it has been pretty chaotic. I should have a bit more time over the coming weeks. This is on my priority list for this weekend, so you'll have a review by Sunday. You can still rebase :)

tests/ui/fn_params_excessive_bools.stderr
```diff
@@ -8,31 +8,31 @@ LL | fn g(_: bool, _: bool, _: bool, _: bool) {}
...
-LL |     fn f(_: bool, _: bool, _: bool, _: bool);
-   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+LL |     fn f(&self, _: bool, _: bool, _: bool, _: bool) {}
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider refactoring bools into two-variant enums
 
 error: more than 3 bools in function parameters
-  --> $DIR/fn_params_excessive_bools.rs:30:5
+  --> $DIR/fn_params_excessive_bools.rs:38:5
```
xFrednet, line 27: This is a report on a trait implementation which might not be changeable by the user. Therefore, I would expect this to be suppressed.

The lint should instead warn the on the trait declaration. It previously did that, but now it doesn't. See line 26

xFrednet (commented):
Excellent work! Only one remark regarding traits. Thank you!

Oh and as said, you're welcome to commit the test for the ticket as well, just to be sure that it never comes up again :)

tests/ui/fn_params_excessive_bools.stderr
kraktus: nice catch! I missed that subtlety

tests/ui/fn_params_excessive_bools.rs
```diff
@@ -24,6 +25,8 @@ struct S;
 trait Trait {
     fn f(_: bool, _: bool, _: bool, _: bool);
```
xFrednet, line 26: This no longer produces a warning. My guess is that trait functions like this one don't run `check_fn` as that function requires a body. That might also be the reason, why the previous implementation iterated over the members of a module/impl block.

You can take a look at [check_trait_item](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/passes/trait.LateLintPass.html#method.check_trait_item). That might help. Otherwise, you might have to iterate over the elements of a trait declaration and emit the lint using [span_lint_hir](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/diagnostics/fn.span_lint_hir.html#) which emits the lint at a given HIR node :upside_down_face:

If you implement either solution, could you also add a test(in the trait) with a default body like this:

```rs
fn i(_: bool, _: bool, _: bool, _: bool) {};
```

I would expect a warning for this one :upside_down_face:

kraktus: Another nice catch and guidance, thanks!

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the fix!

@bors r+

PR #9704, opened 2022-10-24
[`use_self`] fix FP when trait impl defined in macro

kraktus:
changelog: [`use_self`] fix FP when trait impl defined in macro


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #9712, opened 2022-10-25
Remove `lib.register_*` and `src/docs*` in `cargo dev update_lints`

Alexendoo:
Follow up to #9709 / #9541

There's a good number of PRs with some leftover `src/docs` files for example, and as a reviewer it's something we're used to ignoring so it can easily slip through

r? @flip1995 

changelog: none

### DIFF
--- a/clippy_dev/src/update_lints.rs
+++ b/clippy_dev/src/update_lints.rs
@@ -36,6 +36,60 @@ pub enum UpdateMode {
 pub fn update(update_mode: UpdateMode) {
     let (lints, deprecated_lints, renamed_lints) = gather_all();
     generate_lint_files(update_mode, &lints, &deprecated_lints, &renamed_lints);
+    remove_old_files(update_mode);
+}
+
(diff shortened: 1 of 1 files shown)


clippy_dev/src/update_lints.rs
```diff
@@ -36,6 +36,57 @@ pub enum UpdateMode {
 pub fn update(update_mode: UpdateMode) {
     let (lints, deprecated_lints, renamed_lints) = gather_all();
     generate_lint_files(update_mode, &lints, &deprecated_lints, &renamed_lints);
+    remove_old_files(update_mode);
+}
+
+/// Remove files no longer needed after <https://github.com/rust-lang/rust-clippy/pull/9541>
+/// that may be reintroduced unintentionally
+fn remove_old_files(update_mode: UpdateMode) {
```
flip1995, line 44: Please add a fixme comment, that this should only exist temporarily, and should be removed once no PRs that add those files exist anymore

PR #9714, opened 2022-10-25
Fix `bool_to_int_with_if` false positive with `if let`

### LINKED ISSUE #9706: Incorrect code suggestion when using if-let expression
### Summary

Code suggestion causes code not to compile because the suggestion assumes if-let expressions work where normal conditions do, which is not the case.

### Reproducer

```rust
enum Foo {
    Bar,
}

fn bug(x: Foo) -> i32 {
    if let Foo::Bar = x {
        1
    } else {
        0
    }
}
```

Clippy suggests the code below which does not compile.

```rust
enum Foo {
    Bar,
}

fn bug(x: Foo) -> i32 {
    i32::from(let Foo::Bar = x)
}
```

`cargo check` output:
```rust
error: expected expression, found `let` statement
   |
   |     i32::from(let Foo::Bar = x)
   |               ^^^

error: `let` expressions are not supported here
   |
   |     i32::from(let Foo::Bar = x)
   |               ^^^^^^^^^^^^^^^^
   |
   = note: only supported directly in conditions of `if` and `while` expressions

error[E0658]: `let` expressions in this position are unstable
   |
   |     i32::from(let Foo::Bar = x)
   |               ^^^^^^^^^^^^^^^^
   |
   = note: see issue #53667 <https://github.com/rust-lang/rust/issues/53667> for more information
   = help: add `#![feature(let_chains)]` to the crate attributes to enable
```

### Version

```text
rustc 1.66.0-nightly (7fcf850d7 2022-10-23)
binary: rustc
commit-hash: 7fcf850d7942804990a1d2e3fe036622a0fe4c74
commit-date: 2022-10-23
host: x86_64-pc-windows-msvc
release: 1.66.0-nightly
LLVM version: 15.0.2
```

### Additional Labels

@rustbot label +I-suggestion-causes-error +bool_to_int_with_if

Alexendoo:
Fixes #9706

changelog: FP: [`bool_to_int_with_if`]: Now ignores `if let` statements

### DIFF
--- a/clippy_lints/src/bool_to_int_with_if.rs
+++ b/clippy_lints/src/bool_to_int_with_if.rs
@@ -1,3 +1,4 @@
+use clippy_utils::higher::If;
 use rustc_ast::LitKind;
 use rustc_hir::{Block, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
@@ -52,9 +53,9 @@ impl<'tcx> LateLintPass<'tcx> for BoolToIntWithIf {
 }
 
 fn check_if_else<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx rustc_hir::Expr<'tcx>) {
-    if let ExprKind::If(check, then, Some(else_)) = expr.kind
+    if let Some(If { cond, then, r#else: Some(r#else) }) = If::hir(expr)
         && let Some(then_lit) = int_literal(then)
-        && let Some(else_lit) = int_literal(else_)
+        && let Some(else_lit) = int_literal(r#else)
     {
         let inverted = if is_integer_literal(then_lit, 1) && is_integer_literal(else_lit, 0) {
             false
@@ -66,7 +67,7 @@ fn check_if_else<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx rustc_hir::Expr<'tcx>
         };
         let mut applicability = Applicability::MachineApplicable;
         let snippet = {
-            let mut sugg = Sugg::hir_with_applicability(cx, check, "..", &mut applicability);
+            let mut sugg = Sugg::hir_with_applicability(cx, cond, "..", &mut applicability);
             if inverted {
                 sugg = !sugg;
             }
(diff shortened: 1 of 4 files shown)


tests/ui/bool_to_int_with_if.rs
```diff
@@ -123,3 +124,22 @@ fn side_effect() {}
...
+    B,
+}
+
+fn if_let(a: Enum, b: Enum) {
+    if let Enum::A = a {
+        1
+    } else {
+        2
+    };
+
```
xFrednet, line 139: These examples should trigger the lint, as it only detects `if <expr> {1} else {0}` AFAIK.

Could you update the example, to use these literals? :upside_down_face:

clippy_lints/src/bool_to_int_with_if.rs
```diff
@@ -52,9 +53,9 @@ impl<'tcx> LateLintPass<'tcx> for BoolToIntWithIf {
 }
 
 fn check_if_else<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx rustc_hir::Expr<'tcx>) {
-    if let ExprKind::If(check, then, Some(else_)) = expr.kind
+    if let Some(If { cond, then, r#else: Some(r#else) }) = If::hir(expr)
```
xFrednet, line 56: I'm considering if it would be nicer to suggest `i32::from(matches!(X, Y))` instead of ignoring `if let` cases completely. That would be shorter in most cases, but maybe not more readable. This is just a thought, so I'm fine with this as well.

xFrednet (commented):
Once small nit and one question, otherwise looks good to me :)

tests/ui/bool_to_int_with_if.rs
Alexendoo: Whoops 😄

clippy_lints/src/bool_to_int_with_if.rs
Alexendoo: Not sure on that one, I think it's a bit too complex personally

xFrednet: Okay, then let's just ignore this case. Creating the suggestion would also not be too much fun

xFrednet (approved):

xFrednet:
Looks good to me, thank you!

Lord Bors, would you please use your magic powers to merge this PR?

Issue #4140, opened 2019-05-25
Suggestion does not compile: use self with associated types


lopopolo:
```console
$ cargo clippy -V
clippy 0.0.212 (fc96aa03 2019-05-04)
```

I have two traits that have the same associated types that are shaped like `From` and `TryFrom`. I provide a default `TryFrom` implementation for types that implement `From`. Clippy suggests I modify the code below to replace the `To::From` and `To::To` references with `Self::From` and `Self::To`, which causes a recursive overflow.

```rust
pub trait FromMrb<T> {
    type From;
    type To;

    fn from_mrb(interp: &Mrb, value: T) -> Self;
}

pub trait TryFromMrb<T>
where
    Self: Sized,
{
    type From;
    type To;

    unsafe fn try_from_mrb(interp: &Mrb, value: T) -> Result<Self, Error<Self::From, Self::To>>;
}

impl<From, To> TryFromMrb<From> for To
where
    To: FromMrb<From>,
{
    type From = To::From;
    type To = To::To;

    unsafe fn try_from_mrb(interp: &Mrb, value: From) -> Result<Self, Error<Self::From, Self::To>> {
        Ok(FromMrb::from_mrb(interp, value))
    }
}
```

```
error: unnecessary structure name repetition
  --> mruby/src/convert.rs:50:17
   |
50 |     type From = To::From;
   |                 ^^ help: use the applicable keyword: `Self`
   |
note: lint level defined here
  --> mruby/src/lib.rs:1:22
   |
1  | #![deny(clippy::all, clippy::pedantic)]
   |                      ^^^^^^^^^^^^^^^^
   = note: #[deny(clippy::use_self)] implied by #[deny(clippy::pedantic)]
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self

error: unnecessary structure name repetition
  --> mruby/src/convert.rs:51:15
   |
51 |     type To = To::To;
   |               ^^ help: use the applicable keyword: `Self`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self
```

lopopolo:
This is a more minimal and self-contained example:

```rust
#![deny(clippy::all, clippy::pedantic)]

pub struct Error<From, To> {
    _from: From,
    _too: To,
}

pub trait From<T> {
    type From;
    type To;

    fn from(value: T) -> Self;
}

pub trait TryFrom<T>
where
    Self: Sized,
{
[...]

PR #9726, opened 2022-10-26
[`use_self`] fix suggestion when full path to struct was given

kraktus:
Previously the following wrong suggestion was given

```rust
impl Error for std::fmt::Error {
    fn custom<T: std::fmt::Display>(_msg: T) -> Self {
-        std::fmt::Error // Should lint
+        Self::Error // Should lint
    }
}
```

Also remove known problem line related to #4140 since it's been closed, and refactor the lint

changelog: [`use_self`] fix suggestion when full path to struct was given


Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

PR #9728, opened 2022-10-27
move `UNINLINED_FORMAT_ARGS` to pedantic

dswij:
As discussed in zulip, we are moving this lint to pedantic to be backported

changelog: [`UNINLINED_FORMAT_ARGS`]: move to pedantic


flip1995:
https://github.com/rust-lang/rust/pull/103654

PR #9731, opened 2022-10-27
improve `lintcheck`

kraktus:
Two axes were improved (one per commit):
- Logging
- Fixing `--fix` flag

About logging, I'm open to other form and crates to use, my main point was to actually see actual clippy's suggestions, rather than for each crate look for the line

---

changelog: none

### DIFF
--- a/lintcheck/src/main.rs
+++ b/lintcheck/src/main.rs
@@ -28,6 +28,7 @@ use std::time::Duration;
 
 use cargo_metadata::diagnostic::{Diagnostic, DiagnosticLevel};
 use cargo_metadata::Message;
+use log::{debug, error, trace, warn};
 use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use walkdir::{DirEntry, WalkDir};
@@ -163,10 +164,10 @@ fn get(path: &str) -> Result<ureq::Response, ureq::Error> {
         match ureq::get(path).call() {
             Ok(res) => return Ok(res),
             Err(e) if retries >= MAX_RETRIES => return Err(e),
-            Err(ureq::Error::Transport(e)) => eprintln!("Error: {e}"),
+            Err(ureq::Error::Transport(e)) => error!("{}", e),
             Err(e) => return Err(e),
         }
-        eprintln!("retrying in {retries} seconds...");
+        warn!("retrying in {retries} seconds...");
         thread::sleep(Duration::from_secs(u64::from(retries)));
         retries += 1;
     }
@@ -234,7 +235,7 @@ impl CrateSource {
                         .expect("Failed to clone git repo!")
                         .success()
                     {
-                        eprintln!("Failed to clone {url} into {}", repo_path.display());
+                        warn!("Failed to clone {url} into {}", repo_path.display());
                     }
                 }
                 // check out the commit/branch/whatever
@@ -247,7 +248,7 @@ impl CrateSource {
                     .expect("Failed to check out commit")
                     .success()
                 {
-                    eprintln!("Failed to checkout {commit} of repo at {}", repo_path.display());
+                    warn!("Failed to checkout {commit} of repo at {}", repo_path.display());
                 }
 
                 Crate {
@@ -390,6 +391,8 @@ impl Crate {
 
         cargo_clippy_args.extend(clippy_args);
 
+        debug!("Arguments passed to cargo clippy driver: {:?}", cargo_clippy_args);
+
         let all_output = Command::new(&cargo_clippy_path)
             // use the looping index to create individual target dirs
             .env("CARGO_TARGET_DIR", shared_target_dir.join(format!("_{thread_index:?}")))
@@ -409,21 +412,19 @@ impl Crate {
         let status = &all_output.status;
 
         if !status.success() {
-            eprintln!(
-                "\nWARNING: bad exit status after checking {} {} \n",
-                self.name, self.version
-            );
+            warn!("bad exit status after checking {} {} \n", self.name, self.version);
         }
 
         if config.fix {
+            trace!("{}", stderr);
             if let Some(stderr) = stderr
                 .lines()
                 .find(|line| line.contains("failed to automatically apply fixes suggested by rustc to crate"))
             {
                 let subcrate = &stderr[63..];
-                println!(
-                    "ERROR: failed to apply some suggetion to {} / to (sub)crate {subcrate}",
-                    self.name
+                error!(
+                    "failed to apply some suggetion to {} / to (sub)crate {}",
+                    self.name, subcrate
                 );
             }
             // fast path, we don't need the warnings anyway
@@ -449,7 +450,7 @@ fn build_clippy() {
         .status()
         .expect("Failed to build clippy!");
     if !status.success() {
-        eprintln!("Error: Failed to compile Clippy!");
+        error!("Failed to compile Clippy!");
         std::process::exit(1);
     }
 }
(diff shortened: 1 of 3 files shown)


flip1995:
I'm really unsure if we need this level of logging in the `lintcheck` tool. This introduces 2 new dependencies, that I don't think help us much here. I really don't need coloring in my terminal when using the tool.

kraktus:
> I'm really unsure if we need this level of logging in the `lintcheck` tool. This introduces 2 new dependencies, that I don't think help us much here. I really don't need coloring in my terminal when using the tool.

This additional logging already allowed me to fix two issues of `use_self` that went unnoticed: https://github.com/rust-lang/rust-clippy/pull/9726 and https://github.com/rust-lang/rust-clippy/pull/9704.

About the colouring/formatting, I'd be fine with making it similar to plain `(e)println!`

For the two dependencies, I'm not sure why it is an issue since we have no restrain in how we run `lintcheck`. `log` crate is definitely trustworthy, `simplelog` also get a decent amount of download but would be ready to change it to whatever minimalistic log handler we'd like. We could even simulate its behaviour with just plain `write!`, but I feel that would really be reinventing the wheel.

Alexendoo:
I was thinking about adding another output mode alongside normal/markdown that output the rendered diagnostics to the log, would that cover it?

lintcheck/src/main.rs
```diff
@@ -337,7 +348,9 @@ impl Crate {
         let shared_target_dir = clippy_project_root().join("target/lintcheck/shared_target_dir");
 
         let mut cargo_clippy_args = if config.fix {
-            vec!["--fix", "--"]
+            // need to pass `clippy` arg even if already feeding `cargo-clippy`
+            // see https://github.com/rust-lang/rust-clippy/pull/9461
+            vec!["clippy", "--fix", "--allow-no-vcs", "--"]
         } else {
             vec!["--", "--message-format=json", "--"]
```
Alexendoo, line 355: While you're in the neighbourhood, could change this bit to

```suggestion
            vec!["clippy", "--message-format=json", "--"]
```

I did find this quite confusing before the conversation in https://github.com/rust-lang/rust-clippy/pull/9461

flip1995:
There is a list of allowed crates in the Rust compiler: https://github.com/rust-lang/rust/blob/master/src/tools/tidy/src/deps.rs Clippy should also adhere to this list of crates. For more information about this, see https://rustc-dev-guide.rust-lang.org/crates-io.html

I'm ok with the `log` crate as a dependency. But `simplelog` is not part of the above list.

I'll leave the decision whether `log` is actually necessary to Alex/Matthias, since they are the main owners / maintainers of this tool.

kraktus:
> There is a list of allowed crates in the Rust compiler: https://github.com/rust-lang/rust/blob/master/src/tools/tidy/src/deps.rs Clippy should also adhere to this list of crates. For more information about this, see https://rustc-dev-guide.rust-lang.org/crates-io.html
> 
> I'm ok with the `log` crate as a dependency. But `simplelog` is not part of the above list.
> 
> I'll leave the decision whether `log` is actually necessary to Alex/Matthias, since they are the main owners / maintainers of this tool.

This is fair, I can implement a simple log consumer by hand which would avoid that dependency. It looks like lintcheck is already using crates not from the whitelist though (`walkdir` `ureq`, `toml`, `tar`, `clap`), only counting direct deps.

> I was thinking about adding another output mode alongside normal/markdown that output the rendered diagnostics to the log, would that cover it?

Functionality-wise it would, but I feel this would be less-maintainable than leveraging on the rust ecosystem.

kraktus:
First just want to thank you Alexendoo for your PR because I believe `lintcheck` is our cornerstone tool to proactively find new false positives.

Now, I would argue that the difference of complexity between https://github.com/rust-lang/rust-clippy/pull/9764/commits/cea40448baa1c331142ab36aa7f45749c7318b06 (from #9764 which also add 2 new deps and rely on instable features) and https://github.com/rust-lang/rust-clippy/pull/9731/commits/5461e99e65e813c6791e705c440d2ccabd2a9b80 favouring the latter is compelling enough to rely on `log` and `env_logger`.

With latest version of env_logger fixing their issue with piping logs to a file, I can revert to use it and all new deps would be in the rustc whitelist. Will wait for a final call from matthiaskrgr before doing so though.

Alexendoo:
Comparing those commits seems odd, they do completely different things. Adding `log`/`env_logger` would not make `diff`/`strip-ansi-escapes` any less necessary

The sigpipe feature is so that `lintcheck ... | head/less/etc` doesn't panic

xFrednet:
Hey @matthiaskrgr, this is a ping from triage. A previous commit, mentioned that they would like input from you. Would you mind taking a look at it: https://github.com/rust-lang/rust-clippy/pull/9731#issuecomment-1312757489

@kraktus Since this PR was created quite some time ago do you believe this would still be a good addition?

matthiaskrgr:
Oh damn, I have absolute no memory of ever seeing this :sweat:  but I've also turned off all github mail notifications since getting 30-40 mails a day was getting unbearable.

To be honest I'm not sure if I am still fit as "area expert" for lintcheck, I did write the initial impl with a clear goal of what I wanted it to be, but it has also been rewritten several times (at least thats what it feels like) so that I no longer feel familiar with the code, I have also not run it in probably 3-4 years by myself (if others do, that is great though :D) 

I looked over the PR and did not see anything that made me think "omg wtf", so lgtm I guess??

kraktus:
I think fixing —fix would be a good addition yes, eventually running it as CI whenever a new MachineApplicable lint is created to know about cornercases should be a goal.

matthiaskrgr:
Ah yes! 
I've integrated that into icemaker around one and a half years ago (take a `.rs` file,  make sure it compiles, run cargo clippy --fix, and note if suggestion applies or not ) which yielded around 70 tickets already guess :D

With whole crates there is always a higher change though that lints do collide in some weird way, sometimes two lints give conflicting suggestions for the same line, or that you'll have to remove major parts of the program to get a somewhat usable mvce.

xFrednet:
> To be honest I'm not sure if I am still fit as "area expert" for lintcheck, I did write the initial impl with a clear goal of what I wanted it to be, but it has also been rewritten several times (at least thats what it feels like) so that I no longer feel familiar with the code, I have also not run it in probably 3-4 years by myself (if others do, that is great though :D)

Okay, that's understandable!

r? @Alexendoo Would you mind taking over this review? You already left a few comments earlier and know lintcheck a bit AFAIK. If you don't have the time, you can reassign it to me :)

> I think fixing `--fix` would be a good addition yes, eventually running it as CI whenever a new MachineApplicable lint is created to know about cornercases should be a goal.

Sounds reasonable, thank you for the answer :D

Alexendoo:
Yeah I can review it, may need redoing since it's changed a fair bit since the PR was opened, I don't think anybody has touched `--fix` so I assume it's still in need of repair

xFrednet:
Hey @Alexendoo this is a ping from triage. Could you take a look at this issue and indicate what needs to be done next?

kraktus:
I could rebase it and work on next week. I'll need to check back the code as I've lost track of the changes since. I'd like to get a green light on it before though.

flip1995:
As we already use  some deps that are not white listed in the Rust repo and `lintcheck` is not really part of the Rust repo workspace, `simplelog` should be fine. (It's MIT/Apache2.0 licensed). So this should resolve my blocker of this PR :)

Alexendoo:
Personally I'd rather have `-v` print the diagnostics in full in the report than have it go via logging

xFrednet:
Hey @kraktus, could you rebase this PR on master? Recently, we made several changes to lintcheck.

For the changes. I'd like to have `--fix` working. We now run lintcheck in our CI but it doesn't check the suggestions yet :D

For the logging part. Our CI now displays the full error message. You can take a look at https://github.com/xFrednet/rust-clippy/actions/runs/10044508861 from https://github.com/rust-lang/rust-clippy/pull/13139 as an example. Having a local option would also be cool, but I'd also be in favor of a flag instead of logging.

@Alexendoo I've also assigned myself, as I want to keep track of it. We can both do reviews on this one :D

kraktus:
Hey, I don't have plan to work on this in the foreseeable future, feel free to pick it up

xFrednet:
Alright, then I'll close this PR for now. Thank you :D

PR #9733, opened 2022-10-27
Ensure new_ret_no_self is not fired if impl Trait<Self> is returned.

### LINKED ISSUE #7344: `new_ret_no_self` false positive when returning `impl Trait<Self>`
Lint name: new_ret_no_self

```rust
pub struct Foo;

impl Foo {
    pub fn new() -> impl Into<Self> {
        Foo
    }
}
```

causes a warning where it shouldn't. `impl Into` here can be any trait that has `Self` in generic arguments.

This is similar to #4359 but instead of `Future<Output = Self>` it's `Into<Self>`. #4365's title implies that this should be allowed but actually it only deals with associated types, not type parameters.

`cargo clippy -V`: 0.1.54 (2021-06-09 eab201d)

nbdd0121:
Fix #7344: ensure new_ret_no_self is not fired if `impl Trait<Self>` is returned.

changelog: [`new_ret_no_self`]: No longer lints when `impl Trait<Self>` is returned

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -103,7 +103,7 @@ mod zst_offset;
 use bind_instead_of_map::BindInsteadOfMap;
 use clippy_utils::consts::{constant, Constant};
 use clippy_utils::diagnostics::{span_lint, span_lint_and_help};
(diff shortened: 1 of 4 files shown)


dswij (approved):

dswij:
Thanks for this! @bors r+

PR #9744, opened 2022-10-28
add [`permissions_set_readonly_false`] #9702

### LINKED ISSUE #9702: Lint `permissions.set_readonly(false)`
### What it does

The lint would catch any usage of [`Permissions::set_readonly`](https://doc.rust-lang.org/nightly/std/fs/struct.Permissions.html#method.set_readonly) called with `false`

`.set_readonly(false)` on unix platforms sets the file to be world writable - https://github.com/rust-lang/rust/pull/101644

### Lint Name

permissions_set_readonly_false

### Category

suspicious

### Advantage

The behaviour is surprising, so a lint to inform people that it's probably not doing what they expect would be useful

### Drawbacks

If someone knows about that behaviour and is using it intentionally it would be a false positive, however in that case the `#[allow]` would be a good indicator to others that it is intentional

It's possible that the behaviour will be changed or deprecated upstream before the lint is written

### Example

```rust
let f = File::create("foo.txt")?;
let metadata = f.metadata()?;
let mut permissions = metadata.permissions();

permissions.set_readonly(false);
```

dboso:
Closes #9702

---

changelog: New lint [`permissions_set_readonly_false`]
[#9744](https://github.com/rust-lang/rust-clippy/pull/9744)

### DIFF
--- a/clippy_lints/src/permissions_set_readonly_false.rs
+++ b/clippy_lints/src/permissions_set_readonly_false.rs
@@ -1,0 +1,54 @@
+use clippy_utils::diagnostics::span_lint_and_note;
+use clippy_utils::paths;
+use clippy_utils::ty::match_type;
+use rustc_ast::ast::LitKind;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to `std::fs::Permissions.set_readonly` with argument `false`
+    ///
+    /// ### Why is this bad?
+    /// On Unix platforms this results in the file being world writable
+    ///
+    /// ### Example
+    /// ```rust
+    /// let f = File::create("foo.txt").unwrap();
+    /// let metadata = f.metadata().unwrap();
+    /// let mut permissions = metadata.permissions();
+    /// permissions.set_readonly(false);
+    /// ```
+    #[clippy::version = "1.66.0"]
+    pub PERMISSIONS_SET_READONLY_FALSE,
+    suspicious,
+    "Checks for calls to `std::fs::Permissions.set_readonly` with argument `false`"
+}
+declare_lint_pass!(PermissionsSetReadonlyFalse => [PERMISSIONS_SET_READONLY_FALSE]);
+
+impl<'tcx> LateLintPass<'tcx> for PermissionsSetReadonlyFalse {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let ExprKind::MethodCall(path, receiver, args, _) = &expr.kind {
+            if_chain! {
+                if match_type(cx, cx.typeck_results().expr_ty(receiver), &paths::PERMISSIONS);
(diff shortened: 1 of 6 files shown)


clippy_lints/src/permissions_set_readonly_false.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[clippy::version = "1.66.0"]
+    pub PERMISSIONS_SET_READONLY_FALSE,
+    suspicious,
+    "Checks for calls to `std::fs::Permissions.set_readonly` with argument `false`"
+}
+declare_lint_pass!(PermissionsSetReadonlyFalse => [PERMISSIONS_SET_READONLY_FALSE]);
+
+impl<'tcx> LateLintPass<'tcx> for PermissionsSetReadonlyFalse {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let ExprKind::MethodCall(path, receiver, args, _) = &expr.kind {
```
Jarcho, line 32: `args` should be `[arg]`.

clippy_lints/src/permissions_set_readonly_false.rs
```diff
@@ -0,0 +1,54 @@
...
+            if_chain! {
```
Jarcho, line 33: Let-chains should be used here rather than `if_chain`. e.g.
```rust
if let Some(x) = foo
    && let y = bar
    && baz
{}
```

Jarcho (changes requested):
Just a couple of small style nits.

Not 100% sure about `suspicious`, but I don't think any of the other categories fit and the behaviour on unix is pretty bad.

clippy_lints/src/permissions_set_readonly_false.rs
```diff
@@ -0,0 +1,49 @@
...
+    "Checks for calls to `std::fs::Permissions.set_readonly` with argument `false`"
+}
+declare_lint_pass!(PermissionsSetReadonlyFalse => [PERMISSIONS_SET_READONLY_FALSE]);
+
+impl<'tcx> LateLintPass<'tcx> for PermissionsSetReadonlyFalse {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let ExprKind::MethodCall(path, receiver, [arg], _) = &expr.kind
+            && match_type(cx, cx.typeck_results().expr_ty(receiver), &paths::PERMISSIONS)
+                && path.ident.name == sym!(set_readonly)
+                && [arg].len() == 1
```
Jarcho, line 36: There's no reason to use `[arg]` here. You're creating a type `[&Expr; 1]` which will always have a length of one.

Next line also has the same unneeded array creation. Can just be `&arg.kind`.

Jarcho (changes requested):
Another small style nit. You can squash the changes and keep everything as one commit.

Jarcho:
Ping @dboso. Do you plan on continuing this?

dboso:
Feel free to take over :)

Sent from Proton Mail mobile

-------- Original Message --------

chansuke:
I'd like to take over this PR

xFrednet:
I'm closing this in favor of #10063

Thank you for the groundwork @dboso. If the new PR stalls, and you want to continue this work feel free to reopen it or create a new one :)

PR #9745, opened 2022-10-28
Fix `redundant_closure_for_method_calls` suggestion

### LINKED ISSUE #7746: `clippy::redundant-closure-for-method-calls` suggestion does not surround `* mut T` with `<>`
```
Error: Unable to create clippy annotations! Reason: HttpError: Resource not accessible by integration
Warning: It seems that this Action is executed from the forked repository.
Warning: GitHub Actions are not allowed to create Check annotations, when executed for a forked repos. See https://github.com/actions-rs/clippy-check/issues/2 for details.
Posting clippy checks here instead.
error: redundant closure
   --> src/rc.rs:391:17
    |
391 |                 |mem| mem.cast::<RcBox<MaybeUninit<T>>>(),
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace the closure with the method itself: `*mut T::cast`
    |
    = note: `-D clippy::redundant-closure-for-method-calls` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure_for_method_calls
```

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=581dff623478876daa72c1b4c48a0012

The suggestion that is incorrect and also does not compile: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=c690d5e30d2d595ade64eba599ac2ad7

The correct suggestion: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=3eb69e14aa05f18c3f27f09e515c4a4c

### Meta

```
  /usr/share/rust/.cargo/bin/rustc -V
  rustc 1.57.0-nightly (aa7aca3b9 2021-09-30)
  /usr/share/rust/.cargo/bin/cargo -V
  cargo 1.57.0-nightly (d56b42c54 2021-09-27)
  /usr/share/rust/.cargo/bin/rustup -V
  rustup 1.24.3 (ce5817a94 2021-05-31)
```

matttpt:
Fixes #7746. The issue turns out to be more general than raw pointers. The `redundant_closure_for_method_calls` lint produces incorrect suggestions when the method is associated with a type that must be enclosed in angle brackets or must be written with generic arguments substituted. For example:

```rust
fn main() {
    // Clippy's suggestion: [T; N]::as_slice
    // Correct suggestion:  <[u8; 3]>::as_slice
    let array_opt: Option<&[u8; 3]> = Some(&[4, 8, 7]);
    array_opt.map(|a| a.as_slice());

    // Clippy's suggestion: [T]::len
    // Correct suggestion:  <[u8]>::len
    let slice_opt: Option<&[u8]> = Some(b"slice");
    slice_opt.map(|s| s.len());

    // Clippy's suggestion: *const T::is_null
    // Correct suggestion:  <*const usize>::is_null
    let ptr_opt: Option<*const usize> = Some(&487);
    ptr_opt.map(|p| p.is_null());

    // Clippy's suggestion: dyn TestTrait::method_on_dyn
    // Correct suggestion:  <dyn TestTrait>::method_on_dyn
    let test_struct = TestStruct {};
    let dyn_opt: Option<&dyn TestTrait> = Some(&test_struct);
    dyn_opt.map(|d| d.method_on_dyn());
}

// For the trait object example:
trait TestTrait {}
struct TestStruct {}
impl TestTrait for TestStruct {}

impl dyn TestTrait + '_ {
    fn method_on_dyn(&self) -> bool {
        false
    }
}
```

The issue also affects references and tuples, though I had to patch the standard library with non-trait methods for those types to test that. Just in case, I also included handling for `!`, since it appeared to be possible to call methods on it with angle brackets. I just couldn't verify the resulting suggestion, since dead-code analysis eliminates the code first.

This is my first exposure to Rust compiler internals, so please let me know if I'm taking the wrong approach here!

changelog: [`redundant_closure_for_method_calls`]: add angle brackets and substitute generic arguments in suggestion when needed

### DIFF
--- a/clippy_lints/src/eta_reduction.rs
+++ b/clippy_lints/src/eta_reduction.rs
@@ -11,7 +11,7 @@ use rustc_hir::{Closure, Expr, ExprKind, Param, PatKind, Unsafety};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::adjustment::{Adjust, Adjustment, AutoBorrow};
 use rustc_middle::ty::binding::BindingMode;
-use rustc_middle::ty::{self, Ty, TypeVisitable};
+use rustc_middle::ty::{self, EarlyBinder, SubstsRef, Ty, TypeVisitable};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::symbol::sym;
 
@@ -150,7 +150,7 @@ impl<'tcx> LateLintPass<'tcx> for EtaReduction {
             if check_sig(cx, closure_ty, call_ty);
             then {
                 span_lint_and_then(cx, REDUNDANT_CLOSURE_FOR_METHOD_CALLS, expr.span, "redundant closure", |diag| {
-                    let name = get_ufcs_type_name(cx, method_def_id);
+                    let name = get_ufcs_type_name(cx, method_def_id, substs);
                     diag.span_suggestion(
                         expr.span,
                         "replace the closure with the method itself",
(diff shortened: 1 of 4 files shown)


flip1995:
Sorry for taking so long. This is a really nice fix, thanks!

@bors r+

PR #9760, opened 2022-10-31
Add new lint [`let_underscore_future`]

### LINKED ISSUE #9721: Should assigning an unawaited future to `let _ = ...` be it's own lint?
### What it does

I recently came across and fixed a bug in my code where I had an async function that returns a `Result` and meant to ignore the result so I used `let _ = ...`, but forgot to `.await` the function. ([link to full context](https://github.com/BfBB-Clash/BfBB-Clash/pull/107/files).

I know that lints like `let_underscore_drop` exist but this is quite broad and doesn't do what I would want here.

Should there be a new lint that specifically looks for assigning a `Future` to `let _ = ...`?

### Lint Name

let_underscore_future

### Category

suspicious

### Advantage

- Alerts the user that a created `Future` is not being completed.

### Drawbacks

Is there any reason why you might want to immediately drop an un-awaited future with `let _ = ...`? I'm not aware of any.

### Example

```rust
async fn foo() -> Result<(), Error> {}

async fn main() {
  let _ = foo();
}
```

Could be written as:

```rust
async fn foo() -> Result<(), Error> {}

async fn main() {
  let _ = foo().await;
}
```

SquareMan:
This closes #9721 
---

changelog: add new lint [`let_underscore_future`]


clippy_lints/src/let_underscore.rs
```diff
@@ -59,7 +59,31 @@ declare_clippy_lint! {
...
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `let _ = <expr>` where the resulting type of expr implements `Future`
+    ///
+    /// ### Why is this bad?
+    /// Futures must be polled for work to be done. The original intention was most likely to await the future
+    /// and ignore the resulting value.
+    ///
+    /// ### Example
+    /// ```rust,ignore
```
llogiq, line 71: Why is this ignored? Compiling this will make sure the code is correct (for some value of correct).

clippy_lints/src/let_underscore.rs
```diff
@@ -88,6 +112,16 @@ impl<'tcx> LateLintPass<'tcx> for LetUnderscore {
                     "consider using an underscore-prefixed named \
                             binding or dropping explicitly with `std::mem::drop`",
                 );
+            } else if let Some(future_trait_def_id) = cx.tcx.lang_items().future_trait()
+                && implements_trait(cx, cx.typeck_results().expr_ty(init), future_trait_def_id, &[]) {
+                span_lint_and_help(
+                    cx,
+                    LET_UNDERSCORE_FUTURE,
+                    local.span,
+                    "non-binding let on a future",
```
llogiq, line 121: I'd suggest putting the keyword in backticks, as in:
```
"non-binding `let` on a future",
```

llogiq (changes requested):
This looks good apart from a few small nits.

clippy_lints/src/let_underscore.rs
SquareMan: Good call, should I adjust the other lint messages in this file as well?

clippy_lints/src/let_underscore.rs
SquareMan: Makes sense. I originally copied the above lint's comment as a template which actually needs it.

clippy_lints/src/let_underscore.rs
llogiq: Yes, that makes sense.

SquareMan: Gotcha, I figured as much so I did that in my changed commits

clippy_lints/src/let_underscore.rs
```diff
@@ -88,6 +119,16 @@ impl<'tcx> LateLintPass<'tcx> for LetUnderscore {
...
```
llogiq, line 128: Sorry, I missed this one; please also backtick the `let` here.

llogiq (commented):
I think we missed one `let` in the lint message. Apart from that this is good to merge.

clippy_lints/src/let_underscore.rs
SquareMan: Ah good catch, forgot about those messages.

SquareMan:
Took a couple tries to fix the test files but looks like we're good now.

llogiq:
Thank you! And sorry for the inconvenience.

@bors r+

bugadani:
This lint misses `_ = async_fn();` cases. Is this distinction meaningful between the underscore expression and `let _` or is this a false negative this lint should catch?

PR #9770, opened 2022-11-01
Add new lint  [`misnamed-getters`]

### LINKED ISSUE #9769: Lint detecting typos and bad copy-paste in getter methods
### What it does

Credit to `@sdroege` for [the idea](https://twitter.com/sdroege_/status/1586799257968906241)

This new lint would try to detect potential typos when writing getters for structs and unions, where the name of the method doesn't match the name of the field being accessed, and that there is another field whose name does match.

It can also detect the `_mut` versions of these functions.

### Lint Name

misnamed_getters

### Category

suspicious

### Advantage

Detects potential bugs

### Drawbacks

We can't know for sure the intent and context of the method, and there might be valid cases triggering false positives.

### Example

```rust                                                                                                                                                                                                                       
struct A {                                                                                                                                                                                                                    
    a: String,                                                                                                                                                                                                                
    b: String,                                                                                                                                                                                                                
 }                                                                                                                                                                                                                             
                                                                                                                                                                                                                              
impl A {                                                                                                                                                                                                                      
     fn a(&self) -> &str{                                                                                                                                                                                                      
        &self.b                                                                                                                                                                                                             
    }                                                                                                                                                                                                                         
}                                                                                                                                                                                                                             
```

Should probably be:

```rust
impl A {                                                                                                                                                                                                                      
     fn a(&self) -> &str{                                                                                                                                                                                                      
        &self.a                                                                                                                                                                                                             
    }                                                                                                                                                                                                                         
}   
```

sgued:
```
changelog: Add new lint  [`misnamed-getters`]
```

Closes #9769 

The current lint matches all methods with a body of just one expression under the form `(&mut?)? <expr>.field` where field doesn't match the name of the method but there is a field of the same type in `<expr>` that matches the name. This allows matching nested structs, for example for newtype wrappers. This may cast the net a bit too wide and cause false positives. I'll run [clippy_lint_tester](https://github.com/mikerite/clippy_lint_tester) on the top crates to see how frequently false positives happen.

There also may be room for improvement by checking that the replacement field would work taking into account implementations of `Deref` and `DerefMut` even if the types don't exactly match but I don't know yet how this could be done.


llogiq:
Hi there! Thanks for improving clippy! I'd like to note that the prefix "mis-" as in "misnamed" is only written with one 's'. Please rename the lint before someone makes inappropriate jokes about it.

sgued:
Thanks, I fixed it. I don't know if there  is a way to change the branch name without having to create a new PR...

sgued:
I also expended the lint to also match unsafe `_unchecked` getters

llogiq:
Don't worry about the branch name, no one looks at them & they're usually puns anyway.

clippy_lints/src/functions/misnamed_getters.rs
```diff
@@ -0,0 +1,129 @@
...
+        if f.name.as_str() == name {
+            correct_field = Some(f);
+        }
+        if f.name == used_ident.name {
+            used_field = Some(f);
+        }
+    }
+
+    let Some(used_field) = used_field else {
+            if cfg!(debug_assertions) {
```
sgued, line 104: This doesn't work for when the field access happens through a de-reference. For example in [syn](https://github.com/dtolnay/syn/blob/4c0ddcc84ce51a84b57ce54fcfda8893f67dd4a1/src/lit.rs#L476) where `repr` is a box, and the field is obtained by de-reference. I'll fix that.

sgued: Ok I fixed most of the issue by using `cx.typeck_results().expr_adjustements()`. There are still some false negatives left (see `FIXME` comment) but it should still be a pretty useful lint.

Mannually getting the fields reachable through recursive autoderef seems much more complex so I won't do it now.

llogiq:
Thank you!

@bors r+

Issue #9782, opened 2022-11-03
needless_borrows_for_generic_args is wrong


e00E:
### Summary

Started showing up with Rust 1.65.0 .

### Lint Name

needless_borrow

### Reproducer

I tried this code:

```rust
fn foo<T: AsRef<[u8]>>(t: T) {
    println!("{}", std::mem::size_of::<T>());
    let _t: &[u8] = t.as_ref();
}

fn main() {
    let a: [u8; 100] = [0u8; 100];

    // 100
    foo::<[u8; 100]>(a);
    foo(a);

    // 16
    foo::<&[u8]>(&a);
    foo(a.as_slice());

    // 8
    foo::<&[u8; 100]>(&a);
    foo(&a);
}
```

I saw this happen:

```
warning: the borrowed expression implements the required traits
  --> src/main.rs:14:18
   |
14 |     foo::<&[u8]>(&a);
   |                  ^^ help: change this to: `a`
   |
   = note: `#[warn(clippy::needless_borrow)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

warning: the borrowed expression implements the required traits
  --> src/main.rs:18:23
   |
18 |     foo::<&[u8; 100]>(&a);
   |                       ^^ help: change this to: `a`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

warning: the borrowed expression implements the required traits
  --> src/main.rs:19:9
   |
19 |     foo(&a);
   |         ^^ help: change this to: `a`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

warning: `a` (bin "a") generated 3 warnings
```

I expected to see this happen:
> no lint

While it is true that the trait is implemented by the borrowed expression, it is not true that this operation "is going to be dereferenced immediately by the compiler", and it is wrong to suggest that these instances should be changed to not use a reference. First, the calls where the type of T is specified no longer compile with the suggestion, and more importantly this changes the semantics of the code. There might be very different implementations based on the type or a reference was intentionally used to not copy a large object into the function.

### Version

```text
rustc 1.65.0 (897e37553 2022-11-02)
binary: rustc
commit-hash: 897e37553bba8b42751c67658967889d11ecd120
commit-date: 2022-11-02
host: x86_64-unknown-linux-gnu
release: 1.65.0
LLVM version: 15.0.0
```

[...]

PR #9791, opened 2022-11-04
Address issues 9739 and 9782

### LINKED ISSUE #9739: needless_borrow false positive: `if` and `else` have incompatible types
### Summary

If you have a function accepting both a slice and an array, the `needless_borrow` lint suggests you to turn the slice into an array. Most times this works, but if you have an else/if with differently sized slices in the then/else bodies, applying the lint's suggestion causes a compilation error.

### Lint Name

needless_borrow

### Reproducer

I tried this code:

```rust
fn foo<D: std::fmt::Display>(_it: impl IntoIterator<Item = D>) {}

fn main() {
    foo(if std::env::var_os("HI").is_some()  {
        &[0]
    } else {
        &[] as &[u32]
    });
}
```

I saw this happen:

```
error: the borrowed expression implements the required traits
 --> a.rs:7:9
  |
7 |         &[0]
  |         ^^^^ help: change this to: `[0]`
  |
```

### Version

```text
rustc 1.66.0-nightly (0da281b60 2022-10-27)

but also on

rustc 1.65.0-beta.4 (a31705549 2022-10-21)

(no idea about stable)
```

### Additional Labels

_No response_

smoelius:
This PR fixes #9739 in the manner I suggested in https://github.com/rust-lang/rust-clippy/issues/9739#issuecomment-1296802376.

This PR also fixes the compilation failures in #9782 (but doesn't address @e00E's other objections).

Fixes #9739

r? @Jarcho

changelog: Fix two `needless_borrow` false positives, one involving borrows in `if`-`else`s, the other involving qualified function calls

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -805,30 +805,39 @@ fn walk_parents<'tcx>(
                     .position(|arg| arg.hir_id == child_id)
                     .zip(expr_sig(cx, func))
                     .and_then(|(i, sig)| {
-                        sig.input_with_hir(i).map(|(hir_ty, ty)| match hir_ty {
-                            // Type inference for closures can depend on how they're called. Only go by the explicit
-                            // types here.
-                            Some(hir_ty) => binding_ty_auto_deref_stability(cx, hir_ty, precedence, ty.bound_vars()),
-                            None => {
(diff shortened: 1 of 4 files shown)


Jarcho:
Thank you. @bors r+

PR #9796, opened 2022-11-04
Fix #9771 (`unnecessary_to_owned` false positive)

### LINKED ISSUE #9771: unnecessary_to_owned includes invalid `&` in suggestion
### Summary

The `unnecessary_to_owned` lint includes an invalid `&` in the suggestion, making the code not compile. Removing that reference makes the suggestion valid.

### Reproducer

I tried this code:

```rust
use std::marker::PhantomData;

pub struct Key<K: AsRef<[u8]>, V: ?Sized>(K, PhantomData<V>);

impl<K: AsRef<[u8]>, V: ?Sized> Key<K, V> {
    pub fn new(key: K) -> Key<K, V> {
        Key(key, PhantomData)
    }
}

pub fn pkh(pkh: &[u8]) -> Key<Vec<u8>, String> {
    Key::new([b"pkh-", pkh].concat().to_vec())
}
```

The lint should say that `.to_vec()` is redundant and can be removed.

Instead, this happened:

```
warning: unnecessary use of `to_vec`
  --> src/main.rs:13:14
   |
13 |     Key::new([b"pkh-", pkh].concat().to_vec())
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use: `&[b"pkh-", pkh].concat()`
   |
```

It is almost correct, but the suggestion includes a `&`, which is invalid because `Key::new` expects a `Vec<u8>`, not a `&Vec<u8>`. Removing that leading `&` makes the suggestion valid, and the code compiles.

### Version

```text
rustc 1.67.0-nightly (11ebe6512 2022-11-01)
binary: rustc
commit-hash: 11ebe6512b4c77633c59f8dcdd421df3b79d1a9f
commit-date: 2022-11-01
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

smoelius:
Fixes #9771

In that issue's example(s), the lint tried to add a `&` to a value, which implicitly changed the type of a field to a reference. The fix is to add the reference to `receiver_ty` (the type of the receiver of the `to_owned`-like method), before passing `receiver_ty` to `can_change_type`. `can_change_type` properly rejects the modified `receiver_ty`.

cc: @mikerite just because I think he was the author of `can_change_type`.

changelog: fix `unnecessary_to_owned` false positive which implicitly tried to change the type of a field to a reference

### DIFF
--- a/tests/ui/unnecessary_to_owned.fixed
+++ b/tests/ui/unnecessary_to_owned.fixed
@@ -426,3 +426,32 @@ mod issue_9504 {
         foo(std::path::PathBuf::new().to_string_lossy().to_string()).await;
     }
 }
+
+mod issue_9771a {
+    #![allow(dead_code)]
+
+    use std::marker::PhantomData;
+
+    pub struct Key<K: AsRef<[u8]>, V: ?Sized>(K, PhantomData<V>);
+
+    impl<K: AsRef<[u8]>, V: ?Sized> Key<K, V> {
+        pub fn new(key: K) -> Key<K, V> {
+            Key(key, PhantomData)
+        }
+    }
+
+    pub fn pkh(pkh: &[u8]) -> Key<Vec<u8>, String> {
+        Key::new([b"pkh-", pkh].concat().to_vec())
+    }
+}
+
+mod issue_9771b {
+    #![allow(dead_code)]
+
+    pub struct Key<K: AsRef<[u8]>>(K);
+
+    pub fn from(c: &[u8]) -> Key<Vec<u8>> {
+        let v = [c].concat();
+        Key(v.to_vec())
+    }
+}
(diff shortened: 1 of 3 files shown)


tests/ui/unnecessary_to_owned.fixed
```diff
@@ -426,3 +426,32 @@ mod issue_9504 {
...
+}
+
+mod issue_9771b {
+    #![allow(dead_code)]
+
+    pub struct Key<K: AsRef<[u8]>>(K);
+
+    pub fn from(c: &[u8]) -> Key<Vec<u8>> {
+        let v = [c].concat();
+        Key(v.to_vec())
```
tmpolaczyk, line 455: Note that this issue was not a false positive, ideally this should be corrected to `Key(v)`, as `.concat()` already returns a `Vec<u8>`.
```suggestion
        Key(v)
```

smoelius: I take your point, but I think we have to treat the issue as a false positive for now.

Replacing `v.to_vec()` with `v` would move `v`, and `unnecessary_to_owned` doesn't currently have a way to tell whether a value can be moved.

I suspect it would be possible to incorporate something like `possible_borrowers` from `redundant_clone`. But I also think that would be a significant change and, hence, an enhancement.

tmpolaczyk: Alright, just wanted to make sure that you understood the issue correctly. I don't know much about the internals of clippy, so I can't help you with implementing this.

flip1995:
@bors r+

Thanks!

PR #9799, opened 2022-11-04
Update lint `suspicious_to_owned`

bebecue:
changelog: [`suspicious_to_owned`]: Use explicit type annotation on lint description instead of `matches!`


dswij (approved):
Thanks for this, much appreciated!

PR #9800, opened 2022-11-04
Return multiple resolutions from `def_path_res`

Alexendoo:
Changes `def_path_res` to return all the resolutions matching the path rather than the first one (with a namespace hint that covered some cases).  This would fix any issues that come up with multiple versions of the same crate being present as they all have the same crate name

It also adds resolution of `impl _ {}` items for local items, and removes struct field resolution as it didn't seem to be used anywhere

I tested it on a local crate and it worked for the multiple crate issue, but I couldn't come up with a test that worked well with `// aux-build`, maybe `// aux-crate` after https://github.com/rust-lang/rust/pull/103266 could work but I'm not sure on that either

changelog: [`disallowed_methods`], [`disallowed_types`], [`disallowed_macros`]: fix path resolution with multiple versions of the same crate
changelog: [`disallowed_methods`]: Resolve methods in `impl`s in the current crate

### DIFF
--- a/clippy_lints/src/await_holding_invalid.rs
+++ b/clippy_lints/src/await_holding_invalid.rs
@@ -1,7 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::{match_def_path, paths};
 use rustc_data_structures::fx::FxHashMap;
-use rustc_hir::def::{Namespace, Res};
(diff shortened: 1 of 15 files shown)


dswij (approved):
LGTM, thanks for this!

Alexendoo:
Huh I thought this was already merged 😄

PR #9802, opened 2022-11-05
Enable ptr_as_ptr by default

SUPERCILEX:
Under the assumption that the `as` keyword is a mistake, we should be encouraging users to move away from it where possible.

changelog: [`ptr_as_ptr`]: warn by default


Alexendoo:
related zulip thread: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Moving.20away.20from.20.60as.60.20.28cast_lossless.29

SUPERCILEX:
Closing as per https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Enable.20ptr_as_ptr.20by.20default/near/315508408

Issue #4864, opened 2019-11-29
Re-enable cast_lossless lint by default


Shnatsel:
The [cast_lossless](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless) lint has been made pedantic following #4528, with the reasoning being "It will only become a problem if the code changes in the future".

I am a happy user of this lint, and I find that converting always-lossless casts to `from()` calls as the lint suggests clarifies the code a great deal: it makes it explicit where truncation may happen, which makes the code _much_ easier to understand.

As such, I believe this lint is a good candidate for `clippy::style` category. Perhaps its description could be revised to make the readability aspect more explicit.

Shnatsel:
Example from real code to back this up:
Before: https://github.com/fschutt/fastblur/blob/c80b518e9633605546352895fb963dde39ce40f6/src/blur.rs#L203-L213
After: https://github.com/fschutt/fastblur/blob/df2032bf50a30186aa569959e960292b3215fa8d/src/blur.rs#L203-L213

jhpratt:
> it makes it explicit where truncation may happen

That's what the "lossless" part is. Truncation can't happen with the current types. It would lint if the types changes in the future such that truncation is possible. As such, I am personally opposed to this.

llogiq:
I don't think that enabling this lint will be very popular. At best, it improves style (but not much, and it makes the code longer, too), and at worst it is seen as a false positive.

flip1995:
I agree with @llogiq. IMO `pedantic` is the most fitting group for this lint.

sunfishcode:
A common misconception about `cast_lossless` seems to be that it's just about wight might happen if you change your code, and naturally this makes it seem pedantic.

`cast_lossless` is also about what may happen if someone else changes other code. With `as`, your code may change from correct to buggy without you yourself changing it, and without anyone seeing a warning.

llogiq:
> if someone else changes other code

How would that play out? Changing the signature or type of a `let`? That would still be function-local or even scope-local, right?

Shnatsel:
I suppose violation of correctness could happen like this:

1. Third-party library exposes a struct in version 0.5, say `{ data: &[u8], number: u16 }`
1. Client code that uses the library contains code `library_struct.number as usize` - this is sound so far
1. Third-party library changes the struct in version 0.6 to `{ data: &[u8], number: u64 }`
1. Client code upgrades the library dependency from 0.5 to 0.6

[...]

Issue #4528, opened 2019-09-09
cast-lossless is too strict


nrc:
`cast-lossless` includes issuing a warning about what might happen *if* you change your code: `warning: casting u8 to i32 may become silently lossy if you later change the type`. Given that the cast is totally safe with the code as written, this seems like it should be a pedantic lint, rather than a default one.

flip1995:
I agree. Especially because all the other "cast on number to another"-lints are allow-by-default.

PR #9803, opened 2022-11-05
Enable cast_lossless by default

SUPERCILEX:
Under the assumption that the `as` keyword is a mistake, we should be encouraging users to move away from it where possible.

Closes https://github.com/rust-lang/rust-clippy/issues/4864

changelog: [`cast_lossless`]: warn by default

### DIFF
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -152,7 +152,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "pre 1.29.0"]
     pub CAST_LOSSLESS,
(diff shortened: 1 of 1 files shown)


Alexendoo:
Personally I agree with the original reason it was moved there - https://github.com/rust-lang/rust-clippy/issues/4528

You could open a thread [on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy) though if you'd like a wider discussion about it

SUPERCILEX:
The Zulip thread pointed out the current inconsistency of when you can use From as being pretty annoying, so closing for now.

PR #9804, opened 2022-11-05
Update CoC to rust-lang/rust's

Alexendoo:
I spotted mentions of IRC channels that no longer exist when grepping, so this updates it to the same as the one in https://github.com/rust-lang/rust/blob/master/CODE_OF_CONDUCT.md

r? @flip1995

changelog: none


flip1995:
@bors r+

Thanks!

Issue #6880, opened 2021-03-10
Lint request: Safety comment on safe code


alice-i-cecile:
### What it does

Warns when a SAFETY comment is left on a safe function, trait etc.

*What is the advantage of the recommended code over the original code*

Avoids users accidentally using unsafe code in an inappropriate fashion because it's marked safe.

### Drawbacks

Won't work on normal comments, see #4828.

### Example

```rust
/// Safety: This function must be used very carefully 
pub fn my_fn(){
}
```

Should be written as:

```rust
/// Safety: This function must be used very carefully 
pub unsafe fn my_fn(){
}
```

Here's an [example](https://github.com/bevyengine/bevy/blob/main/crates/bevy_ecs/src/archetype.rs#L471-L516) of where this slipped through into real code.

giraffate:
Do you mean the [`missing_safety_doc`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc) lint?

alice-i-cecile:
No: the reverse. A safety doc was left on code that was not marked unsafe, implying that there were UB implications to using it incorrectly.

This is dangerous because safe code should not cause UB, even indirectly by violating unsafe code's assumptions in complex ways.

ojeda:
This is a good idea; in fact, possibly for both `missing_safety_doc` and `undocumented_unsafe_blocks`.

The description is unclear which one it is referring to, though (it says "`SAFETY` comment" and "unsafe code", but the examples seem to imply safety contracts for unsafe functions).

alice-i-cecile:
> The description is unclear which one it is referring to, though (it says "SAFETY comment" and "unsafe code", but the examples seem to imply safety contracts for unsafe functions).

My apologies. I meant the latter, but I suspect both would be useful.

ojeda:
> My apologies. I meant the latter, but I suspect both would be useful.

No need to apologize! And thanks for this idea and opening this issue. I have opened another one for the former, then: https://github.com/rust-lang/rust-clippy/issues/7954.

PR #9822, opened 2022-11-09
Add `unnecessary_safety_doc` lint

Veykril:
changelog: [`unnecessary_safety_doc`]: Add `unnecessary_safety_doc` lint

fixes https://github.com/rust-lang/rust-clippy/issues/6880

This lint does not trigger for private functions, just like `missing_safety_docs`. Reason for that was implementation simplicity and because I figured asking first would make more sense, so if it should trigger for private functions as well let me know and I'll fix that up as well.


Jarcho:
Looks good. Thank you.

@bors r+

Issue #7322, opened 2021-06-04
Normalize semicolon inside/outside blocks for `()` returning expressions


ojeda:
### What it does

For `()` returning expressions, check that the semicolon is inside/outside the block.

Either way it is fine, but consistency is best. Having the semicolon inside may be slightly nicer and more similar to, say, conditionals and loop bodies.

### Categories (optional)

- Kind: `clippy::style`

### Drawbacks

It is subjective, so perhaps we should offer both ways, one enabled by default.

### Example

```rust
unsafe { f(x) };
```

Could be written as:

```rust
unsafe { f(x); }
```

bjorn3:
This should preferably work for all blocks, not just unsafe blocks. Maybe it should also work for if else and match.

mbartlett21:
Having the semicolon outside means that rustfmt then doesn't make the block three lines.

E.g:
```rs
unsafe { my_fn(); }

// Helpfully formats to:
unsafe {
    my_fn();
}
```

1c3t3a:
@rustbot claim 
I'm going to work on this the coming week!

PR #7564, opened 2021-08-13
Add semicolon inside/outside block lints

1c3t3a:
This PR adds two new lints to clippy according to #7322. The two lints offer two flavors of the same thing: they check if for `()` returning blocks the semicolon is inside or outside the block and correct it if possible. At the moment the `semicolon_outside_block` lint fires in the `dogfood` test.

- \[x] Followed [lint naming conventions][lint_naming]
- \[x] Added passing UI tests (including committed `.stderr` file)
- \[ ] `cargo test` passes locally
- \[x] Executed `cargo dev update_lints`
- \[x] Added lint documentation
- \[x] Run `cargo dev fmt`

changelog:
Added two new lints according to #7322.
``[`semicolon_inside_block`]``: Checks for a block with a `()` returning expression if the semicolon is inside the block.
``[`semicolon_outside_block`]``: Checks for a block with a `()` returning expression if the semicolon is outside the block.

Issue #3305, opened 2018-10-12
Useless semicolons


leonardo-m:
I suggest to give some kind of warning or hint to remove the double ;; here:

```
fn main() {
    println!("hello");;
}
```

The D language is different from Rust so the situation isn't directly comparable, but in a case like this gives a warning like this:

temp.d(2,22): Warning: use '{ }' for an empty statement, not a ';'

estebank:
Adding a different case from https://github.com/rust-lang/rust/issues/56338:

```
enum Foo { A, B }

fn bar(f: Foo) {
    match f {
        Foo::A => {},
        Foo::B => {},
    };
}

fn main() {
    bar(Foo::A);
    bar(Foo::B);
}
```

leonardo-m:
In my opinion this is fitter in rustc more than in Clippy. But better to have it in Clippy than having it nowhere.

JohnTitor:
The first case is caught by rustc lint (`redundant_semicolon`) and Clippy lint (`clippy::no_effect`).

phansch:
Shouldn't this be just an extension of the `redundant_semicolons` lint then? It would feel weird to have `clippy::redundant_semicolons` and `redundant_semicolons` in rustc.

camsteffen:
This applies to anything with a block: `if`, `if let`, `while`, `for`. I think it should be a separate lint `needless_semicolon_after_block`.

dswij:
There is now a rustc lint `redundant_semicolons` for this

PR #9826, opened 2022-11-10
Add semicolon-outside/inside-block lints

Veykril:
changelog: Add `semicolon_outside_block` and `semicolon_inside_block` lints

Fixes https://github.com/rust-lang/rust-clippy/issues/7322

An earlier attempt at this can be found here https://github.com/rust-lang/rust-clippy/pull/7564. This PR still implements two separate lints but I am open to merging them into a single one that's configurable.


clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,123 @@
...
+        let expr_snip = snippet_with_macro_callsite(cx, expr.span, "..");
+
+        let mut suggestion: String = snippet_with_macro_callsite(cx, block.span, "..").to_string();
+
+        if let Some((expr_offset, _)) = suggestion.rmatch_indices(&*expr_snip).next()
+        && let Some(semi_offset) = suggestion[expr_offset + expr_snip.len()..].find(';') {
+            suggestion.remove(expr_offset +  expr_snip.len() + semi_offset);
+        } else {
+            return;
+        }
```
Veykril, line 109: I am not sure if the `find` is necessary since in theory the semi should be immediately at the start of the slice, but I wasn't sure if that was always gonna be the case

clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        let expr_snip = snippet_with_macro_callsite(cx, tail.span, "..");
+
+        let mut suggestion: String = snippet_with_macro_callsite(cx, block.span, "..").to_string();
+
+        if let Some((expr_offset, _)) = suggestion.rmatch_indices(&*expr_snip).next() {
+            suggestion.insert(expr_offset + expr_snip.len(), ';');
+        } else {
+            return;
+        }
```
Alexendoo, line 88: This would create equally large suggestions for large blocks

A multispan suggestion could be a good option here, one span to remove the semicolon and one to add a new one

Something along the lines of `span.with_start(block_expr.span.end())` should get the span of the semicolon

clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,131 @@
...
+        );
+    }
+}
+
+fn semicolon_outside_block(cx: &LateContext<'_>, block: &Block<'_>) {
+    if !block.span.from_expansion()
+    && block.expr.is_none()
+    && let [.., Stmt { kind: StmtKind::Semi(expr), .. }] = block.stmts
+    && let Some(block_expr @ Expr { kind: ExprKind::Block(_, _), ..}) = get_parent_expr_for_hir(cx,block.hir_id)
+    && let Some(Node::Stmt(Stmt { kind: StmtKind::Expr(_), .. })) = get_parent_node(cx.tcx, block_expr.hir_id) {
```
Alexendoo, line 107: let chain formatting nit:

```rust
if ...
    && ...
    && ...
{
    ...
}
```

clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,131 @@
...
+    restriction,
+    "add a semicolon outside the block"
+}
+declare_lint_pass!(SemicolonBlock => [SEMICOLON_INSIDE_BLOCK, SEMICOLON_OUTSIDE_BLOCK]);
+
+impl LateLintPass<'_> for SemicolonBlock {
+    fn check_block(&mut self, cx: &LateContext<'_>, block: &Block<'_>) {
+        semicolon_inside_block(cx, block);
+        semicolon_outside_block(cx, block);
+    }
```
Alexendoo, line 71: It may be slightly cleaner to start with `check_stmt` and pattern match the opposite direction rather than using `get_parent*`, I'll leave that decision up to you

Alexendoo (commented):
👋

clippy_lints/src/semicolon_block.rs
Veykril: Done that, though I am having problems with macro test, I tried using `.source_callsite()` on the spans to no avail, as that always spans the entire `macro_call!();` (including the semi). Not sure how to fix that

clippy_lints/src/semicolon_block.rs
Veykril: Yes thats a lot better

clippy_lints/src/semicolon_block.rs
Alexendoo: Hmm, that's difficult. As far as I can see the `.source_callsite()` of the statement doesn't include the semi, but there's nothing I can see that does include it (for `semicolon_outside_block`)

Maybe if the span is from a macro expansion you could use https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/source_map/struct.SourceMap.html#method.mac_call_stmt_semi_span, ignoring macros would also be an option

Veykril: Hm, actually maybe it was the other way around, never containing the semi, will check that tomorrow, thanks!

clippy_lints/src/semicolon_block.rs
```diff
@@ -96,7 +96,8 @@ impl LateLintPass<'_> for SemicolonBlock {
 }
 
 fn semicolon_inside_block(cx: &LateContext<'_>, block: &Block<'_>, tail: &Expr<'_>, semi_span: Span) {
-    let insert_span = tail.span.with_lo(tail.span.hi());
+    let tail_span_end = tail.span.source_callsite().hi();
+    let insert_span = Span::with_root_ctxt(tail_span_end, tail_span_end);
```
Alexendoo, line 100: This can be `tail.span.source_callsite().shrink_to_hi()`

clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,139 @@
...
+                                        kind: StmtKind::Semi(expr),
+                                        span,
+                                        ..
+                                    },
+                                ],
+                            ..
+                        },
+                        _,
+                    ),
+                ..
```
Alexendoo, line 87: This pattern is a bit large, I think you could pull the matching up to `block` into a `let else` or similar then match on `block`

Veykril:
Needed to rebase, changes are in [16b016e](https://github.com/rust-lang/rust-clippy/pull/9826/commits/16b016edf1518b4e1b90469819255dce221e0ab5)

clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,134 @@
...
+use rustc_errors::Applicability;
+use rustc_hir::{Block, Expr, ExprKind, Stmt, StmtKind};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// For () returning expressions, check that the semicolon is inside the block.
```
Alexendoo, line 11: Thanks! I think this is the last thing:

I believe this lint would apply to expressions returning any type rather than just `()`

Veykril: Ah, I forgot to change that good catch. Unsure what wording to take here though, let me know if this works.

Alexendoo: It's tricky, maybe something like "Suggests moving the semicolon from a block's final expression outside of the block" and the opposite, to me the current reading suggests it wouldn't work on things like `{ a; b; }`

Alexendoo: You can rebase on master also to fix the remark CI issue

clippy_lints/src/semicolon_block.rs
```diff
@@ -0,0 +1,134 @@
...
+use rustc_errors::Applicability;
+use rustc_hir::{Block, Expr, ExprKind, Stmt, StmtKind};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Suggests moving the semicolon from a block inside of the block to its kast expression.
```
Alexendoo, line 11:
```suggestion
    /// Suggests moving the semicolon after a block to the inside of the block, after its last expression.
```

oxalica:
IIRC, this lint doesn't trigger if both semicolon is used, yes?
`unsafe { foo(); };` from #3305

clippy_lints/src/semicolon_block.rs
Veykril: Sorry, I was a bit hasty on that commit I think, fixed it

Alexendoo (approved):

Alexendoo:
Great, thanks @Veykril

@bors r+

oxalica:
@Veykril 
> It does not trigger in that case no (I added it as to the tests though, thanks for bringing that up)

I think the double semicolon case perfectly fits the lint name `semicolon_outside_block`, and it's surprising that it fires only when there's no semi inside.

Also I'd expect the following cases to be warn-by-default.
```rust
unsafe { foo(); }; // Double semicolon.
match foo() { _ => foo() }; // The arm returns `()`.
match foo() { _ => foo(), _ => { foo() } }; // All arms return `()`.
for _ in [()] { unit_fn_block()/*semi or no semi*/ }; // Always `()`.
if foo() { unit_fn_block()/*semi or no semi*/ }; // Always `()`.
loop { }; // `!`
```

If we want to warn these cases, while still make it configurable for `unsafe { f() };` v.s. `unsafe { f(); }`. Should we create a separated lint for the these warning cases? I cannot come up with another name besides `semicolon_outside_block`.

oxalica:
:thinking: If the current semicolon-outside/inside-block behaviors rely purely on the syntax, and `unsafe { foo() };` is semantically always the same as `unsafe { foo(); }`, could we move these into `rustfmt` instead?

Jarcho:
Changing `{ foo() };` to `{ foo(); }` is only sementically equivalent when there aren't any bindings inside the block. `rustfmt` also has the restriction of working with an untyped ast, so it's very limited on what transformations it could do with semicolon placement.

oxalica:
> Changing `{ foo() };` to `{ foo(); }` is only sementically equivalent when there aren't any bindings inside the block.

Oh, I forget the drop order stuff. `;`-outside-the-block will delay the drop of the last Expr. So yeah it's more complicated than `rustfmt` typically does.

PR #9829, opened 2022-11-10
Make it clear that `or_fun_call` can be a false-positive

### LINKED ISSUE #8574: `or_fun_call` isn't always a performance lint
Replacing `.unwrap_or(Vec::new())` with `.unwrap_or_default()` doesn't do anything for performance, they produce identical machine code with optimizations: https://rust.godbolt.org/z/hzTdazGqW.

Perhaps the lint category could be changed or it could be changed to not detect cases where it's not actually a performance warning?

hrxi:
Also move it to nursery so that the false-positives can be dealt with.

CC #8574

changelog: [`or_fun_call`]: Mention false-positives, move to nursery.

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -832,32 +832,30 @@ declare_clippy_lint! {
     /// etc. instead.
     ///
     /// ### Why is this bad?
-    /// The function will always be called and potentially
-    /// allocate an object acting as the default.
+    /// The function will always be called. This is only bad if it allocates or
+    /// does some non-trivial amount of work.
     ///
     /// ### Known problems
-    /// If the function has side-effects, not calling it will
-    /// change the semantic of the program, but you shouldn't rely on that anyway.
+    /// If the function has side-effects, not calling it will change the
+    /// semantic of the program, but you shouldn't rely on that.
+    ///
+    /// The lint also cannot figure out whether the function you call is
+    /// actually expensive to call or not.
     ///
     /// ### Example
     /// ```rust
     /// # let foo = Some(String::new());
-    /// foo.unwrap_or(String::new());
+    /// foo.unwrap_or(String::from("empty"));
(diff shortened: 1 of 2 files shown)


llogiq:
I must admit that I don't understand why this lint should be in `style`. I'm fine with mentioning the false positive, but wouldn't putting the lint in `style` imply that we lint those cases for stylistic reasons? That would be wrong in my opinion.

We lint this because of perf (because the function may allocate or perform work), not style. Alas, there are false positives (we are not likely to fix any time soon), but that doesn't mean the reason for linting has changed.

We could reduce the false positive by creating an allow or deny list for this lint. That way, people would be able to specify which functions they want to avoid calling needlessly.

hrxi:
As-is, the lint has many false positives, which probably fails to meet the bar for performance lints? I don't know the bar for these in clippy.

A list of functions not to warn for doesn't seem feasible to me. A list of functions to warn for seems feasible, but will likely always be incomplete.

llogiq:
1. If we worry about too many false positive, the correct course of action would be to move the lint to the nursery.
2. We consider false positives far worse than false negatives. So with a deny list, the lint might be incomplete, but would still be useful.

hrxi:
Removed the recategorization, now it's only a documentation update.

llogiq:
For the record, I'd be OK with recategorizing into `nursery` until we have a way to reduce the false positives.

llogiq:
I think you need to add an explicit `#[warn]` annotation to the test case.

hrxi:
> For the record, I'd be OK with recategorizing into `nursery` until we have a way to reduce the false positives.

Thanks, I wasn't really aware of nursery being a thing. I'm new to clippy.

> I think you need to add an explicit `#[warn]` annotation to the test case.

Thanks, done. :)

llogiq:
Thank you, and sorry it took so much back-and-forth.

@bors r+

PR #9833, opened 2022-11-12
Update lint numbers in README

TennyZhuang:
changelog:none

Based on [the link](https://rust-lang.github.io/rust-clippy/master/lints.json
), we now have 603 rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 A collection of lints to catch common mistakes and improve your [Rust](https://github.com/rust-lang/rust) code.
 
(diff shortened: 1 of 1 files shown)


xFrednet:
The number gets automatically updated by `clippy_dev`. Though you're right that the number seems out of sync. It could be, that this comes from internal lints, which are not exported as part of the lint list, but are detected by `clippy_dev` on the other hand, I don't believe that we have that many. Only 10 - 20. You might want to look into the implementation to adjust the number. You're also welcome to reach out, if you'd like more support :)

TennyZhuang:
> The number gets automatically updated by `clippy_dev`.

Cool!

PR #9837, opened 2022-11-12
fix never_loop false positive

### LINKED ISSUE #9831: never_loop false positive on unconditional break to internal labeled block
### Summary

Having an internal labeled block which is _always_ broken to causes a false positive on never_loop.

### Lint Name

never_loop

### Reproducer

I tried this code: ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=18a2b2acc2e483eb760e71f6a75d335b))

```rust
fn thing(iter: impl Iterator) {
    for _ in iter {
        'b: {
            // error goes away if we just have the block's value be ().
            break 'b;
        }
    }
}
```

I saw this error:

```
error: this loop never actually loops
 --> src/lib.rs:2:5
  |
2 | /     for _ in iter {
3 | |         'b: {
4 | |             break 'b;
5 | |         }
6 | |     }
  | |_____^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#never_loop
  = note: `#[deny(clippy::never_loop)]` on by default
help: if you need the first element of the iterator, try writing
  |
2 |     if let Some(_) = iter.next() {
  |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Version

```text
rustc 1.67.0-nightly (e75aab045 2022-11-09)
binary: rustc
commit-hash: e75aab045fc476f176a58c408f6b06f0e275c6e1
commit-date: 2022-11-09
host: aarch64-apple-darwin
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

DesmondWillowbrook:
fixes #9831

changelog: [`never_loop`]: fixed false positive on unconditional break in internal labeled block

### DIFF
--- a/clippy_lints/src/loops/never_loop.rs
+++ b/clippy_lints/src/loops/never_loop.rs
@@ -169,7 +169,8 @@ fn never_loop_expr(expr: &Expr<'_>, main_loop_id: HirId) -> NeverLoopResult {
                 combine_seq(e, arms)
             }
         },
-        ExprKind::Block(b, _) => never_loop_block(b, main_loop_id),
+        ExprKind::Block(b, None) => never_loop_block(b, main_loop_id),
+        ExprKind::Block(b, Some(_label)) => absorb_break(never_loop_block(b, main_loop_id)),
         ExprKind::Continue(d) => {
             let id = d
                 .target_id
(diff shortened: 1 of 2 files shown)


dswij (approved):
LGTM, thanks for this!

DesmondWillowbrook:
Note:
I have discovered that this fix causes a false negative in the following case:

```rust
pub fn test20() {
    'a: loop {
        'b: {
            break 'a;
        }
    }
}
```
Working on it!

PR #9840, opened 2022-11-13
[`arithmetic-side-effects`]: Consider user-provided pairs

c410-f3r:
Depends on https://github.com/rust-lang/rust-clippy/pull/9592.

Because of https://github.com/rust-lang/rust-clippy/pull/9559#discussion_r985293523, r? @Alexendoo 

```
changelog: [`arithmetic-side-effects`]: Consider user-provided pairs
```

### DIFF
--- a/clippy_lints/src/operators/arithmetic_side_effects.rs
+++ b/clippy_lints/src/operators/arithmetic_side_effects.rs
@@ -5,25 +5,26 @@ use clippy_utils::{
     peel_hir_expr_refs,
 };
 use rustc_ast as ast;
-use rustc_data_structures::fx::FxHashSet;
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::Ty;
 use rustc_session::impl_lint_pass;
 use rustc_span::source_map::{Span, Spanned};
 
-const HARD_CODED_ALLOWED: &[&str] = &[
-    "&str",
-    "f32",
-    "f64",
-    "std::num::Saturating",
-    "std::num::Wrapping",
-    "std::string::String",
+const HARD_CODED_ALLOWED_BINARY: &[[&str; 2]] = &[
+    ["f32", "f32"],
+    ["f64", "f64"],
+    ["std::num::Saturating", "std::num::Saturating"],
+    ["std::num::Wrapping", "std::num::Wrapping"],
+    ["std::string::String", "&str"],
 ];
+const HARD_CODED_ALLOWED_UNARY: &[&str] = &["f32", "f64", "std::num::Saturating", "std::num::Wrapping"];
 
 #[derive(Debug)]
 pub struct ArithmeticSideEffects {
-    allowed: FxHashSet<String>,
+    allowed_binary: FxHashMap<String, FxHashSet<String>>,
+    allowed_unary: FxHashSet<String>,
     // Used to check whether expressions are constants, such as in enum discriminants and consts
     const_span: Option<Span>,
     expr_span: Option<Span>,
@@ -33,19 +34,55 @@ impl_lint_pass!(ArithmeticSideEffects => [ARITHMETIC_SIDE_EFFECTS]);
 
 impl ArithmeticSideEffects {
     #[must_use]
-    pub fn new(mut allowed: FxHashSet<String>) -> Self {
-        allowed.extend(HARD_CODED_ALLOWED.iter().copied().map(String::from));
+    pub fn new(user_allowed_binary: Vec<[String; 2]>, user_allowed_unary: Vec<String>) -> Self {
+        let mut allowed_binary: FxHashMap<String, FxHashSet<String>> = <_>::default();
+        for [lhs, rhs] in user_allowed_binary.into_iter().chain(
+            HARD_CODED_ALLOWED_BINARY
+                .iter()
+                .copied()
+                .map(|[lhs, rhs]| [lhs.to_string(), rhs.to_string()]),
+        ) {
+            allowed_binary.entry(lhs).or_default().insert(rhs);
+        }
+        let allowed_unary = user_allowed_unary
+            .into_iter()
+            .chain(HARD_CODED_ALLOWED_UNARY.iter().copied().map(String::from))
+            .collect();
         Self {
-            allowed,
+            allowed_binary,
+            allowed_unary,
             const_span: None,
             expr_span: None,
         }
     }
 
-    /// Checks if the given `expr` has any of the inner `allowed` elements.
-    fn is_allowed_ty(&self, ty: Ty<'_>) -> bool {
-        self.allowed
-            .contains(ty.to_string().split('<').next().unwrap_or_default())
+    /// Checks if the lhs and the rhs types of a binary operation like "addition" or
+    /// "multiplication" are present in the inner set of allowed types.
+    fn has_allowed_binary(&self, lhs_ty: Ty<'_>, rhs_ty: Ty<'_>) -> bool {
+        let lhs_ty_string = lhs_ty.to_string();
+        let lhs_ty_string_elem = lhs_ty_string.split('<').next().unwrap_or_default();
+        let rhs_ty_string = rhs_ty.to_string();
+        let rhs_ty_string_elem = rhs_ty_string.split('<').next().unwrap_or_default();
+        if let Some(rhs_from_specific) = self.allowed_binary.get(lhs_ty_string_elem)
+            && {
+                let rhs_has_allowed_ty = rhs_from_specific.contains(rhs_ty_string_elem);
+                rhs_has_allowed_ty || rhs_from_specific.contains("*")
+            }
+        {
+           true
+        } else if let Some(rhs_from_glob) = self.allowed_binary.get("*") {
+            rhs_from_glob.contains(rhs_ty_string_elem)
+        } else {
+            false
+        }
+    }
+
+    /// Checks if the type of an unary operation like "negation" is present in the inner set of
+    /// allowed types.
+    fn has_allowed_unary(&self, ty: Ty<'_>) -> bool {
+        let ty_string = ty.to_string();
+        let ty_string_elem = ty_string.split('<').next().unwrap_or_default();
+        self.allowed_unary.contains(ty_string_elem)
     }
 
     // For example, 8i32 or &i64::MAX.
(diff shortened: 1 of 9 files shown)


clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -11,18 +11,19 @@ use rustc_span::{
...
+    "std::num::Saturating,std::num::Saturating",
+    "std::num::Wrapping,std::num::Wrapping",
+    "std::string::String,&str",
 ];
+const HARD_CODED_ALLOWED_UNARY: &[&str] = &["f32", "f64", "std::num::Saturating", "std::num::Wrapping"];
 
 #[derive(Debug)]
 pub struct ArithmeticSideEffects {
-    allowed: FxHashSet<String>,
+    allowed_bin: FxHashMap<String, FxHashSet<String>>,
```
c410-f3r, line 25: I might improve this monstrosity in the future but for now, let's just say that it works as intended.

c410-f3r:
@Alexendoo https://github.com/rust-lang/rust-clippy/pull/9592 has been merged.

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,33 @@ macro_rules! define_Conf {
...
+    ///
+    /// Pairs are asymmetric, which means that `SomeType,AnotherType` is not the same as
+    /// `AnotherType,SomeType`.
+    ///
+    /// # Example
+    ///
+    /// ```toml
+    /// arithmetic-side-effects-allowed-bin = ["SomeType,f32", "AnotherType,*"]
+    /// ```
+    (arithmetic_side_effects_allowed_bin: Vec<String> = <_>::default()),
```
Alexendoo, line 224: Would this work with `Vec<[String; 2]>`? `[["SomeType", "f32"], ..]` would be nicer than splitting on an internal comma IMO

c410-f3r: Surprisingly (at least for me), it is possible. Changed.

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,33 @@ macro_rules! define_Conf {
...
+    ///
+    /// Pairs are asymmetric, which means that `["SomeType", "AnotherType"]` is not the same as
+    /// `["AnotherType", "SomeType"]`.
+    ///
+    /// # Example
+    ///
+    /// ```toml
+    /// arithmetic-side-effects-allowed-bin = [["SomeType" , "f32"], ["AnotherType", "*"]]
+    /// ```
+    (arithmetic_side_effects_allowed_bin: Vec<[String; 2]> = <_>::default()),
```
flip1995, line 224: I wouldn't abreviate this
```suggestion
    (arithmetic_side_effects_allowed_binary: Vec<[String; 2]> = <_>::default()),
```

c410-f3r: Done

c410-f3r:
> We talked about this in [the meeting the other day](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202022-11-29/near/312878404), one thing that came up was that we could use `_` instead of `*` for the wildcard type. There wasn't a conclusion either way though, what do you think about that @c410-f3r?

`*` is the canonical glob character used in strings but now that pairs are described as arrays of two elements, it shouldn't matter which one is better as long as the intended behavior is documented. If desired, I can add support for both.

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,33 @@ macro_rules! define_Conf {
 }
 
 define_Conf! {
-    /// Lint: Arithmetic.
+    /// Lint: arithmetic_side_effects.
```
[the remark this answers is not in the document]
c410-f3r: Done

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,33 @@ macro_rules! define_Conf {
...
+    /// Suppress checking of the passed type pair names grouped by commas in binary operations like
+    /// addition or multiplication.
+    ///
+    /// Supports the "*" wildcard to indicate that a certain type won't trigger the lint regardless
+    /// of the involved counterpart. For example, `["SomeType", "*"]` or `["*", "AnotherType"]`.
+    ///
+    /// Pairs are asymmetric, which means that `["SomeType", "AnotherType"]` is not the same as
+    /// `["AnotherType", "SomeType"]`.
+    ///
+    /// # Example
```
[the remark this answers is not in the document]
c410-f3r: Done

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,33 @@ macro_rules! define_Conf {
 }
 
 define_Conf! {
-    /// Lint: Arithmetic.
+    /// Lint: arithmetic_side_effects.
     ///
-    /// Suppress checking of the passed type names.
-    (arithmetic_side_effects_allowed: rustc_data_structures::fx::FxHashSet<String> = <_>::default()),
```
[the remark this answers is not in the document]
c410-f3r: I regret not categorizing this lint as `nursery` in the first place. As you saw yourself, many things changed, bugs were fixed and probably other things will come into play in the the future.

This lint already broke downstream users in stable toolchains because of https://github.com/rust-lang/rust-clippy/pull/9443 so can you guys allow the introduction of these new parameters without having to keep `arithmetic_side_effects_allowed`? For what is worth, there is only one project relying on the current behavior in all of GitHub.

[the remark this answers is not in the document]
Alexendoo: The changes would get into I think 1.67 or 1.68 meaning there will be a good few months where the lint is available with the current config on stable

My preference would be to keep the existing variable around, maybe renamed with `conf_deprecated` to `arithmetic_side_effects_allowed_any`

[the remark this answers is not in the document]
c410-f3r: Well, at least I tried. Done as requested.

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,41 @@ macro_rules! define_Conf {
...
+    /// Pairs are asymmetric, which means that `["SomeType", "AnotherType"]` is not the same as
+    /// `["AnotherType", "SomeType"]`.
+    ///
+    /// #### Example
+    ///
+    /// ```toml
+    /// arithmetic-side-effects-allowed-binary = [["SomeType" , "f32"], ["AnotherType", "*"]]
+    /// ```
+    (arithmetic_side_effects_allowed_binary: Vec<[String; 2]> = <_>::default()),
+    /// Lint: arithmetic_side_effects.
```
Alexendoo, line 233: Uppercase here too

Alexendoo: The lint description itself could also do with having the `# Allowed types` section removed, the configuration should show up under it properly

clippy_lints/src/utils/conf.rs
```diff
@@ -205,10 +205,41 @@ macro_rules! define_Conf {
...
 define_Conf! {
-    /// Lint: Arithmetic.
+    /// Lint: ARITHMETIC_SIDE_EFFECTS.
     ///
     /// Suppress checking of the passed type names.
+    #[conf_deprecated(
+        "Please use `arithmetic_side_effects_allowed_binary` or `arithmetic_side_effects_allowed_unary` instead",
+        arithmetic_side_effects_allowed_unary
+    )]
     (arithmetic_side_effects_allowed: rustc_data_structures::fx::FxHashSet<String> = <_>::default()),
```
Alexendoo, line 215: I wouldn't say deprecate it for the other two, only if you want to rename it

The description could mention that it's equivalent to specifying both a `_unary` and two `*` `_binary` entries

Alexendoo: Would also be good to have a test to make sure this option still works

clippy_lints/src/utils/conf.rs
c410-f3r: Done

clippy_lints/src/utils/conf.rs
c410-f3r: Done

clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -33,19 +34,67 @@ impl_lint_pass!(ArithmeticSideEffects => [ARITHMETIC_SIDE_EFFECTS]);
...
+            if rhs_from_specific.contains(rhs_ty_string_elem) || rhs_from_specific.contains("*") {
+                true
+            } else {
+                rhs_has_lhs_glob()
+            }
+        } else if let Some(rhs_from_glob) = self.allowed_binary.get("*") {
+            rhs_from_glob.contains(rhs_ty_string_elem)
+        } else {
+            rhs_has_lhs_glob()
+        }
```
Alexendoo, line 89:
```suggestion
        if let Some(rhs_from_specific) = self.allowed_binary.get(lhs_ty_string_elem)
            && (rhs_from_specific.contains(rhs_ty_string_elem) || rhs_from_specific.contains("*"))
        {
           true
        } else if let Some(rhs_from_glob) = self.allowed_binary.get("*") {
            rhs_from_glob.contains(rhs_ty_string_elem)
        } else {
            false
        }
```

clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -33,19 +34,67 @@ impl_lint_pass!(ArithmeticSideEffects => [ARITHMETIC_SIDE_EFFECTS]);
...
+                .copied()
+                .map(|[lhs, rhs]| [lhs.to_string(), rhs.to_string()]),
+        ) {
+            if let Some(hs) = allowed_binary.get_mut(&lhs) {
+                let _ = hs.insert(rhs);
+            } else {
+                let mut hs: FxHashSet<String> = <_>::default();
+                let _ = hs.insert(rhs);
+                let _ = allowed_binary.insert(lhs, hs);
+            }
```
Alexendoo, line 51:
```suggestion
            allowed_binary.entry(lhs).or_default().insert(rhs);
```

Alexendoo (commented):
Just a couple code style things and we can get this merged (once bors becomes unstuck)

clippy_lints/src/operators/arithmetic_side_effects.rs
c410-f3r: Done

clippy_lints/src/operators/arithmetic_side_effects.rs
c410-f3r: Done

Alexendoo (approved):

Alexendoo:
Thanks! @bors r+

PR #9848, opened 2022-11-14
Fix `vec-box-size-threshold` off-by-one error

lukas-code:
The docs for `vec-box-size-threshold` say "The size of the boxed type in bytes, where boxing in a `Vec` is allowed", but previously a boxed type of exactly that size wasn't allowed in `Vec`.

changelog: [`vec_box`]: Fix off-by-one error for `vec-box-size-threshold` configuration.


Jarcho:
Thank you. @bors r+

PR #9858, opened 2022-11-16
`never_loop`: don't emit AlwaysBreaks if it targets a block

DesmondWillowbrook:
ref: https://github.com/rust-lang/rust-clippy/pull/9837#issuecomment-1312788194

The previous fix (#9837) was too simple and ignored all break commands inside a labelled block, regardless of whether their destination was a labelled block or a loop. This fix tracks all the labelled blocks in scope to ensure that only breaks targeting loops are considered.

changelog: [`never_loop`]: prevent false negatives from `breaks` nested in labelled blocks


DesmondWillowbrook:
r? @dswij

For continuity on the same issue.

dswij (approved):
LGTM, thanks for this!

Issue #6920, opened 2021-03-16
MSRV attribute is not scoped


camsteffen:
Example:

```rust
#![feature(custom_inner_attributes)]

mod my_mod {
    #![clippy::msrv = "1.0.0"]
}

fn hi() {
    // the `manual_strip` lint is suppressed by the attribute in `my_mod`
    let s = "hello, world!";
    if s.starts_with("hello, ") {
        assert_eq!(s["hello, ".len()..].to_uppercase(), "WORLD!");
    }
}
```

If we don't want to support this sort of usage, maybe we can just warn or error when the attribute is found in a non-crate.

hellow554:
There's this testfile https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/min_rust_version_attr.rs and it seems that scoping is supported in some way.

It seems, that the inner attribute starts from that line on in the code, because if I move your `hi` function above `my_mod` the lint is triggered correctly.

PR #6201, opened 2020-10-21
Add support for minimum supported rust version

Suyash458:
add configuration option for minimum supported rust version
add msrv attribute to some lints listed in #6097
add tests
addresses #6097 

changelog: Add `msrv` configuration to Clippy. This should get a longer changelog entry.

PR #9860, opened 2022-11-16
Allow using `clippy::msrv` as an outer attribute

Alexendoo:
changelog: Allow specifying `#[clippy::msrv]` as an outer attribute

Probably not too useful to clippy users, but it makes the MSRV tests slightly cleaner

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ in the `Cargo.toml` can be used.
 rust-version = "1.30"
 ```
 
-The MSRV can also be specified as an inner attribute, like below.
+The MSRV can also be specified as an attribute, like below.
 
 ```rust
 #![feature(custom_inner_attributes)]
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -72,7 +72,7 @@ minimum supported Rust version (MSRV) in the clippy configuration file.
 msrv = "1.30.0"
 ```
 
-The MSRV can also be specified as an inner attribute, like below.
+The MSRV can also be specified as an attribute, like below.
 
 ```rust
 #![feature(custom_inner_attributes)]
--- a/book/src/development/adding_lints.md
+++ b/book/src/development/adding_lints.md
@@ -463,7 +463,7 @@ if !self.msrv.meets(msrvs::STR_STRIP_PREFIX) {
 }
 ```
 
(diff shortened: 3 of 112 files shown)


Jarcho:
Why was this limited to inner attributes in the first place?

Alexendoo:
I missed it the first time I looked through #6201 but just found https://github.com/rust-lang/rust-clippy/pull/6201#discussion_r516774498

I think the reasoning applied to an earlier version of the PR, but it ended up using `enter_lint_attrs` which should work fine for both inner & outer attrs

Jarcho:
Just want to make sure my interpretation is correct. This looks like when an msrv attribute is seen, it sets the msrv until the next msrv attribute, totally ignoring the scope of the attribute. This would make something like:
```rust
#[clippy::msrv = "1.28"]
fn foo() {}
fn bar() {}
```
apply the msrv to both `foo` and `bar`. If that's the case this is just encouraging a footgun on stable.

Alexendoo:
Yeah that's https://github.com/rust-lang/rust-clippy/issues/6920, it being used on stable is a good point. I'll get that fixed up first

flip1995:
I made the decision to limit this to inner attributes, because I didn't really see a use case where it would make sense to apply this to items/expressions/... and because the implementation of only inner attributes was simpler to do.

Together with the [`cfg_version`](https://doc.rust-lang.org/beta/unstable-book/language-features/cfg-version.html) feature of `rustc` also having this as an outer attribute makes sense.

Alexendoo:
Opened #9924

Didn't know about that feature, hopefully that gets to stable at some point

Picking it up automatically could be good, I would guess a cfg doesn't get passed to `enter/exit_lint_attrs` though

flip1995:
> In the case of `cfg_version` version would it not be better to set that as the msrv if it's higher than the current value?

As with all `cfg` attributes, we don't even see that attribute. We even see the code, if the feature is enabled or we don't see the code at all. We sadly cannot access `cfg` attributes.

Jarcho (approved):
LGTM. Github seems to be only sending me about half the email notifications right now, so you might need to approve yourself after rebasing

Alexendoo:
Thanks! Hopefully they fix that 👀

@bors r=Jarcho

PR #9865, opened 2022-11-18
Add allow-mixed-uninlined-format-args config

nyurik:
Implement `allow-mixed-uninlined-format-args` config param to change the behavior of the `uninlined_format_args` lint. Now it is a part of `style` per [Zulip chat](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60uninlined_format_args.60.20category), and won't propose inlining in case of a mixed usage, e.g. `print!("{} {}", var, 1+2)`. If the user sets `allow-mixed-uninlined-format-args` config param to `false`, the lint would behave like it did before -- proposing to inline args even in the mixed case.

---

changelog: [`uninlined_format_args`]: Added a new config `allow-mixed-uninlined-format-args` to allow the lint, if only some arguments can be inlined
[#9865](https://github.com/rust-lang/rust-clippy/pull/9865)
changelog: Moved [`uninlined_format_args`] to `style` (Now warn-by-default)
[#9865](https://github.com/rust-lang/rust-clippy/pull/9865)


xFrednet:
Hey @nyurik, thank you for the PR! I've seen it, this is just a small warning, that the review will take some time due to its size. I'll try to get it done by Thursday, but it might take until the end of next weekend. I hope that's alright with you :)

nyurik:
@xFrednet Let me know if there are any blockers on this one.  BTW, note that the actual code change is just a few lines - the rest is just a copy/paste of an integration test - to see that the same test behaves properly with a different configuration parameter.

xFrednet:
There are no blockers, just a busy week on my end and I got a cold. My schedule is mostly free on the weekend. You'll have a review the latest on Sunday. Thank you for your patience. :)

nyurik:
Thanks @xFrednet, didn't mean to push you on this :)   Get well soon!

clippy_lints/src/format_args.rs
```diff
@@ -305,11 +321,23 @@ fn check_uninlined_args(cx: &LateContext<'_>, args: &FormatArgsExpn<'_>, call_si
                 Applicability::MachineApplicable,
                 if multiline_fix { CompletelyHidden } else { ShowCode },
             );
+            if ignore_mixed {
+                // Improve lint config discoverability
+                diag.note_once(
```
xFrednet, line 326: I'm unsure about this part. Configurations are currently hard to discover, and discussing this should be on our agenda. A solution could be these notes, like you also suggested in #9880. However, until that has been discussed (Most likely on Tuesday on Zulip) I would hold off on this one.

Even then, I'd recommend adding the message all at once, to ensure that they're the same/similar. If this should be included in this PR, we should try to find a shorter wording. I don't remember if that's an official guideline, or just one we follow, but messages should ideally be shorter than 80 characters, 100 would also be fine, but this is a bit too long IMO.

We could remove the reference to the `clippy.toml` file, just mention that there is a configuration which is described in the lint description or say that this message could be suppressed via the `XYZ` config value. All of these are viable options with different tradeoffs. Deciding on a phrasing, is one of the reasons why I would like to do this as part of #9880 and not here.

tests/ui-toml/allow_mixed_uninlined_format_args/uninlined_format_args.rs
```diff
@@ -0,0 +1,182 @@
+// aux-build:proc_macro_with_span.rs
+// run-rustfix
```
xFrednet, line 2: This file seems to be the same as the normal ui test one, just with the new config value. Usually, it's enough in the ui-toml tests, to just include the relevant test cases. So have a few that should be linted, and a few which still shouldn't be.

That also clearly indicates what this file actually tests. Like this I had to use a local diff to see that the test covered three examples from this file. Would you mind reducing the scope? Then you can also remove the `with_span` macro magic and so on from this test as it's covered by the other one.

xFrednet (commented):
Thank you for waiting on the review. The implementation itself looks good to me. It was actually less to review then I thought, as most of the changes come from the `.stderr`, which is almost identical to the one in the `/ui/` folder.

nyurik:
@xFrednet thanks, makes sense, i think I addressed all of your feedback points. Thx again!

tests/ui-toml/allow_mixed_uninlined_format_args/uninlined_format_args.rs
nyurik: fixed

xFrednet (approved):

xFrednet:
Looks good to me, thank you very much for the addition!

@bors r+

PR #9873, opened 2022-11-18
Move `line_span` to source.rs

smoelius:
`line_span` is a non-public function used only in source.rs. It seems like it ought to go in source.rs.

changelog: none


flip1995:
@bors r+

Thanks!

PR #9881, opened 2022-11-19
Rustup

flip1995:
r? @ghost

It took >4 weeks, but I finally got to do the sync 🎉 

changelog: none


flip1995:
Further blocked by https://github.com/rust-lang/rust/pull/103660 😭  I try to deal with this tomorrow...

flip1995:
Whoooooooop Whoooooooop :tada:

src/driver.rs
```diff
@@ -252,14 +252,24 @@ pub fn main() {
     exit(rustc_driver::catch_with_exit_code(move || {
         let mut orig_args: Vec<String> = env::args().collect();
 
+        let sys_root_env = std::env::var("SYSROOT").ok();
+        let pass_sysroot_env_if_given = |args: &mut Vec<String>, sys_root_env| {
+            if let Some(sys_root) = sys_root_env {
+                args.extend(vec!["--sysroot".into(), sys_root]);
+            };
+        };
```
jyn514, line 260: This no longer avoids passing `--sysroot` twice if it's already present in the arguments (i.e. `have_sys_root_arg ` is missing). @flip1995 @ozkanonur Is that intentional? It's making https://github.com/rust-lang/rust/pull/97443/ a little tricky to fix.

(I am not sure this is the actual PR where the change was made, it's been pretty hard to track down through the subtree updates.)

ozkanonur: This is implemented intentionally to fix https://github.com/rust-lang/rust/pull/103660#issuecomment-1320974997

ozkanonur: Seems we need to check if `--sysroot` presents before this process. Do you want me to fix it ? @jyn514

flip1995: @ozkanonur The fix is already implemented in #10149 . It's just waiting for my review. (This'll be merged before the sync on Thursday)

PR #9919, opened 2022-11-20
Link to a list of configurable lints in documentation

Alexendoo:
changelog: none

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,8 +197,8 @@ disallowed-names = ["toto", "tata", "titi"]
 cognitive-complexity-threshold = 30
 ```
 
-See the [list of lints](https://rust-lang.github.io/rust-clippy/master/index.html) for more information about which
-lints can be configured and the meaning of the variables.
+See the [list of configurable lints](https://rust-lang.github.io/rust-clippy/master/index.html#Configuration),
+the lint descriptions contain the names and meanings of these configuration variables.
 
 > **Note**
 >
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -197,8 +197,8 @@ disallowed-names = ["toto", "tata", "titi"]
 cognitive-complexity-threshold = 30
 ```
 
-See the [list of lints](https://rust-lang.github.io/rust-clippy/master/index.html) for more information about which
-lints can be configured and the meaning of the variables.
+See the [list of configurable lints](https://rust-lang.github.io/rust-clippy/master/index.html#Configuration)
+for more information about the meaning of the variables.
```
xFrednet, line 201: Do we maybe want to include a comment that they have to open the lint descriptions to see the configs?

```suggestion
open the lint description for more information about the meaning of these variables.
```

Alexendoo: I'm struggling to come up with a phrasing that doesn't sound awkward

xFrednet: What about a simple: "See the _list of configurable lints_, the lint description contains the name and effect of these variables"

We could also try an alternative and add a new part to the list/book, which lists all config variable as first class citizens. This is less related to this change and more a general idea

xFrednet (approved):

xFrednet:
LGTM, thank you, let's hope that it helps a bit :)

@bors r+

Issue #9808, opened 2022-11-07
`indexing_slicing` is not triggered in `const fn`s


KizzyCode:
### Summary

The `indexing_slicing`-lint is not triggered within `const fn`s. In the following example, `is_negative` performs indexing on a slice (`self.value[0]`) which should get linted but isn't. However if I remove the `const`-attribute for `is_negative`, the lint gets triggered as expected.

```rust
#![warn(clippy::indexing_slicing)]

/// An opaque integer representation
pub struct Integer<'a> {
    /// The underlying data
    value: &'a [u8],
}
impl<'a> Integer<'a> {
    // Check whether `self` holds a negative number or not
    pub const fn is_negative(&self) -> bool {
        self.value[0] & 0b1000_0000 != 0
    }
}
```

### Lint Name

indexing_slicing

### Reproducer

I tried this code:

```rust
#![warn(clippy::indexing_slicing)]

/// An opaque integer representation
pub struct Integer<'a> {
    /// The underlying data
    value: &'a [u8],
}
impl<'a> Integer<'a> {
    // Check whether `self` holds a negative number or not
    pub const fn is_negative(&self) -> bool {
        self.value[0] & 0b1000_0000 != 0
    }
}
```

I expected to see this happen: The lint warns me on `self.value[0]`.

Instead, this happened: The lint does not warn me on `self.value[0]` 😅

### Version

```text
rustc 1.65.0 (897e37553 2022-11-02)
binary: rustc
commit-hash: 897e37553bba8b42751c67658967889d11ecd120
commit-date: 2022-11-02
host: aarch64-apple-darwin
release: 1.65.0
LLVM version: 15.0.0
```

xFrednet:
This is intentional, since all alternatives (`get()`, `expect()`) are currently sadly not possible in constant blocks. We could add a config value, to have such lints still report in const context. Would that solution work for you? :upside_down_face:

KizzyCode:
Yes, this would work 😊
I use clippy more as a "reminder" with `#[allow(...)]` and "This is safe because ..."-comments (similar to `unsafe`); so for me, even if there is no other way in `const`-contexts, it forces me to double check if my assumptions really make sense.

xFrednet:
> Additionally it's probably a good idea to explicitly document this behavior?

Agreed, it might be good to also check other lints for this.

---

Adding the config value and adding the documentation should be fairly simple.

@rustbot label +good-first-issue

naosense:
[...]

PR #9920, opened 2022-11-21
add `suppress_restriction_lint_in_const` config

naosense:
According to #9808 , add a new lint `suppress_lint_in_const` to report even in const context. BTW, i am not good at naming either, if anyone have a better idea, i am happy to change it.

This PR is still in progress, so i keep it draft. 

- \[x] Run `cargo dev fmt`

changelog: Enhancement: [`indexing_slicing`]: add new config `suppress-restriction-lint-in-const` to enable restriction lints, even if the suggestion might not be applicable

r? @xFrendet


naosense:
Hey @xFrednet , i think it's ready to go. Could you please review it?

clippy_lints/src/utils/conf.rs
```diff
@@ -402,6 +402,10 @@ define_Conf! {
     /// A list of paths to types that should be treated like `Arc`, i.e. ignored but
     /// for the generic parameters for determining interior mutability
     (ignore_interior_mutability: Vec<String> = Vec::from(["bytes::Bytes".into()])),
+    /// Lint: INDEXING_SLICING
+    ///
+    /// Whether to suppress lint in const function
+    (suppress_lint_in_const: bool = true),
```
xFrednet, line 408: I would include the lint category in the lint name, as it could be possible, that users want restriction lints to trigger in every context while style lints, only if it can be rewritten.

Then we could also think about defaulting this to false, but having it on true is the safe choice :+1:

tests/ui-toml/suppress_lint_in_const/clippy.toml
```diff
@@ -0,0 +1 @@
+suppress-lint-in-const = false
```
xFrednet, line 1: Formatting nit, could you add a new line here?

clippy_lints/src/indexing_slicing.rs
```diff
@@ -146,7 +157,7 @@ impl<'tcx> LateLintPass<'tcx> for IndexingSlicing {
                 // Catchall non-range index, i.e., [n] or [n << m]
                 if let ty::Array(..) = ty.kind() {
                     // Index is a const block.
-                    if let ExprKind::ConstBlock(..) = index.kind {
+                    if self.suppress_lint_in_const && let ExprKind::ConstBlock(..) = index.kind {
                         return;
                     }
```
xFrednet, line 162: The lint emission in line 155 and 170 both use `span_lint_and_help` and suggest using `.get(n)` or `.get_mut(n)`. This suggestion is not applicable in constant context. Could you use [`span_lint_and_then()`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/diagnostics/fn.span_lint_and_then.html) and construct a message using the diagnostic builder? In constant context, it would be good to either not emit the suggestion or add another note saying: "the suggestion might not be applicable in constant blocks". From these options, I prefer the second one.

clippy_lints/src/utils/conf.rs
```diff
@@ -402,6 +402,10 @@ define_Conf! {
     /// A list of paths to types that should be treated like `Arc`, i.e. ignored but
     /// for the generic parameters for determining interior mutability
     (ignore_interior_mutability: Vec<String> = Vec::from(["bytes::Bytes".into()])),
+    /// Lint: INDEXING_SLICING
+    ///
+    /// Whether to suppress lint in const function
```
xFrednet, line 407: I think here it's important to document, why a user might want/not want to enable this. Presently, it sounds like there is no reason to set it to true.

```suggestion
    /// Whether to suppress a restriction lint in constant code. In same
    /// cases the restructed operation might not be unavoidable, as the
    /// suggested counterparts are unavailable in constant code. This
    /// configuration will cause restriction lints to trigger even
    /// if no suggestion can be made.
```

Something like that, I feel like it's not perfect yet, but good enough ^^

xFrednet (commented):
Really solid start! Thank you for working on this issue :)

I hope that these suggestions are fairly simple. You're welcome to reach out if you'd like any help :)

clippy_lints/src/utils/conf.rs
naosense: sorry i dont quite get it, do you mean the name should be something like `suppress_{category}_lint_in_const`?

clippy_lints/src/indexing_slicing.rs
naosense: Do you mean replace `span_lint_and_help` with with [`span_lint_and_then()`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/diagnostics/fn.span_lint_and_then.html) in line 155 and 170?

xFrednet: Yes, then you can include optional information in the closure. Maybe try to search clippy for an example of [`span_lint_and_then`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/diagnostics/fn.span_lint_and_then.html)

naosense:
@xFrednet thanks for your kindness and patience! Not quite sure everything is corrected as you requested, could you please review again?

clippy_lints/src/indexing_slicing.rs
```diff
@@ -141,12 +155,19 @@ impl<'tcx> LateLintPass<'tcx> for IndexingSlicing {
...
                 };
 
-                span_lint_and_help(cx, INDEXING_SLICING, expr.span, "slicing may panic", None, help_msg);
+                span_lint_and_then(cx, INDEXING_SLICING, expr.span, "slicing may panic", |diag| {
+                    let note = if cx.tcx.hir().is_inside_const_context(expr.hir_id) {
+                        "the suggestion might not be applicable in constant blocks"
+                    } else {
+                        ""
+                    };
+                    diag.span_suggestion(expr.span, help_msg, note, Applicability::MachineApplicable);
```
xFrednet, line 164: The applicability is used to indicate if a suggestion can be automatically applied, meaning that the marked span will be replaced by the suggestion. In this case, we don't have an exact suggestion and only a message suggesting the usage of `get(n)` or `get_mut(n)`. Which can't automatically be applied. Therefore, the applicability should be `Unspecified`. This is just some general information.

This is going into the right direction. The `is_inside_const` check is excellent! I was thinking of a different styling, that can be achieved with the following code:

```rs
span_lint_and_then(cx, INDEXING_SLICING, expr.span, "slicing may panic", |diag| {
    diag.help(help_msg);

    if cx.tcx.hir().is_inside_const_context(expr.hir_id) {
        diag.note("the suggestion might not be applicable in constant blocks");
    }
});
```

The same goes for the other `span_lint_and_then`. This one also doesn't require the specification of the applicability. You can also store the note message in a variable, like the current `help_msg` to ensure that both messages stay the same. Then this should be perfect!

xFrednet (commented):
> @xFrednet thanks for your kindness and patience!

You're very welcome. Thank you for your kindness and contribution as well!

> Not quite sure everything is corrected as you requested, could you please review again?

Yes, I have one suggestion, and then it should be good to go!

naosense:
ping ping @xFrednet , let me know if anything wrong with my code.😄

xFrednet:
Hey @naosense, everything is fine. It can sometimes take a few days, until I and other team members have time to review a PR. If you don't hear anything, it's nothing bad and a ping after some time can be helpful.

I generally aim to give a review within 5 days, but it can go up to a week, if life gets busy or there are several other reviews waiting. This one is the next proper review in my queue. Maybe you'll get a review today, but it's more likely that it'll be tomorrow. :)

naosense:
@xFrednet got it, didn't know that before, I'm fine with any time, just follow your schedule

tests/ui/indexing_slicing_index.stderr
```diff
@@ -1,3 +1,22 @@
+error: indexing may panic
+  --> $DIR/indexing_slicing_index.rs:9:20
+   |
+LL | const REF: &i32 = &ARR[idx()]; // Ok, should not produce stderr.
```
xFrednet, line 4: These new messages are a result of the change, as the default for `suppress-restriction-lint-in-const` is false. I believe this default makes sense. However, then these comments should be adjusted. Ideally, they should say that is should be linted, since the config value is set.

tests/ui-toml/suppress_lint_in_const/clippy.toml
```diff
@@ -0,0 +1 @@
+suppress-restriction-lint-in-const = false
```
xFrednet, line 1: This test should test the non default value. Could you set it to true and make sure to include all the tests from `tests/ui/indexing_slicing_index.rs` which were previously suppressed?

xFrednet (commented):
The implementation looks good! The tests still need some adjustment and then we can merge this :)

xFrednet:
Hey @naosense 

> got it, didn't know that before, I'm fine with any time, just follow your schedule

No problem. It's something I needed to get used to as well. You're welcome to reach out to me here or on Zulip if you're unsure about these things :)

clippy_lints/src/utils/conf.rs
```diff
@@ -406,6 +406,14 @@ define_Conf! {
...
     /// Whether to allow mixed uninlined format args, e.g. `format!("{} {}", a, foo.bar)`
     (allow_mixed_uninlined_format_args: bool = true),
+    /// Lint: INDEXING_SLICING
+    ///
+    /// Whether to suppress a restriction lint in constant code. In same
+    /// cases the restructured operation might not be unavoidable, as the
+    /// suggested counterparts are unavailable in constant code. This
+    /// configuration will cause restriction lints to trigger even
+    /// if no suggestion can be made.
+    (suppress_restriction_lint_in_const: bool = false),
```
xFrednet, line 416: The description sounds perfect :)

clippy_lints/src/indexing_slicing.rs
```diff
@@ -141,12 +155,18 @@ impl<'tcx> LateLintPass<'tcx> for IndexingSlicing {
...
+
+                    if cx.tcx.hir().is_inside_const_context(expr.hir_id) {
+                        diag.note(note);
+                    }
+                });
             } else {
                 // Catchall non-range index, i.e., [n] or [n << m]
                 if let ty::Array(..) = ty.kind() {
                     // Index is a const block.
-                    if let ExprKind::ConstBlock(..) = index.kind {
```
xFrednet, line 149: After reading the `.stderr` output again, I understand this check better. This is exclusive for arrays. Arrays can be bounds checked during compile time, which makes indexing with const index values safe again.

This currently triggers the uitest in line 28 and 29 which it shouldn't. Could you remove the `self.suppress_restriction_lint_in_const` check again from this if?

xFrednet (commented):
Almost perfect, I found one last nit, which should be simple to fix.

Could you then squash the `fix test` commits into one? Afterwards, we can merge this. Thank you for the effort you put into this, it's highly appreciated :)

xFrednet:
This looks good to me. Thank you for all the work you put into this and the quick updates. I hope you had fun working on this. It's an excellent addition to Clippy IMO. 

@bors r+

xFrednet:
Looks like a fluke.

@bors retry

Be nice and let the CI be green this time :)

xFrednet:
It looks like something in the CI is broken. This is not specific to this PR. I'll merge it, once the problem has been resolved :)

xFrednet:
Let's give this another try

@bors r+

PR #9943, opened 2022-11-25
manual_let_else: keep macro call on suggestion blocks

### LINKED ISSUE #9940: `manual_let_else` suggestion expands macros
### Summary

If a macro has a `return` statement, the suggestion from the lint will contain the expanded macro content, rather than the macro call.

### Reproducer

I tried this code:

```rust

macro_rules! some_macro {
    () => {
        return ()
    };
}

fn main() {
    let foo = Some(1);

    let _value = match foo {
        Some(value) => value,
        _ => some_macro!()
    };
}
```

```rust

macro_rules! some_macro {
    () => {
        return ()
    };
}

fn main() {
    let foo = Some(1);

    let Some(value) = foo else { some_macro!() };
}
```
This is the suggestion it should emit, the macro is not expanded.

Instead, this happened:

```rust

macro_rules! some_macro {
    () => {
        return ()
    };
}

fn main() {
    let foo = Some(1);

    let Some(value) = foo else { return () };
}
```

The suggestion removes the `some_macro!()`, and replaces it with the expansion.

### Version

```text
rustc 1.67.0-nightly (70f8737b2 2022-11-23)
binary: rustc
commit-hash: 70f8737b2f5d3bf7d6b784fad00b663b7ff9feda
commit-date: 2022-11-23
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

dswij:
Closes #9940

changelog: [`manual_let_else`]: Do not expand macro calls on suggestions

### DIFF
--- a/clippy_lints/src/manual_let_else.rs
+++ b/clippy_lints/src/manual_let_else.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::higher::IfLetOrMatch;
 use clippy_utils::msrvs::{self, Msrv};
 use clippy_utils::peel_blocks;
-use clippy_utils::source::snippet_opt;
+use clippy_utils::source::{snippet, snippet_with_macro_callsite};
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::visitors::{for_each_expr, Descend};
 use if_chain::if_chain;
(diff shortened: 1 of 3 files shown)


Jarcho:
`snippet_with_context` would be better here. `snippet_with_macro_callsite` should be removed since it doesn't handle nested macro calls.

Jarcho:
Thank you. @bors r+

PR #9948, opened 2022-11-25
Support user format-like macros

nyurik:
Add support for `#[clippy::format_args]` attribute that can be attached to any macro to indicate that it functions the same as the built-in format macros like `format!`, `println!` and `write!`

---

changelog: Enhancement: [`format_in_format_args`], [`recursive_format_impl`], [`to_string_in_format_args`], [`uninlined_format_args`], [`unused_format_specs`]: Recognizes `#[clippy::format_args]` to support custom 3rs party format macros.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -7,6 +7,7 @@
 - [Configuration](configuration.md)
     - [Lint Configuration](lint_configuration.md)
 - [Clippy's Lints](lints.md)
+- [Attributes for Crate Authors](attribs.md)
 - [Continuous Integration](continuous_integration/README.md)
     - [GitHub Actions](continuous_integration/github_actions.md)
     - [GitLab CI](continuous_integration/gitlab.md)
--- a/book/src/attribs.md
+++ b/book/src/attribs.md
@@ -1,0 +1,53 @@
+# Attributes for Crate Authors
+
+In some cases it is possible to extend Clippy coverage to 3rd party libraries.
+To do this, Clippy provides attributes that can be applied to items in the 3rd party crate.
+
+## `#[clippy::format_args]`
+
+_Available since Clippy v1.84_
+
+This attribute can be added to a macro that supports `format!`, `println!`, or similar syntax.
+It tells Clippy that the macro is a formatting macro, and that the arguments to the macro
+should be linted as if they were arguments to `format!`. Any lint that would apply to a
+`format!` call will also apply to the macro call. The macro may have additional arguments
+before the format string, and these will be ignored.
+
+### Example
+
+```rust
+/// A macro that prints a message if a condition is true.
+#[macro_export]
+#[clippy::format_args]
+macro_rules! print_if {
+    ($condition:expr, $($args:tt)+) => {{
+        if $condition {
+            println!($($args)+)
+        }
+    }};
+}
+```
+
+## `#[clippy::has_significant_drop]`
+
+_Available since Clippy v1.60_
+
+The `clippy::has_significant_drop` attribute can be added to types whose Drop impls have an important side effect,
+such as unlocking a mutex, making it important for users to be able to accurately understand their lifetimes.
+When a temporary is returned in a function call in a match scrutinee, its lifetime lasts until the end of the match
+block, which may be surprising.
+
+### Example
+
+```rust
+#[clippy::has_significant_drop]
+struct CounterWrapper<'a> {
+    counter: &'a Counter,
+}
+
+impl<'a> Drop for CounterWrapper<'a> {
+    fn drop(&mut self) {
+        self.counter.i.fetch_sub(1, Ordering::Relaxed);
+    }
+}
+```
--- a/clippy_utils/src/attrs.rs
+++ b/clippy_utils/src/attrs.rs
@@ -27,7 +27,10 @@ pub const BUILTIN_ATTRIBUTES: &[(&str, DeprecationStatus)] = &[
     ("cyclomatic_complexity", DeprecationStatus::Replaced("cognitive_complexity")),
     ("dump",                  DeprecationStatus::None),
     ("msrv",                  DeprecationStatus::None),
+    // The following attributes are for the 3rd party crate authors.
+    // See book/src/attribs.md
     ("has_significant_drop",  DeprecationStatus::None),
+    ("format_args",           DeprecationStatus::None),
 ];
 
 pub struct LimitStack {
--- a/clippy_utils/src/macros.rs
+++ b/clippy_utils/src/macros.rs
@@ -1,5 +1,6 @@
 #![allow(clippy::similar_names)] // `expr` and `expn`
 
+use crate::get_unique_attr;
 use crate::visitors::{Descend, for_each_expr_without_closures};
 
 use arrayvec::ArrayVec;
@@ -7,7 +8,7 @@ use rustc_ast::{FormatArgs, FormatArgument, FormatPlaceholder};
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::sync::{Lrc, OnceLock};
 use rustc_hir::{self as hir, Expr, ExprKind, HirId, Node, QPath};
-use rustc_lint::LateContext;
+use rustc_lint::{LateContext, LintContext};
 use rustc_span::def_id::DefId;
 use rustc_span::hygiene::{self, MacroKind, SyntaxContext};
 use rustc_span::{BytePos, ExpnData, ExpnId, ExpnKind, Span, SpanData, Symbol, sym};
@@ -36,7 +37,9 @@ pub fn is_format_macro(cx: &LateContext<'_>, macro_def_id: DefId) -> bool {
     if let Some(name) = cx.tcx.get_diagnostic_name(macro_def_id) {
         FORMAT_MACRO_DIAG_ITEMS.contains(&name)
     } else {
-        false
+        // Allow users to tag any macro as being format!-like
+        // TODO: consider deleting FORMAT_MACRO_DIAG_ITEMS and using just this method
+        get_unique_attr(cx.sess(), cx.tcx.get_attrs_unchecked(macro_def_id), "format_args").is_some()
     }
 }
 
--- a/tests/ui/uninlined_format_args.rs
+++ b/tests/ui/uninlined_format_args.rs
@@ -262,8 +262,6 @@ fn tester2() {
     my_concat!("{}", local_i32);
     my_good_macro!("{}", local_i32);
     my_good_macro!("{}", local_i32,);
-
-    // FIXME: Broken false positives, currently unhandled
     my_bad_macro!("{}", local_i32);
     my_bad_macro2!("{}", local_i32);
     used_twice! {
@@ -271,4 +269,23 @@ fn tester2() {
         small = "small value: {}",
         local_i32,
     };
+}
+
+#[clippy::format_args]
+macro_rules! usr_println {
+    ($target:expr, $($args:tt)*) => {{
+        if $target {
+            println!($($args)*)
+        }
+    }};
+}
+
+fn user_format() {
+    let local_i32 = 1;
+    let local_f64 = 2.0;
+
+    usr_println!(true, "val='{}'", local_i32);
+    usr_println!(true, "{}", local_i32);
+    usr_println!(true, "{:#010x}", local_i32);
+    usr_println!(true, "{:.1}", local_f64);
 }
--- a/tests/ui/format_args_unfixable.rs
+++ b/tests/ui/format_args_unfixable.rs
@@ -119,3 +119,32 @@ fn test2() {
         format!("something failed at {}", Location::caller())
     );
 }
+
+#[clippy::format_args]
+macro_rules! usr_println {
+    ($target:expr, $($args:tt)*) => {{
+        if $target {
+            println!($($args)*)
+        }
+    }};
+}
+
+fn user_format() {
+    let error = Error::new(ErrorKind::Other, "bad thing");
+    let x = 'x';
+
+    usr_println!(true, "error: {}", format!("boom at {}", Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+    usr_println!(true, "{}: {}", error, format!("boom at {}", Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+    usr_println!(true, "{:?}: {}", error, format!("boom at {}", Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+    usr_println!(true, "{{}}: {}", format!("boom at {}", Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+    usr_println!(true, r#"error: "{}""#, format!("boom at {}", Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+    usr_println!(true, "error: {}", format!(r#"boom at "{}""#, Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+    usr_println!(true, "error: {}", format!("boom at {} {0}", Location::caller()));
+    //~^ ERROR: `format!` in `usr_println!` args
+}
--- a/tests/ui/format_args_unfixable.stderr
+++ b/tests/ui/format_args_unfixable.stderr
@@ -174,5 +174,68 @@ LL |     panic!("error: {}", format!("something failed at {}", Location::caller(
    = help: combine the `format!(..)` arguments with the outer `panic!(..)` call
    = help: or consider changing `format!` to `format_args!`
 
-error: aborting due to 18 previous errors
+error: `format!` in `usr_println!` args
+  --> tests/ui/format_args_unfixable.rs:136:5
+   |
+LL |     usr_println!(true, "error: {}", format!("boom at {}", Location::caller()));
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = help: combine the `format!(..)` arguments with the outer `usr_println!(..)` call
+   = help: or consider changing `format!` to `format_args!`
+
+error: `format!` in `usr_println!` args
+  --> tests/ui/format_args_unfixable.rs:138:5
+   |
+LL |     usr_println!(true, "{}: {}", error, format!("boom at {}", Location::caller()));
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = help: combine the `format!(..)` arguments with the outer `usr_println!(..)` call
+   = help: or consider changing `format!` to `format_args!`
+
+error: `format!` in `usr_println!` args
+  --> tests/ui/format_args_unfixable.rs:140:5
+   |
+LL |     usr_println!(true, "{:?}: {}", error, format!("boom at {}", Location::caller()));
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = help: combine the `format!(..)` arguments with the outer `usr_println!(..)` call
+   = help: or consider changing `format!` to `format_args!`
+
+error: `format!` in `usr_println!` args
+  --> tests/ui/format_args_unfixable.rs:142:5
+   |
+LL |     usr_println!(true, "{{}}: {}", format!("boom at {}", Location::caller()));
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = help: combine the `format!(..)` arguments with the outer `usr_println!(..)` call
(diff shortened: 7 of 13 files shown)


Alexendoo:
There are not uncommon situations where inline args can't be used, such as

```rust
macro_rules! error {
    ($fmt:literal $(, $e:expr)* $(,)?) => { println!(concat!("ERROR: ", $fmt), $($e,)*) }
}

fn main() {
    let x = 1;
    error!("{}", x);
}
```

Another thing is that it could break macro invocations themselves, initially I wrote that as `($fmt:literal, $($e:expr),*)` where replacing it with `error!("{x}")` would result in `error: unexpected end of macro invocation`

nyurik:
@Alexendoo I just added your example to my unit tests - it passes just fine.  Do you have a test that fails?

Alexendoo:
`concat!` does seem to be safe since it gets the correct span, for the second one an example would be

```rust
macro_rules! error {
    ($fmt:literal, $($e:expr),*) => { println!($fmt, $($e,)*) }
}

fn main() {
    let x = 1;
    error!("{}", x);
}
```

nyurik:
@Alexendoo that's an excellent catch, thank you!  The issue here is with the bug in macro definition (granted that macros are a special beast and a "bug" could be someone's feature).  TBH, I am surprised there is no lint that tries to catch this mistake. The bigger question though is if this is something we should or should not process?   I.e. something that suggests a cleaner repeated macro arg catch.  I would even make this as a `suspicious` category.

That said, on one hand we have badly created user macros that probably should be fixed. On the other, there is a very common pattern with all sorts of log and other macros that use format strings and could really use this lint. Which of these takes priority?

nyurik:
@Alexendoo I created a lint suggestion #9959 that would solve this types of issues - having both lints on at the same time would fix the issue i think

Alexendoo:
It would be neat to be able to lint macro definitions, however as they are [almost entirely represented](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/struct.MacroDef.html) as a big `TokenStream` that may be expanded into an arbitrary place linting them isn't easy - https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Get.20paths.20in.20a.20macro/near/292040802

Another one to bring up is that a common way of defining things like this is two arms, so that lint wouldn't be generally applicable

```rust
($fmt:literal) => ...
($fmt:literal, $($e:expr),*) => ...
```

e.g. the `format_args` `macro_rules` stub is [defined like that](https://doc.rust-lang.org/std/macro.format_args.html). There are macros out there that assume the single arg version is a plain string much like the old `panic!` macro, I found this one in a search https://github.com/scottlamb/moonfire-nvr/blob/d509d3ff40160a58e4fbdd01be240f2be474b693/server/base/error.rs#L159-L166

That would make the suggestion change the behaviour of the code, silently even if the printed variable is otherwise used

My biggest concern is probably all the proc macros out there, the current implementation is not super robust against them. If a proc macro did an internal format!() but happened to use certain input spans we could be matching up a completely different format string than the one actually being expanded

To that end it may be worth keeping an eye on https://github.com/rust-lang/rust/issues/99012 / https://github.com/rust-lang/compiler-team/issues/541

Alexendoo:
Another fun one:

```rust
macro_rules! used_twice {
    (
        large = $large:literal,
        small = $small:literal,
        $val:expr,
    ) => {{
        if $val < 5 {
            println!($small, $val);
        } else {
            println!($large, $val);
        }
    }}
}

fn main() {
    let x = 1;
    used_twice! {
        large = "large value: {}",
        small = "small value: {}",
        x,
    }
}
```

nyurik:
I have a better idea - let's disable Rust macros, as they are clearly a work of Cthulhu, designed to steer us away from the light... as this is clearly a case of "I miss Perl, so I will make things as interesting as possible".

I don't think Clippy should care about such cases TBH as they probably represent 1% - when on the other hand we have 99% of the simple `log::warn!` type.  Moreover, if our lint is used often enough, the 1% edge cases should disappear out right as abominations against humanity. That said, we may want to detect such cases (e.g. have a list of well known crates), and if the first macro in a call chain is not in the well known list, reduce the certainty of replacement from `MachineApplicable` to `MaybeIncorrect`.

llogiq:
Please keep in mind that we deem false positives a worse outcome than false negatives. I'd be OK with having either a config that makes the lint apply everywhere instead of just format while making the suggestion `MaybeIncorrect` or having a second lint that is `allow`ed by default that catches those cases aside `format` though, with a slight preference towards the latter.

nyurik:
@llogiq I think we should also take into account how often we get false positive vs false negative. I agree that if they are about the same, false positive is worse, but if we get 100x more false negatives, I would prefer get a few more false positives.

I like the idea of an extra config param, something like `include_custom_format_macros` (naming?) - this way all lints that currently use `is_format_arg` would handle 3rd-party format macros.

llogiq:
I would fully support that argument for a `suspicious` lint, but not for a `style` lint. Please note that a lint rarely works in isolation – clippy usually runs many of them. If one of them gives even a modest amount of false positives, this reduces the usefulness of clippy as a whole. While that may be acceptable for code that is likely problematic and warrants a closer look, I don't think it is needed for `style` lints which IMHO are a teaching tool.

With that said, as long as the user explicitly asks for it, I'm ok with changing this balance. Be careful with suggestion applicability though.

nyurik:
@Alexendoo and @llogiq I made some changes to this PR - is this what you had in mind?

llogiq:
Ok, that's the minimal solution. I had a list of paths in mind that should be treated as format!-like macros, possibly with `*` wildcards, which would give users full flexibility.

nyurik:
> Ok, that's the minimal solution. I had a list of paths in mind that should be treated as format!-like macros, possibly with `*` wildcards, which would give users full flexibility.

What do you mean by `paths`?  Fully qualified macro names, or namespaces like `mycrate::*`? Seems a bit excessive tbh. Do any other lint provide something like that? Besides, I would rather start small first - MVP, get it merged and tried by users, and if there is a demand, build on top of it.

llogiq:
The base idea was to have fully qualified paths and then perhaps extend that to allow for wild cards.

But I agree, doing the minimal thing first is entirely acceptable.

Alexendoo:
Do we have a good idea of what the most common uses are? I would be in favour of adding e.g. `log` + `tracing` to the whitelist if it caught most of the cases where it's used in the wild

nyurik:
I'm all for adding more well known format like macros to the allow list, but I think that should be an independent PR. This pr is for the ones that are not explicitly listed

xFrednet:
What do you think about this not being a `true` / `false` flag, but a list of allowed macros instead? This would give users a lot more flexibility, but require some extra implementation work. The config could also be used for the `std`, `log` and `tracing` macros. For other config lists, we have [the `..` syntax](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_names), which means that the defined macros should only expand the default configuration.

Sorry, for suggestion this just now. Taking this version is also fine with me, I just wanted to throw in the idea :)

xFrednet:
@nyurik Thank you, also for all the work you put into the formatting lints!

I suggested this here, as the goal is generally to keep the lint configs stable. We can later add a new configuration value, but removing them should be avoided. That's why Clippy still has old values marked as deprecated. Therefore, I'm not quite sure how to split this in a good way. Adding a config value like this _should theoretically be simple_ :tm:. Though I'm not the reviewer, so suggestion a change that adds another x lines is a bit simpler ^^. I also know that I oversimplify these changes in my head. Are there any more thoughts on this? :)

nyurik:
@xFrednet config values are parsed with Serde.  I think we can start with a bool, and later use an enum to extend it:

```rust
/// config declaration
include_custom_format_macros: IncludeCustomFormatMacrosEnum,

...

#[derive(Debug, Clone, PartialEq, Deserialize)]
#[serde(untagged)]
pub enum IncludeCustomFormatMacrosEnum {
    Bool(bool),
    List(Vec<String>),
}
```

This way users can use either true/false or a list of strings.  If we want, we can go even further with a custom parser.

xFrednet:
I wasn't aware of this option with Serde ^^

Then this is definitely an option. My suggestion at first, also suggested including the `println!` and `write!` macros in the list, to give users a lot of control. Then a different name would be better. However, this level of control is probably not needed. So SGTM :)

llogiq:
@Alexendoo I'd also be OK with merging this now and extending in a follow up PR. What do you think?

llogiq:
Ok, since no one is stepping up, I say let's just merge this now and extend later.

@bors r+

Alexendoo:
Sorry! lost the notification over the holidays, personally I'm not keen on this approach so @bors r-

will type up a reason shortly

Alexendoo:
🤔

@bors ping

Alexendoo:
Ah it did work, r- doesn't have a reply

Okay so for me I don't feel that it's an urgent enough issue to have a temporary solution, I'd rather we go straight to a version that solves it for most people without requiring them to add anything to a `clippy.toml`. That way we don't have to carry the doc/implementation complexity forward or worry about macros triggering it erroneously even if they did opt in 

I picture that as being either a `#[clippy::format_something_something]` attribute that is applied to macro definitions or the previously mentioned `Vec<String>` config that is prepopulated with a list of third party macros

The next meeting is in a few days, so let's get consensus there on what direction we want to go in

nyurik:
@Alexendoo is there a way to subscribe to the meeting calendar events?

Alexendoo:
Yep, you can ask for a calendar invite in the zulip channel

The meeting is on the 10th at 16:00 UTC

m-ou-se:
Hey hey, I figured y'all should know that https://github.com/rust-lang/rust/pull/106745 is about to be merged. It adds `rustc_ast::FormatArgs`, which contains all the information clippy needs, so that clippy no longer needs to go hunt for patterns like `ArgumentV1::new_*` in the hir.

nyurik:
Thx @m-ou-se, this is indeed exciting!  Sounds like you will be able to do a number of additional optimizations once format params are part of AST.

My concern is that it still won't solve the corner cases listed above where inlining a var makes the macro behave differently.  Maybe we should consider adding a core feature attribute to indicate something is a format-args macro, instead of adding an attribute to clippy alone?  Not certain if it would help though...

nyurik:
Took a bit of a year+ pause on this one.. Still, seems there is plenty of interest in this, so now this lint is rebased and cleaned up a bit. The question still remains: how can we progress.  Perhaps someone has a better idea after Clippy has improved so much over the last year?  @Alexendoo had some comments back then - do we still want to go with a one-per-project allow-list of valid crates? Feels really messy. 

Maybe we should duplicate all format lints, e.g. create `uninlined_format_args_ext` in addition to `uninlined_format_args` -- the `_ext` would indicate this lint works on 3rd party crates.  And this way that lint can be `suspicious`.

My original thoughts were [posted](https://github.com/rust-lang/rust-clippy/issues/10165#issuecomment-1377730134) in the log crate.

## automatic
Fully automatic - decide at the linting time if `my_format!("{}", foo}` is the same as `my_format!("{foo}")` - i.e. the "what-if" macro expansion - this way the lints will try the proposed substitutions, and if the resulting code is the same as original, make the suggestion.

**PROs**: easiest to use everyone, and this approach can be used by many other types of lints to validate replacements instead of declaring them as `MachineApplicable` by the lint authors.
**CONs**: hardest to implement, only works for the lints which offer a `MachineApplicable` replacement

## attribute
Create an attribute that macro developers can use to indicate that their macro is format! compatible:

```
#[format]
some_macro!( [...,] string_literal [, args]* )
```

**PROs**: allows macro authors to declare a macro as OK for Clippy to process as a format-like macro for many lints
**CONs**: existing libs don't have it, possibly will raise some warnings if used with older versions, a bit chatty

## configuration-based
add a Clippy config param for end users to list all macros they will want to inline, plus provide a built-in list of all well-known format-like macros in popular crates. If the user wants to extend the built-in ones, they can just use the `format-like-macros: ["..", "some-other-mods", "and-macros"]` syntax. This approach was implemented as a bool flag in #9948.

This may require more "magic" options:  `"*"` = include all, `"-mod.macro"` = exclude a specific mod or macro, etc.

**PROs**: easy to configure by end user, can be applied to older crates
**CONs**: Clippy team would have to maintain it, each user would need to manually add the ones not on the built-in list, macro writers have no way to control the built-in list for a specific version

Alexendoo:
My preferred solution would be the tool attribute applied to macro definitions, e.g.

```rust
#[clippy::format_args_compatible]
macro_rules! log {
// ...
```

Config based relies on assuming third party crate layouts to be widely useful which we're trying not to expand, I think putting it in the macro authors hands is preferable there

If there's a good heuristic that works for automatic detection I would be fine with that, but I'm not sure if there is one

nyurik:
@Alexendoo do you know of any example of an attribute set on a macro rule, and clippy using that attribute while parsing the expansion of the macro?

We could introduce a new `pedantic` (?) lint that would tag all cases as suspiscious if they resolve to `format_args!` internally, but there are guaranteed to be rare false positives.  We could even add a link to a guide on how to write proper format macros so they won't break if they use a single string without args.

Alexendoo:
I don't think we have any cases where it's used, but `MacroCall` contains the macro's `DefId` which can be used to retrieve its attributes

Linting macro definitions is considered too difficult because the only thing you get from unexpanded macros is a token stream

xFrednet:
Hey, this is a ping from triage. @Alexendoo  can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@rustbot ready

Alexendoo:
Ah sorry missed that notification @xFrednet, my position is still if we're doing this it should be through an attribute rather than configuration

nyurik:
@Alexendoo i updated this PR to rely on `#[clippy::format_args]` attribute

tests/ui/format_args_unfixable_user.rs
Alexendoo, line 1: This file produces no lints but unfixable implies it ought to

tests/ui/uninlined_format_args.rs
```diff
@@ -211,12 +213,14 @@ macro_rules! my_println2_args {
...
+#[clippy::format_args]
 macro_rules! my_concat {
     ($fmt:literal $(, $e:expr)*) => {
         println!(concat!("ERROR: ", $fmt), $($e,)*)
     }
 }
 
+#[clippy::format_args]
 macro_rules! my_good_macro {
     ($fmt:literal $(, $e:expr)* $(,)?) => {
```
Alexendoo, line 225: `#[clippy::format_args]` is applied to many of these macros but only `my_good_macro` is linting, is that intentional?

nyurik:
@Alexendoo thx, I think I fixed all the issues with it now, adding a lot of new tests in various relevant lints.

nyurik:
@xFrednet @Alexendoo @llogiq friendly ping - it seems this 2 years old PR might actually get merged / put to vote, as I don't see any outstanding issues with it?

Alexendoo:
https://rust-lang.zulipchat.com/#narrow/channel/257328-clippy/topic/Can.20we.20start.20FCP.20for.20.60.23.5Bclippy.3A.3Aformat_args.5D.60.20in.20user.20code.3F hasn't had any concerns raised so yeah seems fine

We should ensure this is documented somewhere though

nyurik:
@Alexendoo where would be the best place to document that 3rd party crates can now use this new attribute?  Documentation should also include recommended way to declare format-using macros (i.e. to avoid trailing comma gotchas, etc).  The only concern for 3rd party crates is how will they deal with the older Clippy versions - wouldn't there be some warning if clippy doesn't recognize an attribute?

Alexendoo:
Probably a new section in the book for if we add others

> older Clippy versions - wouldn't there be some warning if clippy doesn't recognize an attribute?

If the crate author uses Clippy I believe they would have to be using one that supports the attribute, but consuming the dep shouldn't show a warning if it's older than the attribute. We should add a test for this though

nyurik:
Thx, rebased and added documentation page to the book.

nyurik:
P.S. using undefined attribute is an error, not a warning - that's why `#[rustversion::attr(since(1.84), clippy::format_args)]` should be the recommended usage pattern.

book/src/lib_usage.md
```diff
@@ -0,0 +1,33 @@
+# Modifying Clippy behavior with attributes
+
+In some cases it is possible extend Clippy coverage to include 3rd party libraries.
+At this moment, only one such modification is possible: adding a
```
flip1995, line 4: There's also the `has_significant_drop` attribute.

clippy_utils/src/attrs.rs
```diff
@@ -28,6 +28,7 @@ pub const BUILTIN_ATTRIBUTES: &[(&str, DeprecationStatus)] = &[
     ("dump",                  DeprecationStatus::None),
     ("msrv",                  DeprecationStatus::None),
     ("has_significant_drop",  DeprecationStatus::None),
```
flip1995, line 30: This and the new attribute are different from the others above: they are meant for crate authors, instead of for clippy users.

Can you add a comment about that, before this line, please? Also to separate them in our code.

clippy_utils/src/macros.rs
```diff
@@ -36,7 +37,9 @@ pub fn is_format_macro(cx: &LateContext<'_>, macro_def_id: DefId) -> bool {
     if let Some(name) = cx.tcx.get_diagnostic_name(macro_def_id) {
         FORMAT_MACRO_DIAG_ITEMS.contains(&name)
     } else {
-        false
+        // Allow users to tag any macro as being format!-like
+        // TODO: consider deleting FORMAT_MACRO_DIAG_ITEMS and using just this method
```
flip1995, line 41: This should be a follow up PR to the rust repo, after this PR is merged and synced

book/src/SUMMARY.md
```diff
@@ -4,6 +4,7 @@
 
 - [Installation](installation.md)
 - [Usage](usage.md)
+    - [Extending Clippy coverage](lib_usage.md)
```
flip1995, line 7: I would make this it's own section, as it is not really how to use clippy (as a tool), but more about how create authors can also benefit from some clippy lints by adding attributes.

Section title(maybe) : Attributes for Crate Authors

book/src/lib_usage.md
```diff
@@ -0,0 +1,33 @@
...
+
+This attribute can be added to a macro that supports `format!`-like syntax.
+It tells Clippy that the macro is a formatting macro, and that the arguments to the macro
+should be linted as if they were arguments to `format!`. Any lint that would apply to a
+`format!` call will also apply to the macro call. The macro may have additional arguments
+before the format string, and these will be ignored.
+
+### Example
+
+Note that the `#[clippy::format_args]` is only available in v1.84, and will
```
flip1995, line 17: Is this also a concern when your using a create as a dependency that uses this macro? 

This note should probably at the top of this chapter, as it applies to all attributes. Then the attribute documentation should contain the version it was added.

flip1995 (commented):
The only thing that still needs to be figured out is the behavior when those macros are used in a dependency. Aka would adding this  (without the `rust version::attr`) raise the MSRV of a crate?

book/src/lib_usage.md
Alexendoo: It should be fine for dependencies, they're compiled with rustc which doesn't know about specific clippy attributes. I do think it's worth adding a test in an aux-build, assuming aux-build uses rustc and not clippy-driver

We probably don't need to recommend `rustversion`, that would only be needed if the library author wants to run an old clippy on the library itself

Jarcho: It's only needed if multiple clippy versions are being run on the workspace containing the attribute. Might be worth noting as an addendum to the section about crate attributes, but not really a thing worth drawing much attention to.

clippy_utils/src/attrs.rs
nyurik: thx @flip1995, added docs for it too

book/src/SUMMARY.md
nyurik: done, thx

book/src/attribs.md
```diff
@@ -0,0 +1,52 @@
+# Attributes for Crate Authors
+
+In some cases it is possible extend Clippy coverage to include 3rd party libraries.
```
flip1995, line 3:
```suggestion
In some cases it is possible to extend Clippy coverage to 3rd party libraries. To do this, Clippy provides attributes that can be applied to items in the 3rd party crate.
```

Please format after applying.

clippy_utils/src/attrs.rs
flip1995: Can you add a comment in the code here please, saying the the following macros are for crate authors?

flip1995 (approved):
LGTM. Just 2 minor comments.

flip1995:
I think this is good to go now. @Alexendoo any other change requests? If not r=me

book/src/attribs.md
```diff
@@ -0,0 +1,53 @@
...
+
+## `#[clippy::format_args]`
+
+_Available since Clippy v1.84_
+
+This attribute can be added to a macro that supports `format!`-like syntax.
+It tells Clippy that the macro is a formatting macro, and that the arguments to the macro
+should be linted as if they were arguments to `format!`. Any lint that would apply to a
+`format!` call will also apply to the macro call. The macro may have additional arguments
+before the format string, and these will be ignored.
```
Alexendoo, line 14: There are lints that apply to `format!` that wouldn't apply to macros marked with the attr, might want to list it as `format!`/`println!`/etc or similar

Alexendoo: e.g. `format_in_format_args`

nyurik: @Alexendoo thx, updated

Alexendoo (approved):

nyurik:
only two years in the making :) . Thx for all the feedback!

PR #9962, opened 2022-11-26
`comparison_to_empty`: peel derefs in suggestions

### LINKED ISSUE #9958: Clippy lint "comparison to empty" fails to apply fix when reference is used
I tried this code:

```rust
fn main() {
    let s = "Hello World!";
    let s1 = &s;
    println!("{}", *s1 != "");
}
```

I expected to see this happen: Clippy applies fix successfully

Instead, this happened: 

Clippy fails to automatically apply fix

<details><summary>Output</summary>
<p>

```
> cargo clippy --fix
    Checking clippy_bug v0.1.0 (/home/mdgaziur001/clippy_bug)
warning: failed to automatically apply fixes suggested by rustc to crate `clippy_bug`

  * src/main.rs

error[E0614]: type `bool` cannot be dereferenced
 --> src/main.rs:4:20
  |
4 |     println!("{}", *s1.is_empty());
  |                    ^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0614`.

warning: comparison to empty slice
 --> src/main.rs:4:20
  |
4 |     println!("{}", *s1 == "");
  |                    ^^^^^^^^^ help: using `is_empty` is clearer and more explicit: `*s1.is_empty()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#comparison_to_empty
  = note: `#[warn(clippy::comparison_to_empty)]` on by default

warning: `clippy_bug` (bin "clippy_bug") generated 1 warning (run `cargo fix --bin "clippy_bug"` to apply 1 suggestion)
warning: `clippy_bug` (bin "clippy_bug" test) generated 1 warning (1 duplicate)
    Finished dev [unoptimized + debuginfo] target(s) in 1.20s
```
</p>
</details>

### Meta

`rustc --version --verbose`:

Stable:

```
binary: rustc
commit-date: 2022-11-02
release: 1.65.0
LLVM version: 15.0.0
```

Beta:

```
rustc 1.66.0-beta.2 (0040709e3 2022-11-20)
binary: rustc
commit-hash: 0040709e39bfa159d099ee14628ad4ac72f91406
commit-date: 2022-11-20
release: 1.66.0-beta.2
LLVM version: 15.0.2
```

Nightly:

```
rustc 1.66.0-beta.2 (0040709e3 2022-11-20)
binary: rustc
commit-hash: 0040709e39bfa159d099ee14628ad4ac72f91406
commit-date: 2022-11-20
release: 1.66.0-beta.2
LLVM version: 15.0.2
```

<p>

```
<backtrace>
```

</p>
</details>

mdgaziur:
This PR fixes #9958. In order to fix the issue, the lint will now peel reference operators and enclose the expression with parentheses when necessary.

changelog: [`comparison_to_empty`]: Peel deref operators in suggestions when necessary

### DIFF
--- a/clippy_lints/src/len_zero.rs
+++ b/clippy_lints/src/len_zero.rs
@@ -1,13 +1,13 @@
 use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::source::snippet_with_applicability;
-use clippy_utils::{get_item_name, get_parent_as_impl, is_lint_allowed};
+use clippy_utils::{get_item_name, get_parent_as_impl, is_lint_allowed, peel_ref_operators};
 use if_chain::if_chain;
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
 use rustc_hir::def_id::DefIdSet;
 use rustc_hir::{
     def_id::DefId, AssocItemKind, BinOpKind, Expr, ExprKind, FnRetTy, ImplItem, ImplItemKind, ImplicitSelfKind, Item,
-    ItemKind, Mutability, Node, TraitItemRef, TyKind,
+    ItemKind, Mutability, Node, TraitItemRef, TyKind, UnOp,
 };
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::{self, AssocKind, FnSig, Ty};
@@ -16,6 +16,7 @@ use rustc_span::{
     source_map::{Span, Spanned, Symbol},
     symbol::sym,
 };
+use std::borrow::Cow;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -428,16 +429,23 @@ fn check_len(
 fn check_empty_expr(cx: &LateContext<'_>, span: Span, lit1: &Expr<'_>, lit2: &Expr<'_>, op: &str) {
     if (is_empty_array(lit2) || is_empty_string(lit2)) && has_is_empty(cx, lit1) {
         let mut applicability = Applicability::MachineApplicable;
+
+        let lit1 = peel_ref_operators(cx, lit1);
+        let mut lit_str = snippet_with_applicability(cx, lit1.span, "_", &mut applicability);
+
+        // Wrap the expression in parentheses if it's a deref expression. Otherwise operator precedence will
+        // cause the code to dereference boolean(won't compile).
+        if let ExprKind::Unary(UnOp::Deref, _) = lit1.kind {
+            lit_str = Cow::from(format!("({lit_str})"));
+        }
+
         span_lint_and_sugg(
             cx,
             COMPARISON_TO_EMPTY,
             span,
             "comparison to empty slice",
             &format!("using `{op}is_empty` is clearer and more explicit"),
-            format!(
-                "{op}{}.is_empty()",
-                snippet_with_applicability(cx, lit1.span, "_", &mut applicability)
-            ),
+            format!("{op}{lit_str}.is_empty()"),
             applicability,
         );
     }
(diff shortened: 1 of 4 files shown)


clippy_lints/src/len_zero.rs
```diff
@@ -425,8 +425,12 @@ fn check_len(
     }
 }
 
-fn check_empty_expr(cx: &LateContext<'_>, span: Span, lit1: &Expr<'_>, lit2: &Expr<'_>, op: &str) {
+fn check_empty_expr(cx: &LateContext<'_>, span: Span, mut lit1: &Expr<'_>, lit2: &Expr<'_>, op: &str) {
     if (is_empty_array(lit2) || is_empty_string(lit2)) && has_is_empty(cx, lit1) {
+        if let ExprKind::Unary(UnOp::Deref, expr) = lit1.kind {
+            lit1 = expr;
+        }
```
dswij, line 432: I think this only covers 1 deref, so 2 derefs or more wouldn't be covered, i.e.

```rust
    let s = "Hello, world!";
    let s1 = &&s;
    println!("{}", **s1 == "");
```

tests/ui/len_zero.rs
```diff
@@ -64,6 +64,10 @@ fn main() {
 
     if "".len() == 0 {}
 
+    let s = "Hello, world!";
+    let s1 = &s;
+    println!("{}", *s1 == "");
+
```
dswij, line 70: Let's also add a case for multiple derefs in the test.

dswij (changes requested):
Thanks for the PR, and welcome to Clippy!

Note that for the changelog, it's user-facing changelog. So we'd expect a short explanation of the change itself, rather than pointing it to an issue number/link. Could you help to change that?

Other than that, it's just a small comment on the change itself, and it should be good to go

clippy_lints/src/len_zero.rs
mdgaziur: I think getting the expression that is deref'd in a loop can be a solution

dswij: That's one way to do it, but we have a util function that derefs expr

https://github.com/rust-lang/rust-clippy/blob/3cf2c17e53d115bd8c233a268c0d0ff0fbb2395e/clippy_utils/src/lib.rs#L2276-L2287

Maybe we can use that instead

mdgaziur: Another thing I noticed is that, parentheses aren't removed. So, something like the following:
```rs
**(s1) == ""
```
will become:
```rs
(s1).is_empty()
```
Is there any way to remove these redundant parentheses?

dswij: We have another util to build suggestions, but it needs a little bit more digging:

https://github.com/rust-lang/rust-clippy/blob/78589ffad2c973f7863fdabc2df6e8ab2b54110f/clippy_utils/src/sugg.rs

IIRC, you can remove parentheses with this.

mdgaziur: Can you tell me what `default` parameter in `Sugg::hir_with_applicability` does?

mdgaziur: I found out that other lints don't remove the parentheses. Is this ok if I ignore parentheses?
AFAIK, the util you've suggested directly takes snippet from source code and has no method to remove unnecessary parentheses.

dswij: I think it depends on the lint whether it will suggest removing the parentheses or not. Ideally, we'd have the parentheses removed.

But it's fine if we merge it without removing it in this PR, since the change is already an improvement as it is.

Let me know if you decide not to include the parentheses change in this PR.

mdgaziur: I think we can ignore parentheses now. Maybe this can be done in future?

dswij:
Thanks for this! @bors r+

PR #9966, opened 2022-11-27
Fix manual_let_else produces a wrong suggestion with or-patterns

### LINKED ISSUE #9938: `manual_let_else` produces a wrong suggestion with or-patterns
### Summary

When using an or-pattern with two different variants, `manual_let_else` will not wrap them in parentheses, which are needed for the pattern to be valid in that context.

### Reproducer

I tried this code:

```rust

enum Foo {
    Bar(i32),
    Baz(i32),
    Qux(i32),
}

fn main() {
    let foo = Foo::Bar(1);

    let _value = match foo {
        Foo::Bar(_) | Foo::Baz(_) => (),
        _ => return,
    };
}
```

```rust

enum Foo {
    Bar(i32),
    Baz(i32),
    Qux(i32),
}

fn main() {
    let foo = Foo::Bar(1);

    let (Foo::Bar(_) | Foo::Baz(_)) = foo else { return };
}
```

This is valid, producing no errors.

Instead, this happened:

```rust

enum Foo {
    Bar(i32),
    Baz(i32),
    Qux(i32),
}

fn main() {
    let foo = Foo::Bar(1);

	// Notice the missing parens, which is not valid when using `let else`
    let Foo::Bar(_) | Foo::Baz(_) = foo else { return };
}
```

Which produces the error:

```
error: top-level or-patterns are not allowed in `let` bindings
  --> src/main.rs:23:9
   |
23 |     let Foo::Bar(_) | Foo::Baz(_) = foo else { return };
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^ help: wrap the pattern in parentheses: `(Foo::Bar(_) | Foo::Baz(_))`
```

### Version

```text
rustc 1.67.0-nightly (70f8737b2 2022-11-23)
binary: rustc
commit-hash: 70f8737b2f5d3bf7d6b784fad00b663b7ff9feda
commit-date: 2022-11-23
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

alex-semenyuk:
Fix #9938
changelog: Sugg: [`manual_let_else`]: Suggestions for or-patterns now include required brackets.
[#9966](https://github.com/rust-lang/rust-clippy/pull/9966)

### DIFF
--- a/clippy_lints/src/manual_let_else.rs
+++ b/clippy_lints/src/manual_let_else.rs
@@ -151,7 +151,12 @@ fn emit_manual_let_else(cx: &LateContext<'_>, span: Span, expr: &Expr<'_>, pat: 
             } else {
                 format!("{{ {sn_else} }}")
             };
-            let sugg = format!("let {sn_pat} = {sn_expr} else {else_bl};");
+            let sn_bl = if matches!(pat.kind, PatKind::Or(..)) {
+                format!("({sn_pat})")
+            } else {
+                sn_pat.into_owned()
+            };
+            let sugg = format!("let {sn_bl} = {sn_expr} else {else_bl};");
             diag.span_suggestion(span, "consider writing", sugg, app);
         },
     );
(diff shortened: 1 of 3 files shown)


xFrednet (approved):
LGTM, we can merge this after a rebase :)

alex-semenyuk:
@xFrednet Rebased

PR #9975, opened 2022-11-28
Move `index_refutable_slice` to `pedantic`

xFrednet:
During the creation, I out this lint into the nursery group to let it run in the wild before moving it to a commonly used group. This move never happened until now, though. It should be safe, as Clippy and I have been using it for months and there are no open issues for is :)

---

changelog: Move `index_refutable_slice` to `pedantic` (Now warn-by-default)
[#9975](https://github.com/rust-lang/rust-clippy/pull/9975)


llogiq:
Thank you!

@bors r+

PR #9982, opened 2022-11-29
Don't lint `from_over_into` for opaque types

### LINKED ISSUE #9935: FP from_over_into: type_alias_impl_trait
### Summary

.

### Lint Name

from_over_into

### Reproducer

I tried this code:

```rust
#![feature(type_alias_impl_trait)]

type T = impl Sized;

struct Foo;

impl Into<T> for Foo {
    fn into(self) -> T {}
}

fn main(){
    let _: T = Foo.into();
}
```

I saw this happen:
`cargo clippy --fix -- -Aclippy::all -Wclippy::from_over_into`
```
    Checking f v0.1.0 (/tmp/f)
warning: failed to automatically apply fixes suggested by rustc to crate `f`

  * src/main.rs

error[E0119]: conflicting implementations of trait `std::convert::From<T>` for type `T`
 --> src/main.rs:7:1
  |
7 | impl From<Foo> for T {
  | ^^^^^^^^^^^^^^^^^^^^
  |
  = note: conflicting implementation in crate `core`:
          - impl<T> std::convert::From<T> for T;

For more information about this error, try `rustc --explain E0119`.

warning: an implementation of `From` is preferred since it gives you `Into<_>` for free where the reverse isn't true
 --> src/main.rs:7:1
  |
7 | impl Into<T> for Foo {
  | ^^^^^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#from_over_into
  = note: requested on the command line with `-W clippy::from-over-into`
help: replace the `Into` implentation with `From<Foo>`
  |
7 ~ impl From<Foo> for T {
8 ~     fn from(val: Foo) -> Self {}
  |

warning: `f` (bin "f") generated 1 warning (run `cargo fix --bin "f"` to apply 1 suggestion)
warning: `f` (bin "f" test) generated 1 warning (1 duplicate)
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
```

### Version

_No response_

### Additional Labels

_No response_

Jarcho:
fixes #9935

This is stalled until the next sync. The impl in question can't be written on the pinned nightly.

changelog: Don't lint `from_over_into` for opaque types

### DIFF
--- a/clippy_lints/src/from_over_into.rs
+++ b/clippy_lints/src/from_over_into.rs
@@ -10,7 +10,7 @@ use rustc_hir::{
     TyKind,
 };
 use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::hir::nested_filter::OnlyBodies;
+use rustc_middle::{hir::nested_filter::OnlyBodies, ty};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::symbol::{kw, sym};
 use rustc_span::{Span, Symbol};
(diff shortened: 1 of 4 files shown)


flip1995 (approved):
LGTM. r=me after the next sync (in case I forget)

Issue #9986, opened 2022-11-29
unnecessary_safety_doc considered harmful


asomers:
### Summary

I disagree with the `unnecessary_safety_doc` warning.  That's because a function's `Safety` section can document more than just the minimum requirements for calling the function safely.  For example:

* It can document why a function is safe, even though the programmer expects that it wouldn't be:
```rust
/// # Safety
///
/// Safe because the pointer argument is validated through some mysterious back channel
pub fn foo(*const MyStruct) {...}
```
or this real-life example
```rust
/// # Safety
/// Safe because the pointer argument is unused
// But it's required to be present because the API is defined according to some external
// interface specification
#[no_mangle]
pub extern "C" fn foo(_td: *mut thread_data) -> libc::c_int {...}
```
or this real-life example: https://github.com/asomers/blosc-rs/blob/03c25105e4d1b78af8c1af2cba98f5c390dd7c9c/blosc/src/lib.rs#L363

* It can document how to use a function "safely", even though the compiler doesn't technically consider it `unsafe`.
```rust
/// # Safety
/// Though technically safe to use, this function is vulnerable to a race condition
/// and may return wrong results unless the caller does such and such.
pub fn foo() -> i64 {...}
```
or
```rust
/// # Safety
/// The returned pointer's lifetime will be such and such, and the caller is responsible
/// for freeing it with `Box::from_raw`.
pub fn foo() -> *const MyStruct {...}
```

### Lint Name

unnecessary_safety_doc

### Reproducer

I tried this code:

```rust
/// # Safety
///
/// Safe because the pointer argument is unused.
#[no_mangle]
pub extern "C" fn fio_bfffs_commit(_td: *mut thread_data) -> libc::c_int {
    unimplemented!()
}
```

I saw this happen:

```
warning: safe function's docs have unnecessary `# Safety` section
```

I expected to see this happen:
No warning

### Version

```text
rustc 1.67.0-nightly (2585bcea0 2022-11-28)
binary: rustc
commit-hash: 2585bcea0bc2a9c42a4be2c1eba5c61137f2b167
commit-date: 2022-11-28
host: x86_64-unknown-freebsd
release: 1.67.0-nightly
LLVM version: 15.0.4
```

xFrednet:
I agree with this assessment. The lint should at least not be warn by default.

@rustbot claim

PR #9989, opened 2022-11-29
Move `unnecessary_unsafety_doc` to `pedantic`

xFrednet:
This lint was added in #9822. I like the idea, but also agree with #9986 as well. I think it should at least not be warn-by-default. This is one of these cases, where I'd like a group between pedantic and restriction. But I believe that users using `#![warn(clippy::pedantic)]` will know how to enable the lint if they disagree with it.

---

Since the lint is new:

changelog: none

r? @flip1995 since I'd suggest back porting this, the original PR was merged 16 days ago.

Closes: #9986 (While it doesn't address everything, I believe that this is the best compromise)

### DIFF
--- a/clippy_lints/src/doc.rs
+++ b/clippy_lints/src/doc.rs
@@ -253,7 +253,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.66.0"]
     pub UNNECESSARY_SAFETY_DOC,
-    style,
+    restriction,
     "`pub fn` or `pub trait` with `# Safety` docs"
 }
 
--- a/tests/ui/unnecessary_unsafety_doc.rs
+++ b/tests/ui/unnecessary_unsafety_doc.rs
@@ -1,6 +1,7 @@
 // aux-build:doc_unsafe_macros.rs
 
 #![allow(clippy::let_unit_value)]
+#![warn(clippy::unnecessary_safety_doc)]
 
 #[macro_use]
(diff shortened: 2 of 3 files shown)


flip1995 (commented):
I agree that this lint should not be warn-by-default. I wouldn't even put it in `pedantic` though. I don't see a point in suggesting to remove documentation. If people want to have `Safety` comments on only `unsafe` functions, they should enable this lint  explicitly IMO.

So I would put it in `restriction`.

---

If we merge this before tomorrow, I don't think any backport will be necessary. We had a >4 week sync drought. I will make sure that I'm right though, so I'm keeping the label until I've done that.

xFrednet:
> So I would put it in `restriction`.

I was considering that, but held off on it since it seems some wanted it in style. `pendantic` seemed to be the middle. I've now moved it to `restriction`

> If we merge this before tomorrow, I don't think any backport will be necessary. We had a >4 week sync drought. I will make sure that I'm right though, so I'm keeping the label until I've done that.

Even better :tada:

flip1995:
There was no discussion about the lint group in the PR, so I don't think there was a compelling reason for putting it in `style`.

@bors r+

Thanks!

flip1995:
Just checked: This lint is not in `beta` yet and on the current `rust-lang/rust` master it is already in `restriction`. So no need to backport anything. Thanks for pointing this out anyway!

PR #9996, opened 2022-11-30
Fix `unnecessary_cast` suggestion when taking a reference

### LINKED ISSUE #9906: unnecessary_cast causes borrowcheck problems
### Summary

.

### Lint Name

unnecessary_cast

### Reproducer

I tried this code:

```rust
// run-pass
// check that casts are not being treated as lexprs.

fn main() {
    let mut a = 0i32;
    let b = &(a as i32);
    a = 1;
    assert_ne!(&a as *const i32, b as *const i32);
    assert_eq!(*b, 0);

    assert_eq!(issue_36936(), 1);
}


struct A(u32);

impl Drop for A {
    fn drop(&mut self) {
        self.0 = 0;
    }
}

fn issue_36936() -> u32 {
    let a = &(A(1) as A);
    a.0
}
```

I saw this happen:
`cargo clippy --fix -- -Aclippy::all -Wclippy::unnecessary_cast`
```

  * src/main.rs

error[E0506]: cannot assign to `a` because it is borrowed
 --> src/main.rs:4:5
  |
3 |     let b = &a;
  |             -- borrow of `a` occurs here
4 |     a = 1;
  |     ^^^^^ assignment to borrowed `a` occurs here
5 |     assert_ne!(&a as *const i32, b as *const i32);
  |                                  - borrow later used here

For more information about this error, try `rustc --explain E0506`.

warning: casting to the same type is unnecessary (`i32` -> `i32`)
 --> src/main.rs:3:14
  |
3 |     let b = &(a as i32);
  |              ^^^^^^^^^^ help: try: `a`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
  = note: requested on the command line with `-W clippy::unnecessary-cast`
```

### Version

```text
binary: rustc
commit-date: 2022-11-19
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

Jarcho:
fixes #9906
changelog: `unnecessary_cast`: Fix suggestion when taking a reference

### DIFF
--- a/clippy_lints/src/casts/unnecessary_cast.rs
+++ b/clippy_lints/src/casts/unnecessary_cast.rs
@@ -90,7 +90,11 @@ pub(super) fn check<'tcx>(
             expr.span,
             &format!("casting to the same type is unnecessary (`{cast_from}` -> `{cast_to}`)"),
             "try",
-            cast_str,
+            if get_parent_expr(cx, expr).map_or(false, |e| matches!(e.kind, ExprKind::AddrOf(..))) {
+                format!("{{ {cast_str} }}")
+            } else {
+                cast_str
+            },
             Applicability::MachineApplicable,
         );
         return true;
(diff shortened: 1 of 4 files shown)


Alexendoo (approved):

Alexendoo:
Thanks, interesting issue & fix

@bors r+

PR #10020, opened 2022-12-01
Identify more cases of useless `into_iter()` calls

samueltardieu:
changelog: Sugg: [`useless_conversion`]: Now suggests removing calls to `into_iter()` on an expression implementing `Iterator`
[#10020](https://github.com/rust-lang/rust-clippy/pull/10020)

### DIFF
--- a/clippy_lints/src/useless_conversion.rs
+++ b/clippy_lints/src/useless_conversion.rs
@@ -1,11 +1,11 @@
 use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_sugg};
 use clippy_utils::source::{snippet, snippet_with_macro_callsite};
 use clippy_utils::sugg::Sugg;
-use clippy_utils::ty::{is_type_diagnostic_item, same_type_and_consts};
-use clippy_utils::{get_parent_expr, is_trait_method, match_def_path, paths};
+use clippy_utils::ty::{is_copy, is_type_diagnostic_item, same_type_and_consts};
+use clippy_utils::{get_parent_expr, is_trait_method, match_def_path, path_to_local, paths};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind, HirId, MatchSource};
+use rustc_hir::{BindingAnnotation, Expr, ExprKind, HirId, MatchSource, Node, PatKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty;
 use rustc_session::{declare_tool_lint, impl_lint_pass};
@@ -81,16 +81,24 @@ impl<'tcx> LateLintPass<'tcx> for UselessConversion {
                     }
                 }
                 if is_trait_method(cx, e, sym::IntoIterator) && name.ident.name == sym::into_iter {
-                    if let Some(parent_expr) = get_parent_expr(cx, e) {
-                        if let ExprKind::MethodCall(parent_name, ..) = parent_expr.kind {
-                            if parent_name.ident.name != sym::into_iter {
-                                return;
-                            }
-                        }
+                    if get_parent_expr(cx, e).is_some() &&
+                       let Some(id) = path_to_local(recv) &&
+                       let Node::Pat(pat) = cx.tcx.hir().get(id) &&
+                       let PatKind::Binding(ann, ..) = pat.kind &&
+                       ann != BindingAnnotation::MUT
+                    {
+                        // Do not remove .into_iter() applied to a non-mutable local variable used in
+                        // a larger expression context as it would differ in mutability.
+                        return;
                     }
+
                     let a = cx.typeck_results().expr_ty(e);
                     let b = cx.typeck_results().expr_ty(recv);
-                    if same_type_and_consts(a, b) {
+
+                    // If the types are identical then .into_iter() can be removed, unless the type
+                    // implements Copy, in which case .into_iter() returns a copy of the receiver and
+                    // cannot be safely omitted.
+                    if same_type_and_consts(a, b) && !is_copy(cx, b) {
                         let sugg = snippet(cx, recv.span, "<expr>").into_owned();
                         span_lint_and_sugg(
                             cx,
--- a/tests/ui/useless_conversion.rs
+++ b/tests/ui/useless_conversion.rs
@@ -33,12 +33,71 @@ fn test_issue_3913() -> Result<(), std::io::Error> {
     Ok(())
 }
 
-fn test_issue_5833() -> Result<(), ()> {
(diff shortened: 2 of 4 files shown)


clippy_lints/src/useless_conversion.rs
```diff
@@ -81,9 +81,16 @@ impl<'tcx> LateLintPass<'tcx> for UselessConversion {
                     }
                 }
                 if is_trait_method(cx, e, sym::IntoIterator) && name.ident.name == sym::into_iter {
-                    if let Some(parent_expr) = get_parent_expr(cx, e) {
-                        if let ExprKind::MethodCall(parent_name, ..) = parent_expr.kind {
-                            if parent_name.ident.name != sym::into_iter {
+                    if matches!(recv.kind, ExprKind::Path(..)) && get_parent_expr(cx, e).is_some() {
+                        if_chain! {
```
flip1995, line 85: I don't quite get what this new condition is trying to check for. There also aren't any new tests added that this new code should block the linting for. Can you explain this a bit, please?

samueltardieu: Sure. As explained in the commit message, two problems can occur by removing `.into_iter()` even when it is the identity function:

- If the receiver of `.into_iter()` is an immutable local variable `x`, removing `.into_iter()` in the context of a larger expression can change the mutability. For example, `x.into_iter().next()` works, as `x.into_iter()` won't. Note that if we are not in the context of a larger expression, we can remove the call: `let mut y = x.into_iter();` and `let mut y = x;` are equivalent, same without `mut`.
- If the receiver is a `const`, Rust can start displaying warnings although the transformation is correct. For example, if `R` is a `const` `Range<_>` (which implements `Iterator`), `R.into_iter().next()` doesn't trigger any warning while `R.next()` warns that `R` starts from scratch at every call to `.next()` (which is also the call of `R.into_iter().next()`. Here also this can only happen in the context of a larger expression, `let y = R.into_iter();` and `let y = R;` do the same thing and trigger no warning.

To avoid doing an incorrect transformation (first case) or introducing warnings from the compiler (second case), we skip attempting to removing `.into_iter()` if:

- The receiver is a `ExprKind::Path` (to cover local variables and const items)
- and the call to `.into_iter()` is used in the context of a larger expression

In this case, we want to skip the `.into_iter()` removal unless we can check (`if_chain!` below) that the receiver is a mutable local variable, because there we have no problem removing the `.into_iter()`.

One of the added tests checks the block, in addition to the existing `test_issue_5833()` which already covers the case of the receiver being a immutable local variable.

This one 

```rust
fn test_const_into_iter1() -> Result<(), ()> {
    const NUMBERS: std::ops::Range<i32> = 0..10;
    let _ = NUMBERS.into_iter().next();

    Ok(())
}
```

checks that it is blocked as well (`const` item `NUMBERS`, which is a `ExprKind::Path` and not a mutable local variable). We do not want to remove `.into_iter()` there (although it would be correct) because of the compiler warning it would provoke.

For info, I started this yesterday when I was helping a friend of mine with their Rust code for advent-of-code. They used `s.lines().into_iter().map(…)` and when I told them that `.into_iter()` was unnecessary as `.lines()` returned an `impl Iterator` they told me that Clippy hadn't warned them.

flip1995: Thanks for the great explanation! And sorry for missing the commit message.

Usually we don't care about a follow up warning by rustc (or Clippy), so you don't really have to handle that case. In the range example, I would even argue that suggesting to remove `into_iter` and then get a follow up warning is beneficial, because it points out an additional potential flaw in the code.

samueltardieu: Makes sense. I will push an amended patch. Also, I've identified another problematic corner case that I will integrate: if the receiver `x` implements `Copy`, `x.into_iter().next()` will not advance `x` while `x.next()` will (I identified this while thinking about `static` variables which could be used as receiver if they implement `Copy`). I will include new tests for this as well.

xFrednet:
Hey @samueltardieu, would you mind adding more details to the changelog entry? Some information which cases are detected now would be helpful :)

samueltardieu:
> Hey @samueltardieu, would you mind adding more details to the changelog entry? Some information which cases are detected now would be helpful :)

Done, in the style of published clippy changelog entries.

xFrednet:
Perfect, thank you very much!

xFrednet:
I'll also add this to my reviewing queue. You should have a review by the end of the week :)

r? @xFrednet

tests/ui/useless_conversion.rs
```diff
@@ -41,12 +41,90 @@ fn test_issue_5833() -> Result<(), ()> {
     Ok(())
 }
 
+fn test_issue_5833_alt1() -> Result<(), ()> {
+    let text = "foo\r\nbar\n\nbaz\n";
+    let mut lines = text.lines();
+    if Some("ok") == lines.into_iter().next() {}
+
+    Ok(())
```
xFrednet, line 49: Is there a reason, why all of these test functions return a `Result`? It seems like it would be easier and shorter to have no return at all

tests/ui/useless_conversion.rs
```diff
@@ -41,12 +41,90 @@ fn test_issue_5833() -> Result<(), ()> {
...
+    let text = "foo\r\nbar\n\nbaz\n";
+    if Some("ok") == text.lines().into_iter().next() {}
+
+    Ok(())
+}
+
+#[allow(const_item_mutation)]
+fn test_const_into_iter1() -> Result<(), ()> {
+    const NUMBERS: std::ops::Range<i32> = 0..10;
+    let _ = NUMBERS.into_iter().next();
```
xFrednet, line 70: A small tip for this test and the test above. It's usually helpful to leave a comment in the test indicating if a case should be linted or not. During development, this is often clear, but for the review I had to think about if and why this should be linted. For someone else without this PR as context, understanding this code might be harder.

The way I often choose, is to name the functions/modules accordingly. So `lint_into_iter_on_const_iter` might work or a command like:

```rs
// This should lint, since `NUMBERS` is already an iterator and copy
```

xFrednet (commented):
Two small nits in the tests. The implementation looks good to me :)

tests/ui/useless_conversion.rs
samueltardieu: Done. It was iike this so I went on doing it. I fixed all tests concerned by this patch.

tests/ui/useless_conversion.rs
samueltardieu: Renamed all tests concerned by this change.

xFrednet (approved):

xFrednet:
Everything looks good to me, thank you for the contribution and your patience!

@bors r+

Issue #10021, opened 2022-12-01
ICE `type parameter `Self/#0` (Self/0) out of range when substituting, substs=[]` on November 30th nightly


Tpt:
### Summary

First, thank you so much for the Clippy team for your amazing work.

Sadly, it seems #9505 has not fixed the error in all cases. When running clippy using November 30th nightly I still get the error on [this repository](https://github.com/oxigraph/rio).

To reproduce:
```sh
git clone https://github.com/oxigraph/rio
cd rio
cargo +nightly clippy
```

This is maybe a duplicate of #9675

Feel free to reach out if you need help.

### Version

```text
rustc 1.67.0-nightly (c97b539e4 2022-11-30)
binary: rustc
commit-hash: c97b539e408ea353f4fde2f9251d598291fec421
commit-date: 2022-11-30
host: x86_64-unknown-linux-gnu
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Error output

<details><summary>Backtrace</summary>
  <p>

  ```
error: internal compiler error: compiler/rustc_middle/src/ty/subst.rs:810:9: type parameter `Self/#0` (Self/0) out of range when substituting, substs=[]

thread 'rustc' panicked at 'Box<dyn Any>', /rustc/c97b539e408ea353f4fde2f9251d598291fec421/compiler/rustc_errors/src/lib.rs:1575:9
stack backtrace:
   0: std::panicking::begin_panic::<rustc_errors::ExplicitBug>
   1: std::panic::panic_any::<rustc_errors::ExplicitBug>
   2: <rustc_errors::HandlerInner>::bug::<&alloc::string::String>
   3: <rustc_errors::Handler>::bug::<&alloc::string::String>
   4: rustc_middle::ty::context::tls::with_context_opt::<rustc_middle::ty::context::tls::with_opt<rustc_middle::util::bug::opt_span_bug_fmt<rustc_span::span_encoding::Span>::{closure#0}, ()>::{closure#0}, ()>
   5: rustc_middle::util::bug::opt_span_bug_fmt::<rustc_span::span_encoding::Span>
   6: rustc_middle::util::bug::bug_fmt
   7: <rustc_middle::ty::subst::SubstFolder>::type_param_out_of_range
   8: <rustc_middle::ty::subst::SubstFolder as rustc_middle::ty::fold::TypeFolder>::fold_ty
[...]

PR #9505, opened 2022-09-21
Fix ICE in `unnecessary_to_owned`

mikerite:
Fixes #9504

Compiler generated call `into_future` nodes return empty substs which we need when checking it's predicates. Handle this by simply exitting when we encounter one. This change introduces false negatives in place of the ICEs.

changelog: [`unnecessary_to_owned`]: fix ICE

PR #10027, opened 2022-12-03
Fix 10021

smoelius:
This PR proposes a fix for #10021.

The problem is similar to the one that @mikerite described in #9505. The compiler is generating an empty substitution for a call, even though the type of `Self` seems to be needed. In @mikerite's case, the call was to [`IntoFuture::into_future`](https://doc.rust-lang.org/std/future/trait.IntoFuture.html#tymethod.into_future). In this case, the call is to [`Try::branch`](https://doc.rust-lang.org/std/ops/trait.Try.html#tymethod.branch).

The proposed fix is to verify that the parameter whose type is changing has an index within the substitution. The strikes me as a reasonable approach, since if the check were to fail, the following code would be a no-op:

https://github.com/rust-lang/rust-clippy/blob/4c123a06ba3c2ec17d3a4dfa251dccdc5368b712/clippy_lints/src/methods/unnecessary_to_owned.rs#L420-L428

Like @mikerite's original solution, this solution turns ICEs into false negatives.

changelog: fix `unnecessary_to_owned` false positive involving `Try::branch`


dswij (approved):
Thanks for this!

PR #10028, opened 2022-12-03
Add `extra_unused_type_parameters` lint

### LINKED ISSUE #9240: Lint for unused type parameters on functions
### What it does

Similar to `extra_unused_lifetimes`, checks for unused type parameters on functions and warns that this is likely unintended, suggesting that they be removed. (Note: should this only apply to functions, or are there other places it could be run?)

### Lint Name

extra_unused_type_parameters

### Category

complexity

### Advantage

- Calling functions with unused type parameters requires adding a turbofish, since inferring them is impossible. The turbofish itself is useless other than to satisfy the compiler.
- If refactoring a function with a long signature and removing the use of generics, it's easy to miss removing the corresponding type parameter, especially if the function is `pub` and isn't used elsewhere. This accidentally exposes downstream users to the author's mistake, which is bad news.

### Drawbacks

I can't think of any besides maybe the lint name, which might need some bikeshedding. Also, maybe a `extra_unused_generics` lint could be introduced that implies both `extra_unused_lifetimes` and `extra_unused_type_params`?

### Example

```rust
pub fn foo<T>(x: u8) {
    // T unused in function body as well
    // ..
}
```

Could be written as:

```rust
pub fn foo(x: u8) {
    // ..
}
```

mkrasnitski:
Closes #9240. ~~Keeping this as draft for now, because of a bug that I don't know how to fix.~~ It seems that opaque return types are not walked properly, for some unknown reason. As in, the following:

```rust
fn used_ret_opaque<A>() -> impl Iterator<Item = A> {
    std::iter::empty()
}
```
This triggers the lint even though it shouldn't. Discussion on Zulip didn't illuminate any possible reasons why, so PR-ing this now to increase visibility.

---

changelog: new lint: [`extra_unused_type_parameters`]
[#10028](https://github.com/rust-lang/rust-clippy/pull/10028)

### DIFF
--- a/clippy_lints/src/extra_unused_type_parameters.rs
+++ b/clippy_lints/src/extra_unused_type_parameters.rs
@@ -1,0 +1,178 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::trait_ref_of_method;
+use rustc_data_structures::fx::FxHashMap;
+use rustc_errors::MultiSpan;
+use rustc_hir::intravisit::{walk_impl_item, walk_item, walk_param_bound, walk_ty, Visitor};
+use rustc_hir::{
+    GenericParamKind, Generics, ImplItem, ImplItemKind, Item, ItemKind, PredicateOrigin, Ty, TyKind, WherePredicate,
+};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::hir::nested_filter;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::{def_id::DefId, Span};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for type parameters in generics that are never used anywhere else.
+    ///
+    /// ### Why is this bad?
+    /// Functions cannot infer the value of unused type parameters; therefore, calling them
+    /// requires using a turbofish, which serves no purpose but to satisfy the compiler.
+    ///
+    /// ### Example
+    /// ```rust
+    /// // unused type parameters
+    /// fn unused_ty<T>(x: u8) {
+    ///     // ..
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn no_unused_ty(x: u8) {
+    ///     // ..
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub EXTRA_UNUSED_TYPE_PARAMETERS,
+    complexity,
+    "unused type parameters in function definitions"
+}
+declare_lint_pass!(ExtraUnusedTypeParameters => [EXTRA_UNUSED_TYPE_PARAMETERS]);
+
+/// A visitor struct that walks a given function and gathers generic type parameters, plus any
+/// trait bounds those parameters have.
+struct TypeWalker<'cx, 'tcx> {
+    cx: &'cx LateContext<'tcx>,
+    /// Collection of all the type parameters and their spans.
+    ty_params: FxHashMap<DefId, Span>,
+    /// Collection of any (inline) trait bounds corresponding to each type parameter.
+    bounds: FxHashMap<DefId, Span>,
+    /// The entire `Generics` object of the function, useful for querying purposes.
+    generics: &'tcx Generics<'tcx>,
+    /// The value of this will remain `true` if *every* parameter:
+    ///   1. Is a type parameter, and
+    ///   2. Goes unused in the function.
+    /// Otherwise, if any type parameters end up being used, or if any lifetime or const-generic
+    /// parameters are present, this will be set to `false`.
+    all_params_unused: bool,
+}
+
+impl<'cx, 'tcx> TypeWalker<'cx, 'tcx> {
+    fn new(cx: &'cx LateContext<'tcx>, generics: &'tcx Generics<'tcx>) -> Self {
+        let mut all_params_unused = true;
+        let ty_params = generics
+            .params
+            .iter()
+            .filter_map(|param| {
+                if let GenericParamKind::Type { .. } = param.kind {
+                    Some((param.def_id.into(), param.span))
+                } else {
+                    if !param.is_elided_lifetime() {
+                        all_params_unused = false;
+                    }
+                    None
+                }
+            })
+            .collect();
+        Self {
+            cx,
+            ty_params,
+            bounds: FxHashMap::default(),
+            generics,
+            all_params_unused,
+        }
+    }
+
+    fn emit_lint(&self) {
+        let (msg, help) = match self.ty_params.len() {
+            0 => return,
+            1 => (
+                "type parameter goes unused in function definition",
+                "consider removing the parameter",
+            ),
+            _ => (
+                "type parameters go unused in function definition",
+                "consider removing the parameters",
+            ),
+        };
+
+        let source_map = self.cx.tcx.sess.source_map();
+        let span = if self.all_params_unused {
+            self.generics.span.into() // Remove the entire list of generics
+        } else {
+            MultiSpan::from_spans(
+                self.ty_params
+                    .iter()
+                    .map(|(def_id, &span)| {
+                        // Extend the span past any trait bounds, and include the comma at the end.
+                        let span_to_extend = self.bounds.get(def_id).copied().map_or(span, Span::shrink_to_hi);
+                        let comma_range = source_map.span_extend_to_next_char(span_to_extend, '>', false);
+                        let comma_span = source_map.span_through_char(comma_range, ',');
+                        span.with_hi(comma_span.hi())
+                    })
+                    .collect(),
+            )
+        };
+
+        span_lint_and_help(self.cx, EXTRA_UNUSED_TYPE_PARAMETERS, span, msg, None, help);
+    }
+}
+
+impl<'cx, 'tcx> Visitor<'tcx> for TypeWalker<'cx, 'tcx> {
+    type NestedFilter = nested_filter::OnlyBodies;
+
+    fn visit_ty(&mut self, t: &'tcx Ty<'tcx>) {
+        if let Some((def_id, _)) = t.peel_refs().as_generic_param() {
+            if self.ty_params.remove(&def_id).is_some() {
+                self.all_params_unused = false;
+            }
+        } else if let TyKind::OpaqueDef(id, _, _) = t.kind {
+            // Explicitly walk OpaqueDef. Normally `walk_ty` would do the job, but it calls
+            // `visit_nested_item`, which checks that `Self::NestedFilter::INTER` is set. We're
+            // using `OnlyBodies`, so the check ends up failing and the type isn't fully walked.
+            let item = self.nested_visit_map().item(id);
+            walk_item(self, item);
+        } else {
+            walk_ty(self, t);
+        }
+    }
+
+    fn visit_where_predicate(&mut self, predicate: &'tcx WherePredicate<'tcx>) {
+        if let WherePredicate::BoundPredicate(predicate) = predicate {
+            // Collect spans for bounds that appear in the list of generics (not in a where-clause)
+            // for use in forming the help message
+            if let Some((def_id, _)) = predicate.bounded_ty.peel_refs().as_generic_param()
+                && let PredicateOrigin::GenericParam = predicate.origin
+            {
+                self.bounds.insert(def_id, predicate.span);
+            }
+            // Only walk the right-hand side of where-bounds
+            for bound in predicate.bounds {
+                walk_param_bound(self, bound);
+            }
+        }
+    }
+
+    fn nested_visit_map(&mut self) -> Self::Map {
+        self.cx.tcx.hir()
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for ExtraUnusedTypeParameters {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
+        if let ItemKind::Fn(_, generics, _) = item.kind {
+            let mut walker = TypeWalker::new(cx, generics);
+            walk_item(&mut walker, item);
+            walker.emit_lint();
+        }
+    }
+
+    fn check_impl_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx ImplItem<'tcx>) {
+        // Only lint on inherent methods, not trait methods.
+        if let ImplItemKind::Fn(..) = item.kind && trait_ref_of_method(cx, item.owner_id.def_id).is_none() {
+            let mut walker = TypeWalker::new(cx, item.generics);
+            walk_impl_item(&mut walker, item);
+            walker.emit_lint();
+        }
+    }
+}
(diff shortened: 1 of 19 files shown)


mkrasnitski:
After some digging, it seems the corresponding hir type was added all the way back in rust-lang/rust#54741 (since renamed to `TyKind::OpaqueDef`), and has only ever contained lifetime bounds on the `impl Trait`, not any parameter bounds. Does the `Item = A` count as a parameter bound in this case?

cc-ing @oli-obk

oli-obk:
When you seen an `OpaqueDef` you basically need to fetch the HIR of the item it references and walk that, too. We're mangling the HIR quite a bit here to make the queries like `generics_of` and `predicates_of` simpler. This is in the process of being changed and I don't know what the end result will be. Maybe for now just mark all generic params as used if there's an `OpaqueDef`?

mkrasnitski:
> When you seen an `OpaqueDef` you basically need to fetch the HIR of the item it references and walk that, too.

Turns out `walk_ty` already does this by calling `visit_nested_item` on the id. However, that involves a check that `Self::NestedFilter::INTER` is set, and I'm using `OnlyBodies` which keeps it `false`. That's the reason the walk was stopping short of descending down into the `OpaqueDef`. Maybe that behavior should be changed, since the filter check seems not to have been intended for this edge-case?

mkrasnitski:
This PR is ready for review/merge, I think.

flip1995:
Assigning myself, because we talked about this on Zulip a lot. I will probably be busy with release work in the next week, so review might take a bit.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,118 @@
...
+                    GenericParamKind::Type { .. } => Some((param.def_id.into(), param.span)),
+                    _ => None,
+                })
+                .collect(),
+        }
+    }
+
+    fn emit_lint(&self) {
+        for span in self.map.values() {
+            span_lint(
```
flip1995, line 61: This should be a `span_lint_and_sugg`. However, I can see that producing a suggestion that suggests to remove the parameter might be hard. (how does `extra_unused_lifetimes` do this?). But this should at least be a `span_lint_and_help`.

flip1995 (commented):
Here's an early review. I still have to look at the tests and think a little bit about all the cases I can come up with. Impl looks already really clean AFAICT from a quick glance.

clippy_lints/src/extra_unused_type_parameters.rs
mkrasnitski: Actually, `extra_unused_lifetimes` only uses `span_lint`. I think part of the challenge is in determining the span for the help/suggestion, as it's possible that _all_ the generic parameters on a function go unused, in which case the fix requires also removing the surrounding angle brackets. Or, if there are still some parameters left over, the angle brackets remain. I could certainly add a help string and keep the help span `None`.

Also, what's your opinion on combining multiple unused parameters into a single lint message? I have a test case for this: `unused_multi<T, U>(x: u8) {}`

flip1995: Yes, figuring out spans here will be the hardest part. Also figuring out what `,` to remove together with the generic parameter will be hard.

Also this is Rust, so `fn foo<>() {}` is [totally valid syntax](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b320f376223a5c58a50e2e5203c79321) which will then be removed by `rustfmt`. So you don't have to worry about the brackets. 

So when figuring out the spans, you will only have to care about the bounds and the `,`s. 

But this is all extra credit work. Producing cool suggestions can be fun though :) 

---

> Also, what's your opinion on combining multiple unused parameters into a single lint message?

Also a cool thing to do. You will then have to suggest things with `multipart_suggestion`, which is not `rustfix`able, though.

If you're motivated to try out things like this, go ahead and do it. You can post the results to Zulip and we can figure out what looks and feels best. However, I would already be fine with merging this with just a `span_lint_and_help`.

flip1995:
```rust
pub fn foo<T,>() {}
```

is also valid. So you probably want a span until the next `,` or `>`, if no `,` exists.

Take a look at `SourceMap` and its functions to produce such spans: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/source_map/struct.SourceMap.html#method.span_until_char

mkrasnitski: I've pushed my attempt at doing this; let me know what you think.

mkrasnitski:
I've found an edge case where the lint doesn't behave quite as expected:

```rust
fn nested_bounds<A, B, C, D>()
where
    B: Iterator<Item = A>,
    C: Iterator<Item = B>,
    D: Iterator<Item = C>,
{
}
```

This produces the following lint output:
```
warning: type parameter goes unused in function definition
  --> tests/ui/extra_unused_type_parameters.rs:61:27
   |
61 | fn nested_bounds<A, B, C, D>()
   |                           ^
   |
   = help: consider removing the parameter
```

This is because I mark parameters used if they appear in the right-hand side of a where-clause. What I failed to consider, was if the left side _also_ goes unused, then I shouldn't mark the nested parameter as unused. What is the consensus here, should this be fixed? Subsequent applications of the lint will eventually get rid of all of the parameters, but it would be nice if they were all detected at once. However, I feel like that would require a fair bit of engineering effort.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,177 @@
...
+            .filter_map(|param| match &param.kind {
+                GenericParamKind::Type { .. } => Some((param.def_id.into(), param.span)),
+                non_ty => {
+                    if let GenericParamKind::Lifetime { kind } = non_ty {
+                        match kind {
+                            LifetimeParamKind::Elided => (),
+                            _ => has_non_ty_params = true,
+                        }
+                    }
+                    None
```
mkrasnitski, line 66: Just realized this check is buggy; it doesn't account for constant parameters. Currently away from my PC, but I'll fix it tonight.

mkrasnitski:
This is ready for re-review, when you have the time @flip1995

flip1995:
Sorry for taking so long. Last week before vacation in $day_job is a bit stressful. I'll get back to this during the holidays.

mkrasnitski:
Will this still be in time to get merged into 1.67?

flip1995:
Sadly not. It would've to get merged today to make it in 1.68. I'm so sorry that I didn't get to review this in time.

mkrasnitski:
That's alright, no big deal. So, you're saying that since 1.68 goes into beta tomorrow, this would realistically only make it into 1.69? Should I go ahead and update the version number on the lint then?

flip1995:
> So, you're saying that since 1.68 goes into beta tomorrow, this would realistically only make it into 1.69?

Beta is branched the week before the release. So next Friday (1 week from now). This will be the 1.68 beta. The code that is in this repo is **not** the code that is released. The released code is in the rust-lang/rust repo. We sync this repo to the rust repo every other week. This sync was yesterday and there won't be another sync until the release

More about the Rust release process: https://forge.rust-lang.org/release/process.html 

> Should I go ahead and update the version number on the lint then?

Sure. But it doesn't really matter. Getting the version numbers right is part of writing the changelog. But the more version numbers are correct, the less work for the changelog writer.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,175 @@
...
+            walk_ty(self, t);
+        }
+    }
+
+    fn visit_where_predicate(&mut self, predicate: &'tcx WherePredicate<'tcx>) {
+        if let WherePredicate::BoundPredicate(predicate) = predicate {
+            // Collect spans for bounds that appear in the list of generics (not in a where-clause)
+            // for use in forming the help message
+            if let Some((def_id, _)) = predicate.bounded_ty.peel_refs().as_generic_param()
+                && predicate.span < self.generics.where_clause_span
```
flip1995, line 143: Huh, what does a comparison of spans do? :thinking:

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,175 @@
...
+                        }
+                        span
+                    })
+                    .collect(),
+            )
+        } else {
+            self.generics.span.into()
+        };
+
+        span_lint_and_help(self.cx, EXTRA_UNUSED_TYPE_PARAMETERS, span, msg, None, help);
```
flip1995, line 115: You've put so much effort in crafting this awesome spans and message, I think this should be `span_lint_and_sugg`. The non-multispan cases can then be tested with `// run-rustfix` in the test file. You'll have to split up the test files for that though: fixable and unfixable.

flip1995 (commented):
I finally managed to get to this. Looks really good overall. I have some style questions left.

You're concern about those cascading bounds is not an issue we should address. Just applying this lint multiple times is totally fine and pretty much how Clippy works. It's not worth putting in the effort to try and fix this and then potentially introducing FPs or FNs.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,175 @@
...
+                    .map(|(def_id, &(mut span))| {
+                        if let Some(bound_span) = self.bound_map.get(def_id) {
+                            span = span.with_hi(bound_span.hi());
+                        }
+                        span = source_map.span_extend_to_next_char(span, ',', false);
+                        span = span.with_hi(BytePos(span.hi().0 + 1));
+
+                        let max_hi = self.generics.span.hi();
+                        if span.hi() >= max_hi {
+                            span = span.with_hi(BytePos(max_hi.0 - 1));
```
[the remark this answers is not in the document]
mkrasnitski: The first use there is to include the comma after the parameter when highlighting the span. Since `SourceMap::span_extend_to_next_char` stops at the character immediately prior, advancing one character farther should always work, I think.

The second use occurs if the comma scanning blows past the end of the list of parameters (and into the list of function arguments), and rewinds back to the character just before the last ">".

clippy_lints/src/extra_unused_type_parameters.rs
mkrasnitski: I assumed it's the same as doing a comparison on the tuple `(lo, hi)`, but I'm guessing comparing `.lo()` should work just as well.

flip1995: I worded my question badly: What is this span comparison trying to achieve?

mkrasnitski: Here I'm looking for any bounds that appear inline with the list of generics, instead of in a where-clause. Comparing spans was the easiest way I thought of to rule out any bounds in the where-clause.

clippy_lints/src/extra_unused_type_parameters.rs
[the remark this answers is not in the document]
mkrasnitski: Makes sense, although the presence of any bounds that are themselves generic would complicate this, since they contain `>`, and might also contain commas.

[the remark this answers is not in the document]
flip1995: I see. Yeah, then I think your solution is the best. But please add comments why those `BytePos`es are "safe".

clippy_lints/src/extra_unused_type_parameters.rs
flip1995: A `WhereBoundPredicate` has a [`PredicateOrigin`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.PredicateOrigin.html). Maybe you can use that instead?

mkrasnitski: You're totally right.

clippy_lints/src/extra_unused_type_parameters.rs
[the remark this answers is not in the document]
mkrasnitski: I think it's still possible to get rid of them actually. I'll just have to skip past the bound if it's present, then apply the logic you posted.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// }
+    /// ```
+    #[clippy::version = "1.68.0"]
+    pub EXTRA_UNUSED_TYPE_PARAMETERS,
+    complexity,
+    "unused type parameters in function definitions"
+}
+declare_lint_pass!(ExtraUnusedTypeParameters => [EXTRA_UNUSED_TYPE_PARAMETERS]);
+
+struct TypeWalker<'cx, 'tcx> {
```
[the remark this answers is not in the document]
mkrasnitski: Done.

mkrasnitski:
I've pushed some changes that hopefully address your concerns. It may be a few days before I can get to changing to using `span_lint_and_sugg`, as that would require changes in how I'd form the suggestion, and overall it will take a bit of effort.

Also, I'd like to at some point just squash all these commits before the PR is merged, as I don't think they're super informative on their own.

Let me know what you think of the new changes.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -0,0 +1,177 @@
...
+        } else {
+            MultiSpan::from_spans(
+                self.ty_params
+                    .iter()
+                    .map(|(def_id, &span)| {
+                        // Extend the span past any trait bounds, and include the comma at the end.
+                        let span_to_extend = self.bounds.get(def_id).copied().map_or(span, Span::shrink_to_hi);
+                        let comma_range = source_map.span_extend_to_next_char(span_to_extend, '>', false);
+                        let comma_span = source_map.span_through_char(comma_range, ',');
+                        span.with_hi(comma_span.hi())
```
flip1995, line 111: Does this work for

```rust
fn foo<A: Bound<u32, u32>, B>(_: B) {}
```

I expect that this will produce the span:
```rust
fn foo<A: Bound<u32, u32>, B>(_: B) {}
//     ^^^^^^^^^^^^^
```

flip1995 (commented):
Only one comment left.

flip1995:
> It may be a few days before I can get to changing to using `span_lint_and_sugg`, as that would require changes in how I'd form the suggestion, and overall it will take a bit of effort.

We can also just do this in a follow up PR. This is not blocking for this PR.

> Also, I'd like to at some point just squash all these commits before the PR is merged

Sure, I'll wait for that after approving this PR :+1:

clippy_lints/src/extra_unused_type_parameters.rs
mkrasnitski: I'm able to confirm that the produced span is in fact the following:

```rust
fn foo<A: Bound<u32, u32>, B>(_: B) {}
//     ^^^^^^^^^^^^^^^^^^^
```

mkrasnitski:
I've gone ahead and squashed everything down. Should be ready for merge now.

flip1995 (approved):
Awesome work! Easy to understand code, really concise and well commented+documented. And great suggestion building (one of my favorite parts when working on diagnostics).

mkrasnitski:
@flip1995 Thanks for all your help on this work, I really appreciate it!

flip1995:
Thank you for sticking with it and patiently waiting for my review!

mkrasnitski:
I now realize it's been pretty much 6 months since I opened the tracking issue for this - I went in with basically zero knowledge about compiler internals, but now I feel a lot better equipped to go off and tackle more work :)

flip1995:
Sounds great! I would be happy to see more contributions from you to Clippy! But if you want to tackle bigger challenges also have a look at rustc :)

PR #10038, opened 2022-12-06
`cast_possible_truncation` Suggest TryFrom when truncation possible

xFrednet:
This fixes the last issues from https://github.com/rust-lang/rust-clippy/pull/9664 as the author seems to be inactive. The PR author was sadly kept during the rebase, due to the conflict resolution.

IDK if it's worth it do to a full review, I only added the last commit, everything else remained the same, besides a rebase.

---

changelog: Sugg: [`cast_possible_truncation`]: Now suggests using `try_from` or allowing the lint
[#10038](https://github.com/rust-lang/rust-clippy/pull/10038)

closes: https://github.com/rust-lang/rust-clippy/issues/9231

### DIFF
--- a/clippy_lints/src/casts/cast_possible_truncation.rs
+++ b/clippy_lints/src/casts/cast_possible_truncation.rs
@@ -3,8 +3,6 @@ use clippy_utils::diagnostics::{span_lint, span_lint_and_then};
 use clippy_utils::expr_or_init;
 use clippy_utils::source::snippet;
 use clippy_utils::ty::{get_discriminant_value, is_isize_or_usize};
-use rustc_ast::ast;
-use rustc_attr::IntType;
 use rustc_errors::{Applicability, SuggestionStyle};
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::{BinOpKind, Expr, ExprKind};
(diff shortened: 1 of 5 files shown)


clippy_lints/src/casts/cast_possible_truncation.rs
```diff
@@ -153,5 +155,19 @@ pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, cast_expr: &Expr<'_>,
         _ => return,
     };
 
-    span_lint(cx, CAST_POSSIBLE_TRUNCATION, expr.span, &msg);
+    let snippet = snippet(cx, expr.span, "..");
+    let name_of_cast_from = snippet.split(" as").next().unwrap_or("..");
```
Jarcho, line 159: This doesn't work with chained casts. e.g. `x as u32 as u16`.

clippy_lints/src/casts/cast_possible_truncation.rs
```diff
@@ -153,5 +155,19 @@ pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, cast_expr: &Expr<'_>,
...
+    let suggestion = format!("{cast_to}::try_from({name_of_cast_from})");
```
Jarcho, line 160: Using `cast_to` in the suggestion will lose any type aliases. It would be better to take a snippet.

Jarcho (commented):
Just a couple of edge cases noted. Otherwise LGTM.

xFrednet:
Good points! I've rebased, since it has been over a month. All changes are in the last commit, as well as a test for the `as u32 as u16` expression.

Thank you for your patience, when it came to the update :)

Jarcho:
Thank you. @bors r+

RalfJung:
I am a bit surprised that this was marked as closing https://github.com/rust-lang/rust-clippy/issues/9231; at least I don't see a test for usize-to-u64 casts here.

PR #10043, opened 2022-12-06
Add lint `almost_complete_digit_range`

phtown:
changelog: [`almost_complete_digit_range`]: Add digit analog to `almost_complete_letter_range`

I have added a lint that will detect `'0'..'9'` and suggest converting it to `'0'..='9'`, the same way `almost_complete_letter_range` does for ascii letters. I tied into the implementation of `AlmostCompleteLetterRange` in order to do that, but I didn't change its interface.
This is my first contribution to Clippy, so please let me know if there's anything I should do differently. I'll be happy to incorporate any suggestions you have.
Thanks!


Jarcho:
Just a comment a naming. It would be better to rename `almost_complete_letter_range` to `almost_complete_range`, and then use the same lint to implement both. Changes look good otherwise.

Might be easier to start from `master`, use `cargo dev rename_lint almost_complete_letter_range almost_complete_range` and then add your changes onto that.

phtown:
While making the change you suggested I ran into an issue I hope you can shed some light on. If you run the test `tests/ui/almost_complete_range` and look at the macro `b`, you'll see the lint only triggers on the first instance it should. Is that expected behavior for `macro_rules` macros? I can't see anything in the lint definition that would make it do that.

Jarcho:
I'm surprised the compiler isn't complaining about that. Only the first arm of that macro will ever be expanded.

You probably meant to do something like
```rust
macro_rules! b {
    () => {{
        let _ = 'a'..'z';
        let _ = 'A'..'Z';
        let _ = '0'..'9';
    }};
}
```

Jarcho:
Looks good to go. You can squash the changes. Probably best to have the rename as a separate commit, but not necessary if that's too much effort.

phtown:
It looks like something I did while rebasing closed this merge request. Hopefully I just need to reopen it?

Jarcho:
Weird. I wonder what you did.

Thank you. @bors r+

Issue #9933, opened 2022-11-23
is digits instead of digits range contains


leonardo-m:
### What it does

It suggests to replace a coding pattern of using range + "contains" with a more specific function.

### Lint Name

is_ascii_digit_instead_contains

### Category

style

### Advantage

Less bug-prone, shorter

### Example

```
fn foo1(d: u8) -> bool {
    (b'0' ..= b'9').contains(&d)
}
```

Could be written as:

```
fn foo2(d: u8) -> bool {
    d.is_ascii_digit()
}
```

koka831:
It is supported in https://rust-lang.github.io/rust-clippy/master/#manual_is_ascii_check :)

leonardo-m:
Are you sure that lint fires for that foo1 function? I've tried with this and still I don't see anything.

    #![warn(clippy::all)]
    #![warn(clippy::nursery)]
    #![warn(clippy::pedantic)]

koka831:
Ah, I'm sorry I misunderstood it. The lint only supports for `matches!` macro and does not for `contains` currently.

xFrednet:
This can be an addition to the lint logic. :)

@rustbot label +good-first-issue

naosense:
@rustbot claim 

Any hints where I can start?😀

koka831:
Hi @naosense, you can add check for `contains` method in `is_manual_ascii_check` lint.
Currently it looks for `matches!` macro only.
https://github.com/rust-lang/rust-clippy/blob/1a9657139d1beca1d01486ba360fa17c43b9f58f/clippy_lints/src/manual_is_ascii_check.rs#L78-L80

You may find further informations in the book https://doc.rust-lang.org/nightly/clippy/development/index.html :)

xFrednet:
You can start, by checking for `<expr-1>.contains(<expr-2>)` calls. If `<expr-2>` is a character you can then check if `<expr-1>` is a specific range. When starting on a new lint I often first look at the HIR with the `#[clippy::dump]` attribute. Try running Clippy on this playground: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=cf532662d9bd2c097e6194e808ae3fe4

[...]

PR #10053, opened 2022-12-09
improve `manual_is_ascii_check ` check

naosense:
Sorry, not familiar the api, i can only check the method name of expression `<expr-1>.contains(<expr-2>)` after read clippy book and hints from #9933 . i dont know how to check
1.  if <expr-1> is a specific range
2. <expr-2> is a character

r? @xFrednet could you please provide some more hints? 😝️

---

changelog: Enhancement: [`manual_is_ascii_check`]: Now detects ranges with `.contains()` calls
[#10053](https://github.com/rust-lang/rust-clippy/pull/10053)

### DIFF
--- a/clippy_lints/src/manual_is_ascii_check.rs
+++ b/clippy_lints/src/manual_is_ascii_check.rs
@@ -3,7 +3,7 @@ use clippy_utils::{diagnostics::span_lint_and_sugg, higher, in_constant, macros:
 use rustc_ast::ast::RangeLimits;
 use rustc_ast::LitKind::{Byte, Char};
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind, PatKind, RangeEnd};
+use rustc_hir::{BorrowKind, Expr, ExprKind, PatKind, RangeEnd};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::{def_id::DefId, sym, Span};
@@ -86,8 +86,12 @@ impl<'tcx> LateLintPass<'tcx> for ManualIsAsciiCheck {
             && path.ident.name == sym!(contains)
             && let Some(higher::Range { start: Some(start), end: Some(end), limits: RangeLimits::Closed })
             = higher::Range::hir(receiver) {
-                let range = check_range(start, end);
+            let range = check_range(start, end);
+            if let ExprKind::AddrOf(BorrowKind::Ref, _, e) = arg.kind {
+                check_is_ascii(cx, expr.span, e, &range);
+            } else {
                 check_is_ascii(cx, expr.span, arg, &range);
+            }
         }
     }
 
--- a/clippy_lints/src/manual_is_ascii_check.rs
+++ b/clippy_lints/src/manual_is_ascii_check.rs
@@ -1,11 +1,11 @@
 use clippy_utils::msrvs::{self, Msrv};
-use clippy_utils::{diagnostics::span_lint_and_sugg, in_constant, macros::root_macro_call, source::snippet};
+use clippy_utils::{diagnostics::span_lint_and_sugg, higher, in_constant, macros::root_macro_call, source::snippet};
 use rustc_ast::LitKind::{Byte, Char};
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, PatKind, RangeEnd};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
-use rustc_span::{def_id::DefId, sym};
+use rustc_span::{def_id::DefId, sym, Span};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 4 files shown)


llogiq:
1. We have a `clippy_utils::higher::Range::hir(&Expr)` call that will give you a `higher::Range { start, end, limits }` you can use to match both exclusive and inclusive ranges.
2. you'll be looking at a `ExprKind::Lit` with a node of `LitKind::Char(c)` containing the character.

naosense:
@llogiq Can you take a look at whether I'm in the right direction?

clippy_lints/src/manual_is_ascii_check.rs
```diff
@@ -75,47 +75,54 @@ impl<'tcx> LateLintPass<'tcx> for ManualIsAsciiCheck {
...
-                        "try",
-                        format!("{recv}.{sugg}()"),
-                        applicability,
-                    );
-                }
+                check_is_ascii(cx, macro_call.span, recv, &range);
             }
+        } else if let ExprKind::MethodCall(path, receiver, [arg], ..) = expr.kind
+            && path.ident.name == sym!(contains)
+            && let Some(higher::Range { start: Some(start), end: Some(end), .. }) = higher::Range::hir(receiver) {
```
xFrednet, line 86: This is ignoring the `limits` field, which should be considered while matching the range. Otherwise, `0..9` and `0..=9` are seen as the same :)

tests/ui/manual_is_ascii_check.rs
```diff
@@ -15,6 +15,7 @@ fn main() {
     assert!(matches!('x', 'A'..='Z' | 'a'..='z'));
 
     assert!(matches!('x', 'A'..='Z' | 'a'..='z' | '_'));
+    assert!((b'0'..=b'9').contains(&b'0'));
```
xFrednet, line 18: Could you add more tests with non-exclusive ranges like this: `(b'0'..b'9').contains(&b'0')`.

It would also be good to test all the other ranges:
`'A'..='Z'`, `'a'..='z'` ...

xFrednet (commented):
This looks like a good start. The main thing missing are tests. I've added comments at the related locations :)

tests/ui/manual_is_ascii_check.fixed
```diff
@@ -15,6 +15,14 @@ fn main() {
     assert!('x'.is_ascii_alphabetic());
 
     assert!(matches!('x', 'A'..='Z' | 'a'..='z' | '_'));
+
+    assert!(&b'0'.is_ascii_digit());
```
naosense, line 19: I dont think borrowing format is right, it should be `b'0'.is_ascii_digit()`, `&b'a'.is_ascii_lowercase()` etc, but i dont know how to do that.

clippy_lints/src/manual_is_ascii_check.rs
naosense: Do i just need to verify `RangeLimits::Closed`?

tests/ui/manual_is_ascii_check.fixed
xFrednet: `.fixed` files are checked that they compile. If they don't the test would fail, so this wouldn't be a blocker. The ideal solution would be, to check if the expression is a reference [`ExprKind::AddrOf`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/enum.ExprKind.html#variant.AddrOf), if so take the inner expression of `AddrOf` otherwise just snip it as it is :)

clippy_lints/src/manual_is_ascii_check.rs
xFrednet: Yes, people could also write `0..':'` as `:` is right after the 9, but that would be super weird. It's better not to lint them and only check for `0..9` with `RangeLimits::Closed`

naosense: thanks, resolved

clippy_lints/src/manual_is_ascii_check.rs
```diff
@@ -23,6 +24,14 @@ declare_clippy_lint! {
     ///     assert!(matches!(b'X', b'A'..=b'Z'));
     ///     assert!(matches!('2', '0'..='9'));
     ///     assert!(matches!('x', 'A'..='Z' | 'a'..='z'));
+    ///
+    ///     assert!((b'0'..=b'9').contains(&b'0'));
+    ///     assert!((b'a'..=b'z').contains(&b'a'));
+    ///     assert!((b'A'..=b'Z').contains(&b'A'));
```
xFrednet, line 30: Looking at this example, I'd suggest removing the `assert!()` macro around the expression. The code should still compile and make it easier to focus on the expression in question. I'd also say that just the normal letter ranges below are enough. This should mainly give the user an idea what is being linted.

clippy_lints/src/manual_is_ascii_check.rs
```diff
@@ -32,6 +41,14 @@ declare_clippy_lint! {
...
     ///     assert!('2'.is_ascii_digit());
     ///     assert!('x'.is_ascii_alphabetic());
+    ///
+    ///     assert!(b'0'.is_ascii_digit());
+    ///     assert!(b'a'.is_ascii_lowercase());
+    ///     assert!(b'A'.is_ascii_uppercase());
+    ///
+    ///     assert!('0'.is_ascii_digit());
+    ///     assert!('a'.is_ascii_lowercase());
+    ///     assert!('A'.is_ascii_uppercase());
```
xFrednet, line 51: I think here it would be enough to list only one of each.

tests/ui/manual_is_ascii_check.rs
```diff
@@ -15,6 +15,14 @@ fn main() {
...
 
     assert!(matches!('x', 'A'..='Z' | 'a'..='z' | '_'));
+
+    assert!((b'0'..=b'9').contains(&b'0'));
+    assert!((b'a'..=b'z').contains(&b'a'));
+    assert!((b'A'..=b'Z').contains(&b'A'));
+
+    assert!(('0'..='9').contains(&'0'));
+    assert!(('a'..='z').contains(&'a'));
+    assert!(('A'..='Z').contains(&'A'));
```
xFrednet, line 25: We should have some tests, that handle the non reference case. Something like;

```rs
let cool_letter = &'g';
('0'..='9').contains(cool_letter);
('a'..='z').contains(cool_letter);
('A'..='Z').contains(cool_letter);
```

You also don't need to wrap these into `assert!()` macros either

xFrednet (commented):
The implementation looks good to me, three more nits and then it's ready for merge. Would you mind squashing your commits a bit? Like the two ones adding tests and the two first ones with the same message. :)

tests/ui/manual_is_ascii_check.rs
naosense: Sorry, I dont see the difference between `('0'..='9').contains(&'0')` and `('0'..='9').contains(cool_letter)`, `cool_letter` dont looks like a non reference, is this a typo?

xFrednet: The difference is, that `cool_letter` is already a reference and therefore doesn't need the `&` when it's used as an argument. The implementation handles this fine, I'd just like a test for it, to ensure that it'll be handled even after some changes to the code :)

naosense: Okay, I'll add it

xFrednet (approved):

PR #10055, opened 2022-12-10
uninlined_format_args: Ignore assert! and debug_assert! before 2021 edition

taiki-e:
Similar to https://github.com/rust-lang/rust-clippy/pull/9605, but for `assert!` and `debug_assert!` macros. ([non_fmt_panics lint triggers them](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=18b20c408ec62a67f1503cd5d284424b))

changelog: [`uninlined_format_args`]: Do not inline `assert!` and `debug_assert!` macros before 2021 edition

r? @llogiq


llogiq:
Thank you!

@bors r+

PR #10068, opened 2022-12-12
Fix error caused by recursion into Self.

### LINKED ISSUE #10041: Clippy crashes on simple method returning `impl PartialOrd`
### Summary

This is probably caused by https://github.com/rust-lang/rust-clippy/pull/9733 per a crash analysis:

```
...
#30580 0x0000555c0ff6ebc0 in clippy_utils::ty::contains_ty_adt_constructor_opaque ()
#30581 0x0000555c0ff6ebc0 in clippy_utils::ty::contains_ty_adt_constructor_opaque ()
#30582 0x0000555c0ff6ebc0 in clippy_utils::ty::contains_ty_adt_constructor_opaque ()
#30583 0x0000555c0ff6ebc0 in clippy_utils::ty::contains_ty_adt_constructor_opaque ()
#30584 0x0000555c0fcbb240 in <clippy_lints::methods::Methods as rustc_lint::passes::LateLintPass>::check_impl_item ()
#30585 0x00007fe4894426dd in rustc_hir::intravisit::walk_item::<rustc_lint::late::LateContextAndPass<rustc_lint::late::LateLintPassObjects>> ()
...
```

### Reproducer

I tried this code:

```rust
pub struct Bomb;

impl Bomb {
    pub fn explode(&self) -> impl PartialOrd {
        0i32
    }
}
```

Clippy does not crash.

Instead, this happened:

Clippy crashes with a stack overflow exception.

### Version

```text
binary: rustc
commit-date: 2022-11-02
release: 1.65.0
LLVM version: 15.0.0
```

### Additional Labels

_No response_

Nashenas88:
fixes #10041

Prevent recursion into Self when it's a generic parameter. Added regression test from example in #10041.

- \[x] Followed [lint naming conventions][lint_naming] - **N/A**
- \[x] Executed `cargo dev update_lints` - **N/A**
- \[x] Added lint documentation - **N/A**
- \[x] Run `cargo dev fmt`

changelog: [`new-ret-no-self`] Fix segmentation fault caused when generic parameter defaults to `Self` and is unspecified. For example, `fn uh_oh(&self) -> impl PartialOrd { ... }` has a hidden `Rhs=Self` as the generic parameter for `PartialOrd`.

### DIFF
--- a/clippy_utils/src/ty.rs
+++ b/clippy_utils/src/ty.rs
@@ -90,6 +90,7 @@ pub fn contains_ty_adt_constructor_opaque<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'
                                 .substs
                                 .types()
                                 .skip(1) // Skip the implicit `Self` generic parameter
+                                .filter(|inner_ty| *inner_ty != ty) // Skip any other `Self` generic parameters
                                 .any(|ty| contains_ty_adt_constructor_opaque(cx, ty, needle))
                             {
                                 return true;
--- a/tests/ui/new_ret_no_self.rs
+++ b/tests/ui/new_ret_no_self.rs
@@ -400,3 +400,14 @@ mod issue7344 {
         }
     }
 }
+
+mod issue10041 {
+    struct Bomb;
+
+    impl Bomb {
+        // Hidden <Rhs = Self> default generic paramter.
+        pub fn explode(&self) -> impl PartialOrd {
+            0i32
+        }
+    }
+}


Nashenas88:
If this fix is not sufficient, I also created a revert with #10069

Nashenas88:
Let me know if I should remove the changelog line above since this is a fix.

Nashenas88:
@xFrednet let me know if you'd like me to add any additional detail or change any wording here.

Nashenas88:
Also, do I need to amend the commit with the changelog, or is it pulled from the github description instead?

tmandry:
This is fixing a bug introduced last week, do we need a changelog entry if it isn't released? (Though this will probably need a beta backport, based on timing.)

r? @dswij 
cc @nbdd0121

nbdd0121:
Sorry for the problem caused,  but ideally I'd hope #9733 is not reverted because we have code that uses pattern in #7344.

I did some poking and discovered that similar recursion can happen for assoc types as well, when using TAIT:
```rust
#![feature(type_alias_impl_trait)]
type X = impl std::ops::Add<Output=X>;

struct Foo;

impl Foo {
    fn new() -> X {
        return 1;
    }
}
```

A more reliable solution might be to have a `FxHashSet` to track all checked `Ty`s?

For a temporary solution, I suppose we can just get rid of recursion in `contains_ty_adt_constructor_opaque`, by replacing the recursive `contains_ty_adt_constructor_opaque(cx, ty, needle)` call with `ty == needle || ty.ty_adt_def() == needle.ty_adt_def()`.

xFrednet:
Thank you for the updated and detailed description.

> Also, do I need to amend the commit with the changelog, or is it pulled from the github description instead?

The entry is pulled from the GH description. Otherwise, it would also produce a bunch of merge conflicts.

> This is fixing a bug introduced last week, do we need a changelog entry if it isn't released? (Though this will probably need a beta backport, based on timing.)

It can take some time until PRs land on master. With this one, I hope to have it merged and back ported. This is very likely, and then we won't need a changelog entry. It's still better to have one in the description, in case it does take longer.

Alexendoo:
Thanks for that @nbdd0121, I had been trying to find if there were a case like that but couldn't think of one

Nashenas88:
@matthiaskrgr do we want to go with your PR then, and I'll close this one and #10069 ?

matthiaskrgr:
my pr didn't work fully and I'm not sure how to fix it unfortunately 😅

tmandry:
I think we should go ahead and revert then, all the reasoning in https://forge.rust-lang.org/compiler/reviews.html#reverts applies here.

dswij:
We just merged a forward-fix for this: https://github.com/rust-lang/rust-clippy/pull/10086.

PR #10090, opened 2022-12-16
Warn if unresolved disallowed types/macros/methods are used in clippy.toml for disallowed_* macros

i509VCB:
I am unsure if this is the right level of abstraction to work at for this feature. There doesn't seem to be an easy way to get a span of the config file to point to what directly was unresolved. This does note what entry is unresolved though so the user can see what may have been removed or is invalid for disallowed macros, types and methods.

changelog: [`disallowed_macros`]:  warn if unresolved macro is disallowed, [`disallowed_types`]:  warn if unresolved type is disallowed, [`disallowed_methods`]: warn if unresolved method is disallowed

### DIFF
--- a/clippy_lints/src/disallowed_macros.rs
+++ b/clippy_lints/src/disallowed_macros.rs
@@ -1,10 +1,11 @@
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::diagnostics::{span_lint, span_lint_and_then};
 use clippy_utils::macros::macro_backtrace;
 use rustc_data_structures::fx::FxHashSet;
 use rustc_hir::def_id::DefIdMap;
 use rustc_hir::{Expr, ForeignItem, HirId, ImplItem, Item, Pat, Path, Stmt, TraitItem, Ty};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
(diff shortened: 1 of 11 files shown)


xFrednet:
Hey, I would alter the changelog entry, to something like this:

> changelog: [`disallowed_macros`], [`disallowed_types`], [`disallowed_methods`]: Now emit a warning if the given path couldn't be resolved

Does this capture the change correctly? :upside_down_face:

Alexendoo:
This may not mesh well with workspaces, when multiple crates share a `clippy.toml` some of the paths may be reachable from some but not other crates, e.g. if some crates depend on a certain dependency but others do not

Not sure on the best thing to do here

xFrednet:
We could have the warning, be optional, by another config value, lint or other kind of verification, but that doesn't feel quite right. Ideally, it would emit a warning, if the path couldn't be found in any workspace member. However, this is not viable with the current setup of Clippy AFAIK, as each crate is compiled individually.

We could add some magic to `cargo-clippy`, but that sounds very hacky :thinking:

xFrednet:
As previously stated in the comments, I don't believe that this can be nicely implemented for workspaces with Clippy's current setup. If someone is as a suggestion, we'll be happy to discuss it, as the idea is good IMO.

For now, I'll close this PR, as there hasn't been any activity in the last year.

@i509VCB thank you for the suggestion and the time you put into this!

PR #10097, opened 2022-12-17
Rustup

flip1995:
r? @ghost

I'm on the train and my internet is too bad to download the necessary toolchain, so I have to use CI to find sync fallout.

changelog: none


flip1995:
Ah, now that all fallout is fixed, I finally downloaded the toolchain :tada:

Issue #9995, opened 2022-11-30
`std::mem::size_of_val` w/ double-reference


crepererum:
### What it does

Imagine the following code:

```rust
struct S {
    field: u32,
    data: Vec<u8>,
}

impl S {
    /// Get size of object including `self`, in bytes.
    pub fn size(&self) -> usize {
        std::mem::size_of_val(self) + (std::mem::size_of::<u8>() * self.data.capacity())
    }
}
```

Now with a little typo, this is all wrong:

```rust
impl S {
    /// Get size of object including `self`, in bytes.
    pub fn size(&self) -> usize {
        // typo is here:      V
        std::mem::size_of_val(&self) + (std::mem::size_of::<u8>() * self.data.capacity())
    }
}
```

This is because `&self` will not calculate the size of `S` but of `&S` This is most likely NOT what the user wanted, but it's easy to mess up.

This is confusing because the type signature is [`size_of_val<T>(val: &T)`](https://doc.rust-lang.org/std/mem/fn.size_of_val.html) where the the type of the size calculation is NOT a reference but the parameter is (to prevent the value from being moved just for the calculation).

See [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=33ebc9823b328bd8f1ad6a7e6a67dd6a).

### Lint Name

size_of_val_on_reference

### Category

suspicious

### Advantage

- prevents bug like [this one](https://github.com/apache/arrow-datafusion/pull/4371#discussion_r1032385224)

### Drawbacks

- there might be good reasons why a users calculates the size of a reference.

### Example

```rust
fn f(&self) {
    std::mem::size_of_val(&self)
}
```

Probably meant:

```rust
fn f(&self) {
    std::mem::size_of_val(self)
}
```

PR #10098, opened 2022-12-17
Add size_of_ref lint

lukaslueg:
This addresses #9995, which is likely raising a valid point about `std::mem::size_of_val()`: It's [very easy to use double-references as the argument](https://github.com/apache/arrow-datafusion/pull/4371#discussion_r1032385224), which the function will happily accept and give back the size of _the reference_, not the size of the value _behind_ the reference. In the worst case, if the value matches the programmer's expectation, this seems to work, while in fact, everything will go horribly wrong e.g. on a different platform.

The size of a `&T` is independent of what `T` is, and people might want to use `std::mem::size_of_val()` to actually get the size of _any_ reference (e.g. via `&&()`). I would rather suggest that this is always bad behavior, though ([instead](https://doc.rust-lang.org/reference/type-layout.html#pointers-and-references-layout), [and](https://doc.rust-lang.org/stable/std/primitive.usize.html#associatedconstant.BITS)). I, therefore, put this lint into `correctness`.

Since the problem is usually easily fixed by removing extra `&`, I went light on suggesting code.

---

changelog: New lint: [`size_of_ref`]
[#10098](https://github.com/rust-lang/rust-clippy/pull/10098)

### DIFF
--- a/clippy_lints/src/size_of_ref.rs
+++ b/clippy_lints/src/size_of_ref.rs
@@ -1,0 +1,73 @@
+use clippy_utils::{diagnostics::span_lint_and_help, path_def_id, ty::peel_mid_ty_refs};
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Checks for calls to `std::mem::size_of_val()` where the argument is
+    /// a reference to a reference.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// Calling `size_of_val()` with a reference to a reference as the argument
+    /// yields the size of the reference-type, not the size of the value behind
+    /// the reference.
+    ///
+    /// ### Example
+    /// ```rust
+    /// struct Foo {
+    ///     buffer: [u8],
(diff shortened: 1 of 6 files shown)


Jarcho (commented):
This would be better as `suspicious`. It's an easy mistake to make, but it's not inherently wrong and the alternative requires writing out the type name which could be a long.

clippy_lints/src/size_of_ref.rs
```diff
@@ -0,0 +1,73 @@
...
+            && let arg_ty = cx.typeck_results().expr_ty(arg)
+            && peel_mid_ty_refs(arg_ty).1 > 1
+        {
+            span_lint_and_help(
+                cx,
+                SIZE_OF_REF,
+                expr.span,
+                "argument to `std::mem::size_of_val()` is a reference to a reference",
+                None,
+                "dereference the argument to `std::mem::size_of_val()` to get the size of the value instead of the generic size of any reference-type",
```
Jarcho, line 69: This message needs the same edit as the lint description. Other then that everything looks good.

lukaslueg:
Ah sorry, I was lazy :-/ Will squash if needed

Jarcho:
It's better if it's squashed, but not a big deal.

lukaslueg:
squashed

Jarcho:
Thank you. @bors r+

PR #10099, opened 2022-12-18
Null fn lints

### LINKED ISSUE #1644: Lint null function pointers
https://github.com/rust-lang/rust/issues/40913 showed a real world case of assuming that `fn()` can be null.

patterns that should be linted are

```rust
std::mem::transmute::<Anything, SomeFnPtrType>(std::ptr::null())
```

```rust
(fn_ptr as *mut something).is_null()
```

Niki4tap:
Adds lints to check for code, that assumes nullable `fn()`.

### Lint examples:

`transmute_null_to_fn`:
```rust
error: transmuting a known null pointer into a function pointer
  --> $DIR/transmute_null_to_fn.rs:9:23
   |
LL |         let _: fn() = std::mem::transmute(std::ptr::null::<()>());
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this transmute results in undefined behavior
   |
   = help: try wrapping your function pointer type in `Option<T>` instead, and using `None` as a null pointer value
```

`fn_null_check`:
```rust
error: function pointer assumed to be nullable, even though it isn't
  --> $DIR/fn_null_check.rs:13:8
   |
LL |     if (fn_ptr as *mut ()).is_null() {}
   |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: try wrapping your function pointer type in `Option<T>` instead, and using `is_none` to check for null pointer value
```

Closes #1644

changelog: Fix [`transmuting_null`] false negative if `const` pointer type is not unsigned integer.
changelog: add lint [`transmute_null_to_fn`].
changelog: add lint [`fn_null_check`].

### DIFF
--- a/clippy_lints/src/fn_null_check.rs
+++ b/clippy_lints/src/fn_null_check.rs
@@ -1,0 +1,129 @@
+use clippy_utils::consts::{constant, Constant};
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::{is_integer_literal, is_path_diagnostic_item};
+use if_chain::if_chain;
+use rustc_hir::{BinOpKind, Expr, ExprKind, TyKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for comparing a function pointer to null.
+    ///
+    /// ### Why is this bad?
+    /// Function pointers are assumed to not be null.
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// let fn_ptr: fn() = /* somehow obtained nullable function pointer */
+    ///
+    /// if (fn_ptr as *const ()).is_null() { ... }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let fn_ptr: Option<fn()> = /* somehow obtained nullable function pointer */
+    ///
+    /// if fn_ptr.is_none() { ... }
+    /// ```
+    #[clippy::version = "1.67.0"]
+    pub FN_NULL_CHECK,
+    correctness,
+    "`fn()` type assumed to be nullable"
+}
+declare_lint_pass!(FnNullCheck => [FN_NULL_CHECK]);
+
+const LINT_MSG: &str = "function pointer assumed to be nullable, even though it isn't";
+const HELP_MSG: &str = "try wrapping your function pointer type in `Option<T>` instead, and using `is_none` to check for null pointer value";
+
+fn is_fn_ptr_cast(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    if_chain! {
+        if let ExprKind::Cast(cast_expr, cast_ty) = expr.kind;
+        if let TyKind::Ptr(_) = cast_ty.kind;
+        if cx.typeck_results().expr_ty_adjusted(cast_expr).is_fn();
+        then {
+            true
+        } else {
+            false
+        }
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for FnNullCheck {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        // Catching:
+        // (fn_ptr as *<const/mut> <ty>).is_null()
+        if_chain! {
+            if let ExprKind::MethodCall(method_name, receiver, _, _) = expr.kind;
+            if method_name.ident.as_str() == "is_null";
+            if is_fn_ptr_cast(cx, receiver);
+            then {
+                span_lint_and_help(
+                    cx,
+                    FN_NULL_CHECK,
+                    expr.span,
+                    LINT_MSG,
+                    None,
+                    HELP_MSG
+                );
+            }
+        }
+
+        if let ExprKind::Binary(op, left, right) = expr.kind
+            && let BinOpKind::Eq = op.node
+        {
+            let to_check: &Expr<'_>;
+            if is_fn_ptr_cast(cx, left) {
+                to_check = right;
+            } else if is_fn_ptr_cast(cx, right) {
+                to_check = left;
+            } else {
+                return;
+            }
+
+            // Catching:
+            // (fn_ptr as *<const/mut> <ty>) == <const that evaluates to null_ptr>
+            let c = constant(cx, cx.typeck_results(), to_check);
+            if let Some((Constant::RawPtr(0), _)) = c {
+                span_lint_and_help(
+                    cx,
+                    FN_NULL_CHECK,
+                    expr.span,
+                    LINT_MSG,
+                    None,
+                    HELP_MSG
+                );
+                return;
+            }
+
+            // Catching:
+            // (fn_ptr as *<const/mut> <ty>) == (0 as <ty>)
+            if let ExprKind::Cast(cast_expr, _) = to_check.kind && is_integer_literal(cast_expr, 0) {
+                span_lint_and_help(
+                    cx,
+                    FN_NULL_CHECK,
+                    expr.span,
+                    LINT_MSG,
+                    None,
+                    HELP_MSG
+                );
+                return;
+            }
+
+            // Catching:
+            // (fn_ptr as *<const/mut> <ty>) == std::ptr::null()
+            if let ExprKind::Call(func, []) = to_check.kind &&
+                is_path_diagnostic_item(cx, func, sym::ptr_null)
+            {
+                span_lint_and_help(
+                    cx,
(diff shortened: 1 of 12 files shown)


clippy_lints/src/fn_null_check.rs
```diff
@@ -0,0 +1,129 @@
...
+fn is_fn_ptr_cast(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    if_chain! {
+        if let ExprKind::Cast(cast_expr, cast_ty) = expr.kind;
+        if let TyKind::Ptr(_) = cast_ty.kind;
+        if cx.typeck_results().expr_ty_adjusted(cast_expr).is_fn();
+        then {
+            true
+        } else {
+            false
+        }
```
llogiq, line 48:
```suggestion
        then {
            cx.typeck_results().expr_ty_adjusted(cast_expr).is_fn()
        } else {
            false
        }
```

clippy_lints/src/fn_null_check.rs
```diff
@@ -0,0 +1,129 @@
...
+            if is_fn_ptr_cast(cx, receiver);
+            then {
+                span_lint_and_help(
+                    cx,
+                    FN_NULL_CHECK,
+                    expr.span,
+                    LINT_MSG,
+                    None,
+                    HELP_MSG
+                );
```
llogiq, line 68: It is custom to factor out such things as a plain function instead of using `const`s to repeat the messages through the code.

clippy_lints/src/fn_null_check.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for FnNullCheck {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        // Catching:
+        // (fn_ptr as *<const/mut> <ty>).is_null()
+        if_chain! {
+            if let ExprKind::MethodCall(method_name, receiver, _, _) = expr.kind;
```
llogiq, line 57: I would `match` on `expr.kind` once instead of having multiple `if let`s, but that's just my personal preference.

clippy_lints/src/fn_null_check.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// ```
+    /// Use instead:
+    /// ```rust,ignore
+    /// let fn_ptr: Option<fn()> = /* somehow obtained nullable function pointer */
+    ///
+    /// if fn_ptr.is_none() { ... }
+    /// ```
+    #[clippy::version = "1.67.0"]
+    pub FN_NULL_CHECK,
+    correctness,
```
llogiq, line 31: I would put this into `suspicious` instead of `correctness`. The code may not do what it should, but it's not dangerous in and of itself. `correctness` lints stop the build by default, so they might mask other noteworthy lint messages.

clippy_lints/src/transmute/mod.rs
```diff
@@ -409,6 +410,34 @@ declare_clippy_lint! {
...
+    /// ```rust
+    /// let null_fn: fn() = unsafe { std::mem::transmute( std::ptr::null::<()>() ) };
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let null_fn: Option<fn()> = None;
+    /// ```
+    #[clippy::version = "1.67.0"]
+    pub TRANSMUTE_NULL_TO_FN,
+    correctness,
```
llogiq, line 437: In contrast, here `correctness` is totally OK.

clippy_lints/src/transmute/transmute_null_to_fn.rs
```diff
@@ -0,0 +1,61 @@
...
+    "try wrapping your function pointer type in `Option<T>` instead, and using `None` as a null pointer value";
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, arg: &'tcx Expr<'_>, to_ty: Ty<'tcx>) -> bool {
+    if !to_ty.is_fn() {
+        return false;
+    }
+
+    // Catching transmute over constants that resolve to `null`.
+    let mut const_eval_context = constant_context(cx, cx.typeck_results());
+    if let ExprKind::Path(ref _qpath) = arg.kind &&
```
llogiq, line 23: Again, I would use a single `match` with multiple arms instead of many `if let`s.

llogiq (commented):
This looks good in general, I have a few suggestions regarding code style.

clippy_lints/src/fn_null_check.rs
Niki4tap: Thanks, will do

clippy_lints/src/fn_null_check.rs
Niki4tap: Yeah, you're totally right, no clue how I haven't noticed this pattern myself

clippy_lints/src/fn_null_check.rs
Niki4tap: Not sure if I understand you correctly, are you suggesting something like this:
```rust
match expr.kind {
     ExprKind::MethodCall(method_name, receiver, _, _) if
     method_name.ident.as_str() == "is_null"
     && is_fn_ptr_cast(cx, receiver) =>
     {
         lint_expr(cx, expr);
         return;
     },
     _ => {}
}
```
instead of:
```rust
if let ExprKind::MethodCall(method_name, receiver, _, _) = expr.kind
    && method_name.ident.as_str() == "is_null"
    && is_fn_ptr_cast(cx, receiver)
{
        lint_expr(cx, expr);
        return;
}
```
?

clippy_lints/src/fn_null_check.rs
Niki4tap: I would argue it should stay in `correctness` honestly.

[The clippy rfc](https://github.com/rust-lang/rfcs/blob/master/text/2476-clippy-uno.md?plain=1#L87) clearly defines `correctness` as `Probable bugs`, with the example of:
> calling `.clone()` on `&&T`, which clones the (`Copy`) reference and not the actual type

Which is in my opinion of a similar logical bug category as this lint. Furthermore, [clippy itself](https://github.com/rust-lang/rust-clippy/blob/master/README.md?plain=1#L16) defines `correctness` as:
> code that is outright wrong or useless

Which is also what I think this lint is supposed to detect, checking a function pointer for null is not _just_
> code may not do what it should

It is just an outright wrong code sample. Nobody should ever want to do this, as no function pointer in rust is assumed to be nullable.

clippy_lints/src/fn_null_check.rs
llogiq: Yes, and then you can merge the multiple match expressions.

Niki4tap: Oh, yeah sorry, didn't think of that, `match` will probably be better, will do

clippy_lints/src/transmute/transmute_null_to_fn.rs
Niki4tap: Right, changed to `match`

Niki4tap:
@llogiq I believe all code style issues have been resolved, one thing I would still like to discuss though, is whether `fn_null_check` lint should be in a different category.

Everything I've said in the [suggestion](https://github.com/rust-lang/rust-clippy/pull/10099#discussion_r1051612803) still stands, and I just mostly wanted to mention this concern here specifically, as this is not just a code style issue, and might need to be discussed outside of a suggestion. My opinion is that, while `fn_null_check` might not seem as impactful as `transmute_null_to_fn`, they both are checking for the same root issue, which is assuming that `fn()` can be null, so I believe that their severity should not be separated.

clippy_lints/src/transmute/transmute_null_to_fn.rs
```diff
@@ -0,0 +1,61 @@
...
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::{is_integer_literal, is_path_diagnostic_item};
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_middle::ty::Ty;
+use rustc_span::symbol::sym;
+
+use super::TRANSMUTE_NULL_TO_FN;
+
+const LINT_MSG: &str = "transmuting a known null pointer into a function pointer";
```
llogiq, line 11: I think we can inline those consts now that we use them only once.

Niki4tap: Done

Niki4tap:
Thanks for the review @llogiq! Is this now ready for merge?

llogiq:
Thank you!

I'm a bit wary about the correctness group for the null check lint, but that may be overcautious. We can revisit this if there are any reports of false positives or something.

@bors r+

PR #10107, opened 2022-12-21
Suggest using Path for comparing extensions

### LINKED ISSUE #10042: case_sensitive_file_extension_comparisons hits on a to_lowercase'd string
### Summary

This could be user error but I would have thought that forcing the str to lowercase would mean that it's ok to test for an extension with ends_with.

### Lint Name

_No response_

### Reproducer

I tried this code:

```rust
pub fn is_jpg(filename: &str) -> bool {
    filename.to_lowercase().ends_with(".jpg")
}
```

I saw this happen:

```
warning: case-sensitive file extension comparison
  --> (filename)
   |
(line) |     filename.to_lowercase().ends_with(".jpg")
   |                             ^^^^^^^^^^^^^^^^^
   |
   = note: `-W clippy::case-sensitive-file-extension-comparisons` implied by `-W clippy::pedantic`
   = help: consider using a case-insensitive comparison instead
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#case_sensitive_file_extension_comparisons
```

(no output)

### Version

```text
rustc 1.62.1 (e092d0b6b 2022-07-16)
binary: rustc
commit-hash: e092d0b6b43f2de967af0887873151bb1c0b18d3
commit-date: 2022-07-16
host: x86_64-apple-darwin
release: 1.62.1
LLVM version: 14.0.5
```

### Additional Labels

_No response_

tylerjw:
fixes #10042 

changelog: Sugg: [`case_sensitive_file_extension_comparisons`]: Now displays a suggestion with `Path`
[#10107](https://github.com/rust-lang/rust-clippy/pull/10107)

### DIFF
--- a/clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
+++ b/clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
@@ -1,7 +1,9 @@
-use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_lang_item;
 use if_chain::if_chain;
 use rustc_ast::ast::LitKind;
+use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, LangItem};
 use rustc_lint::LateContext;
 use rustc_span::{source_map::Spanned, Span};
@@ -28,13 +30,39 @@ pub(super) fn check<'tcx>(
         let recv_ty = cx.typeck_results().expr_ty(recv).peel_refs();
         if recv_ty.is_str() || is_type_lang_item(cx, recv_ty, LangItem::String);
         then {
-            span_lint_and_help(
+            span_lint_and_then(
                 cx,
                 CASE_SENSITIVE_FILE_EXTENSION_COMPARISONS,
-                call_span,
+                recv.span.to(call_span),
                 "case-sensitive file extension comparison",
-                None,
-                "consider using a case-insensitive comparison instead",
+                |diag| {
+                    diag.help("consider using a case-insensitive comparison instead");
+                    let mut recv_str = Sugg::hir(cx, recv, "").to_string();
+
+                    if is_type_lang_item(cx, recv_ty, LangItem::String) {
+                        recv_str = format!("&{recv_str}");
+                    }
+
+                    if recv_str.contains(".to_lowercase()") {
+                        diag.note("to_lowercase allocates memory, this can be avoided by using Path");
+                        recv_str = recv_str.replace(".to_lowercase()", "");
+                    }
+
+                    if recv_str.contains(".to_uppercase()") {
+                        diag.note("to_uppercase allocates memory, this can be avoided by using Path");
+                        recv_str = recv_str.replace(".to_uppercase()", "");
+                    }
+
+                    diag.span_suggestion(
+                        recv.span.to(call_span),
+                        "use std::path::Path",
+                        format!("std::path::Path::new({})
+        .extension()
+        .map_or(false, |ext| ext.eq_ignore_ascii_case(\"{}\"))", 
+                            recv_str, ext_str.strip_prefix('.').unwrap()),
+                        Applicability::MaybeIncorrect,
+                    );
+                }
             );
         }
     }
--- a/clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
+++ b/clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::source::snippet_opt;
 use clippy_utils::source::{indent_of, reindent_multiline};
-use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_lang_item;
 use if_chain::if_chain;
 use rustc_ast::ast::LitKind;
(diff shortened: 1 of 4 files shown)


tylerjw:
This should be marked as a Draft or WIP as adding `// run-rustfix` to the test results in code that does not compile.

Secondly, this does not fix the linked to issue as it does not detect calls to `to_lowercase` yet or help the user remove those calls.

tylerjw:
I'm could use some mentoring on what to do next about this PR.

1. Should I be using string operations to build the suggestion or using something like [fold](https://doc.rust-lang.org/beta/nightly-rustc/rustc_middle/ty/fold/index.html) from rustc?  The idea would be to traverse `recv` and remove method calls `to_lowercase` and `to_uppercase`. I looked into doing this and could not hack this together so it would work or find any examples of similar things in clippy. Advice here would be really helpful.

2. On creating a multi-line suggestion. My current approach seems to hard-code the indentation level. I also have a copy of this that makes this a single-line change, but the result is a diagnostic that runs off the right side of my terminal. Is there tooling for creating multi-line suggestions that have the correct indentation, look good in the terminal output, and are not too far from what rust fmt would create?

Alexendoo:
Hi, sorry for taking a while to get to this

For `1.` I would say no, removing a trailing `.to_lowercase()`/`.to_uppercase()` is the only place it can be done safely without other checks that are overkill since it wouldn't be that common, e.g. in `f(s.to_lowercase()).ends_with(".txt")` we don't know what's going to happen in `f`

For `2.` you could look at using [`snippet_indent`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/source/fn.snippet_indent.html) or [`reindent_multiline`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/source/fn.reindent_multiline.html) + [`indent_of`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/source/fn.indent_of.html)

tylerjw:
Thank you for the reply. Are you saying I should adjust this PR to only remove the `to_lowercase` call in the case where it is at the end of the recv code snippet? That makes allot of sense as the only place where it is safe to do this in. Should I still provide a diagnostic about maybe removing it when I detect it in the line but not at the end?

tylerjw:
@Alexendoo I updated this PR based on your feedback. Let me know if you would like me to change anything. Thank you again for the links for indenting, that was easy to fixup once I knew about those functions.

clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
```diff
@@ -28,13 +31,46 @@ pub(super) fn check<'tcx>(
...
+
+                    if recv_source.ends_with(".to_lowercase()") {
+                        diag.note("to_lowercase allocates memory, this can be avoided by using Path");
+                        recv_source = recv_source.strip_suffix(".to_lowercase()").unwrap().to_string();
+                    }
+
+                    if recv_source.ends_with(".to_uppercase()") {
+                        diag.note("to_uppercase allocates memory, this can be avoided by using Path");
+                        recv_source = recv_source.strip_suffix(".to_uppercase()").unwrap().to_string();
+                    }
```
Alexendoo, line 55: I would say detecting the use of `to_lower/uppercase` is good, but personally I think we should not lint when we detect that to be the case. Linting about a case sensitive operation when we know it's case insensitive seems a bit too confusing to me

tylerjw: Ok, I don't understand what you are asking. My understanding of the original issue was that users were using to_lowercase as a work-around to the case-sensitive nature of `ends_with`. The whole point of this lint is to catch when users do that and let them know there is a case-insensitive way to check for extensions without the allocations of is_lowercase.

Are you suggesting that if we detect `is_lowercase` we should assume the user intended to allocate and convert to the lowercase and we should not lint in that case?

Alexendoo: The purpose of the lint is to detect case sensitive comparisons and suggest changing them to case insensitive ones, hence the message being `error: case-sensitive file extension comparison`. It happens to lint `s.to_lowercase().ends_with("...")` currently, but I agree with #10042 that this is a false positive

Yeah, we should not lint in that case

tylerjw: Would you like me to change this to detect that case (`to_lowercase` at the end of the recv) and stop the lint?

tylerjw: I pushed a change that skips the lint if to_lowercase or to_uppercase is the last method called.

Alexendoo: Great, thanks

clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
```diff
@@ -28,13 +38,36 @@ pub(super) fn check<'tcx>(
...
                 cx,
                 CASE_SENSITIVE_FILE_EXTENSION_COMPARISONS,
-                call_span,
+                recv.span.to(call_span),
                 "case-sensitive file extension comparison",
-                None,
-                "consider using a case-insensitive comparison instead",
+                |diag| {
+                    diag.help("consider using a case-insensitive comparison instead");
+                    let mut recv_source = Sugg::hir(cx, recv, "").to_string();
```
Alexendoo, line 48: Pretty minor but since we don't need the `Sugg` methods here:

```suggestion
                    let Some(mut recv_source) = snippet_opt(cx, recv.span) else { return };
```

clippy_lints/src/methods/case_sensitive_file_extension_comparisons.rs
```diff
@@ -28,13 +38,36 @@ pub(super) fn check<'tcx>(
...
+
+                    if is_type_lang_item(cx, recv_ty, LangItem::String) {
```
Alexendoo, line 50: `recv_ty` has had its references peeled, but if it's a `&String` already we don't need to add a `&`, so this could be

```suggestion
                    if !cx.typeck_results().expr_ty(recv).is_ref() {
```

(or change `recv_ty` to not use `.peel_refs()` and peel it separately for line 39)

tylerjw:
I've applied the changes you suggested and tested them. Let me know if I did it correctly and if you'd like any more changes. Thank you for the review.

Alexendoo (approved):

Alexendoo:
Yep looks great thanks!

@bors r+

PR #10109, opened 2022-12-22
Fix FP in needless_return when using yeet

### LINKED ISSUE #9947: YEET!
### Summary

.

### Lint Name

needless_return

### Reproducer

I tried this code:

```rust
#![feature(yeet_expr)]

fn always_yeet() -> Result<i32, String> {
    do yeet "hello";
}

fn main() {
    assert_eq!(always_yeet(), Err("hello".to_string()));
}

```
Interestingly, it seems like `do yeet` is read as return statement (I have no idea about the desugaring here tbh) and removing it causes a type mismatch:

````
    Checking fff v0.1.0 (/tmp/fff)
warning: failed to automatically apply fixes suggested by rustc to crate `fff`

  * src/main.rs

 --> src/main.rs:4:5
  |
3 | fn always_yeet() -> Result<i32, String> {
  |                     ------------------- expected `std::result::Result<i32, std::string::String>` because of return type
4 |     "hello"
  |     ^^^^^^^ expected enum `std::result::Result`, found `&str`
  |
  = note:   expected enum `std::result::Result<i32, std::string::String>`
          found reference `&'static str`

 --> src/main.rs:4:5
  |
4 |     do yeet "hello";
  |     ^^^^^^^^^^^^^^^
  |
  = help: remove `return`

````

### Version

```text
rustc 1.67.0-nightly (b3bc6bf31 2022-11-24)
binary: rustc
commit-hash: b3bc6bf31265ac10946a0832092dbcedf9b26805
commit-date: 2022-11-24
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

Niki4tap:
Fixes #9947

changelog: Fix: [`needless_return`]: don't lint when using `do yeet`
#10109

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -6,7 +6,7 @@ use core::ops::ControlFlow;
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::FnKind;
-use rustc_hir::{Block, Body, Expr, ExprKind, FnDecl, HirId, MatchSource, PatKind, StmtKind};
+use rustc_hir::{Block, Body, Expr, ExprKind, FnDecl, HirId, LangItem, MatchSource, PatKind, QPath, StmtKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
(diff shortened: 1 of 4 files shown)


flip1995:
@bors r+

Thanks!

PR #9867, opened 2022-11-18
[arithmetic-side-effects] Consider negative numbers and add more tests

c410-f3r:
Depends on https://github.com/rust-lang/rust-clippy/pull/9840 which depends on https://github.com/rust-lang/rust-clippy/pull/9592.

Adds more tests from `integer_arithmetic` and also considers more corner-cases.

---

changelog: PF: [`arithmetic_side_effects]: No longer lints on corner cases with negative number literals
[#9867](https://github.com/rust-lang/rust-clippy/pull/9867)

PR #10112, opened 2022-12-23
[arithmetic-side-effects] Consider negative numbers and add more tests

c410-f3r:
Same as #9867.

Opening again because it is not possible to randomly choose a reviewer in an ongoing PR like in the rust repo.

---

changelog: PF: [`arithmetic_side_effects`]: No longer lints on corner cases with negative number literals
[#9867](https://github.com/rust-lang/rust-clippy/pull/9867)

### DIFF
--- a/clippy_lints/src/operators/arithmetic_side_effects.rs
+++ b/clippy_lints/src/operators/arithmetic_side_effects.rs
@@ -2,7 +2,7 @@ use super::ARITHMETIC_SIDE_EFFECTS;
 use clippy_utils::{
     consts::{constant, constant_simple},
     diagnostics::span_lint,
-    peel_hir_expr_refs,
+    peel_hir_expr_refs, peel_hir_expr_unary,
 };
 use rustc_ast as ast;
 use rustc_data_structures::fx::{FxHashMap, FxHashSet};
@@ -98,8 +98,11 @@ impl ArithmeticSideEffects {
     }
 
     /// If `expr` is not a literal integer like `1`, returns `None`.
+    ///
+    /// Returns the absolute value of the expression, if this is an integer literal.
     fn literal_integer(expr: &hir::Expr<'_>) -> Option<u128> {
-        if let hir::ExprKind::Lit(ref lit) = expr.kind && let ast::LitKind::Int(n, _) = lit.node {
+        let actual = peel_hir_expr_unary(expr).0;
+        if let hir::ExprKind::Lit(ref lit) = actual.kind && let ast::LitKind::Int(n, _) = lit.node {
             Some(n)
         }
         else {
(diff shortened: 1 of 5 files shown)


xFrednet:
Hey :wave:, I've added this to my reviewing queue. I'm currently on holiday, would it be okay, if the first review takes until the start of January?

c410-f3r:
> Hey wave, I've added this to my reviewing queue. I'm currently on holiday, would it be okay, if the first review takes until the start of January?

Don't worry. Merry Christmas and happy new year.

clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -98,8 +98,15 @@ impl ArithmeticSideEffects {
     }
 
     /// If `expr` is not a literal integer like `1`, returns `None`.
+    ///
+    /// It is worth nothing that the returned number can or can not be negative.
```
xFrednet, line 102: Would you mind explaining what you meant with the new comment?

If you mean 'noting', I'd suggest the following doc comment for the function:

```suggestion
    /// Returns the absolute value of the expression, if this is an integer literal.
```

clippy_lints/src/operators/arithmetic_side_effects.rs
```diff
@@ -98,8 +98,15 @@ impl ArithmeticSideEffects {
...
     fn literal_integer(expr: &hir::Expr<'_>) -> Option<u128> {
-        if let hir::ExprKind::Lit(ref lit) = expr.kind && let ast::LitKind::Int(n, _) = lit.node {
+        let actual = if let hir::ExprKind::Unary(hir::UnOp::Neg, elem) = expr.kind {
+            elem
+        } else {
+            expr
+        };
```
xFrednet, line 108: Tiny nit, which is probably not worth implementing, but I wanted to mention it. Rust allows multiple unary operators. A user could write `--1` which would return `None`. However, since this is only used to suppress the lint if the expression is definitely safe, I don't think it's important to handle this case :)

xFrednet (approved):
LGTM, one tiny documentation nit and then we can merge this :)

clippy_lints/src/operators/arithmetic_side_effects.rs
c410-f3r: Done

clippy_lints/src/operators/arithmetic_side_effects.rs
c410-f3r: Created an external function that "peels" unary operators,

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update and patience with the review!

@bors r+

c410-f3r:
Thank you

PR #10114, opened 2022-12-25
option_if_let_else: update known problems wording

tgross35:
changelog: none


tgross35:
I fixed the redundant wording here, but I'm not sure why this statement is under "known problems" anyway. Before merging, could someone clarify that's correct?

flip1995:
I'm not sure if the old and new wording are equivalent. The old version (why the wording is indeed bad) says that no suggestion will be emitted, while the new version says that no lint will be emitted. Can you please check, if there are tests for the case described in the `Known Problems` section?

tgross35:
@flip1995 I believe I got this one right, neither a lint nor a suggestion is emitted if it contains loop control statements. See this demo https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=48f12d4b9a1cc2202f2536bd913a469b and I think the relevant test is here https://github.com/rust-lang/rust-clippy/blob/4f3ab69ea0a0908260944443c739426cc384ae1a/tests/ui/option_if_let_else.fixed#L69-L75

tgross35:
My main question though is why is this under "Known Problems" when it's correct behavior? Since you can't `break`/`continue` from within a closure. Seems like a `# Quirks` section would be more fitting, but I don't see any precedent for that

flip1995:
Ok, then you corrected not even the wording, but also the meaning thanks!

I don't have a good explanation. I think you're right, that `# Notes` would be a better heading for this. We usually only have 3 section headers in our documentation, so I guess consistency? 

So please change it to `# Notes` and we can merge this. Thanks for the attention to the detail!

tgross35:
Okiedoke, changed the section name and added a little note about the reasoning. Should be all set 👍

tgross35:
Hey @flip1995 sorry to poke, but was this good to merge?

flip1995:
Oh sorry, totally fell of my radar. Thanks for addressing everything!

@bors r+

tgross35:
Thanks!

Issue #4295, opened 2019-07-22
Consider removing `mutex_atomic`


carllerche:
There are many reasons to use mutexes over atomics. Clippy is unable to correctly determine if an atomic is sufficient to replace the mutex use. Attempting to lint over could lead to pushing users to introduce bugs in their code.

I recommend removing this lint.

Manishearth:
I feel like this lint is correct most of the time; and clippy lints are supposed to be used with a liberal sprinkling of `allow()`. There needs to be stronger, more detailed justification for removing a lint.

carllerche:
My main point is the potential harm of taking action based on a false positive is high compared to other provided lints.

Manishearth:
I disagree, lints aren't as strong signals as compiler warnings, and Clippy's warnings are meant to be taken with a grain of salt. Here the signal is "there might be an issue", if you determine it isn't an issue, `allow()` it.

If you really think there's a problem here a stronger case needs to be made, with lots of examples.

Manishearth:
> I have only experienced a 100% false positive rate

Like I said, clippy is supposed to be used with a liberal sprinkling of `allow()`. Some lints just don't make sense for some codebases. It's not possible for our lints to be totally globally applicable. It's up to the codebase to make this judgement.

carllerche:
Ok, feel free to close if you disagree.

Manishearth:
(leaving open since other maintainers, and other people, may have different opinions)

flip1995:
The lint mentions in its `Known problems` section that
```
This lint cannot detect if the mutex is actually used for waiting before a critical section.
```

I'm with @Manishearth here. I think in the general case this lint is correct and `Atomic`s are sufficient. If you really need the mutex you should just allow the lint.

Removing this lint would require to show (with examples), that you need a Mutex in the general case.

thombles:
[...]

PR #10115, opened 2022-12-25
Move `mutex_atomic` to `restriction`

rdrpenguin04:
By #4295, the general consensus seems to be that `mutex_atomic` is not a useful lint in most cases. If anything, it could be useful as a restriction on code that for whatever reason can't use atomics. Keeping it in `clippy::nursery` is harmful to people attempting to use clippy for soundness.

---

changelog: Moved [`mutex_atomic`] to `restriction`
[#10115](https://github.com/rust-lang/rust-clippy/pull/10115)

### DIFF
--- a/clippy_lints/src/mutex_atomic.rs
+++ b/clippy_lints/src/mutex_atomic.rs
@@ -39,7 +39,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "pre 1.29.0"]
     pub MUTEX_ATOMIC,
-    nursery,
+    restriction,
     "using a mutex where an atomic value could be used instead"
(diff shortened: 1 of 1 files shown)


flip1995:
If this lint is moved to `restriction` at least the documentation should summarize the discussions in the issue, why users of this lint should be careful.

rdrpenguin04:
I have added a comment. I'm still not sure "shooting flies with cannons" is correct language for a formal issue description, but I didn't write it, and it's likely another PR's job to decide to replace it.

flip1995:
@bors r+

Thanks! The description in the documentation LGTM.

PR #10123, opened 2022-12-28
Add `hidden_static_lifetime` lint

### LINKED ISSUE #10108: Detect named lifetimes that are effectively `'static`
### What it does

[Given a function that returns a borrow to a lifetime that isn't used in any parameter](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=6523c695187210473ebc36fd11ca3f10), give and explanation and suggest removing the named lifetime and making it `'static`, as it is the only effective lifetime that will work:

```
fn foo<'a>() -> &'a str { todo!() }
```

### Lint Name

hidden_static_lifetime

### Category

suspicious, style, pedantic

### Advantage

It makes the code easier to follow and is a good thing to teach users about.

### Drawbacks

_No response_

### Example

```rust
fn foo<'a>() -> &'a str { todo!() }
```

Could be written as:

```rust
fn foo() -> &'static str { todo!() }
```

blyxyas:
Fixes #10108 

---

changelog: New lint: [`hidden_static_lifetime`]:
[#10123](https://github.com/rust-lang/rust-clippy/pull/10123)

### DIFF
--- a/clippy_lints/src/functions/hidden_static_lifetime.rs
+++ b/clippy_lints/src/functions/hidden_static_lifetime.rs
@@ -1,15 +1,18 @@
+use crate::rustc_lint::LintContext;
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use rustc_errors::Applicability;
 use rustc_hir::{
-    intravisit::FnKind, FnDecl, FnRetTy, GenericArg, GenericBound, GenericParamKind, LifetimeParamKind, QPath, TyKind,
-    WherePredicate,
+    intravisit::FnKind, FnDecl, FnRetTy, GenericArg, GenericBound, GenericParamKind, LifetimeParamKind, ParamName,
+    /* PredicateOrigin */ QPath, Ty, TyKind, TypeBindingKind, WherePredicate,
 };
 use rustc_lint::LateContext;
+use rustc_middle::lint::in_external_macro;
+use rustc_span::Span;
 
 use super::HIDDEN_STATIC_LIFETIME;
 
-pub(super) fn check_fn<'tcx>(cx: &LateContext<'_>, kind: FnKind<'tcx>, decl: &'tcx FnDecl<'_>) {
-    if let FnKind::ItemFn(_, generics, _) = kind {
+pub(super) fn check_fn<'tcx>(cx: &LateContext<'_>, kind: FnKind<'tcx>, decl: &'tcx FnDecl<'_>, span: Span) {
+    if !in_external_macro(cx.sess(), span) && let FnKind::ItemFn(_, generics, _) = kind {
         let mut lifetime_is_used;
         for generic in generics.params.iter() {
             if let GenericParamKind::Lifetime { kind } = generic.kind &&
@@ -21,25 +24,14 @@ pub(super) fn check_fn<'tcx>(cx: &LateContext<'_>, kind: FnKind<'tcx>, decl: &'t
 						break;
 					}
 					// If input is reference
-					if let TyKind::Rptr(lifetime, _) = input.kind {
-						if lifetime.ident == generic.name.ident() {
-							lifetime_is_used = true;
+					if let TyKind::Rptr(lifetime, mut_ty) = &input.kind {
+						if !lifetime.is_anonymous() && lifetime.ident == generic.name.ident() {
+								lifetime_is_used = true;
+						} else {
+							lifetime_is_used = check_if_uses_lifetime(mut_ty.ty, &generic.name);
 						}
-					}
-					// If input is struct using a lifetime
-					if let TyKind::Path(qpath) = &input.kind {
-						if let QPath::Resolved(_, path) = qpath {
-							for segment in path.segments {
-								for arg in segment.args().args {
-									// If input's lifetime and function's are the same.
-									if let GenericArg::Lifetime(lifetime) = arg {
-										if lifetime.hir_id.owner == generic.hir_id.owner {
-											lifetime_is_used = true;
-										}
-									}
-								}
-							}
-						}
+					} else {
+						lifetime_is_used = check_if_uses_lifetime(input, &generic.name);
 					}
 				}
 
@@ -54,14 +46,27 @@ pub(super) fn check_fn<'tcx>(cx: &LateContext<'_>, kind: FnKind<'tcx>, decl: &'t
 									if let GenericBound::Outlives(_) = bound {
 										lifetime_is_used = true;
 									} else {
-										span_lint_and_sugg(cx,
-											HIDDEN_STATIC_LIFETIME,
-											bound.span(),
-											"This lifetime can be changed to `'static`",
-											"try",
-											"changing this lifetime to 'static".into(),
-											Applicability::MachineApplicable
-										);
+										// Check that generic isn't X<A = B>.
+										if let GenericBound::Trait(poly_trait_ref, _) = bound {
+											for segment in poly_trait_ref.trait_ref.path.segments {
+												if let Some(gen_args) = segment.args {
+													for ty_binding in gen_args.bindings {
+														if let TypeBindingKind::Equality { .. } = ty_binding.kind {
+															lifetime_is_used = true;
+														}
+													}
+												}
+											}
+										} else {
+											span_lint_and_sugg(cx,
+												HIDDEN_STATIC_LIFETIME,
+												bound.span(),
+												"this lifetime can be changed to `'static`",
+												"try",
+												format!("changing the lifetime `{}` to 'static", generic.name.ident().as_str()),
+												Applicability::MaybeIncorrect
+											);
+										}
 									}
 								}
 							} else {
@@ -73,10 +78,10 @@ pub(super) fn check_fn<'tcx>(cx: &LateContext<'_>, kind: FnKind<'tcx>, decl: &'t
 										span_lint_and_sugg(cx,
 											HIDDEN_STATIC_LIFETIME,
 											region_predicate.span,
-											"This lifetime can be changed to `'static`",
+											"this lifetime can be changed to `'static`",
 											"try",
-											"changing this lifetime to 'static".into(),
-											Applicability::MachineApplicable
+											format!("changing the lifetime `{}` to 'static", generic.name.ident().as_str()),
+											Applicability::MaybeIncorrect
 										);
 									}
 								}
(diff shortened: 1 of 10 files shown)


clippy_lints/src/functions/mod.rs
```diff
@@ -326,6 +327,25 @@ declare_clippy_lint! {
     "getter method returning the wrong field"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects lifetimes that are secretly `'static`.
```
lukaslueg, line 332: My two cents: The language "secretly" is extremely vague. Maybe: "Detects generic lifetimes that the compiler will always resolve to be `'static`"

clippy_lints/src/functions/mod.rs
```diff
@@ -326,6 +327,25 @@ declare_clippy_lint! {
...
+    /// ### Why is this bad?
+    /// One could think that the lifetime is dropped eventually, but it isn't, as it is really a `'static`
```
lukaslueg, line 334: More cents: Similar to above, language "lifetime is dropped": lifetimes are never dropped. Maybe: "The introduction of a lifetime parameter that can only resolve to `'static` adds unnecessary complexity."

tests/ui/hidden_static_lifetime.stderr
```diff
@@ -0,0 +1,16 @@
+error: this lifetime can be changed to `'static`
+  --> $DIR/hidden_static_lifetime.rs:17:6
+   |
+LL | fn a<'a>() -> &'a str {
+   |      ^^ help: try: `changing the lifetime `'a` to 'static`
```
lukaslueg, line 5: More cents: Avoid "help: try:". The fix seems incorrect. Shouldn't the correct help be "Try removing the lifetime-parameter and change references to it to `'static`"

clippy_lints/src/loops/never_loop.rs
```diff
@@ -228,6 +228,7 @@ fn never_loop_expr(expr: &Expr<'_>, ignore_ids: &mut Vec<HirId>, main_loop_id: H
     }
 }
 
+#[clippy::dump]
```
lukaslueg, line 231: Leftover?

blyxyas: Yes, it is a leftover, thanks for pointing it out.

tests/ui/hidden_static_lifetime.stderr
blyxyas: Suggestion applied.

clippy_lints/src/functions/mod.rs
blyxyas: Suggestion applied.

clippy_lints/src/functions/mod.rs
blyxyas: Suggestion applied.

Jarcho:
This needs to be limited by quite a bit. The only cases which this is valid for are when the lifetime is used only in immutable references which are neither behind a mutable reference, nor used in function types; and mutable references to static types.

Some valid examples:
* `&'a T` -> `&'static T`
* `&'a &'b T` -> `&'static &'static T`
* `&'a mut T` -> `&'static mut T` where `T: 'static`

Some invalid examples:
* `&'a mut &'b T` -> `&'static mut &'static T`
* `fn(&'a T)` -> `fn(&'static T)`

Whether switching struct lifetime parameters to `'static` is valid or not depends on the content of the struct, so those can't be easily switched either.

blyxyas:
How could I test this?
I'm struggling to test it, because functions cannot return references, so `&'a &'b T`, instead of activating the lint, trigger compiler errors. I cannot use `todo!()` because functions that panic aren't linted.

Jarcho:
The current code doesn't do anything to avoid linting on panic. Worst case you could do `unsafe { *ptr::null() }`.

blyxyas:
The lint doesn't, but it seems that a code that should trigger the lint, doesn't, if the code has a `todo!()`, I assume because of how `cargo test` is made. Maybe there's an option for ignoring function's bodies?

Also, `unsafe {  *std::ptr::null() }` doesn't work, it triggers [E0507]: Cannot move out of a raw pointer:
Using this code:

```rust
fn l<'l, T>() -> &'l mut T where T: 'static + Copy {
	unsafe { *std::ptr::null() }
}
```

I'm clueless, I don't know how to test this lint.

Jarcho:
Try `unsafe { std::ptr::null().read() }` instead. That should have no constraints on the type it can return.

blyxyas:
Thanks for the snippet, it works.

Jarcho:
This needs a test for with something like `struct S<'a>(&'a mut i32)` to make sure that doesn't get linted.

Jarcho:
Struct fields follow the same rules that other types do. Some more examples:

```rust
struct S<T>(T);
struct Ref<'a, T: 'a>(&'a T);

struct Both<'a, 'b, T> {
    owned: T,
    borrow: &'a T,
    mut_borrow: &'b mut T,
}

fn f1<'a, 'b>() -> S<&'a mut Ref<'b, u32>> {} // can't change
fn f2<'a, 'b>() -> &'a Both<'a, 'b, &'a str> {} // can't change
fn f3<'a, 'b, 'c>() -> &'a Both<'a, 'b, &'c str> {} // 'a can be 'static
fn f4<'a, 'b>() -> &'a Both<'a, 'b, &'static str> {} // 'a and 'b can be 'static
fn f5<'a> -> Ref<'a, fn(S<Ref<'a, u32>>)> {} // can't change
```

Handling structs correctly is going to be a fair bit of work and is better handled not at the HIR level since substitutions need to be taken into account.

I would say it's better to just handle the simple case of a top level reference (or multiple references). Unless you really want to go all the way with it.

---

A couple of notes about the the implementation.

* A visitor can be use to simplify looking for a lifetime.
* `where` clauses shouldn't be changed. `T: 'a` isn't the same as `T: 'static`. They still need to be checked for lifetime use though.

blyxyas:
> Better handled not at the HIR level
What other levels can I access? Isn't Clippy specifically for checking the HIR?

blyxyas:
Also, making this lint only trigger for function returns and ignore structs is probably the best idea. A new lint can possibly be implemented in the future for struct checking. But it doesn't make sense to check structs with a lint in the `functions` category. In the new commit, if a lifetime is used in any other expressions that isn't the return type reference, it will be discarded.

Jarcho:
This should use a [`Visitor`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/intravisit/trait.Visitor.html) for most of the implementation. You can see `clippy_utils\visitors` for some examples.

clippy_lints/src/functions/hidden_static_lifetime.rs
```diff
@@ -0,0 +1,280 @@
...
+												if let Some(gen_args) = segment.args {
+													for ty_binding in gen_args.bindings {
+														if let TypeBindingKind::Equality { .. } = ty_binding.kind {
+															lifetime_is_used = true;
+														};
+													};
+												};
+											};
+										} else {
+											span_lint_and_help(cx,
```
Jarcho, line 73: This is not the same as a `'static` bound.

clippy_lints/src/functions/hidden_static_lifetime.rs
```diff
@@ -0,0 +1,280 @@
...
+										};
+									};
+								};
+							} else {
+								// Check for other lifetimes
+								if let WherePredicate::RegionPredicate(region_predicate) = predicate {
+									if region_predicate.lifetime.hir_id.owner == generic.hir_id.owner {
+										lifetime_is_used = true;
+									} else {
+										span_lint_and_help(cx,
```
Jarcho, line 87: This is not the same as a `'static` bound.

clippy_lints/src/functions/hidden_static_lifetime.rs
```diff
@@ -0,0 +1,280 @@
...
+                    }
+                }
+            }
+        }
+    };
+    true
+}
+
+// true = lifetime isn't used (success)
+fn lifetime_not_used_in_ty(ret_ty: &Ty<'_>, generic: &GenericParam<'_>) -> bool {
```
Jarcho, line 179: This should be merged with the previous function. While the type is a reference you're unwrapping it looking for lifetimes to change to `'static`. Once you hit a type which isn't a reference you're then looking to see if the lifetime is used again.

Jarcho:
> What other levels can I access? Isn't Clippy specifically for checking the HIR?

You would use the middle layer's representation of everything. It can all be retrieved from various `TyCtxt` queries. If you're just handling top-level references, there's no need to go this route.

blyxyas:
I've been trying to convert the old code to a new lint using `Visitor` for a few days and I can't really see it.
Is it necessary? The old code worked fine and I don't see the need to use Visitors for this, is it for performance?

The [last commit](https://github.com/rust-lang/rust-clippy/pull/10123/commits/edb006f41828793034d901e8a47e64248aacd4a9) contains my current code for converting this lint to a Visitor, it doesn't work and I don't even know if I'm doing things right.

Is it really necessary? If so, is there something I'm doing really wrong that I should be fixing right now?

Jarcho:
You misunderstood how to use a visitor there. You should only need to override `check_lifetime`, leaving everything else to the default behaviour of walking the AST. You can then call `visit_ty` and `visit_where_predicate` to feed data into the visitor.

Essentially something like (not valid code):
```
struct Visitor {
    lifetimes: Vec<_>,
}
impl Visitor for Visitor {
    fn check_lifetime(&mut self, lifetime) {
        if let Some(i) = self.lifetimes.iter().position(|x| x == lifetime) {
            self.lifetimes.remove(i);
        }
    }
}
fn lint(...) {
    let mut visitor = Visitor { lifetimes: ... };
    visitor.visit_ty(arg1);
    visitor.visit_ty(arg2);
    visitor.visit_where_predicate(pred1);
    // use remaining lifetimes
}
```

blyxyas:
It's been 48 - 49 days since this PR was created. I'm very sorry for wasting your time. Once again I chose an issue with a scope far bigger than I imagined, and thought it would be easy. That's a childish mentality and, in all future PRs I should stick to easier, quicker issues.

**For anyone attempting this in the future**: this [last commit](https://github.com/rust-lang/rust-clippy/commit/8e032a6045c4397c7046d1f8f00351ff3667d0dc) contains my work-in-progress code (It obviously doesn't work). You may also want to check the [last commit before the revamp](https://github.com/rust-lang/rust-clippy/pull/10123/commits/d1b1f3e33bfdcd0d8c7941e6123357585987c60a) for some support functions or notes.

Im very sorry to my reviewer, @Jarcho for wasting their time. I'm going to close this PR and unnassign myself from the issue. If anyone wants to make another PR about #10108, it's completely free.

PR #9491, opened 2022-09-17
[`drop_copy`]: Do not lint idiomatic in match arm

kraktus:
close https://github.com/rust-lang/rust-clippy/issues/9482

changelog: [`drop_copy`]: Do not lint idiomatic in match arm

Issue #9482, opened 2022-09-15
drop_copy should not trigger when drop is used in a match branch


ia0:
### Summary

The lint description says:

> drop does nothing for types that implement Copy

This is wrong, drop can be used to forget that a function returns a non-unit type when the result value is sometimes useless.

### Lint Name

drop_copy

### Reproducer

I tried [this code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=06a56b6a3ad4b2bb610a3e8b557f6885):

```rust
fn foo() -> u8 {
    println!("doing foo");
    0 // result is not always useful, the side-effect matters
}
fn bar() {
    println!("doing bar");
}

fn main() {
    match 42 {
        0 => drop(foo()),  // drop is needed because we only care about side-effects
        1 => bar(),
        _ => (),  // doing nothing (no side-effects needed here)
    }
}
```

I saw this happen:

```
error: calls to `std::mem::drop` with a value that implements `Copy`. Dropping a copy leaves the original intact
  --> src/main.rs:11:14
   |
11 |         0 => drop(foo()),  // drop is needed because we only care about side-effects
   |              ^^^^^^^^^^^
   |
   = note: `#[deny(clippy::drop_copy)]` on by default
note: argument has type `u8`
  --> src/main.rs:11:19
   |
11 |         0 => drop(foo()),  // drop is needed because we only care about side-effects
   |                   ^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_copy
```

I expected to see this happen:

No error

### Version

```text
rustc 1.63.0 (4b91a6ea7 2022-08-08)
binary: rustc
commit-hash: 4b91a6ea7258a947e59c6522cd5898e7c0a6a88f
commit-date: 2022-08-08
host: x86_64-unknown-linux-gnu
release: 1.63.0
LLVM version: 14.0.5
```

xerenahmed:
> drop can be used to forget that a function returns a non-unit

No, if value is Copy then not.

kraktus:
I don't think this use of `drop` is idiomatic, do you have any reference or example in "canonical" code? You would do ` => {foo();}` or `let _ = foo();` which the common way to discard unused return values

ia0:
> I don't think this use of drop is idiomatic

That's a fair point. See below for a solution if we go in that direction.

> do you have any reference or example in "canonical" code

[...]

PR #10142, opened 2023-01-02
[`drop_ref`]: don't lint idiomatic in match arm

### LINKED ISSUE #10122: drop_ref should not trigger when drop is used in a match branch
### Summary

This bug is similar to #9482 but for `drop_ref` instead of `drop_copy`.

I would similarly argue that if `drop()` is called on the result of a function (or other expression with side-effects) as a match branch (i.e. the code contains something like `=> drop(expr_with_side_effects),`) then the `drop_ref` lint should not trigger because `drop()` is not used to drop the value but to ignore the value.

The fix may be similar to #9491.

### Lint Name

drop_ref

### Reproducer

I tried this code:

```rust
fn foo(xs: &mut [u8; 3]) -> &mut u8 {
    println!("doing foo");
    &mut xs[2] // result is not always useful, the side-effect matters
}
fn bar() {
    println!("doing bar");
}

fn main() {
    let mut xs = *b"abc";
    match 42 {
        0 => drop(foo(&mut xs)),  // drop is needed because we only care about side-effects
        1 => bar(),
        _ => (),  // doing nothing (no side-effects needed here)
    }
}
```

I saw this happen:

```
error: calls to `std::mem::drop` with a reference instead of an owned value. Dropping a reference does nothing
  --> src/main.rs:12:14
   |
12 |         0 => drop(foo(&mut xs)),  // drop is needed because we only care about side-effects
   |              ^^^^^^^^^^^^^^^^^^
   |
note: argument has type `&mut u8`
  --> src/main.rs:12:19
   |
12 |         0 => drop(foo(&mut xs)),  // drop is needed because we only care about side-effects
   |                   ^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#drop_ref
```

No error

### Version

```text
rustc 1.67.0-nightly (c97b539e4 2022-11-30)
binary: rustc
commit-hash: c97b539e408ea353f4fde2f9251d598291fec421
commit-date: 2022-11-30
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

EricWu2003:
fixes #10122

As established in issue #9482, it is idiomatic to use a single `drop()` expression in a match arm to achieve a side-effect of a function while discarding its output. This should also apply to cases where the function returns a reference.

The change to the lint's code was less than 1 line, because all the heavy lifting was done in PR #9491.

---

changelog: FP: [`drop_ref`]: No longer lints idiomatic expression in `match` arms
[#10142](https://github.com/rust-lang/rust-clippy/pull/10142)

### DIFF
--- a/clippy_lints/src/drop_forget_ref.rs
+++ b/clippy_lints/src/drop_forget_ref.rs
@@ -206,7 +206,7 @@ impl<'tcx> LateLintPass<'tcx> for DropForgetRef {
             let is_copy = is_copy(cx, arg_ty);
             let drop_is_single_call_in_arm = is_single_call_in_arm(cx, arg, expr);
             let (lint, msg) = match fn_name {
-                sym::mem_drop if arg_ty.is_ref() => (DROP_REF, DROP_REF_SUMMARY),
+                sym::mem_drop if arg_ty.is_ref() && !drop_is_single_call_in_arm => (DROP_REF, DROP_REF_SUMMARY),
                 sym::mem_forget if arg_ty.is_ref() => (FORGET_REF, FORGET_REF_SUMMARY),
                 sym::mem_drop if is_copy && !drop_is_single_call_in_arm => (DROP_COPY, DROP_COPY_SUMMARY),
                 sym::mem_forget if is_copy => (FORGET_COPY, FORGET_COPY_SUMMARY),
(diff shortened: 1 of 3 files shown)


EricWu2003:
@kraktus Could you please take a look at this PR, since you worked on this code previously?

kraktus:
Sounds good to me 👍

llogiq:
Thank you, that looks like a sensible approach.

@bors r+

PR #10149, opened 2023-01-03
Don't pass `--sysroot` twice if SYSROOT is set

jyn514:
This is useful for rust-lang/rust to allow setting a sysroot that's *only* for build scripts, different from the regular sysroot passed in RUSTFLAGS (since cargo doesn't apply RUSTFLAGS to build scripts or proc-macros).

That said, the exact motivation is not particularly important: this fixes a regression from
https://github.com/rust-lang/rust-clippy/pull/9881/commits/5907e9155ed7f1312d108aa2110853472da3b029#r1060215684.

Note that only RUSTFLAGS is tested in the new integration test; passing --sysroot through `clippy-driver` never worked as far as I can tell, and no one is using it, so I didn't fix it here.

Helps with https://github.com/rust-lang/rust/pull/106394.

---

changelog: other: `SYSROOT` and `--sysroot` can now be set at the same time
[#10149](https://github.com/rust-lang/rust-clippy/pull/10149)

### DIFF
--- a/src/driver.rs
+++ b/src/driver.rs
@@ -256,11 +256,14 @@ pub fn main() {
     LazyLock::force(&ICE_HOOK);
     exit(rustc_driver::catch_with_exit_code(move || {
         let mut orig_args: Vec<String> = env::args().collect();
+        let has_sysroot_arg = arg_value(&orig_args, "--sysroot", |_| true).is_some();
 
         let sys_root_env = std::env::var("SYSROOT").ok();
         let pass_sysroot_env_if_given = |args: &mut Vec<String>, sys_root_env| {
             if let Some(sys_root) = sys_root_env {
-                args.extend(vec!["--sysroot".into(), sys_root]);
+                if !has_sysroot_arg {
+                    args.extend(vec!["--sysroot".into(), sys_root]);
+                }
             };
         };
 
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -1,42 +1,28 @@
-//! To run this test, use
+//! This test is meant to only be run in CI. To run it locally use:
+//!
 //! `env INTEGRATION=rust-lang/log cargo test --test integration --features=integration`
 //!
 //! You can use a different `INTEGRATION` value to test different repositories.
+//!
+//! This test will clone the specified repository and run Clippy on it. The test succeeds, if
+//! Clippy doesn't produce an ICE. Lint warnings are ignored by this test.
 
 #![cfg(feature = "integration")]
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
 #![warn(rust_2018_idioms, unused_lifetimes)]
 
+use std::env;
 use std::ffi::OsStr;
-use std::path::{Path, PathBuf};
 use std::process::Command;
-use std::{env, eprintln};
 
 #[cfg(not(windows))]
 const CARGO_CLIPPY: &str = "cargo-clippy";
 #[cfg(windows)]
 const CARGO_CLIPPY: &str = "cargo-clippy.exe";
 
-// NOTE: arguments passed to the returned command will be `clippy-driver` args, not `cargo-clippy`
-// args. Use `cargo_args` to pass arguments to cargo-clippy.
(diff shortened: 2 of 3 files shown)


src/driver.rs
```diff
@@ -256,11 +256,14 @@ pub fn main() {
     LazyLock::force(&ICE_HOOK);
     exit(rustc_driver::catch_with_exit_code(move || {
         let mut orig_args: Vec<String> = env::args().collect();
+        let has_sysroot_arg = arg_value(&orig_args, "--sysroot", |_| true).is_some();
```
jyn514, line 259: Note that if we *do* want to support passing --sysroot to clippy-driver (with `--` to cargo-clippy, not rustflags), this will need to be changed to look at `clippy_args` as well, not just `orig_args`.

tests/integration.rs
```diff
@@ -1,19 +1,42 @@
...
 const CARGO_CLIPPY: &str = "cargo-clippy.exe";
 
-#[cfg_attr(feature = "integration", test)]
-fn integration_test() {
-    let repo_name = env::var("INTEGRATION").expect("`INTEGRATION` var not set");
+// NOTE: arguments passed to the returned command will be `clippy-driver` args, not `cargo-clippy`
+// args. Use `cargo_args` to pass arguments to cargo-clippy.
+fn clippy_command(repo_dir: &Path, cargo_args: &[&str]) -> Command {
+    let root_dir = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+    let target_dir = option_env!("CARGO_TARGET_DIR").map_or_else(|| root_dir.join("target"), PathBuf::from);
```
jyn514, line 24: This is a small unrelated change, I have a global CARGO_TARGET_DIR set. It's not set in CI so I don't think it should matter too much in practice.

tests/integration.rs
```diff
@@ -90,3 +104,42 @@ fn integration_test() {
...
+    let rustc_output = Command::new(rustc)
+        .args(["--print", "sysroot"])
+        .output()
+        .expect("unable to run rustc");
+    assert!(rustc_output.status.success());
+    let sysroot = String::from_utf8(rustc_output.stdout).unwrap();
+    let sysroot = sysroot.trim_end();
+
+    #[track_caller]
+    fn verify_cmd(cmd: &mut Command) {
```
jyn514, line 120: Making this a separate function isn't strictly necessary but it will make it easier to test clippy-driver in the future if you choose to.

tests/integration.rs
```diff
@@ -90,3 +104,42 @@ fn integration_test() {
...
+        }
+        if !output.status.success() {
+            panic!("clippy exited with an error")
+        }
+    }
+
+    // This is a fairly small repo; we want to avoid checking out anything heavy twice, so just
+    // hard-code it.
+    let repo_name = "rust-lang/log";
+    let repo_dir = repo_dir(repo_name);
```
jyn514, line 139: Unfortunately this test will run for each repo in the matrix :/ I'm not sure how to get it to run only once - maybe we can update CI to run it in its own job? If it helps it's very quick to execute, log is quite small.

flip1995 (changes requested):
The `integration` test is not the right place for this. The `integration` test is really just for running Clippy on a repo and check that it doesn't produce an ICE.

IIUC what you want to test is that when running `SYSROOT=/path/to/smth RUSTFLAGS=--sysroot=/path/tp/lib cargo clippy` will only use the sysroot given by `RUSTFLAGS`?

I'll try to move this to `.github/driver.sh` tomorrow before merging.

jyn514:
> IIUC what you want to test is that when running SYSROOT=/path/to/smth RUSTFLAGS=--sysroot=/path/tp/lib cargo clippy will only use the sysroot given by RUSTFLAGS?

Yes, exactly.

> I'll try to move this to .github/driver.sh tomorrow before merging.

❤️

flip1995:
@bors r+ 

Thanks for addressing this and sorry for taking so long to review. I'll sync this to rustc right after this PR is merged.

src/driver.rs
daivinhtran: Could I send a PR to support passing `--sysroot` to `clippy-driver`?

Issue #10089, opened 2022-12-15
Clippy --fix for box_default creates invalid code


nsabovic:
After upgrading to 1.66 I got a few new clippy errors. Running `clippy --fix` produced this error instructing me to file a bug report. 

Seems like it attempted to convert:

```rust
let _: Box<dyn Trait> = Box::new(ConcreteType::default());
```
with
```rust
let _: Box<dyn Trait> = Box::default();
```

...which obviously fails.

```sh
warning: failed to automatically apply fixes suggested by rustc to crate `optimization`

after fixes were automatically applied the compiler reported errors within these files:

  * optimization/src/driver.rs

This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see 
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:
error[E0277]: the trait bound `dyn optimizer::Optimizer + std::marker::Send + std::marker::Sync: std::default::Default` is not satisfied
  --> optimization/src/driver.rs:80:13
   |
80 |             Box::default(),
   |             ^^^^^^^^^^^^ the trait `std::default::Default` is not implemented for `dyn optimizer::Optimizer + std::marker::Send + std::marker::Sync`
   |
   = help: the following other types implement trait `std::default::Default`:
             std::boxed::Box<T>
             std::boxed::Box<[T]>
             std::boxed::Box<std::ffi::CStr>
             std::boxed::Box<std::ffi::OsStr>
             std::boxed::Box<str>
   = note: required for `std::boxed::Box<dyn optimizer::Optimizer + std::marker::Send + std::marker::Sync>` to implement `std::default::Default`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
Original diagnostics will follow.

       Fixed zemax/src/from_zemax/tests/mod.rs (1 fix)
warning: `Box::new(_)` of default value
  --> optimization/src/driver.rs:80:13
   |
80 |             Box::new(NadamOptimizer::default()),
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `Box::default()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#box_default
[...]

PR #10153, opened 2023-01-03
trim paths in `box_default`

llogiq:
This might help with #10089, though I have not tested that yet. In any event, it keeps the suggestion short and to the point.

---
changelog: Trim paths in [`box_default`] suggestion


Jarcho:
How does this handle cases where the path is actually needed?

llogiq:
The macro just reduces paths that have active imports, so the shortened paths are guaranteed to lead to the same item as the fully qualified paths.

It's also used in the compiler, btw.

Jarcho:
Sounds good. @bors r+

PR #10160, opened 2023-01-04
trim paths in `default_trait_access`/`clone_on_copy` suggestions

llogiq:
This should help making the suggestions more palatable. Similar to #10153.

---

changelog: trim paths in [`default_trait_access`]/[`clone_on_copy`] suggestions


dswij (approved):
Nice! LGTM

Thanks for this

PR #10161, opened 2023-01-05
Expand derivable-impls to cover enums with a default unit variant.

khuey:

changelog: [`derivable_impls`]: suggest deriving Default for enums with a default unit variant


llogiq:
Thank you!

@bors r+

PR #10162, opened 2023-01-05
Fix FP of single-element-loop

### LINKED ISSUE #10018: FP single-element-loop: nested control flow
### Summary

We have to traverse the entire body to check for related control flow (break / continue) maybe and not warn then?

### Lint Name

single-element-loop

### Reproducer

I tried this code:

```rust
#![warn(clippy::if_same_then_else)]
#![allow(
    clippy::blacklisted_name,
    clippy::cognitive_complexity,
    clippy::collapsible_if,
    clippy::ifs_same_cond,
    clippy::needless_return
)]

fn if_same_then_else2() -> Result<&'static str, ()> {
    if true {
        for _ in &[42] {
            let foo: &Option<_> = &Some::<u8>(42);
            if true {
                break;
            } else {
                continue;
            }
        }
    } else {

    }
    todo!()
}

fn main() {}
```

I saw this happen:

```

  * src/main.rs

error[E0268]: `break` outside of a loop or labeled block
  --> src/main.rs:16:17
   |
16 |                 break;
   |                 ^^^^^ cannot `break` outside of a loop or labeled block

error[E0268]: `continue` outside of a loop
  --> src/main.rs:18:17
   |
18 |                 continue;
   |                 ^^^^^^^^ cannot `continue` outside of a loop

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0268`.

warning: lint `clippy::blacklisted_name` has been renamed to `clippy::disallowed_names`
 --> src/main.rs:3:5
  |
3 |     clippy::blacklisted_name,
  |     ^^^^^^^^^^^^^^^^^^^^^^^^ help: use the new name: `clippy::disallowed_names`
  |
  = note: `#[warn(renamed_and_removed_lints)]` on by default

warning: unused variable: `foo`
  --> src/main.rs:13:17
   |
13 |             let foo: &Option<_> = &Some::<u8>(42);
   |                 ^^^ help: if this is intentional, prefix it with an underscore: `_foo`
   |

warning: function `if_same_then_else2` is never used
  --> src/main.rs:10:4
   |
10 | fn if_same_then_else2() -> Result<&'static str, ()> {
   |    ^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: for loop over a single element
  --> src/main.rs:12:9
   |
12 | /         for _ in &[42] {
13 | |             let foo: &Option<_> = &Some::<u8>(42);
14 | |             if true {
15 | |                 break;
...  |
18 | |             }
19 | |         }
   | |_________^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_element_loop
   = note: requested on the command line with `-W clippy::single-element-loop`
help: try
   |
12 ~         {
13 +             let _ = &42;
14 +             let foo: &Option<_> = &Some::<u8>(42);
15 +             if true {
16 +                 break;
17 +             } else {
18 +                 continue;
19 +             }
20 +         }
   |
```

### Version

```text
rustc 1.67.0-nightly (c97b539e4 2022-11-30)
binary: rustc
commit-hash: c97b539e408ea353f4fde2f9251d598291fec421
commit-date: 2022-11-30
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

tamaroning:
closes #10018

---

changelog: [`single_element_loop`]: No longer lints, if the loop contains a `break` or `continue`
[#10162](https://github.com/rust-lang/rust-clippy/pull/10162)

### DIFF
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -1,7 +1,7 @@
 use super::SINGLE_ELEMENT_LOOP;
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::{indent_of, snippet_with_applicability};
-use clippy_utils::visitors::for_each_expr;
+use clippy_utils::visitors::contains_break_or_continue;
 use if_chain::if_chain;
 use rustc_ast::util::parser::PREC_PREFIX;
 use rustc_ast::Mutability;
@@ -9,18 +9,6 @@ use rustc_errors::Applicability;
 use rustc_hir::{is_range_literal, BorrowKind, Expr, ExprKind, Pat};
 use rustc_lint::LateContext;
 use rustc_span::edition::Edition;
-use std::ops::ControlFlow;
-
-fn contains_break_or_continue(expr: &Expr<'_>) -> bool {
-    for_each_expr(expr, |e| {
-        if matches!(e.kind, ExprKind::Break(..) | ExprKind::Continue(..)) {
-            ControlFlow::Break(())
-        } else {
-            ControlFlow::Continue(())
-        }
-    })
-    .is_some()
-}
 
 pub(super) fn check<'tcx>(
     cx: &LateContext<'tcx>,
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -1,6 +1,7 @@
 use super::SINGLE_ELEMENT_LOOP;
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::{indent_of, snippet_with_applicability};
+use clippy_utils::visitors::for_each_expr;
 use if_chain::if_chain;
 use rustc_ast::util::parser::PREC_PREFIX;
 use rustc_ast::Mutability;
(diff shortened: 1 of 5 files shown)


clippy_lints/src/loops/single_element_loop.rs
```diff
@@ -1,13 +1,26 @@
...
 use if_chain::if_chain;
 use rustc_ast::util::parser::PREC_PREFIX;
 use rustc_ast::Mutability;
 use rustc_errors::Applicability;
 use rustc_hir::{is_range_literal, BorrowKind, Expr, ExprKind, Pat};
 use rustc_lint::LateContext;
 use rustc_span::edition::Edition;
+use std::ops::ControlFlow;
+
+fn contains_break_or_continue(expr: &Expr<'_>) -> bool {
```
xFrednet, line 14: This looks useful in general, would you mind moving this into `clippy_utils::visitors`?

xFrednet (approved):
LGTM, I only have one small suggestion, then we can merge this.

Side note: I wonder if it would make more sense, to simply change the applicability and point out, that the `break` and `continue`statement should be refactored. However, that seems like a lot of effort for little benefit. Some users might even use this pattern to structure their control flow (Even if that might not be ideal). Therefore, I think we should keep this implementation :)

tamaroning:
@xFrednet thank you for your review. I've finished fixing code.

xFrednet:
You're welcome, thank you for the swift response. This looks good to me.

@bors r+

Issue #9500, opened 2022-09-20
Unknown lint `arithmetic_side_effects` (`rustc` warning)


Rudxain:
### Description

I'm not sure if I should report this issue here or at `/rust-lang/rust/`. I was reading the docs, and found `clippy::arithmetic_side_effects`, set it to level `warn`, and `rustc` says that it can't recognize it. Why does this happen?

### Version

```text
rustc 1.63.0 (4b91a6ea7 2022-08-08)
binary: rustc
commit-hash: 4b91a6ea7258a947e59c6522cd5898e7c0a6a88f
commit-date: 2022-08-08
host: x86_64-unknown-linux-gnu
release: 1.63.0
LLVM version: 14.0.5
```

kraktus:
`arithmetic_side_effects` was added in [rust 1.64](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone), I assume this is the issue, but a reproducible example would be needed.

Rudxain:
> `arithmetic_side_effects` was added in [rust 1.64](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone), I assume this is the issue, but a reproducible example would be needed.

Oh right, I'm gonna close the issue then. Sorry for the inconvenience.

BTW, I ran `rustup update && rustup upgrade`, but it didn't update the version. Am I doing something wrong?

Rudxain:
I've updated to `rustc` 1.64, yet I still get the warning, even after `cargo clean`

kraktus:
What is your clippy version?

Rudxain:
Output of `cargo clippy -V`
```
clippy 0.1.64 (a55dd71 2022-09-19)
```

KizzyCode:
Can confirm – same versions as OP, also running VSCode with `rust-analyzer`. I'm on macOS if it makes any difference.

jmillikin:
I notice the lint was renamed from `arithmetic` to `arithmetic_side_effects` on September 9th (https://github.com/rust-lang/rust-clippy/commit/51d8b6c6643abb048987bd0befdeb6f46db0f6b5). It looks like Clippy v0.1.64 contains the old name:

```
// works
#![warn(clippy::arithmetic)]

// error: unknown lint: `clippy::arithmetic_side_effects`
#![warn(clippy::arithmetic_side_effects)]
```

Seems like a documentation bug that the name is updated immediately even if it stabilized under a different name.

PR #10170, opened 2023-01-06
[arithmetic_side_effects] Add more tests related to custom types

c410-f3r:
Add tests to ensure that custom types are triggered with any type of arithmetic operation as well as combinations with or without references.

---

changelog: none

### DIFF
--- a/tests/ui/arithmetic_side_effects.rs
+++ b/tests/ui/arithmetic_side_effects.rs
@@ -13,30 +13,94 @@
 
 use core::num::{Saturating, Wrapping};
 
+#[derive(Clone, Copy)]
 pub struct Custom;
 
 macro_rules! impl_arith {
-    ( $( $_trait:ident, $ty:ty, $method:ident; )* ) => {
+    ( $( $_trait:ident, $lhs:ty, $rhs:ty, $method:ident; )* ) => {
         $(
-            impl core::ops::$_trait<$ty> for Custom {
-                type Output = Self;
-                fn $method(self, _: $ty) -> Self::Output { Self }
+            impl core::ops::$_trait<$lhs> for $rhs {
+                type Output = Custom;
+                fn $method(self, _: $lhs) -> Self::Output { todo!() }
+            }
+        )*
+    }
+}
+
+macro_rules! impl_assign_arith {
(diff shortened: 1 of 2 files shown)


c410-f3r:
Can you guys close https://github.com/rust-lang/rust-clippy/issues/9500?

xFrednet:
Hey, I'm a bit confused by the changelog entry. Could you explain what you mean with the "mention" part? The `.stderr` files look pretty much the same for me, for normal integers and custom types, or have I missed something? :)

c410-f3r:
> Hey, I'm a bit confused by the changelog entry. Could you explain what you mean with the "mention" part? The `.stderr` files look pretty much the same for me, for normal integers and custom types, or have I missed something? :)

Changelog has been modified.

The `.stderr` file now has more than 300 LOC related to the newly included tests that deal with custom types and their arithmetic/unary operations with or without references.

clippy_lints/src/operators/mod.rs
```diff
@@ -116,6 +116,11 @@ declare_clippy_lint! {
     /// # let a = 0;
     /// a + 1;
     /// ```
+    ///
+    /// ### Noteworthy
+    ///
+    /// The `arithmetic_side_effects` lint is smarter with integer types, can deal with more corner
+    /// cases and is suitable for both numeric and custom types.
```
xFrednet, line 123: Adding a reference ca be good, though based on this description, I'm a bit confused, why a user should pick this lint over the `arithmetic_side_effects`. It might be good to make the difference clearer, or is `arithmetic_side_effects` just better all around?

c410-f3r: There are a bunch of other reasons, in fact, `arithmetic_side_effects` was initially created to replace `integer_arithmetic`.

I just don't have the energy or motivation to search all my previous contributions to compile a list so I will simply remove the mention.

xFrednet: I'm fine with being it mentioned, you could have a simple: `See also` in both lints that link to the other version. Just the way it was described there was a bit confusing. Removing it completely is also fine with me :).

I'd probably set the changelog entry to none in that case, since this PR has no user-facing changes, even though it's always awesome to have more tests :)

dswij (approved):
Thanks for taking the time to improve the tests!

PR #10174, opened 2023-01-07
hotfix: remove `ITER_COUNT` since it is not called

chansuke:
`&paths:ITER_COUNT` is not being called from anywhere, so removing it doesn't seem to be a problem.

---

changelog: none


dswij (approved):
Thanks for this!

PR #8945, opened 2022-06-04
improve [`almost_swapped`] to look for let statement

DevAccentor:
fix  #8151

improve [`almost_swapped`] to also detects code using let statements like
```rust
let b = 1;
let a = b;
let b = a;

let mut b = 1;
let a = b;
b = a;
```

changelog: improve [`almost_swapped`] to look for code using let statement

PR #10177, opened 2023-01-08
Almost swapped

chansuke:
Take over from https://github.com/rust-lang/rust-clippy/pull/8945

Fix https://github.com/rust-lang/rust-clippy/issues/8151

---

changelog: enhancement: [`almost_swapped`]: Now detects almost swaps using `let` statements 
[#10177](https://github.com/rust-lang/rust-clippy/pull/10177)


clippy_lints/src/swap.rs
```diff
@@ -172,15 +172,16 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
         if_chain! {
-            if let StmtKind::Semi(first) = w[0].kind;
-            if let StmtKind::Semi(second) = w[1].kind;
-            if first.span.ctxt() == second.span.ctxt();
-            if let ExprKind::Assign(lhs0, rhs0, _) = first.kind;
-            if let ExprKind::Assign(lhs1, rhs1, _) = second.kind;
+            if let StmtKind::Semi(first) = stmt[0].kind;
+            if let StmtKind::Semi(second) = stmt[1].kind;
+            if let Some((lhs0, rhs0)) = parse(&stmt[0]);
+            if let Some((lhs1, rhs1)) = parse(&stmt[1]);
```
Alexendoo, line 183: Here `parse` should be acting like https://github.com/rust-lang/rust-clippy/pull/8945#issuecomment-1192657937, where it checks for either a `let a = b` or `a = b`. That way we can get rid of https://github.com/rust-lang/rust-clippy/pull/10177/files#diff-342cbd825f201119c85fd8892ed86fb369d63b914c1e1e0179a40ef10215bd44R225-R295

```suggestion
            if let Some((lhs0, rhs0)) = parse(&stmt[0]);
            if let Some((lhs1, rhs1)) = parse(&stmt[1]);
```

clippy_lints/src/swap.rs
```diff
@@ -172,15 +172,16 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
     }
 }
 
+#[allow(clippy::too_many_lines)]
 /// Implementation of the `ALMOST_SWAPPED` lint.
 fn check_suspicious_swap(cx: &LateContext<'_>, block: &Block<'_>) {
-    for w in block.stmts.windows(2) {
+    for stmt in block.stmts.windows(2) {
```
Alexendoo, line 178: While we're here we could change this to use

```suggestion
    for [first, second] in block.stmts.array_windows() {
```

Alexendoo (commented):
Thanks for picking this up!

clippy_lints/src/swap.rs
chansuke: @Alexendoo Thank you so much for your review. I will fix this.

clippy_lints/src/swap.rs
```diff
@@ -172,12 +172,13 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
-    for w in block.stmts.windows(2) {
+    for [first_, second_] in block.stmts.array_windows() {
         if_chain! {
-            if let StmtKind::Semi(first) = w[0].kind;
-            if let StmtKind::Semi(second) = w[1].kind;
+            if let StmtKind::Semi(first) = first_.kind;
+            if let StmtKind::Semi(second) = second_.kind;
             if first.span.ctxt() == second.span.ctxt();
             if let ExprKind::Assign(lhs0, rhs0, _) = first.kind;
             if let ExprKind::Assign(lhs1, rhs1, _) = second.kind;
```
Alexendoo, line 184: Have another look at the comment https://github.com/rust-lang/rust-clippy/pull/8945#issuecomment-1192657937

We shouldn't be checking the `StmtKind` here, rather in a `parse` function that detects both cases. `ExprOrLocal` would be something along the lines of

```rust
enum ExprOrLocal<'a> {
    Expr(&'a Expr<'a>),
    Local(Ident),
}
```

clippy_lints/src/swap.rs
```diff
@@ -220,6 +221,92 @@ fn check_suspicious_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
             }
         }
+
+        if let StmtKind::Local(first) = first_.kind && let StmtKind::Local(second) = second_.kind {
+            parse_mutual(cx, first, second);
+        }
+
+        if let StmtKind::Local(first) = first_.kind && let StmtKind::Semi(second) = second_.kind {
+            parse(cx, first, second);
+        }
```
Alexendoo, line 231: And then we can remove this, there should be only one place where we're emitting the lint

```suggestion
```

chansuke: @Alexendoo I have read the comment and am stucking of refactoring the [L:238~L:293](https://github.com/rust-lang/rust-clippy/pull/8945/files#diff-342cbd825f201119c85fd8892ed86fb369d63b914c1e1e0179a40ef10215bd44R238-R293).How can I simplify these functions...
```
enum ExprOrLocal<'a> {
    Expr(&'a Expr<'a>),
    Local(&'a Local<'a>),
}
// Checking both statement type but probably I'm doing wrong 
fn parse<'a, 'hir>(first: &'a Stmt<'hir>, second: &'a Stmt<'hir>) -> Option<(&'a Local<'hir>, &'a ExprOrLocal<'hir>)> { // 
    if let StmtKind::Local(first_hs) = first.kind {
        if let StmtKind::Local(second_hs) = second.kind {
            if first_hs.span.ctxt() == second_hs.span.ctxt() {
                Some((first_hs, second_hs));
            }
        }
    } else if let StmtKind::Local(first_hs) = first.kind {
        if let StmtKind::Semi(second_hs) = second.kind {
            if first_hs.span.ctxt() == second_hs.span.ctxt() {
                Some((first_hs, second_hs));
            }
        }
    }
    None
}

if let Some((lhs1, rhs1)) = parse(first, second) {
  // Trying to make L:238 ~ L:293 DRY in here.
  // ref: https://github.com/rust-lang/rust-clippy/pull/8945/files#diff-342cbd825f201119c85fd8892ed86fb369d63b914c1e1e0179a40ef10215bd44R238-R293) DRY in here.
}
```

Alexendoo: `parse` should only take a single `Stmt` which you're breaking down into either `let {ident} = {expr}` or `{expr} = {expr}`

`ExprOrLocal::Local` should be an `Ident`, not `Local`, you could rename it to `ExprOrIdent` as well to make that clearer

You'd call it twice, once with the first statement and once with the second. Then you'd write a function

```rust
fn is_same(cx: .., lhs: ExprOrIdent, rhs: &Expr) -> bool { .. }
```

to call instead of `eq_expr_value` now that `lhs0` and `lhs1` are `ExprOrIdent`s instead of `Expr`s

chansuke: @Alexendoo Thank you so much for your advice. I have updated the code and trying to fix these changes now.
https://github.com/rust-lang/rust-clippy/blob/b2482d2db561b456661a910e46c10db45b0abcc5/tests/ui/crate_level_checks/no_std_swap.stderr#L6
https://github.com/rust-lang/rust-clippy/blob/b2482d2db561b456661a910e46c10db45b0abcc5/tests/ui/swap.fixed#L117

clippy_lints/src/swap.rs
```diff
@@ -172,15 +172,16 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
-            if let StmtKind::Semi(second) = w[1].kind;
+            if let StmtKind::Semi(first_semi) = first.kind;
+            if let StmtKind::Semi(second_semi) = second.kind;
             if first.span.ctxt() == second.span.ctxt();
-            if let ExprKind::Assign(lhs0, rhs0, _) = first.kind;
-            if let ExprKind::Assign(lhs1, rhs1, _) = second.kind;
+            if let ExprKind::Assign(lhs0, rhs0, _) = first_semi.kind;
+            if let ExprKind::Assign(lhs1, rhs1, _) = second_semi.kind;
             if eq_expr_value(cx, lhs0, rhs1);
             if eq_expr_value(cx, lhs1, rhs0);
```
Alexendoo, line 186: `parse`/`is_same`  should be replacing the `StmtKind/ExprKind` checks here, not an additional `if`

chansuke: @Alexendoo Replaced these logic, thanks.

clippy_lints/src/swap.rs
```diff
@@ -172,55 +172,167 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
     }
 }
 
+#[allow(clippy::too_many_lines)]
```
Alexendoo, line 175:
```suggestion
```

clippy_lints/src/swap.rs
```diff
@@ -172,55 +172,167 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
+    if let StmtKind::Local(_) = stmt_l.kind && let StmtKind::Local(_) = stmt_r.kind {
+        true
+    } else if let StmtKind::Local(_) = stmt_l.kind && let StmtKind::Semi(_) = stmt_r.kind {
+        true
+    } else {
+        false
+    }
+}
+
+fn is_same<'a, 'hir>(cx: &LateContext<'_>, lhs: &'a ExprOrIdent<'hir>, rhs: &'a Expr<'hir>) -> bool {
```
Alexendoo, line 242: We should be able to simplify this a good bit, something like

```rust
match lhs {
    ExprOrIdent::Expr(expr) => eq_expr_value(cx, expr, rhs)
    ExprOrIdent::Ident(ident) => ...
}
```

Where the `...` checks if `rhs` is a path with a single `PathSegment` in `.segments` and compares that against `ident` rather than joining with `::` (an ident can only ever be a single segment)

Alexendoo: Then you can add a method `span` on `ExprOrIdent` to add the `ctxt` equality check back above https://github.com/rust-lang/rust-clippy/pull/10177/files#diff-342cbd825f201119c85fd8892ed86fb369d63b914c1e1e0179a40ef10215bd44R181 rather than as part of this function

clippy_lints/src/swap.rs
```diff
@@ -174,53 +174,145 @@ fn check_manual_swap(cx: &LateContext<'_>, block: &Block<'_>) {
...
+}
 
-                let span = first.span.to(second.span);
-                let Some(sugg) = std_or_core(cx) else { return };
+fn is_same<'a, 'hir>(cx: &LateContext<'_>, lhs: &'a ExprOrIdent<'hir>, rhs: &'a Expr<'hir>) -> bool {
+    match lhs {
+        ExprOrIdent::Expr(expr) => eq_expr_value(cx, expr, rhs),
+        ExprOrIdent::Ident(ident) =>
+        if let ExprKind::Path(QPath::Resolved(None, path)) = rhs.kind &&
+            &path.segments.iter().map(|segment| segment.ident).collect::<Vec<_>>()[0] == ident {
```
Alexendoo, line 247: This isn't quite equivalent, it would say the path `foo::bar` is equal to `foo` as it's only checking the first segment, it should be something along the lines of `let [segment] = &path.segments && segment.ident == ident`

tests/ui/almost_swapped.rs
```diff
@@ -0,0 +1,32 @@
...
+    let b = a;
+    let a = t;
+
+    let mut b = 1;
+    let mut a = 2;
+
+    let t = b;
+    b = a;
+    a = t;
+}
```
Alexendoo, line 32: Let's stick these in the existing test file `swap.rs`, there might be some where the suggestions don't compile so you can add `mut` to make them compile or put them into a `swap_unfixable.rs`

Alternatively split out the existing tests from `swap.rs` into this `almost_swapped.rs` and create a `manual_swap.rs` for the rest

Alexendoo: There's still this one to cover ^

chansuke: @Alexendoo Sorry again. I moved these(L:26 ~ L:31) into `tests/ui/swap.rs` but I'm not certain this is correct.

Alexendoo: Ah I just clicked on the end of the file to comment, not meaning those couple lines specifically

I mean all the tests so there's one place `almost_swapped` is being tested, whether that's combining it into `swap.rs` or moving the existing `almost_swapped` tests out of `swap.rs` into `almost_swapped.rs` and renaming `swap.rs` to `manual_swap.rs`

chansuke: I appreciate your detailed exmplanation

tests/ui/swap.rs
```diff
@@ -5,9 +5,11 @@
     clippy::disallowed_names,
     clippy::no_effect,
     clippy::redundant_clone,
+    clippy::almost_swapped,
```
Alexendoo, line 8: We're testing for this, so we don't want to `#![allow]` it

Alexendoo: If there are some suggestions that don't compile, create a `swap_unfixable.rs` and put them in there instead

Alexendoo:
Okay great, this looks good thanks. Could you do a squash and then we'll get it merged

chansuke:
@Alexendoo Thank you so much for your patience. I've updated the branch.

Alexendoo (approved):

Alexendoo:
No worries, thanks again for the work

@bors r+

PR #10179, opened 2023-01-08
trim paths in `suspicious_to_owned`

llogiq:
This continues my path trimming spree. I'm not going to add yet another changelog entry, we should have one "trim paths in some applicable lints" entry instead.

---

changelog: none


llogiq:
Sorry! I appear to be doing too many things at once...

xFrednet (approved):

PR #10181, opened 2023-01-08
add a test against #10089

llogiq:
This was just an oversight while doing the box default path trimming PR, so I'm adding it rather late than never.

---

changelog: none


flip1995:
@bors r+

Thanks!

PR #10184, opened 2023-01-09
Allow implementing`Hash` with derived `PartialEq` (`derive_hash_xor_eq`

### LINKED ISSUE #2627: derive_hash_xor_eq deny deriving PartialEq when implementing Hash manually
According to [here](https://rust-lang-nursery.github.io/rust-clippy/v0.0.191/index.html#derive_hash_xor_eq)

derive_hash_xor_eq "Checks for deriving Hash but implementing PartialEq explicitly." however even doing the reverse is denied too. 

I have tried the minimal example below and still faced this problem:

```
use std::hash::{Hash, Hasher};

#[derive(PartialEq)]
struct Duck {}

impl Hash for Duck {
    fn hash<H: Hasher>(&self, _: &mut H) {}
}
fn main() {}
```
Where running clippy results in this:
```

error: you are implementing `Hash` explicitly but have derived `PartialEq`
 --> src/main.rs:6:1
  |
6 | / impl Hash for Duck {
7 | |     fn hash<H: Hasher>(&self, _: &mut H) {}
8 | | }
  | |_^
  |
  = note: #[deny(derive_hash_xor_eq)] on by default
note: `PartialEq` implemented here
 --> src/main.rs:3:10
  |
3 | #[derive(PartialEq)]
  |          ^^^^^^^^^
  = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.191/index.html#derive_hash_xor_eq

error: Could not compile `hashequal`.
```

I am using the latest rust nightly and the latest clippy. (mac os)

I have tried to look at the code and this seems intentional -well even the name seems to imply so- so maybe only the documentation should change?

robertbastian:
This is a common pattern and is totally allowed by the `Hash` trait.

Fixes #2627

changelog: [`derive_hash_xor_eq `]: allow implementing `Hash` with derived `PartialEq`

### DIFF
--- a/clippy_lints/src/derive.rs
+++ b/clippy_lints/src/derive.rs
@@ -243,7 +243,7 @@ fn check_hash_peq<'tcx>(
             cx.tcx.for_each_relevant_impl(peq_trait_def_id, ty, |impl_id| {
                 let peq_is_automatically_derived = cx.tcx.has_attr(impl_id, sym::automatically_derived);
 
-                if peq_is_automatically_derived == hash_is_automatically_derived {
+                if !hash_is_automatically_derived || peq_is_automatically_derived {
                     return;
                 }
 
@@ -252,17 +252,11 @@ fn check_hash_peq<'tcx>(
                 // Only care about `impl PartialEq<Foo> for Foo`
                 // For `impl PartialEq<B> for A, input_types is [A, B]
                 if trait_ref.substs.type_at(1) == ty {
-                    let mess = if peq_is_automatically_derived {
-                        "you are implementing `Hash` explicitly but have derived `PartialEq`"
-                    } else {
-                        "you are deriving `Hash` but have implemented `PartialEq` explicitly"
-                    };
-
                     span_lint_and_then(
                         cx,
                         DERIVE_HASH_XOR_EQ,
                         span,
-                        mess,
+                        "you are deriving `Hash` but have implemented `PartialEq` explicitly",
                         |diag| {
                             if let Some(local_def_id) = impl_id.as_local() {
                                 let hir_id = cx.tcx.hir().local_def_id_to_hir_id(local_def_id);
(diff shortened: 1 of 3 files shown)


Jarcho:
The lint name should be changed while you're at it. Something like `derived_hash_with_manual_eq`. `cargo dev rename_lint` can be used to make the change.

robertbastian:
Done

tests/ui/derived_hash_with_manual_eq.rs
```diff
@@ -34,23 +34,4 @@ impl std::hash::Hash for Bah {
...
-
-mod use_hash {
-    use std::hash::{Hash, Hasher};
-
-    #[derive(PartialEq)]
-    struct Foo3;
-
-    impl Hash for Foo3 {
-        fn hash<H: std::hash::Hasher>(&self, _: &mut H) {}
-    }
```
xFrednet, line 53: I think it would be better to keep this test code and add a comment explaining, why this is okay. Something simple like:

```
/// Implementing `Hash` with a derived `PartialEq` is fine. See #2627
```

tests/ui/rename.rs
```diff
@@ -10,6 +10,7 @@
 #![allow(clippy::box_collection)]
 #![allow(clippy::redundant_static_lifetimes)]
 #![allow(clippy::cognitive_complexity)]
+#![allow(clippy::derived_hash_with_manual_eq)]
```
xFrednet, line 13: These changes currently make the CI fail. You can automatically update the `.stderr` file by first running `cargo uitest` and den `cargo dev bless`. That should update everything

xFrednet (commented):
The logic looks good to me. Two small nits and then this should be ready for master.

@Jarcho Thank you for the suggestion to rename the lint, I probably would have missed that :)

tests/ui/derived_hash_with_manual_eq.rs
xFrednet: Ah, okay I misunderstood this part. Thank you :)

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update :)

@bors r+

PR #10187, opened 2023-01-10
[needless_return]: Remove all semicolons on suggestion 

### LINKED ISSUE #10182: clippy --fix fails for unnecessary return + unnecessary trailing semicolon
### Summary

Hi,
Thanks for your work on clippy.

It seems like clippy's fix for an unnecessary return doesn't account for trailing semicolons and fixes `return true;;` to `true;`, causing an error. I was able to reproduce it on stable, nightly and on master (commit `46981dca3622f5d53c88219f35113e553de20e1a`).

### Reproducer

Minimised:
```rust
// lib.rs
fn _breaks_clippy_fix() -> bool {
    return true;;
}
```
```
〉cargo clippy --fix
    Checking sandbox_rs v0.1.0 (/drive/ssd_ext4/Dev/sandbox/sandbox_rs)
warning: failed to automatically apply fixes suggested by rustc to crate `sandbox_rs`

  * src/lib.rs

 --> src/lib.rs:1:31
  |
1 | pub fn breaks_clippy_fix() -> bool {
  |        -----------------      ^^^^ expected `bool`, found `()`
  |        |
  |        implicitly returns `()` as its body has no tail or `return` expression
2 |     true;
  |         - help: remove this semicolon to return this value

warning: unnecessary trailing semicolon
 --> src/lib.rs:2:17
  |
2 |     return true;;
  |                 ^ help: remove this semicolon
  |
  = note: `#[warn(redundant_semicolons)]` on by default

 --> src/lib.rs:2:5
  |
2 |     return true;;
  |     ^^^^^^^^^^^
  |
  = help: remove `return`

warning: `sandbox_rs` (lib) generated 2 warnings
warning: `sandbox_rs` (lib test) generated 2 warnings (2 duplicates)
    Finished dev [unoptimized + debuginfo] target(s) in 0.35s
```

### Version

`rustc +stable --version --verbose`:
```
rustc 1.66.0 (69f9c33d7 2022-12-12)
binary: rustc
commit-hash: 69f9c33d71c871fc16ac445211281c6e7a340943
commit-date: 2022-12-12
release: 1.66.0
LLVM version: 15.0.2
```

`rustc +nightly --version --verbose`:
```
rustc 1.68.0-nightly (cc47b0699 2023-01-08)
binary: rustc
commit-hash: cc47b069983292e4ee8982d5dabe6301452c5f25
commit-date: 2023-01-08
release: 1.68.0-nightly
LLVM version: 15.0.6
```

dswij:
Closes #10182 

Multiple semicolons currently breaks autofix for `needless_return` suggestions. Any semicolons left after removing return means that the return type will always be `()`, and thus fail to compile.

This PR allows `needless_return` to remove multiple semicolons.

The change won't cover the case where there is multiple line yet.

i.e. 

```rust
fn needless_return() -> bool {
    return true;
;;
}
```

---

changelog: Sugg: [`needless_return`]: Now removes all semicolons on the same line
[#10187](https://github.com/rust-lang/rust-clippy/pull/10187)

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -1,7 +1,7 @@
 use clippy_utils::diagnostics::{span_lint_and_then, span_lint_hir_and_then};
 use clippy_utils::source::{snippet_opt, snippet_with_context};
 use clippy_utils::visitors::{for_each_expr, Descend};
-use clippy_utils::{fn_def_id, path_to_local_id};
+use clippy_utils::{fn_def_id, path_to_local_id, span_find_starting_semi};
 use core::ops::ControlFlow;
 use if_chain::if_chain;
 use rustc_errors::Applicability;
@@ -151,7 +151,7 @@ impl<'tcx> LateLintPass<'tcx> for Return {
         kind: FnKind<'tcx>,
         _: &'tcx FnDecl<'tcx>,
         body: &'tcx Body<'tcx>,
-        _: Span,
+        sp: Span,
         _: HirId,
     ) {
         match kind {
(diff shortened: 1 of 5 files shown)


Jarcho:
It would be better to use the lexer here. Start at the end of the statement, skip all whitespace/comment tokens, and take the offset of all the leading semicolons. You should also bail out of the lint if a `#` is found (e.g. `; #[cfg(...)] {}`).

dswij:
I specifically avoided the lexer because [it has a wide definition of whitespace](https://doc.rust-lang.org/reference/whitespace.html), e.g. line feed also falls into whitespace. 

I'm keeping this PR fairly simple so it won't cover those unnecessary colons spanning multiple lines. Also, rustc lint will fire for these redundant semicolons and rustfmt will fix this. This change is just to allow running clippy fix before all of that. I'd argue that even this scenario is fairly uncommon.

> You should also bail out of the lint if a # is found (e.g. ; #[cfg(...)] {}).

Maybe, yeah. I think this is out of the scope of this PR, but `unreachable_code` will fire here, and the user should address that first before considering clippy lints.

I also tested this out and seems like `needless_return` is already bailing out for these scenarios.

Jarcho:
> Also, rustc lint will fire for these redundant semicolons and rustfmt will fix this.

That's reasonable then. Given that the issue had come up at all I was assuming this wasn't the case.

@bors r+

PR #10199, opened 2023-01-13
Document lint configuration values in Clippy's book

### LINKED ISSUE #9991: Document lint configuration values in Clippy's book
We have a few lints with configuration values, like [borrow_as_ptr](https://rust-lang.github.io/rust-clippy/master/index.html#borrow_as_ptr). We want to increase the visibility of these config values, to encourage the use of them.

As part of this, we'd like a chapter about configuring lints and having a list of available configurations with links to [Clippy's lint list](https://rust-lang.github.io/rust-clippy/master/index.html).

Tasks:
* [x] Write how a user can add configurations: See [Current docs](https://github.com/rust-lang/rust-clippy#configuration)
* [x] Generate list of configurations. This can be done in Clippy's metadata collector. https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/utils/internal_lints/metadata_collector.rs
* [x] Include the generated list in the book
* [x] Be happy!

---

I'm open to mentoring this issue :)

CC: #9880

tylerjw:
changelog: document lint configuration values in Clippy's book

fixes #9991

r? @xFrednet


tylerjw:
@xFrednet I believe I'm ready for a first review.

README.md
```diff
@@ -194,12 +194,23 @@ value` mapping e.g.
...
 avoid-breaking-exported-api = false
 disallowed-names = ["toto", "tata", "titi"]
-cognitive-complexity-threshold = 30
 ```
 
 See the [list of configurable lints](https://rust-lang.github.io/rust-clippy/master/index.html#Configuration),
 the lint descriptions contain the names and meanings of these configuration variables.
 
+See [table of lint configurations](https://doc.rust-lang.org/nightly/clippy/lint_configuration.html)
+to see what configuration options you can set and the lints they configure.
```
xFrednet, line 203: I remember the first reference to Clippy's lint list was mainly added to make people more aware. The documentation added by this PR is better for that. Maybe we can join these two together, saying something like this:

The [table of configurations](https://doc.rust-lang.org/nightly/clippy/lint_configuration.html) contains all config values, their default, and a list of lints they affect. Each [configurable lint](https://rust-lang.github.io/rust-clippy/master/index.html#Configuration), also contains information about these values.

```
The [table of configurations](https://doc.rust-lang.org/nightly/clippy/lint_configuration.html)
contains all config values, their default, and a list of lints they affect.
Each [configurable lint](https://rust-lang.github.io/rust-clippy/master/index.html#Configuration)
, also contains information about these values.
```

book/src/lint_configuration.md
```diff
@@ -0,0 +1,52 @@
...
+| enforced-import-renames | `[]` | The list of imports to always rename, a fully qualified path followed by the rename | [missing_enforced_import_renames](https://rust-lang.github.io/rust-clippy/master/index.html#missing_enforced_import_renames) |
+| allowed-scripts | `["Latin"]` | The list of unicode scripts allowed to be used in the scope | [disallowed_script_idents](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_script_idents) |
+| enable-raw-pointer-heuristic-for-send | `true` | Whether to apply the raw pointer heuristic to determine if a type is `Send` | [non_send_fields_in_send_ty](https://rust-lang.github.io/rust-clippy/master/index.html#non_send_fields_in_send_ty) |
+| max-suggested-slice-pattern-length | `3` | When Clippy suggests using a slice pattern, this is the maximum number of elements allowed in the slice pattern that is suggested | [index_refutable_slice](https://rust-lang.github.io/rust-clippy/master/index.html#index_refutable_slice) |
+| await-holding-invalid-types | `[]` | [ERROR] MALFORMED DOC COMMENT |  |
```
xFrednet, line 42: This is an interesting output, it might be that this is just a problem in the metadata collection and not the generating script itself.

.github/workflows/remark.yml
```diff
@@ -33,6 +33,9 @@ jobs:
         echo `pwd`/mdbook >> $GITHUB_PATH
 
     # Run
+    - name: cargo collect-metadata
```
xFrednet, line 36: Instead of calling `collect-metadata` here, we should add a check to see that the current table is the same as the output after running `cargo collect-metadata`. This could either be implemented by adding a check in the metadata collection, or running the collection and then checking with git if the file was actually changed.

I'm a little unsure which would be cleaner @flip1995, do you have a suggestion as our main CI wizard? :mage:

xFrednet (commented):
A solid start, thank you for the work! I've added some comments, and also gave some feedback about the table/format in the issue. I like the look of the table, but this might be a case where a different format is sadly more functional.

tylerjw:
I updated this to produce both a table and paragraph sections. The table portion looks like this:

image

And then if you were to click on any of those configs it would go to a table section that looks like this:

image

I tried making the table widths work for the names, default values, and lints but couldn't get it to look good, so I moved the lints into the paragraphs and made the table just the config names and their default values. Clicking on any one goes to the paragraph, which now includes not only the list of lints but also the config type and the whole doc string.

README.md
tylerjw: I made this change.

clippy_lints/src/utils/conf.rs
```diff
@@ -219,7 +219,8 @@ define_Conf! {
     ///
     /// #### Noteworthy
     ///
-    /// A type, say `SomeType`, listed in this configuration has the same behavior of `["SomeType" , "*"], ["*", "SomeType"]` in `arithmetic_side_effects_allowed_binary`.
+    /// A type, say `SomeType`, listed in this configuration has the same behavior of
+    /// `["SomeType" , "*"], ["*", "SomeType"]` in `arithmetic_side_effects_allowed_binary`.
```
tylerjw, line 223: The reason for this change is to fix a `> 120 characters line` lint error on the markdown output.

tylerjw:
@xFrednet this is ready for another review. I made some changes to how the generated doc output. Let me know if you have any ideas for how I could format it better and other things you notice that I could improve.

clippy_lints/src/utils/conf.rs
xFrednet: That's totally fine! The documentation should have been wrapped anyways. This is probably an artifact from the time, where the documentation was limited to a single line ^^

.github/workflows/remark.yml
xFrednet: I've done some testing and came up with a working solution:
1. I'd move this test into `clippy_bors.yml` as that script already runs the metadata collection. (This also avoids installing the rust toolchain in this workflow)
2. Add this new step under `Test metadata collection` around line 148
    ```yml
    - name: Test lint_configuration.md is up-to-date
      run: |
        echo "run \`cargo collect-metadata\` if this fails"
        git update-index --refresh
    ```
    This step will fail, if there are any changes after the collection.

	Credit where credit is due: the command has been *stolen* from [Stack overflow](https://stackoverflow.com/questions/3878624/how-do-i-programmatically-determine-if-there-are-uncommitted-changes) and was, of course, tested locally and in the CI by me :)

xFrednet (commented):
This version looks good to me! I have two more required adjustments, and then we can merge this.
1. Is the comment for the CI changes
2. Could you also update the [*add configuration to a lint*](https://doc.rust-lang.org/clippy/development/adding_lints.html#adding-configuration-to-a-lint) section to have a 5th step, to run `cargo collect-metadata`?

Then this should be ready! Thank you very much for this update!!!!

tylerjw:
@xFrednet I pushed the two changes you laid out. Thank you for going and figuring out the github actions magic I needed for this.

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -199,12 +219,27 @@ impl Drop for MetadataCollector {
...
+        // Outputting markdown
+        if Path::new(MARKDOWN_OUTPUT_FILE).exists() {
+            fs::remove_file(MARKDOWN_OUTPUT_FILE).unwrap();
+        }
+        let mut file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .open(MARKDOWN_OUTPUT_FILE)
+            .unwrap();
+        writeln!(file, "{}", self.get_markdown_docs(),).unwrap();
```
xFrednet, line 242: There is one last thing I just remembered. Could you add a comment to the `book/src/configuration.md` file, notifying potential editors, that the content is automatically generated? Something like this:

```md
<!--
This file is generated by `cargo collect-metadata`.
Please use that command to update the file and do not edit it by hand.
-->
```

Then this is really ready to be merged :)

tylerjw: Pushed this change.

xFrednet (approved):

xFrednet:
Everything looks perfect to me! Thank you very much. This is a very nice step forward for Clippy's configuration!

@bors r+

flip1995:
Awesome! Thanks for implementing this :)

@xfrednet Sorry for not replying to your CI question. But the solution you came up with is perfect 👍

xFrednet:
@flip1995 No problem at all, I would have waited if Google didn't turn up such a simple solution. :)

PR #10201, opened 2023-01-14
Add missing arguments to `new_lint` example of `cargo` lints

MariaSolOs:
changelog: Add missing arguments to `new_lint` code example.

From the surrounding text, it seems like this code example is missing the arguments to create a new `cargo` lint.


dswij (approved):
Thanks for this!

PR #10206, opened 2023-01-16
Add `multiple_unsafe_ops_per_block` lint

### LINKED ISSUE #10064: New lint: Limit unsafe blocks to a single unsafe operation
### What it does

Checks for `unsafe` blocks that contain more than one unsafe operation. Unsafe operations include:

- Dereferencing a raw pointer (`*const T` or `*mut T`)
- Calling an unsafe function or method
- Accessing or modifying a mutable static variable
- Accessing fields of `union`s

### Lint Name

`one_unsafe_op_per_unsafe_block`

### Category

restriction

### Advantage

Combined with `undocumented_unsafe_blocks`, this lint ensures that each unsafe operation must be independently justified. Combined with `unused_unsafe`, this lint also ensures that we know when we eliminate unsafe operations through refactoring.

See the "Example" section for comprehensive examples.

### Drawbacks

It may be awkward or noisy to break up a series of related unsafe operations into individual blocks.

## With `undocumented_unsafe_blocks`

```rust
/// Reads a `char` from the given pointer.
///
/// # Safety
///
/// `ptr` must point to four consecutive, initialized bytes which
/// form a valid `char` when interpreted in the native byte order.
unsafe fn read_char(ptr: *const u8) -> char {
    // SAFETY: The caller has guaranteed that the value pointed
    // to by `bytes` is a valid `char`.
    unsafe { char::from_u32_unchecked(*ptr.cast::<u32>()) }
}
```

Although the `unsafe` block is properly justified for `char::from_u32_unchecked`, there are actually two unsafe operations in it. The pointer dereference needs to be justified as well, but it's easy to accidentally include it without updating the justification. By limiting unsafe blocks to a single unsafe operation, this would be rewritten as:

```rust
/// Reads a `char` from the given pointer.
///
/// # Safety
///
/// - `ptr` must be 4-byte aligned, point to four consecutive
///   initialized bytes, and be valid for reads of 4 bytes.
/// - The bytes pointed to by `ptr` must represent a valid
///   `char` when interpreted in the native byte order.
unsafe fn read_char(ptr: *const u8) -> char {
    // SAFETY: `ptr` is 4-byte aligned, points to four consecutive
    // initialized bytes, and is valid for reads of 4 bytes.
    let int_value = unsafe { *ptr.cast::<u32>() };
    // SAFETY: The caller has guaranteed that the four bytes
    // pointed to by `bytes` represent a valid `char`.
    unsafe { char::from_u32_unchecked(int_value) }
}
```

Breaking up our unsafe block led us to justify dereferencing `ptr`, which in turn led us to update the function's safety requirements to accurately reflect the invariants used by the implementation.

## With `unused_unsafe`

Refactoring from the same starting function as the example in `undocumented_unsafe_blocks`:

```rust
/// Reads a `char` from the given reference.
///
/// # Safety
///
/// `int` must represent a valid `char` when interpreted in
/// the native byte order.
unsafe fn read_char(int: &u32) -> char {
    // SAFETY: The caller has guaranteed that the four bytes
    // pointed to by `bytes` represent a valid `char`.
    unsafe { char::from_u32_unchecked(*int) }
}
```

By changing our function to accept `&u32`, we were able to remove the only unsafe operation from our first unsafe block. In turn, we were able to eliminate the unsafe block, prompting us to update our safety requirements for our function. In this way, refactoring a function to remove unsafe helps ensure that the safety documentation also gets updated to reflect the new requirements.

Niki4tap:
Adds a lint, which restricts an `unsafe` block to only one unsafe operation.

Closes #10064

---

changelog: New lint: [`multiple_unsafe_ops_per_block`]
[#10206](https://github.com/rust-lang/rust-clippy/pull/10206)

### DIFF
--- a/clippy_lints/src/multiple_unsafe_ops_per_block.rs
+++ b/clippy_lints/src/multiple_unsafe_ops_per_block.rs
@@ -1,0 +1,185 @@
+use clippy_utils::{
+    diagnostics::span_lint_and_then,
+    visitors::{for_each_expr_with_closures, Descend, Visitable},
+};
+use core::ops::ControlFlow::Continue;
+use hir::{
+    def::{DefKind, Res},
+    BlockCheckMode, ExprKind, QPath, UnOp, Unsafety,
+};
+use rustc_ast::Mutability;
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `unsafe` blocks that contain more than one unsafe operation.
+    ///
+    /// ### Why is this bad?
+    /// Combined with `undocumented_unsafe_blocks`,
+    /// this lint ensures that each unsafe operation must be independently justified.
+    /// Combined with `unused_unsafe`, this lint also ensures
+    /// elimination of unnecessary unsafe blocks through refactoring.
+    ///
+    /// ### Example
+    /// ```rust
+    /// /// Reads a `char` from the given pointer.
+    /// ///
+    /// /// # Safety
+    /// ///
+    /// /// `ptr` must point to four consecutive, initialized bytes which
+    /// /// form a valid `char` when interpreted in the native byte order.
+    /// fn read_char(ptr: *const u8) -> char {
+    ///     // SAFETY: The caller has guaranteed that the value pointed
+    ///     // to by `bytes` is a valid `char`.
+    ///     unsafe { char::from_u32_unchecked(*ptr.cast::<u32>()) }
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// /// Reads a `char` from the given pointer.
+    /// ///
+    /// /// # Safety
+    /// ///
+    /// /// - `ptr` must be 4-byte aligned, point to four consecutive
+    /// ///   initialized bytes, and be valid for reads of 4 bytes.
+    /// /// - The bytes pointed to by `ptr` must represent a valid
+    /// ///   `char` when interpreted in the native byte order.
+    /// fn read_char(ptr: *const u8) -> char {
+    ///     // SAFETY: `ptr` is 4-byte aligned, points to four consecutive
+    ///     // initialized bytes, and is valid for reads of 4 bytes.
+    ///     let int_value = unsafe { *ptr.cast::<u32>() };
+    ///
+    ///     // SAFETY: The caller has guaranteed that the four bytes
+    ///     // pointed to by `bytes` represent a valid `char`.
+    ///     unsafe { char::from_u32_unchecked(int_value) }
+    /// }
+    /// ```
+    #[clippy::version = "1.68.0"]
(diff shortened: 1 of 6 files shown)


clippy_lints/src/one_unsafe_op_per_block.rs
```diff
@@ -0,0 +1,252 @@
...
+        }
+    }
+}
+
+// This whole function is practically one big `match` block
+#[expect(
+    clippy::too_many_lines,
+    reason = "long match means long function, not much you can do here"
+)]
+fn collect_unsafe_exprs(cx: &LateContext<'_>, expr: &hir::Expr<'_>, unsafe_ops: &mut Vec<(&'static str, Span)>) {
```
Alexendoo, line 113: This could use [`for_each_expr_with_closures`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/visitors/fn.for_each_expr_with_closures.html) (or a struct that impls [`Visitor`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/intravisit/trait.Visitor.html)) rather than calling itself recursively to be sure we're catching all the expressions

clippy_lints/src/one_unsafe_op_per_block.rs
```diff
@@ -0,0 +1,252 @@
...
+    ///     // initialized bytes, and is valid for reads of 4 bytes.
+    ///     let int_value = unsafe { *ptr.cast::<u32>() };
+    ///
+    ///     // SAFETY: The caller has guaranteed that the four bytes
+    ///     // pointed to by `bytes` represent a valid `char`.
+    ///     unsafe { char::from_u32_unchecked(int_value) }
+    /// }
+    /// ```
+    #[clippy::version = "1.68.0"]
+    pub ONE_UNSAFE_OP_PER_BLOCK,
```
Alexendoo, line 57: The lint name should be `multiple_unsafe_ops_per_block` or something along those lines, to reflect the thing it's detecting rather than the thing it's suggesting

Alexendoo (commented):
:wave:

Thanks for the PR! It's looking good

Niki4tap:
Thanks for the review, @Alexendoo! I've addressed the comments, can you review it again please?

clippy_lints/src/multiple_unsafe_ops_per_block.rs
```diff
@@ -0,0 +1,210 @@
...
+}
+
+fn collect_unsafe_exprs<'tcx>(
+    cx: &LateContext<'tcx>,
+    expr: &'tcx hir::Expr<'_>,
+    unsafe_ops: &mut Vec<(&'static str, Span)>,
+) {
+    let expr = expr.peel_drop_temps();
+
+    for_each_expr_with_closures(cx, expr, |expr| {
```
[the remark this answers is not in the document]
Niki4tap: Can't take `Block` directly here, as I still have one recursive call at the bottom, but I figured I can take `impl Visitable`, so that works. Thanks for the comment!

Niki4tap:
@Alexendoo can you look again please? I think it should be good now

Alexendoo (approved):

PR #10212, opened 2023-01-18
Lint `zero_ptr` in const contexts

Niki4tap:
Lint `zero_ptr` in const contexts, since `core::ptr::null` is now const stabilized.

This PR doesn't affect `cmp_nan` lint (aside from `fixme` comment) that was disabled as well, because [`{f32,f64}::is_nan`](https://doc.rust-lang.org/stable/std/primitive.f32.html#method.is_nan) is still unstable in const contexts, so this is essentially blocked on https://github.com/rust-lang/rust/issues/57241

---

changelog: Enhancement: [`zero_ptr`]: Now lints in const contexts
[#10212](https://github.com/rust-lang/rust-clippy/pull/10212)


Niki4tap:
Giving it another look, the fixme comment about `cmp_nan` doesn't *really* make sense, since the lint doesn't give you an automatically-applicable suggestion, and the note about `{f32,f64}::is_nan` isn't even correct in cases of `NAN > some_float` and other comparison operations, maybe the lint message should be reworked. I would be glad to do it if this is desirable.

llogiq:
Why did you remove the `crashes` test?

Niki4tap:
I think the test is mostly commented out and wasn't quite working anyways, and what was left of it was testing the behavior of `zero_ptr` and `cmp_nan` lints in const contexts (not sure why these checks are in that test in the first place), which this PR changes.

`cmp_nan` checks were moved to `tests/ui/cmp_nan.rs`.

llogiq:
Ok, looks good otherwise. We'll have to revisit once the Rust stabilization PR is merged.

@rustbot blocked

xFrednet:
It looks like the feature is still blocking this PR and there is nothing indicating that it will be stabilized soon. I'll close this PR for now. Once the feature is stabilized, we can open a new PR to add this behavior.

@Niki4tap thank you for the time you put into this!

PR #10218, opened 2023-01-21
Add machine applicable suggestion for `bool_assert_comparison`

### LINKED ISSUE #7598: `bool_assert_comparison` improve the suggestion to include the expression

### LINKED ISSUE #8118: `bool_assert_comparison`: assert_eq! with a literal bool
### Description

Hi,

i have a macro which generated a function for use in unit tests which is:
```
assert_eq!($MacroInput, my_struct.macro_input());
```

Clippy indicates in this case that i should use `assert!` but this would mean that i have to write an if/else block to handle both cases. 
While this makes sense if i have a hardcoded boolean literal, but in this case i have macro generated boolean literals.

### Version

```text
binary: rustc
commit-date: 2021-11-29
release: 1.57.0
LLVM version: 13.0.0
```

### Additional Labels

_No response_

Alexendoo:
Fixes #7598
Also closes #8118, it had already been fixed by an earlier change but I've added a test for it

changelog: [`bool_assert_comparison`] The suggestion is now machine applicable

### DIFF
--- a/clippy_lints/src/bool_assert_comparison.rs
+++ b/clippy_lints/src/bool_assert_comparison.rs
@@ -1,10 +1,11 @@
+use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::macros::{find_assert_eq_args, root_macro_call_first_node};
-use clippy_utils::{diagnostics::span_lint_and_sugg, ty::implements_trait};
+use clippy_utils::ty::{implements_trait, is_copy};
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, Lit};
-use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::ty;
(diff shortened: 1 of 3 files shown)


dswij (approved):
Nice, LGTM. Thanks for this!

PR #10223, opened 2023-01-23
Add `transmute_tuple_to_slice`

### LINKED ISSUE #3994: [lint request] unnecessary_slice_transmute
Some people transmute a `(T* [const, mut], usize)` into a `&[T]`/`&mut [T]`/`* const[T]`/`*mut [T]` unnecessarily and vice-versa, instead of using the `from_raw_parts` methods and or `(slice.as_ptr(), slice.len())` (yeah, really). 

We should detect this and tell people to use the proper methods instead. We probably want to detect the incorrect `(usize, T* [const, mut])` cases as well, and tell people to also use the correct methods instead. 

If people are doing this inside a macro... we should warn anyways.

robertbastian:
Fixes #3994

---

changelog: New lint [`transmute_tuple_to_slice`]
[#10223](https://github.com/rust-lang/rust-clippy/pull/10223)


robertbastian:
CI is telling me to run `cargo dev update_lints` but doing that doesn't give me a diff.

Jarcho:
This already exists as part of `transmute_undefined_repr`. Currently in the nursery.

For CI that can be fixed by rebasing onto the latest master.

robertbastian:
This one suggests an alternative though, and is way more targeted. IIUC the other is in the nursery because it has a lot of false positives.

Jarcho:
The false positives on it are basically fixed. I think there's only one left with a simple fix. The suggestion to use the correct method when one exists should definitely be added to that lint though.

robertbastian:
I'll close this as I don't have the time to put this into `transmute_undefined_repr`.

PR #10229, opened 2023-01-26
Document `cargo-clippy` feature

### LINKED ISSUE #10220: Ability to skip files or blocks entirely
### Description

Some [experimental code of mine](https://github.com/danielparks/iter-matcher) generates a 48,000+ line function[^1] that takes 20+ minutes to process with `Clippy` with all lints allowed.

[^1]: The code is actually used in a private branch on [another crate](https://github.com/danielparks/htmlize), but there is [an example](https://github.com/danielparks/iter-matcher/blob/main/examples/html-entities.rs) that generates the function. This is all a bit horrific, and I’m still considering just shelving the idea, but it’s much faster than anything else I’ve tried.

It would be really helpful to be able to skip or ignore files or blocks of code entirely.

This seems like a fairly niche use-case, so perhaps this isn’t worth pursuing.

### Discarded idea: using a feature

I considered adding a `clippy` feature, but this is a bit inconvenient. I usually run `cargo clippy --all-features` so hiding the code if `feature = "clippy"` works well, but I also run `cargo test --all-features` so I need to change one or the other. This is additionally complicated by needing to configure CI correctly.

These are not insurmountable problems at all, but they do make it harder, especially for new people who want to contribute to my project.

### Proposal: setting `--cfg clippy` when Clippy runs

Code that you don’t want Clippy to look at could be guarded with `#[cfg(not(clippy))]`.

I haven’t dug into the Clippy code yet, but this seems like it might be relatively easy to implement (I’m happy to volunteer).

I’m not sure if Clippy would need an option to disable the configuration option so that you can force it to look at all code. For my usage, I would be happy to just remove the `#[cfg(not(clippy))]` line and run clippy again, but I could imagine that some people would want to run Clippy on some code only some times.

I suppose another solution would be for `cargo clippy` to accept `--cfg` options.

### Version

```text
rustc 1.66.1 (90743e729 2023-01-10)
binary: rustc
commit-hash: 90743e7298aca107ddaa0c202a4d3604e29bfeb6
commit-date: 2023-01-10
host: x86_64-apple-darwin
release: 1.66.1
LLVM version: 15.0.2
```

### Additional Labels

@rustbot label +C-enhancement

danielparks:
It is possible to use conditional compilation to prevent Clippy from evaluating certain code at all. Unfortunately, it was no longer documented anywhere. This adds a brief explanation of how to use the feature with conditional compilation, and mentions a few downsides.

Fixes #10220 — Ability to skip files or blocks entirely
changelog: none

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -100,3 +100,24 @@ Note: `custom_inner_attributes` is an unstable feature so it has to be enabled e
 
 Lints that recognize this configuration option can be
 found [here](https://rust-lang.github.io/rust-clippy/master/index.html#msrv)
+
+### Disabling evaluation of certain code
+
+> **Note:** This should only be used in cases where other solutions, like `#[allow(clippy::all)]`, are not sufficient.
+
+Very rarely, you may wish to prevent Clippy from evaluating certain sections of code entirely. You can do this with
+[conditional compilation](https://doc.rust-lang.org/reference/conditional-compilation.html) by checking that the
+`cargo-clippy` feature is not set. You may need to provide a stub so that the code compiles:
+
+```rust
+#[cfg(not(feature = "cargo-clippy"))]
+include!(concat!(env!("OUT_DIR"), "/my_big_function-generated.rs"));
+
+#[cfg(feature = "cargo-clippy")]
+fn my_big_function(_input: &str) -> Option<MyStruct> {
+    None
+}
+```
+
+This feature is not actually part of your crate, so specifying `--all-features` to other tools, e.g. `cargo test
+--all-features`, will not disable it.


danielparks:
I’m not entirely satisfied with this, but I wanted to get something out.

Note that the file seems to be wrapped at 120 characters instead of 80 like other book source files. I decided to match the file.

flip1995 (commented):
I would argue that this isn't the right approach. If sections should be ignored by Clippy, you can just add `#[allow(clippy::all)]`. (You may have to other lints or lint groups to that list in order to make it work fully). 

I think we only still support the `cargo-clippy` feature, because of backwards compat, which would also explain why it was removed from the documentation.

danielparks:
This is useful a few different situations that arise when working with generated code (particularly code generated by a tool you didn’t write):

1. The code is slow to evaluate, or otherwise causes problems for Clippy. For example, in #10220 I generated a function that takes stable Clippy 20+ minutes to evaluate when all lints are allowed. Perhaps that will be fixed, but generated code in particular seems likely to trigger edge cases.
2. You want to ensure that Clippy will never lint certain code, even if future groups beyond `all`, `pedantic`, etc. are added.
3. You want to prevent Clippy from linting code that contains attributes with Clippy directives. (This seems unlikely to be an issue to me.)

Mostly I care about number 1, since that’s what I ran into.

Perhaps this is a rare enough need that it’s not worth documenting. It’s possible that #10220 is enough to help somebody who has a similar problem in the future — I spent quite some time searching for a way to do this before I filed that issue.

If you feel that this shouldn’t be documented, please go ahead and close. Otherwise, I’m happy to adjust the documentation, or move it somewhere else, or whatever you think is best.

flip1995:
1. 20+ minutes is definitely concerning. Is the code that causes this openly available. It might be worth doing a perf run with Clippy and try to find out what lint(s) cause this high runtime and maybe fix them. If there are only a few lints causing this, we can just add a check if the lint is allowed and skip the actual lint check for those _expensive lints_.
2. That isn't a real concern. All warn-by-default lints will always be in the `all` group. It is really unlikely that a standalone allow-by-default group gets added. And even if that should happen, everyone first has to enable it for them to also having to disable it again.
3. I don't really know what you mean by that. Why would you put Clippy attributes into code that you then exclude by a feature.

---

I think you convinced me that documenting this is worth it. However I think we shouldn't advertise it so prominently. I would put it at the end of the file you've put it in and add a note that this should only be needed in really special cases. You can add a note to a MD book with markdown quotes:

```markdown
> Text of the note goes here
```

danielparks:
Done. I shortened the copy a little more, but I thought it was probably best to keep the example. That said, I’m happy to change this however you like.

flip1995 (approved):
Thanks! I really like the wording now. This totally slipped through, I have way too many GH notifications...

PR #10231, opened 2023-01-26
`invalid_regex`: Show full error when string value doesn't match source

### LINKED ISSUE #4170: Unrecognized escape sequence
Clippy version: `clippy 0.0.212 (265318d 2019-05-17)`

I get the following error when linting code, and I have a feeling it might be a bug. If it is not, it maybe clippy can give a correct suggestion?

```
error: regex syntax error: unrecognized escape sequence
   --> src/compiler/utilities.rs:187:52
    |
187 |         static ref RE: Regex = Regex::new("^..?($|[\\/])").unwrap();
    |                                                    ^^
    |
    = note: #[deny(clippy::invalid_regex)] on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#invalid_regex
```

Alexendoo:
changelog: [`invalid_regex`]: Show full error when parsing non-literals or regular strings containing escape sequences

Fixes #4170, the escape sequence there causes the span to be incorrect which will have caused most of the confusion

### DIFF
--- a/tests/ui/regex.stderr
+++ b/tests/ui/regex.stderr
@@ -29,7 +29,10 @@ error: regex syntax error: invalid character class range, the start must be <= t
 LL |     let some_unicode = Regex::new("[é-è]");
    |                                     ^^^
 
-error: regex syntax error on position 0: unclosed group
+error: regex parse error:
+           (
+           ^
+       error: unclosed group
   --> $DIR/regex.rs:18:33
    |
 LL |     let some_regex = Regex::new(OPENING_PAREN);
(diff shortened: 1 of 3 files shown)


tests/ui/regex.stderr
```diff
@@ -79,93 +94,110 @@ error: regex syntax error: unrecognized escape sequence
...
+error: regex parse error:
+           /b/c
+             ^^
+       error: unrecognized escape sequence
+  --> $DIR/regex.rs:40:42
+   |
+LL |     let escaped_string_span = Regex::new("/b/c");
+   |                                          ^^^^^^^^
+   |
+   = help: consider using a raw string literal: `r".."`
```
Alexendoo, line 106: compiletest is doing its windows path normalisation thing, it looks like this normally:

```
error: regex parse error:
           \b\c
             ^^
       error: unrecognized escape sequence
 --> ../t.rs:6:42
  |
6 |     let escaped_string_span = Regex::new("\\b\\c");
  |                                          ^^^^^^^^
  |
  = help: consider using a raw string literal: `r".."`
```

tests/ui/regex.stderr
```diff
@@ -29,7 +29,10 @@ error: regex syntax error: invalid character class range, the start must be <= t
 LL |     let some_unicode = Regex::new("[é-è]");
    |                                     ^^^
 
-error: regex syntax error on position 0: unclosed group
+error: regex parse error:
+           (
```
llogiq, line 33: Could we extend the span to cover the quotation marks? The way it looks now throws me off a bit.

llogiq (commented):
Apart from the missing quotation marks when linting a `const`, this looks good to me. Let me know if you want to tackle this as part of this PR or if you want to leave it for a followup.

tests/ui/regex.stderr
Alexendoo: I don't think we can, it's the error as printed by the regex crate which doesn't know of the source representation

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0193cd9f840f62b49211c23e647ae5e0

I'd say it looks slightly better in a terminal where it's grouped together by being bolded, but yeah the formatting is a little out of place

image

llogiq: I see. Ok then.

PR #10237, opened 2023-01-27
Fix styling in documentation for `suspicious_xor_used_as_pow` lint

cstyles:
There was a tab after the three leading slashes which caused the contents of the "Why is this bad?" section to be rendered as a code block.

**Before:**

master

**After:**

fixed

The file still contains a lot of tabs but they don't affect the documentation.

---

changelog: [`suspicious_xor_used_as_pow`]: Fix styling in documentation


Jarcho:
Thanks for the catch. @bors r+

PR #10249, opened 2023-01-28
`doc_markdown` lint: Whitelist names of popular frame servers

Enyium:
changelog: none

Spelling proof:

- https://avs-plus.net/
- https://www.vapoursynth.com/

Popularity proof:

- https://forum.doom9.org/#collapseimg_forumbit_67
- https://en.wikipedia.org/wiki/Doom9

Must the names also be added to these locations?

- https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/doc/doc-fixable.rs#L55
- https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/doc/doc-fixable.fixed#L55

### DIFF
--- a/clippy_lints/src/utils/conf.rs
+++ b/clippy_lints/src/utils/conf.rs
@@ -29,6 +29,7 @@ const DEFAULT_DOC_VALID_IDENTS: &[&str] = &[
     "TeX", "LaTeX", "BibTeX", "BibLaTeX",
     "MinGW",
     "CamelCase",
+    "AviSynth", "VapourSynth",
 ];
 const DEFAULT_DISALLOWED_NAMES: &[&str] = &["foo", "baz", "quux"];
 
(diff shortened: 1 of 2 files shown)


giraffate:
Oh, I'm not familiar with them, and don't know doom9. Any thought? @rust-lang/clippy 

Anyway, the test can be added in https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/doc/doc-fixable.rs#L55. The `.fixed` file will be fixed by `cargo dev bless`: https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/basics.md#building-and-testing

flip1995:
Hm, I think this is a case where those tools might be really popular in the domain they are used in, but I wouldn't consider them generally popular enough to put it in the whitelist. If we'd start that, we have to include a lot of things from different specific domains. I think for this use case you should add a `clippy.toml` and extend this list in your project - that's why it's configurable.

Enyium:
> I think for this use case you should add a `clippy.toml` and extend this list in your project - that's why it's configurable.

Okay, I did that.

PR #10255, opened 2023-01-29
prevents `len_without_is_empty` from yielding positive when `len` takes arguments besides `&self`

### LINKED ISSUE #9520: `len_without_is_empty` triggers when both `len` and `is_empty` have non-standard signatures
### Summary

I have a proxy type (a symbol from a string interner) which has `len` and `is_empty` method which:

- are fallible
- take the interner as a method parameter

`len_without_is_empty` triggers with these methods, which is unexpected since neither method matches the method signature the lint looks for.

### Lint Name

len_without_is_empty

### Reproducer

I tried this code:

```rust
#[warn(clippy::len_without_is_empty)]

pub trait Intern {
    type Symbol: Copy;
    
    fn lookup_symbol(&self, sym: Self::Symbol) -> Result<Option<&[u8]>, String>;
}


#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord)]
pub struct Symbol(u32);

impl Symbol {
    #[inline]
    #[must_use]
    pub fn is_empty<T, U>(self, interner: &T) -> bool
    where
        T: Intern<Symbol = U>,
        U: Copy + From<Symbol>,
    {
        if let Ok(Some(bytes)) = interner.lookup_symbol(self.into()) {
            bytes.is_empty()
        } else {
            true
        }
    }

    #[inline]
    #[must_use]
    // #[allow(clippy::len_without_is_empty)]
    pub fn len<T, U>(self, interner: &T) -> usize
    where
        T: Intern<Symbol = U>,
        U: Copy + From<Symbol>,
    {
        if let Ok(Some(bytes)) = interner.lookup_symbol(self.into()) {
            bytes.len()
        } else {
            0_usize
        }
    }
}
```

I saw this happen:

```
warning: struct `Symbol` has a public `len` method, but the `is_empty` method has an unexpected signature
  --> src/lib.rs:31:5
   |
31 | /     pub fn len<T, U>(self, interner: &T) -> usize
32 | |     where
33 | |         T: Intern<Symbol = U>,
34 | |         U: Copy + From<Symbol>,
   | |_______________________________^
   |
   = note: `#[warn(clippy::len_without_is_empty)]` on by default
note: `is_empty` defined here
  --> src/lib.rs:16:5
   |
16 | /     pub fn is_empty<T, U>(self, interner: &T) -> bool
17 | |     where
18 | |         T: Intern<Symbol = U>,
19 | |         U: Copy + From<Symbol>,
   | |_______________________________^
   = note: expected signature: `(self) -> bool`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty

warning: `playground` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
```

I expected to see this happen: No error because neither `len` nor `is_empty` has the expected signature.

### Version

```text
rustc 1.64.0 (a55dd71d5 2022-09-19)
binary: rustc
commit-hash: a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52
commit-date: 2022-09-19
host: x86_64-apple-darwin
release: 1.64.0
LLVM version: 14.0.6
```

### Additional Labels

_No response_

khayyamsaleem:
Fixes #9520
---

changelog: Fix [`len_without_is_empty`] false positive when len has a non-standard method signature

### DIFF
--- a/tests/ui/len_without_is_empty.rs
+++ b/tests/ui/len_without_is_empty.rs
@@ -282,4 +282,17 @@ impl AsyncLen {
     }
 }
 
+// issue #9520
+pub struct NonStandardLenAndIsEmptySignature;
+impl NonStandardLenAndIsEmptySignature {
+    // don't lint
+    pub fn len(&self, something: usize) -> usize {
+        something
+    }
+
+    pub fn is_empty(&self, something: usize) -> bool {
+        something == 0
+    }
+}
+
 fn main() {}
--- a/clippy_lints/src/len_zero.rs
+++ b/clippy_lints/src/len_zero.rs
@@ -135,6 +135,7 @@ impl<'tcx> LateLintPass<'tcx> for LenZero {
             if item.ident.name == sym::len;
             if let ImplItemKind::Fn(sig, _) = &item.kind;
             if sig.decl.implicit_self.has_implicit_self();
+            if sig.decl.inputs.len() == 1;
             if cx.effective_visibilities.is_exported(item.owner_id.def_id);
             if matches!(sig.decl.output, FnRetTy::Return(_));
             if let Some(imp) = get_parent_as_impl(cx.tcx, item.hir_id());


tests/ui/len_without_is_empty.rs
```diff
@@ -282,4 +282,17 @@ impl AsyncLen {
     }
 }
 
+// issue #9520
+pub struct NonStandardLenAndIsEmptySignature;
```
lopopolo, line 286: Would it be possible to add the reproducer from #9520 as a test case?

khayyamsaleem: Yep, let me add that in.

khayyamsaleem: added

lopopolo (approved):
Thanks for adding the rest case!

lopopolo:
Yes I believe so. I am not a member of the clippy team.

Jarcho:
Thank you for the fix. @bors r+

Just a note for future, it would be better to minimize the test cases. In this case the function bodies for the generic functions could have just been `unimplemented!()`.

Issue #10233, opened 2023-01-26
Migrate format_args!() lints to to ast::FormatArgs


m-ou-se:
Now that https://github.com/rust-lang/rust/pull/106745 is merged, Clippy has access to the original parsed/processed format_args!() invocation. This hopefully means that Clippy no longer needs to do anything that depends on the exact expansion of format_args!() and implementation details of fmt::Arguments (which will undergo changes in the (near!) future).

Details:

As of https://github.com/rust-lang/rust/pull/106745, the `format_args!()` builtin macro expands to a special [`ast::FormatArgs` AST node](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/format/struct.FormatArgs.html), which contains nothing that is specific to the standard library's `fmt::Arguments` implementation. During AST lowering, this node is expanded into HIR that is specific to how `fmt::Arguments` is implemented, and will change in the (near) future.)

Clippy's format_args lints are currently done in late passes that use the HIR. They should be changed to use the information from [ast::FormatArgs](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/format/struct.FormatArgs.html) instead. (Ideally by making them early passes, but that might not always be possible.)

This is part of https://github.com/rust-lang/rust/issues/99012

---

Lints to migrate:

- [x] ManualAssert - MANUAL_ASSERT: https://github.com/rust-lang/rust-clippy/pull/10276
- [x] Write: https://github.com/rust-lang/rust-clippy/pull/10275
  - [x] PRINT_WITH_NEWLINE
  - [x] PRINTLN_EMPTY_STRING
  - [x] PRINT_STDOUT
  - [x] PRINT_STDERR
  - [x] USE_DEBUG
  - [x] PRINT_LITERAL
  - [x] WRITE_WITH_NEWLINE
  - [x] WRITELN_EMPTY_STRING
  - [x] WRITE_LITERAL
- [x] FormatArgs: https://github.com/rust-lang/rust-clippy/pull/10484
  - [x] FORMAT_IN_FORMAT_ARGS
  - [x] TO_STRING_IN_FORMAT_ARGS
  - [x] UNINLINED_FORMAT_ARGS
  - [x] UNUSED_FORMAT_SPECS
- [ ] FormatImpl
  - [ ] RECURSIVE_FORMAT_IMPL
  - [ ] PRINT_IN_FORMAT_IMPL
- [ ] ExplicitWrite - EXPLICIT_WRITE
[...]

PR #10275, opened 2023-02-01
Migrate `write.rs` to `rustc_ast::FormatArgs`

Alexendoo:
changelog: none

Part 1 of #10233

The additions to `clippy_utils` are the main novelty of this PR, there's no removals yet since other parts still rely on `FormatArgsExpn`

The changes to `write.rs` itself are relatively straightforward, as for this one there's no lints in it that rely on type checking format params

r? @flip1995


nyurik:
This is awesome, thanks for the implementation! The code looks reasonable, and I learnt a few things.

If my understanding is correct, you essentially store every format_args invocation in a hashmap as `span -> AST` during the early pass, and later get the AST during the late pass, and you do it in the context of the write lint. My guess is that some other lints may need the same functionality, and it would be a bit wasteful for each lint to generate a full copy of such mapping for all format_args.  Is there a way to have a shared cache of those, where each lint that needs it can set some flag to enable this feature if that lint is being used?

llogiq:
Yeah, I suggested on zulip to have one internal lint pass that stores the mapping in a shared hash map.

nyurik:
Looks good!

Somewhat related -- does clippy optimize lint running if a lint is disabled? Or does it run all lints all the time, but hides output of the disabled lints?  If first, all arg formatting lints would need some way to indicate they are dependent on this internal lint.

flip1995:
> does clippy optimize lint running if a lint is disabled? Or does it run all lints all the time, but hides output of the disabled lints?

https://github.com/rust-lang/rust/issues/106983 Currently the latter. In the future maybe also the former.

Also in the future lint passes might run in parallel. This might lead to problems with the `thread_local`. But I think we should be fine here, because one is a early lint and the others are all late lints. And I think it's safe to assume that early and late lints will never run in parallel.  (Also this is far in the future).

Alexendoo:
It shouldn't be too much of an issue if it does optimise that in the future because you'd have to go pretty far out of your way to disable the collector lint, it's not part of any group like `clippy::all` so you'd have to `#[allow]` it by name, and the name isn't listed on the website or in `clippy-driver -W help`

Is clippy tested anywhere with `parallel_compiler`? That could be an issue if the threads change there, though it should still compile fine

clippy_utils/src/macros.rs
```diff
@@ -339,6 +343,77 @@ fn is_assert_arg(cx: &LateContext<'_>, expr: &Expr<'_>, assert_expn: ExpnId) ->
...
+        } else if ctxt.outer_expn().is_descendant_of(expn_id)
+            && macro_backtrace(expr.span)
+                .map(|macro_call| cx.tcx.item_name(macro_call.def_id))
+                .any(|name| matches!(name, sym::const_format_args | sym::format_args | sym::format_args_nl))
+        {
+            ControlFlow::Break(expr)
+        } else {
+            ControlFlow::Continue(Descend::No)
+        }
+    });
```
Alexendoo, line 387: Worth mentioning that I pulled this from the existing `FormatArgsExpn::find_nested` and the very start of `FormatArgsExpn::parse`:

https://github.com/rust-lang/rust-clippy/blob/8a9860901f0ae9782ff23fb793838a16f733a60b/clippy_utils/src/macros.rs#L989-L1004

https://github.com/rust-lang/rust-clippy/blob/8a9860901f0ae9782ff23fb793838a16f733a60b/clippy_utils/src/macros.rs#L893-L896

nyurik:
Friendly ping @flip1995 :)

nyurik:
:wave:

flip1995 (approved):
I don't really like the `thread_local` variable to store the state, but I can't come up with a better implementation. So let's merge this to move things forward.

flip1995:
@bors r+

Thanks!

Alexendoo:
Damn proc macros

Alexendoo:
Moved that test to `tests/ui/crashes` as it's a false positive now, or at least the suggestion would cause an error. It was an ICE until a week ago (#10401) so it's not too much of a regression at least

flip1995:
Thanks for the explanation! I'm fine with this "regression" since it is only for invalid code anyway IIUC.

Let's get this merged to get it in with this sync.

@bors +

PR #10276, opened 2023-02-01
Don't depend on FormatArgsExpn in ManualAssert.

m-ou-se:
Part of https://github.com/rust-lang/rust-clippy/issues/10233

changelog: none


flip1995:
r? @flip1995 

Assigning to myself, but @Alexendoo might be an even better reviewer of this, if they're interested.

Alexendoo:
The output is now the same on both 2018 and 2021 editions, [that seems correct](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=37b8a231459247a37df834b6caf536b5) so I think the previous difference was just down to the implementation

It'd say it's worth keeping the revisions to ensure they don't drift

Alexendoo (approved):

Alexendoo:
Looks good to me, thanks!

@bors r+

PR #10293, opened 2023-02-04
Negate suggestions when needed in `bool_assert_comparison`

### LINKED ISSUE #10291: Incorrect auto-fix: comparison to `false` breaks
I just stumbled over this and think the test cases specify the wrong behaviour.
This line:

https://github.com/rust-lang/rust-clippy/blob/8a9860901f0ae9782ff23fb793838a16f733a60b/tests/ui/bool_assert_comparison.rs#L89

Gets turned into this:

https://github.com/rust-lang/rust-clippy/blob/8a9860901f0ae9782ff23fb793838a16f733a60b/tests/ui/bool_assert_comparison.fixed#L89

But these are opposite asserts, so this will break the test.

The same applies to all the other calls to `assert_eq!(..., false)` and `assert_neq!(..., true)` of course.

Edit: sorry for the lack of template, I created this from the GH blame view which seems so sidestep that.

**Minimal Reproducer**

```rust
fn main() {
    println!("Hello, world!");
}

#[cfg(test)]
mod test {
    #[test]
    fn assert_correctly_transformed() {
        let must_be_false = false;

        assert_eq!(must_be_false, false);
    }
}
```

**Expected:**

```rust
        assert!(!must_be_false);
```

**Observed:**

```rust
        assert!(must_be_false);
```

(note the lack of `!`)

`clippy 0.1.69 (658fad6 2023-02-03)`

This does NOT happen on stable (yet) since the auto-fix was only introduced recently.

Alexendoo:
changelog: none assuming this gets into the same release as #10218

Fixes #10291

r? @dswij

Thanks to @black-puppydog for spotting it early

### DIFF
--- a/clippy_lints/src/bool_assert_comparison.rs
+++ b/clippy_lints/src/bool_assert_comparison.rs
@@ -1,5 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::macros::{find_assert_eq_args, root_macro_call_first_node};
+use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::{implements_trait, is_copy};
 use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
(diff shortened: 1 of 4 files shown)


dswij (approved):
Good find! Lgtm, thanks for this

PR #10295, opened 2023-02-04
fix(suspicious_to_owned): use span_suggestions to suggest both intents

### LINKED ISSUE #10294: suspicious_to_owned lint for Cow has invalid MaybeIncorrect fix
### Summary

`Applicability::MaybeIncorrect` is meant for fixes which may not be the user's intention, but must result in valid Rust code. However, the current suggestion when applied results in incorrect Rust code.

ref: [`suspicious_to_owned` lint page](https://rust-lang.github.io/rust-clippy/master/#suspicious_to_owned)

### I tried this code:

```rust
let _ = Cow::Borrowed(&()).to_owned();
```

### I expected to see this happen:

I would see two suggested fixes that would replace this by:
```rust
let _ = Cow::Borrowed(&()).clone();
```
or
```rust
let _ = Cow::Borrowed(&()).into_owned();
```

###  Instead, this happened:
Applying the single suggested fix:

```rust
let _ = Cow::Borrowed(&()).clone()` or `Cow::Borrowed(&()).into_owned();
```

### Version

```text
binary: rustc
commit-date: 2022-11-02
release: 1.65.0
LLVM version: 15.0.0
```

### Additional Labels

DesmondWillowbrook:
fixes #10294

changelog: [`suspicious_to_owned`]: suggestions now produce valid Rust code

### DIFF
--- a/clippy_lints/src/methods/suspicious_to_owned.rs
+++ b/clippy_lints/src/methods/suspicious_to_owned.rs
@@ -1,4 +1,4 @@
-use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::is_diag_trait_item;
 use clippy_utils::source::snippet_with_context;
 use if_chain::if_chain;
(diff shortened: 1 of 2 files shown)


clippy_lints/src/methods/suspicious_to_owned.rs
```diff
@@ -17,19 +17,25 @@ pub fn check(cx: &LateContext<'_>, expr: &hir::Expr<'_>, recv: &hir::Expr<'_>) -
...
                 )),
-                "consider using, depending on intent",
-                format!("{recv_snip}.clone()` or `{recv_snip}.into_owned()"),
-                app,
+                |diag| {
+                    diag.span_suggestions(
+                        expr.span,
+                        "depending on intent, either make the Cow an Owned variant or clone the Cow itself",
+                        [format!("{recv_snip}.into_owned()"), format!("{recv_snip}.clone()")],
+                        Applicability::Unspecified
```
Jarcho, line 36: Since this splits the combined suggestion into two overlapping suggestions this should keep the same applicability it had before.

DesmondWillowbrook: You're right, it should. Fixed it now!

Jarcho:
Since you're already changing things, it might look better when given two suggestions with different help messages.

e.g.
```
span_suggestion(..., "either convert it to it's owned type", ...)
span_suggestion(..., "or clone it", ...)
```

DesmondWillowbrook:
Thanks for the suggestion! I initially wasn't sure about using the method, but this looks much better.

Jarcho:
That is nicer to read.

Thank you for the fix. @bors r+

Issue #5736, opened 2020-06-22
Split `missing_docs_in_private_items` into various lints


jhpratt:
Would it be feasible and acceptable to split `missing_docs_in_private_items` into a few different lints? I find myself quite frequently adding `#[allow]` to `enum` variants and `struct` fields, in large part because the naming is typically self-explanatory. Even splitting the lint in two (fields/variants and everything else) would be _incredibly_ useful for me, and would avoid cluttering declarations with unnecessary (imo) `#[allow]` attributes.

Having different lints for modules and top-level crate documentation would also be great, but are less important to me.

jhpratt:
A quick `rg` and `tokei` on a project of mine (admittedly with a large number of structs) shows that I have `#[allow(clippy::missing_docs_in_private_items)]` 64 times out of 1387 total lines of Rust code (4.6% of all lines). All 64 are on `enum` variants and `struct` fields.

If this is something that is desired, I'd be interested in contributing once I get some free time.

flip1995:
Maybe even better: Don't split it, but make it configurable, so that you can configure it to only warn on functions for example.

jhpratt:
That could certainly work, and would be backwards-compatible as well.

jhpratt:
Another situation I've found myself allowing the lint regularly is `static`s, `const`s, `struct`s, etc. that are only used within the function.

flip1995:
I think we should not lint at all on definitions inside a function.

pvdrz:
On `bindgen` we do this hack to keep reusable code documented by  marking it as public but putting it inside a module that's only public if compiled with a certain feature: https://github.com/rust-lang/rust-bindgen/blob/a7ff8e136027c9542f65eca97a7ff3c0b870518d/bindgen/lib.rs#L36-L50

This has the undesired effect of not being able to detect unused items because they are public.

If this lint could be configured to detect missing docs in `pub(crate)` items, we could remove this hack and remove the downside of not detecting unused code.

Let me know if this would be a reasonable change and I will start working on it.

flip1995:
I could see a config option, that you can set to specify if the lint should lint on all private items or only on crate items. All private items should be the default though. Let me know if you need help implementing this. `clippy_lints/src/utils/conf.rs` is a good starting point to look at for adding config options.

PR #10303, opened 2023-02-06
Add configuration to lint missing docs of `pub(crate)` items

pvdrz:
Fixes this: https://github.com/rust-lang/rust-clippy/issues/5736#issuecomment-1412442404

TODO:
- [x] Needs docs
- [x] Needs better names
- [x] Should `pub` items be checked to when this new option is enabled? I'm saying no because `missing_docs` already exists

@flip1995 I'd like to get some input from you :) 

---

changelog: Enhancement: [`missing_docs_in_private_items`]: Added new configuration `missing-docs-in-crate-items` to lint on items visible within the current crate. For example, `pub(crate)` items.
[#10303](https://github.com/rust-lang/rust-clippy/pull/10303)

### DIFF
--- a/book/src/lint_configuration.md
+++ b/book/src/lint_configuration.md
@@ -542,7 +542,8 @@ if no suggestion can be made.
 
 
 ### missing-docs-in-crate-items
-Whether to **only** check for missing documentation in `pub(crate)` items.
+Whether to **only** check for missing documentation in items visible within the current
+crate. For example, `pub(crate)` items.
 
 **Default Value:** `false` (`bool`)
 
--- a/book/src/lint_configuration.md
+++ b/book/src/lint_configuration.md
@@ -53,6 +53,7 @@ Please use that command to update the file and do not edit it by hand.
 | [ignore-interior-mutability](#ignore-interior-mutability) | `["bytes::Bytes"]` |
 | [allow-mixed-uninlined-format-args](#allow-mixed-uninlined-format-args) | `true` |
 | [suppress-restriction-lint-in-const](#suppress-restriction-lint-in-const) | `false` |
+| [missing-docs-in-crate-items](#missing-docs-in-crate-items) | `false` |
 
 ### arithmetic-side-effects-allowed
 Suppress checking of the passed type names in all types of operations.
@@ -538,6 +539,14 @@ if no suggestion can be made.
 **Default Value:** `false` (`bool`)
 
 * [indexing_slicing](https://rust-lang.github.io/rust-clippy/master/index.html#indexing_slicing)
+
+
+### missing-docs-in-crate-items
+Whether to **only** check for missing documentation in `pub(crate)` items.
+
+**Default Value:** `false` (`bool`)
+
+* [missing_docs_in_private_items](https://rust-lang.github.io/rust-clippy/master/index.html#missing_docs_in_private_items)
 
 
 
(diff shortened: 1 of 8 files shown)


flip1995:
> * Should `pub` items be checked to when this new option is enabled? I'm saying no because `missing_docs` already exists

Agreed. I also am fine with the name. I'll let @giraffate do the final review, I only skimmed over the name and documentation.

tests/ui-toml/pub_crate_missing_docs/pub_crate_missing_doc.rs
```diff
@@ -0,0 +1,51 @@
...
+    pub(crate) struct CrateStructNoDocs {
+        /// some docs
+        pub(crate) crate_field_with_docs: (),
+        pub(crate) crate_field_no_docs: (),
+        /// some docs
+        priv_field_with_docs: (),
+        priv_field_no_docs: (),
+    }
+}
+
```
giraffate, line 47: question: This configuration says `Whether to **only** check for missing documentation in pub(crate) items`. Should the case below in top-level module be checked?
```suggestion
type Typedef = String;
pub type PubTypedef = String;

```

giraffate (commented):
Thanks! I asked one question.

The `tests/ui-toml/pub_crate_missing_docs/pub_crate_missing_doc.stdout` file is empty, so it should be removed.

tests/ui-toml/pub_crate_missing_docs/pub_crate_missing_doc.rs
pvdrz: I have mixed feelings about this. On one hand `Typedef` is visible inside of the whole crate so it would be checked. However, some people might want to just check stuff that's marked explicitly as `pub(crate)` or `pub(super)` (where `super == crate`).

giraffate: IMO it's okay that `Typedef` is checked, but it seems to be confusing the note of configuration, for example how about the following:
`Whether to check for missing documentation in items visible within the current crate, for example pub(crate) items`

And can the tests like https://github.com/rust-lang/rust-clippy/pull/10303/files#r1106516655 be added?

pvdrz: sure I'll do both and ping you again

book/src/lint_configuration.md
```diff
@@ -540,4 +541,12 @@ if no suggestion can be made.
 * [indexing_slicing](https://rust-lang.github.io/rust-clippy/master/index.html#indexing_slicing)
 
 
+### missing-docs-in-crate-items
+Whether to **only** check for missing documentation in `pub(crate)` items.
```
giraffate, line 545: Can this be updated in the same way as https://github.com/rust-lang/rust-clippy/pull/10303/files#diff-1d884fd439e15a9e47e1135cdc0ab5ee896a876b9f4c256ecce4be6c69a8831dR459?

clippy_lints/src/missing_doc.rs
```diff
@@ -35,6 +36,8 @@ declare_clippy_lint! {
 }
 
 pub struct MissingDoc {
+    /// Whether to only check for missing docs in `pub(crate)` items.
```
giraffate, line 39: Can this be updated in the same way as https://github.com/rust-lang/rust-clippy/pull/10303/files#diff-1d884fd439e15a9e47e1135cdc0ab5ee896a876b9f4c256ecce4be6c69a8831dR459?

giraffate (approved):
Overall looks good, thanks!

I made small comments. I'll merge this once those are addressed.

pvdrz:
Thanks @giraffate! It seems we're good to go then

giraffate:
@bors r+

Thanks!

PR #10310, opened 2023-02-09
[arithmetic_side_effects] Fix #10209

### LINKED ISSUE #10209: arithmetic_side_effects: can’t see through trivial constant items
### Summary

`arithmetic_side_effects` lint correctly considers operations such as `(somevar: u64) / 2` as non-side-effectful. However as soon as `2` is extracted into a constant, the lint starts firing.

I don't think it is critical that we handle complex constant evaluation here, but it would be nice if it handled at least basic literals like that so that this lint didn't become a motivation to inline magic numbers into the code.

### Lint Name

arithmetic_side_effects

### Reproducer

I tried this code:

```rust
#![deny(clippy::arithmetic_side_effects)]
fn foo(input: u64) -> u64 {
    const A: u64 = 1;
    input / A
}
```

I saw this happen:

```
error: arithmetic operation that can potentially result in unexpected side-effects
 --> src/lib.rs:5:5
  |
5 |     input / A
  |     ^^^^^^^^^
```

### Version

```text
rustc nightly / 0.1.68 (2023-01-17 3984bc5) on the playground
```

### Additional Labels

_No response_

c410-f3r:
Fix #10209

---

changelog: Enhancement: [`arithmetic_side_effects`]: No longer lints, if safe constant values are used.
[#10310](https://github.com/rust-lang/rust-clippy/pull/10310)

### DIFF
--- a/clippy_lints/src/operators/arithmetic_side_effects.rs
+++ b/clippy_lints/src/operators/arithmetic_side_effects.rs
@@ -1,6 +1,6 @@
 use super::ARITHMETIC_SIDE_EFFECTS;
 use clippy_utils::{
-    consts::{constant, constant_simple},
+    consts::{constant, constant_simple, Constant},
     diagnostics::span_lint,
     peel_hir_expr_refs, peel_hir_expr_unary,
 };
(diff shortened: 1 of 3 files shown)


Alexendoo (commented):
Thanks, looks good

Since the lint is doing a moderate amount of work it'd be worth adding an [`is_lint_allowed`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_lint_allowed.html) check in `should_skip_expr`

c410-f3r:
> Thanks, looks good
> 
> Since the lint is doing a moderate amount of work it'd be worth adding an [`is_lint_allowed`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_lint_allowed.html) check in `should_skip_expr`

Done

Alexendoo (approved):

Alexendoo:
👍

@bors r+

PR #10311, opened 2023-02-09
[never_loop] Fix #10304

### LINKED ISSUE #10304: clippy::never_loop has false positive with break from labeled blocks
### Summary

clippy::never_loop is triggering on a loop with a break to a labeled block. i couldnt seem to get a false positive for nested loops with break statements, so its probably a bug specifically with [rfc#2046](https://github.com/rust-lang/rfcs/pull/2046) (stabalized in 1.65).

### Lint Name

never_loop

### Reproducer

I tried this code:

```rust
fn main() {
    for _ in 0..10 {
        'block: {
            break 'block;
            return;
        }
        println!("looped");
    }
}
```
[playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=794ac224fae5cc53f4963cf6f7cb1838)

I saw this happen:

```
error: this loop never actually loops
  --> src/main.rs:4:5
   |
4  | /     for _ in 0..10 {
5  | |         'block: {
6  | |             break 'block;
7  | |             return;
8  | |         }
9  | |         println!("looped");
10 | |     }
   | |_____^
```
This is in spite of this loop looping 10 times when run.

### Version

```text
rustc 1.67.0 (fc594f156 2023-01-24)
binary: rustc
commit-hash: fc594f15669680fa70d255faec3ca3fb507c3405
commit-date: 2023-01-24
release: 1.67.0
LLVM version: 15.0.6
```

### Additional Labels

_No response_

samueltardieu:
It is not sufficient to ignore break from a block inside the loop. Instructions after the break must be ignored, as they are unreachable. This is also true for all instructions in outer blocks and loops until the right block is reached.

Fixes #10304

---

changelog: FP: [`never_loop`]: No longer lints, for statements following break statements for outer blocks.
[#10311](https://github.com/rust-lang/rust-clippy/pull/10311)

### DIFF
--- a/clippy_lints/src/loops/never_loop.rs
+++ b/clippy_lints/src/loops/never_loop.rs
@@ -68,18 +68,6 @@ fn combine_seq(first: NeverLoopResult, second: NeverLoopResult) -> NeverLoopResu
     }
 }
 
-// Combine two results where both parts are called but not necessarily in order.
-#[must_use]
-fn combine_both(left: NeverLoopResult, right: NeverLoopResult) -> NeverLoopResult {
-    match (left, right) {
-        (NeverLoopResult::MayContinueMainLoop, _) | (_, NeverLoopResult::MayContinueMainLoop) => {
-            NeverLoopResult::MayContinueMainLoop
-        },
-        (NeverLoopResult::AlwaysBreak, _) | (_, NeverLoopResult::AlwaysBreak) => NeverLoopResult::AlwaysBreak,
-        (NeverLoopResult::Otherwise, NeverLoopResult::Otherwise) => NeverLoopResult::Otherwise,
-    }
-}
-
 // Combine two results where only one of the part may have been executed.
 #[must_use]
 fn combine_branches(b1: NeverLoopResult, b2: NeverLoopResult) -> NeverLoopResult {
@@ -139,7 +127,7 @@ fn never_loop_expr(expr: &Expr<'_>, ignore_ids: &mut Vec<HirId>, main_loop_id: H
         ExprKind::Struct(_, fields, base) => {
             let fields = never_loop_expr_all(&mut fields.iter().map(|f| f.expr), ignore_ids, main_loop_id);
             if let Some(base) = base {
-                combine_both(fields, never_loop_expr(base, ignore_ids, main_loop_id))
+                combine_seq(fields, never_loop_expr(base, ignore_ids, main_loop_id))
             } else {
                 fields
             }
@@ -218,7 +206,7 @@ fn never_loop_expr(expr: &Expr<'_>, ignore_ids: &mut Vec<HirId>, main_loop_id: H
                 | InlineAsmOperand::SymFn { .. }
                 | InlineAsmOperand::SymStatic { .. } => NeverLoopResult::Otherwise,
             })
-            .fold(NeverLoopResult::Otherwise, combine_both),
+            .fold(NeverLoopResult::Otherwise, combine_seq),
         ExprKind::Yield(_, _)
         | ExprKind::Closure { .. }
         | ExprKind::Path(_)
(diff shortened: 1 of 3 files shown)


Jarcho:
Overall looks good. While you're at changing this you can remove `combine_both`. Sub-expressions have always been executed in order, and it's been defined to be that way for quite a while now.

samueltardieu:
> Overall looks good. While you're at changing this you can remove `combine_both`. Sub-expressions have always been executed in order, and it's been defined to be that way for quite a while now.

Indeed, will do.

samueltardieu:
It is ready now. I first removed `combine_both()` and merged the changes into the existing commit, then I realized it probably deserved a commit on its own for traceability.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -56,35 +59,36 @@ fn absorb_break(arg: NeverLoopResult) -> NeverLoopResult {
...
+        (NeverLoopResult::IgnoreUntilEnd(a), NeverLoopResult::IgnoreUntilEnd(b)) => {
+            if ignore_ids.iter().find(|&e| e == &a || e == &b).unwrap() == &a {
+                NeverLoopResult::IgnoreUntilEnd(b)
+            } else {
+                NeverLoopResult::IgnoreUntilEnd(a)
+            }
+        },
+        (NeverLoopResult::IgnoreUntilEnd(_), NeverLoopResult::AlwaysBreak)
+        | (NeverLoopResult::AlwaysBreak, NeverLoopResult::IgnoreUntilEnd(_) | NeverLoopResult::AlwaysBreak) => {
+            NeverLoopResult::AlwaysBreak
```
Jarcho, line 90: Shouldn't `(IgnoreUntilEnd(_), AlwaysBreak)` return `IgnoreUntilEnd(_)` as it's the smaller of the two scopes?

clippy_lints/src/loops/never_loop.rs
```diff
@@ -176,7 +180,10 @@ fn never_loop_expr(expr: &Expr<'_>, ignore_ids: &mut Vec<HirId>, main_loop_id: H
             }
             let ret = never_loop_block(b, ignore_ids, main_loop_id);
             ignore_ids.pop();
```
Jarcho, line 182: This should have the same condition as the the push call. Looks like you didn't add the bug, but it still needs to be fixed.

clippy_lints/src/loops/never_loop.rs
```diff
@@ -234,16 +241,19 @@ fn never_loop_expr_all<'a, T: Iterator<Item = &'a Expr<'a>>>(
...
-        .fold(NeverLoopResult::Otherwise, combine_both)
+        .fold(NeverLoopResult::Otherwise, combine_seq)
 }
 
 fn never_loop_expr_branch<'a, T: Iterator<Item = &'a Expr<'a>>>(
     e: &mut T,
     ignore_ids: &mut Vec<HirId>,
     main_loop_id: HirId,
 ) -> NeverLoopResult {
+    let ignore_ids_saved = ignore_ids.clone();
```
Jarcho, line 252: You shouldn't need to clone here. By the time `never_loop_expr` returns either `ignore_ids` is back to it's original state, or you have a bug which needs to be fixed.

Jarcho (changes requested):
Took a closer second look at this and there are a few things which need to be fixed.

clippy_lints/src/loops/never_loop.rs
samueltardieu: The `.fold()` executes at the same time as the iterator, not at the end, I cannot borrow `ignore_ids` both in the `.map()` (mutably) and in the `.fold()`. So either I clone `ignore_ids`, or I `.collect()` the results of the map.

Jarcho: You can inline the `map` into the `fold`.

clippy_lints/src/loops/never_loop.rs
samueltardieu: Right. Fixed and checked in `test24()`.

clippy_lints/src/loops/never_loop.rs
samueltardieu: Fixed in a new commit and checked in `test21()`.

clippy_lints/src/loops/never_loop.rs
samueltardieu: Done.

PR #10312, opened 2023-02-10
Add `transmute_slice_to_larger_element_type` lint

### LINKED ISSUE #10285: deny transmute of `&[T]` into `&[U]` where `U` is larger than `T`
### What it does

This prevents a kind of UB.

Examples:

```rs
let bytes = &[1u8, 2u8, 3u8, 4u8] as &[u8];
let alt_slice: &[u32] = unsafe { core::mem::transmute(bytes) };
// expected to be expanded
assert_eq!(alt_slice[0], 1);
assert_eq!(alt_slice[1], 2);
assert_eq!(alt_slice[2], 3);
assert_eq!(alt_slice[3], 4);
```

This is not true. `alt_slice[0]` have (mostly) `0x01020304` or `0x04030201` depending on byte-order, and left is looks like garbage and user thinks it is not what (s)he wanted.

Also, this contains UB. Miri reports call to `transmute` is UB.

I believe: `mem::transmute::<&[T], &[U]>` is UB if and only if `mem::size_of<T> < mem::size_of<U>`.

Alternatives:

This may not be `MachineApplicable`, because:
If user is expected to keep length and transmute each element:

```rs
// let original_slice: &[T] = ...
let alt_slice: &[U] = original_slice
    .iter()
    .map(|original| unsafe { core::mem::transmute(original) })
    .collect::<Vec<_>>()
    .as_slice()
```

If user is expected to transmute entire bit pattern of pointing bit-pattern: 
```rs
// let original_slice: &[T] = ...
let (prefix, align_result, suffix) = original_slice.align_to::<U>();
assert_eq!(prefix.len(), 0);
assert_eq!(suffix.len(), 0);
```

### Lint Name

transmute_to_larger_element_slice

### Category

correctness

### Advantage

- Prevents a kind of UB.

### Drawbacks

_No response_

### Example

```rust
// let bytes = &[1u8, 2u8, 3u8, 4u8] as &[u8];
let alt_slice: &[u32] = unsafe { core::mem::transmute(bytes) };
```

Could be written as:

```rust
let alt_slice: &[u32] = original_slice
    .iter()
    .map(|original| unsafe { core::mem::transmute(original) })
    .collect::<Vec<_>>()
    .as_slice()
```

or

```rust
let (prefix, align_result, suffix) = original_slice.align_to::<u32>();
assert_eq!(prefix.len(), 0);
assert_eq!(suffix.len(), 0);
```

KisaragiEffective:
- [ ] `cargo test` passes locally
  - [ ] except `fmt` check
- [ ] Run `cargo dev fmt`
  - This was not possible, because running `cargo dev fmt` somehow implied `cargo dev fmt --check`

close #10285 

---

changelog: New lint: [`transmute_slice_to_larger_element_type`]
[#10312](https://github.com/rust-lang/rust-clippy/pull/10312)

### DIFF
--- a/clippy_lints/src/transmute/mod.rs
+++ b/clippy_lints/src/transmute/mod.rs
@@ -8,6 +8,7 @@ mod transmute_num_to_bytes;
 mod transmute_ptr_to_ptr;
 mod transmute_ptr_to_ref;
 mod transmute_ref_to_ref;
+mod transmute_slice_to_larger_element_type;
 mod transmute_undefined_repr;
 mod transmutes_expressible_as_ptr_casts;
 mod transmuting_null;
@@ -438,6 +439,25 @@ declare_clippy_lint! {
     "transmute results in a null function pointer, which is undefined behavior"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// ### Why is this bad?
+    ///
+    /// ### Example
+    /// ```rust
+    /// // example code where clippy issues a warning
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// // example code which does not raise clippy warning
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub TRANSMUTE_SLICE_TO_LARGER_ELEMENT_TYPE,
+    correctness,
+    "default lint description"
+}
+
 pub struct Transmute {
     msrv: Msrv,
 }
@@ -458,6 +478,7 @@ impl_lint_pass!(Transmute => [
     TRANSMUTE_UNDEFINED_REPR,
     TRANSMUTING_NULL,
     TRANSMUTE_NULL_TO_FN,
+    TRANSMUTE_SLICE_TO_LARGER_ELEMENT_TYPE,
 ]);
 impl Transmute {
     #[must_use]
@@ -503,6 +524,7 @@ impl<'tcx> LateLintPass<'tcx> for Transmute {
                     | transmute_int_to_float::check(cx, e, from_ty, to_ty, arg, const_context)
                     | transmute_float_to_int::check(cx, e, from_ty, to_ty, arg, const_context)
                     | transmute_num_to_bytes::check(cx, e, from_ty, to_ty, arg, const_context)
+                    | transmute_slice_to_larger_element_type::check(cx, e, from_ty, to_ty, arg)
                     | (
                         unsound_collection_transmute::check(cx, e, from_ty, to_ty)
                         || transmute_undefined_repr::check(cx, e, from_ty, to_ty)
--- a/clippy_lints/src/transmute/mod.rs
+++ b/clippy_lints/src/transmute/mod.rs
@@ -441,21 +441,32 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
+    /// Checks for slice creation which has larger element before transmute.
     ///
     /// ### Why is this bad?
+    /// Creating those slices leads to out-of-bounds read, considered Undefined Behavior.
     ///
     /// ### Example
     /// ```rust
-    /// // example code where clippy issues a warning
+    /// let i8_slice: &[i8] = &[1i8, 2, 3, 4];
+    //  let i32_slice: &[i32] = unsafe { std::mem::transmute(i8_slice) };
     /// ```
+    ///
     /// Use instead:
+    ///
     /// ```rust
-    /// // example code which does not raise clippy warning
+    /// let i32_slice: &[i32] = i8_slice.iter().map(|item| unsafe { std::mem::transmute(item) }).collect::<Vec<_>>().to_slice();
+    /// ```
+    ///
+    /// or, alternatively:
+    ///
+    /// ```rust
+    /// let i32_slice: &[i32] = std::slice::align_to::<i32>(i8_slice).1;
     /// ```
     #[clippy::version = "1.69.0"]
     pub TRANSMUTE_SLICE_TO_LARGER_ELEMENT_TYPE,
     correctness,
-    "default lint description"
+    "transmute leads out-of-bounds read, which is undefined behavior"
 }
 
 pub struct Transmute {
--- a/clippy_lints/src/transmute/mod.rs
+++ b/clippy_lints/src/transmute/mod.rs
@@ -449,19 +449,21 @@ declare_clippy_lint! {
     /// ### Example
     /// ```rust
     /// let i8_slice: &[i8] = &[1i8, 2, 3, 4];
-    //  let i32_slice: &[i32] = unsafe { std::mem::transmute(i8_slice) };
+    /// let i32_slice: &[i32] = unsafe { std::mem::transmute(i8_slice) };
     /// ```
     ///
     /// Use instead:
     ///
     /// ```rust
+    /// let i8_slice: &[i8] = &[1i8, 2, 3, 4];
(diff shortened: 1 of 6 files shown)


KisaragiEffective:
(force pushed to revert local change that was try about `clippy dev fmt`)

clippy_lints/src/transmute/mod.rs
```diff
@@ -438,6 +439,36 @@ declare_clippy_lint! {
...
+    /// ### Example
+    /// ```rust
+    /// let i8_slice: &[i8] = &[1i8, 2, 3, 4];
+    //  let i32_slice: &[i32] = unsafe { std::mem::transmute(i8_slice) };
+    /// ```
+    ///
+    /// Use instead:
+    ///
+    /// ```rust
+    /// let i32_slice: &[i32] = i8_slice.iter().map(|item| unsafe { std::mem::transmute(item) }).collect::<Vec<_>>().to_slice();
```
samueltardieu, line 459: Have you tried this suggestion? I don't think this will do what you want: this attempts to transmute `&i8` to `i32`.

Also, isn't the vector you create dropped at the end of the statement? (and don't you mean `.as_slice()`?)

clippy_lints/src/transmute/mod.rs
```diff
@@ -438,6 +439,36 @@ declare_clippy_lint! {
...
+    /// ```
+    ///
+    /// or, alternatively:
+    ///
+    /// ```rust
+    /// let i32_slice: &[i32] = std::slice::align_to::<i32>(i8_slice).1;
```
samueltardieu: Don't you mean `unsafe { i8_slice.align_to::<i32>().1 }`?

clippy_lints/src/transmute/transmute_slice_to_larger_element_type.rs
```diff
@@ -0,0 +1,71 @@
...
+use clippy_utils::source::reindent_multiline;
+use clippy_utils::sugg;
+use clippy_utils::ty::approx_ty_size;
+use rustc_errors::Applicability;
+use rustc_hir::Expr;
+use rustc_lint::LateContext;
+use rustc_middle::ty::{self, Ty};
+use std::borrow::Cow;
+
+// TODO: Adjust the parameters as necessary
```
samueltardieu, line 12: What is there left to adjust?

clippy_lints/src/transmute/transmute_slice_to_larger_element_type.rs
```diff
@@ -0,0 +1,71 @@
...
+
+                true
+            } else {
+                false
+            }
+        } else {
+            false
+        }
+    } else {
+        false
```
samueltardieu, line 69: You can use `if_chain!` to avoid this cascade fo `else { false }` (or `if let … && let … && …` since feature `let_chains` is enabled).

clippy_lints/src/transmute/transmute_slice_to_larger_element_type.rs
KisaragiEffective: There's nothing, I've forgot to remove :c

KisaragiEffective: fixed in [3bf3cdf](https://github.com/rust-lang/rust-clippy/pull/10312/commits/3bf3cdfdf3f5041592d9403b7acf583fad1a7018)

Jarcho:
This would be fine without a suggestion. It's more likely that the size needs to change than anything.

This would be better as a lint to check if the size changes, not just becomes larger. Both almost certainly do not have the intended result.

KisaragiEffective:
> This would be better as a lint to check if the size changes, not just becomes larger. 

Yes, that will not be what is wanted. However, if slice gets *truncated* by transmute, it does not cause UB immediately (still may cause: `&[T]` to `&[bool]` would cause UB for most `T`). So it would be fall in suspicious category and better to create separate lint IMO.

Jarcho:
It's not instant UB to grow the size either, it depends on the underlying allocation. e.g.

```rust
let x: [u32; 2] = [0, 1];
let y: &[u64] = transmute(&x[0..1]); // This is 'fine'
```

asquared31415:
Note that
> ```rust
> let x: [u32; 2] = [0, 1];
> let y: &[u64] = transmute(&x[0..1]); // This is 'fine'
> ```
is ***not*** fine, as per Miri.  The input slice is only allowed to access the 4 bytes that make up the element from `[0..1]`, and Miri errors before the `transmute` returns, because the transmute is trying to access 8 bytes.  The only way to do this properly is to go through raw pointers, `core::slice::from_raw_parts(x.as_ptr().cast::<u64>(), 1);`

Transmuting a slice to any larger element type is always an error because of this, since the length in *elements* stays the same, but it tries to access more bytes.  This is basically the same as [cast_slice_different_sizes](https://rust-lang.github.io/rust-clippy/master/index.html#cast_slice_different_sizes), but is **immediate** UB because it's references.

(additionally there's alignment issues and such, but those can be checked/fixed more easily)

re: shrinking instead of growing, it's probably useful as a different lint, since it could be intended, but still definitely suspicious, i would be in favor of either warn or deny on such a lint.  But this lint is always correct to be `deny` because the code is always UB to access more bytes.

clippy_lints/src/transmute/mod.rs
asquared31415: Yes, this code does not compile because of the temporary `Vec` being dropped, and the `transmute` on each element is wrong as well.  A "proper" suggestion might be to go through `core::slice::from_raw_parts`, but also because of alignment requirements, I don't think it can be an automated suggestion.  Perhaps "consider `core::slice::from_raw_parts(_mut)` instead" would suffice?

Jarcho:
> is not fine, as per Miri. The input slice is only allowed to access the 4 bytes that make up the element from [0..1], and Miri errors before the transmute returns, because the transmute is trying to access 8 bytes. The only way to do this properly is to go through raw pointers, core::slice::from_raw_parts(x.as_ptr().cast::<u64>(), 1);

Looks like this is much more defined than when I last checked (years ago). Both tree and stack borrows definitely set this as ub with references. As a tangential note, it is valid for raw pointers:

```
fn main(){
    let x = [0u64; 4];
    let x = &x as *const [u64; 4] as *const u32;
    let x = std::ptr::slice_from_raw_parts(x, 4) as *const [u64];
    let x = unsafe { &*x };
    println!("{:?}", x);
}
```

---

I'm still not convinced this is worth a lint separate from just any size changes. Converting to a smaller type without changing the length is incredibly close to certainly wrong. I've never seen a valid use case for it. I can't even imagine what that use case would be.

xFrednet:
Hey @KisaragiEffective this is a ping from triage. Are you planning to continue working on this PR?

KisaragiEffective:
> Hey @KisaragiEffective this is a ping from triage. Are you planning to continue working on this PR?

I will, i am not able to decide whether the lint will fire not only larger, but smaller one. The current impl has a bug that suggests take ref on temp, but other parts still can be used.

xFrednet:
> I will, i am not able to decide whether the lint will fire not only larger, but smaller one.

Interesting, I would lint that as well. Having `&[u32]` -> `&[u8]` will not result in out of bounds reading, but it's still very likely wrong. Then the lint could just be called `transmute_slice` or something similar. 

If there is a valid use case to transmute `&[u32]` -> `&[u8]` we can just add a configuration value to allow this. But I think just linting it is fine and will catch more cases that should be linted :D

xFrednet:

PR #10270, opened 2023-01-31
Update dependencies `cargo_metadata` and `clap`

kadiwa4:
changelog: none

Allows removing some duplicate dependencies in rust-lang/rust's `Cargo.lock`.

PR #10314, opened 2023-02-10
Fix CLI of clippy_dev

flip1995:
Clap was updated in rust-lang/rust-clippy#10270, which broke the command line of clippy_dev. This swaps out contains_id, which now returns always true in the places it was used with get_flag.

r? @xFrednet 

This should also fix https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.60cargo.20dev.20setup.20intellij.60.20is.20degraded/near/325770850

changelog: none

### DIFF
--- a/clippy_dev/src/main.rs
+++ b/clippy_dev/src/main.rs
@@ -11,22 +11,22 @@ fn main() {
 
     match matches.subcommand() {
         Some(("bless", matches)) => {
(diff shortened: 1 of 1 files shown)


xFrednet (approved):

xFrednet:
LGTM, I remember struggling with the new function names as well, when I updated clap in marker. Thank you!

@bors r+

Issue #10267, opened 2023-01-31
Make `SpanlessEq` more consistent


flip1995:
### Description

The `SpanlessEq::eq_expr` implementation has a wildcard match arm that just returns `false`. 

https://github.com/rust-lang/rust-clippy/blob/d227f18c2eb7c1d21d590ea4d8708d8dc5d01308/clippy_utils/src/hir_utils.rs#L305

This is inconsistent with `SpanlessHash`, which implements hashing for each expressions type. 

https://github.com/rust-lang/rust-clippy/blob/d227f18c2eb7c1d21d590ea4d8708d8dc5d01308/clippy_utils/src/hir_utils.rs#L785-L789

`PartialEq` and `Hash` implementations have to be consistent. The same should apply to `SpanlessEq` and `SpanlessHash`

### Version

```text
rustc 1.69.0-nightly (d7948c843 2023-01-26)
binary: rustc
commit-hash: d7948c843de94245c794e8c63dd4301a78bb5ba3
commit-date: 2023-01-26
host: x86_64-unknown-linux-gnu
release: 1.69.0-nightly
LLVM version: 15.0.7
```

llogiq:
Note that SpanlessEq is used in `eq_op`, so changing the default to `true` would introduce false positives there whenever new expr variants are added.

Also the consistency between Hash and Eq goes only one way: Two equal values must hash to the same hash, but two unequal values may hash to the same hash.

So I'd suggest seeking further consistency here is misguided. Perhaps we should add a comment stating that fact.

flip1995:
> so changing the default to `true` would introduce false positives there whenever new expr variants are added.

I don't suggest changing the default. We should remove the wildcard completely. My main motivation for this is, that then, when new `ExprKind`s are added in `rustc`, contributors are forced to also deal with this in `eq_expr`. Currently only `hash_expr` is updated.

I'm fine with as a first step to do a NFC and replace the wildcard with all missing expr kinds.

Jarcho:
Probably worth having the final pattern be `(ExprKind::A | ExprKind::B | ...,  _)`. At least that way adding a new variant will cause an error.

Yoric:
I'd be happy to pick this issue as a way to familiarize myself with clippy's code.

But

> uh, removing the wildcard completely might not even be possible, because we have to deal with different ExprKinds that are never equal.

Does this mean that this issue is actionable or not?

flip1995:
[...]

PR #10315, opened 2023-02-10
Make SpanlessEq more consistent - part 1, issue #10267

Yoric:
1. Get rid of the wildcard `_ => false` in `SpanlessEq::eq_expr`, as requested in #10267.
2. Make sure that `SpanlessEq::eq_expr` branches remain in alphabetical order.

I allowed myself to combine 1. and 2. since this was a style PR anyway. Feel free to reject PR if you wish me to return to the previous order, which mixes alphabetical and random.

Also, note the list of `=> false` at the end, which feels a little suspicious.

If your PR fixes an issue, you can add `fixes #issue_number` into this
your PR is merged.

---
changelog: none

### DIFF
--- a/clippy_utils/src/hir_utils.rs
+++ b/clippy_utils/src/hir_utils.rs
@@ -4,7 +4,7 @@ use crate::source::snippet_opt;
 use rustc_ast::ast::InlineAsmTemplatePiece;
 use rustc_data_structures::fx::FxHasher;
 use rustc_hir::def::Res;
-use rustc_hir::HirIdMap;
+use rustc_hir::{HirIdMap, YieldSource};
 use rustc_hir::{
     ArrayLen, BinOpKind, BindingAnnotation, Block, BodyId, Closure, Expr, ExprField, ExprKind, FnRetTy, GenericArg,
     GenericArgs, Guard, HirId, InlineAsmOperand, Let, Lifetime, LifetimeName, Pat, PatField, PatKind, Path,
@@ -205,6 +205,7 @@ impl HirEqInterExpr<'_, '_, '_> {
         res
     }
 
+    /// Determine whether two expressions are syntactically equal.
     #[expect(clippy::similar_names, clippy::too_many_lines)]
     pub fn eq_expr(&mut self, left: &Expr<'_>, right: &Expr<'_>) -> bool {
         if !self.inner.allow_side_effects && left.span.ctxt() != right.span.ctxt() {
@@ -226,6 +227,8 @@ impl HirEqInterExpr<'_, '_, '_> {
             reduce_exprkind(self.inner.cx, &left.kind),
             reduce_exprkind(self.inner.cx, &right.kind),
         ) {
+            // Style note: These patterns are ordered alphabetically.
+            // If you add new cases, please try and maintain this order!
             (&ExprKind::AddrOf(lb, l_mut, le), &ExprKind::AddrOf(rb, r_mut, re)) => {
                 lb == rb && l_mut == r_mut && self.eq_expr(le, re)
             },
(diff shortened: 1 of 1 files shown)


clippy_utils/src/hir_utils.rs
```diff
@@ -288,21 +290,54 @@ impl HirEqInterExpr<'_, '_, '_> {
...
+            | &ExprKind::Struct(..)
+            | &ExprKind::Tup(..)
+            | &ExprKind::Unary(..)
+
+            // Note: The following do NOT have a positive match.
+            | &ExprKind::ConstBlock(_)
+            | &ExprKind::Closure(_)
+            | &ExprKind::InlineAsm(_)
+            | &ExprKind::Yield(..)
+            | &ExprKind::Type(..)
```
flip1995, line 338: Those are the reason, for which I opened the issue.  Can you try to add an implementation for those? If you're unsure about something just ask and we can come up with something together.

clippy_utils/src/hir_utils.rs
```diff
@@ -288,21 +290,54 @@ impl HirEqInterExpr<'_, '_, '_> {
...
+            | &ExprKind::Index(..)
+            | &ExprKind::If(..)
+            | &ExprKind::Let(..)
+            | ExprKind::Lit(..)
+            | &ExprKind::Loop(..)
+            | &ExprKind::Match(..)
+            | &ExprKind::MethodCall(..)
+            | ExprKind::Path(..)
+            | &ExprKind::Repeat(..)
+            | ExprKind::Ret(..)
```
flip1995, line 328: Some of those don't have a `&`. I would add it for consistency.

flip1995 (commented):
Can you do the same for `eq_stmt` please? Should be way shorter and there shouldn't be anything missing.

flip1995:
And while you at it, and I just saw it in a Rust PR: Can you also do the same for `eq_ty`? :)

Sorry for baiting you with `eq_expr` and then asking for so much more. If you don't want to do all the extra work, that's totally understandable, and we can do that another time :+1:

clippy_utils/src/hir_utils.rs
Yoric: It did feel weird. Should we add them also in the positive cases?

Yoric:
> And while you at it, and I just saw it in a Rust PR: Can you also do the same for `eq_ty`? :)
> 
> Sorry for baiting you with `eq_expr` and then asking for so much more. If you don't want to do all the extra work, that's totally understandable, and we can do that another time +1

No problem. I won't have time to do everything today, but I'll try and do that by mid-next-week.

clippy_utils/src/hir_utils.rs
Yoric: I'm really not sure what definition of equality we're implementing. Can two closures even be equal? Or only if they have the same `def_id`?

clippy_utils/src/hir_utils.rs
flip1995: I wouldn't add them there. Probably Clippy will complain if you do :D

clippy_utils/src/hir_utils.rs
flip1995: An "implementation" for those that is just `(Closure, Closure) => false,` is fine. But I think 2 closures can be equal. E.g.:

```rust
let x = || 4 + 2;
let y = || 4 + 2;
```

The closures have the exact same expressions inside and the same argument count and everything. Only their span is different. Otherwise you could just remove `y` and use `x` everywhere instead.

flip1995: So we pretty much care about the literal content of the expressions if they are equal. (Disregarding formatting and spans)

flip1995:
Thanks a lot! Take your time :)

clippy_utils/src/hir_utils.rs
flip1995: I would just do this in one PR if you don't mind.

Yoric: > An "implementation" for those that is just `(Closure, Closure) => false,` is fine. But I think 2 closures can be equal. E.g.:
> 
> ```rust
> let x = || 4 + 2;
> let y = || 4 + 2;
> ```
> 
> The closures have the exact same expressions inside and the same argument count and everything. Only their span is different. Otherwise you could just remove `y` and use `x` everywhere instead.

Not sure I follow. Does this mean that closure comparison use De Bruijn indices? Feels a bit overkill for what looked like a style PR :)

Yoric:
Alright, here is one version with `eq_ty`. I have left a few comments and `FIXME` where I have doubts.

Yoric:
@flip1995 Gentle ping?

blyxyas:
@Yoric Are you still interested in working on this PR? I'm closing this for inactivity (207 since the last comment by a user), but feel free to re-open it if you are interested! (Maybe in a new branch since I'd suppose that `clippy_utils/src/hir_utils.rs` has changed a lot in that time) ฅ^•ﻌ•^ฅ

PR #10321, opened 2023-02-11
Fix false positives for `extra_unused_type_parameters`

### LINKED ISSUE #10318: extra_unused_type_parameters should not warn external macros
### Summary

extra_unused_type_parameters currently warns of code generated by external macros, but it should not. (there is no way for the user to fix it.)

Mentioning @mkrasnitski who implemented this lint in https://github.com/rust-lang/rust-clippy/pull/10028.

### Lint Name

extra_unused_type_parameters

### Reproducer

I tried this code:

```toml
# Cargo.toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
static_assertions = "=1.1.0"
```

```rust
// src/lib.rs
#![warn(clippy::extra_unused_type_parameters)]

#[allow(dead_code)]
struct S(());

static_assertions::assert_impl_all!(S: Send);
```

I saw this happen:

```
warning: type parameter goes unused in function definition
 --> src/lib.rs:6:1
  |
6 | static_assertions::assert_impl_all!(S: Send);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: consider removing the parameter
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::extra_unused_type_parameters)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  = note: this warning originates in the macro `static_assertions::assert_impl_all` (in Nightly builds, run with -Z macro-backtrace for more info)
```

I expected to see this happen: no warnings

### Version

```text
rustc 1.69.0-nightly (2773383a3 2023-02-10)
binary: rustc
commit-hash: 2773383a314a4b8f481ce2bed12c32de794ffbe9
commit-date: 2023-02-10
host: aarch64-apple-darwin
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

### LINKED ISSUE #10319: extra_unused_type_parameters probably should not warn empty functions
### Summary

It is a popular pattern to define an empty function with an unused type parameter with trait bounds to ensure that the type implements the trait.
extra_unused_type_parameters *probably* should not warn this pattern (at least by default).

- https://github.com/dtolnay/proc-macro2/blob/4822d05db17faa0639a3e0200aaf79cdacac16e0/tests/marker.rs#L12
- https://github.com/tokio-rs/tokio/blob/01bb1ecf4dee073da65dfede4c845ac27c555af0/tokio/src/lib.rs#L633
- https://github.com/tokio-rs/axum/blob/main/axum/src/test_helpers/mod.rs#L8-L10

Mentioning @mkrasnitski who implemented this lint in https://github.com/rust-lang/rust-clippy/pull/10028.

### Lint Name

extra_unused_type_parameters

### Reproducer

I tried this code:

```rust
#![warn(clippy::extra_unused_type_parameters)]

#[allow(dead_code)]
fn assert_send<T: Send>() {}

#[allow(dead_code)]
struct S(());

const _: fn() = || {
    assert_send::<S>();
};
```

I saw this happen:

```
warning: type parameter goes unused in function definition
 --> src/lib.rs:4:15
  |
4 | fn assert_send<T: Send>() {}
  |               ^^^^^^^^^
  |
  = help: consider removing the parameter
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::extra_unused_type_parameters)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I expected to see this happen: no warning

### Version

```text
rustc 1.69.0-nightly (2773383a3 2023-02-10)
binary: rustc
commit-hash: 2773383a314a4b8f481ce2bed12c32de794ffbe9
commit-date: 2023-02-10
host: aarch64-apple-darwin
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

mkrasnitski:
Don't lint external macros. Also, if the function body is empty, any type parameters with bounds on them are not linted. Note that only the body needs be empty - this rule still applies if the function takes any arguments.

fixes #10318
fixes #10319 
changelog: none

### DIFF
--- a/clippy_lints/src/extra_unused_type_parameters.rs
+++ b/clippy_lints/src/extra_unused_type_parameters.rs
@@ -4,10 +4,12 @@ use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::MultiSpan;
 use rustc_hir::intravisit::{walk_impl_item, walk_item, walk_param_bound, walk_ty, Visitor};
 use rustc_hir::{
-    GenericParamKind, Generics, ImplItem, ImplItemKind, Item, ItemKind, PredicateOrigin, Ty, TyKind, WherePredicate,
+    BodyId, ExprKind, GenericParamKind, Generics, ImplItem, ImplItemKind, Item, ItemKind, PredicateOrigin, Ty, TyKind,
+    WherePredicate,
 };
-use rustc_lint::{LateContext, LateLintPass};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::hir::nested_filter;
+use rustc_middle::lint::in_external_macro;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::{def_id::DefId, Span};
 
@@ -55,10 +57,13 @@ struct TypeWalker<'cx, 'tcx> {
     /// Otherwise, if any type parameters end up being used, or if any lifetime or const-generic
     /// parameters are present, this will be set to `false`.
     all_params_unused: bool,
+    /// Whether or not the function has an empty body, in which case any bounded type parameters
+    /// will not be linted.
+    fn_body_empty: bool,
 }
 
 impl<'cx, 'tcx> TypeWalker<'cx, 'tcx> {
-    fn new(cx: &'cx LateContext<'tcx>, generics: &'tcx Generics<'tcx>) -> Self {
+    fn new(cx: &'cx LateContext<'tcx>, generics: &'tcx Generics<'tcx>, body_id: BodyId) -> Self {
         let mut all_params_unused = true;
         let ty_params = generics
             .params
@@ -74,12 +79,18 @@ impl<'cx, 'tcx> TypeWalker<'cx, 'tcx> {
                 }
             })
             .collect();
+
+        let body = cx.tcx.hir().body(body_id).value;
+        let fn_body_empty =
+            matches!(&body.kind, ExprKind::Block(block, None) if block.stmts.is_empty() && block.expr.is_none());
+
         Self {
             cx,
             ty_params,
             bounds: FxHashMap::default(),
             generics,
             all_params_unused,
+            fn_body_empty,
         }
     }
 
@@ -96,7 +107,7 @@ impl<'cx, 'tcx> TypeWalker<'cx, 'tcx> {
             ),
         };
 
-        let source_map = self.cx.tcx.sess.source_map();
+        let source_map = self.cx.sess().source_map();
         let span = if self.all_params_unused {
             self.generics.span.into() // Remove the entire list of generics
         } else {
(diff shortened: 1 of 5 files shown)


flip1995:
@bors r+

Thanks!

flip1995:
This also makes me wonder if we should lint type params with bounds in the first place or if that is a bit too much for a warn-by-default lint. We could make this configurable.

Nominating for the next meeting.

mkrasnitski:
Sounds good - is `lintcheck` like a crater run but for clippy?

flip1995:
Yes, exactly. There is a predefined list in the `lintcheck` dir in the Clippy repo, but also https://github.com/matthiaskrgr/clippy-lintcheck which does things on a slightly bigger scale.

flip1995:
FP in CXX:

```rust
    /// `CxxString` is not constructible via `new`. Instead, use the
    /// [`let_cxx_string!`] macro.
    pub fn new<T: Private>() -> Self {
        unreachable!()
    }
```

`Private` is a public trait in a private module, so that it can't be implemented for external types. Maybe checking if a bound `is_exported` would address this FP.

---

Pretty much the same thing in `matrixmultiply`:

```rust
    /// Decide how many threads to use in each loop
    pub(crate) fn new<K>(m: usize, k: usize, n: usize, max_threads: usize) -> Self
        where K: GemmKernel
```

`GemmKernel` is a `pub(crate)` trait in a private module.

---

In `protobuf`:

```rust
    pub fn new<O>(name: &'static str) -> GeneratedOneofDescriptorData
    where
        O: OneofFull,
```

I think this can be removed. Maybe it is in generated code and that's why it is there? :thinking: 

However, this made me notice that this lint should respect the `avoid_breaking_exported_api` conv value and not lint on public / exported functions.

---

Tokio-io:

```rust
pub use self::async_read::AsyncRead;
pub use self::async_write::AsyncWrite;

fn _assert_objects() {
    fn _assert<T>() {}
    _assert::<Box<dyn AsyncRead>>();
    _assert::<Box<dyn AsyncWrite>>();
}
```

My best guess is, that this asserts that the `AsyncRead` and `AsyncWrite` traits are [object safe](https://doc.rust-lang.org/reference/items/traits.html#object-safety).

So maybe extend the fix in this PR to not only type params with bounds, but general params.

---

@mkrasnitski Will you get to fixing the above issues today / tomorrow before the sync? If not, I would move it to pedantic for now and then we will make it warn-by-default (`complexity`) one release cycle later.

---

I ran this lint on the most popular > 400 crates regarding downloads and dependencies that are out there. 
[crates.toml](https://github.com/rust-lang/rust-clippy/files/10802588/crates.toml.txt) < Those are the crates

mkrasnitski:
Thanks for finding these. I can take care of this today (don't want to wait another 6 weeks to have the fix land).

mkrasnitski:
I'm unsure how to get the required `LocalDefId` for each bounds to check if they're `is_exported`.  Would you happen to know?

mkrasnitski:
I don't think the `matrixmultiply` example is a false positive. Since `new` is also marked `pub(crate)`, I think the lint should apply there. There's a `#[cfg]` branch where K goes unused, so I think it's a valid application of the lint.

So, I think I should maybe check for cases where the function is exported but the bound is not - this by itself is pretty rare I would guess, since you can't just bound on a private trait, as that produces a compiler error saying you aren't allowed to leak the private trait. So, I'm surprised that you're still allowed to leak the trait if it's inside of a private module.

flip1995:
> So, I think I should maybe check for cases where the function is exported but the bound is not - this by itself is pretty rare I would guess

I think that would be overkill. Just checking if the bound is exported should be enough. Maybe this introduces a FN in this case, but as you said, this is pretty rare and then we err on the safe side.



> Should I be using their `TraitRef` in some way?

That sounds reasonable to me.

mkrasnitski:
Opened #10392 to address the new FPs.

PR #10328, opened 2023-02-12
Liberate late-bound regions rather than erasing them in `needless_pass_by_value`

compiler-errors:
changelog: [`needless_pass_by_value`]: fixes an ICE with when there are late-bound regions in function signatures

Fixes rust-lang/rust#107147
r? @matthiaskrgr


compiler-errors:
For the record, I have no idea how to make a clippy PR, or clippy UI test for that matter. Let me know if I did something wrong, otherwise I can always open this against the rust-lang/rust repo :^)

matthiaskrgr:
Looks good to me, yeah, basically just create foo.rs and foo.stderr  in in the test directory and that's basically it. :) 
Just two nites:
* I would like to see the rustc ticket referenced in the .rs file so we know what this test was about, a comment like `// https://github.com/rust-lang/rust/issues/107147` should do imo.
* The test can be moved into `tests/ui/crashes` actually

matthiaskrgr (approved):
Thanks!

PR #10333, opened 2023-02-12
lintcheck: fix clap panic

samueltardieu:
clap 4.1.4 panics if `-` is used at the start of an argument:

```
$ cargo lintcheck
[…]
thread 'main' panicked at 'Argument recursive: long "--recursive" must not start with a `-`, that will be handled by the parser', /home/sam/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-4.1.4/src/builder/debug_asserts.rs:82:13
```

changelog: none


flip1995:
@bors r+

Thanks!

PR #10336, opened 2023-02-12
manual_let_else: do not suggest semantically different replacements

### LINKED ISSUE #10241: `manual_let_else` suggests invalid replacement
### Summary

`manual_let_else` can propose replacements that change the behaviour of the code

Permalink to the reproducer: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b6719a7919c90a17b495f991566e423c, tested 2023-01-26

### Lint Name

manual_let_else

### Reproducer

I tried this code:

```rust

fn main() {
    let data = [1_u8, 2, 3, 4, 0, 0, 0, 0];
    
    // Rewriting as let-else would break this match because [data @ ..] (last pattern) matches everything.
    let data = match data.as_slice() {
        [] | [0, 0] => return,
        [data @ .., 0, 0, 0, 0] | [data @ .., 0, 0] | [data @ ..] => data,
    };
    
    dbg!(data);
}
```

I saw this happen:

```rust

fn main() {
    let data = [1_u8, 2, 3, 4, 0, 0, 0, 0];
    
    // Rewriting as let-else would break this match because [data @ ..] (last pattern) matches everything.
    let ([data @ .., 0, 0, 0, 0] | [data @ .., 0, 0] | [data @ ..]) = data.as_slice() else {
        return;
    };
    
    dbg!(data);
}
```

No replacement is proposed 

### Version

```text
- rust-clippy 0.1.68 (2023-01-26 d7948c8) (playground)
```

### Additional Labels

_No response_

samueltardieu:
The problem is that this lint does not consider the possibility that the divergent branch can come first and that the patterns may overlap. This led to incorrect suggestions, previously registered as correct in the tests themselves:

```rust
let v = match build_enum() {
    _ => continue,
    Variant::Bar(v) | Variant::Baz(v) => v,
};
```

had a `let Variant::Bar(v) | Variant::Baz(v) = v else { continue; }` suggestion, which is obviously wrong as the original code `continue`s in any case. Issue #10241 gives another example.

The code now checks that the divergent branch comes second. It could be extended later (I've added a TODO) to check for non-overlapping patterns.

Fixes #10241.

changelog: [`manual_let_else`] do not suggest non equivalent replacements in `match`

### DIFF
--- a/clippy_lints/src/manual_let_else.rs
+++ b/clippy_lints/src/manual_let_else.rs
@@ -115,6 +115,13 @@ impl<'tcx> LateLintPass<'tcx> for ManualLetElse {
                     .enumerate()
                     .find(|(_, arm)| expr_diverges(cx, arm.body) && pat_allowed_for_else(cx, arm.pat, check_types));
                 let Some((idx, diverging_arm)) = diverging_arm_opt else { return; };
+                // If the non-diverging arm is the first one, its pattern can be reused in a let/else statement.
+                // However, if it arrives in second position, its pattern may cover some cases already covered
+                // by the diverging one.
+                // TODO: accept the non-diverging arm as a second position if patterns are disjointed.
+                if idx == 0 {
+                    return;
+                }
                 let pat_arm = &arms[1 - idx];
                 if !expr_is_simple_identity(pat_arm.pat, pat_arm.body) {
                     return;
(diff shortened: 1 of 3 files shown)


llogiq:
Thanks, that makes sense.

@bors r+

PR #10338, opened 2023-02-12
Ignore synthetic type parameters for `extra_unused_type_parameters`

mkrasnitski:
There was a minor bug around calculating spans when forming the help message. An example:

```rust
fn unused_opaque<A, B>(dummy: impl Default) {}
//               ^^ ^
```

In this case, the entire list of generics should be highlighted, instead of each individual parameter. The culprit is the `impl Default`, which registers as a type parameter but doesn't live within the `<...>`. Because synthetic parameters can't ever be manually created, we just ignore them for this lint.

r? @flip1995 
changelog: none

### DIFF
--- a/clippy_lints/src/extra_unused_type_parameters.rs
+++ b/clippy_lints/src/extra_unused_type_parameters.rs
@@ -23,7 +23,6 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
-    /// // unused type parameters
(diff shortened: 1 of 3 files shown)


mkrasnitski:
Sorry about that - conflicts should be resolved now @flip1995

flip1995:
No worries! Thanks for resolving them so quickly.

@bors r+

Issue #10049, opened 2022-12-07
needless_return autofix causes compilation error with ambiguous `|` chain


ComputerDruid:
### Summary

When clippy's autofix removes the `return`, it breaks the compilation of expressions like `if true {1} else {0} | 2`

### Lint Name

needless_return

### Reproducer

I tried this code https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6026763a3a08d2c4c0dfb0eddc70b565

```rust
// from https://cs.opensource.google/fuchsia/fuchsia/+/master:src/lib/storage/vfs/rust/src/common.rs;l=149;drc=5faabf8b96d05e4d871caf12b7ab7313e8c63b2f
pub fn rights_to_posix_mode_bits(readable: bool, writable: bool, executable: bool) -> u32 {
    return if readable { libc::S_IRUSR } else { 0 }
        | if writable { libc::S_IWUSR } else { 0 }
        | if executable { libc::S_IXUSR } else { 0 };
}
```

I saw this happen:

```
warning: unneeded `return` statement
 --> src/lib.rs:3:5
  |
3 | /     return if readable { libc::S_IRUSR } else { 0 }
4 | |         | if writable { libc::S_IWUSR } else { 0 }
5 | |         | if executable { libc::S_IXUSR } else { 0 };
  | |____________________________________________________^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return
  = note: `#[warn(clippy::needless_return)]` on by default
  = help: remove `return`

warning: `playground` (lib) generated 1 warning (run `cargo fix --lib -p playground` to apply 1 suggestion)
    Finished dev [unoptimized + debuginfo] target(s) in 1.37s
```

Clippy identifies this as a needless_return, and suggests a `MachineApplicable` fix of:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1fadf1df63ff019f4c28d17e22d94cb4

```rust
pub fn rights_to_posix_mode_bits(readable: bool, writable: bool, executable: bool) -> u32 {
    if readable { libc::S_IRUSR } else { 0 }
        | if writable { libc::S_IWUSR } else { 0 }
        | if executable { libc::S_IXUSR } else { 0 }
}
```

Which fails to build with a bunch of errors including:

```
error[[E0308]](https://doc.rust-lang.org/stable/error-index.html#E0308): mismatched types
 --> src/lib.rs:4:9
  |
[...]

PR #10345, opened 2023-02-14
fix [`needless_return`] incorrect suggestion when returning if sequence

J-ZhengLi:
fixes: #10049 

---

changelog: [`needless_return`]: fix incorrect suggestion  on if sequence

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -14,6 +14,7 @@ use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::def_id::LocalDefId;
 use rustc_span::source_map::Span;
 use rustc_span::{BytePos, Pos};
+use std::borrow::Cow;
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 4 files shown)


clippy_lints/src/returns.rs
```diff
@@ -210,28 +214,46 @@ fn check_final_expr<'tcx>(
...
+                    return;
+                }
+
+                let (snippet, _) = snippet_with_context(
+                    cx,
+                    inner_expr.span,
+                    ret_span.ctxt(),
+                    "..",
+                    &mut Applicability::MachineApplicable,
+                );
```
xFrednet, line 238: The applicability might be modified by `snippet_with_context`. The resulting value should be stored, as it indicates if the snippet can be automatically applied. I'd suggest storing it in the `RetReplacement` enum along with the snippet to then use it for the emission in `emit_return_lint`

clippy_lints/src/returns.rs
```diff
@@ -253,29 +275,21 @@ fn check_final_expr<'tcx>(
...
-            let (snippet, _) = snippet_with_context(cx, inner_span, ret_span.ctxt(), "..", &mut applicability);
-            snippet.to_string()
-        },
-    );
-    let sugg_help = if inner_span.is_some() {
-        "remove `return`"
-    } else {
-        replacement.sugg_help()
-    };
+    let applicability = Applicability::MachineApplicable;
```
xFrednet, line 290: This should be the applicability, from the snippet. See comment above

clippy_lints/src/returns.rs
```diff
@@ -253,29 +275,21 @@ fn check_final_expr<'tcx>(
...
-    semi_spans: Vec<Span>,
-    inner_span: Option<Span>,
-    replacement: RetReplacement,
-) {
+fn expr_contains_if<'tcx>(expr: &'tcx Expr<'tcx>) -> bool {
+    match expr.kind {
+        ExprKind::If(..) => true,
+        ExprKind::Binary(_, left, right) => expr_contains_if(left) || expr_contains_if(right),
+        _ => false,
+    }
```
xFrednet, line 283: This will now also suggest brackets, if only a single if statement is present, like

```
fn b() -> u32 {
    return if true {1} else {2};
}
```

Could you adjust the recursion in a way, that it only returns true if they're nested in binary expressions? I think the simplest option would be to pass in a bool value `on_if` that is returned for `ExprKind::If(..)`

Could you than also add the example above as a test?

xFrednet (commented):
Hey, overall, this already looks good. Three small comments, and then this should be ready for merge :)

clippy_lints/src/returns.rs
J-ZhengLi: Thank you for pointing that out~ your are right, my code missed that case.

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the changes :)

@bors r+

Issue #10272, opened 2023-02-01
ifs_same_cond bails out on equal methods that return bool 


matthiaskrgr:
### Summary

apparently we detect  boolan expressions of the form `x == y`  but we don't detect `fn(&self) -> bool` ones???
I get it that this can be iffy if sideeffects are involved but in this example the fn does not even need `&mut self` :/

Could we somehow warn here  for const-fns at least maybe?

### Lint Name

ifs_same_cond

### Reproducer

I tried this code:

```rust
pub fn main() {
    let a = Some(32);
    if let Some(32) = a {
        let b = String::from("x");
        if b.contains("ah") { // same as
            eprintln!("AAA");
        } else if b.contains("ah") { // here
            println!("AAA");
        }
    }
}

```

I expected to see this happen:
clippy warning

Instead, this happened:

### Version

```text
rustc 1.69.0-nightly (dc1d9d50f 2023-01-31)
binary: rustc
commit-hash: dc1d9d50fba2f6a1ccab8748a0050cde38253f60
commit-date: 2023-01-31
host: x86_64-unknown-linux-gnu
release: 1.69.0-nightly
LLVM version: 15.0.7
```

J-ZhengLi:
claim first, question it later~ @rustbot claim

J-ZhengLi:
just want to confirm one thing before I start: what will be the side-effect even if the caller wasn't declared mutable?
Seems like there's already a [test case here](https://github.com/rust-lang/rust-clippy/blob/5b6795f50b521b11393d39f823d8a3e5b262b134/tests/ui/ifs_same_cond.rs#L40-L43) in the file, but for some reason the person though it will be ok to leave it as is, so I'm not sure...

J-ZhengLi:
well, here comes the question 🤣:
If there's a mut pointer, and most of its methods are `const fn`, should the same method calls trigger this lint? like:

```rust
let p: *mut i8 = std::ptr::null_mut();
if p.is_null(){
    unsafe { p.write(..) }
}
else if p.is_null{..}
else {..}
```

PR #10350, opened 2023-02-16
enhance [`ifs_same_cond`] to warn same immutable method calls as well

J-ZhengLi:
fixes: #10272 

---

changelog: Enhancement: [`ifs_same_cond`]: Now also detects immutable method calls.
[#10350](https://github.com/rust-lang/rust-clippy/pull/10350)


J-ZhengLi:
feel very insecure about this :man_facepalming: , is there any side-effect of doing this that I'm not aware of?

PatchMixolydic:
I'm not familiar with the method behind this check, but if it only checks the mutability of the binding, this might lead to false positives with interior mutable types:
```rust
use std::cell::Cell;

fn main() {
    let x = Cell::new(true);

    if !x.get() {
        unreachable!("x should be true");
    } else if !x.take() {
        unreachable!("x should've been true");
    } else if !x.get() {
        println!("OK");
    } else {
        unreachable!("x should be false");
    }
}
```

(Of course, it's possible that this is covered by `SpanlessEq::eq_expr` and I completely missed it ^^;)

J-ZhengLi:
> I'm not familiar with the method behind this check, but if it only checks the mutability of the binding, this might lead to false positives with interior mutable types:
> 
> ```rust
> use std::cell::Cell;
> 
> fn main() {
>     let x = Cell::new(true);
> 
>     if !x.get() {
>         unreachable!("x should be true");
>     } else if !x.take() {
>         unreachable!("x should've been true");
>     } else if !x.get() {
>         println!("OK");
>     } else {
>         unreachable!("x should be false");
>     }
> }
> ```
> 
> (Of course, it's possible that this is covered by `SpanlessEq::eq_expr` and I completely missed it ^^;)

Thank you~ I just tested your example to the test but it didn't trigger any warning, so I guess eq_expr somehow does cover this scenario. 😄

xFrednet:
Hey, thank you for the PR. I would have also brought up the point of inner mutability. Thank you, @PatchMixolydic, for the comment.

I see three solutions:
1. Scan the called method for inner mutability. This will only work for local methods, as external ones are not loaded by rustc AFAIK, and will be complex with little benefit.
2. Create an allow list, which would avoid false positives, but would also be less flexible
3. Add a configuration value that lints these methods by default, but can be turned off, if the project uses a lot of inner mutability.

From these options, I recommend the last one. You can take a look at this link, if you want to implement this solution: https://doc.rust-lang.org/clippy/development/adding_lints.html#adding-configuration-to-a-lint

You are welcome to reach out if you have any questions :)

J-ZhengLi:
> Hey, thank you for the PR. I would have also brought up the point of inner mutability. Thank you, @PatchMixolydic, for the comment.
> 
> I see three solutions:
> 
>     1. Scan the called method for inner mutability. This will only work for local methods, as external ones are not loaded by rustc AFAIK, and will be complex with little benefit.
> 
>     2. Create an allow list, which would avoid false positives, but would also be less flexible
> 
>     3. Add a configuration value that lints these methods by default, but can be turned off, if the project uses a lot of inner mutability.
> 
> 
> From these options, I recommend the last one. You can take a look at this link, if you want to implement this solution: https://doc.rust-lang.org/clippy/development/adding_lints.html#adding-configuration-to-a-lint
> 
> You are welcome to reach out if you have any questions :)

well I do have a question, what if get the type of the caller instead and skip the check of `Cell`, `RefCell`, `*mut` types completely, would that be achievable?

xFrednet:
>  what if get the type of the caller instead

I believe this is already how the lint operates. I thought the question was about structs with field with internal mutability.

> **update**: I noticed that [`mutable_key_type`](https://rust-lang.github.io/rust-clippy/master/index.html#mutable_key_type) already got a configuration called `ignore_interior_mutability`, which seems allowing people to "whitelisting" certain types, maybe I could try using the same conf directly.

That sounds like a perfect approach :+1:

clippy_lints/src/copies.rs
```diff
@@ -547,9 +569,47 @@ fn check_for_warn_of_moved_symbol(cx: &LateContext<'_>, symbols: &[(HirId, Symbo
...
+        true
+    } else {
+        false
+    };
+
+    if is_ignored_ty
+        || caller_ty.is_mutable_ptr()
+        || path_to_local(caller_expr)
+            .and_then(|hid| find_binding_init(cx, hid))
+            .is_none()
```
xFrednet, line 588: Could you explain this check? I understand that it checks if the caller is a local value and if an init expression is present. But why should this be ignored if this is the case?

clippy_lints/src/copies.rs
```diff
@@ -547,9 +569,47 @@ fn check_for_warn_of_moved_symbol(cx: &LateContext<'_>, symbols: &[(HirId, Symbo
...
+    {
+        return true;
+    }
+
+    false
```
xFrednet, line 593: NIT: You can replace the `if` and this false with this:

```suggestion
    is_ignored_ty
        || caller_ty.is_mutable_ptr()
        || path_to_local(caller_expr)
            .and_then(|hid| find_binding_init(cx, hid))
            .is_none()
```

tests/ui-toml/ifs_same_cond/clippy.toml
```diff
@@ -0,0 +1 @@
+ignore-interior-mutability = ["std::cell::Cell"]
```
xFrednet, line 1: Could you add a new line at the end of this file?
```suggestion
ignore-interior-mutability = ["std::cell::Cell"]

```

clippy_lints/src/utils/conf.rs
```diff
@@ -437,7 +437,7 @@ define_Conf! {
     ///
     /// The maximum size of the `Err`-variant in a `Result` returned from a function
     (large_error_threshold: u64 = 128),
-    /// Lint: MUTABLE_KEY_TYPE.
+    /// Lint: MUTABLE_KEY_TYPE, IFS_SAME_COND.
     ///
     /// A list of paths to types that should be treated like `Arc`, i.e. ignored but
```
xFrednet, line 442: The docs imply to me that `Arc` and other pointer types like `Cell` are included in the list, even if they aren't part of the default configuration. Do you maybe want to add them?

xFrednet (commented):
The implementation looks good overall. I have some small suggestions regarding the style and type resolution, but they should hopefully be simple to fix. :)

clippy_lints/src/copies.rs
```diff
@@ -170,7 +185,14 @@ impl<'tcx> LateLintPass<'tcx> for CopyAndPaste {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
         if !expr.span.from_expansion() && matches!(expr.kind, ExprKind::If(..)) && !is_else_clause(cx.tcx, expr) {
             let (conds, blocks) = if_sequence(expr);
-            lint_same_cond(cx, &conds);
+            let mut ignored_ty_ids = FxHashSet::default();
+            for ignored_ty in &self.ignore_interior_mutability {
+                let path: Vec<&str> = ignored_ty.split("::").collect();
+                for id in def_path_def_ids(cx, path.as_slice()) {
```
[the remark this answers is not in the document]
J-ZhengLi: Ohhh oops, you are right, I forgot that `check_expr` runs for every expression it founds

clippy_lints/src/copies.rs
J-ZhengLi: Because the documentation of `find_binding_init` says: `... Return None if the binding is mutable...`, I thought it was pretty convenient lol, unless there are more optimized way to check whether a variable is mutable or not

tests/ui-toml/ifs_same_cond/clippy.toml
J-ZhengLi: Done~

clippy_lints/src/utils/conf.rs
J-ZhengLi: you mean adding Arc. Mutex, RWLock, Cell, RefCell into the list as well?

J-ZhengLi: Oh wait, I think did the opposite of what that config was designed to do... those types in the list were supposed to be checked.

I got confused by the name lol

J-ZhengLi: Oh wait, I think I did the opposite of what that config was designed to do... those types in the list were supposed to be checked.

I got confused by the name lol

clippy_lints/src/copies.rs
xFrednet: Ah, okay, that's interesting. I understand why. I guess this is a good heuristic. It will have some false negatives, but those are better than false positives :)

clippy_lints/src/utils/conf.rs
xFrednet: Now that you say it, it's a bit confusing :sweat_smile:. I guess we'll leave it like this now, since it's not new.

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the change and nice refactoring. :)

@bors r+

J-ZhengLi:
Ahh, looks like the some book test is failing, that's new for me, what should I do 😱

xFrednet:
Right, we've added that recently, and I still forget about mentioning that. Luckily we have bors as our bouncer ^^. Thank you for fixing the CI :)

@bors r+

Issue #6556, opened 2021-01-07
unusual_byte_groupings fires for octal values represented in binary


koutheir:
Lint name: unusual_byte_groupings

I tried this code:

```rust
let unix_mode = 0b111_101_101; // rwx r-x r-x
```

I didn't expect to see the warning `digits of hex or binary literal not grouped by four`, because grouping by 3 bits is very reasonable when binary is clearer than octal.

### Meta

- `cargo clippy -V`:
  ```
  clippy 0.0.212 (e1884a8 2020-12-29)
  ```
- `rustc -Vv`:
  ```
  rustc 1.49.0 (e1884a8e3 2020-12-29)
  binary: rustc
  commit-hash: e1884a8e3c3e813aada8254edfa120e85bf5ffca
  commit-date: 2020-12-29
  host: x86_64-unknown-linux-gnu
  release: 1.49.0
  ```

camsteffen:
Hmm I guess we'll just add a special case for 3 groups of 3? Usually the groups don't have semantic meaning like they do in a unix mode.

koutheir:
I think this doesn't depend on any semantic meaning. Grouping by four bits is a specification of one hexadecimal digit in the binary numeral system. Likewise, grouping by three bits is a specification of one octal digit in binary. The octal numeral system is used often enough to make this warning unexpected in computer science.

mohanson:
Can this check be turned off now? I think this check should not be added, whether it is a 4-bit grouping or a 3-bit grouping.

In virtual machine development, we often need to encode/decode an instruction, and the instruction encoding usually has its own format, such as RISC-V ISA: 

```
// |31|30|29|28|27|26|25|24|23|22|21|20|19|18|17|16|15|14|13|12|11|10|09|08|07|06|05|04|03|02|01|00|
// |funct7              |rs2           |rs1           |funct3  |rd            |opcode              | R-type
// |rs3           |funct|rs2           |rs1           |funct3  |rd            |opcode              | R4-type
// |imm[11:0]                          |rs1           |funct3  |rd            |opcode              | I-type
// |imm[11:5]           |rs2           |rs1           |funct3  |imm[4:0]      |opcode              | S-type
// |imm[31:12]                                                 |rd            |opcode              | U-type
// |12|imm[10:5]        |rs2           |rs1           |funct3  |imm[4:1]   |11|opcode              | B-type
// |20|imm[10:1]                    |11|imm[19:12]             |rd            |opcode              | J-type
```

[...]

PR #10353, opened 2023-02-16
Change unusual_byte_groupings to require byte groupings of equal size

nindalf:
Fixes issue #6556

This lint required byte groupings of size 2 or 4 for `Radix::Binary` and `Radix::Hexadecimal`. Since there are good reasons for allowing groups of other sizes, this PR relaxes the restriction. This lint now requires that 

- group sizes after the first group be of the same size and
- greater or equal in size to the first group.

---

changelog: [`unusual_byte_groupings`]: reduce false positives by relaxing restriction requiring groups of specific sizes.


clippy_lints/src/literal_representation.rs
```diff
@@ -210,7 +210,7 @@ impl WarningType {
                 cx,
                 UNUSUAL_BYTE_GROUPINGS,
                 span,
-                "digits of hex or binary literal not grouped by four",
+                "digits of hex, binary or octal literal not groups of equal size",
```
llogiq, line 213:
```suggestion
                "digits of hex, binary or octal literal not in groups of equal size",
```

clippy_lints/src/literal_representation.rs
```diff
@@ -427,8 +427,12 @@ impl LiteralDigitGrouping {
 
         let first = groups.next().expect("At least one group");
 
-        if (radix == Radix::Binary || radix == Radix::Hexadecimal) && groups.any(|i| i != 4 && i != 2) {
-            return Err(WarningType::UnusualByteGroupings);
+        if radix == Radix::Binary || radix == Radix::Octal || radix == Radix::Hexadecimal {
```
llogiq, line 430:
```suggestion
        if radix != Radix::Decimal {
```

nindalf: I think this keeps the intention clearer.

llogiq:
Thank you! @bors r+

PR #10356, opened 2023-02-16
Add `let_underscore_untyped`

### LINKED ISSUE #6842: Warn about blanket `let _ = `
### What it does

*What does this lint do?*

Suggest replacing `let _ = <expr>` with `let _: <type> = <expr>`

### Categories (optional)

- Kind: Tentatively `clippy::pedantic`

It's somewhat common to use `let _ = <expr>` to ignore results that would otherwise generate a warning when not used. e.g. ignore an unimportant error.

However, if changes elsewhere cause the type of `<expr>` to change, bugs could be silently introduced, as `let _ =` would accept any type.

For example, `<expr>` might go from `Result<_>` to `impl Future<Output=Result<_>>`, because an upstream crate changed a function to `async fn`. And this could cause some operation to never be performed.

Asking the user to always specify the type in `let _ =` would avoid this kind of bugs.

### Drawbacks

Slightly more verbose code.

### Example

```rust
let _ = do_something_and_ignore_error();
```

Could be written as:

```rust
let _: Result<_, _> = do_something_and_ignore_error();
```
### Alternatives

1. Warn when an `impl Future` is dropped in `let _ =`, it almost never makes sense to construct a future then immediately drop it.
2. Suggest using `<expr>.ok()` instead of `let _ = <expr>`, for the case where `<expr>` is `Result<_, _>`

Either option works for the example above, but I think my suggestion is more general.

JirkaVebr:
Fixes #6842

This adds a new pedantic `let_underscore_untyped` lint which checks for `let _ = <expr>`, and suggests to either provide a type annotation, or to remove the `let` keyword. That way the author is forced to specify the type they intended to ignore, and thus get forced to re-visit the decision should the type of `<expr>` change. Alternatively, they can drop the `let` keyword to truly just ignore the value no matter what.

r? @llogiq

changelog: New lint: [let_underscore_untyped]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4494,6 +4494,7 @@ Released 2018-09-13
 [`let_underscore_future`]: https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_future
 [`let_underscore_lock`]: https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_lock
 [`let_underscore_must_use`]: https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_must_use
+[`let_underscore_untyped`]: https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_untyped
 [`let_unit_value`]: https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value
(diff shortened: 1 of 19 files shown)


llogiq:
Thank you!

@bors r+

PR #10358, opened 2023-02-16
Add `unnecessary_literal_unwrap` lint

### LINKED ISSUE #10352: Unnecessary direct unwraps on Option and Result are not warned/fixed
### Summary

When we have direct unwraps on an `Option` or `Result` like the following:

```rust
let val = Some(1).unwrap();
let val = Ok(1).unwrap();
let val = Err(1).unwrap_err();
let val = Ok(1).expect(...);
```

The lint remove the unnecessary unwrap/expect and unnecessary wraps from the above lines.

```rust
let val = 1;
let val = 1;
let val = 1;
let val = 1;
```

### Lint Name

unnecessary_unwrap

### Reproducer

I tried this code:

```rust
let val = Some(1).unwrap();
```

```rust
let val = 1;
```

Instead, this happened:

Nothing changed.

### Version

_No response_

pksunkara:
Add lint for more unnecessary unwraps and suggest fixes for them.

Fixes #10352 

- [x] Run `cargo dev fmt`

r? @llogiq 

---

changelog: New lint [`unnecessary_literal_unwrap`]
[#10358](https://github.com/rust-lang/rust-clippy/pull/10358)


llogiq:
The `unwrap_or_else_default` UI test is still failing.

samueltardieu:
> The `unwrap_or_else_default` UI test is still failing.

As well as many others. I count differences in 21 output files when testing locally.

pksunkara:
Only the documentation is left. But before I do that, I would like to know 2 things:

1. How do I make this lint warn by default?
2. What should I put in place of `#[clippy::version = "1.45.0"]` in `declare_lint!` usage?

clippy_lints/src/methods/mod.rs
```diff
@@ -271,6 +272,56 @@ declare_clippy_lint! {
...
+    /// ```rust,ignore
+    /// # let option = Some(1);
+    /// # let result: Result<usize, ()> = Ok(1);
+    /// option?;
+    ///
+    /// // or
+    ///
+    /// result?;
+    /// ```
+    #[clippy::version = "1.45.0"]
```
samueltardieu, line 319: That should be 1.69.0

clippy_lints/src/methods/mod.rs
```diff
@@ -271,6 +272,56 @@ declare_clippy_lint! {
...
+    pub UNNECESSARY_LITERAL_UNWRAP,
+    complexity,
+    "checks for calls of `unwrap()` or `expect()` on `Some()` that cannot fail"
```
samueltardieu, line 322: Isn't `Ok()` covered as well?

tests/ui/unnecessary_literal_unwrap.rs
```diff
@@ -0,0 +1,49 @@
...
+    let _val = Some(1).expect("this never happens");
+}
+
+fn unwrap_option_none() {
+    None::<usize>.unwrap();
+    None::<usize>.expect("this always happens");
+}
+
+fn unwrap_result_ok() {
+    let _val = Ok::<usize, ()>(1).unwrap();
```
samueltardieu, line 17: What if the `usize` was needed in this context? Wouldn't suggesting `let _val = 1;` loose that information? Could you include examples in which the type information is needed, such as:

```rust
fn context_needed() {
    let _s = Some::<usize>([1, 2, 3].iter().sum()).unwrap();
}
```

as the current suggestion will not compile as-is.

llogiq: Agreed. If the `Some` or `Ok` have given generics (from the hir::Ty, not the ty::Ty), the first of them should be added to the suggestion as type of the binding.

llogiq:
@pksunkara do you intend to revisit this at some point?

pksunkara:
Yup, I can take a look on Monday.

pksunkara:
I would appreciate if I can get some answers for the questions I asked earlier, https://github.com/rust-lang/rust-clippy/pull/10358#issuecomment-1433883966

llogiq:
> 1. How do I make this lint warn by default?
> 2. What should I put in place of #[clippy::version = "1.45.0"] in declare_lint! usage?

1. Do you mean in a test? That would be `#[warn(clippy::unnecessary_literal_unwrap)]`. Otherwise the default lint level is given by the lint group. As this lint is in `complexity`, that should IIRC warn by default.
2. `#[clippy::version = "1.69.0"]`.

tests/ui/unnecessary_literal_unwrap.rs
pksunkara: I am stuck trying to address the review comments. I have an `Expr` from `Expr::call` [here](https://github.com/rust-lang/rust-clippy/pull/10358/files#diff-bf1cd10676ad7ed2fa17ccd518ddaaba2f1a54859f8cbd9f6745d87bc9ef99adR18), but how do I get the generic arguments (`hir::Ty`) from that?

pksunkara:
I have addressed the review comments.

@rustbot r?

blyxyas:
@pksunkara The comment to remove the `L-waiting-on-author` is [rustbot ready](https://github.com/rust-lang/triagebot/wiki/Shortcuts)

llogiq:
There are still merge conflicts, and I don't have time to review right now, sorry. Perhaps I'll find a few minutes tomorrow.

pksunkara:
> There are still merge conflicts

They are very minor ones because of this PR touching a lot of tests and I will fix them once the code is reviewed.

llogiq (approved):
Looks good, I only have a small nit regarding docs.

Also I'm not completely sure about the name: So far we have 23 `needless-*` and 18 `unnecessary-*` lints. I like the former more for both brevity and being easier to write.

pksunkara:
Addressed the conflicts and review comment.

llogiq:
So you intend to keep the name longer. Ah, well, you've done enough rebases as is. Let's roll with it. Thanks for seeing this PR through.

@bors r+

PR #10359, opened 2023-02-16
Include async functions in the len_without_is_empty

### LINKED ISSUE #7232: len_without_is_empty triggers for async len methods, but does not accept async is_empty methods
The following will trigger the `len_without_is_empty` error:

```rust
impl Foo {
    pub async fn len(&self) -> usize { ... }
    pub async fn is_empty(&self) -> bool { ... }
}
```

It probably shouldn't expect the exact signature `fn(&self) -> bool`.

mladedav:
fixes #7232

Changes done to the functionality:

Allowing different error types for the functions was disallowed. So the following was linted before but is not after this change
```
impl Foo {
    pub len(&self) -> Result<usize, Error1> { todo!(); }
    pub is_empty(&self) -> Result<bool, Error2> { todo!(); }
}
```

---

changelog: Enhancement: [`len_without_is_empty`]: Now also detects `async` functions
[#10359](https://github.com/rust-lang/rust-clippy/pull/10359)

### DIFF
--- a/clippy_lints/src/len_zero.rs
+++ b/clippy_lints/src/len_zero.rs
@@ -6,8 +6,9 @@ use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
 use rustc_hir::def_id::DefIdSet;
 use rustc_hir::{
-    def_id::DefId, AssocItemKind, BinOpKind, Expr, ExprKind, FnRetTy, ImplItem, ImplItemKind, ImplicitSelfKind, Item,
-    ItemKind, Mutability, Node, TraitItemRef, TyKind,
+    def::Res, def_id::DefId, lang_items::LangItem, AssocItemKind, BinOpKind, Expr, ExprKind, FnRetTy, GenericArg,
+    GenericBound, ImplItem, ImplItemKind, ImplicitSelfKind, Item, ItemKind, Mutability, Node, PathSegment, PrimTy,
+    QPath, TraitItemRef, TyKind, TypeBindingKind,
 };
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty::{self, AssocKind, FnSig, Ty};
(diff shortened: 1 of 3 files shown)


llogiq:
fmt and dogfood are still failing; those should be easy to fix.

@rustbot author

mladedav:
The tests are fixed now, hope everything looks fine

@llogiq

clippy_lints/src/len_zero.rs
```diff
@@ -251,33 +252,133 @@ fn check_trait_items(cx: &LateContext<'_>, visited_trait: &Item<'_>, trait_items
...
-fn parse_len_output<'tcx>(cx: &LateContext<'_>, sig: FnSig<'tcx>) -> Option<LenOutput<'tcx>> {
+
+fn extract_future_output<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> Option<&'tcx PathSegment<'tcx>> {
+    if let ty::Alias(_, alias_ty) = ty.kind() {
+        let def_id = alias_ty.def_id;
+        let hir_map = cx.tcx.hir();
+        let item = hir_map.get_if_local(def_id);
+        let item = item.unwrap();
+
+        if let Node::Item(item) = item {
```
samueltardieu, line 268: or could use a unique `if let …` here since you are really matching deeper and deeper with one possible pattern.

clippy_lints/src/len_zero.rs
```diff
@@ -251,33 +252,133 @@ fn check_trait_items(cx: &LateContext<'_>, visited_trait: &Item<'_>, trait_items
...
+            return Some(LenOutput::Integral);
+        }
+
+        if let Res::Def(_, def_id) = res {
+            if cx.tcx.is_diagnostic_item(sym::Option, def_id) {
+                if is_first_generic_integral(segment) {
+                    return Some(LenOutput::Option(def_id));
+                }
+
+                return Some(LenOutput::Option(def_id));
```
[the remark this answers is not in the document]
mladedav: Sorry, this is a bug. `is_first_generic_integral` checks that `len()` returns `Option<T>` where T is an integral type. This should ignore things like `pub fn len() -> Option<String>` because that's what happens for the sync version.

I've added a test for this.

clippy_lints/src/len_zero.rs
mladedav: I rewrote this with an if let without nesting. However, it can't be a unique `if let` since at times I want to check some bounds before indexing into an array.

I have collapsed the chains as much as possible though. I personally don't like that much (e.g. I'd rather split
```
        let Some(Node::Item(Item { kind: ItemKind::OpaqueTy(opaque), .. })) = cx.tcx.hir().get_if_local(alias_ty.def_id) &&
        ...
```
into
```
        let Some(Node::Item(item)) = cx.tcx.hir().get_if_local(alias_ty.def_id) &&
        let Item { kind: ItemKind::OpaqueTy(opaque), .. } = item &&
        ...
```
but if you prefer it like this I have no issue with that.

mladedav:
@llogiq

Hi, do you think you will have some time for this PR? I'd like to finish this soI don't forget about it.

clippy_lints/src/len_zero.rs
```diff
@@ -251,33 +252,133 @@ fn check_trait_items(cx: &LateContext<'_>, visited_trait: &Item<'_>, trait_items
...
+enum LenOutput {
     Integral,
     Option(DefId),
-    Result(DefId, Ty<'tcx>),
+    Result(DefId),
 }
-fn parse_len_output<'tcx>(cx: &LateContext<'_>, sig: FnSig<'tcx>) -> Option<LenOutput<'tcx>> {
+
+fn extract_future_output<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> Option<&'tcx PathSegment<'tcx>> {
+    if let ty::Alias(_, alias_ty) = ty.kind() {
```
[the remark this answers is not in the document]
llogiq: Nowadays we can even use let chains.

clippy_lints/src/len_zero.rs
```diff
@@ -251,33 +252,98 @@ fn check_trait_items(cx: &LateContext<'_>, visited_trait: &Item<'_>, trait_items
...
 
-impl<'tcx> LenOutput<'tcx> {
-    fn matches_is_empty_output(self, ty: Ty<'tcx>) -> bool {
+impl LenOutput {
+    fn matches_is_empty_output<'tcx>(self, cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
+        if let Some(segment) = extract_future_output(cx, ty) {
+            return match (self, segment.res) {
+                (_, Res::PrimTy(PrimTy::Bool)) => true,
+                (Self::Option(_), Res::Def(_, def_id)) if cx.tcx.is_diagnostic_item(sym::Option, def_id) => true,
+                (Self::Result(_), Res::Def(_, def_id)) if cx.tcx.is_diagnostic_item(sym::Result, def_id) => true,
```
llogiq, line 338:
```suggestion
                (Self::Option(_), Res::Def(_, def_id)) => cx.tcx.is_diagnostic_item(sym::Option, def_id),
                (Self::Result(_), Res::Def(_, def_id)) => cx.tcx.is_diagnostic_item(sym::Result, def_id),
```
I think we don't need guards here.

llogiq (approved):
r=me after rebase

mladedav:
r= @llogiq

Issue #6207, opened 2020-10-23
Lint `assert!` without message


HMPerson1:
### What it does

Lint `{debug_}assert{_eq,_ne}!` without a custom panic message. (inspired by [this tweet](https://twitter.com/myrrlyn/status/1319672781324324867))

### Categories (optional)

- Kind: pedantic

*What is the advantage of the recommended code over the original code*

If the assertion fails, it might be easier to diagnose why.

### Drawbacks

None.

### Example

```rust
assert!(some_condition(foo));
debug_assert_eq(a, bar(b));
```

Could be written as:

```rust
assert!(some_condition(foo), "foo failed some condition: foo = {}", foo);
debug_assert_eq!(a, bar(b), "failed to find inverse of bar at {}", a);
```

Issue #6610, opened 2021-01-20
Clippy ignores crate wide warn/deny


tcharding:
I put this code in `main.rs`

```rust
#![deny(clippy::print_stdout)]

mod foo;

#[allow(clippy::print_stdout)]
fn main() {
    println!("Hello, world!");
}
```
And this code in `foo.rs`
```rust
pub fn foo() {
    println!("This should trigger clippy");
}
```
I expected to see clippy warn something like
```

error: use of `println!`
 --> src/foo.rs:2:5
  |
6 |     println!("This should trigger clippy");
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
```
Instead all I get is 
```
warning: function is never used: `foo`
  ```

### Meta
```
cargo clippy -V                                                                                                            ✘ 101 
clippy 0.1.51 (4253153 2021-01-17)
rustc -Vv
rustc 1.51.0-nightly (4253153db 2021-01-17)
binary: rustc
commit-hash: 4253153db205251f72ea4493687a31e04a2a8ca0
commit-date: 2021-01-17
host: x86_64-unknown-linux-gnu
release: 1.51.0-nightly
LLVM version: 11.0.1
  ```

giraffate:
It seems to be similar to https://github.com/rust-lang/rust-clippy/issues/5356.

flip1995:
Yes this is the same underlying issue. The solution is to get rid of `pre_expansion_pass` lints.

davepacheco:
I think I ran into this too and reproduced it here: https://github.com/davepacheco/clippy-issue

@giraffate and @flip1995 you said this is the same underlying issue as #5356, but that one looks like it's been fixed.  Do you expect this issue is fixed too?  I'm still seeing it on clippy 0.0.212 (e1884a8e 2020-12-29) (Rust stable 1.49.0).

This doesn't always seem to be a problem and I'm not sure why not.  I have a different crate (dropshot) that uses a crate-wide `#![allow(clippy::style)]` that appears to work.

flip1995:
It's the same underlying issue, but requires a separate fix, which involves quite a bit of work. TL;DR the whole lint has to be rewritten.

davepacheco:
Thanks for clarifying!

jszwedko:
In case it helps others, as a work around I ended up adding these lints to `.cargo/config.toml`:

```toml
[target.'cfg(feature = "cargo-clippy")']
rustflags = [
[...]

PR #10362, opened 2023-02-16
Add `missing_assert_message` lint

unexge:
Fixes https://github.com/rust-lang/rust-clippy/issues/6207.

changelog: new lint: [`missing_assert_message`]: A new lint for checking assertions that doesn't have a custom panic message.
[#10362](https://github.com/rust-lang/rust-clippy/pull/10362)

r? @llogiq

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -927,7 +927,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     });
     store.register_late_pass(|_| Box::new(no_mangle_with_rust_abi::NoMangleWithRustAbi));
     store.register_late_pass(|_| Box::new(collection_is_never_read::CollectionIsNeverRead));
-    store.register_pre_expansion_pass(|| Box::new(missing_assert_message::MissingAssertMessage));
+    store.register_pre_expansion_pass(|| Box::<missing_assert_message::MissingAssertMessage>::default());
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -193,6 +193,7 @@ mod minmax;
 mod misc;
 mod misc_early;
 mod mismatching_type_param_order;
+mod missing_assert_message;
 mod missing_const_for_fn;
 mod missing_doc;
 mod missing_enforced_import_rename;
@@ -926,6 +927,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     });
     store.register_late_pass(|_| Box::new(no_mangle_with_rust_abi::NoMangleWithRustAbi));
     store.register_late_pass(|_| Box::new(collection_is_never_read::CollectionIsNeverRead));
+    store.register_pre_expansion_pass(|| Box::new(missing_assert_message::MissingAssertMessage));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -927,7 +927,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     });
     store.register_late_pass(|_| Box::new(no_mangle_with_rust_abi::NoMangleWithRustAbi));
     store.register_late_pass(|_| Box::new(collection_is_never_read::CollectionIsNeverRead));
-    store.register_pre_expansion_pass(|| Box::<missing_assert_message::MissingAssertMessage>::default());
+    store.register_late_pass(|_| Box::new(missing_assert_message::MissingAssertMessage));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/clippy_lints/src/missing_assert_message.rs
+++ b/clippy_lints/src/missing_assert_message.rs
@@ -10,10 +10,10 @@ use rustc_span::sym;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks assertions that doesn't have a custom panic message.
+    /// Checks assertions without a custom panic message.
     ///
     /// ### Why is this bad?
-    /// If the assertion fails, a custom message may make it easier to debug what went wrong.
+    /// If the assertion fails, the custom message may make it easier to understand what went wrong.
     ///
     /// ### Example
     /// ```rust
@@ -30,7 +30,7 @@ declare_clippy_lint! {
     #[clippy::version = "1.69.0"]
     pub MISSING_ASSERT_MESSAGE,
     pedantic,
-    "checks assertions that doesn't have a custom panic message"
+    "checks assertions without a custom panic message"
 }
 
 #[derive(Default, Clone, Debug)]
--- a/clippy_lints/src/missing_assert_message.rs
+++ b/clippy_lints/src/missing_assert_message.rs
@@ -1,0 +1,123 @@
+use clippy_utils::diagnostics::span_lint;
+use rustc_ast::ast;
+use rustc_ast::{
+    token::{Token, TokenKind},
+    tokenstream::TokenTree,
+};
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks assertions that doesn't have a custom panic message.
+    ///
+    /// ### Why is this bad?
+    /// If the assertion fails, a custom message may make it easier to debug what went wrong.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let threshold = 50;
+    /// let num = 42;
+    /// assert!(num < threshold);
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let threshold = 50;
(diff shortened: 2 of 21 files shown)


clippy_lints/src/missing_assert_message.rs
```diff
@@ -0,0 +1,123 @@
...
+    token::{Token, TokenKind},
+    tokenstream::TokenTree,
+};
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks assertions that doesn't have a custom panic message.
```
llogiq, line 13:
```suggestion
    /// Checks assertions without a custom panic message.
```

clippy_lints/src/missing_assert_message.rs
```diff
@@ -0,0 +1,123 @@
...
+    ///
+    /// ### Why is this bad?
+    /// If the assertion fails, a custom message may make it easier to debug what went wrong.
```
llogiq, line 16:
```suggestion
    /// If the assertion fails, the custom message may make it easier to understand what went wrong.
```

clippy_lints/src/missing_assert_message.rs
```diff
@@ -0,0 +1,123 @@
...
+    /// assert!(num < threshold);
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let threshold = 50;
+    /// let num = 42;
+    /// assert!(num < threshold, "{num} is lower than threshold ({threshold})");
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub MISSING_ASSERT_MESSAGE,
```
llogiq, line 31: I'm not sure whether assert! had messages since 1.0, so I'll check if we need a minimum supported rust version.

clippy_lints/src/missing_assert_message.rs
```diff
@@ -0,0 +1,123 @@
...
+    pedantic,
+    "checks assertions that doesn't have a custom panic message"
```
llogiq, line 33:
```suggestion
    "checks assertions without a custom panic message"
```

llogiq (commented):
Great work! I think the macro stuff is really tricky.

I have but a few small doc suggestions and I'll have to look up if we need an msrv.

clippy_utils/src/numeric_literal.rs
```diff
@@ -219,7 +219,7 @@ impl<'a> NumericLiteral<'a> {
 }
 
 fn split_suffix<'a>(src: &'a str, lit_kind: &LitKind) -> (&'a str, Option<&'a str>) {
-    debug_assert!(lit_kind.is_numeric());
+    debug_assert!(lit_kind.is_numeric(), "`lit_kind` should be numeric");
     lit_suffix_length(lit_kind)
```
matthiaskrgr, line 223: My two cents: these changes display the very problem with this lint.
It encourages people to just add *some kind* of assert-message to silent the lint, but the message provides zero information.

If `foo.is_numeric()` returns `false` and blows up an assert, I already know that `foo` is not numeric and that this is problematic.
Why would would I want to waste time and add the same exact information "foo should be numeric" again, in a textual form (which just adds additional maintenance burden because now I have to additionally update the message every time I change the variable name, etc..) ?

IMO it would be good to add an additional message to the lint, something along the lines of
`help: try to describe why the failing assert is problematic`. The inserted assert-message should be more about the `why is it important to know that this failed` and not `what exactly failed here`, because the code inside the assert already conveys that.

llogiq: I agree that the messages provided here are a bit meh. However, I would not say that this is a problem with the lint; we simply have no way of knowing whether an assert message will satisfy. However, having them at all will at least put them under the eyes of a potential reviewer. Also I think your point should be added to the lint's documentation.

Perhaps tne lint should be put in `restriction` instead of `pedantic`.

llogiq:
I just looked and `assert!`s allowed messages back in 1.0.0, so no MSRV necessary.

clippy_lints/src/lib.rs
```diff
@@ -911,6 +912,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|_| Box::new(size_of_ref::SizeOfRef));
     store.register_late_pass(|_| Box::new(multiple_unsafe_ops_per_block::MultipleUnsafeOpsPerBlock));
     store.register_late_pass(|_| Box::new(extra_unused_type_parameters::ExtraUnusedTypeParameters));
+    store.register_pre_expansion_pass(|| Box::<missing_assert_message::MissingAssertMessage>::default());
```
Alexendoo, line 915: Unfortunately we don't want to be adding more pre expansion passes because it comes with a good number of bugs - #6610, it would want to be a late pass

We do have some utilities that would help out there: [`find_assert_args`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/macros/fn.find_assert_args.html) and [`find_assert_eq_args`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/macros/fn.find_assert_eq_args.html). Usage would be similar to

https://github.com/rust-lang/rust-clippy/blob/6444621c101151b49a5e7c9854d602483000e337/clippy_lints/src/assertions_on_constants.rs#L34-L40

clippy_utils/src/numeric_literal.rs
unexge: Yes you're right about the messages, since we are not very familiar with the codebase we tried to come up with messages by quickly looking at the assertion but it was hard to come up with good messages. 
I've added a help message and updated the lint's description. 
Didn't know about `restriction`, it makes more sense for this lint. I've reverted the "dogfood" commit, I believe someone with more familiarity with the codebase can enable the lint and come up with good assert messages in a follow-up PR.

clippy_lints/src/lib.rs
unexge: We started with a late pass and `find_assert_args` functions, but I think there was a recent change in the compiler and assert macros expand to something else than `find_assert_args` expects. So those functions don't return the correct `PanicExpn`, we tried to fix it but couldn't do it :) I'll give it another go!

unexge: So it seems like starting from Rust 1.52 `assert_{eq,ne}` macros use `panicking::assert_failed` instead of `panic!` (See `assert_eq` in [1.52](https://github.com/rust-lang/rust/blob/1.52.0/library/core/src/macros/mod.rs#L35-L62) and in [1.51](https://github.com/rust-lang/rust/blob/1.51.0/library/core/src/macros/mod.rs#L56-L86)).

https://github.com/rust-lang/rust-clippy/pull/10362/commits/a2502da1c4f6dee0f0444072385588827209671c addresses Rust 1.52 changes
https://github.com/rust-lang/rust-clippy/pull/10362/commits/569fa96915bfddf8724b5141382b3ca91393a858 converts this lint to a late lint pass

unexge:
/ping 😄 I believe all the comments are addressed and it's ready for another round of review

llogiq:
@Alexendoo since you commented on the implementation, do you have an idea on how to get this to work as a LateLintPass?

Alexendoo:
It's working as a late pass from the latest commit, it looks good to me at a glance

clippy_lints/src/missing_assert_message.rs
```diff
@@ -0,0 +1,75 @@
...
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks assertions without a custom panic message.
+    ///
+    /// ### Why is this bad?
+    /// Without a good custom message, it'd be hard to understand what went wrong when the assertion fails.
+    /// A good custom message should be more about why the failure of the assertion is problematic
+    /// and not what is failed because the assertion already conveys that.
+    ///
```
llogiq, line 17: I think we should have a `### Known Problems` section that admits we cannot check the assert messages for quality, and perhaps a link to some docs or guidelines?

llogiq (commented):
This looks good to me, I only have a last small doc suggestion.

clippy_lints/src/missing_assert_message.rs
unexge: I've added the `Known Problems` section but couldn't find a good doc to link. The best thing I can find is the answers to https://softwareengineering.stackexchange.com/questions/301652/purpose-of-assertion-messages. What do you think?

llogiq: Nah, let's not link to a site where the content could change at will. In that case, just leave the quality unspecified.

llogiq:
Thank you!

@bors r+

mickvangelderen:
Would it be possible to not trigger this lint in testing code? In tests the assert expression usually suffices but outside of tests the additional message is much more likely to be exposed to a non-developer.

unexge:
Hey @mickvangelderen, that should [already be the case](https://github.com/rust-lang/rust-clippy/blob/d732cce0d39d49ad07feb45d5f6d966a69d0147c/clippy_lints/src/missing_assert_message.rs#L60-L63). There might be some cases we fail to identify as a test function though, could you please share the code if possible?

mickvangelderen:
@unexge ah, I only read the lint documentation, not the code. Perhaps it should be added because I misjudged the false positive rate of this lint based on the current description.

unexge:
@mickvangelderen Yes, that's definitely a good point! I'll submit a PR. Thanks!

PR #10363, opened 2023-02-16
[significant_drop_tightening] Ignore inexpensive statements

c410-f3r:
Not all statements that follow the last use of a lock guard are expensive and can therefore be ignored by the lint. 

```rust
pub fn foo() -> i32 {
    let mutex = Mutex::new(1);
    let rslt = *lock;
    let another = rslt;
   another
}
```

---

changelog: [`significant_drop_tightening`]: No longer lints for inexpensive statements after the lock guard
[#10363](https://github.com/rust-lang/rust-clippy/pull/10363)


xFrednet (approved):

xFrednet:
Looks good to me, thank you!

@bors r+

PR #10368, opened 2023-02-18
[significant_drop_tightening] Evaluate the return expression of a block

c410-f3r:
For whatever reason, the return expression of a block is not contained inside the slice of statements and because of that the lint wasn't evaluating things that could potentially block the release of a lock.

```rust
pub fn example() -> i32 {
    let mutex = Mutex::new(1);
    let _ = *lock;
    let _ = *lock;
   do_heavy_computation_that_takes_time_and_returns_i32()
}
```

---

changelog: none

### DIFF
--- a/clippy_lints/src/significant_drop_tightening.rs
+++ b/clippy_lints/src/significant_drop_tightening.rs
@@ -61,10 +61,26 @@ pub struct SignificantDropTightening<'tcx> {
 }
 
 impl<'tcx> SignificantDropTightening<'tcx> {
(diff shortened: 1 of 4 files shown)


xFrednet (approved):

xFrednet:
Woops, yeah I missed this in the last review. Blocks are a bit tricky when it comes to this. Thank you, for finding that so quickly and creating a follow-up PR :)

@bors r+

c410-f3r:
Thanks @xFrednet

PR #10385, opened 2023-02-21
Update readme description of `restriction` lints to dissuade casual use.

Jarcho:
Adds some stronger wording about not enabling `restriction` lints. I've seen it come up a few times where people are confused about what the `restriction` category is for and end up casually enabling lints from it (or the whole category).

changelog: None

PR #10388, opened 2023-02-22
Remove restriction group from table and update book

flip1995:
The restriction group should not be listed in the table of lint groups with all the other lint groups. The restriction group is only there for convenient grouping and so that it easier in Clippy to generate documentation and the like. It is not there for Clippy users to enable like every other group.

Follow up to #10385, addressing my own comments

cc @Jarcho 
r? @xFrednet 

changelog: none

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,22 +19,23 @@ You can choose how much Clippy is supposed to ~~annoy~~ help you by changing the
 | `clippy::complexity`  | code that does something simple but in a complex way                                | **warn**      |
 | `clippy::perf`        | code that can be written to run faster                                              | **warn**      |
 | `clippy::pedantic`    | lints which are rather strict or have occasional false positives                    | allow         |
-| `clippy::restriction` | lints which prevent the use of language and library features[^restrict]             | allow         |
 | `clippy::nursery`     | new lints that are still under development                                          | allow         |
 | `clippy::cargo`       | lints for the cargo manifest                                                        | allow         |
 
 More to come, please [file an issue](https://github.com/rust-lang/rust-clippy/issues) if you have ideas!
 
-The `restriction` category should, *emphatically*, not be enabled as a whole. The contained
-lints may lint against perfectly reasonable code, may not have an alternative suggestion,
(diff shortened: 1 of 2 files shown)


xFrednet:
I understand both perspectives. Having the group in the table can give the wrong impression that `#[warn(clippy::restriction)]` is a viable thing to do. The user would then get a warning by a lint we have. On the other hand, not having it in the table makes it inconsistent. Coming from the lint list searching for what the `restriction` group is, it's easy to miss the comment below the table and then also not see the hint. I would be in favor of having it in the table, but I also don't have a strong opinion on this.

r? @Jarcho

flip1995:
I now also removed the footnote. Placing this all the way at the bottom of the file seemed weird to me. I wonder if we should include the examples there in the first place. It's probably better placed here: https://doc.rust-lang.org/nightly/clippy/lints.html#restriction

Jarcho:
I was questioning the footnote as I added it. If the table entry is removed it makes even less sense to keep it.

For the table it seems weird for the category to pretend `restriction` doesn't exist in the table, but treat it's existence equally everywhere else. There's also nothing next to the table to say `#[warn(clippy::group)]` is even a thing you can do.

The `restriction` group is also useful for public facing documentation, not just for internal tooling. In that sense it is a real category, just not one that's meant to control lint levels. We could just not create the category in rustc and actually solve the problem.

---

Unrelated, the order of the two paragraphs under the table should be flipped IMO. The current order is a concluding statement followed by more text about the thing that was just concluded. It's a little awkward to read.

xFrednet:
Hey @flip1995 this is a ping from triage. Would you mind taking a look at @Jarcho's latest comment? 

---

I part makes sense to me:

> The restriction group is also useful for public facing documentation, not just for internal tooling. In that sense it is a real category, just not one that's meant to control lint levels. We could just not create the category in rustc and actually solve the problem.

But I still don't have any strong opinions.

flip1995:
Fair points. Let's keep it in the table then without changes.

PR #10392, opened 2023-02-22
Fix more false positives for `extra_unused_type_parameters`

mkrasnitski:
Builds on #10321. All empty functions are no longer linted, instead of just those that have trait bounds on them. Also, if a trait bound contains a non-public trait (un-exported, but still potentially reachable), then the corresponding type parameter isn't linted.

Finally, added support for the `avoid_breaking_exported_api` config option.

r? @flip1995
changelog: none

### DIFF
--- a/clippy_lints/src/extra_unused_type_parameters.rs
+++ b/clippy_lints/src/extra_unused_type_parameters.rs
@@ -4,14 +4,17 @@ use rustc_data_structures::fx::FxHashMap;
 use rustc_errors::MultiSpan;
 use rustc_hir::intravisit::{walk_impl_item, walk_item, walk_param_bound, walk_ty, Visitor};
 use rustc_hir::{
-    BodyId, ExprKind, GenericParamKind, Generics, ImplItem, ImplItemKind, Item, ItemKind, PredicateOrigin, Ty, TyKind,
-    WherePredicate,
+    BodyId, ExprKind, GenericBound, GenericParamKind, Generics, ImplItem, ImplItemKind, Item, ItemKind,
+    PredicateOrigin, Ty, TyKind, WherePredicate,
 };
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::hir::nested_filter;
 use rustc_middle::lint::in_external_macro;
-use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::{def_id::DefId, Span};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::{
+    def_id::{DefId, LocalDefId},
+    Span,
+};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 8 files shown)


clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -38,7 +41,30 @@ declare_clippy_lint! {
     complexity,
     "unused type parameters in function definitions"
 }
-declare_lint_pass!(ExtraUnusedTypeParameters => [EXTRA_UNUSED_TYPE_PARAMETERS]);
+
+#[derive(Default)]
```
Jarcho, line 45: Shouldn't derive `Default` here.

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -151,11 +175,15 @@ impl<'cx, 'tcx> Visitor<'tcx> for TypeWalker<'cx, 'tcx> {
...
-                if self.fn_body_empty {
+                // If the bound contains non-public traits, err on the safe side and don't lint the
+                // corresponding parameter.
+                if !predicate
+                    .bounds
+                    .iter()
+                    .filter_map(bound_to_trait_def_id)
+                    .all(|def_id| self.cx.effective_visibilities.is_exported(def_id))
+                {
                     self.ty_params.remove(&def_id);
```
Jarcho, line 187: Should also set `all_params_unused`.

Jarcho (commented):
Quick review. Let's see if we can get this in today.

clippy_lints/src/extra_unused_type_parameters.rs
mkrasnitski: Not exactly sure why I added that. I think the example lint I looked at that used a config option (`upper_case_acronyms`) had that, so I figured it was required.

clippy_lints/src/extra_unused_type_parameters.rs
mkrasnitski: Good catch - this was actually introduced in #10321.

Jarcho:
Going to merge this now and save @flip1995 from having to wait a little for it.

Thank you. @bors r+

clippy_lints/src/extra_unused_type_parameters.rs
```diff
@@ -38,7 +41,29 @@ declare_clippy_lint! {
...
+impl ExtraUnusedTypeParameters {
+    pub fn new(avoid_breaking_exported_api: bool) -> Self {
+        Self {
+            avoid_breaking_exported_api,
+        }
+    }
+
+    /// Don't lint external macros or functions with empty bodies. Also, don't lint public items if
+    /// the `avoid_breaking_exported_api` config option is set.
+    fn check_false_positive(&self, cx: &LateContext<'_>, span: Span, def_id: LocalDefId, body_id: BodyId) -> bool {
```
flip1995, line 58: Weird function name IMO. But the best thing I can come up with is: `is_exported_macro_or_empty`.

tests/ui/extra_unused_type_parameters.rs
```diff
@@ -71,7 +79,23 @@ where
...
+    }
+
+    fn priv_trait_bound<T: private::Private>() {
+        unimplemented!();
+    }
+
+    fn unused_with_priv_trait_bound<T: private::Private, U>() {
+        unimplemented!();
+    }
+}
```
flip1995, line 98: Maybe I missed it, but I think there is a test for a `pub` function missing (should not get linted).

flip1995 (commented):
@mkrasnitski Awesome work, thanks for implementing this so quickly. Really a pleasure working with you!

I have two more comments, but they are just NITs. This unblocks the sync. Thanks for helping review @Jarcho

clippy_lints/src/extra_unused_type_parameters.rs
mkrasnitski: yeah, names are hard... I couldn't really come up with anything satisfactory either.

tests/ui/extra_unused_type_parameters.rs
mkrasnitski: As far as I can tell, that relies on the config option, right? Does `uitest` try different values of config vars as listed in `clippy.toml`?

mkrasnitski:
Glad to hear this made it in time. I have plans to try using `span_lint_and_sugg()` like you recommended previously, so these nits can be included along with those changes, when they happen.

tests/ui/extra_unused_type_parameters.rs
flip1995: Have a look in the `test/ui-toml` dir on how to test different config values.

PR #10400, opened 2023-02-25
chore: remove unneeded rustfmt skip

calebcartwright:
---

The associated rustfmt bug that originally necessitated these skips was resolved a while back, so these are no longer necessary

changelog: none


Jarcho:
Thank you. @bors r+

PR #10404, opened 2023-02-26
Add `unused_enumerate_index` lint

dnbln:
A lint for unused `.enumerate()` indexes (`for (_, x) in iter.enumerate()`).

I wasn't able to find a `rustc_span::sym::Enumerate`, so the code for checking that it's the correct `Enumerate` iterator is a bit weird.

---

changelog: New lint: [`unused_enumerate_index`]: A new lint for checking that the indexes from `.enumerate()` calls are used.
[#10404](https://github.com/rust-lang/rust-clippy/pull/10404)

### DIFF
--- a/clippy_lints/src/loops/mod.rs
+++ b/clippy_lints/src/loops/mod.rs
@@ -580,7 +580,7 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for `for (_, v) in a.iter().enumerate()`
+    /// Checks for uses of the `enumerate` method where the index is unused (`_`)
     ///
     /// ### Why is this bad?
     /// The index from `.enumerate()` is immediately dropped.
@@ -589,20 +589,20 @@ declare_clippy_lint! {
     /// ```rust
     /// let v = vec![1, 2, 3, 4];
     /// for (_, x) in v.iter().enumerate() {
-    ///     print!("{x}")
+    ///     println!("{x}");
     /// }
     /// ```
     /// Use instead:
     /// ```rust
     /// let v = vec![1, 2, 3, 4];
     /// for x in v.iter() {
-    ///     print!("{x}")
+    ///     println!("{x}");
     /// }
     /// ```
-    #[clippy::version = "1.69.0"]
+    #[clippy::version = "1.75.0"]
     pub UNUSED_ENUMERATE_INDEX,
     style,
-    "using .enumerate() and immediately dropping the index"
+    "using `.enumerate()` and immediately dropping the index"
 }
 
 declare_clippy_lint! {
--- a/clippy_lints/src/loops/mod.rs
+++ b/clippy_lints/src/loops/mod.rs
@@ -14,6 +14,7 @@ mod needless_range_loop;
 mod never_loop;
 mod same_item_push;
 mod single_element_loop;
+mod unused_enumerate_index;
 mod utils;
 mod while_immutable_condition;
 mod while_let_loop;
@@ -579,6 +580,33 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
+    /// Checks for `for (_, v) in a.iter().enumerate()`
+    ///
+    /// ### Why is this bad?
+    /// The index from `.enumerate()` is immediately dropped.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let v = vec![1, 2, 3, 4];
+    /// for (_, x) in v.iter().enumerate() {
+    ///     print!("{x}")
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let v = vec![1, 2, 3, 4];
+    /// for x in v.iter() {
+    ///     print!("{x}")
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub UNUSED_ENUMERATE_INDEX,
+    style,
+    "using .enumerate() and immediately dropping the index"
+}
+
+declare_clippy_lint! {
+    /// ### What it does
     /// Looks for loops that check for emptiness of a `Vec` in the condition and pop an element
     /// in the body as a separate operation.
     ///
@@ -619,6 +647,7 @@ impl Loops {
         }
     }
 }
+
 impl_lint_pass!(Loops => [
     MANUAL_MEMCPY,
     MANUAL_FLATTEN,
@@ -638,7 +667,8 @@ impl_lint_pass!(Loops => [
     SINGLE_ELEMENT_LOOP,
     MISSING_SPIN_LOOP,
     MANUAL_FIND,
-    MANUAL_WHILE_LET_SOME
+    MANUAL_WHILE_LET_SOME,
+    UNUSED_ENUMERATE_INDEX,
 ]);
 
 impl<'tcx> LateLintPass<'tcx> for Loops {
@@ -717,6 +747,7 @@ impl Loops {
         same_item_push::check(cx, pat, arg, body, expr);
         manual_flatten::check(cx, pat, arg, body, span);
         manual_find::check(cx, pat, arg, body, span, expr);
+        unused_enumerate_index::check(cx, pat, arg, body);
     }
 
     fn check_for_loop_arg(&self, cx: &LateContext<'_>, _: &Pat<'_>, arg: &Expr<'_>) {
--- a/clippy_lints/src/loops/unused_enumerate_index.rs
+++ b/clippy_lints/src/loops/unused_enumerate_index.rs
@@ -1,8 +1,7 @@
 use super::UNUSED_ENUMERATE_INDEX;
 use clippy_utils::diagnostics::{multispan_sugg, span_lint_and_then};
 use clippy_utils::source::snippet;
-use clippy_utils::sugg;
-use clippy_utils::visitors::is_local_used;
+use clippy_utils::{pat_is_wild, sugg};
 use rustc_hir::def::DefKind;
 use rustc_hir::{Expr, ExprKind, Pat, PatKind};
 use rustc_lint::LateContext;
@@ -60,13 +59,4 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>, arg: &'tcx
             );
         },
     );
-}
-
-/// Returns `true` if the pattern is a `PatWild` or an ident prefixed with `_`.
-fn pat_is_wild<'tcx>(cx: &LateContext<'tcx>, pat: &'tcx PatKind<'_>, body: &'tcx Expr<'_>) -> bool {
-    match *pat {
-        PatKind::Wild => true,
-        PatKind::Binding(_, id, ident, None) if ident.as_str().starts_with('_') => !is_local_used(cx, body, id),
-        _ => false,
-    }
 }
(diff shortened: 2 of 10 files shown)


blyxyas:
Hi @dnbln, Philipp doesn't have enough time to review PRs anymore, so I'll take the review from now on. Are you still interested in continuing with this development? ฅ^•ﻌ•^ฅ

r? @blyxyas

dnbln:
Hi @blyxyas, yes, certainly. I have resolved the merge conflicts; do let me know if there's anything else I should do.

clippy_lints/src/loops/mod.rs
```diff
@@ -577,6 +578,33 @@ declare_clippy_lint! {
     "manual implementation of `Iterator::find`"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `for (_, v) in a.iter().enumerate()`
```
blyxyas, line 583: A little bit more textual description would be better. We usually leave the code for the example, if the code is a just a little bit complex.

```suggestion
    /// Checks for uses of the `enumerate` method where the index is unused (`_`)
```

Or maybe a different wording

clippy_lints/src/loops/mod.rs
```diff
@@ -577,6 +578,33 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Why is this bad?
+    /// The index from `.enumerate()` is immediately dropped.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let v = vec![1, 2, 3, 4];
+    /// for (_, x) in v.iter().enumerate() {
+    ///     print!("{x}")
```
blyxyas, line 592: `println!` is far more familiar to users than `print!`, it almost seems like a typo for newer users.

(Also the semi-colon)

```suggestion
    ///     println!("{x}");
```

The same with the next example

clippy_lints/src/loops/mod.rs
```diff
@@ -577,6 +578,33 @@ declare_clippy_lint! {
...
+    /// ```rust
+    /// let v = vec![1, 2, 3, 4];
+    /// for x in v.iter() {
+    ///     print!("{x}")
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub UNUSED_ENUMERATE_INDEX,
+    style,
+    "using .enumerate() and immediately dropping the index"
```
blyxyas, line 605:
```suggestion
    "using `.enumerate()` and immediately dropping the index"
```

clippy_lints/src/loops/mod.rs
```diff
@@ -577,6 +578,33 @@ declare_clippy_lint! {
...
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let v = vec![1, 2, 3, 4];
+    /// for x in v.iter() {
+    ///     print!("{x}")
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
```
blyxyas, line 602:
```suggestion
    #[clippy::version = "1.74.0"]
```

clippy_lints/src/loops/unused_enumerate_index.rs
blyxyas, line 1: Don't worry about setting variables just to point to fields.
Variables like `pat_span = pat.span;` don't really add to readability, so I think we can remove those

clippy_lints/src/loops/unused_enumerate_index.rs
```diff
@@ -0,0 +1,75 @@
...
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>, arg: &'tcx Expr<'_>, body: &'tcx Expr<'_>) {
+    let pat_span = pat.span;
+
+    let PatKind::Tuple(pat, _) = pat.kind else {
+        return;
+    };
+
+    if pat.len() != 2 {
+        return;
+    }
```
blyxyas, line 20: This can be removed, if we check that the `enumerate` method comes from the standard library `Iterator` trait.

blyxyas (changes requested):
Just a first review. Very good for a first review, and welcome to contributing to Clippy! :heart:

clippy_lints/src/loops/mod.rs
dnbln: I'm not sure, but shouldn't this lint be using `clippy::version` for the current nightly (`1.75.0`)?

clippy_lints/src/loops/unused_enumerate_index.rs
```diff
@@ -49,16 +39,16 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>, arg: &'tcx
...
         "you seem to use `.enumerate()` and immediately discard the index",
         |diag| {
             let base_iter = sugg::Sugg::hir(cx, self_arg, "base iter");
             multispan_sugg(
                 diag,
                 "remove the `.enumerate()` call",
                 vec![
-                    (pat_span, snippet(cx, new_pat_span, "value").into_owned()),
-                    (arg_span, base_iter.to_string()),
+                    (pat.span, snippet(cx, tuple[1].span, "value").into_owned()),
```
blyxyas, line 50: We usually use `..` as a snippet default.

```suggestion
                    (pat.span, snippet(cx, tuple[1].span, "..").into_owned()),
```

clippy_lints/src/loops/mod.rs
```diff
@@ -599,10 +599,10 @@ declare_clippy_lint! {
     ///     print!("{x}")
```
blyxyas, line 599: This line also needs the same change.

clippy_lints/src/loops/mod.rs
```diff
@@ -599,10 +599,10 @@ declare_clippy_lint! {
...
     /// }
     /// ```
-    #[clippy::version = "1.69.0"]
+    #[clippy::version = "1.74.0"]
```
blyxyas, line 602: I did what we call a *silly mistake*, this needs to be on `1.75.0` ≽^•⩊•^≼

blyxyas (commented):
I did a little bit of trolling in my last review hehe :3
Seems pretty good, this is the last review before doing the final complete one.

blyxyas (commented):
Just some more tests and this is it! Thanks for your patience and I hope this new lint (quite a useful one, also) gets merged as soon as possible! :heart: ^•ﻌ•^ฅ♡

tests/ui/unused_enumerate_index.rs
```diff
@@ -0,0 +1,9 @@
+#![allow(unused)]
+#![warn(clippy::unused_enumerate_index)]
+
+fn main() {
+    let v = [1, 2, 3];
+    for (_, x) in v.iter().enumerate() {
+        print!("{x}");
+    }
```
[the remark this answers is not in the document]
dnbln: I added some more test cases for those, I'm not sure if I can come up with any others.

clippy_lints/src/loops/unused_enumerate_index.rs
```diff
@@ -0,0 +1,76 @@
...
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>, arg: &'tcx Expr<'_>, body: &'tcx Expr<'_>) {
+    let PatKind::Tuple(tuple, _) = pat.kind else {
+        return;
+    };
+
+    let ExprKind::MethodCall(method, self_arg, [], _) = arg.kind else {
+        return;
+    };
+
+    if method.ident.as_str() != "enumerate" {
```
dnbln, line 21: I'm unsure if this check is really necessary anymore given the `call_name == "{std,core}::iter::Iterator::enumerate"` check below, do let me know if I should remove it.

clippy_lints/src/loops/unused_enumerate_index.rs
```diff
@@ -0,0 +1,76 @@
...
+        return;
+    }
+
+    let Some((DefKind::AssocFn, call_id)) = cx.typeck_results().type_dependent_def(arg.hir_id) else {
+        return;
+    };
+
+    let call_name = cx.tcx.def_path_str(call_id);
+
+    if call_name != "std::iter::Iterator::enumerate" && call_name != "core::iter::Iterator::enumerate" {
```
dnbln, line 46: Also another corner case: what if the `impl Iterator for I` overrides the `fn enumerate(...)`? How should such cases be handled? rustdoc doesn't seem to point out such known overrides (or at least I don't know where to look), so are there any other overrides that should be added here?

blyxyas: AFAIK, there's nothing that we can really do to know if a method has been overriden that wouldn't produce false negatives. Also, it seems like [it isn't a common pattern to override `enumerate`](https://github.com/search?q=lang%3Ars+fn+enumerate%28self%29+-%3E+Enumerate%3CSelf%3E&type=code). So overriden `enumerate` isn't really a problem nor has solution.

clippy_lints/src/loops/unused_enumerate_index.rs
blyxyas: Yep, it can be removed.

blyxyas:
We have a slight block here (don't worry, I'll fix it soon（=´∇｀=）). We have this deprecated [`paths`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/paths/index.html) module, and yesterday I started an effort to convert these paths to diagnostic items and convert existing uses of these paths to use diagnostic items.

Don't worry, it won't take long. I'll update you with time.
There isn't anything more to do on your end.

clippy_lints/src/loops/unused_enumerate_index.rs
Jarcho: There's no way to override any of the adapter methods without unsafe code. The return types don't have any exposed constructors.

blyxyas:
It seems like it will take more than expected (without even counting the reviewing + sync time), so I'll unblock this PR and just fix it later in my rust-lang/rust branch. I don't think it's worth it making you wait even longer haha :sweat_smile:

clippy_lints/src/loops/unused_enumerate_index.rs
```diff
@@ -0,0 +1,72 @@
...
+            );
+        },
+    );
+}
+
+/// Returns `true` if the pattern is a `PatWild` or an ident prefixed with `_`.
+fn pat_is_wild<'tcx>(cx: &LateContext<'tcx>, pat: &'tcx PatKind<'_>, body: &'tcx Expr<'_>) -> bool {
+    match *pat {
+        PatKind::Wild => true,
+        PatKind::Binding(_, id, ident, None) if ident.as_str().starts_with('_') => !is_local_used(cx, body, id),
```
blyxyas, line 69: Pretty smart move.

tests/ui/unused_enumerate_index.rs
```diff
@@ -0,0 +1,37 @@
+#![allow(unused)]
+#![warn(clippy::unused_enumerate_index)]
+
```
blyxyas, line 3:
```suggestion
use std::iter::Enumerator;
```

blyxyas (approved):
Ok, that's it! Add a test implementing `std::iter::Iterator` for a `Dummy3` structure and then squash the branch into 1 or 2 commits. It should be merge-ready then!

clippy_lints/src/loops/unused_enumerate_index.rs
dnbln: I've used the `FOR_KV_MAP` lint for reference, this function is taken straight from there. Maybe it could be moved to an util function, since with this it would already be defined 3 times?

https://github.com/rust-lang/rust-clippy/blob/7d34406015fce08031d0986dec0a4689a3cc407e/clippy_lints/src/loops/for_kv_map.rs#L59-L66

blyxyas: Oh yes then, move to it `clippy_utils`

blyxyas (approved):
LGTM, thanks for the lint, and sorry for taking so long to review! ❤️

PR #10414, opened 2023-02-27
add large future lint

### LINKED ISSUE #5263: Warn about large Future sizes
When using `.await` it's easy to compose several async functions together. The problem is, each function has its own state, and all states of the whole call tree are merged into one giant Future object.

I've ended with async functions that have 60KB of state, and code using them started getting stack overflows.

It would be helpful if Clippy measured sizes of implicit Futures created for async functions and warn when they reach a non-trivial sizes (e.g. > 16KB).

csmoe:
Closes #5263

---

changelog: new lint: [`large_futures`]
[#10414](https://github.com/rust-lang/rust-clippy/pull/10414)


csmoe:
@xFrednet Error fixed, may I have your review? thanks :)

clippy_lints/src/large_future.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// }
+    /// ```
+    ///
+    /// `Box::pin` the big future instead.
+    ///
+    /// ```txt
+    /// let fut = Box::pin(big_fut([0u8; 1024]));
+    /// ```
+    #[clippy::version = "1.68.0"]
+    pub LARGE_FUTURE,
```
xFrednet, line 33: Small naming nit: Lint names should [ideally be plural](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints). So `LARGE_FUTURES` would be a better name. You can try renaming the lint with `cargo dev`. I don't remember the exact comment, but you should find it in the help menu. It'll add some tests, for renamed lints, which will need to be removed, but it does the whole file renaming thing for you :)

clippy_lints/src/large_future.rs
```diff
@@ -0,0 +1,78 @@
...
+    nursery,
```
xFrednet, line 34: The nursery group is for lints, which are currently under active development or have a high number of false positives. For this one, I would say that either `clippy::perf` or `clippy::pedantic` would be correct. I could see this generating quite many warnings in some projects, which might be harder to fix due to the type change. Having it allow-by-default in the `clippy::pedantic` is probably the safe option :)

clippy_lints/src/large_future.rs
```diff
@@ -0,0 +1,78 @@
...
+                        && implements_trait(cx, ty, future_trait_def_id, &[]) {
+                            if let Ok(layout) = cx.tcx.layout_of(cx.param_env.and(ty)) {
+                                let size = layout.layout.size();
+                                if size >= Size::from_bytes(self.future_size_threshold) {
+                                    span_lint(
+                                        cx,
+                                        LARGE_FUTURE,
+                                        expr.span,
+                                        &format!("large future with {size:?}, consider `Box::pin` on it"),
+                                    );
```
xFrednet, line 70: I feel like the `Size(XYZ bytes)` message looks a bit off. I would suggest this message.

```rs
&format!("a large future with a size of {} bytes", size.bytes()),
```

The suggestion should usually be added as a help message. You can take a look at `span_lint_and_help`. Would it maybe be possible to create an suggestion with `Box::pin`? I know that we won't be able to make it machine applicable, as it might require additional modifications. However, displaying a suggestion can sometimes be a big help. Something like this:

```
Box::pin(future)
```

xFrednet (commented):
Hey thank you for the PR. This is already an excellent start. I've added some comments. You're welcome to reach out if you need further assistance :)

clippy_lints/src/utils/conf.rs
```diff
@@ -459,6 +459,10 @@ define_Conf! {
     /// Whether to **only** check for missing documentation in items visible within the current
     /// crate. For example, `pub(crate)` items.
     (missing_docs_in_crate_items: bool = false),
+    /// Lint: LARGE_FUTURE
+    ///
+    /// The maximum size of a `Future` can have.
+    (future_size_threshold: u64 = 16 * 1024),
```
xFrednet, line 465: Also, could you add a test for this config value inside the `tests/ui-toml/` test folder?

tests/ui-toml/large_futures/clippy.toml
```diff
@@ -0,0 +1 @@
+future-size-threshold = 16384
```
xFrednet, line 1:
```suggestion
future-size-threshold = 16384

```

These files are sadly now formatted automatically.

I think for readability, it would also be easier, if the test config value would be more distinct from the 16kb. Maybe 1024 would be better. But keeping the current size is a also fine :)

clippy_lints/src/large_futures.rs
```diff
@@ -0,0 +1,80 @@
...
+                        && implements_trait(cx, ty, future_trait_def_id, &[]) {
+                            if let Ok(layout) = cx.tcx.layout_of(cx.param_env.and(ty)) {
+                                let size = layout.layout.size();
+                                if size >= Size::from_bytes(self.future_size_threshold) {
+                                    span_lint(
+                                        cx,
+                                        LARGE_FUTURES,
+                                        expr.span,
+                                        &format!("large future with a size of {} bytes, consider `Box::pin` on it", size.bytes()),
+                                    );
```
xFrednet, line 72: Are you currently looking into creating a suggestion, as proposed in my last comment? If not, we can just switch it to `span_lint_and_help` creating a proper suggestion might be a bit more complicated.

This code snipped should do it :)

```rust
span_lint_and_help(
    cx,
    LARGE_FUTURES,
    expr.span,
    &format!("large future with a size of {} bytes", size.bytes()),
    None,
    "consider using `Box::pin` to store the future on the heap",
);
```

xFrednet (commented):
Looks good to me overall, two small nits and then it should be ready for merge :)

csmoe:
rustc is trying to address this problem, let's wait for that :)

clippy_lints/src/large_futures.rs
csmoe: yep, still testing locally.

xFrednet:
Do you have an estimate for the timeframe? If that'll take a few releases/months we could still add the lint and deprecate it later. :)

csmoe:
@xFrednet Done~ :)

csmoe:
@xFrednet ping :)

xFrednet:
Hey @csmoe, thank you for your patience, the last two weeks have been very chaotic. You'll have a review by the end of the week! :)

tests/ui/large_futures.fixed
```diff
@@ -0,0 +1,41 @@
...
+
+pub fn foo() -> impl std::future::Future<Output = ()> {
+    async {
+        let x = [0i32; 1024 * 16];
+        async {}.await;
+        dbg!(x);
+    }
+}
+
+fn main() {}
```
xFrednet, line 41: Could you add a test with a multi line async block to see if the output looks good? Something like:

```
async {
    let x = [0i32; 1024 * 16];
    println!("I make this block multiline :3");
}.await;
```

xFrednet (commented):
Looks good to me overall. A few small nits which should hopefully be simple to fix and then we can merge this :)

clippy_lints/src/large_futures.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn new(future_size_threshold: u64) -> Self {
+        Self { future_size_threshold }
+    }
+}
+
+impl_lint_pass!(LargeFuture => [LARGE_FUTURES]);
+
+impl<'tcx> LateLintPass<'tcx> for LargeFuture {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let ExprKind::Match(expr, _, MatchSource::AwaitDesugar) = expr.kind {
```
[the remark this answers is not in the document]
csmoe: if will not works here since we are checking the lowered expr.await by [lower_expr_await](https://doc.rust-lang.org/stable/nightly-rustc/rustc_ast_lowering/struct.LoweringContext.html#method.lower_expr_await), which marks the span of `expr.await` as Desugred as a whole, thus `check_if_expand_from_macro` always returns false.

xFrednet (approved):

xFrednet:
Let's try this again :)

@bors r+

clippy_lints/src/large_futures.rs
```diff
@@ -0,0 +1,87 @@
...
+    ///
+    /// ### Why is this bad?
+    /// Due to the current [unideal implemention](https://github.com/rust-lang/rust/issues/69826) of `Generator`,
+    /// large size of a `Future` may cause stack overflows.
+    ///
+    /// ### Example
+    /// ```rust
+    /// async fn wait(f: impl std::future::Future<Output = ()>) {}
+    ///
+    /// async fn big_fut(arg: [u8; 1024]) {}
```
mickvangelderen, line 21: @csmoe Given the default threshold of 16*1024 bytes, I don't think the example code in the lint documentation will actually trigger the lint. 

Besides this, can you maybe shed some light on why:

```rust
async fn fut() {
    let _x = [0; 16*1024];
    async {}.await;
}

async fn main() {
    fut().await; // <- expect to trigger here
}
```

Does not trigger the lint? In my mind the future returned by `fut` needs to keep around `_x` until after the await point. Is this somehow optimized out or am I just not thinking about this correctly?

csmoe: yes, _x is optimized out as it does not live across await point(see my test cases, I made it by a printf). You can check fut()'s size, it should be very small.

mickvangelderen: Interesting that it can optimize a local but not an argument. 

To address the broken example, I've created opened https://github.com/rust-lang/rust-clippy/pull/11553.

PR #10415, opened 2023-02-27
Add `collection_is_never_read`

### LINKED ISSUE #9267: Lint idea: Collection is only added to, never read

schubart:
Fixes #9267

@flip1995 and @llogiq, I talked with you about this one at Rust Nation in London last week. :-)

This is my first contribution to Clippy, so lots of feedback would be greatly appreciated.

- \[x] Run `cargo dev fmt`

`dogfood` found one true positive (see #9509) and no false positives.

`lintcheck` found no (true or false) positives, even when running on an extended set of crates.

---

changelog: new lint [`collection_is_never_read`]
[#10415](https://github.com/rust-lang/rust-clippy/pull/10415)

### DIFF
--- a/clippy_lints/src/collection_is_never_read.rs
+++ b/clippy_lints/src/collection_is_never_read.rs
@@ -30,6 +30,7 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
+    /// # let samples = vec![3, 1, 2];
     /// let mut sorted_samples = samples.clone();
     /// sorted_samples.sort();
     /// for sample in &samples { // Oops, meant to use `sorted_samples`.
@@ -38,6 +39,7 @@ declare_clippy_lint! {
     /// ```
     /// Use instead:
     /// ```rust
+    /// # let samples = vec![3, 1, 2];
     /// let mut sorted_samples = samples.clone();
     /// sorted_samples.sort();
     /// for sample in &sorted_samples {
--- a/clippy_lints/src/collection_is_never_read.rs
+++ b/clippy_lints/src/collection_is_never_read.rs
@@ -98,10 +98,16 @@ fn has_no_read_access<'tcx>(cx: &LateContext<'tcx>, id: HirId, block: &'tcx Bloc
         // Method call on `id` in a statement ignores any return value, so it's not a read access:
         //
         // id.foo(...); // Not reading `id`.
+        //
+        // Only assuming this for "official" methods defined on the type. For methods defined in extension
+        // traits (identified as local, based on the orphan rule), pessimistically assume that they might
+        // have side effects, so consider them a read.
         if let Some(Node::Expr(parent)) = get_parent_node(cx.tcx, expr.hir_id)
             && let ExprKind::MethodCall(_, receiver, _, _) = parent.kind
             && path_to_local_id(receiver, id)
             && let Some(Node::Stmt(..)) = get_parent_node(cx.tcx, parent.hir_id)
+            && let Some(method_def_id) = cx.typeck_results().type_dependent_def_id(parent.hir_id)
+            && !method_def_id.is_local()
         {
             return ControlFlow::Continue(());
         }
--- a/clippy_lints/src/collection_is_never_read.rs
+++ b/clippy_lints/src/collection_is_never_read.rs
@@ -1,20 +1,11 @@
 use clippy_utils::diagnostics::span_lint;
-use clippy_utils::get_enclosing_block;
-use clippy_utils::get_parent_node;
-use clippy_utils::path_to_local_id;
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::visitors::for_each_expr_with_closures;
+use clippy_utils::{get_enclosing_block, get_parent_node, path_to_local_id};
 use core::ops::ControlFlow;
-use rustc_hir::Block;
-use rustc_hir::ExprKind;
-use rustc_hir::HirId;
-use rustc_hir::Local;
-use rustc_hir::Node;
-use rustc_hir::PatKind;
-use rustc_lint::LateContext;
-use rustc_lint::LateLintPass;
-use rustc_session::declare_lint_pass;
-use rustc_session::declare_tool_lint;
+use rustc_hir::{Block, ExprKind, HirId, Local, Node, PatKind};
+use rustc_lint::{LateContext, LateLintPass};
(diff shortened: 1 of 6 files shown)


xFrednet:
r? @llogiq

(giraffate seems to be busy and you seem to know about this lint based on the description :))

clippy_lints/src/collection_is_never_read.rs
```diff
@@ -0,0 +1,125 @@
...
+use clippy_utils::path_to_local_id;
+use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::visitors::for_each_expr_with_closures;
+use core::ops::ControlFlow;
+use rustc_hir::Block;
+use rustc_hir::ExprKind;
+use rustc_hir::HirId;
+use rustc_hir::Local;
+use rustc_hir::Node;
+use rustc_hir::PatKind;
```
llogiq, line 13:
```suggestion
use rustc_hir::{Block, ExprKind, HirId, Local, Node, PatKind};
```
You can combine multiple imports (also in other places) to reduce the code a bit.

tests/ui/collection_is_never_read.rs
```diff
@@ -0,0 +1,116 @@
+#![allow(unused)]
+#![warn(clippy::collection_is_never_read)]
+
+use std::collections::HashMap;
+
+fn main() {}
```
llogiq, line 6: I may have missed it, but I'd like to see a test using the return value of an `insert` (to some set or map) e.g. for detecting duplicates.

Something like `if my_set.insert(val).is_some() { .. }`

clippy_lints/src/collection_is_never_read.rs
schubart: Done.

tests/ui/collection_is_never_read.rs
schubart: Done.

schubart:
Thanks for your comments. Regarding docs about possible known problems: I honestly can't think of any. Using `lintcheck` found some false positives in earlier versions of the lint (e.g. not searching through closures) but I've addressed everything I found or could think of. I'm very open to ideas as to what might be missing.

schubart:
I came up with a false positive:

The lint assumes that a statement `x.foo(args);` is not reading the collection `x` and that any `x` that is only used in this way is pointless. This assumption is wrong if `foo` has side effects or if it modifies `args`. 

I'm not aware of any methods like that in the container types I'm targeting, but such methods can be added using extension traits:

```rust
    trait VecExt<T> {
        fn print_self(&self);
    }

    impl<T> VecExt<T> for Vec<T> {
        fn print_self(&self) {
            println!("my length: {}", self.len());
        }
    }

    let mut x = vec![1, 2, 3]; // Wrong warning: `collection_is_never_read`
    x.print_self();
```

Perhaps I should check if `foo` is defined via an extension trait and if so, pessimistically assume that it has side effects.

Is there a way to check where `foo` is defined?

llogiq:
Regarding the possible false positive via extension trait, *if* the trait is defined in the local crate, we might be able to get the method by def_id. I would advise to avoid that because it only pushes the boundary one step: If your trait method calls another one, you're back to step one unless you can follow the whole chain of methods, at which point you will run the risk of a) a general performance nightmare and b) possible infinite recursion if methods call themselves. Even if you collect all the needed data on `visit_crate(..)`, you'll recreate most of the information that rustc already has for very little gain; if only one function call is outside of your crate, the chain is broken.

So I would suggest we add that possibility under `### Known Problems` in the docs and then merge this.

schubart:
My idea is simpler than what I think you describe: I don't want to look into the definitions of any methods to see if they have side effects. I only want to identify call to methods that are not part of the "official" collection type. Such calls can be pessimistically considered reads. This avoids the false positive above (but might create false negatives, which seems acceptable).

I think I can identify such methods via the orphan rule: Any method on, say, `Vec` is either part of its definition in `std`, or it's local.

I've pushed a commit that implements this idea. What do you think?

schubart:
How the lint works: For a local variable `x` that is of a container type, raise a warning if there is no read access (but there is some access). The lint considers almost every access of `x` to be a read access, so by default the risk of false positives seems low.

The only way of accessing `x` that the lint does not consider a read access is a *statement* of the form `x.foo(args);`, i.e. a *method call* with receiver `x` where the return value is not used. If all access to `x` is of this form, then the lint raises a warning.

In your example, `foo` is not a method, so the lint *does* consider `foo(x)` a read access and no warning is raised, which we seem to agree is the right behaviour:

```rust
    fn foo<T>(v: &Vec<T>) -> usize {
        v.len()
    }

    let x = vec![1, 2, 3]; // Ok, no warning
    foo(&x);
```

Note that this particular example could be considered a false negative because nothing useful is ever done with the vector. But that's beyond the scope of the lint. If the collection is passed to a function, then the lint assumes that that's meaningful and necessary. It does not try to look into the method. Since `foo` in this example is just a wrapper around `len()`, it should probably be annotated as `must_use` and different lints would catch the pointlessness of this code.

llogiq:
Ah, I missed that. Thank you for writing this lint!

@bors r+

schubart:
Thanks!

A few things I was hoping to discuss before merging:

* Is the name OK?
* It's currently in `nursery`. Could it be in `suspicious` instead?
* Shall I squash the commits before merging? I guess that's a moot point now.

PR #10418, opened 2023-02-27
Run dogfood to completion

Jarcho:
Run dogfood on all packages before failing the test. Failing early is painful on lints which trigger on multiple crates.

changelog: None

### DIFF
--- a/tests/dogfood.rs
+++ b/tests/dogfood.rs
@@ -7,6 +7,7 @@
 #![cfg_attr(feature = "deny-warnings", deny(warnings))]
 #![warn(rust_2018_idioms, unused_lifetimes)]
 
+use itertools::Itertools;
 use std::path::PathBuf;
 use std::process::Command;
 use test_utils::IS_RUSTC_TEST_SUITE;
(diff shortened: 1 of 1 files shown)


.cargo/config.toml
```diff
@@ -3,6 +3,7 @@ uitest = "test --test compile-test"
 dev = "run --package clippy_dev --bin clippy_dev --manifest-path clippy_dev/Cargo.toml --"
 lintcheck = "run --package lintcheck --bin lintcheck --manifest-path lintcheck/Cargo.toml  -- "
 collect-metadata = "test --test dogfood --features internal -- run_metadata_collection_lint --ignored"
+dogfood = "test --test dogfood"
```
flip1995, line 6: We have `dev dogfood` with `--fix` support. So I don't think we need another alias.

flip1995 (approved):
One comment. I'm not opposed with merging it as-is, if you think this is useful. r=me either way.

.cargo/config.toml
Jarcho: Somehow I missed that. I've been using `test` directly this entire time.

flip1995: I also only noticed it 2 weeks ago or so :smile:

flip1995:
@bors r+

Thanks!

PR #10426, opened 2023-02-28
Port clippy away from compiletest to ui_test

oli-obk:
Reasons to do this:

* runs completely on stable Rust
* is easier to extend with new features
* has its own dogfood test suite, so changes can be tested in [the `ui_test` repo](https://github.com/oli-obk/ui_test)
* supports dependencies from crates.io without having to manually fiddle with command line flags
* supports `ui-cargo`, `ui`, `ui-toml` out of the box, no need to find and run the tests ourselves

One thing that is a big difference to `compiletest` is that if a test emits *any* error, you need to mark all of them with `//~ ERROR:` annotations. Since many clippy tests did not have annotations, I changed many lints to be `warn` in their test so that only the `stderr` output is tested.

TODO:

* [ ] check that this still works as a subtree in the rustc repo

changelog: none

Note: at present the latest changes needed for clippy are only available as a git dependency, but I expect to publish a new crates.io version soon


Manishearth:
cc @rust-lang/clippy 

(Could you split the commit between the one that fixes the tests and the one that updates the runner?)

oli-obk:
> `clippy_test_deps/Cargo.lock` doesn't get generated for us

you need to `env BLESS=1` to bless lock files. I will change that blessing to whatever scheme you'd like, but I think clippy had no scheme for blessing while running tests before?

oli-obk:
> The second is that there's no progress output, I don't know how `compiletest-rs` manages it, but it prints the test progress without having to pass `--nocapture`

ah, yea I can do that. Btw: do y'all want dots or filenames by default? I can make it so CI prints filenames and local running prints dots.

Manishearth:
Looks like some tests are broken

image


There should be no changes to blessed files, no?

Alexendoo:
There's some tests that still have `// old-style-headers`:

```
tests/ui-internal/custom_ice_message.rs
tests/ui/almost_complete_range.rs
tests/ui/asm_syntax.rs
tests/ui/author/blocks.rs
tests/ui/cast_size_32bit.rs
tests/ui/cast_size.rs
tests/ui/crashes/ice-4968.rs
tests/ui/crashes/ice-7410.rs
tests/ui/crate_level_checks/entrypoint_recursion.rs
tests/ui/crate_level_checks/no_std_main_recursion.rs
tests/ui/def_id_nocore.rs
tests/ui/empty_loop_no_std.rs
tests/ui/entry.rs
tests/ui/enum_clike_unportable_variant.rs
tests/ui/expect_tool_lint_rfc_2383.rs
tests/ui/fn_to_numeric_cast_32bit.rs
tests/ui/fn_to_numeric_cast.rs
tests/ui/macro_use_imports_expect.rs
tests/ui/macro_use_imports.rs
tests/ui/manual_assert.rs
tests/ui/missing_doc.rs
tests/ui/must_use_unit.rs
tests/ui/needless_parens_on_range_literals.rs
tests/ui/needless_splitn.rs
tests/ui/no_mangle_with_rust_abi.rs
tests/ui/non_octal_unix_permissions.rs
tests/ui/numbered_fields.rs
tests/ui/redundant_clone.rs
tests/ui/to_digit_is_some.rs
tests/ui/transmute_32bit.rs
tests/ui/transmute_64bit.rs
tests/ui/uninlined_format_args_panic.rs
tests/ui/wildcard_imports_2021.rs
tests/ui/wildcard_imports.rs
```

From `rg '^// *(edition|revisions|\w+-)' tests -t rust` if you want something more locally clickable

oli-obk:
oh oops

>  error: 'cargo-clippy' is not installed for the toolchain 'nightly-2023-02-25-x86_64-unknown-linux-gnu'
To install, run `rustup component add clippy --toolchain nightly-2023-02-25-x86_64-unknown-linux-gnu`

I guess for the `ui-cargo` tests I better build `cargo-clippy` first. Or should I revert to the previous behaviour of collecting the `.rs` files and running `clippy-driver` on them?

Alexendoo:
It won't be in the toolchain but `cargo-clippy` should be built alongside `clippy-driver`, so something like

```rust
config.program.set_file_name(if cfg!(windows) {
    "cargo-clippy.exe"
} else {
    "cargo-clippy"
});
```

would hopefully work

tests/compile-test.rs
```diff
@@ -187,177 +88,72 @@ fn run_internal_tests() {
...
-    let res = run_tests(&config, &config.filters, tests);
-    env::set_current_dir(current_dir).unwrap();
+    config.args = vec![
+        "clippy".into(),
+        "--color".into(),
+        "never".into(),
+        "--target-dir".into(),
+        "target/ui_test_cargo".into(),
+        "--manifest-path".into(),
+    ];
```
Alexendoo, line 116: Could we use `--quiet` here? I think that'd get rid of most of the need for the filters while still printing any warnings

Alexendoo:
> should I revert to the previous behaviour of collecting the `.rs` files and running `clippy-driver` on them?

I'd say actually running `cargo-clippy` would be nicer if we can get it to work well

tests/ui-toml/array_size_threshold/array_size_threshold.stderr
```diff
@@ -1,29 +0,0 @@
-error: large array defined as const
-  --> $DIR/array_size_threshold.rs:4:1
-   |
-LL | const ABOVE: [u8; 11] = [0; 11];
-   | -----^^^^^^^^^^^^^^^^^^^^^^^^^^^
-   | |
-   | help: make this a static item: `static`
-   |
-   = note: `-D clippy::large-const-arrays` implied by `-D warnings`
```
Alexendoo, line 9: Looks like the `ui-toml` tests aren't picking up the `clippy.toml`, they need to have `CLIPPY_CONF_DIR` (or `CARGO_MANIFEST_DIR` as in the current implementation) set to the directory they're in

oli-obk: I cheated and made `clippy-driver` look in the directory of the file it's compiling first. I can look into using `CLIPPY_CONF_DIR`, but that would essentially restore the previous system of manually iterating over the folders and running the test suite once per file.

oli-obk:
Ah the remaining failure is a flakiness where `ui-cargo` causes collisions between crates of the same name. Obviously tests should be allowed to have the same name and not conflict. This needs a fix in `ui_test`.

tests/compile-test.rs
```diff
@@ -187,177 +88,79 @@ fn run_internal_tests() {
...
+            .parent()
+            .unwrap()
+            .canonicalize()
+            .unwrap()
+            .parent()
+            .unwrap()
+            .display()
+            .to_string(),
+        "$$DIR",
+    );
```
Alexendoo, line 111: The path part of this will need to be ran through `regex::escape`, e.g. on windows it contains backslashes

Alexendoo:
On Windows some backslashes in paths are left behind:

```
actual output differed from expected
--- tests\ui\approx_const.stderr
+++ <stderr output>
 warning: approximate value of `f{32, 64}::consts::E` found
-  --> $DIR/approx_const.rs:4:16
+  --> $DIR\approx_const.rs:4:16
    |
 LL |     let my_e = 2.7182;
```

clippy_lints/src/utils/conf.rs
```diff
@@ -466,15 +466,18 @@ define_Conf! {
...
     // Start looking for a config file in CLIPPY_CONF_DIR, or failing that, CARGO_MANIFEST_DIR.
-    // If neither of those exist, use ".".
+    // If neither of those exist, use the directory of the input source file or in case the input is
+    // passed via stdin, use ".".
     let mut current = env::var_os("CLIPPY_CONF_DIR")
-        .or_else(|| env::var_os("CARGO_MANIFEST_DIR"))
-        .map_or_else(|| PathBuf::from("."), PathBuf::from);
+        .map(PathBuf::from)
+        .or_else(|| Some(input_file?.parent().unwrap().to_path_buf()))
+        .or_else(|| Some(env::var_os("CARGO_MANIFEST_DIR")?.into()))
```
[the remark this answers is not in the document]
oli-obk: Yea, I think fixing the ui-cargo flakiness will allow me to just overwrite the `CARGO_MANIFEST_DIR` env var per test, so I'm going to revert that change to clippy driver

oli-obk:
CI is green, though if someone could check whether it works on windows that would be great.

There's some follow-ups I want to do, like changing the entire test suite back to `-Dwarnings`, but that requires touching 150 tests or so, so I'd prefer not to do that in this PR.

I'll update the usage instructions now, too, as it's not necessary anymore to run `cargo dev bless`, instead the stderr files are updated by enabling the `BLESS` env var. Any opinions on the scheme for this? name of env var? different scheme like `cargo test -- --bless`?

Alexendoo:
Windows errors:

ui:

```diff
--- tests\ui\track-diagnostics.stderr
+++ <stderr output>
... 3 lines skipped ...
 LL | const S: A = B;
    |              ^ expected `A`, found `B`
--Ztrack-diagnostics: created at compiler/rustc_infer/src/infer/error_reporting/mod.rs:LL:CC
+-Ztrack-diagnostics: created at compiler/rustc_infer\src\infer\error_reporting\mod.rs:LL:CC
```
```diff
--- tests\ui\crashes\ice-7868.stderr
+++ <stderr output>
 warning: unsafe block missing a safety comment
-  --> $DIR/auxiliary/ice-7868-aux.rs:2:5
+  --> $DIR/auxiliary\ice-7868-aux.rs:2:5
    |
 LL |     unsafe { 0 };
```

ui-cargo needs regex escaping as well:

```
regex parse error:
    \\?\D:\Code\Rust\clippy
          ^^
error: unrecognized escape sequence
```

ui-internal:

```diff
actual output differed from expected
--- tests\ui-internal\custom_ice_message.stderr
+++ <stderr output>
-thread '<unnamed>' panicked at 'Would you like some help with that?', clippy_lints/src/utils/internal_lints/produce_ice.rs
+thread '<unnamed>' panicked at 'Would you like some help with that?', clippy_lints/src\utils\internal_lints\produce_ice.rs
 note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

tests/compile-test.rs
```diff
@@ -187,177 +91,96 @@ fn run_internal_tests() {
...
 
-    let res = run_tests(&config, tests);
-    match res {
-        Ok(true) => {},
-        Ok(false) => panic!("Some tests failed"),
-        Err(e) => {
-            panic!("I/O failure during tests: {e:?}");
+    ui_test::run_tests_generic(
+        config,
+        |path| path.ends_with(".rs"),
```
Alexendoo, line 119: The ui-toml suite isn't picking up any tests as this is `Path::ends_with`

```suggestion
        |path| path.extension() == Some("rs".as_ref()),
```

tests/compile-test.rs
```diff
@@ -106,7 +109,8 @@ fn run_ui_toml() {
             .parent()
             .unwrap()
             .display()
-            .to_string(),
+            .to_string()
+            .replace('\\', "/"),
```
Alexendoo, line 113: I think this'd still need to go through `regex::escape`, as canonical windows paths contain `?`, e.g. `\\?\D:\Code\Rust\clippy\`

Manishearth:
What's the status of this from the Clippy side? It needs to be updated, but beyond that is there work to be done?

oli-obk:
I needed to do some refactorings. I do have a windows machine now, so I will test both windows and linux before making a new version. Once that is approved, maybe just delegate to me and I'll do a rebase and land it quickly

Manishearth:
Alright, let us know when ready!

Manishearth:
Needs a lockfile update and a fmt

tests/ui-cargo/cargo_common_metadata/fail/Cargo.stderr
```diff
@@ -0,0 +1,5 @@
+error: unexpected argument '--extern' found
+
+Usage: cargo check --color <WHEN> --quiet
```
Manishearth, line 3: A lot of tests are failing for this reason, probably a bug

Manishearth:
Can the error-to-warning stuff be moved to a separate PR? I'd like to split this up as much as possible, and I don't think the error-to-warning stuff needs uitest to work.

Alternatively, if uitest can get a config to turn off the requirement for `//~ERROR`, that would be great.

oli-obk:
> Needs a lockfile update

Afaict clippy doesn't commit lockfiles. I'll need to teach ui_test to automatically update lockfiles, but not .stderr files

oli-obk:
> Alternatively, if uitest can get a config to turn off the requirement for `//~ERROR`, that would be great.

Yea I think this is the best route. Shouldn't be too hard to add to Yolo mode

oli-obk:
> Can the error-to-warning stuff be moved to a separate PR? I'd like to split this up as much as possible, and I don't think the error-to-warning stuff needs uitest to work.
> 
> Alternatively, if uitest can get a config to turn off the requirement for `//~ERROR`, that would be great.

I reverted all these changes and made it so that `//~ERROR` is not required. Although if `//~ERROR` is specified, it must be correct, so there are a handful (<=5) tests that had their annotations changed.

tests/ui/cast_size_32bit.stderr
```diff
@@ -1,118 +1,183 @@
-error: casting `isize` to `i8` may truncate the value
+warning: casting `isize` to `i8` may truncate the value
```
Alexendoo, line 1: A 32bit only test that needs manually reverting

tests/ui/enum_clike_unportable_variant.stderr
```diff
@@ -1,58 +1,62 @@
-error: C-like enum variant discriminant is not portable to 32-bit targets
+warning: C-like enum variant discriminant is not portable to 32-bit targets
```
Alexendoo, line 1: Similar here, it's not run on x86

tests/ui/fn_to_numeric_cast_32bit.stderr
```diff
@@ -1,144 +1,152 @@
-error: casting function pointer `foo` to `i8`, which truncates the value
+warning: casting function pointer `foo` to `i8`, which truncates the value
```
Alexendoo, line 1: Another 32bit one

tests/ui/macro_use_imports.stderr
```diff
@@ -1,28 +1,16 @@
...
 
 error: `macro_use` attributes are no longer needed in the Rust 2018 edition
   --> $DIR/macro_use_imports.rs:19:5
    |
 LL |     #[macro_use]
-   |     ^^^^^^^^^^^^ help: remove the attribute and import the macro directly, try: `use mac::{pub_macro, function_macro, ty_macro, inner_mod_macro, pub_in_private_macro};`
+   |     ^^^^^^^^^^^^ help: remove the attribute and import the macro directly, try: `use mac::{pub_macro, function_macro, ty_macro, try_err, string_add, mut_mut, inner_mod_macro, pub_in_private_macro};`
 
-error: aborting due to 4 previous errors
+error: aborting due to 2 previous errors
```
Alexendoo, line 15: Not sure why this is no longer working correctly, same with `macro_use_imports_expect.rs`

tests/compile-test.rs
```diff
@@ -166,7 +69,8 @@ fn base_config(test_dir: &str) -> compiletest::Config {
 
 fn run_ui() {
     let mut config = base_config("ui");
-    config.rustfix_coverage = true;
+    config.program.args.push("-Dwarnings".into());
```
Alexendoo, line 72: Could this be moved into `base_config`? So that there's a smaller diff for `ui-toml` tests that emit lints

tests/ui/cast_size_32bit.stderr
oli-obk: I went through all `//@only` and `//@ignore` tests and fixed them.

oli-obk:
> I wonder if we can add a temporary patch to compiletest-rs that allows for the `//@` syntax too

Yea, I wanna do the same thing for the in-rustc-tree compiletest before migrating to ui_test, just to keep the changes happening at the same time small.

I'll create a compiletest-rs PR and a separate clippy PR that does just that.

Manishearth:
If rustc is going to benefit from it too it doesn't even need to be temporary or a setting imo

oli-obk:
Yea, there's an accepted MCP for this: https://github.com/rust-lang/compiler-team/issues/512

oli-obk:
I updated the first commit to also include updated instructions for blessing.

Blessing now works while running tests. So `cargo uibless` or `cargo bless` will do the same thing as `cargo test`, but additionally bless the tests. These commands aren't 100% equivalent though, so if you want to do `cargo bless --features=internal` you need to do the manual thing `cargo test --features=internal -- -- --bless`.

We can pick a different scheme (e.g. use env vars instead of `-- -- --bless`), but I like the command line version as it trivially works on windows, too.

Alexendoo:
bootstrap uses `cargo dev bless` I believe, so we'd want to keep that around temporarily

Alexendoo:
Personally I prefer the environment variable, it works with things like `cargo test -Finternal`/`cargo uitest -Finternal`/`cargo nextest run` without having to pass the rather odd `-- -- --bless` or create a bunch of aliases

tests/compile-test.rs
```diff
@@ -4,160 +4,63 @@
...
-    config.src_base = Path::new("tests").join(test_dir);
-    config.build_base = profile_path.join("test").join(test_dir);
-    config.rustc_path = profile_path.join(if cfg!(windows) {
+    config.program.args.push("--emit=metadata".into());
+    config.program.args.push("-Aunused".into());
+    config.program.args.push("-Zui-testing".into());
+    config.program.args.push("-Dwarnings".into());
+
+    // Normalize away slashes in windows paths.
+    config.stderr_filter(r#"\\"#, "/");
```
Alexendoo, line 59: Could we adapt [the path heuristic from rust's compiletest](https://github.com/rust-lang/rust/blob/6ce22733b973355573efd1e6294e585460e90e17/src/tools/compiletest/src/runtest.rs#L3748-L3775) here? I think it would need support in `ui_test` as it uses a non string `Replacer`

It would be nice to have non path backslashes stay as backslashes as that frequently confuses people writing tests

xFrednet:
> One thing that is a big difference to `compiletest` is that if a test emits _any_ error, you need to mark all of them with `//~ ERROR:` annotations. Since many clippy tests did not have annotations, I changed many lints to be `warn` in their test so that only the `stderr` output is tested.

Would it maybe be possible to make the `//~ ERROR` checks optional? I've been looking at `ui_test` for a different project, where I would like the option not to annotate every test with these

tests/ui/dbg_macro.stderr
```diff
@@ -1,202 +1,2 @@
...
-LL |         dbg!(1);
-   |         ^^^^^^^
-   |
-help: remove the invocation before committing it to a version control system
-   |
-LL |         1;
-   |         ~
-
-error: aborting due to 18 previous errors
+error: Option 'test' given more than once
```
Manishearth: this seems like a bug

tests/ui/crate_level_checks/entrypoint_recursion.stderr
```diff
@@ -1,11 +0,0 @@
-error: recursing into entrypoint `a`
```
Manishearth, line 1: this seems like an unintentional deletion

tests/ui/eq_op.stderr
```diff
@@ -1,172 +1,2 @@
...
-   |                    ^^^^^
-
-error: equal expressions as operands to `==`
-  --> $DIR/eq_op.rs:93:5
-   |
-LL |     (n1.inner.0).0 == (n1.inner.0).0 && (n1.inner.1).0 == (n2.inner.1).0 && (n1.inner.2).0 == (n2.inner.2).0
-   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: aborting due to 28 previous errors
+error: Option 'test' given more than once
```
Manishearth: also a bug

tests/ui/items_after_test_module/block_module.stderr
```diff
@@ -1,17 +1,2 @@
...
-...  |
-LL | |     () => {};
-LL | | }
-   | |_^
-   |
-   = help: move the items to before the testing module was defined
-   = note: `-D clippy::items-after-test-module` implied by `-D warnings`
-
-error: aborting due to previous error
+error: Option 'test' given more than once
```
Manishearth, line 1: here too

tests/ui/macro_use_imports_expect.rs
```diff
@@ -15,16 +15,12 @@ extern crate macro_use_helper as mac;
 extern crate proc_macro_derive as mini_mac;
 
 mod a {
-    #[expect(clippy::macro_use_imports)]
```
Manishearth, line 18: shoudl we not be testing `expect` here, as per the file name?

tests/ui/missing_assert_message.stderr
```diff
@@ -1,131 +1,113 @@
-error: assert without any message
-  --> $DIR/missing_assert_message.rs:14:5
+error[E0425]: cannot find function `foo` in this scope
```
Manishearth, line 1: issue: seems like a bug

tests/ui/missing_inline_proc_macro.stderr
```diff
@@ -0,0 +1,20 @@
+error: the `#[proc_macro]` attribute is only usable with crates of the `proc-macro` crate type
```
Manishearth, line 1: issue: needs to be fixed

tests/ui/needless_pass_by_value_proc_macro.stderr
```diff
@@ -0,0 +1,20 @@
+error: the `#[proc_macro_derive]` attribute is only usable with crates of the `proc-macro` crate type
```
Manishearth, line 1: issue: ditto

tests/ui/proc_macro.stderr
```diff
@@ -1,11 +1,20 @@
-error: approximate value of `f{32, 64}::consts::PI` found
-  --> $DIR/proc_macro.rs:10:14
+error: the `#[proc_macro]` attribute is only usable with crates of the `proc-macro` crate type
```
Manishearth, line 1: issue: ditto

tests/ui/single_call_fn.stderr
```diff
@@ -1,55 +1,2 @@
...
-   | ^^^^^^^^^
-   |
-help: used here
-  --> $DIR/single_call_fn.rs:25:9
-   |
-LL |         j();
-   |         ^
-
-error: aborting due to 4 previous errors
+error: Option 'test' given more than once
```
Manishearth: issue: bug

Manishearth (commented):
I went through all of the changes.

Overall, here's what I see:

 - a bunch of changes wrt the precise compiletest flags
 - some minor changes to output wrt `$DIR` and such, as well as line numbering changes
 - some new `.fixed` files

(Would be nice to have separate commits for these but it's less of a problem now that i've reviewed them; but for future commits might be worth doing commits incrementally and merging in `master` so the review history is preserved)

This all seems okay.

I also found two bugs, one is the "multiple `test` flags" one, and one is where some proc macro errors turn up for a couple tests.

So I think we're quite close to landing this. I'd also be open to landing this as is with an issue filed for fixing those buggy tests.

oli-obk:
You're all too fast ^^ I just pushed it to get it to my windows machine 😅

oli-obk:
> Would it maybe be possible to make the `//~ ERROR` checks optional? I've been looking at `ui_test` for a different project, where I would like the option not to annotate every test with these

yes, that's either `Yolo` or `Fail { require_patterns: false }` mode: https://docs.rs/ui_test/0.11.1/ui_test/enum.Mode.html

xFrednet:
> yes, that's either `Yolo` or `Fail { require_patterns: false }` mode: https://docs.rs/ui_test/0.11.1/ui_test/enum.Mode.html

From the documentation it sounds like `Yolo` requires the annotations. And `Fail`, is fine with missing annotations, as along as the compilation errors. Is my understanding correct?

If so, I was basically looking for `Yolo { require_patterns: false}`

oli-obk:
Yolo is what clippy uses. So it is similar to `Fail { require: false }`, but also allows passing tests.

tests/ui/crate_level_checks/entrypoint_recursion.stderr
oli-obk: I don't think this has worked at all, the other test in the same file doesn't have a `.stderr` file either. I'm guessing compiletest-rs didn't care that there was an extraneous file lying around if the test passes without output?

tests/ui/missing_assert_message.stderr
oli-obk: yea, the test was broken before. It would not compile at all with `--test` passed

tests/compile-test.rs
oli-obk: I opened https://github.com/oli-obk/ui_test/issues/89 to track this

oli-obk:
Ok, CI is happy, testing against rustc CI now, so the next sync doesn't have a nasty surprise

tests/ui/crate_level_checks/entrypoint_recursion.stderr
Manishearth: ah, seems fair. ty!

Manishearth:
I'm ok with merging this now and you and @flip1995 coordinating the sync.

Alexendoo:
Does `cargo dev bless` work after this? If not we can save some upstream headaches by migrating the current test setup to pick up on the `BLESS` env variable before the next sync + update bootstrap to use it ahead of time - https://github.com/rust-lang/rust/blob/6f8c27ae89dfd32895419d7ef5b89844bcad1bcd/src/bootstrap/test.rs#L810C1-L810C1

It probably wouldn't block sync but would block any upstream PR that needs to update clippy tests

oli-obk:
> Does `cargo dev bless` work after this?

no, there is no more way of blessing after running tests. You can only bless *while* running tests. The command for this is `cargo bless` or `cargo uibless` for just blessing the regular `ui` test suite (but not the toml, cargo, ... ones).

flip1995:
I wouldn't block this on the sync. Let's get this merged. :rocket:

Manishearth:
I feel like we have alignment from three team members, so

@bors r+ 

Thanks for your diligence on this!!

PR #10438, opened 2023-03-03
Use `snippet_with_context` more

Jarcho:
No tests at the moment. Need to find a way to write macro tests without writing a pile of macros.

changelog: None


dswij (approved):
LGTM, thanks!

Issue #10433, opened 2023-03-01
Fix tries to omit required binding


HollayHorvath:
### Summary

When running `cargo clippy --fix` on my codebase I get a report that an automatically applied fix breaks the compiler:
```
warning: failed to automatically apply fixes suggested by rustc to crate `lib`

after fixes were automatically applied the compiler reported errors within these files:

  * pangolin/tests/system/mod.rs

This likely indicates a bug in either rustc or cargo itself,
and we would appreciate a bug report! You're likely to see 
a number of compiler warnings after this message which cargo
attempted to fix but failed. If you could open an issue at
https://github.com/rust-lang/rust/issues
quoting the full output of this command we'd be very appreciative!
Note that you may be able to make some more progress in the near-term
fixing code with the `--broken-code` flag

The following errors were reported:
error[E0698]: type inside `async` block must be known in this context
   --> pangolin/tests/system/mod.rs:431:9
    |
431 |         std::future::pending().await;
    |         ^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `T` declared on the function `pending`
    |
note: the type is part of the `async` block because of this `await`
   --> pangolin/tests/system/mod.rs:431:31
    |
431 |         std::future::pending().await;
    |                               ^^^^^^

error[E0698]: type inside `async` block must be known in this context
   --> pangolin/tests/system/mod.rs:431:31
    |
431 |         std::future::pending().await;
    |                               ^^^^^^ cannot infer type for type parameter `T`
    |
note: the type is part of the `async` block because of this `await`
   --> pangolin/tests/system/mod.rs:431:31
    |
431 |         std::future::pending().await;
    |                               ^^^^^^

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0698`.
Original diagnostics will follow.

warning: this let-binding has unit value
   --> pangolin/tests/system/mod.rs:431:9
    |
431 |         let _pending: () = std::future::pending().await;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: omit the `let` binding: `std::future::pending().await;`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value
    = note: `#[warn(clippy::let_unit_value)]` on by default

warning: `pangolin` (test "lib") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 20.68s
```

[...]

PR #10439, opened 2023-03-03
Fix FP for `let_unit_value` when `await` used

giraffate:
Fix https://github.com/rust-lang/rust-clippy/issues/10433

changelog: Fix FP for `let_unit_value` when `await` used


Jarcho:
Thank you. @bors r+

PR #10445, opened 2023-03-04
Lintcheck maintenance

samueltardieu:
Make `cargo lintcheck -j 0` use all threads instead of panicking, and cleanup (and shorten) lintcheck's parsing code.

changelog: none

### DIFF
--- a/lintcheck/src/config.rs
+++ b/lintcheck/src/config.rs
@@ -89,14 +89,11 @@ impl LintcheckConfig {
             if markdown { "md" } else { "txt" }
         ));
 
-        // look at the --threads arg, if 0 is passed, ask rayon rayon how many threads it would spawn and
-        // use half of that for the physical core count
-        // by default use a single thread
+        // look at the --threads arg, if 0 is passed, use the threads count
         let max_jobs = match clap_config.get_one::<usize>("threads") {
             Some(&0) => {
                 // automatic choice
-                // Rayon seems to return thread count so half that for core count
-                rayon::current_num_threads() / 2
+                std::thread::available_parallelism().map(|n| n.get()).unwrap_or(1)
             },
             Some(&threads) => threads,
             // no -j passed, use a single thread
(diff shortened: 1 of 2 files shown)


lintcheck/src/config.rs
```diff
@@ -1,131 +1,70 @@
...
-                // Rayon seems to return thread count so half that for core count
-                rayon::current_num_threads() / 2
-            },
-            Some(&threads) => threads,
-            // no -j passed, use a single thread
-            None => 1,
+        // look at the --threads arg, if 0 is passed, use the threads count
+        if config.max_jobs == 0 {
+            // automatic choice
+            config.max_jobs = num_cpus::get();
```
llogiq, line 55: Why not use `std::thread::available_parallelism()` here and avoid an extra dependency?

llogiq (approved):
Do we need the `num_cpus` dependency here? Otherwise this looks good and I'll gladly r+ it.

lintcheck/src/config.rs
samueltardieu: Sure. I introduced the dependency in the first commit to keep the same semantics as was there before so that commits were self contained, and didn't think twice about removing it. Done.

lintcheck/src/config.rs
```diff
@@ -1,131 +1,70 @@
...
-                // Rayon seems to return thread count so half that for core count
-                rayon::current_num_threads() / 2
-            },
-            Some(&threads) => threads,
-            // no -j passed, use a single thread
-            None => 1,
+        // look at the --threads arg, if 0 is passed, use the threads count
+        if config.max_jobs == 0 {
+            // automatic choice
+            config.max_jobs = std::thread::available_parallelism().map(|n| n.get()).unwrap_or(1);
```
llogiq, line 55:
```suggestion
            config.max_jobs = std::thread::available_parallelism().map_or(1, |n| n.get());
```

samueltardieu: Done. There should be a lint for that.

Neat GitHub interface btw, I didn't know this one.

llogiq:
Thank you! And I think there is a lint for that, I'm curious why it didn't trigger during dogfood.

@bors r+

PR #10448, opened 2023-03-04
Add new `redundant_async_block` lint

### LINKED ISSUE #10444: `async { <FUTURE_EXPR>.await }` can be reduced to `<FUTURE_EXPR>`
### What it does

Awaiting a `Future` a future in a new `async` block is the same as just using the future directly. I think?

### Lint Name

redundant_async_block

### Category

style, complexity

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
block_on(async {
    my_future.await
})
```

Could be written as:

```rust
block_on(my_future)
```

samueltardieu:
Fixes #10444 

changelog: [`redundant_async_block`]: new lint to detect `async { future.await }`

### DIFF
--- a/clippy_lints/src/redundant_async_block.rs
+++ b/clippy_lints/src/redundant_async_block.rs
@@ -1,0 +1,84 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, source::snippet};
+use rustc_ast::ast::*;
+use rustc_ast::visit::Visitor as AstVisitor;
+use rustc_errors::Applicability;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `async` block that only returns `await` on a future.
+    ///
+    /// ### Why is this bad?
+    /// It is simpler and more efficient to use the future directly.
+    ///
+    /// ### Example
+    /// ```rust
+    /// async fn f() -> i32 {
+    ///     1 + 2
+    /// }
+    ///
+    /// let fut = async {
+    ///     f().await
+    /// };
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// async fn f() -> i32 {
+    ///     1 + 2
+    /// }
+    ///
+    /// let fut = f();
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub REDUNDANT_ASYNC_BLOCK,
+    complexity,
+    "`async { future.await }` can be replaced by `future`"
+}
+declare_lint_pass!(RedundantAsyncBlock => [REDUNDANT_ASYNC_BLOCK]);
+
+impl EarlyLintPass for RedundantAsyncBlock {
+    fn check_expr(&mut self, cx: &EarlyContext<'_>, expr: &Expr) {
+        if expr.span.from_expansion() {
+            return;
+        }
+        if let ExprKind::Async(_, _, block) = &expr.kind && block.stmts.len() == 1 &&
+            let Some(Stmt { kind: StmtKind::Expr(last), .. }) = block.stmts.last() &&
+            let ExprKind::Await(future) = &last.kind &&
+            !future.span.from_expansion() &&
+            !await_in_expr(future)
+        {
+            span_lint_and_sugg(
+                cx,
+                REDUNDANT_ASYNC_BLOCK,
+                expr.span,
+                "this async expression only awaits a single future",
+                "you can reduce it to",
+                snippet(cx, future.span, "..").into_owned(),
+                Applicability::MachineApplicable,
+            );
+        }
+    }
+}
+
+/// Check whether an expression contains `.await`
+fn await_in_expr(expr: &Expr) -> bool {
+    let mut detector = AwaitDetector::default();
+    detector.visit_expr(expr);
+    detector.await_found
+}
+
+#[derive(Default)]
+struct AwaitDetector {
+    await_found: bool,
+}
+
+impl<'ast> AstVisitor<'ast> for AwaitDetector {
+    fn visit_expr(&mut self, ex: &'ast Expr) {
+        match (&ex.kind, self.await_found) {
+            (ExprKind::Await(_), _) => self.await_found = true,
+            (_, false) => rustc_ast::visit::walk_expr(self, ex),
+            _ => (),
+        }
+    }
+}
--- a/tests/ui/redundant_async_block.rs
+++ b/tests/ui/redundant_async_block.rs
@@ -1,0 +1,64 @@
+// run-rustfix
+
+#![allow(unused)]
+#![warn(clippy::redundant_async_block)]
+
+async fn func1(n: usize) -> usize {
+    n + 1
+}
+
+async fn func2() -> String {
+    let s = String::from("some string");
+    let f = async { (*s).to_owned() };
+    let x = async { f.await };
+    x.await
+}
+
+macro_rules! await_in_macro {
+    ($e:expr) => {
+        std::convert::identity($e).await
+    };
+}
+
+async fn func3(n: usize) -> usize {
+    // Do not lint (suggestion would be `std::convert::identity(func1(n))`
+    // which copies code from inside the macro)
+    async move { await_in_macro!(func1(n)) }.await
+}
+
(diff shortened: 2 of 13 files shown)


clippy_lints/src/redundant_async_block.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// ### What it does
+    /// Checks for `async` block that only returns `await` on a future.
+    ///
+    /// ### Why is this bad?
+    /// It is shorter to directly use the future.
+    ///
+    /// ### Example
+    /// ```rust
+    /// await {
+    ///     function_returning_a_future().await
```
llogiq, line 17: Might be good if we had an `async fn function_returning_a_future() {}` so the example actually compiles. That applies to both code snippets.

clippy_lints/src/redundant_async_block.rs
```diff
@@ -0,0 +1,51 @@
...
+use rustc_errors::Applicability;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `async` block that only returns `await` on a future.
+    ///
+    /// ### Why is this bad?
+    /// It is shorter to directly use the future.
```
llogiq, line 12: I'm curious: Does the `async await` really get reduced to a single call, or even inlined in release mode? If not, there would even be a perf angle.

clippy_lints/src/redundant_async_block.rs
```diff
@@ -0,0 +1,51 @@
...
+    ///     function_returning_a_future().await
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// function_returning_a_future()
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub REDUNDANT_ASYNC_BLOCK,
+    complexity,
```
llogiq, line 26: Do we have a lintcheck run to ensure we don't get too much churn when introducing this lint?

Nowadays we are quite cautious when adding warn-by-default lints.

clippy_lints/src/redundant_async_block.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+        if let ExprKind::Async(_, _, block) = &expr.kind && block.stmts.len() == 1 &&
+            let Some(Stmt { kind: StmtKind::Expr(last), .. }) = block.stmts.last() &&
+            let ExprKind::Await(future) = &last.kind
+        {
+            span_lint_and_sugg(
+                cx,
+                REDUNDANT_ASYNC_BLOCK,
+                expr.span,
+                "this async-await expression is equivalent to the awaited future itself",
```
llogiq, line 44:
```suggestion
                "this async expression only awaits a single future",
```

tests/ui/redundant_async_block.rs
```diff
@@ -0,0 +1,31 @@
...
+}
+
+fn main() {
+    let fut1 = async { 42 };
+    let fut2 = async { fut1.await };
+
+    let fut1 = async { 42 };
+    let fut2 = async move { fut1.await };
+
+    let fut = async { async { 42 }.await };
```
llogiq, line 24: I'd like to see a test where the future is macro-generated, a test where the whole block *outside* the future / with the future is macro-generated by an crate-internal/external macro each.

The internal ones may lint, but with the correct snippet; the external ones I think should not lint (because we don't expect people to be able to change code external to their crate). I want to be sure the `span.from_expansion()` check is sufficient.

llogiq (commented):
I like this lint, but I think we can make the message shorter. Also I'd want to see some more tests. Finally, before we introduce a warn-by-default lint, we should at least run a lintcheck to get a handle on churn.

clippy_lints/src/redundant_async_block.rs
```diff
@@ -0,0 +1,51 @@
...
+declare_lint_pass!(RedundantAsyncBlock => [REDUNDANT_ASYNC_BLOCK]);
+
+impl EarlyLintPass for RedundantAsyncBlock {
+    fn check_expr(&mut self, cx: &EarlyContext<'_>, expr: &rustc_ast::Expr) {
+        if expr.span.from_expansion() {
+            return;
+        }
+        if let ExprKind::Async(_, _, block) = &expr.kind && block.stmts.len() == 1 &&
+            let Some(Stmt { kind: StmtKind::Expr(last), .. }) = block.stmts.last() &&
+            let ExprKind::Await(future) = &last.kind
```
Alexendoo, line 38: A check for `future.span.from_expansion()` would be good, and an accompanying test for something like

```rust
macro_rules! await_in_macro {
    ($e:expr) => { std::convert::identity($e).await };
} 

async { await_in_macro!(foo()) }
```

This could come up with `futures::poll` [like here](https://github.com/stratis-storage/stratisd/blob/f20f841756f06c310e5352994ef65d1c40040464/src/engine/structures/lock.rs#L1046) but having our own macro definition means we don't have to rely on [`future`'s implementation](https://docs.rs/futures-util/0.3.26/src/futures_util/async_await/poll.rs.html#17) staying the same

samueltardieu:
In addition to all those comments, there is one check missing: there must be no other `.await` in the expression, as-in

```rust
async { func1(func2().await)).await }
```

I'll work on that, probably in the next few days.

clippy_lints/src/redundant_async_block.rs
samueltardieu: Done

clippy_lints/src/redundant_async_block.rs
samueltardieu: If you are talking about a `cargo lintcheck`, yes, I ran one (maybe I should mention it in the description). Is there a way to do something similar to a crater run for lint checking?

clippy_lints/src/redundant_async_block.rs
samueltardieu: Done.

tests/ui/redundant_async_block.rs
samueltardieu: Checking inside a macro needs a bit more work, because the awaiten future must not come from a macro parameter, as it might contain `.await` itself.

clippy_lints/src/redundant_async_block.rs
samueltardieu: You're right, without the test on the `future` itself the proposed solution involved copying code coming from the macro. Done.

clippy_lints/src/redundant_async_block.rs
samueltardieu: I call still see the await process in the MIR expansion in release mode, so I don't think this is optimized away.

tests/ui/redundant_async_block.rs
samueltardieu: Right now, it doesn't look into macros at all. I've added a `MISSED OPPORTUNITY` marker, but I'm not sure how to check inside the macro that no parameter (or another macro) is used as the `.await` receiver.

llogiq: That's all right. :+1:

llogiq:
What did your lintcheck run come up with?

I might do a run to reproduce and see what it turns up.

samueltardieu:
> What did your lintcheck run come up with?

> I might do a run to reproduce and see what it turns up.

Nothing, as expected.

```bash
$ cargo lintcheck -j 0 --filter redundant_async_block
[…]
$ cat lintcheck-logs/lintcheck_crates_logs.txt
clippy 0.1.69

### Reports



### Stats:

### ICEs:
```

> I might do a run to reproduce and see what it turns up.

You need to rebase this branch on `master` to use `-j 0`, tell me if you want me to push my updated branch.

llogiq:
r=me when rebased.

samueltardieu:
Checked with the 500 most recently downloaded crates of crates.io: this lint was not triggered (the complete 470klines report is available at <https://rfc1149.net/tmp/top500_logs.txt.gz>).

llogiq:
Yeah, I also ran lintcheck. Thanks everyone!

@bors r+

samueltardieu:
@llogiq: rebased

llogiq:
Thanks again!

@bors r+

ComputerDruid:
#10509 notes that this lint triggers even in cases where the suggestion is a behavior change.

PR #10454, opened 2023-03-05
Wrap `transmutes_expressible_as_ptr_casts` suggestions in parentheses

### LINKED ISSUE #10449: FP transmutes_expressible_as_ptr_casts: type fn() cannot be dereferenced
### Summary

.

### Lint Name

transmutes_expressible_as_ptr_casts

### Reproducer

I tried this code:

```rust
fn main() {
    fn f() {}

    let _x: u8 = unsafe {
        *std::mem::transmute::<fn(), *const u8>(f)
    };
}

```

I saw this happen:

```
warning: transmute from `fn()` to `*const u8` which could be expressed as a pointer cast instead
 --> src/main.rs:5:10
  |
5 |         *std::mem::transmute::<fn(), *const u8>(f)
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try: `f as *const u8`
  |
```
Suggestion does not compile:
```
error[E0614]: type `fn() {main::f}` cannot be dereferenced
 --> src/main.rs:5:9
  |
5 |         *f as *const u8
  |         ^^

```

### Version

```text
binary: rustc
commit-date: 2023-03-03
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

Alexendoo:
changelog: [`transmutes_expressible_as_ptr_casts`]: Fix suggestion missing wrapping parentheses

Fixes #10449

r? @Jarcho

Is this the best way to go about this? `unused_parens` will catch the unnecessary ones but emitting them in the first place isn't ideal

### DIFF
--- a/clippy_lints/src/transmute/transmutes_expressible_as_ptr_casts.rs
+++ b/clippy_lints/src/transmute/transmutes_expressible_as_ptr_casts.rs
@@ -2,8 +2,9 @@ use super::utils::check_cast;
 use super::TRANSMUTES_EXPRESSIBLE_AS_PTR_CASTS;
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::sugg::Sugg;
+use rustc_ast::ExprPrecedence;
 use rustc_errors::Applicability;
-use rustc_hir::Expr;
+use rustc_hir::{Expr, Node};
 use rustc_lint::LateContext;
 use rustc_middle::ty::{cast::CastKind, Ty};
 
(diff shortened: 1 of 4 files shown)


Jarcho:
You can check if the parent expression's precedence is `>=PREC_PREFIX`. I don't think as casting has a constant.

Jarcho:
ping @Alexendoo

Jarcho:
Thank you. @bors r+

PR #10458, opened 2023-03-05
lintcheck: use multithreading unless --fix or --recursive is used

samueltardieu:
Use multithreading unless there is a reason not to.

changelog: none

### DIFF
--- a/lintcheck/src/config.rs
+++ b/lintcheck/src/config.rs
@@ -1,10 +1,16 @@
 use clap::Parser;
-use std::path::PathBuf;
+use std::{num::NonZeroUsize, path::PathBuf};
 
 #[derive(Clone, Debug, Parser)]
 pub(crate) struct LintcheckConfig {
-    /// Number of threads to use, 0 automatic choice
-    #[clap(long = "jobs", short = 'j', value_name = "N", default_value_t = 1)]
+    /// Number of threads to use (default: all unless --fix or --recursive)
+    #[clap(
+        long = "jobs",
+        short = 'j',
+        value_name = "N",
+        default_value_t = 0,
+        hide_default_value = true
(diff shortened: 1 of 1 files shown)


samueltardieu:
r? @llogiq 
(as you dealt with other lintcheck changes that I proposed recently)

lintcheck/src/config.rs
```diff
@@ -50,9 +55,12 @@ impl LintcheckConfig {
...
         // look at the --threads arg, if 0 is passed, use the threads count
-        if config.max_jobs == 0 {
-            // automatic choice
-            config.max_jobs = std::thread::available_parallelism().map_or(1, |n| n.get());
+        if config.max_jobs.is_none() {
+            config.max_jobs = Some(if config.fix || config.recursive {
+                1
+            } else {
+                std::thread::available_parallelism().map_or(1, |n| n.get())
+            });
```
llogiq, line 63: Why not make `config.max_jobs` a plain `usize` with a `#[default = 0]` and then first set it to `1` if `config.fix || config.recursive`, then if it is still zero, set it to `available_parallelism` if available, 1 otherwise?

llogiq:
I don't think we need an `Option` for `max_jobs`. Otherwise this looks ok to me.

lintcheck/src/config.rs
samueltardieu: Indeed. Done.

llogiq:
Thank you!

@bors r+

samueltardieu:
@llogiq: I have added the missing "changelog: none" in the PR description

llogiq:
Thanks!

@bors r+

llogiq:
Seems like a transient error

@bors retry

PR #10470, opened 2023-03-09
Fix semicolon insertion in `match_single_binding`

### LINKED ISSUE #10447: match_single_binding should ommit entire statement instead of suggestion broken "()" ?
### Summary

.

### Lint Name

match_single_binding

### Reproducer

I tried this code:

```rust
pub fn a() -> usize {
    match 1 {
        _ => (),
    }
    2
}

pub fn main() {}

```

Clippy will suggest 
````rust
pub fn a() -> usize {
    ()
    2
}
````

but this will not build
````

error: expected `;`, found `2`
 --> src/main.rs:2:4
  |
2 |     ()
  |       ^ help: add `;` here
3 |     2
  |     - unexpected token
````

maybe we can make it `();` or just suggest removal of the entire match?

### Version

```text
rustc 1.69.0-nightly (b1719530f 2023-03-04)
binary: rustc
commit-hash: b1719530f44e3c8ec903f76020a52bd8764d5d10
commit-date: 2023-03-04
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

Alexendoo:
changelog: [`match_single_binding`]: Fix missing semicolon after the suggestion

Fixes #10447

Also fixes an edge case for unit returning macros in expression contexts:

```rust
f(match 1 {
    _ => println!("foo"),
});
```

would suggest

```rust
f(println!("foo"););
```

### DIFF
--- a/clippy_lints/src/matches/match_single_binding.rs
+++ b/clippy_lints/src/matches/match_single_binding.rs
@@ -3,7 +3,7 @@ use clippy_utils::macros::HirNode;
 use clippy_utils::source::{indent_of, snippet, snippet_block_with_context, snippet_with_applicability};
 use clippy_utils::{get_parent_expr, is_refutable, peel_blocks};
 use rustc_errors::Applicability;
-use rustc_hir::{Arm, Expr, ExprKind, Node, PatKind};
+use rustc_hir::{Arm, Expr, ExprKind, Node, PatKind, StmtKind};
 use rustc_lint::LateContext;
 use rustc_span::Span;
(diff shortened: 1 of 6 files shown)


giraffate (approved):
Overall looks good, thanks!

Can the conflicts be resolved?

Alexendoo:
Woops, I saw that and then forgot about it. Rebased thanks

giraffate:
@bors r+

Thanks!

PR #10479, opened 2023-03-10
Don't lint `manual_clamp` in const contexts.

### LINKED ISSUE #10474: manual_clamp suggests using clamp() in a const function
### Summary

I'm manually reimplementing `clamp()` because the built-in version is not `const`. The `manual_clamp` rule notifies me about this.
It would be nice if it could recognize that i'm `const` and therefore the suggestion is not applicable.

### Lint Name

manual_clamp

### Reproducer

I tried this code:

```rust
#[deny(clippy::nursery)]
const fn clamp(x: u32, min: u32, max: u32) -> u32 {
    if x < min {
        min
    } else if x > max {
        max
    } else {
        x
    }
}
```

I saw this happen:

```
error: clamp-like pattern without using clamp function
 --> src/main.rs:3:5
  |
3 | /     if x < min {
4 | |         min
5 | |     } else if x > max {
6 | |         max
7 | |     } else {
8 | |         x
9 | |     }
  | |_____^ help: replace with clamp: `x.clamp(min, max)`
```

no warning

### Version

```text
rustc 1.70.0-nightly (39f2657d1 2023-03-09)
binary: rustc
commit-hash: 39f2657d1101b50f9b71ae460b762d330cc8426b
commit-date: 2023-03-09
release: 1.70.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

Jarcho:
fixes #10474

Probably worth including in the sync.
r? @flip1995 

changelog: [`manual_clamp`]: Don't lint in const contexts.

### DIFF
--- a/clippy_lints/src/manual_clamp.rs
+++ b/clippy_lints/src/manual_clamp.rs
@@ -6,7 +6,8 @@ use clippy_utils::ty::implements_trait;
 use clippy_utils::visitors::is_const_evaluatable;
 use clippy_utils::MaybePath;
 use clippy_utils::{
-    eq_expr_value, is_diag_trait_item, is_trait_method, path_res, path_to_local_id, peel_blocks, peel_blocks_with_stmt,
+    eq_expr_value, in_constant, is_diag_trait_item, is_trait_method, path_res, path_to_local_id, peel_blocks,
+    peel_blocks_with_stmt,
 };
 use itertools::Itertools;
 use rustc_errors::Applicability;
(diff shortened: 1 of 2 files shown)


flip1995:
@bors r+

Thanks!

This is a nice fix, but I already did the sync https://github.com/rust-lang/rust/pull/108974

Since the release was yesterday, this is not really time critical IMO. So I will just include it in the next sync.

Jarcho:
I noticed the lint was in the nursery after making this.

PR #10481, opened 2023-03-10
Add `allow_attribute` lint

### LINKED ISSUE #10468: Lint: Suggestion `#[expect]` over `#[allow]`
### What it does

The `#![feature(lint_reasons)]` adds a new `#[expect]` attribute as described in [RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html).

It would be cool to have a restriction lint, that suggests replacing all `#[allow]` attributes with `#[expect]`.

### Lint Name

allow_attribute

### Category

restriction

### Notes

* This should be a restriction lint, as the `allow` attribute has valid use cases.
* The lint has to check for the `lint_reasons` feature.
* The lint should probably ignore crate attributes, or at least have an option to allow them. Crate attributes are `#![attr]` in the root of the crate.

### Example

```rust
#[allow(unused)]
let x = 0;
```

Could be written as:

```rust
#[expect(unused)]
let x = 0;
```

blyxyas:
Fixes #10468 

changelog: new lint: [`allow_attributes`]
[#10481](https://github.com/rust-lang/rust-clippy/pull/10481)


clippy_lints/src/allow_attribute.rs
```diff
@@ -0,0 +1,93 @@
...
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::{symbol::Ident, BytePos};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects uses of the `#[allow]` attribute and suggests to replace it with the new `#[expect]` attribute implemented by `#![feature(lint_reasons)]` ([RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html))
+    /// ### Why is this bad?
+    /// Using `#[allow]` isn't bad, but `#[expect]` may be preferred as it lints if the code **doesn't** produce a warning.
+    /// ### Example
```
xFrednet, line 14: What do you think of this description? I think it's a bit clearer :)
```suggestion
    /// Detects uses of the `#[allow]` attribute and suggests replacing it with
    /// the `#[expect]` (See [RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html))
    ///
    /// The expect attribute is still unstable and requires the `lint_reasons`
    /// on nightly. It can be enabled by adding `#![feature(lint_reasons)]` to
    /// the crate root.
    ///
    /// This lint only warns outer attributes (`#[allow]`), as inner attributes
    /// (`#![allow]`) are usually used to enable or disable lints on a global scale. 
    /// 
    /// ### Why is this bad?
    /// 
    /// `#[expect]` attributes suppress the lint emission, but emit a warning, if
    /// the expectation is unfulfilled. This can be useful to be notified when the
    /// lint is no longer triggered.
    ///
    /// ### Example
```

clippy_lints/src/allow_attribute.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// ```rust,ignore
```
xFrednet, line 15: Do we need the ignore here? :)

clippy_lints/src/allow_attribute.rs
```diff
@@ -0,0 +1,93 @@
...
+                    format!("#[expect{})]", snippet(
+                        cx,
+                        attr.ident().unwrap().span
+                        .with_lo(
+                            attr.ident().unwrap().span.hi() + BytePos(2) // Cut [(
+                        )
+                        .with_hi(
+                            attr.meta().unwrap().span.hi() - BytePos(2) // Cut )]
+                        )
+                        , "...")), Applicability::MachineApplicable);
```
xFrednet, line 89: I think it would be better to use `if Some(ident) = attr.ident()` above and then use the span of the ident. That can avoid the span magic and code snipping. It might also work with `cfg_attr` attributes.

tests/ui/allow_attribute.fixed
```diff
@@ -0,0 +1,18 @@
...
+#![warn(clippy::allow_attribute)]
+#![feature(lint_reasons)]
+
+fn main() {}
+
+// Using clippy::needless_borrow just as a placeholder, it isn't relevant.
+
+// Should lint
+#[expect(dead_code)]
+struct T1;
```
xFrednet, line 12: Could you also add a test with `cfg_attr` I want to see if that work :)

```
// `panic = "unwind"` should always be true
#[cfg_attr(panic = "unwind", allow(dead_code))]
struct CFG_T;
```

I would also like a function with an inner attribute:

```
fn ignore_inner_attr() {
	#![allow(unused)] // Should not lint
}
```

tests/ui/allow_attribute.rs
```diff
@@ -0,0 +1,18 @@
+// run-rustfix
+#![allow(unused)]
+#![warn(clippy::allow_attribute)]
+#![feature(lint_reasons)]
+
+fn main() {}
+
+// Using clippy::needless_borrow just as a placeholder, it isn't relevant.
```
xFrednet, line 8: What do you mean with this comment? :sweat_smile:

xFrednet (commented):
This is a good start, I've left some comments, how I believe the code can be improved. You're welcome to ask if you have any questions :)

tests/ui/allow_attribute.rs
blyxyas: T3 and T4 have attributes `#[deny(clippy::needless_borrow)]` and `#[warn(clippy::needless_borrow)]` to check that the lint doesn't change anything except `allow`. So it means that the lint `needless_borrow` is used but isn't meaningful and could be replaced.

clippy_lints/src/allow_attribute.rs
blyxyas: If `,ignore` is removed CI will fail because it should be:

```rust
 #[allow(unused_mut)]
 fn foo() -> usize {
    let mut a: Vec<TYPE> = Vec::new();
    a.len()
}
```

I consider that adding that type annotation doesn't help showing what the lint does, and clutters the example making it harder to understand these examples.

xFrednet: Makes sense, you could just use `vec![1]` to initialize it, that would give it the default integer type :). I prefer the example to be tested, but it's not necessary :)

tests/ui/allow_attribute.fixed
blyxyas: Added

clippy_lints/src/allow_attribute.rs
blyxyas: Fixed

clippy_lints/src/allow_attribute.rs
blyxyas: Fixed

clippy_lints/src/allow_attribute.rs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        // Check outer attributes
+
+        if_chain! {
+            if let AttrStyle::Outer = attr.style;
+            if attr.ident()
+            .unwrap_or(Ident::with_dummy_span(sym!(empty))) // Will not trigger if doesn't have an ident.
+            .name == sym!(allow);
```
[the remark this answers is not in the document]
blyxyas: Fixed

clippy_lints/src/allow_attribute.rs
```diff
@@ -0,0 +1,93 @@
...
+            .unwrap_or(Ident::with_dummy_span(sym!(empty))) // Will not trigger if doesn't have an ident.
+            .name == sym!(feature);
+            if let ast::AttrKind::Normal(normal) = &attr.kind;
+            if let Some(MetaItemKind::List(list)) = normal.item.meta_kind();
+            if let Some(symbol) = list.get(0);
+            if symbol.ident().unwrap().name == sym!(lint_reasons);
+            then {
+                self.lint_reasons_active = true;
+            }
+        }
```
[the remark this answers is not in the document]
blyxyas: Fixed

CHANGELOG.md
```diff
@@ -4382,6 +4382,7 @@ Released 2018-09-13
 <!-- begin autogenerated links to lint list -->
 [`absurd_extreme_comparisons`]: https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons
 [`alloc_instead_of_core`]: https://rust-lang.github.io/rust-clippy/master/index.html#alloc_instead_of_core
+[`allow_attribute`]: https://rust-lang.github.io/rust-clippy/master/index.html#allow_attribute
```
xFrednet, line 4385: Small NIT that I just noticed. Lint names should ideally be plural. Could you rename the lint to `allow_attributes`? We could also choose `allow_attrs` as an abbreviation, but I think the long name is a bit better, since other lints also use the version :)

clippy_lints/src/allow_attribute.rs
```diff
@@ -0,0 +1,79 @@
...
+                    //     cx,
+                    //     ident.span
+                    //     .with_lo(
+                    //         ident.span.hi() + BytePos(2) // Cut *(
+                    //     )
+                    //     .with_hi(
+                    //         attr.meta().unwrap().span.hi() - BytePos(1) // Cut )
+                    //     )
+                    //     , "..."))
+                    , Applicability::MachineApplicable);
```
xFrednet, line 75: Small formatting nit, and removing commented code :)
```suggestion
                    "replace it with",
                    "expect".into(),
                    Applicability::MachineApplicable,
                );
```

It's a bit annoying, that rustfmt doesn't work with let-chains or the `if_chain` macro -.-

tests/ui/allow_attribute.stderr
```diff
@@ -0,0 +1,16 @@
...
+LL | #[allow(dead_code)]
+   |   ^^^^^ help: replace it with: `expect`
+   |
+   = note: `-D clippy::allow-attribute` implied by `-D warnings`
+
+error: #[allow] attribute found
+  --> $DIR/allow_attribute.rs:20:30
+   |
+LL | #[cfg_attr(panic = "unwind", allow(dead_code))]
+   |                              ^^^^^ help: replace it with: `expect`
```
xFrednet, line 13: Really nice, that this also works!!!

xFrednet (commented):
Four nits, that should be simple to fix and then this should be ready for merge. Thank you for the swift updates! :D

PR #10483, opened 2023-03-11
New lint to detect `&std::path::MAIN_SEPARATOR.to_string()`

### LINKED ISSUE #10480: Suggest `std::path::MAIN_SEPARATOR_STR`
### What it does

Suggests `std::path::MAIN_SEPARATOR_STR` over `std::path::MAIN_SEPARATOR.to_string()`

### Lint Name

_No response_

### Category

_No response_

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
&std::path::MAIN_SEPARATOR.to_string()
```

Could be written as:

```rust
std::path::MAIN_SEPARATOR_STR
```

samueltardieu:
Fixes #10480 

changelog: [`manual_main_separator_str`] new lint

### DIFF
--- a/tests/ui/manual_main_separator_str.fixed
+++ b/tests/ui/manual_main_separator_str.fixed
@@ -1,0 +1,39 @@
+// run-rustfix
+
+#![allow(unused)]
+#![warn(clippy::manual_main_separator_str)]
+
+use std::path::MAIN_SEPARATOR;
+
+fn len(s: &str) -> usize {
+    s.len()
+}
+
+struct U<'a> {
+    f: &'a str,
+    g: &'a String,
+}
+
+struct V<T> {
+    f: T,
+}
+
+fn main() {
+    // Should lint
+    let _: &str = std::path::MAIN_SEPARATOR_STR;
+    let _ = len(std::path::MAIN_SEPARATOR_STR);
+    let _: Vec<u16> = std::path::MAIN_SEPARATOR_STR.encode_utf16().collect();
+
+    // Should lint for field `f` only
+    let _ = U {
+        f: std::path::MAIN_SEPARATOR_STR,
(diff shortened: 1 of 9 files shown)


llogiq:
I wonder if we could extend the `unnecessary_to_owned` lint instead of creating a new one. Also doesn't that work with all `&str` constants? Why single out `MAIN_SEPARATOR`?

samueltardieu:
> I wonder if we could extend the `unnecessary_to_owned` lint instead of creating a new one. Also doesn't that work with all `&str` constants? Why single out `MAIN_SEPARATOR`?

It depends on the existence of a `&str` constant (`std::path::MAIN_SEPARATOR_STR`) with the same content as the `char` constant (`std::path::MAIN_SEPARATOR`).

tests/ui/manual_main_separator_str.fixed
```diff
@@ -0,0 +1,14 @@
+// run-rustfix
+
+#![allow(unused)]
+#![warn(clippy::manual_main_separator_str)]
+
+use std::path::MAIN_SEPARATOR;
+
+fn main() {
```
llogiq, line 8: What about `let`s without a given type? Or other expressions that are not direct local bindings?

llogiq (commented):
Ah, I hadn't noticed the difference here.

I'd like to see some more tests, otherwise this looks good.

tests/ui/manual_main_separator_str.fixed
samueltardieu: I have added new tests, such as initializing a `struct` field with this expression or passing it as a function argument.

I've also made it work when it is a method receiver (see the test with `encode_utf16()`), without an explicit `&`.

samueltardieu:
@llogiq Is there anything more you'd like me to do on this one?

PR #10484, opened 2023-03-11
Migrate `format_args.rs` to `rustc_ast::FormatArgs`

Alexendoo:
changelog: none

Part of https://github.com/rust-lang/rust-clippy/issues/10233

Empty precision specifiers are no longer linted as the span for that isn't present in [`FormatOptions`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/format/struct.FormatOptions.html)

```rust
format!("{:.}", ...)
```

That could be fixed later with some hackery or a change upstream

r? @flip1995


clippy_lints/src/utils/format_args_collector.rs
```diff
@@ -15,9 +20,79 @@ declare_clippy_lint! {
...
+            }
+        }
+
+        if !seen_comma {
+            return true;
+        }
+    }
+
+    false
+}
```
Alexendoo, line 98: This is a slightly simplified version of

https://github.com/rust-lang/rust-clippy/blob/f19db28361c74195e7baffeca23915549c0805ca/clippy_utils/src/macros.rs#L932-L992

tests/ui/uninlined_format_args.stderr
```diff
@@ -774,18 +774,6 @@ LL -     println!("{:w$.p$}", w = width, p = prec);
...
-   |
-LL |     println!("{}", format!("{}", local_i32));
-   |                    ^^^^^^^^^^^^^^^^^^^^^^^^
-   |
-help: change this to
-   |
-LL -     println!("{}", format!("{}", local_i32));
-LL +     println!("{}", format!("{local_i32}"));
-   |
-
```
Alexendoo, line 788: This is a bug with `root_macro_call_first_node`, not sure why but it doesn't pick up that `format!()` as a macro call

clippy_lints/Cargo.toml
```diff
@@ -9,6 +9,7 @@ keywords = ["clippy", "lint", "plugin"]
 edition = "2021"
 
 [dependencies]
+arrayvec = { version = "0.7", default-features = false }
```
Alexendoo, line 12: `clippy_utils` depends on this already so it's not a new dependency

m-ou-se:
> Empty precision specifiers are no longer linted as the span for that isn't present in [`FormatOptions`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/format/struct.FormatOptions.html)
> 
> ```rust
> format!("{:.}", ...)
> ```

We should probably just move that warning to rustc, since an empty `.` specifier is always useless.

nyurik:
I tried to apply a similar change to [explicit_write.rs](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/explicit_write.rs) , and `root_macro_call_first_node` wasn't a drop-in replacement for `FormatArgsExpn::parse`.  Most likely i am doing something totally wrong.  Something like this partially works, but is probably wrong as well:

```rust
if let ExprKind::MethodCall(unwrap_fun, write_call, [], _) = expr.kind
    && unwrap_fun.ident.name == sym::unwrap
    // match call to write_fmt
    && let ExprKind::MethodCall(write_fun, write_recv, [write_arg], _) = look_in_block(cx, &write_call.kind)
    && let ExprKind::Call(_, write_arg2) = look_in_block(cx, &write_arg.kind)
    && write_fun.ident.name == sym!(write_fmt)
    // match calls to std::io::stdout() / std::io::stderr ()
    && let Some(dest_name) = if match_function_call(cx, write_recv, &paths::STDOUT).is_some() {
        Some("stdout")
    } else if match_function_call(cx, write_recv, &paths::STDERR).is_some() {
        Some("stderr")
    } else {
        None
    }
    && let Some(format_args) = root_macro_call_first_node(cx, write_call)
    && is_format_macro(cx, format_args.def_id)
    // && let Some(format_args2) = FormatArgsExpn::find_nested(cx, expr, format_args.expn)
```

Alexendoo:
The replacement would be `find_format_args`. The `root_macro_call_first_node` would be detecting the `write[ln]` macro call there, not `format_args`

You wouldn't need `is_format_macro` as it's already checking the macro kind below that, but if you're using `root_macro_call_first_node` it could replace the two `is_expn_of` checks with a `match cx.tcx.get_diagnostic_name(macro_call.def_id) { .. }`

Manishearth (approved):

PR #10487, opened 2023-03-11
Gramar, and spelin kleanup

nyurik:
A few minor cleanups in various markdown files, mostly focusing on spelling and ignoring non-compilable codeblocks. Also a few markdown tables, etc.

P.S. I'm no grammar expert, do take with a grain of salt.

changelog: none

### DIFF
--- a/book/src/development/common_tools_writing_lints.md
+++ b/book/src/development/common_tools_writing_lints.md
@@ -3,7 +3,7 @@
 You may need following tooltips to catch up with common operations.
 
 - [Common tools for writing lints](#common-tools-for-writing-lints)
-  - [Retrieving the type of an expression](#retrieving-the-type-of-an-expression)
+  - [Retrieving the type of expression](#retrieving-the-type-of-expression)
   - [Checking if an expr is calling a specific method](#checking-if-an-expr-is-calling-a-specific-method)
   - [Checking for a specific type](#checking-for-a-specific-type)
   - [Checking if a type implements a specific trait](#checking-if-a-type-implements-a-specific-trait)
@@ -16,7 +16,7 @@ Useful Rustc dev guide links:
 - [Type checking](https://rustc-dev-guide.rust-lang.org/type-checking.html)
 - [Ty module](https://rustc-dev-guide.rust-lang.org/ty.html)
 
-## Retrieving the type of an expression
+## Retrieving the type of expression
 
 Sometimes you may want to retrieve the type `Ty` of an expression `Expr`, for
 example to answer following questions:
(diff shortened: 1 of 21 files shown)


book/src/development/common_tools_writing_lints.md
```diff
@@ -16,7 +16,7 @@ Useful Rustc dev guide links:
 - [Type checking](https://rustc-dev-guide.rust-lang.org/type-checking.html)
 - [Ty module](https://rustc-dev-guide.rust-lang.org/ty.html)
 
-## Retrieving the type of an expression
+## Retrieving the type of expression
```
xFrednet, line 19:
```suggestion
## Retrieving the type of expressions
```

book/src/development/common_tools_writing_lints.md
```diff
@@ -192,7 +192,7 @@ functions to deal with macros:
 - `span.from_expansion()`: detects if a span is from macro expansion or
   desugaring. Checking this is a common first step in a lint.
 
-   ```rust
+   ```rust,ignore
```
xFrednet, line 195: The `ignore` flag is AFAIK only used by rustdoc, I know that there are tools to also run the code in md book, though. Is there a reason why you think it's better to add them? For me, it seems like this expects a certain type of markdown to be used and makes it less compatible for little reason :thinking: 

The rendering seems to work on GH just fine either way:
```rust,ignore
let x = 0;
```

xFrednet (commented):
One small question, while skimming over the changes :)

book/src/development/common_tools_writing_lints.md
nyurik: @xFrednet oops, just saw your reply, thx for patience :)  The reason I added it there was because some IDEs (and most like all IDEs in the future) treat any code inside the triple backticks as valid code, and try to help with it. In this case, the code is incorrect (e.g. it has a tripple dot instead of a real code: `fn main() { ... }`) -- which means IDEs show it as an error.

book/src/lint_configuration.md
```diff
@@ -195,7 +195,7 @@ The maximum cognitive complexity a function can have
 ### disallowed-names
 The list of disallowed names to lint about. NB: `bar` is not here since it has legitimate uses. The value
 `".."` can be used as part of the list to indicate, that the configured values should be appended to the
-default configuration of Clippy. By default any configuration will replace the default value.
+default configuration of Clippy. By default, any configuration will replace the default value.
```
xFrednet, line 199: This file is auto generated, from the doc comments in `clippy_lints::util::conf`. You can update the strings there and then run `cargo collect-metadata` that should update this file and make bors happy :)

nyurik: thanks @xFrednet , fixed.  Storing auto-generated content in git is evil... but alas.

nyurik:
@dswij or @xFrednet can you re-bors it please?

nyurik:
@dswij I think you need to bors+ it

PR #10490, opened 2023-03-12
Do not propose to remove `async move` if variables are captured by ref

### LINKED ISSUE #10482: redundant_async_block results in broken code for `async fn f(&self)`
### Summary

auto fix of redundant_async_block can lead to broken code in some circumstances

### Lint Name

redundant_async_block

### Reproducer

I tried this code:

```rust
#[derive(Debug, Clone)]
struct F {}

impl F {

    async fn run(&self) {}
}

pub async fn run() {
    let f = F{};
    let c = f.clone();
    tokio::task::spawn(async move {c.run().await});
    let _f = f;
}
```

repro: https://github.com/mattsse/redundant-async-block-broken-code

I saw this happen:

```

  * src/lib.rs

error[E0597]: `c` does not live long enough
  --> src/lib.rs:16:24
   |
16 |     tokio::task::spawn(c.run());
   |                        ^^^^^^^
   |                        |
   |                        borrowed value does not live long enough
   |                        argument requires that `c` is borrowed for `'static`
17 |     let _f = f;
18 | }
   | - `c` dropped here while still borrowed

For more information about this error, try `rustc --explain E0597`.

warning: this async expression only awaits a single future
  --> src/lib.rs:16:24
   |
16 |     tokio::task::spawn(async move {c.run().await});
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: you can reduce it to: `c.run()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_async_block
   = note: `#[warn(clippy::redundant_async_block)]` on by default
```

### Version

```text
cargo 1.70.0-nightly (7d3033d2e 2023-03-08)
rustc 1.70.0-nightly (ff4b772f8 2023-03-10)
```

### Additional Labels

_No response_

samueltardieu:
Fixes #10482 

changelog: FP [`redundant_async_block`] Do not propose to remove `async move` if variables are captured by ref

### DIFF
--- a/clippy_lints/src/redundant_async_block.rs
+++ b/clippy_lints/src/redundant_async_block.rs
@@ -32,7 +32,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.69.0"]
     pub REDUNDANT_ASYNC_BLOCK,
-    complexity,
+    nursery,
     "`async { future.await }` can be replaced by `future`"
 }
 declare_lint_pass!(RedundantAsyncBlock => [REDUNDANT_ASYNC_BLOCK]);
@@ -48,6 +48,11 @@ impl EarlyLintPass for RedundantAsyncBlock {
             !future.span.from_expansion() &&
             !await_in_expr(future)
         {
+            if captures_value(last) {
+                // If the async block captures variables then there is no equivalence.
+                return;
+            }
+
             span_lint_and_sugg(
                 cx,
                 REDUNDANT_ASYNC_BLOCK,
@@ -77,6 +82,36 @@ impl<'ast> AstVisitor<'ast> for AwaitDetector {
     fn visit_expr(&mut self, ex: &'ast Expr) {
         match (&ex.kind, self.await_found) {
             (ExprKind::Await(_), _) => self.await_found = true,
+            (_, false) => rustc_ast::visit::walk_expr(self, ex),
+            _ => (),
+        }
+    }
+}
+
+/// Check whether an expression may have captured a local variable.
+/// This is done by looking for paths with only one segment, except as
+/// a prefix of `.await` since this would be captured by value.
+///
+/// This function will sometimes return `true` even tough there are no
+/// captures happening: at the AST level, it is impossible to
+/// dinstinguish a function call from a call to a closure which comes
+/// from the local environment.
+fn captures_value(expr: &Expr) -> bool {
+    let mut detector = CaptureDetector::default();
+    detector.visit_expr(expr);
+    detector.capture_found
+}
+
+#[derive(Default)]
+struct CaptureDetector {
+    capture_found: bool,
+}
+
+impl<'ast> AstVisitor<'ast> for CaptureDetector {
+    fn visit_expr(&mut self, ex: &'ast Expr) {
+        match (&ex.kind, self.capture_found) {
+            (ExprKind::Await(fut), _) if matches!(fut.kind, ExprKind::Path(..)) => (),
+            (ExprKind::Path(_, path), _) if path.segments.len() == 1 => self.capture_found = true,
             (_, false) => rustc_ast::visit::walk_expr(self, ex),
             _ => (),
         }
--- a/tests/ui/redundant_async_block.rs
+++ b/tests/ui/redundant_async_block.rs
@@ -3,6 +3,8 @@
 #![allow(unused)]
 #![warn(clippy::redundant_async_block)]
 
+use std::future::Future;
+
 async fn func1(n: usize) -> usize {
     n + 1
 }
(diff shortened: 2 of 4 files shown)


samueltardieu:
The `captured_by_ref()` function can have false positives, which errs on the safe side. Is there a way to go from a `NodeId` to the corresponding AST in an `EarlyLintPass`? That would let us check exactly what is captured, by checking the closure built from the `Async` AST node.

OliverNChalk:
Is there an estimated release date for this, our codebase is running into this quite a bit when we launch tasks to drive components

clippy_lints/src/redundant_async_block.rs
```diff
@@ -42,12 +42,17 @@ impl EarlyLintPass for RedundantAsyncBlock {
...
             return;
         }
-        if let ExprKind::Async(_, _, block) = &expr.kind && block.stmts.len() == 1 &&
+        if let ExprKind::Async(capture, _, block) = &expr.kind && block.stmts.len() == 1 &&
             let Some(Stmt { kind: StmtKind::Expr(last), .. }) = block.stmts.last() &&
             let ExprKind::Await(future) = &last.kind &&
             !future.span.from_expansion() &&
             !await_in_expr(future)
         {
+            if matches!(capture, CaptureBy::Value) && captured_by_ref(last) {
```
cynecx, line 51: This fix is rather incomplete because `CaptureBy::Ref` can also move a value into the closure:

```rust
use std::future::Future;

async fn work(_: &str) {}

fn spawn(_: impl Future<Output = ()> + 'static) {}

fn main() {
    let val = "Hello World".to_owned();
    spawn(async { work(&{ val }).await }); // false-positive
}
```

samueltardieu: Indeed, fixed.

samueltardieu:
I also propose to put the lint in nursery for the time being, notably because of https://github.com/rust-lang/rust-clippy/issues/10509.

xFrednet:
Hey, thank you for the PR and comments left by others. I'll sadly be busy until the end of the week, but you'll get a review by the end of next week :)

clippy_lints/src/redundant_async_block.rs
```diff
@@ -48,6 +48,11 @@ impl EarlyLintPass for RedundantAsyncBlock {
             !future.span.from_expansion() &&
             !await_in_expr(future)
         {
+            if captured_by_ref(last) {
```
xFrednet, line 51: I guess that this should also check the async capture mode:

```suggestion
            if CaptureBy::Ref = cap_kind && captured_by_ref(last) {
```

The `cap_kind` can be taken from `ExprKind::Async(cap_kind, _, block)` in line 45

---

When looking at the tests, it looks like this also detects cases with `async move`. Is this intentional? In that case, my suggestion would be wrong, but the function name would also be misleading. `captured_by_move` would then be better or if the `CaptureBy` doesn't matter at all, just something like `captures_value` :)

clippy_lints/src/redundant_async_block.rs
```diff
@@ -82,3 +87,33 @@ impl<'ast> AstVisitor<'ast> for AwaitDetector {
...
 }
+
+/// Check whether an expression may have captured a local variable
+/// by reference. This is done by looking for paths with only one
+/// segment, except as a prefix of `.await` since this would
+/// be captured by value.
+///
+/// There will be false positives, for example when calling a local
+/// function, but at the AST level this cannot be distinguished from
+/// calling a captured closure from the local environment.
```
xFrednet, line 98: Could you give an example when this is a false positive? From the explanation, it sounds like detecting `function` as a ref path, would cause false negatives and not the other way around :)

tests/ui/redundant_async_block.rs
```diff
@@ -62,3 +64,48 @@ fn main() {
...
+}
+
+fn spawn<F: Future + 'static>(_: F) {}
+
+async fn work(_: &str) {}
+
+fn capture() {
+    let val = "Hello World".to_owned();
+    // Do not lint: `val` would not live long enough
+    spawn(async { work(&{ val }).await }); // false-positive
```
xFrednet, line 110:
```suggestion
    spawn(async { work(&{ val }).await }); // false-negative?
```

*false-positive* (FP) refers to triggers that shouldn't happen, while *false-negatives* (FN) are cases that should trigger a lint, but don't. Generally speaking, FNs are better than FPs :)

xFrednet (commented):
Generally looks good to me, I have a few questions, regarding names and documentation, but I'd say this is ready, once they've been cleared up.

Thank you for your patience, while waiting for the review :)

clippy_lints/src/redundant_async_block.rs
xFrednet: Woops, this suggestion has already been made above. Turns out I should have read everything before. In that case I would suggest calling the function `captures_value`

samueltardieu: Renamed.

clippy_lints/src/redundant_async_block.rs
samueltardieu: The false positives comment concerns the function itself, I'll be clearer in the comment. Since this function is used to prevent the lint from triggering, this causes false negatives at the lint level.

I will avoid using "false …" at the function level.

tests/ui/redundant_async_block.rs
samueltardieu: I copied-pasted the code from the issue. It was indeed a false positive, that this patch is fixing. I've removed the obsolete comment.

xFrednet (approved):

xFrednet:
Looks good to me, thank you!

@bors r+

I've added the `beta-nominated` label. It looks like the lint has a few FP and changing the lint group should be able to land before the next release :)

flip1995:
This will make it into the same release as the commit that added the lint. No action required.

Issue #10366, opened 2023-02-18
cast_possible_truncation gives invalid suggestion to use `_::try_from(x)`


ryanavella:
### Summary

`cast_possible_truncation` gives a strange suggestion to replace `x as _` with `_::try_from(x)`, where `x` is `f64` and the placeholder type is `i32`. I don't believe we can use placeholder types in this position currently, and even if this syntax was allowed, there is no `impl TryFrom<f64> for i32`.

### Lint Name

cast_possible_truncation

### Reproducer

I tried this code:

```rust
#![warn(clippy::cast_possible_truncation)]
pub fn f(x: f64) -> i32 {
    x as _
}
```

I saw this happen:

```
warning: casting `f64` to `i32` may truncate the value
 --> src/lib.rs:3:5
  |
3 |     x as _
  |     ^^^^^^
  |
  = help: if this is intentional allow the lint with `#[allow(clippy::cast_precision_loss)]` ...
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_truncation
note: the lint level is defined here
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::cast_possible_truncation)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: ... or use `try_from` and handle the error accordingly
  |
3 |     _::try_from(x)
  |
```

I expected to see this happen: no invalid suggestion, although I suppose the warning itself may still be warranted? 

### Version

```text
rustc 1.69.0-nightly (9a7cc6c32 2023-02-16)
binary: rustc
commit-hash: 9a7cc6c32f1a690f86827e4724bcda85e506ef35
commit-date: 2023-02-16
host: x86_64-pc-windows-msvc
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

 * `I-suggestion-causes-error`

ryanavella:
The suggestion could be slightly improved if it was `<_>::try_from(x)` instead of `_::try_from(x)`.

Here is the diagnostic produced by using the current suggestion:

```
error: expected one of `.`, `;`, `?`, `}`, or an operator, found `::`
 --> src/lib.rs:3:6
  |
3 |     _::try_from(x).unwrap()
  |      ^^ expected one of `.`, `;`, `?`, `}`, or an operator
```

Here is the diagnostic if I instead do `<_>::try_from(x)`

```
[...]

PR #10496, opened 2023-03-13
fix [`cast_possible_truncation`] offering wrong suggestion for casting float to integer

J-ZhengLi:
fixes: #10366 

---

changelog: [`cast_possible_truncation`] Fix incorrect suggestions when casting from float types or to `_`

### DIFF
--- a/clippy_lints/src/casts/cast_possible_truncation.rs
+++ b/clippy_lints/src/casts/cast_possible_truncation.rs
@@ -2,8 +2,9 @@ use clippy_utils::consts::{constant, Constant};
 use clippy_utils::diagnostics::{span_lint, span_lint_and_then};
 use clippy_utils::expr_or_init;
 use clippy_utils::source::snippet;
+use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::{get_discriminant_value, is_isize_or_usize};
-use rustc_errors::{Applicability, SuggestionStyle};
+use rustc_errors::{Applicability, Diagnostic, SuggestionStyle};
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::{BinOpKind, Expr, ExprKind};
 use rustc_lint::LateContext;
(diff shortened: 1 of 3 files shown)


clippy_lints/src/casts/cast_possible_truncation.rs
```diff
@@ -165,10 +173,14 @@ pub(super) fn check(
 
     let name_of_cast_from = snippet(cx, cast_expr.span, "..");
     let cast_to_snip = snippet(cx, cast_to_span, "..");
-    let suggestion = format!("{cast_to_snip}::try_from({name_of_cast_from})");
+    let suggestion = if cast_to_snip == "_" {
+        format!("<{cast_to_snip}>::try_from({name_of_cast_from})")
```
Alexendoo, line 177: We could do something like

```rust
format!("{}.try_into()", Sugg::hir(cx, cast_expr, "..").maybe_par())
```

Would be a bit nicer than `<_>::try_from(x)` in my opinion

J-ZhengLi: 🤔 well, you mean use `try_into()` for that specify scenerio where `cast_to_snip` is a wildcard, or replacing all `try_from` with `try_into` in general? because latter will require the type being known...

Alexendoo: Just the wildcard case

J-ZhengLi: Right, that make sense

xFrednet:
`@rustbot` seems to have been asleep with assigning someone this PR.

r? @Alexendoo Would you mind, since you already gave some feedback? :upside_down_face:

clippy_lints/src/casts/cast_possible_truncation.rs
```diff
@@ -163,12 +172,15 @@ pub(super) fn check(
...
+    } else {
+        format!("{cast_to_snip}::try_from({})", Sugg::hir(cx, cast_expr, ".."))
+    };
 
     span_lint_and_then(cx, CAST_POSSIBLE_TRUNCATION, expr.span, &msg, |diag| {
-        diag.help("if this is intentional allow the lint with `#[allow(clippy::cast_possible_truncation)]` ...");
+        diag.help(help);
         diag.span_suggestion_with_style(
             expr.span,
             "... or use `try_from` and handle the error accordingly",
```
Alexendoo, line 186: Rather than returning early let's add an `if cast_to.is_floating_point() { ... }` around the suggestion creation here, that way we also catch the `f64` -> `f32` case

J-ZhengLi: @Alexendoo  thanks for reviewing btw~

 I somehow thought float to float conversion impls the TryFrom trait but I was wrong :man_facepalming:  also I assume you meant `if !cast_from.is_floating_point()` right?

clippy_lints/src/casts/cast_possible_truncation.rs
```diff
@@ -163,19 +165,34 @@ pub(super) fn check(
...
-        diag.help("if this is intentional allow the lint with `#[allow(clippy::cast_possible_truncation)]` ...");
-        diag.span_suggestion_with_style(
-            expr.span,
-            "... or use `try_from` and handle the error accordingly",
-            suggestion,
-            Applicability::Unspecified,
-            // always show the suggestion in a separate line
-            SuggestionStyle::ShowAlways,
-        );
+        diag.help(help);
```
Alexendoo, line 169: A nit but `help` can be inlined again since it's only used here

With that + a squash it'll be good to merge thanks!

Alexendoo (approved):

Alexendoo:
Thanks @J-ZhengLi!

@bors r+

PR #10497, opened 2023-03-13
new lint: suspicious_doc_comments

### LINKED ISSUE #10485: Warn on accidental triple-slash doc comment `///!`
### What it does

Checks for use of `///!` and warns against it. For example:

```rs
///! This crate does cool things.

pub fn example() {}
```

It’s not immediately obvious, but the doc comment is attached to the wrong item.

### Lint Name

suspicious_doc_comment

### Category

suspicious

### Advantage

Users are less likely to make this mistake and be confused by it / have worse documentation because of it.

### Drawbacks

Nothing I can think of.

### Example

```rust
///! This crate does cool things.

pub fn example() {}
```

Could be written as:

```rust
//! This crate does cool things.

pub fn example() {}
```

y21:
Fixes #10485.

This PR adds a new lint (`suspicious_doc_comments`) that triggers when the user writes `///!` or `/**!`.
This is almost certainly a mistake and the user probably meant to write an inner doc comment (`//!`, `/*!`) to document the module or crate that this comment is contained in.

changelog: [`suspicious_doc_comments`]: new lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4925,6 +4925,7 @@ Released 2018-09-13
 [`suspicious_arithmetic_impl`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_arithmetic_impl
 [`suspicious_assignment_formatting`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_assignment_formatting
 [`suspicious_command_arg_space`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_command_arg_space
+[`suspicious_doc_comments`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_doc_comments
 [`suspicious_else_formatting`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_else_formatting
 [`suspicious_map`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_map
 [`suspicious_op_assign_impl`]: https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_op_assign_impl
--- a/clippy_lints/src/declared_lints.rs
+++ b/clippy_lints/src/declared_lints.rs
@@ -569,6 +569,7 @@ pub(crate) static LINTS: &[&crate::LintInfo] = &[
     crate::strings::STR_TO_STRING_INFO,
     crate::strings::TRIM_SPLIT_WHITESPACE_INFO,
     crate::strlen_on_c_strings::STRLEN_ON_C_STRINGS_INFO,
+    crate::suspicious_doc_comments::SUSPICIOUS_DOC_COMMENTS_INFO,
     crate::suspicious_operation_groupings::SUSPICIOUS_OPERATION_GROUPINGS_INFO,
     crate::suspicious_trait_impl::SUSPICIOUS_ARITHMETIC_IMPL_INFO,
     crate::suspicious_trait_impl::SUSPICIOUS_OP_ASSIGN_IMPL_INFO,
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -284,6 +284,7 @@ mod slow_vector_initialization;
 mod std_instead_of_core;
 mod strings;
 mod strlen_on_c_strings;
+mod suspicious_doc_comments;
 mod suspicious_operation_groupings;
 mod suspicious_trait_impl;
 mod suspicious_xor_used_as_pow;
@@ -958,6 +959,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|_| Box::new(lines_filter_map_ok::LinesFilterMapOk));
     store.register_late_pass(|_| Box::new(tests_outside_test_module::TestsOutsideTestModule));
     store.register_late_pass(|_| Box::new(manual_slice_size_calculation::ManualSliceSizeCalculation));
+    store.register_early_pass(|| Box::new(suspicious_doc_comments::SuspiciousDocComments));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/clippy_lints/src/suspicious_doc_comments.rs
+++ b/clippy_lints/src/suspicious_doc_comments.rs
@@ -1,0 +1,94 @@
+use clippy_utils::diagnostics::{multispan_sugg_with_applicability, span_lint_and_then};
+use if_chain::if_chain;
+use rustc_ast::{token::CommentKind, AttrKind, AttrStyle, Attribute, Item};
+use rustc_errors::Applicability;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects the use of outer doc comments (`///`, `/**`) followed by a bang (`!`): `///!`
+    ///
+    /// ### Why is this bad?
+    /// Triple-slash comments (known as "outer doc comments") apply to items that follow it.
+    /// An outer doc comment followed by a bang (i.e. `///!`) has no specific meaning.
+    ///
+    /// The user most likely meant to write an inner doc comment (`//!`, `/*!`), which
+    /// applies to the parent item (i.e. the item that the comment is contained in,
+    /// usually a module or crate).
+    ///
+    /// ### Known problems
+    /// Inner doc comments can only appear before items, so there are certain cases where the suggestion
+    /// made by this lint is not valid code. For example:
+    /// ```rs
+    /// fn foo() {}
+    /// ///!
+    /// fn bar() {}
+    /// ```
+    /// This lint detects the doc comment and suggests changing it to `//!`, but an inner doc comment
+    /// is not valid at that position.
+    ///
+    /// ### Example
+    /// In this example, the doc comment is attached to the *function*, rather than the *module*.
+    /// ```rust
+    /// pub mod util {
+    ///     ///! This module contains utility functions.
+    ///
+    ///     pub fn dummy() {}
+    /// }
+    /// ```
+    ///
+    /// Use instead:
+    /// ```rust
+    /// pub mod util {
+    ///     //! This module contains utility functions.
+    ///
+    ///     pub fn dummy() {}
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub SUSPICIOUS_DOC_COMMENTS,
+    suspicious,
+    "suspicious usage of (outer) doc comments"
+}
+declare_lint_pass!(SuspiciousDocComments => [SUSPICIOUS_DOC_COMMENTS]);
+
+const WARNING: &str = "this is an outer doc comment and does not apply to the parent module or crate";
+const HELP: &str = "use an inner doc comment to document the parent module or crate";
+
+impl EarlyLintPass for SuspiciousDocComments {
+    fn check_item(&mut self, cx: &EarlyContext<'_>, item: &Item) {
+        let replacements = collect_doc_comment_replacements(&item.attrs);
+
+        if let Some(((lo_span, _), (hi_span, _))) = replacements.first().zip(replacements.last()) {
+            let span = lo_span.to(*hi_span);
+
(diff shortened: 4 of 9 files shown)


clippy_lints/src/outer_doc_comment_bang.rs
```diff
@@ -0,0 +1,99 @@
...
+                        CommentKind::Block => write!(sugg, "/*!{com}*/")
+                    };
+                }
+            }
+        }
+
+        if let Some((lo_span, hi_span)) = lo_span.zip(hi_span) {
+            let span = lo_span.to(hi_span);
+            let indent = indent_of(cx, span);
+            let sugg = reindent_multiline(sugg.into(), true, indent);
```
dswij, line 86: This suggestion takes all lines that are bad, but does not include the one that are legit.

e.g.

```rust
pub mod foo6 {
    ///! sugg1
    /// This is gone in sugg
    ///! sugg2
    
    pub fn bar() {}
}
```

produces

```stderr
error: this is an outer doc comment and does not apply to the parent module or crate
  --> $DIR/outer_doc_comment_bang.rs:36:5
   |
LL | /     ///! sugg1
LL | |     /// This is gone in sugg
LL | |     ///! sugg2
   | |______________^
   |
help: use an inner doc comment to document the parent module or crate
   |
LL ~     //! sugg1
LL +     //! sugg2
   |
```

dswij: I think we can use `multispan` suggestion instead, or rather, just lint the lines individually if that's easier.

clippy_lints/src/outer_doc_comment_bang.rs
```diff
@@ -0,0 +1,99 @@
...
+        let mut hi_span = None;
+        let mut sugg = String::new();
+
+        // multiline doc comments span across multiple attributes
+        // so keep pushing into a string
+        for attr in &item.attrs {
+            if_chain! {
+                if let AttrKind::DocComment(com_kind, sym) = attr.kind;
+                if let AttrStyle::Outer = attr.style;
+                if let Some(com) = sym.as_str().strip_prefix('!');
```
dswij, line 64: maybe move it to a separate function, and then we can use `map(...).collect(...)` to built `Vec<String>`

tests/ui/outer_doc_comment_bang.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub mod foo5 {
+    ///! This module
+    ///! contains
+    ///! useful functions.
+
+    pub fn bar() {}
+}
+
+fn main() {}
```
dswij, line 43: Can we add a case where the outer doc comment bang is not in the first line, and it's not contiguous?

i.e.

```
///!
///
///!
```
and

```
///
///!
///
```

dswij (changes requested):
Thanks for the PR! and welcome to Clippy.

Some minor comments and suggestions.

Also, not sure if I personally love the name `outer_doc_comment_bang`, but combined with the `allow` attribute, it looks neat. I don't know if there is a better name for it though.

clippy_lints/src/outer_doc_comment_bang.rs
y21: > just lint the lines individually if that's easier.

This does seem a lot easier (and that's what I initially had). Gets rid of hacking the lo/hi spans together like this, but IME when someone writes crate-level or module-level documentation, it tends to be a long doc comment with lots of lines so it would report a lot of the same diagnostics potentially. I thought it'd be nicer if it could combine them all into one diagnostic since they are all logically part of the same doc comment.
I'm not sure if this really a problem and there seem to be quite a few corner cases I missed by doing it like this, so just linting each line individually as they are actually represented in the AST might be better.

y21 (commented):
> Also, not sure if I personally love the name outer_doc_comment_bang, but combined with the allow attribute, it looks neat. I don't know if there is a better name for it though.

I agree, I had a hard time coming up with a name for this lint. It seems like the linked issue that initially suggested this lint actually has a name suggestion (`suspicious_doc_comment`), totally missed that. I like that one. I'd be fine with renaming this to `suspicious_doc_comment`.

y21:
Oops, I didn't mean to start  a new review there 😅. The comment I left was in response to changing it from "merging" the spans together into a single suggestion to linting the lines individually.

clippy_lints/src/outer_doc_comment_bang.rs
dswij: > I thought it'd be nicer if it could combine them all into one diagnostic since they are all logically part of the same doc comment

We can use [multispan_sugg](https://github.com/rust-lang/rust-clippy/blob/89160b6de2a62499a5c4a484024cd62061e3d2de/clippy_utils/src/diagnostics.rs#L222-L236) to do this. So it'll be one suggestion but with different spans

tests/ui/outer_doc_comment_bang.rs
y21: Interesting. This is actually a compile error after applying suggestions [[1]](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ea61b61c49e1a3fa86cd865472dc7017) [[2]](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b302c61a683fd3030c8fea8ee7c0ff9a), so presumably the lint should only suggest changing `///!` to `//!` if it isn't preceded by an inline doc comment (i.e. `///` or `/*!`)?

dswij: Huh, TIL. 

If it's a compile error then I think we don't need to cover it. Maybe it's good to put a comment somewhere that indicates we know about it and deliberately chose not to cover it.

y21: Added the comment under "Known problems"

y21:
I changed it to use `mutlispan_sugg` instead. Would `suspicious_doc_comment` be a better name for this lint or is this one fine?

dswij (commented):
Lint looks good! Just a couple small question/comment, and it'd be good to merge aftewards.

I think `suspicious_doc_comments` is nice because we have other similarly-named lints. Also note that the name would be plural because of [Clippy's lint naming convention](https://rust-lang.github.io/rfcs/0344-conventions-galore.html#lints)

clippy_lints/src/outer_doc_comment_bang.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Triple-slash comments (known as "outer doc comments") apply to items that follow it.
+    /// An outer doc comment followed by a bang (i.e. `///!`) has no specific meaning.
+    ///
+    /// The user most likely meant to write an inner doc comment (`//!`, `/*!`), which
+    /// applies to the parent item (i.e. the item that the comment is contained in,
+    /// usually a module or crate).
+    ///
+    /// ### Known problems
+    /// The suggestion made by this lint may be invalid if it sees a normal outer doc comment (`///`)
+    /// followed by `///!`. This lint suggests changing it to `//!`, but that in turn makes the outer doc comment invalid.
```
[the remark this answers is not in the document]
y21: What I meant by "the suggestion may be invalid" is that the suggestion might not be valid code (i.e. does not compile). For example, writing
```rs
///
///!
fn foo() {}
```
compiles as is but triggers this lint and the lint suggests changing the last line to `//!`. Following the lint and changing it to 
```rs
///
//!
fn foo() {}
```
results in a compile error (so the suggestion is invalid).

clippy_lints/src/outer_doc_comment_bang.rs
```diff
@@ -0,0 +1,86 @@
...
+    "outer doc comment followed by a bang symbol"
+}
+declare_lint_pass!(OuterDocCommentBang => [OUTER_DOC_COMMENT_BANG]);
+
+const WARNING: &str = "this is an outer doc comment and does not apply to the parent module or crate";
+const HELP: &str = "use an inner doc comment to document the parent module or crate";
+
+impl EarlyLintPass for OuterDocCommentBang {
+    fn check_item(&mut self, cx: &EarlyContext<'_>, item: &Item) {
+        let spans = collect_bad_doc_comments(&item.attrs);
```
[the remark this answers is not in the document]
y21: Good point... Maybe `sugg` would be better? 🤔
At least that's how `multispan_sugg` names its parameter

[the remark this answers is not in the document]
y21: Oh, I think I misunderstood. Did you mean that there might be a better name for the function? 
In any case, I think we could change both the variable name `spans` and the function name. `let suggs = collect_doc_comment_suggs(&item.attrs);` maybe?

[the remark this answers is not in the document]
Alexendoo: `replacements` is used in a couple other places

clippy_lints/src/outer_doc_comment_bang.rs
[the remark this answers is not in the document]
dswij: Ah, ok. Sorry, I mixed up stuff and misunderstood before.

Then I guess not linting this case is a good idea (while still having the known problem section). A more complete solution would be changing the applicability to `MaybeIncorrect` for these scenarios and then using [`multispan_sugg_with_applicability`](https://github.com/rust-lang/rust-clippy/blob/9e53b6544f77de0f4bfb64413014c9ebd832d133/clippy_utils/src/diagnostics.rs#L227)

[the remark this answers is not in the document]
y21: Sorry, do you mean that I should just change `multispan_sugg` to `multispan_sugg_with_applicability` and use `Applicability::MaybeIncorrect`, or check if the `///!` is preceded by `///` and not lint in those cases?

Alexendoo:
This error [seems to be common](https://github.com/search?q=lang%3Ars+%2F%5C%2F%5C%2F%5C%2F%21%2F&type=code) which makes this a very useful lint

From the search, there's a few styles that would be good to add as tests

Both of these would be joined doc comments but in the first the intention would be to have two separate comments still

https://github.com/Twinklebear/tray_rust/blob/9359cb1868a723431214c9fa1fc8c1aef54adae3/src/sampler/morton.rs#L1-L4
https://github.com/openebs/mayastor/blob/aebfd876be16fa2676f311c36276efd6ad3aa019/sysfs/src/lib.rs#L1-L2

Various gaps

https://github.com/massalabs/massa/blob/b27e8416efac19e8ec470e96bb49c74687092ac3/massa-async-pool/src/changes.rs#L14-L16
https://github.com/Keats/jsonwebtoken/blob/8fba79b25459eacc33a80e1ee37ff8eba64079ca/src/jwk.rs#L2-L6

`////!` is probably a commented out outer doc comment

https://github.com/tanakh/reqwest/blob/147831375613a5e508487b2d85a99104ae1505af/src/lib.rs#L184-L185

It would also be good to see if it works on items that don't generally have doc comments applied to them, like

```rust
///! a
use x;

///! b
macro_rules! x {() => {}}
```

You don't have to explicitly handle all of these, particularly the first two are kind of conflicting, but it'd be good to have tests for them anyway

clippy_lints/src/outer_doc_comment_bang.rs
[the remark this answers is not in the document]
Alexendoo: It may need to be `MaybeIncorrect` normally, it seems pretty common to accidentally put these comments [after an item](https://github.com/lsd-rs/lsd/blob/118faab5339cdd02ef31a47057173a7cbe91534b/src/config_file.rs#L9-L10) which would be a compile error when fixed

y21:
> https://github.com/Keats/jsonwebtoken/blob/8fba79b25459eacc33a80e1ee37ff8eba64079ca/src/jwk.rs#L2-L6

This is one of the "known-but-deliberately-ignored" problems, I think, so I don't think I can add a test for this, can I? `// run-rustfix` would make this fail the test

Alexendoo:
You can put them in a separate `suspicious_doc_comments_unfixable.rs` without `// run-rustfix`

y21:
Re the label: is there something left to do? I think I have a test for all the cases that were mentioned

dswij:
@y21 oh, there's a conflict, so I thought you are still working on it :)

y21:
Fixed the conflicts and cleaned up the commit history a bit.

dswij (approved):

dswij:
@y21 Huge thanks for this! Should be a really nice, useful lint

@bors r+

PR #10499, opened 2023-03-13
Fix `almost_swapped` false positive (`let mut a = b; a = a`)

### LINKED ISSUE #10421: `almost_swapped` false positive `a=a; a=a` (in `clap::Parser` derive)
### Summary

#10177

- [ ] `clippy::almost_swapped`'s error message is wrong when triggered in a derive macro. Not sure about other macros.

> error: this looks like you are trying to swap `__clap_subcommand` and `Parser`

There is no variable called `Parser`, that is the name of the derive macro. Should be:
> error: this looks like you are trying to swap `__clap_subcommand` and `__clap_subcommand`

- [x] `clippy::almost_swapped` triggers when you would be "swapping" the same variable with itself. I.e. the following triggers it:
```rust
let a = a;
let a = a;
```
First impression wouldn't be that we're trying to swap `a` with `a` since they are the same name. And the generated suggestion `std::mem::swap(&mut a, &mut a)` is silly; breaks aliasing rules.

- [ ] `clippy::almost_swapped` triggers within macros in the first place. Should it?

Obviously that code above is a little silly, but can be generated in macros, e.g. in `clap::Parser`'s derive. Not sure if clippy is supposed to trigger within macros in general--I.e. #10318 ?

### Lint Name

almost_swapped

### Reproducer

I tried this code:

```rust
fn main() {
    let a = 0;
    let a = a;
    let a = a;
}
```

I saw this happen:

```
error: this looks like you are trying to swap `a` and `a`
 --> src/main.rs:3:5
  |
3 | /     let a = a;
4 | |     let a = a;
  | |_____________^ help: try: `std::mem::swap(&mut a, &mut a)`
  |
  = note: or maybe you should use `std::mem::replace`?
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#almost_swapped
  = note: `#[deny(clippy::almost_swapped)]` on by default
```

No error

---
### Clap macro

I tried this code:

```rust
use clap::Parser;

#[derive(Parser)]
enum MyEnum {
    MyVariant
}
```

Which expands to:
```rust
// ...
#[allow(dead_code, unreachable_code, unused_variables, unused_braces)]
#[allow(clippy :: style, clippy :: complexity, clippy :: pedantic, clippy ::
restriction, clippy :: perf, clippy :: deprecated, clippy :: nursery, clippy
:: cargo, clippy :: suspicious_else_formatting,)]
#[deny(clippy :: correctness)]
impl clap::Subcommand for MyEnum {
    fn augment_subcommands<'b>(__clap_app: clap::Command) -> clap::Command {
        ;
        let __clap_app = __clap_app;
        let __clap_app =
            __clap_app.subcommand({
                    let __clap_subcommand = clap::Command::new("my-variant");
                    let __clap_subcommand = __clap_subcommand;
                    let __clap_subcommand = __clap_subcommand;
                    __clap_subcommand
                });
        ;
        __clap_app
    }
    fn augment_subcommands_for_update<'b>(__clap_app: clap::Command)
        -> clap::Command {
        ;
        let __clap_app = __clap_app;
        let __clap_app =
            __clap_app.subcommand({
                    let __clap_subcommand = clap::Command::new("my-variant");
                    let __clap_subcommand = __clap_subcommand;
                    let __clap_subcommand = __clap_subcommand;
                    __clap_subcommand
                });
        ;
        __clap_app
    }
    fn has_subcommand(__clap_name: &str) -> bool {
        if "my-variant" == __clap_name { return true }
        false
    }
}
```

I saw this happen:

```
error: this looks like you are trying to swap `__clap_subcommand` and `Parser`
 --> src/lib.rs:3:10
  |
3 | #[derive(Parser)]
  |          ^^^^^^
  |
  = note: or maybe you should use `std::mem::replace`?
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#almost_swapped
 --> src/lib.rs:3:10
  |
3 | #[derive(Parser)]
  |          ^^^^^^
  = note: `#[deny(clippy::almost_swapped)]` implied by `#[deny(clippy::correctness)]`
  = note: this error originates in the derive macro `Parser` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `playground` due to previous error
```

No error

Or

```rust
error: this looks like you are trying to swap `__clap_subcommand` and `__clap_subcommand`
```

### Version

```text
Playground:
Build using the Nightly version: 1.69.0-nightly

(2023-02-27 7281249a19a9755e9d88)

Local windows:
$ rustc +nightly -Vv
rustc 1.69.0-nightly (d962ea578 2023-02-26)
binary: rustc
commit-hash: d962ea57899d64dc8a57040142c6b498a57c8064
commit-date: 2023-02-26
host: x86_64-pc-windows-msvc
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

blyxyas:
Fixes `2` in #10421 
changelog: [`almost_swapped`]: Fix false positive when a variable is changed to itself. (`a = a`)

### DIFF
--- a/clippy_lints/src/swap.rs
+++ b/clippy_lints/src/swap.rs
@@ -190,6 +190,7 @@ fn check_suspicious_swap(cx: &LateContext<'_>, block: &Block<'_>) {
             && first.span.eq_ctxt(second.span)
             && is_same(cx, lhs0, rhs1)
             && is_same(cx, lhs1, rhs0)
+			&& !is_same(cx, lhs1, rhs1) // Ignore a = b; a = a (#10421)
             && let Some(lhs_sugg) = match &lhs0 {
                 ExprOrIdent::Expr(expr) => Sugg::hir_opt(cx, expr),
                 ExprOrIdent::Ident(ident) => Some(Sugg::NonParen(ident.as_str().into())),
--- a/tests/ui/swap.fixed
+++ b/tests/ui/swap.fixed
@@ -186,3 +186,11 @@ const fn issue_9864(mut u: u32) -> u32 {
     v = temp;
     u + v
 }
+
+#[allow(clippy::let_and_return)]
+const fn issue_10421(x: u32) -> u32 {
+    let a = x;
+    let a = a;
+    let a = a;
+    a
+}
--- a/tests/ui/swap.rs
+++ b/tests/ui/swap.rs
@@ -215,3 +215,11 @@ const fn issue_9864(mut u: u32) -> u32 {
     v = temp;
     u + v
 }
+
+#[allow(clippy::let_and_return)]
+const fn issue_10421(x: u32) -> u32 {
+    let a = x;
+    let a = a;
+    let a = a;
+    a
+}


giraffate (approved):
Thanks!

It seems to be only resolve `2` in https://github.com/rust-lang/rust-clippy/issues/10421#issue-1602340635. So I updated notes not to close https://github.com/rust-lang/rust-clippy/issues/10421.

blyxyas:
Yeah I only tackled *2* because I didn't know if it should lint on macros or not.
Anyway, looking again at the problem, *3* should also fix *1*, right? The fix to the third point (ignoring macros) should remove the necessity to correct an error message that happens in macros.

Is this true?

Alexendoo:
Yeah it would, and yeah we don't want to be linting macros

PR #10503, opened 2023-03-14
new lint: needless traits in scope

woshilapin:
changelog: [`needless_traits_in_scope`]: trait is needlessly imported in trait's namespace

This is my first contribution to `rust-clippy`.

I propose a lint to find traits that are imported with a `use` but trait's name is not used afterwards (only methods). Importing traits with ` as _` gives 2 advantages:
- not keeping an unused name in the trait namespace
- helps identifying traits (compared to `struct`) in the `use` statements>

I do not consider this PR finished, because I'm looking for first feedbacks... and a little help for the `FIXME` in the UI test case (how to check the trait's name is actually not used in the rest of the scope).

And don't worry about telling me the lint is dumb, the journey trying stuff with `rust-clippy` so far has been enlightening anyway.

### DIFF
--- a/clippy_lints/src/needless_traits_in_scope.rs
+++ b/clippy_lints/src/needless_traits_in_scope.rs
@@ -1,0 +1,76 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use rustc_errors::Applicability;
+use rustc_hir::*;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Find traits that are not explicitely used in scope (like `AsRef::as_ref()`)
+    /// but directly with the trait method (like `opt.as_ref()`).
+    ///
+    /// These traits can be imported anonymously with `use crate::Trait as _`.
+    /// This avoids name collision with other traits (possibly with the same name).
+    /// It also helps identify the traits in `use` statements.
+    ///
+    /// ### Why is this bad?
+    /// This needlessly brings a trait in trait's namespace.
+    ///
+    /// ### Example
+    /// ```rust
+    /// use std::io::Read;
+    /// fn main() {
+    ///   let mut b = "I'm your father!".as_bytes();
+    ///   let mut buffer = [0; 10];
+    ///   b.read(&mut buffer)
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// use std::io::Read as _;
+    /// fn main() {
+    ///   let mut b = "I'm your father!".as_bytes();
+    ///   let mut buffer = [0; 10];
+    ///   b.read(&mut buffer)
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub NEEDLESS_TRAITS_IN_SCOPE,
+    pedantic,
+    "trait is needlessly imported in trait's namespace, and can be anonymously imported"
+}
+declare_lint_pass!(NeedlessTraitsInScope => [NEEDLESS_TRAITS_IN_SCOPE]);
+
+impl<'tcx> LateLintPass<'tcx> for NeedlessTraitsInScope {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
+        // Only process `use` statements, ignore `UseKind::Glob`
+        let ItemKind::Use(use_path, UseKind::Single) = item.kind else {
+            return
+        };
+        // Check if the `use` is aliased with ` as `.
+        // If aliased, then do not process, it's probably for a good reason
+        if item.ident != use_path.segments.last().unwrap().ident {
+            return;
+        }
+        let path = use_path
+            .segments
+            .iter()
+            .map(|segment| segment.ident)
+            .fold(String::new(), |mut acc, ident| {
+                if !acc.is_empty() {
+                    acc += "::";
+                }
+                acc += ident.as_str();
+                acc
+            });
+        span_lint_and_sugg(
+            cx,
+            NEEDLESS_TRAITS_IN_SCOPE,
+            use_path.span,
+            "trait is needlessly imported in trait's namespace",
+            "you can import the trait anonymously",
+            format!("{path} as _"),
+            Applicability::MachineApplicable,
+        );
+    }
+}
--- a/clippy_lints/src/needless_traits_in_scope.rs
+++ b/clippy_lints/src/needless_traits_in_scope.rs
@@ -14,7 +14,9 @@ declare_clippy_lint! {
     /// It also helps identify the traits in `use` statements.
     ///
     /// ### Why is this bad?
-    /// This needlessly brings a trait in trait's namespace.
+    /// This needlessly brings a trait in trait's namespace, where it could
+    /// shadow other things. This is not really a problem, this lint is just
+    /// for those who like to keep things tidy.
     ///
     /// ### Example
     /// ```rust
--- a/clippy_lints/src/needless_traits_in_scope.rs
+++ b/clippy_lints/src/needless_traits_in_scope.rs
@@ -47,6 +47,14 @@ impl<'tcx> LateLintPass<'tcx> for NeedlessTraitsInScope {
         let ItemKind::Use(use_path, UseKind::Single) = item.kind else {
             return
         };
+        // Check if it's a trait
+        if !use_path
+            .res
+            .iter()
+            .any(|res| matches!(res, def::Res::Def(def::DefKind::Trait, _)))
+        {
+            return;
+        }
         // Check if the `use` is aliased with ` as `.
         // If aliased, then do not process, it's probably for a good reason
         if item.ident != use_path.segments.last().unwrap().ident {
--- a/clippy_lints/src/needless_traits_in_scope.rs
+++ b/clippy_lints/src/needless_traits_in_scope.rs
@@ -14,7 +14,7 @@ declare_clippy_lint! {
     /// It also helps identify the traits in `use` statements.
     ///
     /// ### Why is this bad?
-    /// This needlessly brings a trait in trait's namespace, where it could
+    /// This needlessly brings a trait into the type namespace, where it could
     /// shadow other things. This is not really a problem, this lint is just
     /// for those who like to keep things tidy.
     ///
@@ -39,7 +39,7 @@ declare_clippy_lint! {
     #[clippy::version = "1.69.0"]
     pub NEEDLESS_TRAITS_IN_SCOPE,
     restriction,
-    "trait is needlessly imported in trait's namespace, and can be anonymously imported"
+    "trait is needlessly imported into the type namespace, and can be anonymously imported"
 }
 declare_lint_pass!(NeedlessTraitsInScope => [NEEDLESS_TRAITS_IN_SCOPE]);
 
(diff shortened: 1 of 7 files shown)


clippy_lints/src/needless_traits_in_scope.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// use std::io::Read as _;
+    /// fn main() {
+    ///   let mut b = "I'm your father!".as_bytes();
+    ///   let mut buffer = [0; 10];
+    ///   b.read(&mut buffer)
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub NEEDLESS_TRAITS_IN_SCOPE,
+    pedantic,
```
llogiq, line 39: I think this should be a restriction lint. It's nothing that should be frowned upon in normal Rust code and makes the code longer, but it could bring benefits in certain situations (those you mentioned).

clippy_lints/src/needless_traits_in_scope.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// ### What it does
+    /// Find traits that are not explicitely used in scope (like `AsRef::as_ref()`)
+    /// but directly with the trait method (like `opt.as_ref()`).
+    ///
+    /// These traits can be imported anonymously with `use crate::Trait as _`.
+    /// This avoids name collision with other traits (possibly with the same name).
+    /// It also helps identify the traits in `use` statements.
+    ///
+    /// ### Why is this bad?
+    /// This needlessly brings a trait in trait's namespace.
```
llogiq, line 17:
```suggestion
    /// This needlessly brings a trait in trait's namespace, where it could
    /// shadow other things. This is not really a problem, this lint is just
    /// for those who like to keep things tidy.
```

clippy_lints/src/needless_traits_in_scope.rs
```diff
@@ -0,0 +1,76 @@
...
+    ///   let mut buffer = [0; 10];
+    ///   b.read(&mut buffer)
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub NEEDLESS_TRAITS_IN_SCOPE,
+    pedantic,
+    "trait is needlessly imported in trait's namespace, and can be anonymously imported"
+}
+declare_lint_pass!(NeedlessTraitsInScope => [NEEDLESS_TRAITS_IN_SCOPE]);
```
llogiq, line 42: You'll want to write your own struct so you can `impl_lint_pass!` here instead. Then you can add a `Visitor` to your LintPass that you use on `check_crate` to run through the code and collect any trait use. Please leave a comment or ping me on zulip if you need help with that.

llogiq (commented):
I don't think the lint is stupid. It will be useful for those who like their namespaces tidy. I left a few comments, hope they are helpful. Please comment here or ping us on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy) if you need further assistance.

@rustbot author

clippy_lints/src/needless_traits_in_scope.rs
woshilapin: Indeed, better than `pedantic`. Fixed in 02db9854e.

clippy_lints/src/needless_traits_in_scope.rs
woshilapin: Done in ef7fac20e.

clippy_lints/src/needless_traits_in_scope.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// ### What it does
+    /// Find traits that are not explicitely used in scope (like `AsRef::as_ref()`)
+    /// but directly with the trait method (like `opt.as_ref()`).
+    ///
+    /// These traits can be imported anonymously with `use crate::Trait as _`.
+    /// This avoids name collision with other traits (possibly with the same name).
+    /// It also helps identify the traits in `use` statements.
+    ///
+    /// ### Why is this bad?
+    /// This needlessly brings a trait in trait's namespace, where it could
```
PatchMixolydic, line 17: Trait names do not have their own namespace; rather, [they live in the type namespace](https://doc.rust-lang.org/reference/names/namespaces.html).
```suggestion
    /// This needlessly brings a trait into the type namespace, where it could
```

This could be another reason for someone to use this trait: it'd allow them to avoid name collisions between types and traits.

clippy_lints/src/needless_traits_in_scope.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// fn main() {
+    ///   let mut b = "I'm your father!".as_bytes();
+    ///   let mut buffer = [0; 10];
+    ///   b.read(&mut buffer)
+    /// }
+    /// ```
+    #[clippy::version = "1.69.0"]
+    pub NEEDLESS_TRAITS_IN_SCOPE,
+    restriction,
+    "trait is needlessly imported in trait's namespace, and can be anonymously imported"
```
PatchMixolydic, line 42:
```suggestion
    "trait is needlessly imported into the type namespace, and can be anonymously imported"
```

clippy_lints/src/needless_traits_in_scope.rs
```diff
@@ -0,0 +1,86 @@
...
+                    acc += "::";
+                }
+                acc += ident.as_str();
+                acc
+            });
+        span_lint_and_sugg(
+            cx,
+            NEEDLESS_TRAITS_IN_SCOPE,
+            use_path.span,
+            "trait is needlessly imported in trait's namespace",
```
PatchMixolydic, line 80:
```suggestion
            "trait is needlessly imported into the type namespace",
```

PatchMixolydic (changes requested):

clippy_lints/src/needless_traits_in_scope.rs
woshilapin: Indeed, thank you, I had actually problem to find the information (probably missing the right keywords). I remember there were a few different namespaces, but couldn't remember the exact boundaries of each category.

Fixed in a6e854e62edbabcb33f8f10658c9e098e23d4868.

clippy_lints/src/needless_traits_in_scope.rs
woshilapin: Fixed in a6e854e62edbabcb33f8f10658c9e098e23d4868.

clippy_lints/src/needless_traits_in_scope.rs
woshilapin: Fixed in a6e854e62edbabcb33f8f10658c9e098e23d4868.

llogiq:
If you need further assistance for implementing the visitor, feel free to ping us here or [on zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy).

woshilapin:
> If you need further assistance for implementing the visitor, feel free to ping us here or [on zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy).

Thanks for getting back to me. To be honest, I've had little time since I opened the PR to explore the visitor solution. But the little I looked, I'm sure some help might help me.

Let met start with what I (think I) understood. `check_crate` will run before `check_item` which means I can do a first parsing to store stuff in the `struct NeedlessTraitsInScope` that I will use later in `check_item` to decide if I need to yield the lint or not.

I also understand that instead of 

```
declare_lint_pass!(NeedlessTraitsInScope => [NEEDLESS_TRAITS_IN_SCOPE]);
```
I can write the following
```
struct NeedlessTraitsInScope {
  // some HashMap ?
}
impl_lint_pass!(NeedlessTraitsInScope => [NEEDLESS_TRAITS_IN_SCOPE]);
```

Now, it's fuzzy what I'm missing, but I'll try to make some questions:
- I'm not sure how to parse the entire crate with a visitor?
  - I've seen a `for_each_expr`, but I'm not sure how to detect what I need (should I only focus on `ExprKind::Call` and `ExprKind::MethodCall`)
  - since `check_crate` only give me a `LateContext`, I'm not sure how to use `for_each_expr` which seems to take a `impl Visitable`
- I'm not even sure how to store what I found during `check_crate` parsing
  - because I need to remember that I found a Path used... but I also need to remember where it was used (in which module of which file)

In any case, I'll probably be slow since I don't have a lot of personal time these days to work on this. I hope it's not a problem :relaxed:

llogiq:
No problem. That's a good start; everything you wrote seems correct to me. Now the way using a Visitor works is: in `check_crate(..)` you construct your visitor, call its `visit_crate` method that will walk the whole crate and can collect the data. After calling that, you'll deconstruct the visitor to get the collected data, which you store in your LintPass.

For the visitor itself, you'll want to override the `visit_expr` method to check `Call`s and `MethodCall`s.

woshilapin:
I tried implementing the `Visitor`. With the hands on it, most of the help you provided now make sense (much more that when I read first). But I'm having some problems with the `Visitor`.

I started implementing `rustc_hir::intravisit::Visitor` for `struct UsedTraitVisitor` (a new `struct` for visiting)... but then, I couldn't find the `visit_crate()` for this `Visitor` trait.

Parsing through the documentation, I actually found a `visit_crate` on `rustc_ast::visit::Visitor`, so I tried to implement this instead (not really sure of the difference between the two). With `rustc_ast::visit::Visitor`, I now have a difficulty to call `visit_crate` that needs a `rustc_ast::ast::Crate`. I'm not sure how to access `rustc_ast::ast::Crate` from `LateContext`. I did found `cx.tcx.crate_for_resolver()` that returns a `Steal<(rustc_ast::ast::Crate, _)>` but using `steal.borrow().0` seems to panic at runtime (even looks like an ICE, asking to report to the compiler).

Note: I usually do not like pushing broken code, but tell me if that can help.

llogiq:
Sorry, my bad. I hadn't looked into the intravisit code in a while. `cx.tcx().visit_all_item_likes_in_crate(&mut visitor)` should do the crate walking for you.

woshilapin:
> Sorry, my bad. I hadn't looked into the intravisit code in a while. `cx.tcx().visit_all_item_likes_in_crate(&mut visitor)` should do the crate walking for you.

Hi, thanks for the help. I actually didn't mention it in my previous comment, but I did already explore `visit_all_item_likes_in_crate`: it looked like a dead end when I did. It does explore all `rustc_hir::Item`, but I'm interested in all `rustc_hir::Expr` (the `Visitor::visit_expr` is never called, only `Visitor::visit_item`). I also found out about `clippy_utils::visitors::for_each_expr`, but it does only take a `impl Visitable` which, at the time, didn't seem to be implemented by the things I care (the entire crate or at least, each module :shrug:).

After digging, I'll try to explore the following option:
- `visit_all_item_likes_in_crate` with a visitor implementing `visit_item`
- when visiting `ItemKind::Fn`, use `for_each_expr` on the body (I should be able to get `BodyId` from `ItemKind::Fn`, then get the `Body<'tcx>` from `cx.tcx.hir().body(body_id)`)

That said, this does not look like a general solution. If that does work, I might also need to explore other `ItemKind` (`Macro` and `Impl` for example ? :thinking: )

llogiq:
You can just implement `visit_expr`, the default implementations will take care of the rest

J-ZhengLi:
Hi @woshilapin , just a friendly reminder that this PR is being inactive for a while now (just pass its anniversary 😄). 

So, if you're still interested in working on this, plz let us know. And if you have any further questions, don't hesitate to ask people in [this zulip channel](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy), cheers!

woshilapin:
@J-ZhengLi Thank you for following-up. I don't really plan to work on it anymore, I don't really have the time. And the complexity of the task is a bit more than I had in mind when I started this. Feel free to close it if you prefer.

woshilapin:
For those interested in this work, someone else actually did land a new clippy lint in 1.83 (see https://github.com/rust-lang/rust-clippy/pull/13322).

PR #10516, opened 2023-03-17
Use `split-debuginfo = "unpacked"` for debug builds

Alexendoo:
On Windows this has no effect as it's unsupported. On macOS the default set by cargo is already unpacked so no effect there either

For Linux it shaves a bit off the rebuild time, for me in the case of a simple `touch` + `cargo build` it goes from 12s to 10s

It saves a good amount of disk space too, on `aarch64-unknown-linux-gnu` it saves 1.2GB for a plain `cargo build`, 3GB when also running `cargo dev` and `cargo test --no-run -F internal`

r? @flip1995

changelog: none

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -11,3 +11,6 @@ target-dir = "target"
 
 [unstable]
 binary-dep-depinfo = true
+
+[profile.dev]
+split-debuginfo = "unpacked"
--- a/.github/workflows/clippy_bors.yml
+++ b/.github/workflows/clippy_bors.yml
@@ -180,6 +180,8 @@ jobs:
 
     # Run
     - name: Build Integration Test
+      env:
+        CARGO_PROFILE_DEV_SPLIT_DEBUGINFO: off
       run: cargo test --test integration --features integration --no-run
 
     # Upload


xFrednet:
Out of interest, do you know why `unpacked` is the default on Mac but not on linux? :thinking:

flip1995:
@bors r+

That sounds like a really nice improvement. The Clippy target dir is responsible for 9.5% of the file sizes in my `$HOME` dir. This will improve things :)

Alexendoo:
The extra files in `deps` confused the `mv` command, but I've set the integration build to not use split debuginfo since it's easier to just transfer the binaries

klensy:
Funny thing: on my windows machine with this change rustc(or cargo?) forgets to clean 2.2GB of `*.o` files in deps folder.

klensy:
Looks like it preserved here, but should this work for `x86_64-pc-windows-msvc`? Idk:
https://github.com/rust-lang/rust/blob/0b011b7b7e5d1a1737aa3337f01b79fd5f56cf04/compiler/rustc_codegen_ssa/src/back/link.rs#L1368-L1389

Alexendoo:
hmm I see what's happening I think, we use `-Zunstable-options` to enable internal rustc lints, but that also enables unsupported `split-debuginfo` profiles

I'll see if upstream will accept another -Z flag for that instead of using `-Zunstable-options`

PR #10520, opened 2023-03-18
Use uninit checking from rustc

### LINKED ISSUE #10407: clippy doesn't properly model whether a type is allowed to be uninit
### Summary

The `uninit_vec` and `uninit_assumed_init` lints use a helper to check whether a type may be uninit.

https://github.com/rust-lang/rust-clippy/blob/3d193fa17a7df3256d58b841a2df92af1cbdbea4/clippy_utils/src/ty.rs#L541-L548

This helper is incorrect for newtypes around `MaybeUninit<T>`, ZSTs and unions in general.

This should be fixed by just deleting this helper function and using rustc's `check_validity_of_init` (the name may change in rust-lang/rust#108505). It doesn't currently support uninit checks but that should be really easy to add upstream (and needed for rust-lang/rust#100423 anyways).

### Lint Name

uninit_vec,uninit_assumed_init

### Reproducer

I tried this code:

```rust
use core::mem::MaybeUninit;

#[repr(transparent)]
struct Transparent<T>(MaybeUninit<T>);

pub fn foo<T>() {
    let mut vec = Vec::<Transparent<T>>::new();
    vec.reserve(20);
    unsafe { vec.set_len(20) };
}
```

I saw this happen:

```
error: calling `set_len()` immediately after reserving a buffer creates uninitialized values
 --> src/lib.rs:8:5
  |
8 |     vec.reserve(20);
  |     ^^^^^^^^^^^^^^^^
9 |     unsafe { vec.set_len(20) };
  |              ^^^^^^^^^^^^^^^
  |
  = help: initialize the buffer or wrap the content in `MaybeUninit`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#uninit_vec
  = note: `#[deny(clippy::uninit_vec)]` on by default
```

I expected to see this happen: Everything is fine

### Version

```text
rustc 1.69.0-nightly (34e6673a0 2023-02-25)
binary: rustc
commit-hash: 34e6673a0473e90ef01a18eb575392c9e3859747
commit-date: 2023-02-25
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

Nilstrieb:
rustc has proper heuristics for actually checking whether a type allows being left uninitialized (by asking CTFE). We can now use this for our helper instead of rolling our own bad version with false positives.

I added this in rustc in rust-lang/rust#108669

Fix #10407

changelog: [`uninit_vec`]: fix false positives
changelog: [`uninit_assumed_init`]: fix false positives

### DIFF
--- a/clippy_utils/src/ty.rs
+++ b/clippy_utils/src/ty.rs
@@ -16,9 +16,9 @@ use rustc_infer::infer::{
 use rustc_lint::LateContext;
 use rustc_middle::mir::interpret::{ConstValue, Scalar};
 use rustc_middle::ty::{
-    self, AdtDef, AliasTy, AssocKind, Binder, BoundRegion, FnSig, IntTy, List, ParamEnv, Predicate, PredicateKind,
-    Region, RegionKind, SubstsRef, Ty, TyCtxt, TypeSuperVisitable, TypeVisitable, TypeVisitableExt, TypeVisitor,
-    UintTy, VariantDef, VariantDiscr,
+    self, layout::ValidityRequirement, AdtDef, AliasTy, AssocKind, Binder, BoundRegion, FnSig, IntTy, List, ParamEnv,
+    Predicate, PredicateKind, Region, RegionKind, SubstsRef, Ty, TyCtxt, TypeSuperVisitable, TypeVisitable,
+    TypeVisitableExt, TypeVisitor, UintTy, VariantDef, VariantDiscr,
 };
 use rustc_middle::ty::{GenericArg, GenericArgKind};
 use rustc_span::symbol::Ident;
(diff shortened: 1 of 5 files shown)


tests/ui/uninit.rs
```diff
@@ -3,12 +3,14 @@
...
 
+union MyOwnMaybeUninit {
+    value: u8,
+    uninit: (),
+}
+
 fn main() {
     let _: usize = unsafe { MaybeUninit::uninit().assume_init() };
 
-    // edge case: For now we lint on empty arrays
```
Nilstrieb, line 9: now we do!

clippy_utils/src/ty.rs
```diff
@@ -538,13 +538,10 @@ pub fn same_type_and_consts<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> bool {
...
-    match *ty.kind() {
-        ty::Array(component, _) => is_uninit_value_valid_for_ty(cx, component),
-        ty::Tuple(types) => types.iter().all(|ty| is_uninit_value_valid_for_ty(cx, ty)),
-        ty::Adt(adt, _) => cx.tcx.lang_items().maybe_uninit() == Some(adt.did()),
-        _ => false,
-    }
+pub fn is_uninit_value_valid_for_ty<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
+    cx.tcx
+        .check_validity_requirement((ValidityRequirement::Uninit, cx.param_env.and(ty)))
+        .unwrap_or(true)
```
Alexendoo, line 544: Should we be falling back to `false` here like before? I don't know the circumstances in which the function would return an `Err`

Nilstrieb: It errors when it can't get the layout of the type, for example because it contains generic params. I chose true to avoid false positives, but it probably makes sense to make it false instead, so that clippy complains when using it on generic params (because that will probably be wrong).

Alexendoo: Thanks for explaining, yeah let's go with `false` in that case

tests/ui/uninit.rs
```diff
@@ -3,13 +3,15 @@
...
-    let _: [u8; 0] = unsafe { MaybeUninit::uninit().assume_init() };
-
-    // edge case: For now we accept unit tuples
```
Nilstrieb, line 12: I also changed this comment because this isn't really a "for now", ZSTs are simply valid here.

Alexendoo (approved):

Alexendoo:
Great, thanks!

@bors r+

zesterer:
I believe we've found a false-positive for this [here](https://github.com/zesterer/chumsky/blob/main/src/private.rs#L277-L281).

Nilstrieb:
Uh, that's because this is a polymorphic type and we are conservative around polymorphic types and always say that they are wrong.
That's not very good, open an issue. Maybe we should have/keep the bespoke logic as a fallback for polymorphic types since the rustc checks cannot handle them.

PR #10526, opened 2023-03-21
Really dogfood clippy

samueltardieu:
The dogfood success condition was inverted in `tests/dogfood.rs`:

```rust
  assert!(!failed_packages.is_empty(), …);
```

while instead the `failed_packages` collection must be empty:

```rust
  assert!(failed_packages.is_empty(), …);
```

And indeed, several clippy lint source files were not clean and had to be fixed in the process.

changelog: none


flip1995:
Ah thank you!

@bors r+ p=100

PR #10528, opened 2023-03-22
Clear with drain

### LINKED ISSUE #9339: new lint: `vec.drain(..)` instead of `vec.clear()`

bluthej:
changelog: [`clear_with_drain`]: Add new lint

Fixes #9339

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -78,7 +78,7 @@ use std::sync::OnceLock;
 use std::sync::{Mutex, MutexGuard};
 
 use if_chain::if_chain;
-use rustc_ast::ast::{self, LitKind};
+use rustc_ast::ast::{self, LitKind, RangeLimits};
 use rustc_ast::Attribute;
 use rustc_data_structures::fx::FxHashMap;
 use rustc_data_structures::unhash::UnhashMap;
@@ -115,6 +115,7 @@ use rustc_span::Span;
 use rustc_target::abi::Integer;
 
 use crate::consts::{constant, Constant};
+use crate::higher::Range;
 use crate::ty::{can_partially_move_ty, expr_sig, is_copy, is_recursively_primitive_type, ty_is_fn_once_param};
 use crate::visitors::for_each_expr;
 
@@ -1489,6 +1490,24 @@ pub fn is_else_clause(tcx: TyCtxt<'_>, expr: &Expr<'_>) -> bool {
         )) => else_expr.hir_id == expr.hir_id,
         _ => false,
     }
+}
+
+/// Checks whether the given `Range` is equivalent to a `RangeFull`.
+/// Inclusive ranges are not considered because they already constitute a lint.
+pub fn is_range_full(cx: &LateContext<'_>, container: &Expr<'_>, range: Range<'_>) -> bool {
+    range.start.map_or(true, |e| is_integer_const(cx, e, 0))
+        && range.end.map_or(true, |e| {
+            if range.limits == RangeLimits::HalfOpen
+                && let ExprKind::Path(QPath::Resolved(None, container_path)) = container.kind
+                && let ExprKind::MethodCall(name, self_arg, [], _) = e.kind
+                && name.ident.name == sym::len
+                && let ExprKind::Path(QPath::Resolved(None, path)) = self_arg.kind
+            {
+                container_path.res == path.res
+            } else {
+                false
+            }
+        })
 }
 
 /// Checks whether the given expression is a constant integer of the given value.
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -96,6 +96,7 @@ use rustc_hir::{
 use rustc_lexer::{tokenize, TokenKind};
 use rustc_lint::{LateContext, Level, Lint, LintContext};
 use rustc_middle::hir::place::PlaceBase;
+use rustc_middle::mir::ConstantKind;
 use rustc_middle::ty as rustc_ty;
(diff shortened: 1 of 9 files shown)


tests/ui/clear_with_drain.rs
```diff
@@ -0,0 +1,40 @@
...
+    v.drain(1..); // Yay
+
+    let mut v = vec![1, 2, 3];
+    v.drain(..v.len() - 1); // Yay
+
+    let mut v = vec![1, 2, 3];
+    v.drain(1..v.len() - 1); // Yay
+}
+
+fn main() {}
```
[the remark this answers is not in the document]
bluthej: Would you like me to add some code that uses the `iter` variables in the existing test cases? Or did you mean that I should write some additional tests, like for instance:
```rust
let mut v = vec![1, 2, 3];
for x in v.drain(..) {
    // do something with x
}
```

clippy_utils/src/lib.rs
```diff
@@ -1491,6 +1492,24 @@ pub fn is_else_clause(tcx: TyCtxt<'_>, expr: &Expr<'_>) -> bool {
     }
 }
 
+/// Checks whether the given `Range` is equivalent to a `RangeFull`.
+/// Inclusive ranges are not considered because they already constitute a lint.
+pub fn is_range_full(cx: &LateContext<'_>, container: &Expr<'_>, range: Range<'_>) -> bool {
+    range.start.map_or(true, |e| is_integer_const(cx, e, 0))
```
llogiq, line 1498: Since you made this a `clippy_utils` function, we should take care to either document any possible shortcomings or rule them out. For example, I think the method will fail on a `Range<i8>`.

llogiq:
I only found a few nits, with the additional tests and the wrong lower bound in the `is_range_full` function documented or fixed (it only works correct for unsigned types, because for signed types `MIN` != 0), I'd gladly merge this.

clippy_utils/src/lib.rs
bluthej: Thx for pointing that out! It didn't cross my mind...

Unfortunately I don't see a way to make it work for signed integers because right now the test on the lower bound is performed with `is_integer_const`, which takes a `u128` as a value to check for, so I can't check for `i8::MIN`.

Do you have some way that that could be accomplished in mind? Otherwise, I will document the function more precisely.

I have to admit that the applicability of this function is a bit awkward, for example I don't think it works on `..usize::MAX` either. It was made specifically to detect `RangeFull`s or equivalents used to index `Vec`s and `VecDeque`s, but the way it's written it also works in a few other cases. Maybe I can tweak the implementation to only check for `RangeTo`, `RangeFrom` and `Range` if the range is the argument of a method call of a `Vec` or `VecDeque`, what do you think?

llogiq: I guess the easiest way is to rename to `is_usize_range_full` and document the restricted applicability of the method. Otherwise you can get the type (via `expr_ty`), then the generic, then match on that to figure out the bounds.

bluthej: It might even be better to restrict the check to ranges passed to `drain` for now (apart from `.. ` which is the easy case). That makes it less flexible but at the same time it does prevent any misuse.

bluthej: Thx for the pointer, I will try to get the generic case working and if I really can't I will fall back to the `usize` solution with added documentation.

llogiq: Cool! Feel free to ping us here or on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy) if you need further assistance.

tests/ui/clear_with_drain.rs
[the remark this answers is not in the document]
bluthej: I added some tests where the `drain` iterator is used, in both the 'should trigger' and 'should not trigger' cases

clippy_utils/src/lib.rs
bluthej: It took a fair amount of work but I think I got it! At least it works as expected in my use case, and I added a couple of tests where I use `usize::MIN` as a lower bound to make sure that works as well (it already worked before with the hard-coded 0 but I thought it would be nice to make sure that it works).

I haven't added tests for the case where signed integers are used because this lint only applies in a method call to `drain` which only takes `usize` ranges. Do you need me to add unit tests for the new `is_range_full` function before accepting this PR? If so, where are such tests placed in the repo?

Note that I had to significantly alter the function signature of `is_range_full` to make it more generic. For instance, I made the container optional so that this function can be called on any kind of range.

tests/ui/clear_with_drain.rs
```diff
@@ -0,0 +1,84 @@
...
+    let mut v = vec![1, 2, 3];
+    v.drain(usize::MIN..v.len()); // Nay
+}
+
+fn range_from() {
+    let mut v = vec![1, 2, 3];
+    let iter = v.drain(0..); // Yay
+
+    let mut v = vec![1, 2, 3];
+    let next = v.drain(0..).next(); // Yay
```
llogiq, line 27: Could we have an example where the iter is only used later?

bluthej: Sure thing!

llogiq:
Thank you!

@bors r+

PR #10531, opened 2023-03-22
Add explanation on how to run `cargo-clippy` and `clippy-driver`

Nilstrieb:
I don't know how this would be done on Windows or MacOS, feedback appreciated.

I did find this snippet for MacOS but it seems a little sketchy.

https://github.com/rust-lang/rust-clippy/blob/1d1e72308e5bcdf5abe95c1d29dfe82524940061/.github/workflows/clippy_bors.yml#L98-L102

changelog: none

### DIFF
--- a/book/src/development/adding_lints.md
+++ b/book/src/development/adding_lints.md
@@ -18,6 +18,7 @@ because that's clearly a non-descriptive name.
     - [Cargo lints](#cargo-lints)
   - [Rustfix tests](#rustfix-tests)
   - [Testing manually](#testing-manually)
+  - [Running directly](#running-directly)
   - [Lint declaration](#lint-declaration)
   - [Lint registration](#lint-registration)
   - [Lint passes](#lint-passes)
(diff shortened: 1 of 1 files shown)


book/src/development/adding_lints.md
```diff
@@ -186,6 +186,13 @@ cargo dev lint input.rs
 from the working copy root. With tests in place, let's have a look at
 implementing our lint now.
 
+## Getting `cargo-clippy` and `clippy-driver` to work
```
giraffate, line 189: Can this be moved to https://github.com/rust-lang/rust-clippy/blob/master/book/src/usage.md#using-clippy-without-cargo-clippy-driver?

Nilstrieb: That looks like more of a user guide? This is only really for development

giraffate: Oh, you are right. I misunderstood it.

Question: In what situations would we use it directly during development?

Nilstrieb: I used it for perf testing so I could do `perf record cargo-clippy`. I'm not sure whether it's required but I have some problems with `cargo dev lint`. Although I had a lot of problems with everything to get everything to work with frame pointers :D. I personally also just like having direct access to stuff, but that's not that important.

giraffate (approved):
It looks good, thanks! Can the link be added to here: https://github.com/rust-lang/rust-clippy/pull/10531/files#diff-a5ade8b3f29bfda4754336da44290868956c0617b33e3dab780bee150c5d6208R21

giraffate:
@bors r+

Thanks!

PR #10533, opened 2023-03-22
Significantly optimize `significant_drop_tightening`

### LINKED ISSUE #10532: `significant_drop_tightening` is very slow
### Description

`significant_drop_tightening` is very slow. I tried running clippy with `-Aclippy::all -Wclippy::perf` on `rustc_borrowck`. (note that it seemed very slow on the entire compiler, so I don't think there's anything specific to `borrowck`, it's just very big.

According to a profile, it spends about 66% of time inside `has_sig_drop_attr`:

https://github.com/rust-lang/rust-clippy/blob/1d1e72308e5bcdf5abe95c1d29dfe82524940061/clippy_lints/src/significant_drop_tightening.rs#L307

More precisely, it seems to spend the time getting the type of fields.

https://github.com/rust-lang/rust-clippy/blob/1d1e72308e5bcdf5abe95c1d29dfe82524940061/clippy_lints/src/significant_drop_tightening.rs#L321

Inside this function it substitutes the type of the field with the substs (generic parameters) of the type. This folds the type and is a little work. This lint does that little work on everything, turning it into a lot of work. It also does so eagerly recursively which is just a receipe for slowness.

I've got a simple fix, I'll send in a PR.

### Version

```text
current master, e64c5961dc7640b4d5529608e0f321d586c027a0, 1.70.0
```

### Additional Labels

@rustbot label +C-bug

Nilstrieb:
The lint is very slow as it doesn't cache the deeply nested check for the attribute. If we cache it, we can reduce the time spent on checking `rustc_borrowck` from 28s to 9s, which is a nice improvement. In the profile, the time inside `has_sig_drop_attr` goes from 66% to 0.2%, which is a lot more reasonable.

<details>
<summary>Flame graphs</summary>

Before (all the tall `clippy` towers are `has_sig_drop_attr`):
before

After:
after
</details>

Fixes #10532

changelog: [`significant_drop_tightening`]: significantly optimized

### DIFF
--- a/clippy_lints/src/significant_drop_tightening.rs
+++ b/clippy_lints/src/significant_drop_tightening.rs
@@ -1,9 +1,9 @@
-use crate::FxHashSet;
 use clippy_utils::{
     diagnostics::span_lint_and_then,
     get_attr,
     source::{indent_of, snippet},
 };
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_errors::{Applicability, Diagnostic};
 use rustc_hir::{
     self as hir,
@@ -58,6 +58,7 @@ impl_lint_pass!(SignificantDropTightening<'_> => [SIGNIFICANT_DROP_TIGHTENING]);
 pub struct SignificantDropTightening<'tcx> {
     /// Auxiliary structure used to avoid having to verify the same type multiple times.
     seen_types: FxHashSet<Ty<'tcx>>,
+    type_cache: FxHashMap<Ty<'tcx>, bool>,
 }
 
 impl<'tcx> SignificantDropTightening<'tcx> {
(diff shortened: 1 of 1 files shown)


Nilstrieb:
I'd also like to nominate this for a beta backport (if the lint is in beta already, which I think it is), as this seems to affect uses where nursery was never enabled (which is kinda weird?).

clippy_lints/src/significant_drop_tightening.rs
```diff
@@ -58,6 +58,7 @@ impl_lint_pass!(SignificantDropTightening<'_> => [SIGNIFICANT_DROP_TIGHTENING]);
 pub struct SignificantDropTightening<'tcx> {
     /// Auxiliary structure used to avoid having to verify the same type multiple times.
     seen_types: FxHashSet<Ty<'tcx>>,
+    type_cache: FxHashMap<Ty<'tcx>, bool>,
```
WaffleLapkin, line 61: Why do we need both? Is `type_cache` not enough?

Nilstrieb: We need one for cycle detection (during the recursive call) and one for caching (after the recursive call)

Alexendoo:
> That's odd. How does the lint affect non-nursery uses? Does the lint fire even when nursery is not enabled?

The lint pass still runs but the lint won't be emitted if it's not enabled, related: https://github.com/rust-lang/rust/issues/106983

flip1995:
https://github.com/rust-lang/rust/pull/110273

PR #10534, opened 2023-03-23
Flag `bufreader.lines().filter_map(Result::ok)` as suspicious

samueltardieu:
This lint detects a problem that happened recently in https://github.com/uutils/coreutils and is described in https://github.com/rust-lang/rust/issues/64144. 

changelog: [`lines_filter_map_ok`]: new lint


samueltardieu:
The proposed lint now triggers with anything returning `std::io::Lines` (including `std::io::stdin().lines()`), and detects `|x| x.ok()` in addition to `Result::ok`.

A lintcheck with the 500 top most downloaded recent crates shows no hit.

llogiq:
I wonder why we should lint `stdin.lines()`. The infinite loop problem only appears on reading directories, and I haven't yet been able to pipe one to stdin.

samueltardieu:
> I wonder why we should lint `stdin.lines()`. The infinite loop problem only appears on reading directories, and I haven't yet been able to pipe one to stdin.

On Linux : 

```
$ cat t.rs
fn main() {
    for l in std::io::stdin().lines() {
        println!("l = {l:?}");
    }
}
$ rustc t.rs
$ ./t < /
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
l = Err(Os { code: 21, kind: IsADirectory, message: "Is a directory" })
^C
```

llogiq:
Ok then. I think the docs could use some editing for wording, otherwise it looks ok.

samueltardieu:
> Ok then. I think the docs could use some editing for wording, otherwise it looks ok.

I have reworked the wording, tell me if you'd change other things.

llogiq (commented):
The only possible problem is that the user could check for `in.is_file()` manually, in which case we'd have a false positive. That said, I'd be ok with just adding that to the "known problems" section, as the code to check for this case could be quite hairy

samueltardieu:
Note that even if `in.is_file()`, an I/O error while reading the file (*e.g.*, a USB key removed while reading) will likely persist, in which case the implementation of `Lines.next()` will return an `Err` forever (I have also verified this experimentally by creating a small fuse FS returning a read error).

I've added a "Known problems" section to the lint.

I've used "suspicious" as a category. Is that ok, or do you prefer it to be put in the "nursery"?

llogiq:
I think suspicious is likely fine, as the lint highlights a real problem and I haven't encountered a case where ignoring an error on one line would have produced a good outcome.

@bors r+

PR #10536, opened 2023-03-23
Add suggestions to `extra_unused_type_parameters`

mkrasnitski:
Change the `extra_unused_type_parameters` lint to provide machine applicable suggestions rather than just help messages. Exception to this are cases when any unused type parameters appear bounded in where clauses - for now I've deemed these cases unfixable and separated them out. Future work might be able to provide suggestions in these cases.

Also, added a test case for the `avoid_breaking_exported_api` config option.

r? @flip1995 

changelog: [`extra_unused_type_parameters`]: Now provides fixable suggestions.


flip1995 (approved):
This looks like it was more work than I anticipated. Thanks for doing this. Making this suggestion auto-applicable is really awesome!

flip1995:
@bors r+

Sorry for taking so long.

PR #10543, opened 2023-03-24
Add `tests_outside_test_module` lint

blyxyas:
Adds `tests_outside_test_module` from #10506. This PR **doesn't** close the issue, just resolves task 1.

changelog: [`tests_outside_test_module`]: The lint has been added

### DIFF
--- a/clippy_lints/src/tests_outside_test_module.rs
+++ b/clippy_lints/src/tests_outside_test_module.rs
@@ -6,13 +6,11 @@ use rustc_span::{def_id::LocalDefId, Span};
 
 declare_clippy_lint! {
     /// ### What it does
-    ///
-    /// Triggers when a testing function (marked with the `#[test]` attribute) isn't inside a testing module (marked with `#[cfg(test)]`).
-    ///
+    /// Triggers when a testing function (marked with the `#[test]` attribute) isn't inside a testing module
+    /// (marked with `#[cfg(test)]`).
     /// ### Why is this bad?
-    ///
-    /// The idiomatic (and more performant) way of writing tests is inside a testing module (flagged with `#[cfg(test)]`), having test functions outside of this module is confusing and may lead to them being "hidden".
-    ///
+    /// The idiomatic (and more performant) way of writing tests is inside a testing module (flagged with `#[cfg(test)]`),
+    /// having test functions outside of this module is confusing and may lead to them being "hidden".
     /// ### Example
     /// ```rust
     /// #[test]
@@ -39,7 +37,7 @@ declare_clippy_lint! {
     #[clippy::version = "1.70.0"]
     pub TESTS_OUTSIDE_TEST_MODULE,
     restriction,
-    "The test function `my_cool_test` is outside the testing module `tests`."
+    "A test function is outside the testing module."
 }
 
 declare_lint_pass!(TestsOutsideTestModule => [TESTS_OUTSIDE_TEST_MODULE]);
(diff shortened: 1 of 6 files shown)


clippy_lints/src/tests_outside_test_module.rs
```diff
@@ -0,0 +1,101 @@
...
+impl TestsOutsideTestModule {
+    pub fn new() -> Self {
+        Self { test_mod_exists: false }
+    }
+}
+
+impl_lint_pass!(TestsOutsideTestModule => [TESTS_OUTSIDE_TEST_MODULE]);
+
+impl LateLintPass<'_> for TestsOutsideTestModule {
+    fn check_mod(&mut self, cx: &LateContext<'_>, _: &Mod<'_>, _: HirId) {
```
flip1995, line 59: This should probably be `check_crate` to really only do it once and right at the beginning.

Also: What if the test module is in a completely different file, than the test function? This `check_mod` (or `check_crate`) impl doesn't know the difference. So I'm not sure if checking for it even makes sense. For a restriction lint, I would just not check.

blyxyas: `check_mod` is necessary for cases like `test/outer/inner/file.rs`. It works like that on Clippy UI tests (It fails with `check_crate` if it's in a test sub-sub-directory like `tests/ui/tests_outside_test_module/file.rs`).

---

Then maybe changing the lint to pedantic? I think it's pretty universal that if a testing module exists, and some tests are outside of that module, they should be in there.

flip1995: Thinking about this more: Is it relevant if there is a `cfg(test)` module in the first place? I would say no. 

If it is good advice to move a test function to such a module, then it is probably also a good idea to create such a test module.

What do you think about removing this check completely?

blyxyas: > If it is good advice to move a test function to such a module, then it is probably also a good idea to create such a test module.

You're right. I'm going to remove the check.

clippy_lints/src/tests_outside_test_module.rs
```diff
@@ -0,0 +1,73 @@
...
+    ///     #[test]
+    ///     fn my_cool_test() {
+    ///         // [...]
+    ///     }
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub TESTS_OUTSIDE_TEST_MODULE,
+    restriction,
+    "The test function `my_cool_test` is outside the testing module `tests`."
```
flip1995, line 42:
```suggestion
    "A test function is outside the testing module."
```

clippy_lints/src/tests_outside_test_module.rs
```diff
@@ -0,0 +1,73 @@
...
+use rustc_span::{def_id::LocalDefId, Span};
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Triggers when a testing function (marked with the `#[test]` attribute) isn't inside a testing module (marked with `#[cfg(test)]`).
+    ///
+    /// ### Why is this bad?
+    ///
+    /// The idiomatic (and more performant) way of writing tests is inside a testing module (flagged with `#[cfg(test)]`), having test functions outside of this module is confusing and may lead to them being "hidden".
```
flip1995, line 14: This line should be formatted. I'm surprised this passes CI :thinking:

flip1995 (approved):
Only some major nitpicking.

clippy_lints/src/tests_outside_test_module.rs
blyxyas: Fixed in https://github.com/rust-lang/rust-clippy/pull/10543/commits/b2f919182045773ee2a4091ef071ffd3183efe45

clippy_lints/src/tests_outside_test_module.rs
blyxyas: Fixed in https://github.com/rust-lang/rust-clippy/pull/10543/commits/b2f919182045773ee2a4091ef071ffd3183efe45

blyxyas:
They called me Speedy McQuick in high school

Issue #10548, opened 2023-03-25
unnecessary-struct-initialization changes code to require type annotations (which are not provided)


matthiaskrgr:
### Summary

.

### Reproducer

I tried this code:

```rust
#[derive(Default)]
struct NonGeneric {
    field1: usize,
}

fn main() {
    let default1 = NonGeneric { ..Default::default() };
}

```

clippy suggestion:
````
warning: unnecessary struct building
 --> src/main.rs:9:20
  |
9 |     let default1 = NonGeneric { ..Default::default() };
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace with: `Default::default()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_struct_initialization
```

This gets auto applied with `--fix` but it does not compile since we are lacking type annotations for `Default`

```
The following errors were reported:
error[E0282]: type annotations needed
 --> src/main.rs:9:9
  |
9 |     let _default1 = Default::default();
  |         ^^^^^^^^^
  |
help: consider giving `_default1` an explicit type
  |
9 |     let _default1: /* Type */ = Default::default();
  |                  ++++++++++++

error: aborting due to previous error
```

### Version

```text
rustc 1.70.0-nightly (8be3c2bda 2023-03-24)
binary: rustc
commit-hash: 8be3c2bda6b683f87b24714ba595e8b04faef54c
commit-date: 2023-03-24
host: x86_64-unknown-linux-gnu
release: 1.70.0-nightly
LLVM version: 15.0.7
```


### Additional Labels

_No response_

Issue #10547, opened 2023-03-25
FP unnecessary_struct_initialization, suggestion does not compile


matthiaskrgr:
### Summary

.

### Lint Name

unnecessary_struct_initialization

### Reproducer

I tried this code:

```rust
struct S(i32);

// We should evaluate `x[2]` and copy the value out *before* evaluating the LHS
// and changing its value.
fn _fun() {
    let mut x = &[S(0), S(1), S(2)][..];
    let y = &mut S(7);
    *{
        x = &[S(3), S(4), S(5)];
        &mut *y
    } = S { ..x[2] };
    assert_eq!(2, y.0);
    assert_eq!(5, x[2].0);
}

pub fn main() {}

```

I saw this happen:

```
warning: unnecessary struct building
  --> src/main.rs:11:9
   |
11 |     } = S { ..x[2] };
   |         ^^^^^^^^^^^^ help: replace with: `x[2]`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_struct_initialization
   = note: `#[warn(clippy::unnecessary_struct_initialization)]` on by default
```

The suggested code does not compile:
````
The following errors were reported:
error[E0508]: cannot move out of type `[S]`, a non-copy slice
  --> src/main.rs:11:9
   |
11 |     } = x[2];
   |         ^^^^
   |         |
   |         cannot move out of here
   |         move occurs because `x[_]` has type `S`, which does not implement the `Copy` trait

error: aborting due to previous error

For more information about this error, try `rustc --explain E0508`.
Original diagnostics will follow.
````

### Version

```text
rustc 1.70.0-nightly (8be3c2bda 2023-03-24)
binary: rustc
commit-hash: 8be3c2bda6b683f87b24714ba595e8b04faef54c
commit-date: 2023-03-24
host: x86_64-unknown-linux-gnu
release: 1.70.0-nightly
LLVM version: 15.0.7
```

Alexendoo:
More generally it doesn't handle place expressions where the value cannot be moved out of, or where moving it would cause another error

```rust
#![allow(unused, clippy::no_effect)]

struct A {
    x: i32
}

fn borrowed(a: A) {
    let b = &a;
    A { ..a };
    b;
}

fn would_be_moved(a: A) {
    A { ..a };
    A { ..a };
}

struct S {
    a: A
}

fn field(s: &S) {
    A {..s.a}
}
```

PR #10552, opened 2023-03-26
Move unnecessary_struct_initialization to nursery

Alexendoo:
changelog: none, assuming it makes into the same release as #10489

Mostly because of #10547 but there is also #10548


flip1995:
Thanks!

@bors r+

Side note: I find this lint a bit weird in the first place. `S { ..a }` is not really a syntax Rust beginners are aware of IMO. So if someone uses this _unusual_ way to initialize a struct, there is probably a good reason why they didn't use just `a`. The only other reason I can think of where this could happen is during a refactor removing a struct field.

PR #10553, opened 2023-03-26
In uninit checking, add fallback for polymorphic types

### LINKED ISSUE #10551: False positive for uninit lint
### Summary

Code like [this](https://github.com/zesterer/chumsky/blob/2be5cd420666e9d6ec07cbf89eb6440be2723d1d/src/private.rs#L277-L281) produces an erroneous lint emission despite the code being safe (`MaybeUninit<T>` has no inhabitants).

### Lint Name

uninit_assumed_init

### Reproducer

I tried this code:

```rust
trait MaybeUninitExt<T>: Sized {
    /// Identical to the unstable [`MaybeUninit::uninit_array`]
    fn uninit_array<const N: usize>() -> [Self; N];
}

impl<T> MaybeUninitExt<T> for MaybeUninit<T> {
    fn uninit_array<const N: usize>() -> [Self; N] {
        // SAFETY: Output type is entirely uninhabited - IE, it's made up entirely of `MaybeUninit`
        unsafe { MaybeUninit::uninit().assume_init() }
    }
}
```

I saw this happen:

```
error: this call for this type may be undefined behavior
   --> src/private.rs:280:18
    |
280 |         unsafe { MaybeUninit::uninit().assume_init() }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

*nothing*

### Version

_No response_

### Additional Labels

_No response_

Nilstrieb:
After #10520, we always assumed that polymorphic types do not allow to be left uninitialized. But we can do better, by peeking into polymorphic types and adding a few special cases for going through tuples, arrays (because the length may be polymorphic) and blanket allowing all unions (like MaybeUninit).

fixes #10551

changelog: [uninit_vec]: fix false positive for polymorphic types
changelog: [uninit_assumed_init]: fix false positive for polymorphic types

### DIFF
--- a/clippy_utils/src/ty.rs
+++ b/clippy_utils/src/ty.rs
@@ -541,9 +541,25 @@ pub fn same_type_and_consts<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> bool {
 pub fn is_uninit_value_valid_for_ty<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
     cx.tcx
         .check_validity_requirement((ValidityRequirement::Uninit, cx.param_env.and(ty)))
-        // For types containing generic parameters we cannot get a layout to check.
-        // Therefore, we are conservative and assume that they don't allow uninit.
-        .unwrap_or(false)
+        .unwrap_or_else(|_| is_uninit_value_valid_for_ty_fallback(cx, ty))
+}
+
+/// A fallback for polymorphic types, which are not supported by `check_validity_requirement`.
+fn is_uninit_value_valid_for_ty_fallback<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
+    match *ty.kind() {
+        // The array length may be polymorphic, let's try the inner type.
+        ty::Array(component, _) => is_uninit_value_valid_for_ty(cx, component),
+        // Peek through tuples and try their fallbacks.
+        ty::Tuple(types) => types.iter().all(|ty| is_uninit_value_valid_for_ty(cx, ty)),
+        // Unions are always fine right now.
+        // This includes MaybeUninit, the main way people use uninitialized memory.
+        // For ADTs, we could look at all fields just like for tuples, but that's potentially
+        // exponential, so let's avoid doing that for now. Code doing that is sketchy enough to
+        // just use an `#[allow()]`.
+        ty::Adt(adt, _) => adt.is_union(),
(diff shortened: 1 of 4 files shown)


clippy_utils/src/ty.rs
```diff
@@ -541,9 +541,25 @@ pub fn same_type_and_consts<'tcx>(a: Ty<'tcx>, b: Ty<'tcx>) -> bool {
...
+/// A fallback for polymorphic types, which are not supported by `check_validity_requirement`.
+fn is_uninit_value_valid_for_ty_fallback<'tcx>(cx: &LateContext<'tcx>, ty: Ty<'tcx>) -> bool {
+    match *ty.kind() {
+        // The array length may be polymorphic, let's try the inner type.
+        ty::Array(component, _) => is_uninit_value_valid_for_ty(cx, component),
+        // Peek through tuples and try their fallbacks.
+        ty::Tuple(types) => types.iter().all(|ty| is_uninit_value_valid_for_ty(cx, ty)),
+        // Unions are always fine right now.
+        // This includes MaybeUninit, the main way people use uninitialized memory.
+        ty::Adt(adt, _) if adt.is_union() => true,
```
samueltardieu, line 556: Since other branches don't use guards, why not use `ty::Adt(adt, _) => adt.is_union()`?

Alexendoo:
Would be good to add a test for #10565 as well

Alexendoo (approved):

pickfire:
I believe this still does not fully fix the issue, monoio still get false positive. https://github.com/bytedance/monoio/actions/runs/4635132256/jobs/8205680770?pr=157

```rust
struct Page<T> {
    // continued buffer of fixed size
    slots: Box<[MaybeUninit<Entry<T>>]>,
}

impl<T> Page<T> {
    fn new(size: usize, prev_len: usize) -> Self {
        let mut buffer = Vec::with_capacity(size);
        unsafe { buffer.set_len(size) };
        let slots = buffer.into_boxed_slice();
        Self {
            slots,
        }
    }
}
```

Alexendoo:
The fix is not yet available on nightly, it will be in the one after https://github.com/rust-lang/rust/pull/110003 is merged

I'll add a beta-nominated label to ensure it's in the same release as https://github.com/rust-lang/rust-clippy/pull/10520, not sure how that works out timing wise

flip1995:
This will get into the same beta/stable version as #10520. No action required AFAICT.

Issue #10550, opened 2023-03-26
rustdoc: `synthetic_non_exhaustive` lint


Ezrashaw:
### What it does

See discussion in https://github.com/rust-lang/rust/issues/108925 and https://github.com/rust-lang/rust/pull/109007.

`enum`'s which have a `#[doc(hidden)]` attribute are effectively `#[non_exhaustive]`. Sometimes a `#[doc(hidden)]` `__Nonexhaustive` variant (dating from before `#[non_exhaustive]`) exists, which . Even in the absence of a explicit variant to artificially synthesize a `#[non_exhaustive]` attribute, code which has "hidden" variants should mark itself as non exhaustive to be explicit; they are effectively non-exhaustive.

### Lint Name

synthetic_non_exhaustive

### Category

style

### Advantage

Once again, see discussion in linked issues.

`enum`s which have `#[doc(hidden)]` variants are effectively non-exhaustive and should be marked as such to give clarity to the consumer of the `enum`.

### Drawbacks

See discussion in linked issues (last time I promise).

In summary, we aren't sure about the extent to which this affects existing code and how much of that code is, in actuality, valid. For example, `libcore` and `libbacktrace` *are triggered by this lint*: in `libbacktrace` at least, the offending code is not idiomatic and is a valid example of this lint. 

### Example

```rust
pub enum Foo {
    A,
    B,
    #[doc(hidden)]
    C,
}
```

Could be written as:

```rust
#[non_exhaustive]
pub enum Foo {
    A,
    B,
    #[doc(hidden)]
    C,
}
```

matt115200:
Hi! This didn't seem too tricky, hoping to push a PR for this today.

@rustbot claim

PR #10559, opened 2023-03-28
new lint: `synthetic_non_exhaustive`

matt115200:
Closes: https://github.com/rust-lang/rust-clippy/issues/10550

changelog: [`synthetic_non_exhaustive`]: new line


matt115200:
So if it was decided that we wanted this behaviour, would the best solution be to remove the underscore checks from `manual_non_exhaustive`?

I should have searched harder and found that first :) Looks like I missed some stuff like checking msrv even supports `#[non_exhaustive]`, and creating some nicer output with `span_suggestion`.

Jarcho:
It would probably be ok to remove the underscore check. The check to see if the variant is actually created within the defining crate should be enough to catch where `#[doc(hidden)]` is used to indicate a private variant.

PR #10562, opened 2023-03-29
make `as_underscore` lint on _ as the type of a ptr cast

asquared31415:
This makes `as_underscore` lint on cases where the type of a raw pointer `as` cast is inferred.  This can be very dangerous, such as the following code:
```rust
fn as_ptr(&self) -> *const u8 {
    &self as *const _ as *const u8
}
```
which now shows ``help: consider giving the type explicitly: `*const &UwU` ``, and should hopefully help the programmer notice that they've incorrectly turned a double pointer into a single pointer.

changelog: `as_underscore` now lints on cases where the type of a raw pointer `as` cast is an underscore, such as `x as *const _`.

### DIFF
--- a/clippy_lints/src/casts/as_underscore.rs
+++ b/clippy_lints/src/casts/as_underscore.rs
@@ -1,12 +1,20 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, Ty, TyKind};
+use rustc_hir::{Expr, MutTy, Ty, TyKind};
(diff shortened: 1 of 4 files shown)


asquared31415:
This probably could serve to *also* be some warn by default lint, but `as_underscore` definitely should detect this, whether or not another lint exists.

asquared31415:
Upon further thought, this probably does need to be its own lint only, since it's more dangerous and needs special consideration by the user to allow/deny.  `allow(as_underscore)` should not silence this.

PR #10563, opened 2023-03-29
Partial no-op refactoring of #9948

nyurik:
This contains some prep work for #9948 to keep that change to the minimum, and make it easier to review it.

This should be a noop, but it has some tests from that PR discussion, and should help in the future with the corner case format handling.

cc: @Alexendoo @llogiq @xFrednet  as the 3 people who reviewed the parent PR

----

changelog: none


llogiq:
Ok, so all I see is introducing some inline format args in the code and extending one ui test. All right, let's do this.

@bors r+

PR #10564, opened 2023-03-29
make cast_possible_wrap work correctly for 16 bit {u,i}size

### LINKED ISSUE #9337: `cast_possible_wrap` for convertion from `u8` to `isize`
### Summary

The lint `cast_possible_wrap` is triggered by a conversion from `u8` to `isize`.

The displayed message is:

    casting `u8` to `isize` may wrap around the value on targets with 32-bit wide pointers

On 16/32/64-bits platforms, the message is wrong. An 8-bit unsigned value fits in a 16/32/64-bits signed register.

### Reproducer

```rust
#![warn(clippy::cast_possible_wrap)]
const A: u8 = 0;
const B: isize = A as isize;
```

### Version

```text
rustc 1.63.0 (4b91a6ea7 2022-08-08)
binary: rustc
commit-hash: 4b91a6ea7258a947e59c6522cd5898e7c0a6a88f
commit-date: 2022-08-08
release: 1.63.0
LLVM version: 14.0.5
```

### Additional Labels

@rustbot label +I-false-positive +L-pedantic

asquared31415:
These changes make `cast_possible_wrap` aware of the different pointer widths and fixes the implementation to print the correct pointer widths.

Fixes #9337 

changelog: `cast_possible_wrap` does not lint on `u8 as isize` or `usize as i8`, since these can never wrap.
`cast_possible_wrap` now properly considers 16 bit pointer size and prints the correct bit widths.

### DIFF
--- a/clippy_lints/src/casts/cast_possible_wrap.rs
+++ b/clippy_lints/src/casts/cast_possible_wrap.rs
@@ -1,41 +1,81 @@
 use clippy_utils::diagnostics::span_lint;
-use clippy_utils::ty::is_isize_or_usize;
 use rustc_hir::Expr;
 use rustc_lint::LateContext;
 use rustc_middle::ty::Ty;
 
 use super::{utils, CAST_POSSIBLE_WRAP};
 
+// this should be kept in sync with the allowed bit widths of `usize` and `isize`
+const ALLOWED_POINTER_SIZES: [u64; 3] = [16, 32, 64];
+
+// whether the lint should be emitted, and the required pointer size, if it matters
+enum EmitState {
+    NoLint,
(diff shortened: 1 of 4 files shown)


asquared31415:
r? rust-lang/clippy

It has been a couple months, and there hasn't been a review, so rerolling, unless there's another process to request a review?

giraffate (commented):
Overall looks good, thanks!

Can we add the note about `usize and isize are at least 16-bits wide.` in the reference?
https://github.com/rust-lang/rust-clippy/issues/9337#issuecomment-1332190129

giraffate:
@bors r+

Thanks!

PR #10567, opened 2023-03-30
add new lint `as_underscore_ptr` to check for `as *{const,mut} _`

asquared31415:
This probably could go in `suspicious`, but I was unsure where exactly to put it.  It's not necessarily wrong code on its own, but can be problematic in certain situations.

changelog: [`as_underscore_ptr`]: a new lint to detect use of `as *const _` or `as *mut _` which can be the cause of subtle raw pointer bugs.

### DIFF
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -1,5 +1,6 @@
 mod as_ptr_cast_mut;
 mod as_underscore;
+mod as_underscore_ptr;
 mod borrow_as_ptr;
 mod cast_abs_to_unsigned;
 mod cast_enum_constructor;
@@ -557,6 +558,52 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
+    /// Checks for the usage of `as *const _` or `as *mut _` where the pointed to type is inferred.
+    ///
+    /// ### Why is this bad?
+    /// When converting to a pointer, it can be dangerous to not specify the type. Type inference can
+    /// be affected by many things, types may not always be obvious, and when working with pointers, while
+    /// the pointed to type doesn't technically matter until an access, you should always know what types
+    /// you intend to work with. When using multiple chained `as` casts, it can be especially dangerous,
+    /// because `*const T as *const U` **always compiles** (for Sized types T and U).
+    ///
+    /// ### Example
+    /// ```rust
+    /// // intent: turn a `&u64` into a `*const u8` that points to the bytes of the u64 (⚠️does not work⚠️)
+    /// fn owo(x: &u64) -> *const u8 {
+    ///     // ⚠️ `&x` is a `&&u64`, so this turns a double pointer into a single pointer
+    ///     // ⚠️ This pointer is a dangling pointer to a local
+    ///     &x as *const _ as *const u8
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// // intent: turn a `&u64` into a `*const u8` that points to the bytes of the u64 (⚠️does not work⚠️)
+    /// fn owo(x: &u64) -> *const u8 {
+    ///     // ⚠️ `&x` is a `&&u64`, so this turns a double pointer into a single pointer
+    ///     // ⚠️ This pointer is a dangling pointer to a local
+    ///     &x as *const &u64 as *const u8
+    /// }
+    /// ```
+    /// While this is still buggy, the bug is more visible here: you have a `*const &u64`. To actually fix the
+    /// underlying issue, use:
+    /// ```rust
+    /// // turn a `&u64` into a `*const u8` that points to the bytes of the u64
+    /// fn owo(x: &u64) -> *const u8 {
+    ///     x as *const u64 as *const u8
+    /// //  ^ now `x` instead of `&x`
+    /// }
+    /// ```
+    /// This lint cannot suggest this final code because it's a more broad `restriction` lint that forces the
+    /// user to be more specific, and this transformation is not always valid.
+    #[clippy::version = "1.70.0"]
+    pub AS_UNDERSCORE_PTR,
+    restriction,
+    "detects `as *{const,mut} _ conversions"
+}
+
+declare_clippy_lint! {
+    /// ### What it does
     /// Checks for the usage of `&expr as *const T` or
     /// `&mut expr as *mut T`, and suggest using `ptr::addr_of` or
     /// `ptr::addr_of_mut` instead.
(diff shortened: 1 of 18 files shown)


asquared31415:
It currently is pedantic because I was aware that lots of code uses these casts (not just in the mutability case).  But yes it lints every time the target type of a cast is any pointer to `_`.

asquared31415:
Linting on "a pointer cast that changes the type of the referent" for ptr->ptr is just the current lint that suggests `ptr.cast::<T>()`.  ref->ptr casts cannot change the type of the referent, as enforced by the compiler.
There probably should be a lint for `*const T as *mut T` (or the other way) without changing the type that suggests some unstable ptr methods as well, though idk what the clippy policy on that is.

Does clippy have a crater-like tool that I could use to better understand the cases where this fires and the lint's suggestion would be a downgrade?  I expected the lint to be moderately common, hence an allow-by-default group, but also from my understanding, a significant number of people try to avoid this pattern as much as reasonable, because of how tricky the casts can be.

The goal of this new lint is to find cases where the use of `as * _` produces a type that the user might not expect, or is liable to change without warning.  Some of the specific cases that  I had in mind are:
```rust
struct UwU;
impl UwU {
    // intent: turn a `&UwU` into a `*const u8` that points to the same data
    fn as_ptr(&self) -> *const u8 {
        // ⚠️ `&self` is a `&&UwU`, so this turns a double pointer into a single pointer
        // ⚠️ This pointer is a dangling pointer to a local
        &self as *const _ as *const u8
    }
}
```
:arrow_up: (turns a double pointer into a single pointer because the type of `self` is less obvious)
```rust
fn uwu(data: &[u64]) -> *const u64 { // changing the return type is dangerous
    data.as_ptr() as *const _
}
```
:arrow_up: (the pointer type could be easily changed by inference)
```rust
unsafe fn read_data(data: *const u32) { // 1. if this signature changes
    let _ = *data;
}

let arr = [1_u32, 2, 3, 4];
unsafe {
    let ptr = arr.as_ptr().add(2) as *const _; // 2. this inference can change
    /*
    some more code
    */
    read_data(ptr); // 3. because of this call
}
```
:arrow_up: (inference for a local is determined by how it gets used later)

Cases where the type is not obvious and cases where the type is obvious look *really similar*, and I'm not sure if clippy wants the complexity of picking out specific patterns for this lint, if the information such as "what controls this inference" is even available at all.

I would like to improve the lint so that people can use it via `pedantic` or possibly even `suspicious` if it can be very accurate, but if that's not reasonable/possible, would putting this lint in `restriction` as-is be acceptable, under the same reasoning as `as_underscore`?

asquared31415:
On your point of trying to exclude "the referent is unchanged", in the example I provide in the lint (and the first example in my previous comment) the problematic cast *is* "reference to pointer that does not change the pointed to type", `&self as *const _` (does `&&UwU -> *const &UwU`, and that's all the compiler allows).  The next cast is "cast a pointer to another pointer with an explicit type".  Under that suggested rule as-is, we miss the first cast, which is the suspicious cast, and the second cast always has to be fine, `ptr -> ptr` is a well defined and common pattern, the best we can do is suggest `.cast::<u8>()`, but that doesn't help this specific casting error.

Alexendoo:
Thanks for the examples!

Yeah at the very least this is a useful `restriction` lint, but higher than that would be good if we can get it not too loud

I saw the first example in the test file that didn't contain the `⚠️` comments explaining the issue. But just from the lint message in `.stderr` I think I glossed over the extra `&` in the type, it could be worth adding a note if the type is behind a reference to drive the point home. Maybe that could also be a heuristic itself

For the other two, if the lint didn't fire because the type behind the pointer is unchanged it would at least serve as a guard ensuring it stays that way

Alexendoo:
A route we could go is to land this as-is in `restriction`, then have any heuristic based ones as their own new lints. Because it would still be useful to be able to deny all inferred type casts

blyxyas:
The use of UwU instead of the boring `foo` and `bar` is epic.

KisaragiEffective:
IMHO - if you divide case - casting to nested-pointer (like `&&UwU`) by `as *{const,mut} _` and everything else for the cast, former could be `suspicious` (or, at least `pedantic`).

asquared31415:
I implemented the changes to this lint to show an extra note for some especially suspicious cases, and moved the lint down to `restriction`.  Are there any other concerns with this PR?

tests/ui/zero_ptr_no_std.rs
```diff
@@ -8,6 +8,7 @@
 fn main(_argc: isize, _argv: *const *const u8) -> isize {
     let _ = 0 as *const usize;
     let _ = 0 as *mut f64;
+    #[allow(clippy::as_underscore_ptr)] // trying to test inference with a different lint
```
Alexendoo, line 11: The `allow`s in various other files can be removed since it's `restriction`

clippy_lints/src/casts/mod.rs
```diff
@@ -555,6 +556,47 @@ declare_clippy_lint! {
...
+    /// ```rust
+    /// struct UwU;
+    /// impl UwU {
+    ///     // intent: turn a `&UwU` into a `*const u8` that points to the same data
+    ///     fn as_ptr(&self) -> *const u8 {
+    ///         // ⚠️ `&self` is a `&&UwU`, so this turns a double pointer into a single pointer
+    ///         // ⚠️ This pointer is a dangling pointer to a local
+    ///         &self as *const _ as *const u8
+    ///     }
+    /// }
```
Alexendoo, line 580: How about using something like this as our example? Has a non buggy case to contrast to and it's pretty plain since there's no ZSTs/self params to think about

```rust
fn good_cast(x: &u64) -> *const u8 {
    x as *const _ as *const u8
}

fn bad_cast(x: &u64) -> *const u8 {
    &x as *const _ as *const u8
}
```

Alexendoo (commented):
Thanks for that, just a couple things then it'd good to merge. Sorry for the delay!

asquared31415:
What do I do about that CI failure? It says that the `as_underscore_ptr` UI test has a `fixed` file (true) but is missing a `run-rustfix` annotation (false). However when I look at the "changed files" there's 100% a `// run-rustfix` as the very start of the file?

Alexendoo:
Ahh that's an unfortunate message, it should be `//@run-rustfix` since #10669, the message didn't get updated in `compiletest` but would be resolved by #10426

Alexendoo:
I brought this up in the last meeting re: it being `restriction` and a good way to catch the motivating example [was suggested](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202023-07-11/near/374303667), if we lint casts where the level of indirection decreases but is hidden behind a `_` that would be able to be warn by default

PR #10570, opened 2023-03-30
Add redundant type annotations lint

### LINKED ISSUE #9155: Warn on redundant / needless type annotations
### What it does

Warns about needless / redundant type annotations.

### Lint Name

redundant_type_annotation

### Category

style, pedantic

### Advantage

Code is more idiomatic, shorter, and easier to modify.

### Drawbacks

Types may be added by some novice Rust developers as hints to help remember what type is present.

### Example

```rust
let foo: String = String::new();
```

Could be written as:

```rust
let foo = String::new();
```

AlessioC31:
Hello, I'm trying to add the `redundat_type_annotations` lint.

It's still WIP but I'd like to start gathering some feedbacks to be sure that I'm not doing things 100% wrong :)

Right now it still misses lints like:

- [x] `let foo: u32 = 5_u32`,
- [x] `let foo: String = STest2::func()`
- [x] `let foo: String = self.func()` (`MethodCall`)
- [x] refs
- [ ] Generics

I've some problems regarding the second example above, in the `init` part of the `Local` I have:

```rust
init: Some(
                Expr {
                    hir_id: HirId(DefId(0:24 ~ playground[e1bd]::main).58),
                    kind: Call(
                        Expr {
                            hir_id: HirId(DefId(0:24 ~ playground[e1bd]::main).59),
                            kind: Path(
                                TypeRelative(
                                    Ty {
                                        hir_id: HirId(DefId(0:24 ~ playground[e1bd]::main).61),
                                        kind: Path(
                                            Resolved(
                                                None,
                                                Path {
                                                    span: src/main.rs:77:21: 77:27 (#0),
                                                    res: Def(
                                                        Struct,
                                                        DefId(0:17 ~ playground[e1bd]::STest2),
                                                    ),
                                                    segments: [
                                                        PathSegment {
                                                            ident: STest2#0,
                                                            hir_id: HirId(DefId(0:24 ~ playground[e1bd]::main).60),
                                                            res: Def(
                                                                Struct,
                                                                DefId(0:17 ~ playground[e1bd]::STest2),
                                                            ),
                                                            args: None,
                                                            infer_args: true,
                                                        },
                                                    ],
                                                },
                                            ),
                                        ),
                                        span: src/main.rs:77:21: 77:27 (#0),
                                    },
                                    PathSegment {
                                        ident: get_numb#0,
                                        hir_id: HirId(DefId(0:24 ~ playground[e1bd]::main).62),
                                        res: Err,
                                        args: None,
                                        infer_args: true,
                                    },
                                ),
                            ),
                            span: src/main.rs:77:21: 77:37 (#0),
                        },
                        [],
                    ),
                    span: src/main.rs:77:21: 77:39 (#0),
                },
            ),
```

And I'm not sure how to get the return type of the function `STest2::func()` since the resolved path `DefId` points to the struct itself and not the function. Do you have any idea on how I could get this information in this case?

Thanks!

changelog: changelog: [`redundant_type_annotations`]: New lint to warn on redundant type annotations

fixes #9155

### DIFF
--- a/clippy_lints/src/redundant_type_annotations.rs
+++ b/clippy_lints/src/redundant_type_annotations.rs
@@ -1,0 +1,185 @@
+use clippy_utils::diagnostics::span_lint;
+use rustc_ast::LitKind;
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Warns about needless / redundant type annotations.
+    ///
+    /// ### Why is this bad?
+    /// Code without type annotations is shorter and in most cases
+    /// more idiomatic and easier to modify.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let foo: String = String::new();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let foo = String::new();
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub REDUNDANT_TYPE_ANNOTATIONS,
+    pedantic,
+    "warns about needless / redundant type annotations."
+}
+declare_lint_pass!(RedundantTypeAnnotations => [REDUNDANT_TYPE_ANNOTATIONS]);
+
+fn is_same_type<'tcx>(cx: &LateContext<'tcx>, ty_resolved_path: hir::def::Res, func_return_type: Ty<'tcx>) -> bool {
+    // type annotation is primitive
+    if let hir::def::Res::PrimTy(primty) = ty_resolved_path
+        && func_return_type.is_primitive()
+        && let Some(func_return_type_sym) = func_return_type.primitive_symbol()
+    {
+        return primty.name() == func_return_type_sym;
+    }
+
+    // type annotation is any other non generic type
+    if let hir::def::Res::Def(_, defid) = ty_resolved_path
+        && let Some(annotation_ty) = cx.tcx.type_of(defid).no_bound_vars()
+    {
+        return annotation_ty == func_return_type;
+    }
+
+    false
+}
+
+fn func_hir_id_to_func_ty<'tcx>(cx: &LateContext<'tcx>, hir_id: hir::hir_id::HirId) -> Option<Ty<'tcx>> {
+    if let Some((defkind, func_defid)) = cx.typeck_results().type_dependent_def(hir_id)
+        && defkind == hir::def::DefKind::AssocFn
+        && let Some(init_ty) = cx.tcx.type_of(func_defid).no_bound_vars()
+    {
+        Some(init_ty)
+    } else {
+        None
+    }
+}
+
+fn func_ty_to_return_type<'tcx>(cx: &LateContext<'tcx>, func_ty: Ty<'tcx>) -> Option<Ty<'tcx>> {
+    if func_ty.is_fn()
+        && let Some(return_type) = func_ty.fn_sig(cx.tcx).output().no_bound_vars()
+    {
+        Some(return_type)
+    } else {
+        None
+    }
+}
+
+/// Extracts the fn Ty, e.g. `fn() -> std::string::String {f}`
+fn extract_fn_ty<'tcx>(
+    cx: &LateContext<'tcx>,
+    call: &hir::Expr<'tcx>,
+    func_return_path: &hir::QPath<'tcx>,
+) -> Option<Ty<'tcx>> {
+    match func_return_path {
+        // let a: String = f(); where f: fn f() -> String
+        hir::QPath::Resolved(_, resolved_path) => {
+            if let hir::def::Res::Def(_, defid) = resolved_path.res
+                && let Some(middle_ty_init) = cx.tcx.type_of(defid).no_bound_vars()
+            {
+                Some(middle_ty_init)
+            } else {
+                None
+            }
+        },
+        // Associated functions like
+        // let a: String = String::new();
+        // let a: String = String::get_string();
+        hir::QPath::TypeRelative(..) => func_hir_id_to_func_ty(cx, call.hir_id),
+        hir::QPath::LangItem(..) => None,
+    }
+}
+
+fn is_redundant_in_func_call<'tcx>(
+    cx: &LateContext<'tcx>,
+    ty_resolved_path: hir::def::Res,
+    call: &hir::Expr<'tcx>,
+) -> bool {
+    if let hir::ExprKind::Path(init_path) = &call.kind {
+        let func_type = extract_fn_ty(cx, call, init_path);
+
+        if let Some(func_type) = func_type
+            && let Some(init_return_type) = func_ty_to_return_type(cx, func_type)
+        {
+            return is_same_type(cx, ty_resolved_path, init_return_type);
+        }
+    }
+
+    false
+}
+
+impl LateLintPass<'_> for RedundantTypeAnnotations {
+    fn check_local<'tcx>(&mut self, cx: &LateContext<'tcx>, local: &'tcx rustc_hir::Local<'tcx>) {
+        // type annotation part
+        if !local.span.from_expansion()
+            && let Some(ty) = &local.ty
+
+            // initialization part
+            && let Some(init) = local.init
+        {
+            match &init.kind {
+                // When the initialization is a call to a function
+                hir::ExprKind::Call(init_call, _) => {
+                    if let hir::TyKind::Path(ty_path) = &ty.kind
+                        && let hir::QPath::Resolved(_, resolved_path_ty) = ty_path
+
+                        && is_redundant_in_func_call(cx, resolved_path_ty.res, init_call) {
+                        span_lint(cx, REDUNDANT_TYPE_ANNOTATIONS, local.span, "redundant type annotation");
+                    }
+                },
+                hir::ExprKind::MethodCall(_, _, _, _) => {
+                    if let hir::TyKind::Path(ty_path) = &ty.kind
+                        && let hir::QPath::Resolved(_, resolved_path_ty) = ty_path
+
+                        && let Some(func_ty) = func_hir_id_to_func_ty(cx, init.hir_id)
+                        && let Some(return_type) = func_ty_to_return_type(cx, func_ty)
+                        && is_same_type(cx, resolved_path_ty.res, return_type)
+                    {
+                        span_lint(cx, REDUNDANT_TYPE_ANNOTATIONS, local.span, "redundant type annotation");
+                    }
+                },
+                // When the initialization is a path for example u32::MAX
+                hir::ExprKind::Path(init_path) => {
+                    // TODO: check for non primty
+                    if let hir::TyKind::Path(ty_path) = &ty.kind
+                        && let hir::QPath::Resolved(_, resolved_path_ty) = ty_path
+                        && let hir::def::Res::PrimTy(primty) = resolved_path_ty.res
+
+                        && let hir::QPath::TypeRelative(init_ty, _) = init_path
+                        && let hir::TyKind::Path(init_ty_path) = &init_ty.kind
+                        && let hir::QPath::Resolved(_, resolved_init_ty_path) = init_ty_path
+                        && let hir::def::Res::PrimTy(primty_init) = resolved_init_ty_path.res
+
+                        && primty == primty_init
+                    {
+                        span_lint(cx, REDUNDANT_TYPE_ANNOTATIONS, local.span, "redundant type annotation");
+                    }
+                },
+                hir::ExprKind::Lit(init_lit) => {
+                    match init_lit.node {
+                        // In these cases the annotation is redundant
+                        LitKind::Str(..)
+                        | LitKind::ByteStr(..)
+                        | LitKind::Byte(..)
+                        | LitKind::Char(..)
+                        | LitKind::Bool(..)
+                        | LitKind::CStr(..) => {
+                            span_lint(cx, REDUNDANT_TYPE_ANNOTATIONS, local.span, "redundant type annotation");
+                        },
+                        LitKind::Int(..) | LitKind::Float(..) => {
+                            // If the initialization value is a suffixed literal we lint
+                            if init_lit.node.is_suffixed() {
+                                span_lint(cx, REDUNDANT_TYPE_ANNOTATIONS, local.span, "redundant type annotation");
+                            }
+                        },
+                        LitKind::Err => (),
+                    }
+                }
+                _ => ()
+            }
+        };
+    }
+}
--- a/clippy_lints/src/redundant_type_annotations.rs
+++ b/clippy_lints/src/redundant_type_annotations.rs
@@ -111,6 +111,17 @@ fn is_redundant_in_func_call<'tcx>(
     false
 }
 
+fn extract_primty<'tcx>(ty_kind: &hir::TyKind<'tcx>) -> Option<hir::PrimTy> {
+    if let hir::TyKind::Path(ty_path) = ty_kind
+        && let hir::QPath::Resolved(_, resolved_path_ty) = ty_path
+        && let hir::def::Res::PrimTy(primty) = resolved_path_ty.res
+    {
+        Some(primty)
+    } else {
+        None
+    }
+}
+
 impl LateLintPass<'_> for RedundantTypeAnnotations {
     fn check_local<'tcx>(&mut self, cx: &LateContext<'tcx>, local: &'tcx rustc_hir::Local<'tcx>) {
         // type annotation part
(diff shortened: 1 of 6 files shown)


tests/ui/redundant_type_annotations.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn new() -> Self {
+        Self { inner: 5 }
+    }
+}
+
+fn f_prim() -> u32 {
+    5
+}
+
+fn main() {
```
xFrednet, line 48: I would suggest a few more tests. It's fine if not all of them work right away, but I definitely want to prevent false positives (FPs)

```rs
#[derive(Debug, Default)]
struct Cake<T> {
    _data: T,
}

fn make_something<T: Default>() -> T {
    T::default()
}

fn make_cake<T: Default>() -> Cake<T> {
    Cake::<T>::default()
}

fn test_generics() {
    // The type annotation is needed to determine T
    let _c: Cake<i32> = make_something();

    // The type annotation is needed to determine the topic
    let _c: Cake<u8> = make_cake();
}

fn test_non_locals() {
    // This shouldn't lint
    fn _arg(x: u32) -> u32 {
        x
    }

    // This could lint, but probably shouldn't
    let _closure_arg = |x: u32| x;
}

fn test_complex_types() {
    // Shouldn't lint, since the literal will be i32 otherwise
    let _u8: u8 = 128;

    // Should lint
    let _tuple_i32: (i32, i32) = (12, 13);

    // Shouldn't lint, since the tuple will be i32 otherwise
    let _tuple_u32: (u32, u32) = (1, 2);

    // Should lint, since the type is determined by the init value
    let _tuple_u32: (u32, u32) = (3_u32, 4_u32);

    // Should lint
    let _array: [i32; 3] = [5, 6, 7];

    // Shouldn't lint
    let _array: [u32; 2] = [8, 9];
}
```

I also feel like it would be better, to split up the main method into functions, which specify if and why an example should be linted :)

tests/ui/redundant_type_annotations.rs
```diff
@@ -0,0 +1,67 @@
...
+    let a: String = String::new();
+
+    let st: ATest2 = ATest2::new();
+    let a: u32 = st.func(); // this should lint but doesn't (MethodCall)
+
+    let a: String = String::from("test");
+
+    let a: u32 = u32::MAX;
+
+    let a: u32 = ATest2::get_num(); // this should lint but doesn't (DefId refers to ATest2 and not to the func)
```
xFrednet, line 66: You can take a look at [`type_dependent_def`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/typeck_results/struct.TypeckResults.html#method.type_dependent_def) with this one. Maybe search in Clippy for examples how that is used to resolve the called function :)

clippy_lints/src/redundant_type_annotations.rs
```diff
@@ -0,0 +1,143 @@
...
+
+            false
+        },
+        hir::QPath::LangItem(..) => false,
+    }
+}
+
+impl LateLintPass<'_> for RedundantTypeAnnotations {
+    fn check_local<'tcx>(&mut self, cx: &LateContext<'tcx>, local: &'tcx rustc_hir::Local<'_>) {
+        if_chain! {
```
xFrednet, line 103: Here we should also include a check if the code comes from a macro. I believe it should be:

```rs
local.span.from_expansion()
```

or something close to that :)

xFrednet (commented):
Welcome to Clippy. Thank you for your patience while waiting on a review!

I've left some review comments. Please don't be discouraged by them, I believe this lint can become quite complex depending on how detailed and how many edge cases we want to catch. Please reach out if you require any further assistance. :)

clippy_lints/src/redundant_type_annotations.rs
```diff
@@ -0,0 +1,143 @@
...
+    /// let foo: String = String::new();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let foo = String::new();
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub REDUNDANT_TYPE_ANNOTATIONS,
+    pedantic,
+    "Warns about needless / redundant type annotations."
```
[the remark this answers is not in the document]
AlessioC31: Thanks!

clippy_lints/src/redundant_type_annotations.rs
```diff
@@ -0,0 +1,143 @@
...
+                if let Some(init_return_type) = middle_ty_init.fn_sig(cx.tcx).output().no_bound_vars();
+                then {
+                    return is_redundant_in_resolved_path(cx, res, init_return_type);
+                }
+            }
+
+            false
+        },
+        // let a: String = String::new();
+        hir::QPath::TypeRelative(func_hir_ty, _) => {
```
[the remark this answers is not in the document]
AlessioC31: Uhm in which cases do you think it could create I problem? I added your tests but I don't see anything weird happening, what am I missing?

tests/ui/redundant_type_annotations.rs
AlessioC31: Thanks, this was very useulf, I managed to solve this problem :D

clippy_lints/src/redundant_type_annotations.rs
AlessioC31: Done, thanks

AlessioC31:
> Welcome to Clippy. Thank you for your patience while waiting on a review!
> 
> I've left some review comments. Please don't be discouraged by them, I believe this lint can become quite complex depending on how detailed and how many edge cases we want to catch. Please reach out if you require any further assistance. :)

Thanks for the review!

clippy_lints/src/redundant_type_annotations.rs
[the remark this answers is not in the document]
xFrednet: `QPaths` are always a bit complicated. This comment was a question if you've considered the case and came to the conclusion that the second argument isn't needed, or if there was just no test case that needed that.

Leaving it blank should be fine, I can't think of a test example for this.

xFrednet:
> Thanks for the review!

You're welcome. Are you still working on the implementation or is the branch ready for another review? :)

AlessioC31:
> > Thanks for the review!
> 
> You're welcome. Are you still working on the implementation or is the branch ready for another review? :)

I'm still working on it. Right now I've problems with `Lit` type of expressions (like how to map a `Lit` to a `Ty`) :)
Do you have any suggestion on where I could have a look?

xFrednet:
> I'm still working on it. Right now I've problems with `Lit` type of expressions (like how to map a `Lit` to a `Ty`) :) Do you have any suggestion on where I could have a look?

I don't think there is an easy way to map it. You could request the type of the expression, but that will probably not help too much.

> Or when is a type annotation useful in case of a `Lit`?

A type annotation is useful or even required when the literal is unsuffixed. For example `8` can be several types, the compiler would then look at its usages to determine the type. If the user writes `let x: u8 = 8`, the type should be kept IMO. However, if the literal has a suffix `8_u8` then the type can be removed. You can take a look at [`LitKind::is_unsuffixed()`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/ast/enum.LitKind.html#method.is_unsuffixed) for this one. For `&str` etc it should be fine to remove the type as the literal is clear.

tests/ui/redundant_type_annotations.rs
```diff
@@ -0,0 +1,151 @@
...
+    let _c: Cake<i32> = make_something();
+
+    // The type annotation is needed to determine the topic
+    let _c: Cake<u8> = make_cake();
+
+    // This should lint (doesn't)
+    let _c: Cake<u8> = make_cake::<u8>();
+
+    // This should lint (doesn't)
+    let _c: u8 = make_something::<u8>();
```
xFrednet, line 79: These are false negatives, which are fine to have in my opinion. I'd suggest merging the lint as is, we can always improve it later. Though, the comment should make it a bit clearer then. Same for the cases below.

Maybe:

```suggestion
    // This could be lint, but currently isn't 
    let _c: Cake<u8> = make_cake::<u8>();

    // This could be lint, but currently isn't
    let _c: u8 = make_something::<u8>();
```

tests/ui/redundant_type_annotations.rs
```diff
@@ -0,0 +1,151 @@
...
+fn test_simple_types() {
+    let _var: u32 = u32::MAX;
+
+    // Should lint (doesn't)
+    let _var: u32 = 5_u32;
+}
+
+fn main() {}
+
+// TODO: test refs
```
xFrednet, line 151: Could you address this TODO :)

tests/ui/redundant_type_annotations.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    let _return: u32 = return_an_int();
+
+    let _return: String = String::new();
+
+    let new_pie: Pie = Pie::new();
+
+    let _return: u32 = new_pie.return_an_int();
+
+    let _return: String = String::from("test");
```
xFrednet, line 135: This currently doesn't trigger a lint, even if the function comment says it should. I'm not quite sure why, as `String::new()` etc are covered.

xFrednet (commented):
Generally speaking, I think this is a good start, which we could actually merge as a minimum viable product. There are still several examples that could be improved. However, most of them seem a bit complicated to fix and the question is how much benefit we'll get from them. It should be fine as is, since all ToDos are false negatives. 

What are your thoughts? Do you want to try perfecting it in this PR or get this merge ready and create follow-ups? (Assuming, you're still interested in the lint)

AlessioC31:
> Generally speaking, I think this is a good start, which we could actually merge as a minimum viable product. There are still several examples that could be improved. However, most of them seem a bit complicated to fix and the question is how much benefit we'll get from them. It should be fine as is, since all ToDos are false negatives.
> 
> What are your thoughts? Do you want to try perfecting it in this PR or get this merge ready and create follow-ups? (Assuming, you're still interested in the lint)

Hello @xFrednet , thanks again.

I'm still interested and I plan to perfecting this PR (at least adding support for primitive types and the suggestions you mentioned in the previous reviews).
I had to pause working on this because I had a busy period, it should be better next week!

xFrednet:
Sounds good, that's no problem at all, take your time! :)

tests/ui/redundant_type_annotations.rs
AlessioC31: Wow, I missed that, thanks!

I digged into it a bit, it turns out it is related to generics.
When I do `cx.tcx.type_of(func_defid)` of the function, I get `EarlyBinder(fn(T) -> Self {<Self as std::convert::From<T>>::from})`

If I then execute `subst_identity` or `skip_binder` on this, I get `fn(T) -> Self {<Self as std::convert::From<T>>::from}`

Then, when I do `func_ty.fn_sig(cx.tcx).output()` where `func_ty` is the type I wrote previously, I get `Binder(Self, [])`, then I interpret the return type as `Self` instead of `String`.

Do you know what I could use in this case? I tried having a look at other lints but I was not able to find anything, thanks!

AlessioC31: cc @xFrednet :D

xFrednet: Sorry, I missed your question earlier :sweat_smile:. I'm not totally sure about this one, but you might want to take a look at:
* [`type_dependent_def`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/typeck_results/struct.TypeckResults.html#method.type_dependent_def) to resolve the method target
* Maybe [qpath_res](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/typeck_results/struct.TypeckResults.html#method.qpath_res) to resolve `Self`?

If both of these don't work, I'd probably ask on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy) the others might have a good idea for this one :)

AlessioC31: > Sorry, I missed your question earlier 😅. I'm not totally sure about this one, but you might want to take a look at:
> 
> * [`type_dependent_def`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/typeck_results/struct.TypeckResults.html#method.type_dependent_def) to resolve the method target
> * Maybe [qpath_res](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/typeck_results/struct.TypeckResults.html#method.qpath_res) to resolve `Self`?
> 
> If both of these don't work, I'd probably ask on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy) the others might have a good idea for this one :)

I've tried the two methods you suggested without any luck. I've asked on Zulip, let's see.

Thanks!

AlessioC31:
> Hey @AlessioC31, what is the current status of the PR? Are there still open TODOs, besides the rebase? :)

Hello @xFrednet , there's still the problem with the `String::from` (and other generics since it's the same I guess). I've asked on Zulip but I haven't received any answer, maybe I can wrap this PR, I'll add support for refs, rebase and then it's ready to being merged. It won't lint in case generics are used but maybe I can come back at this when I understand how to do it

xFrednet:
> Hello @xFrednet, I think this is ready for review.
> 
> We miss a couple of things which I think they could be added in the future, like:
> 
>     * Support for generics

As long as we don't have false positives and only false negatives, that's totally fine :). Could you add this limitation, as well as the complex types and path one into the lint documentation? I like to use a `### Limitations` section to list such things. After this PR, we should ideally also create issues for these :)

>     * Refs returned from anything else than a `MethodCall` (maybe it's because of my little experience with rust yet, but I think this shouldn't happen often, what do you think?)

We should make sure that normal function calls are also detected. For one, because they're used quite often and because it should be a simple addition. I can mark the required changes in the next review :)

>     * Complex types (tuples, arrays, etc...)

Complex types are sadly complex to lint. I've been contemplating, how to best handle lints for these, for a different project. I sadly haven't found a good option. This is to say, that it's fine if these are not caught, probably. I guess that Clippy has quite a few lints that can't handle these :)

>     * `Path` to something different than a primitive type. Like for now this lint supports cases like `let a: u32 = u32::MAX` but it doesn't support things like `let a: MyStruct = Somewhere::MyStaticInstance`. Do you think we need to add this?

I believe that should be fine. Such a pattern is rarely used in the projects I'm using. We can leave this as an improvement for the future :)

> Thanks!

You're very welcome, thank you for working on Clippy! I'll review this PR in the upcoming week :)

clippy_lints/src/redundant_type_annotations.rs
```diff
@@ -0,0 +1,201 @@
...
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Warns about needless / redundant type annotations.
+    ///
+    /// ### Why is this bad?
+    /// Code without type annotations is shorter and in most cases
+    /// more idiomatic and easier to modify.
+    ///
+    /// ### Example
```
xFrednet, line 16: It would be good to have a `### Limitations` section, as suggested in this comment:

https://github.com/rust-lang/rust-clippy/pull/10570#issuecomment-1556238326

clippy_lints/src/redundant_type_annotations.rs
```diff
@@ -0,0 +1,201 @@
...
+        // type annotation part
+        if !local.span.from_expansion()
+            && let Some(ty) = &local.ty
+
+            // initialization part
+            && let Some(init) = local.init
+        {
+            match &init.kind {
+                // When the initialization is a call to a function
+                hir::ExprKind::Call(init_call, _) => {
```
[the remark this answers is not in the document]
AlessioC31: Yes, what I meant in the previous comment was that I don't think it will handle refs returned from function calls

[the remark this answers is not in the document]
xFrednet: Ahh okay, that's totally fine :)

clippy_lints/src/redundant_type_annotations.rs
AlessioC31: Done thanks!

AlessioC31:
I think I've addressed all your comments and rebased to `master`. Should I squash the commits?

Thanks!

xFrednet:
> I think I've addressed all your comments and rebased to `master`.

There still seems to be a conflict, could you resolve that one?

> Should I squash the commits?

It's not mandatory, but would be nice, 1 - 5 commits would be a good balance :)

I intend to review this over the weekend :)

AlessioC31:
> > I think I've addressed all your comments and rebased to `master`.
> 
> There still seems to be a conflict, could you resolve that one?
> 
> > Should I squash the commits?
> 
> It's not mandatory, but would be nice, 1 - 5 commits would be a good balance :)
> 
> I intend to review this over the weekend :)

Ok I've rebased to master and reduced the amount of commits, thanks!

AlessioC31:
Rebased again :D

xFrednet (approved):

PR #10573, opened 2023-03-30
Ignore `file!()` macro in `print_literal`, `write_literal`

### LINKED ISSUE #10544: False positive for `file!()` macro in println
### Summary

When using `file!()` as a macro inside of `println!`, it suggest to replace it.

The suggestion leads to invalid code.

The error was probably introduces in 6fc6d87fd0@Alexendoo 

As shown in the code below, other macro calls work fine, it seems to be a problem with the `file!()` macro in particular.

Maybe checking `&& value_string != "file!()"` would be a valid solution here?!
https://github.com/rust-lang/rust-clippy/blob/27edc2af066979ff7b3182b33c62a84d57b907c4/clippy_lints/src/write.rs#L463

### Lint Name

print_literal

### Reproducer

I tried this code:

```rust
macro_rules! static_str {
    () => { "Hello World" }
}

fn main() {
    println!("{}", file!());
    println!("{}", line!());
    println!("{}", static_str!());
    println!("{}", "Hello World");
}
```

I saw this happen:

```
warning: literal with an empty format string
 --> src/main.rs:6:20
  |
6 |     println!("{}", file!());
  |                    ^^^^^^^
  |
  = note: `#[warn(clippy::print_literal)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#print_literal
help: try this
  |
6 -     println!("{}", file!());
6 +     println!("ile!(");
  |

warning: literal with an empty format string
 --> src/main.rs:9:20
  |
9 |     println!("{}", "Hello World");
  |                    ^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#print_literal
help: try this
  |
9 -     println!("{}", "Hello World");
9 +     println!("Hello World");
  |

warning: `abc` (bin "abc") generated 2 warnings
```

Not issuing the first warning, but only the second

### Version

```text
rustc 1.66.0-nightly (f5193a9fc 2022-09-25)
binary: rustc
commit-hash: f5193a9fcc73dc09e41a90c5a2c97fc9acc16032
commit-date: 2022-09-25
release: 1.66.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

Alexendoo:
changelog: [`print_literal`], [`write_literal`]: Ignore the `file!()` macro

`file!()` expands to a string literal with its span set to that of the `file!()` callsite, but isn't marked as coming from an expansion. To fix this we make sure we actually find a string/char literal instead of assuming it's one and slicing

It would also ignore any other macros that result in the same situation, but that shouldn't be common as `proc_macro::Span::call_site()` returns a span that is marked as from expansion

Fixes #10544

### DIFF
--- a/clippy_lints/src/write.rs
+++ b/clippy_lints/src/write.rs
@@ -463,12 +463,18 @@ fn check_literal(cx: &LateContext<'_>, format_args: &FormatArgs, name: &str) {
             && let Some(value_string) = snippet_opt(cx, arg.expr.span)
     {
             let (replacement, replace_raw) = match lit.kind {
-                LitKind::Str | LitKind::StrRaw(_)  => extract_str_literal(&value_string),
+                LitKind::Str | LitKind::StrRaw(_)  => match extract_str_literal(&value_string) {
+                    Some(extracted) => extracted,
+                    None => return,
+                },
                 LitKind::Char => (
                     match lit.symbol.as_str() {
                         "\"" => "\\\"",
                         "\\'" => "'",
-                        _ => &value_string[1..value_string.len() - 1],
+                        _ => match value_string.strip_prefix('\'').and_then(|s| s.strip_suffix('\'')) {
+                            Some(stripped) => stripped,
+                            None => return,
+                        },
                     }
                     .to_string(),
                     false,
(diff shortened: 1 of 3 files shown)


Jarcho:
Looks good. @bors r+

PR #10576, opened 2023-03-31
Make dogfood tests run in `release` mode

blyxyas:
The old PR was about creating a new alias in `.cargo/config.toml`. After some extra testing we noticed that the speedup wasn't that critical and @flip1995 created a new patch to actually make `dogfood` tests run with a `--release` argument.

Ignore any conversation before [this comment](https://github.com/rust-lang/rust-clippy/pull/10576#issuecomment-1492002622), as the PR has been re-done.

changelog:none

### DIFF
--- a/clippy_dev/src/dogfood.rs
+++ b/clippy_dev/src/dogfood.rs
@@ -10,8 +10,14 @@ use std::{
 pub fn dogfood(fix: bool, allow_dirty: bool, allow_staged: bool) {
     let mut cmd = Command::new("cargo");
 
+    #[cfg(not(target_os = "windows"))]
     cmd.current_dir(clippy_project_root())
         .args(["test", "-r", "--test", "dogfood"])
+        .args(["--features", "internal"])
+        .args(["--", "dogfood_clippy"]);
+    #[cfg(target_os = "windows")]
+    cmd.current_dir(clippy_project_root())
+        .args(["test", "--test", "dogfood"])
         .args(["--features", "internal"])
         .args(["--", "dogfood_clippy"]);
 


flip1995:
How long does `dev-r` take from a clean state compared to just `dev`?

Alexendoo:
I wouldn't expect this to have such a large effect on the times since `cargo dev dogfood` is just handing off to a `cargo test` invocation, running

```bash
cargo dev dogfood
touch clippy_lints/src/lib.rs
time cargo dev dogfood

cargo dev-r dogfood
touch clippy_lints/src/lib.rs
time cargo dev-r dogfood
```

Gives me ~40s for both `time`s

flip1995:
Note: dogfood doesn't do anything when you run the command twice without changing anything. Maybe that's where the improvement is coming from.

I ran `cargo clean && time cargo dev dogfood` 3x:
1. No changes, all debug
    ```
    cargo dev dogfood  319.60s user 37.20s system 419% cpu 1:25.11 total
    ```
2. Your changes
    ```
    cargo dev dogfood  368.70s user 39.13s system 450% cpu 1:30.47 total
    ```
3. With the patch below 
    ```
    cargo dev dogfood  208.53s user 21.94s system 513% cpu 44.853 total
    ```

Patch:
```patch
diff --git a/clippy_dev/src/dogfood.rs b/clippy_dev/src/dogfood.rs
index b69e9f649ec..35653c76cc9 100644
--- a/clippy_dev/src/dogfood.rs
+++ b/clippy_dev/src/dogfood.rs
@@ -12,6 +12,10 @@ pub fn dogfood(fix: bool, allow_dirty: bool, allow_staged: bool) {
         .args(["--features", "internal"])
         .args(["--", "dogfood_clippy"]);

+    if !cfg!(debug_assertions) {
+        cmd.arg("--release");
+    }
+
     let mut dogfood_args = Vec::new();
     if fix {
         dogfood_args.push("--fix");
```

---

Another experiment: When I already ran `cargo build` to build Clippy and clippy_dev and then run `time cargo dev dogfood` (assuming the above patch was applied:
1. Just `dev`:
    ```
    cargo dev dogfood  301.85s user 34.24s system 396% cpu 1:24.70 total
    ```
2. With `dev-r`:
    ```
    cargo dev dogfood  180.30s user 18.08s system 491% cpu 40.341 total
    ```
3. With `dev` but the `--release` flag applied unconditionally:
    ```
    cargo dev dogfood  165.80s user 20.05s system 472% cpu 39.327 total
    ```

So I think the best thing to do here would be:

```patch
diff --git a/clippy_dev/src/dogfood.rs b/clippy_dev/src/dogfood.rs
index b69e9f649ec..484e0d4e88b 100644
--- a/clippy_dev/src/dogfood.rs
+++ b/clippy_dev/src/dogfood.rs
@@ -10,7 +10,8 @@ pub fn dogfood(fix: bool, allow_dirty: bool, allow_staged: bool) {
     cmd.current_dir(clippy_project_root())
         .args(["test", "--test", "dogfood"])
         .args(["--features", "internal"])
-        .args(["--", "dogfood_clippy"]);
+        .args(["--", "dogfood_clippy"])
+        .arg("--release");

     let mut dogfood_args = Vec::new();
     if fix {
```

and NOT introduce `dev-r`.

blyxyas:
I just tested again with a clean slate of a recently pulled `master`. It is ~2 seconds faster.
The first testing environment was in a mid-debugging branch, so dogfood was failing faster, that's why I found it so useful.

In a "stable" and after `cargo clean`, it looses all hopes to be useful. So I guess this PR is useless and should be closed.

flip1995 (approved):
I would approve, but since this is my own patch, I don't want to approve my own thing.

flip1995:
Please update the PR description+title before this gets merged though

Alexendoo:
That cuts the runtime down by not running it :P

```
% cargo test --test dogfood --features internal -- dogfood_clippy --release
    Finished test [unoptimized + debuginfo] target(s) in 0.07s
     Running tests/dogfood.rs (target/debug/deps/dogfood-bf1f6a65afcf8f1b)
error: Unrecognized option: 'release'
error: test failed, to rerun pass `--test dogfood`
```

That doesn't get shown by `cargo dev dogfood` though because it only prints the stdout, would be best to change that to inherit stdout + stderr rather than using `.output()` + `println`

Jarcho:
This is probably not worth it. Running dogfood in release mode would mean compiling clippy in release mode, which basically negates all the speed gains from running clippy on itself faster. Basically more than double the compilation time for release mode and dogfood now has to run in zero time.

It also means getting an ICE while running dogfood will be much less pleasant experience.

blyxyas:
> Which basically negates all the speed gains from running clippy on itself faster.

It seems like it doesn't (only compiling Clippy):

dev: `time cargo test --test dogfood --features internal -- dogfood_clippy`: 1:35 total
release: `time cargo test --test dogfood -r --features internal -- dogfood_clippy`: 1:10.50 total

---

On the topic of the unrecognized option, moving `--release` to before `--features` is declared seems to work.

```diff
- cargo test --test dogfood --features internal -- dogfood_clippy --release

+ cargo test --test dogfood --release --features internal -- dogfood_clippy
```

Jarcho:
Just tested on my machine:

`cargo test --test dogfood --features internal` with a single change to `clippy_lints` compiles in 0:20 and takes 1:20 total. \
`cargo test --test dogfood --release --features internal` with a single change to `clippy_lints` compiles in 1:25 and takes 1:58 total.

Both tests were don first by running the command, then changing `clippy_lints` by adding a newline to the start of `lib.rs`. I haven't been able to make debug mode slower overall even when making actual functional changes to the code either, though that does close the gap a bit. Any change to `clippy_utils` makes the gap significantly wider.

Jarcho:
More perf tests on windows:

For debug mode with incremental:

* `cargo clean && cargo build --all-targets --features internal`
* Add newline to start of `clippy_lints/src/lib.rs` (seems to cause full recompile)
* `measure-command { cargo test --test dogfood --features internal }`

Time: 3:13

* `cargo clean && cargo build --all-targets --features internal`
* Add newline to start of `clippy_lints/src/allow_attributes.rs`
* `measure-command { cargo test --test dogfood --features internal }`

Time: 2:25

For debug mode without incremental:

* `cargo clean && cargo build --all-targets --features internal`
* add newline to start of `clippy_lints/src/lib.rs`
* `measure-command { cargo test --test dogfood --features internal }`

Time: 2:41

For release mode:

* `cargo clean && cargo build --all-targets --features internal --release`
* add newline to start of `clippy_lints/src/lib.rs`
* `measure-command { cargo test --test dogfood --features internal --release }`

Time: 3:02

---

Release mode is only beating incremental debug mode when none of the incremental cache seems to get used.

blyxyas:
This differences maybe are caused by platform differences?

These are my times (Using the same commands):

| DEBUG                               | Blyxyas' times (Linux) | Jarcho's times (Windows) |
|-------------------------------------|------------------------|--------------------------|
| Change to `lib.rs`                  | 2:02                   | 3:13                     |
| Change to `lib.rs` (No incremental) | 2:03                   | 2:41                     |
| Change to `allow_attribute.rs`      | 1:34                   | 2:25                     |
|             **RELEASE**             |                        |                          |
| Change to `lib.rs` (Incremental)    | 1:40.93                | 3:02                     |

Difference from Linux `lib.rs` changes (incremental): -18.0%
Difference from Windows `lib.rs` changes (incremental): -5.6%

Difference from Linux `lib.rs` changes (No incremental): -18.7%
Difference from Windows `lib.rs` changes (No incremental): **+13%**

So this PR makes sense only for Linux users.

---

Note that, a "mid-debugging" scenario (adding a new line to `allow_attribute.rs`) does increase dogfood times (In my case, by **+4%**). I don't think this is significant, as the command for quickly testing (for debugging) a lint is `TESTNAME=... cargo uitest`, not `cargo test` or `cargo dev dogfood`

blyxyas:
If someone could verify the results on their machine. Maybe test them on Mac OS?

flip1995:
> That cuts the runtime down by not running it :P

That's not the outcome we want. I should have checked for that :sweat_smile: 

---

In that case, compiling Clippy in release mode and running dogfood in release mode vs both in debug mode took about the same time for me. So I don't think this improves things significantly. Also since it is really machine dependent as it seems, I don't think adding a new alias for it is worth it.

flip1995:
With this patch, same commands as https://github.com/rust-lang/rust-clippy/pull/10576#issuecomment-1494322693

| | `blyxyas:dev_release` | `master` | 
|--------|-------:|------:|
| Debug, clean | 1:18 | 1:19 |
| Debug, rerun | 0:32 | 0:25 |
| Release, clean | 1:22 | 1:23 |
| Release, rerun | 0:40 | 0:40 |

So I would conclude that this PR doesn't improve the situation.

blyxyas:
So Alexendoo's differences are:

| LINUX                         	|            	|
|-------------------------------	|------------	|
| Debug clean - Release clean   	| **+14%**   	|
| Debug re-run - Release re-run 	| **+47.5%** 	|
|          **WINDOWS**          	|            	|
| Debug clean - Release clean   	| -1.8%      	|
| Debug re-run - Release re-run 	| **+28.2%** 	|

So Alexendoo's machine also performes worse, even on Linux.
Seeing that this is extremely machine-dependant, and seeing that it'd take a lot of time to investigate the origin of these differences (number of cores maybe? Maybe a hidden architecture differences? Maybe brand differences between Intel and AMD?) and I'm not even sure conditional compilation could solve the issue, I think this PR should be close.

(Please react if you agree or disagree)

flip1995:
Thanks for investigating this!

PR #10578, opened 2023-03-31
Add `items_after_test_module` lint

blyxyas:
Resolves task *3* of #10506, alongside *1* resolved at #10543 in an effort to help standarize a little bit more testing modules.

---

changelog:[`items_after_test_module`]: Added the lint.

### DIFF
--- a/clippy_lints/src/items_after_test_module.rs
+++ b/clippy_lints/src/items_after_test_module.rs
@@ -1,0 +1,82 @@
+use clippy_utils::{diagnostics::span_lint_and_help, is_in_cfg_test};
+use rustc_hir::{HirId, ItemId, ItemKind, Mod};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Triggers if an item is declared after the testing module marked with `#[cfg(test)]`.
+    /// ### Why is this bad?
+    /// Having items declared after the testing module is confusing and may lead to bad test coverage.
+    /// ### Example
+    /// ```rust
+    /// #[cfg(test)]
+    /// mod tests {
+    ///     // [...]
+    /// }
+    ///
+    /// fn my_function() {
+    ///     // [...]
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn my_function() {
+    ///     // [...]
+    /// }
+    ///
+    /// #[cfg(test)]
+    /// mod tests {
+    ///     // [...]
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub ITEMS_AFTER_TEST_MODULE,
+    style,
+    "An item was found after the testing module `tests`"
+}
+
+declare_lint_pass!(ItemsAfterTestModule => [ITEMS_AFTER_TEST_MODULE]);
+
+impl LateLintPass<'_> for ItemsAfterTestModule {
+    fn check_mod(&mut self, cx: &LateContext<'_>, _: &Mod<'_>, _: HirId) {
+        let mut was_test_mod_visited = false;
+        let mut test_mod_hash: Option<u128> = None;
+
+        let hir = cx.tcx.hir();
+        let items = hir.items().collect::<Vec<ItemId>>();
+
+        for itid in &items {
+            let item = hir.item(*itid);
+
+            if_chain! {
+            if was_test_mod_visited;
+            if cx.sess().source_map().lookup_char_pos(item.span.lo()).file.name_hash
+            == test_mod_hash.unwrap(); // Will never fail
+            if !matches!(item.kind, ItemKind::Mod(_) | ItemKind::Macro(_, _));
+            if !is_in_cfg_test(cx.tcx, itid.hir_id()); // The item isn't in the testing module itself
+
+            if !in_external_macro(cx.sess(), item.span);
+            then {
+                span_lint_and_help(cx, ITEMS_AFTER_TEST_MODULE, item.span, "an item was found after the testing module", None, "move the item to before the testing module was defined");
+            }};
+
+            if matches!(item.kind, ItemKind::Mod(_)) {
+                for attr in cx.tcx.get_attrs(item.owner_id.to_def_id(), sym::cfg) {
+                    if_chain! {
+                                if attr.has_name(sym::cfg);
+                                if let Some(mitems) = attr.meta_item_list();
+                                if let [mitem] = &*mitems;
+                                if mitem.has_name(sym::test);
+                                then {
+                                    was_test_mod_visited = true;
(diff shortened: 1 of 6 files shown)


tests/ui/items_after_test_modules/items_after_test_module.fixed
```diff
@@ -0,0 +1,19 @@
...
+#![warn(clippy::items_after_test_module)]
+
+mod my_mod;
+
+fn main() {}
+
+fn should_not_lint() {}
+
+fn should_lint() {}
+
```
dswij, line 13: Think it's missing `#[cfg(test)]` here

blyxyas: Fixed

blyxyas:
Im thinking about removing suggestions.
The current commit fixes that specific test case, but I tried with putting an attribute over `should_not_lint`, but it completely breaks the program.

I could fix that case, but then if I put another item with an attribute over that one, it will break again.
I thought attributes were also

tests/ui/items_after_test_module.stderr
```diff
@@ -0,0 +1,19 @@
+error: an item was found after the testing module
```
dswij, line 1: Have we considered taking all of the Span between two, and after any testing module, and use that as the lint's Span?

This way the suggestion can be clean when there's many items, imo.

tests/ui/items_after_test_module.rs
```diff
@@ -0,0 +1,20 @@
...
+#[allow(clippy::allow_attributes)]
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn hi() {}
+}
+
+fn should_lint() {}
+
+const SHOULD_ALSO_LINT: usize = 1;
```
dswij, line 20: Can we also add a test case for macros here?

Macros are usually... tricky

clippy_lints/src/items_after_test_module.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[clippy::version = "1.70.0"]
+    pub ITEMS_AFTER_TEST_MODULE,
+    style,
+    "An item was found after the testing module `tests`"
+}
+
+declare_lint_pass!(ItemsAfterTestModule => [ITEMS_AFTER_TEST_MODULE]);
+
+impl LateLintPass<'_> for ItemsAfterTestModule {
+    fn check_mod(&mut self, cx: &LateContext<'_>, _: &Mod<'_>, _: HirId) {
```
dswij, line 44: Some style nit/suggestion dump, feel free to ignore 😅 

Maybe we can take out the `if_chain` check, then do a `skip_while` iterator on `items` iter.

dswij:
> Im thinking about removing suggestions. The current commit fixes that specific test case, but I tried with putting an attribute over `should_not_lint`, but it completely breaks the program.

Yeah, that sounds a bit convoluted. I'd recommend leaving the suggestion for now, and maybe come back to it in the future

tests/ui/items_after_test_module.rs
blyxyas: Added

tests/ui/items_after_test_module.stderr
blyxyas: The suggestion isn't there anymore

dswij: Sorry, I think I'm not being clear.

What I'm trying to say is, rather than linting the items one-by-one (which might cause a blow up in the number of clippy warnings/errors in some scenarios), we could emit the warnings in one warning instead.

Also, I think we'd want to lint the macros too. Might be missing something obvious here, but why are we ignoring macros?

dswij: Sorry, I think I'm not clear enough here.

Rather than linting the items one-by-one (which might cause a blow up in the number of clippy errors/warnings), we can instead collect all the items after the misplaced test module, and lint them as one.

Also, I might be missing something obvious here, but why are we ignoring macros?

blyxyas: Ok, I'm pretty dumb.
The macros were being ignored because I thought that the user couldn't move them, as the place of macro declaration matters a lot to the usage of it. But thinking a little bit more, it doesn't the matter, as everything after the testing module should be moved in pack.

I'm going to fix this now.

blyxyas: Fixed, for real now

dswij: Not a bit of worry. Thanks for incorporating the suggestion above :)

tests/ui/items_after_test_module.stderr
```diff
@@ -0,0 +1,17 @@
...
+  --> $DIR/items_after_test_module.rs:13:1
+   |
+LL | / mod tests {
+LL | |     #[test]
+LL | |     fn hi() {}
+LL | | }
+...  |
+LL | |     () => {};
+LL | | }
+   | |_^
```
dswij, line 11: Just a question: is this the intended span for the lint?

blyxyas: Yes it is.
Would you want it to start after the test module?

dswij: This is fine. I was thinking that it might start a line earlier (from `#[cfg(test)]`) but this'd be fine too.

dswij (commented):
@blyxyas Sorry for the rather slow review.

Just a question, and should be good to merge afterwards :)

clippy_lints/src/items_after_test_module.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        let hir = cx.tcx.hir();
+        let items = hir.items().collect::<Vec<ItemId>>();
+
+        for (i, itid) in items.iter().enumerate() {
+            let item = hir.item(*itid);
+
+            if_chain! {
+            if was_test_mod_visited;
+            if i == (items.len() - 3 /* Weird magic number (HIR-translation behaviour) */);
```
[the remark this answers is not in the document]
dswij: We can use https://doc.rust-lang.org/std/primitive.usize.html#method.checked_sub if we can't guarantee that this won't cause an overflow.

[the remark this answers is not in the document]
blyxyas: Just tested it with this code (The minimum amount of items that meats the requirements for the lint):

```
#[cfg(test)]
mod test {}

const U: usize = 0;
```

In an independent crate using this command: `cargo dev lint ../<test dir>/ -- -- -W clippy::items_after_test_module --test`
Even though the lint, well, lints. It doesn't overflow.

dswij:
Thank you! @bors r+

dswij:
@blyxyas Whoops, CI failed there. Can you help to take a look?

blyxyas:
I don't know how to fix this, it work on my machine and the normal Github CI test 'base'. Maybe retrying?

blyxyas:
@dswij Could you retry the bors tests?

blyxyas:
??? It works on my machine, with a modern 64-bit Ubuntu-based OS, like the CI. I guess I'm going to ask on Zulip, because I can't replicate this CI

blyxyas:
Just asked on Zulip, it's fixed now yay

blyxyas:
It's fixed now, the bug was caused because the test didn't change to the `//@` syntax for headers.

PR #10584, opened 2023-04-01
fix: `wildcard_imports` ignore `test.rs` files

### LINKED ISSUE #10580: `clippy::wildcard_imports` triggers for `use super::super::*;` under a module in a `tests.rs`
### Summary

If you use modules to sort your tests under a `tests.rs` file, Clippy can trigger a false positive if you use `use super::super::*;`. This even includes if you're using it to import items from within the `tests.rs` file.

### Lint Name

`clippy::wildcard_imports`

### Reproducer

I tried this code:

<details><summary><code>/src/lib.rs</code></summary>

```rust
#![warn(clippy::wildcard_imports)]

#[cfg(test)]
mod tests;

#[must_use]
pub fn add(left: usize, right: usize) -> usize {
	left + right
}
```
</details>
<details><summary><code>/src/tests.rs</code></summary>

```rust
mod add {
	use super::super::*;
	
	#[test]
	fn it_works() {
		let result = add(2, 2);
		assert_eq!(result, 4);
	}
}
```
</details>

I saw this happen (make sure you execute `cargo clippy --tests` to compile the tests):

```
warning: usage of wildcard import
 --> src\tests.rs:2:6
  |
2 |     use super::super::*;
  |         ^^^^^^^^^^^^^^^ help: try: `super::super::add`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wildcard_imports
 --> src\lib.rs:1:9
  |
1 | #![warn(clippy::wildcard_imports)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^
```

(No warning)

### Version

```text
rustc 1.68.2 (9eb3afe9e 2023-03-27)
binary: rustc
commit-hash: 9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0
commit-date: 2023-03-27
host: x86_64-pc-windows-msvc
release: 1.68.2
LLVM version: 15.0.6
```

### Additional Labels

@rustbot label +L-pedantic

blyxyas:
Adds a check to see if the building crate is a test one, if so, ignore it

---

Closes #10580 
changelog:[`wildcard_imports`]: Add a check to ignore files named `test.rs` and `tests.rs`


clippy_lints/src/wildcard_imports.rs
```diff
@@ -102,21 +103,40 @@ declare_clippy_lint! {
...
             test_modules_deep: 0,
+            ignore: false,
         }
     }
 }
 
 impl_lint_pass!(WildcardImports => [ENUM_GLOB_USE, WILDCARD_IMPORTS]);
 
 impl LateLintPass<'_> for WildcardImports {
+    fn check_mod(&mut self, cx: &LateContext<'_>, module: &Mod<'_>, _: HirId) {
```
flip1995, line 122: This needs a `check_mod_post`, that sets the `self.ignore` to `false` again. Otherwise, once checking one module that ends with `test.rs`, you'll never lint again in any other module. 

Aaaand I just noticed that there is no `check_mod_post`. Hmmm... 

This should also only be done on a file level. But a file is implicitly either a module itself AFAIK or `lib.rs`/`main.rs`. So checking the mod name should be enough.

flip1995 (commented):
I have doubts, if checking the file name is actually necessary.

clippy_lints/src/wildcard_imports.rs
flip1995: You probably want to do all that with `check_item` and `check_item_post`. You should be able to get the `Mod` name through the `Item` `ident`.

blyxyas: Checking the mod name doesn't work for files because files are not items (sadly).
It seems like the only way to check that the file name isn't `test.rs` or `tests.rs` is to directly check the filename. As there are no "module_name" function.

blyxyas: I can't seem to get a way to get the crate name from `check_item` (or even `check_crate`), I've been trying things for 5 or 6 hours now.
Would checking for `test` in any part of the path fix this issue?

blyxyas:
Oh god I finally got it
Someone un Zulip mentioned those `(lib)`, `(bin)`, etc... labels when building a crate, so I thought that, if the building process knows what it's building, then the crate type must be stored somewhere. That's when I found that a [this function](https://github.com/rust-lang/rust/blob/4cb92cc83ac6b24d83f8a30763b276cadb8dc41d/compiler/rustc_driver_impl/src/lib.rs#L172-L179) in `rust/rust_driver_impl` uses an struct called [`Options`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_session/config/struct.Options.html) from [`rustc_session::config`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_session/config/index.html). And `Options` has this very hand field `test: bool`. And you can access `Options` from `Session`, and `Session` from `LateContext`.

The illuminance came before me when I realize that now, I could check if a crate is a test. It's so beautiful.
I think we **definitely** need to add a function like `is_test_crate(cx)` to `clippy_utils` or something.

I've been at this like 8 hours now, and now... It's fixed... 😌 Peace is within me.

blyxyas:
Is there something left for merging this?

flip1995:
> Is there something left for merging this?

Just me being slow. Sorry :|

@bors r+

blyxyas:
?

flip1995:
Ah thanks for investigating, I planned to do that this evening.

@bors r+

PR #10586, opened 2023-04-02
Add `safe_unchecked` lint

### LINKED ISSUE #9438: suspicious lint: functions ending in _unchecked that are not marked as unsafe
### What it does

check function names that end _unchecked and then squark if they are not marked as unsafe.

Maybe I'm totally out of line here but I think if you're saying 'trust me' it should be marked unsafe so you can say why it's safe to use the unchecked version.

### Lint Name

safe_unchecked

### Category

correctness, suspicious

### Advantage

The suggestion would be to add unsafe to the function call. That could cause knock on compilation failures but they would have to be for the dev to fix manually as we can't say why they think it's safe.

If we have codebases where unsafe is swept under the carpet then all of rust is built on shaky ground.

### Drawbacks

False positives?

### Example

```rust
pub fn byte_to_str_unchecked(bytes: &[u8]) -> &str {
	unsafe {
		#[allow(clippy::transmute_bytes_to_str)]
		mem::transmute(bytes)
	}
}
```

Could be written as:
```rust
pub unsafe fn byte_to_str_unchecked(bytes: &[u8]) -> &str {
	unsafe {
		#[allow(clippy::transmute_bytes_to_str)]
		mem::transmute(bytes)
	}
}
```

blyxyas:
Adds the lint `safe_unchecked`, that triggers when a function which name ends with "_unchecked" isn't marked as unsafe

Resolves #9438 
changelog:[`safe_unchecked`]: The lint was added

### DIFF
--- a/clippy_lints/src/functions/mod.rs
+++ b/clippy_lints/src/functions/mod.rs
@@ -3,6 +3,7 @@ mod misnamed_getters;
 mod must_use;
 mod not_unsafe_ptr_arg_deref;
 mod result;
+mod safe_unchecked;
 mod too_many_arguments;
 mod too_many_lines;
 
@@ -354,6 +355,35 @@ declare_clippy_lint! {
     "`impl Trait` is used in the function's parameters"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Triggers when a function with a name ending with `_unchecked` is not marked as unsafe.
+    ///
+    /// ### Why is this bad?
+    /// `_unchecked` functions that have safety requirements should be marked `unsafe`.
+    ///
+    /// ### Known problems
+    /// An `_unchecked` function can be completely safe, with no conditions to maintain memory safety. In those cases they should not be marked `unsafe`.
+    ///
+    /// ### Example
+    /// ```rust
+    /// fn do_foo_unchecked() {
+    ///     // [...]
+    /// }
+    /// ```
+    ///
+    /// Use instead:
+    /// ```rust
+    /// unsafe fn do_foo_unchecked() {
+    ///     // [...]
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub SAFE_UNCHECKED,
+    restriction,
+    "a function ending in `_unchecked` that isn't marked `unsafe`"
+}
+
 #[derive(Copy, Clone)]
 pub struct Functions {
     too_many_arguments_threshold: u64,
@@ -382,6 +412,7 @@ impl_lint_pass!(Functions => [
     RESULT_LARGE_ERR,
     MISNAMED_GETTERS,
     IMPL_TRAIT_IN_PARAMS,
+    SAFE_UNCHECKED,
 ]);
 
 impl<'tcx> LateLintPass<'tcx> for Functions {
@@ -400,6 +431,7 @@ impl<'tcx> LateLintPass<'tcx> for Functions {
         not_unsafe_ptr_arg_deref::check_fn(cx, kind, decl, body, def_id);
         misnamed_getters::check_fn(cx, kind, decl, body, span);
         impl_trait_in_params::check_fn(cx, &kind, body, hir_id);
+        safe_unchecked::check_fn(cx, &kind, span, hir_id);
     }
 
     fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx hir::Item<'_>) {
--- a/clippy_lints/src/functions/safe_unchecked.rs
+++ b/clippy_lints/src/functions/safe_unchecked.rs
@@ -1,0 +1,42 @@
+use clippy_utils::{diagnostics::span_lint_and_then, is_trait_impl_item};
+use rustc_errors::Applicability;
+use rustc_hir::{intravisit::FnKind, HirId};
+use rustc_lint::LateContext;
+use rustc_span::{symbol::Ident, Span};
+
+use super::SAFE_UNCHECKED;
+
+pub(super) fn check_fn(cx: &LateContext<'_>, kind: &FnKind<'_>, sp: Span, hir_id: HirId) {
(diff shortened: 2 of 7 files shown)


clippy_lints/src/functions/mod.rs
```diff
@@ -354,6 +355,29 @@ declare_clippy_lint! {
     "`impl Trait` is used in the function's parameters"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Triggers when a function with a name ending with `_unchecked` is not marked as unsafe
+    /// ### Why is this bad?
+    /// `_unchecked` functions do not check some invariants which could result in business logic, memory safety or undefined behaviour if the user does not guarantee them.
```
Alexendoo, line 362: Many `_unchecked` functions do not have safety requirements, as logic errors are [not considered unsafe](https://doc.rust-lang.org/reference/behavior-not-considered-unsafe.html#logic-errors) I don't think we should be recommending using `unsafe` to denote that

blyxyas: So maybe "`_unchecked` functions may result in memory safety or undefined behaviour, so marking them with unsafe is a way to alert of their uncertainty"?

Im not good at lint descriptions.

clippy_lints/src/functions/mod.rs
```diff
@@ -354,6 +355,29 @@ declare_clippy_lint! {
     "`impl Trait` is used in the function's parameters"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Triggers when a function with a name ending with `_unchecked` is not marked as unsafe.
+    /// ### Why is this bad?
+    /// `_unchecked` functions may result in memory safety or undefined behaviour, so marking them with unsafe is a way to alert of their uncertainty.
```
Alexendoo, line 362:
```suggestion
    /// ### What it does
    /// Triggers when a function with a name ending with `_unchecked` is not marked as unsafe.
    ///
    /// ### Why is this bad?
    /// `_unchecked` functions that have safety requirements should be marked `unsafe`.
    ///
    /// ### Known problems
    /// An `_unchecked` function can be completely safe, with no conditions to maintain memory safety. In those cases they should not be marked `unsafe`.
```

Plus line breaks as needed

clippy_lints/src/functions/mod.rs
```diff
@@ -354,6 +355,29 @@ declare_clippy_lint! {
...
+    /// Use instead:
+    /// ```rust
+    /// unsafe fn do_foo_unchecked() {
+    ///     // [...]
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub SAFE_UNCHECKED,
+    restriction,
+    "the function `do_foo_unchecked` wasn't marked with 'unsafe'"
```
Alexendoo, line 378:
```suggestion
    "a function ending in `_unchecked` that isn't marked `unsafe`"
```

blyxyas: Fixed

clippy_lints/src/functions/mod.rs
blyxyas: Fixed

clippy_lints/src/functions/safe_unchecked.rs
```diff
@@ -0,0 +1,70 @@
...
+                    |diag| {
+                        diag.span_suggestion_verbose(
+                            sp.shrink_to_lo().with_lo(sp.lo() - BytePos(1)),
+                            "add an unsafe marker",
+                            "unsafe ".to_owned(),
+                            Applicability::MachineApplicable,
+                        );
+                    },
+                );
+            }
```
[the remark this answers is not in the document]
blyxyas: Fixed

clippy_lints/src/functions/safe_unchecked.rs
```diff
@@ -0,0 +1,70 @@
...
+                    cx,
+                    SAFE_UNCHECKED,
+                    ident.span,
+                    "this function ending with '_unchecked' isn't marked with 'unsafe'",
+                    |diag| {
+                        diag.span_suggestion_verbose(
+                            sp.shrink_to_lo().with_lo(sp.lo() - BytePos(1)),
+                            "add an unsafe marker",
+                            "unsafe ".to_owned(),
+                            Applicability::MachineApplicable,
```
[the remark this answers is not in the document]
blyxyas: Fixed

clippy_lints/src/functions/safe_unchecked.rs
```diff
@@ -0,0 +1,70 @@
...
+        FnKind::ItemFn(ident, _, header) => {
+            if !header.is_unsafe() && ident.as_str().ends_with("_unchecked") {
+                span_lint_and_then(
+                    cx,
+                    SAFE_UNCHECKED,
+                    ident.span,
+                    "this function ending with '_unchecked' isn't marked with 'unsafe'",
+                    |diag| {
+                        diag.span_suggestion_verbose(
+                            sp.shrink_to_lo().with_lo(sp.lo() - BytePos(1)),
```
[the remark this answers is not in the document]
blyxyas: It turns out it was doing nothing

tests/ui/safe_unchecked.rs
```diff
@@ -0,0 +1,21 @@
...
+
+trait SomeTrait {
+    fn method_should_not_lint();
+    fn method_should_lint_unchecked();
+    unsafe fn method_should_not_lint_unchecked();
+}
+
+impl SomeTrait for usize {
+    fn method_should_not_lint() {}
+    fn method_should_lint_unchecked() {}
```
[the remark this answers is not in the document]
blyxyas: Fixed

blyxyas:
As it seems this lint is very "controversial" (it was from the very start, even in the original issue) I think I will close the PR.
I also kinda agree that it's going too far. Not a lot of people have this need and it could be fixed by a regex search in a pre-push script.

PR #10591, opened 2023-04-03
Show multiple clippy.toml warnings with `sess.warn` instead of `eprintln!`

beetrees:
Use `sess.warn` to display the multiple clippy.toml warning instead of `eprintln!`.

changelog: none


flip1995:
Uh, nice!

@bors r+

PR #10592, opened 2023-04-03
Fix bug with getting parent directories in `lookup_conf_file`

beetrees:
Currently `lookup_conf_file` doesn't canonicalize the configuration directory before using [`PathBuf::pop`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.pop) to get the parent directory. This isn't usually an issue when clippy is invoked via `cargo clippy` as `CARGO_MANIFEST_DIR` is already canonicalized. However, this currently causes `clippy-driver` to ignore any `clippy.toml` in any parent directories when  `CARGO_MANIFEST_DIR` and `CLIPPY_CONF_DIR` are not set.

changelog: Fix a bug that would cause parent directories not to be searched for `clippy.toml` when using `clippy-driver` directly.


giraffate (approved):

giraffate:
@bors r+

It looks good, thanks!

PR #10593, opened 2023-04-04
fix(needless_return): do not trigger on ambiguous match arms return

### LINKED ISSUE #10546: Bug Report: Clippy `needless_return` incorrectly gets triggered
When using a match statement at the end of a function that has a return statement in a match arm and is followed by a semicolon, `cargo clippy` will suggest replacing the `return` with a unit value. However, it does this regardless of what type the match statement actually returns, so if it is something other than unit, it breaks.

Minimal reproducible example:

```rs
fn main() {
    match 0 {
        0 => 0,
        _ => return,
    };
}
```

`cargo clippy --fix`:

```
    Checking test v0.1.0 (/root/test)
warning: failed to automatically apply fixes suggested by rustc to crate `test`

  * src/main.rs

error[E0308]: `match` arms have incompatible types
 --> src/main.rs:4:14
  |
2 | /     match 0 {
3 | |         0 => 0,
  | |              - this is found to be of type `{integer}`
4 | |         _ => (),
  | |              ^^ expected integer, found `()`
5 | |     }
  | |_____- `match` arms have incompatible types

 --> src/main.rs:4:14
  |
4 |         _ => return,
  |              ^^^^^^
  |
  = help: replace `return` with a unit value

warning: `test` (bin "test") generated 1 warning (run `cargo clippy --fix --bin "test"` to apply 1 suggestion)
warning: `test` (bin "test" test) generated 1 warning (1 duplicate)
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
```

`rustc --version --verbose`:

```
rustc 1.70.0-nightly (8be3c2bda 2023-03-24)
binary: rustc
commit-hash: 8be3c2bda6b683f87b24714ba595e8b04faef54c
commit-date: 2023-03-24
release: 1.70.0-nightly
LLVM version: 15.0.7
```

feniljain:
If we see a case where match returns something other than `()`, we just skip `needless_return` lint in that case

Should fix #10546

Relevant Zulip Discussion: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Issue.20.2310546

---

changelog: FP: [`needless_return`]: No longer lints match statements with incompatible branches
[#10593](https://github.com/rust-lang/rust-clippy/pull/10593)

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -9,7 +9,7 @@ use rustc_hir::intravisit::FnKind;
 use rustc_hir::{Block, Body, Expr, ExprKind, FnDecl, LangItem, MatchSource, PatKind, QPath, StmtKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
-use rustc_middle::ty::subst::GenericArgKind;
+use rustc_middle::ty::{self, subst::GenericArgKind, Ty};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::def_id::LocalDefId;
 use rustc_span::source_map::Span;
@@ -175,7 +175,7 @@ impl<'tcx> LateLintPass<'tcx> for Return {
                 } else {
                     RetReplacement::Empty
                 };
-                check_final_expr(cx, body.value, vec![], replacement);
+                check_final_expr(cx, body.value, vec![], replacement, None);
             },
             FnKind::ItemFn(..) | FnKind::Method(..) => {
                 check_block_return(cx, &body.value.kind, sp, vec![]);
@@ -188,11 +188,11 @@ impl<'tcx> LateLintPass<'tcx> for Return {
 fn check_block_return<'tcx>(cx: &LateContext<'tcx>, expr_kind: &ExprKind<'tcx>, sp: Span, mut semi_spans: Vec<Span>) {
     if let ExprKind::Block(block, _) = expr_kind {
         if let Some(block_expr) = block.expr {
-            check_final_expr(cx, block_expr, semi_spans, RetReplacement::Empty);
+            check_final_expr(cx, block_expr, semi_spans, RetReplacement::Empty, None);
         } else if let Some(stmt) = block.stmts.iter().last() {
             match stmt.kind {
                 StmtKind::Expr(expr) => {
-                    check_final_expr(cx, expr, semi_spans, RetReplacement::Empty);
+                    check_final_expr(cx, expr, semi_spans, RetReplacement::Empty, None);
                 },
                 StmtKind::Semi(semi_expr) => {
                     // Remove ending semicolons and any whitespace ' ' in between.
@@ -202,7 +202,7 @@ fn check_block_return<'tcx>(cx: &LateContext<'tcx>, expr_kind: &ExprKind<'tcx>, 
                             span_find_starting_semi(cx.sess().source_map(), semi_span.with_hi(sp.hi()));
                         semi_spans.push(semi_span_to_remove);
                     }
-                    check_final_expr(cx, semi_expr, semi_spans, RetReplacement::Empty);
+                    check_final_expr(cx, semi_expr, semi_spans, RetReplacement::Empty, None);
                 },
                 _ => (),
             }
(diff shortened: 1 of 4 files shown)


tests/ui/needless_return.stderr
```diff
@@ -136,10 +136,10 @@ LL | |             return;
    = help: remove `return`
 
 error: unneeded `return` statement
-  --> $DIR/needless_return.rs:98:14
+  --> $DIR/needless_return.rs:98:13
    |
 LL |         _ => return,
-   |              ^^^^^^
```
xFrednet, line 142: I'm a little bit confused, that the changes above seem to have changed the span of the lint. Do you know where this comes from?

I'm guessing that these cases were marked as `RetReplacement::Empty` before while they're now `RetReplacement::Unit`. That is totally fine, but it would be good if the span stays the same, or the suggestion at least also includes a leading space :)

tests/ui/needless_return.fixed
```diff
@@ -81,7 +81,7 @@ fn test_void_if_fun(b: bool) {
 fn test_void_match(x: u32) {
```
xFrednet, line 81: I feel like none of these tests cover the actual case of the issue. Could you add this test:

```rs
fn test_match_as_stmt() {
    let x = 9;
    match x {
        1 => 2,
        2 => return,
        _ => 0,
    };
    // The semicolon makes it a statement, and allows the branches to be non-unit
}
```

clippy_lints/src/returns.rs
```diff
@@ -268,8 +284,9 @@ fn check_final_expr<'tcx>(
         // note, if without else is going to be a type checking error anyways
         // (except for unit type functions) so we don't match it
         ExprKind::Match(_, arms, MatchSource::Normal) => {
+            let match_ty = cx.typeck_results().expr_ty(peeled_drop_expr);
             for arm in arms.iter() {
-                check_final_expr(cx, arm.body, semi_spans.clone(), RetReplacement::Unit);
+                check_final_expr(cx, arm.body, semi_spans.clone(), RetReplacement::Empty, Some(match_ty));
```
xFrednet, line 289: I might be mistaken, but I feel like this is not quite the right fix for the issue. The problem, as I understand it, is that the `return` value has a different type than the other match arms. From my understanding, this can only happen, if the match is at the end of the function and has a semicolon, making it a statement. This therefore also implies that the function returns the unit type `()`, as otherwise there would be something after the statement or the types would be compatible.

```rs
fn test_match_as_stmt() {
    let x = 9;
    match x {
        1 => 2,
        _ => return,
    }; // The semicolon makes it a statement, and allows the branches to be non-unit
    // This makes the branches incompatible
}

fn does_not_compile() {
    let x = 9;
    match x {
        1 => 2,
        _ => return,
    }
    // The first branch is incompatible with the function signature.
}

fn does_compile_because_types_are_compatible() -> u32 {
    let x = 9;
    match x {
        1 => 2,
        _ => return 3,
    }
    // Should be linted since the types are compatible
}
```

(Maybe there are some other edge cases with `impl Trait`, that might be worth testing)

Based on this, I think the proper fix would be to check the type of the match statement. If it's `()` the types are compatible. If not and the `return` has a value than they would still be compatible. It would only cause problems, if the `match` type is `!= ()` while the function is `-> ()`.

Does this make sense?

xFrednet (commented):
Thank you for the PR :)

tests/ui/needless_return.stderr
feniljain: Yeah I also was confused with that and even asked on zulip what maybe the possible reason for it 🤔

clippy_lints/src/returns.rs
feniljain: > I think the proper fix would be to check the type of the match statement
 
 That's what I am doing currently in the function `check_final_expr` here:
 https://github.com/feniljain/rust-clippy/blob/c12748fab3a14beae4958f13551e7e6c52298490/clippy_lints/src/returns.rs#L250-L260
 
 😅 
 
 > If not and the return has a value than they would still be compatible.
 
 We would not reach this ( one I linked just above ) if the return is not empty, in which case we just tell clippy to ignore it, at least that's what we discussed in zulip

tests/ui/needless_return.fixed
feniljain: Oops messed up a commit somewhere while git worktree shenanigans, updated the PR with your test, thanks for this! 🙇🏻

tests/ui/needless_return.stderr
feniljain: Fixed this by changing the default we send xD

xFrednet (approved):

xFrednet:
Everything looks good to me, thank you for the PR :)

@bors r+

kadiwa4:
Sadly, this fix is not general enough to also solve #10390. @feniljain would you like to give that a go as well?

Issue #9824, opened 2022-11-09
False negative on mem_replace_option_with_none


and-reas-se:
### Summary

I was going trough the [Learning Rust with entirely too many linked lists](https://rust-unofficial.github.io/too-many-lists/) tutorial. I have my editor set to give me warnings from clippy. In section 3.1 the tutorial has you replacing a type with an Option, and then changing a couple of of mem::replace instances with the take() method on Option. I thought to myself "huh, that's always better, that should be a lint". So I googled, and turns out it is a lint, but for some reason the lint doesn't trigger on the code in the tutorial.

### Lint Name

mem_replace_option_with_none

### Reproducer

This is the code from the tutorial, but with the irrelevant parts cut out:

```rust
use std::mem;

pub struct List {
    head: Option<Box<Node>>,
}

struct Node {
    #[allow(dead_code)] // to get rid of that warning since I cut the code that read from it
    next: Option<Box<Node>>,
}

impl List {
    pub fn push(&mut self) {
        let new_node = Box::new(Node {
            next: mem::replace(&mut self.head, None),
        });

        self.head = Some(new_node);
    }
}
```
To reproduce create a new with cargo library crate, paste into lib.rs, and run cargo clippy. 
The lint is not triggered for the mem::replace on line 15.

### Version

```text
rustc 1.65.0 (897e37553 2022-11-02)
binary: rustc
commit-hash: 897e37553bba8b42751c67658967889d11ecd120
commit-date: 2022-11-02
host: x86_64-unknown-linux-gnu
release: 1.65.0
LLVM version: 15.0.0
```

J-ZhengLi:
this can further simplified as:

```rust
struct T(Option<u8>);

fn main() {
    let mut t = T(Some(1));
    let test = std::mem::replace(&mut t.0, None);
    println!("{:?}", test);
}
```

seems like the original lint ignores member variables like `t.0`, `self.head` for some reason.

PR #10594, opened 2023-04-04
fix [`mem_replace_option_with_none`] not considering field variables

J-ZhengLi:
fixes: #9824 

---

changelog: fix [`mem_replace_option_with_none`] not considering field variables

### DIFF
--- a/clippy_lints/src/mem_replace.rs
+++ b/clippy_lints/src/mem_replace.rs
@@ -1,12 +1,13 @@
 use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_sugg, span_lint_and_then};
 use clippy_utils::msrvs::{self, Msrv};
 use clippy_utils::source::{snippet, snippet_with_applicability};
+use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_non_aggregate_primitive_type;
-use clippy_utils::{is_default_equivalent, is_res_lang_ctor, path_res};
+use clippy_utils::{is_default_equivalent, is_res_lang_ctor, path_res, peel_ref_operators};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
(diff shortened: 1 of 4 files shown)


clippy_lints/src/mem_replace.rs
```diff
@@ -101,40 +101,26 @@ declare_clippy_lint! {
...
+    let mut applicability = Applicability::MachineApplicable;
+    span_lint_and_sugg(
+        cx,
+        MEM_REPLACE_OPTION_WITH_NONE,
+        expr_span,
+        "replacing an `Option` with `None`",
+        "consider `Option::take()` instead",
+        format!(
+            "{}.take()",
+            snippet_with_applicability(cx, sugg_expr_span, "", &mut applicability)
```
Jarcho, line 120: Should be `Sugg::hir_with_context(peel_ref_operators(cx, dest), expr_span.ctxt(), "", &mut applicability).mabe_par()` to handle precedence issues and macro calls.

J-ZhengLi: Thanks~ I do like this better, but I don't know the difference between them and in what case that parenthesis are needed, could you maybe provide one example? Which I'll probably add as one of the test cases.

Jarcho: `replace(&mut *x, None)` is the only practical example. If `x` is a type implementing `Deref` then the deref can't always be stripped out. This leaves you with `(*x).take()` as the required suggestion.

`x + x` could also return `&mut Option<_>`, but that's neither common nor recommended.

J-ZhengLi: Ahh, that's interesting, I completely forgot that's the case.
also, I've add a related test case

Jarcho:
Looks good. Thank you. @bors r+

Issue #10518, opened 2023-03-18
Suggest `size_of_val(a)` for `a.len() * size_of::<T>()`


scottmcm:
### What it does

When `a` is `&[T]`, detect `a.len() * size_of::<T>()` and suggest `size_of_val(a)` instead.

### Lint Name

manual_slice_size_calculation

### Category

complexity, perf

### Advantage

- Shorter to write
- Removes the need for the human and the compiler to worry about overflow in the multiplication
- Potentially faster at runtime as rust emits special no-wrapping flags when it calculates the byte length
- Less turbofishing

### Example

https://github.com/rust-lang/rust/blob/13afbdaa0655dda23d7129e59ac48f1ec88b2084/library/core/src/hash/mod.rs#L837

```rust
let newlen = data.len() * mem::size_of::<$ty>();
```

Could be written as:

```rust
let newlen = mem::size_of_val(data);
```

asquared31415:
`slice.len()`, `usize` multiplication, and `size_of::<T>` are all `const`, but `size_of_val` is only unstably const.  Code may be using this pattern out of necessity.

blyxyas:
> `size_of_val`  is only unstably const.

Then maybe only linting for nightly builds? It can be quickly checked from `cx.sess().is_nightly_build()`. As far as I'm aware, there are no unstable Clippy lints.

Alexendoo:
It can still lint on stable, but only outside of const contexts which can be checked using [`in_constant`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.in_constant.html)

PR #10601, opened 2023-04-06
Add [`manual_slice_size_calculation`]

schubart:
Fixes: #10518

---

changelog: new lint [`manual_slice_size_calculation`]

### DIFF
--- a/clippy_lints/src/manual_slice_size_calculation.rs
+++ b/clippy_lints/src/manual_slice_size_calculation.rs
@@ -1,5 +1,5 @@
 use clippy_utils::diagnostics::span_lint_and_help;
-use clippy_utils::in_constant;
+use clippy_utils::{expr_or_init, in_constant};
 use rustc_hir::{BinOpKind, Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::ty;
@@ -60,6 +60,9 @@ fn simplify<'tcx>(
     expr1: &'tcx Expr<'tcx>,
     expr2: &'tcx Expr<'tcx>,
 ) -> Option<&'tcx Expr<'tcx>> {
+    let expr1 = expr_or_init(cx, expr1);
+    let expr2 = expr_or_init(cx, expr2);
+
     simplify_half(cx, expr1, expr2).or_else(|| simplify_half(cx, expr2, expr1))
 }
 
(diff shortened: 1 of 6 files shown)


clippy_lints/src/manual_slice_size_calculation.rs
```diff
@@ -0,0 +1,90 @@
...
+    }
+}
+
+fn simplify<'tcx>(
+    cx: &LateContext<'tcx>,
+    expr1: &'tcx Expr<'tcx>,
+    expr2: &'tcx Expr<'tcx>,
+) -> Option<&'tcx Expr<'tcx>> {
+    simplify_half(cx, expr1, expr2).or_else(|| simplify_half(cx, expr2, expr1))
+}
```
llogiq, line 64: I swear I've seen this pattern multiple times. Perhaps we should make a generic `clippy_utils` function of it.

llogiq (commented):
I'd like to see a test where either the size or length is bound to a temporary variable first. It's ok for now if that isn't linted, but sufficiently easy to change.

Otherwise the code looks good, and I'll gladly merge it with the additional test added.

schubart:
> I'd like to see a test where either the size or length is bound to a temporary variable first. It's ok for now if that isn't linted, but sufficiently easy to change.
> 
> Otherwise the code looks good, and I'll gladly merge it with the additional test added.

Done.

I feel this kind of data flow analysis (where does a value come from, what do we know about it?) could be a whole new thing, applicable to a lot more than just this lint here.

For example, in the following Java code Intellij warns that the `if` condition is always true:

```java
List<String> strings = Collections.emptyList();
if (strings.isEmpty()) {
    System.out.println("There are no strings");
}
```

and in this code it warns that the list access will always fail:

```java
List<String> strings = foo();
if (strings.isEmpty()) {
    System.out.println(strings.get(0));
}
```

Do we have anything like this in Clippy?

llogiq:
We have, it's `clippy_utils::expr_or_init(..)`. If you want to add this directly to this PR, feel free, otherwise I'll have this merged and we can do a follow-up.

schubart:
Thanks, I'll give it a shot. Let's not merge this yet.

schubart:
Done. Thanks for the suggestion, this is very nice!

llogiq:
Great!

@bors r+

PR #10607, opened 2023-04-07
Add spans to `clippy.toml` error messages

beetrees:
Adds spans to errors and warnings encountered when parsing `clippy.toml`.

changelog: Errors and warnings generated when parsing `clippy.toml` now point to the location in the TOML file the error/warning occurred.


src/driver.rs
```diff
@@ -83,11 +83,6 @@ fn track_files(parse_sess: &mut ParseSess, conf_path_string: Option<String>) {
         file_depinfo.insert(Symbol::intern("Cargo.toml"));
     }
 
-    // `clippy.toml`
-    if let Some(path) = conf_path_string {
-        file_depinfo.insert(Symbol::intern(&path));
-    }
```
Alexendoo, line 89: It'd be worth leaving a comment behind mentioning that `clippy.toml` is being loaded elsewhere

clippy_lints/src/utils/conf.rs
```diff
@@ -67,33 +70,61 @@ impl DisallowedPath {
...
 
+impl From<io::Error> for TryConf {
+    fn from(value: io::Error) -> Self {
+        ConfError::from(value).into()
+    }
+}
+
 #[derive(Debug)]
-struct ConfError(String);
+pub struct ConfError(pub String, pub Option<Span>);
```
giraffate, line 100: IMO it would be easy to understand when accessing fields if its field has name, like `error.message` and `error.span`.
```suggestion
pub struct ConfError {
    pub message: String,
    pub span: Option<Span>,
}
```

giraffate (commented):
Sorry for the late reviewing. Overall looks good, thanks! I made one comment.

giraffate:
@bors r+

Thanks!

PR #10613, opened 2023-04-09
Add lint `assigning_clones`.

### LINKED ISSUE #1709: Lint to suggest clone_from and clone_into
Basically, we'd suggest the following replacements:

`a = b.clone()` => `a.clone_from(&b)`
`a = b.to_owned()` => `b.clone_into(&mut a)`

kpreid:
This lint reports when `*a = b.clone();` can be replaced with `a.clone_from(&b);`, and the same for `ToOwned::clone_into()`. Fixes #1709.

This is my first attempt at writing code for `clippy` or `rustc`, and I could use help with these missing parts:

* [ ] Does not correctly verify that the call is a call to `Clone::clone()` rather than some other similarly named method. I need advice on how to ask rustc for this information; the lint isn't really correct without it.
* [ ] Does not detect function call syntax in addition to method call syntax. I will fix this after the previous item.
* [ ] Does not lint on assignment to a local variable or `DerefMut` coercion, only on assignment to `*{some &mut T}`, because it does not know whether the local variable might be uninitialized. I need advice on how to ask rustc for this information. (But if it's not feasible, then the lint will still work, just not apply to as many things.)

changelog: new lint [`assigning_clones`]

### DIFF
--- a/clippy_lints/src/assigning_clones.rs
+++ b/clippy_lints/src/assigning_clones.rs
@@ -56,7 +56,7 @@ impl<'tcx> LateLintPass<'tcx> for AssigningClones {
             return;
         }
 
-        let op = if method_name.ident.name == sym!(clone) {
+        let op = if method_name.ident.name == sym::clone {
             Op::Clone
         } else if method_name.ident.name == sym!(to_owned) {
             Op::ToOwned
--- a/clippy_lints/src/assigning_clones.rs
+++ b/clippy_lints/src/assigning_clones.rs
@@ -1,0 +1,204 @@
+use clippy_utils::{diagnostics::span_lint_and_then, sugg::Sugg, ty::implements_trait};
+use rustc_errors::Applicability;
+use rustc_hir::{self as hir, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for code like `foo = bar.clone();`
+    ///
+    /// ### Why is this bad?
+    /// If cloning `bar` allocates memory (or other resources), then this code will allocate
+    /// new memory for the clone of `bar`, then drop `foo`, then overwrite `foo` with the clone.
+    /// Instead, `Clone::clone_from()`, or `ToOwned::clone_into()`, may be able to update
+    /// `foo` in-place, reusing existing memory.
+    ///
+    /// Note that this only provides any actual improvement if the type has explicitly implemented
+    /// the `clone_from()` trait method, since the trait-provided implementation will just call
+    /// `clone()`.
+    ///
+    /// ### Example
+    /// ```rust
+    /// #[derive(Clone)]
+    /// struct Thing;
+    ///
+    /// pub fn assign_to_ref(a: &mut Thing, b: Thing) {
+    ///     *a = b.clone();
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// #[derive(Clone)]
+    /// struct Thing;
+    ///
+    /// pub fn assign_to_ref(a: &mut Thing, b: Thing) {
+    ///     a.clone_from(&b);
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub ASSIGNING_CLONES,
+    perf,
+    "assigning the result of cloning may be inefficient"
+}
+declare_lint_pass!(AssigningClones => [ASSIGNING_CLONES]);
+
+impl<'tcx> LateLintPass<'tcx> for AssigningClones {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, assign_expr: &'tcx hir::Expr<'_>) {
+        let ExprKind::Assign(assign_target, clone_call, _span) = assign_expr.kind else { return };
+        // TODO: Also look for `Clone::clone` function calls, not just method calls
+        let ExprKind::MethodCall(method_name, clone_receiver, args, _span) = clone_call.kind else { return };
+
+        // Fast syntactic check: if it has args it can't be the call we are looking for,
+        // so we don't even need to consult the types.
+        if !args.is_empty() {
+            return;
+        }
+
+        let op = if method_name.ident.name == sym!(clone) {
+            Op::Clone
+        } else if method_name.ident.name == sym!(to_owned) {
+            Op::ToOwned
+        } else {
+            return;
+        };
+
+        if ok_to_suggest(cx, op, assign_target, clone_call).is_some() {
+            suggest(cx, op, assign_expr, assign_target, clone_receiver);
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+enum Op {
+    Clone,
+    ToOwned,
+}
+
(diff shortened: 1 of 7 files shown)


Nilstrieb:
Note that the derive macro currently doesn't implement clone_from and it's unlikely to do so in the future: https://github.com/rust-lang/rust/pull/98445#issuecomment-1190681305

kpreid:
> Note that the derive macro currently doesn't implement clone_from and it's unlikely to do so in the future

Indeed, that's why the note says “_may_ be inefficient”. I could imagine the lint pass checking the implementing type to see whether it overrides `clone_from` or not to decide whether to emit the lint (or how to phrase it). I also thought about mentioning the derive's behavior in the lint documentation, but it seems not great to do that when it might, in fact, be wrong in the future without being updated. (But maybe that's worth doing anyway.)

However, my hope here is that the lint will raise awareness of the _existence_ of `clone_from`, and authors of types will more often implement it when it is applicable, thus expanding the performance gains. The rewrite this lint proposes should not significantly _decrease_ performance even if it gains none, so the cost is only in “churn” (people seeing a new lint and fixing/allowing it), which is not to be ignored! But discussing whether this lint should not exist at all should probably happen on the issue, not the PR, yes?

Another idea: there could be a configuration option (or splitting this lint into two lints) to have it lint only on types _known_ to have `clone_from()` implementations.

clippy_lints/src/assigning_clones.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// #[derive(Clone)]
+    /// struct Thing;
+    ///
+    /// pub fn assign_to_ref(a: &mut Thing, b: Thing) {
+    ///     a.clone_from(&b);
+    /// }
+    /// ```
```
xFrednet, line 39:
```suggestion
    /// ### Example
    /// ```rust
    /// # let src = "Cheesecake".to_string();
    /// let mut target = "Hello world".to_string();
    /// // [...]
    /// target = src.clone();
    /// ```
    /// Use instead:
    /// ```rust
    /// # let src = "Cheesecake".to_string();
    /// let mut target = "Hello world".to_string();
    /// // [...]
    /// target.clone_from(&src);
    /// ```
```

clippy_lints/src/assigning_clones.rs
```diff
@@ -0,0 +1,204 @@
...
+    let clone_result_type = cx.typeck_results().expr_ty_adjusted(clone_call);
+    if !(implements_trait(cx, clone_result_type, op.expected_trait(cx)?, &[])) {
+        return None;
+    }
+
+    // If we're assigning to a dereferenced reference, then we know the place is already valid.
+    // On the other hand, if the place is a variable or a Box, it might be uninitialized,
+    // in which case the suggestion might be wrong.
+    // TODO: Actually ask whether the place is uninitialized at this point, instead of
+    // guessing based on the syntax and type.
```
xFrednet, line 98: Finding the origin of a variable can be a bit difficult, when it comes to patterns, function parameters etc. Though, in this case, we only want to find the `let <var>` statement if it's a local value. This is a rough sketch of how to check for this:

```
if let Some(id) = path_to_local(expr) {
    if let Some(Node::Stmt(stmt)) = cx.tcx.hir().find(id) {
        // if stmt is let, check for init expr
    }
}
```

[`clippy_utils::expr_or_init`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.expr_or_init.html) could also be interesting to look at :)

clippy_lints/src/assigning_clones.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn expected_trait(self, cx: &LateContext<'_>) -> Option<hir::def_id::DefId> {
+        match self {
+            Op::Clone => cx.tcx.lang_items().clone_trait(),
+            Op::ToOwned => cx.tcx.get_diagnostic_item(sym::ToOwned),
+        }
+    }
+
+    fn message(self) -> &'static str {
+        // TODO: Use the receiver type to say "is" instead of "may be" for types which
+        // are known to have optimizations (e.g. `String`).
```
xFrednet, line 139: You can use type checking, to get the type of the clone call. If it's an ADT, you can use the `DefId` to check for known types :)

See:
* [Checking for a specific type](https://doc.rust-lang.org/clippy/development/common_tools_writing_lints.html#checking-for-a-specific-type)
* [Diagnostic items](https://rustc-dev-guide.rust-lang.org/diagnostics/diagnostic-items.html)

clippy_lints/src/assigning_clones.rs
```diff
@@ -0,0 +1,204 @@
...
+        assign_target: &hir::Expr<'tcx>,
+        clone_receiver: &hir::Expr<'tcx>,
+        applicability: &mut Applicability,
+    ) -> String {
+        match self {
+            Op::Clone => {
+                // The assignment LHS, which will become the receiver of the `.clone_from()` call,
+                // should lose one level of dereference operator since autoref takes care of that.
+                let target_sugg = if let ExprKind::Unary(hir::UnOp::Deref, ref_expr) = assign_target.kind {
+                    Sugg::hir_with_applicability(cx, ref_expr, "_", applicability)
```
xFrednet, line 165: Nice usage of `hir_with_applicability` :+1: 

You can resolve this directly :)

xFrednet (commented):
Thank you for the PR. I left remarks regarding most todos. As suggested in a comment, it would probably be best to only lint cases, where a custom `clone_from` implementation exists. I'm not 100% sure how to do that right now, but will look it up, until the next review. I hope that this gives you enough hints to continue on the implementation for now.

You're welcome to reach out if you have any questions :)

clippy_lints/src/assigning_clones.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// #[derive(Clone)]
+    /// struct Thing;
+    ///
+    /// pub fn assign_to_ref(a: &mut Thing, b: Thing) {
+    ///     a.clone_from(&b);
+    /// }
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub ASSIGNING_CLONES,
+    perf,
```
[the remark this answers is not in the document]
kpreid: I personally like option 3 for now — focusing on giving definitely-useful advice. I haven't gotten back to working on this yet, but when I do I will look into that, and if doesn't seem feasible to do that then I'll change the category to `pedantic`.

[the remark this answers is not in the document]
xFrednet: We've discussed the topic in the meeting 2 weeks ago, but I missed writing a comment then :sweat_smile:. Here is a link to the discussion:

https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202023-05-02/near/355124756

It was suggested to add a clippy attribute that crates can use, to mark a struct for this lint. Also restricting the lint to custom `clone_from` suggestions was encouraged, IIRC

xFrednet:
Hey @kpreid, this is a ping from triage. Do you need any further assistance with this PR? :)

kpreid:
> Do you need any further assistance with this PR? :)

Not specifically; I just have not gotten back to it yet.

xFrednet:
Hey @kpreid, I'm closing this PR due to inactivity. If you want to continue this, you can add a comment for me to reopen this PR or simply create a new one. Thank you for the time you already put into this lint!

PR #10614, opened 2023-04-09
Clear with drain

### LINKED ISSUE #10572: Want to extend `clear_with_drain` to other types of containers
I implemented [clear_with_drain](https://github.com/rust-lang/rust-clippy/pull/10528) for `Vec`, and then I had a look at #9059 and saw that `manual_retain` applies to a bunch of containers (BTW I admire your work on that lint @kyoto7250 :clap:).

So I'm thinking `clear_with_drain` should probably be extended in a similar fashion to apply to the collections in `std` that have both a `clear` and a `drain` method, i.e.:
- [String](https://doc.rust-lang.org/stable/std/string/struct.String.html#)
- [HashSet](https://doc.rust-lang.org/stable/std/collections/hash_set/struct.HashSet.html#)
- [HashMap](https://doc.rust-lang.org/stable/std/collections/hash_map/struct.HashMap.html#)
- [VecDeque](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html#)
- [BinaryHeap](https://doc.rust-lang.org/stable/std/collections/struct.BinaryHeap.html#)

What do you think?

Also note that the documentations for [Vec](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#examples-34), [VecDeque](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html#examples-23) and [String](https://doc.rust-lang.org/stable/std/string/struct.String.html#examples-36) explicitly mention that `drain` with a full range will clear the container, "like `clear()` does". Does this change anything regarding the category proposed originally in #9339? Style was suggested, but that would mean the examples in the standard library would get warnings :sweat_smile:. Maybe Pedantic is more suitable?

@rustbot label +C-enhancement

bluthej:
Fixes #10572: both the original intent of the issue (extending `clear_with_drain`) and the false negative for `collection_is_never_read` I found in the process are fixed by this PR.

changelog: [`clear_with_drain`]: extend to 5 other types of containers. [`collection_is_never_read`]: fix false negative for `String`s.


schubart (approved):
The change to `collection_is_never_read` looks good to me.

Manishearth (approved):

PR #10616, opened 2023-04-09
new lint: `missing_fields_in_debug`

### LINKED ISSUE #10429: Lint missing fields in manual Debug implementations
### What it does

This lint finds fields that implement Debug but are not printed in a struct's Debug implementation.

This lint would ignore fields in types that do not implement Debug.

### Lint Name

debug_missing_field

### Category

style, pedantic

### Advantage

- When adding a new field to a type that implements Debug manually, you may forget to update the Debug implementation. This lint would ensure new fields are linted if not printed in the debug implementation.

### Drawbacks

- Both tuple and struct debug implementations need to be considered.
- Enums are a little more complicated to lint, since you would need check the fields of each variant.
- The lint may have false positives if the Debug implementation uses manual write! invocations instead of debug_struct and debug_tuple.

### Example

```rust
pub struct Example {
    non_debug: NonDebug,
    debug: u32,
    new: u16,
}

impl Debug for Example {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        f.debug_struct("Example").field("debug", &self.debug).finish_non_exhaustive()
    }
}
```

Could be written as:

```rust
pub struct Example {
    non_debug: NonDebug,
    debug: u32,
    new: u16,
}

impl Debug for Example {
    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
        f.debug_struct("Example").field("debug", &self.debug).field("new", &self.new).finish_non_exhaustive()
    }
}
```

y21:
Fixes #10429

This PR adds a new lint that looks for manual `Debug` implementations that do not "use" all of the fields.
This often happens when adding a new field to a struct.
It also acts as a style lint in case leaving out a field was intentional. In that case, it's preferred to use [`DebugStruct::finish_non_exhaustive`](https://doc.rust-lang.org/stable/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive), which indicates that there are more fields that were explicitly not shown.

```
changelog: [`missing_fields_in_debug`]: missing fields in manual `Debug` implementation
```


clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    let _: Option<!> = for_each_expr(block, |expr| {
+        let ExprKind::MethodCall(path, recv, ..) = &expr.kind else {
+            return ControlFlow::Continue(());
+        };
+
+        if_chain! {
+            if [sym::debug_struct, sym::debug_tuple].contains(&path.ident.name);
+            if let Some(ty) = typeck_results.expr_ty(recv).peel_refs().ty_adt_def();
+            if match_def_path(cx,ty.did(), &paths::FORMATTER);
```
y21, line 121: `is_type_diagnostic_item(cx, ty, sym::Formatter)` doesn't work for some reason. It returns false despite `dbg!(ty)` showing that it's `std::fmt::Formatter<'_>`. Not sure why

clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,291 @@
...
+            then {
+                has_debug_struct_tuple = true;
+            }
+        }
+
+        if_chain! {
+            if path.ident.name.as_str() == "finish_non_exhaustive";
```
y21, line 128: Self reminder: add this as a symbol over at `rust-lang/rust`

clippy_lints/src/missing_fields_in_debug.rs
Jarcho: `Formatter` isn't a diagnostic item. You can use `match_type` and add the item to `clippy_utils::paths`.

tests/ui/missing_field_in_debug.rs
Alexendoo, line 1: Could do with a rename `missing_field_...` -> `missing_fields_...`

clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,293 @@
...
+        .filter_map(|field| {
+            if field_accesses.contains(&field.ident) {
+                None
+            } else {
+                Some((field.span, HELP_FIELD_UNUSED))
+            }
+        })
+        .collect::<Vec<_>>();
+
+    if !span_notes.is_empty() {
```
Alexendoo, line 178: We probably want to check that at least one field is being accessed, so the patterns like the following aren't linted

```rust
use std::fmt::{Debug, Formatter, Result};

struct Inner {
    a: usize,
    b: usize,
}

struct HasInner {
    inner: Inner,
}

impl HasInner {
    fn get(&self) -> &Inner {
        &self.inner
    }
}

impl Debug for HasInner {
    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
        let inner = self.get();

        f.debug_struct("HasInner")
            .field("a", &inner.a)
            .field("b", &inner.b)
            .finish()
    }
}
```

Plus a test for that

clippy_lints/src/missing_fields_in_debug.rs
Alexendoo: Checking like this is fine, we only add symbols upstream when they're needed as part of a diagnostic item

tests/ui/missing_field_in_debug.stderr
```diff
@@ -0,0 +1,156 @@
...
+LL | | }
+   | |_^
+   |
+note: this field is unused
+  --> $DIR/missing_field_in_debug.rs:49:30
+   |
+LL | struct UnnamedStruct1Ignored(String);
+   |                              ^^^^^^
+   = help: consider including all fields in this `Debug` impl
+   = help: consider calling `.finish_non_exhaustive()` if you intend to ignore fields
```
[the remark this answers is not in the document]
y21: Huh, TIL. Do we just not lint tuple structs then, or accept that there's going to be FPs and lint anyway, and have the user manually `#[allow]` them? It is a pedantic lint after all. Not sure what's better in this case.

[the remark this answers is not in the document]
Alexendoo: I'm not sure either, but I'm leaning towards ignoring tuple structs, it seems a lot easier to accidentally miss a struct field and tuple structs changing size is less common IME

clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,276 @@
...
+
+    let _: Option<!> = for_each_expr(block, |expr| {
+        if let ExprKind::Field(target, ident) = expr.kind
+            && let target_ty = typeck_results.expr_ty(target).peel_refs()
+            && target_ty == self_ty
+        {
+            field_accesses.insert(ident);
+        }
+        ControlFlow::Continue(())
+    });
```
Alexendoo, line 153: It would be good to also catch things like `.field("some_field", &())` or where the value is computed without accessing the field

tests/ui/missing_field_in_debug.stderr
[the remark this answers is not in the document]
y21: Alright, I removed all mentions of tuple structs so it lints only `DebugStruct` instances now. I guess we can always add debug tuples later if and when `DebugTuple::finish_non_exhaustive` gets added

clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,298 @@
...
+    self_ty: Ty<'tcx>,
+    item: &'tcx Item<'tcx>,
+    data: &VariantData<'_>,
+) {
+    let mut has_direct_field_access = false;
+    let mut field_accesses = FxHashSet::default();
+
+    for_each_expr(block, |expr| {
+        if let ExprKind::Field(target, ident) = expr.kind
+            && let target_ty = typeck_results.expr_ty(target).peel_refs()
```
Alexendoo, line 163: I had a look through the results of

```sh
cargo lintcheck --filter missing_fields_in_debug --recursive
```

A situation it found was where the fields are accessed via a `Deref` impl such as in `imgui`

```rust
use std::fmt::Debug;
use std::ops::Deref;

struct Inner {
    a: usize,
    b: usize,
}
struct Wrapper(Inner);

impl Deref for Wrapper {
    type Target = Inner;

        &self.0
    }
}

impl Debug for Wrapper {
        f.debug_struct("Wrapper")
            .field("a", &self.a)
            .field("b", &self.b)
            .finish()
    }
}
```

To account for derefs we'd want to change the `expr_ty`s in the lint to `expr_ty_adjusted` instead

Alexendoo: The other was there's a good amount of structs out there just omitting a `PhantomData` field, it may be worth special casing that but I'm not sure

y21: Imo it's fine to leave out `PhantomData`, since it doesn't really add any value and is mostly just for teaching the compiler about certain safety properties of generic parameters? I suppose including it means you can see what variance it has in its generics, but I guess that isn't usually what people are looking for when debug formatting a struct. I special-cased it to allow omitting phantomdata (and also rebased + squash commits to cleanup the history a bit)

y21:
Is there anything else or is it just waiting on a rebase?

clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,311 @@
...
+                check_field_access(sym, expr);
+            }
+            ControlFlow::<!, _>::Continue(())
+        });
+
+        arm.pat.each_binding(|_, _, span, pat_ident| {
+            if !field_accesses.contains(&pat_ident) {
+                span_notes.push((span, "the field referenced by this binding is unused"));
+            }
+        });
```
Alexendoo, line 272: Is this any different to an `unused_variables` warning?

Alexendoo:
Sorry for the delay, can't remember if I had anything else in mind so I just read it over again this weekend 😄

A thought that came to me just now is maybe we don't actually need to lint enums, unlike structs where it's very easy to add a new field and forget to update the `Debug` impl using a `match` already has you consider enum variants fully. The author would have to explicitly ignore the fields with `..`/`_` in order to miss them

clippy_lints/src/missing_fields_in_debug.rs
```diff
@@ -0,0 +1,311 @@
...
+        }
+        ControlFlow::<!, _>::Continue(())
+    });
+
+    let span_notes = data
+        .fields()
+        .iter()
+        .filter_map(|field| {
+            let EarlyBinder(field_ty) = cx.tcx.type_of(field.def_id);
+            if field_accesses.contains(&field.ident.name) || field_ty.is_phantom_data() {
```
Alexendoo, line 189: For the `PhantomData` check we have access to the hir ty so we can use [`is_path_lang_item(cx, field.ty, LangItem::PhantomData)`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_path_lang_item.html)

y21:
> A thought that came to me just now is maybe we don't actually need to lint enums, unlike structs where it's very easy to add a new field and forget to update the Debug impl using a match already has you consider enum variants fully.

That's a very good point, I agree with this. I went ahead and removed the enum checking. That made me realize that one can do something similar with structs by destructuring `self` (you get a compiler error when a new field is added):
```rs
struct F { a: i32, b: i32 }

impl fmt::Debug for F {
  fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
    let Self { a } = self; // error, b not handled
    // ...
  }
}
```
Although a lint for this would still be nice 😅

Alexendoo:
Thanks! Yeah that's a good way of catching it

With a rebase for the conflicts I'd say this is ready to merge

Alexendoo (approved):

Alexendoo:
👍

@bors r+

PR #10618, opened 2023-04-09
[missing_const_for_fn] fix #7121

### LINKED ISSUE #7121: `missing_const_for_fn`: Known Problems should include future compatibility hazard
Marking a private function as const when possible does not cause an issue. But marking a public function as const can become a problem in the future if the implementation changes, as removing const would break API compatibility. It is true that the `missing_const_for_fn` lint is allow by default, but I feel that this future compatibility hazard should be included in the Known Problems.

KisaragiEffective:

changelog: [`missing_const_for_fn`] note about removing const may break compatibility

fix #7121

### DIFF
--- a/clippy_lints/src/missing_const_for_fn.rs
+++ b/clippy_lints/src/missing_const_for_fn.rs
@@ -41,6 +41,7 @@ declare_clippy_lint! {
     /// can't be const as it calls a non-const function. Making `a` const and running Clippy again,
     /// will suggest to make `b` const, too.
     ///
+    /// If you are marking public function with const, doing opposite will break API compatibility.
     /// ### Example
     /// ```rust
     /// # struct Foo {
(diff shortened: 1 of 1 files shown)


clippy_lints/src/missing_const_for_fn.rs
```diff
@@ -41,6 +41,7 @@ declare_clippy_lint! {
     /// can't be const as it calls a non-const function. Making `a` const and running Clippy again,
     /// will suggest to make `b` const, too.
     ///
+    /// If you are marking public function with const, doing opposite will break API compatibility.
```
llogiq, line 44:
```suggestion
    /// If you are marking a public function with `const`, removing it again will break API compatibility.
```

llogiq (commented):
Good idea, I have a grammar improvement though.

KisaragiEffective:
Thank you for quick review (and improve grammer suggestion), I've applied it!

llogiq:
Thank you!

@bors r+

PR #10624, opened 2023-04-10
remove unusued `#![feature(drain_filter)]`

the8472:
The unstable feature does not appear to be used and its presence blocks work in https://github.com/rust-lang/rust/pull/104455

changelog: none


xFrednet (approved):

xFrednet:
Interesting, looks like the feature was once used, but is no longer needed. Thank you for checking this :)

@bors r+

Issue #10597, opened 2023-04-05
[Tracking issue] Project: Clippy Book Chapter Updates Reborn


blyxyas:
Tracking issue for all relating to the Clippy Book Chapter Updates Reborn project.
If you have anything relating to the whole project, please comment it here instead of in an individual PR.

| NAME           	| STATUS             	| REF 	|
|----------------	|--------------------	|----	|
| Defining lints 	| PR Open 🔮          	| #10595    	|
| Writing tests  	| PR Open 🔮 	| #10596   	|
| Lint passes | Needs discussion 🐝 |    	|
| Emitting lints | Working on 🦕 |    	|
|Type Checking| Not started yet 🍅 |    	|
|Trait Checking| Not started yet 🍅 |    	|
|Method Checking| Not started yet 🍅 |    	|
|Macro Expansions| Not started yet 🍅 |    	|
|HirIds| Not started yet 🍅 |    	|
|Example: Early Lint| Not started yet 🍅 |    	|
|Adding documentation| Not started yet 🍅 |    	|

You can comment about which chapters to add or remove.

## Ideas

* "Tips from collaborators / experts" section / chapter? 

## Discussions

* Should these new updates replace the current "Adding lints" chapter? (👍 Y / 👎 N)
* Should they use an storytelling / informal style to welcome new users? (🎉 Y / 😕 N)
* Lint passes: Re-write the current section ([here](https://doc.rust-lang.org/clippy/development/adding_lints.html#lint-passes)) or create a whole new chapter? (❤️ Re-write / 👀 Create)   For this one take into account that this project is basically de-structuring the "Adding Lints" chapter into a lot of smaller ones, so creating a new chapter isn't a big deal.

For more detailed answer about these (or new) discussions, you can also create a comment and I'll add it here.

---

A lot of the discussion about this project was done through the draft PR #9426, but please, use this tracking issue instead.

blyxyas:
|b= idea x

xFrednet:
> Should these new updates replace the current "Adding lints" chapter?

Yes, in my opinion they should. The current "Adding lints" grew over time and is now too long and not well-structured. There is no real separation between that and the "Common Tools" section. Reworking that would be a gigantic help

> Should they use an storytelling / informal style to welcome new users?

This depends on the part of the documentation. I like the storytelling style, but I know that I'm sometimes less formal than others in this project. Here I would put trust in the reviewer :)

[...]

PR #10626, opened 2023-04-10
Clippy Book Chapter Updates Reborn: Trait Checking

blyxyas:
This PR adds a new chapter to the book: "Trait Checking". No major changes from the source (just some typos, re-phrasing, the usual).

## Notes

- Does not require any other PR to be merged.
- To talk about the whole project, please use the tracking issue for the project #10597 (It also contains a timeline, discussions and more information)

changelog: Add a new "Trait Checking" chapter to the book

### DIFF
--- a/book/src/development/trait_checking.md
+++ b/book/src/development/trait_checking.md
@@ -102,5 +102,4 @@ impl LateLintPass<'_> for CheckTokioAsyncReadExtTrait {
 [symbol]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/symbol/struct.Symbol.html
 [symbol_index]: https://doc.rust-lang.org/beta/nightly-rustc/rustc_span/symbol/sym/index.html
 [TyCtxt]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html
-[using_diagnostic_items]: https://rustc-dev-guide.rust-lang.org/diagnostics/diagnostic-items.html#using-diagnostic-items
 [rust]: https://github.com/rust-lang/rust
--- a/book/src/development/trait_checking.md
+++ b/book/src/development/trait_checking.md
@@ -78,8 +78,7 @@ implementation.
 
 > **Note**: This approach should be avoided if possible, the best thing to do would be to make a PR to [`rust-lang/rust`][rust].
 
-Below, we check if the given `expr` implements `tokio`'s
-[`AsyncReadExt`][AsyncReadExt] trait:
+Below, we check if the given `expr` implements the `Iterator`'s trait method `cloned` :
 
 ```rust
 use clippy_utils::{match_trait_method, paths};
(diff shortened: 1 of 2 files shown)


book/src/development/trait_checking.md
```diff
@@ -0,0 +1,112 @@
...
+}
+```
+
+## Using Type Path
+
+If neither diagnostic item or lang item is available, we can use
+[`clippy_utils::paths`][paths] with the `match_trait_method` to determine trait
+implementation.
+
+> **Note**: This approach should be avoided if possible.
```
flip1995, line 80: I would add, that instead one should consider adding a diag item to the rust compiler and then later use that instead. Best thing would be to link a rustc PR that does this as an example.

book/src/development/trait_checking.md
```diff
@@ -0,0 +1,112 @@
...
+[`AsyncReadExt`][AsyncReadExt] trait:
+
+```rust
+use clippy_utils::{match_trait_method, paths};
+use rustc_hir::Expr;
+use rustc_lint::{LateContext, LateLintPass};
+
+impl LateLintPass<'_> for CheckTokioAsyncReadExtTrait {
+    fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
+        if match_trait_method(cx, expr, &paths::TOKIO_IO_ASYNCREADEXT) {
```
flip1995, line 92: NIT: please use a path that is not to an external crate (I don't want to give the idea that adding paths to external crates is accepted. We try to avoid that as much as possible)

book/src/development/trait_checking.md
```diff
@@ -0,0 +1,112 @@
...
+            println!("`expr` implements `TOKIO_IO_ASYNCREADEXT` trait!");
+        }
+    }
+}
+```
+
+> **Note**: Even though all the `clippy_utils` methods we have seen in this
+> chapter takes `expr` as a parameter, these methods are actually using
+> each expression's `HirId` under the hood.
```
flip1995, line 101: Yeah, rather add a list to those methods and explain this there.

book/src/development/trait_checking.md
blyxyas: Ok, just a question. Why is this specific path accepted?

book/src/development/trait_checking.md
blyxyas: I think I'm just going to remove the note. I don't see it being really necessary here (if someone is curious, they'll check the implementation)

book/src/development/trait_checking.md
```diff
@@ -0,0 +1,105 @@
...
+}
+```
+
+## Using Type Path
+
+If neither diagnostic item nor a language item is available, we can use
+[`clippy_utils::paths`][paths] with the `match_trait_method` to determine trait
+implementation.
+
+> **Note**: This approach should be avoided if possible, the best thing to do would be to make a PR to [`rust-lang/rust`][rust].
```
flip1995, line 79:
```suggestion
> **Note**: This approach should be avoided if possible, the best thing to do would be to make a PR to [`rust-lang/rust`][rust] adding a diagnostic item.
```

flip1995 (approved):
This LGTM.

All of the remaining 3 PRs are now ready to be merged. However, some links need to be figured out I think. @blyxyas Can I ask you to do that for those 3 PRs and then `r=me` on them in whatever order makes the most sense.

After all those PRs are merged, I think the only thing left to do is address the `FIXME`s in those chapters and cross link those chapters.

blyxyas:
I'll fix them. Btw, are you talking about the link order in `SUMMARY.md` or the missing links in the `FIXME`s

flip1995:
Mainly missing lints in `FIXME` comments. If you think the order in the ToC needs to be adapted go ahead. I didn't pay much attention to that TBH.

PR #10629, opened 2023-04-11
Fix UB in `as_ptr_cast_mut` documentation

### LINKED ISSUE #10628: "Use instead" example for `as_ptr_cast_mut` has undefined behavior
### Description

The documentation for the [`as_ptr_cast_mut` lint](https://rust-lang.github.io/rust-clippy/master/index.html#as_ptr_cast_mut) suggests the following as an example of correct code:
```rust
let mut string = String::with_capacity(1);
let ptr = string.as_mut_ptr();
unsafe { ptr.write(4) };
```
However, running this code in Miri on the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=530c944d09cf5d84c05b8621359a3bd6) gives the following error, indicating undefined behavior:
<details>
<pre>
error: Undefined Behavior: attempting a write access using <2996> at alloc1527[0x0], but that tag does not exist in the borrow stack for this location
 --> src/main.rs:4:14
  |
4 |     unsafe { ptr.write(4) };
  |              ^^^^^^^^^^^^
  |              |
  |              attempting a write access using <2996> at alloc1527[0x0], but that tag does not exist in the borrow stack for this location
  |              this error occurs as part of an access at alloc1527[0x0..0x1]
  |
  = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
  = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <2996> would have been created here, but this is a zero-size retag ([0x0..0x0]) so the tag in question does not exist anywhere
 --> src/main.rs:3:15
  |
3 |     let ptr = string.as_mut_ptr();
  |               ^^^^^^^^^^^^^^^^^^^
  = note: BACKTRACE (of the first span):
  = note: inside `main` at src/main.rs:4:14: 4:26

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to previous error</pre>
</details>

### Version

_No response_

### Additional Labels

@rustbot label +A-documentation

Alexendoo:
changelog: none

Fixes #10628

There's no `String::as_mut_ptr` surprisingly, so the example is actually calling `str::as_mut_ptr` on an empty `str`

### DIFF
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -625,14 +625,14 @@ declare_clippy_lint! {
     ///
     /// ### Example
     /// ```rust
-    /// let string = String::with_capacity(1);
-    /// let ptr = string.as_ptr() as *mut u8;
+    /// let mut vec = Vec::<u8>::with_capacity(1);
+    /// let ptr = vec.as_ptr() as *mut u8;
     /// unsafe { ptr.write(4) }; // UNDEFINED BEHAVIOUR
     /// ```
     /// Use instead:
     /// ```rust
-    /// let mut string = String::with_capacity(1);
-    /// let ptr = string.as_mut_ptr();
+    /// let mut vec = Vec::<u8>::with_capacity(1);
+    /// let ptr = vec.as_mut_ptr();
     /// unsafe { ptr.write(4) };
     /// ```
     #[clippy::version = "1.66.0"]


Jarcho:
While there isn't an issue with using `Vec`, this really should be considered a bug in `std`.

@bors r+

Jarcho:
Uhmm... one more time. @bors r+

PR #10632, opened 2023-04-11
Add `needless_maybe_sized` lint

### LINKED ISSUE #10600: Warn on useless `?Sized` anti-bound
### What it does

Consider the following (admittedly silly) code:
```rust
fn id<T: Copy + ?Sized>(t: &T) -> &T {
    t
}
```
The author clearly _meant_ for this method to work on unsized `T`, but the `Copy` bound implicitly implies `Clone`, which implies `Sized`, so the compiler will deduce that `T` is `Sized`. Right now, it compiles and passes clippy silently. Instead, clippy should warn that the `?Sized` anti-bound is useless, and suggest removing one of the two bounds.

Here's a silly-but-slightly-less-so example:
```rust
unsafe fn deref<T: Copy + ?Sized>(ptr: *const T) -> T {
    *ptr
}
```
Here, someone might have this function without the `?Sized` anti-bound (`Copy` is required to move out of the pointer), realize they want to call it with an unsized type, and modify it like so. (In my case, I was trying to construct an example of `?Sized` and pointer dereferencing, and ended up surprised for a few minutes that this compiled.)

A similar case is #1463, i.e. bounds where one implies the other, for example `T: Copy + Clone`. That seems potentially less bad, because there's no issue with the code's intent; it's just an unnecessarily complex bound.

### Lint Name

useless_anti_bound

### Category

suspicious

### Advantage

- Clearer intent
- Removes useless code

### Drawbacks

None I can think of.

### Example

```rust
fn id<T: Copy + ?Sized>(t: &T) -> &T {
    t
}
```

Could be written as:

```rust
fn id<T: Copy>(t: &T) -> &T {
    t
}
```

Alexendoo:
changelog: new lint: [`needless_maybe_sized`]

Closes #10600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5523,6 +5523,7 @@ Released 2018-09-13
 [`needless_late_init`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_late_init
 [`needless_lifetimes`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_lifetimes
 [`needless_match`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_match
+[`needless_maybe_sized`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_maybe_sized
 [`needless_option_as_deref`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_option_as_deref
 [`needless_option_take`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_option_take
 [`needless_parens_on_range_literals`]: https://rust-lang.github.io/rust-clippy/master/index.html#needless_parens_on_range_literals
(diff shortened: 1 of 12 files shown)


Jarcho:
r? Jarcho

I'm not sure if the trace with spans is really worth it. It takes up a lot space and most of the info is not very useful. Is there a way to just print the location without the snippet?

One extension would be to check if any of the traits have functions using `Self` by value. e.g.

```rust
trait T {
  fn foo() -> Self;
}
```

There's no `Sized` bound on the trait, but it's also unimplementable for unsized types.

Alexendoo:
I don't think there is a way to keep the location while hiding the snippet, but we could make them regular notes

Jarcho:
Too bad there's no option for printing spans like that.

Jarcho:
Been a week without objections. Getting the spans printed nicely isn't a blocker and can be added in later.

@bors r+

J-ZhengLi:
getting blocked by `bors` for this long is kind of unfortunate.

@Jarcho, could you retry running the tests so this can get through?

PR #10638, opened 2023-04-12
Standardize lint formulation

### LINKED ISSUE #10631: Make lint formulation more consistent
### Description

When trying to figure out how I should formulate the [documentation](https://github.com/rust-lang/rust-clippy/blob/83e42a2337dadac915c956d125f1d69132f36425/clippy_lints/src/methods/mod.rs#L3196) for `clear_with_drain`, I had a look at how other lints are formulated, and I found a few different ways which are pretty much equivalent.

Here are the statistics I gathered:

| Formulation | Number of occurrences |
|---------------|---------------------------|
| "Checks for usage" | 59 |
| "Checks for use" | 16 |
| "Checks for usages" | 12 |
| "Checks for uses" | 6 |
| "Checks for using" | 3 |
| "Detect uses" | 2 |
| "Detects uses" | 2 |
| "Check for use" | 1 |
| "Detect use" | 1 |
| Total | 102 |

> Note: I used case-insensitive search and I made sure plural and singular forms were counted separately by using `[^s]` in the regexes for singular forms. Also, to double check, the Total is not just the sum of all the occurrences, it was obtained by a single call with a regex that matches all these cases (and more which I did not include because they had 0 occurrences).

These represent about 1/6 of the total number of lints, which is quite significant.

I propose to standardize the formulation, opting for the most frequent one, i.e. **Checks for usage**. Of course, I am open to picking a different one if a native English speaker (which I'm not) tells me that it is not the most idiomatic 🙂 

### Version

_No response_

### Additional Labels

@rustbot label +C-enhancement +A-documentation

bluthej:
Fixes #10631 

changelog: Standardize a very frequent lint formulation which had variations

### DIFF
--- a/clippy_lints/src/allow_attributes.rs
+++ b/clippy_lints/src/allow_attributes.rs
@@ -6,7 +6,7 @@ use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 
 declare_clippy_lint! {
-    /// Detects uses of the `#[allow]` attribute and suggests replacing it with
+    /// Checks for usage of the `#[allow]` attribute and suggests replacing it with
     /// the `#[expect]` (See [RFC 2383](https://rust-lang.github.io/rfcs/2383-lint-reasons.html))
     ///
     /// The expect attribute is still unstable and requires the `lint_reasons`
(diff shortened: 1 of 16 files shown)


Alexendoo (approved):

Alexendoo:
Thanks for going through them all!

@bors r+

bluthej:
No pb! Thanks for your approval and thank god for ripgrep which helped a lot along the way :)

PR #10622, opened 2023-04-10
Clippy Book Chapter Updates Reborn: Lint Passes

blyxyas:
This PR adds a new chapter to the book: "Lint passes". No major changes apart from some re-phrasing, fixing typos... etc.

## Notes

- Requires #10595 to be merged before this one (Or else, a link will be broken).
- To talk about the whole project, please use the tracking issue for the project #10597 (It also contains a timeline, discussions and more information)

changelog: Add a new "Lint passes" chapter to the book
r? @flip1995

PR #10595, opened 2023-04-04
Clippy Book Chapter Updates Reborn: Defining Lints

blyxyas:
Revival of #9659, I've made a few changes (mainly ones from reviews from #9426, like removing mid-section storytelling) and some re-writes. The goal of the project would be to slowly re-write the Clippy book (only chapters that need it) to more up-to-date style.

## Notes

- Things like `git status` commands have changed, we need to make sure that they're correct.
- As this is a team-wide effort, I would please ask anyone and everyone to read it and give your opinion.

---

changelog: Add a new "Defining lints" chapter to the book

r? @flip1995

PR #10644, opened 2023-04-13
Clippy Book Chapter Updates Reborn: Method Checking

blyxyas:
This PR adds a new chapter to the book: "Method Checking". Some major re-phrasing was done and some changes in the code comments (apart from grammar and minor changes).

## Notes

- Requires #10595 **and** #10622 to be merged before this, or else several links will be broken
- To talk about the whole project, please use the tracking issue for the project #10597 (It also contains a timeline, discussions and more information)

changelog: Add a new "Method Checking" chapter to the book.

r? @flip1995


book/src/development/method_checking.md
```diff
@@ -0,0 +1,93 @@
+# Method Checking
+
+In some scenarios we might want to check for methods when developing
+a lint. There are two kinds of questions that we might be curious about:
+
+- Invocation: Does an expression call a specific method?
+- Definition: Does the type `Ty` of an expression define a method?
```
flip1995, line 7: This says something different than what is then documented below. Below is documented on how to check an `impl` block. This is pretty trivial and doesn't really need documentation IMO. So I would leave this out.

I think usually, we don't need to check if a `Ty` implements a certain method, but rather if a `Ty` implements a trait. But there's a different chapter for this.

Let me know if you disagree.

book/src/development/method_checking.md
```diff
@@ -0,0 +1,93 @@
...
+use rustc_span::sym;
+
+impl<'tcx> LateLintPass<'tcx> for OurFancyMethodLint {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>) {
+        // Check our expr is calling a method with pattern matching
+        if let hir::ExprKind::MethodCall(path, _, [_self_arg, ..]) = &expr.kind
+            // Check if the name of this method is `our_fancy_method`
+            && path.ident.name == sym!(our_fancy_method)
+            // We can check the type of the self argument whenever necessary.
+			// (It's necessary if we want to check that method is specifically belonging to a specific trait,
```
flip1995, line 27: There is mixed tabs and spaces in this file. Also some other formatting issues, like missing backticks.

book/src/development/method_checking.md
```diff
@@ -0,0 +1,93 @@
...
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_span::sym;
+
+impl<'tcx> LateLintPass<'tcx> for OurFancyMethodLint {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx hir::Expr<'_>) {
+        // Check our expr is calling a method with pattern matching
+        if let hir::ExprKind::MethodCall(path, _, [_self_arg, ..]) = &expr.kind
+            // Check if the name of this method is `our_fancy_method`
+            && path.ident.name == sym!(our_fancy_method)
+            // We can check the type of the self argument whenever necessary.
```
flip1995, line 26: More often then not we want to do this. If we don't this is a major cause of false positives. So I would include that part here.

flip1995 (commented):
Most of it LGTM. Let's fix/discuss the rest.

book/src/development/method_checking.md
blyxyas: I think we can implement both checking if a method call calls a specific method **and** if the type implements a trait in the same example (The comments explain it). And center the other one about checking an `impl` instead of a `Ty` (Changing the wording, keeping the example)

blyxyas:
I think the latest commit fixes the issues shown in the reviews.

flip1995 (approved):
LGTM.

Rebased and formatted.

I'm concerned that PRs for the new chapters might get blocked by our link checker, because those new chapters that are linked in here are not yet upstream. But in the meantime we can just replace them with `FIXME`s.

If anything should fail during merge, feel free to `r=me` after swapping out the links with `FIXME`s.

PR #10647, opened 2023-04-15
new lint: `manual_while_let_some`

y21:
This PR implements the lint I suggested [on zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/lint.20on.20while.20pop.20unwrap).
It looks for while loops like these:
```rs
let mut numbers = vec![0, 1, 2];
while !numbers.is_empty() {
  let number = numbers.pop().unwrap();
  // use `number`
}
```
and suggests replacing it with a while-let loop, like this:
```rs
let mut numbers = vec![0, 1, 2];
while let Some(number) = numbers.pop() {
  // use `number`
}
```
... which is more concise and idiomatic.

It only looks for `Vec::pop()` calls in the first statement of the loop body in an attempt to not trigger FPs (as pop might only be called conditionally).

changelog: new lint [`manual_while_let_some`]


clippy_lints/src/while_pop_unwrap.rs
```diff
@@ -0,0 +1,122 @@
...
+declare_lint_pass!(WhilePopUnwrap => [WHILE_POP_UNWRAP]);
+
+fn report_lint(cx: &LateContext<'_>, pop_span: Span, ident: Option<Ident>, loop_span: Span, receiver_span: Span) {
+    span_lint_and_then(
+        cx,
+        WHILE_POP_UNWRAP,
+        pop_span,
+        "you seem to be trying to pop elements from a `Vec` in a loop",
+        |diag| {
+            diag.span_suggestion(
```
llogiq, line 46: Usually, suggestions should contain all changes. To do this right you could use [`clippy_utils::multispan_sugg`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/diagnostics/fn.multispan_sugg.html), use the `while let Some(..)` message for the `while`loop header and an empty `String` for the pop+unwrapping statement.

There is also a `_with_applicability` variant you could use to make the lint auto-applicable if the snippets are from valid Rust code.

That said, as a style lint, it should probably not apply to macro-generated code. You can check `expr.span.from_expansion()` to rule that out.

clippy_lints/src/while_pop_unwrap.rs
```diff
@@ -0,0 +1,122 @@
...
+        },
+    );
+}
+
+fn match_method_call(cx: &LateContext<'_>, expr: &Expr<'_>, method: &[&str]) -> bool {
+    if let ExprKind::MethodCall(..) = expr.kind
+        && let Some(id) = cx.typeck_results().type_dependent_def_id(expr.hir_id)
+        && match_def_path(cx, id, method)
+    {
+        true
```
llogiq, line 66:
```suggestion
    {
        match_def_path(cx, id, method)
```

clippy_lints/src/while_pop_unwrap.rs
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+        false
+    }
+}
+
+fn is_vec_pop(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
```
llogiq, line 72: This method can probably be inlined, it's only used once, and doesn't have much documentation value.

tests/ui/while_pop_unwrap.rs
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    // This should also not warn. It conditionally pops elements.
+    while !numbers.is_empty() {
+        if false {
+            continue;
+        }
+        accept_i32(numbers.pop().unwrap());
+    }
```
llogiq, line 46: I'd like to see a test that checks for `is_empty()` on a different `Vec`. Also one that `pop()`s, but doesn't `unwrap()`, and one macro-generated (can be with a local macro).

llogiq (changes requested):
This is a good starting point! :+1:

For a `style` lint, I'd like to have a real working suggestion. Otherwise I'd be fine with putting the lint in `nursery` for the time being, and work on it in follow-up PRs. Otherwise a few more tests would be great.

tests/ui/while_pop_unwrap.rs
y21: Oh that's a good point. At the moment this lint doesn't actually check if `Vec::pop()` and `Vec::is_empty()` were called on the same `Vec`, but I'm also not sure how I'd check for that. Would it make sense to just compare the two `Expr`s (`SpanlessEq::eq_expr`?) or is there a better way?

llogiq: Yeah, `SpanlessEq` is the way to go there.

clippy_lints/src/while_pop_unwrap.rs
y21: > That said, as a style lint, it should probably not apply to macro-generated code. You can check expr.span.from_expansion() to rule that out.

The `check()` function is called by the `Loops` pass, which already checks if this is macro-expanded code if I understand it right.

https://github.com/rust-lang/rust-clippy/blob/203c909d04ffc0b2c22f97cb304649aac237087e/clippy_lints/src/loops/mod.rs#L622-L625

llogiq:
I like it! :+1: The only thing I think we might improve is the name. `while_pop_unwrap` looks a bit out of place as a lint name. How about `manual_while_let_some` (we have a good number of `manual_*` lints already)?

y21:
I agree, `manual_while_let_some` sounds better. Changing.

llogiq:
Thank you!

@bors r+

PR #10651, opened 2023-04-16
Add size-parameter to unecessary_box_returns

### LINKED ISSUE #10641: `unnecessary_box_returns` don't lint on large objects
### Description

A lot of the reason you may `Box` something is if it is too large to move cheaply. For large types it wouldn't be very unusual to return a `Box<T>`, especially if the box existed before returning it (as in the example here) to prevent some unnecessary moves or reallocations.

It could be a good idea to not raise this lint if T is e.g. >64 bytes (or reuse the threashold for suggesting boxed enum variants).

Some relevant zulip discussion: https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/returning.20boxed.20values

```rust
#![allow(unused)]
#![warn(clippy::pedantic)]

struct Huge([u8; 500]);
struct HasHuge(Box<Huge> /* ... */);

impl HasHuge {
  fn into_huge(self) -> Box<Huge> {
    self.0
  }
}
```

### Version

```text
rustc 1.68.2 stable
```

### Additional Labels

@rustbot label +C-enhancement

lukaslueg:
Fixes #10641

This adds a configuration-knob to the `unecessary_box_returns`-lint which allows _not_ linting a `fn() -> Box<T>` if `T` is "large". The default byte size above which we no longer lint is 128 bytes (due to https://github.com/rust-lang/rust-clippy/issues/4652#issue-505670554, also used in #9373). The overall rational is given in #10641.

changelog: [`unnecessary_box_returns`]: Don't lint if `T` in `Box<T>` is large


clippy_lints/src/utils/conf.rs
```diff
@@ -463,6 +463,10 @@ define_Conf! {
     ///
     /// The maximum byte size a `Future` can have, before it triggers the `clippy::large_futures` lint
     (future_size_threshold: u64 = 16 * 1024),
+    /// Lint: UNNECESSARY_BOX_RETURNS,
```
xFrednet, line 466:
```suggestion
    /// Lint: UNNECESSARY_BOX_RETURNS.
```

This list requires a dot at the end, to be parsed correctly. Once you do this, you also have to run `cargo collect-matadata` to update the lint configuration list in our book.

xFrednet (approved):
LGTM, tiny nit related to the lint documentation comment. It requires some specific syntax. :)

xFrednet:
You can `r=me` (meaning `bors r=xFrednet` with the at)

xFrednet:
You can `r=me` (meaning `bors r=xFrednet` with an `@` in front)

@bors delegate+

PR #10653, opened 2023-04-16
Book: Improve type checking and lint passes chapters

flip1995:
- Adds the lint passes chapter to the index.
- Moves additional reading material for beginners to the front page of the development section.
- Clarify some details in `hir::Ty` vs `ty::Ty`.

@blyxyas I started reviewing the chapters that were already merged. Great work. I liked the additional reading chapter so much, that I moved it front and center. Other than that I only had some minor corrections to make. Please review.

r? @llogiq 

changelog: none

### DIFF
--- a/book/src/development/type_checking.md
+++ b/book/src/development/type_checking.md
@@ -119,8 +119,9 @@ an `u32`. As far as `hir::Ty` is concerned those might be different types. But a
 understands that they're the same type, in-depth lifetimes, etc...
 
 To get from a `hir::Ty` to a `ty::Ty`, you can use the [`hir_ty_to_ty`][hir_ty_to_ty] function outside of bodies or
-outside of bodies the [`TypeckResults::node_type()`][node_type] method. Don't use `hir_ty_to_ty` inside of bodies,
-because this can cause ICEs.
+outside of bodies the [`TypeckResults::node_type()`][node_type] method.
+
+> **Warning**: Don't use `hir_ty_to_ty` inside of bodies, because this can cause ICEs.
 
 ## Useful Links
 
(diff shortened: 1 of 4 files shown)


book/src/development/type_checking.md
```diff
@@ -107,15 +107,20 @@ impl LateLintPass<'_> for MyStructLint {
...
+```
 
 Here the HIR sees the types without "thinking" about them, it knows that the function takes an `u32` and returns
-an `u32`. But at the `ty::Ty` level the compiler understands that they're the same type, in-depth lifetimes, etc...
+an `u32`. As far as `hir::Ty` is concerned those might be different types. But at the `ty::Ty` level the compiler
+understands that they're the same type, in-depth lifetimes, etc...
 
-you can use the [`hir_ty_to_ty`][hir_ty_to_ty] function to convert from a `hir::Ty` to a `ty::Ty`
+To get from a `hir::Ty` to a `ty::Ty`, you can use the [`hir_ty_to_ty`][hir_ty_to_ty] function outside of bodies or
+outside of bodies the [`TypeckResults::node_type()`][node_type] method. Don't use `hir_ty_to_ty` inside of bodies,
```
llogiq, line 122: I think that would warrant it's own warning ⚠️ symbol.

llogiq (approved):
Looks good, solid improvements.

I'd put the warning into its own warning box.

llogiq:
Cool stuff!

@bors r+

blyxyas:
> Great work

Thanks, but like 90% of this PRs are from @nahuakang, I'm just fixing some typos and actively fixing reviews and going through the PR process. The only full paragraph that I've written was "phrased weirdly" 😅

PR #10656, opened 2023-04-17
Add configuration for `semicolon_block` lints

### LINKED ISSUE #10654: Add new lint for the location of a semicolon depending on whether its block is singleline or multiline
### What it does

This is the same as both semicolon-inside-block and semicolon-outside-block, but decides which one to use based on whether the block is singleline or multiline.

### Lint Name

semicolon-inside-block-if-multiline or semicolon-outside-block-if-singleline (I prefer this one)

### Category

restriction

### Advantage

* rustfmt will make an unsafe block containing only one expression multiline if the semicolon is inside the block, but won't if it's on the outside. Adding this lint will allow swapping the location to allow rustfmt to make it a singleline, while also encouraging some consistency.
* Some may find outside cleaner only in the case that it's singleline.

### Drawbacks

* The descriptions of both semicolon-inside-block and semicolon-outside-block suggest consistency, so this may contradict that.
* If rustfmt is invoked, what was previously fine could become a warning (or an error if a library author is a bit zealous about warnings). This may be unexpected.

### Example

```rust
    unsafe {
        WriteProcessMemory(
            self.inner,
            base as _,
            bytes.as_ptr() as _,
            bytes.len(),
            None,
        )
    };

    unsafe {
        ExitProcess(0i32);
    }
```

Could be written as:

```rust
    unsafe {
        WriteProcessMemory(
            self.inner,
            base as _,
            bytes.as_ptr() as _,
            bytes.len(),
            None,
        );
    }

    unsafe { ExitProcess(0i32) };
```

Centri3:
Does exactly what it says on the tin, suggests moving a block's final semicolon inside if it's multiline and outside if it's singleline.

I don't really like how this is implemented so I'm not too sure if this is ready yet. Alas, it might be ok.

---

fixes #10654

changelog: Enhancement: [`semicolon_inside_block`]: Added `semicolon-inside-block-ignore-singleline` as a new config value.
[#10656](https://github.com/rust-lang/rust-clippy/pull/10656)
changelog: Enhancement: [`semicolon_outside_block`]: Added `semicolon-outside-block-ignore-multiline` as a new config value.
[#10656](https://github.com/rust-lang/rust-clippy/pull/10656)


xFrednet:
Hey! Code written in the half-asleep state can be quite genius. At least there are quite a few PRs I created in that state which weren't too bad ^^. I planned to do a review of this on Wednesday, is that fine with you, or should I wait on the planned rewrite :)

Centri3:
> Hey! Code written in the half-asleep state can be quite genius. At least there are quite a few PRs I created in that state which weren't too bad ^^. I planned to do a review of this on Wednesday, is that fine with you, or should I wait on the planned rewrite :)

Wednesday sounds good, I think I can get it done by then. Thanks!

clippy_lints/src/semicolon_block.rs
```diff
@@ -135,3 +183,45 @@ fn semicolon_outside_block(cx: &LateContext<'_>, block: &Block<'_>, tail_stmt_ex
...
+        );
+    }
+}
+
+fn get_line(cx: &LateContext<'_>, span: Span) -> usize {
+    cx.sess()
+        .source_map()
+        .lookup_line(span.lo())
+        .expect("failed to get span's line")
+        .line
```
xFrednet, line 226: This `expect` looks a bit suspicious, I believe that there are some error cases for macros and there might be others. Generally speaking, it's advisable to handle errors to avoid panics. I think the simplest way would be to simply to not lint if this returns an error.

It can be okay to use `expect()` in clippy, but in that case, it should be explained, why this is safe. The expect doesn't seem to do that in this case.

clippy_lints/src/declared_lints.rs
```diff
@@ -546,6 +546,7 @@ pub(crate) static LINTS: &[&crate::LintInfo] = &[
     crate::self_named_constructors::SELF_NAMED_CONSTRUCTORS_INFO,
     crate::semicolon_block::SEMICOLON_INSIDE_BLOCK_INFO,
     crate::semicolon_block::SEMICOLON_OUTSIDE_BLOCK_INFO,
+    crate::semicolon_block::SEMICOLON_OUTSIDE_BLOCK_IF_SINGLELINE_INFO,
```
xFrednet, line 549: I feel like this shouldn't be a new lint, but much rather a configuration for `semicolon_outside_block`. When I see `#[warn(clippy::semicolon_outside_block_if_singleline)]` I would expect it to only lint single line cases, and not also multiline ones.


Now with a configuration, it would look something like this:

```
// main.rs
#![warn(clippy::semicolon_inside_block)]

// clippy.toml
semicolon-outside-block-for-single-line = true
```

Which is more effort, but that should be fine with a `restriction` lint. This way, it's clear that semicolons are usually preferred inside the block, unless it's a single line. The config name might need some improvement, but that would feel a bit better to me. What do you think? Also, because you suggested the lint in the first case? :)

Here is some documentation, if you want to implement a configuration value instead: [Adding configuration to a lint](https://doc.rust-lang.org/clippy/development/adding_lints.html#adding-configuration-to-a-lint)

xFrednet (commented):
The implementation overall looks good to me. I left two comments, where I would like your feedback on.

Also, because I didn't say that earlier: Welcome to Clippy :wave: :paperclip:

clippy_lints/src/declared_lints.rs
Centri3: I've added that plus another for semicolon_inside_block, for consistency. with both of them enabled it's the same behavior as the original lint.

clippy_lints/src/semicolon_block.rs
Centri3: Wasn't sure what to do here, it doesn't lint over panicking now.

xFrednet (commented):
This generally looks good to me, some nits and a few questions to make sure, that I understand everything correctly :)

clippy_lints/src/semicolon_block.rs
```diff
@@ -83,21 +98,33 @@ impl LateLintPass<'_> for SemicolonBlock {
...
+                semicolon_inside_block(self, cx, block, tail, stmt.span);
+            },
             _ => (),
         }
     }
 }
 
-fn semicolon_inside_block(cx: &LateContext<'_>, block: &Block<'_>, tail: &Expr<'_>, semi_span: Span) {
+fn semicolon_inside_block(
+    conf: &mut SemicolonBlock,
```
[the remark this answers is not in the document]
Centri3: nope, I can't believe I wrote that :sweat_smile:

I moved them into an impl block.

clippy_lints/src/utils/conf.rs
```diff
@@ -277,6 +277,14 @@ define_Conf! {
...
     /// default configuration of Clippy. By default, any configuration will replace the default value.
     (disallowed_names: Vec<String> = super::DEFAULT_DISALLOWED_NAMES.iter().map(ToString::to_string).collect()),
+    /// Lint: SEMICOLON_INSIDE_BLOCK.
+    ///
+    /// Whether to lint only if it's multiline.
+    (semicolon_inside_block_if_multiline: bool = false),
+    /// Lint: SEMICOLON_OUTSIDE_BLOCK.
+    ///
+    /// Whether to lint only if it's singleline.
+    (semicolon_outside_block_if_singleline: bool = false),
```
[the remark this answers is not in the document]
Centri3: I changed the two to `semicolon_inside_block_ignore_singleline` and `semicolon_outside_block_ignore_multiline` since I agree, the fact that the names of both describe the effect is a bit weird.

You do understand the configuration - it ignores it if it's singleline/multiline. I thought to add two instead of only one as maybe someone may wish to always have a semicolon on the outside, but only inside if it's multiline, or some combination of the two.

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update and addressing all my review comments. I hope you had fun working on Clippy :D

@bors r+

Centri3:
whoops, forgot to run cargo `collect-metadata` :sweat_smile: I think you'll need to reapprove this, @xFrednet

PR #10664, opened 2023-04-18
Don't apply `no_effect_underscore_binding` on `__`

Centri3:
`__` is sometimes used to indicate "this is an implementation detail, don't use this", but this lint is incorrectly triggered on such code.

changelog: [`no_effect_underscore_binding`]: Don't apply if there are 2 or more leading underscores


Centri3:
I've just realized this is only emitted on bindings, not items (duh..), sorry about that. This isn't needed

PR #10665, opened 2023-04-18
Don't apply `string_lit_as_bytes` if in macro expansion

Centri3:
The following code will emit a warning on both w! and h!, despite there being nothing the user (or library author) could do about it:
```rust
#![warn(clippy::string_lit_as_bytes)]

use windows::w;
use windows::h;

fn main() {
    let _w = w!("example");
    let _h = h!("example");
}
```
This is because windows-rs will create a binding `const INPUT: &[u8] = $s.as_bytes()`, and changing this to b"$s" is, well, suboptimal. I don't know enough about Rust to know if this is something that can be detected though if it can be I'm happy with closing this in favor of implementing that.

I'm not sure whether this is how it should be done though, as this simply tells clippy to not invoke this even if it's applicable (this also affects the other string lints, but didn't cause any tests to fail).

changelog: [`string_lit_as_bytes`]: Don't lint if in external macro

### DIFF
--- a/clippy_lints/src/strings.rs
+++ b/clippy_lints/src/strings.rs
@@ -253,6 +253,10 @@ impl<'tcx> LateLintPass<'tcx> for StringLitAsBytes {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) {
         use rustc_ast::LitKind;
 
+        if e.span.from_expansion() {
+            return;
+        }
+
         if_chain! {
             // Find std::str::converts::from_utf8
             if let Some(args) = match_function_call(cx, e, &paths::STR_FROM_UTF8);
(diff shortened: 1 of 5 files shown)


giraffate (commented):
Thanks for creating the pull request! I made some comments. If you have any questions, feel free to ask us.

- It would be better to ignore it only in an external macro, so this would be helpful for it: https://github.com/rust-lang/rust-clippy/pull/10502
- It looks like that this fix affects the `string_from_utf8_as_bytes` lint. At first, it would be better to fix only `string_lit_as_bytes`.

Centri3:
> Thanks for creating the pull request! I made some comments. If you have any questions, feel free to ask us.
> 
>     * It would be better to ignore it only in an external macro, so this would be helpful for it: [fix `almost_swapped`: Ignore external macros #10502](https://github.com/rust-lang/rust-clippy/pull/10502)
> 
>     * It looks like that this fix affects the `string_from_utf8_as_bytes` lint. At first, it would be better to fix only `string_lit_as_bytes`.

Thanks, I wasn't sure how to do that. Done

giraffate (approved):
Overall looks good, thanks!

We follow the no merge-commit policy, so can you do the rebase the `master` branch, not merge? Once this is resolved, I'll merge this.
https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/basics.md#pr

giraffate (approved):

PR #10669, opened 2023-04-19
Update to a compiletest-rs version that requires `//@` for commands

oli-obk:
Requires https://github.com/Manishearth/compiletest-rs/pull/261 to get published

This PR is a smaller step towards https://github.com/rust-lang/rust-clippy/pull/10426

changelog: Move to a version of compiletest-rs that allows us to require `//@` for test suite commands.


oli-obk:
Restarting CI

oli-obk:
This works now and depends only on the latest crates.io release of compiletest-rs

tests/ui/wildcard_imports_2021.stderr
```diff
@@ -1,132 +1,73 @@
-error: usage of wildcard import
-  --> $DIR/wildcard_imports_2021.rs:9:5
+error[E0463]: can't find crate for `wildcard_imports_helper`
```
Alexendoo, line 1: This one was real confusing, it looks like it's the 2021 revision of the test but it's not, the test file has 2021 in its name **and** two edition revisions

The `.stderr` must have been left behind from a previous version that didn't have revisions making it trickier by showing a diff instead of a new file creation

oli-obk:
I checked that this PR does not add any new files and deleted the preexisting stray .stderr file.

The other review comments have also been addressed by reverting the changes to what they are on master.

tests/ui/asm_syntax.rs
```diff
@@ -1,5 +1,5 @@
-// only-x86_64
-// ignore-aarch64
+//@only-target-x86_64
+//@ignore-target-aarch64
```
Alexendoo, line 2: `-target` shouldn't be added just yet, there's a few cases of that

oli-obk:
Sorry about that back and forth. This time I went through all 400 changed .rs files and removed all but the `//@` changes (except in one or two cases where command flags were just wrong)

oli-obk:
> p.s. checking out a branch called `@` was a nice little puzzle

I will strive to pick a harder branch name next time

PR #10672, opened 2023-04-20
WIP: `excessive_width` and `excessive_indentation` lints

Centri3:
Not done yet, but looking for feedback. I'm not sure how to go about implementing excessive_indentation, but maybe doing the opposite of excessive_width? Get the start of the statement's span and counting characters?

Anyway, I'm thinking maybe excessive_width is emitted wherever rustfmt could've failed (so stuff like use_small_heuristics and such are taken into account), but I don't know how to implement this without duplicating every one of its width-related settings (and also whether it's necessary, since 100 is already a pretty generous threshold)

changelog: new lints [`excessive_width`] and [`excessive_indentation`]


Centri3:
ok, I see an issue with this. local bindings seem to be checked as well even if they're a block, which makes sense. this means, even if the actual code never reaches above column 100, it'll still be triggered if the length of the entire block is above 100. Not ideal... I'll check back on here later with hopefully the knowledge needed to eliminate that.

Alexendoo:
Although it's unstable rustfmt does have an option to error on lines that go above the `max_width` - [`error_on_line_overflow`](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#error_on_line_overflow). We want to avoid doing things that are/could be covered by rustfmt

For indentation that does seem like something more in our area, at least I wouldn't expect for rustfmt to add an option for that. Does that sound reasonable @flip1995?

Centri3:
Ok, I think this is ready! @Alexendoo, feel free to review this whenever you can and once consensus on whether this should be added has been reached :slightly_smiling_face: the current implementation is a bit verbose but I'm not sure if it's possible to get it better.

I'll also write tests and the description later, once I know whether this should be added.

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,329 @@
...
+            {
+                self.nest_level += 1;
+
+                check_indent(self, *inner_span);
+
+                self.nest_level -= 1;
+            }
+            ItemKind::Trait(trit) => check_trait_and_impl(self, item, &trit.items),
+            ItemKind::Impl(imp) => check_trait_and_impl(self, item, &imp.items),
+            _ => (),
```
Alexendoo, line 291: We can use `walk_item` here & merge the `impl`/`trait`/`mod foo { .. }` cases

```suggestion
            ItemKind::Trait(_) | ItemKind::Impl(_) | ItemKind::Mod(.., ModKind::Loaded(_, Inline::Yes, _)) => {
                self.nest_level += 1;

                if !check_indent(self, item.span) {
                    walk_item(self, item);
                }

                self.nest_level -= 1;
            },
            // Don't visit non-inline modules
            ItemKind::Mod(..) => {},
            _ => walk_item(self, item),
```

Alexendoo (commented):
👋

Centri3:
I'll apply these changes tomorrow :slightly_smiling_face: thanks! Glad to see this could've been done better

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,329 @@
...
+impl EarlyLintPass for ExcessiveNesting {
+    fn check_item(&mut self, cx: &EarlyContext<'_>, item: &Item) {
+        let conf = self;
+        NestingVisitor {
+            conf,
+            cx,
+            nest_level: 0,
+        }
+        .visit_item(item);
+    }
```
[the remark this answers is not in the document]
Centri3: I noticed this while testing, hence the note under "Known issues" (though I thought it was only specific to modules). At first, I was thinking of using `check_crate` instead but thought it wasn't necessary since all it was doing was calling `visit_item` for each item anyway, but I'll change it to that (plus `Inline::No`) once I have the chance. Thanks

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        match &local.kind {
+            LocalKind::Init(expr) => self.visit_expr(expr),
+            LocalKind::InitElse(expr, block) => {
+                self.visit_expr(expr);
+                self.visit_block(block);
+            },
+            LocalKind::Decl => (),
+        }
+    }
```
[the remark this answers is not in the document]
Centri3: Fixed

clippy_lints/src/excessive_nesting.rs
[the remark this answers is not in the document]
Centri3: Fixed, also removed the known issues section since it doesn't apply anymore

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    fn visit_block(&mut self, block: &Block) {
+        // TODO: Can we use some RAII guard instead? Borrow checker seems to hate that
+        // idea but it would be a lot cleaner.
+        self.nest_level += 1;
+
+        if !check_indent(self, block.span) {
+            for stmt in &block.stmts {
+                self.visit_stmt(stmt);
+            }
```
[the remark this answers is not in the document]
Centri3: Fixed, I should've read the module's documentation. It just has `visit_block` and `visit_item` now

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,136 @@
...
+
+                if !self.check_indent(item.span) {
+                    walk_item(self, item);
+                }
+
+                self.nest_level -= 1;
+            },
+            // Mod: Don't visit non-inline modules
+            // ForeignMod: I don't think this is necessary, but just incase let's not take any chances (don't want to
+            // cause any false positives)
```
Alexendoo, line 131: Yeah it's not needed, despite the name `ForeignMod` refers to `extern { .. }` blocks

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,136 @@
...
+            ItemKind::Mod(..) | ItemKind::ForeignMod(..) => {},
```
Alexendoo, line 132: Here we'll want to create a new visitor with the nesting level set to 0 and call `.visit_item(item)` on it so that we pick up nesting in other files than the entry point

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,136 @@
...
+            return true;
+        }
+
+        false
+    }
+}
+
+impl<'conf, 'cx> Visitor<'_> for NestingVisitor<'conf, 'cx> {
+    fn visit_block(&mut self, block: &Block) {
+        self.nest_level += 1;
```
Alexendoo, line 109: Here and in `visit_item` I'd say we should ignore macro expansions when incrementing/decrementing the nesting level, how to handle them at a syntax level is a difficult problem, e.g.

```rust
macro_rules! a { () => {{{ 1 }}} }
macro_rules! b { () => {{{{{{{ a!() }}}}}}} }

b!(); // currently this will cause `1` to be linted with a depth of 8ish
```

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,136 @@
...
+};
+use rustc_lint::{EarlyContext, EarlyLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// Checks for blocks which are indented beyond a certain threshold.
```
Alexendoo, line 14:
```suggestion
    /// Checks for blocks which are nested beyond a certain threshold.
```

clippy_lints/src/excessive_nesting.rs
Alexendoo: You can add a test for this by adding `auxiliary/mod.rs` and having `mod auxiliary` in the main test file

Centri3: Fixed

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,150 @@
...
+            return true;
+        }
+
+        false
+    }
+}
+
+impl<'conf, 'cx> Visitor<'_> for NestingVisitor<'conf, 'cx> {
+    fn visit_block(&mut self, block: &Block) {
+        // TODO: Probably not necessary, since any block would already be ignored by the check in visit_item
```
Alexendoo, line 109: This is needed, as e.g. `let x = some_macro!()` where `some_macro` expands to some block wouldn't have any item to visit

tests/ui-toml/excessive_nesting/excessive_nesting.rs
Alexendoo, line 1: We could do something with revisions here to be sure the depth tracking is working as intended, if you make all the places getting linted have the same depth, say 4 or 5, then you can have one config file of that depth and another below it, e.g.

```toml 
# below/clippy.conf
excessive-nesting-threshold = 3
```

```toml
# above/clippy.conf
excessive-nesting-threshold = 4 # or higher if needed
```

Then in the main test file

```rust
//@revisions: below above
//@[below] rustc-env:CLIPPY_CONF_DIR=below
//@[above] rustc-env:CLIPPY_CONF_DIR=above
```

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn check_indent(&self, span: Span) -> bool {
+        if self.nest_level > self.conf.excessive_nesting_threshold && !in_external_macro(self.cx.sess(), span) {
+            span_lint_and_help(
+                self.cx,
+                EXCESSIVE_NESTING,
+                span,
+                "this block is too nested",
+                None,
+                "try refactoring your code to minimize nesting",
+            );
```
Alexendoo, line 98: It's a bit annoying but in early lints we shouldn't `span_lint` directly from a visitor as it won't take into account lint attrributes (`#[allow/warn/etc]`) properly

`ExcessiveNesting` could store a [`NodeSet`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_ast/node_id/type.NodeSet.html) and hand a `&mut NodeSet` to `NestingVisitor` to populate rather than linting immediately, then it could `span_lint` in its `EarlyLintPass` impls of `check_block`/`check_item` if the node is in that set

Alexendoo: We'd also want a test of `#[allow(clippy::excessive_nesting)] { some nested thing }`

tests/ui-toml/excessive_nesting/excessive_nesting.rs
Centri3: done, though I did have to specify the path relative to the manifest instead of the test file otherwise it wouldn't find it.

clippy_lints/src/excessive_nesting.rs
Centri3: fixed, and added that test. seems to work

I did add initialization for `nodes` where `ExcessiveNesting` is constructed instead of in `check_crate`, is this how it should be done or should I wrap it in `Option` instead?

Alexendoo: That's good yeah, it's a fancy `HashSet` wrapper ultimately so no need to stick it in an `Option`

tests/ui-toml/excessive_nesting/default/clippy.toml
```diff
@@ -0,0 +1 @@
+excessive-nesting-threshold = 0
```
Alexendoo, line 1: It would be good if this were 1 higher than whatever the other file ends up as rather than `0` (effectively every other test in the repo checks that nothing lints on `0` already)

Then if most of the tests only trigger the lower number you can visually scan the file to check if all the indents are at the right number

clippy_lints/src/excessive_nesting.rs
```diff
@@ -115,7 +114,9 @@ struct NestingVisitor<'conf, 'cx> {
 
 impl NestingVisitor<'_, '_> {
     fn check_indent(&mut self, span: Span, id: NodeId) -> bool {
-        if self.nest_level > self.conf.excessive_nesting_threshold && !in_external_macro(self.cx.sess(), span) {
+        let threshold = self.conf.excessive_nesting_threshold;
+
+        if threshold != 0 && self.nest_level > threshold && !in_external_macro(self.cx.sess(), span) {
```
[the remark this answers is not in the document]
Centri3: Fixed

tests/ui-toml/excessive_nesting/auxiliary/proc_macros.rs
```diff
@@ -0,0 +1,476 @@
+//@compile-flags: --emit=link
+//@no-prefer-dynamic
+
+// NOTE: Copied from `ui/auxiliary/proc_macros.rs`, couldn't get `../` to work for some reason
```
Alexendoo, line 4: The path starts at `excessive_nesting/auxiliary`, and `excessive_nesting/auxiliary/..` is considered not to exist on linux (but does exist on windows where paths are first resolved lexically)

A copy is unfortunate but it's something we can look to fix separately

tests/ui-toml/excessive_nesting/excessive_nesting.rs
```diff
@@ -0,0 +1,199 @@
+//@aux-build:proc_macros.rs
+//@revisions: set above
+//@[set] rustc-env:CLIPPY_CONF_DIR=tests/ui-toml/excessive_nesting/set
+//@[above] rustc-env:CLIPPY_CONF_DIR=tests/ui-toml/excessive_nesting/above
```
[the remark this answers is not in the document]
Centri3: Alright, I did check it though and it seems everything works as expected

clippy_lints/src/excessive_nesting.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+impl<'conf, 'cx> Visitor<'_> for NestingVisitor<'conf, 'cx> {
+    fn visit_block(&mut self, block: &Block) {
+        if block.span.from_expansion() {
+            return;
+        }
+
+        // TODO: This should be rewritten using `LateLintPass` so we can use `is_from_proc_macro` instead,
+        // but for now, this is fine.
```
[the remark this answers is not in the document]
Centri3: > maybe that late pass mods don't store whether they're inline or not

Yeah I believe this is it, though we could likely check `inner_span` afaik to see if the file is the same, unless that isn't stored?. This should probably be done in the future ("someday"), though.

Alexendoo (approved):

Alexendoo:
Thanks! @bors r+

PR #10675, opened 2023-04-20
[arithmetic_side_effects] Cache symbols

c410-f3r:
An internal-only modification to speed up the processing of symbols because "intern" isn't very cheap, even more when you are doing the same thing for every method expression.

changelog: none


llogiq:
Ah, we missed the changelog line. @bors retry

klensy:
Isn't it simple to add that strings to symbol table, and that just filter on symbols, skipping str<->String comparison/conversions?

c410-f3r:
That is what I asked at https://github.com/rust-lang/rust-clippy/pull/10615#discussion_r1161276730

IIRC, it is not rare to see PRs in `rust-last/rust` trimming pre-registered symbols and Clippy already has a bunch of interns, therefore, caching seemed to be the natural choice.

PR #10679, opened 2023-04-20
use `is_inside_const_context` for `in_constant` util fn

### LINKED ISSUE #10452: FP bool-to-int-with-if in const context
### Summary

.

### Lint Name

bool-to-int-with-if

### Reproducer

I tried this code:

```rust
fn main() {
    let _ = [0; { 
         if false { 0 } else { 1 }
    }];
}
```

I saw this happen:

```
 --> feature-gate-const-if-match.rs:3:10
  |
3 |          if false { 0 } else { 1 }
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace with from: `usize::from(!false)`
```
The problem here is that the lint probably just checks "am I inside a const fn" but this is some which is false, BUT we are inside an array which still is some sort of const context and the lint misses this.
So the `if false { 0 } else { 1 }` is probably const-eval'd but this does not work with 

````rust
fn main() {
    let _ = [0; { 
         usize::from(!false)
    }];
}
````
```
error[E0015]: cannot call non-const fn `<usize as std::convert::From<bool>>::from` in constants
 --> feature-gate-const-if-match.rs:3:10
  |
3 |          usize::from(!false)
  |          ^^^^^^^^^^^^^^^^^^^
  |
  = note: calls in constants are limited to constant functions, tuple structs and tuple variants
  = help: add `#![feature(const_trait_impl)]` to the crate attributes to enable
````

### Version

```text
binary: rustc
commit-date: 2023-03-03
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

y21:
Fixes #10452.

This PR improves the `in_constant` util function to detect more cases of const contexts. Previously this function would not detect cases like expressions in array length position or expression in an inline const block `const { .. }`.

changelog: [`bool_to_int_with_if`]: recognize array length operand as being in a const context and don't suggest `usize::from` there

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -86,10 +86,10 @@ use rustc_hir::hir_id::{HirIdMap, HirIdSet};
 use rustc_hir::intravisit::{walk_expr, FnKind, Visitor};
 use rustc_hir::LangItem::{OptionNone, ResultErr, ResultOk};
 use rustc_hir::{
-    self as hir, def, Arm, ArrayLen, BindingAnnotation, Block, BlockCheckMode, Body, Closure, Constness, Destination,
-    Expr, ExprKind, FnDecl, HirId, Impl, ImplItem, ImplItemKind, ImplItemRef, IsAsync, Item, ItemKind, LangItem, Local,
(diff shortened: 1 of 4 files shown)


Jarcho:
Nice that there's a function in rustc for that now.

Thank you. @bors r+

PR #10683, opened 2023-04-21
Fix false positive in `allow_attributes`

Centri3:
This would emit a warning if used in a proc-macro with the feature `lint_reasons` enabled. This is now fixed.

changelog: [`allow_attributes`]: Don't lint if in external macro

### DIFF
--- a/tests/ui/allow_attributes_false_positive.rs
+++ b/tests/ui/allow_attributes_false_positive.rs
@@ -1,4 +1,3 @@
-#![allow(unused)]
 #![warn(clippy::allow_attributes)]
 #![feature(lint_reasons)]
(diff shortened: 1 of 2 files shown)


tests/ui/allow_attributes_false_positive.rs
```diff
@@ -0,0 +1,6 @@
+#![allow(unused)]
```
Alexendoo, line 1: Thanks! For clarity lets remove this, I didn't realise it only linted on outer attrs so I was wondering why this wasn't linted

```suggestion
```

Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

PR #10692, opened 2023-04-21
new lint: `missing_asserts_for_indexing`

### LINKED ISSUE #8296: If indexing slice more than once, suggest asserting length beforehand
### What it does

If you assert before indexing, Rust will remove the bounds checks on the indexings.

An example: https://godbolt.org/z/fczP9YTbj (correctly optimizes at opt-level=1 and higher)

### Lint Name

multiple-indexing-assert-length

### Category

_No response_

### Advantage

- Improved performance

### Drawbacks

_No response_

### Example

```rust
let _ = x[0]; // bounds check here
let _ = x[1..4]; // bounds check here
```

Could be written as:

```rust
assert!(x.len() >= 10);

let _ = x[0]; // no bounds check here
let _ = x[1..4]; // or here
```

y21:
Fixes #8296

This lint looks for repeated slice indexing and suggests adding an `assert!` beforehand that helps LLVM elide bounds checks. The lint documentation has an example.

I'm not really sure what category this should be in. It seems like a nice lint for the `perf` category but I suspect this has a pretty high FP rate, so it might have to be a pedantic lint or something.
I'm also not sure about the name. If someone knows a better name for this lint, I'd be fine with changing it.

changelog: new lint [`missing_asserts_for_indexing`]

### DIFF
--- a/tests/ui/missing_asserts_for_indexing.fixed
+++ b/tests/ui/missing_asserts_for_indexing.fixed
@@ -1,0 +1,123 @@
+//@run-rustfix
+
+#![allow(unused)]
+#![warn(clippy::missing_asserts_for_indexing)]
+
+// ok
+fn sum_with_assert(v: &[u8]) -> u8 {
+    assert!(v.len() > 4);
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+// ok
+fn sum_with_assert_other_way(v: &[u8]) -> u8 {
+    assert!(5 <= v.len());
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+// ok
+fn sum_with_assert_ge(v: &[u8]) -> u8 {
+    assert!(v.len() >= 5);
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+// ok
+fn sum_with_assert_ge_other_way(v: &[u8]) -> u8 {
+    assert!(4 < v.len());
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+fn sum_with_assert_lt(v: &[u8]) -> u8 {
+    assert!(v.len() > 4);
+    //~^ ERROR incorrect length
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+fn sum_with_assert_le(v: &[u8]) -> u8 {
+    assert!(v.len() > 4);
+    //~^ ERROR incorrect length
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+fn sum_with_incorrect_assert_len(v: &[u8]) -> u8 {
+    assert!(v.len() > 4);
+    //~^ ERROR incorrect length
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+fn sum_with_incorrect_assert_len2(v: &[u8]) -> u8 {
+    assert!(v.len() > 4);
+    //~^ ERROR incorrect length
+    v[0] + v[1] + v[2] + v[3] + v[4]
+}
+
+// ok, don't lint for single array access
+fn single_access(v: &[u8]) -> u8 {
+    v[0]
+}
+
+// ok
+fn subslice_ok(v: &[u8]) {
+    assert!(v.len() > 3);
+    let _ = v[0];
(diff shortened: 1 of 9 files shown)


y21:
Oh huh, I just tried adding a test for multiple different slices (ie slices from different bindings) in the same block:
```rs
fn foo(a: &[u8], b: &[u8]) {
  let _ = a[0] + a[1];
  let _ = b[0] + b[1];
}
```
and it seems like it thinks all of those array accesses reference the same slice. `SpanlessHash` doesn't hash the local path expressions "correctly" and both expressions (`a` and `b`) result in the same hash:

https://github.com/rust-lang/rust-clippy/blob/928349795f9f28f54e69bd0f0ffeed33de5e272a/clippy_utils/src/hir_utils.rs#L878-L883

Not really sure why it is doing what it is doing (always hashing `1`) or how I'd differentiate between different slices now :/

llogiq:
I think the problem is that if two idents are the same constant value, `SpanlessEq` returns `true`. So if `SpanlessHash` hashed the ident, it would have two different idents hashed to different values and you'd end up with equal values distributed to different buckets. Boom.

What we could do is hash locals to the constant value if any and else hash the ident, I think.

Alexendoo:
There'd be the case where locals are deeper in other expressions like `a.field` vs `b.field`, there's also the possibility of natural hash collisions if we solve that. We probably want to do something similar to https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.search_same.html

y21:
Right now this only works if integer literals are used for indexing and it'd be interesting if this could be extended to also work for some simple arithmetic, like
```rs
fn foo(x: &mut Vec<i32>, n: usize) -> i32 {
  // suggest `assert!(x.len() > n + 3);` here
  x[n] + x[n + 1] + x[n + 2] + x[n + 3]
}
```
But that probably complicates the logic quite a bit, so maybe it's worth doing this in a separate PR?

tests/ui/missing_asserts_for_indexing.fixed
Alexendoo, line 1: Some tests that would be good to have

```rust
// more complex expressions
s.field[0] + s.field[1]
```
```rust
// exprs w/ side effects that shouldn't be linted
f()[0] + f()[1]
```
```rust
// like index_different_slices but shouldn't lint
let _ = a[0] + b[1];
```

Alexendoo:
I'm not sure about the arithmetic case, it would

Alexendoo:
Nominating this for discussion in the next clippy meeting. Tuesday the 13th, 15:00 UTC @ https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy

I'd like to discuss what category this should fit into

y21:
> I'm not sure about the arithmetic case, does that version still get optimised to a single bounds check?

Huh, good point. It doesn't look like it gets optimized to a single bounds check as I've written it with the assert. I am guessing that the fact that `n + 3` can overflow messes with it.
[Writing the assert like this](https://godbolt.org/z/P9jGW9sGd) reduces it to one bounds check (+ the check that it doesn't overflow), but it's kinda weird, so probably better to not do this for now (it also seems way more fragile, very slight changes to the assert miss the optimization, e.g. changing it from `&&` to `&`).


> I'd like to discuss what category this should fit into

👍
I was originally thinking about the perf category, but the more I think about it, I'm guessing that a lot of code does this sort of pattern, and having to assert this everywhere is probably quite annoying, especially when performance isnt' as much of a concern (to the point where one has to worry about branching).

y21:
Changed the category to restriction as decided in the [clippy meeting](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202023-06-13/near/365903765)

Alexendoo (approved):

Alexendoo:
Thanks! LGTM with a rebase

@bors delegate+

y21:
I rebased the PR and also had to change a few things in the tests since uitest has changed quite a bit since I've created this PR (intentionally put this into its own commit so reviewing that is easier).

Alexendoo:
> changed quite a bit since I've created this PR

I wonder how that happened 😅

I really need to start going through my assigned PRs bottom to top so things don't get left like this

Thanks again! @bors r+

Issue #9757, opened 2022-10-31
Possible problems involving `clippy::shadow_reuse` and auto-generated code


c410-f3r:
### Description

According to https://github.com/dtolnay/quote/issues/232#issuecomment-1296451876, `clippy::shadow_reuse` shouldn't be triggered but I am personally not aware of this intended limitation. Any thoughts?

### Version

```text
Latest 1.65 beta
```

Alexendoo:
A fix for this case would be to check `pat.span.from_expansion()` in addition to the current `ident.span.from_expansion()` check

It wouldn't be intended for this to lint, ident + macro interactions are just non obvious

c410-f3r:
Closing due to the lack of activity.

lochetti:
I am trying to take a look at this issue. Although, this would be my first contribution here.

Not linting the code if `pat.span.from_expansion()` does solve the specific issue mentioned for the `quote` crate, but I have one doubt: Which is the best way to increment the tests in order to add this use case? I am really not sure how to do that.

Thanks for the patience in helping a new contributor wannabe!

c410-f3r:
To solve the issue, I **think** `shadow_reuse` should also check `clippy_utils::is_from_proc_macro` along side the already included `ident.span.from_expansion()`. See https://github.com/rust-lang/rust-clippy/pull/10203#issuecomment-1509929940.

In regards to test addition or how to test this issue specifically, I really have no clue because https://github.com/rust-lang/rust-clippy/pull/10203/files#diff-98b6b061d5b311aae09fd482c612aab5544d55b9c0da67438d79ef0fe181d526R129-R159 didn't help.

Perhaps one of the Clippy members can help you? @Alexendoo @Jarcho @xFrednet

Alexendoo:
You can add a test along the lines of

```rust
macro_rules! reuse {
    ($v:ident) => {
        let $v = $v + 1;
    }
}

let x = 1;
reuse!(x);
```

PR #10697, opened 2023-04-22
Ignore `shadow` warns in code from macro expansions

lochetti:
This PR fixes https://github.com/rust-lang/rust-clippy/issues/9757

I am in doubt if just looking for `pat.span.from_expansion()` would be sufficient instead of looking for both `pat.span.desugaring_kind().is_some()` or `pat.span.from_expansion()`. The tests (including the new one) passes if I leave the only `if pat.span.from_expansion()`. Any feedbacks?

Also, this is my first PR here, sorry for anything and thanks for the patience!

changelog: [`shadow_same`, `shadow_reuse`, `shadow_unrelated`]: avoiding warns in macro-generated code

### DIFF
--- a/clippy_lints/src/shadow.rs
+++ b/clippy_lints/src/shadow.rs
@@ -108,7 +108,7 @@ impl<'tcx> LateLintPass<'tcx> for Shadow {
     fn check_pat(&mut self, cx: &LateContext<'tcx>, pat: &'tcx Pat<'_>) {
         let PatKind::Binding(_, id, ident, _) = pat.kind else { return };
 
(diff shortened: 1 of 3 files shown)


dswij (approved):
Thanks for the PR!

For checking if the span is from macros, `from_expansion()` works just fine. I'd guess that the idea behind ignoring any spans from desugar (through `desugaring_kind().is_some()`) is just to be safe - i.e., shadows in desugared code should not be linted.

I'd say that it makes sense and it's fine to leave the desugar check as it is for now.

PR #10698, opened 2023-04-22
Migrate `clippy::author` to `rustc_ast::FormatArgs`

Niki4tap:
This PR aims to migrate `clippy::author` lint to new `FormatArgs` AST, since it's been ICEing after the change.

repro.rs:
```rs
#![feature(stmt_expr_attributes)]

fn main() {
    #[clippy::author]
    {
        println!("Hello, {}!", "ferris");
    }
}
```

<details>
<summary>
Current ICE
</summary>

```
if let ExprKind::Block(block, None) = expr.kind
    && block.stmts.len() == 1
    && let StmtKind::Semi(e) = block.stmts[0].kind
    && let ExprKind::Block(block1, None) = e.kind
    && block1.stmts.len() == 1
    && let StmtKind::Semi(e1) = block1.stmts[0].kind
    && let ExprKind::Call(func, args) = e1.kind
    && let ExprKind::Path(ref qpath) = func.kind
    && match_qpath(qpath, &["$crate", "io", "_print"])
    && args.len() == 1
    && let ExprKind::Call(func1, args1) = args[0].kind
    && let ExprKind::Path(ref qpath1) = func1.kind
thread 'rustc' panicked at 'path_to_string: called for lang item qpath', clippy_lints/src/utils/author.rs:746:36

note: Clippy version: clippy 0.1.70 (928349795 2023-04-21)

query stack during panic:
#0 [analysis] running analysis passes on this crate
end of query stack
```
</details>

<details>
<summary>
After this PR
</summary>

```
if let ExprKind::Block(block, None) = expr.kind
    && block.stmts.len() == 1
    && let StmtKind::Semi(e) = block.stmts[0].kind
    && let Some(macro_call) = macros::root_macro_call_first_node(cx, e)
    && cx.tcx.is_diagnostic_item(sym::println_macro, macro_call.def_id)
    && let args = macros::collect_format_args(cx, |_fmt_args| {}, e, macro_call.expn)
    && args.len() == 1
    && let ExprKind::Lit(ref lit) = args[0].kind
    && let LitKind::Str(s, _) = lit.node
    && s.as_str() == "ferris"
    && block.expr.is_none()
{
    // report your lint here
}
```
</details>

r? @Alexendoo

I've seen you migrate other lints, so I think your feedback will be valuable here. Thanks in advance :)
(if anyone else has an opinion that might be useful here, you're free to comment too)

Also, I'm not exactly sure if a new utility function is needed here, or it would better to reuse existing ones like `clippy_utils::macros::find_format_args`.

note: I think `author` lint is mostly internal, and the change is too, but on the other hand it's a bugfix, so I'm not sure what to put in the changelog, let there be none for now :)

changelog: none

### DIFF
--- a/clippy_lints/src/utils/author.rs
+++ b/clippy_lints/src/utils/author.rs
@@ -1,6 +1,7 @@
 //! A group of attributes that can be attached to Rust code in order
 //! to generate a clippy lint detecting said code automatically.
 
+use clippy_utils::macros::{collect_format_args, is_format_macro, root_macro_call_first_node};
 use clippy_utils::{get_attr, higher};
 use rustc_ast::ast::{LitFloatType, LitKind};
 use rustc_ast::LitIntType;
@@ -239,14 +240,14 @@ impl<'a, 'tcx> PrintVisitor<'a, 'tcx> {
         }
     }
 
-    fn slice<T>(&self, slice: &Binding<&[T]>, f: impl Fn(&Binding<&T>)) {
+    fn slice<T>(&self, slice: &Binding<&[T]>, f: impl Fn(Binding<&T>)) {
         if slice.value.is_empty() {
             chain!(self, "{slice}.is_empty()");
         } else {
             chain!(self, "{slice}.len() == {}", slice.value.len());
             for (i, value) in slice.value.iter().enumerate() {
                 let name = format!("{slice}[{i}]");
-                f(&Binding { name, value });
+                f(Binding { name, value });
             }
         }
     }
@@ -333,6 +334,24 @@ impl<'a, 'tcx> PrintVisitor<'a, 'tcx> {
 
     #[allow(clippy::too_many_lines)]
     fn expr(&self, expr: &Binding<&hir::Expr<'_>>) {
+        if let Some(macro_call) = root_macro_call_first_node(self.cx, expr.value)
+            && is_format_macro(self.cx, macro_call.def_id)
+        {
+            let diag = self.cx.tcx.get_diagnostic_name(macro_call.def_id).unwrap();
+            bind!(self, macro_call);
+            chain!(self, "let Some({macro_call}) = macros::root_macro_call_first_node(cx, {expr})");
+            chain!(self, "cx.tcx.is_diagnostic_item(sym::{diag}, {macro_call}.def_id)");
+            let exprs = collect_format_args(self.cx, |_fmt_args| {}, expr.value, macro_call.value.expn);
+            let args = exprs.as_slice();
+            bind!(self, args);
+            chain!(self, "let {args} = macros::collect_format_args(cx, |_fmt_args| {{}}, {expr}, {macro_call}.expn)");
+            self.slice(args, |e| {
+                let expr = Binding { name: e.name, value: *e.value };
+                self.expr(&expr);
+            });
+            return;
+        }
+
         if let Some(higher::While { condition, body }) = higher::While::hir(expr.value) {
             bind!(self, condition, body);
             chain!(
@@ -398,25 +417,25 @@ impl<'a, 'tcx> PrintVisitor<'a, 'tcx> {
             ExprKind::Array(elements) => {
                 bind!(self, elements);
                 kind!("Array({elements})");
-                self.slice(elements, |e| self.expr(e));
+                self.slice(elements, |e| self.expr(&e));
             },
             ExprKind::Call(func, args) => {
                 bind!(self, func, args);
                 kind!("Call({func}, {args})");
                 self.expr(func);
-                self.slice(args, |e| self.expr(e));
+                self.slice(args, |e| self.expr(&e));
             },
             ExprKind::MethodCall(method_name, receiver, args, _) => {
                 bind!(self, method_name, receiver, args);
                 kind!("MethodCall({method_name}, {receiver}, {args}, _)");
                 self.ident(field!(method_name.ident));
                 self.expr(receiver);
-                self.slice(args, |e| self.expr(e));
+                self.slice(args, |e| self.expr(&e));
             },
             ExprKind::Tup(elements) => {
                 bind!(self, elements);
                 kind!("Tup({elements})");
-                self.slice(elements, |e| self.expr(e));
+                self.slice(elements, |e| self.expr(&e));
             },
             ExprKind::Binary(op, left, right) => {
                 bind!(self, op, left, right);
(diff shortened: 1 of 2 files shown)


clippy_lints/src/utils/author.rs
```diff
@@ -333,6 +334,24 @@ impl<'a, 'tcx> PrintVisitor<'a, 'tcx> {
...
+            let diag = self.cx.tcx.item_name(macro_call.def_id);
+            bind!(self, macro_call);
+            chain!(self, "let Some({macro_call}) = macros::root_macro_call_first_node(cx, {expr})");
+            chain!(self, "cx.tcx.is_diagnostic_item(sym::{diag}_macro, {macro_call}.def_id)");
+            let exprs: Vec<_> = collect_format_args(expr.value).collect();
+            let args = exprs.as_slice();
+            bind!(self, args);
+            chain!(self, "let {args} = macros::collect_format_args({expr}).collect()");
+            self.slice(args, |e| {
+                let expr = Binding { name: e.name.clone(), value: *e.value };
```
Niki4tap, line 349: I really don't like this `clone`, but I'm not really sure how to avoid it here...

clippy_lints/src/utils/author.rs
```diff
@@ -333,6 +334,24 @@ impl<'a, 'tcx> PrintVisitor<'a, 'tcx> {
 
     #[allow(clippy::too_many_lines)]
     fn expr(&self, expr: &Binding<&hir::Expr<'_>>) {
+        if let Some(macro_call) = root_macro_call_first_node(self.cx, expr.value)
+            && is_format_macro(self.cx, macro_call.def_id)
+        {
+            let diag = self.cx.tcx.item_name(macro_call.def_id);
+            bind!(self, macro_call);
+            chain!(self, "let Some({macro_call}) = macros::root_macro_call_first_node(cx, {expr})");
+            chain!(self, "cx.tcx.is_diagnostic_item(sym::{diag}_macro, {macro_call}.def_id)");
```
Niki4tap, line 343: I like `sym::{diag}_macro`, and it *almost* works, for exception of `panic`, it should be `std_panic_macro`, but this implementation returns just `panic_macro`. I would appreciate a better solution here, but if there's none, I will hardcode this specific case, although it seems more like a workaround.

Alexendoo: Would `tcx.get_diagnostic_name` work?

Alexendoo:
Thanks for this, didn't know `clippy::author` had support for that

> Also, I'm not exactly sure if a new utility function is needed here, or it would better to reuse existing ones like `clippy_utils::macros::find_format_args`.

A new utility function could be nice, I'm not that happy with the current API. However we can't match on the HIR of `format_args` as `collect_format_args` does because that was holding back progress in `rustc`, it ought to be based on `find_format_args`/`find_format_arg_expr`

clippy_lints/src/utils/author.rs
Niki4tap: Seems to work like a charm, thanks!

clippy_lints/src/utils/author.rs
Niki4tap: I've changed the signature of `slice` to provide an owned `Binding` as it already has it, not sure why it was by-ref previously, but that solved the issue of unnecessary clone.

Niki4tap:
> A new utility function could be nice, I'm not that happy with the current API. However we can't match on the HIR of `format_args` as `collect_format_args` does because that was holding back progress in `rustc`, it ought to be based on `find_format_args`/`find_format_arg_expr`

Ah, yeah that makes sense. I've changed `collect_format_args` implementation to use these instead of HIR matching, should be fine now ^^

clippy_utils/src/macros.rs
```diff
@@ -425,6 +425,33 @@ pub fn find_format_arg_expr<'hir, 'ast>(
...
+/// println!("Hello, {}!", "ferris")
+/// //                     ^^^^^^^^ returns these expressions
+/// ```
+pub fn collect_format_args<'hir>(
+    cx: &LateContext<'_>,
+    start: &'hir Expr<'hir>,
+    expn_id: ExpnId,
+) -> Vec<&'hir Expr<'hir>> {
+    let mut args = vec![];
+    find_format_args(cx, start, expn_id, |format_args| {
```
Alexendoo, line 441: Since lints would likely want to access the `format_args` part too it would be good to have this API take a callback that has the found args as the second argument. That way we don't have to `find_format_args` twice

Niki4tap: Done

clippy_utils/src/macros.rs
```diff
@@ -425,6 +425,38 @@ pub fn find_format_arg_expr<'hir, 'ast>(
...
+/// // vvvvv any format-like macro
+/// println!("Hello, {}!", "ferris")
+/// //                     ^^^^^^^^ returns these expressions
+/// ```
+pub fn collect_format_args<'hir>(
+    cx: &LateContext<'_>,
+    callback: impl FnOnce(&FormatArgs),
+    start: &'hir Expr<'hir>,
+    expn_id: ExpnId,
+) -> Vec<&'hir Expr<'hir>> {
```
Alexendoo, line 443: It'd want to be  `callback: impl FnOnce(&FormatArgs, Vec<...>)` rather than returning the `Vec`, lints generally want access to both at the same time

Niki4tap: Oh, sorry for misunderstanding then. Although I'm not quite sure how to apply it in the author lint, if we don't return the args, it wouldn't quite fit into the if-chain author lint tries to build, i. e.
```rs
   && macros::collect_from_args(cx, |args, _fmt_args| { if /* another if-chain */ { /* report your lint here */ } }, e, macro_call.expn)
```
\- this does not only introduce another level on indentation, but also requires quite a bit of code refactoring to support, which is, in my opinion, quite a bit out of scope for the author lint.

If you have any ideas on how to mitigate this, I would be happy to hear, otherwise I can close the PR and submit an issue on this.

Alexendoo: For `author` hmm that's tricky, I guess you could always print a comment saying to take a look at the function

Alexendoo:
Since #11444 was merged `find_format_args` now returns a value instead of taking a callback making this easier if you wanted to pick it up again, otherwise I'll add it to my todo list

xFrednet:
Hey @Alexendoo, this is a ping from triage. It seems like @Niki4tap hasn't responded to your last message where you suggested that you'll pick this PR up. Do we want to close this one due to inactivity?

PR #10699, opened 2023-04-22
Clippy Book Chapter Updates Reborn: Refresh Lint Configuration's looks

blyxyas:
This PR modernizes and clears up some confusion with the "Lint Configuration Options" chapter from the book.

### Changes

- **Remove 'Option - Default Value" table**

    - Why was it even there?
    - It shouldn't be the first thing an user sees when they enter the chapter. It's clunky, ugly and not useful. The default values for configs are stated in a per-config basis if needed.

- **Add a simple description of what the chapter contains, and the scheme of each configuration option**

- **Minor formatting, mainly adding code fragments to code text**

    - It seemed weird and jarring not having back-ticks on text like "arithmetic_side_effects".
    -  Improves readability and separation between configs.

- **Separate a little bit the Affected Lints list + "Affected lists" message**
 
    - Not having something indicating that the list is about the lints that use the configuration option is confusing.
    - It isn't as important as the description and example. Therefore should be separated a little bit imo

---

This is an independent effort from #10597, but as it's still a Book Chapter Update, I thought it would be cool to include it here. I'm going to keep the reviewing process for this PR to rustbot's desires.

[Rendered](https://github.com/blyxyas/rust-clippy/blob/book-lint_config/book/src/lint_configuration.md)
[Current](https://github.com/rust-lang/rust-clippy/blob/master/book/src/lint_configuration.md)

changelog: Refresh styling from the "Lint Configuration Options" book chapter.


xFrednet:
Hey @Jarcho, this is a ping from triage. Could you review this PR? If you're busy you can also pass it to me. I have a reviewing queue, but I should be able to get to this one within a week or so :)

flip1995:
This PR also was on my todo list for some time. Just looked over it and this is an easy approval. Thanks for cleaning this up!

@bors r+

PR #10705, opened 2023-04-23
Add a test that checks for old style test headers

Alexendoo:
Follow up to #10669, we're pretty used to them so they're easy to slip through

changelog: none


flip1995:
@bors r+

Thanks!

PR #10706, opened 2023-04-23
Remove check for `lib.register_*` and `src/docs*` in `cargo dev update_lints`

Alexendoo:
The transition period has well passed

changelog: none


llogiq:
Yeah, makes sense. Thank you!

@bors r+

Issue #10713, opened 2023-04-25
items_after_test_module disagrees with widespread practice


dtolnay:
### Summary

```rust
mod bar;
mod foo;

#[cfg(test)]
mod tests;

use bar::Trait;

pub fn fun() {}
```

I believe clippy should not dislike this code. Declaring modules at the top of the file is extremely common practice. People disagree on whether `mod` goes before or after `use`, but both of those ways are way more common than putting `mod` last in a file.

### Lint Name

items_after_test_module

### Reproducer

```rust
// src/lib.rs

#[cfg(test)]
mod tests;

pub fn f() {}
```

```rust
// src/tests.rs (empty)
```

```console
$ cargo clippy --tests

warning: items were found after the testing module
 --> src/lib.rs:2:1
  |
2 | / mod tests;
3 | |
4 | | pub fn f() {}
  | |_____________^
  |
  = help: move the items to before the testing module was defined
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#items_after_test_module
  = note: `#[warn(clippy::items_after_test_module)]` on by default
```

### Version

```text
rustc 1.71.0-nightly (f5559e338 2023-04-24)
binary: rustc
commit-hash: f5559e338256f17ada6d82b429acc2dbd8facc9c
commit-date: 2023-04-24
host: x86_64-unknown-linux-gnu
release: 1.71.0-nightly
LLVM version: 16.0.2
```

dtolnay:
@blyxyas @dswij (#10578)

blyxyas:
I think downgrading here isn't the correct move. But instead ignore module imports.

The lint was never intended to catch

```
#[cfg(test)]
mod tests;
```

It was intended to catch modules with a body.

```
#[cfg(test)]
mod tests {
// [•••]
}
```

So this isn't the lint level's fault, this is just a bug.

ilyagr:
I'm still having problems. I'm not sure if it's a bug in clippy or in the `test-case` library. I filed https://github.com/frondeus/test-case/issues/122.

PR #10714, opened 2023-04-25
Downgrade items_after_test_module to nursery

dtolnay:
I think this lint shouldn't be enabled by default while we reconcile #10713.

changelog: [`items_after_test_module`]: remove from default set of enabled lints


blyxyas:
Maybe this should be closed as the original bug was fixed.

dtolnay:
Thanks!

PR #10720, opened 2023-04-27
run linkcheck in clippy ci

### LINKED ISSUE #10711: run linkcheck in clippy ci
### Summary

This often causes pain when doing clippy->rustc syncs, as @flip1995  pointed out we could we should see to we can run linkcheck in clippy CI to catch this earlier https://github.com/rust-lang/rust/pull/110718#issuecomment-1519557770

SergenKaraoglan:
fixes #10711

changelog: Run [linkcheck.sh](https://github.com/rust-lang/rust/blob/master/src/tools/linkchecker/linkcheck.sh) from rustc repo in Remark.yml to check Clippy book.

### DIFF
--- a/.github/workflows/remark.yml
+++ b/.github/workflows/remark.yml
@@ -36,6 +36,12 @@ jobs:
     - name: Check *.md files
       run: git ls-files -z '*.md' | xargs -0 -n 1 -I {} ./node_modules/.bin/remark {} -u lint -f > /dev/null
 
+    - name: Linkcheck book
+      run: |
+        rustup toolchain install nightly --component rust-docs
+        curl https://raw.githubusercontent.com/rust-lang/rust/master/src/tools/linkchecker/linkcheck.sh -o linkcheck.sh
+        sh linkcheck.sh clippy --path ./book
+        
     - name: Build mdbook
       run: mdbook build book
 
--- a/book/src/development/lint_passes.md
+++ b/book/src/development/lint_passes.md
@@ -50,7 +50,7 @@ questions already, but the parser is okay with it. This is what we
 mean when we say `EarlyLintPass` deals with only syntax on the AST level.
 
 Alternatively, think of the `foo_functions` lint we mentioned in
-[define new lints](define_lints.md#name-the-lint) chapter.
+define new lints <!-- FIXME: add link --> chapter.
 
 We want the `foo_functions` lint to detect functions with `foo` as their name.
 Writing a lint that only checks for the name of a function means that we only


.github/workflows/remark.yml
```diff
@@ -36,6 +36,13 @@ jobs:
     - name: Check *.md files
       run: git ls-files -z '*.md' | xargs -0 -n 1 -I {} ./node_modules/.bin/remark {} -u lint -f > /dev/null
 
+    - name: Linkcheck book
+      run: |
+        rustup default nightly
+        rustup toolchain install nightly --component rust-docs
```
flip1995, line 42: GHA has a stable Rust toolchain installed by default. Does it also work with that or does it require nightly for some reason?

SergenKaraoglan: According to linkcheck.sh: "This requires the rust-docs rustup component to be installed in the nightly
# toolchain."

flip1995:
Can you push a commit, where you change a markdown file? I'd like to see this in action.

flip1995:
Ok, thanks it is working! (There is currently a broken link in the book that I only fixed in the rust-lang/rust repo).

Looking at this: https://github.com/rust-lang/rust-clippy/actions/runs/4830350981/jobs/8606470214#step:7:54 I think adding `rust-docs` as a component to our `rust-toolchain` file should be enough. I don't think you have to install the nightly toolchain.

SergenKaraoglan:
Doesn't appear to work without nightly toolchain. https://github.com/rust-lang/rust-clippy/actions/runs/4830750054/jobs/8607336167?pr=10720#step:7:30

https://github.com/rust-lang/rust/blob/master/src/tools/linkchecker/linkcheck.sh#L26

flip1995:
Thanks! last steps:

- Remove the link from here: https://github.com/rust-lang/rust-clippy/blob/990bbdc2bee264291843497cebe59f307c2ed86f/book/src/development/lint_passes.md?plain=1#L53 and replace it with a `<!-- FIXME: add link -->`
- Revert other markdown change
- Squash the commits

book/src/development/lint_passes.md
```diff
@@ -50,7 +50,7 @@ questions already, but the parser is okay with it. This is what we
 mean when we say `EarlyLintPass` deals with only syntax on the AST level.
 
 Alternatively, think of the `foo_functions` lint we mentioned in
-[define new lints](define_lints.md#name-the-lint) chapter.
+[define new lints](<!-- FIXME: add link -->) chapter.
```
flip1995, line 53:
```suggestion
define new lints <!-- FIXME: add link --> chapter.
```

SergenKaraoglan: Yes that makes much more sense, sorry 😅

SergenKaraoglan:
Thank you for the quick responses.

PR #10725, opened 2023-04-29
don't remove `dbg!` in arbitrary expressions

### LINKED ISSUE #9914: `dbg_macro` removes empty dbg! macro
### Summary

.

### Lint Name

dbg_macro

### Reproducer

I tried this code:

```rust
fn main() {
    let _: () = dbg!();
}
```

I saw this happen:
` cargo clippy --fix -- -Aclippy::all -Wclippy::dbg_macro`
```

  * src/main.rs

error: expected expression, found `;`
 --> src/main.rs:2:17
  |
2 |     let _: () = ;
  |                 ^ expected expression

warning: `dbg!` macro is intended as a debugging tool
 --> src/main.rs:2:17
  |
2 |     let _: () = dbg!();
  |                 ^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#dbg_macro
  = note: requested on the command line with `-W clippy::dbg-macro`
help: ensure to avoid having uses of it in version control
  |
2 -     let _: () = dbg!();
2 +     let _: () = ;
  |

warning: `clpy` (bin "clpy") generated 1 warning (run `cargo fix --bin "clpy"` to apply 1 suggestion)

  * src/main.rs

error: expected expression, found `;`
 --> src/main.rs:2:17
  |
2 |     let _: () = ;
  |                 ^ expected expression

warning: `clpy` (bin "clpy" test) generated 1 warning (1 duplicate)
    Finished dev [unoptimized + debuginfo] target(s) in 0.48s
```

### Version

```text
binary: rustc
commit-date: 2022-11-19
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

y21:
Fixes #9914

The `dbg_macro` lint replaces empty `dbg!` invocations with the empty string in its suggestion, which is not always valid code in certain contexts (e.g. `let _ = dbg!();` becomes `let _ = ;`). This PR changes it to `()`, which should always be valid where `dbg!()` is valid (`dbg!()` with no arguments evaluates to `()`).

It also special-cases "standalone" `dbg!();` expression statements, where it will suggest removing the whole statement entirely like it did before.

changelog: [`dbg_macro`]: don't remove `dbg!()` in arbitrary expressions as it sometimes results in syntax errors

### DIFF
--- a/clippy_lints/src/dbg_macro.rs
+++ b/clippy_lints/src/dbg_macro.rs
@@ -3,10 +3,10 @@ use clippy_utils::macros::root_macro_call_first_node;
 use clippy_utils::source::snippet_with_applicability;
 use clippy_utils::{is_in_cfg_test, is_in_test_function};
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind};
-use rustc_lint::{LateContext, LateLintPass};
+use rustc_hir::{Expr, ExprKind, Node, Stmt, StmtKind};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
-use rustc_span::sym;
+use rustc_span::{sym, Span};
 
 declare_clippy_lint! {
     /// ### What it does
@@ -29,6 +29,11 @@ declare_clippy_lint! {
     pub DBG_MACRO,
     restriction,
     "`dbg!` macro is intended as a debugging tool"
+}
+
+fn span_including_semi(cx: &LateContext<'_>, span: Span) -> Span {
+    let span = cx.sess().source_map().span_extend_to_next_char(span, ';', true);
+    span.with_hi(span.hi() + rustc_span::BytePos(1))
 }
 
 #[derive(Copy, Clone)]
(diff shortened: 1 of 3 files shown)


Jarcho:
Does this work properly when `dbg!()` is expanded from another macro? e.g. `foo!();` where `foo!()` expands to `dbg!()`.

y21:
> Does this work properly when `dbg!()` is expanded from another macro? e.g. `foo!();` where `foo!()` expands to `dbg!()`.

Hmm it seems like it breaks if the `dbg!()` is passed as an expr metavariable
```rs
macro_rules! foo {
    ($x:expr) => { $x; };
}

fn main() {
    foo!(dbg!());
}
```
```
warning: the `dbg!` macro is intended as a debugging tool
  --> test.rs:16:10
   |
16 |     foo!(dbg!());
   |          ^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#dbg_macro
  --> test.rs:1:9
   |
1  | #![warn(clippy::dbg_macro)]
   |         ^^^^^^^^^^^^^^^^^
help: remove the invocation before committing it to a version control system
   |
16 -     foo!(dbg!());
16 +     foo!(
   |
```
so perhaps my way of checking if we're allowed to remove the `dbg!()` by checking if the parent node is a `StmtKind::Semi` is not that reliable.

Jarcho:
I think the only thing you can do here is to start tokenizing from the end of the call site and see if the first non-whitespace token is a semicolon.

y21:
Alright, it's not exactly pretty but it works.

Jarcho:
It works, so...

Thank you. @bors r+

PR #10730, opened 2023-05-01
`imprecise_flops`: Globally ignore `#[no_std]` crates

### LINKED ISSUE #10728: `mul_add` suggested for `no_std` where it is unavailable
### Summary

I think the idea behind this lint is great.  Thank you for your work.  Unfortunately `mul_add` is not available with `no_std`.

`mul_add` should not be suggested for `no_std`

### Lint Name

suboptimal_flops

### Reproducer

I tried this code:

```rust
f64::from(temp_data_raw - 0x8000) * 7.8125e-3f64 - 256.0f64
```

I saw this happen:

```
warning: failed to automatically apply fixes suggested by rustc to crate `p1`

  * p1/src/lib.rs

error[E0599]: no method named `mul_add` found for type `f64` in the current scope
  --> p1/src/lib.rs:43:51
   |
43 |                 f64::from(temp_data_raw - 0x8000).mul_add(7.8125e-3f64, -256.0f64),
   |                                                   ^^^^^^^ method not found in `f64`

For more information about this error, try `rustc --explain E0599`.
```

No suggestion of `mul_add` for `no_std`

### Version

```text
rustc 1.69.0 (84c898d65 2023-04-16)
binary: rustc
commit-hash: 84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc
commit-date: 2023-04-16
host: aarch64-unknown-linux-gnu
release: 1.69.0
LLVM version: 15.0.7
```

### Additional Labels

blyxyas:
Really small fix.
Fixes #10728
changelog: [`imprecise_flops`]: Fix false positives with `#[no_std]`

### DIFF
--- a/clippy_lints/src/floating_point_arithmetic.rs
+++ b/clippy_lints/src/floating_point_arithmetic.rs
@@ -732,14 +732,14 @@ fn check_radians(cx: &LateContext<'_>, expr: &Expr<'_>) {
 impl<'tcx> LateLintPass<'tcx> for FloatingPointArithmetic {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
         // All of these operations are currently not const and are in std.
-        if in_constant(cx, expr.hir_id) || is_no_std_crate(cx) {
+        if in_constant(cx, expr.hir_id) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


Jarcho:
`to_radians` and `to_degrees` are in `core`. Otherwise this is fine.

Jarcho:
Thank you. @bors r+

PR #10734, opened 2023-05-02
Update macros.rs (typo)

smoelius:
r? @Alexendoo

changelog: none


smoelius:
Aside: how do you know this assumption holds?

Alexendoo (approved):

Alexendoo:
Thanks! @bors r+

If the lint passes start being run from different threads we'd see test failures for any lints that use `find_format_args`

smoelius:
> If the lint passes start being run from different threads we'd see test failures for any lints that use `find_format_args`

Ah. Thanks!

PR #10736, opened 2023-05-03
initial clippy::ref_patterns implementation

### LINKED ISSUE #9010: New lint: ref_pattern, for disallowing the ref keyword
### What it does

I would like to have a lint that fires on any use of the `ref` keyword.

### Lint Name

ref_pattern

### Category

restriction

### Advantage

Code that doesn't use it is easier to read for newcomers since they're unlikely to be aware of what it does.

### Drawbacks

_No response_

### Example

```rust
if let Some(ref foo) = *rwlock.read().unwrap() {}
```

Could be written as:

```rust
if let Some(foo) = &*rwlock.read().unwrap() {}
```

NotAPenguin0:
This implements a new lint that discourages the use of the `ref` keyword as outlined in #9010. I think there are still some things to improve about this lint, but I need some feedback before I can implement those.

- [x] ~~Maybe it's desirable that a quick fix is listed too, instead of a generic `avoid using the ref keyword` lint.~~
- [x] `let ref x = y` already has a lint (`clippy::toplevel_ref_arg`). This implementation will report this too, so you get two lints for the same issue, which is not great. I don't really know a way around this though.
- [X] The dogfood test is currently failing locally, though I ran `cargo clippy -- -D clippy::all -D clippy::pedantic` and got no output, so I'm not sure why this is.

Any help with these would be greatly appreciated.

fixes #9010 
changelog: [`ref_patterns`]: newly added lint

### DIFF
--- a/clippy_lints/src/misc.rs
+++ b/clippy_lints/src/misc.rs
@@ -16,8 +16,11 @@ use rustc_span::source_map::{ExpnKind, Span};
 
 use clippy_utils::sugg::Sugg;
 use clippy_utils::{
-    get_parent_expr, in_constant, is_integer_literal, is_no_std_crate, iter_input_pats, last_path_segment, SpanlessEq,
+    get_parent_expr, in_constant, is_integer_literal, is_lint_allowed, is_no_std_crate, iter_input_pats,
+    last_path_segment, SpanlessEq,
 };
+
+use crate::ref_patterns::REF_PATTERNS;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -162,6 +165,10 @@ impl<'tcx> LateLintPass<'tcx> for LintPass {
             return;
         }
         for arg in iter_input_pats(decl, body) {
+            // Do not emit if clippy::ref_patterns is not allowed to avoid having two lints for the same issue.
+            if !is_lint_allowed(cx, REF_PATTERNS, arg.pat.hir_id) {
+                return;
+            }
             if let PatKind::Binding(BindingAnnotation(ByRef::Yes, _), ..) = arg.pat.kind {
                 span_lint(
                     cx,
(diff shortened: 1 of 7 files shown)


Alexendoo:
> * The dogfood test is currently failing locally, though I ran `cargo clippy -- -D clippy::all -D clippy::pedantic` and got no output, so I'm not sure why this is.

The clippy packages aren't in a cargo workspace due to being part of the `rust-lang/rust` repo workspace so that would only be linting the root crate, you can run the dogfood test with `cargo test --features internal --test dogfood`

NotAPenguin0:
I see, thanks. I'll run that and see what I can fix.

Alexendoo:
> Maybe it's desirable that a quick fix is listed too, instead of a generic `avoid using the ref keyword` lint.

It's hard to produce a specific suggestion since it's often not as simple as putting an `&` somewhere in the expression, e.g.

```rust
let x = (String::new(), String::new());
let (a, ref b) = x;
```

> `let ref x = y` already has a lint (`clippy::toplevel_ref_arg`). This implementation will report this too, so you get two lints for the same issue, which is not great. I don't really know a way around this though.

You could check where `toplevel_ref_arg` is emitted if this lint is enabled with [`!is_lint_allowed(cx, REF_PATTERN, pat.hir_id)`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_lint_allowed.html)

clippy_lints/src/misc.rs
```diff
@@ -252,6 +263,10 @@ impl<'tcx> LateLintPass<'tcx> for LintPass {
             // Don't lint things expanded by #[derive(...)], etc or `await` desugaring
             return;
         }
+        // Do not emit if clippy::ref_pattern is not allowed to avoid having two lints for the same issue.
+        if !is_lint_allowed(cx, REF_PATTERN, expr.hir_id) {
+            return;
+        }
```
Alexendoo, line 269: This one appears to be above `used_underscore_binding` which would be unrelated

With that + a rebase/squash this looks good to merge to me!

NotAPenguin0:
Should I do anything on my end to squash and rebase?

Alexendoo:
Yeah though actually one last thing also, could you change the lint + filenames to be `ref_patterns` (plural)

For the rebase/squash it would be something along the lines of

```bash
git fetch upstream
git rebase -i upstream/master
```

Then changing the `pick`s of all but your first commit to be `fixup` (or `f` for short), saving + exiting and resolving any conflicts that arise

Alexendoo (approved):

Issue #10739, opened 2023-05-04
Generalize string literal inlining into format-like macros


nyurik:
### Description

A common pattern `format!("...{}...", "foo")` has a number of related lints like:
* [print_literal](https://rust-lang.github.io/rust-clippy/master/index.html#print_literal)  `println!("{}", "foo");`
* [write_literal](https://rust-lang.github.io/rust-clippy/master/index.html#write_literal)  `writeln!(buf, "{}", "foo");`

we also have the [uninlined_format_args](https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args) which is more generic, but focuses on variables, not string literals.

I think these approaches should be combined, either as part of the `uninlined_format_args`, or as a new lint (and possibly sunset/remove the print/write literal?).

Mixed cases like `format!("{}{}", foo, "bar")` may cause issues with multiple lints trying to fix the same, so a single lint might be better, but uninlined_format_args focuses on vars, so expanding its scope might be confusing? Although technically this is also inlining an argument, so might be ok?

### Version

```text
rustc 1.69.0 (84c898d65 2023-04-16)
binary: rustc
commit-hash: 84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc
commit-date: 2023-04-16
host: x86_64-unknown-linux-gnu
release: 1.69.0
LLVM version: 15.0.7
```

voodookiidoo:
Good evening dear friend! Is this issue/pr still in progress? I was thinking if i could pick up this issue, since i started experimenting with that idea, and then found out that i was about to duplicate one's work. If you need any assist, please let me know, i'd be happy to work in a duet! If you don't wish to work on it, or don't have enough time - i could pick it up for you

CommanderStorm:
Not really progress, no.

The following is the last PRs that was broken out of the effort to make the `useless_fmt`-ish lints more powerfull.

- https://github.com/rust-lang/rust-clippy/pull/16595
- https://github.com/rust-lang/rust-clippy/pull/16540

This is the PR with the actual issues noted in this issue:
- https://github.com/rust-lang/rust-clippy/pull/10740

In general in OSS, don't ask for permission to do experiment/have fun.
The worst thing that can happen is that we make clippy too smart 😉 🤖 

Context: Yuri asked me offline half-ish year ago if this lint area is something I might have interest in this-ish area , I think it is save to say that you are not stepping on anyones shoes.

voodookiidoo:
[...]

PR #10740, opened 2023-05-04
WIP format literal arg inlining

nyurik:
A rough draft of #10739 implementation. The goal is to allow any kind of literal string arguments to be inlined automatically.

```rust
format!("hello {}", "world")
// is replaced with
format!("hello world")
```

```
changelog: [`uninlined_format_args`]: support for inlining string literals
```

### DIFF
--- a/clippy_lints/src/format_args.rs
+++ b/clippy_lints/src/format_args.rs
@@ -1,3 +1,4 @@
+use crate::write::extract_str_literal;
 use arrayvec::ArrayVec;
 use clippy_config::msrvs::{self, Msrv};
 use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
@@ -9,6 +10,7 @@ use clippy_utils::macros::{
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::{implements_trait, is_type_lang_item};
 use itertools::Itertools;
+use rustc_ast::token::LitKind;
 use rustc_ast::{
     FormatArgPosition, FormatArgPositionKind, FormatArgsPiece, FormatArgumentKind, FormatCount, FormatOptions,
     FormatPlaceholder, FormatTrait,
@@ -324,8 +326,8 @@ impl<'a, 'tcx> FormatArgsExpr<'a, 'tcx> {
         // we cannot remove any other arguments in the format string,
         // because the index numbers might be wrong after inlining.
         // Example of an un-inlinable format:  print!("{}{1}", foo, 2)
-        for (pos, usage) in self.format_arg_positions() {
-            if !self.check_one_arg(pos, usage, &mut fixes) {
+        for (pos, usage, pl) in self.format_arg_positions() {
+            if !self.check_one_arg(pos, usage, pl, &mut fixes) {
                 return;
             }
         }
@@ -360,9 +362,19 @@ impl<'a, 'tcx> FormatArgsExpr<'a, 'tcx> {
         );
     }
 
-    fn check_one_arg(&self, pos: &FormatArgPosition, usage: FormatParamUsage, fixes: &mut Vec<(Span, String)>) -> bool {
+    fn check_one_arg(
+        &self,
+        pos: &FormatArgPosition,
+        usage: FormatParamUsage,
+        placeholder: Option<&FormatPlaceholder>,
+        fixes: &mut Vec<(Span, String)>,
+    ) -> bool {
         let index = pos.index.unwrap();
-        let arg = &self.format_args.arguments.all_args()[index];
+        // If the argument is not found by its index, something is weird, ignore and stop processing this
+        // case.
+        let Some(arg) = &self.format_args.arguments.by_index(index) else {
+            return false;
+        };
 
         if !matches!(arg.kind, FormatArgumentKind::Captured(_))
             && let rustc_ast::ExprKind::Path(None, path) = &arg.expr.kind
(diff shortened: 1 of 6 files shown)


clippy_lints/src/format_args.rs
```diff
@@ -375,6 +381,21 @@ fn check_one_arg(
         fixes.push((pos_span, replacement));
         fixes.push((arg_span, String::new()));
         true  // successful inlining, continue checking
+    } else if !matches!(arg.kind, FormatArgumentKind::Captured(_))
+        && let Some(FormatPlaceholder{span: Some(placeholder_span), ..}) = placeholder
+        && let rustc_ast::ExprKind::Lit(lit) = &arg.expr.kind
+        && lit.kind == LitKind::Str // FIXME: lit.kind must match the format string kind
```
llogiq, line 387: This isn't clear to me. What other kinds would there be?

tests/ui/uninlined_format_args.fixed
```diff
@@ -262,3 +262,17 @@ fn tester2() {
         local_i32,
     };
 }
+
+fn literals() {
```
llogiq, line 266: I worry about cases where the format string and literal value come from different macro expansions. We should have a test where that is the case.

tests/ui/uninlined_format_args.stderr
```diff
@@ -177,6 +177,18 @@ LL -     println!("{:.1}", local_f64);
...
+   |
+LL |     println!("Hello {} is {:.*}", "x", local_i32, local_f64);
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+help: change this to
+   |
+LL -     println!("Hello {} is {:.*}", "x", local_i32, local_f64);
+LL +     println!("Hello x is {local_f64:.local_i32$}");
+   |
+
```
llogiq, line 191: Looks pretty good. I have just one request for clarification and I'd like to see a test with macros.

llogiq (commented):
Looks good to me. I have but one request for clarification and I'd like to see a test with macros.

clippy_lints/src/format_args.rs
nyurik: there are also [raw strings](https://doc.rust-lang.org/reference/tokens.html#raw-string-literals)

nyurik: And possibly byte strings too?  Not certain

tests/ui/uninlined_format_args.fixed
nyurik: Yeah, i will need tons more of tests for this - some already exist in
* https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/write_literal.rs
* https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/write_literal_2.rs
* https://github.com/rust-lang/rust-clippy/blob/master/tests/ui/write_with_newline.rs

Could you give an example of what specific macro expansion where you thinking about?

Alexendoo:
I'd say this should be under a new lint name, although it's similar what it's doing is different enough to the existing `uninlined_format_args`

The existing implementation can also be used, it handles the various raw/non-raw permutations of the literal and format string as well as some other literals. You'd have to change the part where it checks the macro names before it's called + do a lint rename for both `print_literal` and `write_literal` to the new one

nyurik:
@Alexendoo after some more thinking, it seems separating it into a new lint would create more problems than solve:
* multiple lints operating on the same expression tend to cause issues, esp in the `--fix` mode
* a format expression with numeric indexes would be ignored by both lints unless used together: `format!("{0} {1}", var, "str")` will be ignored by both the `uninlined_format_args` and the string inlining lint, but when used together it would suggest the `format!("{var} str")`
* the original lint name `uninlined_format_args` fits perfectly for the string inlining as well, so wouldn't cause any confusion

So in short, we would not really gain anything besides problems from multiple lints...  Not worth it IMO :)

Alexendoo:
I don't see what problem that would cause here, overlapping suggestions would cause rustfix ui tests to fail (`cargo fix` can handle them though) but the suggestions here wouldn't have overlapping spans

The main thing is so that you can set the lint levels independently, e.g. `print_literal` and `write_literal` are `style` but `uninlined_format_args` is `pedantic`. It wouldn't be unusual to want to disable one independently if they were in the same group 

It would also make the lint documentation more confusing to have to explain the two different kinds of inlining in one lint, and how `allow-mixed-uninlined-format-args` would only apply to one of those kinds

`format!("{0} {1}", var, "str")` doesn't seem likely enough to occur to make up for that

nyurik:
@Alexendoo you do raise important config point.  I think we can actually achieve both goals if we track the type of expansion needed -- e.g. if both are enabled, and both are needed (the `"literal", var` case above), we can show just one suggestion (under either of the names, or even add a note saying that both were involved).  If only one is enabled, it won't even consider the other in the expansion.

Alexendoo:
To me that sounds too complicated/novel considering the problem is relatively minor

llogiq:
Agreed, let's not complicate the solution. As long as both lints don't step on each other's toes by issuing incompatible suggestions, having two lints that don't care about each other is fine.

xFrednet:
Hey @nyurik, this is a ping from triage. Do you still plan to continue this PR? If you need help, you're always welcome to ask questions here or on Zulip :)

xFrednet:

PR #10747, opened 2023-05-05
Inherit stdout/stderr for `cargo dev dogfood`

Alexendoo:
changelog: none

Prints progress as it happens and in colour, and will also show anything printed to stderr

### DIFF
--- a/clippy_dev/src/dogfood.rs
+++ b/clippy_dev/src/dogfood.rs
@@ -1,4 +1,4 @@
-use crate::clippy_project_root;
+use crate::{clippy_project_root, exit_if_err};
 use std::process::Command;
 
 /// # Panics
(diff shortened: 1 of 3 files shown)


flip1995:
> Prints progress as it happens and in colour, and will also show anything printed to stderr

I didn't see any progress printing on CI and also don't understand how the committed changes should achieve this. Is this only done if there is something emitted to `stderr`?

I think this is a code improvement nonetheless, so r=me, if I missed something obvious.

Alexendoo:
For CI it wouldn't show anything different, but `cargo dev dogfood` was hiding the output of the `cargo test --test dogfood ...` until it was completed, so you wouldn't see the initial clippy + test compilation progress

I've now added `--nocapture` so the test output can be seen

I tried doing the same thing in the dogfood test itself but that seemed to print stuff even when `--nocapture` wasn't passed in which is weird so I reverted that

flip1995:
I see, thanks for the explanation. 

@bors r+

Alexendoo:
Huh, tests capture print macros and panics [by asking `std` to do it](https://doc.rust-lang.org/src/std/io/stdio.rs.html#19-22) 😨

So other ways of writing to stdout/stderr [aren't captured](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8526aa2c1adb0ed704b2edbdab3eccf9)

PR #10752, opened 2023-05-05
Ignore expressions from macros in `default_constructed_unit_structs`

Alexendoo:
changelog: none, should be the same release as the lint itself


giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #10726, opened 2023-04-30
redundant_pattern_matching does not check for match with single arm


disco07:
### Summary

Use of match, matching `Result` or `Option`;  do not trigger `clippy::redundant_pattern_matching`

### Reproducer

I tried this code:

```rust
match Err::<i32, i32>(55) {
        Ok(_) => true,
        _ => false,
    };
```

I expected to see this happen:
```rust
Err::<i32, i32>(55).is_ok();
```

Instead, this happened:
```rust
match Err::<i32, i32>(55) {
        Ok(_) => true,
        _ => false,
    };
```

### Version

```text
rustc 1.71.0-nightly (b628260df 2023-04-22)
binary: rustc
commit-hash: b628260df0587ae559253d8640ecb8738d3de613
commit-date: 2023-04-22
host: x86_64-unknown-linux-gnu
release: 1.71.0-nightly
LLVM version: 16.0.2
```

disco07:
@xFrednet

xFrednet:
Hey @disco07, thank you for the ticket :). Do you have a specific question related to it :)

disco07:
**@xFrednet** Can I work on this ?

xFrednet:
Yes, you're very welcome to! As a start, you can take a look at the lint code: https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/matches/redundant_pattern_match.rs

And look at the AST structure of the code you want to detect. For this, I recommend using `#[clippy::dump]` and running `tool > clippy`. See [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4280d721d559f645ef063ca6f3a92924). Example actually triggers the [`match_like_matches_macro`](https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro) lint.

I guess that the lint ignores the match, as it expects the `Ok` and `Err` branches, and not a wildcard. Maybe you just need to add a check for that :)

PR #10753, opened 2023-05-05
redundant_pattern_matching

disco07:
This PR try to solve this issue https://github.com/rust-lang/rust-clippy/pull/10726, 
but it enter in conflict with another test.

changelog: none

Try to test this:
```
let _w = match x {
     Some(_) => true,
     _ => false,
};
```

this happen:
```
 error: match expression looks like `matches!` macro
   --> $DIR/match_expr_like_matches_macro.rs:21:14
    |
 LL |       let _w = match x {
    |  ______________^
 LL | |         Some(_) => true,
 LL | |         _ => false,
 LL | |     };
    | |_____^ help: try this: `matches!(x, Some(_))`
 
+error: redundant pattern matching, consider using `is_some()`
+  --> $DIR/match_expr_like_matches_macro.rs:21:14
+   |
+LL |       let _w = match x {
+   |  ______________^
+LL | |         Some(_) => true,
+LL | |         _ => false,
+LL | |     };
+   | |_____^ help: try this: `x.is_some()`
+   |
+   = note: `-D clippy::redundant-pattern-matching` implied by `-D warnings`
+
```
I need some help to fix this. Thanks

### DIFF
--- a/clippy_lints/src/matches/match_like_matches.rs
+++ b/clippy_lints/src/matches/match_like_matches.rs
@@ -185,7 +185,7 @@ fn is_some(path_kind: PatKind<'_>) -> bool {
     match path_kind {
         PatKind::TupleStruct(QPath::Resolved(_, path), patterns, _) if is_wild(&patterns[0]) => {
             let name = path.segments[0].ident;
-            if name.as_str() == "Some" {
+            if name.name == rustc_span::sym::Some {
                 return true;
             }
             false
--- a/clippy_lints/src/matches/match_like_matches.rs
+++ b/clippy_lints/src/matches/match_like_matches.rs
@@ -183,15 +183,12 @@ fn find_bool_lit(ex: &ExprKind<'_>) -> Option<bool> {
 
 fn is_some(path_kind: PatKind<'_>) -> bool {
     match path_kind {
-        PatKind::TupleStruct(ref path_left, patterns, _) if is_wild(&patterns[0]) => match path_left {
-            QPath::Resolved(_, path) => {
-                let name = path.segments[0].ident;
-                if name.as_str() == "Some" {
-                    return true;
-                }
-                return false;
-            },
-            _ => false,
+        PatKind::TupleStruct(QPath::Resolved(_, path), patterns, _) if is_wild(&patterns[0]) => {
+            let name = path.segments[0].ident;
+            if name.as_str() == "Some" {
+                return true;
+            }
+            false
         },
         _ => false,
     }
--- a/clippy_lints/src/matches/match_like_matches.rs
+++ b/clippy_lints/src/matches/match_like_matches.rs
@@ -1,10 +1,12 @@
+use super::REDUNDANT_PATTERN_MATCHING;
 use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::is_lint_allowed;
 use clippy_utils::is_wild;
 use clippy_utils::source::snippet_with_applicability;
 use clippy_utils::span_contains_comment;
 use rustc_ast::{Attribute, LitKind};
 use rustc_errors::Applicability;
-use rustc_hir::{Arm, BorrowKind, Expr, ExprKind, Guard, Pat};
+use rustc_hir::{Arm, BorrowKind, Expr, ExprKind, Guard, Pat, PatKind, QPath};
 use rustc_lint::{LateContext, LintContext};
 use rustc_middle::ty;
 use rustc_span::source_map::Spanned;
@@ -99,6 +101,14 @@ where
                 }
             }
 
+            for arm in iter_without_last.clone() {
+                if let Some(pat) = arm.1 {
+                    if !is_lint_allowed(cx, REDUNDANT_PATTERN_MATCHING, pat.hir_id) && is_some(pat.kind) {
+                        return false;
+                    }
+                }
+            }
+
             // The suggestion may be incorrect, because some arms can have `cfg` attributes
             // evaluated into `false` and so such arms will be stripped before.
             let mut applicability = Applicability::MaybeIncorrect;
@@ -168,5 +178,21 @@ fn find_bool_lit(ex: &ExprKind<'_>) -> Option<bool> {
             }
         },
         _ => None,
+    }
+}
+
+fn is_some(path_kind: PatKind<'_>) -> bool {
+    match path_kind {
+        PatKind::TupleStruct(ref path_left, patterns, _) if is_wild(&patterns[0]) => match path_left {
+            QPath::Resolved(_, path) => {
+                let name = path.segments[0].ident;
+                if name.as_str() == "Some" {
+                    return true;
+                }
+                return false;
+            },
+            _ => false,
+        },
+        _ => false,
     }
 }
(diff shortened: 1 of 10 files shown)


disco07:
==@xFrednet== can you help me please ?

xFrednet:
> @xFrednet can you help me please ?

Sure, I'm guessing, your question is how to prevent the other lint from triggering in this case? In the test files, you could just use `#[allow(other_lint)]`, but that wouldn't be the ideal solution here, as the user would still see two lints. Instead, you can take a look at [`is_lint_allowed`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_lint_allowed.html#). In this case, I recommend adding a check to `match_expr_like_matches_macro`, that suppresses the lint, if `redundant_pattern_matching` is currently allowed.

Does this explanation make sense? :)

disco07:
> > @xFrednet can you help me please ?
> 
> Sure, I'm guessing, your question is how to prevent the other lint from triggering in this case? In the test files, you could just use `#[allow(other_lint)]`, but that wouldn't be the ideal solution here, as the user would still see two lints. Instead, you can take a look at [`is_lint_allowed`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_lint_allowed.html#). In this case, I recommend adding a check to `match_expr_like_matches_macro`, that suppresses the lint, if `redundant_pattern_matching` is currently allowed.
> 
> Does this explanation make sense? :)

Thank you so much I will try this :-)

disco07:
@xFrednet Can you check my changes if you have some time please :) 
Thanks in advance

xFrednet:
Sure, I'll take a look at it later today or tomorrow. :)

r? @xFrednet

clippy_lints/src/matches/match_like_matches.rs
```diff
@@ -170,3 +180,16 @@ fn find_bool_lit(ex: &ExprKind<'_>) -> Option<bool> {
         _ => None,
     }
 }
+
+fn is_some(path_kind: PatKind<'_>) -> bool {
+    match path_kind {
+        PatKind::TupleStruct(QPath::Resolved(_, path), patterns, _) if is_wild(&patterns[0]) => {
```
xFrednet, line 186: Optional suggestion: Here you can destruct the patterns slice directly, just a small quality of life thing and to avoid panics from indexing, even though this should be 100 save
```suggestion
        PatKind::TupleStruct(QPath::Resolved(_, path), [first, ..], _) if is_wild(first) => {
```

clippy_lints/src/matches/match_like_matches.rs
```diff
@@ -170,3 +180,16 @@ fn find_bool_lit(ex: &ExprKind<'_>) -> Option<bool> {
...
+            let name = path.segments[0].ident;
+            if name.name == rustc_span::sym::Some {
+                return true;
+            }
+            false
```
xFrednet, line 191: Here you can return the expression result directly, but we should also check for `None`. I believe it would look like this:

```suggestion
            name.name == rustc_span::sym::Some | name.name == rustc_span::sym::None
```

clippy_lints/src/matches/redundant_pattern_match.rs
```diff
@@ -188,72 +188,7 @@ fn find_sugg_for_if_let<'tcx>(
...
-                    })
-                } else {
-                    None
-                }
-            },
-            _ => None,
-        };
+        let found_good_method = found_good_method(cx, arms, node_pair);
 
         if let Some(good_method) = found_good_method {
```
xFrednet, line 193: Now you can inline the call directly

```suggestion
        if let Some(good_method) = found_good_method(cx, arms, node_pair) {
```

xFrednet (commented):
This already looks good overall, I have a few NITs, where I would appreciate it if you could update them. But then it'll be ready to be merged :)

disco07:
Thanks @xFrednet :)

tests/ui/redundant_pattern_matching_result.rs
```diff
@@ -125,3 +126,30 @@ const fn issue6067() {
...
+
+    match Ok::<i32, i32>(42) {
+        Err(_) => true,
+        _ => false,
+    };
+
+    match Err::<i32, i32>(42) {
+        Ok(_) => true,
+        _ => false,
+    };
```
xFrednet, line 144: These lints only check the patterns and branches, the scrutinee (Expression used by the `match` to match) is ignored by the lint. The first and last match in the marked code are equivalent. Could you instead use the following tests:

```
    // This is optional, but it makes the examples easier
    let x: Result<i32, i32> = Ok(42)

    match x {
        Ok(_) => true,
        _ => false,
    };

    match x {
        Ok(_) => false,
        _ => true,
    };

    match x {
        Err(_) => true,
        _ => false,
    };

    match x {
        Err(_) => false,
        _ => true,
    };

    // Don't lint
    match x {
        Err(16) => false,
        _ => true,
    };

    // Don't lint
    match x {
        Ok(16) => false,
        _ => true,
    };
```

And also adjust these for the `Option` tests

xFrednet (commented):
Thank you for the swift update. I still have a NIT when it comes to the tests, but that should be a simple update. Once that is done, we can merge it. Could you also squash your fix commits, to only have one or two?

---

PS, I'll be gone over the weekend, so the next review will be done some time next week :)

tests/ui/redundant_pattern_matching_option.rs
```diff
@@ -103,3 +104,34 @@ fn issue7921() {
...
+
+    match y {
+        Some(_) => true,
+        _ => false,
+    };
+
+    match y {
+        None => true,
+        _ => false,
+    };
```
xFrednet, line 130: These match statements are still equivalent to the ones above

```suggestion
    match x {
        Some(_) => false,
        _ => true,
    };

    match x {
        None => false,
        _ => true,
    };
```

Using y doesn't have any benefit here.

tests/ui/redundant_pattern_matching_result.rs
```diff
@@ -125,3 +126,40 @@ const fn issue6067() {
         Err(_) => true,
     };
 }
+
+fn issue10726() {
+    // This is optional, but it makes the examples easier
+    let x: Result<i32, i32> = Ok(42);
+
+    match x {
```
xFrednet, line 134: These tests are good now!

xFrednet (commented):
The `Option` tests still need some adjustments. Then it should be ready.

Have you seen my request to squash your commits, from the last review?
> Could you also squash your fix commits, to only have one or two?

disco07:
> The `Option` tests still need some adjustments. Then it should be ready.
> 
> Have you seen my request to squash your commits, from the last review?
> 
> > Could you also squash your fix commits, to only have one or two?

I hope I made the right changes this time ;)

xFrednet (approved):

xFrednet:
This version looks good to me, thank you for the PR. I hope you had fun working on Clippy :)

@bors r+

disco07:
@xFrednet I update my PR to add changelog: none

xFrednet:
No problem :)

@bors retry

PR #10759, opened 2023-05-06
Now `option_env_unwrap` warns even if a variable isn't set at compiletime

### LINKED ISSUE #10742: `option_env_unwrap` doesn't warn if the env variable isn't set
### Summary

`option_env_unwrap` only lints if the variable is currently set in the environment. This makes it hard to reproduce clippy output between machines; for example, CI often sets test variables that aren't set locally. Clippy should warn about these consistently, even if the variables aren't set in the environment.

Here's a slightly more realistic example of why it would be useful to warn:
```rust
#[test]
fn test_simple() {
  // ...
}

#[test]
#[ignored = "needs credentials"]
fn test_extended() {
  let credentials = option_env!("CREDENTIALS").unwrap();
  // ...
}
```

### Lint Name

option_env_unwrap

### Reproducer

I tried this code:

```rust
#[allow(unused)]
fn foo() {
    let x = option_env!("CARGO").unwrap();
    let y = option_env!("Y").unwrap();
}
```

I expected to see this happen: Both `CARGO` and `Y` have the lint emitted.

Instead, this happened: Only `CARGO` had the lint emitted.

### Version

```text
Clippy 0.1.70 (2023-05-03 473f916)
```

blyxyas:
Fixes #10742
changelog: Fix false negative where `option_env_unwrap` wouldn't warn if the env variable isn't set at compile-time.

### DIFF
--- a/clippy_lints/src/option_env_unwrap.rs
+++ b/clippy_lints/src/option_env_unwrap.rs
@@ -1,10 +1,9 @@
 use clippy_utils::diagnostics::span_lint_and_help;
 use clippy_utils::is_direct_expn_of;
-use if_chain::if_chain;
 use rustc_ast::ast::{Expr, ExprKind, MethodCall};
 use rustc_lint::{EarlyContext, EarlyLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::sym;
+use rustc_span::{sym, Span};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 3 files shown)


blyxyas:
Note that I needed to undo some "stylistic choices" (remove the `if_chain!`) because Rust pattern matching won't let you have 2 patterns with different types in the same `if let` statement.

That means that the following won't compile:

```rs
if let ExprKind::Call(caller, _) | ExprKind::Path(_, caller) = receiver.kind;
```

Because `Call`'s `caller` and `Path`'s `caller` are different types.

blyxyas:
As @flip1995 is pretty busy, I'll let rustbot choose another reviewer.

r? rust-lang/clippy

tests/ui/option_env_unwrap.rs
```diff
@@ -9,6 +9,7 @@ use proc_macros::{external, inline_macros};
 fn main() {
     let _ = option_env!("PATH").unwrap();
     let _ = option_env!("PATH").expect("environment variable PATH isn't set");
+    let _ = option_env!("Y").unwrap(); // This test only works if you don't have a __Y__do_not_use env variable in your environment.
```
flip1995, line 12: Is the comment outdated and should be

```suggestion
    let _ = option_env!("Y").unwrap(); // This test only works if you don't have a Y env variable in your environment.
```

clippy_lints/src/option_env_unwrap.rs
```diff
@@ -36,21 +35,24 @@ declare_lint_pass!(OptionEnvUnwrap => [OPTION_ENV_UNWRAP]);
...
+            );
         }
+
+        if let ExprKind::MethodCall(box MethodCall { seg, receiver, .. }) = &expr.kind &&
+		matches!(seg.ident.name, sym::expect | sym::unwrap) {
+			if let ExprKind::Call(caller, _) = &receiver.kind && is_direct_expn_of(caller.span, "option_env").is_some() {
+				lint(cx, expr.span);
+			} else if let ExprKind::Path(_, caller) = &receiver.kind && is_direct_expn_of(caller.span, "option_env").is_some() {
+				lint(cx, expr.span);
+			}
```
flip1995, line 55: Can you add 2 short comments here, what cases are covered by Call vs Path?

flip1995:
@bors r+

Thanks!

PR #10760, opened 2023-05-07
Standard lint formulations

### LINKED ISSUE #10660: Check for lint formulations that should be standardized
### What it does

Checks if lint formulations are very close to a standard existing one. For now I can only think of "Checks for usage of..." but this could be extended to other frequent formulations.

This is a follow-up to #10631, where @xFrednet kindly suggested to create a custom lint for the fix I had to do manually.

### Lint Name

almost_standard_lint_formulation

### Category

internal

### Advantage

- Keeps lint formulations consistent
- Reduces the overhead of having to come up with a good lint formulation for new lints
- Reduces the chances of ungrammatical lint formulations

### Drawbacks

Less freedom to formulate lints however you like

### Example

```rust
declare_clippy_lint! {
    /// ### What it does
    /// Checks for using...
}
```

Could be written as:

```rust
declare_clippy_lint! {
    /// ### What it does
    /// Checks for usage of...
}
```

NanthR:
A WIP that fixes #10660. Fix lints that don't conform to the standard formulation. 

changelog: none


clippy_lints/src/utils/internal_lints/almost_standard_lint_formulation.rs
```diff
@@ -0,0 +1,77 @@
...
+                            let starts_with_correct_formulation = cur_line.starts_with(formulation.correction);
+                            if !starts_with_correct_formulation && formulation.wrong_pattern.is_match(cur_line) {
+                                if let Some(ident) = attr.ident() {
+                                    span_lint_and_help(
+                                        cx,
+                                        ALMOST_STANDARD_LINT_FORMULATION,
+                                        ident.span,
+                                        "invalid documentation formulation",
+                                        None,
+                                        &format!("try using {} instead", formulation.correction),
```
xFrednet, line 59:
```suggestion
                                        "non-standard lint description",
                                        None,
                                        &format!("try using `{}` instead", formulation.correction),
```

tests/ui-internal/check_formulation.rs
```diff
@@ -0,0 +1,43 @@
...
+    Warn,
+    "One",
+    report_in_external_macro: true
+}
+
+declare_tool_lint! {
+    /// # What it does
+    /// Detects uses of incorrect formulations
+    #[clippy::version = "pre 1.29.0"]
+    pub clippy::INVALID2,
```
xFrednet, line 35: Can you also add a test to make sure that `#[allow(clippy::almost_standard_lint_formulation)]` works on the lint, just in case we want to allow it in some edge cases?

xFrednet (commented):
The lint implementation is already a good step in the right direction. I have a few suggestions, but nothing major :)

xFrednet:
I belive, I can safely steal this review :)

r? @xFrednet

clippy_lints/src/utils/internal_lints/almost_standard_lint_formulation.rs
```diff
@@ -0,0 +1,77 @@
...
+            correction: "Checks for",
+        }];
+        let mut check_next = false;
+        if let ItemKind::Static(ty, Mutability::Not, _) = item.kind {
+            let lines = cx
+                .tcx
+                .hir()
+                .attrs(item.hir_id())
+                .iter()
+                .filter_map(|attr| ast::Attribute::doc_str(attr).map(|sym| (sym, attr)));
```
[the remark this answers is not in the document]
xFrednet: Ahh, yes you're right. Then this is a better solution :)

tests/ui-internal/allow_incorrect_lint_formulations.rs
```diff
@@ -0,0 +1,22 @@
+#![allow(clippy::almost_standard_lint_formulation)]
```
xFrednet, line 1: I would like this attribute to be more fine-grained, we'll likely only want to enable it on a lint by lint level, and not on a module level like here. Can you try attaching it to the lint name directly:

```
#[allow(clippy::almost_standard_lint_formulation)]
pub clippy::ALLOWED_INVALID,
```

then you can also move it into the same test file as the other tests :)

xFrednet (commented):
A small nit regarding the requested test case, but then I believe this should be ready for merge. Are there any additions you want to make, since the PR is still in draft mode? :)

You can also push the new changes in a separate commit :)

xFrednet (approved):

xFrednet:
Perfect, this version looks good to me. Thank you for the work you put into this!

@bors r+

xFrednet:
Never knew bors checked for that ^^

@bors r+

PR #10768, opened 2023-05-10
[arithmetic_side_effects] Consider referenced allowed or hard-coded types

### LINKED ISSUE #10767: [`arithmetic_side_effects`] Referenced allowed or hard-coded types should not fire the lint
```rust
pub fn float_arith_ref() {
    3.1_f32 + &1.2_f32;
    &3.4_f32 + 1.5_f32;
    &3.5_f32 + &1.3_f32;
}

pub fn float_foo(f: &f32) -> f32 {
    f + f
}
```

c410-f3r:
Fix #10767

```
changelog: [`arithmetic_side_effects`]: Do not fire when dealing with allowed or hard-coded types that are referenced.
```


Jarcho:
This should be remove the references from the type before converting it to a string. A type with a named lifetime has it appear in the string.

c410-f3r:
`peel_hir_expr_refs` is being called before the conversion to string in `has_allowed_binary`, which I am assuming that already  strips all types of lifetimes. Should I use another method?

Jarcho:
`peel_hir_expr_refs` only strips away the `&` operator, not all references on the type. There's a `peel_refs` function on `Ty` you can use.

c410-f3r:
Done. Thank you @Jarcho

Jarcho:
Thank you. @bors r+

PR #10769, opened 2023-05-11
add lint `manual_next_back`

### LINKED ISSUE #10274: suggestion: .rev().next()
### What it does

Suggest `.next_back()` insteaad of `.rev().next()` when that's equivalent.

### Lint Name

iter_rev_next

### Category

pedantic

### Advantage

People will point it out at code review anyway. Might as well spare them some typing

### Drawbacks

I haven't thought about edge cases where applying the suggestion would not be correct

### Example

```rust
foo.iter().rev().next()
```

Could be written as:

```rust
foo.iter().next_back()
```

Icxolu:
changelog: [`manual_next_back`]: checks for manual reverse iteration (`.rev().next()`) of a `DoubleEndedIterator`

fixes #10274


giraffate:
@bors r+

Thanks!

Issue #9399, opened 2022-08-30
[New Lint Request] Suggest explict `drop` call for every sync lock


c410-f3r:
### What it does

Is there a lint that already does such thing or at least a similar issue?

## Problem

Code that heavily depends on synchronous primitives like `Mutex` can suffer unnecessary resource contention if the lock is only released and the end of its scope through `Drop` but could in fact be released early.

```rust
// Could be rewritten as `let owned_rslt = some_sync_resource.lock().do_stuff_with_resource();`
fn inline_lock() {
  let locked = some_sync_resource.lock();
  let owned_rslt = locked.do_stuff_with_resource();
  // Only `owned_rslt` is needed but `locked` is still held
  do_heavy_computation_that_takes_time(owned_rslt);
}

// When the lock is used across many lines, `drop` should be called directly.
fn explict_drop() {
  let locked = some_sync_resource.lock();
  let reference = locked.get_inner_resource_ref();
  let _ = reference.do_stuff();
  let _ = reference.do_other_things();
  // From this point forward, `locked` is not needed anymore
  let _ = compute_foo();
  let _ = compute_bar();
  let _ = compute_baz();
}
```

## Implementation

It is trivial to detect synchronous primitives of the standard library but third-party implementations are tricky to detect. What should be done here? Hard code a list of well-known crates or allow users to provide a list of custom types?

### Lint Name

sync_drop_call

### Category

pedantic

### Example

```rust
<code>
```

Could be written as:

```rust
<code>
```

flip1995:
To implement this, the `has_significant_drop` attribute, that can be attached to types could be used, like the [`clippy::significant_drop_in_scrutinee`](https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_in_scrutinee) lint does.

However, I marked it as `E-hard` because keeping track of references build from the `locked` variables. So checking for the last use of `locked` should be easy. But checking that `reference` (or other bindings) aren't used anymore might be hard.

Thomasdezeeuw:
I've hit a false positive, or at least a bad suggestion, with this lint using `clippy 0.1.72 (f4b80ca 2023-06-30)`.

The related code: 
https://github.com/Thomasdezeeuw/heph/blob/8725fe9ed871f135b7ab87b5969eb1213121a242/inbox/src/lib.rs#L523-L529

For which Clippy suggests:

```
error: temporary with significant `Drop` can be early dropped
   --> inbox/src/lib.rs:523:21
[...]

PR #10774, opened 2023-05-14
[significant_drop_tightening] Fix #10413

### LINKED ISSUE #10413: Incorrect suggestion for has_significant_drop
### Summary

The has_significant_drop lint only (seems to) considers the first usage of a locked value and suggests an incorrect correction of the code.

### Lint Name

has_significant_drop

### Reproducer

The code comes from https://github.com/Thomasdezeeuw/heph/blob/8be6cfa97d6870cc7e086e63e32608976c6d76f1/inbox/src/lib.rs#L519-L529. The relevant part:

```rust
    // self.registered_waker: Option<task::Waker>
    // self.channel.sender_wakers: Mutex<Vec<task::Waker>>
        if let Some(waker) = self.registered_waker.take() {
            let mut sender_wakers = self.channel.sender_wakers.lock().unwrap();
            let idx = sender_wakers.iter().position(|w| w.will_wake(&waker));
            if let Some(idx) = idx {
                let waker = sender_wakers.swap_remove(idx);
                unlock(sender_wakers);
                drop(waker);
            }
        }
```

The suggestion Clippy makes:

```
error: temporary with significant `Drop` can be early dropped
   --> inbox/src/lib.rs:523:21
    |
522 |           if let Some(waker) = self.registered_waker.take() {
    |  ___________________________________________________________-
523 | |             let mut sender_wakers = self.channel.sender_wakers.lock().unwrap();
    | |                     ^^^^^^^^^^^^^
524 | |             let idx = sender_wakers.iter().position(|w| w.will_wake(&waker));
525 | |             if let Some(idx) = idx {
...   |
529 | |             }
530 | |         }
    | |_________- temporary `sender_wakers` is currently being dropped at the end of its contained scope
    |
    = note: this might lead to unnecessary resource contention
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_tightening
    = note: `-D clippy::significant-drop-tightening` implied by `-D clippy::nursery`
help: merge the temporary construction with its single usage
    |
523 ~
524 +             let idx = self.channel.sender_wakers.lock().unwrap().position(|w| w.will_wake(&waker));
    |
help: remove separated single usage
    |
524 -             let idx = sender_wakers.iter().position(|w| w.will_wake(&waker));
524 +
    |
```

We can't move the lock into the line with `idx` as `sender_wakers` (the locked value) is used later on. Clippy however doesn't seem to see/understand this.

I guess the basic problem is that Clippy is only considering a single use of the locked value, while in the code above their are two uses.

### Version

```text
rustc 1.69.0-nightly (d962ea578 2023-02-26)
binary: rustc
commit-hash: d962ea57899d64dc8a57040142c6b498a57c8064
commit-date: 2023-02-26
host: x86_64-apple-darwin
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

@rustbot label+I-suggestion-causes-error

c410-f3r:
Fix #10413

This is quite a rewrite that unfortunately took a  large amount of time. I tried my best to comment what is going on to easy review but feel free to ask any question.

The problem basically is that the current algorithm is only taking into consideration single blocks which means that things like the following don't work or show unpredictable results.

```rust
let mutex = Mutex::new(1);
{
  {
    let _ = *lock;
  }
}
```

The solve the issue, each path that refers a lock is now being tracked individually.

```
changelog: [`significant_drop_tightening`]: Lift the restriction of only considerate single blocks
```

### DIFF
--- a/clippy_lints/src/significant_drop_tightening.rs
+++ b/clippy_lints/src/significant_drop_tightening.rs
@@ -1,10 +1,10 @@
 use clippy_utils::{
     diagnostics::span_lint_and_then,
-    get_attr,
+    expr_or_init, get_attr, path_to_local,
     source::{indent_of, snippet},
 };
-use rustc_data_structures::fx::{FxHashMap, FxHashSet};
-use rustc_errors::{Applicability, Diagnostic};
+use rustc_data_structures::fx::{FxHashMap, FxHashSet, FxIndexMap};
+use rustc_errors::Applicability;
 use rustc_hir::{
     self as hir,
     intravisit::{walk_expr, Visitor},
@@ -13,6 +13,7 @@ use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::ty::{subst::GenericArgKind, Ty, TypeAndMut};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
 use rustc_span::{symbol::Ident, Span, DUMMY_SP};
+use std::borrow::Cow;
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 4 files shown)


c410-f3r:
@dswij I can re-roll if you don't have time to review

dswij:
Sorry for the delay, I had started to look into this but then fell a bit ill last week. I will take a look asap

clippy_lints/src/significant_drop_tightening.rs
```diff
@@ -350,73 +208,244 @@ impl<'cx, 'sdt, 'tcx> SigDropChecker<'cx, 'sdt, 'tcx> {
...
+            self.manage_has_expensive_expr_after_last_attr();
+        }
+        if let Some(expr) = block.expr {
+            self.ap.curr_stmt = Cow::Owned(dummy_stmt_expr(expr));
+            self.visit_expr(expr);
+            self.ap.curr_block_hir_id = block.hir_id;
+            self.ap.curr_block_span = block.span;
+            self.manage_has_expensive_expr_after_last_attr();
         }
+    }
```
dswij, line 282: I'm not sure if I understand completely, but I think this is the difference with the previous implementation? i.e. now we check for blocks and any nested block?

c410-f3r: The previous algorithm used to check all individual occurrences of `Block` through `LateLintPass::check_block` but now everything starts in the root block of `fn`through `LateLintPass::check_fn` to properly track the subsequent nested block children.

c410-f3r:
There is still related work to be done that I would like to continue pursuing with this PR merged, as such, I will try another reviewer.

r? clippy

Jarcho:
With a quick look this seems ok.

With that said, getting this lint to a state where it could be removed from the nursery will requiring rewriting this as a MIR lint. It needs to be able to track lifetimes for it to not have a large number of false positives. e.g. 
```rust
let mut locked = foo.lock();
let borrow = &mut *locked;
let references_borrow = bar(borrow);
// and so on
```

c410-f3r:
Yeah, such thing will allow me to close #9399 but something tells me that it will take a long time to finish :)

c410-f3r:
@Jarcho I can re-roll again if you don't have the time to review

Jarcho:
Sorry, I've been sick the past few weeks.

Since this is in the nursery I'll consider the lint still a WIP. I'll just reiterate that you can't fully fix this lint at the HIR level. You can reference `needless_clone` for how to write a MIR lint the needs to deal with lifetimes.

Since there doesn't seem to be any new issues added with this. @bors r+

c410-f3r:
The CI failure has been fixed.Thank you, @Jarcho!

Issue #10723, opened 2023-04-28
let_underscore_untyped false positive with derive(Pod) on struct


rkevingibson:
### Summary

When using the bytemuck::Pod trait (and corresponding derive macro), the lint points to a struct definition as in violation, despite having no `let` keyword in the source at all. It points to the name of the struct as the violation, and goes away if I remove the Pod trait from the list of derived traits.

### Lint Name

let_underscore_untyped

### Reproducer

I tried this code in a new blank project with the following cargo.toml dependency:
```
bytemuck = {version= "1.12.3", features= ["derive"]}
```

```rust
#![deny(clippy::all)]
#![deny(clippy::pedantic)]

use bytemuck::{Pod, Zeroable};

#[repr(C)]
#[derive(Debug, Clone, Copy, Default, Pod, Zeroable, PartialEq, PartialOrd)]
struct Foo {
    pub x: f32,
}

fn main() {
    println!("Hello, world!");
}
```

I saw this happen:

```
PS C:\Code\clippy_repro> cargo clippy
    Checking clippy_repro v0.1.0 (C:\Code\clippy_repro)
error: non-binding `let` without a type annotation
 --> src\main.rs:8:8
  |
8 | struct Foo {
  |        ^^^
  |
  = help: consider adding a type annotation or removing the `let` keyword
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_untyped
note: the lint level is defined here
 --> src\main.rs:2:9
  |
2 | #![deny(clippy::pedantic)]
  |         ^^^^^^^^^^^^^^^^
  = note: `#[deny(clippy::let_underscore_untyped)]` implied by `#[deny(clippy::pedantic)]`

error: could not compile `clippy_repro` due to previous error
```

I expected to see this happen:
There aren't any let keywords in the source, and as far as I can tell there's nothing in the bytemuck procedural macro that should trigger this lint.

### Version

```text
rustc 1.69.0 (84c898d65 2023-04-16)
binary: rustc
commit-hash: 84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc
commit-date: 2023-04-16
host: x86_64-pc-windows-msvc
release: 1.69.0
LLVM version: 15.0.7
```

PR #10775, opened 2023-05-14
Ignoring `let_underscore_untyped` warnings in code from proc macros

lochetti:
Don't linting let_underscore_untyped if code was generated by procedural macro.

This PR fixes https://github.com/rust-lang/rust-clippy/issues/10723

---

closes: https://github.com/rust-lang/rust-clippy/issues/10723

changelog: Fix: [`let_underscore_untyped`]: No longer lints inside proc macros
[#10775](https://github.com/rust-lang/rust-clippy/pull/10775)

### DIFF
--- a/clippy_lints/src/let_underscore.rs
+++ b/clippy_lints/src/let_underscore.rs
@@ -192,18 +192,18 @@ impl<'tcx> LateLintPass<'tcx> for LetUnderscore {
             if local.pat.default_binding_modes && local.ty.is_none() {
                 // When `default_binding_modes` is true, the `let` keyword is present.
 
-				if let Some(init) = local.init {
-                    // Ignore function calls that return impl traits...
-                    if matches!(init.kind, ExprKind::Call(_, _) | ExprKind::MethodCall(_, _, _, _)) {
-                        let expr_ty = cx.typeck_results().expr_ty(init);
-                        if expr_ty.is_impl_trait() {
-                            return;
-                        }
-                    }
-                    // Ignore if it is from a procedural macro...
-                    if is_from_proc_macro(cx, init) {
-                        return;
-                    }
+				// Ignore function calls that return impl traits...
+				if let Some(init) = local.init &&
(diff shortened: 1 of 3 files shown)


clippy_lints/src/let_underscore.rs
```diff
@@ -191,15 +192,19 @@ impl<'tcx> LateLintPass<'tcx> for LetUnderscore {
...
+                    if matches!(init.kind, ExprKind::Call(_, _) | ExprKind::MethodCall(_, _, _, _)) {
+                        let expr_ty = cx.typeck_results().expr_ty(init);
+                        if expr_ty.is_impl_trait() {
+                            return;
+                        }
+                    }
+                    // Ignore if it is from a procedural macro...
+                    if is_from_proc_macro(cx, init) {
+                        return;
+                    }
```
xFrednet, line 206: Tiny nit: This check is only executed if `local.init` is `Some`. However, the check should be done, whenever a `let_underscore_untyped` lint would be triggered. Could you move this below the if block, directly above the `span_lint_and_help`?

Note, in practice, I believe this doesn't matter, as I believe that locals without an init expression require a type. But moving it out of the block would be a bit cleaner either way.

xFrednet (commented):
Hey, thank you for the PR. I have a tiny nit, afterwards the PR will be ready to be merged :+1: 

---

Also, a small note, the commits are currently marked as *unverified*. We can merge them as is, but I want to mention it, in case that is unintentional :)

clippy_lints/src/let_underscore.rs
lochetti: Hello @xFrednet! Thanks for suggesting this change. If makes a lot of sense. I have done it in the https://github.com/rust-lang/rust-clippy/pull/10775/commits/2d9d81fc1e881f4ccf84c0509ffc23c8086cdcc6 commit.

And my git was configured to sign the commits using a key whoose identity was my work email and not my personal email. Now I've changed it and it will be Verified!

xFrednet:
Alright, this version looks good to me. Thank you for the PR and quick update :)

@bors r+

PR #10778, opened 2023-05-14
Don't emit clippy::useless_conversion on type aliases

### LINKED ISSUE #10773: Don't emit `clippy::useless_conversion` on type aliases
### Description

Currently, `clippy::useless_conversion` will still be emitted even on type aliases, for example:
```rust
HWND::from(0isize)
```
Clippy will suggest:
```rust
0isize
```
This conveys the type worse, but looking over rustc_hir, it doesn't seem whether it's a type alias is available so I don't know if this is possible to implement. I saw a few things relating to whether it's a trait alias but not whether it's a type alias. If this is possible, I'm happy to implement this :slightly_smiling_face:

### Version

```text
rustc 1.71.0-nightly (9ecda8de8 2023-04-30)
binary: rustc
commit-hash: 9ecda8de85ce893cc3fc748ab06be0b8250147a7
commit-date: 2023-04-30
release: 1.71.0-nightly
LLVM version: 16.0.2
```

### Additional Labels

@rustbot label +C-enhancement

Centri3:
Fixes #10773

changelog: Enhancement: [`useless_conversion`]: Don't lint on type aliases

### DIFF
--- a/clippy_lints/src/useless_conversion.rs
+++ b/clippy_lints/src/useless_conversion.rs
@@ -1,12 +1,12 @@
 use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_sugg};
+use clippy_utils::is_ty_alias;
 use clippy_utils::source::{snippet, snippet_with_context};
 use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::{is_copy, is_type_diagnostic_item, same_type_and_consts};
 use clippy_utils::{get_parent_expr, is_trait_method, match_def_path, path_to_local, paths};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
-use rustc_hir::def::{DefKind, Res};
-use rustc_hir::{BindingAnnotation, Expr, ExprKind, HirId, MatchSource, Node, PatKind, QPath, TyKind};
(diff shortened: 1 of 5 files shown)


clippy_lints/src/useless_conversion.rs
```diff
@@ -195,3 +197,12 @@ impl<'tcx> LateLintPass<'tcx> for UselessConversion {
         }
     }
 }
+
+/// `cx.qpath_res` seems to return `AssocFn` so we do this instead
+fn is_ty_alias(qpath: &QPath<'_>) -> bool {
```
llogiq, line 202: This could well be in `clippy_utils`, I think.

llogiq (approved):
Thank you! This looks very good. In fact, the only improvement I can suggest is to move the `is_ty_alias` function into `clippy_utils` so we can use it from other lints.

Are you OK with doing that? Otherwise I'll r+ and make a note to do it next time I refactor stuff.

clippy_lints/src/useless_conversion.rs
Centri3: done, though I did have to enable the feature `if_let_guards`.

llogiq:
Thank you!

@bors r+

PR #10779, opened 2023-05-15
Add new lint `ptr_cast_constness`

Centri3:
This adds a new lint which functions as the opposite side of the coin to `ptr_as_ptr`. Rather than linting only as casts that don't change constness, this lints only constness; suggesting to use `pointer::cast_const` or `pointer::cast_mut` instead.

changelog: new lint [`ptr_cast_constness`]

### DIFF
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -752,7 +752,7 @@ impl<'tcx> LateLintPass<'tcx> for Casts {
                 return;
             }
             cast_slice_from_raw_parts::check(cx, expr, cast_expr, cast_to, &self.msrv);
-            ptr_cast_constness::check(cx, expr, cast_expr, cast_from, cast_to);
+            ptr_cast_constness::check(cx, expr, cast_expr, cast_from, cast_to, &self.msrv);
             as_ptr_cast_mut::check(cx, expr, cast_expr, cast_to);
             fn_to_numeric_cast_any::check(cx, expr, cast_expr, cast_from, cast_to);
             fn_to_numeric_cast::check(cx, expr, cast_expr, cast_from, cast_to);
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -448,7 +448,7 @@ declare_clippy_lint! {
     #[clippy::version = "1.65.0"]
     pub PTR_CAST_CONSTNESS,
     pedantic,
-    "TODO"
+    "casting using `as` from and to raw pointers to change constness when specialized methods apply"
 }
 
 declare_clippy_lint! {
--- a/clippy_lints/src/casts/mod.rs
+++ b/clippy_lints/src/casts/mod.rs
@@ -18,6 +18,7 @@ mod fn_to_numeric_cast;
 mod fn_to_numeric_cast_any;
 mod fn_to_numeric_cast_with_truncation;
 mod ptr_as_ptr;
+mod ptr_cast_constness;
 mod unnecessary_cast;
 mod utils;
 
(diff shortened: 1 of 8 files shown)


clippy_lints/src/casts/ptr_cast_constness.rs
```diff
@@ -0,0 +1,44 @@
...
+    cast_from: Ty<'_>,
+    cast_to: Ty<'_>,
+    msrv: &Msrv,
+) {
+    if_chain! {
+        if msrv.meets(RustcVersion::new(1,65,0));
+        if let ty::RawPtr(TypeAndMut { mutbl: from_mutbl, .. }) = cast_from.kind();
+        if let ty::RawPtr(TypeAndMut { mutbl: to_mutbl, .. }) = cast_to.kind();
+        if !matches!((from_mutbl, to_mutbl),
+            (Mutability::Not, Mutability::Not) | (Mutability::Mut, Mutability::Mut));
```
llogiq, line 25:
```suggestion
        if matches!((from_mutbl, to_mutbl),
            (Mutability::Not, Mutability::Mut) | (Mutability::Mut, Mutability::Not));
```
We can make this a positive `matches!`.

llogiq (commented):
One small nitpick, otherwise this looks good. r=me with the nit fixed. Thank you!

clippy_lints/src/casts/ptr_cast_constness.rs
Centri3: fixed, do agree that negative `matches!` looks weird.

clippy_lints/src/casts/mod.rs
```diff
@@ -422,6 +423,34 @@ declare_clippy_lint! {
...
+    /// Use instead:
+    /// ```rust
+    /// let ptr: *const u32 = &42_u32;
+    /// let mut_ptr = ptr.cast_mut();
+    /// let ptr = mut_ptr.cast_const();
+    /// ```
+    #[clippy::version = "1.65.0"]
+    pub PTR_CAST_CONSTNESS,
+    pedantic,
+    "TODO"
```
llogiq, line 451: How about "Using raw pointer cast to change constness when specialized methods apply"?

tests/ui/ptr_cast_constness.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    // Make sure the lint is triggered inside a macro
+    let _ = inline!($ptr as *const i32);
+
+    // Do not lint inside macros from external crates
+    let _ = external!($ptr as *const i32);
+}
+
+#[clippy::msrv = "1.64"]
+fn _msrv_1_37() {
```
llogiq, line 39: Perhaps rename to `_msrv_1_64`? This looks like copy&pasted from another test.

tests/ui/ptr_cast_constness.rs
```diff
@@ -0,0 +1,55 @@
...
+    let ptr: *const u32 = &42_u32;
+    let mut_ptr: *mut u32 = &mut 42_u32;
+
+    // `pointer::cast_const` and `pointer::cast_mut` were stabilized in 1.65. Do not lint this
+    let _ = ptr as *mut i32;
+    let _ = mut_ptr as *const i32;
+}
+
+#[clippy::msrv = "1.65"]
+fn _msrv_1_38() {
```
llogiq, line 49: Similarly this should probably be named `_msrv_1_65`.

tests/ui/ptr_cast_constness.rs
Centri3: Fixed, was from `ptr_as_ptr`

llogiq:
Sorry for keeping you waiting. Thank you for writing the lint!

@bors r+

PR #10781, opened 2023-05-15
flip1995: Stepping down from the reviewer rotation

flip1995:
A step I was trying to avoid for way too long, but sadly necessary now. I hope I can come back stronger in a few months.

More explanation on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Stepping.20back.20from.20review.20rotation/near/358384096)

changelog: none


xFrednet:
While I'm a bit sad to merge this. I think it's good to take a break from reviews every once in a while. And based on the Zulip post, you'll still stay with us, so take all the time you need. Thank you for all the PRs you have reviewed over the years :heart: 

@bors r+

PR #10782, opened 2023-05-15
Add new lint `as_ptr_cast_underscore`

### LINKED ISSUE #10587: New Lint: `pointer_as_const_underscore`
### What it does

As far as I know, every usage of `.as_ptr() as *const _` is overcomplicated, as it has a pointer (`.as_ptr()`) being casted to a pointer (`as *const _`). So `as_ptr() as *const _` would be equal to `.as_ptr()` ([Example of them being equal](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=336dae9cd37e1d35e19331535c070786)).

Even when they have the same result, [~19.1k files use this pattern](https://github.com/search?q=lang%3ARust+%2F%5Cw*.as_ptr%5C%28%5C%29+as+%5C*const+_%2F&type=code).

---

## Notes

- Similar to #10567, this could cause some issues (as one targets `as *const _` and the other one `.as_ptr() as *const _`)
- I used the common `.as_ptr()` here as an example, but the lint would include any function returning a pointer.
- While `.as_ptr() as *const _` = `.as_ptr()`, it isn't equal to just `as *const _`.

### Lint Name

`pointer_as_const_underscore` or `as_ptr_as_const_underscore`

### Category

_No response_

### Advantage

Cleaner code

### Drawbacks

*None that I can think of.*

### Example

```rust
let a = "a";
let a_ptr = a.as_ptr() as *const _;
```

Could be written as:

```rust
let a = "a";
let a_ptr = a.as_ptr();
```

Centri3:
This will lint when the result of `as *<constness> _` is both the same type and mutability, as it is an unnecessary operation.

Fixes #10587

changelog: new lint [`as_ptr_cast_underscore`]


Alexendoo:
This feels like it should be rolled into `unnecessary_cast` rather than a separate lint

Centri3:
> This feels like it should be rolled into `unnecessary_cast` rather than a separate lint

Potentially, though `unnecessary_cast` currently checks only primitive types and not raw pointers, at all. Just stuff like `1u32 as u32`, not `x.as_ptr() as *const _`. This pattern is so common anyway that it feels like having this warn by default doesn't feel right since many use this to signal "I don't care about the resulting type, just make it work), and it currently doesn't give a suggested fix, but that should be easy to implement

asquared31415:
My lint #10567 checks for *all* uses of `as *const _` (or `*mut`) but it's currently a `restrict` lint for similar reasons that you found, the code pattern is so common, especially in `const`s and with integers cast to pointers.  It would entirely cover this lint by being more general.

Because this case is specifically for not changing the types, it is likely more suspicious, liable for the types to change due to type inference, and especially with pointers, that's probably not desirable, so I think that it should be something that's warn by default, but I understand that it may be very common.

I think that it would be useful for `unnecessary_cast` to be expanded to contain this lint, since it's a no-op cast and start handling `as _` and perhaps have a message when the target type is `_` that "the type was inferred to be X, but `_` is fragile and this may silently change"

There's some times when I think that we should consider moving the `restrict` lint for "use of `as` at all" to a warn by default, since `as` has so many of these suspicious behaviors. :laughing:

Centri3:
> I think that it would be useful for `unnecessary_cast` to be expanded to contain this lint

If this is the general consensus, then I'm open to doing this if an accurate suggestion could be added (which one definitely could), so running `cargo fix` would change it automatically. In that case warn-by-default would be fine.

> start handling as _ and perhaps have a message when the target type is _ that "the type was inferred to be X, but _ is fragile and this may silently change"

There's actually already `as_underscore` which does exactly this

asquared31415:
Sorry for not being clear, I specifically mean that `unnecessary_cast` should have a case for `as _` when `_` is inferred to be the same type as the left hand side of the `as` cast.  `unnecessary_cast` doesn't appear to fire on code like `let x: i32 = 0_i32 as _` or other cases where inference picks the same type for `_`, despite the cast being entirely redundant.

`as_underscore` would still be useful for cases when the `_` inferred type is *not* the same, and probably would need to be kept in `restrict`.  Unfortunately last I remember, `as_underscore` is far too common of a code pattern and feedback will be very negative from people denying clippy warnings in CI (even though we tell people not to do that).

Centri3:
I'll close this and instead add this to `unnecessary_cast` at some point.

PR #10783, opened 2023-05-15
Also inspect doc block comments

### LINKED ISSUE #10277: Clippy treats `/**…*/` doc comments differently from `///…` doc comments
### Summary

I had what seemed like a spurious `clippy::missing_errors_doc` that was resolved by changing `/**… */` style doc comment into a `///…` style doc comment. When I did that, additional clippy lints that were previously silent were revealed

### Reproducer

[playground link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=5e2c1e628428a6fae094123488ba5588) I would have expected the two functions to lint identically, but clippy's missing_errors_doc does not identify the `# Errors` section in one of them. This is not specific to `missing_errors_doc` but seems to apply to lint based on the content of the doc comment. Most of them are false negatives when using `/**…*/`, and as such are silent, but missing_errors_doc is a doc-comment-contents-based false-positive lint.

I would happily contribute a fix with guidance. On initial investigation, maybe [this](https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/doc.rs#L515) is the relevant line?

Thanks!

### Version

```text
rustc 1.69.0-nightly (c8e6a9e8b 2023-01-23)
binary: rustc
commit-hash: c8e6a9e8b6251bbc8276cb78cabe1998deecbed7
commit-date: 2023-01-23
host: x86_64-apple-darwin
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

blyxyas:
The system that checked for certain things in Documentation, wouldn't work with code blocks (with only one `*`). This PR fixes that.

More info in the linked issue.
Fixes #10277 

changelog: Fix lints that were silenced because of the use of `/** ...*/` instead of `///`

### DIFF
--- a/clippy_lints/src/doc.rs
+++ b/clippy_lints/src/doc.rs
@@ -454,6 +454,18 @@ fn lint_for_missing_headers(
 #[expect(clippy::cast_possible_truncation)]
 #[must_use]
 pub fn strip_doc_comment_decoration(doc: &str, comment_kind: CommentKind, span: Span) -> (String, Vec<(usize, Span)>) {
+    fn get_line_indentation(line: &Vec<char>) -> u32 {
+        let mut indent = 0;
+        for c in line {
+            if *c == ' ' {
+                indent += 1;
+            } else if *c == '\n' {
+                indent += 4;
+            }; // Standard "Tab" space
+        }
+        indent
+    }
+
     // one-line comments lose their prefix
     if comment_kind == CommentKind::Line {
         let mut doc = doc.to_owned();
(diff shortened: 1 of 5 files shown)


Alexendoo:
This is bypassing the markdown parsing which I'm not super keen on, without fixing [the root of the problem](https://github.com/rust-lang/rust-clippy/issues/10277#issuecomment-1412988784) we'd still have similar issues for any other doc based lint, e.g. for `missing_safety_doc`

blyxyas:
Sorry for the wait. I was in Portugal.
I'm going to fix this right now.

blyxyas:
Is there something more to change in this PR?

blyxyas:
I forgot to actually add the `#[warn]` attribute to test it :skull:

I'll add it soon

blyxyas:
Just tested it. It **does** indeed bug the spans on `doc_markdown`. I'll prepare a fix

blyxyas:
Fixed =)

Alexendoo:
It doesn't seem to work with

```rust
#![warn(clippy::doc_markdown)]

pub mod a {
    /**
    StructName
    */
    pub fn f() {}
}
```

blyxyas:
Almost 100 days now, I'll change the approach as this isn't working.

blyxyas:
I'll have to close this, sadly I don't have the time to research this issue further, I'll leave it free for someone else that has the time to fix it.

PR #10785, opened 2023-05-15
Fix `diverging_sub_expression` not checking body of block

### LINKED ISSUE #10776: diverging_sub_expression false negative on diverging body of block
### Summary

`diverging_sub_expression` triggers on `… || diverge()` but not `… || { diverge(); }`. It should trigger equivalently on both.

This came up due to https://github.com/rust-lang/rust/pull/104134 breaking this test in clippy:

https://github.com/rust-lang/rust-clippy/blob/a167973e8171bce22ded246283faaf80b729382d/tests/ui/diverging_sub_expression.stderr#L33-L37

due to the `panic!` expansion changing from `core::panicking::panic_fmt(…)` to `{ core::panicking::panic_fmt(…); }`.

### Lint Name

diverging_sub_expression

### Reproducer

**True positive:**

```rust
fn main() {
    let lhs = false;
    let _ = lhs || return;
}
```

```console
$ cargo clippy

warning: sub-expression diverges
 --> src/main.rs:3:20
  |
3 |     let _ = lhs || return;
  |                    ^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#diverging_sub_expression
  = note: `#[warn(clippy::diverging_sub_expression)]` on by default
```

---

**False negative:**

```rust
fn main() {
    let lhs = false;
    let _ = lhs || { return; };
}
```

```console
$ cargo clippy

... nothing
```

### Version

```text
rustc 1.71.0-nightly (2c41369ac 2023-05-13)
binary: rustc
commit-hash: 2c41369acc445d04129db40ba998dd7a89fb0d2e
commit-date: 2023-05-13
host: aarch64-apple-darwin
release: 1.71.0-nightly
LLVM version: 16.0.2
```

Centri3:
Fixes #10776

This also adds a warning to the test `ui/never_loop.rs`, not sure if this is correct or not.

changelog: [`diverging_sub_expression`]: Fix false negatives with body of block


clippy_lints/src/mixed_read_write_in_expression.rs
```diff
@@ -136,6 +137,15 @@ impl<'a, 'tcx> DivergenceVisitor<'a, 'tcx> {
 impl<'a, 'tcx> Visitor<'tcx> for DivergenceVisitor<'a, 'tcx> {
     fn visit_expr(&mut self, e: &'tcx Expr<'_>) {
         match e.kind {
+            // fix #10776
+            ExprKind::Block(block, ..) => {
+                if let Some(stmt) = block.stmts.first() && block.stmts.len() == 1 {
```
Jarcho, line 142: The final expression should also be handled. e.g. `{ return }`

Centri3: thanks! fixed, didn't notice that before since rustfmt would always add a semicolon.

i've decided to make it return after checking the final expression since the original issue was that `return;` and `{ return; }` would be treated differently, yet be the same operation, so something like `{ _ = 1; return }` feels a bit silly.

Jarcho:
Clippy has a no merge commit policy. You'll have to rebase your changes instead.

Centri3:
> Clippy has a no merge commit policy. You'll have to rebase your changes instead.

Could've sworn I did that the first time, not sure what I did. I guess it's clear now I have no idea what I'm doing :smile:

clippy_lints/src/mixed_read_write_in_expression.rs
```diff
@@ -136,6 +137,21 @@ impl<'a, 'tcx> DivergenceVisitor<'a, 'tcx> {
...
         match e.kind {
+            // fix #10776
+            ExprKind::Block(block, ..) => {
+                if let Some(e) = block.expr {
+                    self.visit_expr(e);
+
+                    return;
+                }
+
+                if let Some(stmt) = block.stmts.first() && block.stmts.len() == 1 {
```
Jarcho, line 148: Should be `if let [stmt] = block.stmts`. Everything else looks good.

Centri3: Good point, forgot you could do that

Centri3: Fixed

tests/ui/diverging_sub_expression.rs
```diff
@@ -35,6 +37,18 @@ fn foobar() {
...
             },
+            // lint blocks as well
+            15 => true || { return; },
+            16 => false || { return; },
+            // ... and when it's a single expression
+            17 => true || { return },
+            18 => false || { return },
+            // ... but not when there's both an expression and a statement
+            19 => true || { _ = 1; return },
+            20 => false || { _ = 1; return },
```
Jarcho, line 48: Just noticed this. The comment above says this shouldn't lint, but it currently does. One of the two should be fixed.

Jarcho:
Feel free to squash after fixing the last little nit.

clippy_lints/src/mixed_read_write_in_expression.rs
```diff
@@ -136,6 +137,21 @@ impl<'a, 'tcx> DivergenceVisitor<'a, 'tcx> {
...
         match e.kind {
+            // fix #10776
+            ExprKind::Block(block, ..) => {
+                if let Some(e) = block.expr && block.stmts.is_empty() {
+                    self.visit_expr(e);
+
+                    return;
+                }
+
+                if let [stmt, rest @ ..] = block.stmts && rest.is_empty() {
```
Jarcho, line 148: This ends up linting a block like `{ diverge(); 5 }` which it used to not do.

You can match on multiple fields at the same time which would make this clearer. e.g.
```rust
match (stmts, expr) {
    ([], Some(e)) => {}
    ([s], None) => {}
    _ => {}
}
```

Centri3: > This ends up linting a block like { diverge(); 5 } which it used to not do.

I'm not sure if this is an issue as it's unreachable anyway

I'll change it to that

Centri3: Done, also added a test for that (`{ diverge(); 5 }`) and it's fixed

Jarcho:
Thank you. @bors r+

PR #10786, opened 2023-05-15
Remove unnecessary `clone` from `needless_collect` example

mickvangelderen:
The example for [clippy::needless_collect](https://rust-lang.github.io/rust-clippy/master/#needless_collect) is written as follows:

```rust
let len = iterator.clone().collect::<Vec<_>>().len();
// should be
let len = iterator.count();
```

With this change, the unnecessary `clone()` is removed and the the standard

    ### Example
    ```rust
    // original
    ```
    Use instead:
    ```rust
    // improved
    ```

structure is followed.

Discussion: https://github.com/rust-lang/rust-clippy/discussions/10784#discussion-5198885

changelog: [`needless_collect`]: Cleaned up the example in the lint documentation.

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -3133,8 +3133,11 @@ declare_clippy_lint! {
     /// ### Example
     /// ```rust
     /// # let iterator = vec![1].into_iter();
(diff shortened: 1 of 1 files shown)


Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

PR #10788, opened 2023-05-15
New lint [`manual_partial_ord_and_ord_impl`]

### LINKED ISSUE #10744: Lint that encourages people to implement `PartialOrd` in terms of `Ord` when applicable
### What it does

When a type manually implements both `PartialOrd` and `Ord`, both implementations must agree i.e. behave the same way. Usually, the function bodies of `ord` and `partial_ord` are identical, save for wrapping the return expression in `Some()` for `partial_ord`. This is unnecessarily repetitive and makes the code more prone to error during refactoring, as one may forget to update one of the implementations. The best way of implementing these traits manually is to implement `PartialOrd` in terms of `Ord`, which reduces code duplication and guarantees that they function in the same way.

This lint should not trigger if the bounds on the impl for `Ord` are more restrictive than the bounds on `PartialOrd`.

Real-world example: https://github.com/bevyengine/bevy/pull/8529#discussion_r1183002034.

### Lint Name

duplicate_manual_partial_ord_impl

### Category

style, complexity

### Advantage

* The code is less repetitive.
* Implementations of `PartialOrd` and `Ord` are guaranteed to agree. This is safer for future refactoring, since you can't forget to update one of the two impls.

### Drawbacks

Will trigger on code that isn't broken which may be annoying, but following the lint's advice will always (as far as I can tell) make the code more robust.

### Example

```rust
impl Ord for F {
    fn cmp(&self, other: &Self) -> Ordering {
        match self.a.cmp(&other.b) {
            // .. some complex logic
        }
    }
}

impl PartialOrd for F {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(match self.a.cmp(&other.b) {
            // .. some complex logic
        })
    }
}
```

Could be written as:

```rust
impl Ord for F { /* Unchanged */ }

impl PartialOrd for F {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        Some(self.cmp(other))
    }
}
```

Centri3:
Lints when both `PartialOrd` and `Ord` are manually implemented yet the body of `PartialOrd::partial_cmp` isn't `Some(self.cmp(<other>))`.

This is in `correctness` currently but I'm ok with elsewhere.

Closes #10744

---

changelog: new lint [`needless_partial_ord_impl`]
[#10788](https://github.com/rust-lang/rust-clippy/pull/10788)


xFrednet:
@blyxyas Would you mind reviewing this PR as well? It looks like a nice new lint!

r? @xFrednet

blyxyas:
Yep, will also review it!

blyxyas:
I feel like `manual_partial_ord_and_ord_impl` is pretty long and doesn't really convey anything. Maybe something like `partial_ord_same_as_ord`?

tests/ui/manual_partial_ord_and_ord_impl.rs
blyxyas, line 1: What happens if the `PartialOrd` implementation has side effects? :thinking:
In that case, `PartialOrd` cannot be replaced with `Some(self.cmp(other))`

(this is a very edgy edge-case)

```rs
impl Ord for A {
    fn cmp(&self, other: &Self) -> Ordering {
        todo!();
    }
}

impl PartialOrd for A {
    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
        fn_with_side_effects();
	todo!();
    }
}
```

blyxyas:
Sorry for taking so long to review, it looks like a very useful lint, even tho not a lot of people manually implement `Ord` and `PartialOrd` ^^

tests/ui/manual_partial_ord_and_ord_impl.rs
Centri3: That would break it, yes. Though I'm not sure if this is too big of an issue since implementations of `PartialOrd` and `Ord` must agree; I'm not sure if this extends to printing to stdout or mutating a static and whatnot but generally, this would be a pretty terrible idea anyway.

There's also `derive_ord_xor_partial_ord`, which is similar and disallows the implementation even if it has side effects afaict.

Centri3:
lintcheck reveals 4 instances of `needless_partial_ord_impl` and 1 instance of `needless_clone_impl` in the wild: 2 in the log crate, 2 in regex, and the one for `needless_clone_impl` from ryu. They seem correct, but the suggestion most certainly isn't. If somebody can take a closer look at these that'd be much appreciated :D

blyxyas:
Could you split these two lints into two different PRs so they're easier to review and easier for generating the changelog? (and a simpler Git structure in general)

xFrednet:
> Could you split these two lints into two different PRs so they're easier to review and easier for generating the changelog? (and a simpler Git structure in general)

From the newly created PRs, it looks like the lints were quite different. So asking this is reasonable. For further reference, you can just add multiple `changelog` entries to a PR and I'll combine them:

changelog: new lint XYZ
changelog: new lint ZZZ

works just fine. Also, a small warning, that at one point you might encounter contributors, who are not as proficient with git as you are. And in this regard, just to add even more to this totally side tracked comment: I can recommend playing a bit with aliases for git if you use the console. I have a few for fast rebasing of just fixing up the last commit. :)

clippy_lints/src/needless_impls.rs
```diff
@@ -0,0 +1,145 @@
...
+                    |diag| {
+                        let (help, app) = if let Some(other) = body.params.get(0)
+                            && let PatKind::Binding(_, _, other_ident, ..) = other.pat.kind
+                        {
+                            (
+                                Cow::Owned(format!("{{ Some(self.cmp({})) }}", other_ident.name)),
+                                Applicability::Unspecified,
+                            )
+                        } else {
+                            (Cow::Borrowed("{ Some(self.cmp(...)) }"), Applicability::HasPlaceholders)
```
blyxyas, line 131: Could you add a test where this branch is taken?

clippy_lints/src/needless_impls.rs
```diff
@@ -0,0 +1,145 @@
...
+                && cmp_path.ident.name == sym::cmp
+                && let Res::Local(..) = path_res(cx, other_expr)
+            {} else {
+                span_lint_and_then(
+                    cx,
+                    NEEDLESS_PARTIAL_ORD_IMPL,
+                    item.span,
+                    "manual implementation of `PartialOrd` when `Ord` is already implemented",
+                    |diag| {
+                        let (help, app) = if let Some(other) = body.params.get(0)
```
blyxyas, line 123: (See the comment in the fixed version for more info.)
```suggestion
                        let (help, app) = if let Some(other) = body.params.get(1)
```

tests/ui/needless_partial_ord_impl.stderr
```diff
@@ -0,0 +1,28 @@
+error: manual implementation of `PartialOrd` when `Ord` is already implemented
+  --> $DIR/needless_partial_ord_impl.rs:18:1
+   |
+LL | /  impl PartialOrd for A {
+LL | |      fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+   | | _____________________________________________________________-
+LL | ||         todo!();
+LL | ||     }
+   | ||_____- help: change this to: `{ Some(self.cmp(self)) }`
```
blyxyas, line 9: Shouldn't this be `{ Some(self.cmp(other)) }`?

[`std::cmp::Ord`](https://doc.rust-lang.org/std/cmp/trait.Ord.html) is derived to use `std::cmp::Ord::cmp(&self, &other)`. So  I think it should actually use the second argument (there's a suggestion about this in the main lint file)

blyxyas (changes requested):
Some small changes and I think it's pretty ready to be merged :+1: (The `other` or `self` comment has to be addressed tho, I actually think it's a bug)

clippy_lints/src/needless_impls.rs
Centri3: There already is I think, but it was overshadowed by the `self` bug.

Centri3:
??? what

That makes no sense

blyxyas (approved):
I think this is complete. Thanks! :heart:

tests/ui/incorrect_partial_ord_impl_on_ord_type.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+impl PartialOrd for B {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+// lint, but we can't give a correct suggestion since &Self is not named
```
xFrednet, line 40: I think it would be valid to just add a second suggestion, giving `&Self` a name. Then we can also test the suggestion in the CI. Otherwise, I would suggest splitting the file into `_fixable.rs` and a `unfixable.rs` version :)

clippy_lints/src/incorrect_impls.rs
```diff
@@ -112,13 +161,76 @@ impl LateLintPass<'_> for IncorrectImpls {
...
+                        },
+                        [cmp_expr],
+                    ) = expr.kind
+                && is_res_lang_ctor(cx, cx.qpath_res(some_path, *some_hir_id), LangItem::OptionSome)
+                && let ExprKind::MethodCall(cmp_path, _, [other_expr], ..) = cmp_expr.kind
+                && cmp_path.ident.name == sym::cmp
+                && let Res::Local(..) = path_res(cx, other_expr)
+            {} else {
+                // If lhs and rhs are not the same type, bail. This makes creating a valid
+                // suggestion tons more complex.
```
xFrednet, line 203: NIT: You could mention this limitation in the lint description. I usually use the `### Limitations` section for something like this. But I don't think it's that important, so just an optional suggestion :)

xFrednet (commented):
Looks good overall, I have a few minor comments, and then it should be ready for bors consumption :D

tests/ui/incorrect_partial_ord_impl_on_ord_type.rs
```diff
@@ -0,0 +1,117 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        todo!();
+    }
+}
+
+impl PartialOrd for B {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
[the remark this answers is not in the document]
Centri3: I feel like handling all cases like that is a bit challenging, you could also do `Some(Self::cmp(self, other))` I'm pretty sure which *should* be correct but isn't caught

Centri3:
regex test failed, see #11111

xFrednet (approved):

xFrednet:
This version looks good to me. Could you rebase on master and also squash the changes into 1-4 commits? Then everything should be ready. Here you can also `r=blyxyas,xFrednet` afterwards :)

Thank you for the update and review you two!

---

Bors, you know what to do :D

PR #10789, opened 2023-05-16
fix example heading in `string_slice`

Centri3:
changelog: none


xFrednet (approved):

xFrednet:
Nice catch, thank you!

@bors r+

(Let's hope bors works again)

PR #10793, opened 2023-05-17
[`arithmetic_side_effects`] Fix #10792 

### LINKED ISSUE #10792: [arithmetic_side_effects] Constant fields are ignored
```rust
struct Foo {
  bar: u32,
}
const FOO: Foo = Foo { bar: 2 };
let _ = 10 / FOO.bar;
```

In fact, this is an issue that involves anything that uses `clippy_utils::consts::constant`.

c410-f3r:
Fix #10792 

```
changelog: [`arithmetic_side_effects`]: Retrieve field values of structures that are in constant environments
```

### DIFF
--- a/clippy_utils/src/consts.rs
+++ b/clippy_utils/src/consts.rs
@@ -14,7 +14,7 @@ use rustc_middle::mir::interpret::Scalar;
 use rustc_middle::ty::{self, EarlyBinder, FloatTy, ScalarInt, Ty, TyCtxt};
 use rustc_middle::ty::{List, SubstsRef};
 use rustc_middle::{bug, span_bug};
-use rustc_span::symbol::Symbol;
+use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::SyntaxContext;
 use std::cmp::Ordering::{self, Equal};
 use std::hash::{Hash, Hasher};
@@ -22,7 +22,8 @@ use std::iter;
 
 /// A `LitKind`-like enum to fold constant `Expr`s into.
 #[derive(Debug, Clone)]
-pub enum Constant {
+pub enum Constant<'tcx> {
+    Adt(rustc_middle::mir::ConstantKind<'tcx>),
     /// A `String` (e.g., "abc").
     Str(String),
     /// A binary string (e.g., `b"abc"`).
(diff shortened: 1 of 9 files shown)


clippy_utils/src/consts.rs
```diff
@@ -606,20 +602,18 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
...
-                ))),
-                ty::Float(FloatTy::F64) => Some(Constant::F64(f64::from_bits(
-                    int.try_into().expect("invalid f64 bit representation"),
-                ))),
-                ty::RawPtr(_) => Some(Constant::RawPtr(int.assert_bits(int.size()))),
-                // FIXME: implement other conversions.
-                _ => None,
-            }
+        mir::ConstantKind::Val(ConstValue::Scalar(Scalar::Int(int)), _) => match result.ty().kind() {
+            ty::Adt(_, _) => Some(Constant::Int(int.assert_bits(int.size()))),
```
c410-f3r, line 606: Initially I thought of bringing the field's ident from `ExprKind::Field` here in order to search the actual type with something like `def.all_fields().any(|field| ... match ident logic ...)` but then I realized that someone probably already did this in a clever way.

The problem is, I don't know any other way. Do you have any thoughts about this matter?

Manishearth:
r? @xFrednet 

(limited internet this week, redirecting reviews)

clippy_utils/src/consts.rs
xFrednet: Hey @rust-lang/clippy, has someone maybe worked mir and has an idea how to solve this? :)

c410-f3r: My search in the deepest and darkest parts of Rustc wasn't a success and since no one else replied, I moved forward with the initial idea. Hey, at least it is working 🙄

Jarcho: Would be better to add a variant to `Constant` to hold struct types and then resolve fields as they come up. The variant can just be a reference to an allocation.

c410-f3r: I am sorry @Jarcho, can you elaborate?

xFrednet: I'm guessing he recommends adding a new variant like `Constant::Field(bits-as-int)` and let the lint handle the type of the bits inside. Does that make sense? And @Jarcho was this what you meant?

Jarcho: Add something like `Constant::Adt(ConstantKind<'tcx>)`, have `Constant::fetch_path` return that when needed and then fields as needed in `Constant::expr`.

You need to be handle any nested fields (e.g. `a.b.c.d`) as well arrays and slices (can be done in a different pr).

c410-f3r: > Add something like `Constant::Adt(ConstantKind<'tcx>)`, have `Constant::fetch_path` return that when needed and then fields as needed in `Constant::expr`.

Done as requested

xFrednet (approved):

xFrednet:
This version looks good to me. Thank you for figuring this out :)

Lord bors, would you be so kind and merge this PR :)

PR #10807, opened 2023-05-21
[`unused_async`]: do not consider `await` in nested `async` blocks as used

### LINKED ISSUE #10800: `unused_async` not triggering when function only uses `.await` in an async block
### Summary

it seems the lint is simply checking for an occurrence of `.await` literal in the code, this is not always accurate, we can put an async block inside a sync function and use `.await` in the block, so making that function async is unnecessary but the lint still doesnt trigger

### Lint Name

unused_async

### Reproducer

playground link: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c31e923630b481e7c9d6c40a610d84a2

```rust
#![deny(clippy::unused_async)]
#![allow(dead_code, unused_must_use, clippy::no_effect)]

// works fine
async fn async_block_no_await() {
    async {
        tokio::time::sleep(std::time::Duration::from_secs(1));
    };
}

// false negative
async fn async_block_await() {
    async {
        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
    };
}

// proof that it can be sync
fn async_block_await_sync() {
    async {
        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
    };
}
```

### Version

```text
even on playground
```

y21:
Fixes #10800.
This PR makes sure that `await` expressions inside of inner `async` blocks don't prevent the lint from triggering.
For example
```rs
async fn foo() {
  async {
    std::future::ready(()).await;
  }
}
```
Even though there *is* a `.await` expression in this function, it's contained in an async block, which means that the enclosing function doesn't need to be `async` too.

changelog: [`unused_async`]: do not consider `await` in nested `async` blocks as used

### DIFF
--- a/clippy_lints/src/unused_async.rs
+++ b/clippy_lints/src/unused_async.rs
@@ -44,7 +44,7 @@ struct AsyncFnVisitor<'a, 'tcx> {
     found_await: bool,
     /// Also keep track of `await`s in nested async blocks so we can mention
     /// it in a note
-    found_await_in_async_block: bool,
+    await_in_async_block: Option<Span>,
     async_depth: usize,
 }
 
@@ -55,8 +55,8 @@ impl<'a, 'tcx> Visitor<'tcx> for AsyncFnVisitor<'a, 'tcx> {
         if let ExprKind::Yield(_, YieldSource::Await { .. }) = ex.kind {
             if self.async_depth == 1 {
                 self.found_await = true;
-            } else {
-                self.found_await_in_async_block = true;
+            } else if self.await_in_async_block.is_none() {
+                self.await_in_async_block = Some(ex.span);
             }
         }
         walk_expr(self, ex);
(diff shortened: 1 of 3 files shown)


clippy_lints/src/unused_async.rs
```diff
@@ -42,21 +42,43 @@ declare_lint_pass!(UnusedAsync => [UNUSED_ASYNC]);
 struct AsyncFnVisitor<'a, 'tcx> {
     cx: &'a LateContext<'tcx>,
     found_await: bool,
+    /// Also keep track of `await`s in nested async blocks so we can mention
+    /// it in a note
+    found_await_in_async_block: bool,
```
Jarcho, line 47: I'm not against doing this, but it should be keeping track of the span as well. Just the first occurrence is fine.

y21: Alright, done. Though not sure exactly if this is how you meant it.

Jarcho:
Thank you. @bors r+

PR #10809, opened 2023-05-22
Fix missing block for unsafe code

### LINKED ISSUE #10808: `single-match` makes incorrect suggestion for unsafe code
### Summary

This type of code creates incorrect suggestions for all lints like `single-match` and `single-match-else`:

```rust
match bar {
    Some(v) => unsafe { ... }
    ...
}
```

### Reproducer

Several non-passing tests for this issue have already been created in #10809

Run `cargo clippy -- -W clippy::single-match` with this code:

```rust
fn main() {
    foo(None);
}

pub fn foo(bar: Option<u32>) {
    match bar {
        Some(v) => unsafe {
            let r = &v as *const u32;
            println!("{}", *r);
        },
        _ => {}
    }
}
```

I expected Clippy to suggest this code:

```rust
    if let Some(v) = bar {
        unsafe {
            let r = &v as *const u32;
            println!("{}", *r);
        }
    }
```

Instead, it suggested this broken code:

```rust
    if let Some(v) = bar unsafe {
        let r = &v as *const u32;
        println!("{}", *r);
    }
```

### Version

```text
rustc 1.69.0 (84c898d65 2023-04-16)
binary: rustc
commit-hash: 84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc
commit-date: 2023-04-16
release: 1.69.0
LLVM version: 15.0.7
```

### Additional Labels

nyurik:
If a block is declared as unsafe, it needs an extra layer of curly braces around it.

Fixes #10808

This code adds handling for `UnsafeSource::UserProvided` block, i.e. `unsafe { ... }`. Note that we do not handle the `UnsafeSource::CompilerGenerated` as it seems to not be possible to generate that with the user code (?), or at least doesn't seem to be needed to be handled explicitly.

There is an issue with this code: it does not add an extra indentation for the unsafe blocks. I think this is a relatively minor concern for such an edge case, and should probably be done by a separate PR (fixing compile bug is more important than getting styling perfect especially when `rustfmt` will fix it anyway)

```rust
// original code
unsafe {
  ...
}

// code that is now generated by this PR
{ unsafe {
  ...
} }

// what we would ideally like to get
{
  unsafe {
    ...
  }
}
```

changelog: [`single_match`](https://rust-lang.github.io/rust-clippy/master/#single_match): Fix suggestion for `unsafe` blocks

### DIFF
--- a/tests/ui/single_match.rs
+++ b/tests/ui/single_match.rs
@@ -1,5 +1,5 @@
 #![warn(clippy::single_match)]
-#![allow(clippy::uninlined_format_args)]
+#![allow(unused, clippy::uninlined_format_args)]
 
 fn dummy() {}
 
@@ -243,4 +243,25 @@ fn main() {
         Some(x) => x,
         _ => 0,
     };
+}
+
+fn issue_10808(bar: Option<i32>) {
+    match bar {
+        Some(v) => unsafe {
+            let r = &v as *const i32;
+            println!("{}", *r);
+        },
+        _ => {},
+    }
+
+    match bar {
+        Some(v) => {
+            // this comment prevents rustfmt from collapsing the block
+            unsafe {
+                let r = &v as *const i32;
+                println!("{}", *r);
+            }
+        },
+        _ => {},
+    }
 }
--- a/tests/ui/single_match.rs
+++ b/tests/ui/single_match.rs
@@ -255,8 +255,8 @@ fn issue_10808(bar: Option<i32>) {
     }
 
     match bar {
+        #[rustfmt::skip]
         Some(v) => {
-            // this comment prevents rustfmt from collapsing the block
             unsafe {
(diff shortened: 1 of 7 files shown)


tests/ui/single_match.rs
```diff
@@ -244,3 +244,24 @@ fn main() {
...
+        Some(v) => unsafe {
+            let r = &v as *const i32;
+            println!("{}", *r);
+        },
+        _ => {},
+    }
+
+    match bar {
+        Some(v) => {
+            // this comment prevents rustfmt from collapsing the block
```
Jarcho, line 259: Using `#[rustfmt::skip]` would be better.

nyurik: Thanks @Jarcho, fixed

Jarcho:
Thank you. @bors r+

PR #10810, opened 2023-05-22
needless_else: new lint to check for empty `else` clauses

samueltardieu:
Empty `else` clauses are useless. They happen in the wild and are not linted yet: https://github.com/uutils/coreutils/pull/4880/files

`else` clauses containing or preceded by comments are not linted as the comments might be important.

changelog: [`needless_else`]: new lint

### DIFF
--- a/clippy_lints/src/needless_else.rs
+++ b/clippy_lints/src/needless_else.rs
@@ -1,0 +1,57 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, source::trim_span, span_extract_comment};
+use rustc_ast::ast::{Expr, ExprKind};
+use rustc_errors::Applicability;
+use rustc_lint::{EarlyContext, EarlyLintPass, LintContext};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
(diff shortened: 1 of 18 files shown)


sylvestre:
i wonder how often it is happening :)

clippy_lints/src/needless_else.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// Use instead:
+    /// ```rust
+    ///# fn check() -> bool { true }
+    /// if check() {
+    ///     println!("Check successful!");
+    /// }
+    /// ```
+    #[clippy::version = "1.71.0"]
+    pub NEEDLESS_ELSE,
+    style,
```
llogiq, line 31: I wonder how often that happens. Given that it's a style lint (which is warn by default), I worry that it might drown out more important messages if we leave it that way. On the other hand, if it's rare anyway, perhaps it's not a big deal. I'll run lintcheck and report back.

llogiq:
Ok, so lintcheck didn't find any instance. Instances will probably be mostly in half-refactored code. Thank you!

@bors r+

PR #10813, opened 2023-05-22
[`default_constructed_unit_structs`]: do not lint on type alias paths

### LINKED ISSUE #10755: FP: default_constructed_unit_structs
### Summary

.

### Lint Name

default_constructed_unit_structs

### Reproducer

I tried this code:

```rust
struct Sqlite {}

trait HasArguments<'q> {
    type Arguments;
}

impl<'q> HasArguments<'q> for Sqlite {
    type Arguments = std::marker::PhantomData<&'q ()>;
}

fn foo() {
    let _ = <Sqlite as HasArguments>::Arguments::default();
}

fn main() {}
```

I saw this happen:

```
warning: use of `default` to create a unit struct
  --> elided-lifetime-in-path-in-type-relative-expression.rs:12:48
   |
12 |     let _ = <Sqlite as HasArguments>::Arguments::default();
   |                                                ^^^^^^^^^^^ help: remove this call to `default`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#default_constructed_unit_structs
   = note: requested on the command line with `-W clippy::default-constructed-unit-structs`

warning: 2 warnings emitted
```

The suggestion does not compile
````
error[E0575]: expected method or associated constant, found associated type `HasArguments::Arguments`
  --> elided-lifetime-in-path-in-type-relative-expression.rs:12:13
   |
12 |     let _ = <Sqlite as HasArguments>::Arguments;
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: can't use a type alias as a constructor

````

### Version

```text
rustc 1.71.0-nightly (6d140d59b 2023-05-06)
binary: rustc
commit-hash: 6d140d59bb957903b02ddb993450509d8d92b448
commit-date: 2023-05-06
release: 1.71.0-nightly
LLVM version: 16.0.2
```

### Additional Labels

_No response_

y21:
Fixes #10755.

Type aliases cannot be used as a constructor, so this lint should not trigger in those cases.
I also changed `clippy_utils::is_ty_alias` to also consider associated types since [they kinda are type aliases too](https://github.com/rust-lang/rust/blob/48ec50ae39d0ca0baa0e78f56c395dcc6d7ebd65/compiler/rustc_resolve/src/late/diagnostics.rs#L1520).

changelog: [`default_constructed_unit_structs`]: do not lint on type alias paths

### DIFF
--- a/clippy_lints/src/default_constructed_unit_structs.rs
+++ b/clippy_lints/src/default_constructed_unit_structs.rs
@@ -1,4 +1,4 @@
-use clippy_utils::{diagnostics::span_lint_and_sugg, match_def_path, paths};
+use clippy_utils::{diagnostics::span_lint_and_sugg, is_ty_alias, match_def_path, paths};
 use hir::{def::Res, ExprKind};
 use rustc_errors::Applicability;
 use rustc_hir as hir;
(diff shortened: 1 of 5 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the addition! :+1: 

@bors r+

---

The lint is new and will be released with 1.71 if the version attribute is correct. It might be, that this changes doesn't make it into the beta branch. I'll therefore tag it as a potential back port. As this is a complexity lint, I could see this producing quite a few FPs

@rustbot label +beta-nominated

PR #10814, opened 2023-05-23
new lint: `explicit_into_iter_fn_arg`

### LINKED ISSUE #10743: Lint unnecessary `.into_iter()` in function arguments
### What it does

If someone didn't know that various functions can take an `impl IntoIterator` instead of an `impl Iterator` for their arguments, they might have wrote useless explicit `.into_iter()`s.

See also https://github.com/rust-lang/rust-clippy/issues/1094

### Lint Name

explicit_into_iter_fn_arg

### Category

pedantic

### Advantage

_No response_

### Drawbacks

_No response_

### Example

```rust
fn main() {
    let mut v = vec![];
    v.extend(vec![0].into_iter());
}
```

Could be written as:

```rust
fn main() {
    let mut v = vec![];
    v.extend(vec![0]);
}
```

y21:
Closes #10743.
This adds a lint that looks for `.into_iter()` calls in a call expression to a function that already expects an `IntoIterator`. In those cases, explicitly calling `.into_iter()` is unnecessary.
There were a few instances of this in clippy itself so I fixed those as well in this PR.

changelog: new lint [`explicit_into_iter_fn_arg`]


clippy_lints/src/explicit_into_iter_fn_arg.rs
```diff
@@ -0,0 +1,144 @@
...
+                    cx.typeck_results().type_dependent_def_id(parent_expr.hir_id)
+                        .map(|did| (did, args, MethodOrFunction::Method))
+                }
+                _ => None,
+            };
+
+            if let Some((parent_fn_did, args, kind)) = parent_expr
+                && let sig = cx.tcx.fn_sig(parent_fn_did).skip_binder().skip_binder()
+                && let Some(arg_pos) = args.iter().position(|x| x.hir_id == expr.hir_id)
+                && let Some(&into_iter_param) = sig.inputs().get(match kind {
```
llogiq, line 124: You could impl `Add<usize>` for `MethodOrFunction` to simplify this, but it won't block the PR from merging.

clippy_lints/src/explicit_into_iter_fn_arg.rs
```diff
@@ -0,0 +1,144 @@
...
+                    MethodOrFunction::Function => arg_pos,
+                    MethodOrFunction::Method => arg_pos + 1, // skip self arg
+                })
+                && let ty::Param(param) = into_iter_param.kind()
+                && let Some(span) = into_iter_bound(cx, parent_fn_did, param.index)
+            {
+                let sugg = snippet(cx, recv.span.source_callsite(), "<expr>").into_owned();
```
llogiq, line 131: How about using `snippet_with_applicability` here?

tests/ui/explicit_into_iter_fn_arg.rs
```diff
@@ -0,0 +1,24 @@
...
+fn e<T: IntoIterator<Item = i32>>(_: T) {}
+fn f(_: std::vec::IntoIter<i32>) {}
+
+fn main() {
+    a(vec![1, 2].into_iter());
+    b(vec![1, 2].into_iter());
+    c(vec![1, 2].into_iter());
+    d(vec![1, 2].into_iter());
+    e([&1, &2, &3].into_iter().cloned());
+    f(vec![1, 2].into_iter());
```
llogiq, line 23: I'd like to see an example with two chained `.into_iter()` calls.

llogiq (commented):
This looks good in general, I only have a scant few suggestions.

clippy_lints/src/explicit_into_iter_fn_arg.rs
y21: Hmm, I feel like an `Add` impl could potentially be a little confusing? I personally wouldn't know what it meant to add `MethodOrFunction` to a usize. Maybe it could just be a normal method? i.e.
```rs
impl MethodOrFunction {
    /// Maps the argument positon in `pos` to the parameter position.
    /// For methods, the `self` argument is skipped.
    fn arg_pos(&self, pos: usize) -> usize {
        match self {
            MethodOrFunction::Method => pos + 1,
            MethodOrFunction::Function => pos,
        }
    }
}
```

tests/ui/explicit_into_iter_fn_arg.rs
y21: Seems like that triggers [`useless_conversion`](https://rust-lang.github.io/rust-clippy/master/#useless_conversion) already:
```rs
warning: useless conversion to the same type: `std::vec::IntoIter<i32>`
  --> test.rs:24:14
   |
24 |     v.extend(vec![0].into_iter().into_iter());
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `.into_iter()`: `vec![0].into_iter()`
   |
```
So... maybe we shouldn't lint at all if we see `x.into_iter().into_iter()`, to avoid linting twice?

llogiq: Yeah. That makes me wonder if we should just merge this into the `useless_conversion` lint?

y21: Alright, makes sense, done. Sorry, took a bit longer

llogiq:
One small idea for improvement: If there are multiple `.into_iter()` calls, you could make the suggestion text ``"remove the `into_iter()`s"`` (note the *s*). But I'll have this merged now, so if you want to tackle that, feel free to open a followup PR and r? me. Thanks for seeing this through!

@bors r+

PR #10817, opened 2023-05-23
validate lint name in `clippy_dev`

y21:
This PR adds a little bit of validation to `cargo dev new_lint`. I've had it happen a few times where I wanted to add a new lint, but forgot that lint names cannot contain `-`. If you try to do it anyway, `clippy_dev` will generate illegal syntax (like adding `mod test-lint;` to clippy_lints/src/lib.rs for the module declaration). Maybe having it error out early would be helpful to others too.

changelog: none

### DIFF
--- a/clippy_dev/src/main.rs
+++ b/clippy_dev/src/main.rs
@@ -182,7 +182,7 @@ fn get_clap_config() -> ArgMatches {
                         .long("name")
                         .help("Name of the new lint in snake case, ex: fn_too_long")
                         .required(true)
-                        .value_parser(|name: &str| Ok::<_, Infallible>(name.replace("-", "_"))),
+                        .value_parser(|name: &str| Ok::<_, Infallible>(name.replace('-', "_"))),
                     Arg::new("category")
                         .short('c')
                         .long("category")
--- a/clippy_dev/src/main.rs
+++ b/clippy_dev/src/main.rs
@@ -180,7 +180,14 @@ fn get_clap_config() -> ArgMatches {
                         .short('n')
                         .long("name")
                         .help("Name of the new lint in snake case, ex: fn_too_long")
-                        .required(true),
+                        .required(true)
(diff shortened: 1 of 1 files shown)


clippy_dev/src/main.rs
```diff
@@ -180,7 +180,14 @@ fn get_clap_config() -> ArgMatches {
                         .short('n')
                         .long("name")
                         .help("Name of the new lint in snake case, ex: fn_too_long")
-                        .required(true),
+                        .required(true)
+                        .value_parser(|name: &str| {
+                            if name.contains('-') {
+                                Err("Lint name cannot contain `-`, use `_` instead.")
```
flip1995, line 186: Instead of error, would it make sense to just replace `-` with `_`?

y21: That makes sense to me (assuming that *was* directed at me). I think it's safe to assume that the user probably didn't want the literal character `-` but just meant to have it as a delimiter and replacing it with `_` is fine

flip1995 (approved):

clippy_dev/src/main.rs
flip1995: Yes, that question was for you :)

clippy_dev/src/main.rs
```diff
@@ -180,7 +181,8 @@ fn get_clap_config() -> ArgMatches {
                         .short('n')
                         .long("name")
                         .help("Name of the new lint in snake case, ex: fn_too_long")
-                        .required(true),
+                        .required(true)
+                        .value_parser(|name: &str| Ok::<_, Infallible>(name.replace("-", "_"))),
```
flip1995, line 185:
```suggestion
                        .value_parser(|name: &str| Ok::<_, Infallible>(name.replace('-', "_"))),
```

Clippy is not happy with itself.

flip1995:
@bors r+

Thanks!

PR #10821, opened 2023-05-24
Emit `unnecessary_cast` on raw pointers as well

Centri3:
Supersedes(?) #10782, since this and #10567 will cover the original issue.
Does not lint on type aliases or inferred types.

changelog: [`unnecessary_cast`]: Also emit on casts between raw pointers with the same type and constness

### DIFF
--- a/tests/ui/unnecessary_cast.stderr
+++ b/tests/ui/unnecessary_cast.stderr
@@ -1,5 +1,5 @@
 error: casting integer literal to `i32` is unnecessary
-  --> $DIR/unnecessary_cast.rs:14:5
+  --> $DIR/unnecessary_cast.rs:17:5
    |
 LL |     1i32 as i32;
    |     ^^^^^^^^^^^ help: try: `1_i32`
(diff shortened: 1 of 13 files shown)


asquared31415:
I think it would be useful to have tests for getting a pointer from somewhere else, example:
```rust
let ptr: *const () = 1 as *const ();
let x = ptr as *const ();
```
Does this lint handle generics properly?  The following code should probably lint:
```rust
fn owo<T>(ptr: *const T) {
    let x = ptr as *const T;
}
```
A test case casting to a *different* generic `*const U` in that same style should also be added.

A test case involving a cast like `x: *const u32` `x as *const u32 as *const u8` might also be useful to make sure that the suggestion works properly and only removes the first useless cast.

Also why does this explicitly allow code that uses `as _` when that cast is a no-op?  I would expect both of these test cases to emit a lint, because the code functions the same if the `as _` is removed.
```rust
    let _: *const u8 = [1u8, 2].as_ptr() as _;
    let _: *mut u8 = [1u8, 2].as_mut_ptr() as _;
```

Centri3:
> I think it would be useful to have tests for getting a pointer from somewhere else, example:

I'm pretty sure this would function the same as any other cast but I added one anyway

> Does this lint handle generics properly? The following code should probably lint:

Yes, since it checks `cast_from` and `cast_to`'s kind which afaik will ensure that type, constness and generics will all be the same. I'm not sure if this has any unintended side effects, though

asquared31415:
Thanks for adding those tests! This all looks good to me.

A tiny wording nitpick: I think that rustc uses "mutability" to refer to what you called "constness" (and "const" is normally things like `const fn`), so would a better wording be "casting a raw pointer to the same mutability and type is unnecessary"?

Centri3:
The std library's docs of `pointer::cast_mut` uses constness, so I'm not sure about that. Plus there's now `ptr_cast_constness`, so, they're kind of interchangeable here imo

asquared31415:
Oh look at that, it does.  I guess in the context of raw pointers specifically, that wording is used instead?  Weird.  Then I have absolutely no issue with that in the name of consistency.

tests/ui/unnecessary_cast.stderr
```diff
@@ -1,190 +1,232 @@
+error: casting raw pointers to the same type and constness is unnecessary (`*const T` -> `*const T`)
```
dswij, line 1: Wouldn't the type inside the bracket always be equal? i.e.

(`foo` -> `bar`) always will be the case that `foo==bar` in this lint?

dswij: What do you think of removing this altogether? Seems redundant

dswij (commented):
Thanks for this! This LGTM, just a small comment on the suggestion message.

tests/ui/unnecessary_cast.stderr
Centri3: That was done to remain consistent with the other lint message further down (`&format!("casting to the same type is unnecessary ({cast_from} -> {cast_to})")`). It is redundant yes but imo both should be changed if it's removed

dswij: That's fair. We can leave it for now

dswij:
Thank you! @bors r+

PR #10822, opened 2023-05-24
Ignore `#[cfg]`'d out code in `needless_else`

Alexendoo:
changelog: none (same release as #10810)

`#[cfg]` making things fun once more

This lead me to think about macro calls that expand to nothing as well, but apparently they produce an empty stmt in the AST so are already handled, added a test for that

r? @llogiq


llogiq:
Thank you!

@bors r+

PR #10829, opened 2023-05-26
[bool_comparison] Deal with multiple false-like options

### LINKED ISSUE #10796: New lint that forbids `!bool`
### What it does

`!bool` is hard to see and can lead to possible logical misunderstands. Instead, `expr == false` should be preferable.

This is not just my opinion based on experience, other ppl also think the same way. To list a few:

* https://github.com/hashcat/hashcat#contributing
* https://eev.ee/blog/2016/12/01/lets-stop-copying-c/#_1

### Lint Name

neg_bool

### Category

restriction

### Example

```rust
let expr: bool = ...;
if !expr { ... }
```

Could be written as:

```rust
let expr: bool = ...;
if expr == false { ... }
```

c410-f3r:
Fix #10796

In face of the feedback provided in https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/.E2.9C.94.20New.20possible.20lint.20to.20deal.20with.20.60!bool.60 and in #10796, the most natural solution to maintain the current behaviour, avoid potential conflicts with a new theoretical lint and enable the possibility of specify different behaviours was, IMO, the addition of a new configuration parameter for `bool_comparison` as well as the incorporation of unary checks.

```
changelog: [`bool_comparison`]: Add a new `bool-comparison-false-ty` parameter to specify different false-like behaviours
```

### DIFF
--- a/tests/ui-toml/bool_comparison_equal/bool_comparison_equal.stderr
+++ b/tests/ui-toml/bool_comparison_equal/bool_comparison_equal.stderr
@@ -1,0 +1,52 @@
+error: inequality checks against true can be replaced with a comparison against `false`
+  --> $DIR/bool_comparison_equal.rs:7:13
+   |
+LL |     let _ = x != true;
+   |             ^^^^^^^^^ help: try simplifying it as shown: `x == false`
+   |
+   = note: `-D clippy::bool-comparison` implied by `-D warnings`
+
+error: inequality checks against true can be replaced with a comparison against `false`
+  --> $DIR/bool_comparison_equal.rs:8:13
+   |
+LL |     let _ = true != x;
+   |             ^^^^^^^^^ help: try simplifying it as shown: `x == false`
+
+error: less than comparison against true can be replaced with a comparison against `false`
+  --> $DIR/bool_comparison_equal.rs:9:13
+   |
+LL |     let _ = x < true;
+   |             ^^^^^^^^ help: try simplifying it as shown: `x == false`
+
+error: less than comparison against true can be replaced with a comparison against `false`
+  --> $DIR/bool_comparison_equal.rs:10:13
+   |
+LL |     let _ = true > x;
+   |             ^^^^^^^^ help: try simplifying it as shown: `x == false`
+
+error: order comparisons between booleans can be simplified
+  --> $DIR/bool_comparison_equal.rs:13:13
+   |
+LL |     let _ = x < y;
+   |             ^^^^^ help: try simplifying it as shown: `x == false & y`
+
+error: order comparisons between booleans can be simplified
+  --> $DIR/bool_comparison_equal.rs:14:13
+   |
+LL |     let _ = x > y;
+   |             ^^^^^ help: try simplifying it as shown: `x & y == false`
+
+error: order comparisons between booleans can be simplified
+  --> $DIR/bool_comparison_equal.rs:15:13
+   |
+LL |     let _ = x > y && !x && y == false;
+   |             ^^^^^ help: try simplifying it as shown: `x & y == false`
+
+error: negation (`!`) is hard to see and can lead to possible misunderstandings
+  --> $DIR/bool_comparison_equal.rs:15:22
+   |
+LL |     let _ = x > y && !x && y == false;
+   |                      ^^ help: try replacing it with a comparison against `false`: `x == false`
+
+error: aborting due to 8 previous errors
+
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -595,7 +595,9 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|_| Box::new(float_literal::FloatLiteral));
     store.register_late_pass(|_| Box::new(ptr::Ptr));
     store.register_late_pass(|_| Box::new(needless_bool::NeedlessBool));
-    store.register_late_pass(|_| Box::new(needless_bool::BoolComparison));
+    let bool_comparison_false_ty = conf.bool_comparison_false_ty.clone();
+    store
+        .register_late_pass(move |_| Box::new(needless_bool::BoolComparison::new(bool_comparison_false_ty.as_deref())));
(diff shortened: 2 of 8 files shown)


tests/ui-toml/bool_comparison_equal/bool_comparison_equal.stderr
```diff
@@ -0,0 +1,52 @@
+error: inequality checks against true can be replaced with a comparison against `false`
+  --> $DIR/bool_comparison_equal.rs:7:13
+   |
+LL |     let _ = x != true;
+   |             ^^^^^^^^^ help: try simplifying it as shown: `x == false`
```
xFrednet, line 5: Looking at the lint description, the lint was first desired to catch cases like `x == false` and suggest `!x` instead. I seems weird, that binary operators all of the sudden suggest such code. I think, suggesting `!x` directly would be the better option.

tests/ui-toml/bool_comparison_equal/bool_comparison_equal.stderr
```diff
@@ -0,0 +1,52 @@
...
+  --> $DIR/bool_comparison_equal.rs:10:13
+   |
+LL |     let _ = true > x;
+   |             ^^^^^^^^ help: try simplifying it as shown: `x == false`
+
+error: order comparisons between booleans can be simplified
+  --> $DIR/bool_comparison_equal.rs:13:13
+   |
+LL |     let _ = x < y;
+   |             ^^^^^ help: try simplifying it as shown: `x == false & y`
```
xFrednet, line 31: The `&` operator is semanticly equivalent, so i think it's good to have it as the main suggestion. Especially if the suggestion might become machine applicable at some point. However, it might also be good to include a help comment, suggesting the usage of `&&`. :)

tests/ui-toml/bool_comparison_equal/bool_comparison_equal.stderr
```diff
@@ -0,0 +1,52 @@
...
+  --> $DIR/bool_comparison_equal.rs:15:13
+   |
+LL |     let _ = x > y && !x && y == false;
+   |             ^^^^^ help: try simplifying it as shown: `x & y == false`
+
+error: negation (`!`) is hard to see and can lead to possible misunderstandings
+  --> $DIR/bool_comparison_equal.rs:15:22
+   |
+LL |     let _ = x > y && !x && y == false;
+   |                      ^^ help: try replacing it with a comparison against `false`: `x == false`
```
xFrednet, line 49: It looks like the lint is detecting and suggesting the complete opposite of what what it should do, according to the lint description:
* https://rust-lang.github.io/rust-clippy/master/index.html#bool_comparison

Could you have a look at this?

xFrednet (commented):
Hey, I only skimmed over the test output, and found some inconsistencies with what the lint is intended to do. Could you take a look at the marked cases?

I'll probably wait on a code review until they have been addressed. You're welcome to reach out, if you need assistance with anything :)

c410-f3r:
Based on the provided feedback, widening the scope of `bool_comparison` with flags and unary checks looked like the most sensible approach but I am open to new suggestions/approaches. If the issue is documentation, new notes will de added as needed.

The context is that a new lint (let's call it `bool_false`) demanding `bool == false` or any other similar expression like `bool.not()` will conflict with `bool_comparison` and users will have to choose one of them.

At the same time, choosing `bool_false` over `bool_comparison` make diagnostics like `x != true` disappear. This could in theory be resolved by copying the behaviour of `bool_comparison` to `neg_bool` with the modified `!bool == false` but then, it is much more natural and less redundant to just add configuration options to `bool_comparison` and let the current `!bool` be default.

tests/ui-toml/bool_comparison_equal/bool_comparison_equal.stderr
c410-f3r: The only thing that differs this diagnostic from the current status-quo is `x == false` instead of `!x`.

https://github.com/rust-lang/rust-clippy/blob/dc17e7317b7bd6d585abafa4d927dcf876dc4d18/tests/ui/bool_comparison.stderr#L75-L79

So if possible I prefer to improve messages in a following PR.

xFrednet: > The only thing that differs this diagnostic from the current status-quo is `x == false` instead of `!x`.

Exactly, I think it would be better to suggest `!x` directly, as it would otherwise be linted in the next clippy run

> So if possible I prefer to improve messages in a following PR.

Sure that works for me :)

xFrednet:
> Based on the provided feedback, widening the scope of `bool_comparison` with flags and unary checks looked like the most sensible approach

I'm not sure, which flags and unary checks you're referring to here. Could you explain this a bit?

> The context is that a new lint (let's call it `bool_false`) demanding `bool == false` or any other similar expression like `bool.not()` will conflict with `bool_comparison` and users will have to choose one of them.

I don't think we should restrict this lint based on a lint, which might or might not be added in the future. Unless you're already planning on adding these lints. Even then, I'd say that this lint should have a higher priority, as it's warn by default and `bool_false` would be a `restriction` lint.

This is not to imply that we can't make compromises. It would be easy to add a check in this lint, to see if the `bool_false` lint is enabled, and if so, ignore `x == false` but continue linting `x < y` cases. That seems like the better solution to me.

c410-f3r:
> I'm not sure, which flags and unary checks you're referring to here. Could you explain this a bit?

Flags or configuration options: https://github.com/rust-lang/rust-clippy/pull/10829/files#diff-92569a8dda65b485c7d2e10bf6a341e6d415c670fd1856590ba5bde6d894610b

Additional unary checks: https://github.com/rust-lang/rust-clippy/pull/10829/files#diff-ed7759c6e78c5d300dc98807b3c2eff9d2467f6c6f1bb10ba56e06f62ff37ffbR361-R383

As previously said, this PR was a best-effort trying to avoid conflict with existent behaviour but it seems that the current design is not desired. If it is good to you, I can create a separate lint and just let users choose which one they want.

c410-f3r:
Thank you @xFrednet for your time and feedback on the subject. Unfortunately, I don't have the energy or motivation to keep persuading  this feature.

Issue #10803, opened 2023-05-20
redundant_pattern_match does not trigger on `matches!` macro


disco07:
### Summary

Uses of `matches` macro with `Result` or `Option` in first argument and in the second argument is just here to check if the first argument is `is_ok()` or `is_some()`, do not lint. When I make `cargo expand` on this:
```rust
matches!(Ok::<i32, i32>(10), Ok(_));
```
I have:
```rust
match Ok::<i32, i32>(10) {
        Ok(_) => true,
        _ => false,
    };
```

### Lint Name

redundant_pattern_match

### Reproducer

I tried this code:

```rust
matches!(Ok::<i32, i32>(10), Ok(_));
```

I expected to see this happen:
```rust
Ok::<i32, i32>(10).is_ok()
```
Instead, this happened:
```rust
matches!(Ok::<i32, i32>(10), Ok(_));
```

### Version

```text
rustc 1.71.0-nightly (521f4dae1 2023-05-19)
binary: rustc
commit-hash: 521f4dae1bdf7fe9cf9436ecef9ee7c7442708bf
commit-date: 2023-05-19
host: x86_64-unknown-linux-gnu
release: 1.71.0-nightly
LLVM version: 16.0.4
```

disco07:
@flip1995 I want to work on this, have you some ideas (suggestions) to give me please to resolve this?

flip1995:
This is a bit complicated. The macro expands to a `match` expression. The `redundant_pattern_match` lint i probably has a `span.from_expansion` condition in it, that prevents it from triggering inside of (external) macros. We could loosen that restriction, by checking the `expn_data` if the expansion (directly) comes from the `matches!` macro. I think we have a utility function called `is_expn_of` of something similar for this.

I haven't looked into the code of this lint for quite some time. Let me know, if this explanation already helps you get started, or if you want some more detailed explanation. I'm happy to do that, if necessary :)

disco07:
@flip1995 Thanks I will explore this path

PR #10831, opened 2023-05-26
Fix redundant_pattern_match on matches! macro

disco07:
This PR solve this issue https://github.com/rust-lang/rust-clippy/issues/10803
r? @flip1995 

changelog: none

### DIFF
--- a/clippy_lints/src/matches/mod.rs
+++ b/clippy_lints/src/matches/mod.rs
@@ -25,7 +25,7 @@ mod wild_in_or_pats;
 
 use clippy_utils::msrvs::{self, Msrv};
 use clippy_utils::source::{snippet_opt, walk_span_to_context};
-use clippy_utils::{higher, in_constant, is_span_match, tokenize_with_text};
+use clippy_utils::{higher, in_constant, is_direct_expn_of, is_span_match, tokenize_with_text};
 use rustc_hir::{Arm, Expr, ExprKind, Local, MatchSource, Pat};
 use rustc_lexer::TokenKind;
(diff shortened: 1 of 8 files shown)


clippy_lints/src/matches/mod.rs
```diff
@@ -974,12 +974,16 @@ impl_lint_pass!(Matches => [
 
 impl<'tcx> LateLintPass<'tcx> for Matches {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
-        if in_external_macro(cx.sess(), expr.span) {
+        if is_expn_of(expr.span, "matches").is_none() && in_external_macro(cx.sess(), expr.span) {
```
flip1995, line 977: I'm not sure if this should be `is_direct_expn_of` (also below) Otherwise it might also lint, if `matches!` is used in an external macro.

flip1995 (commented):
That was quick! One small comment.

clippy_lints/src/matches/mod.rs
disco07: Thanks for your response, I'll try this and push the change

flip1995 (approved):

flip1995:
Thanks! Can you squash your commits together? After that this can already be merged.

disco07:
> Thanks! Can you squash your commits together? After that this can already be merged.

I think it's good now ;)

flip1995:
I rebased the branch on top of master and squashed the commits. (We don't want to have merge commits in the history).

@bors r+

PR #10834, opened 2023-05-26
Use URL parameters for filter states

### LINKED ISSUE #8510: "Clippy Lints" page forgets "Lint groups" selection
### Description

If you search for a Clippy lint by selecting one or more "Lint groups," click "View Source," and then click the back button, the lint groups will still be selected, but the selection will not be respected.

What I did:
1. Go to the [Clippy Lints](https://rust-lang.github.io/rust-clippy/master/) page.
2. Unselect all "Lint groups" but "Restriction."
3. Expand "indexing_slicing."
4. Click "View Source."
5. Click the browser back button.

What I saw: more than just "Restriction" lints.

What I expected to see: just "Restriction" lints.

In fact, after clicking the back button, if you select and unselect a lint group, you can see those lints disappearing from the listing.

Clicking "Related Issues" seems to exhibit the same behavior.

### Version

_No response_

### Additional Labels

_No response_

whee:
This fixes #8510 by storing Clippy Lints page filter configuration in the URL parameters.

This includes:
- Lint levels
- Lint groups
- Version filters

"Filter" was already present in the URL and its behavior is retained. There is existing support for passing a `sel` query parameter; this is also retained, but I am not sure if it used in the wild.

The URL parameters only get included if they are modified after loading the page.

I have these changes available here in case people want to play with it: https://whee.github.io/rust-clippy/master/

An example with levels, groups, and versions set (oddly):

https://whee.github.io/rust-clippy/master/#/?groups=pedantic,perf&levels=allow,warn&versions=gte:53,lte:57,eq:54

Adding a filter:

https://whee.github.io/rust-clippy/master/#/manual_str_repeat?groups=pedantic,perf&levels=allow,warn&versions=gte:53,lte:57,eq:54

---

changelog: Docs: [`Clippy's lint list`] now stores filter parameters in the URL, to allow easy sharing
[#10834](https://github.com/rust-lang/rust-clippy/pull/10834)

### DIFF
--- a/util/gh-pages/index.html
+++ b/util/gh-pages/index.html
@@ -517,7 +517,8 @@ Otherwise, have a great day =^.^=
                     <h2 class="panel-title">
                         <div class="panel-title-name">
                             <span>{{lint.id}}</span>
-                            <a href="#{{lint.id}}" class="anchor label label-default" ng-click="open[lint.id] = true; $event.stopPropagation()">&para;</a>
+                            <a href="#{{lint.id}}" class="anchor label label-default"
+                                ng-click="openLint(lint); $event.preventDefault(); $event.stopPropagation()">&para;</a>
                             <a href="" id="clipboard-{{lint.id}}" class="anchor label label-default" ng-click="copyToClipboard(lint); $event.stopPropagation()">
                                 &#128203;
                             </a>
--- a/util/gh-pages/index.html
+++ b/util/gh-pages/index.html
@@ -501,9 +501,11 @@ Otherwise, have a great day =^.^=
                     <div class="col-12 col-md-7 search-control">
                         <div class="input-group">
                             <label class="input-group-addon" id="filter-label" for="search-input">Filter:</label>
-                            <input type="text" class="form-control filter-input" placeholder="Keywords or search string" id="search-input" ng-model="search" ng-model-options="{debounce: 50}"/>
+                            <input type="text" class="form-control filter-input" placeholder="Keywords or search string" id="search-input"
+                                ng-model="search" ng-blur="updatePath()" ng-keyup="$event.keyCode == 13 && updatePath()"
+                                ng-model-options="{debounce: 50}" />
                             <span class="input-group-btn">
-                                <button class="filter-clear btn" type="button" ng-click="search = ''">
+                                <button class="filter-clear btn" type="button" ng-click="search = ''; updatePath();">
                                     Clear
                                 </button>
                             </span>
--- a/util/gh-pages/script.js
+++ b/util/gh-pages/script.js
@@ -156,10 +156,6 @@
                 Object.entries(versionFilterKeyMap).map(([key, value]) => [value, key])
             );
 
-            // An internal URL change occurs when we are modifying the URL parameters in a way
-            // that should not reload parameters from the URL
-            let internalURLChange = false;
-
             // loadFromURLParameters retrieves filter settings from the URL parameters and assigns them
             // to corresponding $scope variables.
             function loadFromURLParameters() {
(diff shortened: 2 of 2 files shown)


xFrednet:
Hey @whee, thank you for the PR, I'm excited to see this idea becoming a reality!!

I haven't looked at the code yet, but played with the linked version. (Thank you for hosting it :D). It already works wonderfully. I have a few suggestions which would make it feel better for me:
1. Currently, every typed letter in the search box updates the domain. I think it would be better, to add a debounce or something similar, so the domain is only updated if there are no changes for a second or so. The search already uses this. Another option, which would probably be easier and work as expected, would be to only update the URL if the search box is unfocussed. There should be a listener for that in JS
2. Selecting the default filters, keeps the levels and groups in the URL. This creates a long URL which is equivalent to not having any filter:
    * https://whee.github.io/rust-clippy/master
    * https://whee.github.io/rust-clippy/master/#/?levels=allow,warn,deny,none&groups=cargo,complexity,correctness,nursery,pedantic,perf,restriction,style,suspicious
    
    It would be nice, if the filters are dropped again, when they're equivalent to the default
3. Clicking the ¶ button next to the link adds the lint name after a `/#/<lint>` and not a `#<lint>` like it used to.

As said, I haven't looked at the code yet, so if any of these changes are intentionally or explained there, I'll see it later. Besides these functionally nits, I do love this functionality!

---

Next week I have exams coming up, so it'll most likely take until the end of the week, until I can review the code. You are welcome to push any changes, if you agree with my suggestions above :)

whee:
I appreciate the feedback @xFrednet! I'll try to address each of your points:

1. I agree that the path updating is a bit noisy in the search case. I think updating when unfocused makes sense. I know I tend to go from editing to copying the URL pretty quickly via shortcuts, so a debounce might add too much of a delay.

2. This is partially intentional, but I think your use case makes sense as a "reset to not including the URL parameters" situation.

The scenario I wanted to address was:

- I've configured a set of filters I like, which coincidentally match the defaults in some way
- I bookmark the page
- I come back later, and the view is different because the defaults changed

Clicking "Default" explicitly is a good indicator that the user wants defaults and not a specific set, though, when compared to clicking the checkboxes and coincidentally matching the defaults.

3. I think this is a side effect of using `$location` service to consistently set the URL. Adding that slash is "standard" for AngularJS in this case. I'll check into whether or not I can change that behavior.

I'll try to get these changes in soon. Good luck with exams!

xFrednet:
Hey, I finally have time to look at this properly, sorry that it took so long. I should be done with my semester now :tada: 

> 2. This is partially intentional, but I think your use case makes sense as a "reset to not including the URL parameters" situation.
>
> The scenario I wanted to address was:
>
> - I've configured a set of filters I like, which coincidentally match the defaults in some way
> - I bookmark the page
> - I come back later, and the view is different because the defaults changed
> 
> Clicking "Default" explicitly is a good indicator that the user wants defaults and not a specific set, though, when compared to clicking the checkboxes and coincidentally matching the defaults.

I don't think many people will bookmark specific filters. And even then, I think it's unlikely that the filters will change much. We had two changes to the default in the last two years and one of them was the addition of a new lint group. I see arguments for both sides here :)

whee:
Thanks for taking a look!

> I would still be in favor of removing the filter from the URL if the default has been manually selected. As said, we rarely change the filter and even in those cases I would guess that the user would like the defaults over their custom filter, if the filter is equivalent to the current default anyways. For example, if a new lint group would be added, they would probably one that one to be selected.
> 
> My main argument for this, is that it would make the URL a lot shorter in most cases.
> 
> Another change I would like to see, is that the All button for lint levels also removes the filter from the URL, as it's equivalent to the Default button for lint groups.

Makes sense to me. I'll let you know when I have the rest of the tweaks in and think it's ready for another peek.

whee:
Simplifying the handling of default values was a good call; the code got a bit simpler as well.

Hopefully this behavior makes a bit more sense:
https://whee.github.io/rust-clippy/master/

(This PR matches that deployed version.)

The only change I haven't been able to make is getting that `/` out of the path. AngularJS's `$location` service seems to expect it when using `path`. If I use `hash` instead, then somehow it's doubling up the `#`. Enabling `html5Mode` on the `$locationProvider` removes the `#` entirely and breaks existing links, and fixing up the href on the window doesn't stick -- Angular ends up overwriting it again anyway. I haven't found a good solution to this yet, other than maybe rolling back to the custom path handling, which might necessitate custom handling of the rest of the parameters.

xFrednet (approved):

xFrednet:
> The only change I haven't been able to make is getting that `/` out of the path. AngularJS's `$location` service seems to expect it when using `path`. If I use `hash` instead, then somehow it's doubling up the `#`. Enabling `html5Mode` on the `$locationProvider` removes the `#` entirely and breaks existing links, and fixing up the href on the window doesn't stick -- Angular ends up overwriting it again anyway. I haven't found a good solution to this yet, other than maybe rolling back to the custom path handling, which might necessitate custom handling of the rest of the parameters.

It looks a bit odd, but the old links still work and the new links also work, so I think it's fine either way. That was a tiny nit, but I don't think we should block the PR based on this :)

---

With that, I can say that this version looks good to me. Thank you for this update, I think it turned out pretty well!

@bors r+

PR #10835, opened 2023-05-26
new lint: `drain_collect`

### LINKED ISSUE #10818: new lint: avoid `vec.drain(..).collect()` in favor of `std::mem::take(&mut vec)`
### What it does

This lint would see the use of `vec.drain(..).collect()`, which moves all the items from a `Vec<T>` into a new `Vec<T>` via `FromIterator`, and suggest to instead use `std::mem::take(&mut vec)`.

This is similar to https://github.com/rust-lang/rust-clippy/issues/9339, implemented as [`clear_with_drain`](https://rust-lang.github.io/rust-clippy/master/index.html#clear_with_drain).

### Lint Name

collect_with_drain

### Category

perf, pedantic

### Advantage

- Avoids `FromIterator` and new allocations.
- (Subjectively) clearer code.

### Drawbacks

- By taking the Vec and replacing it with a new one, you've moved the capacity, but the author may have intended to keep the capacity in place and perform the allocation/memcpy. This situation seems rare to me.

### Example

```rust
// self.vec is a Vec<u32>
fn func(&mut self) -> Vec<u32> {
    self.vec.drain(..).collect()
}
```

Could be written as:

```rust
fn func(&mut self) -> Vec<u32> {
    std::mem::take(&mut self.vec)
}
```

y21:
Closes #10818.

This adds a new lint that looks for `.drain(..).collect()` and suggests replacing it with `mem::take`.

changelog: [`drain_collect`]: new lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4737,6 +4737,7 @@ Released 2018-09-13
 [`double_must_use`]: https://rust-lang.github.io/rust-clippy/master/index.html#double_must_use
 [`double_neg`]: https://rust-lang.github.io/rust-clippy/master/index.html#double_neg
 [`double_parens`]: https://rust-lang.github.io/rust-clippy/master/index.html#double_parens
+[`drain_collect`]: https://rust-lang.github.io/rust-clippy/master/index.html#drain_collect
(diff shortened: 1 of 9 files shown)


dswij (approved):
Looks great, thanks for this!

Issue #10825, opened 2023-05-25
regex::bytes::Regex::new("(?-u)."): false pos.: "pattern can match invalid UTF-8"


safinaskar:
### Summary

This bug is two bugs combined: false positive one and false negative one.

Let's begin with false positive.

Clippy rejects this absolutely legit code: `regex::bytes::Regex::new("(?-u).")`. Clippy says: "pattern can match invalid UTF-8". Yes, the pattern can. And this is intentional. This is why `regex::bytes` exists.

Now let's talk about false negative.

Clippy accepts this buggy code: `regex::Regex::new("(?-u).").unwrap()`. This regex is wrong, because `regex::Regex` regexes should never match invalid UTF-8.

Again:
- `regex::Regex` regexes never match invalid UTF-8. Attempt to create such regex will fail in runtime. For example, `regex::Regex::new("(?-u).").unwrap()` will fail in runtime. And so should clippy, but it does not
- `regex::bytes::Regex` are allowed to match invalid UTF-8. So, `regex::bytes::Regex::new("(?-u).")` is okay, but Clippy rejects it
- Also, note that `regex::bytes::Regex::new(".")` does not match invalid UTF-8

### Lint Name

invalid_regex

### Reproducer

-

### Version

```text
The same play.rust-lang.org uses. (It reports clippy 0.1.71 (2023-05-23 5ea3f0a))
```

safinaskar:
As well as I understand, @TimNN tried to add proper UTF-8 support in https://github.com/rust-lang/rust-clippy/pull/2645 , but for some reasons this currently doesn't work

safinaskar:
Let me again say what should be added to tests:
- `regex::Regex::new(".")` should be ✅accepted
- `regex::Regex::new("(?u).")` should be ✅accepted
- `regex::Regex::new("(?-u).")` should be ✗rejected
- `regex::bytes::Regex::new(".")` should be ✅accepted
- `regex::bytes::Regex::new("(?u)."` should be ✅accepted
- `regex::bytes::Regex::new("(?-u)."` should be ✅accepted

This is whatever crate regex does. Please, add this to tests

lochetti:
Is someone already working on this? If not, I would like to claim the issue and try to work on it.

Alexendoo:
Feel free to claim it!

PR #10839, opened 2023-05-28
Fixing `invalid_regex` with invalid UTF8. Also, adding more test cases

lochetti:
Fixing false positive and false negative when dealing with regex that could match invalid UTF8.

This PR fixes https://github.com/rust-lang/rust-clippy/issues/10825

changelog: [`invalid_regex`]: Fixing false positive and false negative when dealing with regex that could match invalid UTF8


llogiq:
Thank you!

@bors r+

PR #10840, opened 2023-05-28
from_over_into: Show suggestions for non-Self expanded paths

Alexendoo:
changelog: [`from_over_into`]: Show suggestions when the body contains macros not expanding to `Self`

Currently any path in a macro expansion causes the suggestion to be hidden, meaning most macro calls cause it to be hidden

Now it's only hidden if the expansion contains `Self`


dswij (approved):
Nice, thanks for this!

PR #10844, opened 2023-05-30
Don't lint `let_unit_value` when `()` is explicit

### LINKED ISSUE #9048: let_unit_value should be allowed when I explicitly write "let () = ..."
### Summary

Sometimes I write code like this: `let () = some_function_call()`. I do this to make sure that this function returns unit value, i. e. to make sure that I didn't miss any return value. But let_unit_value warns on this code. So, please, make it not to warn on this code

### Lint Name

let_unit_value

### Reproducer

I tried this code:

```rust
fn main() {
    let mut v = vec![2, 1];
    let () = v.sort();
}
```

I saw this happen:

```
warning: this let-binding has unit value
 --> src/main.rs:3:5
  |
3 |     let () = v.sort();
  |     ^^^^^^^^^^^^^^^^^^ help: omit the `let` binding: `v.sort();`
  |
  = note: `#[warn(clippy::let_unit_value)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_unit_value

    Finished dev [unoptimized + debuginfo] target(s) in 0.42s
```

(no warn)

### Version

```text
Version from https://play.rust-lang.org/ as on today (2022-06-25)
```

### Additional Labels

_No response_

Centri3:
since these are explicitly written (and not the result of a function call or anything else), they should be allowed, as they are both useful in some cases described in #9048

Fixes #9048

changelog: [`let_unit_value`]: Skip `let () = { ... }` and `let { ... } = ()`


clippy_lints/src/unit_types/let_unit_value.rs
```diff
@@ -13,56 +13,68 @@ use rustc_middle::ty;
...
+    if let PatKind::Tuple(fields, ..) = local.pat.kind && fields.is_empty() {
+        return;
+    }
+
+    if_chain! {
+        if let Some(init) = local.init;
+        if !local.pat.span.from_expansion();
+        if !in_external_macro(cx.sess(), local.span);
+        if cx.typeck_results().pat_ty(local.pat).is_unit();
+        then {
```
llogiq, line 26:
```suggestion
        if let Some(init) = local.init;
        if !local.pat.span.from_expansion() &&
            !in_external_macro(cx.sess(), local.span) &&
            cx.typeck_results().pat_ty(local.pat).is_unit() {
```

Also remove the stray `}` on the way out.

llogiq (commented):
Thanks for taking this on, also I much like the additions to the tests. One thing that befuddled me is why did you regress from a `let` chain to an `if_chain!` macro invocation?

clippy_lints/src/unit_types/let_unit_value.rs
Centri3: I found it a tad easier to read in this case, but I'm ok with switching it back.

Centri3: Fixed

Centri3: I've just seen this zulip discussion - https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/removing.20if_chain.20dependency/near/292061902, didn't know clippy will at some point transition away from them, so I'll make sure to use let chains instead in future PRs

clippy_lints/src/unit_types/let_unit_value.rs
```diff
@@ -13,17 +13,27 @@ use rustc_middle::ty;
...
+
     if let Some(init) = local.init
         && !local.pat.span.from_expansion()
         && !in_external_macro(cx.sess(), local.span)
         && cx.typeck_results().pat_ty(local.pat).is_unit()
     {
+        // skip `let awa = ()`
+        if let ExprKind::Tup(elements) = init.kind && elements.is_empty() {
+            return;
+        }
```
danielhenrymantilla, line 29: Thanks for including this case ❤️

tests/ui/let_unit.fixed
```diff
@@ -15,16 +15,25 @@ fn main() {
...
+        // do not lint this, since () is explicit
+        let _a = ();
     }
 
+    // do not lint this, since () is explicit
+    let () = dummy();
+    let () = ();
+    () = dummy();
+    () = ();
+
```
danielhenrymantilla, line 27: If we could cover the following as well it would be quite neat:

```suggestion
let _: () = dummy();

```

Centri3: Fine by me

Alexendoo (approved):

Alexendoo:
r=me with rebase/squash

kupiakos:
Any chance this can be rebased and merged?

cocodery:
I have tried locally, merging `master` branch and solving one small conflict will be fine. 
And related [Issue#12017](https://github.com/rust-lang/rust-clippy/issues/12017) can tag fixed in the first comment.
@Centri3

Issue #10836, opened 2023-05-27
Double not incorrectly triggers `nonminimal_bool` when !v returns `bool` but v is not `bool`


endorpersand:
### Summary

When a type implements `Not` and outputs `bool`, Clippy suggests to simplify the expression `!!x` to just `x`, even though that isn't equivalent.

### Lint Name

nonminimal_bool

### Reproducer

I tried this code:

```rust
struct Foo(bool);
impl std::ops::Not for Foo {
    type Output = bool;

    fn not(self) -> Self::Output {
        !self.0
    }
}

fn main() {
    let _: bool = !!Foo(true);
}
```

I saw this happen:

```
warning: this boolean expression can be simplified
  --> src/report.rs:11:19
   |
11 |     let _: bool = !!Foo(true);
   |                   ^^^^^^^^^^^ help: try: `Foo(true)`
   |
```

I expected to see this happen:
This lint shouldn't have triggered. 

### Version

```text
rustc 1.69.0 (84c898d65 2023-04-16)
binary: rustc
commit-hash: 84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc
commit-date: 2023-04-16
host: x86_64-apple-darwin
release: 1.69.0
LLVM version: 15.0.7
```

xFrednet:
This probably only requires a type check for the inner expression and some tests. It should be a good first issue :)

@rustbot label +good-first-issue

lochetti:
Maybe we can close this issue, after the PR https://github.com/rust-lang/rust-clippy/pull/10845?

PR #10845, opened 2023-05-30
nonminimal_bool fix double not

disco07:
fix issue https://github.com/rust-lang/rust-clippy/issues/10836

changelog: Fix [`nonminimal_bool`] false positive when `!!x`, `x` isn't boolean and implements `Not`

### DIFF
--- a/clippy_lints/src/booleans.rs
+++ b/clippy_lints/src/booleans.rs
@@ -8,10 +8,12 @@ use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, FnKind, Visitor};
 use rustc_hir::{BinOpKind, Body, Expr, ExprKind, FnDecl, UnOp};
 use rustc_lint::{LateContext, LateLintPass, Level};
+use rustc_middle::ty::{self, Ty};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::def_id::LocalDefId;
 use rustc_span::source_map::Span;
 use rustc_span::sym;
+use rustc_span::symbol::Ident;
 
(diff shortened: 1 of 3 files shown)


clippy_lints/src/booleans.rs
```diff
@@ -473,6 +496,12 @@ impl<'a, 'tcx> Visitor<'tcx> for NonminimalBoolVisitor<'a, 'tcx> {
                     self.bool_expr(e);
                 },
                 ExprKind::Unary(UnOp::Not, inner) => {
+                    if let ExprKind::Unary(UnOp::Not, ex) = inner.kind {
+                        let ty = self.cx.typeck_results().expr_ty(ex);
+                        if is_impl_not_trait_with_bool_out(self.cx, ty) {
+                            return;
+                        }
```
samueltardieu, line 503: Out of curiosity, why not just check if `ex` is a `bool`?

samueltardieu:
> changelog: none

There probably should be a changelog there as this is a bugfix to an existing lint.

clippy_lints/src/booleans.rs
disco07: Thanks you for your feedback. I try this

disco07: Are you some method to suggest to me please? When I try `cx.typeck_results().expr_ty(..)` on struct, it returns struct name, not bool type

samueltardieu:
Note that this only seems to fix the particular double not case. Other cases are problematic, such as 

```rust
   let _a = !(!S && !S);
```

when `S` implements `std::ops::Not<Output = bool>`: Clippy will suggest to replace this with `S`.

Issue #10838, opened 2023-05-27
auto fix failed converting `Into` to `From` with `Self`


AstrickHarren:
### Summary

Looks like we need to expand `Self` before converting `Into` to `From`, but I could (almost certainly) be inconclusive. 

### Reproducer

I to clippy fix this code:

```rust
impl<X> Into<Rval<Self>> for Lval<X> {
  fn into(self) -> Rval<Self> {
    Rval::Val(self)
  }
}
```

I expected to see this happen:
```rust
impl<X> From<Lval<X>> for Rval<Lval<X>> {
  fn from(lval: Lval<X>) -> Self {
    Rval::Val(lval)
  }
}
```

Instead, this happened:

```rust
impl<X> From<Lval<X>> for Rval<Self> {
  fn from(val: Lval<X>) -> Self {
    Rval::Val(val)
  }
}
```

### Version

```text
rustc 1.69.0 (84c898d65 2023-04-16)
binary: rustc
commit-hash: 84c898d65adf2f39a5a98507f1fe0ce10a2b8dbc
commit-date: 2023-04-16
host: aarch64-apple-darwin
release: 1.69.0
LLVM version: 15.0.7
```

Alexendoo:
TIL you can put `Self` in that position

PR #10853, opened 2023-05-31
Ignore fix for `from_over_into` if the target type contains a `Self` reference

MarcusGrass:
Fixes https://github.com/rust-lang/rust-clippy/issues/10838.

This is my first time contributing here, and the fix is kind of ugly.
I've worked a bit with `quote` and was trying to figure out a way to replace the type in a better way than just a raw string-replace but couldn't quite figure out how to.

The only thing really required to fix this, is to replace all `Self` references with the type stated in the `from` variable, this isn't entirely simple to do with raw strings without creating a mess though. 

We need to find and replace all `Self`'s in a variable with `from` but there could be an arbitrary amount, in a lot of different positions. As well as some type that contains the name self, like `SelfVarSelf` which shouldn't be replaced.

The strategy is essentially, if `"Self"` is surrounded on both sides by something that isn't alphanumeric, then we're golden, then trying to make that reasonably efficient.

I would not be offended if the solution is too messy to accept!

changelog: [from_over_into]: Replace Self with the indicated variable in suggestion and fix.

### DIFF
--- a/clippy_lints/src/from_over_into.rs
+++ b/clippy_lints/src/from_over_into.rs
@@ -80,11 +80,6 @@ impl<'tcx> LateLintPass<'tcx> for FromOverInto {
             && cx.tcx.is_diagnostic_item(sym::Into, middle_trait_ref.def_id)
             && !matches!(middle_trait_ref.substs.type_at(1).kind(), ty::Alias(ty::Opaque, _))
         {
-            if !target_ty.find_self_aliases().is_empty() {
-                // It's tricky to expand self-aliases correctly, we'll ignore it to not cause a
-                // bad suggestion/fix.
-                return;
-            }
             span_lint_and_then(
                 cx,
                 FROM_OVER_INTO,
@@ -161,6 +156,11 @@ fn convert_to_from(
     self_ty: &Ty<'_>,
     impl_item_ref: &ImplItemRef,
 ) -> Option<Vec<(Span, String)>> {
+    if !target_ty.find_self_aliases().is_empty() {
+        // It's tricky to expand self-aliases correctly, we'll ignore it to not cause a
+        // bad suggestion/fix.
+        return None;
+    }
     let impl_item = cx.tcx.hir().impl_item(impl_item_ref.id);
     let ImplItemKind::Fn(ref sig, body_id) = impl_item.kind else { return None };
     let body = cx.tcx.hir().body(body_id);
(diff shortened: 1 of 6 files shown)


Alexendoo:
Tricky one for a first contribution 😄

We might be able to do it without re-lexing by getting the spans of all the `Self`s and calculating the `BytePos` offset from the start of `target_ty`'s span, but that has a bunch of its own problems and is also quite complicated

Maybe we should just stick

```rust
if !target_ty.find_self_aliases().is_empty() {
    return None;
}
```

in `suggest_to_from`, `Self` in that position should be relatively rare so it wouldn't be a huge deal to lose the suggestion, the lint itself will still fire

MarcusGrass:
> Tricky one for a first contribution smile
> 
> We might be able to do it without re-lexing by getting the spans of all the `Self`s and calculating the `BytePos` offset from the start of `target_ty`'s span, but that has a bunch of its own problems and is also quite complicated
> 
> Maybe we should just stick
> 
> ```rust
> if !target_ty.find_self_aliases().is_empty() {
>     return None;
> }
> ```
> 
> in `suggest_to_from`, `Self` in that position should be relatively rare so it wouldn't be a huge deal to lose the suggestion, the lint itself will still fire

Sounds reasonable, now the change doesn't look like a 3AM, 5th-attempt, third-redbull leetcode submission. 

Hopefully this was correct, I did the exit as early as possible, the stderr file is unchanged since the lint is just ignored, and the .fixed file keeps the original code, I think that's how it's supposed to look

clippy_lints/src/from_over_into.rs
```diff
@@ -80,6 +80,11 @@ impl<'tcx> LateLintPass<'tcx> for FromOverInto {
             && cx.tcx.is_diagnostic_item(sym::Into, middle_trait_ref.def_id)
             && !matches!(middle_trait_ref.substs.type_at(1).kind(), ty::Alias(ty::Opaque, _))
         {
+            if !target_ty.find_self_aliases().is_empty() {
+                // It's tricky to expand self-aliases correctly, we'll ignore it to not cause a
+                // bad suggestion/fix.
+                return;
+            }
```
Alexendoo, line 87: It can go in `convert_to_from` so that we still lint these cases, just that instead of a full suggestion it'll ask you to figure it out yourself

The test case will probably need to be moved to `from_over_into_unfixable.rs` (or however it's spelled) also

MarcusGrass: Ah I see, thanks a lot for the patience, now it produces a lint without a suggestion!

Alexendoo (approved):

Alexendoo:
Thanks @MarcusGrass!

@bors r+

PR #10855, opened 2023-06-01
Fix suggestion on fully qualified syntax

### LINKED ISSUE #10850: `explicit_deref_methods` malformed suggestion on fully qualified syntax
### Summary

`explicit_deref_methods` gives a malformed suggestion when it's emitted on fully qualified syntax.

### Reproducer

I tried this code:

```rust
<Foo as Deref>::deref(foo);
```

Where `Foo` implements `Deref`.

I saw this happen:

```
&*<Foo as Deref>::deref;
```

```
&*foo;
```

### Version

```text
rustc 1.71.0-nightly (a2b1646c5 2023-05-25)
binary: rustc
commit-hash: a2b1646c597329d0a25efa3889b66650f65de1de
commit-date: 2023-05-25
host: x86_64-pc-windows-gnu
release: 1.71.0-nightly
LLVM version: 16.0.4
```

### Additional Labels

Centri3:
fixes #10850

changelog: [`explicit_deref_methods`]: Fix malformed suggestion on `Foo::deref(&foo)`

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -56,9 +56,11 @@ declare_clippy_lint! {
     /// let b = &*a;
     /// ```
     ///
-    /// This lint excludes:
+    /// This lint excludes all of:
     /// ```rust,ignore
     /// let _ = d.unwrap().deref();
+    /// let _ = Foo::deref(&foo);
+    /// let _ = <Foo as Deref>::deref(&foo);
     /// ```
     #[clippy::version = "1.44.0"]
     pub EXPLICIT_DEREF_METHODS,
@@ -155,7 +157,7 @@ impl_lint_pass!(Dereferencing<'_> => [
 
 #[derive(Default)]
 pub struct Dereferencing<'tcx> {
-    state: Option<(State<'tcx>, StateData)>,
+    state: Option<(State, StateData)>,
 
     // While parsing a `deref` method call in ufcs form, the path to the function is itself an
     // expression. This is to store the id of that expression so it can be skipped when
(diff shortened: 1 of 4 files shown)


clippy_lints/src/dereference.rs
```diff
@@ -1503,12 +1512,25 @@ fn report<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, state: State, data
...
             } else {
                 expr_str.into_owned()
             };
 
+            // Fix #10850, changes suggestion if it's `Foo::deref` instead of `foo.deref`. Since `Foo::deref` is
+            // a `Call` instead of a `MethodCall` this should catch all instances of this, even if it's fully
+            // qualified or whatnot.
+            if is_final_ufcs && let Some(args) = call_args {
+                // Remove ref if it's there
+                let arg = if let ExprKind::AddrOf(.., arg) = args[0].kind {
```
Centri3, line 1526: Is there a function that does this for me? Since this does feel a tad hacky.

Centri3: I also feel this could break on something like `*foo` if it's `&&Foo` but I haven't tested this yet.

Alexendoo:
The `Foo` part can participate in type inference

```rust
use std::ops::Deref;

fn main() {
    let a = Default::default();
    let b = String::deref(&a);
}
```

Maybe we ought to preserve `Foo::deref` anyway though as unlike `.deref()` it does provide more information to the reader than the equivalent `&*`, what type is being deref'd can be confusing

llogiq:
Thanks @Alexendoo for taking on the review, and thanks @Centri3 for pushing this forward!

@bors r+

Issue #9921, opened 2022-11-21
[documentation] Where to place clippy.toml is hard to find


arifd:
### Summary

I expected to find the documentation where to place the clippy.toml file here: https://doc.rust-lang.org/clippy/configuration.html

I eventually found it here: https://doc.rust-lang.org/clippy/development/adding_lints.html#adding-configuration-to-a-lint
> Clippy supports the configuration of lints values using a clippy.toml file in the workspace directory

Which is deep under development, and far too late for the general user.

Many thanks for the most excellent clippy!

PR #10857, opened 2023-06-01
Explain which paths clippy searches for configuration in docs

MarcusGrass:
Fixes https://github.com/rust-lang/rust-clippy/issues/9921.

Adds information on where to place the configuration files, it may be a bit verbose. Also added a comment to the section of the code where the search happens, to hopefully prevent changing that without updating the docs.

changelog: Make documentation about where to place configuration files clearer.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -2,8 +2,10 @@
 
 > **Note:** The configuration file is unstable and may be deprecated in the future.
 
-Some lints can be configured in a TOML file named `clippy.toml` or `.clippy.toml`. It contains a
-basic `variable = value` mapping e.g.
+Some lints can be configured in a TOML file named `clippy.toml` or `.clippy.toml`, placed in a directory specified by
+the environment variable `CLIPPY_CONF_DIR`, or if that's not found, the environment variable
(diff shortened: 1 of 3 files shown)


book/src/configuration.md
```diff
@@ -2,8 +2,10 @@
 
 > **Note:** The configuration file is unstable and may be deprecated in the future.
 
-Some lints can be configured in a TOML file named `clippy.toml` or `.clippy.toml`. It contains a
-basic `variable = value` mapping e.g.
+Some lints can be configured in a TOML file named `clippy.toml` or `.clippy.toml`, placed in a directory specified by
+the environment variable `CLIPPY_CONF_DIR`, or if that's not found, the environment variable
```
flip1995, line 6: I would put this in a list, so

```
- `CLIPPY_CONF_DIR` ...; or
- ...; or
- ...
```

instead of in text. Would make it easier to read.

flip1995 (commented):
One comment on how to improve this slightly.

flip1995:
Thanks, an ordered list is even better.

@bors r+

PR #10858, opened 2023-06-01
Add `external` feature, separating tests with external dependencies from those that don't have

### LINKED ISSUE #8206: Add a feature to enable/disable external crates in uitests
### Description

In #8179 two new dependencies for our UI-tests have been added,  `futures` and `tokio`. This increases the pipeline runtime by about 3 minutes. The additional time is mostly spent on compiling, from what I can tell. For local tests, I would like to have external crates like these disabled by default and only enable them if I actually develop on related lints and in the pipeline.

This is how I imagine it:

```
cargo uitest
```
This runs all tests which only require Clippy. uitests with external crates are ignored on this case

```
cargo uitest --features=external-dependencies
```
Runs all uitest, including those that need external dependencies. This would be the default option in the pipeline

### Version

_No response_

### Additional Labels

cc: @flip1995 you're our CI wizard :mage:. Maybe you have some thoughts on the matter :upside_down_face:

blyxyas:
I'm sorry to whoever will review this. I'm really sorry.
The main changes are done to `tests/compile-test.rs`, `Cargo.toml`, and creating `tests/ui-external`

This PR would separate tests that have external dependencies, to those that don't have. This should speed up by a lot testing time.

It's still a draft because it doesn't work yet. More info below:

## Known issues:

* When using `serde` as `external::serde::<macro>`, it doesn't work (rustc thinks we're using a sysroot dependency, for some reason??). Maybe enabling `#![feature(rustc_private)]` could fix this, I don't know where to enable it tho?

* [single-component-path-imports](https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports) just doesn't lint anything. This may be related to changing from `extern crate serde` to `use external::serde` (maybe a lint issue?)

Fixes #8206
r? @ghost

changelog:none


Alexendoo:
Worth mentioning #10426 would change how test deps are built

blyxyas:
I guess that PR makes this one pretty much useless.
Should I close it?

Alexendoo:
Oops sorry forgot to reply there, it could probably still be done but would have to be rewritten on top of those changes once they arrive, however personally I'd say we have too many tests now that use dependencies for it to be worth it. The cost of compiling test deps is also a one time thing as far as the developer experience goes so shouldn't be too much of a problem

Issue #4182, opened 2019-06-07
let_and_return "known problems"


alexander-irbis:
Hi!
The documentation page of `let_and_return` https://rust-lang.github.io/rust-clippy/master/index.html#let_and_return , referenced with the lint, mentions that there are no known problems with it. Maybe https://github.com/rust-lang/rust/issues/37612 should be mentioned as a subtle and elusive problem.

PR #10859, opened 2023-06-01
Bring up Rust lang #37612 as a known problem for let_and_return

MarcusGrass:
Fixes https://github.com/rust-lang/rust-clippy/issues/4182.

I don't think conforming to this lint could trigger the issue immediately, only if subsequent code-changes go wrong, but I may be mistaken. 

Since the lint can't trigger it by itself, just closing this issue might be reasonable, if not maybe this PR fixes it.

changelog: Update docs for `let_and_return`, mention rust-lang #37612


giraffate:
@bors r+

Thanks!

PR #10863, opened 2023-06-01
Add lints to disallow non-`#![no_std]` and non-`#![no_core]` dependencies in crates with those attributes

### LINKED ISSUE #1570: Lint non-`#![no_std]` dependencies of a `#![no_std]` crate
A `#![no_std]` crate can depend on crates that aren't `#![no_std]`, which results in libstd being pulled in regardless, defeating the purpose of `#![no_std]`.

Centri3:
These lints are simple, which is essentially required due to how little info is (seemingly) available about the dependency tree, unfortunately. (If this isn't the case, I'm happy to reimplement it!) Therefore we can't even give a helpful suggestion, so something like `cargo-nono` like mentioned in #1570 would likely still be necessary to figure out the offending crate.

Also, tests for `#![no_core]` are TODO, and more complex tests in general.

Closes #1570

changelog: Add `no_core_crate_core_dependent` and `no_std_crate_std_dependent` lints


flip1995:
Would `cargo metadata` help with figuring out the dependency chain and which crate(s) introduce `std`/`core`? In that case this lint could be implemented as a `cargo` lint.

Centri3:
I believe so, yes. But that would be essentially reinventing the wheel as `cargo-nono` does exactly that already, but I'm not too opposed to the idea

flip1995:
> as `cargo-nono` does exactly that already,

Ah, yeah copying another tool into Clippy doesn't make much sense.

Then my question is: Why would we need a less capable version of `cargo-nono` in Clippy? Let's just point to `cargo-nono` if people ask about this functionality in the future.

dswij:
FWIU, we're not checking conditional `no_std` like cargo-nono does

> or the often used conditional #![cfg_attr(not(feature = "std"), no_std)]

Maybe we can improve on this in future PR?

Centri3:
> FWIU, we're not checking conditional no_std like cargo-nono does

I don't think this matters, as the lint specifically looks for if `libstd` is linked; it doesn't do any guesswork.

> It's more of a dependency scan than a lint (I can't recall if we have similar lints like this for Clippy).

There are a couple but they're very old (and in the `cargo` category.)

xFrednet:

PR #10864, opened 2023-06-01
[`unnecessary_lazy_eval`]: don't lint on types with deref impl

### LINKED ISSUE #10437: `unnecessary_lazy_evaluations` and `Deref`
### Summary

Clippy seems to assume that `Deref` is always pure, and suggest rewriting `.unwrap_or_else(|| *x)` to `.unwrap_or(*x)`. This is not correct when for example `lazy_static` is involved.

### Lint Name

unnecessary_lazy_evaluations

### Reproducer

```rust
use std::ops::Deref;

struct A;
impl Deref for A {
	type Target = u32;

		println!("Yes hello am dereffed");
		&0
	}
}

fn main() {
	println!("unwrap_or_else");
	Some(4).unwrap_or_else(|| *A);
	println!("unwrap_or");
	Some(4).unwrap_or(*A);
}
```

Clippy suggest rewriting the `unwrap_or_else` into `unwrap_or`, as a few lines below. But the type `A` has a nontrivial `Deref`, so they are not equal.

### Version

```text
rustc 1.69.0-nightly (7aa413d59 2023-02-19)
binary: rustc
commit-hash: 7aa413d59206fd511137728df3d9e0fd377429bd
commit-date: 2023-02-19
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

y21:
Fixes #10437.
This PR changes clippy's util module `eager_or_lazy` to also consider deref expressions whose type has a non-builtin deref impl and not suggest replacing it as that might have observable side effects.
A prominent example might be the `lazy_static` macro, which creates a newtype with a `Deref` impl that you need to go through to get access to the inner value. Going from lazy to eager can make a difference there.

changelog: [`unnecessary_lazy_eval`]: don't lint on types with non-builtin deref impl

### DIFF
--- a/clippy_utils/src/eager_or_lazy.rs
+++ b/clippy_utils/src/eager_or_lazy.rs
@@ -173,11 +173,15 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
                         self.eagerness = Lazy;
                     }
                 },
-
+                // Custom `Deref` impl might have side effects
+                ExprKind::Unary(UnOp::Deref, e)
+                    if self.cx.typeck_results().expr_ty(e).builtin_deref(true).is_none() =>
+                {
+                    self.eagerness |= NoChange;
+                },
                 // Dereferences should be cheap, but dereferencing a raw pointer earlier may not be safe.
                 ExprKind::Unary(UnOp::Deref, e) if !self.cx.typeck_results().expr_ty(e).is_unsafe_ptr() => (),
                 ExprKind::Unary(UnOp::Deref, _) => self.eagerness |= NoChange,
-
                 ExprKind::Unary(_, e)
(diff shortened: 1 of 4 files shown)


xFrednet:
Assuming it's fine if I steal the PR review :)

@blyxyas Would you mind reviewing this PR?

r? @xFrednet

blyxyas:
Yep, Gonna review it right now ^^

blyxyas (approved):
LGTM, Thanks!

y21:
thanks for the review! not sure if @blyxyas has r+ rights, so maybe @xFrednet you need to if the changes look good? :)

xFrednet (approved):

xFrednet:
Yes, sorry for the delay. This looks good to me as well! Thank you!

@bors r=blyxyas,xFrednet

PR #10875, opened 2023-06-02
Cleanup gh-pages: Remove v0.0.* documentation

flip1995:
The version numbers v0.0.* are pretty meaningless. They were used back in the day, where Clippy was deployed through crates.io. Today, distinguishing between them doesn't serve any value IMO. This is why I propose to unify them under one `pre-1.29.0` version documentation and remove the old documentation.

The next step would be to also remove the tags from the repo and replace the `v0.0.212` tag with a `pre-1.29.0` tag.

---

We probably won't be able to use bors with this PR. If approved by all of @rust-lang/clippy, I will push it to the `gh-pages` branch.

Must be merged together with #10876 (best with a closed tree)


flip1995:
Nominating to talk about this in the next meeting at the latest.

Manishearth (approved):
In favor. Little bit worried about old links but i think everyone used the master link anyway

Alexendoo (approved):
Sounds good to me also, maybe worth keeping the tags for the historically curious but there are rather a lot of them

dswij (approved):
Definitely in favor of this

xFrednet:
cc: @blyxyas @Centri3 since the team was pinged :)

xFrednet (approved):
I'm also in favor of this! :)

blyxyas (approved):
I also see that documentation as just taking space.
"Only" [129 code results](https://github.com/search?q=%2Fhttps%3A%5C%2F%5C%2Frust-lang.github.io%5C%2Frust-clippy%5C%2Fv0%5C.0%5C.%2F&type=code) for links to that v0.0.* documentation, and none of those has a lot of stars.

Centri3 (approved):
Looks like only half a dozen or so instances linking to this aren't from `rust-lang-nursery.github.io`, which just redirects you anyway. I don't see much of a reason to keep these

flip1995:
Here is a preview: https://flip1995.github.io/rust-clippy/

PR #10876, opened 2023-06-02
Adapt versions.html file to cleaned up gh-pages

flip1995:
Companion PR to #10875

changelog: Remove legacy v0.0.* versions from Clippy repository and documentation. 

Must be merged together with #10875 (best with a closed tree)

r? @Alexendoo (because you were randomly selected on the other PR :P)


xFrednet:
I'm also adding the I-nominated label here to indicate that it's waiting on the next meeting :)

@rustbot label +I-nominated

PR #10881, opened 2023-06-03
[`useless_conversion`]: pluralize if there are multiple `.into_iter()` calls

y21:
context: https://github.com/rust-lang/rust-clippy/pull/10814#issuecomment-1575036086

changelog: [`useless_conversion`]: pluralize if there are multiple `.into_iter()` calls in the chain

r? @llogiq

### DIFF
--- a/clippy_lints/src/useless_conversion.rs
+++ b/clippy_lints/src/useless_conversion.rs
@@ -93,14 +93,13 @@ fn into_iter_call<'hir>(cx: &LateContext<'_>, expr: &'hir Expr<'hir>) -> Option<
 /// Same as [`into_iter_call`], but tries to look for the innermost `.into_iter()` call, e.g.:
 /// `foo.into_iter().into_iter()`
 ///  ^^^  we want this expression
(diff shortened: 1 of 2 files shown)


clippy_lints/src/useless_conversion.rs
```diff
@@ -93,12 +93,14 @@ fn into_iter_call<'hir>(cx: &LateContext<'_>, expr: &'hir Expr<'hir>) -> Option<
...
+    let mut depth = 0;
     loop {
         if let Some(recv) = into_iter_call(cx, expr) {
             expr = recv;
+            depth += 1;
         } else {
-            return Some(expr);
+            return Some((expr, depth));
         }
     }
```
llogiq, line 105: While we're at it, could we make this a `while let`? I think it'll look neater.

llogiq (approved):
Thanks, that was quick. One small nit, otherwise r=me.

llogiq:
Thank you!

@bors r+

PR #10884, opened 2023-06-03
New lint [`needless_raw_string_hashes`]

### LINKED ISSUE #10882: Lint to warn against unnecessary extra hashes in raw string literals
### What it does

Checks for the use of more hashes that are necessary in `r#…#`-style literals, and emits a warning about it telling you to use fewer hashes. Suggested name: `needless_raw_string_hashes`; suggested categories: `style` or `pedantic`.

### Advantage

- Code is kept more consistent; there are fewer ways to express one piece of logic.
- Code automatically self-documents the level of “rawness” needed in the string literal.

### Drawbacks

- Sometimes for alignment purposes one might want to add more hashes than necessary.
- Maybe this should be in `rustfmt` instead?

### Example

```rust
let s = r##"Hello "world"!"##;
```

Could be written as:

```rust
let s = r#"Hello "world"!"#;
```

Centri3:
Emits a warning when there are an extraneous number of hashes(?) around a raw string literal, for example `r##"I'm a "raw string literal"!"##` or `cr#"crunb"#`

Closes #10882

I think this could also fit in `style` as well, rather than `complexity`.

changelog: Add [`needless_raw_string_hashes`] and [`needless_raw_string`] lints


Centri3:
Seems the only instance of this from running `lintcheck` is in anyhow's `build.rs`, which is very similar to clippy's `CARGO_CLIPPY_HELP` so maybe this can ignore `r#`?

clippy_lints/src/raw_strings.rs
```diff
@@ -0,0 +1,109 @@
...
+            };
+            if !snippet(cx, expr.span, prefix).trim().starts_with(prefix) {
+                return;
+            }
+
+            #[expect(clippy::cast_possible_truncation)]
+            let req = lit.symbol.as_str().as_bytes()
+                .split(|&b| b == b'"')
+                .skip(1)
+                .map(|bs| 1 + bs.iter().take_while(|&&b| b == b'#').count() as u8)
```
GrishaVar, line 78: As in rust-lang/rust#112373, the value of req has an upper bound, so short-circuiting the max should (I think) make this a lot faster for very long strings.

tests/ui/needless_raw_string.rs
```diff
@@ -0,0 +1,16 @@
+//@run-rustfix
+#![allow(clippy::needless_raw_string_hashes, clippy::no_effect, unused)]
```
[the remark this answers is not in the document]
Centri3: Fixed, though I've kept the `allow`s as otherwise it'll still fire on like `r#"\s"#`

clippy_lints/src/raw_strings.rs
Centri3: Not entirely sure how to go about this :sweat_smile: sorry

dswij:
> I think this could also fit in style as well, rather than complexity

I think this fit in `style` more.

Centri3:
`style`'s fine by me

Centri3:
Now that I think about it, maybe `needless_raw_strings` should be `pedantic` or `restriction`? There are many cases where you may want a raw string even if it's unnecessary (regexes are a good example).

clippy_lints/src/raw_strings.rs
GrishaVar: Given a raw string is surrounded by `n` hashes, the contents of the literal cannot contain a `"` followed by `n` hashes. If it did, the lexer would complain. So we know the result of searching with `.max()` will not be greater than `n`. But the `.max()` doesn't know this and will search through the whole string.

Example:
`r#####"  a string containing "#### followed by an extremely long sequence of text which drones on forever...  "#####`
Once we've found the `"####` we can stop searching because we already know the number of hashes surrounding the literal is correct.

When I wrote this line, I didn't deal with this because the PR was just a proof-of-concept; readability was more important. I think the only nice way of fixing it is looping over the iterator instead of calling `.max()`.

Centri3: I went with a slightly differing approach: to stop iterating using `map_while` once `n` - 1 hashes has been found. Should be similar performance (afaik) but remains using `max`

clippy_lints/src/raw_strings.rs
```diff
@@ -0,0 +1,123 @@
...
+                .skip(1)
+                .map_while(|bs| {
+                    let max = num;
+                    let num = bs.iter().take_while(|&&b| b == b'#').count() as u8;
+                    if num != max {
+                        return Some(num + 1);
+                    }
+
+                    None
+                })
```
GrishaVar, line 104: I don't think this fixes the issue. We know from the lexer that the number of hashes inside the literal will always be strictly smaller than the number of hashes surrounding the literal, so the `num == max` case can never occur. This code will do the same thing the previous one did.

In general, any `while`-based iterators don't work for this task, since they will always allow all items that match some condition. What we want to do is match the _first_ occurrence of `hash_length = num-1`, but not any of the following such occurrences.

Also keep in mind short circuiting the `split` is the same kind of issue. Ideally, searching the literal should stop as soon as the last necessary `#` is found, not when the following `"` is encountered.

Centri3: > This code will do the same thing the previous one did.

This was my bad actually, it was meant to check `num - 1`.

> In general, any while-based iterators don't work for this task,

I don't think this is true (with `num - 1`). `map_while` returns the value from the predicate, once `None` is returned, it will stop searching because it returns `None` from `next`, which would be as if it ran out of items (thus, it will stop iterating). You can see this in its implementation: https://doc.rust-lang.org/src/core/iter/adapters/map_while.rs.html#40-44. Since `None` is only returned in the case that it equals `max - 1`, the last element will either have any number of hashes (if it finished iterating) or `max - 1` (if it returned `None`).

Though it should be possible to remove `max`, by checking whether the last element is `max - 1` before running `max` on the rest.

Centri3: I just benchmarked it; for long strings (ran 1,000,000 times, as well as a couple dozen times to remove any potential randomness) it's *way* slower, UNLESS there's a `"` with `max - 1` hashes, in which case it's between 10000%-10% the speed of the original implementation (depending on where it is in the string). The performance hit for where it's near the end is probably due to it needing to call a closure, over and over again.

The difference is not large enough for other strings (but not small enough to be unnoticeable) that it's worth it imo in its current implementation.

clippy_lints/src/raw_strings.rs
```diff
@@ -0,0 +1,124 @@
...
+                    if num.saturating_sub(1) == max {
+                        // Since `num - 1` is the max number of hashes it can have, we can
+                        // short-circuit with full confidence this is correct
+                        return None;
+                    }
+
+                    Some(num + 1)
+                })
+                .max()
+                .unwrap_or(0);
```
GrishaVar, line 107:
```suggestion
            let req = if max == 0 {
                0
            } else {
                let mut following_quote = false;
                let mut cur_len = 0u8;
                let mut max_len = 0u8;
                for b in lit.symbol.as_str().as_bytes() {
                    match b {
                        b'"' => {
                            following_quote = true;
                            cur_len = 1;
                        },
                        b'#' => if following_quote {
                            cur_len += 1;
                        },
                        _ => if following_quote {
                            following_quote = false;
                            max_len = max_len.max(cur_len);
                            if max_len == max { break }
                        },
                    }
                }
                if following_quote {
                    // extra check in case string ends with hashes
                    max_len = max_len.max(cur_len)
                }
                max_len
            };
```
This is sort of what I had in mind. I had considered some kind of `fold`ing as well, but it would look even worse and short circuiting would be rough. Depends heavily on the benchmark of course, in the stuff I tested, I found a significant speedup compared to the other versions so far for strings with `max-1` hashes and little difference when included hashes were not `max-1`.

clippy_lints/src/raw_strings.rs
GrishaVar: > This was my bad actually, it was meant to check `num - 1`.

You've swapped the `num` and `max` variables in the recent push, so it should be `max - 1` :P the code currently does the same thing again.

But, changing it to map take while `num != max - 1`, what I think you intended, would not work either. In that case, the iterator would finish without returning `num`, so it would not get considered in the max. This is what I meant about matching the same conditions: with `while` can either get all `max-1` and no `max`, or we can get all of the `max`. But we want to find just one of the `max` and stop there. It could be that I'm missing something though, let me know.

> The difference is not large enough for other strings (but not small enough to be unnoticeable) that it's worth it imo in its current implementation.

I found the difference to be unnoticeable with the new suggestion; perhaps you can try it with your benchmarks. If it doesn't look good, maybe my original line is fine to keep.

Centri3: > You've swapped the num and max variables in the recent push

I should not code while half asleep.

> But we want to find just one of the max and stop there.

This is what it does; the `max` would stop on the final element, which happens to be `max - 1`.

clippy_lints/src/raw_strings.rs
Centri3: This is indeed even faster, so thank you! But it's unreadable, so I think I'll at least try to reincorporate it into iterators.

Centri3: I've made it into an iterator, haven't benchmarked it yet but it has the same logic as yours. But at this point, it's both faster and more readable than the previous version so I think this is good enough, optimizing one lint to hell won't make or break clippy's performance :D

clippy_lints/src/raw_strings.rs
Centri3: That isn't true. It returned `num + 1` before.

Centri3: I see what you mean.

CHANGELOG.md
```diff
@@ -5385,4 +5387,7 @@ Released 2018-09-13
 [`allowed-idents-below-min-chars`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allowed-idents-below-min-chars
 [`min-ident-chars-threshold`]: https://doc.rust-lang.org/clippy/lint_configuration.html#min-ident-chars-threshold
 [`accept-comment-above-statement`]: https://doc.rust-lang.org/clippy/lint_configuration.html#accept-comment-above-statement
+[`allow-one-hash-in-raw-string`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allow-one-hash-in-raw-string
+<!-- end autogenerated links to configuration documentation -->
+_string_hashes` when it's unnecessary)
```
Alexendoo, line 5392: Looks unintended

Centri3: Not sure how that happened

CHANGELOG.md
```diff
@@ -5386,4 +5388,5 @@ Released 2018-09-13
 [`min-ident-chars-threshold`]: https://doc.rust-lang.org/clippy/lint_configuration.html#min-ident-chars-threshold
 [`accept-comment-above-statement`]: https://doc.rust-lang.org/clippy/lint_configuration.html#accept-comment-above-statement
 [`accept-comment-above-attributes`]: https://doc.rust-lang.org/clippy/lint_configuration.html#accept-comment-above-attributes
+[`allow-one-hash-in-raw-string`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allow-one-hash-in-raw-string
```
dswij, line 5391: ditto:

`allow-one-hash-in-raw-strings`

dswij (changes requested):
Just some small comments. I think it's good to merge afterwards

clippy_lints/src/utils/conf.rs
```diff
@@ -542,10 +542,14 @@ define_Conf! {
     ///
     /// Whether to accept a safety comment to be placed above the statement containing the `unsafe` block
     (accept_comment_above_statement: bool = false),
-        /// Lint: UNDOCUMENTED_UNSAFE_BLOCKS.
```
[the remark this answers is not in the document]
Centri3: idk how that keeps happening :sweat_smile:

dswij (approved):
Thank you for this!

Centri3:
dogfood failed, seems to pass now @dswij

Issue #5102, opened 2020-01-27
new lint: using `let mut a = a` at start of function


jyn514:
I inherited a codebase from someone new to Rust and there were lots of function definitions that looked like this:

```rust
fn f(a: Vec<u8>) {
    let mut a = a;
    // rest of code
}
```

It would be nice if there were a lint saying this would be better as `fn f(mut a: Vec<u8>)` instead.

sonng:
Looks like a good first issue. Could this be assigned to me please?

Thanks

sonng:
Hey guys,

I'm on the last leg of the work (`cargo test`) and making sure everything is fine. I noticed that while running that, our new lint gets triggered and causes an error, causing the whole thing to fail.

My question is, how do I turn this from an error to a warning?
or alternatively, I could update the problematic code and make them fn parameters mutable.

Cheers

flip1995:
Can you open a WIP PR, so I can have a look at the code and the dogfood error?

sonng:
I'm happy for any feedback on the overall change too.

brightly-salty:
I'm willing to attempt to implement this and submit a PR if it would be accepted if @sonng isn't working on it. I expect to restart (to avoid merge conflicts) using their PR as a guide/inspiration?

flip1995:
@brightly-salty Thanks! Yes you can use they're PR as inspiration. I think my comments on the PR should also help you with dos and don'ts.

max-niederman:
I think this should probably be generalized to redefined locals as well.

maxcnunes:
Is this issue still relevant? It seems https://github.com/rust-lang/rust-clippy/pull/10885 covered what have been requested here and we could close it.

PR #10885, opened 2023-06-04
new lint: `redundant_locals`

max-niederman:
This lint checks for code like the following:

```rs
let x = 1;
let x = x;
```

It checks (afaik) all cases where a binding is shadowed by its own value in the same block, including function parameters. This has no effect and is almost certainly accidental, so it's in the `correctness` category like `self_assignment`.

This also lays the groundwork for a more generalized version of #5102.  

changelog: new lint: [`redundant_local`]

### DIFF
--- a/clippy_lints/src/redundant_locals.rs
+++ b/clippy_lints/src/redundant_locals.rs
@@ -1,0 +1,103 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use clippy_utils::is_from_proc_macro;
+use clippy_utils::ty::needs_ordered_drop;
+use rustc_hir::def::Res;
+use rustc_hir::{BindingAnnotation, ByRef, Expr, ExprKind, HirId, Local, Node, Pat, PatKind, QPath};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::Ident;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for redundant redefinitions of local bindings.
+    ///
+    /// ### Why is this bad?
+    /// Redundant redefinitions of local bindings do not change behavior and are likely to be unintended.
+    ///
+    /// Note that although these bindings do not affect your code's meaning, they _may_ affect `rustc`'s stack allocation.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let a = 0;
+    /// let a = a;
+    ///
+    /// fn foo(b: i32) {
+    ///    let b = b;
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let a = 0;
+    /// // no redefinition with the same name
+    ///
+    /// fn foo(b: i32) {
+    ///   // no redefinition with the same name
+    /// }
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub REDUNDANT_LOCALS,
+    correctness,
+    "redundant redefinition of a local binding"
+}
+declare_lint_pass!(RedundantLocals => [REDUNDANT_LOCALS]);
+
+impl<'tcx> LateLintPass<'tcx> for RedundantLocals {
+    fn check_local(&mut self, cx: &LateContext<'tcx>, local: &'tcx Local<'tcx>) {
+        if_chain! {
+            // the pattern is a single by-value binding
+            if let PatKind::Binding(BindingAnnotation(ByRef::No, mutability), _, ident, None) = local.pat.kind;
+            // the binding is not type-ascribed
+            if local.ty.is_none();
+            // the expression is a resolved path
+            if let Some(expr) = local.init;
+            if let ExprKind::Path(qpath @ QPath::Resolved(None, path)) = expr.kind;
+            // the path is a single segment equal to the local's name
+            if let [last_segment] = path.segments;
+            if last_segment.ident == ident;
+            // resolve the path to its defining binding pattern
+            if let Res::Local(binding_id) = cx.qpath_res(&qpath, expr.hir_id);
+            if let Node::Pat(binding_pat) = cx.tcx.hir().get(binding_id);
+            // the previous binding has the same mutability
+            if find_binding(binding_pat, ident).unwrap().1 == mutability;
+            // the local does not affect the code's drop behavior
+            if !affects_drop_behavior(cx, binding_id, local.hir_id, expr);
+            // the local is user-controlled
+            if !in_external_macro(cx.sess(), local.span);
+            if !is_from_proc_macro(cx, expr);
+            then {
+                span_lint_and_help(
+                    cx,
+                    REDUNDANT_LOCALS,
+                    vec![binding_pat.span, local.span],
+                    "redundant redefinition of a binding",
+                    None,
+                    &format!("remove the redefinition of `{ident}`"),
+                );
+            }
+        }
+    }
+}
+
+/// Find the annotation of a binding introduced by a pattern, or `None` if it's not introduced.
+fn find_binding(pat: &Pat<'_>, name: Ident) -> Option<BindingAnnotation> {
+    let mut ret = None;
+
+    pat.each_binding_or_first(&mut |annotation, _, _, ident| {
+        if ident == name {
+            ret = Some(annotation);
+        }
+    });
+
+    ret
+}
+
+/// Check if a rebinding of a local affects the code's drop behavior.
+fn affects_drop_behavior<'tcx>(cx: &LateContext<'tcx>, bind: HirId, rebind: HirId, rebind_expr: &Expr<'tcx>) -> bool {
+    let hir = cx.tcx.hir();
+
+    // the rebinding is in a different scope than the original binding
+    // and the type of the binding cares about drop order
+    hir.get_enclosing_scope(bind) != hir.get_enclosing_scope(rebind)
+        && needs_ordered_drop(cx, cx.typeck_results().expr_ty(rebind_expr))
+}
--- a/tests/ui/redundant_locals.rs
+++ b/tests/ui/redundant_locals.rs
@@ -1,0 +1,111 @@
+//@aux-build:proc_macros.rs:proc-macro
+#![allow(unused, clippy::no_effect, clippy::needless_pass_by_ref_mut)]
+#![warn(clippy::redundant_locals)]
+
+extern crate proc_macros;
+use proc_macros::{external, with_span};
+
+fn main() {}
+
+fn immutable() {
+    let x = 1;
+    let x = x;
+}
+
+fn mutable() {
+    let mut x = 1;
+    let mut x = x;
+}
+
+fn upgraded_mutability() {
+    let x = 1;
+    let mut x = x;
+}
+
+fn downgraded_mutability() {
+    let mut x = 1;
+    let x = x;
+}
+
+fn coercion(par: &mut i32) {
+    let par: &i32 = par;
+
+    let x: &mut i32 = &mut 1;
+    let x: &i32 = x;
+}
+
+fn parameter(x: i32) {
+    let x = x;
+}
+
+fn many() {
+    let x = 1;
+    let x = x;
+    let x = x;
+    let x = x;
+    let x = x;
+}
+
+fn interleaved() {
+    let a = 1;
+    let b = 2;
+    let a = a;
+    let b = b;
+}
+
+fn block() {
+    {
+        let x = 1;
+        let x = x;
+    }
+}
+
+fn closure() {
+    || {
+        let x = 1;
+        let x = x;
+    };
+    |x: i32| {
+        let x = x;
+    };
+}
+
+fn consequential_drop_order() {
+    use std::sync::Mutex;
+
+    let mutex = Mutex::new(1);
+    let guard = mutex.lock().unwrap();
+
+    {
+        let guard = guard;
+    }
+}
+
+fn inconsequential_drop_order() {
+    let x = 1;
+
+    {
+        let x = x;
+    }
+}
+
+fn macros() {
+    macro_rules! rebind {
+        ($x:ident) => {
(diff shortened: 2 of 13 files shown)


max-niederman:
Okay, so I'm really confused on how the the `redundant_locals` lint is still being emitted for the `popular-crates` even with the `Span::from_expansion` check.

Centri3:
> Okay, so I'm really confused on how the the `redundant_locals` lint is still being emitted for the `popular-crates` test even with the `Span::from_expansion` check. Anybody know why this could happen?

This is likely because `clap` will itself redefine locals in its generated code (it does this in `Args::augment_args`, iirc), therefore you'll need to call `is_from_proc_macro` (which I think will require you to use a `LateLintPass` lint). afaik `Span::from_expansion` isn't enough to figure out if something's from a proc macro or not

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,136 @@
...
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::{symbol::Ident, Span};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for redundant redefinitions of local bindings.
+    ///
+    /// ### Why is this bad?
+    /// Redundant redefinitions of local bindings do nothing and are likely to be unintended.
```
Centri3, line 17: I think this could also mention that it may allocate a new local on the stack (*even in release mode*, at least for `Vec<T>`. [example](https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=ddcfaadc19ad7ce8497bcec03ec8d205))

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,136 @@
...
+                    definition_span: subpat.span,
+                },
+            );
+        }
+        true
+    });
+}
+
+// check if a local is actually controlled by the user
+// otherwise, (macro or compiler generated), we should not lint it
```
Centri3, line 130: nit: this should be a doc comment and "compiler generated" can be removed afaik, as compiler generated locals are not visited (or even available in the AST)

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        // collect the initial bindings from the function's parameters
+        let mut bindings = FxHashMap::default();
+        for param in &func.decl().inputs {
+            collect_introduced_bindings(&mut bindings, &param.pat);
+        }
+
+        check_block_with_bindings(cx, &mut bindings, block);
+    }
+}
```
Centri3, line 72: issue: this won't catch
```rust
fn parameter(x: i32) {
    let x = x;
    {
        let x = x;
    }
}
```
or
```rust
fn block() {
    let x = 1;
    {
        let x = x;
    }
}
```

dswij:
@max-niederman thanks for the PR!

We follow a no merge-commit policy. Can you help to resolve the conflict by rebasing this to `master` branch and drop the merge-commit?

@Centri3 Do you want to help review this PR since you've looked at it?

clippy_lints/src/redundant_local.rs
Centri3: you should also do this for the other functions here, `collect_introduced_bindings` and `check_block_with_bindings`

clippy_lints/src/redundant_local.rs
Centri3: You should check all stmts, not just `StmtKind::Local`, that way you also include any introduced `Block`s (as currently it will only check the fn's outermost block)

Centri3 (commented):
Looks pretty good so far, but it misses some redundant locals.

clippy_lints/src/redundant_local.rs
Centri3: I think this can be done by using `path_res`, then using the `HirId` from that to get that node; using `cx.tcx.hir().get(hir_id)` (maybe also requiring getting the parent node). Getting the mutability is the same as you did in `collect_introduced_bindings`. This should allow you to completely omit storing the bindings, and you can combine this with checking all stmts :) Though I haven't tested this yet so, so there may be some error somewhere..

Note this will require using a late lint.

llogiq:
Please add a test case with a type-ascribed local (e.g. `let x: u8 = x;`). Those should not be linted.

Alexendoo:
Similarly some test cases for `let ref x = x;` and `let ref mut x = x;` would be good

We might need to add a [`needs_ordered_drop`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/ty/fn.needs_ordered_drop.html) check as removing the `let` statement can change the drop order, e.g.

```rust
let guard = mutex.lock().unwrap();

// ..

{
    let guard = guard;
    // ..
}

// mutex is now unlocked
```

max-niederman:
@Centri3 @Alexendoo I intentionally wrote the lint to only check redefinitions in the same block in order to prevent changes to the drop order, but I definitely think it would be better to check all redefinitions and use `needs_ordered_drop` now that I know that's an option. I'm going to rewrite as a late-pass lint and do that.

max-niederman:
I'll also disable linting on type-ascribed locals. Totally forgot about coercion :facepalm:.

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,110 @@
...
+    correctness,
+    "redundant redefinition of a local binding"
+}
+declare_lint_pass!(RedundantLocal => [REDUNDANT_LOCAL]);
+
+impl<'tcx> LateLintPass<'tcx> for RedundantLocal {
+    fn check_local(&mut self, cx: &LateContext<'tcx>, local: &'tcx Local<'tcx>) {
+        if_chain! {
+            // the pattern is a single by-value binding
+            if let PatKind::Binding(BindingAnnotation(ByRef::No, mutability), _, ident, None) = local.pat.kind;
```
Centri3, line 46: This will intentionally miss code like
```rust
let (x, y) = (1, 2);
let (x, y) = (x, y);
```
You can use the aforementioned `pat.walk()` plus `for_each_expr` to get every binding introduced in `init`, but this will likely be a lot more complex and it's fine if it isn't done; just throwing it out there :)

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// ### Example
+    /// ```rust
+    /// let a = 0;
+    /// let a = a;
+    ///
+    /// fn foo(b: i32) {
+    ///    let b = b;
+    /// }
+    /// ```
+    /// Use instead:
```
Centri3, line 26: nit: I'm not sure whether `use instead` is necessary here. Some lints tend to omit it in some cases like these ("no definition with the same name")

Centri3 (commented):
Looks pretty good, I've left some comments on what can be improved.

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,110 @@
...
+                    None,
+                    &format!("remove the redefinition of `{ident}`"),
+                );
+            }
+        }
+    }
+}
+
+/// Find the annotation of a binding introduced by a pattern, or `None` if it's not introduced.
+fn find_binding(pat: &Pat<'_>, name: Ident) -> Option<BindingAnnotation> {
```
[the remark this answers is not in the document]
max-niederman: Yeah that was my first thought but I decided to go with this instead because mutating a value to be returned inside the closure felt kinda gross compared to the functional method. Would be nice if there was an iterator type to do outer iteration over subpatterns.

clippy_lints/src/redundant_local.rs
max-niederman: I have no idea whether it's consistent with other similar lints, but I think it's unlikely to hurt. Also, I think that, in this particular case, it's helpful because a significant fraction of Rust users probably aren't familiar with the term "binding" or "local", so having an example which makes it clear what exactly it means to have a "redundant redefinition of a binding" makes the lint more accessible.

Centri3: that's fine, was just pointing out that others omit it sometimes

clippy_lints/src/redundant_local.rs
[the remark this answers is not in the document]
max-niederman: I rewrote it using `Pat::each_binding_or_first` and I agree that it's way more readable this way.

max-niederman:
I finally got around to working on this again and I'm pretty sure it's in a good state. It would be good to check for multiple-binding local statements, but that's _way_ harder and definitely suffers from diminishing returns in terms of applicability.

Centri3 (approved):
This looks good, thanks! I may do the `(x, y)` idea as a followup if that's fine with you ^^

@dswij, can you take a look as well?

clippy_lints/src/redundant_local.rs
```diff
@@ -0,0 +1,100 @@
...
+            if let Some(expr) = local.init;
+            if let ExprKind::Path(qpath @ QPath::Resolved(None, path)) = expr.kind;
+            // the path is a single segment equal to the local's name
+            if path.segments.len() == 1;
+            if path.segments[0].ident == ident;
+            // resolve the path to its defining binding pattern
+            if let Res::Local(binding_id) = cx.qpath_res(&qpath, expr.hir_id);
+            if let Node::Pat(binding_pat) = cx.tcx.hir().get(binding_id);
+            // the previous binding has the same mutability
+            if find_binding(binding_pat, ident).unwrap().1 == mutability;
```
dswij, line 59: How are we sure that `find_binding(...)` here will always return `Some(...)`?

tests/ui/redundant_local.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+fn inconsequential_drop_order() {
+    let x = 1;
+
+    {
+        let x = x;
+    }
+}
```
dswij, line 86: Can we also add some test for macros? i.e. macro definitions and external macros. Usually those are.. tricky

clippy_lints/src/redundant_local.rs
max-niederman: `binding_pat` is the pattern of the local which first defined the identifier, so it must bind to that identifier.

tests/ui/redundant_local.rs
max-niederman: I added a test for declarative macros and proc macros like `clap`'s which expand to redundant locals are tested by `lintcheck`.

clippy_lints/src/redundant_locals.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// let a = 0;
+    /// // no redefinition with the same name
+    ///
+    /// fn foo(b: i32) {
+    ///   // no redefinition with the same name
+    /// }
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub REDUNDANT_LOCALS,
+    correctness,
```
flip1995, line 37: Is this really a `correctness` lint? What's the danger in doing this? deny-by-default for this lint seems a bit strict to me.

I would rather put this lint into `complexity`. Maybe `suspicious`.

max-niederman: I put it in the correctness category because `self_assignment`, which is almost the same thing, is also in correctness.

Centri3: `self_assignment` is a lot worse (it causes logic bugs) but yeah, I think this should be `correctness` because it's never what the programmer intended

flip1995: Thanks for running `lintcheck`. In that case, I think it can keep `correctness`. I guess "code that is outright useless" fits for this lint.

tests/ui/redundant_locals.rs
```diff
@@ -0,0 +1,97 @@
...
+
+fn inconsequential_drop_order() {
+    let x = 1;
+
+    {
+        let x = x;
+    }
+}
+
+fn macros() {
```
Centri3, line 88: I think tests for external/procedural macros are worthwhile. Add `//@aux-build:proc_macros.rs` to the top of the file, then `extern crate proc_macros`. This will let you use the macros in `tests/ui/auxiliary/proc_macros.rs`.

Something like
```rust
external! {
	let x = 1;
	let x = x;
}
// This will set the token's spans to the end of `span`, thus testing that `is_from_proc_macro` works
with_span! {
	span
	let x = 1;
	let x = x;
}
```

clippy_lints/src/redundant_locals.rs
```diff
@@ -0,0 +1,100 @@
...
+        if_chain! {
+            // the pattern is a single by-value binding
+            if let PatKind::Binding(BindingAnnotation(ByRef::No, mutability), _, ident, None) = local.pat.kind;
+            // the binding is not type-ascribed
+            if local.ty.is_none();
+            // the expression is a resolved path
+            if let Some(expr) = local.init;
+            if let ExprKind::Path(qpath @ QPath::Resolved(None, path)) = expr.kind;
+            // the path is a single segment equal to the local's name
+            if path.segments.len() == 1;
```
Centri3, line 53: We can use `if let [last_ident] = path.segments` followed by `last_ident == ident` (you may need `&*` to coerce it to another type)

Centri3 (approved):
I think this is about ready to be merged :D Just two minor nits, thanks!

tests/ui/redundant_locals.rs
```diff
@@ -0,0 +1,111 @@
+//@aux-build:proc_macros.rs
```
Centri3, line 1:
```suggestion
//@aux-build:proc_macros.rs:proc-macro
```
This needs the crate type now.

Centri3:
Once you rebase and add the crate type this is ready :)

tests/ui/redundant_locals.rs
max-niederman: Thanks, I had more or less given up.

Centri3:
Sorry, also please squash.

Centri3:
Thanks! @bors r+

Issue #10832, opened 2023-05-26
`undocumented_unsafe_blocks` false positive when unsafe block is on a separate line from a `let`


DropDemBits:
### Summary

When a single-line unsafe block is long enough to be formatted on a new line,
it appears that the safety comment isn't found.

### Lint Name

undocumented_unsafe_blocks

### Reproducer

I tried this code:

```rust
#![deny(clippy::undocumented_unsafe_blocks)]

fn main() {
    let aaaaaa = 1;
    // SAFETY: A safety comment
    let _some_var_name =
        unsafe { path::to::aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(aaaaaa, A::CONTEXT_INFO) };
}

pub(crate) mod path {
    pub(crate) mod to {
        pub(crate) unsafe fn aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(_: u32, _: ()) -> u32{
            1
        }
    }
}

struct A {}

impl A {
    const CONTEXT_INFO: () = ();
}
```

I saw this happen:

```
error: unsafe block missing a safety comment
 --> src/main.rs:7:9
  |
7 |         unsafe { path::to::aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(aaaaaa, A::CONTEXT_INFO) };
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: consider adding a safety comment on the preceding line
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks
note: the lint level is defined here
 --> src/main.rs:1:9
  |
1 | #![deny(clippy::undocumented_unsafe_blocks)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I expected to see this happen:

Lint shouldn't fire

### Version

```text
rustc 1.71.0-nightly (1c42cb4ef 2023-04-26)
binary: rustc
commit-hash: 1c42cb4ef0544fbfaa500216e53382d6b079c001
commit-date: 2023-04-26
host: x86_64-pc-windows-msvc
release: 1.71.0-nightly
LLVM version: 16.0.2
```

lochetti:
here it goes a little bit smaller reproducer

```
[...]

PR #10886, opened 2023-06-04
Adding configuration to allow safety comment above stmt containing unsafe block

lochetti:
Adding a new configuration, `accept-comment-above-statement`, to allow a safety comment to be placed before the statement that has the `unsafe` block. It affects the `undocumented_unsafe_blocks` lint.

The default value for this configuration will be `false`. So the user has to opt-in for the change.

This PR fixes https://github.com/rust-lang/rust-clippy/issues/10832

---

changelog: Enhancement [`undocumented_unsafe_blocks`]: Added `accept-comment-above-statement` configuration.
[#10886](https://github.com/rust-lang/rust-clippy/pull/10886)


xFrednet:
Hey @Centri3, this might be an interesting and simple PR to review, if you're interested.

I can assign you to the PR, after you left a comment. That makes it easier to keep track of PRs :).

r? @xFrednet

Centri3:
Alright, I'll review it in a bit :D

clippy_lints/src/utils/conf.rs
```diff
@@ -514,6 +514,10 @@ define_Conf! {
     ///
     /// The byte size a `T` in `Box<T>` can have, below which it triggers the `clippy::unnecessary_box` lint
     (unnecessary_box_size: u64 = 128),
+    /// Lint: UNDOCUMENTED_UNSAFE_BLOCKS.
+    ///
+    /// Whether to accept a safety comment to be placed above the statement containing the `usafe` block
```
Centri3, line 519:
```suggestion
    /// Whether to accept a safety comment to be placed above the statement containing the `unsafe` block
```

tests/ui-toml/undocumented_unsafe_blocks/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,12 @@
+#![deny(clippy::undocumented_unsafe_blocks)]
+#![allow(clippy::missing_safety_doc)]
+
+fn main() {
+    // Safety: A safety comment
+    let _some_variable_with_a_very_long_name_to_break_the_line =
+        unsafe { a_function_with_a_very_long_name_to_break_the_line() };
```
Centri3, line 7: I think this could benefit from an extra test or two; namely, making sure static/consts work as expected as well

Centri3 (commented):
The implementation looks pretty good overall! There's just a minor typo in `accept_comment_above_statement`'s doc comment.

lochetti:
> The implementation looks pretty good overall! There's just a minor typo in `accept_comment_above_statement`'s doc comment.

Hey @Centri3 thank you very much for the review. I will address your comments in the next couple days and I let you know :)

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -546,7 +580,7 @@ fn get_body_search_span(cx: &LateContext<'_>) -> Option<Span> {
     for (_, node) in map.parent_iter(body.hir_id) {
         match node {
             Node::Expr(e) => span = e.span,
-            Node::Block(_) | Node::Arm(_) | Node::Stmt(_) | Node::Local(_) => (),
+            Node::Block(_) | Node::Arm(_) | Node::Stmt(_) | Node::Local(_) | Node::Item(_) => (),
```
lochetti, line 583: I am not 100% sure that this change would not lead to undesirable side effects. This was necessary to make it work for the const/static cases.
Wdyt?

clippy_lints/src/utils/conf.rs
lochetti: Fixed! https://github.com/rust-lang/rust-clippy/pull/10886/commits/d01f49aad11eb505bbd9bb34024cc5e46b875350

tests/ui-toml/undocumented_unsafe_blocks/undocumented_unsafe_blocks.rs
lochetti: Done! https://github.com/rust-lang/rust-clippy/pull/10886/commits/d01f49aad11eb505bbd9bb34024cc5e46b875350

Thanks for sugesting this! In fact, the result was not what we expected for static/const. I had to make another change, as well, for those two cases.

clippy_lints/src/undocumented_unsafe_blocks.rs
Centri3: a `Node::Item` is just an item, i.e., `fn`, `const`, `mod`, etc, so while I don't think that would cause any issues you can probably check that `kind` is either `Static` or `Const`

lochetti: Ok! I have done that and ammend to the previous commit. https://github.com/rust-lang/rust-clippy/pull/10886/commits/e9562675ed0990fe54ffd96fbc2237de803bbc74

tests/ui-toml/undocumented_unsafe_blocks/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,24 @@
...
+        unsafe { a_const_function_with_a_very_long_name_to_break_the_line() };
+}
+
+pub unsafe fn a_function_with_a_very_long_name_to_break_the_line() -> u32 {
+    1
+}
+
+pub const unsafe fn a_const_function_with_a_very_long_name_to_break_the_line() -> u32 {
+    2
+}
```
Centri3, line 24: I think, just in case, this file should be the same as `tests/ui/undocumented_unsafe_blocks.rs` (minus the safety comments). This way if something else breaks with this configuration option we can see.

lochetti: Ok! I understand the idea of having all the same tests with this configuration and this makes sense for me. But I could not understand why `minus the safety comments`. The configuration will affect just a small subset of all tests. So, in summary, we will have almost the same .stderr file. The only difference would be those cases that we already have here at `tests/ui-toml/toml_unknown_key/conf_unknown_key.stderr` that don't lint. But we would have to have the safety comments in all test cases to test is properly, don't we?

Maybe I am just not undestanding your idea.

Centri3: > minus the safety comments

Sorry, I meant that in regard to `should be the same`; For the ones affected by the configuration option, they would be of course different :D

lochetti: Ok, great! I will do that

lochetti: Done: https://github.com/rust-lang/rust-clippy/pull/10886/commits/928c712c0d3a7b9766f38046a573295c11767933
Two interesting things:
- I had to duplicate code and create a new `auxiliary/proc_macro_unsafe.rs` file under the test folder. I don't know if there is another way to use auxiliary functions in ui-toml tests without this duplication
- And there was another test case that don't fire with this configuration. It is the unsafe block at line 498. And it makes sense to don't lint this case with the configuration. It was cool to see that but I was not expecting it :)

lochetti:
@Centri3 I have just rebased from master. Let me know if you have any comments about the current state of the PR :)

Centri3 (approved):
This looks good to me, thanks!

cc @xFrednet

tests/ui-toml/undocumented_unsafe_blocks/undocumented_unsafe_blocks.rs
```diff
@@ -0,0 +1,534 @@
+//@aux-build:proc_macro_unsafe.rs
```
xFrednet, line 1: A small site note @Centri3, AFAIK you requested that this file is the same as the normal test file. Generally speaking, I think it's fine to have some basic tests for the lint, some *don't lint* examples and then the specific case, which is influenced by the config value. Having a lot of output in the `.stderr` file can make it hard to distinguish the actual changes.

This is not to imply that either variant is better. This is up to the reviewer to decide and since you mainly reviewed this one, I'll follow your lead. Just so you know, it's not mandatory :)

xFrednet (approved):
This version LGTM! I believe, you'll need to rebased on master and run `cargo collect-metadata` again due to some recent changes to that script. Then it should be ready to be merged :D

xFrednet:
@Centri3 I'll delegate the PR to you, you can `r=Centri3,xFrednet` after the rebase :D.

@bors delegate=@Centri3

lochetti:
Hey @Centri3 and @xFrednet. I have rebased and have run `cargo collect-metadata`. The last commit is the result of the `cargo collect-metada` command. Let me know if anything else is required :)

Centri3:
Thank you!
(omg I get to merge)

@bors r=Centri3,xFrednet

PR #10889, opened 2023-06-04
Add configuration values auto markdown links to `CHANGELOG.md`

blyxyas:
Related to [this Zulip discussion](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Revisiting.20.20Clippy's.20configuration.20discover/near/363119950).

Now `cargo dev update_lints` also includes configurations as Markdown links, so changes to configuration can be linked directly.
It links to [this file](https://github.com/rust-lang/rust-clippy/blob/master/book/src/lint_configuration.md) (`book/src/lint_configuration.md`) because a new changelog may be deployed before the website is updated. So, in order to avoid broken links, this is the simplest way.

changelog:none


xFrednet:
Hey awesome, that you already made a PR for this!! I believe `cargo dev update_lints` or however the comment is again, would replace everything inside the two doc comments. Could you test that the values remain even after the update? :upside_down_face:

xFrednet:
The implementation looks good to me, but I'm not sure, if we want to add `clippy_lints` as a dependency to `cargo_dev` @flip1995, do you have any thoughts?

flip1995:
I don't really like depending on `clippy_*` crates from `clippy_dev`. This feels a little cyclic. 

Also linking to source code instead of documentation feels wrong. 

It also doesn't make things easier that the stable book is usually updated after the changelog is written.

It doesn't happen that often that a new configuration is added. Couldn't we add the link to it manually, if this happens? `linkcheck` should validate that we only add links that actually exist.

xFrednet:
> It also doesn't make things easier that the stable book is usually updated after the changelog is written.

Is to book updated, at the same time that the new version is released? I'm guessing that most people only check the changelog after a release. This would most likely be fine.

> It doesn't happen that often that a new configuration is added. Couldn't we add the link to it manually, if this happens?

I mean we could, but it's very likely that the list will get outdated fast. Since we already collect the data I think it would be better to automatically add the lints. Maybe we can do it as part of `cargo collect-metadata` since that command has to be run after adding a new configuration anyways

flip1995:
> Maybe we can do it as part of `cargo collect-metadata` since that command has to be run after adding a new configuration anyways

Good idea, let's update the changelog from `cargo collect-metadata` then. I would still link to the book, even if the config is not there yet. At least, if `linkcheck` allows us to do so.



> Is to book updated, at the same time that the new version is released?

IIUC, yes. The Clippy book rides the same release train as normal Clippy.

flip1995:
FWIW, there is also a nightly/beta version of the book: https://doc.rust-lang.org/nightly/clippy/ But it would still have the same link problems as to stable, since the Clippy book is deployed from the Rust repo, not the Clippy repo.

xFrednet:
Nice, I was looking for something like that earlier. It should be fine linking to that one. We also link to the master version of our lint list.

> But it would still have the same link problems as the stable version

Which problem do you mean? At the point when a changelog entry is added, it's already on rustc's master and the nightly book will also have been updated :thinking:

flip1995:
> Which problem do you mean?

We run a link checker in CI, that verifies links in the book, so that we don't link to a 404 page. I'm not sure, if the link checker cares about the `#<section>` part. If it does, we would only be able to add those links, once the config documentation is deployed in the book. If not (what I hope for) we don't have any problem.

xFrednet:
The `#<section>` directive is mainly for browsers, to directly jump to a location. I would strongly guess that it would work perfectly. At the same time, I'm not really web developer ^^

flip1995:
> It should be fine linking to that one

I don't think we have to link to the nightly one in the changelog. The changelog gets written for the stable version. By the time people click on the changelog (through the rustc release notes) the stable book will already have the stable config values of the changelog in it.

blyxyas:
*Sooo...* Is this PR useless?

blyxyas:
?

blyxyas:
Could you re-review it? Now it works with `cargo collect-metadata`

CHANGELOG.md
```diff
@@ -5318,3 +5318,58 @@ Released 2018-09-13
 [`zero_width_space`]: https://rust-lang.github.io/rust-clippy/master/index.html#zero_width_space
 [`zst_offset`]: https://rust-lang.github.io/rust-clippy/master/index.html#zst_offset
 <!-- end autogenerated links to lint list -->
+<!-- Start configuration's links to the book, DO NOT EDIT MANUALLY -->
```
flip1995, line 5321: This is a big NIT, but we usually word this as

```
<!-- begin autogenerated links to configuration documentation -->
```

flip1995 (approved):
LGTM. Only one comment. Please excuse my OCD. :sweat_smile:

clippy_lints/src/utils/internal_lints/metadata_collector.rs
```diff
@@ -260,6 +262,22 @@ Please use that command to update the file and do not edit it by hand.
...
+        let mut changelog_file = OpenOptions::new().read(true).write(true).open(CHANGELOG_PATH).unwrap();
+
+        if let Some(position) = changelog.find("<!-- begin autogenerated links to configuration documentation -->") {
+            // I know this is kinda wasteful, we just don't have regex on `clippy_lints` so... this is the best
+            // we can do AFAIK.
+            changelog = changelog[..position].to_string();
+        }
+        writeln!(
+            changelog_file,
+            "{changelog}<!-- begin autogenerated links to configuration documentation -->\n{}\n<!-- end autogenerated links to configuration documentation -->\n",
```
flip1995, line 277:
```suggestion
            "{changelog}<!-- begin autogenerated links to configuration documentation -->\n{}\n<!-- end autogenerated links to configuration documentation -->",
```

Now remove the extra newline and this is good to go. :shipit:

flip1995:
r=me once my comment is addressed. (You might want to squash the commits)

@bors delegate+

xFrednet:
This new extra comment from bors, about `r=me` is awesome! Good to know :D

blyxyas:
Bors is truly AI blockchain-based machine learning technology

flip1995:
> run `cargo collect-metadata` if this fails

How ironic. Please rebase on current master and run that and then this should merge.

PR #10891, opened 2023-06-05
[`missing_const_for_fn`]: Ensure dropped locals are `~const Destruct`

### LINKED ISSUE #10617: [`missing_const_for_fn`] does not taike into account whether receiver is `~const Destruct`
### Summary

`missing_const_for_fn` does not taike into account whether receiver is `~const Destruct`.

### Lint Name

missing_const_for_fn

### Reproducer

I tried this code:

```rust
#![warn(clippy::missing_const_for_fn)]

pub struct Identifier(String);

impl Identifier {
    #[must_use = "If you don't use it, it will drop entire String"]
    pub fn name(self) -> String {
        self.0
    }
}
```

I saw this happen:

```
warning: this could be a `const fn`
 --> src/lib.rs:7:5
  |
7 | /     pub fn name(self) -> String {
8 | |         self.0
9 | |     }
  | |_____^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn
```

I expected to the lint does not fire.

### Version

```text
rustc 1.70.0-nightly (1459b3128 2023-03-23)
binary: rustc
commit-hash: 1459b3128e288a85fcc4dd1fee7ada2cdcf28794
commit-date: 2023-03-23
release: 1.70.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

Centri3:
this will check every local for `TerminatorKind::Drop` to ensure they can be evaluated at compile time, not sure if this is the best way to do this but MIR is confusing and it works so...

fixes #10617

changelog: [`missing_const_for_fn`]: Ensure dropped locals are `~const Destruct`

### DIFF
--- a/clippy_utils/src/qualify_min_const_fn.rs
+++ b/clippy_utils/src/qualify_min_const_fn.rs
@@ -4,7 +4,7 @@
 // differ from the time of `rustc` even if the name stays the same.
 
 use crate::msrvs::Msrv;
-use hir::{Constness, LangItem};
+use hir::LangItem;
 use rustc_const_eval::transform::check_consts::ConstCx;
 use rustc_hir as hir;
 use rustc_hir::def_id::DefId;
@@ -14,14 +14,14 @@ use rustc_middle::mir::{
     Body, CastKind, NonDivergingIntrinsic, NullOp, Operand, Place, ProjectionElem, Rvalue, Statement, StatementKind,
     Terminator, TerminatorKind,
 };
-use rustc_middle::traits::ObligationCause;
+use rustc_middle::traits::{ImplSource, ObligationCause};
 use rustc_middle::ty::subst::GenericArgKind;
 use rustc_middle::ty::{self, adjustment::PointerCast, Ty, TyCtxt};
 use rustc_middle::ty::{BoundConstness, TraitRef};
 use rustc_semver::RustcVersion;
 use rustc_span::symbol::sym;
 use rustc_span::Span;
-use rustc_trait_selection::traits::SelectionContext;
+use rustc_trait_selection::traits::{ObligationCtxt, SelectionContext};
 use std::borrow::Cow;
 
 type McfResult = Result<(), (Span, Cow<'static, str>)>;
@@ -135,9 +135,9 @@ fn check_rvalue<'tcx>(
     match rvalue {
         Rvalue::ThreadLocalRef(_) => Err((span, "cannot access thread local storage in const fn".into())),
         Rvalue::Len(place) | Rvalue::Discriminant(place) | Rvalue::Ref(_, _, place) | Rvalue::AddressOf(_, place) => {
-            check_place(tcx, *place, span, body)
+            check_place(tcx, *place, span, body, false)
         },
-        Rvalue::CopyForDeref(place) => check_place(tcx, *place, span, body),
+        Rvalue::CopyForDeref(place) => check_place(tcx, *place, span, body, false),
         Rvalue::Repeat(operand, _)
         | Rvalue::Use(operand)
         | Rvalue::Cast(
@@ -230,14 +230,14 @@ fn check_statement<'tcx>(
     let span = statement.source_info.span;
     match &statement.kind {
         StatementKind::Assign(box (place, rval)) => {
-            check_place(tcx, *place, span, body)?;
+            check_place(tcx, *place, span, body, false)?;
             check_rvalue(tcx, body, def_id, rval, span)
         },
 
-        StatementKind::FakeRead(box (_, place)) => check_place(tcx, *place, span, body),
+        StatementKind::FakeRead(box (_, place)) => check_place(tcx, *place, span, body, false),
         // just an assignment
         StatementKind::SetDiscriminant { place, .. } | StatementKind::Deinit(place) => {
-            check_place(tcx, **place, span, body)
+            check_place(tcx, **place, span, body, false)
         },
 
         StatementKind::Intrinsic(box NonDivergingIntrinsic::Assume(op)) => check_operand(tcx, op, span, body),
@@ -263,7 +263,8 @@ fn check_statement<'tcx>(
 
 fn check_operand<'tcx>(tcx: TyCtxt<'tcx>, operand: &Operand<'tcx>, span: Span, body: &Body<'tcx>) -> McfResult {
     match operand {
-        Operand::Move(place) | Operand::Copy(place) => check_place(tcx, *place, span, body),
+        Operand::Move(place) => check_place(tcx, *place, span, body, true),
+        Operand::Copy(place) => check_place(tcx, *place, span, body, false),
         Operand::Constant(c) => match c.check_static_ptr(tcx) {
             Some(_) => Err((span, "cannot access `static` items in const fn".into())),
             None => Ok(()),
@@ -271,13 +272,20 @@ fn check_operand<'tcx>(tcx: TyCtxt<'tcx>, operand: &Operand<'tcx>, span: Span, b
     }
 }
 
-fn check_place<'tcx>(tcx: TyCtxt<'tcx>, place: Place<'tcx>, span: Span, body: &Body<'tcx>) -> McfResult {
+fn check_place<'tcx>(tcx: TyCtxt<'tcx>, place: Place<'tcx>, span: Span, body: &Body<'tcx>, in_move: bool) -> McfResult {
     let mut cursor = place.projection.as_ref();
 
     while let [ref proj_base @ .., elem] = *cursor {
         cursor = proj_base;
         match elem {
-            ProjectionElem::Field(..) => {
+            ProjectionElem::Field(_, ty) => {
+                if !is_ty_const_destruct(tcx, ty, body) && in_move {
+                    return Err((
+                        span,
+                        "cannot drop locals with a non constant destructor in const fn".into(),
+                    ));
+                }
+
                 let base_ty = Place::ty_from(place.local, proj_base, body, tcx).ty;
                 if let Some(def) = base_ty.ty_adt_def() {
                     // No union field accesses in `const fn`
@@ -320,7 +328,7 @@ fn check_terminator<'tcx>(
                     "cannot drop locals with a non constant destructor in const fn".into(),
                 ));
             }
-            check_place(tcx, *place, span, body)
+            check_place(tcx, *place, span, body, false)
         },
         TerminatorKind::SwitchInt { discr, targets: _ } => check_operand(tcx, discr, span, body),
         TerminatorKind::GeneratorDrop | TerminatorKind::Yield { .. } => {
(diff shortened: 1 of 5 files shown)


Alexendoo:
I have no experience on the MIR side of things, r? @Jarcho who I _think_ does

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -80,6 +81,27 @@ fn check_ty<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, span: Span) -> McfResult {
             GenericArgKind::Lifetime(_) | GenericArgKind::Const(_) => continue,
         };
 
+        // Only do this check if we're in `TerminatorKind::Drop`, otherwise rustc will sometimes overflow
+        // its stack. This check is unnecessary outside of a `Drop` anyway so it's faster regardless
+        if in_drop && let ty::Adt(def, subst) = ty.kind() {
+            if def.has_non_const_dtor(tcx) && in_drop {
```
Jarcho, line 87: Is there really a need to walk all the field types after this check?

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -306,7 +328,12 @@ fn check_terminator<'tcx>(
         | TerminatorKind::Terminate
         | TerminatorKind::Unreachable => Ok(()),
 
-        TerminatorKind::Drop { place, .. } => check_place(tcx, *place, span, body),
+        TerminatorKind::Drop { place, .. } => {
+            for local in &body.local_decls {
```
Jarcho, line 332: This should only be checking the type of the place. There's nothing else being dropped at this point.

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -306,7 +328,12 @@ fn check_terminator<'tcx>(
...
+                check_ty(tcx, local.ty, span, true)?;
+            }
+            check_place(tcx, *place, span, body)
```
Jarcho, line 335: This shouldn't be necessary. The compiler won't drop union fields.

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -70,7 +71,7 @@ pub fn is_min_const_fn<'tcx>(tcx: TyCtxt<'tcx>, body: &Body<'tcx>, msrv: &Msrv)
     Ok(())
 }
 
-fn check_ty<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, span: Span) -> McfResult {
+fn check_ty<'tcx>(tcx: TyCtxt<'tcx>, ty: Ty<'tcx>, span: Span, in_drop: bool) -> McfResult {
```
Jarcho, line 74: This is really two functions. One to check that the parameters are valid types for const functions. The other is checking the constness of dropping the type.

clippy_utils/src/qualify_min_const_fn.rs
Centri3: AFAIK yes. from my testing it wouldn't catch it's not `~const Destruct` without it, but I can test again.

Jarcho: That's true. It doesn't take `ParamEnv`.

Jarcho: It would be better to use the trait solver here rather than walking the type. I would make a new util function for it. You can use [rustc's impl](https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_const_eval/transform/check_consts/qualifs.rs.html#150) as a reference.

Centri3: > It would be better to use the trait solver here rather than walking the type.

I'm trying to do this but it seems that it returns true anyway, which is... confusing, since it checks `Vec` and all but still thinks it has `~const Destruct` implemented...

Centri3: It seems `Vec<u16>` returns `Err(Unimplemented)` so I'll likely still need to walk the type... Am I missing something?

Centri3: Well, I just committed the current implementation and it's slow, really slow. And it still needs to walk all fields so I don't think it should do this

Jarcho: You didn't use any of the reference function starting at line 173 and that looks like it's filtering out unusable candidate selections.

Centri3: I actually did before; I removed it because it didn't change the result at all

Jarcho: That is the case. It wouldn't be too bad to do so now as const impls are still unstable.

Now that I'm looking at it, the referenced issue doesn't have a `Drop` terminator in the MIR at all. Which makes sense as the only contained field was moved.

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -398,3 +407,33 @@ fn is_const_fn(tcx: TyCtxt<'_>, def_id: DefId, msrv: &Msrv) -> bool {
...
+            && def.variants().iter()
+                .map(|variant| variant.fields.iter().map(|field| is_ty_const_destruct(tcx, field.ty(tcx, subst), body)))
+                .all(|mut fs| fs.all(|f| f))
+    } else {
+        true
+    };
+
+    let infcx = tcx.infer_ctxt().build();
+    let mut selcx = SelectionContext::new(&infcx);
+    selcx.select(&obligation).is_ok() && fields_all_const_destruct
```
Jarcho, line 438: This call is wrong. Should be `selcx.select(...).ok().flatten().is_some()`. Note that select returns `Result<Option<T>, ...>`

clippy_utils/src/qualify_min_const_fn.rs
Centri3: I'm still trying to fix this but it seems like this won't work (for both `Destruct` and also `Drop`), as it seems nothing checks whether its fields implement `~const Destruct` or not; just the containing type, therefore no matter what it needs to walk it.

If that's the case, I think the original implementation should be used since it's a lot faster...

Centri3: It's worth noting that using the trait solver works if the containing type doesn't implement `~const Destruct`, just not if only the fields don't

Centri3: Wow nevermind, I got it working... I have no clue why `ocx.select_all_or_error()` didn't work before but it's working so I'll take it? That's so bizarre

Centri3:
Ok, I got it working... Sorry for wasting your time with the trait solver shebackle :sweat_smile:

clippy_utils/src/qualify_min_const_fn.rs
Centri3: Fixed

clippy_utils/src/qualify_min_const_fn.rs
Centri3: Don't think this applies anymore

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -256,20 +263,29 @@ fn check_statement<'tcx>(
...
+        Operand::Copy(place) => check_place(tcx, *place, span, body, false),
         Operand::Constant(c) => match c.check_static_ptr(tcx) {
             Some(_) => Err((span, "cannot access `static` items in const fn".into())),
             None => Ok(()),
         },
     }
 }
 
-fn check_place<'tcx>(tcx: TyCtxt<'tcx>, place: Place<'tcx>, span: Span, body: &Body<'tcx>) -> McfResult {
+fn check_place<'tcx>(tcx: TyCtxt<'tcx>, place: Place<'tcx>, span: Span, body: &Body<'tcx>, in_move: bool) -> McfResult {
```
Centri3, line 275: Note that `in_move` here seems to be necessary here, as otherwise, something like `fn name(&self) -> &String { ... }` will not be linted

clippy_utils/src/qualify_min_const_fn.rs
Jarcho: The `check_place` call is still unnecessary.

clippy_utils/src/qualify_min_const_fn.rs
Jarcho: For moves you'll only need to check if the base of the projection has a non-const `Drop`. Only locals, or fields of a local, can be dropped and if a type is const droppable then all of it's fields will be as well.

clippy_utils/src/qualify_min_const_fn.rs
Centri3: Ah, thought you meant checking the locals; I see. That was already there before but I see what you mean where it won't drop union fields (as that's the only place it'll return an error).

clippy_utils/src/qualify_min_const_fn.rs
Centri3: I think it does this now

Jarcho: The only part you need to check for a move is the base of the projection. There's no need to check the type of each projection for const `Drop`.

Centri3: I forgot to remove that one :sweat_smile: Fixed, it only checks the base now

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -256,16 +263,29 @@ fn check_statement<'tcx>(
...
     let mut cursor = place.projection.as_ref();
+
+    if let [proj_base] = cursor
+        && let ProjectionElem::Field(_, ty) = proj_base
+        && !is_ty_const_destruct(tcx, *ty, body)
+        && in_move
```
Jarcho, line 281: This whole part can be inlined to where `Operand::Move` is handled since removing the need for the `is_move` flag.

Everything else looks ok.

clippy_utils/src/qualify_min_const_fn.rs
```diff
@@ -256,7 +263,19 @@ fn check_statement<'tcx>(
 
 fn check_operand<'tcx>(tcx: TyCtxt<'tcx>, operand: &Operand<'tcx>, span: Span, body: &Body<'tcx>) -> McfResult {
     match operand {
-        Operand::Move(place) | Operand::Copy(place) => check_place(tcx, *place, span, body),
+        Operand::Move(place) => {
+            if let [proj_base] = place.projection.as_ref()
```
Jarcho, line 267: This need to handle any number of field projections as long as there is at least one (The `[_, ..]` or `[.., _]` patterns).

I was slightly wrong on an earlier comment. You don't need to check the base type of the projection specifically, only the type of the field being moved.

Centri3: Would that be done by getting the type of `place` or by using `Field(.., ty)` like it does now?

Jarcho: The type of the place is what you want so long as there is a non-empty projection.

Jarcho:
Thank you. @bors r+

PR #10894, opened 2023-06-05
[`type_repetition_in_bounds`]: Don't lint on derived code

### LINKED ISSUE #10504: clippy::type_repetition_in_bounds triggers on derived code
### Summary

clippy::type_repetition_in_bounds triggers on code generated by Serde derive when it clashes with bounds on the type itself.

### Lint Name

type_repetition_in_bounds

### Reproducer

I tried this code:

```rust
#[derive(Debug, Serialize, Deserialize)]
#[serde(bound(
    serialize = "ParamStr: Serialize + Hash + Eq",
    deserialize = "Box<ParamStr>: serde::de::DeserializeOwned + Hash + Eq"
))]
pub(crate) struct OpaqueParams<ParamStr: ?Sized + Debug>(
    pub(crate) IndexMap<Box<ParamStr>, String>,
);
```

I saw this happen:

```
warning: this type has already been used as a bound predicate
   --> [...]/src/params.rs:135:17
    |
135 |     serialize = "ParamStr: Serialize + Hash + Eq",
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: consider combining the bounds: `"ParamStr: Serialize + Hash + Eq": ?Sized + Debug + "ParamStr: Serialize + Hash + Eq" + "ParamStr: Serialize + Hash + Eq" + "ParamStr: Serialize + Hash + Eq"`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#type_repetition_in_bounds
```

No error because this bound is used by derived code.

### Version

```text
rustc 1.68.0 (2c8cc3432 2023-03-06)
binary: rustc
commit-hash: 2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74
commit-date: 2023-03-06
release: 1.68.0
LLVM version: 15.0.6
```

### Additional Labels

@rustbot label +T-macros

Centri3:
fixes #10504.

changelog: [`type_repetition_in_bounds`]: Don't lint on derived code

### DIFF
--- a/clippy_lints/src/trait_bounds.rs
+++ b/clippy_lints/src/trait_bounds.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::{span_lint_and_help, span_lint_and_sugg};
 use clippy_utils::source::{snippet, snippet_opt, snippet_with_applicability};
-use clippy_utils::{SpanlessEq, SpanlessHash};
+use clippy_utils::{is_from_proc_macro, SpanlessEq, SpanlessHash};
 use core::hash::{Hash, Hasher};
 use if_chain::if_chain;
 use itertools::Itertools;
@@ -260,10 +260,7 @@ impl TraitBounds {
                     SpanlessTy { ty: p.bounded_ty, cx },
                     p.bounds.iter().collect::<Vec<_>>()
                 );
-                let bounded_ty = snippet(cx, p.bounded_ty.span, "_");
-                if let TyKind::Path(qpath) = p.bounded_ty.kind;
-                if format!("{}:", rustc_hir_pretty::qpath_to_string(&qpath)) == format!("{bounded_ty}:");
-
+                if !is_from_proc_macro(cx, p.bounded_ty);
                 then {
                     let trait_bounds = v
                         .iter()
@@ -272,7 +269,10 @@ impl TraitBounds {
                         .filter_map(get_trait_info_from_bound)
                         .map(|(_, _, span)| snippet_with_applicability(cx, span, "..", &mut applicability))
                         .join(" + ");
-                    let hint_string = format!("consider combining the bounds: `{bounded_ty}: {trait_bounds}`");
+                    let hint_string = format!(
+                        "consider combining the bounds: `{}: {trait_bounds}`",
+                        snippet(cx, p.bounded_ty.span, "_"),
+                    );
                     span_lint_and_help(
                         cx,
                         TYPE_REPETITION_IN_BOUNDS,
--- a/clippy_lints/src/trait_bounds.rs
+++ b/clippy_lints/src/trait_bounds.rs
@@ -260,6 +260,9 @@ impl TraitBounds {
                     SpanlessTy { ty: p.bounded_ty, cx },
                     p.bounds.iter().collect::<Vec<_>>()
                 );
+                if let TyKind::Path(qpath) = p.bounded_ty.kind;
+                if format!("{}:", rustc_hir_pretty::qpath_to_string(&qpath))
+                    == format!("{}:", snippet(cx, p.bounded_ty.span, "_"));
 
                 then {
                     let trait_bounds = v
--- a/clippy_lints/src/trait_bounds.rs
+++ b/clippy_lints/src/trait_bounds.rs
@@ -262,7 +262,7 @@ impl TraitBounds {
                 );
                 let bounded_ty = snippet(cx, p.bounded_ty.span, "_");
                 if let TyKind::Path(qpath) = p.bounded_ty.kind;
-                if format!("{}:", rustc_hir_pretty::qpath_to_string(&qpath)) == format!("{}:", bounded_ty);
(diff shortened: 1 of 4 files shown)


clippy_lints/src/trait_bounds.rs
```diff
@@ -260,6 +260,9 @@ impl TraitBounds {
                     SpanlessTy { ty: p.bounded_ty, cx },
                     p.bounds.iter().collect::<Vec<_>>()
                 );
+                if let TyKind::Path(qpath) = p.bounded_ty.kind;
+                if format!("{}:", rustc_hir_pretty::qpath_to_string(&qpath))
```
Centri3, line 264: I've never seen the function qpath_to_string before, and is never used in clippy, but it's extremely handy here. Should I change this to something else? Is there a reason it's never used?

xFrednet: I don't know about this specifc case, but I'm very sure, that using paths like this is prone to false positives. I would guess that this wouldn't hold for paths with reexported items, aliasses or some weird relative path like `duck::super::quack`.

Is there a reason why [`clippy_utils::is_from_proc_macro`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.is_from_proc_macro.html#) wouldn't work in this case? It's suppose to handle cases, were the spans are not marked as originating from a macro :)

cc @blyxyas this might be interesting to you and you might also have some thoughts :).

Centri3: We could likely implement `WithSearchPat` for `WherePredicate` relatively easily

blyxyas: [`qpath_to_string`](https://doc.rust-lang.org/beta/nightly-rustc/src/rustc_hir_pretty/lib.rs.html#218-220) is a really expensive (it uses [`print_path`](https://doc.rust-lang.org/beta/nightly-rustc/src/rustc_hir_pretty/lib.rs.html#1603)) function, so I think really the best way to go here, as Centri3 said, is by implementing `WithSearchPat` on `TyKind`.
(I will still benchmark these PR so we are sure that this is a performance problem)

blyxyas: As a kind of side note, the average time for testing `type_repetition_in_bounds` in this PR is about 190 millis (run 1000 times). While in `master`'s branch, it's 168 millis.

Is this significant? *Maybe...* (It would be a nice addition to have `WithSearchPath` implemented for more like `TyKind` or `Attribute` (useful for `allow_attributes`, which is bugged for this exact reason. But it isn't a deal-breaker))

Centri3: > As a kind of side note, the average time for testing `type_repetition_in_bounds` in this PR is about 190 millis (run 1000 times). While in `master'`s branch, it's 168 millis.

I'm curious what the difference would be if `is_from_proc_macro` was used instead.

> is by implementing `WithSearchPat` on `TyKind`

For this PR in particular I think we should implement both on `TyKind` and  `WherePredicate`, since that's what's linted on anyway. afaict it would just need to search for `'` or the type's name (if it's `BoundPredicate` or `RegionPredicate`, respectively). `TyKind` (or `Ty` to be more specific) would likely

> or `Attribute`

This was actually already implemented in #10869 :)

Centri3: Hmm, I'm trying to implement it for `Ty` but am not sure what to use for `Path`... I'll commit the current implementation and hope someone can give their thoughts on it

blyxyas (approved):
I think this is ready to merge. Thanks!

cc @xFrednet

xFrednet:
Sounds like you're back in action. Ain't no rest for the wicket, right? ^^

I'll give this a final review in the next couple of days. Thank you for the reviewing effort @blyxyas

clippy_utils/src/check_proc_macro.rs
```diff
@@ -319,6 +319,44 @@ fn attr_search_pat(attr: &Attribute) -> (Pat, Pat) {
...
+//     match where_pred {
+//         WherePredicate::BoundPredicate(bound) => {
+//             todo!();
+//         },
+//         WherePredicate::RegionPredicate(region) => {
+//
+//         },
+//         WherePredicate::EqPredicate(..) => unimplemented!(),
+//     }
+// }
```
xFrednet, line 333: Should this comment be part of this PR or is it just an artifact of your dev process? :D

xFrednet (commented):
LGTM, just one tiny question before the merge :)

clippy_utils/src/check_proc_macro.rs
Centri3: Forgot to finish that, I think I'll remove it since checking the `Ty` already works

xFrednet (approved):

xFrednet:
This version looks good to me, thank you! @Centri3 

Also thank you, @blyxyas, for the review support!

@bors r=blyxyas,xFrednet

Issue #9311, opened 2022-08-09
Update docs and tooling to focus on `LateLintPasses`


xFrednet:
Our documentation, how to add a lint, uses an `EarlyLintPass` for the example lint. However, most new lints use the `LateLintPass` since they need type information from rustc. Since the `LateLintPass` is the main one used for new lint, I suggest updating the documentation to focus on that pass and to update the `cargo dev new_lint` tool to use the `LateLintPass` by default.

Related discussion on Zulip: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202022-08-09/near/292596176

I'll wait a bit until I do these changes, to potentially get additional feedback, but also because I don't have that much time rn :sweat_smile: :upside_down_face:

nahuakang:
I missed that Zulip chat but I think this update would make perfect sense. 

On the documentation side, could we maybe keep the `EarlyLintPass` as a chapter and have another `LateLintPass` as a chapter. Then newcomers can read whichever one that suits the ticket they take up?

Fully agree that `cargo dev new_lint` could use `LateLintPass` by default (or have `--early` and `--late` flags)?

xFrednet:
> On the documentation side, could we maybe keep the EarlyLintPass as a chapter and have another LateLintPass as a chapter. Then newcomers can read whichever one that suits the ticket they take up?

That would also be an option, i think the important part is, that we clearly indicate when which version is better. And in the big majority the LateLintPass is even reauired. My goal with this suggestions is to simplify contributing a bit and prevent information overload. so if we find a way to simplify everything and keep both chapters that would be even better :)

nahuakang:
@xFrednet @flip1995 👋 Hey everyone, I'm happy to help with iterating over the docs aspect of this issue, i.e. editing and revamping the existing documentation for adding lints, especially switching to emphasis on `LateLintPass`.

Below is a draft outline I organized. Please review it and give your suggestions:

```
- **Adding lints**
    - **Overview: How to define a lint**
        - Standalone
        - Location for lint tests
        - How to add specific type of lint (`methods`, `function`, etc.)
        - Lint passes (late vs early)
    - **Write a lint with `LateLintPass` (elaborative)**
        - Define the lint
        - Lint registration
        - Writing tests
        - Emitting the lint
        - Add lint logic
            - (TBD) Teach how to check for type
            - (TBD) Teach how to check for certain trait
            - (TBD) Teach how to access items by IDs
[...]

PR #10907, opened 2023-06-08
Direct towards late passes in `cargo dev new_lint`

Alexendoo:
changelog: none

This would be the tooling part of #9311

- `--pass late` is now the default
- It prints a message recommending the use of a late pass if you choose `--pass early`

### DIFF
--- a/clippy_dev/src/new_lint.rs
+++ b/clippy_dev/src/new_lint.rs
@@ -37,7 +37,7 @@ impl<T> Context for io::Result<T> {
 ///
 /// This function errors out if the files couldn't be created or written to.
 pub fn create(
-    pass: Option<&String>,
+    pass: &String,
     lint_name: Option<&String>,
     category: Option<&str>,
     mut ty: Option<&str>,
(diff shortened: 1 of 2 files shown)


flip1995:
> * It prints a message recommending the use of a late pass if you choose `--pass early`

I wouldn't add this. If someone passes `--pass early` they probably have a reason to do so already.

clippy_dev/src/new_lint.rs
```diff
@@ -63,6 +63,14 @@ pub fn create(
         add_lint(&lint, msrv).context("Unable to add lint to clippy_lints/src/lib.rs")?;
     }
 
+    if pass == "early" {
+        println!(
+            "\n\
+            NOTE: Use a late pass unless you need something specific from\n\
+            an early pass, as they lack many features and utilities"
```
flip1995, line 70: You convinced me that this is a good idea. But I would remove the `\n` from the message and let the user terminal insert the line ~~breaks~~wraps. Otherwise this will look really weird, if the terminal inserts extra line wraps

PR #10909, opened 2023-06-08
Fix `useless_vec` suggestion in `for _ in vec![..]`

### LINKED ISSUE #2256: useless_vec doesn't suggest a necessary &
In some cases, the useless_vec warning will suggest changing a `vec![...]` to a `&[...]`. In this case, a `*` or a `&` may be required. This is not mentioned in the help text or in the extended help text at https://rust-lang-nursery.github.io/rust-clippy/master/index.html#useless_vec

A test case reproducing this:

```
fn main() {
    for a in vec![0] {
        foo(a)
    }
}

fn foo(a: i64) {
    println!("{}", a)
}
```

Clippy will suggest changing `vec![0]` to `&[0]`. A `&` before the `a` in `for a in` is also needed, or similar. While the compiler will suggest this change, it'd be better to make this clear up front.

Alexendoo:
Fixes https://github.com/rust-lang/rust/issues/111034
Fixes #2256

changelog: [`useless_vec`]: Fix suggestion in `for _ in vec![..]`

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -699,7 +699,12 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     });
     let too_large_for_stack = conf.too_large_for_stack;
     store.register_late_pass(move |_| Box::new(escape::BoxedLocal { too_large_for_stack }));
-    store.register_late_pass(move |_| Box::new(vec::UselessVec { too_large_for_stack }));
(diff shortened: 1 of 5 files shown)


llogiq:
Good improvement! Thanks Alex!

@bors r+

llogiq:
You should `cargo dev fmt` your code and either call `above` and `below` somewhere in your test or `#[allow(unused)]`.

Alexendoo:
Fourth time was the charm 👀

PR #10910, opened 2023-06-08
Ignore more pointer types in `unnecessary_cast`

Alexendoo:
Spotted this because

https://github.com/rust-lang/rust-clippy/blob/e2c655b4c07c912ef749be316aeea8453cb9d840/tests/ui/suspicious_to_owned.rs#L9-L10

currently fails on `aarch64-unknown-linux-gnu` as `c_char` is `u8` there

The current implementation checks for `as alias`, `as _`. This adds things like
- `as *const alias`
- `as *const cfg_dependant`
- `as *const _`

changelog: none


llogiq:
Thank you!

@bors r+

Issue #9657, opened 2022-10-15
as_conversions triggers on derived code


RReverser:
### Summary

I like using as_conversions as a catch-all guard to stop myself from reaching for it in my own code when better solutions are available. However, it triggers on the generated derive code too, e.g. when using [serde-repr](https://github.com/dtolnay/serde-repr).

Worse, it can't even be disabled by putting `#[allow(clippy::as_conversions)]` on the offending type, as it doesn't propagate to the derive-generated code, so the only option is to opt-out at the module level.

### Lint Name

as_conversions

### Reproducer

I tried this code:

```rust
#![warn(clippy::as_conversions)]

use serde_repr::{Deserialize_repr, Serialize_repr};

#[derive(Serialize_repr, Deserialize_repr)]
#[repr(i32)]
pub enum State {
    Error = -1,
    Idle = 0,
    Waiting = 1,
    Done = 2,
}
```

I saw this happen:

```
using a potentially dangerous silent `as` conversion
consider using a safe wrapper for this conversion
for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#as_conversions
```

I expected to see this happen:

(no warning, as the `as` conversion is used internally by an external crate)

### Version

```text
rustc 1.64.0 (a55dd71d5 2022-09-19)
binary: rustc
commit-hash: a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52
commit-date: 2022-09-19
host: x86_64-pc-windows-msvc
release: 1.64.0
LLVM version: 14.0.6
```

### Additional Labels

@rustbot label +T-macros

kraktus:
root issue: https://github.com/dtolnay/serde-repr/blob/78e7199f79bc0ad441fcd8aaa1b2d8a41f3672c7/src/lib.rs#L58

I'm not sure why this is not considered as `in_external_macro`, I assume because it reuses root spans.

RReverser:
> root issue: [dtolnay/serde-repr@`78e7199`/src/lib.rs#L58](https://github.com/dtolnay/serde-repr/blob/78e7199f79bc0ad441fcd8aaa1b2d8a41f3672c7/src/lib.rs#L58)
> 
> I'm not sure why this is not considered as `in_external_macro`, I assume because it reuses root spans.

The `as` itself doesn't come from the source though, I guess clippy should be looking at the keyword's span but doesn't currently.

RReverser:
[...]

PR #10911, opened 2023-06-08
Don't linting `as_conversions` in proc macros

lochetti:
Don't linting `as_conversions` if code was generated by procedural macro.

This PR fixes https://github.com/rust-lang/rust-clippy/issues/9657

I implemented the fix changing the lint code to be a `LateLintPass` in order to be able to use the `is_from_proc_macro` out of the box. If the reviwer thinks that it would be better to do the other way (implementing `WithSearchPat`) just let me know. I might need some help in implementing it for the `ustc_ast::ast::Expr`

changelog: [`as_conversions`] avoiding warnings in macro-generated code


Alexendoo (approved):

Alexendoo:
Thanks! Yeah a late pass sounds like the best way to do it

@bors r+

PR #10916, opened 2023-06-09
New lint [`min_ident_chars`]

### LINKED ISSUE #10915: `single_letter_ident`
### What it does

I'd like to suggest a restriction, `single_letter_ident`, which forbids identifiers that are a single letter

### Advantage

This would allow projects that want to enforce longer identifier names (for readability) to do so without breaking existing projects

### Drawbacks

None that I can think of, as it's a restriction which needs to be opted into

### Example

```rust
impl fmt::Display for MyStruct {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.write_str(self.as_ref())
    }
}
```

Could be written as:

```rust
impl fmt::Display for MyStruct {
    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
        formatter.write_str(self.as_ref())
    }
}
```

This would also forbid the following:

```rust
struct A {}

enum B {}

trait C {}

fn d() {}

const E: &str = "some const";

static F: &str = "some const";
```

TBD for enum members / struct fields / assoc. consts / assoc. types, where this makes a bit more sense, e.g.:

```rust
struct Point {
  x: usize,
  y: usize,
}
```

Generics (e.g. `T`) should be allowed to remain single letter idents (IMO, can also be debated)

Alternatively, could be a `min_ident_len` lint which has configuration (since 2 letter idents might also be undesirable)

The lint could also optionally (or by configuration) whitelist specific identifiers which are common, e.g. `n` for number functions, `i`, `x`, `y`, etc. This could be narrowed to only allow these if they're of specific types to avoid confusion and still restrict the same identifiers for arbitrary cases. TBD whether this is desirable.

Centri3:
Closes #10915

This also implements `WithSearchPat` for `Ident`, as I was going to rewrite this as a late lint to optionally ignore fields and/or generics but this was more complex than anticipated

changelog: New lint [`min_ident_chars`]
[#10916](https://github.com/rust-lang/rust-clippy/pull/10916)


xFrednet:
Hey @blyxyas do you want to review this PR? It sounds like a good and simple lint :)

Assuming it's fine for @Jarcho if we steal the review.

r? @xFrednet

blyxyas:
Yes, will take it :)

clippy_lints/src/single_letter_idents.rs
```diff
@@ -0,0 +1,52 @@
...
+use rustc_lint::{EarlyContext, EarlyLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::symbol::Ident;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for idents which comprise of a single letter.
+    ///
+    /// Note: This lint can be very noisy when enabled; it even lints generics! it may be desirable
```
blyxyas: It should probably not lint generics, as single-letter generics are the idiomatic form to put them.

clippy_lints/src/single_letter_idents.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// to only enable it temporarily.
+    ///
+    /// ### Why is this bad?
+    /// In many cases it's not, but at times it can severely hinder readability. Some codebases may
+    /// wish to disallow this to improve readability.
+    ///
+    /// ### Example
+    /// ```rust,ignore
```
blyxyas:
```suggestion
    /// ```rust,ignore
    /// for m in movies {
    ///		let title = m.t;
    /// }
    /// ```
    /// Use instead:
    /// ```
    /// for movie in movies {
    /// 	let tilte = movie.title;
    /// }
    /// ```
```

clippy_lints/src/utils/conf.rs
```diff
@@ -514,6 +515,11 @@ define_Conf! {
     ///
     /// The byte size a `T` in `Box<T>` can have, below which it triggers the `clippy::unnecessary_box` lint
     (unnecessary_box_size: u64 = 128),
+    /// Lint: SINGLE_LETTER_IDENTS.
+    ///
+    /// Allowed single letter idents.
+    (allowed_idents: rustc_data_structures::fx::FxHashSet<char>
```
blyxyas, line 521: Maybe another config option for a minimal length identifier? Instead of having another lint, I think having compact here is the best option.

If you're going to implement it, lints when ident's length i 1 should be different from messages when ident's length < config. So it's clear that the configuration value is what caused that message.

blyxyas (commented):
Some really *really* nitpicks. Everything else looks fine, Thanks :heart:

clippy_lints/src/single_letter_idents.rs
Centri3: I tried this, but couldn't quite get it to work. Getting the `Res` from a `PathSegment` (field on `Node::TyParam`) doesn't work, but of course, that would be solved by just accessing the `res` field on it :sweat_smile: ... I'll probably add a configuration option for that later, along with fields

blyxyas (approved):
Everything seems pretty fine, thanks! Seems like a very useful lint :heart: 

cc @xFrednet

clippy_lints/src/utils/conf.rs
```diff
@@ -34,6 +34,7 @@ const DEFAULT_DOC_VALID_IDENTS: &[&str] = &[
     "CamelCase",
 ];
 const DEFAULT_DISALLOWED_NAMES: &[&str] = &["foo", "baz", "quux"];
+const DEFAULT_ALLOWED_IDENTS_BELOW_MIN_CHARS: &[&str] = &["i", "j", "x", "y", "z", "w", "n"];
```
xFrednet, line 37: For list configurations with default parameters, it's nice to allow the user to only extend the list and not replace it. Clippy has a nice small util to support the `".."` value to indicate this behavior.

Here is a link to the relevant code snippet:

https://github.com/rust-lang/rust-clippy/blob/d44ea7c5c5d51403baa6eeb0e3bbdb0a1ffeed1f/clippy_lints/src/utils/conf.rs#LL586C1-L587C104

Could you also update the conf documentation, to mention this option? You can just copy that part from the other configs :)

tests/ui-toml/min_ident_chars/clippy.toml
```diff
@@ -0,0 +1,2 @@
+allowed-idents-below-min-chars = ["Owo", "Uwu", "wha", "t_e", "lse", "_do", "_i_", "put", "her", "_e"]
```
xFrednet, line 1: Looking at this list, I would think that we should allow indents starting with an underscore in general, as those are not used later on. I think it would be fine to allow them all the time, but you could also add a config value for that :)

tests/ui/min_ident_chars.rs
```diff
@@ -0,0 +1,81 @@
...
+    let l = 4;
+    let o = 6;
+    // 2 len does not lint
+    let hi = 0;
+    // Lint
+    let (h, o, w) = (1, 2, 3);
+    for (a, (r, e)) in (0..1000).enumerate().enumerate() {}
+    let you = Vec4 { x: 1, y: 2, z: 3, w: 4 };
+    while let (d, o, _i, n, g) = (true, true, false, false, true) {}
+    let today = true;
```
xFrednet, line 63: Quite good, thank you for asking. My semester is finally over and the summer is super beautiful where I live, without it getting too hot. Can recommend!

xFrednet:
You just pushed, while I was doing a review xD

Centri3:
Sorry, had to make sure I ran `collect-metadata` on this one (spoiler alert: I did not.)

tests/ui/min_ident_chars.rs
```diff
@@ -0,0 +1,81 @@
...
+struct Vec4 {
+    x: u32,
+    y: u32,
+    z: u32,
+    w: u32,
+}
+
+struct AA<T, E>(T, E);
+
+fn main() {
```
xFrednet, line 40: Could you add a test for function parameters like this:

```rs
// Should lint:
fn wrong_pythagoras(a: f32, b: f32) -> f32 {
    a * a + a * b
}
```

and one for just the underscore identifier:

```rs
// Should not lint
for _ in 0..10 {}
```

The later one could maybe lint, but then we should add `_` as a default allowed ident

xFrednet (commented):
Hey, so far, I've only looked at the tests and config implementation. It looks like you're also still working on the branch. These are some nits I found rn. Let me know when you're ready for the next review :)

xFrednet:
> Sorry, had to make sure I ran `collect-metadata` on this one (spoiler alert: I did not.)

No problem! It was not the first time this happened, just the first time that GH deleted a comment. It was a simple one to reconstruct though :D

tests/ui-toml/min_ident_chars/clippy.toml
Centri3: That sounds reasonable, I think not having a config value for it is fine

clippy_lints/src/utils/conf.rs
Centri3: Done

tests/ui/min_ident_chars.rs
Centri3: Done, the latter was already handled for by only checking `Binding`s (that will be a `Wild`, afaik)

Centri3:
> Let me know when you're ready for the next review :)

Feel free to take a look whenever, the implementation is already pretty much done.

xFrednet:
I looks like the PR has gotten some conflict, without bors adding a comment. Could you resolve them?

clippy_lints/src/min_ident_chars.rs
```diff
@@ -0,0 +1,154 @@
...
+        let Some(ident) = node.ident() else {
+            return;
+        };
+
+        let str = ident.as_str();
+        if !in_external_macro(cx.sess(), ident.span)
+            && str.len() <= conf.min_ident_chars_threshold as usize
+            && !str.starts_with('_')
+            && !str.is_empty()
+            && conf.allowed_idents_below_min_chars.get(&str.to_owned()).is_none()
```
xFrednet, line 110: NIT: The `check_pat` uses the same checks (in a slightly different order). I think it would be better, to move this into a separate method of `MinIdentChars` and use the same one in both locations :)

xFrednet (commented):
This version looks quite good to me. I have three tiny nits, and then it should be ready :)

clippy_lints/src/min_ident_chars.rs
```diff
@@ -0,0 +1,154 @@
...
+    cx: &'cx LateContext<'tcx>,
+}
+
+impl Visitor<'_> for IdentVisitor<'_, '_> {
+    #[expect(clippy::cast_possible_truncation)]
+    fn visit_id(&mut self, hir_id: HirId) {
+        let Self { conf, cx } = *self;
+        // Reimplementation of `find`, as it uses indexing, which can (and will in async functions) panic.
+        // This should probably be fixed on the rustc side, this is just a temporary workaround.
+        // FIXME: Remove me if/when this is fixed in rustc
```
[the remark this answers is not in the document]
Centri3: Done, rust-lang/rust#112534

clippy_lints/src/min_ident_chars.rs
Centri3: Just didn't feel like doing this before :D I'll finish it now.

xFrednet (approved):

xFrednet:
Looks good to me, thank you!

@bors r=blyxyas,xFrednet

PR #10920, opened 2023-06-10
Add `SPEEDTEST`

blyxyas:
In the `master` branch, we currently don't have any way to test the performance of a single lint in changes.
This PR adds `SPEEDTEST`, the environment variable which lets you do a speed test on a lint / category of tests with various configuration options.

Maybe we should merge this with `lintcheck` :thinking: 
See the book page for more information.

changelog:none

### DIFF
--- a/tests/compile-test.rs
+++ b/tests/compile-test.rs
@@ -213,12 +213,46 @@ fn run_ui_cargo() {
 
 fn main() {
     set_var("CLIPPY_DISABLE_DOCS_LINKS", "true");
-    run_ui();
-    run_ui_toml();
-    run_ui_cargo();
-    run_internal_tests();
-    rustfix_coverage_known_exceptions_accuracy();
-    ui_cargo_toml_metadata();
+    // The SPEEDTEST_* env variables can be used to check Clippy's performance on your PR. It runs the
+    // affected test 1000 times and gets the average.
+    if let Ok(speedtest) = std::env::var("SPEEDTEST") {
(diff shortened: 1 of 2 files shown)


tests/compile-test.rs
```diff
@@ -355,10 +355,42 @@ fn run_ui_cargo() {
...
+        } else {
+            iterations = 1000;
+        }
+
+        let mut sum = 0;
+        for _ in 0..iterations {
+            let start = std::time::Instant::now();
+            f();
+            sum += start.elapsed().as_millis();
+        }
```
y21, line 386: since the timing itself also takes some time to run, would it make sense to move the timing stuff out of the loop and divide the time taken by the number of iterations after the loop? the Instant::now().elapsed() overhead might also be negligible compared to how long a lint pass takes, so maybe this doesn't matter that much (from what I remember, it was around 40ns on the playground...)

y21 (commented):
I usually put a Instant::now().elapsed() check in a lint myself when making a PR to see roughly what impact it has, so this would be really nice to have if it was directly supported! I have one question tho:

tests/compile-test.rs
llogiq: I think the usual way to micro benchmark is to time variable numbers of iterations, then do a linear regression, but this is clearly a macro benchmark, so don't worry about either loop or time measurement overhead.

llogiq:
Thank you! @bors r+

PR #10925, opened 2023-06-11
add lint [`incorrect_clone_impl_on_copy_type`]

### LINKED ISSUE #10700: `Clone` should be `*self` for types which are always `Copy`
### What it does

If a type's `Clone` impl has the same `where` bounds as its `Copy` impl, then this gives a warning unless the `Clone` impl body is exactly
```rust
*self
```

Inspired by the incorrect code in <https://users.rust-lang.org/t/why-is-the-value-expression-with-the-type-implemented-copy-not-desugared-to-e-clone/92921?u=scottmcm>, which currently gets no clippy warnings <https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=3fa61e0abb79c6a444d2ef77a5c6033c>.

### Lint Name

overcomplicated_clone_impl

### Category

suspicious, complexity

### Advantage

It's shorter to write than other options, and it always correct -- anything with different behaviour would be a violation of the `Copy`+`Clone` contract.

Thus using any other implementation is just asking for that more-complicated implementation to have a logic error.

### Drawbacks

Macros might not want to special-case their generation to meet this.

### Example

```rust
use std::marker::PhantomData;
struct Id<T>(u32, PhantomData<T>);

impl<T> Clone for Id<T> {
    fn clone(&self) -> Self {
        Id(self.0.clone(), PhantomData)
    }
}
impl<T> Copy for Id<T> {}
```

Could be written as:

```rust
use std::marker::PhantomData;
struct Id<T>(u32, PhantomData<T>);

impl<T> Clone for Id<T> {
    fn clone(&self) -> Self {
        *self
    }
}
impl<T> Copy for Id<T> {}
```

(Which, notably, can't just use `#[derive(Copy, Clone)]`.)

Centri3:
Split off from #10788.

Closes #10700

----

changelog: new lint [`incorrect_clone_impl_on_copy_type`]
[#10925](https://github.com/rust-lang/rust-clippy/pull/10925)


xFrednet:
@Centri3, you're on a real PR creation spree, damn! I'll review this one over the coming week :D

clippy_lints/src/needless_impls.rs
```diff
@@ -0,0 +1,109 @@
...
+        let node = get_parent_node(cx.tcx, impl_item.hir_id());
+        let Some(Node::Item(item)) = node else {
+            return;
+        };
+        let ItemKind::Impl(imp) = item.kind else {
+            return;
+        };
+        let Some(trait_impl) = cx.tcx.impl_trait_ref(item.owner_id).map(EarlyBinder::skip_binder) else {
+            return;
+        };
```
xFrednet, line 63: Wouldn't it be smarter, to use `check_item`, search for `ItemKind::Impl` and then walk downwards? Then you can do all of these checks only once. The lint would need to be emitted with [`span_lint_hir`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/diagnostics/fn.span_lint_hir.html#) to make sure that the lint level is tracked correctly. But the rest should work just fine :)

clippy_lints/src/needless_impls.rs
```diff
@@ -0,0 +1,109 @@
...
+        let body = cx.tcx.hir().body(impl_item_id);
+        let ExprKind::Block(block, ..) = body.value.kind else {
+            return;
+        };
+        // Above is duplicated from the `duplicate_manual_partial_ord_impl` branch.
+        // Remove it while solving conflicts once that PR is merged.
+
+        // Actual implementation; remove this comment once aforementioned PR is merged
+        if cx.tcx.is_diagnostic_item(sym::Clone, trait_impl_def_id)
```
xFrednet, line 79: `Clone` also has a method called `clone_from`. I wonder if it's worth checking for that, even though, it's used very very rarely

xFrednet (commented):
I have a few smaller comments, but all of them should hopefully be simple to solve :). Please reach out, if you have any questions :)

clippy_lints/src/needless_impls.rs
Centri3: We could probably say to remove the entire implementation if `Copy` is implemented

clippy_lints/src/needless_impls.rs
Centri3: I think this runs into the issue that it needs to check every item of `ItemKind::Impl`, rather than just one. I think that's how it was done but resulted in it being rather hard to read and a bit slow.

clippy_lints/src/needless_impls.rs
```diff
@@ -0,0 +1,109 @@
...
+    /// impl Clone for A {
+    ///     fn clone(&self) -> Self {
+    ///         *self
+    ///     }
+    /// }
+    ///
+    /// impl Copy for A {}
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub NEEDLESS_CLONE_IMPL,
```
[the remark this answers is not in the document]
Centri3: `different_clone_copy_impl` is incorrect for the `PartialOrd` side of the coin, so I think we should go with `incorrect`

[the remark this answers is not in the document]
xFrednet: I'm fine with that :)

clippy_lints/src/needless_impls.rs
xFrednet: Interesting, I would expect it to be faster, as we only check the elements inside impl items and avoid the lookup, but I might be mixing up projects rn xD. I believe we can keep it like this. I'll give you another review this week :)

Centri3: In truth, it's completely possible it is as I hadn't adequately benchmarked the old one. I'll reimplement the previous and see the proper difference

xFrednet: I don't think that's necessary, but the results would be interesting :)

Centri3: I just benchmarked it, and it's ~2x slower than the current implementation, which is very surprising, I have no idea what's causing that.

I think I'll stick with the current implementation because of that

xFrednet: SGTM, I'll give you a final review this week and most likely r+ it :)

xFrednet (approved):

xFrednet:
This looks good to me. Thank you for the new lint!

@bors r+

PR #10931, opened 2023-06-12
[`unnecessary_fold`]: suggest turbofish if necessary

### LINKED ISSUE #10000: FP unnecessary_fold: no longer able to infer type
### Summary

.

### Lint Name

unnecessary_fold

### Reproducer

I tried this code:

```rust
use std::hash::BuildHasher;

fn smoketest_map<S: BuildHasher>(mut map: HashMap<i32, i32, S>) {
    map.insert(0, 0);
    assert_eq!(map.values().fold(0, |x, y| x + y), 0);
}

fn main() {
    smoketest_map(HashMap::new());
}
```

I saw this happen:

```

  * src/main.rs

error[E0283]: type annotations needed
    --> src/main.rs:6:29
     |
6    |     assert_eq!(map.values().sum(), 0);
     |                             ^^^ cannot infer type of the type parameter `S` declared on the associated function `sum`
     |
     = note: cannot satisfy `_: std::iter::Sum<&i32>`
note: required by a bound in `std::iter::Iterator::sum`
    --> /home/matthias/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/iter/traits/iterator.rs:3379:12
     |
3379 |         S: Sum<Self::Item>,
     |            ^^^^^^^^^^^^^^^ required by this bound in `std::iter::Iterator::sum`
help: consider specifying the type argument in the function call
     |
6    |     assert_eq!(map.values().sum::<S>(), 0);
     |                                +++++

For more information about this error, try `rustc --explain E0283`.

warning: this `.fold` can be written more succinctly using another method
 --> src/main.rs:6:29
  |
6 |     assert_eq!(map.values().fold(0, |x, y| x + y), 0);
  |                             ^^^^^^^^^^^^^^^^^^^^^ help: try: `sum()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_fold
  = note: `#[warn(clippy::unnecessary_fold)]` on by default
```

### Version

```text
rustc 1.67.0-nightly (e0098a5cc 2022-11-29)
binary: rustc
commit-hash: e0098a5cc3a87d857e597af824d0ce1ed1ad85e0
commit-date: 2022-11-29
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

y21:
Fixes #10000

This adds turbofish `::<T>` to the suggestion in `unnecessary_fold`. This is necessary because the `Sum` trait is generic, which breaks inference when changing `fold()` to `sum()`.

changelog: [`unnecessary_fold`]: suggest turbofish if necessary

### DIFF
--- a/clippy_lints/src/methods/unnecessary_fold.rs
+++ b/clippy_lints/src/methods/unnecessary_fold.rs
@@ -12,14 +12,10 @@ use rustc_span::{source_map::Span, sym};
 
 use super::UNNECESSARY_FOLD;
 
-/// No turbofish needed in any case.
-fn no_turbofish(_: &LateContext<'_>, _: &hir::Expr<'_>) -> bool {
-    false
-}
-
-/// Turbofish (`::<T>`) may be needed, but can be omitted if we are certain
-/// that the type can be inferred from usage.
-fn turbofish_if_not_inferred(cx: &LateContext<'_>, expr: &hir::Expr<'_>) -> bool {
+/// Do we need to suggest turbofish when suggesting a replacement method?
+/// Changing `fold` to `sum` needs it sometimes when the return type can't be
+/// inferred. This checks for some common cases where it can be safely omitted
+fn needs_turbofish(cx: &LateContext<'_>, expr: &hir::Expr<'_>) -> bool {
     let parent = cx.tcx.hir().get_parent(expr.hir_id);
 
     // some common cases where turbofish isn't needed:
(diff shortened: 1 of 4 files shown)


clippy_lints/src/methods/unnecessary_fold.rs
```diff
@@ -7,10 +7,74 @@ use rustc_errors::Applicability;
...
+    // if it's neither of those, stay on the safe side and suggest turbofish,
+    // even if it could work!
+    true
+}
+
+#[derive(Copy, Clone)]
+struct Replacement {
+    method_name: &'static str,
+    has_args: bool,
+    requires_turbofish: fn(&LateContext<'_>, &hir::Expr<'_>) -> bool,
```
Jarcho, line 56: The normal thing to do here would be to have the field be something like `has_generic_return`. Since you aren't taking advantage of the extra flexibility provided by function types, don't use them.

y21: ok, changed this in the latest commit. Is that better?

Jarcho:
Thank you. @bors r+

PR #10933, opened 2023-06-12
[`useless_vec`]: lint on `vec![_]` invocations that adjust to a slice

### LINKED ISSUE #2262: Lint uses of `Vec` where arrays would do fine

y21:
Fixes #2262 (well, actually my PR over at #10901 did do most of the stuff, but this PR implements the other case mentioned in the comments that my PR didn't fix)

Before this change, it would lint `(&vec![1]).iter().sum::<i32>()`, but not `vec![1].iter().sum::<i32>()`. This PR handles this case.
This also refactors a few things that I wanted to do in my other PR but forgot about.

changelog: [`useless_vec`]: lint on `vec![_]` invocations that adjust to a slice

### DIFF
--- a/clippy_lints/src/vec.rs
+++ b/clippy_lints/src/vec.rs
@@ -8,7 +8,6 @@ use clippy_utils::ty::is_copy;
 use clippy_utils::visitors::for_each_local_use_after_expr;
 use clippy_utils::{get_parent_expr, higher, is_trait_method};
 use if_chain::if_chain;
-use rustc_ast::BindingAnnotation;
 use rustc_errors::Applicability;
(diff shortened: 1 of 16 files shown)


Centri3:
Will take a look in a bit :)

Centri3 (approved):
This looks great, thanks! @Manishearth, can you take a quick look just in case I missed something?

Manishearth (approved):

PR #10934, opened 2023-06-12
new lint [`single_range_in_vec_init`]

### LINKED ISSUE #10932: Add a lint for when ranges are used inside of a Vec
### What it does

When someone goes to put a range inside of a Vec, Clippy should probably check with them to make sure that's what they actually want to do. They may want to go over the range and collect that all into a Vec.

### Advantage

- New programmers to Rust (especially those using Rustlings) can get more information for why using a `vec![0..100]` doesn't work.
- Prevents using the wrong type on accident inside of a Vec.

### Drawbacks

None AFAIK, aside from needing to disable it for valid use cases.

### Example

```rust
let a = [0..200];
```

You may want to make a Vec of integers, with the length specified here, not a Vec of ranges. Here are some potential solutions:

- Initialize a Vec of length 200, with each value set to 0.
```rust
let a = vec![0; 200]
```
- Initialize a Vec, containing the values from 1 to 200.
```rust
let a = (0..200).collect::<Vec<i32>>()
```

Alternatively, you may want to make an Array instead, with each value initialized to 0:
```rust
let a = [0..200]
```

Centri3:
Lints on `vec![0..200]` (or `[0..200]`), suggesting either `(0..200).collect::<Vec<i32>>()` or `[0; 200]`.

Haven't tested it with anything that isn't primitive. Probably should!

Closes #10932

changelog: new lint [`single_range_in_vec_init`]

### DIFF
--- a/clippy_lints/src/single_range_in_vec_init.rs
+++ b/clippy_lints/src/single_range_in_vec_init.rs
@@ -1,10 +1,6 @@
 use clippy_utils::{
-    diagnostics::span_lint_and_then,
-    get_trait_def_id,
-    higher::VecArgs,
-    macros::root_macro_call_first_node,
-    source::{snippet_opt, snippet_with_applicability},
-    ty::implements_trait,
+    diagnostics::span_lint_and_then, get_trait_def_id, higher::VecArgs, macros::root_macro_call_first_node,
+    source::snippet_opt, ty::implements_trait,
 };
 use rustc_ast::{LitIntType, LitKind, UintTy};
 use rustc_errors::Applicability;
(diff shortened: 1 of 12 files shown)


clippy_lints/src/single_range_in_vec_init.rs
```diff
@@ -0,0 +1,147 @@
...
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use std::fmt::{self, Display, Formatter};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for `Vec` or array initializations that contain only one range.
+    ///
+    /// ### Why is this bad?
+    /// This is almost always incorrect, as it will result in a `Vec` that has only element. Almost
```
giraffate, line 17:
```suggestion
    /// This is almost always incorrect, as it will result in a `Vec` that has only one element. Almost
```

giraffate (commented):

giraffate:
@bors r+

Thanks!

PR #10935, opened 2023-06-12
Don't lint non-statement/faux empty `needless_if`s

Alexendoo:
Also has a basic fall-back for `if` statements that have attributes applied to them and incorporates https://github.com/rust-lang/rust-clippy/pull/10921#pullrequestreview-1474008780 while I was there

r? @Manishearth

changelog: none


Manishearth (approved):
there's still a risk of the macros themselves being cfg-replaced but should be fine

PR #10937, opened 2023-06-12
add lint `option_vec`

### LINKED ISSUE #10899: `option_vec`, a lint that checks for `Option<Vec<T>>`
### What it does

Checks for instances of `Option<Vec<T>>`, and recommends that they be converted to `Vec<T>`.

### Advantage

- `Option<Vec<T>>` holds very little advantage over `Vec<T>`, as a `Vec<T>` can already hold zero items.
- If the user wishes to use `Option<Vec<T>>` to denote a vector with zero or more items, and `Vec<T>` to denote a vector with one or more items, they can use Rust's type system to enforce this:
```rust
pub struct MyAst {
    /// Must have at least one value
    identifiers: Vec<String>,
}
// ...can be rewritten as...
pub struct MyAst {
    first_identifier: String,
    other_identifiers: Vec<String>,
}
// which will enforce, via the type system, at least one value.
```

### Drawbacks

The user may deliberately be using `Option<Vec<T>>` to denote a vector with zero or more items, and `Vec<T>` to denote a vector with one or more items (though, as described above, they can enforce this sort of thing using the type system).

### Example

```rust
pub struct MyAst {
    identifiers: Option<Vec<String>>,
}
```

Could be written as:

```rust
pub struct MyAst {
    identifiers: Vec<String>,
}
```

Centri3:
Closes #10899

I don't really agree with this lint all that much, but I guess that's what `restriction` lints are for :D

changelog: New lint [`option_vec`]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5059,6 +5059,7 @@ Released 2018-09-13
 [`option_map_unwrap_or_else`]: https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unwrap_or_else
 [`option_option`]: https://rust-lang.github.io/rust-clippy/master/index.html#option_option
 [`option_unwrap_used`]: https://rust-lang.github.io/rust-clippy/master/index.html#option_unwrap_used
+[`option_vec`]: https://rust-lang.github.io/rust-clippy/master/index.html#option_vec
 [`or_fun_call`]: https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call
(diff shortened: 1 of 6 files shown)


Alexendoo:
Nominated for a general discussion of this style of lint rather than this specific one, the

> I don't really agree with this lint all that much

kind

Centri3:
Closing as decided in https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202023-06-13, in favor of extending `disallowed-types` to work with generics

Issue #10555, opened 2023-03-27
unnecessary_cast on `as u64` for a c_ulong


joshtriplett:
### Summary

unnecessary_cast triggers on casts from C FFI type aliases, even though the type aliases may not always match the type being cast to.

### Lint Name

unnecessary_cast

### Reproducer

I tried this code:

```rust
use core::ffi::c_ulong;

fn main() {
    let x: c_ulong = 0;
    let y = x as u64;
    println!("Hello, {y}!");
}
```

I saw this happen:

```
warning: casting to the same type is unnecessary (`u64` -> `u64`)
 --> src/main.rs:5:13
  |
5 |     let y = x as u64;
  |             ^^^^^^^^ help: try: `x`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast
  = note: `#[warn(clippy::unnecessary_cast)]` on by default
```

I expected to not get an error, because `c_ulong` is a type alias that isn't always `u64`.

I thought https://github.com/rust-lang/rust-clippy/pull/8596 was supposed to fix this.

### Version

```text
rustc 1.68.1 (8460ca823 2023-03-20)
binary: rustc
commit-hash: 8460ca823e8367a30dda430efda790588b8c84d3
commit-date: 2023-03-20
host: x86_64-unknown-linux-gnu
release: 1.68.1
LLVM version: 15.0.6
```

joshtriplett:
Note that `.into()` has a similar false positive:

```rust
use core::ffi::c_ulong;

fn main() {
    let x: c_ulong = 0;
    let y: u64 = x.into();
    println!("Hello, {y}!");
}
```

produces:
```
warning: useless conversion to the same type: `u64`
 --> src/main.rs:5:18
  |
5 |     let y: u64 = x.into();
  |                  ^^^^^^^^ help: consider removing `.into()`: `x`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
  = note: `#[warn(clippy::useless_conversion)]` on by default
```

Alexendoo:
This one is tricky, #8596 would cover the case of `some_u64 as c_ulong` where we can check the `rustc_hir::Ty` of `c_ulong` to see it's an alias, but for the `x` in `x as u64` we only have the `rustc_middle::ty::Ty` which doesn't distinguish `u64` from `c_ulong`

joshtriplett:
[...]

PR #10942, opened 2023-06-13
Ignore more type aliases in `unnecessary_cast`

Centri3:
This is potentially the worst code I've ever written, and even if not, it's very close to being on par with starb. This will ignore `call() as i32` and `local_obtained_from_call as i32` now.

This should fix every reasonable way to reproduce #10555, but likely not entirely.

changelog: Ignore more type aliases in `unnecessary_cast`


llogiq:
I concur that the code isn't exactly beautiful, but it appears to fix the additional test cases, so thank you!

@bors r+

PR #10944, opened 2023-06-13
Remove dead code in `needless_pass_by_value`

GuillaumeGomez:
The `spans_need_deref` is never used so I removed it alongside all linked code as well.


xFrednet:
LGTM, would be interesting, how that evolved, but oh well. Thank you for the PR :)

@bors r+

PR #10945, opened 2023-06-13
[`no_effect`]: Suggest adding `return` if applicable

### LINKED ISSUE #10941: Lint to warn on a suspicious unnecessary operation, that happens to have the same type as the return type
### What it does

if there's `ControlFlow::Break(());`, and it's completely unnecessary, whilst the function returns `ControlFlow`, rather than just emitting `no_effect`, clippy should also say that it's likely intended to be returned.

This could be an enhancement to `no_effect` rather than a new lint.

### Advantage

- Saves the developer precious sanity

### Drawbacks

- None.

### Example

```rust
fn yep() -> bool {
	true;
}
```

Should be written as:

```rust
fn yep() -> bool {
	return true;
}
```
(Minus the `return` if it's unnecessary, but other lints will take care of that)

Centri3:
Closes #10941

Unfortunately doesn't catch anything complex as `no_effect` already wouldn't, but I'm fine with that (it catches `ControlFlow` at least :D)

changelog: [`no_effect`]: Suggest adding `return` if statement has same type as function's return type and is the last statement in a block

### DIFF
--- a/clippy_lints/src/no_effect.rs
+++ b/clippy_lints/src/no_effect.rs
@@ -1,11 +1,16 @@
 use clippy_utils::diagnostics::{span_lint_hir, span_lint_hir_and_then};
-use clippy_utils::is_lint_allowed;
 use clippy_utils::peel_blocks;
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::has_drop;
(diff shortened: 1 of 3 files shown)


llogiq:
Thank you!

@bors r+

PR #10946, opened 2023-06-13
[`match_same_arms`]: don't lint if `non_exhaustive_omitted_patterns`

### LINKED ISSUE #10327: False positive in match_same_arms + non_exhaustive_omitted_patterns
### Summary

Clippy's suggestion makes the code not compile when `non_exhaustive_omitted_patterns` is in use.

### Lint Name

match_same_arms

### Reproducer

```rust
#![feature(non_exhaustive_omitted_patterns_lint)]
#![deny(clippy::match_same_arms)]

use std::sync::atomic::Ordering; // #[non_exhaustive] enum

pub fn f(x: Ordering) {
    match x {
        Ordering::Relaxed => println!("relaxed"),
        Ordering::Release => println!("release"),
        Ordering::Acquire => println!("acquire"),
        Ordering::AcqRel | Ordering::SeqCst => unsupported(x),
        #[deny(non_exhaustive_omitted_patterns)]
        _ => unsupported(x),
    }
}

fn unsupported(x: Ordering) {
    dbg!(x);
}
```

```console
$ cargo clippy
error: this match arm has an identical body to the `_` wildcard arm
  --> src/lib.rs:11:9
   |
11 |         Ordering::AcqRel | Ordering::SeqCst => unsupported(x),
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try removing the arm
   |
   = help: or try changing either arm body
note: `_` wildcard arm here
  --> src/lib.rs:13:9
   |
13 |         _ => unsupported(x),
   |         ^^^^^^^^^^^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#match_same_arms
```

The original code compiles. Clippy's suggested code (remove line 11) does not compile.

```console
error: some variants are not matched explicitly
  --> src/lib.rs:13:9
   |
13 |         _ => unsupported(x),
   |         ^ patterns `std::sync::atomic::Ordering::AcqRel` and `std::sync::atomic::Ordering::SeqCst` not covered
   |
   = help: ensure that all variants are matched explicitly by adding the suggested match arms
   = note: the matched value is of type `std::sync::atomic::Ordering` and the `non_exhaustive_omitted_patterns` attribute was found
```

### Version

```text
rustc 1.69.0-nightly (585f3eef2 2023-02-11)
binary: rustc
commit-hash: 585f3eef26f04440bca726c29193af7b4fa90e54
commit-date: 2023-02-11
release: 1.69.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

Centri3:
Fixes #10327

changelog: [`match_same_arms`]: Don't lint if `non_exhaustive_omitted_patterns` is `warn` or `deny`

### DIFF
--- a/clippy_lints/src/matches/match_same_arms.rs
+++ b/clippy_lints/src/matches/match_same_arms.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::source::snippet;
-use clippy_utils::{path_to_local, search_same, SpanlessEq, SpanlessHash};
+use clippy_utils::{is_lint_allowed, path_to_local, search_same, SpanlessEq, SpanlessHash};
 use core::cmp::Ordering;
 use core::iter;
 use core::slice;
@@ -9,6 +9,7 @@ use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
 use rustc_hir::def_id::DefId;
 use rustc_hir::{Arm, Expr, ExprKind, HirId, HirIdMap, HirIdMapEntry, HirIdSet, Pat, PatKind, RangeEnd};
+use rustc_lint::builtin::NON_EXHAUSTIVE_OMITTED_PATTERNS;
 use rustc_lint::LateContext;
 use rustc_middle::ty;
 use rustc_span::Symbol;
@@ -103,17 +104,21 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'_>]) {
     let indexed_arms: Vec<(usize, &Arm<'_>)> = arms.iter().enumerate().collect();
     for (&(i, arm1), &(j, arm2)) in search_same(&indexed_arms, hash, eq) {
         if matches!(arm2.pat.kind, PatKind::Wild) {
-            span_lint_and_then(
-                cx,
-                MATCH_SAME_ARMS,
-                arm1.span,
-                "this match arm has an identical body to the `_` wildcard arm",
-                |diag| {
-                    diag.span_suggestion(arm1.span, "try removing the arm", "", Applicability::MaybeIncorrect)
-                        .help("or try changing either arm body")
-                        .span_note(arm2.span, "`_` wildcard arm here");
-                },
-            );
+            if !cx.tcx.features().non_exhaustive_omitted_patterns_lint
+                || is_lint_allowed(cx, NON_EXHAUSTIVE_OMITTED_PATTERNS, arm2.hir_id)
+            {
+                span_lint_and_then(
+                    cx,
+                    MATCH_SAME_ARMS,
+                    arm1.span,
+                    "this match arm has an identical body to the `_` wildcard arm",
+                    |diag| {
+                        diag.span_suggestion(arm1.span, "try removing the arm", "", Applicability::MaybeIncorrect)
+                            .help("or try changing either arm body")
+                            .span_note(arm2.span, "`_` wildcard arm here");
+                    },
+                );
+            }
         } else {
             let back_block = backwards_blocking_idxs[j];
             let (keep_arm, move_arm) = if back_block < i || (back_block == 0 && forwards_blocking_idxs[i] <= j) {
--- a/tests/ui/match_same_arms_non_exhaustive.rs
+++ b/tests/ui/match_same_arms_non_exhaustive.rs
@@ -1,0 +1,58 @@
+#![feature(non_exhaustive_omitted_patterns_lint)]
+#![warn(clippy::match_same_arms)]
+#![no_main]
+
+use std::sync::atomic::Ordering; // #[non_exhaustive] enum
+
+pub fn f(x: Ordering) {
+    match x {
+        Ordering::Relaxed => println!("relaxed"),
+        Ordering::Release => println!("release"),
+        Ordering::Acquire => println!("acquire"),
+        Ordering::AcqRel | Ordering::SeqCst => panic!(),
+        #[deny(non_exhaustive_omitted_patterns)]
+        _ => panic!(),
+    }
+}
+
+mod f {
+    #![deny(non_exhaustive_omitted_patterns)]
+
+    use super::*;
+
+    pub fn f(x: Ordering) {
+        match x {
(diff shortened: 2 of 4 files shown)


xFrednet:
Hey @blyxyas, do you want to take a look at this PR? :)

blyxyas:
Yeehaw :cowboy_hat_face: (Yes)

tests/ui/match_same_arms_non_exhaustive.rs
blyxyas, line 1: Instead of cloning the old test, how about just making a new `match_same_arms_non_exhaustive.rs` test with a global `#![deny_non_exhaustive_ommited_patterns]` and putting all tests about this particular bug in there?

`TESTNAME` just checks for tests starting with it, so `TESTNAME="match_same_arms" cargo uitest` would still test both files.

Also, if the files aren't duplicate, tester's machines don't have to do as much work :D

blyxyas (changes requested):
Other than that, everything's good, thanks!
(What a hard PR to review)

tests/ui/match_same_arms_non_exhaustive.rs
Centri3: I'm fine with that, that was just out of laziness :)

Centri3: I merged the two files now. Only line numbers changed

clippy_lints/src/matches/match_same_arms.rs
```diff
@@ -102,45 +103,47 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'_>]) {
...
-                    diag.span_suggestion(arm1.span, "try removing the arm", "", Applicability::MaybeIncorrect)
-                        .help("or try changing either arm body")
-                        .span_note(arm2.span, "`_` wildcard arm here");
-                },
-            );
-        } else {
-            let back_block = backwards_blocking_idxs[j];
-            let (keep_arm, move_arm) = if back_block < i || (back_block == 0 && forwards_blocking_idxs[i] <= j) {
-                (arm1, arm2)
+        if is_lint_allowed(cx, NON_EXHAUSTIVE_OMITTED_PATTERNS, arm2.hir_id) {
```
xFrednet, line 106: The `non_exhaustive_omitted_patterns` lint is still unstable, which means that this check will fail in almost every project unless it has turned on the specific feature.

You can check if the feature is enabled with `cx.tcx.features().non_exhaustive_omitted_patterns_lint`. In this case, I think it would be good to also add a test file, without the feature, to make sure it behaves like normal in those cases.

I think we should also add a note about this, to the lint documentation. Otherwise, it could get confusing why the lint doesn't trigger in some cases

clippy_lints/src/matches/match_same_arms.rs
```diff
@@ -102,45 +103,47 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'_>]) {
...
+                    cx,
+                    MATCH_SAME_ARMS,
+                    arm1.span,
+                    "this match arm has an identical body to the `_` wildcard arm",
+                    |diag| {
+                        diag.span_suggestion(arm1.span, "try removing the arm", "", Applicability::MaybeIncorrect)
+                            .help("or try changing either arm body")
+                            .span_note(arm2.span, "`_` wildcard arm here");
+                    },
+                );
```
xFrednet, line 118: I think the `non_exhaustive_omitted_patterns` lint should only gate this lint trigger with wildcards, and not the one below when we actually have meaningful patters :)

xFrednet (commented):
I like the idea, but I think we should make the lint suppression more fine-grained :)

clippy_lints/src/matches/match_same_arms.rs
Centri3: Fixed

clippy_lints/src/matches/match_same_arms.rs
Centri3: Fixed

clippy_lints/src/matches/match_same_arms.rs
```diff
@@ -103,17 +104,21 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'_>]) {
...
-                MATCH_SAME_ARMS,
-                arm1.span,
-                "this match arm has an identical body to the `_` wildcard arm",
-                |diag| {
-                    diag.span_suggestion(arm1.span, "try removing the arm", "", Applicability::MaybeIncorrect)
-                        .help("or try changing either arm body")
-                        .span_note(arm2.span, "`_` wildcard arm here");
-                },
-            );
+            if !cx.tcx.features().non_exhaustive_omitted_patterns_lint
```
blyxyas, line 107: It should probably be a good idea to add a `FIXME` comment here for when `non_exhaustive_omitted_patterns` gets stabilized [(rust#89554)](https://github.com/rust-lang/rust/issues/89554). That would make Clippy crash AFAIK

Note that this should only be removed if Clippy is being compiled for Rust with that feature. So, unless we have  some magic to check which version is Clippy being installed (or some other methods)

Or we can just not have this stress and just use `cx.tcx.features().active(sym::non_exhaustive_omitted_patterns_lint)`

I'm not sure what branch is the best to follow here.

```suggestion
            // FIXME(Centri3): ...
            if !cx.tcx.features().non_exhaustive_omitted_patterns_lint
```

Centri3: I'll do the latter in a bit, but I'm curious then, why is this used so much? (`cx.tcx.features().<field>`) I haven't seen `.active` before.

Jarcho: There's no ICE risk here. What will happen is the stabilization pr will remove the field which will cause clippy to fail to compile. Since clippy is in the rust repo the same pr will also fix clippy by removing condition.

Basically this doesn't need a note in the source code.

xFrednet (approved):

xFrednet:
LGTM, thank you for this nice update :D

@bors r=blyxyas,xFrednet

PR #10948, opened 2023-06-13
[`borrow_as_ptr`]: Ignore temporaries

### LINKED ISSUE #9884: FP borrow_as_ptr
### Summary

.

### Lint Name

borrow_as_ptr

### Reproducer

I tried this code:

```rust
enum Foo {}
impl Drop for Foo {
    fn drop(&mut self) {}
}

fn main() {
    unsafe { std::ptr::read(&1u8 as *const u8 as *const Foo) };
}
```

I saw this happen:
` cargo clippy --fix -- -Wclippy::borrow_as_ptr`
```
warning: failed to automatically apply fixes suggested by rustc to crate `a`

  * src/main.rs

error[E0745]: cannot take address of a temporary
 --> src/main.rs:7:48
  |
7 |     unsafe { std::ptr::read(std::ptr::addr_of!(1u8) as *const Foo) };
  |                                                ^^^ temporary value

For more information about this error, try `rustc --explain E0745`.

warning: borrow as raw pointer
 --> src/main.rs:7:29
  |
7 |     unsafe { std::ptr::read(&1u8 as *const u8 as *const Foo) };
  |                             ^^^^^^^^^^^^^^^^^ help: try: `std::ptr::addr_of!(1u8)`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#borrow_as_ptr
  = note: requested on the command line with `-W clippy::borrow-as-ptr`

warning: `a` (bin "a") generated 1 warning (run `cargo fix --bin "a"` to apply 1 suggestion)
warning: `a` (bin "a" test) generated 1 warning (1 duplicate)
```

### Version

```text
rustc 1.67.0-nightly (b833ad56f 2022-11-18)
binary: rustc
commit-hash: b833ad56f46a0bbe0e8729512812a161e7dae28a
commit-date: 2022-11-18
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

Centri3:
Fixes #9884

changelog: [`borrow_as_ptr`]: Ignore temporaries

### DIFF
--- a/clippy_lints/src/casts/borrow_as_ptr.rs
+++ b/clippy_lints/src/casts/borrow_as_ptr.rs
@@ -1,11 +1,10 @@
-use std::ops::ControlFlow;
-
+use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::is_no_std_crate;
 use clippy_utils::source::snippet_with_context;
-use clippy_utils::{diagnostics::span_lint_and_sugg, visitors::for_each_unconsumed_temporary};
 use rustc_errors::Applicability;
 use rustc_hir::{BorrowKind, Expr, ExprKind, Mutability, Ty, TyKind};
 use rustc_lint::LateContext;
+use rustc_middle::ty::adjustment::Adjust;
 
 use super::BORROW_AS_PTR;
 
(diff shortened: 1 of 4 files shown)


clippy_lints/src/casts/borrow_as_ptr.rs
```diff
@@ -23,6 +25,10 @@ pub(super) fn check<'tcx>(
         };
         let mut app = Applicability::MachineApplicable;
         let snip = snippet_with_context(cx, e.span, cast_expr.span.ctxt(), "..", &mut app).0;
+        // Fix #9884
+        if for_each_unconsumed_temporary(cx, expr, |_| ControlFlow::Break(true)).is_break() {
```
Alexendoo, line 29: We can copy the criteria rustc uses - https://github.com/rust-lang/rust/blob/3ed2a10d173d6c2e0232776af338ca7d080b1cd4/compiler/rustc_hir_typeck/src/expr.rs#L482-L499

Centri3: I was looking for that for quite a while, only finding that utils function. Thank you!

Alexendoo: A good tip is to search for the error message, here "cannot take address of a temporary", and work your way up from that

Centri3: I often do that after initially searching the docs, so it's a bit weird I couldn't find it then

dswij (approved):
LGTM, thanks!

PR #10949, opened 2023-06-14
[`manual_filter_map`]: lint on `matches` and pattern matching

### LINKED ISSUE #8010: Missed `manual_filter_map` with `matches!`
Lint name:
`manual_filter_map`

I tried this code:

```rust
iter.filter(|x| matches!(x, Enum::A(_)))
    .map(|x| if let Enum::A(s) = x { s } else { unreachable!() })
```
[Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=8a10a5c2d8647c8b3f93e8f64448c961)

I expected to see this happen: `manual_filter_map` lint should be triggered.

Instead, this happened: everything's fine for clippy.

### Meta

```
rustc 1.46.0-nightly (f455e46ea 2020-06-20)
binary: rustc
commit-hash: f455e46eae1a227d735091091144601b467e1565
commit-date: 2020-06-20
release: 1.46.0-nightly
LLVM version: 10.0
```

### Additional information
There are many ways to express the _filter_ part:
* `matches!`
* `if let /* ... */ { true } else { false }`
* `match /* ... */ => true, _ => false `
Similarly, for the _map_ part it is possible to use:
* `if let /* ... */ { inner } else { /* diverge */ }`
* `match /* ... */ => inner, _ => /* diverge */ `
* (soon) `let inner = /* ... */ else { /* diverge */ }`

Moreover, I am not considering eventual helper functions implemented for the enum like `fn is_a(&self) -> bool` or `fn unwrap_a(self) -> A`.
As you can see, the number of cases is combinatorial. Therefore it is reasonable to match the patterns for the _filter_ and the _map_ parts independently, and use the results to perform the final detection. On the other hand, I am worried that trying to _generalize_ the detection of `manual_filter_map` in this way could lead to false positives.

y21:
Fixes #8010

Previously this lint only worked specifically for a very limited set of methods on the filter call (`.filter(|opt| opt.is_some())` and `.filter(|res| res.is_ok())`). This PR extends it to also recognize `matches!` in the `filter` and pattern matching with `if let` or `match` in the `map`.

Example:
```rs
enum Enum {
  A(i32),
  B,
}

let _ = [Enum::A(123), Enum::B].into_iter()
  .filter(|x| matches!(x, Enum::A(_)))
  .map(|x| if let Enum::A(s) = x { s } else { unreachable!() });
```
Now suggests:
```diff
-  .filter(|x| matches!(x, Enum::A(_))).map(if let Enum::A(s) = x { s } else { unreachable!() })
+  .filter_map(|x| match x { Enum::A(s) => Some(s), _ => None })
```

Adding this required a somewhat large change in code because it originally seemed to be specifically written with only method calls in the filter in mind, and `matches!` has different behavior in the map, so this new setup should make it possible to support more "generic" cases that need different handling for the filter and map calls.

changelog: [`manual_filter_map`]: lint on `matches` and pattern matching (and some internal refactoring)

### DIFF
--- a/clippy_lints/src/methods/filter_map.rs
+++ b/clippy_lints/src/methods/filter_map.rs
@@ -1,7 +1,9 @@
-use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
+use clippy_utils::macros::{is_panic, root_macro_call};
 use clippy_utils::source::{indent_of, reindent_multiline, snippet};
 use clippy_utils::ty::is_type_diagnostic_item;
-use clippy_utils::{is_trait_method, path_to_local_id, peel_blocks, SpanlessEq};
+use clippy_utils::{higher, is_trait_method, path_to_local_id, peel_blocks, SpanlessEq};
+use hir::{Body, HirId, MatchSource, Pat};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
@@ -10,7 +12,7 @@ use rustc_hir::{Closure, Expr, ExprKind, PatKind, PathSegment, QPath, UnOp};
 use rustc_lint::LateContext;
 use rustc_middle::ty::adjustment::Adjust;
 use rustc_span::source_map::Span;
-use rustc_span::symbol::{sym, Symbol};
+use rustc_span::symbol::{sym, Ident, Symbol};
 use std::borrow::Cow;
 
 use super::MANUAL_FILTER_MAP;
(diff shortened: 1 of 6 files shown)


clippy_lints/src/methods/filter_map.rs
```diff
@@ -50,6 +52,197 @@ fn is_option_filter_map(cx: &LateContext<'_>, filter_arg: &hir::Expr<'_>, map_ar
...
+            } else if cx.tcx.is_diagnostic_item(sym::Result, recv_ty.did())
+                && path.ident.name == sym!(is_ok)
+            {
+                Some(Self::IsOk { receiver })
+            } else {
+                None
+            }
+        } else if let Some(macro_call) = root_macro_call(expr.span)
+            && is_matches_macro(cx, macro_call.def_id)
+            && let ExprKind::Match(scrutinee, arms, _) = expr.kind
```
Alexendoo, line 232: This can be

```suggestion
            && let ExprKind::Match(scrutinee, [arm, _], _) = expr.kind
```

The `_ => false` arm must come second

clippy_lints/src/methods/filter_map.rs
```diff
@@ -50,6 +52,197 @@ fn is_option_filter_map(cx: &LateContext<'_>, filter_arg: &hir::Expr<'_>, map_ar
...
+                            // check if the other arm has pattern `_` and this arm references the local (modulo blocks)
+                            if let PatKind::Wild = other_arm.pat.kind {
+                                expr_uses_local(arm.pat, arm.body.peel_blocks())
+                            } else {
+                                None
+                            }
+                        };
+                        let (else_arm, (ident, span)) = is_variant_arm(arm1, arm2)
+                            .map(|data| (arm2, data))
+                            .or_else(|| is_variant_arm(arm2, arm1).map(|data| (arm1, data)))?;
```
Alexendoo, line 193: We don't need to test both orientations, the `_` arm must be the second one because

```rust
match x {
    _ => {},
//  ^ this arm always matches
    None => {},
//  ^^^^ unreachable
}
```

clippy_lints/src/methods/filter_map.rs
```diff
@@ -50,6 +52,197 @@ fn is_option_filter_map(cx: &LateContext<'_>, filter_arg: &hir::Expr<'_>, map_ar
...
+                }
+            },
+        }
+    }
+
+    fn hir(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, filter_param_id: HirId) -> Option<Self> {
+        if let ExprKind::MethodCall(path, receiver, [], _) = expr.kind
+            && let Some(recv_ty) = cx.typeck_results().expr_ty(receiver).peel_refs().ty_adt_def()
+            // TODO: we intentionally (?) do not check if `receiver` is a path expression, but maybe we should..?
+            // .filter(|x| effect(x).is_some()).map(|x| effect(x).unwrap()) to filter_map can have different behavior
```
Alexendoo, line 217: True, we could go with checking [`.can_have_side_effects()`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/struct.Expr.html#method.can_have_side_effects) and if so mention it could change the behaviour in a note and adjust the applicability to `MaybeIncorrect`

clippy_lints/src/methods/filter_map.rs
```diff
@@ -50,6 +52,197 @@ fn is_option_filter_map(cx: &LateContext<'_>, filter_arg: &hir::Expr<'_>, map_ar
...
+        filter_param_id: HirId,
+        is_filter_param_ref: bool,
+    ) -> Option<CheckResult<'tcx>> {
+        match *self {
+            OffendingFilterExpr::IsSome { receiver } | OffendingFilterExpr::IsOk { receiver } => {
+                // check if closure ends with expect() or unwrap()
+                if let ExprKind::MethodCall(seg, map_arg, ..) = map_body.value.kind
+                    && matches!(seg.ident.name, sym::expect | sym::unwrap | sym::unwrap_or)
+                    // .map(|y| f(y).copied().unwrap())
+                    //                     ~~~~
```
Alexendoo, line 117: The `~~~~` needs realigning

clippy_lints/src/methods/filter_map.rs
```diff
@@ -50,6 +52,197 @@ fn is_option_filter_map(cx: &LateContext<'_>, filter_arg: &hir::Expr<'_>, map_ar
...
+                // OR
+                // `match enum {   Variant       (v) => v,      _ => unreachable!() }`
+                //        ^^^^     ^^^^^^^        ^                  ^^^^^^^^^^^^^^
+                //     scrutinee  variant_span  variant_ident        else_
+                let (scrutinee, else_, variant_ident, variant_span) =
+                    match higher::IfLetOrMatch::parse(cx, map_body.value) {
+                    Some(higher::IfLetOrMatch::IfLet(sc, pat, then, Some(else_)))
+                        if let Some((ident, span)) = expr_uses_local(pat, then) => {
+                        (sc, else_, ident, span)
+                    },
```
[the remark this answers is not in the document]
y21: it's unfortunate that rustfmt completely ignores code inside macros (`if_chain!` specifically) and let chains... I'm honestly not entirely sure how to format the match here. The line would be >120 columns. Should it just stay on the same line or how is this meant to be formatted?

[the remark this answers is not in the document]
Alexendoo: I think here it's because of the `if let` in the match guard, so you might be able to delete the `let`, rustfmt, add the `let` back

clippy_lints/src/methods/filter_map.rs
```diff
@@ -50,6 +52,218 @@ fn is_option_filter_map(cx: &LateContext<'_>, filter_arg: &hir::Expr<'_>, map_ar
...
+            // we still want to lint if the expression isn't a path to a local variable and possibly contains 
+            // side effects, *but* it can't be machine-applicable then, because that can change the behavior of the program:
+            // .filter(|x| effect(x).is_some()).map(|x| effect(x).unwrap())
+            // vs.
+            // .filter_map(|x| effect(x))
+            // 
+            // the latter only calls `effect` once
+
+            let non_local_expr_span = (!path_to_local_id(receiver, filter_param_id)
+                && receiver.can_have_side_effects())
```
Alexendoo, line 238: I think we can drop the `path_to_local_id` part, unless I'm missing something anything that doesn't contain side effects would be fine as `MachineApplicable`

y21: ah, true, I was concerned that without the `path_to_local_id` check it would lead to this FP:
```rs
let a = None;
[].into_iter().filter(|x| a.is_some()).map(|x| x.unwrap());
```
that is, accessing a different local in the `filter`, *but* I just remembered that that case is already handled in a different place:

https://github.com/rust-lang/rust-clippy/blob/eee3112dc343fc52d2221bffa4d832c9c27df054/clippy_lints/src/methods/filter_map.rs#L149-L151

Alexendoo (commented):
Whoops, forgot to submit these 👀

image

y21:
oops, forgot to bless the tests

Alexendoo:
Thanks, looks good!

@bors r+

PR #10951, opened 2023-06-14
new lint [`single_call_fn`]

### LINKED ISSUE #10861: `single_call_function`
### What it does

Checks for functions which are only called in 1 location.

Enabling this lint can guard against segmentation which can occur stealthily over a long period of development, when a function is split purposefully for testing or readability this lint can be allowed for these specific cases.

### Advantage

Functions that are only called once, may be better suited as scopes or in-lined.

Premature abstraction can occur in development leading to segmentation with many small functions that are only called once, this can introduce significant additional complexity and indirection that makes the code-base less readable.

### Drawbacks

It can be useful to split a large function into sub-functions even when these are only called once each, this can improve test-ability and readability. This is often done purposefully.

### Example

```rust
fn main() {
    one();
}
fn one() {
    // [...]
}
```

Could be written as:

```rust
fn main() {
    // [...]
}
```

Centri3:
Closes #10861

changelog: New lint [`single_call_fn`]

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -287,6 +287,7 @@ mod semicolon_if_nothing_returned;
 mod serde_api;
 mod shadow;
 mod significant_drop_tightening;
+mod single_call_fn;
 mod single_char_lifetime_names;
 mod single_component_path_imports;
 mod single_range_in_vec_init;
@@ -1055,6 +1056,12 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(move |_| Box::new(large_stack_frames::LargeStackFrames::new(stack_size_threshold)));
     store.register_late_pass(|_| Box::new(single_range_in_vec_init::SingleRangeInVecInit));
     store.register_late_pass(|_| Box::new(incorrect_impls::IncorrectImpls));
+    store.register_late_pass(move |_| {
+        Box::new(single_call_fn::SingleCallFn {
+            avoid_breaking_exported_api,
+            def_id_to_usage: rustc_data_structures::fx::FxHashMap::default(),
+        })
+    });
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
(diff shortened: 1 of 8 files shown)


matthiaskrgr:
Should we have a test with closures as well?

Centri3:
Good point. It should not lint on those

clippy_lints/src/lib.rs
```diff
@@ -1045,6 +1046,12 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     });
     let stack_size_threshold = conf.stack_size_threshold;
     store.register_late_pass(move |_| Box::new(large_stack_frames::LargeStackFrames::new(stack_size_threshold)));
+    store.register_late_pass(move |_| {
+        Box::new(single_call_fn::SingleCallFn {
+            avoid_breaking_exported_api,
```
giraffate, line 1051: Can we add this lint to the config file comment?
https://github.com/rust-lang/rust-clippy/blob/cda13a8b26d2d739fe6a4541f5773c8f2f1b3a42/clippy_lints/src/utils/conf.rs#L292-L295

clippy_lints/src/single_call_fn.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    fn check_crate_post(&mut self, cx: &LateContext<'tcx>) {
+        let mut v = FnUsageVisitor {
+            cx,
+            def_id_to_usage: &mut self.def_id_to_usage,
+        };
+        cx.tcx.hir().visit_all_item_likes_in_crate(&mut v);
+
+        for usage in self.def_id_to_usage.values() {
+            let fn_span = usage.0;
```
giraffate, line 91:
```suggestion
            let single_call_fn_span = usage.0;
```

clippy_lints/src/single_call_fn.rs
```diff
@@ -0,0 +1,132 @@
...
+            if let [usage] = *usage.1 {
```
giraffate, line 92:
```suggestion
            if let [caller_span] = *usage.1 {
```

giraffate (commented):
Overall looks good, thanks! I made small comments.

matthiaskrgr:
Oh we could also use a test showing that this does not lint on `#[test]` functions

Centri3:
Yes, it should not do that and there's currently no check for if it's `#[test]` or not, so I'm pretty sure it will lint

giraffate:
@bors r+

Thanks!

Centri3:
@giraffate, please approve again :D
(I need to get into a habit of doing that...)

PR #10960, opened 2023-06-15
new lint `null_pointer_optimization`

### LINKED ISSUE #9983: clippy::large_enum_variant should suggest Option<Box<T>> over Box<Option<T>>
### Description

Consider this example:

```rust
struct LargeValue([u8; 201]);

enum Foo {
    Large(Option<LargeValue>),
    Small,
}
```

Running `cargo clippy` gives this warning:

```
warning: large size difference between variants
 --> src/lib.rs:3:1
  |
3 | / enum Foo {
4 | |     Large(Option<LargeValue>),
  | |     ------------------------- the largest variant contains at least 202 bytes
5 | |     Small,
  | |     ----- the second-largest variant carries no data at all
6 | | }
  | |_^ the entire enum is at least 202 bytes
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
  = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields to reduce the total size of the enum
  |
4 |     Large(Box<Option<LargeValue>>),
  |           ~~~~~~~~~~~~~~~~~~~~~~~
```

Boxing an option is inefficient in the case of `None`, since it still needs to allocate the `None` on the heap and it can't take advantage of [null pointer optimization](https://doc.rust-lang.org/std/option/#representation). Instead, the lint should suggest <code>Option&lt;Box&lt;<i>LargeValue</i>&gt;&gt;</code> when the original field is <code>Option&lt;<i>LargeValue</i>&gt;</code>.

### Version

```text
rustc 1.67.0-nightly (a28f3c88e 2022-11-20)
binary: rustc
commit-hash: a28f3c88e50a77bc2a91889241248c4543854e61
commit-date: 2022-11-20
host: x86_64-apple-darwin
release: 1.67.0-nightly
LLVM version: 15.0.4
```

### Additional Labels

_No response_

Centri3:
Although this doesn't close #9983, I think this is good enough and is more general than such a change to `large_enum_variant` :)

changelog: New lint [`null_pointer_optimization`]


Alexendoo:
Sometimes you do need a `&/&mut Option<T>` which you can't get from the `Option<Box<T>>`, it would be worth mentioning that somewhere, in a note or the description at the very least

Centri3:
IIUC, `into_inner` would let you but it's unstable, so. But yes I will add it to the description

Alexendoo:
`Box::into_inner(b)` is [just `*b`](https://doc.rust-lang.org/src/alloc/boxed.rs.html#599)

So technically yeah you can move the `T` out of the box and place it into an `Option` and take a reference to that, but this consumes the box and can't be done if you have a `&/&mut Option<Box<T>>`

Centri3:
We could probably add an additional check for `#[repr(transparent)]` types around `Box` or `NonNull`, they're likely uncommon but it's still guaranteed to be optimized

xFrednet:

PR #10962, opened 2023-06-15
New lint [`ptr_to_temporary`]

### LINKED ISSUE #10959: Lint for returning raw pointers to temporary stack locations that don't outlive the statement that creates them
### What it does

The general pattern of the code this should complain about is something like

```rust
let x = &(some expression that is stored in a temporary variable on the stack) as *const T;
```

Similar with using `addr_of!` instead of `&` and casts, etc.

Concrete examples this would catch are

```rust
let x = &(1 + 2) as *const i32;
let x = &(x as *const i32) as *const *const i32;
```

In both cases the part in the parenthesis is stored in a temporary stack location that is no longer valid after the whole statement.

It should however not catch

```rust
let x = &(*ptr).x as *const T;
let x = &(some_variable) as *const T;
```

### Advantage

Whatever pointer is created there is pointing to no longer valid stack memory, so any usage afterwards will be unsound

### Drawbacks

Theoretically this could cause false positives but the only case I can see where the resulting code is not unsound is if you cast the pointer to an `usize` and do some calculations with it. I don't see how that could lead to any useful results in such a context though.

### Example

See examples above

Centri3:
This is a small one, but very nice as even `miri` doesn't catch this (it catches it if it's used incorrectly, but not created erroneously).
There were a couple cases of this in other tests, I checked them all and they were 100% valid so I think this is (hopefully) bug-free. Still likely needs more tests though, as always :D (I'm pretty bad at them)

Closes #10959

changelog: New lint [`ptr_to_temporary`]

### DIFF
--- a/clippy_lints/src/ptr_to_temporary.rs
+++ b/clippy_lints/src/ptr_to_temporary.rs
@@ -1,0 +1,355 @@
+use clippy_utils::consts::is_promotable;
+use clippy_utils::diagnostics::{span_lint_and_note, span_lint_hir_and_then};
+use clippy_utils::mir::{location_to_node, StatementOrTerminator};
+use rustc_data_structures::fx::FxHashMap;
+use rustc_hir::def_id::{DefId, LocalDefId};
+use rustc_hir::intravisit::FnKind;
+use rustc_hir::{Body, BorrowKind, Expr, ExprKind, FnDecl, HirId, ItemKind, OwnerNode};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::mir::visit::{MutatingUseContext, PlaceContext, Visitor};
+use rustc_middle::mir::{
+    self, BasicBlock, BasicBlockData, CallSource, Local, Location, Place, PlaceRef, ProjectionElem, Rvalue, SourceInfo,
+    StatementKind, TerminatorKind,
+};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::{sym, Span, Symbol};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for raw pointers pointing to temporary values that will **not** be promoted to a
+    /// constant through
+    /// [constant promotion](https://doc.rust-lang.org/stable/reference/destructors.html#constant-promotion).
+    ///
+    /// ### Why is this bad?
+    /// Usage of such a pointer will result in Undefined Behavior, as the pointer will stop
+    /// pointing to valid stack memory once the temporary is dropped.
+    ///
+    /// ### Known problems
+    /// Expects any call to methods named `as_ptr` or `as_mut_ptr` returning a raw pointer to have
+    /// that raw pointer point to data owned by self. Essentially, it will lint all temporary
+    /// `as_ptr` calls even if the pointer doesn't point to the temporary.
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// fn returning_temp() -> *const i32 {
+    ///     let x = 0;
+    ///     &x as *const i32
+    /// }
+    ///
+    /// let px = returning_temp();
+    /// unsafe { *px }; // ⚠️
+    /// let pv = vec![].as_ptr();
+    /// unsafe { *pv }; // ⚠️
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub PTR_TO_TEMPORARY,
+    // TODO: Let's make it warn-by-default for now, and change this to deny-by-default once we know
+    // there are no major FPs
+    suspicious,
+    "disallows obtaining raw pointers to temporary values"
+}
+declare_lint_pass!(PtrToTemporary => [PTR_TO_TEMPORARY]);
+
+impl<'tcx> LateLintPass<'tcx> for PtrToTemporary {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        check_for_returning_raw_ptr(cx, expr);
+    }
+
+    fn check_fn(
+        &mut self,
+        cx: &LateContext<'tcx>,
+        _: FnKind<'_>,
+        _: &FnDecl<'_>,
+        _: &Body<'_>,
+        _: Span,
+        def_id: LocalDefId,
+    ) {
+        let mir = cx.tcx.optimized_mir(def_id);
+
+        // Collect all local assignments in this body. This is faster than continuously passing over the
+        // body every time we want to get the assignments.
+        let mut assignments = LocalAssignmentsVisitor {
+            results: FxHashMap::default(),
+        };
+        assignments.visit_body(mir);
+
+        let mut v = DanglingPtrVisitor {
+            cx,
+            body: mir,
+            results: vec![],
+            local_assignments: assignments.results,
+        };
+        v.visit_body(mir);
+
+        for (span, hir_id, ident) in v.results {
+            // TODO: We need to lint on the call in question instead, so lint attributes work fine. I'm not sure
+            // how to though
+            span_lint_hir_and_then(
+                cx,
+                PTR_TO_TEMPORARY,
+                hir_id,
+                span,
+                &format!("calling `{ident}` on a temporary value"),
+                |diag| {
+                    diag.note(
+                        "usage of this pointer will cause Undefined Behavior as the temporary will be deallocated at \
+                         the end of the statement, yet the pointer will continue pointing to it, resulting in a \
+                         dangling pointer",
+                    );
+                },
+            );
+        }
+    }
+}
+
+/// Check for returning raw pointers to temporaries that are not promoted to a constant
+fn check_for_returning_raw_ptr<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) -> bool {
+    // Get the final return statement if this is a return statement, or don't lint
+    let expr = if let ExprKind::Ret(Some(expr)) = expr.kind {
+        expr
+    } else if let OwnerNode::Item(parent) = cx.tcx.hir().owner(cx.tcx.hir().get_parent_item(expr.hir_id))
+        && let ItemKind::Fn(_, _, body) = parent.kind
+        && let block = cx.tcx.hir().body(body).value
+        && let ExprKind::Block(block, _) = block.kind
+        && let Some(final_block_expr) = block.expr
+        && final_block_expr.hir_id == expr.hir_id
+    {
+        expr
+    } else {
+        return false;
+    };
+
+    if let ExprKind::Cast(cast_expr, _) = expr.kind
+        && let ExprKind::AddrOf(BorrowKind::Ref, _, e) = cast_expr.kind
+        && !is_promotable(cx, e)
+    {
+        span_lint_and_note(
+            cx,
+            PTR_TO_TEMPORARY,
+            expr.span,
+            "returning a raw pointer to a temporary value that cannot be promoted to a constant",
+            None,
+            "usage of this pointer by callers will cause Undefined Behavior as the temporary will be deallocated at \
+             the end of the statement, yet the pointer will continue pointing to it, resulting in a dangling pointer",
+        );
+
+        return true;
+    }
+
+    false
+}
+
+struct LocalAssignmentsVisitor {
+    results: FxHashMap<Local, Vec<Location>>,
+}
+
+impl Visitor<'_> for LocalAssignmentsVisitor {
+    fn visit_place(&mut self, place: &Place<'_>, ctxt: PlaceContext, loc: Location) {
+        if matches!(
+            ctxt,
+            PlaceContext::MutatingUse(
+                MutatingUseContext::Store | MutatingUseContext::Call | MutatingUseContext::Borrow
+            )
+        ) {
+            self.results.entry(place.local).or_insert(vec![]).push(loc);
+        }
+    }
+}
+
+struct DanglingPtrVisitor<'a, 'tcx> {
+    cx: &'a LateContext<'tcx>,
+    body: &'tcx mir::Body<'tcx>,
+    local_assignments: FxHashMap<Local, Vec<Location>>,
+    results: Vec<(Span, HirId, Symbol)>,
+}
+
+impl<'tcx> Visitor<'tcx> for DanglingPtrVisitor<'_, 'tcx> {
+    fn visit_basic_block_data(&mut self, _: BasicBlock, data: &BasicBlockData<'tcx>) {
+        let Self {
+            cx,
+            body,
+            local_assignments,
+            results,
+        } = self;
+
+        if let Some(term) = &data.terminator
+            && let TerminatorKind::Call {
+                func,
+                args,
+                destination,
+                target: Some(target),
+                call_source: CallSource::Normal,
+                ..
+            } = &term.kind
+            && destination.ty(&body.local_decls, cx.tcx).ty.is_unsafe_ptr()
+            && let [recv] = args.as_slice()
+            && let Some(recv) = recv.place()
+            && let Some((def_id, _)) = func.const_fn_def()
+            && let Some(ident) = returns_ptr_to_self(cx, def_id)
+            && let Ok(recv) = traverse_up_until_owned(body, local_assignments, recv)
+        {
+
+            check_for_dangling(
+                body,
+                *target,
+                recv,
+                destination.as_ref(),
+                term.source_info,
+                ident,
+                results,
+            );
+        }
+    }
+}
+
+fn check_for_dangling<'tcx>(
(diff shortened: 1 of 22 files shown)


clippy_lints/src/casts/ptr_to_temporary.rs
```diff
@@ -0,0 +1,34 @@
...
+                .get(base.hir_id)
+                .is_some_and(|x| x.iter().any(|adj| matches!(adj.kind, Adjust::Deref(_))))
+        }) {
+            span_lint_and_help(
+                cx,
+                PTR_TO_TEMPORARY,
+                expr.span,
+                "raw pointer to a temporary value",
+                None,
+                "this causes Undefined Behavior; create a local binding to make it longer lived",
```
y21, line 30: I don't think creating a dangling raw pointer on its own causes undefined behavior, only actually using the pointer in bad ways does (eg also dereferencing it), as it's also possible entirely in safe code? Ofc I'm not an actual reviewer, but I just thought I'd throw that in

Centri3: That's true, that should be reworded to like "usage of this will cause UB"

clippy_lints/src/casts/mod.rs
```diff
@@ -657,6 +658,31 @@ declare_clippy_lint! {
...
+    /// Checks for raw pointers that point to temporary values.
+    ///
+    /// ### Why is this bad?
+    /// Usage of such a pointer will result in Undefined Behavior, as the pointer will stop pointing
+    /// to valid stack memory once the temporary is dropped.
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// let p = &0u32 as *const u32;
+    /// unsafe { p.read() }; // ⚠️
```
Jarcho, line 672: This isn't actually unsafe. `&0u32` creates a `&'static u32`.

Centri3: huh, is that actually the case? Weird, perhaps it would need a more complex example then

Jarcho: More generally, `let x = &e` will extend the lifetime of `e` the same as if it had first been bound to a local.

Centri3: I've created a better, actually correct example.

Jarcho:
I'm still not sure how the lint as written is useful. Blanket banning `&e as *const _` where `e` is some temporary expression is overkill. Using it as an argument is fine so long as the function doesn't store the pointer somewhere to be used after the function returns. When bound to a local as in `let x = &e as *const _`, `e` will have it's lifetime extended to that of `x`, so there's no issue with that either.

asquared31415:
There's still a case of `&0 as *const i32` in the tests that this is linting on when it's completely sound to use.
[example, run with miri](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=00ed7440fd43c94325bd998de24642fc)

I don't recall exactly where we say what does and doesn't get promoted to statics, and whether that's a stable promise, so I'm not sure that this lint can be useful in that regard.  If those rules ever change, this lint is either going to miss too much (preferable, but not great), or error on code that's perfectly sound.

Centri3:
> Using it as an argument is fine so long as the function doesn't store the pointer somewhere

Huh, weird, I've went with the assumption that giving a function a pointer to temporary would cause immediate UB. Guess not.

> and whether that's a stable promise

If it's not, I see no reason why that shouldn't be linted against (as such code changing at will due to an implementation detail is likely not the intended behavior).

Regardless, though; I've done some testing and I think, even if it's UB, it's impossible for it to be reproduced as allocating stack memory always involves subtracting from the stack pointer; in the case of allocating a temporary, then allocating actual locals, it'll never be overwritten because the temporary will be at `[rsp+04h]` and the rest will be at `[rsp+08h]` and so on. The compiler is free to reorder this however it wants but 2 locals should never be at the same memory region. Whether or not there's an edge-case that makes it UB doesn't matter; Passing a pointer to such a temporary will always be correct (with a modern-day compiler of course), even in called functions as they will subtract from the stack pointer again; not add. (miri still reports this as UB though, which is understandable)

[But; returning a temporary, however, is UB and will always be reproduceable](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021), as it'll both free the temporary and add to the stack pointer which can then in the caller overwrite the pointer's contents if you're unlucky. This should be linted even if it's a static, as for the reason stated above. So, I think we should isolate this to returning temporaries.

y21:
Can this also catch `vec![1].as_ptr()` (or, idk if this can be made more general, maybe if a method is called that takes `&self` and returns a raw pointer?), or is this outside the scope for this? This is a case where temporary lifetime extension doesn't happen,  and I could see a lint being very useful there

Centri3:
I'll look into that, that would be very useful

Alexendoo:
> Huh, weird, I've went with the assumption that giving a function a pointer to temporary would cause immediate UB. Guess not.

Ignoring constant promotion and lifetime extension, temporaries [usually](https://doc.rust-lang.org/stable/reference/destructors.html#temporary-scopes) live until the end of the statement, so they'll be dropped after the function has already returned

> called on a temporary that takes `&self` and returns a raw pointer?

[`Allocators`](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#tymethod.alloc) would be a moderately common counter example here, or any other thing that returns a pointer to something that outlives `&self`

Constant promotion *should* be easy enough to ignore, we can throw in an `is_const_evaluatable` check. Lifetime extension [only happens to `let` statements](https://doc.rust-lang.org/stable/reference/destructors.html#temporary-lifetime-extension) so linting non-`let` statements where the pointer escapes the temporary scope would be a good way around that

One thing to be aware of if you're going for returns is block expressions like so https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9bc3a39542f9977e20653e34d50c9f64

In general let's go with whatever MIRI says as the source of truth, if it says it's UB we could lint it, if not we shouldn't

Centri3:
> [Allocators](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#tymethod.alloc) would be a moderately common counter example here, or any other thing that returns a pointer to something that outlives `&self`

Good point, we could probably only lint if the temporary implements `Drop` but that will likely have the same issue, only a subset of those FPs would be solved, that could maybe be a `restriction`/`nursery` lint

> Constant promotion should be easy enough to ignore, we can throw in an `is_const_evaluatable` check. Lifetime extension [only happens to `let` statements](https://doc.rust-lang.org/stable/reference/destructors.html#temporary-lifetime-extension) so linting non-`let` statements where the pointer escapes the temporary scope would be a good way around that

I don't think we should just ignore constant promotion, again, imo, if code depends on an implementation detail to be correct that should still be disallowed

> In general let's go with whatever MIRI says as the source of truth, if it says it's UB we could lint it, if not we shouldn't

I think for the case of returning pointers to temporaries, we should make it more general to "usage of this pointer may cause Undefined Behavior if it isn't promoted to a constant". Even if it isn't UB in some cases, it's definitely not something intended (nor wanted) by the programmer.

Jarcho:
Both [constant promotion](https://doc.rust-lang.org/stable/reference/destructors.html?highlight=promotion#constant-promotion) and [lifetime extension](https://doc.rust-lang.org/stable/reference/destructors.html?highlight=extension#temporary-lifetime-extension) are defined as part of the language. Even if they weren't in the reference, they would de facto be part of the language; far too much code would break because of it.

Centri3:
So, if I understand correctly, the exact criteria for constant promotion are:
- no interior mutability
- no destructors
- no calls to functions that are not `rustc_promotable`
- it will not change the behavior of the code\*

And this is entirely independent of `opt-level` or any other compiler flags (so, `--debug` and `--release` builds will promote the same constants [minus if they're optimized out entirely, of course]). Also, this behavior will not change; whatever's promoted today will still be (or not be) in 20 years.

If all of those are true, then I can see why we shouldn't lint it. It would be entirely pointless. But IMO, if even one of those aren't it should still be linted.

Also,

> code depends on an implementation detail

This was a bad way to put it. It's not an implementation detail. I more so meant it depends on something outside of the programmer's control to be correct; which, if it can change at will, is not something I would deem "correct".

\* Not always true (admittedly, though, this may be a separate issue, not something constant promotion causes). Have a look at what [this code](https://rust.godbolt.org/z/rT8a3E4qx) outputs with `opt-level` 0, 1, 2 and 3. I have no idea why, but with `opt-level=1`, this will be inlined to `true`. Always.

Alexendoo:
Looking at the optimised output of programs containing UB will [not be that useful](https://rust.godbolt.org/z/bjrMfKz3e), because LLVM is free to optimise assuming it can't happen

It can necessarily change the behaviour of the code, many things won't compile without it (which is why it can't be opt level dependent) and addresses of references can be observed at runtime

Centri3:
> because LLVM is free to optimise assuming it can't happen

Which is what makes it so weird since it should prove that it *can* happen, it's as if it optimizes out storing 110 (which is the only major difference between the two).

> many things won't compile without it

I meant code that would compile without it that's now promoted

In this case, that would be something like `&(100 + 10) as *const i32`, both compile fine but if it's suddenly no longer promoted it's dropped before the function returns. The opposite would be `&(100 + *&(10)) as *const i32`, which while 110 is promoted is still UB as it stores a pointer to it on the stack. Though of course, such a change would likely not happen since it would break `&(100 + 10)` as well (afaik).

> and addresses of references can be observed at runtime

Depending on the exact address of something is a *terrible* idea :D the OS is free to allocate the binary into memory wherever it pleases (address space layout randomization), any code that depends on that is incorrect (and honestly, should be linted in `correctness` too; something like `p == &0x100 as *const i32` is *never* going to succeed but compiles anyway as no address can be on the first page of memory)

If there's an exhaustive list of everything required for constant promotion, then I'm happy to implement it that way but I still believe this should be linted, as a configuration option possibly.

Alexendoo:
> it should prove that it _can_ happen

That's thinking about it from the wrong angle, by definition the compiler can assume UB does not happen. That's precisely the reason it can [consider the 110 a dead store](https://rust.godbolt.org/z/djrx6bfvT) - it would be UB to read from a dangling pointer, therefore it isn't read from, therefore there's no reason to store 110 in it. Or something like that.

I think a lint that suggests transforming cases where the promotion is being relied upon for safety to a separate binding would be fine, `{ let x: &'static _ = ...; &x as *const _ }` or w/e, but for this correctness lint it would be nice if we could say "this is almost certainly UB"

Centri3:
I agree. For this lint in particular, let's go with only linting non-promoted returned pointers then have a separate restriction lint (or, of course a configuration option since they're kinda the same lint anyway).

But in regards to the first paragraph, I meant that LLVM has to prove so, in this case the MIR generated is basically identical for that function with both `opt-level=1` and `opt-level=3`, even the IR generated. But yeah the compiler should assume it doesn't because that's up to the programmer to uphold. I'm guessing that's what it does here on `opt-level=2+`.

Centri3:
I've made it operate on the MIR now for `as_ptr` calls and it works great! It seems to have no FPs I know of (though one minor FN I know how to fix), but there are surely some. Still WIP, so does anybody have some edge-cases I can add to the tests?

The reason why MIR works better is mostly because we know when locals are dropped just by looking at the statements/terminator, if there's a `Drop` for the temporary and no `StorageDead` for the pointer we know it outlives it. In the case there's no `Drop` in the basic block, it's either not temporary (which we know is false) or has no drop glue. In which case we need to be careful as the dead-order(?) is arbitrary, I think anyway.

MIR also has no concept of method calls, just function calls, so we can easily lint both `Vec::as_ptr` and `.as_ptr()` the same.

The earlier `'static` problem was easily solvable by getting the function signature and seeing if there are any late bound or static regions. This is likely wrong but it seems to work. The place's type or typeck results cannot be used unfortunately as these are changed to `ReErased`.

clippy_lints/src/ptr_to_temporary.rs
```diff
@@ -0,0 +1,355 @@
...
+        && final_block_expr.hir_id == expr.hir_id
+    {
+        expr
+    } else {
+        return false;
+    };
+
+    if let ExprKind::Cast(cast_expr, _) = expr.kind
+        && let ExprKind::AddrOf(BorrowKind::Ref, _, e) = cast_expr.kind
+        && !is_promotable(cx, e)
```
Centri3, line 124: We should operate on the MIR here, as after a few reborrows(? stuff like `_ = &(*_)`) there should always be a `const _` if it's promoted. Otherwise, it's temporary. This should be a good starting point.

clippy_lints/src/ptr_to_temporary.rs
```diff
@@ -0,0 +1,355 @@
...
+/// data in the next `Drop` terminator, if not this indicates a bug in our heuristic.
+fn traverse_up_until_owned<'tcx>(
+    body: &'tcx mir::Body<'tcx>,
+    local_assignments: &FxHashMap<Local, Vec<Location>>,
+    start: Place<'tcx>,
+) -> Result<PlaceRef<'tcx>, TraverseError> {
+    traverse_up_until_owned_inner(body, local_assignments, start.as_ref(), 0)
+}
+
+fn traverse_up_until_owned_inner<'tcx>(
```
Centri3, line 268: We should use an analysis pass instead, and track where this owned data goes.

cc @Jarcho, do you have any ideas of what "owned" should be defined as? Basically, for `Drop` types, the place that is dropped. We could just use the place that is dropped for these types but a temporary that doesn't implement `Drop` won't have this, despite being freed. Checking the order of `StorageDead`s works, as the temporary cannot be dead before the pointer, but there may be some nuances to it I'm missing.

clippy_lints/src/ptr_to_temporary.rs
```diff
@@ -0,0 +1,355 @@
...
+        None
+    }) {
+        match (dead_local == recv.local, dead_local == ptr.local) {
+            (true, false) => recv_dead = true,
+            (false, true) if recv_dead => {
+                results.push((
+                    source_info.span,
+                    body.source_scopes[source_info.scope]
+                        .local_data
+                        .clone()
```
Centri3, line 247:
```suggestion
                        .as_ref()
```

clippy_lints/src/ptr_to_temporary.rs
```diff
@@ -0,0 +1,355 @@
...
+        && let ItemKind::Fn(_, _, body) = parent.kind
+        && let block = cx.tcx.hir().body(body).value
+        && let ExprKind::Block(block, _) = block.kind
+        && let Some(final_block_expr) = block.expr
+        && final_block_expr.hir_id == expr.hir_id
+    {
+        expr
+    } else {
+        return false;
+    };
```
Alexendoo, line 120: Could be a use for `find_all_ret_expressions`

Jarcho:
Linting uses of `as_ptr` requires tracking values and not just places. Some simple examples of valid things:
```rust
let x = some_vec.as_ptr();
GLOBAL.lock().some_field = some_vec;
return x;
```
This can be detected by noting `some_vec` is never dropped and therefore the pointer is still valid.

```rust
let x = some_vec.as_ptr();
swap(&mut GLOBAL.lock().some_field, &mut some_vec);
return x;
```
This is basically the same, except `some_vec` is still dropped.

Cases where the local isn't used after the pointer is created can still be linted (e.g. `return some_vec.as_ptr()`). Still has some use, but it needs to be very conservative.

GuillaumeGomez:
Ping @Centri3. I'd love very much to have this lint. Do you need help?

Centri3:
I don't have the time nor energy to work on this. I'd love to finish this but I can't remember what needed fixing or anything about this PR for that matter 😅 If you can take a look that'd be much appreciated, however I should be back soon-ish

xFrednet:
I'm closing this according to the last comment. Thank you for the time, you already put into this lint!

@GuillaumeGomez, since you expressed interest in a previous comment, you're welcome to pick this PR up :)

Centri3:
If anybody tries to continue this, it will need a major rewrite. The current way of traversing backwards is... terse, verbose, and it shouldn't be done this way. There's `rustc_mir_dataflow` for MIR analysis like this, so a full rewrite basically, but the main premise of this will carry over at least.

Since `Drop` terminators aren't emitted if there isn't drop glue, we (unfortunately) can't use that which would work great if it was present. If it is (so it has a `Drop` implementation), then we can check IIRC whether a `StorageDead` is beyond it, or smth. Otherwise `StorageDead`s can be in any order afaik and that doesn't mean it's UB, since it should just be a hint to the optimizer?

So yeah, we need a bad heuristic for this lint. Ideally there'd be more info, and maybe we can have smth like `unleash_the_miri_inside_of_you` but for clippy instead in the future? I would love having *more stable* MIR lints...

The way MIRI detects this ofc only works if it's actually used after being dropped (that meaning, an access after a `StorageDead` i.e. being dropped) - so we can't do that ig.

PR #10965, opened 2023-06-16
Make `--explain` subcommand return 1 for missing lints

not-my-profile:
changelog: The `--explain` subcommand now exits with the 1 exit code for missing lints


Alexendoo (approved):

Alexendoo:
Makes sense, thanks!

@bors r+

PR #10970, opened 2023-06-16
new lint: `read_line_without_trim`

y21:
This adds a new lint that checks for calls to `Stdin::read_line` with a reference to a string that is then attempted to parse into an integer type without first trimming it, which is always going to fail at runtime.
This is something that I've seen happen a lot to beginners, because it's easy to run into when following the example of chapter 2 in the book where it shows how to program a guessing game.
It would be nice if we could point beginners to clippy and tell them "let's see what clippy has to say" and have clippy explain to them why it fails 👀 

I think this lint can later be "generalized" to work not just for `Stdin` but also any `BufRead` (which seems to be where the guarantee about the trailing newline comes from) and also, matching/comparing it to a string slice that doesn't end in a newline character (e.g. `input == "foo"` is always going to fail)

changelog: new lint: [`read_line_without_trim`]

### DIFF
--- a/tests/ui/read_line_without_trim.rs
+++ b/tests/ui/read_line_without_trim.rs
@@ -1,0 +1,19 @@
+//@run-rustfix
+
+#![allow(unused)]
+#![warn(clippy::read_line_without_trim)]
+
+fn main() {
+    let mut input = String::new();
+    std::io::stdin().read_line(&mut input).unwrap();
+    input.pop();
+    let _x: i32 = input.parse().unwrap(); // don't trigger here, newline character is popped
+
+    let mut input = String::new();
+    std::io::stdin().read_line(&mut input).unwrap();
+    let _x: i32 = input.parse().unwrap();
+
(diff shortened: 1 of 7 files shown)


tests/ui/read_line_without_trim.rs
```diff
@@ -0,0 +1,19 @@
...
+
+fn main() {
+    let mut input = String::new();
+    std::io::stdin().read_line(&mut input).unwrap();
+    input.pop();
+    let _x: i32 = input.parse().unwrap(); // don't trigger here, newline character is popped
+
+    let mut input = String::new();
+    std::io::stdin().read_line(&mut input).unwrap();
+    let _x: i32 = input.parse().unwrap();
```
giraffate, line 14: Can we add the case for float, bool and uint? Also, can we add the case for the type which don't match to `parse_fails_on_trailing_newline`?

giraffate:
@bors r+

Thanks!

PR #10974, opened 2023-06-17
Fix integration tests

flip1995:
It sees like the `integration` test binary was no longer uploaded. I wonder how it was the successfully "run". First attempt to fix this.

Also updates the artifacts actions to v3.

r? @xFrednet 

changelog: none


flip1995:
This needs more work. I think we have to set the `LD_LIBRARY_PATH` for the integration tests as well. Also if compilation fails for reasons like this, the tests still are marked successful.

PR #10977, opened 2023-06-17
New lint [`local_assigned_single_value`]

### LINKED ISSUE #4016: Lint variable only ever assigned a single value
In both the rust source and the rustfmt source I encountered code of the sort

```rust
let first = false;
for x in y {
    if first {
        first = false;
        // do some logic
    }
    // do some logic
}
```

If a variable never gets a mutable reference taken to and all assignments assign the same value, we should have a correctness lint tell the user they are doing something that makes no sense in any program.

This is probably easiest implemented as an MIR lint

Centri3:
This one was pretty tough. I still don't know much about MIR so I likely overcomplicated some of it but I hope it's at least of decent quality :)

I don't know of any FPs (that I haven't fixed already), but there's tons of FNs, most notably `let (x, y) = (1, 1)` and the like. It's pretty barebones currently,

Closes #4016

changelog: New lint [`local_assigned_single_value`]

### DIFF
--- a/clippy_lints/src/local_assigned_single_value.rs
+++ b/clippy_lints/src/local_assigned_single_value.rs
@@ -3,6 +3,7 @@ use clippy_utils::fn_has_unsatisfiable_preds;
 use clippy_utils::source::snippet_opt;
 use itertools::Itertools;
 use rustc_const_eval::interpret::Scalar;
+use rustc_data_structures::fx::FxHashSet;
 use rustc_hir::def_id::LocalDefId;
 use rustc_hir::{intravisit::FnKind, Body, FnDecl};
 use rustc_index::IndexVec;
@@ -10,7 +11,7 @@ use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::mir::{
     self, interpret::ConstValue, visit::Visitor, Constant, Location, Mutability, Operand, Place, Rvalue,
 };
-use rustc_middle::mir::{AggregateKind, PlaceElem, Terminator, TerminatorKind};
+use rustc_middle::mir::{AggregateKind, CastKind, PlaceElem, Terminator, TerminatorKind};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::source_map::Spanned;
 use rustc_span::Span;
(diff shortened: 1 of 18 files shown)


Centri3:
The implementation is pretty much done now, I may refactor again though. I still see no FPs as it's overly pessimistic, if it can't prove with 100% confidence it's constant it won't lint.

There are still missed cases, though, like casting a C-like enum or getting the field of a constant ADT

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+        // Building MIR for `fn`s with unsatisfiable preds results in ICE.
+        if fn_has_unsatisfiable_preds(cx, def_id.to_def_id()) {
+            return;
+        }
+
+        let mir = cx.tcx.optimized_mir(def_id.to_def_id());
+        let mut v = V {
+            body: mir,
+            cx,
+            map: mir.local_decls.iter().map(|_| LocalUsageValues::default()).collect(),
```
Jarcho, line 61: The map should be initialized to a state such that only user bound mutable variables check the assigned value. i.e. `matches!(l.local_info, LocalInfo::User(_)) && l.mutability == Mutability::Mut`)

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+impl<'a, 'tcx> Visitor<'tcx> for V<'a, 'tcx> {
+    fn visit_assign(&mut self, place: &Place<'tcx>, rvalue: &Rvalue<'tcx>, loc: Location) {
+        let Self { body, cx, map } = self;
+        let Some(stmt) = body.stmt_at(loc).left() else {
+            return;
+        };
+
+        if stmt.source_info.span.from_expansion() {
```
Jarcho, line 138: You can't just ignore assignments from macros. The most permissive thing you can do here is to block linting for any locals which are assigned in macros. Trying to lint macros will be pain if you want to avoid false positives.

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+            return;
+        }
+
+        let usage = &mut map[place.local];
+        let mut spanned = Spanned {
+            node: LocalUsage::Pending,
+            span: stmt.source_info.span,
+        };
+
+        if let Rvalue::Use(operand) = rvalue {
```
Jarcho, line 156: This whole chain should just be a `match`.

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+                && let Constant { literal, .. } = **constant
+                && let Some(ConstValue::Scalar(scalar)) = literal.try_to_value(cx.tcx)
+            {
+                spanned.node = LocalUsage::Scalar(scalar);
+            } else if let Some(place) = operand.place() {
+                if let [base_proj, ..] = place.projection.as_slice()
+                    // Handle `let [x, y] = [1, 1]` and `let (x, y) = (1, 1)`
+                    && matches!(base_proj, PlaceElem::Field(..) | PlaceElem::Index(..))
+                {
+                    spanned.node = LocalUsage::DependsOnWithProj(place.local, base_proj);
```
Jarcho, line 167: This is only ok if the source projection is not mutable.

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+                } else {
+                    // It seems sometimes a local's just moved, no projections, so let's make sure we
+                    // don't set `assigned_non_const` to true for these cases
+                    spanned.node = LocalUsage::DependsOn(place.local);
```
Jarcho, line 171: This is only ok if the source projection is not mutable.

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+                && let Constant { literal, .. } = **constant
+                && let Some(ConstValue::Scalar(scalar)) = literal.try_to_value(cx.tcx)
+            {
+                spanned.node = LocalUsage::Scalar(scalar);
+            } else if let Operand::Move(place) = operand {
+                if let [base_proj, ..] = place.projection.as_slice()
+                    && matches!(base_proj, PlaceElem::Field(..) | PlaceElem::Index(..))
+                {
+                    // Probably unnecessary
+                    spanned.node = LocalUsage::DependsOnWithProj(place.local, base_proj);
```
Jarcho, line 198: This is only ok if the source projection is not mutable.

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+                } else {
+                    // Handle casts from enum discriminants
+                    spanned.node = LocalUsage::DependsOn(place.local);
```
Jarcho, line 201: This is only ok if the source projection is not mutable.

Jarcho (changes requested):
Just did an initial review. Didn't check the evaluation logic yet, but there's a bit of work to do with the current comments.

clippy_lints/src/local_assigned_single_value.rs
Centri3: Well, `local_info` is actually `ClearCrossCrate` now so we can't. It's not available past borrowck and friends and results in an ICE if used in clippy.

We could perhaps use `var_debug_info` or something.

clippy_lints/src/local_assigned_single_value.rs
```diff
@@ -0,0 +1,326 @@
...
+            }
+        }
+    }
+}
+
+type LocalUsageMap<'tcx> = IndexVec<mir::Local, LocalUsageValues<'tcx>>;
+
+/// Holds the data we have for the usage of a local.
+#[derive(Debug, Default)]
+struct LocalUsageValues<'tcx> {
```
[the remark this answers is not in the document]
Centri3: Agreed

Centri3:
Thanks for reviewing! I agree that there's a ton of work to do on this. I'll probably rewrite what you didn't review yet as it's even uglier. Unfortunately I might not be able to get to this as I've been having issues with both my physical and mental health, but I'll get to it in due time ^^

Centri3:
I'm closing this for now as I have no intention to come back to it. If anybody else wants to work on this in the next few months, you're free to.

Some things I feel like should be note to anybody working on this in the future: MIR might not be the best for this. Currently, `let x = A::B as i32` where `A` is `repr(i32)` and has implicit discriminators is lowered to something like `let _ = 1i32 as i32`, which means we have no way of ignoring this as it shouldn't be linted. This can be fixed upstream, however, there's no real way to determine if an assignment comes before another. So something like
```rust
let mut x = 1;
let mut y = 1;
x = y;
y = x;
```
Will either get into a loop or not lint, despite we should lint it. If this operates on the HIR, we can ignore `y = x` when evaluating `x = y`, but: HIR will be tricky. `let (mut x, mut y) = (1, 2)` will need to be evaluated properly, same with
```rust
let (&&&x, [y, z], A { x: _, w, .. }) = (&&&&&&1, [1, 2], A { x: 0, y: 0, z: 0, w: 0 }
```
and anything else. It'll be a very complex lint no matter what.

Issue #10981, opened 2023-06-18
Flip order of s[a..b].as_bytes()


A-Walrus:
### What it does

Suggests replacing slicing a `str` followed by `as_bytes` with the opposite order
```rust
s[a..b].as_bytes()
```
with
```rust
&s.as_bytes()[a..b]
```

### Advantage

- Removes unnecessary utf8 indexing check 
- won't panic if a and b are not on utf8 boundaries

### Drawbacks

Technically possible someone relied on it panicking, but find that unlikely.

### Example

```rust
let s = "Hello World";
let bytes = s[1..3].as_bytes();
```

Could be written as:

```rust
let s = "Hello World";
let bytes = &s.as_bytes()[1..3];
```

A-Walrus:
I'd like to work on this but I'm new to the codebase and might need some guidance.
Want to make sure I'm on the right track... this lint should implement the `LateLintPass` `check_expr` method right?

y21:
This seems like it would need a `LateLintPass`, since you need to know that the type of the expression being indexed is `str` (type information are only available in late lint passes), and `check_expr` sounds right to me. I would implement the `check_expr` method, check if it's a method call to `str::as_bytes` and the receiver of the call is an index expression, with the indexed type being `str`. The [clippy documentation (specifically "common tools")](https://doc.rust-lang.org/clippy/development/common_tools_writing_lints.html) is probably helpful here. Good luck!

PR #10984, opened 2023-06-18
Add slice_as_bytes lint

A-Walrus:
Detects patterns like
```rust
s[1..5].as_bytes();
```
and suggest replacing with
```rust
&s.as_bytes()[1..5];
```

fixes: #10981 
```
changelog: [`slice_as_bytes`]: add new lilnt
```

### DIFF
--- a/clippy_lints/src/methods/slice_as_bytes.rs
+++ b/clippy_lints/src/methods/slice_as_bytes.rs
@@ -1,0 +1,39 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, source::snippet_with_applicability, ty::is_type_lang_item};
+use rustc_errors::Applicability;
+use rustc_hir::{
+    Expr, ExprKind,
+    LangItem::{self, Range, RangeFrom, RangeFull, RangeTo, RangeToInclusive},
+    QPath,
+};
+use rustc_lint::LateContext;
+
+use super::SLICE_AS_BYTES;
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, recv: &Expr<'_>) {
+    if let ExprKind::Index(indexed, index) = recv.kind {
+        if let ExprKind::Struct(QPath::LangItem(Range | RangeFrom | RangeTo | RangeToInclusive | RangeFull, ..), ..) =
+            index.kind
+        {
+            let ty = cx.typeck_results().expr_ty(indexed).peel_refs();
+            let is_str = ty.is_str();
+            let is_string = is_type_lang_item(cx, ty, LangItem::String);
+            if is_str || is_string {
+                let mut applicability = Applicability::MachineApplicable;
+                let stringish = snippet_with_applicability(cx, indexed.span, "..", &mut applicability);
+                let range = snippet_with_applicability(cx, index.span, "..", &mut applicability);
+                let type_name = if is_str { "str" } else { "String" };
+                span_lint_and_sugg(
+                    cx,
+                    SLICE_AS_BYTES,
+                    expr.span,
+                    &(format!(
+                        "slicing a {type_name} before calling `as_bytes` results in needless UTF-8 alignment checks, and has the possiblity of panicking"
+                    )),
+                    "try",
+                    format!("&{stringish}.as_bytes()[{range}]"),
+                    applicability,
+                );
+            }
+        }
+    }
+}
--- a/clippy_lints/src/methods/slice_as_bytes.rs
+++ b/clippy_lints/src/methods/slice_as_bytes.rs
@@ -1,19 +1,13 @@
 use clippy_utils::{diagnostics::span_lint_and_sugg, source::snippet_with_applicability, ty::is_type_lang_item};
 use rustc_errors::Applicability;
-use rustc_hir::{
-    Expr, ExprKind,
-    LangItem::{self, Range, RangeFrom, RangeFull, RangeTo, RangeToInclusive},
-    QPath,
-};
+use rustc_hir::{is_range_literal, Expr, ExprKind, LangItem};
 use rustc_lint::LateContext;
 
 use super::SLICE_AS_BYTES;
 
 pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, recv: &Expr<'_>) {
     if let ExprKind::Index(indexed, index) = recv.kind {
-        if let ExprKind::Struct(QPath::LangItem(Range | RangeFrom | RangeTo | RangeToInclusive | RangeFull, ..), ..) =
-            index.kind
-        {
+        if is_range_literal(index) {
             let ty = cx.typeck_results().expr_ty(indexed).peel_refs();
             let is_str = ty.is_str();
             let is_string = is_type_lang_item(cx, ty, LangItem::String);
--- a/tests/ui/slice_as_bytes.rs
+++ b/tests/ui/slice_as_bytes.rs
@@ -1,0 +1,35 @@
+//@run-rustfix
+#![allow(unused)]
+#![warn(clippy::slice_as_bytes)]
+
+use std::ops::{Index, Range};
+
+struct Foo;
+
+struct Bar;
+
+impl Bar {
+    fn as_bytes(&self) -> &[u8] {
+        &[0, 1, 2, 3]
+    }
+}
+
+impl Index<Range<usize>> for Foo {
+    type Output = Bar;
+
+    fn index(&self, _: Range<usize>) -> &Self::Output {
+        &Bar
+    }
+}
+
+fn main() {
+    let s = "Lorem ipsum";
+    let string: String = "dolor sit amet".to_owned();
+
+    let bytes = s[1..5].as_bytes();
+    let bytes = string[1..].as_bytes();
+    let bytes = "consectetur adipiscing"[..=5].as_bytes();
+
+    let f = Foo;
+    let bytes = f[0..4].as_bytes();
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5205,6 +5205,7 @@ Released 2018-09-13
 [`size_of_in_element_count`]: https://rust-lang.github.io/rust-clippy/master/index.html#size_of_in_element_count
 [`size_of_ref`]: https://rust-lang.github.io/rust-clippy/master/index.html#size_of_ref
 [`skip_while_next`]: https://rust-lang.github.io/rust-clippy/master/index.html#skip_while_next
+[`slice_as_bytes`]: https://rust-lang.github.io/rust-clippy/master/index.html#slice_as_bytes
 [`slow_vector_initialization`]: https://rust-lang.github.io/rust-clippy/master/index.html#slow_vector_initialization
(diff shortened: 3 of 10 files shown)


clippy_lints/src/slice_as_bytes.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// s[1..5].as_bytes();
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let s = "Lorem ipsum";
+    /// &s.as_bytes()[1..5];
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub SLICE_AS_BYTES,
+    nursery,
```
blyxyas, line 24: It should probably not be `nursery`. `nursery` is left for lints that need further discussion / improving before going public. It should be `pedantic` imo

clippy_lints/src/slice_as_bytes.rs
```diff
@@ -0,0 +1,58 @@
...
+                                        if  is_str|| is_string
+                    {
+                        let mut applicability = MachineApplicable;
+                        let stringish = snippet_with_applicability(cx,indexed.span,"..",&mut applicability);
+                        let range = snippet_with_applicability(cx,index.span,"..",&mut applicability);
+                        let sugg = format!("&{stringish}.as_bytes()[{range}]");
+                        let type_name = if is_str {"str"} else {"String"};
+                        let msg = format!("slicing a {type_name} before calling `as_bytes` results in needless UTF-8 alignment checks, and has the possiblity of panicking");
+                        let sp = expr.span;
+                        span_lint_and_sugg(cx,SLICE_AS_BYTES ,sp , &msg, "try",sugg ,applicability);
```
blyxyas, line 51: No need to assign variables for superfluous values. Generally, we only store the suggestion when emitting a lint (if the suggestion is complex, otherwise don't). But this may be just personal taste

```suggestion
                        span_lint_and_sugg(cx,SLICE_AS_BYTES ,expr.span , format!("slicing a {type_name} before calling `as_bytes` results in needless UTF-8 alignment checks, and may panic"), "try",format!(&{stringish}.as_bytes()[{range}]), Applicability::MachineApplicable);
```

(If you apply this, format using `cargo dev fmt` after)

clippy_lints/src/slice_as_bytes.rs
```diff
@@ -0,0 +1,58 @@
...
+    nursery,
+    "slicing a string and immediantly calling as_bytes is less efficient and can lead to panics"
+}
+declare_lint_pass!(SliceAsBytes => [SLICE_AS_BYTES]);
+
+impl<'tcx> LateLintPass<'tcx> for SliceAsBytes {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let ExprKind::MethodCall(path, recv, [], _) = expr.kind && path.ident.name == sym!(as_bytes){
+            if let ExprKind::Index(indexed, index) = recv.kind {
+                use LangItem::{Range,RangeFrom,RangeFull,RangeTo,RangeToInclusive};
```
blyxyas, line 33: Could you move this import to the start of the file?
Rust is a compiled language, so importing after / before something doesn't make any change on runtime.

clippy_lints/src/slice_as_bytes.rs
```diff
@@ -0,0 +1,58 @@
...
+    #[clippy::version = "1.72.0"]
+    pub SLICE_AS_BYTES,
+    nursery,
+    "slicing a string and immediantly calling as_bytes is less efficient and can lead to panics"
+}
+declare_lint_pass!(SliceAsBytes => [SLICE_AS_BYTES]);
+
+impl<'tcx> LateLintPass<'tcx> for SliceAsBytes {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if let ExprKind::MethodCall(path, recv, [], _) = expr.kind && path.ident.name == sym!(as_bytes){
```
blyxyas, line 31: I would probably change this `if` statement order (it's slightly unoptimal). Also, knowing that `as_bytes` always always takes 0 arguments, you can just ignore the third and fourth arguments (using `..`)

```suggestion
        if path.ident.name == sym!(as_bytes) && let ExprKind::MethodCall(path, recv, ..) = expr.kind {
```

clippy_lints/src/slice_as_bytes.rs
```diff
@@ -0,0 +1,58 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, source::snippet_with_applicability, ty::is_type_lang_item};
+use rustc_errors::Applicability::MachineApplicable;
+use rustc_hir::{Expr, ExprKind, LangItem, QPath};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for string slices immediantly followed by `as_bytes`.
```
blyxyas, line 9:
```suggestion
    /// Checks for string slices immediatly followed by `as_bytes`.
```

clippy_lints/src/slice_as_bytes.rs
```diff
@@ -0,0 +1,58 @@
...
+    /// ### Why is this bad?
+    /// This involves doing an unnecessary UTF-8 alignment check which is less efficient, and can cause a panic.
```
blyxyas, line 11:
```suggestion
    /// It involves doing an unnecessary UTF-8 alignment check which is less efficient, and can cause a panic.
```

tests/ui/slice_as_bytes.rs
blyxyas, line 1: Could you also add a test with `&str`, and another one with a struct / type that implements `as_bytes` but isn't `&str` nor `String`

blyxyas (changes requested):
Very good for a first-time contributor, thanks :heart:! Could you consider these changes?
Also, as a side note, the lint should go in the `methods` type. Try putting it there (maybe creating a new branch and copy-pasting your changes to the new files?) or, if it's too complex, just leave it as is (We'll manually re-type all lints in a few weeks).

(`cargo dev new_lint --name=slice_as_bytes --type=methods --category=pedantic`)

clippy_lints/src/slice_as_bytes.rs
y21: Seems like it could also be `perf`, maybe? https://godbolt.org/z/qd8r8x6hW shows that doing this looks much better perf wise (1 check vs 4 checks in the happy path). Though this is a micro optimization, and I'm not sure if this is enough for the perf category, since most of the perf lints have much bigger impact (something like allocating when it's not necessary). It does seem easy enough of a fix and does not really hurt readability so maybe its fine?

clippy_lints/src/slice_as_bytes.rs
y21: i think it's "immediately"  😅 https://dictionary.cambridge.org/dictionary/english/immediately

clippy_lints/src/methods/slice_as_bytes.rs
```diff
@@ -0,0 +1,39 @@
...
+    LangItem::{self, Range, RangeFrom, RangeFull, RangeTo, RangeToInclusive},
+    QPath,
+};
+use rustc_lint::LateContext;
+
+use super::SLICE_AS_BYTES;
+
+pub(super) fn check(cx: &LateContext<'_>, expr: &Expr<'_>, recv: &Expr<'_>) {
+    if let ExprKind::Index(indexed, index) = recv.kind {
+        if let ExprKind::Struct(QPath::LangItem(Range | RangeFrom | RangeTo | RangeToInclusive | RangeFull, ..), ..) =
```
A-Walrus, line 14: There`s also `RangeInclusiveStruct` and `RangeInclusiveNew` not sure if they're relevant or not

A-Walrus:
I believe I've addressed all the comments, please let me know if I missed something :)

clippy_lints/src/methods/slice_as_bytes.rs
blyxyas: [From the documentation](https://doc.rust-lang.org/beta/nightly-rustc/rustc_hir/lang_items/enum.LangItem.html#variant.RangeInclusiveStruct)
> ### `RangeInclusiveStruct`
> The `RangeInclusive` lang item.

It should be added imo (Although, I don't think it will change anything, as `RangeInclusiveStruct` [isn't really used](https://github.com/search?q=repo%3Arust-lang%2Frust%20RangeInclusiveStruct&type=code))

But `RangeInclusiveNew` is the LangItem used for the `impl<idx> RangeInclusive<idx>` for creating a [new RangeInclusive](https://github.com/rust-lang/rust/blob/405130538938e2d63a761b7d8d6b27d26e634c33/library/core/src/ops/range.rs#L369), it shouldn't be added.

Alexendoo: There's https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/fn.is_range_literal.html which would also catch `..=`

ChrisJefferson:
I'm going to be honest, I don't like this lint without clearer warnings that it could break working code.

The comment is this removes code which can "perform a safety check which can panic". If someone wrote their code depending on that safety check, this clippy silently removes it.

I realise this would be a much bigger change (sorry!), but could clippy have some way of annotating changes which change behaviour in this way?

blyxyas:
Checking every use of that same vector would be too expensive performance-wise, so I don't think it's worth it. We can document it in the lint though. Could you provide an example where this check is needed? I feel like using this check would be just moving the panic from one place to another. Thanks! :heart:

A-Walrus:
> Checking every use of that same vector would be too expensive performance-wise, so I don't think it's worth it. We can document it in the lint though.

I'm not sure what you're referring to, Could you please explain?

> Could you provide an example where this check is needed? I feel like using this check would be just moving the panic from one place to another.

Sure, helix-editor/helix#7417. We were providing chunks of bytes from a string to be processed by some other code. The chunks themselves are not necessarily valid strings, ie can be cut in the middle of a char.
Something similar might happen with the `std::io::Write` trait's `write` function which returns how many bytes where written. If someone was writing from a string they might try taking the remaining bytes of the string like `s[bytes_written..].as_bytes()` which would be incorrect.

blyxyas:
> I'm not sure what you're referring to, Could you please explain?

Referring to: "If someone wrote their code depending on that safety check, this clippy [lint] silently removes it." -- @ChrisJefferson 

I was saying that, if we need to check if all uses of a vector `s` when linting this because maybe the suggestion might break code. Checking all uses of `s` would be too expensive performance-wise and a bad idea.

blyxyas:
I think we should do a big lintcheck and see the repercusions of this change. We can always leave it at `MaybeIncorrect`

Jarcho (commented):
Apologies for the long delay. Just a category change, and some minor cleanup are needed.

Jarcho:
Ping @A-Walrus 

@rustbot  author

xFrednet:
Hey this is triage, I'm closing this due to inactivity. If you want to continue this implementation, you're welcome to create a new PR. @A-Walrus, thank you for the time, you already put into this!

PR #10985, opened 2023-06-18
New lint [`trivial_default_constructed_types`]

Centri3:
Will lint on cases like `<&str>::default` or `u32::default`, etc.

changelog: New lint [`trivial_default_constructed_types`]

### DIFF
--- a/clippy_lints/src/trivial_default_constructed_types.rs
+++ b/clippy_lints/src/trivial_default_constructed_types.rs
@@ -1,0 +1,120 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, is_from_proc_macro, is_lang_item_or_ctor, last_path_segment};
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind, LangItem, QPath};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::{
+    lint::in_external_macro,
+    ty::{self, Ty},
+};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::kw;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for types constructed by `default` that really don't need to be.
+    ///
+    /// ### Why is this bad?
+    /// It's harder for the reader to know what the value is, and it's an unnecessary function call.
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// let a = A(Option::default());
+    /// ```
+    /// Use instead:
+    /// ```rust,ignore
+    /// let a = A(None);
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub TRIVIAL_DEFAULT_CONSTRUCTED_TYPES,
+    pedantic,
+    "checks for usage of `Default::default` to construct trivial types"
+}
+declare_lint_pass!(TrivialDefaultConstructedTypes => [TRIVIAL_DEFAULT_CONSTRUCTED_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for TrivialDefaultConstructedTypes {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &Expr<'tcx>) {
+        if !in_external_macro(cx.sess(), expr.span)
+            && let ExprKind::Call(call, _) = expr.kind
+            && let ExprKind::Path(qpath) = call.kind
+            // `last_path_segment` ICEs if we give it a `LangItem`.
+            && !matches!(qpath, QPath::LangItem(..))
+            && last_path_segment(&qpath).ident.name == kw::Default
+        {
+            let ret_ty = cx
+                .typeck_results()
+                .expr_ty(call)
+                .fn_sig(cx.tcx)
+                .output()
+                .skip_binder()
+                .peel_refs();
+            if let Some(default) = default_value(cx, ret_ty) && !is_from_proc_macro(cx, expr) {
+                span_lint_and_sugg(
+                    cx,
+                    TRIVIAL_DEFAULT_CONSTRUCTED_TYPES,
+                    expr.span,
+                    "constructing a trivial type using `default`",
+                    "try",
+                    default.to_string(),
+                    Applicability::MachineApplicable,
+                );
+            } else if let ty::Tuple(fields) = ret_ty.kind()
+                && let Some(fields_default) = fields.iter()
+                    .map(|field| default_value(cx, field))
+                    .collect::<Option<Vec<&'static str>>>()
+                && !is_from_proc_macro(cx, expr)
+            {
+                let default = if fields.len() == 1 {
+                    // Needs trailing comma to be a single-element tuple
+                    fields_default[0].to_owned() + ","
+                } else {
+                    fields_default.join(", ")
+                };
+
+                span_lint_and_sugg(
+                    cx,
+                    TRIVIAL_DEFAULT_CONSTRUCTED_TYPES,
+                    expr.span,
+                    &format!(
+                        "constructing a {} using `default`",
+                        if fields.is_empty() { "unit" } else { "trivial tuple" },
+                    ),
+                    "try",
+                    format!("({default})"),
+                    Applicability::MachineApplicable,
+                );
+            } else if let ty::Array(ty, len) = ret_ty.kind()
+                && let Some(default) = default_value(cx, *ty)
+                && !is_from_proc_macro(cx, expr)
+            {
+                span_lint_and_sugg(
+                    cx,
+                    TRIVIAL_DEFAULT_CONSTRUCTED_TYPES,
+                    expr.span,
+                    "constructing a trivial array using `default`",
+                    "try",
+                    format!("[{default}; {len}]"),
+                    Applicability::MachineApplicable,
+                );
+            }
+        }
+    }
+}
+
+/// Gets the default value of `ty`.
+fn default_value(cx: &LateContext<'_>, ty: Ty<'_>) -> Option<&'static str> {
+    match ty.kind() {
+        ty::Adt(def, _) => {
+            if is_lang_item_or_ctor(cx, def.did(), LangItem::Option) {
+                return Some("None");
+            }
+
+            None
+        },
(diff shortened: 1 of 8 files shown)


clippy_lints/src/trivial_default_constructed_types.rs
```diff
@@ -0,0 +1,120 @@
...
+                span_lint_and_sugg(
+                    cx,
+                    TRIVIAL_DEFAULT_CONSTRUCTED_TYPES,
+                    expr.span,
+                    "constructing a trivial type using `default`",
+                    "try",
+                    default.to_string(),
+                    Applicability::MachineApplicable,
+                );
+            } else if let ty::Tuple(fields) = ret_ty.kind()
```
Jarcho, line 60: Should probably have a limit on the tuple arity. `(0, 0)` is one thing, `("", false, false, 0, 0.0)` is too much.

clippy_lints/src/trivial_default_constructed_types.rs
```diff
@@ -0,0 +1,120 @@
...
+                && let Some(fields_default) = fields.iter()
+                    .map(|field| default_value(cx, field))
+                    .collect::<Option<Vec<&'static str>>>()
+                && !is_from_proc_macro(cx, expr)
+            {
+                let default = if fields.len() == 1 {
```
Jarcho, line 66: nit: `if let [default] = fields_default`

clippy_lints/src/trivial_default_constructed_types.rs
```diff
@@ -0,0 +1,120 @@
...
+declare_lint_pass!(TrivialDefaultConstructedTypes => [TRIVIAL_DEFAULT_CONSTRUCTED_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for TrivialDefaultConstructedTypes {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &Expr<'tcx>) {
+        if !in_external_macro(cx.sess(), expr.span)
+            && let ExprKind::Call(call, _) = expr.kind
+            && let ExprKind::Path(qpath) = call.kind
+            // `last_path_segment` ICEs if we give it a `LangItem`.
+            && !matches!(qpath, QPath::LangItem(..))
+            && last_path_segment(&qpath).ident.name == kw::Default
```
Jarcho, line 41: This doesn't guarantee `Default::default`. You'll need to resolve the path.

clippy_lints/src/trivial_default_constructed_types.rs
```diff
@@ -0,0 +1,120 @@
...
+            if is_lang_item_or_ctor(cx, def.did(), LangItem::Option) {
+                return Some("None");
+            }
+
+            None
+        },
+        ty::Bool => Some("false"),
+        ty::Str => Some(r#""""#),
+        ty::Int(_) | ty::Uint(_) => Some("0"),
+        ty::Float(_) => Some("0.0"),
```
Jarcho, line 116: This might cause an error if the call to default has a qualified type (e.g. `u32::default()`). I would just ignore linting in those cases.

clippy_lints/src/trivial_default_constructed_types.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[clippy::version = "1.72.0"]
+    pub TRIVIAL_DEFAULT_CONSTRUCTED_TYPES,
+    pedantic,
+    "checks for usage of `Default::default` to construct trivial types"
+}
+declare_lint_pass!(TrivialDefaultConstructedTypes => [TRIVIAL_DEFAULT_CONSTRUCTED_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for TrivialDefaultConstructedTypes {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &Expr<'tcx>) {
+        if !in_external_macro(cx.sess(), expr.span)
```
Jarcho, line 36: It would be better to not lint in any macros unless you want to go through a lot of extra effort. A simple example is a macro expanding to `foo(Default::default())` might call different `foo` functions in each expansion.

clippy_lints/src/trivial_default_constructed_types.rs
Centri3: Fine by me

clippy_lints/src/trivial_default_constructed_types.rs
Centri3: I believe in those cases type inference would do its job (like `struct A { x: 0 }` where `x` is `u32`), but I think we can add the suffix to it

clippy_lints/src/trivial_default_constructed_types.rs
Centri3: Thanks!

(need to remember to use patterns more often :sweat_smile:)

clippy_lints/src/trivial_default_constructed_types.rs
Centri3: I should not code while half asleep ^^

clippy_lints/src/trivial_default_constructed_types.rs
Jarcho: The suffix will be needed when generics are involved. It's also a problem when type inference can't pick a type, in which case `i32` is assumed.

Jarcho: It just occurred to me that the same thing can happen to `Option` as well. You can check if the path to see if it has a type parameter for `Option`.

Centri3: Well, I've made it always add the suffix, does that work? ^^

Centri3: > It just occurred to me that the same thing can happen to `Option` as well. You can check the path to see if it has a type parameter for `Option`.

Oh hmm.. Let me see then

Jarcho: Always adding the suffix is fine.

clippy_lints/src/trivial_default_constructed_types.rs
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+    }
+}
+
+/// Gets the default value of `ty`.
+fn default_value(cx: &LateContext<'_>, ty: Ty<'_>) -> Option<Cow<'static, str>> {
+    match ty.kind() {
+        ty::Adt(def, substs) if let [subst] = substs.as_slice() => {
+            is_lang_item_or_ctor(cx, def.did(), LangItem::Option).then(|| format!("None::<{subst}>").into())
```
Jarcho, line 103: Adding the type argument to `None` when it's not needed isn't really an improvement. `None::<u32>` is ok. `None::<Vec<String>>` is not great. `None::<Arc<Mutex<SomeStruct<String>>>>` is terrible. None of them are better than `Option::default()`.

Centri3: We would likely need to use the `QPath` instead of `Ty` to fix that, but this would ignore type aliases and the like

Centri3: I think the easiest way to fix this is to just make it `HasPlaceholders` in the `None` case, and use `::<_>`.

Jarcho: You bring up a fun issue with type aliases. Numeric literals can't just use type suffixes if the required type is an alias since it could be configuration dependent.

With all the edge cases I don't think there's any way to avoid parsing the type part of the path. Only linting when either the type is missing (e.g. `Default::default()`), or the type is a single identifier matching the expected type's name (e.g. `Option::default()`) is fine. If somebody does something stupid like `type u32 = i32;` then the lint can not trigger. `type MyInt = i32;` is probably better not linted as well.

Centri3: That's a good point, so parsing the path is necessary then. I believe we can use `.into()` for type aliases in particular, or `<name>::from(x)` perhaps (which is usually what I use for creating type aliases). But we probably shouldn't lint them yeah

Centri3: I'm unfortunately not sure how this is supposed to both avoid type aliases and lint `Default::default`. We have to use a ty afaik in that case so we end up skipping type aliases. I have it working for everything else but it's basically useless for arrays, slices, tuples etc if `Default::default` isn't linted

Jarcho: This is easier to do now with `expr_use_ctxt`. You'll have to add a function to get the defined type as a HIR type, but that's mostly copying the existing function. This won't work for anything defined in another crate unfortunately.

Centri3: Honestly I feel like at this point it'd just be easier to store where it's a type alias directly in the `Ty`'s info 😅 It can probably be changed upstream rather easily

I may look into that...

Jarcho: You would still need to take type inference into account, so you would still have to look up the type as it was written either way.

Centri3: Indeed, but at least we'd be able to ignore type aliases more easily. Not a big deal for `u32::default` where `u32` is `type u32 = i32` but `Default::default` is currently pretty annoying

Centri3: I think, for now, I'll change this to lint type aliases for `Default::default` and ignoring them on `u32::default` using the `Res`. One day, once `lazy_type_alias` is stable (and hopefully that normalization is delayed until after typeck :sweat_smile:) then we can ignore them entirely.

xFrednet:

Centri3:
Triaging some of my PRs in case someone wants to pick them up, this one is pretty close to being done I think. The ergonomics are good, It doesn't mess with anything it knows it can't change, and also tries to omit generic parameters, etc.

Mostly, I think we should extract this weird `HIR` + `Ty` abstraction I conjured and maybe also wait until `lazy_type_alias` is stabilized (*if it is*), or just live with the FPs - I feel like they wouldn't be that common, similar to other lints that operate on `Ty`.

If nobody else does, I'll definitely finish this at some point in the future

PR #10987, opened 2023-06-18
new lint: `type_id_on_box`

### LINKED ISSUE #7687: New Lint: Box<dyn Any>::type_id()
### What it does

Detect calls to `Box<dyn Any>::type_id()` 

The intent is likely to get the `TypeId` of the boxed value, but this returns the `TypeId` of `Box<dyn Any>`.
Using `TypeId::of::<Box<dyn Any>>()` is much clearer if that is intended.

### Categories (optional)

- Kind: suspicious

### Drawbacks

None

### Example

[play ground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=74bd5795547270da0f40fc1ae5b28ce4)
```rust
use std::any::{Any,TypeId};

fn main() {
    let any_box:Box<dyn Any>=Box::new(0usize);
    dbg!(
        any_box.type_id(),
        TypeId::of::<usize>(),
        TypeId::of::<Box<dyn Any>>(),
        (*any_box).type_id(),
    );
}
```

y21:
Closes #7687.

A new lint that detects calling `.type_id()` on `Box<dyn Any>` (and not on the underlying `dyn Any`), which can make up for some pretty confusing bugs! 

changelog: new lint: [`type_id_on_box`]


tests/ui/type_id_on_box.rs
```diff
@@ -0,0 +1,20 @@
+//@run-rustfix
+
+#![warn(clippy::type_id_on_box)]
+
+use std::any::{Any, TypeId};
```
llogiq, line 5: I'd like to see a test where the type is behind a type alias, just to be sure we are  privy to possible interactions.

llogiq (commented):
Thank you, this already looks pretty good. I'd just like some more testing.

tests/ui/type_id_on_box.rs
y21: I added a few more tests. Not sure if this is how you meant it though

clippy_lints/src/methods/type_id_on_box.rs
```diff
@@ -0,0 +1,64 @@
...
+use rustc_errors::Applicability;
+use rustc_hir::Expr;
+use rustc_lint::LateContext;
+use rustc_middle::ty::adjustment::Adjust;
+use rustc_middle::ty::adjustment::Adjustment;
+use rustc_middle::ty::Ty;
+use rustc_middle::ty::{self, ExistentialPredicate};
+use rustc_span::{sym, Span};
+
+fn is_dyn_trait(cx: &LateContext<'_>, ty: Ty<'_>) -> bool {
```
Nilstrieb, line 13:
```suggestion
fn is_dyn_any(cx: &LateContext<'_>, ty: Ty<'_>) -> bool {
```

llogiq:
Thank you!

@bors r+

PR #10988, opened 2023-06-19
New lint [`let_else_on_result_ok`]

### LINKED ISSUE #9792: Avoid using `let else` for `Result`
### What it does

source: https://twitter.com/nick_r_cameron/status/1588205358803259392

As suggested in the OP, using `let-else` syntax for `Result` type would be a risk of overlooking proper handling of `Err`.
Also, refactoring, such as replacing `Option` types with `Result` is common thing.
This kind of work is sometimes overlooked, so it is worthwhile to cover in lint.

This lint will help you to notice when `let-else` is applied to the `Result` type.

### Lint Name

`let_else_to_result`

### Category

suspicious

### Advantage

It detects potential risk of missing error handling

### Drawbacks

_No response_

### Example

```rust
enum MyErr { A, B }
fn foo() -> Result<String, MyErr> {
    Ok("foo".into())
}

fn bar() {
    let Ok(foo) = foo() else { return; };
}
```

Could be written as:

```rust
enum MyErr { A, B }
fn foo() -> Result<String, MyErr> {
    Ok("foo".into())
}

fn bar() {
    let foo = match foo() {
        Ok(foo) => foo,
        e => { return; }
    };
}
```

Centri3:
Closes #9792.

changelog: New lint [`let_else_on_result_ok`]


Alexendoo:
You explicitly do have to handle the case where there's an error with let/else already, what's the scenario this lint is expected to benefit?

Centri3:
> what's the scenario this lint is expected to benefit?

This tweet easily explains it best: https://twitter.com/nick_r_cameron/status/1588224853336793089
(though the lint description is incorrect, I wrote that pretty late :D)

Alexendoo:
Looking through [some examples](https://github.com/search?q=lang%3Ars+%2F%5Cs%5Cslet+Ok%5C%28%2F&type=code&p=1) it seems like this would be almost entirely false positives, meaning most of the changes happening would be making code worse to satisfy the lint by adding `.ok()` or similar

Centri3:
We could probably limit it to ones that `return`, rather than `continue`, since almost all of which come from that. Majority of the rest seem to be worth linting (like returning `None` from `Ok`, or constructing an entirely new `Err` that doesn't include the old `Err`'s information)

Centri3:
> e.g. plenty of them are handling `SendError`

I only see a select few that do that. Majority of the others are ignoring an `anyhow::Result` or some custom error type, that does include error information. (I mean, one even ignores a `serde_json::Error` which is very silly)

There really isn't any particularly good reason to remove such information in those cases, at worst you can just print to `stderr` instead

Alexendoo:
I went through a few pages, ignoring any `continue`s/`break`s/`return None` the majority were still false positives which is far too high

Centri3:
That's fair. I do agree that there are many cases where this isn't applicable so I'll see if there's some way to make it more specialized.

Centri3:
So, I've made a couple changes to it. It now only lints if the following are true:
- The containing function returns `Result`
- There's a `return` in the `else` block
- The `Result` type returned from the called function (or just the type of the pattern, really) has a non-zero number of fields in the `Err` variant. That way, something like `SendError` isn't linted

I *think* that should be good now. Is there anything else it should ignore?

xFrednet:

PR #10996, opened 2023-06-20
Lint `mem_forget` if any fields are `Drop`

### LINKED ISSUE #9298: `mem_forget` checks if a type implements `Drop`
### Description

`mem_forget` [currently only triggers][code] if the type passed to `mem::forget` implements `Drop`.

However, if a type doesn't implement `Drop`, then that doesn't mean it lacks destructor code. For instance, `String` and `HashMap` currently don't implement it.

Instead, `core::mem::needs_drop::<T>()` is usually used to check if dropping a `T` invokes a destructor (the compiler-internal equivalent is [`rustc_middle::ty::Ty::needs_drop`][needs_drop]). It is already being used for `(drop|forget)_non_drop` and `let_underscore_drop`, but it is actually intended for optimization purposes only; it might spuriously return true. I don't know if it is the right thing to use here.

[code]: https://github.com/rust-lang/rust-clippy/blob/5721ca9a13562f5168c6ba9f43dc19bb76c6f7bb/clippy_lints/src/mem_forget.rs#L38
[needs_drop]: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.needs_drop

### Version

_No response_

### Additional Labels

@rustbot label: +C-enhancement

Centri3:
Closes #9298
I think this way of doing it (`needs_drop`) should be fine.

---

changelog: Enhancement: [`mem_forget`]: Now lints on types with fields that implement `Drop`
[#10996](https://github.com/rust-lang/rust-clippy/pull/10996)


xFrednet (approved):

tests/ui/mem_forget.stderr
```diff
@@ -4,19 +4,48 @@ error: usage of `mem::forget` on `Drop` type
...
+  --> $DIR/mem_forget.rs:23:5
+   |
+LL |     std::mem::forget(string);
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+note: argument has type `std::string::String`
+  --> $DIR/mem_forget.rs:23:22
+   |
+LL |     std::mem::forget(string);
+   |                      ^^^^^^
```
xFrednet, line 48: I find this output, showing the argument, verbose with little benefit. I believe this now shows up because this PR is reusing the lint pass of an existing lint. Do you think there is a simple way to remove the note? Maybe by using `span_lint_and_then` instead?

It's also fine if you disagree, merging it like this should also be fine :)

Centri3: I'm neutral on it, I think seeing "argument has type `std::vec::Vec`" is helpful but not really if it's one of the fields, maybe we can have something like "whose fields implement `Drop`"?

xFrednet: Sure, I think the not itself is also fine, the span, showcasing the same statement was the part seems verbose to me. Maybe just make the note unspanned in this case?

xFrednet (approved):

xFrednet:
Super swift as always. Thank you!

King bors, would you please accept this beautiful change into the marvelous kingdom of *Master Branch Land*?

xFrednet:
Looks like npm has some problems fetching the remark thingy. Let's try again?

@bors retry

PR #10997, opened 2023-06-20
New lint [`legacy_numeric_constants`]

### LINKED ISSUE #10995: Warn against using legacy std integral constants
### What it does

This lint triggers on usage of legacy integral constants and functions, such as:

- `std::i32::MAX`
- `u64::max_value()`

### Advantage

Enforce a consistent style across the ecosystem.

This lint should probably be placed into the `clippy::style` group.

### Drawbacks

Older projects will experience some churn, but this should be trivially auto-fixable in most cases.

[T-libs-api opposes deprecating this outright](https://github.com/rust-lang/rust/pull/107587#issuecomment-1599053799) so a clippy lint is the next best thing.

### Example

```rust
if x == std::i32::MAX {
```

Could be written using the associated constant as:

```rust
if x == i32::MAX {
```

Centri3:
~~I don't like *one specific part* of its implementation, pretty obvious what it is :) If somebody has a better way to do it, please tell me!~~ Thanks Alex :)
(Still had fun on this though :D)

Closes #10995

changelog: New lint [`legacy_numeric_constants`]


clippy_lints/src/legacy_integral_constants.rs
```diff
@@ -0,0 +1,147 @@
...
+                    .map(|i| ::rustc_span::Symbol::intern(i))
+                    .eq($segments.iter().copied())
+            )||+
+        };
+    }
+
+    // This is awful, but there isn't really much of a better way to do it unfortunately.
+    // FIXME: If possible, let's simplify this.
+    if let [segments @ .., _] = &*cx.get_def_path(def_id)
+        && is_path_in_integral_module![
```
Alexendoo, line 125: Does this work?

```rust
if let [sym::core, sym::u8 | sym::i8 | sym::u16 | etc., last] = .. {
    Some(last)
} else {
    None
}
```

Centri3: Hmm potentially so. I'll see if I can get that to work later

Centri3: Yep it does, thank you!

m-ou-se:
What does this new lint do for this snippet?

```rust
use std::u64;

fn main() {
    let _ = u64::MAX;
}
```

It should suggest simply removing the `use std::u64;` line and not comment on the `u64::MAX` expression, because that will still when `u64` resolves to the type instead of the module.

Centri3:
> What does this new lint do for this snippet?
> 
> ```rust
> use std::u64;
> 
> fn main() {
>     let _ = u64::MAX;
> }
> ```
> 
> It should suggest simply removing the `use std::u64;` line and not comment on the `u64::MAX` expression, because that will still work when `u64` resolves to the type instead of the module.

I think it should just not lint. Ideally it would remove the `use`. Getting the `use` of `u64` would be a bit tricky, IIUC.

Centri3:
I've altered its suggestion on such a case; it now suggests `std::primitive::<etc>` instead, with a note stating that removing the `use` statement will make the above unnecessary

tests/ui/legacy_integral_constants.stderr
```diff
@@ -0,0 +1,72 @@
+error: usage of a legacy integral constant
+  --> $DIR/legacy_integral_constants.rs:10:5
+   |
+LL |     std::f32::EPSILON;
+   |     ^^^^^^^^^^^^^^^^^ help: try: `f32::EPSILON`
+   |
+   = note: `-D clippy::legacy-integral-constants` implied by `-D warnings`
```
pitaj, line 7: The two successive colons in `` help: try: `f32::epsilon` `` looks off to me. I'd also like to see the help message mention "associated constant" somewhere. Something like this:

```
error: usage of a legacy integral constant
  --> $DIR/legacy_integral_constants.rs:10:5
   |
LL |     std::f32::EPSILON;
   |     ^^^^^^^^^^^^^^^^^ help: use the associated constant `f32::EPSILON` instead
   |
```

pitaj:
It's kinda weird to call this `legacy_integral_constants` when it also concerns the floating-point types, which are not integers. Maybe `legacy_numeric_constants` instead?

tests/ui/legacy_integral_constants.stderr
Centri3: That's how inline suggestions look, no real way around it whilst still having it applied unfortunately

pitaj: Alright, in that case the following would be fine

    help: use the associated constant instead: `f32::EPSILON`

pitaj:
> I've altered its suggestion on such a case; it now suggests `std::primitive::<etc>` instead, with a note stating that removing the `use` statement will make the above unnecessary

Not really a huge fan of this, since `std::primitive::i32::MAX` is IMO worse. I think it's fine to not provide a suggestion in this case, and instead just `span_help` to remove `use std::<number>;`.

pitaj:
Thought of another corner case that might require handling separately:

```
// replacing this with just `use u64::MAX` results in an error which ironically suggests `std::u64::MAX`
use std::u64::MAX;

fn main() {
    let _ = MAX;
}
```

tests/ui/legacy_integral_constants.fixed
```diff
@@ -0,0 +1,33 @@
...
+    u8::MAX;
+    u8::MIN;
+    ::std::primitive::u8::MIN;
+    ::std::u8::MIN;
+    ::std::primitive::u8::MIN;
+    std::primitive::u32::MAX;
+    use std::f64;
+    std::primitive::f64::MAX;
+    // Don't lint
+    f32::EPSILON;
```
pitaj, line 26: Might want to make sure that constants in  `std::f64/f32::consts` don't trigger this. And might want to test `use std::f32::consts;` as well just to be safe.

Centri3:
> > I've altered its suggestion on such a case; it now suggests `std::primitive::<etc>` instead, with a note stating that removing the `use` statement will make the above unnecessary
> 
> Not really a huge fan of this, since `std::primitive::i32::MAX` is IMO worse. I think it's fine to not provide a machine-applicable suggestion in this case, and instead just `span_help` to remove `use std::<number>;`.

As said on [zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Getting.20the.20.60use.60.20of.20a.20.60QPath.60), this is tricky. Likely not very plausible. It's either `std::primitive` or no suggestion, really. (Or, of course, finding each `use` statement as said on zulip as well but this will be A: ugly and B: have an ugly suggestion that likely won't even work)

pitaj (changes requested):

tests/ui/legacy_integral_constants.fixed
Centri3: It must be of length 3 so I don't think it will. It also only checks expressions so `use std::f32::consts;` will not be checked. I'll add tests later though

pitaj:
Ah right of getting the span for `span_help` will be an enormous pain, then just providing a dumb help message or note is fine. Something as generic as "remove the `use std::f32` import to use the associated constant instead" would be fine with me.

pitaj:
Actually, is it possible to make the lint trigger on the actual imports like `use std::i32;` directly, and ignore cases where the imports are just used like `i32::MAX`? Would the suggestion to remove the import be easier to implement in that case?

Might want to exclude flats in case someone does something like
```
use std::f32;

fn main() {
    let _ = f32::consts::E;
}
```

Centri3:
What about `std::i32::MAX`, then? If there's no `use` statement then it won't lint it. You *could* lint expressions only if it's an absolute path but that feels hacky imo and would likely lead to an awful suggestion

blyxyas:
>  You could lint expressions only if it's an absolute path but that feels hacky

Maybe this solution is worse, but you could implement a `check_item` for `ItemKind::Use` and store spans to these `use`s. It's a solution I've used a few times and it shouldn't be a performance hit when ignoring allowed lints is implemented.

But that's just a thought.

Centri3:
That's what I was thinking of, but I was worried of cases like
```rust
use std::f64 as _;
// or
pub mod a {
    pub use std::f64;
}

fn a() {
    self::a::f64::MAX;
    use std::f64;
    f64::MAX; // will it always lint the above `use` instead of the `as _` one?
    // etc...
}
```
There are many edgecases to be worried of here, and to ensure that the removed `use` would be the one it originated from would likely require reimplementing resolving, which isn't exactly ideal (we *could* have a single once note that suggests removing them all, or a hidden suggestion, but I don't like either of those ideas all too much)

blyxyas:
>    // as alexendoo also noted, it'll likely be terribly difficult to create an accurate suggestion, so
>    // it might also not be worth the effort at all, like how would:
> use std::{f64, f32::{MAX, self}, borrow::Cow};
    // work?

The `rust-analyzer` team may have some insights here, when using an unknown object, RA may suggest to import it, and the `use` item will be formatted AFAIK.

Alexendoo:
> What about `std::i32::MAX`, then? If there's no `use` statement then it won't lint it. You _could_ lint expressions only if it's an absolute path but that feels hacky imo and would likely lead to an awful suggestion

You can use the current logic where it checks if the first segment is e.g. `u8` to determine if it needs to add `std::primitive`, but instead just return early in that case. The user removing the `use std::u8` will then make it use the associated constant

> The `rust-analyzer` team may have some insights here, when using an unknown object, RA may suggest to import it, and the `use` item will be formatted AFAIK.

The part that makes it difficult is that all the required information to create a good suggestion isn't available in the HIR, it lowers nested `use` items into multiple single ones

pitaj:
> What about `std::i32::MAX`, then? If there's no `use` statement then it won't lint it. You *could* lint expressions only if it's an absolute path but that feels hacky imo and would likely lead to an awful suggestion

Just to be clear, my thought was to keep linting expressions like you already are and only skip ones which look like associated constant usage `i32::MAX`. Anything else like `std::i32::MAX` or `MAX` would still get linted directly with a suggestion.

Again, if providing a machine applicable suggestion is very difficult even doing it this way, I'm okay with just a `span_help` or unspanned help or note instead.

pitaj (changes requested):

Centri3:
> Just to be clear, my thought was to keep linting expressions like you already are and only skip ones which look like associated constant usage i32::MAX. Anything else like std::i32::MAX or MAX would still get linted directly with a suggestion.

Yeah that sounds fine by me

Centri3:
I've made it have a note on every applicable suggestion that some `use` statements may be the culprit, but it doesn't suggest removing them automatically.

clippy_lints/src/legacy_numeric_constants.rs
```diff
@@ -0,0 +1,227 @@
...
+                        if let Some(def_id) = res.opt_def_id()
+                            && let Some(module_name) = is_path_in_integral_module(cx, def_id)
+                            && let _ = use_stmts.insert(module_name, vec![])
+                            && let Some(use_stmts) = use_stmts.get_mut(&module_name)
+                        {
+                            use_stmts.push(item.span);
+                        }
+                    }
+                },
+                UseKind::Glob => glob_use_stmts.push(item.span),
```
Centri3, line 83: Maybe we can use `names_imported_by_glob_use` to filter some of these out? Might not be worth the effort, though

tests/ui/legacy_numeric_constants.rs
```diff
@@ -0,0 +1,52 @@
...
+        ::std::u8::MIN;
+        ::std::primitive::u8::min_value();
+    }
+}
+
+#[clippy::msrv = "1.42.0"]
+fn msrv_too_low() {
+    // FIXME: Why does this lint??
+    std::u32::MAX;
+}
```
xFrednet, line 47: This properly also has a similar reason like `span_lint_hir`. The msrv is tracked by  `extract_msrv_attr!(LateContext);` which does it for every node, but you lint once at the end. Either, you have to do the tracking in the visitor or change the logic to lint in the lint pass

clippy_lints/src/legacy_numeric_constants.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    extract_msrv_attr!(LateContext);
+}
+
+struct V<'a, 'tcx> {
+    cx: &'a LateContext<'tcx>,
+    msrv: &'a Msrv,
+    use_stmts: &'a FxHashMap<Symbol, Vec<Span>>,
+    glob_use_stmts: &'a Vec<Span>,
+}
```
xFrednet, line 113: I'm not sure, if visiting everything again with a Visitor is the best thing to do. I'm guessing, you're doing it this way, to catch cases where the `use std::primitive::u32` is after an item/body?

I can't remember if this has been discussed as part of the previous conversion. But I would guess that just linting on `use std::primitive::u32` should be fine. There might be some uncool cases like `use std::primitive::u32::MAX` but even then, I'd say it's enough to lint that one and state that all uses of `MAX` have to be replaced with `u32::MAX`. Walking all items again will probably have a performance impact and is weird when it comes to msrv tracking and lint emission.

xFrednet (commented):
I like the lint idea, but I'm not too sure about the way it's implemented. I've left some notes. (I've read the discussion a few days ago, so sorry if some of these have already been discussed :sweat_smile: )

clippy_lints/src/legacy_numeric_constants.rs
```diff
@@ -0,0 +1,227 @@
...
+            use_stmts,
+            glob_use_stmts,
+        } = self;
+
+        if !item.span.is_dummy() && let ItemKind::Use(path, kind) = item.kind {
+            match kind {
+                UseKind::Single => {
+                    for res in &path.res {
+                        if let Some(def_id) = res.opt_def_id()
+                            && let Some(module_name) = is_path_in_integral_module(cx, def_id)
```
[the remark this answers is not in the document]
Centri3: Because something like `f32::MAX` where `f32` is `std::f32` won't work. It doesn't include the full path, so anything that isn't just `std::f32::MAX` won't work (which is very rare)

clippy_lints/src/legacy_numeric_constants.rs
Centri3: Yeah we need to do 2 passes to have that work unfortunately, we could lint the `use` specifically if it exists but that might be a bit uglier than just linting the usage

tests/ui/legacy_numeric_constants.rs
Centri3: Yeah it should probably do it like that instead

clippy_lints/src/legacy_numeric_constants.rs
Alexendoo: Linting the `use` seems fine to me since that's the part we want them to remove

Centri3: But we can't really know whether they're using it for `std::f32::MAX` or `std::f32::consts`. Something like
```rust
use std::f32;
use f32::consts;
```
Would be a FP (but not something people use regardless)

Urgau:
Have you considered simply using diagnostics items instead of manually trying to match them ?

Centri3:
They aren't diagnostic items, so we can't unfortunately.

Urgau:
They could be added. I'm sure rust-lang/rust wouldn't mind them, in particular for a lint.

Centri3:
That would require waiting for the next sync in 2 weeks.

xFrednet:
> They could be added. I'm sure rust-lang/rust wouldn't mind them, in particular for a lint.

They wouldn't mind at all. The diagnostic items are exactly intended for this kind of work. Just the sync wait time can be annoying from the Clippy side. I don't think it's a necessity, but definitely a nice idea :)

clippy_lints/src/legacy_numeric_constants.rs
xFrednet: I think this would be an acceptable FP, I can't imagine people really using this. :)

clippy_lints/src/legacy_numeric_constants.rs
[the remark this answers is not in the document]
xFrednet: As mentioned below, I think this would be an acceptable FP. Otherwise, we could consider collecting the IDs of all deprecated modules ones and then only checking that list. That sounds more efficient to me, at least for normal crates.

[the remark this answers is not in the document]
Centri3: This actually wouldn't be a FP, mostly FNs. In the case of `f32::MAX` it wouldn't lint despite looking identical to the correct `f32::MAX` which is *absolutely* not desirable imo. In the case of `std::f32::MAX`, there are many that qualify it like that so I don't think that's acceptable either.

clippy_lints/src/legacy_numeric_constants.rs
Centri3: There's quite a few cases of it. https://github.com/search?q=%2Fuse+f32%3A%3Aconsts%7Cuse+f64%3A%3Aconsts%2F&type=code

pitaj: How about this:

- lint `use std::<int>` directly, but not `std::<float>`
- do not lint `<int>::<CONST>` even if it resolves to `std::<int>::<CONST>`, because it's covered by the previous
- lint all other cases of `std::<int>::<CONST>`: `MAX`, `std::i32::MIN`, etc
- lint all cases of `std::<float>::<CONST>`
- provide an unspanned "remove the import" help or note for cases where `<float>::<CONST>` resolves to `std::<float>::<CONST>`

clippy_lints/src/legacy_numeric_constants.rs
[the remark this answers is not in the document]
xFrednet: Ahh okay, then I misunderstood that part. Thank you for clarifying!

clippy_lints/src/legacy_numeric_constants.rs
xFrednet: I think this is a good idea and should also simplify the implementation :)

Centri3: I don't particularly agree, it's inconsistent with the behavior of `deprecated` which lints on the usage itself which is generally a lot clearer. In the case of `std::<int>` though we should also lint the `use` itself as that's also quasi-deprecated but regardless the usage itself should be linted

Seeing `usage of legacy numeric constants` on `use std::u32::MAX` will likely confuse the programmer if removing that suddenly causes compilation to fail, and they may just silence the lint as a result (adding a single `allow` is easier than manually changing every `MAX` to `u32::MAX`)

This would also make it non-`MachineApplicable` in the case of `std::f32::MAX` and friends, which I don't think is worth the increased simplicity...

pitaj: I think you may have misunderstood the behavior I proposed.

> Seeing `usage of legacy numeric constants` on `use std::u32::MAX` will likely confuse the programmer if removing that suddenly causes compilation to fail, and they may just silence the lint as a result (adding a single `allow` is easier than manually changing every `MAX` to `u32::MAX`)

Under the behavior I suggested, `use std::u32::MAX;` would NOT trigger the lint. `use std::u32;` would.

> This would also make it non-`MachineApplicable` in the case of `std::f32::MAX` and friends, which I don't think is worth the increased simplicity...

The behavior for `std::f32::MAX` would not change at all under my proposal.

Centri3: I see.
But I still don't think we should go with this. In the case of `f32::MAX` or `f64::MAX` it'll result in a much less helpful message, just "remove the import" which is ambiguous; which import? It looks identical to `f32::MAX`, just saying there's some import for `std::f32` *somewhere* isn't all too helpful.

What this simplifies is removing the `use` collection; however, the rest of it is only made more complicated. If the "remove the import" portion is improved, nothing has been simplified.

> But the pattern `i32::MAX` is not the problem, it's actually the prior `use std::i32;` that's the problem because it was made redundant.

And in the cases where they're identical, It should still lint the usage but not give a suggestion, only listing the potentially bad imports. Then, since `std::i32` is quasi-deprecated, that can be linted as well (we could also omit the bad imports in this case, since it'll be linted). In the case of `f32::MAX` it'd always show the bad imports.

pitaj: > But I still don't think we should go with this. In the case of `f32::MAX` or `f64::MAX` it'll result in a much less helpful message, just "remove the import" which is ambiguous; which import? It looks identical to `f32::MAX`, just saying there's some import for `std::f32` *somewhere* isn't all too helpful.

I think you're overstating how hard it is to find and remove an import. And in the case of floats, it will commonly require more work anyways to resolve the `::consts` issue. Maybe we just shouldn't ever lint `f32::MAX`.

> What this simplifies is removing the `use` collection; however, the rest of it is only made more complicated. If the "remove the import" portion is improved, nothing has been simplified.
> 
> And in the cases where they're identical, It should still lint the usage but not give a suggestion, only listing the potentially bad imports. Then, since `std::i32` is quasi-deprecated, that can be linted as well (we could also omit the bad imports in this case, since it'll be linted). In the case of `f32::MAX` it'd always show the bad imports.

In my opinion, linting `i32::MAX` is purely noise. There's nothing to change with that code. All that needs to be done is remove the redundant import. With current behavior, stderr is longer and more repetitive.

It feels like throwing a warning when someone uses `.try_into()` because they don't need to explicitly import the `TryInto` trait, since it's in the prelude.

Here's how I see it. The current behavior is slower, more complicated, and more repetitive for a slightly more useful warning message. I don't think the tradeoffs are worth it.

Centri3:
Well, I completely forgot about this PR :sweat_smile: I should have this back going again within a few days

tests/ui/legacy_numeric_constants.stderr
```diff
@@ -0,0 +1,171 @@
...
+   |
+   = help: use the associated constant `u8::MIN` instead at its usage
+
+error: importing legacy numeric constants
+  --> $DIR/legacy_numeric_constants.rs:77:9
+   |
+LL |     use std::u32;
+   |         ^^^^^^^^
+   |
+   = help: use the associated constants on `u32` instead at their usage
```
pitaj, line 39: Can we just recommend removing the import altogether? Maybe something like
```
help: remove this import
note: `u32::<CONST>` will then resolve to the respective associated constant
```

Centri3: We can't because somebody could do something stupid like `use crate::u32::MAX` after it, which would then break. It would be nice if we could detect that though

Centri3: Also, we should only show the note if it isn't aliased, like `use std::u32 as u32_consts`

pitaj: Maybe something like "remove this import and use the associated constants on `u32` instead"?

pitaj:
Hey @Centri3 any updates on this? If you're busy I may be able to finish it off.

xFrednet:
`@Centri3` has been busy lately. @blyxyas would you mind taking over this review? You can also bounce it to me, if you don't have the time :D

xFrednet:
I'm closing this PR due to inactivity. The implementation is an excellent start, if anyone is interested in picking this PR up, go ahead. If you have questions, you're welcome to ask them here as well.

@Centri3 Thank you for working on this. If you want to continue your work, you can reopen this PR or open a new one :D

PR #11001, opened 2023-06-21
[`question_mark`]: don't lint inside of `try` block

### LINKED ISSUE #8628: `question_mark` inside `try` blocks
### Summary

The rewrites suggested by `question_mark` do not work correctly inside `try` blocks (unstable with `#![feature(try_blocks)]`), since `?` exits the `try` block rather than the function.

I don't know what the policy is for unstable features, so just close this if it's irrelevant.

### Lint Name

question_mark

### Reproducer

With the following function,
```rust
fn foo1(a: Option<u32>) -> Option<u32> {
	let b: Option<u32> = try {
		if a.is_none() {
			return None;
		}
		32
	};
	b.or(Some(128))
}
```
Clippy suggests rewriting it as 
```rust
fn foo2(a: Option<u32>) -> Option<u32> {
	let b: Option<u32> = try {
		a?;
		32
	};
	b.or(Some(128))
}
```
However, `foo1(None)` gives `None`, while `foo2(None)` gives `Some(128)`, which is not the same.

### Version

```text
rustc 1.61.0-nightly (9f4dc0b4d 2022-03-23)
binary: rustc
commit-hash: 9f4dc0b4db892271cd0dada6e072775b5b5d6b1e
commit-date: 2022-03-23
release: 1.61.0-nightly
LLVM version: 14.0.0
```

### Additional Labels

y21:
Fixes #8628.

Diff looks a bit noisy because I had to move the two functions into an impl, because they now need to access the structs `try_block_depth` field to see if they're inside a try block.

changelog: [`question_mark`]: don't lint inside of `try` block

### DIFF
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -770,7 +770,7 @@ pub fn register_plugins(store: &mut rustc_lint::LintStore, sess: &Session, conf:
     store.register_late_pass(|_| Box::<useless_conversion::UselessConversion>::default());
     store.register_late_pass(|_| Box::new(implicit_hasher::ImplicitHasher));
     store.register_late_pass(|_| Box::new(fallible_impl_from::FallibleImplFrom));
-    store.register_late_pass(|_| Box::new(question_mark::QuestionMark));
+    store.register_late_pass(|_| Box::<question_mark::QuestionMark>::default());
     store.register_late_pass(|_| Box::new(question_mark_used::QuestionMarkUsed));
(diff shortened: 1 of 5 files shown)


Jarcho:
Well that diff was useless. Nowhere near the most useless I've seen.

This doesn't handle nested functions (e.g. `fn f() { try { fn f2() {} }`). You'll have to keep a stack of try block depths for this.

Jarcho:
Thank you. @bors r+

PR #11002, opened 2023-06-21
teach `eager_or_lazy` about panicky arithmetic operations

### LINKED ISSUE #9422: `unnecessary_lazy_evaluations` suggestion causes panic on `bool::then`
### Summary

Clippy does not consider the possibility of arithmetic underflow in the non-lazy suggestion

### Lint Name

unnecessary_lazy_evaluations

### Reproducer

```rust
fn f(x: usize) -> Option<usize>
{
    (x >= 5).then(|| x - 5)
    // (x >= 5).then_some(x - 5)  // clippy suggestion panics
}

fn main() {
    let x = f(3);
    println!("{:?}", x);
}
```

clippy output:
```
warning: unnecessary closure used with `bool::then`
 --> src/main.rs:3:5
  |
3 |     (x >= 5).then(|| x - 5)
  |     ^^^^^^^^^--------------
  |              |
  |              help: use `then_some(..)` instead: `then_some(x - 5)`
  |
  = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations

    Finished dev [unoptimized + debuginfo] target(s) in 0.68s
```

### Version

```text
rustc 1.65.0-nightly (8c6ce6b91 2022-09-02)
binary: rustc
commit-hash: 8c6ce6b91b172f77c795a74bfeaf74b865146b3f
commit-date: 2022-09-02
release: 1.65.0-nightly
LLVM version: 15.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #9814: `bool::then` -> `bool::then_some` within `unnecessary_lazy_eval` can suggest fixes that can panic
### Summary

The `bool::then` -> `bool::then_some` suggestions for `unnecessary_lazy_eval` can be problematic if the underlying body of the closure is not actually trivial. See repro below for a concrete example.

This is especially problematic because `cargo clippy --fix` will auto-fix this, which can inadvertently cause a new panic.

### Lint Name

unnecessary_lazy_eval

### Reproducer

I tried this code:

```rust
let foo : usize = 0;
let option = (foo > 0).then_some(|| foo - 1)
```

I saw this happen:

```
warning: unnecessary closure used with `bool::then`
```
This is _incorrect_, since `bool:then` will panic in a case where `then_some` would not have.

No clippy warning suggested.

### Version

```text
binary: rustc
commit-date: 2022-11-02
host: aarch64-apple-darwin
release: 1.65.0
LLVM version: 15.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #11793: `unnecessary_lazy_evaluations` gives wrong suggestions for then
### Summary

In the attached scenario `unnecessary_lazy_evaluations` is triggered, but it gives a wrong suggestion because the lazy expression relies on the fact that's called only when the number is greater than 0 because otherwise always subtracting 1 would result in an underflow.

### Lint Name

unnecessary_lazy_evaluations

### Reproducer

I tried this code:

```rust
pub fn sub(n: u8) -> Option<u8> {
    (n > 0).then(|| n - 1)
}
```

I saw this happen:

```
warning: unnecessary closure used with `bool::then`
 --> src/lib.rs:4:5
  |
4 |     (n > 0).then(|| n - 1)
  |     ^^^^^^^^--------------
  |             |
  |             help: use `then_some(..)` instead: `then_some(n - 1)`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
  = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default

warning: `playground` (lib) generated 1 warning (run `cargo clippy --fix --lib -p playground` to apply 1 suggestion)
    Finished dev [unoptimized + debuginfo] target(s) in 0.61s
```

No lint should be emitted 

### Version

```text
rustc 1.73.0 (cc66ad468 2023-10-03)
binary: rustc
commit-hash: cc66ad468955717ab92600c770da8c1601a4ff33
commit-date: 2023-10-03
release: 1.73.0
LLVM version: 17.0.2
```

### Additional Labels

_No response_

y21:
Fixes #9422
Fixes #9814
Fixes #11793

It's a bit sad that we have to do this because arithmetic operations seemed to me like the prime example where a closure would not be necessary, but this has "side effects" (changes behavior when going from lazy to eager) as some of these panic on overflow/underflow if compiled with `-Coverflow-checks` (which is the default in debug mode).
Given the number of backlinks in the mentioned issues, this seems to be a FP that is worth fixing, probably.

changelog: [`unnecessary_lazy_evaluations`]: don't lint if closure has panicky arithmetic operations

### DIFF
--- a/clippy_utils/src/consts.rs
+++ b/clippy_utils/src/consts.rs
@@ -10,7 +10,7 @@ use rustc_hir::{BinOp, BinOpKind, Block, ConstBlock, Expr, ExprKind, HirId, Item
 use rustc_lexer::tokenize;
 use rustc_lint::LateContext;
 use rustc_middle::mir::interpret::{alloc_range, Scalar};
-use rustc_middle::ty::{self, EarlyBinder, FloatTy, GenericArgsRef, List, ScalarInt, Ty, TyCtxt};
+use rustc_middle::ty::{self, EarlyBinder, FloatTy, GenericArgsRef, IntTy, List, ScalarInt, Ty, TyCtxt, UintTy};
 use rustc_middle::{bug, mir, span_bug};
 use rustc_span::symbol::{Ident, Symbol};
 use rustc_span::SyntaxContext;
@@ -49,6 +49,63 @@ pub enum Constant<'tcx> {
     Ref(Box<Constant<'tcx>>),
     /// A literal with syntax error.
     Err,
+}
+
+trait IntTypeBounds: Sized {
+    type Output: PartialOrd;
+
+    fn min_max(self) -> Option<(Self::Output, Self::Output)>;
+    fn bits(self) -> Self::Output;
+    fn ensure_fits(self, val: Self::Output) -> Option<Self::Output> {
+        let (min, max) = self.min_max()?;
+        (min <= val && val <= max).then_some(val)
+    }
+}
+impl IntTypeBounds for UintTy {
+    type Output = u128;
+    fn min_max(self) -> Option<(Self::Output, Self::Output)> {
+        Some(match self {
+            UintTy::U8 => (u8::MIN.into(), u8::MAX.into()),
+            UintTy::U16 => (u16::MIN.into(), u16::MAX.into()),
+            UintTy::U32 => (u32::MIN.into(), u32::MAX.into()),
+            UintTy::U64 => (u64::MIN.into(), u64::MAX.into()),
+            UintTy::U128 => (u128::MIN, u128::MAX),
+            UintTy::Usize => (usize::MIN.try_into().ok()?, usize::MAX.try_into().ok()?),
+        })
+    }
+    fn bits(self) -> Self::Output {
+        match self {
+            UintTy::U8 => 8,
+            UintTy::U16 => 16,
+            UintTy::U32 => 32,
+            UintTy::U64 => 64,
+            UintTy::U128 => 128,
+            UintTy::Usize => usize::BITS.into(),
+        }
+    }
+}
+impl IntTypeBounds for IntTy {
+    type Output = i128;
+    fn min_max(self) -> Option<(Self::Output, Self::Output)> {
+        Some(match self {
+            IntTy::I8 => (i8::MIN.into(), i8::MAX.into()),
+            IntTy::I16 => (i16::MIN.into(), i16::MAX.into()),
+            IntTy::I32 => (i32::MIN.into(), i32::MAX.into()),
+            IntTy::I64 => (i64::MIN.into(), i64::MAX.into()),
+            IntTy::I128 => (i128::MIN, i128::MAX),
+            IntTy::Isize => (isize::MIN.try_into().ok()?, isize::MAX.try_into().ok()?),
+        })
+    }
+    fn bits(self) -> Self::Output {
+        match self {
+            IntTy::I8 => 8,
+            IntTy::I16 => 16,
+            IntTy::I32 => 32,
+            IntTy::I64 => 64,
+            IntTy::I128 => 128,
+            IntTy::Isize => isize::BITS.into(),
+        }
+    }
 }
 
 impl<'tcx> PartialEq for Constant<'tcx> {
@@ -435,8 +492,15 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
         match *o {
             Int(value) => {
                 let ty::Int(ity) = *ty.kind() else { return None };
+                let (min, _) = ity.min_max()?;
                 // sign extend
                 let value = sext(self.lcx.tcx, value, ity);
+
+                // Applying unary - to the most negative value of any signed integer type panics.
+                if value == min {
+                    return None;
+                }
+
                 let value = value.checked_neg()?;
                 // clear unused bits
                 Some(Int(unsext(self.lcx.tcx, value, ity)))
@@ -572,17 +636,33 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
         match (l, r) {
             (Constant::Int(l), Some(Constant::Int(r))) => match *self.typeck_results.expr_ty_opt(left)?.kind() {
                 ty::Int(ity) => {
+                    let (ty_min_value, _) = ity.min_max()?;
+                    let bits = ity.bits();
                     let l = sext(self.lcx.tcx, l, ity);
                     let r = sext(self.lcx.tcx, r, ity);
+
+                    // Using / or %, where the left-hand argument is the smallest integer of a signed integer type and
+                    // the right-hand argument is -1 always panics, even with overflow-checks disabled
+                    if let BinOpKind::Div | BinOpKind::Rem = op.node
+                        && l == ty_min_value
+                        && r == -1
+                    {
+                        return None;
+                    }
+
                     let zext = |n: i128| Constant::Int(unsext(self.lcx.tcx, n, ity));
                     match op.node {
-                        BinOpKind::Add => l.checked_add(r).map(zext),
-                        BinOpKind::Sub => l.checked_sub(r).map(zext),
-                        BinOpKind::Mul => l.checked_mul(r).map(zext),
+                        // When +, * or binary - create a value greater than the maximum value, or less than
+                        // the minimum value that can be stored, it panics.
+                        BinOpKind::Add => l.checked_add(r).and_then(|n| ity.ensure_fits(n)).map(zext),
+                        BinOpKind::Sub => l.checked_sub(r).and_then(|n| ity.ensure_fits(n)).map(zext),
+                        BinOpKind::Mul => l.checked_mul(r).and_then(|n| ity.ensure_fits(n)).map(zext),
                         BinOpKind::Div if r != 0 => l.checked_div(r).map(zext),
                         BinOpKind::Rem if r != 0 => l.checked_rem(r).map(zext),
-                        BinOpKind::Shr => l.checked_shr(r.try_into().ok()?).map(zext),
-                        BinOpKind::Shl => l.checked_shl(r.try_into().ok()?).map(zext),
+                        // Using << or >> where the right-hand argument is greater than or equal to the number of bits
+                        // in the type of the left-hand argument, or is negative panics.
+                        BinOpKind::Shr if r < bits && !r.is_negative() => l.checked_shr(r.try_into().ok()?).map(zext),
+                        BinOpKind::Shl if r < bits && !r.is_negative() => l.checked_shl(r.try_into().ok()?).map(zext),
                         BinOpKind::BitXor => Some(zext(l ^ r)),
                         BinOpKind::BitOr => Some(zext(l | r)),
                         BinOpKind::BitAnd => Some(zext(l & r)),
@@ -595,24 +675,28 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
                         _ => None,
                     }
                 },
-                ty::Uint(_) => match op.node {
-                    BinOpKind::Add => l.checked_add(r).map(Constant::Int),
-                    BinOpKind::Sub => l.checked_sub(r).map(Constant::Int),
-                    BinOpKind::Mul => l.checked_mul(r).map(Constant::Int),
-                    BinOpKind::Div => l.checked_div(r).map(Constant::Int),
-                    BinOpKind::Rem => l.checked_rem(r).map(Constant::Int),
-                    BinOpKind::Shr => l.checked_shr(r.try_into().ok()?).map(Constant::Int),
-                    BinOpKind::Shl => l.checked_shl(r.try_into().ok()?).map(Constant::Int),
-                    BinOpKind::BitXor => Some(Constant::Int(l ^ r)),
-                    BinOpKind::BitOr => Some(Constant::Int(l | r)),
-                    BinOpKind::BitAnd => Some(Constant::Int(l & r)),
-                    BinOpKind::Eq => Some(Constant::Bool(l == r)),
-                    BinOpKind::Ne => Some(Constant::Bool(l != r)),
-                    BinOpKind::Lt => Some(Constant::Bool(l < r)),
-                    BinOpKind::Le => Some(Constant::Bool(l <= r)),
-                    BinOpKind::Ge => Some(Constant::Bool(l >= r)),
-                    BinOpKind::Gt => Some(Constant::Bool(l > r)),
-                    _ => None,
+                ty::Uint(ity) => {
+                    let bits = ity.bits();
+
+                    match op.node {
+                        BinOpKind::Add => l.checked_add(r).and_then(|n| ity.ensure_fits(n)).map(Constant::Int),
+                        BinOpKind::Sub => l.checked_sub(r).and_then(|n| ity.ensure_fits(n)).map(Constant::Int),
+                        BinOpKind::Mul => l.checked_mul(r).and_then(|n| ity.ensure_fits(n)).map(Constant::Int),
+                        BinOpKind::Div => l.checked_div(r).map(Constant::Int),
+                        BinOpKind::Rem => l.checked_rem(r).map(Constant::Int),
+                        BinOpKind::Shr if r < bits => l.checked_shr(r.try_into().ok()?).map(Constant::Int),
+                        BinOpKind::Shl if r < bits => l.checked_shl(r.try_into().ok()?).map(Constant::Int),
+                        BinOpKind::BitXor => Some(Constant::Int(l ^ r)),
+                        BinOpKind::BitOr => Some(Constant::Int(l | r)),
+                        BinOpKind::BitAnd => Some(Constant::Int(l & r)),
+                        BinOpKind::Eq => Some(Constant::Bool(l == r)),
+                        BinOpKind::Ne => Some(Constant::Bool(l != r)),
+                        BinOpKind::Lt => Some(Constant::Bool(l < r)),
+                        BinOpKind::Le => Some(Constant::Bool(l <= r)),
+                        BinOpKind::Ge => Some(Constant::Bool(l >= r)),
+                        BinOpKind::Gt => Some(Constant::Bool(l > r)),
+                        _ => None,
+                    }
                 },
                 _ => None,
             },
--- a/clippy_utils/src/eager_or_lazy.rs
+++ b/clippy_utils/src/eager_or_lazy.rs
@@ -9,12 +9,13 @@
 //!  - or-fun-call
 //!  - option-if-let-else
 
+use crate::consts::{constant, FullInt};
 use crate::ty::{all_predicates_of, is_copy};
 use crate::visitors::is_const_evaluatable;
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::def_id::DefId;
 use rustc_hir::intravisit::{walk_expr, Visitor};
-use rustc_hir::{Block, Expr, ExprKind, QPath, UnOp};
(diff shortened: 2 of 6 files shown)


Centri3:
Just a thought, but maybe you can call `clippy_utils::consts::constant` to see if it'll panic or not? (It'll return `None` if either it's not a constant or over/underflows), in that case it can still lint.

(That could also be solid new lint potential!)

y21:
@Centri3 Hmm, so you mean, if `constant(lhs, rhs)` returns `Some(_)`, then changing lazy to eager is always safe because it never returns `Some(_)` on overflow? I think that could work

Centri3:
Yep :D Unfortunately won't catch stuff like `x + 2` where `x` is 3 but that should be enough

y21:
It doesn't look like `Some(_)` is guaranteed to mean "no overflow".
Clippy's const eval represents numbers in its bit representation as a `u128` and it does checked operations with `i128`, so it does return `None` if it overflows a `i128`, but not if it overflows an `i32`. So e.g. `i32::MAX + 1` still fits in a i128 so that's fine, but `i128::MAX + 1` is not.

I'm not sure if this could be considered a bug on its own. 🤔
Maybe it could also store the integer type and check that it fits in the target type when evaluating a binop

Centri3:
That's bizarre, that should probably be fixed (as it seems `binop` was intended to return `None` on overflow.)

y21:
@Centri3 just so we don't try to fix something both at the same time, should I fix it or do you want to do it? But also, why a separate function if `binop` was intended to catch overflows (now that I think about it, this should really be considered a bug, because it means const eval has a different value from what it would be at runtime w/o overflow checks)?
And how would that `binop_checked` function be used? `ConstEvalLateContext`  can't be constructed outside of the module, so I'm not sure how one could call `binop_checked` then. Or am I misunderstanding?

Centri3:
Nah I'm not, feel free to :D I was thinking perhaps a flag to `expr`, but it should definitely be changed on `binop` (that was just thrown out in case it was, in fact, not a bug)

clippy_utils/src/consts.rs
```diff
@@ -568,17 +632,30 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
...
+                        BinOpKind::Sub => l.checked_sub(r).and_then(|n| ity.ensure_fits(n)).map(zext),
+                        BinOpKind::Mul => l.checked_mul(r).and_then(|n| ity.ensure_fits(n)).map(zext),
                         BinOpKind::Div if r != 0 => l.checked_div(r).map(zext),
                         BinOpKind::Rem if r != 0 => l.checked_rem(r).map(zext),
-                        BinOpKind::Shr => l.checked_shr(r.try_into().ok()?).map(zext),
-                        BinOpKind::Shl => l.checked_shl(r.try_into().ok()?).map(zext),
+                        // Using << or >> where the right-hand argument is greater than or equal to the number of bits
+                        // in the type of the left-hand argument, or is negative panicks.
+                        BinOpKind::Shr if r < bits && !l.is_negative() => l.checked_shr(r.try_into().ok()?).map(zext),
+                        BinOpKind::Shl if r < bits && !l.is_negative() => l.checked_shl(r.try_into().ok()?).map(zext),
```
guilliamxavier, line 658: should rather be
```suggestion
                        BinOpKind::Shr if r < bits && !r.is_negative() => l.checked_shr(r.try_into().ok()?).map(zext),
                        BinOpKind::Shl if r < bits && !r.is_negative() => l.checked_shl(r.try_into().ok()?).map(zext),
```
I think? (note: I don't see this case in the added tests)

tests/ui/unnecessary_lazy_eval.rs
```diff
@@ -181,10 +183,34 @@ fn main() {
...
+fn panicky_arithmetic_ops(x: usize) {
+    let _x = false.then(|| i32::MAX + 1); // don't lint
+    let _x = false.then(|| i32::MAX * 2); // don't lint
+    let _x = false.then(|| i32::MAX - 1); // lint
+    #[allow(clippy::identity_op)]
+    let _x = false.then(|| (1 + 2 * 3 - 2 / 3 + 9) << 2); // lint
+    let _x = false.then(|| 255u8 << 7); // lint
+    let _x = false.then(|| 255u8 << 8); // don't lint
+    let _x = false.then(|| 255u8 >> 8); // don't lint
+    let _x = false.then(|| 255u8 >> x); // don't lint
```
guilliamxavier, line 212: This is the only line actually using the argument to this function (all other are using literals and/or constants); what about replacing this `x` with e.g. `-7` (negative), removing the parameter, and adding a new function taking `x: i32` (signed) and testing things like `(x != 0).then(|| i32::MAX / x)`, `(x != i32::MIN).then(|| -x)` etc.?

guilliamxavier (commented):
Also bitten by the issue (in a case where `checked_*` would not be appropriate), would love to see this fix land 🙂 Two remarks though

tests/ui/unnecessary_lazy_eval.rs
y21: It's been a while but, FWIW, those other lines that don't use the parameter are there to actually check that the updated const eval logic works the way it should with panicky arithmetic in mind (i.e., when it would or wouldn't panic due to overflow at runtime).

When clippy's const eval sees a path to a local like the `x` in the expression `255 >> x` we're passing to it, it immediately bails out of the evaluator, so I guessed there wouldn't be much of a need/wouldn't be interesting to test different kinds of variations of this and thought one case would be enough, as it just immediately drops out in any case and doesn't even get to the interesting stuff. So for anything involving a local variable such as a parameter, this isn't gonna change eager to lazy or lazy to eager and thus will not emit a lint. Unless I misunderstood your comment?

The test case `255 >> -7` sounds like itd be worth testing though. I haven't yet looked at the other comment but from a quick glance this case (negative RHS) is basically what you were commenting on, I think.

clippy_utils/src/consts.rs
y21: You're right, good catch, thanks! This should be checking the right side, not the left side.

tests/ui/unnecessary_lazy_eval.rs
guilliamxavier: Ah OK, I had ticked at the lone variable in the midst of consts (then suggested adding some "more realistic" uses as variations of the original issue) but that's only to test the "bail out of const eval" thing.

👍 for the added test

clippy_utils/src/consts.rs
```diff
@@ -568,17 +632,30 @@ impl<'a, 'tcx> ConstEvalLateContext<'a, 'tcx> {
...
             (Constant::Int(l), Some(Constant::Int(r))) => match *self.typeck_results.expr_ty_opt(left)?.kind() {
                 ty::Int(ity) => {
+                    let (ty_min_value, _) = ity.min_max()?;
+                    let bits = ity.bits();
                     let l = sext(self.lcx.tcx, l, ity);
                     let r = sext(self.lcx.tcx, r, ity);
+
+                    // Using / or %, where the left-hand argument is the smallest integer of a signed integer type and
+                    // the right-hand argument is -1 always panics, even with overflow-checks disabled
+                    if l == ty_min_value && r == -1 {
```
guilliamxavier, line 642: I just realized that surely this is missing a condition like `matches!(op.node, BinOpKind::Div | BinOpKind::Rem)` (or should be moved inverted to the corresponding guards in the `match` below)?

y21: Huh yeah, this seems to unnecessarily bail out for the other ops where it would be ok. Not sure how I missed that when I wrote this, when the comment up top clearly mentions it 😅 
Will fix this in a bit

holly-hacker:
Hi, are there any blockers for this PR besides the current merge conflicts, or is it just waiting for a review?

y21:
Just waiting for a review. I've been holding off fixing the conflicts because it's been a quiet PR for months and I don't want to rebase this every few days to weeks when it's not clear how long until a review.

@Jarcho Do you plan to review this still? Or should we reroll to get another reviewer to look at this, if you're busy? 🙂

clippy_utils/src/eager_or_lazy.rs
```diff
@@ -207,6 +214,25 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
...
+                ExprKind::Binary(op, ..)
+                    if matches!(
+                        op.node,
+                        BinOpKind::Add
+                            | BinOpKind::Mul
+                            | BinOpKind::Sub
+                            | BinOpKind::Div
+                            | BinOpKind::Rem
+                            | BinOpKind::Shl
+                            | BinOpKind::Shr
```
Jarcho, line 230: These only need to check the second argument.

Jarcho (commented):
This would be better done with `is_const_evaluable`. If evaluating the constant would panic it's most likely a bug as it would always panic when run.

Also floating point math never panics.

y21:
Do we still want to keep the changes in clippy's const eval (returning `None` in case of overflow)? I'd consider it a bug that given the expression `Binary(Add, u32::MAX, 1)` it returns 4294967296, a number that is out of range for a u32. It should really return either `None` (or wrap around), no? `None` seems safer since the result depends on a cfg

Jarcho:
That can be kept as is. Not handling overflow properly was definitely a bug.

y21:
> This would be better done with `is_const_evaluable`. If evaluating the constant would panic it's most likely a bug as it would always panic when run.

I thought about this for a bit. I'm not sure. If I understand this right, doing that would mean we would *still* lint here, right?
```rs
trait Foo {
  const ASSOC: i32;
  fn bar() {
    (Self::ASSOC > 0).then(|| Self::ASSOC - 1);
  }
}
```
`is_const_evaluatable` would return true for the `Self::ASSOC - 1` expression, but I don't think this is enough that we should suggest replacing it with `then_some`. A downstream impl can still implement the trait and give the associated constant the value zero. Then the function will unconditionally overflow.

Jarcho:
Fair point. That just leaves changing the ops that only depend on the second argument.

clippy_utils/src/eager_or_lazy.rs
```diff
@@ -207,6 +214,37 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
...
+                ExprKind::Binary(op, left, right)
+                    if matches!(op.node, BinOpKind::Shl | BinOpKind::Shr)
+                        && let left_ty = self.cx.typeck_results().expr_ty(left)
+                        && let left_bits = left_ty.int_size_and_signed(self.cx.tcx).0.bits()
+                        && constant(self.cx, self.cx.typeck_results(), right)
+                            .and_then(|c| c.int_value(self.cx, left_ty))
+                            .map_or(true, |c| match c {
+                                FullInt::S(i) => i >= i128::from(left_bits) || i.is_negative(),
+                                FullInt::U(i) => i >= u128::from(left_bits),
+                            }) =>
```
Jarcho, line 230: Just FYI. The compiler will already check for this.

clippy_utils/src/eager_or_lazy.rs
```diff
@@ -207,6 +214,37 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
...
+                    self.eagerness |= NoChange;
+                },
+
+                // Arithmetic operations panic on under-/overflow with overflow checks, so don't suggest changing it:
+                // https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#overflow
+                // Using `constant` lets us allow some simple cases where we know for sure it can't overflow
+                ExprKind::Binary(op, ..)
+                    if matches!(
+                        op.node,
+                        BinOpKind::Add | BinOpKind::Mul | BinOpKind::Sub | BinOpKind::Div | BinOpKind::Rem
```
Jarcho, line 241: You forgot about `Div` and `Rem` which also only depend on the second argument.

y21: > Using / or %, where the left-hand argument is the smallest integer of a signed integer type and the right-hand argument is -1 [...]

That sounds a lot like the left hand side matters too?

Jarcho: If the second argument is a constant and isn't `-1` or `0` then the operation will never panic.

clippy_utils/src/eager_or_lazy.rs
```diff
@@ -207,6 +214,37 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
...
                         ty::Bool | ty::Int(_) | ty::Uint(_),
                     ) => {},
+
+                // `>>` and `<<` panic when the right-hand side is greater than or equal to the number of bits in the
+                // type of the left-hand side, or is negative. Doesn't need `constant` on the whole expression
+                // because only the right side matters.
+                ExprKind::Binary(op, left, right)
+                    if matches!(op.node, BinOpKind::Shl | BinOpKind::Shr)
+                        && let left_ty = self.cx.typeck_results().expr_ty(left)
+                        && let left_bits = left_ty.int_size_and_signed(self.cx.tcx).0.bits()
```
y21, line 224: Note to self when I get to addressing the rest: this should make sure that `left_ty` is an int or uint. This probably ices right now for `true.then(|| S1 >> S1)` with a custom shr impl

clippy_utils/src/eager_or_lazy.rs
y21: Can you elaborate? I'm not sure I understand what needs to be changed? Do you mean that rustc's `arithmetic_overflow` lint detects overflowing exprs, so it doesn't matter if we end up linting because the user gets warned anyway when the expr is a constant, so we just need to check for `constant(right).is_none()`?

Jarcho: Since the code as written is almost certainly has the wrong shift amount, it would be better to suggest eagerly evaluating to avoid having the lint only come up after having fixed the first problem.

e.g. When changing `x >> 32` to `x >> 31` it would have been better to get both lints up front.

y21: Do you think we should do the same for `/` and `%`, then? That is, still lint for something like `x / 0`. That would follow the same logic here afaict?

y21: @Jarcho ping? :)

Jarcho: That sounds reasonable.

y21: I was going to do it by checking if `constant(right).is_none()` because only then could it be dependent on runtime values. The compiler will catch it for `Some(_)` constants and we can still early-suggest it like you said because we'd only delay the lint by not immediately suggesting eager evaluation.

I guess if we stay consistent with this, we really should be doing it for all arithmetic operations then, i.e., `u32::MAX + 1` could still be evaluated eagerly. It's a programmer error and the compiler emits an error.

One (more or less -- it's only a FN) problem with this is that something like `true.then(|| 1 >> (2 >> 32))` wouldn't lint, because the RHS of the outer right shift will evaluate to `None` in clippy's const eval.

So in a sense we'd need to know if the failure case is because an error in const eval happened (overflow), or the user tried to do non-const things, such as referring to variables, like `1 >> n`. Only in the latter case do we want to avoid linting.

`is_const_evaluatable` has the problem I mentioned earlier (it can be dependent on downstream consts). There's not really any better util for this, right? I guess we have to either make clippy's const eval return why it failed, or accept the false negative, or come up with some other way.

y21: Maybe we could run `constant()` on the leaves of nested binary expressions, e.g., for `((1 + (n + 3)) >> (5 + (6 * 7))) >> (5 + (6 * 8))` that would be `n`, `3`, `6`, `7`, `6`, `8`. If any of them return `None` in const eval, we should not lint.

If all of them return `Some(_)`, the compiler should be able to either verify that it's okay or it will emit an error that it overflows at runtime. We can still emit a warning in any case.

It's possible that I'm overcomplicating this though ^^

Jarcho: I personally don't think it's worth the effort to propagate the overflow since it's a thing that would be fixed after seeing the error anyways. If you want to code it then go for it. It does make things slightly nicer.

Evaluating the children of a node that failed to evaluate is definitely the wrong way to handle this.

y21: > Evaluating the children of a node that failed to evaluate is definitely the wrong way to handle this.

Why? I'm not seeing where the approach that I mentioned in my last comment would break and I don't think that would add a ton of complexity. That is, checking if all leaves are able to evaluate in `constant()` and if they aren't, overflow can be ruled out and it must mean that it's runtime dependent.

I agree though, this doesn't really matter all that much. Rustc is going to emit an error for overflow and the lint would trigger later after that's fixed no matter if there are FNs here or not.

Jarcho: It would work, but propagating the overflow error back up is both the better solution and not any extra work to code.

y21: Yeah, ok, makes sense now. I feel like that'd involve a lot of changes within, switching from `Option` to `Result`, to propagate it. That could probably be a PR on its own. Let's just stick to only eagerly linting overflow at the root binary expression like `1 >> 32` for now. This all seems like a nice-to-have, not strictly required to fix FNs or FPs.

clippy_utils/src/eager_or_lazy.rs
y21: This seems like it's no longer needed now that we're doing the same for `/` and `%` that we were doing for `>>` (linting as long as the relevant binary op sides are constants)

y21:
I think this should be ready for review now. `/` and `%` was pretty confusing to allow certain cases that rustc can check and disallow others that depend on runtime values and are able to panic.
I also had to fix a bug in `match_same_arms` because it didn't respect `#[allow]` attributes on arms.

clippy_utils/src/eager_or_lazy.rs
```diff
@@ -207,6 +214,53 @@ fn expr_eagerness<'tcx>(cx: &LateContext<'tcx>, e: &'tcx Expr<'_>) -> EagernessS
...
+                            .and_then(|c| c.int_value(self.cx, right_ty))
+                        && match (left, right) {
+                            // `1 / x` -- x might be zero
+                            (_, None) => true,
+                            // `x / -1` -- x might be T::MIN = panic
+                            #[expect(clippy::match_same_arms)]
+                            (None, Some(FullInt::S(-1))) => true,
+                            // anything else is either fine or checked by the compiler
+                            _ => false,
+                        } =>
```
Jarcho, line 246: Little style nit. This could be:
```rust
matches!((left, right),
	// `x / y` -- y might be zero
	(_, None)
	// `x / -1` -- x might be T::MIN 
	| (None, Some(FullInt::S(-1)))
)
```

Might look better. Though I'm currently using github's diff view, so everything looks bad.

Jarcho (approved):

Jarcho:
r=me whether you make the change or not.

y21:
Is this waiting on me to do `r=Jarcho`? Just making sure. Because I can't do that, I don't think. I did make the requested change.

Jarcho:
For some reason I thought you did. @bors r+

Issue #11004, opened 2023-06-22
Clippy incorrectly says that an infinite loop will never loop.


ChaiTRex:
### Summary

When a macro is involved, Clippy incorrectly says that a loop will never loop, but the loop is an infinite loop, so it loops quite a bit.

### Lint Name

never_loop

### Reproducer

I [tried this code](https://play.rust-lang.org/?edition=2021&gist=acffef0d02905565635e63292922b4f4) (which is reduced from [a prime iterator](https://play.rust-lang.org/?edition=2021&gist=5ae227468ea4b0b1c56cbf5e13846eec)):

```rust
#[allow(dead_code)]
fn main() {
    macro_rules! m {
        () => {
            'label: {
                if true {
                    break 'label;
                }
                return;
            }
        };
    }

    loop {
        m!();
    }
}
```

I saw this happen:

```
$ cargo clippy
    Checking never_loop v0.1.0 (/home/chaitrex/dev/never_loop)
error: this loop never actually loops
  --> src/main.rs:14:5
   |
14 | /     loop {
15 | |         m!();
16 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#never_loop
   = note: `#[deny(clippy::never_loop)]` on by default

error: could not compile `never_loop` (bin "never_loop") due to previous error
```

I expected to see this happen:

No errors or warnings from Clippy.

### Version

```text
rustc 1.70.0 (90c541806 2023-05-31)
binary: rustc
commit-hash: 90c541806f23a127002de5b4038be731ba1458ca
commit-date: 2023-05-31
host: x86_64-unknown-linux-gnu
release: 1.70.0
LLVM version: 16.0.2
```

blyxyas:
Expanding this macro will also trigger this lint:

```
#[allow(dead_code)]
fn main() {
    loop {
        'label: {
            if true {
                break 'label;
            }
            return;
        }
    }
}
```

Centri3:
It looks like it doesn't actually evaluate the `if` condition, so it'll just assume it always fails (I guess? More accurately, it never short-circuits if it's always true), maybe for simple expressions like `true` this can be fixed but something more complex is out of scope of clippy

ChaiTRex:
Could it check if there's a `break` (or `continue`, etc.) anywhere before the `return` that would break out past the `return` or continue the `loop`?

digama0:
I also ran across this:
```rust
[...]

PR #11005, opened 2023-06-22
Check if `if` conditions always evaluate to true in `never_loop`

Centri3:
This fixes the example provided in #11004, but it shouldn't be closed as this is still an issue on like
```rust
let x = true;
if x { /* etc */ }`
```
This also makes `clippy_utils::consts::constant` handle `ConstBlock` and `DropTemps`.

changelog: [`never_loop`]: Check if `if` conditions always evaluate to true

### DIFF
--- a/clippy_lints/src/loops/never_loop.rs
+++ b/clippy_lints/src/loops/never_loop.rs
@@ -1,22 +1,23 @@
 use super::utils::make_iterator_snippet;
 use super::NEVER_LOOP;
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::consts::constant;
 use clippy_utils::higher::ForLoop;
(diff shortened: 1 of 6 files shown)


blyxyas:
I'm amazed with the complexity of this lint, I feel like it could be so much simpler :sweat_smile:.
Just checking if there's something before a `return` / `break` should be enough, right? Maybe I'm missing something important here.

Centri3:
I suppose so, but then there's something like
```rust
if false {
	break 'a;
}
return;
```
You're basically switching a few FPs for a whole bunch of FNs, pick your poison I suppose. I think the current way is fine and only has its shortcomings here, with unevaluable conditions

giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #11013, opened 2023-06-23
New lint [`redundant_at_rest_pattern`]

### LINKED ISSUE #11011: New lint to detect usage of `[args @ ..]`
### What it does

Detects usage of `[args @ ..]` and suggests switching it to `args`.

Can probably be named `redundant_rest_pattern`.

@rustbot claim

### Advantage

- Is clearer; a 1-len slice with a rest pattern is(should be) identical to just `args`.

### Drawbacks

None.

### Example

```rust
if let Some([args @ ..]) = init { ... }
```

Could be written as:

```rust
if let Some(args) = init { ... }
```

Centri3:
Closes #11011

It's always a great feeling when a new lint triggers on clippy itself :smile: 

changelog: New lint [`redundant_at_rest_pattern`]


matthiaskrgr:
The first association when I read the title was "REST API" but I don't have any better ideas unfortunately. :sweat_smile:

matthiaskrgr:
The book uses the term `@ binding`. https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#-bindings

Maybe `redundant_pattern_at` `redundant_at_pattern` or `redundant_at_binding` ?

Centri3:
`rest` is for the pattern on the right, i.e., `..`, which is being extraneously binded to on the left. The at operator isn't really the issue

giraffate:
The `rest pattern` is reffered at https://doc.rust-lang.org/reference/patterns.html#rest-patterns. This lint detects the combination of `@`(at binding) and `..` (rest pattern), so how about `redundant_at_rest_pattern`? (`redundant_at_rest_pattern_binding` is a little long?)

Centri3:
I'm fine with `redundant_at_rest_pattern`

clippy_lints/src/misc_early/mod.rs
```diff
@@ -318,6 +319,34 @@ declare_clippy_lint! {
...
+    ///     // NOTE: Type is a slice here
+    ///     println!("all elements: {all:#?}");
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust,ignore
+    /// if let all = v {
+    ///     // NOTE: Type is a `Vec` here
+    ///     println!("all elements: {all:#?}");
+    /// }
```
giraffate, line 342: It warns `irrefutable_let_patterns` in this case, so can we add the case that simply print?

It would be better to detect `irrefutable_let_patterns`, but IMO this lint is still worth it as is.
```suggestion
    /// if let all = v {
    ///     // NOTE: Type is a `Vec` here
    ///     println!("all elements: {all:#?}");
    /// }
    /// // or
    /// println!("all elements: {v:#?}");
```

giraffate (commented):

clippy_lints/src/misc_early/mod.rs
Centri3: I think this is fine. Many lints clippy or not will still lint even if it'll give a warning, and then that lint will hopefully be applied as well by the programmer. Though yeah I'm fine with adding just the `println!`

giraffate:
@bors r+

Thanks!

PR #11016, opened 2023-06-23
[`filter_next`]: suggest making binding mutable if it needs to be

### LINKED ISSUE #10029: filter_next: after fixes were automatically applied the compiler reported errors
### Summary

Clippy suggested a lint to replace `foo.filter().next()` with `foo.find()`. However, it didn't mention anything about requiring `foo` to be mutable.

When I ran `cargo clippy --fix` to automatically apply the lint, it caused a compiler error and told me to file an issue on Github.

[You can view the code here](https://github.com/DarylPinto/advent-of-code-2022/blob/1f8ff55/day04b/src/main.rs#L21-L27), as well as in the `Reproducer` section below.

### Lint Name

filter_next

### Reproducer

I tried this code:

```rust
fn puzzle(input: &str) -> usize {
    input
        .lines()
        .filter(|line| {
            let assignments = line
                .split(',')
                .map(|assignment| {
                    assignment
                        .split('-')
                        .filter_map(|n| n.parse::<i32>().ok())
                        .collect::<Vec<_>>()
                })
                .collect::<Vec<_>>();

            let assignment_one = assignments[0][0]..=assignments[0][1];
            let assignment_two = assignments[1][0]..=assignments[1][1];

            assignment_one
                .filter(|n| assignment_two.contains(n))
                .next()
                .is_some()
        })
        .count()
}
```

I saw this happen:

When I ran `cargo clippy --fix` the following output was written to my terminal:

```

  * day04b/src/main.rs

error[E0596]: cannot borrow `assignment_one` as mutable, as it is not declared as mutable
  --> day04b/src/main.rs:24:13
   |
21 |             let assignment_one = assignments[0][0]..=assignments[0][1];
   |                 -------------- help: consider changing this to be mutable: `mut assignment_one`
...
24 |             assignment_one.find(|n| assignment_two.contains(n))
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot borrow as mutable

For more information about this error, try `rustc --explain E0596`.
```

I expected to see one of the following things happen:
- Clippy to apply the lint without causing a compiler error
- Clippy to not suggest this lint at all

### Version

```text
binary: rustc
commit-date: 2022-11-02
host: aarch64-apple-darwin
release: 1.65.0
LLVM version: 15.0.0
```

### Additional Labels

_No response_

y21:
Fixes #10029

changelog: [`filter_next`]: suggest making binding mutable if it needs to be and adjust applicability

### DIFF
--- a/clippy_lints/src/methods/filter_next.rs
+++ b/clippy_lints/src/methods/filter_next.rs
@@ -1,12 +1,28 @@
-use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg};
+use clippy_utils::diagnostics::{span_lint, span_lint_and_then};
 use clippy_utils::source::snippet;
 use clippy_utils::ty::implements_trait;
+use rustc_ast::{BindingAnnotation, Mutability};
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_lint::LateContext;
 use rustc_span::sym;
 
 use super::FILTER_NEXT;
+
+fn path_to_local(expr: &hir::Expr<'_>) -> Option<hir::HirId> {
+    match expr.kind {
+        hir::ExprKind::Field(f, _) => path_to_local(f),
+        hir::ExprKind::Index(recv, _) => path_to_local(recv),
+        hir::ExprKind::Path(hir::QPath::Resolved(
+            _,
+            hir::Path {
+                res: rustc_hir::def::Res::Local(local),
+                ..
+            },
+        )) => Some(*local),
+        _ => None,
+    }
+}
 
 /// lint use of `filter().next()` for `Iterators`
 pub(super) fn check<'tcx>(
@@ -26,15 +42,30 @@ pub(super) fn check<'tcx>(
         if filter_snippet.lines().count() <= 1 {
             let iter_snippet = snippet(cx, recv.span, "..");
             // add note if not multi-line
-            span_lint_and_sugg(
-                cx,
-                FILTER_NEXT,
-                expr.span,
-                msg,
-                "try",
-                format!("{iter_snippet}.find({filter_snippet})"),
-                Applicability::MachineApplicable,
-            );
+            span_lint_and_then(cx, FILTER_NEXT, expr.span, msg, |diag| {
+                let (applicability, pat) = if let Some(id) = path_to_local(recv)
+                    && let Some(hir::Node::Pat(pat)) = cx.tcx.hir().find(id)
+                    && let hir::PatKind::Binding(BindingAnnotation(_, Mutability::Not), _, ident, _) = pat.kind
+                {
+                    (Applicability::Unspecified, Some((pat.span, ident)))
+                } else {
+                    (Applicability::MachineApplicable, None)
+                };
+
+                diag.span_suggestion(
+                    expr.span,
+                    "try",
+                    format!("{iter_snippet}.find({filter_snippet})"),
+                    applicability,
+                );
+
+                if let Some((pat_span, ident)) = pat {
+                    diag.span_help(
+                        pat_span,
+                        format!("you will also need to make `{ident}` mutable, because `find` takes `&mut self`"),
+                    );
+                }
+            });
         } else {
             span_lint(cx, FILTER_NEXT, expr.span, msg);
         }
--- a/tests/ui/methods_unfixable.rs
+++ b/tests/ui/methods_unfixable.rs
@@ -1,0 +1,10 @@
+#![warn(clippy::filter_next)]
+
+fn main() {
+    issue10029();
+}
+
+pub fn issue10029() {
+    let iter = (0..10);
+    let _ = iter.filter(|_| true).next();
+}
--- a/tests/ui/methods_unfixable.stderr
+++ b/tests/ui/methods_unfixable.stderr
@@ -1,0 +1,15 @@
+error: called `filter(..).next()` on an `Iterator`. This is more succinctly expressed by calling `.find(..)` instead
+  --> $DIR/methods_unfixable.rs:9:13
+   |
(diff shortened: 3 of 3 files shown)


clippy_lints/src/methods/filter_next.rs
```diff
@@ -26,15 +42,30 @@ pub(super) fn check<'tcx>(
...
-                "try this",
-                format!("{iter_snippet}.find({filter_snippet})"),
-                Applicability::MachineApplicable,
-            );
+            span_lint_and_then(cx, FILTER_NEXT, expr.span, msg, |diag| {
+                let (applicability, pat) = if let Some(id) = path_to_local(recv)
+                    && let Some(hir::Node::Pat(pat)) = cx.tcx.hir().find(id)
+                    && let hir::PatKind::Binding(BindingAnnotation(_, Mutability::Not), _, ident, _) = pat.kind
+                {
+                    (Applicability::Unspecified, Some((pat.span, ident)))
```
y21, line 50: It can no longer be a machine applicable suggestion if it needs to be made mutable because of cases like these:
```rs
let (Ok(f) | Err(f)) = ...;
f.filter(...).next();
```
There are multiple bindings that would need to be made mutable, but `cx.tcx.hir().find(id)` only returns one pattern.

`iter_skip_next` is doing this as well, so I think having it be unspecified is fine.
https://github.com/rust-lang/rust-clippy/blob/ce0a48a9b66148fdd96fddf585f9e11850861a47/clippy_lints/src/methods/iter_skip_next.rs#L23-L35

dswij:
Can you help to review this? @blyxyas

blyxyas:
Yep, will review it! :cowboy_hat_face:

clippy_lints/src/methods/filter_next.rs
```diff
@@ -26,15 +42,30 @@ pub(super) fn check<'tcx>(
...
+                    expr.span,
+                    "try this",
+                    format!("{iter_snippet}.find({filter_snippet})"),
+                    applicability,
+                );
+
+                if let Some((pat_span, ident)) = pat {
+                    diag.span_help(
+                        pat_span,
+                        format!("you will also need to make `{ident}` mutable, because `find` takes `&mut self`"),
```
blyxyas, line 65: Knowing `ident`'s span, I think it would be pretty good if it suggested adding a mut directly (instead of a help message, an `Unspecified` suggestion), it would be just doing some span magic and it would be more illustrative.

(Just an opinion, feel free to ignore, the current version is also good)

blyxyas (approved):
Just a mini-nit, but it's very good.. Thanks :heart:!

clippy_lints/src/methods/filter_next.rs
y21: I tried that before, but that didn't work when there is more than one pattern that introduces the identifier, see this comment: https://github.com/rust-lang/rust-clippy/pull/11016#discussion_r1240422104

tests/ui/methods_unfixable.rs
```diff
@@ -0,0 +1,29 @@
+#![warn(clippy::filter_next)]
+
+fn main() {
+    issue10029::puzzle("");
+}
+
+mod issue10029 {
```
dswij, line 7: We can probably make this testcase simpler

clippy_lints/src/methods/filter_next.rs
```diff
@@ -1,13 +1,29 @@
...
 use clippy_utils::ty::implements_trait;
+use rustc_ast::{BindingAnnotation, Mutability};
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_lint::LateContext;
 use rustc_span::sym;
 
 use super::FILTER_NEXT;
 
+fn path_to_local(expr: &hir::Expr<'_>) -> Option<hir::HirId> {
```
dswij, line 12: We have a `path_to_local` helper in `clippy_utils`. Can we use that here?

clippy_lints/src/methods/filter_next.rs
dswij: I gues we can walk the parent to find all the bindings, but that's also not simple. This is good enough for now

clippy_lints/src/methods/filter_next.rs
y21: If I remember correctly, `path_to_local` doesn't "work" if the local is reached through field accesses, so for example it would return `false` for `a.b[0]` if given the id of the local `a`, but we need it to return `true`, because we use this logic to figure out if a binding needs to be made mutable, and `a.b[0].find_mut(..)` requires `a` be mutable. Maybe this could be better named or have a comment explaining it.

tests/ui/methods_unfixable.rs
dswij: Reducing it is great since it makes us understand the problem more (not in this case though), but also makes it simpler for others to understand what's happening and what is being tested in case anyone looks at this in the future :smile:

y21: Ok, fair points, I reduced the test case down to the core problem in the last commit👍

clippy_lints/src/methods/filter_next.rs
blyxyas: `clippy_utils::path_to_local` is too primitive in my opinion. It only handles the case where the expression is a Path, and its Res is a `Res::Local`. Maybe we should modify it and make it handle this new cases.

[(`path_to_local`'s source)](https://doc.rust-lang.org/beta/nightly-rustc/src/clippy_utils/lib.rs.html#450-457)

dswij (approved):

dswij:
Thank you! 

@bors r=blyxyas,dswij

PR #11020, opened 2023-06-24
New lint [`tuple_array_conversions`]

### LINKED ISSUE #10748: Simpler tuple <=> array conversions
### What it does

It suggests better code in cases of conversions from/to tuples to/from arrays.

### Lint Name

simple_tuple_array_conversion

### Category

style

### Advantage

The code is simpler, shorter, less bug-prone.

Since this the array <=> tuple conversions are allowed:
https://github.com/rust-lang/rust/pull/97594

### Drawbacks

None.

### Example

```
#![allow(unused_variables)]
fn main() {
    let t1: &[(u32, u32)] = &[(1, 2), (3, 4)];
    let v1: Vec<[u32; 2]> = t1.iter().map(|&(a, b)| [a, b]).collect();
    let t2: Vec<(u32, u32)> = v1.iter().map(|&[a, b]| (a, b)).collect();
}
```

Could be written as:

```
#![allow(unused_variables)]
fn main() {
    let t1: &[(u32, u32)] = &[(1, 2), (3, 4)];
    let v2: Vec<[u32; 2]> = t1.iter().map(|&t| t.into()).collect();
    let t3: Vec<(u32, u32)> = v2.iter().map(|&v| v.into()).collect();
}
```

The lint should work with longer arrays/tuples too. There are many more complex situations where the .into() could be used to convert between arrays and tuples, but I think this basic case is enough to keep lint logic simple, and it's enough to teach the programmer this transformation is allowed (Clippy is also a teaching tool).

Centri3:
Closes #10748

PS, the implementation is a bit ugly :sweat_smile: ~~I will likely refactor soon enough :)~~ Done :D

changelog: New lint [`tuple_array_conversions`]


llogiq:
Looks good to me. Thank you!

Could we determine whether type annotations are needed and otherwise shorten the help text? In any event, that could be a follow-up PR.

@bors r+

Centri3:
@llogiq, CI unfortunately failed because of `collect-metadata`. It should pass now.

llogiq:
@bors r-

Please rebase.

Centri3:
Was just about to :D

kupiakos:
Given the line `let (a, b) = (buf[0], buf[1]);`, I would not replace it with `let (a, b) = (buf[0], buf[1]);` as this lints suggests, but instead write `let [a, b] = buf;`.

Issue #4559, opened 2019-09-20
Suggest std::iter::{once, empty}


matthiasbeyer:
I don't know whether clippy is capable of doing this, but it would be nice to suggest `std::iter::once()` and `std::iter::empty()`, especially when things like `vec![element].iter()` (or similar) is used.

mikerite:
This is possible but the type of `std::iter::once()` \ `std::iter::empty()` isn't the same as `vec![..].iter()`. This means that the lint needs to be careful when it makes the suggestion. (See #3307 for an example of what can go wrong.) It's probably only safe when the result is immediately used by another iterator method.

jfrimmel:
One thing where this lint could always be applied is this pattern:
```
Some(x).into_iter(); // -> std::iter::once(x)
None.into_iter(); // -> std::iter::empty()
```
I see this relatively common, but the expressed meaning of these statements isn't as clear as it could be if the standard iterator functions would be used.

llogiq:
If returning it as `impl Iterator`, it would also be OK to change, I guess.

jplatte:
Just saw `Some(foo).into_iter().chain(something_else).collect()` today. So in addition to the possibility of returning `impl Iterator`, another (maybe more likely) scenario where this lint could be applied is when the iterator (chain) is consumed in the same expression, though I'm not sure how hard that is to detect.

PR #11023, opened 2023-06-25
New lint [`option_iter`]

Centri3:
Idea came from https://github.com/rust-lang/rust-clippy/issues/4559#issuecomment-538624846.

We can probably extend this in the future to `vec![element].iter()` :D

changelog: New lint [`option_iter`]


Centri3:
Welp, while going through and blessing tests I saw `iter_on_empty_collections` and `iter_on_single_items` :sweat_smile: Damn. I should really check the nursery category before implementing lints! But, these seem to miss many cases; crucially, it won't lint if it's returned or something like `chain(Some(0))`, which this will. I also made it more general and was about to commit that before noticing the previous lints. It's basically a better version of those lints now :D

But, maybe this should update those lints instead? Make them finally `pedantic`, or even `style` potentially?

Centri3:
I'm closing this PR in favor of extending the already existing two lints, and likely bumping them up to `pedantic`/`style`. It should be opened soon enough!

PR #11028, opened 2023-06-25
Skip if_not_else lint for '!= 0'-style checks

BenWiederhake:
Currently, clippy makes unhelpful suggestions such as this:

```
warning: unnecessary `!=` operation
   --> src/vm.rs:598:36
    |
598 |                     *destination = if source & 0x8000 != 0 { 0xFFFF } else { 0 };
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: change to `==` and swap the blocks of the `if`/`else`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#if_not_else
    = note: `-W clippy::if-not-else` implied by `-W clippy::pedantic`
```

Bit tests often take on the form `if foo & 0x1234 != 0 { … } else { … }`, and the `!= 0` part reads as "has any bits set". Therefore, this code already has the "correct" order, and shouldn't be changed.

This PR disables the lint for these cases, and in fact all cases where the condition is "foo is non-zero".

I did my homework:
- \[X] Followed [lint naming conventions][lint_naming] → Not applicable, this PR fixes an existing lint
- \[X] Added passing UI tests (including committed `.stderr` file) → Yes, `tests/ui/if_not_else_bittest.rs`
- \[X] `cargo test` passes locally
- \[X] Executed `cargo dev update_lints`
- \[X] Added lint documentation → Not applicable, this PR fixes an existing lint
- \[X] Run `cargo dev fmt`

changelog: Fix [`if_not_else`] false positive when something like `bitflags != 0` is used

### DIFF
--- a/tests/ui/if_not_else_bittest.rs
+++ b/tests/ui/if_not_else_bittest.rs
@@ -1,0 +1,11 @@
+#![deny(clippy::if_not_else)]
+
+fn show_permissions(flags: u32) {
+    if flags & 0x0F00 != 0 {
+        println!("Has the 0x0F00 permission.");
+    } else {
+        println!("The 0x0F00 permission is missing.");
+    }
+}
+
+fn main() {}
--- a/clippy_lints/src/if_not_else.rs
+++ b/clippy_lints/src/if_not_else.rs
@@ -1,6 +1,7 @@
 //! lint on if branches that could be swapped so no `!` operation is necessary
 //! on the condition
 
+use clippy_utils::consts::{constant_simple, Constant};
 use clippy_utils::diagnostics::span_lint_and_help;
 use clippy_utils::is_else_clause;
 use rustc_hir::{BinOpKind, Expr, ExprKind, UnOp};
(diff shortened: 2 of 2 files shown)


tests/ui/if_not_else_bittest.rs
```diff
@@ -0,0 +1,11 @@
+#![deny(clippy::if_not_else)]
```
llogiq, line 1: Why is this in its own file? You can add it to `if_not_else.rs` instead.

BenWiederhake: It is a separate test (absence of a warning), so putting it in its own file seems more appropriate. If you insist, I can mash it in the other file.

llogiq: No, that's alright.

llogiq:
Ok, this seems like it should reduce the less useful messages from clippy, so thank you!

@bors r+

Issue #9665, opened 2022-10-17
Suggest `{f32, f64}::is_finite` and `{f32, f64}::is_infinite`


Luro02:
### What it does

Detects manual implementations of these functions and suggests replacing them with a call to that function.

### Lint Name

float_manual_finite/float_manual_infinite

### Category

style, pedantic

### Advantage

- easier to read
- less code to work with
- could be faster than the original code [see source of `is_infinite`, which suggests some future optimizations](https://doc.rust-lang.org/src/core/num/f32.rs.html#460)

### Drawbacks

- maybe changes the meaning of the code?

### Example

```rust
fn double(float: f32) -> f32 {
    if float == f32::INFINITY || float == f32::NEG_INFINITY {
        println!("is infinite");
    }
    
    float * 2.0
}


fn main() {
    dbg!(double(6.63E-34));
}
```

Could be written as:

```rust
fn double(float: f32) -> f32 {
    if float.is_infinite() {
        println!("is infinite");
    }
    
    float * 2.0
}


fn main() {
    dbg!(double(6.63E-34));
}
```

------

```rust
fn double(float: f32) -> f32 {
    assert!(float != f32::INFINITY && float != f32::NEG_INFINITY);

    float * 2.0
}


fn main() {
    dbg!(double(6.63E-34));
}
```

Could be written as:

```rust
fn double(float: f32) -> f32 {
    assert!(float.is_finite());

    float * 2.0
}


fn main() {
    dbg!(double(6.63E-34));
}
```

PR #11029, opened 2023-06-26
Make `comparison_to_empty` work on `if let`/`let` chains

Centri3:
This adds `LetChain` to `clippy_utils::higher`, other lints may benefit from such a change as well :D

changelog: Enhancement: [`comparison_to_empty`]: Now lints on `if let`

### DIFF
--- a/clippy_lints/src/len_zero.rs
+++ b/clippy_lints/src/len_zero.rs
@@ -7,11 +7,10 @@ use rustc_ast::ast::LitKind;
 use rustc_errors::Applicability;
 use rustc_hir::def::Res;
 use rustc_hir::def_id::{DefId, DefIdSet};
-use rustc_hir::lang_items::LangItem;
 use rustc_hir::{
     AssocItemKind, BinOpKind, Expr, ExprKind, FnRetTy, GenericArg, GenericBound, ImplItem, ImplItemKind,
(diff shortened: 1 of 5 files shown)


Jarcho:
This would work just as well by checking for `ExprKind::Let` rather then the whole if condition. `ExprKind::Let` can only exist as part or it.

Centri3:
Done, I've kept `LetChain` though as I still think it can be useful in other cases

Jarcho:
I'm not sure how useful it will end up. An iterator over binary operator chains might be useful, but I'd rather wait for a definite use case before adding it.

Generally util functions shouldn't be added without at least a single use case.

Centri3:
> I'm not sure how useful it will end up. An iterator over binary operator chains might be useful, but I'd rather wait for a definite use case before adding it.
> 
> Generally util functions shouldn't be added without at least a single use case.

#9665. It requires something like this, and my prototype implementation uses it.

Jarcho:
If that's the case then it should be added with that PR so it can be reviewed with it's use case.

Jarcho:
ping @Centri3

This is only waiting on a rebase and the removal of the let-chain util.

Jarcho:
Thank you. @bors r+

PR #11030, opened 2023-06-26
suggests `is_some_and` over `map().unwrap`

### LINKED ISSUE #9125: Use is_some_and when appropriate
### What it does

The lint would detect cases where the new Option function is_some_and would be appropriate. Examples would be places where `map()` converts an `Option<T>` into an `Option<bool>` and then `unwrap_or(false)` is used. The library function is still experimental, but once it's stable, this would be a useful modernization suggestion.

### Lint Name

option_is_some_with

### Category

complexity

### Advantage

Reduces function calls (even if they would optimize down), increases clarity.

### Drawbacks

Low priority, since the existing approaches are functional, if less idiomatic/aesthetic.

### Example

```rust
let x: Option<int> = Some(7);
x.map(|val| val > 5).unwrap_or(false)
```

Could be written as:

```rust
let x: Option<int> = Some(7);
x.is_some_and(|val| val > 5)
```

darklyspaced:
changelog:
 * modified option_map_unwrap_or lint to recognise when an `Option<T>` is mapped to an `Option<bool>` with false being used when `None` is detected; suggests the use of `is_some_and` instead
 * msrv is set to 1.70.0 for this lint; when `is_some_and` was stabilised

fixes #9125


darklyspaced:
apologies for the slightly questionably git history.

dogfood was failing because `check` had too many arguments after adding support for msrv discrimination on the lint. i wasn't exactly sure on the most idiomatic way to fix this, so i just wrapped two similar arguments into a tuple which was then destructed inside the function.

please let me know if there is a better way to get around the too many arguments warning.

clippy_lints/src/methods/mod.rs
```diff
@@ -3904,7 +3906,7 @@ impl Methods {
                             manual_saturating_arithmetic::check(cx, expr, lhs, rhs, u_arg, &arith["checked_".len()..]);
                         },
                         Some(("map", m_recv, [m_arg], span, _)) => {
-                            option_map_unwrap_or::check(cx, expr, m_recv, m_arg, recv, u_arg, span);
+                            option_map_unwrap_or::check(cx, expr, m_recv, m_arg, (recv, u_arg), span, &self.msrv);
```
Centri3, line 3909: You should probably add `#[expect(clippy:;too_many_arguments)]` to the function instead.

xFrednet:
@Centri3 Do you want to take over the review for this PR? :)

Centri3:
Sure, I'll take a look in a bit :D

clippy_utils/src/msrvs.rs
```diff
@@ -19,6 +19,7 @@ macro_rules! msrv_aliases {
 
 // names may refer to stabilized feature flags or library items
 msrv_aliases! {
+    1,70,0 { OPT_IS_SOME_AND }
```
Centri3, line 22: nit: We should probably name this `OPTION_IS_SOME_AND` for consistency with the other msrvs.

clippy_lints/src/methods/option_map_unwrap_or.rs
```diff
@@ -74,16 +77,34 @@ pub(super) fn check<'tcx>(
             return;
         }
 
+        let mut suggest_is_some_and = false;
```
Centri3, line 80: nit: As this isn't mutated past the following `if`, let's make this something like
```rust
let suggest_is_some_and = if msrv.meets(msrvs::OPT_IS_SOME_AND) {
} else {
	false
}
```
(@xFrednet, just curious, if you know, can you suggest an easier way to do this? i.e., make a variable `true` if a let chain evaluates to `true`...)

clippy_lints/src/methods/option_map_unwrap_or.rs
```diff
@@ -74,16 +77,34 @@ pub(super) fn check<'tcx>(
...
+        // argument to `unwrap_or` is false & is_some_and is stabilised; should suggest using `is_some_and`
+        if msrv.meets(msrvs::OPT_IS_SOME_AND) {
```
Centri3, line 82: This should be turned into a `let` chain to reduce nesting.

Centri3 (commented):
Looks pretty good overall, I've left some small comments. Thanks! :D

clippy_lints/src/methods/option_map_unwrap_or.rs
darklyspaced: i reckon taking advantage of the `matches!` macro would work quite well as the only purpose of the `if let` blocks are pattern matching. 

something like this would work:
```
let suggest_is_some_and: bool = msrv.meets(msrvs::OPTION_IS_SOME_AND)
    && matches!(&unwrap_arg.kind, ExprKind::Lit(lit) 
    if matches!(lit.node, rustc_ast::LitKind::Bool(false)));
```

idk if its sacrificing readability at thought point though. imho, it's fine.

Centri3: Yeah that could be fine here though let chains with then `true` else `false` is pretty common, so having a better way would be nice

xFrednet: > (@xFrednet, just curious, if you know, can you suggest an easier way to do this? i.e., make a variable true if a let chain evaluates to true...)

Nope, sadly not that I know of. It's possible to use the `if_chain` macro, but that would be less readable than your suggested alternative. I've also noticed that Rust is missing something for this. There was an RFC, to make `let <pat> = <expr>` evaluate to a boolean in every context, but that one got rejected IIRC.

A similar way to write your suggestion would also be:
```rs
let suggest_is_some_and;
if msrv.meets(msrvs::OPT_IS_SOME_AND) {
	suggest_is_some_and = true;
} else {
	suggest_is_some_and = false;
}
```
This can be more readable, in some circumstances, but often it's just a style question :)

clippy_lints/src/methods/option_map_unwrap_or.rs
```diff
@@ -27,6 +29,7 @@ pub(super) fn check<'tcx>(
     unwrap_recv: &rustc_hir::Expr<'_>,
     unwrap_arg: &'tcx rustc_hir::Expr<'_>,
     map_span: Span,
+    msrv: &Msrv,
```
xFrednet, line 32: NIT: Can you add this lint to the configuration documentation, to make sure that it's correctly linked in Clippy's lint list?

You should be able to just add it to the list here:

https://github.com/rust-lang/rust-clippy/blob/ea4c5c5ca55ae9614552c0d285ee6c051cf88b30/clippy_lints/src/utils/conf.rs#L297-L300

(On your branch, that is just a link to the file :))

xFrednet (approved):

xFrednet:
This version looks good to me, just a tiny nit, then it's ready to be merged. Thank you for the PR, I hope you had fun working on Clippy!

@Centri3 Thank you for the review. I leave the final r+ to you.

@bors d=Centri3

Centri3 (approved):
This looks good, thank you :)

@bors r=Centri3,xFrednet

Centri3:
Well bors is still asleep. But, @darklyspaced, I do believe you need to run `cargo collect-metadata`. I looked again and saw no changes to lint configuration, unfortunately dogfood itself doesn't check it, only CI does when testing an approved PR (which should definitely be changed btw, as it's really annoying)

Once that's been ran, I'll try reapproving again :D (also if there's no changes ^^)

Centri3:
Thank you!

@bors r=Centri3,xFrednet

xFrednet:
> Well bors is still asleep

`@bors` comments sadly don't work in the review comments. That's why you always see everyone first approving the PR and then adding a text comment with the LGTM and r+ :)

cc: @blyxyas this might also be interesting for you :)

PR #11031, opened 2023-06-26
New lint [`needless_return_with_try`]

### LINKED ISSUE #10902: Needless return with Err and question mark
### What it does

I sometimes see `return Err(MyError::SomeError(v))?;` -- which seems redundant - it has both the `return` AND the question mark operator.  So this can be rewritten as either
* `Err(MyError::SomeError(v))?;` - without the return
* `return Err(MyError::SomeError(v));` - without the question operator

The second option only works if the error type is the same as in `Result<..., MyError>`, otherwise it only works with the question mark.

So question - does it make sense to suggest needless `return` in the case where it can be expressed with a question mark? Or should the question mark be suggested to be removed in the case where it can? Or neither?

### Advantage

- shorter code - no confusion over two "return" operators

### Drawbacks

- without `return` might be less readable, so perhaps only for same-result-errors and remove the question mark?

### Example

```rust
use thiserror::Error;

#[derive(Error, Debug)]
enum MyTopError {
    #[error("Internal Error: {0}")]
    SubError(#[from] MySubError),
    #[error("Main Error")]
    MainError,
}

#[derive(Error, Debug)]
enum MySubError {
    #[error("Value {0} too small")]
    TooSmall(i32),
}

fn foo(v: i32) -> Result<(), MyTopError> {
    if v < 5 {
        return Err(MySubError::TooSmall(v))?;
        //    This is identical to the above, but too verbose:
        // return Err(MyTopError::SubError(MySubError::TooSmall(v)));
        //    This is identical to the above, but has no "return"
        // Err(MySubError::TooSmall(v))?;
        //    This causes an error: mismatched types
        // return Err(MySubError::TooSmall(v));
    }
    if v > 100 {
        return Err(MyTopError::MainError)?;
        // Err(MyTopError::MainError)?;        // This is identical to the above, but has no "return"
        // return Err(MyTopError::MainError);  // This is identical to the above, but has no question mark
    }
    println!("v: {v}");
    Ok(())
}

fn main() {
    let v = 3;
    let r = foo(v);
    println!("r: {:?}", r);
}
```

Centri3:
Closes #10902

Rather than having a config option, this will just suggest removing the "return"; if `try_err` is used as well, then it'll be added again but without the `?`.

changelog: New lint [`needless_return_with_try`]


giraffate:
I didn't immediately understand `try` in `needless_return_with_try` to mean `?`. IMO, so how about `needless_return_with_question_mark` or just `needless_return_with_question` for the lint naming?

Centri3:
I'm fine with that, `_try` just felt more concise (after all, `?` is the *try* operator)

clippy_lints/src/returns.rs
```diff
@@ -76,6 +78,46 @@ declare_clippy_lint! {
...
+    /// if paired with `try_err`, use instead:
+    /// ```rust,ignore
+    /// fn foo(x: usize) -> Result<(), Box<dyn Error>> {
+    ///     if x == 0 {
+    ///         return Err(...);
+    ///     }
+    ///     Ok(())
+    /// }
+    /// ```
+    #[clippy::version = "1.72.0"]
```
giraffate, line 115: Sorry for the my late review. Can we change it?
```suggestion
    #[clippy::version = "1.73.0"]
```

clippy_lints/src/returns.rs
```diff
@@ -76,6 +78,46 @@ declare_clippy_lint! {
...
+    /// simplify to
+    /// ```rust,ignore
+    /// fn foo(x: usize) -> Result<(), Box<dyn Error>> {
+    ///     if x == 0 {
+    ///         Err(...)?;
+    ///     }
+    ///     Ok(())
+    /// }
+    /// ```
+    /// if paired with `try_err`, use instead:
```
giraffate, line 106: IMO it would be better to change suggestions by using `is_lint_allowed` when warning `try_err`, but it can be done in another pull request.

giraffate (commented):

clippy_lints/src/returns.rs
Centri3: We should probably move around `try_err` then, probably to `returns` instead of `matches`, that way we can reuse the logic. But I'm ok with this.

giraffate:
Okay. Then can we update the version https://github.com/rust-lang/rust-clippy/pull/11031#discussion_r1263151767 ?

Centri3:
Yeah, but I'll also move `try_err`. I'll get back to this soon.

Centri3:
This actually seems too complex to tackle rn, I'll take another look at some point and potentially create a followup PR

giraffate:
@bors r+

Thanks!

PR #11035, opened 2023-06-26
[`option_if_let_else`]: suggest `.as_ref()` if scrutinee is of type `&Option<_>`

### LINKED ISSUE #10729: `clippy::option_if_let_else` suggestion doesn't compile when passing a matched reference to a function
### Summary

If you're `match`ing on some reference type (e.g. `&Option<String>`) and in the `Some` case are passing the produced inner reference (e.g. `&String`) to a function taking a reference, then the result from following `clippy:option_if_let_else`'s suggestion will not compile.

### Reproducer

I tried this code:

```rust
#![warn(clippy::option_if_let_else)]

pub fn reproduce(initial: &Option<String>) {
    match initial {
        Some(value) => do_something(value),
        None => {}
    }
}

fn do_something(_value: &str) {}
```

Clippy produces the following warning:

```log
❯ cargo clippy
warning: use Option::map_or instead of an if let/else
 --> src/lib.rs:4:5
  |
4 | /     match initial {
5 | |         Some(value) => do_something(value),
6 | |         None => {}
7 | |     }
  | |_____^ help: try: `initial.map_or({}, |value| do_something(value))`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::option_if_let_else)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: `clippy-bug-repro` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
```

Following the suggestion results in this code:

```rust
#![warn(clippy::option_if_let_else)]

pub fn reproduce(initial: &Option<String>) {
    initial.map_or({}, |value| do_something(value))
}

fn do_something(_value: &str) {}
```

Which fails to compile with:

```log
❯ cargo build 
   Compiling clippy-bug-repro v0.1.0 (/home/yami/dev/repos/clippy-bug-repro)
 --> src/lib.rs:4:45
  |
4 |     initial.map_or({}, |value| do_something(value))
  |                                ------------ ^^^^^
  |                                |            |
  |                                |            expected `&str`, found `String`
  |                                |            help: consider borrowing here: `&value`
  |                                arguments to this function are incorrect
  |
note: function defined here
 --> src/lib.rs:7:4
  |
7 | fn do_something(_value: &str) {}
  |    ^^^^^^^^^^^^ ------------

error: could not compile `clippy-bug-repro` (lib) due to previous error
```

<details>
<summary>Further "whack-a-mole" results that may not be considered directly relevant to this issues, but are probably still useful as a reference.</summary>
  
Following _that_ suggestion _also_ does not compile:

```log
❯ cargo build
   Compiling clippy-bug-repro v0.1.0 (/home/yami/dev/repos/clippy-bug-repro)
error[E0507]: cannot move out of `*initial` which is behind a shared reference
    --> src/lib.rs:4:5
     |
4    |     initial.map_or({}, |value| do_something(&value))
     |     ^^^^^^^^----------------------------------------
     |     |       |
     |     |       `*initial` moved due to this method call
     |     help: consider calling `.as_ref()` or `.as_mut()` to borrow the type's contents
     |     move occurs because `*initial` has type `Option<String>`, which does not implement the `Copy` trait
     |
note: `Option::<T>::map_or` takes ownership of the receiver `self`, which moves `*initial`
    --> /home/yami/.rustup/toolchains/beta-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1155:31
     |
1155 |     pub const fn map_or<U, F>(self, default: U, f: F) -> U
     |                               ^^^^
help: you can `clone` the value and consume it, but this might not be your desired behavior
     |
4    |     initial.clone().map_or({}, |value| do_something(&value))
     |             ++++++++

For more information about this error, try `rustc --explain E0507`.
error: could not compile `clippy-bug-repro` (lib) due to previous error
```

Following the suggestion to change `initial.map_or(...)` to `initial.as_ref().map_or(...)` starts to compile, but results in two more warnings:

```log
❯ cargo clippy
warning: passing a unit value to a function
 --> src/lib.rs:4:5
  |
4 |     initial.as_ref().map_or({}, |value| do_something(&value))
  |     ^^^^^^^^^^^^^^^^^^^^^^^^--^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |                             |
  |                             help: use a unit literal instead: `()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unit_arg
  = note: `#[warn(clippy::unit_arg)]` on by default

warning: this expression creates a reference which is immediately dereferenced by the compiler
 --> src/lib.rs:4:54
  |
4 |     initial.as_ref().map_or({}, |value| do_something(&value))
  |                                                      ^^^^^^ help: change this to: `value`
  |

warning: `clippy-bug-repro` (lib) generated 2 warnings (run `cargo clippy --fix --lib -p clippy-bug-repro` to apply 2 suggestions)
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
```

The `clippy::unit_arg` is fair, and I imagine is purely an artefact of my reproduction's `None` branch being a no-op.

However, the suggestion for `clippy::needless_borrow` actually undoes the _original_ (incorrect) suggestion made in the context of `clippy::option_if_let_else`. I imagine this would need to be considered when "fixing" this issue.
</details>

### Version

```text
rustc 1.70.0-beta.2 (071f14baa 2023-04-30)
binary: rustc
commit-hash: 071f14baae931e17a5a99366bf216e76384cc4f6
commit-date: 2023-04-30
release: 1.70.0-beta.2
LLVM version: 16.0.2
```

### Additional Labels

y21:
Fixes #10729

`Option::map_or` takes ownership, so if matching on an `&Option<_>`, we need to suggest `.as_ref()` before calling `map_or` to get the same effect and to not cause a borrowck error. 

changelog: [`option_if_let_else`]: suggest `.as_ref()`/`.as_mut()` if scrutinee is of type `&Option<_>`/`&mut Option<_>`

### DIFF
--- a/tests/ui/option_if_let_else.rs
+++ b/tests/ui/option_if_let_else.rs
@@ -239,6 +239,8 @@ fn main() {
         Ok(a) => a + 1,
     };
     let _ = if let Ok(a) = res { a + 1 } else { 5 };
+    issue10729::reproduce(&None);
+    issue10729::reproduce2(&mut None);
 }
 
 #[allow(dead_code)]
@@ -248,4 +250,26 @@ fn issue9742() -> Option<&'static str> {
         Some(name) if name.starts_with("foo") => Some(name.trim()),
         _ => None,
     }
+}
+
+mod issue10729 {
+    #![allow(clippy::unit_arg)]
+
+    pub fn reproduce(initial: &Option<String>) {
+        // 👇 needs `.as_ref()` because initial is an `&Option<_>`
+        match initial {
+            Some(value) => do_something(value),
+            None => {},
+        }
+    }
+
+    pub fn reproduce2(initial: &mut Option<String>) {
+        match initial {
+            Some(value) => do_something2(value),
+            None => {},
+        }
+    }
+
+    fn do_something(_value: &str) {}
+    fn do_something2(_value: &mut str) {}
 }
--- a/tests/ui/option_if_let_else.rs
+++ b/tests/ui/option_if_let_else.rs
@@ -239,8 +239,6 @@ fn main() {
         Ok(a) => a + 1,
     };
     let _ = if let Ok(a) = res { a + 1 } else { 5 };
-    issue10729::reproduce(&None);
-    issue10729::reproduce2(&mut None);
 }
 
 #[allow(dead_code)]
@@ -253,7 +251,7 @@ fn issue9742() -> Option<&'static str> {
 }
 
 mod issue10729 {
-    #![allow(clippy::unit_arg)]
+    #![allow(clippy::unit_arg, dead_code)]
 
     pub fn reproduce(initial: &Option<String>) {
         // 👇 needs `.as_ref()` because initial is an `&Option<_>`
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -140,6 +140,9 @@ fn try_get_option_occurrence<'tcx>(
             let (as_ref, as_mut) = match &expr.kind {
                 ExprKind::AddrOf(_, Mutability::Not, _) => (true, false),
                 ExprKind::AddrOf(_, Mutability::Mut, _) => (false, true),
+                _ if let Some(mutb) = cx.typeck_results().expr_ty(expr).ref_mutability() => {
+                    (mutb == Mutability::Not, mutb == Mutability::Mut)
+                }
                 _ => (bind_annotation == BindingAnnotation::REF, bind_annotation == BindingAnnotation::REF_MUT),
             };
 
--- a/tests/ui/option_if_let_else.fixed
+++ b/tests/ui/option_if_let_else.fixed
@@ -198,8 +198,6 @@ fn main() {
     let _ = res.map_or(1, |a| a + 1);
     let _ = res.map_or(1, |a| a + 1);
     let _ = res.map_or(5, |a| a + 1);
-    issue10729::reproduce(&None);
-    issue10729::reproduce2(&mut None);
 }
 
 #[allow(dead_code)]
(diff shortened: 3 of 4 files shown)


tests/ui/option_if_let_else.rs
```diff
@@ -239,6 +239,8 @@ fn main() {
         Ok(a) => a + 1,
     };
     let _ = if let Ok(a) = res { a + 1 } else { 5 };
+    issue10729::reproduce(&None);
+    issue10729::reproduce2(&mut None);
```
Jarcho, line 243: Small nit. There's no need for this as the test code is only compiled, not run.

Jarcho (commented):
Just to check, this lint currently doesn't handle patterns like `&Some(_)`, correct?

y21:
> Just to check, this lint currently doesn't handle patterns like &Some(_), correct?

Nope, it doesn't lint in that case: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=790624fb7c9a4ddb11753e60e278ec5b

tests/ui/option_if_let_else.rs
y21: I did that to silence the dead code warning, but I guess I could also `#[allow]` it.

Jarcho: You can also prefix names with underscores as well. That silences unused warnings.

Jarcho:
Ok. Not a requirement to handle it, just want to make sure that case doesn't end up with a regressions. Once the little nit is handled then this is good.

tests/ui/option_if_let_else.rs
y21: Decided to go with `#![allow(dead_code)]` for that module because it's also used for other unused functions in that test file, hope that's fine too

Jarcho:
Thank you. @bors r+

PR #11039, opened 2023-06-27
Add `BLESS` for compile-test and some cleanup

Alexendoo:
changelog: none

Allows passing the environment variable `BLESS` to bless tests, which is useful when you want to bless internal tests - `BLESS= cargo uitest -Finternal`

Also updates a place in the docs referring to `cargo dev bless` and removes some unused test deps


flip1995:
There are `cargo bless` and `cargo uibless` aliases already. Is this additional way to bless tests necessary?

Alexendoo:
You can't pass feature flags with those, might also be needed for things not using the alias like nextest - https://github.com/rust-lang/rust-clippy/pull/10426#issuecomment-1516592013

llogiq:
I don't think such a small benign change needs a huge motivation. Alright!

@bors r+

PR #11043, opened 2023-06-27
Use substring matching for TESTNAME

Alexendoo:
Restores the previous behaviour of matching using a substring match rather than needing a full match

changelog: none


Jarcho:
I assume this changed on the switch to `ui_test`?

@bors r+

Alexendoo:
> I assume this changed on the switch to `ui_test`?

Yeah

PR #11046, opened 2023-06-29
New lint [`iter_skip_zero`]

Centri3:
Idea came from my contribution to `unnecessary_cast` recently. Sorry about that :sweat_smile:

Could be an issue if somebody manually implements `skip`, but I don't think anybody will (and this would be an issue with other lints too, afaik)

changelog: New lint [`iter_skip_zero`]

### DIFF
--- a/clippy_lints/src/methods/iter_skip_zero.rs
+++ b/clippy_lints/src/methods/iter_skip_zero.rs
@@ -1,0 +1,34 @@
+use clippy_utils::consts::{constant, Constant};
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::{is_from_proc_macro, is_trait_method};
+use rustc_errors::Applicability;
+use rustc_hir::Expr;
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::ITER_SKIP_ZERO;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, arg_expr: &Expr<'_>) {
+    if !expr.span.from_expansion()
+        && is_trait_method(cx, expr, sym::Iterator)
+        && let Some(arg) = constant(cx, cx.typeck_results(), arg_expr).and_then(|constant| {
+            if let Constant::Int(arg) = constant {
+                Some(arg)
+            } else {
+                None
+            }
+        })
+        && arg == 0
+        && !is_from_proc_macro(cx, expr)
+    {
+        span_lint_and_then(cx, ITER_SKIP_ZERO, arg_expr.span, "usage of `.skip(0)`", |diag| {
+            diag.span_suggestion(
+                arg_expr.span,
+                "if you meant to skip the first element, use",
+                "1",
+                Applicability::MaybeIncorrect,
+            )
+            .note("this call to `skip` does nothing and is useless; remove it");
+        });
+    }
+}
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -45,6 +45,7 @@ mod iter_nth_zero;
 mod iter_on_single_or_empty_collections;
 mod iter_overeager_cloned;
 mod iter_skip_next;
+mod iter_skip_zero;
 mod iter_with_drain;
 mod iterator_step_by_zero;
 mod manual_next_back;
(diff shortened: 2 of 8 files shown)


xFrednet:
`r?` @blyxyas

(xFrednet has picked a reviewer for you, use r? to override)

clippy_lints/src/methods/iter_skip_zero.rs
```diff
@@ -0,0 +1,36 @@
...
+        && !is_from_proc_macro(cx, expr)
+    {
+        span_lint_and_sugg(
+            cx,
+            ITER_SKIP_ZERO,
+            arg_expr.span,
+            "usage of `.skip(0)`",
+            "if you meant to skip the first element, try",
+            "1".to_owned(),
+            Applicability::MaybeIncorrect,
```
blyxyas, line 33: I'm not sure this should be a suggestion, as it changes the program's logic flow.

clippy_lints/src/methods/iter_skip_zero.rs
```diff
@@ -0,0 +1,36 @@
...
+        })
+        && arg == 0
+        && !is_from_proc_macro(cx, expr)
+    {
+        span_lint_and_sugg(
+            cx,
+            ITER_SKIP_ZERO,
+            arg_expr.span,
+            "usage of `.skip(0)`",
+            "if you meant to skip the first element, try",
```
blyxyas, line 31: I feel like this flows better

```suggestion
            "if you meant to skip the first element, use",
```

blyxyas (changes requested):
Very interesting concept for a lint. It has some things I'd like to talk about, but very good overall :heart:

clippy_lints/src/methods/mod.rs
```diff
@@ -3284,6 +3285,27 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Why is this bad?
+    /// This was likely intended to be `.skip(1)` to skip the first element, as `.skip(0)` does
+    /// nothing.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let v = vec![1, 2, 3];
+    /// let x = v.iter().skip(0).collect::<Vec<_>>();
+    /// let y = v.iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
Centri3: There's a good few others like that (only example, mostly self-explanatory lints)

clippy_lints/src/methods/iter_skip_zero.rs
Centri3: There are other lints like that, like `almost_complete_range`. I'm fine with changing it but imo just because it changes the program's behavior doesn't mean it shouldn't give a suggestion

xFrednet: `MaybeIncorrect` is intended for changes, which change the semantics, but it's likely that the change is preferable.

rust-lang/rustfix#200 might be an interesting read, with more context and also something which could be an interesting project :)

clippy_lints/src/methods/iter_skip_zero.rs
```diff
@@ -0,0 +1,36 @@
...
+        && arg == 0
+        && !is_from_proc_macro(cx, expr)
+    {
+        span_lint_and_sugg(
+            cx,
+            ITER_SKIP_ZERO,
+            arg_expr.span,
+            "usage of `.skip(0)`",
+            "if you meant to skip the first element, use",
+            "1".to_owned(),
```
blyxyas, line 32: I'm not sure how to handle this lint. The suggestion it produces different outputs, and I'm afraid that the change may be applied without really thinking and breaking software. I think the best course of action here would be to keep this message, but also add a note that it currently is totally useless and doing absolutely nothing.

Centri3: I suppose we could have two suggestions, one that's the current and one that removes it entirely

blyxyas (approved):
LGTM, Thanks :heart: !

xFrednet:
I should have bors permissions, once the team PR is merged. For now, I can delegate these PRs to you two :)

@bors delegate @Centri3 
@bors delegate @blyxyas

xFrednet:
Theoretically, if I get the comment right

@bors delegate=blyxyas 
@bors delegate=Centri3

blyxyas:
r+

blyxyas:
Life is hard

blyxyas:
Maybe the second delegation overwrote the first one? @Centri3 could you try?

Centri3:
Maybe, @bors r=blyxyas

xFrednet:
WTF? So picky that boy

Hey @bors be nicer to our newest members. Honer their new power!!! :sparkles:

proski:
Just for the record, this lint breaks Alacritty. It has code that skips some non-zero amount of characters in one branch but doesn't skip anything in another. The result is stored in the same structure field of the type `Skip<Chars>`.

`skip(0)` is used to convert the type from `Chars` to `Skip<Chars>`. I wish I could use something else, but I don't see what would be a good replacement. There is no `into()` implementation either from the original iterator to `Skip` or from `Skip` to the original. `advance_by` is nightly-only, so it cannot be used yet.

blyxyas:
This lint can be allowed by using an `#<!>[allow(clippy::iter_skip_zero)]` attribute in the code, `-- -A clippy::iter_skip_zero` in the terminal or CI, or even by adding `lints.clippy.iter_skip_zero = "allow"` in your `Cargo.toml`.

I'm not seeing how this is an issue, or how a lint can break a project other than generating some new error messages in the CI.

Centri3:
Probably the fix that breaks it? We should call `expr_use_ctxt` to handle this

proski:
Thank you for the responses! I didn't mean to say that the issue cannot be ignored in the code. My point is that there are legitimate uses of `skip(0)`. They can seem contrived, but there is a real life example.

PR #11047, opened 2023-06-29
Create `implicit_transmute_types` lint.

Flying-Toast:
Type inference is useful in most cases, but it can cause issues around usage of transmute. Even though a specific usage of transmute may have been confirmed to be sound, someone might unintentionally change a piece of surrounding code that causes the transmuted types to change and possibly become unsound. Despite that, transmute() will happily continue to infer these changed types.

This commit adds a lint `implicit_transmute_types` that catches calls to transmute without an explicit turbofish. The idea is that the initial author of code that uses a transmute will verify the soundness of converting between the two types, and enfore those types on the transmute call. If the types end up changing later, typecheck will fail and force the transmute call to be updated and remind the user to reassess the soundness of transmuting between the new types.

### DIFF
--- a/clippy_lints/src/implicit_transmute_types.rs
+++ b/clippy_lints/src/implicit_transmute_types.rs
@@ -1,0 +1,108 @@
+use clippy_utils::{diagnostics::span_lint_and_then, is_diagnostic_item_or_ctor, last_path_segment};
+use rustc_errors::Applicability;
+use rustc_hir::{self as hir, ExprKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to [`transmute`] without explicit type parameters
+    /// (i.e. without turbofish syntax).
+    ///
+    /// ### Why is this bad?
+    /// In most cases Rust's type inference is helpful, however it can cause
+    /// problems with [`transmute`]. `transmute` is wildly unsafe
+    /// unless the types being transmuted are known to be compatible. As such,
+    /// a seemingly innocent change in something's type can end up making a
+    /// previously-valid transmute suddenly become unsound. Thus it is
+    /// good practice to always be explicit about the types you expect to be
+    /// transmuting between, so that the compiler will force you to
+    /// reexamine the transmute if either type changes.
+    ///
+    /// ### Example
+    /// ```rust
+    /// #[repr(transparent)]
+    /// struct CharWrapper {
+    ///     _inner: char,
+    /// }
+    ///
+    /// let wrapped = CharWrapper { _inner: 'a' };
+    /// let transmuted = unsafe { core::mem::transmute(wrapped) };
+    ///
+    /// // This is sound now, but if it gets changed in the future to
+    /// // something that expects a type other than `char`, the transmute
+    /// // would infer it returns that type, which is likely unsound.
+    /// let _ = char::is_lowercase(transmuted);
+    /// ```
+    ///
+    /// Specify type parameters:
+    /// ```rust
+    /// # #[repr(transparent)]
+    /// # struct CharWrapper {
+    /// #     _inner: char,
+    /// # }
+    /// let wrapped = CharWrapper { _inner: 'a' };
+    /// // Because we explicitly specify the types for the transmute, any change in
+    /// // surrounding code that would cause the transmute call to infer different
+    /// // types will now be caught by typechecking, forcing us to come back and
+    /// // reassess the soundsness of transmuting between the new types.
+    /// let transmuted = unsafe { core::mem::transmute::<CharWrapper, char>(wrapped) };
+    ///
+    /// let _ = char::is_lowercase(transmuted);
+    /// ```
+    ///
+    /// If you decide that you *do* want the types to be inferred,
+    /// you can silence the lint by conveying your intention explicitly:
+    /// ```rust
+    /// # use std::mem::transmute;
+    /// # fn main() {
+    /// #     unsafe {
+    /// #         let foo: i32 = 123;
+    /// #         let _: u32 =
+    /// transmute::<_, _>(foo);
+    /// #     }
+    /// # }
+    /// ```
+    ///
+    /// [`transmute`]: https://doc.rust-lang.org/core/mem/fn.transmute.html
+    #[clippy::version = "1.72.0"]
+    pub IMPLICIT_TRANSMUTE_TYPES,
+    style,
+    "calling mem::transmute without explicit type parameters"
+}
+
+declare_lint_pass!(ImplicitTransmuteTypes => [IMPLICIT_TRANSMUTE_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for ImplicitTransmuteTypes {
+    fn check_expr(&mut self, cx: &LateContext<'_>, e: &hir::Expr<'_>) {
+        if let ExprKind::Call(func, _) = &e.kind
+            && let ExprKind::Path(qpath) = &func.kind
+            && let Some(def_id) = cx.qpath_res(qpath, func.hir_id).opt_def_id()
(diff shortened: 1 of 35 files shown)


clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// // surrounding code that would cause the transmute call to infer different
+    /// // types will now be caught by typechecking, forcing us to come back and
+    /// // reassess the soundsness of transmuting between the new types.
+    /// let transmuted = unsafe { core::mem::transmute::<CharWrapper, char>(wrapped) };
+    ///
+    /// let _ = char::is_lowercase(transmuted);
+    /// ```
+    ///
+    /// If you decide that you *do* want the types to be inferred,
+    /// you can silence the lint by conveying your intention explicitly:
```
Centri3, line 55: I feel like having a configuration option to also disallow this would be nice, you can probably check the 1st and 2nd arguments of the last segment to see if either are `_`

clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,105 @@
...
+declare_lint_pass!(ImplicitTransmuteTypes => [IMPLICIT_TRANSMUTE_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for ImplicitTransmuteTypes {
+    fn check_expr(&mut self, cx: &LateContext<'_>, e: &hir::Expr<'_>) {
+        if let ExprKind::Call(func, _) = &e.kind
+            && let ExprKind::Path(qpath) = &func.kind
+            && let Some(def_id) = cx.qpath_res(qpath, func.hir_id).opt_def_id()
+            && let Some(sym::transmute) = cx.tcx.get_diagnostic_name(def_id)
+            && let hir::QPath::Resolved(_, path) = qpath
+            && let Some(last_segment) = path.segments.last()
```
Centri3, line 83: nit: This could use `last_path_segment` instead. As this is `sym::transmute`, it won't a `LangItem`, so it won't panic (as it will on `LangItem`s.)

clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+declare_lint_pass!(ImplicitTransmuteTypes => [IMPLICIT_TRANSMUTE_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for ImplicitTransmuteTypes {
+    fn check_expr(&mut self, cx: &LateContext<'_>, e: &hir::Expr<'_>) {
+        if let ExprKind::Call(func, _) = &e.kind
+            && let ExprKind::Path(qpath) = &func.kind
+            && let Some(def_id) = cx.qpath_res(qpath, func.hir_id).opt_def_id()
+            && let Some(sym::transmute) = cx.tcx.get_diagnostic_name(def_id)
```
Centri3, line 81: nit: This can use `is_diagnostic_item_or_ctor`, not too big of an improvement here but when you're handling `Some` and such it comes in real handy (and is a bit cleaner)

SImilarly, you can use `path_res` instead of `qpath_res` if you want

clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,105 @@
...
+            && let Some(last_segment) = path.segments.last()
+            && let None = last_segment.args
+        {
+            let suggestion_span = qpath.span().shrink_to_hi();
+
+            let substs = cx.typeck_results().node_substs(func.hir_id);
+            let srctype = substs.type_at(0);
+            let dsttype = substs.type_at(1);
+
+            cx.struct_span_lint(
```
Centri3, line 92: You should use the `span_lint_and_then` function instead.

Centri3 (commented):
This looks pretty good, thank you! I've left some comments.

xFrednet:
Hey @Flying-Toast welcome to Clippy :wave: 

Thank you for starting the review @Centri3. You're welcome to continue it, and ping me once everything is done :)

Centri3:
I think I've already said everything I wanted to ^^ feel free to give your thoughts

clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,108 @@
...
+    pub IMPLICIT_TRANSMUTE_TYPES,
+    style,
+    "calling mem::transmute without explicit type parameters"
+}
+
+declare_lint_pass!(ImplicitTransmuteTypes => [IMPLICIT_TRANSMUTE_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for ImplicitTransmuteTypes {
+    fn check_expr(&mut self, cx: &LateContext<'_>, e: &hir::Expr<'_>) {
+        if let ExprKind::Call(func, _) = &e.kind
```
xFrednet, line 79: Could you add a check, to ensure that the code doesn't come from macros? You can checkout [`in_external_macro`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/lint/fn.in_external_macro.html#) for this :)

clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,108 @@
...
+                cx,
+                IMPLICIT_TRANSMUTE_TYPES,
+                e.span,
+                "`transmute` called without explicit type parameters",
+                |b| {
+                    b.span_suggestion_verbose(
+                        suggestion_span,
+                        "consider specifying the types intended to be transmuted",
+                        format!("::<{srctype}, {dsttype}>"),
+                        Applicability::MachineApplicable
```
xFrednet, line 102: Printing types like this can sometimes cause problems, when rustc can't determine the correct path for an item. This should be:
```suggestion
                        Applicability::MaybeIncorrect
```

tests/ui/implicit_transmute_types.rs
```diff
@@ -0,0 +1,14 @@
...
+#![warn(clippy::implicit_transmute_types)]
+
+use std::mem;
+
+fn main() {
+    unsafe {
+        let _: u32 = mem::transmute(123i32);
+
+        let _: u32 = mem::transmute::<i32, u32>(123i32);
+        let _: u32 = mem::transmute::<_, _>(123i32);
```
xFrednet, line 12: Can you add some more tests, with structs and other more complex types? Maybe play with `String`, `struct A(u32)` etc. These files don't get executed, so the transmutes can be unsound, without spawning nostril dragons :wink:

clippy_lints/src/implicit_transmute_types.rs
```diff
@@ -0,0 +1,108 @@
...
+
+declare_lint_pass!(ImplicitTransmuteTypes => [IMPLICIT_TRANSMUTE_TYPES]);
+
+impl<'tcx> LateLintPass<'tcx> for ImplicitTransmuteTypes {
+    fn check_expr(&mut self, cx: &LateContext<'_>, e: &hir::Expr<'_>) {
+        if let ExprKind::Call(func, _) = &e.kind
+            && let ExprKind::Path(qpath) = &func.kind
+            && let Some(def_id) = cx.qpath_res(qpath, func.hir_id).opt_def_id()
+            && is_diagnostic_item_or_ctor(cx, def_id, sym::transmute)
+            && last_path_segment(qpath).args.is_none()
```
xFrednet, line 83: Here we could optionally check, that the two types are not the infer ones `::<_, _>`. But that's an optional improvement. You're welcome to pick it up, or we just create a follow-up issue for this :)

xFrednet (commented):
Hey, I think this is a good start. I left some comments. You're welcome to reach out to @Centri3 or me if you have any questions! :)

xFrednet:
Ping @Flying-Toast, this is a ping from triage. Do you plan to continue this PR in the near future?

Flying-Toast:
> Ping @Flying-Toast, this is a ping from triage. Do you plan to continue this PR in the near future?

Probably not, so I'm fine closing it or if someone else wants to take it over.

xFrednet:
Okay, thank you for the time and effort you already put into this PR. :D

PR #11051, opened 2023-06-30
Make `eq_op` suggest `.is_nan()`

Centri3:
changelog: Enhancement: [`eq_op`]: Suggests `is_nan()` for `x != x` where `x` is a float


llogiq:
Makes sense to me, and the implementation looks good. Thank you!

@bors r+

PR #11052, opened 2023-06-30
New lint [`string_lit_chars_any`]

### LINKED ISSUE #10389: restrict: `str_literal.chars().any(|x| x == c)` affects performance negatively
### What it does

prevents `str_literal.chars().any(|x| x == c)`.

### Lint Name

str_literal_chars_any

### Category

restriction

### Advantage

- Contributes optimizer: see https://godbolt.org/z/nsdPj3EaP, about x40 speed

### Drawbacks

_No response_

### Example

```rust
let foo: bool = "\\.+*?()|[]{}^$#&-~".chars().any(|x| x == c)
```

Could be written as:

```rust
let foo: bool = match c {
    '\\' | '.' | '+' | '*' | '?' | '(' | ')' | '|' | '[' | ']' | '{'
    | '}' | '^' | '$' | '#' | '&' | '-' | '~' => true,
    _ => false,
}
```

Centri3:
Closes #10389

changelog: New lint [`string_lit_chars_any`]


clippy_lints/src/methods/string_lit_chars_any.rs
```diff
@@ -0,0 +1,80 @@
...
+    param: &'tcx Param<'tcx>,
+    body: &Expr<'_>,
+    msrv: &Msrv,
+) {
+    if !msrv.meets(MATCHES_MACRO) || !is_trait_method(cx, expr, sym::Iterator) {
+        return;
+    };
+
+    // All bindings introduced by `param`
+    let mut bindings = FxHashSet::default();
```
Jarcho, line 31: Using a `HashSet` to hold a single binding is excessive. The pattern should be a single `PatKind::Binding`. The only other valid pattern is `PatKind::Wild`.

clippy_lints/src/methods/string_lit_chars_any.rs
```diff
@@ -0,0 +1,80 @@
...
+            bindings.insert(hir_id);
+        }
+    });
+
+    if let ExprKind::Lit(lit_kind) = recv.kind
+        && let LitKind::Str(val, _) = lit_kind.node
+        && let ExprKind::Binary(kind, lhs, rhs) = body.kind
+        && let BinOpKind::Eq = kind.node
+        && let Some(scrutinee) = lhs_rhs_are_param_and_scrutinee(&bindings, lhs, rhs)
+            .or_else(|| lhs_rhs_are_param_and_scrutinee(&bindings, rhs, lhs))
```
Jarcho, line 44: Once the `HashSet` is changed to a single `HirId` this can be simplified to:
```
&& let Some(lhs) = path_to_local(lhs)
&& let Some(rhs) = path_to_local(rhs)
&& (lhs == arg) != (rhs == arg)
```

clippy_lints/src/methods/string_lit_chars_any.rs
```diff
@@ -0,0 +1,60 @@
...
+    if msrv.meets(MATCHES_MACRO)
+        && is_trait_method(cx, expr, sym::Iterator)
+        && let PatKind::Binding(_, arg, _, _) = param.pat.kind
+        && let ExprKind::Lit(lit_kind) = recv.kind
+        && let LitKind::Str(val, _) = lit_kind.node
+        && let ExprKind::Binary(kind, lhs, rhs) = body.kind
+        && let BinOpKind::Eq = kind.node
+        && let Some(lhs_path) = path_to_local(lhs)
+        && let Some(rhs_path) = path_to_local(rhs)
+        && let scrutinee = if lhs_path == arg {
```
Jarcho, line 31: This doesn't handle the case where `lhs_path == rhs_path`, though I doubt anybody would actually write it. Something like `match (arg == lhs_path, arg == rhs_path)` might be clearer.

Centri3: Yeah I don't think we should handle that either, as wouldn't that have to be `|c| c == c` or `|_| x == x`? It's basically useless and other lints will take care of it anyway ^^

Will make it a `match` though :+1:

Jarcho: I meant it as that case was not handled correctly (i.e. not linted). New version is fine.

Jarcho:
Looks good now. Thank you. @bors r+

PR #11053, opened 2023-06-30
Use `cargo build --tests` in CI

Alexendoo:
I noticed that we run a `cargo build` but end up downloading/building a bunch of deps after that for tests in CI

https://github.com/rust-lang/rust-clippy/actions/runs/5426673277/jobs/9869019973#step:6:12
https://github.com/rust-lang/rust-clippy/actions/runs/5426673277/jobs/9869019973#step:7:11

This builds the tests in the build step too, it may speed up runs by downloading/building more in parallel

changelog: none


flip1995:
@bors r+

Thanks!

PR #11057, opened 2023-07-01
New lint [`inefficient_pow`]

Centri3:
changelog: New lint [`inefficient_pow`]


Centri3:
I'm not entirely sure how useful this lint is, considering the cost of using `.pow()` anyway isn't too large (see [godbolt](https://rust.godbolt.org/z/Gevdvjzc6)) but maybe it's desired somewhere ^^

Centri3:
Huh, I've just looked at the assembly for `1 << 200` and it seems it uses `shl edi,cl`, with `and ecx,31`, which is, uh, weird. That's probably why it breaks in a really weird way when it overflows :woman_shrugging: At least without it, it wouldn't roll back over and instead be `1100` -> `1000`.

Unfortunately `shl` doesn't set the overflow flag afaik so it can't really be a rustc issue.

Centri3:
I don't particularly like the suggested code, and think this should be an optimization pass in rustc instead. I'll work on that instead

PR #11061, opened 2023-07-02
`let_and_return`: lint 'static lifetimes, don't lint borrows in closures

### LINKED ISSUE #11056: `cargo clippy` attempted to fix `let_and_return` but failed
### Summary

Hi, I'm learning rust. I was not able to identify the issue: Clippy attempted to fix a function but it failed. 
It looks similar to [#91523](https://github.com/rust-lang/rust/issues/91523).

### Lint Name

let_and_return

### Reproducer

```bash
git clone git@github.com:Mcdostone/kconfig.git
cd kconfig
cargo clippy --fix
```

<p>

```
cargo clippy   --fix --allow-dirty

    Checking kconfig v0.1.0
warning: failed to automatically apply fixes suggested by rustc to crate `kconfig`

  * src/kconfig.rs

error[E0597]: `file` does not live long enough
  --> src/kconfig.rs:23:19
   |
21 |       map(delimited(ws_comment, many0(parse_entry), ws_comment), |d| {
   |       -                                                          --- value captured here
   |  _____|
   | |
22 | |         Kconfig {
23 | |             file: file.display().to_string(), //input.extra.file.display().to_string(),
   | |                   ^^^^ borrowed value does not live long enough
24 | |             entries: d,
25 | |             //architecture: Noneis_architecture_related(&input.extra.0.clone(), &input.extra.1),
26 | |         }
27 | |     })(input)
   | |______- a temporary with access to the borrow is created here ...
28 |   }
   |   -
   |   |
   |   `file` dropped here while still borrowed
   |   ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `impl FnMut(nom_locate::LocatedSpan<&str, KconfigFile>) -> std::result::Result<(nom_locate::LocatedSpan<&str, KconfigFile>, kconfig::Kconfig), nom::Err<nom::error::Error<nom_locate::LocatedSpan<&str, KconfigFile>>>>`
   |
   = note: the temporary is part of an expression at the end of a block;
           consider forcing this temporary to be dropped sooner, before the block's local variables are dropped
help: for example, you could save the expression's value in a new local variable `x` and then make `x` be the expression at the end of the block
   |
21 ~     let x = map(delimited(ws_comment, many0(parse_entry), ws_comment), |d| {
22 |         Kconfig {
 ...
26 |         }
27 ~     })(input); x
   |

For more information about this error, try `rustc --explain E0597`.
```

</p>
</details>

### Version

```text
binary: rustc
commit-date: 2023-05-31
release: 1.70.0
LLVM version: 16.0.2
```

### Additional Labels

_No response_

Alexendoo:
Fixes #11056

Now also ignores functions returning `'static` lifetimes, since I noticed the `stdin.lock()` example was still being linted but doesn't need to be since https://github.com/rust-lang/rust/pull/93965

changelog: none

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::{span_lint_and_then, span_lint_hir_and_then};
 use clippy_utils::source::{snippet_opt, snippet_with_context};
-use clippy_utils::visitors::{for_each_expr, Descend};
+use clippy_utils::visitors::{for_each_expr_with_closures, Descend};
 use clippy_utils::{fn_def_id, path_to_local_id, span_find_starting_semi};
 use core::ops::ControlFlow;
 use if_chain::if_chain;
@@ -328,7 +328,7 @@ fn emit_return_lint(cx: &LateContext<'_>, ret_span: Span, semi_spans: Vec<Span>,
 }
 
 fn last_statement_borrows<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) -> bool {
-    for_each_expr(expr, |e| {
+    for_each_expr_with_closures(cx, expr, |e| {
         if let Some(def_id) = fn_def_id(cx, e)
             && cx
                 .tcx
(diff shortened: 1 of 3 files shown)


llogiq:
Thank you, LGTM.

@bors r+

PR #11077, opened 2023-07-03
[`arc_with_non_send_sync`]: don't lint if type has nested type parameters

### LINKED ISSUE #11076: arc_with_non_send_sync false positive in generic code
### Summary

I believe there is nothing wrong with the code in the repro below.

Seems like Clippy wants a `T: Send` bound but I don't see a reason why that bound would need to be written.

### Lint Name

arc_with_non_send_sync

### Reproducer

```rust
use std::sync::{Arc, Mutex};
use std::thread;

pub struct Collector<T> {
    out: Arc<Mutex<Vec<T>>>,
}

impl<T> Collector<T> {
    pub fn new() -> Self {
        let out = Arc::new(Mutex::new(Vec::new()));
        Collector { out }
    }
}

fn main() {
    let collector = Collector::new();

    // Do Send stuff with it
    let thread = thread::spawn({
        let out = Arc::clone(&collector.out);
        move || out.lock().unwrap().push("success")
    });
    thread.join().unwrap();
    println!("{:?}", Arc::into_inner(collector.out).unwrap().into_inner().unwrap());
}
```

```console
$ cargo clippy

error: usage of `Arc<T>` where `T` is not `Send` or `Sync`
  --> src/main.rs:10:19
   |
10 |         let out = Arc::new(Mutex::new(Vec::new()));
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: consider using `Rc<T>` instead or wrapping `T` in a std::sync type like `Mutex<T>`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
   = note: `#[deny(clippy::arc_with_non_send_sync)]` on by default
```

### Version

```text
rustc 1.72.0-nightly (839e9a6e1 2023-07-02)
binary: rustc
commit-hash: 839e9a6e1210934fd24b15548b811a97c77138fc
commit-date: 2023-07-02
release: 1.72.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

_No response_

y21:
Fixes #11076

changelog: [`arc_with_non_send_sync`]: don't lint if type has nested type parameters

r? @Manishearth

### DIFF
--- a/clippy_lints/src/arc_with_non_send_sync.rs
+++ b/clippy_lints/src/arc_with_non_send_sync.rs
@@ -7,6 +7,7 @@ use rustc_hir::{Expr, ExprKind};
 use rustc_lint::LateContext;
 use rustc_lint::LateLintPass;
 use rustc_middle::ty;
+use rustc_middle::ty::GenericArgKind;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::symbol::sym;
 
@@ -47,7 +48,10 @@ impl LateLintPass<'_> for ArcWithNonSendSync {
             if let ExprKind::Path(func_path) = func.kind;
             if last_path_segment(&func_path).ident.name == sym::new;
             if let arg_ty = cx.typeck_results().expr_ty(arg);
-            if !matches!(arg_ty.kind(), ty::Param(_));
+            // make sure that the type is not and does not contain any type parameters
+            if arg_ty.walk().all(|arg| {
+                !matches!(arg.unpack(), GenericArgKind::Type(ty) if matches!(ty.kind(), ty::Param(_)))
+            });
             if !cx.tcx
                 .lang_items()
                 .sync_trait()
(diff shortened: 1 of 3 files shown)


Manishearth (approved):

jfgoog:
I agree that this fix needs to go into 1.72.0.

I am testing the 1.72 beta for the Android Rust toolchain, and we have some code that uses Arc<tokio::sync::Mutex> that erroneously triggers this error.

The Tokio docs even specifically mention Arc<Mutex>, so I suspect we aren't the only ones who will encounter this problem:

https://github.com/tokio-rs/tokio/blob/a7d52c2fede5ebd8f6e49d52a4af77138e0bd6e1/tokio/src/sync/mutex.rs

jfgoog:
I have confirmed that this patch eliminates the compile error I was getting.

I also tried updating tokio to the latest version (without the patch), and stlll got the compile error.

flip1995:
https://github.com/rust-lang/rust/pull/114937

PR #11088, opened 2023-07-03
Make `single_range_in_vec_init` ignore type annotations, fn arguments and `ExprField`s

### LINKED ISSUE #11086: `single_range_in_vec_init` stronger than expected, when type is known
### Summary

The `single_range_in_vec_init` warning seems a bit overly cautious, if the type resulting from an invocation of a `vec!` macro is already known "from the outside", e.g. via a type annotation or when it is used as an argument / return value.

For example
```rust
let foo: Vec<Range<usize>> = vec![1..5]; // causes "warning: a `Vec` of `Range` that is only one element"
```
In this case clippy might infer, that the user indeed wanted a `Vec<Range<...>>`, because otherwise the program could not even have type checked. The same is true for this example:

```rust
fn barify(v: Vec<Range<usize>>) {
  ...
}

let v = vec![3..42]; // causes "warning: a `Vec` of `Range` that is only one element"
barify(v);
```
One explicit way to remedy this is using explicit lengths:
```rust
let foo = vec![1..5; 1]; // OK
```
although I am slightly worried that tools like cargo-fmt might at some point opt to remove that.

### Lint Name

single_range_in_vec_init

### Reproducer

I tried this code:

```rust
use std::ops::Range;

fn main() {
    let _foo: Vec<Range<usize>> = vec![1..5];
}
```

I saw this happen:

```
warning: a `Vec` of `Range` that is only one element
 --> src/main.rs:4:35
  |
4 |     let _foo: Vec<Range<usize>> = vec![1..5];
  |                                   ^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_range_in_vec_init
  = note: `#[warn(clippy::single_range_in_vec_init)]` on by default
help: if you wanted a `Vec` that contains the entire range, try
  |
4 |     let _foo: Vec<Range<usize>> = (1..5).collect::<std::vec::Vec<usize>>();
  |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
help: if you wanted a `Vec` of len 5, try
  |
4 |     let _foo: Vec<Range<usize>> = vec![1; 5];
  |                                        ~~~~

warning: `single_range` (bin "single_range") generated 1 warning
```

No warnings produced.

### Version

_No response_

### Additional Labels

_No response_

Centri3:
Closes #11086

This still lints if it's a local and used in a function, as ignoring that would be *very* hard and *very* slow, but it now gives a note to give type annotations if this is intended.

changelog: Enhancement: [`single_range_in_vec_init`]: Ignores if it's a local that has type annotations, or is immediately passed to a function or struct initializer

### DIFF
--- a/clippy_lints/src/single_range_in_vec_init.rs
+++ b/clippy_lints/src/single_range_in_vec_init.rs
@@ -1,9 +1,9 @@
 use clippy_utils::diagnostics::span_lint_and_then;
-use clippy_utils::get_trait_def_id;
 use clippy_utils::higher::VecArgs;
 use clippy_utils::macros::root_macro_call_first_node;
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::implements_trait;
+use clippy_utils::{get_parent_node, get_trait_def_id, peel_blocks};
 use rustc_ast::{LitIntType, LitKind, UintTy};
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind, LangItem, Local, Node, QPath};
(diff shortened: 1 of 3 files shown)


Alexendoo:
#11166 could be useful here

Alexendoo:
#11166 was merged, let's use that here

xFrednet:

xFrednet:
Removing inactive closed in favor of #12611

PR #11094, opened 2023-07-03
[`useless_vec`]: add more tests and don't lint inside of macros

### LINKED ISSUE #11084: [`useless_vec`]: add more tests
### Description

I'm mainly creating this issue so we don't forget about this: https://github.com/rust-lang/rust-clippy/pull/11081#issuecomment-1618074901

We should add these as test cases for the lint.

### Version

```text
n/a
```

### Additional Labels

_No response_

y21:
Closes #11084.

I realized that the fix I added in #11081 itself also causes an error in a suggestion when inside of a macro. Example:
```rs
macro_rules! x {
  () => {
    for _ in vec![1, 2] {}
  }
}
x!();
```
Here it would suggest replacing `vec![1, 2]` with `[x!()]`, because that's what the source callsite is (reminder: it does this to get the correct span of `x!()` for code like `for _ in vec![x!()]`), but that's wrong when *inside* macros, so I decided to make it not lint if the whole loop construct is inside a macro to avoid this issue.

changelog: [`useless_vec`]: add more tests and don't lint inside of macros

r? @Alexendoo since these were your tests, I figured it makes most sense to assign you

### DIFF
--- a/clippy_lints/src/vec.rs
+++ b/clippy_lints/src/vec.rs
@@ -154,6 +154,10 @@ impl UselessVec {
         span: Span,
         suggest_slice: SuggestedType,
     ) {
+        if span.from_expansion() {
+            return;
+        }
+
         let mut applicability = Applicability::MachineApplicable;
 
         let snippet = match *vec_args {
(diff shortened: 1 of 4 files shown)


Alexendoo (approved):

Alexendoo:
Great, thanks!

@bors r+

PR #11096, opened 2023-07-03
[`manual_range_patterns`]: lint negative values

### LINKED ISSUE #11091: FN manual_range_patterns : negative values
### Description

```rust
fn main() {
    let x: i8 = i8::MAX;

    matches!(x, -1 | -2 | -3 | -4);

    let y = -3..5;
    let v = y.collect::<Vec<_>>();
    dbg!(v);
}
````
could warn as ranges can also go negative.

### Version

```text
rustc 1.72.0-nightly (5bd28f5ea 2023-06-28)
binary: rustc
commit-hash: 5bd28f5eac1ba3569bfa8d49ec3f5acbdfdff7a0
commit-date: 2023-06-28
release: 1.72.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

_No response_

y21:
Fixes #11091.

Now also lints negative values in patterns (`-1 | -2 | -3`)

changelog: [`manual_range_patterns`]: lint negative values


y21:
Okay, I think this requires some pretty involved changes to handle "large" numbers correctly. For example:
```rs
    matches!(
        1u128,
        340282366920938463463374607431768211455
            | 340282366920938463463374607431768211454
            | 340282366920938463463374607431768211453
    );
```
This currently emits the warning "help: try: `-3..=-1`", because the integer literal wraps around. This probably needs lots of casting and checks to make all the comparisons etc work. Though maybe there's a nicer way to do this. Will have to think about it.
(or I guess we could also just ignore numbers >=i128::MAX for now 🤔 )

Centri3:
> Okay, I think this requires some pretty involved changes to handle "large" numbers correctly. For example:
> 
> ```rust
>     matches!(
>         1u128,
>         340282366920938463463374607431768211455
>             | 340282366920938463463374607431768211454
>             | 340282366920938463463374607431768211453
>     );
> ```
> 
> This currently emits the warning "help: try: `-3..=-1`", because the integer literal wraps around. This probably needs lots of casting and checks to make all the comparisons etc work. Though maybe there's a nicer way to do this. Will have to think about it. (or I guess we could also just ignore numbers >i128::MAX for now 🤔 )

The type of the integer is known in `LitKind::Int`, you can cast it using that (like `match int_ty { Isize => val as isize, _ => /* etc */ }`)

y21:
yep but what I mean is that, unless I'm overthinking this way too much, the logic will get very complicated very quickly here now that we need these casts and checks basically for every use of `min`/`max` (to get additions/less than/greater than/range checks etc. right) and it probably requires very large changes

Centri3:
That's a good point.

giraffate (commented):
> (or I guess we could also just ignore numbers >i128::MAX for now 🤔 )

IMO it would be better to ignore these numbers or just note it in `Known problems` section.

giraffate:
@bors r+

Thanks!

PR #11097, opened 2023-07-03
Clippy: Remove `TyCtyt::super_traits_of` public comment and test behavior

xFrednet:
The `TyCtyt::super_traits_of` function is actually not needed. This PR removes the reference and adds a test as well :)

changelog: none

### DIFF
--- a/clippy_utils/src/ty.rs
+++ b/clippy_utils/src/ty.rs
@@ -949,16 +949,15 @@ pub fn ty_is_fn_once_param<'tcx>(tcx: TyCtxt<'_>, ty: Ty<'tcx>, predicates: &'tc
         .iter()
         .try_fold(false, |found, p| {
             if let ty::ClauseKind::Trait(p) = p.kind().skip_binder()
-            && let ty::Param(self_ty) = p.trait_ref.self_ty().kind()
-            && ty.index == self_ty.index
-        {
-            // This should use `super_traits_of`, but that's a private function.
-            if p.trait_ref.def_id == fn_once_id {
-                return Some(true);
-            } else if p.trait_ref.def_id == fn_mut_id || p.trait_ref.def_id == fn_id {
-                return None;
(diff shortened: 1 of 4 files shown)


Jarcho:
That note is still accurate. `while_let_on_iterator` is working because the closure's kind is `FnOnce`.

I'm not sure if `ty_is_fn_once_param` is needed any more. It might be needed to handle cases where the closure is first assigned to a local, but that would have to be implemented first.

xFrednet:
Could you give an example for code that wouldn't work? I tested this ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=56922f378b04ac2583fd899a888edb0c)):

```rs
    // Lint if `FnOnce` is a super trait
    trait MySpecialFnMut: FnOnce() {}
    impl<T: FnOnce()> MySpecialFnMut for T {}
    fn f4(_: impl MySpecialFnMut) {}
    let mut it = 0..10;
    let x: Box<dyn MySpecialFnMut> = Box::new(|| {
        while let Some(x) = it.next() {
            if x % 2 == 0 {
                break;
            }
        }
    });
    f4(x);
```

and it lints just fine. I have the feeling I'm just overlooking something :sweat_smile:

Jarcho:
You're not missing anything. Looks like the closure's kind is now being set based on how it's used rather than how it's capable of being used. I guess the function can be removed then.

Jarcho:
Ping @xFrednet

`ty_is_fn_once_param` can just be removed and `get_enclosing_loop_or_multi_call_closure` can be simplified (`is_once` will always be false due to the closure kind check right before). 

@rustbot author

xFrednet:
Hey, year sorry, I wasn't sure if I could just remove the function and then I lost track of it :see_no_evil:

I'll close this in favor of https://github.com/rust-lang/rust-clippy/pull/11812, since that includes the suggested change.

PR #11107, opened 2023-07-04
New lint [`error_impl_error`]

### LINKED ISSUE #11101: New lint suggestion `Error impl Error`
### What it does

Warns on types which implement `std::error::Error` being named `Error`.

### Advantage

Removes needing to document against using this common pattern.
The reasons against this pattern is that it can reduce readability and requires error types be renamed when imported (which can further reduce readability).

When you have 20 types named `Error` in your project it can become annoying.

### Drawbacks

Adds an additional lint to clippy.

### Example

```rust
#[derive(Debug, thiserror::Error)]
pub enum Error { /* ... */ }
```

Could be written as:

```rust
#[derive(Debug, thiserror::Error)]
pub enum MySpecificError { /* ... */ }
```

```rust
#[derive(Debug)]
pub enum Error { /* ... */}
impl std::error::Error for Error { /* ... */ }
```

Could be written as:

```rust
#[derive(Debug)]
pub enum MySpecificError { /* ... */}
impl std::error::Error for Error { /* ... */ }
```

Centri3:
Closes #11101

changelog: New lint [`error_impl_error`]


clippy_lints/src/error_impl_error.rs
```diff
@@ -0,0 +1,76 @@
...
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for types named `Error` that implement `Error`.
+    ///
+    /// ### Why is this bad?
+    /// It can become confusing when a codebase has 20 types all named `Error`, requiring either
+    /// aliasing them in the `use` statement them or qualifying them like `my_module::Error`. This
+    /// severely hinders readability.
```
Jarcho, line 19: This isn't really a much of a readability. The alternative name would be `MyModuleError` or `MyCrateError`, which isn't really any worse to read.

Jarcho (commented):
This really should be restricted to types exported from the current module. Module private types have none of the listed downsides.

clippy_lints/src/error_impl_error.rs
Centri3: I'd say it's more comprehension, which is what I was *trying* to get at. Having a consistent name for a type is better than sometimes using `my_module::Error`, `MyModuleError` or just `Error`. Similar, `missing_enforced_import_renames`

Centri3:
Could be a configuration option

Jarcho:
I was referring to not linting only for module-private types, not crate-private types. I doubt there's any use in having a config for this as they are unusual enough as is, and importing them would be even rarer as only sub-modules could do that.

For types visible outside their own module I would still lint them as normal. If enough people want to ignore them, but still keep the lint otherwise then this can revisited, but I would rather not have a config option unless it's going to be useful.

Centri3:
ah, yeah that would likely be rare enough that it wouldn't be too useful

CHANGELOG.md
```diff
@@ -5464,4 +5465,5 @@ Released 2018-09-13
 [`accept-comment-above-statement`]: https://doc.rust-lang.org/clippy/lint_configuration.html#accept-comment-above-statement
 [`accept-comment-above-attributes`]: https://doc.rust-lang.org/clippy/lint_configuration.html#accept-comment-above-attributes
 [`allow-one-hash-in-raw-strings`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allow-one-hash-in-raw-strings
+[`allow-private-error`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allow-private-error
```
Jarcho, line 5468: Leftover from previous commit.

book/src/lint_configuration.md
```diff
@@ -730,3 +730,13 @@ Whether to allow `r#""#` when `r""` can be used
 * [`unnecessary_raw_string_hashes`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_raw_string_hashes)
 
 
+## `allow-private-error`
```
Jarcho, line 733: Leftover from previous commit.

clippy_lints/src/error_impl_error.rs
```diff
@@ -0,0 +1,87 @@
...
+                    ERROR_IMPL_ERROR,
+                    item.ident.span,
+                    "exported type alias named `Error` that implements `Error`",
+                );
+            },
+            ItemKind::Impl(imp) if let Some(trait_def_id) = imp.of_trait.and_then(|t| t.trait_def_id())
+                && error_def_id == trait_def_id
+                && let Some(def_id) = path_res(cx, imp.self_ty).opt_def_id().and_then(DefId::as_local)
+                && let hir_id = cx.tcx.hir().local_def_id_to_hir_id(def_id)
+                && let Node::Item(ty_item) = cx.tcx.hir().get(hir_id)
```
Jarcho, line 60: This would be better done using `TyCtxt::opt_item_name`. The containing module can be retrieved using `TyCtxt::parent_module_from_def_id`.

Centri3: I've gone with `opt_item_ident` instead as we also need the span of it, but it is definitely cleaner!

(though `parent_module_from_def_id` was already used :thinking:)

CHANGELOG.md
```diff
@@ -5465,3 +5466,5 @@ Released 2018-09-13
 [`accept-comment-above-attributes`]: https://doc.rust-lang.org/clippy/lint_configuration.html#accept-comment-above-attributes
 [`allow-one-hash-in-raw-strings`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allow-one-hash-in-raw-strings
 <!-- end autogenerated links to configuration documentation -->
+onfiguration.html#allow-private-error
```
Jarcho, line 5469: Looks like something went wrong here.

Centri3: I'm not sure why that keeps happening :confused: Will probably need to keep a closer eye on the changelog after rebasing

Jarcho:
Thank you. @bors r+

PR #11110, opened 2023-07-05
[`unnecessary_literal_unwrap`]: don't lint if binding initializer comes from expansion

y21:
Fixes https://github.com/rust-lang/rust-clippy/discussions/11109

changelog: [`unnecessary_literal_unwrap`]: don't lint if binding initializer comes from expansion


Jarcho:
Thank you. @bors r+

PR #11111, opened 2023-07-05
Fix regex lints for regex 1.9.0

Alexendoo:
regex 1.9.0 was [just released](https://blog.burntsushi.net/regex-internals/), which changes where the types are defined. Instead of updating the definitions to the ones in 1.9.0 this PR uses [`def_path_def_ids`](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.def_path_def_ids.html) on the canonical paths so that we don't have to worry about third party crate internals

This means that it still works with older regex versions too, and will for any future layout changes. I tested it with 1.8.4 and 1.9.0

changelog: [`INVALID_REGEX`], [`TRIVIAL_REGEX`]: now works with regex 1.9.0


giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #10847, opened 2023-05-30
New Changelog formatting


blyxyas:
\[Related to this [Zulip Meeting](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202023-05-30)\]

## Grouping issues

This issue is a discussion around changing the changelog's formatting. Several solutions came out in the 2023-05-30 meeting, this issue discusses the last one (the conclusive one).

The conclusion that we reached on this meeting around the changelog is that we should group all small fixes (usually fixing false-positives and very small suggestion changes), that the user will have to click on. This can be done using the `<details>` and `<summary>` tags.

(The person making the changelog will decide what is a small change)

We would keep the current sections, and add a new one.
(Note that it is tempting using the `Others` section to group all small commits, but sometimes, changes that cannot be categorized in any of the other categories are significant. So we'll have to create a new section)

```
### New Lints
### Moves and Deprecations
### Enhancements
### False Positive Fixes
### Suggestion Fixes/Improvements
### ICE Fixes
### Documentation Improvements
### Others

<details><summary><h3>Minor fixes</h3></summary>
</details>
```

<details><summary><h3>Minor fixes</h3></summary>
    
- Someone fixed a typo yay :D #100000

</details>

Note the double space between `### Others` and `<details>`. And the space between `<details>`, the content, and the final `</details>` (Github is pretty picky when rendering Markdown files that use HTML, but there isn't any other way to create dropdowns in Markdown)

---

## The PR range

We would also replace the commit range (which is not very significant, as a single PR can have many commits) with a new merged PRs range.

The filter to use here would be something like this:

```
is:pr is:closed merged:<last release, YYYY-MM-DD>..<this release> base:master sort:merged-desc 
```

We could display it raw as a range:
#10669..#10841 

(This approach doesn't make much sense, as PRs aren't sequential in any way)

Or maybe just provide a link to view it in Github

[**View 80 PRs merged since 1.69**](https://github.com/rust-lang/rust-clippy/pulls?q=is%3Apr+is%3Aclosed+merged%3A2023-04-20..2023-06-01+base%3Amaster+sort%3Amerged-desc+)

---

Joining these 2 concepts would give us a much more legible changelog.

cc @xFrednet

xFrednet:
Hey @blyxyas, thank you for creating the issue. I really like the PR range and will switch to it in the next changelog.

---

For the second part, about minor changes, you wrote:

[...]

PR #11114, opened 2023-07-06
Fix changelog PR listings, create them automatically in `fetch_prs_between.sh`

Alexendoo:
changelog: none

`fetch_prs_between.sh` now finds the date of the newest and oldest merges from GitHub to use in the range, this does add a dependency on https://github.com/cli/cli

It also no longer prints rollups/merges that come from rustc, so only clippy changes should be printed

r? @xFrednet

cc #10847

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ document.
 
 Current stable, released 2023-06-01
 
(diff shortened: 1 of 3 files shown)


xFrednet (approved):

xFrednet:
This is AWESOME! I knew there was a reason, why I haven't started writing the changelog for the next release :wink:. I really like this change, it'll make writing the changelog more present. Thank you very much!

@bors r+

PR #11115, opened 2023-07-06
New lint [`filter_map_bool_then`]

### LINKED ISSUE #9098: Check for filter_map combined with `bool::then`, suggest `filter` then `map`
### What it does

The combination of `filter_map` with `bool::then` can often be more readably written as a `filter` followed by a `map`, which avoids multiple levels of closure nesting and reduces indentation.

### Lint Name

filter_then_map

### Category

style

### Advantage

- Avoids multiple levels of closure and parenthesis nesting.
- Reduces indentation.

### Drawbacks

- While the compiler should generally be able to optimize `filter` then `map`, this may potentially miss optimization opportunities.

### Example

```rust
let v = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30];
collection
    .iter()
    .filter_map(|n| (n % 2 == 0).then(|| complex_calculation(n * 100)))
```

Could be written as:

```rust
let v = vec![1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20, 25, 30];
collection
    .iter()
    .filter(|n| n % 2 == 0)
    .map(|n| complex_calculation(n * 100))
```

(The case I ran into that prompted this had a dozen lines in place of `complex_calculation`.)

Centri3:
Closes #9098

changelog: New lint [`filter_map_bool_then`]

### DIFF
--- a/clippy_lints/src/methods/filter_map_bool_then.rs
+++ b/clippy_lints/src/methods/filter_map_bool_then.rs
@@ -1,0 +1,39 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::paths::BOOL_THEN;
+use clippy_utils::source::snippet_opt;
+use clippy_utils::{is_from_proc_macro, match_def_path, peel_blocks};
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::{LateContext, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_span::Span;
+
+use super::FILTER_MAP_BOOL_THEN;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, arg: &Expr<'_>, call_span: Span) {
+    if !in_external_macro(cx.sess(), expr.span)
+        && let ExprKind::Closure(closure) = arg.kind
+        && let body = peel_blocks(cx.tcx.hir().body(closure.body).value)
+        && let ExprKind::MethodCall(_, recv, [then_arg], _) = body.kind
+        && let ExprKind::Closure(then_closure) = then_arg.kind
+        && let then_body = peel_blocks(cx.tcx.hir().body(then_closure.body).value)
+        && let Some(def_id) = cx.typeck_results().type_dependent_def_id(body.hir_id)
+        && match_def_path(cx, def_id, &BOOL_THEN)
+        && !is_from_proc_macro(cx, expr)
+        && let Some(decl_snippet) = closure.fn_arg_span.and_then(|s| snippet_opt(cx, s))
+        // NOTE: This will include the `()` (parenthesis) around it. Maybe there's some utils method
+        // to remove them? `unused_parens` will already take care of this but it may be nice.
+        && let Some(filter) = snippet_opt(cx, recv.span)
+        && let Some(map) = snippet_opt(cx, then_body.span)
+    {
+        span_lint_and_sugg(
+            cx,
+            FILTER_MAP_BOOL_THEN,
+            call_span,
+            "usage of `bool::then` in `filter_map`",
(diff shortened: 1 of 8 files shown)


clippy_lints/src/methods/mod.rs
```diff
@@ -3378,6 +3379,35 @@ declare_clippy_lint! {
     "calling `Stdin::read_line`, then trying to parse it without first trimming"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `bool::then` in `Iterator::filter_map`.
+    ///
+    /// ### Why is this bad?
+    /// This can be more concisely written with `filter` then `map` instead.
```
Jarcho, line 3387: Minor nit: 'concise' is not the right description. It reduces nesting and has a small readability gain due to separating the filtering step from the transformation step.

Otherwise this looks fine.

Jarcho:
Revisiting this, there is one notable issue. `filter` has a different signature for, taking `Fn(&T)`, instead of `Fn(T)`. Easiest way to handle this would be to limit the lint to `Copy` types and adjust the closure on `filter` to be `|&<decl>|`.

clippy_lints/src/methods/filter_map_bool_then.rs
```diff
@@ -0,0 +1,51 @@
...
+        && match_def_path(cx, def_id, &BOOL_THEN)
+        && !is_from_proc_macro(cx, expr)
+        && let Some(param_snippet) = snippet_opt(cx, param.span)
+        && let Some(filter) =
+            // I'm fairly sure removing all parenthesis is fine, as something like
+            // `(true == true).then(<etc>)` would no longer need them as `bool::then` is removed. I
+            // can't think of any other case that would trigger the lint yet need parenthesis.
+            //
+            // Also, If they were uneven (like `((1 + 1)`) then building the AST would fail anyway
+            // so that's thankfully a non-issue
```
Jarcho, line 37: This is not true (e.g. `((x)?)`). Even removing a matching number of parenthesis doesn't work (e.g. `((x) == (y))`).

clippy_lints/src/methods/filter_map_bool_then.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use super::FILTER_MAP_BOOL_THEN;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, arg: &Expr<'_>, call_span: Span) {
+    if !in_external_macro(cx.sess(), expr.span)
+        && let ExprKind::Closure(closure) = arg.kind
+        && let body = cx.tcx.hir().body(closure.body)
+        && let value = peel_blocks(body.value)
+        // Indexing should be fine as `filter` always has 1 input, we unfortunately need both
+        // `inputs` and `params` here as we need both the type and the span
```
Jarcho, line 20: I don't think the method was ever confirmed to be `Iterator::filter_map`.

Centri3: It is, it's under ("filter_map", [arg])

clippy_lints/src/methods/filter_map_bool_then.rs
Centri3: ah, yeah I'll remove it then. That's unfortunate though

clippy_lints/src/methods/filter_map_bool_then.rs
Jarcho: That only constrains the method name and arity. It doesn't limit it to being the one defined on `Iterator`.

Centri3: Ah I see now. Yeah I'll add an `is_trait_method` call

Jarcho (approved):
Looks good. r=me after rebase.

PR #11131, opened 2023-07-09
[significant_drop_tightening] Add #11125 test

### LINKED ISSUE #11125: Incorrect suggestion for significant-drop-tightening
### Summary

I've hit a false positive, or at least a bad suggestion, with the `significant-drop-tightening` lint where a mutex is used in two locations, but Clippy's suggestion doesn't consider the second usage of the locked value.

Initially reported in https://github.com/rust-lang/rust-clippy/issues/9399#issuecomment-1615885088.
/cc @c410-f3r

### Lint Name

significant-drop-tightening

### Reproducer

The related code: 
https://github.com/Thomasdezeeuw/heph/blob/8725fe9ed871f135b7ab87b5969eb1213121a242/inbox/src/lib.rs#L523-L529

For which Clippy suggests:

```
error: temporary with significant `Drop` can be early dropped
   --> inbox/src/lib.rs:523:21
    |
522 |           if let Some(waker) = self.registered_waker.take() {
    |  ___________________________________________________________-
523 | |             let mut sender_wakers = self.channel.sender_wakers.lock().unwrap();
    | |                     ^^^^^^^^^^^^^
524 | |             let idx = sender_wakers.iter().position(|w| w.will_wake(&waker));
525 | |             if let Some(idx) = idx {
...   |
529 | |             }
530 | |         }
    | |_________- temporary `sender_wakers` is currently being dropped at the end of its contained scope
    |
    = note: this might lead to unnecessary resource contention
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_tightening
    = note: `-D clippy::significant-drop-tightening` implied by `-D clippy::nursery`
help: merge the temporary construction with its single usage
    |
523 ~
524 +             let idx = self.channel.sender_wakers.lock().unwrap().position(|w| w.will_wake(&waker));
    |
help: remove separated single usage
    |
524 -             let idx = sender_wakers.iter().position(|w| w.will_wake(&waker));
524 +
    |
```

It doesn't seem to realize that `sender_wakers` is used twice, once to iterate over (line 524) and then to mutate it (line 526).

### Version

```text
rustc 1.72.0-nightly (cb80ff132 2023-07-07)
binary: rustc
commit-hash: cb80ff132a0e9aa71529b701427e4e6c243b58df
commit-date: 2023-07-07
release: 1.72.0-nightly
LLVM version: 16.0.5

clippy 0.1.72 (cb80ff1 2023-07-07)
```

### Additional Labels

_No response_

c410-f3r:
Fix #11125

---

changelog: none

### DIFF
--- a/tests/ui/significant_drop_tightening.fixed
+++ b/tests/ui/significant_drop_tightening.fixed
@@ -2,7 +2,15 @@
 
 #![warn(clippy::significant_drop_tightening)]
 
-use std::sync::Mutex;
+use std::{
+    cell::UnsafeCell,
+    mem::{drop as unlock, MaybeUninit},
+    sync::{
+        atomic::{AtomicU64, AtomicUsize},
+        Mutex,
+    },
+    task::Waker,
+};
 
 pub fn complex_return_triggers_the_lint() -> i32 {
     fn foo() -> i32 {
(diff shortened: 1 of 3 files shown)


xFrednet:
r% @blyxyas 

(xFrednet has picked a reviewer for you, use r% to override)

blyxyas (changes requested):
Could you simplify this? We only need `sender_wakers` (or a `Mutex<Vec<Waker>>`), the test doesn't need a lot of the original block to still be effective (e.g. the initial `if` stmt. or `SendValue`)

c410-f3r:
Is it really necessary to trim-down a self-contained use-case provided by a user of this lint?

c410-f3r:
I see. Thank you

blyxyas:
But looking at the current `tests/ui/significant_drop_tightening.rs`, it seems like a simplified test was provided in #11129. I'm not sure how to proceed here.

xFrednet:
> But looking at the current `tests/ui/significant_drop_tightening.rs`, it seems like a simplified test was provided in #11129. I'm not sure how to proceed here.

Is my understanding correct, that this PR only adds a test and that a simpler example of the same problem has been added in  #11129?

If this test doesn't check a different behavior or variant of one, I'd suggest closing the PR. If it's a variant, that checks something slightly different, it would be good to merge it. Merging it either way would not be wrong :)

blyxyas:
Well, I think we'll do an exception this time. @c410-f3r, could you fix the conflicting files

blyxyas:
just a little mini-ping
@c410-f3r

c410-f3r:
Thank you for the heads-up.

Unfortunately, I don't have the motivation to continue pursuing this issue.

PR #11135, opened 2023-07-09
Fix `unwrap_or_else_default` false positive

smoelius:
This PR fixes a false positive in the handling of `unwrap_or_else` with a default value when the value is needed for type inference.

An easy example to exhibit the false positive is the following:
```rust
    let option = None;
    option.unwrap_or_else(Vec::new).push(1);
```
The following code would not compile, because the fact that the value is a `Vec` has been lost:
```rust
    let option = None;
    option.unwrap_or_default().push(1);
```
The fix is to:
- implement a heuristic to tell whether an expression's type can be determined purely from its subexpressions, and the arguments and locals they use;
- apply the heuristic to `unwrap_or_else`'s receiver.

The heuristic returns false when applied to `option` in the above example, but it returns true when applied to `option` in either of the following examples:
```rust
    let option: Option<Vec<u64>> = None;
    option.unwrap_or_else(Vec::new).push(1);
```
```rust
    let option = None::<Vec<u64>>;
    option.unwrap_or_else(Vec::new).push(1);
```

(Aside: https://github.com/rust-lang/rust-clippy/pull/10120 unfairly contained multiple changes in one PR. I am trying to break that PR up into smaller pieces.)

---

changelog: FP: [`unwrap_or_else_default`]: No longer lints if the default value is needed for type inference


Centri3:
Will take a look soon enough ^^

clippy_utils/src/ty/type_certainty/certainty.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    }
+
+    pub fn with_def_id(self, def_id: DefId) -> Certainty {
+        match self {
+            Certainty::Uncertain => Certainty::Uncertain,
+            Certainty::Certain(_) => Certainty::Certain(Some(def_id)),
+            Certainty::Contradiction => Certainty::Contradiction,
+        }
+    }
```
Centri3, line 112: nit: I feel like these two should use `if let`/`matches!` instead, since the others are unchanged

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+use rustc_span::{Span, Symbol};
+
+mod certainty;
+use certainty::{join, meet, Certainty, Meet};
+
+pub fn expr_type_is_certain(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
+    expr_type_certainty(cx, expr).is_certain()
+}
+
+fn expr_type_certainty(cx: &LateContext<'_>, expr: &Expr<'_>) -> Certainty {
```
Centri3, line 32: nit: I think this could use a `Visitor` instead, though perhaps I'm wrong

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        return false;
+    };
+
+    let generics = cx.tcx.generics_of(callee_def_id);
+    let fn_sig = cx.tcx.fn_sig(callee_def_id).skip_binder();
+
+    (0..(generics.parent_count + generics.params.len()) as u32).all(|index| {
+        fn_sig.inputs().iter().any(|input_ty| {
+            input_ty.skip_binder().walk().any(|arg| {
+                if let GenericArgKind::Type(ty) = arg.unpack() {
```
Centri3, line 294: We could perhaps use `matches!` here to prevent the `else` branch (like, `matches!(arg.unpack(), GenericArgKind::Type(ty) if ty.is_param(index))`)

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+                    ty.is_param(index)
+                } else {
+                    false
+                }
+            })
+        })
+    })
```
Centri3, line 301: I'm not quite able to suggest something better rn, but this block is quite hard to read.

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+            if let Some(method_def_id) = cx.typeck_results().type_dependent_def_id(expr.hir_id)
+                && let Some(self_ty_def_id) = adt_def_id(self_ty(cx, method_def_id))
+            {
+                receiver_type_certainty = receiver_type_certainty.with_def_id(self_ty_def_id);
+            };
+            let lhs = path_segment_certainty(cx, receiver_type_certainty, method, false);
+            let rhs = if type_is_inferrable_from_arguments(cx, expr) {
+                meet(
+                    std::iter::once(receiver_type_certainty).chain(args.iter().map(|arg| expr_type_certainty(cx, arg))),
+                )
```
Centri3, line 65: nit: This could perhaps use `then` followed by conversion to `Certainty` using `map_or`. Though I'm not sure if that's more readable here.

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        // - `Option::None::<Vec<u64>>` // certain; parent (i.e., `Option`) is uncertain
+        Res::Def(_, def_id) => {
+            // Checking `res_generics_def_id(..)` before calling `generics_of` avoids an ICE.
+            if cx.tcx.res_generics_def_id(path_segment.res).is_some() {
+                let generics = cx.tcx.generics_of(def_id);
+                let lhs = if (parent_certainty.is_certain() || generics.parent_count == 0) && generics.params.is_empty()
+                {
+                    Certainty::Certain(None)
+                } else {
+                    Certainty::Uncertain
```
Centri3, line 210: nit: This could perhaps use `then` followed by conversion to `Certainty` using `map_or`. Though I'm not sure if that's more readable here.

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+        }
+    }
+}
+
+fn type_certainty(cx: &LateContext<'_>, ty: &hir::Ty<'_>) -> Certainty {
+    // Handle `TyKind::Path` specially so that its `DefId` can be preserved.
+    if let TyKind::Path(qpath) = &ty.kind {
+        qpath_certainty(cx, qpath, true)
+    } else {
+        let mut visitor = CertaintyVisitor::new(cx);
```
Centri3, line 136: nit: Unnecessary `else` branch (afaict?)

clippy_utils/src/ty/type_certainty/mod.rs
```diff
@@ -0,0 +1,310 @@
...
+//!
+//! This is, in some sense, a counterpart to `let_unit_value`'s `expr_needs_inferred_result`.
+//! Intuitively, that function determines whether an expression's type is needed for type inference,
+//! whereas `expr_type_is_certain` determines whether type inference is needed for an expression's
+//! type.
+//!
+//! As a heuristic, `expr_type_is_certain` may produce false negatives, but a false positive should
+//! be considered a bug.
+
+#![allow(clippy::wrong_self_convention)]
```
Centri3, line 14: I don't like bulk-allowing a lint like this, especially one like `wrong_self_convention`. We should probably put this on individual methods instead, as `expect` (PS: Which method is actually triggering it? I can't see any out of the ordinary.)

Centri3 (commented):
This is quite the PR, thanks for taking this on!

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: I'm not sure a visitor would make sense here. When a node has multiple children, the results of the respective recursive calls need to be combined in some way (e.g., `meet`ed or `join`ed). IMHO, visitors are not really good for that sort of situation. This could be a failure of imagination on my part, though.

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: Your point is very well taken. I moved the `.walk().any(...)` code into its own function and added a comment to help explain what this block does.

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: Is this what you had in mind?
```rust
            let rhs = type_is_inferrable_from_arguments(cx, expr)
                .then(|| meet(args.iter().map(|arg| expr_type_certainty(cx, arg))))
                .unwrap_or(Certainty::Uncertain);
```
Do you think it is more clear?

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: Same questions as above (is this what you had in mind and it is more clear?):
```rust
                let lhs = ((parent_certainty.is_certain() || generics.parent_count == 0) && generics.params.is_empty())
                    .then(|| Certainty::Certain(None))
                    .unwrap_or(Certainty::Uncertain);
```

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: I think the `else` is necessary. I expanded the comment preceding the `if` in attempt to better explain why.

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: > Which method is actually triggering it?

AFAICT, none are. (I can't remember why I included this originally.)

smoelius:
> This is quite the PR, thanks for taking this on!

Thank you.

FWIW, I am highly sympathetic to comments like https://github.com/rust-lang/rust-clippy/pull/11135#discussion_r1259168431. Please let me know anything I can do to make the code more clear.

clippy_utils/src/ty/type_certainty/mod.rs
Centri3: Yeah it's not :sweat_smile: let's keep it the way it is.

clippy_utils/src/ty/type_certainty/mod.rs
Centri3: Oh, I meant that you could inline the `else`, like,
```rust
if let TyKind::Path(qpath) = &ty.kind {
    return qpath_certainy(cx, qpath, true)
}

let mut visitor = CertainyVisitor::new(cx);
// etc...
```

Centri3:
Yeah it looks good now :+1:

clippy_utils/src/ty/type_certainty/mod.rs
smoelius: Sorry for misunderstanding. Fixed.

clippy_utils/src/ty/type_certainty/certainty.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl TryJoin for Option<DefId> {
+    fn try_join(self, other: Self) -> Option<Self> {
+        match (self, other) {
+            (Some(lhs), Some(rhs)) => {
+                if lhs == rhs {
+                    Some(Some(lhs))
```
Centri3, line 46: nit: This is the perfect place to use `then_some`: `(lhs == rhs).then_some(Some(lhs))`

clippy_utils/src/ty/type_certainty/certainty.rs
```diff
@@ -0,0 +1,128 @@
...
+pub trait TryJoin: Sized {
+    fn try_join(self, other: Self) -> Option<Self>;
+}
+
+impl Meet for Option<DefId> {
+    fn meet(self, other: Self) -> Self {
+        match (self, other) {
+            (None, _) | (_, None) => None,
+            (Some(lhs), Some(rhs)) => {
+                if lhs == rhs {
```
Centri3, line 31: Same here as well (`then_some`); missed this one before

Centri3 (approved):
Looks good ^^ Let's just use `then_some` in `meet` then I think this is done :D

smoelius:
> Looks good ^^ Let's just use `then_some` in `meet` then I think this is done :D

Thanks, @Centri3 . Should I squash this one commit?

Centri3:
Yeah if you could squash them that would be great

Centri3:
Thanks! @bors r+

PR #11136, opened 2023-07-10
[`read_line_without_trim`]: detect string literal comparison and `.ends_with()` calls

y21:
This lint now also realizes that a comparison like `s == "foo"` and calls such as `s.ends_with("foo")` will fail if `s` was initialized by a call to `Stdin::read_line` (because of the trailing newline).

changelog: [`read_line_without_trim`]: detect string literal comparison and `.ends_with()` calls

r? @giraffate assigning you because you reviewed #10970 that added this lint, so this is kinda a followup PR ^^

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -3420,11 +3420,12 @@ declare_clippy_lint! {
 declare_clippy_lint! {
     /// ### What it does
     /// Looks for calls to [`Stdin::read_line`] to read a line from the standard input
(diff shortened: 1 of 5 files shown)


y21:
wow, this PR is ancient :D

(The first commit is the actual relevant addition, the second one moves some things around which causes the diff to look a bit annoying, so it's probably better reviewed commit-by-commit)

dswij (approved):
Thanks for this! Sorry that the review take awhile.

Can you help to rebase this? I think it's good to merge afterward.

dswij (approved):

PR #11137, opened 2023-07-11
Add test dependencies lock file

oli-obk:
This should make the test suite more reliable. CI and new checkouts do not have a lockfile, so they pick up random new crates. This can cause build breakage and random test failures if someone on crates.io breaks semver accidentally. After this PR there should be no more issues like that.

changelog: use `Cargo.lock` for our test suite's dependencies


flip1995:
I guess this is a good chance to point out https://github.com/rust-lang/rust-clippy/pull/11045?notification_referrer_id=NT_kwDOAJSxB7I2OTEwMjgyMTI1Ojk3NDQ2NDc#issuecomment-1612731666

PR #11140, opened 2023-07-12
New lint [`four_forward_slashes`]

### LINKED ISSUE #9212: Lint for 4 / (////) to help detect broken doc comments
### What it does

Detects four consecutive forward slash characters.

This is almost always indicative of a broken doc comment.

### Lint Name

four-forward-slashes

### Category

suspicious

### Advantage

- The doc comment appears as expected.

### Drawbacks

Users may have a non-standard style that uses 4 forward slashes semantically.

### Example

```rust
//// The life total of a unit
struct Life(u8);
```

Was likely intended to be:

```rust
/// The life total of a unit
struct Life(u8);
```

Centri3:
Closes #9212

changelog: New lint [`four_forward_slashes`]

### DIFF
--- a/clippy_lints/src/four_forward_slashes.rs
+++ b/clippy_lints/src/four_forward_slashes.rs
@@ -1,0 +1,82 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, source::snippet_opt};
+use rustc_errors::Applicability;
+use rustc_hir::Item;
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::{Span, SyntaxContext};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for outer doc comments written with 4 forward slashes (`////`).
+    ///
+    /// ### Why is this bad?
+    /// This is (probably) a typo, and results in it not being a doc comment; just a regular
+    /// comment.
+    ///
+    /// ### Example
+    /// ```rust
+    /// //// My amazing data structure
+    /// pub struct Foo {
+    ///     // ...
+    /// }
+    /// ```
+    ///
+    /// Use instead:
+    /// ```rust
+    /// /// My amazing data structure
+    /// pub struct Foo {
+    ///     // ...
+    /// }
+    /// ```
+    #[clippy::version = "1.72.0"]
+    pub FOUR_FORWARD_SLASHES,
+    suspicious,
+    "comments with 4 forward slashes (`////`) likely intended to be doc comments (`///`)"
+}
+declare_lint_pass!(FourForwardSlashes => [FOUR_FORWARD_SLASHES]);
+
+impl<'tcx> LateLintPass<'tcx> for FourForwardSlashes {
+    fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
+        if item.span.from_expansion() {
+            return;
+        }
+        let src = cx.sess().source_map();
+        let item_and_attrs_span = cx
+            .tcx
+            .hir()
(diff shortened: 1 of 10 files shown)


tests/ui/four_forward_slashes.rs
```diff
@@ -0,0 +1,44 @@
+//// first line borked doc comment. doesn't combust!
```
Centri3, line 1: I think we'll need a separate test file for this, as we can't have `#![...]` below items

xFrednet: Yeah, sounds reasonable :)

xFrednet:
r$ @blyxyas 

(xFrednet has picked a reviewer for you, use r$ to override)

clippy_lints/src/four_forward_slashes.rs
```diff
@@ -0,0 +1,82 @@
...
+            };
+            let contents = contents.trim();
+            if contents.is_empty() {
+                break;
+            }
+            if contents.starts_with("////") {
+                let bounds = file.line_bounds(line);
+                let span = Span::new(bounds.start, bounds.end, SyntaxContext::root(), None);
+
+                if snippet_opt(cx, span).is_some_and(|s| s.trim().starts_with("////")) {
```
blyxyas, line 65: Is checking for "////" two times necessary?

tests/ui/four_forward_slashes.stderr
```diff
@@ -0,0 +1,75 @@
...
+error: comment with 4 forward slashes (`////`). This looks like a doc comment, but it isn't
+  --> $DIR/four_forward_slashes.rs:20:1
+   |
+LL | / //// whoops
+LL | | //// two borked comments!
+   | |_
+   |
+help: make this a doc comment by removing one `/`
+   |
+LL + /// whoops
```
blyxyas, line 47: Could this be compacted into a single suggestion / emission?

blyxyas (commented):
I cannot stop thinking about [`rustc_ast::util::comments::gather_comments`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_ast/util/comments/fn.gather_comments.html), it returns a `Vec<Comment>` and that could be used to check if the comment starts with `//` or `////` (I'm not sure `Comment` also includes the initial `//`).

It would be simpler, but we'd have to do a benchmark to verify speed. Maybe SPEEDTEST can help with that?

tests/ui/four_forward_slashes.stderr
Centri3: Tried to, it was `span_suggestion_verbose` before. Would prefer it but not sure if it's possible sadly

clippy_lints/src/four_forward_slashes.rs
Centri3: That's the quasi-`is_from_proc_macro` check, so kinda.

Centri3:
`gather_comments` may work, but from what I can see it'd do a lot more work for the same result. Even just allocating an empty `Vec`, then increasing its capacity every time a comment is found, would be slow.

Perhaps we can use the lexer though. I hadn't thought to and that may be nicer

Centri3:
Unfortunately `gather_comments` won't work. It doesn't give the span of the comment, only the column it starts at (`Comment.pos`).

clippy_lints/src/four_forward_slashes.rs
Centri3: Ok it's unnecessary since I've just realized the prior would already cover that :sweat_smile:

blyxyas (approved):
LGTM, Thanks! :heart: 
cc @xFrednet

xFrednet:
I'm sadly busy until Tuesday/Wednesday, but then I'll do the last check. :)

Issue #11144, opened 2023-07-12
`tuple_array_conversion` false positive on destructured tuple values


MingweiSamuel:
### Summary

The `tuple_array_conversion` seems to be way to zealous in detecting the original tuple identifiers, generating false positives even when the names are assigned through destructuring. These all trigger it even though it makes no sense to:
```rust
let (a, (_b, c)) = get_tuple_a();
let _ = [a, c];

let (a, [_b, c]) = get_tuple_b();
let _ = [a, c];

let (a, Wrap(b)) = get_tuple_c();
let _ = [a, b];
```

full code below

### Lint Name

tuple_array_conversion

### Reproducer

full code:
```rust
fn get_tuple_a() -> (usize, (usize, usize)) {
    (1, (2, 3))
}

fn get_tuple_b() -> (usize, [usize; 2]) {
    (1, [2, 3])
}

struct Wrap<T>(pub T);
fn get_tuple_c() -> (usize, Wrap<usize>) {
    (1, Wrap(2))
}

fn main() {
    let (a, (_b, c)) = get_tuple_a();
    let _ = [a, c];

    let (a, [_b, c]) = get_tuple_b();
    let _ = [a, c];
    
    let (a, Wrap(b)) = get_tuple_c();
    let _ = [a, b];
}
```

```rust
    Checking playground v0.0.1 (/playground)
warning: it looks like you're trying to convert a tuple to an array
  --> src/main.rs:16:13
   |
16 |     let _ = [a, c];
   |             ^^^^^^
   |
   = help: use `.into()` instead, or `<[T; N]>::from` if type annotations are needed
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tuple_array_conversions
   = note: `#[warn(clippy::tuple_array_conversions)]` on by default

warning: it looks like you're trying to convert a tuple to an array
  --> src/main.rs:19:13
   |
19 |     let _ = [a, c];
   |             ^^^^^^
   |
   = help: use `.into()` instead, or `<[T; N]>::from` if type annotations are needed
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tuple_array_conversions

warning: it looks like you're trying to convert a tuple to an array
  --> src/main.rs:22:13
   |
22 |     let _ = [a, b];
   |             ^^^^^^
   |
   = help: use `.into()` instead, or `<[T; N]>::from` if type annotations are needed
[...]

Issue #11124, opened 2023-07-07
`tuple_array_conversions` triggers even when binding is used for more than tuple-array conversion


mwkmwkmwk:
### Summary

A lint about tuple-array conversions triggers when bindings created by tuple destructuring are used to create an array (or the other way around). However, the lint doesn't check whether this is the *only* use of such bindings, in which case using `into()` would harm readability by forcing changes to other binding uses.

### Lint Name

tuple_array_conversions

### Reproducer

I tried this code:

```rust
fn get_tuple() -> (i32, i32) {
    (1, 2)
}

fn iter_arrays() -> impl Iterator<Item = [i32; 2]> {
    [].into_iter()
}

fn do_something(_: i32) -> i32 {
    3
}

fn main() {
    let (a, b) = get_tuple();
    for x in [a, b] {
        println!("use both: {x}");
    }
    do_something(a);
    iter_arrays()
        .map(|[a, b]| (do_something(a), (a, b)))
        .for_each(drop);
}
```

I saw this happen:

```
$ cargo clippy
    Checking clt v0.1.0 (/home/mwk/clt)
warning: it looks like you're trying to convert a tuple to an array
  --> src/main.rs:15:14
   |
15 |     for x in [a, b] {
   |              ^^^^^^
   |
   = help: use `.into()` instead, or `<[T; N]>::from` if type annotations are needed
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tuple_array_conversions
   = note: `#[warn(clippy::tuple_array_conversions)]` on by default

warning: it looks like you're trying to convert an array to a tuple
  --> src/main.rs:20:41
   |
20 |         .map(|[a, b]| (do_something(a), (a, b)))
   |                                         ^^^^^^
   |
   = help: use `.into()` instead, or `<(T0, T1, ..., Tn)>::from` if type annotations are needed
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tuple_array_conversions

warning: `clt` (bin "clt") generated 2 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.08s
```

I expected to see this happen:

(no warning)

### Version

```text
rustc 1.72.0-nightly (85bf07972 2023-07-06)
binary: rustc
commit-hash: 85bf07972a1041b9e25393b803d0e006bec3eaaf
commit-date: 2023-07-06
host: x86_64-unknown-linux-gnu
release: 1.72.0-nightly
[...]

Issue #11100, opened 2023-07-04
tuple_array_conversions misses implicit &mut -> & conversion subtelty


HadrienG2:
### Summary

I have code that uses a double buffering pattern, and as a result has multiple instances of...

- Starting from `&mut [T; 2]` (representing a borrow of the underlying data store) and turning it into the heterogeneous tuple `(&T, &mut T)` (representing input and output accessors).
- Starting from such a heterogeneous tuple and downgrading it into a `[&T; 2]` (in cases where the distinction between input and output is not necessary and I just want read only access to everything).

These conversions are performed using pairs of array and slice patterns, which clippy's `tuple_array_conversions` lint will incorrectly suggest turning into `from()` or `into()` statements. The tuple <-> array From/Into impl actually are not powerful enough to do this.

### Lint Name

tuple_array_conversions

### Reproducer

I tried this code:

```rust
pub struct Evolving([usize; 2]);

impl Evolving {
    pub fn in_out(&mut self) -> (&usize, &mut usize) {
        let [input, output] = &mut self.0;
        (input, output)
    }
}
```

I saw this happen:

```text
warning: it looks like you're trying to convert an array to a tuple
   --> data/src/concentration/mod.rs:124:9
    |
124 |         (input, output)
    |         ^^^^^^^^^^^^^^^
    |
    = help: use `.into()` instead, or `<(T0, T1, ..., Tn)>::from` if type annotations are needed
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tuple_array_conversions
    = note: `#[warn(clippy::tuple_array_conversions)]` on by default

```

I expected to see this happen: clippy understands that this conversion cannot be done via into() and does not suggest it.

### Version

```text
rustc 1.72.0-nightly (0ab38e95b 2023-07-03)
binary: rustc
commit-hash: 0ab38e95bb1cbf0bd038d359bdecbfa501f003a7
commit-date: 2023-07-03
host: x86_64-unknown-linux-gnu
release: 1.72.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

@rustbot label +I-suggestion-causes-error +L-complexity

Issue #11085, opened 2023-07-03
`tuple_array_conversions` suggested code erases binding names and hides asymmetry


oxalica:
### Summary

In case of tuple elements are asymmetric, eg. `(src, dst)`, simply `tuple.into()` erase all binding names and make it seem like all elements are interchangeable, also hiding the asymmetry.

```rust
let [src, dest]: [_; 2] = args.try_into().ok()?;
operations.push((src, dest)); // <- warning
```

Suggested code is not obvious what each element means. It seems like to be an iterator which is general over array length, but is actually not.

```rust
operations.push(<[_; 2]>::try_from(args).ok()?.into());
```

### Lint Name

tuple_array_conversions

### Reproducer

I tried this code:

```rust
use std::path::PathBuf;

pub fn parse_ops_from_args_warn(args: Vec<PathBuf>, operations: &mut Vec<(PathBuf, PathBuf)>) -> Option<()> {
    let [src, dest]: [_; 2] = args.try_into().ok()?;
    operations.push((src, dest)); // <- warning
    Some(())
}

pub fn parse_ops_from_args_suggested(args: Vec<PathBuf>, operations: &mut Vec<(PathBuf, PathBuf)>) -> Option<()> {
    operations.push(<[_; 2]>::try_from(args).ok()?.into());
    Some(())
}
```

I saw this happen:

```
warning: it looks like you're trying to convert an array to a tuple
 --> src/lib.rs:5:21
  |
5 |     operations.push((src, dest));
  |                     ^^^^^^^^^^^
  |
  = help: use `.into()` instead, or `<(T0, T1, ..., Tn)>::from` if type annotations are needed
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#tuple_array_conversions
  = note: `#[warn(clippy::tuple_array_conversions)]` on by default
```

I expected to see this happen: no warnings.

### Version

```text
Clippy 0.1.72 (2023-07-02 839e9a6)

On Rust playground nightly.
```

Centri3:
I don't see any reason you can't do something like:
```rust
let (src, dest) = <[_; 2]>::try_from(args).ok()?.into();
operations.push((src, dest));
```
(untested, could summon demons)

Or even name it `src_and_dest` and don't destructure at all (though that's awful for anything with more than 2 elements)

Centri3:
> No. This is purely making things complicated to workaround the warning.

Completely agree ^^

[...]

PR #11146, opened 2023-07-12
[`tuple_array_conversions`]: move from `complexity` to `nursery`

MingweiSamuel:
The lint suggestion is arguably often less readable and more complex than the original code.

For example, which of the following is the most readable:
```rust
let _vertices = edges.flat_map(|(src, dst)| [src, dst]);
let _vertices = edges.flat_map(<_ as Into<[i32; 2]>>::into);
let _vertices = edges.flat_map(<[i32; 2]>::from);
```

The lint can be useful, but really only applies if the tuple is either long enough that naming the fields is silly (maybe at least 4 entries long), or if the author intends the fields to be homogenous, which is author intent and can't be determined by the lint. Therefore I think the lint should be marked as pedantic.

Currently, there are also a lot of false positives with the lint:
* https://github.com/rust-lang/rust-clippy/issues/11082
* https://github.com/rust-lang/rust-clippy/issues/11085
* https://github.com/rust-lang/rust-clippy/issues/11100 (https://github.com/rust-lang/rust-clippy/pull/11105)
* https://github.com/rust-lang/rust-clippy/issues/11124
* https://github.com/rust-lang/rust-clippy/issues/11144

Should fix those issues before enabling it for everyone.

---

changelog: Move [`tuple_array_conversions`] to `nursery` (Now allow-by-default)

[#11146](https://github.com/rust-lang/rust-clippy/pull/11146)

### DIFF
--- a/clippy_lints/src/tuple_array_conversions.rs
+++ b/clippy_lints/src/tuple_array_conversions.rs
@@ -16,8 +16,8 @@ declare_clippy_lint! {
     /// Checks for tuple<=>array conversions that are not done with `.into()`.
     ///
     /// ### Why is this bad?
-    /// It's unnecessary complexity. `.into()` works for tuples<=>arrays at or below 12 elements and
-    /// conveys the intent a lot better, while also leaving less room for hard to spot bugs!
+    /// It may be unnecessary complexity. `.into()` works for converting tuples
+    /// <=> arrays of up to 12 elements and may convey intent more clearly.
     ///
     /// ### Example
     /// ```rust,ignore
@@ -31,7 +31,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.72.0"]
     pub TUPLE_ARRAY_CONVERSIONS,
-    complexity,
+    pedantic,
     "checks for tuple<=>array conversions that are not done with `.into()`"
 }
 impl_lint_pass!(TupleArrayConversions => [TUPLE_ARRAY_CONVERSIONS]);


clippy_lints/src/tuple_array_conversions.rs
```diff
@@ -31,7 +31,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.72.0"]
     pub TUPLE_ARRAY_CONVERSIONS,
-    complexity,
+    pedantic,
```
Centri3, line 34: This should be `nursery` instead.

I'll try to fix most of these issues before the next sync (in 2 weeks).

MingweiSamuel: Ok

Centri3 (approved):

Centri3:
Bit sad about this but it makes sense. Once it's been rewritten hopefully we can bump it up to `pedantic`, likely not `complexity` as the lost asymmetry is a big issue. Which is unfortunate since this lint was kinda meant as a learning tool.

Centri3:
> For example, which of the following is the most readable:

Will say though that IMO it's the third, it conveys the intent best and even avoids an unnecessary closure which is pretty nice

xFrednet (approved):

xFrednet:
LGTM, thank you for the PR :+1: I believe this move makes sense for the time being.

The version says 1.72.0, we should be able to back port this move to beta.

r? @xFrednet 

@bors r=Centri3,xFrednet

PR #11161, opened 2023-07-14
[significant_drop_tightening] Fix #11160

### LINKED ISSUE #11160: [significant_drop_tightening] Functions that return literals should not trigger the lint
Literals are cheap expressions that won't interfere with the release of a lock.

```rust
pub fn foo() -> bool {
    let mutex = Mutex::new(1i32);
    let _ = lock.abs();
    true
}
```

cc https://github.com/rust-lang/rust-clippy/issues/11125#issuecomment-1635705282

c410-f3r:
Fix #11160

changelog: [`significant_drop_tightening`]: Ignore literals in function returns


dswij (approved):
Makes sense. Thanks for this!

PR #11166, opened 2023-07-16
Refactor some of  `dereference.rs` to util functions

Jarcho:
I've seen a few lints that need to be able to tell if changing the type of an expression would be a vaild suggestion. This extracts part of how that's done from `explicit_auto_deref`.

changelog: None

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -3,21 +3,23 @@ use clippy_utils::mir::{enclosing_mir, expr_local, local_assignments, used_exact
 use clippy_utils::msrvs::{self, Msrv};
 use clippy_utils::source::{snippet_with_applicability, snippet_with_context};
 use clippy_utils::sugg::has_enclosing_paren;
-use clippy_utils::ty::{adt_and_variant_of_res, expr_sig, is_copy, peel_mid_ty_refs, ty_sig};
+use clippy_utils::ty::{is_copy, peel_mid_ty_refs};
 use clippy_utils::{
-    fn_def_id, get_parent_expr, get_parent_expr_for_hir, is_lint_allowed, path_to_local, walk_to_expr_usage,
+    expr_use_ctxt, get_parent_expr, get_parent_node, is_lint_allowed, path_to_local, DefinedTy, ExprUseNode,
 };
 
+use hir::def::DefKind;
+use hir::MatchSource;
 use rustc_ast::util::parser::{PREC_POSTFIX, PREC_PREFIX};
 use rustc_data_structures::fx::FxIndexMap;
 use rustc_data_structures::graph::iterate::{CycleDetector, TriColorDepthFirstSearch};
 use rustc_errors::Applicability;
+use rustc_hir::def::Res;
 use rustc_hir::def_id::{DefId, LocalDefId};
 use rustc_hir::intravisit::{walk_ty, Visitor};
 use rustc_hir::{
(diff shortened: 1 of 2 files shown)


clippy_utils/src/lib.rs
```diff
@@ -2502,6 +2505,263 @@ pub fn walk_to_expr_usage<'tcx, T>(
...
+    Field(&'tcx ExprField<'tcx>),
+    /// An argument to a function.
+    FnArg(&'tcx Expr<'tcx>, usize),
+    /// An argument to a method.
+    MethodArg(HirId, Option<&'tcx GenericArgs<'tcx>>, usize),
+    /// The callee of a function call.
+    Callee,
+    /// Access of a field.
+    FieldAccess(Ident),
+}
```
smoelius, line 2548:
```suggestion
}

```

clippy_lints/src/dereference.rs
```diff
@@ -649,420 +788,164 @@ fn deref_method_same_type<'tcx>(result_ty: Ty<'tcx>, arg_ty: Ty<'tcx>) -> bool {
...
-        matches!(
-            self,
-            Self::MethodReceiver | Self::FieldAccess { of_union: false, .. } | Self::Callee
-        )
+#[derive(Clone, Copy)]
+enum TyCoercionStability {
+    Deref,
+    Reborrow,
+    None,
+}
```
smoelius, line 818:
```suggestion
}

```
Is juxtaposing the the type and its impl intentional?

Centri3: `blank_lines_lower_bound` is definitely relevant here. Is it worth the conflicts though? :eyes:

Jarcho: Formatting the type and it's impl without a blank line in between is used in other parts of clippy. It's not a consistent thing though.

smoelius:
> I've seen a few lints that need to be able to tell if changing the type of an expression would be a vaild suggestion. This extracts part of how that's done from `explicit_auto_deref`.

Can I ask how/where you see the extracted code being applied?

clippy_lints/src/dereference.rs
smoelius: Are the [multiple impls](https://github.com/rust-lang/rust-clippy/blob/e9f65e3315c068505bda195784d89a58b14e8dc3/clippy_utils/src/lib.rs#L2549-L2561) similarly intentional?

Jarcho: That part isn't.

smoelius:
Thanks!

clippy_utils/src/lib.rs
```diff
@@ -2502,6 +2505,261 @@ pub fn walk_to_expr_usage<'tcx, T>(
...
+    pub fn is_return(&self) -> bool {
+        matches!(self, Self::Return(_))
+    }
+
+    /// Checks if the value is used as a method call receiver.
+    pub fn is_recv(&self) -> bool {
+        matches!(self, Self::MethodArg(_, _, 0))
+    }
+
+    /// Get's the needed type as it's defined without any type inference.
```
Centri3, line 2560:
```suggestion
    /// Gets the needed type as it's defined without any type inference.
```

clippy_utils/src/lib.rs
```diff
@@ -2502,6 +2505,261 @@ pub fn walk_to_expr_usage<'tcx, T>(
     None
 }
 
+/// A type definition as it would be viewd from within a function.
```
Centri3, line 2508:
```suggestion
/// A type definition as it would be viewed from within a function.
```

PR #11169, opened 2023-07-16
don't hide lifetimes for `LateContext`

y21:
Running `cargo dev new_lint --type methods` creates the lint file with hidden lifetimes for the `LateContext` parameter (i.e. `&LateContext`, when it should be `&LateContext<'_>`). This is already warned on with `#![warn(rust_2018_idioms)]`, so clippy should not use hidden lifetimes

changelog: none


xFrednet (approved):

xFrednet:
LGTM, thank you for the PR :)

@bors r+

xFrednet:
Just stealing the review, I saw the PR and it looked interesting and then the review was done :D

r? @xFrednet

Issue #11162, opened 2023-07-14
Uplift `undocumented_unsafe_blocks` to `pedantic`


tgross35:
### Description

`missing_safety_doc` is warn by default, I think it makes sense to have its counterpart `undocumented_unsafe_blocks` at least be in pedantic. The other doc related lints like `missing_panic_docs` are in pedantic, so it seems to be a suitible place.

### Additional Labels

@rustbot label +A-category

tgross35:
I think that if this is uplifted, `accept-comment-above-statement` and `accept-comment-above-attributes` should both default to true, to help keep sanity.

Centri3:
`missing_safety_doc` and `missing_panic_docs` are a bit of an unfair comparison, safety docs are for the user of your crate and let them know what safety invariants they must uphold, whilst `SAFETY:` comments help you explain why it's sound to contributors; sometimes, it's also pretty unnecessary sometimes (stuff like `// SAFETY: the caller must uphold the invariants described in the safety docs` or the like, there's tons of those around libstd that don't really provide much clarity)

I'm definitely for making the 2 config options true by default though :+1:

tgross35:
> `missing_safety_doc` and `missing_panic_docs` are a bit of an unfair comparison, safety docs are for the user of your crate and let them know what safety invariants they must uphold, whilst `SAFETY:` comments help you explain why it's sound to contributors; sometimes, it's also pretty unnecessary (stuff like `// SAFETY: the caller must uphold the invariants described in the safety docs` or the like, there's tons of those around libstd that don't really provide much clarity).
> 
> Stuff like the winapi as well would likely be _very_ unnecessary as a good portion of it can be utilized _somewhat_ care-free (despite being unsafe). You'd have to document each and every call, which would likely just be noise.

I do get that the comparison isn't exactly apples to apples - but Clippy of course has a ton of lints that are dedicated to easier development and maintenance. Not that that alone is the strongest justification, but even a simple comment like `// SAFETY: FFI call with no preconditions` goes a long way toward easily answering the question of why something that says `unsafe` is, well, safe.

> I'm definitely for making the 2 config options true by default though 👍

Great! I can submit a PR for this later

Centri3:
> but Clippy of course has a ton of lints that are dedicated to easier development and maintenance

[...]

PR #11170, opened 2023-07-17
Change defaults of `accept-comment-above-statement` and `accept-comment-above-attributes`

tgross35:
This patch sets the two configuration options for `undocumented_unsafe_blocks` to `true` by default: these are `accept-comment-above-statement` and `accept-comment-above-attributes`. Having these values `false` by default prevents what many users would consider clean code, e.g. placing the `// SAFETY:` comment above a single-line functino call, rather than directly next to the argument.

This was originally discussed in https://github.com/rust-lang/rust-clippy/issues/11162

changelog: [`undocumented_unsafe_blocks`]: set
`accept-comment-above-statement` and `accept-comment-above-attributes` to `true` by default.


tgross35:
Conflicts should be resolved

Since you were involved in the original conversation:

r? @Centri3

Centri3 (approved):

Centri3:
LGTM, though we should probably use revisions for the default/custom configuration options instead of duplicating the file with minor comment changes, but that's not a big deal

Ideally however, these wouldn't even be configuration options as I highly doubt there are any cases (other than bugs) where somebody would actually want to have these set to `false`. We could probably deprecate them and say that this behavior will be forced in the future, but that can be done in a future PR.

@bors r+

tgross35:
> LGTM, though we should probably use revisions for the default/custom configuration options instead of duplicating the file with minor comment changes, but that's not a big deal

If this is something that we can do now, is there an example? I have to fix something here anyway it seems so might as well change that too if possible

Centri3:
You usually put the following headers at the top of the file:
```rust
//@revisions: a b
//@[a] rustc-env:CLIPPY_CONF_DIR=tests/ui-toml/mylint/a
//@[b] rustc-env:CLIPPY_CONF_DIR=tests/ui-toml/mylint/b
```
With a `clippy.toml` file existing in both subdirectories `a` and `b`. Doubt you can conditionally add a comment though so just have something like `SAFETY: will be allowed if <configoption> is true`

Centri3:
Hey @tgross35, any progress on the revisions? If not, this just needs a `cargo collect-metadata` run and a rebase and this can be merged.

tgross35:
@Centri3 sorry about that, it fell off my radar. I think I got the dual config working right as you suggested, and did the rebase.

I'm not sure why github shows the four deleted/renamed files as conflicts when my tree is clean, I am trying to figure this out but do you have any ideas?

tgross35:
I guess I was just rebasing against my own fork instead of upstream 🤦 I think this should be good to go

Centri3 (approved):

Centri3:
Thanks!

@bors r+

PR #11183, opened 2023-07-18
Remove `#![allow(unused)]` and `--crate-name` from `cargo dev new_lint` generated tests

Alexendoo:
changelog: none

Also removes some unused flags from `ui-cargo` tests because the entrypoint is now the `Cargo.toml`, not the `.rs` files

### DIFF
--- a/clippy_dev/src/new_lint.rs
+++ b/clippy_dev/src/new_lint.rs
@@ -96,8 +96,7 @@ fn create_test(lint: &LintData<'_>) -> io::Result<()> {
 
         path.push("src");
         fs::create_dir(&path)?;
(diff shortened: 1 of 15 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thanks! 

@bors r+

Alexendoo:
Realised I didn't mention why `#![allow(unused)]` was removed so for posterity: we pass `-A unused` to tests in `compile-test.rs` but the old `compiletest-rs` crate didn't apply that when running `.fixed` files, `ui_test` does so it's no longer needed

Issue #11179, opened 2023-07-18
`needless_pass_by_ref_mut` false positive


iamsauravsharma:
### Summary

I got
```
error: this argument is a mutable reference, but not used mutably
  --> src/migrator/any.rs:22:17
   |
22 |     connection: &mut <Any as sqlx::Database>::Connection,
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&<Any as sqlx::Database>::Connection`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
note: the lint level is defined here
  --> src/lib.rs:3:9
   |
3  | #![deny(clippy::all)]
   |         ^^^^^^^^^^^
   = note: `#[deny(clippy::needless_pass_by_ref_mut)]` implied by `#[deny(clippy::all)]`
```
even when connection is used as mutable https://github.com/iamsauravsharma/sqlx_migrator/blob/master/src/migrator/any.rs#L36
 GitHub actions error https://github.com/iamsauravsharma/sqlx_migrator/actions/runs/5583309753/jobs/10203493083

### Lint Name

needless_pass_by_ref_mut

### Reproducer

1. clone repo at https://github.com/iamsauravsharma/sqlx_migrator
2. run cargo clippy --features any,postgres you will see clippy error
```
error: this argument is a mutable reference, but not used mutably
  --> src/migrator/any.rs:22:17
   |
22 |     connection: &mut <Any as sqlx::Database>::Connection,
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&<Any as sqlx::Database>::Connection`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
```
Expects no error since connection is used mutably and using as suggestion doesn't compile code

### Version

```text
rustc 1.73.0-nightly (da6b55cc5 2023-07-17)
binary: rustc
commit-hash: da6b55cc5eaf76ed6acb7dc2f7d611e32af7c9a7
commit-date: 2023-07-17
host: aarch64-apple-darwin
release: 1.73.0-nightly
LLVM version: 16.0.5
```

Centri3:
cc @GuillaumeGomez

y21:
Minimal reproducible example:
```rs
[...]

PR #11184, opened 2023-07-18
Fix async functions handling for `needless_pass_by_ref_mut` lint

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust-clippy/issues/11179.

The problem with async is that "internals" are actually inside a closure from the `ExprUseVisitor` point of view, meaning we need to actually run the check on the closures' body as well.

changelog: none

r? @llogiq


GuillaumeGomez:
Fixed the failing lints on my code changes. ^^'

llogiq:
Thank you!

@bors r+

GuillaumeGomez:
It failed because of the commit format. I added `changelog: none` to the first comment of the PR. Hopefully it should fix the issue?

PR #11198, opened 2023-07-20
[`slow_vector_initialization`]: catch `Vec::new()` followed by `.resize(len, 0)`

### LINKED ISSUE #10938: `slow-vector-initialization` should catch `Vec::new` followed by `Vec::resize(len, 0)`.
### Summary

There's a lint, `slow-vector-initialization`, which catches if you try to do something like this:

```rust
let mut foo = Vec::with_capacity(1024);
foo.resize(1024, 0);
```

It suggests you use the `vec!` macro instead, which is faster. However, if you replace the call to `Vec::with_capacity` with `Vec::new`, you get an even worse-performing version of the same code, and clippy is no longer able to suggest the better option.

### Lint Name

slow-vector-initialization

### Reproducer

I tried [this code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c1c4375cee754edca3496bb93291d2d8):
```rust
pub fn main() {
    let mut bar = Vec::new();
    bar.resize(1024, 0);
}
```

Raise an instance of `slow-vector-initialization`, something similar to this:
```
warning: slow zero-filling initialization
 --> src/main.rs:3:5
  |
2 |     let mut bar = Vec::new();
  |                   ------------------ help: consider replace allocation with: `vec![0; 1024]`
3 |     bar.resize(1024, 0);
  |     ^^^^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#slow_vector_initialization
  = note: `#[warn(clippy::slow_vector_initialization)]` on by default
```

Instead, this happened:

No lint raised

### Version

```text
binary: rustc
commit-date: 2023-05-31
host: aarch64-apple-darwin
release: 1.70.0
LLVM version: 16.0.2

(this also happens on the version running on playground right now)
```

y21:
Closes #10938 

changelog: [`slow_vector_initialization`]: catch `Vec::new()` followed by `.resize(len, 0)`

### DIFF
--- a/clippy_lints/src/slow_vector_initialization.rs
+++ b/clippy_lints/src/slow_vector_initialization.rs
@@ -2,7 +2,8 @@ use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::sugg::Sugg;
 use clippy_utils::ty::is_type_diagnostic_item;
 use clippy_utils::{
-    get_enclosing_block, is_integer_literal, is_path_diagnostic_item, path_to_local, path_to_local_id, SpanlessEq,
+    get_enclosing_block, is_expr_path_def_path, is_integer_literal, is_path_diagnostic_item, path_to_local,
+    path_to_local_id, paths, SpanlessEq,
 };
 use if_chain::if_chain;
 use rustc_errors::Applicability;
@@ -60,7 +61,11 @@ struct VecAllocation<'tcx> {
 
     /// Reference to the expression used as argument on `with_capacity` call. This is used
     /// to only match slow zero-filling idioms of the same length than vector initialization.
-    len_expr: &'tcx Expr<'tcx>,
+    ///
+    /// Initially set to `None` if initialized with `Vec::new()`, but will always be `Some(_)` by
+    /// the time the visitor has looked through the enclosing block and found a slow
+    /// initialization, so it is safe to unwrap later at lint time.
+    size_expr: Option<&'tcx Expr<'tcx>>,
 }
 
 /// Type of slow initialization
@@ -77,18 +82,14 @@ impl<'tcx> LateLintPass<'tcx> for SlowVectorInit {
         // Matches initialization on reassignments. For example: `vec = Vec::with_capacity(100)`
         if_chain! {
             if let ExprKind::Assign(left, right, _) = expr.kind;
-
-            // Extract variable
             if let Some(local_id) = path_to_local(left);
-
-            // Extract len argument
-            if let Some(len_arg) = Self::is_vec_with_capacity(cx, right);
+            if let Some(size_expr) = Self::as_vec_initializer(cx, right);
 
             then {
                 let vi = VecAllocation {
                     local_id,
                     allocation_expr: right,
-                    len_expr: len_arg,
+                    size_expr,
                 };
 
                 Self::search_initialization(cx, vi, expr.hir_id);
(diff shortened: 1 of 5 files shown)


Centri3:
As requested by Manishearth on zulip,

r? @Centri3

clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -118,6 +120,25 @@ impl<'tcx> LateLintPass<'tcx> for SlowVectorInit {
...
 
 impl SlowVectorInit {
+    /// Given an expression, returns:
+    /// - `Some(Some(size))` if it is a function call to `Vec::with_capacity(size)`
+    /// - `Some(None)` if it is a call to `Vec::new()`
+    /// - `None` if it is neither
+    #[allow(
+        clippy::option_option,
+        reason = "outer option is immediately removed at call site and the inner option is kept around, \
+            so extracting into a dedicated enum seems unnecessarily complicated"
```
Centri3, line 130: Yeah this seems fine, it'd definitely be better if it was an optional two-variant enum though. But not too big of a deal

clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -118,6 +120,25 @@ impl<'tcx> LateLintPass<'tcx> for SlowVectorInit {
 }
 
 impl SlowVectorInit {
+    /// Given an expression, returns:
+    /// - `Some(Some(size))` if it is a function call to `Vec::with_capacity(size)`
+    /// - `Some(None)` if it is a call to `Vec::new()`
+    /// - `None` if it is neither
+    #[allow(
```
Centri3, line 127: nit:
```suggestion
    #[expect(
```

clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -134,6 +155,14 @@ impl SlowVectorInit {
         }
     }
 
+    /// Checks if the given expression is `Vec::new()`
+    fn is_vec_new(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
```
Centri3, line 159: nit: This is only used in `as_vec_initializer` afaict, if it's refactored into a two-variant enum then this should probably be inlined as naming it wouldn't clarify much then

clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -169,12 +198,18 @@ impl SlowVectorInit {
...
-        let len_expr = Sugg::hir(cx, vec_alloc.len_expr, "len");
+        let len_expr = Sugg::hir(
+            cx,
+            vec_alloc
+                .size_expr
+                .expect("length expression must be set by this point"),
+            "len",
+        );
 
         span_lint_and_then(cx, SLOW_VECTOR_INITIALIZATION, slow_fill.span, msg, |diag| {
```
Centri3, line 209: Not part of your PR, but if this could also suggest to remove the following call that'd be nice. Not sure if that'd have adverse side effects though

(should probably also be a separate PR ^^)

Centri3 (approved):
LGTM, thanks! Just a couple nits and I think this is good to go

clippy_lints/src/slow_vector_initialization.rs
y21: it would end up as an enum that is exactly the same as the `Option<_>` enum in std, just with different variant names, but I can do that if it makes it clearer

clippy_lints/src/slow_vector_initialization.rs
y21: True, though I think we should then also do the same for the other already-existing method that checks for `Vec::with_capacity`, as it would be inconsistent/weird if only `is_vec_new` was inlined imo

y21: err, now actually getting to implement it, that `is_vec_with_capacity` function has slightly more stuff going on so it might be better to leave it

Centri3: Hmm.. It doesn't seem like what it does is necessary, that could probably be changed to matching on the path directly (rather than the method name plus the type).

Centri3:
Thanks! @bors r+

PR #11200, opened 2023-07-20
[`unused_async`]: don't lint if paths reference async fn without immediate call

### LINKED ISSUE #9359: unused_async emitted when trait bounds where function is used require it to be async
### Summary

When using actix-web, Clippy warned me that the `async` on one of my handler functions was unused as there was no await. It turned out that the trait bounds on the `.to()` method of `actix-web::route::Route` required that the handler function implemented `Handler<Args>` which requires an async function.

### Lint Name

unused_async

### Reproducer

I tried this code:

```rust
use actix_web::{web, App, HttpServer};
use std::sync::Mutex;

struct AppStateWithCounter {
    counter: Mutex<i32>, // <- Mutex is necessary to mutate safely across threads
}

async fn index(data: web::Data<AppStateWithCounter>) -> String {
    if let Ok(mut counter) = data.counter.lock() { // <- get counter's MutexGuard
        *counter += 1; // <- access counter inside MutexGuard
        format!("Request number: {counter}") // <- response with count
    } else {
        "Unable to access counter".to_string()
    }
}

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    // Note: web::Data created _outside_ HttpServer::new closure
    let counter = web::Data::new(AppStateWithCounter { 
        counter: Mutex::new(0),
    });

    HttpServer::new(move || {
        // move counter into the closure
        App::new()
            .app_data(counter.clone()) // <- register the created data
            .route("/", web::get().to(index))
    })
    .bind(("127.0.0.1", 8080))?
    .run()
    .await
}
```

I saw this happen:

```
warning: unused `async` for function with no await statements
  --> src\main.rs:8:1
   |
8  | / async fn index(data: web::Data<AppStateWithCounter>) -> String {
9  | |     if let Ok(mut counter) = data.counter.lock() { // <- get counter's MutexGuard
10 | |         *counter += 1; // <- access counter inside MutexGuard
11 | |         format!("Request number: {counter}") // <- response with count
...  |
14 | |     }
15 | | }
   | |_^
   |
   = note: `-W clippy::unused-async` implied by `-W clippy::pedantic`
   = help: consider removing the `async` from this function
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_async
```

### Version

```text
rustc 1.63.0 (4b91a6ea7 2022-08-08)
binary: rustc
commit-hash: 4b91a6ea7258a947e59c6522cd5898e7c0a6a88f
commit-date: 2022-08-08
host: x86_64-pc-windows-gnu
release: 1.63.0
LLVM version: 14.0.5
```

### Additional Labels

@rustbot label + `I-suggestion-causes-error`

### LINKED ISSUE #9695: unused `async` for function with no await statements when function is used in contexts requiring async
### Summary

Functions with the `async` keyword get the following error, even when the functions must have async to be used in contexts that are async.

```
unused `async` for function with no await statements
```

### Lint Name

unused_async

### Reproducer

I tried this code:

```toml
[package]
name = "pro"
version = "0.1.0"
edition = "2021"

[dependencies]
axum = { version = "0.5" }
tokio = { version = "1.0", features = ["full"] }
```

```rust
use axum::{routing::get, Router};
use std::net::SocketAddr;

#[tokio::main]
async fn main() {
    let router = Router::new().route("/", get(home));
    let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
    axum::Server::bind(&addr)
        .serve(router.into_make_service())
        .await
        .unwrap();
}

async fn home() -> &'static str {
    "Hello, World!"
}
```

No warnings.

Instead, this happened:

```
warning: unused `async` for function with no await statements
  --> src/main.rs:15:1
   |
15 | / async fn home() -> &'static str {
16 | |     "Hello, World!"
17 | | }
   | |_^
   |
   = note: `-W clippy::unused-async` implied by `-W clippy::pedantic`
   = help: consider removing the `async` from this function
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unused_async
```

### Version

```text
rustc 1.64.0 (a55dd71d5 2022-09-19)
binary: rustc
commit-hash: a55dd71d5fb0ec5a6a3a9e8c27b2127ba491ce52
commit-date: 2022-09-19
host: x86_64-apple-darwin
release: 1.64.0
LLVM version: 14.0.6
```

y21:
Fixes #9695
Fixes #9359

Clippy shouldn't lint unused `async` if there are paths referencing them if that path isn't the receiver of a function call, because that means that the function might be passed to some other function:
```rs
async fn f() {} // No await statements, so unused at this point

fn requires_fn_future<F: Future<Output = ()>>(_: fn() -> F) {}
requires_fn_future(f); // `f`'s asyncness is actually not unused.
```
(This isn't limited to just passing the function as a parameter to another function, it could also first be stored in a variable and later passed to another function as an argument)

This requires delaying the linting until post-crate and collecting path references to local async functions along the way.

changelog: [`unused_async`]: don't lint if paths reference async fn that require asyncness

### DIFF
--- a/clippy_lints/src/unused_async.rs
+++ b/clippy_lints/src/unused_async.rs
@@ -1,6 +1,5 @@
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::diagnostics::span_lint_hir_and_then;
 use clippy_utils::is_def_id_trait_method;
-use rustc_data_structures::fx::FxHashMap;
 use rustc_hir::def::DefKind;
 use rustc_hir::intravisit::{walk_body, walk_expr, walk_fn, FnKind, Visitor};
 use rustc_hir::{Body, Expr, ExprKind, FnDecl, Node, YieldSource};
@@ -47,11 +46,12 @@ pub struct UnusedAsync {
     async_fns_as_value: LocalDefIdSet,
     /// Functions with unused `async`, linted post-crate after we've found all uses of local async
     /// functions
-    unused_async_fns: FxHashMap<LocalDefId, UnusedAsyncFn>,
+    unused_async_fns: Vec<UnusedAsyncFn>,
 }
 
 #[derive(Copy, Clone)]
 struct UnusedAsyncFn {
+    def_id: LocalDefId,
     fn_span: Span,
     await_in_async_block: Option<Span>,
 }
@@ -122,13 +122,11 @@ impl<'tcx> LateLintPass<'tcx> for UnusedAsync {
                 // Don't lint just yet, but store the necessary information for later.
                 // The actual linting happens in `check_crate_post`, once we've found all
                 // uses of local async functions that do require asyncness to pass typeck
-                self.unused_async_fns.insert(
+                self.unused_async_fns.push(UnusedAsyncFn {
+                    await_in_async_block: visitor.await_in_async_block,
+                    fn_span: span,
                     def_id,
-                    UnusedAsyncFn {
-                        await_in_async_block: visitor.await_in_async_block,
-                        fn_span: span,
-                    },
-                );
+                });
             }
         }
     }
@@ -164,12 +162,13 @@ impl<'tcx> LateLintPass<'tcx> for UnusedAsync {
         let iter = self
             .unused_async_fns
             .iter()
-            .filter_map(|(did, item)| (!self.async_fns_as_value.contains(did)).then_some(item));
+            .filter(|UnusedAsyncFn { def_id, .. }| (!self.async_fns_as_value.contains(def_id)));
 
         for fun in iter {
-            span_lint_and_then(
+            span_lint_hir_and_then(
                 cx,
                 UNUSED_ASYNC,
+                cx.tcx.local_def_id_to_hir_id(fun.def_id),
                 fun.fn_span,
                 "unused `async` for function with no await statements",
                 |diag| {
--- a/clippy_lints/src/unused_async.rs
+++ b/clippy_lints/src/unused_async.rs
@@ -1,11 +1,13 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::is_def_id_trait_method;
+use rustc_data_structures::fx::FxHashMap;
+use rustc_hir::def::DefKind;
 use rustc_hir::intravisit::{walk_body, walk_expr, walk_fn, FnKind, Visitor};
-use rustc_hir::{Body, Expr, ExprKind, FnDecl, YieldSource};
+use rustc_hir::{Body, Expr, ExprKind, FnDecl, Node, YieldSource};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::nested_filter;
-use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::def_id::LocalDefId;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::def_id::{LocalDefId, LocalDefIdSet};
 use rustc_span::Span;
 
 declare_clippy_lint! {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/unused_async.rs
```diff
@@ -38,7 +40,23 @@ declare_clippy_lint! {
...
 
-declare_lint_pass!(UnusedAsync => [UNUSED_ASYNC]);
+#[derive(Default)]
+pub struct UnusedAsync {
+    /// Keeps track of async functions used as values (i.e. path expressions to async functions that
+    /// are not immediately called)
+    async_fns_as_value: LocalDefIdSet,
+    /// Functions with unused `async`, linted post-crate after we've found all uses of local async
+    /// functions
+    unused_async_fns: FxHashMap<LocalDefId, UnusedAsyncFn>,
```
Jarcho, line 50: There's no reason for this to be a `HashMap` instead of a `Vec`. Using a `Vec` will also give a stable lint order.

y21: > There's no reason for this to be a `HashMap` instead of a `Vec`. Using a `Vec` will also give a stable lint order.

Oh right, I used a `HashMap` because I was originally thinking that I could remove defs from `unused_async_fns` as we find paths, but realized that it's probably not really worth it.
But what do you mean about a stable lint order? Do you mean that we wouldn't need to do it post-crate?

clippy_lints/src/unused_async.rs
```diff
@@ -101,24 +119,71 @@ impl<'tcx> LateLintPass<'tcx> for UnusedAsync {
...
+    // After collecting all unused `async` and problematic paths to such functions,
+    // lint those unused ones that didn't have any path expressions to them.
+    fn check_crate_post(&mut self, cx: &LateContext<'tcx>) {
+        let iter = self
+            .unused_async_fns
+            .iter()
+            .filter_map(|(did, item)| (!self.async_fns_as_value.contains(did)).then_some(item));
+
+        for fun in iter {
+            span_lint_and_then(
```
Centri3, line 170: This should be `span_lint_hir_and_then`, with the `HirId` obtained from `local_def_id_to_hir_id`. Otherwise this won't take into account lint attributes on the function's signature

clippy_lints/src/unused_async.rs
Centri3: The order of items iterated on for a hashmap is essentially random, I believe it's stable for `FxHashMap` between different runs but if a function's added the order will completely change (which makes the diff pretty annoying)

clippy_lints/src/unused_async.rs
```diff
@@ -101,24 +119,71 @@ impl<'tcx> LateLintPass<'tcx> for UnusedAsync {
...
+                    kind: ExprKind::Call(Expr { span, .. }, _) | ExprKind::MethodCall(_, Expr { span, .. }, ..),
+                    ..
+                }) if *span == expected_receiver
+            )
+        }
+
+        // Find paths to local async functions that aren't immediately called.
+        // E.g. `async fn f() {}; let x = f;`
+        // Depending on how `x` is used, f's asyncness might be required despite not having any `await`
+        // statements, so don't lint at all if there are any such paths.
```
Centri3, line 150: I believe this can use #11166 once that's merged, that way some cases are still linted

clippy_lints/src/unused_async.rs
Jarcho: Stable in this case meaning stable between different builds of clippy and adding an item doesn't rearrange other items are linted in. Changes to the internals of the `HashMap` should not cause clippy's tests to fail.

In the future if you do need to use lint from a `HashMap` you can sort by the `HirId` emitting the lint.

y21: That makes sense, thanks for clarifying

clippy_lints/src/unused_async.rs
Jarcho: This is unlikely to be the case. The signatures of the two functions are incompatible.

Jarcho:
Thank you. @bors r+

y21:
oops, I must have deleted the changelog word from the pr description again somehow. It's fixed now

PR #11204, opened 2023-07-21
new lint: [`should_panic_without_expect`]

### LINKED ISSUE #10956: Lint on should_panic tests without reasons
### What it does

For a `#[test]` that has `#[should_panic]`, this lint would fire unless given a reason, e.g. `#[should_panic = "No such file or directory"]`. 

### Advantage

This prevents a should-panic test from accepting another reason when it fires, and there are many reasons such a test could panic.

### Drawbacks

It could fire on tests that are genuinely expecting many possible panic reasons and don't care which one fires. These are probably very few in practice, but it could justify making it a "pedantic" lint.

### Example

```rust
    #[should_panic]
    fn test_can_panic() {
        panic!("yup")
    }
```

Could be written as:

```rust
    #[should_panic = "yup"]
    fn test_can_panic() {
        panic!("yup")
    }

```

y21:
Closes #10956

changelog: new lint: [`should_panic_without_expect`]

### DIFF
--- a/clippy_lints/src/attrs.rs
+++ b/clippy_lints/src/attrs.rs
@@ -6,7 +6,11 @@ use clippy_utils::macros::{is_panic, macro_backtrace};
 use clippy_utils::msrvs::{self, Msrv};
 use clippy_utils::source::{first_line_of_span, is_present_in_source, snippet_opt, without_block_comments};
 use if_chain::if_chain;
-use rustc_ast::{AttrKind, AttrStyle, Attribute, LitKind, MetaItemKind, MetaItemLit, NestedMetaItem};
+use rustc_ast::token::{Token, TokenKind};
+use rustc_ast::tokenstream::TokenTree;
+use rustc_ast::{
+    AttrArgs, AttrArgsEq, AttrKind, AttrStyle, Attribute, LitKind, MetaItemKind, MetaItemLit, NestedMetaItem,
+};
 use rustc_errors::Applicability;
 use rustc_hir::{
     Block, Expr, ExprKind, ImplItem, ImplItemKind, Item, ItemKind, StmtKind, TraitFn, TraitItem, TraitItemKind,
@@ -341,6 +345,41 @@ declare_clippy_lint! {
 
 declare_clippy_lint! {
     /// ### What it does
+    /// Checks for `#[should_panic]` attributes without specifying the expected panic message.
+    ///
+    /// ### Why is this bad?
+    /// The expected panic message should be specified to ensure that the test is actually
+    /// panicking with the expected message, and not another unrelated panic.
+    ///
+    /// ### Example
+    /// ```rust
+    /// fn random() -> i32 { 0 }
+    ///
+    /// #[should_panic]
+    /// #[test]
+    /// fn my_test() {
+    ///     let _ = 1 / random();
+    /// }
+    /// ```
+    ///
+    /// Use instead:
+    /// ```rust
+    /// fn random() -> i32 { 0 }
+    ///
+    /// #[should_panic = "attempt to divide by zero"]
+    /// #[test]
+    /// fn my_test() {
+    ///     let _ = 1 / random();
+    /// }
+    /// ```
+    #[clippy::version = "1.73.0"]
+    pub SHOULD_PANIC_WITHOUT_EXPECT,
+    pedantic,
+    "ensures that all `should_panic` attributes specify its expected panic message"
+}
+
+declare_clippy_lint! {
+    /// ### What it does
     /// Checks for `any` and `all` combinators in `cfg` with only one condition.
     ///
     /// ### Why is this bad?
@@ -395,6 +434,7 @@ declare_lint_pass!(Attributes => [
     DEPRECATED_SEMVER,
     USELESS_ATTRIBUTE,
     BLANKET_CLIPPY_RESTRICTION_LINTS,
+    SHOULD_PANIC_WITHOUT_EXPECT,
 ]);
 
 impl<'tcx> LateLintPass<'tcx> for Attributes {
@@ -441,6 +481,9 @@ impl<'tcx> LateLintPass<'tcx> for Attributes {
                     }
                 }
             }
+        }
+        if attr.has_name(sym::should_panic) {
+            check_should_panic_reason(cx, attr);
         }
     }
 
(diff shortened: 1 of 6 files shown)


clippy_lints/src/attrs.rs
```diff
@@ -716,6 +756,7 @@ impl EarlyLintPass for EarlyAttributes {
         check_mismatched_target_os(cx, attr);
         check_minimal_cfg_condition(cx, attr);
         check_misused_cfg(cx, attr);
+        check_should_panic_reason(cx, attr);
```
Centri3, line 759: The lint is in the `Attributes` pass but the check is in `EarlyAttributes`.

This should be in `Attributes` instead anyway as [early lints in this file are pre-expansion lints](https://github.com/rust-lang/rust-clippy/blob/ee8a429792eb280d20c5e4f4b1549d1dd3b30335/clippy_lints/src/lib.rs#L370), which have historically had countless problems and are quasi-deprecated in rustc. The implementation should work the same as a late pass lint

clippy_lints/src/attrs.rs
```diff
@@ -767,6 +808,39 @@ fn check_empty_line_after_outer_attr(cx: &EarlyContext<'_>, item: &rustc_ast::It
...
+        }
+
+        span_lint_and_sugg(
+            cx,
+            SHOULD_PANIC_WITHOUT_REASON,
+            attr.span,
+            "#[should_panic] attribute without a reason",
+            "consider specifying the expected panic",
+            r#"#[should_panic(expected="<panic message>")]"#.into(),
+            Applicability::Unspecified
```
Centri3, line 839: nit: This should probably be `HasPlaceholders`

clippy_lints/src/attrs.rs
```diff
@@ -767,6 +808,39 @@ fn check_empty_line_after_outer_attr(cx: &EarlyContext<'_>, item: &rustc_ast::It
...
+            return;
+        }
+
+        span_lint_and_sugg(
+            cx,
+            SHOULD_PANIC_WITHOUT_REASON,
+            attr.span,
+            "#[should_panic] attribute without a reason",
+            "consider specifying the expected panic",
+            r#"#[should_panic(expected="<panic message>")]"#.into(),
```
Centri3, line 838: nit: There should probably be a space on either side of the `=`

clippy_lints/src/attrs.rs
y21: The reason I made it part of the early lint pre-expansion pass is because that makes finding the `#[should_panic]` attribute a lot easier. `#[test]` seems to expand to a `const` item with the `should_panic` attribute part of a field of the struct being initialized in the const item somewhere.
`#[should_panic]` doesn't appear in `Attribute`'s `check_attribute` fn, so I guess we'd have to look for these kinds of const items, which sounds a lot more painful than what it's currently doing. Or am I missing something?

Centri3: Oh yeah that's a good point actually, also more importantly `#[cfg(test)]` and just `#[test]` will cause it to not appear in the HIR *unless* you're running `cargo test`

So this'll unfortunately need to be a pre-expansion lint. To make sure that `#[cfg(test)}` and all work tests should be added for those

Alexendoo:
I'd say call this `should_panic_without_expect`, I wasn't aware you could supply an expected panic message and the title made me think it would be like a comment

clippy_lints/src/attrs.rs
Alexendoo: If it can be done outside of a pre expansion pass it should be, the main pre expansion bug is it not picking up lint levels correctly - https://github.com/rust-lang/rust-clippy/issues/6610, for a lint that you might want to globally allow this is particularly bad

`cargo clippy --tests` (or `--all-targets` etc) can be used to run clippy on tests

y21: Right, okay, that's pretty bad. This makes it quite a bit more complicated, so I suppose we'll have to look for the expanded `#[test]` const item and extract the `should_panic` field that's somewhere in the `test::TestDescAndFn` struct initializer.
I'm not really sure how I'd then get the span of the `#[should_panic]` attribute though, as I don't think this is anywhere in the HIR after `#[test]` expansion

Centri3: All I can think of is having a pre expansion pass to get the span of the attribute, then a late pass for the actual linting. Perhaps you can match the functions together through the `testfn` field.

It's unfortunately gonna require a hacky solution no matter what

y21: > Another thing that's worth a shot is using `get_attrs` on the `DefId` of the function called in `testfn`. _maybe_ that'd have `should_panic`

Oh that's smart, I didn't think of that. I looked at the output of `-Zunpretty=hir-tree` and I can see a `#[should_panic]` attribute with the right span on there. That should work and is probably the least hacky way to make this work

y21:
> I'd say call this `should_panic_without_expect`, I wasn't aware you could supply an expected panic message and the title made me think it would be like a comment

Good point, I renamed the lint and also rewrote it as a late lint now

clippy_lints/src/attrs.rs
```diff
@@ -518,6 +558,42 @@ impl<'tcx> LateLintPass<'tcx> for Attributes {
...
+                    // test::StaticTestFn(|| ..)
+                    && let ExprKind::Call(_, [closure]) = testfn_field.expr.kind
+                    && let ExprKind::Closure(closure) = closure.kind
+                    // test::assert_result(f())
+                    //                     ^ get this `DefId` and its should_panic attr
+                    && let ExprKind::Call(_, [fn_call_closure]) = cx.tcx.hir().body(closure.body).value.kind
+                    && let ExprKind::Call(callee, _) = fn_call_closure.kind
+                    && let ExprKind::Path(QPath::Resolved(_, path)) = callee.kind
+                    && let Some(test_function_did) = path.res.opt_def_id()
+                    && let Some(attribute) = cx.tcx.get_attr(test_function_did, sym::should_panic)
```
Alexendoo, line 592: If the should_panic attribute remains on the function is there a reason we aren't checking that directly? Rather than finding the function via its test description

y21: Huh, okay I think I made a mistake when I rewrote this from pre expansion to late lint. It seems like I was wrong: `check_attribute` *does* get called for `#[should_panic]`. It must have been that I forgot to run with `--test`, in which case the function is completely cfg'd out and does not show up at all (so it makes sense that the attribute didn't get visited).
In that case we can make this a whole lot simpler.

Centri3:
Maybe rebase and run `cargo bless`? rustfix is ran automatically now so it may be creating a `fixed` file

Alexendoo:
You can add `//@no-rustfix` to prevent that

clippy_utils/src/paths.rs
```diff
@@ -166,3 +166,5 @@ pub const DEBUG_STRUCT: [&str; 4] = ["core", "fmt", "builders", "DebugStruct"];
 pub const ORD_CMP: [&str; 4] = ["core", "cmp", "Ord", "cmp"];
 #[expect(clippy::invalid_paths)] // not sure why it thinks this, it works so
 pub const BOOL_THEN: [&str; 4] = ["core", "bool", "<impl bool>", "then"];
+#[allow(clippy::invalid_paths, reason = "internal lints do not always know about ::test")]
+pub const TEST_DESC_AND_FN: [&str; 3] = ["test", "types", "TestDescAndFn"];
```
Alexendoo, line 170: Thanks! r=me with this removed

Alexendoo (approved):

Alexendoo:
👍

@bors r+

PR #11207, opened 2023-07-21
[`needless_pass_by_ref_mut`]: Do not lint if passed as a fn-like argument

### LINKED ISSUE #11182: `needless_pass_by_ref_mut` false positive because of type change
### Summary

A function may require an `&mut` argument so that it has a particular type.

### Lint Name

needless_pass_by_ref_mut

### Reproducer

I tried this code (note that `foo` needs to have the `&mut` argument to be passed to `bar`):

```rust
fn foo(y: &mut i32) -> i32 {
    12 + *y
}

fn bar(_: impl Fn(&mut i32) -> i32) {}

fn main() {
    bar(foo);
}
```

I saw this happen:

```
warning: this argument is a mutable reference, but not used mutably
 --> src/main.rs:1:11
  |
1 | fn foo(y: &mut i32) -> i32 {
  |           ^^^^^^^^ help: consider changing to: `&i32`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
  = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default
```

I expected to see this happen: no warning.

### Version

```text
rustc 1.73.0-nightly (da6b55cc5 2023-07-17)
binary: rustc
commit-hash: da6b55cc5eaf76ed6acb7dc2f7d611e32af7c9a7
commit-date: 2023-07-17
release: 1.73.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

_No response_

### LINKED ISSUE #11199: Suggests removed "unused" mut when not actually able to be removed
### Summary

Clippy will suggest that `mut` be removed from function references if it is not used, even if the mutability cannot be removed because of function type requirements.

### Lint Name

`clippy::needless_pass_by_ref_mut`

### Reproducer

The codebase this occurs in is rather complicated, so I reduced it to a simple test case which reproduces the same false positive.

```rs
#[derive(Debug)]
struct Context {
    a: i32,
    b: i32,
}

type ActionFn = fn(&mut Context);

#[derive(Debug)]
struct Action {
    name: &'static str,
    call: ActionFn,
}

fn action_foo(ctx: &mut Context) {
    println!("Uses r/w access to parameter.");
    ctx.a += 1;
}

fn action_bar(ctx: &mut Context) {
    println!("Only needs r/o access, but must be &mut to match the function definition.");
    let _ = ctx.a * ctx.b;
}

fn main() {
    let action1 = Action {
        name: "Foo",
        call: action_foo,
    };

    let action2 = Action {
        name: "Bar",
        call: action_bar,
    };

    run_action(&action1);
    run_action(&action2);
}

fn run_action(action: &Action) {
    println!("Running action {}", action.name);
    let mut ctx = Context { a: 100, b: -20 };
    (action.call)(&mut ctx);
}
```

Here we have a fixed object `Action` which expects a specific type for its function pointer. However, while most implementers need for full mutable access, some do not, which then triggers the following Clippy lint:

```rs
warning: this argument is a mutable reference, but not used mutably
  --> src/main.rs:20:20
   |
20 | fn action_bar(ctx: &mut Context) {
   |                    ^^^^^^^^^^^^ help: consider changing to: `&Context`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
   = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default

warning: `test2` (bin "test2") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
```

However, this is incorrect, as the `mut` cannot actually be removed, as this will change the function signature, causing a compiler error. That is, if I change line 20 to `fn action_bar(ctx: &Context)`, then I get:

```rs
    Checking test2 v0.1.0 (/tmp/emmie/test2)
  --> src/main.rs:33:15
   |
33 |         call: action_bar,
   |               ^^^^^^^^^^ types differ in mutability
   |
   = note: expected fn pointer `for<'a> fn(&'a mut Context)`
                 found fn item `for<'a> fn(&'a Context) {action_bar}`
   = note: when the arguments and return types match, functions can be coerced to function pointers

error: could not compile `test2` (bin "test2") due to previous error
```

Because there is a legitimate type incompatibility reason why `mut` cannot be removed here, Clippy should not be recommending its removal.

### Version

```text
rustc 1.73.0-nightly (39f42ad9e 2023-07-19)
binary: rustc
commit-hash: 39f42ad9e8430a8abb06c262346e89593278c515
commit-date: 2023-07-19
release: 1.73.0-nightly
LLVM version: 16.0.5

clippy 0.1.73 (39f42ad 2023-07-19)
```

### Additional Labels

Centri3:
Fixes #11182 and also fixes #11199 (though this is kind of a duplicate)

There's likely a case or two I've missed, so this likely needs a bit more work but it seems to work fine with the tests I've added.

PS, the diff for the test is useless because it iterates over a hashmap before linting. Seems to work fine but we could maybe change this for consistency's sake

changelog: [`needless_pass_by_ref_mut`]: No longer lints if the function is passed as a fn-like argument

### DIFF
--- a/clippy_lints/src/needless_pass_by_ref_mut.rs
+++ b/clippy_lints/src/needless_pass_by_ref_mut.rs
@@ -1,16 +1,16 @@
 use super::needless_pass_by_value::requires_exact_signature;
-use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::diagnostics::span_lint_hir_and_then;
 use clippy_utils::source::snippet;
-use clippy_utils::{is_from_proc_macro, is_self};
-use if_chain::if_chain;
-use rustc_data_structures::fx::FxHashSet;
+use clippy_utils::{get_parent_node, is_from_proc_macro, is_self};
+use rustc_data_structures::fx::{FxHashSet, FxIndexMap};
 use rustc_errors::Applicability;
-use rustc_hir::intravisit::FnKind;
-use rustc_hir::{Body, FnDecl, HirId, HirIdMap, HirIdSet, Impl, ItemKind, Mutability, Node, PatKind};
+use rustc_hir::intravisit::{walk_qpath, FnKind, Visitor};
+use rustc_hir::{Body, ExprKind, FnDecl, HirId, HirIdMap, HirIdSet, Impl, ItemKind, Mutability, Node, PatKind, QPath};
 use rustc_hir_typeck::expr_use_visitor as euv;
 use rustc_infer::infer::TyCtxtInferExt;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::map::associated_body;
+use rustc_middle::hir::nested_filter::OnlyBodies;
 use rustc_middle::mir::FakeReadCause;
 use rustc_middle::ty::{self, Ty, UpvarId, UpvarPath};
 use rustc_session::{declare_tool_lint, impl_lint_pass};
@@ -48,20 +48,24 @@ declare_clippy_lint! {
     "using a `&mut` argument when it's not mutated"
 }
 
-#[derive(Copy, Clone)]
-pub struct NeedlessPassByRefMut {
+#[derive(Clone)]
+pub struct NeedlessPassByRefMut<'tcx> {
     avoid_breaking_exported_api: bool,
+    used_fn_def_ids: FxHashSet<LocalDefId>,
+    fn_def_ids_to_maybe_unused_mut: FxIndexMap<LocalDefId, Vec<rustc_hir::Ty<'tcx>>>,
 }
 
-impl NeedlessPassByRefMut {
+impl NeedlessPassByRefMut<'_> {
     pub fn new(avoid_breaking_exported_api: bool) -> Self {
         Self {
             avoid_breaking_exported_api,
+            used_fn_def_ids: FxHashSet::default(),
+            fn_def_ids_to_maybe_unused_mut: FxIndexMap::default(),
         }
     }
 }
 
-impl_lint_pass!(NeedlessPassByRefMut => [NEEDLESS_PASS_BY_REF_MUT]);
+impl_lint_pass!(NeedlessPassByRefMut<'_> => [NEEDLESS_PASS_BY_REF_MUT]);
 
 fn should_skip<'tcx>(
     cx: &LateContext<'tcx>,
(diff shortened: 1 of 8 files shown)


y21:
I think this fix would still lint incorrectly on my provided example in one of the linked issues: https://github.com/rust-lang/rust-clippy/issues/11199#issuecomment-1644634675
Probably makes sense to have that as a test as well

Centri3:
Yeah and I believe it'll still lint stuff like
```rust
fn then(a: &mut u32) {}

fn els(a: &mut u32) {
    *a = 2;
}

if true {
    then
} else {
    els
}
```
Which isn't good

clippy_lints/src/needless_pass_by_ref_mut.rs
```diff
@@ -46,20 +49,26 @@ declare_clippy_lint! {
...
 
-#[derive(Copy, Clone)]
-pub struct NeedlessPassByRefMut {
+#[derive(Clone)]
+pub struct NeedlessPassByRefMut<'tcx> {
     avoid_breaking_exported_api: bool,
+    used_fn_def_ids: FxHashSet<LocalDefId>,
+    // TODO(Centri3): Change this to a `Vec`. This will cause a performance deficit but it'll make the output a lot
+    // more consistent, as adding or removing a function won't change the order of them
+    fn_def_ids_to_maybe_unused_mut: FxHashMap<LocalDefId, Vec<rustc_hir::Ty<'tcx>>>,
```
GuillaumeGomez, line 58: No need to use a `Vec`, simply replace `FxHashMap` with `FxIndexMap` to keep the order.

Centri3: Neat, had no idea that was a crate

GuillaumeGomez: Compiler internals are full of very useful things. :D

tests/ui/needless_pass_by_ref_mut.rs
```diff
@@ -155,15 +156,48 @@ async fn a8(x: i32, a: &mut i32, y: i32, z: &mut i32) {
...
+    foo4(&mut v);
+    foo5(&mut v);
+    alias_check(&mut v);
+    alias_check2(&mut v);
+    println!("{u}");
+    closure_takes_mut(passed_as_closure);
+    A { s: passed_as_field };
+    used_as_path;
+    let _: fn(&mut u32) = passed_as_local;
+    let _ = if true { ty_unify_1 } else { ty_unify_2 };
```
GuillaumeGomez, line 202: Might be better to instead have:

```suggestion
    let _  = if v[0] == 0 { ty_unify_1 } else { ty_unify_2 };
```

Not sure if const removal pass was already run at this point or not.

Centri3: Ah yeah, it's not present so it has

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, great work!

giraffate (approved):

giraffate:
@bors r+

Thanks!

Centri3:
Looks like dogfood failed because of #10120.
@bors r=giraffate

PR #11208, opened 2023-07-21
Added new lint "path_join_correction" 

ofeeg:
r? xFrednet
changelog: Remade #10663 to point to a fresh branch as per request of @blyxyas 

Finally got a break from my job to get around to this. Thank you guys for waiting.


clippy_lints/src/methods/join_absolute_path.rs
```diff
@@ -0,0 +1,30 @@
...
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::{symbol::sym::Path, Span};
+
+use super::JOIN_ABSOLUTE_PATH;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>, span: Span) {
+    let ty = cx.typeck_results().expr_ty(expr);
+    if_chain!(
```
xFrednet, line 12: Rustfmt sadly doesn't work with `if_chains`, could you correct the indention of the code below?

clippy_lints/src/methods/join_absolute_path.rs
```diff
@@ -0,0 +1,30 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, ty::is_type_diagnostic_item};
+use rustc_ast::ast::LitKind;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::{symbol::sym::Path, Span};
+
+use super::JOIN_ABSOLUTE_PATH;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>, span: Span) {
```
xFrednet, line 10: We should make sure that this code is actually written by the user, you can check that it doesn't come from a macro with [`Span::from_expansion`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_span/span_encoding/struct.Span.html#method.from_expansion):
```rs
!expr.span.from_expansion()
```

clippy_lints/src/methods/join_absolute_path.rs
```diff
@@ -0,0 +1,30 @@
...
+     if is_type_diagnostic_item(cx, ty, Path);
+     let applicability = Applicability::MachineApplicable;
+     if let ExprKind::Lit(spanned) = &join_arg.kind;
+     if let LitKind::Str(symbol, _) = spanned.node;
+     if symbol.as_str().starts_with('/') || symbol.as_str().starts_with('\\');
+     then {
+      span_lint_and_sugg(
+      cx,
+      JOIN_ABSOLUTE_PATH,
+      span.with_hi(expr.span.hi()),
```
xFrednet, line 22: Here you can just use `join_arg.span` :)

clippy_lints/src/methods/join_absolute_path.rs
```diff
@@ -0,0 +1,30 @@
...
+      r#"argument in join called on path contains a starting '/'"#,
```
xFrednet, line 23: I would suggest a slightly different message:

```suggestion
      "argument to `Path::join` starts with a path separator",
```

the separator might be platform specific :)

clippy_lints/src/methods/join_absolute_path.rs
```diff
@@ -0,0 +1,30 @@
...
+      "try removing first '/' or '\\'",
+      "join(\"your/path/here\")".to_owned(),
+      applicability
```
xFrednet, line 26: The applicability determines how the suggestion can be used by other tools, like rustfix. In this case, the applicability should be `Applicability::Unspecified`, as the suggested text cannot be substituted for the original.

I would suggest, just displaying a help message right here. We could draft a suggestion if you're interested :). For now, you can use this emission function, to apply all my suggestions above and add more help messages :).
```rs
span_lint_and_then(
    cx,
    JOIN_ABSOLUTE_PATH,
    join_arg.span,
    "argument to `Path::join` starts with a path separator",
    |diag| {
        diag.note("joining a path starting with separator will replace the path instead");
        diag.help(r#"if this is unintentional, try removing the starting separator"#);
        diag.help(r#"if this is intentional, try creating a new Path instead"#);
    },
)
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3261,6 +3262,41 @@ declare_clippy_lint! {
     "calling `drain` in order to `clear` a container"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for initial `'/ or \\'` in an argument to `.join()` on a `Path`.
+    ///
+    /// ### Why is this bad?
+    /// `.join()` comments starting with a separator  (`/` or `\\`) can replace the entire path.
+    /// If this is intentional, prefer creating a new `Path` instead.
```
xFrednet, line 3271:
```suggestion
    /// `.join()` arguments starting with a separator  (`/` or `\\`) can replace the entire path.
    /// If this is intentional, prefer creating `Path::new()` instead.
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3261,6 +3262,41 @@ declare_clippy_lint! {
     "calling `drain` in order to `clear` a container"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for initial `'/ or \\'` in an argument to `.join()` on a `Path`.
```
xFrednet, line 3267:
```suggestion
    /// Checks for calls to `Path::join` that start with a path separator, like `\\` or `/`.
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3261,6 +3262,41 @@ declare_clippy_lint! {
...
+    /// `.join()` comments starting with a separator  (`/` or `\\`) can replace the entire path.
+    /// If this is intentional, prefer creating a new `Path` instead.
+    ///
+    /// See [`Path::join()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join)
+    ///
+    /// ### Example
+    /// ```rust
+    /// let path = std::path::Path::new("/bin");
+    /// let res = path.join("/sh");
+    /// assert_eq!(res, std::path::PathBuf::from("/sh"));
```
xFrednet, line 3279: In these examples, it would be nicer to use `# use std::path::Path` and then remove the leading `std::path`.

```suggestion
    /// # use std::path::{Path, PathBuf}
    /// let path = Path::new("/bin");
    /// let joined_path = path.join("/sh");
    /// assert_eq!(joined_path, PathBuf::from("/sh"));
```

Same in the example below

tests/ui/join_absolute_path.rs
```diff
@@ -0,0 +1,26 @@
+//@run-rustfix
```
xFrednet, line 1: rustfix only works with valid suggestions, for now, the best action is to rust remove it :)
```suggestion
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3832,6 +3869,7 @@ impl Methods {
                     if let Some(("collect", _, _, span, _)) = method_call(recv) {
                         unnecessary_join::check(cx, expr, recv, join_arg, span);
                     }
+                    else {join_absolute_path::check(cx, expr,  join_arg, span);}
```
xFrednet, line 3872: The `expr` value is the whole `<receiver>.join(<arg>)` expression. For the lint we want to check the type of the receiver. You can simply update this to:

```suggestion
                    else {
                        join_absolute_path::check(cx, recv, join_arg, span)
                    }
```

clippy_lints/src/methods/join_absolute_path.rs
```diff
@@ -0,0 +1,30 @@
...
+use rustc_ast::ast::LitKind;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::{symbol::sym::Path, Span};
+
+use super::JOIN_ABSOLUTE_PATH;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>, span: Span) {
+    let ty = cx.typeck_results().expr_ty(expr);
```
xFrednet, line 11: With the updated receiver, this gives you the following types (using `eprintln!("ty:#?")`)

```
&std::path::Path
&std::path::Path
&[&str]
&std::path::Path
```

The reference at the start prevents the `if is_type_diagnostic_item(cx, ty, Path);` check from passing. Instead, we should first remove the outer reference, this can be done with [`Ty::peel_refs`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.peel_refs)

```suggestion
    let ty = cx.typeck_results().expr_ty(expr).peel_refs();
```

xFrednet (commented):
Hey @ofeeg, welcome back! This version is already a good start. I left several comments, once they are addressed, we should have a working lint. Please don't hesitate to reach out if you have any questions!

@rustbot author

blyxyas:
"Pretty good first PR!"

*Proceeds to **completely and utterly destroy** your PR*

xFrednet:
> Is it solved or do I save your fork just in case. Also, @xFrednet I think I'll let you have this one, as I have a lot of reviews currently. Hope it's not a problem :)

I'm happy to do the full review :D

ofeeg:
> There are merge commits (commits with multiple parents) in your changes. We have a [no merge policy](https://rustc-dev-guide.rust-lang.org/git.html#no-merge-policy) so these commits will need to be removed for this pull request to be merged.
> 
> You can start a rebase with the following commands:
> 
> ```
> $ # rebase
> $ git rebase -i master
> $ # delete any merge commits in the editor that appears
> $ git push --force-with-lease
> ```
> 
> The following commits are merge commits:
> 
>     * [1f26d8c](https://github.com/rust-lang/rust-clippy/commit/1f26d8cbd387f0bd459dfdaff44154909b8850df)
> 
>     * [1d77675](https://github.com/rust-lang/rust-clippy/commit/1d7767588cbaa967c8f40128866470cb7ac78256)

I don't know what this is about. It may have to do with me accidentally committing to master instead of the new branch by complete accident. I gotta go to my day job now, though.

xFrednet:
Clippy has a policy that branches should be updated using rebases instead of merges. I think in this case, it would be best to squash all commits into one. That would remove the merge commits, without any hassle. 

Also, a small note, your most recent commits use a different name and don't link to your GH. This is fine for Clippy, but I wanted to mention it in case this was unintenional :)

ofeeg:
> Clippy has a policy that branches should be updated using rebases instead of merges. I think in this case, it would be best to squash all commits into one. That would remove the merge commits, without any hassle.
> 
> Also, a small note, your most recent commits use a different name and don't link to your GH. This is fine for Clippy, but I wanted to mention it in case this was unintenional :)

Thanks for your guidance! Hopefully this next commit is agreeable! Sorry for taking so long; I've been going through a lot of personal changes, and they're mostly related to getting out of intense physical labor for at least a little while. lol

clippy_lints/src/methods/join_absolute_paths.rs~
```diff
@@ -0,0 +1,30 @@
+use clippy_utils::{diagnostics::span_lint_and_sugg, ty::is_type_diagnostic_item};
```
xFrednet, line 1: Can you remove this `rs~` file?

xFrednet (commented):
This is a good step, the lint now detects the test cases correctly. I had some other suggests in the previous review, which were marked as *Outdated* since the file was renamed with the lint rename. Could you take a look at the resolved comments? They're related to the diagnostic and documentation style. Once they're addressed, we should be able to move forward :)

> Sorry for taking so long; I've been going through a lot of personal changes, and they're mostly related to getting out of intense physical labor for at least a little while. lol

No problem at all. I hope this is a good change for you!

clippy_lints/src/methods/mod.rs
```diff
@@ -2974,8 +2975,8 @@ declare_clippy_lint! {
     /// # let mut state = DefaultHasher::new();
     /// # let my_enum = Foo::Empty;
     /// match my_enum {
-    /// 	Empty => ().hash(&mut state),
-    /// 	WithValue(x) => x.hash(&mut state),
+    ///         Empty => ().hash(&mut state),
+    ///         WithValue(x) => x.hash(&mut state),
```
Centri3, line 2979: nit: Can you change this to 4 spaces instead of 8? Seems unintentional

Same with the other change a few dozen lines down

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,27 @@
...
+use rustc_ast::ast::LitKind;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::symbol::sym::Path;
+
+use super::JOIN_ABSOLUTE_PATHS;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>) {
+    let ty = cx.typeck_results().expr_ty(expr).peel_refs();
+    if !expr.span.from_expansion() && is_type_diagnostic_item(cx, ty, Path)
```
Centri3, line 12: This `from_expansion` check is unnecessary as all lints in `methods/` already do this check. It should be removed

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,27 @@
...
+        join_arg.span,
+        "argument to `Path::join` starts with a path separator",
+        |diag| {
+          diag.note("joining a path starting with separator will replace the path instead");
+          diag.help(r#"if this is unintentional, try removing the starting separator"#);
+          diag.help(r#"if this is intentional, try creating a new Path instead"#);
+        },
+      );
+    };
+}
```
Centri3, line 27:
```suggestion
    if is_type_diagnostic_item(cx, ty, Path)
        && let ExprKind::Lit(spanned) = &join_arg.kind
        && let LitKind::Str(symbol, _) = spanned.node
        && (symbol.as_str().starts_with('/') || symbol.as_str().starts_with('\\'))
    {
        span_lint_and_then(
            cx,
            JOIN_ABSOLUTE_PATHS,
            join_arg.span,
            "argument to `Path::join` starts with a path separator",
            |diag| {
                diag
                      .note("joining a path starting with separator will replace the path instead")
                      .help(r#"if this is unintentional, try removing the starting separator"#)
                      .help(r#"if this is intentional, try creating a new Path instead"#);
            },
        );
    }
```
This formatting is pretty weird, in Clippy we tend to put each `&&` on a new line and the `&&` at the start of the line. iirc there's no consensus in the style team on how this should be formatted, so this is just what we do for now.

Usually, when working with let chains, you want to remove the entire let chain (leaving only the first condition) to see what rustfmt would probably do. This is because, as xFrednet said, rustfmt doesn't format let chains currently, including the entire `then` block.

(This suggestion also fixes the 2 nits ^^)

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,27 @@
...
+    if !expr.span.from_expansion() && is_type_diagnostic_item(cx, ty, Path)
+    && let ExprKind::Lit(spanned) = &join_arg.kind && let LitKind::Str(symbol, _) = spanned.node
+    && (symbol.as_str().starts_with('/') || symbol.as_str().starts_with('\\')) {
+      span_lint_and_then(
+        cx,
+        JOIN_ABSOLUTE_PATHS,
+        join_arg.span,
+        "argument to `Path::join` starts with a path separator",
+        |diag| {
+          diag.note("joining a path starting with separator will replace the path instead");
```
Centri3, line 21: nit: These 3 can be chained, like `diag.note(...).help(...).help(...)`

clippy_lints/src/methods/mod.rs
```diff
@@ -3261,6 +3262,43 @@ declare_clippy_lint! {
...
+    /// assert_eq!(joined_path, PathBuf::from("/bin/sh"));
+    ///
+    /// // If this was intentional, create a new path instead
+    /// let new = Path::new("/sh");
+    /// assert_eq!(new, PathBuf::from("/sh"));
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub JOIN_ABSOLUTE_PATHS,
+    pedantic,
+  "arg to .join called on a Path contains separator at the start"
```
Centri3, line 3299:
```suggestion
  "argument to `Path::join()` containing leading separator"
```

tests/ui/join_absolute_paths.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    // should not be linted
+    let path: &[&str] = &["/bin"];
+    path.join("/sh");
+    println!("{:?}", path);
+
+    //should not be linted
+    let path = Path::new("/bin");
+    path.join("sh");
+    println!("{}", path.display());
```
Centri3, line 24: These `println!`s should probably be removed. They're a bit misleading as `join` takes `&self`, not `&mut self`.

clippy_lints/src/methods/mod.rs
```diff
@@ -3261,6 +3262,43 @@ declare_clippy_lint! {
...
+    /// let joined_path = path.join("sh");
+    /// assert_eq!(joined_path, PathBuf::from("/bin/sh"));
+    ///
+    /// // If this was intentional, create a new path instead
+    /// let new = Path::new("/sh");
+    /// assert_eq!(new, PathBuf::from("/sh"));
+    /// ```
+    #[clippy::version = "1.70.0"]
+    pub JOIN_ABSOLUTE_PATHS,
+    pedantic,
```
Centri3, line 3298: Is there a reason this is `pedantic`? I think there's an argument to be had for this to be `suspicious` or `correctness` instead, so it's enabled-by-default.

Centri3 (commented):
Just a couple nits, the actual code looks pretty good

clippy_lints/src/methods/mod.rs
ofeeg: I actually didn't know there was correctness as an option. I went with pedantic as it doesn't seem like a breaking mistake, but correctness makes way more sense here. I'm choosing that.

tests/ui/join_absolute_paths.rs
ofeeg: That makes sense. Not everyone would be using absolute paths when calling join to get this behavior.

xFrednet (commented):
This version looks good to me! Thank you for resolving all our comments. Also, thank you for your review @Centri3. :heart: 

@ofeeg The commit history needs to be cleaned up a bit, before we can merge this. There should be no merge commits and the messages should all be descriptive. I would suggest squashing all commits of this PR into a single commit. This will remove the merge commits easily. Removing them with a rebase is possible, but can be a bit tricky.

And then we can merge it :)

ofeeg:
> This version looks good to me! Thank you for resolving all our comments. Also, thank you for your review @Centri3. heart
> 
> @ofeeg The commit history needs to be cleaned up a bit, before we can merge this. There should be no merge commits and the messages should all be descriptive. I would suggest squashing all commits of this PR into a single commit. This will remove the merge commits easily. Removing them with a rebase is possible, but can be a bit tricky.
> 
> And then we can merge it :)

I'm still  learning how to rebase effectively. I think I did it correctly but even with no changes from the passing commit that I backed up it seems to be failing the test here on github. I'm pretty confused as to how to handle this due to all the squashing I had to do.

xFrednet:
I've tried to look how this can be easily fixed, but I couln't find a good solution. Rebases can be really evil, especially when you only get started using them :see_no_evil:. In this case, I would suggest the following:
1. Create a backup branch
2. Reset this branch to have no changes at all
3. Copy the files you touched onto this branch
4. Add a commit for them

This is far from ideal, but the best way I can find right now. I have a backup of an older version of this PR under https://github.com/xFrednet/rust-clippy/tree/backup/pr/path_join_correct that branch might be a good start, if you want to see which files you touched.

Sorry that the rebase has messed up the history so much. The next time, I'll try to write more detailed instruction :see_no_evil:

ofeeg:
I think I fixed everything, but I did a force push which seems to be an awful idea. I ran cargo test on my own machine and it worked so should I just create a new PR again, or are you able to reopen the PR upon checking the recent commit? @xFrednet

xFrednet:
Hey @ofeeg, something is still of, but I'm not quite sure what. I created https://github.com/rust-lang/rust-clippy/pull/11453 which should include all theses changes :)

PR #11214, opened 2023-07-23
check that the types are equal in `SpanlessEq::eq_expr`

### LINKED ISSUE #11213: Wrong `if_same_then_else` lint error when having two different integer
### Summary

if we call a trait method on two different integer type in a if else condition and specify the types like `0_u8`, `0u16` and... this lint will give us a error about both if and else are the same, when they're not...

### Lint Name

if_same_then_else

### Reproducer

I tried this code:

```rust
let test = if true { 0_u16.is_power_of_two() } else { 0_u8.is_power_of_two() };
```

I saw this happen:

```
error: this `if` has identical blocks
  --> src.rs:44:28
   |
44 |         let test = if true { 0_u16.is_power_of_two() } else { 0_u8.is_power_of_two() };
   |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: same as this
  --> src.rs:44:61
   |
44 |         let test = if true { 0_u16.is_power_of_two() } else { 0_u8.is_power_of_two() };
   |                                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                           
```

I didn't expected to see any lint error here because I'm calling this trait method on two different type.

### Version

```text
rustc 1.71.0 (8ede3aae2 2023-07-12)
binary: rustc
commit-hash: 8ede3aae28fe6e4d52b38157d7bfe0d3bceef225
commit-date: 2023-07-12
host: x86_64-pc-windows-msvc
release: 1.71.0
LLVM version: 16.0.5
```

### Additional Labels

_No response_

y21:
Fixes #11213

changelog: [`if_same_then_else`]: don't lint for integer literals of different types

### DIFF
--- a/clippy_utils/src/hir_utils.rs
+++ b/clippy_utils/src/hir_utils.rs
@@ -252,15 +252,15 @@ impl HirEqInterExpr<'_, '_, '_> {
             return false;
         }
 
-        if let Some((typeck_lhs, typeck_rhs)) = self.inner.maybe_typeck_results {
-            if let (Some(l), Some(r)) = (
+        if let Some((typeck_lhs, typeck_rhs)) = self.inner.maybe_typeck_results
+            && typeck_lhs.expr_ty(left) == typeck_rhs.expr_ty(right)
+            && let (Some(l), Some(r)) = (
                 constant_simple(self.inner.cx, typeck_lhs, left),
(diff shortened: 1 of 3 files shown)


Jarcho:
Thank you. @bors r+

Issue #11181, opened 2023-07-18
ptr_arg should ignore extern functions


david-monroe:
### Summary

There is usually no way for extern code to convert a `&Vec<u8>` to a `&[u8]` as the layout of the latter is unspecified.

>### Why this is bad
>Requiring the argument to be of the specific size makes the function less useful for no benefit;

The benefit is that it can be used from extern code.

### Lint Name

ptr_arg

### Reproducer

I tried this code:

```rust
    #[no_mangle]
    pub extern "C" fn read_buffer(buffer: &Vec<u8>, len: usize) {
        
    }
```

I saw this happen:

```
  |
2 |     pub extern "C" fn read_buffer(buffer: &Vec<u8>, len: usize) {
  |                                           ^^^^^^^^ help: change this to: `&[u8]`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg
  = note: `#[warn(clippy::ptr_arg)]` on by default
```

I expected to see this happen:

No warning.

### Version

```text
Playground
```

PR #11215, opened 2023-07-23
ptr_arg should ignore extern functions

MrNossiom:
Fixes: #11181

changelog: [`ptr_arg`]: ignore extern functions that are not  

I am not sure whether we should ignore other Rust calling conventions like `rust-intrinsic`, `rust-call` or `rust-cold`.


Jarcho:
The call to `check_mut_from_ref` should happen before before the ABI check. `mut_from_ref` shouldn't depend on the function's ABI.

Jarcho:
Thank you. @bors r+

PR #11221, opened 2023-07-24
Remove Gha status emitter in compile-test

Alexendoo:
Disables the github specific output for now since it can be a bit confusing - https://github.com/oli-obk/ui_test/issues/109, in particular the truncation/repetition

r? @flip1995

changelog: none

### DIFF
--- a/tests/compile-test.rs
+++ b/tests/compile-test.rs
@@ -5,7 +5,7 @@
 #![warn(rust_2018_idioms, unused_lifetimes)]
 #![allow(unused_extern_crates)]
 
-use compiletest::{status_emitter, CommandBuilder};
+use compiletest::{status_emitter, CommandBuilder, OutputConflictHandling};
 use ui_test as compiletest;
 use ui_test::Mode as TestMode;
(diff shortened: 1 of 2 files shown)


flip1995:
@bors r+

Thanks!

tests/compile-test.rs
```diff
@@ -296,7 +300,7 @@ fn run_ui_cargo() {
             config.out_dir = PathBuf::from("target/ui_test_cargo/").join(path.parent().unwrap());
             Some(config)
         },
-        (status_emitter::Text, status_emitter::Gha::<true> { name }),
+        status_emitter::Text,
```
RalfJung, line 303: @oli-obk so should we apply the same diff in Miri to reduce the confusing output there?

oli-obk: We can also just fix it ^^

flip1995: Once it is Just Fixed :tm: , we'll also revert this and re-enable it in Clippy :)

PR #11225, opened 2023-07-25
Fix integration tests #2

matthiaskrgr:
fix integration tests.

It turned out that the following tests fail to build at all:

chalk, combine, stdarch and hyper.

This is often a problem of passing `--all-targets --all-features`, in case of combine though, outdated deps were to blame.

I have opened tickets against combine and rustfmt
https://github.com/rust-lang/rustfmt/issues/5859
https://github.com/Marwes/combine/issues/357

should we just remove the other failing repos? :/

changelog: fix integration tests on ci

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -65,6 +65,10 @@ fn integration_test() {
         .expect("unable to run clippy");
 
     let stderr = String::from_utf8_lossy(&output.stderr);
+
+    // debug:
+    eprintln!("{stderr}");
+
     if let Some(backtrace_start) = stderr.find("error: internal compiler error") {
         static BACKTRACE_END_MSG: &str = "end of query stack";
         let backtrace_end = stderr[backtrace_start..]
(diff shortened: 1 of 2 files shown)


matthiaskrgr:
lol

matthiaskrgr:
should be ready now :sparkles:

tests/integration.rs
```diff
@@ -65,6 +65,10 @@ fn integration_test() {
         .expect("unable to run clippy");
 
     let stderr = String::from_utf8_lossy(&output.stderr);
+
+    // debug:
+    eprintln!("{stderr}");
```
flip1995, line 70: Is this actually a leftover for debugging, or do you want to have this in the tests? From the added `--show-output` flag, I assume this is left here intentionally.

I think this could be quite useful, just want to make sure it wasn't forgotten about.

flip1995:
@bors delegate+

r=me with the comment addressed one way or another.

Thank you for working on this and fixing it! :heart:

tests/integration.rs
matthiaskrgr: No, this makes us print the clippy command stderr, but it would still be hidden by the testing harness by default, which is why we also need the other flag.

I think its nice to be verbose by default here so we can see what the problem is right away,in case there are any.

flip1995: I thought so. Makes sense, let's ship it :shipit:

Issue #11185, opened 2023-07-19
`needless_pass_by_ref_mut` should not trigger for cfg-gated mod/fn


oxalica:
### Summary

In target specific modules, some functions may only mutate arguments for a specific platform. Keeping the same signature to take `&mut` for all of them makes higher level abstractions simpler.

If we expose this platform differences, it will propagate upwards and force every users to `#[cfg]` between `&` and `&mut`. This violates the abstraction goal of the library.

A solution to this issue is to treat every `#[cfg] fn` and `#[cfg] mod { pub fn }` as part of public interface, and suppress warnings that would cause signature changes on them.

### Lint Name

needless_pass_by_ref_mut

### Reproducer

```rust
pub struct Resource(i32);

impl Resource {
    // Real public interface.
    pub fn work(&mut self) {
        imp::work(&mut self.0);
    }
}

// Semi-public interface. Abstractions over platforms.
#[cfg(unix)]
mod imp {
    // Immutable for some platforms.
    pub fn work(data: &mut i32) { // <- needless_pass_by_ref_mut
        println!("{data}");
    }
}

#[cfg(windows)]
mod imp {
    // Mutable for some other platforms.
    pub fn work(data: &mut i32) {
        *data += 1;
        println!("{data}");
    }
}
```

I saw this happen:

```
warning: this argument is a mutable reference, but not used mutably
  --> src/lib.rs:13:23
   |
14 |     pub fn work(data: &mut i32) { // <- needless_pass_by_ref_mut
   |                       ^^^^^^^^ help: consider changing to: `&i32`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
   = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default
```

I expected to see this happen:
<No warnings>

### Version

```text
Rust Playground, nightly.
Clippy 0.1.73 (2023-07-18 903e279)
```

### Additional Labels

@rustbot label +I-false-positive +I-suggestion-causes-error

I think this fits `I-suggestion-causes-error` because the change breaks on other cfg-gated platforms.

Centri3:
cc @GuillaumeGomez

Likely you'll want to traverse up the `DefId` tree using `parent_module`/`parent_module_from_def_id` to see if any have a `#[cfg(...)]` attribute, plus the function in question ofc.

GuillaumeGomez:
[...]

PR #11226, opened 2023-07-25
Needless ref mut cfg

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust-clippy/issues/11185.

cc @Centri3 

changelog: Emit note if function is behind a cfg for `NEEDLESS_PASS_BY_REF_MUT` lint.

### DIFF
--- a/clippy_lints/src/needless_pass_by_ref_mut.rs
+++ b/clippy_lints/src/needless_pass_by_ref_mut.rs
@@ -12,11 +12,11 @@ use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::map::associated_body;
 use rustc_middle::hir::nested_filter::OnlyBodies;
 use rustc_middle::mir::FakeReadCause;
-use rustc_middle::ty::{self, Ty, UpvarId, UpvarPath};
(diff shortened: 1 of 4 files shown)


clippy_lints/src/needless_pass_by_ref_mut.rs
```diff
@@ -93,6 +93,16 @@ fn should_skip<'tcx>(
     is_from_proc_macro(cx, &input)
 }
 
+fn inherits_cfg(tcx: TyCtxt<'_>, def_id: LocalDefId) -> bool {
```
Centri3, line 96: This would be cleaner with `tcx.hir().parent_iter()`, Which is what [`is_in_cfg_test`](https://doc.rust-lang.org/nightly/nightly-rustc/src/clippy_utils/lib.rs.html#2430-2446) does. The lint already gets the `HirId` of the function so that can be turned into a local.

(PS: This should probably be a utils function similar to that one)

GuillaumeGomez: Sounds good to me. Moving to the function to utils.

GuillaumeGomez:
Updated.

llogiq:
Looks good to me. Good job on the test updates, too!

@bors r+

PR #11231, opened 2023-07-25
Add *no merge policy* note via rustbot

xFrednet:
I just found out that `@rustbot` can automatically add a note about our *no-merge commits* policy, if it detects a merge commit: https://forge.rust-lang.org/triagebot/no-merge.html

Funnly enough, I found this while writing documentation for Marker. `@rustbot` is a cool tool :+1: 

---

r? @flip1995 since you'll probably be the person that will see this message the most ^^

changelog: none

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -9,6 +9,9 @@ allow-unauthenticated = [
 # See https://github.com/rust-lang/triagebot/wiki/Shortcuts
 [shortcut]
 
+# Have rustbot inform users about the *No Merge Policy*
+[no-merges]
+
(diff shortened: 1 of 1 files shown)


matthiaskrgr:
Curious if  this will break rustc-> clippy syncs 😅

xFrednet:
Based on the documentation, I'm assuming that it only adds a comment to the PR. The PR should still be mergable as usual afterwards. :)

flip1995:
@bors r+

Looking forward to talking to rustbot during syncs! :upside_down_face:

PR #11239, opened 2023-07-27
Update ui test crate

oli-obk:
DO NOT MERGE before https://github.com/console-rs/indicatif/pull/563 has made it to crates.io. While ui-test works fine when piped to a file (or in CI), locally it becomes very weird on test suites with more tests than lines in the terminal. The 0.12 update of ui-test includes progress bars and spinners showing which tests are running and how far along in the test suite you are. 

This update also removes the `//@run-rustfix` flag, and just runs rustfix on all tests. This means I had to opt out of running rustfix on ~100 tests, but it also allowed me to remove the rustfix coverage check entirely, as it is now effectively builtin.

changelog: update ui-test crate to 0.12 (automatically runs rustfix on all tests)


oli-obk:
If you want to try out the new progress bar feature, use `cargo test -- -- --quiet`.

A notable difference (without `--quiet`) to before this PR is that the test names are now printed immediately, but the `ok` part after the `...` is only printed when the test actually finishes.

Alexendoo:
This fails to run on my laptop, it hits

```
thread '<unnamed>' panicked at 'could not execute "/home/alex/rust/clippy/target/debug/clippy-driver" ...: Too many open files (os error 24)', /home/alex/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ui_test-0.13.0/src/lib.rs:620:35
...
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "SendError(..)"', /home/alex/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ui_test-0.13.0/src/status_emitter.rs:255:54
```

It's 2 cores/2 threads so it's not running much in parallel, it seems to be opening a growing number of pipes according to `lsof` so there may be a leak somewhere. My `ulimit -n` is apparently 1024 which is surprisingly small

oli-obk:
I fixed the pipe leak (i'll bump the minor version of ui test in the Cargo.toml here to pull it in), but miri has the same issue for low thread/process limits.

Part of the problem seems to be, that it's per process tree, not per process. So all the clippy processes running at the same time seem to be sharing a single limit.

Alexendoo:
Nice, runs fine now

I've noticed it's quite a bit slower, measuring it with:

```bash
hyperfine -L commit 7c595b4599d546dc5f36bc4bebbbf8ad454fed62,ui_test \
    --prepare 'git checkout {commit}' \
    --warmup 2 \
    -n '{commit}' \
    'cargo uitest'
```

x86_64-pc-windows-msvc, 6 cores/12 threads

```
Benchmark 1: 7c595b4599d546dc5f36bc4bebbbf8ad454fed62
  Time (mean ± σ):     18.325 s ±  0.296 s    [User: 113.464 s, System: 59.451 s]
  Range (min … max):   17.899 s … 18.837 s    10 runs

Benchmark 2: ui_test
  Time (mean ± σ):     24.262 s ±  0.041 s    [User: 61.322 s, System: 39.990 s]
  Range (min … max):   24.208 s … 24.332 s    10 runs

Summary
  '7c595b4599d546dc5f36bc4bebbbf8ad454fed62' ran
    1.32 ± 0.02 times faster than 'ui_test'
```

aarch64-unknown-linux-gnu, 4 cores/4 threads

```
Benchmark 1: 7c595b4599d546dc5f36bc4bebbbf8ad454fed62
  Time (mean ± σ):     55.910 s ±  0.233 s    [User: 146.823 s, System: 58.316 s]
  Range (min … max):   55.658 s … 56.372 s    10 runs
 
Benchmark 2: ui_test
  Time (mean ± σ):     85.066 s ±  0.091 s    [User: 97.486 s, System: 48.101 s]
  Range (min … max):   84.942 s … 85.243 s    10 runs
 
Summary
  '7c595b4599d546dc5f36bc4bebbbf8ad454fed62' ran
    1.52 ± 0.01 times faster than 'ui_test'
```

oli-obk:
I'll investigate, but I won't get to that in the next few weeks as I'm going on vacation starting this weekend.

oli-obk:
looks like there were two separate regressions:

```
  add2722677412c40cb4ef41ac46ede4792442f4d (commit = add2722677412c40cb4ef41ac46ede4792442f4d) ran
    1.06 ± 0.01 times faster than b401c422f22470822f84111d400579b3689b84cd (commit = b401c422f22470822f84111d400579b3689b84cd)
    1.84 ± 0.01 times faster than 7b194241530c7d8cbd7f677d587e3a86e0340114 (commit = 7b194241530c7d8cbd7f677d587e3a86e0340114)
    2.73 ± 0.02 times faster than 17d39fda8312f667a2151d8c03b2d8cba1ec435e (commit = 17d39fda8312f667a2151d8c03b2d8cba1ec435e)
    2.73 ± 0.02 times faster than ui_test (commit = ui_test)
```

oli-obk:
The performance issue has been found and resolved.

Alexendoo:
Very nice, it's now faster than the current version

```
Windows
  'ui_test' ran
    1.42 ± 0.03 times faster than 'add2722677412c40cb4ef41ac46ede4792442f4d'

Linux aarch64
  'ui_test' ran
    1.40 ± 0.01 times faster than 'add2722677412c40cb4ef41ac46ede4792442f4d'
```

flip1995:
@bors r=Alexendoo,flip1995 p=10

Taking over the r+, to avoid more conflicts

Thanks!

oli-obk:
Ready again. I also checked (manually) and haven't found any other flags being used in clippy CI

Issue #11238, opened 2023-07-27
explicit unit type instead of underscore


lengyijun:
### What it does

When use underscore to ignored some parameter in match block or closure definition,  if the ignored parameter is unit type, it should be replaced with `()` explicitly

### Advantage

- explicit () is still zero cost
- helpful in reading code: whenever meet underscore, we always wonder what type is ignored here and why. explict () can clear up misunderstandings

### Drawbacks

`|()|` has one more charactor than `|_|`
length = 4 vs length = 3

### Example

```rust
fn foo() -> Result<() , ()> { unimplemented!() }

match foo() {
  Ok(_) => {}
  Err(_) => {}
}

foo().map_err(|_| { todo!() })
```

Could be written as:

```rust
fn foo() -> Result<() , ()> { unimplemented!() }

match foo() {
  Ok(()) => {}
  Err(()) => {}
}

foo().map_err(|()| { todo!() })
```

PR #11242, opened 2023-07-27
New lint `ignored_unit_patterns`

samueltardieu:
This idea comes from #11238. I've put the lint in `pedantic` as it might trigger numerous positives (three in Clippy itself).

changelog: [`ignored_unit_patterns`]: new lint

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4890,6 +4890,7 @@ Released 2018-09-13
 [`if_same_then_else`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else
 [`if_then_some_else_none`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_then_some_else_none
 [`ifs_same_cond`]: https://rust-lang.github.io/rust-clippy/master/index.html#ifs_same_cond
+[`ignored_unit_patterns`]: https://rust-lang.github.io/rust-clippy/master/index.html#ignored_unit_patterns
 [`impl_trait_in_params`]: https://rust-lang.github.io/rust-clippy/master/index.html#impl_trait_in_params
 [`implicit_clone`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_clone
 [`implicit_hasher`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_hasher
--- a/clippy_dev/src/main.rs
+++ b/clippy_dev/src/main.rs
@@ -41,7 +41,7 @@ fn main() {
                 matches.get_one::<String>("type").map(String::as_str),
                 matches.get_flag("msrv"),
             ) {
-                Ok(_) => update_lints::update(update_lints::UpdateMode::Change),
+                Ok(()) => update_lints::update(update_lints::UpdateMode::Change),
                 Err(e) => eprintln!("Unable to create lint: {e}"),
             }
         },
--- a/clippy_dev/src/setup/vscode.rs
+++ b/clippy_dev/src/setup/vscode.rs
@@ -47,7 +47,7 @@ fn check_install_precondition(force_override: bool) -> bool {
         }
     } else {
         match fs::create_dir(vs_dir_path) {
-            Ok(_) => {
+            Ok(()) => {
                 println!("info: created `{VSCODE_DIR}` directory for clippy");
             },
             Err(err) => {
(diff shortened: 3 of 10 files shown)


Jarcho:
I think it would be better for the name to mention that it's about pattern matching. Maybe `ignored_unit_patterns`.

clippy_lints/src/ignored_unit.rs
```diff
@@ -0,0 +1,54 @@
...
+use rustc_hir as hir;
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `_` in patterns of type `()`.
+    ///
+    /// ### Why is this bad?
+    /// Matching with `()` explicitely instead of `_` outlines
```
Centri3, line 13:
```suggestion
    /// Matching with `()` explicitly instead of `_` outlines
```

Centri3: This should probably also mention that it prevents the type from changing as well. Usually a good thing, other times not.

clippy_lints/src/ignored_unit.rs
```diff
@@ -0,0 +1,54 @@
...
+    #[clippy::version = "1.73.0"]
+    pub IGNORED_UNIT,
+    pedantic,
+    "suggest replacing `_` by `()` in patterns where appropriate"
+}
+declare_lint_pass!(IgnoredUnit => [IGNORED_UNIT]);
+
+impl<'tcx> LateLintPass<'tcx> for IgnoredUnit {
+    fn check_pat(&mut self, cx: &LateContext<'tcx>, pat: &'tcx hir::Pat<'tcx>) {
+        pat.walk(|pat| {
```
Centri3, line 39: nit: There's no point to walking this, iirc, as `check_pat` will already enter paths recursively. Maybe I'm wrong though? :thinking:

clippy_lints/src/ignored_unit.rs
```diff
@@ -0,0 +1,54 @@
...
+            if matches!(pat.kind, PatKind::Wild) && cx.tcx.typeck(pat.hir_id.owner.def_id).pat_ty(pat).is_unit() {
```
Centri3, line 40: nit: You can use `cx.typeck_results()` here.

clippy_lints/src/ignored_unit.rs
```diff
@@ -0,0 +1,54 @@
...
+                    cx,
+                    IGNORED_UNIT,
+                    pat.span,
+                    "matching over `()` is more explicit",
+                    "use `()` instead of `_`",
+                    String::from("()"),
+                    Applicability::MachineApplicable,
+                );
+            }
+            true
```
Centri3, line 51: nit: If removing `pat.walk` doesn't work, this should use `walk_always` instead.

samueltardieu:
> I think it would be better for the name to mention that it's about pattern matching. Maybe `ignored_unit_patterns`.

Name and PR updated

samueltardieu:
@Centri3 Thanks for your review, I think all points have been addressed.

Centri3 (approved):

Centri3:
LGTM, I'll wait until @giraffate reviews as well though I think this should be good to go. Thanks!

Centri3:
Oh yeah btw, I think in the future we can extend this to other patterns, like so: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=55acf7c246efb174d34dfe0c4e177425, in a general pedantic lint for unnecessary wildcards. Though it's probably debatable on whether this is better, it does prevent the type changing as well which may or may not be a good thing.

samueltardieu:
> I think in the future we can extend this to other patterns

Yeah, the first step could be to extend it to other singleton 
types (such as been done for `()`).

giraffate:
Hmm, I'm wondering if `restriction` would be a good category. It looks like a strict coding style.

I would like hear @Centri3's opinion.

Centri3:
`pedantic` is fine imo for the current lint, the only real strictness I think that may be undesirable in some cases is that the type cannot change, but it's opt in so it should be ok.

giraffate (approved):

giraffate:
@bors r=Centri3,giraffate

Thanks!

cloneable:
This check also catches `_: ()` as function parameter. Is this intended? Based on the error help I'm not sure it is.

```
error: matching over `()` is more explicit
 --> src/lib.rs:8:12
  |
8 | pub fn moo(_: ()) {}
  |            ^ help: use `()` instead of `_`: `()`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#ignored_unit_patterns
  = note: `-D clippy::ignored-unit-patterns` implied by `-D clippy::pedantic`
```

Centri3:
`_: ()` is `pat: ty`, so yes, but I'm not sure if it's desirable

cloneable:
Ok, I filed #11403 where this can be discussed, but feel free to close if all is good.

PR #11252, opened 2023-07-29
[`unwrap_used`]: Do not lint unwrapping on `!` or never-like enums

### LINKED ISSUE #11245: Lint to check for unwrap except on Infallible error
### What it does

Behaves exactly like `unwrap_used` _**except**_ if the error type is `Infallible` it does not issue a warning.

### Advantage

- Still warns the user about unsafe unwraps.
- Does not bother the user where `unwrap` cannot fail due to the error being `Infallible`.

### Drawbacks

None (that I can think of)

### Example

```rust
const fn foo() -> Result<&'static str, Infallible> {
    Ok("foo")
}

fn main() {
    let f = foo().expect("This cannot fail.");
    println!("{f}");
}
```

Could be written as:

```rust
const fn foo() -> Result<&'static str, Infallible> {
    Ok("foo")
}

fn main() {
    let f = foo().unwrap();
    println!("{f}");
}
```

Centri3:
Fixes #11245

changelog: [`unwrap_used`]: Do not lint unwrapping on `!` or never-like enums
changelog: [`expect_used`]: Do not lint unwrapping on `!` or never-like enums


clippy_lints/src/methods/unwrap_expect_used.rs
```diff
@@ -0,0 +1,87 @@
...
+    let method_suffix = if is_err { "_err" } else { "" };
+
+    if allow_unwrap_in_tests && (is_in_test_function(cx.tcx, expr.hir_id) || is_in_cfg_test(cx.tcx, expr.hir_id)) {
+        return;
+    }
+
+    let help = if variant == Variant::Unwrap && is_lint_allowed(cx, EXPECT_USED, expr.hir_id) {
+        format!(
+            "if you don't want to handle the `{none_value}` case gracefully, consider \
+                using `expect{method_suffix}()` to provide a better panic message"
```
xFrednet, line 73: NIT: Since this is a long message, I would actually suggest including a line break. Having everything in a long line is often less readable. Alternatively, you could use a shorter version:

```suggestion
            "consider using `expect{method_suffix}()` to provide a better panic message"
```

I probably prefer having the shorter message instead of a line break, but either option is fine :)

clippy_lints/src/methods/unwrap_expect_used.rs
```diff
@@ -0,0 +1,87 @@
...
+        ("an `Option`", "None", "")
+    } else if is_type_diagnostic_item(cx, ty, sym::Result)
+        && let ty::Adt(_, substs) = ty.kind()
+        && let Some(t_or_e_ty) = substs[usize::from(!is_err)].as_type()
+    {
+        // Issue #11245: Do not lint `!` or never-like enums
+        if t_or_e_ty.is_never()
+            || (t_or_e_ty.is_enum() && t_or_e_ty.ty_adt_def().is_some_and(|def| def.variants().is_empty()))
+        {
+            return;
```
xFrednet, line 56: NIT: For readability and consistency, it might be good to extract this into `clippy_utils::ty`

xFrednet (approved):
That's a nice refactoring! I have two small NITs, everything else looks good to me! You can `r=me` once the nits have been addressed.

clippy_lints/src/methods/unwrap_expect_used.rs
Centri3: I definitely prefer the shorter variant too

Centri3:
Somehow managed to put the utils function in just the right spot to cause a conflict. :D

@bors r=xFrednet

PR #11253, opened 2023-07-29
[`no_mangle_with_rust_abi`]: Check for statics with a non `#[repr(Rust)]` type

### LINKED ISSUE #11219: Lint against `#[no_mangle]` for non-`repr(C)` `pub static`s
### What it does

The lint should lint against a `#[no_mangle]` `pub static` which has a type which is not `repr(C)` (or that has no explicit `repr(Rust)` set).

This would be similar to `no_mangle_with_rust_abi` from https://github.com/rust-lang/rust-clippy/issues/10347, but for `pub static`s.

### Advantage

Avoids potential subtle mistakes and UB in programs that mix Rust and other languages like C: [one cannot predict the layout of `repr(Rust)` types](https://doc.rust-lang.org/reference/type-layout.html#the-default-representation) and thus an exported `static` marked with `#[no_mangle]` has a high chance of being a mistake.

### Drawbacks

_No response_

### Example

```rust
pub struct S(u8, u16);

#[no_mangle]
pub static X: S = S(0xFF, 0xFFFF);
```

Should be written as:

```rust
#[repr(C)]
pub struct S(u8, u16);

#[no_mangle]
pub static X: S = S(0xFF, 0xFFFF);
```

Otherwise, a C program with manually written bindings (e.g. projects not using `cbindgen`, which in this case would generate an incomplete type) trying to access `X` will break sooner or later, potentially silently (using `-Zrandomize-layout` would increase the chances of the issue being spotted, though), e.g.

```c
#include <assert.h>
#include <stdint.h>

struct S {
    uint8_t a;
    uint16_t b;
};

extern const struct S X;

int main(void) {
    assert(X.a == 0xFF);
    assert(X.b == 0xFFFF);
    return 0;
}
```

Centri3:
Fixes #11219

changelog: [`no_mangle_with_rust_abi`]: Check for statics with an unstable ABI

### DIFF
--- a/clippy_lints/src/no_mangle_with_rust_abi.rs
+++ b/clippy_lints/src/no_mangle_with_rust_abi.rs
@@ -1,32 +1,41 @@
 use clippy_utils::diagnostics::span_lint_and_then;
-use clippy_utils::source::snippet_with_applicability;
+use clippy_utils::source::snippet_opt;
 use rustc_errors::Applicability;
-use rustc_hir::{Item, ItemKind};
+use rustc_hir::{FnSig, Item, ItemKind, Ty};
+use rustc_hir_analysis::hir_ty_to_ty;
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty;
 use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::{BytePos, Pos};
+use rustc_span::{sym, BytePos, Pos};
 use rustc_target::spec::abi::Abi;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for Rust ABI functions with the `#[no_mangle]` attribute.
+    /// Checks for Rust ABI functions or statics with the `#[no_mangle]` attribute.
     ///
     /// ### Why is this bad?
-    /// The Rust ABI is not stable, but in many simple cases matches
-    /// enough with the C ABI that it is possible to forget to add
-    /// `extern "C"` to a function called from C. Changes to the
-    /// Rust ABI can break this at any point.
+    /// The Rust ABI is not stable, but in many simple cases matches enough with the C ABI
+    /// that it is possible to forget to add `extern "C"` to a function called from C.
+    /// Changes to the Rust ABI can break this at any point.
     ///
     /// ### Example
     /// ```rust
-    ///  #[no_mangle]
-    ///  fn example(arg_one: u32, arg_two: usize) {}
-    /// ```
+    /// #[no_mangle]
+    /// fn example(arg_one: u32, arg_two: usize) {}
     ///
+    /// pub struct UsingMeInCIsUB(u32, u32);
+    /// #[no_mangle]
+    /// pub static ZERO: UsingMeInCIsUB = UsingMeInCIsUB(0, 0);
+    /// ```
     /// Use instead:
     /// ```rust
-    ///  #[no_mangle]
-    ///  extern "C" fn example(arg_one: u32, arg_two: usize) {}
+    /// #[no_mangle]
+    /// extern "C" fn example(arg_one: u32, arg_two: usize) {}
+    ///
+    /// #[repr(C)]
+    /// pub struct UsingMeInCIsFine(u32, u32);
+    /// #[no_mangle]
+    /// pub static ZERO: UsingMeInCIsFine = UsingMeInCIsFine(0, 0);
     /// ```
     #[clippy::version = "1.69.0"]
     pub NO_MANGLE_WITH_RUST_ABI,
@@ -37,33 +46,75 @@ declare_lint_pass!(NoMangleWithRustAbi => [NO_MANGLE_WITH_RUST_ABI]);
 
 impl<'tcx> LateLintPass<'tcx> for NoMangleWithRustAbi {
     fn check_item(&mut self, cx: &LateContext<'tcx>, item: &'tcx Item<'tcx>) {
-        if let ItemKind::Fn(fn_sig, _, _) = &item.kind {
-            let attrs = cx.tcx.hir().attrs(item.hir_id());
-            let mut app = Applicability::MaybeIncorrect;
-            let snippet = snippet_with_applicability(cx, fn_sig.span, "..", &mut app);
-            for attr in attrs {
-                if let Some(ident) = attr.ident()
-                    && ident.name == rustc_span::sym::no_mangle
-                    && fn_sig.header.abi == Abi::Rust
-                    && let Some((fn_attrs, _)) = snippet.split_once("fn")
-                    && !fn_attrs.contains("extern")
-                {
-                    let sugg_span = fn_sig.span
-                        .with_lo(fn_sig.span.lo() + BytePos::from_usize(fn_attrs.len()))
-                        .shrink_to_lo();
-
-                    span_lint_and_then(
-                        cx,
-                        NO_MANGLE_WITH_RUST_ABI,
-                        fn_sig.span,
-                        "`#[no_mangle]` set on a function with the default (`Rust`) ABI",
-                        |diag| {
-                            diag.span_suggestion(sugg_span, "set an ABI", "extern \"C\" ", app)
-                                .span_suggestion(sugg_span, "or explicitly set the default", "extern \"Rust\" ", app);
-                        },
-                    );
-                }
-            }
+        if cx.tcx.get_attr(item.owner_id, sym::no_mangle).is_none() {
+            return;
         }
+
+        match item.kind {
+            ItemKind::Fn(fn_sig, _, _) => check_fn(cx, fn_sig),
+            ItemKind::Static(ty, _, _) => check_static(cx, item, ty),
+            _ => {},
+        }
+    }
+}
+
+/// Check for functions that are implicitly using the Rust ABI.
+fn check_fn(cx: &LateContext<'_>, fn_sig: FnSig<'_>) {
+    if fn_sig.header.abi == Abi::Rust
+        && let Some(snippet) = snippet_opt(cx, fn_sig.span)
+        && let Some((fn_attrs, _)) = snippet.split_once("fn")
+        && !fn_attrs.contains("extern")
+    {
+        let sugg_span = fn_sig.span
+            .with_lo(fn_sig.span.lo() + BytePos::from_usize(fn_attrs.len()))
+            .shrink_to_lo();
+
+        span_lint_and_then(
+            cx,
+            NO_MANGLE_WITH_RUST_ABI,
+            fn_sig.span,
+            "`#[no_mangle]` set on a function with the default (`Rust`) ABI",
+            |diag| {
+                diag.span_suggestion(
+                    sugg_span,
+                    "set an ABI",
+                    "extern \"C\" ",
+                    Applicability::MaybeIncorrect,
+                )
+                .span_suggestion(
+                    sugg_span,
+                    "or explicitly set the default",
+                    "extern \"Rust\" ",
+                    Applicability::MaybeIncorrect,
+                );
+            },
+        );
+    }
+}
+
+/// Check for static items with a type that is implicitly using the Rust ABI.
+fn check_static(cx: &LateContext<'_>, item: &Item<'_>, ty: &Ty<'_>) {
+    // TODO(Centri3): Once/if Rust allows an explicit `#[repr(Rust)]`, change this to check for
+    // that. It should work by checking `ReprOptions::flags` for `IS_EXPLICIT_RUST`, whatever
+    // value that may be.
+
+    if let ty::Adt(def, _) = hir_ty_to_ty(cx.tcx, ty).kind()
+        // No explicit representation. `align` and `pack` don't need to be checked as,
+        // afaik, what mostly matters is just the Rust ABI.
+        && def.repr().flags.bits() == 0
+    {
+        span_lint_and_then(
+            cx,
+            NO_MANGLE_WITH_RUST_ABI,
+            item.span,
+            "`#[no_mangle]` set on a static with the default (`Rust`) ABI",
+            |diag| {
+                diag.span_note(ty.span, "this type is implicitly `#[repr(Rust)]`");
+
+                if def.did().is_local() {
+                    diag.help("set an explicit ABI (like `#[repr(C)]`) on the type's definition");
+                }
+            },
+        );
     }
 }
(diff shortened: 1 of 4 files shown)


clippy_lints/src/no_mangle_with_rust_abi.rs
```diff
@@ -37,33 +46,75 @@ declare_lint_pass!(NoMangleWithRustAbi => [NO_MANGLE_WITH_RUST_ABI]);
...
+    }
+}
+
+/// Check for static items with a type that is implicitly using the Rust ABI.
+fn check_static(cx: &LateContext<'_>, item: &Item<'_>, ty: &Ty<'_>) {
+    // TODO(Centri3): Once/if Rust allows an explicit `#[repr(Rust)]`, change this to check for
+    // that. It should work by checking `ReprOptions::flags` for `IS_EXPLICIT_RUST`, whatever
+    // value that may be.
+
+    if let ty::Adt(def, _) = hir_ty_to_ty(cx.tcx, ty).kind()
```
Alexendoo, line 101: Since we can avoid `hir_ty_to_ty` here:

```suggestion
    if let ty::Adt(def, _) = cx.tcx.type_of(item.owner_id).skip_binder().kind()
```

clippy_lints/src/no_mangle_with_rust_abi.rs
```diff
@@ -37,33 +46,75 @@ declare_lint_pass!(NoMangleWithRustAbi => [NO_MANGLE_WITH_RUST_ABI]);
...
+        // No explicit representation. `align` and `pack` don't need to be checked as,
+        // afaik, what mostly matters is just the Rust ABI.
+        && def.repr().flags.bits() == 0
```
Alexendoo, line 104: Checking the flags directly runs into some implementation details, e.g. if `-Z randomize-layout` is passed the lint won't fire

Let's go with a more high level `!repr.c() && !repr.transparent() && !repr.simd() && repr.int.is_none()`. The latter because `#[repr(int)]` on enums have defined layout

Alexendoo:
We could add a check for (non unit) tuples as well as ADTs

clippy_lints/src/no_mangle_with_rust_abi.rs
Centri3: Out of curiosity, does `#[repr(int)]` define a variant's layout if it's a struct/tuple variant? Or is that what `#[repr(C, int)]` does? Because if so, we should probably lint that too if there are any variants that aren't unit.

Alexendoo: Yep - https://github.com/rust-lang/rfcs/blob/master/text/2195-really-tagged-unions.md#guide-level-explanation, you don't have to check the variant kinds

Centri3:
iirc slices (`&[T]`, not `[T]`) also have an undefined layout too, so we can maybe lint that as well?

clippy_lints/src/no_mangle_with_rust_abi.rs
```diff
@@ -37,33 +45,79 @@ declare_lint_pass!(NoMangleWithRustAbi => [NO_MANGLE_WITH_RUST_ABI]);
...
+    } {
+        span_lint_and_then(
+            cx,
+            NO_MANGLE_WITH_RUST_ABI,
+            item.span,
+            "`#[no_mangle]` set on a static with the default (`Rust`) ABI",
+            |diag| {
+                diag.span_note(ty.span, "this type is implicitly `#[repr(Rust)]`")
+                    .note_once(
+                        "usage of this static in an external context is Undefined Behavior, as its layout is undefined",
```
Centri3, line 114: Don't think this note is necessary, but I think the additional context could be nice.

Centri3: The note is technically wrong though (undefined layout instead of unstable layout), that should be changed

tests/ui/no_mangle_with_rust_abi.stderr
```diff
@@ -1,10 +1,37 @@
...
+
 error: `#[no_mangle]` set on a function with the default (`Rust`) ABI
-  --> $DIR/no_mangle_with_rust_abi.rs:5:1
+  --> $DIR/no_mangle_with_rust_abi.rs:20:1
    |
 LL | fn rust_abi_fn_one(arg_one: u32, arg_two: usize) {}
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
-   = note: `-D clippy::no-mangle-with-rust-abi` implied by `-D warnings`
+   = note: calling this function in an external content is Undefined Behavior, as its calling convention (ABI) is undefined
```
ojeda, line 34:
```suggestion
   = note: calling this function in an external context is Undefined Behavior, as its calling convention (ABI) is undefined
```

tests/ui/no_mangle_with_rust_abi.stderr
```diff
@@ -1,10 +1,37 @@
...
+   |
+LL | pub static ZERO_UB: UsingMeInCIsUB = UsingMeInCIsUB(0, 0);
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+note: this type is implicitly `#[repr(Rust)]`
+  --> $DIR/no_mangle_with_rust_abi.rs:9:21
+   |
+LL | pub static ZERO_UB: UsingMeInCIsUB = UsingMeInCIsUB(0, 0);
+   |                     ^^^^^^^^^^^^^^
+   = note: usage of this static in an external context is Undefined Behavior, as its layout is undefined
```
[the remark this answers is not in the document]
Centri3: Nope, it should probably be reworded to "in foreign code"

clippy_lints/src/no_mangle_with_rust_abi.rs
Alexendoo: Is UB the right term for this? It might lead to UB if it causes a read from padding bytes for example but I don't imagine any usage of it at all is UB

e.g. it might be a valid case to pass a pointer to a rust type to other functions exposed by a rust crate, there's also the case of rust to rust FFI when building both sides with the exact same toolchain

tests/ui/no_mangle_with_rust_abi.rs
```diff
@@ -1,6 +1,21 @@
...
+pub static ZERO_UB: UsingMeInCIsUB = UsingMeInCIsUB(0, 0);
+
+#[repr(C)]
+pub struct UsingMeInCIsFine(u32, u32);
+#[no_mangle]
+pub static ZERO_DB: UsingMeInCIsFine = UsingMeInCIsFine(0, 0);
+
+#[no_mangle]
+pub static ZERO_UB_AGAIN: external::UsingMeInCIsUB = external::UsingMeInCIsUB(0, 0);
+
```
Alexendoo, line 18: Would be good to have a test for a static pointer/primitive

clippy_lints/src/no_mangle_with_rust_abi.rs
Centri3: > It might lead to UB if it causes a read from padding bytes for example

Also if it rearranges the fields, lets say one field has an invariant that it's always even, but the other doesn't, and the compiler's aware of this invariant; if you pass that to a function expecting them to *not* be rearranged, what will happen?

I think it's the right term, any use of it inside of Rust isn't defined behavior, its layout can be whatever it wants, outside of rust it isn't really UB, but I think it still conveys the point better than explicitly stating this.

> when building both sides with the exact same toolchain

Doesn't mean their layouts are the same, it's still unstable even with the same compiler flags and all.

Alexendoo: > it's still unstable even with the same compiler flags and all

It is a grey area but it's still done, rustc and clippy dynamically link to libstd/librustc_driver for example

> outside of rust it isn't really UB, but I think it still conveys the point better than explicitly stating this

UB is a term that is so often misunderstood and I don't think we should worsen that. It's fine to just say that the layout is unspecified

Centri3: > It is a grey area but it's still done, rustc and clippy dynamically link to libstd/librustc_driver for example

Those are `dylib`s. They have the necessary info to ensure everything works fine, though whether or not you can actually read that info is up to whether they're unstable even between the same compiler version (afaik, they are, but in practice they may work fine between the same compiler version. Otherwise you wouldn't be able to link to libstd anyway)

This mostly concerns `cdylib`. If you're using `dylib`, there's no point to using `no_mangle` because you can just access the static idiomatically using a single path.

tests/ui/no_mangle_with_rust_abi.rs
Centri3: Pointers will require a wrapper type since they are `!Send` + `!Sync`, and it only lints ADTs already, so I *think* that should work ok.

clippy_lints/src/no_mangle_with_rust_abi.rs
Alexendoo: I'm aware it's not the same situation, it's just a grey area example

tests/ui/no_mangle_with_rust_abi.rs
Alexendoo: Oh yeah forgot that about pointers, was looking at some pre 1.0 code apparently. Still want a test for some other primitive though, yeah the implementation looks good for them

asquared31415:
I think that the current implementation is correct given a conservative approach to the current unsafe guarantees, but I believe that there are wishes to make more of `repr(Rust)` defined.  If that happens, this lint may need to be updated to be less restrictive.

Alexendoo (approved):

Alexendoo:
Yeah sounds good to me, we can adapt this if future guarantees are ever added

r=me with rebase/squash

Alexendoo:
Oh @asquared31415's comment reminded me of something, `Option<&T>` and such

Centri3:
iirc `Option<&T>` still doesn't have a stable ABI, even if the underlying type is (which, in this case, has a stable ABI). The discriminant can be whatever it wants so I think it should still lint it.

(Unless you meant something other than not linting it, not sure)

Centri3:
Will rebase later today

Alexendoo:
https://doc.rust-lang.org/std/primitive.reference.html says

> In fact, `Option<&T>` has the same memory representation as a nullable but aligned pointer, and can be passed across FFI boundaries as such.

asquared31415:
This is a complete list of the types that have well defined layouts inside `Option`: https://doc.rust-lang.org/std/option/index.html#representation

Centri3:
> [doc.rust-lang.org/std/primitive.reference.html](https://doc.rust-lang.org/std/primitive.reference.html) says
> 
> > In fact, `Option<&T>` has the same memory representation as a nullable but aligned pointer, and can be passed across FFI boundaries as such.

Odd, the type layout section of the reference should really mention that

While `Option<&T>` is the same as `&T`, the underlying type must be `repr(C)` for it to be FFI safe. I mean, you can pass it around, but you can't easily give it to an external function expecting it. So we'll need to probe a little deeper into `Option` types to see if `T` is `repr(Rust)`

asquared31415:
This does not need to do any inspection of the type behind one of the pointer-like types because `Option<&T>` is as FFI-safe as a raw pointer.  We are not be linting on `*const NotFFISafe`, because the inner type of a pointer (when it's `Sized`) does not affect the ABI, So none of the pointer-like `Option`s need inspection besides "is an `Option` of one of these pointer types"
(for `T: Sized`):
- `Option<Box<T>>`
- `Option<&T>`
- `Option<&mut T>`
- `Option<fn>` (with *or without* an `extern` ABI)
- `Option<ptr::NonNull<T>>`

The only remaining case that has guaranteed layout is the `Option<NonZeroX>` types, which are represented as plain integers (and don't have types to inspect further, like references).

Centri3:
> `Option<&T>` is as FFI-safe as a raw pointer.

Yes, the `Option<&...>` part is safe, but the `T` is absolutely not. I was not referring to `Option<&...>` in specific; that's FFI-safe. The underlying type isn't however, and because of that it's basically useless.

You can pass it around safely, but accessing the inner data on its own won't work. Unless all fields are private doing that is a terrible idea.

> We are not be linting on `*const NotFFISafe`

And we probably should with `#[no_mangle]`.

As I said, you can pass the pointer/reference around care-free, but you can't read from it in an external function. You can pass it to a free function defined in the crate that defines `NotFFISafe`, that returns a mutable reference to that field. So it's better than nothing, but not as useful as a `repr(C)` type.

I don't see any reason to assume the author has gone through that effort if there are public fields. They're likely to just do a null check, then read it and cause a multitude of issues.

(If there are public fields, that's a major footgun anyway for rust code in a `cdylib`.)

asquared31415:
In my experience the "opaque pointer" pattern is fairly common.  And while we could say "use a FFI-safe type", that won't satisfy most people.  In fact we even have an exception for `*const T` such that it does not trigger the FFI safety lint in rustc for similar reasons.  Especially `*const ()` which is suggested as a opaque pointer type.

Centri3:
There isn't any issue with using `repr(C)` on an opaque type. The nomicon even suggests it in fact (is it Rustonomicon or the nomicon?), I'd say you should since it immediately represents it comes from/is sent to C. It's definitely pedantic, since it's never constructed, but clippy can be more pedantic than rustc by default.

I think it makes sense to not lint this stuff in rustc. It can be correct, but it most of the time isn't (or could be written better). imo, in clippy it'd make sense to lint this since either the code is incorrect or the purpose of the struct can be stated better (both ways are solved by adding `repr(C)`.

> Especially `*const ()` which is suggested as a opaque pointer type.

We already don't lint `*const ()` or `()`.

Centri3:
For now, yes. I'll update this soon to do so. However, in the future we should have an off-by-default configuration option to make it a bit more pedantic.

xFrednet:

Centri3:
Triaging my old PRs in case someone wants to pick them up, this needs a tiny bit of work. Mostly around opaque pointers, but also I left a TODO for when `repr(Rust)` is explicitly allowed - it now is, so we should handle it.

In terms of opaque pointers, tbh I don't think `*const T` will show up enough for it to matter. By that I mean, we can just ignore it like the other lint does.

I feel like another lint is still in order for this case since opaque pointers feel... weird, considering you can always define another unit struct. Generally rust prefers explicit over implicit, does it not? So I think that makes sense.

PR #11261, opened 2023-07-30
[`unnecessary_find_map`]: look for then_some

### LINKED ISSUE #11260: `unnecessary_find_map` should see through `then_some`
### Summary

It would be nice if Clippy could also warn me about using `then_some` in return of a `find_map` closure, like in this example:

```
std::iter::once(1).find_map(|n| (n > 1).then_some(n));
```

### Lint Name

unnecessary_find_map

### Reproducer

I tried this code:

```rust
    // warns
    std::iter::once(1).find_map(|n| if n > 1 { Some(n) } else { None });
    // does not
    std::iter::once(1).find_map(|n| (n > 1).then_some(n));
    // better solution
    std::iter::once(1).find(|n| *n > 1);
```

The `unnecessary_find_map` lint should flag the second example as well

Instead, this happened:

No warning is emitted

### Version

_No response_

y21:
Closes #11260

changelog: [`unnecessary_find_map`]: lint `.then_some()` in closure

### DIFF
--- a/tests/ui/unnecessary_find_map.rs
+++ b/tests/ui/unnecessary_find_map.rs
@@ -21,3 +21,7 @@ fn main() {
 fn find_map_none_changes_item_type() -> Option<bool> {
     "".chars().find_map(|_| None)
 }
+
+fn issue11260() {
+    let _x = std::iter::once(1).find_map(|n| (n > 1).then_some(n));
+}
--- a/tests/ui/unnecessary_find_map.rs
+++ b/tests/ui/unnecessary_find_map.rs
@@ -23,5 +23,7 @@ fn find_map_none_changes_item_type() -> Option<bool> {
 }
 
 fn issue11260() {
+    let y = Some(1);
     let _x = std::iter::once(1).find_map(|n| (n > 1).then_some(n));
+    let _x = std::iter::once(1).find_map(|n| (n > 1).then_some(y)); // different option, so can't be just `.find()`
 }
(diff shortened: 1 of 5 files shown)


tests/ui/unnecessary_find_map.rs
```diff
@@ -21,3 +21,7 @@ fn main() {
 fn find_map_none_changes_item_type() -> Option<bool> {
     "".chars().find_map(|_| None)
 }
+
+fn issue11260() {
+    let _x = std::iter::once(1).find_map(|n| (n > 1).then_some(n));
```
blyxyas, line 26: Could you add some more tests? :cowboy_hat_face:

blyxyas (commented):
Just a mini-nit, everything else seems pretty good :cow2:

tests/ui/unnecessary_find_map.rs
y21: Do you have anything specific in mind? I realized that this fix also applies to `unnecessary_filter_map`, so I'm going to add a test that it also does the right thing there, but I'm not sure what else we would want to test here

y21: Added a test for unnecessary_filter_map and testing that it doesn't incorrectly suggest `.find()` when calling `.then_some()` with a different option

blyxyas (approved):
LGTM, Thanks :heart:!

PR #11265, opened 2023-07-31
Don't escape unicode escape braces in `print_literal`

### LINKED ISSUE #11264: clippy::write_literal uses format-string escaping incorrectly in suggested fix
### Summary

`clippy::print_literal` and `clippy::write_literal` warn about using a literal as an argument to the formatting macros. Instead, such literals can be inlined directly into the format string, which is more efficient.

These warnings generate a suggested fix which does this inlining. However, they perform escaping incorrectly when constructing the new suggested format string, leading to incorrect code. Braces are both part of the underlying string literal language level syntax, and part of the library level formatting string syntax, and the suggested fix does not correctly handle the interaction between these two different uses of brace characters in the source code.

### Reproducer

I tried this code:

```rust
fn main() {
    use std::io::Write;
    println!("{}", "\u{1F980}");
    writeln!(std::io::stdout(), "{}", "\u{1F980}").ok();
}
```

- Correct fixes suggested to replace the use of a literal as a formatted argument (on writeln! and println!).

Instead, this happened:
- Incorrect fixes were suggested. In particular, note that the suggested fix sets the format string to `"\u{{1F980}}"`. This is wrong, because the braces are part of the Unicode character escape sequence in the string, _not_ part of the format string that is parsed by the std formatting machinery, hence they should not be doubled as they are in the Clippy-suggested fix.

```
warning: literal with an empty format string
 --> src/main.rs:3:20
  |
3 |     println!("{}", "\u{1F980}");
  |                    ^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#print_literal
  = note: `#[warn(clippy::print_literal)]` on by default
help: try this
  |
3 -     println!("{}", "\u{1F980}");
3 +     println!("\u{{1F980}}");
  |

warning: literal with an empty format string
 --> src/main.rs:4:39
  |
4 |     writeln!(std::io::stdout(), "{}", "\u{1F980}").ok();
  |                                       ^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#write_literal
  = note: `#[warn(clippy::write_literal)]` on by default
help: try this
  |
4 -     writeln!(std::io::stdout(), "{}", "\u{1F980}").ok();
4 +     writeln!(std::io::stdout(), "\u{{1F980}}").ok();
  |
```

### Version

_No response_

### Additional Labels

_No response_

Alexendoo:
Fixes #11264

changelog: none

### DIFF
--- a/tests/ui/print_literal.rs
+++ b/tests/ui/print_literal.rs
@@ -51,4 +51,18 @@ fn main() {
     // The string literal from `file!()` has a callsite span that isn't marked as coming from an
     // expansion
     println!("file: {}", file!());
+
+    // Braces in unicode escapes should not be escaped
+    println!("{}", "{} \x00 \u{ab123} \\\u{ab123} {:?}");
+    println!("{}", "\\\u{1234}");
+    // This does not lint because it would have to suggest unescaping the character
+    println!(r"{}", "\u{ab123}");
+    // These are not unicode escapes
+    println!("{}", r"\u{ab123} \u{{");
+    println!(r"{}", r"\u{ab123} \u{{");
+    println!("{}", r"\{ab123} \u{{");
+    println!("{}", "\\u{ab123}");
+    println!("{}", "\\\\u{1234}");
+
+    println!("mixed: {} {world}", "{hello}");
 }
--- a/clippy_lints/src/write.rs
+++ b/clippy_lints/src/write.rs
@@ -486,9 +486,9 @@ fn check_literal(cx: &LateContext<'_>, format_args: &FormatArgs, name: &str) {
             && let rustc_ast::ExprKind::Lit(lit) = &arg.expr.kind
             && !arg.expr.span.from_expansion()
             && let Some(value_string) = snippet_opt(cx, arg.expr.span)
-    {
+        {
             let (replacement, replace_raw) = match lit.kind {
-                LitKind::Str | LitKind::StrRaw(_)  => match extract_str_literal(&value_string) {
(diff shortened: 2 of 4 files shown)


tests/ui/print_literal.rs
```diff
@@ -42,4 +42,7 @@ fn main() {
     // The string literal from `file!()` has a callsite span that isn't marked as coming from an
     // expansion
     println!("file: {}", file!());
+
+    // Braces in unicode escapes should not be escaped
+    println!("{}", "{} \u{ab123} {}");
```
llogiq, line 47: How about a test case with a format string of `"\\u{ab123}"` with binding of `ab123` to something?

Or a test case with additional format arg things like debug output or width specification?

Just to make sure we cover all cases.

llogiq (commented):
This looks ok to me, but I'd like to see more tests.

tests/ui/print_literal.rs
Alexendoo: `\\u{..}` was a good call, it didn't handle that correctly so I've replaced the regex with a state machine

samueltardieu:
Isn't this PR ready to be merged?

llogiq:
I think another test case with `\\\u{1234}` would be useful, otherwise this looks good. Thanks for the heads-up and sorry for the delay.

llogiq:
Thank you!

@bors r+

PR #11284, opened 2023-08-02
Alphabetically order arms in `methods/mod.rs` match

lengyijun:
changelog: none

just order the arms in the match block

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -3881,6 +3881,13 @@ impl Methods {
                         unnecessary_lazy_eval::check(cx, expr, recv, arg, "and");
                     }
                 },
+                ("any", [arg]) if let ExprKind::Closure(arg) = arg.kind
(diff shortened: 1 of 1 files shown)


lengyijun:
not sure the `skip` arm, it looks good to be neighbour to `last` arm

blyxyas:
If we're keeping an alphabetical order, then making an exception just for `skip` would be inconsistent.

Centri3:
While it's pretty, I'm pretty sure having them in alphabetical order would affect performance as match arms are always evaluated in the order they're written. It'd have to go through tons of arms for even the most common methods, like `len`.

For now, let's go with the nice alphabetical order but in the future we should probably reorder them based on how common that method is (and the following checks), like `write` vs `resize`.

blyxyas:
I don't think it matters as the majority of users use Clippy to warn, it doesn't need to fail fast.

It'd still be good to reorder this by use in the future, though.

blyxyas (approved):
LGTM, Thanks ❤️!

PR #11289, opened 2023-08-03
 [iter_overeager_cloned]: detect `.cloned().filter()` and `.cloned().find()`

lengyijun:
changelog: [`iter_overeager_cloned`]: detect `.cloned().filter()` and `.cloned().find()`

Key idea:
```
// before
iter.cloned().filter(|x| unimplemented!() )
// after
iter.filter(|&x| unimplemented!() ).cloned()

// before
iter.cloned().filter( foo )
// after
// notice `iter` must be `Iterator<Item= &T>` (callee of `cloned()`)
// so the parameter in the closure of `filter` must be `&&T`
// so the deref is safe
iter.filter(|&x| foo(x) ).cloned()
```

### DIFF
--- a/clippy_lints/src/methods/iter_overeager_cloned.rs
+++ b/clippy_lints/src/methods/iter_overeager_cloned.rs
@@ -2,7 +2,7 @@ use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::{implements_trait, is_copy};
 use rustc_errors::Applicability;
-use rustc_hir::Expr;
+use rustc_hir::{Expr, ExprKind};
 use rustc_lint::LateContext;
 use rustc_middle::ty;
 use rustc_span::sym;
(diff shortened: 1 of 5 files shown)


clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -47,13 +61,33 @@ pub(super) fn check<'tcx>(
...
+                            diag.span_suggestion(replace_span, "try", snip, Applicability::MachineApplicable);
+                        }
+                    }
                 }
             }
         );
     }
 }
+
+fn transform_closure<'tcx>(cx: &LateContext<'tcx>, closure: Expr<'tcx>) -> Option<String> {
```
blyxyas, line 88: Is there a case where this `None` cases could have some significance? Or it shouldn't never be `None`? In the case it can be `None` it should be accounted in the `match` in the lint emitter closure.

tests/ui/iter_overeager_cloned.fixed
```diff
@@ -21,18 +21,22 @@ fn main() {
...
-    let _ = vec.iter().cloned().filter(|x| x.starts_with('2'));
+    let _ = vec.iter().filter(|&x| x.starts_with('2')).cloned();
+
+    let _ = vec.iter().find(|&x| x == "2").cloned();
+
+    {
+        let f = |x: &String| x.starts_with('2');
+        let _ = vec.iter().filter(|&x| f(x)).cloned();
+        let _ = vec.iter().find(|&x| f(x)).cloned();
+    }
```
blyxyas, line 32: Are we sure dererencing the closure argument wouldn't break logic? What about the argument being a tuple? Could you also add a test with a `move`?

Centri3: This could also be an issue if the dereferenced type doesn't implement some trait: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=91cf6c202658738b0fba1b53c5510d01

Also, if it doesn't implement `Copy` this won't work if it's `&T` -> `T`

blyxyas: @lengyijun Could you add a test for this?

blyxyas (approved):
LGTM, Thanks! :heart:

PR #11293, opened 2023-08-04
feat: add cfg_not_test lint

### LINKED ISSUE #11234: `cfg_not_test`
### What it does

Disallows usage of any conditional compilation that always excludes code from test builds.

### Advantage

- Prevents code being excluded from tests builds. Preventing the case where a codebase appears to have 100% coverage while having untested code.
- Enforces simplicity in testing, guarding against excessive mocking anti-pattern.

### Drawbacks

- May prevent rare legitimate use cases for excluding code from test runs.

### Example

The following would trigger this lint:

```rust
#[cfg(not(test))]
```

```rust
#[cfg(all(debug_assertions, not(test)))]
```

```rust
#[cfg(not(any(not(debug_assertions), test)))]
```

mrnossiom:
Fixes #11234 

changelog: new lint: [`cfg_not_test`]

I don't know whether to lint only the `attr` or also the item associated to it. I guess this would mean putting the check in another place than `check_attribute` but I can't find a way to get the associated item to the attribute.

Also, I'm not sure how to document this lint, I feel like my explications are bad.

### DIFF
--- a/clippy_lints/src/cfg_not_test.rs
+++ b/clippy_lints/src/cfg_not_test.rs
@@ -1,0 +1,60 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use rustc_ast::NestedMetaItem;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::declare_lint_pass;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `cfg` that excludes code from `test` builds. (i.e., `#{cfg(not(test))]`)
+    ///
+    /// ### Why is this bad?
+    /// This may give the false impression that a codebase has 100% coverage, yet actually has untested code.
+    /// Enabling this also guards against excessive mockery as well, which is an anti-pattern.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # fn important_check() {}
+    /// #[cfg(not(test))]
+    /// important_check(); // I'm not actually tested, but not including me will falsely increase coverage!
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # fn important_check() {}
+    /// important_check();
+    /// ```
+    #[clippy::version = "1.73.0"]
+    pub CFG_NOT_TEST,
+    restriction,
+    "enforce against excluding code from test builds"
+}
+
+declare_lint_pass!(CfgNotTest => [CFG_NOT_TEST]);
+
+impl EarlyLintPass for CfgNotTest {
+    fn check_attribute(&mut self, cx: &EarlyContext<'_>, attr: &rustc_ast::Attribute) {
+        if attr.has_name(rustc_span::sym::cfg) && contains_not_test(attr.meta_item_list().as_deref(), false) {
+            span_lint_and_then(
+                cx,
+                CFG_NOT_TEST,
+                attr.span,
+                "code is excluded from test builds",
+                |diag| {
+                    diag.help("consider not excluding any code from test builds");
+                    diag.note_once("this could increase code coverage despite not actually being tested");
+                },
+            );
+        }
+    }
+}
+
+fn contains_not_test(list: Option<&[NestedMetaItem]>, not: bool) -> bool {
+    list.is_some_and(|list| {
+        list.iter().any(|item| {
+            item.ident().is_some_and(|ident| match ident.name {
+                rustc_span::sym::not => contains_not_test(item.meta_item_list(), !not),
+                rustc_span::sym::test => not,
+                _ => contains_not_test(item.meta_item_list(), not),
+            })
+        })
+    })
+}
--- a/tests/ui/cfg_not_test.rs
+++ b/tests/ui/cfg_not_test.rs
@@ -1,0 +1,32 @@
+#![allow(unused)]
+#![warn(clippy::cfg_not_test)]
+
+fn important_check() {}
+
+fn main() {
+    // Statement
+    #[cfg(not(test))]
+    let answer = 42;
+
+    // Expression
+    #[cfg(not(test))]
+    important_check();
+
+    // Make sure only not(test) are checked, not other attributes
+    #[cfg(not(foo))]
+    important_check();
+}
+
+#[cfg(not(not(test)))]
+struct CfgNotTest;
+
+// Deeply nested `not(test)`
+#[cfg(not(test))]
+fn foo() {}
+#[cfg(all(debug_assertions, not(test)))]
+fn bar() {}
+#[cfg(not(any(not(debug_assertions), test)))]
+fn baz() {}
+
+#[cfg(test)]
+mod tests {}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5254,6 +5254,7 @@ Released 2018-09-13
 [`cast_sign_loss`]: https://rust-lang.github.io/rust-clippy/master/index.html#cast_sign_loss
 [`cast_slice_different_sizes`]: https://rust-lang.github.io/rust-clippy/master/index.html#cast_slice_different_sizes
 [`cast_slice_from_raw_parts`]: https://rust-lang.github.io/rust-clippy/master/index.html#cast_slice_from_raw_parts
+[`cfg_not_test`]: https://rust-lang.github.io/rust-clippy/master/index.html#cfg_not_test
 [`char_lit_as_u8`]: https://rust-lang.github.io/rust-clippy/master/index.html#char_lit_as_u8
 [`chars_last_cmp`]: https://rust-lang.github.io/rust-clippy/master/index.html#chars_last_cmp
 [`chars_next_cmp`]: https://rust-lang.github.io/rust-clippy/master/index.html#chars_next_cmp
--- a/clippy_lints/src/declared_lints.rs
+++ b/clippy_lints/src/declared_lints.rs
@@ -104,6 +104,7 @@ pub(crate) static LINTS: &[&crate::LintInfo] = &[
     crate::casts::REF_AS_PTR_INFO,
     crate::casts::UNNECESSARY_CAST_INFO,
     crate::casts::ZERO_PTR_INFO,
+    crate::cfg_not_test::CFG_NOT_TEST_INFO,
     crate::checked_conversions::CHECKED_CONVERSIONS_INFO,
     crate::cognitive_complexity::COGNITIVE_COMPLEXITY_INFO,
     crate::collapsible_if::COLLAPSIBLE_ELSE_IF_INFO,
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -92,6 +92,7 @@ mod box_default;
 mod byte_char_slices;
 mod cargo;
 mod casts;
+mod cfg_not_test;
 mod checked_conversions;
 mod cognitive_complexity;
 mod collapsible_if;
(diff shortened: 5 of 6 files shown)


clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,62 @@
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// // an important that is a pain to test but not testing it can actually be dangerous
+    ///
+    /// #[cfg(not(test))]
+    /// important_check();
+    /// ```
+    ///
+    /// You should instead don't exclude any specific code from your test builds.
```
Centri3, line 25:
```suggestion
    /// ### What it does
    /// Checks for usage of `cfg` that excludes code from `test` builds. (i.e., `#{cfg(not(test))]`)
    ///
    /// ### Why is this bad?
    /// This may give the false impression that a codebase has 100% coverage, yet actually has untested code.
    ///
    /// ### Example
    /// ```rust
    /// #[cfg(not(test))]
    /// important_check(); // I'm not actually tested, but not including me will falsely increase coverage!
    /// ```
    /// Use instead:
    /// ```rust
    /// important_check();
    /// ```
```

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,62 @@
...
+    "enforce against excluding code from test builds"
+}
+
+declare_lint_pass!(CfgNotTest => [CFG_NOT_TEST]);
+
+impl EarlyLintPass for CfgNotTest {
+    fn check_attribute(&mut self, cx: &EarlyContext<'_>, attr: &rustc_ast::Attribute) {
+        if let Some(ident) = attr.ident()
+            && ident.name == rustc_span::sym::cfg
+            && contains_not_test(attr.meta_item_list().as_deref(), false)
```
Centri3, line 38: This won't detect it if `--test` is passed.

This could be a pre-expansion pass lint instead so it's ran before `cfg` is expanded, *but these are very buggy and soft-deprecated*. See #6610 for an example of one of these issues. If it's made into a pre-expansion pass, this should probably be moved to `attrs.rs` and combined with `maybe_misused_cfg`, that way we can minimize its usage to only a few lints.

Whether or not it not detecting it if `--test` is passed is up to you, it's probably not as big of a deal as other lints.

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            span_lint_and_help(
+                cx,
+                CFG_NOT_TEST,
+                attr.span,
+                "code is excluded from test builds",
+                None,
+                "consider not excluding any code from test builds",
```
Centri3, line 46: nit: Imo this should give a note instead, like "this will increase code coverage despite not actually being tested"

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,62 @@
...
+            );
+        }
+    }
+}
+
+fn contains_not_test(list: Option<&[NestedMetaItem]>, not: bool) -> bool {
```
Centri3, line 52: I believe this can use `has_name` on the attribute instead?

clippy_lints/src/cfg_not_test.rs
MrNossiom: maybe but code coverage is not the only issue
(or maybe you meant to add another note, I don't know how to do that... another function, using the `Diasgnostic` struct for more complex lints?)

clippy_lints/src/cfg_not_test.rs
MrNossiom: I could have done that, but I think this doesn't check when the test is deeply nested like in the example `#[cfg(not(any(not(debug_assertions), test)))]`

clippy_lints/src/cfg_not_test.rs
MrNossiom: `span_lint_and_then()`

tests/ui/cfg_not_test.rs
```diff
@@ -0,0 +1,19 @@
+#![allow(unused)]
```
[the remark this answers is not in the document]
MrNossiom: Why souldn't we lint those ?

clippy_lints/src/cfg_not_test.rs
Centri3: > I didn't know about `cfg` expansion, does this only concern `cfg_attr`?

No, it concerns `cfg` as well. In the HIR, and even AST, `cfg`s are expanded to either *nothing* or unchanged depending on whether it should "be in that compilation run". So, if `not(test)` is successful, it'll be entirely gone before linting occurs. Thus, if the user passes `--test`, there will be no warnings.

For something like `not(test)`, not a big deal, but something like `cfg(all(windows, not(test))` is not good.

I'd recommend messing around a bit on godbolt if you're still having trouble understanding: https://godbolt.org/z/WxPWeM416, you can see the AST before macro-expansion, after, and the HIR by passing different things to `unpretty`.

> (btw, why isn't there more folders for lints?)

Only because nobody's taken the time to :D I'd be 100% in favor of merging a PR to clean that up a bit.

> I don't know if grouping lints has an advantage

It's cleaner, but more importantly, it puts them in the same lint pass which has a *huge* performance boost. For on-by-default (warn/deny) lints, that you almost never want to globally disallow, this is great. In the case of this lint, it's allow-by-default, but I'd still say it's ok in the same lint pass as it does the exact same thing as other lints.

> Also, this deserves his own lint, and IMHO has not much to do with `maybe_misused_cfg`.

I actually agree now; that one is warn-by-default and this one is a restriction lint. So, no complaints from me.

tests/ui/cfg_not_test.rs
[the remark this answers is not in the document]
Centri3: Honestly, no idea, can't remember my reasoning :D

Now, I think it's fine and makes sense

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,59 @@
...
+use rustc_ast::NestedMetaItem;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `cfg` that excludes code from `test` builds. (i.e., `#{cfg(not(test))]`)
+    ///
+    /// ### Why is this bad?
+    /// This may give the false impression that a codebase has 100% coverage, yet actually has untested code.
```
Centri3, line 11: I didn't include it here before because I wasn't entirely sure what it meant when writing this, though this should include the 2nd advantage, like
```rust
/// Enabling this also guards against excessive mockery as well, which is an anti-pattern.
```
Could probably be worded better.

tests/ui/cfg_not_test.rs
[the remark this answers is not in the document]
Centri3: I think we also need a test for `#[cfg(not(not(test))]`. Not sure if that'd lint currently, though it shouldn't. If it does currently, you can track the current negation while visiting the meta items in the `CfgNotTest` struct, and only lint it's negative (an odd number of `not`s).

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn check_attribute(&mut self, cx: &EarlyContext<'_>, attr: &rustc_ast::Attribute) {
+        if attr.has_name(rustc_span::sym::cfg) && contains_not_test(attr.meta_item_list().as_deref(), false) {
+            span_lint_and_then(
+                cx,
+                CFG_NOT_TEST,
+                attr.span,
+                "code is excluded from test builds",
+                |diag| {
+                    diag.help("consider not excluding any code from test builds");
+                    diag.note("this could increase code coverage despite not actually being tested");
```
Centri3, line 42: Major nit, but I think these two should be `_once` diagnostics.

clippy_lints/src/cfg_not_test.rs
MrNossiom: > merging a PR to clean that up a bit.
Would definitely be willing to do that 😉 

> In the case of this lint, it's allow-by-default
I thought restriction was warn-by-default (it's not). In the original issue, @blyxyas states "I feel like the drawback is too heavy for an allow-by-default lint" which I understand as "the lint should be warn-by-default"

clippy_lints/src/cfg_not_test.rs
MrNossiom: `help_once` doesn't exist

Centri3: Hmm, thought that was a thing. That's fine then.

clippy_lints/src/cfg_not_test.rs
Centri3: I think she meant enabled-by-default, `restriction` is allow-by-default and generally, lints in that group aren't applicable in every situation or go against idiomatic code. There are valid reasons for using `not(test)`.

blyxyas:
> I thought restriction was warn-by-default (it's not). In the original issue, @blyxyas states "I feel like the drawback is too heavy for an allow-by-default lint" which I understand as "the lint should be warn-by-default"

Totally a typo, it should not be warn by default imo

Jarcho:
LGTM, but since you were already reviewing this. r? @Centri3

xFrednet:
@Centri3 is currently inactive. @Jarcho would you mind taking this review back? If you don't have the time rn, you can reassign it using `r? clippy` or to me :)

r? @Jarcho

mrnossiom:
Hi @Jarcho,

It's been a little time since I've made this PR. I would love to finish it. I remember being blocked by not understanding exactly what was blocking.

If I am correct, early passes are kind of unstable and buggy, but the information I need is not available after `cfg` expansion ? Or are `cfg`'s expanded before early passes ?

Jarcho:
Early passes are fine and have access to the ast before any `cfg` based removal has happened. The passes we don't want to add are pre-expansion passes, but that's not needed for this.

mrnossiom:
So this PR is good as is, then ?

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,60 @@
...
+                    diag.help("consider not excluding any code from test builds");
+                    diag.note_once("this could increase code coverage despite not actually being tested");
+                },
+            );
+        }
+    }
+}
+
+fn contains_not_test(list: Option<&[NestedMetaItem]>, not: bool) -> bool {
+    list.map_or(not, |list| {
```
Jarcho, line 51: Should be `is_some_and`. Right now `#[cfg(not(foo))]` will return true.

clippy_lints/src/cfg_not_test.rs
```diff
@@ -0,0 +1,60 @@
...
+        list.iter().any(|item| {
+            item.ident().map_or(false, |ident| match ident.name {
```
Jarcho, line 53: You can use `is_some_and` instead of `map_or(false, ...)`.

Jarcho (changes requested):

clippy_lints/src/cfg_not_test.rs
mrnossiom: Should I update tests to match such a case?

Jarcho: Definitely add a test for it.

Jarcho:
If you haven't already you'll have to rebase and rebless the tests. There was a `ui_test` update that changed the format a bit.

xFrednet:
Hey, this is a ping from triage. @Jarcho can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@mrnossiom Would you mind rebasing the PR again?

@rustbot ready

Jarcho (approved):

Jarcho:
Marked this as approved. @rust-lang/clippy can someone merge this if they see it before I do since this is now multiple times I've missed it.

Issue #11103, opened 2023-07-04
Optimize HashSet contains+insert usage


nyurik:
### What it does

Optimize hashset (and possibly similar HashMap?) usage when the user first uses `contains()` to check the non-existence of a value, followed by insertion.

### Advantage

- Faster code
- Shorter code

### Example

```rust
use std::collections::HashSet;

fn main() {
    let mut vals = HashSet::new();
    insert_if(&mut vals, 10);
}

fn insert_if(set: &mut HashSet<i32>, value: i32) {
    if !set.contains(&value) {
        set.insert(value);
        println!("inserted {value:?}");
    }
}
```

Could be written as:

```rust
fn insert_if(set: &mut HashSet<i32>, value: i32) {
    if set.insert(value) {
        println!("inserted {value:?}");
    }
}
```

lochetti:
Interesting! I will try to do it. I might need some help, though :)
If you know some other lint that has similar logic (looking if two expressions happen one after another and then linting) let me know.

@rustbot claim

Centri3:
I believe `unnecessary_unwrap`, @lochetti ^^

Jarcho:
`map_entry` does this for `HashMap`/`BTreeMap`.

PR #11296, opened 2023-08-05
New lint `hashset_insert_after_contains`

lochetti:
This PR closes [11103](https://github.com/rust-lang/rust-clippy/issues/11103).

This is my first PR creating a new lint, so I am opening as a draft and thanks for the patience :)

I also have used the simpler `span_lint()` when emitting the lint. If the reviewers think that it is not enough, I can try to evolve the code the use the others span_lint functions.

The idea of the lint is to find `insert` in hashmanps inside if staments that are checking if the hashmap `contains` the same value that is being inserted. This is not necessary since you could simply call the `insert` and check for the bool returned if you still need the if statement.

changelog: new lint: [`hashset_insert_after_contains`]

### DIFF
--- a/clippy_lints/src/hashset_insert_after_contains.rs
+++ b/clippy_lints/src/hashset_insert_after_contains.rs
@@ -1,13 +1,16 @@
-use clippy_utils::diagnostics::span_lint;
+use std::ops::ControlFlow;
+
+use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::visitors::for_each_expr;
 use clippy_utils::{higher, peel_hir_expr_while, SpanlessEq};
-use rustc_hir::intravisit::Visitor;
-use rustc_hir::{Block, Expr, ExprKind, Stmt, StmtKind, UnOp};
+use rustc_hir::{Expr, ExprKind, UnOp};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::sym;
+use rustc_span::{sym, Span};
 
 declare_clippy_lint! {
+    /// ### What it does
     /// Checks for usage of `contains` to see if a value is not
     /// present on `HashSet` followed by a `insert`.
     ///
@@ -36,75 +39,66 @@ declare_clippy_lint! {
     #[clippy::version = "1.73.0"]
     pub HASHSET_INSERT_AFTER_CONTAINS,
     perf,
-    "use of `contains` to see if a value is not present on a `HashSet` followed by a `insert`"
+    "unnecessary call to `HashSet::contains` followed by `HashSet::insert`"
 }
 declare_lint_pass!(HashsetInsertAfterContains => [HASHSET_INSERT_AFTER_CONTAINS]);
 
 impl<'tcx> LateLintPass<'tcx> for HashsetInsertAfterContains {
     fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
-        if expr.span.from_expansion() {
-            return;
-        }
-
-        let Some(higher::If {
+        if !expr.span.from_expansion()
+        && let Some(higher::If {
             cond: cond_expr,
             then: then_expr,
             ..
         }) = higher::If::hir(expr)
-        else {
-            return;
-        };
-
-        let Some(contains_expr) = try_parse_contains(cx, cond_expr) else {
-            return;
-        };
-
-        if !find_insert_calls(cx, &contains_expr, then_expr) {
-            return;
-        };
-        span_lint(
-            cx,
-            HASHSET_INSERT_AFTER_CONTAINS,
-            expr.span,
-            "usage of `HashSet::insert` after `HashSet::contains`. Remove the usage of `HashSet::contains` and just call `HashSet::contains` instead",
-        );
+        && let Some(contains_expr) = try_parse_contains(cx, cond_expr)
+        && find_insert_calls(cx, &contains_expr, then_expr) {
+            span_lint_and_then(
+                cx,
+                HASHSET_INSERT_AFTER_CONTAINS,
+                expr.span,
+                "usage of `HashSet::insert` after `HashSet::contains`",
+                |diag| {
+                    diag.note("`HashSet::insert` returns whether it was inserted")
+                        .span_help(contains_expr.span, "remove the `HashSet::contains` call");
+                },
+            );
+        }
     }
 }
 
 struct ContainsExpr<'tcx> {
     receiver: &'tcx Expr<'tcx>,
     value: &'tcx Expr<'tcx>,
+    span: Span,
 }
 fn try_parse_contains<'tcx>(cx: &LateContext<'_>, expr: &'tcx Expr<'_>) -> Option<ContainsExpr<'tcx>> {
     let expr = peel_hir_expr_while(expr, |e| match e.kind {
         ExprKind::Unary(UnOp::Not, e) => Some(e),
         _ => None,
     });
-    match expr.kind {
-        ExprKind::MethodCall(
-            path,
-            receiver,
-            [
-                Expr {
-                    kind: ExprKind::AddrOf(_, _, value),
-                    span: value_span,
-                    ..
-                },
-            ],
-            _,
-        ) => {
-            let receiver_ty = cx.typeck_results().expr_ty(receiver);
-            if value_span.ctxt() == expr.span.ctxt()
-                && is_type_diagnostic_item(cx, receiver_ty, sym::HashSet)
-                && path.ident.name == sym!(contains)
-            {
-                Some(ContainsExpr { receiver, value })
-            } else {
-                None
-            }
-        },
-        _ => None,
+    if let ExprKind::MethodCall(
+        path,
+        receiver,
+        [
+            Expr {
+                kind: ExprKind::AddrOf(_, _, value),
+                span: value_span,
+                ..
+            },
+        ],
+        span,
+    ) = expr.kind
+    {
+        let receiver_ty = cx.typeck_results().expr_ty(receiver);
+        if value_span.ctxt() == expr.span.ctxt()
+            && is_type_diagnostic_item(cx, receiver_ty, sym::HashSet)
+            && path.ident.name == sym!(contains)
+        {
+            return Some(ContainsExpr { receiver, value, span });
+        }
     }
+    None
 }
 
 struct InsertExpr<'tcx> {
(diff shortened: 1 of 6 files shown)


clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        if !find_insert_calls(cx, &contains_expr, then_expr) {
+            return;
+        };
+        span_lint(
+            cx,
+            HASHSET_INSERT_AFTER_CONTAINS,
+            expr.span,
+            "usage of `HashSet::insert` after `HashSet::contains`. Remove the usage of `HashSet::contains` and just call `HashSet::contains` instead",
+        );
```
Centri3, line 70: Instead of `span_lint`, I'd say we should use `span_lint_and_then`.
Here's what I'd suggest:
```suggestion
        span_lint_and_then(
            cx,
            HASHSET_INSERT_AFTER_CONTAINS,
            expr.span,
            "usage of `HashSet::insert` after `HashSet::contains`,
            |diag| {
                diag.note("`HashSet::insert` returns whether it was inserted").span_help(/* span for `contains` */, "remove the `HashSet::contains` call")
            },
        );
```

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// let mut set = HashSet::new();
+    /// let value = 5;
+    /// if set.insert(&value) {
+    ///     println!("inserted {value:?}");
+    /// }
+    /// ```
+    #[clippy::version = "1.73.0"]
+    pub HASHSET_INSERT_AFTER_CONTAINS,
+    perf,
+    "use of `contains` to see if a value is not present on a `HashSet` followed by a `insert`"
```
Centri3, line 39:
```suggestion
    "unnecessary call to `HashSet::contains` followed by `HashSet::insert`"
```

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+use clippy_utils::ty::is_type_diagnostic_item;
+use clippy_utils::{higher, peel_hir_expr_while, SpanlessEq};
+use rustc_hir::intravisit::Visitor;
+use rustc_hir::{Block, Expr, ExprKind, Stmt, StmtKind, UnOp};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// Checks for usage of `contains` to see if a value is not
```
Centri3, line 11: Missing `### What it does`

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// The receiver expression used in the contains call.
+    receiver: &'tcx Expr<'tcx>,
+    /// The value expression used in the contains call.
+    value: &'tcx Expr<'tcx>,
+    /// Whether or a lint shoud be emitted.
+    should_lint: bool,
+}
+
+impl<'tcx> Visitor<'tcx> for InsertSearcher<'_, 'tcx> {
+    fn visit_block(&mut self, block: &'tcx Block<'_>) {
```
Centri3, line 149: This is unnecessary, the default implementation calls `walk_block` which does something like this already

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+            },
+            _ => {
+                if let ExprKind::Block(block, _) = expr.kind {
+                    self.visit_block(block);
+                }
+            },
+        }
+    }
+
+    fn visit_stmt(&mut self, stmt: &'tcx Stmt<'_>) {
```
Centri3, line 175: This is unnecessary, the default implementation calls `walk_stmt` which does something like this already

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+    cx: &'cx LateContext<'tcx>,
+    /// The receiver expression used in the contains call.
+    receiver: &'tcx Expr<'tcx>,
+    /// The value expression used in the contains call.
+    value: &'tcx Expr<'tcx>,
+    /// Whether or a lint shoud be emitted.
+    should_lint: bool,
+}
+
+impl<'tcx> Visitor<'tcx> for InsertSearcher<'_, 'tcx> {
```
Centri3, line 148: I think, instead of using a visitor, this could use `for_each_expr` on the `then` block. I *think* you could copy the code from `visit_expr` into that closure, and change `Some` to `Break`.

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+struct ContainsExpr<'tcx> {
+    receiver: &'tcx Expr<'tcx>,
+    value: &'tcx Expr<'tcx>,
+}
+fn try_parse_contains<'tcx>(cx: &LateContext<'_>, expr: &'tcx Expr<'_>) -> Option<ContainsExpr<'tcx>> {
+    let expr = peel_hir_expr_while(expr, |e| match e.kind {
+        ExprKind::Unary(UnOp::Not, e) => Some(e),
+        _ => None,
+    });
+    match expr.kind {
```
Centri3, line 83: nit: This should probably be an `if let` instead

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+        ) => {
+            let receiver_ty = cx.typeck_results().expr_ty(receiver);
+            if value_span.ctxt() == expr.span.ctxt()
+                && is_type_diagnostic_item(cx, receiver_ty, sym::HashSet)
+                && path.ident.name == sym!(contains)
+            {
+                Some(ContainsExpr { receiver, value })
+            } else {
+                None
+            }
```
Centri3, line 104: nit: Unnecessary `else`:
```suggestion
                return Some(ContainsExpr { receiver, value });
            }
            
            None
```

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+        for stmt in block.stmts {
+            self.visit_stmt(stmt);
+        }
+        if let Some(expr) = block.expr {
+            self.visit_expr(expr);
+        }
+    }
+
+    fn visit_expr(&mut self, expr: &'tcx Expr<'_>) {
+        match try_parse_insert(self.cx, expr) {
```
Centri3, line 159: nit: This should probably be an `if let` instead

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,184 @@
...
+    #[clippy::version = "1.73.0"]
+    pub HASHSET_INSERT_AFTER_CONTAINS,
+    perf,
+    "use of `contains` to see if a value is not present on a `HashSet` followed by a `insert`"
+}
+declare_lint_pass!(HashsetInsertAfterContains => [HASHSET_INSERT_AFTER_CONTAINS]);
+
+impl<'tcx> LateLintPass<'tcx> for HashsetInsertAfterContains {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if expr.span.from_expansion() {
```
Centri3, line 45: I think we can turn this into an `if let`, like
```rust
if expr.span.from_expansion()
    && let Some(higher::If { ... }) = ...
    && let Some(contains_expr) = ...
    && find_insert_calls(...)
{
    /* lint */
}
```

Centri3 (commented):
Pretty good for a first lint, thanks!

lochetti:
> Pretty good for a first lint, thanks!

Thank you for the review, @Centri3 ! The code is looking way simpler with your suggestions :)

Could you please take a look again and let me know what you think? The changes are [in the commit](https://github.com/rust-lang/rust-clippy/pull/11296/commits/a64e5fef2b8e1011e3dedd52343925d659bf05f5)

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,132 @@
...
+        }
+    } else {
+        None
+    }
+}
+
+fn find_insert_calls<'tcx>(cx: &LateContext<'tcx>, contains_expr: &ContainsExpr<'tcx>, expr: &'tcx Expr<'_>) -> bool {
+    for_each_expr(expr, |e| {
+        if let Some(insert_expr) = try_parse_insert(cx, e)
+        && SpanlessEq::new(cx).eq_expr(contains_expr.receiver, insert_expr.receiver)
```
Centri3, line 124: You should probably call `peel_borrows` before here so that `&x.contains(y)` and `x.insert(y)` are considered the same.

lochetti:
Hey @Centri3 !

I think I have addressed almost everything from you comments.
I have made the lint work for three more cases that were not being linted before:
```
if (&set).contains(&value) {
    set.insert(value);
}
```
```
let borrow_value = &6;
if !set.contains(borrow_value) {
    set.insert(*borrow_value);
}
```
```
let borrow_set = &mut set;
if !borrow_set.contains(&value) {
    borrow_set.insert(value);
}
```

But I could not understand your comment about `from_expansion`, I am sorry.
Do you mind going in more details, please? Probably I am missing some knowledge here. If you could also come up with  a specific test case -if it is not hard, of course- that is not being addressed, it would make it easier for me to understand :)

Thanks! :)

Centri3:
The issue with `ctxt` here is that if they both come from the same context, then I *think* this will wrongly trigger. This is fine for root (no expansions), this is what you want to check for after all, but if they both come from the same expansion it will wrongly trigger.

However, I think ignoring all expansions isn't the right way to go about this; if it happens to expand to code that references a local, and also inserts into it, then it was likely intentionally done (if the local isn't also from an expansion, of course). So if it's not external, it should still lint.

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,127 @@
...
+    #[clippy::version = "1.73.0"]
+    pub HASHSET_INSERT_AFTER_CONTAINS,
+    perf,
+    "unnecessary call to `HashSet::contains` followed by `HashSet::insert`"
+}
+declare_lint_pass!(HashsetInsertAfterContains => [HASHSET_INSERT_AFTER_CONTAINS]);
+
+impl<'tcx> LateLintPass<'tcx> for HashsetInsertAfterContains {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if !expr.span.from_expansion()
```
Centri3, line 48: This probably also needs an `is_from_proc_macro` check right before emitting

clippy_lints/src/hashset_insert_after_contains.rs
```diff
@@ -0,0 +1,132 @@
...
+                kind: ExprKind::AddrOf(_, _, value),
+                span: value_span,
+                ..
+            },
+        ],
+        span,
+    ) = expr.kind
+    {
+        let receiver_ty = cx.typeck_results().expr_ty(receiver);
+        if value_span.ctxt() == expr.span.ctxt()
```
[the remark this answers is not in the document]
lochetti: Let's see if I understood the changes that I need to apply here:
1. Remove the `if value_span.ctxt() == expr.span.ctxt()` comparison.
2. Check if `value_span` is not from an expansion. If it is from an expasion, we will not lint
3. Check if `value_span` is nto from external_macro. If it is from an external macro, we will not lint.

Is it correct?

lochetti:
Hey @Centri3 hello! Sorry for that long delay, but I think I will not be able to finish the PR :grimacing: 
I am sorry.

Centri3:
No worries about it! I'll close it then and others can finish it if they wish. Feel free to create a new PR if you wish :3

lochetti:
I have opened a new PR to try to continue the work :) 
It is the https://github.com/rust-lang/rust-clippy/pull/12873

PR #11301, opened 2023-08-06
[`useless_conversion`]: don't lint if type parameter has unsatisfiable bounds for `.into_iter()` receiver

### LINKED ISSUE #11300: useless conversion wrongly suggests removing .into_iter() when ExactSizeIterator is required
### Summary

I have a function that takes an argument implementing `IntoIterator + ExactSizeIterator`.  When I call it, I supply an array with `.into_iter()`.  Clippy wrongly suggests removing the `.into_iter()` on the grounds that it's automatic.  However, if I do that then the code fails to compile because `ExactSizeIterator` is not satisfied.

### Lint Name

useless_conversion

### Reproducer

I tried this code:

```rust
pub fn foo<I>(i: I)
where I: IntoIterator<Item=i32> + ExactSizeIterator
{
    assert_eq!(i.len(), 3);
}

pub fn bar() {
    foo([1, 2, 3].into_iter());
}
```

I saw this happen:

```
warning: explicit call to `.into_iter()` in function argument accepting `IntoIterator`
 --> src/lib.rs:8:9
  |
8 |     foo([1, 2, 3].into_iter());
  |         ^^^^^^^^^^^^^^^^^^^^^ help: consider removing the `.into_iter()`: `[1, 2, 3]`
  |
note: this parameter accepts any `IntoIterator`, so you don't need to call `.into_iter()`
 --> src/lib.rs:2:10
  |
2 | where I: IntoIterator<Item=i32> + ExactSizeIterator
  |          ^^^^^^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
  = note: `#[warn(clippy::useless_conversion)]` on by default
```

No errors.

If I apply Clippy's suggest and remove the `.into_iter()`, then this happens:
```
error[[E0277]](https://doc.rust-lang.org/stable/error_codes/E0277.html): the trait bound `[i32; 3]: ExactSizeIterator` is not satisfied
 --> src/lib.rs:8:9
  |
8 |     foo([1, 2, 3]);
  |     --- ^^^^^^^^^ the trait `ExactSizeIterator` is not implemented for `[i32; 3]`
  |     |
  |     required by a bound introduced by this call
  |
  = help: the following other types implement trait `ExactSizeIterator`:
            &mut I
            Args
            ArgsOs
            ArrayChunksMut<'_, T, N>
            ArrayWindows<'_, T, N>
            Box<I, A>
            Chunks<'_, T>
            ChunksExact<'_, T>
          and 109 others
note: required by a bound in `foo`
 --> src/lib.rs:2:35
  |
1 | pub fn foo<I>(i: I)
  |        --- required by a bound in this function
2 | where I: IntoIterator<Item=i32> + ExactSizeIterator
  |                                   ^^^^^^^^^^^^^^^^^ required by this bound in `foo`
```

### Version

```text
rustc 1.73.0-nightly (399b06823 2023-07-20)
binary: rustc
commit-hash: 399b068235ceea440540539b3bfd1aeb82214a28
commit-date: 2023-07-20
host: x86_64-unknown-freebsd
release: 1.73.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

@rustbot label +`I-suggestion-causes-error`

y21:
Fixes #11300.

Before this PR, clippy assumed that if it sees a `f(x.into_iter())` call and the type at that argument position is generic over any `IntoIterator`, then the `.into_iter()` call must be useless because `x` already implements `IntoIterator`, *however* this assumption is not right if the generic parameter has more than just the `IntoIterator` bound (because other traits can be implemented for the IntoIterator target type but not the IntoIterator implementor, as can be seen in the linked issue: `<[i32; 3] as IntoIterator>::IntoIter` satisfies `ExactSizeIterator`, but `[i32; 3]` does not).

So, this PR makes it check that the type parameter only has a single `IntoIterator` bound. It *might* be possible to check if the type of `x` in `f(x.into_iter())` satisfies all the bounds on the generic type parameter as defined on the function (which would allow removing the `.into_iter()` call even with multiple bounds), however I'm not sure how to do that, and the current fix should always work.

**Edit:** This PR has been changed to check if any of the bounds don't hold for the type of the `.into_iter()` receiver, so we can still lint in some cases.

changelog: [`useless_conversion`]: don't lint `.into_iter()` if type parameter has multiple bounds

### DIFF
--- a/tests/ui/useless_conversion.rs
+++ b/tests/ui/useless_conversion.rs
@@ -244,6 +244,47 @@ mod issue11300 {
         // and `i32: Helper2<[i32, 3]>` is true, so this call is indeed unncessary.
         foo3([1, 2, 3].into_iter());
     }
+
+    fn ice() {
+        struct S1;
+        impl S1 {
+            pub fn foo<I: IntoIterator>(&self, _: I) {}
+        }
+
+        S1.foo([1, 2].into_iter());
+
+        // ICE that occured in itertools
+        trait Itertools {
+            fn interleave_shortest<J>(self, other: J)
+            where
+                J: IntoIterator,
+                Self: Sized;
+        }
+        impl<I: Iterator> Itertools for I {
+            fn interleave_shortest<J>(self, other: J)
+            where
+                J: IntoIterator,
+                Self: Sized,
+            {
+            }
+        }
+        let v0: Vec<i32> = vec![0, 2, 4];
+        let v1: Vec<i32> = vec![1, 3, 5, 7];
+        v0.into_iter().interleave_shortest(v1.into_iter());
+
+        trait TraitWithLifetime<'a> {}
+        impl<'a> TraitWithLifetime<'a> for std::array::IntoIter<&'a i32, 2> {}
+
+        struct Helper;
+        impl<'a> Helper {
+            fn with_lt<I>(&self, _: I)
+            where
+                I: IntoIterator + TraitWithLifetime<'a>,
+            {
+            }
+        }
+        Helper.with_lt([&1, &2].into_iter());
+    }
 }
 
 #[derive(Copy, Clone)]
--- a/tests/ui/useless_conversion.rs
+++ b/tests/ui/useless_conversion.rs
@@ -206,11 +206,43 @@ mod issue11300 {
         assert_eq!(i.len(), 3);
     }
 
+    trait Helper<T: ?Sized> {}
+    impl Helper<i32> for [i32; 3] {}
+    impl Helper<i32> for std::array::IntoIter<i32, 3> {}
+    impl Helper<()> for std::array::IntoIter<i32, 3> {}
+
+    fn foo2<X: ?Sized, I>(_: I)
+    where
+        I: IntoIterator<Item = i32> + Helper<X>,
+    {
+    }
+
+    trait Helper2<T> {}
+    impl Helper2<std::array::IntoIter<i32, 3>> for i32 {}
+    impl Helper2<[i32; 3]> for i32 {}
+    fn foo3<I>(_: I)
+    where
+        I: IntoIterator<Item = i32>,
+        i32: Helper2<I>,
+    {
+    }
+
     pub fn bar() {
         // This should not trigger the lint:
         // `[i32, 3]` does not satisfy the `ExactSizeIterator` bound, so the into_iter call cannot be
         // removed and is not useless.
         foo([1, 2, 3].into_iter());
+
+        // This should trigger the lint, receiver type [i32; 3] also implements `Helper`
+        foo2::<i32, _>([1, 2, 3].into_iter());
+
+        // This again should *not* lint, since X = () and I = std::array::IntoIter<i32, 3>,
+        // and `[i32; 3]: Helper<()>` is not true (only `std::array::IntoIter<i32, 3>: Helper<()>` is).
+        foo2::<(), _>([1, 2, 3].into_iter());
+
+        // This should lint. Removing the `.into_iter()` means that `I` gets substituted with `[i32; 3]`,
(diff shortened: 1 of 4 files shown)


Centri3:
We should probably check the rest of the bounds instead. I *think* `tcx.param_env` with the `DefId` of the passed function is all you need to pass to `implements_trait`, while making sure the only bounds passed are referencing the index of the `.into_iter()` argument, though I'm not sure if that'll work properly with something like `BoundWIthGeneric<T>` where `T: ?Sized` or something

tests/ui/useless_conversion.rs
```diff
@@ -194,11 +194,24 @@ mod issue11300 {
         assert_eq!(i.len(), 3);
     }
 
+    trait Helper {}
```
Centri3, line 197: Can you also add a test for something like `trait Helper<T: ?Sized>`? Basically any trait with generics.

A quick look at the new version makes me think this *should* work, but a test would be great

y21:
> We should probably check the rest of the bounds instead. 

Yeah, that's better, I changed it to that now. I ended up going with `EarlyBinder::bind(predicate).instantiate(substs)` since that allows directly "replacing" the `T` anywhere in the predicate with the `.into_iter()` receiver very nicely to check if all the bounds for `x.into_iter()` also hold for `x` and it worked out. <sub>(I totally didn't steal it from [another lint](https://github.com/rust-lang/rust-clippy/blob/526d1156bd783244f7e848d53e48ae3f9b32791f/clippy_lints/src/methods/unnecessary_to_owned.rs#L446))</sub>

tests/ui/useless_conversion.rs
y21: I changed the tests a bit, is this what you had in mind? Or something else?

Centri3: I was thinking a test that would lint, basically seeing if it would substitute the `T` in `Helper<T>` properly

y21: I see. Well, I think the tests I just added are also a good addition since that found a bug while I was testing stuff so I'm just going to leave them and add another test with your suggestion

y21: ok, added the test, WDYT now? Did I get it right this time or still wrong? 😅

Centri3: Yeah looks good :+1:, though the comment references `Helper` instead of `Helper2`

dswij (approved):
Thank you. This looks great. 

@y21 Can you help to solve the merge conflict, and I think it's good to merge afterward.

dswij:
Thank you.

@bors r+

Centri3:
Hmm. That's an ICE: https://github.com/rust-lang/rust-clippy/actions/runs/6083961075/job/16505016344#step:8:3267

y21:
Sorry, my bad. The code for getting the node args on method calls was slightly wrong.
To quickly explain what happened in that ICE: in this example code here
```rs
struct S1;
impl S1 {
    pub fn foo<I: IntoIterator>(&self, _: I) {..}
}

S1.foo([1, 2].into_iter());
```
The node args that we *want* are here:
```rs
S1.foo(..)
   ^^^^^^^
```
(this contains `[IntoIter<i32>]`)

However we were getting the node args from here:
```rs
S1.foo([1, 2].into_iter());
       ^^^^^^
```
That didn't really make sense, the expression didn't even have any substs, and of course, using those to substitute the generic parameter list `<I: IntoIterator>` won't work.


We didn't seem to have any tests for method calls that could have caught this, but I added some now.
Also, itertools on its own is a pretty good testsuite for this lint too 😄 
(I also ran the updated commit on itertools, cargo, chalk & lintcheck to make sure that this indeed fixes the ICE)

PR #11310, opened 2023-08-09
[`slow_vector_initialization`]: clarify why `Vec::new()` + resize is worse

y21:
#11198 extended this lint to also warn on `Vec::new()` + `resize(0, len)`, but did not update the lint documentation, so it left some confused (https://github.com/rust-lang/rust-clippy/issues/10938#issuecomment-1663880083).
This PR should make it a bit more clear. (cc @djc @vi what do you think about this?)

<details>
<summary>More details</summary>

Godbolt for `Vec::new()` + `.resize(x, 0)`: https://godbolt.org/z/e7q9xc9rG

The resize call first does a normal allocation (`__rust_alloc`):
```asm
alloc::raw_vec::finish_grow:
  ...
  cmp     qword ptr [rcx + 8], 0
  je      .LBB1_7  ; if capacity == 0 -> LBB1_7

.LBB1_7:
  ...
  call    qword ptr [rip + __rust_alloc@GOTPCREL]
```

*Then* a memset for zero initialization:
```asm
example::f:
  ...
  xor     esi, esi  ; 0
  call    qword ptr [rip + memset@GOTPCREL]
```
------------

Godbolt for `vec![0; len]`: https://godbolt.org/z/M3vr53vWY

Important bit:
```asm
example::f:
  ...
  call    qword ptr [rip + __rust_alloc_zeroed@GOTPCREL]
```

</details>

changelog: [`slow_vector_initialization`]: clarify why `Vec::new()` + resize is worse than `vec![0; len]`

### DIFF
--- a/clippy_lints/src/slow_vector_initialization.rs
+++ b/clippy_lints/src/slow_vector_initialization.rs
@@ -20,7 +20,7 @@ declare_clippy_lint! {
     /// These structures are non-idiomatic and less efficient than simply using
     /// `vec![0; len]`.
     ///
-    /// More specifically, for `vec![0; len]`, the compiler can use a more specialized type of allocation
+    /// Specifically, for `vec![0; len]`, the compiler can use a specialized type of allocation
     /// that also zero-initializes the allocated memory in the same call
     /// (see: [alloc_zeroed](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html#method.alloc_zeroed)).
     ///
--- a/clippy_lints/src/slow_vector_initialization.rs
+++ b/clippy_lints/src/slow_vector_initialization.rs
@@ -20,6 +20,20 @@ declare_clippy_lint! {
     /// These structures are non-idiomatic and less efficient than simply using
     /// `vec![0; len]`.
     ///
+    /// More specifically, for `vec![0; len]`, the compiler can use a more specialized type of allocation
+    /// that also zero-initializes the allocated memory in the same call
+    /// (see: [alloc_zeroed](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html#method.alloc_zeroed)).
+    ///
+    /// Writing `Vec::new()` followed by `vec.resize(len, 0)` is suboptimal because,
+    /// while it does do the same number of allocations,
+    /// it involves two operations for allocating and initializing.
+    /// The `resize` call first allocates memory (since `Vec::new()` did not), and only *then* zero-initializes it.
+    ///
+    /// Writing `Vec::with_capacity(size)` followed by `vec.resize(len, 0)` is similar.
+    /// The allocation shifts from `resize` to `with_capacity`,
+    /// but the zero-initialization still happens separately,
+    /// when it could be done in one call with `vec![0; len]` (`alloc_zeroed`).
+    ///
     /// ### Example
     /// ```rust
     /// # use core::iter::repeat;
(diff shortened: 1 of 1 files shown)


clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -20,6 +20,20 @@ declare_clippy_lint! {
     /// These structures are non-idiomatic and less efficient than simply using
     /// `vec![0; len]`.
     ///
+    /// More specifically, for `vec![0; len]`, the compiler can use a more specialized type of allocation
+    /// that also zero-initializes the allocated memory in the same call
+    /// (see: [alloc_zeroed](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html#method.alloc_zeroed)).
```
djc, line 25: Just nitpicking some language, hope that's okay!

```suggestion
    /// Specifically, for `vec![0; len]`, the compiler can use a specialized type of allocation
    /// that also zero-initializes the allocated memory in the same call
    /// (see: [alloc_zeroed](https://doc.rust-lang.org/stable/std/alloc/trait.GlobalAlloc.html#method.alloc_zeroed)).
```

clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -20,6 +20,20 @@ declare_clippy_lint! {
...
+    ///
+    /// Writing `Vec::new()` followed by `vec.resize(len, 0)` is suboptimal because,
+    /// while it does do the same number of allocations,
+    /// it involves two operations for allocating and initializing.
+    /// The `resize` call first allocates memory (since `Vec::new()` did not), and only *then* zero-initializes it.
+    ///
+    /// Writing `Vec::with_capacity(size)` followed by `vec.resize(len, 0)` is similar.
+    /// The allocation shifts from `resize` to `with_capacity`,
+    /// but the zero-initialization still happens separately,
+    /// when it could be done in one call with `vec![0; len]` (`alloc_zeroed`).
```
djc, line 35: I feel like these paragraphs don't add as much value, I would probably cut them.

vi:
> @vi what do you think about this?

It makes it clear that advantage is because of zero value. I though that zero values matter for `.bss` vs `.data`, but allocation optimisations based on zero values are maybe less typical yet.

Is the lint inhibited for `Vec::new() + resize(len, 1)`, i.e. when it is statically known that the value is not zero, so vector initialisation needs to be an explicit step anyway? If not, the help message may be edited.

clippy_lints/src/slow_vector_initialization.rs
```diff
@@ -32,13 +46,17 @@ declare_clippy_lint! {
     ///
     /// let mut vec2 = Vec::with_capacity(len);
     /// vec2.extend(repeat(0).take(len));
+    ///
+    /// let mut vec3 = Vec::new();
```
djc, line 50: I would make this the second (maybe even the first?) example (and maybe cut the `vec1` example, since it seems fairly obviously similar to the `vec0` example). In contrast, this last example is already more effcient than the other ones because it only allocates once.

djc (approved):
This is great, thanks!

clippy_lints/src/slow_vector_initialization.rs
y21: Updated the examples, is this what you had in mind?

y21:
> Is the lint inhibited for Vec::new() + resize(len, 1), i.e. when it is statically known that the value is not zero, so vector initialisation needs to be an explicit step anyway? If not, the help message may be edited.

It only lints if the value argument in `.resize(len, val)` (or `.extend(repeat(val).take(len))` is the integer literal "0", so yep it wouldn't lint for `.resize(len, 1)`

djc:
Yup, looks great to me!

xFrednet (approved):

Issue #11290, opened 2023-08-03
Spurious "redundant redefinition of a binding" error from redundant_locals


emarsden:
### Summary

Nightly clippy is showing what I believe to be a spurious "redundant redefinition of a binding" error for lexically scoped mutable bindings.

### Lint Name

redundant_locals

### Reproducer

I tried this code:

```rust
fn do_shadow_let() {
    let bazzle = 42;
    {
        let bazzle = 69;
        println!("shadow_let: internal bazzle: {bazzle}");
    }
    println!("shadow_let: external bazzle: {bazzle}");
}

fn do_shadow_letmut() {
    let mut bazzle = 0;
    bazzle += 42;
    {
        let mut bazzle = bazzle;
        bazzle = 69;
        println!("shadow_letmut: internal bazzle: {bazzle}");
    }
    println!("shadow_letmut: external bazzle: {bazzle}");
}

fn main () {
    do_shadow_let();
    do_shadow_letmut();
}

```

I saw this happen:

```
error: redundant redefinition of a binding
  --> src/main.rs:14:9
   |
14 |     let mut bazzle = 0;
   |         ^^^^^^^^^^
...
17 |         let mut bazzle = bazzle;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: remove the redefinition of `bazzle`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_locals
   = note: `#[deny(clippy::redundant_locals)]` on by default

```

I expected to see no error for the shadowed mutable binding case, as for the shadowed immutable binding. 

### Version

```text
rustc 1.73.0-nightly (8131b9774 2023-08-02)
binary: rustc
commit-hash: 8131b9774ebcb6c162fcac71545a13543ec369e7
commit-date: 2023-08-02
host: x86_64-unknown-linux-gnu
release: 1.73.0-nightly
LLVM version: 16.0.5
```

Centri3:
We'll need to be more careful if they're in differing blocks, by making sure they're never mutated or explicitly used.

max-niederman:
Well... this is embarassing.

I think the best way to handle this would be to just not emit `redundant_locals` on mutable bindings in different blocks,
since the case where the binding is needlessly mutable is already covered by rustc's `unused_mut`.

Centri3:
That won't catch it if it's immutable and dropped, though. It'll still lint in those cases

Centri3:
[...]

PR #11316, opened 2023-08-10
Rustup

flip1995:
r? @ghost

cc @max-niederman With the latest sync, I'm getting a lot of FP in the `redundant_locals` lint you recently added. Any ideas where this could come from?

changelog: none


max-niederman:
> r? @ghost
> 
> cc @max-niederman With the latest sync, I'm getting a lot of FP in the `redundant_locals` lint you recently added. Any ideas where this could come from?
> 
> changelog: none

Excuse my ignorance. What is FP in this context?

matthiaskrgr:
FP / FN is usually False Positive and False Negative in clippy context

Centri3:
I think it's #11290 in particular. Maybe there are other issues, @flip1995?

Basically, for shadowing a local in a new scope, we need to ensure that it's never explicitly dropped or mutated within that block to lint it. We can still lint other cases fine.

I will work on this soon, or you can if you want.

flip1995:
I won't get to this over the weekend :neutral_face:

Centri3:
Looks like async functions are expanded to have a redundant redefinition now: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=3cfab52003203ea576c2ebdf97ba6843

Maybe there's a `from_expansion`-like function for *compiler generated*-expansions, regardless of whether it's external or not?

flip1995:
Thanks for investigating. I managed to fix it with your suggested one-liner :heart:

PR #11319, opened 2023-08-11
[iter_overeager_cloned]: detect .cloned().map() and .cloned().for_each()

lengyijun:
changelog: [`iter_overeager_cloned`]

key idea:
for `f` in `.map(f)` and `.for_each(f)`:
1. `f` must be closure
2. don't lint if mutable paramter in clsure `f`: `|mut x| ...`
3. don't lint if parameter is moved
4. maybe incorrect

### DIFF
--- a/clippy_lints/src/methods/iter_overeager_cloned.rs
+++ b/clippy_lints/src/methods/iter_overeager_cloned.rs
@@ -1,20 +1,28 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::source::snippet_opt;
 use clippy_utils::ty::{implements_trait, is_copy};
+use rustc_ast::BindingAnnotation;
 use rustc_errors::Applicability;
-use rustc_hir::{Expr, ExprKind};
+use rustc_hir::{Body, Expr, ExprKind, HirId, HirIdSet, PatKind};
+use rustc_hir_typeck::expr_use_visitor::{Delegate, ExprUseVisitor, PlaceBase, PlaceWithHirId};
 use rustc_lint::LateContext;
-use rustc_middle::ty;
+use rustc_middle::mir::{FakeReadCause, Mutability};
+use rustc_middle::ty::{self, BorrowKind};
 use rustc_span::sym;
 
 use super::ITER_OVEREAGER_CLONED;
 use crate::redundant_clone::REDUNDANT_CLONE;
+use crate::rustc_trait_selection::infer::TyCtxtInferExt;
 
 #[derive(Clone, Copy)]
 pub(super) enum Op<'a> {
     // rm `.cloned()`
     // e.g. `count`
     RmCloned,
+
+    // rm `.cloned()`
+    // e.g. `map` `for_each`
+    NeedlessMove(&'a str, &'a Expr<'a>),
 
     // later `.cloned()`
     // and add `&` to the parameter of closure parameter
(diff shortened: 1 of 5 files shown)


clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -35,10 +58,33 @@ pub(super) fn check<'tcx>(
...
+        if let Op::NeedlessMove(_, expr) = op {
+            if let rustc_hir::ExprKind::Closure(closure) = expr.kind
+               && let Body {params: [p], value: _, generator_kind: _ } = cx.tcx.hir().body(closure.body) {
+                    match p.pat.kind {
+                        rustc_hir::PatKind::Binding(BindingAnnotation(_, Mutability::Not), _, _, _) => {
+                            let body = cx.tcx.optimized_mir(closure.def_id);
+                            if let VarDebugInfoContents::Place(p) = &body.var_debug_info[0].value {
+                                let mut find_move = FindMove::new(p.local);
+                                find_move.visit_body(body);
+                                if find_move.found {
```
Centri3, line 70: If I'm understanding this correctly, this is just checking if it returns something, right? Since otherwise, if it doesn't return a `()`, it'll always have a move/copy, no matter what. If so, then why aren't you checking the return type of the closure?

If you aren't, then I'm really *really* confused on what this does. Maybe you meant to skip `Copy`s? (If so, you probably want to use `ExprUseVisitor` and overwrite `consume` instead of `optimized_mir`.)

lengyijun: I want to check whether the parameter of closure is `move`d in body, `_2`.
I'm not sure a better way to do that.

If it's `Copy`, it think should be checked in line 52:
```
&& matches!(*iter_assoc_ty.kind(), ty::Ref(_, ty, _) if !is_copy(cx, ty))
```

Centri3: > I want to check whether the parameter of closure is moved in body, _2.

Wow, I completely missed this is `var_debug_info`...

This, however, doesn't always work. For example, `_ = s` generates absolutely zero code, no matter what. You're basically relying on MIR not improving, ever, and if it does you're screwed.

MIR is generally not the way to do this. It loses a lot of information and fluff that the HIR has, which is necessary for linting.

> I'm not sure a better way to do that.

Why aren't you using a single `for_each_expr`? You could just check if there are any paths referencing the local.

lengyijun: I'm afraid `for_each_expr` can't detect `move` .
In hir, we can't detect whether `x.len()` use `&self` or `Self` ?

Centri3: > In hir

We can, using the `ExprUseVisitor` and overwriting `consume`.

`&self` and `Self` are both `move` so in MIR, afaik, you're linting both. For `mir-opt-level` > 0, it's lowered to a `Len`, but above that it's a call and a `move`.

clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -51,8 +59,46 @@ pub(super) fn check<'tcx>(
             return;
         }
 
+        if let Op::NeedlessMove(_, expr) = op {
+            let rustc_hir::ExprKind::Closure(closure) = expr.kind else { return } ;
+            let body@ Body {params: [p], .. } = cx.tcx.hir().body(closure.body) else { return };
```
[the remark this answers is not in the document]
lengyijun: `map` and `for_each` require a closure of only one param

tests/ui/iter_overeager_cloned.stderr
```diff
@@ -130,5 +130,21 @@ LL |             iter.cloned().filter(move |S { a, b }| **a == 1 && b == &target
...
+LL |     let _ = vec.iter().cloned().map(|x| x.len());
+   |             ^^^^^^^^^^--------------------------
+   |                       |
+   |                       help: try: `.map(|x| x.len())`
+
+error: unneeded cloning of iterator items
+  --> $DIR/iter_overeager_cloned.rs:65:13
+   |
+LL |     let _ = vec.iter().cloned().for_each(|x| assert!(!x.is_empty()));
+   |             ^^^^^^^^^^----------------------------------------------
```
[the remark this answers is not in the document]
lengyijun: Maybe we need to fix the origin .stderr first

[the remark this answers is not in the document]
blyxyas: Yep, just checked, it is a lint-scoped problem. It's problematic if the user isn't using a tool like `rustfix` to apply the suggestions for them. Could you fix it or do you think it is out of scope for this PR?

clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -51,8 +59,46 @@ pub(super) fn check<'tcx>(
...
+            ExprUseVisitor::new(
+                &mut delegate,
+                &infcx,
+                closure.body.hir_id.owner.def_id,
+                cx.param_env,
+                cx.typeck_results(),
+            )
+            .consume_body(body);
+
+            let mut found = false;
```
blyxyas, line 77: Could this have a better variable name? Maybe something like `to_be_discarded` or similar?

clippy_lints/src/methods/iter_overeager_cloned.rs
```diff
@@ -83,8 +129,33 @@ pub(super) fn check<'tcx>(
...
+        if let PlaceBase::Local(l) = place_with_id.place.base {
+            self.used_move.insert(l);
+        }
+    }
+
+    fn borrow(&mut self, _: &PlaceWithHirId<'tcx>, _: HirId, _: BorrowKind) {}
+
+    fn mutate(&mut self, _: &PlaceWithHirId<'tcx>, _: HirId) {}
+
+    fn fake_read(&mut self, _: &PlaceWithHirId<'tcx>, _: FakeReadCause, _: HirId) {}
```
blyxyas, line 160: Could you put some `unreachable!` in these methods to signify that they won't be used, and shouldn't be used?

Centri3: They'll still be ran anyway by `ExprUseVisitor`

tests/ui/iter_overeager_cloned.stderr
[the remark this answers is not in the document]
lengyijun: A quick review, some other files suffer same problems:

tests/ui/float_equality_without_abs.stderr
tests/ui/unnecessary_lazy_eval_unfixable.stderr
tests/ui/unnecessary_lazy_eval.stderr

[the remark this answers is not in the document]
blyxyas: After some reflexing on this discussion, I just realized this is total bikeshedding. It isn't worth pausing a lint just because of a stylistic error. I'll discuss this stylistic choice in the Zulip chat on the next meeting and decide from that. This discussion can be resolved.

blyxyas (commented):
Just that mini-nit of changing the variable name and I think this is ready, thanks :heart:!

blyxyas (approved):

PR #11333, opened 2023-08-13
Do not bless by default in ui tests

Alexendoo:
This restores the default behaviour to check the `.stderr`, it was changed in #11239 to bless by default in `cargo test` (unless in github actions), but check by default in `cargo uitest` which is fairly confusing

It also meant `cargo uitest -F internal` no longer worked

`--bless` prevents the use of `Args::test` but we can look at reintegrating with that after @oli-obk's vacation

r? @flip1995

changelog: none

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,7 +1,7 @@
 [alias]
-uitest = "test --test compile-test -- --check"
-uibless = "test --test compile-test"
-bless = "test"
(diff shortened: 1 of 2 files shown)


flip1995:
@bors r+

Thanks! I also was surprised by the bless-by-default change, but didn't want to block the bitrotty PR on that. Agreed that the check-by-default behavior feels more natural.

PR #11336, opened 2023-08-14
Use ui_test's Windows path backslash heuristic

Alexendoo:
changelog: none

Instead of unconditionally replacing `\` with `/` we now use [`Match::PathBackslash`](https://docs.rs/ui_test/latest/ui_test/enum.Match.html#variant.PathBackslash) to only replace backslashes in paths that look like windows paths

`ui-toml` and `ui-cargo` tests still use the old way because they produce verbatim paths on windows in some tests (`\\?\C:\foo\...`) which was finnicky to get the replacement order correct with

Also removes the `ui_test` -> `compiletest` alias and `VarGuard`


flip1995 (approved):
Cool stuff! Thanks!

PR #11338, opened 2023-08-14
allow trait alias DefIds in `implements_trait_with_env_from_iter`

### LINKED ISSUE #11337: ICE: debug assertions: trait alias
### Summary

need clippy built with debug assertions
rustcs `./tests/ui/closures/self-supertrait-bounds.rs`

```rust
// check-pass

// Makes sure that we only consider `Self` supertrait predicates while
// elaborating during closure signature deduction.

#![feature(trait_alias)]

trait Confusing<F> = Fn(i32) where F: Fn(u32);

fn alias<T: Confusing<F>, F>(_: T, _: F) {}

fn main() {
    alias(|_| {}, |_| {});
}
```

### Version

```text
rustc @ 3071e0aef6dfd0a150c3fb1da0abad4ec86ca0aa
```

### Error output

  <p>

  ```
  <backtrace>
  ```

  </p>
</details>

y21:
Fixes #11337

changelog: none

### DIFF
--- a/clippy_utils/src/ty.rs
+++ b/clippy_utils/src/ty.rs
@@ -27,6 +27,7 @@ use rustc_target::abi::{Size, VariantIdx};
 use rustc_trait_selection::traits::query::evaluate_obligation::InferCtxtExt as _;
 use rustc_trait_selection::traits::query::normalize::QueryNormalizeExt;
 use rustc_trait_selection::traits::{Obligation, ObligationCause};
+use std::assert_matches::debug_assert_matches;
 use std::iter;
 
 use crate::{match_def_path, path_res, paths};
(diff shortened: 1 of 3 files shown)


clippy_utils/src/ty.rs
```diff
@@ -259,7 +260,11 @@ pub fn implements_trait_with_env_from_iter<'tcx>(
             })),
     );
 
-    debug_assert_eq!(tcx.def_kind(trait_id), DefKind::Trait);
+    debug_assert_matches!(
+        tcx.def_kind(trait_id),
+        DefKind::Trait | DefKind::TraitAlias,
+        "`DefId` must belong to a trait or trait alias"
+    );
```
y21, line 267: we could also just use `debug_assert!(matches!(...));` and avoid the feature, but `debug_assert_matches!` has a better ICE message (it'll print the given DefKind)

Centri3: From the looks of it, `assert_matches` is on track to getting stabilized so there isn't much of a point in doing so. I'd say keep it. The path will change though in the future, but it shouldn't cause issues (it'll be fixed within clippy in the PR that changes it)

PR #11340, opened 2023-08-15
New lint `needless_cow`

### LINKED ISSUE #11328: Lint for needless use of `Cow`
### What it does

When a `Cow`'s `Borrowed` variant is only ever used for empty strings, clippy should suggest to remove the `Cow` and use `String::new()` instead.

### Advantage

- `String::new()` allocates no memory, and is essentially the same speed: https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=93a8649afa1b55fc84cef2829920a932 (Less instructions doesn't mean faster, but imo, even if it's slower it's worth the ergonomics)
- It's simpler.

### Drawbacks

- It may actually end up being ever so slightly slower. In some cases this may be undesirable.

### Example

```rust
if let Some(x) = x {
    x.to_string().into()
} else if let Some(y) = y {
    Cow::Borrowed("")
} else {
    "".into()
}
```

Could be written as:

```rust
if let Some(x) = x {
    x.to_string()
} else if let Some(y) = y {
    String::new()
} else {
    String::new()
}
```

Centri3:
Closes #11328 

This is actually probably easiest as a MIR lint. Most of the problems with it currently would persist if it operated on the HIR anyway.
r? @Jarcho

Currently in `nursery` since there's quite a few issues.

changelog: New lint [`needless_cow`]

### DIFF
--- a/clippy_lints/src/needless_cow.rs
+++ b/clippy_lints/src/needless_cow.rs
@@ -1,0 +1,306 @@
+use clippy_utils::diagnostics::span_lint_hir_and_then;
+use clippy_utils::fn_has_unsatisfiable_preds;
+use clippy_utils::ty::is_type_diagnostic_item;
+use itertools::Itertools;
+use rustc_const_eval::interpret::{AllocId, AllocRange, ConstValue, GlobalAlloc};
+use rustc_data_structures::fx::{FxIndexMap, IndexEntry};
+use rustc_hir::def_id::LocalDefId;
+use rustc_hir::intravisit::FnKind;
+use rustc_hir::{Body, FnDecl};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::mir::{
+    self, AggregateKind, ConstantKind, Location, Operand, Place, Rvalue, SourceInfo, Statement, StatementKind,
+};
+use rustc_middle::ty::TyCtxt;
+use rustc_mir_dataflow::lattice::FlatSet;
+use rustc_mir_dataflow::value_analysis::{Map, State, ValueAnalysis, ValueAnalysisWrapper, ValueOrPlace};
+use rustc_mir_dataflow::{Analysis, Results, ResultsVisitor};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::{sym, Span};
+use rustc_target::abi::{FieldIdx, Size};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for locals who are assigned a `Cow`, yet the `Borrowed` variant is only ever used for
+    /// empty string slices.
+    ///
+    /// ### Why is this bad?
+    /// An empty string does not allocate any memory. This can use `String` instead, with the
+    /// `Borrowed` invocations substituted for `String::new`.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # let my_owned_string = String::new();
+    /// # let something = true;
+    /// # let other_thing = true;
+    /// let _ = if something {
+    ///     Cow::Owned(my_owned_string);
+    /// } else if other_thing {
+    ///     Cow::Owned(String::new());
+    /// } else {
+    ///     Cow::Borrowed("");
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # let my_owned_string = String::new();
+    /// # let something = true;
+    /// # let other_thing = true;
+    /// let _ = if something {
+    ///     my_owned_string;
+    /// } else if other_thing {
+    ///     String::new();
+    /// } else {
+    ///     String::new();
+    /// }
+    /// ```
+    #[clippy::version = "1.73.0"]
+    pub NEEDLESS_COW,
+    nursery,
+    "usage of `Cow` with only empty strings in `Borrow` variant"
+}
+impl_lint_pass!(NeedlessCow => [NEEDLESS_COW]);
+
+pub struct NeedlessCow;
+
+impl<'tcx> LateLintPass<'tcx> for NeedlessCow {
+    fn check_fn(
+        &mut self,
+        cx: &LateContext<'tcx>,
+        _: FnKind<'tcx>,
+        _: &'tcx FnDecl<'tcx>,
+        _: &'tcx Body<'tcx>,
+        _: Span,
+        def_id: LocalDefId,
+    ) {
+        // Probably unnecessary nowadays considering dummy MIR is created for these functions.
+        if fn_has_unsatisfiable_preds(cx, def_id.to_def_id()) {
+            return;
+        }
+
+        let mir = cx.tcx.optimized_mir(def_id);
+        let mut cow_locals_to_assignments = FxIndexMap::default();
+
+        for local in mir
+            .local_decls
+            .iter_enumerated()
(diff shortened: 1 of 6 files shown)


tests/ui/needless_cow.rs
```diff
@@ -0,0 +1,71 @@
...
+    let x = if true { Cow::Borrowed("") } else { Cow::Borrowed("") };
+    let mut y = x;
+
+    helper(&mut y);
+}
+
+// FIXME: Should not lint.
+fn returning() -> Cow<'static, str> {
+    if true { Cow::Borrowed("") } else { Cow::Borrowed("") }
+}
```
Centri3, line 34: We can probably do another analysis pass for these, but I'm unsure if that's the best way to fix this. The issue is that it doesn't *have* to refer to the local initialized by `Borrowed`, it could be something like `_0 = _2` where `_2 = _3` which makes it a bit annoying. There's probably an analysis pass that gives you which locals it "depends" on

Jarcho:
Just some initial comments.

* Since the only value you care about is the empty string you can shrink `FlatSet` to only storing a single value. This would represent uninitialized, empty string, and any other value.
* You can track `Cow<str>` values in the same pass that you track `&str` values. You would need to treat `Cow::Owned` as equivalent to the uninitialized state. The result can be extracted by joining the state of all the terminating blocks.

Centri3:
> * Since the only value you care about is the empty string you can shrink `FlatSet` to only storing a single value. This would represent uninitialized, empty string, and any other value.

Do you mean removing the `FlatSet`, using something like a bool (but whose implementations of `HasTop` and `HasBottom` are the `false` variant), or using something like `FlatSet<()>`?

Jarcho:
`FlatSet<()>` can represent the needed state.

clippy_lints/src/needless_cow.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn check_fn(
+        &mut self,
+        cx: &LateContext<'tcx>,
+        _: FnKind<'tcx>,
+        _: &'tcx FnDecl<'tcx>,
+        _: &'tcx Body<'tcx>,
+        _: Span,
+        def_id: LocalDefId,
+    ) {
+        // Probably unnecessary nowadays considering dummy MIR is created for these functions.
```
Centri3, line 87: I'd like to know whether this is the case. This is the code I was referencing: https://github.com/rust-lang/rust/blob/d7e751006cb3691d1384b74196a9cb45447acfa8/compiler/rustc_mir_transform/src/lib.rs#L381-L423

clippy_lints/src/needless_cow.rs
```diff
@@ -0,0 +1,384 @@
...
+                continue;
+            };
+
+            span_lint_hir_and_then(
+                cx,
+                NEEDLESS_COW,
+                mir.source_scopes[first.source_info().scope]
+                    .local_data
+                    .as_ref()
+                    .assert_crate_local()
```
Centri3, line 133: This call feels suspicious. Isn't `ClearCrossCrate` meant to prevent usage from other crates (In this case, I suppose clippy)? For `local_info` this would ICE. Was this not cleared so that clippy can access this?

clippy_lints/src/needless_cow.rs
```diff
@@ -0,0 +1,384 @@
...
+        let Some(target_idx) = self.map.find(target.as_ref()) else {
+            return;
+        };
+
+        let result = match rvalue {
+            Rvalue::Use(Operand::Constant(box constant)) => ValueOrPlace::Value(FlatSet::Elem(constant.literal)),
+            Rvalue::Use(Operand::Move(place) | Operand::Copy(place))
+            | Rvalue::Ref(_, _, place)
+            | Rvalue::CopyForDeref(place) => {
+                if (place.projection.is_empty() || place.is_indirect_first_projection() && place.projection.len() == 1)
```
Centri3, line 191: We can change this to always clearing projections, as all projections are either nops or do nothing for `str`:

- `Deref`: `str` doesn't implement `Deref` and can't by external code;
- `Field`: `str` has no fields since it's a primitive;
- `Index`: Indexing an empty `str` will always panic;
- `ConstantIndex`: Same as above;
- `Subslice`: Same as above;
- `Downcast`: `str` is not an enum or generator;
- `OpaqueCast`: Since we only track `str` it must already be a concrete type;

However, I'm not sure if this is what we *should* do or if it's correct. Am I missing something that allows us to track the value for these places easier?

clippy_lints/src/needless_cow.rs
```diff
@@ -0,0 +1,384 @@
...
+        state.flood(target.as_ref(), &self.map);
+
+        let Some(target_idx) = self.map.find(target.as_ref()) else {
+            return;
+        };
+
+        let result = match rvalue {
+            Rvalue::Use(Operand::Constant(box constant)) => ValueOrPlace::Value(FlatSet::Elem(constant.literal)),
+            Rvalue::Use(Operand::Move(place) | Operand::Copy(place))
+            | Rvalue::Ref(_, _, place)
```
Centri3, line 189: Value analysis doesn't track these by default, however, for constant empty strings there's always a reference. This seems to work, though I'd like clarification that this is what we should do.

clippy_lints/src/needless_cow.rs
```diff
@@ -0,0 +1,384 @@
...
+                },
+            );
+        }
+    }
+}
+
+type CowAssignments = Vec<CowAssignment>;
+
+#[derive(Clone, Copy)]
+enum CowAssignment {
```
Centri3, line 152: To properly handle `async` correctly, we need another variant that's for closures. If this closure isn't used in a larger `if`/`match`/etc., we would handle it as if it's its own separate "thing", but if it's part of a larger structure, we'll handle assignments within it as if there's no closure at all. Basically:

```rust
if true { (|| Cow::Borrowed(""))() } else { Cow::Borrowed("") }
```

Should lint, but:

```rust
(|| Cow::Borrowed(""))();
```

Should not, since the closure isn't declared and called within an `if`/`match`/etc.

Also, if the closure contains its own `if` expression, we'd handle that too.

clippy_lints/src/needless_cow.rs
```diff
@@ -0,0 +1,384 @@
...
+                }) else {
+                    return;
+                };
+                let val = val.eval(cx.tcx, cx.param_env);
+
+                if is_empty_str(cx.tcx, val) {
+                    entry.get_mut().push(CowAssignment::Borrowed(stmt.source_info));
+                } else {
+                    entry.remove_entry();
+                }
```
Centri3, line 271: Should be extracted into its own function

Centri3 (commented):
I've left some comments for things I'd either like clarification on or which we need to do for this to be warn-by-default.

xFrednet:

Centri3:
I think this one, despite being a MIR lint, is pretty solid (if I do say so myself). If anyone who knows MIR wants to pick this up, it shouldn't take *too* much effort to resolve my previous comments and make it ready

Could also rewrite this to use the HIR, but this means you have to parse destructuring (so `let (a, _) = (Cow::Borrowed(""), ())` would be handled) and also `if` statements and such. This would be well appreciated tbh as other lints need this too (see `local_assigned_single_value`), but it's quite a big undertaking

PR #11350, opened 2023-08-18
[`type_id_on_box`]: lint on any `Box<dyn _>`

### LINKED ISSUE #11349: `type_id_on_box` for supertraits of Any
### What it does

Detect `.type_id()` calls on `Box<dyn Trait>` when `trait Trait: Any {...}`

### Advantage

Less annoying-to-find mistakes on Any traits

### Drawbacks

_No response_

### Example

```rust
use std::any::Any;

trait Test: Any {}

impl Test for i32 {}

fn main() {
    let b: Box<dyn Test> = Box::new(0);
    let ty = b.type_id();
    dbg!(ty);
}
```
[playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=708cd57949ee5a693d72d76b96509ee8)

y21:
Closes #11349.

It now not only lints when calling `.type_id()` on the type `Box<dyn Any>`, but also on any `Box<dyn Trait>` where `Trait` is a subtrait of `Any`

changelog: FN: [`type_id_on_box`]: lint if `Any` is a sub trait

### DIFF
--- a/tests/ui/type_id_on_box.fixed
+++ b/tests/ui/type_id_on_box.fixed
@@ -22,41 +22,35 @@ fn existential() -> impl Any {
 trait AnySubTrait: Any {}
 impl<T: Any> AnySubTrait for T {}
 
-// `Any` is an indirect supertrait
-trait AnySubSubTrait: AnySubTrait {}
-impl<T: AnySubTrait> AnySubSubTrait for T {}
-
-// This trait mentions `Any` in its predicates, but it is not a subtrait of `Any`.
-trait NormalTrait
-where
-    i32: Any,
-{
-}
-impl<T> NormalTrait for T {}
-
 fn main() {
+    // Don't lint, calling `.type_id()` on a `&dyn Any` does the expected thing
+    let ref_dyn: &dyn Any = &42;
+    let _ = ref_dyn.type_id();
+
     let any_box: Box<dyn Any> = Box::new(0usize);
     let _ = (*any_box).type_id();
-    let _ = TypeId::of::<Box<dyn Any>>(); // Don't lint. We explicitly say "do this instead" if this is intentional
+    //~^ ERROR: calling `.type_id()` on
+
+    // Don't lint. We explicitly say "do this instead" if this is intentional
+    let _ = TypeId::of::<Box<dyn Any>>();
     let _ = (*any_box).type_id();
+
+    // 2 derefs are needed here to get to the `dyn Any`
     let any_box: &Box<dyn Any> = &(Box::new(0usize) as Box<dyn Any>);
-    let _ = (**any_box).type_id(); // 2 derefs are needed here to get to the `dyn Any`
+    let _ = (**any_box).type_id();
+    //~^ ERROR: calling `.type_id()` on
 
     let b = existential();
-    let _ = b.type_id(); // Don't lint.
+    let _ = b.type_id(); // Don't
+
+    let b: Box<dyn AnySubTrait> = Box::new(1);
+    let _ = (*b).type_id();
+    //~^ ERROR: calling `.type_id()` on
 
     let b: SomeBox = Box::new(0usize);
     let _ = (*b).type_id();
+    //~^ ERROR: calling `.type_id()` on
 
     let b = BadBox(Box::new(0usize));
     let _ = b.type_id(); // Don't lint. This is a call to `<BadBox as Any>::type_id`. Not `std::boxed::Box`!
-
-    let b: Box<dyn AnySubTrait> = Box::new(1);
-    let _ = (*b).type_id(); // Lint if calling `type_id` on a `dyn Trait` where `Trait: Any`
-
-    let b: Box<dyn AnySubSubTrait> = Box::new(1);
-    let _ = b.type_id(); // Known FN - Any is not an "immediate" supertrait
-
-    let b: Box<dyn NormalTrait> = Box::new(1);
-    let _ = b.type_id(); // `NormalTrait` does not have `Any` as its supertrait (even though it mentions it in `i32: Any`)
 }
--- a/tests/ui/type_id_on_box.fixed
+++ b/tests/ui/type_id_on_box.fixed
@@ -19,6 +19,21 @@ fn existential() -> impl Any {
     Box::new(1) as Box<dyn Any>
 }
 
+trait AnySubTrait: Any {}
+impl<T: Any> AnySubTrait for T {}
+
+// `Any` is an indirect supertrait
+trait AnySubSubTrait: AnySubTrait {}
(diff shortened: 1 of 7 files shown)


xFrednet:
Let's see if we can revive this PR. @y21 would you mind rebasing on master?

r? @xFrednet

tests/ui/type_id_on_box.fixed
```diff
@@ -19,6 +19,9 @@ fn existential() -> impl Any {
     Box::new(1) as Box<dyn Any>
 }
 
+trait AnySubTrait: Any {}
+impl<T: Any> AnySubTrait for T {}
```
xFrednet, line 23: Can you add a test, where the `Any` trait is down one or two super traits?

I'm also wondering if a combination of generics in the trait declaration can somehow mess this up, but I don't think so.

xFrednet (commented):
The implementation looks good to me. I have one small comment and then it should be good to go :)

tests/ui/type_id_on_box.fixed
y21: Looks like it currently doesn't lint in that case. Should we? I'm not sure if we'd need to be a bit more careful when recursively checking them to not stack overflow or something like that to avoid cycles, specially when there are generics involved like you said. But I also can't really come up with a case that could run into that (and not get rejected by the compiler first). Leaving it as a FN would be safe I think, but I could also try it if you want that implemented

xFrednet: Based on the implementation, I didn't expect it to work. I mainly wanted to confirm my suspicion. Having a test that just confirms this FN with a command seems all right to me. 

Now that I'm thinking about it, is there a reason, why we can't *simply* check if the inner type implements the `Any` trait? (I haven't looked at the implementation for this comment, so sorry if I missed something obvious)

y21: What's the inner type? The type that the user is calling `.type_id()` on? I don't think it would work. `Any` is also implemented `&dyn Any`. So I think we would also lint on the following code, which doesn't have the issue this lint looks for
```rs
let x: &dyn Any = &1;
assert_eq!(x.type_id(), TypeId::of::<i32>()); // works as expected
```

It's more of an issue with method resolution.

I think this lint has a bit of an unfortunate name. As far as I understand it, this happens when calling `.type_id()` on *anything* that isn't `&dyn Any`.
`.type_id()` on  `Rc<dyn Any>` has the same issue. It might make sense to just rename this lint in a followup?

xFrednet: Yeah, the name is a bit unfortunate, renaming it in a followup PR sounds good to me, you can `r? me` if you want.

> What's the inner type?

I meant the generic type. So first look for `Box<&dyn SomethingTrait>` and then get the `&dyn SomethingTrait` from the generics of `Box` and check if `&dyn SomethingTrait` implements `Any`

That shouldn't lint the code example you gave in your comment, but correctly detect these lint cases right?

y21: Oh I see what you mean now. Well, we would lint on `Box<i32>` then, but I guess we can first check if its also `dyn`. Then that sounds like it could work

y21: (I assume you meant `Box<dyn>` not `Box<&dyn>`, since the latter isn't really useful to have anyway)

Now implementing it, it does however mean that we would lint on this test case I added:

https://github.com/rust-lang/rust-clippy/blob/37be3e4dd5ec4aa11959f9fd24d1c1dc53da44bd/tests/ui/type_id_on_box.rs#L60-L61

Since `dyn NormalTrait: Any` (since it is implemented for anything), but it's not a subtrait. I suppose that might still be worth linting even if it isn't a subtrait. I don't really have strong feelings about it, we could do it like that and forget about subtraits.

xFrednet: > (I assume you meant Box<dyn> not Box<&dyn>, since the latter isn't really useful to have anyway)

Your assumption is correct.

Hmm, I sadly haven't used `Any` before, so I'm not sure what the correct behavior should be. But if `Any` is implemented for anything, it feels weird to restrict it only to lint `dyn Trait` things, which have it explicitly as a super trait.

Do we maybe want to ask on Zulip if someone has experience with `Any`? Otherwise, I think it's okay to merge the current version. Since this is an warn-by-default lint it's better to have FNs over FPs

y21: I also did a lintcheck run on the top 200 crates and it found one match in ahash:

https://docs.rs/ahash/latest/src/ahash/random_state.rs.html#192

As far as I can tell, I *think* (though I could be wrong, hard to tell in the docs.rs web view) the lint is correct that this is suspicious, that `.type_id() != TypeId::of::<&DefaultRandomSource>()` should always return `true`.

Given this, I would say it makes sense to do it how you suggested, but I can also ask on zulip

y21: Ah, one thing that I missed is that we *do* still need to check if `Any` is a supertrait for the suggestion that we are making.

For some random `Box<dyn Trait>` (where `dyn Trait` implements `Any`), the suggestion to dereference it first and then call `type_id` is not valid (as in, doesn't do what you would expect). It only works when `Any` is explicitly a supertrait because only then is it also put in the vtable.

That is, this is not valid 
```rs
trait Trait {} // Not `: Any`
impl Trait for i32 {}

let b = Box::new(42i32) as Box<dyn Trait>;
assert_eq!((*b).type_id(), TypeId::of::<i32>()); // fails
```
but when  `: Any` is added to the trait definition the assertion passes.

-----

So I think for this we can use the existing function that is in the PR currently but only use it for adding a suggestion. We could still lint w/o a suggestion if that function has false negatives (like when `Any` is not directly a supertrait). Does that make sense to you?

y21: Alright, done. As mentioned on Zulip, checking with `implements_trait` is redundant, since any `dyn Trait` automatically implements `Any` because of its blanket impl, so it would always be true. So we now simply lint when calling `.type_id()` on any `Box<dyn Trait>`. I also updated the lint documentation to to reflect this.

And as mentioned, the suggestion to dereference the box only makes sense when the `Any` supertrait exists (for example in ahash's case it doesn't), so we still use the old logic for that

xFrednet (approved):

xFrednet:
This version looks good to me, thank you for also updating the lint description. I like these changes. It looks like the suggestion good some support on Zulip, if something comes up, we can just create a followup.

Thank you!

@bors r+

y21:
> It looks like the suggestion good some support on Zulip, if something comes up, we can just create a followup.

I cant quite parse this sentence, can you rephrase that ? "Could have some support"?

PR #11358, opened 2023-08-19
Rename incorrect_impls to non_canonical_impls, move them to warn by default

Alexendoo:
The wording/category of these feel too strong to me, I would expect most of the time it's linting the implementations aren't going to be *incorrect*, just unnecessary

changelog: rename `incorrect_clone_impl_on_copy_type` to [`non_canonical_clone_impl`]
changelog: rename `incorrect_partial_ord_impl_on_ord_type` to [`non_canonical_partial_ord_impl`]
changelog: Move [`non_canonical_clone_impl`], [`non_canonical_partial_ord_impl`] to suspicious


Jarcho:
Both of the lints are checking the function body for the most obvious implementation so the name and lint category are both justified.

Alexendoo:
(added a note mentioning that it's a requirement for `Clone`/`Copy` and `Ord`/`PartialOrd` to be consistent)

Not being the most obvious implementation doesn't make them incorrect, the [before example](https://rust-lang.github.io/rust-clippy/master/index.html#/incorrect_clone_impl_on_copy_type) in `incorrect_clone_impl_on_copy_type` is correct, just not in our canonical form

Jarcho:
The lint on `PartialOrd` really would be better as `suspicious`. Assuming the impl is correct now, it's still something which could drift in the future when something changes.

`manual_clone_impl_on_copy_type` also doesn't really make sense as a name when `expl_impl_clone_on_copy` is already a lint. `non_canonical_clone_impl` would be a better name for what's actually happening.

Alexendoo:
Whoops, apparently if you forget to commit your changes and push a commit equal to the target branch it'll close the PR for you

Jarcho:
Yay, more perplexing behaviour from github!

@bors r+

Jarcho:
Looks like you need to re-bless the tests. Should also change the error message which still mentions that it's an incorrect impl.

Alexendoo:
I do after rebasing 🤷

Jarcho:
One more time. @bors r+

sfackler:
It looks like this missed the 1.73.0 release?

Alexendoo:
Yeah, it's still `incorrect_...` on 1.73

PR #11359, opened 2023-08-19
Check that the suggested method exists in unwrap_or_default

### LINKED ISSUE #11355: Missing Entry::or_default on LinearMap
### Summary

Lint [unwrap_or_default](https://rust-lang.github.io/rust-clippy/master/index.html#/unwrap_or_default) suggests use of `or_default` with [`LinearMap::Entry`](https://docs.rs/linear-map/latest/linear_map/enum.Entry.html) which does not support this method.

It may be easier to add the method to `linear-map`, but the crate is in maintenance-only mode.

### Lint Name

unwrap_or_default

### Reproducer

```rust
let mut map = LinaerMap::new();
map.entry("x").or_insert_with(Default::default);
```

### Version

_No response_

### Additional Labels

_No response_

Alexendoo:
Fixes #11355

changelog: none

### DIFF
--- a/clippy_lints/src/methods/or_fun_call.rs
+++ b/clippy_lints/src/methods/or_fun_call.rs
@@ -65,10 +65,25 @@ pub(super) fn check<'tcx>(
         };
 
         let sugg = match (name, call_expr.is_some()) {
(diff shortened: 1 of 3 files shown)


dswij (approved):

dswij:
Thanks for this! @bors r+

PR #11363, opened 2023-08-20
[`redundant_closure_call`]: avoid duplicated `async` keyword when triggering on closure that returns `async` block

### LINKED ISSUE #11357: redundant_closure_call: suggestion does not work if inner expression is async block
### Summary

Suggestion does not work if the inner expression is `async` block.

### Reproducer

I tried this code:

```rust
let x = (|| async {})().await;
```

```rust
let x = async {}.await;
```

Instead, this happened:

```rust
let x = async async {}.await;
```

### Version

```text
rustc 1.73.0-nightly (31395ec38 2023-07-24)
binary: rustc
commit-hash: 31395ec38250b60b380fd3c27e94756aba3557de
commit-date: 2023-07-24
release: 1.73.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

KisaragiEffective:
close #11357

----

changelog: [`redundant_closure_call`]: avoid duplicated `async` keyword when triggering on closure that returns `async` block

### DIFF
--- a/clippy_lints/src/redundant_closure_call.rs
+++ b/clippy_lints/src/redundant_closure_call.rs
@@ -5,7 +5,7 @@ use clippy_utils::sugg::Sugg;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
 use rustc_hir::intravisit::{Visitor as HirVisitor, Visitor};
-use rustc_hir::{intravisit as hir_visit, CoroutineKind, CoroutineSource};
+use rustc_hir::{intravisit as hir_visit, CoroutineKind, CoroutineSource, Node};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::nested_filter;
 use rustc_middle::lint::in_external_macro;
@@ -176,12 +176,19 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClosureCall {
                             hint = hint.asyncify();
                         }
 
-                        diag.span_suggestion(
-                            full_expr.span,
-                            "try doing something like",
-                            hint.maybe_par(),
-                            applicability,
-                        );
+                        let is_in_fn_call_arg = clippy_utils::get_parent_node(cx.tcx, expr.hir_id)
+                            .map(|x| match x {
+                                Node::Expr(expr) => matches!(expr.kind, hir::ExprKind::Call(_, _)),
+                                _ => false,
+                            })
+                            .unwrap_or(false);
+
+                        // avoid clippy::double_parens
+                        if !is_in_fn_call_arg {
+                            hint = hint.maybe_par()
+                        };
+
+                        diag.span_suggestion(full_expr.span, "try doing something like", hint, applicability);
                     }
                 },
             );
--- a/clippy_lints/src/redundant_closure_call.rs
+++ b/clippy_lints/src/redundant_closure_call.rs
@@ -4,8 +4,8 @@ use clippy_utils::get_parent_expr;
 use clippy_utils::sugg::Sugg;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
-use rustc_hir::intravisit as hir_visit;
 use rustc_hir::intravisit::{Visitor as HirVisitor, Visitor};
+use rustc_hir::{intravisit as hir_visit, CoroutineKind, CoroutineSource};
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::nested_filter;
 use rustc_middle::lint::in_external_macro;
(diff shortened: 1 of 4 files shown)


clippy_lints/src/redundant_closure_call.rs
```diff
@@ -62,9 +62,9 @@ impl<'tcx> Visitor<'tcx> for ReturnVisitor {
 
 /// Checks if the body is owned by an async closure
 fn is_async_closure(body: &hir::Body<'_>) -> bool {
-    if let hir::ExprKind::Closure(closure) = body.value.kind
-        && let [resume_ty] = closure.fn_decl.inputs
-        && let hir::TyKind::Path(hir::QPath::LangItem(hir::LangItem::ResumeTy, ..)) = resume_ty.kind
+    if let hir::ExprKind::Closure(_) = body.value.kind
+        && let Some(kind) = body.generator_kind
+        && kind == GeneratorKind::Async(AsyncGeneratorKind::Closure)
```
llogiq, line 67: Could we pull those statements together? As in

```suggestion
        && let Some(GeneratorKind::Async(AsyncGeneratorKind::Closure)) = body.generator_kind
```

llogiq (commented):
I'm not sure how to make it work right now. Let's see if the discussion on zulip brings a solution.

clippy_lints/src/redundant_closure_call.rs
```diff
@@ -61,10 +61,15 @@ impl<'tcx> Visitor<'tcx> for ReturnVisitor {
...
-        && let hir::TyKind::Path(hir::QPath::LangItem(hir::LangItem::ResumeTy, ..)) = resume_ty.kind
+fn is_async_closure(cx: &LateContext<'_>, body: &hir::Body<'_>) -> bool {
+    if let hir::ExprKind::Closure(desugared_outer_closure) = body.value.kind
+        && let desugared_outer_closure_body = cx.tcx.hir().body(desugared_outer_closure.body)
+        && let hir::ExprKind::Closure(desugared_inner_closure) = desugared_outer_closure_body.value.kind
+        // NB: `async || 42` is desugared into `|| |mut __context: #[lang = "ResumeTy"] | 42`; so
+        //     we need to check if the only input was that lang item.
+        && let [compiler_generated_resume_ty] = desugared_inner_closure.fn_decl.inputs
+        && let hir::TyKind::Path(hir::QPath::LangItem(hir::LangItem::ResumeTy, ..)) =
+            compiler_generated_resume_ty.kind
```
y21, line 72: This fixes it for me and seems to pass all tests (modulo nits; might need to fmt, etc...)
```suggestion
    if let hir::ExprKind::Closure(desugared_inner_closure) = body.value.kind
        && let desugared_inner_closure_body = cx.tcx.hir().body(desugared_inner_closure.body)
        && let Some(GeneratorKind::Async(AsyncGeneratorKind::Closure)) = desugared_inner_closure_body.generator_kind
```

I think I explained it confusingly on zulip yesterday, sorry for that. The `body: &hir::Body<'_>` that this function receives is already the **outer** closure body, so the variable name on line 65 "desugared_outer_closure" is not right. That is already the inner closure. There's no other closures after it, so trying to get another closure won't work.
This is also why the async closure `(async || 2)().await` tests failed: this function would never return true. It's trying to get one closure too much

clippy_lints/src/redundant_closure_call.rs
```diff
@@ -155,14 +157,15 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClosureCall {
...
+                            if let hir::ExprKind::Closure(closure) = body.kind {
+                                let async_closure_body = cx.tcx.hir().body(closure.body);
 
-                            // `async x` is a syntax error, so it becomes `async { x }`
-                            if !matches!(async_closure_body.value.kind, hir::ExprKind::Block(_, _)) {
-                                hint = hint.blockify();
+                                // `async x` is a syntax error, so it becomes `async { x }`
+                                if !matches!(async_closure_body.value.kind, hir::ExprKind::Block(_, _)) {
+                                    hint = hint.blockify();
+                                }
```
y21, line 168: Does this part of the diff functionally change anything? As far as I can tell, only indendation changed and the closure check was moved into its own `if`, right? Is that necessary? I believe this isn't technically right or wrong, because this check always succeeds if `generator_kind` is async (this check was already done in `is_async_closure`), but afaict it does it here again to get the inner closure body to figure out if we have `async || 1`, in which case it needs to add a block: `async { 1 }`, or if the body is already a block: `async || { 1 }`, in which case calling `blockify` is unnecessary (would result in `async { { 1 } }`

KisaragiEffective:
rebased.

y21:
CI is failing, do you need help with this or is this still a wip?

KisaragiEffective:
This is still wip, I think I have reverted necessary commit.

llogiq:
@rustbot author

@KisaragiEffective Please tell rustbot "review" when you've got the PR ready for review. Or do you need any help going forward?

tests/ui/redundant_closure_call_fixable.rs
```diff
@@ -84,3 +85,17 @@ fn issue9956() {
...
+    (|| async {})().await;
+}
+
+mod issue11707 {
+    use core::future::Future;
+
+    fn spawn_on(fut: impl Future<Output = ()>) {}
+
+    fn demo() {
+        spawn_on(((|| async move {})()));
```
KisaragiEffective, line 99: Note for self: remove double-paren

llogiq:
Looks good to me. Thank you!

@bors r+

PR #11364, opened 2023-08-20
Convert `redundant_clone` to an analysis pass

### LINKED ISSUE #10517: FP redundant_clone: Box
### Summary

.

### Lint Name

redundant_clone

### Reproducer

I tried this code:

```rust
pub fn main() {
    let i: Box<_> = Box::new(1);
    let j = i.clone();
    assert_eq!(*i, 2);
}
```

I saw this happen:

```
warning: redundant clone
 --> unique-decl-init-copy.rs:3:14
  |
3 |     let j = i.clone();
  |              ^^^^^^^^ help: remove this
  |
note: cloned value is neither consumed nor mutated
 --> unique-decl-init-copy.rs:3:13
  |
3 |     let j = i.clone();
  |             ^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
  = note: `#[warn(clippy::redundant_clone)]` on by defaul
```

But 
````rust
pub fn main() {
    let i: Box<_> = Box::new(1);
    let j = i;
    assert_eq!(*i, 2);
}

`````
does not compile

````
warning: unused variable: `j`
 --> unique-decl-init-copy.rs:3:9
  |
3 |     let j = i;
  |         ^ help: if this is intentional, prefix it with an underscore: `_j`
  |

error[E0382]: borrow of moved value: `i`
 --> unique-decl-init-copy.rs:4:5
  |
2 |     let i: Box<_> = Box::new(1);
  |         - move occurs because `i` has type `std::boxed::Box<i32>`, which does not implement the `Copy` trait
3 |     let j = i;
  |             - value moved here
4 |     assert_eq!(*i, 2);
  |     ^^^^^^^^^^^^^^^^^ value borrowed here after move
  |
  = note: this error originates in the macro `assert_eq` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider cloning the value if the performance cost is acceptable
  |
3 |     let j = i.clone();
  |              ++++++++

error: aborting due to previous error; 1 warning emitted
````

### Version

```text
rustc 1.70.0-nightly (511364e78 2023-03-16)
binary: rustc
commit-hash: 511364e7874dba9649a264100407e4bffe7b5425
commit-date: 2023-03-16
release: 1.70.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

### LINKED ISSUE #10545: redundant_clone: False positive when original value is only used in macro
### Summary

redundant_clone warns when both the cloned and modified value and the original value are used in a macro, but should not warn because the cloned value has been modified.

### Lint Name

redundant_clone

### Reproducer

I tried this code:

```rust
#![warn(clippy::redundant_clone)]

use quote::quote;
use proc_macro2::TokenStream;

pub fn f(args: TokenStream, function1: syn::ItemFn) -> TokenStream {
    let mut function2 = function1.clone();
    function2.sig.constness = None; // modify function2

    quote! {
        #[cfg(not(#args))]
        #function2 // fn ...
        #[cfg(#args)]
        #function1 // const fn ...
    }
}

```

[playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=fc6baff5f044ba98032a254f9b352c3b)

I saw this happen:

```
warning: redundant clone
 --> src/lib.rs:7:34
  |
7 |     let mut function2 = function1.clone();
  |                                  ^^^^^^^^ help: remove this
  |
note: cloned value is neither consumed nor mutated
 --> src/lib.rs:7:25
  |
7 |     let mut function2 = function1.clone();
  |                         ^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::redundant_clone)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^
```

I expected to see this happen: no warning

### Version

```text
rustc 1.70.0-nightly (8be3c2bda 2023-03-24)
binary: rustc
commit-hash: 8be3c2bda6b683f87b24714ba595e8b04faef54c
commit-date: 2023-03-24
host: aarch64-apple-darwin
release: 1.70.0-nightly
LLVM version: 15.0.7
```

### Additional Labels

_No response_

### LINKED ISSUE #10577: False positive: redundant_clone
### Summary

I looked and this didn't look like any of the existing `redundant_clone` false positives. Sometime between `nightly-2022-10-01` and `nightly-2023-03-18` (sorry for the wide range) `clippy::redundant_clone` started false-positive firing here:
https://github.com/readysettech/readyset/blob/82c3e22f64610cfdebed7d3f227ff477b25c5b10/readyset-server/src/controller/migrate/materialization/plan.rs#L277-L278

If that `clone()` call is removed, compilation fails due to `index_on` being used here:
https://github.com/readysettech/readyset/blob/82c3e22f64610cfdebed7d3f227ff477b25c5b10/readyset-server/src/controller/migrate/materialization/plan.rs#LL427C36-L427C44

### Lint Name

clippy::redundant_clone

### Reproducer

See links

### Version

```text
❯ rustc -Vv
rustc 1.70.0-nightly (4a04d086c 2023-03-18)
binary: rustc
commit-hash: 4a04d086cac54a41517d5657b59d5fe2caca2d71
commit-date: 2023-03-18
release: 1.70.0-nightly
LLVM version: 16.0.0
```

### Additional Labels

_No response_

### LINKED ISSUE #10870: [regression in 1.70] [clippy] redundant clone ; cloned value is neither consumed nor mutated

### LINKED ISSUE #10893: Warning on `redundant_clone` when required with scopes
### Summary

The following code requires that `let mut state = orig_state.clone();` is cloned.
```rust
        {
            // The frequency difference is within tolerance.
            let mut state = orig_state.clone();
            state.tsc_khz = Some(state.tsc_khz.unwrap() + (TSC_KHZ_TOL / 2.0).round() as u32);
            assert!(!vcpu
                .is_tsc_scaling_required(state.tsc_khz.unwrap())
                .unwrap());
        }

        {
            // The frequency difference is over the tolerance.
            let mut state = orig_state;
            state.tsc_khz = Some(state.tsc_khz.unwrap() + (TSC_KHZ_TOL * 2.0).round() as u32);
            assert!(!vcpu
                .is_tsc_scaling_required(state.tsc_khz.unwrap())
                .unwrap());
        }
```
Without it an error is emitted:
```
error[E0382]: use of moved value: `orig_state`
   --> src/vmm/src/vstate/vcpu/x86_64.rs:919:29
    |
906 |         let orig_state = vcpu.save_state().unwrap();
    |             ---------- move occurs because `orig_state` has type `vstate::vcpu::x86_64::VcpuState`, which does not implement the `Copy` trait
...
910 |             let mut state = orig_state;
    |                             ---------- value moved here
...
919 |             let mut state = orig_state;
    |                             ^^^^^^^^^^ value used here after move
    |
help: consider cloning the value if the performance cost is acceptable
    |
910 |             let mut state = orig_state.clone();
    |                                       ++++++++

For more information about this error, try `rustc --explain E0382`.
```
With it, clippy emits a warning:
```
warning: redundant clone
   --> src/vmm/src/vstate/vcpu/x86_64.rs:910:39
    |
910 |             let mut state = orig_state.clone();
    |                                       ^^^^^^^^ help: remove this
    |
note: cloned value is neither consumed nor mutated
   --> src/vmm/src/vstate/vcpu/x86_64.rs:910:29
    |
910 |             let mut state = orig_state.clone();
    |                             ^^^^^^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
    = note: `#[warn(clippy::redundant_clone)]` on by default

warning: `vmm` (lib test) generated 1 warning (run `cargo clippy --fix --lib -p vmm --tests` to apply 1 suggestion)
```

See: https://github.com/firecracker-microvm/firecracker/blob/main/src/vmm/src/vstate/vcpu/x86_64.rs#L908:L924

### Lint Name

clippy::redundant_clone

### Reproducer

_No response_

### Version

```text
binary: rustc
commit-date: 2023-05-31
release: 1.70.0
LLVM version: 16.0.2
```

### Additional Labels

_No response_

### LINKED ISSUE #10940: False positive on redundant clone: borrow of moved value
### Summary

Clippy falsely reports a "redundant clone" warning. Removing the clone leads to a `borrow of moved value` error.

This bug occurs on 1.70.0 as well as on nightly.

### Reproducer

I tried this code:

```rust
fn generate_for_loop(ident: Variable, expr: Expression, body: Statement) -> String {
    // Assign expression to variable to access it from within the loop
    let mut expr_ident = ident.clone();
    expr_ident.name = format!("loop_orig_{}", ident.name);
    let mut out_str = format!("{};\n", generate_declare(&expr_ident, Some(expr)));

    // Loop signature
    out_str += &format!(
        "for (let iter_{I} = 0; iter_{I} < {E}.length; iter_{I}++)",
        I = ident.name,
        E = expr_ident.name
    );

    // Block with prepended declaration of the actual variable
    out_str += &generate_block(
        body,
        Some(format!(
            "let {I} = {E}[iter_{I}];\n",
            I = ident.name,
            E = expr_ident.name,
        )),
    );
    out_str
}
```

I expected to see this happen: The code to run successfully and apply fixes suggested by `cargo clippy --fix`.

Instead, this happened: I encountered the following error:

```
warning: failed to automatically apply fixes suggested by rustc to crate `sb`

  * src/generator/js.rs

https://github.com/rust-lang/rust-clippy/issues

error[E0382]: borrow of moved value: `ident`
   --> src/generator/js.rs:188:47
    |
185 | fn generate_for_loop(ident: Variable, expr: Expression, body: Statement) -> String {
    |                      ----- move occurs because `ident` has type `ast::Variable`, which does not implement the `Copy` trait
186 |     // Assign expression to variable to access it from within the loop
187 |     let mut expr_ident = ident;
    |                          ----- value moved here
188 |     expr_ident.name = format!("loop_orig_{}", ident.name);
    |                                               ^^^^^^^^^^ value borrowed here after move
    |
    = note: this error originates in the macro `$crate::__export::format_args` which comes from the expansion of the macro `format` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider cloning the value if the performance cost is acceptable
    |
187 |     let mut expr_ident = ident.clone();
    |                               ++++++++

For more information about this error, try `rustc --explain E0382`.

warning: redundant clone
   --> src/generator/js.rs:187:31
    |
187 |     let mut expr_ident = ident.clone();
    |                               ^^^^^^^^ help: remove this
    |
note: cloned value is neither consumed nor mutated
   --> src/generator/js.rs:187:

26
    |
187 |     let mut expr_ident = ident.clone();
    |                          ^^^^^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
    = note: `#[warn(clippy::redundant_clone)]` on by default

warning: `antimony-lang` (bin "sb") generated 1 warning (run `cargo clippy --fix --bin "sb"` to apply 1 suggestion)
```

### Version

```text
rustc 1.72.0-nightly (df77afbca 2023-06-12)
binary: rustc
commit-hash: df77afbcaf3365a32066a8ca4a00ae6fc9a69647
commit-date: 2023-06-12
host: aarch64-apple-darwin
release: 1.72.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

Jarcho:
fixes #10940
fixes #10893
fixes #10870
fixes #10577
fixes #10545
fixes #10517

This is mostly finished, but blocked until rust-lang/rust#114900 is merged (or something similar). This isn't worth reviewing until then.

changelog:

### DIFF
--- a/clippy_lints/src/redundant_clone.rs
+++ b/clippy_lints/src/redundant_clone.rs
@@ -1,28 +1,27 @@
 use clippy_utils::diagnostics::{span_lint_hir, span_lint_hir_and_then};
-use clippy_utils::mir::{visit_local_usage, LocalUsage, PossibleBorrowerMap};
+use clippy_utils::mir::PossibleBorrowerMap;
 use clippy_utils::source::snippet_opt;
-use clippy_utils::ty::{has_drop, is_copy, is_type_diagnostic_item, is_type_lang_item, walk_ptrs_ty_depth};
+use clippy_utils::ty::{is_type_diagnostic_item, is_type_lang_item};
 use clippy_utils::{fn_has_unsatisfiable_preds, match_def_path, paths};
-use if_chain::if_chain;
+use core::mem::take;
+use rustc_data_structures::fx::{FxHashMap, FxHashSet};
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::FnKind;
-use rustc_hir::{def_id, Body, FnDecl, LangItem};
+use rustc_hir::{self as hir, def_id, FnDecl, LangItem, Mutability};
+use rustc_index::bit_set::{BitSet, HybridBitSet};
 use rustc_lint::{LateContext, LateLintPass};
-use rustc_middle::mir;
-use rustc_middle::ty::{self, Ty};
+use rustc_middle::mir::{
+    BasicBlock, BasicBlockData, Body, BorrowKind, CallReturnPlaces, InlineAsmOperand, Local, Location,
+    NonDivergingIntrinsic, Operand, Rvalue, Statement, StatementKind, Terminator, TerminatorEdges, TerminatorKind,
+};
+use rustc_middle::ty::{self, ParamEnv, TyCtxt};
+use rustc_mir_dataflow::fmt::DebugWithContext;
+use rustc_mir_dataflow::{Analysis, AnalysisDomain, JoinSemiLattice};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::def_id::LocalDefId;
 use rustc_span::source_map::{BytePos, Span};
 use rustc_span::sym;
-
-macro_rules! unwrap_or_continue {
-    ($x:expr) => {
-        match $x {
-            Some(x) => x,
-            None => continue,
-        }
-    };
-}
+use std::collections::hash_map::Entry;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -70,7 +69,7 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
         cx: &LateContext<'tcx>,
         _: FnKind<'tcx>,
         _: &'tcx FnDecl<'_>,
-        _: &'tcx Body<'_>,
+        _: &'tcx hir::Body<'_>,
         _: Span,
         def_id: LocalDefId,
     ) {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +68,724 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+        }
+        changed
+    }
+}
+
+#[derive(Default, Debug, Clone)]
+struct CloneDomain {
+    links: FxHashSet<(Local, Local, BasicBlock)>,
+    values: FxHashMap<Local, Value>,
+    candidates: Candidates,
```
cjgillot, line 243: This field should not be part of the analysis domain, but computed in post-processing.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +68,724 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+}
 
-            let from_borrow = match_def_path(cx, fn_def_id, &paths::CLONE_TRAIT_METHOD)
-                || match_def_path(cx, fn_def_id, &paths::TO_OWNED_METHOD)
-                || (match_def_path(cx, fn_def_id, &paths::TO_STRING_METHOD)
-                    && is_type_lang_item(cx, arg_ty, LangItem::String));
+#[derive(Debug, Clone)]
+struct Value {
+    last_modified: Option<BasicBlock>,
+    lint_unread: HybridBitSet<BasicBlock>,
```
cjgillot, line 187: This field looks like it attempts to model a liveness analysis. Can it be replaced by `MayleLiveLocals` that we already have in store?

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +68,724 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+            }
+        }
+        changed
+    }
+}
+
+#[derive(Default, Debug, Clone)]
+struct CloneDomain {
+    links: FxHashSet<(Local, Local, BasicBlock)>,
+    values: FxHashMap<Local, Value>,
```
cjgillot, line 242: IIUC, you are trying to compute which locals store the same value.
Can this be achieved by using an `IndexVec<Local, FlatSet<Location>>`?
`FlatSet::Bottom` means uninitialized / moved-out.
`FlatSet::Elem(location)` means "assigned some unique value created at this location".
`FlatSet::Top` means "could store anything".
And in case of an assignment, propagate the `FlatSet` value.

cjgillot (commented):
I left a few comments and questions, to gain some perspective on the use case of https://github.com/rust-lang/rust/pull/114900.
My impression is that you are trying to bend the dataflow API to do too much at once.
This analysis could be much simpler, and not require any change to the dataflow API.

clippy_lints/src/redundant_clone.rs
Jarcho: This is tracking if a read has occurred since the paired value was first modified.

clippy_lints/src/redundant_clone.rs
Jarcho: Not just which locals are duplicates of each other. This also needs to track how the duplicate values are used in relation to each other. We also need to know where the duplicate was made in order to lint that span.

clippy_lints/src/redundant_clone.rs
Jarcho: I think that would work.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +68,724 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+                if let TerminatorKind::Call { args, .. } = &data.terminator().kind
+                    && let [Operand::Copy(arg) | Operand::Move(arg), ..] = &**args
+                    && let Some(Some(targets)) = ref_target_results.get().targets.get(&arg.local)
+                {
+                    Some((block, (arg.local, targets.clone())))
+                } else {
+                    None
+                }
+            })
+            .collect::<FxHashMap<_, _>>();
```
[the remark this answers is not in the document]
Jarcho: Typically yes. If there ends up being perf problem this would be worth not linting the few extra cases this could catch. Limiting the analysis to just types which are cloned could also get a decent speed up

Jarcho:
Thanks for taking a look. Assuming pulling `Candidates` out of the analysis works that does unblock this.

It would still be nice to be able to change the links stored in the domain to a bitset and have the analysis convert between the `(Local, Local, Block)` triplet and the index. Doing so would require the join to have access to the analysis as it needs the triplet in order to work.

---

Just as further elaboration the basic idea for the lint it starts from the clone call and tracks both values. When one of the two values is modified, it then switches to checking if the non-modified version is read. If it is, then the clone is considered necessary. e.g. 
```rust
// `x` is assigned a duplicate value of `y`
// a new value in `links` keeps track of the relation and the clone location
let x = y.clone(); 

// all values linked to `y` have `lint_unread` modified here 
y.modify();

// since `y` doesn't have the original value, the clone is necessary
// any clone call in `x`'s `lint_unread` field is a necessary clone
x.read();
```
If the order of the read and write are swapped the clone call isn't necessary since the read could have occurred on the original value.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+                    };
 
-            if terminator.source_info.span.from_expansion() {
-                continue;
+                    if !borrowers.bounded_borrowers(&[], &[], local, loc) {
+                        required_clones.insert(block);
+                    }
+                }
             }
+        }
```
Jarcho, line 168: The `required_clones.insert(..)` part should occur inside the join, but the `PossibleBorrowerMap` can't be accessed from there.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+}
+impl DebugWithContext<CloneAnalysis<'_, '_, '_>> for CloneDomain {}
+
+struct CloneAnalysis<'a, 'mir, 'tcx> {
+    body: &'mir Body<'tcx>,
+    tracked: BitSet<Local>,
+    clone_sources: &'a FxHashMap<BasicBlock, (Local, HybridBitSet<Local>)>,
+    link_buf: Vec<(Local, Local, BasicBlock)>,
+    borrowers: PossibleBorrowerMap<'mir, 'tcx>,
+    required_clones: FxHashSet<BasicBlock>,
```
Jarcho, line 257: This could be calculated in a `ResultsVisitor` thereby resolving the earlier issue with `PossibleBorrowerMap`, but that requires the visitor to detect reads and moves the same way the analysis does.

Given that handling each read is a small amount of work, and using a visitor requires a whole extra pass over every block it's probably better to do it this way. It's also just less work to implement/maintain.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+            span_lint_hir(cx, REDUNDANT_CLONE, node, info.span, "redundant clone");
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct Value {
+    // Two values which were last modified at different locations may contain different values.
+    // Joining two such values needs to be treated as a modification.
+    last_modified: Option<Location>,
```
Jarcho, line 191: This really should be something like:
```rust
enum Index {
    Statement(usize),
    Merge(Local),
}
struct Location {
    block: BasicBlock,
    index: Index,
}
```

This would allow distinguishing different merged values from each other, but would require `join` to know which block it's working on.

Jarcho (commented):
@cjgillot I've made some changes to the analysis. Though not strictly required, the changes to the dataflow api would still be useful.

clippy_lints/src/redundant_clone.rs
cjgillot: I would represent a value with a simple index. A first pass collects value roots. Those would be the value of a variable where it is borrowed before cloning.

The state would be something like `IndexVec<Local, FlatSet<Value>>`.

```rust
let mut x = if _ { foo() } else { bar() };
let y = if _ { foo2() } else { bar2() };

// `x` is borrowed here. `x` is `FlatSet::Elem(Value(0))` here.
let z = x.clone();
// `z` is `FlatSet::Elem(Value(0))`.

// assuming the match doesn't modify `x`, `y`, and `z`, then the join when merging
// the arms shouldn't detect a modification
match _ {
    _ => { ... },
    _ => { ... },
    _ => { ... },
};
// `x` and `z` are still both `FlatSet::Elem(Value(0))`

// here the clone isn't needed: `x` and `y` are both `FlatSet::Elem(Value(0))`.
do_something(&z);

if _ {
    x = y;
    // `x` would be `FlatSet::Top` if `y` is not tracked.
}
// By joining, `x` is `FlatSet::Top`.

// here the clone is needed: `x` is `FlatSet::Top`, while `z` is `FlatSet::Elem(Value(0))`
do_something(&z);
```

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
-        if let ty::FnDef(def_id, _) = *func.ty(mir, cx.tcx).kind();
-        if let (inner_ty, 1) = walk_ptrs_ty_depth(args[0].ty(mir, cx.tcx));
-        if !is_copy(cx, inner_ty);
-        then {
-            Some((def_id, *local, inner_ty, destination.as_local()?))
-        } else {
-            None
+#[derive(Default, Debug, Clone)]
+struct RefTargetDomain {
+    targets: FxHashMap<Local, Option<HybridBitSet<Local>>>,
```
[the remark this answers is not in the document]
Jarcho: That is correct.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+}
+impl JoinSemiLattice for RefTargetDomain {
+    fn join(&mut self, other: &Self) -> bool {
```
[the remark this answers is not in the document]
Jarcho: That is correct. Basically `uninit < local < maybe non-local`

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+                            state.targets.get(&src).cloned().flatten()
+                        }
+                        Rvalue::Use(Operand::Move(src)) if let Some(src) = src.as_local() => {
+                            state.targets.remove(&src).flatten()
+                        }
+                        _ => None,
+                    };
+                    state.targets.insert(dst, targets);
+                } else if let Rvalue::Use(Operand::Move(src))
+                    | Rvalue::Ref(_, BorrowKind::Mut { .. }, src) = a.1
```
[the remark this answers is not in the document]
Jarcho: That is an error. It should be setting the value to `None`, not removing it.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+                        Rvalue::Ref(_, BorrowKind::Shallow | BorrowKind::Shared, target) => {
+                            if let Some(target) = target.as_local()
+                                && !self.raw_ptr_taken.contains(target)
+                            {
+                                let mut targets = HybridBitSet::new_empty(self.tracked.domain_size());
+                                targets.insert(target);
+                                Some(targets)
+                            } else {
+                                None
+                            }
```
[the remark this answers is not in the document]
Jarcho: Not currently handled.

clippy_lints/src/redundant_clone.rs
```diff
@@ -70,317 +69,729 @@ impl<'tcx> LateLintPass<'tcx> for RedundantClone {
...
+                if let TerminatorKind::Call { args, .. } = &data.terminator().kind
+                    && let [Operand::Move(arg), ..] = &**args
+                    && let Some(Some(targets)) = ref_target_results.get().targets.get(&arg.local)
+                {
+                    Some((block, (arg.local, targets.clone())))
+                } else {
+                    None
+                }
+            })
+            .collect::<FxHashMap<_, _>>();
```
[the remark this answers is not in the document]
Jarcho: The mapping is `block -> possible clone sources`. The argument local doesn't strictly need to be there.

clippy_lints/src/redundant_clone.rs
Jarcho: That wouldn't handle cloning a value twice correctly, nor would it handle a chain of clones. e.g.
```rust
let x = foo();
let y = if bar {
    x.clone()
} else {
    x.clone()
};
let z = y.clone();
// `x`, `y` and `z` should all be seen as the same value here.
```

xFrednet:
Hey, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

xFrednet:

apiraino:
(t-compiler triage :wave: )

I just noticed this PR was closed for inactivity. IIUC this was part of the work on [rust#114900](https://github.com/rust-lang/rust/pull/114900) (which admittedly was opened long ago). I'll try to get eyes on that PR to clarify its status and - by extension - this one, too.

Issue #11330, opened 2023-08-12
Concisely creating a vector of a known capacity


Kittenlover229:
### What it does

Informs the user about a more concise way to create a vector with a known capacity.

I often wrote code like the following: 
```rs
{ 
    let mut v = vec![];
    v.reserve(space_hint);
    v
}
```
since I didn't know about the existence of `Vec::with_capacity`, maybe clippy can warn about things like that? It also seems like this can be generalized to cases when we are pushing into an empty vector for a known amount of times. 

### Example

```rust
{ 
    let mut v = vec![];
    v.reserve(space_hint);
    v
}
```

Could be written as:

```rust
Vec::with_capacity(space_hint)
```

PR #11373, opened 2023-08-21
Added new lint: `reserve_after_initialization`

Red-Rapious:
Closes https://github.com/rust-lang/rust-clippy/issues/11330.

A new lint that informs the user about a more concise way to create a vector with a known capacity.
Example:
```rust
let mut v: Vec<usize> = vec![];
v.reserve(10);
```

Produces the following help:
```rust
  |
2 | /     let mut v: Vec<usize> = vec![];
3 | |     v.reserve(10);
  | |__________________^ help: consider using `Vec::with_capacity(space_hint)`: `let v: Vec<usize> = Vec::with_capacity(10);`
  |
```

And can be rewritten as:
```rust
let v: Vec<usize> = Vec::with_capacity(10);
```

changelog: new lint [`reserve_after_initialization`]

### DIFF
--- a/clippy_lints/src/reserve_after_initialization.rs
+++ b/clippy_lints/src/reserve_after_initialization.rs
@@ -1,0 +1,242 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::higher::{get_vec_init_kind, VecInitKind};
+use clippy_utils::source::snippet;
+use clippy_utils::visitors::for_each_local_use_after_expr;
+use clippy_utils::{get_parent_expr, path_to_local_id};
+use core::ops::ControlFlow;
+use rustc_ast::LitKind;
+use rustc_errors::Applicability;
+use rustc_hir::def::Res;
+use rustc_hir::{
+    BindingAnnotation, Block, Expr, ExprKind, HirId, Local, Mutability, PatKind, QPath, Stmt, StmtKind, UnOp,
+};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::{Span, Symbol};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Informs the user about a more concise way to create a vector with a known capacity.
+    ///
+    /// ### Why is this bad?
+    /// The `Vec::with_capacity` constructor is easier to understand.
+    ///
+    /// ### Example
+    /// ```rust
+    /// {
+    ///     let mut v = vec![];
+    ///     v.reserve(space_hint);
+    ///     v
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// Vec::with_capacity(space_hint)
+    /// ```
+    #[clippy::version = "1.73.0"]
+    pub RESERVE_AFTER_INITIALIZATION,
+    complexity,
+    "`reserve` called immediatly after `Vec` creation"
+}
+impl_lint_pass!(ReserveAfterInitialization => [RESERVE_AFTER_INITIALIZATION]);
+
+/*impl<'tcx> LateLintPass<'tcx> for ReserveAfterInitialization {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx rustc_hir::Expr<'_>) {
+        if let ExprKind::Assign(left, right, _) = expr.kind
+        && let ExprKind::Path(QPath::Resolved(None, path)) = left.kind
+        && let [name] = &path.segments
+        && let Res::Local(id) = path.res
+        && let Some(init) = get_vec_init_kind(cx, right)
+        && !matches!(init, VecInitKind::WithExprCapacity(_)) {
+            span_lint_and_help(
+                cx,
+                RESERVE_AFTER_INITIALIZATION,
+                expr.span,
+                "`reserve` called just after the initialisation of the vector",
+                None,
+                "use `Vec::with_capacity(space_hint)` instead"
+            );
+        }
+    }
+
+    /*fn check_block(&mut self, cx: &LateContext<'_>, block: &'_ rustc_hir::Block<'_>) {
+        span_lint_and_help(
+            cx,
+            RESERVE_AFTER_INITIALIZATION,
+            block.span,
+            "`reserve` called just after the initialisation of the vector",
+            None,
+            "use `Vec::with_capacity(space_hint)` instead"
+        );
+    }*/
+}*/
+
+#[derive(Default)]
+pub struct ReserveAfterInitialization {
+    searcher: Option<VecReserveSearcher>,
+}
+
+struct VecReserveSearcher {
+    local_id: HirId,
+    lhs_is_let: bool,
+    let_ty_span: Option<Span>,
+    name: Symbol,
+    err_span: Span,
+    last_reserve_expr: HirId,
+    space_hint: usize,
+}
+impl VecReserveSearcher {
+    fn display_err(&self, cx: &LateContext<'_>) {
+        if self.space_hint == 0 {
+            return;
+        }
+
+        let mut needs_mut = false;
+        let _res = for_each_local_use_after_expr(cx, self.local_id, self.last_reserve_expr, |e| {
+            let Some(parent) = get_parent_expr(cx, e) else {
+                return ControlFlow::Continue(());
+            };
+            let adjusted_ty = cx.typeck_results().expr_ty_adjusted(e);
+            let adjusted_mut = adjusted_ty.ref_mutability().unwrap_or(Mutability::Not);
+            needs_mut |= adjusted_mut == Mutability::Mut;
+            match parent.kind {
+                ExprKind::AddrOf(_, Mutability::Mut, _) => {
+                    needs_mut = true;
+                    return ControlFlow::Break(true);
+                },
+                ExprKind::Unary(UnOp::Deref, _) | ExprKind::Index(..) if !needs_mut => {
+                    let mut last_place = parent;
+                    while let Some(parent) = get_parent_expr(cx, last_place) {
+                        if matches!(parent.kind, ExprKind::Unary(UnOp::Deref, _) | ExprKind::Field(..))
+                            || matches!(parent.kind, ExprKind::Index(e, _, _) if e.hir_id == last_place.hir_id)
+                        {
+                            last_place = parent;
+                        } else {
+                            break;
+                        }
+                    }
+                    needs_mut |= cx.typeck_results().expr_ty_adjusted(last_place).ref_mutability()
+                        == Some(Mutability::Mut)
+                        || get_parent_expr(cx, last_place)
+                            .map_or(false, |e| matches!(e.kind, ExprKind::AddrOf(_, Mutability::Mut, _)));
+                },
+                ExprKind::MethodCall(_, recv, ..)
+                    if recv.hir_id == e.hir_id
+                        && adjusted_mut == Mutability::Mut
+                        && !adjusted_ty.peel_refs().is_slice() =>
+                {
+                    // No need to set `needs_mut` to true. The receiver will be either explicitly borrowed, or it will
+                    // be implicitly borrowed via an adjustment. Both of these cases are already handled by this point.
+                    return ControlFlow::Break(true);
+                },
+                ExprKind::Assign(lhs, ..) if e.hir_id == lhs.hir_id => {
+                    needs_mut = true;
+                    return ControlFlow::Break(false);
+                },
+                _ => (),
+            }
+            ControlFlow::Continue(())
+        });
+
+        let mut s = if self.lhs_is_let {
+            String::from("let ")
+        } else {
+            String::new()
+        };
+        if needs_mut {
+            s.push_str("mut ");
+        }
+        s.push_str(self.name.as_str());
+        if let Some(span) = self.let_ty_span {
+            s.push_str(": ");
+            s.push_str(&snippet(cx, span, "_"));
+        }
+        s.push_str(format!(" = Vec::with_capacity({});", self.space_hint).as_str());
+
+        span_lint_and_sugg(
+            cx,
+            RESERVE_AFTER_INITIALIZATION,
+            self.err_span,
+            "calls to `reverse` immediately after creation",
+            "consider using `Vec::with_capacity(space_hint)`",
+            s,
+            Applicability::HasPlaceholders,
+        );
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for ReserveAfterInitialization {
+    fn check_block(&mut self, _: &LateContext<'tcx>, _: &'tcx Block<'tcx>) {
+        self.searcher = None;
+    }
+
+    fn check_local(&mut self, cx: &LateContext<'tcx>, local: &'tcx Local<'tcx>) {
+        if let Some(init_expr) = local.init
+            && let PatKind::Binding(BindingAnnotation::MUT, id, name, None) = local.pat.kind
+            && !in_external_macro(cx.sess(), local.span)
+            && let Some(init) = get_vec_init_kind(cx, init_expr)
+            && !matches!(init, VecInitKind::WithExprCapacity(_))
+        {
+            self.searcher = Some(VecReserveSearcher {
+                local_id: id,
+                lhs_is_let: true,
+                name: name.name,
+                let_ty_span: local.ty.map(|ty| ty.span),
+                err_span: local.span,
+                last_reserve_expr: init_expr.hir_id,
+                space_hint: 0
+            });
+        }
+    }
+
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if self.searcher.is_none()
+            && let ExprKind::Assign(left, right, _) = expr.kind
+            && let ExprKind::Path(QPath::Resolved(None, path)) = left.kind
+            && let [name] = &path.segments
+            && let Res::Local(id) = path.res
+            && !in_external_macro(cx.sess(), expr.span)
+            && let Some(init) = get_vec_init_kind(cx, right)
+            && !matches!(init, VecInitKind::WithExprCapacity(_))
+        {
+            self.searcher = Some(VecReserveSearcher {
+                local_id: id,
+                lhs_is_let: false,
+                let_ty_span: None,
+                name: name.ident.name,
+                err_span: expr.span,
+                last_reserve_expr: expr.hir_id,
+                space_hint: 0
+            });
+        }
+    }
+
+    fn check_stmt(&mut self, cx: &LateContext<'tcx>, stmt: &'tcx Stmt<'_>) {
+        if let Some(searcher) = self.searcher.take() {
+            if let StmtKind::Expr(expr) | StmtKind::Semi(expr) = stmt.kind
+                && let ExprKind::MethodCall(name, self_arg, other_args, _) = expr.kind
+                && other_args.len() == 1
+                && let ExprKind::Lit(lit) = other_args[0].kind
+                && let LitKind::Int(space_hint, _) = lit.node
+                && path_to_local_id(self_arg, searcher.local_id)
+                && name.ident.as_str() == "reserve"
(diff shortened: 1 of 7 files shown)


clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,225 @@
...
+                    return ControlFlow::Break(true);
+                },
+                ExprKind::Assign(lhs, ..) if e.hir_id == lhs.hir_id => {
+                    needs_mut = true;
+                    return ControlFlow::Break(false);
+                },
+                _ => (),
+            }
+            ControlFlow::Continue(())
+        });
```
y21, line 106: I think this doesn't handle  `let ref mut y = v;` correctly (i.e. will not detect `v` as requiring mutability after a reserve call).
But also, this mutability checking seems like a lot of code with lots of logic for a suggestion that isn't machine applicable. I feel like that's not really necessary for a `HasPlaceholders` suggestion and clippy could just leave `mut` out and expect the user to fill it in if needed, or maybe `snippet` could be used to get the source string from `let` to `=`.

y21 (commented):
Looks pretty good to me already for a first contribution 👀 but I have one question

clippy_lints/src/reserve_after_initialization.rs
Red-Rapious: Thank you for your feedback! I will indeed remove this `mut` checking and let the user adapt it.

Red-Rapious: I decided to go for the `snippet` option instead, and added 3 more tests to show the different cases.

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+use rustc_middle::lint::in_external_macro;
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Informs the user about a more concise way to create a vector with a known capacity.
+    ///
+    /// ### Why is this bad?
+    /// The `Vec::with_capacity` constructor is easier to understand.
```
blyxyas, line 19: "easier to understand" seems kinda subjective, asserting that it's `less complex` or `more readable` would transmit the idea better.

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+        {
+            self.searcher = Some(VecReserveSearcher {
+                local_id: id,
+                err_span: local.span,
+                init_part: format!("let {}: {}", 
+                snippet(cx, local.pat.span, ""),
+                match local.ty {
+                    Some(type_inference) => snippet(cx, type_inference.span, "").to_string(),
+                    None => String::new()
+                }),
```
y21, line 89: Instead of getting two snippets here, this could probably pass a single larger span to `snippet`, from `let` to the low part of the initializer expression. This might simplify it a bit further and also make it slightly nicer in case there is no type annotation (right now it would suggest `let x: = Vec::with_capacity()`)
```suggestion
                init_part: snippet(cx, local.span.shrink_to_lo().to(init_expr.span.shrink_to_lo()), "").into_owned(),
```
This is what `init_part` would store with this change:
```rs
let mut v = Vec::new();
^^^^^^^^^^^^
```
Then, in `display_err` it could be just `format!("{}Vec::with_capacity({})", self.init_part, self.space_hint)` (without the `=` sign and whitespaces, since that will already be there).
This would also need to be done in `check_expr` if this is implemented in this place.

(It might even be possible to use `write!` or `String::push_str` instead of `format!` since this is just appending to an already existing string)

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.searcher = None;
+    }
+
+    fn check_local(&mut self, cx: &LateContext<'tcx>, local: &'tcx Local<'tcx>) {
+        if let Some(init_expr) = local.init
+            && let PatKind::Binding(BindingAnnotation::MUT, id, _name, None) = local.pat.kind
+            && !in_external_macro(cx.sess(), local.span)
+            && let Some(init) = get_vec_init_kind(cx, init_expr)
+            && !matches!(init, VecInitKind::WithExprCapacity(_))
+            && !matches!(init, VecInitKind::WithConstCapacity(_))
```
blyxyas, line 79: This can be optimized with some magic patterns

```suggestion
            && !matches!(init, VecInitKind::WithExprCapacity(_)
               | VecInitKind::WithConstCapacity(_)
            )
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+impl<'tcx> LateLintPass<'tcx> for ReserveAfterInitialization {
+    fn check_block(&mut self, _: &LateContext<'tcx>, _: &'tcx Block<'tcx>) {
+        self.searcher = None;
+    }
+
+    fn check_local(&mut self, cx: &LateContext<'tcx>, local: &'tcx Local<'tcx>) {
+        if let Some(init_expr) = local.init
+            && let PatKind::Binding(BindingAnnotation::MUT, id, _name, None) = local.pat.kind
```
blyxyas, line 75: This `_name` is never used, it can be discarded.
```suggestion
            && let PatKind::Binding(BindingAnnotation::MUT, id, _, None) = local.pat.kind
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+            && !matches!(init, VecInitKind::WithExprCapacity(_))
+            && !matches!(init, VecInitKind::WithConstCapacity(_))
+        {
+            self.searcher = Some(VecReserveSearcher {
+                local_id: id,
+                err_span: local.span,
+                init_part: format!("let {}: {}", 
+                snippet(cx, local.pat.span, ""),
+                match local.ty {
+                    Some(type_inference) => snippet(cx, type_inference.span, "").to_string(),
```
blyxyas, line 87: AFAIK it's good practice to use `".."` as snippet defaults. If that default should never be met, you can check for the snippet with [`snippet_opt`](https://doc.rust-lang.org/beta/nightly-rustc/clippy_utils/source/fn.snippet_opt.html) in the initial_matching.

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        let s = format!("{} = Vec::with_capacity({});", self.init_part, self.space_hint);
+
+        span_lint_and_sugg(
+            cx,
+            RESERVE_AFTER_INITIALIZATION,
+            self.err_span,
+            "calls to `reverse` immediately after creation",
+            "consider using `Vec::with_capacity(space_hint)`",
```
blyxyas, line 61: `rust-analyzer` in IDE integrations can automatically place the cursor over the placeholder if you use this format. (If the user has integrated Clippy with their RA)

```suggestion
            "consider using `Vec::with_capacity(/* Space hint */)`",
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+            return;
+        }
+
+        let s = format!("{} = Vec::with_capacity({});", self.init_part, self.space_hint);
+
+        span_lint_and_sugg(
+            cx,
+            RESERVE_AFTER_INITIALIZATION,
+            self.err_span,
+            "calls to `reverse` immediately after creation",
```
blyxyas, line 60:
```suggestion
            "call to `reserve` immediately after creation",
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+                space_hint: String::new()
+            });
+        }
+    }
+
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if self.searcher.is_none()
+            && let ExprKind::Assign(left, right, _) = expr.kind
+            && let ExprKind::Path(QPath::Resolved(None, path)) = left.kind
+            && let [_name] = &path.segments
```
blyxyas, line 99: This named pattern destructuring could just be a wildcard, as it isn't used elsewhere. It will still have the same destructuring meaning.

```suggestion
            && let [_] = &path.segments
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    fn check_stmt(&mut self, cx: &LateContext<'tcx>, stmt: &'tcx Stmt<'_>) {
+        if let Some(searcher) = self.searcher.take() {
+            if let StmtKind::Expr(expr) | StmtKind::Semi(expr) = stmt.kind
+                && let ExprKind::MethodCall(name, self_arg, other_args, _) = expr.kind
+                && other_args.len() == 1
+                && path_to_local_id(self_arg, searcher.local_id)
+                && name.ident.as_str() == "reserve"
```
blyxyas, line 121: Using the [`sym!`](https://doc.rust-lang.org/beta/nightly-rustc/clippy_utils/macro.sym.html) macro here would be faster (`Symbol::as_str` is slow as it has to lock the symbol interner). Creating a symbol from a string is faster than creating a string from a symbol :cat:

```suggestion
                && name.ident.as_str() == sym!(reserve)
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+            {
+                self.searcher = Some(VecReserveSearcher {
+                    err_span: searcher.err_span.to(stmt.span),
+                    space_hint: snippet(cx, other_args[0].span, "").to_string(),
```
blyxyas, line 125:
```suggestion
                    space_hint: snippet(cx, other_args[0].span, "..").to_string(),
```

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+                space_hint: String::new()
+            });
+        }
+    }
+
+    fn check_stmt(&mut self, cx: &LateContext<'tcx>, stmt: &'tcx Stmt<'_>) {
+        if let Some(searcher) = self.searcher.take() {
+            if let StmtKind::Expr(expr) | StmtKind::Semi(expr) = stmt.kind
+                && let ExprKind::MethodCall(name, self_arg, other_args, _) = expr.kind
+                && other_args.len() == 1
```
y21, line 119: mini nit
```suggestion
                && let ExprKind::MethodCall(name, self_arg, [space_hint], _) = expr.kind
```
then `space_hint` can be used instead of `other_args[0]`

blyxyas (commented):
WOW, really cool for a first lint, just some very minor nits, but very **very** good overall :heart:, I'm really impressed!

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+
+struct VecReserveSearcher {
+    local_id: HirId,
+    err_span: Span,
+    init_part: String,
+    space_hint: String,
+}
+impl VecReserveSearcher {
+    fn display_err(&self, cx: &LateContext<'_>) {
+        if self.space_hint.is_empty() {
```
blyxyas, line 50:
```suggestion
        if self.space_hint.is_empty() || is_from_proc_macro(err_span) {
```

tests/ui/reserve_after_initialization.rs
blyxyas, line 1: This should include some tests with macros, you can get some inspiration with other test files like `tests/ui/allow_attributes.rs` (that's a simple one)

tests/ui/reserve_after_initialization.rs
```diff
@@ -0,0 +1,19 @@
+#![warn(clippy::reserve_after_initialization)]
```
y21, line 1: A test for assign expressions would be nice to have.  Something like
```rs
let mut x: Vec<usize> = Vec::new();
x = Vec::new();
x.reserve(10);
```

clippy_lints/src/reserve_after_initialization.rs
y21: Oh, the init_expr span part probably needs to be `init_expr.span.source_callsite().shrink_to_lo()`, since the initializer expression can be part of a macro expansion (`vec![]`). In those cases we need the real source span that invoked that macro and not some span inside the `vec![]` macro.

clippy_lints/src/reserve_after_initialization.rs
y21: Hmm, doesn't `sym!` do the same (locking the interner)? 🤔 
https://doc.rust-lang.org/beta/nightly-rustc/src/rustc_span/symbol.rs.html#1914-1916

tests/ui/reserve_after_initialization.rs
Red-Rapious: How do I manually test an example file containing a macro? When running:
```console
$ cargo dev lint tests/ui/allow_attributes.rs
```
I get the following error:
```rust
error[E0463]: can't find crate for `proc_macros`
 --> tests/ui/allow_attributes.rs:7:1
  |
7 | extern crate proc_macros;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
```

clippy_lints/src/reserve_after_initialization.rs
blyxyas: Mmmm... You're right, `sym!` expands to `intern` that also uses `self.0.lock()`. Maybe it should be mentioned on the `sym!` docs. / `intern`'s documentation. So yeah, it doesn't matter if it's changed.

Red-Rapious:
I added some tests and fixed everything you mentioned except for:
- the use of `syml!`, since it apparently will not improve performance (I'm glad, I couldn't get it to work properly anyway)
- the macros tests: I have trouble using an external crate while running tests manually, as I explained above

tests/ui/reserve_after_initialization.rs
blyxyas: To run a specific lint use the [`TESTNAME` env. var.](https://doc.rust-lang.org/cargo/commands/cargo-test.html) and the `cargo uitest` command.

```sh
$ TESTNAME="allow_attributes" cargo uitest
```

Using `cargo dev lint` is used to execute your lint on a normal file AFAIK, it isn't really used (at least I've never used it)

y21:
Btw I think you can mark things that are done as resolved, that might make it a little easier to read through and see what's left ^^ (saying this because my very first comment is taking up lots of space and that comment is not relevant anymore/already applied, I can't click "resolve" myself it looks like)

clippy_lints/src/reserve_after_initialization.rs
blyxyas: This still needs to be addressed. Having some pattern magic here (`|`) makes faster both compiling (less macros) and more runtime (less branches)

Red-Rapious: My bad, I changed it in `check_expr` but not in `check_local`.

blyxyas:
I just noticed that actually testing with macros isn't really documented anywhere, so I'll do a tutorial here.

In our test suite headers can be used with the "//@" syntax. E.g. "//@run-rustfix" as the first line is a header.
You can use the header `//@aux-build:proc_macros.rs:proc-macro` to import the auxiliary create just for testing procedural macros and how lints interact with them.

So, after using `//@aux-build:proc_macros.rs:proc-macro` you can then `extern macro proc_macro;`, and then `use proc_macros::{external, with_span}`.

The auxiliary proc macro `external!` simulates a proc macro that is external to the crate (I think), and the `with_span` macro simulates a proc macro that changes the span, so you can verify if you're linting the correct expression and such.

We really should document this, #10597 
If in the future something isn't in the book or other documentation (and a reviewer isn't online), you can always check the `rustc-dev-guide`, it's a book mainly for developing rustc and maybe cargo, but it has some things that also apply to Clippy. (if a reviewer is online, you can always ask for help here or on the [Zulip server](https://rust-lang.zulipchat.com/), we all check that chat more or less periodically :heart: )

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    fn check_local(&mut self, cx: &LateContext<'tcx>, local: &'tcx Local<'tcx>) {
+        if let Some(init_expr) = local.init
+            && let PatKind::Binding(BindingAnnotation::MUT, id, _, None) = local.pat.kind
+            && !in_external_macro(cx.sess(), local.span)
+            && let Some(init) = get_vec_init_kind(cx, init_expr)
+            && !matches!(init, VecInitKind::WithExprCapacity(_)
+               | VecInitKind::WithConstCapacity(_)
+            )
```
y21, line 79:
```suggestion
            && !matches!(init, VecInitKind::WithExprCapacity(_) | VecInitKind::WithConstCapacity(_))
```
Sadly rustfmt doesn't know about let chains yet so this formatting needs to be done manually for now😢

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+                space_hint: String::new()
+            });
+        }
+    }
+
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if self.searcher.is_none()
+            && let ExprKind::Assign(left, right, _) = expr.kind
+            && let ExprKind::Path(QPath::Resolved(None, path)) = left.kind
+            && let [_] = &path.segments
```
y21, line 96: Is this `let [_]` needed? Can there ever be more or less than one segment in a path to a local or what does this do?

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,139 @@
...
+        if let Some(searcher) = self.searcher.take() {
+            if let StmtKind::Expr(expr) | StmtKind::Semi(expr) = stmt.kind
+                && let ExprKind::MethodCall(name, self_arg, [space_hint], _) = expr.kind
+                && path_to_local_id(self_arg, searcher.local_id)
+                && !is_from_proc_macro(cx, expr)
+                && name.ident.as_str() == "reserve"
+            {
+                self.searcher = Some(VecReserveSearcher {
+                    err_span: searcher.err_span.to(stmt.span),
+                    space_hint: snippet(cx, space_hint.span, "..").to_string(),
```
y21, line 125:
```suggestion
                    space_hint: snippet(cx, space_hint.span, "..").into_owned(),
```
This allows it to reuse the string if owned (which afaik is basically always the case) and is consistent with the other usages of snippet here

y21 (commented):
Just the last few tiny nits I could find ^^. After these it looks good to me 👍

clippy_lints/src/reserve_after_initialization.rs
Red-Rapious: I previously used this `let` to retrieve the name of the first segment, but I think there's not use to check that `path.segments.len() == 1`; I'll remove the line.

clippy_lints/src/reserve_after_initialization.rs
```diff
@@ -0,0 +1,134 @@
...
+            });
+        }
+    }
+
+    fn check_stmt(&mut self, cx: &LateContext<'tcx>, stmt: &'tcx Stmt<'_>) {
+        if let Some(searcher) = self.searcher.take() {
+            if let StmtKind::Expr(expr) | StmtKind::Semi(expr) = stmt.kind
+                && let ExprKind::MethodCall(name, self_arg, [space_hint], _) = expr.kind
+                && path_to_local_id(self_arg, searcher.local_id)
+                && !is_from_proc_macro(cx, expr)
```
blyxyas, line 115: `is_from_proc_macro` should go last, as it's very expensive, there are some efforts on optimizing it, but for the moment, it should go last.

blyxyas (approved):
Just this mini-nit and it should be good to go! Really impressive for a first contribution :heart:

y21 (approved):
LGTM!

> your reviews were very helpful!

Thanks :D Glad to be of help

Red-Rapious:
Thank you so much for your help, @blyxyas and @y21!

Issue #11246, opened 2023-07-28
`undocumented_unsafe_blocks` false positive on `const`s


jhpratt:
### Summary

See repro, as it's self-explanatory.

### Lint Name

undocumented_unsafe_blocks

### Reproducer

```rust
#![deny(clippy::undocumented_unsafe_blocks)]

// Safety: foo
const _: () = unsafe {};
```

The lint fires, but it should not, as a safety comment is present on the preceding line. This is the case regardless of configuration.

### Version

```text
rustc 1.73.0-nightly (500647fd8 2023-07-27)
binary: rustc
commit-hash: 500647fd8138cc09e87edb08d62f81654fbf6ef8
commit-date: 2023-07-27
host: x86_64-unknown-linux-gnu
release: 1.73.0-nightly
LLVM version: 16.0.5
```

J-ZhengLi:
always wonder how this lint was able to check comments 🤔 , so I'm gonna try @rustbot claim

J-ZhengLi:
as it turns out, there was a similar issue (#10832) which got fixed by adding a configuration to this lint.

Try adding a `accept-comment-above-statement = true` to your clippy.toml.

jhpratt:
As I said in the original issue, this occurs regardless of configuration.

J-ZhengLi:
> As I said in the original issue, this occurs regardless of configuration.

oops, I didn't read it through, my bad

PR #11375, opened 2023-08-22
fix fp when [`undocumented_unsafe_blocks`] not able to detect comment on globally defined const/static variables

J-ZhengLi:
fixes: #11246

changelog: fix detection on global variables for [`undocumented_unsafe_blocks`]

### DIFF
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -341,44 +341,21 @@ fn block_parents_have_safety_comment(
     id: hir::HirId,
 ) -> bool {
     if let Some(node) = get_parent_node(cx.tcx, id) {
-        return match node {
-            Node::Expr(expr) => {
-                if let Some(
-                    Node::Local(hir::Local { span, .. })
-                    | Node::Item(hir::Item {
-                        kind: hir::ItemKind::Const(..) | ItemKind::Static(..),
-                        span,
-                        ..
-                    }),
-                ) = get_parent_node(cx.tcx, expr.hir_id)
-                {
-                    let hir_id = match get_parent_node(cx.tcx, expr.hir_id) {
-                        Some(Node::Local(hir::Local { hir_id, .. })) => *hir_id,
-                        Some(Node::Item(hir::Item { owner_id, .. })) => {
-                            cx.tcx.hir().local_def_id_to_hir_id(owner_id.def_id)
-                        },
-                        _ => unreachable!(),
-                    };
-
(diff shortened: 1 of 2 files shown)


clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -639,29 +615,42 @@ fn get_body_search_span(cx: &LateContext<'_>) -> Option<Span> {
...
     let mut span = map.body(body).value.span;
+    let mut maybe_global_var = false;
     for (_, node) in map.parent_iter(body.hir_id) {
         match node {
-            Node::Expr(e) => span = e.span,
+            Node::Expr(e) => {
+                span = e.span;
+                // Note: setting this to `false` is to making sure a "in-function defined"
+                // const/static variable not mistakenly processed as global variable,
+                // since global var doesn't have an `Expr` parent as its parent???
```
J-ZhengLi, line 625: I'm 70% sure of this... so, could someone confirm this for me? 😆

Centri3: I *think* this is because a `const`'s initializer is a body, and when getting a body's `value`'s parent it'll get wherever the body's used; see the HIR for `const A: u32 = { 0 }`: https://rust.godbolt.org/z/3cY4vaq6T

So this is getting the `Expr` every other time, yet here, it's getting the `Const`/`Static`, afaik?

Centri3:
Interestingly, `unnecessary_safety_comment` doesn't have the same issue despite mostly sharing the same code. I'll see if we can replicate what it does since it looks like there is an issue in `block_has_safety_comment` or `block_parents_have_safety_comment` that `stmt_has_safety_comment` doesn't have.

J-ZhengLi:
> Interestingly, `unnecessary_safety_comment` doesn't have the same issue despite mostly sharing the same code. I'll see if we can replicate what it does since it looks like there is an issue in `block_has_safety_comment` or `block_parents_have_safety_comment` that `stmt_has_safety_comment` doesn't have.

it is the `get_body_search_span` function that's causing troubles, I think...

clippy_lints/src/undocumented_unsafe_blocks.rs
J-ZhengLi: wait a minute, I think unintentionally increase the search span for other scenarios, let me see if I can find a better solution~

Centri3:
This version looks fine to me. I don't particularly like `maybe_global_var` but ah well; I have nothing to suggest over it.

@bors r+

PR #11385, opened 2023-08-23
skip float_cmp check if lhs is a custom type

markhuang1212:

changelog: [`float_cmp`]: allow float eq comparison when lhs is a custom type that implements PartialEq<f32/f64>

If the lhs of a comparison is not float, it means there is a user implemented PartialEq, and the caller is invoking that custom version of `==`, instead of the default floating point equal comparison. 

People may wrap f32 with a struct (say `MyF32`) and implement its PartialEq that will do the `is_close()` check, so that `MyF32` can be compared with either f32 or `MyF32`.

### DIFF
--- a/clippy_lints/src/operators/float_cmp.rs
+++ b/clippy_lints/src/operators/float_cmp.rs
@@ -17,7 +17,7 @@ pub(crate) fn check<'tcx>(
     left: &'tcx Expr<'_>,
     right: &'tcx Expr<'_>,
 ) {
-    if (op == BinOpKind::Eq || op == BinOpKind::Ne) && (is_float(cx, left) || is_float(cx, right)) {
+    if (op == BinOpKind::Eq || op == BinOpKind::Ne) && is_float(cx, left) {
         let left_is_local = match constant_with_source(cx, cx.typeck_results(), left) {
             Some((c, s)) if !is_allowed(&c) => s.is_local(),
             Some(_) => return,
--- a/tests/ui/float_cmp.rs
+++ b/tests/ui/float_cmp.rs
@@ -41,12 +41,25 @@ impl PartialEq for X {
     }
 }
 
+impl PartialEq<f32> for X {
+    fn eq(&self, o: &f32) -> bool {
+        if self.val.is_nan() {
+            o.is_nan()
+        } else {
+            self.val == *o // no error, inside "eq" fn
+        }
+    }
+}
+
 fn main() {
     ZERO == 0f32; //no error, comparison with zero is ok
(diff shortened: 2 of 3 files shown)


blyxyas:
This would fail if the `lhs` doesn't implement `PartialEq`, so we should use the [`implements_trait`](https://doc.rust-lang.org/beta/nightly-rustc/clippy_utils/ty/fn.implements_trait.html) function if the `lhs` implements it.

[You can check this example for reference](https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/common_tools_writing_lints.md#checking-if-a-type-implements-a-specific-trait), we should also add a test where the `lhs` doesn't implement `PartialEq<f32|f64>` to check that it doesn't lint.

markhuang1212:
Not sure I understand this correctly, we already checked that lhs is a float before we enter the lint, if lhs doesn't implement PartialEq, nothing should happen. Before this pr lint will be performed if either lhs or rhs is float

blyxyas:
Oh I get it now! If the second value doesn't implement PartialEq, it will have a compiler error before linting. Right?

markhuang1212:
> Oh I get it now! If the second value doesn't implement PartialEq, it will have a compiler error before linting. Right?

Yes!

blyxyas (approved):
Well, I think this is good then :heart:.
Could you squash these 2 commits into a single one?

blyxyas:
*meow* :cat:
<!-- @bors r+ --!>

markhuang1212:
@blyxyas I think this pr might caused some confusion due to a previous typo in the title. Anyway, my motivation is this: in a project I'm working on, we use a custom type to wrap f64 to add some functionality, which include a `impl PartialEq<f64>` that returns true if `(a-b).abs() < episolon`, and when I'm using it, it triggers a warning. From my understanding, if people is intentionally invoking a custom PartiqlEq, they are trying to acheive something that they have customized, and therefore should not be warned against. The rationale of this change is somewhat similar to the existing feature where the check is not enforced if the function name containes `eq`.

blyxyas:
I'll do some more testing on this patch. I'm not sure it really is caused because of this `if` statement. The logic would be the same as the `||` is inclusive. If `is_float(cx, left)` is `true`, then it doesn't matter the result `is_float(cx, right)`. idk :woman_shrugging:

blyxyas:
Mmmmmm... I just checked it, it **does** change the logic. I find it pretty weird that this inclusive operator changes the outcome in exclusive-like behaviours.

LGTM! Thanks :heart: 
@bors r+

PR #11387, opened 2023-08-23
[`unnecessary_unwrap`]: lint on `.as_ref().unwrap()`

### LINKED ISSUE #11371: `unnecessary_unwrap` doesn't work with `.as_ref().unwrap()` and `.as_mut().unwrap()`
### Summary

The lint should recognize `Option::as_ref` and `Option::as_mut` methods and suggest code with `if let`, matching on `&foo`/`&mut foo`, or alternatively with `ref`/`ref mut` in the pattern.

### Lint Name

unnecessary_unwrap

### Reproducer

I tried this code:

```rust
let a = Some(1);
if a.is_some() {
    let x = a.as_ref().unwrap();
    println!("{x}");
}
```

The lint would trigger and suggest the following code:
```rust
let a = Some(1);
if let Some(x) = &a {
    println!("{x}");
}
```

Instead, the lint didn't trigger.

### Version

```text
0.1.73 (2023-08-20 5c6a7e7) on playground
```

y21:
Closes #11371 

This turned out to be a little more code than I originally thought, because the lint also makes sure to not lint if the user tries to mutate the option:
```rs
if option.is_some() {
  option = None;
  option.unwrap(); // don't lint here
}
```
... which means that even if we taught this lint to recognize `.as_mut()`, it would *still* not lint because that would count as a mutation. So we need to allow `.as_mut()` calls but reject other kinds of mutations.
Unfortunately it doesn't look like this is possible with `is_potentially_mutated` (seeing what kind of mutation happened).
This replaces it with a custom little visitor that does basically what it did before, but also allows `.as_mut()`.

changelog: [`unnecessary_unwrap`]: lint on `.as_ref().unwrap()`

### DIFF
--- a/clippy_lints/src/unwrap.rs
+++ b/clippy_lints/src/unwrap.rs
@@ -1,15 +1,18 @@
 use clippy_utils::diagnostics::span_lint_hir_and_then;
 use clippy_utils::ty::is_type_diagnostic_item;
-use clippy_utils::usage::is_potentially_mutated;
+use clippy_utils::usage::is_potentially_local_place;
 use clippy_utils::{higher, path_to_local};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_expr, walk_fn, FnKind, Visitor};
-use rustc_hir::{BinOpKind, Body, Expr, ExprKind, FnDecl, HirId, PathSegment, UnOp};
+use rustc_hir::{BinOpKind, Body, Expr, ExprKind, FnDecl, HirId, Node, PathSegment, UnOp};
+use rustc_hir_typeck::expr_use_visitor::{Delegate, ExprUseVisitor, PlaceWithHirId};
+use rustc_infer::infer::TyCtxtInferExt;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::hir::nested_filter;
 use rustc_middle::lint::in_external_macro;
-use rustc_middle::ty::Ty;
+use rustc_middle::mir::FakeReadCause;
+use rustc_middle::ty::{self, Ty, TyCtxt};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::def_id::LocalDefId;
 use rustc_span::source_map::Span;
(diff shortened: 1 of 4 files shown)


blyxyas:
> This turned out to be a little more code than I originally thought, because the lint also makes sure to not lint if the user tries to mutate the option

Wouldn't that also be... an ***unnecessary unwrap***?

y21:
Oh true 😄 that example might've been confusing... I suppose
```rs
if option.is_some() {
  option = calculate_option(); // some kind of expression mutating option
  option.unwrap();
}
```
would be more realistic. Suggesting to change that to an if let wouldnt preserve semantics since that mutation may change it to `None`

clippy_lints/src/unwrap.rs
```diff
@@ -192,6 +195,43 @@ fn collect_unwrap_info<'tcx>(
...
 
+/// A HIR visitor delegate that checks if a local variable of type `Option<_>` is mutated,
+/// unless `Option::as_mut` is called.
+struct MutationVisitor<'tcx> {
+    is_mutated: bool,
+    local_id: HirId,
+    tcx: TyCtxt<'tcx>,
+}
+
+fn is_option_as_mut_use(tcx: TyCtxt<'_>, expr_id: HirId) -> bool {
```
blyxyas, line 206: Could you document this function, even a single line will make it easier to understand.

clippy_utils/src/usage.rs
```diff
@@ -37,6 +37,17 @@ pub fn is_potentially_mutated<'tcx>(variable: HirId, expr: &'tcx Expr<'_>, cx: &
...
+    match place.base {
+        PlaceBase::Local(id) => id == local_id,
+        PlaceBase::Upvar(_) => {
+            // Conservatively assume yes.
+            true
+        },
+        _ => false,
+    }
+}
+
```
blyxyas, line 50: Seems like this wouldn't support static variables? Could you add a test for `static` places?

blyxyas (commented):
Wow, it really is a bit of code, but it seems really good. Just a couple questions :cat:

clippy_utils/src/usage.rs
y21: Wdym by this? If the passed place is a static variable, then it sounds right that this should return false, since it's not a local place. It's ok for this lint specifically to only consider locals and ignore statics, because it doesn't/didn't look for statics anyway. That is, this already doesn't lint:
```rs
fn f() {
  static X: Option<i32> = Some(123);

  if X.is_some() {
    X.unwrap();
  }
}
```
so there's no way to really get FPs here if this returns `false` for a static place there I think.
I suppose it would still be worth having a test that makes sure that something like
```rs
static mut X: Option<i32> = Some(123);
if X.is_some() {
  X = None;
  X.unwrap();
}
```
doesn't lint

clippy_lints/src/unwrap.rs
y21: Added some more documentation, that should hopefully clarify it a bit. Is is this better?

clippy_utils/src/usage.rs
blyxyas: It would be nice if we could also lint static places, although calling unwrap on an static variable isn't common (nor a good practice). Why are not linting on static places, just because of backwards compatbility?

y21: I'm not sure why it doesn't lint on those, it doesn't mention that anywhere in the lint as far as I can see. Maybe it was just forgotten as it's probably very uncommon to do this on statics like you said. I think it would be too big of a change for this PR to make that work imo. There seem to be quite a few places in this lint that assume it's a local

clippy_lints/src/unwrap.rs
blyxyas: Yep, it's better ❤️

blyxyas (approved):
I think this is pretty good to go then, thanks! :heart:

y21:
Looks like a spurious error?

blyxyas:
Looks like one :3
@bors r+

PR #11391, opened 2023-08-24
fix the uitest `enum_clike_unportable_variant`

mojave2:

changelog: none


flip1995:
Thanks!

I just noticed that this test case is ignored on x86. This means that we never run it on CI, because there we only have x86 machines. :thinking:

Alexendoo:
I think that's because ignoring x86 includes x86_64 in ui_test but didn't in compiletest, `//@ignore-32bit` is probably what it wants to be

Alexendoo:
Thanks! @bors r+

PR #11396, opened 2023-08-24
new lint: `iter_out_of_bounds`

### LINKED ISSUE #11345: Lint when `.take(x)` or `.skip(x)` is out of bounds for an array
### What it does

`.take()` and `.skip()` (gracefully) don't panic when doing something like `[1, 2, 3].iter().skip(4)`. We should lint this and say that it makes no sense since it'll either return the full iterator or an empty iterator.

### Advantage

- The code is useless in both scenarios

### Drawbacks

_No response_

### Example

```rust
for _ in [1, 2, 3].iter().skip(4) {
    unreachable!();
}
for (i, _) in [1, 2, 3].iter().take(4).enumerate() {
    assert!(i <= 2);
}
```

Could be written as:

y21:
Closes #11345

The original idea in the linked issue seemed to be just about arrays afaict, but I extended this to catch some other iterator sources such as `iter::once` or `iter::empty`.

I'm not entirely sure if this name makes a lot of sense now that it's not just about arrays anymore (specifically, not sure if you can call `.take(1)` on an `iter::Empty` to be "out of bounds"?).

changelog: [`iter_out_of_bounds`]: new lint

### DIFF
--- a/clippy_lints/src/methods/iter_out_of_bounds.rs
+++ b/clippy_lints/src/methods/iter_out_of_bounds.rs
@@ -1,5 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_note;
-use clippy_utils::{is_trait_method, match_def_path, paths};
+use clippy_utils::higher::VecArgs;
+use clippy_utils::{expr_or_init, is_trait_method, match_def_path, paths};
 use rustc_ast::LitKind;
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::LateContext;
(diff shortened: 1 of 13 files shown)


clippy_lints/src/methods/iter_out_of_bounds.rs
```diff
@@ -0,0 +1,91 @@
...
+        } else if match_def_path(cx, did, &paths::SLICE_ITER)
+            && let ExprKind::MethodCall(_, recv, ..) = iter.kind
+            && let ExprKind::Array(array) = recv.peel_borrows().kind
+        {
+            // For slice::Iter<'_, T>, the receiver might be an array literal: [1,2,3].iter().skip(..)
+            array.len().try_into().ok()
+        } else if match_def_path(cx, did, &paths::ITER_EMPTY) {
+            Some(0)
+        } else if match_def_path(cx, did, &paths::ITER_ONCE) {
+            Some(1)
```
Centri3, line 34: Maybe we can lint `vec!` as well? There is the `VecArgs` utils type and, since we have the `iter` expr, I *think* we can use that (plus an `expr_or_init` call)

clippy_lints/src/methods/iter_out_of_bounds.rs
```diff
@@ -0,0 +1,91 @@
...
+use rustc_middle::ty::{self};
+use rustc_span::sym;
+
+use super::ITER_OUT_OF_BOUNDS;
+
+/// Attempts to extract the length out of an iterator expression.
+fn get_iterator_length<'tcx>(cx: &LateContext<'tcx>, iter: &'tcx Expr<'tcx>) -> Option<u128> {
+    let iter_ty = cx.typeck_results().expr_ty(iter);
+
+    if let ty::Adt(adt, substs) = iter_ty.kind() {
```
Centri3, line 15: nit: This could probably be `let-else`.

clippy_lints/src/methods/iter_out_of_bounds.rs
```diff
@@ -0,0 +1,91 @@
...
+    recv: &'tcx Expr<'tcx>,
+    arg: &'tcx Expr<'tcx>,
+    message: &'static str,
+    note: &'static str,
+) {
+    if is_trait_method(cx, expr, sym::Iterator)
+        && let Some(len) = get_iterator_length(cx, recv)
+        && let ExprKind::Lit(lit) = arg.kind
+        && let LitKind::Int(skip, _) = lit.node
+        && skip > len
```
Centri3, line 55:
```suggestion
        && let LitKind::Int(skipped, _) = lit.node
        && skipped > len
```
`skipped` is probably misleading considering this is also called for `take()` but this should be ok.

Centri3: We should also call `expr_or_init` here as well

clippy_lints/src/methods/iter_out_of_bounds.rs
Centri3: nit: I think we can also call `ty_adt_def` instead.

`peel_refs` is also probably ok too, but I don't think there are many cases where that'll even come up if it changes anything (why would somebody do `(&[1, 2, 3].iter()).skip(x)`?)

clippy_lints/src/methods/mod.rs
```diff
@@ -3538,6 +3539,30 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Example
+    /// ```rust
+    /// for _ in [1, 2, 3].iter().take(4) {}
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// for _ in [1, 2, 3].iter() {}
+    /// ```
+    #[clippy::version = "1.73.0"]
```
Centri3, line 3560: Needs to be bumped, sync just happened and we're on 1.74.0 now

tests/ui/iter_out_of_bounds.rs
Centri3, line 1: Can you add a test for something like `x.iter().skip(y)` as well?

clippy_lints/src/methods/iter_out_of_bounds.rs
y21: > nit: I think we can also call `ty_adt_def` instead.

I don't think so, since we need the generic args, and we don't get those with `ty_adt_def`

Centri3: Ah, missed that that was used. That's fine then

clippy_lints/src/methods/iter_out_of_bounds.rs
```diff
@@ -0,0 +1,106 @@
...
+    arg: &'tcx Expr<'tcx>,
+    message: &'static str,
+    note: &'static str,
+) {
+    if is_trait_method(cx, expr, sym::Iterator)
+        && let Some(len) = get_iterator_length(cx, recv)
+        && let Some(skipped) = expr_as_u128(cx, arg)
+        && skipped > len
+    {
+        span_lint_and_note(cx, ITER_OUT_OF_BOUNDS, expr.span, message, None, note);
```
Jarcho, line 72: This really should have a `MaybeIncorrect` suggestion added to it.

Jarcho (commented):
This should have some more negative tests added. e.g. `[1, 2, 3].iter().take(2)`

clippy_lints/src/methods/iter_out_of_bounds.rs
y21: I'm not sure what the suggestion should be or if it makes sense to have one. At least for something like `[1,2,3].iter().take(4)` it could probably suggest removing the `.take()` call as that'll still yield the same items with or without it, but I don't know what the user could have meant with `[1,2,3].iter().skip(4)`. Should it suggest and empty iterator? 🤔

Jarcho: I was only considering `take` here. Given the suggestion isn't going to have a lot of value (decent chance of being the wrong fix), I guess it's fine to not have one.

Jarcho:
Thank you. @bors r+

PR #11401, opened 2023-08-24
[`if_then_some_else_none`]: look into local initializers for early returns

### LINKED ISSUE #11394: False positive with ? in `clippy::if_then_some_else_none`
### Summary

Probably related to #7870, if I remove `let _ = ` then the warning disappears

### Lint Name

clippy::if_then_some_else_none

### Reproducer

I tried this code:

```rust
#![warn(clippy::if_then_some_else_none)]

fn f(b: bool, v: Result<(), ()>) -> Result<(), ()> {
    let x = if b {
        let _ = v?;
        Some(())
    } else {
        None
    };

    Ok(())
}
```

I saw this happen:

```
warning: this could be simplified with `bool::then`
```

I expected to see this happen: No warning

### Version

```text
rustc 1.73.0-nightly (1b198b3a1 2023-08-13)
binary: rustc
commit-hash: 1b198b3a196442e14fb06978166ab46a4618d131
commit-date: 2023-08-13
host: x86_64-pc-windows-msvc
release: 1.73.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

_No response_

y21:
Fixes #11394

As the PR title says, problem was that it only looked for early returns in semi statements. Local variables don't count as such, so it didn't count `let _v = x?;` (or even just `let _ = return;`) as a possible early return and didn't realize that it can't lint then.

Imo the `stmts_contains_early_return` function that was used before is redundant. `contains_return` could already do that if we just made the parameter a bit more generic, just like `for_each_expr`, which can already accept `&[Stmt]`

changelog: [`if_then_some_else_none`]: look into local initializers for early returns

### DIFF
--- a/clippy_lints/src/if_then_some_else_none.rs
+++ b/clippy_lints/src/if_then_some_else_none.rs
@@ -6,7 +6,7 @@ use clippy_utils::sugg::Sugg;
 use clippy_utils::{contains_return, higher, is_else_clause, is_res_lang_ctor, path_res, peel_blocks};
 use rustc_errors::Applicability;
 use rustc_hir::LangItem::{OptionNone, OptionSome};
-use rustc_hir::{Expr, ExprKind, Stmt, StmtKind};
+use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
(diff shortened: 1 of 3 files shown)


xFrednet (approved):

xFrednet:
LGTM, and I agree that the `stmts_contains_early_return` seems redundant and error prune. Thank you for the nice fix :D

@bors r+

Issue #11402, opened 2023-08-25
False Positive: needless_raw_string_hashes


FintanH:
### Summary

On our `radicle-surf` project, the lint job failed due to this clippy warning. This is the [job](https://github.com/radicle-dev/radicle-git/actions/runs/5965259965/job/16182278180?pr=139) and the offending code is:

```rust
#[test]
fn test_both_missing_eof_newline() {
    let buf = r#"
diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=123
\ No newline at end of file
+hello=1234
\ No newline at end of file
"#;
    let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
    let diff = Diff::try_from(diff).unwrap();
    assert_eq!(
        diff.modified().next().unwrap().diff.eof(),
        Some(EofNewLine::BothMissing)
    );
}
```

What's strange is there is a very similar test below that does not set off the clippy warning:
```rust
#[test]
fn test_none_missing_eof_newline() {
    let buf = r#"
diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=123
+hello=1234
"#;
    let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
    let diff = Diff::try_from(diff).unwrap();
    assert_eq!(
        diff.modified().next().unwrap().diff.eof(),
        Some(EofNewLine::NoneMissing)
    );
}
```

### Lint Name

needless_raw_string_hashes

### Reproducer

I tried this code:

```rust
#[test]
fn test_both_missing_eof_newline() {
    let buf = r#"
diff --git a/.env b/.env
index f89e4c0..7c56eb7 100644
--- a/.env
+++ b/.env
@@ -1 +1 @@
-hello=123
\ No newline at end of file
+hello=1234
\ No newline at end of file
"#;
    let diff = git2::Diff::from_buffer(buf.as_bytes()).unwrap();
    let diff = Diff::try_from(diff).unwrap();
    assert_eq!(
        diff.modified().next().unwrap().diff.eof(),
        Some(EofNewLine::BothMissing)
    );
}
```

I saw this happen:

```
error: unnecessary hashes around raw string literal
   --> radicle-surf/t/src/diff.rs:363:15
    |
363 |       let buf = r#"
    |  _______________^
[...]

PR #11415, opened 2023-08-27
Move `needless_raw_string_hashes` to `pedantic`

Alexendoo:
IMO it doesn't improve code enough to be warn by default. [It seems to be unclear to some also](https://github.com/rust-lang/rust-clippy/issues/11402), but that can probably be remedied separately

changelog: Moved [`needless_raw_string_hashes`] to `pedantic` (Now allow-by-default)
[#11415](https://github.com/rust-lang/rust-clippy/pull/11415)

r? @flip1995

### DIFF
--- a/clippy_lints/src/raw_strings.rs
+++ b/clippy_lints/src/raw_strings.rs
@@ -49,7 +49,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.72.0"]
     pub NEEDLESS_RAW_STRING_HASHES,
-    style,
+    pedantic,
     "suggests reducing the number of hashes around a raw string literal"
 }
 impl_lint_pass!(RawStrings => [NEEDLESS_RAW_STRINGS, NEEDLESS_RAW_STRING_HASHES]);


Centri3 (approved):

Centri3:
I'm fine with pedantic. Not sure what's wrong in #11402 though, as removing hashes should make no difference unless there's a `"` with that many or more hashes following it.

Alexendoo:
My guess is people are removing the `r` as well, I opened https://github.com/rust-lang/rust-clippy/pull/11416 but didn't link the issue as being fixed by it since I think there's further improvements to be made/maybe there's something we're missing

flip1995:
@bors r+

Reading through the issues, I figure this is required/wanted.

flip1995:
I'm leaning towards letting this change ride the release train without a backport. This lint is already in stable and I don't think the FN/confusion is so bad that we need a beta backport for this. However I'll cherry-pick this commit, so that it at least get's into 1.74.

flip1995:
https://github.com/rust-lang/rust/pull/116245

PR #11416, opened 2023-08-27
Use multipart suggestions for raw string lints

Alexendoo:
Should make it slightly easier to see the suggested edit

Before/after for `needless_raw_string_hashes`:

| Before| After |
|--------|--------|
| before | after | 

changelog: none

### DIFF
--- a/tests/ui/needless_raw_string.rs
+++ b/tests/ui/needless_raw_string.rs
@@ -9,8 +9,13 @@ fn main() {
     br#"aaa"#;
     br#""aaa""#;
     br#"\s"#;
-    // currently disabled: https://github.com/rust-lang/rust/issues/113333
-    // cr#"aaa"#;
-    // cr#""aaa""#;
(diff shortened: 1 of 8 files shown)


tests/ui/needless_raw_string.rs
```diff
@@ -9,8 +9,7 @@ fn main() {
     br#"aaa"#;
```
xFrednet, line 9: All these tests seem to be single line examples. Could you add a test where the string spans multiple lines?

xFrednet (approved):
LGTM, one small nit, then we can r+ this :)

tests/ui/needless_raw_string.rs
Alexendoo: 👍 added

xFrednet (approved):

xFrednet:
LGTM, thank you for the update :D

@bors r+

PR #11419, opened 2023-08-28
fix: needless_range_loop's FP on the `Field` expr

### LINKED ISSUE #11399: False positive: needless_range_loop
### Summary

Related code:
```rust
for i in 0..message.len() {
        block.data[i] = message[i] as char;
 }
 ```

Clippy begs me to use ```rust
for x in data```
without the 'i' enumerator, however this would make fetching message[i] impossible.

### Lint Name

needless_range_loop

### Reproducer

I tried this code:

```rust
for i in 0..message.len() {
        block.data[i] = message[i] as char;
}
```

I saw this happen:

```
the loop variable `i` is used to index `message` Help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_range_loop Help: consider using an iterator and enumerate()
```

no complaints from Clippy

### Version

```text
rustc 1.72.0-nightly (8c74a5d27 2023-06-14)
binary: rustc
commit-hash: 8c74a5d27c644a0f7a22bb2fa8dd3ff8257bc220
commit-date: 2023-06-14
host: x86_64-pc-windows-msvc
release: 1.72.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

_No response_

kiscad:

fix #11399 

changelog: [`needless_range_loop`]: Fix the FP warning when the index-var is used in another "field" sequence.

### DIFF
--- a/tests/ui/needless_range_loop2.rs
+++ b/tests/ui/needless_range_loop2.rs
@@ -117,3 +117,17 @@ mod issue2277 {
         }
     }
 }
+
+mod issue11399 {
+    pub struct Block {
+        data: [i32; 3],
+    }
+
+    pub fn example() {
+        let values: [i8; 3] = [1, 2, 3];
+        let mut block = Block { data: [0; 3] };
+        for i in 0..values.len() {
+            block.data[i] = values[i] as i32;
+        }
+    }
+}
--- a/clippy_lints/src/loops/needless_range_loop.rs
+++ b/clippy_lints/src/loops/needless_range_loop.rs
@@ -295,6 +295,11 @@ impl<'a, 'tcx> VarVisitor<'a, 'tcx> {
                 }
             }
         }
+
+        // check whether the struct/tuple is referenced by a name, when the seqexpr is a field expr.
+        if let ExprKind::Field(sq_expr, _) = seqexpr.kind {
+            return self.check(idx, sq_expr, expr);
+        }
         true
     }
 }


tests/ui/needless_range_loop2.rs
```diff
@@ -117,3 +117,17 @@ mod issue2277 {
...
+    pub struct Block {
+        data: [i32; 3],
+    }
+
+    pub fn example() {
+        let values: [i8; 3] = [1, 2, 3];
+        let mut block = Block { data: [0; 3] };
+        for i in 0..values.len() {
+            block.data[i] = values[i] as i32;
+        }
```
flip1995, line 131: This could be written as

```rust
for (i, value) in values.iter().enumerate() {
     block.data[i] = value as i32;
}
```

Which is more idiomatic IMO and I think the lint already suggests to do. 

Even better might be to suggest using `zip`, but that might change semantics without length information. So I wouldn't go that route.

flip1995: Yeah, they should be treated the same. Now I'm wondering why `data[i]` is not linted :thinking: 

But in this case, I think the fix is incomplete. What about a case like this:

```rust
for i in 0..values.len() {
    some_func()[i] = values[i] as i32;
}
```

Or any other expression that returns something that can be indexed and assigned to. Is the lint triggering for those as well?

mojave2: It looks that the indirect indexing means indexing any expression containing the index pat-var, like `[i + 1]`, other than `[i]`.

flip1995: Ah thanks for the clarification! Then I think there are 2 proper ways to fix this:

1. Make it lint less by just searching for `ExprKind::Index` and check if `i` is used in the index expression and if it is used more than once, don't lint
2. Make it lint more by doing the same check, but then suggest to use `enumerate`.

I think for this PR option 1 is the better / quicker one to fix the issue. Option 2 is more a long term thing.

xFrednet:
Hey, this is a ping from triage. @flip1995 can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@rustbot ready

xFrednet:
It looks like `flip1995` is currently busy. Sorry for the delay, let's pick a new reviewer. 

r? clippy

flip1995:
So sorry that I didn't get to this. My comment here: https://github.com/rust-lang/rust-clippy/pull/11419#discussion_r1332746225 wasn't addressed yet. I think it's good that some other reviewer also takes a look at this. Maybe my understanding here is wrong.

xFrednet:
Hey @kiscad, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

@rustbot author

samueltardieu:
No activity for more than one year.

PR #11427, opened 2023-08-30
Bump ui_test

oli-obk:
This makes `ui_test` parse `--bless` and allows a follow up change to use `Mode::Error` (instead of `Mode::Yolo`) with `RustfixMode::Everything`

changelog: none

### DIFF
--- a/tests/compile-test.rs
+++ b/tests/compile-test.rs
@@ -114,20 +114,8 @@ fn canonicalize(path: impl AsRef<Path>) -> PathBuf {
 }
 
 fn base_config(test_dir: &str) -> (Config, Args) {
-    let bless = var_os("RUSTC_BLESS").is_some_and(|v| v != "0") || env::args().any(|arg| arg == "--bless");
-
-    let args = Args {
-        filters: env::var("TESTNAME")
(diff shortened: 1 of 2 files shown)


Alexendoo:
Opened https://github.com/oli-obk/ui_test/pull/157 to get `OutputConflictHandling::Error` back, particularly for the internal tests where the command is difficult to remember

tests/compile-test.rs
```diff
@@ -302,8 +291,7 @@ fn run_ui_cargo() {
 
     ui_test::run_tests_generic(
         vec![config],
-        args,
-        |path, args, _config| path.ends_with("Cargo.toml") && ui_test::default_filter_by_arg(path, args),
+        |path, config| path.ends_with("Cargo.toml") && ui_test::default_file_filter(path, config),
```
Alexendoo, line 294: This ought to be `default_any_file_filter` (or whatever I called it), the one that doesn't require `.rs`

oli-obk: whoops fixed

Alexendoo:
Thanks! @bors r+

PR #11444, opened 2023-09-01
Return a value from find_format_args instead of using a callback

Alexendoo:
r? @flip1995

changelog: none

### DIFF
--- a/clippy_lints/src/explicit_write.rs
+++ b/clippy_lints/src/explicit_write.rs
@@ -57,54 +57,52 @@ impl<'tcx> LateLintPass<'tcx> for ExplicitWrite {
             } else {
                 None
             }
+            && let Some(format_args) = find_format_args(cx, write_arg, ExpnId::root())
         {
-            find_format_args(cx, write_arg, ExpnId::root(), |format_args| {
-                let calling_macro =
-                    // ordering is important here, since `writeln!` uses `write!` internally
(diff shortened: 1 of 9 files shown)


flip1995:
I haven't reviewed the PR yet. But I don't like the doc comment of `Symbol::as_str`

> in practice it works out ok.

I'll have to look more closely on this PR though, before making a decision.

Alexendoo:
Assuming I've done it correctly it should be less of a crime than `Symbol::as_str`, since it's tied to a lifetime that's shorter than the thread local

Jarcho:
There's nothing that actually stops a `&'static LateContext<'_>` value from existing. `Box::leak` is the easy way to do it, but there are others. `Symbol::as_ref` is unsafe for exactly the same reason as there's nothing stopping `&'static Symbol` from existing.

Neither of these are really a problem in practice, but there isn't a lifetime you can extend to that isn't technically unsafe.

Alexendoo:
Yeah that's true, I've updated the comment a bit

Alexendoo:
I thought about this again today and realised we could stick the `FormatArgs` in an `Rc`, which means it no longer needs any unsafe code since it's not borrowing from the thread local. Updated the PR to use that

flip1995 (approved):
Wow, turns out if you wait long enough, PR reviews resolve themselves. Sorry for taking so long.

Since this PR doesn't add unsafe code now, it was an easy review.

Thanks!

PR #11448, opened 2023-09-02
DefaultUnionRepresentation: explain why we only warn about unions with at least 2 non-ZST fields

RalfJung:
changelog: none


blyxyas (approved):
LGTM, Thanks! :heart:

PR #11450, opened 2023-09-02
`never_loop` catches `loop { panic!() }`

digama0:
* Depends on: #11447 

This is an outgrowth of #11447 which I felt would best be done as a separate PR because it yields significant new results.

This uses typecheck results to determine divergence, meaning we can now detect cases like `loop { std::process::abort() }` or `loop { panic!() }`. A downside is that `loop { unimplemented!() }` is also being linted, which is arguably a false positive. I'm not really sure how to check this from HIR though, and it seems best to leave this epicycle for a later PR.

changelog: [`never_loop`]: Now lints on `loop { panic!() }` and similar constructs

### DIFF
--- a/clippy_lints/src/loops/never_loop.rs
+++ b/clippy_lints/src/loops/never_loop.rs
@@ -1,6 +1,5 @@
 use super::utils::make_iterator_snippet;
 use super::NEVER_LOOP;
-use clippy_utils::consts::{constant, Constant};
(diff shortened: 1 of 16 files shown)


llogiq:
I think finding `unimplemented!()` from the expansion chain should be possible, but I think that's a niche case, so fixing it in a later PR is fine.

PR #11451, opened 2023-09-02
[`slow_vector_initialization`]: use the source span of vec![] macro and fix another FP

### LINKED ISSUE #11408: slow_vector_initialization emits an internal warning ("rustc_errors::emitter Invalid span")
### Summary

I see this warning message printed:

```
WARN rustc_errors::emitter Invalid span /rustc/58eefc33adf769a1abe12ad94b3e6811185b4ce5/library/alloc/src/macros.rs:44:36: 44:59 (#4), error=SourceNotAvailable { filename: Real(Remapped { local_path: None, virtual_name: "/rustc/58eefc33adf769a1abe12ad94b3e6811185b4ce5/library/alloc/src/macros.rs" }) }
```

(This is also confusing my tooling that uses JSON output format since this line is not JSON :p)

### Reproducer

I tried [this code](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021&gist=c07cd6613fccb2ef29d5cca88adf9197):

```rust
pub fn wat() {
    let mut contents = vec![];
    contents.resize(123, 0);
}
```

In addition to a (correct) lint message, it produces the warning.

### Version

```text
1.74.0-nightly (2023-08-24 58eefc33adf769a1abe1)

(per playground, I didn't test locally)
```

### Additional Labels

_No response_

y21:
Fixes #11408

<details>
<summary>Also fixes a FP when the vec initializer comes from a macro other than `vec![]`</summary>

```rs
macro_rules! x {
  () => { vec![] }
}
fn f() {
  let mut v = x!();
  v.resize(10, 0);
}
```
This shouldn't warn. The `x!` macro might be doing other things, so just replacing `x!()` with `vec![0; 10]` is not always an option.
</details>

I added some test cases for macro expansions, however I don't think there's a way to write a test for that specific warning that appeared in the linked issue. As far as I understand, that happens when the rust-src rustup component isn't installed (so the stdlib source is unavailable) and the span points to the `vec![]` *expansion*, instead of the `vec![]` that the user wrote.

changelog: [`slow_vector_initialization`]: use the source span of `vec![]` macro
changelog: [`slow_vector_initialization`]: only warn on `vec![]` expansions and allow other macros

### DIFF
--- a/clippy_lints/src/slow_vector_initialization.rs
+++ b/clippy_lints/src/slow_vector_initialization.rs
@@ -1,4 +1,5 @@
 use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::macros::root_macro_call;
 use clippy_utils::sugg::Sugg;
 use clippy_utils::{
     get_enclosing_block, is_expr_path_def_path, is_integer_literal, is_path_diagnostic_item, path_to_local,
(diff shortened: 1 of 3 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update!

@bors r+

PR #11452, opened 2023-09-02
[`len_without_is_empty`]: follow type alias to find inherent `is_empty` method

### LINKED ISSUE #11165: clippy::len_without_is_empty bug when implementing on a type alias
### Summary

When a type alias is created, say `type B = A;` and both `len` and `is_empty` are implemented on `B`, clippy still complain that type `B` has a public `len` method, but no `is_empty` method  .

### Reproducer

I tried this code:

```rust
pub struct A;
pub type B = A;

impl B {
    /// Common method
    pub fn len(&self) -> usize {
        todo!()
    }

    pub fn is_empty(&self) -> bool {
        todo!()
    }
}
```

I expected to see this happen: No warnings, specifically, `clippy::len_without_is_empty`, is emitted.

Instead, this happened: Clippy returns the warning  `clippy::len_without_is_empty`.

Clippy version:

```shell
$ cargo clippy --version               
clippy 0.1.72 (ad96323 2023-07-14)
```

Result of running `cargo clippy`:

```shell
$ cargo clippy          
warning: type `B` has a public `len` method, but no `is_empty` method
 --> src/lib.rs:6:5
  |
6 |     pub fn len(&self) -> usize {
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#len_without_is_empty
  = note: `#[warn(clippy::len_without_is_empty)]` on by default

warning: `rust-doc-repro` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
  ```

### Version

```text
rustc 1.73.0-nightly (ad963232d 2023-07-14)
binary: rustc
commit-hash: ad963232d9b987d66a6f8e6ec4141f672b8b9900
commit-date: 2023-07-14
release: 1.73.0-nightly
LLVM version: 16.0.5
```

### Additional Labels

_No response_

y21:
Fixes #11165

When we see an `impl B` and `B` is a type alias to some type `A`, then we need to follow the type alias to look for an `is_empty` method on the aliased type `A`. Before this PR, it'd get the inherent impls of `B`, which there aren't any and so it would warn that there isn't an `is_empty` method even if there was one.
Passing the type alias `DefId` to `TyCtxt::type_of` gives us the aliased `DefId` (or simply return the type itself if it wasn't a type alias) so we can just use that

changelog: [`len_without_is_empty`]: follow type alias to find inherent `is_empty` method

### DIFF
--- a/clippy_lints/src/len_zero.rs
+++ b/clippy_lints/src/len_zero.rs
@@ -424,6 +424,14 @@ fn check_for_is_empty(
     item_name: Symbol,
     item_kind: &str,
 ) {
+    // Implementor may be a type alias, in which case we need to get the `DefId` of the aliased type to
+    // find the correct inherent impls.
+    let impl_ty = if let Some(adt) = cx.tcx.type_of(impl_ty).skip_binder().ty_adt_def() {
+        adt.did()
+    } else {
+        return;
+    };
+
     let is_empty = Symbol::intern("is_empty");
     let is_empty = cx
         .tcx
--- a/tests/ui/len_without_is_empty.rs
+++ b/tests/ui/len_without_is_empty.rs
@@ -436,4 +436,27 @@ impl DifferingErrors {
     }
 }
 
+// Issue #11165
+pub struct Aliased1;
+pub type Alias1 = Aliased1;
+
+impl Alias1 {
+    pub fn len(&self) -> usize {
+        todo!()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        todo!()
+    }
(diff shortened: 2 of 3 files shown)


Centri3 (approved):

Centri3:
Bit weird how `inherent_impls` doesn't already do this (i.e., strip type aliases), but looks fine to me regardless. Thanks!

@bors r+

Issue #10655, opened 2023-04-17
Detect path join with hardcoded absolute path


joshtriplett:
### What it does

Detect calls to `.join` with an absolute path as an argument. This won't do what the user likely expects. `join` with an absolute path will ignore the left-hand side and just use the absolute path.

For example, `root.join("/bin/sh")` will return the path `"/bin/sh"`.

In an ideal world, this would also detect calls like `root.join(SOME_CONSTANT)` where the constant is a hardcoded absolute path, as well, but that may be harder. This would still provide value even if it only detects absolute paths specified directly in the `join` call.

The lint should explain the behavior of `join`, and should suggest removing the leading `/` from the absolute path.

### Lint Name

join_absolute_path

### Category

correctness

### Advantage

Removing the leading `/` from the user's path will make the join create a path relative to the path given on the left-hand side of the join operation.

### Example

```rust
some_path.join("/bin/sh")
```

Could be written as:

```rust
some_path.join("bin/sh")
```

xFrednet:
Hey @anthonygedeon, there is already an open PR https://github.com/rust-lang/rust-clippy/pull/10663 to implement the lint. This is just a quick warning to avoid duplicate work :)

anthonygedeon:
@xFrednet Thanks for letting me know! I'll try to look for another good-first-issue then

PR #11453, opened 2023-09-02
New lint `clippy::join_absolute_paths`

xFrednet:
Hey @ofeeg, this PR is a copy of all the changes you did in https://github.com/rust-lang/rust-clippy/commit/47171d3c6366dd08476b2ab1e1661950d13a4218 from PR https://github.com/rust-lang/rust-clippy/pull/11208. I wasn't sure how to fix the git history. Hope you're okay with me figuring this out in this separate PR

---

changelog: New lint [`join_absolute_paths`]
[#11453](https://github.com/rust-lang/rust-clippy/pull/11453)

Closes: https://github.com/rust-lang/rust-clippy/issues/10655

r? @Centri3 Since you also gave feedback on the other PR. I hope that I copied everything correctly, but a third pair of eyes would be appreciated :D

### DIFF
--- a/clippy_lints/src/methods/join_absolute_paths.rs
+++ b/clippy_lints/src/methods/join_absolute_paths.rs
@@ -1,0 +1,52 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::expr_or_init;
+use clippy_utils::source::snippet_opt;
+use clippy_utils::ty::is_type_diagnostic_item;
+use rustc_ast::ast::LitKind;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::symbol::sym;
+use rustc_span::Span;
+
+use super::JOIN_ABSOLUTE_PATHS;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, recv: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>, expr_span: Span) {
+    let ty = cx.typeck_results().expr_ty(recv).peel_refs();
+    if (is_type_diagnostic_item(cx, ty, sym::Path) || is_type_diagnostic_item(cx, ty, sym::PathBuf))
+        && let ExprKind::Lit(spanned) = expr_or_init(cx, join_arg).kind
+        && let LitKind::Str(symbol, _) = spanned.node
+        && let sym_str = symbol.as_str()
+        && sym_str.starts_with(['/', '\\'])
+    {
+        span_lint_and_then(
+            cx,
+            JOIN_ABSOLUTE_PATHS,
+            join_arg.span,
+            "argument to `Path::join` starts with a path separator",
+            |diag| {
+                let arg_str = snippet_opt(cx, spanned.span).unwrap_or_else(|| "..".to_string());
+
+                let no_separator = if sym_str.starts_with('/') {
+                    arg_str.replacen('/', "", 1)
+                } else {
+                    arg_str.replacen('\\', "", 1)
+                };
+
+                diag.note("joining a path starting with separator will replace the path instead")
+                    .span_suggestion(
+                        spanned.span,
+                        "if this is unintentional, try removing the starting separator",
+                        no_separator,
+                        Applicability::Unspecified,
+                    )
+                    .span_suggestion(
+                        expr_span,
+                        "if this is intentional, try using `Path::new` instead",
+                        format!("PathBuf::from({arg_str})"),
+                        Applicability::Unspecified,
+                    );
+            },
+        );
+    }
+}
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -49,6 +49,7 @@ mod iter_skip_next;
 mod iter_skip_zero;
 mod iter_with_drain;
 mod iterator_step_by_zero;
+mod join_absolute_paths;
 mod manual_next_back;
 mod manual_ok_or;
 mod manual_saturating_arithmetic;
@@ -3684,6 +3685,46 @@ declare_clippy_lint! {
     "calling the `try_from` and `try_into` trait methods when `From`/`Into` is implemented"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to `Path::join` that start with a path separator (`\\` or `/`).
+    ///
+    /// ### Why is this bad?
+    /// If the argument to `Path::join` starts with a separator, it will overwrite
+    /// the original path. If this is intentional, prefer using `Path::new` instead.
+    ///
+    /// Note the behavior is platform dependent. A leading `\\` will be accepted
+    /// on unix systems as part of the file name
+    ///
+    /// See [`Path::join`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join)
+    ///
+    /// ### Example
+    /// ```rust
+    /// # use std::path::{Path, PathBuf};
+    /// let path = Path::new("/bin");
+    /// let joined_path = path.join("/sh");
+    /// assert_eq!(joined_path, PathBuf::from("/sh"));
+    /// ```
+    ///
+    /// Use instead;
+    /// ```rust
+    /// # use std::path::{Path, PathBuf};
+    /// let path = Path::new("/bin");
+    ///
+    /// // If this was unintentional, remove the leading separator
+    /// let joined_path = path.join("sh");
+    /// assert_eq!(joined_path, PathBuf::from("/bin/sh"));
+    ///
+    /// // If this was intentional, create a new path instead
+    /// let new = Path::new("/sh");
+    /// assert_eq!(new, PathBuf::from("/sh"));
+    /// ```
+    #[clippy::version = "1.76.0"]
+    pub JOIN_ABSOLUTE_PATHS,
+    suspicious,
+    "calls to `Path::join` which will overwrite the original path"
+}
+
 pub struct Methods {
     avoid_breaking_exported_api: bool,
     msrv: Msrv,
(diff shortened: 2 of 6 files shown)


clippy_lints/src/methods/mod.rs
```diff
@@ -3563,6 +3564,43 @@ declare_clippy_lint! {
...
+    /// assert_eq!(joined_path, PathBuf::from("/bin/sh"));
+    ///
+    /// // If this was intentional, create a new path instead
+    /// let new = Path::new("/sh");
+    /// assert_eq!(new, PathBuf::from("/sh"));
+    /// ```
+    #[clippy::version = "1.74.0"]
+    pub JOIN_ABSOLUTE_PATHS,
+    correctness,
+    "arg to .join called on a `Path` contains leading separator"
```
Centri3, line 3601: Maybe:
```suggestion
    "call to `Path::join` which will overwrite the original path"
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3563,6 +3564,43 @@ declare_clippy_lint! {
     "calls to `.take()` or `.skip()` that are out of bounds"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to `Path::join` that start with a path separator, like `\\` or `/`..
```
Centri3, line 3569:
```suggestion
    /// Checks for calls to `Path::join` where the `path` argument starts with a path separator (`\\` or `/`).
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3563,6 +3564,43 @@ declare_clippy_lint! {
...
+    ///
+    /// ### Why is this bad?
+    /// `.join()` arguments starting with a separator  (`/` or `\\`) can replace the entire path.
+    /// If this is intentional, prefer using `Path::new()` instead.
```
Centri3, line 3573:
```suggestion
    /// If the argument to `Path::join` starts with a separator, this will overwrite the original path. If this is intentional, prefer using `Path::new` instead.
```

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,30 @@
...
+        span_lint_and_then(
+            cx,
+            JOIN_ABSOLUTE_PATHS,
+            join_arg.span,
+            "argument to `Path::join` starts with a path separator",
+            |diag| {
+                diag
+                      .note("joining a path starting with separator will replace the path instead")
+                      .help("if this is unintentional, try removing the starting separator")
+                      .help("if this is intentional, try creating a new Path instead");
```
Centri3, line 26:
```suggestion
                diag.note("joining a path starting with separator will replace the path instead")
                    .help("if this is unintentional, try removing the starting separator")
                    .help("if this is intentional, try creating a new Path instead");
```
Still a bit weird in terms of formatting. I also think we should also add suggestions here instead of just `.help(...)`.

tests/ui/join_absolute_paths.rs
```diff
@@ -0,0 +1,26 @@
+#![allow(unused)]
+#![warn(clippy::join_absolute_paths)]
+
+use std::path::{Path, PathBuf};
+
+fn main() {
+    // should be linted
```
Centri3, line 7: Let's use `ui_test` annotations:
```suggestion
```

tests/ui/join_absolute_paths.rs
```diff
@@ -0,0 +1,26 @@
...
+    let path = Path::new("/bin");
+    path.join("/sh");
```
Centri3, line 9:
```suggestion
    path.join("/sh");
    //~^ ERROR: argument to `Path::join` starts with a path separator
```
Same with the rest.

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,30 @@
...
+use rustc_span::symbol::sym::Path;
+
+use super::JOIN_ABSOLUTE_PATHS;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>) {
+    let ty = cx.typeck_results().expr_ty(expr).peel_refs();
+    if is_type_diagnostic_item(cx, ty, Path)
+        && let ExprKind::Lit(spanned) = &join_arg.kind
+        && let LitKind::Str(symbol, _) = spanned.node
+        && (symbol.as_str().starts_with('/') || symbol.as_str().starts_with('\\'))
```
Centri3, line 15: Does `\\` do this on unix as well? Or is it normalized to `/` immediately? If it doesn't, we could say this behavior would only be present on windows to remove any potential confusion.

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,30 @@
...
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::symbol::sym::Path;
+
+use super::JOIN_ABSOLUTE_PATHS;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, join_arg: &'tcx Expr<'tcx>) {
+    let ty = cx.typeck_results().expr_ty(expr).peel_refs();
+    if is_type_diagnostic_item(cx, ty, Path)
+        && let ExprKind::Lit(spanned) = &join_arg.kind
```
Centri3, line 13: We should call `expr_to_init` here so that `Path::join(x)` where `x` is `let x = "/a"` is linted too.

Centri3:
Everything looks fine to me, minus the nits. CI seems to agree too.

clippy_lints/src/methods/join_absolute_paths.rs
xFrednet: On unix it just accepts it as a file name

```rs
    let path = Path::new("C:\\Users");
    let path2 = path.join("\\Ducks");
    println!("{path2:#?}"); -> "C:\\Users/\\Ducks"
```

This behavior actually makes sense to me. I'll add a note to the lint description. I wouldn't add it to the lint emission, as it's already quite verbose.

ofeeg:
Looks like it all got figured out. Sorry for the trouble.

clippy_lints/src/methods/join_absolute_paths.rs
lengyijun: Can we use `[has_root](https://github.com/rust-lang/rust-clippy/pull/11453#method.has_root)` to check ?

xFrednet: @lengyijun I assume you mean [`Path::has_root()`](https://doc.rust-lang.org/std/path/struct.Path.html#method.has_root)?

I don't think that it would be good to be platform specific here. I can imagine that several projects just run Clippy on linux without specifying `--all-targets`. This assumption comes from the fact that I often forget to specify that flag in the CI. There might also be users, who only run Clippy locally. However, this can lead to false positives, even if I don't believe that anybody wants to start their folder name with `\\` on linux. So, I think we should keep the current behavior, but move the lint to `suspicious`, to have the lint warn by default and not deny. Does that sound good to everyone?

xFrednet:
I've rebased and moved the lint to `suspicious`, the rest should be the same as before

xFrednet:
@Centri3 Seems to be busy lately, @blyxyas would you mind taking over the review?

r? @blyxyas

blyxyas (changes requested):
Meow

clippy_lints/src/methods/join_absolute_paths.rs
```diff
@@ -0,0 +1,52 @@
...
+            join_arg.span,
+            "argument to `Path::join` starts with a path separator",
+            |diag| {
+                let arg_str = snippet_opt(cx, spanned.span).unwrap_or_else(|| "..".to_string());
+
+                let no_separator = if sym_str.starts_with('/') {
+                    arg_str.replacen('/', "", 1)
+                } else {
+                    arg_str.replacen('\\', "", 1)
+                };
```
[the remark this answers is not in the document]
xFrednet: Assume, you suggest this to safe performance? I feel like this refactoring would make the code less readable while having almost no impact. The closure is only called if the lint is actually emitted. I doubt that this would even be measurable.

I noticed another "optimization", of only calling `starts_with(&['/', '\\'])` instead of two separate times. But that rules out this suggestion, from what I can tell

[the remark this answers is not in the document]
blyxyas: Not really performance, it's just that seeing a function two times with the same arguments hurts my feelings :cry:.  starts_with(&['/', '\\']) is fine though :) Please apply that change ₍˄·͈༝·͈˄₎ฅ˒˒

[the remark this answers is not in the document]
xFrednet: Already done :D

blyxyas (approved):
LGTM, thanks! :heart:
Could you squash these and add a `Co-Authored-By: ofeeg <mhanna0000@gmail.com>` to the commit message, in honor of those who are no longer listed as authors.

xFrednet:
Meow @blyxyas 
```
  ^~^  ,
 ('Y') )
 /   \/ 
(\|||/)
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3684,6 +3685,46 @@ declare_clippy_lint! {
...
+    ///
+    /// // If this was unintentional, remove the leading separator
+    /// let joined_path = path.join("sh");
+    /// assert_eq!(joined_path, PathBuf::from("/bin/sh"));
+    ///
+    /// // If this was intentional, create a new path instead
+    /// let new = Path::new("/sh");
+    /// assert_eq!(new, PathBuf::from("/sh"));
+    /// ```
+    #[clippy::version = "1.75.0"]
```
blyxyas, line 3722:
```suggestion
    #[clippy::version = "1.76.0"]
```

PR #11455, opened 2023-09-03
skip `todo!()` in  `never_loop`

digama0:
As promised in #11450, here is an implementation which skips occurrences of the `todo!()` macro.

changelog: [`never_loop`]: skip loops containing `todo!()`

### DIFF
--- a/clippy_lints/src/loops/never_loop.rs
+++ b/clippy_lints/src/loops/never_loop.rs
@@ -2,6 +2,7 @@ use super::utils::make_iterator_snippet;
 use super::NEVER_LOOP;
 use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::higher::ForLoop;
+use clippy_utils::macros::root_macro_call_first_node;
 use clippy_utils::source::snippet;
 use rustc_errors::Applicability;
 use rustc_hir::{Block, Destination, Expr, ExprKind, HirId, InlineAsmOperand, Pat, Stmt, StmtKind};
@@ -263,13 +264,25 @@ fn never_loop_expr<'tcx>(
         | ExprKind::Lit(_)
         | ExprKind::Err(_) => NeverLoopResult::Normal,
     };
-    combine_seq(result, || {
+    let result = combine_seq(result, || {
         if cx.typeck_results().expr_ty(expr).is_never() {
             NeverLoopResult::Diverging
         } else {
             NeverLoopResult::Normal
         }
-    })
+    });
+    if  let NeverLoopResult::Diverging = result &&
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) &&
+        let "todo" | "unimplemented" = cx.tcx.item_name(macro_call.def_id).as_str()
+    {
+        // We return MayContinueMainLoop here because we treat `todo!()` and
+        // `unimplemented!()` macros as potentially containing any code,
+        // including a continue of the main loop. This effectively silences the lint
+        // whenever a loop contains one of these macros anywhere.
+        NeverLoopResult::MayContinueMainLoop
+    } else {
+        result
+    }
 }
 
 fn never_loop_expr_all<'tcx, T: Iterator<Item = &'tcx Expr<'tcx>>>(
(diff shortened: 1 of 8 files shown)


clippy_lints/src/loops/never_loop.rs
```diff
@@ -263,13 +264,25 @@ fn never_loop_expr<'tcx>(
...
         if cx.typeck_results().expr_ty(expr).is_never() {
             NeverLoopResult::Diverging
         } else {
             NeverLoopResult::Normal
         }
-    })
+    });
+    if  let NeverLoopResult::Diverging = result &&
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) &&
+        let "todo" | "unimplemented" = cx.tcx.item_name(macro_call.def_id).as_str()
```
Centri3, line 276: You can use diagnostic items instead; `todo_macro`/`unimplemented_macro`.

Centri3 (changes requested):
This should only skip `todo!` imo. `unimplemented!` is used for stuff that won't be implemented, not for WIP code.

clippy_lints/src/loops/never_loop.rs
digama0: I tried that, it didn't work. Could that `sym` be referring to something else?

PS: this is also the method used in the todo restriction lint, so if we can get to the bottom of this we may also consider fixing that code too.

Centri3: Odd, since it's declared here: https://doc.rust-lang.org/nightly/src/core/macros/mod.rs.html#767

Not sure why that wouldn't work then, since `macro_call.def_id` should refer to the `todo!` this is assigned to.

Alexendoo:
```rust
let Some(sym::todo_macro | sym::unimplemented_macro) = cx.tcx.get_diagnostic_name(macro_call.def_id)
```

seems to work fine for me

digama0: From testing and from reading the documentation, it appears that `sym::todo_macro` represents the string "todo_macro", which... seems useless? And I tried exactly that @Alexendoo , it fails the test on my machine

Centri3: yes, `sym::todo_macro` represents the string "todo_macro", *but* that string represents the diagnostic item `todo!`.

I'll try out the code Alex suggested, I don't see why it wouldn't work.

Centri3: The suggested code works fine for me.

digama0: Oh, sorry I used `item_name` rather than `get_diagnostic_name`

Centri3 (approved):

Centri3:
This version LGTM. r=me once `vec.rs` has been blessed

clippy_lints/src/loops/never_loop.rs
```diff
@@ -263,13 +264,25 @@ fn never_loop_expr<'tcx>(
...
             NeverLoopResult::Normal
         }
-    })
+    });
+    if  let NeverLoopResult::Diverging = result &&
+        let Some(macro_call) = root_macro_call_first_node(cx, expr) &&
+        let Some(sym::todo_macro) = cx.tcx.get_diagnostic_name(macro_call.def_id)
+    {
+        // We return MayContinueMainLoop here because we treat `todo!()` and
+        // `unimplemented!()` macros as potentially containing any code,
```
dswij, line 279:
```suggestion
        // We return MayContinueMainLoop here because we treat `todo!()`
        // as potentially containing any code,
```

dswij (commented):
Just a small comment suggestion

dswij (approved):

dswij:
Thank you. @bors r=Centri3,dswij

digama0:
how do I kick the bot to try again?

Jarcho:
@bors retry

The changelog line needs to start with `changlog:` (note the colon). I've fixed that for you.

PR #11459, opened 2023-09-03
[`implied_bounds_in_impls`]: include (previously omitted) associated types in suggestion

### LINKED ISSUE #11435: `implied_bounds_in_impls` false positive
### Summary

The lint suggests removing a specified bound, however, removing the bound breaks code that then uses the function.

### Lint Name

implied_bounds_in_impls

### Reproducer

I tried this code:

```rust
use std::{
    cell::RefCell,
    fmt,
    ops::{Deref, DerefMut},
};

pub struct HeavyState<T: fmt::Debug> {
    current: RefCell<T>,
}

impl<T: fmt::Debug> HeavyState<T> {
    pub fn get_mut(&self) -> impl Deref<Target = T> + DerefMut + '_ {
        self.current.borrow_mut()
    }
}

pub fn foo(bindings: HeavyState<Vec<u32>>) {
   let mut b = bindings.get_mut();
   b.push(0);
}

fn main() {}
```

Clippy reports the following:

```
warning: this bound is already specified as the supertrait of `DerefMut`
  --> src/main.rs:14:35
   |
14 |     pub fn get_mut(&self) -> impl Deref<Target = T> + DerefMut + '_ {
   |                                   ^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#implied_bounds_in_impls
   = note: `#[warn(clippy::implied_bounds_in_impls)]` on by default
help: try removing this bound
   |
14 -     pub fn get_mut(&self) -> impl Deref<Target = T> + DerefMut + '_ {
14 +     pub fn get_mut(&self) -> impl DerefMut + '_ {
   |
```

However, applying the suggestion results in the following:
```
error[E0599]: no method named `push` found for opaque type `impl std::ops::DerefMut + '_` in the current scope
  --> src/main.rs:23:7
   |
23 |     b.push(0);
   |       ^^^^ method not found in `impl DerefMut`
```

The suggestion should be to transfer the associated type bound `Target = T` onto `DerefMut` rather than entirely removing the bound. This would be the correct code:
```rs
impl<T: fmt::Debug> HeavyState<T> {
    pub fn get_mut(&self) -> impl DerefMut<Target = T> + '_ {
        self.current.borrow_mut()
    }
}
```

### Version

```text
rustc 1.74.0-nightly (84a9f4c6e 2023-08-29)
binary: rustc
commit-hash: 84a9f4c6e6f0f02bff9acc9a5e0305a1554a87fc
commit-date: 2023-08-29
host: x86_64-apple-darwin
release: 1.74.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

y21:
Fixes #11435

It now includes associated types from the implied bound that were omitted in the second bound. Example:
```rs
fn f() -> impl Iterator<Item = u8> + ExactSizeIterator> {..}
```
Suggestion before this change:
```diff
- pub fn my_iter() -> impl Iterator<Item = u32> + ExactSizeIterator {
+ pub fn my_iter() -> impl ExactSizeIterator {
```
It didn't include `<Item = u32>` on `ExactSizeIterator`. Now, with this change, it does.
```diff
- pub fn my_iter() -> impl Iterator<Item = u32> + ExactSizeIterator {
+ pub fn my_iter() -> impl ExactSizeIterator<Item = u32> {
```

We also now extend the span to include not just possible `+` ahead of it, but also behind it (an example for this is in the linked issue as well).
**Note:** The overall diff is a bit noisy, because building up the suggestion involves quite a bit more logic now and I decided to extract that into its own function. For that reason, I split this PR up into two commits. The first commit contains the actual "logic" changes. Second commit just moves code around.

changelog: [`implied_bounds_in_impls`]: include (previously omitted) associated types in suggestion
changelog: [`implied_bounds_in_impls`]: include the `+` behind bound if it's the last bound

### DIFF
--- a/clippy_lints/src/implied_bounds_in_impls.rs
+++ b/clippy_lints/src/implied_bounds_in_impls.rs
@@ -56,7 +56,9 @@ fn emit_lint(
     poly_trait: &rustc_hir::PolyTraitRef<'_>,
     opaque_ty: &rustc_hir::OpaqueTy<'_>,
     index: usize,
+    // The bindings that were implied
     implied_bindings: &[rustc_hir::TypeBinding<'_>],
+    // The original bindings that `implied_bindings` are implied from
     implied_by_bindings: &[rustc_hir::TypeBinding<'_>],
     implied_by_args: &[GenericArg<'_>],
     implied_by_span: Span,
@@ -69,7 +71,7 @@ fn emit_lint(
         poly_trait.span,
         &format!("this bound is already specified as the supertrait of `{implied_by}`"),
         |diag| {
-            // If we suggest removing a bound, we may also need extend the span
+            // If we suggest removing a bound, we may also need to extend the span
             // to include the `+` token that is ahead or behind,
             // so we don't end up with something like `impl + B` or `impl A + `
 
--- a/clippy_lints/src/implied_bounds_in_impls.rs
+++ b/clippy_lints/src/implied_bounds_in_impls.rs
@@ -50,6 +50,93 @@ declare_clippy_lint! {
 }
 declare_lint_pass!(ImpliedBoundsInImpls => [IMPLIED_BOUNDS_IN_IMPLS]);
 
+#[allow(clippy::too_many_arguments)]
+fn emit_lint(
+    cx: &LateContext<'_>,
+    poly_trait: &rustc_hir::PolyTraitRef<'_>,
+    opaque_ty: &rustc_hir::OpaqueTy<'_>,
+    index: usize,
+    implied_bindings: &[rustc_hir::TypeBinding<'_>],
+    implied_by_bindings: &[rustc_hir::TypeBinding<'_>],
+    implied_by_args: &[GenericArg<'_>],
+    implied_by_span: Span,
+) {
+    let implied_by = snippet(cx, implied_by_span, "..");
+
+    span_lint_and_then(
+        cx,
+        IMPLIED_BOUNDS_IN_IMPLS,
+        poly_trait.span,
+        &format!("this bound is already specified as the supertrait of `{implied_by}`"),
+        |diag| {
+            // If we suggest removing a bound, we may also need extend the span
+            // to include the `+` token that is ahead or behind,
+            // so we don't end up with something like `impl + B` or `impl A + `
+
+            let implied_span_extended = if let Some(next_bound) = opaque_ty.bounds.get(index + 1) {
+                poly_trait.span.to(next_bound.span().shrink_to_lo())
+            } else if index > 0
+                && let Some(prev_bound) = opaque_ty.bounds.get(index - 1)
+            {
+                prev_bound.span().shrink_to_hi().to(poly_trait.span.shrink_to_hi())
+            } else {
+                poly_trait.span
+            };
+
+            let mut sugg = vec![(implied_span_extended, String::new())];
+
+            // We also might need to include associated type binding that were specified in the implied bound,
+            // but omitted in the implied-by bound:
+            // `fn f() -> impl Deref<Target = u8> + DerefMut`
+            // If we're going to suggest removing `Deref<..>`, we'll need to put `<Target = u8>` on `DerefMut`
+            let omitted_assoc_tys: Vec<_> = implied_bindings
+                .iter()
+                .filter(|binding| !implied_by_bindings.iter().any(|b| b.ident == binding.ident))
+                .collect();
+
+            if !omitted_assoc_tys.is_empty() {
+                // `<>` needs to be added if there aren't yet any generic arguments or bindings
+                let needs_angle_brackets = implied_by_args.is_empty() && implied_by_bindings.is_empty();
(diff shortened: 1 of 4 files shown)


clippy_lints/src/implied_bounds_in_impls.rs
```diff
@@ -162,10 +249,14 @@ fn check(cx: &LateContext<'_>, decl: &FnDecl<'_>) {
...
                 && let implied_args = path.args.map_or([].as_slice(), |a| a.args)
+                && let implied_bindings = path.args.map_or([].as_slice(), |a| a.bindings)
                 && let Some(def_id) = poly_trait.trait_ref.path.res.opt_def_id()
-                && let Some(implied_by_span) = implied_bounds
+                && let Some((implied_by_span, implied_by_args, implied_by_bindings)) = implied_bounds
                     .iter()
-                    .find_map(|&(span, implied_by_args, preds, implied_by_def_id)| {
+                    .find_map(|&(span, implied_by_path, preds, implied_by_def_id)| {
+                        let implied_by_args = implied_by_path.args.map_or([].as_slice(), |a| a.args);
+                        let implied_by_bindings = implied_by_path.args.map_or([].as_slice(), |a| a.bindings);
```
blyxyas, line 258: As these `.as_slice()` do not convey useful info but clutter the code, I think we can replace them with `&[]` (IIRC, `&[]` is equivalent to `[].as_slice()`)

tests/ui/implied_bounds_in_impls.rs
```diff
@@ -83,4 +83,43 @@ mod issue11422 {
...
+    trait Trait4<T>: Trait3<T, i16, i64> {}
+    impl Trait3<i8, i16, i64> for () {
+        type X = i32;
+        type Y = i128;
+    }
+    impl Trait4<i8> for () {}
+
+    // Associated type `X` is specified, but `Y` is not, so only that associated type should be moved
+    // over
+    fn f3() -> impl Trait3<i8, i16, i64, X = i32, Y = i128> + Trait4<i8, X = i32> {}
```
blyxyas, line 122: To test that this actually works, `Trait3`'s `X` and `Trait4`'s `X` should be distinct.

blyxyas (commented):
Pretty good first revision, thanks for being so committed to this one lint :cat: (really, all current commits affecting `implied_bounds_in_impls.rs` are by you :cactus: )

clippy_lints/src/implied_bounds_in_impls.rs
y21: This sadly [doesn't work](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=024fdf2625542008e7ffaae2fa912466).

 `&[]` is more like of type `&[_; 0]`, which can usually coerce to `&[_]`, but not here (probably because it's in a generic context), and `[].as_slice()` does the coercion explicitly to avoid the type check error. We could also do `&[] as &[_]`, but imo it looks weird and it's not that much better than `[].as_slice()`

tests/ui/implied_bounds_in_impls.rs
y21: AFAIK it's not possible for `Trait4::X` to be different from `Trait3::X` ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3a7240598d8246f53ee4b79e0f890b96))
This test was just to make sure that it doesn't "duplicate" the `X` associated type

blyxyas: the Rust compiler saving the day once again :tada:

clippy_lints/src/implied_bounds_in_impls.rs
```diff
@@ -50,6 +50,95 @@ declare_clippy_lint! {
...
+
+            let implied_span_extended = if let Some(next_bound) = opaque_ty.bounds.get(index + 1) {
+                poly_trait.span.to(next_bound.span().shrink_to_lo())
+            } else if index > 0
+                && let Some(prev_bound) = opaque_ty.bounds.get(index - 1)
+            {
+                prev_bound.span().shrink_to_hi().to(poly_trait.span.shrink_to_hi())
+            } else {
+                poly_trait.span
+            };
```
blyxyas, line 86: I'm kinda worried about this check, it's kinda inconsistent (one checking index and the other not), but it may not matter if we make sure it doesn't fail.

Could you add a test with 3 bounds? Maybe some with generics and some without them?

y21: The reason why index - 1 has a check and the other doesn't is mainly that `- 1` (I assume you're referring to this) can panic with debug assertions due to overflow if it's the first bound, while `+ 1` in this context can't.
I mean, a slice with `usize::MAX` elements (which is needed here for an overflow to occur) is not only realistically impossible because nobody has that much ram but also theoretically impossible, since afaik allocations in rust can never be bigger than `isize::MAX`

blyxyas (approved):
LGTM, thanks :heart:!

Issue #11429, opened 2023-08-30
`option_if_let_else` suggestion causes `redundant_closure` warning


Kriskras99:
### Summary

The `option_if_let_else` suggestions do not check if the closure can be replaced by the function itself.

### Reproducer

I tried this code:

```rust
#[warn(clippy::option_if_let_else)]
fn main() {
    let option: Option<&std::collections::HashMap<&str, &str>> = None;
    let mut hashmap = if let Some(hashmap) = option {
        hashmap.clone()
    } else {
        std::collections::HashMap::new()
    };
    hashmap.insert("Hello", "Clippy");
    println!("{hashmap:?}");
}
```

I expected to see this happen:
```
warning: use Option::map_or_else instead of an if let/else
 --> src/main.rs:4:23
  |
4 |       let mut hashmap = if let Some(hashmap) = option {
  |  _______________________^
5 | |         hashmap.clone()
6 | |     } else {
7 | |         std::collections::HashMap::new()
8 | |     };
  | |_____^ help: try: `option.map_or_else(|| std::collections::HashMap::new(), |hashmap| hashmap.clone())`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#option_if_let_else
```
After applying the fix I expect no warnings from Clippy.
Instead, this happened:
```
warning: redundant closure
 --> src/main.rs:4:42
  |
4 |     let mut hashmap = option.map_or_else(|| std::collections::HashMap::new(), |hashmap| hashmap.clone());
  |                                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace the closure with the function itself: `std::collections::HashMap::new`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
  = note: `#[warn(clippy::redundant_closure)]` on by default
```

### Version

```text
rustc 1.74.0-nightly (84a9f4c6e 2023-08-29)
binary: rustc
commit-hash: 84a9f4c6e6f0f02bff9acc9a5e0305a1554a87fc
commit-date: 2023-08-29
host: x86_64-unknown-linux-gnu
release: 1.74.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

@rustbot label + L-suggestion

J-ZhengLi:
@rustbot claim for now, ask question later.

**later**:

So we need to check if the function call has any parameters, and suggesting pass the function itself if not?
[...]

PR #11460, opened 2023-09-04
suggest passing function instead of calling it in closure for [`option_if_let_else`]

J-ZhengLi:
fixes: #11429 

changelog: suggest passing function instead of calling it in closure for [`option_if_let_else`]

### DIFF
--- a/clippy_lints/src/option_if_let_else.rs
+++ b/clippy_lints/src/option_if_let_else.rs
@@ -166,10 +166,9 @@ fn try_get_option_occurrence<'tcx>(
 
             let mut app = Applicability::Unspecified;
 
-            let (none_body, is_argless_call) = if let Some(call_expr) = try_get_argless_call_expr(none_body) {
-                (call_expr, true)
-            } else {
-                (none_body, false)
+            let (none_body, is_argless_call) = match none_body.kind {
+                ExprKind::Call(call_expr, []) if !none_body.span.from_expansion() => (call_expr, true),
+                _ => (none_body, false),
             };
 
             return Some(OptionOccurrence {
@@ -193,14 +192,6 @@ fn try_get_option_occurrence<'tcx>(
     }
 
     None
-}
-
-/// Gets the call expr iff it does not have any args and was not from macro expansion.
-fn try_get_argless_call_expr<'tcx>(expr: &Expr<'tcx>) -> Option<&'tcx Expr<'tcx>> {
-    match expr.kind {
-        ExprKind::Call(call_expr, []) if !expr.span.from_expansion() => Some(call_expr),
-        _ => None,
-    }
 }
 
 fn try_get_inner_pat_and_is_result<'tcx>(cx: &LateContext<'tcx>, pat: &Pat<'tcx>) -> Option<(&'tcx Pat<'tcx>, bool)> {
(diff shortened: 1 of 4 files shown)


clippy_lints/src/option_if_let_else.rs
```diff
@@ -188,6 +195,16 @@ fn try_get_option_occurrence<'tcx>(
...
 
+/// Gets the call expr iff it does not have any args and was not from macro expansion.
+fn try_get_argless_call_expr<'tcx>(expr: &Expr<'tcx>) -> Option<&'tcx Expr<'tcx>> {
+    if !expr.span.from_expansion() &&
+        let ExprKind::Call(call_expr, []) = expr.kind
+    {
+        return Some(call_expr);
+    }
+    None
+}
```
samueltardieu, line 206: This snippet may help avoid a function call if `expr` is not an `ExprKind::Call` and looks more balanced IMO:

```suggestion
    match expr.kind {
        ExprKind::Call(call_expr, []) if !expr.span.from_expansion() => Some(call_expr),
        _ => None,
    }
```

J-ZhengLi: good catch! it might or might not triggers another clippy warning, I'm not sure, let's find out 😆

J-ZhengLi: there was an extra bracket at the end, but it's all good now~ thanks!

clippy_lints/src/option_if_let_else.rs
```diff
@@ -165,6 +165,13 @@ fn try_get_option_occurrence<'tcx>(
             }
 
             let mut app = Applicability::Unspecified;
+
+            let (none_body, is_argless_call) = if let Some(call_expr) = try_get_argless_call_expr(none_body) {
+                (call_expr, true)
+            } else {
+                (none_body, false)
+            };
```
samueltardieu, line 173: In fact, `try_get_argless_call_expr()` is now so short and used at only one place that it can probably be inlined (and the function removed):

```suggestion
            let (none_body, is_argless_call) = match none_body.kind {
                ExprKind::Call(call_expr, []) if !none_body.span.from_expansion() => (call_expr, true),
                _ => (none_body, false),
            };
```

clippy_lints/src/option_if_let_else.rs
```diff
@@ -188,6 +195,14 @@ fn try_get_option_occurrence<'tcx>(
     None
 }
 
+/// Gets the call expr iff it does not have any args and was not from macro expansion.
```
Centri3, line 198:
```suggestion
/// Gets the call expr if it does not have any args and was not from macro expansion.
```

Centri3 (approved):
LGTM other than @samueltardieu's suggestion and a typo. Thanks!

J-ZhengLi:
> LGTM other than @samueltardieu's suggestion and a typo. Thanks!

The typo was part of a function that was no longer needed, and I... ended up delete it, but still thank you~ 😺

clippy_lints/src/option_if_let_else.rs
J-ZhengLi: The typo was part of a function that was no longer needed, and I... ended up delete it, but still thank you~ 😺

J-ZhengLi:
@Centri3 Perhaps this was forgotten?

Centri3:
Haven't had the time. This version still looks good to me so @bors r+

PR #11468, opened 2023-09-07
add extra `byref` checking for the guard's local

### LINKED ISSUE #11465: `redundant_guards` doesn't take `ref` into account
### Summary

If I have an enum variant storing an array, whose value is compared to byte strings, the comparison must happen in a guard.

For example, I'd do:

`Foo(ref arr) if arr == b"bar"`

Where `arr` is `[u8; 3]`.

The lint will incorrectly assume that `b"bar"` can simply be moved into the pattern, which is not the case, as its type would be `&[u8; 3]`.

### Lint Name

redundant_guards

### Reproducer

I tried this code:

```rust
enum Test {
    Foo([u8; 3])
}

let foo = Test::Foo([98, 97, 114]);

match foo {
    Test::Foo(ref arr) if arr == b"bar" => {}
    _ => {}
}
```

I saw this happen:

```
warning: redundant guard
  --> src/main.rs:11:27
   |
11 |     Test::Foo(ref arr) if arr == b"bar" => {}
   |                           ^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_guards
   = note: `#[warn(clippy::redundant_guards)]` on by default
help: try
   |
11 -     Test::Foo(ref arr) if arr == b"bar" => {}
11 +     Test::Foo(b"bar") => {}
   |
```

No warning

### Version

```text
rustc 1.74.0-nightly (ec08a0337 2023-09-04)
binary: rustc
commit-hash: ec08a0337f3556212525dbf1d3b41e19bdf27621
commit-date: 2023-09-04
release: 1.74.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

mojave2:
changelog: [`redundant_guards`]: Now checks if the variable is bound using `ref` before linting.

The lint should not be emitted, when the local variable is bind by-ref in the pattern.

fixes #11465

### DIFF
--- a/clippy_lints/src/matches/redundant_guards.rs
+++ b/clippy_lints/src/matches/redundant_guards.rs
@@ -123,7 +123,7 @@ fn get_pat_binding<'tcx>(
                 if matches!(bind_annot.0, rustc_ast::ByRef::Yes) {
                     let _ = byref_ident.insert(ident);
                 }
-                // the second call of `replce()` returns a `Some(span)`, meaning a multi-binding pattern
+                // the second call of `replace()` returns a `Some(span)`, meaning a multi-binding pattern
                 if span.replace(pat.span).is_some() {
                     multiple_bindings = true;
                     return false;
--- a/clippy_lints/src/matches/redundant_guards.rs
+++ b/clippy_lints/src/matches/redundant_guards.rs
@@ -34,24 +34,45 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'tcx>]) {
                 ],
                 MatchSource::Normal,
             ) = if_expr.kind
+            && let Some((binding_span, is_field, is_byref)) = get_pat_binding(cx, scrutinee, outer_arm)
         {
+            if is_field && is_byref { return; }
+            let pat_span = if let PatKind::Ref(pat, _) = arm.pat.kind {
+                if is_byref { pat.span } else { continue; }
+            } else {
+                if is_byref { continue; }
+                arm.pat.span
+            };
+
             emit_redundant_guards(
                 cx,
                 outer_arm,
                 if_expr.span,
-                scrutinee,
-                arm.pat.span,
+                pat_span,
+                binding_span,
+                is_field,
                 arm.guard,
             );
         }
         // `Some(x) if let Some(2) = x`
-        else if let Guard::IfLet(let_expr) = guard {
+        else if let Guard::IfLet(let_expr) = guard
+            && let Some((binding_span, is_field, is_byref)) = get_pat_binding(cx, let_expr.init, outer_arm)
+        {
+            if is_field && is_byref { return; }
+            let pat_span = if let PatKind::Ref(pat, _) = let_expr.pat.kind {
+                if is_byref && !is_field { pat.span } else { continue; }
+            } else {
+                if is_byref { continue; }
+                let_expr.pat.span
+            };
+
             emit_redundant_guards(
                 cx,
                 outer_arm,
                 let_expr.span,
-                let_expr.init,
-                let_expr.pat.span,
+                pat_span,
+                binding_span,
+                is_field,
                 None,
             );
         }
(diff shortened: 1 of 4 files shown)


clippy_lints/src/matches/redundant_guards.rs
```diff
@@ -80,6 +81,37 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'tcx>]) {
...
+        },
+        ..
+    } = arm
+    else {
+        return false;
+    };
+
+    for pat in *pats {
+        if let PatKind::Binding(annot, pat_id, ..) = &pat.kind
+            && matches!(annot.0, rustc_ast::ByRef::No)
```
y21, line 107: Would it make sense to have this ByRef check in the `get_pat_binding` function, instead of duplicating the code for finding the right binding in a new function? I.e. having it here:
https://github.com/rust-lang/rust-clippy/blob/415ba21c3b78125cf583ca0ef7a318a8f347a282/clippy_lints/src/matches/redundant_guards.rs#L89-L90

clippy_lints/src/matches/redundant_guards.rs
```diff
@@ -80,6 +81,37 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'tcx>]) {
     }
 }
 
+fn is_the_local_binded_by_value_in_pat(local: &Expr<'_>, arm: &Arm<'_>) -> bool {
```
blyxyas, line 84: This name probably should be more illustrative.

tests/ui/redundant_guards.rs
blyxyas, line 1: Could you add some more tests? Like 2 simpler tests that are not specifically from #11465, checking that everything still works.

clippy_lints/src/matches/redundant_guards.rs
```diff
@@ -67,31 +88,48 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'tcx>]) {
...
         outer_arm.pat.walk(|pat| {
-            if let PatKind::Binding(_, hir_id, _, _) = pat.kind
+            if let PatKind::Binding(bind_annot, hir_id, _, _) = pat.kind
                 && hir_id == local
-                && span.replace(pat.span).is_some()
             {
-                multiple_bindings = true;
-                return false;
+                is_byref = matches!(bind_annot.0, rustc_ast::ByRef::Yes);
+                if span.replace(pat.span).is_some() {
```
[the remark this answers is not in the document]
y21: It gets set to `Some(_)` by the replace call, so the condition will be true when it reaches the replace call for the 2nd time or any other time after the first replace call (ie when the pattern is found twice, for stuff like `let Ok(v) | Err(v)` I imagine).

I got confused by this before when I looked at the code, maybe it could be written clearer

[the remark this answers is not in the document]
mojave2: How about the code below?
```rust
if span.is_some() {
    multiple_bindings = true;
    return false;
} else {
    span.insert(pat.span);
}
```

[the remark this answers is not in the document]
blyxyas: No, but a comment linking to @y21's comment would be really helpful for future reference would help future reviewers / contributors?

clippy_lints/src/matches/redundant_guards.rs
```diff
@@ -67,31 +88,48 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, arms: &'tcx [Arm<'tcx>]) {
...
         }
     }
 }
 
-fn get_pat_binding<'tcx>(cx: &LateContext<'tcx>, guard_expr: &Expr<'_>, outer_arm: &Arm<'tcx>) -> Option<(Span, bool)> {
+fn get_pat_binding<'tcx>(
+    cx: &LateContext<'tcx>,
+    guard_expr: &Expr<'_>,
+    outer_arm: &Arm<'tcx>,
+) -> Option<(Span, bool, bool)> {
```
blyxyas, line 118: Probably a good idea to write a comment describing what are these 2 `bool`s / what are their function

tests/ui/redundant_guards.fixed
```diff
@@ -143,3 +143,44 @@ fn g(opt_s: Option<S>) {
...
+
+        let struct_b = B {
+            b: "bar".to_string(),
+            c: 42,
+        };
+        match struct_b {
+            B { ref b, .. } if b == "bar" => {},
+            B { ref c, .. } if c == &1 => {},
+            B { ref c, .. } if let &1 = c => {},
+            B { ref c, .. } if matches!(c, &1) => {},
```
blyxyas, line 182: Shouldn't these have the same treatment as the first `match c`? Is this intended because it's from a struct?

blyxyas (commented):
This is pretty close to being finished, just a couple questions :heart:

clippy_lints/src/matches/redundant_guards.rs
mojave2: I just made a custom type `PatBindingInfo` for the return type.

tests/ui/redundant_guards.fixed
mojave2: Yes. These cases are handled now.

blyxyas (approved):
Just this mini-typo and this should be ready to go, thanks! :heart:

PR #11062, opened 2023-07-02
New lint [`bare_dos_device_names`]

Centri3:
Closes #8511

I've made this a draft as it's not quite ready yet. But whoever's randomly picked, feel free to review ^^ Everything I've left out so far and/or haven't finished is really because I didn't get enough sleep :D I've left some TODOs about the place and will likely make it a tad more general soon enough.

changelog: New lint [`bare_dos_device_names`]

PR #11476, opened 2023-09-10
new lint: `zombie_processes`

### LINKED ISSUE #10754: Is it possible to generate a warning about std::process::Command::spawn() returned Child is never waited?
### What it does

If there is no wait on the Child handle(and/or before it goes out of scope, the destructor won't wait() it either.), generate a warning about possible consequences: zombie process.

### Lint Name

process_child_not_waited

### Category

_No response_

### Advantage

_No response_

### Drawbacks

None.

### Example

```rust
thread::spawn(move || {
                std::process::Command::new(&command[0])
                .args(&command[1..])
                .spawn()
                .unwrap_or_else(|e| panic!("Failed to execute {}: {}", command.join(" "), e));
            });
```

Could be written as:

```rust
thread::spawn(move || {
                let child = std::process::Command::new(&command[0])
                .args(&command[1..])
                .spawn()
                .unwrap_or_else(|e| panic!("Failed to execute {}: {}", command.join(" "), e));
                child.wait();
            });
```

y21:
Closes #10754

Lint description should explain this PR, so I think there's nothing else to say here ^^

changelog: new lint: [`zombie_processes`]


Centri3:
Out of curiosity, does aborting the program also finally release these "zombies" in any way? If so, I think we could check for an unconditional `panic!`, `exit`, `abort`, etc., after it.

I ask because I was creating a modloader for a game, it'd be loaded then reopen the program with a different entry point before aborting the current program (to run our own code after the loader finishes loading DLLs, or something. No idea why I didn't use function hooking, probably wanted to feel clever ^^). Not really the greatest example, but if the parent process closing allows these zombies to terminate properly, I think it makes sense to check for this.

Regardless though, this is a nice lint and I'll review it soon.

y21:
On [waitpid's man page](https://linux.die.net/man/2/waitpid) there's this note:
> If a parent process terminates, then its "zombie" children (if any) are adopted by [init](https://linux.die.net/man/8/init)(8), which automatically performs a wait to remove the zombies.

So yes, I guess that releases them :)
I also did a quick experiment and I can see that this does indeed happen:
<details>
<summary>Experiment details</summary>

main.rs:
```rs
use std::process::Command;

fn main() {
  for _ in 0..5 {
    let _ = Command::new("echo").spawn().unwrap(); // <- no wait()
  }
  println!("Zombies spawned. Press enter to release them");
  std::io::stdin().read_line(&mut String::new()).unwrap();
}
```
This doesn't have an explicit exit() call, but this is all at the end of a main function and I think that'd have the same effect.

```sh
$ rustc main.rs
$ ./main
Zombies spawned. Press enter to release them

# at this point, run `ps aux` in another session and look for 5 `[echo] <defunct>` entries (zombies). 
#
# <enter>
# 
# now that the process has exited, run `ps aux` again and see that there are no <defunct> entries

```
(This seems kinda specific to linux, idk if this is needed at all on Windows...)
</details>

>  If so, I think we could check for an unconditional panic!, exit, abort, etc., after it.

By this I assume you mean that we shouldn't lint if there is one of these after spawning the process.
I agree with exit and abort, but I'm not sure about `panic!`, since you can catch those and so it's not always guaranteed that the process terminates (which means that the zombie processes might remain after the panic)

Centri3:
TIL :)

Platform-specific lints are fine. For example, `non_octal_unix_permissions` is basically a nop on windows (or other platforms) since you can't actually access the method it checks. There's another similar to this in #11062 too. `wait`-ing is also usually what the programmer intended even if they're on Windows (I can't really think of many reasons you wouldn't)

Catching a panic is only possible on `-Cpanic=unwind` and should really only be reserved to FFI (or otherwise wherever unwinding is silly, like inside the panic handler). You use this to stop unwinding, to instead abort the program immediately. In the case somebody actually uses this as a try-catch, this is probably a reasonable FN (and we should probably have a lint for this as well)

We could probably extend this to anything that is unconditionally `!` (which `catch_unwind` wouldn't be)

y21:
Yeah I agree catching panics really shouldnt be done for error handling à la try-catch, but panics are also per thread, so you can panic on one thread and the others (the rest of the program) will live on, and I would say a large number of rust programs involve threads in some way.
Tokio also uses catch_unwind so that stuff like webservers don't go down when a panic occurs. So imo a `panic!` is quite often not something that can stop the process (and therefore not always something that removes zombies), but if you think we should still do that anyway, I'm also fine with that.

Another question re. the proposed behavior: would this look at just the next statement after the spawn or how far are we willing to lookahead for an exit? All the way to the end of the enclosing block?
Would it consider `exit()` in an if too?
I imagine this would become kinda complex if we want to look across multiple statements, since some statements can diverge or exit only in an "unlikely" case

> Catching a panic is only possible on `-Cpanic=unwind`

Also to me this makes it sound like that's a rare thing to have but it's the default value ^^, so you have to manually set panic=abort to get it to not unwind


> We could probably extend this to anything that is unconditionally ! (which catch_unwind wouldn't be)

Not sure on this, since expressions like `return`, `continue`, `loop {}` etc. are also `!` but they don't necessarily stop the process (code could be getting into an intentional infinite loop that processes inputs in a loop forever or something like that)

Centri3:
Yeah I was wrong, always forget `!` extends to *anything* that is never constructed ^^

Also forgot `panic!` is specific to a thread; I think that's best to be propagated to other threads but yeah, we should still lint it in case of `panic!`s because of that

Couple thoughts on the behavior:
1. If it's the final call expression in the main function, it shouldn't lint since returning from main will (ignoring `#[start]`) effectively immediately exit the process.
2. It shouldn't consider conditional `exit`s, even something like `if true`
3. Any call between the `exit` and the `spawn` should be pessimistically considered always diverging, i.e., it'll never return

Stuff like loop expressions should ofc be considered infinitely looping. I think this doesn't have to extend to `for` loops though, as you'd have to intentionally make an iterator that never advances for that to loop forever.

clippy_lints/src/zombie_processes.rs
```diff
@@ -0,0 +1,253 @@
...
+    "not waiting on a spawned child process"
+}
+declare_lint_pass!(ZombieProcesses => [ZOMBIE_PROCESSES]);
+
+fn emit_lint(cx: &LateContext<'_>, span: Span) {
+    span_lint_and_then(
+        cx,
+        ZOMBIE_PROCESSES,
+        span,
+        "spawned process is never `wait()`-ed on and leaves behind a zombie process",
```
Centri3, line 49: Should mention this only happens on unix (or might leave behind), a line break would also be appreciated

clippy_lints/src/zombie_processes.rs
```diff
@@ -0,0 +1,253 @@
...
+        |diag| {
+            diag.help("consider calling `.wait()`")
+                .note("also see https://doc.rust-lang.org/stable/std/process/struct.Child.html#warning");
```
Centri3, line 52: This is a bit noisy imo:
```suggestion
            diag.help("see https://doc.rust-lang.org/stable/std/process/struct.Child.html#warning");
```
Ideally it'd also have a suggestion, but are there any cases where adding it to the end of the span wouldn't work?

clippy_lints/src/zombie_processes.rs
```diff
@@ -0,0 +1,253 @@
...
+                    // on the variable, and lint if not.
+                    // This is difficult to do because expressions can be arbitrarily complex
+                    // and the variable can "escape" in various ways, e.g. you can take a `&mut` to the variable
+                    // and call `.wait()` through that, or pass it to another function...
+                    // So instead we do the inverse, checking if all uses are either:
+                    // - a field access (`child.{stderr,stdin,stdout}`)
+                    // - calling `id` or `kill`
+                    // - no use at all (e.g. `let _x = child;`)
+                    // - taking a shared reference (`&`), `wait()` can't go through that
+                    // Neither of these is sufficient to prevent zombie processes
```
Centri3, line 76: very nitpicky:
```suggestion
                    // None of these are sufficient to prevent zombie processes
```

Centri3 (commented):
(meow meow), long wait sorry - Looks good overall

clippy_lints/src/zombie_processes.rs
y21: can you elaborate on what you mean by line break?

Centri3: Making it span 2 lines, it's a bit long for one imo. Not necessary though

y21: You mean 2 lines for the *message*? Does that actually work? Never seen that, but I'm also not really sure if that's what you mean (probably not). I suppose "might leave behind a zombie process" could be its own note. Then the message is shorter: just "spawned process is never `wait()`-ed on"

clippy_lints/src/zombie_processes.rs
y21: Yea, I guess we can have a suggestion for `let _ = <expr that spawns child>;` and `<expr that spawns child>;` specifically. But probably not for when it's assigned to a binding, since it's used elsewhere and finding an appropriate place to put the call seems tricky.

It should probably also be `MaybeIncorrect`. Maybe the user wants `kill()` instead 🤷

y21:
Had to rebase because the PR was getting too far behind. The last commit should contain the changes for (most of) the review comments. Also had to split the test up into two files since some of them have suggestions now

clippy_lints/src/zombie_processes.rs
Centri3: Now that I think about it, I don't think I have either. Yea that's what I meant. A separate note is good too :+1:

xFrednet:
I'll put this on my todo list for the next week :D

@y21 would you mind rebasing on master once again?

r? @xFrednet

tests/ui/zombie_processes.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // Conditionally exiting
+    // It should assume that it might not exit and still lint
+    {
+        let mut x = Command::new("").spawn().unwrap();
+        //~^ ERROR: spawned process is never `wait()`ed on
+        if true {
+            std::process::exit(0);
+        }
```
xFrednet, line 63: Could you also add a test like this:

```rs
let mut x = Command::new("").spawn().unwrap();
//~^ ERROR: spawned process is never `wait()`ed on
if true {
    return; // Maybe also `break` && `continue` but that might be overkill
}
x.wait().unwrap();
```

tests/ui/zombie_processes.rs
```diff
@@ -0,0 +1,87 @@
+#![warn(clippy::zombie_processes)]
+
+use std::process::{Child, Command};
+
+fn main() {
+    {
+        let mut x = Command::new("").spawn().unwrap();
```
xFrednet, line 7: Could you add one test with `#[allow]` on the command creation to ensure that it can be correctly suppressed at this node? I'm pretty sure it will be, but it's better to check this explicitly :)

xFrednet (commented):
The implementation looks good to me. I would like more tests to ensure that this doesn't result in several FPs, they might require some additional logic in the lint implementation :)

tests/ui/zombie_processes.rs
```diff
@@ -0,0 +1,87 @@
+#![warn(clippy::zombie_processes)]
+
+use std::process::{Child, Command};
+
+fn main() {
```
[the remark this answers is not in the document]
y21: > Can you add tests where the created process is returned from a function or stored in a struct?

A test for functions exists: https://github.com/rust-lang/rust-clippy/blob/81acbcb62fd457c325cf75de6271cab8db28b4f6/tests/ui/zombie_processes_fixable.rs#L19

I can add another one for structs and blocks.

I think those are currently false negatives if you don't call `wait()` there, but I assume you're more interested in the FP case where it *shouldn't* warn with a `wait()` call, like in the given example.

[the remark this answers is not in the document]
xFrednet: I think the test would be better without the `todo!()` macro:

```rust
fn spawn_proc() -> Child {
    Command::new("").spawn()
}
```

I want to make sure that this expression doesn't trigger the lint.

tests/ui/zombie_processes.rs
y21: Making it emit a warn here added quite a bit of complexity (adding another visitor), but done.

clippy_lints/src/zombie_processes.rs
```diff
@@ -0,0 +1,361 @@
...
+/// }
+/// c.wait(); // this might not be reachable
+/// ```
+enum WaitFinder<'a, 'tcx> {
+    WalkUpTo(&'a LateContext<'tcx>, HirId),
+    Found(&'a LateContext<'tcx>, HirId),
+}
+
+impl<'a, 'tcx> Visitor<'tcx> for WaitFinder<'a, 'tcx> {
+    type Result = ControlFlow<BreakReason>;
```
y21, line 122: Nothing related to this PR, but now that rustc supports returning `ControlFlow` from visitors for short-circuiting, maybe we should start doing this in more places. I think `for_each_expr` is used quite commonly and would benefit from this since it already expects the closure to return `ControlFlow`

tests/ui/zombie_processes.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    {
+        let mut x = Command::new("").spawn().unwrap();
+        if true {
+            x.wait().unwrap();
+            return;
+        }
+        x.wait().unwrap();
+    }
```
xFrednet, line 134: Could you add another test, where the `Command` is returned from the function with a return statement? In those cases we shouldn't warn and assume that wait is called on the returned value.

xFrednet (commented):
The update looks good. One comment and then it should be good to go =^.^=

The next review should also be a bit quicker again. Sorry for the delay.

y21:
Sorry for the wait, am a bit busy again, I'll try to get to the rest of this in 2-3 weeks

@rustbot author

xFrednet:
Hey @y21 this is another careful ping. Could you look at the last comment? The changes should be simple IIRC

tests/ui/zombie_processes.rs
y21: Do you mean just something like:
```rs
fn spawn_proc_2() -> Child {
    return Command::new("").spawn().unwrap();
}
```
?

(I guess I'm just a bit confused as to why you commented on these lines 127-134 specifically, because that test sounds unrelated to the selected test case?)

y21:
After having thought about this more for a while, I'm not sure if some of the logic in here is really necessary/worth the complexity.

For example, checking if the `.spawn()` call is the last expression in `main` is a fair bit of code for what I think is fairly unlikely to occur in practice.

The logic implies that you `.spawn()` a process and don't do *anything* with it (if you do, then it won't be the last statement of the program), which on its own doesn't make a lot of sense, because (without waiting) the host process will likely exit before the child process gets to do anything.

It could work if the spawned process finishes quickly enough, but I'd even go as far as saying that it probably should warn on that, exactly because it isn't waiting on the subprocess and the current process might exit prematurely depending on how long it takes. (Then the problem doesn't really have to do with zombie processes anymore, but it would still encourage adding a `.wait()`)

xFrednet:
This sounds reasonable to me, feel free to remove that part. If this really comes up as a comment, we can always add it as a limitation to the lint description. :)

tests/ui/zombie_processes.rs
y21: (not sure if you saw this message, but @xFrednet mentioning you just to make sure)

xFrednet: Nope, I missed this comment. It has nothing to do with this specific test, just that the `return` statement reminded me of this option :D

tests/ui/zombie_processes_fixable.rs
```diff
@@ -0,0 +1,26 @@
...
+fn not_main() {
+    Command::new("").spawn().unwrap();
+}
+
+fn spawn_proc() -> Child {
+    Command::new("").spawn().unwrap()
+}
+
+fn spawn_proc_2() -> Child {
+    return Command::new("").spawn().unwrap();
```
y21, line 25: Really wanted to use `#[expect(clippy::needless_return)]` here now that it's stable but #12998 :(

xFrednet: That PR should be ready tomorrow. Then we can do that before the merge :P

You can and probably should try `#[expect(clippy::zombie_processes)]`

xFrednet (commented):
Looks good to me :D

I've started the FCP: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/FCP.3A.20New.20lint.20.60zombie_processes.60.20rust-clippy.2311476/near/450804727

You can wait with the rebase until the FCP has concluded.

y21:
@xFrednet The FCP thread hasn't gotten comments but 4 upvote reactions and it's been about a month -- do you think we're good? I guess it's also waiting on me to do a final rebase.

xFrednet:
Ahh, I totally forgot about this Pr, yes we can count the FCP as accepted. You can `Roses = me` after a rebase :D

xFrednet (approved):

xFrednet (approved):

Issue #6425, opened 2020-12-06
Suggest handling case insensitivity on .ends_with(".ext")


kangalioo:
### What it does

When `.ends_with()` is called on a String or &str, and the parameter looks like a file extension (regex `\.([a-z0-9]{1,5})|([A-Z0-9]{1,5})`), inform the programmer that this code will not handle differently-cased filenames well -  potentially a bug. Additionally, a correct replacement is suggested (see below).

### Categories

- Kind: `clippy::pedantic` or perhaps `clippy::correctness`

*What is the advantage of the recommended code over the original code*

It handles file extensions correctly regardless of case. In Windows, this behavior is very much expected since the entire operating system has case-insensitive filenames. In Linux, filenames are usually case-sensitive, but I think even on Linux, it's standard behavior to recognize file extensions regardless of case.

### Drawbacks

- The suggested improvement is harder to read than the naive case-sensitive version
- The line `.ends_with(".ext")` might not be a file extension check at all. It could mean something else for which case-sensitivity makes sense

### Example

```rust
if !filename.ends_with(".abc") {
    panic!("Please provide an ABC file");
}
```

Could be written as:

```rust
if filename.rsplit('.').next().map(|ext| ext.eq_ignore_ascii_case("abc")) != Some(true) {
    panic!("Please provide an ABC file");
}
```

Or like this (nightly-only because of `Path::eq_ignore_ascii_case`):
```rust
let filename = Path::new(filename);
if filename.extension().map(|ext| ext.eq_ignore_ascii_case("abc")) != Some(true) {
    panic!("Please provide an ABC file");
}
```

flip1995:
If we should lint this, we should only lint the case, where `filename` is a `Path`. If we would also lint strings, this will have too many FPs, I think.

kangalioo:
So like, the lint should trigger only on this kind of thing?
```rust
// filename: Path
filename.ends_with(".ext")
```
Probably nobody is going to write this code, because `ends_with` on Path doesn't work with the file extension at all. So the above code would fail to return the expected value for almost _all_ inputs, so the programmer will know by themselves that they did something wrong.

The reason for this is that `Path::ends_with` [only considers whole path components to match](https://doc.rust-lang.org/std/path/struct.Path.html#method.ends_with).

If it would produce too many false positive's to have the lint trigger on strings, I don't think there's any purpose left for this lint suggestion.

flip1995:
[...]

PR #11483, opened 2023-09-11
new lint: `path_ends_with_ext`

### LINKED ISSUE #11479: path.ends_with(".ext") won't match file extension
### What it does

`Path.ends_with` works differently than `str.ends_with`, and an attempt to match a file extension can unexpectedly only compare to the whole file name.

### Advantage

It catches a potential gotcha.

### Drawbacks

`path.ends_with(".ext")` is a mistake when used with file name extensions, but could be valid for a dot-file, like `path.ends_with(".rustup")`.

### Example

```rust
fn is_markdown(path: &Path) -> bool {
    path.ends_with(".md")
}
```

Could be written as:

```rust
fn is_markdown(path: &Path) -> bool {
    path.extension().map_or(false, |ext| ext == "md") // then run clippy --fix to make it case-insensitive ;)
}
```

y21:
Closes #11479

Not sure if it needs more test cases. I couldn't come up with any other ones, but it is a pretty simple lint logic wise with not too many checks

changelog: new lint: [`path_ends_with_ext`]


Alexendoo:
This would have a fair few false positives with stuff like `.git` or any other dotfile type things

If we can source one shipping a big list of file extensions would probably be the ideal solution since it wouldn't take up all that much space, the problem being how easily it is to source such a list

y21:
Maybe a configuration option for additional dotfiles to allow would also be useful.

kornelski:
I suggest limiting it to 2-char and 3-char extensions. The longer the name, the more likely it is a dotfile.

I don't have any 2-char dotfiles. The 3-letter dotfiles I could find:

.git .svn .gem	.npm .vim .env .rnd .ssh	.vnc

y21:
> I suggest limiting it to 2-char and 3-char extensions. The longer the name, the more likely it is a dotfile.
> 
> I don't have any 2-char dotfiles. The 3-letter dotfiles I could find:
> 
> .git .svn .gem .npm .vim .env .rnd .ssh .vnc

That sounds great, I implemented this in the last commit (that and the ability to add custom dotfiles that clippy shouldn't warn).

clippy_lints/src/methods/path_ends_with_ext.rs
```diff
@@ -0,0 +1,50 @@
...
+use clippy_utils::ty::is_type_diagnostic_item;
+use rustc_ast::{LitKind, StrStyle};
+use rustc_data_structures::fx::FxHashSet;
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::sym;
+use std::fmt::Write;
+
+pub const DEFAULT_ALLOWED_FILE_EXTENSIONS: &[&str] = &["git", "svn", "gem", "npm", "vim", "env", "rnd", "ssh", "vnc"];
```
Alexendoo, line 15: Could also add `smb`, `nvm`, `bin`

clippy_lints/src/methods/path_ends_with_ext.rs
```diff
@@ -0,0 +1,50 @@
...
+    expr: &Expr<'_>,
+    msrv: &Msrv,
+    allowed_paths: &FxHashSet<String>,
+) {
+    if is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(recv).peel_refs(), sym::Path)
+        && !path.span.from_expansion()
+        && let ExprKind::Lit(lit) = path.kind
+        && let LitKind::Str(path, StrStyle::Cooked) = lit.node
+        && let Some(path) = path.as_str().strip_prefix('.')
+        && (path.len() == 2 || path.len() == 3)
```
Alexendoo, line 30: Is there a reason we're excluding single char extensions?

clippy_lints/src/utils/conf.rs
```diff
@@ -561,6 +561,11 @@ define_Conf! {
     /// Which crates to allow absolute paths from
     (absolute_paths_allowed_crates: rustc_data_structures::fx::FxHashSet<String> =
         rustc_data_structures::fx::FxHashSet::default()),
+    /// Lint: PATH_ENDS_WITH_EXT.
+    ///
+    /// Additional file extensions to allow
+    (allowed_file_extensions: rustc_data_structures::fx::FxHashSet<String> =
+        rustc_data_structures::fx::FxHashSet::default()),
```
Alexendoo, line 568: I think it's worth re-wording this, since they're not file extensions if they're added to this config option, they're files or directories

tests/ui/path_ends_with_ext.rs
```diff
@@ -0,0 +1,29 @@
...
+    //~^ ERROR: this looks like a failed attempt at checking for the file extension
+
+    // some "extensions" are allowed by default
+    path.ends_with(".git");
+
+    // most legitimate "dotfiles" are longer than 3 chars, so we allow them as well
+    path.ends_with(".bashrc");
+
+    path.ends_with(arg!());
+}
```
Alexendoo, line 21: Could do with some cases like `..`, `./a`

clippy_lints/src/methods/path_ends_with_ext.rs
y21: Not really, I just went with @kornelski's suggestion. It probably makes sense to lint single char extensions too.

tests/ui/path_ends_with_ext.rs
y21: Good point, I guess we shouldn't lint on these since its not an extension. Maybe we could just check if all chars in the path (without the dot) match `char::is_alphanumeric` or smth 🤔

CHANGELOG.md
```diff
@@ -5574,5 +5575,7 @@ Released 2018-09-13
 [`allow-one-hash-in-raw-strings`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allow-one-hash-in-raw-strings
 [`absolute-paths-max-segments`]: https://doc.rust-lang.org/clippy/lint_configuration.html#absolute-paths-max-segments
 [`absolute-paths-allowed-crates`]: https://doc.rust-lang.org/clippy/lint_configuration.html#absolute-paths-allowed-crates
+[`allowed-dotfiles`]: https://doc.rust-lang.org/clippy/lint_configuration.html#allowed-dotfiles
 [`enforce-iter-loop-reborrow`]: https://doc.rust-lang.org/clippy/lint_configuration.html#enforce-iter-loop-reborrow
 <!-- end autogenerated links to configuration documentation -->
+mentation -->
```
y21, line 5581: huh, I'm not sure what happened here..? I'm pretty sure I've seen this happen in another PR before.
Seems to happen when renaming a config option and re-running `cargo collect-metadata`? Probably worth looking into at some point another time

jdh8:
How is this different from #6425 :monocle_face:

y21:
> How is this different from #6425 🧐

The description and code looks like it has a little bit of overlap and sounds similar, but that one seems to look for a different kind of suspicious pattern and I would say this one is different enough to be its own lint.

This lint looks for `Path::ends_with` calls and checks if the argument looks like a file extension and warns the user that this isn't what `Path::ends_with` *does*, which has nothing to do with case sensitivity like the other lint IMO.
The other lint looks for `str::ends_with`, which at least does the right thing in that it actually checks the extension (just not case insensitive).

We may want to change the suggestion to use `str::eq_ignore_ascii_case` instead of `==` though.

clippy_lints/src/methods/mod.rs
```diff
@@ -3563,11 +3566,47 @@ declare_clippy_lint! {
...
+    /// File extensions are often two to three characters long, so this only lints in those cases
+    /// in an attempt to avoid false positives.
+    /// The `allowed-file-extensions` configuration can be used to allow additional
+    /// file extensions that Clippy should not lint.
+    ///
+    /// ### Why is this bad?
+    /// This doesn't actually compare file extensions. Rather, `ends_with` compares the given argument
+    /// to the last **component** of the path and checks if it matches exactly.
+    ///
+    /// ### Example
```
Alexendoo, line 3584: Probably worth sticking a known issues section in that mentions it won't pick up longer extension names, with that and a squash it looks good to go

y21: This is already mentioned in "What it does", but I guess that can be moved to its own "Known issues" section

Alexendoo (approved):

Alexendoo:
Thanks!

@bors r+

PR #11484, opened 2023-09-12
[`extra_unused_type_parameters`]: Fix edge case FP for parameters in where bounds

### LINKED ISSUE #11302: `extra_unused_type_parameters` + where clause requiring the parameter
### Summary

When a generic parameter is mentioned in a where clause, Clippy seems to not notice that.
The suggestion of removing the parameter would result in non-compiling code.

The lint tries to protect the user from signatures where rustc cannot infer the type.

Here is the full example that also shows rustc can indeed infer the type of `Signature`:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=902fccc27c7495faca8e0863ba13486b

### Lint Name

extra_unused_type_parameters

### Reproducer (incomplete, full code above)

```rs
fn store<T: 'static, Signature: 'static>(v: &mut Vec<Box<dyn Closure>>, func: T)
where
    WrappedFn<T, Signature>: Closure,
{
    v.push(Box::new(WrappedFn {
        inner: func,
        _p: PhantomData,
    }));
}
```
```
warning: type parameter `Signature` goes unused in function definition
  --> src/main.rs:50:20
   |
50 | fn store<T: 'static, Signature: 'static>(v: &mut Vec<Box<dyn Closure>>, func: T)
   |                    ^^^^^^^^^^^^^^^^^^^^ help: consider removing the parameter
   |
   = note: `#[warn(clippy::extra_unused_type_parameters)]` on by default
```

### Version

```text
Stable Playground with Rust 1.71.1
```

### Additional Labels

_No response_

mkrasnitski:
Generic parameters can end up being used on the left side of where-bounds if they are not directly bound but instead appear nested in some concrete generic type. Therefore, we should walk the left side of where bounds, but only if the bounded type is *not* a generic param, in which case we still need to ignore the bound.

Fixes #11302 

changelog: [`extra_unused_type_parameters`]: Fix edge case false positive for parameters in where bounds

### DIFF
--- a/clippy_lints/src/extra_unused_type_parameters.rs
+++ b/clippy_lints/src/extra_unused_type_parameters.rs
@@ -246,8 +246,13 @@ impl<'cx, 'tcx> Visitor<'tcx> for TypeWalker<'cx, 'tcx> {
                 {
                     self.ty_params.remove(&def_id);
                 }
+            } else {
+                // If the bounded type isn't a generic param, but is instead a concrete generic
+                // type, any params we find nested inside of it are being used as concrete types,
+                // and can therefore can be considered used. So, we're fine to walk the left-hand
+                // side of the where bound.
+                walk_ty(self, predicate.bounded_ty);
             }
-            // Only walk the right-hand side of where bounds
             for bound in predicate.bounds {
                 walk_param_bound(self, bound);
             }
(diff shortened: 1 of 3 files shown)


Jarcho:
Thank you. @bors r+

Issue #11494, opened 2023-09-13
enum_variant_names lint of a variant starting or ending with the name of the enum can depend on the order of the variants


jonboh:
### Summary

The lint bails out if the variant name is just a word. But this prevents linting later fields that contain the name of the Enum.
This is done to avoid linting prefix/postfix that would otherwise leave empty variants. I think this was introduced in #8324.

I found this while preparing a PR for #2555, I'll try to also submit a PR to fix this behavior. 

### Reproducer

This code will not lint the third variant name
```rusts
enum Data {
    Some,
    Other,
    DataTracker,
} 
```

I expected to see this happen:
`warning: variant name starts with the enum's name`

Instead, this happened:
Nothing

However this other definition will be linted correctly:
```
enum Data {
    DataTracker,
    Some,
    Other,
} 
```

### Version

```text
rustc 1.74.0-nightly (249595b75 2023-08-23)
binary: rustc
commit-hash: 249595b7523fc07a99c1adee90b1947739ca0e5b
commit-date: 2023-08-23
host: x86_64-unknown-linux-gnu
release: 1.74.0-nightly
LLVM version: 17.0.0
```

Issue #11481, opened 2023-09-11
allow-one-hash-in-raw-strings in clippy.toml not respected for needless_raw_string_hashes


gwy15:
### Summary

in https://doc.rust-lang.org/clippy/lint_configuration.html#allow-one-hash-in-raw-strings, there's a configuration which, as the name suggests, should allow using one hash for raw strings. But it is not currently respected for lint `needless_raw_string_hashes`.

### Reproducer

I tried this code:

`./clippy.toml`:
```
allow-one-hash-in-raw-strings = true
```

`./src/main.rs`:
```rust
const S: &str = r#"
\d+
"#;

fn main() {
    println!("s = {S}");
}
```

I expected to see this happen: No lint.

Instead, this happened:
```
warning: unnecessary hashes around raw string literal
 --> src/main.rs:1:17
  |
1 |   const S: &str = r#"
  |  _________________^
2 | | \d+
3 | | "#;
  | |__^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_raw_string_hashes
  = note: `#[warn(clippy::needless_raw_string_hashes)]` on by default
help: try
  |
1 ~ const S: &str = r"
2 + \d+
3 ~ ";
  |
```

### Version

```text
rustc 1.72.0 (5680fa18f 2023-08-23)
binary: rustc
commit-hash: 5680fa18feaa87f3ff04063800aec256c3d4b4be
commit-date: 2023-08-23
host: aarch64-apple-darwin
release: 1.72.0
LLVM version: 16.0.5
```

krtab:
A commit fixing this issue is in review in the following rust-lang/rust PR clippy subtree:
https://github.com/rust-lang/rust/pull/119552

Issue #1138, opened 2016-08-05
enum_variant_names: warn about prefix/postfix only if >N variants present


birkenfeld:
With enums that have only a few (2 or 3?) variants, the probability is large that the naming will not be improved by moving the common parts into the enum name.

An example is vulkano's `DeviceCreationError::OutOfHostMemory` and `DeviceCreationError::OutOfDeviceMemory` - it is a generic device creation error, not specifically an "out of memory on device creation" error, and it is incidental that the only two current cases are out-of-memory cases.

oli-obk:
Sounds reasonable to me to exclude enums with few variants. The variant count should obviously be configurable

Manishearth:
wfm

PR #11496, opened 2023-09-13
add lint for struct field names

### LINKED ISSUE #2555: complain if all fields of a struct have the same prefix or postfix
We already do this for enum variants. The code should be reusable

jonboh:
changelog: [`struct_field_names`]: lint structs with the same pre/postfix in all fields or with fields that are pre/postfixed with the name of the struct.

fixes #2555 

I've followed general structure and naming from the code in [enum_variants](https://github.com/rust-lang/rust-clippy/blob/b788addfcc955368b9771b77d312c248fab60253/clippy_lints/src/enum_variants.rs) lint, which implements the same logic for enum variants.


clippy_utils/src/str_utils.rs
```diff
@@ -236,6 +236,30 @@ pub fn count_match_end(str1: &str, str2: &str) -> StrCount {
...
 
+/// Returns a `snake_case` version of the input
+/// ```
+/// use clippy_utils::str_utils::to_snake_case;
+/// assert_eq!(to_snake_case("AbcDef"), "abc_def");
+/// assert_eq!(to_snake_case("ABCD"), "a_b_c_d");
+/// assert_eq!(to_snake_case("AbcDD"), "abc_d_d");
+/// assert_eq!(to_snake_case("Abc1DD"), "abc1_d_d");
+/// ```
+pub fn to_snake_case(name: &str) -> String {
```
y21, line 247: rustc has a similar function for its nonstandard_style lint: https://doc.rust-lang.org/stable/nightly-rustc/src/rustc_lint/nonstandard_style.rs.html#237-265
it would be nice if we could somehow reuse that logic, but it might be good enough for this PR (`enum_variant` also has its own version of to_camel_case).

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,168 @@
...
+        );
+    }
+}
+
+fn check_struct_start(cx: &LateContext<'_>, item_name: &str, field: &FieldDef<'_>) {
+    let item_name_snake = to_snake_case(item_name);
+    let name = field.ident.name.as_str();
+    let item_name_chars = item_name_snake.chars().count();
+
+    if count_match_start(&item_name_snake, name).char_count == item_name_chars
```
y21, line 126: Is this different from just `name.starts_with(&item_name_snake)`? `starts_with` would be easier to read imo

y21 (commented):
One question: would it make sense to allow `is_*` fields? I imagine this would be a common case where people would want to allow the lint

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,168 @@
...
+impl_lint_pass!(StructFields => [STRUCT_FIELD_NAMES]);
+
+fn is_snake_case(name: &str) -> bool {
+    name.chars().all(|c| !c.is_uppercase() || c == '_')
+}
+
+fn check_fields(cx: &LateContext<'_>, threshold: u64, fields: &[FieldDef<'_>], item_name: &str, item_span: Span) {
+    if (fields.len() as u64) < threshold {
+        return;
+    }
```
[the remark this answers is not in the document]
jonboh: Yes, it is expected, I've implemented the threshold imitating the behavior of the enum_variants lint and the reasoning given in the issue that generated the option (that in enums with low variant counts the change doesn't improve the quality of the names that much) https://github.com/rust-lang/rust-clippy/issues/1138. 
I think it makes sense to make the threshold affect at the lint as a whole. Otherwise it might make sense to have two independent lints.

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    for field in fields {
+        if is_snake_case(field.ident.name.as_str()) {
+            check_struct_start(cx, item_name, field);
+            check_struct_end(cx, item_name, field);
+        }
+    }
+
+    let first_field = fields[0].ident.name.as_str();
```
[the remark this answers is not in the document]
jonboh: right, fixed in [c652eca](https://github.com/rust-lang/rust-clippy/pull/11496/commits/c652eca2756cee4f6cbe1b7b3abc1d7bfb20e5a8)

I've created another PR fixing it in enum_variants: https://github.com/rust-lang/rust-clippy/pull/11552
I've tried setting you as reviewer as well, but it has failed for some reason.

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,168 @@
...
+use clippy_utils::str_utils::{count_match_end, count_match_start, to_snake_case};
+use rustc_hir::{FieldDef, ItemKind, VariantData};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+use rustc_span::Span;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects struct fields that are prefixed or suffixed
+    /// by the same characters.
```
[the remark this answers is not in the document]
jonboh: updated description on [05fd236](https://github.com/rust-lang/rust-clippy/pull/11496/commits/05fd236e5581db1493a5e287f27ffcf9c432e5d8)

In the issue that gave rise to the threshold I interpreted that both parts of the lint should be disabled when under the threshold. 
Also from the description of the option I think that's the intended meaning:
`The minimum number of enum variants for the lints about variant names to trigger`

jonboh:
> One question: would it make sense to allow `is_*` fields? I imagine this would be a common case where people would want to allow the lint

Yes, I agree it would make sense to silence the lint in this case. I could add that prefix as an exception so that it does not trigger the lint, but I wasn't sure if adding particular cases was appropiate. 

I think even in the case in which `is_*` makes sense it is possible to factor out that meaning into the struct name.
```
struct Data {
    is_correct: bool
    is_duplicaded: bool
    is_encrypted: bool
}

struct DataChecks {
    correct: bool
    duplicated: bool
    encrypted: bool
}
```
I don't have a strong preference for the second case. 
Do you think it is worthwhile adding the specific case for `is_*`?

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,168 @@
...
+            .iter()
+            .zip(field_split.iter().rev())
+            .take_while(|(a, b)| a == b)
+            .map(|e| *e.0)
+            .collect();
+    }
+    let prefix = pre.join("");
+    let postfix = post.join("");
+    if fields.len() > 1 {
+        let (what, value) = match (prefix.is_empty() || prefix == "_", postfix.is_empty()) {
```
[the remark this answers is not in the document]
jonboh: you are right, this check was only needed in the enum_variants case that uses CamelCase. In this case prefixes that are only `_` will already be empty (and not be linted, which was the reason why this check was introduced in enum_variants).

tests/ui/struct_fields.rs
```diff
@@ -0,0 +1,122 @@
+#![warn(clippy::struct_field_names)]
```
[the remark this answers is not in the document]
jonboh: thanks, those case were not being linted correctly. fixed on [9f54c59](https://github.com/rust-lang/rust-clippy/pull/11496/commits/9f54c59f2e223c84107631e8254f70511dff8654)

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,168 @@
...
+            (true, true) => return,
+            (false, _) => ("pre", prefix),
+            (true, false) => {
+                post.reverse();
```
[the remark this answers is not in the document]
jonboh: fixed on [9f54c59](https://github.com/rust-lang/rust-clippy/pull/11496/commits/9f54c59f2e223c84107631e8254f70511dff8654)

clippy_utils/src/str_utils.rs
jonboh: I agree it would be better to use a single implementation both from `struct_fields` and for `enum_variant`. However those functions (`to_snake_case` and `to_camel_case`) are private. I think it would make sense trying to factor them out into a style utilities (similar to `clippy_utils::str_utils`) module that is accessible from clippy. 
I haven't yet contributed to rustc so I don't know if making that change for a need in clippy would be accepted. I would happily prepare a PR in rustc and clippy that unifies these implementations into a single one in rustc.
I think (from memory) there are a bunch of places in clippy that have similar ad-hoc functions related to these transformations.

y21:
The idea of allowing `is_*` came from [this `#[allow]` here](https://github.com/rust-lang/rust-clippy/pull/11496/files#diff-96b9ee5ecc7448a11cf3413d24308d08e213d75cff7cb4c0e999476201b4d922L585), I was of the opinion that changing the `is_local` field to `local` would make it less clear. Thinking about it more though and looking at your examples, this might just be fine and we could also keep it like that

clippy_lints/src/struct_fields.rs
jonboh: yes, it's much better with starts/ends_with. fixed in [654cea6](https://github.com/rust-lang/rust-clippy/pull/11496/commits/654cea62fdffba12178a41968457b69571df3cf3)

clippy_utils/src/str_utils.rs
y21: Yeah I think we could move those casing functions to some kind of util module of a crate that both clippy and rustc_lint can depend on. `rustc_ast::util::case` is an option, and if there's a better place then the reviewer over at rust-lang/rust can say it, should be trivial to move around as these are very self contained. It's probably not worth blocking this PR on syncing those two up, that could be done later on after the PR has landed.

clippy_lints/src/struct_fields.rs
```diff
@@ -0,0 +1,164 @@
...
+fn is_snake_case(name: &str) -> bool {
+    name.chars().all(|c| !c.is_uppercase() || c == '_')
+}
+
+fn check_fields(cx: &LateContext<'_>, threshold: u64, fields: &[FieldDef<'_>], item_name: &str, item_span: Span) {
+    if (fields.len() as u64) < threshold {
+        return;
+    }
+
+    for field in fields {
```
y21, line 69: Is there a reason that this struct name prefix checking is its own loop, and not run together with the other loop that shrinks the shared pre/postfix in fields (like how enum_variant_names does it)?

y21 (commented):
We should also have a test for the configuation option (in tests/ui-toml) to make sure it works and continues to work to avoid issues such as #11481

y21:
Also I wonder if we can merge this file with enum_variants since they do seem very related and share a bit of logic. There's also other lints as part of the same lint pass such as module_name_repetitions, for catching a shared {pre,post}fix between modules and structs or other items

clippy_lints/src/struct_fields.rs
jonboh: Yes, it prevents this issue: https://github.com/rust-lang/rust-clippy/issues/11494
I opened a PR for enum_variants to fix it there as well: https://github.com/rust-lang/rust-clippy/pull/11498

y21: I see, makes sense.

jonboh:
Yeah, they could be in the same file, although it might be good to change the name `enum_variants` at that point. For me `enum_variants` its already kind of a misnomer when the file contains a bunch more things than checking the enum variants. 
Initially I thought it might be better to split the enum_variants file so that the logic related to module inception would be in a different file.

The module inception part already checks the struct names as well, so we could merge everything together with a more general name for the file (instead of `enum_variants`).

y21:
I agree that the lint pass name `enum_variants` is not great for what it is now. All lints in it are checking for common prefixes and suffixes in items (not necessarily enums). Renaming it to something more general and putting `struct_field_names` in it sounds good. I'm not sure what a good name for it could be. Maybe `item_name_repetitions`? Would include both modules, struct (fields) and enum (variants)

y21 (commented):
The change from enum_variants to item_name_repetitions is great, I like that. It's probably worth including another changelog entry in the description that says that the `enum_variant_name_threshold` config option was renamed to `item_name_threshold`.

It also looks like the error in CI happens because it's testing that each directory in `tests/ui-toml` has an associated test, but `tests/ui-toml/item_name_repetitions` doesn't have any direct test files, only directories. Might have to flatten this to `item_name_repetitions_threshold_0` and `item_name_repetitions_threshold_5` 🤔

jonboh:
> The change from enum_variants to item_name_repetitions is great, I like that. It's probably worth including another changelog entry in the description that says that the `enum_variant_name_threshold` config option was renamed to `item_name_threshold`.
> 
> It also looks like the error in CI happens because it's testing that each directory in `tests/ui-toml` has an associated test, but `tests/ui-toml/item_name_repetitions` doesn't have any direct test files, only directories. Might have to flatten this to `item_name_repetitions_threshold_0` and `item_name_repetitions_threshold_5` 🤔

I think it was failing because I had one stderr file with the old name that did not have any associated test file. I think I've fixed it now. I'm going to check that the tests are actually running in the CI (running `cargo test` both of them run)

y21 (commented):
Some last comments from my side, mostly just nits and a question :)

There's still some duplicate looking code in `check_{variants,fields}`, but it's subtly different in places that I'm not sure it's easily refactorable into its own function, so I'm also not sure it's worth blocking the PR over it (but if you have ideas for addressing this, feel free to do that too). There's the other PR which makes parts of `check_variant` match the behavior from `check_fields` more closely, so it might be easier to refactor it with those changes.

I also did a lintcheck run and there didn't appear to be any FPs (actually there weren't any matches at all, which is not necessarily a bad thing), though only on the 26 default crates

@Alexendoo do you have some other things to comment on that I might have missed?

Alexendoo:
`cargo lintcheck --recursive --filter struct_field_names` gives some results to look at, didn't look through all of them but there were some that could do with looking at like definitions in macro expansions

y21:
Oh, I forgot `--recursive` was a thing (but that's on me, it is documented in the readme now that I'm reading it). I did another lintcheck run but with `--recursive` and the top 100 crates this time and that does indeed show a lot more matches (135), though a lot of them are from libc in macro expansions.
A lot of them are also legitimate warnings, but I'm now wondering if this is a bit too strong for a style lint?

I looked through all of them and here's some interesting cases worth mentioning:

### Shared prefix of struct and field might be in the middle of a word
```rs
struct Foo {
  food: i32,
  a: i32,
  b: i32
}
```
Should something like this lint? It does right now, because it's a simple starts_with check, and I'm thinking we might want to use the same logic we're using for the fields: splitting the field name and snake case struct name by `_` and use `take_while` to look for the longest prefix. With that logic, it wouldn't lint here. 

### A field that has the same name as the struct
Another case that seems to be warned on a lot:
```rs
struct Proxy {
  proxy: i32,
  unrelated1: bool,
  unrelated2: bool
}
```
I'm thinking this might just be fine to lint. This could use `inner` as the field name or something. But right now it lints twice, once because it has the same prefix *and* the same postfix. This should at the very least only lint once.

### Structs in macro expansions
It currently lints if the struct is from a macro expansion. This happens a lot in libc because they have a macro that takes idents and creates a struct from it. I'm a bit split on this. We should definitely not lint if it's from a foreign macro expansion that we don't get to change or control, but imo it might still make sense to lint if the struct name and all field names are from the same syntax context. For example,
```rs
macro_rules! mk_struct {
  () => {
    struct Foo {
      foo_a: i32,
      foo_b: i32,
      foo_c: i32,
    }
  }
}
mk_struct!();
```
It might be nice to lint here since we own the macro and can change it, but as soon as only some of the fields or the struct name are `ident` macro inputs or are from other expansions, we probably shouldn't.

### Two underscores
Another instance in libc
```rs
    pub struct regex_t {
        __buffer: i32,
        __allocated: i32,
        __used: i32,
    }
```
This complains that all fields have the same prefix `_`. Similar to how it already ignores a single `_` prefix, maybe it should ignore this too?

jonboh:
## Middle word prefix
Yeah, this doesn't make sense to lint. Now the check is done in terms of words so that case should not be linted anymore.

## Field with the same name as the struct
Fixed by only checking the end if the field has more words than the struct. (otherwise it will be linted with the check from the beginning of the field)

--
As a result of the changes in these two points I've merged the `check_struct_start/end` into `check_struct_name_repetition` as they had to share a bit of logic and having them separate didn't make sense at that point.

## Structs in macro expansion
I've added the `is_present_in_source` check so I understand only macros that we "own" will be linted now. I haven't worked much with macros so let me know if there's some extra checks that we have to add.

## Underscore prefix
Now prefixes consisting only of underscores should no longer be linted

y21:
> I've added the `is_present_in_source` check so I understand only macros that we "own" will be linted now. I haven't worked much with macros so let me know if there's some extra checks that we have to add.

Hmm, I'm not *sure* if that function does what we want. It looks like that checks if we have access to the source of that span, which, to me seems like it could sometimes still return true for some macros from other crates. 
I was thinking of [`clippy_utils::macros::span_is_local`](https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/macros/fn.span_is_local.html), which should be a more reliable way of checking if the span is from the current crate.

But, in some cases that span check for the whole item alone is probably not enough. The struct field names themselves might be from a macro, too, so those need to be checked as well.

To make this more general, I feel like linting on something like the following could potentially open a can of worms, so maybe we should not:
```rs
macro_rules! mk_struct {
  ($field:ident) => {
    struct Foo {
      $field: i32,
      // ... 2 more unrelated fields ...
    }
  }
}

mk_struct!(foo); // expands to `struct Foo { foo: i32, ... }`
```
That is, a local macro that takes in an `ident`ifier and creates a struct with the given identifier as a field. Since the struct name and field name that share a prefix are from different "places" (macro input and macro body), I think it might not always be possible to fix the warning (changing the struct)

So, I was thinking if we could make sure to only lint if the place (the syntax context) is the same for the item, the struct name and all field names. In practice that would mean that, for example, that all the struct pieces are all defined in the same macro (or outside of any macros entirely) and not from mixed expansions like in my example above. We could do that by getting the syntax context for the item (`item_span.ctxt()`) and the field ident (`field.ident.span.ctxt()`) (or the struct name, depending on what we're checking) and then comparing them. If they're not the same, avoid linting

There would also ofc be the option to just completely ignore macros altogether by adding something like `if span.from_expansion() { return; }`, and that would be a safe option (prevents FPs), but I think it'd be nice to lint in some macro cases?

-------

Also, unrelated but it looks like you (accidentally?) committed an empty foo.txt file? 😄

jonboh:
I've added a bunch of tests that I hope capture the behavior that you described. I agree that it makes sense to lint only when the origin of the identifiers is the same.

Right now, checking the `SyntaxContext` of the identifiers of both the struct and fields something like this gets linted:
```
    macro_rules! mk_struct_full_def {
        ($struct_name:ident, $field1:ident, $field2:ident, $field3:ident) => {
            struct $struct_name {
                $field1: i32,
                $field2: i32,
                $field3: i32,
            }
        };
    }
    mk_struct_full_def!(PrefixData, some_data, some_meta, some_other);
```
It is possible to avoid linting this case by checking if the fields SyntaxContext is the same as the one from the `Item` itself (not its identifier). This might be overdoing it, but in essence the user has control over these names and they don't come from different places (just not the usual place in which you'd expect this lint).

There are still a lot of places in libc in which the lint triggers. Mostly in code like this:
```
s! {
    pub struct in_pktinfo {
        pub ipi_ifindex: ::c_int,
        pub ipi_spec_dst: ::in_addr,
        pub ipi_addr: ::in_addr,
    }
}
```
I would say that in these cases (libc) it makes sense to disable the lint (with an #[allow()]). In the general case, I think its ok to lint this.

y21:
Yeah, IMO it seems fine to lint the `mk_struct_full_def` case because the user gets to control all the idents from the same context. And yes, libc is kind of a special case by itself because it does a lot of stuff that normal Rust code probably doesn't do. So it seems fine for crates like libc to just allow the lint if needed

tests/ui/struct_fields.rs
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// should not lint
+pub struct RegexT {
+    __buffer: i32,
+    __allocated: i32,
+    __used: i32,
+}
+
+mod macro_tests {
```
y21, line 172: To make sure that the `span_is_local` check works, we should also have a test with external spans. We could do that by using the `external` proc macro, which just echos the given tokens but with an external span. Something like:
```rs
external! {
  struct ExternalFoo {
    external_foo_a: i32,
    external_foo_b: i32,
    external_foo_c: i32,
  }
}
```
There are a lot of other tests that use this macro, so could take some inspiration from there

clippy_lints/src/item_name_repetitions.rs
```diff
@@ -123,12 +159,119 @@ impl EnumVariantNames {
...
+            .map(|e| e.0)
+            .collect();
+    }
+    let prefix = pre.join("_");
+    post.reverse();
+    let postfix = post.join("_");
+    if fields.len() > 1 {
+        let (what, value) = match (
+            prefix.is_empty() || prefix.chars().all(|c| c == '_'),
+            postfix.is_empty(),
```
y21, line 218: This should also check if the postfix is just underscores like with prefix above, right? It currently ignores a single underscore postfix, but not two

clippy_lints/src/item_name_repetitions.rs
```diff
@@ -123,12 +159,119 @@ impl EnumVariantNames {
...
+                        cx,
+                        STRUCT_FIELD_NAMES,
+                        field.hir_id,
+                        field.span,
+                        "field name starts with the struct's name",
+                    );
+                }
+                if field_words.len() > item_name_words.len() {
+                    // lint only if the end is not covered by the start
+                    if field_words
```
y21, line 255:
```suggestion
                } else if field_words
```
I found this logic a bit confusing to understand. It seems like we can just use `else if` here to only lint same postfix if we didn't lint a prefix? Or is this doing something else that I misunderstood?

y21 (commented):
The way it deals with macros now seems good to me 👍
Just some minor comments.

clippy_lints/src/item_name_repetitions.rs
y21: Do you disagree with this suggested change? I see it was marked as resolved, but the code is still the same :)

clippy_lints/src/utils/conf.rs
```diff
@@ -356,10 +356,10 @@ define_Conf! {
     ///
     /// The maximum size of objects (in bytes) that will be linted. Larger objects are ok on the heap
     (too_large_for_stack: u64 = 200),
-    /// Lint: ENUM_VARIANT_NAMES.
+    /// Lint: ENUM_VARIANT_NAMES, STRUCT_FIELD_NAMES.
     ///
-    /// The minimum number of enum variants for the lints about variant names to trigger
-    (enum_variant_name_threshold: u64 = 3),
+    /// The minimum number of enum variants or struct fields for the lints about variant and field names to trigger
+    (item_name_threshold: u64 = 3),
```
Alexendoo, line 362: This will need a `#[conf_deprecated]` entry if it's being renamed, though it would also be reasonable to have a separate config variable for structs and enums to avoid that

clippy_lints/src/item_name_repetitions.rs
y21: ok, the fact that your comments were in the pending state lead to a pretty confusing conversation here 😅, but now that I can actually see the replies:

I see the problem now, thanks for clarifying that. You mean something like
```rs
struct Foo {
  foo_something_foo: i32,
  a: i32,
  b: i32
}
```
would only be linted for its shared prefix and not also its postfix if we changed this to `else if`. Yeah, makes sense to keep it like this then and tell the user right away that it has a common prefix + postfix

jonboh: yeah, good thing [zulip](https://rust-lang.zulipchat.com/) exists 😅

clippy_lints/src/item_name_repetitions.rs
y21: Right, yeah, I guess we could also lint single `_` postfix. I'm not sure if anyone does that, since it doesn't have a specific meaning, unlike a *pre*fix `_`. It was just odd to me that it would only ignore one underscore specifically.

jonboh: ok, after [1c891ea](https://github.com/rust-lang/rust-clippy/pull/11496/commits/1c891ea4d17a24bfafd973e76250b3746009556a) any `_` postfix is linted

clippy_lints/src/utils/conf.rs
jonboh: I've splitted the option to avoid having to deprecate the other one

clippy_lints/src/utils/conf.rs
```diff
@@ -359,7 +359,11 @@ define_Conf! {
     /// Lint: ENUM_VARIANT_NAMES.
     ///
     /// The minimum number of enum variants for the lints about variant names to trigger
-    (enum_variant_name_threshold: u64 = 3),
+    (enum_name_threshold: u64 = 3),
```
y21, line 362: The config variable from the other lint is still different now. This should probably be changed back to `enum_variant_name_threshold` (else we'd need the `#[conf_deprecated]`, and avoiding it is why we changed back in the first place, right?).

clippy_lints/src/item_name_repetitions.rs
```diff
@@ -103,32 +104,184 @@ declare_clippy_lint! {
...
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// struct Cake {
+    ///     sugar: u8,
+    ///     flour: u8,
+    ///     eggs: u8
+    /// }
+    /// ```
+    #[clippy::version = "1.74.0"]
```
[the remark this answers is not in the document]
y21: (or we can also just leave it, I just learned that this gets corrected in the changelog PRs anyway)

clippy_lints/src/utils/conf.rs
jonboh: my bad, I forgot the variant part of the name, I should have look at the PR diff

y21:
I did one more lintcheck run for the new changes, 31 results this time. The issues I pointed out in an earlier comment seem to be resolved now. Most of the matches are in libc and linux-raw-sys now, which seems fine to me, and the rest are legitimate warnings. Might want to squash commits a bit though. @Alexendoo do you have something else to comment on? 

<details>
<summary>Lintcheck logs</summary>

```
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.112/src/unix/linux_like/linux/gnu/mod.rs:353:9 clippy::struct_field_names "all fields have the same prefix: `si`"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.134/src/unix/linux_like/linux/gnu/mod.rs:385:9 clippy::struct_field_names "all fields have the same prefix: `si`"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.139/src/unix/linux_like/linux/gnu/mod.rs:385:9 clippy::struct_field_names "all fields have the same prefix: `si`"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.148/src/unix/linux_like/linux/gnu/mod.rs:390:9 clippy::struct_field_names "all fields have the same prefix: `si`"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/libc-0.2.149/src/unix/linux_like/linux/gnu/mod.rs:390:9 clippy::struct_field_names "all fields have the same prefix: `si`"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/thiserror-impl-1.0.49/src/attr.rs:23:5 clippy::struct_field_names "field name ends with the struct's name"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/typenum-1.17.0/build/op.rs:15:5 clippy::struct_field_names "field name starts with the struct's name"
$CARGO_HOME/registry/src/index.crates.io-6f17d22bba15001f/unicode-normalization-0.1.22/src/stream_safe.rs:66:5 clippy::struct_field_names "field name starts with the struct's name"
target/lintcheck/sources/aho-corasick-1.1.1/src/util/special.rs:10:1 clippy::struct_field_names "all fields have the same postfix: `id`"
target/lintcheck/sources/anyhow-1.0.75/src/error.rs:595:1 clippy::struct_field_names "all fields have the same prefix: `object`"
target/lintcheck/sources/idna-0.4.0/src/uts46.rs:26:1 clippy::struct_field_names "all fields have the same prefix: `byte`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:134:1 clippy::struct_field_names "all fields have the same prefix: `l`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:143:1 clippy::struct_field_names "all fields have the same prefix: `l`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:278:1 clippy::struct_field_names "all fields have the same postfix: `files`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:292:1 clippy::struct_field_names "all fields have the same prefix: `fsx`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:337:1 clippy::struct_field_names "all fields have the same prefix: `nr`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:426:1 clippy::struct_field_names "all fields have the same prefix: `ru`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:473:1 clippy::struct_field_names "all fields have the same prefix: `sa`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:481:1 clippy::struct_field_names "all fields have the same prefix: `ss`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:619:1 clippy::struct_field_names "all fields have the same prefix: `c`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:629:1 clippy::struct_field_names "all fields have the same prefix: `c`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:641:1 clippy::struct_field_names "all fields have the same prefix: `c`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:653:1 clippy::struct_field_names "all fields have the same prefix: `ws`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:661:1 clippy::struct_field_names "all fields have the same prefix: `c`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:767:1 clippy::struct_field_names "all fields have the same prefix: `d`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:798:1 clippy::struct_field_names "all fields have the same prefix: `st`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:813:1 clippy::struct_field_names "all fields have the same prefix: `f`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:829:1 clippy::struct_field_names "all fields have the same prefix: `f`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:845:1 clippy::struct_field_names "all fields have the same prefix: `f`"
target/lintcheck/sources/linux-raw-sys-0.5.4/src/x86_64/general.rs:876:1 clippy::struct_field_names "all fields have the same prefix: `sa`"
target/lintcheck/sources/rustls-0.22.0-alpha.3/src/crypto/cipher.rs:61:1 clippy::struct_field_names "all fields have the same postfix: `len`"


### Stats:

| clippy::struct_field_names                         |    31 |


### ICEs:

```
</details>

jonboh:
Is there any general guideline to what extent squash the commits? In my last PR I just squashed everything into a single commit.

clippy_lints/src/item_name_repetitions.rs
```diff
@@ -103,32 +104,184 @@ declare_clippy_lint! {
...
+
+    check_struct_name_repetition(cx, item, fields);
+
+    // if the SyntaxContext of the identifiers of the fields and struct differ dont lint them.
+    // this prevents linting in macros in which the location of the field identifier names differ
+    if !fields.iter().all(|field| item.ident.span.eq_ctxt(field.ident.span)) {
+        return;
+    }
+
+    let mut pre: Vec<&str> = match fields.get(0) {
```
y21, line 194:
```suggestion
    let mut pre: Vec<&str> = match fields.first() {
```
now that #11609 is merged, this will need to be fixed

clippy_lints/src/item_name_repetitions.rs
```diff
@@ -103,32 +104,184 @@ declare_clippy_lint! {
...
+    /// Detects struct fields that are prefixed or suffixed
+    /// by the same characters or the name of the struct itself.
+    ///
+    /// ### Why is this bad?
+    /// Information common to all struct fields is better represented in the struct name.
+    ///
+    /// ### Limitations
+    /// Characters with no casing will be considered when comparing prefixes/suffixes
+    /// This applies to numbers and non-ascii characters without casing
+    /// e.g. `Foo1` and `foo2` is considered to have different prefixes
```
y21, line 118: I'm a bit confused by this as it is currently written. Should `Foo1` here be `foo1` (lowercase)?

tests/ui-toml/item_name_repetitions/threshold0/item_name_repetitions.rs
```diff
@@ -0,0 +1,3 @@
+struct Data {}
```
y21, line 1: Can we also add `enum Actions {}` here? This PR also deletes the enum_variant_threshold0 uitest from the other PR, so right now this doesn't test empty enums anymore.

y21 (commented):
Did another quick look through the code (missed these things in my earlier reviews, sorry). Nothing big though.
Also regarding squashing (I forgot to respond 😅), I don't think it's mandatory or that there are any rules around this other than that commit messages should be descriptive and that there shouldn't be commits like "fmt", "fix tests", etc.

tests/ui-toml/item_name_repetitions/threshold0/item_name_repetitions.rs
jonboh: ah yes, I hadn't initially added because it always failed (because of the ice fixed in the other pr). During the rebase I must have missed it

clippy_lints/src/item_name_repetitions.rs
jonboh: yes, missed that one

Alexendoo:
I usually squash to a single commit unless I think it would be helpful to have them split up

One thing to bring up again is the lint group, did we still want to move it to `pedantic`?

Otherwise @bors delegate=y21

y21:
Right, the lint group. I'm not 100% sure here. `enum_variant_names` is a style lint, so it seems consistent to have them in the same category, but on the other hand, `module_name_repetitions` is in pedantic.
I would put this in pedantic since it does seem like there are cases where it's not so clear (as can be seen by the `#[allow]`s on this PR alone) and leave it there for a bit, since that seems like a safe choice and we can always move it. Happy to also hear @jonboh's opinion though

tests/ui-toml/item_name_repetitions/threshold0/clippy.toml
```diff
@@ -0,0 +1 @@
+struct-field-name-threshold = 0
```
y21, line 1:
```suggestion
struct-field-name-threshold = 0
enum-variant-name-threshold = 0
```
It seems you forgot to add the config option for enums

y21: Also, while we're at it, can the `item_names.rs` file in the threshold 5 test also be renamed `item_name_repetitions.rs` for consistency?

y21 (commented):
Before we can merge this, one last comment about the empty enum test. After this I think we're good. 😄 
(Also, the changelog in the description still says that a config option was renamed. Not true anymore)

tests/ui-toml/item_name_repetitions/threshold0/clippy.toml
jonboh: should be fixed now :)

y21 (approved):
Alright, this looks ready to be merged 🎉. Thank you for implementing this lint and for your patience!

@bors r+

y21:
🤔 @bors ping

y21:
Let's try again? @bors r+

y21:
> ERROR: PR body must contain 'changelog: ...'

I guess it doesn't like that linebreak after changelog? @jonboh can you try removing that?

jonboh:
linebreak removed.

Thank you for taking the time to review the PR :D

y21:
Yep, just running `cargo collect-metadata` should do it all for you

y21:
Third time's the charm. @bors retry

y21:
🎉 

That took a few tries.
Thanks again for your work here @jonboh !

PR #11506, opened 2023-09-15
fix filter_map_bool_then with a bool reference

### LINKED ISSUE #11503: `filter_map_bool_then` generates incorrect autofix when the `bool` in question is a reference
### Summary

Method calls implicitly dereference `self`, so `then` can be called on (possibly nested) references to `bool`. The autofix for `filter_map_bool_then` doesn’t take this into account and tries to return references to `bool` from `filter`.

### Reproducer

I tried this code:

```rust
fn main() {
    let bools: &[bool] = &[true, false, false, true];
    let result: Vec<usize> = bools
        .iter()
        .enumerate()
        .filter_map(|(i, b)| b.then(|| i))
        .collect();
    dbg!(result);
}
```

`cargo clippy --fix` should generate code like this:
```rust
fn main() {
    let bools: &[bool] = &[true, false, false, true];
    let result: Vec<usize> = bools
        .iter()
        .enumerate()
        .filter(|&(i, b)| *b)
        .map(|(i, b)| i)
        .collect();
    dbg!(result);
}
```

Instead, this happened:
```
warning: failed to automatically apply fixes suggested by rustc to crate `project`

  * src/main.rs

https://github.com/rust-lang/rust-clippy/issues

 --> src/main.rs:6:27
  |
6 |         .filter(|&(i, b)| b).map(|(i, b)| i)
  |                           ^ expected `bool`, found `&bool`
  |
help: consider dereferencing the borrow
  |
6 |         .filter(|&(i, b)| *b).map(|(i, b)| i)
  |                           +

warning: usage of `bool::then` in `filter_map`
 --> src/main.rs:6:10
  |
6 |         .filter_map(|(i, b)| b.then(|| i))
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `filter` then `map` instead: `filter(|&(i, b)| b).map(|(i, b)| i)`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#filter_map_bool_then
  = note: `#[warn(clippy::filter_map_bool_then)]` on by default
```

### Version

```text
rustc 1.74.0-nightly (8142a319e 2023-09-13)
binary: rustc
commit-hash: 8142a319ed5c1d1f96e5a1881a6546e463b77c8f
commit-date: 2023-09-13
release: 1.74.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

mojave2:
changelog: [`filter_map_bool_then`]: Fix the incorrect autofix when the `bool` in question is a reference.

fix #11503

### DIFF
--- a/clippy_lints/src/methods/filter_map_bool_then.rs
+++ b/clippy_lints/src/methods/filter_map_bool_then.rs
@@ -8,7 +8,7 @@ use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::{LateContext, LintContext};
 use rustc_middle::lint::in_external_macro;
-use rustc_middle::ty::Binder;
+use rustc_middle::ty::{self, Binder};
 use rustc_span::{sym, Span};
 
 pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, arg: &Expr<'_>, call_span: Span) {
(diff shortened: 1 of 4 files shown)


giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #11485, opened 2023-09-12
[`transmute_null_to_fn`] does not recognize chain of casts


J-ZhengLi:
### Summary

Just happend to see the code and realize that it didn't take a chain of `as` into consideration, see the reproducer, or [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=056c207bbfc133ee623afc514bdbc377), if one of the cast gets removed, this lint will work as normal.

### Lint Name

transmute_null_to_fn

### Reproducer

I tried this code:

```rust
let null_fn: fn() = unsafe {
    std::mem::transmute( 0 as *const i8 as *const u8 )
};
```

I expected to see this happen:

gives warning

Instead, this happened:

nothing happened

J-ZhengLi:
I believe the `ExprKind::Cast`s need to be peeled off before passing to `is_integer_literal` in [this line](https://github.com/rust-lang/rust-clippy/blob/98363cbf6a7c3f8b571a7d92a3c645bb4376e4a6/clippy_lints/src/transmute/transmute_null_to_fn.rs#L41)

PR #11507, opened 2023-09-15
trigger [`transmute_null_to_fn`] on chain of casts

J-ZhengLi:
fixes: #11485 

changelog: trigger [`transmute_null_to_fn`] on chain of casts


dswij (approved):
Nice spot! Thanks for this

Issue #11474, opened 2023-09-08
explicit_auto_deref interaction with ManuallyDrop


talagrand:
### Summary

Explicit derefs are recommended when using ManuallyDrop types, but the lint does not factor this in when recommending to auto deref.

### Lint Name

explicit_auto_deref

### Reproducer

Sample repro:

```rust

pub struct Variant {
    pub anonymous: Variant0,
}
pub union Variant0 {
    pub anonymous: std::mem::ManuallyDrop<Variant00>,
}

pub struct Variant00 {
    pub anonymous: Variant000,
}

pub union Variant000 {
    pub val: i32,
}

fn main() {
    unsafe {
        let mut p = core::mem::zeroed::<Variant>();
        (*p.anonymous.anonymous).anonymous.val = 1;
    }
}
```

Warning produced:

```
warning: deref which would be done by auto-deref
  --> src\main.rs:19:9
   |
19 |         (*p.anonymous.anonymous).anonymous.val = 1;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `p.anonymous.anonymous`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#explicit_auto_deref
   = note: `#[warn(clippy::explicit_auto_deref)]` on by default

```

When attempting to fix this we get:

```
error: not automatically applying `DerefMut` on `ManuallyDrop` union field
  --> src\main.rs:19:9
   |
19 |         p.anonymous.anonymous.anonymous.val = 1;
   |         ^^^^^^^^^^^^^^^^^^^^^
   |
   = help: writing to this reference calls the destructor for the old value
   = help: add an explicit `*` if that is desired, or call `ptr::write` to not run the destructor
```

So it seems like this case should be ignored by the rule, given the recommendation to explicitly deref.

### Version

```text
rustc 1.70.0-dev
binary: rustc
commit-hash: 23b6f0bea6
commit-date: 2023-06-19
host: x86_64-pc-windows-msvc
release: 1.70.0-dev
LLVM version: 16.0.2
```

PR #11508, opened 2023-09-15
Lint `needless_borrow` and `explicit_auto_deref` on most union field accesses

Jarcho:
Changes both lints to follow rustc's rules around auto-deref through `ManuallyDrop` union fields rather than just bailing on union fields.

changelog: [`needless_borrow`] & [`explicit_auto_deref`]: Lint on most union field accesses

### DIFF
--- a/clippy_lints/src/dereference.rs
+++ b/clippy_lints/src/dereference.rs
@@ -5,6 +5,7 @@ use clippy_utils::ty::{implements_trait, peel_mid_ty_refs};
 use clippy_utils::{
     expr_use_ctxt, get_parent_expr, get_parent_node, is_lint_allowed, path_to_local, DefinedTy, ExprUseNode,
 };
+use core::mem;
 use rustc_ast::util::parser::{PREC_POSTFIX, PREC_PREFIX};
 use rustc_data_structures::fx::FxIndexMap;
 use rustc_errors::Applicability;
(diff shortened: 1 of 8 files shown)


blyxyas:
It seems like #11477 fixed #11474. Is this PR still applicable?

Jarcho:
It's still relevant. Both the PR and the issue in question seem to have some misunderstandings why the suggestion doesn't compile. Auto-deref only doesn't apply on `ManuallyDrop` fields of unions when the result is used to access a field. In other words, `x.y.z` isn't valid when `x` is a union, `y` is a `ManuallyDrop` field, and `z` is used mutably.

blyxyas (commented):
The first commit is looking pretty good, just some preliminary feedback

clippy_lints/src/dereference.rs
```diff
@@ -1022,53 +1055,61 @@ fn report<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, state: State, data
...
+                // field expressions when the base type is a union and the parent
+                // expression is also a field access.
+                //
+                // e.g. `&mut x.y.z` where `x` is a union, and accessing `z` requires a
+                // deref through `ManuallyDrop<_>` will not compile.
+                let parent_id = cx.tcx.hir().parent_id(expr.hir_id);
+                if parent_id == data.first_expr.hir_id {
+                    return;
+                }
+                (cx.tcx.hir().get(parent_id).expect_expr().span, true)
```
blyxyas, line 1094: Could you replace this `expect_expr` with a fallible alternative? Or is there a guarantee that the parent id is an expression?

tests/ui/needless_borrow.stderr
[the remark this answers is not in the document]
Jarcho: GitHub isn't showing code here, so I assume you're referring to the mid expression blocks. The final expression in a block is moved before the parent expression runs. Since rustc only disallows `x.y.z` not any other form (e.g. `{ x.y }.z`) tests around that are needed.

clippy_lints/src/dereference.rs
Jarcho: There has to be at least one deref expression as direct parents.

tests/ui/needless_borrow.rs
```diff
@@ -190,27 +190,48 @@ fn issue9383() {
...
+            u: Wrap(ManuallyDrop::new(Foo { x: 0 })),
+        };
+        let _ = &mut (&mut x.u).x;
+        let _ = &mut (&mut { x.u }).x;
+        let _ = &mut ({ &mut x.u }).x;
+
+        let mut x = U { u: Wrap(Foo { x: 0 }) };
+        let _ = &mut (&mut x.u).x;
+        let _ = &mut (&mut { x.u }).x;
+        let _ = &mut ({ &mut x.u }).x;
```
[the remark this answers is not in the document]
Jarcho: I thought we weren't going to enforce this until `ui_test` worked the way we wanted it.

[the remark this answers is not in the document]
blyxyas: If I understand correctly, the conclusions from [the Zulip thread talking about this](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Improve.20UI.20error.20checks) were that we soft-mandated `//~ERROR` annotations (without a body). Even if they aren't obligated by the CI, I just find them to increase legibility :cat:

clippy_lints/src/dereference.rs
```diff
@@ -703,6 +713,14 @@ fn try_parse_ref_op<'tcx>(
...
 }
 
+// Checks if the adjustments contains a deref of `ManuallyDrop<_>`
+fn adjust_derefs_manually_drop<'tcx>(adjustments: &'tcx [Adjustment<'tcx>], mut ty: Ty<'tcx>) -> bool {
+    adjustments.iter().any(|a| {
+        let ty = mem::replace(&mut ty, a.target);
+        matches!(a.kind, Adjust::Deref(Some(ref op)) if op.mutbl == Mutability::Mut) && is_manually_drop(ty)
+    })
+}
+
```
blyxyas, line 723: I'm not sure about the logic of this function. The original `ty` passed into the function is never read because from the first iteration it is already replaced with an adjustment.

Also, the doc-comment isn't a doc-comment

blyxyas:
I think this two commits can be squashed, as all changes made to `clippy_lints/src/dereference.rs` made by the first one are deleted by the second one.

clippy_lints/src/dereference.rs
Jarcho: The first iteration doesn't necessarily happen.

Jarcho: The result of `mem::replace` on the first iteration is the passed in value.

Jarcho:
The first commit made changes to some logic. The second commit uses the changed logic, but extracted into a function. Diffs sometimes do a bad job at showing things.

blyxyas (approved):
LGTM, thanks! :heart:

PR #11511, opened 2023-09-15
Split `needless_borrow` into two lints

Jarcho:
Splits off the case where the borrow is used as a generic argument to a function. I think the two cases are different  enough to warrant a separate lint.

The tests for the new lint have been reordered to group related parts together. Two warning have been dropped, one looked like it was testing the generic argument form, but it ends up triggering the auto-deref variant. The second was just a redundant test that didn't do anything interesting.

An issue with cycle detection is also included. The old version was checking if a cycle was reachable from a block when it should have been checking if the block is part or a cycle.

As a side note, I'm liking the style of just jamming all the tests into separate scopes in main.

changelog: Split off `needless_borrows_for_generic_args` from `needless_borrow`


llogiq:
The code looks good, but I wonder if we shouldn't name the new lint `needless_generic_arg_borrow` (without the s) to be more alike the `needless_borrow` lint. Alternative (but longer) would be `needless_borrow_by_generic_arg`.

Jarcho:
`needless_borrow` really should plural according to the lint naming guidelines. `needless_borrows_for_generic_args` would work.

llogiq:
Thank you!

@bors r+

PR #11514, opened 2023-09-16
triagebot no-merges: exclude "Rustup"s, add labels

pitaj:
https://github.com/rust-lang/triagebot/pull/1720

changelog: none

### DIFF
--- a/triagebot.toml
+++ b/triagebot.toml
@@ -11,6 +11,8 @@ allow-unauthenticated = [
 
 # Have rustbot inform users about the *No Merge Policy*
 [no-merges]
+exclude_titles = ["Rustup"] # exclude syncs from rust-lang/rust
+labels = ["has-merge-commits", "S-waiting-on-author"]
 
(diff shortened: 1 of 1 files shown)


xFrednet:
Clippy doesn't have the `has-merge-commits` label, not sure if it's worth adding it. :thinking:  The `Rustup` title check might be nice, though.

r? @flip1995 since you usually do the updates. Do you want to have fewer conversations with `@rustbot`? ^^

pitaj:
Adding the label is low cost IMO. It's a visible reminder to the reviewer that can't get lost in a sea of comments, and bors may eventually become aware of it (reusing to merge a PR with it).

flip1995:
Thanks! I added the label (copied from rust repo, same color and everything)

Will this PR work before merging the triagebot PR or will it break triagebot? r=me once the triagebot PR is merge or if this already works without that PR @bors delegate+

pitaj:
It technically won't break anything but I'll wait until the triagebot change is merged.

PR #11515, opened 2023-09-16
[`filter_map_bool_then`]: peel as many refs as needed

y21:
In #11506 this lint was improved to suggest one deref if the bool is behind references (fixed the FP #11503), however it might need multiple dereferences if the bool is behind multiple layers of references. E.g. `&&&bool` needs `***b`.

changelog: [`filter_map_bool_then`]: peel as many references as needed and suggest multiple dereferences


Jarcho:
Thank you. @bors r+

PR #11518, opened 2023-09-17
fix FP with needless_raw_string_hashes

### LINKED ISSUE #11420: False negative with needless_raw_string_hashes
### Summary

The `needless_raw_string_hashes` warning isn't warning me about some cases I thought it would.

### Lint Name

needless_raw_string_hashes

### Reproducer

I tried this code:

```rust
let _y = r###"wea ther"###;
```

I expected it to tell me that three hashes were unnecessary, and that it could actually be no hashes, and suggest:

```rust
let _y = r"wea ther";
```

which also compiles.

Instead, this happened:

Clippy did not give me any warnings.

### Version

```text
rustc 1.72.0 (5680fa18f 2023-08-23)
binary: rustc
commit-hash: 5680fa18feaa87f3ff04063800aec256c3d4b4be
commit-date: 2023-08-23
host: aarch64-apple-darwin
release: 1.72.0
LLVM version: 16.0.5
```

mojave2:
changelog: Fix [`needless_raw_string_hashes`]: Continue the lint checking of raw string when `needless_raw_strings` is allowed.

fix #11420

### DIFF
--- a/clippy_lints/src/raw_strings.rs
+++ b/clippy_lints/src/raw_strings.rs
@@ -96,8 +96,9 @@ impl EarlyLintPass for RawStrings {
                         );
                     },
                 );
-
-                return;
+                if !matches!(cx.get_lint_level(NEEDLESS_RAW_STRINGS), rustc_lint::Allow) {
+                    return;
+                }
             }
 
             let req = {
--- a/clippy_lints/src/attrs.rs
+++ b/clippy_lints/src/attrs.rs
@@ -616,7 +616,7 @@ fn check_should_panic_reason(cx: &LateContext<'_>, attr: &Attribute) {
             attr.span,
             "#[should_panic] attribute without a reason",
             "consider specifying the expected panic",
-            r#"#[should_panic(expected = /* panic message */)]"#.into(),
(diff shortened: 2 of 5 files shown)


clippy_lints/src/raw_strings.rs
```diff
@@ -96,8 +96,9 @@ impl EarlyLintPass for RawStrings {
                         );
                     },
                 );
-
-                return;
+                if !matches!(cx.get_lint_level(NEEDLESS_RAW_STRINGS), rustc_lint::Allow) {
```
J-ZhengLi, line 99: There's a clippy wrapper called `is_lint_allowed`, it should work here with the expr's HirId, I think.

mojave2: It looks that `is_lint_allowed` need a `LateContext` argument, but we have only `EarlyContext` here.

J-ZhengLi: oh~ right, my bad

Alexendoo (approved):

Alexendoo:
Thanks! @bors r+

PR #11521, opened 2023-09-17
[`map_identity`]: allow closure with type annotations

### LINKED ISSUE #9122: False positive in map_identity
### Summary

With the code below, clippy suggests removing the `map_err()` call. However, in this case `map_err()` is the only thing telling the compiler the type of the `E` in the relevant  `Result<T, E>`, so removing it will result in an error. This lint is machine-applicable, so that generates an error.

I guess this regressed from when #8478 was fixed.

### Lint Name

map_identity

### Reproducer

I tried this code:

```rust
#[test]
fn test() {
    block_on_all(async move { Ok(()) })
        .map_err(|err: std::io::Error| err)
        .unwrap();
}

fn block_on_all<F>(f: F) -> F::Output
where
    F: std::future::Future,
{
    current_thread_runtime().block_on(f)
}

fn current_thread_runtime() -> tokio::runtime::Runtime {
    let mut builder = tokio::runtime::Builder::new_current_thread();

    #[cfg(feature = "aio")]
    builder.enable_io();

    builder.build().unwrap()
}
```

I saw this happen:

```
warning: unnecessary map of the identity function
 --> tests/foo.rs:3:40
  |
3 |       block_on_all(async move { Ok(()) })
  |  ________________________________________^
4 | |         .map_err(|err: std::io::Error| err)
  | |___________________________________________^ help: remove the call to `map_err`
  |
  = note: `#[warn(clippy::map_identity)]` on by default
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#map_identity
```

I expected to see this happen: nothing, or potentially it could instead suggest turbofishing the `Ok()` call, as in `Ok::<_, std::io::Error>(())`.

### Version

```text
binary: rustc
commit-date: 2022-06-27
host: aarch64-apple-darwin
release: 1.62.0
LLVM version: 14.0.5
```

### Additional Labels

y21:
Fixes #9122

`.map(|a: u32| a)` can help type inference, so we should probably allow this and not warn about "unnecessary map of the identity function"

changelog: [`map_identity`]: allow closure with type annotations

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -2027,36 +2027,31 @@ pub fn is_must_use_func_call(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
     did.map_or(false, |did| cx.tcx.has_attr(did, sym::must_use))
 }
 
-/// Checks if an expression represents the identity function
-/// Only examines closures and `std::convert::identity`
+/// Checks if a function's body represents the identity function. Looks for bodies of the form:
+/// * `|x| x`
+/// * `|x| return x`
+/// * `|x| { return x }`
+/// * `|x| { return x; }`
 ///
-/// Closure bindings with type annotations and `std::convert::identity` with generic args
-/// are not considered identity functions because they can guide type inference,
-/// and removing it may lead to compile errors.
-pub fn is_expr_identity_function(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
-    /// Checks if a function's body represents the identity function. Looks for bodies of the form:
-    /// * `|x| x`
-    /// * `|x| return x`
-    /// * `|x| { return x }`
-    /// * `|x| { return x; }`
-    fn is_body_identity_function(cx: &LateContext<'_>, func: &Body<'_>) -> bool {
-        let id = if_chain! {
-            if let [param] = func.params;
-            if let PatKind::Binding(_, id, _, _) = param.pat.kind;
-            then {
-                id
-            } else {
-                return false;
-            }
-        };
+/// Consider calling [`is_expr_untyped_identity_function`] or [`is_expr_identity_function`] instead.
+fn is_body_identity_function(cx: &LateContext<'_>, func: &Body<'_>) -> bool {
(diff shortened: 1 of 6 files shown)


clippy_utils/src/lib.rs
```diff
@@ -2007,6 +2007,10 @@ pub fn is_must_use_func_call(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
 
 /// Checks if an expression represents the identity function
 /// Only examines closures and `std::convert::identity`
+///
+/// Closure bindings with type annotations and `std::convert::identity` with generic args
+/// are not considered identity functions because they can guide type inference,
+/// and removing it may lead to compile errors.
 pub fn is_expr_identity_function(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
```
llogiq, line 2014: Please note that this utils function might be used from other lints, too (even if it isn't now). So this might inadvertently introduce false negatives. I'd be more comfortable with having the check directly in the lint.

y21: This is used only by all the other `*_identity` lints, so would need to be repeated in the other lints as well. So I thought it'd make more sense to have it here

y21: A (somewhat arbitrary) example for how the same issue happens in `flat_map_identity`:
```rs
[[]].iter().flat_map(|a: &[i32; 0]| a);
```
clippy recommends replacing it with `.flatten()`, however that fails to compile.

Maybe we could keep the behavior of the current function as it is and have another one with a more obvious name? 🤔

y21: I moved this new "breaking" logic into its own util function with a hopefully clear name. The old util function (even if unused now) will continue to behave like it used to, but references the new function in its doc comment because that's likely what one wants to use.
The diff is a bit confusing now because I moved the inner function out so that it can be called from both functions, and the deindenting must be confusing git.

What do you think about this change? There shouldn't and won't be any accidental FNs now this way I think.

llogiq:
Thank you!

@bors r+

PR #11522, opened 2023-09-17
[`redundant_guards`]: lint if the pattern is on the left side

y21:
A tiny improvement to the `redundant_guards` lint. There's no associated issue for this, just noticed it while going through the code.
Right now it warns on `Some(x) if x == 2` and suggests `Some(2)`, but it didn't do that for `Some(x) if 2 == x` (i.e. when the local is on the right side and the pattern on the left side).

changelog: [`redundant_guards`]: also lint if the pattern is on the left side


giraffate (approved):

giraffate:
@bors r+

Thanks!

PR #11523, opened 2023-09-17
used_underscore_bindings: respect lint levels on the binding definition

### LINKED ISSUE #947: Fix `used_underscore_binding`
In #944, `used_underscore_binding` was made allow and to ignore `_result` as a ugly hack. This should be fixed.

### LINKED ISSUE #11520: Check if `clippy::used_underscore_binding` is allowed on the underscore field
### Description

It would be cool, if the [clippy::used_underscore_binding](https://rust-lang.github.io/rust-clippy/master/index.html#/used_underscore_binding) lint would also check the lint level at the field itself. This would make it simple to allow the lint, if it's used by a derive macro

### Example

```
#![warn(clippy::used_underscore_binding)]

pub struct Bar {
    // This allow should also suppress the lint
    // vvv
    #[allow(clippy::used_underscore_binding)]
    _cool: u32,
}

impl Bar {
    pub fn something(&self) {
        println!("{}", self._cool);
    } 
}

```

### Hints:

The lint level at a given note, can be checked with [TyCtxt::lint_level_at_node](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/context/struct.TyCtxt.html#method.lint_level_at_node) accessed via `cx.tcx.lint_level_at_node` with the HirId of the field. If the level is `Level::Expect`, the expectation should be marked as fulfilled by calling [`LateContext::fulfill_expectation`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_lint/context/trait.LintContext.html#method.fulfill_expectation)

### Version

```text
1.74.0
```

### Additional Information

Context: https://github.com/idanarye/rust-typed-builder/issues/113

Alexendoo:
Fixes #11520
Fixes #947

Also ignores usages of `PhantomData`

changelog: none

### DIFF
--- a/clippy_lints/src/misc.rs
+++ b/clippy_lints/src/misc.rs
@@ -1,24 +1,23 @@
-use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg, span_lint_hir_and_then};
+use clippy_utils::diagnostics::{span_lint, span_lint_and_sugg, span_lint_and_then, span_lint_hir_and_then};
 use clippy_utils::source::{snippet, snippet_opt, snippet_with_context};
+use clippy_utils::sugg::Sugg;
+use clippy_utils::{
+    any_parent_is_automatically_derived, fulfill_or_allowed, get_parent_expr, in_constant, is_integer_literal,
+    is_lint_allowed, is_no_std_crate, iter_input_pats, last_path_segment, SpanlessEq,
+};
 use if_chain::if_chain;
 use rustc_errors::Applicability;
+use rustc_hir::def::Res;
 use rustc_hir::intravisit::FnKind;
 use rustc_hir::{
-    self as hir, def, BinOpKind, BindingAnnotation, Body, ByRef, Expr, ExprKind, FnDecl, Mutability, PatKind, Stmt,
-    StmtKind, TyKind,
+    BinOpKind, BindingAnnotation, Body, ByRef, Expr, ExprKind, FnDecl, Mutability, PatKind, QPath, Stmt, StmtKind, Ty,
+    TyKind,
 };
-use rustc_lint::{LateContext, LateLintPass};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
 use rustc_session::{declare_tool_lint, impl_lint_pass};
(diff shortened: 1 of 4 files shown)


clippy_lints/src/misc.rs
```diff
@@ -257,46 +256,65 @@ impl<'tcx> LateLintPass<'tcx> for LintPass {
...
+            && !name.starts_with("__")
+            && let definition_span = cx.tcx.hir().span(definition_hir_id)
+            && !definition_span.from_expansion()
+        {
+            let (level, _) = cx.tcx.lint_level_at_node(USED_UNDERSCORE_BINDING, definition_hir_id);
+            match level {
+                Level::Allow => return,
+                Level::Expect(id) => {
+                    cx.fulfill_expectation(id);
+                    return;
```
xFrednet, line 302: This check might miss a second expectation, if there is also one on the statement level. I think in 99.9% of the cases it should be fine, but it might be safer to also check the level at `expr` and potentially fulfill a second expectation. :)

Otherwise very nice check :)

Alexendoo: Aha makes sense, does the `ForceWarn` expectation ever need to be fulfilled?

xFrednet: The force warn expectation should be handled exactly the same as a normal one. That one becomes relevant if there is an `#[expect]` and the user specifies a `--force-warn` in via the console arguments

Alexendoo: Am I understanding correctly that `--force-warn foo` turns all the `#[expect(foo)]`s into `Level::ForceWarn(Some(_))`?

Personally I would still expect the lint to emit if `--force-warn` was passed, but it could probably still fulfill the expectation to not create a new warning/error

xFrednet: Yeah, sorry, I typed my answer on the go :sweat_smile:. With `--force-warn` the lint should be emitted as normal, that will emit the lint and fulfill the lint expectation IIRC

xFrednet (approved):

xFrednet:
LGTM, thank you for the update. I really like the `fulfill_or_allowed` function :+1: 

Looking forward to have this fix in the next rustup of Marker :D

@bors r+

PR #11527, opened 2023-09-17
new lint: `iter_without_into_iter`

### LINKED ISSUE #9736: Suggest implementing `IntoIterator for &'_ T` where `T` has `iter()` method returning an iterator (or vice versa)
### What it does

If the trait implementation is missing it suggests adding it by forwarding to the method (can be automated).
If the trait implementation is present but the inherent method is not suggest adding it forwarding to trait implementation.
Same for mutable versions.

Both of these APIs are idiomatic.

Probably should be only triggered for public `T`.

### Lint Name

impl_iter_into_iterator

### Category

style

### Advantage

- The API is idiomatic and intuitive because all types in `std` and many other crates behave the same
- It's possible to pass `&T` to `for` or functions expecting `IntoIterator` avoiding typing `.iter()`

### Drawbacks

- Increases boilerplate in the library

### Example

```rust
pub struct Foo([u8; 32]);

impl Foo {
    fn iter(&self) -> Iter<'_> { // Iter implements Iterator<Item=&u8>
        Iter::new(&self.0)
    }
}

```

Could be written as:

```rust
pub struct Foo([u8; 32]);

impl Foo {
    fn iter(&self) -> Iter<'_> { // Iter implements Iterator<Item=&u8>
        Iter::new(&self.0)
    }
}

impl<'a> IntoIterator for &'a Foo {
    type IntoIter = Iter<'a>;
    type Item = &'a u8;

        self.iter()
    }
}
```

y21:
Closes #9736

A new lint that looks for `iter` (and `iter_mut`) method implementations without the type implementing `IntoIterator` for `&Type`.
Imo this seems rather pedantic, so I went with that, but I can be convinced to change it to `style` like the linked issue asked for.
Writing a machine applicable suggestion seems a bit tricky and tedious, so for now this relies on the user adding remaining lifetimes.

changelog: new lint: `iter_without_into_iter`

### DIFF
--- a/clippy_lints/src/iter_without_into_iter.rs
+++ b/clippy_lints/src/iter_without_into_iter.rs
@@ -1,0 +1,135 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::get_parent_as_impl;
+use clippy_utils::source::snippet;
+use clippy_utils::ty::{implements_trait, make_normalized_projection};
+use rustc_errors::Applicability;
+use rustc_hir::{FnRetTy, ImplItemKind, ImplicitSelfKind, TyKind};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Looks for `iter` and `iter_mut` methods without an associated `IntoIterator for (&|&mut) Type` implementation.
+    ///
+    /// ### Why is this bad?
+    /// It's not bad, but having them is idiomatic and allows the type to be used in for loops directly
+    /// (`for val in &iter {}`), without having to first call `iter()` or `iter_mut()`.
+    ///
+    /// ### Example
+    /// ```rust
+    /// struct MySlice<'a>(&'a [u8]);
+    /// impl<'a> MySlice<'a> {
+    ///     pub fn iter(&self) -> std::slice::Iter<'a, u8> {
+    ///         self.0.iter()
+    ///     }
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// struct MySlice<'a>(&'a [u8]);
+    /// impl<'a> MySlice<'a> {
+    ///     pub fn iter(&self) -> std::slice::Iter<'a, u8> {
+    ///         self.0.iter()
+    ///     }
+    /// }
+    /// impl<'a> IntoIterator for &MySlice<'a> {
+    ///     type Item = &'a u8;
+    ///     type IntoIter = std::slice::Iter<'a, u8>;
+    ///     fn into_iter(self) -> Self::IntoIter {
+    ///         self.iter()
+    ///     }
+    /// }
+    /// ```
+    #[clippy::version = "1.74.0"]
+    pub ITER_WITHOUT_INTO_ITER,
+    pedantic,
+    "implementing `iter(_mut)` without an associated `IntoIterator for (&|&mut) Type` impl"
+}
+declare_lint_pass!(IterWithoutIntoIter => [ITER_WITHOUT_INTO_ITER]);
+
+/// Checks if a given type is nameable in a trait (impl).
+/// RPIT is stable, but impl Trait in traits is not (yet), so when we have
+/// a function such as `fn iter(&self) -> impl IntoIterator`, we can't
+/// suggest `type IntoIter = impl IntoIterator`.
+fn is_nameable_in_impl_trait(ty: &rustc_hir::Ty<'_>) -> bool {
+    !matches!(ty.kind, TyKind::OpaqueDef(..))
+}
+
+impl LateLintPass<'_> for IterWithoutIntoIter {
+    fn check_impl_item(&mut self, cx: &LateContext<'_>, item: &rustc_hir::ImplItem<'_>) {
+        let item_did = item.owner_id.to_def_id();
+        let (borrow_prefix, expected_implicit_self) = match item.ident.name {
+            sym::iter => ("&", ImplicitSelfKind::ImmRef),
+            sym::iter_mut => ("&mut ", ImplicitSelfKind::MutRef),
+            _ => return,
+        };
+
+        if let ImplItemKind::Fn(sig, _) = item.kind
+            && let FnRetTy::Return(ret) = sig.decl.output
+            && is_nameable_in_impl_trait(ret)
+            && cx.tcx.generics_of(item_did).params.is_empty()
+            && sig.decl.implicit_self == expected_implicit_self
+            && sig.decl.inputs.len() == 1
+            && let Some(imp) = get_parent_as_impl(cx.tcx, item.hir_id())
+            && imp.of_trait.is_none()
+            && let sig = cx.tcx.liberate_late_bound_regions(
+                item_did,
+                cx.tcx.fn_sig(item_did).instantiate_identity()
+            )
+            && let ref_ty = sig.inputs()[0]
+            && let Some(into_iter_did) = cx.tcx.get_diagnostic_item(sym::IntoIterator)
+            && let Some(iterator_did) = cx.tcx.get_diagnostic_item(sym::Iterator)
+            && let ret_ty = sig.output()
+            // Order is important here, we need to check that the `fn iter` return type actually implements `IntoIterator`
+            // *before* normalizing `<_ as IntoIterator>::Item` (otherwise make_normalized_projection ICEs)
+            && implements_trait(cx, ret_ty, iterator_did, &[])
+            && let Some(iter_ty) = make_normalized_projection(
+                cx.tcx,
+                cx.param_env,
+                iterator_did,
+                sym!(Item),
+                [ret_ty],
+            )
+            // Only lint if the `IntoIterator` impl doesn't actually exist
+            && !implements_trait(cx, ref_ty, into_iter_did, &[])
+        {
+            let self_ty_snippet = format!("{borrow_prefix}{}", snippet(cx, imp.self_ty.span, ".."));
+
+            span_lint_and_then(
+                cx,
+                ITER_WITHOUT_INTO_ITER,
+                item.span,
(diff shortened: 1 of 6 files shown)


clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,149 @@
...
+        && let TyKind::Path(QPath::Resolved(_, path)) = imp.self_ty.kind
+        && let Some(ty_did) = path.res.opt_def_id()
+        && let middle_ty = cx.tcx.type_of(ty_did).skip_binder()
+        && let ref_ty = cx.tcx.mk_ty_from_kind(ty::Ref(cx.tcx.lifetimes.re_erased, middle_ty, expected_mtbl))
+        && let Some(into_iter_did) = cx.tcx.get_diagnostic_item(sym::IntoIterator)
+        && let bound_vars = cx.tcx.late_bound_vars(item.hir_id())
+        // We need to erase late bounds regions so that `-> Iter<'_, T>` works
+        && let ret_middle_ty = cx.tcx.erase_late_bound_regions(
+            ty::Binder::bind_with_vars(hir_ty_to_ty(cx.tcx, ret), bound_vars)
+        )
```
y21, line 84: I'm not 100% confident if this is right. `implements_trait` just returns false if we're passing it a type like `Iter<'_, T>`, probably because of this:
https://github.com/rust-lang/rust-clippy/blob/251a475b724a81ffcfd0113820492a329feeef5c/clippy_utils/src/ty.rs#L242-L243

And erasing the regions seems to make it work. Is it right to use the bound vars from the fn?

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Checks if a given type is nameable in a trait (impl).
+/// RPIT is stable, but impl Trait in traits is not (yet), so when we have
+/// a function such as `fn iter(&self) -> impl IntoIterator`, we can't
+/// suggest `type IntoIter = impl IntoIterator`.
+fn is_nameable_in_impl_trait(ty: &rustc_hir::Ty<'_>) -> bool {
+    !matches!(ty.kind, TyKind::OpaqueDef(..))
+}
+
+fn check_for_mutability(cx: &LateContext<'_>, item: &rustc_hir::ImplItem<'_>, expected_mtbl: Mutability) {
```
Jarcho, line 61: This whole thing really should be inlined.

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,149 @@
...
+            cx.param_env,
+            into_iter_did,
+            sym!(Item),
+            [ret_middle_ty]
+        )
+        // Only lint if the `IntoIterator` impl doesn't actually exist
+        && !implements_trait(cx, ref_ty, into_iter_did, &[])
+    {
+        let (self_ty_snippet, into_iter_snippet) = with_forced_trimmed_paths!((
+            format!("{borrow_prefix}{}", cx.tcx.def_path_str(ty_did)),
```
Jarcho, line 99: You can take a snippet from the impl block to get this instead.

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,149 @@
...
+            cx.tcx.def_path_str(into_iter_did)
```
Jarcho, line 100: Should just be `IntoIterator`.

clippy_lints/src/iter_without_into_iter.rs
Jarcho: You'll want to use `liberate_late_bound_regions` on the functions signature and get the types from there.

Jarcho (changes requested):

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,149 @@
...
+        // Order is important here, we need to check that the `fn iter` return type actually implements `IntoIterator`
+        // *before* normalizing `<_ as IntoIterator>::Item` (otherwise make_normalized_projection ICEs)
+        && implements_trait(cx, ret_middle_ty, into_iter_did, &[])
+        && let Some(ret_iter_ty) = make_normalized_projection(
+            cx.tcx,
+            cx.param_env,
+            into_iter_did,
+            sym!(Item),
+            [ret_middle_ty]
+        )
```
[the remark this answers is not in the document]
y21: [That doesn't seem to exist](https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/ty/fn.make_normalized_projection.html?search=try_make_normalized_projection)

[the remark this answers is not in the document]
Jarcho: I thought I added that at the same time.

clippy_lints/src/iter_without_into_iter.rs
y21: Did that in a separate commit since it creates annoying diffs that probably make it harder to review.

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,134 @@
...
+            && let sig = cx.tcx.liberate_late_bound_regions(
+                item_did,
+                cx.tcx.fn_sig(item_did).skip_binder()
+            )
+            && let ref_ty = sig.inputs()[0]
+            && let Some(into_iter_did) = cx.tcx.get_diagnostic_item(sym::IntoIterator)
+            && let ret_ty = sig.output()
+            // Order is important here, we need to check that the `fn iter` return type actually implements `IntoIterator`
+            // *before* normalizing `<_ as IntoIterator>::Item` (otherwise make_normalized_projection ICEs)
+            && implements_trait(cx, ret_ty, into_iter_did, &[])
```
Jarcho, line 85: Missed this last time. This should be checking for `Iterator`, not `IntoIterator`. This is unlikely to cause an FP in practice.

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -0,0 +1,134 @@
...
+            && let FnRetTy::Return(ret) = sig.decl.output
+            && is_nameable_in_impl_trait(ret)
+            && cx.tcx.generics_of(item_did).params.is_empty()
+            && sig.decl.implicit_self == expected_implicit_self
+            && sig.decl.inputs.len() == 1
+            && let Some(imp) = get_parent_as_impl(cx.tcx, item.hir_id())
+            && imp.of_trait.is_none()
+            && let sig = cx.tcx.liberate_late_bound_regions(
+                item_did,
+                cx.tcx.fn_sig(item_did).skip_binder()
```
Jarcho, line 78: This should be `instantiate_identity`, not `skip_binder`. They currently are (and will probably continue to be) the same, but logically you want the instantiation.

Jarcho (changes requested):
Just two minor things, otherwise this looks good. Feel free to squash everything down.

Jarcho:
Thank you. @bors r+

Kixunil:
The reason I picked `style` is that not doing this causes `for x in &container` to not work. Also do I understand correctly that this doesn't do the opposite - suggest `iter()` when there's `IntoIter` for `&T`?

y21:
Right, this doesn't lint the opposite, must have missed that when reading the issue 😅. Could reopen the issue then, and mention that only the lint for a missing `IntoIterator` impl is implemented now

Kixunil:
Yes please, reopen.

y21:
I can't do that since I'm not on the team and don't have perms for it. I think you as the issue creator can do it though? Or is that different when the issue is autoclosed by a PR? 🤔 I'm not sure.
If not, maybe we need a new issue for it, or someone else with permissions can do it

y21:
Regardless, I'm going to try to implement the other version later today and submit another PR for it

Kixunil:
@y21 I couldn't, thanks anyway!

lopopolo:
Hey folks, the suggested fix this lint produces doesn't compile in at least 5 ways on some toy code I've tried:

- https://github.com/rust-lang/rust-clippy/issues/11692

It looks like the suggested code doesn't have any tests asserting that it compiles.

y21:
As I've said in the issue, it's expected that the current suggestion does not compile (it even has a comment up top), and has the unspecified applicability. It's mostly meant as a nudge in the right direction to the user, not an autofix.
So testing the suggestion doesn't really make sense since we know it won't compile in any case. 

You do make some valid points in the issue and at least some trivial things should have been included in the suggestion, and I'll try to look into improving the suggestion here a bit later.

lopopolo:
> As I've said in the issue, it's expected that the current suggestion does not compile (it even has a comment up top), and has the unspecified applicability. It's mostly meant as a nudge in the right direction to the user, not an autofix.
> So testing the suggestion doesn't really make sense since we know it won't compile in any case.

ahh gotcha. Thanks for clarifying @y21. I didn't read this PR thread or the code. I only left a comment here to cross-post my bug to make sure the rule author (you 😄) came across it.

FWIW, that lack of machine applicability is not made clear in the warning output. Based on my familiarity with clippy's other suggestions, I expected to be able to copy and paste:

```
warning: `iter` method without an `IntoIterator` impl for `&Links<T>`
  --> src/link.rs:68:5
   |
68 | /     pub fn iter(&self) -> Iter<'_, Link<T>, usize> {
69 | |         self.registry.iter()
70 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#iter_without_into_iter
  --> src/lib.rs:11:9
   |
11 | #![warn(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::iter_without_into_iter)]` implied by `#[warn(clippy::pedantic)]`
help: consider implementing `IntoIterator` for `&Links<T>`
   |
33 +
34 + impl IntoIterator for &Links<T> {
35 +     type IntoIter = hashbrown::hash_map::Iter<'_, link::Link<T>, usize>;
36 +     type Iter = (&link::Link<T>, &usize);
37 +     fn into_iter() -> Self::IntoIter {
38 +         self.iter()
39 +     }
40 + }
   |
```

PR #11528, opened 2023-09-17
New lint: `ambiguous_method_names`

### LINKED ISSUE #11499: Lint for ambiguous method defined in trait impl for struct & on struct itself
### What it does

This was inspired by [rust-lang #37741](https://github.com/rust-lang/rust/issues/37741). In short, it's very easy to accidentally override a trait method impl. by adding a method of the same name to a struct impl.. This causes unqualified call sites to land in the struct's method instead of the trait method impl for the struct. (Generally call sites will be unqualified because naming conflicts between traits are rare, so this will happen silently.) Since clippy generally warns for potentially incorrect code, this feels like a good candidate for a new lint.

The problematic situation is shown in [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=24851dd74560c3418e3f7e416a3c05b0). For simplicity, here's the code directly:

```rust
trait DataExt {
    fn ambiguous(&self);
}

struct BaseData {}

impl BaseData {
    // This method shadows DataExt::ambiguous(self).
    fn ambiguous(&self) {
        println!("struct impl");
    }
}

impl DataExt for BaseData {
    fn ambiguous(&self) {
        println!("trait impl");
    }
}

fn main() {
    let data = BaseData{};
    
    // This calls BaseData::ambiguous instead of DataExt::ambiguous & generates no warnings,
    // despite the trait being in scope.
    data.ambiguous();
}
```

The proposal would be to flag `BaseData::ambiguous` as creating ambiguity with `DataExt::ambiguous`. In other words, if a struct `S` implements a trait `T` with method `T::foo`, then a method `S::foo` should generate an ambiguity warning.

### Advantage

- warns about unintentional "overriding" of a trait method already defined on a struct.
- encourages writing easier to understand code

### Drawbacks

- complex object hierarchies may intentionally want this behavior (this is why we probably want the lint to be default allowed)

### Example

The lint should trigger as follows:

```rust
trait DataExt {
    fn ambiguous(&self);
}

struct BaseData {}

impl BaseData {
    // This method shadows DataExt::ambiguous(self) and should fail the lint.
    fn ambiguous(&self) {
        println!("struct impl");
    }
}

impl DataExt for BaseData {
    fn ambiguous(&self) {
        println!("trait impl");
    }
}
```

The developer could then rewrite as:

```rust
trait DataExt {
    fn ambiguous(&self);
}

struct BaseData {}

impl BaseData {
    fn no_longer_ambiguous(&self) {
        println!("struct impl");
    }
}

impl DataExt for BaseData {
    fn ambiguous(&self) {
        println!("trait impl");
    }
}
```

sjwang05:
Closes #11499 

This lint checks for methods in trait impls and struct impls with the same name. See lint description for more info and examples.

changelog: 
new lint: [`ambiguous_method_calls`]


clippy_lints/src/ambiguous_method_calls.rs
```diff
@@ -0,0 +1,179 @@
...
+fn has_ambiguous_name(ty: Ty<'_>, ident: Ident) -> bool {
+    let ty_str = format!("{ty}");
+    let trait_methods = trait_methods().lock().unwrap();
+    let struct_methods = struct_methods().lock().unwrap();
+
+    trait_methods.contains_key(&(ty_str.clone(), ident.name)) && struct_methods.contains_key(&(ty_str, ident.name))
+}
+
+fn trait_methods() -> &'static Mutex<FxHashMap<(String, Symbol), SpanData>> {
+    static NAMES: OnceLock<Mutex<FxHashMap<(String, Symbol), SpanData>>> = OnceLock::new();
```
y21, line 160: Is there a reason for using `static`s instead of storing the map directly in the `AmbiguousMethodCalls` struct? The lint structs can have fields and take `&mut self` in their check methods so its state can be accessed and mutated at any point, which means the `Mutex` could go too.

A problem with the current approach is that it depends on the order in which it visits the items. For example, if a `Base.ambiguous();` expression appears before an ambiguous method implementation, I think it wouldn't lint on it because clippy hasn't seen it yet.

I wonder if maintaining these name maps is necessary though. Rustc already has its own maps for trait/inherent impls per crate for its analysises and typechecking, and they can be accessed through the `TyCtxt`.

For example, there's the [`all_local_trait_impls`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.all_local_trait_impls) query that returns all trait impls for a given type's `DefId`. Maybe when visiting a method in an inherent impl, this could be used to look for traits that have methods with the same name.
I'm not sure if there's a better query to use for this... A problem that I can see here is that it probably wouldn't lint when calling ambiguous unqualified trait methods from foreign crates.

And for linting method call expressions, it could perhaps use `all_local_trait_impls` to get all trait impls + [`inherent_impls`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/ty/struct.TyCtxt.html#method.inherent_impls) for the normal impls, then check if there's duplicate entries between the two.

But maybe having these two maps is faster over having to go through the impls and compare a bunch of `Symbol`s that it justifies having them.

y21: > so linting happens only after visiting all implementations and exprs.

Yep that's a neat way to solve it, FWIW [some other lints follow the same idea](https://github.com/rust-lang/rust-clippy/blob/ddbe1100813d75898d6cde56a498813ac5799e9d/clippy_lints/src/unused_async.rs#L159). For example `unused_async` doesn't lint immediately once it sees an async function without `.await` exprs, but instead stores the relevant stuff for later + delays linting until `check_crate_post` so that we can be sure there aren't any path exprs to an async function in a context that requires a `fn() -> impl Future`.

If going with manual maps, there's one subtle but important thing: the iteration order of a HashMap is not stable, so the order in which it then emits lints can "randomly" change. That makes it annoying for e.g. ui tests where it checks for the exact stderr literally and a different order could then make the tests fail.
I ran into the exact same problem before ^^ see the relevant comment by Jarcho here https://github.com/rust-lang/rust-clippy/pull/11200#discussion_r1271338932

The `inherent_methods` field could just be a `Vec`, right? Seems like it's only inserted into and iterated over.


> As for using rustc's own maps, I'm currently trying to figure something out with the all_traits and all_local_trait_impls queries, and iterating through methods in trait impls to find those that are ambiguous with inherent methods (i.e., same name and same parent type) with something like this in check_fn

Oh huh I was under the impression that the key of `all_local_trait_impls` was the DefId of the type and not the trait, which would have been way more convenient (wouldn't require going through all of the traits). Definitely changes the way I think about this now. All of the trait impl queries on the tyctxt seem to search by a trait id (unless I missed something), which might make this more complicated then..

sjwang05: > The inherent_methods field could just be a `Vec`, right? Seems like it's only inserted into and iterated over.

Yeah, I think I was initially worried about getting the correct spans for each inherent impl (hence the `self.inherent_methods.get(k)`), but now that I think about it, this could also be accomplished using something like a `Vec<(String, Symbol, Span)>` instead of a `FxHashMap`--I believe this also solves the random ordering problem.

> I was under the impression that the key of `all_local_trait_impls`'s map was the DefId of the type and not the trait

That tripped me up as well when I was working on the rustc-based approach. I think this fact would also make linting call sites a lot more difficult, though I haven’t gotten around to writing that yet—do you think it’s still worth pursuing, or should we just stick with manual maps? I’m personally more in favor of manual maps—I feel like querying `all_local_trait_impls`, `all_traits`, and `inherent_impls` all at the same time when linting method calls could get complicated pretty quickly.

clippy_lints/src/ambiguous_method_calls.rs
```diff
@@ -0,0 +1,179 @@
...
+                        span_lint_and_help(
+                            cx,
+                            AMBIGUOUS_METHOD_CALLS,
+                            *span,
+                            "ambiguous method call",
+                            None,
+                            "consider renaming the struct impl's method or explicitly qualifying the call site",
+                        );
+                    }
+                }
```
Jarcho, line 175: This should be a single lint call. I would lint on the inherent method since that's the most likely one to rename. The trait can be trait method(s) can be called out in a note. The call sites are probably not worth mentioning due to the amount of unhelpful noise that it can generate.

Jarcho (changes requested):
Two inherent methods can share the same name assuming the impl blocks have different generic arguments. e.g.
```rust
struct S<T>(T);
impl S<i32> { fn f(self) {} }
impl S<u32> { fn f(self) {} }
```

If you aren't tracking the call sites, then I don't think you need to do anything to account for this, but I would have a test in case future changes need to do something special here.

sjwang05:
Thanks for all the feedback!

Regarding structs with generic arguments and method call sites, I decided to remove call site linting and all related code, so I don't think that'll be a concern for now. I've added a few test cases regardless. 

I've also split up the code in `check_fn` to `check_item` and `check_impl_item`, and I updated `check_crate_post` accordingly. However, `implements_trait` ICEs when a trait has generic arguments and the arguments supplied to the function differ from those of the trait. Do you have any suggestions on how I should proceed? Thanks!

sjwang05:
Disregard the part about the ICE, actually--I was able to find a fix :slightly_smiling_face: . I also added some tests to check for the ICE.

lengyijun:
FYI there is a similar lint `same_name_method`

sjwang05:
Huh, I wasn't aware. If y'all think it'd be best to close this PR as a duplicate, I'm perfectly fine doing so.

Jarcho:
@lengyijun Thanks for pointing that out.

@sjwang05 Sorry you got this far into the lint before this came up. That's a failure on our part for not triaging the issue properly.

PR #11530, opened 2023-09-18
Move zero_ptr to the casts module

Alexendoo:
Also a bit of a touch up to the code while I was there

changelog: none


blyxyas (approved):
LGTM, thanks! :heart:

PR #11539, opened 2023-09-20
Warn missing_enforced_import_renames by default

taiki-e:
Similar to https://github.com/rust-lang/rust-clippy/pull/8261 that did the same thing to disallowed_methods & disallowed_types.
This lint is also only triggered if import renames are defined in the `clippy.toml` file.

changelog: Moved [`missing_enforced_import_renames`] to `style` (Now warn-by-default)
[#11539](https://github.com/rust-lang/rust-clippy/pull/11539)


Centri3 (approved):

Centri3:
Thank you! @bors r+

Issue #11443, opened 2023-09-01
Lint to disallow renaming function parameters


Centri3:
### What it does

The name for parameters of a function are entirely arbitrary; however, in many cases the default for that trait is more desirable for consistency's sake.

Should not lint if the pattern is not a single binding or if its name starts with `_` in either the default implementation or the new name.

Should also be allow-by-default.

### Advantage

- Consistency, for example `from_str`'s parameter is `s` which some may wish to always use. Similarly, `From`'s `value`.

### Drawbacks

- It's really really pedantic.

### Example

```rust
struct A(u32);

impl From<A> for String {
    fn from(a: A) -> Self {
        a.0.to_string()
    }
}
```

Could be written as:

```rust
struct A(u32);

impl From<A> for String {
    fn from(value: A) -> Self {
        value.0.to_string()
    }
}
```

Centri3:
By very very pedantic, I didn't mean too pedantic for `pedantic`; I mean it may be too pedantic even for `restriction` ^^

We could allow library authors to opt into this through an attribute, and just annotate those traits with it.

J-ZhengLi:
@rustbot claim

I'll give it a try 😅

Issue #486, opened 2015-12-08
lint against renaming of method arguments in trait impls


oli-obk:
for consistency one should be using the same method arg names when implementing a trait as the ones used in the trait. 

possible cases:
- if the trait doesn't name them (`_`) an arbitrary name should be allowed,
- if the trait silences an unused warning in a default impl by prepending and underscore, allow the version without an underscore prefix
- allow prepending an underscore

Manishearth:
I'm not sure how I feel about this lint, but okay as an `Allow` lint. Unsure about `Warn`.

(At some point we'll need to make a Rust RfC to figure out sensible defaults that the community agrees on.)

fhartwig:
I would prefer this to be `Allow`. For most traits, this doesn't really matter since the methods only take `self` anyway, and for many of the others I don't think that consistency is important enough to warrant the cognitive overhead of remembering the exact argument names (quick, what's the name of the iterator argument to `Extend::extend`?)

Manishearth:
(Well, clippy can give suggestions, so you don't need to look it up manually, but I agree in general)

PR #11540, opened 2023-09-20
add new lint that disallow renaming parameters in trait functions

J-ZhengLi:
fixes: #11443 
fixes: #486 

changelog: add new lint [`renamed_function_params`]

Note that the lint name is not final, because I have a bad reputation in naming things, and I don't trust myself.

### DIFF
--- a/clippy_config/src/conf.rs
+++ b/clippy_config/src/conf.rs
@@ -40,6 +40,8 @@ const DEFAULT_DOC_VALID_IDENTS: &[&str] = &[
 const DEFAULT_DISALLOWED_NAMES: &[&str] = &["foo", "baz", "quux"];
 const DEFAULT_ALLOWED_IDENTS_BELOW_MIN_CHARS: &[&str] = &["i", "j", "x", "y", "z", "w", "n"];
 const DEFAULT_ALLOWED_PREFIXES: &[&str] = &["to", "as", "into", "from", "try_into", "try_from"];
+const DEFAULT_ALLOWED_TRAITS_WITH_RENAMED_PARAMS: &[&str] =
+    &["core::convert::From", "core::convert::TryFrom", "core::str::FromStr"];
 
 /// Conf with parse errors
 #[derive(Default)]
@@ -613,6 +615,23 @@ define_Conf! {
     /// - Use `".."` as part of the list to indicate that the configured values should be appended to the
     /// default configuration of Clippy. By default, any configuration will replace the default value
     (allowed_prefixes: Vec<String> = DEFAULT_ALLOWED_PREFIXES.iter().map(ToString::to_string).collect()),
+    /// Lint: RENAMED_FUNCTION_PARAMS.
+    ///
+    /// List of trait paths to ignore when checking renamed function parameters.
+    ///
+    /// #### Example
+    ///
+    /// ```toml
+    /// allow-renamed-params-for = [ "std::convert::From" ]
+    /// ```
+    ///
+    /// #### Noteworthy
+    ///
+    /// - By default, the following traits are ignored: `From`, `TryFrom`, `FromStr`
+    /// - `".."` can be used as part of the list to indicate that the configured values should be appended to the
+    /// default configuration of Clippy. By default, any configuration will replace the default value.
+    (allow_renamed_params_for: Vec<String> =
+        DEFAULT_ALLOWED_TRAITS_WITH_RENAMED_PARAMS.iter().map(ToString::to_string).collect()),
 }
 
 /// Search for the configuration file.
@@ -674,6 +693,10 @@ fn deserialize(file: &SourceFile) -> TryConf {
             extend_vec_if_indicator_present(&mut conf.conf.doc_valid_idents, DEFAULT_DOC_VALID_IDENTS);
             extend_vec_if_indicator_present(&mut conf.conf.disallowed_names, DEFAULT_DISALLOWED_NAMES);
             extend_vec_if_indicator_present(&mut conf.conf.allowed_prefixes, DEFAULT_ALLOWED_PREFIXES);
+            extend_vec_if_indicator_present(
+                &mut conf.conf.allow_renamed_params_for,
+                DEFAULT_ALLOWED_TRAITS_WITH_RENAMED_PARAMS,
+            );
             // TODO: THIS SHOULD BE TESTED, this comment will be gone soon
             if conf.conf.allowed_idents_below_min_chars.contains("..") {
                 conf.conf
--- a/clippy_lints/src/functions/renamed_function_params.rs
+++ b/clippy_lints/src/functions/renamed_function_params.rs
@@ -1,18 +1,26 @@
 use clippy_utils::diagnostics::span_lint_and_then;
 use rustc_errors::{Applicability, MultiSpan};
-use rustc_hir::def_id::DefId;
+use rustc_hir::def_id::{DefId, DefIdSet};
 use rustc_hir::hir_id::OwnerId;
-use rustc_hir::{ImplItem, ImplItemKind, ItemKind, Node};
+use rustc_hir::{Impl, ImplItem, ImplItemKind, ImplItemRef, ItemKind, Node, TraitRef};
 use rustc_lint::LateContext;
 use rustc_span::symbol::{kw, Ident, Symbol};
 use rustc_span::Span;
 
 use super::RENAMED_FUNCTION_PARAMS;
 
-pub(super) fn check_impl_item(cx: &LateContext<'_>, item: &ImplItem<'_>) {
+pub(super) fn check_impl_item(cx: &LateContext<'_>, item: &ImplItem<'_>, ignored_traits: &DefIdSet) {
     if !item.span.from_expansion()
         && let ImplItemKind::Fn(_, body_id) = item.kind
-        && let Some(did) = trait_item_def_id_of_impl(cx, item.owner_id)
+        && let parent_node = cx.tcx.parent_hir_node(item.hir_id())
+        && let Node::Item(parent_item) = parent_node
+        && let ItemKind::Impl(Impl {
+            items,
+            of_trait: Some(trait_ref),
+            ..
+        }) = &parent_item.kind
+        && let Some(did) = trait_item_def_id_of_impl(items, item.owner_id)
+        && !is_from_ignored_trait(trait_ref, ignored_traits)
     {
         let mut param_idents_iter = cx.tcx.hir().body_param_names(body_id);
         let mut default_param_idents_iter = cx.tcx.fn_arg_names(did).iter().copied();
@@ -25,7 +33,7 @@ pub(super) fn check_impl_item(cx: &LateContext<'_>, item: &ImplItem<'_>) {
                 cx,
                 RENAMED_FUNCTION_PARAMS,
                 multi_span,
-                &format!("renamed function parameter{plural} of trait impl"),
+                format!("renamed function parameter{plural} of trait impl"),
                 |diag| {
                     diag.multipart_suggestion(
                         format!("consider using the default name{plural}"),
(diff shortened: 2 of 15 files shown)


J-ZhengLi:
> I haven't looked at the implementation, so this might already be implemented: I would recommend that it allows removing leading underscores from the names. Here is an example, of what I mean:
> 
> ```rust
> trait Foo {
>     fn foo(_param: u32) {}
> }
> 
> struct FooImpl;
> 
> impl Foo for FooImpl {
>     //     vvvvv This should be allowed
>     fn foo(param: u32) {
>         println!("{param}");
>     }
> }
> ```
> 
> Potentially, it could also allow adding an underscore, but here I'm uncertain if that's a good idea. Still, this would be the example for that case:
> 
> ```rust
> trait Bar {
>     fn bar(param: u32) {
>         println!("{param}");
>     }
> }
> 
> struct BarImpl;
> 
> impl Bar for BarImpl {
>     //     vvvvvv Maybe this could be allowed?
>     fn bar(_param: u32) {}
> }
> ```

It should already be implemented :D 
in fact any name starts with underscore will be skipped, that includes names in trait def and trait impl

J-ZhengLi:
oops, didn't know the doc comments are being checked now, will fix tomorrow

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,82 @@
...
+                default_name,
+            });
+        }
+    }
+    renamed
+}
+
+fn is_ignored_or_empty_symbol(symbol: Symbol) -> bool {
+    let s = symbol.as_str();
+    s.is_empty() || s.starts_with('_')
```
Centri3, line 64: An identifier can't be empty, so this always returns false

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,82 @@
...
+    default_name: Symbol,
+}
+
+fn renamed_params<I, T>(default_names: &mut I, current_names: &mut T) -> Vec<RenamedParam>
+where
+    I: Iterator<Item = Ident>,
+    T: Iterator<Item = Ident>,
+{
+    let mut renamed = vec![];
+    // FIXME: Should we stop if they have different length?
```
Centri3, line 45: I'd say this should be a `debug_assert!`. This is impossible with trait methods so this would only happen if this is incorrectly called on smth else.

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+}
+
+struct RenamedParam {
+    renamed_span: Span,
+    default_name: Symbol,
+}
+
+fn renamed_params<I, T>(default_names: &mut I, current_names: &mut T) -> Vec<RenamedParam>
```
Centri3, line 39: Can we use impl trait here?

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,82 @@
...
+use super::RENAMED_FUNCTION_PARAMS;
+
+pub(super) fn check_impl_item(cx: &LateContext<'_>, item: &ImplItem<'_>) {
+    if let ImplItemKind::Fn(_, body_id) = item.kind &&
+        let Some(did) = impled_item_def_id(cx, item.owner_id)
+    {
+        let mut param_idents_iter = cx.tcx.hir().body_param_names(body_id);
+        let mut default_param_idents_iter = cx.tcx.fn_arg_names(did).iter().copied();
+
+        let renames = renamed_params(&mut default_param_idents_iter, &mut param_idents_iter);
```
Centri3, line 18: nit:
```suggestion
        let renamed = renamed_fn_args(&mut default_param_idents_iter, &mut param_idents_iter);
```
`fn_args` is more consistent with many queries, and `renames` feels off.

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,82 @@
...
+        // FIXME: Should we use `MultiSpan` to combine output together?
+        // But how should we display help message if so.
```
Centri3, line 20: It should be a suggestion imo, similar to `needless_lifetimes`. ofc with `Applicability::Unspecified`.

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+    renamed
+}
+
+fn is_ignored_or_empty_symbol(symbol: Symbol) -> bool {
+    let s = symbol.as_str();
+    s.is_empty() || s.starts_with('_')
+}
+
+fn impled_item_def_id(cx: &LateContext<'_>, impl_item_id: OwnerId) -> Option<DefId> {
```
Centri3, line 67:
```suggestion
fn impl_item_def_id(cx: &LateContext<'_>, impl_item_id: OwnerId) -> Option<DefId> {
```
I thought this was `implied_item` at first ^^

Centri3 (commented):
Just some minor nits.

I still think we should restrict this to some traits (by default) using an attribute like `clippy::cant_rename` (bikeshedding pending!). We can add this to many libstd traits and crate authors can opt into this if they really want. Consistency isn't as big of a deal when the trait is rarely used already.

clippy_lints/src/functions/renamed_function_params.rs
J-ZhengLi: Well, I think the str form of a wildcard is empty, because this happened before I change the lint level:

```
error: function parameter name was renamed from its trait default
  --> src\unicode.rs:78:30
   |
78 |     fn check_expr(&mut self, cx: &LateContext<'_>, expr: &'_ Expr<'_>) {
   |                              ^^
   |
   = help: consider changing the name to: ''
```

clippy_lints/src/functions/renamed_function_params.rs
J-ZhengLi: I messed up the naming again 😭 , it should be trait_item_def_id consider it's what this function trying to get... I'll change it

clippy_lints/src/functions/renamed_function_params.rs
Centri3: Maybe `fn_arg_names` does that on a `Wildcard` pattern :thinking: That's pretty peculiar

J-ZhengLi: @Centri3 huh, seems like both `body_param_names` and `fn_arg_names` return an empty ident with dummy span for underscore ('_').

[the line in `body_param_names` function](https://doc.rust-lang.org/beta/nightly-rustc/src/rustc_middle/hir/map/mod.rs.html#436)

J-ZhengLi: I've added one as "fixme" on my last commit, should do the trick, I think...

Jarcho:
Since you've already looked at this. r? @Centri3

J-ZhengLi:
Perhaps this was forgotten 😭 , even I couldn't remember what I did

@rustbot r? @rust-lang/clippy

Manishearth:
r? clippy

still busy

xFrednet:
Sure, it's on my todo list for the next week :D

xFrednet (approved):
Thank you for the implementation, this looks good to me. Time to get the FCP started.

@rustbot label +S-final-comment-period -S-waiting-on-author

FCP: https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/FCP.20.60renamed_function_params.60/near/431407447

xFrednet:
On zulip it was mentioned that this lint should ideally have a configuration to ignore some lints, like `From`,  `TryFrom`, and `FromStr`. Once that is done I believe we can move forward with this. There was also this suggestion:

> It looks a bit similar to `disallowed_names`. Perhaps the name could be `disallow_parameter_renames` or `disallow_renames`?

I like the current name, but can understand the reasoning :)

J-ZhengLi:
@xFrednet Finally got some time working on this~ I've added a configuration for it, but again... not confident about that config's name, which is `ignored-traits` for now, if there's better name plz let me know~

some time I really wish there's an AI that could tell me what should I name my variables/functions/lint/config lol

clippy_lints/src/functions/renamed_function_params.rs
```diff
@@ -0,0 +1,123 @@
...
+use rustc_hir::hir_id::OwnerId;
+use rustc_hir::{Impl, ImplItem, ImplItemKind, ImplItemRef, ItemKind, Node, TraitRef};
+use rustc_lint::LateContext;
+use rustc_span::symbol::{kw, Ident, Symbol};
+use rustc_span::Span;
+use std::sync::OnceLock;
+
+use super::RENAMED_FUNCTION_PARAMS;
+
+static IGNORED_TRAIT_IDS: OnceLock<DefIdSet> = OnceLock::new();
```
J-ZhengLi, line 14: I'm not sure if I should use `OnceLock` here, haven't seen them in other lints, maybe there's some kind of concern? idk

clippy_config/src/conf.rs
```diff
@@ -610,6 +611,22 @@ define_Conf! {
...
+    /// ```toml
+    /// ignored-traits = [ "std::convert::From" ]
+    /// ```
+    ///
+    /// #### Noteworthy
+    ///
+    /// - By default, the following traits are ignored: `From`, `TryFrom`, `FromStr`
+    /// - `".."` can be used as part of the list to indicate that the configured values should be appended to the
+    /// default configuration of Clippy. By default, any configuration will replace the default value.
+    (ignored_traits: Vec<String> = DEFAULT_IGNORED_TRAITS.iter().map(ToString::to_string).collect()),
```
xFrednet, line 629: This name is too general, maybe `allow_renamed_params_for` or `renamed_function_params_alllowed_for`?

clippy_config/src/conf.rs
```diff
@@ -671,6 +688,7 @@ fn deserialize(file: &SourceFile) -> TryConf {
             extend_vec_if_indicator_present(&mut conf.conf.doc_valid_idents, DEFAULT_DOC_VALID_IDENTS);
             extend_vec_if_indicator_present(&mut conf.conf.disallowed_names, DEFAULT_DISALLOWED_NAMES);
             extend_vec_if_indicator_present(&mut conf.conf.allowed_prefixes, DEFAULT_ALLOWED_PREFIXES);
+            extend_vec_if_indicator_present(&mut conf.conf.ignored_traits, DEFAULT_IGNORED_TRAITS);
```
xFrednet, line 691: Nice! :D

clippy_lints/src/functions/renamed_function_params.rs
xFrednet: I believe some other lints store this in the lint pass struct instead. The IDs can then be collected in `fn check_crate()`. I would prefer that solution over having a static object. Otherwise, I don't believe there is a reason against `OnceLock`

xFrednet (commented):
Two NITs that should hopefully be easy to fix and then this should be good to go :D

(Sorry for the delay with the review, my past few weeks have been busy)

clippy_lints/src/functions/renamed_function_params.rs
J-ZhengLi: Oh I see, I just thought it might be redundant to having a list of trait paths, and a set of their ids stored in the same struct

J-ZhengLi:
> (Sorry for the delay with the review, my past few weeks have been busy)

no worries, me too XD

xFrednet (approved):

xFrednet:
In Clippy's realm, a rule takes flight,
No renaming parameters, to keep code tight.
Thanks for the effort, I merge with delight.

=^.^=

J-ZhengLi:
ahh, `cargo collect-metadata` again, third times a charm I guess~

xFrednet:
Feel free to `r=me` after the collection :)

@bors delegate+

J-ZhengLi:
should be good now

@bors r=@xFrednet

PR #11543, opened 2023-09-20
Fix needless_continue

lengyijun:
changelog: [`needless_continue`]: don't rm the last `continue` in loop if the continue label is not the loop label


lengyijun:
if the continue target is other label, it should be captured by clippy::never_loop

PR #11544, opened 2023-09-20
Remove most usage of `hir_ty_to_ty`

Alexendoo:
Removes the usages where there's a suitable query or the type was already available elsewhere. The remaining cases would all require more involved changes

changelog: none

r? @Jarcho


Jarcho:
It would be nice to be able to remove it completely given the number of crashes that come from it.

@bors r+

Issue #4077, opened 2019-05-10
Needless continue for match block


pickfire:
tl;dr Fix simple needless continue

```rust
for _ in 0..1 {
    // do something
    continue;
}
```

Currently, `needless_continue` only checks possible removal of `continue` by reordering `if-else` block. One more simpler needless continue to be removed:

From https://github.com/Gymmasssorla/finshir/blob/eae5d0c6761f5558c8ed33ef098d8bd13a07e64f/src/testing/mod.rs#L124-L136

```rust
        for _ in 0..(failed_count.get() - 1) {
            match socket.write_all(portion) {
                Ok(_) => return SendPortionResult::Success,
                Err(err) => {
                    error!(
                        "Failed to send {} byte(s) >>> {}! Retrying the operation...",
                        helpers::cyan(portion.len()),
                        err
                    );
                    continue;
                }
            }
        }
```

Which `continue` can be dropped.

```rust
        for _ in 0..(failed_count.get() - 1) {
            match socket.write_all(portion) {
                Ok(_) => return SendPortionResult::Success,
                Err(err) => {
                    error!(
                        "Failed to send {} byte(s) >>> {}! Retrying the operation...",
                        helpers::cyan(portion.len()),
                        err
                    );
                }
            }
        }
```

One more in https://github.com/Gymmasssorla/finshir/blob/eae5d0c6761f5558c8ed33ef098d8bd13a07e64f/src/testing/mod.rs#L150-L164

What needs to be achieve here is to check if there are any `continue` at the end of each branches (including `if-else` and `match` blocks).

Possible code changes to clippy can probably be done in https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/needless_continue.rs#L374

Xaeroxe:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e1f764cbed2b7227df7a748aa6a2b2dd

Similarly, any continue, conditional or not, which is the last statement of the loop it is in, is useless and should be omitted.

PR #11546, opened 2023-09-21
[`needless_continue`]: lint if the last stmt in for/while/loop is `co…

lengyijun:
changelog: [`needless_continue`]: lint if the last stmt in for/while/loop is `continue`, recursively
Fixes: #4077

### DIFF
--- a/clippy_lints/src/needless_continue.rs
+++ b/clippy_lints/src/needless_continue.rs
@@ -35,7 +35,7 @@
 //! This lint is **warn** by default.
 use clippy_utils::diagnostics::span_lint_and_help;
 use clippy_utils::source::{indent_of, snippet, snippet_block};
-use rustc_ast::ast;
+use rustc_ast::{ast, Block, Label};
 use rustc_lint::{EarlyContext, EarlyLintPass, LintContext};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
 use rustc_span::Span;
@@ -46,6 +46,7 @@ declare_clippy_lint! {
     /// that contain a `continue` statement in either their main blocks or their
     /// `else`-blocks, when omitting the `else`-block possibly with some
     /// rearrangement of code can make the code easier to understand.
+    /// The lint also checks if the last statement in the loop is a `continue`
     ///
     /// ### Why is this bad?
     /// Having explicit `else` blocks for `if` statements
@@ -108,6 +109,45 @@ declare_clippy_lint! {
     ///     }
     ///     // Do something useful
     ///     # break;
+    /// }
+    /// ```
+    ///
+    /// ```rust
+    /// fn foo() -> std::io::ErrorKind { std::io::ErrorKind::NotFound }
+    /// for _ in 0..10 {
+    ///     match foo() {
+    ///         std::io::ErrorKind::NotFound => {
+    ///             eprintln!("not found");
+    ///             continue
+    ///         }
+    ///         std::io::ErrorKind::TimedOut => {
+    ///             eprintln!("timeout");
+    ///             continue
+    ///         }
+    ///         _ => {
+    ///             eprintln!("other error");
+    ///             continue
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    /// Could be rewritten as
+    ///
+    ///
+    /// ```rust
+    /// fn foo() -> std::io::ErrorKind { std::io::ErrorKind::NotFound }
+    /// for _ in 0..10 {
+    ///     match foo() {
+    ///         std::io::ErrorKind::NotFound => {
+    ///             eprintln!("not found");
+    ///         }
+    ///         std::io::ErrorKind::TimedOut => {
+    ///             eprintln!("timeout");
+    ///         }
+    ///         _ => {
+    ///             eprintln!("other error");
+    ///         }
+    ///     }
     /// }
     /// ```
     #[clippy::version = "pre 1.29.0"]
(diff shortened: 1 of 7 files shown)


Alexendoo:
This would be another good one to look at @y21

tests/ui/needless_continue.rs
```diff
@@ -151,3 +151,25 @@ mod issue_2329 {
...
+
+mod issue_4077 {
+    fn main() {
+        loop {
+            do_something();
+            if some_expr() {
+                continue;
+            }
+        }
+    }
```
y21, line 163: A test that makes use of labels would be nice. Smth like
```rs
fn f() {
  'outer: loop {
    'inner: loop {
      if some_expr() {
        continue 'inner; // lint
      } else {
        continue 'outer; // don't lint
      }
    }
  }
}
```

clippy_lints/src/needless_continue.rs
```diff
@@ -361,24 +361,68 @@ fn suggestion_snippet_for_continue_inside_else(cx: &EarlyContext<'_>, data: &Lin
...
+                        rustc_ast::ExprKind::Block(b, _) => {
+                            check_last_stmt(b, func);
+
+                        }
+                        _ => {}
+                    }
+
+                }
+            }
+            _ => {},
```
y21, line 405: Probably makes sense to check for `Block` here too? Would allow us to lint
```rs
loop {
  { continue; }
}
```
as well

y21 (commented):
Just a few things, other than that it looks good to me!

y21 (commented):
I believe the lint description could also use an update now that it lints additional cases and no longer just `if` exprs.

xFrednet:
Hey @lengyijun, the CI failed on your last commit. Do you need any help to address the issue? :)

y21 (approved):
Looks good to me!

clippy_lints/src/needless_continue.rs
```diff
@@ -361,24 +401,79 @@ fn suggestion_snippet_for_continue_inside_else(cx: &EarlyContext<'_>, data: &Lin
...
+
+                        }
+                        _ => {}
+                    }
+
+                }
+            }
+            ast::ExprKind::Block(b, _) => {
+                check_last_stmt(b, func);
+            }
```
Alexendoo, line 455: It would be good to flatten this since it's currently handling the same expressions in a few different places, you could split it into two functions `check_last_stmt` and `check_last_stmt_in_expr` (replacing `check_last_stmt_in_if`)

e.g. roughly

```rust
fn check_last_stmt_in_expr(..) {
    match &expr.kind {
        ExprKind::Continue(continue_label) => {
            // lint, could inline it since now it's only happening in one place
        },
        ExprKind::If(_, then_block, else_block) => {
            check_last_stmt(..);
            if let Some(else_block) = else_block {
                check_last_stmt_in_expr(..);
            }
        },
        ExprKind::Match(_, arms) => {
            for arm in arms {
                check_last_stmt_in_expr(..);
            }
        },
        ExprKind::Block(b, _) => {
            check_last_stmt(..);
        },
        _ => {},
    }
}

fn check_last_stmt(..) {
    if let Some(last_stmt) = b.stmts.last()
        && let ast::StmtKind::Expr(expr) | ast::StmtKind::Semi(expr) = &last_stmt.kind
    {
        check_last_stmt_in_expr(expr, ..);
    }
}
```

Alexendoo:
@bors delegate=y21

I think a good follow up to this if you're interested would be to split this lint up, this addition should be warn by default IMO but the lint is currently `pedantic`. The suggestions that require rearranging code instead of just removing a `continue`/`else { continue }` could go in their own lint

clippy_lints/src/needless_continue.rs
```diff
@@ -361,24 +401,58 @@ fn suggestion_snippet_for_continue_inside_else(cx: &EarlyContext<'_>, data: &Lin
...
+                    span,
+                    MSG_REDUNDANT_CONTINUE_EXPRESSION,
+                    None,
+                    DROP_CONTINUE_EXPRESSION_MSG,
+                );
+            }
+        };
+        check_last_stmt_in_block(loop_block, &p);
+
         for (i, stmt) in loop_block.stmts.iter().enumerate() {
```
y21, line 456: Right now it's possible for the same lint to emit two warnings for the following code:
```rs
loop {
  if true {

  } else { // redundant `else`
    continue; // redundant `continue`
  }
}
```

<details>
<summary>Output</summary>

```
warning: this `continue` expression is redundant
   --> test.rs:171:13
    |
171 |             continue;
    |             ^^^^^^^^

warning: this `else` block is redundant
   --> test.rs:170:16
    |
170 |           } else {
    |  ________________^
171 | |             continue;
172 | |         }
    | |_________^
    |
```
</details>

I like the other warning a bit more in this case because it suggests removing the whole `else` block altogether.

@Alexendoo What do you think? Is this fine to ignore for now or would you consider it a potential blocker? Looks like it possibly has a bit of overlap with your suggestion to split up the lint

If this is a good idea and worth fixing in this PR, we could probably just move this `check_last_stmt_in_expr` call into the loop and guard it behind an `i == stmts.len() - 1` check so that it only runs on the last statement, inline the `with_if_expr` function (it's really short and only used here, anyway) and add `continue;` in the one lint case that is emitted here so that it skips checking the new case.

y21 (commented):
I looked through this one last time and found a potential edge case I wanted to mention (sorry! 😅)

clippy_lints/src/needless_continue.rs
Alexendoo: Yeah, good catch, we don't want to be linting the same thing twice

The else winning would be preferable, but if it's easier implementation wise to ignore it that's fine too, there's a lint for empty else blocks that would pick it up

xFrednet:
Hey @lengyijun, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

@rustbot author

xFrednet:

profetia:
Hi forks! I would like to pick up this implementation, looks like there aren't much left.

PR #11550, opened 2023-09-21
`impl_trait_in_params` now supports impls and traits

### LINKED ISSUE #11548: impl_trait_in_params should warn associated functions
### Summary

This lint currently seems to warn only for freestanding functions.

It is correct to not warn associated functions of trait impls (since the user has no control over it), but I think it would be reasonable to warn associated functions of normal impls.

It is also probably reasonable to warn associated functions of trait definitions if the `avoid-breaking-exported-api` config is set to `false`.

Mentioning @blyxyas, who implemented this lint in #10197.

### Lint Name

impl_trait_in_params

### Reproducer

I tried this code:

```rust
#![warn(clippy::impl_trait_in_params)]

pub fn a(_s: impl AsRef<str>) {}

pub struct B;

impl B {
    pub fn b(_s: impl AsRef<str>) {}
}
```

I expected to see this happen: warnings for both `a` and `b`.

Instead, this happened:

clippy emits warning only for `a`.

```
warning: '`impl Trait` used as a function parameter'
 --> src/lib.rs:3:14
  |
3 | pub fn a(_s: impl AsRef<str>) {}
  |              ^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#impl_trait_in_params
 --> src/lib.rs:1:9
  |
1 | #![warn(clippy::impl_trait_in_params)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: add a type parameter
  |
3 | pub fn a<{ /* Generic name */ }: AsRef<str>>(_s: impl AsRef<str>) {}
  |         ++++++++++++++++++++++++++++++++++++
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=239af06037ff9c308e4355c7dc4e2a96)

### Version

```text
rustc 1.74.0-nightly (3223b0b5e 2023-09-20)
binary: rustc
commit-hash: 3223b0b5e8dadda3f76c3fd1a8d6c5addc09599e
commit-date: 2023-09-20
host: aarch64-apple-darwin
release: 1.74.0-nightly
LLVM version: 17.0.0
```

blyxyas:
Before this PR, the lint `impl_trait_in_params`. This PR gives the lint support for functions in impls and traits. (Also, some pretty heavy refactor)

fixes #11548
changelog:[`impl_trait_in_params`] now supports `impl` blocks and functions in traits


taiki-e:
Thanks! 

I recommend adding a test for trait implementations -- I believe this lint should not warn for trait implementations because traits may be defined in the upstream crate, and the user who implements them may be unable to change them.

tests/ui/impl_trait_in_params.rs
```diff
@@ -32,4 +35,11 @@ impl S {
     pub fn k<K: AnotherTrait<u32>>(_: K, _: impl AnotherTrait<u32>) {} //~ ERROR: `impl Trait` used as a function parameter
 }
 
+// Trying with imported traits
+impl ToString for S {
+    fn to_string(&self) -> String {
+        String::from("a")
+    }
+}
```
taiki-e, line 43: This lint warns for impl trait in parameters, so I think we also need to check for a case where parameters include impl trait.

blyxyas: Oh what a brain melt. I really didn't notice that 🐱

blyxyas:
I just noticed the failing CI, I'll get to work on it right now :cat2:

dswij (approved):
LGTM, thanks for this!

dswij:
@blyxyas can you squash some commits? And then you can r=me

tests/ui-toml/impl_trait_in_params/impl_trait_in_params.stderr
```diff
@@ -0,0 +1,41 @@
+error: trait `Trait` is more private than the item `Public::t`
```
taiki-e, line 1: This error seems unrelated to what we want to test here. I guess you have to make `Trait` trait public.

blyxyas:
Wow I really fucked up the Git history for a couple minutes (like, adding 21 commits, 2K lines changed that aren't mine).
But don't worry, the good ol' reliable "entering panic mode" saved the day again

PR #11551, opened 2023-09-22
fixed fp caused by moving &mut reference inside of a closure

### LINKED ISSUE #11545: False Positive: needless_pass_by_ref_mut when mut-requiring use is via closure in async function
### Summary

When an async function takes a `&mut` parameter, so it can call another function which requires an `&mut` parameter, via a closure, clippy falsely reports the parameter doesn't need to be `&mut`.

This only seems to happen both if the function is `async`, and the call is via a closure.

### Lint Name

needless_pass_by_ref_mut

### Reproducer

I tried this code:

```rust
#![allow(clippy::redundant_closure_call)]
#![deny(clippy::needless_pass_by_ref_mut)]

fn pop_last(str: &mut String) -> bool {
    str.pop().is_some()
}

pub async fn pop_last_wrapper(str: &mut String) -> bool {
    (|| {
        pop_last(str)
    })()
}
```

I saw this happen:

```
error: this argument is a mutable reference, but not used mutably
 --> src/lib.rs:8:36
  |
8 | pub async fn pop_last_wrapper(str: &mut String) -> bool {
  |                                    ^^^^^^^^^^^ help: consider changing to: `&String`
  |
  = warning: changing this function will impact semver compatibility
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
 --> src/lib.rs:2:9
  |
2 | #![deny(clippy::needless_pass_by_ref_mut)]
  |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `rustplay` (lib) due to previous error
```

Code should not raise a lint - the function in use requires a `&mut String`, so this should be allowed.

### Version

```text
rustc 1.73.0-beta.6 (9b9539791 2023-09-17)
binary: rustc
commit-hash: 9b95397915e672e76c43eb013289bca7b9a23370
commit-date: 2023-09-17
host: aarch64-apple-darwin
release: 1.73.0-beta.6
LLVM version: 17.0.0
```

### Additional Labels

_No response_

Meczka:
changelog: [`needless_pass_by_ref mut`]: fixes false positive caused by not covering mutable references passed to a closure inside of a fuction
fixes #11545

### DIFF
--- a/clippy_lints/src/needless_pass_by_ref_mut.rs
+++ b/clippy_lints/src/needless_pass_by_ref_mut.rs
@@ -336,7 +336,12 @@ impl<'tcx> euv::Delegate<'tcx> for MutablyUsedVariablesCtxt<'tcx> {
     fn borrow(&mut self, cmt: &euv::PlaceWithHirId<'tcx>, _id: HirId, borrow: ty::BorrowKind) {
         self.prev_bind = None;
         if let euv::Place {
-            base: euv::PlaceBase::Local(vid),
+            base:
+                euv::PlaceBase::Local(vid)
+                | euv::PlaceBase::Upvar(UpvarId {
+                    var_path: UpvarPath { hir_id: vid },
+                    ..
+                }),
             base_ty,
             ..
         } = &cmt.place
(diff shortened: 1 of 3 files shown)


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the nice fix and adding tests :)

@bors r+

PR #11553, opened 2023-09-22
Fix large_futures example

mickvangelderen:
The value used in the large_futures example was not large enough to trigger the lint given the default threshold. The example also contained more code than necessary. This PR changes the value size from 1kB to 16kB and reduces the example in size.

changelog: [`large_futures`]: Fix and simplify example


xFrednet (approved):

xFrednet:
Looks good to me, thank you for the update :)

@bors r+

PR #11560, opened 2023-09-24
Add missing tests for configuration options

y21:
I noticed that a lot of lints didn't have test(s) for their configuration. This leads to issues like #11481 where the lint just does nothing with it.

This PR adds tests for *almost*[^1] all of the lints with a configuration that didn't have a test in ui-toml.
The tests that I wrote here are usually two cases: one for where it's right above or under the limit set by the config where it shouldn't lint and another one for right above where it should.

changelog: none

[^1]: allow-one-hash-in-raw-strings is ignored by needless_raw_string_hashes

### DIFF
--- a/tests/ui-toml/decimal_literal_representation/clippy.toml
+++ b/tests/ui-toml/decimal_literal_representation/clippy.toml
@@ -1,0 +1,1 @@
+literal-representation-threshold = 0xFFFFFF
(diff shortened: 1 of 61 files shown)


Alexendoo (approved):

Alexendoo:
Nice, thanks!

@bors r+

y21:
Right, usize can be 32 bits, looks like that's why CI is failing for i686 😅 . Will change the test to use some smaller number

Alexendoo:
🤔

@bors r+

PR #11564, opened 2023-09-25
Test that each config value exists in a test clippy.toml

Alexendoo:
Inspired by #11560, adds a test that each config option exists in some form in a `clippy.toml` in `tests/` (currently some are in `ui-toml`, some in `ui-cargo`)

changelog: none


giraffate (approved):

giraffate:
@bors r+

Nice!

Issue #10794, opened 2023-05-17
New lint suggestion: You're using a blocking operation when you could be using an async one


omarandlorraine:
### What it does

A lint which checks for blocking functions inside an async block.

One I know of (and the cause of a bug I ran into today) is `std::thread::sleep`, but there are others. (file operations, networking, ...). If such a function has an async counterpart, then the user could be warned of this fact.

### Lint Name

unwise-block

### Category

suspicious

### Advantage

- warn of badly behaving async programs

### Drawbacks

- might be too noisy

### Example

```rust
use std::thread::sleep;
use std::time::Duration;
use tokio::signal::unix::signal;
use tokio::signal::unix::SignalKind;
use tokio::sync::mpsc;

async fn rx(rx: &mut mpsc::Receiver<()>) {
    loop {
        sleep(Duration::new(1, 0));
        if rx.try_recv().is_ok() {
            println!("got one splivet");
        }
    }
}

async fn tx(tx: mpsc::Sender<()>) {
    let mut stream = signal(SignalKind::user_defined1()).unwrap();

    loop {
        stream.recv().await;
        println!("sending one splivet");
        tx.send(()).await.unwrap();
    }
}

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let (stx, mut srx) = mpsc::channel::<()>(5);

    tokio::join!(tx(stx), rx(&mut srx));
}
```

Could be written as:

```rust
use std::thread::sleep;
use std::time::Duration;
use tokio::signal::unix::signal;
use tokio::signal::unix::SignalKind;
use tokio::sync::mpsc;

async fn rx(rx: &mut mpsc::Receiver<()>) {
    loop {
        tokio::time::sleep(Duration::new(1, 0)).await;
        if rx.try_recv().is_ok() {
            println!("got one splivet");
        }
    }
}

async fn tx(tx: mpsc::Sender<()>) {
    let mut stream = signal(SignalKind::user_defined1()).unwrap();

    loop {
        stream.recv().await;
        println!("sending one splivet");
        tx.send(()).await.unwrap();
    }
}

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let (stx, mut srx) = mpsc::channel::<()>(5);

    tokio::join!(tx(stx), rx(&mut srx));
}
```

Note the different way to sleep in the `rx` function.

y21:
[...]

Issue #4377, opened 2019-08-13
Lint idea: find known-blocking constructs in async contexts


skade:
With `async/await` coming up, we will probably see an uptake of such code:

```
async hello() {
    println!("hello").
}
```

Users will most likely not be aware that the call to `println` are _blocking_. The same goes to any use of the stdlib io apis, most notably `ToSocketAddr` calls.

I think there is a chance for clippy to at least find obvious cases of this, for a list of known functions.

hwchen:
`thread::sleep` seems to be a pretty common mistake:

https://old.reddit.com/r/rust/comments/e1gxf8/not_understanding_asyncawait_properly/
https://old.reddit.com/r/rust/comments/dtp6z7/what_can_i_actually_do_with_the_new_async_fn/
https://old.reddit.com/r/rust/comments/dt0ruy/how_to_await_futures_concurrently/

kornelski:
`thread::sleep` is a major offender. I've found that mistake done several times in a real-world codebase.

estebank:
I'm currently working on a linter that does exhaustive call-graph analysis to detect indirect usages of blocking calls in async functions. Right now this is working well, and requires either annotations (`#[redpen::blocking]`) or adding a fully-qualified path to a list `::std::thread::sleep`).

The exhaustive check might never land on `rustc` itself due to perf considerations (although I would love if it could be upstreamed to clippy once its more mature) *but* we *can* lint on the "simple"/direct case (`async fn foo(t: Duration) { sleep(t) }`) relatively easily. In order to do that, it'd be a good idea to add an annotation for `blocking` to the language.

Adding a top-level attribute feels like a waste, and restricting it to the `clippy` feels wrong, *but* we now have [the `#[diagnostic]` namespace](https://github.com/rust-lang/rust/issues/111996), so we could annotate `std::thread::sleep` and the like with `#[diagnostic::blocking]` for the benefit of both `rustc` and `clippy` to lint against *direct* uses (and my tool can get those attributes itself for its analysis and lint against *indirect* uses) *and* it provides crate authors a way to specify that their functions are blocking and their incorrect direct usage will also get linted.

CC @weiznich

PR #11578, opened 2023-09-28
new lint that detects blocking operations in async

J-ZhengLi:
fixes: #10794 

changelog: new lint [`unnecessary_blocking_ops`]

Ok, this is a rough one, I have no confident in many things (the name ofc), and there are certain things I wanna address before I put too much effort in it.

### DIFF
--- a/clippy_lints/src/unnecessary_blocking_ops.rs
+++ b/clippy_lints/src/unnecessary_blocking_ops.rs
@@ -1,0 +1,162 @@
+use std::ops::ControlFlow;
+
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::visitors::for_each_expr_with_closures;
+use clippy_utils::{def_path_def_ids, fn_def_id, is_lint_allowed};
+use rustc_data_structures::fx::FxHashMap;
+use rustc_errors::Applicability;
+use rustc_hir::def_id::DefId;
+use rustc_hir::hir_id::CRATE_HIR_ID;
+use rustc_hir::{Body, ExprKind, GeneratorKind, HirIdSet};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_session::{declare_tool_lint, impl_lint_pass};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for async function or async closure with blocking operations that
+    /// could be replaced with their async counterpart.
+    ///
+    /// ### Why is this bad?
+    /// Blocking a thread prevents tasks being swapped, causing other tasks to stop running
+    /// until the thread is no longer blocked, which might not be as expected in an async context.
+    ///
+    /// ### Known problems
+    /// Not all blocking operations can be detected, as for now, this lint only detects a small
+    /// set of functions in standard library by default. And some of the suggestions might need
+    /// additional features to work properly.
+    ///
+    /// ### Example
+    /// ```rust
+    /// use std::time::Duration;
+    /// pub async fn foo() {
+    ///     std::thread::sleep(Duration::from_secs(5));
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// use std::time::Duration;
+    /// pub async fn foo() {
+    ///     tokio::time::sleep(Duration::from_secs(5));
+    /// }
+    /// ```
+    #[clippy::version = "1.74.0"]
+    pub UNNECESSARY_BLOCKING_OPS,
+    nursery,
+    "blocking operations in an async context"
+}
+
+pub(crate) struct UnnecessaryBlockingOps {
+    blocking_ops: Vec<String>,
+    blocking_ops_with_suggs: Vec<[String; 2]>,
+    /// Map of resolved funtion def_id with suggestion string after checking crate
+    id_with_suggs: FxHashMap<DefId, String>,
+    /// Keep track of visited block ids to skip checking the same bodies in `check_body` calls
+    visited_block: HirIdSet,
+}
+
+impl UnnecessaryBlockingOps {
+    pub(crate) fn new(blocking_ops: Vec<String>, blocking_ops_with_suggs: Vec<[String; 2]>) -> Self {
+        Self {
+            blocking_ops,
+            blocking_ops_with_suggs,
+            id_with_suggs: FxHashMap::default(),
+            visited_block: HirIdSet::default(),
+        }
+    }
+}
+
+impl_lint_pass!(UnnecessaryBlockingOps => [UNNECESSARY_BLOCKING_OPS]);
+
+// TODO: Should we throw away all suggestions and and give full control to user configurations?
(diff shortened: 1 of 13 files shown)


clippy_lints/src/unnecessary_blocking_ops.rs
```diff
@@ -0,0 +1,173 @@
...
+    &["std", "fs", "write"],
+    // IO functions
+    &["std", "io", "copy"],
+    &["std", "io", "read_to_string"],
+];
+
+impl<'tcx> LateLintPass<'tcx> for UnnecessaryBlockingOps {
+    fn check_crate(&mut self, cx: &LateContext<'tcx>) {
+        // Avoids processing and storing a long list of paths if this lint was allowed entirely
+        if is_lint_allowed(cx, UNNECESSARY_BLOCKING_OPS, CRATE_HIR_ID) {
```
Jarcho, line 101: Lints levels can change anywhere.

clippy_lints/src/utils/conf.rs
```diff
@@ -586,6 +586,26 @@ define_Conf! {
...
+    /// Lint: UNNECESSARY_BLOCKING_OPS.
+    ///
+    /// List of additional blocking function paths to check, with replacement suggestion function paths.
+    ///
+    /// #### Example
+    ///
+    /// ```toml
+    /// blocking-ops-with-suggestions = [["my_crate::some_blocking_fn" , "my_crate::use_this_instead"]]
+    /// ```
+    (blocking_ops_with_suggestions: Vec<[String; 2]> = <_>::default()),
```
Jarcho, line 608: It would be nicer if these were the same config value. Easiest would be interpreting `["foo::bar", ""]` as having no suggestion. A custom deserializer would let you accept both `"foo::bar"` and `["foo::bar", "baz::bar"]`.

Jarcho (commented):
You can avoid the extra HIR traversals by keeping a stack of bodies and whether that body is async. `check_body` and `check_body_post` would maintain the stack and `check_expr` would be used to detect the calls.

You'll also need to make a distinction between async and maybe async contexts. Closures and nested functions inside an async context might run in an async context, but they may also not.

clippy_lints/src/unnecessary_blocking_ops.rs
J-ZhengLi: @Jarcho Sorry for late response, I was a bit busy dealing with other stuffs.

I'm not sure if I fully understand this phrase:
> Closures and nested functions inside an async context might run in an async context, but they may also not.

So, if there's an async function like this

```rust
async fn foo() {
    fn bar() {
        // hehe, sleep!
        std::time::sleep(INF)
    }
}
```

Then it shouldn't trigger any warning unless the function `bar` was called in `foo` later, correct?

clippy_lints/src/utils/conf.rs
J-ZhengLi: I have modified the `clippy_lints::conf::DisallowedPath` enum and re-use it... hope it'll be fine

clippy_lints/src/unnecessary_blocking_ops.rs
Jarcho: That is correct. Unless the call occurs directly within `foo` there's no way to know if the function is called in an async context. Something like `thread::spawn(|| sleep(10))` or `thread::spawn(bar)` would be fine since the `sleep` call doesn't block the current thread.

J-ZhengLi: > That is correct. Unless the call occurs directly within `foo` there's no way to know if the function is called in an async context. Something like `thread::spawn(|| sleep(10))` or `thread::spawn(bar)` would be fine since the `sleep` call doesn't block the current thread.

Ok, I think my code logic already prevent linting on such case, but I'm not sure, I'll add a few test cases tomorrow just in case

clippy_lints/src/unnecessary_blocking_ops.rs
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+    }
+
+    fn check_body(&mut self, cx: &LateContext<'tcx>, body: &'tcx Body<'tcx>) {
+        if is_lint_allowed(cx, UNNECESSARY_BLOCKING_OPS, body.value.hir_id) {
+            return;
+        }
+
+        if let Some(CoroutineKind::Async(_)) = body.coroutine_kind() {
```
J-ZhengLi, line 112: wait... where did `coroutine_kind` go in nightly 1.77.0?

clippy_lints/src/unnecessary_blocking_ops.rs
```diff
@@ -0,0 +1,182 @@
...
+    "std::fs::remove_dir",
+    "std::fs::remove_dir_all",
+    "std::fs::remove_file",
+    "std::fs::rename",
+    "std::fs::set_permissions",
+    "std::fs::symlink_metadata",
+    "std::fs::write",
+    // IO functions
+    "std::io::copy",
+    "std::io::read_to_string",
```
teor2345, line 93: These are only blocking if the underlying reader or writer are blocking. For example, copying between `Vec`s or reading from a `Vec` into a string isn't blocking in practice.

J-ZhengLi: > ...only blocking if the underlying reader or writer are blocking.

That's interesting, can you further explain this? 

Initially, I list those methods purely base on the observation from the `tokio` crate, since there are `AsyncRead` and `AsyncWrite` in the crate, I assume the `Read` and `Write` traits from `std::io` are blocking :smile: , but now that you mention it, I wonder if copying between slices would actually affect anything.

xFrednet:
Hey @Jarcho, this is a ping from triage. Can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@J-ZhengLi could you rebase on master again? Since only `conf.rs` and `lib.rs` need to be updated it should be safe to already review this :D

@rustbot ready

J-ZhengLi:
> ...
> @J-ZhengLi could you rebase on master again? Since only `conf.rs` and `lib.rs` need to be updated it should be safe to already review this :D
> ...

yeah sure, I forgot about this one, still feel very unconfident about this one, running `lintcheck` gives nothing lol (maybe because not many of those use `async` at all).

xFrednet:
I believe none of the lints in lintcheck use async extensively, but I could be wrong.

xFrednet:
Hey, this is a ping from triage. @Jarcho  can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@J-ZhengLi Sorry for the wait, could you rebase this PR?

@rustbot ready

J-ZhengLi:
Looking at current implementation, combining with the other issue #4377, I kinda want to close this and maybe re-work on it later :thinking:

PR #11580, opened 2023-09-28
[`manual_let_else`]: only omit block if span is from same ctxt

### LINKED ISSUE #11579: Manual let else suggests fix that doesn't compile
### Summary

While running clippy in pedantic mode, I got this "manual_let_else" diagnostic from clippy (see below).
The fix suggestion surprised me. I don't think it is valid syntax (though it does look very nice).

### Reproducer

I tried this code:

```rust
let msg = match Some("hi") {
    Some(m) => m,
    _ => unreachable!("can't happen"),
};
```

```
warning: this could be rewritten as `let...else`
 --> src\main.rs:2:5
  |
2 | /     let msg = match Some("hi") {
3 | |         Some(x) => x,
4 | |         _ => unreachable!("can't happen"),
5 | |     };
  | |______^ help: consider writing: `let Some(msg) = Some("hi") else { unreachable!("can't happen") };`
```

Instead, this happened:

```
warning: this could be rewritten as `let...else`
 --> src\main.rs:2:5
  |
2 | /     let msg = match Some("hi") {
3 | |         Some(x) => x,
4 | |         _ => unreachable!("can't happen"),
5 | |     };
  | |______^ help: consider writing: `let Some(msg) = Some("hi") else unreachable!("can't happen");`
```

### Version

```text
rustc 1.74.0-nightly (3223b0b5e 2023-09-20)
binary: rustc
commit-hash: 3223b0b5e8dadda3f76c3fd1a8d6c5addc09599e
commit-date: 2023-09-20
host: x86_64-pc-windows-msvc
release: 1.74.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

_No response_

y21:
Fixes #11579.

The lint already had logic for omitting a block in `else` if a block is already present, however this didn't handle the case where the block is from a different expansion/syntax context. E.g.
```rs
macro_rules! panic_in_block {
  () => { { panic!() } }
}

let _ = match Some(1) {
  Some(v) => v,
  _ => panic_in_block!()
};
```
It would see this in its expanded form as `_ => { panic!() }` and think it doesn't have to include a block in its suggestion because it is already there, however that's not true if it's from a different expansion like in this case.

changelog: [`manual_let_else`]: only omit block in suggestion if the block is from the same expansion

### DIFF
--- a/clippy_lints/src/manual_let_else.rs
+++ b/clippy_lints/src/manual_let_else.rs
@@ -136,9 +136,9 @@ fn emit_manual_let_else(
             // for this to be machine applicable.
             let mut app = Applicability::HasPlaceholders;
             let (sn_expr, _) = snippet_with_context(cx, expr.span, span.ctxt(), "", &mut app);
-            let (sn_else, _) = snippet_with_context(cx, else_body.span, span.ctxt(), "", &mut app);
+            let (sn_else, else_is_mac_call) = snippet_with_context(cx, else_body.span, span.ctxt(), "", &mut app);
 
-            let else_bl = if matches!(else_body.kind, ExprKind::Block(..)) {
+            let else_bl = if matches!(else_body.kind, ExprKind::Block(..)) && !else_is_mac_call {
                 sn_else.into_owned()
             } else {
                 format!("{{ {sn_else} }}")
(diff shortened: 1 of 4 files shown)


Jarcho:
`snippet_with_context` already returns whether the snippet is a macro call or not. That should be used instead.

y21:
Nice, that's even better.

Jarcho:
Thank you. @bors r+

PR #11587, opened 2023-09-30
new lint: `into_iter_without_iter`

### LINKED ISSUE #9736: Suggest implementing `IntoIterator for &'_ T` where `T` has `iter()` method returning an iterator (or vice versa)

y21:
Closes #9736 (part 2)

This implements the other lint that my earlier PR missed: given an `IntoIterator for &Type` impl, check that there exists an inherent `fn iter(&self)` method.

changelog: new lint: `into_iter_without_iter`

r? @Jarcho since you reviewed #11527 I figured it makes sense for you to review this as well?

### DIFF
--- a/clippy_lints/src/iter_without_into_iter.rs
+++ b/clippy_lints/src/iter_without_into_iter.rs
@@ -2,11 +2,13 @@ use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::get_parent_as_impl;
 use clippy_utils::source::snippet;
 use clippy_utils::ty::{implements_trait, make_normalized_projection};
+use rustc_ast::Mutability;
 use rustc_errors::Applicability;
-use rustc_hir::{FnRetTy, ImplItemKind, ImplicitSelfKind, TyKind};
+use rustc_hir::{FnRetTy, ImplItemKind, ImplicitSelfKind, ItemKind, TyKind};
 use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::{self, Ty};
 use rustc_session::{declare_lint_pass, declare_tool_lint};
-use rustc_span::sym;
+use rustc_span::{sym, Symbol};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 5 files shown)


clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -56,7 +102,77 @@ fn is_nameable_in_impl_trait(ty: &rustc_hir::Ty<'_>) -> bool {
...
+fn type_has_inherent_method(cx: &LateContext<'_>, ty: Ty<'_>, method_name: Symbol) -> bool {
+    if let Some(ty_did) = ty.ty_adt_def().map(ty::AdtDef::did) {
+        cx.tcx.inherent_impls(ty_did).iter().any(|did| {
+            cx.tcx
+                .hir()
+                .expect_item(did.expect_local())
+                .expect_impl()
+                .items
+                .iter()
+                .any(|item| item.ident.name == method_name)
```
Jarcho, line 114: This would be slightly better using the `associated_items` query rather than going through the HIR tree.

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -56,7 +102,77 @@ fn is_nameable_in_impl_trait(ty: &rustc_hir::Ty<'_>) -> bool {
     !matches!(ty.kind, TyKind::OpaqueDef(..))
 }
 
+fn type_has_inherent_method(cx: &LateContext<'_>, ty: Ty<'_>, method_name: Symbol) -> bool {
+    if let Some(ty_did) = ty.ty_adt_def().map(ty::AdtDef::did) {
+        cx.tcx.inherent_impls(ty_did).iter().any(|did| {
+            cx.tcx
+                .hir()
+                .expect_item(did.expect_local())
```
Jarcho, line 110: Might be better to return `None` for non-local types. The current orphan rules wouldn't allow it, but those could change in the future.

Jarcho (commented):
Just a couple of minor nits this time.

clippy_lints/src/iter_without_into_iter.rs
y21: Probably don't need this code if we're going with `associated_items`, since we can just give that a normal DefId

Jarcho: It will at least not be an ICE. Good enough given such a change is unlikely.

Jarcho:
Thank you. @bors r+

PR #11596, opened 2023-10-02
Move `needless_pass_by_ref_mut`: `suspicious` -> `nursery`

blyxyas:
[Related to [this Zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/needless_pass_by_ref_mut.20isn't.20ready.20for.20stable)]

`needless_pass_by_ref_mut` has been released with some important bugs (notably having a lot of reported false positives and an ICE). So it may not be really ready for being in stable until these problems are solved. This PR changes the lint's category from `suspicious` to `nursery`, just that.
changelog: none

### DIFF
--- a/clippy_lints/src/needless_pass_by_ref_mut.rs
+++ b/clippy_lints/src/needless_pass_by_ref_mut.rs
@@ -49,7 +49,7 @@ declare_clippy_lint! {
     /// ```
     #[clippy::version = "1.72.0"]
     pub NEEDLESS_PASS_BY_REF_MUT,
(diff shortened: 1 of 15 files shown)


Jarcho:
The changes to the other tests isn't needed, they should only need to be blessed. `needless_pass_by_ref_mut` should have been allowed on those tests in the first place.

blyxyas:
Oh, you're right. I'll change those right now.

Jarcho:
One more time. @bors r+

PR #11600, opened 2023-10-02
new lint: `danger_not_accepted`

Radbuglet:
Adds the `danger_not_accepted` lint:

> ### What it does
> 
> Checks for uses of functions, inherent methods, and trait methods which have been marked as
> dangerous with the `#[clippy::dangerous(...)]` attribute and whose dangers have not been
> explicitly accepted.
> 
> Each `#[clippy::dangerous(reason_1, reason_2, ...)]` attribute specifies a list of dangers
> that the user must accept using the `#[clippy::accept_danger(reason_1, reason_2, ...)]`
> attribute before using the dangerous item to avoid triggering this lint.
> 
> ### Why is this bad?
> 
> Some functionality in a project may be dangerous to use without giving it the appropriate
> caution, even if its misuse does not cause undefined behavior—for example, the method could
> be the source of tricky logic bugs. Other functionality may be dangerous in some contexts
> but not others. This lint helps ensure that users do not unknowingly call into these
> dangerous functions while still allowing users who know what they're doing to call these
> functions without issue.
> 
> ### Example
> ```rust
> #[clippy::dangerous(use_of_lib_1_dangerous_module)]
> mod dangerous_module {
>     #[clippy::dangerous(may_break_program)]
>     pub fn do_something_innocuous_looking() {
>         break_the_program();
>     }
> }
> 
> use unsuspecting_module {
>    fn do_something() {
>        // This function call causes clippy to issue a warning
>        super::dangerous_module::do_something_innocuous_looking();
>    }
> }
> ```
> Use instead:
> ```rust
> #[clippy::dangerous(use_of_lib_1_dangerous_module)]
> mod dangerous_module {
>     #[clippy::dangerous(may_break_program)]
>     pub fn do_something_innocuous_looking() {
>         break_the_program();
>     }
> }
> 
> // This entire module can use functions with the danger `use_of_lib_1_dangerous_module`.
> #[clippy::accept_danger(use_of_lib_1_dangerous_module)]
> use unsuspecting_module {
>    fn do_something() {
>        // Only this statement can call functions with the danger `may_break_program`.
>        #[clippy::accept_danger(may_break_program)]
>        super::dangerous_module::do_something_innocuous_looking();
>    }
> }
> ```

The following features have been deemed as future improvements and have not been included in this MVP:

- Allow users to override modules as *not* having a specific danger.
- Allow users to specify additional dangerous items in the clippy config.
- Devise a scheme (maybe path compression?) to reduce the amount of ancestry tracing we have to do to determine the dangers posed by a method.
- Implement a way to forbid `accept_danger` in a given module.
- Allow `accept_danger` and `dangerous` to be used as inner attributes on stable Rust.

changelog: new lint: `danger_not_accepted`


Radbuglet:
I am having trouble getting the doc-tests to pass locally. After fixing a typo where I declared my module with `use` rather than `mod`, the doc-test runner complains about not being able to find the `dangerous_module` in the crate root.

Here are the offending snippets:

```rust
#[clippy::dangerous(use_of_lib_1_dangerous_module)]
pub mod dangerous_module {
    # fn break_the_program() {}
    #[clippy::dangerous(may_break_program)]
    pub fn do_something_innocuous_looking() {
        break_the_program();
    }
}

pub mod unsuspecting_module {
   fn do_something() {
       // This function call causes clippy to issue a warning
       crate::dangerous_module::do_something_innocuous_looking();
   }
}
```
...produces the error:
```
---- src/danger_not_accepted.rs - danger_not_accepted::DANGER_NOT_ACCEPTED (line 42) stdout ----
error[E0433]: failed to resolve: could not find `dangerous_module` in the crate root
  --> src/danger_not_accepted.rs:55:15
   |
15 |        crate::dangerous_module::do_something_innocuous_looking();
   |               ^^^^^^^^^^^^^^^^ could not find `dangerous_module` in the crate root

For more information about this error, try `rustc --explain E0433`.
Couldn't compile the test.
```
...and the snippet:
```rust
#[clippy::dangerous(use_of_lib_1_dangerous_module)]
pub mod dangerous_module {
    # fn break_the_program() {}

    #[clippy::dangerous(may_break_program)]
    pub fn do_something_innocuous_looking() {
        break_the_program();
    }
}

// This entire module can use functions with the danger `use_of_lib_1_dangerous_module`.
#[clippy::accept_danger(use_of_lib_1_dangerous_module)]
pub mod unsuspecting_module {
    fn do_something() {
        // Only this statement can call functions with the danger `may_break_program`.
        #[clippy::accept_danger(may_break_program)]
        crate::dangerous_module::do_something_innocuous_looking();
    }
}
```
...produces the error:
```
---- src/danger_not_accepted.rs - danger_not_accepted::DANGER_NOT_ACCEPTED (line 60) stdout ----
error[E0433]: failed to resolve: could not find `dangerous_module` in the crate root
  --> src/danger_not_accepted.rs:76:15
   |
18 |        crate::dangerous_module::do_something_innocuous_looking();
   |               ^^^^^^^^^^^^^^^^ could not find `dangerous_module` in the crate root

For more information about this error, try `rustc --explain E0433`.
Couldn't compile the test.
```

These two snippets pass this fork of Clippy when checked on a standalone project so I'm not sure what's happening.

Alexendoo:
The doctests don't get run on a local `cargo test` due to the weird workspace setup we have, but it'll be because doctests wrap the contents in `fn main() { }` the module doesn't exist in the crate root, you can put `# fn main() {}` at the end of the doctest to disable the automatic wrapping

Alexendoo:
Nominating this for discussion in tomorrow's meeting (Tuesday, 15:00 UTC in https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy) since it's a novel lint

Centri3:
> Implement a way to forbid `accept_danger` in a given module.

This is already possible with `#![forbid(clippy::accept_danger)]`.

Radbuglet:
> > Implement a way to forbid `accept_danger` in a given module.
> 
> This is already possible with `#![forbid(clippy::accept_danger)]`.

Whoops—that explanation was a bit ambiguous. What I meant by that post-MVP item was that, in much the same way that some users may wish to `#[forbid(...)]` a specific lint, some users may find it helpful to forbid people from accepting certain dangers with another custom attribute like `#[clippy::forbid_danger(...)]`. That would look something like this:

```rust
#[clippy::dangerous(d)]
mod whee {
    pub fn woo() {}
}

#[clippy::forbid_danger(my_danger)]
mod i_am_super_safe {
    pub fn waz() {
        #[clippy::accept_danger(my_danger)]  // This would be forbidden.
        super::whee::woo();
    }
}
```

I could also see the benefit in forbidding users from accepting any more dangers beyond the ones declared by a similar attribute like `#[forbid_dangers_besides(allow_list_here)]`.

Again, however, I'm under the impression that making too big an addition to the Rust programming language is frowned upon so I'm leaving these as post-MVP features that can be taken up later.

Centri3:
> I think the best solution in the long-term would be to wait for Rust to add support for at least marking block expressions with attributes. That way, the user could type:
> 
> ```rust
> fn main() {
>     let foo = {
>         println!("Hi!");
>         #[clippy::accept_danger(waz)] {
>             dangerous() + 3
>         }
>     };
> }
> 
> #[clippy::dangerous(waz)]
> fn dangerous() -> i32 {
>     4
> }
> ```

Isn't this already possible? This is what you usually do to allow a lint for a macro (though rustfmt moves the opening bracket onto its own line; which, honestly, I hate).

I'm a bit unsure what you mean by "only happening in return positions", as an attribute there is perfectly fine:

```rust
fn awa() {
    #[allow(warnings)]
    ()
}
```

The reason why it didn't work there was because it could either be applied to the function call or the addition. The parser currently goes with the function call with `feature(stmt_expr_attributes)`.

I do agree that clippy shouldn't expose a macro like this, and I'm unsure how it would even work.

I see what you mean now about forbidding specific dangers. The reverse (`allow`ing) could be added to the `disallowed_*` lints, too

Radbuglet:
> I'm a bit unsure what you mean by "only happening in return positions", as an attribute there is perfectly fine:
> 
> ```rust
> fn awa() {
>     #[allow(warnings)]
>     ()
> }
> ```

`allow` is a special rustc attribute which gets special rustc treatment. If I try `awa` with a clippy-specific attribute, I get an error:

```rust
fn awa() -> () {
    #[clippy::msrv = ":)"]  // attributes on expressions are experimental
    ()
}
```

[Link to playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9124936873488a482ef47fc9313a6c4a)

It may be worthwhile seeing if we can get `clippy`'s attribute namespace promoted to exhibit some of the properties of builtin attributes. That way, `#[clippy::accept_danger(...)]` and company could be used in all contexts where an `allow` attribute is allowed. If that were the case, we would be able to write:

```rust
fn woo() -> i32 {
    id(
        #[clippy::accept_danger(my_danger)]
        (dangerous() + 1),
    )
}

#[clippy::dangerous(my_danger)]
fn dangerous() -> i32 {
    3
}

fn id<T>(v: T) -> T {
    v
}
```

...which feels pretty convenient to me.

Alexendoo:
> `allow` is a special rustc attribute which gets special rustc treatment

Huh, TIL

From the meeting we'd like for it to have a human readable message for more detailed error messages similar to `#[must_use]`

We also discussed that it should work across crate boundaries, but it looks like that should already work, I misread it as using a HIR visitor for both attributes initially

Alexendoo:
re: the attributes, yeah some of those are unfortunate but it's still the only real option for syntax. Some of those issues like the macro expansion one are shared with `unsafe` also

Radbuglet:
> From the meeting we'd like for it to have a human readable message for more detailed error messages similar to `#[must_use]`

This should be implemented now on `#[clippy::dangerous]` and `#[clippy::accept_danger]` although only `#[clippy::dangerous]`'s reasons show up in lint messages.

clippy_lints/src/danger_not_accepted.rs
```diff
@@ -0,0 +1,402 @@
...
+
+    // Create a lint
+    span_lint_and_then(
+        cx,
+        DANGER_NOT_ACCEPTED,
+        expr.span,
+        &format!(
+            "called a function marked with `#[clippy::dangerous(...)]` without blessing the calling \
+             module with `#![clippy::accept_danger({})]`",
+            fmt_inline(|f| {
```
Nilstrieb, line 373: this could just map the dangers to a `Vec<String>` and then `<[T]>::join` them instead

tests/ui/danger_not_accepted.rs
```diff
@@ -0,0 +1,78 @@
...
+
+trait FazTrait {
+    #[clippy::dangerous(may_delete_system)]
+    fn faz2(&self);
+}
+
+// Edge case attr tests
+#[rustfmt::skip]
+#[clippy::dangerous(whee, woo,)]
+#[clippy::dangerous(whee, woo, reason = "sdfhsdf",)]
```
Alexendoo, line 64: I think the reason should be per danger, i.e. `#[clippy::dangerous(foo = "reason")]`

Bikeshedding a bit, what about something like `#[clippy::danger::foo = "reason"]`? It's more similar to `#[must_use]` that way

clippy_lints/src/danger_not_accepted.rs
```diff
@@ -0,0 +1,375 @@
...
+    reason: Option<token::Lit>,
+}
+
+impl DangerNotAccepted {
+    fn get_unaccepted_dangers(&self, cx: &LateContext<'_>, item_id: def_id::DefId) -> Option<Vec<UnacceptedDanger>> {
+        let mut unaccepted_dangers = Vec::new();
+        let mut item_iter = Some(item_id);
+
+        while let Some(item_id) = item_iter {
+            item_iter = cx.tcx.opt_parent(item_id);
```
Alexendoo, line 171: What do you think about only supporting the danger definitions directly on functions? To me it feels similar to `unsafe`/`#[must_use]` which also apply individually

tests/ui/danger_not_accepted.rs
Radbuglet: Hmm. I like the `#[clippy::dangerous(a, b, reason = "foo")]` syntax because it allows you to share the same reason among multiple dangers. For example, `foo_internal_api` and `raw_foo_access` could share the reason "low level building block; use bar instead." If you want several different reasons, you could always write it as:

```rust
#[clippy::dangerous(foo, reason = "reason 1")]
#[clippy::dangerous(bar, reason = "reason 2")]
fn baz() {}
```

clippy_lints/src/danger_not_accepted.rs
Radbuglet: I think, for dangers like `unchecked_foo_api`, users are likely going to reuse that same danger for many different functions (or even entire modules!) and would likely benefit from things like this:

```rust
#[clippy::dangerous(unchecked_foo_api)]
impl Foo {
    pub fn get_raw(&self) {}

    pub fn set_raw(&self) {}

    pub fn update_raw(&self) {}
}
```

Would the benefit to this be mainly a performance one? If so, there's always the possibility of applying path compression to reduce the amount of looking one may have to do.

jhpratt:
What if two crates use the same ident? Should they be namespaced to the crate?

Radbuglet:
> What if two crates use the same ident? Should they be namespaced to the crate?

I don't think they should _have_ to be namespaced to a crate since some dangers may be quite common (e.g. `main_thread_exclusive` could be used by e.g. `libc` and `winit`). However, I do see the value of implementing some generic way to namespace things, even if it doesn't have any particularly strong enforcement. Maybe we could add support for `::` in danger idents?

jhpratt:
Sure, supporting `::` without enforcing anything would also work. I would personally adopt the convention of prefixing the crate name.

tests/ui/danger_not_accepted.rs
Alexendoo: Personally I don't think it will be that common to want to use multiple dangers on the same node, particularly so where the reason can be the same for all of them

clippy_lints/src/danger_not_accepted.rs
Alexendoo: The perf aspect is the larger part of the concern yeah, though I haven't ran any kind of benchmark to see how much of an issue it actually is

tests/ui/danger_not_accepted.rs
Alexendoo: Relatedly there was discussion about whether or not the reason should be required, I don't think we fully came to an agreement on that but initially let's make it required, it can be made optional later if we want that but not the other way around

Radbuglet:
Sorry for ghosting this thread for a while. Danger identifiers can now take the form of paths and reasons are mandatory for `#[dangerous]` attributes and use the `danger::path = "justification literal"`.

Alexendoo:
Did the `#[clippy::danger::foo = "reason"]` change get pushed? Sorry for the large delay I thought that was being waited on as the tests still show the original syntax, but I missed that you commented that it was implemented

xFrednet:
Hey @Radbuglet , this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

@rustbot author

Radbuglet:
Whoops. I meant to respond to this thread but forgot. Sorry!

Yes, the `#[clippy::danger::foo = "reason"]` syntax is implemented and currently replaces the old `#[clippy::danger(clippy_danger_foo, reason = "reason")]` syntax. The parser for this custom syntax provides good diagnostics, too.

The following, however, are not implemented:
- I think I forgot to update the documentation but we should probably wait on that until all the other details are finalized.
- I don't yet support `accept_danger` on entire namespaces—maybe that's a good idea?
- You can still place `#[clippy::danger]` attributes on any HIR item and all the descendants will still inherit the danger set.

Radbuglet:
Bumping this PR a bit... are there any additional changes I should make to this PR? My [previous comment](https://github.com/rust-lang/rust-clippy/pull/11600#issuecomment-2028879266) includes a list of incomplete things I remember being mentioned.

Alexendoo:
> * I think I forgot to update the documentation but we should probably wait on that until all the other details are finalized.

Yeah that's fine, the PR itself is still using [the old syntax](https://github.com/rust-lang/rust-clippy/blob/a1497a461f41a0a3b55d64ffe7c9d48491d50417/tests/ui/danger_not_accepted.rs#L40-L41) though

> * I don't yet support `accept_danger` on entire namespaces—maybe that's a good idea?

What do you mean by namespace?

> * You can still place `#[clippy::danger]` attributes on any HIR item and all the descendants will still inherit the danger set.

Let's start conservative and have it be direct only like `#[must_use]`

Radbuglet:
> > * I think I forgot to update the documentation but we should probably wait on that until all the other details are finalized.
> 
> Yeah that's fine, the PR itself is still using [the old syntax](https://github.com/rust-lang/rust-clippy/blob/a1497a461f41a0a3b55d64ffe7c9d48491d50417/tests/ui/danger_not_accepted.rs#L40-L41) though

Huh? The line you shared uses the new syntax:

```rust
#[clippy::dangerous(may_deadlock = "your program may deadlock in calling this function")]
pub fn woo2() {}
```

The new syntax is:

```rust
fn caller() {
    #[clippy::accept_danger(path::to::danger, reason = "this part is optional")]
    callee();
}

#[clippy::dangerous(path::to::danger = "why it's dangerous")]
fn callee() {

}
```

> 
> > * I don't yet support `accept_danger` on entire namespaces—maybe that's a good idea?
> 
> What do you mean by namespace?

I think it might be useful to be able to accept entire classes of dangers. For example:

```rust
#[clippy::accept_danger(my_lib, reason = "we know what we're doing")]
mod trusted {
    fn whee() {
        super::func_1();
        super::func_2();
    }
}

#[clippy::dangerous(my_lib::func1 = "reason 1")]
fn func_1() { ... }

#[clippy::dangerous(my_lib::func2 = "reason 2")]
fn func_2() { ... }
```

However, if this is out-of-scope for an initial implementation, we can skip it.

> > * You can still place `#[clippy::danger]` attributes on any HIR item and all the descendants will still inherit the danger set.
> 
> Let's start conservative and have it be direct only like `#[must_use]`

Sounds good!

Alexendoo:
Ah I see the confusion, the intention was to have the danger be part of the attr path e.g.

```rust
#[clippy::danger::may_deadlock = "your program may deadlock in calling this function"]
pub fn woo2() {}
```

We could accept paths on top of just idents, but it may be worth avoiding so people don't feel the need to add namespaces everywhere

xFrednet:
Hey @Radbuglet , this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

@rustbot author

xFrednet:

PR #11609, opened 2023-10-04
[`get_first`]: lint on non-primitive slices

### LINKED ISSUE #11594: `get_first` doesn't lint on slices of non-primitive types
### Summary

We have a lint for `slice.get(0)`, but I've noticed that in many cases this just doesn't warn.  I looked into the code and there's this check:
https://github.com/rust-lang/rust-clippy/blob/aee3daf90ba36fb14e3b9f04ea0eea97fd2ec6ad/clippy_lints/src/methods/get_first.rs#L23

I'm honestly not sure why this is there. `.first()` should work just as well with non-primitives?
I removed the check locally and all tests still pass so it remains a mystery to me. Doesn't look like that was mentioned in the initial PR or in any issues. Opening an issue first because someone might have a clue as to why it's necessary.

### Lint Name

`get_first`

### Reproducer

I tried this code:

```rust
let strings = &[String::new()];
let v: Option<&String> = strings.get(0);
```

```rs
warning: accessing first element with `strings.get(0)`
   --> test.rs:164:30
    |
164 |     let v: Option<&String> = strings.get(0);
    |                              ^^^^^^^^^^^^^^ help: try: `strings.first()`
```

Instead, this happened:
nothing.

### Version

```text
rustc 1.74.0-nightly (37390d656 2023-09-24)
binary: rustc
commit-hash: 37390d65636dd67e263753a3c04fbc60dcc4348e
commit-date: 2023-09-24
release: 1.74.0-nightly
LLVM version: 17.0.0
```

y21:
Fixes #11594

I left the issue open for a couple days before making the PR to see if anyone has something to say, but it looks like there aren't any objections to removing this check that prevented linting on non-primitive slices, so here's the PR now.
There's a couple of instances in clippy itself where we now emit the lint. The actual relevant change is in the first commit and fixing the `.get(0)` instances in clippy itself is in the 2nd commit.

changelog: [`get_first`]: lint on non-primitive slices

### DIFF
--- a/clippy_lints/src/loops/same_item_push.rs
+++ b/clippy_lints/src/loops/same_item_push.rs
@@ -185,7 +185,7 @@ fn get_vec_push<'tcx>(
             if let StmtKind::Semi(semi_stmt) = &stmt.kind;
             if let ExprKind::MethodCall(path, self_expr, args, _) = &semi_stmt.kind;
             // Figure out the parameters for the method call
-            if let Some(pushed_item) = args.get(0);
+            if let Some(pushed_item) = args.first();
             // Check that the method being called is push() on a Vec
             if is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(self_expr), sym::Vec);
             if path.ident.name.as_str() == "push";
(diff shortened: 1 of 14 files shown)


blyxyas (approved):

giraffate (approved):

giraffate:
@bors r+

Thanks!

Issue #11610, opened 2023-10-04
FP with `needless_pass_by_ref_mut`, async function and closures


WaffleLapkin:
### Summary

`needless_pass_by_ref_mut` lint is wrongfully emitted in some weird cases involving `async` functions and closures.

Note that  #11551 was supposed to fix a similar issue, however the latest nightly still wrongfully emits a lint in the example below.

I did not debug it, but the lint seems to be extremely flaky, for example at one point during the minimization removing an unused `use` resulted in a lint not being emitted, at another point removing unused field caused a similar effect, but then removing another unused field brought the lint back... 

### Lint Name

`needless_pass_by_ref_mut`

### Reproducer

I tried this code:

```rust
async fn _f(v: &mut Vec<()>) {
    let x = || v.pop();
    _ = || || x;
}
```
([play](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=88c19ed7c78a0291ad0bc5bfc1756b7f))

I saw this happen:

```
warning: this argument is a mutable reference, but not used mutably
 --> src/lib.rs:1:16
  |
1 | async fn _f(v: &mut Vec<()>) {
  |                ^^^^^^^^^^^^ help: consider changing to: `&Vec<()>`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
  = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default
```

I expected to see this happen: no warning, `v` is used mutably via `.pop()`.

### Version

```text
1.75.0-nightly (2023-10-03 187b8131d4f760f856b2)
```

GuillaumeGomez:
It was already fixed. Sending a PR to add this test case.

GuillaumeGomez:
In https://github.com/rust-lang/rust-clippy/pull/11492. Someone asked the same question as you in the PR and you have the answer in it. ;)

PR #11622, opened 2023-10-06
Add regression test for #11610 about mutable usage of argument in async function for the `needless_pass_by_ref_mut` lint

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust-clippy/issues/11610.

This was already fixed. I simply added a regression test.

changelog: Add regression test for #11610 about mutable usage of argument in async function for the `needless_pass_by_ref_mut` lint

### DIFF
--- a/tests/ui/needless_pass_by_ref_mut.rs
+++ b/tests/ui/needless_pass_by_ref_mut.rs
@@ -270,6 +270,12 @@ pub async fn closure4(n: &mut usize) {
     })();
 }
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Seems like Centri3 is busy so I was recommended another reviewer.

r? @blyxyas

blyxyas (approved):
LGTM!	Thanks :heart: (=^･^=)

Issue #11586, opened 2023-09-29
`needless_pass_by_ref_mut` should not be emitted for functions containing `unsafe`


Ralith:
### Description

A unique borrow can be required for soundness in a function that performs unsafe operations which do not directly use a reference, e.g. using raw pointers or over FFI.

### Version

```text
rustc 1.74.0-nightly (2f5df8a94 2023-08-31)
binary: rustc
commit-hash: 2f5df8a94bb3c5fae4e3fcbfc8ef20f1f976cb19
commit-date: 2023-08-31
host: x86_64-unknown-linux-gnu
release: 1.74.0-nightly
LLVM version: 17.0.0
```

Issue #11180, opened 2023-07-18
needless_pass_by_ref_mut suggests (potentially unsound) code that triggers mut_from_ref


taiki-e:
### Summary

needless_pass_by_ref_mut (warn by default) suggests code that triggers mut_from_ref (error by default).

The suggested code itself is also suspicious from a soundness point of view.

Mentioning @GuillaumeGomez, who implement this lint in #10900.

### Reproducer

I tried this code:

```rust
#![allow(dead_code)]

use std::ptr::NonNull;

struct Data<T: ?Sized> {
    value: T,
}

unsafe fn get_mut_unchecked<T>(ptr: &mut NonNull<Data<T>>) -> &mut T {
    unsafe { &mut (*ptr.as_ptr()).value }
}
```

This happened:

```
warning: this argument is a mutable reference, but not used mutably
 --> src/lib.rs:7:37
  |
7 | unsafe fn get_mut_unchecked<T>(ptr: &mut NonNull<Data<T>>) -> &mut T {
  |                                     ^^^^^^^^^^^^^^^^^^^^^ help: consider changing to: `&NonNull<Data<T>>`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_ref_mut
  = note: `#[warn(clippy::needless_pass_by_ref_mut)]` on by default
```

And when I applied the suggestion:

```rs
#![allow(dead_code)]

use std::ptr::NonNull;

struct Data<T: ?Sized> {
    value: T,
}

unsafe fn get_mut_unchecked<T>(ptr: &NonNull<Data<T>>) -> &mut T {
    unsafe { &mut (*ptr.as_ptr()).value }
}
```

Then, this happened:

```
error: mutable borrow from immutable input(s)
 --> src/lib.rs:9:59
  |
9 | unsafe fn get_mut_unchecked<T>(ptr: &NonNull<Data<T>>) -> &mut T {
  |                                                           ^^^^^^
  |
note: immutable borrow here
 --> src/lib.rs:9:37
  |
9 | unsafe fn get_mut_unchecked<T>(ptr: &NonNull<Data<T>>) -> &mut T {
  |                                     ^^^^^^^^^^^^^^^^^
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#mut_from_ref
  = note: `#[deny(clippy::mut_from_ref)]` on by default
```

### Version

```text
[...]

PR #11624, opened 2023-10-06
Don't emit `needless_pass_by_ref_mut` if the variable is used in an unsafe block or function

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust-clippy/issues/11586.
Fixes https://github.com/rust-lang/rust-clippy/issues/11180.

As suggested in the two issues above, this lint should not be emitted if this an unsafe function or if the argument is used in an unsafe block.

changelog: [`needless_pass_by_ref_mut`]: Don't emit if the variable is used in an unsafe block or function

### DIFF
--- a/clippy_lints/src/needless_pass_by_ref_mut.rs
+++ b/clippy_lints/src/needless_pass_by_ref_mut.rs
@@ -7,7 +7,8 @@ use rustc_data_structures::fx::{FxHashSet, FxIndexMap};
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::{walk_qpath, FnKind, Visitor};
 use rustc_hir::{
-    Body, Closure, Expr, ExprKind, FnDecl, HirId, HirIdMap, HirIdSet, Impl, ItemKind, Mutability, Node, PatKind, QPath,
+    BlockCheckMode, Body, Closure, Expr, ExprKind, FnDecl, HirId, HirIdMap, HirIdSet, Impl, ItemKind, Mutability, Node,
+    PatKind, QPath,
 };
 use rustc_hir_typeck::expr_use_visitor as euv;
 use rustc_infer::infer::{InferCtxt, TyCtxtInferExt};
@@ -139,13 +140,23 @@ impl<'tcx> LateLintPass<'tcx> for NeedlessPassByRefMut<'tcx> {
         let hir_id = cx.tcx.hir().local_def_id_to_hir_id(fn_def_id);
         let is_async = match kind {
             FnKind::ItemFn(.., header) => {
+                if header.is_unsafe() {
+                    // We don't check unsafe functions.
+                    return;
+                }
                 let attrs = cx.tcx.hir().attrs(hir_id);
                 if header.abi != Abi::Rust || requires_exact_signature(attrs) {
                     return;
                 }
                 header.is_async()
             },
-            FnKind::Method(.., sig) => sig.header.is_async(),
+            FnKind::Method(.., sig) => {
+                if sig.header.is_unsafe() {
+                    // We don't check unsafe functions.
+                    return;
+                }
+                sig.header.is_async()
+            },
             FnKind::Closure => return,
         };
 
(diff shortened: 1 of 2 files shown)


clippy_lints/src/needless_pass_by_ref_mut.rs
```diff
@@ -304,6 +315,23 @@ impl<'tcx> MutablyUsedVariablesCtxt<'tcx> {
         }
         self.aliases.insert(alias, target);
     }
+
+    // The goal here is to find if the current scope is unsafe or not. It stops when it finds
+    // a function or an unsafe block.
+    fn is_in_unsafe_block(&self, mut item: HirId) -> bool {
+        let hir = self.tcx.hir();
+        while let Some(parent) = hir.get_enclosing_scope(item) {
```
Centri3, line 323: This should use `parent_iter` imo, mostly to make `item` immutable. `get_enclosing_scope` just iterates over the parents until it finds a block/item, but returns an `HirId`. Then you can call `fn_sig` on the `Node`, which is what `fn_sig_by_hir_id` does.

Centri3 (approved):
Ignoring all `unsafe` functions is the right move. LGTM, just one nit

clippy_lints/src/needless_pass_by_ref_mut.rs
GuillaumeGomez: I keep forgetting about `parent_iter`... Thanks for the lead!

GuillaumeGomez:
Updated!

tests/ui/needless_pass_by_ref_mut.rs
```diff
@@ -270,6 +275,18 @@ pub async fn closure4(n: &mut usize) {
...
+}
+// Unsafe functions should not warn.
+unsafe fn get_mut_unchecked<T>(ptr: &NonNull<Data<T>>) -> &mut T {
+    &mut (*ptr.as_ptr()).value
+}
+// Unsafe blocks should not warn.
+fn get_mut_unchecked2<T>(ptr: &NonNull<Data<T>>) -> &mut T {
+    unsafe { &mut (*ptr.as_ptr()).value }
+}
+
```
Centri3, line 289: Shouldn't these be `&mut NonNull<Data<T>>`, or am I missing something?

Also, I think we could use 2 more tests here - one which ensures it doesn't lint even without any `unsafe` operations (this could be linted in the future, but being pessimistic is best) and one that ensures that, even if the prior does, rebinding it to another local (like, say, `let pdata = ptr;`) then doing smth unsafe won't lint, to show that it would track where these bindings go.

Centri3: This also creates a mutable reference from an immutable reference, that seems like UB ^^ I presume this is what the lint suggested before?

Centri3 (changes requested):
On a closer look these tests don't seem correct (`&` instead of `&mut`). Also I have a suggestion for the tests

GuillaumeGomez:
Updated the test and fixed the code (the check was done on the wrong `HirId`).

GuillaumeGomez:
Seems like Centri3 is busy so I was recommended another reviewer.

r? @blyxyas

GuillaumeGomez:
Fixed conflict.

blyxyas (approved):
LGTM, thanks! :heart:

clippy_lints/src/needless_pass_by_ref_mut.rs
```diff
@@ -304,10 +315,27 @@ impl<'tcx> MutablyUsedVariablesCtxt<'tcx> {
         }
         self.aliases.insert(alias, target);
     }
+
+    // The goal here is to find if the current scope is unsafe or not. It stops when it finds
+    // a function or an unsafe block.
```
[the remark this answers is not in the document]
GuillaumeGomez: I didn't intend them to be doc comments but I can change them if you prefer?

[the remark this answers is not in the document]
blyxyas: I thought it was a typo :sweat_smile:. Don't worry, as it isn't a public API, rustdoc documentation isn't really important. I'll just merge it as-is.

PR #11627, opened 2023-10-06
[`needless_return_with_question_mark`]: don't lint if never type is used for coercion

### LINKED ISSUE #11616: `needless_return_with_question_mark` lints when never type is used for coercion
### Summary

Inspired by #11613. It's similar but imo not quite the same issue. This one is more about coercions:
```rs
fn foo() -> Result<(), ()> {
    let _x: String = {
        return Err(())?; // <--- should not lint here
    };
    Ok(())
}
```
`return`'s never type is used to get typeck to coerce the type of the block to something else, but that won't happen with its suggestion (just `Err(...)?`).

It might be that the block here has the `NeverToAny` adjust that we can use to suppress the lint in these cases.

### Lint Name

needless_return_with_question_mark

### Reproducer

🆙 

### Version

```text
rustc 1.74.0-nightly (37390d656 2023-09-24)
binary: rustc
commit-hash: 37390d65636dd67e263753a3c04fbc60dcc4348e
commit-date: 2023-09-24
release: 1.74.0-nightly
LLVM version: 17.0.0
```

### Additional Labels

y21:
Fixes #11616

When we have something like
```rs
let _x: String = {
  return Err(())?;
};
```
we shouldn't suggest removing the `return` because the `!`-ness of `return` is used to coerce the enclosing block to some other type. That will lead to a typeck error without a diverging expression like `return`.

changelog: [`needless_return_with_question_mark`]: don't lint if `return`s never typed-ness is used for coercion

### DIFF
--- a/clippy_lints/src/returns.rs
+++ b/clippy_lints/src/returns.rs
@@ -7,10 +7,12 @@ use core::ops::ControlFlow;
 use rustc_errors::Applicability;
 use rustc_hir::intravisit::FnKind;
 use rustc_hir::{
-    Block, Body, Expr, ExprKind, FnDecl, ItemKind, LangItem, MatchSource, OwnerNode, PatKind, QPath, Stmt, StmtKind,
+    Block, Body, Expr, ExprKind, FnDecl, HirId, ItemKind, LangItem, MatchSource, Node, OwnerNode, PatKind, QPath, Stmt,
+    StmtKind,
 };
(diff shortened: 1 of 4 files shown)


giraffate (approved):
Sorry for the late reviewing. Can you resolve the conflicts?

giraffate:
@bors r+

Thanks!

PR #11633, opened 2023-10-06
WIP: empty_doc

### LINKED ISSUE #9931: Detect documentation that is empty
### What it does

Detects documentation that is empty.

### Lint Name

empty-docs

### Category

suspicious

### Advantage

- Empty docs are often pointless
- The developer may have forgotten to fill in the documentation.

### Drawbacks

- Could result in the commonly used `missing_docs` lint being fired. However it should be possible to just use the attribute `#[allow(missing_docs)]` where needed.
- If `#[forbid(missing_docs]` is enabled, removing the empty doc comments will fire the lint.
- What applies as "empty documentation" would need to consider whitespace and possibly odd characters. Using [`str::is_whitespace`](https://doc.rust-lang.org/std/primitive.char.html#method.is_whitespace) would probably satisfy that case.

### Example

```rust
//!
```

Could be written as (if `#![warn(missing_docs)]` or `#![deny(missing_docs)]`):

```rust
#![allow(missing_docs)]
```

or a message:
> help: fill in the documentation

Could be written as (if missing_docs is not enabled)

```rust
```

If  `#[forbid(missing_docs)]`, then we probably should advise changing forbid to deny or show the message to fill in the documentation.

SET001:
Fixes #9931 
```
changelog: [`empty_doc`]: Detects documentation that is empty.
```

### DIFF
--- a/clippy_lints/src/empty_docs.rs
+++ b/clippy_lints/src/empty_docs.rs
@@ -41,20 +41,17 @@ fn trim_comment(comment: &str) -> String {
 
 impl EarlyLintPass for EmptyDocs {
     fn check_attribute(&mut self, cx: &EarlyContext<'_>, attribute: &Attribute) {
-        match attribute.kind {
-            AttrKind::DocComment(_line, comment) => {
-                if trim_comment(comment.as_str()).len() == 0 {
-                    span_lint_and_help(
-                        cx,
-                        EMPTY_DOCS,
-                        attribute.span,
-                        "empty doc comment",
-                        None,
-                        "consider removing or fill it",
-                    );
-                }
-            },
-            _ => {},
+        if let AttrKind::DocComment(_line, comment) = attribute.kind {
+            if trim_comment(comment.as_str()).len() == 0 {
+                span_lint_and_help(
+                    cx,
+                    EMPTY_DOCS,
+                    attribute.span,
+                    "empty doc comment",
+                    None,
+                    "consider removing or fill it",
+                );
+            }
         }
     }
 }
--- a/clippy_lints/src/empty_docs.rs
+++ b/clippy_lints/src/empty_docs.rs
@@ -1,0 +1,61 @@
+use clippy_utils::diagnostics::span_lint_and_help;
+use rustc_ast::ast::*;
+use rustc_lint::{EarlyContext, EarlyLintPass};
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Detects documentation that is empty.
+
+    /// ### Why is this bad?
+    /// It is unlikely there is any reason to have empty documentation for an entity
+    ///
+    /// ### Example
+    /// ```rust
+    /// ///
+    /// fn returns_true() {
+    ///     true
+    /// }
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// fn returns_true() {
+    ///     true
+    /// }
+    /// ```
+    #[clippy::version = "1.74.0"]
+    pub EMPTY_DOCS,
+    suspicious,
+    "docstrings exist but documentation is empty"
+}
+
+declare_lint_pass!(EmptyDocs => [EMPTY_DOCS]);
+
+fn trim_comment(comment: &str) -> String {
+    comment
+        .trim()
+        .split("\n")
+        .map(|comment| comment.trim().trim_matches('*').trim_matches('!'))
+        .collect::<Vec<&str>>()
+        .join("")
+}
+
+impl EarlyLintPass for EmptyDocs {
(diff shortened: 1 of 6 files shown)


clippy_lints/src/empty_docs.rs
```diff
@@ -0,0 +1,61 @@
...
+                        EMPTY_DOCS,
+                        attribute.span,
+                        "empty doc comment",
+                        None,
+                        "consider removing or fill it",
+                    );
+                }
+            },
+            _ => {},
+        }
```
blyxyas, line 59: An `if let AttrKind::DocComment(_, comment) = attribute.kind` would make this more readable :cat2:

blyxyas (changes requested):
Very good for a first patch :heart, mainly just some documentation rephrasing and refactoring to fit better with the rest of the codebase :cat2:

tests/ui/empty_docs.rs
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     *
+     */
+
+    /**
+     * valid block doc comment
+     */
+
+    pub mod inner_module {}
```
Alexendoo, line 26: None of these are actually empty documentation, this is ~equivalent to

```rust
//!
//! valid doc comment
//!!
//!! valid doc comment

///
/// valid doc comment
/// 
/// valid block doc comment
```

Because doc comments and `#[doc]` attributes are merged together

This really should go in `docs.rs` where this merging/parsing is all handled already

SET001: @Alexendoo according to https://doc.rust-lang.org/reference/comments.html there are one line and block doc comments.
What you show in your example is 8 separate one-line doc comments. They are not being merged in 2, just because there are no line breaks within them. So lint in this PR will trigger on lines 1, 3, 5, 6 (and I'm updating the test to demonstrate this).

As for block doc comments, I assume they might have empty lines (for formatting purposes) because the definition of `empty` for a block comment is when all lines are empty.

Tbh I do not understand the part about merging with `#[doc]` attributes. Can you please explain?

SET001:
```
This really should go in doc.rs where this merging/parsing is all handled already
```
there is no much merging/parsing happening in this current lint. The idea is to have all doc-related lints in doc.rs?

Alexendoo: I'm not saying the attributes are merged in the AST/HIR, but the multiple attributes representing doc comments are merged into a single piece of documentation by rustdoc, be they line comments, block comments or [`#[doc = ".."]` attributes](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html)

My example was to represent what rustdoc approximately sees as the documentation, not to say that doc comments get converted to line comments. A screenshot may be clearer:

```rust
    ///

    /// valid doc comment

    /**
     *
     */

    /**
     * valid block doc comment
     */

    pub mod inner_module {}
```

image

The whitespace between the doc comments are not significant, e.g.


```rust
/// a

///

/// b
```

is the same as

```rust
/// a
///
/// b
```

But we certainly don't want to call the middle line comment empty, it's significant because it causes `a` and `b` to be separate paragraphs

Alexendoo: > there is no much merging/parsing happening in this current lint. The idea is to have all doc-related lints in doc.rs?

Yeah, it could go here

https://github.com/rust-lang/rust-clippy/blob/3662bd8f587c9305e4d231a9456e63b92b8a7729/clippy_lints/src/doc.rs#L493-L495

With `.is_empty()` being changed to check that `doc` only contains whitespace if needed, you can use https://doc.rust-lang.org/nightly/nightly-rustc/rustc_resolve/rustdoc/fn.span_of_fragments.html on `fragments` to get the span

clippy_lints/src/empty_docs.rs
```diff
@@ -0,0 +1,57 @@
...
+    "docstrings exist but documentation is empty"
+}
+
+declare_lint_pass!(EmptyDocs => [EMPTY_DOCS]);
+
+fn trim_comment(comment: &str) -> String {
+    comment
+        .trim()
+        .split("\n")
+        .map(|comment| comment.trim().trim_matches('*').trim_matches('!'))
```
GuillaumeGomez, line 37: I think that one shouldn't be needed.

xFrednet:
Hey @SET001, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

@rustbot author

y21:
Looks like this was continued in #12342 because of inactivity. The lint exists now: https://rust-lang.github.io/rust-clippy/master/index.html#/empty_docs

xFrednet:
Thank you for the comment @y21. Then let's close this one.

@SET001 thank you for the time you put into this PR!

PR #11656, opened 2023-10-10
[`unnecessary_to_owned`]: catch `to_owned` on byte slice to create temporary `&str`

### LINKED ISSUE #11648: taking reference on String::from_utf8 result whose argument is converted from slice can be rewritten in borrowed version of function
### What it does

`&String::from_utf8(slice.to_vec()).unwrap_friend()` can be `core::str::from_utf8(slice).unwrap_friend()` where `&str` is expected.

### Advantage

- Remove additional allocation
- Reduces instruction count by 2x on release mode ([godbolt](https://rust.godbolt.org/z/P9Mqas753))

### Drawbacks

n/a

### Example

```rust
fn opaque(buf: &mut [u8]) {}

let mut foo = String::new();
let slice = [0u8; 1024];
opaque(&mut slice);

let ref_str: &str = &String::from_utf8(slice.to_vec()).expect("not UTF-8");
```

Could be written as:

```rust
let ref_str: &str = core::str::from_utf8(slice).expect("not UTF-8");
```

y21:
Closes #11648

Detects the pattern `&String::from_utf8(bytes.to_vec()).unwrap()` and suggests `core::str::from_utf8(bytes).unwrap()`, which avoids the unnecessary intermediate allocation.

I decided to put this in the existing `unnecessary_to_owned` lint (rather than creating a new lint) for a few reasons:
- we get to use some of its logic (for example, recognizing any of the functions in the `to_owned` family, e.g. `to_vec`)
- the actual inefficient operation that can be avoided here is the call to `.to_vec()`, so this is in a way similar to the other cases caught by `unnecessary_to_owned`, just through a bunch of type conversions
- we can make this more "generic" later and catch other cases, so imo it's best not to tie this lint specifically to the `String` type

changelog: [`unnecessary_to_owned`]: catch `&String::from_utf8(bytes.to_vec()).unwrap()` and suggest `core::str::from_utf8(bytes).unwrap()`

### DIFF
--- a/clippy_lints/src/methods/unnecessary_to_owned.rs
+++ b/clippy_lints/src/methods/unnecessary_to_owned.rs
@@ -1,13 +1,15 @@
 use super::implicit_clone::is_clone_like;
 use super::unnecessary_iter_cloned::{self, is_into_iter};
 use clippy_config::msrvs::{self, Msrv};
-use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::source::snippet_opt;
+use clippy_utils::diagnostics::{span_lint_and_sugg, span_lint_and_then};
+use clippy_utils::source::{snippet, snippet_opt};
 use clippy_utils::ty::{
     get_iterator_item_ty, implements_trait, is_copy, is_type_diagnostic_item, is_type_lang_item, peel_mid_ty_refs,
 };
 use clippy_utils::visitors::find_all_ret_expressions;
-use clippy_utils::{fn_def_id, get_parent_expr, is_diag_item_method, is_diag_trait_item, return_ty};
+use clippy_utils::{
+    fn_def_id, get_parent_expr, is_diag_item_method, is_diag_trait_item, match_def_path, paths, return_ty,
+};
 use rustc_errors::Applicability;
 use rustc_hir::def::{DefKind, Res};
 use rustc_hir::def_id::DefId;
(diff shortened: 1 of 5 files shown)


xFrednet:
(I claimed a few PRs, let's give this one to the lottery)

r? clippy

clippy_lints/src/methods/unnecessary_to_owned.rs
```diff
@@ -242,6 +247,65 @@ fn check_into_iter_call_arg(
...
+/// Checks for `&String::from_utf8(bytes.{to_vec,to_owned,...}()).unwrap()` coercing to `&str`,
+/// which can be written as just `std::str::from_utf8(bytes).unwrap()`.
+fn check_string_from_utf8<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, receiver: &'tcx Expr<'tcx>) -> bool {
+    if let Some((call, arg)) = skip_addr_of_ancestors(cx, expr)
+        && !arg.span.from_expansion()
+        && let ExprKind::Call(callee, _) = call.kind
+        && fn_def_id(cx, call).is_some_and(|did| match_def_path(cx, did, &paths::STRING_FROM_UTF8))
+        && let Some(unwrap_call) = get_parent_expr(cx, call)
+        && let ExprKind::MethodCall(unwrap_method_name, ..) = unwrap_call.kind
+        && [sym::unwrap, sym::expect].contains(&unwrap_method_name.ident.name)
```
Jarcho, line 259: Could be `matches!(unwrap_method_name.ident.name, sym::unwrap | sym::expect)`

Jarcho (approved):
Looks fine. Just one small nit.

Jarcho:
Whoops. I thought I already merged this. @bors r+

PR #11661, opened 2023-10-12
`unusable_matches_binding` lint implementation

### LINKED ISSUE #7351: detect unusable bindings in arguments to `matches!`
I opened an issue about this on Rust itself, because it may be possible to fix over there at some point (maybe in an edition): https://github.com/rust-lang/rust/issues/86112

However, in the mean time, a clippy lint is probably more expedient.

### What it does

Warns if the pattern provided to the `matches!` macro (or possibly to nay macro) contains a variable binding which is not also used within the pattern (such as a guard check).

### Categories (optional)

- Kind: clippy::correctness

As explained in https://github.com/rust-lang/rust/issues/86112: 

> A co-worker of mine tried this code:
> 
> ```rust
> if !matches!(binding_to_uuid, binding_to_other_uuid) {
>     // ...
> }
> ```
> 
> I expected to see this happen:
> 
> Compiler should say that this is an obvious error, because there is no way to use the `binding_to_other_uuid` binding because it is internal to the match statement (and just shadows the actual `binding_to_other_uuid`).
> 
> Instead, this happened:
> 
> The `matches!` macro here expands to something that is never useful:
> 
> ```rust
> if !match binding_to_uuid {
>     binding_to_other_uuid => true,
>     _ => false,
> } { ... }
> ```
> 
> Clippy suggests underscoring the `binding_to_other_uuid` because it's unused, but that's not super clear unless you are familiar with the macro expansion. Ideally providing variable bindings that are not used in a guard would be disallowed in `matches!` because they cannot otherwise be used.

### Drawbacks

None. The binding is unusable within that scope if it is not used as part of a guard.

p-fuchs:
Hey,
This PR contains new lint implementation along with UI tests.
Fixes #7351.

changelog: [`unusable_matches_binding`]: Lint implementation and UI test

### DIFF
--- a/clippy_lints/src/matches/mod.rs
+++ b/clippy_lints/src/matches/mod.rs
@@ -974,7 +974,8 @@ declare_clippy_lint! {
     ///
     /// ### Why is this bad?
     /// It could be hard to determine why the compiler complains about an unused variable,
-    /// also it could introduce behavior that was not intened by the programmer.
+    /// which was introduced by the `matches!` macro expansion, also, it could introduce
+    /// behavior that was not intended by the programmer.
     ///
     /// ### Example
     /// ```rust
@@ -1055,7 +1056,7 @@ impl<'tcx> LateLintPass<'tcx> for Matches {
         if let ExprKind::Match(ex, arms, source) = expr.kind {
             if is_direct_expn_of(expr.span, "matches").is_some() {
                 redundant_pattern_match::check_match(cx, expr, ex, arms);
-                unusable_matches_binding::check_matches(cx, ex, arms, expr);
+                unusable_matches_binding::check_matches(cx, arms);
             }
 
             if source == MatchSource::Normal && !is_span_match(cx, expr.span) {
--- a/clippy_lints/src/matches/mod.rs
+++ b/clippy_lints/src/matches/mod.rs
@@ -22,6 +22,7 @@ mod rest_pat_in_fully_bound_struct;
 mod significant_drop_in_scrutinee;
 mod single_match;
 mod try_err;
+mod unusable_matches_binding;
 mod wild_in_or_pats;
 
 use clippy_utils::msrvs::{self, Msrv};
@@ -967,6 +968,38 @@ declare_clippy_lint! {
     "checks for unnecessary guards in match expressions"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for redundant and error prone bindings inside `matches!` macro.
+    ///
+    /// ### Why is this bad?
+    /// It could be hard to determine why the compiler complains about an unused variable,
+    /// also it could introduce behavior that was not intened by the programmer.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let data_source = Some(5);
+    /// let unrelated_data_source = 5;
+    ///
+    /// let x = matches!(data_source, unused_binding);
+    /// let y = matches!(data_source, used_binding if used_binding.is_some());
+    /// let z = matches!(data_source, unused_binding if unrelated_data_source > 6);
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let data_source = Some(5);
+    /// let unrelated_data_source = 5;
+    ///
+    /// let x = true;
+    /// let y = data_source.is_some();
+    /// let z = unrelated_data_source > 6;
+    /// ```
+    #[clippy::version = "1.75.0"]
+    pub UNUSABLE_MATCHES_BINDING,
+    correctness,
+    "default lint description"
+}
+
 pub struct Matches {
     msrv: Msrv,
     infallible_destructuring_match_linted: bool,
@@ -1009,6 +1042,7 @@ impl_lint_pass!(Matches => [
     MANUAL_MAP,
     MANUAL_FILTER,
     REDUNDANT_GUARDS,
+    UNUSABLE_MATCHES_BINDING,
 ]);
 
 impl<'tcx> LateLintPass<'tcx> for Matches {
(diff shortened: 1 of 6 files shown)


clippy_lints/src/matches/unusable_matches_binding.rs
```diff
@@ -0,0 +1,61 @@
...
+                        UNUSABLE_MATCHES_BINDING,
+                        arm.pat.span,
+                        &format!(
+                            "using identificator (`{}`) as matches! pattern argument without guard matches all cases",
+                            ident.name
+                        ),
+                        Some(expr.span),
+                        "matches! invocation always evaluates to `true`",
+                    );
+                }
```
dswij, line 57: Would probably be easier to read if we inverse the predicate and then return early for this block.

tests/ui/unusable_matches_binding.stderr
```diff
@@ -0,0 +1,42 @@
+error: using identificator (`unusable_binding`) as matches! pattern argument without guard matches all cases
```
dswij, line 1: Let's be brief for error messages. I think something like

```
error: binding `unusable_binding` is not used
```

Then we can explain more in detail in `note` or `help`

tests/ui/unusable_matches_binding.stderr
```diff
@@ -0,0 +1,42 @@
...
+note: matches! invocation always evaluates to `true`
+  --> $DIR/unusable_matches_binding.rs:18:13
+   |
+LL |     let _ = matches!(matching_data_source, unusable_binding);
+   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+   = note: `-D clippy::unusable-matches-binding` implied by `-D warnings`
+   = help: to override `-D warnings` add `#[allow(clippy::unusable_matches_binding)]`
+   = note: this error originates in the macro `matches` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: using identificator (`used_binding`) as matches! pattern argument matches all cases
```
dswij, line 16: The error message is pretty confusing for this one. The argument here won't match all cases.

The only special case is probably some mutating expression in `matches!` macro. But, I don't think this is a common case, and shouldn't be written that way anyway. We should only care about this when we are writing clippy suggestions.

Simplifying the logic to just check unused binding should be good enough for this lint.

clippy_lints/src/matches/mod.rs
```diff
@@ -967,6 +968,38 @@ declare_clippy_lint! {
     "checks for unnecessary guards in match expressions"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for redundant and error prone bindings inside `matches!` macro.
+    ///
+    /// ### Why is this bad?
+    /// It could be hard to determine why the compiler complains about an unused variable,
+    /// also it could introduce behavior that was not intened by the programmer.
```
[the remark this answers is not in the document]
dswij: Can we also add that this is bad because of `matches!` expansion?

dswij (commented):
Thanks for the PR! Overall, the implementation looks good, but I have some comments.

tests/ui/unusable_matches_binding.stderr
y21: The current message is a bit verbose, I agree, but in my opinion that suggested message isn't really much better because rustc already emits a warning here with an almost identical message (unused variable: `unusable_binding`), so I don't think that the suggested message would provide much more value than what rustc already says about this.

The problem imo is not so much that the binding isn't used, but rather that the user didn't mean to create a binding in the first place, and someone that doesn't know how `matches!` works likely won't understand why a binding is created either.
The user more likely meant to compare two values (I've seen this confusion happen a few times in forums, thinking that `matches!(1, x)` behaves like `1 == x`), so I think it'd be better if this message focused on the fact that it won't do that and that this `matches!` is useless/always evaluates to true.

How about something like
```
error: identifier pattern in `matches!` macro always evaluates to true

note: the identifier pattern matches any value and creates an unusable binding in the process
help: if you meant to compare two values, use `x == y` or `discriminant(x) == discriminant(y)`
```

dswij: > and someone that doesn't know how matches! works likely won't understand why a binding is created either.

Fair point

> rustc already emits a warning here with an almost identical message (unused variable: unusable_binding)

Regarding this, I wonder if we should have this lint in Clippy at all. Users would stumble into rustc's lint first before running clippy. Having this as rustc_lint might be more beneficial

Fuchczyk: I've changed the error message here.
Isn't this lint too "specific" to be included in the `rustc`?

tests/ui/unusable_matches_binding.stderr
Fuchczyk: After the PR change we need also to check the case when the guard is present because in that situation different error message should be displayed (compared to the case when the guard is missing).

tests/ui/unusable_matches_binding.stderr
y21: I agree that it's a bit "unfortunate" that rustc's unused_variables lint triggers here and prevents us from emitting better diagnostics. It could do a much better job at explaining what the issue is and how to fix it. Might make sense to specialize the behavior of the `unused_variables` lint, if it sees that the parent node of the identifier pattern is a `matches!` expansion?

OTOH, the lint is in the correctness category (deny by default), so I think that will stand out more than rustc's lint? I'm not sure.

dswij: I guess it's better to have something than nothing. Let's see how this lint fares, maybe if it's good enough we can have it uplifted to rustc lint

dswij:
@y21 Would you mind helping to review the code and run lintcheck for this? :)

y21 (commented):
Some initial thoughts :)

y21:
> @y21 Would you mind helping to review the code and run lintcheck for this? :)

Re. lintcheck, it looks like there aren't any matches (no pun intended), at least not for the 100 most downloaded crates. I'm not surprised though, considering the `unused_variables` rustc lint should've sort of warned them

GuillaumeGomez:
This is actually already triggered by rustc's `unused_variable` so can be closed.

dswij:
Thank you for the PR @p-fuchs , it's really appreciated. But since a rustc lint already checks for this, we shouldn't add another redundant one.

smoelius:
Should #7351 be closed as well?

GuillaumeGomez:
Yes it should.

Issue #11662, opened 2023-10-12
Incosistency between `else_if_without_else` and `needless_else`


c410-f3r:
`else_if_without_else` is allowed in `restriction` while `needless_else` warns in `style` although both deal with the same pattern. It is then probably a good move to put both in the same category.

y21:
I personally disagree. They deal with the opposite pattern rather, and one is (in my experience) more idiomatic than the other.
Requiring an `else` on every single `else if` even when it's not needed sounds quite restrictive. An empty `else` OTOH seems much more sane to warn and is likely a mistake, possibly leftover from refactoring.

The same sort of argument can be made for `implicit_return` (restriction) and `needless_return` (style). Omitting `return;` as the last statement of a function is idiomatic Rust so it seems good to have `needless_return` be warn by default.

Alexendoo:
Yeah having both of these warn by default would create a cycle where we repeatedly ask the user to add an `else {}` then on the next run ask them to remove it

c410-f3r:
Looks like this is a desired undocumented behaviour, or is there any?

For those interested in consistency, `semicolon_inside_block` and `semicolon_inside_block` among possibly others, are in the same category with the same level of restriction.

Alexendoo:
There are several pairs of lints that suggest opposites to each other, but the only case where they should be in the same category is if that category is `restriction`, meaning they're both considered individually opt in

c410-f3r:
@Alexendoo https://github.com/rust-lang/rust-clippy/pull/11666

PR #11666, opened 2023-10-13
Document conflicting lints

c410-f3r:
Documents the outcome described in https://github.com/rust-lang/rust-clippy/issues/11662 so that the situation can be avoided in the future by other individuals.

changelog: none

### DIFF
--- a/book/src/development/adding_lints.md
+++ b/book/src/development/adding_lints.md
@@ -627,6 +627,10 @@ classification:
 * For all the other cases, they should be in different categories with different levels of allowance. For example,
 `implicit_return` (restriction, allow) and `needless_return` (style, warn).
 
+For lints that are in different categories, it is also recommended that at least one of them should be in the
+`restriction` category. The reason for this is that the `restriction` group is the only group where we don't
+recommend to enable the entire set, but cherry pick lints out of.
+
 ## PR Checklist
 
 Before submitting your PR make sure you followed all the basic requirements:
--- a/book/src/development/adding_lints.md
+++ b/book/src/development/adding_lints.md
@@ -30,6 +30,7 @@ because that's clearly a non-descriptive name.
   - [Documentation](#documentation)
   - [Running rustfmt](#running-rustfmt)
   - [Debugging](#debugging)
+  - [Conflicting lints](#conflicting-lints)
   - [PR Checklist](#pr-checklist)
   - [Adding configuration to a lint](#adding-configuration-to-a-lint)
   - [Cheat Sheet](#cheat-sheet)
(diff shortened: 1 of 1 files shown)


book/src/development/adding_lints.md
```diff
@@ -612,6 +613,16 @@ output in the `stdout` part.
 
 [`dbg!`]: https://doc.rust-lang.org/std/macro.dbg.html
 
+## Opposing lints
+
+There are several lints that deal with the same pattern but suggest different approaches. When you are creating a
+lint that interact with this scenario, the following tips should be encouraged to guide classification:
+
+* The only case where they should be in the same category is if that category is `restriction`. For example,
+`semicolon_inside_block` and `semicolon_inside_block`.
```
flip1995, line 622: Typo: same lint name 2 times

book/src/development/adding_lints.md
```diff
@@ -612,6 +613,16 @@ output in the `stdout` part.
 
 [`dbg!`]: https://doc.rust-lang.org/std/macro.dbg.html
 
+## Opposing lints
```
flip1995, line 616: I would call this `Conflicting Lints` and rephrase the paragraph below a little, that a new lint might conflict with an existing lint, as it suggests something an existing lint would complain about and if that's the case, the 2 rules below should be applied

flip1995 (changes requested):
Good idea documenting this!

book/src/development/adding_lints.md
```diff
@@ -612,6 +613,20 @@ output in the `stdout` part.
...
+There are several lints that deal with the same pattern but suggest different approaches. In other words, some lints
+may suggest modifications that go in the opposite direction to what some other lints already propose for the same
+code, creating conflicting diagnostics.
+
+When you are creating a lint that ends up in this scenario, the following tips should be encouraged to guide
+classification:
+
+* The only case where they should be in the same category is if that category is `restriction`. For example,
+`semicolon_inside_block` and `semicolon_outside_block`.
+* For all the other cases, they should be in different categories with different levels of allowance. For example,
```
flip1995, line 627: I would even go as far as saying that one of those lints should be in `restriction`. The reason for this is, that the restriction group is the only group where we don't allow/recommend to enable the entire group, but cherry pick lints out of.

Enabling the pedantic group and then getting conflicting lints also doesn't sound like a good UX to me.

c410-f3r: Done

flip1995 (approved):

flip1995:
@bors r+

Thanks!

PR #11669, opened 2023-10-14
new lint: `unnecessary_fallible_conversions`

### LINKED ISSUE #11577: new lint: `unnecessary_faillible_conversion`
### What it does

This lint would warn against using `try_into` from the `TryInto` or `TryFrom` trait when there is an available `Into` or `From` implementation.

### Advantage

Remove the need for handling an error case that can never happen (if there is a `From` impl, the `Error` case of the `TryFrom` impl is `Infallible`.

### Drawbacks

_No response_

### Example

```rust
let a: u32 = 32;
let b: Result<u64, Infaillible> = a.try_into();
```

Could be written as:

```rust
let a: u32 = 32;
let b: u64 = a.into();
```

y21:
Closes #11577

A new lint that looks for calls such as `i64::try_from(1i32)` and suggests `i64::from(1i32)`. See lint description (and linked issue) for more details for why.

There's a tiny bit of overlap with the `useless_conversion` lint, in that the other one warns `T::try_from(T)` (i.e., fallibly converting to the same type), so this lint ignores cases like `i32::try_from(1i32)` to avoid emitting two warnings for the same expression.

Also, funnily enough, with this one exception, this lint would warn on exactly every case in the `useless_conversion_try` ui test that `useless_conversion` didn't cover (but never two warnings at the same time), which is neat. I did add an `#![allow]` though since we don't want interleaved warnings from multiple lints in the same uitest.

changelog: new lint: `unnecessary_fallible_conversions`

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -99,6 +99,7 @@ mod suspicious_to_owned;
 mod type_id_on_box;
 mod uninit_assumed_init;
 mod unit_hash;
+mod unnecessary_fallible_conversions;
 mod unnecessary_filter_map;
 mod unnecessary_fold;
 mod unnecessary_iter_cloned;
@@ -3655,6 +3656,33 @@ declare_clippy_lint! {
     "cloning a `Waker` only to wake it"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to `TryInto::try_into` and `TryFrom::try_from` when their infallible counterparts
+    /// could be used.
+    ///
+    /// ### Why is this bad?
+    /// In those cases, the `TryInto` and `TryFrom` trait implementation is a blanket impl that forwards
+    /// to `Into` or `From`, which always succeeds.
+    /// The returned `Result<_, Infallible>` requires error handling to get the contained value
+    /// even though the conversion can never fail.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let _: Result<i64, _> = 1i32.try_into();
+    /// let _: Result<i64, _> = <_>::try_from(1i32);
+    /// ```
+    /// Use `from`/`into` instead:
+    /// ```rust
+    /// let _: i64 = 1i32.into();
+    /// let _: i64 = <_>::from(1i32);
+    /// ```
+    #[clippy::version = "1.75.0"]
+    pub UNNECESSARY_FALLIBLE_CONVERSIONS,
+    style,
+    "calling the `try_from` and `try_into` trait methods when `From`/`Into` is implemented"
+}
+
 pub struct Methods {
     avoid_breaking_exported_api: bool,
     msrv: Msrv,
@@ -3801,6 +3829,7 @@ impl_lint_pass!(Methods => [
     PATH_ENDS_WITH_EXT,
     REDUNDANT_AS_STR,
     WAKER_CLONE_WAKE,
+    UNNECESSARY_FALLIBLE_CONVERSIONS,
 ]);
 
 /// Extracts a method call name, args, and `Span` of the method name.
@@ -3827,6 +3856,7 @@ impl<'tcx> LateLintPass<'tcx> for Methods {
         match expr.kind {
             hir::ExprKind::Call(func, args) => {
                 from_iter_instead_of_collect::check(cx, expr, args, func);
+                unnecessary_fallible_conversions::check_function(cx, expr, func);
             },
             hir::ExprKind::MethodCall(method_call, receiver, args, _) => {
                 let method_span = method_call.ident.span;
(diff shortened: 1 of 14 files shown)


clippy_lints/src/methods/unnecessary_fallible_conversions.rs
```diff
@@ -0,0 +1,106 @@
...
+            FunctionKind::TryFromFunction => sym::From,
+            FunctionKind::TryIntoMethod | FunctionKind::TryIntoFunction => sym::Into,
+        })
+        // If `T: TryFrom<U>` and `T: From<U>` both exist, then that means that the `TryFrom`
+        // _must_ be from the blanket impl and cannot have been manually implemented
+        // (else there would be conflicting impls, even with #![feature(spec)]), so we don't even need to check
+        // what `<T as TryFrom<U>>::Error` is.
+        && implements_trait(cx, self_ty, from_into_trait, &[other_ty])
+        // Avoid linting if this is an argument to a function: `f(i32.try_into())`.
+        // The function expects a `Result` either way, so users would need to wrap it back in `Ok(..)`
```
sgued, line 46: Is this motivated by a specific example? I don't see when any function taking a `Result<T, Infaillible>` would make sense.

clippy_lints/src/methods/unnecessary_fallible_conversions.rs
```diff
@@ -0,0 +1,106 @@
...
+                | FunctionKind::TryIntoFunction => "calling `TryInto::try_into` when `Into` could be used",
+            },
+            "use",
+            match kind {
+                FunctionKind::TryFromFunction => "From::from".into(),
+                FunctionKind::TryIntoFunction => "Into::into".into(),
+                FunctionKind::TryIntoMethod => "into".into(),
+            },
+            // Suggestion likely results in compile errors, so it needs to be adjusted a bit by the user
+            // (e.g. removing `unwrap()`s as a result, changing type annotations, ...)
```
sgued, line 66: Maybe the lint could detect specifically `.try_into().unwrap()` and `.try_into().expect()` and lint those with `Applicability::MachineApplicable`.

sgued (commented):
Thanks for the implementation of my suggested lint!

clippy_lints/src/methods/unnecessary_fallible_conversions.rs
y21: Hm yeah, I think you're right. The user may have had some kind of helper function called from multiple places that handles a result in the same way. But that doesn't really make a lot of sense to call with an infallible Result in the first place

clippy_lints/src/methods/mod.rs
```diff
@@ -3632,6 +3633,36 @@ declare_clippy_lint! {
...
+    /// - Calling `<T as TryInto<U>>::try_into` where `T` also implements `Into<U>`
+    /// - Calling `<T as TryFrom<U>>::try_from` where `T` also implements `From<U>`
+    ///
+    /// ### Why is this bad?
+    /// In those cases, the `TryInto` and `TryFrom` trait implementation is a blanket impl which forwards
+    /// to `Into` or `From`, which always succeeds.
+    /// Indeed, its associated `Error` type is the empty enum `Infallible`,
+    /// which is impossible to construct by nature, so the error handling that is then imposed by
+    /// the `Result<_, Infallible>` is unnecessary and can be avoided simply by calling `.into()` or `::from`
+    /// directly in the first place.
```
Jarcho, line 3648: This also has more detail than is necessary to get the point across. The key point being the need to unwrap the result even though it's always going to be the `Ok` variant.

clippy_lints/src/methods/unnecessary_fallible_conversions.rs
```diff
@@ -0,0 +1,123 @@
...
+
+        span_lint_and_sugg(
+            cx,
+            UNNECESSARY_FALLIBLE_CONVERSIONS,
+            span,
+            match kind {
+                FunctionKind::TryFromFunction => "calling `TryFrom::try_from` when `From` could be used",
+                FunctionKind::TryIntoMethod
+                | FunctionKind::TryIntoFunction => "calling `TryInto::try_into` when `Into` could be used",
+            },
```
Jarcho, line 81: A more general message such as `use of a fallible conversion when an infallible one could be used`. The highlighted span and the suggestion already have the specifics.

Jarcho (commented):
Looks good overall. Just a couple of small nits. If you don't feel like rewording things it's workable the way it is.

clippy_lints/src/methods/unnecessary_fallible_conversions.rs
y21: Sounds good

clippy_lints/src/methods/mod.rs
y21: Changed the documentation a bit. Is this better?

Jarcho:
Thanks for the small cleanup. @bors r+

alexanderkjeldaas:
Just a maybe unrelated comment on this, but would a lint that finds unnecessary TryFrom implementations that instead could be just From be implemented alongside this one?

```
impl TryFrom<X> for Y {
    type Error = <Foo as FromStr>::Err;
    fn try_from(x: X) -> Result<Self, Self::Error> {
        Ok(Self {
            height: x.height.0.as_u64(),
            hash: x.hash.0,
        })
    }
}
```

Easy to detect when the return value is always `Ok`?

flip1995:
Looks like a possible lint. If you want this, please open an issue or if you want to get into Clippy development consider opening a PR for it :)

PR #11679, opened 2023-10-16
Fix failure of new internal lint `rustc::span_use_eq_ctxt`

a-lafrance:
I've implemented a new Rustc internal lint that forbids usage of `.ctxt() == .ctxt()` for spans, and suggests using `Span::eq_ctxt` directly instead (see https://github.com/rust-lang/rust/issues/49509 and https://github.com/rust-lang/rust/pull/116787).

Unfortunately the PR can't land yet because there's a violation of the lint in the Clippy codebase which (as it should) is causing CI failures, so this PR fixes that lint preemptively to unblock it from landing.

changelog: none

### DIFF
--- a/clippy_utils/src/macros.rs
+++ b/clippy_utils/src/macros.rs
@@ -245,7 +245,7 @@ impl<'a> PanicExpn<'a> {
             return None;
         };
         let result = match name {
-            "panic" if arg.span.ctxt() == expr.span.ctxt() => Self::Empty,
+            "panic" if arg.span.eq_ctxt(expr.span) => Self::Empty,
             "panic" | "panic_str" => Self::Str(arg),
             "panic_display" | "panic_cold_display" => {
                 let ExprKind::AddrOf(_, _, e) = &arg.kind else {


flip1995:
You can just include those changes in your PR towards `rust-lang/rust`. Otherwise you'll have to wait until Thursday, when I sync this PR to the Rust repo.

r? @flip1995 

So the changes LGTM, but I'm holding off with merging, as I think that it's easier for everyone if you just apply those changes in the Rust repo (`src/tools/clippy`)

a-lafrance:
Ah whoops, thanks for the heads up. Didn't know I could update Clippy in-tree, that does seem much smoother.

Closing this now that I've updated the Rustc PR with the Clippy changes.

flip1995:
> Didn't know I could update Clippy in-tree,

Yes, this is encouraged for fixing up breaking changes in rustc. But functional changes like lint additions are not allowed in rustc and must be done in the Clippy repo.

a-lafrance:
Good to know, thanks for the info :)

PR #11683, opened 2023-10-18
Deserialize `Msrv` directly in `Conf`

Alexendoo:
Gives the error a span pointing to the invalid config value

Also puts `Conf` itself in the `OnceLock` rather than just the `Msrv` for [the `register_late_mod_pass` work](https://github.com/rust-lang/rust/pull/116731) since it will be used from two different callbacks

changelog: none

### DIFF
--- a/book/src/lint_configuration.md
+++ b/book/src/lint_configuration.md
@@ -100,7 +100,7 @@ Suppress lints whenever the suggested change would cause breakage for other crat
 ## `msrv`
 The minimum rust version that the project supports
 
-**Default Value:** `None` (`Option<String>`)
+**Default Value:** `Msrv { stack: [] }` (`crate::Msrv`)
 
 ---
 **Affected lints:**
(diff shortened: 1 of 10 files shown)


Manishearth (approved):

book/src/lint_configuration.md
```diff
@@ -100,7 +100,7 @@ Suppress lints whenever the suggested change would cause breakage for other crat
 ## `msrv`
 The minimum rust version that the project supports
 
-**Default Value:** `None` (`Option<String>`)
+**Default Value:** `Msrv { stack: [] }` (`crate::Msrv`)
```
flip1995, line 103: Huh, why is the default value of MSRV an empty stack? 🤔

Alexendoo: Ah damn I didn't think about that, I think we need to change that in general since other config values also show internal type names which isn't helpful for public documentation

flip1995: Right, I don't think this has to be done in this PR though. So feel free to r=me here now. But if you feel like doing it in this PR, I won't stop you :upside_down_face:

Alexendoo:
@bors r=Manishearth,flip1995

I'll think about the config thing separately, I think it would need website changes as well

PR #11685, opened 2023-10-19
Create `clippy_config` crate, hide internal representation from docs

Alexendoo:
r? @flip1995 

The first commit moves a decent chunk of code out of `clippy_lints`/`clippy_utils` into a new crate `clippy_config`

The second commit changes how `--explain`, the book and the site render configuration values. The internal type is now hidden and the displayed defaults are now valid TOML values instead of `Debug` output

changelog: none


tests/dogfood.rs
flip1995, line 1: You also have to add the new `clippy_config` great to the `dogfood_clippy` test function above.

clippy_lints/src/utils/format_args_collector.rs
```diff
@@ -5,27 +5,20 @@ use rustc_ast::{Crate, Expr, ExprKind, FormatArgs};
...
 use rustc_span::{hygiene, Span};
 use std::iter::once;
 use std::mem;
 use std::rc::Rc;
 
-declare_clippy_lint! {
-    /// ### What it does
-    /// Collects [`rustc_ast::FormatArgs`] so that future late passes can call
-    /// [`clippy_utils::macros::find_format_args`]
-    pub FORMAT_ARGS_COLLECTOR,
```
flip1995, line 18: Why could this be removed?

flip1995 (commented):
Mostly NITs.

clippy_lints/src/utils/format_args_collector.rs
Alexendoo: The lint is never actually used, and a lint pass doesn't need to have associated lints to be registered & run for side effects

I removed it because it was producing an error for the metadata collector which was previously ignored:

```
error: metadata collection error for `FORMAT_ARGS_COLLECTOR`: Unable to find lint in lint_store
  --> src/utils/format_args_collector.rs:19:9
   |
19 |     pub FORMAT_ARGS_COLLECTOR,
   |         ^^^^^^^^^^^^^^^^^^^^^
```

flip1995:
@bors r+

Thanks!

Alexendoo:
Needed an `extern crate rustc_driver` like the other crates

@bors r=flip1995

PR #11687, opened 2023-10-20
explicit_reinitialization: Cutoff reinit

lengyijun:
changelog: [`explicit_reinitialization`]: new lint: introduce  fresh variable instead of reinitialization

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5013,6 +5013,7 @@ Released 2018-09-13
 [`explicit_deref_methods`]: https://rust-lang.github.io/rust-clippy/master/index.html#explicit_deref_methods
 [`explicit_into_iter_loop`]: https://rust-lang.github.io/rust-clippy/master/index.html#explicit_into_iter_loop
 [`explicit_iter_loop`]: https://rust-lang.github.io/rust-clippy/master/index.html#explicit_iter_loop
+[`explicit_reinitialization`]: https://rust-lang.github.io/rust-clippy/master/index.html#explicit_reinitialization
 [`explicit_write`]: https://rust-lang.github.io/rust-clippy/master/index.html#explicit_write
 [`extend_from_slice`]: https://rust-lang.github.io/rust-clippy/master/index.html#extend_from_slice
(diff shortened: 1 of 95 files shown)


lengyijun:
Algorithm summary:
1. in hir, filter suspious hir stmt
2. locate the related mir
3. in mir cfg, check the assignment dominate all reachable usages

Alexendoo:
I'm not sure about this, it can change the behaviour of code with regards to drop order for example

```rust
let mut x = some_vec();
x = some_other_vec(); // x is dropped here
// ...
```
```rust
let x = some_vec();
let x = some_other_vec(); // x is no longer dropped here, but at the end of the scope
// ...
```

For a `Vec` this could increase the peak memory usage, with similar issues for other kinds of resources

I also don't think we should be suggesting to shadow one `mut` binding with another, that makes things more confusing I would say

lengyijun:
Q&A
1. why use mir / why not analysis purely in hir ?
No dominance tree in hir

xFrednet:
Hey, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

I'm guessing this PR stalled in part due to the usage of MIR. I currently use MIR for a different project and am not sure if it's the right choice for Clippy. As you say there is no one-to-one relation between MIR and HIR stuff which makes linting hard. There is also the problem that most of the Clippy team are not too familiar with MIR. 

If you're planning to continue this, I would suggest adding the `I-nominate` label, to nominate the discussion if we want to use MIR for lints for the next Clippy meeting. 

@rustbot author

flip1995:
As this lint is close to the `shadow` lints (just the reverse) it probably should go into a allow-by-default group as I feel like this is pretty opinionated. For MIR lints we have a  "should be worth the complexity" rule (discussed in the [last meeting](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy/topic/Meeting.202024-04-02)). For an allow-by-default lint, I don't think this is worth the complexity. For this reason and because there was no response by the author, I'm closing this PR, without a `S-inactive-closed` label.

lengyijun:
https://github.com/lengyijun/explicit_reinitialization

PR #11691, opened 2023-10-20
Lint `flatten()` under `lines_filter_map_ok`

### LINKED ISSUE #11686: lines_filter_map_ok should also warn for flatten
### Summary

When stumbling about the clippy warning `lines_filter_map_ok` , I tried a different solution by using `flatten()`. The clippy warning disappears

I tested the code with a directory (as described in the lint) and it still loops forever.

### Lint Name

lines_filter_map_ok

### Reproducer

I tried this code snippet:

```rust
let f = File::open(".")?; // open a directory
let f = BufReader::new(f);

let numbered_lines_iter = f
    .lines()
    .flatten() // <- has same behavior as .flat_map(Result::ok)
    .enumerate();

let first_line = numbered_lines_iter.next(); // this loops forever
```

Since flatten() is more or less the same as flat_map, it should warn as well

Instead, this happened:
No warnings

### Version

```text
rustc 1.75.0-nightly (0039d739d 2023-10-18)
binary: rustc
commit-hash: 0039d739d40a076334e111488946441378d11cd7
commit-date: 2023-10-18
release: 1.75.0-nightly
LLVM version: 17.0.3
```

sjwang05:
Fixes #11686 

changelog: [`lines_filter_map_ok`]: Also lint calls to `flatten()`

### DIFF
--- a/clippy_lints/src/lines_filter_map_ok.rs
+++ b/clippy_lints/src/lines_filter_map_ok.rs
@@ -53,40 +53,41 @@ declare_clippy_lint! {
     #[clippy::version = "1.70.0"]
     pub LINES_FILTER_MAP_OK,
     suspicious,
-    "filtering `std::io::Lines` with `filter_map()` or `flat_map()` might cause an infinite loop"
+    "filtering `std::io::Lines` with `filter_map()`, `flat_map()`, or `flatten()` might cause an infinite loop"
 }
 declare_lint_pass!(LinesFilterMapOk => [LINES_FILTER_MAP_OK]);
 
 impl LateLintPass<'_> for LinesFilterMapOk {
     fn check_expr(&mut self, cx: &LateContext<'_>, expr: &Expr<'_>) {
-        if let ExprKind::MethodCall(fm_method, fm_receiver, [fm_arg], fm_span) = expr.kind
-            && is_trait_method(cx, expr, sym::Iterator)
-            && (fm_method.ident.as_str() == "filter_map" || fm_method.ident.as_str() == "flat_map")
-            && is_type_diagnostic_item(cx, cx.typeck_results().expr_ty_adjusted(fm_receiver), sym::IoLines)
+        if let ExprKind::MethodCall(fm_method, fm_receiver, fm_args, fm_span) = expr.kind &&
+            is_trait_method(cx, expr, sym::Iterator) &&
+            let fm_method_str = fm_method.ident.as_str() &&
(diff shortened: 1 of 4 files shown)


clippy_lints/src/lines_filter_map_ok.rs
```diff
@@ -53,35 +53,41 @@ declare_clippy_lint! {
...
-                        let ExprKind::MethodCall(method, receiver, [], _) = value.kind &&
-                        path_to_local_id(receiver, param.pat.hir_id) &&
-                        let Some(method_did) = cx.typeck_results().type_dependent_def_id(value.hir_id)
-                    {
-                        is_diag_item_method(cx, method_did, sym::Result) && method.ident.as_str() == "ok"
-                    } else {
-                        false
-                    }
-                _ => false,
+            let lint = if let [fm_arg] = fm_args {
```
Centri3, line 68: Can we match on this instead? Like, `[fm_arg] => {}` then `[] => fm_method_str == "flatten"` for the `flatten` case.

Also, regardless, the `flatten` case should be first imo as it's only a single line, for readability's sake.

Centri3: Not exactly related to this PR but imo this should be a separate function

Centri3 (approved):
Only 2 nits, thank you!

clippy_lints/src/lines_filter_map_ok.rs
sjwang05: Thanks for the feedback! I've converted the `if let` into a `match` and split it into its own `should_lint` function as suggested.

clippy_lints/src/lines_filter_map_ok.rs
```diff
@@ -53,18 +53,48 @@ declare_clippy_lint! {
...
         {
-            let lint = match &fm_arg.kind {
+            span_lint_and_then(
+                cx,
+                LINES_FILTER_MAP_OK,
+                fm_span,
+                &format!(
+                    "`{}()` will run forever if the iterator repeatedly produces an `Err`",
+                    fm_method.ident
+                ),
```
Centri3, line 76: One more nit:
```suggestion
                &format!(
                    "`{fm_method_str}()` will run forever if the iterator repeatedly produces an `Err`",
                ),
```
This prevents calling `as_str` again which requires locking the symbol interner

Centri3:
Thanks, @bors r+

Issue #11692, opened 2023-10-21
`clippy::iter_without_into_iter` suggestion does not compile


lopopolo:
### Summary

The suggested implementation of `IntoIterator` does not compile.

### Reproducer

I tried this code:

```rust
#![warn(clippy::all)]
#![warn(clippy::pedantic)]

pub struct Foo;

pub struct FooIter<'a>(Option<&'a Foo>);

impl<'a> Iterator for FooIter<'a> {
    type Item = &'a Foo;
    
    fn next(&mut self) -> Option<Self::Item> {
        self.0.take()
    }
}

impl Foo {
    #[must_use]
    pub fn iter(&self) -> FooIter<'_> {
        FooIter(Some(self))
    }
}


fn main() {
    println!("Hello, world!");
}
```

I expected to see this happen: clippy emits a `clippy::iter_without_into_iter` and gives a suggested fix that compiles.

Instead, this happened:

```
    Checking playground v0.0.1 (/playground)
warning: `iter` method without an `IntoIterator` impl for `&Foo`
  --> src/main.rs:18:5
   |
18 | /     pub fn iter(&self) -> FooIter<'_> {
19 | |         FooIter(Some(self))
20 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#iter_without_into_iter
note: the lint level is defined here
  --> src/main.rs:2:9
   |
2  | #![warn(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::iter_without_into_iter)]` implied by `#[warn(clippy::pedantic)]`
help: consider implementing `IntoIterator` for `&Foo`
   |
16 + 
17 + impl IntoIterator for &Foo {
18 +     type IntoIter = FooIter<'_>;
19 +     type Iter = &Foo;
20 +     fn into_iter() -> Self::IntoIter {
21 +         self.iter()
22 +     }
23 + }
   |

warning: `playground` (bin "playground") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
```

Adding the suggestion results in this code:

```rust
#![warn(clippy::all)]
#![warn(clippy::pedantic)]

pub struct Foo;

pub struct FooIter<'a>(Option<&'a Foo>);

impl<'a> Iterator for FooIter<'a> {
    type Item = &'a Foo;
    
    fn next(&mut self) -> Option<Self::Item> {
[...]

PR #11696, opened 2023-10-22
[`iter_without_into_iter`]: fix papercuts in suggestion and restrict linting to exported types

y21:
See #11692 for more context.

tldr: the lint `iter_without_into_iter` has suggestions that don't compile, which imo isn't that problematic because it does have the appropriate `Applicability` that tells external tools that it shouldn't be auto-applied.
However there were some obvious "errors" in the suggestion that really should've been included in my initial PR adding the lint, which is fixed by this PR:
- `IntoIterator::into_iter` needs a `self` argument.
- `IntoIterator::Iter` associated type doesn't exist. This should've just been `Item`.

This still doesn't make it machine applicable, and the remaining things are imho quite non-trivial to implement, as I've explained in https://github.com/rust-lang/rust-clippy/issues/11692#issuecomment-1773886111.
I personally think it's fine to leave it there and let the user change the remaining errors when copy-pasting the suggestion (e.g. errors caused by lifetimes that were permitted in fn return-position but are not in associated types).
This is how many of our other lint suggestions already work.

Also, we now restrict linting to only exported types. This required moving basically all of the tests around since they were previously in the `main` function. Same for `into_iter_without_iter`. The git diff is a bit useless here...

changelog: [`iter_without_into_iter`]: fix papercuts in suggestion and restrict linting to exported types

(cc @lopopolo, figured I should mention you since you created the issue)

### DIFF
--- a/clippy_lints/src/iter_without_into_iter.rs
+++ b/clippy_lints/src/iter_without_into_iter.rs
@@ -19,6 +19,11 @@ declare_clippy_lint! {
     /// It's not bad, but having them is idiomatic and allows the type to be used in for loops directly
     /// (`for val in &iter {}`), without having to first call `iter()` or `iter_mut()`.
     ///
+    /// ### Limitations
+    /// This lint is restricted to exported types only, because it is aimed at guiding towards an
+    /// idiomatic, _public_ API.
+    /// Implementing the `IntoIterator` trait when it is not needed or used anywhere doesn't help or improve the code.
+    ///
     /// ### Example
     /// ```rust
     /// struct MySlice<'a>(&'a [u8]);
@@ -60,6 +65,12 @@ declare_clippy_lint! {
     /// It's not bad, but having them is idiomatic and allows the type to be used in iterator chains
     /// by just calling `.iter()`, instead of the more awkward `<&Type>::into_iter` or `(&val).into_iter()` syntax
     /// in case of ambiguity with another `IntoIterator` impl.
+    ///
+    /// ### Limitations
+    /// This lint is restricted to exported types only, because it is aimed at guiding towards an
+    /// idiomatic, _public_ API.
+    /// Adding an `iter` or `iter_mut` for private types when it is not needed or used doesn't improve code,
+    /// and in fact, is linted against by the `dead_code` lint.
     ///
     /// ### Example
     /// ```rust
--- a/clippy_lints/src/iter_without_into_iter.rs
+++ b/clippy_lints/src/iter_without_into_iter.rs
@@ -20,9 +20,9 @@ declare_clippy_lint! {
     /// (`for val in &iter {}`), without having to first call `iter()` or `iter_mut()`.
     ///
     /// ### Limitations
-    /// This lint is restricted to exported types only, because it is aimed at guiding towards an
-    /// idiomatic, _public_ API.
-    /// Implementing the `IntoIterator` trait when it is not needed or used anywhere doesn't help or improve the code.
+    /// This lint focuses on providing an idiomatic API. Therefore, it will only
+    /// lint on types which are accessible outside of the crate. For internal types,
+    /// the `IntoIterator` trait can be implemented on demand if it is actually needed.
     ///
     /// ### Example
     /// ```rust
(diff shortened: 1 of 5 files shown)


clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -104,6 +104,12 @@ fn is_nameable_in_impl_trait(ty: &rustc_hir::Ty<'_>) -> bool {
     !matches!(ty.kind, TyKind::OpaqueDef(..))
 }
 
+fn is_ty_exported(cx: &LateContext<'_>, ty: Ty<'_>) -> bool {
```
xFrednet, line 107: I think it would be good to document this restriction to public items in the lint description. Maybe under a `### Limitations` section to just under the normal `### What it does?` section.

xFrednet (approved):
LGTM, I have one small suggestion, and then you can `r=me`.

Clippy's documentation says that this lint will be stabilized in rust 1.74. Do you think this is worth a beta backport? It's a pedantic lint, which doesn't make it super necessary, but the restriction to only exported items, might make a big difference, to the amount of lint triggers :thinking:

clippy_lints/src/iter_without_into_iter.rs
y21: Good idea. In my experience, "Limitations" is under "Why is this bad?". I'll put that there instead, for consistency's sake.

y21: Done. What do you think about the wording? Not gonna approve just yet because I'd like to hear your opinion

y21:
> Do you think this is worth a beta backport?

I think it's a good idea, even for a pedantic lint. It's not all that useful for private types imo and I think it could make a noticeable difference in how often the lint occurs. The errors in the suggestion were also pretty obvious so it would be nice if this fix gets into beta along with the PR that introduced this.
But if it's too much of a hassle or this is the only PR that is gonna be nominated for backporting, it might be unnecessary.

lopopolo:
Please backport this. Paper cuts in clippy lints lead to them being disabled in my codebases. I almost never revisit an `!#[allow(clippy::xxx)]` once it gets added

xFrednet:
I agree and this shouldn't be a problem, that's why I asked. `@flip1995` Usually does the beta back ports, a week or so before the release. To indicate that a fix should be back ported, we just need to add the `beta-nominated` label.

@rustbot label +beta-nominated

clippy_lints/src/iter_without_into_iter.rs
y21: @xFrednet ? :)
(I'm not sure if you saw the comment or if you're just busy)

xFrednet: I actually missed your comment, thank you for pointing that out. It's correct, that the `### Limitations` section comes after `### Why is it bad?`

clippy_lints/src/iter_without_into_iter.rs
```diff
@@ -61,6 +66,12 @@ declare_clippy_lint! {
     /// by just calling `.iter()`, instead of the more awkward `<&Type>::into_iter` or `(&val).into_iter()` syntax
     /// in case of ambiguity with another `IntoIterator` impl.
     ///
+    /// ### Limitations
+    /// This lint is restricted to exported types only, because it is aimed at guiding towards an
+    /// idiomatic, _public_ API.
+    /// Adding an `iter` or `iter_mut` for private types when it is not needed or used doesn't improve code,
+    /// and in fact, is linted against by the `dead_code` lint.
+    ///
```
xFrednet, line 74: The current version is fine. I would formulate it a bit differently, but this is mostly a question of taste. I leave the decision up to you, which one you like more, or if you want to combine the two versions :)

```
    /// ### Limitations
    /// This lint focusess on providing an idiomatic API. Therefore, it will only
    /// lint on types, which are accessible outside of the crate. For internal types
    /// these methods can be added on demand, if they are acually needed. Otherwise,
    /// it would trigger the [`dead_code`] lint for the unused method.
    ///
    /// [`dead_code`]: https://doc.rust-lang.org/rustc/lints/listing/warn-by-default.html#dead-code
```

y21: Adjusted the documentation a bit with your suggestions in mind.

xFrednet (approved):

xFrednet:
LGTM, thank you for the update :)

@bors r+

flip1995:
This lint does not exist in `beta` yet and this PR was already synced to rustc. So this PR will be included when this lint gets to `beta` and ultimately `stable`. => Nothing to do for me. My favorite kind of backports :upside_down_face:

PR #11698, opened 2023-10-23
Add `waker_clone_and_wake` lint to check needless `Waker` clones

a1phyr:
Check for patterns of `waker.clone().wake()` and replace them with `waker.wake_by_ref()`.

An alternative name could be `waker_clone_then_wake`

changelog: [ `waker_clone_wake`]: new lint

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -112,6 +112,7 @@ mod useless_asref;
 mod utils;
 mod vec_resize_to_zero;
 mod verbose_file_reads;
+mod waker_clone_wake;
 mod wrong_self_convention;
 mod zst_offset;
 
@@ -3632,6 +3633,28 @@ declare_clippy_lint! {
     "`as_str` used to call a method on `str` that is also available on `String`"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `waker.clone().wake()`
+    ///
+    /// ### Why is this bad?
+    /// Cloning the waker is not necessary, `wake_by_ref()` enables the same operation
+    /// without extra cloning/dropping.
+    ///
+    /// ### Example
+    /// ```rust,ignore
+    /// waker.clone().wake();
+    /// ```
+    /// Should be written
+    /// ```rust,ignore
+    /// waker.wake_by_ref();
+    /// ```
+    #[clippy::version = "1.75.0"]
+    pub WAKER_CLONE_WAKE,
+    perf,
+    "cloning a `Waker` only to wake it"
+}
+
 pub struct Methods {
     avoid_breaking_exported_api: bool,
     msrv: Msrv,
@@ -3777,6 +3800,7 @@ impl_lint_pass!(Methods => [
     ITER_OUT_OF_BOUNDS,
     PATH_ENDS_WITH_EXT,
     REDUNDANT_AS_STR,
+    WAKER_CLONE_WAKE,
 ]);
 
 /// Extracts a method call name, args, and `Span` of the method name.
@@ -4365,6 +4389,9 @@ impl Methods {
                     }
                     unnecessary_literal_unwrap::check(cx, expr, recv, name, args);
                 },
+                ("wake", []) => {
+                    waker_clone_wake::check(cx, expr, recv);
+                }
                 ("write", []) => {
                     readonly_write_lock::check(cx, expr, recv);
                 }
--- a/clippy_lints/src/methods/waker_clone_wake.rs
+++ b/clippy_lints/src/methods/waker_clone_wake.rs
@@ -1,6 +1,6 @@
 use clippy_utils::diagnostics::span_lint_and_sugg;
 use clippy_utils::source::snippet_with_applicability;
-use clippy_utils::{match_def_path, paths};
+use clippy_utils::{is_trait_method, match_def_path, paths};
 use rustc_errors::Applicability;
 use rustc_hir::{Expr, ExprKind};
 use rustc_lint::LateContext;
@@ -13,10 +13,11 @@ pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, recv: &'
 
     if let Some(did) = ty.ty_adt_def()
         && match_def_path(cx, did.did(), &paths::WAKER)
-        && let ExprKind::MethodCall(func, waker_ref, &[], _) = recv.kind
-        && func.ident.name == sym::clone
+        && let ExprKind::MethodCall(_, waker_ref, &[], _) = recv.kind
+        && is_trait_method(cx, recv, sym::Clone)
     {
         let mut applicability = Applicability::MachineApplicable;
+        let snippet = snippet_with_applicability(cx, waker_ref.span.source_callsite(), "..", &mut applicability);
 
         span_lint_and_sugg(
             cx,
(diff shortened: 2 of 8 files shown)


Alexendoo:
This'd be a good one @y21

clippy_lints/src/methods/mod.rs
```diff
@@ -3632,6 +3633,28 @@ declare_clippy_lint! {
     "`as_str` used to call a method on `str` that is also available on `String`"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `clone().wake_by_ref()`
```
y21, line 3638:
```suggestion
    /// Checks for usage of `waker.clone().wake()`
```
I assume you meant to use `wake` here. (Also adding a receiver makes it clear that it's about `Waker`s)

clippy_lints/src/methods/waker_clone_and_wake.rs
```diff
@@ -0,0 +1,34 @@
...
+use clippy_utils::{match_def_path, paths};
+use rustc_errors::Applicability;
+use rustc_hir::{Expr, ExprKind};
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::WAKER_CLONE_AND_WAKE;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, recv: &'tcx Expr<'_>) {
+    let ty = cx.typeck_results().expr_ty(recv).peel_refs();
```
y21, line 12: Nit: the `peel_refs()` is pretty much a no op here, isn't it? `waker.clone()` always returns an owned `Waker`.

tests/ui/waker_clone_and_wake.rs
```diff
@@ -0,0 +1,5 @@
+pub fn wake(cx: &mut std::task::Context) {
+    cx.waker().clone().wake()
+}
```
y21, line 3: It'd be good to have some test cases for where this shouldn't lint.
- `waker.wake();`
- `waker.clone().wake_by_ref();`

clippy_lints/src/methods/waker_clone_and_wake.rs
```diff
@@ -0,0 +1,34 @@
...
+
+    if let Some(did) = ty.ty_adt_def() &&
+        match_def_path(cx, did.did(), &paths::WAKER) &&
+        let ExprKind::MethodCall(func, waker_ref, &[], _) = recv.kind &&
+        func.ident.name == sym::clone
```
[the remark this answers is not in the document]
a1phyr: Nice ! I copied this formatting from some other lint and found it very annoying

clippy_lints/src/methods/mod.rs
a1phyr: Thanks !

clippy_lints/src/methods/waker_clone_and_wake.rs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        let mut applicability = Applicability::MachineApplicable;
+
+        span_lint_and_sugg(
+            cx,
+            WAKER_CLONE_AND_WAKE,
+            expr.span,
```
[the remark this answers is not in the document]
a1phyr: I did that because some other lints already do.

For example,
```rust
pub fn test(seeker: &mut dyn io::Seek) -> io::Result<u64> {
    seeker.seek(io::SeekFrom::Current(0))
}
```

results in
```
warning: using `SeekFrom::Current` to start from current position
  --> src/main.rs:18:5
   |
   |     seeker.seek(io::SeekFrom::Current(0))
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace with: `seeker.stream_position()`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#seek_from_current
   = note: `#[warn(clippy::seek_from_current)]` on by default
```

But both are probably used, so just tell me.

tests/ui/waker_clone_and_wake.rs
a1phyr: I added those tests cases.

But maybe they should probably be linted too !

Especially the second one, which could be part of a larger `useless_clone` lint (I am surprised that it doesn't exist).

y21: > Especially the second one, which could be part of a larger `useless_clone` lint (I am surprised that it doesn't exist).

FWIW, this exists: [`redundant_clone`](https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_clone), but it's in the nursery category because of too many false positives and is also undergoing a large rewrite. It looks for a different kind of pattern though (cloning a value and letting the old value drop w/o further use).

y21:
Can you add a changelog entry to the PR description?
Something like
> changelog: [`waker_clone_wake`]: new lint

should be good enough

a1phyr:
Isn't that the role of release ?

Where should I put that ? Section for 1.75 isn't there yet.

y21:
> Where should I put that ? Section for 1.75 isn't there yet.

Sorry, I meant the PR body. The text area where you're currently describing what the PR does.
image

The changelog entry should go there (doesn't really matter where I don't think, could put that at the bottom). This is all normally explained in the input box when you (are about to) open a PR, it might be that you skipped that part. :)
You can take a look at other open PR descriptions for how this should look like.


> Isn't that the role of release ?

For every release, someone manually reviews all the changelog comments from all merged PRs in that timeframe and prepares the changelog.md for that release.
These changelog comments are automatically extracted by [a script](https://github.com/rust-lang/rust-clippy/blob/master/util/fetch_prs_between.sh).
So, every PR must have a `changelog: ...` entry in the description, or else it can't be merged.
(That's how I understand it, anyway)

a1phyr:
Done!

clippy_lints/src/methods/waker_clone_wake.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use super::WAKER_CLONE_WAKE;
+
+pub(super) fn check<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>, recv: &'tcx Expr<'_>) {
+    let ty = cx.typeck_results().expr_ty(recv);
+
+    if let Some(did) = ty.ty_adt_def()
+        && match_def_path(cx, did.did(), &paths::WAKER)
+        && let ExprKind::MethodCall(func, waker_ref, &[], _) = recv.kind
+        && func.ident.name == sym::clone
```
y21, line 17: We should probably use [`is_trait_method(cx, recv, sym::Clone)`](https://doc.rust-lang.org/stable/nightly-rustc/clippy_utils/fn.is_trait_method.html) here, instead of only checking the identifier.

It avoids linting in one edge-case where it shouldn't:
```rs
#![feature(noop_waker)]
use std::task::Waker;

struct MyWaker;
impl MyWaker {
  pub fn clone(&self) -> Waker {
    Waker::noop()
  }
}

MyWaker.clone().wake();
```
This would also be a good test.

(I don't think anyone is going to write code like this, but `MachineApplicable` suggestions are expected to always compile)

y21 (approved):
This looks good to me. cc @Alexendoo

Alexendoo (approved):

y21:
Thanks for implementing this lint @a1phyr! :)

PR #11699, opened 2023-10-23
Set doc-tests to `no_run`

Alexendoo:
This excludes `should_panic` tests, those are still run to ensure they panic. Most of our other doc snippets don't gain much from being run though so this frees up a nice bit of CI time

It also fixes the occasional issue such as `foo.txt`s being created https://github.com/rust-lang/rust-clippy/blob/f942470ca774b9648bac042f5d4c4ec74b81b61a/clippy_lints/src/permissions_set_readonly_false.rs#L19

changelog: none

### DIFF
--- a/book/src/lint_configuration.md
+++ b/book/src/lint_configuration.md
@@ -774,7 +774,7 @@ Additional dotfiles (files or directories starting with a dot) to allow
 
 ## `enforce-iter-loop-reborrow`
 #### Example
-```
+```no_run
 let mut vec = vec![1, 2, 3];
 let rmvec = &mut vec;
 for _ in rmvec.iter() {}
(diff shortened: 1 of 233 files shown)


llogiq:
I wonder if it wouldn't be better to either use `rust,ignore` or just change CI to only run `--lib` tests, as both would still allow us to run the doctests manually if we so desire without changing a lot of lines.

Alexendoo:
I think we should still be checking that the examples compile in CI, `ignored` is used for doc snippets that would not compile, so manually running ignored tests is not going to result in a lot of errors

llogiq:
Ah, yes, we want to compile but not necessarily run the tests. That makes sense.

@bors r+

llogiq:
For some reason the collect\_metadata check failed.

Alexendoo:
Ah, there were a couple code blocks that got copied to `book/src/lint_configuration.md`

PR #11700, opened 2023-10-23
[`pathbuf_init_then_push`]:  Checks for calls to `push` immediately a…

lengyijun:
changelog: [`pathbuf_init_then_push`]: new lint: Checks for calls to `push` immediately after creating a new `PathBuf`

Just a mirror of VEC_INIT_THEN_PUSH

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5712,6 +5712,7 @@ Released 2018-09-13
 [`partialeq_to_none`]: https://rust-lang.github.io/rust-clippy/master/index.html#partialeq_to_none
 [`path_buf_push_overwrite`]: https://rust-lang.github.io/rust-clippy/master/index.html#path_buf_push_overwrite
 [`path_ends_with_ext`]: https://rust-lang.github.io/rust-clippy/master/index.html#path_ends_with_ext
+[`pathbuf_init_then_push`]: https://rust-lang.github.io/rust-clippy/master/index.html#pathbuf_init_then_push
 [`pattern_type_mismatch`]: https://rust-lang.github.io/rust-clippy/master/index.html#pattern_type_mismatch
 [`permissions_set_readonly_false`]: https://rust-lang.github.io/rust-clippy/master/index.html#permissions_set_readonly_false
 [`positional_named_format_parameters`]: https://rust-lang.github.io/rust-clippy/master/index.html#positional_named_format_parameters
--- a/clippy_lints/src/pathbuf_init_then_push.rs
+++ b/clippy_lints/src/pathbuf_init_then_push.rs
@@ -1,0 +1,193 @@
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::path_to_local_id;
+use clippy_utils::source::{snippet, snippet_opt};
+use clippy_utils::ty::is_type_diagnostic_item;
+use rustc_ast::{LitKind, StrStyle};
+use rustc_errors::Applicability;
+use rustc_hir::def::Res;
+use rustc_hir::{BindingMode, Block, Expr, ExprKind, HirId, LetStmt, PatKind, QPath, Stmt, StmtKind, TyKind};
+use rustc_lint::{LateContext, LateLintPass, LintContext};
+use rustc_middle::lint::in_external_macro;
+use rustc_session::impl_lint_pass;
+use rustc_span::{sym, Span, Symbol};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for calls to `push` immediately after creating a new `PathBuf`.
+    ///
+    /// ### Why is this bad?
+    /// Multiple `.join()` calls are usually easier to read than multiple `.push`
+    /// calls across multiple statements. It might also be possible to use
+    /// `PathBuf::from` instead.
+    ///
+    /// ### Known problems
+    /// `.join()` introduces an implicit `clone()`. `PathBuf::from` can alternativly be
+    /// used when the `PathBuf` is newly constructed. This will avoild the implicit clone.
+    ///
+    /// ### Example
+    /// ```rust
+    /// # use std::path::PathBuf;
+    /// let mut path_buf = PathBuf::new();
+    /// path_buf.push("foo");
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// # use std::path::PathBuf;
+    /// let path_buf = PathBuf::from("foo");
+    /// // or
+    /// let path_buf = PathBuf::new().join("foo");
+    /// ```
+    #[clippy::version = "1.81.0"]
+    pub PATHBUF_INIT_THEN_PUSH,
+    restriction,
+    "`push` immediately after `PathBuf` creation"
+}
+
+impl_lint_pass!(PathbufThenPush<'_> => [PATHBUF_INIT_THEN_PUSH]);
+
+#[derive(Default)]
+pub struct PathbufThenPush<'tcx> {
+    searcher: Option<PathbufPushSearcher<'tcx>>,
+}
+
+struct PathbufPushSearcher<'tcx> {
+    local_id: HirId,
+    lhs_is_let: bool,
+    let_ty_span: Option<Span>,
+    init_val: Expr<'tcx>,
+    arg: Option<Expr<'tcx>>,
+    name: Symbol,
+    err_span: Span,
+}
+
+impl<'tcx> PathbufPushSearcher<'tcx> {
+    /// Try to generate a suggestion with `PathBuf::from`.
+    /// Returns `None` if the suggestion would be invalid.
+    fn gen_pathbuf_from(&self, cx: &LateContext<'_>) -> Option<String> {
+        if let ExprKind::Call(iter_expr, []) = &self.init_val.kind
+            && let ExprKind::Path(QPath::TypeRelative(ty, segment)) = &iter_expr.kind
+            && let TyKind::Path(ty_path) = &ty.kind
+            && let QPath::Resolved(None, path) = ty_path
+            && let Res::Def(_, def_id) = &path.res
+            && cx.tcx.is_diagnostic_item(sym::PathBuf, *def_id)
+            && segment.ident.name == sym::new
+            && let Some(arg) = self.arg
+            && let ExprKind::Lit(x) = arg.kind
+            && let LitKind::Str(_, StrStyle::Cooked) = x.node
+            && let Some(s) = snippet_opt(cx, arg.span)
+        {
+            Some(format!(" = PathBuf::from({s});"))
+        } else {
+            None
+        }
+    }
+
+    fn gen_pathbuf_join(&self, cx: &LateContext<'_>) -> Option<String> {
+        let arg = self.arg?;
+        let arg_str = snippet_opt(cx, arg.span)?;
+        let init_val = snippet_opt(cx, self.init_val.span)?;
+        Some(format!(" = {init_val}.join({arg_str});"))
+    }
+
+    fn display_err(&self, cx: &LateContext<'_>) {
+        if clippy_utils::attrs::span_contains_cfg(cx, self.err_span) {
+            return;
+        }
+        let mut sugg = if self.lhs_is_let {
+            String::from("let mut ")
+        } else {
+            String::new()
+        };
+        sugg.push_str(self.name.as_str());
+        if let Some(span) = self.let_ty_span {
+            sugg.push_str(": ");
+            sugg.push_str(&snippet(cx, span, "_"));
+        }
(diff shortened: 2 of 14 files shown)


clippy_lints/src/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,151 @@
...
+    /// The `.join()` is easier to read than multiple `push` calls.
+    ///
+    /// ### Example
+    /// ```rust
+    /// let mut path_buf = PathBuf::new();
+    /// path_buf.push("foo");
+    /// ```
+    /// Use instead:
+    /// ```rust
+    /// let path_buf = PathBuf::new().join("foo");
```
Centri3, line 27: Why wouldn't you use `PathBuf::from` here? We should lint both of these and suggest it instead, imo.

I suppose this would insert `/` or `\\` automatically but that's all I can really think of.

CHANGELOG.md
```diff
@@ -5340,6 +5340,8 @@ Released 2018-09-13
 [`partialeq_to_none`]: https://rust-lang.github.io/rust-clippy/master/index.html#partialeq_to_none
 [`path_buf_push_overwrite`]: https://rust-lang.github.io/rust-clippy/master/index.html#path_buf_push_overwrite
 [`path_ends_with_ext`]: https://rust-lang.github.io/rust-clippy/master/index.html#path_ends_with_ext
+[`pathbuf_init_then_push`]: https://rust-lang.github.io/rust-clippy/master/index.html#pathbuf_init_then_push
+[`pathbuf_then_push`]: https://rust-lang.github.io/rust-clippy/master/index.html#pathbuf_then_push
```
Centri3, line 5344: Please run `cargo dev update_lints`

lengyijun:
Someone may care about the performance: .join() introduce an implicit clone
But personally, I don't care

clippy_lints/src/pathbuf_init_then_push.rs
lengyijun: I think this is very rare
Moreover, I'm not familar with the details of pathbuf

lengyijun: I doubt it's not trial after reading https://github.com/rust-lang/rust-clippy/pull/11453

Centri3: wdym? I'm not sure why `PathBuf::new().join("foo").join("bar")` can't be `PathBuf::from("foo/bar")` (`/` is crossplatform iirc)

lengyijun: I don't like overeager suggestion

xFrednet:
Hey @lengyijun, would you mind rebasing this on master?

@ARandomDev99 This might be a good PR to review. There are some conflicts, but the implementation should be unaffected by the rebase.

If you comment something on this PR, I can also add you as an assignee

r? xFrednet

ARandomDev99:
I concur with Centri3. Instead of multiple calls to `PathBuf::join`, it is better to suggest using `PathBuf::from` while dealing with only string literals. Additionally, I believe this lint should be in the `pedantic` category instead of `complexity`.

xFrednet:
>  Additionally, I believe this lint should be in the `pedantic` category instead of `complexity`.

Could you explain why you feel this should be `pedantic`?

`clippy::vec_init_then_push` is warn-by-default and if this is free from FPs I think it's reasonable to have it as warn-by-default as well. We can bikeshed this part and leave it to the FCP before we merge this :)

lengyijun:
blocked by https://github.com/rust-lang/rust/issues/123352

ARandomDev99:
> Could you explain why you feel this should be `pedantic`?

@lengyijun said:
> Someone may care about the performance: .join() introduce an implicit clone

This doesn't seem like a change that everyone would prefer. Having it allowed by default seems more appropriate to me. Unlike `clippy::vec_init_then_push`, which improves performance, this lint only suggests replacing calls to `push` with `join`.

xFrednet:
> This doesn't seem like a change that everyone would prefer.

Ahh, yep you're right. Then `pedantic` would be the safe choice :+1: 

> I'm still relatively new around here. What does FCP mean? 😅

No problem, FCP stands for *final comment period* it's a new concept for Clippy. When a new lint is ready to be merged, we create a topic on zulip to get feedback. If no concerns are raised within ~1 week we merge the PR, otherwise we discuss the issues :)

clippy_lints/src/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    fn gen_pathbuf_join(&self, cx: &LateContext<'_>) -> Option<String> {
+        let arg = self.arg?;
+        let arg_str = snippet_opt(cx, arg.span)?;
+        let init_val = snippet_opt(cx, self.init_val.span)?;
+        Some(format!(" = {init_val}.join({arg_str});"))
+    }
+
+    fn display_err(&self, cx: &LateContext<'_>) {
```
xFrednet, line 86: I think here it would be good to also check for `#[cfgs]` as the `push` might be conditional: 

```suggestion
    fn display_err(&self, cx: &LateContext<'_>) {
        if clippy_utils::attrs::span_contains_cfg(self.err_span) { return; }
```

tests/ui/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,15 @@
+#![warn(clippy::pathbuf_init_then_push)]
```
xFrednet, line 1: Could you also add some tests for code snippets that shouldn't trigger the lint. Something like:

```rs
let mut x = PathBuf::new();
println!("{}", x.display());
x.push("Duck");
```

And also this test, which should trigger the lint:

```rs
let mut path_buf = PathBuf::from("foo").join("bar");
path_buf.push("buz");
```

xFrednet (commented):
The implementation looks good to me overall. I would like to see more tests, also negative examples of code that shouldn't trigger the lint. I left some example comments. 

There is also a random binary file called `c`. Can you remove that one?

If you rebase, we could also take a look at the lintcheck output in our CI :D

----

Sorry, that it took this long, it again slipped my mind :see_no_evil:

xFrednet:
Let me know when this PR is ready to be reviewed again. I'm assuming the last commits are not ready yet due to their name :)

tests/ui/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,15 @@
...
+
+use std::path::PathBuf;
+
+fn main() {
+    let mut path_buf = PathBuf::new();
+    path_buf.push("foo");
```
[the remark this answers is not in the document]
lengyijun: no interest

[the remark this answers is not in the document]
lengyijun: no interest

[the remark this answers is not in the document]
lengyijun: ready to review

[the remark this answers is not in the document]
xFrednet: I don't understand the "no interest" comments, were they intended for me?

Could you also add the example from my comment as a test?

clippy_lints/src/pathbuf_init_then_push.rs
xFrednet: Now it would make sense to change this to `PathBuf::from` since that's an actual suggestion of the lint.

You can add two suggestions in the code description. Just have a comment like:

```
// Use PathBuf::from:
...

// You can also use `.join()`
...
```

clippy_lints/src/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,190 @@
...
+    lhs_is_let: bool,
+    let_ty_span: Option<Span>,
+    init_val: Expr<'tcx>,
+    arg: Option<Expr<'tcx>>,
+    name: Symbol,
+    err_span: Span,
+}
+
+impl<'tcx> PathbufPushSearcher<'tcx> {
+    // try to generate `PathBuf::from`
```
xFrednet, line 59: NIT:
```suggestion
    /// Try to generate a suggestion with `PathBuf::from`.
    /// Returns `None` if the suggestion would be invalid. 
```

clippy_lints/src/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,190 @@
...
+                let_ty_span: local.ty.map(|ty| ty.span),
+                err_span: local.span,
+                init_val: *init_expr,
+                arg: None,
+            });
+        }
+    }
+
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'_>) {
+        if self.searcher.is_none()
```
xFrednet, line 150: This should probably also clear the `searcher`. If there is an assign expression between the initial `PathBuf` creation and the `x.push(y)` call, it's better not to lint. This could for instance modify `y`. 

```suggestion
        if let Some(mut searcher) = self.searcher.take() 
```

clippy_lints/src/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,190 @@
...
+                searcher.display_err(cx);
+            }
+        }
+    }
+
+    fn check_block_post(&mut self, cx: &LateContext<'tcx>, _: &'tcx Block<'tcx>) {
+        if let Some(searcher) = self.searcher.take() {
+            searcher.display_err(cx);
+        }
+    }
```
xFrednet, line 189: The `searcher.display_err(cx)` should never be executed, since the `x.push();` statement from `check_stmt` always emits the lint if there is something to lint.

```suggestion
    fn check_block_post(&mut self, cx: &LateContext<'tcx>, _: &'tcx Block<'tcx>) {
        self.searcher = None;
    }
```

clippy_lints/src/pathbuf_init_then_push.rs
xFrednet: If the lint no longer triggers after this, it could be that the `check_expr` call from the variable `init` expression calls this function and clears the searcher.

I think the easiest solution would be to move this and the `check_local` code into `check_stmt` and handle everything from the statement level downwards

xFrednet (commented):
I left some more comments, they should hopefully all be easy to fix. They are hopefully not too nitpicky ^^'

Let me know, if you have any question :D

clippy_lints/src/pathbuf_init_then_push.rs
xFrednet: @lengyijun have you seen this comment?

lengyijun:
https://github.com/rust-lang/rust-clippy/blob/279494eced419dca11c4b8e589fc6cec9dc82f46/lintcheck/src/input.rs#L182C16-L187C19

Emmm

xFrednet:
I assume the `Emmm` is referring to the comment?

You can use [span_contains_comment](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.span_contains_comment.html#) to check if there is a comment between the creation and the `push`

xFrednet:
Okay, I'd say the FCP has concluded. It sounds like the lint is okay, but we had others suggest the `restriction` group instead, since it might be more readable but has a performance overhead. Could you change the lint group? Then it should be ready for merge. With the addition of [span_contains_comment](https://doc.rust-lang.org/nightly/nightly-rustc/clippy_utils/fn.span_contains_comment.html#) as mentioned above :D

clippy_lints/src/pathbuf_init_then_push.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    fn gen_pathbuf_join(&self, cx: &LateContext<'_>) -> Option<String> {
+        let arg = self.arg?;
+        let arg_str = snippet_opt(cx, arg.span)?;
+        let init_val = snippet_opt(cx, self.init_val.span)?;
+        Some(format!(" = {init_val}.join({arg_str});"))
+    }
+
+    fn display_err(&self, cx: &LateContext<'_>) {
+        if clippy_utils::attrs::span_contains_cfg(cx, self.err_span) {
```
xFrednet, line 93: NIT: Could you add a test for this? Something like this:

```rs
   let mut path_buf = PathBuf::new();
   #[cfg(cats)]
   path_buf.push("foo");
```

xFrednet (approved):

xFrednet:
This version looks good to me, and the two detected cases in our CI are also very good: https://github.com/rust-lang/rust-clippy/actions/runs/10004116830?pr=11700

Thank you for the last changes. I left a tiny nit, which should be simple to fix. Then you can `r=me`

---

Roses are red,
This PR is good,
I delegate this,
To bors the bot

PR #11701, opened 2023-10-23
Use multiple pushes in `vec_init_then_push` example

Alexendoo:
Makes the perf argument clearer, since a single push doesn't have unnecessary allocations compared to `vec![x]`

changelog: none


Jarcho:
Thank you. @bors r+

PR #11712, opened 2023-10-25
Remove the `internal_warn` lint category

Alexendoo:
`LINT_AUTHOR` and `DUMP_HIR` can be removed since their corresponding lint passes are only used for side effects

The metadata collector lint can be made an `internal` lint since `clippy::internal` is set to deny when running the collector

Also renames some internal lints/passes

changelog: none


dswij (approved):
Makes sense. Thanks for this!

PR #11723, opened 2023-10-27
Remove internal feature from clippy_utils

Alexendoo:
It's only used to gate a few `const`s, removing the feature gate means it doesn't have to be recompiled when moving between a normal and `-F internal` build/test/etc

changelog: none


Jarcho:
Looks reasonable. @bors r+

PR #11724, opened 2023-10-27
Fix missing parenthesis in suboptimal floating point help

### LINKED ISSUE #11559: `mul_add` suggestion from `suboptimal_flops` applies unary minus to unsigned integer, leading to compiler error
### Summary

This issue is triggered when subtracting a cast from an unsigned type in a `mul_add` context. Once again, parentheses around this cast to the rescue 😅 

### Reproducer

I tried this code:

```rust
#[must_use]
pub fn proportion(start: f64, drop: f64, index: usize) -> f64 {
    #[allow(clippy::cast_precision_loss)]
    let result = start - (drop * index as f64);

    result
}
```

```rust
#[must_use]
pub fn proportion(start: f64, drop: f64, index: usize) -> f64 {
    #[allow(clippy::cast_precision_loss)]
    let result = drop.mul_add(-(index as f64), start);

    result
}
```

Instead, this happened:

```rust
#[must_use]
pub fn proportion(start: f64, drop: f64, index: usize) -> f64 {
    #[allow(clippy::cast_precision_loss)]
    let result = drop.mul_add(-index as f64, start);

    result
}
```

which is a compiler error.

### Version

_No response_

### Additional Labels

_No response_

llogiq:
This fixes #11559 by adding a branch in the `Neg` implementation for `Sugg` that adds parentheses to keep precedence in order, then using that in the suggestion. I also removed some needless `.to_string()`s while I was at it.

---

changelog: none

### DIFF
--- a/clippy_lints/src/floating_point_arithmetic.rs
+++ b/clippy_lints/src/floating_point_arithmetic.rs
@@ -323,9 +323,9 @@ fn check_powi(cx: &LateContext<'_>, expr: &Expr<'_>, receiver: &Expr<'_>, args: 
                     let maybe_neg_sugg = |expr, hir_id| {
                         let sugg = Sugg::hir(cx, expr, "..");
                         if matches!(op, BinOpKind::Sub) && hir_id == rhs.hir_id {
-                            format!("-{}", sugg.maybe_par())
+                            -sugg
                         } else {
-                            sugg.to_string()
+                            sugg
                         }
                     };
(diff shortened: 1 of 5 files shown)


llogiq:
I think Jarcho may be overworked a bit, so

@bors r? @blyxyas

blyxyas (approved):
LGTM, thanks! :heart:

PR #11726, opened 2023-10-27
Expand docs on clippy::todo

Wilfred:
https://doc.rust-lang.org/nightly/core/macro.todo.html describes that `todo!()` is intended for explicitly unfinished code. Explain this, and mention `unimplemented!()` as an alternative.

Whilst we're here, improve the punctuation on the other lints.

changelog: [`todo`]: expand docs


dswij (approved):

dswij:
Thanks! @bors r+

Issue #9274, opened 2022-08-01
`read_zero_byte_vec` emitted when `read` is in the next block


unrealhoang:
### Summary

As the heuristic is checking if the next statement of an empty vec includes `read(&mut v)`, if the next statement/expression is a block with `resize` before `read`, the lint is still fired. 

### Lint Name

read_zero_byte_vec

### Reproducer

I tried this code:

```rust
    let mut v = Vec::new();
    {
        v.resize(10, 0);
        r.read(&mut v).unwrap();
    }
```

I saw this happen:

```
error: reading zero byte data to `Vec`
   --> src/main.rs:10:5
    |
130 | /     {
131 | |         v.resize(10, 0);
132 | |         r.read(&mut v).unwrap();
133 | |     }
    | |_____^
    |
    = note: `#[deny(clippy::read_zero_byte_vec)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#read_zero_byte_vec
```

I expected to not seeing this lint fired off.

### Version

```text
rustc 1.64.0-nightly (0f4bcadb4 2022-07-30)
binary: rustc
commit-hash: 0f4bcadb46006bc484dad85616b484f93879ca4e
commit-date: 2022-07-30
host: x86_64-unknown-linux-gnu
release: 1.64.0-nightly
LLVM version: 14.0.6
```

tamaroning:
I added this lint. (#8964)
For now, this lint checks reads immidiately after creating `Vec`s and dosen't catch other many cases. 
e.g.
```rust
// cannot lint
let mut data = Vec::with_capacity(100);
();
f.read(&mut data).unwrap();
``` is because linting all cases is tough.

Linting all cases is tough so I made a compromise.

**Comment by wildbook** [comment]
I just discovered this lint through false positives when running clippy against older code.

All of my so far hits can be simplified down to variants of this:
```rs
let mut data = vec![];
match () {
    () => {
        data.resize(10, 0);
        input.read_exact(&mut data)?;
    },
}
```

It'd be nice if there was a mention of the false positives or restrictions in the lint index, similar to most other lints with known failure cases. I'm also not sure that a lint with so easy-to-trigger false positives should be deny by default, but perhaps I'm just unlucky with it.

joshtriplett:
It's fine if this has false negatives, but it needs to not have false positives. This shouldn't be checking "the next statement" in a way that includes the whole next block.

joshtriplett:
[...]

PR #11727, opened 2023-10-28
move `read_zero_byte_vec` to nursery

dswij:
I think the concerns in #9274 are valid, and we should move this to nursery while we're reworking this.

changelog: [`read_zero_byte_vec`] moved to nursery


Centri3 (approved):

Centri3:
Thank you!

@bors r+

PR #11743, opened 2023-11-01
Fix `dbg_macro` semi span calculation

Alexendoo:
`span_including_semi` was using a `BytePos` to index into a file's source which happened to work because the root file of the test started at `BytePos` 0, it didn't work for other files

changelog: none


xFrednet (approved):

xFrednet:
LGTM, thank you!

@bors r+

PR #11744, opened 2023-11-01
Fix get_first false negative for VecDeque

### LINKED ISSUE #11695: get_first false negative: VecDeque.get(0)
### Summary

The rationale for preferring `vec.first()` over `vec.get(0)` also applies to preferring `vecdeque.front()` over `vecdeque.get(0)`.

### Lint Name

get_first

### Reproducer

```rust
use std::collections::VecDeque;

fn main() {
    let vec = Vec::from(['a', 'b', 'c']);
    println!("{:?}", vec.get(0));

    let vecdeque = VecDeque::from(['a', 'b', 'c']);
    println!("{:?}", vecdeque.get(0));
}
```

Today, `cargo clippy` only suggests replacing `vec.get(0)` with `vec.first()`. It does not suggest replacing `vecdeque.get(0)` with `vecdeque.front()`.

### Version

```text
rustc 1.75.0-nightly (1c05d50c8 2023-10-21)
binary: rustc
commit-hash: 1c05d50c8403c56d9a8b6fb871f15aaa26fb5d07
commit-date: 2023-10-21
host: aarch64-unknown-linux-gnu
release: 1.75.0-nightly
LLVM version: 17.0.3
```

matthri:
fixes #11695 

Also run the lint on `VecDeque` and suggest using `.front()` instead of `.get(0)` when trying to access the first element.

PS: At first I implemented the VecDeque Lint in a separate `if_chain` (see the previous commit).
Let me know if thats the preferred way, then I will remove the refactoring into one block.

changelog: [`get_first`]: fix false negative: Also lint `VecDeque` and suggest using `front()`

### DIFF
--- a/clippy_lints/src/methods/get_first.rs
+++ b/clippy_lints/src/methods/get_first.rs
@@ -20,30 +20,23 @@ pub(super) fn check<'tcx>(
     if_chain! {
         if let Some(method_id) = cx.typeck_results().type_dependent_def_id(expr.hir_id);
         if let Some(impl_id) = cx.tcx.impl_of_method(method_id);
-        if cx.tcx.type_of(impl_id).instantiate_identity().is_slice();
+        let identity = cx.tcx.type_of(impl_id).instantiate_identity();
         if let hir::ExprKind::Lit(Spanned { node: LitKind::Int(0, _), .. }) = arg.kind;
         then {
-            let mut app = Applicability::MachineApplicable;
-            let slice_name = snippet_with_applicability(cx, recv.span, "..", &mut app);
-            span_lint_and_sugg(
-                cx,
-                GET_FIRST,
-                expr.span,
-                &format!("accessing first element with `{slice_name}.get(0)`"),
-                "try",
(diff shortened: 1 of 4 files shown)


clippy_lints/src/methods/get_first.rs
```diff
@@ -18,20 +20,34 @@ pub(super) fn check<'tcx>(
...
+                    span_lint_and_sugg(
+                        cx,
+                        GET_FIRST,
+                        expr.span,
+                        &format!("accessing first element with `{slice_name}.get(0)`"),
+                        "try",
+                        format!("{slice_name}.front()"),
+                        app,
+                    );
+            }
```
Jarcho, line 50: Is there any reason for this to be two blocks?

matthri: Not sure if I got the question correctly, but for slice types the suggested fix is to use `first()` and for the VecDeque the fix is `front()`.

Jarcho:
Thank you. @bors r+

PR #11750, opened 2023-11-02
Replace if_chain with let chains

### LINKED ISSUE #9353: Remove all `if_chain!` macros
### Description

Current Clippy code has a lot of legacy `if_chain!` macros even though the nightly [if let chains](https://rust-lang.github.io/rfcs/2497-if-let-chains.html) is already in use. I would like to remove all `if_chain!` macro usage, replacing it with the regular if statement, unless of course there is any reason to avoid that?

One thing to note is that `cargo fmt` still hasn't been decided (see [rfc](https://github.com/rust-dev-tools/fmt-rfcs/issues/169)) on how to handle it, and doesn't seem likely to be decided soon. This is not a blocker, as the code can always be reformatted later.

#### Possible Lint?
P.S. I just realized that this might be a good automated lint because it would be useful to other existing code as well. See #9354

Alexendoo:
Closes #9353

Let chains are now supported by rustfmt 🎉

The PR is split into two commits
1. The result of running [`if-to-let-chain clippy*/**/*.rs`](https://github.com/Alexendoo/if-to-let-chain)
2. The manual clean up: fixing some errors/formatting, dogfood lints, removing the if_chain internal lint

r? @flip1995

changelog: none


flip1995:
I will review this today. No need to rebase before that. Once I'm done reviewing, you can rebase and directly merge :shipit:

clippy_lints/src/utils/internal_lints/if_chain_style.rs
flip1995, line 1: Non-blocking Question: IIRC this lint checked that infallible `let` bindings at the start or end of the `if_chain` macro were moved before or after the `if_chain` resp.? Would this same lint also make sense for `let-chains`?

flip1995 (approved):
LGTM overall. `rg if_chain` in the repo still shows a few `if_chain` usages. But those are in tests and necessary. However, I think once let-chains are stabilized, we can also remove this one special check for `if_chain` in the `index_refutable_slice` lint implementation.

The `needless_borrow_pat` test implements its own `if_chain` macro for testing. Maybe we can reuse that in `index_refutable_slice` for now and with that get rid of the `if_chain` dep in the top level `Cargo.toml`.

All of this is non-blocking for this PR. Feel free to address it here or we can deal with it in a follow-up PR.  r=me after rebase :shipit:

clippy_lints/src/utils/internal_lints/if_chain_style.rs
Jarcho: That's already caught by rustc as `irrefutable_let_patterns`.

Alexendoo:
@bors r=flip1995

The `if_chain` implementation from `index_refutable_slice` didn't seem to trigger the lint so I've left the `if_chain` dev dep there for now

Issue #11752, opened 2023-11-03
`iter_kv_map` needs to be MSRV gated on 1.54


apoelstra:
### Summary

The new iter_kv_map lint triggers even when clippy.toml MSRV is set before the `into_keys` method exists.

### Reproducer

```
cargo new clippy-demo
echo 'msrv = "1.48.0"' > clippy.toml
echo 'use std::collections::HashMap; fn main() { let x = HashMap::<(), ()>::new(); x.into_iter().map(|(k, _)| k); }' > src/main.rs
cargo clippy
```

You will get a suggestion to use `into_keys` rather than `into_iter().map`. But `into_keys` is not available until Rust 1.54.

### Version

```text
rustc 1.73.0 (cc66ad468 2023-10-03)
binary: rustc
commit-hash: cc66ad468955717ab92600c770da8c1601a4ff33
commit-date: 2023-10-03
host: x86_64-unknown-linux-gnu
release: 1.73.0
LLVM version: 17.0.2
```

PR #11757, opened 2023-11-03
Fix iter_kv_map false positive into_keys and into_values suggestion

matthri:
fixes: #11752 

changelog: [`iter_kv_map`]: fix false positive: Don't suggest `into_keys()` and `into_values()` if the MSRV is to low


y21 (commented):
This should probably also [update the lint documentation](https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv):
https://github.com/rust-lang/rust-clippy/blob/902c79c654b45ab9b8ae425bd2c1fa0c50a3edf7/clippy_config/src/conf.rs#L252

matthri:
Thanks for pointing that out, I updated the documentation

clippy_lints/src/methods/iter_kv_map.rs
```diff
@@ -21,7 +22,11 @@ pub(super) fn check<'tcx>(
     expr: &'tcx Expr<'tcx>,  // .iter().map(|(_, v_| v))
     recv: &'tcx Expr<'tcx>,  // hashmap
     m_arg: &'tcx Expr<'tcx>, // |(_, v)| v
+    msrv: &Msrv,
 ) {
+    if map_type == "into_iter" && !msrv.meets(msrvs::INTO_KEYS) {
+        return;
+    }
     if !expr.span.from_expansion()
         && let ExprKind::Closure(c) = m_arg.kind
```
y21, line 31: This lint should probably check that the closure doesn't have parameter- or return type annotations since that can help type inference or make the `map` closure act as a coercion site.
Not your problem though, that could be a PR on its own, unless you also want to try to fix that here ^^.

Examples where the machine applicable suggestion breaks: [coercions](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d7178f8b562958d52c82af1e40091ff0), [inference](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a67abf071b2ef645d886e8cab5fa2da1)

y21 (approved):
Changes look good to me 👍
I have one comment but that's not really concerned with the PR here, rather the lint itself, so feel free to ignore.

clippy_lints/src/methods/iter_kv_map.rs
matthri: I will have a try fixing this, but  I think I will open a separate PR so this one doesn't get too cluttered ^^

Alexendoo (approved):

Alexendoo:
Thanks! @bors r+

matthri:
I ran the `cargo collect-metadata` command to update `book/src/lint_configuration.md`. 
Should this maybe get mentioned in the [documentation](https://github.com/rust-lang/rust-clippy/blob/master/book/src/development/adding_lints.md#specifying-the-lints-minimum-supported-rust-version-msrv)?

Alexendoo:
Yeah a note would be good, would be good for us to see if we can make it automatic also

PR #11758, opened 2023-11-04
update references of old `msrvs` and `conf` paths

y21:
In #11685, `clippy_lints::utils::conf` and `clippy_utils::msrvs` were moved to a separate `clippy_config` crate.
I noticed that not all references to those paths were updated, so this small PR intends to fix those.

changelog: none


flip1995:
Thank you!

@bors r+

PR #11759, opened 2023-11-04
New lint: `needless_move`

### LINKED ISSUE #11721: lint suggestion: unnecessary_move
### What it does

Lints against uses of `move` before a closure or `async` block that captures nothing by reference. 

### Advantage

This is likely the result of a mistake from the implementer, which does not have the proper intuition about how closure borrow their contents. Sometimes it can be hard to reason about it.

For example the developer might not realise that a capture that is passed by value is always captured by value.

### Drawbacks

_No response_

### Example

```rust
    let a = String::new();
    let closure = move || {
        drop(a);  
    };
```

Could be written as:

```rust
    let a = String::new();
    let closure = || {
        drop(a);  
    };
```

Because the String is always captured by value, since it needs to be passed by value to `drop`.

dnbln:
A lint to check for uses of `move` on closures which aren't necessary. See #11721.

Fixes #11721.
~Currently blocked on https://github.com/rust-lang/rust/pull/117585, and an eventual sync.~

changelog: [`needless_move`]: A lint for unnecessary `move`s on closures / async blocks.

### DIFF
--- a/clippy_lints/src/needless_move.rs
+++ b/clippy_lints/src/needless_move.rs
@@ -33,9 +33,9 @@
 //! try to infer the actual min capture kind needed.
 
 use clippy_utils::diagnostics::span_lint_and_then;
-use clippy_utils::sugg::DiagnosticExt;
+use clippy_utils::sugg::DiagExt;
 use rustc_errors::Applicability;
-use rustc_hir::{CaptureBy, Closure, Expr, ExprKind, HirId};
+use rustc_hir::{CaptureBy, Expr, ExprKind, HirId};
 use rustc_hir_typeck::expr_use_visitor as euv;
 use rustc_infer::infer::TyCtxtInferExt;
 use rustc_lint::{LateContext, LateLintPass};
@@ -46,7 +46,7 @@ use rustc_session::declare_lint_pass;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Checks for closures and `async` blocks where the `move` is not necessary.
+    /// Checks for closures and `async` blocks where capturing by value (the `move` keyword) is unnecessary.
     /// E.g. all the values are captured by value into the closure / `async` block.
     ///
     /// ### Why is this bad?
@@ -96,8 +96,16 @@ declare_clippy_lint! {
 
 declare_lint_pass!(NeedlessMove => [NEEDLESS_MOVE]);
 
-impl NeedlessMove {
-    fn check_closure<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, closure: &'tcx Closure<'tcx>) {
+impl<'tcx> LateLintPass<'tcx> for NeedlessMove {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if expr.span.from_expansion() {
+            return;
+        }
+
+        let ExprKind::Closure(closure) = &expr.kind else {
+            return;
+        };
+
         let CaptureBy::Value { move_kw } = closure.capture_clause else {
             return;
         };
@@ -140,9 +148,9 @@ impl NeedlessMove {
         }
 
         let note_msg = match lint_result {
-            LintResult::NothingCaptured => "there were no captured variables, so the `move` is unnecessary",
+            LintResult::NothingCaptured => "there are no captured variables, so the `move` keyword is unnecessary",
             LintResult::Consumed => {
-                "there were consumed variables, but no borrowed variables, so the `move` is unnecessary"
+                "there are consumed variables, but no borrowed variables, so the `move` keyword is unnecessary"
             },
             LintResult::NeedMove => {
                 // there was a value which would be borrowed if it weren't for the move keyword,
@@ -155,24 +163,17 @@ impl NeedlessMove {
             cx,
             NEEDLESS_MOVE,
             expr.span,
-            "you seem to use `move`, but the `move` is unnecessary",
+            "this closure does not need to capture by value",
             |diag| {
-                diag.suggest_remove_item(cx, move_kw, "remove the `move`", Applicability::MachineApplicable);
+                diag.suggest_remove_item(
+                    cx,
+                    move_kw,
+                    "remove the `move` keyword",
+                    Applicability::MachineApplicable,
+                );
                 diag.note(note_msg);
             },
         );
-    }
-}
-
-impl<'tcx> LateLintPass<'tcx> for NeedlessMove {
-    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
-        if expr.span.from_expansion() {
-            return;
-        }
-
-        if let ExprKind::Closure(closure) = &expr.kind {
-            Self::check_closure(cx, expr, closure);
-        }
     }
 }
 
--- a/clippy_lints/src/needless_move.rs
+++ b/clippy_lints/src/needless_move.rs
@@ -42,7 +42,7 @@ use rustc_lint::{LateContext, LateLintPass};
 use rustc_middle::mir::FakeReadCause;
 use rustc_middle::ty;
 use rustc_middle::ty::UpvarCapture;
-use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_session::declare_lint_pass;
 
 declare_clippy_lint! {
     /// ### What it does
@@ -55,24 +55,25 @@ declare_clippy_lint! {
     /// the programmer adds the `move` keyword to move the variables into the closure, but
     /// then later decides that he no longer needs the variables in question, so he removes them
     /// from the body of the closure, but forgets to also remove the `move` keyword.
-    /// 
+    ///
     /// This is really just a strict coding style issue.
-    /// 
+    ///
     /// ### Caveats
     /// There are some cases where this lint will suggest removing the `move` keyword,
     /// but it would be considered idiomatic to keep it.
-    /// 
+    ///
     /// For example, the closure passed to `std::thread::spawn` is usually always written
     /// with the `move` keyword, even if it's not necessary:
-    /// 
+    ///
     /// ```no_run
+    /// # fn function_that_does_something_with(_: String) {}
     /// let a = String::new();
     /// std::thread::spawn(move || {
     ///    // ...
     ///   function_that_does_something_with(a); // a is moved into the closure
     /// });
     /// ```
-    /// 
+    ///
     /// ### Example
     /// ```no_run
     /// let a = String::new();
@@ -139,9 +140,7 @@ impl NeedlessMove {
         }
 
         let note_msg = match lint_result {
-            LintResult::NothingCaptured => {
-                "there were no captured variables, so the `move` is unnecessary"
-            },
+            LintResult::NothingCaptured => "there were no captured variables, so the `move` is unnecessary",
             LintResult::Consumed => {
                 "there were consumed variables, but no borrowed variables, so the `move` is unnecessary"
             },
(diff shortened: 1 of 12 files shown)


tests/ui/needless_move.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    // owned + NonCopy
+    let a = NonCopy;
+    let closure = move || {
+        with_owned(a);
+    };
+
+    // owned + Copy
+    let a = Copy;
+    let closure = move || {
```
y21, line 54: I don't think we should lint here. Removing the `move` in this case makes the closure no longer `'static`:
```rs
fn assert_static<T: 'static>(_: T) {}

let a = Copy;
assert_static(move || { with_owned(a); }); // removing the `move` causes a compiler error
```

y21 (commented):
Some drive-by comments, I noticed the pr is marked ready now :)

tests/ui/needless_move.rs
dnbln: Hmm that's an interesting case. I thought `Copy` types always get moved into the closure, even without `move`, but it seems I was wrong.

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,188 @@
...
+    fn check_closure<'tcx>(&self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, closure: &'tcx Closure<'tcx>) {
+        let CaptureBy::Value { move_kw } = closure.capture_clause else {
+            return;
+        };
+
+        // Collect moved & borrowed variables from the closure, which the closure *actually* needs.
+        let MovedVariablesCtxt {
+            moved_vars,
+            mut captured_vars,
+        } = {
```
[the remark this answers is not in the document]
y21: oh right, that's unfortunate

tests/ui/needless_move.rs
dnbln: Should be fixed now + I made all the test cases `assert_static` in 06d74077679adbc7ac3c9ad7040a618aa8d2b2b0

clippy_lints/src/needless_move.rs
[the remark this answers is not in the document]
dnbln: Actually I've found the file that does most of that analysis: https://github.com/rust-lang/rust/blob/master/compiler/rustc_hir_typeck/src/upvar.rs

Maybe exposing a few of those functions would be helpful for this lint, so I'm gonna look into it soon.

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// by value (and therefore moved into the closure).
+    captured: FxIndexMap<HirId, Vec<(ty::UpvarId, HirId, ty::BorrowKind)>>,
+
+    mutated: FxIndexMap<HirId, Vec<(ty::UpvarId, HirId)>>,
+}
+
+impl MovedVariablesCtxt {
+    fn move_common(&mut self, cmt: &euv::PlaceWithHirId<'_>, _: HirId) {
+        if let euv::PlaceBase::Upvar(vid) = cmt.place.base {
+            self.moved.entry(vid.var_path.hir_id).or_default().push(vid);
```
y21, line 168: Can you add this as a test? I suspect that this is currently a FP but I could be wrong.
```rs
let v = (String::new(), String::new());
assert_static(move || {
  let _w = v.0;
  let _h = &v.1;
});
```

dnbln: Yeah, that's indeed a FP; gonna look into it

dnbln: This should be fixed now

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Use instead:
+    /// ```no_run
+    /// let a = String::new();
+    /// let closure = || {
+    ///     drop(a);
+    /// };
+    /// ```
+    #[clippy::version = "1.76.0"]
+    pub NEEDLESS_MOVE,
+    restriction,
```
Centri3, line 70: Is there any disadvantage to applying this lint? If not, could this be `pedantic` instead?

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,222 @@
...
+use rustc_middle::ty::UpvarCapture;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for closures and `async` blocks where the `move` is not necessary.
+    /// E.g. all the values are captured by value into the closure / `async` block.
+    ///
+    /// ### Why is this bad?
+    /// Pedantry
```
Centri3, line 53: This should be expanded upon

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,222 @@
...
+                |diag| {
+                    diag.suggest_remove_item(cx, move_kw, "remove the `move`", Applicability::MachineApplicable);
+                    diag.note(note_msg);
+                },
+            );
+        };
+
+        match lint_result {
+            LintResult::NothingCaptured => {
+                lint("there were no captured variables, so the `move` is unnecessary");
```
Centri3, line 134: This could probably be a `let note_msg = match lint_result ...` instead, and early return on `NeedMove`.

Using a closure is a bit more complex, and imo harder to read

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+
+impl<'tcx> LateLintPass<'tcx> for NeedlessMove {
+    fn check_expr(&mut self, cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>) {
+        if expr.span.from_expansion() {
+            return;
+        }
+
+        let ExprKind::Closure(closure) = &expr.kind else {
```
Centri3, line 153: This could be a let-chain instead. However since this is only ever called once, maybe you can inline it?

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+impl<'tcx> euv::Delegate<'tcx> for MovedVariablesCtxt<'tcx> {
+    fn consume(&mut self, cmt: &euv::PlaceWithHirId<'tcx>, hir_id: HirId) {
+        self.move_common(cmt, hir_id);
+    }
+
+    fn borrow(&mut self, cmt: &euv::PlaceWithHirId<'tcx>, hir_id: HirId, bk: ty::BorrowKind) {
+        self.borrow_common(cmt, hir_id, bk);
+    }
```
Centri3, line 215: These should be inlined too

tests/ui/needless_move.rs
```diff
@@ -0,0 +1,1896 @@
...
+
+            #[cfg(cls)]
+            assert_eq!(panicked_at(|| block_on(foo_block())), 96);
+        }
+    }
+
+    // rust/tests/ui/async-await/deep-futures-are-freeze.rs
+
+    fn _deep_futures_are_freeze() {
+        // no-prefer-dynamic
```
Centri3, line 1463: I would like to ensure `x test` won't read these, somehow, I dunno, very cautious 😅 I doubt it does when running tests in subtrees or when they aren't at the top of the file

Or just remove them

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,232 @@
...
+    #[clippy::version = "1.76.0"]
+    pub NEEDLESS_MOVE,
+    restriction,
+    "checks for needless `move`s on closures / `async` blocks"
+}
+
+declare_lint_pass!(NeedlessMove => [NEEDLESS_MOVE]);
+
+impl NeedlessMove {
+    fn check_closure<'tcx>(cx: &LateContext<'tcx>, expr: &'tcx Expr<'tcx>, closure: &'tcx Closure<'tcx>) {
```
Centri3, line 100: Does this really need to be extracted? Can this be in `check_expr` instead?

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,232 @@
...
+                    UpvarCapture::ByRef(it.2)
+                })
+        }
+    }
+}
+
+impl<'tcx> euv::Delegate<'tcx> for MovedVariablesCtxt<'tcx> {
+    fn consume(&mut self, cmt: &euv::PlaceWithHirId<'tcx>, hir_id: HirId) {
+        if let euv::PlaceBase::Upvar(_) = cmt.place.base {
+            self.moved.push((cmt.place.clone(), hir_id));
```
Centri3, line 217: Is there any reason we can't store `&'tcx Place<'tcx>` to prevent cloning? This is cheap if projections is empty, tho you're cloning a `Vec` otherwise

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,232 @@
...
+                "there were consumed variables, but no borrowed variables, so the `move` is unnecessary"
+            },
+            LintResult::NeedMove => {
+                // there was a value which would be borrowed if it weren't for the move keyword,
+                // so we should keep it, as removing it would change semantics.
+                return;
+            },
+        };
+
+        span_lint_and_then(
```
Centri3, line 154: We should call `is_from_proc_macro` here. You can implement `WithSearchPat` for `Closure` by reusing the code for `ExprKind::Closure` in `expr_search_pat`

clippy_lints/src/needless_move.rs
```diff
@@ -0,0 +1,232 @@
...
+            let body = cx.tcx.hir().body(closure.body);
+            let infcx = cx.tcx.infer_ctxt().build();
+            euv::ExprUseVisitor::new(&mut ctx, &infcx, closure.def_id, cx.param_env, cx.typeck_results())
+                .consume_body(body);
+            ctx
+        };
+
+        let mut lint_result = LintResult::NothingCaptured;
+
+        for captured_place in cx.typeck_results().closure_min_captures_flattened(closure.def_id) {
```
Centri3, line 123: Can you run me down on this? What difference does using `ExprUseVisitor` make to just calling this, and checking if any are `UpvarCapture::ByRef(_)`? Is it because we need to filter out the captured variables which aren't used (which would somehow show up in `closure_min_captures_flattened`)?

Centri3 (commented):
Nya

clippy_lints/src/needless_move.rs
Centri3: Ah, nvm, I assume this is relevant to https://github.com/rust-lang/rust-clippy/pull/11759#discussion_r1385165647?

xFrednet:
Hey @dnbln, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?

@rustbot author

---

Also this PR needs a new reviewer since `@Centri3` is currently busy.

r? @y21 would you mind taking this review? You can also ping `@ARandomDev99` or `@GuillaumeGomez` for a pre-review if that helps :)

clippy_lints/src/needless_move.rs
dnbln: The `cmt` is not a `&'tcx` reference, so the lifetimes don't add up and the borrow checker is unhappy.

dnbln:
> Hey @dnbln, this is a ping from triage, since there hasn't been any activity in some time. Are you still planning to continue this implementation?
> 
> If you have any questions, you're always welcome to ask them in this PR or on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/257328-clippy).

Hi @xFrednet. I've honestly really began to doubt the usefulness of this lint, so I don't think I'll be investing any more time into it.

Centri3:
I do still quite like this as a restriction (very pedantic style) lint. I might get around to it eventually if nobody else picks it up :3

dnbln:
I'll just leave the script I used to extract the tests from the rust repo here, if anyone else is interested in continuing this and might find it useful:

https://gist.github.com/dnbln/6b03d98bf189d76236a51ae9e2082433

PR #11760, opened 2023-11-05
Don't check for late-bound vars, check for escaping bound vars

compiler-errors:
Fixes an assertion that didn't make sense. Many valid and well-formed types *have* late-bound vars (e.g. `for<'a> fn(&'a ())`), they just must not have *escaping* late-bound vars in order to be normalized correctly.

Addresses rust-lang/rust-clippy#11230, cc @jyn514 and @matthiaskrgr

changelog: don't check for late-bound vars, check for escaping bound vars. Addresses rust-lang/rust-clippy#11230


matthiaskrgr:
There was a test here https://github.com/rust-lang/rust-clippy/issues/11230#issuecomment-1650591002  can you just throw that into `src/tools/c`lippy/tests/ui/crashes` ?

compiler-errors:
I actually just removed the `cfg(debug_assertion)` -- I don't think they were helping much. Added the test too.

Jarcho:
The `cfg(debug_assertions)` are there because it will change an ICE into a false negative on releases.

jyn514:
i confirmed that this fixes https://github.com/rust-lang/rust-clippy/issues/11230 when applied on top of https://github.com/rust-lang/rust/pull/117595 :)
(test with `debug-assertions = true` and ` git subtree pull --prefix=src/tools/clippy/ https://github.com/rust-lang/rust-clippy/ pull/11760/head`)

Jarcho:
Ping @compiler-errors. Can you re-add the debug `cfg`s.

@rustbot author

Jarcho:
Thank you. @bors r+

I assume an escaped variable is one that is missing it's corresponding binder, correct?

matthiaskrgr:
lol, I have added the missing `changelog: ...`
@bors retry

Issue #11714, opened 2023-10-25
`arc_with_non_send_sync` can mislead users into `unsafe impl`s


kpreid:
### Description

I have, twice now, seen users encounter the `arc_with_non_send_sync` warning and conclude that the way to fix it is to `unsafe impl Send` `unsafe impl Sync` on their type. (The second occurrence is <https://github.com/AccessKit/accesskit/pull/301>; I do not recall the first.)

When I try to take the perspective of someone not knowing what this is about, it's pretty scary and cryptic:

```
warning: usage of an `Arc` that is not `Send` or `Sync`

    = note: the trait `Send` is not implemented for `MyType`
    = note: the trait `Sync` is not implemented for `MyType`
    = note: required for `Arc<MyType>` to implement `Send` and `Sync`
    = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
    = note: `#[warn(clippy::arc_with_non_send_sync)]` on by default
```

This could easily be taken as meaning that “your `Arc` is broken and you need to implement `Send` and `Sync` for it to work”.

On review, I think the clarification that is needed is that this is a _performance and style issue_, not intrinsically a bug, and that the first assumption should be that (presuming that the code otherwise compiles successfully) the `Arc` is unnecessary (or that, as with any compilation warning, something else application-specific needs to be changed), not that the code needs to be changed to force `Send + Sync` bounds by any means necessary. I'm not sure exactly how to phrase this but perhaps something like this revised message and additional help:

```
warning: this `Arc` will not be `Send` or `Sync` and could be replaced with `Rc`

    = help: `Arc<T>` does not provide any benefits over `Rc<T>` unless the contained type `T` implements `Send + Sync`
```

This way, the reader has enough information to make some guesses what a reasonable path forward is, whether or not they actually understand `Send` and `Sync` yet.

### Version

```text
rustc 1.73.0 (cc66ad468 2023-10-03)
```

### Additional Labels

@rustbot label +C-bug +good-first-issue

DataTriny:
Since a PR of mine is mentioned here, let me offer my experience encountering this warning as a casual Rust developer.

[...]

PR #11772, opened 2023-11-07
[arc_with_non_send_sync] Improve suggested resolution

partiallytyped:
Fixes: #11714 

Improved the lint message for [`arc_with_non_send_sync`] to suggest using `RC` unless user needs an Arc, then suggests wrapping in a mutex, and then suggests implementing `Sync` and `Send`.

---

changelog: [`arc_with_non_send_sync`]: Suggest RC over unsafe impl of Send and Sync

### DIFF
--- a/clippy_lints/src/arc_with_non_send_sync.rs
+++ b/clippy_lints/src/arc_with_non_send_sync.rs
@@ -14,7 +14,9 @@ declare_clippy_lint! {
     /// This lint warns when you use `Arc` with a type that does not implement `Send` or `Sync`.
     ///
     /// ### Why is this bad?
-    /// `Arc<T>` is only `Send`/`Sync` when `T` is [both `Send` and `Sync`](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-Send-for-Arc%3CT%3E),
+    /// `Arc<T>` is an Atomic `RC<T>` and guarantees that updates to the reference counter are
+    /// Atomic. This is useful in multiprocessing scenarios. To send an `Arc<T>` across processes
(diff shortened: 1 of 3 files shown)


clippy_lints/src/arc_with_non_send_sync.rs
```diff
@@ -61,19 +63,25 @@ impl<'tcx> LateLintPass<'tcx> for ArcWithNonSendSync {
...
                 |diag| {
                     with_forced_trimmed_paths!({
+                        diag.note(format!("`Arc<{arg_ty}>` is not `Send` and `Sync` as"));
+
                         if !is_send {
                             diag.note(format!("the trait `Send` is not implemented for `{arg_ty}`"));
                         }
                         if !is_sync {
                             diag.note(format!("the trait `Sync` is not implemented for `{arg_ty}`"));
                         }
```
blyxyas, line 76: Some list formatting here would be very appreciated for legibility.

```suggestion
                        diag.note(format!("`Arc<{arg_ty}>` is not `Send` and `Sync` as:"));

                        if !is_send {
                            diag.note(format!("- the trait `Send` is not implemented for `{arg_ty}`"));
                        }
                        if !is_sync {
                            diag.note(format!("- the trait `Sync` is not implemented for `{arg_ty}`"));
                        }
```

blyxyas (commented):
This is pretty much ready, just a little improvement that I think would improve readability.

blyxyas (approved):
LGTM, thanks! :heart: Could you squash the 4 commits into a single one?

blyxyas:
Yeah don't worry, use `git rebase -i HEAD~4` (replace "pick" with "s" in those four commits), adjust the commit messages and just `git push -f`. Rebasing is *the* scenario where a force push would be accepted.

Issue #11689, opened 2023-10-20
manual_memcpy suggest slicing array when not necessary


sosthene-nitrokey:
### Summary

The `manual_memcpy` can suggest slicing an array with `[..N]` even if the type of the array is `[T; N]`, which makes the slicing unnecessary.

### Reproducer

I tried this code:

```rust
    let src: &[u8] = todo!();
    let mut dest = [0; 4];
    
    for i in 0..4 {
        dest[i] = src[i];
    }
```

Clippy suggests `dest[..4].copy_from_slice(&src[..4]);`, but the `dest[..4]` is unnecessary as `dest` as a compile-time known length of `4`.

The same happens when reversing the roles with:

```rust
    let dest: &mut [u8] = todo!();
    let src = [0; 4];
    
    for i in 0..4 {
        dest[i] = src[i];
    }
```

The suggestion is the same but now the `src[..4]` slicing is unnecessary.

### Version

```text
rustc 1.73.0 (cc66ad468 2023-10-03)
binary: rustc
commit-hash: cc66ad468955717ab92600c770da8c1601a4ff33
commit-date: 2023-10-03
host: x86_64-unknown-linux-gnu
release: 1.73.0
LLVM version: 17.0.2
```

PR #11778, opened 2023-11-08
fix: [manual_memcpy] reduce indexing suggestions when array length is equal to loop range

granddaifuku:
fixes: #11689 

This PR improves `manual_memcpy` suggestions by reducing unnecessary indexing.

For example, 
```rust
let src = [0, 1, 2, 3, 4];
let mut dest = [0; 4];
for i in 0..4 {
    dest[i] = src[i];
{
```
Clippy suggests `dest[..4].copy_from_slice(&src[..4]);`.
I reduced this suggestion as `dest.copy_from_slice(&src[..4]);`. (Removed needless indexing.)

changelog: [`manual_memcpy`]: Reduce indexing suggestions when array length is equal to loop range

### DIFF
--- a/clippy_lints/src/loops/manual_memcpy.rs
+++ b/clippy_lints/src/loops/manual_memcpy.rs
@@ -178,7 +178,9 @@ fn build_manual_memcpy_suggestion<'tcx>(
     let dst_base_str = snippet(cx, dst.base.span, "???");
     let src_base_str = snippet(cx, src.base.span, "???");
 
-    let dst = if dst_offset == sugg::EMPTY && dst_limit == sugg::EMPTY {
+    let dst = if (dst_offset == sugg::EMPTY && dst_limit == sugg::EMPTY)
+        || is_array_length_equal_to_range(cx, start, end, dst.base)
+    {
         dst_base_str
     } else {
         format!("{dst_base_str}[{}..{}]", dst_offset.maybe_par(), dst_limit.maybe_par()).into()
@@ -190,11 +192,13 @@ fn build_manual_memcpy_suggestion<'tcx>(
         "clone_from_slice"
     };
 
-    format!(
-        "{dst}.{method_str}(&{src_base_str}[{}..{}]);",
-        src_offset.maybe_par(),
-        src_limit.maybe_par()
-    )
+    let src = if is_array_length_equal_to_range(cx, start, end, src.base) {
+        src_base_str
+    } else {
+        format!("{src_base_str}[{}..{}]", src_offset.maybe_par(), src_limit.maybe_par()).into()
+    };
+
+    format!("{dst}.{method_str}(&{src});")
 }
 
 /// a wrapper of `Sugg`. Besides what `Sugg` do, this removes unnecessary `0`;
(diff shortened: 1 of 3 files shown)


clippy_lints/src/loops/manual_memcpy.rs
```diff
@@ -186,11 +189,13 @@ fn build_manual_memcpy_suggestion<'tcx>(
...
 
-    format!(
-        "{dst}.{method_str}(&{src_base_str}[{}..{}]);",
-        src_offset.maybe_par(),
-        src_limit.maybe_par()
-    )
+    let src = if is_array_length_equal_to_range(cx, start, end, src.base) {
+        src_base_str
+    } else {
+        format!("{src_base_str}[{}..{}]", src_offset.maybe_par(), src_limit.maybe_par()).into()
```
blyxyas, line 195: This `.into()` can be converted to a reference (If it's converting that `format!` to `&str`, which I think that it's doing).

```suggestion
        &format!("{src_base_str}[{}..{}]", src_offset.maybe_par(), src_limit.maybe_par())
```

blyxyas (changes requested):
Pretty good first iteration, just some minor fixes and I think this should be merge-ready!

clippy_lints/src/loops/manual_memcpy.rs
```diff
@@ -117,6 +118,41 @@ fn build_manual_memcpy_suggestion<'tcx>(
...
+        if let ty::Array(_, s) = arr_ty.kind() {
+            let size: u128 = if let Some(size) = s.try_eval_target_usize(cx.tcx, cx.param_env) {
+                size.into()
+            } else {
+                return false;
+            };
+
+            let range = match (extract_lit_value(start), extract_lit_value(end)) {
+                (Some(start_value), Some(end_value)) => {
+                    cmp::max(start_value, end_value) - cmp::min(start_value, end_value)
```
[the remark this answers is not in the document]
granddaifuku: Thank you for your review!
I remove `cmp` and use `end_value - start_value` to calculate the range.

clippy_lints/src/loops/manual_memcpy.rs
granddaifuku: I appreciate your review, but converting `format!("{src_base_str}[{}..{}]", src_offset.maybe_par(), src_limit.maybe_par()).into()` to `&format!("{src_base_str}[{}..{}]", src_offset.maybe_par(), src_limit.maybe_par())` does not work because it requires `Cow<'_, str>`, so I didn't apply the suggestion.
If you have other ideas, I'm open to making the necessary changes.

clippy_lints/src/loops/manual_memcpy.rs
```diff
@@ -446,3 +450,36 @@ fn get_loop_counters<'a, 'tcx>(
...
+                Some(value)
+            } else {
+                None
+            }
+        }
+    }
+
+    let arr_ty = cx.typeck_results().expr_ty(arr).peel_refs();
+
+    if let ty::Array(_, s) = arr_ty.kind() {
```
blyxyas, line 469: Is there any case in which this isn't an array? If so, we should make a first matching on `arr` to see if it's kind `ExprKind::Array` before querying the type (which is a more expensive function)

granddaifuku: I think the `arr` type could be a `std::vec::Vec<i32, std::alloc::Global>` or other types.
However, I'm actually not sure about how to classify `arr` type before querying it, as `src.base.kind` or `dst.base.kind` is `ExprKind::Path`

blyxyas: The indexing operation is its own kind, [`ExprKind::Index`](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_hir/hir/enum.ExprKind.html#variant.Index). Maybe there's something there?

If you cannot find any `ExprKind::Index` expression (which should be either `src` or `dst` without `.base`) it's fine, it can be accepted with a little performance oversight (but I'd really like for it to be pre-matched).

granddaifuku: I couldn't find any expressions of type `ExprKind::Index` in `src` or `dst`. 
I think the `src` and `dst` refer to `base_right` and `base_left` respectively, as provided in https://github.com/rust-lang/rust-clippy/blob/4e85e41e58832d1b5b8379dd508b4908e5562d2b/clippy_lints/src/loops/manual_memcpy.rs#L63-L66

and these are analyzed as elements similar to slices.

blyxyas: Okis, don't worry then.

blyxyas (approved):
LGTM, Thanks for you first contribution and welcome to Clippy! :heart:
Could you please squash these 4 commits into a single one?

granddaifuku:
It appears that the `if_chain` macro has been removed in #11750, leading to a CI failure. I will provide a fix for this issue.

granddaifuku:
I squashed my commits into one, and I really appreciate your guidance!

Issue #11775, opened 2023-11-08
Allow mod files within tests directory


TomFryersMidsummer:
### Summary

Within the `tests/` directory, `common/mod.rs` has an advantage over `common.rs` for shared code: it doesn't try to run `common.rs` as a test file, which would add noise to test runs. So, perhaps `mod_module_files` should exclude this case.

This Rust Book includes this recommendation: https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-tests.

### Lint Name

mod_module_files

### Reproducer

I tried this code:

```
├── Cargo.lock
├── Cargo.toml
├── src
│   └── lib.rs
└── tests
    ├── common
    │   └── mod.rs
    └── test.rs
```
with `mod common;` in `test.rs`. Other files are as generated by `cargo new`, or empty.
```sh
cargo clippy --tests -- -W clippy::mod-module-files
```
I saw this happen:

```
warning: `mod.rs` files are not allowed, found `tests/common/mod.rs`
 --> tests/common/mod.rs:0:1
  |
  |
  = help: move `tests/common/mod.rs` to `tests/common.rs`
  = note: requested on the command line with `-W clippy::mod-module-files`
```

I expected to see this happen:

[no error]

### Version

```text
rustc 1.73.0 (cc66ad468 2023-10-03)
binary: rustc
commit-hash: cc66ad468955717ab92600c770da8c1601a4ff33
commit-date: 2023-10-03
host: x86_64-unknown-linux-gnu
release: 1.73.0
LLVM version: 17.0.2
```

miguelpduarte:
I have also faced this issue. Versions:
```
clippy 0.1.88 (6b00bc3880 2025-06-23)
rustc 1.88.0 (6b00bc388 2025-06-23)
```

This seems like the offending section: https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/module_style.rs#L153

For example for the workspace I'm working in, paths can be `crates/<some-crate>/tests/common/mod.rs`

Since the path starts with `crates/<some-crate>` instead of just `tests`, `!path.starts_with("tests")` is not fulfilled.

I guess it might make sense to either reopen this, or create a separate issue for tracking this specifically for workspace crates.

PR #11779, opened 2023-11-08
[`mod_module_files`] Don't emit lint for mod.rs in tests

partiallytyped:
fixes: #11775

current state: indiscriminately emits the lint for mod files in tests.

The following

```
tests/
  common/
    mod.rs
  test.rs
```

is a common pattern for code shared across the tests and is suggested in the rust book. The change adds an additional check to verify that the mod file is not in tests.

changelog: Fix [`mod_module_files`]: false positive for mod files in tests folder

### DIFF
--- a/clippy_lints/src/module_style.rs
+++ b/clippy_lints/src/module_style.rs
@@ -153,8 +153,11 @@ fn process_paths_for_mod_files<'a>(
 }
 
 /// Checks every path for the presence of `mod.rs` files and emits the lint if found.
+/// We should not emit a lint for test modules in the presence of `mod.rs`.
+/// Using `mod.rs` in integration tests is a [common pattern](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-test)
+/// for code-sharing between tests.
 fn check_self_named_mod_exists(cx: &EarlyContext<'_>, path: &Path, file: &SourceFile) {
-    if path.ends_with("mod.rs") {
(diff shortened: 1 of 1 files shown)


clippy_lints/src/module_style.rs
```diff
@@ -153,8 +153,11 @@ fn process_paths_for_mod_files<'a>(
 }
 
 /// Checks every path for the presence of `mod.rs` files and emits the lint if found.
+/// We should not emit a lint for test modules in the presence of `mod.rs`.
+/// Using `mod.rs` in integration tests is a [common pattern](https://doc.rust-lang.org/book/ch11-03-test-organization.html#submodules-in-integration-test)
+/// for code-sharing between tests.
 fn check_self_named_mod_exists(cx: &EarlyContext<'_>, path: &Path, file: &SourceFile) {
-    if path.ends_with("mod.rs") {
+    if path.ends_with("mod.rs") && !path.starts_with("tests") {
```
blyxyas, line 160: There's a similar pattern on line 148, would it also need this check?

partiallytyped: I don't think it is needed. The pattern above collects the paths to the folders to be enforced when the old style is used.

The pattern here emits an error when the new style is used in the code but within the tests folder the old style with `mod.rs` is used.

I created a structure as in #11775 and the lint wasn't emitted.

blyxyas (approved):
LGTM, thanks! :heart:

PR #11780, opened 2023-11-08
Disable `vec_box` when using different allocators

### LINKED ISSUE #7114: `vec_box` lint should not fire if the allocators between the `Vec` and `Box` are different
As the title suggests, now that there are custom allocators available to `Box` and `Vec`, the lint should take those into account when deciding to suggest `vec_box`. I hit this when I had a `Vec<Box<MyType, MyTypeAllocator>>` where it was not valid to use a `Vec<MyType, MyTypeAllocator>`.

### Meta

- `cargo clippy -V`: clippy 0.1.53 (d0695c9 2021-04-12)
- `rustc -Vv`:
```
rustc 1.53.0-nightly (d0695c908 2021-04-12)
binary: rustc
commit-hash: d0695c9081b16077d0aed368bccaf437d77ff497
commit-date: 2021-04-12
release: 1.53.0-nightly
LLVM version: 12.0.0
```

Jacherr:
Fixes #7114 

This PR disables the `vec_box` lint when the `Box` and `Vec` use different allocators (but not when they use the same - custom - allocator).

For example - `Vec<Box<i32, DummyAllocator>>` will disable the lint, and `Vec<Box<i32, DummyAllocator>, DummyAllocator>` will not disable the lint. 

In addition, the applicability of this lint has been changed to `Unspecified` due to the automatic fixes potentially breaking code such as the following:

```rs
fn foo() -> Vec<Box<i32>> { // -> Vec<i32>
  vec![Box::new(1)]
}
``` 

It should be noted that the `if_chain->let-chains` fix has also been applied to this lint, so the diff does contain many changes.

changelog: disable `vec_box` lint when using nonstandard allocators


tests/ui/vec_box_sized.rs
```diff
@@ -33,6 +59,10 @@ mod should_not_trigger {
         // Regression test for #3720. This was causing an ICE.
         inner: Vec<Box<T>>,
     }
+
+    fn allocator_mismatch() -> Vec<Box<i32, DummyAllocator>> {
+        Vec::new()
+    }
```
xFrednet, line 65: Could you add a second test, where the allocator of the vector mismatches, like his:
```
    fn allocator_mismatch() -> Vec<Box<i32>, DummyAllocator> {
        Vec::new()
    }
```

xFrednet (approved):
Welcome to Clippy :wave:. The changes look good to me, there is one small additional tests I would like to see and then we can merge this :D

xFrednet (approved):

xFrednet:
Looks good to me, thank you for the swift update :D

@bors r+

PR #11781, opened 2023-11-09
Verify Borrow<T> semantics for types that implement Hash, Borrow<str> and Borrow<[u8]>.

### LINKED ISSUE #11710: A lint to detect when a type implements all three of `Hash`, `Borrow<str>` and `Borrow<[u8]>`
### What it does

Detects when a type which implements `Hash` also implements both `Borrow<str>` and `Borrow<[u8]>`.

### Advantage

It's impossible to implement both of these correctly, as `str` and `[u8]` hash differently in incompatible ways, and implementing `Borrow<T>` for your type guaratees that you hash equivalent to `T`.

This is worth adding because this impl is tempting to add, as the distinction between `Borrow` and `AsRef` is very subtle, and the `AsRef` impl for both is correct (as is implementing both for types which do not implement `Hash`).

See https://github.com/BurntSushi/bstr/issues/168 for some discussion (As this would not *quite* have caught the problem in that issue, it's possible my formulation of it is too narrow. For the most part, I am chucking this lint proposal over the wall, and am not tied to the specifics here).

### Drawbacks

It's rather niche.

### Example

```rust
impl Hash for MyType {
    // ...
}

impl Borrow<str> for MyType {
    // ...
}

impl Borrow<[u8]> for MyType {
    // ...
}
```

is wrong, either remove any one of those impls to be correct.

partiallytyped:
Fixes #11710 

The essence of the issue is that types that implement Borrow<T> provide a facet or a representation of the underlying type. Under these semantics `hash(a) == hash(a.borrow())`.

This is a problem when a type implements `Borrow<str>`, `Borrow<[u8]>` and Hash, it is expected that the hash of all three types is identical. The problem is that the hash of [u8] is not the same as that of a String, even when the byte reference ([u8]) is derived from `.as_bytes()`

- [x] Run `cargo dev fmt`

---

 - [x] Explanation of the issue in the code
 - [x] Tests reproducing the issue
 - [x] Lint rule and emission

---

changelog: New lint: [`impl_hash_borrow_with_str_and_bytes`]
[#11781](https://github.com/rust-lang/rust-clippy/pull/11781)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5128,6 +5128,7 @@ Released 2018-09-13
 [`if_then_some_else_none`]: https://rust-lang.github.io/rust-clippy/master/index.html#if_then_some_else_none
 [`ifs_same_cond`]: https://rust-lang.github.io/rust-clippy/master/index.html#ifs_same_cond
 [`ignored_unit_patterns`]: https://rust-lang.github.io/rust-clippy/master/index.html#ignored_unit_patterns
+[`impl_hash_borrow_with_str_and_bytes`]: https://rust-lang.github.io/rust-clippy/master/index.html#impl_hash_borrow_with_str_and_bytes
 [`impl_trait_in_params`]: https://rust-lang.github.io/rust-clippy/master/index.html#impl_trait_in_params
 [`implicit_clone`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_clone
 [`implicit_hasher`]: https://rust-lang.github.io/rust-clippy/master/index.html#implicit_hasher
--- a/clippy_lints/src/declared_lints.rs
+++ b/clippy_lints/src/declared_lints.rs
@@ -204,6 +204,7 @@ pub(crate) static LINTS: &[&crate::LintInfo] = &[
     crate::if_not_else::IF_NOT_ELSE_INFO,
     crate::if_then_some_else_none::IF_THEN_SOME_ELSE_NONE_INFO,
     crate::ignored_unit_patterns::IGNORED_UNIT_PATTERNS_INFO,
+    crate::impl_hash_with_borrow_str_and_bytes::IMPL_HASH_BORROW_WITH_STR_AND_BYTES_INFO,
     crate::implicit_hasher::IMPLICIT_HASHER_INFO,
     crate::implicit_return::IMPLICIT_RETURN_INFO,
     crate::implicit_saturating_add::IMPLICIT_SATURATING_ADD_INFO,
--- a/clippy_lints/src/impl_hash_with_borrow_str_and_bytes.rs
+++ b/clippy_lints/src/impl_hash_with_borrow_str_and_bytes.rs
@@ -1,0 +1,106 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::ty::implements_trait;
+use rustc_hir::def::{DefKind, Res};
+use rustc_hir::{Item, ItemKind, Path, TraitRef};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    ///
+    /// This lint is concerned with the semantics of `Borrow` and `Hash` for a
+    /// type that implements all three of `Hash`, `Borrow<str>` and `Borrow<[u8]>`
+    /// as it is impossible to satisfy the semantics of Borrow and `Hash` for
+    /// both `Borrow<str>` and `Borrow<[u8]>`.
+    ///
+    /// ### Why is this bad?
+    ///
+    /// When providing implementations for `Borrow<T>`, one should consider whether the different
+    /// implementations should act as facets or representations of the underlying type. Generic code
+    /// typically uses `Borrow<T>` when it relies on the identical behavior of these additional trait
+    /// implementations. These traits will likely appear as additional trait bounds.
+    ///
+    /// In particular `Eq`, `Ord` and `Hash` must be equivalent for borrowed and owned values:
+    /// `x.borrow() == y.borrow()` should give the same result as `x == y`.
+    /// It follows then that the following equivalence must hold:
+    /// `hash(x) == hash((x as Borrow<[u8]>).borrow()) == hash((x as Borrow<str>).borrow())`
+    ///
+    /// Unfortunately it doesn't hold as `hash("abc") != hash("abc".as_bytes())`.
+    /// This happens because the `Hash` impl for str passes an additional `0xFF` byte to
+    /// the hasher to avoid collisions. For example, given the tuples `("a", "bc")`, and `("ab", "c")`,
+    /// the two tuples would have the same hash value if the `0xFF` byte was not added.
+    ///
+    /// ### Example
+    ///
+    /// ```
+    /// use std::borrow::Borrow;
+    /// use std::hash::{Hash, Hasher};
+    ///
+    /// struct ExampleType {
+    ///     data: String
+    /// }
+    ///
+    /// impl Hash for ExampleType {
+    ///     fn hash<H: Hasher>(&self, state: &mut H) {
+    ///         self.data.hash(state);
+    ///     }
+    /// }
+    ///
+    /// impl Borrow<str> for ExampleType {
+    ///     fn borrow(&self) -> &str {
+    ///         &self.data
+    ///     }
+    /// }
+    ///
+    /// impl Borrow<[u8]> for ExampleType {
+    ///     fn borrow(&self) -> &[u8] {
+    ///         self.data.as_bytes()
+    ///     }
+    /// }
+    /// ```
+    /// As a consequence, hashing a `&ExampleType` and hashing the result of the two
+    /// borrows will result in different values.
+    ///
+    #[clippy::version = "1.76.0"]
+    pub IMPL_HASH_BORROW_WITH_STR_AND_BYTES,
+    correctness,
+    "ensures that the semantics of `Borrow` for `Hash` are satisfied when `Borrow<str>` and `Borrow<[u8]>` are implemented"
+}
+
+declare_lint_pass!(ImplHashWithBorrowStrBytes => [IMPL_HASH_BORROW_WITH_STR_AND_BYTES]);
+
+impl LateLintPass<'_> for ImplHashWithBorrowStrBytes {
+    /// We are emitting this lint at the Hash impl of a type that implements all
+    /// three of `Hash`, `Borrow<str>` and `Borrow<[u8]>`.
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if let ItemKind::Impl(imp) = item.kind
+            && let Some(TraitRef {path: Path {span, res, ..}, ..}) = imp.of_trait
+            && let ty = cx.tcx.type_of(item.owner_id).instantiate_identity()
+            && let Some(hash_id) = cx.tcx.get_diagnostic_item(sym::Hash)
+            && Res::Def(DefKind::Trait, hash_id) == *res
+            && let Some(borrow_id) = cx.tcx.get_diagnostic_item(sym::Borrow)
+            // since we are in the `Hash` impl, we don't need to check for that.
+            // we need only to check for `Borrow<str>` and `Borrow<[u8]>`
+            && implements_trait(cx, ty, borrow_id, &[cx.tcx.types.str_.into()])
+            && implements_trait(cx, ty, borrow_id, &[Ty::new_slice(cx.tcx, cx.tcx.types.u8).into()])
+        {
+            span_lint_and_then(
+                cx,
+                IMPL_HASH_BORROW_WITH_STR_AND_BYTES,
+                *span,
+                "the semantics of `Borrow<T>` around `Hash` can't be satisfied when both `Borrow<str>` and `Borrow<[u8]>` are implemented",
+                |diag| {
+                    diag.note("the `Borrow` semantics require that `Hash` must behave the same for all implementations of Borrow<T>");
+                    diag.note(
+          "however, the hash implementations of a string (`str`) and the bytes of a string `[u8]` do not behave the same ..."
+      );
+                    diag.note("... as (`hash(\"abc\") != hash(\"abc\".as_bytes())`");
+                    diag.help("consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...");
+                    diag.help("... or not implementing `Hash` for this type");
+                },
+            );
+        }
+    }
+}
--- a/clippy_lints/src/lib.rs
+++ b/clippy_lints/src/lib.rs
@@ -144,6 +144,7 @@ mod if_let_mutex;
 mod if_not_else;
 mod if_then_some_else_none;
 mod ignored_unit_patterns;
+mod impl_hash_with_borrow_str_and_bytes;
 mod implicit_hasher;
 mod implicit_return;
 mod implicit_saturating_add;
@@ -1066,6 +1067,7 @@ pub fn register_lints(store: &mut rustc_lint::LintStore, conf: &'static Conf) {
     store.register_late_pass(move |_| Box::new(manual_hash_one::ManualHashOne::new(msrv())));
     store.register_late_pass(|_| Box::new(iter_without_into_iter::IterWithoutIntoIter));
     store.register_late_pass(|_| Box::new(iter_over_hash_type::IterOverHashType));
+    store.register_late_pass(|_| Box::new(impl_hash_with_borrow_str_and_bytes::ImplHashWithBorrowStrBytes));
     // add lints here, do not remove this comment, it's used in `new_lint`
 }
 
--- a/tests/ui/impl_hash_with_borrow_str_and_bytes.rs
+++ b/tests/ui/impl_hash_with_borrow_str_and_bytes.rs
@@ -1,0 +1,136 @@
+#![warn(clippy::impl_hash_borrow_with_str_and_bytes)]
+
+use std::borrow::Borrow;
+use std::hash::{Hash, Hasher};
+
+struct ExampleType {
+    data: String,
+}
+
+impl Hash for ExampleType {
+    //~^ ERROR: can't
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.data.hash(state);
+    }
+}
+
+impl Borrow<str> for ExampleType {
+    fn borrow(&self) -> &str {
+        &self.data
+    }
+}
+
+impl Borrow<[u8]> for ExampleType {
+    fn borrow(&self) -> &[u8] {
+        self.data.as_bytes()
+    }
+}
+
+struct ShouldNotRaiseForHash {}
+impl Hash for ShouldNotRaiseForHash {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        todo!();
+    }
+}
+
+struct ShouldNotRaiseForBorrow {}
+impl Borrow<str> for ShouldNotRaiseForBorrow {
+    fn borrow(&self) -> &str {
+        todo!();
+    }
+}
+impl Borrow<[u8]> for ShouldNotRaiseForBorrow {
+    fn borrow(&self) -> &[u8] {
+        todo!();
+    }
+}
+
+struct ShouldNotRaiseForHashBorrowStr {}
+impl Hash for ShouldNotRaiseForHashBorrowStr {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        todo!();
+    }
+}
+impl Borrow<str> for ShouldNotRaiseForHashBorrowStr {
+    fn borrow(&self) -> &str {
+        todo!();
+    }
+}
+
+struct ShouldNotRaiseForHashBorrowSlice {}
+impl Hash for ShouldNotRaiseForHashBorrowSlice {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        todo!();
+    }
+}
+
+impl Borrow<[u8]> for ShouldNotRaiseForHashBorrowSlice {
+    fn borrow(&self) -> &[u8] {
+        todo!();
+    }
+}
+
+#[derive(Hash)]
+//~^ ERROR: can't
+struct Derived {
+    data: String,
+}
+
+impl Borrow<str> for Derived {
+    fn borrow(&self) -> &str {
+        self.data.as_str()
+    }
+}
+
+impl Borrow<[u8]> for Derived {
+    fn borrow(&self) -> &[u8] {
+        self.data.as_bytes()
+    }
+}
+
+struct GenericExampleType<T> {
+    data: T,
+}
+
+impl<T: Hash> Hash for GenericExampleType<T> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.data.hash(state);
+    }
+}
+
+impl Borrow<str> for GenericExampleType<String> {
+    fn borrow(&self) -> &str {
+        &self.data
+    }
+}
+
+impl Borrow<[u8]> for GenericExampleType<&'static [u8]> {
+    fn borrow(&self) -> &[u8] {
+        self.data
+    }
+}
+
+struct GenericExampleType2<T> {
+    data: T,
+}
+
+impl Hash for GenericExampleType2<String> {
+    //~^ ERROR: can't
+    // this is correctly throwing an error for generic with concrete impl
+    // for all 3 types
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.data.hash(state);
+    }
+}
+
+impl Borrow<str> for GenericExampleType2<String> {
+    fn borrow(&self) -> &str {
+        &self.data
+    }
+}
+
+impl Borrow<[u8]> for GenericExampleType2<String> {
+    fn borrow(&self) -> &[u8] {
+        self.data.as_bytes()
+    }
+}
--- a/tests/ui/impl_hash_with_borrow_str_and_bytes.stderr
+++ b/tests/ui/impl_hash_with_borrow_str_and_bytes.stderr
@@ -1,0 +1,41 @@
+error: the semantics of `Borrow<T>` around `Hash` can't be satisfied when both `Borrow<str>` and `Borrow<[u8]>` are implemented
+  --> $DIR/impl_hash_with_borrow_str_and_bytes.rs:10:6
+   |
+LL | impl Hash for ExampleType {
+   |      ^^^^
+   |
+   = note: the `Borrow` semantics require that `Hash` must behave the same for all implementations of Borrow<T>
+   = note: however, the hash implementations of a string (`str`) and the bytes of a string `[u8]` do not behave the same ...
+   = note: ... as (`hash("abc") != hash("abc".as_bytes())`
+   = help: consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...
+   = help: ... or not implementing `Hash` for this type
+   = note: `-D clippy::impl-hash-borrow-with-str-and-bytes` implied by `-D warnings`
+   = help: to override `-D warnings` add `#[allow(clippy::impl_hash_borrow_with_str_and_bytes)]`
+
+error: the semantics of `Borrow<T>` around `Hash` can't be satisfied when both `Borrow<str>` and `Borrow<[u8]>` are implemented
+  --> $DIR/impl_hash_with_borrow_str_and_bytes.rs:73:10
+   |
+LL | #[derive(Hash)]
+   |          ^^^^
+   |
+   = note: the `Borrow` semantics require that `Hash` must behave the same for all implementations of Borrow<T>
+   = note: however, the hash implementations of a string (`str`) and the bytes of a string `[u8]` do not behave the same ...
+   = note: ... as (`hash("abc") != hash("abc".as_bytes())`
+   = help: consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...
+   = help: ... or not implementing `Hash` for this type
+   = note: this error originates in the derive macro `Hash` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+error: the semantics of `Borrow<T>` around `Hash` can't be satisfied when both `Borrow<str>` and `Borrow<[u8]>` are implemented
+  --> $DIR/impl_hash_with_borrow_str_and_bytes.rs:117:6
+   |
+LL | impl Hash for GenericExampleType2<String> {
+   |      ^^^^
+   |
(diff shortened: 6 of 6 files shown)


clippy_lints/src/types/hash_borrow_str_soundness.rs
```diff
@@ -0,0 +1,38 @@
...
+/// implementations should act as facets or representations of the underlying type. Generic code
+/// typically uses Borrow<T> when it relies on the identical behavior of these additional trait
+/// implementations. These traits will likely appear as additional trait bounds.
+///
+/// In particular Eq, Ord and Hash must be equivalent for borrowed and owned values:
+/// `x.borrow() == y.borrow()` should give the same result as `x == y`.
+///
+/// It follows then that the following equivalence must hold:
+/// `hash(x) == hash(x.borrow<str>()) == hash(x.borrow<[u8]>())`,
+/// however, `hash("123".as_str()) != hash(x.borrow<[u8]>()`
```
partiallytyped, line 31: This needs some touch ups.

partiallytyped:
Okay, so as far as I understand this needs to be a latelint.

We could do this on the early pass and check for impl of all three, but that could cause issues if there are name conflicts.

e.g. Two Structs with the same name in different modules, both impl Hash and Borrow, except one is Borrow<str> and the other is Borrow<[u8]>.

So instead we do a latelint, and check that for the type we don't have both impls.
Does this make sense?

clippy_lints/src/hash_borrow_str_soundness.rs
```diff
@@ -0,0 +1,84 @@
...
+            // if Res::Def(DefKind::Trait, hash_id) == *res;
+            if let Some(borrow_id) = cx.tcx.get_diagnostic_item(sym::Borrow);
+            if implements_trait(cx, ty, hash_id, &[]);
+            if implements_trait(cx, ty, borrow_id, &[
+                cx.tcx.types.str_.into()
+            ]);
+            // if implements_trait(cx, ty, borrow_id, &[
+                // cx.tcx.types.slice(cx.tcx.types.u8).into()
+            // ]);
+            // now we need only to see that it implements Borrow<str> and Borrow<[u8]>
```
partiallytyped, line 67: Slice doesn't exist in `cx.txc.types`, I am kinda stuck now 😅.

xFrednet:
Welcome to Clippy and thank you for the PR. I didn't have time to look at this until now, but it seems like you found some excellent help on Zulip, which is awesome.

@y21 would you like to take over the review, since you already helped on Zulip? Once you're done, I can do the final review and r+ for us two :)

y21:
I see this is still in draft mode. I can give some feedback now, or do you want me to wait until this is marked ready?

xFrednet:
> I can give some feedback now, or do you want me to wait until this is marked ready?

This usually depends on the author of the PR. Some use draft mode, when they are still working on the PR and some use it, to request feedback. @partiallytyped posted some questions here, which makes me guess that they wanted to get feedback on their questions. However, they have also asked them on Zulip where you have been helpful. In this case, I would monitor the PR or ask the author if they would already like to have a review: :)

partiallytyped:
I think we can go ahead with the review while I do some finishing touch ups on the notes :D

partiallytyped:
I think this issue pretty much opens a whole branch of checks. This one is specific to `Borrow<[u8]>` and `Borrow<str>`, but in general types that impl `Borrow<T>` and `Hash` must have identical hashing behaviour.

clippy_lints/src/hash_borrow_str_soundness.rs
```diff
@@ -0,0 +1,84 @@
...
+declare_lint_pass!(HashBorrowStrSoundness => [HASH_BORROW_STR_SOUNDNESS]);
+
+impl LateLintPass<'_> for HashBorrowStrSoundness {
+    /// We are emitting this lint at the Hash impl of a type that implements all
+    /// three of Hash, Borrow<str> and Borrow<[u8]>.
+    ///
+    /// We check that we are in the Hash impl
+    /// Then we check that the type implements Borrow<str> and Borrow<[u8]>
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if_chain! {
```
y21, line 54: We've been moving away from the `if_chain!` macro, over to let chains, since it provides better DX (rustfmt doesn't support macros) and we'd also like to test the feature in general. Can this be changed?

clippy_lints/src/hash_borrow_str_soundness.rs
```diff
@@ -0,0 +1,82 @@
...
+            // we need only to check for Borrow<str> and Borrow<[u8]>
+            if implements_trait(cx, ty, borrow_id, &[cx.tcx.types.str_.into()]);
+            if implements_trait(cx, ty, borrow_id, &[Ty::new_slice(cx.tcx, cx.tcx.types.u8).into()]);
+
+            then {
+                span_lint_and_then(
+                    cx,
+                    HASH_BORROW_STR_SOUNDNESS,
+                    *span,
+                    "implementation of Hash for type that implements Borrow<str> and Borrow<[u8]> is unsound",
```
y21, line 71: "unsound" has a very specific meaning in Rust and it's concerned with memory safety.
Many Rust devs associate "unsound" with [the ability to cause Undefined Behavior in safe code](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#soundness-of-code--of-a-library).
This lint can be triggered in safe code and safe code cannot be unsound.

Unsafe code cannot rely on the correctness of `Hash` or `Borrow` traits, even if it's non-sensical.
I think it would be better to reword this.

clippy_lints/src/hash_borrow_str_soundness.rs
```diff
@@ -0,0 +1,82 @@
...
+use clippy_utils::ty::implements_trait;
+use rustc_hir::def::{DefKind, Res};
+use rustc_hir::{Item, ItemKind, Path, TraitRef};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    ///
```
y21, line 12:
```suggestion
    /// ### What it does
```
For consistency with all other lints

clippy_lints/src/hash_borrow_str_soundness.rs
```diff
@@ -0,0 +1,82 @@
...
+    ///  This lint is concerned with the soundness of types that implement traits
+    ///  Hash, Borrow<u8> and Borrow<str>:
+    ///  ```
+    /// impl Hash for MyType {
```
y21, line 16: I would maybe move this example to its own `### Example` section.

y21 (commented):
Some initial feedback :)
(Sorry for the chaotic order, had to jump around a bit in the code)

clippy_lints/src/hash_borrow_str_soundness.rs
partiallytyped: Sure yeah. Already done! I just saw the commit removing `if_chain` and I was wondering why my build was, well, not building :D

clippy_lints/src/hash_borrow_str_soundness.rs
partiallytyped: Maybe the example is a bit redundant?

clippy_lints/src/hash_borrow_str_soundness.rs
partiallytyped: I have cleaned up the explanation and added the header. I think the new one is better.

y21: Yeah, looks good to me now.

clippy_lints/src/hash_borrow_str_soundness.rs
y21: I'd keep it, It can hurt to have it imo. Though it's currently failing in CI. You can add `ignore` to the codeblock, since it looks like a visual example.

partiallytyped: I made the example complete to show that it breaks. I think that's better.

partiallytyped:
Rebase and forcepush added all the commits from whence I began. Hope that's fine.

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,114 @@
...
+    ///     }
+    /// }
+    ///
+    /// impl Borrow<[u8]> for ExampleType {
+    ///     fn borrow(&self) -> &[u8] {
+    ///         self.data.as_bytes()
+    ///     }
+    /// }
+    ///
+    /// fn hash(x: impl Hash) -> u64 {
```
y21, line 59: There's a lot of detail in the example that I think can be removed, such as this hash function and the main function. How about trimming this down to some `struct MyType;` and then those three impls?
If you do want to include a full "PoC" kind of example, perhaps it could also have a link to the a gist on the Rust playground.

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,114 @@
...
+            // since we are in the Hash impl, we don't need to check for that.
+            // we need only to check for Borrow<str> and Borrow<[u8]>
+            implements_trait(cx, ty, borrow_id, &[cx.tcx.types.str_.into()]) &&
+            implements_trait(cx, ty, borrow_id, &[Ty::new_slice(cx.tcx, cx.tcx.types.u8).into()])
+        {
+            span_lint_and_then(
+                cx,
+                HASH_BORROW_STR_SEMANTICS,
+                *span,
+                "can't satisfy the semantics of Borrow for Borrow<str> and Borrow<[u8]> for Hash",
```
y21, line 103:
```suggestion
                "can't satisfy the semantics of `Hash` when both `Borrow<str>` and `Borrow<[u8]>` are implemented",
```
I feel like phrasing it like this makes it clearer, considering that this lint is primarily looking for the `Hash` impl and uses that span.

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,114 @@
...
+declare_lint_pass!(HashBorrowStrSemantics => [HASH_BORROW_STR_SEMANTICS]);
+
+impl LateLintPass<'_> for HashBorrowStrSemantics {
+    /// We are emitting this lint at the Hash impl of a type that implements all
+    /// three of Hash, Borrow<str> and Borrow<[u8]>.
+    ///
+    /// We check that we are in the Hash impl
+    /// Then we check that the type implements Borrow<str> and Borrow<[u8]>
+    fn check_item(&mut self, cx: &LateContext<'_>, item: &Item<'_>) {
+        if let ItemKind::Impl(imp) = item.kind &&
```
y21, line 88: Nit (I'm surprised rustfmt didn't pick this up?): The `&&` should normally go on a new line. See e.g. here

https://github.com/rust-lang/rust-clippy/blob/d487579efd3c6b7766f820bbe4b6be0f6d420dad/clippy_lints/src/attrs.rs#L661-L663

y21 (commented):
Last few things from my side and then I think it looks good to me!

y21:
Does this also work when the `Hash` impl is derived? I think it's worth having a test for that, too.

clippy_lints/src/hash_borrow_str_semantics.rs
partiallytyped: I agree, but then `cargo dev fmt` pushes the line down in the ui test 😓

clippy_lints/src/hash_borrow_str_semantics.rs
partiallytyped: Good idea!

partiallytyped:
Surprisingly it does! :D

clippy_lints/src/hash_borrow_str_semantics.rs
y21: >  I removed the line from the error check, I think it's fine to rely on just the ui test, right?

They aren't required yet, but I believe the idea is that soon it will be.

I assume this a problem because of the `//~^ ERROR:` pattern?
At least in rustc's test suite, IIRC, the annotation can be split up to multiple lines by using `//~|` on the next line and it inherits the level.

If that doesn't work, there's also the option to simply only include as many words of the error message as it fits in one line.
uitest only checks if the error message starts with the pattern.

Or, I guess, we can also just not specify it like you have it right now and make it the problem of whoever wants to make them required 😄 

@xFrednet do you know what the best solution to this problem is?

partiallytyped:
That’s exactly it, the problem is that pattern.

I think we should maybe have all these and the issues I faced earlier into some kind of guide for newbies like me :D

clippy_lints/src/hash_borrow_str_semantics.rs
partiallytyped: Yes it's the //~^ pattern :D

y21:
Looks like there are still merge commits somehow, according to rustbot. I think you can just rebase onto master to fix that.

partiallytyped:
This should do it.

clippy_lints/src/hash_borrow_str_semantics.rs
xFrednet: Interesting, no I don't know the best solution. I think, I would prefer the `//~^ ERROR:` for lints with a static message. We had a discussion, to maybe add an option to uitest, that allows us to specify the lint name instead of the message, but I believe this hasn't been implemented yet. Just having the `//~^ ERROR:` without a message SGTM :)

partiallytyped: I tried an empty `//~^ ERROR:` and it failed the bless check. Instead I used `//~^ ERROR: can't` and that seems to be okay.

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            span_lint_and_then(
+                cx,
+                HASH_BORROW_STR_SEMANTICS,
+                *span,
+                "can't satisfy the semantics of `Hash` when both `Borrow<str>` and `Borrow<[u8]>` are implemented",
+                |diag| {
+                    diag.note("types that implement Hash and Borrow<T> should have equivalent Hash results for borrowed and owned values");
+
+                    diag.note("this is not the case for Borrow<str> and Borrow<[u8]> as the two types result in different Hash values");
```
y21, line 96: I feel like these two notes could be one. The first one reads a little confusing since it seems unrelated to the situation of having two conflicting `Borrow` impls.
Maybe something along the lines of "`Borrow` requires `Hash` values of all borrowed types to be the same, which is not the case for `Borrow<str>` and `Borrow<[u8]>`?

tests/ui/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+impl Borrow<[u8]> for ShouldNotRaiseForHashBorrowSlice {
+    fn borrow(&self) -> &[u8] {
+        todo!();
+    }
+}
+
+#[derive(Hash)]
```
y21, line 73: If we use error annotations, we should also have one here :)

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,103 @@
...
+use rustc_hir::{Item, ItemKind, Path, TraitRef};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint is concerned with the semantics of Borrow and Hash for a
+    /// type that implements all three of Hash, Borrow<str> and Borrow<[u8]>
```
y21, line 13: Can you put the trait references between ticks, like `Hash`? There's a lint for this, but it doesn't seem smart enough to trigger here

y21 (commented):
Thought about it for a bit and I left some last comments. Most of them are just nits or non-blocking questions/possible improvements though 🙃

tests/ui/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,88 @@
...
+struct Derived {
```
partiallytyped, line 74:
```suggestion
//~^ ERROR: can't
```

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// typically uses Borrow<T> when it relies on the identical behavior of these additional trait
+    /// implementations. These traits will likely appear as additional trait bounds.
+    ///
+    /// In particular Eq, Ord and Hash must be equivalent for borrowed and owned values:
+    /// `x.borrow() == y.borrow()` should give the same result as `x == y`.
+    /// It follows then that the following equivalence must hold:
+    /// `hash(x) == hash((x as Borrow<[u8]>).borrow()) == hash((x as Borrow<str>).borrow())`
+    ///
+    /// Unfortunately it doesn't hold as `hash("abc") != hash("abc".as_bytes())`.
+    ///
```
partiallytyped, line 30:
```suggestion
    /// this is because the that Hash impl for str passes an additional 0xFF byte to
    /// the hasher to avoid collisions. For example, given a tuple ("a", "bc"), and ("ab", "c"),
    /// the two tuples would have the same hash value if the 0xFF byte was not added.
    /// https://doc.rust-lang.org/stable/core/hash/trait.Hash.html#prefix-collisions
```

clippy_lints/src/hash_borrow_str_semantics.rs
partiallytyped: That sounds better, yes!

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+declare_lint_pass!(HashBorrowStrSemantics => [HASH_BORROW_STR_SEMANTICS]);
+
+impl LateLintPass<'_> for HashBorrowStrSemantics {
+    /// We are emitting this lint at the Hash impl of a type that implements all
+    /// three of Hash, Borrow<str> and Borrow<[u8]>.
+    ///
+    /// We check that we are in the Hash impl
+    /// Then we check that the type implements Borrow<str> and Borrow<[u8]>
```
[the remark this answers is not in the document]
partiallytyped: How'd that get in here 😅

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,103 @@
...
+        if let ItemKind::Impl(imp) = item.kind
+            && let Some(TraitRef {path: Path {span, res, ..}, ..}) = imp.of_trait
+            && let ty = cx.tcx.type_of(item.owner_id).instantiate_identity()
+            && let Some(hash_id) = cx.tcx.get_diagnostic_item(sym::Hash)
+            && Res::Def(DefKind::Trait, hash_id) == *res
+            && let Some(borrow_id) = cx.tcx.get_diagnostic_item(sym::Borrow)
+            // since we are in the Hash impl, we don't need to check for that.
+            // we need only to check for Borrow<str> and Borrow<[u8]>
+            && implements_trait(cx, ty, borrow_id, &[cx.tcx.types.str_.into()])
+            &&implements_trait(cx, ty, borrow_id, &[Ty::new_slice(cx.tcx, cx.tcx.types.u8).into()])
```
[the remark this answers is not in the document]
partiallytyped: Maybe in a follow up when I get more comfortable with this ? :D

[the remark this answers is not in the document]
y21: That's fine :)

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// implementations should act as facets or representations of the underlying type. Generic code
+    /// typically uses Borrow<T> when it relies on the identical behavior of these additional trait
+    /// implementations. These traits will likely appear as additional trait bounds.
+    ///
+    /// In particular Eq, Ord and Hash must be equivalent for borrowed and owned values:
+    /// `x.borrow() == y.borrow()` should give the same result as `x == y`.
+    /// It follows then that the following equivalence must hold:
+    /// `hash(x) == hash((x as Borrow<[u8]>).borrow()) == hash((x as Borrow<str>).borrow())`
+    ///
+    /// Unfortunately it doesn't hold as `hash("abc") != hash("abc".as_bytes())`.
```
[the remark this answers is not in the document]
partiallytyped:
```suggestion
    /// this is because the that Hash impl for str passes an additional 0xFF byte to
    /// the hasher to avoid collisions. For example, given the tuples ("a", "bc"), and ("ab", "c"),
    /// the two tuples would have the same hash value if the 0xFF byte was not added.
    /// https://doc.rust-lang.org/stable/core/hash/trait.Hash.html#prefix-collisions
```

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,105 @@
...
+use rustc_hir::{Item, ItemKind, Path, TraitRef};
+use rustc_lint::{LateContext, LateLintPass};
+use rustc_middle::ty::Ty;
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::symbol::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This lint is concerned with the semantics of Borrow and Hash for a
+    /// type that implements all three of `Hash`, `Borrow<str>` and `Borrow<[u8]>``
```
y21, line 13: I see you changed it to use ticks on this line specifically. It should probably be done for *all* trait references and types.
Also a small typo:
```suggestion
    /// type that implements all three of `Hash`, `Borrow<str>` and `Borrow<[u8]>`
```

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// typically uses Borrow<T> when it relies on the identical behavior of these additional trait
+    /// implementations. These traits will likely appear as additional trait bounds.
+    ///
+    /// In particular Eq, Ord and Hash must be equivalent for borrowed and owned values:
+    /// `x.borrow() == y.borrow()` should give the same result as `x == y`.
+    /// It follows then that the following equivalence must hold:
+    /// `hash(x) == hash((x as Borrow<[u8]>).borrow()) == hash((x as Borrow<str>).borrow())`
+    ///
+    /// Unfortunately it doesn't hold as `hash("abc") != hash("abc".as_bytes())`.
+    /// this is because the that Hash impl for str passes an additional 0xFF byte to
```
y21, line 30:
```suggestion
    /// This is because the that Hash impl for str passes an additional 0xFF byte to
```

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// the hasher to avoid collisions. For example, given the tuples ("a", "bc"), and ("ab", "c"),
+    /// the two tuples would have the same hash value if the 0xFF byte was not added.
+    /// https://doc.rust-lang.org/stable/core/hash/trait.Hash.html#prefix-collisions
```
y21, line 33: The explanation looks good, but now I think the link is not needed anymore, since this covers everything :D

y21 (approved):
Just some minor documentation nits. Thanks for your patience here.
I'm going to tag @xFrednet since these are probably my last comments.
Aside from the small things in the documentation, this looks great!
It's a very nice lint to have.

partiallytyped:
Don't worry about it :D it shows how much you care for this which makes me more excited to help :D

xFrednet:
The start of my week is quite full, it might take until Thursday or the weekend until I have time to give you a full review. Hope that's okay with you :)

clippy_lints/src/hash_borrow_str_semantics.rs
y21: @partiallytyped I just noticed that the typo is still there ^^ (two backticks)

partiallytyped: Oops 🤣

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,105 @@
...
+                "can't satisfy the semantics of `Hash` when both `Borrow<str>` and `Borrow<[u8]>` are implemented",
+                |diag| {
+                    diag.note(
+                        "`Borrow` requires `Hash` values for all borrowed types to be the same. This is not the case for `Borrow<str>` and `Borrow<[u8]>`"
+                    );
+                    diag.note("as the two types result in different `Hash` values, e.g. `hash(\"abc\") != hash(\"abc\".as_bytes())`");
+
+                    diag.help("consider removing one implementation of `Borrow` (`Borrow<str>` or `Borrow<[u8]>`) or not implementing `Hash` for this type");
+                },
+            );
```
xFrednet, line 102: Creating a good message for this lint is hard, due to the complex nature of the issue. I have the feeling, that the following code snippet will make the issue a bit clearer and is more inline to the [message guidelines](https://rustc-dev-guide.rust-lang.org/diagnostics.html#error-messages), but I can't really define what makes it better :sweat_smile: 

```rs
span_lint_and_then(
    cx,
    HASH_BORROW_STR_SEMANTICS,
    *span,
    "implementation of `Hash` that doesn't satisfy the semantics of the trait due to other impls",
    |diag| {
        diag.note("the type of this implementation implements `Borrow<str>` and `Borrow<[u8]>`");
        diag.note("the `Borrow` semantics define that the `Hash` values for all borrowed types should be the same");
        diag.note(
            "however, a string and the bytes of a string have different hash values (`hash(\"abc\") != hash(\"abc\".as_bytes())`)."
        );

        diag.help("consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...");
        diag.help("... or don't implement `Hash` for this type");
    },
);
```

The resulting message would roughly look like this:

```
error: implementation of `Hash` that doesn't satisfy the semantics of the trait due to other impls
  --> $DIR/hash_borrow_str_semantics.rs:10:6
   |
LL | impl Hash for ExampleType {
   |      ^^^^
   |
   = note: the type of this implementation implements `Borrow<str>` and `Borrow<[u8]>`
   = note: the `Borrow` trait define that the `Hash` values for all borrowed types should be the same
   = note: however, a string and the bytes of a string have different hash values (`hash(\"abc\") != hash(\"abc\".as_bytes())`)
   = help: consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...
   = help: ... or don't implement `Hash` for this type
```

tests/ui/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,89 @@
...
+    fn borrow(&self) -> &str {
+        &self.data
+    }
+}
+
+impl Borrow<[u8]> for ExampleType {
+    fn borrow(&self) -> &[u8] {
+        self.data.as_bytes()
+    }
+}
```
xFrednet, line 27: Could you also check an example with generics, like this:

```rs
use std::borrow::Borrow;
use std::hash::Hasher;
use std::hash::Hash;

struct GenericExampleType<T> {
    data: T,
}

impl<T: Hash> Hash for GenericExampleType<T> {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.data.hash(state);
    }
}

impl Borrow<str> for GenericExampleType<String> {
    fn borrow(&self) -> &str {
        &self.data
    }
}

impl Borrow<[u8]> for GenericExampleType<&'static [u8]> {
    fn borrow(&self) -> &[u8] {
        self.data
    }
}
```

I would expect this not to emit a lint message. If it does, it might be difficult to counteract this, since I don't know if we can ask rustc if two generic declarations/specifications are mutually exclusive. We could maybe just accept this as a false positive, if we change the lint group to be warn-by-default instead.

xFrednet (commented):
The lint logic looks good to me. I have some remarks regarding the cosmetics of the lint. Afterward, it should be good to go.

Thank you for waiting on my reivew :)

clippy_lints/src/hash_borrow_str_semantics.rs
```diff
@@ -0,0 +1,105 @@
...
+    ///     fn borrow(&self) -> &[u8] {
+    ///         self.data.as_bytes()
+    ///     }
+    /// }
+    /// ```
+    /// As a consequence, hashing a `&ExampleType` and hashing the result of the two
+    /// borrows will result in different values.
+    ///
+    #[clippy::version = "1.75.0"]
+    pub HASH_BORROW_STR_SEMANTICS,
```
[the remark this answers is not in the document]
partiallytyped: I like the second one, I think it's clear because it indicates that the issue is impl'ing Hash.

tests/ui/hash_borrow_str_semantics.rs
partiallytyped: I don't think it throws an error as it must be considered a different type in each case 🤔.

clippy_lints/src/hash_borrow_str_semantics.rs
partiallytyped: I think it's the wording of the semantic s, in particular the whole "`Hash` values" sounds a bit wrong.

xFrednet: You're right, is "hashes" or something better?

partiallytyped: I am not sure, current lint is

```rust
 "the semantics of `Borrow<T>` around `Hash` can't be satisfied when both `Borrow<str>` and `Borrow<[u8]>` are implemented",
  |diag| {
      diag.note("the `Borrow` semantics define that the `Hash` values for all borrowed types should be the same");
      diag.note(
          "however, a string and the bytes of a string have different hash values (`hash(\"abc\") != hash(\"abc\".as_bytes())`)"
      );
      diag.help("consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...");
      diag.help("... or don't implement `Hash` for this type");
  },
```

perhaps this is more appropriate?

```rust
 "the semantics of `Borrow<T>` around `Hash` can't be satisfied when both `Borrow<str>` and `Borrow<[u8]>` are implemented",
  |diag| {
      diag.note("the `Borrow` semantics define that `Hash` must behave the same for all implementations of Borrow<T>");
      diag.note(
          "however, the hash implementations of a string (`str`) and the bytes of a string `[u8]` do not behave the same"
      );
      diag.note("as (`hash(\"abc\") != hash(\"abc\".as_bytes())`");
      diag.help("consider either removing one of the  `Borrow` implementations (`Borrow<str>` or `Borrow<[u8]>`) ...");
      diag.help("... or don't implement `Hash` for this type");
  },
```

xFrednet: Yeah, I think having it like the second example would be good. Then everything should be ready :)

partiallytyped: Great! :D I adjusted the last sentence a bit as it was grammatically incorrect 😓 given the changes.

xFrednet (approved):

xFrednet:
LGTM, thank you for the update!

@bors r+

PR #11782, opened 2023-11-09
Make `macro_use_imports` lint ordering more stable

Alexendoo:
changelog: none

Fixes [the `macro_use_imports` ordering dependence](https://github.com/rust-lang/rust/pull/117649#issuecomment-1797716088) on the hash of `Span`s

### DIFF
--- a/clippy_lints/src/macro_use.rs
+++ b/clippy_lints/src/macro_use.rs
@@ -3,13 +3,14 @@ use clippy_utils::source::snippet;
 use hir::def::{DefKind, Res};
 use if_chain::if_chain;
 use rustc_ast::ast;
-use rustc_data_structures::fx::{FxHashMap, FxHashSet};
+use rustc_data_structures::fx::FxHashSet;
 use rustc_errors::Applicability;
 use rustc_hir as hir;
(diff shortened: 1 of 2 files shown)


clippy_lints/src/macro_use.rs
```diff
@@ -142,7 +143,7 @@ impl<'tcx> LateLintPass<'tcx> for MacroUseImports {
         }
     }
     fn check_crate_post(&mut self, cx: &LateContext<'_>) {
-        let mut used = FxHashMap::default();
+        let mut used = BTreeMap::new();
```
Centri3, line 146: Is `BTreeMap` faster for this than `FxIndexMap`? If so there are some cases where this should be used instead, but I'm unsure if this is the case..

Centri3 (approved):
One question

clippy_lints/src/macro_use.rs
Alexendoo: Could go either way and would vary based on the size of the map/complexity of the keys. An index map would work as well but I didn't pick one over the other for any particular reason

Centri3:
Thanks, @bors r+

PR #11787, opened 2023-11-09
Fixes to `manual_let_else`'s divergence check

Jarcho:
A few changes to the divergence check in `manual_let_else` and moves it the implementation to `clippy_utils` since it's generally useful:
* Handle internal `break` and `continue` expressions.
    e.g. The first loop is divergent, but the second is not.
    ```rust
    {
        loop {
            break 'outer;
        };
    }
    {
        loop {
            break;
        };
    }
    ```
* Match rust's definition of divergence which is defined via the type system.
    e.g. The following is not considered divergent by rustc as the inner block has a result type of `()`:
    ```rust
    {
        'a: {
            panic!();
            break 'a;
        };
    }
    ```
* Handle when adding a single semicolon would make the expression divergent.
    e.g. The following would be a divergent if a semicolon were added after the `if` expression:
    ```rust
    { if panic!() { 0 } else { 1 } }
    ```

changelog: None

### DIFF
--- a/clippy_lints/src/manual_let_else.rs
+++ b/clippy_lints/src/manual_let_else.rs
@@ -5,18 +5,15 @@ use clippy_utils::diagnostics::span_lint_and_then;
 use clippy_utils::higher::IfLetOrMatch;
 use clippy_utils::source::snippet_with_context;
 use clippy_utils::ty::is_type_diagnostic_item;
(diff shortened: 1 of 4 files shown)


dswij (approved):
Looks great, thanks for this!

Issue #11788, opened 2023-11-10
Lint for iterating over the keys in an unordered collection such as HashSet or HashMap


kayabaNerve:
### What it does

A (possibly pedantic) lint for iterating over a HashSet/HashMap would be appreciated. Given these do not stable iteration orders across program instances, any usage of them in redundant systems may cause differing results and inconsistencies. They also have an unclear iteration order (not lexicographical to keys nor by order) which may give adverse effects in a variety of other use cases.

In effect, the same as https://github.com/rust-lang/rust/pull/92584 (which I would appreciate if anyone could give me insight on how to fork/run over my own project now, possibly via dylint).

### Example

```rust
for (key, value) in my_map {}
```

Could be written as:

```rust
let mut keys = my_map.keys().clone().collect::<Vec<_>();
keys.sort();
for key in keys {
  let value = &my_map[value];
}
```

Please note I don't actually endorse the suggested alternative. My usages of map iteration frequently have context which show how iteration *should* occur. Without said context, my example is forced to create a slow but safe migration for this lint topic.

Jarcho:
This is definitely a restriction lint, but it does have it's uses.

kayabaNerve:
Thank you @Jacherr!

PR #11791, opened 2023-11-11
Implement new lint `iter_over_hash_type`

Jacherr:
Implements and fixes https://github.com/rust-lang/rust-clippy/issues/11788

This PR adds a new *restriction* lint `iter_over_hash_type` which prevents `Hash`-types (that is, `HashSet` and `HashMap`) from being used as the iterator in `for` loops.

The justification for this is because in `Hash`-based types, the ordering of items is not guaranteed and may vary between executions of the same program on the same hardware. In addition, it reduces readability due to the unclear iteration order.

The implementation of this lint also ensures the following:
- Calls to `HashMap::keys`, `HashMap::values`, and `HashSet::iter` are also denied when used in `for` loops,
- When this expression is used in procedural macros, it is not linted/denied.

changelog: add new `iter_over_hash_type` lint to prevent unordered iterations through hashed data structures


clippy_lints/src/iter_over_hash_type.rs
```diff
@@ -0,0 +1,63 @@
...
+        if let Some(for_loop) = ForLoop::hir(expr)
+            && !for_loop.body.span.from_expansion()
+            && let ty = cx.typeck_results().expr_ty(for_loop.arg).peel_refs()
+            && let Some(adt) = ty.ty_adt_def()
+            && let did = adt.did()
+            && (match_def_path(cx, did, &HASHMAP_KEYS)
+                || match_def_path(cx, did, &HASHMAP_VALUES)
+                || match_def_path(cx, did, &HASHSET_ITER_TY)
+                || is_type_diagnostic_item(cx, ty, sym::HashMap)
+                || is_type_diagnostic_item(cx, ty, sym::HashSet))
```
Jarcho, line 53: This is missing `HashSet::IterMut`, `HashMap::Iter`, `HashMap::IterMut` and `HashMap::ValuesMut`. `Drain` for both collections should probably also be caught.

clippy_lints/src/iter_over_hash_type.rs
```diff
@@ -0,0 +1,63 @@
...
+use rustc_session::{declare_lint_pass, declare_tool_lint};
+use rustc_span::sym;
+
+declare_clippy_lint! {
+    /// ### What it does
+    /// This is a restriction lint which prevents the use of hash types (i.e., `HashSet` and `HashMap`) in for loops.
+    ///
+    /// ### Why is this bad?
+    /// Because hash types are unordered, when iterated through such as in a for loop, the values are returned in
+    /// a pseudo-random order. As a result, on redundant systems this may cause inconsistencies and anomalies.
```
Jarcho, line 16: nit: an undefined order

'pseudo-random' is not technically wrong, but very misleading.

clippy_lints/src/iter_over_hash_type.rs
```diff
@@ -0,0 +1,63 @@
...
+                || match_def_path(cx, did, &HASHMAP_VALUES)
+                || match_def_path(cx, did, &HASHSET_ITER_TY)
+                || is_type_diagnostic_item(cx, ty, sym::HashMap)
+                || is_type_diagnostic_item(cx, ty, sym::HashSet))
+        {
+            span_lint(
+                cx,
+                ITER_OVER_HASH_TYPE,
+                expr.span,
+                "iterating over unordered hash-based type",
```
Jarcho, line 59: nit: iteration over

Jarcho (changes requested):
Overall fine other than some small changes. It would be nice if this could handle adapters being used (e.g. `map`), but that can be done as a future extension.

@rustbot author

clippy_lints/src/iter_over_hash_type.rs
Jacherr: `HashSet::IterMut` doesn't appear to exist, but I'll add the rest.

Jarcho: So it does not.

tests/ui/iter_over_hash_type.rs
Centri3, line 1: I'd like tests with type aliases as well, like `FxHashMap`. This should work since you use the typeck table but it should be linted

Jarcho:
Thank you. @bors r+

PR #11792, opened 2023-11-11
[`map_identity`]: respect match ergonomics

### LINKED ISSUE #11764: `map_identity` does not consider reference matching ergonomics
### Summary

`map_identity` is falsely reported for the cases where reference matching ergonomics ([RFC 2005](https://rust-lang.github.io/rfcs/2005-match-ergonomics.html)) changes where a reference is introduced in tuples or arrays (e.g., `&(T,)` -> `(&T,)`).

### Lint Name

map_identity

### Reproducer

I tried this code:

```rust
pub fn foo<'a>(it: impl Iterator<Item = &'a (i32,)>) -> impl Iterator<Item = (&'a i32,)> {
    it.map(|(x,)| (x,))
    // RFC 2005 transforms the pattern to `&(ref x,)`
}
```

I saw this happen:

```
warning: unnecessary map of the identity function
 --> src/lib.rs:2:7
  |
2 |     it.map(|(x,)| (x,))
  |       ^^^^^^^^^^^^^^^^^ help: remove the call to `map`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#map_identity
  = note: `#[warn(clippy::map_identity)]` on by default
```

I expected to see this happen: No lints reported

### Version

```text
rustc 1.75.0-nightly (fee5518cd 2023-11-05)
binary: rustc
commit-hash: fee5518cdd4435c60a57fe3bb734fc1a14abeb7a
commit-date: 2023-11-05
host: x86_64-pc-windows-gnu
release: 1.75.0-nightly
LLVM version: 17.0.4
```

### Additional Labels

y21:
Fixes #11764

Note: the original tests before this were slightly wrong themselves already and had to be changed. They were calling `map` on an iterator of `&(i32, i32)`s, so this PR would stop linting there, but they were meant to test something else unrelated to binding modes, so I just changed them to remove the references so that it iterates over owned values and they all bind by value. This way they continue to test what they checked for before: the identity function for tuple patterns.

changelog: [`map_identity`]: respect match ergonomics

### DIFF
--- a/clippy_utils/src/lib.rs
+++ b/clippy_utils/src/lib.rs
@@ -2034,6 +2034,18 @@ pub fn is_must_use_func_call(cx: &LateContext<'_>, expr: &Expr<'_>) -> bool {
 /// Consider calling [`is_expr_untyped_identity_function`] or [`is_expr_identity_function`] instead.
 fn is_body_identity_function(cx: &LateContext<'_>, func: &Body<'_>) -> bool {
     fn check_pat(cx: &LateContext<'_>, pat: &Pat<'_>, expr: &Expr<'_>) -> bool {
+        if cx
+            .typeck_results()
+            .pat_binding_modes()
+            .get(pat.hir_id)
+            .is_some_and(|mode| matches!(mode, BindingMode::BindByReference(_)))
+        {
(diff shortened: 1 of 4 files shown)


llogiq:
Thank you, that's a solid improvement!

@bors r+

Issue #10118, opened 2022-12-25
`==` check instead of `opt.map_or(false, |val| val == n)` or `if let Some(n) = opt`


laralove143:
### What it does

It converts a code like
```rust
opt.map_or(false, |val| val == 5)
```
to
```rust
opt == Some(5);
```
This could apply to `Result` and technically any enum wrapping a value, for example
```rust
matches!(err, Error::Io(5));
```
to
```rust
err == Error::Io(5);
```

### Lint Name

Unnecessary pattern matching

### Category

style

### Advantage

- Readability
- Conciseness
- Possible optimization (Could someone that can read assembly verify this? The resulting assembly has fewer lines)

### Drawbacks

The enum needs to implement `PartialEq`

### Example

```rust
#[derive(PartialEq)]
enum Error {
    Io(u16),
    Os(u16),
}

fn unnecessary_pattern_match() {
    let err = Error::Os(1);
   matches!(err, Error::Io(5));
}

fn unnecessary_map_or() {
    let opt = Some(1);
    opt.map_or(false, |val| val == 5);
}
```

Could be written as:

```rust
#[derive(PartialEq)]
enum Error {
    Io(u16),
    Os(u16),
}

fn unnecessary_pattern_match() {
    let err = Error::Os(1);
    err == Error::Io(5);
}

fn unnecessary_map_or() {
    let opt = Some(1);
    opt == Some(5);
}
```

llogiq:
Note that depending on the `PartialEq` implementation, this lint might change behavior! There is no guarantee that those will be equal. Consider:

    enum AllTheSame {
        One(u32),
        Other(bool),
    }

    impl core::ops::PartialEq for AllTheSame {
        fn eq(&self, other: &Self) -> bool { true }
    }

This is a valid type with a valid `PartialEq`  impl, yet changing `matches!(ats, AllTheSame::Other(true))` to `ats == AllTheSame::Other(true)` would effectively remove the check.

What we might try is check if the `PartialEq` impl was derived and only lint for those types.

laralove143:
Good point but yes we can do that, or we can check if it's equal to the pattern match (not sure if you can run code while linting)

Another consideration though is if they implemented it custom, maybe they don't want you to use pattern matching, for example:

```rust
enum Num {
    Small(u8),
    Large(u64),
}

assert_eq!(Num::Small(5), Num::Large(5));
```

llogiq:
[...]

PR #11796, opened 2023-11-12
New lint: `unnecessary_map_or`

Jacherr:
Closes https://github.com/rust-lang/rust-clippy/issues/10118

This lint checks `map_or` method calls to check if they can be consolidated down to something simpler and/or more readable.

For example, the code 
```rs
let x = Some(5);
x.map_or(false, |n| n == 5)
```

can be rewritten as 
```rs
let x = Some(5);
x == Some(5)
```

In addition, when the closure is more complex, the code can be altered from, say,
```rs
let x = Ok::<Vec<i32>, i32>(vec![5]);
x.map_or(false, |n| n == [5])
```
into
```rs
let x = Ok::<Vec<i32>, i32>(vec![5]);
x.is_some_and(|n| n == [5])
```

This lint also considers cases where the `map_or` can be chained with other method calls, and accommodates accordingly by adding extra parentheses as needed to the suggestion.

changelog: add new lint `unnecessary_map_or`

### DIFF
--- a/clippy_lints/src/methods/unnecessary_map_or.rs
+++ b/clippy_lints/src/methods/unnecessary_map_or.rs
@@ -1,0 +1,131 @@
+use std::borrow::Cow;
+
+use clippy_config::msrvs::{self, Msrv};
+use clippy_utils::diagnostics::span_lint_and_sugg;
+use clippy_utils::eager_or_lazy::switch_to_eager_eval;
+use clippy_utils::source::snippet_opt;
+use clippy_utils::sugg::{Sugg, make_binop};
+use clippy_utils::ty::{get_type_diagnostic_name, implements_trait};
+use clippy_utils::visitors::is_local_used;
+use clippy_utils::{is_from_proc_macro, path_to_local_id};
+use rustc_ast::LitKind::Bool;
+use rustc_errors::Applicability;
+use rustc_hir::{BinOpKind, Expr, ExprKind, PatKind};
+use rustc_lint::LateContext;
+use rustc_span::sym;
+
+use super::UNNECESSARY_MAP_OR;
+
+pub(super) enum Variant {
+    Ok,
+    Some,
+}
+impl Variant {
+    pub fn variant_name(&self) -> &'static str {
+        match self {
+            Variant::Ok => "Ok",
+            Variant::Some => "Some",
+        }
+    }
+
+    pub fn method_name(&self) -> &'static str {
+        match self {
+            Variant::Ok => "is_ok_and",
+            Variant::Some => "is_some_and",
+        }
+    }
+}
+
+pub(super) fn check<'a>(
+    cx: &LateContext<'a>,
+    expr: &Expr<'a>,
+    recv: &Expr<'_>,
+    def: &Expr<'_>,
+    map: &Expr<'_>,
+    msrv: &Msrv,
+) {
+    let ExprKind::Lit(def_kind) = def.kind else {
+        return;
+    };
+
+    let recv_ty = cx.typeck_results().expr_ty(recv);
+
+    let Bool(def_bool) = def_kind.node else {
+        return;
+    };
+
+    let variant = match get_type_diagnostic_name(cx, recv_ty) {
+        Some(sym::Option) => Variant::Some,
+        Some(sym::Result) => Variant::Ok,
+        Some(_) | None => return,
+    };
+
+    let (sugg, method) = if let ExprKind::Closure(map_closure) = map.kind
+            && let closure_body = cx.tcx.hir().body(map_closure.body)
+            && let closure_body_value = closure_body.value.peel_blocks()
+            && let ExprKind::Binary(op, l, r) = closure_body_value.kind
+            && let Some(param) = closure_body.params.first()
+            && let PatKind::Binding(_, hir_id, _, _) = param.pat.kind
+            // checking that map_or is one of the following:
+            // .map_or(false, |x| x == y)
+            // .map_or(false, |x| y == x) - swapped comparison
+            // .map_or(true, |x| x != y)
+            // .map_or(true, |x| y != x) - swapped comparison
+            && ((BinOpKind::Eq == op.node && !def_bool) || (BinOpKind::Ne == op.node && def_bool))
+            && let non_binding_location = if path_to_local_id(l, hir_id) { r } else { l }
+            && switch_to_eager_eval(cx, non_binding_location)
+            // xor, because if its both then thats a strange edge case and
+            // we can just ignore it, since by default clippy will error on this
+            && (path_to_local_id(l, hir_id) ^ path_to_local_id(r, hir_id))
+            && !is_local_used(cx, non_binding_location, hir_id)
+            && let typeck_results = cx.typeck_results()
+            && typeck_results.expr_ty(l) == typeck_results.expr_ty(r)
+            && let Some(partial_eq) = cx.tcx.get_diagnostic_item(sym::PartialEq)
+            && implements_trait(cx, recv_ty, partial_eq, &[recv_ty.into()])
+    {
+        let wrap = variant.variant_name();
+
+        // we may need to add parens around the suggestion
+        // in case the parent expression has additional method calls,
+        // since for example `Some(5).map_or(false, |x| x == 5).then(|| 1)`
+        // being converted to `Some(5) == Some(5).then(|| 1)` isnt
+        // the same thing
+
+        let inner_non_binding = Sugg::NonParen(Cow::Owned(format!(
+            "{wrap}({})",
+            Sugg::hir(cx, non_binding_location, "")
+        )));
+
+        let binop = make_binop(op.node, &Sugg::hir(cx, recv, ".."), &inner_non_binding)
+            .maybe_par()
+            .into_string();
+
+        (binop, "a standard comparison")
+    } else if !def_bool
+        && msrv.meets(msrvs::OPTION_RESULT_IS_VARIANT_AND)
+        && let Some(recv_callsite) = snippet_opt(cx, recv.span.source_callsite())
+        && let Some(span_callsite) = snippet_opt(cx, map.span.source_callsite())
+    {
+        let suggested_name = variant.method_name();
+        (
+            format!("{recv_callsite}.{suggested_name}({span_callsite})",),
+            suggested_name,
+        )
+    } else {
+        return;
+    };
+
+    if is_from_proc_macro(cx, expr) {
+        return;
+    }
+
+    span_lint_and_sugg(
+        cx,
+        UNNECESSARY_MAP_OR,
+        expr.span,
+        "this `map_or` is redundant",
+        format!("use {method} instead"),
+        sugg,
+        Applicability::MaybeIncorrect,
+    );
+}
--- a/tests/ui/unnecessary_map_or.fixed
+++ b/tests/ui/unnecessary_map_or.fixed
@@ -1,0 +1,64 @@
+//@aux-build:proc_macros.rs
+#![warn(clippy::unnecessary_map_or)]
+#![allow(clippy::no_effect)]
+#![allow(clippy::eq_op)]
+#![allow(clippy::unnecessary_lazy_evaluations)]
+#[clippy::msrv = "1.70.0"]
+#[macro_use]
+extern crate proc_macros;
+
+fn main() {
+    // should trigger
+    let _ = (Some(5) == Some(5));
+    let _ = (Some(5) != Some(5));
+    let _ = (Some(5) == Some(5));
+    let _ = Some(5).is_some_and(|n| {
+        let _ = n;
+        6 >= 5
+    });
+    let _ = Some(vec![5]).is_some_and(|n| n == [5]);
+    let _ = Some(vec![1]).is_some_and(|n| vec![2] == n);
+    let _ = Some(5).is_some_and(|n| n == n);
+    let _ = Some(5).is_some_and(|n| n == if 2 > 1 { n } else { 0 });
+    let _ = Ok::<Vec<i32>, i32>(vec![5]).is_ok_and(|n| n == [5]);
+    let _ = (Ok::<i32, i32>(5) == Ok(5));
+    let _ = (Some(5) == Some(5)).then(|| 1);
+
+    // shouldnt trigger
+    let _ = Some(5).map_or(true, |n| n == 5);
+    let _ = Some(5).map_or(true, |n| 5 == n);
+    macro_rules! x {
+        () => {
+            Some(1)
+        };
+    }
+    // methods lints dont fire on macros
+    let _ = x!().map_or(false, |n| n == 1);
+    let _ = x!().map_or(false, |n| n == vec![1][0]);
+
+    msrv_1_69();
+
+    external! {
+        let _ = Some(5).map_or(false, |n| n == 5);
+    }
+
+    with_span! {
+        let _ = Some(5).map_or(false, |n| n == 5);
+    }
+
+    // check for presence of PartialEq, and alter suggestion to use `is_ok_and` if absent
+    struct S;
+    let r: Result<i32, S> = Ok(3);
+    let _ = r.is_ok_and(|x| x == 7);
+
+    #[derive(PartialEq)]
+    struct S2;
+    let r: Result<i32, S2> = Ok(4);
+    let _ = (r == Ok(8));
+}
+
+#[clippy::msrv = "1.69.0"]
+fn msrv_1_69() {
+    // is_some_and added in 1.70.0
+    let _ = Some(5).map_or(false, |n| n == if 2 > 1 { n } else { 0 });
+}
--- a/tests/ui/unnecessary_map_or.rs
+++ b/tests/ui/unnecessary_map_or.rs
@@ -1,0 +1,67 @@
+//@aux-build:proc_macros.rs
+#![warn(clippy::unnecessary_map_or)]
+#![allow(clippy::no_effect)]
+#![allow(clippy::eq_op)]
+#![allow(clippy::unnecessary_lazy_evaluations)]
+#[clippy::msrv = "1.70.0"]
+#[macro_use]
+extern crate proc_macros;
+
+fn main() {
+    // should trigger
+    let _ = Some(5).map_or(false, |n| n == 5);
+    let _ = Some(5).map_or(true, |n| n != 5);
+    let _ = Some(5).map_or(false, |n| {
+        let _ = 1;
+        n == 5
+    });
+    let _ = Some(5).map_or(false, |n| {
+        let _ = n;
+        6 >= 5
+    });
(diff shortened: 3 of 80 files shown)


clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+            let sugg = if is_type_diagnostic_item(cx, recv_ty, sym::Option) {
+                "is_some_and"
+            } else {
+                "is_ok_and"
+            };
+
+            (
+                format!(
+                    "{}.{}({})",
+                    snippet(cx, recv.span.source_callsite(), ".."),
```
Centri3, line 85: This should be either `snippet_with_applicability` or `snippet_opt`

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+                    snippet(cx, recv.span, ".."),
+                    comparator,
+                    wrap,
+                    snippet(cx, non_binding_location.span.source_callsite(), ".."),
+                    if should_add_parens { ")" } else { "" }
+                ),
+                "standard comparison",
+            )
+        } else if !def_bool && msrv.meets(msrvs::OPTION_IS_SOME_AND) {
+            let sugg = if is_type_diagnostic_item(cx, recv_ty, sym::Option) {
```
Centri3, line 76: I don't really like checking this again. imo you should create an enum based on the results of the first check - i.e., `Result`/`Option`, then give it a method like `sugg`. As an example take a look at `unwrap_used`: https://github.com/Centri3/rust-clippy/blob/71c54137eadc730d01fe15c17df8bf3ba5d3b04b/clippy_lints/src/methods/unwrap_expect_used.rs#L12-L15

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+        && ((BinOpKind::Eq == op.node && !def_bool) || (BinOpKind::Ne == op.node && def_bool))
+        && let non_binding_location = if path_to_local_id(l, hir_id) { r } else { l }
+        && switch_to_eager_eval(cx, non_binding_location)
+        // xor, because if its both then thats a strange edge case and
+        // we can just ignore it, since by default clippy will error on this
+        && (path_to_local_id(l, hir_id) ^ path_to_local_id(r, hir_id))
+        && !is_local_used(cx, non_binding_location, hir_id)
+        && typeck_results.expr_ty(l) == typeck_results.expr_ty(r)
+        {
+            let wrap = if is_type_diagnostic_item(cx, recv_ty, sym::Option) {
```
Centri3, line 49: Same here, maybe `method_name`?

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+            };
+            let comparator = if BinOpKind::Eq == op.node { "==" } else { "!=" };
+
+            // we may need to add parens around the suggestion
+            // in case the parent expression has additional method calls,
+            // since for example `Some(5).map_or(false, |x| x == 5).then(|| 1)`
+            // being converted to `Some(5) == Some(5).then(|| 1)` isnt
+            // the same thing
+            let should_add_parens = get_parent_expr(cx, expr)
+                .is_some_and(|expr| expr.precedence().order() > i8::try_from(AssocOp::Equal.precedence()).unwrap_or(0));
```
Centri3, line 62: Could we use `maybe_par` with `Sugg` instead?

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+            (
+                format!(
```
Centri3, line 64: nit: Please inline some of these, like `"{comparator}"`.

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+        let (sugg, method) = if let ExprKind::Closure(map_closure) = map.kind
+        && let closure_body = cx.tcx.hir().body(map_closure.body)
+        && let closure_body_value = closure_body.value.peel_blocks()
+        && let ExprKind::Binary(op, l, r) = closure_body_value.kind
+        && let Some(param) = closure_body.params.first()
+        && let PatKind::Binding(_, hir_id, _, _) = param.pat.kind
+        // checking that map_or is either:
+        // .map_or(false, |x| x == y) OR .map_or(true, |x| x != y)
+        && ((BinOpKind::Eq == op.node && !def_bool) || (BinOpKind::Ne == op.node && def_bool))
+        && let non_binding_location = if path_to_local_id(l, hir_id) { r } else { l }
```
Centri3, line 41: Can you add a comment here explaining that this allows linting both `y == x` and `x == y`? I also think "binding" here is misleading, this should be `param` instead

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,105 @@
...
+        && (path_to_local_id(l, hir_id) ^ path_to_local_id(r, hir_id))
+        && !is_local_used(cx, non_binding_location, hir_id)
+        && typeck_results.expr_ty(l) == typeck_results.expr_ty(r)
+        {
+            let wrap = if is_type_diagnostic_item(cx, recv_ty, sym::Option) {
+                "Some"
+            } else {
+                "Ok"
+            };
+            let comparator = if BinOpKind::Eq == op.node { "==" } else { "!=" };
```
Centri3, line 54: We can use `as_str` here

WaffleLapkin:
@Jacherr this seems to be waiting on you, since Catherine left you a review.
@rustbot author

Jacherr:
Hi, apologies for the lack of movement on this. I’ll make changes based on the review now, and once everything is up to standard I’ll go through the conflicts.

clippy_lints/src/methods/unnecessary_map_or.rs
Jacherr: Could you clarify exactly what you mean here? Would something like `make_binop(BinOpKind::Eq, Sugg::hir(recv), Sugg::hir(non_binding_location)).maybe_par()` be what you're looking for?

xFrednet:
Hey @Jacherr Sorry for the delay. @Centri3 was sadly busy, but now she indicated that she'll have time for reviews again. Would you mind rebasing again? Next time there is a long delay, you can also pick a new reviewer with `r? clippy`.

Jacherr:
I did rebase but it's not showing up in the PR for some reason 😕

tests/ui/unnecessary_map_or.rs
Centri3, line 1: Just needs tests for external and procedural macros

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,119 @@
...
+            && let Some(recv_callsite) = snippet_opt(cx, recv.span.source_callsite())
+            && let Some(span_callsite) = snippet_opt(cx, map.span.source_callsite())
+        {
+            let sugg = variant.method_name();
+            (format!("{recv_callsite}.{sugg}({span_callsite})",), sugg)
+        } else {
+            return;
+        };
+
+        span_lint_and_sugg(
```
Centri3, line 109: We should call `is_from_proc_macro` here

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,119 @@
...
+                ),
+                "standard comparison",
+            )
+        } else if !def_bool
+            && msrv.meets(msrvs::OPTION_RESULT_IS_VARIANT_AND)
+            && let Some(recv_callsite) = snippet_opt(cx, recv.span.source_callsite())
+            && let Some(span_callsite) = snippet_opt(cx, map.span.source_callsite())
+        {
+            let sugg = variant.method_name();
+            (format!("{recv_callsite}.{sugg}({span_callsite})",), sugg)
```
Centri3, line 104:
```suggestion
            let suggested_method = variant.method_name();
            (format!("{recv_callsite}.{sugg}({span_callsite})",), suggested_method)
```
Since this has the same name as the `sugg` above, which is the other value in the tuple

clippy_lints/src/methods/unnecessary_map_or.rs
Centri3: Basically, checking precedence here I think is so the code doesn't change its meaning. Using `maybe_par` will ensure (maybe while requiring an extra run of `check`) that the precedence isn't changed.

Yeah, that snippet actually looks perfect

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,119 @@
...
+        // .map_or(true, |x| x != y)
+        // .map_or(true, |x| y != x) - swapped comparison
+        && ((BinOpKind::Eq == op.node && !def_bool) || (BinOpKind::Ne == op.node && def_bool))
+        && let non_binding_location = if path_to_local_id(l, hir_id) { r } else { l }
+        && switch_to_eager_eval(cx, non_binding_location)
+        // xor, because if its both then thats a strange edge case and
+        // we can just ignore it, since by default clippy will error on this
+        && (path_to_local_id(l, hir_id) ^ path_to_local_id(r, hir_id))
+        && !is_local_used(cx, non_binding_location, hir_id)
+        && typeck_results.expr_ty(l) == typeck_results.expr_ty(r)
```
Centri3, line 76: Small formatting change, surprised rustfmt didn't catch this
```suggestion
            && let closure_body = cx.tcx.hir().body(map_closure.body)
            && let closure_body_value = closure_body.value.peel_blocks()
            && let ExprKind::Binary(op, l, r) = closure_body_value.kind
            && let Some(param) = closure_body.params.first()
            && let PatKind::Binding(_, hir_id, _, _) = param.pat.kind
            // checking that map_or is one of the following:
            // .map_or(false, |x| x == y)
            // .map_or(false, |x| y == x) - swapped comparison
            // .map_or(true, |x| x != y)
            // .map_or(true, |x| y != x) - swapped comparison
            && ((BinOpKind::Eq == op.node && !def_bool) || (BinOpKind::Ne == op.node && def_bool))
            && let non_binding_location = if path_to_local_id(l, hir_id) { r } else { l }
            && switch_to_eager_eval(cx, non_binding_location)
            // xor, because if its both then thats a strange edge case and
            // we can just ignore it, since by default clippy will error on this
            && (path_to_local_id(l, hir_id) ^ path_to_local_id(r, hir_id))
            && !is_local_used(cx, non_binding_location, hir_id)
            && typeck_results.expr_ty(l) == typeck_results.expr_ty(r)
```

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,119 @@
...
+// Only checking map_or for now
+pub(super) fn check(
+    cx: &LateContext<'_>,
+    expr: &Expr<'_>,
+    recv: &Expr<'_>,
+    def: &Expr<'_>,
+    map: &Expr<'_>,
+    msrv: &Msrv,
+) {
+    if let ExprKind::Lit(def_kind) = def.kind
```
Centri3, line 46: Would prefer using a bunch of `let..else`s over this

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            return;
+        };
+
+        span_lint_and_sugg(
+            cx,
+            UNNECESSARY_MAP_OR,
+            expr.span,
+            format!("`map_or` is redundant, use {method} instead"),
+            "try",
```
Centri3, line 114:
```suggestion
            format!("this `map_or` is redundant"),
            "use {method} instead",
```

Centri3 (approved):
Just a couple more changes then this is good <3

xFrednet:
Hey, this is a ping from triage. @Centri3 can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@rustbot ready

xFrednet:
Hey, this is a ping from triage. @Centri3  can you give this PR a review? It's totally fine if you don't have the time right now, you can reassign the PR to a random team member using `r? clippy`.

@rustbot ready

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,132 @@
...
+
+pub(super) fn check<'a>(
+    cx: &LateContext<'a>,
+    expr: &Expr<'a>,
+    recv: &Expr<'_>,
+    def: &Expr<'_>,
+    map: &Expr<'_>,
+    msrv: &Msrv,
+) {
+    if is_from_proc_macro(cx, expr) {
```
Centri3, line 45: Should be the final check before linting since this is rather expensive

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    let Bool(def_bool) = def_kind.node else {
+        return;
+    };
+
+    let variant = if is_type_diagnostic_item(cx, recv_ty, sym::Option) {
+        Variant::Some
+    } else {
+        Variant::Ok
+    };
```
Centri3, line 66: We can refactor this to instead call `type_diagnostic_name` in a match

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,132 @@
...
+    {
+        let wrap = variant.variant_name();
+        let comparator = op.node.as_str();
+
+        // we may need to add parens around the suggestion
+        // in case the parent expression has additional method calls,
+        // since for example `Some(5).map_or(false, |x| x == 5).then(|| 1)`
+        // being converted to `Some(5) == Some(5).then(|| 1)` isnt
+        // the same thing
+        let should_add_parens = get_parent_expr(cx, expr)
```
Centri3, line 97: This can still use `Sugg::maybe_par`

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,132 @@
...
+            .is_some_and(|expr| expr.precedence().order() > i8::try_from(AssocOp::Equal.precedence()).unwrap_or(0));
+        (
+            format!(
+                "{}{} {comparator} {wrap}({}){}",
+                if should_add_parens { "(" } else { "" },
+                snippet(cx, recv.span, ".."),
+                snippet(cx, non_binding_location.span.source_callsite(), ".."),
+                if should_add_parens { ")" } else { "" }
+            ),
+            "standard comparison",
```
Centri3, line 107:
```suggestion
            "a standard comparison",
```

clippy_lints/src/methods/unnecessary_map_or.rs
```diff
@@ -0,0 +1,132 @@
...
+    };
+
+    span_lint_and_sugg(
+        cx,
+        UNNECESSARY_MAP_OR,
+        expr.span,
+        "this `map_or` is redundant",
+        format!("use {method} instead"),
+        sugg,
+        Applicability::MachineApplicable,
```
Centri3, line 130: I just noticed that this might not be machine applicable. Don't we need to check for `PartialEq` somewhere if we'll change it to `opt == <something>`?

samueltardieu: Yes. This is not a problem for an `Option<T>` as we know that `T` already implements `PartialEq` if the original code was `x.map_or(false, |x| == value)`, and thus `Option<T>` implements `PartialEq` as well. However, for `Result<T, E>`, `E` must implement `PartialEq` as well, or the comparison cannot be done at the `Result<T, E>` level.

For example, the following code will lead to an incorrect suggestion:

```rust
fn main() {
    struct S;
    let r: Result<i32, S> = Ok(3);
    let _ = r.map_or(false, |x| x == 7);
}
```

because the suggested replacement `r == Ok(7)` won't work.

Jacherr:
@rustbot ready

Apologies for the slow turnover again. I hope this is the last of the issues now and this can finally be merged 😅

tests/ui/unnecessary_map_or.fixed
```diff
@@ -9,9 +9,9 @@ extern crate proc_macros;
 
 fn main() {
     // should trigger
-    let _ = Some(5) == Some(5);
-    let _ = Some(5) != Some(5);
-    let _ = Some(5) == Some(5);
+    let _ = (Some(5) == Some(5));
```
samueltardieu, line 12: These `(` should be removed, shouldn't they?

Jacherr: I think this is just an effect of `maybe_par` deciding they should be present. While I'd usually agree, I'm not sure there is a better way to tackle that?

Centri3: `maybe_par` is rather overeager, but `unused_parens` will usually catch them anyway. New lints triggering on fixed code in a cascade is fine

samueltardieu:
Is it time to open a FCP for this lint? I keep seeing `.map_or(false, …)` used everywhere.

Jacherr:
Squash done, ready for final comments/changes then hopefully can finally get this merged. Apologies again for how long this took.

samueltardieu:
@Jacherr It looks like `.map_or(true, …)` can be linted as well, using `.is_none_or(…)`. I played with it a bit but I have to go AFK right now, so feel free to integrate https://github.com/samueltardieu/rust-clippy/commit/a468081648451c93b1ae334e17923d7b062c2a90 in your code if you wish, and update all the lints which trigger this code. Otherwise, I'll submit a complete PR after yours is merged.

Note that tests that don't involve `PartialEq` are machine applicable as far as I can see, so I've changed this as well, but not looked closely at why the others aren't.

Centri3:
I am happy with merging this as-is if no concerns come up, and having that done later.

samueltardieu:
> I am happy with merging this as-is if no concerns come up, and having that done later.

This seems reasonable indeed.

samueltardieu:
@Jacherr Commit https://github.com/rust-lang/rust-clippy/pull/13653/commits/430235e80d022faefb866b1e8e193e1d49f1501f is an updated version of your commit compatible with the current `master`

samueltardieu:
(an alternative is to merge https://github.com/rust-lang/rust-clippy/pull/13653 which contains this PR, updated, and the `.map_or(true, …)` case as well)

Jacherr:
Hopefully, everything is in order for merge now. Ideally, as mentioned in the FCP, this can be merged and the additional `map_or(true` case added in post. It would be nice to put this PR to bed properly given how long it's been going back and forth.

samueltardieu (approved):

Centri3:
Thanks for your work on this :) I doubt anybody will object the lint being added so let's put it to rest :heart:

Issue #11493, opened 2023-09-13
[Tracking issue] Lints that are not grouped


blyxyas:
Currently, the lints that are not grouped are the following (got using a regex search for `define_clippy_lint!`)

- `disallowed_names.rs`
- `extra_unused_type_parameters.rs`
- `fallible_impl_from.rs`
- ~~`default_numeric_fallback.rs`~~
- `unused_unit.rs`
- `suspicious_doc_comments.rs`
- `implicit_saturating_add.rs`
- ~~`missing_doc.rs`~~
- `ignored_unit_patterns.rs`
- `almost_complete_range.rs`
- `derivable_impls.rs`
- ~~`if_then_some_else_none.rs`~~
- `excessive_nesting.rs`
- `manual_slice_size_calculation.rs`
- `enum_clike.rs`
- `default_instead_of_iter_empty.rs`
- ~~`error_impl_error.rs`~~
- `manual_is_ascii_check.rs`
- `implied_bounds_in_impls.rs`
- `default_constructed_unit_structs.rs`
- `redundant_async_block.rs`
- ~~`create_dir.rs`~~
- `large_stack_frames.rs`
- `semicolon_if_nothing_returned.rs`
- `reference.rs`
- `large_const_arrays.rs`
- `precedence.rs`
- `blocks_in_if_conditions.rs`
- `use_self.rs`
- `ptr_offset_with_cast.rs`
- `manual_main_separator_str.rs`
- `non_send_fields_in_send_ty.rs`
- ~~`missing_asserts_for_indexing.rs`~~
- `read_zero_byte_vec.rs`
- `option_env_unwrap.rs`
- `unnecessary_wraps.rs`
- `needless_else.rs`
- `items_after_test_module.rs`
- ~~`absolute_paths.rs`~~
- `double_parens.rs`
- ~~`unwrap_in_result.rs`~~
- `needless_for_each.rs`
- ~~`single_char_lifetime_names.rs`~~
- ~~`min_ident_chars.rs`~~
- `self_named_constructors.rs`
- `unnested_or_patterns.rs`
- `reserve_after_initialization.rs`
- `let_with_type_underscore.rs`
- `cognitive_complexity.rs`
- `escape.rs`
- ~~`multiple_unsafe_ops_per_block.rs`~~
- ~~`partial_pub_fields.rs`~~
- `items_after_statements.rs`
- ~~`large_include_file.rs`~~
- `disallowed_methods.rs`
[...]

PR #11798, opened 2023-11-13
move `suspicious_doc_comments` to doc pass

y21:
This was my first lint. I've been meaning to move it over to `doc.rs` since that's a better place.
There weren't any changes made to the lint logic itself.

I guess this can be considered part of #11493

changelog: none


flip1995:
The `doc.rs` file is now almost 1kLoC. It might make sense to split it up into a `doc/` module. This doesn't really have anything to do with this PR. Just some food for thought.

flip1995:
@bors r+

While I'm already here, I went ahead and reviewed the PR. Thanks for cleaning this up!

y21:
@flip1995 That's a good idea. Same for `attr.rs` probably?

flip1995:
`attrs.rs` also makes sense, looking at it. :+1:

PR #11801, opened 2023-11-13
Split `doc.rs` up into a subdirectory

y21:
So, first, sorry for the bad diff. 😅 

In #11798, @flip1995  suggested splitting `doc.rs` up, much like how we have the `methods/`, `matches/`, `types/` subdirectories.
I agree with this, the file is getting bigger as we add more and more doc lints that it makes sense to do this refactoring.

This is purely an internal change that moves things around a bit. 
(**EDIT:** depending on the outcome of https://github.com/rust-lang/rust-clippy/pull/11801#issuecomment-1816715615 , this may change the lint group name from `doc_markdoc` to `doc`).

I tried to not change any of the actual logic of the lints and as such some things weren't as easy to move to a separate file. So we still have some `span_lint*` calls in the `doc/mod.rs` file, which I think is fine. This is also the case in `methods/mod.rs`.

Also worth mentioning that the lints missing_errors_doc, missing_panics_doc, missing_safety_doc and unnecessary_safety_doc have a lot of the same logic so it didn't make much sense for each of these to be in their own file. Instead I just put them all in `missing_headers.rs`

I also added a bit of documentation to the involved `check_{attrs,doc}` methods.

changelog: none

### DIFF
--- a/clippy_lints/src/doc/suspicious_doc_comments.rs
+++ b/clippy_lints/src/doc/suspicious_doc_comments.rs
@@ -1,0 +1,48 @@
+use clippy_utils::diagnostics::span_lint_and_then;
+use rustc_ast::token::CommentKind;
+use rustc_ast::{AttrKind, AttrStyle, Attribute};
+use rustc_errors::Applicability;
+use rustc_lint::LateContext;
+use rustc_span::Span;
+
+use super::SUSPICIOUS_DOC_COMMENTS;
+
+pub fn check(cx: &LateContext<'_>, attrs: &[Attribute]) {
+    let replacements: Vec<_> = collect_doc_replacements(attrs);
+
+    if let Some((&(lo_span, _), &(hi_span, _))) = replacements.first().zip(replacements.last()) {
+        span_lint_and_then(
+            cx,
+            SUSPICIOUS_DOC_COMMENTS,
+            lo_span.to(hi_span),
+            "this is an outer doc comment and does not apply to the parent module or crate",
+            |diag| {
+                diag.multipart_suggestion(
+                    "use an inner doc comment to document the parent module or crate",
+                    replacements,
+                    Applicability::MaybeIncorrect,
+                );
+            },
+        );
+    }
+}
+
+fn collect_doc_replacements(attrs: &[Attribute]) -> Vec<(Span, String)> {
+    attrs
+        .iter()
+        .filter_map(|attr| {
+            if let AttrKind::DocComment(com_kind, sym) = attr.kind
+                && let AttrStyle::Outer = attr.style
+                && let Some(com) = sym.as_str().strip_prefix('!')
+            {
+                let sugg = match com_kind {
+                    CommentKind::Line => format!("//!{com}"),
+                    CommentKind::Block => format!("/*!{com}*/"),
+                };
+                Some((attr.span, sugg))
+            } else {
+                None
+            }
+        })
+        .collect()
+}
--- a/clippy_lints/src/doc/link_with_quotes.rs
+++ b/clippy_lints/src/doc/link_with_quotes.rs
@@ -1,0 +1,20 @@
+use std::ops::Range;
+
+use clippy_utils::diagnostics::span_lint;
+use rustc_lint::LateContext;
+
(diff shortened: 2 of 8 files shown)


clippy_lints/src/doc/suspicious_doc_comments.rs
```diff
@@ -0,0 +1,48 @@
...
+                    "use an inner doc comment to document the parent module or crate",
+                    replacements,
+                    Applicability::MaybeIncorrect,
+                );
+            },
+        );
+    }
+}
+
+fn collect_doc_replacements(attrs: &[Attribute]) -> Vec<(Span, String)> {
```
y21, line 30: The code in here was previously in the same function where we emit the lint but I extracted that into a separate function because I thought that looks better. If you disagree, I can change it back to have it all in the same function.

blyxyas:
Could you please split this into a lint per commit :pray: so that I can review better and the commit history is clearer (they won't be squashed on merge) (^..^)ﾉ

clippy_lints/src/doc/suspicious_doc_comments.rs
blyxyas: I don't disagree, but if we're not just moving some functions into other files, you could also refactor [this commit](https://github.com/rust-lang/rust-clippy/pull/11801/commits/709ab04e510b2ff62ecd745fe9ad2225abd5d21e).

blyxyas (commented):
Is there any other code change other than just changing code to files? If there isn't (outside that comment), I think this is pretty much ready.

y21:
> Is there any other code change other than just changing code to files?

Only in `suspicious_doc_comments`. I also added a `FindPanicUnwrap::find_span` function that can be used in a bunch of places in the main `mod.rs`.
Other lints are unchanged.

y21:
Oh, the lint group was also changed from `DocMarkdown` to `Doc`. I initially thought that could be a small internal change that the user wouldn't notice, but `#![warn(clippy::doc_markdown)]` works, and won't anymore after this. Should I just change it back? I think doc is a better name for the group since it's really not about markdown anymore at that point, but I'm not sure if that's reason enough to do this change.

blyxyas:
I'll nominate this issue, let's see what the team thinks about this. If we were to change its name, it would need to be deprecated some time before being removed =＾● ⋏ ●＾=

y21:
If this ends up blocking the PR, I guess I could also just revert it for this PR, just so that this can remain an "internal-only" PR.
Renaming from `DocMarkdown` to `Doc`, if the team agrees, could be its own PR since that's a rather small thing.

Since this touches quite a lot of code and is prone to bit rot whenever someone makes a change to doc lints (and solving conflicts here is kind of a chore), I wouldn't want to leave this open for *too* long

blyxyas:
Okis, rename the pass to `DocMarkdown` for now, we'll still discuss it with the rest of the team, but let's keep the ball rolling (=^･ω･^)y＝

clippy_lints/src/doc/suspicious_doc_comments.rs
y21: @blyxyas ? :)

blyxyas (approved):
LGTM, thanks! :heart: I've revised these like 5 times and I still feel like there's something obvious that I don't see. Quite *the GuillaumeGomez syndrome* (=^-ω-^=)

GuillaumeGomez:
:laughing: Careful, it doesn't get better with time. ^^'

blyxyas:
Oh no, I just remember that I said that they wouldn't be squashed ;(
Sorry for the chore, Timo, it's just my Alzheimer. I feel guilty now D:

y21:
Did bors get stuck? 🤔 
It's been going for a while

flip1995:
It lost track of this PR somehow. I re-synchronized bors for Clippy. Let's see if that helps.

flip1995:
Now it at least shows up as `approved` in the queue. Let's 

@bors retry

so that it will be merged.

flip1995:
First bors didn't want to do anything and now they do it twice. Weird fella.

PR #11802, opened 2023-11-13
`needless_return_with_question_mark` ignore let-else

### LINKED ISSUE #11765: Incorrect suggestion for `needless_return_with_question_mark` with `let else`
### Summary

In `let-else` construct, `else` branch must always diverge. Sometimes, it is done by returning an error. Sometimes, this error can be converted to the upper-layer error with the question mark. Clippy doesn't like `return` and a question mark in the same line, but it suggests to drop the `return`, which causes compilation error.

Initially found [on URLO](https://users.rust-lang.org/t/analyzer-quirk/102200).

### Lint Name

needless_return_with_question_mark

### Reproducer

I tried this code:

```rust
fn foo() -> Result<(), ()> {
    let Some(s) = Some("") else {
        return Err(())?;
    };
    
    Ok(())
}
```

I saw this happen:

```
warning: unneeded `return` statement with `?` operator
 --> src/lib.rs:3:9
  |
3 |         return Err(())?;
  |         ^^^^^^^ help: remove it
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_return_with_question_mark
  = note: `#[warn(clippy::needless_return_with_question_mark)]` on by default
```

Problem is that the suggestion here is incorrect - it will prevent code from compiling, since `else` branch of `let-else` will no longer diverge.

It _might_ be a good idea to suggest `return Err(()).map_err(Into::into)` instead, but as-is it's a false positive.

### Version

_No response_

### Additional Labels

dswij:
Fixes #11765

This PR makes `needless_return_with_question_mark` to ignore expr inside let-else.

changelog: [`needless_return_with_question_mark`] ignore let-else


clippy_lints/src/returns.rs
```diff
@@ -435,3 +437,13 @@ fn extend_span_to_previous_non_ws(cx: &LateContext<'_>, sp: Span) -> Span {
...
+
+fn is_let_else_block(tcx: TyCtxt<'_>, e: &Expr<'_>) -> bool {
+    for (_, node) in tcx.hir().parent_iter(e.hir_id) {
+        if let Node::Local(Local { els: Some(_), .. }) = node {
+            return true;
+        }
+    }
+
+    false
+}
```
xFrednet, line 449: Pretty nit picky:

I feel like the `if let Node::Local(Local { els: Some(_), .. })` check could be a bit too general, as it doesn't check, that the expression comes from the else branch:

```rs
fn x() -> Result<(), ()> {
    let Some(_): Option<()> = ({
        return Err(())?;
    }) else {
        panic!();
    };
}
```

Even if that's pretty weird code...

Could you try if this works:

```rs
fn is_let_else_block(tcx: TyCtxt<'_>, e: &Expr<'_>) -> bool {
    let mut child_id = e.hir_id;
    for (parent_id, node) in tcx.hir().parent_iter(e.hir_id) {
        if let Node::Local(Local { els: Some(els), .. }) = node
            && els.hir_id == child_id
        {
            return true;
        }
        child_id = parent_id;
    }

    false
}
```

Then we could probably move this into `clippy_utils` in case other lints need something like this :). If not, it's totally fine. I think this code is already good enough of a heuristic, to avoid FPs and r+ the PR :)

dswij: 🤔 yeah fair enough, if we're going to move this to util, then this needs to be as precise as possible.

```rust
fn x() -> Result<(), ()> {
    let Some(_): Option<()> = ({
        return Err(())?;
    }) else {
        panic!();
    };
}
```

This is kind of outrageous for an FN of this lint, there's more to worry about than a needless return 😅

dswij: Oh, just remembered that we'd need to ignore this as well

```rust
fn x() -> Result<(), ()> {
    let Some(_): Option<()> = ({
        return Err(())?;
    }) else {
        panic!();
    };
}
```

removing the return here leads to more complications than benefits; e.g. here it leads to compilation error, and if we have the same type, it might change the behavior.

> Could you try if this works:

But this does work

xFrednet (approved):

xFrednet:
LGTM, thank you for the update :)

@bors r+

Issue #3770, opened 2019-02-16
if_same_then_else Should not warn for `else if` blocks.


kevincox:
```rust
if condition1() {
  1
} else if condition2() {
  1
} else if condition3() {
  2
} else {
  3
}
```

There are many cases where code like this shouldn't be warned on. While you could do `condition1() || condition2()` often times it is more clear of the intent if they are separate. Especially if the conditions are non-trivial compared to the value being returned.

I think the right time for this warning to warn is if the last `if` has the same condition as the `else`. For example the following two cases would warn.

```rust
if condition1() {
  1
} else {
  1
}
```
```rust
if condition1() {
  1
} else if condition2() {
  2
} else {
  2
}
```

flip1995:
IMHO this lint is good as it is. If you have a situation where separating the if conditions makes the logic more readable you should allow this lint.

From my experience this lint has in fact two purposes:
1. The one that you described, which is: "Hey, you can merge this if conditions"
2. These two if bodies are the same, but maybe that's unintended

Example for 2.:
```rust
if c1 {
    // some code
} else if c2 {
    // some other code
} else if c3 {
    // same code as 1, because copy pasted, but forget to adapt that one line
} else {
    // ...
}
```

If we'd change the lint in a way that only consecutive branches would get linted, the second case would never be caught. And the second case can even lead to bugs, while the first one is just code style.

---
TL;DR: I wouldn't change the lint, but add the second point to the "Why is this bad?" section of the documentation.

kevincox:
In the example you gave I also think that it would have too many false positives. There are lots of cases where the actual value is trivial compared to the complexity of the condition so you wouldn't want to merge it. Additionally the more conditions that are between the two identical branches the more complicated the condition would have to become. The "correct" fix in the example you provided is:

```rust
If c1 || (c3 && !c2) {
  // some code
} else if c2 {
 // some other code
} else {
  // ...
}
```

Unless the code in the off statement is quite complicated this is much harder to understand. And if the conditions have side effects it isn't even correct. 

Also I'm not proposing warning on any consecutive branches. I would only warn when the **last two** branches are identical because in this case you could just omit the condition which will (almost) always result in simpler code. 

I guess it comes down to if you want to avoid false positives or false negatives. Maybe I can propose that this is split into two lints. One will warn when two consecutive branches are identical but not be in `all`. Then the one that only warns for the last two branches only can be included in `all`. I think this is reasonable because `all` claims "everything that has no false positives".

[...]

PR #11809, opened 2023-11-14
Change `if_same_then_else` to be a `style` lint

hrxi:
CC #3770

From https://github.com/rust-lang/rust-clippy/issues/3770#issuecomment-687565594 (@flip1995):

> Oh I thought I replied to this: I definitely see now that having this
> as a correctness lint might be the wrong categorization. What we might
> want to do is to just allow this lint, if there are comments in the
> arm bodies. But a good first step would be to downgrade this lint to
> style or complexity. I would vote for style since merging two arms is
> not always less complex.

changelog: [`if_same_then_else`]: Change to be a `style` lint


Alexendoo (approved):

Alexendoo:
Yeah seems reasonable to move it out of correctness

@bors r+

Issue #11709, opened 2023-10-24
undocumented_unsafe_blocks doesn't detect safety comment above associated constant


ketsuban:
### Summary

I make use of the `voladdress` crate which uses a pattern of "unsafe construction, safe use", but when I put a `VolAddress` in a constant clippy doesn't detect the safety comment I add above it.

### Reproducer

I tried this code:

```rust
#![warn(clippy::undocumented_unsafe_block)]

use voladdress::{Safe, VolAddress};

struct Control(u16);

impl Control {
    // SAFETY: this is safe
    const REGISTER: VolAddress<Self, Safe, Safe> = unsafe { VolAddress::new(0x0400_0204) };
}
```

I expected to see this happen:
no clippy warning

Instead, this happened:
a clippy warning

### Version

```text
rustc 1.75.0-nightly (1c05d50c8 2023-10-21)
binary: rustc
commit-hash: 1c05d50c8403c56d9a8b6fb871f15aaa26fb5d07
commit-date: 2023-10-21
host: x86_64-unknown-linux-gnu
release: 1.75.0-nightly
LLVM version: 17.0.3
```

PR #11813, opened 2023-11-15
fix [`undocumented_unsafe_block`] and [`unnecessary_safety_comment`]  not detecting associated constants

J-ZhengLi:
fixes: #11709 

changelog: fix FP on [`undocumented_unsafe_block`] and [`unnecessary_safety_comment`]

### DIFF
--- a/clippy_lints/src/undocumented_unsafe_blocks.rs
+++ b/clippy_lints/src/undocumented_unsafe_blocks.rs
@@ -4,10 +4,9 @@ use clippy_utils::diagnostics::span_lint_and_help;
 use clippy_utils::source::walk_span_to_context;
 use clippy_utils::visitors::{for_each_expr_with_closures, Descend};
 use clippy_utils::{get_parent_node, is_lint_allowed};
-use hir::HirId;
 use rustc_data_structures::sync::Lrc;
 use rustc_hir as hir;
-use rustc_hir::{Block, BlockCheckMode, ItemKind, Node, UnsafeSource};
+use rustc_hir::{Block, BlockCheckMode, HirId, ItemKind, Node, OwnerNode, UnsafeSource};
 use rustc_lexer::{tokenize, TokenKind};
 use rustc_lint::{LateContext, LateLintPass, LintContext};
 use rustc_middle::lint::in_external_macro;
@@ -185,19 +184,7 @@ impl<'tcx> LateLintPass<'tcx> for UndocumentedUnsafeBlocks {
             return;
         }
 
-        let mk_spans = |pos: BytePos| {
-            let source_map = cx.tcx.sess.source_map();
-            let span = Span::new(pos, pos, SyntaxContext::root(), None);
-            let help_span = source_map.span_extend_to_next_char(span, '\n', true);
-            let span = if source_map.is_multiline(item.span) {
-                source_map.span_until_char(item.span, '\n')
-            } else {
-                item.span
-            };
-            (span, help_span)
-        };
-
-        let item_has_safety_comment = item_has_safety_comment(cx, item);
+        let item_has_safety_comment = item_has_safety_comment(cx, item.hir_id(), item.into());
         match (&item.kind, item_has_safety_comment) {
             // lint unsafe impl without safety comment
             (hir::ItemKind::Impl(impl_), HasSafetyComment::No) if impl_.unsafety == hir::Unsafety::Unsafe => {
@@ -224,7 +211,7 @@ impl<'tcx> LateLintPass<'tcx> for UndocumentedUnsafeBlocks {
             // lint safe impl with unnecessary safety comment
             (hir::ItemKind::Impl(impl_), HasSafetyComment::Yes(pos)) if impl_.unsafety == hir::Unsafety::Normal => {
                 if !is_lint_allowed(cx, UNNECESSARY_SAFETY_COMMENT, item.hir_id()) {
-                    let (span, help_span) = mk_spans(pos);
+                    let (span, help_span) = mk_spans(cx, item.span, pos);
 
                     span_lint_and_help(
                         cx,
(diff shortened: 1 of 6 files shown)


J-ZhengLi:
The code seems confusing for me... TBO I had trouble with identifying the problem twice :smile: (happened on a previous PR also). I think it could use some refactoring, which I might open a separated PR if that's ok? I just afraid if I put them here will make the changes less visible...

blyxyas:
Sorry, I don't really have the time to review a PR with this size (=^-ω-^=) I have some important deadlines before mid-december and a lot of work. Sorry for the wait :heart: 

r? @rust-lang/clippy

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -609,24 +732,32 @@ fn get_body_search_span(cx: &LateContext<'_>) -> Option<Span> {
...
+                    ItemKind::Mod(_) | ItemKind::Trait(..) | ItemKind::Impl(..) => span = item.span,
+                    _ => (),
+                }
                 break;
             },
-            Node::Crate(mod_) if maybe_global_var => {
+            Node::Crate(mod_) if is_const_or_static => {
                 span = mod_.spans.inner_span;
             },
             _ => break,
```
Alexendoo, line 763: Not caused by your PR but I found this function a bit confusing, what do you think of

```rust
fn get_body_search_span(cx: &LateContext<'_>) -> Option<Span> {
    let body = cx.enclosing_body?;
    for (_, parent_node) in cx.tcx.hir().parent_iter(body.hir_id) {
        match parent_node {
            Node::Item(hir::Item {
                kind: hir::ItemKind::Const(..) | ItemKind::Static(..),
                ..
            })
            | Node::ImplItem(hir::ImplItem {
                kind: hir::ImplItemKind::Const(..),
                ..
            })
            | Node::TraitItem(hir::TraitItem {
                kind: hir::TraitItemKind::Const(..),
                ..
            }) => {},
            Node::Item(item) => return Some(item.span),
            Node::Crate(mod_) => return Some(mod_.spans.inner_span),
            _ => {},
        }
    }
    None
}
```

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -440,26 +518,33 @@ enum HasSafetyComment {
...
-fn item_has_safety_comment(cx: &LateContext<'_>, item: &hir::Item<'_>) -> HasSafetyComment {
-    match span_from_macro_expansion_has_safety_comment(cx, item.span) {
+fn item_has_safety_comment(cx: &LateContext<'_>, hir_id: HirId, owner: hir::OwnerNode<'_>) -> HasSafetyComment {
+    match span_from_macro_expansion_has_safety_comment(cx, owner.span()) {
         HasSafetyComment::Maybe => (),
         has_safety_comment => return has_safety_comment,
     }
 
-    if item.span.ctxt() != SyntaxContext::root() {
+    if owner.span().ctxt() != SyntaxContext::root() {
```
Alexendoo, line 527: Another nit while we're in the area

```suggestion
    if owner.span().from_expansion() {
```

clippy_lints/src/undocumented_unsafe_blocks.rs
```diff
@@ -542,32 +627,70 @@ fn stmt_has_safety_comment(cx: &LateContext<'_>, span: Span, hir_id: HirId) -> H
...
+impl CheckableItemId for hir::ImplItemRef {
+    fn owner_id(&self) -> hir::OwnerId {
+        self.id.owner_id
+    }
+}
+impl CheckableItemId for hir::TraitItemRef {
+    fn owner_id(&self) -> hir::OwnerId {
+        self.id.owner_id
+    }
+}
```
Alexendoo, line 648: An option would be to have `ids` be an iterator rather than a slice, and find the previous node with `ids.take_while(|id| id != owner_id).last()`, if it's `None` then you're in the first entry

May or may not be nicer than the traits + indexing into the slice, so feel free to keep it as is if you prefer it as is

clippy_lints/src/undocumented_unsafe_blocks.rs
J-ZhengLi: wow this is interesting, I would never realize that they  does the same thing, really appreciate your effort in this~ :smile:

J-ZhengLi: I think it's awesome, I got confused first but I only just rename a couple variables then add my code on top of it and call it a day, so,  thank you :smile:

clippy_lints/src/undocumented_unsafe_blocks.rs
J-ZhengLi: Seems better and a lot shorter, thanks, although the logic is harder to understand, took my dumb brain a while to figure out what is happening and why it works :rofl:

Alexendoo: The current `enumerate().find_map()` finds the index of the current id then subtracts 1 to get the index of the previous node, may be clearer as ~

```rust
let idx = ids.iter().pos(|id| id.owner_id() == owner.def_id())?;
if idx == 0 {
    walk_span_to_context(search_span, SyntaxContext::root()).map(Span::lo)
} else {
    let span = cx.tcx.hir().span(ids[idx - 1]);
    walk_span_to_context(span, SyntaxContext::root()).map(Span::lo)
}
```

The `.take_while().last()` finds the id *preceding* the current one so doesn't require that re-indexing. If it's `None` then the current id was the first in `siblings`, since the `take_while` would return an iterator of 0 items


To make it clearer you could have it be `if let Some(prev_sibling_id) = siblings...`

J-ZhengLi: > The `.take_while().last()` finds the id _preceding_ the current one so doesn't require that re-indexing...

Thanks, wouldn't the preceding id be a one that `!= owner_id` tho? I guess what gets me confused is that it was previously accepting only when it `== owner_id`

Alexendoo: Say you have an iterator of `['a', 'b', 'c', 'd', 'e']`, `.take_while(|n| n != 'd')` would give you

```text
['a', 'b', 'c', 'd', 'e']
 ^^^^^^^^^^^^^ an iterator of these three items
```

The `.last()` would then give you the `3`

Feel free to use a different way to find the previous item if you think of a clearer one though, another way could be to make it a `DoubleEndedIterator` and do

```rust
siblings.rfind(|id| id == owner_id);
if let Some(prev_sibling_id) = siblings.next_back() {
...
```

J-ZhengLi: ahhh, that make so much more sense! thanks again for your patience. 

Also I ran a simple bench with these two methods and found out that using `DoubleEndedIterator` is faster, so I think I'm gonna use that

xFrednet:
Hey @J-ZhengLi, this is a ping from triage, since there hasn't been any activity in some time. Could you rebase on master?

@rustbot author

J-ZhengLi:
Oh nooo, because I already forgot what I did thus it's kinda hard to distinguish which lines are mine and which are add by others after this PR, at this point it might even be easier if I redo the whole thing (but with less changes I promise). It should be a simple fix tho, I don't know what have I done to make it this complicated 🤦‍♂️

I'm gonna close this and open a new PR tomorrow (hopefully) fixing this again, and sorry @Alexendoo for wasting effort XD

PR #11818, opened 2023-11-15
[`redundant_guards`]: catch `is_empty`, `starts_with` and `ends_with` on slices and `str`s

### LINKED ISSUE #11807: redundant_guards could lint against more checks in the standard library
### Summary

The [redundant_guards](https://rust-lang.github.io/rust-clippy/master/index.html#/redundant_guards) lint could lint against more checks than it does.

Currently it only lints against `if <capture> == pattern`, but it could also lint against some `if` guards that use `std` functions, notably `str::is_empty` and `slice::is_empty`.

### Lint Name

redundant_guards

### Reproducer

I tried this code:

```rust
fn inner() -> Option<(&'static str> {
    todo!();
}
let a = inner();
match a {
    Some(s) if s.is_empty() => todo!("1"),
    Some(_) => todo!("5"),
    None => todo!("4")
}
```

I expected the `redundant_guard` lint to trigger and suggest using `Some("") => todo!("1"),` in the first `match` arm.

Instead, the lint didn't trigger.

### Version

```text
Nightly version: 1.76.0-nightly

(2023-11-13 ba7c7a301984967c8c13)

(From playground)
```

y21:
Fixes #11807

Few things worth mentioning:
- Taking `snippet`s is now done at callsite, instead of passing a span and doing it in `emit_redundant_guards`. This is because we now need custom suggestion strings in certain places, like `""` for `str::is_empty`.
- This now uses `snippet` instead of `snippet_with_applicability`. I don't think this really makes any difference for `MaybeIncorrect`, though?
- This could also lint byte strings, as they're of type `&[u8; N]`, but that can be ugly so I decided to leave it out for now

changelog: [`redundant_guards`]: catch `str::is_empty`, `slice::is_empty`, `slice::starts_with` and `slice::ends_with`

### DIFF
--- a/clippy_lints/src/manual_async_fn.rs
+++ b/clippy_lints/src/manual_async_fn.rs
@@ -187,14 +187,11 @@ fn desugared_async_block<'tcx>(cx: &LateContext<'tcx>, block: &'tcx Block<'tcx>)
 }
 
 fn suggested_ret(cx: &LateContext<'_>, output: &Ty<'_>) -> Option<(&'static str, String)> {
-    match output.kind {
-        TyKind::Tup(tys) if tys.is_empty() => {
-            let sugg = "remove the return type";
-            Some((sugg, String::new()))
-        },
-        _ => {
-            let sugg = "return the output of the future directly";
-            snippet_opt(cx, output.span).map(|snip| (sugg, format!(" -> {snip}")))
-        },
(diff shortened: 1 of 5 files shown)


llogiq:
This looks good to me. Thank you!

@bors r+

PR #11821, opened 2023-11-16
Extend `maybe_misused_cfg` lint over `cfg(test)`

### LINKED ISSUE #11240: Extend `maybe_misused_cfg` to cover `#[cfg(tests)]`
### Description

An easy misspelling to make is to write something like:

```rs
pub fn add(left: usize, right: usize) -> usize {
    left + right
}

#[cfg(tests)]
mod tests {
    use super::*;

    #[test]
    fn it_works() {
        let result = add(2, 2);
        assert_eq!(result, 4);
    }
}
```

This will not result in any warnings or errors currently, but will simply not run any of the tests. Clippy should instead output a warning:

```
warning: test may misspelled as tests
 --> src/lib.rs:5:7
  |
5 | #[cfg(tests)]
  |       ^^^^^ help: use: `test`
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/
index.html#maybe_misused_cfg
  = note: `#[warn(clippy::maybe_misused_cfg)]` on by default
```

### Version

```text
rustc 1.71.0 (8ede3aae2 2023-07-12)
binary: rustc
commit-hash: 8ede3aae28fe6e4d52b38157d7bfe0d3bceef225
commit-date: 2023-07-12
release: 1.71.0
LLVM version: 16.0.5
```

### Additional Labels

@rustbot label +C-enhancement

GuillaumeGomez:
Fixes #11240.

One thought I had is that we could use the levenshtein distance (of 1) to ensure this is indeed `test` that was targeted. But maybe it's overkill, not sure.

changelog: [`maybe_misused_cfg`]: Extend lint over `cfg(test)`

r? @blyxyas

### DIFF
--- a/clippy_lints/src/attrs.rs
+++ b/clippy_lints/src/attrs.rs
@@ -407,20 +407,26 @@ declare_clippy_lint! {
     /// Checks for `#[cfg(features = "...")]` and suggests to replace it with
     /// `#[cfg(feature = "...")]`.
     ///
+    /// It also checks if `cfg(test)` was misspelled.
+    ///
     /// ### Why is this bad?
-    /// Misspelling `feature` as `features` can be sometimes hard to spot. It
+    /// Misspelling `feature` as `features` or `test` as `tests` can be sometimes hard to spot. It
     /// may cause conditional compilation not work quietly.
     ///
     /// ### Example
     /// ```no_run
     /// #[cfg(features = "some-feature")]
     /// fn conditional() { }
+    /// #[cfg(tests)]
+    /// mod tests { }
     /// ```
     ///
     /// Use instead:
     /// ```no_run
     /// #[cfg(feature = "some-feature")]
     /// fn conditional() { }
+    /// #[cfg(test)]
+    /// mod tests { }
     /// ```
     #[clippy::version = "1.69.0"]
     pub MAYBE_MISUSED_CFG,
--- a/clippy_lints/src/attrs.rs
+++ b/clippy_lints/src/attrs.rs
@@ -946,6 +946,19 @@ fn check_nested_misused_cfg(cx: &EarlyContext<'_>, items: &[NestedMetaItem]) {
             }
             if let MetaItemKind::List(list) = &meta.kind {
                 check_nested_misused_cfg(cx, list);
+            // If this is not a list, then we check for `cfg(test)`.
+            } else if let Some(ident) = meta.ident()
+                && matches!(ident.name.as_str(), "tests" | "Test")
+            {
+                span_lint_and_sugg(
+                    cx,
+                    MAYBE_MISUSED_CFG,
+                    meta.span,
+                    &format!("'test' may be misspelled as '{}'", ident.name.as_str()),
+                    "do you mean",
+                    "test".to_string(),
+                    Applicability::MaybeIncorrect,
+                );
             }
         }
     }
--- a/tests/ui/cfg_features.fixed
+++ b/tests/ui/cfg_features.fixed
@@ -14,4 +14,16 @@ fn main() {
     //~^ ERROR: feature may misspelled as features
     //~| ERROR: feature may misspelled as features
     let _ = 1 + 2;
+
+    #[cfg(test)]
+    //~^ ERROR: 'test' may be misspelled as 'tests'
+    let _ = 2;
+    #[cfg(test)]
+    //~^ ERROR: 'test' may be misspelled as 'Test'
+    let _ = 2;
+
+    #[cfg(all(test, test))]
+    //~^ ERROR: 'test' may be misspelled as 'tests'
+    //~| ERROR: 'test' may be misspelled as 'Test'
+    let _ = 2;
 }
--- a/tests/ui/cfg_features.rs
+++ b/tests/ui/cfg_features.rs
@@ -14,4 +14,16 @@ fn main() {
     //~^ ERROR: feature may misspelled as features
     //~| ERROR: feature may misspelled as features
     let _ = 1 + 2;
+
+    #[cfg(tests)]
+    //~^ ERROR: 'test' may be misspelled as 'tests'
+    let _ = 2;
+    #[cfg(Test)]
+    //~^ ERROR: 'test' may be misspelled as 'Test'
+    let _ = 2;
+
+    #[cfg(all(tests, Test))]
+    //~^ ERROR: 'test' may be misspelled as 'tests'
+    //~| ERROR: 'test' may be misspelled as 'Test'
+    let _ = 2;
 }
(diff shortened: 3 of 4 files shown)


blyxyas (changes requested):
Pretty good first iteration (^･o･^)ﾉ”

clippy_lints/src/attrs.rs
```diff
@@ -946,6 +946,19 @@ fn check_nested_misused_cfg(cx: &EarlyContext<'_>, items: &[NestedMetaItem]) {
...
+            // If this is not a list, then we check for `cfg(test)`.
+            } else if let Some(ident) = meta.ident()
+                && (ident.name == sym!(tests) || ident.name == sym!(Test))
+            {
+                span_lint_and_sugg(
+                    cx,
+                    MAYBE_MISUSED_CFG,
+                    meta.span,
+                    &format!("test may be misspelled as {}", ident.name.as_str()),
+                    "use",
```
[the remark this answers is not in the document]
GuillaumeGomez: :+1:

clippy_lints/src/attrs.rs
```diff
@@ -407,20 +407,28 @@ declare_clippy_lint! {
...
     /// `#[cfg(feature = "...")]`.
     ///
+    /// It also checks if `cfg(test)` was misspelled.
+    ///
     /// ### Why is this bad?
     /// Misspelling `feature` as `features` can be sometimes hard to spot. It
     /// may cause conditional compilation not work quietly.
     ///
+    /// Same goes for misspelling `test`.
+    ///
```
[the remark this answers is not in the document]
GuillaumeGomez: :+1:

GuillaumeGomez:
Thanks for the improvements, updated!

clippy_lints/src/attrs.rs
```diff
@@ -946,6 +946,19 @@ fn check_nested_misused_cfg(cx: &EarlyContext<'_>, items: &[NestedMetaItem]) {
             }
             if let MetaItemKind::List(list) = &meta.kind {
                 check_nested_misused_cfg(cx, list);
+            // If this is not a list, then we check for `cfg(test)`.
+            } else if let Some(ident) = meta.ident()
+                && (ident.name == sym!(tests) || ident.name == sym!(Test))
```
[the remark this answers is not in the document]
blyxyas: It's only like 1 millisecond of a difference, the main advantage is that it can be checked against more easily (like in this case)

clippy_lints/src/attrs.rs
```diff
@@ -946,6 +954,19 @@ fn check_nested_misused_cfg(cx: &EarlyContext<'_>, items: &[NestedMetaItem]) {
             }
             if let MetaItemKind::List(list) = &meta.kind {
                 check_nested_misused_cfg(cx, list);
+            // If this is not a list, then we check for `cfg(test)`.
+            } else if let Some(ident) = meta.ident()
+                && matches!(ident.name.as_str(), "tests" | "Test")
```
emilk, line 959: Instead of deny-listing `tests` and `Test`, would it be possible to allow-list all configs passed in by `--cfg` plus the built-in ones (`test`, `debug_assertions`, …)?

clippy_lints/src/attrs.rs
```diff
@@ -946,6 +954,19 @@ fn check_nested_misused_cfg(cx: &EarlyContext<'_>, items: &[NestedMetaItem]) {
...
+            {
+                span_lint_and_sugg(
+                    cx,
+                    MAYBE_MISUSED_CFG,
+                    meta.span,
+                    &format!("test may be misspelled as {}", ident.name.as_str()),
```
emilk, line 965:
```suggestion
                    &format!("'test' may be misspelled as '{}'", ident.name.as_str()),
```

emilk (approved):
Wonderful ❤️

clippy_lints/src/attrs.rs
```diff
@@ -407,20 +407,28 @@ declare_clippy_lint! {
...
     /// `#[cfg(feature = "...")]`.
     ///
+    /// It also checks if `cfg(test)` was misspelled.
+    ///
     /// ### Why is this bad?
-    /// Misspelling `feature` as `features` can be sometimes hard to spot. It
+    /// Misspelling `feature` as `features` or `test` as `tests` can be sometimes hard to spot. It
     /// may cause conditional compilation not work quietly.
     ///
+    /// Same goes for misspelling `test`.
```
blyxyas, line 416: This should be removed, we already mentioned `test` and `tests` in the earlier paragraph

blyxyas (approved):
Just a nit and I think this is merge-ready!

clippy_lints/src/attrs.rs
GuillaumeGomez: I'll check for the full list then.

GuillaumeGomez: So back to this, I'm not sure it's a good approach because you might use a valid `cfg` that is not passed to the compiler (so it's `cfg`ed-out) and in this case it would be considered as an error since this `cfg` isn't known to clippy. So false-positive.

However on the other hand, it's already possible to have this issue currently, so not sure if we want to dive into this any further.

I think a next iteration on this would be to eventually check for all existing ones (the builtin ones and the ones provided to rustc) but we need to check pros and cons.

Does that sound acceptable?

clippy_lints/src/attrs.rs
GuillaumeGomez: :+1:

clippy_lints/src/attrs.rs
GuillaumeGomez: :+1:

GuillaumeGomez:
I applied suggestions except one which I think deserves more discussion and should happen in a follow-up PR. What do you think?

clippy_lints/src/attrs.rs
emilk: You are right; catching all `cfg` errors is definitely a different and much larger issue

emilk (approved):

GuillaumeGomez:
I added the missing changelog in my first comment. Sorry about that.

GuillaumeGomez:
I was planning to open an issue to have a discussion about a wider check for all `cfg`s. However it would have brought issues in case what we thought was a typo was actually another `cfg` which was not activated.

But luckily for us, I just learned that the `--check-cfg` option in `cargo` is being stabilized so this issue just got solved without needing us to do anything. \o/

GuillaumeGomez:
Here's the cargo issue: https://github.com/rust-lang/rust/issues/82450

clippy_lints/src/attrs.rs
pietroalbini: Isn't that https://doc.rust-lang.org/beta/unstable-book/compiler-flags/check-cfg.html?

GuillaumeGomez: Yes it is. I discovered it afterwards and I linked to the tracking issue in my last comment of this PR.

PR #11826, opened 2023-11-16
Fix typos in recent lint documentation.

kpreid:
Fixes typos and markup errors, and also makes the examples more realistic by hiding the `;`s so as not to visibly be discarding the computed value. Affected lints:

* `redundant_as_str`
* `unnecessary_map_on_constructor`

changelog: none

### DIFF
--- a/clippy_lints/src/unnecessary_map_on_constructor.rs
+++ b/clippy_lints/src/unnecessary_map_on_constructor.rs
@@ -10,11 +10,11 @@ use rustc_span::sym;
 declare_clippy_lint! {
     /// ### What it does
     /// Suggests removing the use of a `map()` (or `map_err()`) method when an `Option` or `Result`
-    ///  is being constructed.
+    /// is being constructed.
     ///
     /// ### Why is this bad?
-    /// It introduces unnecessary complexity. In this case the function can be used directly and
-    /// construct the `Option` or `Result` from the output.
+    /// It introduces unnecessary complexity. Instead, the function can be called before
+    /// constructing the `Option` or `Result` from its return value.
     ///
     /// ### Example
     /// ```no_run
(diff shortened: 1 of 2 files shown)


kpreid:
The original typo was fixed by the other PR but this PR still contains other improvements. Rebased.

clippy_lints/src/unnecessary_map_on_constructor.rs
```diff
@@ -9,19 +9,22 @@ use rustc_span::sym;
 
 declare_clippy_lint! {
     /// ### What it does
-    /// Suggest removing the use of a map (or map_err) method when an Option or Result is being constructed.
+    /// Suggests removing the use of a `map()` (or `map_err()`) method when an `Option` or `Result`
+    ///  is being constructed.
```
Alexendoo, line 13: There's an extra space here

clippy_lints/src/unnecessary_map_on_constructor.rs
```diff
@@ -9,19 +9,22 @@ use rustc_span::sym;
...
     ///
     /// ### Why is this bad?
     /// It introduces unnecessary complexity. In this case the function can be used directly and
-    /// construct the Option or Result from the output.
+    /// construct the `Option` or `Result` from the output.
```
Alexendoo, line 17: I think it'd be good to reword the second sentence a bit here, maybe something like

> In this case the function can be used directly in the `Some`, `Ok` or `Err` constructors.

clippy_lints/src/unnecessary_map_on_constructor.rs
kpreid: Fixed.

clippy_lints/src/unnecessary_map_on_constructor.rs
kpreid: I put in a different new wording:

> It introduces unnecessary complexity. Instead, the function can be called before
> constructing the `Option` or `Result` from its return value.

Alexendoo (approved):

Issue #11438, opened 2023-08-31
Detects usage of infinite loops without indicating that function never returns


J-ZhengLi:
### What it does

checks for infinite loops

### Advantage

Could reduce unintentional mistakes where the programmer simply forgor to break

### Drawbacks

Perhaps being noisy when the infinite loops are used as signal handling, or waiting server respond etc...

### Example

```rust
fn loop_this() {
    loop {
        // do something
    }
}
```

Could be written as:

```rust
fn loop_this() {
    loop {
        // do something
        if condition {
            break; // or return
        }
    }
}
```

Or

```rust
fn loop_this() -> ! {
    loop {
        // do something
    }
}
```

PR #11829, opened 2023-11-17
new lint to detect infinite loop

J-ZhengLi:
closes: #11438 

changelog: add new lint to detect infinite loop

~*I'll change the lint name*~. Should I name it  `infinite_loop` or `infinite_loops` is fine? Ahhhh, English is hard...

### DIFF
--- a/clippy_lints/src/loops/infinite_loops.rs
+++ b/clippy_lints/src/loops/infinite_loops.rs
@@ -34,27 +34,15 @@ pub(super) fn check<'tcx>(
         return;
     }
 
-    // First, find any `break` or `return` without entering any inner loop,
-    // then, find `return` or labeled `break` which breaks this loop with entering inner loop,
-    // otherwise this loop is a infinite loop.
-    let mut direct_visitor = LoopVisitor {
+    let mut loop_visitor = LoopVisitor {
         cx,
         label,
         is_finite: false,
-        enter_nested_loop: false,
+        loop_depth: 0,
     };
-    direct_visitor.visit_block(loop_block);
+    loop_visitor.visit_block(loop_block);
 
-    let is_finite_loop = direct_visitor.is_finite || {
-        let mut inner_loop_visitor = LoopVisitor {
-            cx,
-            label,
-            is_finite: false,
-            enter_nested_loop: true,
-        };
-        inner_loop_visitor.visit_block(loop_block);
-        inner_loop_visitor.is_finite
-    };
+    let is_finite_loop = loop_visitor.is_finite;
 
     if !is_finite_loop {
         span_lint_and_then(cx, INFINITE_LOOPS, expr.span, "infinite loop detected", |diag| {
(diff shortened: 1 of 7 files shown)


matthiaskrgr:
What about loops that have todo!() / panic!() / unreachable() / assert!() in them? We have at least a test imo :slightly_smiling_face:

Manishearth:
r? @matthiaskrgr 

might not have time to get around to this soon

clippy_lints/src/loops/infinite_loops.rs
```diff
@@ -0,0 +1,99 @@
...
+                }
+            },
+            ExprKind::Ret(..) => self.found = true,
+            ExprKind::Loop(..) if !self.enter_nested_loop => (),
+            // `panic` like macro calls
+            ExprKind::Call(caller, _) => {
+                println!("caller: {:#?}", caller);
+            }
+            _ => {
+                println!("expr: {:?}, span: {:?}", ex.precedence(), ex.span);
```
matthiaskrgr, line 94: in this leftover debug stuff (might be why the tests fail)

J-ZhengLi: Ah this :rofl: , haha, I was testing the things that you suggested, and suddenly decided I want to save the work and sleep, so this happened, don't worry I'll fix it~

J-ZhengLi: Tests should be good now.

Btw, how can I optimize this? :thinking: Right now it has to run the same visitor with similar behavior twice, and might revisiting the same loop multiple times

matthiaskrgr: No ideas sorry, :sweat: 
Maybe someone in the clippy channel on zulip has an idea?

y21: Why does the visitor run twice?
Can't you look for a `break` anywhere, even in nested loops, in a single visitor "run"?

I think you could do that by storing an int in the visitor for the loop depth, increment/decrement when visiting/leaving one and then if the depth is 0 then `break` without a label means that it breaks the relevant loop

J-ZhengLi: > Why does the visitor run twice? Can't you look for a `break` anywhere, even in nested loops, in a single visitor "run"?
> 
> I think you could do that by storing an int in the visitor for the loop depth, increment/decrement when visiting/leaving one and then if the depth is 0 then `break` without a label means that it breaks the relevant loop

Well, I've thought about that, but unfortunately I don't know how to identify where the visitor is leaving a loop... I tried to find methods that are similar to `check_xxx_post` etc, but no luck :disappointed:... So I gave up

y21: >  I don't know how to identify where the visitor is leaving a loop

`check_expr_post`,  then checking `if let ExprKind::Loop = ex.kind`.
Either that, or you could do it directly within `check_expr` I guess
```rs
match ex.kind {
  // ...
  ExprKind::Loop(..) => {
    self.loop_depth += 1;
    intravisit::walk_expr(self, ex);
    self.loop_depth -= 1;
  }
}
```

J-ZhengLi: > or you could do it directly within `check_expr` I guess

Excellent! It works, thank you~

clippy_lints/src/loops/mod.rs
```diff
@@ -635,6 +636,48 @@ declare_clippy_lint! {
...
+    ///         if condition {
+    ///             break;
+    ///         }
+    ///     }
+    /// }
+    /// ```
+    #[clippy::version = "1.75.0"]
+    pub INFINITE_LOOP,
+    restriction,
+    "possibly unintended infinite loops"
```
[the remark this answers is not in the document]
matthiaskrgr: I also wondering about weird fn like
````rust
fn loopfn() -> Result<(), i32> {
    Ok(loop {})
}
````

afaik something like ` -> Result <!, i32>` is not stabilized yet.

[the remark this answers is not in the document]
J-ZhengLi: > I also wondering about weird fn like

Hmmm, I wasn't aware such cases, currently it would just prints help message suggesting add a break condition...
What would be an ideal solution here :thinking: ?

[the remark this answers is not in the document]
J-ZhengLi: Happy holiday :smile: 

As of your concern, yes it will lint, right now the visitor only looks for following occurrence in a loop block:

1. A return or a break
2. A function which never return, which was inspired be one of your early suggestion such as the one in `panic!`, `unreachable!`, also `std::process::exit`.

And if non of these could be find then... well, triggers the warning.

matthiaskrgr (approved):

matthiaskrgr:
Thanks!
@bors r+

Issue #11830, opened 2023-11-17
Override `clippy::pedantic` in `Cargo.toml`


davidhewitt:
### Description

I'm testing the new `[lints]` functionality in `Cargo.toml`, which I'm very excited to have on stable!

I've just tried the following `Cargo.toml` configuration:

```rust
[lints.clippy]
pedantic = "warn"
similar_names = "allow"
# and a number of other "allow" lints from the pedantic category
```

The idea is that I want to lint against the whole pedantic category but then whitelist specific lints from that category which I'm ok with not complying with.

However, this doesn't work, the `simliar_names` lint is still reported:

```
warning: binding's name is too similar to existing binding
   --> src/validators/union.rs:353:17
    |
353 |             let tag_repr = choice_key.repr()?.to_string();
    |                 ^^^^^^^^
    |
note: existing binding defined here
   --> src/validators/union.rs:344:17
    |
344 |         let mut tags_repr = String::with_capacity(50);
    |                 ^^^^^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#similar_names
    = note: `-W clippy::similar-names` implied by `-W clippy::pedantic`
    = help: to override `-W clippy::pedantic` add `#[allow(clippy::similar_names)]`
```

I believe this is because the order which `Cargo` has passed the flags to clippy on the command line has placed the allow lints before the warn.

```
 Running `/home/david/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/clippy-driver [...] '--allow=clippy::similar_names' '--warn=clippy::pedantic' [...]
```

One fix I can imagine is that clippy should prioritize individual lint flags over their category flags, ignoring the command line order for this case, but this is probably a (minor) breaking change, so there would need to be discussion on whether it's valid to do so.

### Version

```text
rustc 1.74.0 (79e9716c9 2023-11-13)
binary: rustc
commit-hash: 79e9716c980570bfd1f666e3b16ac583f0168962
commit-date: 2023-11-13
host: x86_64-unknown-linux-gnu
release: 1.74.0
LLVM version: 17.0.4
```

### Additional Labels

@rustbot-label +C-enhancement

y21:
[...]

Issue #11751, opened 2023-11-02
`doc-valid-idents` not respected in Cargo.toml [lints] table


rockboynton:
### Summary

See title

### Reproducer

I tried this code:

```toml
[lints.clippy]
pedantic    = "deny"
doc_markdown = { level = "deny", doc_valid_idents = ["GHz", ".."] }
```
```rust
//! GHz
```

I expected to see this happen:

A clean build

Instead, this happened:

I still see the error on the `GHz` in my doc comment

### Version

```text
❯ rustc -Vv
rustc 1.74.0-nightly (58eefc33a 2023-08-24)
binary: rustc
commit-hash: 58eefc33adf769a1abe12ad94b3e6811185b4ce5
commit-date: 2023-08-24
host: x86_64-unknown-linux-gnu
release: 1.74.0-nightly
LLVM version: 17.0.0
```

Stoned33:
Lints in Rust can be controlled and customized using the `#[allow(lint_name)]` attribute in the code. This attribute allows you to selectively disable specific lints for specific sections of code. While this attribute is typically used within the code itself, it may be worth exploring if it can be applied to the Cargo.toml file as well.

rockboynton:
Hi @Stoned33, thanks for your reply. While that does work per crate, other lints work with overriding them in the Cargo.toml as in my example, although I do notice it only seems to work when it's only the bare level key.

For example, this works as expected, to allow denying the pedantic group with an exception, since the order matters in the file

```toml
[lints.clippy]
pedantic    = "deny"
borrow_as_ptr = "allow" 
```

samueltardieu:
Isn't lint extra configuration supposed to be given in `clippy.toml`? If you create one with

```toml
doc-valid-idents = ["GHz", ".."]
```

it should do what you want.

Alexendoo:
Opened a cargo issue for this: https://github.com/rust-lang/cargo/issues/12917

Issue #11237, opened 2023-07-27
`cargo clippy` not obeying `[lints.clippy]` from `Cargo.toml`, but only for some lints


Kmeakin:
### Summary

Clippy does not obey the `lint_name = "allow"` key under `[lints.clippy]` from `Cargo.toml` for certain lints, but does work as expected for others. The lints have to be silenced with `#[allow(...)]` instead.

In the example below, `must_use_candidate` obeys the `[lints.clippy]` settings, but `wildcard_imports` doesn't.

Other lints that don't obey `[lints.clippy]` that I have encountered:
* `too_many_lines`
* `option_if_let_else`

### Reproducer

`Cargo.toml`:
```toml
[package]
name = "clippy-test"
version = "0.1.0"
edition = "2021"

[lints.clippy]
pedantic = "warn"

must_use_candidate = "allow"
wildcard_imports = "allow"
```

`.cargo/config.toml`:
```toml
[unstable]
lints = true

```

`src/lib.rs`:
```rust
mod foo {
    pub struct A;
}

// "usage of wildcard import" reported
use foo::*;

// "this function could have a `#[must_use]` attribute" not reported
pub fn a() -> A {
    A
}
```

### Version

```text
rustc 1.73.0-nightly (864bdf784 2023-07-25)
binary: rustc
commit-hash: 864bdf7843e1ceabc824ed86d97006acad6af643
commit-date: 2023-07-25
host: x86_64-unknown-linux-gnu
release: 1.73.0-nightly
LLVM version: 16.0.5
```
```


### Additional Labels

_No response_

**Comment by Alexendoo** [comment]
Looks like a cargo issue, it sorts the flags before passing them to clippy but the order matters

```
'--allow=clippy::wildcard_imports' '--warn=clippy::pedantic' '--allow=clippy::must_use_candidate'
```

Alexendoo:
Closing in favour of https://github.com/rust-lang/cargo/issues/12918

kristof-mattei:
As posted here https://github.com/rust-lang/cargo/issues/12918#issuecomment-1818071667

The behavior here seems different than rustc. rustc seems to consolidate the lints, apply groups and then the individual settings, whereas Clippy seems to go by the order of the params (and thus the priority).

Wouldn't it be better to match rustc's behavior?

Alexendoo:
[...]

Issue #10306, opened 2023-02-07
Workspace dependencies


stanislav-tkach:
### What it does

I'm not sure if it is a Clippy responsibility, but I think it would be nice to suggest to use "[workspace dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace)" if some library is used in multiple workspace crates.

Perhaps the opposite changes can be suggested if a dependency is used only once or less than some threshold value, though this logic should probably be a separate lint. Ideally both values should be configurable.

### Lint Name

workspace_dependencies

### Category

pedantic

### Advantage

- It is much more convenient to change the dependency version in one place.

### Drawbacks

The only disadvantage I can see is that such lint can be noisy for small workspaces, so it probably should be disabled by default.

### Example

I don't think this example is needed, but nevertheless: 
```toml
# [PROJECT_DIR]/Cargo.toml
[workspace]
members = ["foo", "bar"]

# [PROJECT_DIR]/foo/Cargo.toml
[package]
name = "foo"

[dependencies]
rand = "0.8.5"

# [PROJECT_DIR]/bar/Cargo.toml
[package]
name = "bar"

[dependencies]
rand = "0.8.5"
```

Could be written as:

```toml
# [PROJECT_DIR]/Cargo.toml
[workspace]
members = ["foo", "bar"]

[workspace.dependencies]
rand = "0.8.5"

# [PROJECT_DIR]/foo/Cargo.toml
[package]
name = "foo"

[dependencies]
rand.workspace = true

# [PROJECT_DIR]/bar/Cargo.toml
[package]
name = "bar"

[dependencies]
rand.workspace = true
```

epage:
There are several potential strategies for workspace dependenices
- Put **everything** (except maybe path dependencies) in them allowing a central place to see everything
  - path dependencies are dependent on whether you want your published version requirements to all point to the same version or nit
- Put "this version must be shared" dependencies in them, this is context specific
- Put nothing in them

For me, putting only duplicated dependencies in as workspace dependencies just seems like a source of churn as that is dependent on the current implementation of your code base and can change.

stanislav-tkach:
While I agree with most of arguments, I'm still not sure about this part:

> For me, putting only duplicated dependencies in as workspace dependencies just seems like a source of churn as that is dependent on the current implementation of your code base and can change.

[...]

PR #11832, opened 2023-11-17
Add `lint_groups_priority` lint

Alexendoo:
Warns when a lint group in Cargo.toml's `[lints]` section shares the same priority as a lint. This is in the cargo section but is categorised as `correctness` so it's on by default, it doesn't call `cargo metadata` though and parses the `Cargo.toml` directly

The lint should be temporary until https://github.com/rust-lang/cargo/issues/12918 is resolved, but in the meanwhile this is an common issue to run into

- #11237
- #11751
- #11830

changelog: Add [`lint_groups_priority`] lint

r? @flip1995


flip1995:
Ugh, I really dislike re-parsing the `Cargo.toml` file. Especially something as complex as the lint table. This should be extended in the future AFAIK. Keeping up with this might be hard. How much we have to keep up with it depends on how temporary this temporary lint is.

Is there the possibility to forward this through `cargo-metadata` somehow? But then again, calling `cargo-metadata` on a enabled-by-default lint can have a huge perf impact, making it not really an option... But then again, having this lint allow-by-default would be pretty useless.[^1]

The problem with the "temporary" part that I see is: `cargo` doesn't have the knowledge of what is and isn't a lint group. So moving this lint/warning to `cargo` seems unrealistic in the near future.

I acknowledge that this is an actual problem with the lint table and that we have to do something about this though.

@epage Asking for your guidance here: Do you have better ideas on how to address this issue in the short-term? 

[^1]: I think there were plans to put the `Cargo.toml` parsing code in a crate and publish it on crates.io. But I also don't think this will happen any time soon(?)

Alexendoo:
Yeah redefining the lint table isn't ideal, it's defined slightly lax by having `[lints.rust]` and `[lints.clippy]` hardcoded, so the addition of other keys that don't correspond to tools in `[lints]` wouldn't break the lint

 It does give nicer diagnostics than the cargo lints that read from `cargo metadata` though

flip1995:
@epage I saw that linked in the PR description. My question was more about your expectations on when this will be addressed in `cargo`. Re-parsing parts of the `Cargo.toml` file in Clippy is really not ideal. So for me to agree to merge this "temporarily" requires one of two things:
1. Some approximate timeline on when `cargo` wants to address this issue; or
2. Share the `Cargo.toml` parsing code between Clippy and `cargo`

The second one would be a long term solution and we could keep this lint permanently in Clippy. But achieving that also seems unrealistic in the short term.

epage:
For 1, that is too large of an effort for me to slip in among the other work I'm trying to finish up.

For 2, making the schema available as a package is part of that work I'm trying to finish up so crates.io can use it. It won't have any of the other logic in it like filling in of inferred fields or inheriting from workspaces.

flip1995:
> For 2, making the schema available as a package is part of that work I'm trying to finish up so crates.io can use it.

That's great to hear! In that case, IMO we can merge this now and rework it with the package later. (Maybe together with the other cargo lints we have)

@epage Last request: Can I ask you or someone else from the `cargo` team to review the toml-parsing code in this PR? I don't want to miss something important there.

clippy_lints/src/cargo/mod.rs
```diff
@@ -162,6 +163,43 @@ declare_clippy_lint! {
...
+    /// ### What it does
+    /// Checks for lint groups with the same priority as lints in the `Cargo.toml`
+    /// [`[lints]` table](https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section).
+    ///
+    /// This lint will be removed once [cargo#12918](https://github.com/rust-lang/cargo/issues/12918)
+    /// is resolved.
+    ///
+    /// ### Why is this bad?
+    /// The order of lints in the `[lints]` is ignored, to have a lint override a group the
+    /// `priority` field needs to be used, otherwise the names are sorted alphabetically.
```
epage, line 176: The sort order is undefined

epage (commented):
Ill have to check `check_table` later. As for the serde types, they are more loosely defined which is fine.

Note that this doesn't help with rustdoc.

epage (approved):
Sorry, lost track of finishing my look at this

flip1995:
I saw Ed point out the `cargo_utils_schema` crate here: https://github.com/rust-lang/rust-clippy/issues/10306#issuecomment-1912670329

Can the lint implementation be updated with this, instead of implementing the parsing logic yourself?

https://docs.rs/cargo-util-schemas/latest/cargo_util_schemas/manifest/struct.TomlManifest.html

Alexendoo:
I don't think we would be able to get span information from `cargo_utils_schema`, so we could do it using that but we wouldn't be able to produce nice error messages or have an edit suggestion

epage:
The straightforward way of getting spans for a `toml` file is to define serde types with fields wrapped by `serde_spanned`.  This is invasive enough that I'm trying to avoid it for `cargo-util-schema`. 

In cargo, we will be experimenting with a less straightforward approach.  You can define a serde deserializer that accepts a path to a field and it will error when it gets to that path.  You can then get the span from the error.

clippy_lints/src/cargo/lint_groups_priority.rs
```diff
@@ -0,0 +1,171 @@
...
+                },
+                Some(("clippy", group)) => {
+                    clippy_groups.insert(group);
+                },
+                _ => {},
+            }
+        }
+
+        // internally warnings is considered a lint rather than a group
+        rustc_groups.insert("warnings");
```
flip1995, line 166: Does this matter? I think that you can pass in `-Dwarnings` anywhere and it won't change the lint level of any other lint you set before/after, as `warnings` is not actually a group but some special handled keyword/lint.

Alexendoo: Yeah looks like it doesn't matter, removed it 👍

flip1995:
@bors r+ 

Thanks! Also a big thanks to Ed for helping with the review!

PR #11839, opened 2023-11-18
Extend `unused_io_amount` with more usecases

### LINKED ISSUE #11713: `unused_io_amount` does not fire on `match file.write() { Ok(_) }`
### Summary

Clippy treats `Ok(_)` as a use of the number of bytes written.

### Lint Name

unused_io_amount

### Reproducer

I tried this code:

```rust
use std::io::Write;

fn main() {
    let mut file = std::fs::File::open("/foo").unwrap();
    match file.write(b"very large string") {
        Ok(_) => println!("yay it works"), // oops
        Err(err) => eprintln!("error {err}"),
    }
}
```

I expected to see this happen: Clippy warns that `write` is only called once and the byte size is discarded.

Instead, this happened: No warning.

### Version

```text
clippy 0.1.75 (cd674d61 2023-10-24)
rustc 1.75.0-nightly (cd674d617 2023-10-24)
binary: rustc
commit-hash: cd674d61790607dfb6faa9d754bd3adfa13aea7c
commit-date: 2023-10-24
host: aarch64-apple-darwin
release: 1.75.0-nightly
LLVM version: 17.0.3
```

PartiallyTyped:
Fixes #11713

Adds some more examples for `unused_io_amount` and some documentation.

changelog: Extended [`unused_io_amount`] cases

### DIFF
--- a/clippy_lints/src/unused_io_amount.rs
+++ b/clippy_lints/src/unused_io_amount.rs
@@ -1,9 +1,10 @@
-use clippy_utils::diagnostics::{span_lint, span_lint_and_help};
-use clippy_utils::{is_trait_method, is_try, match_trait_method, paths};
+use clippy_utils::diagnostics::span_lint_and_then;
+use clippy_utils::{is_res_lang_ctor, is_trait_method, match_trait_method, paths};
+use hir::{ExprKind, PatKind};
 use rustc_hir as hir;
 use rustc_lint::{LateContext, LateLintPass};
 use rustc_session::declare_lint_pass;
-use rustc_span::sym;
+use rustc_span::{sym, Span};
 
 declare_clippy_lint! {
     /// ### What it does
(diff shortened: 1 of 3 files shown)


clippy_lints/src/unused_io_amount.rs
```diff
@@ -45,126 +46,241 @@ declare_clippy_lint! {
...
+        },
+        _ => {},
+    };
+}
+
+fn should_lint<'a>(cx: &LateContext<'a>, mut inner: &'a hir::Expr<'a>) -> Option<IoOp> {
+    inner = unpack_match(inner);
+    inner = unpack_try(inner);
+    inner = unpack_call_chain(inner);
+    inner = unpack_await(inner);
```
PartiallyTyped, line 145: There's probably an easier / cleaner way to do this.

PartiallyTyped, line 145: There's probably an easier or cleaner way to do this.

clippy_lints/src/unused_io_amount.rs
```diff
@@ -45,126 +46,241 @@ declare_clippy_lint! {
...
+            PatKind::TupleStruct(
+                _,
+                [hir::Pat {
+                    kind: PatKind::Wild,
+                    ..
+                }],
+                _
+            )
+        )
+}
```
PartiallyTyped, line 174: Should these 2 move to `clippy_utils` ?

PartiallyTyped (commented):
This achieves the goal, but I think it can be done in a better way.

clippy_lints/src/unused_io_amount.rs
PartiallyTyped: As Alejandra pointed out on Zulip, there's also the case of `Ok(_@bind)`.

y21: I haven't yet fully looked through the code so may be missing context, but if the intention here is to check if the pattern ignores the bytes, then this shouldn't return true for `Ok(bind @ _)`. It's basically the same as `Ok(bind)`, which should be considered read. I think this is handled correctly right now though. Wouldn't hurt to have a test for that in any case

PartiallyTyped: It shouldn't because it is matching the wild card. If for some reason the user has a bind there, then either they get a lint about unused var - which is why I didn't add a check here for that case - or they handle it, so it should be cool. 

I will add the test case it tomorrow.

y21: > It shouldn't because it is matching the wild card

Is this in response to "shouldn't return true for ...", as in, you're agreeing with my reasoning and this shouldn't return true, or is that in response to "should be considered read", as in, you disagree and it shouldn't consider this as being read? Wasn't able to read it out of that sentence😅 

To make my point clearer, it shouldn't lint because:
```rs
match read() {
  Ok(bytes @ _) => println!("Read {bytes}"),
  _ => todo!()
}
```
Admittedly, it's weird code to write in the first place, but the bytes are being handled.

As for if this should be a util function, I'm not sure if it's worth. Is that useful in other lints? It seems rather niche to me to want to check if a pattern is `Ok(_)` specifically.

PartiallyTyped: Yes, I am agreeing with your reasoning :D 

As for the suggestion to be a util; there's an `is_ok` function that works only for match arms and is part of another a different function so perhaps there's use for one of the two.

PR #11844, opened 2023-11-20
Remove underscore check for `manual_non_exhaustive` lint

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust-clippy/issues/10550.

As indicated in https://github.com/rust-lang/rust-clippy/pull/10559, the underscore check should be removed.

changelog: remove underscore check for `manual_non_exhaustive` lint

r? @blyxyas

### DIFF
--- a/clippy_lints/src/manual_non_exhaustive.rs
+++ b/clippy_lints/src/manual_non_exhaustive.rs
@@ -118,7 +118,6 @@ impl EarlyLintPass for ManualNonExhaustiveStruct {
             if let Some(Ok(field)) = iter.next()
                 && iter.next().is_none()
                 && field.ty.kind.is_unit()
-                && field.ident.map_or(true, |name| name.as_str().starts_with('_'))
             {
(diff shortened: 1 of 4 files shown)


blyxyas (commented):
Shouldn't this check also be removed from line 121? Or am I missing something?

GuillaumeGomez:
@blyxyas told me they're busy IRL for the next weeks so I'll assign someone else to prevent taking time from them (thanks for all your reviews!).

I applied the suggestion, thanks for that!

r? @flip1995

flip1995:
I looked for the motivation behind checking for `_` prefixes, but couldn't find any comments about this when this lint was implemented.  

I also ran lintcheck for this on the top 400 crates. It triggered in 2 more cases. Both were legit:

```
proc-macro-error-1.0.4/src/diagnostic.rs:13:1
rustls-0.22.0-alpha.4/src/versions.rs:11:1
```

@bors r+

Thanks!

PR #11850, opened 2023-11-21
[`deprecated_semver`]: Allow `#[deprecated(since = "TBD")]`

Nilstrieb:
"TBD" is allowed by rustdoc, saying that it will be deprecated in a future version. rustc will also not actually warn on it.
I found this while checking the rust-lang/rust with clippy.

changelog: [`deprecated_semver`]: allow using `since = "TBD"`


dswij (approved):
Thank you!

Issue #11814, opened 2023-11-15
expending lint [`blocks_in_if_conditions`] to check not only `if`'s conditions, but also `match`'s as welll


J-ZhengLi:
### What it does

Currently only the blocks in if conditions are warned, however, if readability was a concern then those blocks after  `match` keyword should be treated the same right?

[playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9b7b4a803ef2789325ec59e5796353a9)

### Advantage

- Readability
- Consistancy with [`blocks_in_if_conditions`]

### Drawbacks

~Could breaks auto-fix capability.~

~Check the second example, which a block was used in a `while` condition, however, if we move the block up and bind it as suggested, the code will not work, as x does not changes anymore causing an infinite loop.~

~But maybe we can skip such senario.~

### Example

```rust
fn match_block() {
    let _x: i32 = match {
        let opt: Option<u8> = Some(1);
        opt
    } {
        Some(num) => num.into(),
        None => -1
    };
}
```

Could be written as:

```rust
fn match_block() {
    let opt: Option<u8> = Some(1);
    let _x: i32 = match opt {
        Some(num) => num.into(),
        None => -1
    };
}
```

Jarcho:
The change on `while` is not the same since it needs to be evaluated every iteration.

J-ZhengLi:
> The change on `while` is not the same since it needs to be evaluated every iteration.

Good point! I've removed `while` from description

PR #11853, opened 2023-11-22
expending lint [`blocks_in_if_conditions`] to check match expr as well

J-ZhengLi:
closes: #11814 

changelog: expending lint [`blocks_in_if_conditions`] to check block as match expression


clippy_lints/src/blocks_in_if_conditions.rs
```diff
@@ -52,88 +52,89 @@ impl<'tcx> LateLintPass<'tcx> for BlocksInIfConditions {
...
-                        if span.from_expansion() || expr.span.from_expansion() {
+        let Some((cond, keyword)) = higher::If::hir(expr).map(|hif| (hif.cond, "if")).or_else(|| {
+            if let ExprKind::Match(match_ex, _, MatchSource::Normal) = expr.kind {
+                Some((match_ex, "match"))
+            } else {
+                None
+            }
+        }) else {
+            return;
+        };
```
J-ZhengLi, line 63: I've highlighted the part where I changed hoping it'll make things clear,

the rest are just auto formatted, seems like GH is having trouble render it properly

clippy_lints/src/blocks_in_if_conditions.rs
```diff
@@ -52,88 +52,89 @@ impl<'tcx> LateLintPass<'tcx> for BlocksInIfConditions {
...
-                        if span.from_expansion() || expr.span.from_expansion() {
+        let Some((cond, keyword)) = higher::If::hir(expr).map(|hif| (hif.cond, "if")).or(
+            if let ExprKind::Match(match_ex, _, MatchSource::Normal) = expr.kind {
+                Some((match_ex, "match"))
+            } else {
+                None
+            },
+        ) else {
+            return;
+        };
```
J-ZhengLi, line 63: Hmmm, the rendering is messed up by auto formatting, 
but here are what's changed, hoping it'll make things a bit clearer

clippy_lints/src/blocks_in_if_conditions.rs
```diff
@@ -52,88 +52,89 @@ impl<'tcx> LateLintPass<'tcx> for BlocksInIfConditions {
         if in_external_macro(cx.sess(), expr.span) {
             return;
         }
-        if let Some(higher::If { cond, .. }) = higher::If::hir(expr) {
```
J-ZhengLi, line 55: *deleted*

clippy_lints/src/blocks_in_if_conditions.rs
```diff
@@ -52,88 +52,89 @@ impl<'tcx> LateLintPass<'tcx> for BlocksInIfConditions {
...
+                    // move block higher
+                    let mut applicability = Applicability::MachineApplicable;
+                    span_lint_and_sugg(
+                        cx,
+                        BLOCKS_IN_IF_CONDITIONS,
+                        expr.span.with_hi(cond.span.hi()),
+                        COMPLEX_BLOCK_MESSAGE,
+                        "try",
+                        format!(
+                            "let res = {}; {keyword} res",
```
J-ZhengLi, line 107: if -> {keyword}

clippy_lints/src/blocks_in_if_conditions.rs
```diff
@@ -52,88 +52,89 @@ impl<'tcx> LateLintPass<'tcx> for BlocksInIfConditions {
...
                                 snippet_block_with_applicability(
                                     cx,
-                                    block.span,
+                                    ex.span,
                                     "..",
                                     Some(expr.span),
                                     &mut applicability
-                                ),
+                                )
                             ),
```
llogiq, line 89: This `format!("{}", _)` can be reduced to `_.to_string()`.

tests/ui/blocks_in_if_conditions.stderr
```diff
@@ -32,5 +32,22 @@ LL |     if true && x == 3 { 6 } else { 10 }
    = note: `-D clippy::nonminimal-bool` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::nonminimal_bool)]`
 
-error: aborting due to 3 previous errors
+error: in an `if` condition, avoid complex blocks or closures with blocks; instead, move the block or closure higher and bind it with a `let`
```
llogiq, line 35: We should also change the error message for `match`, as there is no `if` condition. Instead let's call it `match` scrutinee.

tests/ui/blocks_in_if_conditions.rs
```diff
@@ -61,4 +61,16 @@ fn block_in_assert() {
     );
 }
 
+// issue #11814
+fn block_in_match_expr(num: i32) -> i32 {
+    match {
```
llogiq, line 66: I'd like to see a test with an unsafe block, too (just so be sure it's handled correctly).

llogiq (commented):
This looks mostly good, but the error message for matches needs some work. Also we may want to rename the lint to `block_in_condition` or some such, because it now isn't restricted to if conditions.

J-ZhengLi:
> ... we may want to rename the lint to `block_in_condition` or some such

I agree, I'll use that name as I couldn't think of a better name~

> just find it interesting that this lint has been renamed from `block_in_if_condition_expr` to `block_in_if_condition`, and now it's becoming `block_in_condition` :rofl: , maybe one day it'll becomes just "block"

llogiq:
Thank you, this looks good to merge now.

@bors r+

Issue #7125, opened 2021-04-22
Warn when iterating an array of ranges


teor2345:
### What it does

Warns users if they accidentally iterate range struct items, rather than iterating through the indexes in the range:

Suggests:
- iterating through the indexes in the range, or
- disambiguating the range struct iteration by providing a range type.

This lint will become more important once `rustc` allows iterating arrays by value, because that allows code like:
```rust
for _ in [0..5] {} // iterates through 1 range, not 5 indexes
```

For context, see https://github.com/rust-lang/rust/pull/84147#discussion_r612045984

### Categories (optional)

- Kind: `clippy::correctness`

*What is the advantage of the recommended code over the original code*

The recommended code either:
- does what the user actually intended, or
- makes it clear that the user intended something really unusual (iterating ranges themselves, not their indexes).

### Drawbacks

Expressing iterating through an array of ranges becomes more verbose. But this disambiguation makes the intent of the code much clearer.

### Example

```rust
for _ in [0..5] {} // iterates through 1 range, clippy warning
```

Is probably intended to be:

```rust
for _ in 0..5 {} // iterates through 5 indexes
```

And if range iteration is genuinely intended, it could be disambiguated as:

```rust
for _: Range<_> in [0..5] {} // iterates through 1 range, no clippy warning
```

scottmcm:
nit: you can't actually annotate a type for the pattern in a `for` loop -- it's not supported in the grammar.

matthiaskrgr:
i guess this should only lint on arrays of len 1?

what about code like this:
```
pub fn main() {
    for r in [0..7, 8..5, 37..42] {
        println!("{:?}",r);
    }
}
```

llogiq:
Shouldn't that one be caught by `single_element_loop`?

llogiq:
Looking into that, it catches `for r in &[0..5]`, but not (yet) `for r in [0..5].iter()`.

llogiq:
I'll extend `single_element_loop` to also catch the `.iter()` variant.

camsteffen:
What if we just lint `for _ in [any_single_element]` since that is never really useful.

llogiq:
Yes, but that appears to only work if the element is an `ExprPath` for some reason.

pitaj:
[...]

PR #11862, opened 2023-11-23
suggest alternatives to iterate an array of ranges

christophbeberweil:
works towards #7125
changelog: [`single_element_loop`]: suggest better syntax when iterating over an array of a single range

@thinkerdreamer and myself worked on this issue during a workshop by @llogiq at the RustLab 2023 conference. It is our first contribution to clippy. 

When iterating over an array of only one element, _which is a range_, our change suggests to replace the array with the contained range itself. Additionally, a hint is printed stating that the user probably intended to iterate over the range and not the array. If the single element in the array is not a range, the previous suggestion in the form of `let {pat_snip} = {prefix}{arg_snip};{block_str}`is used.

This change lints the array with the single range directly, so any prefixes or suffixes are covered as well.

### DIFF
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -95,7 +95,7 @@ pub(super) fn check<'tcx>(
                 cx,
                 SINGLE_ELEMENT_LOOP,
                 arg.span,
-                format!("This loops only once with {pat_snip} being {range_expr}").as_str(),
+                format!("this loops only once with {pat_snip} being {range_expr}").as_str(),
                 "did you mean to iterate over the range instead?",
                 sugg.to_string(),
                 Applicability::Unspecified,
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -95,7 +95,7 @@ pub(super) fn check<'tcx>(
                 cx,
                 SINGLE_ELEMENT_LOOP,
                 arg.span,
-                format!("this loops only once with {pat_snip} being {range_expr}").as_str(),
+                format!("this loops only once with `{pat_snip}` being `{range_expr}`").as_str(),
                 "did you mean to iterate over the range instead?",
                 sugg.to_string(),
                 Applicability::Unspecified,
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -96,7 +96,7 @@ pub(super) fn check<'tcx>(
                 "for loop over a single range inside an array, rather than iterating over the elements in the range directly",
                 "did you mean to iterate over the range instead?",
                 sugg.to_string(),
-                applicability,
+                Applicability::Unspecified,
             );
         } else {
             span_lint_and_sugg(
--- a/clippy_lints/src/loops/single_element_loop.rs
+++ b/clippy_lints/src/loops/single_element_loop.rs
@@ -1,6 +1,6 @@
 use super::SINGLE_ELEMENT_LOOP;
 use clippy_utils::diagnostics::span_lint_and_sugg;
-use clippy_utils::source::{indent_of, snippet_with_applicability};
+use clippy_utils::source::{indent_of, snippet, snippet_with_applicability};
 use clippy_utils::visitors::contains_break_or_continue;
 use rustc_ast::util::parser::PREC_PREFIX;
 use rustc_ast::Mutability;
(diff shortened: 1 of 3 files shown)


pitaj:
Generally I think suggestions should not be machine applicable if they change the semantics of the program, so you might want to change those so they aren't applied automatically.

clippy_lints/src/loops/single_element_loop.rs
```diff
@@ -87,14 +87,27 @@ pub(super) fn check<'tcx>(
...
+        if clippy_utils::higher::Range::hir(arg_expression).is_some() {
+            let sugg = snippet(cx, arg_expression.span, "..");
+            span_lint_and_sugg(
+                cx,
+                SINGLE_ELEMENT_LOOP,
+                arg.span,
+                "for loop over a single range inside an array, rather than iterating over the elements in the range directly",
+                "did you mean to iterate over the range instead?",
+                sugg.to_string(),
+                applicability,
```
llogiq, line 99:
```suggestion
                Applicability::Unspecified,
```

clippy_lints/src/loops/single_element_loop.rs
```diff
@@ -87,14 +87,27 @@ pub(super) fn check<'tcx>(
...
-            format!("{{\n{indent}let {pat_snip} = {prefix}{arg_snip};{block_str}}}"),
-            applicability,
-        );
+        if clippy_utils::higher::Range::hir(arg_expression).is_some() {
+            let sugg = snippet(cx, arg_expression.span, "..");
+            span_lint_and_sugg(
+                cx,
+                SINGLE_ELEMENT_LOOP,
+                arg.span,
+                "for loop over a single range inside an array, rather than iterating over the elements in the range directly",
```
llogiq, line 96: Perhaps we can find a shorter way to phrase this? I like the clarity though. Maybe something like "This loops only once with <loop pattern> being <range expr>", which is admittedly harder to construct.

llogiq (commented):
This looks good for a start! :+1: I do concur with @pitaj's suggestion about applicability, and I have an idea on improving the error message, and we may want to add other tests where e.g. the range or even the whole array is in a variable, but otherwise this is mostly merge-worthy already.

clippy_lints/src/loops/single_element_loop.rs
christophbeberweil: Done

clippy_lints/src/loops/single_element_loop.rs
christophbeberweil: We discussed a lot about the message of this lint, but opted for a longer and more verbose one. I adjusted it to contain your version, which is shorter, but now the text "0..5" is written twice in the lint message, which feels redundant to me.

ThinkerDreamer: I see it as "item" (the looping variable name) and "0..5", rather than 0..5 twice. Am I missing something?

christophbeberweil: This is the current output of the lint:
```
error: this loops only once with item being 0..5
  --> $DIR/single_element_loop.rs:24:17
   |
LL |     for item in [0..5] {
   |                 ^^^^^^ help: did you mean to iterate over the range instead?: `0..5`
```

"0-5" is mentioned in the lint error message and in the suggestion. I thing mentioning it once should be enough, but mentioning it twice will probably not hurt.

christophbeberweil: ... and with a lowercase "t" the error message actually passes the tests, which were failing with my commit from earlier today. :)

clippy_lints/src/loops/single_element_loop.rs
```diff
@@ -87,14 +87,29 @@ pub(super) fn check<'tcx>(
...
-        );
+        if clippy_utils::higher::Range::hir(arg_expression).is_some() {
+            let range_expr = snippet(cx, arg_expression.span, "?").to_string();
+
+            let sugg = snippet(cx, arg_expression.span, "..");
+            span_lint_and_sugg(
+                cx,
+                SINGLE_ELEMENT_LOOP,
+                arg.span,
+                format!("this loops only once with {pat_snip} being {range_expr}").as_str(),
```
llogiq, line 98:
```suggestion
                format!("this loops only once with `{pat_snip}` being `{range_expr}`").as_str(),
```

llogiq (commented):
Ah, there's only one thing I missed, the code snippets should be in backticks in the text.

clippy_lints/src/loops/single_element_loop.rs
christophbeberweil: Done :)

llogiq:
Thank you!

@bors r+

llogiq:
Congratulations, @christophbeberweil and @ThinkerDreamer to your first successful clippy PR! 🎉

If any of you like to contribute further to clippy and need help to either pick or solve another issue, either ping me anywhere or go to the clippy channel on the rust-zulip.

ThinkerDreamer:
Thanks @llogiq! And thanks @christophbeberweil for the great pairing! 😊

christophbeberweil:
Thanks @llogiq and @ThinkerDreamer, I enjoyed hacking on clippy with both of you 🥳

PR #11864, opened 2023-11-24
Create new lint `option_map_or_err_ok`

### LINKED ISSUE #10045: Suggest `ok_or` over `map_or` when possible
### What it does

Check for usage of `_.map_or(Err(_), Ok)`.

### Lint Name

`option_map_or_err_ok`

### Category

pedantic

### Advantage

- Readability: this can be written more concisely as `_.ok_or(_)`.

### Drawbacks

_No response_

### Example

```rust
opt.map_or(Err(e), Ok);
```

Could be written as:

```rust
opt.ok_or(e);
```

GuillaumeGomez:
Fixes #10045.

For the following code:

```rust
let opt = Some(1);
opt.map_or(Err("error"), Ok);
```

It suggests to instead write:

```rust
let opt = Some(1);
opt.ok_or("error");
```

r? @flip1995 

changelog: Create new lint `option_map_or_err_ok`

### DIFF
--- a/clippy_lints/src/methods/mod.rs
+++ b/clippy_lints/src/methods/mod.rs
@@ -70,6 +70,7 @@ mod obfuscated_if_else;
 mod ok_expect;
 mod open_options;
 mod option_as_ref_deref;
+mod option_map_or_err_ok;
 mod option_map_or_none;
 mod option_map_unwrap_or;
 mod or_fun_call;
@@ -3726,6 +3727,31 @@ declare_clippy_lint! {
     "calls to `Path::join` which will overwrite the original path"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `_.map_or(Err(_), Ok)`.
+    ///
+    /// ### Why is this bad?
+    /// Readability, this can be written more concisely as
+    /// `_.ok_or(_)`.
+    ///
+    /// ### Example
+    /// ```no_run
+    /// # let opt = Some(1);
+    /// opt.map_or(Err("error"), Ok);
+    /// ```
+    ///
+    /// Use instead:
+    /// ```no_run
+    /// # let opt = Some(1);
+    /// opt.ok_or("error");
+    /// ```
+    #[clippy::version = "1.76.0"]
+    pub OPTION_MAP_OR_ERR_OK,
+    style,
+    "using `Option.map_or(Err(_), Ok)`, which is more succinctly expressed as `Option.ok_or(_)`"
+}
+
 pub struct Methods {
     avoid_breaking_exported_api: bool,
     msrv: Msrv,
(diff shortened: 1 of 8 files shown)


clippy_lints/src/methods/mod.rs
```diff
@@ -3726,6 +3727,31 @@ declare_clippy_lint! {
     "calls to `Path::join` which will overwrite the original path"
 }
 
+declare_clippy_lint! {
+    /// ### What it does
+    /// Checks for usage of `_.map_or(Err(), Ok)`.
+    ///
+    /// ### Why is this bad?
+    /// Readability, this can be written more concisely as
+    /// `_.ok()`.
```
flip1995, line 3736:
```suggestion
    /// Checks for usage of `_.map_or(Err(_), Ok)`.
    ///
    /// ### Why is this bad?
    /// Readability, this can be written more concisely as
    /// `_.ok_or(_)`.
```

clippy_lints/src/methods/mod.rs
```diff
@@ -3726,6 +3727,31 @@ declare_clippy_lint! {
...
+    ///
+    /// Use instead:
+    /// ```no_run
+    /// # let opt = Some(1);
+    /// opt.ok_or("error");
+    /// ```
+    #[clippy::version = "1.76.0"]
+    pub OPTION_MAP_OR_ERR_OK,
+    style,
+    "using `Option.map_or(Err(), Ok)`, which is more succinctly expressed as `Option.ok()`"
```
flip1995, line 3752:
```suggestion
    "using `Option.map_or(Err(_), Ok)`, which is more succinctly expressed as `Option.ok_or(_)`"
```

clippy_lints/src/methods/option_map_or_err_ok.rs
```diff
@@ -0,0 +1,45 @@
...
+) {
+    let is_option = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(recv), sym::Option);
+
+    if !is_option {
+        return;
+    }
+
+    let map_arg_is_ok = is_res_lang_ctor(cx, path_res(cx, map_expr), ResultOk);
+
+    if map_arg_is_ok
```
flip1995, line 28: I think those could all be pulled into the same `if` condition, no need to define variables.

I guess this would also apply to your previous PR about another of those lints, but no need to go back to that in this PR.

GuillaumeGomez: Ok, I'll send a follow-up. I took this approach because I thought it made code easier to read. I'll just add comments instead.

clippy_lints/src/methods/mod.rs
GuillaumeGomez: :+1:

clippy_lints/src/methods/mod.rs
GuillaumeGomez: :+1:

GuillaumeGomez:
Applied suggestions and simplified code. Like I said, I'll send a follow-up for my previous lint as well.

flip1995:
@bors r+

Thanks!

PR #11865, opened 2023-11-25
feat: add `manual_is_variant_and` lint

yuxqiu:
changelog: add a new lint [`manual_is_variant_and`].
- Replace `option.map(f).unwrap_or_default()` and `result.map(f).unwrap_or_default()` with `option.is_some_and(f)` and `result.is_ok_and(f)` where `f` is a function or closure that returns `bool`.
- MSRV is set to 1.70.0 for this lint; when `is_some_and` and `is_ok_and` was stabilised

---

For example, for the following code:

```rust
let opt = Some(0);
opt.map(|x| x > 1).unwrap_or_default();
```

It suggests to instead write:

```rust
let opt = Some(0);
opt.is_some_and(|x| x > 1)
```


clippy_lints/src/methods/map_unwrap_or_default.rs
```diff
@@ -0,0 +1,73 @@
...
+    let is_result = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(map_recv), sym::Result);
+    if !is_option && !is_result {
+        return;
+    }
+
+    // 3. the `map_arg` function/closure doesn't return `bool`
+    let map_arg_return_bool = expr_sig(cx, map_arg).is_some_and(|fn_sig| {
+        fn_sig
+            .output()
+            .is_some_and(|ret| ret.no_bound_vars().is_some_and(Ty::is_bool))
```
Jarcho, line 38: `instantiate_identity` should be used instead of `no_bound_vars`. This won't be needed with the previous change, but this is still useful to know.

clippy_lints/src/methods/map_unwrap_or_default.rs
```diff
@@ -0,0 +1,73 @@
...
+    });
+    if !map_arg_return_bool {
+        return;
+    }
+
+    // 4. msrv doesn't meet either `OPTION_IS_SOME_AND` or `RESULT_IS_OK_AND`
+    if !msrv.meets(OPTION_IS_SOME_AND) || !msrv.meets(RESULT_IS_OK_AND) {
+        return;
+    }
```
Jarcho, line 47: This doesn't matter since they have the same version, but this is logically the wrong condition. A correct version would first check which function is being suggested and only gate on the function's msrv.

With the previous suggestion combining these it doesn't matter at all.

clippy_lints/src/methods/map_unwrap_or_default.rs
```diff
@@ -0,0 +1,73 @@
...
+        |diag| {
+            let suggestion = vec![
+                (map_span, suggest.into()),
+                (expr.span.with_lo(unwrap_recv.span.hi()), String::new()),
+            ];
+            diag.multipart_suggestion(
+                format!("use {suggest} instead"),
+                suggestion,
+                Applicability::MachineApplicable,
+            );
```
Jarcho, line 70: There isn't a need for a multipart suggestion. Both of the replaced parts are side-by-side.

Jarcho (commented):
Sorry for the long delay. The implementation is fine, just have some small nits about it.

Can you change the name to something like `manual_is_variant_and`. Otherwise we'll end up with a pile of lints that do basically the same thing.

yuxqiu:
> Sorry for the long delay. The implementation is fine, just have some small nits about it.
> 
> Can you change the name to something like `manual_is_variant_and`. Otherwise we'll end up with a pile of lints that do basically the same thing.

No worries! Thank you for your review Jarcho! I have tried to resolved the problems you raised. Can you please see if the new code looks good?

clippy_lints/src/methods/manual_is_variant_and.rs
```diff
@@ -0,0 +1,68 @@
...
+    let is_option = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(map_recv), sym::Option);
+    let is_result = is_type_diagnostic_item(cx, cx.typeck_results().expr_ty(map_recv), sym::Result);
+    if !is_option && !is_result {
+        return;
+    }
+
+    // 3. the caller of `unwrap_or_default` is neither `Option<bool>` nor `Result<bool, _>`
+    // - we don't need to check the type of AdtDef here as we know from 2 that it must be either
+    //   `Option` or `Result`.
+    let map_arg_return_bool = match cx.typeck_results().expr_ty(unwrap_recv).kind() {
```
Jarcho, line 39: `expr_ty(expr)` would get the inner type directly.

yuxqiu: Thanks for the heads up!

Jarcho (approved):
Things look good. Just need to squash down the commits and then it's good.

yuxqiu:
> Things look good. Just need to squash down the commits and then it's good.

All done! Have a good holiday!

Jarcho:
Thank you. @bors r+

PR #11870, opened 2023-11-25
Use absolute path for `declare_tool_lint` in `declare_clippy_lint`

Alexendoo:
A minor change to hide that implementation detail

changelog: none


Jarcho:
I definitely read all those changes. @bors r+

Issue #9965, opened 2022-11-27
str::split instead of to_string().split()


lexeyOK:
### What it does

It suggests to use `str::split` instead of chaining `.to_string().split()`.

### Lint Name

str_split_instead_of_to_string_chain

### Category

style

### Advantage

shorter, there is no overhead as far as i know
<https://godbolt.org/z/ET7Y55P4E>

### Example

```rust
fn count_string(s: &str) -> usize {
    s.to_string().split(",").count()
}
```

Could be written as:

```rust
fn count_str(s: &str) -> usize {
    s.split(",").count()
}
```

Alexendoo:
Would be a good enhancement to `unnecessary_to_owned`

GuillaumeGomez:
Sending a PR shortly to handle this case.

PR #11871, opened 2023-11-25
Extend `UNNECESSARY_TO_OWNED` to handle `split`

GuillaumeGomez:
Fixes https://github.com/rust-lang/rust-clippy/issues/9965.

When you have `to_string().split('a')` or equivalent, it'll suggest to remove the `to_owned`/`to_string` part.

r? @flip1995 

changelog: Extend `UNNECESSARY_TO_OWNED` to handle `split`

### DIFF
--- a/tests/ui/unnecessary_to_owned_on_split.rs
+++ b/tests/ui/unnecessary_to_owned_on_split.rs
@@ -1,0 +1,21 @@
+#[allow(clippy::single_char_pattern)]
+
+fn main() {
+    let _ = "a".to_string().split('a').next().unwrap();
+    //~^ ERROR: unnecessary use of `to_string`
+    let _ = "a".to_string().split("a").next().unwrap();
+    //~^ ERROR: unnecessary use of `to_string`
+    let _ = "a".to_owned().split('a').next().unwrap();
+    //~^ ERROR: unnecessary use of `to_owned`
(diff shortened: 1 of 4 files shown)


tests/ui/unnecessary_to_owned_on_split.rs
```diff
@@ -0,0 +1,6 @@
+fn main() {
+    let _ = "a".to_string().split('a').next().unwrap();
+    //~^ ERROR: unnecessary use of `to_string`
+    let _ = "a".to_owned().split('a').next().unwrap();
```
llogiq, line 4: I'd like to see a test splitting by a &str and perhaps a counter-test that calls split by `()` which is implemented by a custom trait (if that is even possible?).

llogiq (commented):
The code looks good, some more tests would be nice though.

tests/ui/unnecessary_to_owned_on_split.rs
GuillaumeGomez: Implementing the `Pattern` trait is... quite a lot of code. So instead, I added checks for both `char` and `str` and also added the equivalent tests on `slice`.

GuillaumeGomez:
Updated!

llogiq:
Thank you!

@bors r+

